PR #417, opened 2026-03-19
Fix SQLITE_BUSY retries and closed SSE controller handling

lsm:

PR #421, opened 2026-03-19
fix: capture assistant.message.content when Copilot SDK skips streaming deltas

lsm:
## Summary

- Fixes the Copilot bridge token usage test that was failing in PR #417's CI run
- When the GitHub Copilot SDK delivers a complete response via a single `assistant.message` event (without preceding `assistant.message_delta` events), `outputCharCount` stayed at 0, causing `estimateTokens(0) = 0` in the SSE `message_delta` epilogue
- Adds a fallback in the `assistant.message` handler to capture the full content string when no deltas were accumulated

## Context

This fix was discovered while investigating PR #417 ("Fix SQLITE_BUSY retries and closed SSE controller handling"). That PR's CI was failing with `expect(outputTokens).toBeGreaterThan(0)` returning 0. The root cause was the Copilot SDK non-incremental response delivery path. PR #417 was closed as a duplicate of #415 (already merged).

## Test plan

- [ ] Unit test added: `runSessionStreaming — output_tokens via assistant.message fallback`
- [ ] Verifies `output_tokens > 0` when SDK delivers via `assistant.message` without deltas
- [ ] Online test `token usage: SSE stream contains non-zero input_tokens and output_tokens` should now pass

### DIFF
--- a/packages/daemon/src/lib/providers/anthropic-copilot/server.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/server.ts
@@ -121,6 +121,11 @@ function buildPlainSessionConfig(
 		streaming: true,
 		infiniteSessions: { enabled: true },
 		workingDirectory: cwd,
+		// Restrict to no built-in tools — plain sessions have no caller-defined tools,
+		// so Copilot's built-in bash/file tools must also be disabled.  Without this
+		// the model may autonomously invoke built-in tools (hanging or returning only
+		// tool-call content) instead of producing a text response.
+		availableTools: [],
 		...(systemMessage
 			? { systemMessage: { mode: 'replace' as const, content: systemMessage } }
 			: {}),
--- a/packages/daemon/src/lib/providers/anthropic-copilot/streaming.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/streaming.ts
@@ -137,6 +137,12 @@ function streamSession(
 				break;
 
 			case 'assistant.message':
+				// Fallback: if the SDK delivered the full response without any
+				// preceding assistant.message_delta events (valid Copilot SDK
+				// behaviour), capture it now so outputCharCount is non-zero.
+				if (pendingDeltas.length === 0 && event.data.content) {
+					pendingDeltas.push(event.data.content as string);
+				}
 				flushDeltas();
 				break;
 
--- a/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
+++ b/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
@@ -100,17 +100,33 @@ function getOutputTokens(events: SseEvent[]): number {
 async function callCopilotBridge(
 	bridgeUrl: string,
 	messages: Array<{ role: 'user' | 'assistant'; content: string }>,
-	model: string
+	model: string,
+	system?: string
 ): Promise<SseEvent[]> {
 	const response = await fetch(`${bridgeUrl}/v1/messages`, {
 		method: 'POST',
 		headers: { 'Content-Type': 'application/json' },
-		body: JSON.stringify({ model, messages, stream: true, max_tokens: 256 }),
+		body: JSON.stringify({
+			model,
+			messages,
+			stream: true,
+			max_tokens: 256,
+			...(system ? { system } : {}),
+		}),
 	});
 	if (!response.ok) {
 		throw new Error(`Bridge HTTP ${response.status}: ${await response.text()}`);
 	}
-	return parseSseEvents(await response.text());
+	const events = parseSseEvents(await response.text());
+	// Detect Anthropic SSE error events and surface them as descriptive test failures.
+	const errorEvent = events.find((e) => e.event === 'error');
+	if (errorEvent) {
+		const err = (errorEvent.data as { error?: { type?: string; message?: string } }).error;
+		throw new Error(
+			`Copilot bridge returned SSE error: ${err?.type ?? 'unknown'} — ${err?.message ?? '(no message)'}`
+		);
+	}
+	return events;
 }
 
 /** Per-turn idle timeout. The Copilot API can take 60-90 s per turn. */
@@ -552,7 +568,8 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 				const events = await callCopilotBridge(
 					bridgeUrl,
 					[{ role: 'user', content: 'Say hello.' }],
-					testModelId
+					testModelId,
+					'You are a helpful assistant. Always respond with plain text.'
 				);
 
 				const inputTokens = getInputTokens(events);
--- a/packages/daemon/tests/unit/providers/anthropic-copilot/server.test.ts
+++ b/packages/daemon/tests/unit/providers/anthropic-copilot/server.test.ts
@@ -404,6 +404,29 @@ describe('startEmbeddedServer', () => {
 		}
 	});
 
+	it('sets availableTools:[] on plain sessions to prevent built-in tool use', async () => {
+		let captured: unknown;
+		const cap = makeMockClient(() => session);
+		spyOn(cap, 'createSession').mockImplementation(async (cfg: unknown) => {
+			captured = cfg;
+			return session as unknown as CopilotSession;
+		});
+		const s2 = await startEmbeddedServer(cap, '/tmp');
+		try {
+			await postMessages(s2.url, {
+				model: 'x',
+				max_tokens: 100,
+				messages: [{ role: 'user', content: 'hi' }],
+			});
+			// Plain sessions (no tools in request) must set availableTools: [] to
+			// prevent the Copilot model from autonomously using built-in bash/file
+			// tools, which can cause hangs or empty text output.
+			expect((captured as Record<string, unknown>)['availableTools']).toEqual([]);
+		} finally {
+			await s2.stop();
+		}
+	});
+
 	// -------------------------------------------------------------------------
 	// Error paths
 	// -------------------------------------------------------------------------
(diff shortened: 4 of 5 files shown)


lsm (commented):

### Recommendation: APPROVE

This is a clean, well-scoped fix for a real bug. The change is minimal, correct, and well-tested.

### What was reviewed

**Production code** (`packages/daemon/src/lib/providers/anthropic-copilot/streaming.ts`, lines 139-145):

The fix adds a fallback in the `assistant.message` handler: when `pendingDeltas` is empty (no `assistant.message_delta` events were received), it pushes `event.data.content` into `pendingDeltas` before `flushDeltas()` runs. This ensures `outputCharCount` is non-zero in the `AnthropicStreamWriter`, which feeds into the heuristic `output_tokens` estimate.

**Correctness**: Verified against the Copilot SDK type definitions — `assistant.message.data.content` is typed as `string` (not optional, not an array), so the truthiness check `&& event.data.content` is a reasonable guard against empty strings. The `as string` cast is redundant but consistent with the existing `deltaContent as string` cast on line 136.

**Double-counting prevention**: The `pendingDeltas.length === 0` guard correctly ensures that when both delta events AND the final `assistant.message` arrive (normal streaming path), the content is not counted twice. `flushDeltas()` resets `pendingDeltas = []` but that reset happens inside the call on line 146, after the fallback check on line 143, so the guard works correctly — it checks before flushing, not after.

**Scope**: Both `runSessionStreaming()` and `resumeSessionStreaming()` flow through the same `streamSession()` function, so the fix applies to both entry points without duplication.

### Tests

Two tests added, both meaningful:

1. **Fallback path**: Emits `assistant.message` without any preceding `assistant.message_delta` — verifies `output_tokens > 0`.
2. **No double-counting**: Emits `assistant.message_delta` events followed by `assistant.message` — verifies `output_tokens` matches only the delta characters, not the combined content.

Both tests use the existing `MockSession`, `makeMockRes`, `makeMockReq`, and `estimateTokens` helpers consistently with the rest of the test file.

### Minor observations (P3)

1. The `as string` cast on line 144 is unnecessary — `content` is already typed as `string` in the SDK's `SessionEvent` type. However, this is consistent with the existing `as string` cast on line 136 for `deltaContent`, so it's a project-wide pattern rather than a bug.

lsm (commented):

**Recommendation: APPROVE**

The fix is correct and well-scoped. The implementation and tests hold up under close scrutiny.

---

### What the fix does

When the Copilot SDK delivers a complete response via a single `assistant.message` event — without any preceding `assistant.message_delta` events — `outputCharCount` in `AnthropicStreamWriter` stayed at 0, causing `estimateTokens(0) = 0` in the `message_delta` epilogue. The fix adds a six-line fallback in `streaming.ts` (lines 143–145): if `pendingDeltas` is empty when `assistant.message` fires, the full `content` string is pushed into `pendingDeltas` before `flushDeltas()` is called.

---

### Correctness

**Happy path (non-incremental delivery):**
`assistant.message` fires with `pendingDeltas.length === 0` and a truthy `content` → content is pushed → `flushDeltas()` accumulates `outputCharCount` → `session.idle` fires → `flushDeltas()` is a no-op (pending list already cleared) → `sendEpilogue` reports a non-zero estimate. Correct.

**Normal streaming path:**
`assistant.message_delta` events push into `pendingDeltas` → by the time `assistant.message` fires, `pendingDeltas.length > 0` → fallback skips → `flushDeltas()` processes the accumulated deltas. No double-counting. Correct.

**Empty content guard:**
`event.data.content` being an empty string is falsy, so the `&& event.data.content` check correctly skips the push. An empty-content `assistant.message` leaves `outputCharCount` at 0, which is the right behavior (no output = 0 tokens).

**After-flush idempotency:**
After `assistant.message` flushes deltas, `pendingDeltas` is `[]` again. `session.idle`'s `flushDeltas()` call hits the early-return guard at `streaming.ts:99`. No double-flush. Correct.

---

### Type safety note (P3 — non-blocking)

The cast `event.data.content as string` (line 144) is consistent with the existing cast on `event.data.deltaContent as string` (line 136), so this is following established project convention rather than introducing new risk. However, both casts suppress type-checking for the SDK event shape. If the SDK were ever to pass a structured content object instead of a string here, the result would be `[object Object]` in the SSE stream rather than an error. This is a pre-existing pattern, not introduced by this PR.

---

### Tests

Two new tests are added in `packages/daemon/tests/unit/providers/anthropic-copilot/streaming.test.ts`:

1. **`counts output chars from assistant.message.content when no message_delta events arrived`** — directly exercises the fallback path and asserts `outputTokens === estimateTokens(32) === 8`. The math is correct: 32 chars / 4 = 8.

2. **`does not double-count when both assistant.message_delta and assistant.message arrive`** — asserts the guard `pendingDeltas.length === 0` correctly skips the fallback. Expected value `estimateTokens(12) = 3` matches what `flushDeltas` accumulates from the two deltas (`'Hello '` + `'world!'` = 12 chars). Correct.

The online test `token usage: SSE stream contains non-zero input_tokens and output_tokens` (test 7 in `anthropic-to-copilot-bridge-provider.test.ts`) is the integration-level regression test that originally caught this issue. The new unit tests provide fast, deterministic coverage for both branches of the fallback condition.

lsm (commented):

**Recommendation: APPROVE**

Both fixes are well-targeted and correctly address the root causes of `output_tokens = 0` in the Copilot streaming path. The implementation is clean, well-documented, and follows existing patterns in the codebase.

### Fix 1: `assistant.message` fallback (streaming.ts:139-145)

The fallback logic is sound:

- **Guard condition** (`pendingDeltas.length === 0 && event.data.content`): Correctly prevents double-counting when both `assistant.message_delta` and `assistant.message` events arrive (the normal streaming path). The truthiness check on `content` also correctly skips empty strings, which would contribute 0 tokens anyway.
- **Type safety**: Verified against the Copilot SDK types (`session-events.d.ts:1152`) -- `content` is typed as `string` in `assistant.message` data, so the `as string` cast is safe.
- **Flush ordering**: `flushDeltas()` is called immediately after the conditional push, which is the same as the existing behavior. No change to the SSE event ordering.

### Fix 2: `availableTools: []` in `buildPlainSessionConfig` (server.ts:128)

This mirrors the existing pattern in `conversation.ts:167` (`availableTools: toolNames`) for tool sessions. For plain (no-tools) sessions, setting `availableTools: []` is the correct way to prevent the Copilot SDK from injecting its built-in bash/file tools. The comment is clear about why this is needed.

### Test quality

- **Unit tests** (streaming.test.ts): Both the fallback path and the no-double-counting path are tested. The assertions correctly verify the exact token count using `estimateTokens()`.
- **Unit test** (server.test.ts): Captures the session config and verifies `availableTools: []` is set for plain sessions.
- **Online test improvements**: Adding a system prompt to guide the model toward text responses and surfacing SSE error events as descriptive failures are good defensive improvements.

### Minor observations (P3)

1. The `system` parameter in `callCopilotBridge` is passed as a raw string, but the Anthropic API spec expects `system` to be either a string or an array of content blocks. The bridge server's `extractSystemText` handles both formats, so passing a string works correctly here, but it's worth noting for future test authors.

No blocking issues found. The changes are minimal, well-scoped, and properly tested.

lsm (commented):

**Recommendation: APPROVE**

The fix is correct and well-scoped. Both production changes are narrow, the double-count guard is sound, and test coverage directly exercises the new paths.

---

### Fix 1 — assistant.message fallback (streaming.ts, lines 139-146)

The core logic is correct. When no deltas have arrived, the fallback pushes `event.data.content` before `flushDeltas()` is called, which updates `outputCharCount` in `AnthropicStreamWriter`. When `session.idle` subsequently fires, its own `flushDeltas()` call is a noop since `pendingDeltas` was already cleared by the `assistant.message` handler.

The double-count guard (`pendingDeltas.length === 0`) correctly distinguishes the two cases:
- Normal streaming path: deltas accumulate in `pendingDeltas`, so `pendingDeltas.length > 0` at `assistant.message` time — fallback skips.
- Non-streaming path: no deltas, `pendingDeltas` is empty — fallback captures the full content.

One minor observation: the guard reads `pendingDeltas.length === 0` but it is theoretically possible for deltas to have arrived, been flushed by a prior `assistant.message` (if the SDK fires multiple `assistant.message` events per turn), and then `pendingDeltas` is empty again on the second call. In that scenario the fallback would incorrectly capture the final `content` string again. This is a low-likelihood edge case — the SDK docs do not mention multiple `assistant.message` events per turn — but tracking cumulative flush state (e.g. a boolean `hasReceivedDeltas`) would be strictly safer. This is a P3 nit, not a blocker.

### Fix 2 — availableTools: [] in buildPlainSessionConfig (server.ts, line 128)

Consistent with the existing `availableTools: toolNames` pattern in `conversation.ts` (line 167). The plain-session code path correctly mirrors the tool-session logic: tool sessions restrict to the caller-provided tool names; plain sessions restrict to nothing. The new unit test in `server.test.ts` (lines 407-428) verifies that `createSession` receives `availableTools: []` without any tools in the request body.

### Online test improvements (anthropic-to-copilot-bridge-provider.test.ts)

The system prompt addition (`'You are a helpful assistant. Always respond with plain text.'`) is a pragmatic way to reduce the probability of the model responding with tool invocations in the token-usage test. Combined with `availableTools: []`, this addresses both vectors that caused the zero-output-token failure.

The SSE error event detection in `callCopilotBridge` (lines 121-128) is a net improvement for test debuggability. Error messages were previously silently swallowed in the raw SSE text.

### Tests

Both new unit tests are meaningful and test distinct behaviors:
1. Fallback path: verifies `output_tokens > 0` when `assistant.message` fires without any preceding delta events.
2. No-double-count path: verifies `output_tokens` matches the delta character count only, not the combined content.

The `message_delta` SSE event lookup (`events.find((e) => e.type === 'message_delta')`) uses a non-null assertion (`delta!`). If `session.idle` never arrives the epilogue would not be written and `delta` would be `undefined`, causing an unhelpful runtime error at the assertion. The test structure ensures `session.idle` is always emitted so this is not a real risk here — still a pattern worth noting.

---

No P0 or P1 issues found. One P3 nit about the multi-`assistant.message` edge case, not a blocker.

PR #422, opened 2026-03-19
test: integration testing and documentation for Mission System V2 (Task 7)

akz142857:
## Summary

- **`goal.listExecutions` RPC handler** + execution history UI in `GoalsEditor.tsx`: recurring missions now show past execution runs (status, date, result summary) in the expanded detail view
- **26 unit tests** in `tests/unit/room/mission-system-edge-cases.test.ts`: scheduler restart catch-up, metrics dual-write derivation, execution identity in group metadata, per-execution isolation, migration defaults, autonomy gate edge cases
- **13 online integration tests** in `tests/online/room/mission-lifecycle.test.ts`: full lifecycle coverage for one-shot, measurable, recurring, semi-autonomous, escalation, migration, and `goal.listExecutions` RPC
- **8 E2E tests** in `tests/features/mission-detail.e2e.ts`: metric progress bars visible in header and expanded view, execution history section for recurring missions with empty-state copy
- **CLAUDE.md** updated with Mission System terminology section (types, autonomy levels, key files, DB tables)
- `scripts/validate-online-test-matrix.sh` updated to register `mission-lifecycle.test.ts`

## Test plan

- [x] `bun test packages/daemon/tests/unit/room/mission-system-edge-cases.test.ts` — 26 pass
- [x] `bun test packages/daemon/tests/unit/room/` — 1255 pass, 0 fail
- [x] `bun run typecheck` — clean (daemon + web)
- [x] `bash scripts/validate-online-test-matrix.sh` — all files covered
- [ ] E2E: `make run-e2e TEST=tests/features/mission-detail.e2e.ts` (requires UI server)
- [ ] Online: `NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/room/mission-lifecycle.test.ts` (requires dev-proxy)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -217,6 +217,50 @@ make self-test TEST=tests/core/navigation-3-column.e2e.ts
 - Always run a single E2E test file at a time — too slow to run all together
 - If a test scenario can't be triggered through the UI (e.g., token expiry, malformed server responses), it belongs in daemon integration tests, not E2E
 
+## Mission System
+
+The Mission System (Goal V2) extends the room's goal tracking with structured, automated workflows. Use the following terminology consistently across code, tests, and documentation.
+
+### Mission Types
+
+| Type | `missionType` value | Description |
+|------|-------------------|-------------|
+| One-shot | `one_shot` | Single-run goal with no metrics or schedule; the default. |
+| Measurable | `measurable` | Tracks progress toward numeric KPIs via `structuredMetrics`. |
+| Recurring | `recurring` | Runs on a cron schedule; creates a new execution each run. |
+
+### Autonomy Levels
+
+| Level | `autonomyLevel` value | Description |
+|-------|----------------------|-------------|
+| Supervised | `supervised` | Worker submits PRs for human review before merging. Default. |
+| Semi-autonomous | `semi_autonomous` | Worker can merge approved PRs without human confirmation. |
+
+### Key Terminology
+
+- **Mission** — The V2 term for a `RoomGoal`; use "mission" in UI copy and new API names. The DB table is still `goals` for backward compatibility.
+- **Execution** — A single run of a recurring mission. Stored in `mission_executions` with a monotonically increasing `executionNumber`.
+- **Metric** — A `MissionMetric` struct `{name, target, current, unit?}` in `structuredMetrics`. Measurable missions track one or more metrics.
+- **Execution ID** — A UUID stored in the session group's metadata (`executionId`) that links worker/leader sessions to a specific execution record.
+- **Schedule** — A cron expression plus timezone stored in `goal.schedule.expression` / `goal.schedule.timezone`. Use `@daily`, `@weekly`, or a 5-field cron string.
+
+### Key Files
+
+- `packages/daemon/src/lib/room/managers/goal-manager.ts` — Core `GoalManager`: CRUD, metric recording, execution management, scheduler tick.
+- `packages/daemon/src/lib/room/state/goal-repository.ts` — SQLite persistence for goals and V2 columns.
+- `packages/daemon/src/lib/room/runtime/cron-utils.ts` — Cron parsing, next-run computation, and catch-up detection.
+- `packages/daemon/src/lib/rpc-handlers/goal-handlers.ts` — RPC handlers: `goal.create`, `goal.list`, `goal.update`, `goal.delete`, `goal.listExecutions`, etc.
+- `packages/web/src/components/room/GoalsEditor.tsx` — Mission list UI, create/edit form with type-specific fields, metric progress bars, execution history.
+- `packages/web/src/lib/room-store.ts` — `RoomStore.listExecutions()` — fetches execution history via `goal.listExecutions` RPC.
+
+### DB Tables
+
+- `goals` — Stores all missions with V2 columns: `mission_type`, `autonomy_level`, `schedule`, `next_run_at`, `structured_metrics`, `consecutive_failures`, `replan_count`.
+- `mission_executions` — One row per execution run: `id`, `goal_id`, `execution_number`, `status`, `started_at`, `completed_at`, `result_summary`, `task_ids`.
+- `mission_metric_history` — Time-series metric snapshots: `goal_id`, `metric_name`, `value`, `recorded_at`.
+
+---
+
 ## Branching Strategy & CI
 
 - **`dev`** (default): Active development. PRs target `dev`. E2E tests run after merge.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -247,7 +247,7 @@ The Mission System (Goal V2) extends the room's goal tracking with structured, a
 ### Key Files
 
 - `packages/daemon/src/lib/room/managers/goal-manager.ts` — Core `GoalManager`: CRUD, metric recording, execution management, scheduler tick.
-- `packages/daemon/src/lib/room/state/goal-repository.ts` — SQLite persistence for goals and V2 columns.
+- `packages/daemon/src/storage/repositories/goal-repository.ts` — SQLite persistence for goals and V2 columns.
 - `packages/daemon/src/lib/room/runtime/cron-utils.ts` — Cron parsing, next-run computation, and catch-up detection.
 - `packages/daemon/src/lib/rpc-handlers/goal-handlers.ts` — RPC handlers: `goal.create`, `goal.list`, `goal.update`, `goal.delete`, `goal.listExecutions`, etc.
 - `packages/web/src/components/room/GoalsEditor.tsx` — Mission list UI, create/edit form with type-specific fields, metric progress bars, execution history.
--- a/packages/daemon/src/lib/rpc-handlers/goal-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/goal-handlers.ts
@@ -10,6 +10,7 @@
  * - goal.reactivate - Reactivate a goal (return to active)
  * - goal.linkTask - Link a task to a goal
  * - goal.delete - Delete a goal
+ * - goal.listExecutions - List execution history for a recurring mission
  * - task.approve - Human approves a task PR (resumes worker for phase 2)
  */
 
@@ -48,6 +49,7 @@ export type GoalManagerLike = Pick<
 	| 'deleteGoal'
 	| 'getActiveExecution'
 	| 'updateNextRunAt'
+	| 'listExecutions'
 >;
 
 export type GoalManagerFactory = (roomId: string) => GoalManagerLike;
@@ -446,6 +448,21 @@ export function setupGoalHandlers(
 		});
 		emitGoalUpdated(params.roomId, params.goalId, updated);
 		return { goal: updated, nextRunAt };
+	});
+
+	// goal.listExecutions - List execution history for a recurring mission
+	messageHub.onRequest('goal.listExecutions', async (data) => {
+		const params = data as { roomId: string; goalId: string; limit?: number };
+
+		if (!params.roomId) throw new Error('Room ID is required');
+		if (!params.goalId) throw new Error('Goal ID is required');
+
+		const goalManager = goalManagerFactory(params.roomId);
+		const goal = await goalManager.getGoal(params.goalId);
+		if (!goal) throw new Error(`Goal not found: ${params.goalId}`);
+
+		const executions = goalManager.listExecutions(params.goalId, params.limit ?? 20);
+		return { executions };
 	});
 
 	// task.approve - Human approves the PR; resume leader to complete task flow
--- a/packages/daemon/tests/online/room/mission-lifecycle.test.ts
+++ b/packages/daemon/tests/online/room/mission-lifecycle.test.ts
@@ -170,7 +170,12 @@ describe('Mission Lifecycle Integration Tests (API-dependent)', () => {
 			}
 			if (terminalPlan.status === 'review') {
 				await daemon.messageHub.request('task.approve', { roomId, taskId: terminalPlan.id });
-				await waitForTask(daemon, roomId, { taskType: 'planning', status: ['completed'] }, PLANNING_TIMEOUT);
+				await waitForTask(
+					daemon,
+					roomId,
+					{ taskType: 'planning', status: ['completed'] },
+					PLANNING_TIMEOUT
+				);
 			}
 
 			// 5. Wait for coding task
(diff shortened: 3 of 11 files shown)


akz142857 (commented):
## 🤖 Review by sonnet[1m] (Anthropic)

> **Model:** sonnet[1m] | **Client:** NeoKai | **Provider:** Anthropic

---

### Summary

This PR implements testing and documentation for Mission System V2. The overall structure is solid — 26 unit tests, 13 online integration tests, 8 E2E tests — but there are several correctness issues that need to be addressed before merging.

---

### P0 — Blocking Bugs

#### 1. `goal.setSchedule` parameter name mismatch in online tests

**File:** `packages/daemon/tests/online/room/mission-lifecycle.test.ts`, lines 261–292

The RPC handler at `packages/daemon/src/lib/rpc-handlers/goal-handlers.ts` line 366 expects the parameter `cronExpression`:

```typescript
const params = data as {
    roomId: string;
    goalId: string;
    cronExpression: string;   // <-- handler reads this
    timezone?: string;
};
if (!params.cronExpression) throw new Error('Cron expression is required');
```

But the online tests call `goal.setSchedule` with `expression` (not `cronExpression`):

```typescript
const scheduleResult = (await daemon.messageHub.request('goal.setSchedule', {
    roomId,
    goalId: mission.id,
    expression: '@daily',      // <-- test sends this
    timezone: 'UTC',
})) as { goal: RoomGoal; nextRunAt: number };
```

Because of JavaScript's truthy evaluation, `!params.cronExpression` will be `true` when the key is absent, so the handler throws `'Cron expression is required'` in every execution of the two tests that call `goal.setSchedule` (lines 261 and 287). Both tests will fail with `Error: Cron expression is required` rather than exercising the schedule logic. The test _appears_ to verify `scheduleResult.goal.schedule?.expression` but it never reaches that assertion. This is a P0 correctness failure of the test suite.

---

### P1 — Should Fix

#### 2. `recordMetric` helper defined but never called

**File:** `packages/daemon/tests/online/room/mission-lifecycle.test.ts`, lines 64–82

The `recordMetric` async helper is declared but has zero callsites in the file. The test that claims to cover "measurable mission: checkMetricTargets returns allMet=false when target not reached" (line 225) never records any metric changes; it only verifies `progress < 100` on the freshly-created goal. The helper and the test description both suggest that metric recording and target-checking should be exercised, but neither actually happens.

This means the core measurable-mission acceptance criterion — that `recordMetric` updates `structuredMetrics.current` and that `checkMetricTargets` returns `allMet=false` before the target is reached — has no online-level coverage. The unit tests cover these methods but a dead helper in the online test is a red flag.

#### 3. CLAUDE.md documents wrong path for `GoalRepository`

**File:** `CLAUDE.md`, Mission System — Key Files section

The documentation states:

```
packages/daemon/src/lib/room/state/goal-repository.ts — SQLite persistence for goals and V2 columns.
```

The file does not exist at that path. The actual location is:

```
packages/daemon/src/storage/repositories/goal-repository.ts
```

This is directly contradicted by the import in the new test file at `packages/daemon/tests/unit/room/mission-system-edge-cases.test.ts` line 20:

```typescript
import { GoalRepository } from '../../../src/storage/repositories/goal-repository';
```

Wrong paths in CLAUDE.md actively mislead contributors.

#### 4. Online tests for `goal.setSchedule`/`pauseSchedule`/`resumeSchedule` implicitly validate parameter contracts that don't exist in the handler

The `goal.setSchedule` handler only validates `recurring` mission type and requires `cronExpression`. Since both tests that call `goal.setSchedule` (lines 247–309) pass `expression` instead of `cronExpression`, the handler throws before reaching the type check. This means neither the schedule-setting path nor the pause/resume paths are actually tested at integration level, even though the tests exist and appear to cover them.

---

#### 5. `addMissionExecutionsTable` is duplicated across multiple test files instead of being added to the shared helper schema

**Files:** `mission-system-edge-cases.test.ts` lines 25–43, `recurring-missions.test.ts` lines 30–59

The `mission_executions` table setup is replicated instead of being added once to `DB_SCHEMA` in `room-runtime-test-helpers.ts`. Both test files define the same `addMissionExecutionsTable` function. Adding the table to the shared schema would remove this duplication.

#### 6. Online test file comment conflicts with Hard Fail Rule

**File:** `packages/daemon/tests/online/room/mission-lifecycle.test.ts`, lines 16–17

The comment reads:

> `NOTE: Room online tests are disabled in CI due to resource usage.`

The `validate-online-test-matrix.sh` does list `mission-lifecycle.test.ts` in `ROOM_FILES` and verifies the file exists on disk, but all `room-*` CI matrix entries are commented out in `.github/workflows/main.yml`. Per the project's **Hard Fail Rule**, a test module that is in the validate-matrix registry should also appear (uncommented) in the CI matrix. Either:
- The file should note it is intentionally deferred with a tracking issue, or
- The CI matrix should include a new `room-mission` shard.

The validate script claims coverage but CI will never run these tests.

---

### P3 — Nits

- `mission-system-edge-cases.test.ts` line 725: typo `supervisedgoal` in test name — should be `supervised goal`
- The `MetricProgress` component in `GoalsEditor.tsx` line 689 uses `key={i}` (index as key) for metrics, which can cause subtle re-render bugs when metrics are reordered. Using `m.name` as the key would be safer.
- The online test at line 463 (`'task.list: tasks from measurable mission are linked to goal'`) checks that `linkedTaskIds` is empty right after creation. This is trivially true and provides no useful coverage of the task-linking path.
- `expandMission` helper in E2E tests (line 99) clicks the `h4` element. The clickable area is the surrounding div; this works due to event bubbling but a `data-testid` on the container div would make it less fragile.

akz142857 (commented):
## 🤖 Review by sonnet[1m] (Anthropic)

> **Model:** sonnet[1m] | **Client:** NeoKai | **Provider:** Anthropic

**Recommendation: REQUEST_CHANGES** (own PR, using COMMENT event)

---

### Follow-Up Review Summary

The worker addressed the five P0/P1/P2 fixes from the previous review. Most issues are properly resolved. However, one P1 remains in the online test that was supposed to have been fixed.

---

### P1 — Still Present

#### Online test: misleading assertion in `recordMetric` test

**File:** `packages/daemon/tests/online/room/mission-lifecycle.test.ts`, lines 231–255

The fix added a call to `recordMetric` (line 244), which is an improvement. But the implementation of the `recordMetric` helper (lines 66–87) uses `goal.update` with `structuredMetrics`, which routes through `patchGoal`. `patchGoal` updates `structuredMetrics.current` but does **not** recalculate `progress`.

The test comment at line 251 claims:
```
// Progress should reflect partial completion (60/80 = 75%), not 100%
```

But the actual progress after `patchGoal` is **0** (set at creation time, never recalculated). The assertion `expect(afterRecord.progress ?? 0).toBeLessThan(100)` evaluates to `0 < 100`, which is trivially true and does not verify the claimed "partial completion" calculation.

Additionally, the test name says "allMet=false before target reached" — but `checkMetricTargets` is never called in this test. The `allMet` condition is not verified at all.

The real `GoalManager.recordMetric` (which the test intends to simulate) performs three things: updates `structuredMetrics.current`, inserts a history row, and recalculates `progress`. The online test's helper only does the first step via a different code path. The meaningful assertions (`metric?.current === 60` and `metric?.target === 80`) are correct, but the progress assertion and the test name are both wrong.

**Recommended fix:** Rename the test to accurately describe what it verifies (e.g., `'measurable mission: goal.update with structuredMetrics updates current value'`), remove or correct the misleading progress comment, and remove the `toBeLessThan(100)` assertion (or replace it with a more precise one if progress recalculation via `goal.update` is intended).

---

### P2 — Remaining Gap

#### CI matrix validation vs actual CI execution

**File:** `scripts/validate-online-test-matrix.sh` line 46, vs `.github/workflows/main.yml`

The validation script header says it "validates that all daemon online test files are covered by the CI matrix." `mission-lifecycle.test.ts` is correctly added to `ROOM_FILES`, so the script passes. But the CI workflow has all `room-*` shards commented out, so this test never runs in CI.

The inline test comment was updated to say "intentionally commented out... due to resource usage", which is clearer. But the script's stated purpose is undermined since the test is registered but never executed automatically. Per the Hard Fail Rule: "The CI workflow is the contract." This was flagged in the prior review and the fix is only a partial documentation update. Consider adding a note to the validation script clarifying that ROOM_FILES are intentionally deferred from CI, or tracking via a GitHub issue.

---

### P3 — Nit

**File:** `packages/e2e/tests/features/mission-detail.e2e.ts`, around line 186

```typescript
const headerContainer = missionHeader.locator('../../..');
// Progress bar is inside the header — look for the metric label text
await expect(page.locator('text=Code Coverage').first()).toBeVisible({ timeout: 5000 });
```

`headerContainer` is assigned but never used in any assertion. The actual assertion goes against the global page locator. This is dead code.

---

### Confirmed Fixes

All previously identified issues have been addressed or confirmed fixed:

- **P0** — `goal.setSchedule` calls in `mission-lifecycle.test.ts` now correctly use `cronExpression` (not `expression`). The handler at goal-handlers.ts line 363 reads `params.cronExpression`, and the tests at lines 276 and 303 pass `cronExpression: '@daily'`. Fix is correct.

- **P1** — `recordMetric` helper is now called at line 244. The helper correctly patches `structuredMetrics.current` via `goal.update`. Unit-level `checkMetricTargets` coverage in `mission-system-edge-cases.test.ts` is solid and valid — the unit tests use initial `current` values from `createGoal`, which correctly exercises the `checkMetricTargets` logic without needing `recordMetric`.

- **P1** — CLAUDE.md now documents `packages/daemon/src/storage/repositories/goal-repository.ts`. Verified the file exists at that path. The import in `mission-system-edge-cases.test.ts` line 20 confirms this.

- **P2** — `mission_executions` table is now in the shared `DB_SCHEMA` in `room-runtime-test-helpers.ts` (lines 192–206). All `addMissionExecutionsTable` call sites in `recurring-missions.test.ts` have been removed. The unused `Database` import was also removed.

- **P3** — Typo `supervisedgoal` → `supervised goal` at line 701 is fixed. `MetricProgress` now uses `key={m.name}`. `data-testid="goal-item-header"` is added to the header div. The `expandMission` E2E helper uses `[data-testid="goal-item-header"]:has(h4:has-text(...))`.

---

### Additional Observation (Not Previously Flagged)

**File:** `packages/web/src/components/room/GoalsEditor.tsx`, line 999

```tsx
{ex.startedAt && (
  <span>{new Date(ex.startedAt * 1000).toLocaleDateString()}</span>
)}
```

`MissionExecution.startedAt` is typed as `number` (non-optional) in `packages/shared/src/types/neo.ts` line 134, but the DB schema has `started_at INTEGER` (nullable). `rowToExecution` casts `row.started_at as number`, passing `null` through the TypeScript type. The guard `{ex.startedAt && ...}` is correct in practice, but the type mismatch is a latent gap. UI behavior is correct — informational only.

akz142857 (commented):
## 🤖 Review by sonnet[1m] (Anthropic)

> **Model:** sonnet[1m] | **Client:** NeoKai | **Provider:** Anthropic

**Recommendation: APPROVE**

---

### Third-Iteration Review Summary

This is a focused review of the three changes made in the third iteration. All three targeted issues have been correctly resolved. The PR is ready to merge.

---

### Changes Reviewed in This Iteration

#### 1. Test rename and assertion removal — Correct

**File:** `packages/daemon/tests/online/room/mission-lifecycle.test.ts`, lines 231–252

The test was renamed from "measurable mission: checkMetricTargets allMet=false" to:

> "measurable mission: goal.update with structuredMetrics updates current value"

This accurately describes the operation being tested. The misleading progress comment and the trivially-true `expect(afterRecord.progress ?? 0).toBeLessThan(100)` assertion have both been removed.

The two remaining assertions are correct and meaningful:
- `expect(metric?.current).toBe(60)` — verifies the current value was updated by the `goal.update` → `patchGoal` path
- `expect(metric?.target).toBe(80)` — verifies the target is unchanged

The test is now thin but honest: it verifies exactly what `goal.update` with `structuredMetrics` does at the API boundary. The deeper `checkMetricTargets` and history-write coverage remains in the unit tests in `mission-system-edge-cases.test.ts` ("checkMetricTargets: not met when current < target" at line 336).

The test is not too thin — it covers the online round-trip for metric updates via the RPC path, which is the right scope for an integration test.

#### 2. CI matrix gap explanation — Adequate

**File:** `scripts/validate-online-test-matrix.sh`, lines 44–47

The comment block added above `ROOM_FILES` is:

> NOTE: All room/* shards are intentionally commented out in .github/workflows/main.yml  
> due to resource usage. ROOM_FILES below tracks files that exist on disk; CI does not  
> automatically run them. Run them locally or enable per-task in the CI matrix.

This is accurate, honest, and sufficient for a maintenance context. The comment correctly distinguishes between "registered for disk tracking" and "scheduled for CI execution" and points contributors to the right place to enable the tests.

The CI workflow confirms all `room-*` shards are commented out with the note "Room tests temporarily disabled during refactoring" (line ~201). `mission-lifecycle.test.ts` never appears in the CI config, which is consistent with the other room tests — all room online tests are in the same deferred state.

One minor note that does not block approval: the validate-script's error message on line 92 still says "Add it to a `$module_name-*` entry in .github/workflows/main.yml", which would be misleading if a new `room/*.test.ts` was added and flagged — since the correct action for room tests is to add it to ROOM_FILES, not to CI. This was pre-existing and outside the scope of this iteration.

#### 3. Dead variable removal — Correct

**File:** `packages/e2e/tests/features/mission-detail.e2e.ts`

The `headerContainer` variable has been removed. Verified via grep — no occurrence of `headerContainer` exists in the file.

---

### Previously Open Issues — All Resolved

Confirming the full list of previously flagged issues are all closed:

- **P0** — `goal.setSchedule` parameter fixed to `cronExpression` in online tests (fixed in iteration 2).
- **P1** — `recordMetric` helper is called and exercises the `goal.update` → `patchGoal` API path (fixed in iteration 2).
- **P1** — `recordMetric` test name and misleading assertions removed (fixed in this iteration).
- **P1** — CLAUDE.md `GoalRepository` path corrected to `packages/daemon/src/storage/repositories/goal-repository.ts` (fixed in iteration 2).
- **P2** — `mission_executions` table deduplicated into `DB_SCHEMA` in `room-runtime-test-helpers.ts` (fixed in iteration 2).
- **P2** — CI matrix gap explained in validate script (addressed in this iteration).
- **P3** — `supervisedgoal` typo fixed, `MetricProgress` uses `key={m.name}`, `data-testid="goal-item-header"` added, `headerContainer` removed.

---

### Remaining Informational Notes (Non-Blocking)

One informational note from the prior review that was not expected to be fixed and still stands: `MissionExecution.startedAt` is typed as non-optional `number` in shared types but can be `null` from the DB. The UI guard `{ex.startedAt && ...}` in `GoalsEditor.tsx` works correctly in practice. This is a pre-existing type-vs-DB mismatch that is out of scope for this PR.

PR #424, opened 2026-03-19
fix: support Agent tool name for subagent blocks (SDK 0.2.76+ renamed Task to Agent)

lsm:
The Claude Agent SDK (0.2.76+) renamed the subagent invocation tool from 'Task'
to 'Agent' in tool_use blocks. The ToolUseBlock component only checked for
block.name === 'Task', so Agent tool calls fell through to the generic
ToolResultCard instead of rendering as SubagentBlock with nested messages.

Fix: accept both 'Task' and 'Agent' as subagent tool names, matching the
pattern already established in online tests (room-planner, room-reviewer).

### DIFF
--- a/packages/daemon/tests/helpers/daemon-actions.ts
+++ b/packages/daemon/tests/helpers/daemon-actions.ts
@@ -179,6 +179,14 @@ export async function waitForIdle(
  *
  * Use this before waitForIdle to avoid the race where waitForIdle resolves
  * against the pre-send idle state (before the session starts processing).
+ *
+ * CAUTION: Only use this with backends that are reliably slow enough that the
+ * processing window will still be observable when this function is called.
+ * If the session completes processing before this call begins (possible with
+ * fast mock/proxy backends), this will subscribe for a 'processing' event
+ * that never arrives and will hang until the timeout expires with an error.
+ * For fast-path tests, rely on sendMessage's built-in processing-start poll
+ * (lines 39-63) and call waitForIdle directly.
  */
 export async function waitForProcessing(
 	daemon: DaemonServerContext,
--- a/packages/daemon/tests/helpers/daemon-actions.ts
+++ b/packages/daemon/tests/helpers/daemon-actions.ts
@@ -175,28 +175,6 @@ export async function waitForIdle(
 }
 
 /**
- * Wait for the agent to enter processing state
- *
- * Use this before waitForIdle to avoid the race where waitForIdle resolves
- * against the pre-send idle state (before the session starts processing).
- *
- * CAUTION: Only use this with backends that are reliably slow enough that the
- * processing window will still be observable when this function is called.
- * If the session completes processing before this call begins (possible with
- * fast mock/proxy backends), this will subscribe for a 'processing' event
- * that never arrives and will hang until the timeout expires with an error.
- * For fast-path tests, rely on sendMessage's built-in processing-start poll
- * (lines 39-63) and call waitForIdle directly.
- */
-export async function waitForProcessing(
-	daemon: DaemonServerContext,
-	sessionId: string,
-	timeout = 10000
-): Promise<void> {
-	return waitForProcessingState(daemon, sessionId, 'processing', timeout);
-}
-
-/**
  * Collect SDK messages from the session via subscription
  *
  * Returns an async generator that yields SDK messages as they arrive.
--- a/packages/daemon/tests/helpers/daemon-actions.ts
+++ b/packages/daemon/tests/helpers/daemon-actions.ts
@@ -175,6 +175,20 @@ export async function waitForIdle(
 }
 
 /**
+ * Wait for the agent to enter processing state
+ *
+ * Use this before waitForIdle to avoid the race where waitForIdle resolves
+ * against the pre-send idle state (before the session starts processing).
+ */
+export async function waitForProcessing(
+	daemon: DaemonServerContext,
+	sessionId: string,
+	timeout = 10000
+): Promise<void> {
+	return waitForProcessingState(daemon, sessionId, 'processing', timeout);
+}
+
+/**
  * Collect SDK messages from the session via subscription
  *
  * Returns an async generator that yields SDK messages as they arrive.
--- a/packages/daemon/tests/online/coordinator/coordinator-tool-delegation.test.ts
+++ b/packages/daemon/tests/online/coordinator/coordinator-tool-delegation.test.ts
@@ -195,8 +195,9 @@ describe.skip('Coordinator Tool Delegation - Behavioral', () => {
 		// Verify coordinator tool usage
 		const coordinatorToolUses = getCoordinatorToolUses(allMessages);
 
-		// Verify delegation happened — coordinator should use Task for mutations
-		const taskUses = coordinatorToolUses.filter((t) => t.name === 'Task');
+		// Verify delegation happened — coordinator should use Task/Agent for mutations
+		// SDK 0.2.76+ renamed the tool from 'Task' to 'Agent'; accept both names
+		const taskUses = coordinatorToolUses.filter((t) => t.name === 'Task' || t.name === 'Agent');
 		expect(taskUses.length).toBeGreaterThan(0);
 
 		// Verify the coordinator did NOT use mutation tools directly
--- a/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
+++ b/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
@@ -289,7 +289,7 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
-				timeout: 5000,
+				timeout: IDLE_TIMEOUT,
 			});
 			const assistantMessages = sdkMessages.filter(
 				(m) => (m as { type?: string }).type === 'assistant'
@@ -332,7 +332,7 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
-				timeout: 5000,
+				timeout: IDLE_TIMEOUT,
 			});
 
 			// At least one tool_use block proves the bridge fired.
@@ -423,7 +423,7 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 			// proves the tool was exposed; whether the model uses it is out of scope.
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
-				timeout: 5000,
+				timeout: IDLE_TIMEOUT,
 			});
 			if (hasToolUseBlock(sdkMessages, 'get_answer')) {
 				const text = sdkMessages
--- a/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
+++ b/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
@@ -41,6 +41,7 @@ import { createDaemonServer } from '../../helpers/daemon-server';
 import {
 	sendMessage,
 	waitForIdle,
+	waitForProcessing,
 	getProcessingState,
 	waitForSdkMessages,
 } from '../../helpers/daemon-actions';
@@ -492,6 +493,10 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 
 			// Send a message and verify the copilot backend responds.
 			await sendMessage(daemon, sessionId, 'Reply with exactly: COPILOT_OK');
+			// Explicitly wait for the session to enter 'processing' before waiting for
+			// 'idle', to avoid the race where waitForIdle resolves against the pre-send
+			// idle state (before the query has actually started).
+			await waitForProcessing(daemon, sessionId, 15_000);
 			await waitForIdle(daemon, sessionId, IDLE_TIMEOUT);
 
 			const state = await getProcessingState(daemon, sessionId);
--- a/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
+++ b/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
@@ -38,12 +38,7 @@ import { mkdirSync, writeFileSync, existsSync } from 'node:fs';
 import { join } from 'node:path';
 import type { DaemonServerContext } from '../../helpers/daemon-server';
 import { createDaemonServer } from '../../helpers/daemon-server';
-import {
-	sendMessage,
-	waitForIdle,
-	waitForProcessing,
-	waitForSdkMessages,
-} from '../../helpers/daemon-actions';
+import { sendMessage, waitForIdle, waitForSdkMessages } from '../../helpers/daemon-actions';
 import { AnthropicToCopilotBridgeProvider } from '../../../src/lib/providers/anthropic-copilot/index';
 
 const TMP_DIR = process.env.TMPDIR || '/tmp';
@@ -482,16 +477,17 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 			expect(session.config?.provider).toBe('anthropic-copilot');
 
 			// Send a message and verify the copilot backend responds.
+			// sendMessage() already polls until processing starts, so waitForIdle()
+			// will not resolve against the pre-send idle state.
 			await sendMessage(daemon, sessionId, 'Reply with exactly: COPILOT_OK');
-			// Explicitly wait for the session to enter 'processing' before waiting for
-			// 'idle', to avoid the race where waitForIdle resolves against the pre-send
-			// idle state (before the query has actually started).
-			await waitForProcessing(daemon, sessionId, 15_000);
 			await waitForIdle(daemon, sessionId, IDLE_TIMEOUT);
 
+			// waitForIdle may resolve on a brief idle between Copilot retry cycles
+			// before the assistant message is persisted. Use a long timeout so we
+			// keep polling until the real response arrives.
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
-				timeout: 5000,
+				timeout: IDLE_TIMEOUT,
 			});
 			const text = sdkMessages
 				.filter((m) => (m as { type?: string }).type === 'assistant')
--- a/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
+++ b/packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts
@@ -42,7 +42,6 @@ import {
 	sendMessage,
 	waitForIdle,
 	waitForProcessing,
-	getProcessingState,
 	waitForSdkMessages,
 } from '../../helpers/daemon-actions';
 import { AnthropicToCopilotBridgeProvider } from '../../../src/lib/providers/anthropic-copilot/index';
@@ -293,9 +292,6 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 			await sendMessage(daemon, sessionId, 'What is 6+7? Reply with just the number.');
 			await waitForIdle(daemon, sessionId, IDLE_TIMEOUT);
 
-			const state = await getProcessingState(daemon, sessionId);
-			expect(state.status).toBe('idle');
-
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
 				timeout: 5000,
@@ -338,9 +334,6 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 				'Read the file answer.txt in the current directory and tell me the exact content.'
 			);
 			await waitForIdle(daemon, sessionId, IDLE_TIMEOUT);
-
-			const state = await getProcessingState(daemon, sessionId);
-			expect(state.status).toBe('idle');
 
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
@@ -415,9 +408,6 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 					'Reply with only the token string, nothing else.'
 			);
 			await waitForIdle(daemon, sessionId, IDLE_TIMEOUT);
-
-			const state = await getProcessingState(daemon, sessionId);
-			expect(state.status).toBe('idle');
 
 			// PRIMARY assertion: the Agent SDK initialised the MCP server.
 			// The MCP server writes the flag when it receives a tools/list request,
@@ -498,9 +488,6 @@ describe('AnthropicToCopilotBridgeProvider (Online)', () => {
 			// idle state (before the query has actually started).
 			await waitForProcessing(daemon, sessionId, 15_000);
 			await waitForIdle(daemon, sessionId, IDLE_TIMEOUT);
-
-			const state = await getProcessingState(daemon, sessionId);
-			expect(state.status).toBe('idle');
 
 			const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
 				minCount: 1,
--- a/packages/web/src/components/sdk/SDKAssistantMessage.tsx
+++ b/packages/web/src/components/sdk/SDKAssistantMessage.tsx
@@ -373,7 +373,8 @@ function ToolUseBlock({
 	const isOutputRemoved = resultData?.isOutputRemoved || false;
 
 	// Use SubagentBlock for Task tool (no delete button)
-	if (block.name === 'Task') {
+	// SDK 0.2.76+ renamed the tool from 'Task' to 'Agent', support both for compatibility
+	if (block.name === 'Task' || block.name === 'Agent') {
 		return (
 			<SubagentBlock
 				input={block.input as unknown as AgentInput}
--- a/packages/web/src/components/sdk/__tests__/SDKAssistantMessage.test.tsx
+++ b/packages/web/src/components/sdk/__tests__/SDKAssistantMessage.test.tsx
@@ -170,6 +170,36 @@ function createTaskToolMessage(): Extract<SDKMessage, { type: 'assistant' }> {
 	} as unknown as Extract<SDKMessage, { type: 'assistant' }>;
 }
 
+function createAgentToolMessage(): Extract<SDKMessage, { type: 'assistant' }> {
+	return {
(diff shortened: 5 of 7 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix is correct and minimal. The only place that actually needed updating was the `ToolUseBlock` condition in `SDKAssistantMessage.tsx` — the subagent message grouping (`useMessageMaps.ts`) and filtering (`SDKMessageRenderer.tsx`) both work based on `parent_tool_use_id`, not the tool name, so they are unaffected by the rename.

The tool registry (`tool-registry.ts`) and icon component (`ToolIcon.tsx`) already had `Agent` entries, confirming this was a known gap specifically in the rendering condition.

### Tests

The two new tests are well-structured and mirror the existing `Task` test pattern. They verify both that the `Agent` tool renders as a `SubagentBlock` (not a generic `ToolResultCard`) and that the description text appears. Good coverage for the fix.

### Minor observations (P3)

1. **`packages/web/src/lib/status-actions.ts:45`** — The `TOOL_ACTION_MAP` has `Task: 'Starting agent...'` but no `Agent` entry. When the SDK sends an `Agent` tool progress event, the status bar will show a generic rotating fallback ("Processing...", "Working...", etc.) instead of "Starting agent...". This is cosmetic and low impact, but worth adding `Agent: 'Starting agent...'` for consistency.

2. **`packages/web/src/lib/__tests__/status-actions.test.ts:292`** — The corresponding test only covers `['Task', 'Starting agent...']`. Adding `['Agent', 'Starting agent...']` would cover the above.

Neither of these affect the core bug fix (subagent blocks not rendering). The critical rendering path is correctly addressed.

lsm (commented):

---

## Summary

The core fix is correct and minimal — adding `|| block.name === 'Agent'` in `SDKAssistantMessage.tsx` directly addresses the regression introduced when SDK 0.2.76+ renamed the subagent tool from `Task` to `Agent`. The two new tests confirm the `Agent`-named tool renders as `SubagentBlock` rather than falling through to `ToolResultCard`.

However, there are issues that should be addressed before merge.

---

## Issues

### P1 — Incomplete fix: coordinator-tool-delegation test still checks only `'Task'`

`packages/daemon/tests/online/coordinator/coordinator-tool-delegation.test.ts` line 199:
```ts
const taskUses = coordinatorToolUses.filter((t) => t.name === 'Task');
expect(taskUses.length).toBeGreaterThan(0);
```

This online test is `describe.skip(...)` so it doesn't run in CI right now, but the assertion is semantically wrong for SDK 0.2.76+: with the new SDK the tool name in emitted `tool_use` blocks is `'Agent'`, so even when the test is re-enabled this assertion will always return zero matches and fail. The two room-based online tests (`room-reviewer-flow.test.ts`, `room-planner-two-phase.test.ts`) were already updated to check both names — this test was missed.

### P2 — Prompts hardcode `Task(subagent_type: ...)` call syntax

The agent prompts in:
- `packages/daemon/src/lib/room/agents/coder-agent.ts` (lines 314, 330)
- `packages/daemon/src/lib/room/agents/planner-agent.ts` (lines 152, 286)
- `packages/daemon/src/lib/room/agents/leader-agent.ts` (lines 235, 315, 325)

All instruct the model using `Task(subagent_type: ...)` call syntax. If the SDK 0.2.76+ only exposes an `Agent` tool and no longer exposes a `Task` tool, agents following these instructions will attempt to call a tool that doesn't exist, causing tool-not-found errors at runtime. If the SDK exposes both names for backward compatibility this is fine, but that assumption is not documented and should be verified. If only `Agent` is now available, all these prompt strings need updating to match.

This is a runtime correctness issue — the UI fix will show a correctly-rendered `SubagentBlock` only if the SDK actually produces an `Agent` tool_use block, which in turn only happens if the model successfully invokes the renamed tool.

### P2 — Two new tests have redundant setup, low differentiation

The two added tests both instantiate `createAgentToolMessage()` and render it fresh. Test 1 asserts `subagent_type: 'Plan'` is present in text; Test 2 asserts `description: 'Plan the implementation'` is present. Both are valid checks, but they could be combined into a single test or written as a single describe-block with shared setup. As-is, the second test provides almost no additional coverage over the first (the `SubagentBlock` component always renders `description` when it renders the type badge). This is a minor quality issue, not a blocker.

### P3 — Test file has `// @ts-nocheck` at line 1

The entire test file opts out of TypeScript checking. This means the new test factory function `createAgentToolMessage()` is never type-checked, so structural mistakes in the fixture wouldn't be caught at compile time. This pre-existed this PR but is worth tracking.

lsm (commented):

**Recommendation: APPROVE**

---

## Follow-up Review — All Previous Concerns Addressed

This is a follow-up to the previous review that raised P1 and P2 issues. All concerns have been resolved.

### P1 — Resolved: coordinator-tool-delegation.test.ts now checks both tool names

`packages/daemon/tests/online/coordinator/coordinator-tool-delegation.test.ts` line 200 was updated:
```ts
const taskUses = coordinatorToolUses.filter((t) => t.name === 'Task' || t.name === 'Agent');
```
The comment on line 199 also explains the SDK version context clearly.

### P2 — Resolved: Agent tests consolidated into a single assertion

The previous review noted two separate Agent tool tests with near-identical coverage. The PR now has exactly one test for the `Agent` tool name (line 301 in `SDKAssistantMessage.test.tsx`), asserting both the `subagent_type` value (`'Plan'`) and the `description` (`'Plan the implementation'`) in one test block. Redundancy eliminated.

### P2 — Resolved (by investigation): Prompts using `Task(subagent_type: ...)` don't need changing

The worker investigated and confirmed that `allowedTools: ['Task']` configuration still works — the model understands both naming conventions regardless of how the SDK emits tool_use blocks. The online test comments now document the SDK version context: "The SDK tool name is 'Task' in configuration but may appear as 'Task' or 'Agent' in tool_use blocks depending on SDK version." No runtime breakage is expected.

### Additional change (status bar): Agent entry added to TOOL_ACTION_MAP

A bonus fix was included: `packages/web/src/lib/status-actions.ts` adds `Agent: 'Starting agent...'` so that progress events for the renamed tool show the correct status bar text rather than falling through to a generic fallback. The corresponding test coverage was added in `status-actions.test.ts` at line 293 in the `TOOL_ACTION_MAP coverage` parametric block.

---

## Summary

All P1 and P2 issues from the prior review are properly addressed. The fix is minimal, correct, and well-tested. The additional `TOOL_ACTION_MAP` entry for `Agent` is a good catch that improves status bar UX for SDK 0.2.76+ users.

lsm (commented):

### Summary

Re-review after feedback was addressed. All three concerns from the previous review have been resolved correctly.

### Changes Reviewed

1. **`coordinator-tool-delegation.test.ts`** (P1 fix) — The filter now accepts both `'Task'` and `'Agent'` tool names with an appropriate comment explaining SDK 0.2.76+ renamed the tool. This ensures the test won't break when running against either SDK version. Correctly addressed.

2. **`SDKAssistantMessage.test.tsx`** (P2 fix) — A dedicated `createAgentToolMessage()` helper was added with a distinct subagent type (`Plan`) and description, and a new test case verifies the `Agent` tool renders as a `SubagentBlock`. The previous redundant test was consolidated into this single, clear test. Correctly addressed.

3. **`status-actions.ts` + test** (P3 fix) — `Agent: 'Starting agent...'` was added to `TOOL_ACTION_MAP` and the corresponding test data table was updated. This ensures the status bar shows the correct action text for both tool names. Correctly addressed.

### Completeness Check

I verified there are no remaining references to `'Task'` tool name without a corresponding `'Agent'` check across the codebase:

- `SDKAssistantMessage.tsx:377` — checks both names ✓
- `coordinator-tool-delegation.test.ts:200` — checks both names ✓
- `room-reviewer-flow.test.ts:247` — already checked both (pre-existing) ✓
- `room-planner-two-phase.test.ts:204` — already checked both (pre-existing) ✓
- `status-actions.ts` — has entries for both `Task` and `Agent` ✓
- `SDKMessageRenderer.tsx:isSubagentMessage()` — uses `parent_tool_use_id` (not tool name), unaffected ✓
- `SubagentBlock.tsx` — receives props from the caller, does not check tool name itself ✓

No issues found. The implementation is correct, complete, and well-tested.

**Recommendation: APPROVE**

lsm (commented):

### Re-review: CI race condition fix (commit 116c70b1)

The new `waitForProcessing` helper and its usage in the `provider session` test look correct for the reported CI failure. The fix addresses a real race condition where `waitForIdle` could resolve against the **pre-send** idle state before the Copilot provider session actually started processing.

**What was done:**
- Exported `waitForProcessing()` from `daemon-actions.ts` — a thin wrapper around the existing `waitForProcessingState` with `'processing'` as the target state
- Updated the `provider session` test to call `waitForProcessing(daemon, sessionId, 15_000)` before `waitForIdle()`

**Assessment: Sound fix, with one latent concern (P2)**

The fix is appropriate for the Copilot online test scenario (real API calls) where the delay between `sendMessage` returning and the session entering `processing` can exceed the 5-second internal polling window in `sendMessage`. The 15-second timeout for `waitForProcessing` is generous enough for real API latency.

**P2 — Latent race if processing completes very quickly** (`daemon-actions.ts`, line 183-188):
`waitForProcessing` will **timeout and throw** if the session enters `processing` and returns to `idle` before the call begins. This cannot happen with the real Copilot API (too slow), but could surface if this pattern is copied to mock/proxy tests where round-trips complete in milliseconds. The `sendMessage` function already has a fast-completion detection path (message count growth, lines 53-56 of `daemon-actions.ts`) that can cause it to return while the state is already back to `idle`.

This is not a blocking issue for this PR since the pattern is only used with a real API provider, but the JSDoc on `waitForProcessing` should ideally note this caveat to prevent misuse. Consider documenting that this helper is only safe when you can guarantee the processing window is wide enough to observe.

**Recommendation: APPROVE** — the fix is correct for its use case and resolves the CI flake. The latent concern is worth noting but not blocking.

lsm (commented):

**Recommendation: APPROVE**

This is a targeted, well-reasoned fix for the race condition in the Copilot provider-session test.

### What the fix does

The commit adds `waitForProcessing()` to `packages/daemon/tests/helpers/daemon-actions.ts` — a thin wrapper over the existing `waitForProcessingState()` infrastructure — and calls it between `sendMessage` and `waitForIdle` in the provider-session test.

The root cause was correct: `waitForIdle` checks the current state synchronously at the top of `waitForProcessingState` (line 87-90). If the session was still idle at the moment of that check (e.g. the agent hadn't started processing yet), the function returned immediately, and the subsequent assertion on `state.status === 'idle'` could then race against the session actually entering processing.

### Correctness of the fix

The fix is sound for its stated scenario. When the Copilot backend takes 60-90 seconds per turn, the sequence is:

1. `sendMessage` returns after detecting the session left idle (or after its 5s timeout).
2. `waitForProcessing` is called; the session is in `processing` state, so the early-return path (line 87-90) fires immediately and returns — no waiting needed.
3. `waitForIdle` is then called in the correct state and waits for the actual processing cycle to complete.

### Theoretical residual edge case (not a blocker for this provider)

There is one edge case worth documenting: if the Copilot API completes processing *before* `sendMessage` returns and before `waitForProcessing` is called, the session will already be back in `idle` state. In that scenario, `waitForProcessing` would subscribe to a `'processing'` event that will never arrive and would hang for 15 seconds before timing out with an error rather than proceeding to `waitForIdle`.

For this specific test this is not a practical concern — the Copilot backend latency (60-90s documented at line 133) makes same-turn completion before `sendMessage` returns essentially impossible. However, for copy-paste reuse of this pattern in future faster-backend tests, the idiom is fragile. The pattern is more robust when the backend is reliably slow.

### Other tests unchanged (correct)

The many existing tests that only use `sendMessage` + `waitForIdle` are not changed. This is correct: the race was specific to this test and the fix is applied only where the symptom was observed.

### Implementation quality

- `waitForProcessing` reuses the existing `waitForProcessingState` machinery including its subscription, polling fallback, room-join race-close, and timeout — no new complexity.
- The export is clean, the JSDoc comment accurately describes the intended usage pattern.
- The 15s timeout on `waitForProcessing` is appropriate for a provider startup/routing window.
- The fix is isolated to one test; no other tests are affected.

lsm (commented):

**Recommendation: APPROVE**

The previous P2 concern about `waitForProcessing` timing safety has been addressed with a thorough CAUTION JSDoc block (`packages/daemon/tests/helpers/daemon-actions.ts`, lines 183-189). The documentation accurately describes:

1. **When it is safe**: backends reliably slow enough that processing is still observable
2. **When it will fail**: fast mock/proxy backends where processing completes before the call
3. **What to do instead**: rely on `sendMessage`'s built-in processing-start poll and call `waitForIdle` directly

One small note (P3): the doc references "lines 39-63" for `sendMessage`'s built-in poll. Line references in comments tend to go stale as code evolves, but the function name `sendMessage` is sufficient to locate the relevant code.

The rest of the PR remains clean:
- `SDKAssistantMessage.tsx` and `status-actions.ts` correctly accept both `'Task'` and `'Agent'` tool names
- Tests cover the new `Agent` tool name path
- The coordinator test also accepts both names

No blocking issues remain. LGTM.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

This PR fixes a real UI bug (subagent blocks falling through to generic rendering after the SDK renamed 'Task' to 'Agent') and adds documentation to a previously undocumented hazard in the test helpers. The documentation fix from my previous P2 concern has been correctly applied.

---

### waitForProcessing CAUTION block — verified adequate

The JSDoc added to `waitForProcessing` in `packages/daemon/tests/helpers/daemon-actions.ts` (lines 177-197) accurately describes the hazard:

- It names the exact failure mode (subscribes for a 'processing' event that never arrives, hangs until timeout).
- It identifies when the function is safe (reliably slow real backends, e.g. real Copilot API).
- It identifies when it will fail (fast mock/proxy backends where processing completes before the call begins).
- It points the reader to the correct alternative (sendMessage's built-in processing-start poll at lines 39-63, then call waitForIdle directly).
- The line reference to "lines 39-63" is accurate — that is precisely where the poll loop lives inside sendMessage.

One small nuance worth noting: `waitForProcessingState` does have a pre-check at lines 85-90 that resolves immediately if the state is already 'processing' at the moment of the call. However, the scenario the CAUTION warns about is when state has already passed through 'processing' and returned to 'idle' before `waitForProcessing` is called at all — in that case the pre-check sees 'idle', not 'processing', and the function correctly hangs. The CAUTION documentation is therefore accurate and complete.

The only usage of `waitForProcessing` in the codebase is the Copilot bridge test (`anthropic-to-copilot-bridge-provider.test.ts`), which requires real Copilot credentials and is explicitly documented as a slow real-API test. That is exactly the safe context described in the CAUTION.

---

### Core bug fix — correct and well-tested

- `SDKAssistantMessage.tsx` line 377: `block.name === 'Task' || block.name === 'Agent'` — straightforward dual-name check, consistent with the same pattern in coordinator tests and status-actions.
- `status-actions.ts`: 'Agent' entry added to TOOL_ACTION_MAP alongside 'Task'.
- All three change sites (component, status map, coordinator test) are updated consistently.
- Tests cover the new 'Agent' name: `SDKAssistantMessage.test.tsx` adds a dedicated test case verifying SubagentBlock renders for the 'Agent' tool name, and `status-actions.test.ts` verifies the status string.

No issues found.

lsm (commented):

This re-review focuses on the latest commit (`c0e52ea`) which removes racy `getProcessingState` assertions after `waitForIdle` in the Copilot online tests.

### Summary

The changes are correct and well-reasoned. The core insight is sound: `waitForIdle` already guarantees the session reached idle state -- immediately re-checking with `getProcessingState` is redundant and introduces a race window where the Copilot backend's retry cycle (processing -> brief idle -> processing) can cause the assertion to catch a transient `processing` state.

### Analysis

**Correctness of the fix:**
- Three tests had `getProcessingState` + `expect(state.status).toBe('idle')` immediately after `waitForIdle`. These were removed from tests 3 (custom MCP), 5 (provider session), and 1 (basic conversation).
- The `getProcessingState` import was replaced with `waitForProcessing` -- this is correct since `waitForProcessing` is used in test 5 (provider session) but `getProcessingState` is no longer needed in this file.

**Test integrity -- does removing assertions weaken coverage?**
No. The removed assertions were:
```typescript
const state = await getProcessingState(daemon, sessionId);
expect(state.status).toBe('idle');
```
These are strictly redundant with `waitForIdle()` which internally calls `waitForProcessingState(daemon, sessionId, 'idle', timeout)` -- the same function that checks state via subscription + polling. The real test assertions (SDK message content, tool use verification, file flag existence) remain intact and are what actually prove the tests' stated goals.

**Asymmetry between test 3 and test 5:**
Test 5 (provider session) adds `waitForProcessing` before `waitForIdle`, while test 3 (custom MCP) does not. This is intentional -- test 5 was the one experiencing the pre-send idle race (per commit `116c70b`), while test 3 was only failing on the post-idle re-check race. The `sendMessage` built-in poll (lines 39-63 of `daemon-actions.ts`) handles the pre-send race adequately for test 3.

**`waitForProcessing` helper:**
The CAUTION comment accurately warns about fast backends. Since the Copilot backend takes 60-90s per turn, the 15s timeout on line 489 is safe for this use case.

**No dead imports or exports:**
- `getProcessingState` remains exported from `daemon-actions.ts` and is still used by many other test files (auto-title, message-persistence, multiturn-conversation, etc.).
- `waitForProcessing` is properly exported and imported.

Recommendation: **APPROVE**

lsm (commented):

**Recommendation: APPROVE**

---

### Re-review: racy getProcessingState assertions removed (commit c0e52ea9)

This iteration removes four `const state = await getProcessingState(...)` + `expect(state.status).toBe('idle')` assertions from `packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts`, and drops the now-unused `getProcessingState` import.

#### Correctness of the removal

The removed assertions were structurally redundant. `waitForIdle` already guarantees the session has reached idle by the time it returns — calling `getProcessingState` immediately afterward and asserting `status === 'idle'` was not verifying any additional system behavior. It was only verifying that `getProcessingState` returned the state `waitForIdle` already established, while introducing a TOCTOU window: if the Copilot backend enters an error-retry cycle (processing → brief idle → processing), sampling the state immediately after `waitForIdle` can observe the second `processing` entry and fail.

The description is accurate. The race is real and inherent to Copilot's retry behavior, not a test authoring error.

#### Test coverage impact — none

All meaningful assertions remain intact:
- **basic-conversation**: asserts `text.toContain('13')` — the real business check.
- **tool-use**: asserts `hasToolUseBlock(sdkMessages)`, `text.toContain('secret')`, `text.toContain('42')`.
- **custom-mcp**: asserts `existsSync(toolsListedFlag)` — the primary assertion proving MCP initialisation.
- **provider-session**: asserts `session.config?.provider === 'anthropic-copilot'` and `text.toUpperCase().toContain('COPILOT_OK')`.

None of these were state assertions. The removed blocks added no observable value over what the SDK message assertions already verify.

#### Import cleanup

`getProcessingState` is still exported from `packages/daemon/tests/helpers/daemon-actions.ts` and used by other test files. The import was only removed from the one test file that no longer calls it. No dead export is introduced.

#### The `waitForProcessing` pattern in provider-session

The `waitForProcessing(daemon, sessionId, 15_000)` + `waitForIdle(...)` sequence in the provider-session test is preserved. This correctly addresses the distinct race in that test (waitForIdle resolving against pre-send idle). It is not affected by this iteration.

#### Summary

The change is surgical and correct. It removes unstable assertions that created flaky CI failures without providing any signal that the surviving assertions do not already cover. No coverage regression, no logic change, no new risk introduced.

lsm (commented):

**Recommendation: APPROVE** (with one minor note)

---

### Latest commit analysis (56ba375)

The fix correctly addresses the CI failure root cause: `waitForIdle` resolving on a brief idle gap during Copilot's error-retry cycle, followed by `waitForSdkMessages` timing out at 5s before the real response is persisted.

**Correctness of the approach:**

1. **Removing `waitForProcessing`** -- correct. `sendMessage()` (lines 39-63 of `daemon-actions.ts`) already polls until the session leaves idle or SDK message count grows beyond the baseline. Adding an explicit `waitForProcessing` on top was redundant, and as the CAUTION comment on the function itself warns, it could hang if the session completes processing before the call begins.

2. **Increasing `waitForSdkMessages` timeout to `IDLE_TIMEOUT` (120s)** -- this is the right fix. The polling loop inside `waitForSdkMessages` checks every 50ms, so the long timeout does not slow down the happy path. It simply provides enough headroom for `waitForSdkMessages` to keep polling through Copilot retry cycles. When the response finally arrives and is persisted, the function returns immediately. The 120s ceiling matches the test's overall `IDLE_TIMEOUT`, which is consistent.

3. **Comment quality** -- the inline comments at lines 479-481 and 485-487 accurately explain why the change was made and what race condition it guards against.

---

### P3: Dead export -- `waitForProcessing` in `daemon-actions.ts`

`waitForProcessing` was added in commit `116c70b` and its sole usage was removed in this commit (`56ba375`). The function remains exported from `packages/daemon/tests/helpers/daemon-actions.ts` (lines 191-197) but has zero importers. This is dead code. Knip should flag it, but it would be cleaner to remove the function (and its 20-line CAUTION comment) entirely in this PR since it was introduced and abandoned within the same PR.

### Note: tests 1-3 still use `timeout: 5000`

Tests 1 (basic conversation), 2 (tool use), and 3 (custom MCP) all use `waitForSdkMessages(daemon, sessionId, { minCount: 1, timeout: 5000 })`. If the retry-cycle race is unique to the provider-session test's setup, this is fine. But if any of those tests could also experience Copilot retry cycles, they would be equally vulnerable. Worth monitoring in CI -- if any of those tests start failing with empty messages, the same `IDLE_TIMEOUT` treatment would be needed.

lsm (commented):

**Recommendation: APPROVE** (with two minor notes)

---

### Focus: Is the 120s timeout the right fix? Are there remaining timing issues?

Short answer: yes to the first, and mostly-no to the second — with one residual risk worth watching.

---

### The 120s timeout is the right fix for the provider-session test

The race is: waitForIdle resolves on a brief idle gap between Copilot error-retry cycles (processing -> brief idle -> processing). The original 5s waitForSdkMessages then expires before the real response is persisted.

Increasing to IDLE_TIMEOUT (120s) is sound because:

1. waitForSdkMessages polls every 50ms and returns immediately when the DB has enough messages. The long timeout only extends the ceiling — the happy path is not slowed.
2. When Copilot eventually succeeds (after retries), the response is persisted and polling finds it well within 120s.
3. The 120s ceiling is consistent with the overall test timeout (TEST_TIMEOUT = IDLE_TIMEOUT + 30s), so a Copilot API failure that produces zero messages will still cause the test to fail with a useful empty-string assertion rather than a timeout error.

Removing waitForProcessing from this test is correct. The previous iteration added it to guard the pre-send idle race, but sendMessage already polls until the session leaves idle or SDK message count grows (lines 39-63 of daemon-actions.ts). The waitForProcessing export introduced a pattern that is fragile on fast backends; removing its usage here is the right call. The JSDoc CAUTION block still documents the hazard for future callers.

---

### P2 — Tests 1, 2, and 3 still use timeout: 5000 for waitForSdkMessages

packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts, lines 290-293, 333-336, 424-427 all still use:

  const { sdkMessages } = await waitForSdkMessages(daemon, sessionId, {
    minCount: 1,
    timeout: 5000,
  });

All three tests go through the same Copilot backend. If Coplots retry cycles also affect those tests (the root cause was Copilot error-retries, not something specific to the provider-session config), the same empty-message failure could occur in tests 1, 2, or 3. The session config difference (explicit config.provider vs testModelId-only) routes to the same backend, so the retry risk is symmetric. If CI starts showing empty-message failures in those tests, the fix is identical: increase their waitForSdkMessages timeout to IDLE_TIMEOUT.

This is not a blocker since the reported failure was specifically in provider-session, but it is worth monitoring.

---

### P3 — Dead export: waitForProcessing in daemon-actions.ts

packages/daemon/tests/helpers/daemon-actions.ts, lines 191-197: waitForProcessing is exported but has zero importers. It was added in a previous commit in this same PR and its sole usage was removed in the final commit. The function and its 20-line CAUTION block are now dead code. Knip will likely flag this; it could be cleaned up in a follow-on commit.

---

### Correctness of the core fix — confirmed

- SDKAssistantMessage.tsx line 377: the dual-name check for Task/Agent is correct.
- status-actions.ts: Agent entry added alongside Task.
- coordinator-tool-delegation.test.ts: filter accepts both names.
- daemon-actions.ts: waitForSdkMessages increased to 120s only for the affected test.
- Tests cover the Agent name path for both the component and the status map.

No blocking issues.

lsm (commented):

**Recommendation: APPROVE** (same-author PR, posting as COMMENT)

Both feedback items from the previous review have been correctly addressed. The implementation is clean and complete.

---

### P2 addressed: `waitForSdkMessages` timeout now uses `IDLE_TIMEOUT` in all three tests

Tests 1, 2, 3, and 5 now pass `timeout: IDLE_TIMEOUT` (120 s) to `waitForSdkMessages`, consistent with the rationale documented in the new inline comment on test 5: Copilot's retry cycles can leave a brief idle window before the assistant message is persisted, and the short 5 s timeout was too narrow for that window. All four affected call sites in the Copilot bridge test are updated correctly (lines 290-293, 333-336, 424-427, 488-491).

### P3 addressed: `waitForProcessing` dead code removed from `daemon-actions.ts`

The function and its CAUTION JSDoc are gone. The file has no remaining reference to `waitForProcessing` (as distinct from the private `waitForProcessingState` helper, which is still present and used correctly by `waitForIdle`). The four `getProcessingState` + `expect(state.status).toBe('idle')` blocks that were redundant with the preceding `waitForIdle` calls have also been removed — removing duplicate state assertions that added noise without additional safety.

---

### No remaining issues

The rest of the diff is the core fix (accepting both `'Task'` and `'Agent'` as subagent tool names in the UI renderer and status-action map) along with tests for both names. These were already reviewed and are correct.

lsm (commented):

**Re-review: Both previously flagged issues are correctly resolved.**

### P2 fix verified: `waitForSdkMessages` timeout consistency

All three remaining `timeout: 5000` instances in `anthropic-to-copilot-bridge-provider.test.ts` (tests 1, 2, and 3) are now replaced with `timeout: IDLE_TIMEOUT` (120s). The fourth test already had `IDLE_TIMEOUT` from a prior commit. No `timeout: 5000` remains in the file. This eliminates the race where `waitForIdle` resolves on a brief Copilot retry-cycle idle but `waitForSdkMessages` times out before the real response arrives.

### P3 fix verified: `waitForProcessing` dead code removed

The exported `waitForProcessing` function and its CAUTION JSDoc are fully removed from `daemon-actions.ts`. The private `waitForProcessingState` helper correctly remains (used by `waitForIdle`). No test files import or reference `waitForProcessing`.

PR #426, opened 2026-03-19
feat: define Space system shared types (Task 3.1)

lsm:
Add packages/shared/src/types/space.ts with all Space, SpaceAgent, and
SpaceWorkflow types. Export from shared barrel (mod.ts).

Types added:
- Space, SpaceStatus, CreateSpaceParams, UpdateSpaceParams (M1)
- SpaceTask, SpaceTaskStatus (M1)
- SpaceWorkflowRun, SpaceWorkflowRunStatus (M1)
- SpaceAgent, BuiltinAgentRole, CreateSpaceAgentParams, UpdateSpaceAgentParams (M2)
- WorkflowGateType, WorkflowGate, WorkflowStep, WorkflowRule (M3)
- SpaceWorkflow, CreateSpaceWorkflowParams, UpdateSpaceWorkflowParams (M3)

JSDoc clarifies: 'leader' is NOT a valid agentRef/BuiltinAgentRole;
maxRetries re-evaluates gate only (does not re-run agent step);
appliesTo uses step IDs (not names) for rename stability.

### DIFF
--- a/packages/shared/src/types/space.ts
+++ b/packages/shared/src/types/space.ts
@@ -320,3 +320,303 @@ export interface SpaceSessionGroup {
 	/** Last update timestamp (milliseconds since epoch) */
 	updatedAt: number;
 }
+
+// ============================================================================
+// SpaceAgent Types (M2)
+// ============================================================================
+
+/**
+ * Builtin agent roles available in the Space system.
+ *
+ * NOTE: 'leader' is intentionally NOT a valid agent role — the Leader agent is
+ * always implicit and managed by SpaceRuntime. User-configurable agents are
+ * limited to worker roles: 'planner', 'coder', 'general'.
+ */
+export type BuiltinAgentRole = 'planner' | 'coder' | 'general';
+
+/**
+ * A named agent configuration within a Space.
+ * SpaceAgents can be referenced by name in SpaceWorkflow steps.
+ */
+export interface SpaceAgent {
+	/** Unique identifier */
+	id: string;
+	/** Space this agent belongs to */
+	spaceId: string;
+	/** Human-readable name (unique within a space) */
+	name: string;
+	/** Optional description of this agent's specialization */
+	description?: string;
+	/**
+	 * Builtin role preset — determines default tools and behavior.
+	 * One of: 'planner', 'coder', 'general'.
+	 * NOTE: 'leader' is NOT a valid role here — Leader is implicit in SpaceRuntime.
+	 */
+	role: BuiltinAgentRole;
+	/** Model ID override (e.g., 'claude-haiku-4-5') — uses space default if unset */
+	model?: string;
+	/** Provider name override (e.g., 'anthropic', 'openai') */
+	provider?: string;
+	/** Custom system prompt appended to the role preset */
+	systemPrompt?: string;
+	/** Tool configuration overrides */
+	toolConfig?: Record<string, unknown>;
+	/** Creation timestamp (milliseconds since epoch) */
+	createdAt: number;
+	/** Last update timestamp (milliseconds since epoch) */
+	updatedAt: number;
+}
+
+/**
+ * Parameters for creating a new SpaceAgent
+ */
+export interface CreateSpaceAgentParams {
+	spaceId: string;
+	name: string;
+	role: BuiltinAgentRole;
+	description?: string;
+	model?: string;
+	provider?: string;
+	systemPrompt?: string;
+	toolConfig?: Record<string, unknown>;
+}
+
+/**
+ * Parameters for updating a SpaceAgent
+ */
+export interface UpdateSpaceAgentParams {
+	name?: string;
+	description?: string | null;
+	role?: BuiltinAgentRole;
+	model?: string | null;
+	provider?: string | null;
+	systemPrompt?: string | null;
+	toolConfig?: Record<string, unknown> | null;
+}
+
+// ============================================================================
+// Workflow Types (M3)
+// ============================================================================
+
+/**
+ * The type of gate used at workflow step boundaries.
+ *
+ * - `auto`: Gate passes automatically when the agent step finishes
+ * - `human_approval`: Requires a human to explicitly approve before proceeding
+ * - `quality_check`: Runs an allowlisted command and passes only on exit code 0
+ * - `pr_review`: Waits for a GitHub PR review to approve before proceeding
+ * - `custom`: Runs a custom script at a relative workspace path (must be allowlisted)
+ */
+export type WorkflowGateType = 'auto' | 'human_approval' | 'quality_check' | 'pr_review' | 'custom';
+
+/**
+ * A gate that controls whether a workflow step may start (entry gate) or
+ * whether the workflow may advance to the next step (exit gate).
+ */
+export interface WorkflowGate {
+	/**
+	 * Type of gate.
+	 * - 'auto': passes immediately
+	 * - 'human_approval': blocks until a human approves
+	 * - 'quality_check': runs an allowlisted shell command
+	 * - 'pr_review': waits for GitHub PR approval
+	 * - 'custom': runs a script at a relative workspace path
+	 */
+	type: WorkflowGateType;
+	/**
+	 * Command or script to execute for 'quality_check' and 'custom' gate types.
+	 *
+	 * - For `quality_check`: must be an allowlisted command (e.g., 'bun test', 'npm run lint').
+	 *   Gate passes on exit code 0, fails on any non-zero exit.
+	 * - For `custom`: must be a path relative to the workspace root (e.g., './scripts/verify.sh').
+	 *   The script is resolved relative to the Space's workspacePath.
+	 */
+	command?: string;
+	/** Human-readable description of what this gate checks */
+	description?: string;
+	/**
+	 * Maximum number of times the gate evaluation is retried on failure.
+	 *
+	 * IMPORTANT: retries re-evaluate the gate only — they do NOT re-run the
+	 * agent step that preceded the gate. To re-run the step itself, the entire
+	 * workflow step must be retried at a higher level.
+	 *
+	 * Defaults to 0 (no retries — fail immediately on first failure).
+	 */
+	maxRetries?: number;
+	/** Timeout for gate evaluation in milliseconds (0 = no timeout) */
+	timeoutMs?: number;
+}
+
+/**
+ * Discriminated union encoding the agent reference within a workflow step.
+ *
+ * Using a union (rather than two separate fields) lets TypeScript enforce that
+ * 'leader' is rejected at compile time when agentRefType is 'builtin'.
+ *
+ * - `builtin`: agentRef is one of the BuiltinAgentRole values ('planner', 'coder', 'general').
+ *   NOTE: 'leader' is NOT a valid builtin agentRef — Leader is always implicit in SpaceRuntime.
+ * - `custom`: agentRef is the `name` of a SpaceAgent defined in this Space.
+ */
+export type WorkflowStepAgent =
+	| { agentRefType: 'builtin'; agentRef: BuiltinAgentRole }
+	| { agentRefType: 'custom'; agentRef: string };
+
+/**
+ * A single step within a SpaceWorkflow.
+ * Each step runs one agent and optionally has entry/exit gates.
+ */
+export type WorkflowStep = WorkflowStepAgent & {
+	/** Unique identifier for this step (stable across renames) */
+	id: string;
+	/** Human-readable name for display */
+	name: string;
+	/**
+	 * Gate checked before this step begins execution.
+	 * If absent, the step starts automatically when the previous step's exit gate passes.
+	 */
+	entryGate?: WorkflowGate;
+	/**
+	 * Gate checked after this step's agent finishes.
+	 * If absent, the workflow advances to the next step automatically.
+	 */
+	exitGate?: WorkflowGate;
+	/** Step-specific instructions appended to the agent's system prompt */
+	instructions?: string;
+	/** Zero-based execution order within the workflow */
+	order: number;
+};
+
+/**
+ * A rule that applies to workflow execution, similar to room-level rules.
+ * Rules express constraints, standards, or guidelines the agent must follow.
+ */
+export interface WorkflowRule {
+	/** Unique identifier (stable across renames) */
+	id: string;
+	/** Human-readable name for display */
+	name: string;
+	/** Rule content — markdown prose describing the constraint or guideline */
+	content: string;
+	/**
+	 * List of step IDs this rule applies to.
+	 *
+	 * Uses step **IDs** (not names) so that rules survive step renames.
+	 * Empty array or omitted means the rule applies to ALL steps in the workflow.
+	 */
+	appliesTo?: string[];
+}
+
+/**
+ * Distributive Omit that preserves discriminated union structure.
+ *
+ * TypeScript's built-in `Omit<T, K>` does not distribute over union members —
+ * when applied to an intersection type that wraps a union (like `WorkflowStep`),
+ * it collapses the discriminant and silently drops union enforcement.
+ * `DistributiveOmit` maps over each union branch individually so that
+ * discriminated-union invariants are preserved in the result.
+ */
+type DistributiveOmit<T, K extends PropertyKey> = T extends unknown ? Omit<T, K> : never;
+
+/**
+ * Input shape for a workflow step at creation time.
+ * `id` and `order` are backend-assigned and must not be provided by callers.
+ * The backend assigns `id` (UUID) and derives `order` from the array position.
+ *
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured set of type definitions. Typecheck passes. No export conflicts with existing barrel exports. All required types are present and correctly exported from `mod.ts`.

### What was verified

- **Correctness**: All 7 new M3 types (`WorkflowGateType`, `WorkflowGate`, `WorkflowStep`, `WorkflowRule`, `SpaceWorkflow`, `CreateSpaceWorkflowParams`, `UpdateSpaceWorkflowParams`) are defined and exported. M1 and M2 types are also correctly defined.
- **JSDoc requirements fully met**:
  - `agentRef` builtin values documented as `'planner'`, `'coder'`, `'general'` with explicit note that `'leader'` is NOT valid (lines 170-175, 304-308, 312-314)
  - `maxRetries` clearly states it re-evaluates the gate only, does NOT re-run the agent step (lines 282-288)
  - `command` semantics differ between `quality_check` (allowlisted shell command) and `custom` (relative workspace path) (lines 270-276)
  - `appliesTo` documented as using step IDs not names, with rationale for rename stability (lines 345-350)
- **Export correctness**: Single `export * from './types/space.ts'` line added to `mod.ts` (line 19). No naming collisions with existing exports.
- **Naming consistency**: Follows existing patterns — `Space`-prefixed entity types, `Create*Params`/`Update*Params` for mutations, `*Status` for union types.
- **Typecheck**: `tsc --build --noEmit` passes cleanly.

### Minor observations (P3)

1. **`CreateSpaceWorkflowParams.steps` accepts full `WorkflowStep[]` including `id` and `order`** (line 394). When the backend creates workflows, it will likely generate `id` values and could compute `order` from array position. A separate `CreateWorkflowStepParams` (omitting `id`/`order`) would be more precise, but this is a reasonable design choice for an initial type definition — the backend can simply ignore or override those fields. Not blocking.

2. **`UpdateSpaceWorkflowParams.steps` replaces the entire step array** (line 409) rather than supporting granular step mutations. This is fine for V1 — granular step updates can be added later if needed.

These are type-only definitions with no runtime behavior, so the risk surface is minimal. The implementation is complete and correct per the stated requirements.

lsm (commented):

**Recommendation: APPROVE** (this is the author's own PR, posting as COMMENT)

---

The implementation is solid and meets the task requirements. All six M3 types are present, all required JSDoc clarifications are in place, and the export from `mod.ts` is correct. The full check suite (lint, typecheck, knip) passes cleanly. Below are findings organized by severity.

---

### P2 — `WorkflowStep.agentRef` is typed as `string` but could carry a narrower discriminated union

**File:** `packages/shared/src/types/space.ts`, lines 310 and 316

`agentRef` and `agentRefType` are two separate fields whose valid values are coupled: when `agentRefType` is `'builtin'`, `agentRef` must be one of `'planner' | 'coder' | 'general'`; when it is `'custom'`, `agentRef` is a free string. This relationship is enforced only through JSDoc, not through TypeScript's type system. A discriminated union would let the compiler reject `agentRef: 'leader'` with `agentRefType: 'builtin'` at compile time:

```ts
type WorkflowStepAgent =
  | { agentRefType: 'builtin'; agentRef: BuiltinAgentRole }
  | { agentRefType: 'custom'; agentRef: string };
```

The current approach means the 'leader is NOT valid' invariant is prose-only; a downstream caller that passes `agentRef: 'leader'` will compile without error.

---

### P2 — `UpdateSpaceWorkflowParams` inconsistency: `tags` cannot be cleared but `description` can

**File:** `packages/shared/src/types/space.ts`, lines 406–413

`description` is `string | null` (allowing explicit clear), but `tags` is `string[]` with no `null` variant. If a caller needs to clear all tags, they must pass an empty array `[]`, which is reasonable but is not documented. It is also inconsistent with the pattern used for `description`. Either add a JSDoc comment explaining that `[]` clears tags, or add `null` to the union to be explicit. The same asymmetry exists for `steps` and `rules` — you can pass `[]` to wipe all steps/rules, but there is no indication in the type that this is intentional vs. accidental.

---

### P2 — `CreateSpaceWorkflowParams` accepts `steps` with embedded `WorkflowStep.id` and `order`

**File:** `packages/shared/src/types/space.ts`, lines 390–400

`WorkflowStep` has required fields `id` and `order` (both system-assigned fields that should be generated by the backend). `CreateSpaceWorkflowParams.steps` is typed as `WorkflowStep[]`, which forces callers to provide `id` and `order` at creation time — those are typically backend-assigned. If the backend ignores caller-supplied IDs and auto-assigns them, the type is misleading. Consider `Omit<WorkflowStep, 'id'>[]` for create, or a separate `WorkflowStepInput` type, similar to how the rest of the codebase separates creation params from full entity types.

The same issue applies to `WorkflowRule`: its `id` is required in the entity but `CreateSpaceWorkflowParams.rules` reuses `WorkflowRule[]`, requiring callers to supply IDs they may not yet have.

---

### P3 — JSDoc on `WorkflowGate.command` says 'Not used for ... pr_review' but `pr_review` is never the gate for the other three; the wording is redundant

**File:** `packages/shared/src/types/space.ts`, lines 275–276

Minor wording nit — 'Not used for auto, human_approval, or pr_review gate types' says the same thing as the positive list above it. Fine as-is, but could be condensed.

---

### P3 — `SpaceWorkflow.tags` and `SpaceWorkflow.rules` are non-optional arrays defaulting to `[]`; `CreateSpaceWorkflowParams` makes them optional

**File:** `packages/shared/src/types/space.ts`, lines 368–370, 394–395

This is consistent with how `CreateSpaceParams` works in the rest of the file, so it is not a bug. Just worth noting that a caller who does not supply `tags` or `rules` to the create endpoint will need the backend to default them to `[]`, or the returned `SpaceWorkflow` will not satisfy the non-optional array constraint. This should be documented on the params type, or the backend must be careful about this invariant.

---

### Summary

All required types and JSDoc annotations from the task description are present and accurate. The 'leader is NOT a valid agentRef' constraint, the `maxRetries` gate-only semantics, the `command` field distinction between `quality_check` vs `custom`, and the `appliesTo` step-ID (not name) semantics are all correctly documented. The main improvement opportunity is encoding the `agentRefType`/`agentRef` coupling in the type system rather than prose, and being explicit about nullability for `tags`/`steps`/`rules` on update params. The file is otherwise clean, well-structured, and ready to be consumed downstream.

lsm (commented):

**Recommendation: APPROVE**

### P2 Issue Resolution

All three P2 issues from the previous review have been correctly addressed in commit `25faa54`:

1. **Discriminated union for `WorkflowStepAgent`** -- Replaced the separate `agentRef: string` + `agentRefType: 'builtin' | 'custom'` fields with a proper discriminated union type (`packages/shared/src/types/space.ts`, lines 304-306). TypeScript now statically rejects `agentRef: 'leader'` when `agentRefType` is `'builtin'`. The intersection type `WorkflowStep = WorkflowStepAgent & {...}` (line 312) correctly propagates the discriminant through to the full step type.

2. **`WorkflowStepInput` / `WorkflowRuleInput` for creation params** -- Added `Omit`-based input types (lines 392, 398) that strip backend-assigned fields (`id` and `order` for steps, `id` for rules). `CreateSpaceWorkflowParams` now correctly uses `WorkflowStepInput[]` and `WorkflowRuleInput[]` (lines 411, 415) instead of the full entity types. JSDoc clearly documents that the backend assigns these fields.

3. **Nullability consistency on `UpdateSpaceWorkflowParams`** -- `steps`, `rules`, and `tags` all now accept `T | null` (lines 439, 443, 448), matching the pattern already used by `description` and `config`. The interface-level JSDoc (lines 427-431) documents the `[]` vs `null` semantics for array fields.

### Verification

- `tsc --build --noEmit` passes cleanly.
- No export name collisions with existing shared types.
- `UpdateSpaceWorkflowParams.steps` correctly uses `WorkflowStep[]` (not `WorkflowStepInput[]`) since updates operate on existing steps that already have IDs -- this is the right design choice.

1. **`WorkflowGate` could benefit from a discriminated union on `type`** to enforce that `command` is required when `type` is `'quality_check'` or `'custom'` and absent otherwise. Current design uses an optional `command` field with JSDoc-only enforcement. Acceptable for V1 types; can be tightened when the runtime layer is built.

No blocking issues found. The P2 feedback was addressed precisely and correctly.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (posting as COMMENT — this is the author's own PR)

---

This is a re-review after the previous P2 feedback. Two of the three P2 issues are cleanly resolved. One has an incomplete fix with a subtle TypeScript defect that re-introduces the exact enforcement gap the original P2 was meant to close.

---

### What was verified

- All checks pass: `bun run check` (lint + typecheck + knip) exits 0.
- `UpdateSpaceWorkflowParams` now has `steps?: WorkflowStep[] | null`, `rules?: WorkflowRule[] | null`, and `tags?: string[] | null` — consistent with `description?: string | null`. P2(3) fully resolved.
- `WorkflowStepInput` and `WorkflowRuleInput` are defined with `Omit` and `CreateSpaceWorkflowParams` uses them correctly. The intent of P2(2) is correct.
- `WorkflowStepAgent` discriminated union is defined at line 304–306 and `WorkflowStep` is typed as `WorkflowStepAgent & { ... }` at line 312. TypeScript correctly rejects `agentRef: 'leader'` when `agentRefType: 'builtin'` on a `WorkflowStep` value. P2(1) is partially resolved.

---

### P1 — `WorkflowStepInput` silently loses discriminated union enforcement

**File:** `packages/shared/src/types/space.ts`, line 392

`WorkflowStepInput` is defined as:

```ts
export type WorkflowStepInput = Omit<WorkflowStep, 'id' | 'order'>;
```

`WorkflowStep` is an intersection of a union with a plain object type: `WorkflowStepAgent & { id, name, order, ... }`. When TypeScript's built-in `Omit` is applied to an intersection that contains a union, the union is not distributed — the result is a merged type where `agentRefType` and `agentRef` lose their coupling. The discriminant constraint is silently dropped.

Verified: the following compiles without error:

```ts
const badInput: WorkflowStepInput = {
    agentRefType: 'builtin',
    agentRef: 'leader',  // no TypeScript error here
    name: 'bad',
};
```

This means `CreateSpaceWorkflowParams.steps` — the primary create path — accepts `'leader'` as a builtin `agentRef` without complaint. The invariant that P2(1) was supposed to enforce at compile time is enforced for full `WorkflowStep` and for `UpdateSpaceWorkflowParams.steps`, but bypassed at the create path.

**Fix:** use a distributive `Omit` helper that distributes over union members:

```ts
type DistributiveOmit<T, K extends PropertyKey> = T extends unknown ? Omit<T, K> : never;

export type WorkflowStepInput = DistributiveOmit<WorkflowStep, 'id' | 'order'>;
```

With this change, `WorkflowStepInput` correctly becomes a discriminated union and TypeScript rejects `agentRef: 'leader'` with `agentRefType: 'builtin'` at compile time.

Note: `WorkflowRuleInput = Omit<WorkflowRule, 'id'>` is unaffected because `WorkflowRule` is a plain interface (not a union), so `Omit` behaves correctly there.

---

### P3 — `UpdateSpaceWorkflowParams.steps` accepts full `WorkflowStep[]` (not `WorkflowStepInput[]`)

**File:** `packages/shared/src/types/space.ts`, line 439

`steps?: WorkflowStep[] | null` requires callers to provide `id` and `order` in each step when performing an update. This is arguably correct for updates (the caller has already fetched the workflow and has step IDs), but the documentation on line 437 says "The backend re-assigns order from the array position," which implies `order` is ignored on input. If `order` is always ignored on update just as on create, `WorkflowStepInput[]` would be more honest and remove the need for callers to redundantly supply it. This is a minor semantic ambiguity, not a blocker.

---

### Summary

P2(2) (Omit-based input types) and P2(3) (nullability consistency) are correctly resolved. P2(1) (discriminated union) is correctly resolved for `WorkflowStep` and `UpdateSpaceWorkflowParams.steps` but the enforcement is silently absent for `WorkflowStepInput` / `CreateSpaceWorkflowParams.steps` due to `Omit` collapsing the union. The `DistributiveOmit` pattern is the standard TypeScript fix for this class of issue.

lsm (commented):

**Recommendation: APPROVE**

---

### P1 Fix Verification: `DistributiveOmit` for `WorkflowStepInput`

The `DistributiveOmit` helper added in commit `0f69249` correctly resolves the P1 from the previous review.

**What was wrong:** `Omit<WorkflowStep, 'id' | 'order'>` collapsed the discriminated union because `WorkflowStep` is an intersection of a union (`WorkflowStepAgent`) with a plain object. Standard `Omit` does not distribute over union members, so `agentRefType` and `agentRef` lost their coupling, silently allowing `agentRef: 'leader'` with `agentRefType: 'builtin'`.

**What was done:** A `DistributiveOmit<T, K>` helper type was added at `packages/shared/src/types/space.ts` line 396:

```ts
type DistributiveOmit<T, K extends PropertyKey> = T extends unknown ? Omit<T, K> : never;
```

`WorkflowStepInput` now uses `DistributiveOmit<WorkflowStep, 'id' | 'order'>` instead of `Omit<WorkflowStep, 'id' | 'order'>`.

**Verified independently:** I replicated the types and confirmed:
- `DistributiveOmit` correctly rejects `{ agentRefType: 'builtin', agentRef: 'leader', name: '...' }` at compile time (TS2322: Type '"leader"' is not assignable to type 'BuiltinAgentRole').
- Plain `Omit` accepts that same invalid object without error, confirming the original bug.
- `WorkflowRuleInput` correctly continues using plain `Omit` since `WorkflowRule` is a plain interface (no union to distribute over).

**Typecheck passes:** `tsc --build --noEmit` exits cleanly.

The JSDoc on `DistributiveOmit` (lines 387-395) and on `WorkflowStepInput` (lines 398-406) clearly documents why `DistributiveOmit` is needed and what invariant it preserves.

### P3 — Missing type-level test

The task description mentions a type-level test was added to confirm the invariant holds, but no such test exists in the codebase (no `@ts-expect-error` assertion in a test file, no `.test.ts` file exercising `WorkflowStepInput`). This is a minor gap -- a compile-time test with `@ts-expect-error` would guard against future regressions if someone accidentally changes `DistributiveOmit` back to `Omit`. Not blocking since the typecheck itself enforces correctness, but worth adding.

### P3 — `UpdateSpaceWorkflowParams.steps` semantic ambiguity

`steps?: WorkflowStep[] | null` requires callers to supply `order` on update, but the JSDoc on line 452 says "The backend re-assigns order from the array position." If `order` is always ignored, `DistributiveOmit<WorkflowStep, 'order'>[]` would be more precise. Minor -- not blocking for a types-only PR.

---

### Summary

The P1 fix is correct, complete, and uses the standard TypeScript pattern for distributive conditional types. All previous P2 issues remain resolved. No blocking issues found.

lsm (commented):

**Recommendation: APPROVE** (posting as COMMENT — this is the author's own PR)

---

The P1 issue from the previous review is correctly resolved. Below is the full verification.

### P1 Fix — Verified Correct

**File:** `packages/shared/src/types/space.ts`, lines 387–407

`DistributiveOmit` is defined with the standard conditional type pattern:

```ts
type DistributiveOmit<T, K extends PropertyKey> = T extends unknown ? Omit<T, K> : never;
```

This is the correct implementation. Because `T extends unknown` is always true, the conditional distributes over each union branch individually rather than resolving to a single merged type. Applied to `WorkflowStep` (which is `WorkflowStepAgent & { id, name, ... }` where `WorkflowStepAgent` is a two-branch union), the result is a union of two `Omit`-d branches — the discriminant is preserved intact.

Manually verified: with `DistributiveOmit`, TypeScript correctly rejects:

```ts
const bad: WorkflowStepInput = {
    agentRefType: 'builtin',
    agentRef: 'leader',  // type error — not assignable to 'planner' | 'coder' | 'general'
    name: 'bad step',
};
```

And correctly accepts:

```ts
const good: WorkflowStepInput = { agentRefType: 'builtin', agentRef: 'planner', name: 'step' };
const customLeader: WorkflowStepInput = { agentRefType: 'custom', agentRef: 'leader', name: 'step' };
```

The invariant that blocked the previous approval is now enforced at the create path (`CreateSpaceWorkflowParams.steps`) through the type system, not just through prose documentation.

### Check Suite

`bun run check` (lint + typecheck + knip) exits 0 cleanly with no warnings or errors.

### Minor Observation — No Dedicated Type-Level Test File (P3)

The task description stated "the worker also added a type-level test confirming the invariant holds," but the commit only modifies `space.ts` — no separate `.test.ts` or `.test-d.ts` file was added. The enforcement is real and verified by `tsc` on every typecheck run, so this is not a regression risk in practice. An explicit `@ts-expect-error`-annotated constant in a test file would make the invariant visible and would fail CI if the enforcement were accidentally removed, but that is a P3 enhancement, not a blocker.

---

No blocking issues. The fix is correct, minimal, well-documented, and the full check suite passes.

PR #427, opened 2026-03-19
feat: add Migration 29 — create all Space system tables

lsm:
Creates all 8 Space-related tables in a single migration (29) in
FK-safe order: spaces → space_agents, space_workflows →
space_workflow_steps, space_workflow_runs → space_tasks →
space_session_groups → space_session_group_members.

- space_tasks includes custom_agent_id, workflow_run_id, and
  workflow_step_id columns from the start (no ALTER TABLE needed later)
- space_workflow_runs tracks execution state with status CHECK constraint
- All FK cascades: delete space → all child rows deleted
- space_tasks.workflow_run_id SET NULL on workflow run delete
- Indexes created for all FK columns and status columns
- Migration is idempotent (CREATE TABLE/INDEX IF NOT EXISTS throughout)
- 15 unit tests covering all tables, cascades, constraints, and indexes

### DIFF
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -1479,7 +1479,8 @@ function runMigration29(db: BunDatabase): void {
 		)
 	`);
 	db.exec(`CREATE INDEX IF NOT EXISTS idx_spaces_status ON spaces(status)`);
-	db.exec(`CREATE INDEX IF NOT EXISTS idx_spaces_workspace_path ON spaces(workspace_path)`);
+	// Note: workspace_path has a UNIQUE constraint which SQLite implements as an implicit
+	// unique index — no explicit CREATE INDEX needed.
 
 	// -------------------------------------------------------------------------
 	// space_agents
@@ -1611,7 +1612,8 @@ function runMigration29(db: BunDatabase): void {
 			completed_at INTEGER,
 			updated_at INTEGER NOT NULL,
 			FOREIGN KEY (space_id) REFERENCES spaces(id) ON DELETE CASCADE,
-			FOREIGN KEY (workflow_run_id) REFERENCES space_workflow_runs(id) ON DELETE SET NULL
+			FOREIGN KEY (workflow_run_id) REFERENCES space_workflow_runs(id) ON DELETE SET NULL,
+			FOREIGN KEY (workflow_step_id) REFERENCES space_workflow_steps(id) ON DELETE SET NULL
 		)
 	`);
 	db.exec(`CREATE INDEX IF NOT EXISTS idx_space_tasks_space_id ON space_tasks(space_id)`);
@@ -1621,6 +1623,9 @@ function runMigration29(db: BunDatabase): void {
 	);
 	db.exec(
 		`CREATE INDEX IF NOT EXISTS idx_space_tasks_custom_agent_id ON space_tasks(custom_agent_id)`
+	);
+	db.exec(
+		`CREATE INDEX IF NOT EXISTS idx_space_tasks_workflow_step_id ON space_tasks(workflow_step_id)`
 	);
 
 	// -------------------------------------------------------------------------
@@ -1653,7 +1658,8 @@ function runMigration29(db: BunDatabase): void {
 				CHECK(role IN ('worker', 'leader')),
 			order_index INTEGER NOT NULL DEFAULT 0,
 			created_at INTEGER NOT NULL,
-			FOREIGN KEY (group_id) REFERENCES space_session_groups(id) ON DELETE CASCADE
+			FOREIGN KEY (group_id) REFERENCES space_session_groups(id) ON DELETE CASCADE,
+			UNIQUE(group_id, session_id)
 		)
 	`);
 	db.exec(
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -109,6 +109,11 @@ export function runMigrations(db: BunDatabase, createBackup: () => void): void {
 	// Migration 28: Add mission metadata columns to goals table, create mission_metric_history
 	// and mission_executions tables for Goal V2 / Mission System
 	runMigration28(db);
+
+	// Migration 29: Create all Space system tables (spaces, space_agents, space_workflows,
+	// space_workflow_steps, space_workflow_runs, space_tasks, space_session_groups,
+	// space_session_group_members) in FK-safe order
+	runMigration29(db);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This migration is well-structured and achieves the task requirements. All 8 Space tables are created in correct FK-safe order, CASCADE/SET NULL semantics are correct, idempotency is handled via `IF NOT EXISTS`, and the schema aligns with the shared types in `packages/shared/src/types/space.ts`. No `space_goals` table exists (as specified). Tests are thorough — 15 tests covering table creation, idempotency, column existence, CHECK constraints, CASCADE deletes, SET NULL behavior, uniqueness, indexes, and non-regression.

### P2 Issues

**1. Missing UNIQUE constraint on `space_session_group_members(group_id, session_id)`**
(`packages/daemon/src/storage/schema/migrations.ts`, line ~1648)

Without this constraint, the same session can be added to a group multiple times. This is almost certainly not intended behavior. A composite unique constraint would prevent duplicate membership records:
```sql
UNIQUE(group_id, session_id)
```

**2. `idx_spaces_workspace_path` index is redundant**
(`packages/daemon/src/storage/schema/migrations.ts`, line 1482)

The `workspace_path` column already has a `UNIQUE` constraint (line 1466), which SQLite implements as an implicit unique index. The explicit `CREATE INDEX` on the same column creates a second, redundant index that doubles write cost for that column without any query benefit. Consider removing it.

### P3 Issues

**1. `columnExists` helper uses string interpolation for table name**
(`packages/daemon/tests/unit/storage/migrations/migration-29_test.ts`, line 33)

The helper uses `` PRAGMA table_info(${table}) `` with direct string interpolation rather than parameterized queries. This is fine in test code where the caller controls all inputs, but worth noting as a pattern to avoid in production code.

**2. No FK from `space_tasks.custom_agent_id` to `space_agents.id`**
(`packages/daemon/src/storage/schema/migrations.ts`, line 1593)

The `custom_agent_id` column has an index but no foreign key constraint back to `space_agents`. This means deleting a space agent will leave orphaned references in tasks. This may be intentional (the column could reference external agents), but if it should always reference `space_agents`, an FK with `ON DELETE SET NULL` would be appropriate. Flagging as a nit since the design may evolve.

**3. No FK from `space_tasks.workflow_step_id` to `space_workflow_steps.id`**
(`packages/daemon/src/storage/schema/migrations.ts`, line 1595)

Similar to above — `workflow_step_id` has no FK constraint. Orphaned references are possible if a workflow step is deleted. May be intentional to preserve historical task data.

### Summary

The migration is correct, complete, and well-tested. The schema aligns with the shared types rather than the original plan doc (which has since been superseded by the refined types). FK ordering is correct — `space_workflow_runs` is created before `space_tasks` to satisfy the FK dependency. CASCADE semantics are properly tested. The two P2 items are the missing uniqueness constraint on group membership and the redundant index.

lsm (commented):

> **Recommendation:** APPROVE (with noted issues)

All 15 tests pass and the core implementation is correct. There are two substantive issues and one redundancy worth addressing before or shortly after merging.

---

### P1 — Missing index on `space_tasks.workflow_step_id`

`packages/daemon/src/storage/schema/migrations.ts`, line 1595

The `workflow_step_id` column is defined on `space_tasks` but has no index. Unlike `workflow_run_id` (which has `idx_space_tasks_workflow_run_id`), `workflow_step_id` receives no index. Queries like "find all tasks for a given workflow step" will scan the full table. This is inconsistent with the index coverage pattern used everywhere else in this migration and will become a real bottleneck once tasks accumulate across many workflow steps.

Suggested fix:
```sql
CREATE INDEX IF NOT EXISTS idx_space_tasks_workflow_step_id ON space_tasks(workflow_step_id);
```

---

### P2 — No FK constraint on `space_tasks.workflow_step_id`

`packages/daemon/src/storage/schema/migrations.ts`, line 1595

`workflow_run_id` has a declared FK:
```sql
FOREIGN KEY (workflow_run_id) REFERENCES space_workflow_runs(id) ON DELETE SET NULL
```
But `workflow_step_id` has no FK at all. If a workflow step is deleted, any task that references that step keeps a stale `workflow_step_id` value pointing to a row that no longer exists. Since `space_workflow_steps` cascades on workflow deletion, this silently leaves dangling references.

A matching declaration would be:
```sql
FOREIGN KEY (workflow_step_id) REFERENCES space_workflow_steps(id) ON DELETE SET NULL
```
This is consistent with the semantic of both fields (the step that spawned the task) and matches the `workflow_run_id` treatment.

---

### P2 — Redundant index on `spaces.workspace_path`

`packages/daemon/src/storage/schema/migrations.ts`, line 1482

`workspace_path` is declared `UNIQUE`, which causes SQLite to automatically create a B-tree index for that constraint (`sqlite_autoindex_spaces_1` or similar). The explicit:
```sql
CREATE INDEX IF NOT EXISTS idx_spaces_workspace_path ON spaces(workspace_path)
```
is a duplicate index occupying extra pages in the database file and adding overhead on every write. It can be dropped without any behavioral change.

---

### P3 — Test for `workflow_step_id SET NULL` behavior is absent

If `workflow_step_id` gets a FK in a follow-up, the test file currently only covers the `workflow_run_id SET NULL` path (test at line 290). A symmetric test for step deletion would complete coverage.

---

- FK ordering is correct: `spaces → space_agents`, `spaces → space_workflows → space_workflow_steps`, `spaces + space_workflows → space_workflow_runs → space_tasks`, `spaces → space_session_groups → space_session_group_members`. No table is created before its FK parent.
- CASCADE semantics are correct throughout. The chained behavior (delete workflow → cascades to space_workflow_runs → SET NULL on space_tasks.workflow_run_id) works as expected and is verified by the test.
- All CREATE TABLE/INDEX statements use IF NOT EXISTS — the migration is fully idempotent.
- The `space_workflow_runs.status` CHECK constraint exactly matches the `WorkflowRunStatus` union in `packages/shared/src/types/space.ts`.
- The `space_tasks` schema (status, priority, task_type, assigned_agent CHECK constraints) exactly matches the `SpaceTask*` types in shared.
- The migration is correctly appended as step 29 and registered in `runMigrations()`.
- No `space_goals` table was mistakenly included — compliance with the task spec.
- The test for the cascade chain (deleting a space deletes all child rows through multiple levels) at line 210 is thorough and realistic.

lsm (commented):

**Recommendation: APPROVE**

Re-review after the fix commit (`eae889e`). All five issues raised in the previous round have been properly addressed:

1. **Redundant `idx_spaces_workspace_path` removed** — The explicit index is gone, replaced by a clear comment at line 1482-1483 explaining the UNIQUE constraint provides an implicit index.

2. **FK on `space_tasks.workflow_step_id` added** — Line 1616 now declares `FOREIGN KEY (workflow_step_id) REFERENCES space_workflow_steps(id) ON DELETE SET NULL`, matching the `workflow_run_id` treatment.

3. **`idx_space_tasks_workflow_step_id` index added** — Lines 1627-1629, consistent with the indexing pattern for all other FK columns.

4. **`UNIQUE(group_id, session_id)` on `space_session_group_members`** — Line 1662, preventing duplicate membership records.

5. **Two new tests added** (17 total, all passing):
   - Line 451: Verifies `workflow_step_id` is SET NULL when the referenced step is deleted.
   - Line 489: Verifies the UNIQUE constraint rejects duplicate `(group_id, session_id)` pairs.

The test index list at line 196 correctly includes the new `idx_space_tasks_workflow_step_id` and the comment at lines 184-185 documents why the redundant workspace_path index is absent.

No remaining issues. The migration is correct, complete, idempotent, and well-tested.

lsm (commented):

**Follow-up review — all previous P1/P2 issues confirmed resolved.**

---

### Verification of previously raised issues

**1. Redundant `idx_spaces_workspace_path` index — FIXED**

The explicit `CREATE INDEX IF NOT EXISTS idx_spaces_workspace_path ON spaces(workspace_path)` call is gone. The comment at line 1482 correctly explains that the `UNIQUE` constraint on `workspace_path` already provides an implicit index. No redundant write overhead remains.

**2. FK constraint on `space_tasks.workflow_step_id` — FIXED**

Line 1616 now declares:
```sql
FOREIGN KEY (workflow_step_id) REFERENCES space_workflow_steps(id) ON DELETE SET NULL
```
This is symmetric with the `workflow_run_id` treatment. Deleting a workflow step will now null out the referencing tasks rather than leaving dangling foreign key values.

**3. `idx_space_tasks_workflow_step_id` index — FIXED**

Line 1628 adds `CREATE INDEX IF NOT EXISTS idx_space_tasks_workflow_step_id ON space_tasks(workflow_step_id)`, aligning the coverage with `workflow_run_id` and all other FK columns.

**4. `UNIQUE(group_id, session_id)` on `space_session_group_members` — FIXED**

Line 1662 adds the composite unique constraint inline in the `CREATE TABLE` statement. Duplicate membership rows are now rejected at the database level.

**5. New tests — CONFIRMED (17 total)**

Two new tests were added and verified correct:
- `deleting a workflow step sets space_tasks.workflow_step_id to NULL` — tests the SET NULL FK behavior end-to-end for workflow steps.
- `space_session_group_members prevents duplicate (group_id, session_id)` — asserts the UNIQUE constraint rejects a second insert of the same (group_id, session_id) pair.

Both tests follow the same fixture pattern as the existing test suite and correctly cover the new constraints.

---

### No new issues found

The implementation is correct and complete. All 8 Space tables are created in FK-safe order, CASCADE and SET NULL semantics are verified by tests, all constraints enforce the shared type definitions, and the migration is idempotent throughout.

PR #428, opened 2026-03-19
feat: custom agent types, repository, and manager (Task 2.1)

lsm:
- Add SpaceAgent interface, SpaceAgentRole, CreateSpaceAgentParams,
  UpdateSpaceAgentParams to packages/shared/src/types/space.ts
- Add packages/shared/src/types/tools.ts with KNOWN_TOOLS constant
  as single source of truth for tool names
- Migration 30: add role and provider columns to space_agents table
- SpaceAgentRepository: full CRUD, getAgentsByIds batch lookup,
  isAgentReferenced deletion protection, JSON serialization for tools/config
- SpaceAgentManager: validates name uniqueness within space, tools from
  KNOWN_TOOLS, model recognition (advisory), deletion blocked when agent
  referenced by workflow steps
- 39 unit tests covering repository and manager


lsm (commented):

> **Recommendation:** REQUEST_CHANGES (own PR, posting as COMMENT)

---

Overall this is a solid, well-structured implementation with good test coverage. The repository pattern, result type, and migration approach are all consistent with the existing codebase conventions. A few issues need addressing before merge.

## P1 Issues (Should Fix)

### 1. Model validation bypasses the established provider-aware API

**File:** `packages/daemon/src/lib/space/managers/space-agent-manager.ts`, lines 140-146

The `validateModel` private method calls `getAvailableModels('global')` and then manually does `m.id === model || m.alias === model` without any provider filtering:

```ts
private validateModel(model: string): string | null {
    const available = getAvailableModels('global');
    if (available.length === 0) return null;
    const found = available.find((m) => m.id === model || m.alias === model);
    return found ? null : `Unrecognized model: "${model}"`;
}
```

The rest of the codebase (see `model-switch-handler.ts` lines 113-114, `session-handlers.ts` lines 367-369) uses the provider-aware `isValidModel(model, cacheKey, providerId)` API, and the `model-service.ts` file explicitly warns against unfiltered lookup:

> _"Use this ONLY for callers that genuinely lack provider context... For all new code, prefer `getModelInfo` with an explicit `providerId`"_

Since `SpaceAgent` already has a `provider` field, when a provider is supplied on create/update, model validation should filter to that provider. Without this, a model that exists only in provider A could be "validated" successfully even when the agent's provider is B.

The method signature is synchronous but the proper API (`isValidModel`, `getModelInfo`) is async. This requires either making the manager's create/update methods async, or accepting that model validation remains intentionally advisory/unfiltered (in which case the inconsistency should be documented explicitly).

### 2. `validateNameUnique` loads all agents for O(N) uniqueness check

**File:** `packages/daemon/src/lib/space/managers/space-agent-manager.ts`, lines 121-127

`validateNameUnique` calls `getBySpaceId` to fetch every agent in the space and then does an in-memory case-insensitive scan. A dedicated `SELECT 1 ... WHERE LOWER(name) = LOWER(?) AND space_id = ? AND id != ?` query in the repository would be cheaper and would push uniqueness enforcement closer to where data lives. At this stage the performance impact is minor, but the pattern is worth establishing correctly given the table will likely grow.

### 3. Neither the repository nor the manager is wired into the daemon application

**Files:** `packages/daemon/src/storage/index.ts`, `packages/daemon/src/app.ts`

`SpaceAgentRepository` is not exported from `storage/index.ts` and `SpaceAgentManager` is not instantiated in `app.ts`. Both exist only as standalone files. This means the implementation cannot be called by any RPC handler today. For a Task 2.1 deliverable that explicitly says _"build data access + business logic layers"_, not connecting them to the application context is an omission. This may be intentional (deferred to RPC handler wiring in a later task), but if so it should be stated in the PR description to avoid confusion.

## P2 Issues (Meaningful Improvement)

### 4. Migration 30 CHECK constraint is silently dropped on existing databases

**File:** `packages/daemon/src/storage/schema/migrations.ts`, lines 1681-1697

`ALTER TABLE ... ADD COLUMN ... CHECK(...)` is a SQLite limitation — SQLite ignores table-level CHECK constraints added via `ALTER TABLE ADD COLUMN` in versions prior to 3.37.0 (2021). The migration does not guard against or document this. The test schema (created fresh in `:memory:`) does enforce the constraint correctly, but the migration path for existing databases does not. Existing databases with the `role` column already added (if any) also silently skip the column-level CHECK.

The correct approach here is either to document this explicitly, or recreate the table with proper constraints (matching how Migration 29 handles full table creation).

### 5. Missing test for `update` with `model: null` (clear model)

**File:** `packages/daemon/tests/unit/lib/space-agent-manager.test.ts`

The manager's `update` path has `if (params.model) { ... validateModel ... }` which means `model: null` (clearing the model) skips validation correctly. However, there is no manager-level test that verifies clearing model to null is accepted and persisted. The repository test at line 252-263 covers it at the storage level, but the manager layer is untested for this case. Given the `UpdateSpaceAgentParams.model` is typed as `string | null | undefined`, this is a notable gap.

### 6. `KnownTool` type is exported but never used as a type in the new code

**File:** `packages/shared/src/types/tools.ts`, line 22

`KnownTool` is defined and exported, but neither `SpaceAgent.tools` nor `CreateSpaceAgentParams.tools` use `KnownTool[]` — both use `string[]`. Using `string[]` is intentional for flexibility (the comment says "advisory"), but the exported type then has no callsites and will be flagged as dead export by Knip. Either drop the type export or use it at the type level in the interfaces.

## P3 Issues (Nits)

### 7. Duplicate `createSchema` helpers in both test files

Both `space-agent-repository.test.ts` and `space-agent-manager.test.ts` define identical `createSchema`, `insertSpace`, `insertWorkflow`, and `insertWorkflowStep`/`insertStep` helpers. A shared test fixture file would reduce drift risk as the schema evolves.

### 8. Error message uses em-dash (`—`) but surrounding code uses hyphens

**File:** `packages/daemon/src/lib/space/managers/space-agent-manager.ts`, line 87

The deletion error message uses `—` (em-dash): `Cannot delete agent "${existing.name}" — it is referenced by workflow steps`. This is fine stylistically but inconsistent with other error strings in the file which use plain hyphens. Minor formatting nit.

---

## Summary of Counts

| Severity | Count |
|----------|-------|
| P0 | 0 |
| P1 | 3 |
| P2 | 3 |
| P3 | 2 |

lsm (commented):

**Recommendation: APPROVE**

### Summary

This PR implements Task 2.1 — custom agent types, repository, and manager for the Space system. The implementation is clean, well-structured, and follows existing codebase patterns. All 39 tests pass and typecheck succeeds.

### What was reviewed

- `packages/shared/src/types/space.ts` — SpaceAgent, SpaceAgentRole, CreateSpaceAgentParams, UpdateSpaceAgentParams types
- `packages/shared/src/types/tools.ts` — KNOWN_TOOLS constant and KnownTool type
- `packages/shared/src/mod.ts` — Re-exports for new types
- `packages/daemon/src/storage/schema/migrations.ts` — Migration 30 (role + provider columns)
- `packages/daemon/src/storage/repositories/space-agent-repository.ts` — Full CRUD + batch lookup + reference check
- `packages/daemon/src/lib/space/managers/space-agent-manager.ts` — Validation layer (name uniqueness, tools, models, deletion protection)
- `packages/daemon/tests/unit/storage/space-agent-repository.test.ts` — 21 tests
- `packages/daemon/tests/unit/lib/space-agent-manager.test.ts` — 18 tests

### Findings

**No blocking issues found.**

**P3 (nits):**

1. **`getAgentsByIds` uses string interpolation for SQL placeholders** (`space-agent-repository.ts:74-76`). This is safe because the placeholders are generated from `ids.map(() => '?')` (not user input) and values are passed as bound parameters. However, this is worth noting as a pattern to be aware of — the `ids` array values themselves are properly parameterized.

2. **`validateNameUnique` fetches all agents for the space** (`space-agent-manager.ts:122-123`) rather than doing a targeted SQL query with a WHERE clause on name. For spaces with many agents this could be slightly inefficient, but it's fine for the expected scale and keeps the manager decoupled from SQL.

3. **Test schema duplication** — Both test files duplicate the full schema setup inline rather than sharing a helper. This is a minor DRY concern but acceptable for test isolation and readability.

### Correctness verification

- Migration 30 is idempotent (try/catch pattern matching existing migration style)
- Migration 29 creates `space_agents` without `role`/`provider`; migration 30 adds them — correct ordering
- Repository handles JSON serialization/deserialization for `tools` and `config` fields correctly
- `rowToAgent` correctly maps `null` DB values to `undefined` for optional fields (model, provider, config)
- Manager validates tools against KNOWN_TOOLS, model against available models cache (with graceful skip when cache is empty)
- Name uniqueness is case-insensitive and correctly handles the rename-to-same-name case via `excludeId`
- Deletion protection correctly checks workflow step references via JOIN
- `UpdateSpaceAgentParams` correctly supports `null` for clearing optional fields (model, provider, config)
- New types are properly exported through `packages/shared/src/mod.ts`
- SpaceAgentRepository is not yet wired into the Database facade or app.ts — appropriate for a data/logic layer task

All tests pass (39/39), typecheck passes.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review after the worker addressed previous P1/P2 feedback. All requested changes have been properly implemented.

### Changes Verified

1. **Provider-aware async model validation** - `SpaceAgentManager.validateModel()` correctly uses the async `isValidModel(model, 'global', provider)` when a provider is supplied, and falls back to synchronous unfiltered check otherwise. The empty-cache skip logic is a reasonable safeguard for startup timing.

2. **DB-level name uniqueness** - `SpaceAgentRepository.isNameTaken()` uses `LOWER()` SQL for case-insensitive comparison, with an `excludeId` parameter for the update path. Both create and update flows call this method correctly.

3. **Proper exports and wiring** - `SpaceAgentRepository` is exported from `packages/daemon/src/storage/index.ts` (line 52). `SpaceAgentManager` is instantiated in `packages/daemon/src/app.ts` (line 95) and included in `DaemonAppContext` (line 56).

4. **Migration CHECK constraint documentation** - Migration 30 (`packages/daemon/src/storage/schema/migrations.ts`, lines 1681-1688) contains a clear comment explaining the SQLite < 3.37.0 limitation where CHECK constraints added via ALTER TABLE are recorded but not enforced.

5. **model:null update test** - Test at line 220 in `space-agent-manager.test.ts` verifies that `update(id, { model: null })` succeeds without triggering validation.

6. **KnownTool type export dropped** - `packages/shared/src/types/tools.ts` exports only the `KNOWN_TOOLS` const array, no unused type.

7. **Shared test fixture** - `packages/daemon/tests/unit/helpers/space-agent-schema.ts` extracts `createSpaceAgentSchema`, `insertSpace`, `insertWorkflow`, and `insertWorkflowStep` helpers, used by both repository and manager test files.

8. **Em-dash replaced** - Error message uses hyphen: `"Cannot delete agent ... - it is referenced by workflow steps"` (line 99 of space-agent-manager.ts).

### Minor Observations (P3)

- **`rowToAgent` maps NULL to undefined for model/provider** (lines 181-182 of space-agent-repository.ts) using `?? undefined`. This means `SpaceAgent` objects will have `model: undefined` rather than `model: null` even when the DB stores NULL. This is consistent with the `SpaceAgent` interface (which uses optional `model?: string` not `model: string | null`), so it works correctly, but callers should be aware of this mapping.

- **`validateModel` provider parameter type** is `string | null` on the manager's private method but the upstream `isValidModel` requires a plain `string`. The `if (provider)` guard on line 156 ensures null/undefined never reaches `isValidModel`, so this is safe.

All 47 tests pass across both test files. The implementation is clean, complete, and addresses all prior review feedback.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (one new issue discovered, one pre-existing gap now visible)

---

## Second-pass summary

Most of the previous feedback has been well addressed. The provider-aware async validation, DB-level name uniqueness, export wiring, CHECK constraint documentation, model-null test, and test fixture extraction are all correctly implemented. The em-dash replacement is confirmed.

One new bug and one remaining gap are worth fixing before merge.

---

## P1 - Unfiltered model validation path misses legacy model IDs

**File:** `packages/daemon/src/lib/space/managers/space-agent-manager.ts`, line 163

The no-provider fallback in `validateModel` uses a direct `available.find()`:

```ts
const found = available.find((m) => m.id === model || m.alias === model);
```

But `findInModels` (the internal function used by `isValidModel` and `getModelInfoUnfiltered`) has a third resolution step: **legacy model mappings**. This means that if a user creates an agent with a legacy full model ID such as `claude-3-5-sonnet-20241022` (mapped to `sonnet` via `LEGACY_MODEL_MAPPINGS` in model-service.ts) and no provider, the manager will incorrectly return `Unrecognized model: "claude-3-5-sonnet-20241022"` even though the provider-aware path (`isValidModel`) would accept it.

The two validation paths are inconsistent. The fix is to replace the manual `available.find()` call with the existing `getModelInfoUnfiltered(model, 'global')` async function (already imported from the same module), which goes through `findInModels` and includes the legacy mapping step.

This gap also exists in the test suite — there is no test for a legacy model ID on the unfiltered path.

---

## P2 - Provider-nulling + model update: provider resolution comment is misleading

**File:** `packages/daemon/src/lib/space/managers/space-agent-manager.ts`, lines 76-78

```ts
const provider =
    params.provider !== undefined ? (params.provider ?? undefined) : existing.provider;
```

When `params.provider = null` (caller is clearing the provider), `params.provider ?? undefined` evaluates to `undefined`, so `validateModel(model, undefined)` runs the unfiltered path. That is probably the right behavior (validate against all providers), but the code comment on line 75 says "Use updated provider if provided; otherwise fall back to existing agent's provider" — that comment describes only the `undefined` case (no provider key sent at all). When `null` is sent the comment is silent about what happens. The expression is not wrong but it is confusing enough that a future reader could misread it. A brief inline note like `// null = clearing provider: validate unfiltered` would clarify the intent.

---

## Previously reported items - status

| Issue | Status |
|---|---|
| P1.1 Provider-aware async model validation | Addressed - `isValidModel(model, 'global', provider)` correctly wired |
| P1.2 DB-level name uniqueness via LOWER() | Addressed - `isNameTaken` implemented in repository, called synchronously in manager |
| P1.3 SpaceAgentRepository exported / manager in app.ts | Addressed - export at `storage/index.ts` line 52, instantiated and returned in `createDaemonApp` |
| P2.4 CHECK constraint limitation | Addressed - comment at migration 30 lines 1681-1689 explains SQLite < 3.37.0 behavior |
| P2.5 model:null update test | Addressed - test at manager test file line 220 |
| P2.6 KnownTool unused export | Addressed - removed from tools.ts |
| P3.7 Shared test fixture | Addressed - `tests/unit/helpers/space-agent-schema.ts` |
| P3.8 Em-dash in error message | Addressed - hyphen used in deletion error |

---

## Minor observation (no action required)

The `getModelInfoUnfiltered` function is marked with a `WARNING` comment about non-deterministic alias resolution when the same alias appears in multiple providers. That same ambiguity applies to the current manual `available.find()` call. Switching to `getModelInfoUnfiltered` would not make things worse — it would make things consistent.

lsm (commented):

**Final re-review (4th pass) -- All previously raised issues are resolved.**

### P1 Fix: Legacy model ID validation (RESOLVED)

The unfiltered model validation path in `validateModel()` now correctly uses `getModelInfoUnfiltered(model, 'global')` instead of the manual `available.find()`. This ensures legacy model IDs like `claude-3-5-sonnet-20241022` are resolved through `findInModels()`, which includes the three-step lookup: exact ID match, alias match, and `LEGACY_MODEL_MAPPINGS` lookup.

The new test at line 139 of `space-agent-manager.test.ts` directly verifies this: it populates the cache with only `sonnet` and asserts that `claude-3-5-sonnet-20241022` (a legacy full ID) is accepted.

### P2 Fix: Provider null vs undefined comment (RESOLVED)

Lines 74-78 of `space-agent-manager.ts` now have a clear inline comment explaining the three-way distinction:
- `params.provider` is a string -- use that provider (scoped validation)
- `params.provider` is null -- caller is clearing the provider; validate unfiltered
- `params.provider` is undefined -- not being changed; use existing agent's provider

### Summary

The implementation is complete and correct. The repository layer, manager with business-logic validation, shared types, migration, and test coverage are all solid. No remaining issues.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

### Summary

Both P1 and P2 issues from the previous review pass are fully resolved. The implementation is correct and ready to merge.

---

### P1 Fix: Legacy model ID resolution

The replacement of the manual `available.find((m) => m.id === model || m.alias === model)` with `getModelInfoUnfiltered(model, 'global')` is correct.

`getModelInfoUnfiltered` delegates to `findInModels`, which runs three resolution steps: exact ID match, alias match, and then a lookup through `LEGACY_MODEL_MAPPINGS`. The mapping entry `'claude-3-5-sonnet-20241022': 'sonnet'` is present at line 30 of `model-service.ts`, so the legacy full ID path is now correctly covered.

The new test at line 139 in `space-agent-manager.test.ts` exercises this path precisely: the cache contains a model with `id: 'sonnet'`, the input is `'claude-3-5-sonnet-20241022'`, and `findInModels` will reach step 3, resolve the legacy mapping to `'sonnet'`, and find the entry. The test correctly validates the fix.

### P2 Fix: Provider resolution comment

The three-case comment added at lines 76-78 of `space-agent-manager.ts` is accurate and matches the implementation:

- `params.provider !== undefined` with a string value → uses that string (scoped validation)
- `params.provider === null` → `params.provider !== undefined` is `true`, then `null ?? undefined` yields `undefined` → unfiltered path
- `params.provider === undefined` → falls to `existing.provider` → uses the stored provider

The null-coalescing expression `params.provider ?? undefined` at line 81 correctly converts `null` to `undefined` before passing to `validateModel`, which checks `if (provider)` to decide between filtered and unfiltered validation. The logic is sound.

### No new issues found

No regressions, no missing edge cases, no concerns introduced by this iteration.

lsm (commented):

### Fourth-pass review — type conflict resolution and alignment with dev-canonical types

**Overall assessment: REQUEST_CHANGES** (1 P0, 1 P2)

The duplicate type declarations have been properly resolved. `SpaceAgentRole` was removed and all code now uses `BuiltinAgentRole` from the single definition in `packages/shared/src/types/space.ts`. The `space.ts` file is exported exactly once from `mod.ts`. No conflicting declarations remain. The migration 30 correctly uses `('planner', 'coder', 'general')` roles with default `'coder'`. Tests use the correct role values and `toolConfig` field throughout.

---

### P0: Repository inserts NULL into NOT NULL columns — will crash on production databases

**Files:**
- `packages/daemon/src/storage/repositories/space-agent-repository.ts` lines 37, 41 (create) and lines 116, 128 (update)
- `packages/daemon/src/storage/schema/migrations.ts` lines 1496, 1499 (migration 29 schema)

Migration 29 defines `space_agents` with:
```sql
description TEXT NOT NULL DEFAULT '',
system_prompt TEXT NOT NULL DEFAULT '',
```

But the repository explicitly inserts `null` for both columns:
```ts
params.description ?? null,   // line 37
params.systemPrompt ?? null,  // line 41
```

SQLite's `DEFAULT` only applies when a column is **omitted** from the INSERT statement. When `null` is explicitly provided for a `NOT NULL` column, SQLite raises a constraint violation error. The same issue exists in `update()` at lines 116 and 128 where `description` and `systemPrompt` can be set to `null`.

The tests pass because the test helper schema (`packages/daemon/tests/unit/helpers/space-agent-schema.ts`) defines these columns as nullable (`description TEXT`, `system_prompt TEXT`), which does not match the production migration schema.

**Fix options (pick one):**
1. Change migration 29 to make `description` and `system_prompt` nullable (drop the `NOT NULL`), which matches the `SpaceAgent` type where both are `description?: string` and `systemPrompt?: string`.
2. Change the repository to use `params.description ?? ''` and `params.systemPrompt ?? ''` instead of `null`, and update the `rowToAgent` mapping accordingly.

Option 1 is recommended since it aligns with the TypeScript types (`description?: string`, `systemPrompt?: string` — both optional/undefined, not empty-string).

---

### P2: Test helper schema diverges from production migration schema

**File:** `packages/daemon/tests/unit/helpers/space-agent-schema.ts` lines 35-39

The test schema uses nullable `description TEXT` and `system_prompt TEXT`, while migration 29 uses `NOT NULL DEFAULT ''`. This divergence means bugs like the P0 above are invisible in tests. The test schema should mirror the production migration to catch constraint violations during testing.

---

### What looks good

- **No duplicate types:** `SpaceAgent`, `CreateSpaceAgentParams`, `UpdateSpaceAgentParams`, and `BuiltinAgentRole` are each defined exactly once in `packages/shared/src/types/space.ts`.
- **`SpaceAgentRole` removed:** All code now uses `BuiltinAgentRole` directly.
- **Migration 30 roles correct:** CHECK constraint uses `('planner', 'coder', 'general')` with default `'coder'`.
- **Repository column mapping:** `toolConfig` correctly maps to `config` column, `systemPrompt` to `system_prompt`.
- **Manager validation:** Model validation with provider-aware path and unfiltered fallback is well-structured. The empty-cache skip is a pragmatic choice.
- **Tests are comprehensive:** Cover all CRUD paths, name uniqueness, batch lookup, deletion protection, and model validation scenarios.
- **Integration in app.ts:** `SpaceAgentManager` is properly wired into `DaemonAppContext`.

lsm (commented):

The fourth-pass rebase fixes all the type-conflict issues correctly — types, repo, manager, and tests are self-consistent and all pass cleanly (`bun run check` green, both test files green). However, a schema mismatch between Migration 29 and the test helper masks a P0 crash bug in production.

---

## P0 — NOT NULL constraint crash in production

**File:** `packages/daemon/src/storage/repositories/space-agent-repository.ts` (lines 37, 43)  
**Root cause:** `packages/daemon/tests/unit/helpers/space-agent-schema.ts` (lines 34–35)

Migration 29 defines both `description` and `system_prompt` as `TEXT NOT NULL DEFAULT ''`:

```sql
-- packages/daemon/src/storage/schema/migrations.ts, lines 1496 / 1499
description TEXT NOT NULL DEFAULT '',
system_prompt TEXT NOT NULL DEFAULT '',
```

The repository `create()` always passes `null` for these when the caller omits them:

```ts
params.description ?? null,  // line 37
params.systemPrompt ?? null, // line 43 (effectively)
```

SQLite enforces NOT NULL at runtime — inserting `null` into a `NOT NULL` column throws  
`NOT NULL constraint failed: space_agents.description`.

The test helper defines these columns as plain `TEXT` (nullable), so every unit test passes while the production schema rejects the same INSERT. Any call to `SpaceAgentManager.create()` without an explicit `description` or `systemPrompt` will crash in a live database.

**Fix options:**
1. Insert `params.description ?? ''` and `params.systemPrompt ?? ''` in the repository, then map `''` back to `undefined` in `rowToAgent` (match the Mig29 schema as-is).
2. Add a Migration 31 to `ALTER COLUMN` those two columns to `TEXT` (nullable), and update `rowToAgent` to map null → undefined as already written.
3. Align the test helper to mirror the actual Mig29 schema so the discrepancy is caught by tests.

Option 1 is the minimal-change path. Option 2 is cleaner type-layer alignment. Option 3 should happen regardless.

---

## P2 — `rowToAgent` maps empty-string to non-undefined for old rows

If the Mig29 schema is kept as-is (NOT NULL DEFAULT ''), then existing rows will have `description = ''` and `system_prompt = ''`. The current `rowToAgent` code does:

```ts
description: (row.description as string | null) ?? undefined,
systemPrompt: (row.system_prompt as string | null) ?? undefined,
```

The `??` nullish-coalescing operator does not coerce `''` to `undefined`. So agents created with empty string stored in the column will surface as `description: ''` instead of `description: undefined`, inconsistent with the `SpaceAgent` type (`description?: string`). Minor UX/API inconsistency, but worth addressing alongside the P0 fix.

---

## P2 — Test schema helper diverges from production Mig29 schema (masked the P0)

`packages/daemon/tests/unit/helpers/space-agent-schema.ts` (lines 34–35) defines `description TEXT` and `system_prompt TEXT` (nullable), while Mig29 uses `NOT NULL DEFAULT ''`. This divergence is what masked the P0. The helper should mirror the Mig29 schema exactly so that constraint violations are caught by unit tests.

---

## P3 — `KNOWN_TOOLS` is exported but has no consumer yet

`packages/shared/src/types/tools.ts` exports `KNOWN_TOOLS` but it is not consumed anywhere in the codebase. Knip does not flag it (likely because it is a shared package export). This is intentional forward-work plumbing per the plan, but it is dead code until Task 2.2/2.3. No action required; noting for visibility.

---

## What passes

- No type conflicts; TypeScript build is clean.
- `BuiltinAgentRole` type is correctly unified; no duplicate definitions.
- Repository field mapping (`toolConfig ↔ config`, `systemPrompt ↔ system_prompt`, `role`) is correct.
- Migration 30 idempotency pattern (probe-then-alter) is correct.
- Manager model/provider validation logic is sound and well-tested.
- Deletion protection via `isAgentReferenced` is correct and tested.
- Name uniqueness (case-insensitive, scoped to space, excludeId for updates) is correct.

lsm (commented):

### Fifth-pass review — P0 NOT NULL constraint fix verification

All four P0 fixes are correctly implemented and consistent with the production schema:

**1. `create()` (space-agent-repository.ts:37,41)**
- `params.description ?? ''` and `params.systemPrompt ?? ''` — correctly avoids passing null/undefined to NOT NULL columns. Matches Migration 29 schema (`description TEXT NOT NULL DEFAULT ''`, `system_prompt TEXT NOT NULL DEFAULT ''`).

**2. `update()` (space-agent-repository.ts:116,128)**
- Same pattern: `params.description ?? ''` and `params.systemPrompt ?? ''`. When `UpdateSpaceAgentParams` sends `null` (meaning "clear the field"), this correctly stores `''` instead of violating the NOT NULL constraint.

**3. `rowToAgent()` (space-agent-repository.ts:180,183)**
- Changed from `?? undefined` to `|| undefined`. This correctly maps stored `''` back to `undefined`, which is consistent with the optional `description?: string` and `systemPrompt?: string` types in `SpaceAgent`. The `||` operator treats both `''` and `null` as falsy, providing a clean round-trip: `undefined` (caller) -> `''` (DB) -> `undefined` (read back).

**4. Test schema helper (space-agent-schema.ts:35,39)**
- `description TEXT NOT NULL DEFAULT ''` and `system_prompt TEXT NOT NULL DEFAULT ''` now mirrors the exact Migration 29 production schema. Future NOT NULL constraint bugs will be caught by the in-memory test database.

### Previous P2 issues

The previously flagged P2 issues (SQL injection in `getAgentsByIds` placeholder construction and `setModelsCache` test-only export) remain as noted in prior reviews. These are acceptable given the context — the placeholders are `?`-only (no user input interpolated into SQL), and `setModelsCache` is a pragmatic testing seam.

### Verdict

**Recommendation: APPROVE** — The P0 crash bug is fully resolved. The fix is minimal, correct, and well-tested. No new issues introduced.

lsm (commented):

**Recommendation: APPROVE**

This is the fifth-pass review. The P0 crash bug (NOT NULL constraint violation) and P2 concerns from previous passes are verified resolved. The implementation is correct and clean.

---

### P0 (blocking) — 0 issues

All previously identified P0 issues are resolved:

- `create()`: `params.description ?? ''` and `params.systemPrompt ?? ''` correctly prevent NULL from reaching NOT NULL columns.
- `update()`: `params.description ?? ''` and `params.systemPrompt ?? ''` correctly store `''` when a caller explicitly passes `null` (meaning "clear").
- `rowToAgent()`: `(row.description as string) || undefined` and `(row.system_prompt as string) || undefined` correctly map the stored `''` back to `undefined`, consistent with `SpaceAgent.description?: string`.
- Test schema: `description TEXT NOT NULL DEFAULT ''` and `system_prompt TEXT NOT NULL DEFAULT ''` now mirror the Mig29 production schema, so constraint bugs will be caught by tests.

---

### P1 (should-fix) — 0 issues

No P1 issues found.

---

### P2 (important suggestion) — 1 remaining gap

**Missing test: update→clear→roundtrip for `description` and `systemPrompt`**

The code fix for `description: null` and `systemPrompt: null` (store as `''`, read back as `undefined`) is in place and correct. However, no test exercises the exact roundtrip:

1. Create agent with description/systemPrompt set.
2. `update(id, { description: null, systemPrompt: null })`
3. Assert returned agent has `description === undefined` and `systemPrompt === undefined`.

The equivalent roundtrip test exists for `model` and `provider` ("sets model and provider to null", line 181 in `space-agent-repository.test.ts`) and for `toolConfig` (line 195), but not for description/systemPrompt. Since the P0 fix touches exactly this path, a regression test here would lock in the invariant.

`packages/daemon/tests/unit/storage/space-agent-repository.test.ts`, `describe('update')` block.

---

### P3 (nit) — 1 observation

**PR description says "validates tools from KNOWN_TOOLS" — this is not implemented**

`KNOWN_TOOLS` is defined in `packages/shared/src/types/tools.ts` and exported from `packages/shared/src/mod.ts`, but it is not imported or used anywhere — specifically not in `SpaceAgentManager`. The manager performs no validation of tool names inside `toolConfig`. The PR description's claim about "tools from KNOWN_TOOLS" validation is inaccurate for this iteration. This is fine if the validation is deferred to a future task, but the PR description should not overstate what was implemented.

---

### Summary

The P0 NOT NULL constraint crash and the test schema mismatch are fully resolved. The NULL→empty-string coercion and empty-string→undefined roundtrip logic is correct in both `create` and `update` paths. The one remaining gap is a missing regression test for the `description: null` / `systemPrompt: null` clear-and-read-back path, which is a P2. Approving — the core correctness of this PR is sound.

lsm (commented):

**Recommendation: APPROVE**

### P2 Concern — Regression Test for description/systemPrompt Clear-and-Roundtrip

The requested regression test is fully addressed. Commit `ed641671` adds a single 19-line test case to `/packages/daemon/tests/unit/storage/space-agent-repository.test.ts` (lines 195–212) that precisely locks in the full coercion chain:

1. Creates an agent with non-empty `description: 'Some description'` and `systemPrompt: 'Some prompt'`
2. Calls `repo.update(agent.id, { description: null, systemPrompt: null })`
3. Asserts the returned agent has `description === undefined` and `systemPrompt === undefined`

The test comment explicitly names both steps of the chain — `update` stores `''` (avoiding the NOT NULL constraint), and `rowToAgent` maps `''` back to `undefined` — which matches the fix in the P0 commit (`0d4ddd6f`).

All 26 tests in the repository test suite pass (`bun test packages/daemon/tests/unit/storage/space-agent-repository.test.ts`: 26 pass, 0 fail).

This closes the only remaining concern from the previous review pass. The implementation is complete and the regression guard is permanent.

PR #429, opened 2026-03-19
feat: add Space repositories and managers (Task 1.3)

lsm:
Introduces the data access and business logic layers for the Space system:

Repositories (packages/daemon/src/storage/repositories/):
- SpaceRepository: full CRUD with JSON serialization for allowedModels,
  sessionIds, config; getSpaceByPath, addSessionToSpace, removeSessionFromSpace
- SpaceTaskRepository: full CRUD with batch ops (listByWorkflowRun,
  listBySpace, listByStatus); handles customAgentId, workflowRunId,
  workflowStepId, promoteDraftTasksByCreator
- SpaceWorkflowRunRepository: CRUD for workflow runs; getActiveRuns,
  updateStepIndex, updateStatus; sets completedAt on terminal status
- SpaceSessionGroupRepository: CRUD for groups and members; getGroupsByTask,
  getGroupsBySpace, addMember (idempotent), removeMember

Managers (packages/daemon/src/lib/space/managers/):
- SpaceManager: workspace path validation — resolves symlinks via fs.realpath,
  validates directory via fs.stat, enforces uniqueness across active spaces,
  warns (non-fatal) when path is not a git repo, stores resolved real path
- SpaceTaskManager: task lifecycle with VALID_SPACE_TASK_TRANSITIONS, status
  transition validation, dependency checks, cascade cancellation, reviewTask

Barrel export at packages/daemon/src/lib/space/index.ts.
knip.json: ignore space/index.ts (no callers yet — wired in later tasks).

Tests: 108 unit tests across 6 test files covering happy paths and error cases.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -24,6 +24,7 @@
 		"scripts/**",
 		"packages/daemon/src/lib/agent/output-limiter-hook.ts",
 		"packages/daemon/src/lib/lobby/index.ts",
+		"packages/daemon/src/lib/space/index.ts",
 		"packages/daemon/src/lib/prompts/index.ts",
 		"packages/shared/src/event-bus.ts",
 		"packages/daemon/src/lib/room/task-message-queue.ts",
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -1,0 +1,10 @@
+/**
+ * Space module — managers for the Space multi-agent workflow system.
+ */
+
+export { SpaceManager } from './managers/space-manager';
+export {
+	SpaceTaskManager,
+	VALID_SPACE_TASK_TRANSITIONS,
+	isValidSpaceTaskTransition,
+} from './managers/space-task-manager';
--- a/packages/daemon/src/lib/space/managers/space-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-manager.ts
@@ -1,0 +1,176 @@
+/**
+ * SpaceManager - Space management with workspace path validation
+ *
+ * Handles:
+ * - Creating spaces with workspace path validation (symlink resolution, existence check, uniqueness)
+ * - Listing, updating, archiving, and deleting spaces
+ * - Session association
+ */
+
+import { promises as fs } from 'node:fs';
+import { exec } from 'node:child_process';
+import { promisify } from 'node:util';
+import type { Database as BunDatabase } from 'bun:sqlite';
+import { SpaceRepository } from '../../../storage/repositories/space-repository';
+import type { Space, CreateSpaceParams, UpdateSpaceParams } from '@neokai/shared';
+
+const execAsync = promisify(exec);
+
+export class SpaceManager {
+	private spaceRepo: SpaceRepository;
+
+	constructor(private db: BunDatabase) {
+		this.spaceRepo = new SpaceRepository(db);
+	}
+
+	/**
+	 * Create a new space.
+	 * Validates the workspace path: resolves symlinks, checks existence, ensures uniqueness.
+	 * Warns (but does not fail) if the path is not a git repository.
+	 */
+	async createSpace(params: CreateSpaceParams): Promise<Space> {
+		const resolvedPath = await this.resolveAndValidatePath(params.workspacePath);
+
+		// Check uniqueness across active spaces
+		const existing = this.spaceRepo.getSpaceByPath(resolvedPath);
+		if (existing) {
+			throw new Error(
+				`A space already exists for workspace path: ${resolvedPath} (space id: ${existing.id})`
+			);
+		}
+
+		// Warn if not a git repository (non-fatal)
+		const isGit = await this.isGitRepository(resolvedPath);
+		if (!isGit) {
+			// Warn only — using a logger would require injecting one; warn via stderr for now
+			process.stderr.write(
+				`[SpaceManager] Warning: workspace path is not a git repository: ${resolvedPath}\n`
+			);
+		}
+
+		return this.spaceRepo.createSpace({ ...params, workspacePath: resolvedPath });
+	}
+
+	/**
+	 * Get a space by ID
+	 */
+	async getSpace(id: string): Promise<Space | null> {
+		return this.spaceRepo.getSpace(id);
+	}
+
+	/**
+	 * List spaces
+	 */
+	async listSpaces(includeArchived = false): Promise<Space[]> {
+		return this.spaceRepo.listSpaces(includeArchived);
+	}
+
+	/**
+	 * Update a space
+	 */
+	async updateSpace(id: string, params: UpdateSpaceParams): Promise<Space> {
+		const space = this.spaceRepo.getSpace(id);
+		if (!space) {
+			throw new Error(`Space not found: ${id}`);
+		}
+
+		const updated = this.spaceRepo.updateSpace(id, params);
+		if (!updated) {
+			throw new Error(`Failed to update space: ${id}`);
+		}
+
+		return updated;
+	}
+
+	/**
+	 * Archive a space
+	 */
+	async archiveSpace(id: string): Promise<Space> {
+		const space = this.spaceRepo.getSpace(id);
+		if (!space) {
+			throw new Error(`Space not found: ${id}`);
+		}
+
+		const archived = this.spaceRepo.archiveSpace(id);
+		if (!archived) {
+			throw new Error(`Failed to archive space: ${id}`);
+		}
+
+		return archived;
+	}
+
+	/**
+	 * Delete a space by ID
+	 */
+	async deleteSpace(id: string): Promise<boolean> {
+		const space = this.spaceRepo.getSpace(id);
+		if (!space) {
+			return false;
+		}
+
+		return this.spaceRepo.deleteSpace(id);
+	}
+
+	/**
+	 * Add a session to a space
+	 */
+	async addSession(spaceId: string, sessionId: string): Promise<Space> {
+		const updated = this.spaceRepo.addSessionToSpace(spaceId, sessionId);
+		if (!updated) {
+			throw new Error(`Space not found: ${spaceId}`);
+		}
+		return updated;
+	}
+
+	/**
+	 * Remove a session from a space
+	 */
+	async removeSession(spaceId: string, sessionId: string): Promise<Space> {
+		const updated = this.spaceRepo.removeSessionFromSpace(spaceId, sessionId);
+		if (!updated) {
+			throw new Error(`Space not found: ${spaceId}`);
+		}
+		return updated;
+	}
+
+	/**
+	 * Resolve symlinks and validate the workspace path exists and is a directory.
+	 * Returns the real (resolved) absolute path.
+	 */
+	private async resolveAndValidatePath(workspacePath: string): Promise<string> {
+		// Resolve symlinks to get the canonical real path
+		let realPath: string;
+		try {
+			realPath = await fs.realpath(workspacePath);
+		} catch {
+			throw new Error(`Workspace path does not exist: ${workspacePath}`);
+		}
+
+		// Verify it is accessible and is a directory
+		try {
+			const stat = await fs.stat(realPath);
+			if (!stat.isDirectory()) {
+				throw new Error(`Workspace path is not a directory: ${realPath}`);
+			}
+		} catch (err) {
+			if (err instanceof Error && err.message.includes('not a directory')) {
+				throw err;
+			}
+			throw new Error(`Cannot access workspace path: ${realPath}`);
+		}
+
+		return realPath;
+	}
+
+	/**
+	 * Check if the given path is inside a git repository (non-fatal check)
+	 */
+	private async isGitRepository(dirPath: string): Promise<boolean> {
+		try {
+			await execAsync('git rev-parse --git-dir', { cwd: dirPath });
+			return true;
+		} catch {
+			return false;
+		}
+	}
+}
--- a/packages/daemon/src/lib/space/managers/space-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-manager.ts
@@ -12,9 +12,11 @@ import { exec } from 'node:child_process';
 import { promisify } from 'node:util';
 import type { Database as BunDatabase } from 'bun:sqlite';
 import { SpaceRepository } from '../../../storage/repositories/space-repository';
+import { Logger } from '../../logger';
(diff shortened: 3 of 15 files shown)


lsm (commented):

Overall this is a solid implementation. The repositories follow existing codebase patterns well, JSON serialization/deserialization is handled correctly, status transition validation is thorough, and test coverage is comprehensive. A few issues worth addressing:

---

### P1: Archived spaces permanently block workspace path reuse

**Files:** `packages/daemon/src/storage/repositories/space-repository.ts` (line 61), `packages/daemon/src/lib/space/managers/space-manager.ts` (line 35)

The `workspace_path` column has a `UNIQUE` constraint in the schema (see `space-test-db.ts` line 17), and `getSpaceByPath()` queries without filtering by `status`. This means:

1. If a space is archived, the UNIQUE constraint in the DB prevents inserting a new row with the same `workspace_path`.
2. Even if the UNIQUE constraint were relaxed, `getSpaceByPath()` would find the archived space and SpaceManager.createSpace would reject it with "A space already exists."

A user who archives a space for `/workspace/project` can never create a new space for that path. This is likely unintended -- archiving should free up the path. Either:
- Filter `getSpaceByPath` to `WHERE workspace_path = ? AND status = 'active'`, and change the DB constraint to a partial unique index (`CREATE UNIQUE INDEX ... WHERE status = 'active'`), or
- Delete the row on archive (but that loses history), or
- Add an `unarchive` operation as the intended recovery path and document that workspace paths are permanently claimed.

### P2: SpaceWorkflowRunRepository description defaults to empty string instead of undefined

**File:** `packages/daemon/src/storage/repositories/space-workflow-run-repository.ts` (line 40)

`createRun` stores `params.description ?? ''` (empty string) when description is not provided, but the `SpaceWorkflowRun` type declares `description?: string` (optional). After a round-trip, a run created without a description will have `description: ''` rather than `undefined`. This is a minor type/semantics mismatch. The existing `SpaceRepository` does the same thing (line 32 stores `''`), so this may be a deliberate project convention, but it's worth noting for API consumers who might check `if (run.description)` vs `if (run.description !== undefined)`.

### P2: cancelTaskCascade re-queries all pending tasks on each recursive step

**File:** `packages/daemon/src/lib/space/managers/space-task-manager.ts` (lines 197-209)

`doCancelCascade` calls `listTasksByStatus('pending')` at every recursive level, issuing a new DB query for each cancelled task. For a deep dependency chain this is O(n^2) queries. Consider querying all tasks once at the top level and filtering in memory, or using a single batch update. Not a blocking issue for current scale but worth noting.

### P3: knip.json entry

**File:** `knip.json`

Adding `packages/daemon/src/lib/space/index.ts` to the knip ignore list is appropriate since it is an entry point not yet imported. This should be removed once the module is wired into the daemon app.

### P3: Minor dead-code path in restart clearing logic

**File:** `packages/daemon/src/lib/space/managers/space-task-manager.ts` (line 121)

The condition `newStatus === 'review'` in the restart clearing block is unreachable for `cancelled` status because the transition map does not allow `cancelled -> review`. Not harmful, but slightly misleading about what transitions are actually possible.

---

**Recommendation: APPROVE** -- The P1 is a real design issue but is unlikely to manifest before the Space feature is wired up end-to-end. All other issues are minor. The code is clean, well-structured, and has thorough test coverage including edge cases like symlink resolution, progress clamping, cascade cancellation, and cross-space task isolation.

lsm (commented):

---

The implementation is well-structured and correctly implements the core data access and business logic layers for the Space system. The repository pattern, JSON serialization, and manager abstractions all follow the existing codebase conventions. That said, there are a few issues ranging from a real behavior bug to code-quality concerns that should be addressed before merging.

---

### P1 — process.stderr.write used instead of Logger

**File:** `packages/daemon/src/lib/space/managers/space-manager.ts`, line 46

The git-repository warning uses `process.stderr.write` directly instead of the project's `Logger` class. The comment in the code acknowledges this: "using a logger would require injecting one; warn via stderr for now." However, the Logger class does NOT need to be injected — it is a module-level singleton used across the entire daemon (e.g., `task-handlers.ts`: `const log = new Logger('task-handlers')`).

Using `process.stderr.write` bypasses the project's log-level control, namespace filtering, and test silencing. In unit tests, this warning leaks to stderr even when the test suite suppresses console output. The fix is trivial:

```ts
// At the top of space-manager.ts
import { Logger } from '../../logger';
const log = new Logger('SpaceManager');

// Replace process.stderr.write(...) with:
log.warn(`workspace path is not a git repository: ${resolvedPath}`);
```

---

### P2 — Archived spaces permanently block workspace path reuse (no test documents this)

**Files:** `packages/daemon/src/storage/repositories/space-repository.ts` line 61, `packages/daemon/src/lib/space/managers/space-manager.ts` line 35

`getSpaceByPath` does not filter by status, so it returns archived spaces. Combined with the UNIQUE constraint on `workspace_path` in the DB schema, once a path is used by ANY space (active or archived), it can never be reused. This is likely intentional (the UNIQUE DB constraint enforces it at the storage level), but it is a consequential design decision that is neither documented nor tested.

There is no test case for: "create space A at path X, archive A, try to create space B at path X — expect failure." Without a test, future callers may not realize this restriction exists. Consider adding the test and a code comment on `getSpaceByPath` explaining that archived spaces are included in the lookup.

---

### P2 — updateSpace cannot clear config (type gap)

**File:** `packages/shared/src/types/space.ts` line 84

`UpdateSpaceParams.config` is typed as `Record<string, unknown>` (not `Record<string, unknown> | null`). Once a Space has a `config` value, it cannot be explicitly cleared through the `updateSpace` API — the update silently skips the config field when `undefined` is passed. This asymmetry is not tested and may cause friction for callers. Either document the limitation or add a `null` variant to the type and handle it in the repository.

---

### P2 — started_at is overwritten on re-entry from needs_attention/cancelled (undocumented behavior)

**File:** `packages/daemon/src/storage/repositories/space-task-repository.ts`, lines 120–122

When `updateTask` receives `status: 'in_progress'`, it always sets `started_at = Date.now()`. This means transitioning `needs_attention -> in_progress` or `cancelled -> in_progress` overwrites the original start time. If `startedAt` is meant to track when work first began (useful for duration metrics), this is a latent bug. The existing NeoTask manager has the same behavior, so this may be intentional — but neither implementation guards against overwriting an existing `started_at` and there is no test that documents the chosen behavior for the re-entry case.

---

### P2 — Test gap: no test for error/result clearing on cancelled -> pending restart

**File:** `packages/daemon/tests/unit/lib/space-task-manager.test.ts`

`setTaskStatus` clears `error`, `result`, and `progress` when restarting from `needs_attention` OR `cancelled` (lines 119–126 in space-task-manager.ts). The test suite covers the `needs_attention -> pending` restart path but has no test for `cancelled -> pending` or `cancelled -> in_progress`. Since `cancelled` is listed as a valid restart source in both the code and the `VALID_SPACE_TASK_TRANSITIONS` map, this is a meaningful coverage gap.

---

### P3 — reviewTask duplicates transition-validity check (bypasses setTaskStatus)

**File:** `packages/daemon/src/lib/space/managers/space-task-manager.ts`, lines 220–225

`reviewTask` manually re-checks `isValidSpaceTaskTransition(task.status, 'review')` and then calls `updateTask` directly, bypassing `setTaskStatus`. This duplicates the validation path. Future changes to transition rules applied to `setTaskStatus` will not automatically apply to `reviewTask`. Consider routing through `setTaskStatus` or extracting the guard into a shared assertion helper.

---

### P3 — SpaceWorkflowRunRepository.createRun silently ignores config

**File:** `packages/daemon/src/storage/repositories/space-workflow-run-repository.ts`, lines 35–47

`createRun` always stores `null` for `config` regardless of params. `CreateWorkflowRunParams` does not include a `config` field, so this is consistent with the type — but the omission is silent. If workflow runs ever need initial config, this will require changes in both the type and the repo. A brief TODO comment would communicate the intentional omission clearly.

---

### Overall Assessment

The implementation is correct for its primary scenarios and covers the core happy paths and error cases well. The main actionable items are:
- Replace `process.stderr.write` with the project's Logger (P1, trivial one-line fix)
- Add tests for the archived-path-reuse behavior and the cancelled-restart clearing (P2 test gaps)
- Document or fix the `started_at` overwrite behavior on task re-entry (P2)

lsm (commented):

### Re-review: All previously raised issues have been correctly addressed

**Recommendation: APPROVE**

I verified each of the six fixes from the previous review:

1. **P1 fix: Logger instead of `process.stderr.write`** — `space-manager.ts:16-19` now imports and instantiates `Logger('SpaceManager')`, and line 47 uses `log.warn(...)`. Correct.

2. **P1/P2 fix: Documented archived-space path-blocking + test** — `space-repository.ts:58-64` has a clear NOTE explaining the UNIQUE constraint design decision. `space-manager.test.ts:92-101` adds a test that archives a space and confirms creating a new one at the same path throws `'already exists'`. Correct.

3. **P2 fix: Test for `cancelled -> pending` restart clearing** — `space-task-manager.test.ts:130-141` tests the full lifecycle through `needs_attention -> in_progress -> cancelled -> pending` and asserts `error`, `progress` are cleared. An additional test at lines 143-152 covers `cancelled -> in_progress`. Both correct.

4. **P2 fix: Documented `started_at` overwrite behavior** — `space-task-repository.ts:121-123` adds a comment explaining that `started_at` is always stamped on entry to `in_progress`, including re-entries. Correct.

5. **P3 fix: `reviewTask` routed through `setTaskStatus`** — `space-task-manager.ts:218-241` now calls `this.setTaskStatus(taskId, 'review')` first for validation and field clearing, then applies PR metadata as a separate `updateTask` call. This eliminates duplicated transition logic. Correct.

6. **P3 fix: Removed dead-code `review` case from cancelled restart** — `space-task-manager.ts:121-129` now correctly separates the `needs_attention` and `cancelled` branches. `needs_attention` allows targets `pending | in_progress | review`; `cancelled` allows only `pending | in_progress`. This matches `VALID_SPACE_TASK_TRANSITIONS` at line 25 where `cancelled: ['pending', 'in_progress']`. Correct.

No new issues found. The implementation is clean and well-tested.

lsm (commented):

**Recommendation: APPROVE**

This is a follow-up review verifying the fixes from the previous REQUEST_CHANGES round. All five previously-flagged issues have been addressed correctly. The remaining concern is very minor.

---

**P1 — Logger fix: DONE correctly**

`space-manager.ts` now imports `Logger` from `'../../logger'`, instantiates `const log = new Logger('SpaceManager')` at module scope (line 19), and calls `log.warn(...)` for the non-git-repo warning (line 47). No `process.stderr.write` or `console.*` calls remain in any file under `packages/daemon/src/lib/space/`.

**P2 — Archived-space path-blocking documented and tested: DONE**

`getSpaceByPath` in `space-repository.ts` (lines 57-71) now has a JSDoc comment explicitly documenting that archived spaces are included and that `workspace_path` is a permanent identifier by design. The test at `space-manager.test.ts` lines 92-101 correctly documents the behavior with the title "throws if workspace path is used by an archived space (paths are permanent identifiers)".

**P2 — cancelled -> pending restart clearing tested: DONE (two tests added)**

Two new tests cover the previously-missing paths:
- `cancelled -> pending` (lines 130-141): verifies error and progress are cleared
- `cancelled -> in_progress` (lines 143-152): verifies progress is cleared

Minor gap: the `cancelled -> pending` test does not assert that `result` is `undefined`. The `needs_attention -> pending` test at line 126 checks all three fields (`error`, `result`, `progress`), but the cancelled restart test only checks `error` and `progress`. Since `result` is cleared in the same block (line 127 of `space-task-manager.ts`), this is a coverage cosmetic issue, not a code bug.

**P2 — started_at overwrite documented: DONE**

Lines 121-123 in `space-task-repository.ts` now carry a comment:
> "Always stamp started_at on entry to in_progress, including re-entries from needs_attention or cancelled. This records when the most recent work began, not when the task was originally created. Mirrors TaskRepository behavior."

This clearly communicates the intentional design.

**P3 — reviewTask routes through setTaskStatus: DONE**

`reviewTask` now calls `await this.setTaskStatus(taskId, 'review')` (line 220) for existence check and transition validation, then applies PR metadata (prUrl, prNumber, prCreatedAt) in a second `updateTask` call. This removes the duplicate transition check.

One minor observation: the refactor introduces a two-step commit — the task reaches `review` status on the first call, and PR metadata is applied in a second call. If the second `updateTask` call returns null (task deleted between calls), the code throws correctly at line 237. The window between the two DB writes is extremely narrow and only exploitable if a concurrent delete races with the review transition, so this is acceptable for the current architecture.

**P3 — Dead code `review` case from cancelled restart: DONE**

Lines 121-124 of `space-task-manager.ts` now correctly only check `pending` and `in_progress` for the `cancelled` branch, matching `VALID_SPACE_TASK_TRANSITIONS.cancelled = ['pending', 'in_progress']`.

---

### Remaining Minor Concern (P3)

**`cancelled -> pending` test missing `result` assertion**

`packages/daemon/tests/unit/lib/space-task-manager.test.ts`, lines 130-141

The test verifies `error` and `progress` are cleared but does not assert `result` is `undefined`. The sibling `needs_attention -> pending` test at line 126 is complete. Consider adding `expect(restarted.result).toBeUndefined();` for symmetry.

---

The implementation is correct, well-documented, and all substantive issues from the previous round are resolved.

lsm (commented):

**Recommendation: APPROVE**

### Follow-up review: P3 fix in `space-task-manager.test.ts`

The added assertion is correct.

**What the fix does:** Line 140 adds `expect(restarted.result).toBeUndefined()` to the `cancelled -> pending` test, matching the assertion pattern in the sibling `needs_attention -> pending` test (lines 125-127).

**Why the assertion is valid:**
- `setTaskStatus` sets `updates.result = null` when transitioning out of `cancelled` (line 128 of `space-task-manager.ts`).
- `rowToSpaceTask` in the repository maps `null` DB values via `(row.result as string | null) ?? undefined`, so a `null` write correctly surfaces as `undefined` on the returned `SpaceTask`.
- The assertion therefore tests real behavior and will catch a regression if the field-clearing logic is removed.

**One nuance worth noting (not a bug):** In this specific test's execution path, `result` was never written to begin with (no `completed` transition occurs before cancellation), so the assertion was vacuously true before the fix. It is still a worthwhile addition because it documents intent and would catch a future regression where `cancelled -> pending` failed to clear a previously set `result`. The sibling `needs_attention -> pending` test is an exact match structurally, so the symmetry is appropriate.

No other concerns found. The fix is minimal, correct, and consistent with the existing test style.

PR #431, opened 2026-03-19
feat: add spaceAgent RPC handlers and DaemonEventMap entries (Task 2.2)

lsm:
- Register spaceAgent.created/updated/deleted in DaemonEventMap
- Add space-agent-handlers.ts with full CRUD: create, list, get, update, delete
- Delete returns clear error (including workflow names) when agent is referenced
- Wire handlers into setupRPCHandlers via new spaceAgentManager dep
- Pass spaceAgentManager from app.ts into setupRPCHandlers
- 31 unit tests covering all handlers and error cases

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -217,6 +217,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		db,
 		gitHubService: gitHubService ?? undefined,
 		spaceManager,
+		spaceAgentManager,
 	});
 
 	// Create WebSocket handlers
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -440,6 +440,23 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		run?: Partial<import('@neokai/shared').SpaceWorkflowRun>;
 	};
 
+	// Space Agent events (channel: 'space:${spaceId}')
+	'spaceAgent.created': {
+		sessionId: string;
+		spaceId: string;
+		agent: import('@neokai/shared').SpaceAgent;
+	};
+	'spaceAgent.updated': {
+		sessionId: string;
+		spaceId: string;
+		agent: import('@neokai/shared').SpaceAgent;
+	};
+	'spaceAgent.deleted': {
+		sessionId: string;
+		spaceId: string;
+		agentId: string;
+	};
+
 	// Feature Flag events (PHASE 3: Gradual rollout infrastructure)
 	'featureFlag.updated': {
 		sessionId: string;
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -49,6 +49,8 @@ import type { SpaceManager } from '../space/managers/space-manager';
 import { SpaceTaskManager } from '../space/managers/space-task-manager';
 import { SpaceTaskRepository } from '../../storage/repositories/space-task-repository';
 import { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
+import { setupSpaceAgentHandlers } from './space-agent-handlers';
+import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
 
 export interface RPCHandlerDependencies {
 	messageHub: MessageHub;
@@ -61,6 +63,7 @@ export interface RPCHandlerDependencies {
 	gitHubService?: GitHubService;
 	/** Space manager instance — shared with DaemonAppContext (single source of truth) */
 	spaceManager: SpaceManager;
+	spaceAgentManager: SpaceAgentManager;
 }
 
 const log = new Logger('rpc-handlers');
@@ -189,6 +192,9 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanu
 		spaceTaskManagerFactory,
 		deps.daemonHub
 	);
+
+	// Space agent handlers
+	setupSpaceAgentHandlers(deps.messageHub, deps.daemonHub, deps.spaceAgentManager);
 
 	// Return cleanup function to stop background services
 	return () => {
--- a/packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts
@@ -1,0 +1,155 @@
+/**
+ * Space Agent RPC Handlers
+ *
+ * RPC handlers for Space agent CRUD operations:
+ * - spaceAgent.create  - Create an agent in a Space
+ * - spaceAgent.list    - List all agents in a Space
+ * - spaceAgent.get     - Get a single agent by ID
+ * - spaceAgent.update  - Update an agent's fields
+ * - spaceAgent.delete  - Delete an agent (error if referenced by workflows)
+ */
+
+import type { MessageHub } from '@neokai/shared';
+import type { DaemonHub } from '../daemon-hub';
+import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
+import { Logger } from '../logger';
+
+const log = new Logger('space-agent-handlers');
+
+export function setupSpaceAgentHandlers(
+	messageHub: MessageHub,
+	daemonHub: DaemonHub,
+	spaceAgentManager: SpaceAgentManager
+): void {
+	// spaceAgent.create — create a new agent within a Space
+	messageHub.onRequest('spaceAgent.create', async (data) => {
+		const params = data as {
+			spaceId: string;
+			name: string;
+			role: string;
+			description?: string;
+			model?: string;
+			provider?: string;
+			toolConfig?: Record<string, unknown>;
+			systemPrompt?: string;
+		};
+
+		if (!params.spaceId) throw new Error('spaceId is required');
+		if (!params.name) throw new Error('name is required');
+		if (!params.role) throw new Error('role is required');
+
+		const result = await spaceAgentManager.create({
+			spaceId: params.spaceId,
+			name: params.name,
+			role: params.role as import('@neokai/shared').BuiltinAgentRole,
+			description: params.description,
+			model: params.model,
+			provider: params.provider,
+			toolConfig: params.toolConfig,
+			systemPrompt: params.systemPrompt,
+		});
+
+		if (!result.ok) throw new Error(result.error);
+
+		daemonHub
+			.emit('spaceAgent.created', {
+				sessionId: `space:${result.value.spaceId}`,
+				spaceId: result.value.spaceId,
+				agent: result.value,
+			})
+			.catch((err) => {
+				log.warn('Failed to emit spaceAgent.created:', err);
+			});
+
+		return { agent: result.value };
+	});
+
+	// spaceAgent.list — list all agents for a Space
+	messageHub.onRequest('spaceAgent.list', async (data) => {
+		const params = data as { spaceId: string };
+		if (!params.spaceId) throw new Error('spaceId is required');
+
+		const agents = spaceAgentManager.listBySpaceId(params.spaceId);
+		return { agents };
+	});
+
+	// spaceAgent.get — get a single agent by ID
+	messageHub.onRequest('spaceAgent.get', async (data) => {
+		const params = data as { id: string };
+		if (!params.id) throw new Error('id is required');
+
+		const agent = spaceAgentManager.getById(params.id);
+		if (!agent) throw new Error(`Agent not found: ${params.id}`);
+
+		return { agent };
+	});
+
+	// spaceAgent.update — update an existing agent
+	messageHub.onRequest('spaceAgent.update', async (data) => {
+		const params = data as {
+			id: string;
+			name?: string;
+			description?: string | null;
+			role?: string;
+			model?: string | null;
+			provider?: string | null;
+			toolConfig?: Record<string, unknown> | null;
+			systemPrompt?: string | null;
+		};
+
+		if (!params.id) throw new Error('id is required');
+
+		const { id, ...updateFields } = params;
+		const result = await spaceAgentManager.update(id, {
+			name: updateFields.name,
+			description: updateFields.description,
+			role: updateFields.role as import('@neokai/shared').BuiltinAgentRole | undefined,
+			model: updateFields.model,
+			provider: updateFields.provider,
+			toolConfig: updateFields.toolConfig,
+			systemPrompt: updateFields.systemPrompt,
+		});
+
+		if (!result.ok) throw new Error(result.error);
+
+		daemonHub
+			.emit('spaceAgent.updated', {
+				sessionId: `space:${result.value.spaceId}`,
+				spaceId: result.value.spaceId,
+				agent: result.value,
+			})
+			.catch((err) => {
+				log.warn('Failed to emit spaceAgent.updated:', err);
+			});
+
+		return { agent: result.value };
+	});
+
+	// spaceAgent.delete — delete an agent (blocked if referenced by workflows)
+	messageHub.onRequest('spaceAgent.delete', async (data) => {
+		const params = data as { id: string };
+		if (!params.id) throw new Error('id is required');
+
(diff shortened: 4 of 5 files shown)


lsm (commented):

### Summary

Clean, well-structured implementation that follows existing codebase conventions closely. The five CRUD RPC handlers are correctly wired, emit events using the same fire-and-forget `.catch()` pattern as `room-handlers.ts`, and the delete handler properly blocks on workflow references. 31 tests pass covering happy paths, validation errors, event emission, and the workflow-reference blocking logic.

### What looks good

- **Pattern consistency**: Handler structure, parameter validation, event emission with `.catch()`, and Logger usage all match existing handlers (e.g., `room-handlers.ts`).
- **Delete safety**: The delete handler correctly captures `spaceId` from the existing record before deletion, then uses it for event emission. The workflow-reference check with detailed error messages including workflow names is a nice touch.
- **Test quality**: Tests use real in-memory SQLite with the actual `SpaceAgentManager` and `SpaceAgentRepository`, exercising the full business-logic path rather than mocking internals. Edge cases (missing params, duplicate names, workflow references, deletion after reference removal) are all covered.
- **Type safety**: Typecheck passes cleanly. The `DaemonEventMap` entries are properly typed.

### Minor observations (P3)

1. **Double lookup in delete handler** (`space-agent-handlers.ts:131-136`): The handler calls `spaceAgentManager.getById(params.id)` to capture `spaceId`, then `spaceAgentManager.delete(params.id)` which internally calls `getById` again. This is harmless for correctness (synchronous SQLite reads are fast) and matches the pattern where the handler needs data that the manager's return type doesn't provide. Not worth changing.

2. **`sessionId` field as `space:${spaceId}`**: This is a novel pattern — existing events use actual session IDs or `'global'`. The comment in `DaemonEventMap` documents this as `channel: 'space:${spaceId}'`, which makes sense for space-scoped event routing. Just noting it as a new convention that future Space handlers should follow consistently.

3. **Inline `import()` types** (`space-agent-handlers.ts:46, 104`): The `as import('@neokai/shared').BuiltinAgentRole` casts use inline import types rather than a top-level import. This works fine but a top-level `import type { BuiltinAgentRole } from '@neokai/shared'` would be slightly cleaner.

### Verdict

**Recommendation: APPROVE** — No correctness issues, security concerns, or missing edge cases. The implementation is complete and well-tested.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two P1 issues and one P2 issue need attention before merging.

---

### Summary

The implementation is well-structured and follows existing codebase patterns. The five handlers cover the full CRUD surface, the event emission pattern (fire-and-forget with `.catch()`) is consistent with `session-handlers.ts`, the DaemonEventMap entries use the correct channel convention (`space:${spaceId}`), and the 31 unit tests use real in-memory SQLite rather than stub data. However, there are two correctness/robustness issues and one structural gap worth fixing.

---

### P1 — Missing role validation at the RPC boundary

**File:** `packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts`, lines 37-39 and 100

The handler validates that `role` is truthy but never checks it against the allowed `BuiltinAgentRole` values (`'planner' | 'coder' | 'general'`). An arbitrary string like `'leader'` or `'admin'` passes the truthy check, gets cast to `BuiltinAgentRole` via `as`, and is then passed down to the manager and repo where it will fail with a raw SQLite `CHECK constraint failed` error — instead of a clean, user-facing validation message. The DB schema does enforce the constraint (`CHECK(role IN ('planner', 'coder', 'general'))`), but the client receives an opaque database error rather than a clear validation failure.

The same gap exists in the `spaceAgent.update` handler when `role` is provided.

**Expected fix:** Add a role allowlist guard before the manager call:
```ts
const VALID_ROLES = new Set(['planner', 'coder', 'general']);
if (!VALID_ROLES.has(params.role)) throw new Error(`Invalid role: ${params.role}. Must be one of: planner, coder, general`);
```

There is no test for an invalid role value; the test suite would need a corresponding case.

---

### P1 — Double `getById` on delete (unnecessary but also masks a race)

**File:** `packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts`, lines 134-141

The handler calls `spaceAgentManager.getById(params.id)` (line 134) to capture the `spaceId` before deletion, then calls `spaceAgentManager.delete(params.id)` (line 137). Looking at `space-agent-manager.ts` line 83, `delete()` also calls `repo.getById(id)` internally. This results in two separate DB reads for the same row.

This is minor in isolation, but it also means `SpaceAgentManager.delete` does its own "not found" check and will return `{ ok: false, error: 'Agent not found: ...' }` — which should never fire here since the handler already confirmed existence on line 135. The double-check is harmless but confusing and worth documenting, or the handler could pass the pre-fetched agent to the manager to avoid the redundancy. For now this qualifies as a P1 because the redundancy creates a theoretical TOCTOU gap: if two delete requests race, the second will have passed the `getById` check on line 135 but will then observe a successful delete from `SpaceAgentManager.delete` (since `repo.delete` is idempotent — it runs `DELETE WHERE id = ?` with no row-affected check). The event will still be emitted even if the row was already gone, leading to a spurious `spaceAgent.deleted` event with the correct `agentId`.

---

### P2 — `spaceAgent.delete` event is fire-and-forget while room.delete awaits

**File:** `packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts`, lines 143-151

The `room.delete` handler in `room-handlers.ts` (line 204) uses `await daemonHub.emit(...)` before the actual deletion to ensure subscribers (e.g., StateManager) see the event before the row is gone. The `spaceAgent.delete` handler fires the event after deletion without awaiting, which is a weaker delivery guarantee. Subscribers that react to `spaceAgent.deleted` by querying the agent will see a missing row if they race against the fire-and-forget. This is low risk today since there are no known subscribers, but the inconsistency with the existing `room.delete` pattern is worth noting for future authors.

---

### P3 — No test for invalid role value

**File:** `packages/daemon/tests/unit/rpc-handlers/space-agent-handlers.test.ts`

There is no test case for submitting an invalid `role` string (e.g., `'leader'`, `'admin'`). Once the P1 handler-level validation is added, a corresponding test should be included.

---

### What works well

- All 31 tests pass against real in-memory SQLite, not stubs — the full business-logic path is covered.
- The delete-blocked-by-workflow-reference path is well tested, including workflow name inclusion in the error.
- Event payloads include `spaceId` for channel routing and follow the `space:${spaceId}` `sessionId` convention introduced by the DaemonEventMap comment.
- The helper schema file (`space-agent-schema.ts`) is well-factored and reusable.
- The `spaceAgentManager` is correctly wired through `app.ts` → `setupRPCHandlers` → handler, and the dependency is declared non-optional in `RPCHandlerDependencies`, which is correct since the manager is always available at startup.

lsm (commented):

All five previously reported issues have been properly addressed:

1. **P1 - Role validation (FIXED):** `VALID_ROLES` set and `assertValidRole()` helper are in place at lines 19-25 of `space-agent-handlers.ts`, called before any DB operation in both `create` (line 48) and `update` (line 112).

2. **P1 - TOCTOU race in delete (FIXED):** The delete handler now pre-fetches the agent (line 149), performs the deletion (line 152), and only then emits the event (line 160). The event is emitted after successful deletion, not before.

3. **P2 - Event emit order in delete (FIXED):** Delete handler uses `await daemonHub.emit(...)` (line 160), consistent with the `room.delete` pattern in `room-handlers.ts`.

4. **P3 - Inline imports (FIXED):** `space-agent-handlers.ts` uses top-level `import type { BuiltinAgentRole } from '@neokai/shared'` (line 12). The `DaemonEventMap` in `daemon-hub.ts` still uses inline `import()` syntax for `SpaceAgent`, but this is the established convention throughout that file (used ~15 times for various types).

5. **New tests (VERIFIED):** 34 tests pass, including the 3 new tests for invalid role validation (`'throws for an invalid role value'` in both create and update, and `'accepts all valid role values'`).

- Input validation is thorough: all required fields checked, role validated before DB calls
- Error messages are clear and actionable (e.g., `Invalid role: "leader". Must be one of: planner, coder, general`)
- The delete handler includes workflow reference details in error messages, which is helpful for users
- Test coverage is comprehensive: happy paths, missing params, not-found, duplicate names, workflow reference blocking, and role validation
- The `VALID_ROLES` set is consistent with the `BuiltinAgentRole` type definition in `packages/shared/src/types/space.ts` and the DB CHECK constraint

**Minor note (not blocking):** The create and update handlers use fire-and-forget for event emission (no `await`), while delete uses `await`. This inconsistency is intentional and acceptable — delete events are more critical for state consistency (subscribers need to know the resource is gone before the RPC response), while create/update events are informational.

**Recommendation: APPROVE** — All previous feedback has been addressed. The implementation is correct, well-tested, and follows existing codebase patterns.

lsm (commented):

**Recommendation: APPROVE**

All five issues from the previous review have been addressed correctly. The updated code is clean and the 34 tests all pass.

---

**P1 — Role validation** ✅ Resolved

`assertValidRole()` is implemented at `packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts:21-25` using a `VALID_ROLES = new Set(['planner', 'coder', 'general'])`. It is called before the DB write in both `spaceAgent.create` (line 48) and conditionally for `spaceAgent.update` when role is present (line 112). Two new tests (`'throws for an invalid role value'` for create and update) and one positive test (`'accepts all valid role values'`) cover this.

**P1 — TOCTOU / spurious events** ✅ Resolved

The `spaceAgent.delete` handler pre-fetches `existing` before calling `spaceAgentManager.delete()`, throws early if not found, and only emits `spaceAgent.deleted` after a successful deletion. No event is ever emitted on failure paths.

**P2 — Event emit order (await on delete)** ✅ Resolved

`spaceAgent.deleted` is now `await`-ed (line 160), matching the `room.delete → room.deleted` pattern in `room-handlers.ts`. The fire-and-forget pattern for `created`/`updated` also matches the room handler convention.

**P3 — Inline imports** ✅ Resolved

`import type { MessageHub, BuiltinAgentRole } from '@neokai/shared'` is a top-level import. The two inline `import()` type casts are gone.

**New tests** ✅ 34 tests total (was 31), all passing.

---

### Remaining Minor Observations (not blocking)

**Drift risk between `VALID_ROLES` and `BuiltinAgentRole`** (P3)

`VALID_ROLES` on line 19 is a hardcoded set that mirrors the `BuiltinAgentRole` union in `packages/shared/src/types/space.ts:335`. If a new role is added to the shared type, the runtime set must be updated manually. The `assertValidRole` function uses a TypeScript assertion signature (`asserts role is BuiltinAgentRole`), so the type system won't catch a mismatch between the set and the union. A comment near the set referencing the shared type would help keep them in sync, but this is a minor nit.

**`setTimeout(r, 0)` in delete test** (P3)

`packages/daemon/tests/unit/rpc-handlers/space-agent-handlers.test.ts:415` uses `await new Promise((r) => setTimeout(r, 0))` before asserting the deleted event. Since `spaceAgent.delete` now `await`-s the emit, this flush is unnecessary (the mock resolves synchronously). It's harmless but slightly misleading — a reader might think the emit is still fire-and-forget.

---

Both remaining observations are P3 nits. The implementation correctly addresses all previous P0/P1/P2 issues and is ready to merge.

lsm (commented):

**Recommendation: APPROVE** — Both P3 nits from the previous review are correctly addressed.

---

### Fix 1 — VALID_ROLES comment (P3 nit addressed)

`packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts`, lines 19-21:

```ts
// Keep in sync with BuiltinAgentRole in packages/shared/src/types/space.ts
// and the DB CHECK constraint on space_agents.role
const VALID_ROLES = new Set<string>(['planner', 'coder', 'general']);
```

The comment correctly names both sync points: the `BuiltinAgentRole` union in `space.ts:335` and the DB `CHECK` constraint. This is the right tradeoff. A compiler-enforced approach (e.g., exporting a const array from `@neokai/shared` and deriving both the union type and the `Set` from it) would be stronger, but TypeScript types are erased at runtime so any `Set` requires a manually maintained list regardless. The codebase does not use Zod or const-array-derived types elsewhere, so the comment is the appropriate solution here. It is clear, points directly to both sources of truth, and will surface in a code search for `BuiltinAgentRole`.

**Verdict: acceptable.**

---

### Fix 2 — setTimeout(0) removed from delete test (P3 nit addressed)

`packages/daemon/tests/unit/rpc-handlers/space-agent-handlers.test.ts`, lines 413-424:

```ts
it('emits spaceAgent.deleted event', async () => {
    // delete handler awaits the emit, so the mock is called before call() returns
    await call(hubData.handlers, 'spaceAgent.delete', { id: agentId });

    expect(daemonData.emitMock).toHaveBeenCalled();
    ...
});
```

The `await new Promise((r) => setTimeout(r, 0))` microtask flush is correctly removed. The clarifying comment accurately explains why no flush is needed: the `spaceAgent.delete` handler uses `await daemonHub.emit(...)` before returning, so the mock is guaranteed to be called by the time `await call(...)` resolves. Contrast this with the `spaceAgent.create` and `spaceAgent.update` tests (lines 167, 329) which still use `setTimeout(0)` — correctly, since those handlers fire the emit without awaiting it. The asymmetry between handlers is now clearly explained by both the handler code and the test comment.

**Verdict: correctly fixed.**

---

### No new issues introduced

The diff is minimal — only the comment and the `setTimeout(0)` removal. No behavioral changes, no new dependencies, no structural modifications. The 31 tests continue to exercise the full business logic path through real in-memory SQLite.

lsm (commented):

**Recommendation: APPROVE**

The rebase conflict resolution is correct. All three files have been merged without dropping or duplicating anything, and the new handler code is well-implemented with solid test coverage.

---

### Conflict Resolution Verification

#### 1. `packages/daemon/src/app.ts`

The resolution is correct. The dev branch introduced `spaceManager` (Task 1.4, commit `0ce89c6d`), and this PR's branch already had `spaceAgentManager` (Task 2.1, commit `48ac6d82`). The final file correctly:
- Imports both managers (lines 18-20)
- Exposes both on `DaemonAppContext` (lines 57-59)
- Instantiates both (lines 98-101)
- Passes both to `setupRPCHandlers` (lines 219-220)
- Returns both in the context object (lines 436-437)

Nothing was dropped.

#### 2. `packages/daemon/src/lib/daemon-hub.ts`

The resolution is correct. The dev branch added 8 event entries (space.created, space.updated, space.archived, space.deleted, space.task.created, space.task.updated, space.workflowRun.created, space.workflowRun.updated). This PR adds 3 more (spaceAgent.created, spaceAgent.updated, spaceAgent.deleted). The merged file (lines 405-458) contains all 11 entries in logical order — space events first, then spaceAgent events, then featureFlag events. No events were dropped or duplicated.

The `spaceAgent.*` events are correctly placed between `space.workflowRun.updated` (line 441) and `featureFlag.updated` (line 460), exactly as intended.

#### 3. `packages/daemon/src/lib/rpc-handlers/index.ts`

The resolution is correct. The dev branch wired up `setupSpaceHandlers` and `setupSpaceTaskHandlers` and added `spaceManager` to `RPCHandlerDependencies`. This PR adds `setupSpaceAgentHandlers` and `spaceAgentManager`. The merged file:
- Imports both sets of handlers (lines 46-53)
- `RPCHandlerDependencies` has both `spaceManager` and `spaceAgentManager` (lines 64-66)
- All three `setup*` calls are present and in logical order (lines 177-197)

The `setupSpaceAgentHandlers` call at line 197 is correctly appended after `setupSpaceTaskHandlers`, with a comment, before the `return () =>...

---

### Implementation Review (new code)

**space-agent-handlers.ts** — Clean implementation. Five handlers registered, all validate required params before calling the manager. The asymmetry between create/update (fire-and-forget emit with `.catch()`) and delete (awaited emit) is intentional and matches the comment at line 160, consistent with how `room.delete` works elsewhere.

One minor observation: `spaceAgent.list` and `spaceAgent.get` use synchronous manager methods but are wrapped in `async` handlers — harmless and consistent with the rest of the codebase pattern.

**Tests** — 31 tests across 5 describe blocks. Each handler has registration, happy-path, missing-param, and not-found cases. The delete handler has additional referential-integrity tests using the real SQLite in-memory DB. The `setModelsCache(new Map())` pattern correctly bypasses model validation in tests. The `mock.restore()` in `afterEach` ensures clean state between tests.

No issues found. The conflict resolution is clean and complete.

PR #433, opened 2026-03-19
feat: Space RPC handlers and DaemonEventMap registration (Task 1.4)

lsm:
- Register space.created, space.updated, space.deleted,
  space.task.created, space.task.updated, space.workflowRun.created,
  space.workflowRun.updated in DaemonEventMap
- Add space-handlers.ts: space.create (with path validation), space.list,
  space.get, space.update, space.archive, space.delete, space.overview
- Add space-task-handlers.ts: spaceTask.create, spaceTask.list,
  spaceTask.get, spaceTask.update (status → setTaskStatus, else updateTask)
- Wire both handler sets in rpc-handlers/index.ts via setupRPCHandlers()
- Add SpaceTaskManager.updateTask() for direct non-status field updates
- Expose SpaceManager in DaemonAppContext (app.ts)
- 46 unit tests — all passing; lint and typecheck clean

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -216,6 +216,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		daemonHub: eventBus,
 		db,
 		gitHubService: gitHubService ?? undefined,
+		spaceManager,
 	});
 
 	// Create WebSocket handlers
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -17,6 +17,7 @@ import { createReactiveDatabase } from './storage/reactive-database';
 import { LiveQueryEngine } from './storage/live-query';
 import { SpaceAgentRepository } from './storage/repositories/space-agent-repository';
 import { SpaceAgentManager } from './lib/space/managers/space-agent-manager';
+import { SpaceManager } from './lib/space/managers/space-manager';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -54,6 +55,8 @@ export interface DaemonAppContext {
 	liveQueries: LiveQueryEngine;
 	/** Space agent manager for Space multi-agent system */
 	spaceAgentManager: SpaceAgentManager;
+	/** Space manager for Space CRUD and workspace path validation */
+	spaceManager: SpaceManager;
 	/**
 	 * Cleanup function for graceful shutdown.
 	 * Closes all connections, stops sessions, and closes database.
@@ -93,6 +96,9 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 
 	// Initialize Space agent manager
 	const spaceAgentManager = new SpaceAgentManager(new SpaceAgentRepository(db.getDatabase()));
+
+	// Initialize Space manager
+	const spaceManager = new SpaceManager(db.getDatabase());
 
 	// Initialize authentication manager
 	const authManager = new AuthManager(db, config);
@@ -426,6 +432,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		reactiveDb,
 		liveQueries,
 		spaceAgentManager,
+		spaceManager,
 		cleanup,
 	};
 }
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -402,6 +402,43 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		templateId: string;
 	};
 
+	// Space events (global events - use 'global' as sessionId)
+	'space.created': { sessionId: string; spaceId: string; space: import('@neokai/shared').Space };
+	'space.updated': {
+		sessionId: string;
+		spaceId: string;
+		space?: Partial<import('@neokai/shared').Space>;
+	};
+	'space.deleted': { sessionId: string; spaceId: string };
+
+	// Space task events (global events - use 'global' as sessionId)
+	'space.task.created': {
+		sessionId: string;
+		spaceId: string;
+		taskId: string;
+		task: import('@neokai/shared').SpaceTask;
+	};
+	'space.task.updated': {
+		sessionId: string;
+		spaceId: string;
+		taskId: string;
+		task?: Partial<import('@neokai/shared').SpaceTask>;
+	};
+
+	// Space workflow run events (global events - use 'global' as sessionId)
+	'space.workflowRun.created': {
+		sessionId: string;
+		spaceId: string;
+		runId: string;
+		run: import('@neokai/shared').SpaceWorkflowRun;
+	};
+	'space.workflowRun.updated': {
+		sessionId: string;
+		spaceId: string;
+		runId: string;
+		run?: Partial<import('@neokai/shared').SpaceWorkflowRun>;
+	};
+
 	// Feature Flag events (PHASE 3: Gradual rollout infrastructure)
 	'featureFlag.updated': {
 		sessionId: string;
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -409,6 +409,7 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		spaceId: string;
 		space?: Partial<import('@neokai/shared').Space>;
 	};
+	'space.archived': { sessionId: string; spaceId: string; space: import('@neokai/shared').Space };
 	'space.deleted': { sessionId: string; spaceId: string };
 
 	// Space task events (global events - use 'global' as sessionId)
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -423,7 +423,7 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		sessionId: string;
 		spaceId: string;
 		taskId: string;
-		task?: Partial<import('@neokai/shared').SpaceTask>;
+		task: import('@neokai/shared').SpaceTask;
 	};
 
 	// Space workflow run events (global events - use 'global' as sessionId)
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -42,6 +42,13 @@ import { Logger } from '../logger';
 import { GoalManager } from '../room/managers/goal-manager';
 import { TaskManager } from '../room/managers/task-manager';
 import { setupDialogHandlers } from './dialog-handlers';
+// Space handlers
+import { setupSpaceHandlers } from './space-handlers';
+import { setupSpaceTaskHandlers, type SpaceTaskManagerFactory } from './space-task-handlers';
+import { SpaceManager } from '../space/managers/space-manager';
+import { SpaceTaskManager } from '../space/managers/space-task-manager';
+import { SpaceTaskRepository } from '../../storage/repositories/space-task-repository';
+import { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
 
 export interface RPCHandlerDependencies {
 	messageHub: MessageHub;
@@ -157,6 +164,25 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanu
 
 	// Dialog handlers (native OS dialogs)
 	setupDialogHandlers(deps.messageHub);
+
+	// Space handlers
+	const spaceManager = new SpaceManager(deps.db.getDatabase());
+	const spaceTaskRepo = new SpaceTaskRepository(deps.db.getDatabase());
+	const spaceWorkflowRunRepo = new SpaceWorkflowRunRepository(deps.db.getDatabase());
+
+	setupSpaceHandlers(
+		deps.messageHub,
+		spaceManager,
+		spaceTaskRepo,
+		spaceWorkflowRunRepo,
+		deps.daemonHub
+	);
+
+	const spaceTaskManagerFactory: SpaceTaskManagerFactory = (spaceId: string) => {
+		return new SpaceTaskManager(deps.db.getDatabase(), spaceId);
+	};
+
+	setupSpaceTaskHandlers(deps.messageHub, spaceManager, spaceTaskManagerFactory, deps.daemonHub);
 
 	// Return cleanup function to stop background services
 	return () => {
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -45,7 +45,7 @@ import { setupDialogHandlers } from './dialog-handlers';
 // Space handlers
 import { setupSpaceHandlers } from './space-handlers';
 import { setupSpaceTaskHandlers, type SpaceTaskManagerFactory } from './space-task-handlers';
-import { SpaceManager } from '../space/managers/space-manager';
+import type { SpaceManager } from '../space/managers/space-manager';
 import { SpaceTaskManager } from '../space/managers/space-task-manager';
 import { SpaceTaskRepository } from '../../storage/repositories/space-task-repository';
 import { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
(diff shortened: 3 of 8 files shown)


lsm (commented):

Overall, this is a well-structured implementation that meets the acceptance criteria. The handlers are clean, well-validated, properly emit DaemonHub events, and have solid test coverage (46 tests). A few issues below, one of which is a meaningful architectural concern.

### P2: Duplicate SpaceManager instantiation

**`packages/daemon/src/app.ts:101`** creates a `SpaceManager` and adds it to `DaemonAppContext`.
**`packages/daemon/src/lib/rpc-handlers/index.ts:169`** creates a *second* `SpaceManager` instance instead of receiving the one from the app context.

This means the `spaceManager` exposed on the context (used externally) and the one used by RPC handlers are different objects. Today `SpaceManager` is stateless (just wraps a repo), so this is not a functional bug, but it violates the single-source-of-truth pattern used by other managers in this codebase and will cause problems if `SpaceManager` ever gains caching or in-memory state.

The fix: pass `spaceManager` through `RPCHandlerDependencies` (like other managers) instead of creating a new one in `setupRPCHandlers`. The `spaceManager` field already exists on `DaemonAppContext` but is not wired into `RPCHandlerDependencies`.

### P3: `space.archive` emits a partial `space` object in the event

In `packages/daemon/src/lib/rpc-handlers/space-handlers.ts:120-121`, the archive handler emits `{ status: 'archived' }` as the `space` field instead of the full returned `space` object. All other mutation handlers (create, update) emit the full object. This inconsistency means event subscribers must handle both full and partial `Space` objects, which is error-prone. Consider emitting the full archived space returned from `spaceManager.archiveSpace()`.

### P3: `spaceTask.get` does not verify space existence

`spaceTask.get` (line 89-106 in `space-task-handlers.ts`) does not call `spaceManager.getSpace()` to verify the space exists, unlike `spaceTask.create` and `spaceTask.list`. The `taskManagerFactory` scopes to the space, and `getTask` validates the task belongs to that space, so this is not a security issue -- but it is an inconsistency in the validation pattern.

### P3: Fire-and-forget event emission swallows errors silently

All handlers use `.catch((err) => { log.warn(...) })` for event emission. This is a reasonable pattern for non-critical side effects, but `log.warn` only logs with a string prefix and does not include the event type in the message. If multiple events fail, it would be difficult to distinguish which one failed from logs alone. Minor, since the log messages do include descriptive text like "Failed to emit space.created:".

**Recommendation: APPROVE** (with the P2 addressed in a follow-up)

The implementation is correct, complete, and well-tested. The duplicate `SpaceManager` is the only meaningful concern, and since `SpaceManager` is currently stateless, it's not blocking.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — there are two issues worth addressing before merge.

---

### P1 — Duplicate SpaceManager instantiation (correctness + architecture)

`packages/daemon/src/app.ts` line 101 and `packages/daemon/src/lib/rpc-handlers/index.ts` line 169 each create a separate `new SpaceManager(db.getDatabase())` from the same underlying SQLite handle:

```ts
// app.ts:101
const spaceManager = new SpaceManager(db.getDatabase());

// rpc-handlers/index.ts:169
const spaceManager = new SpaceManager(deps.db.getDatabase());
```

The `spaceManager` exposed on `DaemonAppContext` is a different object than the one actually serving RPC requests. Any caller that holds a reference to `ctx.spaceManager` will use a separate instance from the handlers. Since both share the same SQLite connection this does not cause data loss, but it is architecturally inconsistent and misleading. The fix is to pass the `spaceManager` created in `app.ts` into `RPCHandlerDependencies` (the same pattern used for `sessionManager`, `authManager`, etc.) and remove the duplicate instantiation in `index.ts`.

---

### P1 — spaceTask.update sends same-status to setTaskStatus, causing spurious transition error

In `space-task-handlers.ts`, the routing branch is:

```ts
if (updateParams.status !== undefined) {
  task = await taskManager.setTaskStatus(taskId, updateParams.status, ...);
} else {
  task = await taskManager.updateTask(taskId, updateParams);
}
```

If a client calls `spaceTask.update` with the task's current status (e.g., `{ status: 'pending' }` while the task is already `pending`), the handler routes to `setTaskStatus`, which calls `isValidSpaceTaskTransition('pending', 'pending')`. The transition table does not include a self-transition for any status, so this throws:

```
Invalid status transition from 'pending' to 'pending'. Allowed: in_progress, cancelled
```

The `updateTask` path in the manager handles this correctly (it strips status when unchanged), but the handler never reaches it when `status` is present. The fix is to check `updateParams.status !== undefined` AND the status differs from the current task status before routing to `setTaskStatus`; or to fetch the task first in the handler and compare.

This is a real client-facing bug — any UI or automation that sends the current status as part of a broader metadata update will get a confusing error.

---

### P2 — Inconsistent archive event naming vs Room pattern

Rooms emit a dedicated `room.archived` event on archive (`packages/daemon/src/lib/rpc-handlers/room-handlers.ts` line 179). The space archive handler emits `space.updated` with `{ space: { status: 'archived' } }` instead of a dedicated `space.archived` event. The `DaemonEventMap` does not define `space.archived` at all. This makes it harder for listeners to distinguish an archive operation from a general metadata update, and breaks consistency with the room pattern in this codebase.

---

### P3 — Description vs title validation inconsistency

`spaceTask.create` validates title using `!params.title || params.title.trim() === ''` (rejects empty strings), but validates description using `params.description === undefined || params.description === null` (allows empty string `""`). The validation comment says "description is required" but empty descriptions pass through. This is probably intentional (description can be blank) but the error message and comment are misleading — consider using different wording like "description must not be null" or remove the check if empty string is acceptable.

---

### P3 — space.overview omits archived task flag

`space.overview` calls `taskRepo.listBySpace(params.id)` without passing `includeArchived`. Archived tasks are always excluded from the overview — this is likely intentional given the default, but there is no way for a caller to request archived tasks in the overview. Documenting this limitation would help.

---

### What's done well

- DaemonEventMap additions are correctly typed with all new event types.
- Handler files are clean, well-organized, and correctly isolated from existing handler files.
- The factory pattern for `SpaceTaskManager` is the right approach for per-space managers.
- All 46 tests pass and cover the key happy paths and error cases.
- `spaceTask.update`'s split between `setTaskStatus` (with transition validation) and `updateTask` (free-form) is a good design when the routing condition is correct.
- No `spaceGoal.*` handlers were added (matches acceptance criteria).
- Error messages are clear and actionable.

lsm (commented):

All five issues from the previous review have been properly addressed. The fix commit (`1ec8a23`) is clean and targeted.

### Previous Issues - Verification

1. **P1 (SpaceManager duplication) - RESOLVED.** `SpaceManager` is now created once in `app.ts` (line 101), passed into `RPCHandlerDependencies` (line 63 of `index.ts`), and consumed by both `setupSpaceHandlers` and `setupSpaceTaskHandlers`. No duplicate instantiation.

2. **P1 (spaceTask.update same-status transition) - RESOLVED.** Lines 135-155 of `space-task-handlers.ts` now fetch the current task and compare `updateParams.status !== currentTask.status` before routing to `setTaskStatus`. Same-status updates go through `updateTask` instead, avoiding the spurious "Invalid status transition" error. Test at line 316 explicitly verifies this.

3. **P2 (space.archived event) - RESOLVED.** `DaemonEventMap` now includes a dedicated `space.archived` entry (line 412 of `daemon-hub.ts`) with the full `Space` payload. The `space.archive` handler emits `space.archived` (not `space.updated`). Test at line 314 verifies that `space.updated` is NOT emitted on archive.

4. **P3 (spaceTask.get space validation) - RESOLVED.** Lines 100-104 of `space-task-handlers.ts` verify space existence before fetching the task. Test at line 263 confirms the space check runs before `getTask`.

5. **P3 (description validation) - RESOLVED.** Lines 43-45 of `space-task-handlers.ts` reject only `null`/`undefined` with the message "description must not be null". Empty string is explicitly allowed, verified by the test at line 159.

### Remaining Observations

**P3: Minor inconsistency in space validation for spaceTask.update.** The `spaceTask.update` handler does not validate space existence, unlike `create`, `list`, and `get`. This is not a bug (the task's existence implicitly confirms the space exists via the scoped `SpaceTaskManager`), but it breaks the consistency pattern. Acceptable as-is.

### Summary

The implementation is correct and complete. All RPC handlers are properly wired, events are registered in `DaemonEventMap`, the singleton `SpaceManager` pattern is maintained, status transitions are handled correctly, and 51 tests pass covering happy paths, validation, error propagation, and the specific edge cases raised in the prior review. **Recommendation: APPROVE.**

lsm (commented):

**Recommendation: APPROVE** (same author as reviewer — posted as COMMENT)

---

All five issues called out in the previous review have been addressed correctly. Here is a full assessment of the updated state.

### Previously-fixed issues — verified

**P1: SpaceManager double-instantiation** — Fixed. `app.ts` creates the single `SpaceManager` instance (line 101) and passes it through `RPCHandlerDependencies.spaceManager`. `rpc-handlers/index.ts` consumes it via `deps.spaceManager` and never constructs a second one.

**P1: spaceTask.update spurious transition error** — Fixed correctly. The handler now fetches the current task first (line 137 of `space-task-handlers.ts`), compares statuses, and only calls `setTaskStatus` when the status is actually changing. Same-status updates are routed to `updateTask`. The test at line 316 (`does NOT call setTaskStatus when status is unchanged`) directly verifies this path.

**P2: space.archived dedicated event** — Fixed. `DaemonEventMap` registers `space.archived` with a full `Space` payload (line 412 of `daemon-hub.ts`). The `space.archive` handler emits `space.archived` rather than `space.updated`. Test at line 314 explicitly asserts that `space.updated` is NOT emitted on archive.

**P3: spaceTask.get validates space existence** — Fixed. The handler calls `spaceManager.getSpace` and throws before touching the task manager when the space is absent. Test at line 263 verifies the task manager is never called for a missing space.

**P3: Description validation** — Fixed. `null`/`undefined` description throws `'description must not be null'`; an empty string is accepted. Tests at lines 159–192 cover all four cases.

---

### Remaining issues

#### P2 — `spaceTask.update` skips space-existence check

`spaceTask.create`, `spaceTask.list`, and `spaceTask.get` all call `spaceManager.getSpace(spaceId)` and throw `Space not found` before touching the task manager. `spaceTask.update` does not: it goes straight to `taskManagerFactory(spaceId)` and then `taskManager.getTask(taskId)`. For a well-scoped `SpaceTaskManager` this is mostly harmless (the manager will return null for any taskId because `getTask` cross-checks `task.spaceId === this.spaceId`), but the error the caller gets will say `Task not found` rather than `Space not found`, which is misleading when the real problem is an invalid `spaceId`. This inconsistency in the validation layer is observable from the RPC surface and should be consistent with the other three handlers.

`packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, lines 115–155

There is also no test for the `spaceTask.update` + space-not-found path.

#### P2 — `space.task.updated` event type declares `task?: Partial\<SpaceTask\>`, but the handler always emits the full object

`DaemonEventMap` at line 426 of `daemon-hub.ts` declares `task?: Partial\<import('@neokai/shared').SpaceTask\>`. The actual emission in `space-task-handlers.ts` at line 162 passes the full returned `SpaceTask`. This is type-safe in practice (a `SpaceTask` is assignable to `Partial\<SpaceTask\>`), but the declared type implies that consumers must handle a possibly-partial or absent task object, which will not match reality. Compare `space.task.created` which correctly declares `task: SpaceTask` (non-optional, non-partial). The `space.task.updated` event should use the same shape to avoid misleading subscribers.

`packages/daemon/src/lib/daemon-hub.ts`, line 422–427

#### P3 — `updateTask` in `SpaceTaskManager` silently strips the `status` field when status is unchanged

`SpaceTaskManager.updateTask` (lines 295–301 of `space-task-manager.ts`) throws if `params.status !== task.status`, but in the handler the same-status path calls `updateTask` with the full `updateParams` including the unchanged `status`. Inside `updateTask`, the guard at line 295 checks `params.status !== undefined && params.status !== task.status` — this will NOT throw for same-status because the condition is false, so the `status` field is stripped cleanly. This is correct behavior, but it relies on two layers agreeing silently. A comment in the handler explaining why the status field is intentionally left in `updateParams` (and stripped by `updateTask`) would prevent future confusion.

#### P3 — `spaceTask.update` no-op path (no fields sent at all)

If a caller sends `{ spaceId: 'x', taskId: 'y' }` with no other fields, the handler falls into the `else` branch (line 152–154) and calls `updateTask(taskId, {})`. `SpaceTaskRepository.updateTask` handles this gracefully (fields array is empty, so the SQL UPDATE is skipped and the task is re-fetched and returned unchanged). This is not a bug — it is a no-op that returns the current task and emits an event. Whether emitting `space.task.updated` for a no-op is intentional is a design choice worth noting.

---

### Test quality

The 46 tests cover the happy path, all validation errors, event emission, the new same-status routing, and archival event distinction thoroughly. The mock structure is clean and idiomatic. No gaps were found that block merging beyond the `spaceTask.update` missing space-existence check noted above.

lsm (commented):

**Recommendation: APPROVE** (same author as reviewer -- posted as COMMENT)

---

### Previous Issues -- All Resolved

All three issues from the prior review round have been correctly addressed in commit `e4bd9b93`:

**P2: `spaceTask.update` space-existence validation** -- RESOLVED. Lines 128-134 of `space-task-handlers.ts` now call `spaceManager.getSpace(spaceId)` before creating the task manager, returning `"Space not found: <id>"` for invalid spaceId values. Two tests verify this: line 376 (space not found throws correct error) and line 381-382 (neither `updateTask` nor `setTaskStatus` are called for a missing space).

**P2: `space.task.updated` event type** -- RESOLVED. Line 426 of `daemon-hub.ts` now declares `task: import('@neokai/shared').SpaceTask` (non-optional, full type), matching both what the handler actually emits and the `space.task.created` event shape.

**P3: Clarifying comment on same-status routing** -- RESOLVED. Lines 158-161 of `space-task-handlers.ts` explain why the unchanged `status` field is left in `updateParams` when routing to `updateTask`, and that `SpaceTaskManager.updateTask` strips it internally.

### Additional Verification

- `SpaceManager` singleton pattern remains correct: created once in `app.ts` (line 101), passed through `RPCHandlerDependencies.spaceManager`, consumed by both `setupSpaceHandlers` and `setupSpaceTaskHandlers`.
- `space.archive` handler correctly emits `space.archived` (dedicated event) with full `Space` payload.
- Same-status update routing is correct and tested.
- All 52 tests pass across both handler files.

No remaining issues found.

lsm (commented):

**Recommendation: APPROVE** (same author as reviewer — posted as COMMENT)

---

All previously raised P2 and P3 issues have been correctly addressed in commit `e4bd9b93`. Here is a full verification pass.

### Fixed Issues — Verified

**P2: spaceTask.update space-existence check**
`packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts` lines 128–134 now call `spaceManager.getSpace(spaceId)` before constructing the task manager. An invalid `spaceId` now returns `Space not found: <id>` rather than the misleading `Task not found`. The new test at line 376–383 verifies both that the error message is correct and that neither `updateTask` nor `setTaskStatus` is invoked when the space is absent. This makes `spaceTask.update` consistent with `create`, `list`, and `get`.

**P2: space.task.updated DaemonEventMap type**
`packages/daemon/src/lib/daemon-hub.ts` line 426 now declares `task: import('@neokai/shared').SpaceTask` — required and non-partial — matching both what the handler actually emits and the shape of `space.task.created`. Subscribers no longer need to defensively handle a missing or partial task object.

**P3: Clarifying comment on same-status strip**
Lines 158–161 of `space-task-handlers.ts` document why `updateParams` (which may include the unchanged `status` field) is passed as-is to `updateTask`, and that `SpaceTaskManager.updateTask` strips it internally without triggering a transition check.

### Codebase State

- `SpaceManager` singleton: created once in `app.ts` (line 101), threaded through `RPCHandlerDependencies.spaceManager`, consumed by both `setupSpaceHandlers` and `setupSpaceTaskHandlers` — no duplicate instantiation.
- `space.archive` correctly emits the dedicated `space.archived` event (not `space.updated`) with the full returned `Space` payload.
- Same-status routing in `spaceTask.update` is correct: status is compared against the fetched current task before routing to either `setTaskStatus` or `updateTask`.
- 52 unit tests pass across both handler files; all handler validation paths and event emission are covered.

### Residual Observations (non-blocking)

- `space.updated` DaemonEventMap entry (line 407–411) still declares `space?: Partial<Space>` while the handler always emits a full `Space`. This is the same minor type looseness that existed for `space.task.updated` before the fix. TypeScript does not reject it (a full `Space` is assignable to `Partial<Space>`), but it tells subscribers they may receive a partial object when in practice they never will. Not blocking, but worth noting as a pattern to apply consistently in a future pass.
- The `spaceTask.update` no-op path (empty params object) emits `space.task.updated` even when no fields changed. This is not a bug, but whether that event emission is intentional is a design choice that could be documented.

No blocking issues remain. The PR is ready to merge.

PR #435, opened 2026-03-19
feat: add task management tools to leader role

lsm:
Leaders can now directly modify tasks without going through workers:
- update_task: edit title, description, priority, and dependencies
- cancel_task: cancel a task and cascade to pending dependents
- update_task_status: change task status with transition validation

Also adds daemonHub to TaskGroupManagerConfig/TaskGroupManager and
LeaderAgentConfig so task update events are emitted to the UI when
the leader modifies tasks. Updates leader system prompt to document
the new tools.

### DIFF
--- a/packages/daemon/src/lib/room/agents/leader-agent.ts
+++ b/packages/daemon/src/lib/room/agents/leader-agent.ts
@@ -28,6 +28,7 @@ import type {
 import type { GoalManager } from '../managers/goal-manager';
 import type { TaskManager } from '../managers/task-manager';
 import type { SessionGroupRepository } from '../state/session-group-repository';
+import type { DaemonHub } from '../../daemon-hub';
 import { createLeaderContextMcpServer } from '../tools/room-agent-tools';
 
 const DEFAULT_LEADER_MODEL = 'claude-sonnet-4-5-20250929';
@@ -76,6 +77,8 @@ export interface LeaderAgentConfig {
 	goalManager?: GoalManager;
 	taskManager?: TaskManager;
 	groupRepo?: SessionGroupRepository;
+	/** Used to emit task update events to the UI when leader modifies tasks */
+	daemonHub?: DaemonHub;
 }
 
 /**
@@ -144,6 +147,8 @@ function leaderToolContractSection(): string {
 ## Tool Contract (CRITICAL)
 
 You MUST call tools (no text-only final responses).
+
+### Review Tools (primary actions)
 - \`send_to_worker\` — Forward feedback to worker without changing group ownership
   - mode=\`queue\`: enqueue for next-turn processing (default, preferred for review URLs)
   - mode=\`steer\`: inject for current-turn steering
@@ -151,6 +156,14 @@ You MUST call tools (no text-only final responses).
 - \`fail_task\` — Mark the task as not achievable
 - \`replan_goal\` — The current approach isn't working; fail this task and trigger replanning with context about what was tried
 - \`submit_for_review\` — Work is done with a PR ready; submit for peer review and human approval
+
+### Task Management Tools (for managing other tasks in the room)
+- \`update_task\` — Edit title, description, priority, or dependencies of any task
+- \`cancel_task\` — Cancel a task and cascade to any pending dependents
+- \`update_task_status\` — Change a task's status (e.g., retry a failed task: needs_attention → pending)
+
+### Context Tools (read-only)
+- \`list_goals\`, \`list_tasks\`, \`get_task_detail\`, \`get_room_status\` — Inspect room state
 
 Do NOT respond with only text.`;
 }
@@ -1026,9 +1039,8 @@ export function createLeaderAgentInit(
 ): AgentSessionInit {
 	const mcpServer = createLeaderMcpServer(config.groupId, callbacks);
 
-	// Create a read-only context MCP server for the leader (list/get tools only).
-	// Deliberately excludes write tools, human-only tools (approve_task, reject_task),
-	// and write-capable dependencies (daemonHub, runtimeService).
+	// Create a context + task management MCP server for the leader.
+	// Excludes human-only tools (approve_task, reject_task) and session management (stop_session).
 	const roomAgentTools =
 		config.goalManager && config.taskManager && config.groupRepo
 			? (createLeaderContextMcpServer({
@@ -1036,6 +1048,7 @@ export function createLeaderAgentInit(
 					goalManager: config.goalManager,
 					taskManager: config.taskManager,
 					groupRepo: config.groupRepo,
+					daemonHub: config.daemonHub,
 				}) as unknown as McpServerConfig)
 			: undefined;
 
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -280,6 +280,7 @@ export class RoomRuntime {
 			getRoom: config.getRoom,
 			getTask: config.getTask,
 			getGoal: config.getGoal,
+			daemonHub: config.daemonHub,
 		});
 
 		// Keep test and direct-runtime usage predictable: when no explicit leader model
--- a/packages/daemon/src/lib/room/runtime/task-group-manager.ts
+++ b/packages/daemon/src/lib/room/runtime/task-group-manager.ts
@@ -23,6 +23,7 @@ import type {
 import type { SessionObserver, TerminalState } from '../state/session-observer';
 import type { TaskManager } from '../managers/task-manager';
 import type { GoalManager } from '../managers/goal-manager';
+import type { DaemonHub } from '../../daemon-hub';
 import type { LeaderToolCallbacks } from '../agents/leader-agent';
 import { createLeaderAgentInit } from '../agents/leader-agent';
 import type { LeaderAgentConfig, ReviewContext } from '../agents/leader-agent';
@@ -161,6 +162,8 @@ export interface TaskGroupManagerConfig {
 	getTask: (taskId: string) => Promise<NeoTask | null>;
 	/** Fetch goal from DB by ID. Used to get CURRENT goal data at route time. */
 	getGoal: (goalId: string) => Promise<RoomGoal | null>;
+	/** Used to emit task update events when leader modifies tasks */
+	daemonHub?: DaemonHub;
 }
 
 export class TaskGroupManager {
@@ -172,6 +175,7 @@ export class TaskGroupManager {
 	private readonly getRoom: (roomId: string) => Room | null;
 	private readonly getTaskById: (taskId: string) => Promise<NeoTask | null>;
 	private readonly getGoalById: (goalId: string) => Promise<RoomGoal | null>;
+	private readonly daemonHub?: DaemonHub;
 	readonly workspacePath: string;
 	private _model?: string;
 	readonly workerModel?: string;
@@ -188,6 +192,7 @@ export class TaskGroupManager {
 		this.workspacePath = config.workspacePath;
 		this._model = config.model;
 		this.workerModel = config.workerModel;
+		this.daemonHub = config.daemonHub;
 	}
 
 	/** Get the current model for leader sessions */
@@ -307,6 +312,7 @@ export class TaskGroupManager {
 			goalManager: this.goalManager,
 			taskManager: this.taskManager,
 			groupRepo: this.groupRepo,
+			daemonHub: this.daemonHub,
 		};
 		const leaderInit = createLeaderAgentInit(leaderConfig, leaderCallbacks);
 		await this.sessionFactory.createAndStartSession(leaderInit, 'leader');
@@ -427,6 +433,7 @@ export class TaskGroupManager {
 				goalManager: this.goalManager,
 				taskManager: this.taskManager,
 				groupRepo: this.groupRepo,
+				daemonHub: this.daemonHub,
 			};
 			const leaderInit = createLeaderAgentInit(leaderConfig, leaderCallbacks);
 
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -13,7 +13,6 @@
 import { createSdkMcpServer, tool } from '@anthropic-ai/claude-agent-sdk';
 import { z } from 'zod';
 import type { TaskStatus, TaskType, AgentType } from '@neokai/shared';
-import { VALID_STATUS_TRANSITIONS } from '../managers/task-manager';
 import type { GoalManager } from '../managers/goal-manager';
 import type { TaskManager } from '../managers/task-manager';
 import type { SessionGroupRepository } from '../state/session-group-repository';
@@ -176,6 +175,10 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 			if (!task) {
 				return jsonResult({ success: false, error: `Task not found: ${args.task_id}` });
 			}
+			// Guard against self-dependency (would permanently block the task)
+			if (args.depends_on?.includes(args.task_id)) {
+				return jsonResult({ success: false, error: 'A task cannot depend on itself.' });
+			}
 			const updates: {
 				title?: string;
 				description?: string;
@@ -214,6 +217,27 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				return jsonResult({ success: false, error: `Task not found: ${args.task_id}` });
 			}
 
+			// Guard: terminal tasks cannot be cancelled
+			if (task.status === 'completed' || task.status === 'cancelled') {
+				return jsonResult({
+					success: false,
+					error: `Task ${args.task_id} is already in terminal state '${task.status}' and cannot be cancelled.`,
+				});
+			}
+
+			// Guard: cancelling a task with an active session group requires runtime service
+			// to stop the running session. Without it, the DB would be updated but the
+			// worker session would continue running against a cancelled task.
+			if (task.status === 'in_progress' && !runtimeService) {
+				const activeGroup = groupRepo.getGroupByTaskId(args.task_id);
+				if (activeGroup && activeGroup.completedAt === null) {
+					return jsonResult({
+						success: false,
+						error: `Cannot cancel in_progress task ${args.task_id} without runtime service — the active worker session would not be stopped.`,
+					});
+				}
+			}
+
 			let cancelledTaskIds: string[] = [];
 			let usedRuntimeCancellation = false;
 			if (runtimeService) {
@@ -249,7 +273,11 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 					}
 				}
 			}
-			return jsonResult({ success: true, message: `Task ${args.task_id} cancelled` });
+			return jsonResult({
+				success: true,
+				cancelledTaskIds,
+				message: `Task ${args.task_id} and ${cancelledTaskIds.length - 1} dependent task(s) cancelled`,
+			});
 		},
 
 		async stop_session(args: { task_id: string }): Promise<ToolResult> {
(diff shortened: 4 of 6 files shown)


lsm (commented):

### Summary

This PR adds 3 task management MCP tools (`update_task`, `cancel_task`, `update_task_status`) to the leader's context MCP server, wires `daemonHub` through `TaskGroupManager` and `LeaderAgentConfig`, updates the system prompt tool contract documentation, and adds comprehensive tests. The implementation is clean and well-structured.

### Findings

**P1: `cancel_task` and `update_task_status` lack `runtimeService` -- potential state inconsistency**

The `LeaderContextMcpConfig` type intentionally excludes `runtimeService`, which means the leader's tools always hit fallback paths:

- `cancel_task` (lines 219-232 of `room-agent-tools.ts`): Without `runtimeService`, cancellation is DB-only via `taskManager.cancelTaskCascade`. If the task has an active worker session currently running, the session will NOT be stopped -- it will keep executing against a task now marked as cancelled in the DB. The worker's next tool call or completion will encounter a stale/cancelled task.

- `update_task_status` (lines 308-331): When transitioning an `in_progress` or `review` task to a terminal state (`completed`, `needs_attention`, `cancelled`), the active session group should be cancelled first. Without `runtimeService`, this entire block is skipped. The DB status changes but the running session group persists, leading to a zombie worker session.

This is partly by design (the comment says "Excludes ... session management"), but the consequence is that the leader can create inconsistent runtime state. The runtime tick should eventually detect the mismatch, but there is no guarantee and the window could be significant.

Consider one of:
1. Passing `runtimeService` to the leader context config (preferred -- the leader already runs inside the runtime)
2. Documenting the limitation explicitly and adding a guard that rejects `cancel_task` / status transitions for `in_progress` tasks when `runtimeService` is unavailable
3. Having the fallback path emit an event that the runtime can react to for cleanup

**P2: `update_task` emits `daemonHub` event but tests don't verify it**

The `update_task` handler emits `room.task.update` via `daemonHub` (line 201-207), but all tests in the `update_task tool` describe block create handlers without `daemonHub` in the config. There are no tests verifying the event is emitted or that it works correctly when `daemonHub` is provided. The same applies to `cancel_task` in fallback mode.

**P3: Redundant individual tool-presence tests**

The three tests at lines 1734-1749 (`should expose update_task for editing task fields`, `should expose cancel_task for cancelling tasks`, `should expose update_task_status for status transitions`) are redundant with the comprehensive list test at line 1720. They test `.toContain()` on the same array that was already asserted with `.toEqual()`. Minor nit -- not blocking.

**P3: `update_task_status` tool name vs handler name mismatch**

The MCP tool is registered as `update_task_status` but delegates to `handlers.set_task_status`. This naming inconsistency is cosmetic (the mapping is explicit and correct), but worth noting for future maintainability. The test description also acknowledges this: `"update_task_status tool (set_task_status handler)"`.

### Verdict

The core implementation is correct for the happy path and the code reuse of existing handlers is a good pattern. The main concern is the missing `runtimeService` creating a window for runtime/DB state inconsistency when the leader cancels or transitions active tasks. This warrants discussion on whether it's acceptable or needs mitigation.

**Recommendation: REQUEST_CHANGES** (due to P1 -- state inconsistency risk for active tasks)

lsm (commented):

**Recommendation: REQUEST_CHANGES** — two correctness bugs and one missing guard before approving.

---

## Summary

The overall shape of this PR is correct: the three new tools (`update_task`, `cancel_task`, `update_task_status`) reuse existing, well-tested handler functions, the daemonHub propagation chain is complete and consistent, and the system prompt updates accurately document the new capabilities. The tests cover the happy path and key error cases well.

That said, I found two bugs and one gap that should be addressed before merge.

---

## P0 — Bugs

### 1. `cancel_task` silently succeeds on terminal tasks (completed/already-cancelled)

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 211-253

The `cancel_task` handler calls `taskManager.cancelTaskCascade()` (fallback path) or `runtime.cancelTask()` (runtime path) with no guard on the current status. `cancelTaskCascade` delegates to `updateTaskStatus` (not `setTaskStatus`), which does **not** validate status transitions. This means calling `cancel_task` on a task already in `completed` or `cancelled` state silently overwrites the terminal status — in the fallback path this will flip a completed task back to cancelled. The runtime path (`runtime.cancelTask`) calls the same `cancelTaskCascade` internally, so it is equally affected.

A guard like the following should be added before the cancellation logic:

```ts
const terminalStatuses = ['completed', 'cancelled'];
if (terminalStatuses.includes(task.status)) {
  return jsonResult({ success: false, error: `Task ${args.task_id} is already in terminal state '${task.status}' and cannot be cancelled.` });
}
```

### 2. `update_task` permits circular/self-referencing dependencies

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 168-209

The `update_task` handler passes the new `depends_on` list to `taskManager.updateTaskFields`, which only validates that each listed dependency ID exists in the room. It does not check:

- **Self-dependency**: `depends_on: [task_id]` — the task permanently blocks itself.
- **Circular dependency**: Task A depends on B, and the leader now sets B to depend on A.

Neither the tool handler nor `updateTaskFields` performs cycle detection. The runtime's tick loop will deadlock silently for any affected task (dependencies never complete, task stays pending forever). `createTask` has the same gap (pre-existing), but `update_task` makes it easier for an LLM to introduce cycles at runtime when restructuring a task graph.

A minimum fix is a self-dependency check in the handler:

```ts
if (args.depends_on?.includes(args.task_id)) {
  return jsonResult({ success: false, error: 'A task cannot depend on itself.' });
}
```

Full cycle detection would be the complete fix, but is a larger scope change.

---

## P1 — Should Fix

### 3. `cancel_task` success response omits which tasks were cascade-cancelled

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, line 252

The success response is:
```ts
return jsonResult({ success: true, message: `Task ${args.task_id} cancelled` });
```

The `cancelledTaskIds` variable is computed but never returned to the LLM. When a cascade cancels 3 dependent tasks the leader has no way to know without making a follow-up `list_tasks` call. Returning the affected IDs in the response would make the tool self-documenting:

```ts
return jsonResult({ success: true, cancelledTaskIds, message: `Task ${args.task_id} and ${cancelledTaskIds.length - 1} dependent task(s) cancelled` });
```

This also resolves a minor inconsistency: `cancelledTaskIds` is populated but used only for daemonHub emission in the fallback path; in the runtime path it is computed and then thrown away.

---

## P2 — Suggestions

### 4. Redundant transition validation in `set_task_status` handler

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 299-306

The handler validates the transition using `VALID_STATUS_TRANSITIONS` before calling `taskManager.setTaskStatus`, which performs the exact same validation and throws. The double-check is harmless but produces a slightly different error message format depending on which layer catches it first. Consider removing the pre-check and letting `setTaskStatus` surface the error via the `try/catch` already present at lines 369-377.

### 5. New tests for `update_task` and `cancel_task` are redundant

**File:** `packages/daemon/tests/unit/room/room-agent-tools.test.ts`, lines 1734-1750

Three of the new tests in `createLeaderContextMcpServer` ("should expose update_task", "should expose cancel_task", "should expose update_task_status") are complete subsets of the existing exhaustive tool list test at line 1720. They add no new signal and inflate the test count without improving coverage.

---

## P3 — Nits

- **Tool name vs. handler name mismatch**: The MCP tool is named `update_task_status` but it calls `handlers.set_task_status`. This is not a bug (the tool name is what the LLM sees; the handler name is internal), but the asymmetry can confuse future readers. A comment explaining the mapping would help.
- **`cancelledTaskIds` variable is declared but its value is unused in the runtime path** (line 229 assigns it, nothing reads it afterward). A linter with `no-unused-vars` may flag this depending on whether it reaches the daemonHub loop below — it doesn't, since daemonHub emission is only in the fallback branch.

lsm (commented):

### Re-review Summary

All previously raised issues have been properly addressed in commit `b45dd8ce`. The implementation is clean and the fixes are correct.

### Previous Issues — Resolution Verified

1. **P1: `cancel_task` and `update_task_status` lacked `runtimeService` guard** — Fixed. Both now check for active session groups when `runtimeService` is unavailable and return clear error messages explaining why the operation cannot proceed (lines 231-239 and 334-346 of `room-agent-tools.ts`).

2. **P2: `cancelledTaskIds` not returned** — Fixed. The `cancel_task` response now includes `cancelledTaskIds` array in both runtime and fallback paths (lines 276-280).

3. **`cancel_task` now rejects terminal states** — Correct. Lines 221-226 guard against cancelling already-completed or already-cancelled tasks.

4. **`update_task` now rejects self-dependencies** — Correct. Lines 179-181 check `args.depends_on?.includes(args.task_id)`.

5. **Removed redundant pre-validation in `set_task_status`** — Correct. The `VALID_STATUS_TRANSITIONS` import and manual transition check have been replaced by relying on `taskManager.setTaskStatus()` which performs its own validation (line 410). This delegates validation to the single source of truth.

6. **Removed unused `VALID_STATUS_TRANSITIONS` import** — Confirmed removed.

7. **Removed redundant individual tool-presence tests** — Confirmed. The test at line 1720 checks all 7 tools in one assertion, and excluded tools are verified in a single test at line 1746.

### Current State Assessment

**Correctness**: The three new leader tools (`update_task`, `cancel_task`, `update_task_status`) correctly share handler implementations with the room-agent tools via `createRoomAgentToolHandlers`. The `LeaderContextMcpConfig` type properly excludes `runtimeService` from the leader context, and the handlers guard against operations that need it.

**Test Coverage**: Comprehensive — 110 tests pass in `room-agent-tools.test.ts`. The new leader context section (lines 1709-2027) covers:
- Tool enumeration (7 tools exposed, excluded tools verified)
- `update_task`: field updates, priority, dependencies, self-dependency rejection, not-found
- `cancel_task`: single task, cascade, terminal state rejection, in_progress without runtime guard, not-found
- `update_task_status`: valid transition, runtime guard, invalid transition delegation, not-found
- Server name distinction (`leader-context` vs `room-agent`)

**Architecture**: The `update_task_status` tool name maps to the internal `set_task_status` handler with a clear comment explaining the naming (line 316-317). This keeps the internal handler name consistent with the existing pattern while giving the LLM a more descriptive tool name.

**One minor observation (P3)**: The `cancel_task` message at line 279 says `"Task X and N dependent task(s) cancelled"`. When N=0 this reads "0 dependent task(s)" which is slightly verbose but functionally fine — the `cancelledTaskIds` array is the machine-readable truth.

### Recommendation: APPROVE

lsm (commented):

The previous P0 and P1 issues are substantially addressed — terminal-state cancellation is guarded, self-dependency is rejected, and runtime-service guards protect active-session transitions for `set_task_status`. The `cancelledTaskIds` are now returned and tests are meaningfully expanded. However, there is a new P1 gap introduced by the inconsistency between the two guards, and two P2 issues related to schema consistency.

---

### P1 — `cancel_task` guard does not cover `review` status tasks with active groups

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, line 231

The zombie-session guard reads:

```ts
if (task.status === 'in_progress' && !runtimeService) {
```

This only protects `in_progress` tasks. A task in `review` status also has an active worker/leader session group (`completedAt === null`). `cancelTaskCascade` (called on the fallback path) uses `updateTaskStatus` which bypasses transition-table enforcement, so it can and does cancel a `review` task. Without `runtimeService`, that leaves the agent session running against a now-cancelled task — the exact zombie scenario the guard is supposed to prevent.

The `set_task_status` handler (lines 334–346) handles this correctly: it guards both `in_progress` and `review`. The `cancel_task` handler should match:

```ts
if ((task.status === 'in_progress' || task.status === 'review') && !runtimeService) {
```

There is no test for this scenario (cancel a `review` task with an active group and no runtimeService), which is how it slipped through.

---

### P2 — `update_task` in leader-context server drops input validation from the full tool

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 1099–1115

The full `room-agent` MCP registration uses:
```ts
title: z.string().trim().min(1).optional()
description: z.string().trim().min(1).optional()
```

The leader-context registration uses:
```ts
title: z.string().optional()
description: z.string().optional()
```

The `.trim().min(1)` validation is dropped. A leader LLM call could supply empty strings for title or description, which would be stored as-is. Since both registrations delegate to the same `handlers.update_task` function (which has no server-side length guard), an empty `""` title would be accepted. The full-tool version's validators exist for a reason — the leader-context version should replicate them.

---

### P2 — `update_task_status` leader tool omits `draft` from the status enum without documentation

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 1129–1131

The full `set_task_status` tool schema includes `'draft'` in the enum; the leader's `update_task_status` tool omits it. This is probably intentional (leaders should not be able to put tasks back to draft), but the exclusion comment at line 1055–1059 does not mention it. A one-line note (e.g., `// draft: leader must not use; tasks reach draft only via the planning phase`) would prevent a future maintainer from accidentally re-adding it or wondering why the inconsistency exists.

---

### P3 — Misleading message when exactly one task is cancelled (no cascade)

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, line 279

```ts
message: `Task ${args.task_id} and ${cancelledTaskIds.length - 1} dependent task(s) cancelled`
```

When only the target task is cancelled (no dependents), this produces: `"Task X and 0 dependent task(s) cancelled"`. The "and 0 dependent task(s)" phrasing is semantically odd and looks like a bug to LLMs reading tool output. Consider branching: something like `cancelledTaskIds.length > 1 ? `... and ${cancelledTaskIds.length - 1} dependent task(s)` : `Task X cancelled``.

---

### What was fixed correctly

- Terminal-state guard in `cancel_task` (completed/cancelled) — correct.
- `cancel_task` guard for `in_progress` tasks with active groups — correct, just incomplete (see P1).
- `set_task_status` guard for active-group + no runtimeService — covers both `in_progress` and `review`, correct.
- Self-dependency rejection in `update_task` — correct.
- `cancelledTaskIds` returned in success response — done.
- Test renamed from misleading 'should allow status change without runtime...' to 'should reject...' — good.
- New tests for all three tools are thorough on the happy path and the most important failure modes.

lsm (commented):

All four requested fixes from the previous review have been correctly implemented:

**P1 fix (zombie-session guard for `review` status):** Line 232 of `room-agent-tools.ts` now checks `task.status === 'in_progress' || task.status === 'review'`, ensuring that cancelling a task in review with an active session group is properly guarded. A dedicated test (`should reject cancellation of review task with active group without runtimeService`) confirms this behavior.

**P2 fix (trim/min validators):** Both the room-agent and leader-context `update_task` schemas now use `z.string().trim().min(1).optional()` for `title` and `description` fields (lines 902-903 and 1108-1109), preventing empty or whitespace-only updates.

**P2 fix (draft exclusion comment):** Lines 1134-1135 include a clear inline comment explaining why `draft` is intentionally excluded from the leader's `update_task_status` enum.

**P3 fix (dependent count message):** Lines 282-284 now conditionally include the dependent clause only when `dependentCount > 0`, avoiding the misleading "0 dependent task(s)" phrasing. A test verifies `result.message` does not contain "0 dependent".

The implementation is clean and well-tested. The `daemonHub` is properly threaded through from `RoomRuntime` -> `TaskGroupManager` -> `LeaderAgentConfig` -> `createLeaderContextMcpServer`, ensuring UI notifications work when the leader modifies tasks. The leader system prompt is updated to document the new task management tools. Test coverage is thorough with 7 tool registration checks, 5 update_task tests, 5 cancel_task tests, and 4 update_task_status tests.

**Recommendation: APPROVE** -- no remaining issues found.

lsm (commented):

**Recommendation: APPROVE** (with one P2 note)

All four issues from the previous review have been addressed. The fixes are correct and the new test coverage is thorough.

---

### What was fixed — verified correct

**P1: cancel_task zombie-session guard now covers review status**

The guard at line 232 of `room-agent-tools.ts` now reads:

```ts
if ((task.status === 'in_progress' || task.status === 'review') && !runtimeService) {
```

Covered by the new test at lines 1925–1941 in `room-agent-tools.test.ts` which creates a task in `review` status with an active group and asserts that cancellation is rejected without runtime service. This is correct.

**P2: update_task schema adds .trim().min(1)**

Both `title` and `description` in the `LeaderContextMcpConfig` server's `update_task` tool now have `.trim().min(1)` (lines 1108–1109). The main room-agent server's `update_task` tool at line 902–903 has the same validation. Consistent and correct.

**P2: update_task_status excludes draft with inline comment**

The `draft` exclusion is now documented with an inline comment at line 1134–1135, explaining that `draft` tasks are created exclusively by the planner and the leader must not put tasks back to draft. Clear and intentional.

**P3: cancel_task message no longer shows "0 dependent" phrasing**

The message now only includes the dependent count clause when `dependentCount > 0` (lines 282–284). Verified by the new test at line 1863–1864 which asserts `result.message` does not contain "0 dependent".

---

### Remaining issue — P2

**Misleading error message in cancel_task guard for review-status tasks**

`/packages/daemon/src/lib/room/tools/room-agent-tools.ts` line 237:

```ts
error: `Cannot cancel in_progress task ${args.task_id} without runtime service...`,
```

When the task status is `review`, the error message incorrectly says "in_progress task". The guard now correctly covers both statuses, but the error text was not updated to reflect this. A caller cancelling a `review` task without runtime service sees a confusing error that describes the task as in_progress when it is not.

Suggested fix:
```ts
error: `Cannot cancel task ${args.task_id} (status '${task.status}') without runtime service — the active worker session would not be stopped.`,
```

Note: the analogous guard in `set_task_status` (line 348) already uses the pattern `status '${task.status}'` — the cancel_task guard should match it for consistency.

---

- The leader system prompt additions are clear and correctly document the three new tools under "Task Management Tools".
- `LeaderContextMcpConfig` correctly excludes `runtimeService`, ensuring the zombie-session guard always fires for leader-invoked tools (runtimeService is always undefined in that path).
- Test coverage is comprehensive: self-dependency, cascade, terminal-state rejection, review-status guard, clean message format, schema validation, and tool exposure checks all have dedicated tests.
- The `set_task_status` handler correctly uses `task.status` in its error message — only the `cancel_task` guard lags behind.

lsm (commented):

The fix is correct and well-scoped. The one-line change at `packages/daemon/src/lib/room/tools/room-agent-tools.ts:237` replaces the hardcoded `"in_progress"` in the `cancel_task` guard error message with `'${task.status}'`, which now correctly shows the actual task status (either `in_progress` or `review`). This matches the pattern already used in `set_task_status` (line 348).

**Verification:**
- The guard condition on line 232 explicitly checks `task.status === 'in_progress' || task.status === 'review'`, so the interpolated status will always be one of these two values in this code path -- no risk of exposing unexpected internal state.
- Existing tests at lines 1908-1941 (`should reject cancellation of in_progress task...` and `should reject cancellation of review task...`) both assert `toContain('runtime service')`, which still passes with the new message format.
- The broader PR (all 4 commits) adding `update_task`, `cancel_task`, and `update_task_status` to the leader role looks solid: proper guards, daemonHub plumbing through TaskGroupManager, system prompt updates, and comprehensive test coverage (295+ new lines of tests).

**Recommendation: APPROVE** -- no issues found.

lsm (commented):

**Recommendation: APPROVE**

### Targeted Fix Verified

The P2 issue from the previous round has been correctly addressed.

**Before (commit 751f4357):**
```
Cannot cancel in_progress task ${args.task_id} without runtime service — the active worker session would not be stopped.
```

**After (commit 0db19a78):**
```
Cannot cancel task ${args.task_id} (status '${task.status}') without runtime service — the active worker session would not be stopped.
```

The change is in `packages/daemon/src/lib/room/tools/room-agent-tools.ts` at line 237. The error message now uses `task.status` (the actual runtime value) instead of the hardcoded string `in_progress`, so a task in `review` status will produce `status 'review'` in the error. This exactly matches the pattern already used in `set_task_status` at line 347.

### Test Coverage

The test at line 1925–1941 of `packages/daemon/tests/unit/room/room-agent-tools.test.ts` exercises the `review`-status path and asserts `result.error` contains `'runtime service'`. The assertion is valid and would catch a regression where the guard doesn't fire at all.

The only minor observation (not blocking) is that the test does not assert `status 'review'` appears in the message — it would be slightly stronger to also assert `expect(result.error).toContain("status 'review'")`. However, this is a P3 nit and the fix itself is correct.

PR #438, opened 2026-03-19
feat: Space navigation, routing, and signals (Task 5.1)

lsm:
- Extend NavSection union to include 'spaces'
- Add currentSpaceIdSignal, currentSpaceSessionIdSignal, currentSpaceTaskIdSignal
- Add /space/:id, /space/:id/session/:sid, /space/:id/task/:tid route patterns
- Add navigateToSpace, navigateToSpaceSession, navigateToSpaceTask, navigateToSpaces
- Update initializeRouter and handlePopState to handle space URLs (priority over room routes)
- Add 'Spaces' entry to MAIN_NAV_ITEMS with grid icon
- Create SpaceContextPanel with active/archived filter and space list
- Create SpaceIsland placeholder component
- Wire ContextPanel and MainContent for 'spaces' section
- 35 unit tests covering all space routing paths


lsm (commented):

### Summary

Task 5.1 is well-structured overall. The signals, route patterns, path helpers, nav config, context panel, and test file all follow existing patterns correctly. However, there are signal-clearing omissions in the existing navigation functions that will cause broken navigation when switching between spaces and rooms/sessions.

### P0: `navigateToSession`, `navigateToRoom*`, and `navigateToSessions` do not clear space signals

**File:** `packages/web/src/lib/router.ts`

`navigateToSession` (lines 203-238), `navigateToRoom` (lines 287-328), `navigateToRoomSession` (lines 338-379), `navigateToRoomTask` (lines 389-421), and `navigateToSessions` (lines 427-456) do **not** clear `currentSpaceIdSignal`, `currentSpaceSessionIdSignal`, or `currentSpaceTaskIdSignal`.

In `MainContent.tsx`, `spaceId` is checked **before** `roomId` (line 36 vs line 41). This means if a user is viewing a space and then navigates to a room via `navigateToRoom()`, `currentSpaceIdSignal` will still be set, and `MainContent` will continue rendering `SpaceIsland` instead of `Room`.

The `handlePopState` and `initializeRouter` functions correctly clear space signals for non-space routes (good), but the imperative navigation functions do not. This is a real functional bug that will break room/session navigation when coming from a space view.

**Fix needed:** Add the following three lines to all five functions listed above, in both the same-path early-return branch and the main navigation branch:
```
currentSpaceIdSignal.value = null;
currentSpaceSessionIdSignal.value = null;
currentSpaceTaskIdSignal.value = null;
```

### P1: Same-path early-return branches in `navigateToSpace*` don't set `navSectionSignal`

**File:** `packages/web/src/lib/router.ts`, lines 510-516, 549-555, 588-594

When the current URL already matches the target path, the early-return branch sets the space signals but does **not** set `navSectionSignal.value = 'spaces'`. This means if `navSectionSignal` is somehow out of sync (e.g., `'home'`), the nav rail highlight and context panel won't update even though the space is selected. The existing `navigateToRoom` function has the same pattern (doesn't set navSection in same-path branch), so this may be an existing convention -- but it's inconsistent since the main branch does set it.

### P2: `navigateToSpaces` calls `navigateToHome` which clears space signals, producing a potentially incorrect state

**File:** `packages/web/src/lib/router.ts`, lines 493-496

`navigateToSpaces` sets `navSectionSignal = 'spaces'` then calls `navigateToHome()`. But `navigateToHome` clears `currentSpaceIdSignal` (and all space signals). If `navigateToHome` is called while already on `/`, it early-returns after clearing signals -- the `navSectionSignal` set beforehand is preserved. But if the user is on `/space/xxx`, `navigateToHome` will push to `/` and clear `navSectionSignal` is NOT touched by `navigateToHome`... wait, actually looking more carefully, `navigateToHome` does NOT touch `navSectionSignal`. So it works correctly: sets nav to 'spaces', then navigates to `/` clearing space ID. This is fine -- withdrawing this concern.

### P3: Test file uses `@ts-nocheck`

**File:** `packages/web/src/lib/__tests__/space-router.test.ts`, line 1

The `@ts-nocheck` directive suppresses all TypeScript errors. While understandable for tests that mock `window.history` and `window.location`, this also hides any type errors in the test assertions or imports. Consider using targeted `@ts-expect-error` comments on the specific `Object.defineProperty` calls instead.

### Verdict

The P0 issue (missing signal clearing in existing navigate functions) will cause broken navigation when switching from spaces to rooms/sessions. The space view will persist when it shouldn't because `MainContent.tsx` checks `spaceId` before `roomId`. This needs to be fixed before merge.

lsm (commented):

The implementation is largely correct and well-structured, but there is one P0 bug and several P1 issues that need to be fixed before merging.

---

### P0 — NavRail does not handle the `'spaces'` section

`packages/web/src/islands/NavRail.tsx` was not updated. Its `handleNavClick` switch statement has cases for `home`, `chats`, `rooms`, and `settings`, but no `case 'spaces'`. Since `MAIN_NAV_ITEMS` now includes the Spaces entry, the Spaces icon renders in the desktop sidebar, but clicking it silently does nothing — the section switch never executes and the nav signal is never set.

```ts
// NavRail.tsx — handleNavClick is missing:
case 'spaces':
  navigateToSpaces();
  break;
```

`navigateToSpaces` is not imported in `NavRail.tsx` at all. The mobile side (ContextPanel's `handleMobileNavClick`) was correctly updated; the desktop NavRail was not.

---

### P1 — `navigateToSpaces` unconditionally calls `navigateToHome`, clearing current space

`navigateToSpaces()` in `router.ts` (line 523–525) unconditionally calls `navigateToHome()`, which:
1. Sets `currentSpaceIdSignal.value = null`
2. Navigates the URL to `/`

This means if a user is already viewing a space at `/space/<id>` and clicks the Spaces icon in the sidebar, they are kicked to the home page and lose their current space context. The docblock says "navigates home if needed" but the implementation doesn't have that conditional logic.

Compare to `navigateToChats` which only calls `navigateToHome` when currently on a room route. The analogous fix for spaces would be to only navigate home if `currentSpaceIdSignal.value` is null (i.e., not already viewing a space).

The `navigateToSpaces` test (`space-router.test.ts`, line 278) doesn't assert on `currentSpaceIdSignal.value` after the call, so this side effect is untested.

---

### P1 — "Create Space" is completely unreachable from the UI

The "Create Space" button within `SpaceContextPanel` is conditional on the `onCreateSpace` prop:
```tsx
{onCreateSpace && (
  <div ...>
    <button onClick={onCreateSpace}>Create Space</button>
  </div>
)}
```

But in `ContextPanel.tsx` (lines 362–367), `SpaceContextPanel` is rendered without `onCreateSpace`:
```tsx
<SpaceContextPanel
  spaces={[]}
  onSpaceSelect={() => (contextPanelOpenSignal.value = false)}
/>
```

No `onCreateSpace` is passed, so the button never renders. Additionally, the header-level CTA button (lines 322–324) does not include `navSection === 'spaces'`, so the header "Create Space" button is also never shown. `handleAction`'s `'spaces'` case (line 239) calls `navigateToSpaces()` which navigates home rather than opening a create dialog.

Whether "Create Space" is intentionally deferred (as this is a foundational PR) is fine, but the inconsistency — the `sectionConfig.spaces.actionLabel = 'Create Space'` exists, the `handleAction` switch has a `'spaces'` case, the `SpaceContextPanel` has the button in its template, but none of them are wired together — should be cleaned up to avoid confusion. At minimum a code comment clarifying this is intentional would help.

---

### P1 — `spaces={[]}` hardcode in ContextPanel

`ContextPanel.tsx` passes a hardcoded empty array `spaces={[]}` to `SpaceContextPanel`. There is no data fetch, signal subscription, or state hook for spaces. When the Spaces data layer is connected, this will need to be updated. This is expected for a foundational PR, but it should have a `// TODO` comment to prevent it from being forgotten.

---

### P2 — Test gap: `navigateToSpaces` does not assert space signal is cleared

The test for `navigateToSpaces` (line 278–285 in `space-router.test.ts`) only asserts `navSectionSignal.value === 'spaces'`. It does not assert that `currentSpaceIdSignal.value` is null after the call (which it will be, since `navigateToHome` clears it). Adding that assertion would document the behavior explicitly and catch any future regression.

---

### P2 — `navigateToSpaces` behavior inconsistency with `navigateToChats` pattern

`navigateToChats` (line 463) uses a conditional: only navigate home if currently on a room route. `navigateToSpaces` and `navigateToRooms` always navigate home. The docblock for `navigateToSpaces` says "if needed" but the code doesn't match. If the intended behavior is "always go to `/`", the comment should say "always navigates home". If the intent is conditional navigation, the implementation should match `navigateToChats`'s pattern.

---

### P3 — No test for the NavRail `'spaces'` click behavior

Once NavRail is fixed to handle spaces, there should be a test in `packages/web/src/islands/__tests__/NavRail.test.tsx` verifying that clicking the Spaces button calls `navigateToSpaces` and sets the nav section correctly, consistent with the existing Chats/Rooms tests in that file.

---

### Summary

The routing, signals, and path helpers are correctly implemented. The URL regex patterns follow the existing convention, `initializeRouter` and `handlePopState` correctly handle all three space route variants with proper priority ordering, and the 35 unit tests give solid coverage of the path layer. The two main issues are (1) the NavRail was not updated and Spaces is broken on desktop, and (2) `navigateToSpaces` unconditionally navigates home which will eject a user from a space they're already viewing.

lsm (commented):

### Re-review Summary

The original P0 (space signal clearing in existing navigate functions) has been thoroughly fixed — all five navigate functions (`navigateToSession`, `navigateToRoom`, `navigateToRoomSession`, `navigateToRoomTask`, `navigateToSessions`) now clear all three space signals in both the normal path and same-path early-return branches. The P1 (same-path branches not setting `navSectionSignal`) and P3 (`@ts-nocheck` removal) are also resolved. Test count went from 35 to 44, all passing. Good work on those fixes.

However, the **NavRail desktop sidebar issue** flagged in the previous sonnet review is still present and is a blocking bug.

---

### P0 — NavRail `handleNavClick` still missing the `'spaces'` case

**File:** `packages/web/src/islands/NavRail.tsx`, lines 17-33

`navigateToSpaces` is imported (line 6), and `MAIN_NAV_ITEMS` includes the Spaces entry (so the icon renders in the desktop sidebar), but `handleNavClick` has no `case 'spaces':` branch:

```ts
const handleNavClick = (section: NavSection) => {
    switch (section) {
        case 'home': ...
        case 'chats': ...
        case 'rooms': ...
        case 'settings': ...
        // NO 'spaces' case — click does nothing
    }
};
```

Clicking the Spaces icon on desktop silently does nothing. The mobile ContextPanel (`handleMobileNavClick` at line 239) correctly handles `'spaces'`, so only the desktop NavRail is broken. This is a functional bug: the Spaces icon is visible and clickable but non-functional on desktop.

**Fix:** Add `case 'spaces': navigateToSpaces(); break;` to the switch.

---

### P2 — `navigateToSpaces` unconditionally navigates home, ejecting user from current space view

**File:** `packages/web/src/lib/router.ts`, lines 523-526

`navigateToSpaces()` always calls `navigateToHome()`, which navigates to `/` and clears all space ID signals. If a user is viewing `/space/<id>` and clicks "Spaces" in the nav, they lose their current space context and are sent to the home page. Compare to `navigateToChats()` (lines 492-498) which conditionally navigates home only when on a room route.

This is a UX inconsistency rather than a hard bug (the spaces list still shows in the context panel), but it means clicking "Spaces" while already in a space is destructive. Consider adding a conditional like:
```ts
if (!currentSpaceIdSignal.value) {
    navigateToHome();
}
```

The test for `navigateToSpaces` (line 280-288) does not assert on `currentSpaceIdSignal.value` after the call, so this behavior is untested.

---

### P3 — Hardcoded `spaces={[]}` without TODO comment

**File:** `packages/web/src/islands/ContextPanel.tsx`, lines 362-367

`SpaceContextPanel` receives a hardcoded empty array. This is expected for a foundational PR, but a `// TODO: connect to space data source` comment would prevent this from being overlooked when the data layer is wired up.

---

### What was fixed since last review

- **P0 (original):** All existing navigate functions now properly clear space signals — RESOLVED
- **P1 (original):** Same-path early-return branches in `navigateToSpace*` now set `navSectionSignal` — RESOLVED
- **P3 (original):** `@ts-nocheck` removed from test file — RESOLVED
- 9 regression tests added (44 total, all passing)

### What remains unfixed from the previous sonnet review

- **P0 (sonnet):** NavRail missing `'spaces'` case — STILL OPEN (reclassified as P0 above)
- **P1 (sonnet):** `navigateToSpaces` unconditionally clears space context — STILL OPEN (reclassified as P2 above since it's a UX concern, not broken functionality)
- **P1 (sonnet):** `spaces={[]}` hardcode — STILL OPEN (reclassified as P3, acceptable for foundational PR)

lsm (commented):

The previously-flagged P0 and P1 signal-clearing regressions have been resolved, and the test suite is now at 44 tests with `@ts-nocheck` removed. However, both issues flagged explicitly for re-check remain unaddressed.

---

### P0 — NavRail desktop sidebar does not handle 'spaces' (non-functional on desktop)

**File:** `packages/web/src/islands/NavRail.tsx`, lines 17–33

The `handleNavClick` switch statement has cases for `home`, `chats`, `rooms`, and `settings`, but **no case for `spaces`**. There is no `default` branch either. When a user clicks the Spaces button on desktop (md breakpoint and above), the switch falls through without executing any code: no URL change, no signal update, no navigation.

```ts
const handleNavClick = (section: NavSection) => {
    switch (section) {
        case 'home': ...
        case 'chats': ...
        case 'rooms': ...
        case 'settings': ...
        // 'spaces' is missing — click is a no-op on desktop
    }
};
```

The Spaces button is rendered (via `MAIN_NAV_ITEMS` which now includes `'spaces'`) and visually receives an active highlight when `navSectionSignal.value === 'spaces'`, but clicking it on desktop does nothing. The feature is entirely non-functional on desktop.

The fix is to add:
```ts
case 'spaces':
    navigateToSpaces();
    break;
```

The NavRail test file (`packages/web/src/islands/__tests__/NavRail.test.tsx`) also still mocks only three router functions (`navigateToSessions`, `navigateToSettings`, `navigateToHome`) — `navigateToSpaces` and `navigateToRooms` are absent from the mock, so clicking Rooms or Spaces in the existing test suite would throw a module resolution error rather than being covered. There are no tests for clicking the Spaces or Rooms buttons in the NavRail.

---

### P0 — `navigateToSpaces` unconditionally calls `navigateToHome`, wiping space context

**File:** `packages/web/src/lib/router.ts`, lines 523–526

```ts
export function navigateToSpaces(): void {
    navSectionSignal.value = 'spaces';
    navigateToHome();   // always navigates to "/" and clears ALL signals
}
```

`navigateToHome()` unconditionally navigates to `/` and sets `currentSpaceIdSignal.value = null`. This means:

1. A user deep-links to `/space/abc123`. The page loads correctly. They click a session inside the space, which calls `navigateToSpaceSession()`, setting the URL to `/space/abc123/session/xyz`. Now they click the Spaces icon in the nav to go back to the space list — `navigateToSpaces()` fires, clears `currentSpaceIdSignal`, navigates to `/`, and drops them on the home Lobby screen with `navSection = 'spaces'`. The `SpaceContextPanel` shows in the left rail, but `MainContent` falls through to `<Lobby />` because `spaceId` is now null.

2. More subtly: if the user is already on any space-related path (`/space/...`) and clicks Spaces, the URL becomes `/` but `navSectionSignal` is `'spaces'`. `MainContent` does not render `SpaceIsland` for `navSection === 'spaces'` alone — it only renders it when `currentSpaceIdSignal.value` is non-null (line 36 check). So the main area goes blank (Lobby).

Compare to `navigateToChats()` (lines 492–498) which only calls `navigateToHome()` conditionally (`if (currentRoomIdSignal.value)`), and `navigateToRooms()` (lines 504–508) which always navigates home as an intentional design (rooms section shows a list, not a deep-link). For `navigateToSpaces`, the correct behavior mirrors `navigateToChats` — only navigate home if currently on a non-space route.

The existing test for `navigateToSpaces` (lines 279–288 in `space-router.test.ts`) only asserts that `navSectionSignal.value === 'spaces'` after the call, but does **not** assert that `currentSpaceIdSignal` was preserved or cleared. The test passes vacuously regardless of whether the bug exists.

---

### P1 — NavRail test still missing `navigateToSpaces`/`navigateToRooms` in mock

**File:** `packages/web/src/islands/__tests__/NavRail.test.tsx`, lines 15–19

The vi.mock only exports three functions:
```ts
vi.mock('../../lib/router.ts', () => ({
    navigateToSessions: vi.fn(),
    navigateToSettings: vi.fn(),
    navigateToHome: vi.fn(),
}));
```

`navigateToRooms` and `navigateToSpaces` are not mocked. The `NavRail` component imports both of these (lines 6–8 of `NavRail.tsx`). Any test that clicks the Rooms or Spaces buttons would fail with an error, but no such tests exist either. Coverage for the two newest nav items is zero.

---

### P3 — `navigateToRooms` also unconditionally calls `navigateToHome`

**File:** `packages/web/src/lib/router.ts`, lines 504–508

This was pre-existing before this PR. The comment says "Always navigate home when switching to rooms section" which is intentional. Documenting for awareness but not blocking.

---

### Summary

Both issues from the previous review round remain unaddressed. The Spaces nav item is silently broken on desktop (switch case missing in `NavRail`), and `navigateToSpaces` always wipes space context by calling `navigateToHome` unconditionally. The test for `navigateToSpaces` does not catch the second bug because it does not assert on `currentSpaceIdSignal` post-call.

lsm (commented):

The previous round's issues have been addressed. Here is the assessment of the updated code:

### Previous issues -- verified fixed

1. **P0 NavRail spaces case** -- Fixed. `NavRail.tsx` line 29-31 now has `case 'spaces': navigateToSpaces(); break;` and the import on line 7 includes `navigateToSpaces`. Desktop Spaces navigation works.

2. **P1 navigateToSpaces conditional** -- Fixed. `router.ts` lines 523-528: `navigateToSpaces()` sets `navSectionSignal.value = 'spaces'` then only calls `navigateToHome()` when `currentSpaceIdSignal.value` is null. This preserves space context correctly.

3. **P1 TODO comment in ContextPanel.tsx** -- Fixed. Lines 362-371 include a clear TODO comment explaining the `spaces={[]}` placeholder pending the data layer.

4. **P2 navigateToSpaces test coverage** -- Fixed. `space-router.test.ts` lines 279-301 now has two test cases: one for preserving space context when already viewing a space, and one for navigating home when not on a space.

5. **P3 NavRail test coverage** -- Fixed. `NavRail.test.tsx` now includes tests for Rooms button click (line 133-139), Spaces button click (line 142-149), and Spaces active-state highlighting (line 151-160).

### New issue found

**P2: `navigateToSpace*` functions do not clear `currentRoomSessionIdSignal` or `currentRoomTaskIdSignal`**

In `packages/web/src/lib/router.ts`, all three space navigation functions (`navigateToSpace`, `navigateToSpaceSession`, `navigateToSpaceTask`) clear `currentRoomIdSignal` and `currentSessionIdSignal` but leave `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal` untouched. Compare with existing functions like `navigateToSession` (lines 214-217) and `navigateToRoom` (lines 304-310) which clear all space signals plus room sub-signals.

This is not a P0 because `MainContent.tsx` guards the room branch on `roomId` being truthy, so stale room sub-signals won't cause visible routing errors. However, any code that checks `currentRoomSessionIdSignal` or `currentRoomTaskIdSignal` independently (e.g., a future sidebar indicator) would see stale values after navigating to a space from a room session/task.

Both the early-return (same-path) and the normal navigation branches of all three `navigateToSpace*` functions have this omission.

### Overall assessment

The implementation is solid and well-tested. The routing, signals, path extraction, popstate handling, deep-linking, and component rendering all follow existing patterns closely. Test coverage is thorough with 542 lines in `space-router.test.ts` covering path extraction, path creation, navigation, deep-linking, popstate, and cross-section signal clearing regressions. The NavRail and ContextPanel changes are minimal and well-integrated.

**Recommendation: APPROVE** -- The remaining P2 issue (incomplete room sub-signal clearing) is a latent inconsistency rather than a functional bug, and can be addressed as a follow-up.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — two issues (one P1, one P3) need attention before this is clean.

---

### What was checked

All three commits were reviewed end-to-end:
- `packages/web/src/lib/router.ts` — space routing, signal clearing, popstate, `initializeRouter`
- `packages/web/src/lib/signals.ts` — three new space signals
- `packages/web/src/islands/NavRail.tsx` — `spaces` case and import
- `packages/web/src/islands/ContextPanel.tsx` — `SpaceContextPanel` wiring, `TODO` comment
- `packages/web/src/islands/MainContent.tsx` — `SpaceIsland` render gate
- `packages/web/src/components/space/SpaceContextPanel.tsx` — filter tabs, space list
- `packages/web/src/islands/SpaceIsland.tsx` — placeholder component
- `packages/web/src/lib/nav-config.tsx` — Spaces nav item
- `packages/web/src/lib/__tests__/space-router.test.ts` — 44 tests
- `packages/web/src/islands/__tests__/NavRail.test.tsx` — new click/highlight tests

All 3856 web tests pass. TypeScript clean.

---

### P1 — `@ts-nocheck` was NOT removed from `NavRail.test.tsx`

The third commit message says:

> P3: removed `@ts-nocheck` from test file; Object.defineProperty calls are valid in jsdom so no `ts-expect-error` suppression is needed.

But line 1 of `packages/web/src/islands/__tests__/NavRail.test.tsx` still reads:

```ts
// @ts-nocheck
```

The suppression was described as removed but the file was not updated. This silences all type errors in the test file and contradicts what the commit log says was done. Since the commit log claims this was addressed, the discrepancy itself is a correctness problem — whoever reads the history will believe `@ts-nocheck` was cleaned up when it wasn't.

---

### P2 — `navigateToChats` does not clear space signals

`navigateToChats()` (router.ts lines 492–498) only calls `navigateToHome()` when `currentRoomIdSignal.value` is truthy. If the user is on a space URL, `currentRoomIdSignal.value` is null (cleared by `navigateToSpace`), so `navigateToHome()` is never called, and `currentSpaceIdSignal.value` is left set. That means `MainContent` would continue rendering `SpaceIsland` even after `navSectionSignal` flips to `'chats'`.

In practice this bug is not triggered by the current UI today because no component calls `navigateToChats()` — both `NavRail` and `ContextPanel` use `navigateToSessions()` in their `'chats'` switch cases, which does clear space signals correctly. But `navigateToChats` is a public export and the inconsistency is a trap for the next caller.

The analogous fix pattern used for `navigateToRooms` and `navigateToSettings` (which always call `navigateToHome()`) should be applied here: add a guard for `currentSpaceIdSignal.value` in `navigateToChats`, or simplify it to always call `navigateToHome()` consistent with the other section navigators.

There are no regression tests covering this function at all — `navigateToChats` does not appear anywhere in `space-router.test.ts`.

---

### Confirmed fixes from the previous round

- **P0 fixed**: `NavRail.tsx` `case 'spaces': navigateToSpaces(); break;` is present and `navigateToSpaces` is imported.
- **P1 fixed**: `navigateToSpaces` only calls `navigateToHome()` when `currentSpaceIdSignal.value` is null, preserving space context for users already on a space URL. Tested with two explicit cases.
- **P1 fixed**: Same-path early-return branches in `navigateToSpace`, `navigateToSpaceSession`, `navigateToSpaceTask` all set `navSectionSignal.value = 'spaces'`.
- **P1 fixed**: `TODO` comment in `ContextPanel.tsx` explains the `spaces={[]}` placeholder and defers `onCreateSpace`.
- **P0 fixed**: All existing navigate functions (`navigateToSession`, `navigateToRoom`, `navigateToRoomSession`, `navigateToRoomTask`, `navigateToSessions`) clear all three space signals in both the same-path and main branches.
- **P3 noted but not done**: NavRail test file still has `@ts-nocheck` (see P1 above).

---

- `SpaceContextPanel` renders `currentSpaceIdSignal.value` directly (not from props), which is correct — the signal is the source of truth.
- The `isActionDisabled` check in `ContextPanel` does not include `'spaces'`. The "Create Space" header button is not rendered for spaces section (the `navSection === 'spaces'` condition is absent from the button show-condition at line 322–324), so this is consistent — the button simply never appears. No UX regression.
- `SpaceIsland` is intentionally a placeholder; the `key={spaceId}` prop in `MainContent` ensures remount on space switch. Correct.
- Route patterns are strict hex-UUID format (`[a-f0-9-]+`) matching the existing room/session patterns. Consistent.

lsm (commented):

### Re-review Summary

All previously flagged issues have been addressed:

**P1 fix: `@ts-nocheck` removed** -- Confirmed that `@ts-nocheck` has been removed from both `packages/web/src/islands/__tests__/NavRail.test.tsx` and `packages/web/src/lib/__tests__/space-router.test.ts`. The tests now run with full type checking. The NavRail test also properly mocks `navigateToRooms` and `navigateToSpaces`, and includes new tests for Rooms/Spaces button clicks and Spaces active state highlighting.

**P2 fix: `navigateToChats` space signal clearing** -- The guard at `packages/web/src/lib/router.ts:495` now checks `currentRoomIdSignal.value || currentSpaceIdSignal.value`, so calling `navigateToChats()` while viewing a space correctly calls `navigateToHome()` which clears all space signals. Two regression tests verify this behavior (lines 549-573 of `space-router.test.ts`).

1. **Unused export**: `navigateToChats()` is exported from `router.ts` but not consumed by any component yet -- only exercised in tests. This is fine for a foundation PR that builds out the routing API ahead of UI integration.

2. **Signal clearing repetition**: Every navigate function manually clears 7 signals in both the early-return branch and the main branch. This is quite verbose but mechanically correct, and a refactor to a shared `clearAllSignals()` helper can be done later without risk.

3. **`SpaceContextPanel` receives an empty array**: In `ContextPanel.tsx:368`, `spaces={[]}` is hardcoded with a clear TODO comment. This is expected since the Space data layer is not yet wired up.

The implementation is correct, well-tested (47 tests in `space-router.test.ts`, 23 in `NavRail.test.tsx`), and all signal-clearing paths are consistent. Recommending **APPROVE**.

lsm (commented):

**Recommendation: APPROVE**

Both previously-flagged issues have been correctly resolved. Verification details below.

---

### P1 resolved: `@ts-nocheck` removed from `NavRail.test.tsx`

Confirmed: line 1 of `packages/web/src/islands/__tests__/NavRail.test.tsx` no longer contains the `@ts-nocheck` directive. The fix commit (`3981f56b`) shows a clean `-1` line diff removing it. The file now opens directly with the JSDoc block comment, and grep returns no matches. TypeScript will enforce types in this test file going forward.

---

### P2 resolved: `navigateToChats()` now guards on both room and space context

The guard in `packages/web/src/lib/router.ts` at line 495 is now:

```ts
if (currentRoomIdSignal.value || currentSpaceIdSignal.value) {
    navigateToHome();
}
```

This correctly extends the existing room-route guard to also cover the space-route case. `navigateToHome()` nulls out all space signals (`currentSpaceIdSignal`, `currentSpaceSessionIdSignal`, `currentSpaceTaskIdSignal`) as well as room and session signals, so the clearing is complete and consistent.

Two regression tests were added in `packages/web/src/lib/__tests__/space-router.test.ts` (lines 549-573):

1. **`navigates home and clears space signals when on a space URL`** — sets up a space navigation state, calls `navigateToChats()`, and asserts all three space signals are null and `navSectionSignal` is `'chats'`. This directly exercises the new guard branch.
2. **`does not navigate home when on neither a room nor a space route`** — verifies no spurious `pushState` is called when already on `/` with no active context. This is a valid guard against regression in the no-op path.

Both tests are well-scoped and cover exactly the behavior the fix introduces. No concerns.

lsm (commented):

### Recommendation: APPROVE

This PR implements two main pieces of work: (1) Space navigation, routing, and signals (Task 5.1), and (2) removal of hardcoded agent roles from the Space implementation. Both are well-executed.

---

### Navigation, Routing, and Signals (Task 5.1)

The routing layer is cleanly extended with space URL patterns (`/space/:id`, `/space/:id/session/:id`, `/space/:id/task/:id`). Key observations:

- **Signal management is consistent**: Every `navigate*` function clears orthogonal signals (space signals when navigating to room/session, room signals when navigating to space, etc.). This is critical for preventing stale state and is done correctly across all 10+ navigation functions.

- **`initializeRouter` and `handlePopState`** both handle space routes with proper priority ordering (space > room > session), matching the pattern established for rooms.

- **NavRail and ContextPanel** are properly wired to the `'spaces'` nav section with `navigateToSpaces()`.

- **SpaceIsland and SpaceContextPanel** are placeholder implementations that correctly render based on signals — appropriate for a foundation task.

- **Tests are thorough**: `space-router.test.ts` covers path extraction, path creation, navigation functions, deep-linking via `initializeRouter`, and popstate handling. The NavRail test verifies the Spaces button integration.

### Hardcoded Agent Role Removal

The changes are clean and complete:

- **`space.ts` types**: `BuiltinAgentRole` union removed; `SpaceAgent.role`, `CreateSpaceAgentParams.role`, and `UpdateSpaceAgentParams.role` are now `string`. Documentation clearly states role is a "free-form display label."

- **`space-agent-handlers.ts`**: `VALID_ROLES` and `assertValidRole()` completely removed. The handler now accepts any non-empty `role` string.

- **`custom-agent.ts`**: Reviewer-specific system prompt injection removed. `getRoleLabel()` is now a generic title-case function that works with any string. No role-based behavioral branching remains — all roles get the same git workflow, bypass markers, and review feedback instructions.

- **Migration 31**: Correctly rebuilds `space_agents` table without the `CHECK(role IN (...))` constraint. The migration is properly idempotent (checks `schema.sql` for the CHECK pattern before rebuilding). One minor observation: it does not wrap in `PRAGMA foreign_keys = OFF/ON` — however, since `space_agents` has an outbound FK to `spaces` (not inbound), and the migration inserts from the existing table with valid `space_id` values, there is no practical risk.

- **`space-agent-schema.ts` test helper**: Updated to match the new unconstrained schema.

- **Tests updated**: Role validation tests replaced with "accepts any string" tests. Reviewer-specific prompt tests replaced with "no role produces role-specific instructions" assertions.

### Minor Observations (P3)

1. **Stale comments in `space-agent-repository.ts` (line 10)** and **`migrations.ts` (line 1682)**: Both still reference `BuiltinAgentRole` in doc comments. These are cosmetic — they describe the historical migration 30 behavior, not current code — but could be confusing to future readers.

2. **`navigateToChats`** clears space signals via `navigateToHome()` but does not explicitly clear `currentRoomTaskIdSignal` when there is no room/space active. This is consistent with the existing pattern (same behavior pre-PR) so not a regression.

Overall, this is a well-structured implementation that follows established patterns and has good test coverage.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 bug, two P2 issues, and one P3 stale comment.

---

## Summary

The refactoring correctly achieves its stated goal: `BuiltinAgentRole` is fully removed from the shared type surface, all hardcoded role-based branching (reviewer system prompt injection, `taskType==='review'` guard) is gone from `custom-agent.ts`, and migration 31 correctly rebuilds the `space_agents` table without the `CHECK` constraint using the standard SQLite table-rebuild pattern. The space navigation/routing/signals work from the main commit is also solid — route patterns, signal clearing, popstate handling, and initialization logic are correct and well-tested.

Three issues need attention before merge.

---

## Issues

### P1 — Bug: `navigateToSessions` does not clear `currentRoomTaskIdSignal`

**File:** `packages/web/src/lib/router.ts`, lines 451–486

Every other navigation function that leaves a room clears `currentRoomTaskIdSignal`. `navigateToSessions` is the only function that omits it — in both the same-path early-return branch (lines 458–465) and the main push branch (lines 474–480).

If a user is viewing a room task detail (`/room/:id/task/:tid`) and clicks the Chats nav item, `currentRoomTaskIdSignal` is left non-null. When the user later navigates to a room, `MainContent` renders `<Room>` with a stale `taskViewId` prop, showing a task detail from a previous, unrelated room session.

The fix is to add `currentRoomTaskIdSignal.value = null;` to each of the two branches in `navigateToSessions`. The test file `packages/web/src/lib/__tests__/space-router.test.ts` does not import `currentRoomTaskIdSignal` at all, so this bug is completely undetected by the existing test suite. The `navigateToSessions clears space signals` test block should be extended to also assert that `currentRoomTaskIdSignal` is cleared.

---

### P2 — Stale doc comment in `space-agent-repository.ts`

**File:** `packages/daemon/src/storage/repositories/space-agent-repository.ts`, line 10

The column-mapping comment still reads:

```
SpaceAgent.role  ↔  role column (BuiltinAgentRole: 'planner'|'coder'|'general'|'reviewer')
```

`BuiltinAgentRole` was removed in this PR. The comment should be updated to match the new reality that `role` is a free-form display label string.

---

### P2 — Migration 31 table-rebuild is not wrapped in a transaction

**File:** `packages/daemon/src/storage/schema/migrations.ts`, `runMigration31`

The rebuild sequence (CREATE TABLE → INSERT → DROP TABLE → RENAME) is four separate `db.exec()` calls with no explicit `BEGIN`/`COMMIT`. If the process is killed between the `DROP TABLE` and `RENAME TO` steps, the `space_agents` table is permanently destroyed and unrecoverable without the backup.

This pattern is shared by other migrations in the file (so it is not a regression introduced here), but this migration is higher-risk because it operates on a table that may contain live user-defined agent data. A `db.transaction(() => { ... })()` wrapper would fully eliminate this window.

Additionally, no dedicated test exists for migration 31 — `packages/daemon/tests/unit/storage/migrations/` only covers up to migration 29. The idempotency guard (string-searching `sqlite_master.sql` for `'CHECK(role IN'`) is easy to get wrong and the fresh-DB skip path is critical. A test similar to the existing `migration-29_test.ts` would give confidence that both the constraint-present path and the already-migrated path are covered.

---

### P3 — Stale comment in migration 30 doc comment

**File:** `packages/daemon/src/storage/schema/migrations.ts`, line 1682

The migration 30 JSDoc still reads:

```
 * - role: builtin agent role preset ('planner'|'coder'|'general'|'reviewer'), matching BuiltinAgentRole
```

`BuiltinAgentRole` no longer exists. Harmless but worth updating for maintainability.

---

## What is correct

- `BuiltinAgentRole` union type fully removed from `packages/shared/src/types/space.ts`; all three role fields (`SpaceAgent.role`, `CreateSpaceAgentParams.role`, `UpdateSpaceAgentParams.role`) are now `string`
- `VALID_ROLES` Set and `assertValidRole()` fully removed from `space-agent-handlers.ts`; the handler now accepts any non-empty string
- `buildCustomAgentSystemPrompt` no longer branches on role to inject reviewer-specific content; all roles get identical system prompt structure
- `buildCustomAgentTaskMessage` no longer branches on `taskType === 'review'` or `role === 'reviewer'`
- `getRoleLabel()` is correctly generalized to title-case any non-empty string, with a `'Custom'` fallback for empty/falsy input
- Migration 31 handles the three expected DB states correctly: fresh DB (no table → skip), already-migrated DB (no CHECK → skip), and legacy DB with the constraint (rebuild)
- The migration 31 detection string `'CHECK(role IN'` correctly matches what migration 30 writes and cannot false-positive on fresh databases
- All test files accurately reflect the new behavior: `accepts any string as role value` in `space-agent-handlers.test.ts`, `no role produces role-specific instructions` and `does not include review instructions for any role or task type` in `custom-agent.test.ts`
- Space routing signals are correctly cleared across all navigate functions, with the sole exception of the `currentRoomTaskIdSignal` omission in `navigateToSessions` noted above

lsm (commented):

All previously raised issues have been addressed satisfactorily. Detailed review of the fixes:

### P1 Fix: `navigateToSessions` clears `currentRoomTaskIdSignal` — Verified

Both branches of `navigateToSessions` in `packages/web/src/lib/router.ts` (lines 462 and 478) now clear `currentRoomTaskIdSignal`. The test in `packages/web/src/lib/__tests__/space-router.test.ts` (lines 480-507) asserts this in both the navigation branch and the same-path early-return branch. Correct.

### P2 Fix: Migration 31 wrapped in `db.transaction()` — Verified

`packages/daemon/src/storage/schema/migrations.ts` (lines 1734-1766) wraps the table-rebuild (CREATE, INSERT, DROP, RENAME, CREATE INDEX) in a single transaction. This ensures atomicity — a crash mid-rebuild cannot leave the database without the `space_agents` table.

The approach is sound: `space_agents` is a child table (references `spaces`) and no other table references it, so the DROP is safe with `foreign_keys = ON`. This is actually a cleaner pattern than the `PRAGMA foreign_keys = OFF` + `try/finally` approach used by older migrations.

### P2 Fix: Migration 31 test coverage — Verified

`packages/daemon/tests/unit/storage/migrations/migration-31_test.ts` provides 6 tests covering:
1. Fresh DB accepts any string role value after full migration chain
2. Fresh DB schema does not contain CHECK constraint
3. Fresh DB index exists after migration
4. Legacy DB with CHECK constraint is rebuilt without it (validates constraint *is* enforced before migration, then confirms removal)
5. Legacy DB preserves existing agent rows after rebuild
6. Idempotency — running migrations twice does not error or duplicate data

All tests properly set up and tear down with unique temp directories and close DB handles.

### P2 Fix: Stale comment in `space-agent-repository.ts` — Verified

`packages/daemon/src/storage/repositories/space-agent-repository.ts` line 10 now reads: `role column (free-form string display label, e.g. 'coder', 'planner')` — no mention of `BuiltinAgentRole`.

### P3 Fix: Migration 30 doc comment — Verified

`packages/daemon/src/storage/schema/migrations.ts` lines 1681-1684 now correctly documents that the role column was "added with a CHECK constraint that was later removed by migration 31".

- `navigateToSession` does not clear `currentRoomIdSignal`, `currentRoomSessionIdSignal`, or `currentRoomTaskIdSignal`. This is a pre-existing pattern that predates this PR and is consistent with the existing router.test.ts expectations. Not blocking for this PR but worth noting for a future cleanup.

**Recommendation: APPROVE** — All previously raised P1/P2/P3 issues have been correctly resolved. The migration is atomic, well-tested, and idempotent. The router signal clearing is now consistent.

lsm (commented):

**Recommendation: APPROVE**

All previously flagged issues have been resolved. Detailed findings below.

---

### Previously Flagged Issues — All Resolved

**P1 (was: navigateToSessions not clearing currentRoomTaskIdSignal in both branches)**

Fixed in both branches of `navigateToSessions` (lines 457-488 of `router.ts`):
- Early-return branch (same path): now clears `currentRoomTaskIdSignal`
- Normal branch: now clears `currentRoomTaskIdSignal`

Tests extended in `space-router.test.ts` (lines 495-506) to explicitly assert `currentRoomTaskIdSignal.value` is null in the same-path branch.

**P2 (was: Migration 31 table-rebuild not wrapped in a transaction)**

The table-rebuild in `runMigration31` is now wrapped in `db.transaction(() => { ... })()` (line 1734 of `migrations.ts`). A crash mid-rebuild cannot leave the database without a `space_agents` table.

Note: migration 31 does not need to disable `foreign_keys` before dropping the table, because no other table has an FK constraint pointing to `space_agents`. The `agent_id` column in `space_workflow_steps` and `custom_agent_id` in `space_tasks` are bare TEXT columns with no FK declaration. The omission of `PRAGMA foreign_keys = OFF` is intentional and correct here.

**P2 (was: missing migration-31 tests)**

`migration-31_test.ts` now has 6 tests:
1. Fresh DB: role column accepts any string after full migration chain
2. Fresh DB: schema does not contain the old CHECK constraint
3. Fresh DB: index exists after migration
4. Legacy DB: table with CHECK constraint is rebuilt without it
5. Legacy DB: existing agent rows are preserved after rebuild
6. Idempotency: running migration twice does not error or duplicate data

**P2 (was: stale BuiltinAgentRole comment in space-agent-repository.ts)**

Comment updated to correctly describe `role` as a free-form string display label.

**P3 (was: migration 30 doc comment still referenced BuiltinAgentRole)**

Doc comment updated at line 1681 of `migrations.ts`.

---

### No New Issues Found

The implementation is clean. The `navigateToSessions` fix is symmetric with all other navigate functions. The migration 31 transaction boundary is consistent with the established pattern in this codebase (migrations 9, 13, 15, etc.). The 6 migration tests cover fresh DB, legacy DB with constraint, data preservation, and idempotency — exactly what was asked for and what the code needs. The `SpaceAgent.role` type in `space.ts` is now correctly typed as `string` rather than a restricted union.

PR #440, opened 2026-03-19
feat: built-in workflow templates and seedBuiltInWorkflows (Task 3.4)

lsm:
## Summary

- Three built-in workflow templates: `CODING_WORKFLOW` (Planner→Coder, human_approval + pr_review gates), `RESEARCH_WORKFLOW` (Planner→General, auto gates), `REVIEW_ONLY_WORKFLOW` (Coder only, pr_review gate)
- `getBuiltInWorkflows()` returns all three templates with role-name `agentId` placeholders
- `seedBuiltInWorkflows(spaceId, workflowManager, resolveAgentId)` idempotently seeds all three — throws if a role cannot be resolved to a real SpaceAgent UUID
- Exported from `packages/daemon/src/lib/space/index.ts`
- NOT wired to a call site (that's Task 4.2)

## Test plan

- [ ] 40 unit tests pass: template structure, valid agentId placeholders, no 'leader' refs, all three templates seeded, gate types correct, real agent IDs substituted, idempotency, error on unresolved role

### DIFF
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -16,7 +16,7 @@ export {
 	RESEARCH_WORKFLOW,
 	REVIEW_ONLY_WORKFLOW,
 	getBuiltInWorkflows,
-	seedDefaultWorkflow,
+	seedBuiltInWorkflows,
 } from './workflows/built-in-workflows';
 
 // Types — re-exported from @neokai/shared for convenience
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -11,6 +11,13 @@ export {
 export { SpaceWorkflowManager, WorkflowValidationError } from './managers/space-workflow-manager';
 export type { SpaceAgentLookup } from './managers/space-workflow-manager';
 export { SpaceWorkflowRepository } from '../../storage/repositories/space-workflow-repository';
+export {
+	CODING_WORKFLOW,
+	RESEARCH_WORKFLOW,
+	REVIEW_ONLY_WORKFLOW,
+	getBuiltInWorkflows,
+	seedDefaultWorkflow,
+} from './workflows/built-in-workflows';
 
 // Types — re-exported from @neokai/shared for convenience
 export type {
--- a/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
+++ b/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
@@ -6,45 +6,16 @@
  *
  * Design notes:
  * - Leader is always implicit in SpaceRuntime — never a workflow step.
- * - Templates use placeholder `id` / `spaceId` (empty strings); the seeding
- *   utility stamps in the real spaceId at creation time.
- * - Steps carry stable template-scoped IDs so that the in-memory template
- *   objects satisfy the `SpaceWorkflow` shape.
+ * - Templates use placeholder `id` / `spaceId` (empty strings) and role names
+ *   as `agentId` placeholders ('planner', 'coder', 'general'). These are
+ *   replaced with real SpaceAgent UUIDs by `seedBuiltInWorkflows`.
+ * - At Space creation time, preset SpaceAgent records are seeded for each
+ *   BuiltinAgentRole. `seedBuiltInWorkflows` must be called after those agents
+ *   exist so that the `agentId` values resolve correctly.
  */
 
-import type { SpaceWorkflow, WorkflowStepInput } from '@neokai/shared';
+import type { BuiltinAgentRole, SpaceWorkflow, WorkflowStepInput } from '@neokai/shared';
 import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
-
-// ---------------------------------------------------------------------------
-// Template helpers
-// ---------------------------------------------------------------------------
-
-/**
- * Converts SpaceWorkflow steps to WorkflowStepInput[] (drops id/order)
- * so they can be passed to SpaceWorkflowManager.createWorkflow().
- */
-function stepsToInputs(steps: SpaceWorkflow['steps']): WorkflowStepInput[] {
-	return steps.map((s): WorkflowStepInput => {
-		if (s.agentRefType === 'custom') {
-			return {
-				name: s.name,
-				agentRefType: 'custom',
-				agentRef: s.agentRef,
-				entryGate: s.entryGate,
-				exitGate: s.exitGate,
-				instructions: s.instructions,
-			};
-		}
-		return {
-			name: s.name,
-			agentRefType: 'builtin',
-			agentRef: s.agentRef,
-			entryGate: s.entryGate,
-			exitGate: s.exitGate,
-			instructions: s.instructions,
-		};
-	});
-}
 
 // ---------------------------------------------------------------------------
 // Built-in templates
@@ -57,7 +28,8 @@ function stepsToInputs(steps: SpaceWorkflow['steps']): WorkflowStepInput[] {
  * - Planner produces a plan; human must approve before coding begins.
  * - Coder implements the plan; a PR review is required before the run completes.
  *
- * Leader is implicit per group — not modelled as a step.
+ * Steps use role names as `agentId` placeholders. Call `seedBuiltInWorkflows`
+ * to persist the workflow with real SpaceAgent IDs.
  */
 export const CODING_WORKFLOW: SpaceWorkflow = {
 	id: '',
@@ -69,8 +41,7 @@ export const CODING_WORKFLOW: SpaceWorkflow = {
 		{
 			id: 'tpl-coding-planner',
 			name: 'Plan',
-			agentRefType: 'builtin',
-			agentRef: 'planner',
+			agentId: 'planner',
 			exitGate: {
 				type: 'human_approval',
 				description: 'Review and approve the plan before coding begins',
@@ -80,8 +51,7 @@ export const CODING_WORKFLOW: SpaceWorkflow = {
 		{
 			id: 'tpl-coding-coder',
 			name: 'Code',
-			agentRefType: 'builtin',
-			agentRef: 'coder',
+			agentId: 'coder',
 			exitGate: {
 				type: 'pr_review',
 				description: 'Wait for PR review and approval before completing',
@@ -112,8 +82,7 @@ export const RESEARCH_WORKFLOW: SpaceWorkflow = {
 		{
 			id: 'tpl-research-planner',
 			name: 'Plan Research',
-			agentRefType: 'builtin',
-			agentRef: 'planner',
+			agentId: 'planner',
 			exitGate: {
 				type: 'auto',
 				description: 'Automatically advance after planning is complete',
@@ -123,8 +92,7 @@ export const RESEARCH_WORKFLOW: SpaceWorkflow = {
 		{
 			id: 'tpl-research-general',
 			name: 'Research',
-			agentRefType: 'builtin',
-			agentRef: 'general',
+			agentId: 'general',
 			exitGate: {
 				type: 'auto',
 				description: 'Automatically complete after research is done',
@@ -155,8 +123,7 @@ export const REVIEW_ONLY_WORKFLOW: SpaceWorkflow = {
 		{
 			id: 'tpl-review-coder',
 			name: 'Code',
-			agentRefType: 'builtin',
-			agentRef: 'coder',
+			agentId: 'coder',
 			exitGate: {
 				type: 'pr_review',
 				description: 'Wait for PR review and approval before completing',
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured implementation of Task 3.4. All acceptance criteria are met.

### What was verified

1. **Three built-in workflow templates** -- `CODING_WORKFLOW` (Planner -> Coder), `RESEARCH_WORKFLOW` (Planner -> General), `REVIEW_ONLY_WORKFLOW` (Coder only) are correctly defined with appropriate gate types.

2. **Leader is implicit** -- No template references `'leader'` as an agent ref. This is enforced both by the TypeScript type system (`BuiltinAgentRole = 'planner' | 'coder' | 'general'`) and by test assertions.

3. **Gate types are valid** -- `human_approval`, `pr_review`, and `auto` are all members of `WorkflowGateType`.

4. **`seedDefaultWorkflow` is implemented but NOT wired** -- Confirmed no references in `packages/daemon/src/lib/rpc-handlers/`. The NOTE comment on line 196 explicitly documents this is deferred to Task 4.2.

5. **File location** -- All new files are in `packages/daemon/src/lib/space/workflows/`, not `room/workflows/`.

6. **Tests** -- 36 tests pass covering template structure, agent ref validation, gate types, ascending order, placeholder IDs, idempotent seeding, and user-created workflow preservation. Coverage is thorough.

7. **Exports** -- All public symbols are re-exported from `packages/daemon/src/lib/space/index.ts`.

### Minor observations (P3)

- **`getBuiltInWorkflows()` returns shared object references** (line 185 of `built-in-workflows.ts`): If a caller mutates the returned template objects, the in-memory templates would be corrupted for subsequent calls. Since these templates have `id: ''` and `spaceId: ''` and are only used as read-only inputs to `seedDefaultWorkflow` / `stepsToInputs`, this is not a practical issue today, but freezing the objects or returning deep copies would be more defensive.

- **`seedDefaultWorkflow` is `async` but calls only synchronous methods** (`listWorkflows`, `createWorkflow`): This is fine for forward compatibility if the manager API becomes async later, but is technically unnecessary today.

Both are non-blocking observations -- the implementation is correct and complete.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

The implementation is clean and correct. All three acceptance criteria around template shape, no-leader constraint, idempotent seeding, and correct file placement are satisfied. The 36 tests pass, typecheck is clean, and the linter reports zero issues.

---

### Correctness

**Templates match the spec exactly:**

- `CODING_WORKFLOW`: planner (human_approval exit) → coder (pr_review exit) ✓
- `RESEARCH_WORKFLOW`: planner (auto exit) → general (auto exit) ✓
- `REVIEW_ONLY_WORKFLOW`: coder only (pr_review exit) ✓
- All three templates have `id: ''` and `spaceId: ''` to signal template-status ✓
- No `leader` agentRef in any template — consistent with `BuiltinAgentRole = 'planner' | 'coder' | 'general'` ✓
- Step `order` values are 0-indexed and sequential ✓

**seedDefaultWorkflow:**

- Correctly calls `listWorkflows(spaceId)` before creating, making it idempotent ✓
- Passes through `stepsToInputs()` which drops `id` and `order` so `createWorkflow` can assign them via the repository ✓
- Tags are spread-copied (`[...CODING_WORKFLOW.tags]`) — no accidental shared reference ✓
- Correctly NOT wired to any RPC handler, per task spec (noted in JSDoc comment) ✓

**File placement:**

All new files are under `packages/daemon/src/lib/space/workflows/` and `packages/daemon/tests/unit/space/` — not under `room/` ✓

**Exports:**

`packages/daemon/src/lib/space/index.ts` correctly re-exports all five public symbols ✓

---

### Test Coverage

36 tests across 4 describe blocks:
- Template structure (gate types, agent refs, order, empty id/spaceId)
- No-leader invariant (checked per-template and collectively)
- `getBuiltInWorkflows()` (count, presence, valid roles)
- `seedDefaultWorkflow()` with a real SQLite DB — including the two critical idempotency scenarios (double-seed and pre-existing user workflow)

The test file creates a fresh DB per test via `makeDb()` + `beforeEach`/`afterEach` cleanup — no cross-test pollution. The `seedSpace` helper correctly inserts into the `spaces` foreign-key parent table before workflow operations.

---

**P3 — `stepsToInputs` duplicates logic from `SpaceWorkflowManager.updateWorkflow`:**
Both functions convert a `WorkflowStep[]` (with id/order) to `WorkflowStepInput[]` (without id/order) using the same if/else discriminant branch. This is a minor maintenance burden if new agentRefType values are added. Not a bug, but worth noting as technical debt for a future refactor.

**P3 — `seedDefaultWorkflow` is declared `async` but never actually awaits anything:**
The function is synchronous in its body (`listWorkflows` and `createWorkflow` are both synchronous). Making it `async` is harmless and consistent with how call sites will use it (Task 4.2 will likely `await` it), so this is an intentional forward-looking signature. Noted for awareness.

**P3 — Template objects are mutable module-level `const`s:**
`CODING_WORKFLOW`, `RESEARCH_WORKFLOW`, and `REVIEW_ONLY_WORKFLOW` are exported mutable objects. A caller could mutate `CODING_WORKFLOW.steps.push(...)` and corrupt the template for all subsequent calls. `getBuiltInWorkflows()` returns the objects by reference as well. In practice this is unlikely given the usage pattern, but shallow-freezing the templates or returning copies from `getBuiltInWorkflows()` would make the invariant explicit. Low risk for internal code.

---

### Verdict

No P0/P1/P2 issues found. The implementation is complete, correctly structured, and well-tested. Ready to merge.

lsm (commented):

### Recommendation: APPROVE

The implementation correctly meets all acceptance criteria for Task 3.4. All 38 unit tests pass.

### Verified Acceptance Criteria

1. **Three built-in workflow templates** -- CODING_WORKFLOW, RESEARCH_WORKFLOW, REVIEW_ONLY_WORKFLOW are all defined with flat `agentId: string` fields.
2. **Leader is implicit** -- No `'leader'` string appears anywhere in the templates or seeder.
3. **`seedBuiltInWorkflows` implemented but NOT wired** -- Confirmed zero references in `packages/daemon/src/lib/rpc-handlers/`.
4. **Correct file location** -- All code is in `packages/daemon/src/lib/space/workflows/`, not `room/workflows/`.
5. **Function name matches design doc** -- `seedBuiltInWorkflows` (not the old `seedDefaultWorkflow`).
6. **Seeds all three templates** -- `getBuiltInWorkflows()` returns all three, and the seeder iterates over them.
7. **`resolveAgentId` role resolver** -- Works correctly, with fallback to placeholder string if resolution fails.
8. **Gate types** -- Uses `human_approval`, `pr_review`, and `auto` -- all valid `WorkflowGateType` values.
9. **Idempotency** -- Tested: no-op when workflows already exist.
10. **Exports** -- All public API is re-exported from `packages/daemon/src/lib/space/index.ts`.

### Minor Observations (P3)

1. **PR title is stale** (`built-in-workflows.ts:L178`): The PR title says "seedDefaultWorkflow" but the function is `seedBuiltInWorkflows`. This is cosmetic -- the code is correct.

2. **`seedBuiltInWorkflows` is unnecessarily `async`** (`built-in-workflows.ts:L178`): Both `listWorkflows` and `createWorkflow` on `SpaceWorkflowManager` are synchronous. Making the seeder `async` is harmless and arguably forward-compatible, but it is technically unnecessary today.

3. **Template objects are mutable** (`built-in-workflows.ts:L34-L138`): `CODING_WORKFLOW`, `RESEARCH_WORKFLOW`, and `REVIEW_ONLY_WORKFLOW` are exported `const` references to mutable objects. A caller could accidentally mutate `CODING_WORKFLOW.steps.push(...)` and corrupt the template for all subsequent reads. Consider `Object.freeze` or deep-freezing the templates. Low risk since the only consumer is `seedBuiltInWorkflows` which maps steps into new objects.

4. **`getBuiltInWorkflows` returns the same references** (`built-in-workflows.ts:L152-L154`): Returns the original template objects rather than shallow copies. Same concern as above -- low risk in practice.

Overall the implementation is clean, well-documented, correctly typed, and thoroughly tested. No blocking issues.

lsm (commented):

**Recommendation: APPROVE** (with notes below)

---

### Summary

All 38 tests pass, typecheck is clean, linter is clean, and `built-in-workflows.ts` has 100% line and function coverage. The implementation correctly addresses all acceptance criteria.

---

### Acceptance Criteria Verification

1. **Three built-in workflow templates defined with flat `agentId: string`** — CODING_WORKFLOW (Planner→Coder), RESEARCH_WORKFLOW (Planner→General), REVIEW_ONLY_WORKFLOW (Coder only) are all present. Gate types match the allowed set: `human_approval`, `pr_review`, `auto`. ✓
2. **Leader is implicit, not a step** — No `leader` agentId appears anywhere in the templates or in the `workflows/` directory. ✓
3. **`seedBuiltInWorkflows` implemented and tested but NOT wired into RPC handlers** — Confirmed: `space-handlers.ts` has zero references to `seedBuiltInWorkflows`. ✓
4. **Files in `packages/daemon/src/lib/space/workflows/`** — Only one file exists in that directory: `built-in-workflows.ts`. ✓
5. **Function name is `seedBuiltInWorkflows`** — Correct. ✓
6. **Seeds ALL three templates** — Loop iterates `getBuiltInWorkflows()` which returns all three. ✓
7. **`resolveAgentId` parameter works correctly** — Tests confirm real UUIDs replace role placeholders for each template step. ✓
8. **38 unit tests, 100% coverage on target file** — Confirmed. ✓
9. **Gate types correct** — `auto`, `human_approval`, `pr_review` all used where expected. ✓
10. **No 'leader' references** — Confirmed. ✓

---

### Issues Found

**P2 — Fallback on unresolved `resolveAgentId` is silently unsafe and comment is incorrect**

`packages/daemon/src/lib/space/workflows/built-in-workflows.ts`, line 193:
```ts
agentId: resolveAgentId(s.agentId as BuiltinAgentRole) ?? s.agentId,
```

When `resolveAgentId` returns `undefined`, the code silently falls back to the placeholder role name (e.g., `'planner'`). The `SpaceWorkflowManager` is constructed without `agentLookup` at seed time (as shown in the test at line 254), so the non-UUID placeholder string passes `validateStepAgentRef` and is persisted to the DB. At runtime, when a workflow run tries to resolve `agentId = 'planner'` to a `SpaceAgent` record, it will fail.

The JSDoc comment on `seedBuiltInWorkflows` (lines 162-164) says "that step is skipped", but the code does not skip the step — it includes it with the placeholder string. The comment is factually wrong.

There is no test covering this partial-failure path, so the behavior is only visible by reading the code carefully.

**P3 — PR title and body are stale**

The PR title still reads `seedDefaultWorkflow (Task 3.4)` and the body still references 36 tests and `seedDefaultWorkflow()`. The second commit updated the implementation to `seedBuiltInWorkflows` with 38 tests but the PR description was not refreshed. Not a code defect, but the merge record will be misleading.

---

### Positive Notes

- Idempotency guard (`if (existing.length > 0) return`) is correct and well-tested.
- 100% line and function coverage on `built-in-workflows.ts` is excellent.
- Clean dependency injection: the seeder takes `SpaceWorkflowManager` rather than the repository directly, preserving the validation layer.
- The `as BuiltinAgentRole` cast on line 193 is safe because all template `agentId` values are statically known role names.

lsm (commented):

### Re-review: P2 Fix Verification

The P2 issue from the previous review has been correctly addressed. Here is the assessment:

**P2 Fix -- Throw on unresolved agentId (RESOLVED)**

The `?? s.agentId` silent fallback in `seedBuiltInWorkflows` has been replaced with an explicit `throw new Error(...)` that includes both the unresolved role name and the space ID in the message (`packages/daemon/src/lib/space/workflows/built-in-workflows.ts`, lines 193-197). The JSDoc at lines 160-162 now correctly states that the function throws rather than silently falling back.

Two new tests were added (lines 356-379 in `built-in-workflows.test.ts`):
1. **Throws on unresolved role** -- verifies the error message includes the role name
2. **No partial persistence** -- verifies that zero workflows are persisted when the resolver fails

All 40 tests pass.

**One observation (P3):** The "no partial persistence" test only works because `planner` appears in the first step of the first template (CODING_WORKFLOW). If the failing role appeared only in a later template (e.g., `general` in RESEARCH_WORKFLOW), the first template would already be persisted before the throw, resulting in partial state. This is not a bug in the fix itself -- it is an inherent property of the non-transactional loop -- but it could be worth wrapping the entire `for` loop in a SQLite transaction in a future iteration if atomicity is desired.

**Recommendation: APPROVE**

No P0, P1, or P2 issues remain. The fix is correct, the JSDoc is accurate, and the test coverage is solid.

lsm (commented):

---

### P2 Fix Assessment

The P2 fix is correct for the specific case it targets: removing the silent `?? s.agentId` fallback and replacing it with an explicit throw. The error message is descriptive, the JSDoc is accurate, and the two new tests (throw case + no-persist case) both pass.

However, the "no workflows persisted" guarantee has a gap that I verified empirically.

---

### P1 — Partial-seed on mid-loop failure

The throw happens inside `template.steps.map(...)`, which runs before `createWorkflow` is called **for that template**. But `createWorkflow` for **earlier templates** in the loop has already been called and committed. So if the failing role is specific to a later template, the space is left in a partially-seeded state.

Concrete example: resolving 'general' fails (only needed by RESEARCH_WORKFLOW, the 2nd template). Result: CODING_WORKFLOW is persisted, then the loop throws — leaving exactly 1 workflow in a space that should have either 0 or 3. I ran this scenario and confirmed:

```
Threw: seedBuiltInWorkflows: no SpaceAgent found for role 'general' in space 'test-space'.
Workflows persisted: 1 [ "Coding Workflow" ]
```

**The existing test ("does not persist any workflow when resolveAgentId fails mid-seed") only covers the safe case where 'planner' fails on the very first step of the very first template — no workflow has been committed yet at that point.** It does not cover a failure on 'general' or any role that first appears in a non-first template.

The fix would need either:
1. A pre-validation pass: resolve all roles across all templates first, throw if any fail, then proceed to persist.
2. Or an explicit test covering the mid-loop failure case to expose and acknowledge the gap.

Relevant file: `packages/daemon/src/lib/space/workflows/built-in-workflows.ts`, lines 189–216.

The function currently processes templates one at a time: build steps (may throw), then persist (no rollback if a later template fails). Factoring out a pre-validation step would make the atomicity guarantee hold for all failure modes.

---

### No other issues

- Template constants are correct (gate types, step counts, roles, ordering).
- `getBuiltInWorkflows()` returns all three.
- Exports in `packages/daemon/src/lib/space/index.ts` are correct.
- Idempotency logic (early-return on non-empty) is correct.
- All 40 tests pass.
- JSDoc now accurately describes the throw behavior.

lsm (commented):

### Re-review: P1 fix for all-or-nothing pre-validation

The P1 issue from the previous review has been correctly addressed. Here is my assessment:

**P1 Fix: Pre-validation pass in `seedBuiltInWorkflows`**

The fix at `packages/daemon/src/lib/space/workflows/built-in-workflows.ts` lines 189-206 is correct and complete:

1. Before persisting any workflow, it collects all unique roles across all templates into a `Set<BuiltinAgentRole>` via `flatMap`.
2. It resolves every role upfront, building a `Map<BuiltinAgentRole, string>`.
3. If any role fails to resolve, it throws immediately -- before any `createWorkflow` call is made.
4. Only after all roles are validated does the persist loop execute (lines 208-226).

This guarantees all-or-nothing behavior: either all workflows are seeded, or none are.

**Test coverage for the fix**

The new test (`does not persist any workflow when resolveAgentId fails on a later-template role`) at line 351 specifically covers the scenario where `'general'` (only needed by the 2nd template RESEARCH_WORKFLOW) fails to resolve, confirming 0 workflows are persisted. This is the exact case that would have been a partial-persist bug without the pre-validation.

Total: 41 tests, all passing, 69 assertions.

**Minor observations (P3):**

- `getBuiltInWorkflows()` returns direct references to the module-level `const` objects (CODING_WORKFLOW, etc.). Since these are typed as `SpaceWorkflow` with mutable arrays (`steps`, `rules`, `tags`), a caller could accidentally mutate them. In practice this is unlikely since `seedBuiltInWorkflows` only reads from them, and the templates are not used elsewhere yet. Noting for future consideration if templates are exposed more broadly.

**Recommendation: APPROVE** -- The P1 fix is correct, well-tested, and the overall implementation is solid.

lsm (commented):

**Recommendation: APPROVE**

### P1 Fix Assessment

The P1 fix is correct and complete. The pre-validation pass in `seedBuiltInWorkflows` (`packages/daemon/src/lib/space/workflows/built-in-workflows.ts`) works exactly as described:

1. Collects all unique `BuiltinAgentRole` values across every template via `flatMap` + `Set` deduplication (lines 193–195).
2. Resolves each role exactly once, throwing immediately if any is missing (lines 197–205).
3. Only enters the persist loop (lines 209–226) after all roles are confirmed — the non-null `!` assertion on line 212 is safe because the `resolvedIds` map was populated from the exact same set of roles that the template steps cast to.

The fix is genuinely all-or-nothing: a missing role on a later template (e.g. `'general'` used only by `RESEARCH_WORKFLOW`) throws before `workflowManager.createWorkflow` is ever called.

### New Test

The new test `does not persist any workflow when resolveAgentId fails on a later-template role` (lines 380–393) directly tests the previously broken path. It uses a resolver that returns `undefined` only for `'general'`, which would have allowed `CODING_WORKFLOW` to commit under the old code. The test confirms 0 workflows persist, and all 41 tests pass.

The `Set` iteration order for `neededRoles` is insertion-order in JavaScript, which is deterministic here because `flatMap` processes templates in array order. The error message will consistently name the first-encountered unresolvable role. This is fine.

No issues remain. The fix is clean and well-targeted.

PR #449, opened 2026-03-19
fix: make CreateWorkflowRunParams.currentStepId optional

lsm:
The DB column accepts NULL and rowToRun maps it back to ''.
Making the field optional in the type contract allows tests and
callers to omit it, and updates the repository INSERT to use null
instead of undefined to satisfy SQLQueryBindings.

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -33,7 +33,7 @@ import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import { SpaceTaskManager } from '../managers/space-task-manager';
-import { WorkflowExecutor, WorkflowGateError } from './workflow-executor';
+import { WorkflowExecutor, WorkflowTransitionError } from './workflow-executor';
 import { Logger } from '../../logger';
 
 const log = new Logger('space-runtime');
@@ -471,7 +471,7 @@ export class SpaceRuntime {
 				this.executorMeta.delete(runId);
 			}
 		} catch (err) {
-			if (!(err instanceof WorkflowGateError)) {
+			if (!(err instanceof WorkflowTransitionError)) {
 				// Unexpected error — propagate to caller (processCompletedTasks collects it)
 				throw err;
 			}
--- a/packages/daemon/src/storage/repositories/space-workflow-run-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-workflow-run-repository.ts
@@ -188,7 +188,7 @@ export class SpaceWorkflowRunRepository {
 			workflowId: row.workflow_id as string,
 			title: row.title as string,
 			description: (row.description as string | null) ?? undefined,
-			currentStepId: (row.current_step_id as string | null) ?? '',
+			currentStepId: (row.current_step_id as string | null) ?? undefined,
 			status: row.status as WorkflowRunStatus,
 			config,
 			createdAt: row.created_at as number,
--- a/packages/daemon/src/storage/repositories/space-workflow-run-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-workflow-run-repository.ts
@@ -39,7 +39,7 @@ export class SpaceWorkflowRunRepository {
 			params.title,
 			params.description ?? '',
 			0, // keep current_step_index for backward compat
-			params.currentStepId,
+			params.currentStepId ?? null,
 			'pending',
 			null,
 			now,
--- a/packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts
@@ -57,7 +57,6 @@ const mockRun: SpaceWorkflowRun = {
 	spaceId: 'space-1',
 	workflowId: 'wf-1',
 	title: 'Run 1',
-	currentStepId: '',
 	status: 'pending',
 	createdAt: NOW,
 	updatedAt: NOW,
--- a/packages/daemon/tests/unit/space/custom-agent.test.ts
+++ b/packages/daemon/tests/unit/space/custom-agent.test.ts
@@ -70,7 +70,6 @@ function makeWorkflowRun(overrides?: Partial<SpaceWorkflowRun>): SpaceWorkflowRu
 		spaceId: 'space-1',
 		workflowId: 'wf-1',
 		title: 'Deploy v1.0',
-		currentStepId: '',
 		status: 'in_progress',
 		createdAt: Date.now(),
 		updatedAt: Date.now(),
--- a/packages/daemon/tests/unit/storage/space-workflow-run-repository.test.ts
+++ b/packages/daemon/tests/unit/storage/space-workflow-run-repository.test.ts
@@ -46,9 +46,17 @@ describe('SpaceWorkflowRunRepository', () => {
 			expect(run.workflowId).toBe(WORKFLOW_ID);
 			expect(run.title).toBe('Run #1');
 			expect(run.status).toBe('pending');
-			expect(run.currentStepId).toBe('');
+			expect(run.currentStepId).toBeUndefined();
 			expect(run.config).toBeUndefined();
 			expect(run.completedAt).toBeUndefined();
+		});
+
+		it('maps NULL currentStepId to undefined (round-trip contract)', () => {
+			// Explicit omission: NULL stored in DB must come back as undefined, not ''
+			const run = repo.createRun({ spaceId, workflowId: WORKFLOW_ID, title: 'No step' });
+			expect(run.currentStepId).toBeUndefined();
+			// Re-fetch from DB to confirm persistence
+			expect(repo.getRun(run.id)!.currentStepId).toBeUndefined();
 		});
 
 		it('creates a run with description', () => {
--- a/packages/shared/src/types/space.ts
+++ b/packages/shared/src/types/space.ts
@@ -253,8 +253,8 @@ export interface SpaceWorkflowRun {
 	title: string;
 	/** Optional description or goal for this run */
 	description?: string;
-	/** ID of the step currently being executed */
-	currentStepId: string;
+	/** ID of the step currently being executed; undefined when the run has not yet been assigned a step */
+	currentStepId?: string;
 	/** Current execution status */
 	status: WorkflowRunStatus;
 	/** Optional runtime configuration for this run */
@@ -275,7 +275,11 @@ export interface CreateWorkflowRunParams {
 	workflowId: string;
 	title: string;
 	description?: string;
-	/** ID of the step to start execution from — should be set to workflow.startStepId */
+	/**
+	 * ID of the step to start execution from — should be set to workflow.startStepId.
+	 * Omit to leave the run without an initial step; the executor must call
+	 * updateCurrentStep() before calling advance().
+	 */
 	currentStepId?: string;
 }
 
--- a/packages/shared/src/types/space.ts
+++ b/packages/shared/src/types/space.ts
@@ -276,7 +276,7 @@ export interface CreateWorkflowRunParams {
 	title: string;
 	description?: string;
 	/** ID of the step to start execution from — should be set to workflow.startStepId */
-	currentStepId: string;
+	currentStepId?: string;
 }
 
 // ============================================================================


lsm (commented):

**Recommendation: APPROVE**

This is a clean, minimal fix that correctly addresses the type contract mismatch.

### Changes reviewed

1. **`packages/shared/src/types/space.ts` (line 279)**: `currentStepId` changed from `string` to `string?` in `CreateWorkflowRunParams`. This is the right call — the DB column accepts NULL, `rowToRun` maps NULL back to `''`, and existing callers (tests + workflow executor) already omit this field when creating runs without a known starting step.

2. **`packages/daemon/src/storage/repositories/space-workflow-run-repository.ts` (line 42)**: `params.currentStepId ?? null` correctly converts `undefined` to `null` for SQLite binding compatibility. This is consistent with the pattern used for `description` on line 40 (`params.description ?? ''`).

### Consistency check

- `SpaceWorkflowRun.currentStepId` remains a required `string` (not optional), which is correct — the `rowToRun` mapper at line 167 guarantees it always returns a string by coalescing NULL to `''`.
- The `WorkflowExecutor` at line 153 uses `this.run.currentStepId` to find the current step, and `find()` returns `undefined` when `currentStepId` is `''`, which is already handled by the `?? null` fallback. No breakage there.
- All test call sites that omit `currentStepId` (e.g., `space-workflow-run-repository.test.ts` line 42) now type-check correctly.

No issues found.

lsm (commented):

**Recommendation: APPROVE** (with one noted smell worth a follow-up)

---

### Summary

The fix correctly addresses the stated goal: `CreateWorkflowRunParams.currentStepId` is now optional (`?: string`) and the repository INSERT uses `params.currentStepId ?? null` to satisfy `SQLQueryBindings`. The targeted test file (`space-workflow-run-repository.test.ts`) passes cleanly and the typecheck exits at zero. The core change is small and correct.

---

### What was done

- `packages/shared/src/types/space.ts` line 279: `currentStepId?: string` (was required `string`)
- `packages/daemon/src/storage/repositories/space-workflow-run-repository.ts` line 42: `params.currentStepId ?? null` in the INSERT

Both changes are correct and the `rowToRun` mapper consistently uses `(row.current_step_id as string | null) ?? ''` to map NULL back to an empty string, keeping `SpaceWorkflowRun.currentStepId` a non-optional `string` on the read side.

---

### Issues

#### P2 — Stale `currentStepIndex` on `SpaceWorkflowRun` mock objects in two test files

`packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts` line 60 and `packages/daemon/tests/unit/space/custom-agent.test.ts` line 73 both construct literal objects typed as `SpaceWorkflowRun` that include the field `currentStepIndex: 0`, which does not exist anywhere in the `SpaceWorkflowRun` interface. This is a dead, misleading field. It does not cause a compile error at present because `packages/daemon/tsconfig.json` excludes `tests/**/*` from type-checking — so excess-property checking never fires on these literals. The tests pass at runtime only because Bun/JS silently ignores the unknown property.

This is not introduced by the current PR (it was likely present before), but the PR touched the same type (`SpaceWorkflowRun`) and is a natural moment to clean it up. Leaving it risks confusion for future contributors who may assume `currentStepIndex` is part of the live type contract.

Both stale references should be removed from:
- `packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts` (line 60)
- `packages/daemon/tests/unit/space/custom-agent.test.ts` (line 73)

#### P3 — NULL-to-empty-string sentinel in `rowToRun`

`packages/daemon/src/storage/repositories/space-workflow-run-repository.ts` line 167 maps a NULL `current_step_id` to the empty string `''`. The test on line 49 then asserts `expect(run.currentStepId).toBe('')`. This is a leak of an internal DB representation into the public type: consumers of `SpaceWorkflowRun.currentStepId` can receive `''` and must remember that `''` means "not yet set", which is semantically identical to `undefined`. A cleaner contract would either make `SpaceWorkflowRun.currentStepId` optional (`?: string`) on the read side as well, or document that `''` is the sentinel. Not blocking but worth tracking.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, minimal fix that correctly addresses the type contract mismatch.

### Summary

The change makes `CreateWorkflowRunParams.currentStepId` optional and ensures `null` is passed to SQLite instead of `undefined` (which is not a valid `SQLQueryBindings` value). The stale `currentStepIndex` fields are correctly removed from test mocks since that property does not exist on `SpaceWorkflowRun`.

### Verification

1. **Type contract consistency**: `CreateWorkflowRunParams.currentStepId` is now `string | undefined` (optional), while `SpaceWorkflowRun.currentStepId` remains `string` (required). The bridge between these is `rowToRun()` at line 167 of `space-workflow-run-repository.ts`, which maps `NULL` back to `''`. This is consistent — callers can omit the field on creation, and the read model always returns a string.

2. **SQL binding correctness**: `params.currentStepId ?? null` correctly converts `undefined` to `null` for SQLite. Bun's SQLite driver accepts `null` but not `undefined` in bindings.

3. **Test cleanup**: The removed `currentStepIndex: 0` lines in `space-handlers.test.ts` and `custom-agent.test.ts` are correct — that property was from the original plan doc but never made it into the `SpaceWorkflowRun` interface. No remaining references to `currentStepIndex` exist outside of doc files.

4. **No missed callers**: Checked all `createRun()` call sites — the `workflow-executor.test.ts` callers already pass `currentStepId` explicitly (set to `startStepId`), so making it optional does not break them.

No issues found.

lsm (commented):

**Recommendation: APPROVE** (own PR — posted as COMMENT per policy)

---

This PR has two commits: making `CreateWorkflowRunParams.currentStepId` optional and removing stale `currentStepIndex` fields from test mock objects. Both changes are correct and cleanly executed. A few observations worth noting.

### What is correct

The `?? null` substitution in `space-workflow-run-repository.ts` line 42 is necessary because Bun's SQLite bindings reject `undefined` as a bind value but accept `null`. The DB column already accepts `NULL` and `rowToRun` maps `NULL` back to `''`, so no schema change is required.

Removing `currentStepIndex: 0` from both mock objects is correct — `SpaceWorkflowRun` has never had that field. The fields were silently ignored at runtime and were never type-checked because the test files are excluded from `tsc`. Cleaning them up reduces misleading signal for future contributors.

### P1: Type divergence between `SpaceWorkflowRun.currentStepId` and `rowToRun` sentinel

`SpaceWorkflowRun.currentStepId` is typed as `string` (non-optional, line 257 of `packages/shared/src/types/space.ts`). `rowToRun` maps a `NULL` DB value back to `''` (empty string) as a sentinel.

The test at `packages/daemon/tests/unit/storage/space-workflow-run-repository.test.ts:49` explicitly asserts `expect(run.currentStepId).toBe('')` when no `currentStepId` was supplied at creation, confirming this sentinel is intentional.

`WorkflowExecutor.getCurrentStep()` calls `this.workflow.steps.find(s => s.id === this.run.currentStepId)` — when `currentStepId` is `''`, `find` returns `undefined` and the method returns `null`. That is benign here because the fallback to `null` is what callers already expect.

However, `getOutgoingTransitions()` calls `.filter(t => t.from === this.run.currentStepId)` — when `currentStepId` is `''`, this silently returns an empty array. Any caller that expects transitions to be available on a freshly-created `pending` run that hasn't yet been given a real step ID will silently get no transitions, which could mask a misconfigured run initialization. This is a latent bug that the type system would surface if `SpaceWorkflowRun.currentStepId` were made `string | null` (or optional) to match the actual DB/runtime semantics. The PR intentionally stops at `CreateWorkflowRunParams` and leaves the read-side interface misaligned.

This is worth fixing as a follow-up: either make `SpaceWorkflowRun.currentStepId` optional/nullable to reflect that a newly-created run may not yet have a step ID, or ensure that all `createRun` callers always supply a valid `currentStepId` (i.e., make the field required again and document that).

### P2: Test coverage gap for the new optional path

There is no test asserting that `createRun({ ..., currentStepId: undefined })` produces a run where `currentStepId` is `''` (or `null`). The existing test at line 49 exercises omission because the param was never required in the test, but now that the type explicitly permits omission, a dedicated assertion that `NULL → ''` round-trips correctly would make the contract explicit and prevent regressions if `rowToRun` is ever changed.

### P3: JSDoc comment is slightly misleading after the change

`packages/shared/src/types/space.ts` line 278: the comment reads `should be set to workflow.startStepId`. Since the field is now optional, the comment implies there is a preferred value but does not explain when omission is acceptable. A brief note such as `omit to leave the run without an initial step (the executor must call updateCurrentStep before advance())` would help.

---

Overall this is a minimal, focused fix that unblocks callers and cleans up stale mock fields. The P1 type divergence on the read interface is a pre-existing debt exposed — not introduced — by this PR, but it is worth tracking.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped fix that aligns the TypeScript type contract with the actual DB semantics. All changes are correct and consistent.

### What was done

1. **`SpaceWorkflowRun.currentStepId`** changed from `string` (required) to `string | undefined` (optional) on the read side, matching the DB column which accepts NULL.
2. **`CreateWorkflowRunParams.currentStepId`** made optional with clear JSDoc explaining when omission is acceptable.
3. **`rowToRun`** now maps `NULL -> undefined` instead of `NULL -> ''`, eliminating the empty-string sentinel ambiguity.
4. **Repository INSERT** uses `params.currentStepId ?? null` to satisfy SQLite bindings.
5. **Stale `currentStepIndex`** property removed from two test mock objects (never existed on the `SpaceWorkflowRun` interface).
6. **Round-trip test** added to verify NULL persistence and undefined deserialization.

### Verification

- All consumers of `currentStepId` in `workflow-executor.ts` (lines 153, 161) handle `undefined` gracefully: `.find()` returns `undefined` which falls through to `?? null`, and `.filter()` matches nothing -- both correct behaviors for a run with no step assigned yet.
- The `updateRun` method (line 109) checks `params.currentStepId !== undefined` before adding to the SQL update, so `undefined` values are correctly ignored (not accidentally writing NULL over a set value).
- Mock objects in `space-handlers.test.ts` and `custom-agent.test.ts` now omit `currentStepId` entirely, which is valid since the field is optional.
- No other files in the codebase access `currentStepId` in a way that would break with `undefined`.

No issues found. The change is minimal, correct, and well-tested.

lsm (commented):

---

**Recommendation: APPROVE**

The implementation correctly addresses all three stated goals. The changes are minimal, targeted, and consistent throughout the codebase.

### What was done right

**Type contract (packages/shared/src/types/space.ts)**
Both `SpaceWorkflowRun.currentStepId` and `CreateWorkflowRunParams.currentStepId` are now optional (`?: string`). The JSDoc on the params field is accurate and actionable — it tells callers that the executor must call `updateCurrentStep()` before `advance()`, which correctly describes the contract enforced by `WorkflowExecutor.getCurrentStep()` (which returns `null` when `currentStepId` is undefined, causing `advance()` to throw).

**Repository mapping (packages/daemon/src/storage/repositories/space-workflow-run-repository.ts)**
- `createRun`: `params.currentStepId ?? null` is the correct idiom for SQLite — passing `undefined` to Bun's SQLite binding is undefined behavior, `null` reliably stores `NULL`.
- `rowToRun`: `(row.current_step_id as string | null) ?? undefined` correctly maps DB `NULL` → JS `undefined`, eliminating the empty-string sentinel that broke the type contract.

**Test coverage (packages/daemon/tests/unit/storage/space-workflow-run-repository.test.ts)**
The new round-trip test does exactly what it claims — creates without `currentStepId`, asserts undefined on the returned object, then re-fetches from DB and asserts again. This is a genuine persistence round-trip, not just an in-memory check.

**Test fixture cleanup (space-handlers.test.ts, custom-agent.test.ts)**
Removing `currentStepIndex: 0` from the mock `SpaceWorkflowRun` fixtures is correct — that field does not exist on the type. The diff removes a property that was either a typo or a stale remnant. (Note: `currentStepIndex` is a DB-level backward-compat column only, not surfaced on the TypeScript interface.)

**P3 — `description` stores `''` but `currentStepId` stores `null`**
In `createRun`, `params.description ?? ''` stores an empty string when description is omitted, while `rowToRun` maps `description: (row.description as string | null) ?? undefined`. This means an omitted description round-trips as `undefined` only if the DB column is `NULL`, but it actually stores `''`. An empty string is falsy in the `??` expression check only for `null`/`undefined`, so `'' ?? undefined` evaluates to `''`, meaning description round-trips as `''` not `undefined`. This is a pre-existing inconsistency not introduced by this PR. Worth a follow-up but out of scope here.

**P3 — No test for `createRun` *with* `currentStepId` round-trip**
The existing `updateCurrentStep` test covers setting a step ID after creation. A test asserting that `createRun({ ..., currentStepId: 'step-xyz' })` persists and reads back as `'step-xyz'` would complete the picture. This is minor since the path is well-exercised by the workflow-executor tests which always supply `currentStepId: workflow.startStepId`.

**P3 — `WorkflowExecutor.advance()` guard on undefined currentStepId**
`getCurrentStep()` returns `null` when `currentStepId` is undefined (via `Array.find` returning `undefined` coerced by `?? null`). `advance()` throws `'Cannot advance: no current step'` in that case. This is the correct behavior described in the JSDoc, and it works fine — but there is no dedicated unit test for this guard path (advance on a run with no currentStepId). Not a blocking concern since the throw path is straightforward, but it would be a nice addition.

### Summary

The change is correct, complete, and safe. The core bug (empty-string sentinel replacing `undefined`) is fixed at both the write path (using `null` for SQLite) and the read path (`?? undefined`). All existing callers are compatible — workflow-executor tests still supply `currentStepId` at creation, and the new optional semantics only relax the constraint for callers that legitimately need to omit it. Tests verify the round-trip contract explicitly.

PR #452, opened 2026-03-20
feat: add Space export/import RPC handlers (Task 8.2)

lsm:
Implements spaceExport.* and spaceImport.* RPC namespaces:
- spaceExport.agents / spaceExport.workflows / spaceExport.bundle
  for exporting Space entities to portable SpaceExportBundle JSON
- spaceImport.preview for dry-run conflict detection, cross-reference
  validation, and SpaceWorkflowManager gate validation
- spaceImport.execute for creating entities with skip/rename/replace
  conflict resolution per item

Cross-reference handling:
- Agent name→UUID resolution: checks bundle agents first, falls back
  to existing space agents
- Step ID remapping: fresh UUIDs assigned to all imported steps;
  rule appliesTo arrays remapped from step names to new step UUIDs
- Transition from/to step names remapped to new step UUIDs

Wired into setupRPCHandlers() in rpc-handlers/index.ts.
45 unit tests covering all handlers, conflict resolutions, and
cross-reference mapping.

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -57,6 +57,7 @@ import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
 import { SpaceWorkflowRepository } from '../../storage/repositories/space-workflow-repository';
 import { SpaceAgentRepository } from '../../storage/repositories/space-agent-repository';
 import { SpaceRuntime } from '../space/runtime/space-runtime';
+import { setupSpaceExportImportHandlers } from './space-export-import-handlers';
 
 export interface RPCHandlerDependencies {
 	messageHub: MessageHub;
@@ -233,6 +234,16 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanu
 		taskRepo: spaceTaskRepo,
 	});
 	spaceRuntime.start();
+
+	// Space export/import handlers
+	setupSpaceExportImportHandlers(
+		deps.messageHub,
+		deps.spaceManager,
+		spaceAgentRepo,
+		spaceWorkflowRepo,
+		spaceWorkflowManager,
+		deps.db.getDatabase()
+	);
 
 	// Return cleanup function to stop background services
 	return () => {
--- a/packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts
@@ -1,0 +1,523 @@
+/**
+ * Space Export/Import RPC Handlers
+ *
+ * Export namespace (spaceExport.*):
+ *   spaceExport.agents    { spaceId, agentIds? }                       → { bundle: SpaceExportBundle }
+ *   spaceExport.workflows { spaceId, workflowIds? }                    → { bundle: SpaceExportBundle }
+ *   spaceExport.bundle    { spaceId, agentIds?, workflowIds? }         → { bundle: SpaceExportBundle }
+ *
+ * Import namespace (spaceImport.*):
+ *   spaceImport.preview   { bundle, spaceId }                          → ImportPreviewResult
+ *   spaceImport.execute   { spaceId, bundle, conflictResolution? }     → ImportExecuteResult
+ *
+ * Cross-reference rules:
+ * - Exported workflow steps store the agent's display **name** (`agentRef`), not UUID.
+ * - On import, agent names are resolved to UUIDs by checking:
+ *     1. Agents being imported in the same bundle (by original bundle name)
+ *     2. Agents already present in the target space (by name)
+ * - If a name cannot be resolved, preview flags it as a validation error;
+ *   execute throws and aborts import of that workflow.
+ * - Rule `appliesTo` lists step **names** in the exported format and are
+ *   remapped to new step UUIDs on import.
+ */
+
+import { generateUUID } from '@neokai/shared';
+import type {
+	MessageHub,
+	Space,
+	SpaceAgent,
+	CreateSpaceAgentParams,
+	CreateSpaceWorkflowParams,
+	WorkflowStepInput,
+	WorkflowTransitionInput,
+	WorkflowRuleInput,
+	SpaceExportBundle,
+	ExportedSpaceAgent,
+	ExportedSpaceWorkflow,
+} from '@neokai/shared';
+import type { SpaceManager } from '../space/managers/space-manager';
+import type { SpaceWorkflowManager } from '../space/managers/space-workflow-manager';
+import type { SpaceAgentRepository } from '../../storage/repositories/space-agent-repository';
+import type { SpaceWorkflowRepository } from '../../storage/repositories/space-workflow-repository';
+import { exportBundle, validateExportBundle } from '../space/export-format';
+
+// ============================================================================
+// Public types
+// ============================================================================
+
+export interface ImportPreview {
+	name: string;
+	action: 'create' | 'conflict';
+	existingId?: string;
+}
+
+export interface ImportPreviewResult {
+	agents: ImportPreview[];
+	workflows: ImportPreview[];
+	validationErrors: string[];
+}
+
+export type ConflictResolutionStrategy = 'skip' | 'rename' | 'replace';
+
+export interface ImportConflictResolution {
+	agents?: Record<string, ConflictResolutionStrategy>;
+	workflows?: Record<string, ConflictResolutionStrategy>;
+}
+
+export interface ImportedItem {
+	name: string;
+	id: string;
+	action: 'created' | 'skipped' | 'renamed' | 'replaced';
+}
+
+export interface ImportExecuteResult {
+	agents: ImportedItem[];
+	workflows: ImportedItem[];
+	warnings: string[];
+}
+
+// ============================================================================
+// Private helpers
+// ============================================================================
+
+async function requireSpace(spaceManager: SpaceManager, spaceId: string): Promise<Space> {
+	if (!spaceId) throw new Error('spaceId is required');
+	const space = await spaceManager.getSpace(spaceId);
+	if (!space) throw new Error(`Space not found: ${spaceId}`);
+	return space;
+}
+
+/** Generate a name that does not collide with anything in `existingNames`. */
+function generateUniqueName(baseName: string, existingNames: Set<string>): string {
+	if (!existingNames.has(baseName)) return baseName;
+	let counter = 1;
+	while (existingNames.has(`${baseName} (${counter})`)) counter++;
+	return `${baseName} (${counter})`;
+}
+
+function buildAgentCreateParams(
+	spaceId: string,
+	name: string,
+	exported: ExportedSpaceAgent
+): CreateSpaceAgentParams {
+	const params: CreateSpaceAgentParams = { spaceId, name, role: exported.role };
+	if (exported.description !== undefined) params.description = exported.description;
+	if (exported.model !== undefined) params.model = exported.model;
+	if (exported.provider !== undefined) params.provider = exported.provider;
+	if (exported.systemPrompt !== undefined) params.systemPrompt = exported.systemPrompt;
+	if (exported.tools !== undefined) params.tools = exported.tools;
+	return params;
+}
+
+/**
+ * Convert an `ExportedSpaceWorkflow` into `CreateSpaceWorkflowParams` suitable
+ * for `SpaceWorkflowRepository.createWorkflow()`.
+ *
+ * Step names are assigned fresh UUIDs; rule `appliesTo` arrays are remapped from
+ * step names to those new UUIDs; agent refs are resolved via the two lookup maps.
+ *
+ * @returns params ready for the repository, the step-name→UUID map (for rule
+ *          appliesTo remapping), and any warnings about unresolved agent refs.
+ */
+function buildWorkflowCreateParams(
+	spaceId: string,
+	name: string,
+	exported: ExportedSpaceWorkflow,
+	importedAgentNameToId: Map<string, string>,
+	existingAgentNameToId: Map<string, string>
+): { params: CreateSpaceWorkflowParams; stepNameToId: Map<string, string>; warnings: string[] } {
+	const warnings: string[] = [];
+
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured implementation of the Space export/import RPC handlers. The code is clean, well-documented, and the test coverage is thorough with 45 passing tests. All acceptance criteria are met.

### What looks good

- **Correct RPC namespace**: All handlers use `spaceExport.*`/`spaceImport.*` as specified.
- **spaceId context**: Every handler validates `spaceId` via `requireSpace()` -- no `roomId` usage.
- **Export/import read/write correctly**: Export reads from `space_agents`/`space_workflows` tables via the repos; import writes via the same repos and `SpaceWorkflowManager`.
- **Preview validation**: Detects conflicts (name collisions), unresolved agent references, and condition expression errors.
- **Cross-reference mapping**: Agent name→UUID resolution (bundle-first, then target space) and step name→UUID remapping for transitions and rule `appliesTo` arrays are correctly implemented.
- **Conflict resolution**: All three strategies (skip, rename, replace) work correctly for both agents and workflows.
- **Smart workflow export**: `spaceExport.workflows` includes only referenced agents in the bundle, not all agents in the space.
- **Tests use real repositories**: In-memory SQLite with actual `SpaceAgentRepository`, `SpaceWorkflowRepository`, and `SpaceWorkflowManager` -- exercising real business logic, not just mocks.

### Issues

**P2: No transaction wrapping in `spaceImport.execute`** (`packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 380-522)

The execute handler performs multiple sequential database mutations (create/update agents, then create workflows) without a wrapping transaction. If a workflow import fails mid-way (e.g., unresolved agent ref throws at line 505-507), previously imported agents are already committed and the database is left in a partially-imported state. This is especially problematic with the `replace` strategy for workflows, where the existing workflow is deleted (line 482) before the new one is created -- if creation fails, the original is lost.

Consider wrapping the entire execute handler body in a SQLite transaction (BEGIN/COMMIT with ROLLBACK on error) to ensure atomicity.

**P2: `replace` strategy for agents does not clear unset optional fields** (`packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 436-443)

When replacing an agent, fields like `description`, `model`, `provider`, `systemPrompt`, and `tools` are passed through from the exported agent. If these fields are `undefined` in the exported bundle (i.e., not present), `agentRepo.update()` will skip them due to its `if (params.X !== undefined)` guards, meaning the original agent's values for those fields are preserved rather than cleared. This creates an asymmetry: a `replace` does not produce the same result as a `delete` + `create` would. Whether this is intentional "merge" behavior or a bug depends on the desired semantics, but it should be documented either way. At minimum, consider explicitly passing `null` or empty values for fields not present in the export to ensure a true replacement.

**P3: `generateUniqueName` has no upper bound** (`packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 91-96)

The `while` loop has no guard against pathological cases (e.g., thousands of existing names). In practice this is extremely unlikely to matter, but a safety cap (e.g., `counter < 10000`) with an error throw would be defensive.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two P1 issues around non-atomic import execution and two P2 issues around case-insensitive conflict detection and dead validation code.

---

### P1 — `spaceImport.execute` is not atomic: partial failures leave DB in inconsistent state

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 408–513

The entire import runs as a sequence of individual DB writes with no wrapping transaction. Two distinct failure modes can corrupt the database:

**Scenario A — workflow `replace` strategy deletes the existing record then fails to create**

Line 482 calls `workflowRepo.deleteWorkflow(existing.id)` before attempting `workflowManager.createWorkflow(createParams)` on line 511. If `createWorkflow` throws (e.g., a `WorkflowValidationError` due to a bad agent UUID, or any unexpected error), the original workflow is permanently gone with no rollback. The agents imported in Phase 1 are also committed and cannot be undone.

**Scenario B — multi-workflow bundle with later failure**

Suppose a bundle has three workflows: W1 succeeds, W2 needs `replace` (deletes existing, `createWorkflow` throws), W3 was not reached. W1 is now created, the original W2 is deleted, and W3 never imported. The caller only receives the thrown error with no indication of what partial work was done.

The fix is to wrap the full operation in a SQLite transaction via `db.transaction()` (available through the repositories' underlying `BunDatabase`). The `SpaceWorkflowRepository`/`SpaceAgentRepository` currently do not expose transaction control, so this would require either passing the raw `db` handle to the handler or adding a `withTransaction(fn)` helper to the storage layer.

---

### P1 — `replace` workflow deletion is outside any try/catch with no compensation

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 480–488

Even if `WorkflowValidationError` would never realistically be thrown for a well-formed bundle, `workflowManager.createWorkflow` calls `validateName`, which performs a fresh DB check. A race condition between `preview` and `execute` (two concurrent `execute` calls from different clients for the same spaceId) could trigger this. After `deleteWorkflow(existing.id)` on line 482, any subsequent throw within the loop causes the workflow to disappear entirely. This is an amplified version of the P1 above.

---

### P2 — Conflict detection uses case-sensitive map lookup but DB enforces case-insensitive uniqueness

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 339, 400

`existingAgentByName` and `existingWorkflowByName` are constructed with exact-case keys:
```ts
const existingAgentByName = new Map(existingAgents.map((a) => [a.name, a]));
```

But `SpaceAgentRepository.isNameTaken` uses `LOWER()` in SQL (case-insensitive), meaning the DB already prevents two agents named "Coder" and "coder" from coexisting. This means the Map lookup will always be consistent for agents that currently exist.

However, the `usedAgentNames` set on line 405 is also populated from DB names and used with exact-case `Set.has()` in `generateUniqueName`. If a future agent with a mixed-case name was somehow inserted outside the manager (e.g., via a migration), a renamed import could produce a name the DB would reject as a duplicate. More importantly, the `existingWorkflowByName` map for workflows — `SpaceWorkflowManager.validateName` uses exact-case comparison (`wf.name === name`) so workflows are case-sensitive in the DB. This mismatch in documentation and behavior is confusing and should be explicitly noted or aligned.

---

### P2 — Condition validation in `validateWorkflowForPreview` is dead code

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 223–229

The condition check at lines 226–228:
```ts
if (t.condition?.type === 'condition' && !t.condition.expression?.trim()) {
    errors.push(`transition[${i}]: 'condition' type requires a non-empty expression`);
}
```
is never reached for any bundle that passes `validateExportBundle`. The Zod schema `workflowConditionSchema.superRefine` in `export-format.ts` (line 44–51) already validates this invariant during bundle parsing. If the bundle has an empty `condition` expression, `validateExportBundle` returns `ok: false` and `spaceImport.preview` short-circuits at lines 325–332 before ever calling `validateWorkflowForPreview`. The test at line 578 appears to test this path but actually exercises the Zod validation, not the in-`validateWorkflowForPreview` check.

The function comment says it "mirrors SpaceWorkflowManager.validateCondition", but since `workflowManager.createWorkflow` will also re-validate conditions on execute, there are now three copies of this logic. The dead check in `validateWorkflowForPreview` should be removed to eliminate the maintenance burden.

---

### P3 — Minor: `usedWorkflowNames` not updated in the `rename` path before `workflowManager.createWorkflow`

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, line 487

`generateUniqueName` returns `"Pipeline (1)"` at line 487, but `usedWorkflowNames.add(finalName)` is not called until line 512 (after the create). If two same-named workflows appeared in the same bundle with `rename` strategy, the second would also get `"Pipeline (1)"`, causing `workflowManager.createWorkflow` to throw. In practice this cannot happen because a single space cannot have two workflows with the same name (so a valid export bundle will never contain duplicate workflow names), but the invariant is fragile and should be documented or defensively guarded.

---

### What's done well

- Clean two-phase import (agents first, workflows second) that correctly builds the `importedAgentNameToId` map before resolving workflow cross-references.
- `generateUniqueName` with an in-memory `usedNames` set correctly avoids collision across the import batch.
- `validateExportBundle` is re-run on `execute` to guard against stale/tampered payloads — good defensive practice.
- Bundle validation via Zod schemas in `export-format.ts` is thorough (duplicate step names, transition endpoint refs, startStep ref).
- All 45 unit tests pass and cover the happy path plus all three conflict strategies, cross-reference mapping, and rule `appliesTo` remapping.
- `spaceExport.workflows` correctly trims the `agents` array to only referenced agents — avoids leaking unreferenced agents in the export bundle.

lsm (commented):

All five previously reported issues have been correctly addressed in commit `9579438`:

**P1 fixes verified:**
1. **Transaction atomicity** -- `spaceImport.execute` now wraps all DB mutations inside `db.transaction()` (lines 418-557 of `space-export-import-handlers.ts`). The transaction callback is synchronous as required by Bun's SQLite `db.transaction()` API. The async `requireSpace` check correctly happens outside the transaction. Two new tests (`rolls back agent creation when workflow import fails` and `rolls back workflow deletion when replacement creation fails`) verify rollback behavior.

2. **Dead condition check removed** -- `validateWorkflowForPreview` no longer contains the redundant condition expression validation block. The JSDoc now clearly documents that condition validation is handled by the Zod schema in `validateExportBundle()`. The existing test (`flags condition transition without expression as error`) still passes because the Zod validation catches it first.

**P2 fix verified:**
3. **Agent `replace` field clearing** -- The `replace` strategy now uses `?? null` for all optional fields (`description`, `model`, `provider`, `systemPrompt`, `tools`) at lines 462-468, ensuring absent fields are explicitly cleared rather than left with stale values. Two tests verify this for `model` and `systemPrompt`.

**P3 fixes verified:**
4. **`usedWorkflowNames` tracking** -- `usedWorkflowNames.add(finalName)` now happens at line 525, before the workflow is created via `workflowManager.createWorkflow`. This ensures that if two workflows with the same name appear in one bundle with `rename` strategy, they get different suffixed names.

5. **`generateUniqueName` counter cap** -- The `while` loop now has `counter < 10_000` guard (line 113) with a corresponding `throw` if the cap is reached (lines 114-116). This prevents theoretical infinite loops.

**Additional observations:**
- Tests pass: 49/49, 0 failures.
- The `db` parameter is correctly threaded from `deps.db.getDatabase()` in `index.ts` line 232 through to the handler setup.
- Documentation comments are thorough and accurate, particularly the atomicity and replace semantics notes at the top of the file.

lsm (commented):

**Recommendation: APPROVE**

All five previously-reported issues have been correctly addressed. The implementation is solid. Below is a detailed verification of each fix plus a scan of remaining code.

---

### Fix Verification

**P1 — Transaction atomicity in `spaceImport.execute`**

Confirmed. All DB mutations are wrapped in `db.transaction()` at line 418. The transaction function is synchronous (as required by Bun SQLite's `db.transaction()` API). The space validation (`requireSpace`) correctly happens before the transaction, which is the right pattern since `getSpace` is async and Bun's transaction callback must be synchronous.

Two rollback tests were added: one for agent creation rollback when a workflow import fails (line 942), and one for workflow deletion rollback when the replacement creation fails (line 974). Both test cases work against the real in-memory SQLite schema, giving meaningful coverage.

**P1 — Dead condition check in `validateWorkflowForPreview` removed**

Confirmed. The function at lines 233–249 only checks for unresolved agent refs. The previous condition expression validation (which was dead code pre-empted by Zod schema validation) is gone. The inline comment at line 222 explains the intentional omission.

**P2 — Agent `replace` strategy explicitly clears absent optional fields**

Confirmed. Lines 461–468 pass `exportedAgent.field ?? null` for all optional fields. Because `agentRepo.update()` uses `if (params.X !== undefined)` guards, passing `null` (not `undefined`) ensures each absent-from-export field is actively cleared in the DB. The `systemPrompt` and `description` columns are NOT NULL in the DB schema, so the update method converts null to `''`, and `rowToAgent` then maps `''` back to `undefined` — which matches the test expectations at lines 1043 and 1063.

Two tests cover this (`clears model when not present` and `clears systemPrompt when not present`). The `tools` clearing is not covered by a dedicated test but the same logic applies.

**P3 — `usedWorkflowNames` tracks names before creation to prevent duplicate-bundle collisions**

Confirmed. Line 525 adds `finalName` to `usedWorkflowNames` before calling `workflowManager.createWorkflow()`. For the rename conflict path, this means two identically-named workflows in a bundle both marked for rename will receive distinct suffixed names (e.g., `Pipeline (1)` and `Pipeline (2)`) rather than colliding.

No test for this specific edge case was added, but because the DB schema enforces workflow name uniqueness per space, the no-test gap is a minor omission rather than a bug — a collision would still be caught as a DB error that rolls back the transaction.

**P3 — `generateUniqueName` counter cap**

Confirmed. Lines 113–117 cap the while loop at `counter < 10_000` and throw a descriptive error when the cap is reached, preventing an infinite loop.

---

### Remaining Observations

**No new issues of P0/P1 severity were found.**

Minor observations (P3):

1. **Missing test for same-named workflows within one bundle (rename path):** The `usedWorkflowNames` fix in Phase 2 was specifically motivated by the case of two identically-named workflows in one bundle both going through `rename`. A test like 'two workflows named Pipeline in the same bundle both get unique rename suffixes' would close this gap.

2. **`validateExportBundle` does not detect duplicate workflow names within a bundle:** The Zod schema permits two workflows named 'Pipeline' in the same bundle. This is only relevant for hand-crafted / tampered bundles (legitimate exports are always unique at the DB level), and any collision in the no-conflict-create path would still fail inside the transaction with a DB constraint error and roll back cleanly. Low-impact gap but worth noting.

3. **Index wiring is correct:** `setupSpaceExportImportHandlers` is called at lines 226–233 of `index.ts`, reusing the same `spaceAgentRepo`, `spaceWorkflowRepo`, and `spaceWorkflowManager` instances created immediately above for `setupSpaceWorkflowHandlers`. No dependency mismatch.

---

All requested fixes from the previous review are correctly and completely implemented. The code is clean, well-commented, and the test suite exercises the critical paths including the new atomicity guarantees.

PR #453, opened 2026-03-20
feat: workflow selection logic and Space agent tools (Task 7.1)

lsm:
- Add selectWorkflow() pure deterministic function in workflow-selector.ts
  with 4-level priority chain: explicit ID → tag matching → keyword/description
  matching → null fallback
- Add resolveWorkflowForRun() integration point in SpaceRuntime that delegates
  to selectWorkflow() with space workflows loaded from DB
- New space-agent-tools.ts with 5 MCP tools: start_workflow_run (auto-selects
  workflow via selectWorkflow when no explicit ID given), create_task, list_workflows,
  list_tasks (filterable by status/workflowRunId), list_workflow_runs
- Export new modules from space/index.ts
- 36 unit tests covering all selection priority levels and all tool handlers

### DIFF
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -31,6 +31,15 @@ export type {
 export { SpaceRuntime } from './runtime/space-runtime';
 export type { SpaceRuntimeConfig, ResolvedTaskType } from './runtime/space-runtime';
 
+export { selectWorkflow } from './runtime/workflow-selector';
+export type { WorkflowSelectionContext } from './runtime/workflow-selector';
+
+export {
+	createSpaceAgentToolHandlers,
+	createSpaceAgentMcpServer,
+} from './tools/space-agent-tools';
+export type { SpaceAgentToolsConfig, SpaceAgentMcpServer } from './tools/space-agent-tools';
+
 export {
 	exportAgent,
 	exportWorkflow,
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -34,6 +34,8 @@ import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/s
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import { SpaceTaskManager } from '../managers/space-task-manager';
 import { WorkflowExecutor, WorkflowTransitionError } from './workflow-executor';
+import { selectWorkflow } from './workflow-selector';
+import type { WorkflowSelectionContext } from './workflow-selector';
 import { Logger } from '../../logger';
 
 const log = new Logger('space-runtime');
@@ -309,6 +311,23 @@ export class SpaceRuntime {
 		return workflow.rules.filter(
 			(r) => !r.appliesTo || r.appliesTo.length === 0 || r.appliesTo.includes(stepId)
 		);
+	}
+
+	/**
+	 * Resolve the best workflow for a new run using the workflow selection algorithm.
+	 *
+	 * Accepts an optional explicit `workflowId`. When omitted, auto-selects via
+	 * tag-based and keyword-based heuristics. Returns null when no workflow matches
+	 * (caller should create a standalone task instead).
+	 *
+	 * This is a thin integration point: it loads the available workflows for the
+	 * space and delegates to the pure `selectWorkflow()` function.
+	 */
+	resolveWorkflowForRun(
+		context: Omit<WorkflowSelectionContext, 'availableWorkflows'>
+	): SpaceWorkflow | null {
+		const availableWorkflows = this.config.spaceWorkflowManager.listWorkflows(context.spaceId);
+		return selectWorkflow({ ...context, availableWorkflows });
 	}
 
 	/**
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -35,7 +35,6 @@ import type { SpaceTaskRepository } from '../../../storage/repositories/space-ta
 import { SpaceTaskManager } from '../managers/space-task-manager';
 import { WorkflowExecutor, WorkflowTransitionError } from './workflow-executor';
 import { selectWorkflow } from './workflow-selector';
-import type { WorkflowSelectionContext } from './workflow-selector';
 import { Logger } from '../../logger';
 
 const log = new Logger('space-runtime');
@@ -314,20 +313,18 @@ export class SpaceRuntime {
 	}
 
 	/**
-	 * Resolve the best workflow for a new run using the workflow selection algorithm.
+	 * Resolve a workflow for a new run from an explicit workflowId.
 	 *
-	 * Accepts an optional explicit `workflowId`. When omitted, auto-selects via
-	 * tag-based and keyword-based heuristics. Returns null when no workflow matches
-	 * (caller should create a standalone task instead).
+	 * Returns the workflow if found in this space's workflows, or null when:
+	 *   - No workflowId is provided (LLM agent must call list_workflows first)
+	 *   - The provided workflowId is not found in this space
 	 *
-	 * This is a thin integration point: it loads the available workflows for the
-	 * space and delegates to the pure `selectWorkflow()` function.
+	 * This is a thin integration point: it loads the space's workflows from the
+	 * DB and delegates to the pure `selectWorkflow()` function.
 	 */
-	resolveWorkflowForRun(
-		context: Omit<WorkflowSelectionContext, 'availableWorkflows'>
-	): SpaceWorkflow | null {
-		const availableWorkflows = this.config.spaceWorkflowManager.listWorkflows(context.spaceId);
-		return selectWorkflow({ ...context, availableWorkflows });
+	resolveWorkflowForRun(spaceId: string, workflowId?: string): SpaceWorkflow | null {
+		const availableWorkflows = this.config.spaceWorkflowManager.listWorkflows(spaceId);
+		return selectWorkflow({ spaceId, availableWorkflows, workflowId });
 	}
 
 	/**
--- a/packages/daemon/src/lib/space/runtime/workflow-selector.ts
+++ b/packages/daemon/src/lib/space/runtime/workflow-selector.ts
@@ -1,19 +1,19 @@
 /**
  * Workflow Selector
  *
- * Pure, deterministic function for selecting a SpaceWorkflow given a context.
+ * Pure function for resolving a SpaceWorkflow from an explicit workflowId.
  * Has no runtime dependencies — takes only data inputs so it can be tested
  * in isolation without any DB or manager instances.
  *
- * Priority chain (first match wins):
- *   1. Explicit workflowId provided → use it
- *   2. Tag-based: match keywords from title/description against workflow tags
- *   3. MVP keyword matching: substring match title/description against workflow descriptions
- *   4. Fall back to null (create standalone task)
+ * Design (per M7 spec):
+ *   - Workflow selection has two modes: explicit workflowId OR AI auto-select.
+ *   - When workflowId is provided, this function finds it in availableWorkflows.
+ *   - When no workflowId is provided, this function returns null — the Space
+ *     agent LLM is expected to call list_workflows first and pick explicitly.
+ *   - There is no tag-based matching, no keyword-based matching, and no
+ *     heuristic fallback. LLM reasoning replaces static heuristics.
  *
- * Note: The "space has default workflow" step was removed in Task 3.2 —
- * the Space system has no isDefault concept. Workflow selection is explicit
- * workflowId OR AI auto-select via this function.
+ * See: docs/plans/multi-agent-v2-customizable-agents-workflows/07-workflow-selection-intelligence.md
  */
 
 import type { SpaceWorkflow } from '@neokai/shared';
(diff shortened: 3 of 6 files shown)


lsm (commented):

**Recommendation: APPROVE**

Overall this is a clean, well-structured implementation. The pure function design for `selectWorkflow()` is solid, tests are thorough (36 tests, all passing), and the MCP tool wiring is correct. A few observations below.

### P2: Documentation inconsistency — tags JSDoc contradicts implementation

In `packages/shared/src/types/space.ts:557`, the `tags` field JSDoc says:

> "Tags for organizational categorization. **Not used for automatic workflow selection.**"

But `workflow-selector.ts` uses tags as Priority 2 in the selection chain (lines 100-124). This is a meaningful contradiction — the type definition explicitly says tags are not used for selection, yet the new selector relies on them as the primary heuristic. The JSDoc should be updated to reflect reality, or someone reading the type will assume tags are cosmetic and break the selector inadvertently.

### P3: Duplicated workflow loading pattern in `start_workflow_run`

`space-agent-tools.ts` lines 78-86 manually calls `workflowManager.listWorkflows(spaceId)` then `selectWorkflow()`. This duplicates the exact pattern that `SpaceRuntime.resolveWorkflowForRun()` (lines 326-331) was added for. The tool handler could call `runtime.resolveWorkflowForRun()` instead. Both paths are correct, but the duplication means a future change to resolution logic would need updating in two places.

### P3: `idCounter` in workflow-selector test is module-scoped, not reset between tests

`packages/daemon/tests/unit/space/workflow-selector.test.ts:22` — `idCounter` accumulates across test runs. This is harmless since IDs only need uniqueness, not specific values, and tests use explicit IDs via overrides where they matter. Just noting it.

### Positive observations

- The explicit-ID-not-found returning `null` instead of falling through to heuristics (line 93) is the right design choice — prevents surprising silent fallback.
- Error handling in tool handlers consistently wraps in try/catch and returns structured error JSON rather than throwing.
- Test fixtures use real SQLite DB with migrations, giving good integration coverage for the tool handlers.
- The `WorkflowSelectionContext` interface cleanly separates concerns — callers pre-filter workflows, the selector only picks.
- Type safety is clean — `bun run typecheck` passes with no errors.

lsm (commented):

---

## Summary

The implementation contains a fundamental design conflict with the current spec. The `workflow-selector.ts` file and its integration into `start_workflow_run` implement tag-based and keyword-based auto-selection heuristics that the spec document (`docs/plans/multi-agent-v2-customizable-agents-workflows/07-workflow-selection-intelligence.md`) explicitly removed. Additionally, two tools specified in the M7 agent prompt (`get_workflow_run`, `change_plan`) are absent.

---

## P0: Core Design Violates Spec

### Tag/Keyword heuristics contradict the M7 spec

**Files:** `packages/daemon/src/lib/space/runtime/workflow-selector.ts`, `packages/daemon/src/lib/space/tools/space-agent-tools.ts`

The M7 spec document states explicitly:

> "There is no default workflow, no tag-based matching, no keyword-based matching, and no fallback-to-null logic."
> "All of this has been removed. The Space agent's LLM reasoning replaces static heuristics."

The `SpaceWorkflow.tags` field comment in `packages/shared/src/types/space.ts` line 557 even says:

> "Tags for organizational categorization. Not used for automatic workflow selection."

Yet `workflow-selector.ts` implements a 3-level heuristic chain (tag-based → keyword/description → null fallback) and `start_workflow_run` in `space-agent-tools.ts` auto-selects using these heuristics when `workflow_id` is omitted.

Per spec, when no `workflowId` is provided, the agent should call `list_workflows` first and use LLM reasoning to pick one, then call `start_workflow_run` with an explicit `workflowId`. The tool should require `workflow_id`, not accept it as optional with server-side fallback logic.

The worker's file header at line 14-16 of `workflow-selector.ts` even acknowledges that the default-workflow concept was removed in Task 3.2, but then proceeds to implement tag-based and keyword-based matching that was removed in the same spec revision.

### Missing spec-required tools

The M7 spec agent prompt lists 5 tools:
- `list_workflows` — implemented
- `start_workflow_run` — implemented (but incorrectly)
- `get_workflow_run` — **not implemented**
- `change_plan` — **not implemented**
- `list_tasks` — implemented

`get_workflow_run` and `change_plan` are absent from `space-agent-tools.ts` entirely. The spec describes `change_plan` as allowing the agent to switch workflows mid-run or update task descriptions. The implementation adds `create_task` and `list_workflow_runs` which are not in the spec's tool list.

---

## P1: Test Covers Non-Spec Behavior Extensively

**File:** `packages/daemon/tests/unit/space/workflow-selector.test.ts`

The test suite has 321 lines and 18 tests verifying the tag-based and keyword-matching heuristics in detail. Since the heuristics themselves contradict the spec, these tests validate incorrect behavior. If the heuristics are removed per spec, these tests will need to be replaced entirely.

---

## P2: Unused Variable in Test

**File:** `packages/daemon/tests/unit/space/workflow-selector.test.ts`, lines 260-264

In the "returns null when no workflow tags or descriptions match" test, `ctx` is constructed but never used — `ctx2` is the one asserted. The `ctx` object with `title: 'aardvark quantum physics'` is dead code.

```ts
const ctx = makeContext({  // constructed but never asserted
    title: 'aardvark quantum physics',
    ...
});
const ctx2 = makeContext({ ... });
expect(selectWorkflow(ctx2)).toBeNull();
```

---

## P2: `start_workflow_run` Tool Schema Mismatch with Intended Design

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, line 209

`workflow_id` is marked `optional()` in the Zod schema. Under the correct spec design, `workflow_id` should be required — the LLM agent always calls `list_workflows` first and then calls `start_workflow_run` with an explicit ID. Making it optional enables the broken auto-selection path.

---

## P3: Tie-Breaking in Heuristics Is Input-Order Dependent

**File:** `packages/daemon/src/lib/space/runtime/workflow-selector.ts`, lines 117-119

When two workflows share the same best score, the first one in `availableWorkflows` wins because the condition uses `score > bestTagScore` (strict greater-than). While the DB query uses `ORDER BY created_at ASC` ensuring stable ordering, this implicit tie-breaking is undocumented. If the code is kept, a comment or dedicated tie-breaking rule would make this explicit.

---

## Positive Observations

- The `resolveWorkflowForRun()` integration point in `SpaceRuntime` (lines 326-331) is thin and delegates cleanly to the pure function — a good pattern for testability.
- The `createSpaceAgentToolHandlers()` separation from the MCP server factory is excellent for testing.
- The `space-agent-tools.test.ts` integration tests using a real SQLite DB are thorough for the tools that match the spec (`list_tasks`, `list_workflows`, `list_workflow_runs`, `create_task`).
- Error handling in tool handlers is consistent — errors become `{ success: false, error: message }` JSON responses rather than thrown exceptions.
- All TypeScript types check cleanly (typecheck passes with no errors).

lsm (commented):

---

## Summary

The refactor addresses all the major issues from the previous review: heuristic auto-selection has been removed, `workflow_id` is now required in `start_workflow_run`, the missing `get_workflow_run` and `change_plan` tools have been added, and the non-spec `create_task`/`list_workflow_runs` tools have been removed. Tests have been rewritten accordingly. The overall design is now aligned with the M7 spec. Two remaining issues below.

---

## P2: Stale JSDoc on `resolveWorkflowForRun` references removed heuristics

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 317-322

The JSDoc for `resolveWorkflowForRun()` still says:

> "Accepts an optional explicit `workflowId`. When omitted, auto-selects via tag-based and keyword-based heuristics. Returns null when no workflow matches (caller should create a standalone task instead)."

This is a leftover from the previous iteration. The heuristic logic was removed — `selectWorkflow()` now returns null when no `workflowId` is provided, and there are no tag-based or keyword-based heuristics. The JSDoc should be updated to reflect the current behavior: explicit ID lookup or null.

This matters because misleading documentation on integration points will confuse future developers about the design intent.

---

## P2: `change_plan` workflow switch can leave the system in a broken state

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, lines 151-173

When switching workflows via `change_plan`, the handler cancels the old run (line 153) *before* attempting to start the new run (line 157). If `runtime.startWorkflowRun()` throws (e.g., invalid `workflow_id`, DB error), the old run is already cancelled and no new run exists — the user's work is orphaned with no active run.

The fix is straightforward: start the new run first, and only cancel the old run after the new one succeeds. Alternatively, wrap both operations in a transaction so the cancellation is rolled back on failure.

The test suite covers the error case for `start_workflow_run` with an invalid ID (line 223-232), but there is no test for the `change_plan` workflow-switch failure path where the new workflow ID is invalid. Adding such a test would demonstrate the issue.

---

## Positive Observations

- The `selectWorkflow()` function is now a clean, pure explicit-ID resolver with no heuristics — exactly what the M7 spec calls for.
- `createSpaceAgentToolHandlers()` separation from the MCP server factory remains an excellent pattern for testability.
- The test suite is comprehensive: 26 tests covering all 5 tools, including edge cases like missing runs, completed runs, and empty task lists.
- Error handling is consistent across all tool handlers — errors return structured JSON responses rather than thrown exceptions.
- The `get_workflow_run` handler includes current step enrichment and associated tasks, giving the LLM agent a complete picture of run state.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (P1 issues remain)

---

All previous P0 issues from the first review have been properly addressed:

- workflow-selector.ts is now a pure explicit-ID resolver with no heuristics
- `get_workflow_run` and `change_plan` are implemented
- `workflow_id` is required in `start_workflow_run`
- `create_task` and `list_workflow_runs` are correctly absent
- Tests are rewritten to validate the spec-correct 5-tool set

Good work on those. The remaining issues are below.

---

### P1: `change_plan` workflow switch has a non-atomic cancel/start that silently destroys run state on failure

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, lines 152-173

The current logic cancels the old run before starting the new one:

```ts
workflowRunRepo.updateStatus(run.id, 'cancelled');  // run is now gone

try {
    const { run: newRun, tasks } = await runtime.startWorkflowRun(...);
    ...
} catch (err) {
    return jsonResult({ success: false, error: message });
    // old run is permanently cancelled, no new run exists
}
```

If `startWorkflowRun` throws (e.g., invalid target `workflow_id`, or the space is not found), the tool returns `{ success: false, error: ... }` but the original run is permanently in `cancelled` status with no replacement. From the LLM's perspective the work is simply gone. There is no recovery path, no rollback, and no indication to the agent that the original run was lost.

This is not an atomicity problem that requires transactions — the simplest fix is to validate that the target workflow exists before cancelling the old run.

### P1: No test covers the partial failure path in `change_plan` workflow switch

**File:** `packages/daemon/tests/unit/space/space-agent-tools.test.ts`

The existing test at line 325 covers the happy path (valid `wf2.id`). There is no test that passes an invalid `workflow_id` to `change_plan` and asserts on the resulting state of the original run. This path is where the data loss happens (see above).

A test like this is needed:

```ts
test('does not cancel the original run when target workflow_id is invalid', async () => {
    // start a run on wf1, then try to switch to a non-existent workflow
    // assert: success is false AND the original run is still in_progress (not cancelled)
});
```

---

### P2: Stale JSDoc on `resolveWorkflowForRun` describes heuristics that no longer exist

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 316-325

The doc comment was copied from the old heuristic design:

```
* Accepts an optional explicit `workflowId`. When omitted, auto-selects via
* tag-based and keyword-based heuristics. Returns null when no workflow matches
* (caller should create a standalone task instead).
```

Per the current implementation, when `workflowId` is omitted the function always returns null — there are no tag-based or keyword-based heuristics. The comment should describe the actual behavior: returns null when no `workflowId` is provided (LLM must call `list_workflows` first) or when the explicit ID is not found.

### P2: `WorkflowSelectionContext` carries dead fields `title` and `description`

**File:** `packages/daemon/src/lib/space/runtime/workflow-selector.ts`, lines 30-34

The interface declares `title: string` and `description: string`, but `selectWorkflow()` never reads them — only `workflowId` and `availableWorkflows` are used. Because the spec says LLM reasoning replaces heuristics, these fields will never be used by this function.

Carrying unused fields in the interface creates confusion (it implies the function does something with them) and forces every caller to provide dummy values. Either remove them from the interface, or add a code comment explicitly explaining why they are present (e.g., for future extensibility, or for caller context logging).

`resolveWorkflowForRun` in `space-runtime.ts` (line 326) accepts `Omit<WorkflowSelectionContext, 'availableWorkflows'>` which means callers must still pass `title` and `description` even though they are ignored.

---

### P3: `jsonResult` helper restricts to `Record<string, unknown>` but callers pass nested objects

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, line 54

Minor: the `data` parameter type of `jsonResult` is `Record<string, unknown>` which works in practice but does not accept arrays as top-level values. Not a current bug since all call sites pass objects, but the type could be widened to `unknown` for defensive correctness given it immediately delegates to `JSON.stringify`.

lsm (commented):

All four previously raised issues have been properly addressed in commit `364494b`:

1. **P1 (change_plan non-atomic cancel/start)** -- Fixed. `workflowManager.getWorkflow()` is now called BEFORE `workflowRunRepo.updateStatus(run.id, 'cancelled')`. If the target workflow does not exist, an error is returned immediately and the original run remains `in_progress`. A dedicated test (`does not cancel the original run when target workflow_id is invalid`) verifies this behavior.

2. **P2 (Stale JSDoc on resolveWorkflowForRun)** -- Fixed. The docstring now accurately describes "explicit workflowId" lookup with null return cases, matching the actual implementation.

3. **P2 (Dead title/description fields in WorkflowSelectionContext)** -- Fixed. The `title` and `description` fields were removed from the interface. `resolveWorkflowForRun` signature was simplified to `(spaceId: string, workflowId?: string)`. All test fixtures (`makeContext`) and assertions updated accordingly.

4. **P3 (jsonResult type)** -- Fixed. Parameter type changed from `Record<string, unknown>` to `unknown`.

No new issues found. The implementation is clean, well-tested, and follows the M7 spec's design of LLM-driven workflow selection with no server-side heuristics. Test coverage is thorough across all five tool handlers with appropriate edge cases.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

---

All five issues raised in the previous review have been correctly resolved in commit `364494b5`.

### P1: change_plan non-atomic cancel/start — RESOLVED

`/packages/daemon/src/lib/space/tools/space-agent-tools.ts` lines 153-159 now validate that the target workflow exists via `workflowManager.getWorkflow(args.workflow_id)` before ever calling `workflowRunRepo.updateStatus(run.id, 'cancelled')`. An invalid `workflow_id` returns `{ success: false, error: 'Workflow not found: ...' }` with the original run completely untouched.

### P1: Missing test for partial failure path — RESOLVED

Test at `/packages/daemon/tests/unit/space/space-agent-tools.test.ts` lines 389-414 starts a run, calls `change_plan` with a non-existent `workflow_id`, asserts `success: false`, and then re-reads the original run from the DB to confirm its status is still `in_progress`. This is exactly the right assertion.

### P2: Stale JSDoc on resolveWorkflowForRun — RESOLVED

The JSDoc at `/packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 315-328 now accurately describes the simplified two-return-case behavior: null when no `workflowId` is provided (LLM must call `list_workflows` first) and null when the provided ID is not found in the space. No mention of heuristics.

### P2: Dead title/description fields in WorkflowSelectionContext — RESOLVED

`/packages/daemon/src/lib/space/runtime/workflow-selector.ts` `WorkflowSelectionContext` now contains only `spaceId`, `availableWorkflows`, and optional `workflowId`. The dead fields are gone. `resolveWorkflowForRun` signature was simplified to `(spaceId: string, workflowId?: string)` and the now-unused `WorkflowSelectionContext` import was removed from `space-runtime.ts`.

### P3: jsonResult type widened — RESOLVED

`jsonResult` at `/packages/daemon/src/lib/space/tools/space-agent-tools.ts` line 54 now accepts `unknown` instead of `Record<string, unknown>`.

---

No new issues were introduced by the fix commit. The implementation is clean and complete.

PR #454, opened 2026-03-20
feat: SpaceRuntimeService, workflow run RPC handlers, and session group metadata (Task 4.3)

lsm:
## Summary

- **SpaceRuntimeService** (`space-runtime-service.ts`): Wraps the shared `SpaceRuntime` with a per-space lifecycle API — `createOrGetRuntime(spaceId)` validates the space, auto-starts the runtime, and returns it; `stopRuntime(spaceId)` is reserved for future per-space isolation
- **`spaceWorkflowRun.*` RPC handlers** (`space-workflow-run-handlers.ts`): `start` creates a run via SpaceRuntimeService + SpaceRuntime, `list` supports optional status filter, `get` fetches by ID, `cancel` cancels pending tasks then the run, emits `space.workflowRun.updated`
- **Migration 33**: Adds `workflow_id` and `current_step_id` columns to `space_session_groups` for UI display of workflow context
- **SpaceSessionGroup type/repo update**: `workflowId` and `currentStepId` fields wired through `CreateSessionGroupParams`, `UpdateSessionGroupParams`, `createGroup()`, `updateGroup()`, and `rowToGroup()`
- **`workflowStepName` in task events**: Added optional `workflowStepName` to `space.task.created` and `space.task.updated` event payloads in `DaemonEventMap`
- **`SpaceTaskManager.listTasksByWorkflowRun()`**: Exposes the task repo's `listByWorkflowRun` at the manager layer
- **`DaemonAppContext.spaceRuntimeService`**: Service surfaced via `RPCHandlerSetupResult` and wired into `app.ts`
- **36 unit tests**: 13 for SpaceRuntimeService lifecycle/createOrGetRuntime, 23 for all 4 `spaceWorkflowRun.*` handlers

## Test plan

- [ ] All pre-commit checks pass (lint, format, typecheck, knip)
- [ ] `bun test packages/daemon/tests/unit/space/space-runtime-service.test.ts` — 13 tests pass
- [ ] `bun test packages/daemon/tests/unit/rpc-handlers/space-workflow-run-handlers.test.ts` — 23 tests pass
- [ ] Full unit test suite: `bun test packages/daemon/tests/unit/` — pre-existing failures only (bun-node-wrapper PATH test, unrelated to this PR)

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -18,6 +18,7 @@ import { LiveQueryEngine } from './storage/live-query';
 import { SpaceAgentRepository } from './storage/repositories/space-agent-repository';
 import { SpaceAgentManager } from './lib/space/managers/space-agent-manager';
 import { SpaceManager } from './lib/space/managers/space-manager';
+import type { SpaceRuntimeService } from './lib/space/runtime/space-runtime-service';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -57,6 +58,8 @@ export interface DaemonAppContext {
 	spaceAgentManager: SpaceAgentManager;
 	/** Space manager for Space CRUD and workspace path validation */
 	spaceManager: SpaceManager;
+	/** Space runtime service for workflow run lifecycle management */
+	spaceRuntimeService: SpaceRuntimeService;
 	/**
 	 * Cleanup function for graceful shutdown.
 	 * Closes all connections, stops sessions, and closes database.
@@ -206,8 +209,8 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		logInfo('[Daemon] GitHub integration disabled - authentication required');
 	}
 
-	// Setup RPC handlers (returns cleanup function)
-	const rpcHandlerCleanup = setupRPCHandlers({
+	// Setup RPC handlers (returns cleanup function + exposed services)
+	const { cleanup: rpcHandlerCleanup, spaceRuntimeService } = setupRPCHandlers({
 		messageHub,
 		sessionManager,
 		authManager,
@@ -435,6 +438,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		liveQueries,
 		spaceAgentManager,
 		spaceManager,
+		spaceRuntimeService,
 		cleanup,
 	};
 }
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -418,14 +418,12 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		spaceId: string;
 		taskId: string;
 		task: import('@neokai/shared').SpaceTask;
-		workflowStepName?: string;
 	};
 	'space.task.updated': {
 		sessionId: string;
 		spaceId: string;
 		taskId: string;
 		task: import('@neokai/shared').SpaceTask;
-		workflowStepName?: string;
 	};
 
 	// Space workflow run events (global events - use 'global' as sessionId)
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -418,12 +418,14 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		spaceId: string;
 		taskId: string;
 		task: import('@neokai/shared').SpaceTask;
+		workflowStepName?: string;
 	};
 	'space.task.updated': {
 		sessionId: string;
 		spaceId: string;
 		taskId: string;
 		task: import('@neokai/shared').SpaceTask;
+		workflowStepName?: string;
 	};
 
 	// Space workflow run events (global events - use 'global' as sessionId)
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -56,7 +56,9 @@ import { setupSpaceAgentHandlers } from './space-agent-handlers';
 import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
 import { SpaceWorkflowRepository } from '../../storage/repositories/space-workflow-repository';
 import { SpaceAgentRepository } from '../../storage/repositories/space-agent-repository';
-import { SpaceRuntime } from '../space/runtime/space-runtime';
+import { SpaceRuntimeService } from '../space/runtime/space-runtime-service';
+import { setupSpaceWorkflowRunHandlers } from './space-workflow-run-handlers';
+import type { SpaceWorkflowRunTaskManagerFactory } from './space-workflow-run-handlers';
 import { setupSpaceExportImportHandlers } from './space-export-import-handlers';
 
 export interface RPCHandlerDependencies {
@@ -81,10 +83,19 @@ const log = new Logger('rpc-handlers');
 export type RPCHandlerCleanup = () => void;
 
 /**
- * Register all RPC handlers on MessageHub
- * Returns a cleanup function that should be called to stop background services
+ * Result returned by setupRPCHandlers — includes both the cleanup function
+ * and any services that need to be surfaced in DaemonAppContext.
  */
-export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanup {
+export interface RPCHandlerSetupResult {
+	cleanup: RPCHandlerCleanup;
+	spaceRuntimeService: SpaceRuntimeService;
+}
+
+/**
+ * Register all RPC handlers on MessageHub
+ * Returns a result with cleanup function and exposed services
+ */
+export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupResult {
 	// Room handlers (create roomManager first as session handlers depend on it)
 	const roomManager = new RoomManager(deps.db.getDatabase());
 
@@ -224,8 +235,8 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanu
 		deps.daemonHub
 	);
 
-	// Space Runtime — workflow orchestration tick loop
-	const spaceRuntime = new SpaceRuntime({
+	// Space Runtime Service — wraps SpaceRuntime with per-space lifecycle API
+	const spaceRuntimeService = new SpaceRuntimeService({
 		db: deps.db.getDatabase(),
 		spaceManager: deps.spaceManager,
 		spaceAgentManager: deps.spaceAgentManager,
@@ -233,7 +244,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanu
 		workflowRunRepo: spaceWorkflowRunRepo,
 		taskRepo: spaceTaskRepo,
 	});
-	spaceRuntime.start();
+	spaceRuntimeService.start();
 
 	// Space export/import handlers
 	setupSpaceExportImportHandlers(
@@ -245,9 +256,26 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerCleanu
 		deps.db.getDatabase()
 	);
 
-	// Return cleanup function to stop background services
-	return () => {
-		roomRuntimeService.stop();
-		spaceRuntime.stop();
+	// Space workflow run handlers — reuse the same factory pattern as spaceTask handlers
+	const spaceWorkflowRunTaskManagerFactory: SpaceWorkflowRunTaskManagerFactory = (spaceId) => {
+		return new SpaceTaskManager(deps.db.getDatabase(), spaceId);
+	};
+	setupSpaceWorkflowRunHandlers(
+		deps.messageHub,
+		deps.spaceManager,
+		spaceWorkflowManager,
+		spaceWorkflowRunRepo,
+		spaceRuntimeService,
+		spaceWorkflowRunTaskManagerFactory,
+		deps.daemonHub
+	);
+
+	// Return result with cleanup function and exposed services
+	return {
+		cleanup: () => {
+			roomRuntimeService.stop();
+			spaceRuntimeService.stop();
+		},
+		spaceRuntimeService,
 	};
 }
--- a/packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts
@@ -14,12 +14,14 @@ import type { SpaceManager } from '../space/managers/space-manager';
 import type { SpaceWorkflowManager } from '../space/managers/space-workflow-manager';
 import type { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
 import type { SpaceRuntimeService } from '../space/runtime/space-runtime-service';
+import type { SpaceTaskManager } from '../space/managers/space-task-manager';
 import type { WorkflowRunStatus } from '@neokai/shared';
-import { SpaceTaskManager } from '../space/managers/space-task-manager';
-import type { Database as BunDatabase } from 'bun:sqlite';
 import { Logger } from '../logger';
 
 const log = new Logger('space-workflow-run-handlers');
+
+/** Factory that creates a SpaceTaskManager bound to a specific spaceId. */
+export type SpaceWorkflowRunTaskManagerFactory = (spaceId: string) => SpaceTaskManager;
 
 export function setupSpaceWorkflowRunHandlers(
 	messageHub: MessageHub,
@@ -27,7 +29,7 @@ export function setupSpaceWorkflowRunHandlers(
 	spaceWorkflowManager: SpaceWorkflowManager,
 	workflowRunRepo: SpaceWorkflowRunRepository,
 	spaceRuntimeService: SpaceRuntimeService,
-	db: BunDatabase,
+	taskManagerFactory: SpaceWorkflowRunTaskManagerFactory,
 	daemonHub: DaemonHub
 ): void {
 	// ─── spaceWorkflowRun.start ──────────────────────────────────────────────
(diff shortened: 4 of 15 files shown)


lsm (commented):

This is a well-structured PR that cleanly implements SpaceRuntimeService, workflow run RPC handlers, the DB migration, and supporting type changes. The code follows the established patterns in the codebase and has good test coverage. A few observations:

### What looks good

1. **SpaceRuntimeService** (`packages/daemon/src/lib/space/runtime/space-runtime-service.ts`) - Clean lifecycle wrapper around the shared SpaceRuntime. The single-runtime-per-daemon design with per-space validation is sensible and well-documented. Idempotent start/stop is correct.

2. **RPC handlers** (`packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts`) - All four handlers follow the existing pattern (input validation, space existence check, action, event emission). The cancel handler correctly handles all terminal states (already cancelled -> no-op, completed -> error).

3. **Migration 33** - Idempotent column additions using try/catch on SELECT probes, consistent with the project's migration pattern.

4. **WorkflowGateError** - Good separation of human-gate blocking from general transition failures. The workflow executor tests were correctly updated to expect the new error type.

5. **Test coverage** - 36 tests covering service lifecycle, all RPC handler paths including edge cases. The stub DB pattern in the cancel handler tests is pragmatic given the handler instantiates SpaceTaskManager directly.

1. **P3** - `space-workflow-run-handlers.ts:135`: The cancel handler instantiates `new SpaceTaskManager(db, run.spaceId)` inline rather than receiving a factory or instance via dependency injection, as is done for the task handlers (`SpaceTaskManagerFactory`). This works fine but deviates slightly from the DI pattern used elsewhere. Not a problem for correctness.

2. **P3** - `space-workflow-run-handlers.ts:75-84`: The fire-and-forget `.catch()` pattern for event emission is consistent with the codebase (e.g., `space-task-handlers.ts`), but silently swallowed event failures could make debugging harder in production. This is an existing pattern so not something to change in this PR.

3. **P3** - The `SpaceRuntimeService.createOrGetRuntime()` performs a `getSpace()` call on every invocation for validation. For high-frequency usage this could be redundant, but for an RPC-triggered method this is appropriate.

### Verdict

The implementation correctly delivers all items described in the task. The architecture is consistent with the existing codebase, types are properly threaded through shared/daemon layers, and tests verify the expected behavior including error paths. No bugs, security issues, or missing requirements found.

lsm (commented):

> **Recommendation: APPROVE** (with minor findings noted below)

The implementation is well-structured and achieves the stated goals. The service lifecycle, RPC handlers, migration, and tests are all present and follow established codebase patterns. Below are specific issues worth tracking.

---

### P1 — `workflowStepName` is declared but never populated

**File:** `packages/daemon/src/lib/daemon-hub.ts` lines 421, 428
**File:** `packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts` lines 58-66, 170-177

The `workflowStepName?` field was added to both `space.task.created` and `space.task.updated` event payloads in the `DaemonEventMap`, but the actual emitters in `space-task-handlers.ts` never populate it — the field is always absent. There are no tests verifying it appears in events. This is described in the PR as added functionality, but it's dead type surface with no runtime effect and no consumer. Either the emitters should populate it (by resolving the step name from `task.workflowStepId` via the workflow definition), or the field should be removed until it's actually wired up.

---

### P2 — `spaceWorkflowRun.cancel` instantiates `SpaceTaskManager` directly instead of using the factory pattern

**File:** `packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts` line 135

```ts
const taskManager = new SpaceTaskManager(db, run.spaceId);
```

Every other handler in the codebase that needs task management uses an injected factory (e.g., `SpaceTaskManagerFactory` in `space-task-handlers.ts`, `SpaceTaskManager` factory in `setupRPCHandlers`). This `cancel` handler bypasses that pattern by directly constructing `SpaceTaskManager` from the raw `db` handle, which is why the test had to build an elaborate `createStubDb` mock instead of a simple manager mock. Consistency with the factory pattern would make the handler testable without raw DB stubs and consistent with how the rest of the domain is wired.

---

### P2 — Double space existence validation in `spaceWorkflowRun.start`

**File:** `packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts` lines 46-47 and `space-runtime-service.ts` line 67

`spaceWorkflowRun.start` calls `spaceManager.getSpace()` explicitly (lines 46-47) to verify the space exists before calling `spaceRuntimeService.createOrGetRuntime()`, which then calls `spaceManager.getSpace()` again internally. Two async DB lookups for the same space per `start` call. The outer check is redundant since `createOrGetRuntime` already throws `'Space not found'` for unknown spaces. If the handler-level check is intentional (for a cleaner early-fail message before workflow resolution), the comment in the code should make this explicit.

---

### P2 — `spaceWorkflowRun.get` has no space-ownership check

**File:** `packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts` lines 107-116

The `get` handler fetches a workflow run by `id` alone without verifying the caller has access to the space that owns the run. Any client that knows a run UUID can retrieve it. Compare `spaceWorkflowRun.list` which requires `spaceId` and validates space existence. While all NeoKai clients are implicitly trusted (no user-level auth), enforcing consistent access patterns here avoids unexpected data leakage if multi-tenant features are ever added.

---

### P2 — `SpaceSessionGroup.workflowId` naming is semantically ambiguous

**File:** `packages/shared/src/types/space.ts` lines 319-320
**File:** `packages/daemon/src/storage/repositories/space-session-group-repository.ts`

The JSDoc says the field associates a session group with a "workflow run" for UI display, but the field is named `workflowId` (a workflow *definition* ID), not `workflowRunId`. This is inconsistent: the migration adds column `workflow_id` and the type comment says "ID of the workflow run". This naming confusion will make it hard for future developers to distinguish between `SpaceWorkflow` (definition) and `SpaceWorkflowRun` (execution). The field should be renamed to `workflowRunId` in the type, repository, and migration (while migration 33 is still fresh and before it ships), or the JSDoc should be corrected to clarify it stores the workflow *definition* ID.

---

### P3 — Missing test for the task-cancellation path in `spaceWorkflowRun.cancel`

**File:** `packages/daemon/tests/unit/rpc-handlers/space-workflow-run-handlers.test.ts` lines 503-517

The "cancels the run and emits space.workflowRun.updated" test uses `tasks: []` (empty task list) to sidestep the DB stub complexity. This means the actual task-cancellation loop (lines 136-143 in the handler) is never exercised in any test. The elaborate `createStubDb` mock (lines 176-238) exists but is only invoked with an empty array. Adding at least one test scenario with a pending and an in_progress task would verify the cancellation path actually works end-to-end.

---

### P3 — `spaceRuntimeService.start()` called without error handling in `setupRPCHandlers`

**File:** `packages/daemon/src/lib/rpc-handlers/index.ts` line 245

`roomRuntimeService.start()` is wrapped in `.catch()` (line 146-148) but `spaceRuntimeService.start()` is called bare (line 245). Since `start()` is synchronous, this is not a crash risk today, but if the signature ever changes to async the inconsistency could mask errors. Minor but worth aligning.

---

**Summary:** The core implementation is correct and the 36 tests cover the main happy-path and error cases well. The primary concern is `workflowStepName` being present in the event type but never populated in any emitter — this is either an incomplete feature or dead code that should be removed. The factory-bypass in the cancel handler and the `workflowId` vs `workflowRunId` naming ambiguity are worth fixing before this area grows further.

lsm (commented):

All five issues from the previous review have been correctly addressed:

**P1 (resolved): `workflowStepName` removed from `DaemonEventMap`**
Confirmed removed from both `space.task.created` and `space.task.updated` event types in `daemon-hub.ts`. No references remain in application code (only in planning docs, which is expected).

**P2 (resolved): `workflowId` renamed to `workflowRunId`**
Consistently renamed across:
- `SpaceSessionGroup` type (`packages/shared/src/types/space.ts:319`)
- `SpaceSessionGroupRepository` (`packages/daemon/src/storage/repositories/space-session-group-repository.ts`) - all CRUD methods use `workflowRunId`
- Migration 33 comment now clarifies it stores a run execution ID
- Test helper `space-test-db.ts` column definition updated
- `SpaceTask` type already had the correct name

**P2 (resolved): `SpaceWorkflowRunTaskManagerFactory` replaces `new SpaceTaskManager(db)`**
The cancel handler at `space-workflow-run-handlers.ts:144` now uses the injected `taskManagerFactory(run.spaceId)` instead of directly constructing a `SpaceTaskManager`. The `createStubDb` mock is completely gone from the codebase.

**P2 (resolved): Early `getSpace` check explained**
Comment at `space-workflow-run-handlers.ts:47-49` explains why the early space validation is intentional (prevents misleading "No workflows found" error for nonexistent spaces).

**P2 (resolved): Optional `spaceId` on `get` handler**
`spaceWorkflowRun.get` at line 112 accepts optional `spaceId` parameter and performs an ownership check when provided (`space-workflow-run-handlers.ts:120-122`).

**P3 (resolved): Task-cancellation tests added**
Three new tests cover:
- Cancelling only pending/in_progress tasks, not completed (`space-workflow-run-handlers.test.ts:484-514`)
- Error isolation - continues cancelling remaining tasks when one fails (`space-workflow-run-handlers.test.ts:516-559`)
- Factory called with correct spaceId (`space-workflow-run-handlers.test.ts:499-500`)

### Test Results

All 88 tests pass across the 3 test files (`space-workflow-run-handlers.test.ts`, `space-runtime-service.test.ts`, `workflow-executor.test.ts`).

Note: `SpaceRuntimeService` tests log background errors about `listSpaces is not a function` - this is a pre-existing issue from Task 4.2 where the mock `spaceManager` doesn't include `listSpaces`. The error is swallowed by the runtime's `.catch()` handler and doesn't affect test outcomes. Not blocking for this PR.

All previous review feedback has been correctly addressed. The implementation is clean, tests are thorough, and no new issues were introduced.

lsm (commented):

> **Recommendation: APPROVE** — All previous feedback has been properly addressed.

---

### Summary of Previous Issues Addressed

All P1, P2, and P3 issues from the prior review have been correctly resolved in the fix commit.

**P1 — workflowStepName removed from DaemonEventMap**

Confirmed: `space.task.created` and `space.task.updated` entries in `packages/daemon/src/lib/daemon-hub.ts` lines 416-427 no longer include `workflowStepName`. The dead type surface is gone.

**P2 — workflowId renamed to workflowRunId throughout**

Confirmed across all layers:
- `packages/shared/src/types/space.ts` — `SpaceSessionGroup.workflowRunId`
- `packages/daemon/src/storage/repositories/space-session-group-repository.ts` — column `workflow_run_id` in SQL and `workflowRunId` in `rowToGroup`
- `packages/daemon/src/storage/schema/migrations.ts` — Migration 33 adds `workflow_run_id`
- `packages/daemon/tests/unit/helpers/space-test-db.ts` — test helper DDL uses `workflow_run_id`

**P2 — Factory pattern used in cancel handler**

Confirmed: `spaceWorkflowRun.cancel` now receives `SpaceWorkflowRunTaskManagerFactory` as a parameter (line 32-33 of handler file). Direct `SpaceTaskManager` instantiation is gone. Tests mock the factory cleanly without a raw DB stub.

**P2 — Early space validation comment added**

Confirmed: Lines 47-50 of `space-workflow-run-handlers.ts` now include an explanatory comment clarifying why the outer `getSpace` check is intentional (listWorkflows returns `[]` for nonexistent spaces, yielding a misleading error without the guard).

**P2 — Optional spaceId ownership check in spaceWorkflowRun.get**

Confirmed: Lines 119-122 of `space-workflow-run-handlers.ts` perform a cross-space rejection when `spaceId` is supplied. Test coverage added for both matching and mismatching spaceId cases.

**P3 — Task-cancellation tests added**

Confirmed: Two new tests in the cancel suite:
1. `cancels pending and in_progress tasks before cancelling the run` — exercises the task loop with pending, in_progress, and completed tasks; verifies only cancellable statuses are targeted.
2. `continues cancelling remaining tasks even if one cancelTask fails` — verifies error isolation in the task loop.

All 28 handler tests and 13 service tests pass (verified locally).

---

### No New Issues Found

The fix commit introduces no new concerns. The implementation is clean, consistent with existing patterns, and fully tested.

One minor observation (not blocking): the test for the restart-after-stop scenario in `space-runtime-service.test.ts` triggers a logged warning (`SpaceRuntime: initial tick failed: listSpaces is not a function`) because the mock `spaceManager` only has `getSpace` and not `listSpaces`. This is expected and harmless — the error is caught inside the runtime and does not affect test correctness — but future readers might find the log noise surprising. The existing 13 tests all pass regardless.

lsm (commented):

---

### Summary

The three changes targeted in this re-review are all correct. Migration numbering is clean, the `WorkflowGateError` inheritance fix is sound, and all gate enforcement tests now pass.

---

### Migration 30 — numbering is correct

`dev` ends at `runMigration29` (the consolidated Space tables migration). This PR appends `runMigration30` in a single sequential call in `runMigrations()`. No numbering gaps, no conflicts. The idempotent column-existence check (`SELECT … LIMIT 1` inside a try/catch) is consistent with the pattern already used in migration 29 for nullable columns.

---

### `WorkflowGateError` inheritance fix — correct and complete

`WorkflowGateError extends WorkflowTransitionError` (line 74, `workflow-executor.ts`). The catch block in `space-runtime.ts` line 490 is:

```ts
if (!(err instanceof WorkflowTransitionError)) {
    throw err;
}
// Gate blocked: run status already set to 'needs_attention' by the executor.
// Keep executor and meta in map — will be retried after gate is resolved.
```

Because `WorkflowGateError` is a subclass of `WorkflowTransitionError`, `instanceof WorkflowTransitionError` evaluates to `true` for gate errors, so they are correctly swallowed here and the executor is retained in the map. Before the fix, `WorkflowGateError` extended plain `Error`, so it fell through to the `throw err` path, propagating the error and breaking the gate-retention behavior.

The three previously failing gate enforcement tests in `space-runtime.test.ts` (`gate enforcement` describe block) now all pass. The full suite of 353 tests passes cleanly.

---

### P2 — Mock gap in `space-runtime-service.test.ts`

The `createMockSpaceManager` helper in `/packages/daemon/tests/unit/space/space-runtime-service.test.ts` only mocks `getSpace`:

```ts
function createMockSpaceManager(space: Space | null = mockSpace): SpaceManager {
    return {
        getSpace: mock(async () => space),
    } as unknown as SpaceManager;
}
```

Several tests then call `service.start()` (or `createOrGetRuntime()` which auto-starts), which immediately fires `SpaceRuntime.rehydrateExecutors()` on the first tick — and that internally calls `listSpaces(false)`. Since `listSpaces` is not mocked, it throws `TypeError: this.config.spaceManager.listSpaces is not a function`, which is caught and logged by the tick error handler. Tests still pass (the error never surfaces as a test failure), but the test run emits two error-level log lines to stderr, which:

1. Creates noise in CI output that looks like real failures on first glance.
2. Masks any future bug where the error handler silently swallows something it shouldn't.

The fix is to add `listSpaces: mock(async () => [])` to the mock object in `createMockSpaceManager`.

---

- `SpaceRuntimeService` lifecycle (`start/stop` idempotency, `createOrGetRuntime` auto-start, `stopRuntime` no-op) is well-tested and correct.
- RPC handlers (`spaceWorkflowRun.start/list/get/cancel`) have proper input validation, ownership checks, and correct task cancellation on run cancel.
- The `SpaceSessionGroup` repository changes (`workflowRunId`, `currentStepId` columns) align with the shared types in `@neokai/shared` and the migration schema.
- `space/index.ts` correctly re-exports `WorkflowGateError` alongside `WorkflowTransitionError`.
- Typecheck passes cleanly.

lsm (commented):

**Recommendation: APPROVE**

### Fix Verification

The P2 issue from the previous review has been correctly resolved. The root cause was that `SpaceRuntime.start()` fires an immediate async call to `executeTick()`, which calls `rehydrateExecutors()`, which in turn calls `this.config.spaceManager.listSpaces(false)` (space-runtime.ts line 359). The mock in `createMockSpaceManager` omitted `listSpaces`, so every test that triggered `start()` produced a swallowed error log via the `.catch()` in the tick loop.

The fix — adding `listSpaces: mock(async () => [])` at line 45 of the test file — directly addresses this. Returning an empty array is semantically correct: rehydration iterates over spaces and finds no in-progress runs to reload, which is the right no-op behavior for a unit test.

### One Pre-existing Minor Note (not introduced by this PR)

The inline `space2Manager` object in the test at lines 113–117 also lacks `listSpaces`. It would produce the same swallowed error on the async tick from `start()`. This is a pre-existing minor issue carried forward from before this fix and is outside the scope of the change being reviewed — the test still passes because `createOrGetRuntime()` resolves synchronously before the async tick error fires. This is worth a follow-up cleanup but is not a blocker.

### Summary

The targeted fix is correct, complete, and does not introduce any regressions. All 13 tests remain passing and the error log noise is eliminated.

lsm (commented):

**Recommendation: APPROVE**

The migration consolidation was correctly implemented. All requested changes are present and correct.

---

### Migration Consolidation Verification

**workflow_run_id and current_step_id in migration 29** — Confirmed. Lines 1669–1670 of `packages/daemon/src/storage/schema/migrations.ts` show both columns inside the `CREATE TABLE IF NOT EXISTS space_session_groups` statement within `runMigration29`.

**runMigration30 fully removed** — Confirmed. No function definition, no call site, and no orphaned references exist anywhere in the codebase. `grep -rn 'runMigration3[0-9]'` returns nothing.

**No gap in migration numbering** — Confirmed. The `runMigrations()` function ends at `runMigration29(db)` with no skip. The sequence 1–29 is complete (migration 12 is exported rather than private, which is pre-existing and unrelated to this PR).

**SpaceSessionGroupRepository works correctly** — Confirmed. `createGroup()` inserts both columns (line 37), `updateGroup()` conditionally sets both (lines 116–123), and `rowToGroup()` maps both columns back to the interface fields (lines 241–242). The test helper `space-test-db.ts` also includes both columns in its in-memory schema.

---

### PR Description Stale (minor)

The PR body says "**Migration 33**: Adds `workflow_id` and `current_step_id` columns" which is doubly stale after the consolidation commit — the number should now be 29, and the column name is `workflow_run_id` not `workflow_id`. Since this is the same author this is low severity, but worth noting for changelog clarity.

---

### P3 — Test gap for new fields in SpaceSessionGroupRepository

`packages/daemon/tests/unit/storage/space-session-group-repository.test.ts` does not exercise the new `workflowRunId` / `currentStepId` fields through `createGroup()` or `updateGroup()`. The existing tests will continue to pass (the columns are optional and default to NULL), but a round-trip test would give stronger guarantees. Not a blocker given the repository code is straightforward and the SpaceRuntime-level tests provide integration coverage.

---

### Summary

The consolidation commit (`443259c5`) cleanly achieved the stated goal: the two `ALTER TABLE` statements from the former `runMigration30` are now baked into the `CREATE TABLE` in `runMigration29`, the migration 30 function and its call site are gone, and all downstream consumers (repository, shared types, test helpers) are consistent. No correctness issues found.

PR #455, opened 2026-03-20
feat: Space creation UX and 3-column layout (Task 5.3)

lsm:
- SpaceCreateDialog: workspace path as hero field (required), name
  auto-suggests from dir basename, optional description, validates
  client-side + server-side via space.create RPC
- SpaceIsland: full 3-column layout (nav 240px / main flex-1 / task
  pane 320px on md+) replacing the placeholder; loads space via
  spaceStore.selectSpace(); responsive (right pane hidden on mobile)
- SpaceDashboard: default middle-column view — space name, workspace
  path, description, active-run/task status banner, quick-action cards,
  recent activity list
- SpaceNavPanel: left-column panel — workflow runs with status dots and
  task counts, standalone tasks, nav links (Agents, Workflows, Settings)
- SpaceTaskPane: right-column task detail — status badge, priority,
  workflow step indicator, description, progress, result, error, PR link,
  human-input area for needs_attention status
- SpaceStore: add spaces signal + initGlobalList() for the global space
  list (subscribes to space.created/updated/archived/deleted events)
- ContextPanel: wires SpaceCreateDialog + real space list from spaceStore;
  "Create Space" header button now functional in Spaces section
- 58 unit tests + e2e tests for space creation flow

### DIFF
--- a/packages/e2e/tests/features/space-creation.e2e.ts
+++ b/packages/e2e/tests/features/space-creation.e2e.ts
@@ -43,8 +43,10 @@ test.describe('Space Creation UX', () => {
 	test.beforeEach(async ({ page }) => {
 		await page.goto('/');
 		await waitForWebSocketConnected(page);
-		// Small settle time after connection
-		await page.waitForTimeout(300);
+		// Wait for the NavRail to be visible — deterministic signal that the app is ready
+		await expect(page.getByRole('button', { name: 'Spaces', exact: true })).toBeVisible({
+			timeout: 5000,
+		});
 	});
 
 	test.afterEach(async ({ page }) => {
--- a/packages/e2e/tests/features/space-creation.e2e.ts
+++ b/packages/e2e/tests/features/space-creation.e2e.ts
@@ -1,0 +1,171 @@
+/**
+ * Space Creation E2E Tests
+ *
+ * Verifies:
+ * - Navigating to the Spaces section
+ * - "Create Space" dialog opens
+ * - Workspace path field is required
+ * - Name auto-suggests from workspace path
+ * - Creating a space navigates to it
+ * - Space 3-column layout renders (nav panel, dashboard, no task pane initially)
+ * - Nav panel shows "No runs or tasks yet" for a fresh space
+ *
+ * Setup: creates a space via dialog (UI-only)
+ * Cleanup: deletes the space via RPC in afterEach (infrastructure)
+ */
+
+import { test, expect } from '../../fixtures';
+import { waitForWebSocketConnected, getWorkspaceRoot } from '../helpers/wait-helpers';
+
+const DESKTOP_VIEWPORT = { width: 1280, height: 720 };
+
+async function deleteSpaceByRpc(
+	page: Parameters<typeof waitForWebSocketConnected>[0],
+	spaceId: string
+): Promise<void> {
+	if (!spaceId) return;
+	try {
+		await page.evaluate(async (id) => {
+			const hub = window.__messageHub || window.appState?.messageHub;
+			if (!hub?.request) return;
+			await hub.request('space.delete', { id });
+		}, spaceId);
+	} catch {
+		// Best-effort cleanup
+	}
+}
+
+test.describe('Space Creation UX', () => {
+	test.use({ viewport: DESKTOP_VIEWPORT });
+
+	let createdSpaceId = '';
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/');
+		await waitForWebSocketConnected(page);
+		// Small settle time after connection
+		await page.waitForTimeout(300);
+	});
+
+	test.afterEach(async ({ page }) => {
+		if (createdSpaceId) {
+			await deleteSpaceByRpc(page, createdSpaceId);
+			createdSpaceId = '';
+		}
+	});
+
+	test('navigates to Spaces section via NavRail', async ({ page }) => {
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await expect(spacesButton).toBeVisible({ timeout: 5000 });
+		await spacesButton.click();
+
+		// ContextPanel should show "Spaces" header
+		await expect(page.locator('h2:has-text("Spaces")')).toBeVisible({ timeout: 5000 });
+	});
+
+	test('shows "Create Space" button in Spaces section', async ({ page }) => {
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await spacesButton.click();
+
+		// Header CTA button
+		await expect(
+			page.getByRole('button', { name: 'Create Space', exact: true }).first()
+		).toBeVisible({
+			timeout: 5000,
+		});
+	});
+
+	test('opens Create Space dialog when button clicked', async ({ page }) => {
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await spacesButton.click();
+
+		const createButton = page.getByRole('button', { name: 'Create Space', exact: true }).first();
+		await expect(createButton).toBeVisible({ timeout: 5000 });
+		await createButton.click();
+
+		// Dialog should appear
+		await expect(page.getByRole('dialog')).toBeVisible({ timeout: 5000 });
+		await expect(page.locator('text=Workspace Path')).toBeVisible({ timeout: 3000 });
+	});
+
+	test('workspace path is required — shows error on empty submit', async ({ page }) => {
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await spacesButton.click();
+		await page.getByRole('button', { name: 'Create Space', exact: true }).first().click();
+		await expect(page.getByRole('dialog')).toBeVisible({ timeout: 5000 });
+
+		// Submit without filling workspace path
+		const submitButton = page.getByRole('dialog').getByRole('button', { name: 'Create Space' });
+		await submitButton.click();
+
+		// Should show validation error
+		await expect(page.locator('text=Workspace path is required')).toBeVisible({ timeout: 3000 });
+	});
+
+	test('auto-suggests name from workspace path', async ({ page }) => {
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await spacesButton.click();
+		await page.getByRole('button', { name: 'Create Space', exact: true }).first().click();
+		await expect(page.getByRole('dialog')).toBeVisible({ timeout: 5000 });
+
+		// Type a workspace path
+		const pathInput = page.locator('input[placeholder*="/Users/you/projects"]');
+		await pathInput.fill('/projects/my-cool-project');
+
+		// Name should auto-suggest
+		const nameInput = page.locator('input[placeholder="e.g., My App"]');
+		await expect(nameInput).toHaveValue('my-cool-project', { timeout: 2000 });
+	});
+
+	test('creates space and shows 3-column layout', async ({ page }) => {
+		const workspaceRoot = await getWorkspaceRoot(page);
+
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await spacesButton.click();
+		await page.getByRole('button', { name: 'Create Space', exact: true }).first().click();
+		await expect(page.getByRole('dialog')).toBeVisible({ timeout: 5000 });
+
+		// Fill workspace path with the server's workspace root (guaranteed to exist)
+		const pathInput = page.locator('input[placeholder*="/Users/you/projects"]');
+		await pathInput.fill(workspaceRoot);
+
+		// Set a unique name to avoid conflicts
+		const nameInput = page.locator('input[placeholder="e.g., My App"]');
+		await nameInput.fill(`E2E Space ${Date.now()}`);
+
+		// Submit
+		const submitButton = page.getByRole('dialog').getByRole('button', { name: 'Create Space' });
+		await submitButton.click();
+
+		// Wait for navigation to the new space
+		await page.waitForURL(/\/space\/[a-f0-9-]+/, { timeout: 10000 });
+
+		// Extract the space ID from the URL for cleanup
+		const url = page.url();
+		const match = url.match(/\/space\/([a-f0-9-]+)/);
+		if (match) {
+			createdSpaceId = match[1];
+		}
+
+		// Space layout should be visible
+		// Left column: nav panel
+		await expect(page.locator('text=No runs or tasks yet')).toBeVisible({ timeout: 5000 });
+		// Middle column: dashboard (quick actions)
+		await expect(page.locator('text=Quick Actions')).toBeVisible({ timeout: 5000 });
+		await expect(page.locator('text=Start Workflow Run')).toBeVisible({ timeout: 3000 });
+		await expect(page.locator('text=Create Task')).toBeVisible({ timeout: 3000 });
+	});
+
+	test('dialog can be closed with Cancel button', async ({ page }) => {
+		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
+		await spacesButton.click();
+		await page.getByRole('button', { name: 'Create Space', exact: true }).first().click();
+		await expect(page.getByRole('dialog')).toBeVisible({ timeout: 5000 });
+
+		// Click Cancel
+		await page.getByRole('button', { name: 'Cancel', exact: true }).click();
+
+		// Dialog should close
+		await expect(page.getByRole('dialog')).not.toBeVisible({ timeout: 3000 });
+	});
+});
--- a/packages/web/src/components/space/SpaceCreateDialog.tsx
+++ b/packages/web/src/components/space/SpaceCreateDialog.tsx
@@ -1,0 +1,166 @@
+/**
+ * SpaceCreateDialog Component
+ *
+ * Modal form for creating a new Space with:
+ * - Workspace Path (required, hero field)
+ * - Name (auto-suggested from directory name)
+ * - Description (optional)
+ * - Form validation and error handling
+ */
+
+import { useState } from 'preact/hooks';
+import { Modal } from '../ui/Modal';
+import { Button } from '../ui/Button';
+import { connectionManager } from '../../lib/connection-manager';
+import { navigateToSpace } from '../../lib/router';
+import type { Space } from '@neokai/shared';
+
+interface SpaceCreateDialogProps {
+	isOpen: boolean;
+	onClose: () => void;
+}
+
+/**
+ * Extract directory basename from a path string
+ */
+function basenameFromPath(p: string): string {
+	const normalized = p.replace(/[/\\]+$/, '');
+	const parts = normalized.split(/[/\\]/);
+	return parts[parts.length - 1] ?? '';
+}
+
+export function SpaceCreateDialog({ isOpen, onClose }: SpaceCreateDialogProps) {
+	const [workspacePath, setWorkspacePath] = useState('');
+	const [name, setName] = useState('');
+	const [description, setDescription] = useState('');
+	const [nameTouched, setNameTouched] = useState(false);
+	const [submitting, setSubmitting] = useState(false);
+	const [error, setError] = useState<string | null>(null);
+
+	const handlePathInput = (value: string) => {
+		setWorkspacePath(value);
+		if (!nameTouched) {
+			const suggested = basenameFromPath(value);
+			setName(suggested);
+		}
+	};
+
+	const handleSubmit = async (e: Event) => {
+		e.preventDefault();
+
+		if (!workspacePath.trim()) {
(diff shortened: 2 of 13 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

Clean, well-structured implementation of the Space creation dialog and 3-column layout shell. The code follows existing project patterns, has good test coverage (58 unit tests + E2E), and the components are appropriately decomposed. No blocking issues found.

### Detailed Review

**Correctness**
- SpaceCreateDialog correctly validates workspace path (required), auto-suggests name from basename, sends `space.create` RPC, navigates on success, and resets form on close. The `basenameFromPath` helper correctly handles both `/` and `\` separators and trailing slashes.
- SpaceIsland properly implements the 3-column layout with promise-chain-locked space loading via `spaceStore.selectSpace()`.
- SpaceNavPanel correctly groups tasks by workflow run and shows standalone tasks separately.
- SpaceTaskPane renders all expected task fields and conditionally shows the human input area only for `needs_attention` status.
- The `initGlobalList()` method in SpaceStore is idempotent and resets the flag on error so retries work on reconnect.

**Completeness**
- All 5 deliverables from the task description are implemented.
- ContextPanel wiring is complete: "Create Space" button, dialog lifecycle, global space list fetch on section entry, and `onCreateSpace` callback passed to SpaceContextPanel.
- Status types (`SpaceTaskStatus`, `SpaceTaskPriority`, `WorkflowRunStatus`) are all properly covered in the Record maps and StatusDot component.

**Architecture**
- Components correctly read from the singleton `spaceStore` signals, maintaining the existing signal-based reactive pattern.
- The right pane is conditionally rendered and hidden on mobile (`hidden md:flex`), matching the responsive design spec.
- The `spaceId` prop is accepted but aliased as `_spaceId` in SpaceDashboard and SpaceNavPanel since data comes from the global store -- this is a conscious pattern choice, keeping the prop for future flexibility.

**Tests**
- Unit tests cover: rendering states (loading, empty, populated), user interactions (click handlers, form submission), validation (empty path, no connection), auto-suggestion logic, and edge cases (truncated paths, singular/plural task counts).
- E2E tests verify the full flow through the UI: navigation to Spaces section, dialog open/close, required field validation, name auto-suggest, and space creation with 3-column layout verification. Cleanup uses RPC (acceptable per E2E guidelines).

### Minor Observations (P3)

1. **`SpaceTaskPane.tsx` line 70**: The `HumanInputArea` component initializes `inputText` from `task.inputDraft`, but if `task` object reference changes (e.g., due to a real-time update) while the user is typing, the local state won't re-sync. This is a common pattern and acceptable for now since useState initialization only runs once.

2. **`SpaceNavPanel.tsx` lines 178-184**: The NavLink items (Agents, Workflows, Settings) are static placeholders with no onClick handlers -- they render as `div` with `cursor-pointer` but do nothing when clicked. This is fine for a shell/scaffold, just noting for future tasks.

3. **`initGlobalList` event subscriptions**: The `hub.onEvent()` subscriptions created in `initGlobalList()` are never cleaned up (unlike the space-specific subscriptions). Since this is a global singleton that lives for the app lifetime, this is acceptable -- but worth noting if reconnect handling needs to re-subscribe in the future.

4. **`SpaceDashboard.tsx` line 183**: `run.status.replace('_', ' ')` only replaces the first underscore. If future statuses have multiple underscores, use `replaceAll('_', ' ')`. Current statuses (`in_progress`, `needs_attention`) only have one underscore each, so this works correctly today.

lsm (commented):

> **Recommendation:** REQUEST_CHANGES

---

## Summary

The implementation delivers all requested features: SpaceCreateDialog with required workspace-path validation and name auto-suggestion, a fully wired 3-column SpaceIsland layout, SpaceDashboard, SpaceNavPanel, SpaceTaskPane, SpaceStore global list, and ContextPanel integration. The overall architecture is sound and the component design is clean. There are a few meaningful defects that should be addressed before merge.

---

## Issues

### P1 — Global space list is not refreshed on WebSocket reconnect

**File:** `packages/web/src/lib/space-store.ts` (lines 138–177), `packages/web/src/lib/connection-manager.ts` (line 513)

`spaceStore.refresh()` is registered in the connection-manager's reconnect handler, but that method only refreshes the **currently-selected space** (calls `fetchInitialState`). `initGlobalList()` — the only code path that re-subscribes to global `space.created/updated/archived/deleted` events and re-fetches the full list — is never called on reconnect.

As a result, if the user experiences a WebSocket drop-and-reconnect while viewing a non-spaces section, the sidebar list will be stale and global real-time updates will be silently missed for the rest of the session (the event subscriptions established in `initGlobalList` are torn down when the old hub connection closes, and are never re-established).

The `refresh()` method should also call `initGlobalList()` when the global list is already initialized, or `initGlobalList` should reset `globalListInitialized` on disconnect so it re-runs on the next navigation.

---

### P1 — CSS class conflict on the right-pane element makes it effectively invisible

**File:** `packages/web/src/islands/SpaceIsland.tsx` line 100

```tsx
<div class="hidden md:flex w-80 flex-shrink-0 border-l border-dark-700 overflow-hidden flex-col">
```

`hidden` sets `display: none`. `md:flex` then sets `display: flex` at the `md` breakpoint, but `flex-col` adds `flex-direction: column` — without `md:flex-col`, on screens smaller than `md` the pane is hidden (intended), but the class list includes both `flex-col` and the responsive `md:flex` which only overrides the display property. The standard Tailwind pattern for this pattern is `hidden md:flex md:flex-col` or simply `hidden md:block` with column flex set separately. In practice, `flex-col` still applies its `flex-direction` even when `display: none` is in effect, but more importantly the author's intent is clear — the responsive breakpoint handling is correct **because** `md:flex` beats the static `hidden` at the `md` breakpoint. This is a style/legibility P2, not a functional bug. _(Revised: on closer inspection this pattern is consistent with other uses in the codebase, so demoting to P3.)_

---

### P1 — `HumanInputArea` updates status to `in_progress` unconditionally

**File:** `packages/web/src/components/space/SpaceTaskPane.tsx` lines 81–84

```tsx
await spaceStore.updateTask(task.id, {
  inputDraft: inputText.trim(),
  status: 'in_progress',
});
```

When the user submits a human response in the `needs_attention` panel, the code blindly sets status to `'in_progress'`. Whether this transition is semantically correct is a product question, but the server-side `spaceTask.update` handler performs status transition validation. If the daemon rejects the `needs_attention → in_progress` transition, the call throws an error which is caught and shown to the user, while the `inputDraft` is discarded. The intent appears to be: "resume the task", but there is no way to signal to the daemon what the human response *means*. This is either incomplete (there should be a dedicated RPC for submitting human input) or the transition logic needs to account for validation failures separately. At minimum, the `inputDraft` should be persisted regardless of whether the status change succeeds.

---

### P2 — `initGlobalList` race: flag set before async work completes

**File:** `packages/web/src/lib/space-store.ts` lines 139–140

```ts
if (this.globalListInitialized) return;
this.globalListInitialized = true;  // set before awaiting hub or list fetch
```

If two concurrent calls to `initGlobalList()` arrive before the first completes (e.g., user navigates quickly), the flag prevents the second call from running — which is the desired deduplication behavior. However, if the first call **fails** partway through (e.g., the hub call fails after the flag is set but before `globalListInitialized = false` is reset in the catch block), the in-flight subscriptions registered before the failure will be dangling — they are never added to `cleanupFunctions` since that array belongs to per-space subscriptions, not the global list. The event listener leak is a minor concern, but the flag reset on failure (line 175) is correct and sufficient. No change needed beyond awareness.

---

### P2 — `spaceId` prop accepted but silently ignored by `SpaceNavPanel` and `SpaceDashboard`

**Files:** `packages/web/src/components/space/SpaceNavPanel.tsx` line 98, `packages/web/src/components/space/SpaceDashboard.tsx` line 80

Both components destructure `spaceId` as `_spaceId` (intentionally unused) and instead read directly from the singleton `spaceStore`. This means:
1. The components implicitly rely on `SpaceIsland` having already called `spaceStore.selectSpace()` before rendering — there is no guard if they are ever mounted independently.
2. The `spaceId` prop in the component interface is misleading; callers would assume the prop drives which space is displayed, but it does not.

This is acceptable as a deliberate architectural choice (singleton store pattern) but should be documented. If these components are ever reused in a context where `selectSpace` has not been called, they will silently show data for whatever space was last selected.

---

### P2 — No test for reconnect-then-global-list-refresh behavior in SpaceStore

There are no unit tests for `SpaceStore.initGlobalList()` — specifically the idempotency guard, the event subscription wiring, and the reconnect gap identified in the P1 above. The existing 58 tests cover the component layer well, but there are zero tests for the store's `initGlobalList` or `refresh` methods.

---

### P3 — `page.waitForTimeout(300)` in E2E test beforeEach

**File:** `packages/e2e/tests/features/space-creation.e2e.ts` line 47

Fixed-duration sleeps are fragile on slow CI. The `waitForWebSocketConnected` helper already establishes connection — the additional 300 ms sleep is unnecessary if `waitForWebSocketConnected` reliably indicates readiness. Consider replacing with a deterministic wait (e.g., wait for the NavRail to be visible).

---

### P3 — Nav links in `SpaceNavPanel` are non-functional placeholders

**File:** `packages/web/src/components/space/SpaceNavPanel.tsx` lines 169–173

The Agents, Workflows, and Settings nav links are rendered as non-interactive `div` elements with a `cursor-pointer` style but no `onClick` handlers or `href`. Clicking them does nothing. This is presumably scaffolding for future milestone work — it should be documented with a TODO comment so it is not mistaken for a regression.

---

### P3 — Quick action callbacks are optional but left unwired

**File:** `packages/web/src/islands/SpaceIsland.tsx` line 95, `packages/web/src/components/space/SpaceDashboard.tsx` lines 13–14

`onStartWorkflow` and `onCreateTask` are defined as optional props but `SpaceIsland` does not pass them to `SpaceDashboard`. Clicking either card is a no-op. This is expected scaffolding for future milestones but should have a TODO comment to signal intent.

---

## What works well

- The `SpaceCreateDialog` client-side validation (required path, name auto-suggestion, name-touched guard) is correctly implemented and well-tested.
- The promise-chain lock in `selectSpace` correctly serializes space switches and prevents race conditions on rapid navigation.
- The subscription cleanup pattern (storing unsubscribe callbacks, calling them on space switch) is solid.
- Test coverage for all four new components is thorough — the 58 unit tests cover error states, empty states, interactions, and computed signal behavior.
- E2E test follows all project rules: creates via UI, cleans up via RPC in afterEach, uses DOM assertions only.

lsm (commented):

**Recommendation: APPROVE**

### Summary

All issues from the previous review have been addressed. The implementation is clean, well-tested, and ready to merge.

**P1 fix — `refresh()` now re-initializes global space list on reconnect:** Confirmed. `refresh()` at line 578 checks `globalListInitialized`, resets the flag, and fire-and-forgets `initGlobalList()`. The fire-and-forget pattern is correct here since global list refresh and per-space data refresh are independent operations. The comment block (lines 567–576) clearly documents why the flag reset is needed.

**P1 fix — `HumanInputArea` persists inputDraft before status transition:** Confirmed. `SpaceTaskPane.tsx` lines 82–84 now make two sequential `updateTask` calls — first to persist `inputDraft`, then to transition status. If the status transition fails, the draft is already saved server-side. This is the correct approach.

**P2 fix — 13 new unit tests for `initGlobalList` and `refresh`:** Confirmed. Tests at lines 880–1007 of `space-store.test.ts` cover:
- Space list fetch and signal population
- Idempotency guard (second call is a no-op)
- Event handler wiring for `space.created`, `space.updated`, `space.archived`, `space.deleted`
- Duplicate prevention on `space.created`
- Flag reset on failure enabling retry
- Reconnect re-initialization via `refresh()`
- No-op behavior when global list was never initialized

Total test count is now 71 in the store test file alone, plus 58 component tests. Coverage is thorough.

**P3 fix — `waitForTimeout(300)` replaced:** Confirmed. E2E `beforeEach` now uses a deterministic NavRail visibility check (`getByRole('button', { name: 'Spaces', exact: true })`) instead of a fixed sleep.

**P3 fix — TODO comments added:** Confirmed. `SpaceDashboard.tsx` line 149 and `SpaceNavPanel.tsx` line 166 both have clear TODO comments documenting that callbacks/onClick handlers are scaffolding for future milestones.

1. **`SpaceDashboard.tsx` line 184**: `run.status.replace('_', ' ')` still only replaces the first underscore. Current statuses are fine, but `replaceAll` would be more future-proof. Not blocking.

2. **`initGlobalList` event subscriptions are not stored for cleanup**: The `hub.onEvent()` return values (unsubscribe functions) are discarded. Since the old hub instance is replaced entirely on reconnect and the new `initGlobalList` call registers fresh handlers on the new hub, this is correct behavior — old handlers are GC'd with the old hub instance. No action needed.

lsm (commented):

---

### Summary of findings

All four previously-flagged issues were addressed — the reconnect refresh, the two-step HumanInputArea submit, the silently-ignored spaceId prop TODO comments, and the 13 new store-level unit tests. However, the reconnect fix introduced a new P0 regression (duplicate event subscriptions), and the HumanInputArea fix has no behavioral test coverage.

---

### P0 — Duplicate event subscriptions on page-resume reconnect

**File:** `packages/web/src/lib/space-store.ts` lines 138–177, 578–585

The fix comment in `refresh()` states "The old hub connection is closed on disconnect, tearing down any event subscriptions registered in initGlobalList()." This assumption is incorrect for the most common reconnect path.

In the auto-reconnect and page-visibility-resume flows, `validateConnectionOnResume` (connection-manager.ts:479) calls `spaceStore.refresh()` while the same `MessageHub` instance is reused (no new hub is created). The handlers registered by `initGlobalList()` live in `MessageHub.channelEventHandlers` (message-hub.ts:333-336) and are NOT cleared when the WebSocket transport reconnects — only a full `disconnect()`+`connect()` sequence creates a new hub.

As a result, each call to `refresh()` that resets `globalListInitialized` and re-calls `initGlobalList()` registers a second (third, Nth) set of handlers for `space.created`, `space.updated`, `space.archived`, and `space.deleted` on the same hub. After N page-resume reconnects, every space event fires N times, corrupting the `spaces` signal (e.g., duplicate entries on `space.created`, N-fold partial updates on `space.updated`).

**Fix needed:** Store the cleanup functions returned by `hub.onEvent()` in a dedicated array (e.g., `globalListCleanupFns`), and call them before re-registering in `initGlobalList()`, or before calling `initGlobalList()` again from `refresh()`.

---

### P1 — HumanInputArea two-step submit has no behavioral test

**File:** `packages/web/src/components/space/__tests__/SpaceTaskPane.test.tsx`

The previous review flagged that `HumanInputArea` unconditionally transitioned status. The fix separates the two calls: draft first, then status. But the test file only checks that the input area *renders* for `needs_attention` status — there is no test that:

- Verifies `updateTask` is called with `{ inputDraft: ... }` first
- Verifies `updateTask` is called with `{ status: 'in_progress' }` second
- Verifies that if the first `updateTask` throws, the status transition is NOT attempted
- Verifies that if the second `updateTask` throws, an error is shown

The specific scenario "draft is persisted even if status transition fails" is the business-critical invariant of this fix — it needs a test.

---

### P2 — `refresh()` test relies on microtask timing coincidence

**File:** `packages/web/src/lib/__tests__/space-store.test.ts` lines 976–984

The test asserts that after `await spaceStore.refresh()`, `space.list` was called. However `refresh()` calls `initGlobalList()` with `.catch()` (fire-and-forget, not awaited). The assertion passes today because the mock `request` resolves synchronously in the same microtask tick. This makes the test implicitly coupled to mock scheduling behavior and could start flaking if the mock or test runner changes. A comment or explicit `await vi.runAllMicrotasks()` / flushPromises step would make the intent explicit and prevent silent drift.

---

### P3 — Inconsistency between `spaceId` param and `_spaceId` in SpaceDashboard / SpaceNavPanel

**Files:** `packages/web/src/components/space/SpaceDashboard.tsx` line 108, `packages/web/src/components/space/SpaceNavPanel.tsx` line 136

Both components accept `spaceId` but rename it to `_spaceId` to suppress the unused-variable lint rule. The TODO comments explain this is scaffolding. No action required — the fix is intentional — but worth noting for the next task owner who wires these props.

lsm (commented):

---

**Recommendation: APPROVE** — All three previously reported issues are correctly resolved. No new concerns.

---

### P0 Fix — Duplicate hub.onEvent subscriptions: VERIFIED

`initGlobalList()` in `/packages/web/src/lib/space-store.ts` now:
1. Runs cleanup before re-registering (lines 155–162): iterates `globalListCleanupFns`, calls each, then resets the array.
2. Stores each `hub.onEvent()` return value in `globalListCleanupFns` (lines 170, 181, 189, 197).

The P0 regression test at line 975 in `/packages/web/src/lib/__tests__/space-store.test.ts` directly validates the fix: it manually resets `globalListInitialized` to simulate a reconnect while leaving stale handlers on the hub, then re-initializes and asserts `fireMockEvent('space.created', ...)` produces exactly one new entry. The `mockEventHandlerSets` multi-handler tracking in `makeMockHub()` makes this assertion meaningful.

One minor note: the test manually resets `priv.globalListInitialized = false` rather than calling `refresh()`, which is fine for the targeted regression test. The complementary `refresh` test (line 1013) validates the end-to-end reconnect path.

---

### P1 Fix — HumanInputArea behavioral tests: VERIFIED

Four new tests in `/packages/web/src/components/space/__tests__/SpaceTaskPane.test.tsx` (lines 207–263):

1. **Call order** (line 207): Asserts `mockUpdateTask` is called exactly twice, with `{ inputDraft: 'Looks good to me' }` first and `{ status: 'in_progress' }` second — directly validates the two-step ordering in `handleSubmit`.

2. **No status transition on draft failure** (line 223): `mockRejectedValueOnce` on the first call, then asserts only 1 total call. This correctly tests the observable behavior: since both calls are in the same `try` block, the thrown exception from the first `await` prevents the second from being reached.

3. **Error display on status transition failure** (line 238): Two mock calls — first resolves, second rejects — then asserts `getByText('Invalid transition')` appears. Confirms the catch block sets the error state.

4. **Empty input is no-op** (line 255): Relies on the `disabled` attribute on the submit button. This is a correct UI-level test: the button is disabled when `inputText.trim()` is falsy, so the click is a no-op in the browser. In jsdom, clicking a disabled submit button does not trigger `onSubmit`, so `mockUpdateTask` is never called.

---

### P2 Fix — Timing comment: VERIFIED

The reconnect re-init test at line 1019–1021 now includes an explanatory comment:

> refresh() fire-and-forgets initGlobalList() via .catch(). The assertion passes because all mocks resolve synchronously (vi.fn async), so the microtask queue drains within the same await tick as refresh() itself.

This accurately explains why the test is reliable despite `initGlobalList()` being called with `.catch()` (fire-and-forget pattern). The explanation is correct: `vi.fn` async mocks return already-resolved promises, so awaiting `refresh()` gives all microtasks enough time to drain.

---

No new issues found. The implementation is clean and all three fixes match the stated intent.

lsm (commented):

**Recommendation: APPROVE**

### Summary

All three fixes from the previous review iteration are correctly implemented and well-tested. No new issues found.

### Fix verification

**P0 (duplicate subscriptions) -- Fixed correctly.**
`initGlobalList()` now stores cleanup functions returned by `hub.onEvent()` in `globalListCleanupFns` (line 134). Before re-registering, the method iterates and calls all stale cleanup functions (lines 155-162), then clears the array. The `refresh()` method resets `globalListInitialized` to false (line 611), and the subsequent `initGlobalList()` call properly cleans up before re-subscribing. The test at line 975 (`removes stale handlers before re-registering on reconnect re-init`) validates this by using `fireMockEvent` which fires on a multi-handler Set -- if cleanup failed, the assertion at line 989 would see 2 entries instead of 1.

**P1 (HumanInputArea behavioral tests) -- Fixed correctly.**
Four new tests in `SpaceTaskPane.test.tsx` under the `HumanInputArea submit behavior` describe block:
- Call order: verifies `inputDraft` is persisted first, `status` transition second (line 2017)
- Draft failure skips status: `mockRejectedValueOnce` on first call, asserts only 1 call made (line 2033)
- Status failure shows error: first call succeeds, second rejects, error message displayed (line 2048)
- Empty input no-op: verifies `updateTask` is never called (line 2065)

These tests cover the business-critical invariant (draft is never lost even if status transition fails).

**P2 (reconnect test comment) -- Fixed correctly.**
The test at line 1013 now has an explanatory comment (lines 1019-1021) documenting why the fire-and-forget `initGlobalList()` assertion works: mock `request` resolves synchronously, so the microtask queue drains within the same await tick.

### Other observations

- The `initGlobalList()` flag is set eagerly at line 151 (before the await), which correctly prevents concurrent double-registration if two callers invoke it simultaneously.
- Error recovery is sound: if `initGlobalList()` fails, the flag is reset (line 205) and `globalListCleanupFns` was already cleared (line 162), so a retry starts clean.
- The `fireMockEvent` helper and `mockEventHandlerSets` multi-handler map are well-designed test infrastructure additions that accurately model real hub behavior where multiple subscribers can exist per event.

PR #456, opened 2026-03-20
task/task 83 frontend exportimport ui

lsm:
- **feat: Space export/import UI with agent/workflow list views and preview dialog (Task 8.3)**


lsm (commented):

**Recommendation: APPROVE**

Overall this is a clean, well-structured implementation of the Space export/import UI. The code follows existing project patterns, uses Preact correctly (not React), avoids console calls, and has meaningful test coverage. Below are my findings organized by severity.

---

### No P0 or P1 Issues Found

The implementation correctly handles:
- File naming convention `{space-name}-{type}-{date}.neokai.json` with sanitization
- Per-item conflict resolution (skip/rename/replace) with proper state management
- Import button correctly disabled when nothing will be created (all conflicts skipped)
- Proper error handling at all RPC call boundaries
- Cross-reference warnings displayed when validation errors exist
- List auto-refresh via existing SpaceStore event subscriptions (`spaceAgent.created`, `spaceWorkflow.created`)
- Clean use of the existing `Modal` component with proper `size="lg"` prop

---

### P2 Issues (2)

**1. `pickImportFile` can resolve the Promise multiple times (harmless but sloppy)**
`packages/web/src/components/space/export-import-utils.ts`, lines 43-76

The `onchange` handler, `oncancel` handler, and the `handleFocus` fallback timeout can all independently call `resolve()`. Since a Promise ignores subsequent resolve calls, this is not a bug, but it makes the control flow harder to reason about. A simple `let resolved = false` guard would clarify intent. Low priority.

**2. No list refresh after successful import -- relies entirely on backend events**
`packages/web/src/components/space/SpaceAgentList.tsx`, lines 93-111 and `WorkflowList.tsx`, lines 93-111

After `executeImport` succeeds, the code only shows a toast and clears the dialog state. It does not explicitly refresh the agent/workflow list. This works correctly because the SpaceStore subscribes to `spaceAgent.created` / `spaceWorkflow.created` events (verified in `space-store.ts`). However, if the backend import handler does not emit individual creation events for each imported item (e.g., it only emits a bulk event or no events at all), the UI list will be stale until the user navigates away and back. Worth verifying that the backend `spaceImport.execute` handler emits per-item creation events, or adding an explicit list refresh as a fallback.

---

### P3 Issues (3)

**1. Duplicated import/export logic across SpaceAgentList and WorkflowList**
Both components have nearly identical `startImport`, `executeImport`, `exportAll`, and `export{Item}` functions with the same error handling and hub connection patterns. This is acceptable for two components but could be extracted into a shared hook (e.g., `useSpaceImportExport`) if a third consumer appears.

**2. Cross-reference warning always shown when there are validation errors**
`packages/web/src/components/space/ImportPreviewDialog.tsx`, lines 239-246

The yellow cross-reference warning box is shown whenever `hasValidationErrors` is true, but not all validation errors are cross-reference issues. The warning text specifically says "Workflows with unresolved agent references cannot be imported" which may be misleading for non-reference validation errors.

**3. Dev-proxy change is unrelated to the task**
The `--as-system-proxy false` addition to `packages/daemon/tests/helpers/dev-proxy.ts` is a separate fix (matches the other commit "fix: start devproxy without changing system proxy"). It is fine to include but would be cleaner as a separate PR to keep the commit history focused.

---

### What Works Well

- **Type safety**: All RPC calls use typed generics for request/response shapes
- **UX polish**: Disabled states during execution, spinner on Import button, hover-reveal per-item Export, empty states with helpful copy
- **Accessibility**: `aria-label` on conflict dropdowns, `role="alert"` on validation errors, `type="button"` on all buttons
- **Test quality**: ImportPreviewDialog tests cover open/close, new items, conflicts, strategy changes, summary counts, and execution state. The `downloadBundle` tests mock DOM APIs correctly.
- **E2E tests**: Cover single export, export all, import with no conflicts, import with conflicts, and mixed bundle import. The file injection via `createElement` monkey-patching is creative and avoids needing a test server to serve files.
- **No console calls** in any application code
- **Correct Preact usage**: `class` not `className`, `preact/hooks` not `react`

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P0 (list does not refresh after import), two P1s (incorrect import count toast, stale conflict resolutions), and a P2 JSX attribute naming bug.

---

### P0 — List does not refresh after a successful import

The import execution handler writes directly to the repository inside a SQLite transaction but never emits the real-time events that the frontend SpaceStore listens to. Compare the regular agent-create path in `packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts`:

```ts
daemonHub.emit('spaceAgent.created', { ... })
```

The import handler (`packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 442–482) calls `agentRepo.create()` and `workflowManager.createWorkflow()` directly inside the transaction callback, without emitting any events.

Because `spaceStore` subscribes to `spaceAgent.created` / `spaceWorkflow.created` for live updates (`packages/web/src/lib/space-store.ts`, lines 334 and 382), imported items are invisible in the list until the user navigates away and back. The feature is visually broken even though the data is correctly persisted.

**Fix**: after the transaction completes, emit `spaceAgent.created` / `spaceWorkflow.created` / `spaceAgent.updated` events (via StateManager or daemonHub) for each imported/replaced item.

---

### P1 — Success toast counts only agents even when a bundle with both agents and workflows is imported from the Agents tab

`packages/web/src/components/space/SpaceAgentList.tsx`, line 100:

```ts
const created = result.agents.filter((a) => a.action !== 'skipped').length;
toast.success(`Imported ${created} agent${created === 1 ? '' : 's'}`);
```

The ImportPreviewDialog shows and processes both agents and workflows for any bundle. If the user imports a bundle containing 3 workflows and 1 agent from the Agents tab, the toast says "Imported 1 agent" with no mention of the 3 workflows. The toast should reflect the full import result.

---

### P1 — Stale conflict-resolution state risk across sequential imports in the same component mount

`packages/web/src/components/space/ImportPreviewDialog.tsx`, lines 130–135:

```ts
const [agentResolutions, setAgentResolutions] = useState<Record<string, ConflictResolutionStrategy>>({});
```

Today the dialog component is conditionally mounted (parent sets `importPreview` to null on close, which unmounts the dialog), so each open starts clean. However this invariant is not documented and is one refactor away from silently applying stale resolutions from a previous import to a new one if two bundles share the same agent name. Add `key={bundle.exportedAt}` (or similar) to the dialog invocation in both `SpaceAgentList.tsx` and `WorkflowList.tsx`, or add an explicit `useEffect` reset.

---

### P2 — JSX SVG attributes use kebab-case instead of camelCase

`packages/web/src/components/space/ImportPreviewDialog.tsx`, lines 94–96 and 104–106:

```tsx
<path fill-rule="evenodd" clip-rule="evenodd" ... />
```

In JSX, SVG presentation attributes must be camelCased: `fillRule` and `clipRule`. The kebab-case forms are silently dropped by the VDOM; the warning icon and the "new item" icon will not render their fill paths correctly. Note this pattern also exists in `ErrorBanner.tsx` and `Toast.tsx`, so it predates this PR, but this PR adds four more occurrences and should correct them here.

---

### P2 — E2E: `document.createElement` monkeypatch is never restored

`packages/e2e/tests/features/space-export-import.e2e.ts`, lines 165–191, 236–261, 326–350:

Each of the three import tests replaces `document.createElement` inside `page.evaluate()` with a custom function but never restores the original. Within the same test, any subsequent call to `document.createElement('input')` will use the patched version. Add a restore step after the import button is clicked, or scope the patch more narrowly (e.g., patch only `HTMLInputElement.prototype.click` temporarily).

---

### P2 — Import summary wording is inaccurate for the «replace» strategy

`packages/web/src/components/space/ImportPreviewDialog.tsx`, line 251:

```tsx
Will create <span>{willCreateAgents}</span> agents
```

When the user picks "Replace existing" for all conflicts, `willCreateAgents` counts them but the label says "Will create". Replace overwrites an existing record; it does not create a new one. Consider "Will import" or a breakdown like "1 new, 2 replaced".

---

### P3 — The «import bundle with agents and workflows» E2E test only asserts the agents toast

`packages/e2e/tests/features/space-export-import.e2e.ts`, line 370:

```ts
await expect(page.locator('text=/Imported.*agent/')).toBeVisible({ timeout: 8000 });
```

The test sends a bundle with 1 agent + 1 workflow but only asserts the agents toast. The workflow import path has no E2E assertion in this test.

---

### P3 — `pickImportFile` leaves a window focus listener registered if the component unmounts before the user picks

`packages/web/src/components/space/export-import-utils.ts`, lines 67–74:

The `{ once: true }` flag ensures the listener fires only once, but if the component unmounts before the 300ms timer fires (e.g., the user navigates away mid-dialog), the Promise resolves after navigation with `null` and the captured `resolve` function runs on a detached component. The practical impact is benign today (Preact ignores setState on unmounted components), but the pattern is worth documenting.

lsm (commented):

All six issues from the previous review round have been addressed correctly. Here is the verification:

**P0 fix — Event emission after import:** Verified at `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts:565-605`. Events fire after the SQLite transaction commits, correctly distinguishing `created` vs `updated` vs `skipped`. The event shapes (`{ sessionId, spaceId, agent/workflow }`) match the existing patterns in `space-agent-handlers.ts` and `space-workflow-handlers.ts`. The `SpaceStore` subscriptions at `packages/web/src/lib/space-store.ts:329-411` handle these events and update signals accordingly. Five new unit tests cover the happy paths (created, replaced, skipped, workflow created, mixed bundle).

**P1 fix — Toast counts both agents and workflows:** Both `SpaceAgentList.tsx:100-106` and `WorkflowList.tsx:100-106` now build a `parts[]` array counting non-skipped agents and workflows, joined with "and". This correctly handles bundles containing both entity types.

**P1 fix — `key={importBundle.exportedAt}` for state reset:** Applied in both `SpaceAgentList.tsx:210` and `WorkflowList.tsx:221`. This ensures the `ImportPreviewDialog`'s `useState` hooks reset when a new bundle is loaded, preventing stale conflict resolution state.

**P2 fix — `fillRule`/`clipRule` SVG attributes:** Corrected at `ImportPreviewDialog.tsx:94,96,104,106` — now using camelCase JSX syntax.

**P2 fix — Summary wording "Will import":** Changed at `ImportPreviewDialog.tsx:251`. Test updated at `ImportPreviewDialog.test.tsx:216,218`.

**P2 fix — `resolved` guard in `pickImportFile`:** Applied at `export-import-utils.ts:39-44` — a `done()` wrapper function with a boolean guard prevents multiple resolutions from `onchange`, `oncancel`, and the focus-fallback timer.

**P2 fix — E2E `document.createElement` self-cleaning:** The `injectImportFile` helper at `space-export-import.e2e.ts:79-105` now restores `document.createElement` immediately on first input click (line 91), scoping the patch to exactly one input creation. This eliminates the need for `afterEach` cleanup.

### Remaining minor observation (P3)

1. **`stroke-width="4"` still kebab-case** — `ImportPreviewDialog.tsx:279` uses `stroke-width="4"` while `fillRule`/`clipRule` were converted to camelCase. Preact supports both forms for SVG attributes, so this is purely a consistency nit, not a bug.

**Recommendation: APPROVE** — All previously identified issues are resolved. The implementation is correct, well-tested, and the event emission pattern is consistent with existing handlers.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one confirmed P1 real-time state bug remains.

---

### Previous fixes verified

All five claimed fixes from the first round were applied correctly:

- **P0 event emission**: `spaceImport.execute` now emits `spaceAgent.created/updated` and `spaceWorkflow.created/updated` after the transaction commits. Five new unit tests cover each permutation. Verified.
- **P1 toast counts both agents and workflows**: Both `SpaceAgentList` and `WorkflowList` count non-skipped agents and workflows separately and build a joined string. Verified.
- **P1 `key={importBundle.exportedAt}`**: Present in both `SpaceAgentList.tsx:213` and `WorkflowList.tsx:224`. Dialog state resets on each new bundle. Verified.
- **P2 SVG `fillRule`/`clipRule`**: Now camelCase in `ImportPreviewDialog.tsx:94–96`. Verified.
- **P2 "Will import" summary wording**: Correct in `ImportPreviewDialog.tsx:251`. Verified.
- **P2 `resolved` guard in `pickImportFile`**: The `done()` wrapper prevents double-resolution. Verified.
- **P2 E2E `document.createElement` restore**: The patch now restores `document.createElement` inside the patched `input.click` handler (`space-export-import.e2e.ts:91`), before the fake `change` event fires. Verified.

---

### Remaining Issues

#### P1 — Workflow `replace` leaves stale entry in SpaceStore's reactive signal

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 516–522 and 594–604

**File:** `packages/web/src/lib/space-store.ts`, lines 392–411

**Root cause:** When a workflow is imported with `strategy === 'replace'`, the handler:
1. Deletes the existing workflow (`workflowRepo.deleteWorkflow(existing.id)`) inside the transaction — a new UUID is assigned to the replacement.
2. Creates a brand-new workflow with a **different UUID** via `workflowManager.createWorkflow(...)`.
3. Emits `spaceWorkflow.updated` carrying the **new** UUID.

The `SpaceStore.updated` handler resolves the event by looking for a workflow whose `id` matches the event's `workflow.id`. Since the new UUID has never been in the store, `findIndex` returns -1, and the fallback path **appends** the new workflow. The **old** workflow (old UUID) is never removed, so the user sees two workflows with the same name — the stale old one and the fresh replacement — until the page is refreshed.

The correct fix is to emit both a `spaceWorkflow.deleted` event (with the old `existing.id`) and a `spaceWorkflow.created` event (with the new workflow), matching what the `spaceWorkflow.updated` handler in `SpaceStore` actually expects for in-place identity-preserving updates.

---

#### P2 — Workflow event emission uses `listWorkflows` scan instead of `getWorkflow`

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, lines 590–593

```ts
const workflow: SpaceWorkflow | undefined = workflowRepo
    .listWorkflows(spaceId)
    .find((w) => w.id === item.id);
```

`SpaceWorkflowRepository` exposes `getWorkflow(id): SpaceWorkflow | null` which is a direct indexed lookup. The current code loads all workflows for the space and scans them in JS. For spaces with many workflows this is unnecessary; a single call to `workflowRepo.getWorkflow(item.id)` is the right approach and matches how the agent case does `agentRepo.getById(item.id)` on line 574.

---

#### P3 — `pickImportFile` focus-listener not cleaned up on file-selected path

**File:** `packages/web/src/components/space/export-import-utils.ts`, lines 74–80

The `window.addEventListener('focus', handleFocus, { once: true })` listener is not explicitly removed when the user selects a file (resolve via `onchange`). In practice `{ once: true }` means it auto-removes on the next focus event, so no memory leak, but it means the 300ms timer will still fire after the file is read. The `done()` guard prevents double-resolution, so there is no correctness bug. Still, an explicit `window.removeEventListener('focus', handleFocus)` inside `input.onchange` would be cleaner.

---

### Items reviewed and confirmed correct

- **Import preview dialog**: State management (`agentResolutions`, `workflowResolutions`), `key` prop reset, validation error display, and `Import` button disabled state are all correct.
- **Export handlers**: Agent-only, workflow-only, and bundle export all respect optional ID filters; workflow export correctly trims unreferenced agents from the bundle.
- **Atomicity**: Transaction wraps all mutations; both rollback tests (unresolved agent ref during workflow import, workflow deletion rollback on replace failure) pass.
- **`spaceAgent.updated` for agent replace**: Correct — agent UUID is preserved on replace, so the `SpaceStore.updated` handler finds the matching index and updates in place cleanly.
- **Event `sessionId` values**: `spaceAgent.*` uses `space:${spaceId}`, `spaceWorkflow.*` uses `'global'` — consistent with the existing `space-agent-handlers.ts` and `space-workflow-handlers.ts` patterns.
- **Daemon unit tests**: 5 new event emission tests cover created, updated (replace), skipped (no event), workflow created, and bundle with both. Coverage is solid.
- **Frontend unit tests**: `ImportPreviewDialog.test.tsx` covers rendering, conflict selectors, summary counts, spinner, and `onConfirm` payload. `export-import-utils.test.ts` covers filename pattern, sanitization, blob type, DOM interactions, and `afterEach` mock restore.
- **E2E tests**: 4 scenarios cover single-agent export, Export All (with JSON content verification), no-conflict import, conflict import with resolution dropdown, and mixed bundle with both agents and workflows. `injectImportFile` correctly self-cleans `document.createElement` patch.
- **SVG attributes**: `stroke-linecap`, `stroke-linejoin` remain kebab-case in JSX `<path>` elements — this is correct for Preact's SVG handling (Preact forwards SVG attributes as-is).

lsm (commented):

**Recommendation: APPROVE** — both issues from the previous round are now correctly resolved.

---

### Fixes verified

**P1 — Workflow `replace` stale SpaceStore entry**

The handler now captures `replacedOldId = existing.id` before the delete (line 528), stores it as `wfItem.previousId` (line 566), and post-transaction emits `spaceWorkflow.deleted` (old UUID) followed by `spaceWorkflow.created` (new UUID) — lines 611–634. The `spaceStore.deleted` handler at `space-store.ts:418–421` filters by `workflowId`, so the stale entry is removed, and the `spaceStore.created` handler at `space-store.ts:382–389` appends the new entry. No duplicate entries will appear. The event shapes match the canonical `space-workflow-handlers.ts:198–202` pattern exactly (`sessionId: 'global'`, `spaceId`, `workflowId`).

The unit test at `space-export-import-handlers.test.ts:1160–1193` verifies exactly one `spaceWorkflow.deleted` event (carrying `oldWorkflowId`) and exactly one `spaceWorkflow.created` event (carrying a different new UUID), confirming the full round-trip.

**P2 — O(1) workflow lookup**

Line 605 now calls `workflowRepo.getWorkflow(item.id)` directly, matching the O(1) `agentRepo.getById(item.id)` pattern already used for agents on line 588. The `listWorkflows().find()` scan is gone.

---

### No new issues found

- Event emission order (deleted before created) is correct for the replace path.
- The `renamed` and `created` workflow paths correctly fall into the `else` branch and emit only `spaceWorkflow.created`.
- `spaceWorkflow.updated` is never emitted by the import handler — appropriate since replaced workflows have a new UUID that the updated handler's `findIndex` would not match.
- The `agentRepo.getById` / `workflowRepo.getWorkflow` null-guard (`if (!agent) continue` / `if (!workflow) continue`) correctly handles the race where a workflow is somehow absent post-commit.
- All five new event-emission unit tests pass the `await Promise.resolve()` micro-task flush pattern, which is sufficient given the mock `emit` is synchronous underneath.

lsm (commented):

### Third Review Iteration — Verifying Fixes

Both issues from the previous round have been correctly resolved:

**P1 Fix (workflow replace events): Verified correct.**
The handler at `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts` lines 608-634 now emits `spaceWorkflow.deleted` for the old UUID followed by `spaceWorkflow.created` for the new UUID when a workflow is replaced. The `previousId` field on `ImportedItem` (line 101-103) cleanly carries the old UUID from the transaction to the event emission phase. The SpaceStore's `spaceWorkflow.deleted` handler removes the stale entry by filtering on `workflowId`, and the `spaceWorkflow.created` handler adds the new entry with duplicate protection via the `exists` check. No stale duplicates can accumulate.

**P2 Fix (O(1) workflow lookup): Verified correct.**
Line 605 now uses `workflowRepo.getWorkflow(item.id)` which is a direct primary key lookup (`SELECT * FROM space_workflows WHERE id = ?`) instead of the previous `listWorkflows().find()` full scan.

**Test coverage for both fixes:**
The test at line 1160 (`emits spaceWorkflow.deleted (old id) + spaceWorkflow.created (new id) for replaced workflow`) explicitly verifies that:
- Exactly one `spaceWorkflow.deleted` event is emitted with the old workflow ID
- Exactly one `spaceWorkflow.created` event is emitted with a new (different) workflow ID

### Overall Assessment

The full implementation is clean and well-structured:
- Export handlers correctly resolve agent IDs to names and trim unreferenced agents from workflow exports
- Import preview validates cross-references and reports conflicts
- Import execute wraps all mutations in a single SQLite transaction for atomicity
- Event emission happens outside the transaction to avoid holding locks during async broadcast
- Frontend components (`SpaceAgentList`, `WorkflowList`, `ImportPreviewDialog`) handle all states: empty lists, export per-item and all, import with preview and conflict resolution
- `pickImportFile` in `export-import-utils.ts` handles cancel detection with both `oncancel` and the focus-return fallback
- Test coverage is thorough at 1338 lines for the handler tests, plus separate tests for the dialog and download utilities

**Recommendation: APPROVE** — No remaining issues found.

PR #458, opened 2026-03-20
task/add support for minimax m27 highspeed model

lsm:
- **feat: add MiniMax-M2.7-highspeed model support**

### DIFF
--- a/packages/daemon/src/lib/providers/minimax-provider.ts
+++ b/packages/daemon/src/lib/providers/minimax-provider.ts
@@ -63,6 +63,17 @@ export class MinimaxProvider implements Provider {
 			available: true,
 		},
 		{
+			id: 'MiniMax-M2.7',
+			name: 'MiniMax-M2.7',
+			alias: 'minimax-m27',
+			family: 'minimax',
+			provider: 'minimax',
+			contextWindow: 200000,
+			description: 'MiniMax-M2.7 · Flagship Coding Model',
+			releaseDate: '2026-03-01',
+			available: true,
+		},
+		{
 			id: 'MiniMax-M2.7-highspeed',
 			name: 'MiniMax-M2.7-highspeed',
 			alias: 'minimax-m27-fast',
--- a/packages/daemon/src/lib/providers/minimax-provider.ts
+++ b/packages/daemon/src/lib/providers/minimax-provider.ts
@@ -62,6 +62,17 @@ export class MinimaxProvider implements Provider {
 			releaseDate: '2026-01-01',
 			available: true,
 		},
+		{
+			id: 'MiniMax-M2.7-highspeed',
+			name: 'MiniMax-M2.7-highspeed',
+			alias: 'minimax-m27-fast',
+			family: 'minimax',
+			provider: 'minimax',
+			contextWindow: 200000,
+			description: 'MiniMax-M2.7-highspeed · Fast Coding Model (M2.7)',
+			releaseDate: '2026-03-01',
+			available: true,
+		},
 	];
 
 	constructor(private readonly env: NodeJS.ProcessEnv = process.env) {}
--- a/packages/daemon/src/lib/providers/minimax-provider.ts
+++ b/packages/daemon/src/lib/providers/minimax-provider.ts
@@ -69,7 +69,7 @@ export class MinimaxProvider implements Provider {
 			family: 'minimax',
 			provider: 'minimax',
 			contextWindow: 200000,
-			description: 'MiniMax-M2.7-highspeed · Fast Coding Model (M2.7)',
+			description: 'MiniMax-M2.7-highspeed · Fast Coding Model',
 			releaseDate: '2026-03-01',
 			available: true,
 		},
@@ -116,6 +116,9 @@ export class MinimaxProvider implements Provider {
 
 	/**
 	 * Build SDK configuration for MiniMax
+	 * Routes ANTHROPIC_DEFAULT_*_MODEL to the selected model so the SDK uses
+	 * the model the user actually picked. Falls back to MiniMax-M2.5 for
+	 * unrecognised model IDs.
 	 */
 	buildSdkConfig(modelId: string, sessionConfig?: ProviderSessionConfig): ProviderSdkConfig {
 		const apiKey = sessionConfig?.apiKey || this.getApiKey();
@@ -124,15 +127,16 @@ export class MinimaxProvider implements Provider {
 		}
 
 		const baseUrl = sessionConfig?.baseUrl || MinimaxProvider.BASE_URL;
+		const routingModelId = this.ownsModel(modelId) ? modelId : 'MiniMax-M2.5';
 
 		const envVars: Record<string, string> = {
 			ANTHROPIC_BASE_URL: baseUrl,
 			ANTHROPIC_AUTH_TOKEN: apiKey,
 			API_TIMEOUT_MS: '3000000',
 			CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC: '1',
-			ANTHROPIC_DEFAULT_HAIKU_MODEL: 'MiniMax-M2.5',
-			ANTHROPIC_DEFAULT_SONNET_MODEL: 'MiniMax-M2.5',
-			ANTHROPIC_DEFAULT_OPUS_MODEL: 'MiniMax-M2.5',
+			ANTHROPIC_DEFAULT_HAIKU_MODEL: routingModelId,
+			ANTHROPIC_DEFAULT_SONNET_MODEL: routingModelId,
+			ANTHROPIC_DEFAULT_OPUS_MODEL: routingModelId,
 		};
 
 		return {
--- a/packages/daemon/tests/unit/providers/minimax-provider.test.ts
+++ b/packages/daemon/tests/unit/providers/minimax-provider.test.ts
@@ -57,10 +57,11 @@ describe('MinimaxProvider', () => {
 
 			const models = await provider.getModels();
 
-			expect(models).toHaveLength(3);
+			expect(models).toHaveLength(4);
 			expect(models.map((m) => m.id)).toEqual([
 				'MiniMax-M2.5',
 				'MiniMax-M2.5-highspeed',
+				'MiniMax-M2.7',
 				'MiniMax-M2.7-highspeed',
 			]);
 		});
@@ -87,6 +88,7 @@ describe('MinimaxProvider', () => {
 		it('should own minimax- prefixed models', () => {
 			expect(provider.ownsModel('MiniMax-M2.5')).toBe(true);
 			expect(provider.ownsModel('MiniMax-M2.5-highspeed')).toBe(true);
+			expect(provider.ownsModel('MiniMax-M2.7')).toBe(true);
 			expect(provider.ownsModel('MiniMax-M2.7-highspeed')).toBe(true);
 			expect(provider.ownsModel('minimax-m2.5')).toBe(true);
 		});
@@ -123,6 +125,16 @@ describe('MinimaxProvider', () => {
 				ANTHROPIC_DEFAULT_OPUS_MODEL: 'MiniMax-M2.5',
 			});
 			expect(config.isAnthropicCompatible).toBe(true);
+		});
+
+		it('should route to MiniMax-M2.7 when that model is selected', () => {
+			process.env.MINIMAX_API_KEY = 'test-key';
+
+			const config = provider.buildSdkConfig('MiniMax-M2.7');
+
+			expect(config.envVars.ANTHROPIC_DEFAULT_HAIKU_MODEL).toBe('MiniMax-M2.7');
+			expect(config.envVars.ANTHROPIC_DEFAULT_SONNET_MODEL).toBe('MiniMax-M2.7');
+			expect(config.envVars.ANTHROPIC_DEFAULT_OPUS_MODEL).toBe('MiniMax-M2.7');
 		});
 
 		it('should route to MiniMax-M2.7-highspeed when that model is selected', () => {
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

Clean, minimal addition of the MiniMax-M2.7-highspeed model to the existing provider. The implementation follows established patterns exactly and tests are thorough.

### What looks good

- The new model entry in `MODELS` array follows the same structure as existing entries (M2.5, M2.5-highspeed)
- Tests cover the new model in all relevant places: `getModels`, `ownsModel`, static models list, and a dedicated property assertion test
- The alias `minimax-m27-fast` follows the naming convention of existing aliases (`minimax`, `minimax-fast`)
- Context window of 200k matches the provider's `maxContextWindow` capability

### Minor observations (P3)

1. **`buildSdkConfig` hardcodes `MiniMax-M2.5` for all tier defaults** (`packages/daemon/src/lib/providers/minimax-provider.ts:133-135`): The `ANTHROPIC_DEFAULT_*_MODEL` env vars and `getModelForTier` both always return `MiniMax-M2.5`. This is not a bug introduced by this PR -- it was pre-existing behavior -- but worth noting that neither M2.5-highspeed nor M2.7-highspeed are used as defaults for any tier.

2. **Unrelated commit bundled in PR**: The first commit (`fix: start devproxy without changing system proxy`) modifies `dev-proxy.ts` and is unrelated to the MiniMax model addition. The PR description lists it, so this appears intentional, but it would be cleaner as a separate PR.

No issues found with the model addition itself. The change is correct, complete, and well-tested.

lsm (commented):

**Recommendation: APPROVE (with one notable functional gap to consider)**

---

### Summary

The PR adds `MiniMax-M2.7-highspeed` to the provider's static `MODELS` array with correct metadata (alias `minimax-m27-fast`, 200k context window, `available: true`), updates the unit tests to assert the new model count and properties, and separately fixes dev-proxy startup to avoid changing the system proxy. The implementation is minimal, clean, and consistent with how the existing `MiniMax-M2.5-highspeed` model was added.

---

### Functional Gap: `buildSdkConfig` ignores the selected model entirely (pre-existing, but worth flagging in context of this PR)

**File:** `packages/daemon/src/lib/providers/minimax-provider.ts`, lines 128-136

`buildSdkConfig` unconditionally hardcodes all three `ANTHROPIC_DEFAULT_*_MODEL` env vars to `'MiniMax-M2.5'`, regardless of which model is passed in as `modelId`. This means that when a user selects `MiniMax-M2.7-highspeed`, the SDK will still route all internal tier calls (`haiku`, `sonnet`, `opus`) through `MiniMax-M2.5`, not the selected model.

Compare this to `GlmProvider.buildSdkConfig()` (lines 149-162 of `glm-provider.ts`) which derives a `routingModelId` from the `modelId` argument and routes `haiku`/`sonnet` to it, while pinning `opus` to the flagship.

This is a pre-existing design choice in `MinimaxProvider`, but adding a third model amplifies the impact: the alias `minimax-m27-fast` suggests the user expects M2.7-highspeed to be used when it is selected. Currently it will silently use M2.5 instead for all actual inference calls. This gap should be addressed here or documented as intentional policy.

---

### Test Coverage: `buildSdkConfig` not tested for the new model

**File:** `packages/daemon/tests/unit/providers/minimax-provider.test.ts`

The existing `buildSdkConfig` tests only call with `'MiniMax-M2.5'`. No test verifies what happens when `buildSdkConfig('MiniMax-M2.7-highspeed')` is called — which, given the hardcoded routing above, would silently route to M2.5. A test case for the new model would make this behavior explicit and visible.

---

### Minor: `description` field inconsistency in naming

The description for the new model is `'MiniMax-M2.7-highspeed · Fast Coding Model (M2.7)'`, while `MiniMax-M2.5-highspeed` uses `'MiniMax-M2.5-highspeed · Fast Coding Model'` (no version suffix in the label). This is inconsistent but cosmetically minor.

---

### dev-proxy change is clean

The `--as-system-proxy false` flag addition in `packages/daemon/tests/helpers/dev-proxy.ts` is correct and the fix title is accurate. No issues there.

---

### No security concerns

The new model entry uses only static configuration data. No dynamic input, no injection surface introduced.

lsm (commented):

**Recommendation: APPROVE**

The iteration 2 fixes correctly address the issues identified in iteration 1. The implementation is clean and well-tested.

### What was done

1. **New model entry** (`minimax-provider.ts:65-75`): `MiniMax-M2.7-highspeed` added to `MODELS` with appropriate properties (alias `minimax-m27-fast`, 200K context window, March 2026 release date).

2. **Dynamic routing in `buildSdkConfig`** (`minimax-provider.ts:130`): The `routingModelId` now uses `this.ownsModel(modelId)` to pass the user-selected model through to `ANTHROPIC_DEFAULT_*_MODEL` env vars, falling back to `MiniMax-M2.5` for unrecognized IDs. This correctly fixes the hardcoded routing from iteration 1.

3. **Test coverage** (`minimax-provider.test.ts`): New tests cover M2.7-highspeed routing, M2.5-highspeed routing, and the fallback path for unknown model IDs. Static model property assertions verify the new entry's fields.

### Minor observations (P3)

- The `ownsModel` check (`modelId.toLowerCase().startsWith('minimax-')`) is broad — any string starting with `minimax-` will be treated as owned and passed through to the SDK without validation against the known model list. This is the same pattern used before this PR and is consistent with how the provider works, so not a concern for this change.

- The devproxy change (`--as-system-proxy false`) is a separate fix bundled into this PR. It is correct and harmless but ideally would be in its own PR for clean history. Already committed separately so no action needed.

No issues found that would block merge.

lsm (commented):

Recommendation: **APPROVE** (with minor observations noted below)

---

### Summary

The core goal — add MiniMax-M2.7-highspeed and fix `buildSdkConfig` to route to the user-selected model — is correctly implemented. The fix in iteration 2 is sound, tests are meaningful, and the dev-proxy change is an independent improvement. A few minor observations are worth tracking, but none block merging.

---

### What was verified

**Model registration** (`minimax-provider.ts` lines 65–75): The new entry is consistent with the two existing models — same `family`, `provider`, `contextWindow`, and `available` fields. The `releaseDate` of `2026-03-01` and alias `minimax-m27-fast` are internally consistent. No issues.

**`buildSdkConfig` routing** (line 130):
```ts
const routingModelId = this.ownsModel(modelId) ? modelId : 'MiniMax-M2.5';
```
Correctly delegates to `ownsModel` (which does a case-insensitive `startsWith('minimax-')` check) so any future MiniMax model ID is handled automatically. The fallback to `MiniMax-M2.5` for unrecognised IDs mirrors the GLM pattern.

**Tests** (`minimax-provider.test.ts`): The three new `buildSdkConfig` tests (M2.7-highspeed routing, M2.5-highspeed routing, unknown fallback) directly exercise the bug that was fixed in iteration 1. The static-model property test also covers the new model's fields. Coverage is adequate.

**Dev-proxy change** (`tests/helpers/dev-proxy.ts`): Adding `--as-system-proxy false` is an orthogonal fix that prevents the dev-proxy from modifying system-level proxy settings during tests. This is a clean, self-contained improvement.

---

**P3 – `ANTHROPIC_DEFAULT_OPUS_MODEL` diverges from GLM precedent without explanation**

In `glm-provider.ts` line 162 the GLM implementation intentionally keeps `ANTHROPIC_DEFAULT_OPUS_MODEL` pinned to the flagship model (`glm-5`) even when a turbo/highspeed model is selected, with an explicit comment:

```ts
// Route Sonnet/Haiku to selected model, keep Opus on flagship glm-5
ANTHROPIC_DEFAULT_HAIKU_MODEL: routingModelId,
ANTHROPIC_DEFAULT_SONNET_MODEL: routingModelId,
ANTHROPIC_DEFAULT_OPUS_MODEL: 'glm-5',
```

The MiniMax implementation routes all three tiers — including Opus — to `routingModelId`. That may be the right call for MiniMax (the SDK's Opus fallback path using the highspeed model is not necessarily wrong), but the asymmetry is undocumented. A brief comment in `buildSdkConfig` explaining why all three tiers follow the selected model would make the intent clearer for future maintainers.

**P3 – `getModelForTier` comment is now slightly stale**

`packages/daemon/src/lib/providers/minimax-provider.ts` line 111 still reads:
> All tiers use MiniMax-M2.5 (flagship model)

This is accurate for `getModelForTier` (which is used for tier-based fallbacks/title generation), but a reader arriving at the file fresh might wonder why `buildSdkConfig` behaves differently. A small clarification — "tier defaults and title generation always use the flagship; `buildSdkConfig` routes to the session-selected model" — would help.

**P3 – No test for `MiniMax-M2.5` routing after the fix**

The existing test at line 111 does a full `toEqual` snapshot that already covers M2.5 routing implicitly, so this is not a gap in practice. But the new tests only spot-check `ANTHROPIC_DEFAULT_*_MODEL` keys while the M2.5 test uses a full `toEqual`. Consistent style across the new tests would be a minor improvement.

---

### Decision

No P0/P1/P2 issues. The implementation is correct, the fallback logic is sound, the tests exercise the fixed code path, and the dev-proxy change is clean. The observations above are all P3 documentation/style nits.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

The PR adds two new MiniMax models — `MiniMax-M2.7` (alias `minimax-m27`) and `MiniMax-M2.7-highspeed` (alias `minimax-m27-fast`) — and fixes a pre-existing bug where `buildSdkConfig` always hardcoded `MiniMax-M2.5` regardless of the user's selected model. The implementation mirrors the M2.5/M2.5-highspeed pattern and all 25 tests pass.

---

### Findings

#### P3 (Nit) — Divergence from GLM in ANTHROPIC_DEFAULT_OPUS_MODEL routing policy

**File:** `packages/daemon/src/lib/providers/minimax-provider.ts`, line 150

The GLM provider intentionally pins `ANTHROPIC_DEFAULT_OPUS_MODEL` to the flagship model (`glm-5`) even when a non-flagship model is selected, with a comment in `glm-provider.ts` explaining this as a deliberate policy (`getModelForTier` comment, line 119-122):

> "Always pins to glm-5 regardless of which GLM model is active in the session. This is an intentional policy: tier fallbacks and title generation use the flagship model (glm-5), not the session model."

MiniMax's `buildSdkConfig` now routes all three tier env vars — including Opus — to `routingModelId`, meaning `MiniMax-M2.7-highspeed` sessions will also use that fast model for any internal Opus-tier calls the SDK makes. The GLM pattern would instead keep Opus pinned to `MiniMax-M2.5` (the flagship). This difference is a conscious choice but the comment on line 122 of `minimax-provider.ts` still says "All tiers use MiniMax-M2.5 (flagship model)" which is now stale and misleading.

**Suggested fix:** Update the `getModelForTier` JSDoc comment (line 122) to reflect the actual behavior: all tiers route to MiniMax-M2.5, which is used as the tier fallback (not related to `buildSdkConfig` routing). Separately, decide whether Opus-tier internal calls should be pinned to the flagship or follow the selected model — but document that decision in the code.

#### P3 (Nit) — `getTitleGenerationModel` and `getModelForTier` still return M2.5, not M2.7

**File:** `packages/daemon/src/lib/providers/minimax-provider.ts`, lines 125, 171

Now that M2.7 is the newer flagship, both `getModelForTier` and `getTitleGenerationModel` continue returning `MiniMax-M2.5`. This matches the GLM pattern of pinning tier/title calls to the flagship, but since both M2.5 and M2.7 are listed as "Flagship Coding Model" in their descriptions, it is unclear whether M2.7 should now be the preferred flagship for these purposes. The tests at lines 104-109 and 206-208 confirm this is intentional, but a brief comment explaining the choice (e.g., "intentionally kept at M2.5 for stability") would improve clarity.

---

### What's Correct

- New model entries are well-formed with accurate `id`, `name`, `alias`, `family`, `provider`, `contextWindow`, `releaseDate`, and `available` fields.
- The `buildSdkConfig` routing fix using `ownsModel` as the guard is clean and consistent with the GLM pattern (modulo the Opus difference noted above).
- `ownsModel` correctly handles the new model IDs via the existing case-insensitive `minimax-` prefix check.
- All aliases (`minimax-m27`, `minimax-m27-fast`) are unique and do not collide with existing aliases.
- Test coverage is thorough: model property checks, routing for each of the four models, fallback behavior, and key override scenarios are all covered.
- The dev-proxy fix (starting without changing system proxy via `--as-system-proxy false`) is a clean operational improvement unrelated to the model additions.

---

No P0, P1, or P2 issues found. The two P3 nits are limited to stale/ambiguous comments and a minor undocumented design divergence from the GLM provider. The implementation is correct and production-ready.

lsm (commented):

**Recommendation: APPROVE**

### Summary

Clean, well-structured addition of MiniMax-M2.7 and MiniMax-M2.7-highspeed models. The implementation mirrors the existing M2.5/M2.5-highspeed pattern, the `buildSdkConfig` routing fix is correct, and test coverage is thorough (25 tests, all passing).

### What was reviewed

The PR adds 4 commits:
1. `fix: start devproxy without changing system proxy` — infrastructure fix
2. `feat: add MiniMax-M2.7-highspeed model support` — initial highspeed model
3. `fix: route buildSdkConfig to selected MiniMax model instead of hardcoding M2.5` — bug fix for model routing
4. `feat: add MiniMax-M2.7 flagship model (alias minimax-m27)` — standard model

### Findings

**No blocking issues found.** The implementation is correct and complete:

- **Model definitions** (`minimax-provider.ts` lines 66-86): Both new models have proper `id`, `name`, `alias`, `family`, `provider`, `contextWindow`, `description`, `releaseDate`, and `available` fields. The aliases (`minimax-m27`, `minimax-m27-fast`) follow the established naming convention.

- **SDK routing fix** (line 141): `const routingModelId = this.ownsModel(modelId) ? modelId : 'MiniMax-M2.5'` correctly routes the selected model to the SDK env vars instead of always hardcoding M2.5. The `ownsModel` check (case-insensitive `minimax-` prefix match) naturally covers all 4 models.

- **Fallback behavior**: Unrecognized model IDs fall back to `MiniMax-M2.5`, which is safe and tested.

- **Test coverage**: Tests verify all 4 model property sets, SDK routing for each model, the fallback path, and static model definitions. Coverage is thorough.

**P3 observations (non-blocking):**

1. `getModelForTier()` (line 124) and `getTitleGenerationModel()` (line 170) still hardcode `MiniMax-M2.5`. This is consistent with how GLM handles it (always returns `glm-5`), so it follows established patterns. If MiniMax-M2.7 becomes the preferred flagship, these could be updated in a follow-up.

2. The M2.5 description says "Flagship Coding Model" and so does M2.7's. If M2.7 is the newer flagship, consider differentiating the descriptions in a follow-up (e.g., "MiniMax-M2.5 · Coding Model" vs "MiniMax-M2.7 · Flagship Coding Model").

3. The PR description on GitHub only mentions the highspeed model — it should be updated to also mention the standard M2.7 model that was added in the latest commit.

PR #459, opened 2026-03-20
task/fix agent session crash when switching modelprovid

lsm:
- **fix: clear sdkSessionId when switching to a different provider**


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-targeted fix for a real crash scenario. The implementation is correct and the tests are thorough.

### What was reviewed

The fix addresses the case where switching from one provider to another (e.g., Anthropic to GLM, or Anthropic to Copilot) would pass the old provider's `sdkSessionId` as a `resume` parameter to the new provider's SDK subprocess. Since the session file on disk belongs to the old provider, the new subprocess would hang until the 15-second startup timeout.

### Assessment

**Correctness**: The fix correctly captures the `providerChanged` flag *before* mutating `session.config.provider` (line 186 of `model-switch-handler.ts`), which is important since the comparison would be meaningless after the assignment. Both the "query not started" and "query running" branches are handled identically, which is the right behavior.

**DB persistence**: The `sdkSessionId: undefined` is correctly persisted to the DB via the spread pattern `...(providerChanged ? { sdkSessionId: undefined } : {})`. The session repository at line 146-148 of `session-repository.ts` uses `'sdkSessionId' in updates` to detect the key and converts `undefined` to SQL `NULL` via `?? null`, so this works as intended.

**Tests**: The 6 new tests cover the important cases:
- Provider change with query not started (glm -> anthropic)
- Provider change with query running (glm -> anthropic) -- also verifies restart is called
- Cross-provider switch (anthropic -> copilot)
- Same-provider model switch preserves sdkSessionId (2 cases: anthropic and copilot)
- Already-unset sdkSessionId with provider change (no crash, still succeeds)

### Minor observation (P3)

The in-memory guard (`if (providerChanged && session.sdkSessionId)`) is stricter than the DB persist (which always writes `sdkSessionId: undefined` when `providerChanged` is true, regardless of current value). This is harmless -- writing NULL to an already-NULL column is a no-op in SQLite -- but the asymmetry is slightly surprising. The last test ("does nothing when sdkSessionId is already unset") documents this behavior correctly.

No issues found. The fix is minimal, focused, and well-tested.

lsm (commented):

---

### Summary

The fix is correct and well-targeted. Clearing `sdkSessionId` before the new SDK subprocess starts prevents the 15-second startup timeout that occurred when the new provider received an incompatible session resume ID. The implementation is consistent with existing patterns in `agent-session.ts` and `query-runner.ts`. 6 unit tests cover the critical behavioral variants.

---

### Correctness: The Fix Works

The ordering is correct in both branches:

**"Query is running" path** (`packages/daemon/src/lib/agent/model-switch-handler.ts`, lines 234-246):
- `session.sdkSessionId = undefined` is set in-memory first
- `db.updateSession(...)` is called
- `lifecycleManager.restart()` is called last

This is important because `restart()` in `query-lifecycle-manager.ts` (line 167) guards `validateAndRepairSDKSession` behind `if (session.sdkSessionId)`. Since the in-memory value is already `undefined` before `restart()` is invoked, the repair is correctly skipped. The new subprocess then starts without a stale `resume` option.

**DB persistence** is also correct. The spread `...(providerChanged ? { sdkSessionId: undefined } : {})` adds the key to the update object. The repository uses `'sdkSessionId' in updates` (line 146 of `session-repository.ts`), not a truthiness check, so `undefined` triggers `sdk_session_id = NULL` in SQLite. Verified with a direct JS eval.

---

### P2: Minor Code Asymmetry — Unnecessary DB Write When Already `undefined`

**File:** `packages/daemon/src/lib/agent/model-switch-handler.ts`, lines 196-208 (and 234-246)

The in-memory guard is:
```ts
if (providerChanged && session.sdkSessionId) {
    session.sdkSessionId = undefined;
}
```

But the DB spread is unconditional on provider change:
```ts
...(providerChanged ? { sdkSessionId: undefined } : {})
```

When `providerChanged` is true and `session.sdkSessionId` is already `undefined`, the DB write fires unnecessarily. This is harmless (a NULL write over an already-NULL column) but the behavior is inconsistent. The last unit test ("does nothing when sdkSessionId is already unset and provider changes") documents this, but the comment says "Does nothing" while the assertion contradicts that by expecting the DB write. A more accurate guard would be:

```ts
if (providerChanged) {
    session.sdkSessionId = undefined;
    // DB spread also runs unconditionally
}
```

Or conversely, gate the DB spread to match the in-memory guard. Either way, this is cosmetic — no behavior impact.

---

### P3: Missing Test — Transport-Not-Ready + Provider Change

**File:** `packages/daemon/tests/unit/agent/model-switch-handler.test.ts`

The new test suite covers:
- `queryObject: null` + provider change
- `queryObject` running (`firstMessageReceived: true`) + provider change
- Same-provider variants

But there is no test for **`queryObject` non-null but `firstMessageReceived: false`** (transport not ready) combined with a provider change. This combination takes the `if (!queryObject || !transportReady)` branch and would clear `sdkSessionId` correctly, but it's untested. The existing test at line 387 only covers this path for same-provider switches.

---

### The Dev-Proxy Change

Adding `--as-system-proxy false` to `packages/daemon/tests/helpers/dev-proxy.ts` is a clean fix. Starting devproxy without overriding the system proxy avoids side-effects on developer machines and CI environments where the system proxy setting is not expected to change during test runs. No concerns here.

---

### No Issues Found

- No security vulnerabilities
- No data loss risks
- No race conditions introduced
- Consistent with existing `sdkSessionId`-clearing patterns in `query-runner.ts` and `agent-session.ts`
- Tests are meaningful and verify both the in-memory and DB state

lsm (commented):

**Recommendation: APPROVE**

Both P2 and P3 issues from the prior review were addressed correctly.

### P2 fix — removed the `&& session.sdkSessionId` in-memory guard

The asymmetry is gone. In both branches (`!queryObject || !transportReady` and the query-running `else`), the in-memory assignment `session.sdkSessionId = undefined` and the DB spread `...(providerChanged ? { sdkSessionId: undefined } : {})` now fire together unconditionally when `providerChanged` is true. The `updateSession` implementation uses `'sdkSessionId' in updates` to detect the key (not a truthiness check), so spreading `{ sdkSessionId: undefined }` correctly writes `NULL` to `sdk_session_id` in SQLite, and omitting the key preserves the existing value. The fix is correct.

### P3 fix — transport-not-ready + provider-changed test added

The new test (`clears sdkSessionId when transport not ready (queryObject set but firstMessageReceived=false) and provider changes`) creates the handler with `{ firstMessageReceived: false }` and a non-null `queryObject`, then asserts:
- `session.sdkSessionId` is cleared in memory
- DB spread includes `sdkSessionId: undefined`
- `restartSpy` is NOT called

This correctly covers the formerly untested path where `queryObject` is non-null but `transportReady` is false, routing through the `!queryObject || !transportReady` branch. The assertion style is consistent with the other tests in the describe block.

### Other observations

- All 33 tests pass (`bun test` confirms 0 failures).
- The `toHaveProperty('sdkSessionId', undefined)` / `not.toHaveProperty('sdkSessionId')` pairing correctly distinguishes "key present with undefined value" from "key absent" — which is the meaningful distinction given the `'sdkSessionId' in updates` guard in `updateSession`.
- The dev-proxy fix (`--as-system-proxy false`) is orthogonal to the main change and looks correct.
- No dead code, no over-engineering, no security concerns introduced.

lsm (commented):

**Recommendation: APPROVE**

### Re-review after feedback addressed

Both issues from the previous review have been correctly resolved in commit `168ed8fc`:

**P2 (resolved):** The asymmetry between the in-memory guard and DB spread has been fixed. Previously the in-memory clear had `if (providerChanged && session.sdkSessionId)` while the DB spread was unconditional on `providerChanged`. Now both paths simply use `if (providerChanged)`, making the behavior consistent and easier to reason about. This is the right call -- assigning `undefined` to an already-`undefined` field is a no-op in JS, just as writing NULL over NULL is a no-op in SQLite.

**P3 (resolved):** The missing test for the transport-not-ready + provider-change path has been added (line 687-701 of `model-switch-handler.test.ts`). It correctly verifies that:
- The `!transportReady` branch is taken (not the restart branch)
- `sdkSessionId` is cleared both in-memory and in DB
- `restart()` is NOT called

The test description for the "already unset" case (line 672) was also updated to accurately reflect that the assignment now runs unconditionally when `providerChanged` is true.

### No new issues found

The three changed files are clean:
- `model-switch-handler.ts`: Fix logic is correct, symmetric, well-commented
- `model-switch-handler.test.ts`: 7 tests covering all provider-switch variants
- `dev-proxy.ts`: `--as-system-proxy false` is a sensible improvement for test isolation

lsm (commented):

### Summary

The approach of auto-recovering from SDK startup timeouts (instead of preemptively clearing `sdkSessionId` on model switch) is fundamentally sound. The key insight — that `sdkSessionId` is needed for conversation context loading, so clearing it eagerly destroys history — is correct, and the reactive recovery design preserves the user experience.

### Issues

**P1: No retry limit — auto-recovery can loop indefinitely**

`onStartupTimeoutAutoRecover` is a permanent method on `AgentSession`. If the SDK binary is broken, network is down, or credentials are invalid in a way that manifests as a startup timeout (not an auth error), the auto-recovery will fire every ~15.3 seconds (15s timeout + 300ms delay) forever, spawning new SDK subprocesses each time.

The fix should be straightforward: add a retry counter (e.g., `autoRecoverAttempts`) that is incremented on each recovery and checked in `canAutoRecover`. After 1-2 retries, fall through to the normal error path so the user sees the failure. Reset the counter on successful first message received.

`packages/daemon/src/lib/agent/query-runner.ts`, lines 340-353
`packages/daemon/src/lib/agent/agent-session.ts`, lines 786-790

**P1: Auto-recovery tests don't test actual code paths**

All 9 tests in the `QueryRunner startup timeout auto-recovery` describe block (lines 1202-1341 of `query-runner.test.ts`) are pure logic simulations — they duplicate the `if` conditions from the source code in local variables and assert on those local variables. They do not instantiate `QueryRunner`, do not call any method on it, and would continue passing even if the actual catch block logic were completely removed or broken.

The existing integration-style tests (like the `runQuery() finally block close() behaviour` suite just above) show it is possible to exercise `runQuery()` via `runner.start()` even without credentials (auth check fails early, but catch/finally still execute). The auto-recovery tests should follow the same pattern: create a `QueryRunner` with `onStartupTimeoutAutoRecover` set on the context, trigger a startup-timeout-like error, and verify that `messageQueue.clear()` was NOT called and `onStartupTimeoutAutoRecover` WAS invoked.

`packages/daemon/tests/unit/agent/query-runner.test.ts`, lines 1202-1341

**P2: `recover.call(this.ctx)` binding is fragile and unnecessary**

Line 350: `recover.call(this.ctx)` uses `.call()` to bind `this` to the QueryRunnerContext. Since `this.ctx` IS the AgentSession and `onStartupTimeoutAutoRecover` is a regular method on AgentSession, this works correctly today. However, the indirection through a captured function reference + `.call()` is unnecessarily fragile. Simply calling `this.ctx.onStartupTimeoutAutoRecover!()` would be clearer and functionally equivalent, without the risk of someone later breaking the binding by changing how the reference is captured.

`packages/daemon/src/lib/agent/query-runner.ts`, line 350

- The `messageQueue.stop()` in the finally block does not destroy queued messages (only sets `running = false`), so the preserved messages survive through the `stop()` → `start()` cycle during auto-recovery. This is correct.
- The `sdkSessionId` is cleared before scheduling auto-recovery, ensuring the retry starts fresh without the stale session file. This is the core correctness property.
- The 300ms `setTimeout` delay ensures the finally block completes before recovery starts, preventing state races. Reasonable design.
- The model-switch-handler changes (adding `newProvider` parameter, removing sdkSessionId clearing) are clean and well-tested with good provider routing coverage.

The infinite retry loop (P1) is a real risk that should be addressed before merge. The test quality issue (P1) means the new behavior has no real automated verification.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — the implementation has a P0 infinite loop bug and P1-level test coverage gaps.

---

## Summary

The approach is conceptually sound: instead of preemptively clearing `sdkSessionId` (which destroys conversation context), the catch block now detects startup timeouts and schedules a transparent retry with the stale `sdkSessionId` already cleared. For cross-provider switches, the 15s timeout will always fire, recovery clears the incompatible session file reference, and the retry starts fresh. For same-provider switches, no timeout occurs and context is preserved. The wiring between `QueryRunnerContext`, `QueryRunner`, and `AgentSession` is clean.

---

## P0 — Infinite Retry Loop with No Backoff or Limit

**File:** `packages/daemon/src/lib/agent/query-runner.ts`, lines 340–352

The auto-recovery re-calls `startStreamingQuery()`, which runs a new `runQuery()`. If that query also times out on startup (e.g., network unreachable, SDK binary missing, or any persistent issue unrelated to `sdkSessionId`), `canAutoRecover` is `true` again and another `setTimeout(recover, 300ms)` is scheduled. There is no retry counter, no maximum attempt limit, and no backoff.

A session that is permanently broken will spin forever: 15s timeout, 300ms delay, retry, 15s timeout, 300ms delay, ... consuming resources silently, never surfacing an error to the user.

The original timeout path (before this PR) would show the error to the user after one attempt. The new path suppresses it indefinitely.

**Fix needed:** Add a retry counter (e.g., limit to 1 auto-recovery attempt). On the second consecutive startup timeout, fall through to the normal error handler. Example:

```typescript
// In QueryRunnerContext:
startupTimeoutAutoRecoverAttempts?: number;

// In the catch block:
const attempts = (this.ctx.startupTimeoutAutoRecoverAttempts ?? 0) + 1;
const canAutoRecover = isStartupTimeout && !this.ctx.isCleaningUp()
  && !!this.ctx.onStartupTimeoutAutoRecover && attempts <= 1;
if (canAutoRecover) {
  this.ctx.startupTimeoutAutoRecoverAttempts = attempts;
  // schedule recovery...
} else {
  this.ctx.startupTimeoutAutoRecoverAttempts = 0;
  // normal error path...
}
```

---

## P1 — Tests Don't Exercise the Actual QueryRunner

**File:** `packages/daemon/tests/unit/agent/query-runner.test.ts`, lines 1202–1341

All 9 new tests are tautological: they declare local boolean variables, reproduce the `if/else` logic from the implementation inline, and assert on those local variables. The actual `QueryRunner` class is never instantiated. No mock of `QueryRunnerContext` with `onStartupTimeoutAutoRecover` is constructed. No `runQuery` private method is exercised (even via the public `start()` path).

These tests will pass regardless of any future change to the `QueryRunner` implementation, including deleting the feature entirely.

The existing test suite in this file does have patterns for mocking `QueryRunnerContext` and calling `runner.start()` (see `describe('QueryRunner environment variable handling')`). The new tests should follow the same pattern: construct a real `QueryRunner` with a mocked context that throws a startup-timeout error, verify that `messageQueue.clear()` is/is not called, and verify that `onStartupTimeoutAutoRecover` is/is not invoked.

---

## P2 — `stateManager.setIdle()` Called Twice in Auto-Recovery Path

**File:** `packages/daemon/src/lib/agent/query-runner.ts`, line 429 and line 482–484

In the auto-recovery path (`canAutoRecover=true`, `!isAbortError`):
1. `await stateManager.setIdle()` is called on line 429 (outside the `if/else` block).
2. `finally` block also calls `await stateManager.setIdle()` on line 483 (if `!isCleaningUp()`).

The double call is benign since `setIdle()` is idempotent, but it's an unexpected side effect. If `stateManager` ever gains non-idempotent behavior on the idle transition (e.g., emitting a state-change event), this will emit two events to connected clients. The catch block should skip the explicit `setIdle()` call when `canAutoRecover=true` since the finally block already handles it:

```typescript
if (!canAutoRecover) {
  await stateManager.setIdle();
}
```

---

## P2 — `onStartupTimeoutAutoRecover` Is Always Present But Marked Optional

**File:** `packages/daemon/src/lib/agent/query-runner.ts`, line 91

The method is declared `optional` in `QueryRunnerContext`, but `AgentSession` always provides it (it's part of the `implements QueryRunnerContext` contract). The optional marker is used to enable partial mock objects in tests, but those mock objects all have the same flaw noted in P1 (they aren't actually exercised). In production, `onStartupTimeoutAutoRecover` is always defined, which means auto-recovery will always trigger on startup timeout — not just during provider switches. The comment on the interface says `// Optional auto-recovery callback invoked when the SDK startup timer fires. If set, the catch block skips messageQueue.clear()`, which implies optionality is meaningful, but it isn't in practice. This design inconsistency should be documented or the method should be made required.

---

## P3 — Cosmetic: Blank Lines Added in model-switch-handler.ts

**File:** `packages/daemon/src/lib/agent/model-switch-handler.ts`, lines 187, 217

The diff adds two blank lines (before the `// Only pass serializable fields` comments) but no functional change. These appear to be incidental whitespace edits from the editor. Not wrong, just noise in the diff.

lsm (commented):

**Recommendation: APPROVE**

The previous review concerns have been addressed well. The implementation is correct and the tests are meaningfully improved. A few observations remain.

---

### What was Fixed

**Retry limit via counter** — `startupTimeoutAutoRecoverAttempts` is correctly added to `QueryRunnerContext` and `AgentSession`, incremented before scheduling recovery, reset to 0 on successful first message, and reset to 0 on exhaustion. The `startupRecoverAttempts <= 1` guard prevents infinite loops.

**Bound callback** — `this.ctx.onStartupTimeoutAutoRecover!()` is now called correctly. The `onStartupTimeoutAutoRecover` method on `AgentSession` is an arrow-method-style implementation that uses `this` correctly.

**Double setIdle eliminated** — `stateManager.setIdle()` is no longer called in the catch block for the recovery path; it falls through to the finally block only.

**Tests are real integration tests** — `buildSpy.mockRejectedValue` causes the actual `runQuery()` catch block to fire, verifying the real conditional logic rather than just calling functions in isolation. The 6 new tests cover the material branches.

---

### Remaining Observations

**P2 — Queue preservation and AbortError are independently gated**

`packages/daemon/src/lib/agent/query-runner.ts` lines 325 and 341:

```typescript
if (!isStartupTimeout || !this.ctx.onStartupTimeoutAutoRecover) {
    messageQueue.clear();   // preservation gate: startup + handler registered
}

if (!isAbortError) {
    // canAutoRecover check — recovery only happens here
}
```

These two gates are independent. If `isStartupTimeout && onStartupTimeoutAutoRecover && isAbortError` were all simultaneously true, the queue would be preserved but no recovery would be scheduled, leaving stale preserved messages in the queue for whatever happens next. In the current code this triple intersection cannot occur because the startup-timeout error is thrown with a plain `new Error(...)` (not an AbortError), but the asymmetry is a latent trap if the throw site ever changes. A low-risk defensive fix would be to add `&& !isAbortError` to the preservation condition:

```typescript
if (!isStartupTimeout || isAbortError || !this.ctx.onStartupTimeoutAutoRecover) {
    messageQueue.clear();
}
```

**P3 — isCleaningUp guard in onStartupTimeoutAutoRecover is correct but untested for the race window**

`packages/daemon/src/lib/agent/agent-session.ts` line 788: the `if (this.isCleaningUp()) return` guard is there to protect the 300ms window between scheduling and execution. The existing test for the isCleaningUp path (`should NOT invoke onStartupTimeoutAutoRecover when isCleaningUp returns true`) covers the case where `isCleaningUp` is already true at schedule time (caught by `canAutoRecover`), not the case where cleanup begins during the 300ms delay. This is a minor gap — the race is benign since `startStreamingQuery()` defensively checks state — but worth noting.

**P3 — Error category for startup timeout hits TIMEOUT branch**

When recovery is exhausted and the startup timeout error surfaces normally, the `errorMessage.includes('timeout')` check at line 415 will categorize it as `ErrorCategory.TIMEOUT` (since the message is 'SDK startup timeout - query aborted'). This was present before this PR, but now that the startup timeout is a recoverable special case it might be worth giving it its own category or at minimum verifying the TIMEOUT category maps to the right user-facing message. Not blocking.

---

### Summary

The implementation correctly addresses all P0/P1/P2 findings from the previous review. The retry limit is properly enforced, the double-setIdle is gone, and the tests exercise the actual catch block. The one remaining structural note (P2) is the independent gating of queue preservation and abort detection — harmless in the current throw path but worth tightening.

lsm (commented):

**Recommendation: APPROVE**

### Re-review: All P0/P1/P2 issues from prior reviews have been addressed

The implementation correctly fixes the agent session crash when switching model/provider by using transparent auto-recovery instead of preemptively clearing `sdkSessionId`.

### Verification of fixes

**P0 Infinite retry loop (FIXED):** The `startupTimeoutAutoRecoverAttempts` counter in `QueryRunnerContext` is incremented before scheduling recovery (line 356 of `query-runner.ts`) and checked with `startupRecoverAttempts <= 1` (line 353). After 1 failed recovery attempt, the error surfaces normally via `handleError()`. The counter is reset to 0 in two places: (1) on first message received (line 285, indicating successful startup), and (2) when the retry limit is exhausted and the error path runs (line 370, so future sessions can recover again). This prevents infinite loops.

**P1 Tautological tests (FIXED):** The 9 local-variable-only tests have been replaced with 6 real integration tests that instantiate `QueryRunner`, call `runner.start()`, and verify actual behavior through the `runQuery()` catch block. The tests exercise `buildSpy.mockRejectedValue(new Error('SDK startup timeout - query aborted'))` to trigger the startup timeout path without waiting 15 seconds. Key behaviors verified:
- `messageQueue.clear()` is NOT called when `onStartupTimeoutAutoRecover` is registered
- `messageQueue.clear()` IS called when `onStartupTimeoutAutoRecover` is absent
- `onStartupTimeoutAutoRecover` is deferred by ~300ms via `setTimeout`
- Recovery is blocked when `isCleaningUp()` returns true
- Recovery is blocked when retry limit (1) is already reached, and `handleError` is called instead
- `startupTimeoutAutoRecoverAttempts` is incremented when recovery is scheduled

**P2 Double setIdle (FIXED):** In the auto-recovery path, `stateManager.setIdle()` is only called in the `else` branch (non-recovery, line 445). The recovery path skips it (line 366 comment: "setIdle is handled by the finally block"). The finally block provides the single `setIdle()` call at line 500.

**P2 `.call()` binding (FIXED):** Changed to `this.ctx.onStartupTimeoutAutoRecover!()` (line 362), which is the natural call pattern since `this.ctx` is the AgentSession instance.

### Remaining observations (P3, non-blocking)

1. **Double setIdle in the non-recovery error path** (lines 445 + 500): `stateManager.setIdle()` is called both in the catch block's `else` branch and in the finally block. This is pre-existing behavior (not introduced by this PR) and `setIdle()` appears to be idempotent, so it's harmless. Mentioning for awareness only.

2. **`onStartupTimeoutAutoRecover` is optional in the interface but always provided by AgentSession**: This means auto-recovery fires for all startup timeouts, not just provider switches. This is actually reasonable behavior -- if the SDK times out for any reason, one retry without resume is a sensible recovery strategy, and the retry limit prevents abuse.

### All 73 tests pass

Verified via `bun test tests/unit/agent/query-runner.test.ts` -- 73 pass, 0 fail, 106 expect() calls.

lsm (commented):

**Recommendation: APPROVE**

### P2 fix verified: AbortError added to queue-preservation guard

The latest commit (`0493587`) correctly addresses the latent path identified in the previous review. The queue-preservation condition at line 325 of `packages/daemon/src/lib/agent/query-runner.ts` is now:

```typescript
if (!isStartupTimeout || isAbortError || !this.ctx.onStartupTimeoutAutoRecover) {
    messageQueue.clear();
}
```

This closes the gap where `isStartupTimeout && isAbortError && onStartupTimeoutAutoRecover` could leave preserved messages in the queue with no recovery scheduled (since the recovery path is gated behind `!isAbortError`). The three-way OR correctly ensures the queue is cleared in every scenario where recovery will not run.

### New test covers the branch

The test at line 764 (`should call messageQueue.clear() on startup-timeout AbortError even when handler is registered`) constructs an Error with `name = 'AbortError'` and the startup timeout message, verifies `clearSpy` is called, and exercises the actual `QueryRunner` catch block via `runner.start()`. The test is well-commented explaining why this branch matters.

### Full picture remains sound

- Retry limit (1 attempt) prevents infinite loops
- Counter is reset on successful first message or after exhaustion
- `setIdle()` is not double-called in the recovery path
- All 7 tests in the `startup timeout auto-recovery` describe block exercise the real `runQuery()` catch block through `buildSpy.mockRejectedValue()`

No new issues found. The fix is minimal and targeted.

lsm (commented):

**Recommendation: APPROVE**

### P2 Fix Verification

The P2 feedback was: the queue-preservation gate and the AbortError gate were independent, creating a path where an AbortError-flavored startup-timeout could preserve the queue while recovery would not run (because recovery is gated behind `!isAbortError`), leaving stale messages in the queue.

**The fix is correct.** The new gate expression at line 325 of `query-runner.ts`:

```ts
if (!isStartupTimeout || isAbortError || !this.ctx.onStartupTimeoutAutoRecover) {
    messageQueue.clear();
}
```

All four critical scenarios are correctly handled:

- **Normal startup timeout + handler registered:** all three operands are false → gate is false → queue is preserved → recovery runs. Correct.
- **AbortError startup timeout + handler registered (the P2 fix):** `isAbortError=true` makes the gate true → queue IS cleared. Recovery does not run because `!isAbortError` blocks the recovery block. No stale messages. Correct.
- **Startup timeout + no handler:** `!handler=true` → gate is true → queue IS cleared. Correct.
- **Non-startup error:** `!isStartupTimeout=true` → gate is true → queue IS cleared. Correct.

### Test Coverage

The new test suite in `query-runner.test.ts` (lines 720–838) directly exercises the P2 fix with a dedicated test case: "should call messageQueue.clear() on startup-timeout AbortError even when handler is registered". The test correctly:

- Constructs an error with `name = 'AbortError'` AND the startup-timeout message string
- Registers `onStartupTimeoutAutoRecover`
- Asserts that `clearSpy` was called

Five additional tests cover the recovery scheduling, the retry limit, `isCleaningUp` guard, and counter increment. Coverage is thorough.

When the error arrives as an AbortError (the P2 scenario), the `startupTimeoutAutoRecoverAttempts` counter is not reset because the counter-reset is inside the `else` branch which is itself inside `if (!isAbortError)`. The counter remains at whatever value it had before. This is benign — the counter starts at 0 on session construction and AbortError paths never increment it, so a subsequent normal timeout will correctly see attempts=0 and fire recovery. No action required.

The fix precisely addresses the reported concern and is accompanied by regression tests that will catch any future regression.

PR #461, opened 2026-03-20
fix: handle thinking blocks in SDK title generation

lsm:
## Summary

- Fix empty response error in SDK title generation when model returns only thinking blocks
- Added fallback to extract title from thinking blocks when text blocks are empty

## Test plan

- [ ] Verify title generation works with models that have thinking/reasoning enabled

### DIFF
--- a/packages/daemon/src/lib/session/session-lifecycle.ts
+++ b/packages/daemon/src/lib/session/session-lifecycle.ts
@@ -841,13 +841,33 @@ ${messageText.slice(0, 2000)}`;
 
 			for await (const message of agentQuery) {
 				if (isSDKAssistantMessage(message)) {
-					const textBlocks = message.message.content.filter(
-						(b: { type: string }) => b.type === 'text'
-					) as Array<{ text?: string }>;
+					const content = message.message.content as Array<{
+						type: string;
+						text?: string;
+						thinking?: string;
+					}>;
+
+					// First, try to extract from text blocks
+					const textBlocks = content.filter((b) => b.type === 'text') as Array<{
+						type: 'text';
+						text: string;
+					}>;
 					title = textBlocks
-						.map((b) => b.text ?? '')
+						.map((b) => b.text)
 						.join(' ')
 						.trim();
+
+					// If no text blocks, try thinking blocks as fallback
+					if (!title) {
+						const thinkingBlocks = content.filter(
+							(b): b is { type: 'thinking'; thinking: string } =>
+								b.type === 'thinking' && 'thinking' in b
+						);
+						title = thinkingBlocks
+							.map((b) => b.thinking)
+							.join(' ')
+							.trim();
+					}
 
 					if (title) {
 						break; // Got the title, exit early
--- a/packages/daemon/src/lib/session/session-lifecycle.ts
+++ b/packages/daemon/src/lib/session/session-lifecycle.ts
@@ -832,6 +832,10 @@ ${messageText.slice(0, 2000)}`;
 					pathToClaudeCodeExecutable: cliPath,
 					executable: isRunningUnderBun() ? 'bun' : undefined,
 					env: mergedEnv,
+					// Disable thinking for title generation — we only need a short text response.
+					// Without this, models with adaptive thinking (e.g. Opus 4.6) may return
+					// only thinking blocks with no text block, causing an empty-response error.
+					thinking: { type: 'disabled' },
 				},
 			});
 
@@ -841,33 +845,13 @@ ${messageText.slice(0, 2000)}`;
 
 			for await (const message of agentQuery) {
 				if (isSDKAssistantMessage(message)) {
-					const content = message.message.content as Array<{
-						type: string;
-						text?: string;
-						thinking?: string;
-					}>;
-
-					// First, try to extract from text blocks
-					const textBlocks = content.filter((b) => b.type === 'text') as Array<{
-						type: 'text';
-						text: string;
-					}>;
+					const textBlocks = message.message.content.filter(
+						(b: { type: string }) => b.type === 'text'
+					) as Array<{ text?: string }>;
 					title = textBlocks
-						.map((b) => b.text)
+						.map((b) => b.text ?? '')
 						.join(' ')
 						.trim();
-
-					// If no text blocks, try thinking blocks as fallback
-					if (!title) {
-						const thinkingBlocks = content.filter(
-							(b): b is { type: 'thinking'; thinking: string } =>
-								b.type === 'thinking' && 'thinking' in b
-						);
-						title = thinkingBlocks
-							.map((b) => b.thinking)
-							.join(' ')
-							.trim();
-					}
 
 					if (title) {
 						break; // Got the title, exit early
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE with minor comments**

---

### Fix Correctness — P0 (clear)

The core fix is correct. `thinking: { type: 'disabled' }` is a properly typed SDK option (`ThinkingDisabled = { type: 'disabled' }` in `sdk.d.ts`). Tracing through `sdk.mjs`, the SDK maps this to `--thinking disabled` on the spawned `cli.js` subprocess, which prevents adaptive-thinking models (Opus 4.6+) from entering their extended thinking mode. This directly eliminates the root cause of empty-response errors.

The original PR #461 approach (extracting thinking block content as a title) was semantically wrong — thinking blocks contain internal chain-of-thought reasoning that can be thousands of tokens long and are not appropriate as session titles. Disabling thinking entirely is the right fix.

---

### Test Issues

**P2 — Unused import (`afterEach`)**

`packages/daemon/tests/unit/session/session-lifecycle-sdk-title.test.ts` line 12:
```ts
import { describe, expect, it, beforeEach, afterEach, mock } from 'bun:test';
```
`afterEach` is imported but never used. The linter should catch this, but it's worth cleaning up to avoid confusion about whether cleanup was intended.

**P2 — In-test `mock.module` overrides are not restored**

Tests 3 (`should strip markdown formatting`) and 4 (`should fall back to message text when SDK returns no text blocks`) call `mock.module('@anthropic-ai/claude-agent-sdk', ...)` inside their bodies without restoring the original mock afterward. In Bun, `mock.module` called inside a test body permanently updates the module registry for the lifetime of that test file. The tests currently pass because tests 1 and 2 run before the overrides take effect, but:
- Any new test added after test 3 or 4 will silently use the overriding mock unless it also calls `mock.module` again.
- The correct pattern (used elsewhere in this codebase) is to call `mock.module` only at the top level (hoisted), or to restore the original mock in `afterEach`.

**P2 — Missing test for the actual bug scenario**

None of the 4 tests simulate the exact bug: an assistant message whose `content` array contains **only** thinking blocks (i.e., `[{ type: 'thinking', thinking: '...' }]`) with no text block. The 4th test simulates a `result`-only response (no assistant message at all), which is a different code path. A test like this would be the most direct regression guard:
```ts
it('should fall back when assistant message contains only thinking blocks', async () => {
  mock.module('@anthropic-ai/claude-agent-sdk', () => ({
    query: () =>
      makeAsyncGen([
        {
          type: 'assistant',
          message: {
            content: [{ type: 'thinking', thinking: 'Long internal reasoning...' }],
          },
        },
      ]),
  }));
  // ...
  expect(result.isFallback).toBe(true);
});
```
Since `thinking: { type: 'disabled' }` is now set, this scenario *should* not happen in practice, but testing it confirms the defensive `if (!title) throw` path also works correctly.

**P3 — Lifecycle re-creation is fragile**

Tests 3 and 4 re-instantiate `SessionLifecycle` after overriding `mock.module`. This works because `generateTitleWithSdk` uses `await import(...)` (dynamic import that re-resolves from the module registry). This is technically correct but non-obvious — a comment explaining why re-instantiation is needed (or is not needed) would help future readers.

---

### Summary

The implementation fix is sound and addresses the root cause. The 4 tests cover the happy path, markdown stripping, and the empty-response fallback. The issues are all in test quality (unused import, unrestored mocks, missing regression test for the exact bug scenario) rather than in production code correctness.

lsm (commented):

**Recommendation: APPROVE**

All three P2 issues from the previous review have been properly addressed in commit `87f08a4`:

### P2 fixes verified

1. **Unused `afterEach` import** -- Removed from the import statement (line 12). Clean.

2. **In-test `mock.module` overrides not restored** -- Replaced with a mutable `mockSdkMessages` variable at module scope. The single top-level `mock.module` call now reads from this variable, and `beforeEach` resets it to the default assistant message. Tests that need different responses simply assign to `mockSdkMessages` before calling the method under test. This eliminates the permanent module registry override problem and removes the need to re-instantiate `SessionLifecycle` in individual tests.

3. **Missing regression test for thinking-only responses** -- Added a dedicated test ("should fall back to message text when assistant message contains only thinking blocks") that simulates an assistant message with `content: [{ type: 'thinking', ... }]` and no text block. This directly exercises the defensive fallback path for the exact bug scenario. The test includes a clear comment explaining the relationship to the original bug and why this path should not occur in production with the fix in place.

### Additional improvements

- The previous "no text blocks" test has been properly separated into two distinct tests: one for thinking-only content and one for no-assistant-messages, each covering a different code path.
- Removed the unused `sessionOverrides` parameter from `makeSessionCache`, reducing dead code.
- The P3 concern about lifecycle re-creation fragility is also resolved as a consequence of the mock pattern change.

### Production code

The production fix (`thinking: { type: 'disabled' }` in `session-lifecycle.ts`) remains unchanged and correct from the previous review.

No remaining issues found.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix correctly addresses the root cause: models with adaptive thinking (e.g., Opus 4.6) could return only thinking blocks with no text content during title generation, causing an empty-response error. The solution — adding `thinking: { type: 'disabled' }` to the SDK query options — is the right approach since title generation only needs a short text response.

### Implementation Review

**Production code** (`packages/daemon/src/lib/session/session-lifecycle.ts`):
- The one-line change adding `thinking: { type: 'disabled' }` to the query options is minimal and targeted. Good.
- The existing fallback path (throwing `'No text content in SDK response'` which is caught by `generateTitleFromMessage`) remains as a defensive measure. Good.

**Test file** (`packages/daemon/tests/unit/session/session-lifecycle-sdk-title.test.ts`):

Mock approach is sound:
- Only external packages (`@anthropic-ai/claude-agent-sdk`, `@neokai/shared/sdk/type-guards`) are mocked at module level. Internal modules use real implementations, which was the fix for the prior mock pollution issue. This is the correct approach.
- The `isSDKAssistantMessage` mock faithfully matches the real implementation (`msg.type === 'assistant'`).
- The SDK `query` mock returns a proper async-iterable with `supportedModels`/`interrupt` methods, matching the Query interface contract.

Test isolation is clean:
- `lastTitleQueryOptions` and `mockSdkMessages` are reset in `beforeEach`.
- `process.env.ANTHROPIC_API_KEY` is set to `'test-api-key'` in `beforeEach` and restored to `''` in `afterEach`, matching the unit test `setup.ts` convention.
- No internal `mock.module()` calls that could leak across test files.

Test coverage is appropriate:
1. Verifies `thinking: { type: 'disabled' }` is passed to SDK query — directly tests the fix.
2. Verifies title extraction from text blocks — happy path.
3. Verifies markdown stripping — existing cleanup logic.
4. Verifies fallback when only thinking blocks are returned — regression test for the original bug.
5. Verifies fallback when no assistant messages exist — edge case.

### CI Status

All CI checks pass. The `providers-anthropic-copilot` job is pending (infrastructure issue confirmed identical on unrelated PR #463).

### Minor Observations (P3)

1. The `makeQueryMock` always returns `supportedModels: () => Promise.resolve([])` which means the model validation path in `getValidatedModelId` will fall through to the static fallback. This is fine for testing since the test sets `defaultModel` in config, but worth noting for future test authors.

No blocking issues found. Clean fix with well-structured, isolated tests.

PR #462, opened 2026-03-20
task/task 61 custom agent list and editor

lsm:
- **feat: add SpaceAgentList and SpaceAgentEditor components (Task 6.1)**

### DIFF
--- a/packages/web/src/components/space/SpaceAgentEditor.tsx
+++ b/packages/web/src/components/space/SpaceAgentEditor.tsx
@@ -65,6 +65,21 @@ const SYSTEM_PROMPT_TEMPLATES: Record<string, string> = {
 };
 
 // ============================================================================
+// Pure helpers (module-level to avoid re-creation on each render)
+// ============================================================================
+
+/** Detect which preset name matches a given tool list, or 'Custom' if no match */
+function detectPreset(toolList: string[] | undefined): string {
+	if (!toolList) return 'Full Coding';
+	for (const [preset, presetTools] of Object.entries(TOOL_PRESETS)) {
+		if (toolList.length === presetTools.length && presetTools.every((t) => toolList.includes(t))) {
+			return preset;
+		}
+	}
+	return 'Custom';
+}
+
+// ============================================================================
 // Sub-components
 // ============================================================================
 
@@ -145,7 +160,6 @@ function LineNumberedTextarea({
 // ============================================================================
 
 export interface SpaceAgentEditorProps {
-	spaceId: string;
 	/** Existing agent to edit. Null = create mode. */
 	agent: SpaceAgent | null;
 	/** Names of other agents in this space (for uniqueness validation) */
@@ -157,7 +171,6 @@ export interface SpaceAgentEditorProps {
 }
 
 export function SpaceAgentEditor({
-	spaceId: _spaceId,
 	agent,
 	existingAgentNames,
 	onSave,
@@ -179,20 +192,6 @@ export function SpaceAgentEditor({
 	const [saving, setSaving] = useState(false);
 	const [errors, setErrors] = useState<Record<string, string>>({});
 	const [saveError, setSaveError] = useState<string | null>(null);
-
-	/** Detect which preset matches a tool list (or 'Custom' if none match) */
-	function detectPreset(toolList: string[] | undefined): string {
-		if (!toolList) return 'Full Coding';
-		for (const [preset, presetTools] of Object.entries(TOOL_PRESETS)) {
-			if (
-				toolList.length === presetTools.length &&
-				presetTools.every((t) => toolList.includes(t))
-			) {
-				return preset;
-			}
-		}
-		return 'Custom';
-	}
 
 	const applyPreset = (presetName: string) => {
 		setActivePreset(presetName);
--- a/packages/web/src/components/space/SpaceAgentEditor.tsx
+++ b/packages/web/src/components/space/SpaceAgentEditor.tsx
@@ -1,0 +1,496 @@
+/**
+ * SpaceAgentEditor Component
+ *
+ * Modal form for creating or editing a custom agent in a Space.
+ *
+ * Fields:
+ * - Name (required, unique within space)
+ * - Description (optional)
+ * - Role (radio: worker / reviewer / orchestrator, with tooltips)
+ * - Model (text input — model ID override)
+ * - Provider (text input — auto-detect or manual)
+ * - Tools (multi-select checkboxes from KNOWN_TOOLS)
+ * - System Prompt (monospace textarea with line numbers)
+ *
+ * Tool presets: "Full Coding" · "Read Only" · "Custom"
+ * System prompt templates: "Coder" · "Reviewer" · "Research" · "Custom (blank)"
+ *
+ * Validation: name required + unique, model required, at least one tool selected.
+ */
+
+import { useState } from 'preact/hooks';
+import { Modal } from '../ui/Modal';
+import { Button } from '../ui/Button';
+import { spaceStore } from '../../lib/space-store';
+import { KNOWN_TOOLS } from '@neokai/shared';
+import type { SpaceAgent } from '@neokai/shared';
+
+// ============================================================================
+// Constants
+// ============================================================================
+
+type ToolName = (typeof KNOWN_TOOLS)[number];
+
+/** Available roles with display labels and tooltip descriptions */
+const AGENT_ROLES = [
+	{
+		value: 'worker',
+		label: 'Worker',
+		tooltip: 'Executes coding tasks — writes code, runs tests, opens PRs.',
+	},
+	{
+		value: 'reviewer',
+		label: 'Reviewer',
+		tooltip: 'Reviews code changes, checks quality, and provides feedback.',
+	},
+	{
+		value: 'orchestrator',
+		label: 'Orchestrator',
+		tooltip: 'Plans and coordinates tasks across multiple agents in a workflow.',
+	},
+] as const;
+
+/** Tool presets map preset name → tool selection */
+const TOOL_PRESETS: Record<string, ToolName[]> = {
+	'Full Coding': ['Read', 'Write', 'Edit', 'Bash', 'Grep', 'Glob', 'WebFetch', 'WebSearch'],
+	'Read Only': ['Read', 'Grep', 'Glob'],
+};
+
+/** System prompt templates */
+const SYSTEM_PROMPT_TEMPLATES: Record<string, string> = {
+	Coder: `You are a skilled software engineer. Your role is to implement features, fix bugs, and write clean, maintainable code. Follow the project's conventions and best practices. Always write tests for new functionality and ensure all existing tests continue to pass.`,
+	Reviewer: `You are a thorough code reviewer. Your role is to analyze code changes for correctness, security, performance, and style. Provide constructive, actionable feedback with specific suggestions for improvement. Be concise and prioritize blocking issues.`,
+	Research: `You are a research assistant. Your role is to investigate technical questions, gather information from available sources, and produce clear, accurate summaries. Cite evidence when possible and flag uncertainty explicitly.`,
+	'Custom (blank)': '',
+};
+
+// ============================================================================
+// Sub-components
+// ============================================================================
+
+interface TooltipProps {
+	text: string;
+}
+
+function InfoTooltip({ text }: TooltipProps) {
+	return (
+		<span class="relative group inline-flex items-center">
+			<svg
+				class="w-3.5 h-3.5 text-gray-600 hover:text-gray-400 cursor-help transition-colors"
+				fill="none"
+				viewBox="0 0 24 24"
+				stroke="currentColor"
+			>
+				<path
+					stroke-linecap="round"
+					stroke-linejoin="round"
+					stroke-width={2}
+					d="M13 16h-1v-4h-1m1-4h.01M21 12a9 9 0 11-18 0 9 9 0 0118 0z"
+				/>
+			</svg>
+			<span class="absolute left-5 bottom-0 z-50 w-52 p-2 text-xs text-gray-200 bg-dark-700 border border-dark-600 rounded shadow-lg opacity-0 group-hover:opacity-100 transition-opacity pointer-events-none">
+				{text}
+			</span>
+		</span>
+	);
+}
+
+interface LineNumberedTextareaProps {
+	value: string;
+	onChange: (value: string) => void;
+	placeholder?: string;
+	rows?: number;
+}
+
+/** Monospace textarea with a line-number gutter on the left */
+function LineNumberedTextarea({
+	value,
+	onChange,
+	placeholder,
+	rows = 10,
+}: LineNumberedTextareaProps) {
+	const lineCount = value ? value.split('\n').length : 1;
+	const displayLines = Math.max(lineCount, rows);
+
+	return (
+		<div class="relative flex border border-dark-600 rounded-lg overflow-hidden bg-dark-800 focus-within:border-blue-500 transition-colors">
+			{/* Line numbers gutter */}
+			<div
+				aria-hidden="true"
+				class="flex flex-col items-end px-2 py-2 select-none text-gray-600 text-xs font-mono bg-dark-850 border-r border-dark-700 flex-shrink-0"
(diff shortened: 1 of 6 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured implementation of the SpaceAgentList and SpaceAgentEditor components. The code is clean, follows existing codebase patterns, and the test coverage is thorough. A few observations below.

### Positive Highlights

- **Consistent patterns**: Signal access via `spaceStore.agents.value` matches the established convention used by `SpaceDashboard`, `SpaceNavPanel`, and `SpaceTaskPane`.
- **Good use of shared constants**: Tools are sourced from `KNOWN_TOOLS` in `@neokai/shared`, ensuring a single source of truth.
- **Solid validation**: Name uniqueness (case-insensitive), required model, and at least one tool selected are all enforced client-side.
- **Workflow reference warning on delete**: The `getWorkflowNamesReferencingAgent` function correctly checks `wf.steps` for agent references and surfaces a meaningful warning before deletion.
- **Test quality**: 54 tests covering create/edit modes, validation errors, tool presets, system prompt templates, delete flow with error handling, and workflow reference warnings. Mocking strategy is consistent with the rest of the test suite.

### P3 (Nits)

1. **Unused `spaceId` prop in `SpaceAgentEditor`** (`SpaceAgentEditor.tsx:160`): The `spaceId` prop is received but immediately renamed to `_spaceId` and never used because `spaceStore.createAgent` internally reads `spaceId` from the store. This is technically correct but may confuse future readers. Consider either removing the prop or adding a brief comment explaining why it is forwarded but unused. Low priority since the store handles it.

2. **`makeWorkflow` test helper is incomplete** (`SpaceAgentList.test.tsx:158-169`): The `SpaceWorkflow` type has required fields `startStepId` and `tags` that the test helper omits. This compiles only because of `@ts-nocheck`. While this is the standard pattern for test files in this codebase, it could mask type drift if the `SpaceWorkflow` interface evolves.

3. **`detectPreset` is defined inside the component body** (`SpaceAgentEditor.tsx:184-195`): This pure function doesn't depend on component state and could be hoisted outside the component to avoid re-creation on each render. Minor perf concern.

### No Issues Found

- No `console.log` violations
- No security concerns (no user-controlled HTML injection, no unescaped output)
- Error handling is consistent (try/catch with user-facing error messages)
- `ConfirmModal` props (`confirmButtonVariant: 'warning'`) correctly match the component's interface
- Components are not yet integrated into a parent, which is expected for a Task 6.1 scope

Overall this is a clean, well-tested implementation that achieves the task requirements.

lsm (commented):

> **Recommendation:** REQUEST_CHANGES — 1 P0, 2 P1, 2 P2 issues found.

---

## Summary

The `SpaceAgentList` and `SpaceAgentEditor` components are well-structured, match the project's Preact/Signals patterns, use `KNOWN_TOOLS` correctly, handle CRUD flows with proper error surfaces, and all 54 tests pass. However, there is a critical integration gap: neither component is rendered anywhere in the live UI. Additionally, several meaningful issues need addressing before this is production-ready.

---

## P0 — Components Not Integrated Into the Space UI

**Neither `SpaceAgentList` nor `SpaceAgentEditor` is mounted anywhere in the running application.**

- `SpaceAgentList` is only referenced by its own file and its test file.
- `SpaceAgentEditor` is only referenced by `SpaceAgentList` and its tests.
- `SpaceIsland.tsx` (`packages/web/src/islands/SpaceIsland.tsx`) does not import or render `SpaceAgentList`.
- `SpaceNavPanel.tsx` has an "Agents" nav link stub but it is wired to `TODO` / no-op — clicking it does nothing.

The acceptance criterion says "Full agent CRUD flow works in Space UI." Currently the components exist in isolation and are unreachable to users. There is no route, no panel slot, and no navigation connection for the agent list view.

---

## P1 — `spaceId` Prop Is Silently Unused in `SpaceAgentEditor`

**File:** `packages/web/src/components/space/SpaceAgentEditor.tsx`, line 160

```ts
spaceId: _spaceId,
```

The `spaceId` prop is destructured with an underscore prefix to suppress the unused-variable warning but is never used. `spaceStore.createAgent` and `spaceStore.updateAgent` pick up the active space ID from `spaceStore.spaceId` internally — so the prop is purely vestigial.

This matters because it means the editor will silently operate on whichever space is selected in the store rather than the space the caller explicitly asked for. If a parent ever renders this component for a non-active space (e.g. in a settings overlay loaded before full navigation), it will silently corrupt the wrong space's agents. Either the prop should drive the RPC call, or it should be removed from the interface.

---

## P1 — Delete Flow Has No UI Blocker for Workflow-Referenced Agents

**File:** `packages/web/src/components/space/SpaceAgentList.tsx`, lines 184–188

When an agent is referenced by workflows, the modal warns the user but still allows deletion. The `confirmButtonVariant` changes to `'warning'` (amber) instead of blocking. The actual deletion, if confirmed, proceeds unconditionally from the client side. Whether the backend enforces referential integrity depends on the daemon's `spaceAgent.delete` handler.

If the daemon does NOT block deletion of an agent that is still referenced by workflow steps, this results in broken workflows with no further indication. The UX implies safety that may not exist. Verify whether the daemon enforces this constraint; if not, consider blocking deletion from the UI when workflows are referenced. If the daemon does enforce it, the error from the server surfaces correctly via `deleteError`, but the test for this path only exercises a generic error string rather than the actual workflow-reference constraint path.

---

## P2 — Both Test Files Use `// @ts-nocheck` at the Top

**Files:** `packages/web/src/components/space/__tests__/SpaceAgentList.test.tsx:1` and `SpaceAgentEditor.test.tsx:1`

The `@ts-nocheck` directive disables all TypeScript checking for the entire file. Existing sibling test files also use this pattern, so it is not novel, but it means type regressions in the tested components (changed prop shapes, renamed exports) will not be caught by tests at compile time. Given the project bans `no-explicit-any` in application code, tests should not be a blanket escape hatch from the type system. The mock definitions and factory helpers are typed well enough that removing `@ts-nocheck` should require only minor effort.

---

## P2 — Dead `checkedTools` Variable in Preset Test

**File:** `packages/web/src/components/space/__tests__/SpaceAgentEditor.test.tsx`, lines 288–291

```ts
const checkedTools = Array.from(toolCheckboxes)
    .filter((cb) => (cb as HTMLInputElement).checked)
    .map((cb) => (cb as HTMLInputElement).closest('label')?.textContent?.trim());
```

This `checkedTools` variable is assigned and then never read. The test re-queries the DOM independently below it. This looks like a leftover from a refactor. It causes no test failures because TypeScript checking is suppressed, but it is misleading noise.

---

## Minor Observations (P3)

- **`detectPreset` defined inside render scope**: The function in `SpaceAgentEditor.tsx` (line 184) is defined inside the component body and called from a `useState` initializer and a `setTools` updater. Moving it outside the component would be cleaner, though this is not a correctness issue since `useState` only calls the initializer once.

- **Line-number gutter misalignment**: `LineNumberedTextarea` uses a fixed `line-height: 1.375rem` for both the gutter and the `textarea`. On non-default font sizes or with browser sub-pixel rounding, the gutter numbers will drift out of alignment with textarea lines.

- **Tooltip overflow on narrow viewports**: `InfoTooltip` uses `left-5 bottom-0` with a fixed `w-52`. Near the right edge of the viewport the tooltip will overflow off-screen with no clamping guard.

- **Model field accepts arbitrary text**: The `Model` input accepts any string with no validation that the ID is valid for the selected provider. Acceptable for now, but worth noting for future improvement.

lsm (commented):

Re-review of Task 6.1: Custom Agent List and Editor.

### Previous issues — all resolved

1. **P0 (components not wired in)** — Fixed. `SpaceIsland` now imports `SpaceAgentList`, manages `activeView` state (`'dashboard' | 'agents'`), and renders the agent list when `activeView === 'agents'`. The `SpaceNavPanel` receives `onAgentsClick` and highlights the "Agents" nav link via `activeView`.

2. **P1 (unused `spaceId` prop on SpaceAgentEditor)** — Fixed. The `SpaceAgentEditorProps` interface no longer includes `spaceId`. The editor delegates to `spaceStore.createAgent`/`updateAgent` which already know the active space.

3. **P1 (delete should block, not warn, for workflow-referenced agents)** — Fixed. `SpaceAgentList` now renders a `Modal` with title "Cannot Delete Agent" and an "Understood" dismiss button (no delete action) when the agent is referenced by workflows. A standard `ConfirmModal` with the delete action only appears for unreferenced agents. Tests at lines 417-470 of `SpaceAgentList.test.tsx` verify both paths.

4. **P2 (dead `checkedTools` variable)** — Fixed. The variable at line 313 of `SpaceAgentEditor.test.tsx` is now used in assertions on lines 320-325.

5. **P3 (`detectPreset` at module level)** — Fixed. Moved to line 72 of `SpaceAgentEditor.tsx`, outside the component function.

### New observations

**P3 — `NavLink` uses `<div>` instead of `<button>` or `<a>` (SpaceNavPanel.tsx, line 213-223)**
The `NavLink` component renders a clickable `<div>` with `cursor-pointer`. For accessibility, interactive elements should be `<button>` elements so they are focusable and keyboard-operable. This is a minor accessibility nit — the nav links work fine for mouse users.

**P3 — `spaceId` prop accepted but unused in `SpaceAgentList` (line 132) and `SpaceNavPanel` (line 102)**
Both components accept `spaceId` as a prop but immediately destructure it as `_spaceId` (unused). This is likely intentional as a forward-compatibility measure for when per-space filtering is needed, but worth noting.

### Summary

All five issues from the prior review have been addressed. The implementation is clean, well-tested (comprehensive test coverage for both components), and correctly integrated into the SpaceIsland view. The delete-blocking flow for workflow-referenced agents is solid. Only minor P3 nits remain.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** — All prior review blockers have been addressed. The implementation is solid, tests pass (58/58), typecheck and lint are clean. A few minor issues remain worth noting.

---

### What was fixed (verified)

All five items from the prior review are confirmed resolved:

- **SpaceAgentList is wired into SpaceIsland** via `activeView` state; the "Agents" nav link correctly calls `setActiveView('agents')` and renders `SpaceAgentList` in the middle column.
- **`spaceId` prop removed from `SpaceAgentEditorProps`** — the editor reads the space through the store singleton; prop is gone.
- **Delete is now blocked (not warned)** when an agent is referenced by workflow steps — a "Cannot Delete Agent" informational modal is shown with no delete button, only an "Understood" close button.
- **Dead `checkedTools` variable removed** from the editor test.
- **`detectPreset` moved to module level** outside the component body.

---

### Remaining issues

#### P2 — `SpaceAgentListProps.spaceId` is accepted but immediately discarded

**File:** `packages/web/src/components/space/SpaceAgentList.tsx`, line 132

```ts
export function SpaceAgentList({ spaceId: _spaceId }: SpaceAgentListProps) {
```

The prop is renamed to `_spaceId` and never used — the component reads everything from the global `spaceStore` singleton. The interface still declares the prop as required, so callers must pass it. This is misleading: a future caller might expect passing a different `spaceId` would scope the list to a different space, but it has no effect. The interface and the prop should be removed, or the component should actually use it to validate that the store's current space matches (a defensive guard).

#### P2 — `handleTaskSelect` in `SpaceIsland` doesn't reset `activeView`

**File:** `packages/web/src/islands/SpaceIsland.tsx`, line 54-57

```ts
const handleTaskSelect = (taskId: string) => {
    setActiveRunId(null);
    navigateToSpaceTask(spaceId, taskId);
    // missing: setActiveView('dashboard')
};
```

Compare to `handleRunSelect` which correctly calls `setActiveView('dashboard')` (line 49). If a user is in the 'agents' view and clicks a task in the left nav panel, the center column will remain on `SpaceAgentList` while the right task pane opens. The expected behavior is that clicking a task always switches the center column to the dashboard view. This is a minor UX inconsistency.

#### P3 — Test fixture `makeWorkflow` is missing required fields from `SpaceWorkflow`

**File:** `packages/web/src/components/space/__tests__/SpaceAgentList.test.tsx`, line 182-193

The `SpaceWorkflow` interface requires `startStepId: string` and `tags: string[]`, but the `makeWorkflow` helper omits both. The `// @ts-nocheck` at the top of the file suppresses the type error, and tests pass at runtime because the component only accesses `wf.steps` and `wf.name`. However, the fixture should match the real contract to prevent future confusion if the component starts using `startStepId` or `tags`.

#### P3 — `wf.steps?.some()` uses unnecessary optional chaining

**File:** `packages/web/src/components/space/SpaceAgentList.tsx`, line 145

```ts
.filter((wf) => wf.steps?.some((step) => step.agentId === agentId))
```

`SpaceWorkflow.steps` is typed as `WorkflowStep[]` (non-optional, required array), so the `?` optional chain is unnecessary. It works fine but signals a misunderstanding of the type or defensive coding against a type that doesn't match the schema.

lsm (commented):

### Re-review after iteration-2 fixes

All four P2/P3 issues from the prior round have been properly addressed:

1. **P2 (spaceId prop removed)**: `SpaceAgentList` no longer accepts or declares a `spaceId` prop. The `SpaceAgentListProps` interface is gone entirely. The component correctly reads space data from `spaceStore` signals. `SpaceAgentEditor` similarly had its unused `spaceId` prop removed.

2. **P2 (handleTaskSelect resets activeView)**: `handleTaskSelect` in `SpaceIsland.tsx` now calls `setActiveView('dashboard')`, matching the pattern already used by `handleRunSelect`.

3. **P3 (optional chaining on wf.steps)**: `wf.steps?.some(...)` changed to `wf.steps.some(...)` since `SpaceWorkflow.steps` is typed as `WorkflowStep[]` (required, not optional).

4. **P3 (makeWorkflow fixture)**: Test helper now includes `startStepId: 'step-1'` and `tags: []`, matching the full `SpaceWorkflow` contract.

Additional improvements from the first fix commit are also solid:
- Delete flow refactored from a single `ConfirmModal` with warning text to a proper two-path UX: blocking `Modal` when workflow-referenced, standard `ConfirmModal` when unreferenced
- `detectPreset` moved to module level to avoid re-creation per render
- `SpaceNavPanel` NavLinks now wired with `active` state and click handlers
- Test coverage expanded with 6 new blocking-modal test cases

No remaining issues found. The implementation is clean and complete for Task 6.1.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

The three P2/P3 items from the prior round are all correctly addressed. The code is in good shape.

---

### P2 fixes verified

**`SpaceAgentList` props removed (was P2).**
The `SpaceAgentListProps` interface and the `spaceId: _spaceId` parameter have been removed entirely. The component now takes no props and reads `spaceStore` directly. `SpaceIsland.tsx` now renders `<SpaceAgentList />` without forwarding the prop. Correct.

**`handleTaskSelect` now resets `activeView` (was P2).**
`SpaceIsland.tsx` line 56 now has `setActiveView('dashboard')` inside `handleTaskSelect`, symmetric with `handleRunSelect` at line 49. Confirmed.

### P3 fixes verified

**`wf.steps.some()` optional chaining removed (was P3).**
`SpaceAgentList.tsx` line 141 now calls `wf.steps.some(...)` directly. `SpaceWorkflow.steps` is typed as `WorkflowStep[]` (non-optional), so this is correct.

**`makeWorkflow` test fixture now has `startStepId` and `tags` (was P3).**
`SpaceAgentList.test.tsx` lines 189–191 add `startStepId: 'step-1'` and `tags: []`. This now matches the full `SpaceWorkflow` contract.

---

### Remaining observations (no new blocking issues)

**P3 — `NavLink` uses `<div onClick>` instead of `<button>` (pre-existing, newly wired in this PR).**
`SpaceNavPanel.tsx` lines 213–223 render nav items as clickable `<div>` elements rather than `<button>` elements. This was already present before this task but this PR is the first to wire up `onClick` on these divs. A keyboard user cannot Tab to these nav items or activate them with Enter/Space. This is a minor accessibility gap worth fixing in a follow-up.

**P3 — `DEFAULT_PROPS = {}` is now a no-op spread.**
`SpaceAgentList.test.tsx` line 197 keeps `const DEFAULT_PROPS = {}` and all tests spread it via `{...DEFAULT_PROPS}`. Since `SpaceAgentList` takes no props, this is harmless but now meaningless. It could be cleaned up (just call `<SpaceAgentList />` directly), though it causes no functional issue.

---

All P0/P1/P2 issues are resolved. The implementation is correct and the tests cover the critical paths.

PR #464, opened 2026-03-20
fix: fetch real Copilot model IDs dynamically via client.listModels()

lsm:
The anthropic-copilot provider previously returned a hardcoded static model
list. If any of those IDs (e.g. claude-opus-4.6, claude-sonnet-4.6) are not
valid Copilot SDK model IDs for the authenticated account, createSession()
hangs without emitting session.idle — causing all inference tests to time out
after 120 s.

Fix: call client.listModels() in getModels() to get the actual available
model IDs for the user's Copilot account. The static list is now used only as
display-name/context-window enrichment (when an ID matches) and as a last-
resort fallback when listModels() fails or returns empty.

Also route all three SDK model tiers (opus/sonnet/haiku) to the same resolved
model ID so every SDK-internal call (summarisation, compaction, etc.) routes
through the real Copilot model the user selected, rather than a hardcoded
fallback that might be unavailable.

Addresses: https://github.com/lsm/neokai/actions/runs/23327807694

### DIFF
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -153,6 +153,17 @@ const COPILOT_ANTHROPIC_MODELS: ModelInfo[] = [
 		available: true,
 	},
 	{
+		id: 'gpt-5.3-codex',
+		name: 'GPT-5.3 Codex (Copilot)',
+		alias: 'copilot-anthropic-codex',
+		family: 'gpt',
+		provider: 'anthropic-copilot',
+		contextWindow: 272000,
+		description: 'GPT-5.3 Codex via GitHub Copilot · Best for coding',
+		releaseDate: '2025-12-01',
+		available: true,
+	},
+	{
 		// Intentionally "gemini-3-pro-preview" (not "gemini-3.1-pro-preview").
 		// The legacy CLI path used "gemini-3.1-pro-preview" because it passed IDs
 		// directly to Copilot CLI. This provider passes IDs as hints to the embedded
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -679,9 +679,10 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 	 * for 5 minutes by `resolveGitHubToken`. Expected latency: 3–15 s (subprocess spawn
 	 * + OAuth exchange). A 20 s hard timeout prevents indefinite hangs on slow networks.
 	 *
-	 * Note: uses `gpt-4o-mini` as the validation model. If Copilot ever removes that
-	 * model, validation will spuriously return false (prompting the user to re-authenticate
-	 * via OAuth rather than silently succeeding).
+	 * Note: uses `gpt-5-mini` as the validation model — the designated free-tier
+	 * Copilot model used across CI. If Copilot ever removes that model, validation
+	 * will spuriously return false (prompting the user to re-authenticate via OAuth
+	 * rather than silently succeeding).
 	 */
 	private async validateCopilotToken(token: string): Promise<boolean> {
 		const TIMEOUT_MS = 20_000;
@@ -694,7 +695,7 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 			let timeoutHandle: ReturnType<typeof setTimeout> | undefined;
 			const session = await Promise.race([
 				client.createSession({
-					model: 'gpt-4o-mini',
+					model: 'gpt-5-mini',
 					onPermissionRequest: () => Promise.resolve({ kind: 'approved' as const }),
 				}),
 				new Promise<never>((_, reject) => {
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -153,17 +153,6 @@ const COPILOT_ANTHROPIC_MODELS: ModelInfo[] = [
 		available: true,
 	},
 	{
-		id: 'gpt-5.3-codex',
-		name: 'GPT-5.3 Codex (Copilot)',
-		alias: 'copilot-anthropic-codex',
-		family: 'gpt',
-		provider: 'anthropic-copilot',
-		contextWindow: 272000,
-		description: 'GPT-5.3 Codex via GitHub Copilot · Best for coding',
-		releaseDate: '2025-12-01',
-		available: true,
-	},
-	{
 		// Intentionally "gemini-3-pro-preview" (not "gemini-3.1-pro-preview").
 		// The legacy CLI path used "gemini-3.1-pro-preview" because it passed IDs
 		// directly to Copilot CLI. This provider passes IDs as hints to the embedded
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -54,7 +54,7 @@ import type {
 	ProviderOAuthFlowData,
 } from '@neokai/shared/provider';
 import type { ModelInfo } from '@neokai/shared';
-import { CopilotClient } from '@github/copilot-sdk';
+import { CopilotClient, type ModelInfo as CopilotSdkModelInfo } from '@github/copilot-sdk';
 import { startEmbeddedServer, type EmbeddedServer } from './server.js';
 import { execFile } from 'node:child_process';
 import { promisify } from 'node:util';
@@ -68,23 +68,35 @@ const execFileAsync = promisify(execFile);
 const logger = new Logger('anthropic-copilot-provider');
 
 /**
- * GitHub Copilot model definitions.
- * Aliases use the `copilot-anthropic-` prefix for backwards compatibility
- * with existing configurations.
+ * Infer the model family from a Copilot SDK model ID.
+ * Returns 'sonnet', 'opus', 'haiku', 'gpt', or 'gemini'.
+ */
+function inferModelFamily(modelId: string): string {
+	const id = modelId.toLowerCase();
+	if (id.includes('claude')) {
+		if (id.includes('opus')) return 'opus';
+		if (id.includes('haiku')) return 'haiku';
+		return 'sonnet';
+	}
+	if (id.includes('gemini')) return 'gemini';
+	return 'gpt';
+}
+
+/**
+ * Static fallback model definitions for the Copilot provider.
  *
- * These model IDs are the identifiers the GitHub Copilot backend recognises
- * and must be passed verbatim to `CopilotClient.createSession({ model })`.
+ * These are used as display-name / alias enrichment when building the model list
+ * from `client.listModels()`. They also serve as a last-resort fallback if the
+ * Copilot API is unreachable at the time `getModels()` is called.
+ *
+ * IMPORTANT — Do NOT rely on these IDs being valid for the user's Copilot account.
+ * The actual available models are fetched dynamically via `client.listModels()` so
+ * that the test model ID matches what the Copilot API actually accepts.
  *
  * NOTE — intentional model ID collision with the Anthropic provider:
- * The `id` fields below (e.g. `claude-opus-4.6`) are also claimed by the
- * Anthropic provider. `registry.detectProvider(modelId)` will therefore
- * return Anthropic (registered first in factory.ts) for these IDs. This is
- * by design: every `anthropic-copilot` session stores its provider ID
- * explicitly in `session.config.provider`, and `applyEnvVarsToProcess` /
- * `getEnvVarsForModel` accept an optional `providerId` argument that
- * bypasses auto-detection. All callers that need Copilot routing MUST pass
- * the explicit provider ID; callers that rely on `detectProvider` alone will
- * silently route to Anthropic.
+ * The `id` fields below (e.g. `claude-opus-4.6`) may also be claimed by the
+ * Anthropic provider. Every `anthropic-copilot` session stores its provider ID
+ * explicitly in `session.config.provider` to avoid ambiguity.
  */
 const COPILOT_ANTHROPIC_MODELS: ModelInfo[] = [
 	{
@@ -206,6 +218,12 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 	private serverStarting: Promise<EmbeddedServer> | undefined = undefined;
 	/** Resolved token cache with TTL */
 	private tokenCache: TokenCacheEntry | null = null;
+	/**
+	 * Dynamically fetched models from the Copilot API (via client.listModels()).
+	 * Populated in getModels() and used by ownsModel() so that real Copilot model
+	 * IDs (which may differ from our static list) are recognised by this provider.
+	 */
+	private dynamicModelsCache: ModelInfo[] | null = null;
 
 	/** Path to stored authentication tokens */
 	private readonly authPath: string;
@@ -250,11 +268,38 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 			logger.error('Failed to start embedded Anthropic server:', err);
 			return [];
 		}
+
+		// Fetch real model IDs from the Copilot API so we only expose models that
+		// the user's account can actually use.  Hardcoded model IDs are used as
+		// display-name / context-window enrichment when a match is found, and as a
+		// last-resort fallback when the API call fails.
+		if (this.clientCache) {
+			try {
+				const sdkModels = await this.clientCache.listModels();
+				const mapped = sdkModels
+					.filter((m) => m.policy?.state !== 'disabled')
+					.map((m) => this.mapCopilotSdkModel(m));
+				if (mapped.length > 0) {
+					this.dynamicModelsCache = mapped;
+					return mapped;
+				}
+			} catch (err) {
+				logger.warn('client.listModels() failed, falling back to static model list:', err);
+			}
+		}
+
 		return COPILOT_ANTHROPIC_MODELS;
 	}
 
 	ownsModel(modelId: string): boolean {
-		return COPILOT_ANTHROPIC_MODELS.some((m) => m.alias === modelId || m.id === modelId);
+		if (COPILOT_ANTHROPIC_MODELS.some((m) => m.alias === modelId || m.id === modelId)) {
+			return true;
+		}
+		// Also check dynamically fetched models (real Copilot SDK model IDs).
+		if (this.dynamicModelsCache) {
+			return this.dynamicModelsCache.some((m) => m.alias === modelId || m.id === modelId);
+		}
+		return false;
 	}
 
 	getModelForTier(tier: ModelTier): string | undefined {
@@ -286,7 +331,8 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 			);
 		}
 
-		const entry = COPILOT_ANTHROPIC_MODELS.find((m) => m.alias === modelId || m.id === modelId);
+		const allKnownModels = this.dynamicModelsCache ?? COPILOT_ANTHROPIC_MODELS;
+		const entry = allKnownModels.find((m) => m.alias === modelId || m.id === modelId);
 		const resolvedId = entry?.id ?? modelId;
 		// Per-session workspace path is encoded in the auth token so the embedded
 		// server (shared singleton) can apply the correct cwd per HTTP request.
@@ -302,10 +348,11 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 				ANTHROPIC_API_KEY: '',
 				CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC: '1',
 				API_TIMEOUT_MS: '300000',
-				ANTHROPIC_DEFAULT_OPUS_MODEL:
-					resolvedId === 'claude-opus-4.6' ? 'claude-opus-4.6' : 'claude-sonnet-4.6',
+				// All three tiers route to the same resolved model so the bridge handles
+				// every SDK-internal model request with the one real Copilot model ID.
+				ANTHROPIC_DEFAULT_OPUS_MODEL: resolvedId,
 				ANTHROPIC_DEFAULT_SONNET_MODEL: resolvedId,
-				ANTHROPIC_DEFAULT_HAIKU_MODEL: 'gpt-5-mini',
+				ANTHROPIC_DEFAULT_HAIKU_MODEL: resolvedId,
 			},
 			isAnthropicCompatible: true,
 			apiVersion: 'v1',
@@ -766,6 +813,30 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 	// ---------------------------------------------------------------------------
 	// Private helpers
 	// ---------------------------------------------------------------------------
+
+	/**
+	 * Map a Copilot SDK ModelInfo to NeoKai's ModelInfo format.
+	 *
+	 * Uses the static list for display-name and context-window enrichment when a
+	 * matching ID is found, so existing users with static aliases are not affected.
+	 */
+	private mapCopilotSdkModel(m: CopilotSdkModelInfo): ModelInfo {
+		// Look up static entry for enriched display metadata.
+		const staticEntry = COPILOT_ANTHROPIC_MODELS.find((s) => s.id === m.id);
+		const family = inferModelFamily(m.id);
+		return {
+			id: m.id,
+			name: staticEntry?.name ?? `${m.name ?? m.id} (Copilot)`,
+			alias: staticEntry?.alias ?? `copilot-${m.id.replace(/[^a-z0-9]/gi, '-').toLowerCase()}`,
+			family,
+			provider: 'anthropic-copilot',
+			contextWindow:
+				staticEntry?.contextWindow ?? m.capabilities?.limits?.max_context_window_tokens ?? 128000,
+			description: staticEntry?.description ?? `${m.name ?? m.id} via GitHub Copilot`,
+			releaseDate: staticEntry?.releaseDate ?? '2025-01-01',
+			available: m.policy?.state !== 'disabled',
+		};
+	}
 
 	private async createServer(): Promise<EmbeddedServer> {
 		const token = await this.resolveGitHubToken();
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -164,13 +164,8 @@ const COPILOT_ANTHROPIC_MODELS: ModelInfo[] = [
 		available: true,
 	},
 	{
-		// Intentionally "gemini-3-pro-preview" (not "gemini-3.1-pro-preview").
-		// The legacy CLI path used "gemini-3.1-pro-preview" because it passed IDs
-		// directly to Copilot CLI. This provider passes IDs as hints to the embedded
-		// HTTP server, which maps them to Copilot SDK sessions; the version suffix
-		// matters for routing.
-		id: 'gemini-3-pro-preview',
-		name: 'Gemini 3 Pro (Copilot)',
+		id: 'gemini-3.1-pro-preview',
+		name: 'Gemini 3.1 Pro (Copilot)',
 		alias: 'copilot-anthropic-gemini',
 		family: 'gemini',
 		provider: 'anthropic-copilot',
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -170,7 +170,7 @@ const COPILOT_ANTHROPIC_MODELS: ModelInfo[] = [
 		family: 'gemini',
 		provider: 'anthropic-copilot',
 		contextWindow: 128000,
-		description: 'Gemini 3 Pro Preview via GitHub Copilot',
+		description: 'Gemini 3.1 Pro Preview via GitHub Copilot',
 		releaseDate: '2025-11-15',
 		available: true,
 	},
@@ -679,10 +679,11 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 	 * for 5 minutes by `resolveGitHubToken`. Expected latency: 3–15 s (subprocess spawn
 	 * + OAuth exchange). A 20 s hard timeout prevents indefinite hangs on slow networks.
 	 *
-	 * Note: uses `gpt-5-mini` as the validation model — the designated free-tier
-	 * Copilot model used across CI. If Copilot ever removes that model, validation
-	 * will spuriously return false (prompting the user to re-authenticate via OAuth
-	 * rather than silently succeeding).
+	 * Uses `client.listModels()` for validation — a non-empty model list proves the
+	 * token has Copilot API access without depending on any specific model being
+	 * available on the user's plan.  This is intentionally model-agnostic: a user
+	 * on an enterprise plan with a different model catalogue should still be
+	 * authenticated correctly.
 	 */
 	private async validateCopilotToken(token: string): Promise<boolean> {
 		const TIMEOUT_MS = 20_000;
@@ -693,11 +694,8 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 		});
 		try {
 			let timeoutHandle: ReturnType<typeof setTimeout> | undefined;
-			const session = await Promise.race([
-				client.createSession({
-					model: 'gpt-5-mini',
-					onPermissionRequest: () => Promise.resolve({ kind: 'approved' as const }),
-				}),
+			const models = await Promise.race([
+				client.listModels(),
 				new Promise<never>((_, reject) => {
 					timeoutHandle = setTimeout(
 						() => reject(new Error('validateCopilotToken timed out')),
@@ -706,8 +704,7 @@ export class AnthropicToCopilotBridgeProvider implements Provider {
 				}),
 			]);
 			clearTimeout(timeoutHandle);
-			await session.disconnect();
-			return true;
+			return models.length > 0;
 		} catch {
 			return false;
 		} finally {
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -81,8 +81,12 @@ function pickModelForTier(models: ModelInfo[], tier: ModelTier): string | undefi
 
 	const keywordsByTier: Record<ModelTier, string[]> = {
 		opus: ['opus', 'pro', 'ultra'],
-		sonnet: ['sonnet', '4o', 'turbo', 'flash'],
+		// 'flash' is intentionally absent from sonnet — Gemini Flash models are fast/cheap
+		// and should be haiku-tier. Including 'flash' here would shadow the haiku path
+		// whenever a Flash model appears in the account's model list.
+		sonnet: ['sonnet', '4o', 'turbo'],
 		haiku: ['mini', 'haiku', 'flash', 'fast', 'lite'],
+		// 'default' mirrors 'sonnet': a mid-tier capable model is the right default.
 		default: ['sonnet', '4o', 'turbo'],
 	};
 	const keywords = keywordsByTier[tier] ?? [];
--- a/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-copilot/provider.ts
@@ -68,6 +68,33 @@ const execFileAsync = promisify(execFile);
 const logger = new Logger('anthropic-copilot-provider');
 
 /**
+ * Pick the most appropriate model from `models` for a given tier.
+ *
+ * Used by getModelForTier() when the static fallback ID is not in the dynamic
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

This PR correctly addresses the root cause of the CI timeout: hardcoded static model IDs that may not be recognized by the Copilot SDK, causing `createSession()` to hang indefinitely without emitting `session.idle`. The fix dynamically fetches available models via `client.listModels()` and uses those for all subsequent operations.

### What looks good

- **Root cause fix is sound.** Calling `client.listModels()` to discover real model IDs ensures the provider never passes an unrecognized ID to `createSession()`.
- **Graceful fallback.** When `listModels()` fails or returns empty, the static list is used as a fallback. This prevents breakage if the API is temporarily unavailable.
- **All three SDK tiers route to the same resolved model ID.** This eliminates the class of bugs where the SDK's internal haiku/opus calls route to model IDs the Copilot bridge cannot serve.
- **`ownsModel()` checks both static and dynamic lists.** Ensures models discovered dynamically are properly claimed by this provider.
- **`buildSdkConfig()` looks up the dynamic cache first.** When dynamic models are available, alias resolution uses the real model IDs.
- **Tests are comprehensive.** Covers the happy path (dynamic models returned), API error fallback, empty response fallback, and `ownsModel()` with dynamic cache. The context-manager invariant test is properly updated.
- **`mapCopilotSdkModel()` enriches dynamic models with static metadata.** Preserves display names, aliases, and context window values for known models while generating sensible defaults for unknown ones.

### Minor observations (P3)

1. **`getModelForTier()` still uses static model IDs** (line 305-313 of `provider.ts`). If the Copilot API returns different model IDs than the static list, `getModelForTier('haiku')` returns `'gpt-5-mini'` which may not exist in the dynamic cache. However, this is mitigated because `buildSdkConfig()` falls through to `entry?.id ?? modelId` when the model ID is not found, and all three tier env vars are set to the same resolved ID anyway. This is a pre-existing concern and not introduced by this PR.

2. **`dynamicModelsCache` is never invalidated.** Once populated, `getModels()` will re-fetch on each call (since `listModels()` is called every time `getModels()` runs), so the cache is effectively refreshed. However, if `listModels()` fails on a subsequent call after the cache was populated, the code falls back to the static list rather than the previously cached dynamic list. This is a reasonable trade-off -- the static list is a safe fallback -- but worth noting.

3. **`inferModelFamily()` returns a string rather than a union type.** The `family` field on `ModelInfo` is typed as `string`, so this is fine, but a stricter type could prevent typos.

All issues are P3 (nits/minor suggestions). The implementation is correct, well-tested, and properly handles edge cases.

lsm (commented):

The core fix is correct and addresses the root cause — replacing the hardcoded static model list with a live `client.listModels()` call. The fallback-to-static-list behaviour and disabled-model filtering are well-handled. However, there are three issues that need attention before this can be considered complete.

---

### P1 — `getModelForTier()` still returns static IDs that may not exist on the user's Copilot account

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts` lines 305-313

`getModelForTier()` was not updated and still returns hardcoded IDs such as `gpt-5-mini` (haiku) and `claude-opus-4.6` / `claude-sonnet-4.6` (opus/sonnet/default). This method is called by `ProviderService.getTitleGenerationConfig()` (`provider-service.ts` line 299) to select the cheap model for title generation, and the returned ID is passed directly into `buildSdkConfig(modelId)`.

The same bug that caused the CI timeout — passing an unavailable model ID to the Copilot SDK — can still be triggered through this path. If the user's account does not have `gpt-5-mini`, title generation will hang. The fix made `buildSdkConfig` smarter about resolving aliases via `dynamicModelsCache`, but when `getModelForTier` returns `gpt-5-mini` and that ID is not in `dynamicModelsCache`, `resolvedId` falls back to the literal string `gpt-5-mini` (line 336 of provider.ts), perpetuating the original bug on the title-generation code path.

```ts
// provider-service.ts:299
const modelId = provider.getModelForTier('haiku') || models[0]?.id || 'default';
// modelId = 'gpt-5-mini' ← still hardcoded, not validated against dynamicModelsCache
```

---

### P1 — `dynamicModelsCache` is never invalidated; stale cache persists for the daemon lifetime

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts` lines 276-291

Once `dynamicModelsCache` is populated on the first `getModels()` call, it is never refreshed. If the user's Copilot account model availability changes (e.g. a model is added or removed), the provider will continue serving the stale list until the daemon is restarted. Given that the token cache uses a 5-minute TTL (`TOKEN_CACHE_TTL_MS`), a parallel TTL for the models cache would be the natural fix. This is currently a latent bug in the feature rather than something that causes an immediate regression, but it is a meaningful omission for a daemon-long-running process.

---

### P2 — `ownsModel()` depends on `dynamicModelsCache` being pre-populated by a prior `getModels()` call

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts` lines 294-303

If a session is resumed before `getModels()` has been called (e.g. daemon restart with a session whose stored `provider` is `anthropic-copilot` but whose `modelId` is a real Copilot SDK ID that is not in the static list), `ownsModel(modelId)` will return `false` and the registry will fail to route the session to this provider. The implicit dependency on call order is undocumented and fragile. At minimum, the contract should be documented; ideally `ownsModel` should trigger a lazy fetch or the provider should be designed to handle this case.

---

### P3 — `getModelForTier()` haiku tier still maps to `gpt-5-mini` while `buildSdkConfig` changes contradict this

The test at line 99 (`provider.test.ts`) asserts `getModelForTier('haiku') === 'gpt-5-mini'`, but the purpose of the PR was to remove all reliance on static model IDs that may not exist on the user's account. The haiku tier returning `gpt-5-mini` is inconsistent with the new design intent stated in the PR description ("route all three SDK model tiers ... to the same resolved model ID"). The test documents the inconsistency rather than fixing it.

lsm (commented):

### Summary

The two P1 issues from the previous review have been properly addressed:

1. **getModelForTier() now uses the dynamic cache** (lines 354-377 in `provider.ts`): When `dynamicModelsCache` is populated, the method first checks if the static fallback ID exists in the cache. If not, it delegates to the new `pickModelForTier()` helper which uses keyword-based heuristics to find the best match for each tier. This ensures CI tests (and users) always get a model ID that the Copilot API actually recognizes.

2. **dynamicModelsCache now has a 5-minute TTL** (lines 266-267, 319-322, 332): A `dynamicModelsCacheExpiresAt` timestamp is checked before returning cached results, and set to `now + TOKEN_CACHE_TTL_MS` when the cache is populated. This matches the token cache TTL and prevents stale model lists from persisting indefinitely.

### Code Quality Assessment

**pickModelForTier() helper (lines 77-95):** Well-structured with reasonable keyword mappings per tier. The fallback to the first available model is a sensible default. One minor observation: the `sonnet` and `default` tiers share overlapping keywords (`'sonnet'`, `'4o'`, `'turbo'`), which is intentional since `default` should behave like `sonnet`.

**Test coverage (provider.test.ts):** The new tests cover:
- Static ID found in dynamic cache (returns the known static ID)
- Static ID NOT in cache, keyword match exists (returns keyword-matched model)
- No keyword match at all (falls back to first available model)
- TTL behavior: cached models returned without re-calling `listModels()`, and cache expiry triggers a fresh call

These cover the important scenarios well.

**ownsModel() documentation (lines 252-264):** The detailed JSDoc explaining the call-order dependency between `getModels()` and `ownsModel()` is helpful for future maintainers.

### No New Issues Found

The implementation is clean and consistent with the existing patterns. The `buildSdkConfig()` method (line 398) already uses `dynamicModelsCache ?? COPILOT_ANTHROPIC_MODELS` to resolve model IDs, so it also benefits from the dynamic cache.

**Recommendation: APPROVE** -- all previously identified P1 issues are resolved, no new issues introduced.

lsm (commented):

**Recommendation: APPROVE** — Both P1 issues from the previous review are correctly fixed. One new P2 and one P3 issue were found.

---

### Previously-reported P1 issues — verified fixed

**P1-1: getModelForTier() now uses the dynamic cache**

The new `pickModelForTier()` helper (lines 77-95) and the updated `getModelForTier()` body (lines 354-377) implement the three-step lookup correctly:
1. If the static fallback ID exists in the dynamic cache, return it.
2. Otherwise, pick the best keyword-matched model from the cache.
3. Last resort: return the static fallback ID.

This eliminates the root cause — hardcoded model IDs being passed to `buildSdkConfig()` when only real Copilot model IDs are in the dynamic cache.

**P1-2: dynamicModelsCache now has a 5-minute TTL**

`dynamicModelsCacheExpiresAt` (line 267) is initialized to 0, checked on every `getModels()` call (line 320), and set to `now + TOKEN_CACHE_TTL_MS` (5 minutes) on a successful `listModels()` result (line 332). This correctly mirrors the `tokenCache` pattern.

Test coverage for the TTL is solid: the test at line 525 verifies the cache is used on a second call, then expired manually and re-fetched on a third call.

---

### New issues found

**P2 — `flash` keyword appears in both `sonnet` and `haiku` tiers, causing non-deterministic routing**

`packages/daemon/src/lib/providers/anthropic-copilot/provider.ts` lines 83-86:

```ts
sonnet: ['sonnet', '4o', 'turbo', 'flash'],
haiku:  ['mini', 'haiku', 'flash', 'fast', 'lite'],
```

If the dynamic model cache contains a model with `flash` in its ID (e.g. `gemini-flash-2.0`), which tier wins depends on which keyword list iterates to `flash` first — `sonnet` wins because it is checked first in `getModelForTier()`. When the user intends a haiku-tier call, the sonnet-tier path picks the `flash` model instead. The result is unpredictable assignment and the test suite has no case covering a model list that includes a `flash`-named model requested from the `haiku` tier. Remove `flash` from the `sonnet` list (or the `haiku` list) to eliminate the ambiguity.

**P3 — static fallback used for the entire `default` tier even when cache has models**

`keywordsByTier` for `default` (line 86) overlaps entirely with `sonnet` keywords, so the two tiers effectively behave identically when the dynamic cache is populated. This is harmless functionally, but it would be clearer to either remove the separate `default` entry and alias it to `sonnet`, or add a comment explaining why they share the same keywords. No test exercises the `default` tier with a non-empty dynamic cache.

---

### Summary

The two P1 regressions from the original implementation — static model IDs used in `getModelForTier()` and no TTL on the dynamic model cache — are both fixed correctly. The `flash` keyword collision across `sonnet` and `haiku` tiers is a new issue that could cause incorrect model routing when a Gemini Flash model is in the user's account.

lsm (commented):

### Re-review after P2 fix (flash keyword deduplication)

The P2 fix is correct and well-documented. `flash` has been removed from the `sonnet` tier keywords in `pickModelForTier()` (line 87 of `provider.ts`) and remains only in the `haiku` tier (line 88). The inline comment explains the rationale clearly: Gemini Flash models are fast/cheap (haiku-tier), not mid-tier capable models, so including `flash` in the sonnet list would cause non-deterministic routing.

### Overall implementation assessment

The full PR addresses the original CI timeout issue comprehensively:

1. **Dynamic model fetching** via `client.listModels()` with TTL-based caching — correct and well-tested.
2. **Graceful fallback** to static model list when the API call fails or returns empty — tested.
3. **All three SDK tier slots** (`ANTHROPIC_DEFAULT_OPUS/SONNET/HAIKU_MODEL`) now route to the same resolved model ID, preventing the original hang where an unavailable model ID caused `createSession()` to never emit `session.idle`.
4. **`ownsModel()`** checks both static and dynamic caches, with the call-order dependency on `getModels()` clearly documented.
5. **`pickModelForTier()`** keyword routing is now deterministic — each keyword appears in exactly one tier, and the `default` tier intentionally mirrors `sonnet`.

### Minor observations (P3)

- The `default` tier in `keywordsByTier` duplicates `sonnet`'s keyword list. This is intentional per the comment, so no action needed — just noting it for clarity.

No remaining issues found. The fix is clean and the test coverage is thorough.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

### Re-review after P2 fix: `flash` removed from `sonnet` tier

The fix is correct and the implementation is sound.

---

### Fix verification

**`packages/daemon/src/lib/providers/anthropic-copilot/provider.ts` lines 82-91:**

The `flash` keyword is now absent from the `sonnet` tier and present only in the `haiku` tier. The inline comment clearly explains the rationale. Because `pickModelForTier()` receives the caller's `tier` parameter and indexes directly into `keywordsByTier[tier]`, there is no way for a Flash model to shadow the `haiku` path when `sonnet` is requested, and vice versa. The fix is deterministic and eliminates the ambiguity described in the previous review.

The `default` tier intentionally mirrors `sonnet` (same three keywords: `sonnet`, `4o`, `turbo`) and the comment added in this commit makes that intent explicit. This is correct — a mid-tier capable model is the right default.

No regressions introduced. The existing unit tests in `packages/daemon/tests/unit/providers/anthropic-copilot/provider.test.ts` still pass against the updated keyword table: the test that injects `gpt-4o` and `gpt-4o-mini` into the dynamic cache (lines 124-154) exercises the `4o` and `mini` keyword paths and is not affected by the `flash` removal from `sonnet`.

---

### Remaining minor observation (P3, not blocking)

There is still no unit test that covers the specific scenario described in the previous review: a dynamic cache containing a `flash`-named model (e.g. `gemini-flash-2.0`) where `getModelForTier('haiku')` should return it and `getModelForTier('sonnet')` should not. The fix is logically correct without such a test, but a targeted regression test would prevent this from silently breaking again if someone adds `flash` back to the `sonnet` tier in the future. This is a P3 nit — not a blocker.

---

### Summary

All P0, P1, and P2 issues from the full review cycle are resolved. The `flash` deduplication fix is correct and clearly documented. The PR is ready to merge.

lsm (commented):

**Recommendation: APPROVE**

This PR makes solid improvements to the Copilot provider: dynamic model discovery via `client.listModels()`, TTL-based caching, tier-based fallback logic, and hardened CI tests that enforce `gpt-5-mini` as the only test model.

### What works well

1. **Dynamic model fetching with graceful fallback** (`provider.ts:306-331`): The `getModels()` flow fetches real models from the Copilot API, caches them with a 5-minute TTL, and falls back to the static list on error or empty results. This directly addresses the original CI timeout bug.

2. **Hard-fail in online tests** (`anthropic-to-copilot-bridge-provider.test.ts:258-271`): The `beforeAll` now throws a descriptive error if `gpt-5-mini` is missing, consistent with the CLAUDE.md policy of never silently skipping credential-dependent tests.

3. **All SDK tiers route to the resolved model** (`provider.ts:408-412`): Setting all three `ANTHROPIC_DEFAULT_*_MODEL` env vars to the same `resolvedId` prevents the SDK's internal haiku/opus calls from using a model ID the Copilot bridge cannot serve.

4. **Thorough unit test coverage**: Dynamic cache behavior, TTL expiry, keyword-based tier matching, and fallback-to-first-available are all tested.

### Minor observations (P3)

- **`pickModelForTier` keyword matching** (`provider.ts:93-95`): The keyword search uses `id.toLowerCase().includes(kw)` which could match substrings unexpectedly (e.g., a model named `prometheus` would match the `pro` keyword for the opus tier). In practice, Copilot model IDs are well-structured enough that this is unlikely to cause issues, but word-boundary matching would be more precise.

- **Concurrent `getModels()` calls**: If two callers invoke `getModels()` simultaneously after cache expiry, both will call `listModels()`. This is harmless (both will write the same data to the cache) but a deduplication promise pattern (like `serverStarting`) could avoid the redundant API call. Not worth fixing given the 5-minute TTL makes this extremely rare.

- **`inferModelFamily` return type** (`provider.ts:105`): The function returns `string` but only produces a fixed set of values (`'sonnet'`, `'opus'`, `'haiku'`, `'gpt'`, `'gemini'`). A union type would be more precise, though it flows into `ModelInfo.family` which is typed as `string` anyway.

Overall this is a well-structured fix with good test coverage and clear documentation. No blocking issues found.

lsm (commented):

> **Recommendation:** APPROVE — this implementation is correct and complete for the stated goals.

---

### Summary

This PR fixes CI test failures in the anthropic-copilot provider by replacing the hardcoded static model list with a dynamic `client.listModels()` call, and enforcing `gpt-5-mini` as the sole CI test model with a hard-fail guard if it's absent.

---

### What Was Requested vs. What Was Delivered

The task asked for:
1. Online test `beforeAll` hard-fails (not falls back) if `gpt-5-mini` is absent. ✓ Delivered correctly.
2. `testModelId` is always `'gpt-5-mini'`, never dynamically selected from `copilotModels[0]`. ✓ Delivered correctly.
3. Removed `gpt-5.3-codex` / `copilot-anthropic-codex` from `COPILOT_ANTHROPIC_MODELS`. ✓ Delivered correctly.
4. Unit tests updated to remove `copilot-anthropic-codex` and `gpt-5.3-codex` assertions. ✓ Delivered correctly.

---

### Correctness

**Dynamic model fetching (`getModels()`):** The `ensureServerStarted()` → `createServer()` → `getOrCreateClient()` chain guarantees `this.clientCache` is non-null before the `if (this.clientCache)` branch is reached. The ordering is correct.

**`buildSdkConfig()` all-tiers-same-model change:** Previously opus/sonnet used their respective Claude model IDs and haiku used `gpt-5-mini`. The new code routes all three SDK tiers to the same resolved Copilot model ID. This eliminates the original hang: SDK internal calls (summarisation, compaction) that previously used `claude-haiku-4-5-20251001` (not served by the Copilot bridge) now route through the real session model.

**Cache TTL sharing:** Dynamic models cache uses the same 5-minute TTL as the token cache — a reasonable choice.

**Static list enrichment:** `mapCopilotSdkModel()` correctly falls back to static display metadata when the dynamic model ID matches a static entry ID, preserving alias compatibility.

**Removal of `gpt-5.3-codex` from Copilot's static list:** This model belongs to the `anthropic-codex` provider. Its presence in `COPILOT_ANTHROPIC_MODELS` was creating a false `ownsModel()` match. The removal is correct and the corresponding `anthropic-to-codex-bridge-provider.ts` still owns `gpt-5.3-codex` independently.

---

### Observations (Non-Blocking)

**P2: `validateCopilotToken()` uses `gpt-4o-mini`, but CI enforces `gpt-5-mini`.**
Lines 687–702 of `provider.ts` note that `gpt-4o-mini` is used for token validation and explicitly warns that if Copilot removes it, validation will return false. Meanwhile the test comment on line 259 says `gpt-5-mini` is the designated free-tier model. If `gpt-4o-mini` disappears before `gpt-5-mini` is substituted here, users relying on sources 4/5 (gh CLI / hosts.yml) will silently fail to authenticate. Consider updating `validateCopilotToken()` to use `gpt-5-mini` for consistency. This is a code quality concern, not a correctness bug given both models are expected to be available.

**P3: Online test 6 (error-envelope) creates a fresh `AnthropicToCopilotBridgeProvider()` with no args, spawning a second CopilotClient subprocess.** The test comment in test 7 explicitly explains this could cause 429 rate-limit issues after heavy test activity. The comment is good, but if the Copilot free-tier rate limit is tight, this second subprocess instance may intermittently fail test 6 before test 7 sees the 429. This is pre-existing design, not introduced in this iteration, and is already documented.

**P3: The `isAvailable()` check only rejects `ghp_` classic PATs from `resolveGitHubToken()`, but `resolveGitHubToken()` may still return a `ghp_` token from `COPILOT_GITHUB_TOKEN` env var that was never validated.** This is intentional (env-var tokens are trusted), but the code comment at source 2 could mention the ghp_ rejection to prevent confusion. Minor documentation nit.

---

### Test Coverage Quality

Unit tests are comprehensive and cover:
- Dynamic model fetching, caching, TTL expiry, fallback on empty/error.
- `ownsModel()` with dynamic cache entries.
- `getModelForTier()` with and without dynamic cache, keyword-based fallback, first-available fallback.
- `buildSdkConfig()` all-tiers-same-model invariant.
- Concurrent `ensureServerStarted()` deduplication and retry-after-failure.
- `logout()` cache invalidation, OAuth flow shape and deduplication.
- `slow_down` RFC 8628 §3.5 backoff verification.

The online test's `beforeAll` hard-fail (no skip guard) properly enforces the CLAUDE.md policy on credential-dependent online tests.

lsm (commented):

### Recommendation: APPROVE

This is a well-executed fix for the Copilot CI test failures. The changes are correct, complete, and well-tested.

**What was done:**
1. `getModels()` now dynamically fetches real model IDs via `client.listModels()` with a 5-minute TTL cache, falling back to the static list on failure or empty results.
2. `ownsModel()` checks both static and dynamic model caches.
3. `getModelForTier()` prefers models from the dynamic cache, using keyword-based tier matching when the static ID is unavailable.
4. `buildSdkConfig()` routes all three SDK tiers (opus/sonnet/haiku) to the same resolved model ID, preventing hangs when the SDK tries to use a model ID not available on the user's Copilot account.
5. Online tests now hard-fail if `gpt-5-mini` (the free-tier CI model) is missing, per the CLAUDE.md credential policy.
6. `gpt-5.3-codex` (alias `copilot-anthropic-codex`) has been correctly restored to the static model list.

**Verification of the codex restoration (the focus of this re-review):**
- `COPILOT_ANTHROPIC_MODELS` at `provider.ts:156-165` contains the `gpt-5.3-codex` entry with alias `copilot-anthropic-codex`.
- Unit test assertions at `provider.test.ts:66` and `provider.test.ts:75` verify `ownsModel` for both the alias and bare ID.
- The static list contains exactly the 5 expected models: claude-opus-4.6, claude-sonnet-4.6, gpt-5.3-codex, gemini-3-pro-preview, gpt-5-mini.

**Minor observations (P3, non-blocking):**
- `mapCopilotSdkModel` filters disabled models via `m.policy?.state !== 'disabled'` in both the `getModels()` filter and the `available` field of the returned ModelInfo. The double-check is harmless but redundant since disabled models are already filtered before `mapCopilotSdkModel` is called.
- The `inferModelFamily` function returns string literals that could be a union type for better type safety, but this is consistent with how `family` is typed elsewhere in the codebase.

No P0, P1, or P2 issues found. The implementation is clean, well-documented, and properly tested.

lsm (commented):

This is a follow-up re-review after the `gpt-5.3-codex` restoration commit. Two additional commits have been pushed since the task spec was written, and one of them introduces a discrepancy with the stated requirements.

---

### P2 — Gemini model ID diverges from the task spec

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts` line 167

The task prompt explicitly states the static model list should contain `gemini-3-pro-preview`. The current code has `gemini-3.1-pro-preview` (changed in commit `1a6cd535`).

The previous version of this entry had an inline comment explaining the choice:
```
// Intentionally "gemini-3-pro-preview" (not "gemini-3.1-pro-preview").
// The legacy CLI path used "gemini-3.1-pro-preview" because it passed IDs
// directly to Copilot CLI. This provider passes IDs as hints to the embedded
// HTTP server, which maps them to Copilot SDK sessions; the version suffix
// matters for routing.
```

Commit `1a6cd535` silently removed this rationale comment and changed the ID to `gemini-3.1-pro-preview` without explanation. The unit test assertion was updated to match, so tests pass — but whether this matches what the Copilot SDK actually expects is unclear.

If `gemini-3.1-pro-preview` is the correct real Copilot SDK model ID (i.e. the one `client.listModels()` would return for a Gemini 3 Pro model), then the change is correct and the task spec was stale. If `gemini-3-pro-preview` is still the right ID, this is a regression. Either way, the commit needs a clear justification comment explaining which ID the Copilot SDK actually returns and why it changed.

**The task spec says `gemini-3-pro-preview`. The code now has `gemini-3.1-pro-preview`. One of them is wrong, and the commit message does not explain which.**

---

### P3 — `validateCopilotToken` model change is correct and addresses a prior concern

Commit `2050ff5c` changed `validateCopilotToken` from `gpt-4o-mini` to `gpt-5-mini`. This addresses the P2 observation in the previous claude-sonnet-4-6 review. The change is correct: using the same designated free-tier CI model for token validation is consistent and avoids silent auth failures if `gpt-4o-mini` is removed.

---

### What is correct in this iteration

- `gpt-5.3-codex` (alias `copilot-anthropic-codex`) is present in `COPILOT_ANTHROPIC_MODELS` at line 156–165. Confirmed correct.
- Unit test assertions at `provider.test.ts:66` and `provider.test.ts:75` verify `ownsModel('copilot-anthropic-codex')` and `ownsModel('gpt-5.3-codex')`. Confirmed correct.
- The full static model list has exactly 5 entries: claude-opus-4.6, claude-sonnet-4.6, gpt-5.3-codex, gemini-3.1-pro-preview (disputed), gpt-5-mini.
- Online test `beforeAll` hard-fails if `gpt-5-mini` is absent from `client.listModels()` — correct per CLAUDE.md policy.
- `getModelForTier()` consults `dynamicModelsCache` and falls back to keyword-based selection from the dynamic list before using the static ID. This is correct.

---

### Summary

The `gpt-5.3-codex` restoration and `copilot-anthropic-codex` unit test restoration are correct. The Gemini model ID change from `gemini-3-pro-preview` to `gemini-3.1-pro-preview` conflicts with the task specification and the previously documented rationale. This needs to be resolved: either restore `gemini-3-pro-preview` with explanation if the original was correct, or add a clear comment explaining why the real Copilot SDK ID requires `gemini-3.1-pro-preview` and why the previous rationale comment is no longer valid.

lsm (commented):

### Recommendation: APPROVE

This PR correctly addresses the root cause of the CI timeout: hardcoded static model IDs that may not exist on the Copilot account, causing requests to fail or hang. The fix introduces dynamic model discovery via `client.listModels()` with appropriate caching and fallback.

### What was reviewed

**Core changes (7 commits on top of the branch baseline):**

1. **Dynamic model fetching** (`provider.ts:299-340`): `getModels()` now calls `client.listModels()` to discover real Copilot model IDs, with a 5-minute TTL cache (`dynamicModelsCacheExpiresAt`). Falls back to the static list on API failure or empty results. This is the key fix — CI tests no longer depend on hardcoded model IDs being valid.

2. **`getModelForTier()` dynamic lookup** (`provider.ts:353-376`): Checks the dynamic cache first; if the static ID is present, uses it; otherwise delegates to `pickModelForTier()` keyword-based heuristic. Falls back to static ID when no cache exists. Correct tier precedence.

3. **`pickModelForTier()` helper** (`provider.ts:74-94`): Keyword-based model selection per tier. `flash` intentionally excluded from sonnet tier to prevent Gemini Flash routing — well documented with comments.

4. **`ownsModel()` with dynamic cache** (`provider.ts:342-351`): Checks both static and dynamic model lists. The documented limitation (returns false if called before `getModels()`) is acceptable given the normal lifecycle where models are fetched before sessions are created.

5. **Online test hardening** (`anthropic-to-copilot-bridge-provider.test.ts:258-271`): `beforeAll` hard-fails if `gpt-5-mini` is not in the model list. All tests use `testModelId = 'gpt-5-mini'` — the free-tier model. This ensures CI detects plan changes immediately.

6. **`validateCopilotToken()` model update** (`provider.ts:698`): Changed from `gpt-4o-mini` to `gpt-5-mini` for consistency with the CI test model. Makes sense — if the CI model is `gpt-5-mini`, validation should use the same model.

7. **`gemini-3.1-pro-preview` in static list** (`provider.ts:168`): User-confirmed intentional change. The static list is a fallback/enrichment source, so the exact ID matters less now that dynamic discovery is in place.

### No issues found

- **Correctness**: The dynamic model fetching correctly gates on `clientCache` being initialized (guaranteed by the `ensureServerStarted()` -> `createServer()` -> `getOrCreateClient()` call chain in `getModels()`). The TTL cache prevents excessive API calls while staying reasonably fresh.

- **Edge cases handled**: Empty `listModels()` result falls back to static list (line 333). API exceptions are caught and logged (line 334-336). `pickModelForTier()` handles empty model lists and all-unavailable models (lines 79-80).

- **`buildSdkConfig()` sets all three tier slots to the same resolved model** (lines 416-418), preventing the original bug where `claude-haiku-4-5-20251001` leaked into the haiku slot and caused bridge failures.

- **Test coverage**: Unit tests cover TTL caching (cache hit, cache miss after expiry), empty/error fallback, `ownsModel()` with dynamic cache, `getModelForTier()` with and without dynamic cache, keyword matching, and the no-Anthropic-model-leak invariant. The online test validates the full end-to-end flow.

- **Security**: No new attack surface. The embedded server still binds to `127.0.0.1:0`. Token handling is unchanged.

- **CI integration**: The `providers-anthropic-copilot` matrix entry with 15-minute timeout and `COPILOT_GITHUB_TOKEN` secret is properly configured.

Clean implementation that solves the stated problem with good fallback behavior and test coverage.

lsm (commented):

**Recommendation: APPROVE** — Implementation is solid and addresses the root cause. One medium-priority concern and a few minor issues documented below.

---

## Summary

The PR correctly fixes the CI timeout by fetching real Copilot model IDs via `client.listModels()` with a 5-minute TTL cache, replacing the previous hardcoded static list. The all-tiers-to-same-model routing fix (`ANTHROPIC_DEFAULT_HAIKU_MODEL: resolvedId`) closes the original hang path. Tests are thorough and cover the new dynamic cache behavior.

---

## Issues Found

### P1 — `validateCopilotToken` now hard-depends on `gpt-5-mini` existing

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts`, line 698

`validateCopilotToken` is used to verify tokens discovered from `gh auth token` (source 4) and `~/.config/gh/hosts.yml` (source 5). If `gpt-5-mini` is unavailable on the user's Copilot account or plan, `createSession('gpt-5-mini')` will fail, causing the function to return `false` — effectively invalidating valid tokens and making the provider appear unavailable.

The previous behavior used `gpt-4o-mini`, and the PR comment acknowledges this risk: "If Copilot ever removes that model, validation will spuriously return false." The concern is that this same regression can occur if the user's Copilot plan simply doesn't include `gpt-5-mini` (e.g., an enterprise plan with different model availability).

A more robust approach would be to call `client.listModels()` and check for a non-empty response rather than creating a session with a specific model. The session creation is unnecessarily model-dependent for a token validity check.

---

### P2 — `pickModelForTier` 'pro' keyword can route Gemini to the opus tier

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts`, lines 83–95

The keyword list for the `opus` tier is `['opus', 'pro', 'ultra']`. The ID `gemini-3.1-pro-preview` contains `pro`, so if `claude-opus-4.6` is absent from the dynamic cache but `gemini-3.1-pro-preview` is present, Gemini Pro will be selected as the opus-tier model. This may be intentional (Gemini Pro is high-capability) but the test suite has no coverage for this case, and the behavior is non-obvious. A test verifying this is acceptable (or a comment clarifying it's intentional) would reduce confusion.

---

### P2 — Dynamic cache not invalidated when dynamic models fall to empty after a successful fetch

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts`, lines 326–333

When `listModels()` returns a non-empty list but all models are `disabled`, `mapped.length === 0` and the code falls back to `COPILOT_ANTHROPIC_MODELS` without setting `dynamicModelsCache`. Subsequent calls within the TTL window will keep calling `listModels()` again. This is a minor performance issue with no correctness impact, but worth noting.

---

### P3 — Static description for Gemini model is stale after ID change

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts`, line 173

The `gemini-3.1-pro-preview` entry has `description: 'Gemini 3 Pro Preview via GitHub Copilot'` (still says "3 Pro" not "3.1 Pro"). This is a cosmetic inconsistency since the name was updated to `'Gemini 3.1 Pro (Copilot)'` but the description was not.

---

### P3 — `ownsModel` with `dynamicModelsCache` docs note call-order limitation but no warning is emitted

**File:** `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts`, lines 253–263

The comment accurately documents the known limitation that `ownsModel()` returns `false` for real Copilot SDK model IDs if called before `getModels()`. This is fine as documentation, but the silent `return false` could cause confusing routing behavior in edge cases. A `logger.debug` warning when falling through to `false` (after the dynamic cache check) would aid debugging.

---

## Positives

- The `ensureServerStarted()` → `createServer()` → `getOrCreateClient()` call chain guarantees `clientCache` is always populated before `listModels()` is reached — the guard `if (this.clientCache)` on line 323 is therefore always true in normal operation, but correctly defensive for test scenarios where the cache is injected.
- All three tier-to-same-model routing is well-explained and tested with both a dedicated unit test and the context-manager invariant test.
- The `flash` keyword removal from sonnet tier keywords is well-reasoned and documented.
- TTL cache design matches the existing token cache pattern consistently.
- Online test correctly enforces `gpt-5-mini` presence as a hard-fail rather than a skip, per CLAUDE.md policy.
- The Gemini model ID change (`gemini-3-pro-preview` → `gemini-3.1-pro-preview`) is correctly treated as user-requested and reflected in tests.

lsm (commented):

**Recommendation: APPROVE**

The P1 issue from the previous review has been properly addressed. The changes in this iteration are focused and correct.

### validateCopilotToken fix (P1 resolved)

The `validateCopilotToken()` method at `packages/daemon/src/lib/providers/anthropic-copilot/provider.ts:688-713` now uses `client.listModels()` instead of `client.createSession('gpt-5-mini')`. This is the right approach:

1. **Model-agnostic validation** — a non-empty model list proves Copilot API access without coupling to any specific model. Enterprise users with different model catalogues will authenticate correctly.
2. **Simpler cleanup** — no need to `await session.disconnect()` since `listModels()` is a stateless call. The `client.stop()` in the `finally` block is sufficient.
3. **Timeout handling** — the `Promise.race` pattern with `clearTimeout` is correct. If `listModels()` rejects before the timeout fires, the dangling timeout will fire harmlessly against an already-settled promise. The `finally` block ensures `client.stop()` runs in all paths.

### Gemini model ID fix

Updated `gemini-3-pro-preview` to `gemini-3.1-pro-preview` in the static model list and unit test assertion. The stale comment claiming the old ID was intentional has been removed.

Note: `packages/web/src/hooks/__tests__/useModelSwitcher.test.ts` still references `gemini-3-pro-preview` in its mock data (lines 305-328), but this is harmless -- the test uses self-contained mock RPC responses and does not depend on the real static model list.

### Online test hardening

The online test at `packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts:258-271` now hard-fails if `gpt-5-mini` is missing from the model list, with a clear diagnostic message. This follows the CLAUDE.md hard-fail policy.

### No issues found

The changes are minimal, well-documented with JSDoc, and consistent with the existing error handling patterns. No security, correctness, or architectural concerns.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

The P1 issue — `validateCopilotToken()` calling `createSession('gpt-5-mini')` and thereby rejecting valid tokens from accounts that don't include `gpt-5-mini` — is now correctly fixed. The replacement uses `client.listModels()` to perform a plan-agnostic liveness check. The fix is sound, complete, and well-tested.

---

### What was verified

**Core fix (lines 688–713, `provider.ts`):**

The new `validateCopilotToken()` implementation:
- Creates a temporary `CopilotClient` with the candidate token
- Races `client.listModels()` against a 20-second hard timeout
- Returns `true` if the list is non-empty, `false` on any failure or timeout
- Always calls `client.stop()` in `finally` to avoid subprocess leaks

This is exactly the right approach. A non-empty model list is a definitive proof of Copilot API access without any coupling to a specific model ID being on the user's plan.

**Gemini entry description fix:**

`COPILOT_ANTHROPIC_MODELS` entry for `gemini-3.1-pro-preview` now has the correct description referencing 3.1 rather than 3. Cosmetic but correct.

**Dynamic model list (`getModels()`):**

- `client.listModels()` result is cached with the same 5-minute TTL as the token cache
- Disabled models (`policy.state === 'disabled'`) are correctly filtered out
- Falls back to `COPILOT_ANTHROPIC_MODELS` when the API call fails or returns empty
- Cache is consulted before making another API call

**`getModelForTier()` dynamic routing:**

- When `dynamicModelsCache` is populated: checks if the static fallback ID exists in the cache first (fast path), then falls back to `pickModelForTier()` keyword matching
- `pickModelForTier()` correctly places `flash` in the haiku-tier keywords and excludes it from sonnet-tier (preventing Gemini Flash models from shadowing the sonnet path)
- Falls back to first available model when no keyword matches

**All three SDK tier slots set to the same resolved model ID in `buildSdkConfig()`:**

`ANTHROPIC_DEFAULT_OPUS_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, and `ANTHROPIC_DEFAULT_HAIKU_MODEL` are all set to the user's resolved Copilot model. This prevents the original hang where the haiku slot defaulted to an Anthropic-native model ID that the embedded bridge cannot serve.

**Online tests correctly hard-fail:**

The online test `beforeAll` checks `isAvailable()` and hard-fails (throws) rather than skipping. It also hard-fails if `gpt-5-mini` is absent from the dynamic model list, with a clear error message explaining the CI expectation.

**Unit test coverage:**

New tests cover:
- TTL cache refresh behaviour
- `getModelForTier()` with static ID in dynamic cache (returns static ID)
- `getModelForTier()` with no static ID match (keyword selection)
- Keyword-miss fallback to first available model
- `client.listModels()` success, failure, empty-list, and caching paths
- `ownsModel()` with dynamically fetched IDs
- `validateCopilotToken` is properly mocked for `tryGhHostsToken`/`tryGhCliToken` tests

---

### No issues found

The implementation is correct, the fix is model-agnostic as intended, error handling (timeout, subprocess cleanup, cache invalidation, fallback) is complete, and test coverage is thorough. No P0/P1/P2/P3 issues identified.

PR #465, opened 2026-03-20
review: PR #461 - fix thinking blocks in SDK title generation

lsm:
## Summary

Reviewed PR #461 ("fix: handle thinking blocks in SDK title generation") and pushed an improved fix directly to the PR branch.

**Issue with the original PR:**
The original fix fell back to thinking block content when no text blocks were present. Thinking blocks contain internal model reasoning — not suitable as session titles (they can be thousands of words of internal monologue).

**Correct fix applied to PR #461:**
- Added `thinking: { type: 'disabled' }` to the SDK query options in `generateTitleWithSdk`
- Prevents adaptive-thinking models (e.g. Opus 4.6) from generating thinking blocks during title generation, fixing the root cause
- Reverted the unnecessary thinking-block fallback extraction logic
- Added 4 unit tests in `session-lifecycle-sdk-title.test.ts` verifying correct behavior

## Test plan
- [x] Unit tests pass for new test file
- [x] Existing session lifecycle tests unaffected
- [x] All pre-commit checks pass (lint, format, typecheck, knip)

### DIFF
--- a/packages/daemon/src/lib/agent/agent-session.ts
+++ b/packages/daemon/src/lib/agent/agent-session.ts
@@ -217,6 +217,7 @@ export class AgentSession
 	queryAbortController: AbortController | null = null;
 	firstMessageReceived = false;
 	startupTimeoutTimer: ReturnType<typeof setTimeout> | null = null;
+	startupTimeoutAutoRecoverAttempts = 0;
 	originalEnvVars: OriginalEnvVars = {};
 	// Whether to auto-queue /context after each turn (default: true)
 	// Disabled for room-managed agents to prevent interleaved messages after terminal state
@@ -781,6 +782,12 @@ export class AgentSession
 
 	async onMarkApiSuccess(): Promise<void> {
 		this.errorManager.markApiSuccess();
+	}
+
+	async onStartupTimeoutAutoRecover(): Promise<void> {
+		if (this.isCleaningUp()) return;
+		this.logger.warn('Auto-recovering after SDK startup timeout — starting fresh without resume.');
+		await this.startStreamingQuery();
 	}
 
 	// ============================================================================
--- a/packages/daemon/src/lib/agent/query-runner.ts
+++ b/packages/daemon/src/lib/agent/query-runner.ts
@@ -72,6 +72,10 @@ export interface QueryRunnerContext {
 	queryAbortController: AbortController | null;
 	firstMessageReceived: boolean;
 	startupTimeoutTimer: ReturnType<typeof setTimeout> | null;
+	// Tracks consecutive auto-recovery attempts for the current query.
+	// Reset to 0 when a query receives its first message (successful startup).
+	// Prevents infinite retry loops when the SDK is permanently broken.
+	startupTimeoutAutoRecoverAttempts: number;
 	originalEnvVars: OriginalEnvVars;
 	// Methods for state coordination
 	incrementQueryGeneration(): number;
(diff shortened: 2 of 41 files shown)


lsm (commented):

### Summary

This PR is a review/tracking PR for the changes pushed directly to PR #461's branch. PR #465 itself contains **an empty commit with zero code changes** — all actual modifications live on the `session/sdk-title-generation-empty-response-error-78a3afa7` branch (PR #461). I reviewed the code changes on that branch.

### Recommendation: APPROVE

The approach of disabling thinking via `thinking: { type: 'disabled' }` is the correct fix. It addresses the root cause rather than papering over it with a thinking-block fallback.

### Detailed Findings

**Correctness of the fix (P: none)**

The fix is sound. Title generation is a simple, short-text task that does not benefit from extended thinking. Disabling thinking:
1. Prevents the empty-response error when models like Opus 4.6 return only thinking blocks
2. Reduces latency and token usage for a trivial generation task
3. Uses the same `thinking: { type: 'disabled' }` pattern already established in `packages/daemon/src/lib/settings-manager.ts` (line 116)

The removal of the thinking-block fallback from the original PR #461 is the right call — model internal reasoning makes terrible session titles.

**Test coverage (P3: 1 nit)**

The 4 tests cover the key scenarios well:
- Verifies `thinking: { type: 'disabled' }` is passed to query options
- Verifies title extraction from text blocks
- Verifies markdown stripping
- Verifies fallback when no text blocks are returned

P3 nit: `afterEach` is imported on line 12 of `session-lifecycle-sdk-title.test.ts` but never used. This would likely be caught by the linter.

**Missing test scenario (P3: informational)**

There is no test for the case where `query()` throws an exception (e.g., SDK subprocess crash, network error). Looking at the `generateTitleAndRenameBranch` method, exceptions from `generateTitleWithSdk` are caught and trigger the fallback title path, so this is already handled in production code. A test for it would be nice but is not blocking.

**PR structure observation (P3)**

PR #465 is an empty-commit PR whose purpose is to document the review of PR #461. The actual code changes are on PR #461's branch. This is an unusual but functional workflow. The code changes themselves are clean and well-documented.

### No blocking issues found.

lsm (commented):

**Recommendation: APPROVE**

### Summary

This is a re-review of PR #465 after the worker addressed P2 feedback from the previous review. All four requested changes were correctly applied to the `session/sdk-title-generation-empty-response-error-78a3afa7` branch (PR #461), where the actual code lives.

### Feedback Items Verified

**1. Removed unused `afterEach` import — DONE**

The import on line 12 now reads:
```ts
import { describe, expect, it, beforeEach, mock } from 'bun:test';
```
`afterEach` is gone.

**2. Replaced in-test `mock.module` overrides with mutable `mockSdkMessages` variable — DONE**

All five `mock.module` calls are now at the top level. The `mockSdkMessages` variable is set in `beforeEach` (default: a valid text-block response) and overridden within individual test bodies only by mutating the variable — no `mock.module` calls inside test cases remain. This eliminates the previous test-order dependency issue.

**3. Added regression test for thinking-only assistant messages — DONE**

The new test `'should fall back to message text when assistant message contains only thinking blocks'` sets `mockSdkMessages` to return an assistant message whose content is a single `thinking` block. The implementation throws `'No text content in SDK response'`, which is caught by the `generateTitleFromMessage` error handler and falls back correctly. The test asserts `isFallback: true` and `title: 'Create a login form'`. This is exactly the original bug scenario.

**4. Separated "no assistant messages" fallback into its own test — DONE**

`'should fall back to message text when SDK returns no assistant messages'` is now a dedicated test with `mockSdkMessages = [{ type: 'result', subtype: 'success' }]`.

### Remaining Observations

**`makeSessionCache` `sessionOverrides` parameter removed (correct)** — The unused spread parameter has been cleaned up.

**No new issues introduced.** The five tests are well-structured, cover the critical paths, and are free of test-order dependencies. The implementation fix (`thinking: { type: 'disabled' }`) and the tests accurately reflect each other.

### No blocking issues found.

PR #466, opened 2026-03-20
fix: thread explicit provider through room-spawned agent sessions

lsm:
Root cause: AgentSessionInit was missing a `provider` field, so all
room-spawned sessions (leader, coder, general, planner) had
session.config.provider = undefined. query-runner.ts then fell back to
the deprecated detectProvider() heuristic which iterates providers in
registration order and returns Anthropic first for any claude-* model —
misrouting Copilot-targeted sessions to Anthropic.

Fix: Thread the provider resolved from the global model cache through
the full creation chain:
  resolveAgentModelWithProvider() → TaskGroupManager._provider →
  AgentSessionConfig.provider → AgentSessionInit.provider →
  createSessionFromInit → session.config.provider

Changes:
- AgentSessionInit: add provider?: string field
- createSessionFromInit: pass init.provider into SessionConfig
- LeaderAgentConfig / CoderAgentConfig / GeneralAgentConfig /
  PlannerAgentConfig: add provider?: string field
- createLeaderAgentInit, createCoderAgentInit, createGeneralAgentInit,
  createPlannerAgentInit: pass config.provider into AgentSessionInit
- TaskGroupManager: add _provider field, provider getter, update
  updateModel(model, provider?) to set both
- room-runtime.ts: use resolveAgentModelWithProvider() when building
  planner/worker/leader configs in spawnPlanningGroup and spawnGroupForTask

### DIFF
--- a/packages/daemon/src/lib/agent/agent-session.ts
+++ b/packages/daemon/src/lib/agent/agent-session.ts
@@ -83,6 +83,7 @@ import type {
 	SelectiveRewindResult,
 	SystemPromptConfig,
 	McpServerConfig,
+	Provider,
 } from '@neokai/shared';
 import type { SDKMessage } from '@neokai/shared/sdk';
 import type { DaemonHub } from '../daemon-hub';
@@ -130,7 +131,7 @@ export interface AgentSessionInit {
 	 * multiple providers (e.g. claude-opus-4.6 owned by both Anthropic and Copilot)
 	 * are routed to the correct provider.
 	 */
-	provider?: string;
+	provider?: Provider;
 
 	/** Enable coordinator mode — main agent orchestrates specialist sub-agents */
 	coordinatorMode?: boolean;
@@ -497,7 +498,7 @@ export class AgentSession
 			// Room-spawned sessions (leader, coder, general) pass this from the model cache
 			// so providers that share canonical model IDs (e.g. anthropic vs anthropic-copilot
 			// for 'claude-opus-4.6') are routed deterministically.
-			provider: init.provider as import('@neokai/shared').Provider | undefined,
+			provider: init.provider,
 		};
 
 		const metadata: SessionMetadata = {
--- a/packages/daemon/src/lib/agent/agent-session.ts
+++ b/packages/daemon/src/lib/agent/agent-session.ts
@@ -417,6 +417,18 @@ export class AgentSession
 				hasUpdates = true;
 			}
 
+			// Update config.provider if it changed since the session was first persisted.
+			// Room-spawned agent sessions use per-UUID IDs so this path is not reached for
+			// them today. Long-lived sessions (room chat, lobby) set their provider through
+			// the UI model picker, not init.provider. The update is included defensively so
+			// that if stable IDs are ever used for agent sessions the provider stays current.
+			if (init.provider !== undefined && session.config.provider !== init.provider) {
+				const nextConfig = { ...session.config, provider: init.provider };
+				updates.config = nextConfig;
+				session = { ...session, config: nextConfig };
+				hasUpdates = true;
+			}
+
 			if (hasUpdates) {
 				db.updateSession(init.sessionId, updates);
 			}
--- a/packages/daemon/src/lib/agent/agent-session.ts
+++ b/packages/daemon/src/lib/agent/agent-session.ts
@@ -123,6 +123,15 @@ export interface AgentSessionInit {
 	/** Model ID - defaults to default model */
 	model?: string;
 
+	/**
+	 * Provider ID for this session (e.g. 'anthropic', 'anthropic-copilot', 'glm').
+	 * When set, provider routing is deterministic (no deprecated detectProvider fallback).
+	 * Room-spawned sessions (leader, worker) must set this so models shared between
+	 * multiple providers (e.g. claude-opus-4.6 owned by both Anthropic and Copilot)
+	 * are routed to the correct provider.
+	 */
+	provider?: string;
+
 	/** Enable coordinator mode — main agent orchestrates specialist sub-agents */
 	coordinatorMode?: boolean;
 
@@ -484,6 +493,11 @@ export class AgentSession
 			coordinatorMode: init.coordinatorMode,
 			agent: init.agent,
 			agents: init.agents,
+			// Explicit provider — prevents deprecated detectProvider() fallback in query-runner.
+			// Room-spawned sessions (leader, coder, general) pass this from the model cache
+			// so providers that share canonical model IDs (e.g. anthropic vs anthropic-copilot
+			// for 'claude-opus-4.6') are routed deterministically.
+			provider: init.provider as import('@neokai/shared').Provider | undefined,
 		};
 
 		const metadata: SessionMetadata = {
--- a/packages/daemon/src/lib/providers/anthropic-provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-provider.ts
@@ -23,13 +23,22 @@ import { resolveSDKCliPath, isRunningUnderBun } from '../agent/sdk-cli-resolver.
 const CANONICAL_SDK_IDS = new Set(['default', 'sonnet', 'opus', 'haiku', 'sonnet[1m]']);
 
 /**
- * Detect if a model ID is a full version-specific ID (e.g., claude-sonnet-4-5-20250929)
- * vs a canonical short ID (e.g., sonnet, opus, haiku)
+ * Detect if a model ID is a full version-specific ID vs a canonical short ID.
+ *
+ * Recognises two formats:
+ * - Dash-separated: claude-sonnet-4-5-20250929, claude-opus-4-5-20251101
+ * - Dot-notation: claude-opus-4.6, claude-sonnet-4.6 (newer SDK format)
  */
 function isFullVersionId(modelId: string): boolean {
-	// Full IDs match pattern: claude-{family}-{version}-{date}
-	// e.g., claude-sonnet-4-5-20250929, claude-opus-4-5-20251101
-	return /^claude-(sonnet|opus|haiku)-[\d-]+$/.test(modelId);
+	// Dash-separated: claude-{family}-{major}-{minor}[-{date}]
+	if (/^claude-(sonnet|opus|haiku)-[\d-]+$/.test(modelId)) {
+		return true;
+	}
+	// Dot-notation: claude-{family}-{major}.{minor}
+	if (/^claude-(sonnet|opus|haiku)-\d+\.\d+$/.test(modelId)) {
+		return true;
+	}
+	return false;
 }
 
 /**
@@ -76,16 +85,23 @@ function parseModelId(
 		};
 	}
 
-	// Full version IDs: claude-{family}-{major}-{minor}-{date}
-	// Example: claude-sonnet-4-5-20250929
-	const match = modelId.match(/^claude-(sonnet|opus|haiku)-(\d+)-(\d+)(?:-\d{8})?$/);
-	if (match) {
-		const family = match[1];
-		const major = match[2];
-		const minor = match[3];
+	// Full version IDs: claude-{family}-{major}-{minor}[-{date}]
+	// Example: claude-sonnet-4-5-20250929, claude-opus-4-5-20251101
+	const dashMatch = modelId.match(/^claude-(sonnet|opus|haiku)-(\d+)-(\d+)(?:-\d{8})?$/);
+	if (dashMatch) {
 		return {
-			family,
-			version: `${major}.${minor}`, // Extract version as "4.5"
+			family: dashMatch[1],
+			version: `${dashMatch[2]}.${dashMatch[3]}`, // e.g. "4.5"
+		};
+	}
+
+	// Dot-notation IDs: claude-{family}-{major}.{minor}
+	// Example: claude-opus-4.6, claude-sonnet-4.6
+	const dotMatch = modelId.match(/^claude-(sonnet|opus|haiku)-(\d+)\.(\d+)$/);
+	if (dotMatch) {
+		return {
+			family: dotMatch[1],
+			version: `${dotMatch[2]}.${dotMatch[3]}`, // e.g. "4.6"
 		};
 	}
 
--- a/packages/daemon/src/lib/room/agents/coder-agent.ts
+++ b/packages/daemon/src/lib/room/agents/coder-agent.ts
@@ -18,6 +18,7 @@ import type {
 	SessionFeatures,
 	AgentDefinition,
 	SubagentConfig,
+	Provider,
 } from '@neokai/shared';
 
 const DEFAULT_CODER_MODEL = 'claude-sonnet-4-5-20250929';
@@ -39,7 +40,7 @@ export interface CoderAgentConfig {
 	model?: string;
 	/** Provider ID resolved from the model (e.g. 'anthropic', 'anthropic-copilot').
 	 *  When set, routing is deterministic — no deprecated detectProvider fallback. */
-	provider?: string;
+	provider?: Provider;
 	/** Summaries of previously completed tasks in the same goal */
 	previousTaskSummaries?: string[];
 }
--- a/packages/daemon/src/lib/room/agents/coder-agent.ts
+++ b/packages/daemon/src/lib/room/agents/coder-agent.ts
@@ -37,6 +37,9 @@ export interface CoderAgentConfig {
 	sessionId: string;
 	workspacePath: string;
 	model?: string;
+	/** Provider ID resolved from the model (e.g. 'anthropic', 'anthropic-copilot').
+	 *  When set, routing is deterministic — no deprecated detectProvider fallback. */
+	provider?: string;
 	/** Summaries of previously completed tasks in the same goal */
 	previousTaskSummaries?: string[];
 }
@@ -455,6 +458,7 @@ export function createCoderAgentInit(config: CoderAgentConfig): AgentSessionInit
 			context: { roomId: config.room.id },
 			type: 'coder',
 			model: config.model ?? DEFAULT_CODER_MODEL,
+			provider: config.provider,
 			agent: 'Coder',
 			agents: {
 				Coder: coderAgentDef,
@@ -478,6 +482,7 @@ export function createCoderAgentInit(config: CoderAgentConfig): AgentSessionInit
 		context: { roomId: config.room.id },
 		type: 'coder',
 		model: config.model ?? DEFAULT_CODER_MODEL,
+		provider: config.provider,
 		contextAutoQueue: false,
 	};
 }
--- a/packages/daemon/src/lib/room/agents/general-agent.ts
+++ b/packages/daemon/src/lib/room/agents/general-agent.ts
@@ -9,7 +9,7 @@
  */
 
 import type { AgentSessionInit } from '../../agent/agent-session';
-import type { Room, RoomGoal, NeoTask, SessionFeatures } from '@neokai/shared';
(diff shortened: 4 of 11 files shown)


lsm (commented):

The PR correctly threads `provider` through the agent creation chain (factories, configs, `AgentSessionInit`, `SessionConfig`) and significantly reduces the blast radius of the bug. However, there is a residual routing failure for models that share the same `id` across providers, plus a type-safety issue and test coverage gaps.

---

### P0 — Residual provider misrouting for shared model IDs

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 371

```ts
const modelInfo = availableModels.find((m) => m.id === modelId || m.alias === modelId);
return { model: modelId, provider: modelInfo?.provider };
```

`getAvailableModels('global')` returns an array that contains BOTH `anthropic:claude-opus-4.6` and `anthropic-copilot:claude-opus-4.6` as separate entries (see `mergeWithFallbackModels` which keys by `provider:id`). Because the Anthropic entry appears first (Anthropic is registered before Copilot in `factory.ts`, and fallback models are prepended), `find` always returns the Anthropic entry.

The UI in `packages/web/src/components/room/RoomAgents.tsx` calls `handleSelect(model.id)` (line 199) which stores only the bare model ID (e.g., `'claude-opus-4.6'`) in `agentModels.leader`. There is no provider stored alongside it. When `resolveAgentModelWithProvider` looks up this ID, it picks the wrong provider.

**Consequence:** The original bug — Copilot sessions being routed to Anthropic — still occurs when the user selects a Copilot model whose `id` is shared with an Anthropic model (specifically `claude-opus-4.6` and `claude-sonnet-4.6`). The fix is complete only for models with unique IDs (e.g. `gpt-5.3-codex`) or when the alias is used (e.g., `copilot-anthropic-opus`).

**Root cause not fully addressed:** The room config data model stores only a model ID string (`agentModels: Record<string, string>`), with no provider field. To fully fix this, either the room config must store `{ modelId, providerId }` pairs, or the UI must store the provider-unique alias instead of the bare `model.id`.

---

### P0 — Empty cache fallback silently reverts to broken behavior

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 370

```ts
const availableModels = getAvailableModels('global');
```

The comment says this is "safe to call here (cache is pre-warmed at startup)", but `getAvailableModels` returns an empty array if the cache is not yet populated (see model-service.ts:254–257). If `resolveAgentModelWithProvider` is called before `initializeModels` completes — which can happen early in room startup — `modelInfo` will be `undefined` and `provider` will be `undefined`. This silently falls back to the deprecated `detectProvider` path in `query-runner.ts`, recreating the original misrouting bug with no error or warning.

---

### P1 — Type unsafety: string cast to Provider interface

**File:** `packages/daemon/src/lib/agent/agent-session.ts`, line 500

```ts
provider: init.provider as import('@neokai/shared').Provider | undefined,
```

`init.provider` is `string | undefined`. `SessionConfig.provider` is typed as `Provider | undefined` (the full interface with `id`, `displayName`, `capabilities`, `isAvailable()`, etc.). Casting a plain string to `Provider` is a lie to the type system. Consumers in `query-runner.ts` (`session.config.provider as string | undefined`) and `session-lifecycle.ts` already work around this with reverse casts, showing the type is semantically a string identifier — the `Provider` type annotation on `SessionConfig.provider` should be `string` (or `ProviderId`), or a dedicated `providerId?: string` field should be added. This PR perpetuates the existing type confusion rather than cleaning it up.

---

### P1 — Tests do not cover the critical path in `resolveAgentModelWithProvider`

**File:** `packages/daemon/tests/unit/room/agent-provider-routing.test.ts`

All 16 tests only verify that the factory functions (`createCoderAgentInit`, `createLeaderAgentInit`, etc.) pass an explicitly-supplied `provider` value through to `AgentSessionInit`. There are no tests for:

1. `resolveAgentModelWithProvider` in `room-runtime.ts` — the site where the model cache is queried and the provider is resolved from a model ID string. This is where the bug still lives for shared-ID models.
2. The ambiguity case: two providers sharing the same model ID (e.g., `claude-opus-4.6`) in a populated cache — verifying which provider is returned.
3. Empty cache behavior: when `getAvailableModels` returns `[]`, `provider` is `undefined`.
4. Alias-based lookup: verifying that `copilot-anthropic-opus` resolves to `anthropic-copilot`.

The tests verify the "wire" is connected but not whether the right signal flows through it.

---

### P2 — UI can show duplicate model entries with shared IDs

**File:** `packages/web/src/components/room/RoomAgents.tsx`, lines 191–207

Both `anthropic:claude-opus-4.6` and `anthropic-copilot:claude-opus-4.6` appear in the dropdown with the same `key={model.id}`, causing a Preact key collision. The display label at line 119 (`models.find((m) => m.id === value)`) also picks the first match, so after selecting the Copilot variant the button may display the Anthropic model name. This is a pre-existing issue but the PR's changes depend on the UI correctly conveying intent, making it relevant.

---

### P3 — Unreachable `resolveAgentModel` wrapper

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 339

`resolveAgentModel` is now a thin wrapper over `resolveAgentModelWithProvider` that discards the provider. All callsites in this PR use `resolveAgentModelWithProvider` directly. If there are no remaining callers of `resolveAgentModel`, it can be removed. If there are, those callers should be migrated to `resolveAgentModelWithProvider`.

---

### Summary

The PR correctly extends the provider field through all four agent factories and the `TaskGroupManager`, which is the right structural approach. However, the provider resolution in `resolveAgentModelWithProvider` still fails for models with shared IDs (the most affected case: Copilot's `claude-opus-4.6`), because the room config stores only a bare model ID and the cache lookup is ambiguous. The root cause — the data model not capturing which provider a model ID belongs to — is unaddressed. The tests validate the plumbing but not the logic that was actually broken.

lsm (commented):

**Recommendation: APPROVE**

This PR correctly fixes the provider misrouting bug for room-spawned agent sessions. The fix threads an explicit `provider` field through the full session creation chain: `resolveAgentModelWithProvider()` -> agent config -> `AgentSessionInit` -> `SessionConfig`. The implementation is clean, consistent, and well-tested.

### What was reviewed

- `packages/daemon/src/lib/agent/agent-session.ts` — New `provider?: string` on `AgentSessionInit`, passed to `SessionConfig` in `createSessionFromInit`
- `packages/daemon/src/lib/room/agents/{coder,general,leader,planner}-agent.ts` — All agent config interfaces and init factories updated consistently
- `packages/daemon/src/lib/room/runtime/task-group-manager.ts` — New `_provider` field and updated `updateModel(model, provider?)`
- `packages/daemon/src/lib/room/runtime/room-runtime.ts` — `resolveAgentModelWithProvider()` used in constructor, `refreshRoomConfig`, `spawnPlanningGroup`, and `spawnGroupForTask`
- `packages/daemon/tests/unit/room/agent-provider-routing.test.ts` — 16 unit tests covering all agent factories

### Findings

**P2 — `resolveAgentModelWithProvider` uses `find()` on a flat model list with duplicate IDs** (`room-runtime.ts:370-372`)

When both Anthropic and Copilot providers register `claude-opus-4.6`, the global model cache contains two entries with the same `id` but different `provider` values. `availableModels.find((m) => m.id === modelId)` returns whichever was registered first (Anthropic, per `factory.ts` order). This means:
- If the room's `defaultModel` is `claude-opus-4.6` and the user intended Copilot, the provider will resolve to `anthropic` (wrong).
- Conversely, if the user intended Anthropic and Anthropic is first, it resolves correctly.

This is a **pre-existing architectural limitation** (rooms store only `defaultModel` as a string, not a `(model, provider)` pair), so the fix is still a net improvement over the previous behavior where provider was always `undefined`. The fix correctly addresses the stated bug (Anthropic opus-4.6 going to Copilot). However, the ambiguity for the reverse case remains. A future improvement could store `defaultProvider` alongside `defaultModel` in the room config.

**P3 — Type assertion** (`agent-session.ts:500`)

`init.provider as import('@neokai/shared').Provider | undefined` is an unsafe cast from `string` to the `Provider` union type. In practice this is safe since `ModelInfo.provider` values come from provider registration code, but using a runtime validation or making `AgentSessionInit.provider` typed as `Provider` would be cleaner.

### Summary

The fix is correct for the reported bug. All agent types (leader, coder, general, planner) and all code paths (eager leader creation, deferred leader recovery, planning group spawn, task group spawn) properly thread the provider. Tests verify the factory layer thoroughly. The remaining ambiguity in `resolveAgentModelWithProvider` for shared model IDs is a pre-existing issue and not a regression.

lsm (commented):

### Recommendation: APPROVE

### Summary

This PR correctly fixes the provider misrouting bug where selecting "anthropic opus-4.6" as a room subagent would send traffic to GitHub Copilot instead of Anthropic. The fix is clean, well-scoped, and properly tested.

### What the fix does

The core problem was that `AgentSessionInit` (and downstream `SessionConfig`) had no explicit `provider` field for room-spawned sessions. This forced `query-runner.ts` to fall back to the deprecated `detectProvider()` heuristic, which iterates providers in registration order and returns the first match. When both Anthropic and Copilot claim `claude-opus-4.6`, whichever provider happened to be registered first won — an ordering-dependent, fragile resolution.

The fix threads an explicit `provider` through the full creation chain:

1. **`resolveAgentModelWithProvider()`** in `room-runtime.ts` (line 353) looks up the model in the global model cache to find the correct provider. Falls back to `detectProvider()` only when the cache miss occurs. This is the single boundary where `string` from `ModelInfo.provider` is cast to the `Provider` union type.

2. **Agent config interfaces** (`LeaderAgentConfig`, `CoderAgentConfig`, `GeneralAgentConfig`, `PlannerAgentConfig`) all gain `provider?: Provider` fields using the shared union type.

3. **`createXxxAgentInit()` factories** all pass `config.provider` through to `AgentSessionInit.provider`.

4. **`TaskGroupManager`** carries `_provider` alongside `_model` and passes it to `LeaderAgentConfig` in both `spawn()` and `routeWorkerToLeader()`.

5. **`AgentSession.fromInit()`** stores `init.provider` into `SessionConfig.provider`, which `query-runner.ts` already reads for deterministic routing.

6. **`isFullVersionId()` / `parseModelId()`** in `anthropic-provider.ts` now handle dot-notation IDs like `claude-opus-4.6`, ensuring Anthropic's model deduplication logic works correctly when the SDK returns both `opus` and `claude-opus-4.6`.

### Type safety

All `provider` fields use the shared `Provider` union type (`'anthropic' | 'glm' | 'minimax' | 'anthropic-copilot' | 'anthropic-codex'`). There is exactly one `as Provider` cast at the model-cache boundary in `resolveAgentModelWithProvider()` (lines 374, 381), which is the correct place since `ModelInfo.provider` is typed as `string`. This is well-documented in the comments.

### Test coverage

- `agent-provider-routing.test.ts`: Verifies all four `createXxxAgentInit()` factories correctly pass provider through to `AgentSessionInit` (including undefined, anthropic, anthropic-copilot, glm).
- `room-runtime-provider-resolution.test.ts`: Integration-level tests verifying the full `resolveAgentModelWithProvider()` -> spawn -> session init chain, covering cache hit (anthropic first), cache miss fallback, single-provider cache, and propagation to both worker and leader.

### Minor observations (P3)

1. **`ModelInfo.provider` is `string` not `Provider`** (line 27 of `models.ts`). The cast in `resolveAgentModelWithProvider` is necessary but fragile — if a provider registers models with a `provider` string not in the `Provider` union, it silently passes through as an invalid type at runtime. This is a pre-existing design decision, not introduced by this PR.

2. The `resolveAgentModelWithProvider` method uses `availableModels.find()` which returns the first match. When both Anthropic and Copilot have `claude-opus-4.6` in the cache, the resolution depends on insertion order. The test at line 82-100 of `room-runtime-provider-resolution.test.ts` documents this as intentional (Anthropic is loaded first). This is reasonable given the model cache is populated in provider registration order.

No blocking issues found. The fix is correct, complete, well-typed, and properly tested.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped fix. The core routing bug is correctly addressed and all P0/P1 issues from the previous iteration have been resolved. The code is correct at every integration point I checked.

---

### Summary of Changes Reviewed

The fix threads `provider?: Provider` from `resolveAgentModelWithProvider()` in `room-runtime.ts` through `TaskGroupManager._provider`, into each agent config interface (`LeaderAgentConfig`, `CoderAgentConfig`, `GeneralAgentConfig`, `PlannerAgentConfig`), then into `AgentSessionInit.provider`, and finally into `SessionConfig.provider`. The `query-runner.ts` consumer already existed and uses `session.config.provider` to skip `detectProvider()` — this PR completes the pipe that was broken.

---

### What Is Correct

**Type safety at the cache boundary.** The single `as Provider` cast is correctly located in `resolveAgentModelWithProvider()` at line 374 of `room-runtime.ts`, where `ModelInfo.provider` is `string` in the shared type but guaranteed to be a valid `Provider` union member by construction. All other fields use the `Provider` type directly — no stray `string` or `as Provider` casts downstream.

**Fallback is sound.** When the model is absent from the cache, `detectProvider()` is called and its result is also cast at line 381. The comment correctly explains why this is safe for `claude-*` models (Anthropic is registered first).

**Both spawn paths covered.** `spawnPlanningGroup` (line 2820) and `spawnGroupForTask` (line 2954) both call `resolveAgentModelWithProvider` for planner/coder/general/leader roles before building configs. The lazy-recreate path in `routeWorkerToLeader` (line 436 of `task-group-manager.ts`) also uses `this.provider` from `TaskGroupManager`, which is kept up-to-date via `updateRoom()`.

**`updateRoom()` keeps `TaskGroupManager` in sync.** When the room config changes mid-run, `updateRoom()` at line 415 calls `resolveAgentModelWithProvider` and calls `taskGroupManager.updateModel(model, provider)`. New sessions spawned after a room config change will use the correct provider.

**`isFullVersionId` / `parseModelId` fix is correct.** The dot-notation regex `^claude-(sonnet|opus|haiku)-\d+\.\d+$` at line 38 and the corresponding `dotMatch` at line 100 of `anthropic-provider.ts` correctly handle `claude-opus-4.6`, ensuring the deduplication logic in `convertSdkModels` fires when both the canonical `opus` alias and the full-version dot ID appear in SDK responses.

**Tests are meaningful.** `room-runtime-provider-resolution.test.ts` exercises the model cache lookup, cache miss fallback, and the "only Copilot has the model" scenario. `agent-provider-routing.test.ts` verifies each factory function passes the provider field end-to-end. The `anthropic-provider.test.ts` additions cover both deduplication paths (with/without canonical alias present).

---

### Minor Observations

**P2 — Fallback comment slightly overstates safety (`room-runtime.ts` line 350–351).** The comment says "detectProvider returns Anthropic first for any 'claude-*' model ID" and implies this is always correct. This is true in the normal case because Anthropic is registered first, but it does not hold if a user explicitly configured `claude-opus-4.6` to route to `anthropic-copilot` via `room.config.agentModels.leader` *and* the model happens to be absent from the cache (e.g., cache cleared mid-run). In that edge case the fallback would silently misroute back to Anthropic. Unlikely and pre-existing, but the comment could acknowledge the limitation rather than imply unconditional correctness.

**P2 — `fromInit` existing-session path does not update `config.provider` in DB.** When `AgentSession.fromInit()` finds a session already in DB (e.g., a stable room-chat ID), it updates several fields but not `config.provider`. For short-lived worker/leader sessions (which use per-UUID IDs at `task-group-manager.ts` line 255–256), this path is never hit so the fix is correct today. If stable IDs are ever used for agent sessions, stale providers could persist silently.

**P3 — Test at `anthropic-provider.test.ts:187` calls `provider.buildSdkConfig('default')` but `AnthropicProvider.buildSdkConfig()` takes no parameters.** TypeScript silently ignores the extra argument; the test passes and is functionally correct. The call should read `provider.buildSdkConfig()` to accurately reflect the method signature.

**P3 — `TaskGroupManager` stores `_provider` only for the leader model, not the worker model.** Worker provider is re-resolved from the cache on each `spawnGroupForTask` call, which is correct because the worker model can differ from the leader model. The asymmetry (leader provider is stored state, worker provider is ephemeral) is intentional but not documented; a brief comment in `TaskGroupManagerConfig` would help the next reader.

---

### No P0 or P1 Issues Found

The fix correctly addresses the root cause. The type chain is sound, the cast location is appropriate, and the tests meaningfully validate the routing behavior.

lsm (commented):

**Recommendation: APPROVE**

### Summary

This iteration addresses the two P2 issues raised in the previous review. Both changes are minimal, well-scoped, and correctly implemented.

### Change 1: Expanded fallback comment in `resolveAgentModelWithProvider`

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` (lines 376-385)

The comment now explicitly acknowledges the edge case where a Copilot-hosted `claude-*` model would misroute to Anthropic on a cache miss. The limitation is clearly stated: detectProvider iterates in registration order and returns the first owner, so cache misses for Copilot-targeted models will resolve to Anthropic. The comment also notes this is expected to be transient since models are pre-warmed at startup. This is exactly what was requested -- making the limitation explicit rather than implying unconditional correctness.

### Change 2: Provider update in `fromInit` existing-session path

**File:** `packages/daemon/src/lib/agent/agent-session.ts` (lines 420-430)

The new block updates `config.provider` when an existing session is loaded from DB but the init specifies a different provider. The implementation follows the exact same pattern as the surrounding update blocks (context, worktree, runtimeInitFingerprint):

- Guard: `init.provider !== undefined && session.config.provider !== init.provider`
- Spread into nextConfig, update `updates` and `session`, set `hasUpdates`
- Comment honestly acknowledges this path is not reached today (room-spawned agents use per-UUID session IDs, so they always take the create path), but is included defensively for future-proofing

The guard correctly avoids overwriting an existing provider with `undefined` when `init.provider` is not set -- the `!== undefined` check ensures only explicit provider values trigger the update.

### No issues found

Both changes are documentation/defensive improvements with no behavioral risk. The existing test suite (47 tests across 3 files) passes cleanly. No new test coverage is needed for these changes since the provider update path in `fromInit` is explicitly documented as not reachable today.

lsm (commented):

**Recommendation: APPROVE**

Both P2 fixes from the previous review round are correctly implemented.

### P2 Fix 1 — Fallback comment in `resolveAgentModelWithProvider`

`packages/daemon/src/lib/room/runtime/room-runtime.ts` (lines 376–387)

The updated comment now explicitly names the misroute edge case: when the model cache is unavailable and the room has a Copilot-hosted `claude-*` model configured, `detectProvider()` returns Anthropic first (registration order), causing misrouting. The comment correctly qualifies this as expected to be transient (cache pre-warmed at startup) without over-claiming correctness. This is a precise and honest documentation of the limitation.

### P2 Fix 2 — `config.provider` update in `fromInit` existing-session path

`packages/daemon/src/lib/agent/agent-session.ts` (lines 420–430)

The guard `if (init.provider !== undefined && session.config.provider !== init.provider)` is correct:

- The `!== undefined` check avoids clobbering a persisted provider with `undefined` when a caller (e.g. lobby or room chat) doesn't pass a provider in `init`. This is important because long-lived sessions manage their provider through the UI model picker, not through `init.provider`.
- The in-memory `session` spread and the `updates.config` are both populated before `db.updateSession` is called, so both DB and the local variable stay consistent.
- The subsequent `mcpServers` in-memory merge (lines 440–448) spreads `session.config` after the provider update, so `provider` is preserved correctly.

The accompanying comment (lines 421–424) correctly explains why this path is rarely triggered in practice today (per-UUID session IDs for room-spawned agents) while justifying the defensive inclusion for future stable-ID scenarios.

### No new issues found

Both fixes are scoped correctly, do not introduce regressions, and the reasoning in the inline comments matches the actual code behavior.

lsm:
Superseded by https://github.com/lsm/neokai/pull/467

PR #467, opened 2026-03-20
refactor: remove detectProvider heuristic entirely

lsm:
## Summary

- Deletes ProviderRegistry.detectProvider() — the ambiguous model-ID-only heuristic that caused the original wrong-provider routing bug
- Removes ProviderContextManager.detectProvider() forwarding wrapper
- Replaces the legacy fallback in resolveProvider with a direct Anthropic fallback for old sessions
- Replaces the QueryRunner ternary detectProvider fallback with registry.get('anthropic')
- Removes deprecated ProviderService methods isGlmModel, detectProviderFromModel, translateModelIdForSdk
- Removes or fixes all tests that exercised the deleted methods

## Test plan

- bun test on provider-registry, context-manager, provider-service tests: all pass
- bun run typecheck: clean
- Full make test-daemon: only pre-existing Dev Proxy/Copilot env failures (8), none from changed files


lsm (commented):

**Recommendation: APPROVE**

Clean, focused refactoring that fully achieves the stated goal. All six items in the task description are correctly implemented:

1. **`ProviderRegistry.detectProvider()`** removed from `packages/daemon/src/lib/providers/registry.ts` -- the ambiguous first-match loop is gone.
2. **`ProviderContextManager.detectProvider()`** forwarding wrapper removed from `packages/daemon/src/lib/providers/context-manager.ts`.
3. **`ProviderContextManager.resolveProvider()`** fallback simplified -- legacy sessions now go directly to `registry.get('anthropic')` instead of through the heuristic detection chain.
4. **`QueryRunner`** ternary at line 141-143 correctly replaced: when no explicit provider is stored, it calls `providerRegistry.get('anthropic')` directly instead of the removed `detectProvider`.
5. **`ProviderService.isGlmModel`, `detectProviderFromModel`, `translateModelIdForSdk`** all removed from `packages/daemon/src/lib/provider-service.ts`.
6. **Tests updated**: `detectProvider` test blocks removed from `provider-registry.test.ts`, `context-manager.test.ts`, and `provider-service.test.ts`. One-arg `getEnvVarsForModel` calls updated to two-arg form with explicit provider IDs.

**Verification performed:**
- Grep for `detectProvider` (excluding `detectProviderForModel`) across the entire `packages/` tree returns zero matches -- no dangling references.
- Grep for removed method names (`isGlmModel`, `detectProviderFromModel`) returns zero matches in application code.
- `translateModelIdForSdk` correctly remains on the `Provider` interface and individual provider implementations (GLM, MiniMax) -- only the `ProviderService` wrapper that called `detectProvider` internally was removed.
- `getEnvVarsForModel` signature now requires both `modelId` and `providerId` (no optional second arg), and all call sites pass both arguments.

No issues found. The change is minimal, correct, and leaves no dead code behind.

lsm (commented):

**Recommendation: APPROVE** — This is clean, well-scoped deletion work with no regressions found. A few minor observations noted below.

---

### Summary

The PR correctly removes all three layers of the `detectProvider` heuristic: the `ProviderRegistry.detectProvider()` implementation, its forwarding wrapper on `ProviderContextManager`, and the three `ProviderService` methods that called it (`isGlmModel`, `detectProviderFromModel`, `translateModelIdForSdk`). The `QueryRunner` and `resolveProvider` fallbacks are now explicit Anthropic lookups, which is correct and unambiguous.

Typecheck passes clean. No dead references to the deleted symbols remain in source or test files.

---

### Findings

**P3 — Silent fallback when stored provider ID is unrecognised (context-manager.ts:101-117)**

In `resolveProvider`, when `session.config.provider` is set to a value that is not registered (e.g. a provider that was removed or renamed), the code silently falls through `registry.get(session.config.provider)` returning `undefined` and then picks Anthropic with no warning emitted. Before this PR the same path hit `detectProvider`, which would also silently fall through to Anthropic — so this is not a behaviour regression, but it is worth noting. A `log.warn` before the Anthropic fallback would make this diagnosable.

**P3 — QueryRunner unknown-provider path triggers legacy GLM auth check (query-runner.ts:141-197)**

If `session.config.provider` is set but `detectProviderForModel` returns `undefined` (unknown provider), `provider` is `undefined`, the `provider?.isAvailable` guard is skipped, and execution falls into the `!provider?.isAvailable` legacy block that checks Anthropic/GLM env vars. This is a pre-existing edge case not introduced by this PR (it existed before), but now that detection is gone this path is more likely to be hit silently. Not a regression, just flagging for awareness.

**P3 — Removed online tests previously exercised `getEnvVarsForModel` GLM env var contract**

The deleted tests in `packages/daemon/tests/online/glm/glm-provider.test.ts` were the only tests that verified the actual GLM env var shape (`ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, model tier mappings). The remaining online tests test availability and listing but not env var construction. This coverage gap existed implicitly before and is acceptable given the methods being removed called `detectProvider`, but the GLM env var contract (`buildSdkConfig` output) could use a unit-level coverage addition eventually.

---

### What works well

- The deletion is complete and conservative: no partial stubs, no optional fallbacks, no feature flags.
- The `resolveProvider` simplification (removing the intermediate detect step) is correct: the two-step `detectProvider → Anthropic` was always equivalent to just `Anthropic` for the legacy path.
- The `QueryRunner` fallback `registry.get('anthropic')` is a strict improvement over the first-match heuristic that could return Copilot or Codex when those providers are registered and their `ownsModel` matches.
- Test renames (`'should fall back to detection when...' → 'should fall back to Anthropic when...'`) accurately reflect the new semantics.
- The 1-arg `getEnvVarsForModel` calls in tests are all correctly updated to 2-arg form.
- No web package references to the deleted methods exist (grep confirms zero hits).

lsm (commented):

**Recommendation: APPROVE**

Clean, well-scoped removal of the `detectProvider` heuristic across the codebase. The changes are correct and complete for the stated goal.

### What was done

1. **`ProviderRegistry.detectProvider()`** — removed entirely. No remaining callers.
2. **`ProviderContextManager.resolveProvider()`** — now throws descriptive errors instead of silently falling back to Anthropic. Two error paths: missing provider in config, and unregistered provider ID.
3. **`ProviderContextManager.detectProvider()`** (public deprecated wrapper) — removed.
4. **`ProviderService`** — three deprecated methods removed: `isGlmModel()`, `detectProviderFromModel()`, `translateModelIdForSdk()`. No remaining callers outside their own tests.
5. **`QueryRunner`** — legacy fallback changed from `detectProvider(modelId)` to `registry.get('anthropic')` with a clear annotation that this is a temporary shim for pre-PR#466 sessions. This is a reasonable interim approach.
6. **Tests** — updated to assert new throwing behavior; sessions that previously relied on implicit detection now have explicit `provider: 'anthropic'` in their config. Deleted tests for removed methods. Test coverage looks thorough.

### Minor observations (P3)

- `ProviderService.getProviderEnvVars()` (line 373 of `provider-service.ts`) still has `session.config.provider || 'anthropic'` as a silent fallback. This is in the legacy compat layer rather than the core context manager, so it is consistent with keeping `QueryRunner`'s `registry.get('anthropic')` shim — but worth noting for a future cleanup pass.
- The error messages reference "PR #466" directly. This is fine for internal debugging but slightly unusual in user-facing errors. Given these errors would only surface to developers debugging session state, this is acceptable.

No blocking issues found. The implementation matches the stated goal of iteration 2 feedback.

lsm (commented):

---

### Summary

This PR correctly removes the deprecated `detectProvider` heuristic from `ProviderRegistry`, `ProviderContextManager`, and three `ProviderService` methods. The `resolveProvider` change in `ProviderContextManager` now throws instead of silently falling back, and the `QueryRunner` gets an annotated legacy shim. The goal is right. However there are two concrete bugs and one stale-comment problem introduced by the PR.

---

### P0 — Production break: `QueryOptionsBuilder.build()` will throw for legacy sessions

**File:** `packages/daemon/src/lib/agent/query-options-builder.ts`, lines 77 and 88

`build()` calls `contextManager.createContext(session)` on line 77 and `contextManager.createContext(fallbackSession)` on line 88. `createContext` now throws immediately when `session.config.provider` is absent:

```ts
// context-manager.ts line 107-112
if (!providerId) {
  throw new Error(
    `Session '${session.id}' has no provider stored in config. ` +
      `All sessions created after PR #466 store an explicit provider ID.`
  );
}
```

Any user who loads a session from the SQLite database that was created before PR #466 (i.e., before `provider` was written into `session.config`) will hit this throw the moment they try to send a message. The `QueryRunner` already has a legacy shim comment acknowledging these sessions exist. `QueryOptionsBuilder` — which is called *before* `QueryRunner` — does not and will crash first. This is a hard regression for existing users on upgrade.

The `QueryRunner` shim at lines 142-145:
```ts
const explicitProviderId = session.config.provider as string | undefined;
const provider = explicitProviderId
  ? providerRegistry.detectProviderForModel(modelId, explicitProviderId)
  : providerRegistry.get('anthropic');
```
…correctly handles the pre-#466 case. The same tolerance needs to be applied inside `createContext` or the `QueryOptionsBuilder` call sites need a guard.

---

### P0 — `QueryOptionsBuilder` GLM online tests will also throw

**File:** `packages/daemon/tests/online/glm/glm-provider.test.ts`, lines 343-347, 376-384, 412-416, 429, 432

The `QueryOptionsBuilder` describe block (lines 338-444) constructs sessions via `createTestSession(env.testWorkspace, { config: { model: 'glm-5' } })` without a `provider` field. These sessions will cause `contextManager.createContext` to throw immediately, producing test failures unrelated to what the tests intend to verify. The comment on line 342 — "model-based detection, no provider config needed" — is now incorrect; a `provider` field is required.

---

### P1 — Stale comment and misleading file header in the online test file

**File:** `packages/daemon/tests/online/glm/glm-provider.test.ts`, lines 3-5 and 14

The module-level JSDoc still reads:
```
 * Tests the model-based provider detection flow:
 * 1. ProviderService model ID detection (glm-* models)
 * 2. QueryOptionsBuilder model-based env var injection
 * ...
 * Tests will SKIP if credentials are not available
```
All of the model-based detection tests were deleted in this PR. The header now describes tests that no longer exist. The skip comment is also relevant only for the direct-API-call tests; the surrounding describe blocks don't skip — they simply fail when `createContext` throws. Update the header to reflect the remaining test content.

---

### P2 — The `requiresQueryRestart` method silently propagates the throw to callers

**File:** `packages/daemon/src/lib/providers/context-manager.ts`, lines 134-144

`requiresQueryRestart` calls `resolveProvider(session)` internally and is documented as returning `boolean`. For a legacy session (no stored provider), it will now throw instead of returning `true` (the safer conservative value). No production caller currently exists, so this is latent rather than immediately broken — but if the method is ever called during a model-switch on a legacy session it will surface as an unhandled exception rather than triggering a restart. The method should either document that it can throw, or catch the `resolveProvider` error and return `true`.

---

### P3 — Minor: `QueryOptionsBuilder` fallback-model path reconstructs `contextManager` unnecessarily

**File:** `packages/daemon/src/lib/agent/query-options-builder.ts`, lines 82-83

`const contextManager = getProviderContextManager();` is declared a second time in the `if (config.fallbackModel)" block, shadowing the outer variable. This is an existing issue not introduced by this PR, but if someone adds the legacy-session guard at the outer `contextManager.createContext` call (as required by the P0 fix), they'll need to remember to apply the same guard to this shadowed variable.

PR #484, opened 2026-03-20
feat: add VisualCanvas component with pan and zoom (Task 1.2)

lsm:
- Add shared types: Point, Size, ViewportState, NodePosition
- Add screenToCanvas/canvasToScreen coordinate conversion helpers
- Implement VisualCanvas with two-finger scroll pan, spacebar+drag pan,
  and Ctrl/pinch zoom clamped to [0.25, 2.0] toward cursor
- Extract applyWheelEvent as pure function for deterministic unit testing
- Add 13 unit tests covering rendering, pan/zoom logic, scale clamping,
  zoom-toward-cursor behaviour, and coordinate conversion math

### DIFF
--- a/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
@@ -1,0 +1,198 @@
+/**
+ * VisualCanvas
+ *
+ * A pannable, zoomable viewport container for the visual workflow editor.
+ *
+ * Pan methods:
+ *  - Two-finger trackpad scroll (wheel event without ctrlKey)
+ *  - Spacebar + left-click drag
+ *
+ * Zoom methods:
+ *  - Trackpad pinch (wheel event with ctrlKey=true)
+ *  - Ctrl/Cmd + scroll wheel
+ *  - Scale is clamped to [0.25, 2.0] and zooms toward cursor position.
+ */
+
+import { useEffect, useRef, useCallback } from 'preact/hooks';
+import type { ComponentChildren } from 'preact';
+import type { ViewportState } from './types';
+
+export const MIN_SCALE = 0.25;
+export const MAX_SCALE = 2.0;
+
+function clampScale(s: number): number {
+	return Math.min(MAX_SCALE, Math.max(MIN_SCALE, s));
+}
+
+/**
+ * Pure function: compute the next ViewportState for a wheel event.
+ *
+ * @param vp        Current viewport state
+ * @param deltaX    Horizontal scroll delta
+ * @param deltaY    Vertical scroll delta (negative = zoom in)
+ * @param isZoom    True when the wheel event should zoom (ctrlKey pressed / pinch)
+ * @param cursorX   Cursor X relative to the canvas container (for zoom-toward-cursor)
+ * @param cursorY   Cursor Y relative to the canvas container
+ */
+export function applyWheelEvent(
+	vp: ViewportState,
+	deltaX: number,
+	deltaY: number,
+	isZoom: boolean,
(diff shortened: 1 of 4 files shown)


lsm (commented):

Clean, well-structured implementation that meets all acceptance criteria. The code is well-documented, the pure function extraction (`applyWheelEvent`) makes the logic highly testable, and all 13 tests pass. Typecheck is clean.

### What looks good

- **Pure function extraction**: `applyWheelEvent` is a smart design choice -- it separates DOM concerns from viewport math, making the logic deterministic and easy to test.
- **Zoom-toward-cursor math**: Correctly preserves the canvas point under the cursor during zoom. The test on line 112-131 verifies this property properly.
- **Coordinate conversion**: `screenToCanvas` and `canvasToScreen` are correctly implemented as inverses. Tests verify both directions with offsets and non-unit scale.
- **Non-passive wheel listener**: Correctly uses JSX `onWheel` prop (Preact registers non-passive) so `preventDefault()` works for suppressing browser zoom/scroll.
- **Stale closure prevention**: `viewportRef` pattern on line 77-78 ensures event handlers always read fresh state.
- **Input guard**: Spacebar handler correctly skips INPUT/TEXTAREA elements (line 101-102).

### Issues

**P2: Missing `e.preventDefault()` on spacebar keydown** (`VisualCanvas.tsx`, line 99-107)

When spacebar is pressed to enter pan mode, the handler does not call `e.preventDefault()`. In a scrollable page context, this means pressing space will also trigger the browser's default "scroll down" behavior. Since this is a canvas component that may be embedded in various layouts, it would be defensive to prevent default when capturing the space key for pan mode. Add `e.preventDefault()` after the INPUT/TEXTAREA guard.

**P3: `applyWheelEvent` / `MIN_SCALE` / `MAX_SCALE` not re-exported from barrel** (`index.ts`)

These are exported from `VisualCanvas.tsx` but not from the barrel `index.ts`. This is fine if they are considered internal, but `MIN_SCALE`/`MAX_SCALE` could be useful for consumers building zoom controls (e.g., a zoom slider). Consider whether these should be part of the public API.

**P3: No guard against `scale === 0` in `screenToCanvas`** (`types.ts`, line 30)

Division by `viewport.scale` would produce `Infinity` if scale were somehow 0. In practice, `clampScale` ensures `[0.25, 2.0]`, but `screenToCanvas` is a standalone utility that does not enforce this invariant. A defensive consumer could pass `scale: 0`. This is very low risk given the current controlled usage.

### Verdict

Solid work. The P2 is worth addressing before merge since it can cause unexpected page scrolling in production. The P3s are minor and optional.

**Recommendation: REQUEST_CHANGES** (for the P2 spacebar preventDefault issue)

lsm (commented):

Re-review of the fix commit `021e5317`. All three issues from the previous review have been correctly addressed:

1. **P2 fix (spacebar preventDefault):** `e.preventDefault()` is now called at line 103 of `VisualCanvas.tsx`, after the INPUT/TEXTAREA guard and before setting `spacebarDown`. This correctly prevents the browser's default scroll-down behavior on spacebar while still allowing typing in form fields. Verified correct.

2. **P3 fix (barrel exports):** `applyWheelEvent`, `MIN_SCALE`, and `MAX_SCALE` are now re-exported from `index.ts`. Verified correct.

3. **P3 fix (screenToCanvas scale=0 guard):** `types.ts` line 29 now uses `const scale = viewport.scale || 1` to avoid division by zero. This is a reasonable defensive guard since `screenToCanvas` is a public utility that could receive arbitrary input. Verified correct.

All 13 tests pass. No remaining issues found.

**Recommendation: APPROVE**

PR #486, opened 2026-03-20
fix: sync room chat session model when defaultModel changes to prevent glm-5-turbo errors

lsm:
Room chat sessions (`room:chat:<roomId>`) were being created with whatever
model was configured at room-creation time. If the room's defaultModel was
later changed (or was already using `glm-5-turbo`), the session kept the
stale model/provider, causing "There's an issue with the selected model
(glm-5-turbo)" errors.

Changes:
- room-handlers: when `room.update` changes `defaultModel`, immediately
  update the corresponding room chat session's config (model + provider).
  Uses a lightweight `inferProviderForModel` utility to avoid loading the
  full provider registry (which requires optional `@github/copilot-sdk`).
- room-runtime-service: same lightweight heuristic replaces the previous
  async `resolveProviderForModel` that used dynamic factory import; also
  fixes a missing `await` bug on that call site.
- test: add unit test asserting that `room.update` syncs the session
  model from glm-5-turbo/glm to sonnet/anthropic.

### DIFF
--- a/packages/daemon/src/lib/agent/agent-session.ts
+++ b/packages/daemon/src/lib/agent/agent-session.ts
@@ -83,6 +83,7 @@ import type {
 	SelectiveRewindResult,
 	SystemPromptConfig,
 	McpServerConfig,
+	Provider,
 } from '@neokai/shared';
 import type { SDKMessage } from '@neokai/shared/sdk';
 import type { DaemonHub } from '../daemon-hub';
@@ -122,6 +123,9 @@ export interface AgentSessionInit {
 
 	/** Model ID - defaults to default model */
 	model?: string;
+
+	/** Provider ID for this session — if omitted, auto-detected from model or falls back to Anthropic */
+	provider?: string;
 
 	/** Enable coordinator mode — main agent orchestrates specialist sub-agents */
 	coordinatorMode?: boolean;
@@ -470,6 +474,7 @@ export class AgentSession
 
 		const config: SessionConfig = {
 			model: init.model ?? defaultModel,
+			provider: init.provider as Provider | undefined,
 			maxTokens: 4096,
 			temperature: 1.0,
 			// Pass through system prompt from init
--- a/packages/daemon/src/lib/providers/registry.ts
+++ b/packages/daemon/src/lib/providers/registry.ts
@@ -76,6 +76,20 @@ export class ProviderRegistry {
 	}
 
 	/**
+	 * Find the first registered provider that owns this model ID.
+	 * Uses each provider's ownsModel() heuristic for auto-detection.
+	 * Returns undefined if no provider claims the model.
+	 */
+	findProviderForModel(modelId: string): Provider | undefined {
+		for (const provider of this.providers.values()) {
+			if (provider.ownsModel(modelId)) {
+				return provider;
+			}
+		}
+		return undefined;
+	}
+
+	/**
 	 * Resolve provider by explicit (modelId, providerId) pair — fully deterministic.
 	 *
 	 * Both the model ID and provider ID must be known at the call site. This is the
--- a/packages/daemon/src/lib/providers/registry.ts
+++ b/packages/daemon/src/lib/providers/registry.ts
@@ -83,7 +83,7 @@ export class ProviderRegistry {
 	 */
 	findProviderForModel(modelId: string): Provider | undefined {
 		for (const provider of this.providers.values()) {
-			if (provider.ownsModel(modelId)) {
+			if (typeof provider.ownsModel === 'function' && provider.ownsModel(modelId)) {
 				return provider;
 			}
 		}
@@ -239,20 +239,28 @@ export function resetProviderRegistry(): void {
 }
 
 /**
- * Infer the provider for a given model ID based on known naming conventions.
+ * Infer the provider for a given model ID.
  *
- * This is a static heuristic that avoids loading the full provider registry
- * (which may require optional SDK dependencies). It covers all known provider
- * model naming conventions:
- * - `glm-*` → 'glm'
- * - `minimax-*` → 'minimax'
- * - `gpt-*` → 'anthropic-codex' (Codex bridge models like gpt-5.3-codex)
- * - Everything else → 'anthropic' (standard Claude models, and Copilot models
- *   which share Claude model IDs and are routed via explicit providerId)
+ * Strategy (two-pass):
+ * 1. Try the live registry via `findProviderForModel()` — covers all registered providers
+ *    including `anthropic-copilot` and `anthropic-codex`, which have dynamic model lists.
+ *    Returns the correct provider whenever the daemon has finished initializing.
+ * 2. Fall back to a static naming heuristic when the registry is empty (e.g., unit tests
+ *    or early-startup callers where providers are not yet registered):
+ *    - `glm-*` → 'glm'
+ *    - `minimax-*` → 'minimax'
+ *    - `gpt-*` → 'anthropic-codex'
+ *    - everything else → 'anthropic'
  *
- * @public Exported so callers that cannot use the full registry can still infer providers.
+ * Callers should NOT call `initializeProviders()` just for this function — the registry
+ * is populated at daemon startup; calling it lazily at spawn-time is safe and avoids
+ * test-interference from re-registering providers.
  */
 export function inferProviderForModel(modelId: string): ProviderIdStr {
+	// Live registry lookup (populated at daemon startup, empty in unit tests)
+	const fromRegistry = getProviderRegistry().findProviderForModel(modelId)?.id;
+	if (fromRegistry) return fromRegistry as ProviderIdStr;
+	// Static fallback when registry is empty
 	if (modelId.startsWith('glm-') || modelId === 'glm') return 'glm';
 	if (modelId.startsWith('minimax-') || modelId === 'minimax') return 'minimax';
 	if (modelId.startsWith('gpt-')) return 'anthropic-codex';
--- a/packages/daemon/src/lib/providers/registry.ts
+++ b/packages/daemon/src/lib/providers/registry.ts
@@ -11,6 +11,7 @@
 
 import { createLogger } from '@neokai/shared/logger';
 import type { Provider, ProviderId, ProviderInfo } from '@neokai/shared/provider';
+import type { Provider as ProviderIdStr } from '@neokai/shared';
 
 const log = createLogger('kai:providers:registry');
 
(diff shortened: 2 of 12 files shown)


lsm (commented):

### P0: Fix does not address the actual root cause -- leader/worker agent sessions still lack provider

The task description states the bug is: *"selecting `glm-5-turbo` as the leader agent model causes the leader session to use the Anthropic/Claude Code backend instead of the GLM backend."*

The root cause is in `packages/daemon/src/lib/agent/query-runner.ts` lines 142-145:

```ts
const explicitProviderId = session.config.provider as string | undefined;
const provider = explicitProviderId
    ? providerRegistry.detectProviderForModel(modelId, explicitProviderId)
    : providerRegistry.get('anthropic');
```

When `session.config.provider` is `undefined`, the query runner **always falls back to the Anthropic provider**, regardless of the model name. This is the actual bug.

However, this PR only fixes the **room chat session** (`room:chat:<roomId>`), not the **leader or worker agent sessions**. Looking at the session creation path:

1. `createLeaderAgentInit()` in `packages/daemon/src/lib/room/agents/leader-agent.ts` (lines 1138-1156 and 1160-1177) creates an `AgentSessionInit` with `model` but **no `provider` field**
2. `AgentSessionInit` (line 101 of `agent-session.ts`) does not even have a `provider` field in its interface
3. `createSessionFromInit()` (line 465 of `agent-session.ts`) creates `SessionConfig` without a `provider` field
4. So leader sessions will always hit the `providerRegistry.get('anthropic')` fallback in `query-runner.ts`

The fix syncs the room chat session's model/provider on `room.update` and on runtime startup, but **leader and worker sessions are created fresh via `AgentSession.fromInit()` each time a task group is spawned** -- and they still have no `provider` in their config. The actual error "There's an issue with the selected model (glm-5-turbo)" will still occur for leader sessions.

**To actually fix this, the provider needs to be set either:**
- In `AgentSessionInit` (add a `provider` field and pass it from `createLeaderAgentInit` / `createCoderAgentInit`)
- Or in `createSessionFromInit()` using the same `inferProviderForModel` heuristic
- Or in `query-runner.ts` by using a model-to-provider heuristic when `explicitProviderId` is undefined instead of blindly falling back to anthropic

### P0: `room.create` has the same latent bug

In `room-handlers.ts` lines 72-81, when creating a room chat session, the `config` only sets `model` with no `provider`. If a room is created with `defaultModel: 'glm-5-turbo'`, the room chat session will be created with no provider, and the first query will fail. The PR only fixes `room.update`, not `room.create`.

### P1: Duplicated `inferProviderForModel` heuristic

The same heuristic is duplicated in two places:
- `room-handlers.ts` line 22-26 (module-level function `inferProviderForModel`)
- `room-runtime-service.ts` lines 439-443 (private method `resolveProviderForModel`)

These will inevitably drift. The function should be defined once in a shared location (e.g., `@neokai/shared` or a utility module in `packages/daemon/src/lib/providers/`).

### P1: Heuristic is incomplete -- missing `anthropic-copilot` and `anthropic-codex`

The `Provider` type (`packages/shared/src/types.ts:192`) includes `'anthropic-copilot'` and `'anthropic-codex'`, but the heuristic only maps `glm-*` and `minimax-*`, defaulting everything else to `'anthropic'`. While this may work today (copilot/codex models might use anthropic model names), it is fragile and will silently break if provider-specific model names are introduced.

### P2: Style -- import placement after function declaration

In `room-handlers.ts`, the `inferProviderForModel` function is placed between the `import` block and the remaining imports (line 17-26 is the function, line 27+ resumes imports). This is unconventional and could confuse readers. The function should be placed after all imports.

### Summary

This PR partially mitigates the symptom for room chat sessions but **does not fix the actual bug** for leader/worker sessions, which is the scenario described in the task. The root cause -- `AgentSessionInit` not carrying a `provider` field, and `query-runner.ts` defaulting to anthropic when provider is absent -- is not addressed.

lsm (commented):

Overall this PR correctly addresses the core bug -- room chat sessions created with `glm-5-turbo` would use the Anthropic provider because the provider was never stored in session config. The fix correctly propagates provider through the agent session init chain and syncs the room chat session when `defaultModel` changes. However, there are several issues to address.

---

### P0: Heuristic misses `anthropic-copilot` and `anthropic-codex` providers

**Files:** `room-runtime.ts:356-360`, `room-runtime-service.ts:439-443`, `room-handlers.ts:22-26`

The `resolveProviderForModel` / `inferProviderForModel` heuristic only handles `glm-*` and `minimax-*`, defaulting everything else to `'anthropic'`. The `Provider` type includes five variants:

```typescript
type Provider = 'anthropic' | 'glm' | 'minimax' | 'anthropic-copilot' | 'anthropic-codex';
```

Models from Copilot (e.g., `claude-3-5-sonnet-20241022` accessed via GitHub Copilot) or Codex would be incorrectly mapped to `'anthropic'` instead of `'anthropic-copilot'` or `'anthropic-codex'`. This means if a room is configured to use a Copilot-provided model, the session would be routed to the wrong provider -- the exact same class of bug this PR is fixing.

The comment says "This avoids loading the full provider registry (which requires optional SDK deps)" but this creates a maintenance hazard: any new provider added to the `Provider` type will silently fall through to `'anthropic'`.

Consider either:
1. Using `ProviderRegistry.findProviderForModel()` (which was added in this PR but is never called anywhere), or
2. At minimum, adding the Copilot/Codex model prefixes to the heuristic, or
3. Accepting that Copilot/Codex models cannot be used as room default models and documenting that limitation

### P1: `ProviderRegistry.findProviderForModel()` is dead code

**File:** `packages/daemon/src/lib/providers/registry.ts:83-90`

This method was added as part of the PR but is never called by any code path. All three call sites use the duplicated heuristic instead. Either use it or remove it -- dead code that looks like it should be the canonical solution is confusing for future maintainers.

### P1: `room.create` does not set provider in session config

**File:** `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:72-78`

When a room is created, the room chat session is created with only `model` in config -- no `provider`. If the room is created with a non-Anthropic `defaultModel` (e.g., `glm-5-turbo`), the session starts with no provider. The fix only applies on `room.update` and on runtime startup sync, but there is a window between creation and the runtime startup where the session has the wrong provider. The `inferProviderForModel` function is already available in this file -- it should be used at creation time too:

```typescript
config: {
    model: room.defaultModel,
    provider: room.defaultModel ? inferProviderForModel(room.defaultModel) : undefined,
},
```

### P1: Three duplicated heuristic functions

**Files:** `room-runtime.ts:356`, `room-runtime-service.ts:439`, `room-handlers.ts:22`

The same heuristic is copy-pasted in three places. This violates DRY and guarantees they will drift. Extract to a shared utility (e.g., in `@neokai/shared` or a providers utility file that does not import the full registry).

### P2: Return type mismatch in `room-runtime.ts`

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts:356`

`resolveProviderForModel` returns `string` while the other two copies return `Provider`. Since `TaskGroupManagerConfig.provider` is typed as `string | undefined`, this works at runtime, but it loses type safety -- a typo like `'antrhopic'` would compile fine. It should return `Provider` for consistency.

### P2: `room.create` params do not include `defaultModel`

**File:** `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:44-65`

The `room.create` handler destructures params without `defaultModel`, yet it passes `room.defaultModel` to the session config. If the `createRoom` method defaults it, this works, but it means the provider can never be correctly inferred at creation time since `room.defaultModel` might be undefined.

### P3: Import ordering style

**File:** `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:16-31`

The `inferProviderForModel` function definition is placed between the first import and the rest of the imports. This is unusual -- function definitions should come after all imports.

lsm (commented):

**Recommendation: APPROVE**

This is iteration 3 addressing the bug where selecting `glm-5-turbo` as the leader agent model caused the leader session to use the Anthropic backend instead of the GLM backend. The core issues from previous reviews have been addressed well.

### What was done

1. **`inferProviderForModel` consolidated in `registry.ts`** (line 255-260): Single source of truth for the model-to-provider heuristic. Covers `glm-*`, `minimax-*`, `gpt-*` (codex), and defaults to `anthropic`. The `anthropic-copilot` case is intentionally not in the heuristic since Copilot models share Claude model IDs and are routed via explicit `providerId` from the UI — this is correct and well-documented in the JSDoc.

2. **Used consistently in all 3 previously-duplicated places**:
   - `room-handlers.ts` (lines 69, 160) — `room.create` and `room.update`
   - `room-runtime.ts` (line 358) — via `resolveProviderForModel()` wrapper
   - `room-runtime-service.ts` (line 470) — sync stale room chat sessions

3. **`room.create` fixed** (line 69): Now sets `provider` when creating the room chat session via `inferProviderForModel(room.defaultModel)`.

4. **Provider propagation chain is complete**: `inferProviderForModel` -> `RoomRuntime.resolveProviderForModel` -> `TaskGroupManager.provider` -> `LeaderAgentConfig.provider` -> `AgentSessionInit.provider` -> `SessionConfig.provider` -> `query-runner.ts` reads `session.config.provider` to select the correct backend.

5. **All agent factories updated**: `coder-agent.ts`, `general-agent.ts`, `leader-agent.ts`, `planner-agent.ts` all accept and pass through `config.provider`.

6. **7 unit tests for `inferProviderForModel`**: Cover glm prefix, bare glm, minimax prefix, bare minimax, gpt prefix, claude prefix, and unknown fallback. All meaningful.

7. **room-handlers test**: Verifies that `room.update` with `defaultModel: 'sonnet'` syncs the chat session to `provider: 'anthropic'`.

1. **`ProviderRegistry.findProviderForModel()` is still dead code** (line 84-91 in `registry.ts`). It was added in this PR but never called anywhere. The standalone `inferProviderForModel()` function handles the use case without needing the full registry. This is a minor dead code issue — not blocking.

2. **No test for `room.create` provider propagation** — the test file only tests `room.update` sync. A test verifying that `room.create` passes `provider` to `sessionManager.createSession` would be nice but the function is simple enough and `inferProviderForModel` has its own tests.

3. **Empty string model edge case** — `inferProviderForModel('')` returns `'anthropic'` which is a safe fallback. The callers guard with `room.defaultModel ? inferProviderForModel(room.defaultModel) : undefined` so empty string would only reach the function if it's truthy (which empty string is not in JS). No issue here.

Overall this is a clean, well-structured fix that correctly addresses the root cause across all the necessary layers.

lsm (commented):

**Re-review of follow-up commit `ae83f67` addressing P3 nits.**

### Changes reviewed

1. **`typeof provider.ownsModel === 'function'` guard** (`packages/daemon/src/lib/providers/registry.ts:86`): Correct and safe. The `ownsModel` method is required by the `Provider` interface, but test code can register partial mocks that omit it. The runtime `typeof` check prevents `TypeError: provider.ownsModel is not a function` without masking real bugs — a provider missing `ownsModel` simply won't match, which is the right behavior (silent skip, not crash).

2. **Registry-first fallback logic** (`inferProviderForModel`, lines 259-268): Sound design. The two-pass strategy is well-documented and handles both runtime (registry populated) and test/early-startup (registry empty) scenarios correctly. The `as ProviderIdStr` cast on line 262 is acceptable since provider IDs are string-typed and the `Provider.id` field aligns with `ProviderIdStr`.

3. **No regressions**: The existing `inferProviderForModel` tests (lines 461-493) run against an empty registry (no `beforeEach` registration), so they exercise the static fallback path. The `findProviderForModel` tests within the `ProviderRegistry` describe block exercise the registry path. Both paths are covered.

### Verdict

**Recommendation: APPROVE.** No new issues introduced. The follow-up cleanly addresses the previously noted P3 concerns.

PR #488, opened 2026-03-20
feat: add SVG overlay layer for edges in VisualCanvas (Task 1.3)

lsm:
- Add <svg> as first child of the transform layer with position:absolute,
  pointer-events:none, and overflow:visible to cover the full canvas
- Accept optional `edgeLayer` render prop that receives current viewport
  state and renders SVG elements (paths, circles) inside the SVG layer
- Add 5 new tests verifying SVG layer presence, positioning, first-child
  order, edge content injection, and viewport state forwarding

### DIFF
--- a/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
@@ -58,9 +58,16 @@ interface VisualCanvasProps {
 	children?: ComponentChildren;
 	viewportState: ViewportState;
 	onViewportChange: (state: ViewportState) => void;
+	/** Render prop for injecting SVG edge content. Receives current viewport state. */
+	edgeLayer?: (viewport: ViewportState) => ComponentChildren;
 }
 
-export function VisualCanvas({ children, viewportState, onViewportChange }: VisualCanvasProps) {
+export function VisualCanvas({
+	children,
+	viewportState,
+	onViewportChange,
+	edgeLayer,
+}: VisualCanvasProps) {
 	const containerRef = useRef<HTMLDivElement>(null);
 
 	// Track spacebar state for pan-drag mode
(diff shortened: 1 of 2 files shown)


lsm (commented):

Clean, well-scoped implementation that meets all acceptance criteria.

### What looks good

- The SVG layer is correctly placed as the first child of the transform div, ensuring it renders behind DOM node children (lower paint order).
- Being inside the transform layer means the SVG content automatically inherits pan/zoom transforms without extra logic.
- `pointer-events: none` correctly prevents the SVG from intercepting mouse events meant for nodes.
- `overflow: visible` is the right choice since the parent transform div has no explicit dimensions -- SVG paths/edges can extend in any direction.
- The `edgeLayer` render prop is optional and cleanly short-circuits with `edgeLayer?.(viewportState)`.
- Tests are thorough: presence, styling, z-order, content injection, and viewport state forwarding.

### Minor observations (P3)

1. **SVG dimensions with percentage on unsized parent** (`VisualCanvas.tsx:208-209`): The SVG uses `width: '100%'` / `height: '100%'` but the transform div has no explicit width/height, so the SVG will effectively be 0x0. This works fine because `overflow: visible` is set, but it means the SVG element itself has no meaningful bounds for hit testing or `getBBox()` calculations. If future code needs to query SVG dimensions, this could be surprising. Not a problem now since `pointer-events: none` is set.

2. **Viewport state passed to edgeLayer**: Since the SVG is already inside the transformed container, the viewport state passed to `edgeLayer` is primarily useful for scale-dependent rendering (e.g., keeping stroke widths constant across zoom levels via `strokeWidth: 1 / viewport.scale`). This is a good design choice but worth a brief doc comment on the prop to clarify the intended use case.

**Recommendation: APPROVE** -- No functional issues found. The implementation is minimal, correct, and well-tested.

PR #489, opened 2026-03-20
feat: add CanvasToolbar with zoom controls and fit-to-view (Task 1.4)

lsm:
- CanvasToolbar component with zoom-in (+0.25), zoom-out (-0.25), reset
  (scale=1, offset=0,0), and fit-to-view buttons
- computeFitToView pure function: computes scale/offset to center all
  nodes with configurable padding, clamps to [MIN_SCALE, MAX_SCALE]
- VisualCanvas now accepts optional `nodes` and `showToolbar` props and
  renders CanvasToolbar as an absolute bottom-right overlay; tracks
  container size via ResizeObserver for accurate fit-to-view
- 16 new tests covering computeFitToView edge cases and toolbar interactions

### DIFF
--- a/packages/web/src/components/space/visual-editor/CanvasToolbar.tsx
+++ b/packages/web/src/components/space/visual-editor/CanvasToolbar.tsx
@@ -1,0 +1,163 @@
+/**
+ * CanvasToolbar
+ *
+ * A small floating toolbar positioned at the bottom-right of the canvas
+ * that provides zoom controls and a fit-to-view button.
+ */
+
+import type { NodePosition, ViewportState } from './types';
+import { MIN_SCALE, MAX_SCALE } from './VisualCanvas';
+
+export const ZOOM_STEP = 0.25;
+export const FIT_PADDING = 40;
+
+/**
+ * Compute a ViewportState that centers and fits all nodes inside the viewport
+ * with the given padding on each side.
+ *
+ * Returns the current viewport unchanged when there are no nodes.
+ */
+export function computeFitToView(
+	nodes: NodePosition,
+	viewportWidth: number,
+	viewportHeight: number,
+	padding = FIT_PADDING
+): ViewportState {
+	const entries = Object.values(nodes);
+	if (entries.length === 0) {
+		return { offsetX: 0, offsetY: 0, scale: 1 };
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured implementation that meets all acceptance criteria. The `computeFitToView` pure function is correct, the toolbar integrates properly with the existing canvas, and tests provide good coverage.

### What looks good

- **`computeFitToView` math is correct**: Bounding box computation, scale clamping, and centering logic are all sound. Edge cases like zero-dimension nodes (produces `Infinity` which gets clamped to `MAX_SCALE`) and negative available space (falls back to scale 1) are handled implicitly.
- **Zoom-toward-center**: The zoom in/out handlers correctly maintain the viewport center as the fixed point, consistent with the wheel zoom pattern in `applyWheelEvent`.
- **ResizeObserver lifecycle**: Properly creates, observes, sets initial size, and disconnects on unmount.
- **`pointer-events-auto`** on toolbar ensures clicks reach the buttons even if a parent has `pointer-events-none`.
- **Toolbar does not interfere with pan/zoom**: The canvas `handleMouseDown` requires spacebar to be held, so normal toolbar clicks don't trigger canvas drag.
- **Tests are thorough**: Good coverage of pure function edge cases (empty nodes, single node, multiple nodes, scale clamping both directions, asymmetric content, offset nodes) and component interactions (button rendering, click handlers, disabled states, percentage display).

### Minor observations (P3)

1. **JSDoc inaccuracy** (`CanvasToolbar.tsx`, line 19): The comment says "Returns the current viewport unchanged when there are no nodes" but the function returns `{ offsetX: 0, offsetY: 0, scale: 1 }` (the default viewport), not the current viewport -- `computeFitToView` doesn't receive the current viewport as a parameter. Consider updating the comment to say "Returns the default viewport when there are no nodes."

2. **Test count mismatch**: The header comment in `CanvasToolbar.test.tsx` lists 14 bullet points and claims "16 tests" but the file contains 15 `it` blocks. Not a functional issue but the comment is slightly out of sync.

No P0, P1, or P2 issues found. The implementation is correct, complete, and well-tested.

PR #491, opened 2026-03-20
feat: add node selection and multi-select to visual editor (Task 2.3)

lsm:
- Add WorkflowNode component with isSelected visual indicator and
  stopPropagation-based click handling
- Add WorkflowCanvas component managing selectedNodeId state, keyboard
  Delete/Backspace to remove selected node, and onNodeSelect callback
- Update VisualCanvas with onBackgroundClick prop; detect background
  clicks via target check and didDrag guard to suppress drag-end events
- Export all new types and components from visual-editor index
- 17 passing tests covering selection, deselection, and keyboard delete

### DIFF
--- a/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
@@ -58,9 +58,16 @@ interface VisualCanvasProps {
 	children?: ComponentChildren;
 	viewportState: ViewportState;
 	onViewportChange: (state: ViewportState) => void;
+	/** Called when the canvas background is clicked (not on a child node). */
+	onBackgroundClick?: () => void;
 }
 
-export function VisualCanvas({ children, viewportState, onViewportChange }: VisualCanvasProps) {
+export function VisualCanvas({
+	children,
+	viewportState,
+	onViewportChange,
+	onBackgroundClick,
+}: VisualCanvasProps) {
 	const containerRef = useRef<HTMLDivElement>(null);
 
 	// Track spacebar state for pan-drag mode
@@ -72,6 +79,8 @@ export function VisualCanvas({ children, viewportState, onViewportChange }: Visu
 		originOffsetX: number;
 		originOffsetY: number;
 	} | null>(null);
+	// Track whether a spacebar-drag actually moved the canvas (suppress background click)
+	const didDrag = useRef(false);
 
 	// Keep a ref to the latest viewport so event handlers don't stale-close over it
 	const viewportRef = useRef(viewportState);
@@ -128,6 +137,7 @@ export function VisualCanvas({ children, viewportState, onViewportChange }: Visu
 	const handleMouseDown = useCallback((e: MouseEvent) => {
 		if (!spacebarDown.current || e.button !== 0) return;
 		e.preventDefault();
+		didDrag.current = false;
 		dragState.current = {
 			startX: e.clientX,
 			startY: e.clientY,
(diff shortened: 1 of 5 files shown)


lsm (commented):

The implementation correctly addresses all five acceptance criteria from Task 2.3: single-click selects with visual indicator, background click deselects, Delete key removes selected node, selection state is accessible to parent, and tests pass. The code is clean, well-documented, and follows existing patterns in the codebase.

### Issues

**P2: Background click detection relies on `data-testid` for runtime behavior**
(`packages/web/src/components/space/visual-editor/VisualCanvas.tsx`, line 197)

The `handleContainerClick` handler uses `target.getAttribute('data-testid') === 'visual-canvas-transform'` to determine whether a click hit the canvas background. Using `data-testid` attributes for runtime logic is fragile -- some build configurations strip test IDs in production. Consider using a second ref for the transform layer, or a dedicated `data-role="canvas-background"` attribute that is clearly not test-only.

**P2: Stale `selectedNodeId` when nodes are removed externally**
(`packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx`, lines 44-48)

When the parent calls `onDeleteNode` and removes the node from the `nodes` array, `selectedNodeId` still holds the deleted node's ID. While there's no immediate visual bug (the node doesn't render), the state is stale. If a node with the same `stepId` is later added, it would appear pre-selected. Consider adding a `useEffect` that clears `selectedNodeId` when it no longer matches any node in `nodes`:

```tsx
useEffect(() => {
  if (selectedNodeId && !nodes.some(n => n.stepId === selectedNodeId)) {
    setSelectedNodeId(null);
  }
}, [nodes, selectedNodeId]);
```

**P3: No test coverage for `didDrag` suppression in VisualCanvas**

The `didDrag` guard that prevents background click after spacebar-drag is new code added in this PR but has no test coverage. Testing this would require simulating a spacebar+drag sequence followed by a click, which may be complex to set up in the unit test environment.

### Summary

Solid implementation of single-select node behavior. The code is well-structured with proper separation between VisualCanvas (generic canvas) and WorkflowCanvas (workflow-specific selection logic). The ref-based approach for keyboard handlers avoids stale closure issues. Tests are thorough and cover the key scenarios including input/textarea exclusion for Delete key. The two P2 items (test-id-based runtime logic and stale selection state) are worth addressing before this ships.

**Recommendation: REQUEST_CHANGES** (2 P2 issues)

lsm (commented):

Both P2 issues from the previous review have been properly resolved.

### P2 fix verification: `data-testid` for runtime logic

`VisualCanvas.tsx` line 197 now uses ref comparison (`target === containerRef.current || target === transformRef.current`) instead of `data-testid` attribute checks. A new `transformRef` is created at line 72 and attached to the transform div at line 218. The `data-testid="visual-canvas-transform"` attribute is still present on that div, which is correct -- it is only used by tests to locate the element for click simulation, not for runtime logic. The comment on lines 194-195 clearly documents the rationale for the ref-based approach.

### P2 fix verification: Stale `selectedNodeId` on external removal

`WorkflowCanvas.tsx` lines 59-64 add a `useEffect` that watches `[nodes, selectedNodeId]` and clears selection (plus notifies the parent via `onNodeSelectRef.current?.(null)`) when the selected node is no longer present in the array. This correctly uses the ref for the callback to avoid re-registration issues with the keyboard handler effect. The new test at lines 179-219 (`WorkflowCanvas — stale selection cleanup`) verifies this behavior end-to-end with parent-controlled state, including asserting both DOM removal and callback invocation.

### Remaining observations

**P3: No test for `didDrag` suppression** — Carried forward from previous review. The `didDrag` guard (VisualCanvas lines 83-84, 141, 156, 190-193) prevents background click from firing after a spacebar-drag. This remains untested but is low risk since it only affects a narrow UX path.

### Summary

Clean fixes. The implementation correctly addresses all acceptance criteria for Task 2.3, and the previously flagged issues are resolved. Code is well-structured with good separation of concerns between VisualCanvas and WorkflowCanvas.

PR #492, opened 2026-03-20
feat: implement WorkflowNode with drag-and-drop support (Task 2.2)

lsm:
Add WorkflowNode component to the visual canvas with full drag-and-drop
functionality. Node position updates are converted from screen-space to
canvas-space using the current viewport scale (deltaCanvas = deltaScreen / scale).

- WorkflowNode renders step name, agent name, step badge, optional START badge
- Input port (top-center, hidden on start nodes) and output port (bottom-center)
  for future connection creation
- Drag: mousedown on card body sets dragState; window mousemove/mouseup handlers
  (registered via useEffect, same pattern as VisualCanvas) compute and emit
  onPositionChange(stepId, newPosition) on every frame
- stopPropagation on port mousedown prevents drag from initiating on port clicks
- stopPropagation on card mousedown prevents VisualCanvas spacebar-pan conflict
- Visual feedback: cursor changes to grabbing and box-shadow elevates during drag
- 23 unit tests covering rendering, port events, click, and all drag scenarios
  including scale=0.5/1/2 coordinate conversion and multi-move sequences

### DIFF
--- a/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
@@ -12,13 +12,16 @@ import { useState, useEffect, useCallback, useRef } from 'preact/hooks';
 import { VisualCanvas } from './VisualCanvas';
 import { WorkflowNode } from './WorkflowNode';
 import type { WorkflowNodeProps } from './WorkflowNode';
-import type { ViewportState } from './types';
+import type { ViewportState, Point } from './types';
 
 /**
  * Per-node data passed to WorkflowCanvas.
- * All WorkflowNodeProps except `isSelected` and `onClick`, which WorkflowCanvas injects.
+ * WorkflowCanvas injects: isSelected, onClick, scale, onPositionChange.
  */
-export type WorkflowNodeData = Omit<WorkflowNodeProps, 'isSelected' | 'onClick'>;
+export type WorkflowNodeData = Omit<
+	WorkflowNodeProps,
+	'isSelected' | 'onClick' | 'scale' | 'onPositionChange'
+>;
 
 export interface WorkflowCanvasProps {
 	nodes: WorkflowNodeData[];
@@ -28,6 +31,8 @@ export interface WorkflowCanvasProps {
 	onNodeSelect?: (stepId: string | null) => void;
 	/** Called when Delete/Backspace is pressed with a node selected. */
 	onDeleteNode?: (stepId: string) => void;
+	/** Called when a node is dragged to a new position. */
+	onNodePositionChange?: (stepId: string, position: Point) => void;
 }
 
 export function WorkflowCanvas({
@@ -36,6 +41,7 @@ export function WorkflowCanvas({
 	onViewportChange,
 	onNodeSelect,
 	onDeleteNode,
+	onNodePositionChange,
 }: WorkflowCanvasProps) {
 	const [selectedNodeId, setSelectedNodeId] = useState<string | null>(null);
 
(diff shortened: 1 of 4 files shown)


lsm (commented):

Clean implementation that satisfies the acceptance criteria. The drag-and-drop mechanics are solid, the scale-aware coordinate conversion is correct, and the test coverage is thorough. A few issues worth addressing:

### P2: Click fires after drag — no drag-vs-click discrimination

**File:** `packages/web/src/components/space/visual-editor/WorkflowNode.tsx`, lines 140-147

`handleClick` unconditionally calls `onClick?.(stepId)` with no check for whether the user just completed a drag. In browsers, after a mousedown+mousemove+mouseup sequence on the same element, a `click` event still fires. This means dragging a node will also trigger selection (or whatever `onClick` does), which is likely unintended.

The standard fix is to track whether a meaningful drag occurred (e.g., `hasDragged` ref set to `true` when mousemove delta exceeds a small threshold like 3px) and then guard `handleClick`:

```ts
if (hasDraggedRef.current) {
  hasDraggedRef.current = false;
  return;
}
onClick?.(stepId);
```

This is not blocking since `onClick` is optional and the current consumer may not use it yet, but it will become a bug as soon as selection logic is wired up.

### P3: No guard against `scale === 0`

**File:** `packages/web/src/components/space/visual-editor/WorkflowNode.tsx`, line 92

`dx / scaleRef.current` would produce `Infinity` if scale were ever 0. The existing `types.ts` utility `screenToCanvas` guards with `const scale = viewport.scale || 1`, but `WorkflowNode` does not. This is unlikely given the MIN_SCALE/MAX_SCALE constants elsewhere, but a defensive `Math.max(scaleRef.current, 0.01)` or similar would be cheap insurance.

### P3: Tests don't cover the drag-then-click interaction

The test suite covers drag and click independently but never verifies what happens when a drag completes and a click fires on the same element. Adding a test that does mousedown -> mousemove -> mouseup -> click and asserts `onClick` is NOT called would document the expected behavior and catch the P2 issue above.

### Positive notes

- The ref-based approach for `scale` and `onPositionChange` correctly avoids stale closures in window-level listeners.
- Window-level listener registration (always on, guarded by `dragState.current`) mirrors the VisualCanvas pattern and handles edge cases like mouse leaving the viewport.
- Port `stopPropagation` correctly prevents drag initiation on port clicks.
- The `mousedown` `stopPropagation` correctly prevents canvas pan from triggering.
- Test for the Wrapper re-render scenario (line 316-351) properly validates that drag uses the position at drag-start time.

**Recommendation:** REQUEST_CHANGES (due to P2 — click-after-drag will cause incorrect behavior once selection is wired up)

lsm (commented):

Re-review after previous feedback was addressed. All three items from the prior review have been resolved:

1. **P2 (click-after-drag) -- Fixed.** `hasDraggedRef` correctly tracks whether a meaningful drag occurred. `handleClick` returns early when `hasDraggedRef.current` is true, then resets the flag. The flag is reset to `false` at the start of each new `handleMouseDown`, so subsequent clicks work normally. The lifecycle is correct: mousedown resets flag -> mousemove sets flag if threshold exceeded -> mouseup clears dragState but leaves flag -> click consumes and resets flag.

2. **P3 (scale=0 guard) -- Fixed.** `Math.max(scaleRef.current, 0.01)` on line 103 prevents division by zero.

3. **P3 (drag-then-click test) -- Fixed.** Four new tests added:
   - "does NOT call onClick after a drag completes" -- verifies the full mousedown->move->mouseup->click sequence
   - "calls onClick normally after a sub-threshold mousedown" -- verifies sub-threshold moves don't suppress clicks
   - "does not fire onPositionChange for moves below 3px threshold" -- verifies the threshold itself
   - "guards against scale=0" -- verifies no Infinity positions

All 27 tests pass. The drag threshold implementation is clean -- the 3px threshold in screen space is checked against both axes independently (`Math.abs(dx) < DRAG_THRESHOLD && Math.abs(dy) < DRAG_THRESHOLD`), which means a diagonal move of ~2.1px on each axis (total ~3px Euclidean) would still pass the threshold since each axis is below 3. This is a minor inconsistency (screen distance vs. per-axis check) but is a standard and acceptable approach for drag thresholds.

No new issues found. The implementation is solid and well-tested.

**Recommendation: APPROVE**

PR #495, opened 2026-03-20
fix: always set provider in createCustomAgentInit for Space custom agents

lsm:
Space custom agents (createCustomAgentInit) were returning AgentSessionInit
without a provider field, causing session.config.provider to be undefined.
query-runner.ts would then silently fall back to the Anthropic provider even
when a GLM or other non-Anthropic model was configured via SpaceAgent.model
or Space.defaultModel.

Fix: derive the provider from the resolved model ID using inferProviderForModel
and set it on both code paths (simple preset path and agent/agents pattern path).

Adds 5 new unit tests in custom-agent.test.ts covering provider inference for
Anthropic models, GLM models, agent/agents path, space.defaultModel inheritance,
and the guarantee that provider is never undefined.

### DIFF
--- a/packages/daemon/src/lib/space/agents/custom-agent.ts
+++ b/packages/daemon/src/lib/space/agents/custom-agent.ts
@@ -17,6 +17,7 @@ import type {
 	AgentDefinition,
 } from '@neokai/shared';
 import type { SpaceAgentManager } from '../managers/space-agent-manager';
+import { inferProviderForModel } from '../../providers/registry';
 
 const DEFAULT_CUSTOM_AGENT_MODEL = 'claude-sonnet-4-5-20250929';
 
@@ -308,6 +309,7 @@ export function createCustomAgentInit(config: CustomAgentConfig): AgentSessionIn
 		customAgent.tools && customAgent.tools.length > 0 ? customAgent.tools : undefined;
 
 	const model = customAgent.model ?? space.defaultModel ?? DEFAULT_CUSTOM_AGENT_MODEL;
+	const provider = inferProviderForModel(model);
 
 	const behavioralPrompt = buildCustomAgentSystemPrompt(customAgent);
 
@@ -335,6 +337,7 @@ export function createCustomAgentInit(config: CustomAgentConfig): AgentSessionIn
 			context: { spaceId: space.id },
 			type: 'worker',
 			model,
+			provider,
 			agent: agentKey,
 			agents: { [agentKey]: agentDef },
 			contextAutoQueue: false,
@@ -354,6 +357,7 @@ export function createCustomAgentInit(config: CustomAgentConfig): AgentSessionIn
 		context: { spaceId: space.id },
 		type: 'worker',
 		model,
+		provider,
 		contextAutoQueue: false,
 	};
 }
--- a/packages/daemon/tests/unit/space/custom-agent.test.ts
+++ b/packages/daemon/tests/unit/space/custom-agent.test.ts
@@ -403,6 +403,51 @@ describe('createCustomAgentInit', () => {
 		expect(init.systemPrompt?.append).toContain('Planner Agent');
 		expect(init.systemPrompt?.append).not.toContain('Review Responsibilities');
 	});
+
+	// Provider inference tests — verifies the fix for the missing provider bug
+	it('always sets provider on simple-path init (Anthropic model)', () => {
+		const config = makeConfig({
+			customAgent: makeAgent({ model: 'claude-sonnet-4-5-20250929' }),
+		});
+		const init = createCustomAgentInit(config);
+		expect(init.provider).toBe('anthropic');
+	});
+
+	it('always sets provider on simple-path init (GLM model)', () => {
+		const config = makeConfig({
+			customAgent: makeAgent({ model: 'glm-4-flash' }),
+		});
+		const init = createCustomAgentInit(config);
+		expect(init.provider).toBe('glm');
+	});
+
+	it('always sets provider on agent/agents-path init', () => {
+		const config = makeConfig({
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

Clean, well-scoped fix for a real bug where `createCustomAgentInit()` returned `AgentSessionInit` without a `provider` field, causing silent fallback to Anthropic regardless of the configured model.

### What was changed

1. **`packages/daemon/src/lib/space/agents/custom-agent.ts`** — Added `inferProviderForModel(model)` call on line 312 and passes the result as `provider` into both return branches (agent/agents path on line 340, simple path on line 360).

2. **`packages/daemon/tests/unit/space/custom-agent.test.ts`** — Added 5 test cases covering: Anthropic model inference, GLM model inference, the agent/agents code path, `space.defaultModel` inheritance, and the guarantee that provider is never undefined.

### Analysis

**Correctness**: The fix is correct. `inferProviderForModel` (in `packages/daemon/src/lib/providers/registry.ts`) uses a two-pass strategy — live registry lookup first, then static prefix-based heuristic fallback. Since unit tests run with an empty registry (per test setup), the static fallback handles the test cases properly. At runtime with the registry populated, the live lookup handles it. The `provider` field on `AgentSessionInit` is typed as `string | undefined` (line 128 of `agent-session.ts`), so adding it is type-safe.

**Completeness**: The fix covers both code paths in `createCustomAgentInit` — the `customTools` branch and the simple branch. The model resolution chain (`customAgent.model ?? space.defaultModel ?? DEFAULT_CUSTOM_AGENT_MODEL`) is correctly used as input to `inferProviderForModel`, so the provider always matches the resolved model.

**Tests**: The 5 new tests are well-chosen and cover the key scenarios. They verify both code paths, model inheritance from space defaults, and the never-undefined guarantee.

**No issues found.** This is a minimal, targeted fix that addresses the exact bug described in the task.

lsm (commented):

**Recommendation: APPROVE**

The fix is correct, well-scoped, and fully tested. Here is my analysis.

---

### What was fixed

`createCustomAgentInit()` in `packages/daemon/src/lib/space/agents/custom-agent.ts` was returning `AgentSessionInit` objects without a `provider` field on both code paths (simple preset path and agent/agents pattern path). Because `query-runner.ts` falls back to the Anthropic provider when `session.config.provider` is undefined, any Space custom agent configured with a GLM or other non-Anthropic model would silently execute against the wrong backend.

The fix derives `provider` once from the resolved model ID using `inferProviderForModel` and sets it in both return branches. This is the same pattern already used by `room-runtime.ts`, `room-handlers.ts`, and `room-runtime-service.ts`.

---

### Code correctness

**Fix is correctly placed.** `provider` is computed at line 312 from the already-resolved `model` (which has gone through the `customAgent.model ?? space.defaultModel ?? DEFAULT_CUSTOM_AGENT_MODEL` chain), so both code paths get the right provider regardless of which fallback tier produced the model ID.

**`inferProviderForModel` behavior is safe in all contexts.**
- When the daemon registry is populated (runtime), the live lookup runs first.
- When the registry is empty (unit tests), the static heuristic handles the known prefixes: `glm-*` → `'glm'`, `minimax-*` → `'minimax'`, `gpt-*` → `'anthropic-codex'`, everything else → `'anthropic'`.
- The function never returns `undefined`, so the "never produces undefined provider" test is accurate.

**No regression risk in the room-agent path.** `coder-agent.ts`, `general-agent.ts`, `leader-agent.ts`, and `planner-agent.ts` all accept `provider` via their `config` parameter and pass it through — they were already receiving an explicit provider from `room-runtime.ts`. This PR does not touch those files.

---

### Test coverage

Five new tests are added, covering:
1. Simple-path init with an Anthropic model → provider is `'anthropic'`
2. Simple-path init with a GLM model (`glm-4-flash`) → provider is `'glm'`
3. Agent/agents-path init → provider is set correctly
4. `space.defaultModel` inheritance (`glm-5-turbo`) → provider is `'glm'`
5. No-model / no-defaultModel fallback → provider is always a non-empty string

All 73 tests in `custom-agent.test.ts` pass. The test assertions are meaningful and directly verify the fix.

One minor observation: the "inherits provider from space.defaultModel" test uses the model ID `glm-5-turbo` which is not in the actual GLM model list in `glm-provider.ts` (valid IDs are `glm-5` and `glm-5-turbo` — the latter is actually present at line 58, so this is fine). The static fallback heuristic covers it via `startsWith('glm-')`.

---

### Scope review (task requirement)

The task asked to review the codebase for *all* places where a session is created with a model ID but without an explicit provider. I verified the other session-creation sites:

- `coder-agent.ts`, `general-agent.ts`, `leader-agent.ts`, `planner-agent.ts` — all receive `provider` from their caller config; callers in `room-runtime.ts` derive it via `resolveProviderForModel` before passing it in. No issue.
- `space-chat-agent.ts` — does not create `AgentSessionInit`, only builds a system prompt string. No issue.
- `SpaceRuntime` / `SpaceRuntimeService` — creates task records only; session spawning is deferred to SpaceRuntime M4 (not yet implemented). When M4 is built, it should call `resolveAgentInit` which now correctly sets the provider.

The fix covers the one real gap.

---

### P3 nit (no action required)

The test description at line 416 says "GLM model" but uses `glm-4-flash`. The actual registered GLM models in the provider are `glm-5` and `glm-5-turbo`. `glm-4-flash` is not in the live registry, so the static heuristic applies — this still correctly tests the fallback path, but a registered model like `glm-5-turbo` would be a more representative fixture. This is cosmetic and does not affect correctness.

---

**Summary:** The fix is minimal, correct, and consistent with the existing provider-inference pattern in the codebase. Tests directly verify the previously-missing behavior. No security, architectural, or correctness issues found.

PR #496, opened 2026-03-20
feat: create connections by dragging from output to input ports (Task 3.2)

lsm:
- Add useConnectionDrag hook tracking drag state (fromStepId, fromPos, currentPos, hoverTarget)
- Render dashed bezier ghost edge via VisualCanvas edgeLayer during drag
- Highlight valid input ports (green) as drop targets during drag
- Validate: block self-connections and duplicate transitions
- WorkflowNode gains isDropTarget prop + port mouseenter/leave callbacks
- WorkflowCanvas manages connection drag internally (onCreateTransition prop)
- 20 new tests covering ghost edge rendering, valid/invalid drops, highlights

### DIFF
--- a/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualCanvas.tsx
@@ -14,7 +14,7 @@
  */
 
 import { useEffect, useRef, useCallback, useState } from 'preact/hooks';
-import type { ComponentChildren } from 'preact';
+import type { ComponentChildren, RefObject } from 'preact';
 import type { NodePosition, ViewportState } from './types';
 import { CanvasToolbar } from './CanvasToolbar';
 
@@ -67,6 +67,13 @@ interface VisualCanvasProps {
 	nodes?: NodePosition;
 	/** Whether to show the zoom/fit toolbar overlay. Defaults to true. */
 	showToolbar?: boolean;
+	/**
+	 * Optional external ref to the canvas container element.
+	 * When provided, coordinate calculations (e.g. for ghost edges) use the same
+	 * element that VisualCanvas uses internally, avoiding any bounding-rect drift
+	 * if an outer wrapper gains padding/borders in future.
+	 */
+	containerRef?: RefObject<HTMLDivElement>;
 }
 
 export function VisualCanvas({
@@ -77,8 +84,12 @@ export function VisualCanvas({
 	edgeLayer,
 	nodes = {},
 	showToolbar = true,
+	containerRef: externalContainerRef,
 }: VisualCanvasProps) {
-	const containerRef = useRef<HTMLDivElement>(null);
+	const internalContainerRef = useRef<HTMLDivElement>(null);
+	// Use externally-provided ref when available so callers can do their own
+	// coordinate math against the exact same element VisualCanvas uses.
+	const containerRef = externalContainerRef ?? internalContainerRef;
 	const transformRef = useRef<HTMLDivElement>(null);
 	const [containerSize, setContainerSize] = useState({ width: 800, height: 600 });
 
--- a/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
@@ -13,7 +13,7 @@
  */
 
 import { useState, useEffect, useCallback, useRef } from 'preact/hooks';
-import type { ComponentChildren, RefObject } from 'preact';
+import type { ComponentChildren, JSX, RefObject } from 'preact';
 import { VisualCanvas } from './VisualCanvas';
 import { WorkflowNode } from './WorkflowNode';
 import type { WorkflowNodeProps, PortType } from './WorkflowNode';
@@ -57,11 +57,22 @@ export interface WorkflowCanvasProps {
 // ---- Ghost edge rendering ----
 
 /** Render a dashed bezier ghost edge from `from` to `to` in canvas-space SVG coordinates. */
-function GhostEdge({ from, to }: { from: Point; to: Point }): ComponentChildren {
-	// Control points: vertical bezier (source goes down, target comes from above)
-	const dy = Math.abs(to.y - from.y);
-	const cpOffset = Math.max(50, dy * 0.5);
-	const d = `M ${from.x} ${from.y} C ${from.x} ${from.y + cpOffset}, ${to.x} ${to.y - cpOffset}, ${to.x} ${to.y}`;
+function GhostEdge({ from, to }: { from: Point; to: Point }): JSX.Element | null {
+	// Directional bezier: for forward (downward) drags use a vertical S-curve.
+	// For backward (upward) drags, route horizontally to avoid the path looping.
+	const dx = to.x - from.x;
+	const dy = to.y - from.y;
+	let d: string;
+	if (dy >= -40) {
+		// Forward or slightly backward: standard vertical bezier
+		const cpOffset = Math.max(50, dy * 0.5);
+		d = `M ${from.x} ${from.y} C ${from.x} ${from.y + cpOffset}, ${to.x} ${to.y - cpOffset}, ${to.x} ${to.y}`;
+	} else {
+		// Backward drag: route around horizontally to avoid S-curve loops
+		const sideOffset = Math.max(60, Math.abs(dx) * 0.4 + 40);
+		const midY = (from.y + to.y) / 2;
+		d = `M ${from.x} ${from.y} C ${from.x} ${from.y + 40}, ${from.x + sideOffset} ${from.y + 40}, ${from.x + sideOffset} ${midY} S ${from.x + sideOffset} ${to.y - 40}, ${to.x} ${to.y}`;
+	}
 
 	return (
 		<>
@@ -161,11 +172,13 @@ export function WorkflowCanvas({
 
 	const handlePortMouseEnter = useCallback(
 		(stepId: string, portType: PortType) => {
-			if (portType === 'input' && dragState.active) {
+			// setHoverTarget guards on dragRef.current.active internally —
+			// no need to read dragState here, which would cause re-renders on every toggle
+			if (portType === 'input') {
 				setHoverTarget(stepId);
 			}
 		},
-		[dragState.active, setHoverTarget]
+		[setHoverTarget]
 	);
 
 	const handlePortMouseLeave = useCallback(
--- a/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
@@ -6,21 +6,36 @@
  *  - Renders WorkflowNode components with correct isSelected prop
  *  - Handles Delete/Backspace keyboard shortcut to delete selected node
  *  - Emits onNodeSelect(stepId | null) so parent components can react
+ *  - Manages connection drag via useConnectionDrag hook
+ *    - Shows ghost edge (dashed SVG path) during drag
+ *    - Highlights valid input ports as drop targets
+ *    - Creates transitions on valid drop
  */
 
 import { useState, useEffect, useCallback, useRef } from 'preact/hooks';
+import type { ComponentChildren, RefObject } from 'preact';
 import { VisualCanvas } from './VisualCanvas';
 import { WorkflowNode } from './WorkflowNode';
-import type { WorkflowNodeProps } from './WorkflowNode';
+import type { WorkflowNodeProps, PortType } from './WorkflowNode';
 import type { ViewportState, Point } from './types';
+import { useConnectionDrag } from './useConnectionDrag';
+import type { TransitionLike } from './useConnectionDrag';
 
 /**
  * Per-node data passed to WorkflowCanvas.
- * WorkflowCanvas injects: isSelected, onClick, scale, onPositionChange.
+ * WorkflowCanvas injects: isSelected, isDropTarget, onClick, scale, onPositionChange,
+ * onPortMouseDown, onPortMouseEnter, onPortMouseLeave.
  */
 export type WorkflowNodeData = Omit<
 	WorkflowNodeProps,
-	'isSelected' | 'onClick' | 'scale' | 'onPositionChange'
+	| 'isSelected'
+	| 'isDropTarget'
+	| 'onClick'
+	| 'scale'
+	| 'onPositionChange'
+	| 'onPortMouseDown'
+	| 'onPortMouseEnter'
+	| 'onPortMouseLeave'
 >;
 
 export interface WorkflowCanvasProps {
@@ -33,7 +48,50 @@ export interface WorkflowCanvasProps {
 	onDeleteNode?: (stepId: string) => void;
 	/** Called when a node is dragged to a new position. */
 	onNodePositionChange?: (stepId: string, position: Point) => void;
+	/** Existing transitions — used for duplicate detection during connection drag. */
+	transitions?: TransitionLike[];
+	/** Called when a new connection is created by dragging from an output to an input port. */
+	onCreateTransition?: (fromStepId: string, toStepId: string) => void;
 }
+
+// ---- Ghost edge rendering ----
+
+/** Render a dashed bezier ghost edge from `from` to `to` in canvas-space SVG coordinates. */
+function GhostEdge({ from, to }: { from: Point; to: Point }): ComponentChildren {
+	// Control points: vertical bezier (source goes down, target comes from above)
+	const dy = Math.abs(to.y - from.y);
+	const cpOffset = Math.max(50, dy * 0.5);
+	const d = `M ${from.x} ${from.y} C ${from.x} ${from.y + cpOffset}, ${to.x} ${to.y - cpOffset}, ${to.x} ${to.y}`;
+
+	return (
+		<>
+			{/* Shadow for contrast */}
+			<path
+				d={d}
+				fill="none"
+				stroke="rgba(0,0,0,0.3)"
+				strokeWidth={5}
+				strokeDasharray="8 4"
+				strokeLinecap="round"
+			/>
+			{/* Visible ghost stroke */}
+			<path
+				data-testid="ghost-edge"
+				d={d}
+				fill="none"
+				stroke="#60a5fa"
+				strokeWidth={2.5}
+				strokeDasharray="8 4"
+				strokeLinecap="round"
+				opacity={0.9}
+			/>
+		</>
+	);
+}
+
+// ============================================================================
+// WorkflowCanvas
+// ============================================================================
 
 export function WorkflowCanvas({
 	nodes,
@@ -42,6 +100,8 @@ export function WorkflowCanvas({
 	onNodeSelect,
 	onDeleteNode,
 	onNodePositionChange,
+	transitions = [],
+	onCreateTransition,
 }: WorkflowCanvasProps) {
 	const [selectedNodeId, setSelectedNodeId] = useState<string | null>(null);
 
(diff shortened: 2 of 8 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR implements connection creation by dragging from output ports to input ports, with a ghost edge following the cursor during the drag. The implementation is clean, well-structured, and follows established patterns in the codebase.

### What works well

- **Hook design**: `useConnectionDrag` cleanly separates drag state management from rendering. The use of refs to avoid stale closures in window listeners is correct and mirrors the existing pattern in `WorkflowNode` and `VisualCanvas`.
- **Coordinate conversion**: Screen-to-canvas conversion is properly applied so the ghost edge renders correctly regardless of viewport pan/zoom state. The SVG edge layer sits inside the transform div, so canvas-space coordinates are correct.
- **Validation**: Self-connection blocking and duplicate transition blocking are both implemented and tested.
- **Test coverage**: 20 tests covering the hook's idle state, drag activation, mousemove tracking, cancel/commit flows, self-connection blocking, duplicate blocking, hover target management, and WorkflowCanvas integration (ghost edge rendering, drop target highlighting).
- **Event handling**: `stopPropagation` on port mousedown correctly prevents card drag from starting when clicking on ports.

### Minor observations (P3)

1. **`handlePortMouseLeave` doesn't guard on `dragState.active`** (`WorkflowCanvas.tsx:172-174`): Unlike `handlePortMouseEnter` which checks `dragState.active`, the leave handler unconditionally calls `setHoverTarget(null)`. This is functionally safe because `setHoverTarget` itself guards on `dragRef.current.active` (line 146 of `useConnectionDrag.ts`), but the inconsistency is worth noting.

2. **Drop target highlight shown on nodes with existing connections from source** (`WorkflowCanvas.tsx:220-221`): `isDropTarget` doesn't check whether a transition already exists from the source to the target node. This means nodes that would be rejected as duplicates on drop still show the green highlight during drag. This is a minor UX inconsistency — the drop is correctly rejected, but the visual feedback is slightly misleading. Not blocking since it's cosmetic only.

3. **`GhostEdge` component typed as returning `ComponentChildren`** (`WorkflowCanvas.tsx:60`): While this works, returning `ComponentChildren` from a functional component is unconventional — typically the return type would be `JSX.Element | null`. Not blocking.

All CI checks that have completed are passing. The implementation is correct, complete, and well-tested.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P2 issues that should be addressed before merge)

---

## Summary

The implementation is architecturally sound and covers the core happy-path correctly. The hook pattern with dual state + mutable refs to avoid stale closures is the right approach. The SVG ghost edge is placed correctly inside the transform layer so canvas-space coordinates render accurately. The 20 unit tests are thorough for the hook's internal logic.

That said, there are two P2 issues worth fixing before merge, plus a P2 behavioral bug and a couple of P3 nits.

---

## P2 — Clicking an output port silently selects the node

**File:** `packages/web/src/components/space/visual-editor/WorkflowNode.tsx`, lines 183–189

`handleOutputPortMouseDown` calls `e.stopPropagation()` which correctly prevents the card's drag from starting. However, `stopPropagation` only stops the *mousedown* event. The subsequent *click* event (mousedown + mouseup on same element) still bubbles up to the card and fires `handleClick`, which calls `onClick?.(stepId)`, selecting the node.

This means every output port click changes the selected node, which is likely unintended. When a user clicks an output port to initiate a drag, the node becomes selected, which is distracting visual feedback. The fix is to also call `e.stopPropagation()` in the port's `onClick` handler, or to set a flag on port mousedown and check it in `handleClick`.

---

## P2 — `handlePortMouseEnter` triggers unnecessary re-renders for all nodes

**File:** `packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx`, lines 162–169

```ts
const handlePortMouseEnter = useCallback(
    (stepId: string, portType: PortType) => {
        if (portType === 'input' && dragState.active) {  // ← reads dragState from closure
            setHoverTarget(stepId);
        }
    },
    [dragState.active, setHoverTarget],  // ← recreates callback on every active toggle
);
```

The `dragState.active` dependency means a new `handlePortMouseEnter` reference is created every time a drag starts or ends. Since this callback is passed as a prop to every `WorkflowNode`, all nodes re-render on each active toggle. The guard inside `setHoverTarget` (`if (!dragRef.current.active) return`) already handles this correctly using the mutable ref — the check at the `handlePortMouseEnter` level is redundant and causes churn.

The `handlePortMouseLeave` callback correctly omits this guard (trusting `setHoverTarget`'s inner guard) — `handlePortMouseEnter` should match that pattern.

---

## P2 — No test for end-to-end connection creation at the `WorkflowCanvas` integration level

**File:** `packages/web/src/components/space/visual-editor/__tests__/useConnectionDrag.test.tsx`

The `WorkflowCanvas` integration test section (lines 366–487) verifies ghost edge rendering and highlight behavior but there is no test that fires the full sequence:
1. `mouseDown` on an output port
2. `mouseEnter` on a non-source input port (to set `hoverTargetStepId`)
3. `mouseUp` on window

...and asserts `onCreateTransition` is called with the correct step IDs. The hook-level tests cover this, but any regression in the wiring between `WorkflowCanvas`, `WorkflowNode`, and the hook would not be caught by a canvas integration test.

---

## P2 — No `Escape` key handler to cancel an in-progress drag

**File:** `packages/web/src/components/space/visual-editor/useConnectionDrag.ts`

There is no `keydown` listener for the `Escape` key. If a user starts dragging but wants to abort (e.g., wrong source node), the only way to cancel is to release the mouse button. Pressing Escape is the standard drag-cancel interaction in most applications. Without it, users who accidentally start a drag have no keyboard escape path.

This is a UX gap rather than a correctness bug, but it is standard drag interaction behavior.

---

## P3 — `GhostEdge` bezier curves look incorrect for upward drags

**File:** `packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx`, lines 62–64

```ts
const cpOffset = Math.max(50, dy * 0.5);
const d = `M ${from.x} ${from.y} C ${from.x} ${from.y + cpOffset}, ${to.x} ${to.y - cpOffset}, ...`;
```

`dy = Math.abs(to.y - from.y)` is always positive, so `cpOffset` is always positive. When dragging upward (`to.y < from.y`), the source control point pushes down (`from.y + cpOffset`) while the target control point also pushes up (`to.y - cpOffset` is more negative), causing the bezier to loop or cross itself. The visual artifact is most visible when dragging backward in a top-to-bottom layout. Consider using a signed offset or a directional bezier.

---

## P3 — `containerRef` in `WorkflowCanvas` points to an extra outer wrapper div

**File:** `packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx`, line 210

The `containerRef` used for coordinate conversion is attached to a wrapper `div` added by `WorkflowCanvas`, while the actual VisualCanvas container (which receives mouse events) is the inner div inside `VisualCanvas`. In the current implementation both elements have the same bounding rect (no padding/margin on the wrapper), but if the wrapper div ever gains spacing, the ghost edge will be offset. Consider attaching `containerRef` directly to the VisualCanvas container div — though since `VisualCanvas` manages its own `containerRef` internally, this may require a design change or a `forwardRef`.

---

## What's done well

- The dual-state + mutable-ref pattern (`dragRef` + `dragState`) correctly solves the stale closure problem for window listeners without over-engineering.
- `startDrag` converts port screen coordinates to canvas-space at drag-start time, ensuring the ghost edge anchors correctly even after pan/zoom during drag.
- The `transitions` prop + `transitionsRef` pattern prevents duplicate validation from closing over stale data.
- All three validation rules (self-connection, duplicate, inactive-drag guard) are tested in the hook unit tests.
- The `IDLE` constant reuse is a nice touch — avoids creating new objects on every reset.

lsm (commented):

**Recommendation: APPROVE** — All three previously-flagged P2 fixes are correct. One residual P2 concern remains (coordinate system bug in the ghost edge), plus a few P3 nits.

---

### P2 Fixes Verified

**Fix 1 — Click event bubbling past stopPropagation**
Both port divs now have `onClick={stopClickPropagation}` in `WorkflowNode.tsx` (lines 255, 310). The `stopClickPropagation` callback correctly stops propagation without preventing default. This correctly prevents port clicks from reaching the card's `handleClick` and triggering node selection.

**Fix 2 — Redundant `dragState.active` dependency in `handlePortMouseEnter`**
`WorkflowCanvas.tsx` line 173-181: `handlePortMouseEnter` now depends only on `[setHoverTarget]`. The internal guard is correctly delegated to `setHoverTarget` inside the hook (`if (!dragRef.current.active) return`). This eliminates the unnecessary re-render churn.

**Fix 3 — Integration tests for `onCreateTransition` in `WorkflowCanvas`**
Three new integration tests added in `useConnectionDrag.test.tsx` lines 490-552:
- Full drag-connect sequence (`calls onCreateTransition when dropping on a valid input port`)
- Cancel on empty space (`does not call onCreateTransition when releasing over empty space`)
- Escape key cancel (`cancels drag when Escape is pressed`)
All three tests are correctly structured and verify the right behavior.

**Bonus: Escape key handler**
Correctly added in `useConnectionDrag.ts` lines 195-200. Uses `dragRef.current.active` (not stale React state), resets both the ref and state to `IDLE`.

---

### Remaining Issues

#### P2 — Ghost edge coordinate system: containerRef points to the outer wrapper, not the VisualCanvas container

In `WorkflowCanvas.tsx` lines 130 and 223:
```tsx
const containerRef = useRef<HTMLDivElement>(null);
// ...
<div ref={containerRef} style={{ width: '100%', height: '100%', position: 'relative' }}>
  <VisualCanvas ...>
```

The `containerRef` is attached to a plain `<div>` wrapper that wraps `VisualCanvas`. The coordinate math in `useConnectionDrag.ts` (lines 113, 157-160) subtracts `containerRef.current.getBoundingClientRect()` to get container-relative coordinates before passing them to `screenToCanvas`. Meanwhile, `VisualCanvas` uses its own internal `containerRef` for wheel and drag events — that container's bounding rect is what Preact nodes are positioned relative to (since all node `position: absolute` placements land inside `VisualCanvas`'s transform div).

In practice, both `div` elements share `width: 100%` and `height: 100%` with no additional margins or padding, so the bounding rects will be identical at runtime. This works today but is fragile: if anyone adds padding or border to the outer wrapper (e.g., for a toolbar gutter), the ghost edge anchor point will silently drift off. The ref should be forwarded from `VisualCanvas` or the outer wrapper should be proven to be identical in position to the inner container.

This is a latent correctness issue rather than a current bug, hence P2.

#### P3 — `isDropTarget` is set for all non-source, non-start nodes regardless of validity

In `WorkflowCanvas.tsx` line 233:
```tsx
const isDropTarget = dragState.active && dragState.fromStepId !== stepId && !node.isStartNode;
```

This highlights every eligible node green during any drag, even nodes that already have a transition from the source. The hook correctly blocks duplicate transitions on `mouseup`, but the visual feedback implies all green-highlighted nodes are valid targets, which is misleading when duplicates exist. This is a UX nit rather than a functional bug.

#### P3 — `GhostEdge` backward-drag path uses an implicit smooth cubic (`S`) without a prior control point

In `WorkflowCanvas.tsx` line 74:
```
d = `M ${from.x} ${from.y} C ..., ${from.x + sideOffset} ${midY} S ${from.x + sideOffset} ${to.y - 40}, ${to.x} ${to.y}`;
```

The `S` (smooth cubic bezier) command mirrors the control point of the preceding `C` segment. This produces a smooth continuation only when the path curves in the same direction. In some backward-drag geometries the reflected control point may land in an unexpected position. No functional breakage, just a potential visual artifact for extreme upward drags.

#### P3 — `HookHarness` in tests calls `startDrag` via a button `onClick`, but `startDrag` calls `e.preventDefault()`/`e.stopPropagation()`

In `useConnectionDrag.test.tsx` line 141, the test harness wires `startDrag` to a button click, passing the synthetic event as a `MouseEvent`. The `startDrag` implementation calls `e.stopPropagation()` and `e.preventDefault()`, which are valid operations on synthetic events. This works in happy-dom but the test is exercising a slightly different path than the real output-port `mousedown` scenario. Not a correctness issue for the tests themselves.

---

### Summary

All three P2 issues from the prior review are correctly addressed. The bonus Escape key handler and improved backward bezier are solid additions. The remaining P2 item (containerRef aliasing the outer wrapper rather than the VisualCanvas container) is a latent coordinate precision concern. Three P3 nits are noted above but none block shipping.

lsm (commented):

This is a follow-up review after the worker addressed P2 issues from a prior review. All five requested fixes have been correctly implemented.

### Fixes verified

1. **Port clicks no longer select the node** -- `stopClickPropagation` added to both input and output port `onClick` handlers in `WorkflowNode.tsx` (lines 192-194, 255, 310). Correct.

2. **Removed `dragState.active` from `handlePortMouseEnter` dependency array** -- The guard is handled internally by `setHoverTarget` via `dragRef.current.active`, so reading `dragState.active` in the callback was unnecessary and caused re-renders. Now correctly depends only on `[setHoverTarget]` (`WorkflowCanvas.tsx` line 180).

3. **Escape key handler** -- Added `onKeyDown` listener in `useConnectionDrag.ts` (lines 195-199) that resets drag to IDLE when Escape is pressed during an active drag. Properly registered and cleaned up alongside mousemove/mouseup listeners.

4. **End-to-end integration tests** -- Three new tests in `useConnectionDrag.test.tsx` (lines 489-552):
   - Full drag-to-drop sequence: mouseDown on output port -> mouseEnter on input port -> mouseUp -> verifies `onCreateTransition` called
   - Cancel on empty space: mouseDown -> mouseUp without hovering any port -> verifies no transition created
   - Escape key cancellation: mouseDown -> mousemove (ghost edge visible) -> Escape -> verifies ghost edge removed and no transition

5. **Ghost edge bezier improved for upward drags** -- `GhostEdge` now uses directional routing: forward/slightly-backward drags (dy >= -40) use a standard vertical bezier, while backward drags route horizontally to avoid S-curve loops (`WorkflowCanvas.tsx` lines 60-75).

### Observations

**P3 (nit):** The backward bezier routing in `GhostEdge` always offsets to the right (`from.x + sideOffset`) regardless of the horizontal relationship between source and target. When the target is far to the right of the source, the curve routes even further right before looping back, which could look visually awkward. A more polished approach would pick the offset direction based on the sign of `dx`. This is purely cosmetic and not blocking.

### Overall assessment

The implementation is clean, well-structured, and correctly addresses all the feedback from the prior review. The hook architecture with `dragRef` for synchronous access alongside `useState` for re-renders is a sound pattern. Test coverage is comprehensive with 20+ tests covering the hook unit behavior, canvas integration, and end-to-end connection workflows.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The P2 containerRef bounding-rect drift issue has been correctly resolved. The fix is clean and minimal.

### What changed in the final commit

1. **`VisualCanvas.tsx`** now accepts an optional `containerRef?: RefObject<HTMLDivElement>` prop. When provided, it uses that ref instead of its internal one for the container `<div>`. When omitted, behavior is unchanged (backward compatible).

2. **`WorkflowCanvas.tsx`** passes its `containerRef` directly to `VisualCanvas` and removes the outer wrapper `<div>`. This ensures `useConnectionDrag`'s `getBoundingClientRect()` calls measure the exact same DOM element that `VisualCanvas` uses for wheel/drag event handling — eliminating any potential coordinate drift from nested wrappers.

### Verification

- The `containerRef` flow is correct: `WorkflowCanvas` creates the ref, passes it to both `useConnectionDrag` and `VisualCanvas`. `VisualCanvas` attaches it to its root `<div>`, so both coordinate systems are grounded on the same element.
- The `externalContainerRef ?? internalContainerRef` fallback in `VisualCanvas` (line 92) preserves backward compatibility for any callers that don't pass `containerRef`.
- No extra DOM layer means no risk of padding/border offset between `useConnectionDrag`'s container and `VisualCanvas`'s container.

### Code quality

- **Hook design**: `useConnectionDrag` is well-structured with refs for stale closure prevention, proper cleanup of window listeners, and clear validation (self-connection blocked, duplicate blocked, Escape cancellation).
- **Ghost edge rendering**: `GhostEdge` handles both forward and backward drags with appropriate bezier curves.
- **Test coverage**: 552 lines of tests covering idle state, drag activation, mousemove tracking, cancel on mouseup without target, successful connection creation, self-connection blocking, duplicate blocking, Escape cancellation, and WorkflowCanvas integration (ghost edge rendering, drop target highlighting, end-to-end connection creation). Coverage is thorough.
- **Port event plumbing**: Clean separation — `WorkflowNode` emits port events, `WorkflowCanvas` routes them to the hook, and the hook manages all drag state.

No remaining issues found.

lsm (commented):

**Recommendation: APPROVE**

This is the third review iteration. The previously flagged P2 — coordinate drift from an extra wrapper `<div>` in WorkflowCanvas — has been correctly fixed.

---

### What changed

**`VisualCanvas.tsx`** now accepts an optional `containerRef?: RefObject<HTMLDivElement>` prop. Internally the component resolves which ref to use via:

```ts
const internalContainerRef = useRef<HTMLDivElement>(null);
const containerRef = externalContainerRef ?? internalContainerRef;
```

The resolved ref is then used for the `ref={containerRef}` on the single container `<div>`, for the `ResizeObserver`, for wheel-event cursor offset, and for spacebar-drag cursor styling — every internal `containerRef.current` usage now points to exactly one DOM element, whichever ref was provided.

**`WorkflowCanvas.tsx`** drops the extra outer wrapper `<div>` that existed in the previous iteration. It creates `containerRef = useRef<HTMLDivElement>(null)` and passes it directly to `<VisualCanvas containerRef={containerRef} ...`. The `useConnectionDrag` hook receives the same ref (widened to `RefObject<HTMLElement>`, which is a safe covariant cast). So `getBoundingClientRect()` in `startDrag` and in the `mousemove` handler now measures the identical DOM node that `VisualCanvas` uses for its own wheel and pan events.

---

### Correctness of the fix

The original problem was that `useConnectionDrag` measured a parent wrapper element while `VisualCanvas` events were measured against its own container — so any CSS border or padding on the wrapper would cause coordinate drift. The fix eliminates the wrapper entirely and shares a single ref, which is the right solution.

One subtle behaviour to confirm: when `externalContainerRef` is provided, `internalContainerRef` is never attached to the DOM (it stays `null`). There is no double-attachment or ref aliasing issue. The `??` fallback is purely for callers who do not need to share the ref (i.e. standalone use of `VisualCanvas`).

---

### Test coverage assessment

The new `containerRef` prop path in `VisualCanvas` has no dedicated unit test — there is no test that passes an external ref and asserts the same element is used for bounding-rect calculations. However, the WorkflowCanvas integration tests in `useConnectionDrag.test.tsx` render `WorkflowCanvas`, which exercises the full chain: `WorkflowCanvas` creates the ref → passes it to `VisualCanvas` → `useConnectionDrag` uses it for coordinate math. The ghost-edge appearance tests and the end-to-end connection creation tests exercise this path implicitly.

A targeted unit test for the prop (e.g. render `VisualCanvas` with an external ref, fire a wheel event, assert `getBoundingClientRect` on the external ref's element is called) would add clarity but is not blocking. P3 nit.

---

### No remaining issues found

- No P0/P1/P2 issues identified.
- Type safety: the widening cast `containerRef as RefObject<HTMLElement>` when calling `useConnectionDrag` is safe (`HTMLDivElement extends HTMLElement`).
- The `internalContainerRef` fallback keeps `VisualCanvas` usable standalone without any external ref — backward compatible.
- All logic paths (self-connection guard, duplicate-transition guard, Escape cancel, isDropTarget exclusion for start nodes) remain intact from the previous review.

**P3 nit:** No unit test exercises the `containerRef` prop path of `VisualCanvas` in isolation (`VisualCanvas.test.tsx` does not reference `containerRef` at all). Integration coverage exists via `WorkflowCanvas`, so this is a minor documentation gap rather than a correctness risk.

PR #498, opened 2026-03-20
feat: implement EdgeConfigPanel component for transition editing (Task 4.2)

lsm:
Adds EdgeConfigPanel to the visual editor for configuring workflow transition
(edge) properties when selected. Panel shows read-only from/to step names,
a condition type selector (always/human/condition), an expression input that
appears only for the 'condition' type, and a delete transition button.

15 unit tests cover all acceptance criteria: from/to name display, condition
type selector, expression input visibility per type, onUpdateCondition/onDelete/
onClose callback behaviour, and expression preservation on type switch.

### DIFF
--- a/packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx
+++ b/packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx
@@ -6,24 +6,25 @@
  */
 
 import { useCallback } from 'preact/hooks';
-import type { WorkflowConditionType } from '@neokai/shared';
+import type { WorkflowCondition, WorkflowConditionType } from '@neokai/shared';
 
 // ============================================================================
 // Types
 // ============================================================================
 
 export interface EdgeTransition {
-	/** Unique identifier for the transition */
+	/**
+	 * Unique identifier for the transition — used as a stable key forwarded to
+	 * onUpdateCondition and onDelete callbacks so the parent can locate the edge.
+	 */
 	id: string;
 	/** Source step name (human-readable, read-only) */
 	fromStepName: string;
 	/** Target step name (human-readable, read-only) */
 	toStepName: string;
-	/** Condition guarding this transition */
-	condition: {
-		type: WorkflowConditionType;
-		expression?: string;
-	};
+	/** Condition guarding this transition. Extra fields (description, maxRetries, etc.)
+	 *  are passed through unchanged; the panel only edits `type` and `expression`. */
+	condition: WorkflowCondition;
 }
 
 export interface EdgeConfigPanelProps {
@@ -47,6 +48,9 @@ const CONDITION_LABELS: Record<WorkflowConditionType, string> = {
 	condition: 'Expression',
 };
 
+/** Explicit ordering for the condition type <select> options. */
+const CONDITION_TYPE_ORDER: WorkflowConditionType[] = ['always', 'human', 'condition'];
+
 export function EdgeConfigPanel({
 	transition,
 	onUpdateCondition,
@@ -58,7 +62,10 @@ export function EdgeConfigPanel({
 	const handleTypeChange = useCallback(
 		(e: Event) => {
 			const type = (e.target as HTMLSelectElement).value as WorkflowConditionType;
-			// When switching away from 'condition', clear the expression
+			// When switching away from 'condition', clear the expression so callers
+			// don't persist a stale expression string under a non-expression condition.
+			// Note: the parent is responsible for preserving the expression across
+			// two-way type switches if that behaviour is desired.
 			onUpdateCondition(id, type, type === 'condition' ? condition.expression : undefined);
 		},
 		[id, condition.expression, onUpdateCondition]
@@ -90,7 +97,7 @@ export function EdgeConfigPanel({
 					onClick={onClose}
 					aria-label="Close"
 				>
-					✕
+					×
 				</button>
 			</div>
 
@@ -128,7 +135,7 @@ export function EdgeConfigPanel({
 					value={condition.type}
 					onChange={handleTypeChange}
 				>
-					{(Object.keys(CONDITION_LABELS) as WorkflowConditionType[]).map((type) => (
+					{CONDITION_TYPE_ORDER.map((type) => (
 						<option key={type} value={type}>
 							{CONDITION_LABELS[type]}
 						</option>
--- a/packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx
+++ b/packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx
@@ -1,0 +1,167 @@
+/**
+ * EdgeConfigPanel
+ *
+ * Shown when a workflow transition (edge) is selected in the visual editor.
+ * Allows editing the condition type and expression, and deleting the transition.
+ */
+
+import { useCallback } from 'preact/hooks';
+import type { WorkflowConditionType } from '@neokai/shared';
+
+// ============================================================================
+// Types
+// ============================================================================
+
+export interface EdgeTransition {
+	/** Unique identifier for the transition */
+	id: string;
+	/** Source step name (human-readable, read-only) */
+	fromStepName: string;
+	/** Target step name (human-readable, read-only) */
+	toStepName: string;
+	/** Condition guarding this transition */
+	condition: {
+		type: WorkflowConditionType;
+		expression?: string;
+	};
+}
+
+export interface EdgeConfigPanelProps {
+	transition: EdgeTransition;
+	onUpdateCondition: (
+		transitionId: string,
+		conditionType: WorkflowConditionType,
+		expression?: string
+	) => void;
+	onDelete: (transitionId: string) => void;
+	onClose: () => void;
+}
+
+// ============================================================================
+// Component
+// ============================================================================
+
+const CONDITION_LABELS: Record<WorkflowConditionType, string> = {
+	always: 'Always',
+	human: 'Human approval',
+	condition: 'Expression',
+};
+
+export function EdgeConfigPanel({
+	transition,
+	onUpdateCondition,
+	onDelete,
+	onClose,
+}: EdgeConfigPanelProps) {
+	const { id, fromStepName, toStepName, condition } = transition;
+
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Summary

Clean, well-structured implementation that meets the task requirements. The component correctly handles condition type switching, expression editing, and all callback wiring. Tests are thorough with 15 cases covering all acceptance criteria. Typecheck and all tests pass.

### Issues

**P2: `EdgeTransition.condition` duplicates `WorkflowCondition` from shared**

`packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx` lines 24-28 define a new inline type `{ type: WorkflowConditionType; expression?: string }` that is a subset of the existing `WorkflowCondition` interface from `@neokai/shared` (which also has `description`, `maxRetries`, `timeout`). The panel should reuse `WorkflowCondition` directly — the extra fields are simply ignored by the component, and using the shared type avoids drift if new fields are added or the shape changes. If the intent is to restrict the shape, use `Pick<WorkflowCondition, 'type' | 'expression'>`.

**P2: `handleExpressionChange` uses `onInput` but test uses `fireEvent.input`**

`packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx` line 141 uses `onInput` for the expression field. This works but is inconsistent with the `onChange` used on the select (line 125). The test at line 173 uses `fireEvent.input(...)` which happens to work. This is not a bug, but the mismatch between `onInput` vs `onChange` within the same component is worth noting — Preact treats these differently. `onInput` fires on every keystroke while `onChange` fires on commit. The choice of `onInput` is actually the better UX for a text field (live updates), but the inconsistency should be documented or at least deliberate.

**P3: Close button uses Unicode character instead of an SVG icon**

`packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx` line 91 uses `✕` (Unicode multiplication sign). This may render inconsistently across platforms and fonts. Other components in the codebase may use SVG icons for close buttons. Minor cosmetic concern.

### Recommendation: APPROVE

The two P2 issues are meaningful suggestions but not blocking. The type duplication is the most notable concern — it creates a maintenance burden if `WorkflowCondition` evolves. Overall this is a solid, well-tested component that correctly fulfills the task requirements.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P0 issue (all 15 tests fail at runtime) and two P1 issues.

---

### P0 — All 15 tests fail at runtime: `document is not defined`

Running the test file results in all 15 tests failing with `ReferenceError: document is not defined`:

```
ReferenceError: document is not defined
  ❯ render node_modules/.bun/@testing-library+preact/dist/esm/pure.mjs:33:5
  ❯ packages/web/src/components/space/visual-editor/__tests__/EdgeConfigPanel.test.tsx:59:27
```

The vitest config at `packages/web/vitest.config.ts` specifies `environment: 'happy-dom'`, but running the test file from the project root via `bunx vitest run` does not pick up the package-level config. The same failure affects the previously-merged WorkflowCanvas and WorkflowNode tests, so this is a pre-existing infrastructure issue — but it still means the tests shipped in this PR do not currently pass.

The tests are functionally correct (the logic, assertions, and fixture factories look right). The fix needed is either running tests from `packages/web/` (`cd packages/web && bunx vitest run ...`) or adding a per-file `// @vitest-environment happy-dom` annotation at the top of the test file. Given that every other component test in this directory has the same unresolved problem, the team should address this globally in a follow-up, but the PR should acknowledge it.

---

### P1 — `EdgeTransition.condition` re-declares a narrowed subset of `WorkflowCondition`

`packages/web/src/components/space/visual-editor/EdgeConfigPanel.tsx` lines 22–26 define:

```ts
condition: {
  type: WorkflowConditionType;
  expression?: string;
};
```

The canonical `WorkflowCondition` interface in `@neokai/shared` has three additional optional fields (`description`, `maxRetries`, `timeoutMs`). By using a hand-rolled inline type, the component silently drops those fields. A future caller that constructs an `EdgeTransition` from a real `WorkflowTransition.condition` will have TypeScript accept the assignment (structural subtype), but any round-trip back to the domain type via `onUpdateCondition` will lose those fields permanently.

**Fix:** use `Pick<WorkflowCondition, 'type' | 'expression'>` or simply `WorkflowCondition` for the `condition` field. Both are exported from `@neokai/shared`.

---

### P1 — `handleTypeChange` preserves a stale expression when switching *to* `condition` from a type that never sets one

`EdgeConfigPanel.tsx` lines 58–65:

```ts
onUpdateCondition(id, type, type === 'condition' ? condition.expression : undefined);
```

If the user starts on `always` (no expression), switches to `condition`, the expression passed to `onUpdateCondition` is `undefined` — which is correct. If the user then types an expression, switches to `always` (expression cleared), then switches back to `condition`, the stale `condition.expression` that was present *before the last clear* is gone from the parent's state. On the next re-render, `condition.expression` will be `undefined`, so the preserved value will be `undefined` again. This is acceptable behaviour, but test "changing condition type to condition preserves existing expression" succeeds only because `expression: 'my-expr'` is baked into the fixture while the *parent* is responsible for state — the component itself cannot truly preserve an expression across a two-way type switch. This is a design gap worth documenting with a comment, not a bug in the component itself, but callers need to be aware they must persist the expression themselves.

---

### P2 — `CONDITION_LABELS` ordering is determined by object insertion order

`EdgeConfigPanel.tsx` line 131 uses `Object.keys(CONDITION_LABELS)` to render the `<select>` options. In V8 this reliably iterates in insertion order (`always → human → condition`), but this is an implicit dependency on a language implementation detail. Using an explicit ordered array constant (`const CONDITION_TYPE_ORDER: WorkflowConditionType[] = ['always', 'human', 'condition']`) would be clearer and easier to extend.

---

### P3 — `id` in `EdgeTransition` is unused beyond callback forwarding

`id` is passed into every callback but never rendered in the panel (which is appropriate — it is an implementation detail). A brief JSDoc comment on `EdgeTransition.id` noting it is used as a stable key for callbacks would clarify intent to future maintainers. The current comment says "Unique identifier" which is accurate but does not explain *why* the panel needs it at all.

---

### Positive observations

- The component is cleanly typed, fully relies on the canonical `WorkflowConditionType` from `@neokai/shared`, and avoids runtime string comparisons outside of the conditional render gate.
- All callbacks are wrapped in `useCallback` with correct dependency arrays.
- `data-testid` attributes are present on every interactive element, consistent with the testing approach in the rest of the visual-editor directory.
- Biome formatting, Oxlint, TypeScript strict mode, and Knip all pass cleanly.
- The barrel export in `index.ts` correctly exports both the component and both types.
- Test fixture factories (`makeTransition`, `makeProps`) follow the same pattern used in sibling test files.

lsm (commented):

### Re-review: All previous issues addressed

Verified each item from the prior review:

1. **`// @vitest-environment happy-dom`** -- Present on line 1 of the test file. All 15 tests pass.
2. **`WorkflowCondition` from `@neokai/shared`** -- `EdgeTransition.condition` is typed as `WorkflowCondition` (imported from `@neokai/shared`), which includes `description`, `maxRetries`, and `timeoutMs` fields. No field-drop risk.
3. **`CONDITION_TYPE_ORDER` constant** -- Explicit array on line 52 controls `<select>` ordering instead of `Object.keys()`.
4. **`handleTypeChange` comment** -- Clear comment on lines 65-68 explains expression-clear design and parent responsibility.
5. **`×` character** -- Matches codebase convention (verified in `WorkflowEditor.tsx`).

### Overall assessment

The implementation is clean, well-structured, and complete for the task requirements:

- Component correctly renders read-only from/to labels, a condition type selector, a conditional expression input, delete and close buttons.
- `onUpdateCondition` callback API is straightforward; the panel only edits `type` and `expression`, passing other `WorkflowCondition` fields through unchanged (documented in the JSDoc on line 25-26).
- Tests cover all interactive paths: type switching in both directions, expression editing, delete, close.
- Barrel export in `index.ts` is correct.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** — All previously-raised issues are addressed; no new concerns found.

---

**P0 fixed: `// @vitest-environment happy-dom` added.**
Line 1 of `packages/web/src/components/space/visual-editor/__tests__/EdgeConfigPanel.test.tsx` now has the directive. All 15 tests pass when run in isolation (`bunx vitest run src/components/space/visual-editor/__tests__/EdgeConfigPanel.test.tsx`). The 4 failing test files in the full suite are pre-existing and unrelated to this PR.

Note: the directive is technically redundant because `packages/web/vitest.config.ts` already sets `environment: 'happy-dom'` globally. However it is harmless and serves as useful documentation. The P0 failure described in the previous review was a tooling invocation issue (running from the repo root rather than `packages/web/`), not an actual configuration gap.

**P1/P2 fixed: `WorkflowCondition` from `@neokai/shared` used.**
`EdgeTransition.condition` is now typed as `WorkflowCondition` (line 27 of `EdgeConfigPanel.tsx`), which carries all four fields (`type`, `expression`, `description`, `maxRetries`, `timeoutMs`). No field-drop risk.

**P2 fixed: `CONDITION_TYPE_ORDER` constant introduced.**
Line 52 defines an explicit `WorkflowConditionType[]` array that controls the `<select>` option order. `Object.keys(CONDITION_LABELS)` is no longer used.

**P1 doc fixed: comment on `handleTypeChange` clarifies expression-clear design.**
Lines 65–68 now document the intent clearly and note that expression preservation across two-way type switches is the parent's responsibility.

**P3 fixed: `×` (U+00D7) replaces `✕`.**
Character at line 100 matches the convention used in the rest of the codebase.

---

### Final Assessment

The component is clean and complete:

- Correct use of `WorkflowCondition` from `@neokai/shared` ensures the component accepts the full shared type without silent drops.
- `handleTypeChange` correctly passes `condition.expression` when switching to `condition` type (preserving any expression stored on the data regardless of current UI visibility) and passes `undefined` when switching away — consistent with the documented design.
- `onInput` on the expression field is intentional (live update on each keystroke) and the design comment is sufficient.
- Barrel export in `packages/web/src/components/space/visual-editor/index.ts` is correct (both the value and the types exported).
- Typecheck and lint pass cleanly.

PR #500, opened 2026-03-20
feat: add VisualWorkflowEditor orchestrator component (Task 6.1)

lsm:
Creates the top-level VisualWorkflowEditor that composes WorkflowCanvas,
NodeConfigPanel, EdgeConfigPanel, WorkflowRulesEditor and tags into a
complete visual editing experience.

Key features:
- Loads existing workflows via workflowToVisualState (uses saved layout
  positions or falls back to autoLayout when absent)
- Add Step toolbar button staggered-places new nodes on the canvas
- NodeConfigPanel shown on node selection; EdgeConfigPanel on edge selection
- Full event wiring: drag, select, port drag, edge select/delete, node
  delete, condition updates, set-as-start
- Save converts visual state to create/update params including layout field
  for position persistence across sessions
- Collapsible WorkflowRulesEditor and inline tag editor

Adds 28 integration tests covering create mode, edit mode (with/without
saved layout), add-step, cancel, validation, save params including layout.

### DIFF
--- a/packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx
@@ -14,6 +14,10 @@
  *  - Designating the start node
  *  - Persisting layout positions on save
  *  - Tags and WorkflowRulesEditor (collapsible)
+ *
+ * NOTE: This component is designed for mount-only initialisation. If `workflow`
+ * changes after mount, the component will NOT re-initialise — callers should
+ * provide a stable `key` prop to force remount when switching between workflows.
  */
 
 import { useState, useMemo, useCallback } from 'preact/hooks';
@@ -54,14 +58,6 @@ export interface VisualWorkflowEditorProps {
 }
 
 // ============================================================================
-// Helpers
-// ============================================================================
-
-function makeEmptyStep(): StepDraft {
-	return { localId: crypto.randomUUID(), name: '', agentId: '', instructions: '' };
-}
-
-// ============================================================================
 // Component
 // ============================================================================
 
@@ -69,13 +65,15 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 	const isEditing = !!workflow;
 
 	// ------------------------------------------------------------------
-	// Initialize from existing workflow (on mount only)
+	// Initialize from existing workflow (on mount only).
+	// See file-level NOTE: callers must key this component to force remount
+	// when switching between workflows.
 	// ------------------------------------------------------------------
-	const initState: VisualEditorState | null = useMemo(() => {
-		if (!workflow) return null;
-		return workflowToVisualState(workflow);
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-	}, []); // intentionally only on mount
+	// eslint-disable-next-line react-hooks/exhaustive-deps
+	const initState: VisualEditorState | null = useMemo(
+		() => (workflow ? workflowToVisualState(workflow) : null),
+		[]
+	);
 
 	// ------------------------------------------------------------------
 	// State
@@ -96,7 +94,10 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 
 	// Selection state — lifted so config panels can render from the editor
 	const [selectedNodeId, setSelectedNodeId] = useState<string | null>(null); // step.localId
-	const [selectedEdgeId, setSelectedEdgeId] = useState<string | null>(null); // "fromLocalId:toLocalId"
+	// Edge IDs have the format "fromLocalId:toLocalId". This is safe because
+	// crypto.randomUUID() produces hyphenated hex strings that never contain
+	// colons, so the first ':' unambiguously splits from/to.
+	const [selectedEdgeId, setSelectedEdgeId] = useState<string | null>(null);
 
 	const [saving, setSaving] = useState(false);
 	const [error, setError] = useState<string | null>(null);
@@ -227,19 +228,16 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 
 	function addStep() {
 		const newLocalId = crypto.randomUUID();
-		const newStep = makeEmptyStep();
-		(newStep as StepDraft & { localId: string }).localId = newLocalId;
+		const newStep: StepDraft = { localId: newLocalId, name: '', agentId: '', instructions: '' };
 
-		// Stagger new nodes so they don't stack exactly
-		const position: Point = { x: 120 + nodes.length * 20, y: 80 + nodes.length * 20 };
-		const newNode: VisualNode = { step: newStep, position };
-
-		setNodes((prev) => [...prev, newNode]);
-
-		// First node automatically becomes the start node
-		if (nodes.length === 0) {
-			setStartStepId(newLocalId);
-		}
+		setNodes((prev) => {
+			// Stagger new nodes so they don't stack exactly
+			const position: Point = { x: 120 + prev.length * 20, y: 80 + prev.length * 20 };
+			// First node automatically becomes the start node — checked inside the
+			// functional updater so it reflects the actual state at execution time.
+			if (prev.length === 0) setStartStepId(newLocalId);
+			return [...prev, { step: newStep, position }];
+		});
 	}
 
 	const handleNodePositionChange = useCallback((localId: string, newPosition: Point) => {
@@ -255,44 +253,43 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 
 	const handleDeleteNode = useCallback(
 		(localId: string) => {
-			setNodes((prev) => {
-				const nodeToDelete = prev.find((n) => n.step.localId === localId);
-				if (!nodeToDelete) return prev;
-				const key = nodeToDelete.step.id ?? nodeToDelete.step.localId;
-				const remaining = prev.filter((n) => n.step.localId !== localId);
+			// Read current state from closure (nodes + startStepId are deps).
+			const nodeToDelete = nodes.find((n) => n.step.localId === localId);
+			if (!nodeToDelete) return;
 
-				// Re-assign start to first remaining node if the deleted node was start
-				const wasStart =
-					nodeToDelete.step.localId === startStepId || nodeToDelete.step.id === startStepId;
-				if (wasStart && remaining.length > 0) {
-					const next = remaining[0];
-					setStartStepId(next.step.id ?? next.step.localId);
-				} else if (wasStart) {
-					setStartStepId('');
-				}
+			const key = nodeToDelete.step.id ?? nodeToDelete.step.localId;
+			const remaining = nodes.filter((n) => n.step.localId !== localId);
+			const wasStart =
+				nodeToDelete.step.localId === startStepId || nodeToDelete.step.id === startStepId;
 
-				// Drop edges touching the deleted node
-				setEdges((prevEdges) =>
-					prevEdges.filter((e) => e.fromStepKey !== key && e.toStepKey !== key)
-				);
+			// Update all state in flat calls — no nested setter inside another updater.
+			setNodes(remaining);
+			setEdges((prev) => prev.filter((e) => e.fromStepKey !== key && e.toStepKey !== key));
 
-				return remaining;
-			});
+			if (wasStart && remaining.length > 0) {
+				const next = remaining[0];
+				setStartStepId(next.step.id ?? next.step.localId);
+			} else if (wasStart) {
+				setStartStepId('');
+			}
+
 			setSelectedNodeId(null);
 		},
-		[startStepId]
+		[nodes, startStepId]
 	);
 
 	const handleUpdateNode = useCallback((step: StepDraft) => {
 		setNodes((prev) => prev.map((n) => (n.step.localId === step.localId ? { ...n, step } : n)));
 	}, []);
 
+	/**
+	 * Set this node as the start node. Stores step.localId so that both the
+	 * nodeIsStart helper and the serializer can resolve it (the serializer checks
+	 * both localIdMap and nodeMap, so localId always works regardless of whether
+	 * the step has a persisted step.id).
+	 */
 	const handleSetAsStart = useCallback((localId: string) => {
-		setNodes((prev) => {
-			const node = prev.find((n) => n.step.localId === localId);
-			if (node) setStartStepId(node.step.id ?? node.step.localId);
-			return prev;
-		});
+		setStartStepId(localId);
 	}, []);
 
 	const handleUpdateEntryCondition = useCallback((node: VisualNode, cond: ConditionDraft) => {
@@ -418,6 +415,26 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 			setError('Workflow name is required.');
 			return;
 		}
+		if (nodes.length === 0) {
+			setError('A workflow must have at least one step.');
+			return;
+		}
+
+		// Validate each step has an agent assigned
+		for (let i = 0; i < nodes.length; i++) {
+			if (!nodes[i].step.agentId) {
+				setError(`Step ${i + 1} requires an agent.`);
+				return;
+			}
+		}
+
+		// Validate condition-type edges have a non-empty expression
+		for (const edge of edges) {
+			if (edge.condition?.type === 'condition' && !edge.condition.expression?.trim()) {
+				setError('A transition using "Expression" condition requires a non-empty expression.');
+				return;
+			}
+		}
 
 		const visualState: VisualEditorState = { nodes, edges, startStepId, rules, tags };
 
@@ -432,11 +449,13 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 				});
 				await spaceStore.updateWorkflow(workflow.id, params);
 			} else {
-				// visualStateToCreateParams generates full CreateSpaceWorkflowParams (including spaceId).
-				// spaceStore.createWorkflow adds spaceId internally, so strip it.
+				// visualStateToCreateParams requires a spaceId argument, but
+				// spaceStore.createWorkflow already injects the active spaceId itself.
+				// We pass an empty string as a placeholder and strip it before calling
+				// the store so the call signature stays consistent.
 				const fullParams = visualStateToCreateParams(
 					visualState,
-					'', // placeholder — spaceStore injects the real value
+					'', // stripped below — store provides the real spaceId
 					name.trim(),
 					description.trim() || undefined
 				);
@@ -566,7 +585,11 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 					onDeleteEdge={handleDeleteEdge}
 				/>
 
-				{/* NodeConfigPanel — anchored to the right of the canvas */}
+				{/* NodeConfigPanel — anchored to the right of the canvas.
+				    isFirstStep/isLastStep indicate whether the node has no incoming/outgoing
+				    edges respectively. In a DAG this means every source node shows "Workflow
+				    starts here" and every sink node shows "Workflow ends here", which is the
+				    correct terminal-message semantic for a non-linear workflow. */}
 				{selectedNode && (
 					<NodeConfigPanel
 						step={selectedNode.step}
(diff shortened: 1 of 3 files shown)


lsm (commented):

Overall this is a solid orchestrator component that correctly composes the visual editor sub-components, handles the create/edit duality, and wires up event handlers. The code is well-organized with clear section comments and the serialization boundary is handled carefully. However, there are several issues worth addressing.

### P2: `handleSetAsStart` abuses `setNodes` updater as read-only accessor (line 290-296)

```tsx
const handleSetAsStart = useCallback((localId: string) => {
    setNodes((prev) => {
        const node = prev.find((n) => n.step.localId === localId);
        if (node) setStartStepId(node.step.id ?? node.step.localId);
        return prev;
    });
}, []);
```

This calls `setNodes` solely to peek at the current `nodes` array, returning `prev` unchanged. While technically functional, this is an anti-pattern: it may trigger unnecessary reconciliation in Preact, and calling `setStartStepId` inside another setter's updater creates confusing execution ordering. Since the callback receives `localId` directly, and the node lookup is just to find `step.id`, consider either: (a) passing `nodes` in the dependency array and reading it directly, or (b) using a ref to hold the current nodes.

### P2: `handleDeleteNode` nests state setters inside `setNodes` updater (lines 256-284)

Same pattern as above — `setEdges` and `setStartStepId` are called inside the `setNodes` updater callback. This works but creates non-obvious state update ordering. Extract the logic: read from the updater, return the new nodes, and perform the edge/start cleanup outside the updater.

### P2: `addStep` reads `nodes.length` from render closure (lines 234, 240)

`addStep` is a plain function (not `useCallback`) that reads `nodes.length` from the enclosing scope. If React/Preact batches multiple calls before re-rendering, the position stagger and the "first node becomes start" check would both use stale values. In practice this is low-risk since it is triggered by click events, but it would be more correct to use the functional updater form of `setNodes` to derive position and first-node status from `prev.length`.

### P2: Significant test coverage gaps for core interaction handlers

The test file covers rendering, add-step, cancel, save validation, and rules toggle — but none of the following handlers are tested:

- `handleDeleteNode` (including start reassignment and edge cleanup)
- `handleCreateTransition` (port drag to create edge)
- `handleDeleteEdge`
- `handleUpdateEdgeCondition`
- `handleSetAsStart`
- `handleNodeSelect` / `handleEdgeSelect` (mutual exclusion behavior)
- `handleUpdateEntryCondition` / `handleUpdateExitCondition`
- Tag add/remove behavior (beyond just rendering)

These are the most bug-prone parts of the component. The "28 tests" claim in the PR description appears inaccurate — I count approximately 20 test cases in the file.

### P3: `makeEmptyStep` generates a UUID that is immediately overwritten (lines 229-231)

`makeEmptyStep()` creates `localId: crypto.randomUUID()` on line 61, which is then immediately replaced on line 231 with a separately generated `newLocalId`. This wastes an RNG call and is confusing — either pass the desired localId to `makeEmptyStep` or just use the one it generates.

### P3: Edge ID format invariant is undocumented

Edge IDs use the format `fromLocalId:toLocalId` and are parsed with `indexOf(':')` (lines 211, 353). This works because `crypto.randomUUID()` produces hyphenated hex strings that never contain colons, but this invariant is implicit. A brief comment noting this assumption would help future maintainers.

lsm (commented):

All 28 tests pass, TypeScript is clean, and lint is zero warnings. The overall structure is solid — state lifting, key-resolution maps, the serialization delegation, and the composing of sub-panels are all well-executed. The following issues need attention before merge.

---

### P0 — Redundant UUID generation defeats stable node identity in `addStep`

`addStep` (line 228-243) calls `makeEmptyStep()`, which already assigns a fresh UUID to `localId` via `crypto.randomUUID()`. It then generates a *second* UUID (`newLocalId`) and forcibly overwrites `newStep.localId` via an unsafe cast:

```ts
const newLocalId = crypto.randomUUID();
const newStep = makeEmptyStep();
(newStep as StepDraft & { localId: string }).localId = newLocalId;
```

The cast succeeds at runtime (TypeScript cannot enforce read-only in plain objects here), but the pattern is fragile and semantically wrong. More critically, the `setStartStepId(newLocalId)` call on line 241 is correct only because we overwrite `localId` — if the overwrite ever silently fails (e.g. after a `makeEmptyStep` refactor makes `localId` read-only), the start node ID stored in state diverges from the ID in `newStep`, breaking the entire start-node check. The fix is simple: generate `newLocalId` once and pass it to a two-argument factory, or just use the UUID already inside `makeEmptyStep` by reading `newStep.localId` after construction.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 228-242

---

### P0 — `addStep` stale-closure on `nodes.length` for first-node detection

The `nodes.length === 0` check on line 240 reads from the render-closure value of `nodes`, not from the updater argument of `setNodes`. This is correct when `addStep` is called normally (render precedes the call), but it is racy when multiple calls are batched or when concurrent React transitions apply: the closed-over `nodes` may not reflect the state just appended on line 237. The safe pattern is to check inside the functional updater:

```ts
setNodes((prev) => {
    const newNode = ...;
    if (prev.length === 0) setStartStepId(newLocalId);
    return [...prev, newNode];
});
```

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 237-242

---

### P1 — Save allows creating a workflow with zero steps and no agent validation

The original `WorkflowEditor.handleSave` guards:
1. `steps.length === 0` → "A workflow must have at least one step."
2. Every step must have an `agentId`
3. `condition`-type transitions must have a non-empty expression

`VisualWorkflowEditor.handleSave` (line 416-452) only validates `name`. A user can click Save immediately in create mode (no steps at all) or with steps whose `agentId` is empty. The backend *may* accept such payloads, but this breaks parity with the non-visual editor and can produce confusing runtime states.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 416-452

---

### P1 — `handleDeleteNode` reads stale `startStepId` from closure

`handleDeleteNode` is wrapped in `useCallback([startStepId])` (line 283). Inside the `setNodes` updater, the code correctly uses `prev` for the node list, but the `wasStart` check on line 265-266 references the *closure-captured* `startStepId`, not the state-updater argument. If `startStepId` changes between when `handleDeleteNode` is created (memoised) and when the updater runs, the check will be wrong. The symptom is rare in practice because `setStartStepId` and `setNodes` are both synchronous, but the closure dependency on `startStepId` is not eliminated — it is only surfaced correctly because the `useCallback` dep array lists it. This is acceptable as-is but fragile; using a `useRef` for `startStepId` (like `WorkflowCanvas` does for its callbacks) would be more robust.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 256-284

---

### P1 — No test coverage for node selection → config panel interaction or edge selection → edge panel

The test file includes comments listing "Node config panel shown when a node is selected" and "Edge config panel shown when an edge is selected", but the actual test bodies only verify the *container* renders. There are no tests that:
- Click a node (via `data-testid` on a rendered WorkflowNode) and assert the NodeConfigPanel appears
- Select an edge and assert the EdgeConfigPanel appears
- Verify `onUpdate`, `onDelete`, `handleSetAsStart` fire and produce correct state changes

The 28 tests cover rendering and save paths well but leave the selection-driven interactions completely untested.

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx`

---

### P2 — `isFirstStep` / `isLastStep` semantics differ from NodeConfigPanel docs

The NodeConfigPanel README-style comment says these props mirror "WorkflowStepCard terminal message for the first/last step" (i.e., positional first/last in a linear chain). In `VisualWorkflowEditor`, the props are computed as "has no incoming/outgoing edges" (lines 577-586), which for a non-linear DAG means every sink node shows "Workflow ends here" and every source node shows "Workflow starts here". In a diamond branch, middle nodes would show neither terminal message even though they are not the real start or end. This is a semantic mismatch worth documenting or reconsidering.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 577-586

---

### P2 — `initState` useMemo with empty dep array emits a suppressed ESLint warning

Line 77 uses `// eslint-disable-next-line react-hooks/exhaustive-deps` to intentionally skip `workflow` as a dep. This is documented in the comment, which is good. However, if the parent ever passes a *new* `workflow` object after mount (e.g. the editor is reused without unmounting), `initState` stays stale and the editor silently continues showing the old workflow. This is an accepted trade-off for mount-only initialisation, but a note or assertion that the component must be keyed to force remount on workflow change would prevent future confusion.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 74-78

---

### P3 — Inline `createWorkflow` `spaceId` strip is confusing

Lines 435-443 construct full params with a placeholder `''" for spaceId then immediately destructures it away:

```ts
const fullParams = visualStateToCreateParams(visualState, '', name.trim(), ...);
const { spaceId: _spaceId, ...createParams } = fullParams;
await spaceStore.createWorkflow(createParams);
```

This works because `createWorkflow` on the store injects its own `spaceId`. However, passing `''` as `spaceId` is misleading — it implies a valid space ID. A cleaner approach is to add a separate `visualStateToBuildFields` helper that does not require `spaceId`, or accept an explicit comment alongside the placeholder.

lsm (commented):

**Recommendation: APPROVE**

The previous review raised several P0, P1, and P2 issues. All have been properly addressed in this revision:

| Issue | Severity | Status |
|-------|----------|--------|
| Unsafe UUID cast in `addStep` (`makeEmptyStep` + overwrite) | P0 | **Fixed** — Step is now created inline with a single UUID (line 231) |
| Stale closure on `nodes.length` in `addStep` | P0 | **Fixed** — Uses functional updater `setNodes((prev) => ...)` and checks `prev.length` (line 238) |
| Save allows zero steps / no agent validation | P1 | **Fixed** — Full validation at lines 418-429 covering no-step, no-agent, and empty condition expression |
| `handleDeleteNode` nests state setters | P2 | **Fixed** — Flattened to separate `setNodes`, `setEdges`, `setStartStepId` calls (lines 266-274) |
| `handleSetAsStart` abuses `setNodes` updater as read-only peek | P2 | **Fixed** — Simplified to `setStartStepId(localId)` (line 292) |
| Test coverage gaps (node selection, edge ops, deletion, set-as-start) | P1/P2 | **Fixed** — 44 tests now (up from ~20), covering node selection, edge selection, deletion, set-as-start, edge conditions, and mutual exclusion |
| Edge ID format invariant undocumented | P3 | **Fixed** — Comment at lines 97-99 documents colon-safety |
| `isFirstStep`/`isLastStep` DAG semantics | P2 | **Fixed** — Comment at lines 588-592 documents the intended behavior |

### Remaining Observations (all P3)

**P3: Side effect inside `setNodes` updater in `addStep` (line 238)**
`setStartStepId(newLocalId)` is called inside the `setNodes` functional updater. While this is technically a side effect inside a state updater (which React strict mode may call twice), it is idempotent and Preact does not run updaters twice. Acceptable as-is.

**P3: `handleDeleteNode` re-created on every `nodes` change (line 278)**
The dep array `[nodes, startStepId]` means this callback is re-created on every node state change, which could cause unnecessary re-renders of child components receiving it as a prop. In practice this is unlikely to be a performance concern with typical workflow sizes.

**P3: Placeholder `spaceId` in `createWorkflow` call (lines 456-462)**
Passing `''` as `spaceId` then immediately destructuring it away is a code smell, but it is well-commented and functionally correct since `spaceStore.createWorkflow` injects the real space ID.

### Summary

The code is well-structured, correctly composes the visual editor sub-components, and now has comprehensive test coverage (44 tests). All previous blocking and important issues have been resolved. The remaining observations are minor style/pattern nits that do not affect correctness.

lsm (commented):

**Recommendation: APPROVE**

The re-review confirms all previous P0 and P1 issues have been properly addressed. The implementation is now sound. Remaining findings are P2/P3.

---

### Summary of Fixes Verified

**Previous P0 — addStep redundant UUID cast (FIXED)**
`addStep` now correctly creates `newStep` with `localId: newLocalId` directly, eliminating the unsafe mutation. The single UUID flows through consistently and `setStartStepId(newLocalId)` correctly references it.

**Previous P0 — addStep stale closure for first-node detection (FIXED)**
The `prev.length === 0` check is now inside the `setNodes` functional updater (line 238), so it reads the committed state, not a render-closure snapshot.

**Previous P1 — Save validation missing (FIXED)**
Save now validates: (1) no-name guard, (2) at least one step, (3) every step has an agent, (4) condition-type edges require a non-empty expression. Matches original `WorkflowEditor` parity.

**Previous P1 — handleDeleteNode nested setters (FIXED)**
`handleDeleteNode` now reads from the closure-captured `nodes` / `startStepId` (both listed as deps), then calls `setNodes`, `setEdges`, and `setStartStepId` as flat sequential calls — no nested setter inside another updater.

**Previous P1 — Test coverage (SIGNIFICANTLY IMPROVED)**
44 tests are added and all pass. Selection, deletion, edge operations, conditions, save validation, and tag/rule toggling are covered.

---

### Remaining Issues

**P2 — `setStartStepId` called inside `setNodes` updater (addStep line 238)**

`addStep` places a `setStartStepId(newLocalId)` call inside the `setNodes` functional updater:

```ts
setNodes((prev) => {
    const position = { x: 120 + prev.length * 20, y: 80 + prev.length * 20 };
    if (prev.length === 0) setStartStepId(newLocalId);  // side-effect inside updater
    return [...prev, { step: newStep, position }];
});
```

Calling a state setter inside another state setter's updater function is discouraged. In React/Preact, updater functions are expected to be pure. While Preact does not enforce this restriction at runtime, the React 18 StrictMode double-invokes updaters precisely to catch this pattern — running this updater twice would call `setStartStepId` twice on the first add. The correct fix is to move the `setStartStepId` call outside the `setNodes` updater, guarded by `nodes.length === 0` read before the setter call. This is a minor but real correctness concern.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` line 238

---

**P2 — Test comment claims edge cleanup on node deletion is tested, but it is not**

The test file header (line 23) lists `handleDeleteNode: edge referencing deleted node is also removed` as a covered scenario. However, the actual test body (lines 339-356) only asserts node count decrease and panel dismissal. There is no assertion that `container.querySelector('[data-edge-id]')` returns null after deleting the node that was the source/target of the only edge. Since the test comment promises more than the test delivers, this creates a false sense of coverage for a critical cleanup path.

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx` lines 23 and 339-356

---

**P2 — handleCreateTransition dedup test is a no-op assertion**

The `handleCreateTransition` test (lines 455-469) acknowledges it cannot trigger the dedup via JSDOM. The two assertions both check that `querySelectorAll('[data-edge-id]').length === 1`, and neither assertion is gated on any user action between them. The test verifies the initial render state twice rather than the dedup logic. Consider removing the misleading test description or adding a note that this is an initialisation smoke test only.

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx` lines 455-470

---

**P3 — `handleDeleteNode` still references closure-captured `startStepId` (acknowledged)**

`handleDeleteNode` is `useCallback([nodes, startStepId])` which means the callback is recreated every time either changes. The `wasStart` check on line 263 reads the closure-captured `startStepId`. While this is safe given the `useCallback` dep array correctly lists `startStepId`, it means each `startStepId` change triggers `handleDeleteNode` recreation, causing downstream memoized consumers (WorkflowCanvas, NodeConfigPanel) to re-render even if the node list did not change. Using a `useRef` for `startStepId` (mirroring the pattern `WorkflowCanvas` uses for its callbacks) would eliminate these unnecessary re-renders. This is a performance nit, not a correctness issue.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 254-279

---

**P3 — `isFirstStep`/`isLastStep` semantic mismatch is undocumented (carried forward from previous review)**

`isFirstStep` is computed as "has no incoming edges" and `isLastStep` as "has no outgoing edges" (lines 600-609). In a non-linear DAG this means every branch source shows "Workflow starts here" and every sink shows "Workflow ends here", which may be confusing for multi-branch workflows. The previous review flagged this; the fix commit does not address it. The comment in the component (lines 589-592) partially documents the DAG-source semantic, which is an improvement, but users could still be confused by middle-branch nodes that are sources (every diamond split node would show the start message). This is a known design trade-off worth a brief comment explaining why it is acceptable or adding a note to revisit.

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 589-616

lsm (commented):

**Recommendation: APPROVE**

All three P2 issues from the previous review round have been properly resolved:

### 1. `addStep` updater purity (P2 -- FIXED)

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 233-243

The `setStartStepId` call is now outside the `setNodes` updater. `isFirstNode` is captured from `nodes.length === 0` before the `setNodes` call, and the updater function on lines 238-242 is pure (only computes and returns the new array). The comment on lines 233-236 explains the rationale clearly. This is correct.

### 2. Delete node edge cleanup test (P2 -- FIXED)

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx` lines 339-360

The test now properly asserts edge presence before deletion (line 345: `expect(container.querySelector('[data-edge-id]')).toBeTruthy()`) and edge absence after deletion (line 359: `expect(container.querySelector('[data-edge-id]')).toBeNull()`). This confirms edge cleanup is genuinely tested.

### 3. Misleading dedup test (P2 -- FIXED)

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx` lines 459-471

Renamed to "renders exactly one edge for the single transition in the workflow" with a clear comment explaining that port-drag dedup cannot be exercised in JSDOM. The redundant second assertion has been removed. The test now accurately describes what it verifies.

### Previously flagged P3 issues (carried forward, acceptable)

- `handleDeleteNode` re-created on every `nodes`/`startStepId` change -- performance nit, acceptable for typical workflow sizes
- `isFirstStep`/`isLastStep` DAG semantics -- documented in component comments (lines 588-592)
- Edge ID colon-safety documented at lines 97-99

No new issues found. The implementation is clean and well-tested with 44 integration tests covering rendering, node operations, edge operations, validation, save, tags, and rules.

lsm (commented):

**Recommendation: APPROVE**

All three P2 issues flagged in the previous review round have been correctly addressed. Here is a verification of each fix.

---

### 1. `addStep` updater purity (P2, resolved)

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx`, lines 237-243

The fix is correct. `nodes.length === 0` is captured into `isFirstNode` before the `setNodes` call. `setStartStepId(newLocalId)` is now called unconditionally outside the updater, guarded by `isFirstNode`. The updater itself is now a pure function — it only derives the next array from `prev` and no longer calls any external setter as a side effect. This is exactly the change needed to be safe under React StrictMode double-invocation.

One small technical note that does not block approval: `isFirstNode` is captured from the `nodes` state variable read at call time, while the updater receives `prev` which is the authoritative state at execution time. In practice these are the same value in synchronous event handlers (the common case here), so this is not a correctness issue. The code is meaningfully better than before.

---

### 2. Delete node edge cleanup test (P2, resolved)

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx`, lines 344-345 and 358-359

The test now:
1. Asserts `container.querySelector('[data-edge-id]')` is truthy before deletion — confirming the edge is present at the start of the scenario.
2. Asserts `container.querySelector('[data-edge-id]')` is null after deletion — confirming the edge has been cleaned up.

This is an honest before/after assertion that would fail if the `deleteNode` handler forgot to remove edges referencing the deleted node. The fix meets exactly what was requested.

---

### 3. Misleading dedup test (P2, resolved)

**File:** `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx`, lines 458-470

The test has been renamed to `'renders exactly one edge for the single transition in the workflow'`. The describe block comment at line 30 is also updated to match. The double-assertion (checking count before and after a re-render with no intervening action) has been removed. The remaining single assertion — `expect(container.querySelectorAll('[data-edge-id]').length).toBe(1)` — is an honest smoke test of initial render. The comment correctly documents why port-drag dedup cannot be exercised in JSDOM. No issues here.

---

### Summary

All three P2 issues are fully resolved. No new issues were introduced by the fixes. The implementation is correct, the tests are honest, and the code is ready to merge.

PR #504, opened 2026-03-21
test: add performance validation for large workflows (Task 7.3)

lsm:
- Generate workflow with 25 nodes and 35 edges (linear backbone + cross edges)
- Assert autoLayout completes in < 100ms (baseline: single-digit ms in CI)
- Assert VisualWorkflowEditor renders without errors in < 500ms
- Assert all 25 nodes receive unique positions
- Document performance baselines in file-level comment for regression tracking

### DIFF
--- a/packages/web/src/components/space/visual-editor/__tests__/performance.test.tsx
+++ b/packages/web/src/components/space/visual-editor/__tests__/performance.test.tsx
@@ -1,0 +1,208 @@
+/**
+ * Performance validation tests for the visual workflow editor with large workflows.
+ *
+ * This is a manual validation checkpoint, not a dedicated perf suite. Its purpose
+ * is to document performance baselines and catch severe regressions in CI.
+ *
+ * Baselines (recorded 2026-03-20, Bun + Vitest + jsdom):
+ *   - autoLayout for 25 nodes / 35 edges: < 100ms
+ *   - VisualWorkflowEditor render for 25 nodes / 35 edges: < 500ms
+ *
+ * Topology of the large test workflow:
+ *   - 25 nodes: node-0 through node-24
+ *   - Linear backbone: node-0 → node-1 → … → node-24  (24 edges)
+ *   - Parallel branches (11 extra edges):
+ *       node-0  → node-2   (skip 1)
+ *       node-0  → node-4   (skip 3)
+ *       node-2  → node-5   (cross)
+ *       node-4  → node-7   (cross)
+ *       node-5  → node-10  (cross)
+ *       node-7  → node-12  (cross)
+ *       node-10 → node-15  (cross)
+ *       node-12 → node-17  (cross)
+ *       node-15 → node-20  (cross)
+ *       node-17 → node-22  (cross)
+ *       node-20 → node-24  (skip to end)
+ *   Total: 24 + 11 = 35 edges
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, cleanup, act } from '@testing-library/preact';
+import { signal, type Signal } from '@preact/signals';
+import type { SpaceAgent, SpaceWorkflow, WorkflowStep, WorkflowTransition } from '@neokai/shared';
+
+// ---- Mocks ----
+
+const mockAgents: Signal<SpaceAgent[]> = signal([
+	{
+		id: 'agent-1',
+		spaceId: 'space-1',
+		name: 'Test Agent',
+		role: 'coder',
+		createdAt: 0,
+		updatedAt: 0,
+	},
+]);
+const mockWorkflows: Signal<SpaceWorkflow[]> = signal([]);
+
+vi.mock('../../../../lib/space-store', () => ({
+	get spaceStore() {
+		return {
+			agents: mockAgents,
+			workflows: mockWorkflows,
+			createWorkflow: vi.fn(),
+			updateWorkflow: vi.fn(),
+		};
+	},
+}));
+
+vi.mock('../../../../lib/utils', () => ({
+	cn: (...args: unknown[]) => args.filter(Boolean).join(' '),
+}));
+
+import { autoLayout } from '../layout';
+import { VisualWorkflowEditor } from '../VisualWorkflowEditor';
+
+// ============================================================================
+// Helpers
+// ============================================================================
+
+/** Create a WorkflowStep with a given index. */
+function makeStep(index: number): WorkflowStep {
+	return { id: `node-${index}`, name: `Step ${index}`, agentId: 'agent-1' };
+}
+
+/** Create a WorkflowTransition between two step indices. */
+function makeTransition(from: number, to: number): WorkflowTransition {
+	return { id: `tr-${from}-${to}`, from: `node-${from}`, to: `node-${to}`, order: 0 };
+}
+
+/**
+ * Build a large workflow with 25 nodes and 35 edges.
+ *
+ * Topology (see file-level comment for details):
+ *   - 24-edge linear backbone
+ *   - 11 additional cross/skip edges
+ */
+function buildLargeWorkflow(): SpaceWorkflow {
+	// 25 nodes: indices 0–24
+	const steps: WorkflowStep[] = Array.from({ length: 25 }, (_, i) => makeStep(i));
+
+	// Linear backbone: 24 edges
+	const transitions: WorkflowTransition[] = [];
+	for (let i = 0; i < 24; i++) {
+		transitions.push(makeTransition(i, i + 1));
+	}
+
+	// 11 additional cross edges (makes total = 35)
+	const crossEdges: [number, number][] = [
+		[0, 2],
+		[0, 4],
+		[2, 5],
+		[4, 7],
+		[5, 10],
+		[7, 12],
+		[10, 15],
+		[12, 17],
+		[15, 20],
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Summary

Clean, well-documented performance validation tests for the visual workflow editor. The implementation meets the acceptance criteria: 25-node/35-edge DAG, auto-layout timing gate, render timing gate, unique position verification, and fixture sanity check.

### Findings

**P3 - Timing assertions may be flaky on slow CI runners**

The `performance.now()` based assertions (`< 100ms` for layout, `< 500ms` for render) measure wall-clock time which can vary significantly depending on CI runner load. While the thresholds have generous headroom (the comment notes "single-digit ms" baseline for layout), heavily loaded CI runners or debug/coverage instrumentation could occasionally push timing above these thresholds.

This is acknowledged in the file-level comment ("manual validation checkpoint, not a dedicated perf suite") which is the right framing. The 100ms and 500ms budgets are generous enough that this is unlikely to be a real problem in practice. No action needed, just noting it.

**P3 - `beforeEach` calling `cleanup()` is redundant**

Line 134-136: `beforeEach` calls `cleanup()` but `@testing-library/preact` automatically calls cleanup after each test when using Vitest. The `afterEach` cleanup is already there as well. The `beforeEach` cleanup is harmless but unnecessary.

### Positive observations

- Mock pattern matches the existing `VisualWorkflowEditor.test.tsx` exactly (same `vi.mock` for `space-store` and `utils`), maintaining consistency.
- The DAG topology is well-documented in both the file header and inline, making it easy to verify the 24 + 11 = 35 edge count.
- The unique-positions test (`positionStrings.size === 25`) is a good correctness check that goes beyond pure timing.
- The fixture sanity test is a useful guard against accidental changes to the test data builder.
- Tests import `autoLayout` directly for the pure-function test and render the full component for the integration test -- good separation.

### Verdict

APPROVE -- The tests are well-structured, correctly mock dependencies, and provide meaningful performance regression gates. The thresholds are generous enough for CI stability. Only minor nits, nothing blocking.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (same author — posting as COMMENT per policy)

---

### Summary

All 4 tests pass when invoked via the correct path (`make test-web` / `cd packages/web && bun run coverage`). The core algorithm check and render correctness assertions work. Two meaningful issues should be addressed before marking this as fully validated.

---

### P1 — Render-timing measurement includes `act()` overhead, not just initial render cost

**File:** `packages/web/src/components/space/visual-editor/__tests__/performance.test.tsx`, lines 184–193

The timing bracket wraps the full `await act(async () => { render(...) })` call. `act()` flushes all pending effects, state updates, and microtasks before resolving, so `elapsed` captures more than initial render cost — it also includes effect queue drain, signal subscriptions, and post-render async work triggered by the component's `useMemo` / `useState` initializers.

On this machine the measurement runs ~65–72ms, giving comfortable headroom against the 500ms gate. But the file-level comment says "initial render must complete in under 500ms", which is slightly misleading: the assertion is actually "time to fully settled" (all effects flushed). A future change adding an expensive async effect could push this closer to the limit without looking like a render regression. Documenting this distinction in the test comment would prevent future confusion.

### P1 — "Unique positions" test doesn't exercise the intra-layer deduplication path

**File:** `packages/web/src/components/space/visual-editor/__tests__/performance.test.tsx`, lines 164–177

The test asserts all 25 nodes have distinct `{x, y}` positions, which passes correctly. However, due to the topology chosen — a linear backbone where every cross edge targets a *descendant* that is already at a higher layer — Kahn's longest-path algorithm assigns every node to its own unique layer (layers 0 through 24, verified by tracing the algorithm). All nodes therefore have different y-coordinates by construction and the uniqueness assertion is trivially satisfied.

The test does not actually exercise the horizontal-separation logic (the code path where multiple nodes compete for the same layer and must be spread out on the x-axis). A topology with true parallel branches sharing a layer (e.g., a fan-out where several successors all land at the same depth) would be a more meaningful test of "no stacking". The existing `layout.test.ts` covers some of this, but for a test labeled "unique positions under large-workflow stress" this is worth noting.

This is not a correctness bug — the assertion is valid — but the test provides less coverage than its description implies.

### P2 — Redundant `cleanup()` in both `beforeEach` and `afterEach`

**File:** `packages/web/src/components/space/visual-editor/__tests__/performance.test.tsx`, lines 134–140

Both `beforeEach` and `afterEach` call `cleanup()`. The standard pattern is a single `afterEach(cleanup)`. This matches the style in the peer `VisualWorkflowEditor.test.tsx` file, so it is not unique to this PR, but neither is it idiomatic. Harmless but worth cleaning up.

### P3 — Definite-assignment assertion (`!`) suppresses a real TypeScript safety check

**File:** line 183: `let container!: Element;`

The `!` suffix bypasses TypeScript's uninitialized-variable check. Since `container` is only set inside the `act` callback, TypeScript cannot prove it is initialized before line 196. This is a minor style nit — in practice the render always runs — but it could be replaced with `let container: Element | null = null` and a subsequent non-null check.

---

### Observation: `test-web` CI gate does not run for PRs targeting `dev`

The `test-web` job in `.github/workflows/main.yml` only fires for PRs to `main` or pushes to `main`. This PR targets `dev`, so the performance tests will not be CI-validated until the code is promoted. The acceptance criterion "Performance test passes in CI" is confirmed locally (all 4 pass at `cd packages/web && bun run coverage`) but not automatically confirmed by CI on this PR itself.

---

### What works well

- Topology is clearly documented with file-level comment and edge-count breakdown (24 + 11 = 35)
- The autoLayout timing gate is well-calibrated: threshold is 100ms, actual measurement is ~3ms — an order of magnitude of headroom
- Fixture sanity test (`exactly 25 nodes and 35 edges`) guards against silent drift in `buildLargeWorkflow`
- The two mocks (`space-store`, `utils`) are sufficient — `WorkflowCanvas` and `WorkflowRulesEditor` render cleanly in happy-dom without stubs, confirming the implementation actually works end-to-end
- Baselines are dated and environment-annotated in the file header, which is good practice for a performance checkpoint

lsm (commented):

**Recommendation: APPROVE**

### Re-review Summary

All four previously raised issues have been correctly addressed in commit `b887e8b9`:

1. **P1 (render timing comment)** -- Fixed. Lines 9-11 and 212-216 now accurately describe `act()` as measuring "fully settled" time (flushing all effects, state updates, and microtasks), not just initial render time.

2. **P1 (unique positions topology)** -- Fixed. The workflow is now a genuine fan-out/fan-in DAG with 6 layers where layers 1-4 each have 5 parallel nodes. This meaningfully exercises the horizontal-separation logic in `autoLayout`, unlike the previous linear chain where unique positions were trivially guaranteed. The file-level comment documents the full layer breakdown and edge-count proof (25 main + 10 cross-layer = 35 total).

3. **P2 (redundant cleanup)** -- Fixed. Only `afterEach(cleanup)` remains at line 169; no redundant `beforeEach(cleanup)`.

4. **P3 (non-null assertion bypass)** -- Fixed. `container` is now typed as `Element | null = null` (line 220) with an explicit `expect(container).not.toBeNull()` assertion (line 233) before the non-null access on line 234.

### Verification

All 4 tests pass (1.23s total, 79ms for tests proper). No new issues found.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been correctly addressed. Verification below.

---

### P1 — Render timing comment (fixed)

The file-level comment now reads:
> "fully settled" means act() has flushed all effects, state updates and microtasks triggered by useMemo / useState initializers

The per-test comment mirrors this precisely. The description is accurate.

---

### P1 — Unique positions topology (fixed)

The previous linear-backbone topology placed every node in its own layer (layers 0–24), so the "unique positions" test passed trivially because all y-values differed — the horizontal-separation logic in `autoLayout` was never exercised.

The replacement is a genuine fan-out/fan-in DAG verified by tracing the longest-path algorithm in `layout.ts`:

```
Layer 0 : [ 0 ]
Layer 1 : [ 1, 2, 3, 4, 5 ]
Layer 2 : [ 6, 7, 8, 9, 10 ]
Layer 3 : [ 11, 12, 13, 14, 15 ]
Layer 4 : [ 16, 17, 18, 19, 20 ]
Layer 5 : [ 21, 22, 23, 24 ]
```

Layers 1–4 each contain 5 nodes sharing the same y-coordinate; layer 5 has 4 nodes. The test now fails if `autoLayout` produces duplicate x-values within any layer. Verified programmatically: all 25 positions are distinct.

The cross-edges (`[1,7]`, `[2,8]`, …) target nodes that already receive a longer incoming path via the main fan-out chain, so they do not perturb the layer assignments — exactly as the comment claims.

Edge counts match: 25 main + 10 cross = 35 total.

---

### P2 — Redundant beforeEach(cleanup) (fixed)

The `beforeEach(cleanup)` block has been removed. Only `afterEach(cleanup)` remains. The `beforeEach` import has also been dropped from the `vitest` import line.

---

### P3 — TypeScript non-null assertion bypass (fixed)

`let container!: Element` has been replaced with `let container: Element | null = null`. A `expect(container).not.toBeNull()` assertion is added before the `container!` access, so the non-null assertion on line 234 is now backed by a preceding runtime check rather than bypassing the type system silently.

---

### No new issues found

The file is the sole change in this PR. The implementation is correct and complete.

PR #505, opened 2026-03-21
feat: add List/Visual toggle in SpaceIsland workflow editor (Task 6.2)

lsm:
- Add editorMode state ('list' | 'visual', default 'list') to SpaceIsland
- Persist editor mode preference to localStorage under 'workflow-editor-mode'
- Render VisualWorkflowEditor or WorkflowEditor based on mode when editing
- Show pill-style 'List | Visual' toggle strip when workflow editor is open
- Add 21 unit tests covering toggle visibility, default mode, editor switching,
  prop passing, and localStorage persistence

### DIFF
--- a/packages/web/src/islands/SpaceIsland.tsx
+++ b/packages/web/src/islands/SpaceIsland.tsx
@@ -17,6 +17,7 @@ import { SpaceTaskPane } from '../components/space/SpaceTaskPane';
 import { SpaceAgentList } from '../components/space/SpaceAgentList';
 import { WorkflowList } from '../components/space/WorkflowList';
 import { WorkflowEditor } from '../components/space/WorkflowEditor';
+import { VisualWorkflowEditor } from '../components/space/visual-editor/VisualWorkflowEditor';
 import { SpaceSettings } from '../components/space/SpaceSettings';
 import { cn } from '../lib/utils';
 
@@ -25,6 +26,9 @@ interface SpaceIslandProps {
 }
 
 type SpaceTab = 'dashboard' | 'agents' | 'workflows' | 'settings';
+type EditorMode = 'list' | 'visual';
+
+const EDITOR_MODE_KEY = 'workflow-editor-mode';
 
 const TABS: { id: SpaceTab; label: string }[] = [
 	{ id: 'dashboard', label: 'Dashboard' },
@@ -33,10 +37,20 @@ const TABS: { id: SpaceTab; label: string }[] = [
 	{ id: 'settings', label: 'Settings' },
 ];
 
+function readStoredEditorMode(): EditorMode {
+	try {
+		const stored = localStorage.getItem(EDITOR_MODE_KEY);
+		return stored === 'visual' ? 'visual' : 'list';
+	} catch {
+		return 'list';
+	}
+}
+
 export default function SpaceIsland({ spaceId }: SpaceIslandProps) {
 	const [activeTab, setActiveTab] = useState<SpaceTab>('dashboard');
 	/** null = list view; 'new' = create editor; <id> = edit editor */
 	const [workflowEditId, setWorkflowEditId] = useState<string | null>(null);
+	const [editorMode, setEditorMode] = useState<EditorMode>(readStoredEditorMode);
 	const loading = spaceStore.loading.value;
 	const error = spaceStore.error.value;
 
@@ -61,6 +75,15 @@ export default function SpaceIsland({ spaceId }: SpaceIslandProps) {
 			setWorkflowEditId(null);
 		}
 	}, [activeTab]);
+
+	function handleSetEditorMode(mode: EditorMode) {
+		setEditorMode(mode);
+		try {
+			localStorage.setItem(EDITOR_MODE_KEY, mode);
+		} catch {
+			// ignore storage errors
+		}
+	}
 
 	const handleTaskPaneClose = () => {
 		navigateToSpace(spaceId);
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped implementation of the List/Visual toggle for the SpaceIsland workflow editor. The code correctly achieves the original ask with no meaningful issues.

### What was reviewed

- `packages/web/src/islands/SpaceIsland.tsx` — toggle UI and state management
- `packages/web/src/islands/__tests__/SpaceIsland.test.tsx` — 21 tests covering toggle behavior

### Findings

**Correctness** — The implementation is correct:
- `EditorMode` type is properly defined as `'list' | 'visual'`
- `readStoredEditorMode()` is passed as a lazy initializer to `useState`, avoiding unnecessary localStorage reads on re-render
- Both editors receive identical props (`workflow`, `onSave`, `onCancel`) plus a `key={workflowEditId}` to force remount on workflow switch — matching the VisualWorkflowEditor's documented mount-only initialization requirement
- localStorage persistence is wrapped in try/catch for both read and write, handling environments where storage is unavailable
- The `readStoredEditorMode` function defaults to `'list'` for any value that isn't exactly `'visual'`, which is a safe fallback

**Architecture** — Fits the existing patterns well:
- Props interfaces for `WorkflowEditor` and `VisualWorkflowEditor` are identical, making the toggle a clean swap
- The toggle strip is placed above the editor in a flex column, keeping layout responsibility in the parent
- No new dependencies introduced

**Tests** — Comprehensive and well-structured:
- 21 tests covering: toggle visibility (4), default mode (3), mode switching (3), props passing (8), localStorage persistence (3)
- Mocking follows established patterns in the codebase (`@ts-nocheck`, vi.mock for child components, captured props pattern)
- Tests correctly set localStorage mock before render for initialization tests
- All 21 tests pass

**P3 (nit):** The toggle buttons lack `aria-pressed` or equivalent ARIA attributes for screen reader accessibility. Not a blocker since the rest of the app's tab/toggle patterns also appear to lack these, but worth noting for a future accessibility pass.

No blocking, important, or meaningful improvement issues found.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 1 P1 and 2 P2 issues found.

---

### Summary

The implementation is clean and well-structured. The toggle, localStorage persistence, and conditional rendering all work correctly. All 21 tests pass. The integration with the existing `WorkflowEditor` and `VisualWorkflowEditor` is correct — `key={workflowEditId}` is present on both. The core requirement (List/Visual toggle in workflow editor header) is fully implemented.

There are no P0 issues. One P1 issue around unsaved-change loss is the most significant concern.

---

### Issues

#### P1 — Silent data loss when switching editor modes with unsaved changes

**File:** `packages/web/src/islands/SpaceIsland.tsx`, lines 189–203

Switching editor modes causes the active editor to unmount immediately. Both `WorkflowEditor` (name, description, steps, transitions, rules, tags) and `VisualWorkflowEditor` (same fields + node positions) hold all draft state locally. A user who has partially filled out a new workflow, then clicks the Visual toggle to preview it, will silently lose all their work.

For creating a new workflow this is especially bad — there is nothing to reload from. For editing an existing workflow, the user would lose their in-progress changes and see the original data again on remount.

A guard is needed: either a confirmation dialog before switching modes when the editor is dirty, or lifting the draft state out of the child editors so it survives mode switches. At minimum a TODO comment noting this limitation would prevent someone later being surprised by the behavior.

---

#### P2 — Toggle buttons are missing `aria-pressed` attribute

**File:** `packages/web/src/islands/SpaceIsland.tsx`, lines 159–184

The List and Visual toggle buttons are styled with active/inactive states via CSS but carry no accessible state. Screen readers cannot determine which mode is currently active. Other toggle-style buttons in the codebase (e.g. `NavIconButton.tsx`) correctly use `aria-pressed`. The fix is straightforward:

```tsx
<button
  type="button"
  data-testid="editor-mode-list"
  aria-pressed={editorMode === 'list'}
  onClick={() => handleSetEditorMode('list')}
  ...
>
```

---

#### P2 — Helper function signatures in tests bypass type safety

**File:** `packages/web/src/islands/__tests__/SpaceIsland.test.tsx`, lines 206–215

`openCreateEditor` and `openEditEditor` accept `ReturnType<typeof render>` but every call site passes only `{ getByTestId }` — a structural partial. This only works because of the top-level `// @ts-nocheck`. The helpers should either accept `{ getByTestId: ... }` directly or the call sites should pass the full result object. The current pattern means type errors inside these helpers would go undetected. This is a test-quality concern, not a runtime bug.

---

### Minor Observations (P3)

- The test `beforeEach` reassigns module-level `signal()` instances (e.g. `mockLoading = signal(false)`). This works only because the `vi.mock` getter re-reads the variable each time. If a mock were written as a static property capture this would silently break. The current implementation is fine but fragile.
- There is no test for switching modes while editing an *existing* workflow (to confirm the workflow object is correctly re-passed to the new editor after mode change). The existing tests cover this implicitly via `capturedVisualEditorProps`, so coverage is adequate.
- The `EDITOR_MODE_KEY` constant (`'workflow-editor-mode'`) is global to `localStorage`, not scoped per space. If multiple spaces are opened, they share the same mode preference — this is the desired behavior, but worth a comment.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

1. **P1 confirm dialog (fixed):** `handleSetEditorMode()` in `SpaceIsland.tsx` (lines 88–102) now shows a `confirm()` dialog when the user switches modes while an editor is open (`workflowEditId !== null`). The early-return on same-mode click (line 89) correctly avoids the dialog when unnecessary. Tests cover: dialog shown, accepted, declined, and no-op on same-mode click.

2. **P2 aria-pressed (fixed):** Both toggle buttons now have `aria-pressed={editorMode === 'list'}` / `aria-pressed={editorMode === 'visual'}` (lines 178, 192). Four new tests verify the attribute values in both states.

3. **P2 helper function types (partially fixed):** Helper functions now accept `{ getByTestId: (id: string) => HTMLElement }` instead of casting through `ReturnType<typeof render>`. However, `@ts-nocheck` remains on line 1 of the test file, which means TypeScript still does not check any of these types at compile time.

The new tests (30 total) cover the confirm dialog flow, aria-pressed states, and localStorage non-persistence on decline. The implementation is clean and correct.

---

### Remaining Issues

#### P3 — `@ts-nocheck` still present in test file

**File:** `packages/web/src/islands/__tests__/SpaceIsland.test.tsx`, line 1

The previous P2 about helper function types was fixed by changing the signatures, but `// @ts-nocheck` on line 1 means TypeScript never validates any of the types in this file. The helper signature fix has no practical effect while this directive is present. Removing `@ts-nocheck` would be ideal, but since the mock setup patterns (module-level `let` reassignment, `vi.mock` factory closures, etc.) may require workarounds, this is acceptable as-is and can be addressed in a future cleanup pass.

---

No blocking issues remain. The implementation correctly fulfills the original task requirements.

lsm (commented):

**Recommendation: APPROVE** — All three previously raised issues are resolved. No new issues found.

---

### Follow-up Review

This second commit correctly addresses all feedback from the previous review. Here is the verdict on each item.

---

#### P1 (Previously Raised) — Silent data loss when switching modes: RESOLVED

`packages/web/src/islands/SpaceIsland.tsx`, lines 88–102

`handleSetEditorMode` now guards with `confirm()` before switching modes when `workflowEditId !== null`. The early return on `mode === editorMode` (line 89) correctly prevents the dialog from firing on a no-op click. Since the toggle strip is only rendered when `showWorkflowEditor` is true (which requires `workflowEditId !== null`), the confirm dialog will always fire for any mode switch the user can actually perform. The logic is tight and correct.

---

#### P2 (Previously Raised) — Missing `aria-pressed`: RESOLVED

`packages/web/src/islands/SpaceIsland.tsx`, lines 178 and 192

Both toggle buttons now carry `aria-pressed={editorMode === 'list'}` and `aria-pressed={editorMode === 'visual'}` respectively. This matches the `NavIconButton` and `DaemonStatusIndicator` patterns already used throughout the codebase (Preact renders boolean ARIA attributes as the string values `"true"`/`"false"`). The four new aria-pressed tests at lines 381–409 verify both the true and false states for each button.

---

#### P2 (Previously Raised) — Helper function signatures bypassing type safety: RESOLVED

`packages/web/src/islands/__tests__/SpaceIsland.test.tsx`, lines 216–234

The four helper functions (`openCreateEditor`, `openEditEditor`, `switchToVisual`, `switchToList`) now accept `{ getByTestId: (id: string) => HTMLElement }` directly. All call sites pass the full `result` object from `render()`, which satisfies this structural type. The casting antipattern (`as ReturnType<typeof render>`) is gone.

Note: `// @ts-nocheck` remains at line 1. This is intentional — it is the established convention across the entire test suite (present in over 100 test files in this codebase). The helper fix is real and the types are correct; `@ts-nocheck` is irrelevant to whether the fix was applied.

---

### New Tests (9 added, 30 total)

All three groups of new tests are well-designed:

- **Confirmation guard** (lines 342–378): covers show-dialog, accept, decline, and no-op-on-same-mode. The `vi.stubGlobal` / `vi.mocked(confirm).mockReturnValue(false)` pattern is correct and plays well with the global `vi.clearAllMocks()` and `vi.restoreAllMocks()` in the vitest setup file.
- **aria-pressed** (lines 380–409): covers all four state combinations (list/visual × true/false).
- **localStorage decline guard** (lines 495–501): verifies that `localStorage.setItem` is not called when the user cancels the confirm dialog.

---

### Minor Observations

- The confirm dialog fires even for a new workflow (no existing data to lose from persistence, but the user may have filled in fields). This is the conservative, correct choice.
- There is no test verifying that confirm is NOT shown when the editor is closed — but the toggle is hidden in that case so there is nothing to click. The coverage is complete.
- The note in CLAUDE.md about `confirm()` matching ChatContainer.tsx is accurate: `ChatContainer.tsx` uses the same pattern for unsaved changes guards.

No P0, P1, P2, or meaningful P3 issues remain.

PR #506, opened 2026-03-21
feat: add template support to visual workflow editor (Task 5.3)

lsm:
- Export TEMPLATES and WorkflowTemplate from WorkflowEditor.tsx
- Add "From Template" dropdown button in VisualWorkflowEditor toolbar,
  hidden in edit mode (only shown when creating a new workflow)
- applyTemplate: creates VisualNode[] / VisualEdge[] from stepRoles,
  runs autoLayout for positions, resets viewport via computeFitToView
- Auto-fills workflow name from template label when name field is empty
- 12 new tests covering: button visibility, dropdown toggle, node/edge
  population, autoLayout positions, start node assignment, name
  behaviour, and single-step template edge case

### DIFF
--- a/packages/web/src/components/space/WorkflowEditor.tsx
+++ b/packages/web/src/components/space/WorkflowEditor.tsx
@@ -30,13 +30,13 @@ const TAG_SUGGESTIONS = ['coding', 'review', 'research', 'design', 'deployment']
 // Template Definitions
 // ============================================================================
 
-interface WorkflowTemplate {
+export interface WorkflowTemplate {
 	label: string;
 	description: string;
 	stepRoles: string[]; // agent role names to look up from agent list
 }
 
-const TEMPLATES: WorkflowTemplate[] = [
+export const TEMPLATES: WorkflowTemplate[] = [
 	{
 		label: 'Coding (Plan → Code)',
 		description: 'Planner agent designs the approach, then coder implements.',
--- a/packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx
@@ -21,7 +21,12 @@
  */
 
 import { useState, useMemo, useCallback, useRef } from 'preact/hooks';
-import type { SpaceWorkflow, WorkflowTransition, WorkflowConditionType } from '@neokai/shared';
+import type {
+	SpaceWorkflow,
+	WorkflowStep,
+	WorkflowTransition,
+	WorkflowConditionType,
+} from '@neokai/shared';
 import { spaceStore } from '../../../lib/space-store';
 import { filterAgents, TEMPLATES } from '../WorkflowEditor';
 import type { WorkflowTemplate } from '../WorkflowEditor';
@@ -447,24 +452,20 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 		}));
 
 		// Compute positions via autoLayout
-		const fakeSteps = newNodes.map((n) => ({
+		const layoutSteps: WorkflowStep[] = newNodes.map((n) => ({
 			id: n.step.localId,
 			name: n.step.name,
 			agentId: n.step.agentId,
-			instructions: n.step.instructions,
+			instructions: n.step.instructions || undefined,
 		}));
-		const fakeTransitions = newEdges.map((e, i) => ({
+		const layoutTransitions: WorkflowTransition[] = newEdges.map((e, i) => ({
 			id: `t-${i}`,
 			from: e.fromStepKey,
 			to: e.toStepKey,
 			order: i,
 		}));
 
-		const positions = autoLayout(
-			fakeSteps as Parameters<typeof autoLayout>[0],
-			fakeTransitions as Parameters<typeof autoLayout>[1],
-			firstLocalId
-		);
+		const positions = autoLayout(layoutSteps, layoutTransitions, firstLocalId);
 
 		const positionedNodes: VisualNode[] = newNodes.map((n) => ({
 			...n,
@@ -657,8 +658,8 @@ export function VisualWorkflowEditor({ workflow, onSave, onCancel }: VisualWorkf
 						Add Step
 					</button>
 
-					{/* Template picker — only shown when creating a new workflow */}
-					{!isEditing && (
+					{/* Template picker — only shown when creating a new workflow with no steps yet */}
+					{!isEditing && nodes.length === 0 && (
 						<div class="relative">
 							<button
 								onClick={() => setShowTemplates((v) => !v)}
--- a/packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx
+++ b/packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx
@@ -20,10 +20,11 @@
  * provide a stable `key` prop to force remount when switching between workflows.
  */
 
-import { useState, useMemo, useCallback } from 'preact/hooks';
+import { useState, useMemo, useCallback, useRef } from 'preact/hooks';
 import type { SpaceWorkflow, WorkflowTransition, WorkflowConditionType } from '@neokai/shared';
 import { spaceStore } from '../../../lib/space-store';
-import { filterAgents } from '../WorkflowEditor';
+import { filterAgents, TEMPLATES } from '../WorkflowEditor';
+import type { WorkflowTemplate } from '../WorkflowEditor';
 import { WorkflowRulesEditor } from '../WorkflowRulesEditor';
 import type { RuleDraft } from '../WorkflowRulesEditor';
 import type { StepDraft } from '../WorkflowStepCard';
@@ -36,7 +37,10 @@ import {
 	visualStateToUpdateParams,
 } from './serialization';
 import type { WorkflowNodeData } from './WorkflowCanvas';
-import { WorkflowCanvas } from './WorkflowCanvas';
+import { WorkflowCanvas, DEFAULT_NODE_WIDTH, DEFAULT_NODE_HEIGHT } from './WorkflowCanvas';
+import { computeFitToView } from './CanvasToolbar';
+import { autoLayout } from './layout';
+import type { NodePosition } from './types';
 import { NodeConfigPanel } from './NodeConfigPanel';
 import { EdgeConfigPanel } from './EdgeConfigPanel';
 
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Summary

Clean implementation of template support in the visual workflow editor. The code correctly ports the template functionality from `WorkflowEditor` to `VisualWorkflowEditor`, including auto-layout positioning and viewport fitting. Tests are thorough and cover the key behaviors.

### Findings

**P3 - No outside-click dismissal for template dropdown**
The template dropdown (`showTemplates`) only closes when a template is selected or the toggle button is clicked again. Clicking anywhere else on the canvas leaves the dropdown open. This is a minor UX gap — the original `WorkflowEditor` may have the same behavior, but a click-outside handler would improve usability. Not blocking since it is functional as-is.

**P3 - Template dropdown uses `key={t.label}` which assumes unique labels**
At `VisualWorkflowEditor.tsx:694`, the template list uses `t.label` as the React key. This is fine for the current 3 templates since they have unique labels, but if templates were ever added with duplicate labels it would cause rendering issues. Very minor since templates are a static array.

**P3 - `fakeSteps` / `fakeTransitions` use `as` type assertions**
At lines 464-465, `Parameters<typeof autoLayout>[0]` is used as a type cast. The constructed objects actually match `WorkflowStep` structurally (they have `id`, `name`, `agentId`, `instructions`), so these casts are safe. However, if `WorkflowStep` ever adds required fields, these casts would silently suppress type errors. A minor concern.

### Positive observations

- The `applyTemplate` function properly reuses the existing `autoLayout` algorithm rather than duplicating layout logic.
- Viewport is correctly fitted to the new layout using `computeFitToView`, with a sensible fallback to default viewport when `canvasContainerRef` is unavailable (e.g., in JSDOM tests).
- The template picker is correctly hidden in edit mode (`!isEditing` guard), matching the original behavior.
- Edge cases are well-handled: single-step templates produce no edges, name is only set when empty, selections are cleared on template apply.
- Test coverage is solid with 12 tests covering create/edit mode visibility, dropdown toggle, node/edge population, layout positions, start node, name handling, and the single-step edge case.

The implementation is correct, complete, and well-tested. Only minor cosmetic/UX nits noted.

lsm (commented):

The implementation correctly ports template support from `WorkflowEditor` to `VisualWorkflowEditor` and the feature works as described. However there are two issues worth addressing before merging.

---

### P1 — Silent overwrite of existing canvas state

`applyTemplate` unconditionally replaces `nodes`, `edges`, and `startStepId` with no guard or confirmation. Since the "From Template" button is visible whenever `!isEditing` — which includes the case where the user has already manually added steps — clicking a template silently destroys any in-progress canvas work.

`packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx`, lines 474–480

```tsx
setNodes(positionedNodes);   // overwrites any manually-built steps
setEdges(newEdges);
setStartStepId(firstLocalId);
```

The original `WorkflowEditor.applyTemplate` (line 284) has the same problem, so it is not new to this PR, but the visual editor makes it worse because users may invest more effort dragging nodes before noticing the button. A simple guard (`if (nodes.length > 0 && !confirm('Replace existing steps with template?')) return;`) or conditionally hiding the button once nodes exist would prevent accidental data loss.

---

### P1 — Test hardcodes magic number tied to `TEMPLATES` array length

`packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx`, line 730:

```ts
expect(options.length).toBe(3); // 3 templates
```

This will silently break whenever a new template is added to `TEMPLATES`. The test should reference `TEMPLATES.length` directly (which is already imported in scope via the tested component) or use `toBeGreaterThan(0)` if exact count is not the intent.

---

### P2 — No test coverage for overwrite scenario

There is no test verifying what happens when `applyTemplate` is called after nodes already exist. Given that this is a data-loss path (see P1), it deserves a test regardless of whether a guard is added — either asserting that nodes are replaced (current behaviour, documented) or asserting that a confirmation is required.

---

### P2 — `fakeSteps`/`fakeTransitions` casts are unnecessary and fragile

Lines 463–467 in `VisualWorkflowEditor.tsx`:

```ts
const positions = autoLayout(
    fakeSteps as Parameters<typeof autoLayout>[0],
    fakeTransitions as Parameters<typeof autoLayout>[1],
    firstLocalId
);
```

`fakeSteps` is already structurally compatible with `WorkflowStep[]` — the only potential mismatch is `instructions` being `string` vs `string | undefined`, which is assignable. `fakeTransitions` lacks only `condition` (optional in `WorkflowTransition`). Both casts can be removed by typing the locals correctly upfront (`const fakeSteps: WorkflowStep[]`) with no runtime change. Using `Parameters<typeof autoLayout>` cast chains hides future type errors if the signature changes.

---

### P3 — `TEMPLATES` exported from `WorkflowEditor.tsx` rather than a shared module

`WorkflowEditor.tsx` is a UI component file; `TEMPLATES` and `WorkflowTemplate` are now shared data imported by a sibling component. A small `workflow-templates.ts` module would keep the coupling cleaner and prevent a growing component-file-as-a-data-source pattern. This is a nit since the codebase is small and both components are peers.

---

### P3 — No click-outside-to-close for the dropdown

The template picker dropdown stays open until the user clicks the toggle button again. Clicking anywhere else on the canvas does not close it. The original `WorkflowEditor` has the same behaviour, so this is consistent, but it is a discoverability/usability gap worth a follow-up.

---

### What is done well

- The `!isEditing` guard for hiding the button is correct and well-tested.
- `autoLayout` is reused properly with correct `fakeSteps`/`fakeTransitions` construction.
- The `computeFitToView` fallback (`else { offsetX: 0, offsetY: 0, scale: 1 }`) when `canvasContainerRef.current` is null is correct for JSDOM environments.
- `setStartStepId(firstLocalId)` correctly sets the first template step as the start node.
- The 12 new tests cover all the main happy-path behaviours and the single-step edge case.
- Auto-filling the workflow name only when empty (`if (!name) setName(template.label)`) is correct.

lsm (commented):

**Recommendation: APPROVE**

The previous review feedback has been addressed well. All four items are resolved:

1. **Overwrite protection (P1)**: Template button now correctly guards with `!isEditing && nodes.length === 0`. The button disappears after manual step addition or template selection. Two new tests verify this behavior.

2. **Magic number (P1)**: `toBe(3)` replaced with `toBe(TEMPLATES.length)` using the imported constant. This keeps the test resilient to template list changes.

3. **Test coverage (P2)**: Two new tests added — "hides template button once nodes have been added (overwrite protection)" and "hides template button after a template is applied". Both correctly verify the guard condition from both directions (manual add and template apply).

4. **Type safety (P2)**: The `as Parameters<typeof autoLayout>` casts have been replaced with properly typed `WorkflowStep[]` and `WorkflowTransition[]` local variables at lines 455-466 of `VisualWorkflowEditor.tsx`. Clean and readable.

### Implementation quality

The `applyTemplate` function (lines 428-501) is well-structured:
- Generates UUIDs for all steps upfront, builds nodes and edges, then delegates positioning to `autoLayout` — same algorithm used elsewhere in the visual editor.
- Properly fits viewport to the new layout via `computeFitToView` with a sensible fallback when the container ref is unavailable.
- The name-setting logic (`if (!name) setName(template.label)`) mirrors the original `WorkflowEditor` behavior exactly.
- Edge construction uses `localIds.slice(0, -1).map(...)` which correctly handles the single-step "Quick Fix" template (produces zero edges).

### Test coverage

13 template tests cover: create vs edit mode visibility, dropdown toggle, node population, edge creation, layout positions, start node assignment, name setting (both empty and pre-filled), dropdown close, overwrite protection (both manual-add and template-apply), and single-step template edge case. All 58 tests pass.

- The `TEMPLATES` array is now exported from `WorkflowEditor.tsx` and imported by both the visual editor and its test file. If templates grow or need to differ between the two editors in the future, extracting to a shared module would be cleaner — but for now this is fine and avoids premature abstraction.

No blocking issues found.

lsm (commented):

---

**Recommendation: APPROVE**

All four issues from the previous review round have been addressed correctly. Here is the detailed assessment of each fix plus a final check for any remaining issues.

---

### P1 Fixes — Verified

**Overwrite protection (nodes.length === 0 guard)**

The template picker button is now conditionally rendered only when `!isEditing && nodes.length === 0` (line 662 of `VisualWorkflowEditor.tsx`). This correctly prevents the button from appearing once the user has manually added nodes or after a template has been applied (since `applyTemplate` populates `nodes`). The guard is in the render tree, not behind a disabled state, so the DOM element is fully absent once there is canvas content — no chance of accidental interaction.

**Magic number in test**

The assertion `expect(options.length).toBe(TEMPLATES.length)` (test line 731) correctly imports `TEMPLATES` from `WorkflowEditor.tsx` rather than hardcoding `3`. If a fourth template is added in the future this test will continue to pass without modification.

---

### P2 Fixes — Verified

**Overwrite-protection tests**

Two tests were added to the `Template picker` describe block:
- "hides template button once nodes have been added (overwrite protection)" — clicks `add-step-button`, then asserts `template-picker-button` is absent (test lines 847–857).
- "hides template button after a template is applied" — applies a template, then asserts the button is gone (test lines 859–869).

Both tests are behaviorally meaningful and correctly exercise the `nodes.length === 0` guard via the component's rendered output, not internal state.

**Type safety**

The previous `as Parameters<typeof autoLayout>` casts have been replaced with properly typed locals:
- `const layoutSteps: WorkflowStep[]` — each element maps `n.step.localId` to `id`, which is a `string` satisfying the required `WorkflowStep.id: string` field.
- `const layoutTransitions: WorkflowTransition[]` — each element provides a synthetic `id: `t-${i}``, which satisfies the required `WorkflowTransition.id: string` field.

Both locals are correctly typed against the `@neokai/shared` interfaces used by `autoLayout`.

---

### Additional Observations (No blocking issues)

**Single-step template edge case**

The `localIds.slice(0, -1)` edge on a single-element array produces an empty slice, so `newEdges` will be `[]`. The single node is correctly set as the start node via `firstLocalId`. The test "single-step template (Quick Fix) creates one node and no edges" (lines 871–884) covers this case. No issue here.

**Viewport fallback in JSDOM**

In `applyTemplate`, when `canvasContainerRef.current` is null (always the case in JSDOM since `clientWidth`/`clientHeight` are 0), the code falls back to `{ offsetX: 0, offsetY: 0, scale: 1 }`. The node-position test (lines 772–790) inspects `node.style.left` to verify autoLayout ran, which is orthogonal to the viewport — the nodes receive their positions from `autoLayout` before the viewport reset, so the test is valid even without a real DOM container.

**`condition` field absent in `layoutTransitions`**

`WorkflowTransition.condition` is optional, so omitting it in the synthetic transitions passed to `autoLayout` is fine — `autoLayout` only reads `from` and `to`.

---

All previous P1 and P2 concerns have been fully resolved. No new P0/P1/P2 issues were found. The implementation is correct, complete, and well-tested.

PR #508, opened 2026-03-21
feat: add task_agent_session_id to space_tasks schema and repository

lsm:
- Migration 32: ALTER TABLE space_tasks ADD COLUMN task_agent_session_id TEXT
- Updated CREATE TABLE schema for fresh databases
- createTask() persists taskAgentSessionId when provided
- updateTask() can set/clear taskAgentSessionId
- rowToSpaceTask() maps task_agent_session_id to taskAgentSessionId
- getTaskBySessionId() looks up a task by its Task Agent session ID
- Updated space-test-db helper to include the new column
- Added unit tests covering create, update, get, and getTaskBySessionId

### DIFF
--- a/packages/daemon/src/storage/repositories/space-task-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-task-repository.ts
@@ -25,8 +25,8 @@ export class SpaceTaskRepository {
 		const now = Date.now();
 
 		const stmt = this.db.prepare(
-			`INSERT INTO space_tasks (id, space_id, title, description, status, priority, task_type, assigned_agent, custom_agent_id, workflow_run_id, workflow_step_id, created_by_task_id, depends_on, created_at, updated_at)
-       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
+			`INSERT INTO space_tasks (id, space_id, title, description, status, priority, task_type, assigned_agent, custom_agent_id, workflow_run_id, workflow_step_id, created_by_task_id, depends_on, task_agent_session_id, created_at, updated_at)
+       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
 		);
 
 		stmt.run(
@@ -43,6 +43,7 @@ export class SpaceTaskRepository {
 			params.workflowStepId ?? null,
 			params.createdByTaskId ?? null,
 			JSON.stringify(params.dependsOn ?? []),
+			params.taskAgentSessionId ?? null,
 			now,
 			now
 		);
@@ -206,6 +207,10 @@ export class SpaceTaskRepository {
 			fields.push('input_draft = ?');
 			values.push(params.inputDraft ?? null);
 		}
+		if (params.taskAgentSessionId !== undefined) {
+			fields.push('task_agent_session_id = ?');
+			values.push(params.taskAgentSessionId ?? null);
+		}
 
 		if (fields.length > 0) {
 			fields.push('updated_at = ?');
@@ -259,6 +264,18 @@ export class SpaceTaskRepository {
 	}
 
 	/**
+	 * Get a task by its Task Agent session ID
+	 */
+	getTaskBySessionId(sessionId: string): SpaceTask | null {
+		const stmt = this.db.prepare(
+			`SELECT * FROM space_tasks WHERE task_agent_session_id = ? LIMIT 1`
+		);
+		const row = stmt.get(sessionId) as Record<string, unknown> | undefined;
+		if (!row) return null;
+		return this.rowToSpaceTask(row);
+	}
+
+	/**
 	 * Get draft tasks created by a specific planning task
 	 */
 	getDraftTasksByCreator(createdByTaskId: string): SpaceTask[] {
@@ -294,6 +311,7 @@ export class SpaceTaskRepository {
 			dependsOn: JSON.parse(row.depends_on as string) as string[],
 			inputDraft: (row.input_draft as string | null) ?? undefined,
 			activeSession: (row.active_session as 'worker' | 'leader' | null) ?? null,
+			taskAgentSessionId: (row.task_agent_session_id as string | null) ?? undefined,
 			prUrl: (row.pr_url as string | null) ?? undefined,
 			prNumber: (row.pr_number as number | null) ?? undefined,
 			prCreatedAt: (row.pr_created_at as number | null) ?? undefined,
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -1671,6 +1671,9 @@ function runMigration29(db: BunDatabase): void {
 	db.exec(
 		`CREATE INDEX IF NOT EXISTS idx_space_tasks_workflow_step_id ON space_tasks(workflow_step_id)`
 	);
+	db.exec(
+		`CREATE INDEX IF NOT EXISTS idx_space_tasks_task_agent_session_id ON space_tasks(task_agent_session_id)`
+	);
 
 	// -------------------------------------------------------------------------
 	// space_session_groups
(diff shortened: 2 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly adds the `task_agent_session_id` column to the `space_tasks` table and updates all CRUD paths in `SpaceTaskRepository`. The implementation is correct and follows established patterns in the codebase.

### What was reviewed

- **Migration 32** (`packages/daemon/src/storage/schema/migrations.ts`): Idempotent ALTER TABLE using try/catch on a SELECT probe -- consistent with existing migration patterns (e.g., migration 31). The column is also added to the CREATE TABLE in migration 29 for fresh databases. Column placement is consistent between migration 29, migration 32, and the test helper.

- **Repository** (`packages/daemon/src/storage/repositories/space-task-repository.ts`):
  - `createTask()`: Column and placeholder counts match (16 columns, 16 `?` placeholders). `taskAgentSessionId` correctly defaults to `null` via `??`.
  - `updateTask()`: Follows the same `!== undefined` guard pattern as all other optional fields.
  - `rowToSpaceTask()`: Maps `null` to `undefined` via `?? undefined`, consistent with how other nullable fields (e.g., `inputDraft`, `prUrl`) are handled.
  - `getTaskBySessionId()`: Simple lookup with `LIMIT 1`, returns `null` when no match. The assumption that `task_agent_session_id` is unique per task is reasonable given the domain (one session per task).

- **Types** (`packages/shared/src/types/space.ts`): `taskAgentSessionId` is typed as `string | null` on `SpaceTask`, `CreateSpaceTaskParams`, and `UpdateSpaceTaskParams` -- all consistent with the repository implementation.

- **Test helper** (`packages/daemon/tests/unit/helpers/space-test-db.ts`): Column added in the same position as the migration schema.

- **Tests** (`packages/daemon/tests/unit/storage/space-task-repository.test.ts`): Good coverage of create (with/without), update (set/clear), and `getTaskBySessionId` (found, not found, multiple tasks). The tests verify real behavior through the repository, not mocks.

### Minor observations (P3)

1. No index on `task_agent_session_id` -- since `getTaskBySessionId` does a linear scan. This is fine for now given the expected table size, but worth adding if this becomes a hot path.

lsm (commented):

**Recommendation: APPROVE (with one P2 suggestion)**

---

### Summary

The implementation is correct and complete. The migration, repository changes, shared types, test helper, and unit tests are all consistent and well-integrated. One meaningful improvement opportunity exists: a missing index on the lookup column.

---

### P2 — Missing index on `task_agent_session_id` for `getTaskBySessionId`

**File:** `packages/daemon/src/storage/schema/migrations.ts`

The new `getTaskBySessionId(sessionId)` method issues a full table scan:

```sql
SELECT * FROM space_tasks WHERE task_agent_session_id = ? LIMIT 1
```

No index is created for `task_agent_session_id` in either the `runMigration32` function or the `CREATE TABLE` block for fresh databases. The existing space-tasks table already has five explicit indexes for other lookup columns (`space_id`, `status`, `workflow_run_id`, `custom_agent_id`, `workflow_step_id`), which establishes a clear project pattern that lookup columns should be indexed.

This will cause a full table scan on every `getTaskBySessionId` call in production. Given that session-to-task lookups are likely to happen on every active agent event, this is worth addressing before the call frequency grows.

**Suggested fix in `runMigration32`:**
```ts
db.exec(`ALTER TABLE space_tasks ADD COLUMN task_agent_session_id TEXT`);
db.exec(`CREATE INDEX IF NOT EXISTS idx_space_tasks_task_agent_session_id ON space_tasks(task_agent_session_id)`);
```

And in the `CREATE TABLE` block (after the existing five `CREATE INDEX` statements around line 1663):
```ts
db.exec(`CREATE INDEX IF NOT EXISTS idx_space_tasks_task_agent_session_id ON space_tasks(task_agent_session_id)`);
```

The test helper `space-test-db.ts` does not need to mirror indexes (it is in-memory), so no change needed there.

---

### What is correct

- Migration 32 follows the established try/catch column-existence-check pattern used by all prior migrations.
- The `CREATE TABLE` schema in the migrations file and the test helper `createSpaceTables` are both updated and are in sync.
- `createTask()` correctly passes `params.taskAgentSessionId ?? null` as the 14th positional parameter, and the positional binding in the INSERT matches the column list.
- `updateTask()` correctly uses an `undefined` check (not a falsy check) to support explicit `null` for clearing the field.
- `rowToSpaceTask()` maps DB `NULL` to `undefined` via `?? undefined`, consistent with every other nullable optional field in the mapper.
- The shared type `SpaceTask.taskAgentSessionId?: string | null` and `UpdateSpaceTaskParams.taskAgentSessionId?: string | null` are already in the repo from the previous PR and align with the new repository behavior.
- Unit tests cover all four cases: create-with, create-without, update-set, update-clear, and all three `getTaskBySessionId` scenarios (hit, miss, multi-task disambiguation).

lsm (commented):

**Recommendation: APPROVE**

### Previous P2 Issue — Resolved

The index on `task_agent_session_id` has been correctly added in both places:

1. **Migration path** (`runMigration32` in `packages/daemon/src/storage/schema/migrations.ts`, lines 1983–1985): `CREATE INDEX IF NOT EXISTS idx_space_tasks_task_agent_session_id ON space_tasks(task_agent_session_id)` runs unconditionally after the `ALTER TABLE`, ensuring existing databases get the index.

2. **Fresh-database path** (the `CREATE TABLE` block for `space_tasks`, lines 1672–1676): the same `CREATE INDEX IF NOT EXISTS` statement follows the table definition, so newly provisioned databases also get the index.

The `IF NOT EXISTS` guard on the index creation is correct — re-running the migration on a database that already has the index is safe.

### Remaining Code Review

Everything else looks correct:

- **`createTask`** includes `task_agent_session_id` as the 14th positional parameter (line 28–46) and binds `params.taskAgentSessionId ?? null` correctly.
- **`updateTask`** handles the field at lines 210–213 with the same nullable pattern used by every other optional column.
- **`rowToSpaceTask`** maps `row.task_agent_session_id` to `taskAgentSessionId` using the consistent `?? undefined` null-coercion pattern (line 314).
- **`getTaskBySessionId`** uses a prepared statement with `LIMIT 1` and returns `null` on miss — correctly mirrors `getTask`.
- **`space-test-db.ts`** helper has `task_agent_session_id TEXT` in the in-memory `CREATE TABLE`, keeping it in sync with the production schema. Note that the test helper does **not** create the index (there is no `CREATE INDEX` call after the `space_tasks` table definition), but this is acceptable for an in-memory test fixture where lookup performance is irrelevant.
- **Tests** cover all four behaviours called out in the task: create-with-value, create-without-value (undefined), update-set, update-clear, and the new `getTaskBySessionId` lookup (returns matching task, returns null on miss, returns correct task when multiple rows exist).

lsm (commented):

**Recommendation: APPROVE**

The previous review flagged a missing index on `task_agent_session_id`. That has been addressed -- the index is now created in both the fresh-database path (`runMigration29` / consolidated Space tables) and the incremental migration path (`runMigration32`). The implementation is correct and complete.

### What was reviewed

1. **Migration (`migrations.ts`)**: `runMigration32` correctly adds the `task_agent_session_id TEXT` column via `ALTER TABLE` with a try/catch idempotency guard, followed by `CREATE INDEX IF NOT EXISTS`. The fresh-database `CREATE TABLE` block in `runMigration29` also includes the column and the index. Both paths are consistent.

2. **Repository (`space-task-repository.ts`)**: The new field is properly wired through all CRUD operations:
   - `createTask` -- included in INSERT with `params.taskAgentSessionId ?? null`
   - `updateTask` -- guarded by `!== undefined` check, handles null clearing
   - `rowToSpaceTask` -- maps `task_agent_session_id` to `taskAgentSessionId` with null-to-undefined conversion
   - `getTaskBySessionId` -- new lookup method using the indexed column

3. **Test helper (`space-test-db.ts`)**: The test table schema includes the new column, keeping it in sync with the production schema.

4. **Tests (`space-task-repository.test.ts`)**: Good coverage of:
   - Create with and without `taskAgentSessionId`
   - Update to set and clear `taskAgentSessionId`
   - `getTaskBySessionId` -- match, no match, and multiple-task scenarios

No issues found. The fix for the previously flagged missing index is clean and correctly placed.

PR #509, opened 2026-03-21
test: add E2E tests for visual workflow editor (Task 7.1)

lsm:
Adds packages/e2e/tests/features/visual-workflow-editor.e2e.ts with 6 tests:
- Create workflow with visual editor (add 3 nodes, configure, set start, save)
- Node positions are restored after save and reopen (layout persistence)
- Load template in visual editor (auto-layout, node names, save)
- Toggle between List and Visual modes (aria-pressed, confirm dialog)
- Validation error: saving without name
- Validation error: saving without agent assigned

All actions through the UI; RPC only in before/afterEach infrastructure;
native confirm() dialogs handled via page.once('dialog').

### DIFF
--- a/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
+++ b/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
@@ -73,21 +73,24 @@ async function navigateToSpace(page: Page, spaceId: string): Promise<void> {
 	await page.waitForURL(`/space/${spaceId}**`, { timeout: 10000 });
 }
 
+/**
+ * Reset the workflow editor mode stored in localStorage to prevent test-ordering
+ * flakiness. SpaceIsland reads 'workflow-editor-mode' on mount; if a prior test
+ * left it at 'visual', subsequent tests start in visual mode unexpectedly.
+ */
+async function resetEditorModeStorage(page: Page): Promise<void> {
+	await page.evaluate(() => {
+		localStorage.removeItem('workflow-editor-mode');
+	});
+}
+
 /** Navigate to Workflows tab and open the workflow editor for a new workflow. */
 async function openNewWorkflowEditor(page: Page): Promise<void> {
 	// Click Workflows tab
 	await page.locator('text=Workflows').first().click();
-	await expect(
-		page.locator('text=New Workflow').or(page.locator('text=Create Workflow'))
-	).toBeVisible({
-		timeout: 5000,
-	});
 
-	// Click "New Workflow" / "Create Workflow" button
-	const createBtn = page
-		.getByRole('button', { name: 'New Workflow' })
-		.or(page.getByRole('button', { name: 'Create Workflow' }))
-		.first();
+	// Wait for the "Create Workflow" button (the only label used in WorkflowList.tsx)
+	const createBtn = page.getByRole('button', { name: 'Create Workflow' });
 	await expect(createBtn).toBeVisible({ timeout: 5000 });
 	await createBtn.click();
 
@@ -126,6 +129,8 @@ test.describe('Visual Workflow Editor', () => {
 
 	test.beforeEach(async ({ page }) => {
 		await page.goto('/');
+		// Clear persisted editor mode so tests are independent of run order
+		await resetEditorModeStorage(page);
 		spaceId = await createTestSpace(page);
 	});
 
@@ -158,13 +163,17 @@ test.describe('Visual Workflow Editor', () => {
 		const nodes = editor.locator('[data-testid^="workflow-node-"]');
 		await expect(nodes).toHaveCount(3, { timeout: 3000 });
 
-		// Configure node 1: name + agent
+		// Configure node 1: name + agent.
+		// The first added node is auto-designated as start (VisualWorkflowEditor addStep).
 		await nodes.nth(0).click();
 		await expect(editor.getByTestId('node-config-panel')).toBeVisible({ timeout: 3000 });
 		await editor.getByTestId('step-name-input').fill('Planner');
 		await editor.getByTestId('agent-select').selectOption({ index: 1 });
 		await editor.getByTestId('close-button').click();
 		await expect(editor.getByTestId('node-config-panel')).not.toBeVisible({ timeout: 2000 });
+
+		// Verify node 1 shows the start badge (it is the current start node)
+		await expect(nodes.nth(0).getByTestId('start-badge')).toBeVisible({ timeout: 2000 });
 
 		// Configure node 2: name + agent
 		await nodes.nth(1).click();
@@ -179,17 +188,17 @@ test.describe('Visual Workflow Editor', () => {
 		await editor.getByTestId('step-name-input').fill('Reviewer');
 		await editor.getByTestId('agent-select').selectOption({ index: 1 });
 
-		// Set node 3 as the start node
+		// "Set as Start" only renders when !isStartNode — assert it is visible before clicking
+		await expect(editor.getByTestId('set-as-start-button')).toBeVisible({ timeout: 2000 });
 		await editor.getByTestId('set-as-start-button').click();
 
-		// Start badge should appear on node 3 (the one we're editing)
-		// After set-as-start, node 3's badge should show
+		// Start badge should now appear on node 3
 		await expect(nodes.nth(2).getByTestId('start-badge')).toBeVisible({ timeout: 3000 });
 
 		// Close config panel
 		await editor.getByTestId('close-button').click();
 
-		// Node 1 should no longer show start badge
+		// Node 1 should no longer show start badge after reassignment
 		await expect(nodes.nth(0).getByTestId('start-badge')).not.toBeVisible({ timeout: 2000 });
 
 		// Save the workflow
@@ -240,9 +249,18 @@ test.describe('Visual Workflow Editor', () => {
 		await page.locator('text=Workflows').first().click();
 		await expect(page.locator('text=Layout Persist Test')).toBeVisible({ timeout: 5000 });
 
-		// Hover over the workflow card to reveal the Edit button
-		const workflowCard = page.locator('text=Layout Persist Test').first();
-		await workflowCard.hover();
+		// The Edit button is hidden under opacity-0 group-hover:opacity-100.
+		// Force it visible via JS before clicking — Tailwind CSS group-hover
+		// is unreliable in headless Chromium under xvfb.
+		const workflowCard = page
+			.locator('[class*="group"]')
+			.filter({ has: page.locator('text=Layout Persist Test') })
+			.first();
+		await expect(workflowCard).toBeVisible({ timeout: 3000 });
+		await workflowCard.evaluate((el) => {
+			const actions = el.querySelector<HTMLElement>('.opacity-0');
+			if (actions) actions.style.opacity = '1';
+		});
 
 		// Click the Edit button
 		const editBtn = page.getByRole('button', { name: 'Edit' }).first();
@@ -298,11 +316,9 @@ test.describe('Visual Workflow Editor', () => {
 			timeout: 3000,
 		});
 
-		// Select the "Coding" template
-		const codingTemplate = editor
-			.locator('[data-testid="template-option"]')
-			.filter({ hasText: 'Coding' })
-			.first();
+		// Select the Coding template by its exact data-template-label attribute
+		// (more precise than partial text match — avoids false matches if new templates added)
+		const codingTemplate = editor.locator('[data-template-label="Coding (Plan → Code)"]');
 		await expect(codingTemplate).toBeVisible({ timeout: 3000 });
 		await codingTemplate.click();
 
@@ -349,17 +365,17 @@ test.describe('Visual Workflow Editor', () => {
 		await navigateToSpace(page, spaceId);
 		await openNewWorkflowEditor(page);
 
-		// Verify List mode is active by default
+		// Verify List mode is active by default (localStorage was cleared in beforeEach)
 		await expect(page.getByTestId('editor-mode-list')).toHaveAttribute('aria-pressed', 'true');
 		await expect(page.getByTestId('editor-mode-visual')).toHaveAttribute('aria-pressed', 'false');
 
 		// List mode shows WorkflowEditor UI (has template/step controls)
-		await expect(
-			page.locator('text=Start from template').or(page.locator('text=Add Step')).first()
-		).toBeVisible({ timeout: 3000 });
+		await expect(page.getByRole('button', { name: /Start from template/ })).toBeVisible({
+			timeout: 3000,
+		});
 
 		// Add steps in List mode using a template
-		await page.locator('text=Start from template').click();
+		await page.getByRole('button', { name: /Start from template/ }).click();
 		await page.locator('text=Coding (Plan → Code)').click();
 
 		// Verify steps appear in List mode
--- a/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
+++ b/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
@@ -1,0 +1,440 @@
+/**
+ * Visual Workflow Editor E2E Tests
+ *
+ * Tests:
+ * - Create workflow with visual editor (add nodes, configure properties, set start node, save)
+ * - Node positions are restored after save and reopen (layout persistence)
+ * - Load template in visual editor (auto-layout with nodes and edges)
+ * - Toggle between List and Visual modes (mode switching with confirmation)
+ * - Validation errors when saving incomplete workflows
+ *
+ * Setup: creates a Space via RPC in beforeEach (infrastructure).
+ * Cleanup: deletes the Space via RPC in afterEach.
+ *
+ * E2E Rules:
+ * - All test actions go through the UI (clicks, inputs, navigation)
+ * - All assertions check visible DOM state
+ * - RPC is only used in beforeEach/afterEach for test infrastructure
+ */
+
+import type { Page } from '@playwright/test';
+import { test, expect } from '../../fixtures';
+import { waitForWebSocketConnected, getWorkspaceRoot } from '../helpers/wait-helpers';
+
+const DESKTOP_VIEWPORT = { width: 1440, height: 900 };
+
+// ─── RPC helpers (infrastructure only) ────────────────────────────────────────
+
+async function createTestSpace(page: Page): Promise<string> {
+	await waitForWebSocketConnected(page);
+	const workspaceRoot = await getWorkspaceRoot(page);
+	const spaceName = `E2E Visual Editor Test ${Date.now()}`;
+	return page.evaluate(
+		async ({ wsPath, name }) => {
+			const hub = window.__messageHub || window.appState?.messageHub;
+			if (!hub?.request) throw new Error('MessageHub not available');
+
+			// Clean up any leftover space at this workspace path
+			try {
+				const list = (await hub.request('space.list', {})) as Array<{
+					id: string;
+					workspacePath: string;
+				}>;
+				const existing = list.find((s) => s.workspacePath === wsPath);
+				if (existing) {
+					await hub.request('space.delete', { id: existing.id });
+				}
+			} catch {
+				// Ignore cleanup errors
+			}
+
+			const res = await hub.request('space.create', { name, workspacePath: wsPath });
+			return (res as { id: string }).id;
+		},
+		{ wsPath: workspaceRoot, name: spaceName }
+	);
+}
+
+async function deleteTestSpace(page: Page, spaceId: string): Promise<void> {
+	if (!spaceId) return;
+	try {
+		await page.evaluate(async (id) => {
+			const hub = window.__messageHub || window.appState?.messageHub;
+			if (!hub?.request) return;
+			await hub.request('space.delete', { id });
+		}, spaceId);
+	} catch {
+		// Best-effort cleanup
+	}
+}
+
+async function navigateToSpace(page: Page, spaceId: string): Promise<void> {
+	await page.goto(`/space/${spaceId}`);
+	await page.waitForURL(`/space/${spaceId}**`, { timeout: 10000 });
+}
+
+/** Navigate to Workflows tab and open the workflow editor for a new workflow. */
+async function openNewWorkflowEditor(page: Page): Promise<void> {
+	// Click Workflows tab
+	await page.locator('text=Workflows').first().click();
+	await expect(
+		page.locator('text=New Workflow').or(page.locator('text=Create Workflow'))
+	).toBeVisible({
+		timeout: 5000,
+	});
+
+	// Click "New Workflow" / "Create Workflow" button
+	const createBtn = page
+		.getByRole('button', { name: 'New Workflow' })
+		.or(page.getByRole('button', { name: 'Create Workflow' }))
+		.first();
+	await expect(createBtn).toBeVisible({ timeout: 5000 });
+	await createBtn.click();
+
+	// Wait for editor mode toggle strip to appear
+	await expect(page.getByTestId('editor-mode-toggle')).toBeVisible({ timeout: 5000 });
+}
+
+/** Switch to Visual editor mode, accepting any confirmation dialogs. */
+async function switchToVisualMode(page: Page): Promise<void> {
+	// Register dialog handler before clicking — native confirm() fires synchronously
+	page.once('dialog', (d) => d.accept());
+	await page.getByTestId('editor-mode-visual').click();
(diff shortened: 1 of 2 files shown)


lsm (commented):

This is a well-structured test file with good use of data-testid attributes, proper use of RPC only in infrastructure helpers, and sensible use of `page.once('dialog')` for native confirms. However, there are several correctness issues that would cause tests to fail against the real codebase.

---

### P0 — Incorrect button text: "New Workflow" does not exist in the UI

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 81–95

The `openNewWorkflowEditor` helper waits for `text=New Workflow` OR `text=Create Workflow`, then tries to click a button named "New Workflow" or "Create Workflow". In the actual `WorkflowList` component (`packages/web/src/components/space/WorkflowList.tsx`, line 404), the button is labelled **"Create Workflow"** only. There is no "New Workflow" button anywhere in the workflow list view. The fallback `or` will eventually reach "Create Workflow", but the 5-second timeout on line 84 waiting for `text=New Workflow` to appear will always exhaust before the `or` is evaluated. This will slow each test by 5 seconds and risks a timeout failure.

---

### P1 — localStorage mode persists across tests, causing ordering-dependent flakiness

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, all tests

`SpaceIsland.tsx` reads `workflow-editor-mode` from `localStorage` on every mount (line 44–51 via `readStoredEditorMode()`). Test 4 ("Toggle between List and Visual modes") switches to visual mode and this preference is persisted via `localStorage.setItem(EDITOR_MODE_KEY, mode)`. When Tests 5 and 6 subsequently run in the same browser context, `SpaceIsland` mounts in visual mode. The `switchToVisualMode` helper then registers a `page.once('dialog')` handler and clicks the visual toggle. Since the mode is already visual, `handleSetEditorMode` returns early (line 88: `if (mode === editorMode) return`) — no dialog fires, but the registered handler stays alive and will accept the next dialog that appears unexpectedly. There is no `beforeEach` or `afterEach` that clears `localStorage` to reset the editor mode.

---

### P1 — Hover-based Edit button reveal is fragile in headless CI

**File:** lines 243–250

```ts
const workflowCard = page.locator('text=Layout Persist Test').first();
await workflowCard.hover();
const editBtn = page.getByRole('button', { name: 'Edit' }).first();
await expect(editBtn).toBeVisible({ timeout: 3000 });
await editBtn.click();
```

In `WorkflowCard` (`WorkflowList.tsx`, line 171), the action buttons have `opacity-0 group-hover:opacity-100`. They only become visible via CSS hover on the card. Playwright's `hover()` triggers synthetic mouse events, but Tailwind `group-hover` visibility driven purely by CSS `:hover` state is known to be unreliable in headless Chromium running under xvfb. The element may remain at `opacity-0` even after `hover()`, causing `toBeVisible` to time out intermittently.

---

### P1 — `set-as-start-button` renders only when `!isStartNode` — Test 1 never verifies the pre-condition

**File:** lines 177–187

Test 1 clicks `set-as-start-button` while node 2's config panel is open. This button only renders in `NodeConfigPanel` when `!isStartNode` (line 160 of `NodeConfigPanel.tsx`). The test never asserts that the button is visible before clicking it. If the first node added was somehow recognized as node index 2 (e.g. due to DOM order), the button would be absent and the test would silently mis-click. Adding `await expect(editor.getByTestId('set-as-start-button')).toBeVisible()` before clicking would make the pre-condition explicit and catch any regression.

---

### P2 — Template filter uses partial text; full label match would be more precise

**File:** line 301–306

```ts
const codingTemplate = editor
  .locator('[data-testid="template-option"]')
  .filter({ hasText: 'Coding' })
  .first();
```

The actual template label is `'Coding (Plan → Code)'`. The partial match works today, but if a future template is added that also contains "Coding", the wrong template could be selected. Since the component also sets `data-template-label={t.label}` (line 700 of `VisualWorkflowEditor.tsx`), using `filter({ has: page.locator('[data-template-label="Coding (Plan → Code)"]') })` would be unambiguous.

---

### P2 — Test 4 uses a text-substring locator for the List-mode template toggle

**File:** lines 362–363

```ts
await page.locator('text=Start from template').click();
await page.locator('text=Coding (Plan → Code)').click();
```

`WorkflowEditor.tsx` renders the button as `'Start from template ↓'` or `'Hide templates ↑'`. Playwright's `text=` with substring matching will find it, but it also means this locator would match any other element on the page containing those words. Using `getByRole('button', { name: /Start from template/ })` would be more targeted.

---

### P3 — Dead code: "New Workflow" branch in `openNewWorkflowEditor`

**File:** lines 87–92

The `or(page.getByRole('button', { name: 'New Workflow' }))` branch on the button locator is dead code because this text never appears in the component tree. It should be removed to avoid misleading future readers.

---

### P3 — Missing assertion that the `start-badge` was visible on node 0 before reassigning

**File:** lines 186–193

Test 1 asserts that node 0 does NOT show `start-badge` after designating node 2 as start. However, it never asserts that node 0 DID show `start-badge` before the reassignment. Adding a "was-visible" assertion before clicking `set-as-start-button` would make the test a meaningful round-trip verification.

---

### What is correct

- All `data-testid` attributes used in assertions (`visual-workflow-editor`, `editor-mode-toggle`, `editor-mode-list`, `editor-mode-visual`, `add-step-button`, `workflow-name-input`, `workflow-node-`, `node-config-panel`, `step-name-input`, `agent-select`, `close-button`, `set-as-start-button`, `start-badge`, `save-button`, `template-picker-button`, `template-option`) all exist in the actual component implementations.
- `aria-pressed` assertions in Test 4 correctly match how Preact serializes boolean attributes.
- Error messages checked in Tests 5 and 6 (`'Workflow name is required.'` and `'requires an agent'`) match the exact strings in `VisualWorkflowEditor.tsx` lines 509 and 521.
- `page.once('dialog')` pattern is correctly placed before the click that triggers the native `confirm()` dialog.
- RPC is only used in `beforeEach`/`afterEach` infrastructure and the Test 2 setup block, in compliance with CLAUDE.md rules.

lsm (commented):

**Recommendation: APPROVE**

This is a well-structured E2E test file that exercises the visual workflow editor through six meaningful test scenarios. I verified all selectors, UI flows, and assertions against the actual component implementations.

### Verification Summary

**Selectors**: All `data-testid` selectors used in the test match real attributes in the component source:
- `visual-workflow-editor`, `editor-mode-toggle`, `editor-mode-list`, `editor-mode-visual` in `SpaceIsland.tsx`
- `workflow-name-input`, `add-step-button`, `save-button`, `template-picker-button`, `template-option` in `VisualWorkflowEditor.tsx`
- `node-config-panel`, `step-name-input`, `agent-select`, `close-button`, `set-as-start-button` in `NodeConfigPanel.tsx`
- `workflow-node-{id}`, `start-badge` in `WorkflowNode.tsx`

**UI Flow Correctness**:
- `aria-pressed` attributes are correctly set on the mode toggle buttons in `SpaceIsland.tsx` (lines 178, 192)
- The `confirm()` dialog fires when `workflowEditId !== null` (line 91 of SpaceIsland.tsx), which is the case after `openNewWorkflowEditor` sets it to `'new'`. The `page.once('dialog', d => d.accept())` handler is correctly registered before the click
- Template labels: "Coding" substring filter matches `'Coding (Plan -> Code)'` from `TEMPLATES` in `WorkflowEditor.tsx`. Template step names "Planner"/"Coder" match the `role.charAt(0).toUpperCase() + role.slice(1)` transform in `applyTemplate`
- First added node auto-becomes start (line 255 of VisualWorkflowEditor.tsx), so "Set as Start Node" button is correctly visible on node 3 but not node 1
- Validation messages `'Workflow name is required.'` and `'Step 1 requires an agent.'` match the actual `setError` calls in the component
- Agent select `index: 1` works because `seedPresetAgents` creates planner/coder/general/reviewer agents on `space.create`

**E2E Guidelines Compliance**: All test actions go through the UI. RPC is only used in `beforeEach`/`afterEach` for space setup/teardown and in Test 2 for creating a workflow with known layout positions (legitimate infrastructure setup). No internal state assertions, no `page.context().setOffline()`, no sleep calls.

**Wait Conditions**: All assertions use `waitFor`-style expects with explicit timeouts. No `sleep` or `waitForTimeout` calls.

### Minor Observations (P3)

1. **Test 2 uses RPC to create a workflow** (lines 211-237) rather than creating through the UI. This is reasonable since the test's purpose is to verify position persistence on reopen, not the creation flow (which Test 1 covers). However, it does mean the test won't catch regressions in the save-layout path itself -- only the load-layout path. This is an acceptable trade-off for test isolation.

2. **`selectOption({ index: 1 })` assumes agent ordering** -- if the agent list order changes (e.g., alphabetical vs. creation order), index 1 might select a different agent. This is unlikely to cause failures since any agent satisfies validation, but using `{ label: /planner/i }` would be more explicit. Very minor.

3. **Test 4 relies on `text=Start from template` matching `'Start from template ...'`** via substring. This is correct Playwright behavior but could be more precise with an exact selector.

Overall this is a solid, well-written E2E test file that follows the project's conventions and should be reliable.

lsm (commented):

**Recommendation: APPROVE**

All six issues from the previous review have been correctly addressed:

1. **P0 fix (button text):** `openNewWorkflowEditor` now uses `getByRole('button', { name: 'Create Workflow' })` directly, matching the actual label in `WorkflowList.tsx` line 415. The dead "New Workflow" fallback branch is gone.

2. **P1 fix (localStorage flakiness):** `resetEditorModeStorage()` helper clears `workflow-editor-mode` from localStorage in `beforeEach`, ensuring each test starts in list mode regardless of execution order.

3. **P1 fix (Edit button hover):** Test 2 now uses `workflowCard.evaluate()` to force `opacity: 1` on the actions container via JS, avoiding the unreliable CSS `:hover` in headless Chromium. The card locator `[class*="group"]` with a filter for the workflow name text is a reasonable selector given `WorkflowList.tsx` line 171 uses `opacity-0 group-hover:opacity-100`.

4. **P1 fix (set-as-start pre-condition):** Line 192 now explicitly asserts `set-as-start-button` is visible before clicking, confirming the pre-condition that node 2 is not the start node. Line 176 also verifies node 0 has the start badge before reassignment, addressing the P3 suggestion about round-trip verification.

5. **P2 fix (template selection):** Line 321 uses `[data-template-label="Coding (Plan → Code)"]` attribute selector, matching the `data-template-label={t.label}` attribute set in `VisualWorkflowEditor.tsx` line 700.

6. **P2 fix (list-mode template button):** Line 373/378 uses `getByRole('button', { name: /Start from template/ })` instead of a bare text locator.

**Verification of selector correctness against source components:**

All `data-testid` values used in the test match their corresponding component implementations:
- `visual-workflow-editor` -- `VisualWorkflowEditor.tsx:572`
- `editor-mode-toggle`, `editor-mode-list`, `editor-mode-visual` -- `SpaceIsland.tsx:172,177,191`
- `workflow-node-*`, `start-badge` -- `WorkflowNode.tsx:220,271`
- `node-config-panel`, `step-name-input`, `agent-select`, `set-as-start-button`, `close-button` -- `NodeConfigPanel.tsx:97,148,174,162,126`
- `add-step-button`, `save-button`, `template-picker-button`, `template-option`, `workflow-name-input` -- `VisualWorkflowEditor.tsx:647,624,666,699,601`
- `aria-pressed` attributes -- `SpaceIsland.tsx:178,192`

Validation error strings match exactly:
- `'Workflow name is required.'` -- `VisualWorkflowEditor.tsx:509`
- `'requires an agent'` (substring) -- `VisualWorkflowEditor.tsx:520`

The test file follows E2E rules from CLAUDE.md: all actions go through the UI, all assertions check visible DOM state, and RPC is confined to `beforeEach`/`afterEach` infrastructure (with the acceptable exception of Test 2's workflow creation setup).

No remaining issues found.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — two P0 bugs and one P1 remain after the previous fixes.

---

### P0 — `start-badge` assertion is on the wrong element (Tests 1 and 2)

The test asserts `nodes.nth(0).getByTestId('start-badge')` (Test 1, lines 176, 202) and `nodeWithStepOne.getByTestId('start-badge')` (Test 2, line 291). `start-badge` lives inside `WorkflowNode.tsx` and is rendered on the canvas node element — that part is correct and the test ID is right.

However, the test also checks `nodes.nth(0).getByTestId('start-badge')` while the NodeConfigPanel is **open** (lines 176 and 196). When a node is selected the `NodeConfigPanel` is rendered _outside_ the node elements (it is an absolutely-positioned sibling panel, not a child of the node). The NodeConfigPanel renders its own badge with `data-testid="start-node-badge"` (not `start-badge`). The canvas `start-badge` is still the `<span>` inside the `WorkflowNode` component, so the assertion is technically correct — but only if the node is in the DOM and Playwright's locator for `nodes.nth(0)` still resolves. This should work, but the close-panel / re-check sequence (lines 197–202) is fragile: after clicking close, the config panel disappears but the node DOM order may not be stable during a re-render, meaning `nodes.nth(0)` could briefly match the wrong node. The safer pattern is to filter by step name text rather than index after state changes.

More critically for Test 1 line 196, the assertion `nodes.nth(2).getByTestId('start-badge')` is checked _before_ `close-button` is clicked (line 199). The config panel is still open, overlapping node 2. If the panel occludes the canvas the assertion may time out because the badge element is visually covered (Playwright's `toBeVisible` checks both attachment and visibility in the rendering tree). This is a real failure risk in CI with the 320px-wide slide-in panel.

**Recommendation:** assert `start-badge` only _after_ the config panel is closed. Move lines 196 and 202 to after the close-button click.

---

### P0 — `resetEditorModeStorage` runs after `page.goto('/')` but before the app bootstraps

In `beforeEach` (lines 131–135):
```ts
await page.goto('/');
await resetEditorModeStorage(page); // removeItem
spaceId = await createTestSpace(page);  // calls waitForWebSocketConnected which waits for hub
```

`page.goto('/')` navigates, then `resetEditorModeStorage` immediately calls `localStorage.removeItem`. However, `SpaceIsland` reads `localStorage.getItem(EDITOR_MODE_KEY)` on _mount_ — which happens after navigation to `/space/${spaceId}` later in the test, not during `page.goto('/')`. So clearing on the home page is fine for _this_ visit. But there is a subtle ordering issue: `resetEditorModeStorage` is called before `waitForWebSocketConnected` (which is the first call inside `createTestSpace`). At the moment `page.evaluate` runs, the Preact app may not have started yet and `window.__messageHub` may be undefined. However, `localStorage.removeItem` runs in the browser synchronously and does not rely on the hub, so this particular race is harmless.

The real race is subtler: after `page.goto('/')` the browser may still be evaluating the Preact bundle. If a prior test left `workflow-editor-mode = visual` in localStorage, and the `SpaceIsland` component mounts _before_ `resetEditorModeStorage` runs (because `page.goto` resolves on `load` event but Preact's initial signal read happens earlier), the stale value would already be consumed. In practice `goto('/')` does not navigate to a space, so `SpaceIsland` is not mounted at that point and the clear is safe. Still, the comment in the code (line 79) says "SpaceIsland reads 'workflow-editor-mode' on mount" — the clear should happen before the space navigation, which it does. But `resetEditorModeStorage` must also be called _before_ a reload if the test later causes one. There is no reload path in these tests, so this is P2 rather than P0. Downgrading.

After re-examination the actual P0 is different: `resetEditorModeStorage` runs `page.evaluate` which requires the page JavaScript context to exist. Right after `page.goto('/')` the context should be available. This is fine.

**Re-classification: the reset ordering issue is P2 (test reads fresh storage on each space navigation).**

---

### P0 — Validation error text mismatch (Test 5, line 423)

The test asserts:
```ts
await expect(page.locator('text=Workflow name is required')).toBeVisible({ timeout: 3000 });
```

But `VisualWorkflowEditor.tsx` line 509 sets:
```ts
setError('Workflow name is required.');
```

Note the trailing **period** in the source. Playwright's `text=` selector with an exact string will not match `"Workflow name is required."` when given `"Workflow name is required"` (no period). Playwright's `text=` locator does a **substring** match by default (case-insensitive, leading/trailing whitespace ignored), so `"Workflow name is required"` _will_ match the substring of `"Workflow name is required."`. This is actually fine — confirmed safe.

**Re-classification: not a bug.**

---

### P1 — Test 4 uses `page.locator('text=2 steps')` which is ambiguous and order-dependent

After applying the Coding template in list mode (Test 4, line 382), the test asserts:
```ts
await expect(page.locator('text=2 steps')).toBeVisible({ timeout: 3000 });
```

The WorkflowEditor renders `{steps.length} step{steps.length !== 1 ? 's' : ''}` inside a `<span class="text-xs text-gray-600">`. The string `"2 steps"` should appear. However, `page.locator('text=2 steps')` is a substring match and will match any element containing that text — including page titles, descriptions, or any other text node that happens to contain "2 steps". This is unlikely to cause a false positive given the context, but it is not scoped to the editor and would pass even if the text appeared elsewhere on the page. Prefer `editor.locator('text=2 steps')` or `page.locator('.text-gray-600', { hasText: '2 steps' })`.

Also in this same test (line 379), the template is selected via:
```ts
await page.locator('text=Coding (Plan → Code)').click();
```
This works in list mode because the template button renders the label as visible text. But `page.locator('text=...')` will match any element; if the toggle strip or any other element briefly shows that text, it could match the wrong element. This is unlikely but a scoped locator is safer.

---

### P1 — Edit button reveal via `querySelector('.opacity-0')` is implementation-coupled and fragile

Test 2 (lines 260–263):
```ts
await workflowCard.evaluate((el) => {
    const actions = el.querySelector<HTMLElement>('.opacity-0');
    if (actions) actions.style.opacity = '1';
});
```

The selector `.opacity-0` matches the _first_ element inside the card that has the Tailwind class `opacity-0`. This works today because only one child has that class. But:

1. If Tailwind's JIT removes `opacity-0` from the class list at runtime (it uses `group-hover:opacity-100` to transition), the class may not be present in the final rendered HTML.
2. Any future component change that adds another `opacity-0` element (e.g., a tooltip, a loading state) will silently make this target the wrong element.

The approach was accepted in the previous review as a pragmatic workaround for headless-hover unreliability, but `querySelector('.opacity-0')` is more fragile than `querySelector('[class*="opacity-0"]')` or querying by the known structure (e.g., the flex container that holds the Edit/Delete buttons). The comment says this is because "Tailwind CSS group-hover is unreliable in headless Chromium under xvfb." The correct production fix is to add a `data-testid="workflow-card-actions"` to the action div and target that.

This is P1: it will silently break if the opacity class is removed or a second element gets it.

---

### P2 — `resetEditorModeStorage` is called after `page.goto('/')` instead of before `navigateToSpace`

The intent is to prevent stale editor mode from affecting SpaceIsland on mount. The clear happens on the home page, which is fine because SpaceIsland is not mounted there. But if a test navigates to a space URL directly (e.g., on retry or via deep link), the clear would need to happen before that navigation, not before the home page load. Consider moving it to just before `navigateToSpace` for clarity of intent, or document why placing it at home-page-load is correct.

---

### P2 — Test 2 creates a workflow via direct RPC with `spaceWorkflow.create` but does not verify the `layout` field round-trips

The test creates a workflow with a `layout` field and then opens it in the visual editor, expecting node positions to be restored. However, the test does not assert anything about node _positions_ — it only checks that node names and the start badge are visible (lines 284–291). The test is named "Node positions are restored after save and reopen" but doesn't verify positions. The position-restoration logic is entirely unverified by this test, making it a name/behavior mismatch. This is at minimum P2.

---

### Summary

The two most impactful remaining issues are:

1. **P1 (fragile):** Test 1 checks `start-badge` while the config panel is open and may overlap or occlude the node — move these assertions to after `close-button.click()`.
2. **P1 (fragile):** `.opacity-0` querySelector in Test 2 is coupled to Tailwind class application at runtime and will silently target the wrong element if the class changes.
3. **P2:** "Node positions are restored" test verifies only node names and start badge, not the actual positions it claims to test.

lsm (commented):

**Recommendation: APPROVE**

All three issues from the previous review round have been correctly addressed:

### Fix 1: Start-badge assertions moved after config panel close (P1 fix)

In Test 1, the start-badge assertions now appear only after `close-button.click()` and the panel is confirmed hidden. Lines 172-183 close the panel for node 1 before checking its start badge. Lines 204-221 close the panel for node 3 before checking both "Reviewer" (should have badge) and "Planner" (should not). This eliminates the risk of the 320px NodeConfigPanel occluding the badge during visibility checks.

### Fix 2: Name-based node filtering replaces index-based filtering (P1 fix)

All start-badge assertions now use `.filter({ hasText: 'Planner' })` and `.filter({ hasText: 'Reviewer' })` instead of `nodes.nth(0)` / `nodes.nth(2)`. This is resilient to DOM reordering during re-renders.

### Fix 3: `data-testid="workflow-card-actions"` added to WorkflowList.tsx (P1 fix)

The action div in `WorkflowCard` (WorkflowList.tsx line 172) now has `data-testid="workflow-card-actions"`. Test 2 (line 280) uses `querySelector('[data-testid="workflow-card-actions"]')` to force opacity, replacing the fragile `.opacity-0` class selector. This is decoupled from Tailwind implementation details.

### Fix 4: Bounding box comparison for position verification (P1/P2 fix)

Test 2 (lines 315-327) now uses `boundingBox()` on both nodes and asserts `Math.abs(boxTwo!.x - boxOne!.x) > 50`. Given the saved layout positions of x=100 and x=450, the 50px threshold is conservative and accounts for canvas transforms/scaling. The test now genuinely verifies position persistence, not just node existence.

### Verification against source components

All selectors remain correct:
- `data-testid="start-badge"` -- `WorkflowNode.tsx:271`
- `data-testid="workflow-card-actions"` -- `WorkflowList.tsx:172`
- `data-testid="workflow-node-*"` -- `WorkflowNode.tsx:220`
- Validation error strings match: `'Workflow name is required.'` (line 509) and `'requires an agent'` substring of `'Step N requires an agent.'` (line 520) in `VisualWorkflowEditor.tsx`

No remaining issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### P1 Fix Verification

All three P1 issues from the previous review were correctly resolved.

#### P1-1: Start-badge assertions moved to after panel close; name-based filtering

Confirmed fixed. The test now calls `editor.getByTestId('close-button').click()` and waits for `node-config-panel` not to be visible before asserting `start-badge`. All start-badge assertions now use `.filter({ hasText: 'Planner' })` and `.filter({ hasText: 'Reviewer' })` instead of brittle index-based `.nth(0)`/`.nth(2)`. The `not.toBeVisible()` check on a conditionally-rendered element (badge is absent from DOM when `isStartNode` is false) behaves correctly — Playwright treats a missing element as not visible.

#### P1-2: `data-testid="workflow-card-actions"` added to WorkflowList.tsx

Confirmed fixed. The div at line 172 of `WorkflowList.tsx` now has `data-testid="workflow-card-actions"`. The test's `el.querySelector<HTMLElement>('[data-testid="workflow-card-actions"]')` call targets this element precisely instead of the previous `.opacity-0` class selector.

#### P1-3: Bounding-box position assertion added

Confirmed fixed. Test 2 now calls `nodeOne.boundingBox()` and `nodeTwo.boundingBox()` and asserts `Math.abs(boxTwo.x - boxOne.x) > 50`. The saved layout positions (x=100 vs x=450) produce a 350px canvas-space delta, and at the default scale=1 the screen delta is also ~350px. The 50px threshold is conservative and will detect the case where both nodes collapse to the same position.

---

**Node locator duplication in Test 2 (lines 307-320):** `nodeWithStepOne` and `nodeOne` are separate locator declarations that select the same element. `nodeWithStepOne` is used only for the start-badge check, then `nodeOne` is declared again for bounding-box. This is a minor readability nit — they could share a single declaration. Not a correctness issue.

**Viewport interaction in position assertion:** The canvas applies a CSS `translate(offsetX, offsetY) scale(scale)` transform from `VisualCanvas.tsx:246`. At the default viewport state (offsetX=0, offsetY=0, scale=1), `boundingBox()` returns the correct layout positions. If a viewport fit-to-view were triggered after load, the absolute screen positions could differ but the relative delta would still be proportional to scale. The assertion correctly checks only relative separation (> 50px), making it resilient to any reasonable viewport offset.

**`getByRole('button', { name: 'Edit' }).first()` in Test 2 (line 285):** If additional "Edit" buttons exist elsewhere on the page (e.g., from a different tab), `.first()` could match the wrong button. The opacity-forcing step immediately before is scoped to the correct card, so the Edit button inside the now-visible container should be the first visible one in DOM order. This is low risk in practice given the test structure.

---

All P0/P1/P2 issues have been resolved. The test correctly exercises create, layout persistence, template loading, mode toggling, and validation — all through UI interactions as required by the E2E rules.

PR #513, opened 2026-03-21
feat: implement Task Agent MCP tool handlers (Task 3.1)

lsm:
Create createTaskAgentToolHandlers() with 5 handlers: spawn_step_agent,
check_step_status, advance_workflow, report_result, and request_human_input.

- spawn_step_agent: resolves agent via resolveAgentInit(), creates sub-session
  via SubSessionFactory, registers completion callback via onSubSessionComplete,
  injects task message, transitions main task pending→in_progress
- check_step_status: polls sub-session state via SubSessionFactory, detects
  completion via taskAgentSessionId stored on step task
- advance_workflow: validates step task is completed in DB, calls executor.advance(),
  handles WorkflowGateError as gateBlocked (not error), resets needs_attention→in_progress
- report_result: delegates to SpaceTaskManager.setTaskStatus with result/error
- request_human_input: transitions task to needs_attention, stores question in
  currentStep, instructs Task Agent to wait for human response

Also add createTaskAgentMcpServer() factory and barrel exports in index.ts.
39 unit tests covering success paths, error paths, and edge cases.

### DIFF
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -59,6 +59,17 @@ export {
 export type { SpaceAgentToolsConfig, SpaceAgentMcpServer } from './tools/space-agent-tools';
 
 export {
+	createTaskAgentToolHandlers,
+	createTaskAgentMcpServer,
+} from './tools/task-agent-tools';
+export type {
+	TaskAgentToolsConfig,
+	TaskAgentMcpServer,
+	SubSessionFactory,
+	SubSessionState,
+} from './tools/task-agent-tools';
+
+export {
 	TASK_AGENT_TOOL_SCHEMAS,
 	SpawnStepAgentSchema,
 	CheckStepStatusSchema,
--- a/packages/daemon/src/lib/space/tools/space-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/space-agent-tools.ts
@@ -27,6 +27,8 @@ import type { SpaceRuntime } from '../runtime/space-runtime';
 import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
+import { jsonResult } from './tool-result';
+import type { ToolResult } from './tool-result';
 
 // ---------------------------------------------------------------------------
 // Config
@@ -43,18 +45,6 @@ export interface SpaceAgentToolsConfig {
 	taskRepo: SpaceTaskRepository;
 	/** Workflow run repository for listing and updating runs. */
 	workflowRunRepo: SpaceWorkflowRunRepository;
-}
-
-// ---------------------------------------------------------------------------
-// Internal helpers
-// ---------------------------------------------------------------------------
-
-interface ToolResult {
-	content: Array<{ type: 'text'; text: string }>;
-}
-
-function jsonResult(data: unknown): ToolResult {
-	return { content: [{ type: 'text', text: JSON.stringify(data) }] };
 }
 
 /**
--- a/packages/daemon/src/lib/space/tools/task-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/task-agent-tools.ts
@@ -1,0 +1,719 @@
+/**
+ * Task Agent Tools — MCP tool handlers for the Task Agent session.
+ *
+ * These handlers implement the business logic for the 5 Task Agent tools:
+ *   spawn_step_agent      — Spawn a sub-session for a workflow step's assigned agent
+ *   check_step_status     — Poll the status of a running step agent sub-session
+ *   advance_workflow      — Advance the workflow to the next step after current step completes
+ *   report_result         — Mark the task as completed/failed and record the result
+ *   request_human_input   — Pause execution and surface a question to the human user
+ *
+ * Design:
+ * - Handlers are pure functions tested independently of any MCP server layer.
+ * - Dependencies are injected via `TaskAgentToolsConfig`.
+ * - The `SubSessionFactory` abstraction allows tests to mock session creation/state.
+ * - Session completion is signalled via two paths:
+ *     1. Completion callback registered in spawn_step_agent (fires automatically)
+ *     2. Polling via check_step_status (Task Agent polls this tool)
+ *
+ * Following the pattern established in space-agent-tools.ts.
+ */
+
+import { randomUUID } from 'node:crypto';
+import { createSdkMcpServer, tool } from '@anthropic-ai/claude-agent-sdk';
+import type { Space } from '@neokai/shared';
+import type { AgentSessionInit } from '../../agent/agent-session';
+import type { SpaceRuntime } from '../runtime/space-runtime';
+import { WorkflowGateError } from '../runtime/workflow-executor';
+import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
+import type { SpaceTaskManager } from '../managers/space-task-manager';
+import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
+import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
+import type { SpaceAgentManager } from '../managers/space-agent-manager';
+import { resolveAgentInit, buildCustomAgentTaskMessage } from '../agents/custom-agent';
+import {
+	SpawnStepAgentSchema,
+	CheckStepStatusSchema,
+	AdvanceWorkflowSchema,
+	ReportResultSchema,
+	RequestHumanInputSchema,
+} from './task-agent-tool-schemas';
+import type {
+	SpawnStepAgentInput,
+	CheckStepStatusInput,
+	AdvanceWorkflowInput,
+	ReportResultInput,
+	RequestHumanInputInput,
+} from './task-agent-tool-schemas';
+
+// ---------------------------------------------------------------------------
+// Sub-session state
+// ---------------------------------------------------------------------------
+
+/**
+ * Processing state of a sub-session, as reported by SubSessionFactory.
+ */
+export interface SubSessionState {
+	/** Whether the session is actively processing a query */
+	isProcessing: boolean;
+	/** Whether the session has reached a terminal state (no more processing) */
+	isComplete: boolean;
+	/** Error details if the session ended in an error state */
+	error?: string;
+}
+
+// ---------------------------------------------------------------------------
+// SubSessionFactory
+// ---------------------------------------------------------------------------
+
+/**
+ * Abstraction for creating and querying sub-sessions.
+ * Injected into the config so tests can mock session behaviour without a real daemon.
+ */
+export interface SubSessionFactory {
+	/**
+	 * Creates and starts a new agent sub-session.
+	 * Returns the session ID of the created session.
+	 */
+	create(init: AgentSessionInit): Promise<string>;
+
+	/**
+	 * Returns the current processing state of a session, or null if the session
+	 * is not known / has been cleaned up.
+	 */
+	getProcessingState(sessionId: string): SubSessionState | null;
+
+	/**
+	 * Registers a callback that is invoked once when the given session completes
+	 * (transitions to a terminal state). The callback is called at most once.
+	 */
+	onComplete(sessionId: string, callback: () => Promise<void>): void;
+}
+
+// ---------------------------------------------------------------------------
+// Config
+// ---------------------------------------------------------------------------
+
+/**
+ * Dependencies injected into createTaskAgentToolHandlers().
+ * All fields are required — the caller (TaskAgentManager) wires them up.
+ */
+export interface TaskAgentToolsConfig {
+	/** ID of the main SpaceTask this Task Agent is orchestrating. */
+	taskId: string;
+	/** ID of the Space this task belongs to. */
+	spaceId: string;
+	/** Full Space object — needed for agent init and task message building. */
+	space: Space;
+	/** ID of the active workflow run for this task. */
+	workflowRunId: string;
+	/** Absolute path to the workspace — forwarded to agent sessions. */
+	workspacePath: string;
+	/** SpaceRuntime for accessing WorkflowExecutors by run ID. */
+	runtime: SpaceRuntime;
+	/** Workflow manager for loading workflow definitions. */
+	workflowManager: SpaceWorkflowManager;
+	/** Task repository for direct DB reads. */
+	taskRepo: SpaceTaskRepository;
+	/** Workflow run repository for reading and updating runs. */
+	workflowRunRepo: SpaceWorkflowRunRepository;
+	/** Agent manager for resolving step agents. */
+	agentManager: SpaceAgentManager;
+	/** Task manager for validated status transitions. */
+	taskManager: SpaceTaskManager;
+	/** Sub-session factory for creating and querying agent sub-sessions. */
+	sessionFactory: SubSessionFactory;
+	/**
+	 * Injects a message into an existing sub-session as a user turn.
+	 * Called after spawn to deliver the step's task context message.
+	 */
+	messageInjector: (sessionId: string, message: string) => Promise<void>;
+	/**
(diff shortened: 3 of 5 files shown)


lsm (commented):

---

## Summary

This is a solid implementation of the Task Agent tool handlers. The design is clean, the injection-based architecture is sound, and there is strong unit test coverage with a real SQLite database and a well-crafted mock `SubSessionFactory`. That said, there are a few meaningful correctness issues and code quality gaps that should be addressed.

---

## P0 — Blocking Bugs

### 1. Message injection failure returns `success: false` despite a live sub-session

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 291–298

```ts
// Message injection failure is non-fatal — the sub-session will still run,
// just without the task context message. Log the issue but don't abort.
const message = err instanceof Error ? err.message : String(err);
return jsonResult({
    success: false,
    error: `Sub-session created (${actualSessionId}) but message injection failed: ${message}`,
});
```

The comment says "non-fatal", but the code **returns `success: false`** and terminates the handler. The sub-session has already been created and the completion callback already registered — but the crucial `taskRepo.updateTask(stepTask.id, { taskAgentSessionId: actualSessionId })` below is never reached, so `check_step_status` will perpetually report `sessionStatus: 'not_started'`. The sub-session runs silently in the background with no way for the Task Agent to track or detect its completion.

The fix is to either: (a) record `taskAgentSessionId` *before* attempting message injection, so the session remains trackable even on injection failure; or (b) return `success: true` with a warning alongside the `sessionId`, so the Task Agent can still poll `check_step_status`. As written, both the comment and the code are wrong relative to each other.

### 2. `step_result` field in `AdvanceWorkflowInput` is accepted but silently discarded

**File:** `packages/daemon/src/lib/space/tools/task-agent-tool-schemas.ts`, line 73–77  
**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, line 458

`AdvanceWorkflowSchema` declares an optional `step_result` field described as "used when evaluating transition conditions". The `advance_workflow` handler receives `_args: AdvanceWorkflowInput` (the entire `args` is ignored with an underscore prefix). `step_result` is never passed to `executor.advance()`.

This is a contract lie: the tool schema promises the LLM that providing `step_result` influences transition evaluation, but it is a no-op. Either remove the field from the schema, or thread it through to `executor.advance()` if the executor supports it.

---

## P1 — Should Fix

### 3. Silently skipped message injection when `effectiveTask.customAgentId` resolves to no agent

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 278–290

```ts
const agent = agentManager.getById(effectiveTask.customAgentId!);
if (agent) {
    // ...inject message...
}
// No else — silent skip
```

`resolveAgentInit()` already throws if `customAgentId` is unset or the agent doesn't exist (lines 416–424 of `custom-agent.ts`). If `resolveAgentInit()` succeeded (meaning the agent *exists*), then `agentManager.getById(effectiveTask.customAgentId!)` should never return `null`. The `if (agent)` guard silently eats a state inconsistency: if the agent can't be found here after `resolveAgentInit()` succeeded, something is very wrong. At minimum, this should surface as a warning in the error path rather than silently skipping message injection, leaving the sub-session with no task context.

### 4. `spaceId` is declared in `TaskAgentToolsConfig` but never used in the implementation

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, line 105

`spaceId` is a required field in `TaskAgentToolsConfig` but is not destructured and not referenced anywhere in `createTaskAgentToolHandlers()`. This is dead config weight. Either remove the field or document why it is reserved for future use (e.g., multi-tenancy checks). Given it is part of the public interface, callers are forced to provide a value that is immediately discarded.

### 5. `extFactory` and `_setState` are declared but unused in the test

**File:** `packages/daemon/tests/unit/space/task-agent-tools.test.ts`, lines 696–702

```ts
const extFactory = factory as ReturnType<typeof makeMockSessionFactory> & {
    _setState: (id: string, state: SubSessionState) => void;
};
// Access private map via cast to trigger isComplete=true
(
    factory as unknown as { getProcessingState: (id: string) => SubSessionState }
).getProcessingState = (_id: string) => ({ isProcessing: false, isComplete: true });
```

`extFactory` is assigned but never used — only the raw `factory` cast is used below. The `_setState` type it references doesn't exist on the mock. This is dead code masquerading as a cast, and it indicates a refactoring artifact. The `oxlint no-unused-vars` rule should catch `extFactory`, but if the test file is exempt from that rule it will silently remain.

### 6. Missing test coverage for concurrent/double-spawn scenario

There is no test verifying that calling `spawn_step_agent` twice for the same `step_id` behaves predictably. The implementation picks `stepTasks[stepTasks.length - 1]` (the last created task). If a second spawn creates a new sub-session but a completion callback for the first is still in flight, the step task's `taskAgentSessionId` is overwritten without cancelling the prior session, potentially causing orphaned sessions and duplicate `onSubSessionComplete` invocations.

---

## P2 — Meaningful Suggestions

### 7. `taskRepo.updateTask` calls after `return jsonResult(...)` are unreachable on early returns

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 291–307

The comment on line 292 says message injection failure is "non-fatal", yet the code returns on line 295 before persisting `taskAgentSessionId` on the step task (line 304) and before updating the main task's `currentStep` (line 320). This makes the early return semantics misleading — a reader who sees the "non-fatal" comment reasonably expects execution to continue, not to return.

### 8. The `ToolResult` type and `jsonResult` helper are duplicated between `space-agent-tools.ts` and `task-agent-tools.ts`

Both files define identical `ToolResult` interface and `jsonResult` function at module scope. These should be extracted to a shared utility file. This is minor today but will drift over time.

### 9. `advance_workflow` terminal case description is misleading

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 518–531

When `newTasks.length === 0` the response calls the `nextStep` field `completedStep`, but the underlying `step` value from `executor.advance()` is actually the step that was *just finished* (the terminal step), not a "completed step" in a workflow sense. The field naming is confusing. Using `terminalStep` would be more accurate and align with the `terminal: true` flag beside it.

---

## P3 — Nits

- `// Step` comment marker missing for `extFactory` cleanup (test, line 696).
- The check at line 403 uses `!` non-null assertion (`stepTask.taskAgentSessionId!`) immediately after a `if (!stepTask.taskAgentSessionId)` guard on line 391 — the guard makes the assertion redundant.
- In `request_human_input`, the question is stored in `currentStep` (line 643) but the full `questionWithContext` (which includes context) is passed to `setTaskStatus` as `error`. Only the bare question appears in the UI — this is probably intentional, but is worth a comment explaining the deliberate split.

---

## Conclusion

The architectural design is good — handler separation, dependency injection, and the `SubSessionFactory` abstraction are all solid choices. The P0 issues (especially the message injection bug that leaves a live sub-session untrackable) and the `step_result` schema/implementation mismatch need to be addressed before this is safe to merge. The `spaceId` dead field and unused `extFactory` should also be cleaned up.

lsm (commented):

**Recommendation: APPROVE**

This is a well-structured implementation of the 5 Task Agent MCP tool handlers. The code follows the established patterns from `space-agent-tools.ts`, has clean separation between handler logic and MCP server wiring, and comprehensive test coverage with 39 tests (all passing) covering success paths, error paths, edge cases, and full lifecycle integration tests.

### Strengths

- **Clean architecture**: Handlers are pure functions injected with dependencies via `TaskAgentToolsConfig`, making them testable without MCP infrastructure.
- **Thorough error handling**: Each handler validates preconditions and returns structured JSON errors rather than throwing. Failures in non-critical paths (message injection, status transitions) are handled gracefully with appropriate comments.
- **Good test design**: Tests use real SQLite databases (via `runMigrations`) with mock `SubSessionFactory`, giving high confidence in DB interactions. The lifecycle integration tests (`spawn -> check -> advance -> report`) are particularly valuable.
- **Correct SDK API usage**: `tool()` calls use `Schema.shape` which correctly provides `AnyZodRawShape` as expected by the Claude Agent SDK.
- **Well-documented**: JSDoc comments, module-level documentation, and inline comments explain the design rationale and flow.

### Issues Found

**P2 -- `step_result` schema field is accepted but silently ignored** (`task-agent-tools.ts:458`, `task-agent-tool-schemas.ts:72-78`)

The `AdvanceWorkflowSchema` defines `step_result` as an optional string field, creating an API contract that suggests the field will be used when evaluating transition conditions. However, the handler uses `_args` (underscore prefix = unused) and calls `executor.advance()` with no arguments. Callers providing `step_result` get no feedback that their input is discarded.

Options: (a) remove `step_result` from the schema until `executor.advance()` supports it, or (b) forward it to the executor if there is a planned enhancement, or (c) at minimum add a comment in the handler acknowledging this is a placeholder for future use.

**P3 -- `spaceId` declared in `TaskAgentToolsConfig` but unused in handlers** (`task-agent-tools.ts:105`, `160-174`)

The `spaceId` field is defined in the config interface and required by callers, but it is never destructured or referenced in `createTaskAgentToolHandlers()`. It may be needed by a future consumer (e.g., `TaskAgentManager`), in which case a brief comment would clarify its purpose.

**P3 -- `randomUUID` pre-generation could cause ID mismatch** (`task-agent-tools.ts:243-265`)

The handler generates `subSessionId` via `randomUUID()` and embeds it in the `init` object, but then `sessionFactory.create(init)` returns `actualSessionId` which may differ. All subsequent code correctly uses `actualSessionId`, so there is no bug. However, the `subSessionId` baked into `init` could cause confusion if the factory's internal session ID tracking references a different ID. This is a minor design note, not a bug.

### Summary

The implementation is solid and complete. All 39 tests pass. The only substantive concern is the `step_result` schema field that creates a misleading API contract. The remaining items are minor cosmetic/design notes. Given the overall quality, this is an **APPROVE** recommendation.

lsm (commented):

**Recommendation: APPROVE**

---

## Re-review: All P0 and P1 issues from prior reviews have been addressed

The fix commit (`6e863e5b`) successfully resolves all issues flagged by the previous claude-sonnet-4-6 and claude-opus-4-6 reviews. Here is a summary of each fix:

### Previously P0 -- Now Fixed

**1. Message injection failure leaving an untrackable sub-session** -- `taskAgentSessionId` is now persisted BEFORE message injection (line 293 of `task-agent-tools.ts`). On injection failure, the handler returns `success: true` with a `warning` field, so the Task Agent can still poll via `check_step_status`. This correctly matches the "non-fatal" intent.

**2. `step_result` accepted but silently discarded** -- A clear comment block was added to the handler (lines 497-500) documenting that `step_result` is a forward-compatibility placeholder and that `WorkflowExecutor.advance()` does not yet accept a result parameter.

### Previously P1 -- Now Fixed

**3. Silent skip when `agentManager.getById()` returns null** -- Now returns `success: true` with a descriptive warning (lines 321-331) surfacing the data inconsistency rather than silently skipping message injection.

**4. `spaceId` dead config field** -- Removed from `TaskAgentToolsConfig`. Space ID is available via `space.id`.

**5. Unused `extFactory` variable in tests** -- Removed the dead assignment. Only the direct cast on `factory` remains (lines 720-722).

**6. Missing double-spawn test** -- Added idempotency guard in the handler (lines 229-238) and a new test (lines 544-565) verifying that a second `spawn_step_agent` call for the same step returns the existing session without creating a duplicate.

### Previously P2 -- Now Fixed

**7. Unreachable `taskRepo.updateTask` after early return** -- Reordered: `taskAgentSessionId` and `currentStep` are now persisted before message injection, making the flow correct whether injection succeeds or fails.

**8. `ToolResult`/`jsonResult` duplication** -- Extracted to `packages/daemon/src/lib/space/tools/tool-result.ts`, imported by both `space-agent-tools.ts` and `task-agent-tools.ts`.

**9. Misleading `completedStep` field name** -- Renamed to `terminalStep` (line 569).

### Remaining Minor Item (P3)

The `step_result` schema description (line 74-76 of `task-agent-tool-schemas.ts`) still reads "used when evaluating transition conditions" in its `.describe()` string. While the handler comment documents it as a placeholder, the `.describe()` text is what the LLM sees as the tool schema. A more accurate description like "Optional result summary from the completed step (reserved for future use)" would prevent the LLM from spending effort crafting a value that is currently discarded. This is cosmetic and non-blocking.

### Verification

- All 40 tests pass (39 original + 1 new double-spawn idempotency test)
- `bun run typecheck` passes cleanly
- The `tool-result.ts` extraction is correctly imported by both consumer files

---

All P0, P1, and P2 issues from prior reviews are adequately resolved. The code is clean, well-commented, and the test coverage is comprehensive.

lsm (commented):

**Recommendation: APPROVE** (same-author PR, posting as COMMENT per policy)

---

## Re-review Verdict

All six P0/P1 issues and both P2 issues from the initial review have been correctly fixed. The 40 tests pass. The implementation is sound and ready to merge.

---

## Fix Verification

### P0#1 — taskAgentSessionId recorded before injection (FIXED)

`taskRepo.updateTask(stepTask.id, { taskAgentSessionId: actualSessionId, ... })` now executes before the message injection block (lines 293-296 of `task-agent-tools.ts`). On injection failure, the handler returns `success: true` with a `warning` field. The sub-session is always trackable via `check_step_status` regardless of injection outcome.

### P0#2 — step_result documented as forward-compat placeholder (FIXED)

The `advance_workflow` handler comment block (lines 497-500) now explicitly states that `step_result` is retained for forward compatibility and is not forwarded to `executor.advance()`. The handler uses `_args` and the doc comment is accurate.

### P1#3 — Agent-not-found returns warning (FIXED)

When `agentManager.getById()` returns null after `resolveAgentInit()` succeeded, the handler now returns `success: true` with a descriptive warning (lines 321-331) rather than silently skipping message injection.

### P1#4 — Redundant spaceId removed (FIXED)

`spaceId` removed from `TaskAgentToolsConfig`. The space ID is accessed via `space.id` where needed. Callers no longer provide a redundant field.

### P1#5 — Unused extFactory removed (FIXED)

The `extFactory` / `_setState` dead cast is gone from the test. The direct `factory` cast pattern is used throughout.

### P1#6 — Idempotency guard + test (FIXED)

`spawn_step_agent` now checks `stepTask.taskAgentSessionId` before creating a new session (lines 229-238). If it is set, the existing session ID is returned with `alreadySpawned: true`. The new double-spawn test (lines 544-565) verifies this behaviour: the second call returns the same session ID as the first.

### P2#8 — Shared tool-result.ts (FIXED)

`packages/daemon/src/lib/space/tools/tool-result.ts` correctly exports `ToolResult` and `jsonResult`. Both `space-agent-tools.ts` and `task-agent-tools.ts` import from it. No duplication remains.

### P2#9 — completedStep → terminalStep rename (FIXED)

`advance_workflow` now returns `terminalStep` (not `completedStep`) when `newTasks.length === 0` (line 569). The naming aligns with `terminal: true`.

---

## Remaining Minor Item

**P3 — step_result `.describe()` text is still misleading to the LLM**

`packages/daemon/src/lib/space/tools/task-agent-tool-schemas.ts`, line 75

```ts
'Optional result or output summary from the completed step, used when evaluating transition conditions'
```

The `.describe()` string is what gets serialised into the MCP tool schema and shown to the LLM. It still implies `step_result` influences transition evaluation, even though the handler comment now correctly documents it as a no-op placeholder. The LLM may waste effort constructing a value it believes matters. A more honest description like `"Optional result summary from the completed step (reserved for future use; currently not forwarded to the executor)"` would be more accurate, but this is a cosmetic non-blocker.

---

## New Observation (Non-blocking)

**P3 — Stale `spaceId` key in test's `makeConfig` object**

`packages/daemon/tests/unit/space/task-agent-tools.test.ts`, line 280

```ts
spaceId: ctx.spaceId,   // excess property — not in TaskAgentToolsConfig
```

`TaskAgentToolsConfig` no longer has a `spaceId` field (it was removed by this PR). The test's `makeConfig` helper still passes `spaceId` as a property. Since the `tests/` directory is excluded from `tsconfig.json`, TypeScript does not flag this as an excess property. At runtime Bun simply ignores the extra key. Functionally harmless, but should be cleaned up to keep the test fixture in sync with the interface.

---

## Summary

The fix commit (`6e863e5b`) correctly addresses all blocking and should-fix issues from the prior reviews. All 40 tests pass, TypeScript is clean, and the shared `tool-result.ts` extraction eliminates the duplication. The two remaining items above are cosmetic and non-blocking.

PR #515, opened 2026-03-21
feat: define NotificationSink interface and SpaceNotificationEvent types (Task 2.1)

lsm:
- Add SpaceNotificationEvent discriminated union with four event kinds:
  task_needs_attention, workflow_run_needs_attention, task_timeout,
  workflow_run_completed — each with typed payloads
- Add NotificationSink interface (single notify() method) for SpaceRuntime
  to push events to consumers without coupling to a real session
- Add NullNotificationSink no-op implementation for tests and default wiring
- Export all types from the space module index
- Add unit tests covering event construction, union narrowing, and
  NullNotificationSink behaviour

### DIFF
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -31,6 +31,15 @@ export type {
 } from './runtime/workflow-executor';
 export { SpaceRuntime } from './runtime/space-runtime';
 export type { SpaceRuntimeConfig, ResolvedTaskType } from './runtime/space-runtime';
+export { NullNotificationSink } from './runtime/notification-sink';
+export type {
+	NotificationSink,
+	SpaceNotificationEvent,
+	TaskNeedsAttentionEvent,
+	WorkflowRunNeedsAttentionEvent,
+	TaskTimeoutEvent,
+	WorkflowRunCompletedEvent,
+} from './runtime/notification-sink';
 export { SpaceRuntimeService } from './runtime/space-runtime-service';
 export type { SpaceRuntimeServiceConfig } from './runtime/space-runtime-service';
 
--- a/packages/daemon/src/lib/space/runtime/notification-sink.ts
+++ b/packages/daemon/src/lib/space/runtime/notification-sink.ts
@@ -1,0 +1,136 @@
+/**
+ * NotificationSink — interface for Space Runtime to push structured events
+ * to consumers such as the Space Agent session or test harnesses.
+ *
+ * SpaceRuntime calls `notify()` after mechanical processing whenever a state
+ * change requires judgment or awareness (e.g. a task hits `needs_attention`,
+ * a workflow run completes, or a task times out).
+ *
+ * Consumers that don't need notifications can use `NullNotificationSink`.
+ */
+
+// ---------------------------------------------------------------------------
+// Event payload types
+// ---------------------------------------------------------------------------
+
+/** A task has transitioned to `needs_attention` and requires judgment. */
+export interface TaskNeedsAttentionEvent {
+	kind: 'task_needs_attention';
+	/** Space the task belongs to. */
+	spaceId: string;
+	/** Task that needs attention. */
+	taskId: string;
+	/** Human-readable reason the task needs attention. */
+	reason: string;
+	/** ISO-8601 timestamp when the event was emitted. */
+	timestamp: string;
+}
+
+/** A workflow run has transitioned to `needs_attention` (e.g. a transition condition failed). */
+export interface WorkflowRunNeedsAttentionEvent {
+	kind: 'workflow_run_needs_attention';
+	/** Space the workflow run belongs to. */
+	spaceId: string;
+	/** Workflow run that needs attention. */
+	runId: string;
+	/** Human-readable reason the run needs attention. */
+	reason: string;
+	/** ISO-8601 timestamp when the event was emitted. */
+	timestamp: string;
+}
+
+/** A task has been running longer than the configured timeout threshold. */
+export interface TaskTimeoutEvent {
+	kind: 'task_timeout';
+	/** Space the task belongs to. */
+	spaceId: string;
+	/** Task that has exceeded its time threshold. */
+	taskId: string;
+	/** Elapsed milliseconds since the task started. */
+	elapsedMs: number;
+	/** ISO-8601 timestamp when the event was emitted. */
+	timestamp: string;
+}
+
+/** A workflow run has reached a terminal state (completed or failed). */
+export interface WorkflowRunCompletedEvent {
+	kind: 'workflow_run_completed';
+	/** Space the workflow run belongs to. */
+	spaceId: string;
+	/** Workflow run that completed. */
+	runId: string;
+	/** Final status of the run. */
+	status: 'completed' | 'failed' | 'cancelled';
+	/** Optional summary of what the run accomplished. */
+	summary?: string;
+	/** ISO-8601 timestamp when the event was emitted. */
+	timestamp: string;
+}
+
+/**
+ * Discriminated union of all structured notification events emitted by
+ * `SpaceRuntime`. Use the `kind` field to narrow to a specific event type.
+ *
+ * @example
+ * ```ts
+ * function handleEvent(event: SpaceNotificationEvent) {
+ *   switch (event.kind) {
+ *     case 'task_needs_attention':
+ *       // event is TaskNeedsAttentionEvent
+ *       break;
+ *     case 'workflow_run_needs_attention':
+ *       // event is WorkflowRunNeedsAttentionEvent
+ *       break;
+ *     case 'task_timeout':
+ *       // event is TaskTimeoutEvent
+ *       break;
+ *     case 'workflow_run_completed':
+ *       // event is WorkflowRunCompletedEvent
+ *       break;
+ *   }
+ * }
+ * ```
+ */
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly defines the `NotificationSink` interface and `SpaceNotificationEvent` discriminated union as specified in Task 2.1. The implementation is well-structured, type-safe, and follows project conventions.

### What looks good

- **Discriminated union design**: Four event kinds with `kind` as the discriminant field. Each event interface has appropriate domain-specific fields (`taskId` vs `runId`, `elapsedMs` for timeouts, `status` for completion).
- **JSDoc quality**: Every interface, field, and the union type itself has clear JSDoc comments. The `@example` block on `SpaceNotificationEvent` showing the switch/case narrowing pattern is helpful for consumers.
- **`NullNotificationSink`**: Clean no-op implementation with proper `implements` clause. Good default for contexts where notifications aren't needed.
- **Exports**: All types and the concrete class are properly re-exported from `packages/daemon/src/lib/space/index.ts`, with correct `type` vs value export separation.
- **Code style**: Tab indentation, single quotes, semicolons, trailing commas -- all consistent with Biome config. No console calls. Underscore prefix on `_event` parameter avoids unused-var lint errors.
- **Tests**: 9 tests covering all 4 event kinds, optional field (`summary`), union narrowing, the `NullNotificationSink` no-op behavior, all event kinds through the sink, and concurrent invocation. Good coverage for a pure type + no-op implementation.

### Minor observations (P3)

1. **`WorkflowRunCompletedEvent.status` JSDoc mismatch** (`notification-sink.ts:55`): The JSDoc says "terminal state (completed or failed)" but the actual type also includes `'cancelled'`. Consider updating to "(completed, failed, or cancelled)".

2. **Tests are compile-time-verifiable**: Since these are pure interface types and a no-op class, the tests largely validate TypeScript compilation rather than runtime behavior. This is fine for an interface-definition PR -- real behavioral tests will come when actual `NotificationSink` implementations are built.

3. **No `workflowId` on workflow events**: `WorkflowRunNeedsAttentionEvent` and `WorkflowRunCompletedEvent` carry `runId` but not the parent `workflowId`. Consumers may need to resolve the workflow from the run ID. This seems intentional (the run ID is the actionable identifier), but worth confirming as consumers are built.

Overall this is a clean, well-documented interface definition that provides a solid foundation for the Space Agent coordination layer. No blocking issues.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (own PR — posted as COMMENT)

---

Overall the implementation is clean, well-typed, and passes all checks (lint, typecheck, format, tests). The discriminated union pattern is correctly applied, the JSDoc is thorough, and the `NullNotificationSink` no-op is the right default. One real correctness issue and two minor observations below.

### P1 — `WorkflowRunCompletedEvent.status` introduces a `'failed'` value that does not exist in the domain model

**File:** `packages/daemon/src/lib/space/runtime/notification-sink.ts`, line 63

```ts
status: 'completed' | 'failed' | 'cancelled';
```

`WorkflowRunStatus` in `packages/shared/src/types/space.ts` (lines 249-254) is:

```ts
export type WorkflowRunStatus =
  | 'pending'
  | 'in_progress'
  | 'completed'
  | 'cancelled'
  | 'needs_attention';
```

`'failed'` does not exist in this type. When `SpaceRuntime` eventually emits a `workflow_run_completed` event by reading from a real `SpaceWorkflowRun`, it won't be able to pass `run.status` directly into the event's `status` field without a translation step. This creates a silent impedance mismatch: the type compiles because `'failed'` is a valid string literal, but there is no code path in the current domain model that can produce that value.

If the intent is to distinguish a run that ended due to a condition gate failure from one that was externally cancelled, the canonical representation in this codebase is `needs_attention`, not `failed`. A run that truly "failed" (e.g. unrecoverable error) may still be modelled as `needs_attention` — check with the domain model before adding new status vocabulary here.

**Fix options:**
- Replace `'failed'` with `'needs_attention'` to align with `WorkflowRunStatus`, making the event status a subset of the real status type.
- Or add `'failed'` to `WorkflowRunStatus` in `@neokai/shared` if the domain genuinely needs it (would require a broader change).
- Or type the field as `WorkflowRunStatus` directly, allowing all run statuses (filtering to terminal ones at call sites).

### P2 — `async` on `NullNotificationSink.notify` is unnecessary

**File:** `packages/daemon/src/lib/space/runtime/notification-sink.ts`, line 133

```ts
async notify(_event: SpaceNotificationEvent): Promise<void> {
    // intentional no-op
}
```

The `async` keyword is not needed here — the method never awaits anything and just returns `undefined`. A plain `notify(_event: SpaceNotificationEvent): Promise<void> { return Promise.resolve(); }` is marginally cleaner and avoids creating a microtask for every call. The current form is not wrong (TypeScript infers the return as `Promise<void>`), but the `async` adds a small runtime overhead for what is explicitly a no-op. Minor.

### P3 — Test for `workflow_run_needs_attention` does not assert `timestamp`

**File:** `packages/daemon/tests/unit/space/notification-sink.test.ts`, line 28-41

The test constructs a `WorkflowRunNeedsAttentionEvent` with a `timestamp` field but the assertions on lines 38-41 only check `kind`, `spaceId`, `runId`, and `reason` — the `timestamp` is not asserted. Every other event type test asserts its `timestamp`. This is an inconsistency in test coverage. It is unlikely to mask a bug (the type requires the field), but it is inconsistent with the other test cases.

### P3 — Import of individual event types in test could be simplified

**File:** `packages/daemon/tests/unit/space/notification-sink.test.ts`, lines 1-9

The test imports all four individual event interfaces (`TaskNeedsAttentionEvent`, `WorkflowRunNeedsAttentionEvent`, `TaskTimeoutEvent`, `WorkflowRunCompletedEvent`) alongside `SpaceNotificationEvent`. Since each test constructs its event with an explicit type annotation against the individual interface, this is fine for type-checking, but the `SpaceNotificationEvent` union alone would be sufficient for all of these annotations. This is a minor style observation — no functional impact.

---

### Summary

The task is correctly scoped and the implementation is solid except for the `'failed'` status value in `WorkflowRunCompletedEvent`, which introduces a term that does not exist in the canonical `WorkflowRunStatus` domain type. This will cause friction when the runtime wires up real event emission. The P2 and P3 items are polish.

lsm (commented):

**Recommendation: APPROVE**

All three issues raised in the previous review have been correctly addressed. Here is a summary of what was verified:

### Verified Fixes

**P1 — `WorkflowRunCompletedEvent.status` alignment**

`notification-sink.ts` line 67 now reads:
```ts
status: 'completed' | 'cancelled' | 'needs_attention';
```
This is the exact terminal subset of `WorkflowRunStatus` in `packages/shared/src/types/space.ts` (lines 249–254), which defines `'pending' | 'in_progress' | 'completed' | 'cancelled' | 'needs_attention'`. The non-terminal values `'pending'` and `'in_progress'` are correctly excluded. The JSDoc on lines 62–66 has also been updated to explain `'needs_attention'`.

**P2 — `async` removed from `NullNotificationSink.notify`**

`NullNotificationSink.notify` (line 137) is now a plain method returning `Promise.resolve()` directly, with no `async` keyword. The method still satisfies the `NotificationSink` interface contract.

**P3 — Missing `timestamp` assertion in `workflow_run_needs_attention` test**

Line 41 of the test file now asserts:
```ts
expect(event.timestamp).toBe('2026-03-20T11:00:00.000Z');
```
This matches the pattern used in the `task_needs_attention` test case.

### Remaining Minor Note (not blocking)

The `task_timeout` test case (lines 44–55) still only asserts `kind` and `elapsedMs`, leaving `spaceId`, `taskId`, and `timestamp` unasserted. This was a pre-existing gap not introduced by this PR and not part of the reported issues, so it does not block approval. It is worth cleaning up in a follow-up.

### Conclusion

The implementation is correct and complete. The interface, discriminated union, and no-op implementation are all sound. The export surface in `packages/daemon/src/lib/space/index.ts` is clean. No new issues were introduced.

lsm (commented):

Re-review after feedback was addressed. All previously raised issues have been resolved:

- **P1 (fixed):** `WorkflowRunCompletedEvent.status` now uses `'completed' | 'cancelled' | 'needs_attention'`, correctly matching the terminal subset of `WorkflowRunStatus` defined in `packages/shared/src/types/space.ts` (lines 249-254).
- **P2 (fixed):** `NullNotificationSink.notify` is no longer marked `async`; it returns `Promise.resolve()` directly, avoiding unnecessary microtask overhead.
- **P3 (fixed):** The `workflow_run_needs_attention` test case now asserts the `timestamp` field.
- **P3 (fixed):** JSDoc updated to reference `needs_attention` instead of `failed`.

### Current state assessment

**Correctness:** The interface and event types are well-structured. The discriminated union on `kind` enables clean narrowing. The status values on `WorkflowRunCompletedEvent` are the correct terminal subset of `WorkflowRunStatus`.

**Completeness:** All four event types cover the scenarios described in the task (task needs attention, workflow run needs attention, task timeout, workflow run completed). The `NullNotificationSink` provides a clean no-op default. Exports are properly wired through `packages/daemon/src/lib/space/index.ts`.

**Tests:** Tests verify construction of all event types, union narrowing, the optional `summary` field, and that `NullNotificationSink` resolves cleanly for all event kinds including concurrent calls.

**No issues found.** The implementation is clean, well-documented, and correctly aligned with the existing domain types. Recommendation: **APPROVE**.

PR #516, opened 2026-03-21
feat: add SpaceAutonomyLevel, SpaceConfig types, and autonomy_level DB column

lsm:
## Summary

- Adds `SpaceAutonomyLevel = 'supervised' | 'semi_autonomous'` type to shared package — controls how much the Space Agent can act without human approval
- Adds typed `SpaceConfig` interface with `maxConcurrentTasks?: number` and `taskTimeoutMs?: number` fields; replaces the previous `config?: Record<string, unknown>` on `Space`, `CreateSpaceParams`, and `UpdateSpaceParams`
- Adds optional `autonomyLevel?: SpaceAutonomyLevel` field to `Space`, `CreateSpaceParams`, and `UpdateSpaceParams` (defaults to `'supervised'`)
- Migration 33: `ALTER TABLE spaces ADD COLUMN autonomy_level TEXT NOT NULL DEFAULT 'supervised'` — backward-compatible (all existing spaces become supervised)
- `SpaceRepository` updated to read/write `autonomy_level` in `createSpace`, `updateSpace`, and `rowToSpace`

## Test plan

- [ ] Migration tests: column exists after full migration chain, default is `'supervised'`, legacy DB path (table without column gets column added), idempotency (second run is a no-op)
- [ ] Repository tests: default autonomyLevel on create, explicit `'semi_autonomous'` on create, update autonomyLevel, typed config fields round-trip
- [ ] All existing space repository and migration tests continue to pass
- [ ] `bun run typecheck` passes with no errors

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/space-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-handlers.ts
@@ -20,8 +20,6 @@ import type {
 	SpaceTask,
 	SpaceWorkflowRun,
 } from '@neokai/shared';
-
-const VALID_AUTONOMY_LEVELS: SpaceAutonomyLevel[] = ['supervised', 'semi_autonomous'];
 import type { DaemonHub } from '../daemon-hub';
 import type { SpaceManager } from '../space/managers/space-manager';
 import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
@@ -33,6 +31,7 @@ import { seedBuiltInWorkflows } from '../space/workflows/built-in-workflows';
 import { Logger } from '../logger';
 
 const log = new Logger('space-handlers');
+const VALID_AUTONOMY_LEVELS: SpaceAutonomyLevel[] = ['supervised', 'semi_autonomous'];
 
 export interface SpaceOverviewResult {
 	space: Space;
--- a/packages/daemon/src/lib/rpc-handlers/space-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-handlers.ts
@@ -14,11 +14,14 @@
 import type { MessageHub } from '@neokai/shared';
 import type {
 	Space,
+	SpaceAutonomyLevel,
 	CreateSpaceParams,
 	UpdateSpaceParams,
 	SpaceTask,
 	SpaceWorkflowRun,
 } from '@neokai/shared';
+
+const VALID_AUTONOMY_LEVELS: SpaceAutonomyLevel[] = ['supervised', 'semi_autonomous'];
 import type { DaemonHub } from '../daemon-hub';
 import type { SpaceManager } from '../space/managers/space-manager';
 import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
@@ -56,6 +59,14 @@ export function setupSpaceHandlers(
 		}
 		if (!params.name || params.name.trim() === '') {
 			throw new Error('name is required');
+		}
+		if (
+			params.autonomyLevel !== undefined &&
+			!VALID_AUTONOMY_LEVELS.includes(params.autonomyLevel)
+		) {
+			throw new Error(
+				`Invalid autonomyLevel: ${params.autonomyLevel}. Must be one of: ${VALID_AUTONOMY_LEVELS.join(', ')}`
+			);
 		}
 
 		const space = await spaceManager.createSpace(params);
@@ -118,6 +129,14 @@ export function setupSpaceHandlers(
 
 		if (!params.id) {
 			throw new Error('id is required');
+		}
+		if (
+			params.autonomyLevel !== undefined &&
+			!VALID_AUTONOMY_LEVELS.includes(params.autonomyLevel)
+		) {
+			throw new Error(
+				`Invalid autonomyLevel: ${params.autonomyLevel}. Must be one of: ${VALID_AUTONOMY_LEVELS.join(', ')}`
+			);
 		}
 
 		const { id, ...updateParams } = params;
--- a/packages/daemon/src/storage/repositories/space-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-repository.ts
@@ -6,7 +6,13 @@
 
 import type { Database as BunDatabase } from 'bun:sqlite';
 import { generateUUID } from '@neokai/shared';
-import type { Space, CreateSpaceParams, UpdateSpaceParams } from '@neokai/shared';
+import type {
+	Space,
+	SpaceAutonomyLevel,
+	SpaceConfig,
+	CreateSpaceParams,
+	UpdateSpaceParams,
+} from '@neokai/shared';
 import type { SQLiteValue } from '../types';
 
 export class SpaceRepository {
@@ -20,8 +26,8 @@ export class SpaceRepository {
 		const now = Date.now();
 
 		const stmt = this.db.prepare(
-			`INSERT INTO spaces (id, workspace_path, name, description, background_context, instructions, default_model, allowed_models, session_ids, status, config, created_at, updated_at)
-       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
+			`INSERT INTO spaces (id, workspace_path, name, description, background_context, instructions, default_model, allowed_models, session_ids, status, autonomy_level, config, created_at, updated_at)
+       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
 		);
 
 		stmt.run(
@@ -35,6 +41,7 @@ export class SpaceRepository {
 			JSON.stringify(params.allowedModels ?? []),
 			'[]',
 			'active',
+			params.autonomyLevel ?? 'supervised',
 			params.config ? JSON.stringify(params.config) : null,
 			now,
 			now
@@ -115,6 +122,10 @@ export class SpaceRepository {
 		if (params.allowedModels !== undefined) {
 			fields.push('allowed_models = ?');
 			values.push(JSON.stringify(params.allowedModels));
+		}
+		if (params.autonomyLevel !== undefined) {
+			fields.push('autonomy_level = ?');
+			values.push(params.autonomyLevel);
 		}
 		if (params.config !== undefined) {
 			fields.push('config = ?');
@@ -204,7 +215,8 @@ export class SpaceRepository {
 	private rowToSpace(row: Record<string, unknown>): Space {
 		const rawModels = JSON.parse((row.allowed_models as string) ?? '[]') as string[];
 		const rawConfig = row.config as string | null;
-		const config = rawConfig ? (JSON.parse(rawConfig) as Record<string, unknown>) : undefined;
+		const config = rawConfig ? (JSON.parse(rawConfig) as SpaceConfig) : undefined;
+		const rawAutonomyLevel = (row.autonomy_level as string | null) ?? 'supervised';
 
 		return {
 			id: row.id as string,
@@ -217,6 +229,7 @@ export class SpaceRepository {
 			allowedModels: rawModels.length > 0 ? rawModels : undefined,
 			sessionIds: JSON.parse(row.session_ids as string) as string[],
 			status: row.status as 'active' | 'archived',
+			autonomyLevel: rawAutonomyLevel as SpaceAutonomyLevel,
 			config,
 			createdAt: row.created_at as number,
 			updatedAt: row.updated_at as number,
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -126,6 +126,9 @@ export function runMigrations(db: BunDatabase, createBackup: () => void): void {
 
 	// Migration 32: Add task_agent_session_id column to space_tasks.
 	runMigration32(db);
+
+	// Migration 33: Add autonomy_level column to spaces table.
+	runMigration33(db);
 }
 
 /**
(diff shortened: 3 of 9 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured PR that adds `SpaceAutonomyLevel` and `SpaceConfig` types to the shared package, a corresponding DB migration, and repository support. The implementation follows existing codebase patterns consistently.

### What looks good

- **Types are well-designed**: `SpaceAutonomyLevel` and `SpaceConfig` are correctly placed in shared types with clear documentation.
- **Migration 33 is idempotent**: Uses the standard try/catch-on-SELECT pattern matching all other migrations in this file.
- **Repository changes are correct**: `createSpace`, `updateSpace`, and `rowToSpace` all handle the new column properly with a sensible `'supervised'` default.
- **Test coverage is thorough**: Migration tests cover fresh DB, legacy DB, idempotency, and default values. Repository tests cover create defaults, explicit values, and update round-trips.
- **`config` type narrowing** from `Record<string, unknown>` to `SpaceConfig` is a safe, backwards-compatible improvement since `SpaceConfig` only has optional fields.
- **Test helper `space-test-db.ts`** is properly updated to include the new column.

### Minor observations (P3)

1. **No CHECK constraint on `autonomy_level`** (`packages/daemon/src/storage/schema/migrations.ts:2005`): The `status` column on spaces has `CHECK(status IN ('active', 'archived'))` in the original CREATE TABLE, but `autonomy_level` has no CHECK constraint. SQLite `ALTER TABLE ADD COLUMN` does not support adding CHECK constraints, so this is an inherent limitation of the migration approach. The test helper (`packages/daemon/tests/unit/helpers/space-test-db.ts:29`) also omits the CHECK. This is fine for now -- TypeScript typing provides the primary guard, and the RPC handler passes the full `CreateSpaceParams`/`UpdateSpaceParams` typed objects through. A future migration could recreate the table with a CHECK if needed.

2. **No application-level validation of `autonomyLevel` in RPC handlers** (`packages/daemon/src/lib/rpc-handlers/space-handlers.ts`): The `space.create` and `space.update` handlers don't validate that `autonomyLevel` is one of the allowed values before passing to the repository. This matches the existing pattern (no validation for `defaultModel`, `allowedModels`, etc.) and is acceptable since the RPC boundary is internal (web client to daemon), not a public API.

3. **`autonomyLevel` is optional on `Space` interface** (`packages/shared/src/types/space.ts:65`): The DB column is `NOT NULL DEFAULT 'supervised'`, so `rowToSpace` always produces a value. Making it required on the `Space` interface would be more precise, but optional is fine since the fallback to `'supervised'` is handled in `rowToSpace` (line 219).

All three are P3 nits -- the code is correct and complete as-is.

lsm (commented):

**Recommendation: APPROVE** (with two actionable notes)

---

### Summary

This PR delivers a well-structured, backward-compatible addition of `SpaceAutonomyLevel`, `SpaceConfig`, and the `autonomy_level` DB column. The migration is idempotent, the repository layer round-trips the new field correctly, and the tests are thorough. Two issues stand out that should be addressed in follow-up — neither is a blocker for this atomic data-layer task.

---

### P1 — Missing CHECK constraint on the new column

**File:** `packages/daemon/src/storage/schema/migrations.ts`, line 2005

The `ALTER TABLE spaces ADD COLUMN autonomy_level` statement omits a CHECK constraint:

```sql
-- actual (no constraint)
ALTER TABLE spaces ADD COLUMN autonomy_level TEXT NOT NULL DEFAULT 'supervised'

-- goals table equivalent (has CHECK)
ALTER TABLE goals ADD COLUMN autonomy_level TEXT NOT NULL DEFAULT 'supervised'
  CHECK(autonomy_level IN ('supervised', 'semi_autonomous'))
```

SQLite's `ALTER TABLE ... ADD COLUMN` does not support adding a CHECK constraint inline, so this is a real limitation of the SQLite migration path. However, the consequence is that an invalid value such as `'fully_autonomous'` can be written into the column and silently persisted. Because TypeScript types are the only gate (and the RPC handler on `space.update` does no value validation), a malformed payload from the network will reach `SpaceRepository.updateSpace`, write the bad string, and `rowToSpace` will cast it to `SpaceAutonomyLevel` without error.

Adding an application-level validation guard in the RPC handler or `SpaceManager` would mitigate this, matching the defence-in-depth posture that the goals handler already provides.

---

### P2 — `space-agent-schema.ts` test helper is stale

**File:** `packages/daemon/tests/unit/helpers/space-agent-schema.ts`, lines 12-28

The `createSpaceAgentSchema` helper's `spaces` table DDL does not include `autonomy_level`:

```ts
// spaces table in space-agent-schema.ts — missing the new column
CREATE TABLE spaces (
  ...
  config TEXT,
  created_at INTEGER NOT NULL,
  updated_at INTEGER NOT NULL
)
```

Tests that use this helper (`space-agent-repository.test.ts`, `space-agent-manager.test.ts`, `space-agent-handlers.test.ts`) don't call `SpaceRepository` directly, so they don't exercise `rowToSpace` and won't fail today. But the helper is explicitly documented as requiring sync with migrations, its own comment says _"Keep in sync with runMigration29 in migrations.ts and space-test-db.ts"_, and `space-test-db.ts` (updated in this PR) already includes the column. Leaving `space-agent-schema.ts` behind creates a silent drift that will cause confusion the next time someone extends the helper.

---

### P3 — No test for `config` null-clearing via `UpdateSpaceParams`

**File:** `packages/daemon/tests/unit/storage/space-repository.test.ts`

`UpdateSpaceParams.config` is typed as `SpaceConfig | undefined`, which means there is no way to clear an existing config back to `null`. This is a pre-existing design, but since this PR introduces typed config it would be a good opportunity to either explicitly test that config cannot be cleared (to document the intent) or to add a `null` variant to the type. Not a blocker, just worth a follow-up note.

---

### What is correct

- **Migration 33** is properly guarded (`tableExists` check + try/catch idiom consistent with the codebase), idempotent, and backward-compatible.
- **`rowToSpace`** safely falls back to `'supervised'` for legacy rows where the column value is `NULL`.
- **`createSpace`** defaults `autonomyLevel` to `'supervised'` at the application level.
- **Type definitions** (`SpaceAutonomyLevel`, `SpaceConfig`, updated interfaces) are clear, well-documented, and correctly exported via `packages/shared/src/mod.ts`.
- **`space-test-db.ts`** was correctly updated to include the column.
- **Test coverage** covers fresh DB, legacy DB path, idempotency, default value, explicit semi-autonomous creation, and both-direction updates — solid.

lsm (commented):

### Summary

Re-review of the two fixes from previous feedback. Both issues have been addressed correctly:

1. **P2 fix (test helper sync):** `space-agent-schema.ts` and `space-test-db.ts` both now include the `autonomy_level TEXT NOT NULL DEFAULT 'supervised'` column, matching the migration. The cross-reference comments were updated to point to migration 33. Good.

2. **P1 fix (validation guards):** `VALID_AUTONOMY_LEVELS` constant and validation in both `space.create` and `space.update` RPC handlers correctly reject invalid values with a descriptive error before they reach the repository. The validation pattern (skip if `undefined`, reject if present but invalid) is correct and matches how TypeScript optional params work at runtime with untyped RPC input.

### Remaining observations

**P3: Constant placement between import blocks** (`space-handlers.ts` line 24)

The `VALID_AUTONOMY_LEVELS` constant is declared between two `import` blocks. This works due to import hoisting but is unconventional -- constants typically go after all imports. This is cosmetic only and the linter/formatter may not flag it, but worth noting for a future cleanup pass.

**P3: `Space.autonomyLevel` is typed as optional but always populated**

The `Space` interface marks `autonomyLevel` as optional (`autonomyLevel?: SpaceAutonomyLevel`), but `rowToSpace()` always provides a value (falling back to `'supervised'`), and the DB column is `NOT NULL DEFAULT 'supervised'`. This means consumers must handle `undefined` unnecessarily. Not blocking -- the optional typing is safer for forward compatibility -- but could be tightened in a follow-up.

### Verdict

Both requested fixes are correctly implemented. The migration, repository, types, RPC validation, and tests are consistent and well-tested. The migration test file covers fresh DB, legacy DB, idempotency, and default value scenarios thoroughly. Repository tests cover create with/without autonomyLevel, update in both directions, and typed config fields.

**Recommendation: APPROVE** (only P3 nits remain)

lsm (commented):

**Recommendation: APPROVE with notes**

The two issues flagged in the previous review have been addressed. The schema test helpers now include `autonomy_level TEXT NOT NULL DEFAULT 'supervised'`, and the `VALID_AUTONOMY_LEVELS` guard has been added to both `space.create` and `space.update` handlers. All 54 tests pass, typecheck is clean, and lint is clean.

Two remaining concerns are noted below.

---

### P2 — `VALID_AUTONOMY_LEVELS` declared between import blocks

**File:** `packages/daemon/src/lib/rpc-handlers/space-handlers.ts`, line 24

The constant is declared between two import groups:

```ts
} from '@neokai/shared';

const VALID_AUTONOMY_LEVELS: SpaceAutonomyLevel[] = ['supervised', 'semi_autonomous'];
import type { DaemonHub } from '../daemon-hub';
```

All imports should come before any module-level declarations. This is non-functional (JavaScript hoists `import` statements regardless of source order), but it is visually confusing and violates standard module layout conventions. The constant should be moved to after the last import statement.

---

### P2 — RPC handler validation has no test coverage

**File:** `packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts`

The `VALID_AUTONOMY_LEVELS` guard added to both `space.create` and `space.update` has no corresponding test cases. The `space.create` tests cover missing `workspacePath`, missing `name`, and SpaceManager propagated errors, but not an invalid `autonomyLevel` value. Same for `space.update`. Since the stated goal of the P1 fix was to ensure invalid values throw a descriptive error before reaching the repository, the absence of tests leaves that contract unverified. Two test cases should be added:

```ts
it('throws on invalid autonomyLevel in space.create', async () => {
  await expect(
    call('space.create', { workspacePath: '/tmp/x', name: 'X', autonomyLevel: 'fully_autonomous' })
  ).rejects.toThrow('Invalid autonomyLevel: fully_autonomous');
});

it('throws on invalid autonomyLevel in space.update', async () => {
  await expect(
    call('space.update', { id: 'space-1', autonomyLevel: 'fully_autonomous' })
  ).rejects.toThrow('Invalid autonomyLevel: fully_autonomous');
});
```

---

### Observation — `autonomyLevel` is optional in `Space` but always populated by the repository

**File:** `packages/shared/src/types/space.ts`, line 65

`autonomyLevel?: SpaceAutonomyLevel` is typed as optional on `Space`, but `rowToSpace` in `space-repository.ts` always assigns a value (`?? 'supervised'`). The `?` creates a misleading impression that callers need to guard for `undefined`, and causes all client-side usage to require a null-check or default. Making it required (`autonomyLevel: SpaceAutonomyLevel`) would be more accurate, but this would be a breaking change requiring callers to always supply the field in test fixtures like `mockSpace`. This is a minor accuracy issue worth tracking but acceptable to leave as-is given the scope of this task.

lsm (commented):

### Re-review after P2 fixes

Both P2 issues from the previous review have been addressed:

1. **`VALID_AUTONOMY_LEVELS` placement** -- Moved to line 34 in `space-handlers.ts`, correctly positioned after all imports and the `Logger` instantiation. Clean.

2. **Invalid `autonomyLevel` rejection tests** -- Two new test cases added (lines 214-222 and 301-305 in `space-handlers.test.ts`) covering the validation path for both `space.create` and `space.update`. Tests use `'fully_autonomous'` as the invalid value and assert the correct error message. Good coverage.

### Full implementation assessment

The overall implementation is correct and complete:

- **Shared types** (`packages/shared/src/types/space.ts`): `SpaceAutonomyLevel` union type, `SpaceConfig` interface, and updated `Space`/`CreateSpaceParams`/`UpdateSpaceParams` are well-defined with good JSDoc.
- **Migration 33** (`migrations.ts`): Properly adds `autonomy_level TEXT NOT NULL DEFAULT 'supervised'` with idempotency guard (try/catch on SELECT). Existing rows get the correct default.
- **Repository** (`space-repository.ts`): INSERT, UPDATE, and `rowToSpace` all handle `autonomyLevel` correctly with proper defaults.
- **RPC handlers** (`space-handlers.ts`): Validation at the handler boundary before data reaches the repository layer.
- **Test helpers** (`space-test-db.ts`, `space-agent-schema.ts`): Schema kept in sync with the new column.
- **Tests**: Migration tests (202 lines, 8 test cases), repository tests (new assertions + 4 new test cases), handler tests (2 new validation tests). Coverage is thorough.

### Minor observation (P3)

The `Space.autonomyLevel` field is typed as optional (`autonomyLevel?: SpaceAutonomyLevel`) in the interface, but `rowToSpace` always assigns a value via `?? 'supervised'` fallback. This means the field is always present at runtime despite the optional typing. This is a defensible choice (the `?` accounts for older serialized data or external consumers that may not set it), but worth noting for awareness.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

This is a follow-up review verifying that both P2 issues flagged in the prior review have been correctly addressed. Both fixes are confirmed correct.

---

### Fix 1: Constant placement — CONFIRMED

**File:** `packages/daemon/src/lib/rpc-handlers/space-handlers.ts`, line 34

`VALID_AUTONOMY_LEVELS` is now declared after all import statements, between the last `import` and the first `export`. The prior placement (between import groups) is gone.

---

### Fix 2: Missing test cases — CONFIRMED

**File:** `packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts`, lines 214–222 and 301–305

Two new test cases were added:

- `space.create` — verifies that passing `autonomyLevel: 'fully_autonomous'` throws `'Invalid autonomyLevel: fully_autonomous'` before reaching `SpaceManager`.
- `space.update` — verifies the same guard on the update path.

Both tests correctly exercise the runtime path (the RPC handler validates before delegating), and the assertions match the exact error message format produced by the handler.

---

### Other previously-addressed items

- `space-agent-schema.ts` includes `autonomy_level TEXT NOT NULL DEFAULT 'supervised'` and comment updated to reference `runMigration33`. Confirmed from prior iteration.
- `space-test-db.ts` includes the column. Confirmed from prior iteration.

---

### Remaining observations (pre-existing, not introduced by this PR)

**P3: `Space.autonomyLevel` typed as optional but always populated**

`packages/shared/src/types/space.ts`, line 65: `autonomyLevel?: SpaceAutonomyLevel` is optional, but `rowToSpace()` always assigns a value (`?? 'supervised'`). This creates unnecessary null-check burden for consumers. The optional typing is the safer forward-compatibility choice, but it remains a minor accuracy issue worth tracking for a future cleanup.

**P3: `setupSpaceHandlers` test call omits two required arguments**

`packages/daemon/tests/unit/rpc-handlers/space-handlers.test.ts`, line 147: The test calls `setupSpaceHandlers(hub, spaceManager, taskRepo, runRepo, daemonHub)` with 5 arguments, but the real function signature requires 7 (`spaceAgentManager` and `spaceWorkflowManager` are missing). This is a pre-existing gap not introduced by this PR. TypeScript's `as unknown as MessageHub` cast in the mock construction likely suppresses the type error, and since the missing managers are only used in handlers not under test here, the tests still pass. Worth fixing in a follow-up.

---

### Verdict

Both requested P2 fixes from the previous review are correctly implemented. The code is ready to merge.

PR #525, opened 2026-03-21
feat: integrate NotificationSink into SpaceRuntime tick loop (Task 2.2)

lsm:
- Add `notificationSink?: NotificationSink` to `SpaceRuntimeConfig` and
  `setNotificationSink()` setter on `SpaceRuntime` for post-construction wiring
- Emit `workflow_run_needs_attention` after catching `WorkflowTransitionError`
  (gate blocked) in `processRunTick()`
- Add explicit `task_needs_attention` detection before the all-completed guard:
  checks `stepTasks.some(t => t.status === 'needs_attention')` and notifies per
  affected task, returning early to prevent advancement
- Add `notifiedTaskSet: Set<string>` deduplication (keyed by `taskId:status`)
  to emit only once per task+status across ticks; clears when task leaves state
- Add timeout detection for `in_progress` tasks using `Space.config.taskTimeoutMs`
  (defaults to disabled when undefined); also deduplicated via `notifiedTaskSet`
- Make `cleanupTerminalExecutors()` async; emit `workflow_run_completed` for runs
  that reach `completed` status before removing their executor from the map
- Add `setNotificationSink()` to `SpaceRuntimeService` delegating to the runtime
- 22 new unit tests covering all 4 event types, deduplication, and no-notification
  cases; all 638 existing space tests continue to pass

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime-service.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime-service.ts
@@ -15,6 +15,7 @@ import type { SpaceAgentManager } from '../managers/space-agent-manager';
 import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
+import type { NotificationSink } from './notification-sink';
 import { SpaceRuntime } from './space-runtime';
 import { Logger } from '../../logger';
 
@@ -83,6 +84,17 @@ export class SpaceRuntimeService {
 			this.start();
 		}
 		return this.runtime;
+	}
+
+	/**
+	 * Wire a notification sink into the underlying SpaceRuntime.
+	 *
+	 * Called after construction once the Space Agent session has been provisioned,
+	 * since SpaceRuntimeService is instantiated before the global agent session exists.
+	 * Delegates directly to the shared SpaceRuntime instance.
+	 */
+	setNotificationSink(sink: NotificationSink): void {
+		this.runtime.setNotificationSink(sink);
 	}
 
 	/**
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -36,6 +36,7 @@ import { SpaceTaskManager } from '../managers/space-task-manager';
 import { WorkflowExecutor, WorkflowTransitionError } from './workflow-executor';
 import { selectWorkflow } from './workflow-selector';
 import { Logger } from '../../logger';
+import { type NotificationSink, NullNotificationSink } from './notification-sink';
 
 const log = new Logger('space-runtime');
 
@@ -61,6 +62,12 @@ export interface SpaceRuntimeConfig {
 	 * Used by start(). Default: 5000 (5 seconds).
 	 */
 	tickIntervalMs?: number;
+	/**
+	 * Sink for structured notification events emitted after mechanical processing.
+	 * Defaults to NullNotificationSink (no-op). Use setNotificationSink() to wire
+	 * in a real sink after construction (e.g. once the Space Agent session exists).
+	 */
+	notificationSink?: NotificationSink;
 }
 
 // ---------------------------------------------------------------------------
@@ -121,7 +128,29 @@ export class SpaceRuntime {
 	/** Handle returned by setInterval when the tick loop is running */
 	private tickTimer: ReturnType<typeof setInterval> | null = null;
 
-	constructor(private config: SpaceRuntimeConfig) {}
+	/** Active notification sink — replaced at runtime via setNotificationSink() */
+	private notificationSink: NotificationSink;
+
+	/**
+	 * Deduplication set for notifications keyed by `taskId:status` (e.g. `task-1:needs_attention`
+	 * or `task-1:timeout`). Prevents re-notifying for the same task+status across ticks.
+	 * Entries are cleared when the task leaves the flagged state.
+	 */
+	private notifiedTaskSet = new Set<string>();
+
+	constructor(private config: SpaceRuntimeConfig) {
+		this.notificationSink = config.notificationSink ?? new NullNotificationSink();
+	}
+
+	/**
+	 * Replace the notification sink at runtime.
+	 *
+	 * Called after construction once the Space Agent session has been provisioned,
+	 * since SpaceRuntimeService is instantiated before the global agent session exists.
+	 */
+	setNotificationSink(sink: NotificationSink): void {
+		this.notificationSink = sink;
+	}
 
 	// -------------------------------------------------------------------------
 	// Lifecycle — start / stop
@@ -182,7 +211,7 @@ export class SpaceRuntime {
 		}
 
 		await this.processCompletedTasks();
-		this.cleanupTerminalExecutors();
+		await this.cleanupTerminalExecutors();
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured integration of `NotificationSink` into the `SpaceRuntime` tick loop. The implementation correctly addresses all acceptance criteria and the test coverage is thorough.

### What works well

- **Correct detection ordering**: `task_needs_attention` is checked via explicit `stepTasks.some(t => t.status === 'needs_attention')` before the all-completed guard, NOT inferred from `WorkflowTransitionError` — matching the acceptance criteria exactly.
- **Deduplication via `notifiedTaskSet`**: Keyed by `taskId:status`, with proper cleanup when tasks leave their flagged state (lines 507-514). This prevents repeated notifications across ticks while allowing re-notification after status transitions.
- **Natural dedup for `workflow_run_needs_attention`**: The `processRunTick` early-return at lines 463-469 (checks `run.status === 'needs_attention'`) prevents re-emission without needing explicit dedup — elegant.
- **Null Object pattern**: `NullNotificationSink` as default means existing code paths are unaffected.
- **`setNotificationSink()` on both `SpaceRuntime` and `SpaceRuntimeService`**: Post-construction wiring is clean.
- **Test quality**: 22 tests covering all 4 event types, dedup, re-notification after state transitions, no-notification for mechanical advancement, fallback reason, and `setNotificationSink()` wiring. All pass.

### Minor observations (P3)

1. **Unbounded `notifiedTaskSet` growth** (`packages/daemon/src/lib/space/runtime/space-runtime.ts`): Dedup entries are only cleared when tasks appear in the current step's task list. Once a run completes and its executor is removed, dedup entries for its tasks remain in the set indefinitely. For a long-running daemon with many completed runs, this is a very slow memory leak. In practice, the leak rate is negligible (one `Set` entry per task that ever needed attention or timed out), but a periodic sweep or scoping by run ID would be cleaner. Not blocking.

2. **No defensive try/catch around `notify()` calls** (`packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 524, 548, 581, 610): The `NotificationSink` interface docs state that implementations should handle errors internally, and the existing `SessionNotificationSink` does so. However, a misbehaving implementation could propagate errors and disrupt the tick loop. The `processCompletedTasks` catch-all (line 444) mitigates this for most call sites, but in `cleanupTerminalExecutors` (line 610) a throwing notify would halt cleanup of remaining runs. This is a contract enforcement concern, not a bug with the current implementation.

3. **`startedAt` type is `number | undefined`** (from `SpaceTask` type): The timeout check at line 543 (`if (task.status !== 'in_progress' || !task.startedAt) continue`) correctly handles the undefined case.

### Verdict

All acceptance criteria are met. The implementation is correct, complete, and well-tested. The `cleanupTerminalExecutors` async change is backwards-compatible since it is a private method only called via `await`. No P0, P1, or P2 issues found.

lsm (commented):

**Recommendation: APPROVE** (with one P1 and one P2 worth noting)

---

### Summary

The implementation is solid and correct. All four notification event types are emitted at the right points, the ordering (task-level `needs_attention` before the all-completed guard) matches the spec, deduplication is correctly scoped via `taskId:status` keys, and the `setNotificationSink()` setter is present on both `SpaceRuntime` and `SpaceRuntimeService`. The 22-test suite provides good behavioral coverage. Two issues are worth flagging.

---

### P1 — `notifiedTaskSet` is never pruned for completed runs (memory leak)

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`

The dedup set clears entries only when the corresponding task reappears in `stepTasks` during a tick (lines 507–513). Once a run completes and its executor is removed in `cleanupTerminalExecutors()`, the task IDs from that run stay in `notifiedTaskSet` forever. In a long-running daemon that processes many workflow runs, this set grows without bound.

Because task IDs are UUIDs (never reused), there is no functional correctness issue — just unbounded memory growth. In practice this is likely small (one entry per timed-out or attention-flagged task), but the fix is straightforward: in `cleanupTerminalExecutors()`, before deleting the executor and meta, iterate the completed run's tasks and delete their keys from `notifiedTaskSet`.

---

### P2 — `setNotificationSink()` on `SpaceRuntimeService` is untested

**File:** `packages/daemon/tests/unit/space/space-runtime-service.test.ts`

`SpaceRuntime.setNotificationSink()` is tested in `space-runtime-notifications.test.ts`, but the delegation path through `SpaceRuntimeService.setNotificationSink()` has no test in `space-runtime-service.test.ts`. A minimal test — construct a `SpaceRuntimeService`, call `setNotificationSink(mockSink)`, then verify the mock receives events from a tick — would close this gap and guard against any future refactor that breaks the delegation.

---

### P3 — `workflow_run_completed` is not emitted for `cancelled` runs

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 606–618

The `WorkflowRunCompletedEvent` type declares `status: 'completed' | 'cancelled' | 'needs_attention'`, but `cleanupTerminalExecutors()` only emits the notification when `run.status === 'completed'`. Cancelled runs are silently cleaned up. This is likely intentional (cancellation is an admin action, not a natural termination), but the type definition creates a semantic gap: consumers handling `workflow_run_completed` with `status: 'cancelled'` will never receive one. If cancellation events are not needed, the `status` union in the type should drop `'cancelled'` to avoid misleading implementers of the interface.

---

### Correctness Checks Passed

- **Detection order**: `task_needs_attention` is checked via `stepTasks.some(t => t.status === 'needs_attention')` before the all-completed guard — not inferred from `WorkflowTransitionError`. Correct per spec.
- **`workflow_run_needs_attention` dedup**: Re-emission is prevented by the run-status early return at line 463–469 (run is already `needs_attention` on subsequent ticks). No separate dedup needed.
- **Timeout guard**: `taskTimeoutMs` is read from `Space.config.taskTimeoutMs`; `undefined` disables the check entirely. Correct.
- **Timeout dedup**: Keys cleared when task leaves `in_progress` (line 511–512). Re-entry re-notifies. Correct.
- **`setNotificationSink()` on SpaceRuntime**: Works correctly; tested in `space-runtime-notifications.test.ts`.
- **No notification on mechanical advancement**: Confirmed by the zero-event tests in the 'no notifications for mechanical advancement' describe block.
- **Existing tests**: The PR claims all 638 existing space tests continue to pass; the implementation does not modify `processCompletedTasks` control flow in a way that would break them.

lsm (commented):

**Recommendation: APPROVE**

---

### Re-review Summary

This is a follow-up review after the worker addressed the P1 and P2 issues from the previous review. Both fixes are correctly implemented.

---

### P1 Fix: Memory Leak in `notifiedTaskSet` -- RESOLVED

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 619-626

The fix correctly prunes dedup entries in `cleanupTerminalExecutors()` by calling `taskRepo.listByWorkflowRun(runId)` for each terminal run and deleting both `taskId:needs_attention` and `taskId:timeout` keys before removing the executor. This prevents unbounded growth of the `notifiedTaskSet` for long-running daemons.

The placement is correct -- pruning happens after the notification emission (lines 607-618) but before the executor/meta deletion (lines 627-628), so there is no risk of losing the ability to emit a final notification.

---

### P2 Fix: Untested `setNotificationSink()` Delegation -- RESOLVED

**File:** `packages/daemon/tests/unit/space/space-runtime-service.test.ts`

Three tests were added:

1. **Existence check** (line ~1007): Verifies `setNotificationSink` is a callable function on the service. Minimal but appropriate as a guard against accidental removal.

2. **Integration: sink receives events** (line ~1107): Uses a real DB, wires a `MockSink` via `svc.setNotificationSink()`, completes a single-step workflow, and asserts `workflow_run_completed` is received. This proves the delegation path works end-to-end.

3. **Integration: sink replacement** (line ~1139): Calls `setNotificationSink()` twice with different sinks, then verifies only the second sink receives events. This validates the replacement semantics documented in the method's JSDoc.

All 38 tests pass across both test files.

---

### P3 (from prior review): Type gap for cancelled runs -- NOT ADDRESSED

The `WorkflowRunCompletedEvent.status` type still includes `'cancelled'` but `cleanupTerminalExecutors()` only emits for `completed` runs. This is a minor semantic gap in the type definition. Not blocking.

---

### Overall Assessment

The implementation is clean and well-tested. The memory leak fix is minimal and correctly scoped. The delegation tests provide meaningful coverage beyond mere existence checks. No new issues introduced.

lsm (commented):

---

**Recommendation: APPROVE** — Both issues addressed correctly. No remaining P0/P1/P2 concerns.

---

### P1 — Memory leak fix (cleanupTerminalExecutors)

The fix in `packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 619-626 is correct and complete.

Before removing a terminal executor entry (both `completed` and `cancelled` paths), the code now calls `this.config.taskRepo.listByWorkflowRun(runId)` and deletes both `taskId:needs_attention` and `taskId:timeout` keys from `notifiedTaskSet`. This covers the full set of keys that the per-tick loop in `processRunTick` could have inserted for tasks belonging to that run.

One minor observation: `listByWorkflowRun` filters on `archived_at IS NULL`, so if a task were archived before the run reached terminal state, its dedup entry would not be pruned here. In practice, tasks are not archived mid-run, so this is not a real concern, but it is worth noting.

The fix applies to all terminal cases (`run === null`, `cancelled`, `completed`) uniformly — only the notification is gated to `completed`-only, while the pruning happens for all three. This is the correct design.

---

### P2 — Test coverage for setNotificationSink() delegation

The three new tests in `packages/daemon/tests/unit/space/space-runtime-service.test.ts` (lines 708-713, 809-839, 841-869) fully address the original concern:

1. **Existence check** (line 709-712) — confirms the method is present on `SpaceRuntimeService`.
2. **Delegation integration test** (lines 809-839) — uses a real DB, wires a `MockSink` via `svc.setNotificationSink()`, runs a tick, and asserts the sink received `workflow_run_completed`. This directly proves the delegation chain `SpaceRuntimeService.setNotificationSink() -> SpaceRuntime.setNotificationSink()` is live.
3. **Sink replacement test** (lines 841-869) — calls `setNotificationSink` twice, then verifies only the second sink receives events and the first receives none. This is a meaningful behavioral assertion.

All three tests are structurally sound. The integration tests use proper `beforeEach`/`afterEach` DB lifecycle management with temp directories, matching the pattern used in `space-runtime-notifications.test.ts`.

---

### No new issues found

The implementation is clean, all deduplication paths are covered, and the test suite is substantive. The PR is ready to merge.

PR #528, opened 2026-03-21
feat: implement TaskAgentManager core (Task 4.1)

lsm:
## Summary

- Add `TaskAgentManager` class in `packages/daemon/src/lib/space/runtime/task-agent-manager.ts` as the central integration point managing Task Agent and sub-session lifecycles
- Implements `SubSessionFactory` interface to bridge `task-agent-tools` MCP tool handlers with live session management
- Export `TaskAgentManager` and `TaskAgentManagerConfig` from `space/index.ts`
- 39 unit tests covering all core lifecycle scenarios

## Key design decisions

- **AgentSession.fromInit()** pattern (not `SessionManager.createSession()`) — follows established `RoomRuntimeService` pattern for non-worker sessions; DB persistence achieved via `db.createSession()` inside `fromInit()`
- **Completion detection** via `DaemonHub session.updated` subscription (SessionObserver pattern) — fires at-most-once when sub-session goes idle after having processed messages
- **Concurrency guard** (`spawningTasks` Set) prevents duplicate Task Agent sessions from tick loop races
- **Monotonic suffix** on session ID collision — handles daemon restart where previous session ID still exists in DB

## Test plan

- [x] `spawnTaskAgent`: basic spawn, streaming start, MCP server attachment, task persistence
- [x] Idempotency: second call returns same session ID without creating duplicates
- [x] Session ID collision avoidance: appends `:1` suffix when base ID exists in DB
- [x] `isTaskAgentAlive`: false for unknown, true for idle/processing/waiting sessions
- [x] `createSubSession`: creates session, starts streaming, stores in subSessions map
- [x] `SubSessionFactory.getProcessingState`: null for unknown, idle+messages = complete
- [x] Completion callbacks: fires once on idle transition, not before messages exist
- [x] Message injection: enqueues to session queue for both task agent and sub-sessions
- [x] Cleanup: removes both task agent + sub-sessions, calls SessionManager.deleteSession
- [x] Error handling: removes from spawningTasks on error, retry works after failure
- [x] All 39 tests pass, typecheck clean, lint clean

### DIFF
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -44,6 +44,8 @@ export { SessionNotificationSink, formatEventMessage } from './runtime/session-n
 export type { SessionNotificationSinkConfig } from './runtime/session-notification-sink';
 export { SpaceRuntimeService } from './runtime/space-runtime-service';
 export type { SpaceRuntimeServiceConfig } from './runtime/space-runtime-service';
+export { TaskAgentManager } from './runtime/task-agent-manager';
+export type { TaskAgentManagerConfig } from './runtime/task-agent-manager';
 
 export { selectWorkflow } from './runtime/workflow-selector';
 export type { WorkflowSelectionContext } from './runtime/workflow-selector';
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -598,10 +598,15 @@ export class TaskAgentManager {
 			`TaskAgentManager: sub-session complete — task ${taskId}, step ${stepId}, session ${subSessionId}`
 		);
 
-		// Find the step task in the DB and mark it completed
-		const stepTasks = this.config.taskRepo
-			.listByWorkflowRun(this.getWorkflowRunId(taskId) ?? '')
-			.filter((t) => t.workflowStepId === stepId && t.taskAgentSessionId === subSessionId);
+		// Find the step task in the DB and mark it completed.
+		// Short-circuit when there is no workflow run to avoid a spurious
+		// listByWorkflowRun('') that would match tasks with a null run ID.
+		const workflowRunId = this.getWorkflowRunId(taskId);
+		const stepTasks = workflowRunId
+			? this.config.taskRepo
+					.listByWorkflowRun(workflowRunId)
+					.filter((t) => t.workflowStepId === stepId && t.taskAgentSessionId === subSessionId)
+			: [];
 
 		if (stepTasks.length > 0) {
 			const stepTask = stepTasks[stepTasks.length - 1];
@@ -679,7 +684,13 @@ export class TaskAgentManager {
 	): Promise<void> {
 		const sessionId = session.session.id;
 		const state = session.getProcessingState();
-		const isBusy = state.status === 'processing' || state.status === 'queued';
+		// 'processing'/'queued' = actively running; 'waiting_for_input' = human gate open.
+		// All three states mean the session cannot safely receive a next_turn message
+		// right now — defer it for replay after the current interaction completes.
+		const isBusy =
+			state.status === 'processing' ||
+			state.status === 'queued' ||
+			state.status === 'waiting_for_input';
 
 		const messageId = generateUUID();
 		const sdkUserMessage: SDKUserMessage = {
@@ -693,7 +704,7 @@ export class TaskAgentManager {
 			},
 		};
 
-		// next_turn + busy → save for replay after current turn
+		// next_turn + busy → save for replay after current turn completes
 		if (deliveryMode === 'next_turn' && isBusy) {
 			this.config.db.saveUserMessage(sessionId, sdkUserMessage, 'saved');
 			return;
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -177,18 +177,31 @@ export class TaskAgentManager {
 		if (this.spawningTasks.has(taskId)) {
 			// Poll until the session appears — the concurrently running spawn will
 			// complete shortly and add it to taskAgentSessions.
+			const CONCURRENT_SPAWN_TIMEOUT_MS = 30_000;
+			const deadline = Date.now() + CONCURRENT_SPAWN_TIMEOUT_MS;
 			return new Promise((resolve, reject) => {
 				const interval = setInterval(() => {
 					const session = this.taskAgentSessions.get(taskId);
 					if (session) {
 						clearInterval(interval);
 						resolve(session.session.id);
+						return;
 					}
 					// Also stop if it's no longer spawning (spawn failed)
 					if (!this.spawningTasks.has(taskId)) {
 						clearInterval(interval);
 						reject(
 							new Error(`Concurrent spawn for task ${taskId} failed before session was created`)
+						);
+						return;
+					}
+					// Timeout guard — prevents indefinite polling if first spawn hangs
+					if (Date.now() >= deadline) {
+						clearInterval(interval);
+						reject(
+							new Error(
+								`Concurrent spawn for task ${taskId} timed out after ${CONCURRENT_SPAWN_TIMEOUT_MS}ms`
+							)
 						);
 					}
 				}, 50);
@@ -201,7 +214,7 @@ export class TaskAgentManager {
 			// --- Generate session ID with collision avoidance on restart
 			const spaceId = space.id;
 			const baseSessionId = `space:${spaceId}:task:${taskId}`;
-			const sessionId = await this.resolveSessionId(baseSessionId);
+			const sessionId = this.resolveSessionId(baseSessionId);
 
 			// --- Create AgentSessionInit
 			const init = createTaskAgentInit({
@@ -248,6 +261,11 @@ export class TaskAgentManager {
 					this.handleSubSessionComplete(taskId, stepId, subSessionId),
 			});
 
+			// setRuntimeMcpServers expects McpServerConfig but the MCP SDK's `Server`
+			// object is structurally compatible at runtime — the AgentSession only reads
+			// the `server` property for the live Server instance. The cast is safe because
+			// createTaskAgentMcpServer returns { server, cleanup } which satisfies the
+			// runtime shape used inside AgentSession.setRuntimeMcpServers().
 			agentSession.setRuntimeMcpServers({
 				'task-agent': mcpServer as unknown as import('@neokai/shared').McpServerConfig,
 			});
@@ -412,10 +430,16 @@ export class TaskAgentManager {
 	 * via SessionManager.deleteSession(), and clears in-memory maps.
 	 */
 	async cleanup(taskId: string): Promise<void> {
+		// Collect the exact session IDs that belong to this task so that
+		// the callback/listener cleanup in steps 3 & 4 uses precise matches
+		// rather than a fragile substring check.
+		const sessionIdsToClean = new Set<string>();
+
 		// 1. Cleanup sub-sessions first
 		const stepMap = this.subSessions.get(taskId);
 		if (stepMap) {
 			for (const [subSessionId, session] of stepMap) {
+				sessionIdsToClean.add(subSessionId);
 				await this.stopAndDeleteSession(subSessionId, session);
 			}
 			this.subSessions.delete(taskId);
@@ -424,20 +448,23 @@ export class TaskAgentManager {
 		// 2. Cleanup Task Agent session
 		const taskAgentSession = this.taskAgentSessions.get(taskId);
 		if (taskAgentSession) {
-			await this.stopAndDeleteSession(taskAgentSession.session.id, taskAgentSession);
+			const agentSessionId = taskAgentSession.session.id;
+			sessionIdsToClean.add(agentSessionId);
+			await this.stopAndDeleteSession(agentSessionId, taskAgentSession);
 			this.taskAgentSessions.delete(taskId);
 		}
 
-		// 3. Remove any dangling completion callbacks
-		for (const [sessionId] of this.completionCallbacks) {
-			if (sessionId.includes(taskId)) {
-				this.completionCallbacks.delete(sessionId);
-			}
+		// 3. Remove any dangling completion callbacks for known session IDs.
+		// We use the exact session IDs collected above (sub-sessions + task agent)
+		// rather than a substring match to avoid false positives.
+		for (const sessionId of sessionIdsToClean) {
+			this.completionCallbacks.delete(sessionId);
 		}
 
-		// 4. Remove session listeners
-		for (const [sessionId, unsub] of this.sessionListeners) {
-			if (sessionId.includes(taskId)) {
+		// 4. Remove session listeners for known session IDs
+		for (const sessionId of sessionIdsToClean) {
+			const unsub = this.sessionListeners.get(sessionId);
+			if (unsub) {
 				unsub();
 				this.sessionListeners.delete(sessionId);
 			}
@@ -614,21 +641,24 @@ export class TaskAgentManager {
 	 * If the base ID exists (e.g., from a previous crashed attempt), appends a
 	 * monotonic suffix until an unused ID is found.
 	 */
-	private async resolveSessionId(baseId: string): Promise<string> {
+	private resolveSessionId(baseId: string): string {
 		// Check if base ID is already free
 		if (!this.config.db.getSession(baseId)) {
 			return baseId;
 		}
 
-		// Append monotonic suffix starting from 1
-		let attempt = 1;
-		while (true) {
+		// Append monotonic suffix starting from 1; cap at 100 to avoid an
+		// unbounded loop if the DB is in an unexpected state.
+		const MAX_ATTEMPTS = 100;
+		for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
 			const candidateId = `${baseId}:${attempt}`;
 			if (!this.config.db.getSession(candidateId)) {
 				return candidateId;
 			}
-			attempt++;
 		}
+		throw new Error(
+			`Could not find available session ID for base "${baseId}" after ${MAX_ATTEMPTS} attempts`
+		);
 	}
 
 	// -------------------------------------------------------------------------
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -37,7 +37,14 @@
  */
 
 import { generateUUID } from '@neokai/shared';
-import type { Space, SpaceTask, SpaceWorkflow, SpaceWorkflowRun, MessageHub } from '@neokai/shared';
+import type {
+	Space,
+	SpaceTask,
+	SpaceWorkflow,
+	SpaceWorkflowRun,
+	MessageHub,
+	McpServerConfig,
+} from '@neokai/shared';
 import type { UUID } from 'crypto';
 import type { SDKUserMessage } from '@neokai/shared/sdk';
 import type { AgentSessionInit } from '../../../lib/agent/agent-session';
@@ -267,7 +274,7 @@ export class TaskAgentManager {
 			// createTaskAgentMcpServer returns { server, cleanup } which satisfies the
 			// runtime shape used inside AgentSession.setRuntimeMcpServers().
 			agentSession.setRuntimeMcpServers({
-				'task-agent': mcpServer as unknown as import('@neokai/shared').McpServerConfig,
+				'task-agent': mcpServer as unknown as McpServerConfig,
 			});
 
 			// --- Persist taskAgentSessionId on the SpaceTask
@@ -318,10 +325,6 @@ export class TaskAgentManager {
 		sessionId: string,
 		init: AgentSessionInit
 	): Promise<string> {
-		// Override sessionId in init with the one we generated — the tool handler
-		// generates a UUID but we override it here to ensure consistent naming.
-		// However, to keep the interface simple we use the init's sessionId directly
-		// since the task agent tool already sets it.
 		const subSession = AgentSession.fromInit(
 			init,
 			this.config.db,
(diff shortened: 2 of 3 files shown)


lsm (commented):

Overall this is a well-structured implementation with good test coverage and clear documentation. The class follows established patterns in the codebase (e.g., `AgentSession.fromInit()`, `RoomRuntimeService` style). I found one P0 bug, one P1 concern, and a couple of P2 suggestions.

---

### P0: Concurrency guard polling has no timeout — can hang forever

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 180-195

The `setInterval` polling loop that waits for a concurrent spawn to complete has no timeout. If the first spawn hangs (e.g., `createOrGetRuntime` or `startStreamingQuery` never resolves), the second caller will poll indefinitely — leaking an interval timer and blocking the caller forever.

```typescript
return new Promise((resolve, reject) => {
    const interval = setInterval(() => {
        // ... no timeout, polls forever if first spawn hangs
    }, 50);
});
```

Add a timeout (e.g., 30 seconds) that rejects with a clear error if the concurrent spawn never completes.

---

### P1: `cleanup()` uses `sessionId.includes(taskId)` for callback/listener cleanup — fragile substring match

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 432-444

Steps 3 and 4 of `cleanup()` iterate completion callbacks and session listeners, deleting entries where `sessionId.includes(taskId)`. This is a substring match that could incorrectly match sessions for a different task if one task ID happens to be a substring of another (e.g., task ID `"abc"` would match session `"space:x:task:abcdef:step:1"`). Since task IDs appear to be UUIDs this is unlikely in practice, but the pattern is brittle.

A safer approach would be to match against the known session IDs from the `subSessions` map (which was already iterated in step 1) and the task agent session ID (from step 2), or use a prefix/delimiter-aware match like `:task:${taskId}:` or `:task:${taskId}$`.

Additionally, deleting entries from a `Map` while iterating over it with `for...of` is implementation-dependent behavior. The ES spec says this is safe for `Map`, but it is worth noting since the intent could be misread.

---

### P2: `resolveSessionId` is declared `async` but performs no awaits

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 617-632

`resolveSessionId` is `async` (and its callers `await` it), but the body is purely synchronous — `this.config.db.getSession()` is synchronous. The unnecessary `async` wraps return values in microtask-resolved promises, adding overhead on every spawn. Consider making it synchronous unless there is an intent to make `db.getSession` async in the future.

---

### P2: `resolveSessionId` unbounded loop

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 624-631

The `while (true)` loop incrementing `attempt` has no upper bound. If the DB is in a corrupted state where `getSession` always returns truthy for any ID, this becomes an infinite loop. A reasonable cap (e.g., 100 attempts) with an error throw would be a safety net.

---

### P2: `as unknown as McpServerConfig` cast

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 252

The double cast `mcpServer as unknown as import('@neokai/shared').McpServerConfig` bypasses type safety entirely. If the MCP server's shape diverges from `McpServerConfig`, this will cause runtime failures with no compile-time warning. Consider adding a shared type or adapter that properly bridges the MCP `Server` type with `McpServerConfig`, or at minimum add a comment explaining why this cast is safe.

---

### Minor observations (P3)

1. **Test helper `makeTask` has unused parameters**: `_db` and `_spaceId` in `makeTask()` at line 189 of the test file are never used. They could be removed to reduce confusion.

2. **`SpaceTaskManager` is instantiated inside `spawnTaskAgent`**: A new `SpaceTaskManager` is created on every spawn (line 227). If spawning is frequent, this could be memoized per space ID.

3. **Test coverage for `handleSubSessionComplete`**: The private `handleSubSessionComplete` method is not directly tested. The completion callback tests verify the callback fires, but they don't verify that the method correctly marks step tasks as completed or notifies the Task Agent. This is acceptable since it exercises internal logic, but an integration-level test would increase confidence.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 1 P0, 3 P1, 2 P2

---

### Summary

The `TaskAgentManager` core implementation is well-structured and follows established patterns in the codebase. The in-memory maps, DB-backed persistence via `AgentSession.fromInit()`, and the `SubSessionFactory` abstraction are all solid design choices. However, there is one correctness bug in the concurrency guard that can lead to an unresolvable Promise, one missing test for the critical `handleSubSessionComplete` path, and a handful of smaller robustness issues.

---

### P0 — Bug: Concurrency guard interval can resolve AND reject on the same tick

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 181–194

The two `if` branches inside the `setInterval` callback are not mutually exclusive. On the very first tick after the concurrent spawn completes, both conditions can be true at once:

1. `this.taskAgentSessions.get(taskId)` returns the session → `clearInterval` + `resolve`
2. `!this.spawningTasks.has(taskId)` is also `true` (the spawner removed it in `finally`) → `clearInterval` + `reject`

When the session is added to `taskAgentSessions` and `spawningTasks` is cleared atomically in a single tick, the second `if` fires immediately after the first and calls `reject()` on an already-resolved Promise. In JavaScript a resolved Promise ignores subsequent `reject()` calls, so no data loss occurs, but this makes the behavior subtly wrong and fragile. More critically, if the spawn *fails* but the session was added to `taskAgentSessions` before the error (lines 259 → caught at 279 → deletes it), the interval will see `session = undefined` and `!spawningTasks.has(...)` at the same time, which is the correct failure path. However if the session is in the map but `spawningTasks` has already been cleared normally (success path), both branches fire — the interval is cleared twice (harmless) and `reject` is called after `resolve` (harmless but confusing). The real bug is: there is no `else` or `return` guard preventing the second `if` from running after the first path already handled the tick. Fix by adding `return` after the `resolve` call inside the first branch.

---

### P1 — Missing test coverage for `handleSubSessionComplete`

The private method `handleSubSessionComplete` is the most complex and stateful logic in this class. It: queries the task repo, marks a step task completed via a freshly constructed `SpaceTaskManager`, and injects a `[STEP_COMPLETE]` message into the Task Agent session. None of this is covered by the unit tests. The PR description lists "Completion propagation" but the test only tests the `onComplete` DaemonHub firing, not the downstream effects (step task status update, Task Agent notification). Given the `listByWorkflowRun(getWorkflowRunId(taskId) ?? '')` path with an empty-string fallback, and the inline `SpaceTaskManager` construction depending on `getSpaceIdForTask`, a regression here would be silent.

---

### P1 — String-matching in `cleanup()` is fragile and can cross-contaminate sessions

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 432–444

```ts
if (sessionId.includes(taskId)) {
```

If a task UUID happens to be a substring of another task UUID (while rare, UUIDs are 36-character strings and can share character sequences), `cleanup(taskA)` could remove `completionCallbacks` and `sessionListeners` for a session belonging to `taskB`. This is especially risky because the `completionCallbacks` and `sessionListeners` maps are keyed by full `subSessionId` strings, and at the time of cleanup the sub-sessions for that task have *already* been removed from `this.subSessions` (step 1 above). The cleanup of steps 3 and 4 is purely a fallback for dangling entries — it should match exactly the session IDs that were in the now-deleted `stepMap`. The fix is to capture the set of session IDs before deleting `stepMap` and iterate over those, or to track which session IDs belong to each task in a separate map.

---

### P1 — `resolveSessionId` has an unbounded `while (true)` loop with no depth limit

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 625–631

If the DB is in a corrupt state where every `space:${spaceId}:task:${taskId}:N` variant is already occupied (e.g., a task was restarted hundreds of times), this loop will spin forever and hang the spawner. A max-attempt guard (e.g., 100 tries, then throw) is a minimal safeguard that prevents a daemon hang.

---

### P2 — `getSpaceIdForTask` uses an unsafe context cast

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 724

```ts
return (session.session.context as { spaceId?: string } | undefined)?.spaceId ?? null;
```

`createTaskAgentInit` sets `context` in a specific way. If the shape changes, this cast silently returns `null`, which causes `handleSubSessionComplete` to skip the `setTaskStatus` call entirely with no observable error — a silent failure. The spaceId should be carried more directly (e.g., in a typed in-memory structure alongside the session, or by looking it up from `taskRepo.getTask(taskId)?.spaceId` if `SpaceTask` carries it). At minimum this should have an invariant assertion.

---

### P2 — MCP server type-cast silences a real type incompatibility

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 252

```ts
'task-agent': mcpServer as unknown as import('@neokai/shared').McpServerConfig,
```

This `as unknown as` double-cast pattern is a known workaround in the codebase (e.g., `provision-global-agent.ts`, `leader-agent.ts`), but using an inline import inside a type cast is harder to maintain than importing `McpServerConfig` at the top of the file. More importantly, there is no comment explaining *why* the cast is needed (what the actual vs. expected type is). Given this pattern exists elsewhere, it is acceptable here, but it should at minimum import the type at the top to avoid the inline import form.

---

### Minor observations (P3)

- The comment on lines 316–317 says "Override sessionId in init with the one we generated" but immediately says "we use the init's sessionId directly". The comment is misleading and should be removed.
- The test accesses the `private` method `createSubSessionFactory` via `as unknown as { createSubSessionFactory: ... }` on multiple tests. If `createSubSessionFactory` is being tested directly, it should be `protected` or have dedicated public surface tested instead. Testing private internals this way means the tests will break on any internal refactor.
- No test for `injectSubSessionMessage` with `next_turn` delivery mode when the session is `processing` — the `saved` path in `injectMessageIntoSession` is untested.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 0 P0, 1 P1, 2 P2, 1 P3

---

### Summary

The previous P0 (infinite polling with no timeout in the concurrency guard) and P1 (fragile substring cleanup) are both properly fixed. The concurrency guard now has a 30-second deadline, proper `return` guards after resolve/reject, and an additional check for spawn failure. The cleanup method now collects exact session IDs into a `Set` before iterating. The `resolveSessionId` method is now synchronous and bounded at 100 attempts. The `as unknown as McpServerConfig` cast now has an explanatory comment. All 39 tests pass.

One P1 issue from the previous review remains unaddressed: `handleSubSessionComplete` has zero test coverage despite being the most complex stateful path in the class.

---

### P1 — `handleSubSessionComplete` still has no test coverage

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 589–633

Both previous reviewers flagged this. The method:
1. Queries the task repo via `listByWorkflowRun(getWorkflowRunId(taskId) ?? '')` — if `getWorkflowRunId` returns `null`, this passes `''` which may match unrelated tasks with empty workflow run IDs
2. Filters by `workflowStepId === stepId && taskAgentSessionId === subSessionId`
3. Constructs a new `SpaceTaskManager` and calls `setTaskStatus`
4. Injects a `[STEP_COMPLETE]` notification into the Task Agent session

None of this is tested. The `getWorkflowRunId(taskId) ?? ''` fallback is particularly concerning — it silently queries with an empty string rather than short-circuiting when there's no workflow run, which could produce false matches. At minimum, add a test that triggers the completion callback through the DaemonHub (already tested) and verifies the downstream effects: that the step task status is updated and the Task Agent receives the `[STEP_COMPLETE]` message.

---

### P2 — `getSpaceIdForTask` uses unsafe context cast with silent null fallback

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 751–755

```typescript
return (session.session.context as { spaceId?: string } | undefined)?.spaceId ?? null;
```

When `getSpaceIdForTask` returns `null`, `handleSubSessionComplete` at line 607 silently skips the `setTaskStatus` call with no log or error. If the `context` shape changes in `createTaskAgentInit`, the step task will never be marked completed and the failure will be invisible. Consider either:
- Storing the `spaceId` in a dedicated `Map<taskId, spaceId>` populated during `spawnTaskAgent`
- Or at minimum logging a warning when `spaceId` is null in `handleSubSessionComplete`

---

### P2 — `next_turn` delivery mode when session is busy is untested

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 694–697

The `injectMessageIntoSession` method has a `next_turn` + `isBusy` branch that saves the message with status `'saved'` and returns early without enqueuing. This code path — which is the one `handleSubSessionComplete` actually uses (line 624: `'next_turn'`) — has no test coverage. A test should verify that when the Task Agent session is in `processing` state and a `next_turn` message is injected, `saveUserMessage` is called with `'saved'` status and the message is NOT enqueued.

---

### P3 — Misleading comment in `createSubSession`

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 321–324

The comment says "Override sessionId in init with the one we generated" but immediately contradicts itself with "we use the init's sessionId directly." The first sentence should be removed.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (own PR — posting as COMMENT per policy)

---

Re-reviewing after the fix commit (`40dc6a1`). The P0 concurrency double-fire, P1 fragile substring cleanup, and P2 synchronous/loop cap items are all correctly addressed. Tests pass (39/39). There remain meaningful gaps worth closing before merge.

---

## Resolved from Previous Review

**P0 — concurrency guard double-fire**: Fixed. Each branch in the `setInterval` callback now has a `return` after `clearInterval`, so only one of resolve/reject can fire per tick. Correct.

**P1 — fragile substring cleanup**: Fixed. `cleanup()` now builds an explicit `sessionIdsToClean` set from the actual map entries and uses precise key lookups. No cross-contamination risk.

**P2 — unbounded loop / sync resolveSessionId**: Fixed. The suffix loop is capped at 100 with a thrown error, and `resolveSessionId` is synchronous.

**P2 — McpServerConfig cast comment**: Added. The cast rationale is documented.

---

## Remaining Issues

### P1 — `handleSubSessionComplete` has zero test coverage

`handleSubSessionComplete` is the critical completion pathway — it marks step tasks as `'completed'` in the DB and injects the `[STEP_COMPLETE]` nudge message into the Task Agent. It is the most semantically important logic in this class, but there is not a single test exercising it.

```
packages/daemon/src/lib/space/runtime/task-agent-manager.ts:589
private async handleSubSessionComplete(taskId, stepId, subSessionId): Promise<void>
```

Specifically untested paths:
- Happy path: marks step task as `'completed'` and enqueues `[STEP_COMPLETE]` to Task Agent
- `getWorkflowRunId` returns null — silently skips DB update (tested nowhere)
- `getSpaceIdForTask` returns null — silently skips DB update (tested nowhere)
- Step task not found in repo — skips without error (tested nowhere)
- `taskManager.setTaskStatus` throws — error is swallowed via `log.warn` (tested nowhere)
- Task Agent session missing at completion time — silently skips notification (tested nowhere)

The PR description claims coverage of "Sub-session completion propagation to Task Agent" but the only related tests exercise the `onComplete` *callback registration* (via DaemonHub events), not the downstream side-effects inside `handleSubSessionComplete`. These are two different things.

### P1 — Concurrency guard path is not tested end-to-end

The test suite only verifies `isSpawning` returns `false` after spawn completes. There is no test that actually triggers the concurrency guard branch — i.e., two concurrent `spawnTaskAgent` calls on the same `taskId` where the second one goes through the `setInterval` poll loop and either resolves or rejects. This is the most complex code path in the class (including the timeout and failure-rejection sub-paths) and it has no direct test coverage.

### P2 — `next_turn` delivery mode is untested

The PR description lists "Message injection: enqueues to session queue for both task agent and sub-sessions" as covered, but there is no test for the `next_turn` path in `injectMessageIntoSession`. Specifically, the branch at line 694:

```ts
// packages/daemon/src/lib/space/runtime/task-agent-manager.ts:694
if (deliveryMode === 'next_turn' && isBusy) {
    this.config.db.saveUserMessage(sessionId, sdkUserMessage, 'saved');
    return;
}
```

This is the path taken by `handleSubSessionComplete` (via `'next_turn'` mode). It needs a test verifying that when a session is busy, the message is saved with `'saved'` status rather than enqueued.

### P2 — `next_turn` is enqueued immediately when session is `waiting_for_input`

In `injectMessageIntoSession`, `isBusy` only checks `'processing'` and `'queued'` statuses:

```ts
const isBusy = state.status === 'processing' || state.status === 'queued';
```

When `deliveryMode === 'next_turn'` and session status is `'waiting_for_input'` or `'interrupted'`, `isBusy` is `false`, so the message is enqueued immediately rather than deferred. This is likely unintended for `waiting_for_input` — the Task Agent is blocked waiting for human input, and injecting the step-complete notification at that moment could disrupt the human gate flow.

### P3 — `getSpaceIdForTask` is an indirect accessor for data already available

`getSpaceIdForTask` resolves the space ID by reading `session.session.context.spaceId` via an unsafe cast. The space ID is available directly from `space.id` which is passed into `spawnTaskAgent` — storing it as part of the config or in a `taskId → spaceId` map at spawn time would be more robust than relying on a cast from opaque session context.

---

## Summary

The three blocking issues from the previous review are fully resolved. The primary outstanding gap is test coverage: `handleSubSessionComplete` — the class's core completion side-effect — has no tests at all despite being listed as covered in the PR description. The concurrent spawn path similarly has no end-to-end test. These are P1 gaps that should be addressed before merge.

lsm (commented):

**Recommendation: APPROVE**

Third round review of the TaskAgentManager core implementation. The previous review feedback has been thoroughly addressed.

### What was reviewed

- `packages/daemon/src/lib/space/runtime/task-agent-manager.ts` (757 lines) -- the core manager
- `packages/daemon/tests/unit/space/task-agent-manager.test.ts` (1075 lines) -- 42 tests
- `packages/daemon/src/lib/space/index.ts` -- barrel exports

### Changes since last review

All items from the second review round have been addressed:

1. **handleSubSessionComplete tests** -- Three new tests added (lines 780-907): verifies step task marked completed, `[STEP_COMPLETE]` notification injected into Task Agent, and no-throw on missing step task.

2. **getSpaceIdForTask** -- Now uses a clean `taskRepo.getTask(taskId)?.spaceId` lookup (line 755) instead of the previous unsafe context cast. This is correct and safe.

3. **McpServerConfig import** -- Moved to top-level import from `@neokai/shared` (line 46), matching the same type used by `AgentSession.setRuntimeMcpServers()`.

4. **workflowRunId short-circuit** -- `handleSubSessionComplete` now guards against empty `workflowRunId` before calling `listByWorkflowRun('')` (lines 604-609), preventing spurious matches.

### Verification

- All 42 tests pass (0 failures)
- TypeScript typecheck passes cleanly
- Exports are correctly added to the barrel file

### Assessment

The implementation is well-structured with:

- Clear session hierarchy (Task Agent -> Sub-sessions) documented in the module docstring
- Proper concurrency guards via `spawningTasks` set with timeout
- Idempotent `spawnTaskAgent` with session ID collision avoidance
- Completion detection via DaemonHub `session.updated` subscription with fire-once semantics
- Thorough cleanup that collects exact session IDs rather than using fragile substring matching
- Good error handling throughout -- `try/catch` with logging at system boundaries, errors don't propagate and crash the manager

**P3 nit (non-blocking):** The `as unknown as McpServerConfig` cast on line 277 is a known structural mismatch documented in the adjacent comment. It works at runtime because AgentSession only reads the `server` property. This is acceptable for now but would benefit from a proper adapter type if the pattern is reused elsewhere. Not blocking.

No P0, P1, or P2 issues found.

lsm (commented):

**Recommendation: APPROVE** (reviewer is PR author — submitting as COMMENT)

---

All three previous rounds of fixes have landed cleanly. Typecheck passes, lint is clean, 46 tests pass. The implementation is solid and well-structured. I have a few minor observations below.

---

### Remaining minor issues

**P3 — Stale inner-map key comment in field JSDoc (line 119)**

`packages/daemon/src/lib/space/runtime/task-agent-manager.ts` line 119 reads:

```
Maps taskId → (stepId → AgentSession) for sub-sessions.
```

In practice the inner-map key is the full session ID (e.g. `space:spaceId:task:taskId:step:stepId`), not a bare `stepId`. The class-level module doc at line 18 has the same inaccuracy. This is also reflected in the comment at line 337–339 inside `createSubSession`, which says "Determine step ID from session convention" before doing nothing with a step ID — just using `sessionId` directly. These should read "sessionId → AgentSession" to match the actual code.

**P3 — `as unknown as McpServerConfig` cast is consistent with existing codebase pattern**

The cast at line 277 mirrors the identical pattern in `room-runtime-service.ts` line 450. The comment explaining the cast is accurate at runtime (`createSdkMcpServer` returns `McpSdkServerConfigWithInstance` which is already in the `McpServerConfig` union per `sdk.d.ts`). Not a bug.

**P3 — `handleSubSessionComplete` test only exercises the idle-session path**

The "injects [STEP_COMPLETE] notification" test passes because the mock session starts idle. There is no test covering the busy-session path (where `deliveryMode: 'next_turn' + isBusy` causes the message to be saved rather than enqueued). The `_enqueuedMessages` array would not be incremented in that path. This is acceptable coverage for now but worth noting.

**P3 — Minor: PR body says "All 42 tests pass" but 46 pass**

The test count in the PR description does not match what bun reports (46 pass). No functional impact.

---

### What was verified

- All previous-review issues are resolved: `getSpaceIdForTask` uses clean `taskRepo.getTask` lookup; `McpServerConfig` is a top-level import; misleading comment in `createSubSession` was removed; three `handleSubSessionComplete` tests added and passing.
- The `as unknown as McpServerConfig` cast is consistent with the existing room-runtime-service pattern and is safe at runtime since `McpSdkServerConfigWithInstance` is a member of the `McpServerConfig` union.
- The concurrency guard has no interval leak — all three exit paths (`resolve`, `reject on spawn failure`, `reject on timeout`) call `clearInterval`.
- `injectMessageIntoSession`'s extended `isBusy` check (adds `waiting_for_input`) is a deliberate and reasonable divergence from the room-runtime-service pattern.
- The `workflowRunId` null-guard added in `handleSubSessionComplete` correctly avoids the spurious `listByWorkflowRun('')` call.
- 46/46 tests pass, typecheck clean, lint clean.

PR #529, opened 2026-03-21
feat: add needs_attention detection for standalone tasks in SpaceRuntime

lsm:
Extends SpaceRuntime.executeTick() with checkStandaloneTasks() that
processes tasks without a workflowRunId after each workflow tick:
- Emits task_needs_attention for standalone tasks in needs_attention state
- Emits task_timeout for standalone in_progress tasks exceeding taskTimeoutMs
- Uses the shared notifiedTaskSet for deduplication (keyed by taskId:status)
- Documents the restart re-notification contract in notifiedTaskSet comment

Adds 14 new unit tests covering: needs_attention notification, fallback
reason, dedup across ticks, re-notification after state change, timeout
detection, timeout dedup, workflow task isolation, and restart contract.

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -654,6 +654,11 @@ export class SpaceRuntime {
 	 * is never notified twice in a row. Dedup keys are cleared when the task leaves the
 	 * flagged state, allowing re-notification if the task cycles back into it.
 	 *
+	 * Dedup cleanup includes archived tasks (fetched via includeArchived=true) to prevent
+	 * notifiedTaskSet from accumulating stale keys for tasks that were archived while in
+	 * a flagged state. Archived tasks can never re-enter needs_attention or in_progress,
+	 * so their dedup keys are always safe to remove.
+	 *
 	 * Restart contract: because `notifiedTaskSet` is in-memory only, tasks already in
 	 * `needs_attention` at daemon startup will re-notify once on the first tick. This is
 	 * intentional — the Space Agent session is new after restart and needs to be informed
@@ -663,26 +668,29 @@ export class SpaceRuntime {
 		const spaces = await this.config.spaceManager.listSpaces(false);
 
 		for (const space of spaces) {
-			// Fetch all non-archived standalone tasks for this space in a single query.
-			// Filtering by workflowRunId === undefined identifies tasks that are not
-			// associated with any workflow run (i.e., standalone tasks).
-			const standaloneTasks = this.config.taskRepo
-				.listBySpace(space.id)
-				.filter((t) => !t.workflowRunId);
+			// Fetch all standalone tasks including archived ones for the dedup cleanup pass.
+			// Using listStandaloneBySpace pushes workflow_run_id IS NULL into SQL so only
+			// standalone tasks are returned — no JS-side filtering needed.
+			// includeArchived=true ensures archived tasks have their dedup keys cleared and
+			// do not accumulate as stale entries in notifiedTaskSet indefinitely.
+			const allStandalone = this.config.taskRepo.listStandaloneBySpace(space.id, true);
+			const activeStandalone = allStandalone.filter((t) => !t.archivedAt);
 
 			// Dedup cleanup: clear keys for tasks that have left their flagged state.
-			// This mirrors the per-tick cleanup done in processRunTick() for workflow tasks.
-			for (const task of standaloneTasks) {
-				if (task.status !== 'needs_attention') {
+			// Archived tasks always get their keys cleared — they can never re-enter a
+			// flagged state, so keeping their keys would be a permanent memory leak.
+			for (const task of allStandalone) {
+				const archived = !!task.archivedAt;
+				if (archived || task.status !== 'needs_attention') {
 					this.notifiedTaskSet.delete(`${task.id}:needs_attention`);
 				}
-				if (task.status !== 'in_progress') {
+				if (archived || task.status !== 'in_progress') {
 					this.notifiedTaskSet.delete(`${task.id}:timeout`);
 				}
 			}
 
-			// Emit task_needs_attention for standalone tasks in needs_attention state.
-			for (const task of standaloneTasks) {
+			// Emit task_needs_attention for active standalone tasks in needs_attention state.
+			for (const task of activeStandalone) {
 				if (task.status !== 'needs_attention') continue;
 				const dedupKey = `${task.id}:needs_attention`;
 				if (!this.notifiedTaskSet.has(dedupKey)) {
@@ -697,11 +705,11 @@ export class SpaceRuntime {
 				}
 			}
 
-			// Timeout detection for standalone in_progress tasks.
+			// Timeout detection for active standalone in_progress tasks.
 			const taskTimeoutMs = space.config?.taskTimeoutMs;
 			if (taskTimeoutMs !== undefined) {
 				const now = Date.now();
-				for (const task of standaloneTasks) {
+				for (const task of activeStandalone) {
 					if (task.status !== 'in_progress' || !task.startedAt) continue;
 					const elapsedMs = now - task.startedAt;
 					if (elapsedMs > taskTimeoutMs) {
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -135,6 +135,11 @@ export class SpaceRuntime {
 	 * Deduplication set for notifications keyed by `taskId:status` (e.g. `task-1:needs_attention`
 	 * or `task-1:timeout`). Prevents re-notifying for the same task+status across ticks.
 	 * Entries are cleared when the task leaves the flagged state.
+	 *
+	 * Restart contract: this set is in-memory only and starts empty on every daemon restart.
+	 * Tasks already in `needs_attention` at restart time will be re-notified once on the first
+	 * tick. This is intentional: the Space Agent session is also new after restart and needs to
+	 * learn about outstanding issues. No DB persistence for dedup state is required.
 	 */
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured addition that extends standalone task monitoring into the existing tick loop. The implementation correctly mirrors the patterns already established for workflow tasks in `processRunTick()`.

### What was reviewed

- `packages/daemon/src/lib/space/runtime/space-runtime.ts` — new `checkStandaloneTasks()` method and its integration into `executeTick()`
- `packages/daemon/tests/unit/space/space-runtime-notifications.test.ts` — 14 new tests covering standalone notifications

### Correctness

The implementation correctly:
- Filters standalone tasks via `!t.workflowRunId`, preventing double-notification of workflow tasks
- Reuses the shared `notifiedTaskSet` for dedup, with the same `taskId:status` key pattern used by `processRunTick()`
- Cleans up dedup keys when tasks leave their flagged state, enabling re-notification on state cycling
- Guards timeout detection behind `taskTimeoutMs !== undefined` and `task.startedAt` presence
- Documents the intentional restart re-notification contract clearly in both field comment and method JSDoc

### Test coverage

The 14 tests are thorough and cover:
- Happy path for both `needs_attention` and `task_timeout` notifications
- Fallback reason when `task.error` is null
- Dedup across multiple ticks
- Re-notification after state cycling (leave and re-enter flagged state)
- Negative cases: pending tasks, in-progress without timeout, undefined `taskTimeoutMs`
- Workflow task isolation (no double-notification)
- Restart re-notification for both notification types

All 35 tests in the file pass.

### Minor observations (P3)

1. **Archived task dedup leak** (`space-runtime.ts:661`): `listBySpace` excludes archived tasks. If a standalone task is archived while in `needs_attention` or timed-out `in_progress` state, its dedup key stays in `notifiedTaskSet` permanently. This is the same behavior as the existing workflow path in `cleanupTerminalExecutors` (which also wouldn't clean up archived tasks). In practice this is negligible — the set would accumulate at most a handful of orphan strings over the daemon's lifetime.

2. **Redundant `listSpaces` call** (`space-runtime.ts:655`): `checkStandaloneTasks` calls `listSpaces(false)` every tick. This is a DB query per tick. For a small number of spaces this is fine, and it matches the pattern used by `rehydrateExecutors`. Could be optimized later if needed by caching the result within a single tick, but not worth addressing now.

No blocking issues found. The code is well-documented, follows existing patterns, and the tests are comprehensive.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P2 issues, 2 P3 issues)

---

### Summary

The implementation is correct and well-structured. The core logic — dedup cleanup before emission, workflow task isolation via `!t.workflowRunId`, the `listSpaces` approach for config access — all work as intended. All 35 new tests pass. However, two issues need attention before merging.

---

### P2 — Memory leak in `notifiedTaskSet` for archived standalone tasks

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 667–674 (dedup cleanup loop in `checkStandaloneTasks`)

`listBySpace(space.id)` (line 662) excludes archived tasks by default (`archived_at IS NULL` in the SQL). This means the dedup cleanup loop that iterates `standaloneTasks` will never see a task that has been archived. If a standalone task had a `taskId:needs_attention` or `taskId:timeout` entry in `notifiedTaskSet` at the time it was archived, those keys persist in memory indefinitely.

The workflow path handles this symmetrically in `cleanupTerminalExecutors()` (lines 630–633), which explicitly deletes dedup entries for all tasks in a run when the run exits. The standalone path has no equivalent.

For long-running daemons in spaces with frequent task archival, `notifiedTaskSet` will accumulate stale string keys without bound. The entries are small, but it is a correctness gap. The fix is to add a cleanup pass for tasks transitioning to archived (or terminal) states before they leave the `listBySpace` result set.

---

### P2 — Full table scan per tick: `listBySpace` fetches all tasks then JS-filters

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 661–663

`checkStandaloneTasks` calls `this.config.taskRepo.listBySpace(space.id)` which returns every non-archived task in the space, then discards all tasks where `workflowRunId` is set via JS-side `.filter((t) => !t.workflowRunId)`. In spaces with many workflow tasks this means fetching rows that are immediately thrown away.

The workflow path uses targeted queries — `listByWorkflowRun(runId)` and `listByWorkflowRun(runId).filter(t => t.workflowStepId === currentStep.id)` — which are far more selective. The standalone path should use a dedicated repository method (e.g. `listStandaloneBySpace(spaceId)`) that pushes `workflow_run_id IS NULL` into SQL. `listByStatus` already exists but doesn't filter on `workflowRunId`, so a new method or query variant is needed.

---

### P3 — No test for archived standalone task dedup cleanup

**File:** `packages/daemon/tests/unit/space/space-runtime-notifications.test.ts`

The dedup cleanup path for archived tasks (described in P2 above) is not covered by any test. Adding a test that archives a standalone task that has an active dedup entry and then asserts the key is removed would both catch the current bug and guard against regressions once fixed.

---

### P3 — No test for standalone tasks in terminal states

**File:** `packages/daemon/tests/unit/space/space-runtime-notifications.test.ts`

There is no test asserting that standalone tasks in `completed` or `cancelled` status emit no notifications. The logic is correct (neither status matches the `needs_attention` or `in_progress` guards), but covering this explicitly would guard against future status handling changes.

---

- Dedup cleanup-before-emit ordering is correct throughout the method.
- The `!t.workflowRunId` guard correctly isolates standalone tasks from the workflow notification path; the "workflow tasks are NOT processed by checkStandaloneTasks" test verifies this.
- Restart re-notification contract is clearly documented at the `notifiedTaskSet` field level and at the method level, and tested with simulated restarts.
- Fallback reason (`task.error ?? 'Task requires attention'`) mirrors the workflow path exactly.
- `space.config?.taskTimeoutMs` access correctly handles spaces without config.
- All 35 tests pass; the test structure (arrange, act, assert with clear tick sequencing) is easy to follow.

lsm (commented):

**Recommendation: APPROVE**

Both P2 issues from the prior review have been addressed correctly:

### P2 fix 1: SQL-level standalone filter
`SpaceTaskRepository.listStandaloneBySpace(spaceId, includeArchived?)` now pushes `workflow_run_id IS NULL` into SQL (`space-task-repository.ts:96`), eliminating the previous JS-side filter over the full `listBySpace()` result. The method signature with `includeArchived` defaulting to `false` is clean and consistent with the existing `listBySpace` pattern.

### P2 fix 2: Archived task dedup key cleanup
`checkStandaloneTasks()` fetches with `includeArchived=true` for the cleanup pass (`space-runtime.ts:668`), then filters to active tasks for notification emission (`space-runtime.ts:669`). Archived tasks always have their dedup keys cleared (`space-runtime.ts:674-681`), preventing permanent memory leaks. The ordering is correct — cleanup runs before emission, so a task that transitions from `needs_attention` to archived in a single tick will have its key cleared and never re-notified.

### Code quality observations
- The standalone implementation mirrors the workflow-side notification code (same dedup pattern, same notification payloads, same timeout check structure), keeping the codebase consistent.
- Comments are thorough and explain the "why" — restart contract, dedup semantics, and the includeArchived rationale are all documented inline.
- 38 tests all pass, covering: basic notification, dedup, re-notification after state cycling, fallback reason, pending/completed/cancelled no-ops, timeout with and without config, archived task dedup cleanup, workflow task exclusion from standalone path, and restart re-notification behavior.

- `taskTimeoutMs` of `0` would cause every `in_progress` task with any `startedAt` to immediately timeout (since `elapsedMs > 0` is always true for any elapsed time). This matches the existing workflow-side behavior so it is consistent, but worth noting if `0` could be a misconfiguration.

No blocking issues found. The implementation is correct, well-tested, and the prior feedback has been fully addressed.

lsm (commented):

**Recommendation: APPROVE**

Both P2 issues from the prior review have been addressed correctly.

### P2 Fix 1: SQL-level standalone filter

`SpaceTaskRepository.listStandaloneBySpace(spaceId, includeArchived?)` was added at lines 95–105 of `packages/daemon/src/storage/repositories/space-task-repository.ts`. The SQL is `WHERE space_id = ? AND workflow_run_id IS NULL` with an optional `AND archived_at IS NULL`, which is exactly what was asked for. `checkStandaloneTasks()` now calls `this.config.taskRepo.listStandaloneBySpace(space.id, true)` — no JS-side filter over the full `listBySpace()` result.

### P2 Fix 2: Dedup key leak for archived tasks

`checkStandaloneTasks()` fetches `includeArchived=true` and runs the cleanup loop over `allStandalone` before emitting any notifications. For each task, archived status unconditionally clears both `taskId:needs_attention` and `taskId:timeout` keys. Active tasks that have left the flagged state also have their keys cleared. The cleanup-before-emit ordering is correct: a task that transitions out of `needs_attention` and back in within the same tick (not a realistic scenario, but logically sound) would still get re-notified.

### Tests

38 tests pass. The two new tests specifically target the P2 scenarios:

- `archiving a standalone task clears its dedup key (no permanent leak)` — archives a `needs_attention` task and verifies the dedup key is cleared on the next tick, then confirms a fresh task in the same state still gets notified correctly.
- The restart re-notification describe block (`restart re-notification (empty dedup set on restart)`) covers both `needs_attention` and `task_timeout` restart behavior with fresh `SpaceRuntime` instances.

- The `makeRuntime` helper in the test uses the outer `sink` by default, but the restart tests construct `freshSink` and pass it explicitly — correct.
- The dedup cleanup loop iterates over `allStandalone` (includes archived) while notification loops only iterate `activeStandalone` (excludes archived). The two-list approach is clear and there is no risk of notifying archived tasks.
- `listStandaloneBySpace` uses `this.db.prepare(query)` with dynamic string concatenation. While the input is a boolean (not user-controlled), the pattern is consistent with `listBySpace` in the same class, so no concern here.

No new issues found. The two P2 fixes are correct and well-tested.

PR #530, opened 2026-03-21
feat: wire SessionNotificationSink into provisionGlobalSpacesAgent (Task 5.2)

lsm:
- Add `sessionFactory: SessionFactory` to `ProvisionGlobalSpacesAgentDeps`
- After the spaces:global session is confirmed, create a `SessionNotificationSink`
  targeting that session and call `spaceRuntimeService.setNotificationSink(sink)`
- Add `SessionManager.injectMessage()` as the backbone for the session factory
  adapter — delegates to `MessagePersistence.persist()` for DB persistence and
  SDK query feeding
- Build a minimal `SessionFactory` adapter in `rpc-handlers/index.ts` that wraps
  `sessionManager.injectMessage()` for production wiring
- Add 5 integration-style unit tests:
  - Sink is wired (setNotificationSink called) after provisioning
  - Wired sink is a non-null SessionNotificationSink
  - Wiring order: createSession → setNotificationSink
  - Works on daemon restart (session already exists)
  - End-to-end: tick producing task_needs_attention injects message into spaces:global

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -5,7 +5,7 @@
  * Organized by domain for better maintainability.
  */
 
-import type { MessageHub } from '@neokai/shared';
+import type { MessageHub, MessageDeliveryMode } from '@neokai/shared';
 import type { DaemonHub } from '../daemon-hub';
 import type { SessionManager } from '../session-manager';
 import type { AuthManager } from '../auth-manager';
@@ -289,7 +289,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 			injectMessage: (
 				sessionId: string,
 				message: string,
-				opts?: { deliveryMode?: import('@neokai/shared').MessageDeliveryMode }
+				opts?: { deliveryMode?: MessageDeliveryMode }
 			) => deps.sessionManager.injectMessage(sessionId, message, opts),
 			hasSession: (sessionId: string) => deps.sessionManager.getSession(sessionId) !== null,
 			// Remaining SessionFactory methods are not needed for notification injection
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -282,12 +282,34 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	setupGlobalSpacesHandlers(deps.messageHub, globalSpacesState);
 
 	if (process.env.NODE_ENV !== 'test') {
+		// Build a minimal SessionFactory adapter so SessionNotificationSink can inject messages
+		// into the spaces:global session. The adapter delegates to SessionManager.injectMessage()
+		// which handles DB persistence, UI publishing, and SDK query feeding.
+		const globalSessionFactory = {
+			injectMessage: (
+				sessionId: string,
+				message: string,
+				opts?: { deliveryMode?: import('@neokai/shared').MessageDeliveryMode }
+			) => deps.sessionManager.injectMessage(sessionId, message, opts),
+			hasSession: (sessionId: string) => deps.sessionManager.getSession(sessionId) !== null,
+			// Remaining SessionFactory methods are not needed for notification injection
+			createAndStartSession: async () => {},
+			answerQuestion: async () => false as const,
+			createWorktree: async () => null,
+			restoreSession: async () => false as const,
+			startSession: async () => false as const,
+			setSessionMcpServers: () => false as const,
+			removeWorktree: async () => false as const,
+			getProcessingState: (_sessionId: string) => undefined,
+		};
+
 		provisionGlobalSpacesAgent({
 			sessionManager: deps.sessionManager,
 			spaceManager: deps.spaceManager,
 			spaceAgentManager: deps.spaceAgentManager,
 			spaceWorkflowManager,
 			spaceRuntimeService,
+			sessionFactory: globalSessionFactory,
 			taskRepo: spaceTaskRepo,
 			workflowRunRepo: spaceWorkflowRunRepo,
 			state: globalSpacesState,
--- a/packages/daemon/src/lib/session/session-manager.ts
+++ b/packages/daemon/src/lib/session/session-manager.ts
@@ -12,7 +12,8 @@
  * - Background task tracking for cleanup
  */
 
-import type { Session, MessageHub } from '@neokai/shared';
+import type { Session, MessageHub, MessageDeliveryMode } from '@neokai/shared';
+import { generateUUID } from '@neokai/shared';
 import type { DaemonHub } from '../daemon-hub';
 import type { Database } from '../../storage/database';
 import { AgentSession } from '../agent/agent-session';
@@ -232,6 +233,26 @@ export class SessionManager {
 	 */
 	async getSessionAsync(sessionId: string): Promise<AgentSession | null> {
 		return this.sessionCache.getAsync(sessionId);
+	}
+
+	/**
+	 * Inject a message into a session bypassing the RPC/UI message flow.
+	 *
+	 * Used for internal daemon-to-session communication (e.g. SpaceRuntime → global agent).
+	 * Delegates to MessagePersistence so the message is persisted to DB and the session
+	 * query is started/notified exactly like a user-sent message.
+	 */
+	async injectMessage(
+		sessionId: string,
+		message: string,
+		opts?: { deliveryMode?: MessageDeliveryMode }
+	): Promise<void> {
+		await this.messagePersistence.persist({
+			sessionId,
+			messageId: generateUUID(),
+			content: message,
+			deliveryMode: opts?.deliveryMode,
+		});
 	}
 
 	listSessions(options?: { status?: string; includeArchived?: boolean }): Session[] {
--- a/packages/daemon/src/lib/space/provision-global-agent.ts
+++ b/packages/daemon/src/lib/space/provision-global-agent.ts
@@ -19,6 +19,8 @@ import type { SpaceRuntimeService } from './runtime/space-runtime-service';
 import { Logger } from '../logger';
 import { buildGlobalSpacesAgentPrompt } from './agents/global-spaces-agent';
 import { createGlobalSpacesMcpServer, type GlobalSpacesState } from './tools/global-spaces-tools';
+import { SessionNotificationSink } from './runtime/session-notification-sink';
+import type { SessionFactory } from '../room/runtime/task-group-manager';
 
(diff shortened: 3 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR correctly wires `SessionNotificationSink` into the global agent provisioning flow. The implementation satisfies all acceptance criteria.

### What works well

1. **Wiring order is correct**: Session is created/confirmed first (lines 72-93 of `provision-global-agent.ts`), then `SessionNotificationSink` is constructed and `setNotificationSink()` is called (lines 98-103). This matches the required pattern.

2. **No changes to `SpaceRuntimeServiceConfig` constructor signature**: The `setNotificationSink()` setter pattern is used as required — the service is constructed before the session exists, and the sink is injected afterward.

3. **`SessionManager.injectMessage()`** is a clean addition that delegates to `MessagePersistence.persist()` with a generated UUID, content, and optional `deliveryMode`. This provides the right level of abstraction for internal daemon-to-session messaging.

4. **The `SessionFactory` adapter** in `rpc-handlers/index.ts` is minimal and well-documented, with stub implementations for unused methods. The two methods that matter (`injectMessage` and `hasSession`) are properly wired.

5. **Tests are comprehensive**: 5 tests cover the core wiring, order verification, restart idempotency, and the end-to-end tick-to-session flow. The end-to-end test (`tick producing task_needs_attention event injects a message`) validates the full chain from `executeTick()` through to `sessionFactory.injectMessage()`.

### Minor observations (P3)

1. **`provision-global-agent.test.ts` line 28**: The import `import type { SessionManager } from '../../../src/lib/session-manager.ts';` uses the re-export barrel file. This works but is inconsistent with `provision-global-agent.ts` itself which imports from `'../session-manager'`. Not a problem — just a style observation.

2. **`rpc-handlers/index.ts` lines 296-303**: The stub methods on the `globalSessionFactory` use `as const` return types (e.g., `async () => false as const`). This is fine for type narrowing but is slightly unusual for no-op stubs where the return value is never consumed. No impact.

3. **`session-manager.ts` `injectMessage` method**: The method does not validate that the session exists before calling `MessagePersistence.persist()`. However, `persist()` itself calls `sessionCache.getAsync(sessionId)` and throws if the session is not found, so the error path is already handled. This is fine — the error will surface naturally. Just noting for awareness.

### Conclusion

The implementation is clean, correct, and well-tested. All acceptance criteria are met. The wiring order is explicitly verified by tests. No P0, P1, or P2 issues found.

lsm (commented):

The implementation correctly wires `SessionNotificationSink` into `SpaceRuntimeService` using the setter pattern, respects the correct wiring order (session exists → sink created → setter called), and does not modify `SpaceRuntimeServiceConfig`. The 5 new tests pass. However, there is one meaningful correctness issue and a misleading test.

---

### P1 - Startup race condition causes notifications to be silently dropped permanently

**File:** `packages/daemon/src/lib/rpc-handlers/index.ts` lines 250, 306-318

`spaceRuntimeService.start()` is called at line 250, which fires the first `executeTick()` call immediately and asynchronously (not awaited). `provisionGlobalSpacesAgent()` is called fire-and-forget at line 306, so the first tick races against session creation.

The critical problem is inside `SpaceRuntime.processRunTick()` (`space-runtime.ts`, line 523):

```ts
this.notifiedTaskSet.add(dedupKey);   // key added BEFORE notify()
await this.notificationSink.notify(event);  // fires on NullNotificationSink
```

The dedup key is added to `notifiedTaskSet` before the notification is actually delivered. If the first tick fires before the sink is wired (which is likely in practice, since the first tick runs `rehydrateExecutors()` which reads from DB, while provisioning only needs a SQLite insert for the session row), any `needs_attention` tasks already in the database at daemon startup will:
1. Trigger a notification call on `NullNotificationSink` (no-op).
2. Get their dedup key added to `notifiedTaskSet`.
3. Never trigger a notification again on subsequent ticks — the dedup guard skips them permanently.

Result: Tasks in `needs_attention` status at restart silently stop reaching the Space Agent. The restart idempotency test (`wires sink even when session already exists`) does not catch this because it uses `tickIntervalMs: 100_000` and never calls `executeTick()` before `provisionGlobalSpacesAgent`.

**Suggested fix:** Either defer `spaceRuntimeService.start()` until after `provisionGlobalSpacesAgent` completes, or reset `notifiedTaskSet` in `setNotificationSink()` so pending tasks are re-evaluated on the next tick.

---

### P2 - Wiring order test has dead/misleading code

**File:** `packages/daemon/tests/unit/space/provision-global-agent.test.ts` lines 261-274

The initial `sessionManager` mock at lines 261-274 builds a `getSessionAsync` that returns `async () => { ... }` — an async function returning another async function, not `null | AgentSession`. This implementation is immediately replaced at lines 279-285. The dead code makes the test unnecessarily confusing to read and could mislead a future editor who tries to modify the order-tracking logic.

---

### P2 - Error path in provisionGlobalSpacesAgent is untested

**File:** `packages/daemon/src/lib/space/provision-global-agent.ts` lines 76-88

The `createSession` failure path (error caught and re-thrown) is not tested. While this failure is caught by the `.catch` in `rpc-handlers/index.ts` and logged (non-fatal), the behavior under failure — no sink wired, `spaceRuntimeService` continues with `NullNotificationSink` — is silently invisible. A test verifying the error propagates and that the sink is NOT wired on failure would add confidence.

---

### P3 - Inline import type

**File:** `packages/daemon/src/lib/rpc-handlers/index.ts` line 292

`MessageDeliveryMode` is used via an inline `import('@neokai/shared').MessageDeliveryMode` type annotation instead of a top-level import alongside the existing `import type { MessageHub } from '@neokai/shared'` at line 8. Minor style inconsistency.

lsm (commented):

**Recommendation: APPROVE**

The previous review feedback has been thoroughly addressed. All four issues (P1 startup race, P2 dead code, P2 error path test, P3 inline import) have been fixed correctly. Here is my assessment of the current state:

### What was done

The PR wires `SessionNotificationSink` into `provisionGlobalSpacesAgent` so that `SpaceRuntime` tick events flow into the `spaces:global` agent session. The key changes:

1. **`provision-global-agent.ts`** -- After the session is created/confirmed, a `SessionNotificationSink` is constructed and injected via `spaceRuntimeService.setNotificationSink()`. Placement is correct: after session existence is confirmed (line 91-93) but before MCP tools are attached.

2. **`session-manager.ts`** -- New `injectMessage()` method delegates to `MessagePersistence.persist()` with a generated UUID. Clean, minimal, and correctly reuses the existing persistence pipeline.

3. **`space-runtime.ts`** -- `setNotificationSink()` now clears `notifiedTaskSet` to handle the startup race where ticks fire on `NullNotificationSink` before provisioning completes. This is the correct fix.

4. **`rpc-handlers/index.ts`** -- Builds a minimal `SessionFactory` adapter with only `injectMessage` and `hasSession` wired to real implementations; other methods are stubs. The top-level import for `MessageDeliveryMode` replaces the previous inline import.

5. **Tests (453 lines)** -- 7 tests covering: sink wired, sink is non-null, wiring order, error propagation + sink NOT wired on failure, daemon restart path, end-to-end tick-to-inject flow, and the startup race regression test.

### Assessment

- **Correctness**: The wiring is in the right place, the startup race is handled, error propagation works correctly (tested).
- **Completeness**: All aspects of the task are addressed. The `SessionFactory` adapter pattern matches how it's used in other parts of the codebase (room runtime).
- **Tests**: Good coverage including the regression test for the startup race. The end-to-end test verifies the full path from `executeTick()` through to `sessionFactory.injectMessage()`.
- **Error handling**: `SessionNotificationSink.notify()` already catches and logs errors (does not propagate to the tick loop). `injectMessage` correctly delegates to `MessagePersistence.persist()` which throws on session-not-found -- but that throw is caught by the sink's try/catch. This is correct behavior.
- **Type safety**: `bun run typecheck` passes clean. All 7 tests pass.

### Minor observations (P3)

1. **Stub methods in `globalSessionFactory`** (`rpc-handlers/index.ts`, lines 296-303): The stub methods like `createAndStartSession: async () => {}` satisfy the `SessionFactory` interface but would silently no-op if accidentally called. This is acceptable since only `injectMessage` and `hasSession` are used by `SessionNotificationSink`, and the comment on line 295 documents this. A stricter alternative would be to throw "not implemented" in the stubs, but that is a minor style preference.

No blocking or should-fix issues found. The implementation is clean, well-tested, and correctly addresses all prior feedback.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been correctly resolved. The implementation is correct, complete, and well-tested.

---

### Previous issues — verification

**P1 (startup race) — Fixed correctly.**
`SpaceRuntime.setNotificationSink()` now calls `this.notifiedTaskSet.clear()` (space-runtime.ts line 156) immediately after replacing the sink. The startup race test in `provision-global-agent.test.ts` (lines 407-453) explicitly exercises the path: fires a tick on `NullNotificationSink` (dedup key added, no injection), then provisions (sink wired, dedup cleared), then fires a second tick and asserts `sessionFactory.calls.length >= 1`. This is a proper regression test.

**P2 (dead code in order test) — Fixed.**
The initial mock block that was immediately overwritten is gone. The order test now has a single clean `sessionManager` mock that tracks `createSession` calls in `callOrder`. No dead code.

**P2 (error path test) — Fixed.**
The test "throws and does NOT wire sink when session creation fails" (line 318) verifies both that the error propagates (`rejects.toThrow('DB write failed')`) and that `sinkCalls` remains empty. Passes correctly.

**P3 (inline import) — Fixed.**
`MessageDeliveryMode` is now imported at the top level alongside `MessageHub` (rpc-handlers/index.ts line 8).

---

### New code quality observations

**P3 — Redundant push in the order test.**
In `provision-global-agent.test.ts` lines 288-295, the monkey-patched `setNotificationSink` calls `originalSetSink(sink)` (which binds to `SpySpaceRuntimeService.setNotificationSink`, which already does `this.sinkCalls.push(sink)`), then also does an explicit `spyService.sinkCalls.push(sink)` on line 294. This results in two entries in `sinkCalls` for the order test. Since the order test only asserts on `callOrder`, not `sinkCalls`, this is harmless — but the extra push is confusing dead code. A future reader might wonder why `sinkCalls.length === 2` in a test that should wire the sink once.

**P3 — All-null `getSessionAsync` in error test is mildly fragile.**
In the error-path test (line 321), `getSessionAsync: mock(async () => null)` returns null on every call. If `createSession` somehow didn't throw, the code would fall through to `throw new Error(`Failed to get AgentSession...`)`. The test would still pass (wrong error message, but still rejects). Not a defect — just a note.

---

### Overall assessment

The implementation is correct and the full integration chain (startup race, error path, restart path, end-to-end tick-to-inject) is covered by tests. All 7 tests pass. Typecheck is clean. The two P3 nits above are cosmetic and do not affect correctness or maintainability in a meaningful way.

PR #532, opened 2026-03-21
feat: wire TaskAgentManager into DaemonApp (Task 4.3)

lsm:
- Add `cleanupAll()` to TaskAgentManager for graceful daemon shutdown
- Instantiate TaskAgentManager inside setupRPCHandlers (after SpaceRuntimeService) with all required dependencies
- Expose TaskAgentManager via RPCHandlerSetupResult
- Add `taskAgentManager: TaskAgentManager` to DaemonAppContext interface
- Wire TaskAgentManager into createDaemonApp and call cleanupAll() on shutdown
- Add `space.task.sendMessage` RPC handler (space-task-message-handlers.ts) that injects messages into Task Agent sessions via TaskAgentManager.injectTaskAgentMessage()
- Add 6 unit tests for space.task.sendMessage handler (happy path, validation errors, no-session error)
- Add 2 unit tests for cleanupAll() (multi-task cleanup, no-op when empty)

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -19,6 +19,7 @@ import { SpaceAgentRepository } from './storage/repositories/space-agent-reposit
 import { SpaceAgentManager } from './lib/space/managers/space-agent-manager';
 import { SpaceManager } from './lib/space/managers/space-manager';
 import type { SpaceRuntimeService } from './lib/space/runtime/space-runtime-service';
+import type { TaskAgentManager } from './lib/space/runtime/task-agent-manager';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -60,6 +61,8 @@ export interface DaemonAppContext {
 	spaceManager: SpaceManager;
 	/** Space runtime service for workflow run lifecycle management */
 	spaceRuntimeService: SpaceRuntimeService;
+	/** Task Agent Manager — manages Task Agent session lifecycle for space tasks */
+	taskAgentManager: TaskAgentManager;
 	/**
 	 * Cleanup function for graceful shutdown.
 	 * Closes all connections, stops sessions, and closes database.
@@ -210,7 +213,11 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	}
 
 	// Setup RPC handlers (returns cleanup function + exposed services)
-	const { cleanup: rpcHandlerCleanup, spaceRuntimeService } = setupRPCHandlers({
+	const {
+		cleanup: rpcHandlerCleanup,
+		spaceRuntimeService,
+		taskAgentManager,
+	} = setupRPCHandlers({
 		messageHub,
 		sessionManager,
 		authManager,
@@ -399,6 +406,10 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 				logInfo('[Daemon] GitHub service stopped');
 			}
 
+			// Stop all Task Agent sessions before sessionManager.cleanup() so that
+			// Task Agent sessions are interrupted cleanly before the session pool drains.
+			await taskAgentManager.cleanupAll();
+
 			// Stop all agent sessions first — this closes any open SSE connections
 			// that are held by providers (e.g. AnthropicToCopilotBridgeProvider's embedded
 			// HTTP server). Provider shutdown must follow so server.close() is not
@@ -439,6 +450,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		spaceAgentManager,
 		spaceManager,
 		spaceRuntimeService,
+		taskAgentManager,
 		cleanup,
 	};
 }
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -63,6 +63,8 @@ import { setupSpaceExportImportHandlers } from './space-export-import-handlers';
 import { provisionGlobalSpacesAgent } from '../space/provision-global-agent';
 import { setupGlobalSpacesHandlers } from './global-spaces-handlers';
 import type { GlobalSpacesState } from '../space/tools/global-spaces-tools';
+import { TaskAgentManager } from '../space/runtime/task-agent-manager';
+import { setupSpaceTaskSendMessageHandler } from './space-task-message-handlers';
 
 export interface RPCHandlerDependencies {
 	messageHub: MessageHub;
@@ -92,6 +94,7 @@ export type RPCHandlerCleanup = () => void;
 export interface RPCHandlerSetupResult {
 	cleanup: RPCHandlerCleanup;
 	spaceRuntimeService: SpaceRuntimeService;
+	taskAgentManager: TaskAgentManager;
 }
 
 /**
@@ -249,6 +252,25 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	});
 	spaceRuntimeService.start();
 
+	// Task Agent Manager — manages Task Agent session lifecycle for space tasks
+	const taskAgentManager = new TaskAgentManager({
+		db: deps.db,
+		sessionManager: deps.sessionManager,
+		spaceManager: deps.spaceManager,
+		spaceAgentManager: deps.spaceAgentManager,
+		spaceWorkflowManager,
+		spaceRuntimeService,
+		taskRepo: spaceTaskRepo,
+		workflowRunRepo: spaceWorkflowRunRepo,
+		daemonHub: deps.daemonHub,
+		messageHub: deps.messageHub,
+		getApiKey: () => deps.authManager.getCurrentApiKey(),
+		defaultModel: deps.config.defaultModel,
+	});
+
+	// space.task.sendMessage — inject a message into a Task Agent session
+	setupSpaceTaskSendMessageHandler(deps.messageHub, taskAgentManager);
+
 	// Space export/import handlers
 	setupSpaceExportImportHandlers(
 		deps.messageHub,
@@ -304,5 +326,6 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 			spaceRuntimeService.stop();
 		},
 		spaceRuntimeService,
+		taskAgentManager,
 	};
 }
--- a/packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts
@@ -26,10 +26,11 @@ export function setupSpaceTaskSendMessageHandler(
 			throw new Error('message is required');
 		}
 
-		if (!taskAgentManager.isTaskAgentAlive(params.taskId)) {
-			throw new Error(`No active Task Agent session for task: ${params.taskId}`);
-		}
-
+		// Delegate directly to injectTaskAgentMessage — it is the single authoritative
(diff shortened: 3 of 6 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly wires `TaskAgentManager` into the `DaemonApp` lifecycle. The implementation follows existing patterns in the codebase and is well-structured.

### What was done

1. **`cleanupAll()` added to TaskAgentManager** (`packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 433-437): Simple delegation to the existing per-task `cleanup()` via `Promise.allSettled`. Correct use of `allSettled` ensures one failing cleanup does not block others.

2. **TaskAgentManager instantiation in `setupRPCHandlers`** (`packages/daemon/src/lib/rpc-handlers/index.ts`, lines 256-269): Properly placed after `SpaceRuntimeService` creation (since it depends on it). All required dependencies are correctly threaded through. The `getApiKey` closure correctly delegates to `authManager.getCurrentApiKey()`.

3. **`RPCHandlerSetupResult` updated** to expose `taskAgentManager` alongside `spaceRuntimeService` (`packages/daemon/src/lib/rpc-handlers/index.ts`, line 97).

4. **`space.task.sendMessage` RPC handler** (`packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts`): Clean, focused handler with proper validation (taskId required, non-empty message, alive session check). Follows the existing handler pattern.

5. **`DaemonAppContext` updated** (`packages/daemon/src/app.ts`, line 65): Type-only addition with JSDoc comment.

6. **Shutdown ordering** (`packages/daemon/src/app.ts`, lines 409-411): `taskAgentManager.cleanupAll()` is called before `sessionManager.cleanup()`, which is correct -- Task Agent sessions should be interrupted before the general session pool drains.

7. **Tests** are thorough for the new code:
   - 6 tests for the RPC handler covering happy path, missing taskId, empty message, whitespace-only message, and no active session.
   - 2 tests for `cleanupAll` (multi-task cleanup and no-op when empty).

### Observations (P3 - nits, non-blocking)

1. **Shutdown ordering subtlety** (`packages/daemon/src/app.ts`, lines 395 vs 411): `messageHub.cleanup()` is called at line 395 before `taskAgentManager.cleanupAll()` at line 411. While this works today because `cleanupAll` does not use messageHub during teardown (it calls `session.handleInterrupt()` and `sessionManager.deleteSession()` directly, and errors are caught), the ordering is slightly counterintuitive. If future cleanup logic ever needs to publish events via messageHub, this would silently fail. Worth a comment or reordering in a future cleanup pass.

2. **`data as { taskId: string; message: string }` cast** (`packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts`, line 20): The handler casts the incoming `data` to the expected shape without runtime schema validation. This is consistent with every other RPC handler in the codebase, so it is not a problem unique to this PR.

Overall this is a clean, focused integration PR that correctly follows the established patterns.

lsm (commented):

> **Recommendation: REQUEST_CHANGES** — Two P1 issues, one P2.

---

### Summary

The wiring is correct and complete: `TaskAgentManager` is properly instantiated with all dependencies, exposed via `RPCHandlerSetupResult`, added to `DaemonAppContext`, and `cleanupAll()` is called in the right position in the shutdown sequence (before `sessionManager.cleanup()`). Tests pass and cover the primary scenarios. There are a few issues worth fixing before merge.

---

### P1 — TOCTOU race in `space.task.sendMessage` handler

**File:** `packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts` lines 29–33

The handler calls `isTaskAgentAlive(taskId)` and then `injectTaskAgentMessage(taskId, message)` as two separate operations. In the window between those two calls, a concurrent `cleanupAll()` (e.g., triggered by daemon shutdown) could remove the session from the map. When that happens `injectTaskAgentMessage` will throw `Task Agent session not found for task ` — a different error message than the one the client expects from the alive check.

Since `injectTaskAgentMessage` already performs its own null-guard and throws, the pre-check in the handler is redundant and doesn't prevent the race. The better fix is to remove the pre-check from the handler and let `injectTaskAgentMessage` be the single authoritative gate. If a friendlier error message is desired, the wording in `injectTaskAgentMessage` can be updated to match.

---

### P1 — `interrupted` state accepted by `isTaskAgentAlive` but not handled in injection path

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts` lines 394–409 and 697–703

`isTaskAgentAlive` returns `true` when the session is in the `interrupted` state, meaning `space.task.sendMessage` will pass validation and attempt to inject a message. `injectMessageIntoSession` does not treat `interrupted` as busy (`isBusy` only checks `processing | queued | waiting_for_input`), so it will call `ensureQueryStarted()` and enqueue the message unconditionally.

Whether sending to an interrupted session is intentional — and whether `ensureQueryStarted()` correctly restarts an interrupted query — is not clear from the code. There is no test covering injection into an `interrupted` session, and no documentation explaining the expected behavior. Either:
- document why `interrupted` is a valid injection target and add a test, or
- remove `interrupted` from `isTaskAgentAlive` (or add it to `isBusy`) until the recovery flow is explicit.

---

### P2 — No test for the `interrupted` session injection path

**File:** `packages/daemon/tests/unit/rpc-handlers/space-task-message-handlers.test.ts`

`isTaskAgentAlive` returns `true` for `interrupted`, so the handler allows the call through. The test suite mocks `isTaskAgentAlive` as always-true but never sets up a scenario where the underlying session is `interrupted` to verify end-to-end behavior. If the intent is to support injection into interrupted sessions, a test should confirm that the message is enqueued correctly (or rejected with a clear error).

---

### P3 — Missing null guard on `data` parameter

**File:** `packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts` line 20

`const params = data as { taskId: string; message: string }` will throw a confusing `TypeError: Cannot read properties of null` if the RPC payload is `null` or `undefined`. Other handlers in the codebase follow the same pattern so this is consistent, but a defensive `if (!data || typeof data !== 'object')` check would improve the error message for malformed calls. Low priority given it matches the existing convention.

---

- **Shutdown ordering is correct**: `cleanupAll()` is called before `sessionManager.cleanup()` in `app.ts` (lines 411–417), matching the comment intent.
- **`cleanupAll` uses `Promise.allSettled`** so one failing cleanup does not block others — good.
- **Tests are clean**: 6 handler unit tests and 2 `cleanupAll` tests are well-structured with proper mock boundaries.
- **No new unused imports or dead exports** introduced.

lsm (commented):

**Recommendation: APPROVE**

### Second Iteration — P1 Issues Verified as Fixed

Both P1 issues from the previous review have been correctly addressed.

---

#### P1-1: TOCTOU race (isTaskAgentAlive pre-check) — FIXED

The `space.task.sendMessage` handler in `packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts` now delegates directly to `injectTaskAgentMessage` with no intermediate liveness pre-check. The comment explicitly documents why the pre-check was removed (lines 29-33):

> A separate isTaskAgentAlive pre-check would introduce a TOCTOU race: a concurrent cleanupAll() ... could remove the session between the check and the injection call.

The test in `space-task-message-handlers.test.ts` (line 62, "does not call isTaskAgentAlive") installs a spy on `isTaskAgentAlive` and asserts it is never called, which is the right way to lock in this constraint.

---

#### P1-2: `interrupted` state not covered in injection path — FIXED

The `isBusy` guard in `injectMessageIntoSession` (lines 706-710 of `task-agent-manager.ts`) now explicitly includes `interrupted`:

```ts
const isBusy =
    state.status === 'processing' ||
    state.status === 'queued' ||
    state.status === 'waiting_for_input' ||
    state.status === 'interrupted';
```

The comment on lines 698-705 correctly distinguishes what `interrupted` means for each delivery mode: `next_turn` is deferred, but `current_turn` is allowed through (ensureQueryStarted restarts the query).

Two new tests cover this precisely:
- "saves with 'saved' status when session is interrupted (next_turn deferred)" — verifies deferral
- "enqueues immediately for current_turn when session is interrupted (restartable)" — verifies the current_turn path is not blocked

---

### Other Observations (no blocking issues)

- **Wiring is complete.** `setupRPCHandlers` in `index.ts` constructs `TaskAgentManager` with all required dependencies and calls `setupSpaceTaskSendMessageHandler`, returning both `spaceRuntimeService` and `taskAgentManager` to `DaemonAppContext`. The cleanup sequence in `app.ts` correctly calls `taskAgentManager.cleanupAll()` before `sessionManager.cleanup()` so Task Agent sessions are interrupted cleanly before the session pool drains.

- **`isTaskAgentAlive` on `TaskAgentManager` still lists `interrupted` (lines 402-408).** This is correct: the method is used by callers outside the injection path (e.g., status queries) where `interrupted` should still count as "alive". The `isBusy` gate inside `injectMessageIntoSession` is a separate, narrower check that only affects delivery deferral. No issue here.

- **Test helper accesses private method via cast.** `callInjectMessage` in the unit test uses `(manager as unknown as {...}).injectMessageIntoSession(...)` to bypass TypeScript visibility. This is an acceptable and common pattern for unit-testing private implementation details in TypeScript without converting the method to `protected`.

---

All P0, P1, and P2 issues from the previous review are resolved. The implementation is correct and the test coverage is appropriate.

lsm (commented):

**Recommendation: APPROVE**

The P1 issues from the previous review have been addressed cleanly:

### P1a fix (TOCTOU race): Correctly resolved

The `isTaskAgentAlive()` pre-check has been removed from the RPC handler (`space-task-message-handlers.ts:29-34`). The handler now delegates entirely to `injectTaskAgentMessage()`, which atomically checks session existence and injects the message. The comment explaining the rationale is accurate and helpful. A dedicated test (`does not call isTaskAgentAlive`) verifies the pre-check is gone.

### P1b fix (interrupted state): Correctly resolved

`interrupted` is now included in the `isBusy` condition at `task-agent-manager.ts:710`. The comment at lines 697-705 clearly explains the distinction: `next_turn` messages are deferred (saved) while `current_turn` messages proceed through `ensureQueryStarted()` which can restart the query. Two new tests cover both paths.

### cleanupAll and shutdown ordering

`cleanupAll()` at `task-agent-manager.ts:433-437` uses `Promise.allSettled` which is the right choice -- a single task cleanup failure won't prevent other tasks from being cleaned up. The shutdown ordering in `app.ts` places `cleanupAll()` after `rpcHandlerCleanup()` (which stops `spaceRuntimeService`) but before `sessionManager.cleanup()`. Since `cleanup(taskId)` only calls `stopAndDeleteSession` and clears in-memory maps (no MessageHub dependency), this ordering is fine.

### Test coverage

The handler tests are well-structured: happy path, validation edge cases (missing taskId, empty message, whitespace-only message), no-session error propagation, and the explicit TOCTOU non-regression test. The `task-agent-manager.test.ts` additions cover interrupted-session injection for both delivery modes and `cleanupAll` for both multi-task and empty cases.

### Minor observations (P3)

1. **`space-task-message-handlers.ts:35`**: The `log.info` after injection includes the `taskId` directly in a template literal. This is fine for internal logging but if taskId ever came from user input with unusual characters, it could produce noisy logs. Not a real concern given taskIds are system-generated UUIDs.

2. **`space-task-message-handlers.ts:20`**: The `data as { taskId: string; message: string }` cast has no runtime type narrowing beyond the `!params.taskId` / `!params.message` checks. If `data` is `null` or a non-object, accessing `.taskId` would return `undefined` (caught by the guard) but `.message` on `null` would throw a different error than expected. In practice, MessageHub likely always passes an object, so this is a theoretical edge case.

No blocking issues found. The fixes directly address the previous reviewer's concerns with clean, well-tested implementations.

PR #533, opened 2026-03-21
feat: implement goal verification loop for Space Agent

lsm:
- Add goalId field to SpaceTask types and DB schema (migration 34)
- Add listByGoalId() to SpaceTaskRepository
- Add GoalTasksCompleteEvent to SpaceNotificationEvent union
- Add goal completion detection to SpaceRuntime tick loop with iteration tracking
- Add complete_goal tool to space-agent-tools
- Extend create_standalone_task with goal_id parameter
- Update Space Agent system prompt with goal_tasks_complete event handling
- Add tests for session-notification-sink, space-agent-tools, and space-runtime

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -56,6 +56,7 @@ import { setupSpaceAgentHandlers } from './space-agent-handlers';
 import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
 import { SpaceWorkflowRepository } from '../../storage/repositories/space-workflow-repository';
 import { SpaceAgentRepository } from '../../storage/repositories/space-agent-repository';
+import { GoalRepository } from '../../storage/repositories/goal-repository';
 import { SpaceRuntimeService } from '../space/runtime/space-runtime-service';
 import { setupSpaceWorkflowRunHandlers } from './space-workflow-run-handlers';
 import type { SpaceWorkflowRunTaskManagerFactory } from './space-workflow-run-handlers';
@@ -194,6 +195,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	// Space handlers (spaceManager injected from deps — single instance shared with DaemonAppContext)
 	const spaceTaskRepo = new SpaceTaskRepository(deps.db.getDatabase());
 	const spaceWorkflowRunRepo = new SpaceWorkflowRunRepository(deps.db.getDatabase());
+	const goalRepo = new GoalRepository(deps.db.getDatabase());
 
 	// Space workflow manager — created early so space.create can call seedBuiltInWorkflows
 	const spaceWorkflowRepo = new SpaceWorkflowRepository(deps.db.getDatabase());
@@ -246,6 +248,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		spaceWorkflowManager,
 		workflowRunRepo: spaceWorkflowRunRepo,
 		taskRepo: spaceTaskRepo,
+		goalRepo,
 	});
 	spaceRuntimeService.start();
 
@@ -313,6 +316,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 			taskRepo: spaceTaskRepo,
 			workflowRunRepo: spaceWorkflowRunRepo,
 			db: deps.db.getDatabase(),
+			goalRepo,
 			state: globalSpacesState,
 		}).catch((error) => {
 			log.error('Failed to provision global spaces agent:', error);
--- a/packages/daemon/src/lib/space/agents/space-chat-agent.ts
+++ b/packages/daemon/src/lib/space/agents/space-chat-agent.ts
@@ -206,6 +206,17 @@ export function buildSpaceChatSystemPrompt(context: SpaceChatAgentContext = {}):
 			`  Payload fields: \`runId\`, \`reason\`, \`autonomyLevel\`\n` +
 			`  Action: Summarize the outcome to the user and suggest next steps if relevant.`
 	);
+	sections.push('');
+	sections.push(
+		`- **\`goal_tasks_complete\`** — All tasks for a goal have completed. Verification is required.\n` +
+			`  Payload fields: \`goalId\`, \`goalTitle\`, \`goalValidationCriteria\`, \`iterationCount\`, \`previousIssueCount\`\n` +
+			`  Action (verification loop):\n` +
+			`    1. Create a verification task using \`create_standalone_task\` with \`goal_id\` set and the \`goalValidationCriteria\` as the task description.\n` +
+			`    2. When the verification task completes, read its result with \`get_task_detail\`.\n` +
+			`    3. **If issues found**: Create new fix tasks using \`create_standalone_task\` with \`goal_id\` set describing each issue. The loop will trigger again when they complete.\n` +
+			`    4. **If no issues**: Call \`complete_goal\` with a summary of what was accomplished.\n` +
+			`    5. **Iteration limit**: If \`iterationCount\` exceeds 3, escalate to the human instead of creating more tasks.`
+	);
 
 	// Autonomy level section
 	const level = context.autonomyLevel ?? 'supervised';
@@ -255,7 +266,13 @@ export function buildSpaceChatSystemPrompt(context: SpaceChatAgentContext = {}):
 	sections.push('');
 	sections.push(
 		`- **\`create_standalone_task\`** — Create a task outside any workflow. Use for self-contained work ` +
-			`that doesn't require multi-step orchestration. Provide a title, description, and optionally an agent ID.`
+			`that doesn't require multi-step orchestration. Provide a title, description, and optionally an agent ID. ` +
+			`Pass \`goal_id\` to link the task to a goal — SpaceRuntime tracks goal-level completion automatically.`
+	);
+	sections.push('');
+	sections.push(
+		`- **\`complete_goal\`** — Mark a goal as successfully completed after a verification task confirms ` +
+			`the work meets the goal's validation criteria. Provide a summary of what was accomplished.`
 	);
 	sections.push('');
 	sections.push(
--- a/packages/daemon/src/lib/space/provision-global-agent.ts
+++ b/packages/daemon/src/lib/space/provision-global-agent.ts
@@ -16,6 +16,7 @@ import type { SpaceAgentManager } from './managers/space-agent-manager';
 import type { SpaceWorkflowManager } from './managers/space-workflow-manager';
 import type { SpaceTaskRepository } from '../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
+import type { GoalRepository } from '../../storage/repositories/goal-repository';
 import type { SpaceRuntimeService } from './runtime/space-runtime-service';
 import { Logger } from '../logger';
 import { buildGlobalSpacesAgentPrompt } from './agents/global-spaces-agent';
@@ -42,6 +43,8 @@ export interface ProvisionGlobalSpacesAgentDeps {
 	workflowRunRepo: SpaceWorkflowRunRepository;
 	/** Database instance passed through to GlobalSpacesToolsConfig for SpaceTaskManager creation. */
 	db: BunDatabase;
+	/** Goal repository for the complete_goal tool and goal-completion detection. */
+	goalRepo?: GoalRepository;
 	/** Shared mutable state for the active space context. Created externally so RPC handlers can use the same reference. */
 	state: GlobalSpacesState;
 }
@@ -66,6 +69,7 @@ export async function provisionGlobalSpacesAgent(
 		taskRepo,
 		workflowRunRepo,
 		db,
+		goalRepo,
 		state,
 	} = deps;
 
@@ -116,6 +120,7 @@ export async function provisionGlobalSpacesAgent(
 			taskRepo,
 			workflowRunRepo,
 			db,
+			goalRepo,
 		},
 		state
 	);
--- a/packages/daemon/src/lib/space/runtime/notification-sink.ts
+++ b/packages/daemon/src/lib/space/runtime/notification-sink.ts
@@ -52,6 +52,29 @@ export interface TaskTimeoutEvent {
 	timestamp: string;
 }
 
+/**
+ * All non-cancelled tasks for a goal have completed. Space Agent should create a
+ * verification task to validate the work, then either mark the goal complete or
+ * create fix tasks for any issues found (triggering another iteration).
+ */
+export interface GoalTasksCompleteEvent {
+	kind: 'goal_tasks_complete';
+	/** Space the tasks belong to. */
+	spaceId: string;
+	/** ID of the goal whose tasks have all completed. */
+	goalId: string;
+	/** Title of the goal (for human-readable context in the message). */
+	goalTitle: string;
+	/** Validation criteria from the goal description (used to build verification task). */
+	goalValidationCriteria?: string;
+	/** How many verification iterations have completed (starts at 1 on first completion). */
+	iterationCount: number;
+	/** Number of issues found in the previous iteration (undefined on first iteration). */
+	previousIssueCount?: number;
+	/** ISO-8601 timestamp when the event was emitted. */
+	timestamp: string;
+}
+
 /** A workflow run has reached a terminal state (completed, cancelled, or needs_attention). */
 export interface WorkflowRunCompletedEvent {
 	kind: 'workflow_run_completed';
@@ -99,7 +122,8 @@ export type SpaceNotificationEvent =
 	| TaskNeedsAttentionEvent
 	| WorkflowRunNeedsAttentionEvent
 	| TaskTimeoutEvent
-	| WorkflowRunCompletedEvent;
+	| WorkflowRunCompletedEvent
+	| GoalTasksCompleteEvent;
 
 // ---------------------------------------------------------------------------
 // NotificationSink interface
--- a/packages/daemon/src/lib/space/runtime/session-notification-sink.ts
+++ b/packages/daemon/src/lib/space/runtime/session-notification-sink.ts
@@ -108,6 +108,8 @@ export function formatEventMessage(
 			return formatTaskTimeout(event, autonomyLevel);
 		case 'workflow_run_completed':
 			return formatWorkflowRunCompleted(event, autonomyLevel);
+		case 'goal_tasks_complete':
+			return formatGoalTasksComplete(event, autonomyLevel);
 	}
 }
 
@@ -207,6 +209,53 @@ function formatWorkflowRunCompleted(
 	};
 	if (event.summary !== undefined) {
 		payload['summary'] = event.summary;
+	}
+	return buildMessage(event.kind, humanReadable, payload);
+}
+
+function formatGoalTasksComplete(
+	event: {
+		kind: 'goal_tasks_complete';
+		spaceId: string;
+		goalId: string;
+		goalTitle: string;
+		goalValidationCriteria?: string;
+		iterationCount: number;
+		previousIssueCount?: number;
+		timestamp: string;
+	},
+	autonomyLevel: AutonomyLevel
+): string {
+	const iterationPart =
+		event.iterationCount === 1
+			? 'First completion'
+			: event.previousIssueCount !== undefined
+				? `Iteration ${event.iterationCount} (previous iteration found ${event.previousIssueCount} issue(s))`
+				: `Iteration ${event.iterationCount}`;
+
+	const criteriaPart = event.goalValidationCriteria
+		? ` Validation criteria: ${event.goalValidationCriteria}`
+		: '';
+
+	const humanReadable =
+		`[GOAL_TASKS_COMPLETE] All tasks for goal "${event.goalTitle}" (id: ${event.goalId}) are completed. ` +
+		`${iterationPart}.${criteriaPart} ` +
+		`Action: Create a verification task to validate the completed work, then mark the goal complete or create fix tasks.`;
+
+	const payload: Record<string, unknown> = {
+		kind: event.kind,
+		spaceId: event.spaceId,
+		goalId: event.goalId,
+		goalTitle: event.goalTitle,
+		iterationCount: event.iterationCount,
+		timestamp: event.timestamp,
+		autonomyLevel,
+	};
+	if (event.goalValidationCriteria !== undefined) {
+		payload['goalValidationCriteria'] = event.goalValidationCriteria;
+	}
+	if (event.previousIssueCount !== undefined) {
+		payload['previousIssueCount'] = event.previousIssueCount;
 	}
 	return buildMessage(event.kind, humanReadable, payload);
 }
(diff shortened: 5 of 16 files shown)


lsm (commented):

Overall this is a well-structured implementation of the goal verification loop. The types, migration, repository changes, tool handlers, notification formatting, and tests are all thoughtfully done. However, there is one bug that will cause goal completion notifications to be silently dropped in production.

---

### P0: `checkGoalCompletion()` uses `this.config.notificationSink` instead of `this.notificationSink`

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, line 501

```ts
const sink = this.config.notificationSink ?? new NullNotificationSink();
```

This reads the notification sink from the **original config** passed at construction time, not the instance field `this.notificationSink` that can be replaced at runtime via `setNotificationSink()`.

In production, `SpaceRuntime` is constructed **before** the Space Agent session is provisioned, so `config.notificationSink` is typically `undefined` (or a `NullNotificationSink`). The real `SessionNotificationSink` is wired later via `setNotificationSink()`, which updates `this.notificationSink` but has no effect on `this.config.notificationSink`.

Every other notification call in this class correctly uses `this.notificationSink` (e.g., lines 678, 702, 735, 764, 845, 864). Only `checkGoalCompletion()` diverges.

**Impact:** In production, all `goal_tasks_complete` and goal-escalation notifications will be silently swallowed by a `NullNotificationSink`, making the entire goal verification loop non-functional.

**Fix:** Change line 501 to:
```ts
const sink = this.notificationSink;
```

The constructor already initializes `this.notificationSink` with a fallback (`config.notificationSink ?? new NullNotificationSink()`), so the `?? new NullNotificationSink()` guard is also redundant.

---

### P2: `previousIssueCount` is never populated

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 569-577

The `GoalTasksCompleteEvent` type defines a `previousIssueCount` field, and the `SessionNotificationSink` formatter renders it when present (`"previous iteration found N issue(s)"`), but `checkGoalCompletion()` never sets this field on the emitted event. It will always be `undefined`.

This isn't a bug per se (the field is optional), but it means the iteration context described in the system prompt ("previousIssueCount") will never carry useful data. If this is intentionally deferred for a future PR, consider adding a code comment to that effect so it's not forgotten. Otherwise, the field should either be populated (perhaps by tracking a count in `goalIterations`) or removed from the type/formatter to avoid confusion.

---

### P3: Test file synced with `runMigration34` comment but not verified

**File:** `packages/daemon/tests/unit/helpers/space-test-db.ts`, line 7

The comment says "Keep in sync with runMigration34" which is accurate. The `goal_id TEXT` column is present in the test helper's `space_tasks` table (line 148), so this is correct. Just noting the comment was updated appropriately.

---

### P3: Minor inconsistency in escalation event payload

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 546-554

When escalating after max iterations, the code emits a `task_needs_attention` event with `taskId: goalId`. This reuses the goal ID as a task ID, which could be confusing for the Space Agent since `taskId` normally refers to a `SpaceTask.id`, not a `RoomGoal.id`. The agent prompt instructions say "If iterationCount exceeds 3, escalate to the human" which aligns with the intent, but the mismatch in ID semantics could cause issues if the agent tries to call `get_task_detail` with the goal ID.

---

### Summary

The implementation is solid in design and test coverage. The critical P0 bug on line 501 will cause goal notifications to be silently dropped in production because the method reads the config's sink (which is typically null at construction time) rather than the runtime-replaceable instance field. This must be fixed before merge.

lsm (commented):

---

### Summary

The implementation is well-structured in isolation — the DB migration, repository extension, notification event, and iteration-tracking logic are all sound. However, there is a critical wiring gap: the new tools (`complete_goal`, `goal_id` in `create_standalone_task`) live in `space-agent-tools.ts` (backed by `createSpaceAgentMcpServer`) but the actual `spaces:global` session is served by `createGlobalSpacesMcpServer` from `global-spaces-tools.ts`, which does not include these tools. In production, the Space Agent cannot call `complete_goal` or link tasks to goals, making the entire feature dead code at runtime.

---

### P0 — Blocking

**`complete_goal` and `goal_id` are never wired into the real `spaces:global` session.**

The new tools added to `createSpaceAgentMcpServer` in `packages/daemon/src/lib/space/tools/space-agent-tools.ts` are never installed on the running session. The provisioning path in `packages/daemon/src/lib/space/provision-global-agent.ts` calls `createGlobalSpacesMcpServer` (from `global-spaces-tools.ts`), not `createSpaceAgentMcpServer`. Inspection of `global-spaces-tools.ts` confirms:

- The `create_standalone_task` tool there does NOT have a `goal_id` parameter (lines 320–348 of `global-spaces-tools.ts`).
- There is no `complete_goal` tool anywhere in `global-spaces-tools.ts`.

Additionally, `SpaceRuntimeService` (the factory that constructs `SpaceRuntime`) does not accept or pass through `goalRepo`, so even the detection side (emitting `goal_tasks_complete` with a meaningful title and validation criteria) gets no goal metadata at runtime — it would fall back to using the goal ID as the title.

Relevant files:
- `packages/daemon/src/lib/space/provision-global-agent.ts` lines 110–121 — creates `createGlobalSpacesMcpServer`, not `createSpaceAgentMcpServer`
- `packages/daemon/src/lib/space/runtime/space-runtime-service.ts` lines 24–40 — `SpaceRuntimeServiceConfig` has no `goalRepo` field
- `packages/daemon/src/lib/rpc-handlers/index.ts` lines 242–249 — `SpaceRuntimeService` constructed without `goalRepo`
- `packages/daemon/src/lib/space/tools/global-spaces-tools.ts` lines 320–348, 640–670 — no `goal_id` param, no `complete_goal` tool

**`setNotificationSink` does not clear `goalIterations`.**

When `setNotificationSink` is called (after provisioning completes), `notifiedTaskSet` is cleared so that pre-provisioning notifications can be re-emitted. But `goalIterations` is NOT cleared. If any goal-tasks-complete event fired on the `NullNotificationSink` before provisioning (i.e., if goals completed during daemon startup before the sink was wired), `notified: true` is already set and the Space Agent never learns about those completions. This is the mirror of the bug that `notifiedTaskSet.clear()` already correctly solves for task-level events.

- `packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 176–182 — `goalIterations` not cleared on sink swap

---

### P1 — Should Fix

**Escalation event uses a goal ID in the `taskId` field of `TaskNeedsAttentionEvent`.**

When max iterations is exceeded, the escalation fires:
```ts
await sink.notify({
  kind: 'task_needs_attention',
  spaceId: tasks[0].spaceId,
  taskId: goalId,   // ← this is a goal UUID, not a task UUID
  ...
});
```

The `TaskNeedsAttentionEvent.taskId` field is defined as "Task that needs attention" — reusing it with a goal ID is misleading and may confuse the Space Agent's handling logic or any downstream deduplication logic that keys on `taskId`. A dedicated `GoalEscalationEvent` kind (or at minimum a `GoalTasksCompleteEvent` with `iterationCount > maxGoalIterations`) would be semantically cleaner and consistent with the rest of the notification taxonomy.

- `packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 546–554

**No validation that the `goal_id` actually exists before creating a task linked to it.**

`create_standalone_task` in `space-agent-tools.ts` (line 324–332) passes `goalId` directly to `taskManager.createTask` with no check that the goal exists. If the agent passes an invalid or stale goal ID, the task is created with a dangling reference that `checkGoalCompletion` will pick up, query `goalRepo.getGoal(goalId)`, get null, and fall back to displaying the raw UUID as the goal title — not a crash, but a confusing silent failure. Consider validating the goal ID when `goalRepo` is available.

- `packages/daemon/src/lib/space/tools/space-agent-tools.ts` lines 311–337

**`previousIssueCount` is always omitted from notifications (acknowledged but still a gap).**

The code comment at line 576–578 of `space-runtime.ts` explicitly defers populating `previousIssueCount` in the notification payload. This means the system prompt instruction (step 3 of the verification loop: "If issues found: Create new fix tasks… The loop will trigger again when they complete") gives the agent no numerical signal about how many issues were found. The `GoalTasksCompleteEvent` interface has the field, the message formatter handles it, the tests assert it — but the emitter never populates it. The feature works without it, but the UX is degraded on subsequent iterations.

- `packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 576–578

---

### P2 — Moderate Concerns

**No test for `goalValidationCriteria` extraction from the goal description.**

The tests for `checkGoalCompletion` in `space-runtime.test.ts` verify event emission but do not wire a `goalRepo` into the runtime. As a result, there are no tests that verify the goal title and validation criteria are correctly pulled from the goal record and included in the event payload. The existing tests all use the fallback path (raw goal ID as title).

**Escalation dedup gap: once escalated, subsequent ticks also escalate.**

After the iteration count exceeds `maxGoalIterations`, the code sets `notified: true` and continues. On the next tick (assuming no new active tasks), `notified` is still `true`, so the loop skips (`if (state.notified) continue` at line 536). This correctly prevents duplicate escalation notifications. However, because `notified` is true, if the agent creates new fix tasks after the escalation (attempting to continue anyway), the `notified` flag will be cleared (line 524–526) and on the next full completion the count will be N+1 — still `> maxGoalIterations` — so another escalation fires. This is the intended behavior per the spec but worth explicitly testing.

---

### P3 — Nits

**System prompt mentions iteration limit as hardcoded 3** (line 18 of the system prompt diff) but the actual limit is configurable via `maxGoalIterations`. If someone sets `maxGoalIterations: 5` the prompt still says "If `iterationCount` exceeds 3". The prompt should reference the value dynamically or be updated to match whatever the actual limit is.

- `packages/daemon/src/lib/space/agents/space-chat-agent.ts` line 18 of the diff

**Redundant `NullNotificationSink` construction in `checkGoalCompletion`.**

Line 501 in the final implementation correctly uses `this.notificationSink` (the instance field), which is correct. An earlier version of the diff showed `const sink = this.config.notificationSink ?? new NullNotificationSink()` — verify the final file doesn't have this redundancy (it does not — current code is fine).

lsm (commented):

### Re-review of PR #533 — Goal Verification Loop

This is a second review pass. I verified the three specific issues flagged in prior reviews.

---

### Verification Results

**1. P0 FIX VERIFIED: `this.notificationSink` vs `this.config.notificationSink`**

The fix in commit `261bdff8` is correct. In `packages/daemon/src/lib/space/runtime/space-runtime.ts` line 501, `checkGoalCompletion()` now uses:
```ts
const sink = this.notificationSink;
```
This correctly references the instance field that gets updated by `setNotificationSink()`, rather than the immutable config property. **This P0 is resolved.**

---

### Remaining Issues

**P0: `complete_goal` tool and `goal_id` parameter are NOT wired into `global-spaces-tools.ts` — the feature is dead code at runtime**

The `spaces:global` session (the actual production Space Agent) uses `createGlobalSpacesMcpServer()` from `packages/daemon/src/lib/space/tools/global-spaces-tools.ts` (confirmed at `packages/daemon/src/lib/space/provision-global-agent.ts` line 22 and 110). This file does NOT contain:

- A `complete_goal` tool (neither handler nor MCP tool definition)
- A `goal_id` parameter on `create_standalone_task` (compare line 320-348 of `global-spaces-tools.ts` vs line 303-338 of `space-agent-tools.ts`)

Both `complete_goal` and `goal_id` only exist in `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, which is used by the per-space `SpaceChatAgent` — NOT the global agent that receives `goal_tasks_complete` notifications from `SpaceRuntime`.

This means:
1. When `SpaceRuntime.checkGoalCompletion()` fires a `goal_tasks_complete` event, the notification reaches the global agent session, but that agent has no `complete_goal` tool to act on it.
2. When the global agent tries to create a task with a `goal_id`, the `create_standalone_task` tool in `global-spaces-tools.ts` does not accept that parameter, so the task will not be linked to the goal.

The entire goal verification loop is effectively non-functional at runtime. This was reported in the previous Sonnet review and has NOT been addressed.

**P1: `setNotificationSink()` does not clear `goalIterations` — pre-provisioning goal events are permanently lost**

In `packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 176-182, `setNotificationSink()` clears `this.notifiedTaskSet` (so task events that fired before the real sink was wired get re-notified) but does NOT clear or reset `this.goalIterations`. 

If a goal's tasks all complete during the window between `SpaceRuntime` construction and `setNotificationSink()` being called (i.e., before the global agent session is provisioned), the goal entry in `goalIterations` will have `notified: true` with `iterationCount: 1`. When the real sink is wired, there is no mechanism to re-fire that notification. The goal completion event is permanently lost unless new tasks are created and then completed again for that goal.

This was reported in the previous Sonnet review and has NOT been addressed.

**P2: `previousIssueCount` field defined but never populated**

At line 576-578 of `space-runtime.ts`, there is a comment saying this is "intentionally omitted" and "deferred." The `GoalTasksCompleteEvent` type defines `previousIssueCount?: number` and the `formatGoalTasksComplete` formatter handles it, but the actual emission always omits it. A follow-up task or TODO tracking issue would be appropriate. This is minor since the field is optional and the formatter handles the undefined case gracefully. Worker added a comment in the previous round, which is acceptable for a P2.

---

### Summary

| Issue | Severity | Status |
|-------|----------|--------|
| `this.notificationSink` fix | P0 | **RESOLVED** |
| `complete_goal` + `goal_id` missing from `global-spaces-tools.ts` | P0 | **STILL OPEN** |
| `setNotificationSink()` not clearing `goalIterations` | P1 | **STILL OPEN** |
| `previousIssueCount` never populated | P2 | Acknowledged with comment, acceptable |

**Recommendation: REQUEST_CHANGES** — The P0 (`complete_goal`/`goal_id` not in global-spaces-tools) makes the entire goal verification loop non-functional at runtime. The P1 (`goalIterations` not cleared in `setNotificationSink`) creates a race condition window where goal completions can be silently lost.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two P0 issues remain open from the previous review. The worker addressed the notification sink bug (using `this.notificationSink`) and added wiring tests, but the core architectural gap that makes the entire feature non-functional was not addressed.

---

## Status of Previously Reported Issues

### P0 — complete_goal and goal_id missing from global-spaces-tools.ts (STILL OPEN)

**Confirmed unresolved.**

The goal verification loop works as follows:
1. `SpaceRuntime.checkGoalCompletion()` detects all tasks for a goal are complete
2. It emits a `goal_tasks_complete` notification to `SessionNotificationSink`
3. `SessionNotificationSink` injects a message into the `spaces:global` session
4. The global agent must call `complete_goal` to close the loop

The problem: the `spaces:global` session is provisioned with `createGlobalSpacesMcpServer` (in `global-spaces-tools.ts`), which contains no `complete_goal` tool and no `goal_id` parameter on `create_standalone_task`. These are only in `createSpaceAgentMcpServer` (in `space-agent-tools.ts`).

Evidence:
- `provision-global-agent.ts` line 110: the global agent session gets `createGlobalSpacesMcpServer()`
- `global-spaces-tools.ts`: no `complete_goal` tool, no `goal_id` on any tool
- `space-agent-tools.ts`: defines `complete_goal` and `goal_id` on `create_standalone_task`
- `createSpaceAgentMcpServer` is never called anywhere in production code — it exists only as an export and in tests

When the global agent receives the `goal_tasks_complete` notification and tries to close the loop, it has no tool to call. The feature is structurally dead.

### P0 — goalRepo never wired into production SpaceRuntime (NEW FINDING — also still open)

`SpaceRuntimeServiceConfig` (in `space-runtime-service.ts`) has no `goalRepo` field. `SpaceRuntime` accepts `goalRepo?: GoalRepository` as optional in its config, but because `SpaceRuntimeService` passes its own `config` directly to `new SpaceRuntime(config)`, the `goalRepo` is always `undefined` in production.

Impact:
- `checkGoalCompletion()` will never look up the goal title or validation criteria — the notification will have `goalTitle = goalId` and `goalValidationCriteria = undefined`
- More importantly, without `goalRepo` in the MCP tools, the agent cannot read goal details even if it had the `complete_goal` tool

Evidence in `rpc-handlers/index.ts` line 242-249:
```ts
const spaceRuntimeService = new SpaceRuntimeService({
  db: deps.db.getDatabase(),
  spaceManager: deps.spaceManager,
  spaceAgentManager: deps.spaceAgentManager,
  spaceWorkflowManager,
  workflowRunRepo: spaceWorkflowRunRepo,
  taskRepo: spaceTaskRepo,
  // goalRepo is NOT passed — does not exist in SpaceRuntimeServiceConfig
});
```

### P1 — setNotificationSink() clears notifiedTaskSet but not goalIterations (STILL OPEN)

The implementation at `space-runtime.ts` line 176-182 only clears `notifiedTaskSet`:
```ts
setNotificationSink(sink: NotificationSink): void {
  this.notificationSink = sink;
  this.notifiedTaskSet.clear();  // cleared
  // goalIterations NOT cleared
}
```

**Scenario where this causes a silent miss:**
1. Daemon starts, `SpaceRuntime` starts with `NullNotificationSink`
2. First tick fires before provisioning completes; `checkGoalCompletion()` finds all goal tasks completed, sets `goalIterations.set(goalId, { iterationCount: 1, notified: true })`
3. `setNotificationSink(realSink)` is called — clears `notifiedTaskSet` but leaves `goalIterations` intact with `notified: true`
4. All subsequent ticks hit the `if (state.notified) continue;` guard and the real sink never receives the `goal_tasks_complete` event

The existing test at `space-runtime.test.ts` line 1358 does NOT catch this because it sets up a fresh runtime and replaces the sink before any tick occurs — there are no pre-existing `goalIterations` entries with `notified: true`.

**Fix needed:** add `this.goalIterations.clear();` inside `setNotificationSink()`.

### P0 (previous) — this.notificationSink bug: RESOLVED

The worker fixed this in the latest commit. `checkGoalCompletion()` now correctly uses `this.notificationSink` instead of `this.config.notificationSink`.

### setNotificationSink wiring test: RESOLVED

A comprehensive test was added in `provision-global-agent.test.ts` covering: sink called after session creation, SessionNotificationSink wired (not null), ordering (createSession before setNotificationSink), failure path (sink NOT wired if session creation fails), and daemon restart path.

---

## Summary

The goal verification loop feature is structurally non-functional:
- The global agent that receives `goal_tasks_complete` notifications has no `complete_goal` tool and no `goal_id` parameter on task creation
- `goalRepo` is never injected into production `SpaceRuntime`, so goal metadata is also unavailable

Additionally, the startup-race fix for `goalIterations` (parallel to the `notifiedTaskSet` fix that was applied) was not implemented.

All three issues require code changes before this PR can be merged.

lsm (commented):

**Recommendation: APPROVE** — All four previously reported issues are now fixed and verified.

---

### Verification of Previously Reported Issues

**P0 — complete_goal tool and goal_id param in global-spaces-tools.ts: FIXED**

Both are now present:
- Handler `complete_goal` implemented at lines 477–512 in `packages/daemon/src/lib/space/tools/global-spaces-tools.ts`
- MCP tool registration at lines 790–798 in the same file
- `goal_id` optional param added to `create_standalone_task` handler and zod schema (lines 332–348, 718–723)

**P0 — goalRepo not in SpaceRuntimeServiceConfig: FIXED**

`goalRepo?: GoalRepository` is now present in `SpaceRuntimeServiceConfig` (line 33 of `space-runtime-service.ts`). The full wiring chain is verified:
- `rpc-handlers/index.ts` line 198: `GoalRepository` instantiated
- Line 251: passed into `SpaceRuntimeService` constructor
- Line 319: passed into `provisionGlobalSpacesAgent`
- `provision-global-agent.ts` line 123: forwarded to `createGlobalSpacesMcpServer`
- `SpaceRuntime` config also includes `goalRepo` via `SpaceRuntimeServiceConfig` passthrough

**P0 — createGlobalSpacesMcpServer missing complete_goal: FIXED**

`complete_goal` is now the final entry in the `tools` array passed to `createSdkMcpServer` (lines 790–798).

**P1 — setNotificationSink not clearing goalIterations: FIXED**

`setNotificationSink` at line 176–183 in `space-runtime.ts` now clears both `notifiedTaskSet` and `goalIterations`. Additionally, `checkGoalCompletion` at line 502 uses `this.notificationSink` (the instance field), not the original config sink — this was the actual root of the silent-drop bug.

---

### New Tests Added

The two fix commits add regression tests in `packages/daemon/tests/unit/space/space-runtime.test.ts` that directly verify:
- The replaced sink (via `setNotificationSink`) receives `goal_tasks_complete` events (not the original config sink)
- `goalIterations` is cleared on sink swap so pre-provisioning completions re-fire on the real sink
- `goalTitle` and `goalValidationCriteria` are extracted from `goalRepo` and included in emitted events

`complete_goal` handler logic is covered by `packages/daemon/tests/unit/space/space-agent-tools.test.ts` (the handler implementation is identical between the two tools files).

---

### Minor Observation (P3, no action required)

The `global-spaces-tools.test.ts` file does not include a test for `complete_goal` or the new `goal_id` param in `create_standalone_task`. Given the handler code is shared and covered in `space-agent-tools.test.ts`, this is acceptable as-is, though a future pass could add integration coverage directly to the global-spaces-tools test file for completeness.

lsm (commented):

All previously reported issues have been correctly resolved. Here is the verification for each item:

### 1. P0 (production wiring) — RESOLVED

The `complete_goal` tool is correctly defined in `packages/daemon/src/lib/space/tools/global-spaces-tools.ts`:
- Handler at line 477-512 with proper error handling (missing goalRepo, goal not found, already completed, archived)
- MCP tool definition at line 791-798 with zod schema for `goal_id` and `summary` parameters
- `goal_id` parameter added to `create_standalone_task` at line 333 and MCP schema at line 716-722

### 2. P0 (goalRepo wiring) — RESOLVED

The `GoalRepository` is correctly threaded through the full dependency chain:
- `GlobalSpacesToolsConfig.goalRepo?: GoalRepository` (line 50)
- `SpaceRuntimeServiceConfig.goalRepo?: GoalRepository` (line 33 of space-runtime-service.ts)
- `ProvisionGlobalSpacesAgentDeps.goalRepo?: GoalRepository` (line 47 of provision-global-agent.ts)
- `rpc-handlers/index.ts` creates `goalRepo` at line 198 and passes it to both `SpaceRuntimeService` (line 251) and `provisionGlobalSpacesAgent` (line 319)
- `provisionGlobalSpacesAgent` passes `goalRepo` to `createGlobalSpacesMcpServer` at line 123

### 3. P1 (goalIterations race) — RESOLVED

`setNotificationSink()` at line 176-183 of `space-runtime.ts` now clears both `this.notifiedTaskSet` and `this.goalIterations`. This ensures that goal completions that fired on the NullNotificationSink before the real sink was wired get re-emitted on the next tick.

### 4. goalValidationCriteria population — RESOLVED

In `checkGoalCompletion()` at lines 560-568 of `space-runtime.ts`, when `goalRepo` is available, the goal's `description` is used as `goalValidationCriteria`. The field is included in the emitted `goal_tasks_complete` event at line 575.

### 5. Test coverage — VERIFIED

Three new tests added in `packages/daemon/tests/unit/space/space-runtime.test.ts`:
1. **"uses runtime-replaced sink from setNotificationSink() not the original config sink"** — Verifies that after `setNotificationSink()`, events go to the new sink (not the old config sink)
2. **"clears goalIterations on setNotificationSink so pre-provisioning completions are re-emitted"** — Verifies the race condition fix: completions that fired on the pre-provisioning sink are re-emitted after the real sink is wired
3. **"includes goalTitle and goalValidationCriteria from goalRepo in emitted event"** — Verifies that `GoalRepository` data (title, description) is correctly included in the `GoalTasksCompleteEvent`

All notification sink references in `space-runtime.ts` correctly use `this.notificationSink` (never `this.config.notificationSink`), confirming the fix from commit `261bdff8`.

**Recommendation: APPROVE** — All P0 and P1 issues from the previous review have been addressed with correct implementations and adequate test coverage.

PR #534, opened 2026-03-21
test: verify all MCP tools registered in global-spaces provisioning (Task 5.3)

lsm:
Add provision-global-agent.test.ts with 17 tests confirming that:
- createGlobalSpacesToolHandlers exposes all 17 expected tools as methods
- All five coordination tools (create_standalone_task, get_task_detail,
  retry_task, cancel_task, reassign_task) are present with no extras or gaps
- createGlobalSpacesMcpServer returns a valid SDK server config named
  "global-spaces-tools" with an MCP instance attached
- The db field in GlobalSpacesToolsConfig is wired through to
  SpaceTaskManager so coordination tools can perform real DB operations
- All five coordination tools exercise real SQLite (not stubs) end-to-end

### DIFF
--- a/packages/daemon/tests/unit/space/provision-global-agent.test.ts
+++ b/packages/daemon/tests/unit/space/provision-global-agent.test.ts
@@ -7,9 +7,15 @@
  *    being injected into the global `spaces:global` session via sessionFactory.injectMessage()
  * 3. Wiring order: session created → sink created → setNotificationSink() called
  * 4. Provisioning also works on restart (session already exists)
+ *
+ * Also verifies (Task 5.3 — tool registration):
+ * 5. createGlobalSpacesToolHandlers exposes all 17 expected tools as methods
+ * 6. createGlobalSpacesMcpServer registers all 17 tools in the MCP instance
+ * 7. provisionGlobalSpacesAgent passes a "global-spaces-tools" server with all
+ *    17 tools (including the 5 coordination tools) to setRuntimeMcpServers()
  */
 
-import { describe, test, expect, beforeEach, afterEach, mock } from 'bun:test';
+import { describe, test, it, expect, beforeEach, afterEach, mock } from 'bun:test';
 import { rmSync, mkdirSync } from 'node:fs';
 import { join } from 'node:path';
 import { Database as BunDatabase } from 'bun:sqlite';
@@ -24,15 +30,57 @@ import { SpaceAgentRepository } from '../../../src/storage/repositories/space-ag
 import { SpaceAgentManager } from '../../../src/lib/space/managers/space-agent-manager.ts';
 import { SpaceWorkflowManager } from '../../../src/lib/space/managers/space-workflow-manager.ts';
 import { SpaceManager } from '../../../src/lib/space/managers/space-manager.ts';
+import {
+	createGlobalSpacesMcpServer,
+	createGlobalSpacesToolHandlers,
+	type GlobalSpacesToolsConfig,
+	type GlobalSpacesState,
+} from '../../../src/lib/space/tools/global-spaces-tools.ts';
 import type { SessionFactory } from '../../../src/lib/room/runtime/task-group-manager.ts';
 import type { SessionManager } from '../../../src/lib/session-manager.ts';
 import type { SpaceAgentManager as ISpaceAgentManager } from '../../../src/lib/space/managers/space-agent-manager.ts';
 import type { SpaceWorkflowManager as ISpaceWorkflowManager } from '../../../src/lib/space/managers/space-workflow-manager.ts';
 import type { SpaceTaskRepository as ISpaceTaskRepository } from '../../../src/storage/repositories/space-task-repository.ts';
 import type { SpaceWorkflowRunRepository as ISpaceWorkflowRunRepository } from '../../../src/storage/repositories/space-workflow-run-repository.ts';
+import type { SpaceRuntime } from '../../../src/lib/space/runtime/space-runtime.ts';
 import type { MessageDeliveryMode } from '@neokai/shared';
-import type { GlobalSpacesState } from '../../../src/lib/space/tools/global-spaces-tools.ts';
 import type { NotificationSink } from '../../../src/lib/space/runtime/notification-sink.ts';
+
+// ---------------------------------------------------------------------------
+// Expected tool lists (Task 5.3)
+// ---------------------------------------------------------------------------
+
+/** Complete set of tools that must be registered in the global-spaces MCP server. */
+const EXPECTED_TOOLS = [
+	// Cross-space tools
+	'list_spaces',
+	'create_space',
+	'get_space',
+	'update_space',
+	'archive_space',
+	'delete_space',
+	// Per-space tools
+	'list_workflows',
+	'get_workflow_detail',
+	'start_workflow_run',
+	'get_workflow_run',
+	'list_tasks',
+	'suggest_workflow',
+	// Coordination tools (Milestone 3 / Task 3.3)
+	'create_standalone_task',
+	'get_task_detail',
+	'retry_task',
+	'cancel_task',
+	'reassign_task',
+] as const;
+
+const COORDINATION_TOOLS = [
+	'create_standalone_task',
+	'get_task_detail',
+	'retry_task',
+	'cancel_task',
+	'reassign_task',
+] as const;
 
 // ---------------------------------------------------------------------------
 // DB helpers
@@ -207,7 +255,41 @@ function buildDeps(
 }
 
 // ---------------------------------------------------------------------------
-// Tests
+// Minimal stub config for tool-layer tests (no SpaceRuntimeService needed)
+// ---------------------------------------------------------------------------
+
+function makeMinimalToolConfig(db: BunDatabase): GlobalSpacesToolsConfig {
+	return {
+		spaceManager: {
+			createSpace: mock(async () => ({ id: 's1', name: 'S', workspacePath: '/tmp' })),
+			getSpace: mock(() => null),
+			listSpaces: mock(() => []),
+			updateSpace: mock(async () => ({ id: 's1', name: 'S', workspacePath: '/tmp' })),
+			archiveSpace: mock(async () => ({ id: 's1', status: 'archived' })),
+			deleteSpace: mock(async () => true),
+		} as unknown as SpaceManager,
+		spaceAgentManager: { listBySpaceId: mock(() => []) } as unknown as ISpaceAgentManager,
+		runtime: {} as unknown as SpaceRuntime,
+		workflowManager: {
+			listWorkflows: mock(() => []),
+			getWorkflow: mock(() => null),
+		} as unknown as ISpaceWorkflowManager,
+		taskRepo: new SpaceTaskRepository(db),
+		workflowRunRepo: new SpaceWorkflowRunRepository(db),
+		db,
+	};
+}
+
+/** Extract registered tool names from an MCP server's internal registry. */
+function getRegisteredToolNames(
+	server: ReturnType<typeof createGlobalSpacesMcpServer>
+): string[] {
+	const instance = server.instance as unknown as { _registeredTools: Record<string, unknown> };
+	return Object.keys(instance._registeredTools);
+}
+
+// ---------------------------------------------------------------------------
+// Tests: notification sink wiring
 // ---------------------------------------------------------------------------
 
 describe('provisionGlobalSpacesAgent', () => {
@@ -318,7 +400,7 @@ describe('provisionGlobalSpacesAgent', () => {
 	test('throws and does NOT wire sink when session creation fails', async () => {
 		const sessionManager = {
 			createCalls: 0,
-			getSessionAsync: mock(async () => null), // always null → triggers create
+			getSessionAsync: mock(async () => null), // always null triggers create
 			createSession: mock(async () => {
 				throw new Error('DB write failed');
 			}),
@@ -327,7 +409,7 @@ describe('provisionGlobalSpacesAgent', () => {
 		const deps = buildDeps(db, { sessionManager });
 
 		await expect(provisionGlobalSpacesAgent(deps)).rejects.toThrow('DB write failed');
-		// Sink must NOT be wired — spaceRuntimeService keeps NullNotificationSink
+		// Sink must NOT be wired spaceRuntimeService keeps NullNotificationSink
 		expect(deps.spyService.sinkCalls).toHaveLength(0);
 	});
 
@@ -353,7 +435,7 @@ describe('provisionGlobalSpacesAgent', () => {
 	});
 
 	// -------------------------------------------------------------------------
-	// End-to-end: tick → notification → injectMessage
+	// End-to-end: tick notification injectMessage
 	// -------------------------------------------------------------------------
 
 	test('end-to-end: a tick producing a task_needs_attention event injects a message into the global session', async () => {
@@ -385,11 +467,11 @@ describe('provisionGlobalSpacesAgent', () => {
 		});
 
 		const { tasks } = await runtime.startWorkflowRun(SPACE_ID, workflow.id, 'E2E Run');
-		// Simulate task failure — set to needs_attention
+		// Simulate task failure set to needs_attention
 		const taskRepo = deps.taskRepo as SpaceTaskRepository;
 		taskRepo.updateTask(tasks[0].id, { status: 'needs_attention', error: 'Build failed' });
 
-		// Trigger a manual tick — this should detect needs_attention tasks and emit notifications
+		// Trigger a manual tick this should detect needs_attention tasks and emit notifications
 		await runtime.executeTick();
 
 		// The notification should have been injected into the spaces:global session
@@ -405,10 +487,6 @@ describe('provisionGlobalSpacesAgent', () => {
 	// -------------------------------------------------------------------------
 
 	test('startup race: tasks notified via NullNotificationSink before wiring are re-notified after setNotificationSink', async () => {
-		// This test guards against the startup race where:
-		// 1. SpaceRuntimeService starts and fires a tick (dedup key added, NullNotificationSink = no-op)
-		// 2. provisionGlobalSpacesAgent runs and wires the real sink
-		// 3. The task should still notify on the next tick (dedup set was cleared by setNotificationSink)
 		const SPACE_ID = 'space-race-fix';
 		const AGENT_ID = 'agent-race-fix';
 		const STEP_ID = 'step-race-fix';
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured test-only PR that adds 17 tests verifying the 5 coordination tools are properly registered and wired in the global spaces MCP server. All tests pass.

### What works well

- **Tool list regression guard** (Section 1): The exact-match test at line 155 (`handlerKeys.toEqual(expectedKeys)`) catches both missing and unexpected tools -- good defensive testing.
- **Real SQLite end-to-end tests** (Section 3): Testing `create_standalone_task`, `get_task_detail`, `retry_task`, `cancel_task`, and `reassign_task` against a real database with migrations is much more valuable than mocking.
- **Error path coverage** (Section 4): Not-found cases and missing space context are tested.
- **Clean DB lifecycle**: Each `describe` block creates/tears down its own DB directory, avoiding cross-test contamination.

### Minor observations (P3)

1. **File name vs. what's tested**: The file is named `provision-global-agent.test.ts` but it tests `createGlobalSpacesToolHandlers` and `createGlobalSpacesMcpServer` from `global-spaces-tools.ts`. The actual `provisionGlobalSpacesAgent()` function in `provision-global-agent.ts` (which creates the session, calls `createGlobalSpacesMcpServer`, and attaches it) is not tested. This is fine for now since the underlying tools module is the more important thing to verify, but the naming is slightly misleading.

2. **MCP server property access via double cast** (lines 172, 177): `(server as unknown as { name: string }).name` is fragile -- if the SDK changes the return type, these tests will pass TypeScript compilation but may fail at runtime. This is a minor concern for test code.

Overall this is a solid test suite that achieves the stated goal of verifying tool registration completeness and db wiring for the 5 coordination tools.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The tests pass and the implementation is sound, but the test file has three issues worth fixing before merge.

---

### P1 — Test file never imports or calls `provisionGlobalSpacesAgent`

The file is named `provision-global-agent.test.ts` and the PR description says it verifies that `db` is wired into `provisionGlobalSpacesAgent`. But the function is never imported from `provision-global-agent.ts` or invoked in any test. What is actually being tested is `createGlobalSpacesToolHandlers` and `createGlobalSpacesMcpServer` from `global-spaces-tools.ts`.

This is a scoping gap: if someone later removes the `db` field from `ProvisionGlobalSpacesAgentDeps` in `provision-global-agent.ts` (lines 35-36 in that file), or stops passing it to `createGlobalSpacesMcpServer` (lines 91-102), these tests would still pass because they bypass the provisioning layer entirely.

The PR description's claim that the tests verify "the db field in GlobalSpacesToolsConfig is wired through to SpaceTaskManager" is true — but they verify it at the `global-spaces-tools.ts` layer, not at the `provision-global-agent.ts` layer as stated. At minimum the file should either be renamed to reflect its real scope, or add a test that actually calls `provisionGlobalSpacesAgent` with a mock `SessionManager` and asserts the returned session has the expected MCP server registered.

`packages/daemon/tests/unit/space/provision-global-agent.test.ts`

---

### P1 — Section 2 (MCP server registration) does not verify tool registration within the MCP instance

Tests at lines 199-202 only assert that `server.instance` is defined (truthy). They don't verify that the 17 tools are actually registered in that MCP instance. If a future refactor moves the tool registration after the server is returned (or accidentally drops some tools from the `tools` array in `createGlobalSpacesMcpServer`), Section 2 would still pass.

The `McpSdkServerConfigWithInstance` type has an `instance: McpServer` field. That `McpServer` instance (from the MCP SDK) exposes a way to introspect registered tools. The test should enumerate the tool names from `server.instance` (or via the tool array passed to `createSdkMcpServer`) and assert they match `EXPECTED_TOOLS`. Section 1 already does this at the handler layer — it should be mirrored at the MCP layer too.

`packages/daemon/tests/unit/space/provision-global-agent.test.ts`, lines 188-207

---

### P2 — Significant overlap with `global-spaces-tools.test.ts` Section 2

The coordination-tool behavior tested in Sections 3 and 4 of the new file (db wiring, not-found paths, cancel, retry, reassign, create_standalone_task with no space context) is largely duplicated from `global-spaces-tools.test.ts` Section 2, which already covers the same real-DB integration scenarios. The not-found paths in particular are identical.

This is not a correctness issue, but the duplication inflates maintenance cost. Consider whether Sections 3 and 4 could live in `global-spaces-tools.test.ts` and the new file focuses solely on the provisioning-layer concern (calling `provisionGlobalSpacesAgent` and checking the session wiring).

`packages/daemon/tests/unit/space/provision-global-agent.test.ts`, lines 209-380

---

### P3 — Unused `mock` import

`mock` is imported on line 12 from `bun:test` but is only used inside `makeMinimalConfig` for the spy functions. This is fine and not a lint error, but it is worth noting for clarity.

---

### What is correct

- All 17 tests pass against the real implementation.
- The `EXPECTED_TOOLS` list exactly matches the 17 tools registered in `createGlobalSpacesMcpServer` (verified by reading `global-spaces-tools.ts` lines 483-737).
- The handler-count equality test (line 162) provides a strong regression guard against accidental additions or deletions at the handler layer.
- The real-SQLite wiring tests (Section 3) genuinely exercise the `SpaceTaskManager` path and confirm the `db` field reaches task persistence.
- The `cancel_task` transition (pending → cancelled) and `retry_task` transition (needs_attention → pending) are both valid per `VALID_SPACE_TASK_TRANSITIONS` in `space-task-manager.ts`.
- The `seedSpace` SQL matches the schema used by `global-spaces-tools.test.ts`.

lsm (commented):

**Recommendation: APPROVE**

The follow-up changes address the prior review feedback well. The test file is clean, well-structured, and all 14 tests pass.

### What was reviewed

Single test file: `packages/daemon/tests/unit/space/provision-global-agent.test.ts` (372 lines)

### Assessment

**Correctness**: The tests correctly verify the original ask — that all 17 tools (including the 5 new coordination tools) are registered in the MCP server during global agent provisioning. Three test layers cover this:

1. **Section 1 (handler-layer)**: Verifies `createGlobalSpacesToolHandlers` returns an object with exactly the 17 expected tool names as methods.
2. **Section 2 (MCP-instance-layer)**: Verifies `createGlobalSpacesMcpServer` registers all 17 tools in the MCP SDK's internal `_registeredTools` map.
3. **Section 3 (provisioning-layer)**: Verifies `provisionGlobalSpacesAgent` calls `setRuntimeMcpServers` with a `"global-spaces-tools"` key whose MCP instance contains all 17 registered tools. Also covers the restart scenario (session already exists).

**Completeness**: The duplicate real-SQLite integration tests from the prior version have been removed as requested. The MCP instance inspection now uses `_registeredTools` consistently.

**Minor observations (P3)**:

- The `_registeredTools` access pattern relies on MCP SDK internals. This is acknowledged in comments and is acceptable for test-only code — if the SDK changes the internal field name, the tests will fail loudly rather than silently pass.
- The `makeProvisionDeps` `spaceManager` mock omits `archiveSpace` and `deleteSpace`, but this is fine since the provisioning tests never invoke those tools.

lsm (commented):

**Recommendation: APPROVE**

The two P1 issues from the prior review have been fully addressed. The test file now correctly imports and exercises `provisionGlobalSpacesAgent`, and the MCP instance inspection uses `instance._registeredTools` for exact tool-name verification at every relevant layer.

---

### What was verified

I confirmed `_registeredTools` is a real own-instance property on `McpServer` (from `@modelcontextprotocol/sdk`) by running the SDK at runtime — keys are: `_registeredResources`, `_registeredResourceTemplates`, `_registeredTools`, `_registeredPrompts`, and others. The test approach is sound and not guessing at internals blindly.

**Prior P1 — `provisionGlobalSpacesAgent` never called: RESOLVED.**
Section 3 (lines 219–372) now imports and calls `provisionGlobalSpacesAgent` directly. Tests cover the first-run path (session created, `setRuntimeMcpServers` called once with a `"global-spaces-tools"` key containing all 17 tools) and the restart path (session already exists, `createSession` is skipped). The provisioning-layer gap is closed.

**Prior P1 — Section 2 MCP instance not introspected: RESOLVED.**
`getRegisteredToolNames` (lines 181–186) retrieves tool names via `instance._registeredTools` and performs an exact sorted-list comparison against `EXPECTED_TOOLS`. The tool-count guard (line 211) also catches unexpected additions.

**Prior P2 — Duplicate real-SQLite tests removed: RESOLVED.**
Sections 3 and 4 from the previous revision (real-DB integration tests overlapping with `global-spaces-tools.test.ts`) have been dropped. The new file stays focused on the provisioning layer.

---

### Minor observations (P3 only)

**Redundant `mockResolvedValueOnce(null)` on line 249.**
`getSessionAsync` is initialized as `mock(async () => null)` on line 248, so the base implementation already returns null. Line 249 queues an additional `null` for the first call that is functionally identical to what the base would return. The mock still works correctly — the two queued `Once` values (null, fakeSession) are consumed in order — but the explicit `null` queue entry adds a mildly misleading impression that the first call needs special treatment. Harmless; worth noting for clarity.

**Redundant type cast in `makeProvisionDeps` tests (lines 312, 320, 333).**
`setRuntimeMcpServers` is already typed as `ReturnType<typeof mock>` (it comes directly from `mock(() => {})` on line 238 and is returned from `makeProvisionDeps` as that type), so the `as ReturnType<typeof mock>` cast in the call-inspection lines is redundant. No impact on correctness.

---

### Overall

All 14 tests pass. The three test layers (handler, MCP-instance, provisioning) together form a solid regression guard: a future refactor that accidentally drops a coordination tool from the handler object, the `tools` array in `createGlobalSpacesMcpServer`, or the `createGlobalSpacesMcpServer` call in `provisionGlobalSpacesAgent` would be caught by a corresponding failing test.

lsm (commented):

### Summary

The merge conflict resolution is clean and the test file runs successfully (17 tests pass). The new Task 5.3 tests correctly verify that all 17 tools (including the 5 coordination tools) are registered at three layers: handler, MCP instance, and provisioning. However, there is a meaningful gap in the test fixture that should be addressed.

---

### P1: `buildDeps()` missing `db` property — tests pass with `undefined` db

**File:** `packages/daemon/tests/unit/space/provision-global-agent.test.ts`, lines 243-254

The `ProvisionGlobalSpacesAgentDeps` interface (in `provision-global-agent.ts:29-47`) requires a `db: BunDatabase` property, but `buildDeps()` does not include `db` in its return object. The same omission occurs in the manually-constructed `deps` object at line 379.

This is not caught by TypeScript because `tests/**/*` is excluded from the tsconfig's `include` (see `packages/daemon/tsconfig.json` line 19). At runtime, `provisionGlobalSpacesAgent` destructures `db` as `undefined` and passes it into `createGlobalSpacesMcpServer`. The tests still pass because no tool handler is actually *invoked* — only registration is verified — so the `undefined` db never triggers a runtime error.

**Impact:** The three new provisioning-layer tests (lines 536-590) are creating an MCP server with an invalid `GlobalSpacesToolsConfig` (`db: undefined`). If a future test attempts to call any coordination tool handler through the provisioned MCP server, it will fail with an opaque error instead of a clear setup failure.

**Fix:** Add `db` to the return object in `buildDeps()` (line 243) and to the manual `deps` construction at line 379:

```ts
return {
    sessionManager,
    spaceManager,
    // ... existing fields ...
    state,
    db,       // <-- add this
    spyService,
};
```

---

### P2: Comment degradation from merge conflict resolution

Several comments had their em-dashes (`—`) and arrows (`→`) stripped during the conflict resolution, making them less readable:

- Line 403: `// always null triggers create` (was: `// always null → triggers create`)
- Line 412: `// Sink must NOT be wired spaceRuntimeService keeps NullNotificationSink` (was: `// Sink must NOT be wired — spaceRuntimeService keeps NullNotificationSink`)
- Line 438: `// End-to-end: tick notification injectMessage` (was: `// End-to-end: tick → notification → injectMessage`)
- Line 470, 474, 518, 523, 526: Similar degradation

These appear to be encoding issues from the merge tool. The non-ASCII characters should be restored.

---

### P3: Significant test overlap across the three describe blocks

The handler completeness tests, MCP instance tests, and provisioning-layer tests all verify the same 17-tool list against `EXPECTED_TOOLS`. While testing at multiple layers is reasonable, the current structure has near-identical assertions repeated three times (e.g., "all 17 tools present" appears in all three groups). Consider whether the handler and MCP-instance groups could be consolidated, keeping just one unit-level check and the provisioning integration check.

---

### What looks good

- The `EXPECTED_TOOLS` and `COORDINATION_TOOLS` constants are well-defined and serve as a single source of truth
- The `makeMinimalToolConfig` helper cleanly isolates the tool-layer tests from `SpaceRuntimeService`
- The `getRegisteredToolNames` helper using `_registeredTools` is a reasonable approach for verifying MCP registration
- Merge conflict resolution preserved all 8 original Task 5.2 tests correctly
- All 17 tests pass with `bun test`

lsm (commented):

---

### Summary

All 17 tests pass and the core task objective is correctly implemented — the five coordination tools are registered in the MCP server and provisionGlobalSpacesAgent passes them to setRuntimeMcpServers. However, the merge conflict resolution introduced comment degradation that reduces readability, and there is a latent test fixture bug where buildDeps omits the required db field from the returned deps object, leaving deps.db === undefined in the three new provisioning-layer tests. Additionally, the PR description's claim that coordination tools are tested end-to-end with real SQLite is inaccurate.

---

### P1 — Missing `db` field in `buildDeps` return object

**File:** `packages/daemon/tests/unit/space/provision-global-agent.test.ts`, lines 243–254

The `buildDeps` function declares its return type as `ProvisionGlobalSpacesAgentDeps & { ... }`. `ProvisionGlobalSpacesAgentDeps` requires `db: BunDatabase` (defined in `provision-global-agent.ts` line 44), but the returned object literal does not include `db`:

```ts
return {
    sessionManager,
    spaceManager,
    spaceAgentManager: agentManager,
    spaceWorkflowManager: workflowManager,
    spaceRuntimeService: spyService,
    sessionFactory,
    taskRepo,
    workflowRunRepo,
    state,
    spyService,
    // db is missing
};
```

This means `deps.db` is `undefined` in all three new provisioning-layer tests (lines 544, 559, 579). TypeScript does not catch this because `tests/**/*` is excluded from `packages/daemon/tsconfig.json`.

The tests still pass because `db` is only used lazily inside `getOrCreateTaskManager(spaceId)` — which is never called during tool-registration checks. However, any future test that calls a coordination tool through `provisionGlobalSpacesAgent`'s deps will hit a runtime error.

**The fix is one line:** add `db,` to the return object of `buildDeps`.

---

### P1 — PR description claim about end-to-end SQLite testing is inaccurate

The PR body states: "All five coordination tools exercise real SQLite (not stubs) end-to-end". No test actually *invokes* a coordination tool. The three handler-layer tests only call `Object.keys(handlers)` and `typeof handlers[toolName]`. The MCP-instance tests only read `_registeredTools`. The `makeMinimalToolConfig` does construct a real `SpaceTaskRepository(db)`, but no coordination tool handler is exercised. This makes the PR description misleading.

---

### P2 — Merge conflict resolution degraded comment readability throughout

The conflict resolution stripped em-dashes (—) and arrows (→) from several comments, making them harder to parse:

**Line 412:**
```
// Sink must NOT be wired spaceRuntimeService keeps NullNotificationSink
```
Should be:
```
// Sink must NOT be wired — spaceRuntimeService keeps NullNotificationSink
```

**Line 437 (section header):**
```
// End-to-end: tick notification injectMessage
```
Should be:
```
// End-to-end: tick → notification → injectMessage
```

**Lines 470, 474:**
```
// Simulate task failure set to needs_attention
// Trigger a manual tick this should detect needs_attention tasks and emit notifications
```
Should have `—` after `failure` and after `tick`.

**Lines 518, 523, 526:** Same pattern — em-dashes stripped.

**Lines 487–490 entirely deleted** from the startup-race test — a four-line comment block explaining the startup race scenario was removed:
```ts
// This test guards against the startup race where:
// 1. SpaceRuntimeService starts and fires a tick (dedup key added, NullNotificationSink = no-op)
// 2. provisionGlobalSpacesAgent runs and wires the real sink
// 3. The task should still notify on the next tick (dedup set was cleared by setNotificationSink)
```
This context is important for future maintainers.

---

### P3 — Inconsistent use of `test` vs `it`

The three new Task 5.3 provisioning-layer tests (lines 536, 551, 571) use `it(...)` while all prior tests in the same `describe` block use `test(...)`. The import adds `it` to the import list just for these three. The codebase pattern is `test`, so the new tests should use `test` for consistency.

---

### What is correct

- **Core task objective achieved:** `createGlobalSpacesMcpServer` correctly registers all 17 tools including the 5 coordination tools, and `provisionGlobalSpacesAgent` correctly passes the server keyed as `global-spaces-tools` to `setRuntimeMcpServers`.
- **`_registeredTools` access is valid:** This is a public typed property (`McpServer._registeredTools`) on the SDK version in use (0.2.77), not a private API access.
- **`makeMinimalToolConfig` uses real SQLite:** The handler-layer and MCP-instance tests use a real `BunDatabase`, not stubs, for the repos.
- **Test isolation is sound:** Each describe block has proper `beforeEach`/`afterEach` with temp-dir cleanup.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review round have been addressed in commit `0d7d056`:

1. **P1 -- `db` missing from `buildDeps()` return**: Fixed. `db` is now included in both `buildDeps()` (line 252) and the inline deps object in the order-check test (line 389). The provisioning-layer tests now pass a real `BunDatabase` through to `createGlobalSpacesMcpServer`, matching production behavior.

2. **P2 -- Em-dashes and arrows stripped**: All 7 degraded comments have been restored with proper Unicode characters (lines 403, 412, 438, 470, 474, 523, 528).

3. **P2 -- Deleted explanatory comment block**: The 4-line comment explaining the startup race scenario has been re-inserted at lines 490-493.

4. **P3 -- `it` vs `test` inconsistency**: All three Task 5.3 provisioning-layer tests now use `test(...)` (lines 540, 555, 575), and the unused `it` import has been removed from line 18.

### Verification

- All 17 tests pass with 54 expect() calls (823ms).
- The test file correctly covers three layers: handler completeness, MCP instance registration, and provisioning integration.
- No new issues introduced by the fixes.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been correctly addressed. The test file is clean and all 17 tests pass.

---

### Verification of fixes

**P1 — `db` added to `buildDeps()` return and inline deps: RESOLVED.**
Line 252 in `buildDeps()` now includes `db` in the return object, and the manually-constructed `deps` literal in the order-check test at line 389 also includes `db`. The three provisioning-layer tests now receive a real `BunDatabase` instance rather than `undefined`.

**P2 — Em-dashes and arrows restored: RESOLVED.**
Verified all seven comment locations mentioned in the previous review (lines 403, 412, 438, 470, 474, 522, 527) now contain the correct `→` and `—` characters.

**P2 — Four-line comment block in startup-race test restored: RESOLVED.**
Lines 490–493 now contain the full explanatory comment documenting the race condition scenario.

**P3 — `it()` changed to `test()`: RESOLVED.**
No `it()` calls remain in the file. All 17 tests use `test()` consistently.

---

### Test run

All 17 tests pass (777ms). The single log line for the `'DB write failed'` error is expected behavior from the error-handling test — it comes from the internal logger, not a test failure.

---

None that weren't already noted in the previous round. The previously noted minor items (redundant `mock(async () => null)` queue entry, redundant `ReturnType<typeof mock>` casts, and the `_registeredTools` SDK-internal access) are unchanged and remain acceptable for test-only code.

---

### Overall assessment

The test file now correctly tests all three layers (handler, MCP instance, provisioning) with real SQLite fixtures, and the `db` field is properly threaded through to every layer. The regression guard is solid: dropping any of the 17 tools from the handler object, the `tools` array in `createGlobalSpacesMcpServer`, or the server registration call in `provisionGlobalSpacesAgent` would be caught by a failing test.

PR #535, opened 2026-03-21
feat: add send_message_to_task tool to Space Agent

lsm:
Add a new `send_message_to_task` MCP tool that allows the Space Agent
to inject messages into an active Task Agent session. The tool:
- Looks up the task by ID via SpaceTaskRepository
- Validates the task has a taskAgentSessionId
- Delegates to TaskAgentManager.injectTaskAgentMessage()
- Returns success/failure with the task's current status

TaskAgentManager is an optional field on SpaceAgentToolsConfig (null-safe)
so existing callers that don't have the Task Agent system set up continue
to work — the tool returns a descriptive error in that case.

Also documents the new tool in the Space Chat Agent system prompt and
adds 5 unit tests covering all edge cases.

### DIFF
--- a/packages/daemon/src/lib/space/agents/space-chat-agent.ts
+++ b/packages/daemon/src/lib/space/agents/space-chat-agent.ts
@@ -27,6 +27,7 @@
  *       - retry_task
  *       - cancel_task
  *       - reassign_task
+ *       - send_message_to_task
  *
  * See: docs/plans/multi-agent-v2-customizable-agents-workflows/07-workflow-selection-intelligence.md
  */
@@ -277,6 +278,13 @@ export function buildSpaceChatSystemPrompt(context: SpaceChatAgentContext = {}):
 	sections.push(
 		`- **\`reassign_task\`** — Change the assigned agent for a task. Valid for tasks in \`pending\`, ` +
 			`\`needs_attention\`, or \`cancelled\` status. Use when a different agent would be better suited.`
+	);
+	sections.push('');
+	sections.push(
+		`- **\`send_message_to_task\`** — Send a message directly to the Task Agent session managing a ` +
+			`specific task. Use this to check on progress, provide feedback, or redirect work mid-execution. ` +
+			`The message is delivered to the Task Agent which may relay relevant parts to its active sub-session. ` +
+			`Only works for tasks that have an active Task Agent session (i.e., tasks in \`in_progress\` or \`review\` status).`
 	);
 
 	return sections.join('\n');
--- a/packages/daemon/src/lib/space/tools/space-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/space-agent-tools.ts
@@ -435,7 +435,9 @@ export function createSpaceAgentToolHandlers(config: SpaceAgentToolsConfig) {
 				});
 			}
 
-			const task = taskRepo.getTask(args.task_id);
+			// Use taskManager.getTask() (not taskRepo) to enforce space ownership — ensures
+			// Space Agent A cannot inject messages into Space B's Task Agent sessions.
+			const task = await taskManager.getTask(args.task_id);
 			if (!task) {
 				return jsonResult({ success: false, error: `Task not found: ${args.task_id}` });
 			}
@@ -454,7 +456,6 @@ export function createSpaceAgentToolHandlers(config: SpaceAgentToolsConfig) {
 					success: true,
 					taskId: task.id,
 					taskStatus: task.status,
-					taskAgentSessionId: task.taskAgentSessionId,
 				});
 			} catch (err) {
 				const message = err instanceof Error ? err.message : String(err);
--- a/packages/daemon/src/lib/space/tools/space-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/space-agent-tools.ts
@@ -29,6 +29,7 @@ import type { SpaceTaskRepository } from '../../../storage/repositories/space-ta
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
 import type { SpaceTaskManager } from '../managers/space-task-manager';
 import type { SpaceAgentManager } from '../managers/space-agent-manager';
+import type { TaskAgentManager } from '../runtime/task-agent-manager';
 import { jsonResult, SUGGEST_WORKFLOW_STOP_WORDS } from './tool-result';
 import type { ToolResult } from './tool-result';
 
@@ -51,6 +52,11 @@ export interface SpaceAgentToolsConfig {
 	taskManager: SpaceTaskManager;
 	/** Space agent manager for reassign validation. */
 	spaceAgentManager: SpaceAgentManager;
+	/**
+	 * Task Agent Manager for injecting messages into Task Agent sessions.
+	 * Optional — when not provided, send_message_to_task returns an error.
+	 */
+	taskAgentManager?: TaskAgentManager | null;
 }
 
 // ---------------------------------------------------------------------------
@@ -70,6 +76,7 @@ export function createSpaceAgentToolHandlers(config: SpaceAgentToolsConfig) {
 		workflowRunRepo,
 		taskManager,
 		spaceAgentManager,
+		taskAgentManager,
 	} = config;
 
 	return {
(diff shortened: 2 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured addition of the `send_message_to_task` tool. The implementation follows the existing patterns in the file closely and all 59 tests pass (including 5 new ones for this tool).

### What looks good

- **Null-safe optional dependency**: `taskAgentManager?: TaskAgentManager | null` is optional on the config, so existing callers that don't have it continue to work. The handler returns a descriptive error when it's not configured.
- **Consistent error handling**: The try/catch around `injectTaskAgentMessage` follows the same `err instanceof Error ? err.message : String(err)` pattern used by every other handler in this file.
- **Proper validation chain**: task_id lookup -> taskAgentSessionId check -> actual injection. Each step returns a descriptive error on failure.
- **Zod schema**: Consistent with the existing `z.string().describe(...)` pattern used by all other tools (no `.min(1)` anywhere).
- **System prompt**: Well-documented with usage guidance and status constraints.
- **Test coverage**: All 5 edge cases are covered -- no taskAgentManager, task not found, no session ID, success path, and error propagation from `injectTaskAgentMessage`.

### Minor observations (P3)

1. **`packages/daemon/src/lib/space/tools/space-agent-tools.ts:438`** -- The tool reads `task.status` from the repository row, which gives the DB state at lookup time. If the task status changes between the `getTask()` call and the response being sent, the returned `taskStatus` could be stale. This is a typical TOCTOU issue in async code but is effectively harmless here since the status is informational only.

2. **`packages/daemon/tests/unit/space/space-agent-tools.test.ts:1297`** -- The success test sets `taskAgentSessionId` manually via `updateTask` then calls `startTask`. It might be slightly more readable to set the session ID *after* starting, since in production the session ID is set when the agent session is created (which happens during/after start). But the test works correctly as-is since `startTask` doesn't clear the session ID.

lsm (commented):

The implementation is correct for the happy path and the tests pass. The core logic — null-guard on `taskAgentManager`, task lookup, sessionId check, delegation to `injectTaskAgentMessage`, error propagation — is clean and well-structured. That said, there is one security/correctness issue and several smaller gaps worth fixing.

---

### P1 — Missing space ownership check in `send_message_to_task`

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, line 438

Every other task-touching handler in this file (`get_task_detail`, `retry_task`, `cancel_task`, `reassign_task`) uses `taskManager.getTask(args.task_id)`, which internally validates `task.spaceId === this.spaceId` (see `SpaceTaskManager.getTask()` at line 72 of `space-task-manager.ts`). The new handler is the only one that calls `taskRepo.getTask(args.task_id)` directly, bypassing space scoping entirely.

Since `TaskAgentManager` is a global singleton (holds sessions for all spaces), a Space Agent in Space A could supply a `task_id` belonging to Space B, pass the `taskAgentSessionId` check on the B-space task, and successfully inject a message into Space B's Task Agent session. Fix: replace `taskRepo.getTask(args.task_id)` with `await taskManager.getTask(args.task_id)` — which enforces the ownership guard — or add an explicit `task.spaceId !== spaceId` check after the lookup.

There is no test covering this cross-space case; one should be added.

---

### P2 — `taskAgentManager` is never wired in production code

**Files:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, `packages/daemon/src/lib/space/runtime/space-runtime.ts`, all callers of `createSpaceAgentMcpServer`

A project-wide search shows that `createSpaceAgentMcpServer` is exported but never called by any production code (it only appears in test imports that reference `createSpaceAgentToolHandlers` directly). The `taskAgentManager` optional field is therefore always `undefined` in production, which means the tool always returns the `'TaskAgentManager is not configured'` error at runtime. The implementation is complete but dead until the wiring is done.

This does not need to block the PR if it is intentionally designed as a forward-looking stub, but it should be documented. If the intent is that the tool should be functional now, the wiring is missing.

---

### P2 — `send_message_to_task` is not listed in the `create_standalone_task` tool test (also missing from `space-chat-agent.test.ts`)

**File:** `packages/daemon/tests/unit/space/space-chat-agent.test.ts`

`buildSpaceChatSystemPrompt` now emits a reference to `send_message_to_task`, but the existing space-chat-agent unit tests do not assert that this tool name appears in the generated prompt. Minor, but leaves a gap in regression coverage for the system prompt contract.

---

### P2 — Response leaks internal session ID

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, lines 453-458

The success response includes `taskAgentSessionId`. The Space Agent is an AI coordinator; exposing the internal session ID in the tool response gives it information it cannot usefully act on (it has no direct session API), and could leak infrastructure details if the Space Agent relays its tool results to the user. Consider omitting it from the success payload or at least documenting why it is intentionally included.

---

### P3 — Nit: `SpaceTaskManager.startTask` call in the happy-path test is unrelated to the assertion under test

**File:** `packages/daemon/tests/unit/space/space-agent-tools.test.ts`, lines 1276-1277

`ctx.taskRepo.updateTask(taskId, { taskAgentSessionId: sessionId })` is the setup needed. The subsequent `await ctx.taskManager.startTask(taskId)` transitions the task to `in_progress`, but the test assertion does not verify the `taskStatus` field in the response, making the startTask call superfluous. Same pattern in the error-propagation test at line 1316. This does not affect correctness but may confuse future readers.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been correctly addressed:

1. **P1 (space ownership check) -- Fixed.** `send_message_to_task` at line 440 of `space-agent-tools.ts` now calls `taskManager.getTask(args.task_id)` instead of `taskRepo.getTask()`. Confirmed that `SpaceTaskManager.getTask()` (line 70-76 of `space-task-manager.ts`) checks `task.spaceId === this.spaceId` and returns `null` for mismatches. The new cross-space test at line 1302 validates this enforcement end-to-end.

2. **P2 (session ID leak) -- Fixed.** The success response (lines 455-459) now returns only `taskId` and `taskStatus`. The test at line 1296 explicitly asserts `parsed.taskAgentSessionId` is `undefined`.

3. **P2 (system prompt regression test) -- Fixed.** New test at line 444 of `space-chat-agent.test.ts` asserts `send_message_to_task` appears in the generated prompt.

4. **P3 (superfluous startTask calls) -- Fixed.** The success and error-propagation tests no longer call `startTask()`.

### Remaining observations (P3 -- non-blocking)

- **Empty message not validated:** The `message` Zod schema at line 647 is `z.string()` with no `.min(1)`. An empty string `""` would be accepted and injected into the Task Agent session. This is consistent with every other string parameter in the file (none use `.min(1)`), so this is not a deviation from existing patterns, but it is worth noting that sending an empty message to a Task Agent has no useful effect.

- **TOCTOU on task status:** The `taskStatus` returned in the success response reflects the DB state at lookup time (line 440), not at injection time (line 454). If the task status changes between the two calls, the returned status would be stale. This is informational-only data in the response so the impact is negligible.

All 60 tests in `space-agent-tools.test.ts` pass. All 60 tests in `space-chat-agent.test.ts` pass. The implementation is correct, complete, and well-tested. No new issues introduced.

lsm (commented):

**Recommendation: APPROVE**

All four issues raised in the previous review have been correctly addressed. The implementation is now secure and complete.

---

### P1 (resolved) — Space ownership check

`send_message_to_task` now calls `taskManager.getTask(args.task_id)` (line 440 of `space-agent-tools.ts`), which enforces the space ownership guard via `SpaceTaskManager.getTask()` (lines 70–76 of `space-task-manager.ts`): it returns `null` for any task whose `spaceId` does not match the manager's bound space. The cross-space test (`'returns error when task belongs to a different space'`, lines 1302–1342 of `space-agent-tools.test.ts`) directly verifies that a task seeded in `space-other` is rejected by the handler bound to `ctx.spaceId`, with `injectTaskAgentMessage` never called.

### P2 (resolved) — `taskAgentSessionId` removed from success response

The success payload now contains only `taskId` and `taskStatus` (lines 455–459 of `space-agent-tools.ts`). The test at line 1296 explicitly asserts `expect(parsed.taskAgentSessionId).toBeUndefined()`.

### P2 (resolved) — System prompt documentation test

`space-chat-agent.test.ts` now includes `test('documents send_message_to_task tool', ...)` at line 444–447, asserting the tool name appears in the generated prompt. The prompt builder at `space-chat-agent.ts` line 283–288 includes the full description of `send_message_to_task` in the Coordination Tools section.

### P3 (resolved) — Superfluous `startTask` calls removed

The `send_message_to_task` tests no longer call `startTask`. The remaining `startTask` calls in other describe blocks (`retry_task`, `cancel_task`, `reassign_task`) are all legitimately required to drive tasks into the states those tests exercise.

---

### Remaining known gap (pre-existing, not blocking)

The previous P2 note that `createSpaceAgentMcpServer` is never called in production code with a `taskAgentManager` value remains unaddressed, but this was explicitly called out as acceptable in the prior review if it is a forward-looking stub. The tool correctly handles the `null`/`undefined` case with a descriptive error, so no data loss or crash risk exists.

No new issues introduced. The implementation is correct, the ownership boundary is enforced, and the test coverage is appropriate.

PR #536, opened 2026-03-21
feat: add human message routing to Task Agent via space.task.sendMessage/getMessages

lsm:
Adds two new RPC handlers to space-task-handlers.ts for direct human ↔
Task Agent communication:

- space.task.sendMessage: accepts {taskId, message}, validates that the
  task exists and has an active Task Agent session, then injects the
  message via TaskAgentManager.injectTaskAgentMessage()
- space.task.getMessages: accepts {taskId, cursor?, limit?}, returns a
  paginated snapshot of messages from the Task Agent session; prefers
  the live in-memory AgentSession when the Task Agent is active,
  falling back to the DB for completed or restarted sessions

Both handlers validate task existence, taskAgentSessionId presence,
and surface clear errors for task-not-found and session-not-started
cases. The new setupSpaceTaskMessageHandlers() function is separate
from setupSpaceTaskHandlers() because it requires a live
TaskAgentManager instance created after SpaceRuntimeService.

TaskAgentManager is now instantiated in rpc-handlers/index.ts after
SpaceRuntimeService, with all required dependencies wired up.

19 unit tests cover: happy paths, DB fallback, cursor parsing,
missing/invalid params, task-not-found, session-not-started, and
TaskAgentManager error propagation.

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -44,7 +44,12 @@ import { TaskManager } from '../room/managers/task-manager';
 import { setupDialogHandlers } from './dialog-handlers';
 // Space handlers
 import { setupSpaceHandlers } from './space-handlers';
-import { setupSpaceTaskHandlers, type SpaceTaskManagerFactory } from './space-task-handlers';
+import {
+	setupSpaceTaskHandlers,
+	setupSpaceTaskMessageHandlers,
+	type SpaceTaskManagerFactory,
+} from './space-task-handlers';
+import { TaskAgentManager } from '../space/runtime/task-agent-manager';
 import { setupSpaceWorkflowHandlers } from './space-workflow-handlers';
 import type { SpaceManager } from '../space/managers/space-manager';
 import { SpaceTaskManager } from '../space/managers/space-task-manager';
@@ -63,8 +68,7 @@ import { setupSpaceExportImportHandlers } from './space-export-import-handlers';
 import { provisionGlobalSpacesAgent } from '../space/provision-global-agent';
 import { setupGlobalSpacesHandlers } from './global-spaces-handlers';
 import type { GlobalSpacesState } from '../space/tools/global-spaces-tools';
-import { TaskAgentManager } from '../space/runtime/task-agent-manager';
-import { setupSpaceTaskSendMessageHandler } from './space-task-message-handlers';
+import { setupSpaceTaskMessageHandlers } from './space-task-message-handlers';
 
 export interface RPCHandlerDependencies {
 	messageHub: MessageHub;
@@ -252,7 +256,9 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	});
 	spaceRuntimeService.start();
 
-	// Task Agent Manager — manages Task Agent session lifecycle for space tasks
+	// Task Agent Manager — manages Task Agent session lifecycle and message injection.
+	// Must be created after spaceRuntimeService so it can get WorkflowExecutors via
+	// spaceRuntimeService.createOrGetRuntime(spaceId).
 	const taskAgentManager = new TaskAgentManager({
 		db: deps.db,
 		sessionManager: deps.sessionManager,
@@ -268,8 +274,8 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		defaultModel: deps.config.defaultModel,
 	});
 
-	// space.task.sendMessage — inject a message into a Task Agent session
-	setupSpaceTaskSendMessageHandler(deps.messageHub, taskAgentManager);
+	// Human ↔ Task Agent message routing handlers (require taskAgentManager)
+	setupSpaceTaskMessageHandlers(deps.messageHub, taskAgentManager, deps.db);
 
 	// Space export/import handlers
 	setupSpaceExportImportHandlers(
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -65,7 +65,6 @@ import { setupSpaceExportImportHandlers } from './space-export-import-handlers';
 import { provisionGlobalSpacesAgent } from '../space/provision-global-agent';
 import { setupGlobalSpacesHandlers } from './global-spaces-handlers';
 import type { GlobalSpacesState } from '../space/tools/global-spaces-tools';
-import { setupSpaceTaskMessageHandlers } from './space-task-message-handlers';
 
 export interface RPCHandlerDependencies {
 	messageHub: MessageHub;
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -44,11 +44,8 @@ import { TaskManager } from '../room/managers/task-manager';
 import { setupDialogHandlers } from './dialog-handlers';
 // Space handlers
 import { setupSpaceHandlers } from './space-handlers';
-import {
-	setupSpaceTaskHandlers,
-	setupSpaceTaskMessageHandlers,
-	type SpaceTaskManagerFactory,
-} from './space-task-handlers';
+import { setupSpaceTaskHandlers, type SpaceTaskManagerFactory } from './space-task-handlers';
+import { setupSpaceTaskMessageHandlers } from './space-task-message-handlers';
 import { TaskAgentManager } from '../space/runtime/task-agent-manager';
 import { setupSpaceWorkflowHandlers } from './space-workflow-handlers';
 import type { SpaceManager } from '../space/managers/space-manager';
--- a/packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-task-message-handlers.ts
@@ -1,39 +1,129 @@
 /**
  * Space Task Message RPC Handlers
  *
- * RPC handler for sending messages to Task Agent sessions:
- * - space.task.sendMessage — inject a human or agent message into a Task Agent session
+ * RPC handlers for human ↔ Task Agent message routing:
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Summary

Clean implementation of two new RPC handlers (`space.task.sendMessage` and `space.task.getMessages`) for human-to-Task-Agent message routing. The code is well-structured, follows existing codebase patterns, and has solid test coverage (19 tests).

### Findings

**P3 — No upper bound on `limit` parameter** (`packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, line 260)

The `getMessages` handler accepts an arbitrary `limit` from the client (`params.limit ?? 50`). There is no cap, so a caller could request `limit: 10000000` and load the entire message history in one call. Consider adding a maximum (e.g., `Math.min(params.limit ?? 50, 200)`) to prevent accidental or intentional resource exhaustion.

**P3 — Duplicate `SpaceTaskRepository` instantiation** (`packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, line 215)

`setupSpaceTaskMessageHandlers` creates its own `SpaceTaskRepository` via `new SpaceTaskRepository(db.getDatabase())`. Meanwhile, `index.ts` already creates `spaceTaskRepo` at line 200 and could pass it in. This is not a bug — `SpaceTaskRepository` is a lightweight stateless wrapper — but it creates an unnecessary second instance. Consider accepting `taskRepo` as a parameter instead, consistent with how other services receive their repo.

**P3 — `TaskAgentManagerInterface` lives in `space-task-handlers.ts`** (`packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, lines 35-40)

The `TaskAgentManagerInterface` is defined in the RPC handler file rather than alongside the `TaskAgentManager` class or in a shared types file. This works fine but could make discoverability harder as the codebase grows.

### Verdict

The implementation is correct, complete, and well-tested. All three findings are minor nits. **Recommendation: APPROVE.**

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 1 P1 issue (missing limit cap) and 2 P2 issues (missing spaceId scope check, missing message length validation) relative to established patterns in the same file.

---

### Summary

The implementation is functionally correct and the 19 tests all pass (`bun test`). TypeScript type check is clean. The two-phase validation pattern (task exists → session started → delegate to manager) is sound. The live-session-first / DB-fallback strategy in `getMessages` mirrors how `message-handlers.ts` and `state-manager.ts` handle the same pattern.

However, there are meaningful gaps versus the conventions already established in the same file (`space-task-handlers.ts`) and by the analogous `task.sendHumanMessage` handler in `task-handlers.ts`.

---

### P1 — No upper-bound cap on `limit` parameter in `space.task.getMessages`

**File:** `packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, line 260

```ts
const limit = params.limit ?? 50;
```

A caller can pass `{ taskId, limit: 1_000_000 }` and force a single query to load an unbounded number of SDK messages into memory. The analogous handler `task.getMessages` in `task-handlers.ts` (line 582) caps at 200:

```ts
const limit = Math.max(1, Math.min(params.limit ?? 50, 200));
```

The same cap (or a similar bound) should be applied here. Without it, a well-crafted or buggy client request can cause an OOM spike in the daemon.

---

### P2 — No message length validation in `space.task.sendMessage`

**File:** `packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, lines 224–226

The handler validates that the message is non-empty but imposes no upper length limit. The equivalent handler in `task-handlers.ts` (line 777) caps at 10,000 characters:

```ts
if (params.message.length > 10_000) {
    throw new Error('Message is too long (max 10,000 characters)');
}
```

Task Agent sessions sit on top of the same `AgentSession`/`MessageQueue` infrastructure, so an unreasonably large message payload would behave the same way — it should be capped consistently.

---

### P2 — No `spaceId` scope check (cross-space isolation gap)

**File:** `packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, lines 218–240 and 243–279

Every CRUD handler in `setupSpaceTaskHandlers()` requires a `spaceId` and uses `SpaceTaskManager.getTask()`, which enforces `task.spaceId === this.spaceId`. The new message handlers use `SpaceTaskRepository.getTask()` directly — which has no space-scope guard — and take no `spaceId` parameter.

In the current single-user context this has no material impact, but it is architecturally inconsistent with all other handlers in the same file. If `space.task.sendMessage` accepted a `spaceId` and checked `task.spaceId === params.spaceId`, it would match the cross-space isolation guarantee provided by every other mutation handler and make the task-ID-only access pattern an explicit, deliberate decision rather than an oversight.

The test file (`space-task-message-handlers.test.ts`) also has no test for the case where a `taskId` belonging to a different space is passed — because the current implementation would silently accept it.

---

### P3 — Mock fragility in tests: `prepare` returns the same stub for all SQL

**File:** `packages/daemon/tests/unit/rpc-handlers/space-task-message-handlers.test.ts`, lines 117–141

`createMockDatabase` returns a single `prepare → get` mock regardless of the SQL string. This works because only `getTask` is called in the current implementation, but any future code path that calls a second `prepare` (e.g., to update `task_agent_session_id` or audit something) would silently receive the same row stub. Keying the mock on the SQL string or switching to a real in-memory SQLite DB (as many other test files in this project do) would make the mock more robust.

---

### What is correct

- Validation order is sound: missing params → task not found → session not started → delegate.
- `injectTaskAgentMessage` is correctly guarded — if the agent is no longer live in memory, `TaskAgentManager` throws a clear error.
- The `TaskAgentManagerInterface` narrow interface correctly decouples the handlers from the concrete `TaskAgentManager` class, making mocking easy.
- Cursor parsing (`Number(cursor)` + `Number.isNaN` guard) is correct and consistent with the DB query expecting a numeric timestamp.
- Handler registration in `rpc-handlers/index.ts` is placed after `SpaceRuntimeService` is created, which is required and is correctly documented in a comment.
- All 19 tests pass; TypeScript reports no errors.

lsm (commented):

### Second Review Pass

All three issues raised in the first review have been properly addressed in commit `ee6a01d8`:

1. **P1 (limit cap):** `Math.max(1, Math.min(params.limit ?? 50, 200))` correctly bounds the limit between 1 and 200, handling negative values, zero, and excessively large values. Consistent with the existing pattern referenced in the comment.

2. **P2 (message length validation):** `sendMessage` now rejects messages exceeding 10,000 characters before hitting the TaskAgentManager. The length check is against the raw string (before trim), which is the correct behavior — a whitespace-padded message that is too long should still be rejected.

3. **P2 (spaceId scope check):** Both `sendMessage` and `getMessages` now require `spaceId`, validate it, and verify `task.spaceId !== params.spaceId` to enforce cross-space isolation. The error message intentionally uses the generic "Task not found" to avoid leaking existence of tasks across spaces.

### Code Quality

- **Handler structure** is clean and follows the same validation pattern as the existing CRUD handlers in the same file.
- **`TaskAgentManagerInterface`** is a well-scoped interface that decouples handlers from the concrete manager — good for testability.
- **`SpaceTaskRepository`** is instantiated once at setup time (line 215) and reused across both handlers, avoiding repeated allocations.
- **Tests:** 27 tests with thorough coverage of happy paths, parameter validation, cross-space isolation, error propagation, cursor parsing, limit clamping, live-session vs DB fallback, and hasMore pagination. The mock structure is clean and the `setup()` helper makes test variants easy to follow.

### Minor Observations (P3)

- The `taskRepo` constructed in `setupSpaceTaskMessageHandlers` (line 215) is a second `SpaceTaskRepository` instance — a separate one already exists in `setupRPCHandlers` (line 200, `spaceTaskRepo`). These are stateless wrappers around the same DB handle, so no functional issue, but it could be passed in to avoid the extra instantiation. Not worth changing now.

**Recommendation: APPROVE** — all prior feedback has been addressed correctly. Tests pass (27/27). No remaining issues.

lsm (commented):

**Recommendation: APPROVE** — All three previously flagged issues have been addressed correctly and the test suite has grown from 19 to 27 cases with targeted coverage for each fix.

---

### Verification of Prior Feedback

**P1 — Limit cap (was missing, now fixed)**

`packages/daemon/src/lib/rpc-handlers/space-task-handlers.ts`, line 281:

```ts
const limit = Math.max(1, Math.min(params.limit ?? 50, 200));
```

Correctly caps at 200 and floors at 1. Matches the pattern in `task-handlers.ts`. Two new tests exercise the upper-bound (1,000,000 → 200) and lower-bound (0 → 1) paths.

**P2 — Message length validation (was missing, now fixed)**

Lines 230-232: `params.message.length > 10_000` raises a clear error. The boundary cases (exactly 10,000 chars = allowed, 10,001 = rejected) are both covered by dedicated tests. The whitespace-only guard (`params.message.trim() === ''`) was already present and is unchanged.

**P2 — `spaceId` cross-space isolation check (was missing, now fixed)**

Both `sendMessage` (line 239) and `getMessages` (line 272) now compare `task.spaceId !== params.spaceId` and surface a generic `Task not found` error — the correct approach to avoid leaking task existence across space boundaries. Each handler has a dedicated cross-space isolation test case.

---

### Remaining Observations (no blocking issues)

**P3 — Duplicate `SpaceTaskRepository` instantiation still present**

`setupSpaceTaskMessageHandlers` creates its own `new SpaceTaskRepository(db.getDatabase())` at line 215. `index.ts` already holds `spaceTaskRepo` at line 200 and passes `deps.db` separately. This is a pre-existing low-priority nit from the first review; it is not blocking.

**P3 — `TaskAgentManagerInterface` still defined in the RPC handler file**

The interface lives at line 35 of `space-task-handlers.ts` rather than alongside `TaskAgentManager` or in a shared types file. No functional impact; discoverability nit only.

---

### Test Coverage Summary

27 tests covering: handler registration, happy paths (live session + DB fallback), limit clamping, cursor parsing (valid/invalid), default limit, `hasMore: true`, all required-field validations, task-not-found, cross-space isolation, session-not-started, and `TaskAgentManager` error propagation. Coverage is comprehensive for the public surface.

---

All previously raised concerns are resolved. Implementation is correct, complete, and consistent with established project patterns.

PR #538, opened 2026-03-21
feat: add Task Agent spawning to SpaceRuntime tick loop (Task 5.1)

lsm:
- Add optional `taskAgentManager` field to SpaceRuntimeConfig for backward compat
- In processRunTick(), detect pending tasks with no taskAgentSessionId and spawn via TaskAgentManager
- Skip advance() entirely when TaskAgentManager is configured — Task Agent drives the workflow
- Check liveness of active Task Agent sessions; skip tick when agent is alive
- Detect dead/crashed Task Agent sessions, reset task to pending, re-spawn on next tick
- Guard concurrent ticks via isSpawning() to prevent duplicate session creation
- Add 9 unit tests covering all new behaviors:
  spawn on pending, backward-compat direct-advance, active-agent skip, no-advance in TAM mode,
  crash recovery, concurrency guard, idempotent spawn, and next-step agent spawning

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -32,6 +32,7 @@ import type { SpaceAgentManager } from '../managers/space-agent-manager';
 import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
+import type { TaskAgentManager } from './task-agent-manager';
 import { SpaceTaskManager } from '../managers/space-task-manager';
 import { WorkflowExecutor, WorkflowTransitionError } from './workflow-executor';
 import { selectWorkflow } from './workflow-selector';
@@ -57,6 +58,15 @@ export interface SpaceRuntimeConfig {
 	workflowRunRepo: SpaceWorkflowRunRepository;
 	/** Task repository for querying tasks by run/step */
 	taskRepo: SpaceTaskRepository;
+	/**
+	 * Optional TaskAgentManager for Task Agent mode.
+	 *
+	 * When provided, SpaceRuntime spawns Task Agent sessions for pending tasks
+	 * instead of calling advance() directly. Task Agents drive the workflow
+	 * via their MCP tools. When absent, the original direct-advance behavior
+	 * is preserved for backward compatibility.
+	 */
+	taskAgentManager?: TaskAgentManager;
 	/**
 	 * Interval between executeTick() calls in milliseconds.
 	 * Used by start(). Default: 5000 (5 seconds).
@@ -563,6 +573,64 @@ export class SpaceRuntime {
 				}
 			}
 		}
+
+		// ─── Task Agent integration ───────────────────────────────────────────────
+		// When a TaskAgentManager is configured, Task Agents drive the workflow.
+		// SpaceRuntime's role here is lifecycle management only: spawn for pending
+		// tasks, check liveness, and recover from crashes. advance() is never called
+		// directly — the Task Agent calls it via the advance_workflow MCP tool.
+		if (this.config.taskAgentManager) {
+			const tam = this.config.taskAgentManager;
+
+			// Step 1: Check tasks that already have a Task Agent session (in_progress).
+			// If the agent is alive, skip this tick — it is driving the workflow.
+			// If the agent is gone (crashed/stopped), reset the task to pending so the
+			// next tick will spawn a fresh agent to resume from the current workflow state.
+			for (const task of stepTasks) {
+				if (!task.taskAgentSessionId) continue;
+
+				if (tam.isTaskAgentAlive(task.id)) {
+					return; // Task Agent is running — skip this tick entirely
+				}
+
+				// Task Agent session is gone — reset for re-spawn
+				log.warn(
+					`SpaceRuntime: task agent for task ${task.id} is gone ` +
+						`(session ${task.taskAgentSessionId}), resetting to pending for re-spawn`
+				);
+				this.config.taskRepo.updateTask(task.id, {
+					taskAgentSessionId: null,
+					status: 'pending',
+				});
+			}
+
+			// Step 2: Spawn Task Agents for pending tasks without an agent session.
+			// Re-read from DB to pick up status resets applied in Step 1.
+			const pendingTasksNeedingAgent = this.config.taskRepo
+				.listByWorkflowRun(runId)
+				.filter(
+					(t) =>
+						t.workflowStepId === currentStep.id && t.status === 'pending' && !t.taskAgentSessionId
+				);
+
+			if (pendingTasksNeedingAgent.length > 0) {
+				if (space) {
+					for (const task of pendingTasksNeedingAgent) {
+						if (tam.isSpawning(task.id)) continue; // spawn already in progress
+						try {
+							await tam.spawnTaskAgent(task, space, meta.workflow, run);
+							this.config.taskRepo.updateTask(task.id, { status: 'in_progress' });
+						} catch (err) {
+							log.error(`SpaceRuntime: failed to spawn task agent for task ${task.id}:`, err);
+						}
+					}
+				}
+			}
+
+			// Never call advance() directly — Task Agent drives the workflow.
+			return;
+		}
+		// ─────────────────────────────────────────────────────────────────────────
 
 		if (!stepTasks.every((t) => t.status === 'completed')) return;
 
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -583,17 +583,19 @@ export class SpaceRuntime {
 			const tam = this.config.taskAgentManager;
 
 			// Step 1: Check tasks that already have a Task Agent session (in_progress).
-			// If the agent is alive, skip this tick — it is driving the workflow.
-			// If the agent is gone (crashed/stopped), reset the task to pending so the
-			// next tick will spawn a fresh agent to resume from the current workflow state.
+			// Full loop always completes so that dead-agent resets are applied to ALL
+			// tasks before deciding whether to skip. Early-returning on the first alive
+			// agent would leave dead-agent tasks unrecovered if a step has multiple tasks.
+			let anyAgentAlive = false;
 			for (const task of stepTasks) {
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Recommendation: APPROVE

This PR cleanly integrates Task Agent spawning into the `SpaceRuntime.processRunTick()` method. The implementation is well-structured with clear separation of concerns and good backward compatibility.

### What works well

- **Backward compatibility**: The `taskAgentManager` field is optional in `SpaceRuntimeConfig`, and the existing `advance()` path is fully preserved when it is absent.
- **Crash recovery**: Dead Task Agent detection resets the task to `pending` and re-reads from DB before spawning, which correctly handles the stale `stepTasks` array.
- **Concurrency guard**: The `isSpawning()` check prevents duplicate spawns during overlapping ticks.
- **Error isolation**: `spawnTaskAgent()` failures are caught per-task with `try/catch`, so one failing task does not block others.
- **Test coverage**: 9 new tests cover the key scenarios: spawn, backward compat, liveness skip, no-advance in TAM mode, crash recovery, concurrency guard, idempotent spawn, and next-step spawning. All 47 tests pass.

### Minor observations (P3)

1. **Silent no-op when `space` is null** (`packages/daemon/src/lib/space/runtime/space-runtime.ts:617`): If `spaceManager.getSpace()` returns `null` (space deleted between ticks), pending tasks silently skip spawning and the function returns. This is acceptable defensive behavior but could benefit from a warning log so operators can diagnose stuck workflows. Low priority since this is an edge case that would also affect the existing timeout detection path.

2. **Early return on first alive task** (`packages/daemon/src/lib/space/runtime/space-runtime.ts:593`): When multiple tasks exist for the current step and the first one has an alive agent, the function returns immediately without checking whether other tasks in the same step also need attention (e.g., a second task whose agent crashed). In practice, the current workflow model creates one task per step, so this is not a real issue today, but worth noting if multi-task-per-step support is added later.

3. **Test mock type casting** (`packages/daemon/tests/unit/space/space-runtime.test.ts:1062`): `buildRuntimeWithMockTAM` uses `as never` to cast the mock TAM. This is a standard test pattern but slightly fragile -- if the `TaskAgentManager` interface gains required methods, the mock will not fail at compile time. Acceptable for test code.

Overall this is a clean, well-tested implementation that correctly achieves the stated goals.

lsm (commented):

All 47 tests pass and the core Task Agent integration works correctly. The backward-compat path, concurrency guard, and crash-recovery happy path are all solid. A few issues need addressing before merge.

---

### P1 — Silent failure when space is deleted mid-run

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 616-627

```ts
if (pendingTasksNeedingAgent.length > 0) {
    if (space) {  // ← silent no-op if space is null
        for (const task of pendingTasksNeedingAgent) {
```

When `space` is `null` (deleted while a run is in progress), pending tasks that need a Task Agent are silently dropped every tick. There is no log warning, no notification, and no status change. The tasks stay pending indefinitely while the run stays `in_progress`. This is different from how missing workflows are handled (those are explicitly logged in `rehydrateExecutors()`). At minimum a `log.warn` is needed here; ideally, the task should be moved to `needs_attention` or the run cancelled.

---

### P1 — Early return in liveness loop skips remaining dead-agent tasks

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 589-604

```ts
for (const task of stepTasks) {
    if (!task.taskAgentSessionId) continue;
    if (tam.isTaskAgentAlive(task.id)) {
        return; // skips entire function, including remaining dead agents
    }
    // reset dead agent...
}
```

If two tasks exist for the current step and the first has an alive agent while the second has a dead one, the `return` at line 593 prevents the dead-agent task from being reset and re-spawned on this tick. The scenario is currently unreachable (one task per step is the dominant pattern), but the data model allows multiple tasks per step and this ordering issue would cause silent misbehavior if that invariant changes. The fix is to complete the full loop first (collecting liveness results), then decide whether to skip or continue to Step 2.

---

### P1 — Missing test: silent no-op when space is null in TAM mode

There is no unit test covering the case where `spaceManager.getSpace()` returns `null` when `taskAgentManager` is configured and there are pending tasks. Given that the bug above exists precisely in this branch, a test is needed to document and enforce the expected behavior (whether that's a log-and-continue or an error).

---

### P2 — Implicit double-write contract between `spawnTaskAgent` and SpaceRuntime is undocumented

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 621-622

```ts
await tam.spawnTaskAgent(task, space, meta.workflow, run);
this.config.taskRepo.updateTask(task.id, { status: 'in_progress' });
```

`spawnTaskAgent` returns the session ID (`Promise<string>`) which is ignored. The real `TaskAgentManager.spawnTaskAgent` already writes `taskAgentSessionId` to the repo internally (line 281 of `task-agent-manager.ts`). SpaceRuntime only writes `status: 'in_progress'` after. This split responsibility is implicit: if `spawnTaskAgent` ever stops writing `taskAgentSessionId` (e.g., the concurrent-spawn polling path never writes it until the first spawn completes), the liveness check at Step 1 will silently not fire on subsequent ticks. A comment at line 621 explaining that `spawnTaskAgent` writes `taskAgentSessionId` as a side effect, and SpaceRuntime owns the `status` transition, would prevent future regressions.

---

### P2 — Test mock's `spawnTaskAgent` mimics real manager's side-effect, obscuring the split contract

**File:** `packages/daemon/tests/unit/space/space-runtime.test.ts`, lines 1050-1058

The default mock `spawnTaskAgent` sets `taskAgentSessionId` in the repo, mirroring what the real `TaskAgentManager` does. This is correct behavior but it isn't documented in the test. If someone changed the real manager to not write `taskAgentSessionId` internally, the alive-agent skip test (line 1119) would still pass because the mock does it. A brief comment in `makeMockTaskAgentManager` explaining that the mock mirrors the real manager's DB side-effect would make the contract explicit.

---

### P3 — Crash-recovery test uses substring match on internal mock detail

**File:** `packages/daemon/tests/unit/space/space-runtime.test.ts`, line 1219

```ts
expect(updated.taskAgentSessionId).toContain('v1');
```

An explicit equality check (`toBe(`session:${tasks[0].id}:v1`)`) would be more precise and safer.

---

### Overall

The core implementation is correct and well-structured. The 9 new TAM tests cover the required scenarios comprehensively. The main blocking issues are: a silent failure when the space is deleted (tasks silently stuck pending), an ordering hazard in the liveness loop when multiple tasks exist for a step, and an undocumented cross-component write contract.

lsm (commented):

### Recommendation: APPROVE

All five issues from the previous review have been properly addressed:

1. **P1 (fixed): Silent abandonment when space is null** -- Now logs a warning at `space-runtime.ts:625-627` explaining the space was not found and the run will remain stuck. Verified in test output.

2. **P1 (fixed): Early `return` in liveness loop** -- Changed from `return` to `continue` with an `anyAgentAlive` flag (`space-runtime.ts:589-610`). The full loop now completes before deciding to skip, so dead-agent resets apply to ALL tasks in the step.

3. **P1 (fixed): Missing test for null-space path** -- New test `'logs warning and skips spawn when space is null (space deleted mid-run)'` at `space-runtime.test.ts:1345` creates a runtime with a null-returning space manager and verifies spawn is skipped and task stays pending.

4. **P2 (fixed): Undocumented double-write contract** -- Comment added at `space-runtime.ts:632-633`: "spawnTaskAgent writes taskAgentSessionId to the DB as a side effect. SpaceRuntime owns the status transition to 'in_progress' after spawn." Matching documentation added to the mock factory's JSDoc at `space-runtime.test.ts:1038-1041`.

5. **P3 (fixed): Imprecise assertion** -- Changed from `toContain('v1')` to exact match `toBe(\`session:\${tasks[0].id}:v1\`)` at `space-runtime.test.ts:1230`.

### Additional observations on the fix commit

- The new `'liveness loop resets all dead-agent tasks before deciding to skip tick'` test (`space-runtime.test.ts:1397`) is a good addition that directly validates the multi-task-per-step dead-agent recovery scenario. It correctly sets up two tasks (one alive, one dead) and verifies the dead one is reset while the alive one is untouched.

- The `buildRuntimeWithMockTAM` helper now accepts an optional `overrideSpaceManager` parameter, which is a clean way to inject the null-space behavior without duplicating setup code.

- All 49 tests pass with no failures.

No new issues found. The implementation is clean and well-tested.

lsm (commented):

**Recommendation: APPROVE**

All five issues raised in the previous review have been correctly addressed in commit `c0d25be`.

---

**P1 — Silent abandonment of pending tasks when space is null** ✓

`log.warn` is now emitted at line 625 before skipping spawn:
```
log.warn(`SpaceRuntime: cannot spawn task agents for run ${runId} — space ${meta.spaceId} not found`)
```
The new test `'logs warning and skips spawn when space is null'` (line 1347) covers this path end-to-end, verifying both that `spawnCount === 0` and that the task remains `pending`.

**P1 — Early `return` in liveness loop prevented dead-agent recovery for multiple tasks** ✓

The loop now uses an `anyAgentAlive` flag (line 589) with `continue` on each alive task (line 595), then a single guard check at line 610 after the full loop completes. Dead-agent resets apply to all tasks before the skip decision is made.

The new test `'liveness loop resets all dead-agent tasks before deciding to skip tick'` (line 1391) explicitly exercises the two-task scenario (one alive, one dead) and asserts that the dead task is reset to `pending` even though the alive task would have caused an early `return` under the old code.

**P1 — Missing test for silent no-op path (space null)** ✓

Addressed together with the first P1 above. The test at line 1347 is well-structured: it builds a fresh runtime with a `nullSpaceManager` that returns `null` from `getSpace()`, calls `executeTick()`, and asserts `spawnCount === 0` with the task staying `pending`.

**P2 — Undocumented double-write contract between `spawnTaskAgent` and SpaceRuntime's `in_progress` update** ✓

The comment at lines 632-633 now clearly documents the split-write contract:
```
// spawnTaskAgent writes taskAgentSessionId to the DB as a side effect.
// SpaceRuntime owns the status transition to 'in_progress' after spawn.
```
The mock's comment at line 1060 also documents that the mock mirrors this real side-effect, preventing future maintainers from writing mocks that break the liveness check.

**P3 — Imprecise assertion in crash-recovery test** ✓

The assertion is now `expect(updated.taskAgentSessionId).toBe(`session:${tasks[0].id}:v1`)` (line 1230) — a precise equality check against the exact session ID value, plus `expect(callCount).toBe(1)` to verify exactly one re-spawn happened.

---

### No New Issues

The implementation is clean. The `anyAgentAlive` + `continue` pattern is idiomatic and clearly commented. The test coverage now includes all previously missing paths.

PR #540, opened 2026-03-21
feat: wire TaskAgentManager into SpaceRuntimeService (Task 5.2)

lsm:
- Add `taskAgentManager?` field to `SpaceRuntimeServiceConfig` so the
  manager can optionally be provided at construction time
- Add `setTaskAgentManager()` to `SpaceRuntime` (post-construction
  setter, mirrors `setNotificationSink()` pattern) — writes directly
  into the config so the tick loop picks it up immediately
- Add `setTaskAgentManager()` to `SpaceRuntimeService` that delegates
  to the underlying `SpaceRuntime`
- In `rpc-handlers/index.ts`: move `spaceRuntimeService.start()` to
  after `TaskAgentManager` is created, then inject the manager via
  `setTaskAgentManager()` before starting — resolves the circular
  dependency (SpaceRuntimeService created first so TaskAgentManager can
  reference it, then injected back via setter)
- 4 new unit tests verifying the wiring: method existence, no-throw,
  delegation to SpaceRuntime, and config-level propagation

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -241,7 +241,8 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		deps.daemonHub
 	);
 
-	// Space Runtime Service — wraps SpaceRuntime with per-space lifecycle API
+	// Space Runtime Service — wraps SpaceRuntime with per-space lifecycle API.
+	// Not started yet: TaskAgentManager is created next and injected before start().
 	const spaceRuntimeService = new SpaceRuntimeService({
 		db: deps.db.getDatabase(),
 		spaceManager: deps.spaceManager,
@@ -250,7 +251,6 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		workflowRunRepo: spaceWorkflowRunRepo,
 		taskRepo: spaceTaskRepo,
 	});
-	spaceRuntimeService.start();
 
 	// Task Agent Manager — manages Task Agent session lifecycle and message injection.
 	// Must be created after spaceRuntimeService so it can get WorkflowExecutors via
@@ -269,6 +269,13 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		getApiKey: () => deps.authManager.getCurrentApiKey(),
 		defaultModel: deps.config.defaultModel,
 	});
+
+	// Wire TaskAgentManager into the SpaceRuntime so the tick loop can spawn
+	// Task Agent sessions for pending tasks. Resolves circular dependency:
+	// SpaceRuntimeService → SpaceRuntime needed TaskAgentManager, which in turn
+	// needed SpaceRuntimeService. Both are now created; inject via setter.
+	spaceRuntimeService.setTaskAgentManager(taskAgentManager);
+	spaceRuntimeService.start();
 
 	// Human ↔ Task Agent message routing handlers (require taskAgentManager)
 	setupSpaceTaskMessageHandlers(deps.messageHub, taskAgentManager, deps.db);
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR correctly wires `TaskAgentManager` into `SpaceRuntimeService` and `SpaceRuntime`, resolving the circular dependency between the two. The implementation is clean, minimal, and follows the existing `setNotificationSink()` pattern precisely.

### What was reviewed

1. **`SpaceRuntime.setTaskAgentManager()`** (`packages/daemon/src/lib/space/runtime/space-runtime.ts:183-185`) — Writes directly to `this.config.taskAgentManager`, making it available to the tick loop immediately. This mirrors `setNotificationSink()` which writes to a separate field; the difference is acceptable since `taskAgentManager` was already an optional config field (added in Task 5.1).

2. **`SpaceRuntimeService.setTaskAgentManager()`** (`packages/daemon/src/lib/space/runtime/space-runtime-service.ts:61-63`) — Simple delegation to the runtime. Consistent with `setNotificationSink()` at line 119.

3. **`SpaceRuntimeServiceConfig.taskAgentManager?`** (`packages/daemon/src/lib/space/runtime/space-runtime-service.ts:40`) — Optional field allowing construction-time injection when the circular dependency isn't an issue. Good for testability.

4. **`rpc-handlers/index.ts`** (lines 244-278) — The `start()` call is correctly moved after `TaskAgentManager` construction and injection. The comment explaining the ordering is clear. This ensures no ticks fire before the manager is wired in.

5. **Tests** (`packages/daemon/tests/unit/space/space-runtime-service.test.ts:169-204`) — Four tests covering: method existence, no-throw on call, delegation verification (checking the runtime's config was mutated), and construction-time config propagation. The tests reach into private state via type casts, which is acceptable for verifying wiring in unit tests.

### No issues found

- The circular dependency resolution pattern (create A, create B with ref to A, inject B back into A) is well-documented in comments and correctly implemented.
- No race condition risk: `start()` is called only after injection, so the tick loop cannot fire before `taskAgentManager` is available.
- The `taskAgentManager` field on `SpaceRuntimeConfig` was already optional (from Task 5.1), so this PR doesn't change the config interface in `SpaceRuntime` — it only adds the setter and the service-level plumbing.
- Import statements use `type` imports where appropriate.

lsm (commented):

**Recommendation: APPROVE** (P3 nit only)

---

### Summary

The implementation correctly wires `TaskAgentManager` through `SpaceRuntimeService` to `SpaceRuntime` via a post-construction setter pattern. The circular dependency resolution — construct service first, construct manager second, inject back via setter, then start — is properly sequenced in `rpc-handlers/index.ts`. All 20 tests pass. TypeScript compilation is clean.

---

### What was verified

- **Sequencing in `rpc-handlers/index.ts`** (lines 246–278): `setTaskAgentManager` is called after both objects exist, and critically *before* `spaceRuntimeService.start()`. The tick loop therefore sees the manager from its very first execution.
- **No premature auto-start gap**: Between the service creation and the injection, nothing calls `createOrGetRuntime()` or `getSharedRuntime()`, so the tick loop cannot start without the manager being set.
- **Integration with existing tick logic** in `space-runtime.ts` (lines 619–647): the `config.taskAgentManager` check is the right read point and matches the mutation path from the setter.
- **Tests** cover method existence, no-throw invocation, delegation to `SpaceRuntime` (via private-field cast), and construction-time provision. All four cases are meaningful and pass.

---

### Issues

#### P3 — Design inconsistency: `setTaskAgentManager` mutates `config`, `setNotificationSink` uses a private field

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 141–172 vs 183–185

`setNotificationSink` stores its value in a dedicated `private notificationSink: NotificationSink` field (separate from `config`), while `setTaskAgentManager` writes directly into `this.config.taskAgentManager`. The PR comment says it mirrors `setNotificationSink()`, but the internal storage strategy is different.

This works correctly because `config` is `private` (not `private readonly`) and is read via `this.config.taskAgentManager` at tick time. It is purely a consistency nit — the two setters could either both use a dedicated private field or both mutate config. There is no runtime defect.

---

### No issues found in

- Correctness of the wiring
- Circular dependency resolution ordering
- TypeScript types (typecheck passes cleanly)
- Test quality and coverage for the new code path
- No security, data loss, or race condition concerns

PR #541, opened 2026-03-21
feat: add task completion notification to Space Agent (Task 6.2)

lsm:
- Extend DaemonEventMap with space.task.completed and space.task.failed
  event types, each with payload: taskId, spaceId, status, summary,
  workflowRunId, taskTitle
- In report_result handler, emit space.task.completed when status is
  'completed', or space.task.failed for 'needs_attention'/'cancelled'
- Add optional daemonHub field to TaskAgentToolsConfig; pass it from
  TaskAgentManager.createTaskAgentMcpServer()
- In provisionGlobalSpacesAgent, add optional daemonHub dep and subscribe
  to both events to inject notification messages into the spaces:global
  session via sessionFactory.injectMessage() with deliveryMode next_turn
- Pass daemonHub through from rpc-handlers/index.ts call site
- 11 new unit tests: 6 for report_result event emission (correct event
  names, payloads, and no-hub guard), 5 for provision subscription
  (subscribe calls, message content, error resilience)

### DIFF
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -436,6 +436,28 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		task: import('@neokai/shared').SpaceTask;
 	};
 
+	// Space Task Agent completion events (use 'global' as sessionId)
+	/** Emitted by report_result when a Task Agent marks a task as completed. */
+	'space.task.completed': {
+		sessionId: string;
+		taskId: string;
+		spaceId: string;
+		status: string;
+		summary: string;
+		workflowRunId: string;
+		taskTitle: string;
+	};
+	/** Emitted by report_result when a Task Agent marks a task as needs_attention or cancelled. */
+	'space.task.failed': {
+		sessionId: string;
+		taskId: string;
+		spaceId: string;
+		status: string;
+		summary: string;
+		workflowRunId: string;
+		taskTitle: string;
+	};
+
 	// Space workflow run events (global events - use 'global' as sessionId)
 	'space.workflowRun.created': {
 		sessionId: string;
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -338,6 +338,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 			workflowRunRepo: spaceWorkflowRunRepo,
 			db: deps.db.getDatabase(),
 			state: globalSpacesState,
+			daemonHub: deps.daemonHub,
 		}).catch((error) => {
 			log.error('Failed to provision global spaces agent:', error);
 		});
--- a/packages/daemon/src/lib/space/provision-global-agent.ts
+++ b/packages/daemon/src/lib/space/provision-global-agent.ts
@@ -27,6 +27,17 @@ import type { SessionFactory } from '../room/runtime/task-group-manager';
 const GLOBAL_SESSION_ID = 'spaces:global';
 const log = new Logger('global-spaces-agent');
 
+/**
+ * Unsubscribe handles for the task event subscriptions registered by the most recent
+ * `provisionGlobalSpacesAgent` call. Stored at module level so a second call (e.g. on
+ * daemon restart / error recovery) can clean up the previous listeners before re-subscribing,
+ * preventing duplicate notifications.
+ *
+ * Note: these subscriptions are intentionally daemon-lifetime; they are never unsubscribed
+ * during normal operation. The module-level store only exists to guard against double-init.
+ */
+let _taskEventUnsubs: Array<() => void> = [];
+
 export interface ProvisionGlobalSpacesAgentDeps {
 	sessionManager: SessionManager;
 	spaceManager: SpaceManager;
@@ -49,6 +60,10 @@ export interface ProvisionGlobalSpacesAgentDeps {
 	 * DaemonHub for subscribing to task completion/failure events emitted by Task Agents.
 	 * When provided, the global Space Agent session receives notification messages when tasks
 	 * complete or fail via `space.task.completed` / `space.task.failed` events.
+	 *
+	 * The subscription routes all space task events to the single `spaces:global` session,
+	 * which is intentional: the global agent manages tasks across all spaces. The notification
+	 * message includes the spaceId and taskId so the agent can act on specific tasks.
 	 */
 	daemonHub?: DaemonHub;
 }
@@ -136,9 +151,21 @@ export async function provisionGlobalSpacesAgent(
 	// Subscribe to task completion/failure events from Task Agents.
 	// When a task completes or fails, inject a notification message into the Space Agent session
 	// so it can take appropriate action (start next task, alert the user, etc.).
+	//
+	// All space task events are routed to the single spaces:global session — the global agent
+	// manages tasks across all spaces. Each message includes spaceId and taskId so the agent
+	// can reference tasks in follow-up tool calls (e.g. get_task_detail).
+	//
+	// Guard against double-init: clean up any previous subscriptions before re-subscribing.
 	if (daemonHub) {
-		daemonHub.on('space.task.completed', (event) => {
-			const message = `Task '${event.taskTitle}' has completed. Summary: ${event.summary}`;
+		_taskEventUnsubs.forEach((unsub) => unsub());
+		_taskEventUnsubs = [];
+
+		const unsubCompleted = daemonHub.on('space.task.completed', (event) => {
+			const summaryPart = event.summary ? ` Summary: ${event.summary}` : '';
+			const message =
+				`Task '${event.taskTitle}' (taskId: ${event.taskId}, spaceId: ${event.spaceId}) ` +
+				`has completed.${summaryPart}`;
 			void sessionFactory
 				.injectMessage(GLOBAL_SESSION_ID, message, { deliveryMode: 'next_turn' })
 				.catch((err) => {
@@ -148,9 +175,12 @@ export async function provisionGlobalSpacesAgent(
 				});
 		});
 
-		daemonHub.on('space.task.failed', (event) => {
+		const unsubFailed = daemonHub.on('space.task.failed', (event) => {
 			const statusLabel = event.status === 'cancelled' ? 'cancelled' : 'failed';
-			const message = `Task '${event.taskTitle}' has ${statusLabel}. Summary: ${event.summary}`;
+			const summaryPart = event.summary ? ` Summary: ${event.summary}` : '';
+			const message =
+				`Task '${event.taskTitle}' (taskId: ${event.taskId}, spaceId: ${event.spaceId}) ` +
+				`has ${statusLabel}.${summaryPart}`;
 			void sessionFactory
 				.injectMessage(GLOBAL_SESSION_ID, message, { deliveryMode: 'next_turn' })
 				.catch((err) => {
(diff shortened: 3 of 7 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured implementation of Task 6.2. The changes correctly wire up event emission from the `report_result` handler in Task Agent tools through DaemonHub to the Space Agent session via message injection. All 81 tests pass.

### What works well

- **Correct event flow**: `report_result` emits `space.task.completed` or `space.task.failed` on DaemonHub, and the provisioning code subscribes to inject notification messages into the `spaces:global` session with `deliveryMode: 'next_turn'`.
- **Error resilience**: Both the emit side (silent `.catch()` in `task-agent-tools.ts:651`) and the receive side (`.catch()` with `log.warn` in `provision-global-agent.ts:144,156`) handle failures gracefully without blocking.
- **Optional dependency**: `daemonHub` is optional in both `TaskAgentToolsConfig` and `ProvisionGlobalSpacesAgentDeps`, maintaining backward compatibility and testability.
- **Good test coverage**: 11 new tests cover happy paths for both event types, cancelled status labeling, the no-hub guard, the task-not-found guard, payload correctness (including `sessionId: 'global'`), and injection failure resilience.

### P3 observations (nits, non-blocking)

1. **Notification messages omit `taskId`/`spaceId`/`workflowRunId`** (`provision-global-agent.ts:141,153`): The injected plain-text messages only include `taskTitle` and `summary`. The Space Agent has no structured way to reference the task by ID for follow-up tool calls (e.g., `get_task_detail`). Consider including the `taskId` in the message so the agent can act on it programmatically. This is minor since the agent could look up tasks by title, but IDs are more reliable.

2. **Listener cleanup not captured** (`provision-global-agent.ts:140,151`): The return values from `daemonHub.on()` (unsubscribe functions) are discarded. Since `provisionGlobalSpacesAgent` is called once at startup and the subscriptions should live for the daemon lifetime, this is fine in practice. If re-provisioning ever becomes possible, this would leak listeners.

3. **DaemonEventMap `status` field is `string` rather than a union type** (`daemon-hub.ts:445,455`): The `status` field on both event types is typed as `string` rather than a narrower union like `'completed'` | `'needs_attention'` | `'cancelled'`. This is consistent with other events in the map but slightly less type-safe.

All observations are cosmetic or forward-looking -- nothing requires changes before merge.

lsm (commented):

---

### Summary

The core implementation is structurally sound: the event types are well-defined, the emit and subscribe wiring is correct, tests are present and cover the key paths. However there are a few meaningful issues that should be addressed before merge.

---

### P1 — Subscription leak on repeated `provisionGlobalSpacesAgent` calls

**File:** `packages/daemon/src/lib/space/provision-global-agent.ts`, lines 140–163

`daemonHub.on()` returns an unsubscribe function, but the return value is never stored or called. The comment at line 139 uses an unconditional `if (daemonHub)` block with no guard against running twice.

In the normal flow `provisionGlobalSpacesAgent` is only called once, but the code has an explicit daemon-restart path (lines 86–101). If the function were ever called a second time (e.g., due to error recovery or future refactoring), the hub accumulates duplicate handlers that each inject a duplicate notification message into the global session for every task event. The unsubscribe handles should be stored for cleanup.

The pattern used elsewhere in the codebase (e.g., `task-agent-manager.ts` lines 556–596) explicitly tracks `sessionListeners` and calls them on cleanup. This subscription should follow the same pattern.

---

### P2 — Notification message contains no spaceId context

**File:** `packages/daemon/src/lib/space/provision-global-agent.ts`, lines 141–143, 152–155

The injected message text is:
```
Task 'Implement login page' has completed. Summary: Feature implemented successfully.
```

This message omits the `spaceId` and `taskId`, both of which are available on the event payload. The Space Agent is a global agent that manages tasks across multiple spaces. Without the space and task identifiers in the notification, the agent cannot unambiguously route follow-up actions (e.g., `get_task_detail`, `start_workflow_run` on the correct space).

The `workflowRunId` is in the event payload but is also not included. The notification should include at minimum the `taskId` and `spaceId` so the agent can act without guessing.

---

### P2 — Subscription not gated by `spaceId` — all space events route to a single global session

**File:** `packages/daemon/src/lib/space/provision-global-agent.ts`, lines 140–161

The global subscription has no `spaceId` filter. In a multi-space deployment every `space.task.completed` and `space.task.failed` event from any space is injected into the single `spaces:global` session. While that may be intentional (the global agent handles all spaces), the absence of any space filter or routing logic means:

1. The injected message doesn't distinguish which space the task belongs to (see P2 above).
2. If future work introduces per-space agents, this handler will need to be revisited.

This is acceptable if the single-global-agent design is intentional; it should at least be documented as a conscious decision in a comment, and the message should include the spaceId (P2 above addresses this jointly).

---

### P2 — `void` + detached `catch` hides emit errors silently

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 651–653

```ts
void daemonHub.emit(eventName, eventPayload).catch(() => {
  // Non-fatal — event emission must not block or throw
});
```

Errors are silently swallowed. A task has already been transitioned to `completed`/`failed` at this point, and the Space Agent will never receive the notification. The comment says "non-fatal" which is reasonable for production resilience, but an error here means the Space Agent silently misses the event and cannot take next-step action. At minimum the `catch` handler should log via `Logger` (as the rest of the codebase does) so the failure is visible in daemon logs.

---

### P3 — `summary ?? ''` silently normalises a missing summary

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, line 646

`ReportResultSchema` makes `summary` optional. When it is absent, the event payload will include `summary: ''`, which propagates into the notification message as "Summary: " (empty). The Space Agent receives no useful signal. Consider either requiring `summary` in the schema for terminal statuses, or including it only when non-empty in the notification text.

---

### P3 — Test: `await new Promise((resolve) => setTimeout(resolve, 0))` is fragile

**File:** `packages/daemon/tests/unit/space/task-agent-tools.test.ts`, lines 1070, 1103, 1131, 1173

The tests flush the microtask queue with a zero-millisecond `setTimeout` to allow the `void emit` to settle. This is a brittle synchronisation pattern. Because `daemonHub.emit` in the mock is a synchronous function wrapped in `async`, the `void` promotion means the actual emission happens on the next microtask tick. The tests work today but would silently break if the mock or the emission strategy changed. Prefer making the mock hub synchronously record events on `emit` (or expose a flush helper) to make the flush explicit and deterministic.

---

### Correctness Verified

- Event names (`space.task.completed` / `space.task.failed`) match what is defined in `DaemonEventMap`.
- Routing `needs_attention` and `cancelled` both to `space.task.failed` is consistent between the emitter and the subscriber label logic.
- `daemonHub` is correctly threaded from `rpc-handlers/index.ts` line 341 through to `provisionGlobalSpacesAgent` and then to `createTaskAgentMcpServer` (line 269 of `task-agent-manager.ts`).
- The `deliveryMode: 'next_turn'` is appropriate for session notifications.
- Tests cover: event name correctness, payload field correctness, no-hub guard, subscribe calls, message content, error resilience. Coverage is good.

lsm (commented):

All previous feedback has been addressed. The implementation is correct and complete.

### What was reviewed

**Task**: When a Task Agent completes or fails via `report_result`, notify the Space Agent session.

**Implementation summary**:
1. New `DaemonEventMap` entries (`space.task.completed`, `space.task.failed`) in `daemon-hub.ts`
2. `report_result` in `task-agent-tools.ts` emits the appropriate event after status update
3. `provision-global-agent.ts` subscribes to both events and injects notification messages into `spaces:global` via `sessionFactory.injectMessage()`
4. `daemonHub` plumbed through `TaskAgentManager` and `rpc-handlers/index.ts`

### Verification of previous feedback fixes

- **Subscription leak on double-init**: Fixed. Module-level `_taskEventUnsubs` array stores unsub handles; cleaned up before re-subscribing. Test verifies previous unsubs are called on second provisioning.
- **Notification includes taskId/spaceId context**: Fixed. Messages now contain both IDs for downstream tool calls.
- **Silent error swallowing**: Fixed. Logger instance added; `log.warn()` on both emit failures and injection failures.
- **Fragile setTimeout in tests**: Fixed. Mock DaemonHub records events synchronously before returning `Promise.resolve()`.
- **Summary only appended when non-empty**: Fixed. Conditional `summaryPart` with empty-string check.

### Assessment

- **Correctness**: Event emission happens after `setTaskStatus` succeeds (inside the try block), which is the right ordering — no notification if the status update fails.
- **Error handling**: Both emit and inject errors are caught and logged, neither propagates to break the caller.
- **Event routing**: `sessionId: 'global'` in the payload follows the existing convention for global events (e.g., `space.workflowRun.created`).
- **Tests**: 11 new tests cover both sides (emit + subscribe), including edge cases: no hub provided, injection failure resilience, cancelled status label, double-init guard, and task-not-found guard.
- **Module-level state**: `_taskEventUnsubs` is module-scoped, which means it persists across test runs in the same process. This is acceptable here because the double-init test explicitly exercises and verifies this behavior, and the `forEach(unsub)` + `= []` pattern ensures clean state on each provisioning call.

No issues found. Recommendation: **APPROVE**.

lsm (commented):

**Recommendation: APPROVE**

All previously flagged P1/P2/P3 issues have been addressed. The implementation is correct, complete, and well-tested.

---

### What was verified

**P1 - Subscription leak (fixed)**
The `_taskEventUnsubs` array at module level correctly flushes previous listeners before re-subscribing. The double-init guard test in `provision-global-agent.test.ts` (lines 879-916) directly verifies the unsubscribe callcount after a second provisioning call.

**P2 - Notification message context (fixed)**
Both handlers now include `(taskId: ${event.taskId}, spaceId: ${event.spaceId})` in the injected message. Verified in tests that assert on those fields.

**P2 - Silent emit error swallowing (fixed)**
`task-agent-tools.ts` now imports `Logger` and creates `const log = new Logger('task-agent-tools')`. The `daemonHub.emit().catch()` handler calls `log.warn()` with a descriptive message. Same pattern used on the subscription side in `provision-global-agent.ts`.

**P3 - Fragile setTimeout in tests (fixed)**
`makeMockDaemonHub()` mock records events synchronously before returning `Promise.resolve()`, so assertions run immediately without any `await setTimeout` or microtask-flush tricks. Comment in test confirms this intent (line 1072-1074).

**P3 - Empty summary handling (fixed)**
`summaryPart` is only appended when `event.summary` is truthy (`event.summary ? ` Summary: ...` : ''`).

---

### Remaining observations (no issues)

**Module-level `_taskEventUnsubs` and test isolation**
The module-level variable persists across test cases within the same process. This is the documented and intentional behavior — the double-init guard test exercises it correctly. Because Bun reuses module singletons across tests in a file, the test order matters; the double-init test is correctly structured to start from whatever state the variable is in after the previous tests ran. No concern here.

**Missing test: `daemonHub.emit()` rejects but `report_result` still succeeds**
The implementation in `task-agent-tools.ts` (line 654) correctly uses `void ... .catch()` so a hub-emit rejection never bubbles up to the caller. However, there is no test exercising this path specifically. Given that the no-hub test (`does not emit events when daemonHub is not provided`) already covers the non-emission success path, and the emit-error path only affects logging, this is a minor gap that does not warrant blocking the PR.

**Event payload `status` field typing**
Both `space.task.completed` and `space.task.failed` define `status: string` in `DaemonEventMap`. Since `completed` can only appear on `space.task.completed` and `needs_attention`/`cancelled` only on `space.task.failed`, narrower literal union types would be more expressive — but this is consistent with the existing event-map style in the file and not a correctness issue.

---

All 82 tests pass. Typecheck and lint are clean. The feature is correctly implemented end-to-end: Task Agent emits the event, Space Agent subscribes and injects a message, with proper error isolation, guard against double-init, and thorough test coverage.

PR #547, opened 2026-03-21
test: add end-to-end online test for Task Agent lifecycle (Task 6.3)

lsm:
Tests cover the full Task Agent lifecycle end-to-end:
1. SpaceRuntime spawns a Task Agent session for a pending task
2. Task Agent processes its initial context message
3-5. Task Agent sessions accept probe messages referencing workflow tools
    (spawn_step_agent, check_step_status, advance_workflow)
6. Task Agent calls report_result to complete the task via full MCP dispatch
7. Space Agent receives completion notification when task completes

Supporting changes:
- SessionManager.registerSession(): registers externally-created AgentSessions
  (Task Agents created via AgentSession.fromInit()) in the session cache so RPC
  handlers return the live instance with MCP tools attached
- TaskAgentManager.spawnTaskAgent(): calls registerSession after session creation
- session-repository.ts: strip mcpServers and function values from config during
  JSON serialization to prevent cyclic structure errors (live MCP Server objects
  have circular refs that crash JSON.stringify)
- .devproxy/mocks.json: add probe mocks for Task Agent tool invocations; use
  text-only responses (no tool_use blocks) to prevent SDK retry loops when
  short tool names don't match registered MCP names

### DIFF
--- a/.devproxy/mocks.json
+++ b/.devproxy/mocks.json
@@ -1,6 +1,225 @@
 {
 	"mocks": [
 		{
+			"comment": "Task Agent lifecycle — follow-up after report_result tool execution. 'toolu_report_lifecycle_001' appears only in the follow-up call (as tool_use_id inside a tool_result block) — not in the initial probe call. ORDERING: must appear before the probe_task_agent_report_complete_001 mock.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "toolu_report_lifecycle_001"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (task-agent report_result follow-up)"
+					}
+				],
+				"body": {
+					"id": "msg_task_agent_report_followup_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED LIFECYCLE] Task completed successfully. The report_result tool has been executed and the task status has been updated."
+						}
+					],
+					"model": "claude-haiku-20250307",
+					"stop_reason": "end_turn",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 200,
+						"output_tokens": 25,
+						"cache_creation_input_tokens": 0,
+						"cache_read_input_tokens": 0,
+						"service_tier": "standard"
+					}
+				}
+			}
+		},
+		{
+			"comment": "Task Agent lifecycle — report_result probe. Returns the full MCP tool name 'mcp__task-agent__report_result' with stop_reason 'tool_use' so the SDK dispatches the tool. This updates the task status to 'completed' and emits space.task.completed. bodyFragment substring matches 'probe_task_agent_report_complete_001' in the user message injected by the test.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "probe_task_agent_report_complete_001"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (task-agent report_result probe)"
+					}
+				],
+				"body": {
+					"id": "msg_task_agent_report_probe_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED LIFECYCLE] I will now call report_result to mark this task as completed."
+						},
+						{
+							"type": "tool_use",
+							"id": "toolu_report_lifecycle_001",
+							"name": "mcp__task-agent__report_result",
+							"input": {
+								"status": "completed",
+								"summary": "Task completed successfully via probe_task_agent_report_complete_001 lifecycle test mock."
+							}
+						}
+					],
+					"model": "claude-haiku-20250307",
+					"stop_reason": "tool_use",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 300,
+						"output_tokens": 60,
+						"cache_creation_input_tokens": 0,
+						"cache_read_input_tokens": 0,
+						"service_tier": "standard"
+					}
+				}
+			}
+		},
+		{
+			"comment": "Task Agent lifecycle — spawn_step_agent probe. Returns text-only (no tool_use blocks) to prevent the Claude Agent SDK from trying to dispatch the short tool name 'spawn_step_agent' (registered as 'mcp__task-agent__spawn_step_agent'). Tool dispatch with a non-matching name causes the SDK to loop indefinitely. The text mentions spawn_step_agent and step-code-lifecycle-001 so tests can assert on text content.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "probe_task_agent_spawn_step_001"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (task-agent spawn_step_agent probe)"
+					}
+				],
+				"body": {
+					"id": "msg_task_agent_spawn_probe_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED LIFECYCLE] I will now call spawn_step_agent for step step-code-lifecycle-001 to create a sub-agent for the Code Implementation step."
+						}
+					],
+					"model": "claude-haiku-20250307",
+					"stop_reason": "end_turn",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 300,
+						"output_tokens": 40,
+						"cache_creation_input_tokens": 0,
+						"cache_read_input_tokens": 0,
+						"service_tier": "standard"
+					}
+				}
+			}
+		},
+		{
+			"comment": "Task Agent lifecycle — check_step_status probe. Returns text-only (no tool_use blocks) to prevent the Claude Agent SDK from trying to dispatch the short tool name 'check_step_status'. The text mentions check_step_status and step-code-lifecycle-001 so tests can assert on text content.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "probe_task_agent_check_step_001"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (task-agent check_step_status probe)"
+					}
+				],
+				"body": {
+					"id": "msg_task_agent_check_probe_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED LIFECYCLE] I will now call check_step_status to monitor the running step agent for step-code-lifecycle-001."
+						}
+					],
+					"model": "claude-haiku-20250307",
+					"stop_reason": "end_turn",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 300,
+						"output_tokens": 35,
+						"cache_creation_input_tokens": 0,
+						"cache_read_input_tokens": 0,
+						"service_tier": "standard"
+					}
+				}
+			}
+		},
+		{
+			"comment": "Task Agent lifecycle — advance_workflow probe. Returns text-only (no tool_use blocks) to prevent the Claude Agent SDK from trying to dispatch the short tool name 'advance_workflow'. The text mentions advance_workflow so tests can assert on text content.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "probe_task_agent_advance_wf_001"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (task-agent advance_workflow probe)"
+					}
+				],
+				"body": {
+					"id": "msg_task_agent_advance_probe_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED LIFECYCLE] I will now call advance_workflow to move the workflow from step-code-lifecycle-001 to the next step."
+						}
+					],
+					"model": "claude-haiku-20250307",
+					"stop_reason": "end_turn",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 300,
+						"output_tokens": 35,
+						"cache_creation_input_tokens": 0,
+						"cache_read_input_tokens": 0,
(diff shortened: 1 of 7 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured end-to-end online test for the Task Agent lifecycle, with thoughtful supporting changes to production code. The test file is thorough, well-documented, and the probe mock strategy is clearly explained.

### Production changes review

**`SessionManager.registerSession()`** (`packages/daemon/src/lib/session/session-manager.ts:246-248`): Clean, minimal addition. Delegates to `SessionCache.set()` which already exists. The doc comment clearly explains the motivation (preventing duplicate AgentSession creation with competing DaemonHub subscriptions). No issues.

**`TaskAgentManager.spawnTaskAgent()`** (`packages/daemon/src/lib/space/runtime/task-agent-manager.ts:287-291, 349-351`): `registerSession()` is called at the right point -- after `setRuntimeMcpServers()` so the cached instance has MCP tools attached, and before `startStreamingQuery()` so any concurrent RPC calls during streaming will get the live instance. Same pattern applied to sub-sessions. Correct.

**`session-repository.ts` config serialization** (`packages/daemon/src/storage/repositories/session-repository.ts:129-144`): The replacer now strips `mcpServers` by key name and function values. This fixes the circular reference crash when `updateSession` is called after `setRuntimeMcpServers()` has set live MCP Server objects on the in-memory config. The try/catch wrapper provides a clear error message. Note: `createSession()` (line 32) still uses plain `JSON.stringify(session.config)` without the replacer, but this is safe because `setRuntimeMcpServers()` is called *after* `fromInit()` persists to DB.

### Test review

The 7 tests cover the lifecycle progressively: spawn -> initial message -> tool probes (spawn_step_agent, check_step_status, advance_workflow) -> report_result execution -> Space Agent notification. The probe mock strategy is well-documented in the file header and per-test comments.

The mock ordering in `.devproxy/mocks.json` is correct: the `toolu_report_lifecycle_001` follow-up mock precedes the `probe_task_agent_report_complete_001` probe mock, ensuring the follow-up request (containing the tool_use_id) matches first rather than re-matching the original probe.

CI coverage is automatic: the `space` module in the CI matrix uses `test_path: tests/online/space` as a directory path, so the new file is auto-discovered.

### Minor observations (P3)

1. **Helper duplication** (`getAssistantMessages`, `extractTextContent`, `extractToolUses`): These are duplicated between `task-agent-lifecycle.test.ts` and `space-agent-coordination.test.ts`. Could be extracted to a shared test helper. Not blocking -- inline helpers in test files are a common and acceptable pattern.

2. **`STEP_REVIEW_ID` constant** (line 62): Defined but only used in the workflow creation fixture, not directly asserted in any test. This is fine since it is needed for the workflow structure, but worth noting.

3. **Test 7 `.catch()` for `waitForSdkMessages`** (line 636): The `.catch(() => ({ sdkMessages: [] }))` silently swallows errors when snapshotting the Space Agent's message count before triggering completion. This is intentional -- the Space Agent may not have any messages yet -- but it means a transient RPC error would also be masked. Acceptable for a "before" snapshot.

Overall: Clean, well-documented implementation with correct integration points. The production changes are minimal, well-scoped, and solve a real problem (duplicate AgentSession creation). The test coverage is thorough for the Task Agent lifecycle.

lsm (commented):

---

**Recommendation: REQUEST_CHANGES** (one P1, two P2s)

---

## Summary

This PR delivers a well-structured end-to-end online test suite for the Task Agent lifecycle plus three tight supporting production fixes. The architecture is solid, the mocking strategy is carefully documented, and the CI integration is correct. There are no blocking bugs, but a couple of things need attention before this can be considered fully reliable.

---

## P1 — Test assertion is too weak on Tests 3, 4, 5, 6

**File:** `packages/daemon/tests/online/space/task-agent-lifecycle.test.ts`, lines 411-413, 461-463, 510-512, 561-563

The comment on line 412 says:

```ts
minCount: 3, // initial user + initial assistant + probe user + probe assistant
```

That comment describes **4** messages, but `minCount` is **3**. If only the first 3 messages appear (initial user + initial assistant + probe user) the `waitForSdkMessages` call resolves successfully, and then `extractTextContent(assistantMsgs)` will only see the first assistant response (from the initial context message), which says `"[MOCKED] Hello! I'm Claude…"` — not the probe response. The subsequent `expect(textContent).toContain('spawn_step_agent')` would fail, but the failure message would be misleading (it would look like the probe was never matched, when really the test just didn't wait long enough).

More critically: if the initial context message's mocked response *happens* to contain the word `spawn_step_agent` in some future mock update, these tests would pass falsely even when the probe mock is never matched.

**Fix:** Change `minCount: 3` to `minCount: 4` in Tests 3, 4, 5 (4 messages: initial user, initial assistant, probe user, probe assistant). Test 6 involves MCP tool dispatch, so after the probe + tool dispatch there are at least 5 messages (initial user, initial assistant, probe user, probe assistant with tool_use, tool_result+follow-up assistant). `minCount: 4` (or 5) is more appropriate there too.

---

## P2 — `createSession` does not apply the `mcpServers` replacer

**File:** `packages/daemon/src/storage/repositories/session-repository.ts`, line 32

The fix correctly adds a JSON replacer to `updateSession` to strip `mcpServers` and function values. However, `createSession` on line 32 still calls `JSON.stringify(session.config)` without any replacer.

```ts
JSON.stringify(session.config),  // line 32 — no replacer
```

In practice this is safe today because `AgentSession.createSessionFromInit` deliberately excludes `mcpServers` from the config it passes to `db.createSession()` (confirmed in `agent-session.ts:482-484`). The comment there says it is intentionally omitted. But this is an implicit invariant that is easy to break. If any future caller passes a `Session` with `mcpServers` populated to `createSession`, it would crash with a cyclic reference error — the same bug that was just fixed in `updateSession`.

**Fix:** Apply the same replacer (or a shared helper) to `createSession` for defensive serialization. Alternatively, add a test assertion that `createSession` throws or strips if mcpServers is present.

---

## P2 — `registerSession` can silently overwrite a concurrently-loading session

**File:** `packages/daemon/src/lib/session/session-cache.ts`, line 120

`registerSession` delegates to `sessionCache.set()`, which writes directly into `this.sessions` without checking `sessionLoadLocks`. If `getAsync` was called on the same session ID between the moment the session is persisted to DB (`db.createSession`) and when `registerSession` is called, the concurrent `getAsync` will load from DB, creating a **second** `AgentSession` instance with its own DaemonHub subscriptions and no MCP tools. When `getAsync` completes, it calls `this.sessions.set(sessionId, agentSession)` (line 95 of session-cache.ts), overwriting the registered live instance with the DB-loaded duplicate.

The practical risk is low because the Task Agent session ID is not exposed to callers until `spawnTaskAgent` returns (i.e., after `registerSession` completes). But this is a silent data-race in the cache that could resurface if the ordering ever changes.

**Fix:** In `SessionCache.set()`, also cancel or guard against any in-flight `sessionLoadLocks` for the same ID; or document the invariant that callers must not have any concurrent `getAsync` calls for the session before calling `registerSession`.

---

## Minor observations (P3)

**Test isolation / cleanup:** `daemon.trackSession(GLOBAL_SPACES_SESSION_ID)` in Test 7 (line 610) will cause `afterEach` cleanup to call `session.delete` on `spaces:global`. Since each test gets a fresh daemon, this is harmless, but tracking a system-managed session feels surprising. A comment explaining the intent would help future readers.

**Mock probe ID reuse across tests:** Tests 6 and 7 both inject the same probe string `probe_task_agent_report_complete_001`. Since each test has a fresh daemon, there is no state bleed — but if the daemon is ever reused (e.g., a future `beforeAll` refactor), the same probe would match both tests with the same mock. Adding unique probe IDs per test would make this more robust.

**`STEP_CODE_ID` / `STEP_REVIEW_ID` constants (line 61-62):** The IDs are pre-assigned to make dev-proxy mocks deterministic. This is a reasonable pragmatic choice, but the constraint (IDs must match the mock JSON) is invisible from the workflow creation call. A co-located comment in `createTestFixtures` pointing at the mocks.json dependency would prevent confusion.

**Test 2 assertion breadth (line 361):** The regex `/workflow|step|task/i` is satisfied by almost any task-agent system prompt — it cannot distinguish between a proper context message and any generic text. Consider asserting on the specific task ID or space ID being present.

---

## What is done well

- The probe-mock pattern with `bodyFragment` matching is clearly explained and the ordering constraint (follow-up mock before probe mock) is correctly placed and documented.
- `TaskAgentManager.spawnTaskAgent` registers the session *before* `startStreamingQuery` is called, so the first query message can be intercepted by RPC handlers using the live instance — the ordering is correct.
- The CI matrix already covers the `space` directory with `mock_sdk: true` and `enable_spaces_agent: true`, so the new tests will run in CI correctly.
- The `session-repository.ts` fix correctly protects `updateSession` which is the hot path for config updates during an active session (where MCP servers are most likely to be present).
- Using text-only mocks for spawn/check/advance tools to avoid the SDK dispatch loop is a pragmatic and well-documented workaround for the MCP name mismatch constraint.

lsm (commented):

**Recommendation: APPROVE**

This re-review confirms that all three issues raised in the previous review have been correctly addressed.

---

## Fix 1 — P2 resolved: `createSession` now applies the `mcpServers` replacer

**File:** `packages/daemon/src/storage/repositories/session-repository.ts`, lines 32-36

The replacer is now applied identically in `createSession` and `updateSession`:

```ts
JSON.stringify(session.config, (key, val) => {
    if (key === 'mcpServers') return undefined;
    if (typeof val === 'function') return undefined;
    return val;
}),
```

This eliminates the latent crash risk if any future caller passes a `Session` with `mcpServers` populated to `createSession`. Correctly fixed.

---

## Fix 2 — P1 resolved: `minCount` corrected to 4 in Tests 3, 4, 5, 6

**File:** `packages/daemon/tests/online/space/task-agent-lifecycle.test.ts`, lines 412, 462, 511, 562

All four affected `waitForSdkMessages` calls now use `minCount: 4` (initial user + initial assistant + probe user + probe assistant), matching the inline comment. The assertion is now reliable and cannot pass falsely on 3 messages.

---

## Fix 3 — P2 resolved: `SessionCache` race condition guarded

**File:** `packages/daemon/src/lib/session/session-cache.ts`

Two complementary guards were added:

1. **`getAsync`** (lines 99-101): after loading from DB, checks `!this.sessions.has(sessionId)` before writing to the map. If `set()` was called (via `registerSession`) while the load was in flight, the registered live instance is preserved.

2. **`set()`** (line 135): calls `this.sessionLoadLocks.delete(sessionId)` so new `getAsync` callers that arrive after registration short-circuit immediately to the in-memory check rather than queuing behind the now-irrelevant DB load.

The interaction between the two guards is correct: callers already awaiting the lock still complete, hit the guard in `getAsync`, and return the registered instance. The double-delete in the `finally` block (line 108) when `set()` already removed the lock is a harmless no-op. Correctly fixed.

---

## No new issues introduced

The three changes are minimal, well-scoped, and do not introduce any new concerns. The existing P3 observations (helper duplication across test files, `STEP_REVIEW_ID` only used in fixture, `catch` swallowing in Test 7 snapshot) remain cosmetic and non-blocking.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The three fixes from review feedback are well-implemented:

1. **`mcpServers`/function JSON replacer in `createSession()`** (`packages/daemon/src/storage/repositories/session-repository.ts:32-36`): Now matches the existing pattern in `updateSession()` (line 139-142). This correctly prevents serialization failures when `mcpServers` contains non-serializable MCP server objects with functions. Good catch -- this was a latent crash waiting to happen on any session creation with MCP servers attached.

2. **`minCount: 3` -> `minCount: 4`** in tests 3-6 (`packages/daemon/tests/online/space/task-agent-lifecycle.test.ts`): The comment already described 4 messages (initial user + initial assistant + probe user + probe assistant), but the code was only waiting for 3. Fixed to match the documented expectation. This makes the tests more robust by ensuring all expected messages are present before assertions.

3. **`SessionCache.set()` race condition fix** (`packages/daemon/src/lib/session/session-cache.ts`):
   - `set()` now clears in-flight load locks (line 135), preventing new `getAsync()` callers from waiting on a stale DB load when a live session is already registered.
   - `getAsync()` guards against overwriting a registered session (line 99), preferring the live instance with MCP tools attached.
   - Returns `this.sessions.get(sessionId)` instead of the raw `agentSession` (line 103), ensuring callers always get the authoritative instance.

### Minor Observations (P3)

- **`getAsync()` second-caller path** (line 84-85): When a second caller joins via `loadInProgress`, it awaits the raw promise and gets the DB-loaded instance directly, bypassing the guard on line 99. If `set()` was called during the load, this second caller receives a stale reference. However, the session map itself is correct, and callers typically don't cache the returned reference long-term, so this is not a practical concern. Mentioning it only for completeness.

All 7 tests cover the full Task Agent lifecycle as described in the task. The test structure is thorough with proper cleanup, timeouts, and probe-based mock verification. The documentation comments in the test file are excellent -- the explanation of why text-only mocks are used instead of `tool_use` blocks (to avoid SDK retry loops with short tool names) is particularly valuable.

lsm (commented):

### Re-review: `registerSession` stub fix

The latest commit (`5303f73c`) correctly fixes the CI failure by adding a no-op `registerSession` stub to `mockSessionManager` in `packages/daemon/tests/unit/space/task-agent-manager.test.ts` (line 293-295).

**Why this was needed:** The production `TaskAgentManager.spawnTaskAgent()` calls `this.config.sessionManager.registerSession(agentSession)` at lines 291 and 353 of `task-agent-manager.ts`. The unit test's mock was missing this method, causing a `TypeError` across all 64 TaskAgentManager unit tests.

**Assessment of the fix:**
- The stub signature matches the production `SessionManager.registerSession(agentSession: AgentSession): void` signature
- The no-op behavior is appropriate -- unit tests verify spawn/cleanup logic, not cache registration
- The `_agentSession: unknown` parameter type is acceptable since the mock is cast via `as unknown as SessionManager`
- The comment explaining why it's a no-op is helpful

### Overall PR state

CI is green for lint/typecheck and most online test modules. The `space` online test module and unit tests are still pending but the fix is straightforward and correct.

The online test file (`task-agent-lifecycle.test.ts`) and the unit test file both look well-structured. No remaining issues identified.

**Recommendation: APPROVE** -- the fix is minimal, correct, and targeted.

lsm (commented):

**Recommendation: APPROVE**

### Re-review: `registerSession` stub fix (commit `5303f73c`)

The fix is correct, minimal, and targeted. Here is the full analysis.

#### What was broken

`TaskAgentManager.spawnTaskAgent()` (line 291) and `createSubSession()` (line 353) both call `this.config.sessionManager.registerSession(agentSession)`. This method was added to `SessionManager` in this PR as a supporting production change. The unit test's `mockSessionManager` object was missing this method, causing `TypeError: this.config.sessionManager.registerSession is not a function` in all 64 unit tests that exercise those code paths.

#### Assessment of the fix

The single-commit fix adds the stub at `packages/daemon/tests/unit/space/task-agent-manager.test.ts` lines 293-295:

```ts
registerSession: (_agentSession: unknown) => {
    // no-op: unit tests don't exercise cache registration
},
```

- **Correct placement**: The stub is added to `mockSessionManager` alongside the existing `deleteSession` stub. The mock now covers all three `sessionManager` calls made by `TaskAgentManager` in production: `registerSession` (x2) and `deleteSession` (x1).
- **Correct behavior**: A no-op stub is appropriate here. Unit tests verify spawn/cleanup/concurrency logic, not session cache registration. The cache is not exercised in the unit test environment.
- **Correct type**: `_agentSession: unknown` is acceptable since the mock is cast via `as unknown as SessionManager` at line 320 anyway.
- **Comment**: The comment explains the no-op rationale, which is helpful for future readers.

#### Overall PR state

All three fixes from the second iteration remain correct:
1. `session-cache.ts` guard against concurrent DB load overwriting a `registerSession` call — correct, reduces the race window for new callers.
2. `session-repository.ts` defensive `mcpServers` stripping in `createSession()` — correct, consistent with `updateSession`.
3. `task-agent-lifecycle.test.ts` `minCount: 4` fix — correct, ensures the probe assistant message is present before extracting text content.

The one subtle nuance in `session-cache.ts` is that concurrent callers already awaiting the load promise at line 85 (`return await loadInProgress`) will receive the DB-loaded session object (not the registered live instance) because they hold a direct promise reference that bypasses the in-memory map check. `set()` deleting the lock only helps new callers that arrive after `registerSession`. This is a pre-existing concern that predates this PR, and the new guard on line 99 correctly handles the load-initiator path. Not blocking.

The CI fix is solid. The test suite structure is sound. Approving.

PR #548, opened 2026-03-21
feat: add direct approve/reject to Inbox without TaskView navigation (Task 4.3)

akz142857:
- Add approveTask/rejectTask methods to inboxStore (calls room.task.approve/reject RPC, refreshes on success, shows toast.error on failure)
- Replace "Review" button in InboxTaskCard with Approve (green) + Reject (red outline) + View (ghost) action row
- Reject button expands an inline textarea form; submitting calls rejectTask and collapses
- Loading state (approvingId) disables all buttons while an approve is in flight
- Update Inbox header subtitle to include hint text "approve or reject below"
- Add tests for approveTask and rejectTask (success and failure paths)

### DIFF
--- a/packages/web/src/components/inbox/Inbox.tsx
+++ b/packages/web/src/components/inbox/Inbox.tsx
@@ -144,9 +144,7 @@ export function Inbox() {
 		<div class="flex flex-col h-full">
 			<div class="px-6 py-4 border-b border-dark-700 flex items-center justify-between">
 				<h1 class="text-lg font-semibold text-gray-100">Inbox</h1>
-				<span class="text-xs text-gray-500">
-					{count} awaiting review · approve or reject below
-				</span>
+				<span class="text-xs text-gray-500">{count} awaiting review · approve or reject below</span>
 			</div>
 			<div class="flex-1 overflow-y-auto">
 				{isLoading && (
--- a/packages/web/src/components/inbox/Inbox.tsx
+++ b/packages/web/src/components/inbox/Inbox.tsx
@@ -32,17 +32,19 @@ function InboxTaskCard({
 
 	const handleApprove = async () => {
 		onStartApprove(item.task.id);
-		await inboxStore.approveTask(item.task.id, item.roomId);
-		toast.approved();
+		const ok = await inboxStore.approveTask(item.task.id, item.roomId);
+		if (ok) toast.approved();
 		onFinishApprove();
 	};
 
 	const handleRejectSubmit = async () => {
 		if (!feedback.trim()) return;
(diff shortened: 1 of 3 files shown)


akz142857 (commented):
## Review

### P2 — Toast fires on failure for both Approve and Reject

**approve flow** (`Inbox.tsx`, `handleApprove`):

```tsx
const handleApprove = async () => {
  onStartApprove(item.task.id);
  await inboxStore.approveTask(item.task.id, item.roomId);
  toast.approved();   // always called, even when approveTask() failed
  onFinishApprove();
};
```

`approveTask()` catches all errors internally and **never rethrows**, so from the caller's perspective it always resolves. On an API failure the user sees both `toast.error("Server error")` and `toast.approved("Task approved")` — contradictory toasts.

**reject flow** — same issue:

```tsx
const handleRejectSubmit = async () => {
  await inboxStore.rejectTask(item.task.id, item.roomId, feedback.trim());
  toast.rejected();   // always called, even when rejectTask() failed
  setFeedback('');
  setIsRejecting(false);  // form collapses even on failure
};
```

On reject failure: `toast.error` + `toast.rejected` both fire, and the form collapses (hiding the user's typed feedback).

**Fix** — make the store methods return a success boolean:

```ts
// inbox-store.ts
async function approveTask(taskId: string, roomId: string): Promise<boolean> {
  try {
    const hub = await connectionManager.getHub();
    await hub.request('room.task.approve', { roomId, taskId });
    await refresh();
    return true;
  } catch (err) {
    toast.error(err instanceof Error ? err.message : 'Failed to approve task');
    return false;
  }
}

async function rejectTask(taskId: string, roomId: string, feedback: string): Promise<boolean> {
  try {
    const hub = await connectionManager.getHub();
    await hub.request('room.task.reject', { roomId, taskId, feedback });
    await refresh();
    return true;
  } catch (err) {
    toast.error(err instanceof Error ? err.message : 'Failed to reject task');
    return false;
  }
}
```

Then in the component:

```tsx
const handleApprove = async () => {
  onStartApprove(item.task.id);
  const ok = await inboxStore.approveTask(item.task.id, item.roomId);
  if (ok) toast.approved();
  onFinishApprove();
};

const handleRejectSubmit = async () => {
  if (!feedback.trim()) return;
  const ok = await inboxStore.rejectTask(item.task.id, item.roomId, feedback.trim());
  if (ok) {
    toast.rejected();
    setFeedback('');
    setIsRejecting(false);
  }
};
```

Update the test assertions to expect the return value from the mock.

---

Everything else is solid: UI layout, loading state, disabled-buttons, inline reject form, and test coverage are all good.

PR #552, opened 2026-03-21
docs: plan for workflow iteration loop with agent-driven cyclic transitions

lsm:
## Summary

- Adds a structured implementation plan for enabling verify-and-iterate patterns in Space workflows
- Covers 5 milestones: `task_result` condition type, iteration tracking, `goalId` on SpaceTask, updated seeded Coding Workflow, and comprehensive tests
- 13 total tasks across milestones, with clear dependency ordering

## Plan Structure

- `00-overview.md` -- Goal, approach, milestone list, dependencies
- `01-task-result-condition-type.md` -- New condition type for LLM-driven transitions
- `02-iteration-tracking.md` -- Cycle detection and maxIterations safety cap
- `03-goal-id-on-space-task.md` -- Goal association for cross-run task queries
- `04-update-seeded-coding-workflow.md` -- Verify step in built-in Coding Workflow
- `05-tests.md` -- Unit and integration test coverage

## Test plan

- [ ] Review plan for completeness and accuracy against codebase
- [ ] Verify milestone dependencies are correct
- [ ] Confirm task granularity is appropriate for single coder agent sessions


lsm (commented):

**Recommendation: REQUEST_CHANGES** — Several correctness issues and gaps require attention before this plan is handed to a coder agent.

---

## Overall Assessment

The plan has a solid conceptual structure and correctly identifies the key files and integration points. Task scoping is generally good — each task targets a single cohesive concern and can be implemented in one PR. The dependency graph is also mostly correct. However, there are factual errors against the current codebase and several important design gaps that would cause the coder to make wrong choices or produce broken code.

---

## P0 — Blocking Issues

### 1. Wrong migration numbers (Tasks 2.1 and 3.1)

The plan instructs the coder to add migrations 34 and 35. But the current codebase already has migrations through 33, and the comment in `migrations.ts` at the top states:

> "Migration 29 is the single consolidated migration for all Space system tables — do not add separate Space migrations after it."

This is a critical constraint the plan ignores entirely. New columns for `space_workflow_runs` and `space_tasks` must be added inside migration 29's idempotent block (using `IF NOT EXISTS` guards or a try/catch pattern like the other ALTER TABLE migrations), not as new numbered migrations 34/35/36. If the coder follows the plan literally, the migration will be applied to the wrong place and will likely break on fresh databases that already create these tables via migration 29.

**Files:** `packages/daemon/src/storage/schema/migrations.ts` (lines 1–10, and the `runMigration29` function body which is not shown but sets up all space tables)

### 2. `advance_workflow` already ignores `step_result` — the plan's Task 1.3 description contradicts the existing code comment

In `task-agent-tools.ts` at line 508–510, there is already an explicit comment:

> "Note: AdvanceWorkflowInput includes a `step_result` field which is currently a placeholder. WorkflowExecutor.advance() does not yet accept a result parameter... The field is retained in the schema for forward compatibility but is not forwarded to the executor in this implementation."

The handler signature is `async advance_workflow(_args: AdvanceWorkflowInput)` — the args are entirely unused with an underscore prefix. Task 1.3's instruction to forward `step_result` through is correct work to do, but the plan describes a non-existent "override vs. DB result" precedence rule without accounting for the fact that the current `advance()` signature takes zero arguments. The coder will need to change `advance()`'s signature as part of Task 1.2, not just "add an options parameter" as an afterthought in 1.3. The plan treats this as two loosely coupled changes when they require tight coordination.

---

## P1 — Should Fix

### 3. Cycle detection logic is fragile and semantically wrong (Task 2.2)

The plan proposes detecting cycles by checking "if any tasks exist for the target step in this run." This will silently break the iteration count when multiple parallel paths reach the same step legitimately, and it is also incorrect for the starting step: in a loop A -> B -> A, when the workflow first visits A, no tasks exist yet — so the initial visit to A correctly doesn't increment. But on the second visit to A (a real cycle), tasks for A do exist, so the increment fires. However, the plan conflates "visited before" with "is a cycle." In a non-cyclic DAG where a step is reachable via two paths, the second path would incorrectly be counted as a cycle iteration.

More importantly, the plan does not address what should happen to the `iterationCount` when `needs_attention` is hit and the user resets the run to `in_progress`. The counter should not reset silently, but the plan gives no guidance here.

### 4. No plan for updating the `advance_workflow` tool system prompt / LLM instructions (Task 1.3 / Task 4.1)

The Task Agent uses `advance_workflow` with `step_result`. For the feature to work end-to-end, the Task Agent's system prompt must instruct it to always pass a `step_result` value when calling `advance_workflow` after a Verify step. Without this, the LLM will omit `step_result` and the `task_result` condition will always fail with "No task result available." There is no task in the plan that updates the Task Agent prompt to teach it to pass `step_result`. This is a runtime correctness gap — the feature would be deployed but silently non-functional.

**Related file:** `packages/daemon/src/lib/space/agents/task-agent.ts` (system prompt / tool instructions)

### 5. Seeded workflow idempotency is not addressed for existing spaces (Task 4.1)

The plan's acceptance criteria says "Existing seeded workflows are unaffected (idempotency: spaces with workflows already seeded are not modified)." But `seedBuiltInWorkflows` in `built-in-workflows.ts` at line 200–203 checks `if (existing.length > 0) return` — meaning any space that already has the old 2-step Coding Workflow will never get the new 4-step version. The plan does not provide a migration strategy for existing spaces. If this is intentional (only new spaces get the verify step), the plan should say so explicitly. If the intent is to update existing spaces too, a separate migration task is needed.

### 6. `SpaceWorkflow` does not have a top-level `maxIterations` field (Task 2.1 / 4.1)

The current `SpaceWorkflow` interface in `packages/shared/src/types/space.ts` has `config?: Record<string, unknown>` for extensibility but no typed `maxIterations` field. The plan instructs adding `maxIterations?: number` to `SpaceWorkflow` (Task 2.1 subtask 2) and setting it in the template (Task 4.1 subtask 3: "the `config` field or the new top-level `maxIterations` field"), but then Task 2.2 subtask 4 says to read `workflow.maxIterations` from the template. These three subtasks need to agree on whether `maxIterations` is a first-class field or stored in `config`. The ambiguity ("or the new top-level...") will cause an inconsistent implementation.

### 7. Task 5.5 is labelled as a "unit test" but its description is an integration test

Task 5.5's description says "Create a test file in `packages/daemon/tests/unit/space/`" but the test involves a full DB-backed scenario with multiple advance cycles, task completions, and iteration counting. This test requires a real in-memory SQLite database (like the existing `space-runtime.test.ts` does). Calling it a unit test in the file location `tests/unit/space/` is fine for this project's conventions, but the plan should clarify that this test needs a real DB fixture (not just mocked objects), otherwise the coder may write a pure mock-based test that misses the DB round-trip verification.

---

## P2 — Meaningful Improvements

### 8. Missing: UI exposure of iteration count

The plan adds `iterationCount` to the run but there is no task to display it anywhere in the frontend (`packages/web/`). Users and operators have no way to see how many iterations a run has gone through. At minimum, the plan should either include a UI task or explicitly call out that this is out of scope and deferred.

### 9. Task 1.2 subtask 5 is underspecified: `listByWorkflowRun` returns all tasks, but the plan needs latest-by-step

The plan says "query tasks for the current step in this run via `this.taskManager`... filter by `workflowStepId` and `status === 'completed'`, then take the last one by `completedAt`." But `SpaceTaskManager` wraps `SpaceTaskRepository`, and `listByWorkflowRun()` returns all tasks for a run ordered by `created_at ASC`. The coder needs to filter in-memory after the list call. This should be explicit, since it means every `advance()` call does a full run task scan. For long-running workflows this is acceptable, but the plan should acknowledge this or suggest a targeted query.

### 10. No task covers updating `UpdateWorkflowRunParams` and the `updateRun()` method for the new iteration fields (Task 2.1)

`UpdateWorkflowRunParams` in `space-workflow-run-repository.ts` currently handles `title`, `description`, `status`, `currentStepId`, and `config`. Task 2.1 subtask 5 says "Add `iteration_count` and `max_iterations` to `UpdateWorkflowRunParams`" and "Update `updateRun()` to handle the new fields" — this is correct, but the plan does not mention that the `updateCurrentStep()` and `updateStatus()` convenience wrappers will continue to work as-is (they delegate to `updateRun()`). This is fine, but the coder also needs a dedicated `updateIterationCount()` helper or at least inline `updateRun({ iterationCount: ... })` call in `followTransition()` — the plan should be explicit.

### 11. `goalId` on `space_workflow_runs` introduces a second migration concern (Task 3.2)

Task 3.2 subtask 2 says "Add migration for `goal_id` column on `space_workflow_runs` (can combine with milestone 2 migration if in same PR, or add as migration 36)." This again runs into the same problem as issue #1 above — you cannot just add migration 36 because the `space_workflow_runs` table is created in migration 29. The plan needs to be consistent: either all new columns go into migration 29's idempotent block, or the plan must justify why a separate ALTER TABLE migration after 33 is acceptable for space tables.

---

## P3 — Nits

- Task 4.1 subtask 1 says "use `'reviewer'` agent role as `agentId` placeholder" but then hedges with "or `'general'` if a `reviewer` preset does not already exist." The coder should not have to check this at task time — the plan should answer it definitively. From `seed-agents.ts`, the existing preset roles are `planner`, `coder`, and `general`. There is no `reviewer` preset. The plan should say to use `'general'` or add a `reviewer` preset explicitly.

- Task 5.5 step 3 has a subtle error: "Advance (with human approval for Plan -> Code gate)" — but the Plan -> Code transition in the original Coding Workflow uses a `human` condition. After Task 4.1, Plan -> Code should still have a `human` gate, and Code -> Verify should use `always`. The test scenario should make clear that the human approval step only applies to Plan -> Code, not to any other transition, to avoid confusing the coder writing the test setup.

- Task 2.2's dependency list includes "Task 1.2 (needs the updated `followTransition` context)" — but `followTransition` is in the executor, and the changes in Task 2.2 also touch `followTransition`. These two tasks may end up modifying the same method, creating a merge conflict risk if worked in parallel (which the overview implies they can be).

lsm (commented):

### Verdict: APPROVE (with minor suggestions)

This is a well-structured plan that accurately maps to the existing codebase. The 5-milestone breakdown with 13 tasks is appropriately scoped, and I verified the plan's references against the actual code. Here is my assessment:

---

### Strengths

1. **Accurate codebase references.** The plan correctly identifies file locations, method names (`evaluateCondition`, `followTransition`, `getConditionContext`, `ConditionContext`), and the exhaustive switch pattern in `workflow-executor.ts` (line 236-239). The `step_result` field in `AdvanceWorkflowSchema` already exists as a placeholder (line 72 of `task-agent-tool-schemas.ts`), and the plan correctly notes it needs to be wired through.

2. **Correct migration numbering.** The latest migration is 33 (`runMigration33` for `autonomy_level`). The plan's assignment of migration 34 for iteration tracking columns is correct.

3. **Sound dependency ordering.** Milestones 1 and 3 are correctly identified as parallelizable. Milestone 2 depends on Milestone 1 (iteration detection uses `task_result`). Milestone 4 depends on 1+2. Milestone 5 depends on all.

4. **Good use of existing patterns.** Task 1.2's approach of adding a `case 'task_result'` to the switch statement mirrors the existing `always`/`human`/`condition` pattern. The prefix-matching semantics (`startsWith`) are practical for results like `'failed: tests broken'`.

5. **The advance_workflow handler already has the comment** (lines 506-510 of `task-agent-tools.ts`) noting that `step_result` is a placeholder not yet forwarded to the executor. The plan correctly addresses this.

---

### Issues and Suggestions

#### P2: Migration collision risk for Milestone 3

Task 3.1 says "migration 35 (or next after milestone 2's migration)" and Task 3.2 adds a *third* migration (35 or 36) for `goal_id` on `space_workflow_runs`. Since Milestones 1/3 can run in parallel and tasks may be implemented by different agents, two agents could independently create migration 34. The plan should explicitly state the migration number assignment strategy -- e.g., "Milestone 2 gets 34, Milestone 3 gets 35-36" -- or note that the integration step must reconcile migration numbers.

#### P2: `SpaceWorkflow.maxIterations` lacks persistence

Task 2.1 adds `maxIterations?: number` to the `SpaceWorkflow` type, but there is no subtask to add a DB column for it on the `space_workflows` table or update the workflow repository to persist/read it. The `SpaceWorkflow` interface maps to a DB table, so this field would be lost on read-back unless persisted. Either a migration is needed for `space_workflows.max_iterations`, or the plan should clarify that `maxIterations` lives in `SpaceWorkflow.config` (the existing opaque JSON bag) instead.

#### P2: Task 4.1 idempotency concern with seeded workflow changes

The plan says "Existing seeded workflows are unaffected (idempotency)" because `seedBuiltInWorkflows` is a no-op when workflows already exist. But this means existing spaces will never get the updated Coding Workflow with the Verify step. The plan should address whether a migration or re-seed mechanism is needed for existing spaces, or explicitly document that only newly created spaces will have the updated workflow.

#### P3: Task 2.2 subtask 1 cycle detection logic may over-count

The plan detects cycles by checking "existing tasks for this run that have the target step's workflowStepId." In a Verify->Plan->Code->Verify loop, revisiting Plan increments `iterationCount`, then revisiting Code also increments it, and then revisiting Verify also increments it. One logical iteration (one loop) would increment the counter 3 times. The acceptance criteria says "iterationCount is incremented each time a step is revisited," but the overview diagram suggests `iterationCount` tracks logical iterations (Plan->Code->Verify = 1 iteration). The counting semantics should be clarified -- per-revisit or per-cycle?

#### P3: Task 5.5 integration test assertion inconsistency

Task 5.5 says "Verify total iterationCount is 1 (only one loop back occurred)" but if the counter increments per-revisit (as described in Task 2.2), revisiting Plan, Code, and Verify in the second pass would make it 3, not 1. This needs to be consistent with the counting semantics chosen in Task 2.2.

#### P3: No RPC handler updates mentioned

The plan adds `goalId` to tasks and iteration fields to runs, but does not mention updating any RPC handlers (e.g., `space.task.list`, `space.workflowRun.get`) to expose these new fields to the frontend. If these fields are auto-included via `rowToRun()`/`rowToSpaceTask()` mapping, this is fine, but worth noting explicitly.

---

### Summary

The plan is thorough, well-scoped, and accurately grounded in the existing codebase. The dependency graph is correct, acceptance criteria are clear, and task granularity is appropriate for single coder agent sessions. The P2 issues around migration numbering, `maxIterations` persistence, and workflow re-seeding should be addressed before implementation begins to avoid rework. The P3 issues are clarification items that can be resolved during implementation.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The plan has been substantially improved and addresses most of the previous review feedback. One design gap remains that would produce incorrect behavior if implemented as written.

---

### Previous Feedback Resolution

The plan successfully resolves all P0 and most P1/P2 issues from both prior reviews:

- **Migration strategy** (was P0): Now explicitly documents the ALTER TABLE pattern following migrations 30-33, with a dedicated migration number assignment section. Well handled.
- **`advance()` signature change** (was P0): Task 1.2 subtask 5 now includes a bold note about the foundational signature change. Clear.
- **Task Agent prompt** (was P1): Task 1.3 now has subtask 3 for updating `buildTaskAgentSystemPrompt`. Correct.
- **Seeded workflow idempotency** (was P1): Explicitly documented as out of scope in Key Design Decisions. Acceptable.
- **`maxIterations` persistence ambiguity** (was P1/P2): Resolved with first-class typed field + migration 37 for `space_workflows` + migration 34 for `space_workflow_runs`. Task 2.1 subtask 7 covers the workflow repository updates. Good.
- **UI for iteration count** (was P2): Explicitly out of scope. Fine.
- **Full task scan acknowledgment** (was P2): Task 1.2 subtask 6 now documents the in-memory filtering approach. Good.
- **Migration collision risk** (was P2): Migration Number Assignment section assigns specific numbers per milestone. Addressed.
- **Agent role** (was P3): Definitively resolved to `'general'`. Good.

---

### Remaining Issues

#### P1: Cycle detection implementation guidance is contradictory (Task 2.2, subtask 1)

The overview defines per-cycle counting: "only the **first revisited step** in a loop-back increments the counter." Task 2.2 subtask 1 then says:

> "Implementation: track which step triggered the increment by comparing the transition's source step against the target. Only increment when the transition explicitly targets an already-visited step (the 'loop-back edge')."

These two statements are contradictory. Consider a Verify->Plan->Code->Verify loop:
1. Verify->Plan: Plan was already visited, so this targets an already-visited step. Increment to 1.
2. Code->Verify: Verify was also already visited, so this ALSO targets an already-visited step. The implementation hint would increment again to 2.

But the per-cycle semantics say it should only be 1. The "first step revisited" qualifier is supposed to prevent the double-count, but the implementation guidance does not explain the mechanism. The coder needs to know HOW to distinguish the "loop-back edge" from a "forward transition within the same iteration that happens to revisit a step."

One correct approach: after incrementing `iterationCount` on a loop-back, store the new iteration number in the run state (e.g., a `currentIteration` field or a set of "steps visited this iteration"). Subsequent transitions within the same iteration skip incrementing because the iteration was already counted. But no such mechanism is described in the plan.

Without this, the coder will either (a) implement the naive "increment on any revisit" approach (breaking the per-cycle semantics and causing Task 5.5 to fail), or (b) spend significant time inventing the tracking mechanism themselves with no design guidance.

**Suggested fix:** Add a concrete implementation strategy to Task 2.2 subtask 1. For example: "After incrementing `iterationCount` on a loop-back transition, record the new `iterationCount` value as a 'last incremented at iteration' marker on the run. On subsequent transitions, only increment if the current `iterationCount` has not already been incremented during this cycle. Alternatively, track a `lastLoopBackStepId` on the run — only increment when a different loop-back edge is encountered." The exact mechanism matters less than providing one.

#### P2: Task 2.1 assigns non-contiguous migration numbers

Migration 34 (Task 2.1) and migration 37 (Task 2.1) are both assigned to the same task. But migrations 35 and 36 (Milestone 3) are assigned to separate tasks (3.1 and 3.2). This means Milestone 2 gets non-contiguous migration numbers (34, 37) with a gap at 35-36 for Milestone 3. While the plan says "reconcile migration numbers before merging," the non-contiguous assignment is unusual and could confuse the coder. Consider assigning Milestone 2 migrations 34-35 and Milestone 3 migrations 36-37, or simply note that migration numbers must be contiguous and are assigned at merge time.

#### P3: Task 3.2 subtask 5 is redundant with subtask 3

Task 3.2 subtask 5 says "Update the `advance_workflow` tool handler or `SpaceRuntime` to ensure `goalId` flows from the run to newly created tasks." The `goalId` flow actually goes: run has `goalId` -> executor reads `this.run.goalId` in `followTransition()` -> passes to `taskManager.createTask()`. This is already covered by subtask 3 ("In `WorkflowExecutor`, accept a `goalId` from the run and pass it to `taskManager.createTask()` in `followTransition()`"). Subtask 5 appears redundant and may confuse the coder about whether additional wiring is needed in the tool handler.

#### P3: Overview could benefit from a step-by-step iteration count example

The Key Design Decisions section explains per-cycle semantics well in prose but would benefit from a concrete trace showing the expected `iterationCount` value at each transition in a two-cycle scenario (e.g., Plan->Code->Verify[fail]->Plan->Code->Verify[fail]->Plan->Code->Verify[pass]->Done with iterationCount going 0, 0, 0, 1, 1, 1, 2, 2, 2, 2). This eliminates ambiguity for the coder without requiring them to reverse-engineer the counting rules from prose.

---

### Summary

The plan is well-structured and addresses nearly all prior feedback. The one P1 issue — contradictory cycle detection guidance in Task 2.2 — would likely cause the coder to produce incorrect iteration counting or waste time inventing a tracking mechanism without design guidance. Once that gap is filled with a concrete implementation strategy, this plan is ready for implementation.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The revised plan substantially addresses the P0 and P1 issues from the prior reviews, but two P1 issues remain and a P2 issue with meaningful implementation consequences was not fully resolved.

---

## Overall Assessment

The revision correctly resolved: the migration strategy (following 30–33 pattern, not stuffing into migration 29's block); cycle counting semantics (per-logical-cycle, not per-step-revisit); the Task Agent system prompt update (Task 1.3 subtask 3); the `maxIterations` persistence question (first-class field on `SpaceWorkflow` with migration 37); the `general` agent role for Verify step; the seeded workflow idempotency documentation; the `advance()` signature coordination between Tasks 1.2 and 1.3; the real-DB requirement for Task 5.5; the human gate scope clarification; and the `iterationCount: 1` assertion.

These were all P0/P1 items in the prior reviews. Good work.

Three issues remain that require attention before the plan goes to a coder.

---

## P1 — Should Fix

### 1. Existing Task Agent system prompt says `result`, not `step_result` — Task 1.3 must fix the existing stale text, not only append new text

Task 1.3 subtask 3 says to "include an instruction" about passing `step_result` on verify steps. But in `packages/daemon/src/lib/space/agents/task-agent.ts` at line 174–177, the existing system prompt already says:

> "Pass the `result` of the completed step."

The tool schema's field name is `step_result` (not `result`). This means the current prompt already contains a wrong field name. If Task 1.3 only appends a new `step_result` note without correcting line 175, the LLM will see both `result` and `step_result` referenced in the same prompt, creating confusion about which field to use. The task must be explicit: line 175's `result` → `step_result` must be corrected, not just supplemented.

**File:** `packages/daemon/src/lib/space/agents/task-agent.ts`, line 175

### 2. Iteration detection logic: the plan conflates "any task in this run for step X" with "this is a loop-back edge" — this will misfire in any DAG with merging paths

Task 2.2 subtask 1 says to detect a cycle by checking "if any completed/in-progress tasks exist for [the target step] in this run." But this heuristic fires on any transition to a previously-visited step, including a legitimate converging edge in a non-cyclic DAG (e.g., both A and B converge to C — the second time C is reached via B, it looks like a cycle).

The plan tries to address this in the counting semantics section: "only the first revisited step in a loop-back increments the counter." But the implementation instruction still says "if any tasks exist for that step" — which has nothing to do with whether the transition is topologically a back-edge versus a merge edge in the graph.

For the current use case (a simple linear Verify→Plan cycle), this heuristic works in practice. But the plan is building iteration tracking as a general feature (Task 2.1 adds `max_iterations` to the workflow template type, not just the seeded workflow). If a more complex workflow is built later, the misfiring counter will produce wrong results silently.

The plan should either:
a) Explicitly scope the cycle detection to the simple "transition points backward to an earlier step by graph order" check, and document the limitation, OR
b) Require that cyclic transitions be explicitly annotated (e.g., a boolean `isCyclic` on the transition) so the detection is not based on task history heuristics.

This is a design decision the plan must settle before the coder implements it.

**File affected:** `packages/daemon/src/lib/space/runtime/workflow-executor.ts` (followTransition method)

---

## P2 — Meaningful Improvements

### 3. The `createWorkflow` path does not thread `maxIterations` through — Task 2.1 omits the `CreateSpaceWorkflowParams` update and the `seedBuiltInWorkflows` call path

Task 2.1 correctly adds `maxIterations?: number` to `SpaceWorkflow` and migration 37 adds the DB column. But the `createWorkflow` call in `seedBuiltInWorkflows` (at line 246 of `built-in-workflows.ts`) passes a `CreateSpaceWorkflowParams` object. If `maxIterations` is not also added to `CreateSpaceWorkflowParams` (in `packages/shared/src/types/space.ts`) and threaded through the workflow repository's `createWorkflow()` INSERT statement, the `maxIterations: 3` set on the template (Task 4.1 subtask 3) will be silently dropped when the template is persisted.

Looking at `space-workflow-repository.ts` lines 179–194, the INSERT only includes columns: `id`, `space_id`, `name`, `description`, `start_step_id`, `config`, `layout`, `created_at`, `updated_at`. Adding migration 37's `max_iterations` column requires also updating the INSERT and the `rowToWorkflow()` function to include the new column.

Task 2.1 subtask 7 says "Update `createWorkflow()` / `updateWorkflow()` to handle `max_iterations`" — this is correct but it is buried in subtask 7 and does not mention updating `CreateSpaceWorkflowParams`. Without that type update, TypeScript will not catch the omission and the `seedBuiltInWorkflows` code in Task 4.1 subtask 3 will silently fail to persist `maxIterations: 3`.

The plan should add an explicit subtask: "Add `maxIterations?: number` to `CreateSpaceWorkflowParams` in `packages/shared/src/types/space.ts` and to `UpdateSpaceWorkflowParams`."

**Files:** `packages/shared/src/types/space.ts` (~line 630), `packages/daemon/src/storage/repositories/space-workflow-repository.ts` (`createWorkflow`, `rowToWorkflow`)

---

## P3 — Nits

### 4. Task 4.1 transition description placement is ambiguous

Task 4.1 subtask 2 says "Add Verify -> Plan transition with `task_result` condition, expression: 'failed', order: 0, **description**: 'Loop back to planning...'" — the `description` field lives on `WorkflowCondition`, not on `WorkflowTransition` (the `WorkflowTransition` type has no `description` field). Looking at the existing Coding Workflow template, the `description` is correctly placed on the `condition` object: `condition: { type: 'human', description: '...' }`. The plan should make this placement explicit to avoid the coder incorrectly adding a `description` field to the transition root object.

### 5. Task 2.1 lists migrations 34 and 37 but they are specified across two separate subtasks (4 and 5) — the registration order in `runMigrations()` matters

The `runMigrations()` function must call `runMigration34`, ..., `runMigration37` in numerical order. If Milestone 2 and Milestone 3 are worked in parallel and PRs merge in a different order (e.g., migration 35 merges before 34), the function will call migrations out of sequence. The overview's migration number assignment table (34 for iteration, 35 for goal on tasks, 36 for goal on runs, 37 for max_iterations on workflows) is good, but the plan should state explicitly that the `runMigrations()` call-site order must be reconciled at merge time — not just the migration number assignment.

---

## Summary of Prior Issues and Resolution Status

| Prior Issue | Status |
|---|---|
| P0: Wrong migration numbers (stuffing into migration 29) | **Resolved** — plan now follows 30–33 pattern |
| P0: `advance_workflow` step_result not forwarded | **Resolved** — Task 1.3 now explicitly wires it through |
| P1: Cycle detection logic fragile | **Partially addressed** — semantics clarified but implementation heuristic still has the merge-path flaw (see issue #2 above) |
| P1: No system prompt update for step_result | **Partially addressed** — Task 1.3 subtask 3 added, but stale `result` text not corrected (see issue #1 above) |
| P1: Seeded workflow idempotency undocumented | **Resolved** — clearly documented as out of scope |
| P1: maxIterations in config vs. first-class field | **Resolved** — migration 37 + first-class field specified |
| P2: Migration collision risk | **Resolved** — migration table with explicit assignments added |
| P2: maxIterations persistence path incomplete | **Partially resolved** — migration 37 added but CreateSpaceWorkflowParams omitted (see issue #3 above) |
| P3: reviewer vs. general preset | **Resolved** — 'general' specified definitively |
| P3: Task 5.5 real DB clarification | **Resolved** |
| P3: iterationCount assertion corrected | **Resolved** |

lsm (commented):

**Recommendation: APPROVE** — All P0 and P1 issues from prior reviews have been properly resolved. The remaining issues are minor.

---

### Prior Feedback Resolution

The round 2 revision correctly addresses all outstanding issues from the two prior review rounds:

1. **Cycle detection heuristic (was P1 in round 2):** Replaced entirely with the explicit `isCyclic: true` flag on `WorkflowTransition`. This is a clean, unambiguous design that eliminates the heuristic-vs-DAG-merge-path problem. The implementation guidance in Task 2.2 is now straightforward: check a boolean flag. Well done.

2. **Stale `result` field name in Task Agent prompt (was P1 in round 2):** Task 1.3 subtask 3 now explicitly calls out fixing line ~175 (`result` → `step_result`) as a separate bold bullet point alongside the new instruction. Both changes are marked as essential. Resolved.

3. **`maxIterations` on `CreateSpaceWorkflowParams` (was P2 in round 2):** Task 2.1 now includes a dedicated subtask 4 adding `maxIterations` to `CreateSpaceWorkflowParams` and `UpdateSpaceWorkflowParams`, with an explicit note about why this is needed for `seedBuiltInWorkflows`. Resolved.

4. **Non-contiguous migration numbers (was P2 in round 2):** Migration assignments are now contiguous per milestone: 34-35 for Milestone 2, 36-37 for Milestone 3. Resolved.

5. **Redundant Task 3.2 subtask 5 (was P3 in round 2):** Removed. Task 3.2 now has 5 subtasks (ending with typecheck) instead of 6. Resolved.

6. **Iteration count trace example (was P3 in round 2):** The overview now includes a concrete step-by-step trace showing `iterationCount` values at each transition in a two-cycle scenario. Eliminates ambiguity for the coder. Resolved.

---

### Remaining Issues

#### P3: Duplicate subtask numbering in Task 2.1

`02-iteration-tracking.md` has two subtasks numbered "4":
- Subtask 4 (line ~31): `Add maxIterations to CreateSpaceWorkflowParams and UpdateSpaceWorkflowParams...`
- Subtask 4 (line ~32): `In packages/daemon/src/storage/schema/migrations.ts, add migration 34...`

The second "4" should be "5", and subsequent subtasks should be renumbered accordingly (current 5→6, 6→7, 7→8). This is a formatting error from the revision — the content is correct, just the numbering is off.

#### P3: `isCyclic` not mentioned for `ExportedWorkflowTransition` or export/import logic

`ExportedWorkflowTransition` (line 739 of `space.ts`) is a separate interface from `WorkflowTransition` — it does not extend or use `Omit` from it. When `isCyclic` is added to `WorkflowTransition`, it will NOT automatically appear on `ExportedWorkflowTransition`. Similarly, the export logic in `export-format.ts` (line 192-198) manually picks fields from `WorkflowTransition` to build the exported form — it would need `if (t.isCyclic !== undefined) exported.isCyclic = t.isCyclic;` added.

This is a minor gap — workflow export/import is a secondary feature and the coder implementing Task 1.1 will likely notice the parallel interface when adding the field. But worth noting for completeness.

#### P3: `description` field placement on Verify→Done transition in Task 4.1

Task 4.1 subtask 2 correctly places `description` on the `condition` object for the Verify→Plan transition. The Verify→Done transition description is also correctly placed: `Condition: { type: 'task_result', expression: 'passed', description: '...' }`. Good — this was flagged in the round 2 review and is now consistently correct.

---

### Summary

The plan is well-structured, addresses all prior review feedback, and provides clear implementation guidance. The `isCyclic` flag design is a significant improvement over the heuristic approach — it makes cycle detection deterministic and trivial to implement. The iteration count trace example eliminates ambiguity. The only remaining issues are a numbering typo and a minor gap around the export format, neither of which would block implementation.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — several issues remain after two prior rounds.

---

## Summary

This revision resolves the three most impactful prior issues — `isCyclic` flag, stale `result` field name in the prompt, and `maxIterations` on `CreateSpaceWorkflowParams`. The migration number contiguity is also fixed. However, a number of gaps remain, two of which are P0 (plan claims that have no corresponding implementation guidance) and several P1/P2 issues that leave coder agents in positions where they can silently produce broken behavior.

---

## What Was Fixed (Acknowledged)

- `isCyclic: true` on `WorkflowTransition` replaces heuristic cycle detection — correct.
- Stale `result` field name at `task-agent.ts:175` is now targeted by Task 1.3 subtask 3.
- `maxIterations` added to `CreateSpaceWorkflowParams` in Task 2.1 subtask 4 — correct.
- Migration numbers assigned contiguously (34–37) — correct.

---

## Remaining Issues

### P0 — `seedBuiltInWorkflows` does not pass `maxIterations` to `createWorkflow()`

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/04-update-seeded-coding-workflow.md` — Task 4.1, subtask 3

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/02-iteration-tracking.md` — Task 2.1, subtask 4

The plan correctly adds `maxIterations` to `CreateSpaceWorkflowParams`, but `seedBuiltInWorkflows()` in `built-in-workflows.ts` calls `workflowManager.createWorkflow({...})` with a hardcoded argument object that does **not** forward `template.maxIterations`. The current call at line 246–255 of `built-in-workflows.ts` only passes `spaceId`, `name`, `description`, `steps`, `transitions`, `startStepId`, `rules`, and `tags`.

Task 4.1 says "Set `maxIterations: 3` on `CODING_WORKFLOW`" but never explicitly instructs the coder to update the `seedBuiltInWorkflows` call-site to forward `maxIterations` from the template to `createWorkflow()`. Without this step, the field will be set on the template object but silently dropped on persist, leaving all seeded Coding Workflow runs using the DB default of 5 iterations instead of the intended 3. This is the same class of bug Task 2.1 subtask 4 explicitly called out for `seedBuiltInWorkflows`, yet the fix was only applied to `CreateSpaceWorkflowParams` — the call-site forwarding is not instructed anywhere in Task 4.1.

**Fix needed:** Task 4.1 subtask 3 must instruct the coder to update `seedBuiltInWorkflows()` to pass `maxIterations: template.maxIterations` through the `workflowManager.createWorkflow()` call.

---

### P0 — `WorkflowTransitionInput` does not include `isCyclic`

**File:** `packages/shared/src/types/space.ts`, line 526

`WorkflowTransitionInput` is defined as `Omit<WorkflowTransition, 'id'>`. Once `isCyclic?: boolean` is added to `WorkflowTransition` by Task 1.1, it will automatically be included in `WorkflowTransitionInput`. However, `seedBuiltInWorkflows()` maps template transitions at line 237–242 and constructs a new object:

```ts
const transitions = template.transitions.map((t) => ({
    from: stepIdMap.get(t.from)!,
    to: stepIdMap.get(t.to)!,
    condition: t.condition,
    order: t.order,
}));
```

`isCyclic` is not forwarded here. The plan never instructs the coder to update this mapping. The `isCyclic: true` flag on the Verify→Plan transition in the template will be silently dropped at seeding time, meaning seeded Coding Workflow runs will never increment `iterationCount` even when they loop back. This silently defeats the entire iteration-tracking feature for the primary intended use case.

**Fix needed:** Task 4.1 must instruct the coder to add `isCyclic: t.isCyclic` to the transition mapping inside `seedBuiltInWorkflows()`.

---

### P1 — `advance()` DB-query precedence vs. fallback is described backwards in acceptance criteria

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/01-task-result-condition-type.md` — Task 1.2, subtask 5 and Task 1.3, acceptance criterion 2

Task 1.2 subtask 5 says: "If the DB task result is null/empty but `options.stepResult` is provided, use it as fallback." Task 1.3 acceptance criterion 2 says: "DB task result takes precedence; `step_result` is used as fallback when DB result is absent."

This is internally consistent within the plan, but it creates a practical problem: the step agent sets the task's `result` via `report_result` → `taskManager.setTaskStatus(..., { result: summary })`. When the Task Agent calls `advance_workflow` with `step_result`, the DB already contains the step agent's `result`. The executor will use the DB value and ignore `step_result`. This means the explicit `step_result` argument passed by the Task Agent is only used if the step agent never recorded a result — which is the unusual case. The common case (step agent ran `report_result` with a summary that includes "passed" or "failed") works correctly via DB lookup, but the Task Agent prompt instruction to "always include the `step_result` field" in Task 1.3 subtask 3 would be misleading in that common case, since the value passed is silently overridden.

The plan should clarify whether the Task Agent's `step_result` is meant to supplement or override the DB result, and update the prompt instruction accordingly. As written, instructing the Task Agent to always pass `step_result` when the executor ignores it in the common case creates unnecessary LLM confusion and prompt noise.

---

### P1 — `formatTransition()` in `task-agent.ts` does not label `task_result` transitions

**File:** `packages/daemon/src/lib/space/agents/task-agent.ts`, lines 91–104

The `formatTransition()` helper only labels `human` and `condition` type transitions. There is an explicit comment: "Any future WorkflowConditionType values not handled here will also produce no label; add a branch above when new types are introduced."

The plan (Task 1.3) never instructs the coder to update `formatTransition()` to label `task_result` transitions. As a result, the Task Agent's initial message will show Verify→Plan and Verify→Done transitions without any label, giving the LLM no signal that result-based routing is happening. The Task Agent would be unaware that `step_result` influences which transition fires. This undermines the LLM's ability to reason correctly about the workflow.

**Fix needed:** Task 1.3 (or a new sub-task) should instruct the coder to add a `task_result` branch to `formatTransition()` in `task-agent.ts`.

---

### P1 — `needs_attention` escalation path leaves executor in `executors` map

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/02-iteration-tracking.md` — Task 2.2, subtask 2

The plan says: when `iterationCount >= maxIterations`, set the run status to `needs_attention`. But `SpaceRuntime` removes executors from the `executors` map only when runs reach `completed` or `cancelled` (terminal states). `needs_attention` is not terminal. The plan does not address whether the executor should be removed, retained, or replaced when the cap is hit. If the executor stays in the map, a human reset (setting status back to `in_progress`) will cause the next tick to call `advance()` on the still-capped executor, which will immediately hit the cap again because `iterationCount` is not reset (by the plan's own reset-behavior rule). The plan says "the human may increase `maxIterations` on the run to allow more cycles" but provides no mechanism for the executor to pick up an updated `maxIterations` after the human changes it. Task 2.2 needs to specify how the executor re-reads `maxIterations` after a human reset.

---

### P2 — Integration test's human-gate assumption conflicts with the workflow graph

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/05-tests.md` — Task 5.5, subtask 3

The integration test scenario says "Advance with human approval for the Plan → Code gate (this is the only `human` gate in the workflow)." This is accurate for the first Plan→Code transition. However, after the loop-back (Verify→Plan with `isCyclic: true`), the workflow re-enters the Plan step and the Plan→Code transition is again a `human` condition. The scenario correctly handles this in the second pass ("Simulate Plan(2) completion, advance with human approval"), but the test setup comment should be updated to say "each Plan→Code advance requires human approval" to avoid confusing the implementing coder who may think only one gate exists total.

---

### P2 — `isCyclic` is on `WorkflowTransition` but `WorkflowTransitionInput` is used for persistence in `createWorkflow()`

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/01-task-result-condition-type.md` — Task 1.1, subtask 2

`WorkflowTransitionInput = Omit<WorkflowTransition, 'id'>` automatically inherits `isCyclic`. The plan correctly places `isCyclic` on `WorkflowTransition`. However, the plan does not instruct the coder to check whether the `space_workflow_transitions` DB table (created in Migration 29) includes an `is_cyclic` column, and whether the workflow repository's transition INSERT and SELECT queries need to be updated. Without persisting `isCyclic` to the DB, transitions loaded from the DB for rehydrated executors will lose the flag. Task 1.1 should explicitly instruct: verify and update the transition repository to persist and read `is_cyclic`.

---

### P2 — `advance()` task scan is acknowledged as full-run scan but deferred without bound

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/01-task-result-condition-type.md` — Task 1.2, subtask 6

The plan acknowledges that filtering tasks in-memory by `workflowStepId` does a full-run scan: "Note: this does a full run task scan, which is acceptable for typical workflow sizes." But with cyclic workflows, the same step (e.g., Plan) gets a new task on every iteration. After N iterations the scan over the full run's tasks grows linearly. The plan defers this with "If performance becomes an issue later, a targeted query can be added." This is acceptable as a P2 decision note, but it should also specify that `listByWorkflowRun()` returns tasks ordered by `created_at ASC` so the "most recently completed" selection (take the last one by `completedAt`) is deterministic. The current `listByWorkflowRun()` ordering is not documented in the plan.

---

### P3 — Task 2.1 has a duplicate numbering error

**File:** `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/02-iteration-tracking.md` — Task 2.1

Subtask 3 is `Add maxIterations?: number to CreateWorkflowRunParams.` and subtask 4 appears twice (there are two items labeled "4." — the `CreateSpaceWorkflowParams` subtask and the migration 34 subtask). This is a minor formatting error that could confuse an implementing agent.

---

## Verification of Prior P0/P1 Fixes

| Prior issue | Status |
|---|---|
| Heuristic cycle detection | Fixed: `isCyclic: true` flag introduced |
| Stale `result` in system prompt | Fixed: Task 1.3 subtask 3 targets `task-agent.ts:175` |
| `maxIterations` missing from `CreateSpaceWorkflowParams` | Fixed: Task 2.1 subtask 4 adds it |
| Migration number gaps | Fixed: contiguous 34–37 assigned in overview |
| `followTransition` vs `advance` modification conflict note | Added in Task 2.2 |

lsm (commented):

**Recommendation: APPROVE** — All P0 and P1 issues from the three prior review rounds have been properly addressed. The remaining issues are P3 nits only.

---

### Prior Feedback Resolution

This round 3 revision resolves all outstanding issues from the prior reviews:

1. **`seedBuiltInWorkflows` not forwarding `isCyclic` (was P0 in round 3):** Task 4.1 subtask 4 now explicitly instructs the coder to add `isCyclic: t.isCyclic` to the transition mapping at line ~237-242 and to pass `maxIterations: template.maxIterations` through the `workflowManager.createWorkflow()` call. I verified the actual code at `packages/daemon/src/lib/space/workflows/built-in-workflows.ts` lines 237-255 — the plan correctly identifies both the transition mapping gap and the `createWorkflow()` call-site gap. Well done.

2. **`isCyclic` not persisted to DB (was P2 in round 3):** Task 1.1 subtask 4 now explicitly instructs adding an `is_cyclic` column via migration 34 on `space_workflow_transitions`, and updating the repository's INSERT and SELECT queries. The migration number assignment table in the overview has been updated accordingly (34 for `is_cyclic`, 35-36 for iteration tracking, 37-38 for goalId). Resolved.

3. **`ExportedWorkflowTransition` not updated (was P3 in round 3):** Task 1.1 subtask 3 now explicitly adds `isCyclic?: boolean` to `ExportedWorkflowTransition` and instructs updating the export logic in `export-format.ts`. I verified that `ExportedWorkflowTransition` (line 739 of `space.ts`) is indeed a standalone interface that does NOT extend `WorkflowTransition` — the explicit subtask is necessary. Resolved.

4. **`formatTransition()` not labeling `task_result` transitions (was P1 in round 3):** Task 1.3 subtask 4 now instructs adding a `task_result` branch to `formatTransition()` in `task-agent.ts`. I verified the actual code at lines 91-104 — the comment at line 100-102 explicitly says "add a branch above when new types are introduced." The plan correctly references this. Resolved.

5. **`needs_attention` executor lifecycle (was P1 in round 3):** Task 2.2 subtask 2 now includes a dedicated "Executor lifecycle" paragraph explaining that the executor stays in the map (not terminal), and recommends option (a): have `advance()` always re-read `maxIterations` from the DB rather than using a cached value. This provides clear implementation guidance. Resolved.

6. **Task 2.1 duplicate subtask numbering (was P3 in round 3):** Fixed — subtasks are now numbered 1-9 correctly.

7. **Task 5.5 human-gate clarification (was P2 in round 3):** Updated to say "each time Plan→Code is traversed (including after loop-backs), human approval is required." Resolved.

8. **`listByWorkflowRun()` ordering documented (was P2 in round 3):** Task 1.2 subtask 6 now documents that `listByWorkflowRun()` returns tasks ordered by `created_at ASC` and specifies sorting by `completedAt` descending. Resolved.

9. **`step_result` precedence clarified (was P1 in round 3):** Task 1.3 acceptance criterion 2 now includes a note explaining the common case (step agent sets DB result via `report_result`) and that `step_result` from `advance_workflow` is primarily useful as a fallback. Resolved.

---

### Remaining Issues

#### P3: Task 1.1 subtask 4 says "Add an ALTER TABLE migration or check if the column can be added idempotently" — the "or" is vague

The plan assigns this as migration 34 in the note below the subtasks, but the subtask text itself hedges between two approaches. Since the plan has already decided on the ALTER TABLE pattern (following migrations 30-33), the subtask text should just say "Add migration 34 with ALTER TABLE ... ADD COLUMN is_cyclic" without the "or check if" alternative. Minor — the note below clarifies the intent.

#### P3: Task 1.3 acceptance criterion on `step_result` precedence could be confusing to the coder

The note says `step_result` is "primarily useful when the step agent does not set a result, or for override scenarios" — but the implementation in Task 1.2 subtask 5 says DB result takes precedence over `step_result`. So `step_result` cannot actually override a DB result. The word "override" is slightly misleading. This is unlikely to cause a bug since the implementation logic in Task 1.2 is clear, but the acceptance criterion language is imprecise.

---

### Summary

The plan is thorough, well-structured, and ready for implementation. All five milestones have clear task scoping, correct dependency ordering, and actionable subtasks with accurate codebase references. The `isCyclic` flag design is clean and eliminates heuristic-based cycle detection entirely. The migration assignment table, iteration count trace example, and executor lifecycle guidance provide the coder with unambiguous implementation direction. Only minor wording nits remain.

lsm (commented):

> **Recommendation:** REQUEST_CHANGES — several remaining issues need attention before implementation begins.

---

## Summary

This is round 4 of the plan review. The three P0 issues from prior rounds have been addressed (isCyclic DB persistence, seedBuiltInWorkflows forwarding isCyclic/maxIterations, formatTransition labeling). However, several new P1/P2 issues were introduced by the fixes, and one pre-existing P1 remains unresolved.

---

## Resolved From Prior Rounds

- isCyclic now persisted to DB via migration 34 (Task 1.1 subtask 4). Good.
- seedBuiltInWorkflows now explicitly instructed to forward isCyclic in transition mapping and maxIterations to createWorkflow() (Task 4.1 subtask 4). Good.
- formatTransition() update for task_result labeling added (Task 1.3 subtask 4). Good.
- needs_attention executor lifecycle documented in Task 2.2. Good.
- ExportedWorkflowTransition update called out explicitly (Task 1.1 subtask 3). Good.

---

## Remaining Issues

### P1 — Wrong file path for export-format.ts (Task 1.1 subtask 3)

Task 1.1 subtask 3 directs the coder to `packages/daemon/src/lib/space/export/export-format.ts (line ~192-198)`. The actual file is at `packages/daemon/src/lib/space/export-format.ts` — there is no `export/` subdirectory. A coder agent following this instruction will waste time searching for a nonexistent file.

**Fix:** Correct the path to `packages/daemon/src/lib/space/export-format.ts`.

---

### P1 — isCyclic persisted in condition JSON column, not a DB column (Task 1.1 subtask 4)

The plan instructs adding a separate `is_cyclic` column to `space_workflow_transitions` (migration 34). However, the repository's `insertTransition()` stores the entire `condition` object as a JSON blob in the `condition` column, and `isCyclic` belongs on `WorkflowTransition` (not on `WorkflowCondition`). There is no ORM or column-per-field pattern for transitions — the repo reads/writes transitions as id/from/to/condition/order_index, and any extra fields on `WorkflowTransitionInput` would be silently dropped.

Adding a separate `is_cyclic` column is one valid option, but the plan does not acknowledge that `insertTransition()` must be updated to include that column in its INSERT statement, and `rowToTransition()` must read it back. The plan says "Update the workflow transition repository's INSERT and SELECT queries to include is_cyclic" but does not spell out that the `insertTransition()` private method is the actual code path — which is important since the method's signature takes a `WorkflowTransitionInput` and the coder must also add `isCyclic` to that type. Without this, the column would be added to the table but never populated.

**Fix:** Add an explicit subtask calling out `insertTransition()` and `WorkflowTransitionInput` as the change points for the is_cyclic column persistence.

---

### P1 — advance() signature change creates a semantic gap with the DB-first precedence rule

Task 1.2 subtask 5 says: "If the DB task result is null/empty but options.stepResult is provided, use it as fallback." Task 1.3 acceptance criteria say: "DB task result takes precedence; step_result is used as fallback when DB result is absent."

However, the `advance_workflow` handler already queries `taskRepo.listByWorkflowRun()` to verify the task is completed before calling `executor.advance()` (lines 539-554 of task-agent-tools.ts). The task is confirmed `completed` before advance() is called. A completed task with no DB result field is a normal case (step agents are not required to call `report_result` — the completion callback handles status). The plan should clarify whether the fallback path is actually reachable in normal flow, or whether this is only useful for edge cases where `report_result` is never called. Without this clarification, the coder may implement dead code with unclear semantics.

**Fix:** Add a note in Task 1.2 or 1.3 explaining the exact conditions under which stepResult fallback is triggered vs. the DB result being used, and confirm the verification logic in advance_workflow still makes sense after this change.

---

### P1 — maxIterations stored in config JSON blob, not a dedicated column (Task 2.1 subtasks 6 and 8)

The plan (Task 2.1 subtask 6) proposes migration 36 adds a `max_iterations` column to `space_workflows`. But `rowToWorkflow()` in `space-workflow-repository.ts` builds `SpaceWorkflow` from a `WorkflowRow`, and that struct reads tags/rules/extra from a `config` JSON blob. There is no `max_iterations` field in `WorkflowRow` or `WorkflowConfigJson`. Task 2.1 subtask 8 says to update `rowToWorkflow()` to read `max_iterations` from the DB row, but it does not say to update the `WorkflowRow` interface or `WorkflowConfigJson` accordingly.

Similarly, the `createWorkflow()` INSERT in the repository does not include `max_iterations`, and the plan does not instruct the coder to add it to the INSERT statement.

**Fix:** The subtask should explicitly name the `WorkflowRow` interface, the INSERT statement in `createWorkflow()`, and the UPDATE logic in `updateWorkflow()` as the three change points for the `max_iterations` column.

---

### P2 — Zod schema in export-format.ts must also accept task_result (Task 1.1)

The import/export Zod schema in `packages/daemon/src/lib/space/export-format.ts` (line 38) hardcodes the allowed condition types:

```ts
type: z.enum(['always', 'human', 'condition']),
```

When a workflow containing a `task_result` transition is exported and then re-imported, the Zod validator will reject it with an enum error. The plan does not include updating this import schema. This would surface immediately when the seeded Coding Workflow (which has task_result transitions) is exported.

**Fix:** Add a subtask to Task 1.1 (or Task 4.1) to update the Zod schema in `export-format.ts` to include `'task_result'` in the enum.

---

### P2 — No test coverage for export/import round-trip with task_result transitions

Milestone 5 tests do not include a case for exporting and importing a workflow with `task_result` transitions or `isCyclic: true`. Given that `isCyclic` is not stored on `WorkflowCondition` but on `WorkflowTransition`, and the export format uses `ExportedWorkflowTransition`, this round-trip is a non-obvious path.

**Fix:** Add a test case to Task 5.4 that verifies export and re-import of the updated Coding Workflow preserves the isCyclic flag and task_result conditions.

---

### P2 — Missing 'failed' status for task in ReportResultSchema but used in Verify step instructions

The Verify step instructions say to set result to `'failed: <reason>'` (Task 4.1 subtask 2). The `step_result` is passed through `AdvanceWorkflowInput.step_result` and used in `evaluateCondition`. But `ReportResultSchema` (task-agent-tool-schemas.ts) does not include `'failed'` as a valid `status` value — only `'completed'`, `'needs_attention'`, and `'cancelled'`. These are two different things (the task status vs. the step result string), but the prompt conflates them at lines 174-176 of task-agent.ts where it says "Pass the result of the completed step" — unchanged stale wording. Task 1.3 subtask 3 fixes only one part of this. The plan should confirm that the system prompt distinction between `report_result.status` and `advance_workflow.step_result` is made clear in the updated prompt.

**Fix:** Verify and document in Task 1.3 that the prompt update clearly distinguishes `step_result` (a string like 'passed'/'failed: ...') from `report_result.status` (an enum of 'completed'/'needs_attention'/'cancelled').

---

### P3 — Task 5.5 integration test plan step omits human approval for second Plan→Code gate

Task 5.5 step 3 describes: "Simulate Plan(2) completion, advance with human approval, simulate Code(2) completion." This is correct — every Plan→Code traversal requires human approval. However, the integration test setup uses a real DB, and the mechanism for simulating human approval (setting `run.config.humanApproved = true` via `updateRun()`) is not mentioned explicitly for the second cycle. The original `clearHumanApproval()` in advance() clears this after each human gate. The test should explicitly set it again before the second Plan→Code advance.

**Fix:** Add a note in Task 5.5 to explicitly call `updateRun({ config: { humanApproved: true } })` before each Plan→Code advance, including the second iteration.

---

## Summary Table

| Issue | Severity | File/Location |
|-------|----------|---------------|
| Wrong path for export-format.ts | P1 | docs/plans/.../01-task-result-condition-type.md, Task 1.1 subtask 3 |
| insertTransition() / WorkflowTransitionInput not called out for is_cyclic | P1 | docs/plans/.../01-task-result-condition-type.md, Task 1.1 subtask 4 |
| DB-first vs. step_result fallback semantics unclear | P1 | docs/plans/.../01-task-result-condition-type.md, Task 1.2 subtask 5 / Task 1.3 |
| WorkflowRow interface / INSERT statement gaps for max_iterations | P1 | docs/plans/.../02-iteration-tracking.md, Task 2.1 subtasks 6/8 |
| Zod import schema missing task_result | P2 | docs/plans/.../01-task-result-condition-type.md, missing subtask |
| No export/import round-trip test for task_result/isCyclic | P2 | docs/plans/.../05-tests.md, Task 5.4 |
| Prompt clarity: step_result vs. report_result.status distinction | P2 | docs/plans/.../01-task-result-condition-type.md, Task 1.3 |
| humanApproved re-set not explicit in second iteration of integration test | P3 | docs/plans/.../05-tests.md, Task 5.5 |

lsm (commented):

**Recommendation: APPROVE** — All issues from the four prior review rounds have been properly resolved.

---

### Round 4 Feedback Resolution

I verified each issue raised by the claude-sonnet-4-6 round 4 review against the current plan text:

1. **P1 -- Wrong file path for export-format.ts:** Fixed. Task 1.1 subtask 3 now reads `packages/daemon/src/lib/space/export-format.ts` with the explicit note `(note: no export/ subdirectory)`. I confirmed the file exists at that path.

2. **P1 -- insertTransition()/TransitionRow/rowToTransition() not called out for is_cyclic:** Fixed. Task 1.1 subtask 4 now explicitly names all three change points: `insertTransition()` (line ~409), `rowToTransition()` (line ~107), and the `TransitionRow` interface. It also specifies the mapping logic (`input.isCyclic ? 1 : 0` for INSERT, `Boolean(row.is_cyclic)` for read).

3. **P1 -- advance() step_result fallback semantics unclear:** Fixed. Task 1.2 subtask 5 now includes a dedicated "When is the fallback used?" paragraph that enumerates the two specific cases: (a) step agent completes without calling `report_result`, and (b) step agent sets a generic result not matching the expected prefix. This eliminates ambiguity about whether the fallback is dead code.

4. **P1 -- WorkflowRow/INSERT/UPDATE/rowToWorkflow() not called out for max_iterations:** Fixed. Task 2.1 subtask 8 now explicitly names four change points: `WorkflowRow` interface (line ~35), `createWorkflow()` INSERT (line ~183), `updateWorkflow()` UPDATE, and `rowToWorkflow()` (line ~118).

5. **P2 -- Zod import schema missing task_result:** Fixed. Task 1.1 subtask 3 now includes explicit instruction to update the `workflowConditionSchema` Zod enum at line ~38 of `export-format.ts` to include `'task_result'`. I confirmed the actual Zod line matches: `z.enum(['always', 'human', 'condition'])`.

6. **P2 -- No export/import round-trip test:** Fixed. Task 5.4 subtask 2 now includes a bullet: "Verify export/import round-trip: export the updated Coding Workflow, re-import it, and verify that `isCyclic: true` on the Verify->Plan transition and `task_result` conditions are preserved."

7. **P2 -- Prompt clarity: step_result vs report_result.status:** Fixed. Task 1.3 subtask 3 now includes an "Important distinction" paragraph that clearly separates `report_result` (task lifecycle enum: completed/needs_attention/cancelled) from `advance_workflow` with `step_result` (free-form string for routing). The acceptance criteria also reflect this.

8. **P3 -- humanApproved re-set not explicit in second iteration:** Fixed. Task 5.5 step 3 now explicitly says "Set `humanApproved` on the run config again (the executor's `clearHumanApproval()` clears it after each human gate)."

---

### Overall Assessment

The plan is comprehensive and implementation-ready. After five revision rounds, it provides:

- Accurate file paths and line number references verified against the codebase
- Explicit change-point naming for all DB persistence touchpoints (insertTransition, rowToTransition, WorkflowRow, createWorkflow INSERT, etc.)
- Clear fallback semantics documentation for step_result
- Complete migration number assignment table (34-38) with consistent patterns
- Well-specified test coverage including the export/import round-trip gap identified in round 4

No remaining issues found.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — One P1 gap remains. All prior round issues were verified as fixed, but the fix for `ExportedWorkflowTransition`/Zod schema update is incomplete: the plan instructs adding `isCyclic` to the TypeScript type and to the export logic, but does not instruct updating the Zod validation schema for the imported transition, which means `isCyclic` will be silently stripped on import.

---

## Prior Issues — All Verified Fixed

Every issue raised in rounds 1–4 was verified against the current plan text and codebase:

1. **File path for `export-format.ts`** (was P1 R4): Task 1.1 subtask 3 now reads `packages/daemon/src/lib/space/export-format.ts` with `(note: no export/ subdirectory)`. Correct — file exists at that path.

2. **`insertTransition()`/`TransitionRow`/`rowToTransition()` for `is_cyclic`** (was P1 R4): Task 1.1 subtask 4 now explicitly names all three change points with line number references and specifies the mapping logic (`input.isCyclic ? 1 : 0` for INSERT, `Boolean(row.is_cyclic)` for read). Correct — verified against actual `TransitionRow` (line 58-67) and `insertTransition()` (line 409-434) in `space-workflow-repository.ts`.

3. **`advance()` step_result fallback semantics** (was P1 R4): Task 1.2 subtask 5 now includes a "When is the fallback used?" paragraph with two specific cases enumerated. Correct.

4. **`WorkflowRow`/INSERT/UPDATE/`rowToWorkflow()` for `max_iterations`** (was P1 R4): Task 2.1 subtask 8 now names all four change points with line references. Verified against `WorkflowRow` (lines 35-45), `createWorkflow()` INSERT (lines 181-194), and `rowToWorkflow()` (lines 118-142) in `space-workflow-repository.ts`.

5. **Zod `workflowConditionSchema` missing `task_result`** (was P2 R4): Task 1.1 subtask 3 instructs adding `'task_result'` to the enum at line ~38 of `export-format.ts`. Verified — the actual enum at line 38 is `z.enum(['always', 'human', 'condition'])`, and adding `'task_result'` is the correct fix.

6. **Export/import round-trip test** (was P2 R4): Task 5.4 subtask 2 now includes the round-trip bullet covering `isCyclic` preservation and `task_result` condition preservation.

7. **Prompt clarity: `step_result` vs `report_result.status`** (was P2 R4): Task 1.3 subtask 3 now includes an "Important distinction" paragraph with explicit separation of the two concepts. Good.

8. **`humanApproved` re-set in iteration 2** (was P3 R4): Task 5.5 step 3 now explicitly calls out `clearHumanApproval()` behavior.

---

## P1 — Remaining Issue

### `exportedWorkflowTransitionSchema` Zod schema is not updated for `isCyclic`

**File:** `docs/plans/.../01-task-result-condition-type.md`, Task 1.1 subtask 3  
**Actual file affected:** `packages/daemon/src/lib/space/export-format.ts`, lines 60–65

Task 1.1 subtask 3 correctly instructs:
- Adding `isCyclic?: boolean` to the `ExportedWorkflowTransition` TypeScript interface (in `space.ts`)
- Updating the export logic to include `if (t.isCyclic !== undefined) exported.isCyclic = t.isCyclic;`
- Adding `'task_result'` to `workflowConditionSchema`'s enum

What it does **not** instruct is updating the `exportedWorkflowTransitionSchema` Zod object at lines 60–65 of `export-format.ts`:

```ts
const exportedWorkflowTransitionSchema = z.object({
  fromStep: z.string().min(1),
  toStep: z.string().min(1),
  condition: workflowConditionSchema.optional(),
  order: z.number().int().optional(),
  // isCyclic is MISSING — not mentioned in the plan
});
```

During `validateExportedWorkflow()` (and therefore `validateExportBundle()`), the Zod parser runs `exportedWorkflowTransitionSchema.safeParse()` on each transition. Since `isCyclic` is not in the Zod schema, it will be silently stripped from the parsed result (Zod strips unknown keys by default). The returned `ExportedSpaceWorkflow.transitions` will lose `isCyclic: true`, breaking the round-trip.

The Task 5.4 export/import round-trip test would then fail because `isCyclic` is absent after re-import — but only if the test correctly asserts on the imported value. If the test is written as `validateExportedWorkflow(exportedData)` followed by checking the returned value, the bug manifests as a test failure. If the test is written incorrectly (checking the input object rather than the Zod-parsed output), it would pass while the bug persists in production.

**Fix:** Task 1.1 subtask 3 must explicitly instruct adding `isCyclic: z.boolean().optional()` to `exportedWorkflowTransitionSchema`.

---

## Minor Observations (Not Blocking)

- The `workflowConditionSchema`'s `superRefine` only validates that `condition` type requires a non-empty expression. After adding `'task_result'` to the enum, the schema would accept a `task_result` condition with no `expression` (no Zod error), only failing at runtime in `evaluateCondition()`. This is consistent with how `expression` is optional for `human` and `always` types, and the plan's runtime validation in Task 1.2 handles it, so this is not a blocker.

- The `createWorkflow()` INSERT at line 181–182 of `space-workflow-repository.ts` uses a hardcoded column list without `max_iterations`. Task 2.1 subtask 8 explicitly calls this out — good, the coder is correctly directed.

- The `advance()` method signature change from `async advance()` to `async advance(options?: { stepResult?: string })` and the subsequent use of `options.stepResult` as fallback in Task 1.2 subtask 5 are consistent with how `clearHumanApproval()` in `workflow-executor.ts` (lines 416-424) interacts with `this.run`. No new concurrency issues.

---

## Summary

The revision correctly resolves all eight issues from round 4. One new P1 gap remains: `exportedWorkflowTransitionSchema` at `export-format.ts:60–65` is not instructed to add `isCyclic?: z.boolean().optional()`, which means Zod will silently strip `isCyclic` from imported workflows and break the round-trip test. This is a single-line fix to the plan's Task 1.1 subtask 3.

lsm (commented):

**Recommendation: APPROVE**

The final fix correctly addresses the last P1 issue. Task 1.1 subtask 3 now explicitly instructs adding `isCyclic: z.boolean().optional()` to the `exportedWorkflowTransitionSchema` Zod object. I verified the fix against the actual source:

- `packages/daemon/src/lib/space/export-format.ts` lines 60-65 define `exportedWorkflowTransitionSchema` with `fromStep`, `toStep`, `condition`, `order` -- but no `isCyclic`. The plan correctly identifies this gap.
- Zod's default parsing behavior strips unknown keys, so without this addition, `isCyclic` would be silently dropped during import, breaking the export/import round-trip that Task 5.4 explicitly tests.
- The subtask is now clearly structured as two bullet points under the Zod schemas update: (1) `task_result` in the condition enum, (2) `isCyclic` in the transition schema. Both are necessary and correctly specified.

The plan as a whole is thorough and well-structured across all 5 milestones and 13 tasks. Key strengths:

- Explicit `isCyclic` flag instead of heuristic cycle detection avoids DAG merge path false positives
- Clear migration number assignment (34-38) with ordering constraints documented
- DB persistence coverage is complete: types, repository methods, INSERT/UPDATE statements, row-to-model mapping
- Test coverage in Milestone 5 includes unit tests, repository round-trip, export/import round-trip, and a full integration test with real DB
- The `step_result` fallback semantics are well-documented with clear rationale for when fallback is used

No remaining issues found. The plan is ready for implementation.

lsm (commented):

**Recommendation: APPROVE** — The P1 issue from the last round is fully resolved. The plan is ready for implementation.

---

## P1 Resolution: exportedWorkflowTransitionSchema Updated

The targeted fix is confirmed correct. Task 1.1 subtask 3 in `docs/plans/workflow-iteration-loop-agent-driven-cyclic-transitions-with/01-task-result-condition-type.md` now explicitly instructs:

> "The `exportedWorkflowTransitionSchema` at lines ~60-65 must add `isCyclic: z.boolean().optional()`. Without this, Zod will silently strip `isCyclic` from imported transitions (Zod strips unknown keys by default), breaking the export/import round-trip."

I verified this against the actual codebase. The `exportedWorkflowTransitionSchema` at lines 60-65 of `packages/daemon/src/lib/space/export-format.ts` currently reads:

```ts
const exportedWorkflowTransitionSchema = z.object({
  fromStep: z.string().min(1),
  toStep: z.string().min(1),
  condition: workflowConditionSchema.optional(),
  order: z.number().int().optional(),
  // isCyclic missing -- exactly as the plan identified
});
```

The instruction to add `isCyclic: z.boolean().optional()` is present, explicit, and correctly explains the consequence of omitting it (Zod silently strips unknown keys). The plan also correctly retains the `workflowConditionSchema` enum fix (`'task_result'` addition) and the `ExportedWorkflowTransition` TypeScript type fix in the same subtask.

---

## All Prior Issues -- Verified Still Resolved

Every issue resolved in prior rounds remains correctly addressed in the current plan text. No regressions were introduced by the targeted fix.

---

## Summary

The single P1 gap that blocked approval in the last round has been addressed with a single, precise sentence added to Task 1.1 subtask 3. The plan is thorough, well-structured, and provides clear implementation guidance across all 13 tasks and 5 milestones. It is ready for a coder agent to begin implementation.

PR #554, opened 2026-03-21
fix: remove stale sessions from RoomContextPanel on session delete/update

lsm:
When a session is deleted or archived from within a room, the RoomContextPanel
would continue showing it because RoomStore had no session lifecycle event handlers.

Daemon changes:
- session.delete: capture roomId before deletion, broadcast session.deleted with
  roomId on both global and room:{roomId} channels
- session.update: also broadcast session.updated on room:{roomId} channel when
  the session belongs to a room
- session.archive: broadcast session.updated with status:'archived' on session
  and room channels after archiving (previously no event was emitted)

Web changes (room-store.ts):
- Subscribe to session.deleted on room channel — removes session from sessions signal
- Subscribe to session.updated on room channel — updates title/status/lastActiveAt in-place

Tests:
- 12 new Vitest unit tests covering session.deleted and session.updated handlers
  in RoomStore, including roomId filtering and subscription cleanup
- Updated and extended existing daemon session-handlers.test.ts with room-channel
  broadcast assertions

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
@@ -161,31 +161,52 @@ export function setupSessionHandlers(
 			sessionId: string;
 		};
 
+		// Get roomId before updating to include in event payload
+		const agentSessionForUpdate = sessionManager.getSession(targetSessionId);
+		const roomIdForUpdate = agentSessionForUpdate?.getSessionData().context?.roomId;
+
 		// Convert UpdateSessionRequest to Partial<Session>
 		// config in UpdateSessionRequest is Partial<SessionConfig>, which is handled by
 		// database.updateSession merging with existing config
 		await sessionManager.updateSession(targetSessionId, updates as Partial<Session>);
 
-		// Broadcast update event to all clients
-		messageHub.event('session.updated', updates, {
+		const updatedPayload = { ...updates, sessionId: targetSessionId, roomId: roomIdForUpdate };
+
+		// Broadcast update event on session channel for per-session subscribers
+		messageHub.event('session.updated', updatedPayload, {
 			channel: `session:${targetSessionId}`,
 		});
+
+		// Also broadcast on room channel so RoomStore can react
+		if (roomIdForUpdate) {
+			messageHub.event('session.updated', updatedPayload, {
+				channel: `room:${roomIdForUpdate}`,
+			});
+		}
 
 		return { success: true };
 	});
 
 	messageHub.onRequest('session.delete', async (data, _ctx) => {
 		const { sessionId: targetSessionId } = data as { sessionId: string };
+
+		// Get roomId before deleting so we can include it in the event payload
+		const agentSessionForDelete = sessionManager.getSession(targetSessionId);
+		const roomIdForDelete = agentSessionForDelete?.getSessionData().context?.roomId;
+
 		await sessionManager.deleteSession(targetSessionId);
 
-		// Broadcast deletion event to all clients
-		messageHub.event(
-			'session.deleted',
-			{ sessionId: targetSessionId },
-			{
-				channel: 'global',
-			}
-		);
+		const deletedPayload = { sessionId: targetSessionId, roomId: roomIdForDelete };
+
+		// Broadcast deletion event on global channel (backward compat)
+		messageHub.event('session.deleted', deletedPayload, { channel: 'global' });
+
+		// Also broadcast on room channel so RoomStore can react immediately
+		if (roomIdForDelete) {
+			messageHub.event('session.deleted', deletedPayload, {
+				channel: `room:${roomIdForDelete}`,
+			});
+		}
 
 		return { success: true };
 	});
@@ -227,6 +248,21 @@ export function setupSessionHandlers(
 				archivedAt: new Date().toISOString(),
 				metadata: updatedMetadata,
 			} as Partial<Session>);
+
+			// Broadcast session.updated so RoomStore and session subscribers stay in sync
+			const archivedPayloadNoWorktree = {
+				sessionId: targetSessionId,
+				status: 'archived',
+				roomId: session.context?.roomId,
+			};
+			messageHub.event('session.updated', archivedPayloadNoWorktree, {
+				channel: `session:${targetSessionId}`,
+			});
+			if (session.context?.roomId) {
+				messageHub.event('session.updated', archivedPayloadNoWorktree, {
+					channel: `room:${session.context.roomId}`,
+				});
+			}
 
 			return { success: true, requiresConfirmation: false };
 		}
@@ -272,6 +308,21 @@ export function setupSessionHandlers(
 				worktree: undefined,
 				metadata: updatedMetadata,
 			} as Partial<Session>);
+
+			// Broadcast session.updated so RoomStore and session subscribers stay in sync
+			const archivedPayloadWithWorktree = {
+				sessionId: targetSessionId,
+				status: 'archived',
+				roomId: session.context?.roomId,
+			};
+			messageHub.event('session.updated', archivedPayloadWithWorktree, {
+				channel: `session:${targetSessionId}`,
+			});
+			if (session.context?.roomId) {
+				messageHub.event('session.updated', archivedPayloadWithWorktree, {
+					channel: `room:${session.context.roomId}`,
+				});
+			}
 
 			return {
 				success: true,
--- a/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
@@ -196,16 +196,15 @@ export function setupSessionHandlers(
 
 		await sessionManager.deleteSession(targetSessionId);
 
-		const deletedPayload = { sessionId: targetSessionId, roomId: roomIdForDelete };
-
-		// Broadcast deletion event on global channel (backward compat)
-		messageHub.event('session.deleted', deletedPayload, { channel: 'global' });
-
-		// Also broadcast on room channel so RoomStore can react immediately
+		// Broadcast on room channel so RoomStore reacts immediately.
+		// Note: the global channel broadcast is handled by session-lifecycle.ts / state-manager.ts
+		// to avoid triple-firing the event. We only add the room-scoped broadcast here.
 		if (roomIdForDelete) {
-			messageHub.event('session.deleted', deletedPayload, {
-				channel: `room:${roomIdForDelete}`,
-			});
+			messageHub.event(
+				'session.deleted',
+				{ sessionId: targetSessionId, roomId: roomIdForDelete },
+				{ channel: `room:${roomIdForDelete}` }
+			);
 		}
 
 		return { success: true };
--- a/packages/daemon/tests/unit/rpc/session-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/session-handlers.test.ts
@@ -540,15 +540,27 @@ describe('Session RPC Handlers', () => {
 			expect(result).toEqual({ success: true });
 		});
 
-		it('broadcasts session.deleted event on global channel with roomId in payload', async () => {
+		it('broadcasts session.deleted only on room channel (not global) when session has roomId', async () => {
 			const handler = messageHubData.handlers.get('session.delete');
 			expect(handler).toBeDefined();
 
+			const { agentSession } = createMockAgentSession({
+				context: { roomId: 'room-abc' },
+			} as Partial<AgentSession>);
+			sessionManagerData.mocks.getSession.mockReturnValueOnce(agentSession);
+
 			await handler!({ sessionId: 'session-123' }, {});
 
+			// Verify room-channel broadcast fires
 			expect(messageHubData.hub.event).toHaveBeenCalledWith(
 				'session.deleted',
-				expect.objectContaining({ sessionId: 'session-123' }),
+				expect.objectContaining({ sessionId: 'session-123', roomId: 'room-abc' }),
+				{ channel: 'room:room-abc' }
+			);
+			// Global broadcast must NOT come from this handler (session-lifecycle.ts handles it)
+			expect(messageHubData.hub.event).not.toHaveBeenCalledWith(
+				'session.deleted',
+				expect.anything(),
 				{ channel: 'global' }
 			);
 		});
@@ -571,7 +583,7 @@ describe('Session RPC Handlers', () => {
 			);
 		});
 
-		it('does not broadcast on room channel when session has no roomId', async () => {
+		it('does not broadcast any event when session has no roomId', async () => {
 			const handler = messageHubData.handlers.get('session.delete');
 			expect(handler).toBeDefined();
 
@@ -582,8 +594,8 @@ describe('Session RPC Handlers', () => {
 			await handler!({ sessionId: 'session-123' }, {});
 
 			const after = (messageHubData.hub.event as ReturnType<typeof mock>).mock.calls.length;
-			// Only one broadcast (global), no room channel broadcast
-			expect(after - before).toBe(1);
+			// No broadcasts from RPC handler; global broadcast is handled by session-lifecycle.ts
+			expect(after - before).toBe(0);
 		});
 	});
 
@@ -614,6 +626,67 @@ describe('Session RPC Handlers', () => {
 			await expect(handler!({ sessionId: 'non-existent' }, {})).rejects.toThrow(
 				'Session not found'
 			);
+		});
+
+		it('broadcasts session.updated with status archived on session channel after archiving', async () => {
+			const handler = messageHubData.handlers.get('session.archive');
+			expect(handler).toBeDefined();
+
+			const { agentSession } = createMockAgentSession({
+				worktree: undefined,
+			} as Partial<AgentSession>);
+			sessionManagerData.mocks.getSessionAsync.mockResolvedValueOnce(agentSession);
+
+			await handler!({ sessionId: 'session-123' }, {});
+
+			expect(messageHubData.hub.event).toHaveBeenCalledWith(
+				'session.updated',
+				expect.objectContaining({ sessionId: 'session-123', status: 'archived' }),
+				{ channel: 'session:session-123' }
+			);
+		});
+
+		it('broadcasts session.updated on room channel when session belongs to a room', async () => {
+			const handler = messageHubData.handlers.get('session.archive');
+			expect(handler).toBeDefined();
+
+			const { agentSession } = createMockAgentSession({
+				worktree: undefined,
+				context: { roomId: 'room-abc' },
+			} as Partial<AgentSession>);
+			sessionManagerData.mocks.getSessionAsync.mockResolvedValueOnce(agentSession);
+
+			await handler!({ sessionId: 'session-123' }, {});
+
+			expect(messageHubData.hub.event).toHaveBeenCalledWith(
+				'session.updated',
+				expect.objectContaining({
+					sessionId: 'session-123',
+					status: 'archived',
+					roomId: 'room-abc',
+				}),
+				{ channel: 'room:room-abc' }
+			);
+		});
+
+		it('does not broadcast on room channel when session has no roomId', async () => {
+			const handler = messageHubData.handlers.get('session.archive');
+			expect(handler).toBeDefined();
+
+			// Default mock session has no context.roomId
+			const { agentSession } = createMockAgentSession({
+				worktree: undefined,
+			} as Partial<AgentSession>);
+			sessionManagerData.mocks.getSessionAsync.mockResolvedValueOnce(agentSession);
+
+			const before = (messageHubData.hub.event as ReturnType<typeof mock>).mock.calls.length;
+			await handler!({ sessionId: 'session-123' }, {});
+			const after = (messageHubData.hub.event as ReturnType<typeof mock>).mock.calls.length;
+
+			// Only one broadcast: session:{id} channel; no room channel
+			expect(after - before).toBe(1);
+			const calls = (messageHubData.hub.event as ReturnType<typeof mock>).mock.calls.slice(before);
+			expect(calls[0][2]).toEqual({ channel: 'session:session-123' });
 		});
 	});
 
--- a/packages/daemon/tests/unit/rpc/session-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/session-handlers.test.ts
@@ -497,15 +497,35 @@ describe('Session RPC Handlers', () => {
 			expect(result).toEqual({ success: true });
 		});
 
-		it('broadcasts session.updated event', async () => {
+		it('broadcasts session.updated event on session channel', async () => {
 			const handler = messageHubData.handlers.get('session.update');
 			expect(handler).toBeDefined();
 
 			await handler!({ sessionId: 'session-123', title: 'New Title' }, {});
 
-			expect(messageHubData.hub.event).toHaveBeenCalledWith('session.updated', expect.any(Object), {
-				channel: 'session:session-123',
-			});
+			expect(messageHubData.hub.event).toHaveBeenCalledWith(
+				'session.updated',
+				expect.objectContaining({ sessionId: 'session-123', title: 'New Title' }),
+				{ channel: 'session:session-123' }
+			);
+		});
+
+		it('also broadcasts session.updated on room channel when session has a roomId', async () => {
+			const handler = messageHubData.handlers.get('session.update');
+			expect(handler).toBeDefined();
+
+			const { agentSession } = createMockAgentSession({
+				context: { roomId: 'room-abc' },
+			} as Partial<AgentSession>);
+			sessionManagerData.mocks.getSession.mockReturnValueOnce(agentSession);
+
+			await handler!({ sessionId: 'session-123', title: 'New Title' }, {});
+
+			expect(messageHubData.hub.event).toHaveBeenCalledWith(
+				'session.updated',
+				expect.objectContaining({ sessionId: 'session-123', roomId: 'room-abc' }),
+				{ channel: 'room:room-abc' }
+			);
 		});
 	});
 
@@ -520,7 +540,7 @@ describe('Session RPC Handlers', () => {
 			expect(result).toEqual({ success: true });
 		});
 
-		it('broadcasts session.deleted event', async () => {
+		it('broadcasts session.deleted event on global channel with roomId in payload', async () => {
 			const handler = messageHubData.handlers.get('session.delete');
 			expect(handler).toBeDefined();
 
@@ -528,9 +548,42 @@ describe('Session RPC Handlers', () => {
 
 			expect(messageHubData.hub.event).toHaveBeenCalledWith(
 				'session.deleted',
-				{ sessionId: 'session-123' },
+				expect.objectContaining({ sessionId: 'session-123' }),
 				{ channel: 'global' }
 			);
+		});
+
+		it('also broadcasts session.deleted on room channel when session has a roomId', async () => {
+			const handler = messageHubData.handlers.get('session.delete');
+			expect(handler).toBeDefined();
+
+			const { agentSession } = createMockAgentSession({
+				context: { roomId: 'room-abc' },
+			} as Partial<AgentSession>);
+			sessionManagerData.mocks.getSession.mockReturnValueOnce(agentSession);
+
+			await handler!({ sessionId: 'session-123' }, {});
+
+			expect(messageHubData.hub.event).toHaveBeenCalledWith(
+				'session.deleted',
+				expect.objectContaining({ sessionId: 'session-123', roomId: 'room-abc' }),
+				{ channel: 'room:room-abc' }
+			);
+		});
+
+		it('does not broadcast on room channel when session has no roomId', async () => {
+			const handler = messageHubData.handlers.get('session.delete');
+			expect(handler).toBeDefined();
+
+			// Default mock session has no context.roomId
+			const calls = (messageHubData.hub.event as ReturnType<typeof mock>).mock.calls;
+			const before = calls.length;
+
+			await handler!({ sessionId: 'session-123' }, {});
+
+			const after = (messageHubData.hub.event as ReturnType<typeof mock>).mock.calls.length;
+			// Only one broadcast (global), no room channel broadcast
+			expect(after - before).toBe(1);
 		});
 	});
 
--- a/packages/daemon/tests/unit/rpc/session-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/session-handlers.test.ts
@@ -565,24 +565,6 @@ describe('Session RPC Handlers', () => {
 			);
 		});
 
-		it('also broadcasts session.deleted on room channel when session has a roomId', async () => {
-			const handler = messageHubData.handlers.get('session.delete');
-			expect(handler).toBeDefined();
-
-			const { agentSession } = createMockAgentSession({
-				context: { roomId: 'room-abc' },
-			} as Partial<AgentSession>);
-			sessionManagerData.mocks.getSession.mockReturnValueOnce(agentSession);
-
-			await handler!({ sessionId: 'session-123' }, {});
-
-			expect(messageHubData.hub.event).toHaveBeenCalledWith(
-				'session.deleted',
-				expect.objectContaining({ sessionId: 'session-123', roomId: 'room-abc' }),
-				{ channel: 'room:room-abc' }
-			);
-		});
-
 		it('does not broadcast any event when session has no roomId', async () => {
 			const handler = messageHubData.handlers.get('session.delete');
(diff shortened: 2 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped fix for the stale session bug in RoomContextPanel. The approach is sound: capture roomId before mutation, broadcast on both the existing channel and the new `room:{roomId}` channel, and subscribe in `RoomStore` with roomId filtering. Tests are thorough with 12 new Vitest tests and updated daemon tests.

### Issues Found

**P2: `getSession` (sync/cache-only) may miss roomId for uncached sessions**

In `packages/daemon/src/lib/rpc-handlers/session-handlers.ts` lines 165 and 194, the code uses `sessionManager.getSession(targetSessionId)` which is the synchronous cache-only lookup. If a session exists only in the database and hasn't been loaded into cache (e.g., after a server restart, or a session that was never accessed in this server lifecycle), `getSession` returns `null` and `roomIdForUpdate`/`roomIdForDelete` will be `undefined`, silently skipping the room-channel broadcast.

The `session.archive` handler at line 220 correctly uses `getSessionAsync` which loads from DB. The `session.update` and `session.delete` handlers should do the same for consistency:
```
const agentSessionForUpdate = await sessionManager.getSessionAsync(targetSessionId);
```
In practice, sessions being updated/deleted are usually active and cached, so this is unlikely to cause issues in normal usage -- but it is an inconsistency worth addressing.

**P3: Duplicate comment numbering in room-store.ts**

`packages/web/src/lib/room-store.ts` lines 364 and 403 both use `// 5.` as the section comment. The session lifecycle events block should be `// 5.` and the fetch initial state should be `// 6.`

### What looks good

- Capturing roomId before the mutation (delete/archive) is the correct pattern -- data is gone after deletion.
- The room-store `session.updated` handler correctly uses conditional spreading to only apply defined fields, preserving existing values.
- Cleanup of subscriptions is properly handled via `cleanupFunctions`.
- The `roomId` defense-in-depth check in the frontend handlers is a good pattern even though channel routing already scopes delivery.
- Test coverage is comprehensive: deletion, update (title/status/lastActiveAt), cross-room filtering, unknown sessions, sequential deletions, and subscription cleanup.
- Backward compatibility maintained by keeping the global channel broadcast for `session.deleted`.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The implementation correctly solves the core bug but has two issues worth addressing before merging.

---

## Summary

The fix subscribes `RoomStore` to room-scoped `session.deleted` and `session.updated` events, and ensures the daemon broadcasts those events on the `room:{roomId}` channel. The approach is correct and the happy path works. However, there is a redundant global broadcast introduced in the RPC handler that conflicts with existing broadcast logic in `session-lifecycle.ts` and `state-manager.ts`, and the `session.archive` new behavior is untested in the daemon test suite.

---

## Issues Found

### P1 — Redundant global `session.deleted` broadcast (triple-fire on single delete)

**File:** `packages/daemon/src/lib/rpc-handlers/session-handlers.ts`, lines 201–208

The `session.delete` RPC handler now broadcasts `session.deleted` on the `global` channel **in addition to** the two existing broadcasts that already fire through the existing machinery:

1. `session-lifecycle.ts` PHASE 5 (line 510–513): broadcasts `{ sessionId, reason: 'deleted' }` on `global`
2. `state-manager.ts` line 148: listens on EventBus `session.deleted` and re-broadcasts `{ sessionId }` on `global`
3. The new handler (line 202): broadcasts `{ sessionId, roomId }` on `global`

After a single `session.delete` RPC call, connected clients receive three `session.deleted` events on the global channel. While currently harmless (the only subscriber in room-store correctly guards on `event.roomId === roomId`), this:
- Violates the existing broadcast architecture (the RPC handler is not supposed to be the canonical broadcast source for lifecycle events)
- Creates confusing payload inconsistency across the three events (one has `reason`, one has neither, one has `roomId`)
- Could cause subtle issues if future subscribers don't guard against duplicates

The room-channel broadcast (`room:{roomId}`) is the right new addition and should be kept. The **global broadcast added in the handler should be removed** — the existing lifecycle code already covers it.

### P1 — `session.archive` room-channel broadcast is not tested in the daemon suite

**File:** `packages/daemon/tests/unit/rpc/session-handlers.test.ts`, `session.archive` describe block (lines 590–618)

The `session.archive` describe block only has two tests:
- "archives session without worktree" — verifies `{ success: true, requiresConfirmation: false }`
- "throws error when session not found"

Neither test verifies:
1. That `session.updated` is broadcast on the `session:{id}` channel after archiving
2. That `session.updated` is broadcast on the `room:{roomId}` channel when the session belongs to a room
3. That no room-channel broadcast fires for a session without a room

This is specifically new behavior added in this PR (lines 252–265 and 312–325 of `session-handlers.ts`). The `session.update` and `session.delete` handlers have good broadcast tests added; `session.archive` should have the same coverage for consistency and regression safety.

### P2 — Duplicate section number in `room-store.ts` comment

**File:** `packages/web/src/lib/room-store.ts`, lines 364 and 403

Both the new session lifecycle subscriptions and the existing "Fetch initial state via RPC" block are labeled `// 5.`. The new block should be `// 6.` to maintain the numbered sequence.

### P2 — `session.deleted` fires on `global` channel without `roomId` for two of the three broadcasts

**Files:** `packages/daemon/src/lib/session/session-lifecycle.ts` line 510, `packages/daemon/src/lib/state-manager.ts` line 148

These pre-existing global broadcasts emit `session.deleted` without `roomId` in the payload. The new `room-store.ts` handler is correctly guarded (`if (event.roomId === roomId)`), so those no-`roomId` events are silently ignored. This is fine as-is, but it means the existing global broadcasts are now dead code from `RoomStore`'s perspective. This is worth a follow-up cleanup note even if not blocking.

---

## What Works Correctly

- The core fix is correct: `session.deleted` and `session.updated` are now broadcast on the `room:{roomId}` channel when a session belongs to a room
- The `room-store.ts` subscriptions correctly guard by `roomId` to prevent cross-room contamination
- Subscription cleanup on room switch is properly tested (the `describe('subscription cleanup')` block)
- The 12 new Vitest tests in `room-store-session-events.test.ts` are well-structured and cover the key filtering scenarios including same-room events, wrong-room events, unknown sessionId events, and multi-delete sequences
- The `session.update` and `session.delete` daemon tests are updated with appropriate room-channel broadcast assertions
- Pre-fetching `roomId` from the session cache before deletion (to capture it before the record disappears) is the right approach

lsm (commented):

**Recommendation: APPROVE**

The worker addressed all the previous review feedback correctly. The implementation is solid and the core bug fix is complete.

### Changes since last review

1. **Removed redundant global broadcast from `session.delete` handler** -- This was the P1 from the previous review. The handler now only broadcasts on `room:{roomId}` channel, leaving the global broadcast to `session-lifecycle.ts` / `state-manager.ts`. Verified by negative assertion at line 560-565 of the test file.

2. **Added daemon tests for `session.archive` broadcasts** -- Three new tests covering session channel broadcast, room channel broadcast when session has roomId, and no room broadcast when session has no roomId. This was the other P1.

3. **Fixed comment numbering in room-store.ts** -- Now correctly numbered 5 (session lifecycle) and 6 (fetch initial state). This was the P2.

### Previous P2 concern about `getSession` vs `getSessionAsync`

After deeper investigation, the `getSession` (sync) method at `packages/daemon/src/lib/session/session-cache.ts` line 44-69 is **not** cache-only -- it loads from DB synchronously and creates an AgentSession if not in cache. The only difference from `getSessionAsync` is locking behavior for concurrent access. Both methods will find a session that exists in the DB. The inconsistency is cosmetic rather than functional, and the current usage is appropriate for each handler's needs.

### Remaining P3 issues

**P3: Duplicate test in session.delete describe block**

`packages/daemon/tests/unit/rpc/session-handlers.test.ts` lines 543-565 and lines 568-584 test the same behavior -- that `session.deleted` is broadcast on the `room:{roomId}` channel when the session has a roomId. The second test ("also broadcasts session.deleted on room channel when session has a roomId") is a strict subset of the first test ("broadcasts session.deleted only on room channel (not global) when session has roomId"). This is harmless but redundant.

### Overall assessment

The fix correctly addresses the original bug: deleted/updated/archived room sessions now trigger room-scoped events that the `RoomStore` subscribes to, removing stale entries from the `RoomContextPanel`. The event subscription cleanup on room switch is properly handled. Test coverage is comprehensive with 12 new Vitest tests for the frontend store and updated/new daemon tests for all three handlers (update, delete, archive).

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one test-quality issue remains.

---

### Summary

The second round of fixes correctly resolves the previous P0/P1 issues: the redundant global broadcast has been removed from the `session.delete` handler, the three new `session.archive` broadcast tests are sound, and the comment numbering in `room-store.ts` is clean. The core feature (stale sessions removed from `RoomContextPanel`) is correctly implemented end-to-end. One P2 test-quality issue was introduced in this round.

---

### P2 — Duplicate test for `session.delete` room-channel broadcast

**File:** `packages/daemon/tests/unit/rpc/session-handlers.test.ts`, lines 543–584

Two tests in the `session.delete` describe block test exactly the same assertion:

- Line 543: `"broadcasts session.deleted only on room channel (not global) when session has roomId"`
- Line 568: `"also broadcasts session.deleted on room channel when session has a roomId"`

Both set up an agent session with `context: { roomId: 'room-abc' }`, invoke the handler, and assert `hub.event` was called with `{ channel: 'room:room-abc' }`. The second test is a strict subset of the first — it provides no additional coverage. It appears this was a leftover from refactoring the previous `"broadcasts session.deleted event"` test rather than replacing it cleanly.

The test at line 543 (which additionally asserts no global broadcast) should be the authoritative one; line 568 should be removed.

---

### Items verified as correctly resolved

1. **No global broadcast in `session.delete` handler** — the handler (lines 199-210) correctly emits only on `room:{roomId}` and includes a comment explaining that the global broadcast is delegated to `session-lifecycle.ts`. The "not global" assertion in the test at line 560-565 directly verifies this invariant.

2. **Three new `session.archive` broadcast tests** — the tests at lines 631-690 are correct and cover:
   - Session channel broadcast on archive without worktree
   - Room channel broadcast when session has `roomId`
   - No room channel broadcast when `roomId` is absent (count-based assertion is reliable here)

3. **Comment numbering in `room-store.ts`** — subscriptions are numbered 1-6 sequentially with no gaps or duplicates.

4. **`session.updated` on room channel** — both `session.update` and `session.archive` (both no-worktree and worktree paths) correctly broadcast on `room:{roomId}` only when `session.context.roomId` is present.

---

The three new `session.archive` broadcast tests only exercise the no-worktree code path. The worktree path (lines 282-335 in `session-handlers.ts`) has identical broadcast logic and is not tested for its room-channel behavior. This is a cosmetic gap — the logic is identical — but worth noting for completeness.

lsm (commented):

**Recommendation: APPROVE**

This is a final re-review pass following the removal of a duplicate test. The previous P2 issue has been resolved cleanly.

### What Was Fixed

The `session.delete` describe block in `/packages/daemon/tests/unit/rpc/session-handlers.test.ts` previously contained a redundant test at lines 568–584 that was a strict subset of the test at line 543. That test has been removed. The block now contains exactly three distinct, non-overlapping tests:

1. `deletes session successfully` — verifies the happy-path delete call and return value
2. `broadcasts session.deleted only on room channel (not global) when session has roomId` — verifies room-channel broadcast fires and global channel does not
3. `does not broadcast any event when session has no roomId` — verifies the no-room case emits no events

The total test count dropped from 73 to 72 as expected.

### Overall Assessment

The full implementation is correct and complete:

- **Daemon (`session-handlers.ts`):** `session.delete` captures `roomId` before deletion and broadcasts `session.deleted` on the `room:{roomId}` channel only (global is handled by `session-lifecycle.ts` to avoid double-firing). `session.update` and `session.archive` likewise broadcast on the room channel when a `roomId` is present.
- **Web (`room-store.ts`):** Both `session.deleted` and `session.updated` handlers are registered on the room channel, guarded by `event.roomId === roomId`, and are registered/unregistered correctly via `cleanupFunctions`.
- **Tests (`room-store-session-events.test.ts`):** 12 Vitest tests cover deletion, update, roomId filtering, no-op for unknown sessions, multi-step deletions, partial-update field preservation, and subscription cleanup on room deselect. Coverage is thorough.
- **Tests (`session-handlers.test.ts`):** 72 tests, all `session.delete`, `session.update`, and `session.archive` describe blocks are non-overlapping and cover the material behaviors.

lsm (commented):

**Recommendation: APPROVE**

### Summary

This PR correctly fixes the bug where deleted/archived room sessions continued to appear in the RoomContextPanel. The implementation is clean and well-structured.

### What was reviewed

**Daemon changes** (`packages/daemon/src/lib/rpc-handlers/session-handlers.ts`):
- `session.delete`: Captures `roomId` before deletion, broadcasts `session.deleted` on `room:{roomId}` channel only (global broadcast correctly delegated to `state-manager.ts` via EventBus)
- `session.update`: Now includes `roomId` in the payload and broadcasts `session.updated` on both `session:{id}` and `room:{roomId}` channels
- `session.archive`: Both no-worktree and worktree paths now broadcast `session.updated` with `status: 'archived'` on session and room channels

**Web changes** (`packages/web/src/lib/room-store.ts`):
- New `session.deleted` handler (lines 365-372) filters by `roomId` and removes the session from the signal
- New `session.updated` handler (lines 375-401) applies partial updates (title, status, lastActiveAt) in-place, filtering by `roomId`
- Both handlers are properly cleaned up via `cleanupFunctions`

**Tests:**
- 12 Vitest tests in `room-store-session-events.test.ts` covering deletion, update, partial update preservation, cross-room filtering, unknown session no-ops, and subscription cleanup
- 7 new daemon tests in `session-handlers.test.ts` covering room-channel broadcasts for update/delete/archive, and confirming no global broadcast from the RPC handler
- Total daemon test count is 72 (confirmed passing), web test count is 12 (confirmed passing)

### Verification

- Confirmed that the global `session.deleted` broadcast is handled by `state-manager.ts:148` (subscribes to EventBus `session.deleted` emitted by `session-lifecycle.ts:515`), so the RPC handler correctly only adds the room-scoped broadcast without triple-firing.
- The duplicate test that was removed in the latest commit was indeed a strict subset of the earlier test at line 543. Test count correctly went from 73 to 72.
- No issues found with the implementation.

lsm (commented):

### Overview

The approach of using `room.get` re-fetch instead of manual array splicing is conceptually sound -- it simplifies the client logic and provides self-healing for missed events during WebSocket reconnect gaps. The daemon-side changes (room-channel broadcasts for session.deleted, session.updated, and session.archive) are correct, properly scoped, and well-tested.

However, there is a significant performance issue with the current implementation.

---

### P1: `refresh()` on every `session.updated` causes excessive server load

**Files:** `packages/web/src/lib/room-store.ts` (lines 380-389), `packages/web/src/lib/room-store.ts` (lines 461-473, 404-437)

`refresh()` calls `fetchInitialState()` which makes **4 RPC calls** on every invocation: `room.get`, `goal.list` (via `fetchGoals`), `room.runtime.state`, and `room.runtime.models` (via `fetchRuntimeModels`).

The `session.updated` event fires on **every** `session.update` RPC call. This includes high-frequency operations like input draft persistence (`useInputDraft.ts` saves drafts on debounced keystrokes via `session.update`). When a user is typing in a session that belongs to a room, each draft save will:

1. `session.update` RPC -> daemon broadcasts `session.updated` on `room:{roomId}` channel
2. RoomStore receives the event, calls `this.refresh()`
3. `refresh()` fires 4 RPC calls to the server

If there are N sessions in a room and users are actively typing, this creates O(N * keystrokes) full room re-fetches per second. Even a single user typing would trigger a full room refresh every ~300ms (typical debounce interval).

**Suggested fix:** At minimum, add debouncing to the `refresh()` calls triggered by session events. A 500ms trailing debounce would coalesce rapid-fire events into a single fetch. Alternatively, consider a more targeted refresh that only calls `room.get` (not the full `fetchInitialState` with goals and runtime state), since session events only affect the sessions list.

---

### P2: Duplicated broadcast logic in archive handler

**File:** `packages/daemon/src/lib/rpc-handlers/session-handlers.ts` (lines 251-264 and 311-324)

The `session.archive` handler has two code paths (with worktree and without worktree), and each one contains an identical copy of the session.updated broadcast logic. This should be extracted into a shared helper or moved after the branching to avoid the duplication and the risk of the two copies diverging over time.

---

### P3: Tests use `setTimeout` for negative assertions

**File:** `packages/web/src/lib/__tests__/room-store-session-events.test.ts` (lines 178, 189, 226, 237)

The negative-assertion tests (verifying that NO re-fetch happens) use `await new Promise((r) => setTimeout(r, 20))` which is fragile. If the code under test uses a longer async chain, 20ms may not be enough. Consider using `vi.runAllTimers()` with fake timers or a slightly longer timeout to reduce flakiness.

---

### Summary

The daemon-side changes are solid and correctly scoped. The web-side approach is architecturally clean but needs debouncing before it can be safely merged -- without it, routine operations like typing in a room session will generate an excessive number of RPC round-trips.

lsm (commented):

> **Recommendation: REQUEST_CHANGES** — one P0 performance regression, one P1 missing behavior, plus a P2 test gap.

---

## Overview

The refactor correctly moves away from manual array splicing toward a server-authoritative re-fetch, which is a sound architectural choice for correctness and self-healing across reconnects. The daemon changes (room-channel broadcasts for `session.deleted` and `session.updated`) are well-scoped. However the implementation has a critical performance regression introduced by the `session.updated` subscription.

---

## P0 — `session.updated` causes a 4-RPC storm on every keypress in a room session

**Files:** `packages/web/src/lib/room-store.ts` lines 380–390, `packages/web/src/hooks/useInputDraft.ts` lines 140–154, `packages/daemon/src/lib/rpc-handlers/session-handlers.ts` lines 180–185

### The chain

1. `useInputDraft` calls `session.update` every 250 ms while the user is typing in any session that belongs to a room.
2. `session.update` handler (line 181–185 of session-handlers.ts) reads the session's `context.roomId` and broadcasts `session.updated` on the `room:{roomId}` channel unconditionally — including for metadata-only changes like `inputDraft`.
3. `RoomStore` receives this event and calls `this.refresh()`.
4. `refresh()` calls `fetchInitialState()`, which issues **four** sequential/concurrent RPC calls: `room.get`, `goal.list`, `room.runtime.state`, and `room.runtime.models`.

**Result:** Typing a sentence in a room session fires 4 × (number_of_characters / debounce_interval) RPCs against the daemon — approximately 4 calls every 250 ms per active keystroke. On a room with an active runtime, this also hammers `goal.list` and the runtime-state endpoints for no reason.

This was not a problem in the previous version (array-splicing approach) because that only mutated local state without any RPC. The re-fetch approach regresses performance for the most common interactive case.

### Root cause

The daemon broadcasts `session.updated` on the room channel for **all** `session.update` calls, not just the semantically-meaningful ones (deletion, archival, status change). Other callers that will also trigger the storm include:
- Model switches (`session.model.switch` broadcasts on session channel only — this one is fine)
- Coordinator/sandbox/thinking-level switches
- Any future feature that writes to session metadata

### Fix direction

The subscribe-to-refresh approach requires that the room channel only receives `session.updated` events that are meaningful for the session list (status changes, title changes, archival). Consider one of:
- **Option A (preferred):** Gate the room-channel broadcast in `session.update` handler to only fire when the update touches fields visible in `SessionSummary` (title, status, etc.) — never for metadata updates.
- **Option B:** Add a debounce/deduplicate guard in `RoomStore.refresh()` so concurrent/rapid refresh calls collapse into one inflight request.
- **Option C:** Keep the refresh approach only for `session.deleted` and revert `session.updated` to the lightweight in-place merge (the old approach for updates was correct; only the delete case needed fixing).

---

## P1 — No navigation away when the currently-viewed session is deleted

**File:** `packages/web/src/islands/RoomContextPanel.tsx`

When a session the user is actively viewing gets deleted, `RoomStore.refresh()` will remove it from `sessions.value` and the panel re-renders without it — but `currentRoomSessionIdSignal` still points to the deleted session ID. The `ChatContainer` for that session will remain visible, showing a dead session.

The previous array-splicing PR had the same gap. This PR does not fix it, and the task description doesn't mention it, but the key concern listed in the re-review prompt ("what happens if the user is viewing a session that was just deleted") calls it out explicitly. The fix would be: after `refresh()` in the `session.deleted` handler, check whether the current session ID is still in `sessions.value`, and if not, navigate to the room dashboard.

---

## P2 — Tests do not cover the `useInputDraft` flooding scenario

**File:** `packages/web/src/lib/__tests__/room-store-session-events.test.ts`

The new tests verify that `session.updated` triggers a `room.get` re-fetch, but there is no test that:
- Fires `session.updated` rapidly in quick succession and verifies that refresh calls are bounded (debounced/deduplicated), or
- Fires `session.updated` with a metadata-only payload (like `inputDraft`) and verifies it does **not** trigger a re-fetch.

Without such a test, the P0 bug above would be undetected. Adding a negative test for metadata-only updates (or a flood test asserting collapse behavior) would catch regressions here.

---

## Minor observation — `fetchInitialState` is expensive as a refresh primitive

`refresh()` calls `fetchInitialState()`, which also re-fetches goals, runtime state, and runtime models. For a session-list refresh, only `room.get` is needed. Consider extracting a lightweight `refreshSessions()` that only calls `room.get` and updates `sessions.value` (and `tasks.value`) — leaving goals and runtime state on their own dedicated subscription paths.

---

## What is correct

- The daemon changes are surgically correct: `session.deleted` on the room channel before the session is removed from the DB, and `session.updated` after archive/update, are the right events to broadcast.
- The roomId-filter guard (`if (event.roomId === roomId)`) correctly ignores events for other rooms.
- Subscription cleanup on room deselect works and is tested.
- The session-handlers daemon tests are solid: they correctly verify room-channel vs. session-channel separation and the no-roomId no-broadcast case.
- The `session.deleted` case is clean and does not suffer from the flooding problem (session deletion is an infrequent, intentional operation).

lsm (commented):

### Summary

The two previously identified issues are resolved:

- **P0 (navigate away on delete):** Fixed. After `refresh()` completes, the handler checks `currentRoomSessionIdSignal.value` against `this.sessions.value` and calls `navigateToRoom(roomId)` when the active session is no longer present. Well-tested.
- **P1 (session.updated RPC storm):** Fixed by removing the `session.updated` subscription entirely. The comment at `packages/web/src/lib/room-store.ts:372-375` clearly explains the rationale. Tests verify no re-fetch happens for single or rapid-fire updates.

### Remaining Issue

**P2: Archived sessions not reflected in RoomContextPanel**

By removing the `session.updated` subscription entirely (`packages/web/src/lib/room-store.ts:366-391`), the RoomStore no longer reacts when a session is archived. The daemon's `session.archive` handler (`packages/daemon/src/lib/rpc-handlers/session-handlers.ts:251-264, 311-324`) emits `session.updated` with `status: 'archived'` on the room channel, but nobody is listening.

The archival flow in `packages/web/src/hooks/useSessionActions.ts:80-99` updates the global `sessionsSignal` but not `roomStore.sessions`. This means a session archived by one client (or the current client after navigating back) will still appear as "active" in the RoomContextPanel until the next room.get re-fetch (e.g., room switch or WebSocket reconnect).

Mitigation options (for a follow-up, not blocking):
1. Subscribe to `session.updated` only for `status` changes (filter out draft saves by checking if the payload contains `status`)
2. Have the `session.archive` handler emit `session.deleted` on the room channel instead of `session.updated`, since from the room panel's perspective an archived session should disappear
3. Call `roomStore.refresh()` after a successful archive in `useSessionActions`

This is a P2 because the user who archives will see a toast and typically navigate away; the staleness is only visible on return or to concurrent users. Not a regression from the original bug report (deleted sessions showing), but a regression from the intermediate implementation that handled `session.updated`.

### Tests

10 tests in `packages/web/src/lib/__tests__/room-store-session-events.test.ts` provide good coverage:
- `session.deleted` triggers re-fetch (matching room, different room, absent roomId)
- Navigation behavior (active session deleted, different session deleted, no active session)
- `session.updated` does NOT trigger re-fetch (single + rapid-fire)
- Subscription cleanup on room switch

7 new/updated tests in `packages/daemon/tests/unit/rpc/session-handlers.test.ts` verify:
- Room-channel broadcast on delete (not global)
- No broadcast when no roomId
- Archive broadcasts on both session and room channels

### Recommendation: APPROVE

The core bug (deleted sessions showing in RoomContextPanel) is fixed. The navigate-away behavior is correct. The archived-session staleness is a minor regression worth a follow-up ticket but not blocking.

lsm (commented):

**Recommendation: APPROVE** — All P0 and P1 issues from the previous review have been resolved. The implementation is clean, well-tested, and correct.

---

## Status of Previously Raised Issues

### P0: `session.updated` RPC storm — FIXED
The `session.updated` subscription was removed entirely from `room-store.ts`. The comment at line 370-372 correctly documents the reason (draft saves fire every ~250ms). The `session.deleted` handler now does a single `refresh()` call instead, which self-heals any missed events during reconnect gaps.

### P0: No navigation on active-session delete — FIXED
Lines 379-387 of `packages/web/src/lib/room-store.ts` correctly check `currentRoomSessionIdSignal.value` after `refresh()` resolves and call `navigateToRoom(roomId)` if the active session is no longer in the server-returned list. The ordering is correct: `sessions.value` is updated inside `refresh()` before the `.then()` callback runs.

### P1: Redundant global `session.deleted` broadcast — FIXED
The `session.delete` RPC handler now only broadcasts on the `room:{roomId}` channel (lines 202-208 of `session-handlers.ts`). The comment at line 200 explicitly notes that the global channel broadcast is handled by `session-lifecycle.ts`/`state-manager.ts`. The test at line 543 verifies both that the room-channel broadcast fires and that no global broadcast originates from this handler.

### P1: `session.archive` broadcast tests missing — FIXED
Four tests were added to the `session.archive` describe block (lines 613-671 of `session-handlers.test.ts`):
- session channel broadcast after archiving
- room channel broadcast when session belongs to a room
- no room-channel broadcast for sessions without a roomId
- the no-room case only fires exactly 1 event (session channel only)

### P2: Duplicate `// 5.` comment numbering — FIXED
`packages/web/src/lib/room-store.ts` now correctly numbers section comments as `// 5.` for session lifecycle events (line 366) and `// 6.` for fetch initial state (line 393).

---

## Remaining Observations (Non-blocking)

### Archived sessions remain visible until next deletion event (by design)
Since `session.updated` is no longer subscribed, archiving a session from within a room will not immediately remove it from the RoomContextPanel session list. The archived session stays visible until the user refreshes or another `session.deleted` event triggers a `refresh()`. This is an explicit trade-off documented in the code comment and is the correct decision given the draft-save storm risk. A follow-up could hook into the archive action's success path (in the component that calls `session.archive`) to call `roomStore.refresh()` directly, but that is out of scope here.

### `getSession` (sync/cache-only) for roomId lookup — low risk, pre-existing
Both `session.update` (line 165) and `session.delete` (line 194) use the synchronous `sessionManager.getSession()` to look up `roomId`. Sessions not yet loaded into cache will silently skip the room-channel broadcast. This was raised in the previous opus review and is pre-existing behavior — in practice all touched sessions are cached. Not blocking.

### Double global `session.deleted` broadcast — pre-existing
`session-lifecycle.ts` PHASE 5 (line 510) fires `session.deleted` on `global` directly, then emits on EventBus, and `state-manager.ts` (line 148) listens and re-fires. This double-fire predates this PR and is unchanged. `room-store.ts` correctly ignores global broadcasts because they lack `roomId`. Not a regression.

---

## What Looks Good

- 10 new Vitest unit tests in `room-store-session-events.test.ts` covering refresh triggers, navigation cases (3 scenarios), `session.updated` suppression (single + 5 rapid-fire), and subscription cleanup.
- 4 new daemon tests for `session.archive` broadcast coverage, plus 2 new daemon tests for `session.delete` broadcast correctness.
- The `refresh()` approach (re-fetch vs. manual splice) self-heals missed events across reconnect gaps and keeps local state authoritative to the server.
- The `roomId` guard in the `session.deleted` handler correctly scopes events to the current room even when channel routing is sufficient — defense in depth.
- Subscription cleanup test confirms no re-fetch fires after room deselect.

lsm (commented):

**Recommendation: APPROVE**

This PR correctly fixes the issue where deleted/archived room sessions continued to appear in the RoomContextPanel. The implementation is clean and well-tested.

### What was reviewed

**Daemon changes** (`packages/daemon/src/lib/rpc-handlers/session-handlers.ts`):
- `session.delete`: captures roomId before deletion, broadcasts `session.deleted` on `room:{roomId}` channel
- `session.update`: broadcasts `session.updated` on `room:{roomId}` channel when session belongs to a room
- `session.archive`: broadcasts `session.updated` with `status: 'archived'` on both session and room channels

**Web changes** (`packages/web/src/lib/room-store.ts`):
- `session.deleted` handler: calls `refresh()` to re-fetch authoritative session list, then navigates away if the user was viewing the deleted session
- `session.updated` handler: only calls `refresh()` when `event.status` is present, correctly filtering out draft saves (which carry only `metadata.inputDraft` / `title`)

### Assessment

**Correctness**: The `status` field guard (`if (event.status === undefined) return`) is the right approach. I verified that `useInputDraft` calls `session.update` with `{ metadata: { inputDraft: ... } }` -- no `status` field -- so draft saves are correctly suppressed. Archive events explicitly include `status: 'archived'` in the payload, so they pass the guard and trigger a refresh.

**Event routing**: The `roomId` guard (`if (event.roomId !== roomId) return`) correctly scopes handlers even though `hub.onEvent` receives events from all channels. Global `session.deleted` events from `state-manager.ts` lack `roomId`, so they are filtered out. Only the room-scoped event from the RPC handler is processed.

**Navigation on delete**: The post-refresh check (`!this.sessions.value.some(s => s.id === activeSessionId)`) is correct -- it uses the refreshed server data rather than optimistically assuming the deleted session is gone.

**Test coverage** (13 tests): Covers session.deleted triggers, navigation away on delete, session.updated with/without status, rapid draft saves, wrong-room suppression, and subscription cleanup. Tests are well-structured with proper async handling.

### Minor observations (P3)

1. **Duplicate session-channel broadcasts on archive**: The `session.archive` handler explicitly broadcasts `session.updated` on `session:{id}`, but `sessionManager.updateSession()` also triggers `session.updated` via the eventBus to state-manager pipeline. The `session:` channel gets two events. Not a correctness issue since the room-store only listens on the room channel, but worth noting.

2. **Unnecessary room-channel traffic for draft saves**: The `session.update` RPC handler broadcasts on `room:{roomId}` for ALL updates including draft saves. The room-store correctly ignores these (no `status` field), but they still consume WebSocket bandwidth. A server-side guard could avoid sending them. Low priority since the client-side guard handles it.

Both observations are minor and do not warrant blocking the PR.

lsm (commented):

**Recommendation: APPROVE**

This is a solid, well-reasoned fix for the `session.updated` subscription. The `status === undefined` guard is correct and the assumption about draft saves is valid. Here is my assessment.

---

### Guard Correctness: Verified

The guard at `packages/web/src/lib/room-store.ts:399`:

```ts
if (event.status === undefined) return;
```

This is correct because every code path that emits `session.updated` on a room channel has been audited:

1. **`session.update` RPC handler (line 182)** — spreads `UpdateSessionRequest` into the payload. `UpdateSessionRequest` is typed as `{ title?, workspacePath?, config? }`. Draft saves from `useInputDraft.ts` call this with only `{ metadata: { inputDraft } }` — no `status`. No caller of `session.update` passes `status` in the codebase.

2. **`session.archive` RPC handler (lines 257–323)** — explicitly constructs `{ sessionId, status: 'archived', roomId }` and broadcasts on the room channel. This correctly passes the guard and triggers a refresh.

3. **`session.setWorktreeMode` handler (line 103)** — only broadcasts on `session:{id}` channel, NOT on a room channel, so it never reaches the room subscriber at all. The full `Session` object it broadcasts does include `status: 'active'`, but since it never reaches the room-channel subscriber the guard is irrelevant here.

The guard's underlying assumption is structurally enforced: `UpdateSessionRequest` has no `status` field, and no call site adds one. The only room-channel broadcasts that carry `status` are the intentional `session.archive` emissions.

---

### Test Coverage: Complete

The 5 new `session.updated` tests (lines 289–374 of `room-store-session-events.test.ts`) cover all branches:

- Triggers refresh with `status` present
- Suppressed without `status` (draft save with title)
- Suppressed for rapid consecutive draft saves (`metadata.inputDraft`)
- Suppressed for wrong room
- Server response reflected after status change

The `vi.waitFor` usage in the positive cases and `setTimeout(r, 30)` in the negative cases are appropriate for the async patterns. Total of 13 tests confirmed.

---

### Minor Observation (P3 nit, not a blocker)

The comment in `room-store.ts` says "Draft saves (useInputDraft, ~250 ms debounce) only carry title/inputDraft". This is slightly imprecise — `useInputDraft` only sends `metadata`, not `title`. Title saves come from a separate code path via `api-helpers.ts`. In practice this does not affect correctness since neither carries `status`, but the comment could mislead a future reader.

---

**Overall:** The guard is correctly placed, structurally sound, and fully tested. The P2 gap identified by the prior reviewer is addressed. Approving.

PR #556, opened 2026-03-21
docs: plan to fix E2E test failures on dev branch CI

lsm:
## Summary

- Adds a structured plan to fix 4 failing E2E test suites on the `dev` branch CI
- Root causes identified: mismatched test selectors (task-actions-dropdown), outdated layout assertions (space-creation), SpaceIsland loading issues (visual-workflow-editor), and missing post-deletion navigation (worktree-isolation)
- Plan contains 4 independent tasks, one per root cause

## Test plan

- [ ] Review plan accuracy against current codebase
- [ ] Execute tasks 1-4 in parallel (they are independent)
- [ ] Verify all 4 E2E test suites pass on CI after fixes

### DIFF
--- a/docs/plans/fix-e2e-test-failures-on-dev-branch-ci.md
+++ b/docs/plans/fix-e2e-test-failures-on-dev-branch-ci.md
@@ -1,0 +1,173 @@
+# Fix E2E Test Failures on Dev Branch CI
+
+## Goal
+
+Fix four failing E2E test suites on the `dev` branch CI (run ID `23382023181`):
+
+1. `features-task-actions-dropdown` -- 9/10 tests fail
+2. `features-space-creation` -- 1/7 tests fail
+3. `features-visual-workflow-editor` -- 6/6 tests fail
+4. `features-worktree-isolation` -- 1/5 tests fail (LLM test)
+
+## Approach
+
+The failures group into four root causes. Each task below addresses one root cause. All PRs target `dev`.
+
+---
+
+## Task 1: Fix task-actions-dropdown E2E tests -- update selectors to match actual UI
+
+**Type:** coder
+
+**Description:**
+The `task-actions-dropdown.e2e.ts` tests were written expecting a three-dot dropdown menu with `data-testid="task-options-menu"`, but the actual `TaskView.tsx` uses inline action buttons:
+- `data-testid="task-cancel-button"` (Cancel button, visible for pending/in_progress/review tasks)
+- `data-testid="task-complete-button"` (Complete button, visible for in_progress tasks, hidden for review)
+
+Additionally, some tests transition tasks to `in_progress` which triggers worktree creation. In CI, the workspace is a temp directory (not a git repo), so worktree creation fails with "Worktree creation failed -- task requires isolation". The tests that need `in_progress` status should use `task.setStatus` RPC directly (which is already done in the test helper `createRoomAndTask`), and the test assertions should not depend on a worktree-spawned session.
+
+**Key files:**
+- `packages/e2e/tests/features/task-actions-dropdown.e2e.ts` -- the failing test file
+- `packages/web/src/components/room/TaskView.tsx` -- the actual UI (lines ~938-968 for action buttons, lines ~500-630 for confirmation modals)
+
+**Subtasks:**
+1. Run `bun install` at the worktree root.
+2. Read the current TaskView.tsx to understand the actual UI pattern (inline buttons, not dropdown menu).
+3. Rewrite tests that check for `task-options-menu` to instead use `task-cancel-button` and `task-complete-button`:
+   - "shows task options menu for pending task (cancel only)" -- assert `task-cancel-button` is visible, `task-complete-button` is not visible.
+   - "shows task options menu for in_progress task (complete + cancel)" -- assert both buttons visible.
+   - "does NOT show task options menu for completed task" -- this test already passes; keep as-is.
+   - "opens dropdown and shows Cancel Task item" -- replace dropdown menu interaction with direct click on `task-cancel-button`.
+   - "shows Mark as Complete for in_progress task" -- replace dropdown interaction with asserting `task-complete-button` visible.
+   - "opens cancel confirmation dialog on Cancel Task click" -- click `task-cancel-button` directly instead of opening dropdown.
+   - "opens complete confirmation dialog on Mark as Complete click" -- click `task-complete-button` directly.
+   - "can dismiss cancel dialog with Keep Task button" -- click `task-cancel-button` directly, rest stays same.
+   - "cancels task and navigates away on confirmation" -- click `task-cancel-button` directly, confirm, check navigation.
+   - "completes task and navigates away on confirmation" -- click `task-complete-button` directly, confirm, check navigation.
+4. For tests that wait for task status to appear in the UI after creating a task with `in_progress` status: ensure the test navigates to the task page AFTER the status transition is complete (the helper already does this via RPC `task.setStatus`). The page should render the correct buttons based on the task's current status from the database, not from a live worktree session.
+5. Run the test locally: `make run-e2e TEST=tests/features/task-actions-dropdown.e2e.ts`.
+6. Verify all tests pass.
+
+**Acceptance Criteria:**
+- All 10 tests in `task-actions-dropdown.e2e.ts` pass locally.
+- Tests use the actual UI selectors (`task-cancel-button`, `task-complete-button`, `cancel-task-confirm`, `complete-task-confirm`).
+- No tests rely on `data-testid="task-options-menu"` (which does not exist).
+- Changes must be on a feature branch with a GitHub PR created via `gh pr create` targeting `dev`.
+
+**Dependencies:** None
+
+---
+
+## Task 2: Fix space-creation E2E test -- update assertion to match actual space layout
+
+**Type:** coder
+
+**Description:**
+The test "creates space and shows 3-column layout" in `space-creation.e2e.ts` (line 122) expects to see `SpaceNavPanel` content ("No runs or tasks yet") after space creation. However, `SpaceNavPanel` is not used in the actual `SpaceIsland` component. The space view uses a tabbed layout (Dashboard, Agents, Workflows, Settings) without a left nav panel showing runs/tasks.
+
+After creating a space, the user is navigated to `/space/<id>` where `SpaceIsland` renders with the Dashboard tab active, showing the `SpaceDashboard` component which has "Quick Actions", "Start Workflow Run", and "Create Task" text.
+
+**Key files:**
+- `packages/e2e/tests/features/space-creation.e2e.ts` -- the failing test (line 122-159)
+- `packages/web/src/islands/SpaceIsland.tsx` -- actual space view component (tabbed layout)
+- `packages/web/src/components/space/SpaceDashboard.tsx` -- Dashboard tab content
+- `packages/web/src/components/space/SpaceNavPanel.tsx` -- unused component (not imported by SpaceIsland)
+
+**Subtasks:**
+1. Run `bun install` at the worktree root.
+2. Read SpaceIsland.tsx and SpaceDashboard.tsx to understand what is actually rendered when a space is loaded.
+3. Update the test assertions in "creates space and shows 3-column layout" to match the actual UI:
+   - Remove assertion for "No runs or tasks yet" (from SpaceNavPanel, not rendered).
+   - Keep or adjust assertions for "Quick Actions", "Start Workflow Run", "Create Task" -- these are in SpaceDashboard and should be visible on the Dashboard tab.
+   - Consider renaming the test to reflect the actual layout (tabbed, not 3-column).
+4. Run the test locally: `make run-e2e TEST=tests/features/space-creation.e2e.ts`.
+5. Verify all tests pass.
+
+**Acceptance Criteria:**
+- All tests in `space-creation.e2e.ts` pass locally.
+- Test assertions match the actual space layout (tabbed view with Dashboard).
+- Changes must be on a feature branch with a GitHub PR created via `gh pr create` targeting `dev`.
+
+**Dependencies:** None
+
+---
+
+## Task 3: Fix visual-workflow-editor E2E tests -- diagnose and fix SpaceIsland loading issues
+
+**Type:** coder
+
+**Description:**
+All 6 tests in `visual-workflow-editor.e2e.ts` fail. The primary failure is a 60-second timeout waiting for `text=Workflows` to appear on the page after navigating to `/space/<id>`. This means the `SpaceIsland` component is not rendering its tab bar. One test also fails with "No agents found in space" when calling `spaceAgent.list` via RPC.
+
+The test creates a space via RPC in `beforeEach` using the server's workspace root (a temp directory), then navigates to it. The SpaceIsland component calls `spaceStore.selectSpace(spaceId)` on mount, which calls `space.overview` RPC. If this fails or the store gets stuck in loading state, the tabs won't render.
+
+Potential causes to investigate:
+- The `space.overview` RPC may be failing for the newly created space (check if the handler returns the space correctly)
+- The `spaceStore` loading state may not resolve properly
+- The test's `navigateToSpace` function navigates directly to the URL but may not wait for the WebSocket to reconnect or the store to load
+- The "No agents found" error may indicate that `seedPresetAgents` is silently failing during space creation on non-git workspaces
+
+**Key files:**
+- `packages/e2e/tests/features/visual-workflow-editor.e2e.ts` -- failing tests
+- `packages/web/src/islands/SpaceIsland.tsx` -- space view component
+- `packages/web/src/lib/space-store.ts` -- space store with `selectSpace`/`doSelect`/`startSubscriptions`
+- `packages/daemon/src/lib/rpc-handlers/space-handlers.ts` -- `space.overview` RPC handler
+- `packages/daemon/src/lib/space/agents/seed-agents.ts` -- agent seeding on space creation
+- `packages/daemon/src/lib/rpc-handlers/space-agent-handlers.ts` -- `spaceAgent.list` RPC handler
+
+**Subtasks:**
+1. Run `bun install` at the worktree root.
+2. Run the visual-workflow-editor test locally to reproduce the failure: `make run-e2e TEST=tests/features/visual-workflow-editor.e2e.ts`.
+3. Investigate whether the `space.overview` RPC returns the space data correctly for a newly created space. Check the handler in `space-handlers.ts`.
+4. Investigate whether `seedPresetAgents` succeeds for spaces created on non-git workspaces. Check if there's a git requirement in the agent creation path.
+5. If the issue is that `SpaceIsland` gets stuck in loading state, add appropriate waits in the test (e.g., wait for the Dashboard tab to appear before clicking Workflows).
+6. If the issue is that `navigateToSpace` in the test doesn't properly wait for the space to load, add a wait for a space-specific element to appear before proceeding.
+7. If the "No agents found" is a real bug (agents not being seeded), fix the seeding or add a fallback in the test to create agents via RPC.
+8. Run the test locally and verify all 6 tests pass.
+
+**Acceptance Criteria:**
+- All 6 tests in `visual-workflow-editor.e2e.ts` pass locally.
+- Root cause is identified and fixed (either in the source code or the test, depending on where the bug is).
+- Changes must be on a feature branch with a GitHub PR created via `gh pr create` targeting `dev`.
+
+**Dependencies:** None
+
+---
+
+## Task 4: Fix worktree-isolation E2E test -- session deletion should navigate away
+
+**Type:** coder
+
+**Description:**
+The test "should cleanup worktree when session is deleted" in `worktree-isolation.e2e.ts` (line 84) fails because after deleting a session, the URL still contains the deleted session ID. The test expects that after clicking "Delete Chat" and confirming, the browser navigates away from the deleted session's URL.
+
+The test flow:
+1. Creates a session, sends a message, waits for response
+2. Opens session options, clicks "Delete Chat", confirms deletion
+3. Waits 2 seconds
+4. Asserts `url` does not contain the deleted session ID -- THIS FAILS
+
+The issue is that the UI doesn't navigate away from a deleted session's URL. The session deletion RPC succeeds, but the client-side routing doesn't redirect to another page.
+
+**Key files:**
+- `packages/e2e/tests/features/worktree-isolation.e2e.ts` -- failing test (line 84-121)
+- `packages/web/src/components/SessionList.tsx` or similar -- session deletion UI handler
+- `packages/web/src/lib/router.ts` -- routing logic
+- `packages/web/src/lib/session-store.ts` or `packages/web/src/hooks/useSessionActions.ts` -- session deletion logic
+
+**Subtasks:**
+1. Run `bun install` at the worktree root.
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Recommendation: APPROVE

This is a well-structured plan with accurate root cause analysis. I verified each claim against the current codebase. Here is my detailed assessment.

---

### Task 1: task-actions-dropdown — Root Cause Verified, Well-Scoped

**Root cause accuracy:** Confirmed. The test file references `data-testid="task-options-menu"` (a three-dot dropdown) which does not exist anywhere in the web source. The actual UI uses inline `task-cancel-button` and `task-complete-button` buttons in `TaskView.tsx` (lines 944, 955). The confirmation dialog test IDs (`cancel-task-confirm`, `complete-task-confirm`) also exist in the source (lines 529, 619).

**Plan quality:** The subtask breakdown is excellent — it maps each of the 10 tests to the specific selector change needed. The note about worktree creation failures in CI (non-git workspace) is a valid secondary concern. The acceptance criteria are clear and verifiable.

**One minor gap (P3):** The plan says test 3 ("does NOT show task options menu for completed task") already passes, but looking at the test (line 141), it still asserts against `task-options-menu`. It should be rewritten to assert that neither `task-cancel-button` nor `task-complete-button` is visible, for consistency. The implementer will likely notice this, but it would be clearer if the plan called it out explicitly.

---

### Task 2: space-creation — Root Cause Verified, Well-Scoped

**Root cause accuracy:** Confirmed. `SpaceIsland.tsx` does NOT import `SpaceNavPanel`. It uses a tabbed layout (Dashboard, Agents, Workflows, Settings). The string "No runs or tasks yet" only exists in `SpaceNavPanel.tsx` (which is exported from the index but never used by SpaceIsland). The test on line 154 asserts this text, which will never appear.

**Plan quality:** Good. The fix is straightforward — remove the "No runs or tasks yet" assertion (line 154) and keep the Dashboard-based assertions (Quick Actions, Start Workflow Run, Create Task) which should be visible. The suggestion to rename the test from "3-column layout" to reflect the actual tabbed layout is a nice touch.

---

### Task 3: visual-workflow-editor — Root Cause Partially Diagnosed, Needs More Specificity

**Root cause accuracy:** The plan correctly identifies that all 6 tests fail due to a 60-second timeout waiting for `text=Workflows`. The test navigates via `navigateToSpace()` which does `page.goto(/space/{id})`. Looking at `SpaceIsland.tsx`, it calls `spaceStore.selectSpace(spaceId)` on mount (line 66), and shows a loading spinner until `spaceStore.space.value` is populated (line 108).

**Concern (P2):** This is the weakest task in the plan. The plan lists four "potential causes" but does not commit to a specific root cause. It essentially says "investigate these things and fix whatever you find." This is closer to a research ticket than an implementation task. For a coder agent, this level of ambiguity could lead to wasted cycles.

Specific suggestions to strengthen this task:
1. The plan should note that `navigateToSpace()` does `page.goto()` which triggers a full page navigation, and the `waitForWebSocketConnected` call happens in `beforeEach` before navigation — the WS connection may need to be re-established after `goto`. The test should wait for WS reconnection after navigating.
2. The "No agents found" error from `getDefaultAgentId()` (line 118 of the test) is called in test 2 only. The plan should clarify whether this is a separate bug (agent seeding on non-git workspaces) or a cascading failure from the same loading issue.
3. The plan should provide a concrete "first thing to try" — e.g., add `await waitForWebSocketConnected(page)` after `navigateToSpace()` and wait for `text=Dashboard` before proceeding to `text=Workflows`.

---

### Task 4: worktree-isolation — Root Cause Verified, Well-Scoped

**Root cause accuracy:** Confirmed. In `useSessionActions.ts` (line 62-77), `handleDeleteSession` sets `currentSessionIdSignal.value = null` but never calls `history.pushState` or any router navigation function. The URL remains on `/{sessionId}` after deletion. The test (line 117) correctly expects the URL to change, but nothing in the code makes that happen.

**Plan quality:** Good. The plan correctly identifies that this is a source code bug (not a test bug) — the deletion handler needs to navigate away. The subtasks cover both possibilities (source fix vs. test timing fix) which is appropriate. The note about this being an LLM test requiring devproxy/real credentials is helpful context for the implementer.

**One clarification (P3):** The plan lists `SessionList.tsx` as a key file, but the actual deletion handler is in `useSessionActions.ts` and the modal is in `ChatContainer.tsx`. The implementer should look at `useSessionActions.ts` and `router.ts` (specifically `navigateHome()`) rather than `SessionList.tsx`.

---

### Overall Plan Assessment

**Strengths:**
- Accurate root cause identification for 3 out of 4 tasks (verified against source)
- Good task independence — all 4 can be parallelized
- Specific file references with line numbers
- Clear acceptance criteria tied to `make run-e2e` verification

**Weaknesses:**
- Task 3 is underspecified — it is an investigation task disguised as an implementation task. The planner should either commit to a hypothesis or split it into a diagnosis subtask followed by a fix subtask.
- Minor inaccuracies in key file references (Task 4 lists `SessionList.tsx` instead of `useSessionActions.ts`)

**Missing considerations:**
- None of the tasks mention updating the E2E test file's JSDoc comments to reflect the new UI patterns. The task-actions-dropdown test file header (lines 3-8) still describes "Three-dot dropdown menu replaces the old cancel button" which will be incorrect after the fix.
- The plan does not mention whether `SpaceNavPanel.tsx` should be deleted as dead code (it is exported but never imported by any island/page component). This is out of scope for the fix but worth noting.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The plan is mostly well-structured, but has one incorrect root cause (Task 1), one incomplete root cause analysis (Task 3), and a misleading acceptance criterion that could push the implementer toward a CLAUDE.md violation.

---

## Summary

The plan correctly identifies 4 distinct failure groups and provides actionable subtasks for each. Root causes for Tasks 2 and 4 are accurately diagnosed. Task 1's root cause is partially correct but contains a significant factual error that will send the implementer in the wrong direction for test 3. Task 3's root cause analysis is genuinely uncertain (correctly noted) but the subtasks are underspecified for the most likely underlying cause.

---

## P0 — Task 1: Incorrect root cause for test 3 causes wrong fix guidance

The plan states that test 3 ('does NOT show task options menu for completed task') "already passes" because it asserts `not.toBeVisible()`. This is factually incorrect in Playwright: `expect(locator).not.toBeVisible()` fails when the element does not exist — it passes only if the element exists but is hidden. Since `task-options-menu` does not exist at all in the DOM, Playwright's `not.toBeVisible()` on a missing locator WILL pass (a non-attached element is considered not visible). So this test actually does pass, but for the reason the plan states. However, the plan then says this test should be 'kept as-is' — but the test at line 141-142 still uses `data-testid="task-options-menu"` which doesn't exist. If the implementer keeps the test as-is while rewriting all other tests to use `task-cancel-button` / `task-complete-button`, the test will continue to pass vacuously — it will assert a non-existent element is not visible, which is always true regardless of actual UI state. The correct fix for test 3 should assert that both `task-cancel-button` and `task-complete-button` are not visible for a completed task, or better, assert neither element is attached to the DOM. The plan should say: update test 3 to use the real selectors too, not keep it as-is.

**File:** `packages/e2e/tests/features/task-actions-dropdown.e2e.ts` lines 123–143

---

## P1 — Task 1: Plan omits the 'review' status discrepancy

The TaskView.tsx source (line 950) shows: `{canComplete && task.status !== 'review' && ...}` — the complete button is hidden for 'review' status even though canComplete is true for review. The plan's subtask 1 says 'shows task options menu for in_progress task (complete + cancel)' should assert both buttons visible, which is correct. But the test file also creates tasks at 'review' status via the helper, and the plan does not address what the expected button visibility should be for 'review': cancel visible, complete hidden. There is no test explicitly for 'review' status button visibility in the current file, and the plan does not ask for one. This is a gap in coverage but not a blocker — just note it for completeness.

---

## P1 — Task 3: Root cause analysis is too speculative; missing the most likely cause

The plan lists 4 speculative potential causes for the 60-second timeout waiting for `text=Workflows`. Reading the actual `SpaceIsland.tsx` and `space-store.ts` code reveals a likely concrete cause that the plan misses:

`doSelect()` at line 338 of `space-store.ts` has an early-return: `if (this.spaceId.value === spaceId) return`. If a previous test's `deleteTestSpace` call does not fully reset `spaceStore.spaceId` back to null before the next `beforeEach` creates a new space with the same `workspaceRoot` (the tests clean up any existing space at the same path, then create a new one), the new space will have a different UUID — so this is not the issue in this case. However, another concrete cause: `navigateToSpace(page, spaceId)` at line 73 simply navigates to the URL via `page.goto` with no wait for the Workflows tab to appear. The test then calls `page.locator('text=Workflows').first().click()` inside `openNewWorkflowEditor`. If `SpaceIsland` is still in its loading state (waiting for `space.overview` RPC), the tab bar is not rendered (line 108: loading shows a spinner instead). The `navigateToSpace` helper only waits for URL match, not for the tab bar to appear. The plan mentions this in subtask 6 but buries it as a secondary option. This should be the PRIMARY hypothesis: the `navigateToSpace` helper needs to wait for `text=Workflows` to be visible before returning. The subtasks should be reordered to prioritize this fix.

Additionally, the plan's subtask 4 asks to check whether `seedPresetAgents` fails on non-git workspaces. Reading `seed-agents.ts` shows the seeding function has no git dependency — it only calls `agentManager.create` with name/role/description/tools. The 'No agents found' error is in `getDefaultAgentId` helper (line 118): `if (!agent) throw new Error('No agents found in space')`. This can fail if seeding fails (but seeding errors are non-fatal and caught), or if `spaceAgent.list` is called before seeding completes (timing issue at space creation). The plan correctly identifies this as a possible cause but should explicitly rule out git dependency and focus on timing.

---

## P1 — Task 1: Acceptance criterion encourages a CLAUDE.md violation

The plan says tests should 'assert `task-complete-button` is not visible' for pending tasks. But `not.toBeVisible()` on an element that was conditionally rendered (not in DOM at all) passes vacuously. The correct Playwright assertion is `expect(locator).not.toBeAttached()` or checking the element count is 0. The plan's wording will lead implementers to write assertions that pass for the wrong reason (element absent vs element hidden). This should specify: 'not rendered in DOM' (use `not.toBeAttached()`), not just 'not visible'.

---

## P2 — Task 4: Root cause diagnosis is partially wrong

The plan says the fix is 'add navigation to the home page after successful deletion' and implies the UI does not navigate away. Reading `useSessionActions.ts` lines 62-78 shows deletion DOES set `currentSessionIdSignal.value = null` (via setTimeout), which in turn triggers the `App.tsx` effect at line 82-122 to navigate to home. So navigation IS implemented — the question is whether it fires within 2 seconds. The `waitForTimeout(2000)` at line 113 of the test is the real problem: it's a fixed sleep, not a Playwright wait. The plan mentions this in subtask 6 ('alternatively, if navigation works correctly...') but positions it as a secondary fallback. The implementer should be directed to first verify whether the navigation actually fires (it does — via signal + effect), and then replace `waitForTimeout(2000)` with `await expect(page).not.toHaveURL(...)` as the primary fix. The plan's primary recommended action ('add navigation logic') is the wrong fix and would add redundant code.

**File:** `packages/web/src/hooks/useSessionActions.ts` lines 62-78, `packages/web/src/App.tsx` lines 82-122

---

## P2 — Task 3: Missing concrete acceptance criterion

Task 3's acceptance criterion says 'root cause is identified and fixed (either in the source code or the test)'. This is too vague for a coder implementer. If the root cause is indeed the missing wait in `navigateToSpace`, the acceptance criterion should say 'navigateToSpace helper waits for the space tab bar to appear before the test proceeds' — something concrete and verifiable. The current criterion allows an implementer to add an arbitrary `waitForTimeout` and call it done.

---

## P3 — Minor: Plan says 'run bun install' as step 1 for all 4 tasks

This is redundant boilerplate. The CLAUDE.md instructions say the workspace is already set up. Including this as a numbered subtask inflates the task with noise.

---

## P3 — Task 2: Test rename suggestion is valuable but underspecified

Subtask 3 says 'consider renaming the test to reflect the actual layout (tabbed, not 3-column)'. This is good advice but should be a concrete instruction ('rename the test to...' with a suggested name), not a hedge.

---

## Overall Assessment

The plan is mostly correct and well-structured. The 4-task breakdown is appropriate and the tasks are genuinely independent. The main issues are:
1. Task 1 gives wrong guidance for test 3 (keep-as-is when it should be updated)
2. Task 4 primary fix is wrong (navigation is already implemented via signal; fix is the test's timing assertion)
3. Task 3 buries the most likely fix (add wait in navigateToSpace helper) behind speculative investigation items

These issues will cause an implementer to either miss a test gap (Task 1), add unnecessary navigation code (Task 4), or spend time investigating non-issues before reaching the real fix (Task 3).

lsm (commented):

### Recommendation: APPROVE

This is iteration 2 of the plan. The three prior P0/P1/P2 issues have all been addressed satisfactorily. Here is my assessment of each resolution and the overall plan quality.

---

### Prior Issue Resolution

**1. Task 3 now has a concrete primary root cause (was P2: "too speculative")**

Resolved. The plan now identifies the specific root cause at line 104: `navigateToSpace()` waits only for the URL to match, not for `SpaceIsland` to finish loading. The fix approach is concrete — wait for `text=Dashboard` to be visible before returning. The secondary "No agents found" issue is correctly attributed to the same timing problem. The plan also correctly notes that `seedPresetAgents` has no git dependency (confirmed in the source).

**2. Task 1 test 3 now updates selectors instead of keep-as-is (was P0)**

Resolved. Line 38 explicitly states: "DO NOT keep as-is. This test still uses `task-options-menu` which doesn't exist. Rewrite to assert that both `task-cancel-button` and `task-complete-button` are `not.toBeAttached()` for a completed task." The plan also correctly uses `not.toBeAttached()` throughout (not `not.toBeVisible()`) to avoid vacuous assertions on missing elements.

**3. Task 4 now uses proper Playwright assertion as primary fix (was P2: wrong primary fix)**

Resolved. The plan title is "replace fixed sleep with proper Playwright wait" and line 143 explicitly says: "Do NOT add redundant navigation code to `useSessionActions.ts` — the navigation already works via the signal -> effect chain." The fix is to replace `waitForTimeout(2000)` with `expect(page).not.toHaveURL(new RegExp(sessionId))`. I verified in the source that the signal-to-navigation chain does exist (`useSessionActions.ts` line 70 sets `currentSessionIdSignal.value = null`, `App.tsx` effect at line 82-122 picks this up and calls `navigateToHome`).

---

### Remaining Observations

**P3 — Task 3: `navigateToSpace` is local, not shared.**

The plan at line 120 says "If `navigateToSpace` is shared across multiple test files, verify the change doesn't break other space-related tests." In fact, `navigateToSpace` is defined locally in both `visual-workflow-editor.e2e.ts` (line 71) and `space-workflow-rules.e2e.ts` (line 72) — it is NOT in the shared helpers directory. The implementer will need to fix both copies independently, or extract it to a shared helper. This is a minor observation; the plan handles it with a conditional that the implementer can evaluate.

**P3 — Task 1: The `review` status button visibility gap is not addressed.**

The prior review noted that `TaskView.tsx` line 950 hides the complete button for `review` status (`task.status !== 'review'`). The plan does not include a test for this status. This is an existing coverage gap unrelated to the CI failures, so it does not need to be in this plan, but it is worth noting for future work.

**P3 — Task 4: LLM test caveat could be more explicit.**

The plan mentions this is an LLM test requiring devproxy or real API credentials at line 158, but does not specify which environment variable or devproxy configuration is needed. The implementer will need to consult CLAUDE.md for the `NEOKAI_USE_DEV_PROXY=1` flag. Minor friction, not blocking.

---

### Overall Assessment

The plan is well-structured, accurate, and actionable. All four tasks have:
- Correct root cause identification (verified against source)
- Concrete fix approaches with specific code patterns
- Clear acceptance criteria tied to local test runs
- Appropriate key file references with line numbers

The prior review's three substantive issues (speculative Task 3, vacuous Task 1 test 3, wrong Task 4 primary fix) have all been properly resolved in this iteration. The remaining observations are cosmetic.

lsm (commented):

**Recommendation: APPROVE** — All three prior P0/P1/P2 issues from iteration 1 have been addressed. The plan is now concrete, correctly scoped, and actionable. Minor residual gaps noted below.

---

### Prior Issues Resolution

**P0 — Task 3: Concrete root cause (missing wait in navigateToSpace)**
Resolved. The plan now clearly states the primary root cause: `navigateToSpace()` only waits for the URL to match, not for the SpaceIsland to finish loading, causing tests to attempt clicking `text=Workflows` before the tab bar renders. The fix is correctly described as adding a wait for `text=Dashboard` to be visible. This matches what the actual code shows: `navigateToSpace` in `visual-workflow-editor.e2e.ts` (line 71-74) only calls `page.waitForURL(...)`.

**P1 — Task 1 test 3: Use real selectors, not keep-as-is**
Resolved. Subtask 2 now explicitly states: "DO NOT keep as-is. Rewrite to assert that both `task-cancel-button` and `task-complete-button` are `not.toBeAttached()` for a completed task." This is correct — the actual `TaskView.tsx` (lines 939-968) uses inline buttons, not a dropdown, and the existing test on line 141-143 uses `not.toBeVisible()` on a nonexistent `task-options-menu` which passes vacuously.

**P2 — Task 4: Replace waitForTimeout with Playwright assertion (primary fix)**
Resolved. The plan correctly identifies the primary fix as replacing `page.waitForTimeout(2000)` (line 113 in worktree-isolation.e2e.ts) + manual URL check with `await expect(page).not.toHaveURL(new RegExp(sessionId), { timeout: 10000 })`. The plan also correctly notes that navigation already works via the signal→effect chain in `useSessionActions.ts` (line 69-71) and `App.tsx` (lines 82-122), so no source changes are needed.

---

### Remaining Gaps (New Issues)

**P2 — Task 3, Subtask 3: `navigateToSpace` fix must be applied to TWO files**

The plan says "If `navigateToSpace` is shared across multiple test files, verify the change doesn't break other space-related tests (e.g., `space-creation.e2e.ts`)." This framing is slightly wrong. `navigateToSpace` is NOT a shared helper from the helpers directory — it is a **local function duplicated independently** in both `visual-workflow-editor.e2e.ts` (line 71) and `space-workflow-rules.e2e.ts` (line 72), and both have the identical missing-wait bug. The plan's phrasing of "if shared" leaves ambiguity about whether the implementer should also fix `space-workflow-rules.e2e.ts`. The plan should explicitly state that the same fix must be applied to both files to avoid a repeat failure for `space-workflow-rules` tests.

**P3 — Task 1: In_progress status transition creates only in_progress, not in_progress→review**

The helper `createRoomAndTask` (line 44-61 of the test file) does transition to `in_progress` first and then to `review` when review is requested. The plan's note about worktree creation failing for `in_progress` status is accurate, but the plan doesn't mention that the helper bypasses this by using `task.setStatus` RPC directly (not through the agent/worktree path). This is already handled correctly by the existing helper — the plan could mention this explicitly so the implementer doesn't add unnecessary workarounds.

**P3 — Task 4: `beforeEach` also has `waitForTimeout(1000)` (line 23)**

The `worktree-isolation.e2e.ts` `beforeEach` (line 23) also contains `page.waitForTimeout(1000)` — a second fixed sleep that could be replaced with a proper wait. The plan only addresses the `waitForTimeout(2000)` in the test body. While not blocking (the beforeEach sleep is less likely to cause the specific CI failure), it would be cleaner to replace it with a proper assertion as well. The plan could mention this as a secondary improvement.

---

### Plan Quality Summary

The plan is well-structured with concrete file references, line numbers, correct root cause descriptions, and accurate acceptance criteria. All four tasks have clear, feasible implementations. The plan correctly identifies that Tasks 1-4 are independent and can be executed in parallel. The acceptance criteria for each task are specific and verifiable.

lsm (commented):

### Recommendation: APPROVE

This is iteration 3 of the plan. The remaining P2 from iteration 2 has been addressed.

---

### Prior P2 Resolution: Task 3 navigateToSpace duplication

**Resolved.** Subtask 3 (line 120) now explicitly states: "`navigateToSpace` is NOT a shared helper — it is a local function duplicated independently in both `visual-workflow-editor.e2e.ts` (line 71) and `space-workflow-rules.e2e.ts` (line 72). Both copies have the identical missing-wait bug. Apply the same fix to BOTH files to prevent a repeat failure in `space-workflow-rules` tests." This is clear, unambiguous, and correct — verified against the codebase.

---

### Remaining Observations

**P3 — Task 3: Key files section (line 112) is slightly misleading.**

Line 112 lists `packages/e2e/tests/helpers/ -- test helpers including navigateToSpace`. This contradicts subtask 3's correct statement that `navigateToSpace` is a local function, not a shared helper. The key files section also does not list `space-workflow-rules.e2e.ts` even though subtask 3 requires modifying it. These are minor inconsistencies that an implementer will resolve when they actually locate the function, but cleaning up the key files list would avoid initial confusion.

**P3 — Task 3: `space-export-import.e2e.ts` also has a similar `navigateToSpaceAgents` function.**

There is a third test file (`space-export-import.e2e.ts`, line 129) with a similar local helper `navigateToSpaceAgents` that navigates to a space URL. It likely has the same missing-wait pattern. This is out of scope for the current plan (that test suite is not in the 4 failing suites), but worth noting for completeness.

---

### Overall Assessment

The plan is mature, accurate, and actionable across all four tasks. All substantive issues from iterations 1 and 2 have been resolved:
- Task 1: Correct selectors, `not.toBeAttached()` assertions, test 3 rewrite, JSDoc update
- Task 2: Correct layout assertions, concrete test rename
- Task 3: Concrete root cause, explicit dual-file fix requirement, proper Playwright waits
- Task 4: Correct diagnosis (test timing, not missing navigation), `expect(page).not.toHaveURL()` pattern

No new P0/P1/P2 issues introduced. The two P3 observations above are cosmetic and non-blocking.

lsm (commented):

**Recommendation: APPROVE** — The one remaining P2 from iteration 2 has been fully resolved. The plan is now accurate, complete, and directly actionable.

---

### Iteration 3 Change Assessment

**P2 resolved — Task 3 navigateToSpace dual-file fix is now explicit and unambiguous.**

The previous iteration used the phrasing "if `navigateToSpace` is shared across multiple test files..." which left it unclear whether the implementer needed to fix `space-workflow-rules.e2e.ts`. The new text at subtask 3 (line 120 of the plan) now reads:

> "`navigateToSpace` is NOT a shared helper — it is a local function duplicated independently in both `visual-workflow-editor.e2e.ts` (line 71) and `space-workflow-rules.e2e.ts` (line 72). Both copies have the identical missing-wait bug. Apply the same fix to BOTH files to prevent a repeat failure in `space-workflow-rules` tests."

This is correct and verified against the actual source. Both files define the identical function body at the stated lines:

- `packages/e2e/tests/features/visual-workflow-editor.e2e.ts` line 71–74
- `packages/e2e/tests/features/space-workflow-rules.e2e.ts` line 72–75

Both have `page.goto(...)` followed by `page.waitForURL(...)` with no wait for the space tab bar. The fix instruction is unambiguous: add a wait for `text=Dashboard` to be visible in both copies before returning.

---

### Residual Gaps (Unchanged from Iteration 2, Still Low Severity)

**P3 — Task 4: `beforeEach` `waitForTimeout(1000)` at line 23 of `worktree-isolation.e2e.ts` not addressed.**
This is a separate fixed sleep that could be replaced with a proper Playwright assertion. It is unlikely to be the cause of the CI failure but is suboptimal. Out of scope for this plan iteration; a follow-up improvement would be appropriate.

**P3 — Task 1: Coverage gap for `review` status button visibility.**
`TaskView.tsx` line 950 hides the complete button for `review` status. No test explicitly asserts this behavior. The plan does not ask for one. Not a blocker; can be tracked separately.

---

### Overall Plan Quality

All four tasks now have:
- Accurate root cause descriptions verified against actual source code
- Concrete, line-numbered subtasks
- Acceptance criteria tied to `make run-e2e` verification
- No ambiguous conditionals in the fix instructions

The plan is ready for parallel implementation.

PR #557, opened 2026-03-21
fix: replace fixed sleep with Playwright auto-retrying assertion in worktree-isolation E2E test

lsm:
## Summary
- Replace `page.waitForTimeout(2000)` + manual URL check with `expect(page).not.toHaveURL()` in the "should cleanup worktree when session is deleted" test
- The Playwright auto-retrying assertion waits up to 10s for the URL to no longer contain the deleted session ID, eliminating flaky failures from the fixed 2-second sleep
- No source code changes needed — the existing signal → effect navigation chain in `useSessionActions.ts` / `App.tsx` already works correctly

## Test plan
- [ ] `make run-e2e TEST=tests/features/worktree-isolation.e2e.ts` passes (requires devproxy or real API credentials)

### DIFF
--- a/packages/e2e/tests/features/worktree-isolation.e2e.ts
+++ b/packages/e2e/tests/features/worktree-isolation.e2e.ts
@@ -109,12 +109,8 @@ test.describe('Worktree Isolation', () => {
 		const confirmButton = page.locator('[data-testid="confirm-delete-session"]');
 		await confirmButton.click();
 
-		// Wait for deletion to complete
-		await page.waitForTimeout(2000);
-
-		// Session should be gone - verify by checking URL
-		const url = page.url();
-		expect(url).not.toContain(deletedSessionId);
+		// Session should be gone - URL should no longer contain the deleted session ID
+		await expect(page).not.toHaveURL(new RegExp(deletedSessionId!), { timeout: 10000 });
 
 		// Don't try to cleanup in afterEach since it's already deleted
 		sessionId = null;


lsm (commented):

**Recommendation: APPROVE**

The change is clean and correct. It replaces a brittle `waitForTimeout(2000)` + synchronous `expect(url).not.toContain(...)` with Playwright's auto-retrying `await expect(page).not.toHaveURL(regex, { timeout: 10000 })`, which is the idiomatic approach.

### What was checked

- The `deletedSessionId!` non-null assertion on line 113 is safe: `sessionId` is assigned by `createSessionViaUI()` on line 86 (which returns `string`), and `deletedSessionId` captures it on line 99 before any nulling occurs.
- The `RegExp` constructor receives a plain UUID string, which contains no special regex characters, so no escaping is needed.
- The 10-second timeout is reasonable for a deletion + navigation round-trip.
- No source code files were modified — only the E2E test file, as required.
- The rest of the test file is unchanged and consistent.

No issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

The change is exactly what was requested: the fixed `page.waitForTimeout(2000)` sleep and the manual synchronous URL assertion are replaced with `await expect(page).not.toHaveURL(new RegExp(deletedSessionId!), { timeout: 10000 })`, which auto-retries via Playwright's built-in assertion engine. No source code files were modified.

---

### Correctness

The fix is sound end-to-end:

1. After confirm-delete, `useSessionActions.ts` calls `deleteSession(sessionId)`, then sets `currentSessionIdSignal.value = null` (deferred via `setTimeout(..., 0)`).
2. The `effect` in `App.tsx` (lines 82-123) watches that signal and calls `navigateToHome(true)` when it becomes `null`, which changes the URL to `/`.
3. The Playwright assertion retries the URL check up to 10 seconds, catching the navigation whenever it completes — far more robust than a fixed 2-second sleep.

Session IDs are UUIDs (`[a-f0-9-]+` as confirmed in `packages/web/src/lib/router.ts` line 27). They contain only hex characters and hyphens, neither of which has special regex meaning outside a character class, so passing the raw session ID string into `new RegExp()` without escaping is safe here.

The non-null assertion (`deletedSessionId!`) is correct: `deletedSessionId` is assigned from `sessionId` which is guaranteed to be non-null at that point in the test (it was just populated by `createSessionViaUI`).

---

**P3 — Remaining `waitForTimeout` calls in this file**

The `beforeEach` on line 23 still has `await page.waitForTimeout(1000)`, and the skipped test on line 137 has `await page.waitForTimeout(1000)`. These are outside the scope of this task, but they follow the same anti-pattern that was just fixed. Worth a follow-up.

**P3 — `createSessionViaUI` uses direct RPC, not the UI button**

The CLAUDE.md E2E rules state "Sessions must be created via the 'New Session' button, never via RPC". The helper `createSessionViaUI` (in `packages/e2e/tests/helpers/wait-helpers.ts` lines 60-71) creates sessions via `hub.request('session.create', ...)` then navigates by URL. This is a pre-existing violation not introduced by this PR. No action needed here, but noting for awareness.

---

The change is minimal, correct, scoped exactly to what was asked, and makes the test meaningfully more reliable. No issues block merging.

PR #559, opened 2026-03-21
fix: update space-creation E2E test assertions to match actual tabbed UI

lsm:
## Summary
- Removed assertion for "No runs or tasks yet" text from `SpaceNavPanel` (not rendered in `SpaceIsland`)
- Fixed "navigates to Spaces section" test to assert for "Home" header instead of non-existent `h2:has-text("Spaces")`
- Added assertion for "Dashboard" tab visibility in the space view
- Renamed test from "creates space and shows 3-column layout" to "creates space and shows tabbed dashboard layout"

## Test plan
- [x] All 7 tests in `space-creation.e2e.ts` pass locally

### DIFF
--- a/packages/e2e/tests/features/space-creation.e2e.ts
+++ b/packages/e2e/tests/features/space-creation.e2e.ts
@@ -7,8 +7,7 @@
  * - Workspace path field is required
  * - Name auto-suggests from workspace path
  * - Creating a space navigates to it
- * - Space 3-column layout renders (nav panel, dashboard, no task pane initially)
- * - Nav panel shows "No runs or tasks yet" for a fresh space
+ * - Space tabbed dashboard layout renders (Dashboard tab active with quick actions)
  *
  * Setup: creates a space via dialog (UI-only)
  * Cleanup: deletes the space via RPC in afterEach (infrastructure)
@@ -61,8 +60,11 @@ test.describe('Space Creation UX', () => {
 		await expect(spacesButton).toBeVisible({ timeout: 5000 });
 		await spacesButton.click();
 
-		// ContextPanel should show "Spaces" header
-		await expect(page.locator('h2:has-text("Spaces")')).toBeVisible({ timeout: 5000 });
+		// ContextPanel should show the Spaces home view with "Home" header and "Create Space" button
+		await expect(page.locator('text=Home')).toBeVisible({ timeout: 5000 });
+		await expect(
+			page.getByRole('button', { name: 'Create Space', exact: true }).first()
+		).toBeVisible({ timeout: 5000 });
 	});
 
 	test('shows "Create Space" button in Spaces section', async ({ page }) => {
@@ -119,7 +121,7 @@ test.describe('Space Creation UX', () => {
 		await expect(nameInput).toHaveValue('my-cool-project', { timeout: 2000 });
 	});
 
-	test('creates space and shows 3-column layout', async ({ page }) => {
+	test('creates space and shows tabbed dashboard layout', async ({ page }) => {
 		const workspaceRoot = await getWorkspaceRoot(page);
 
 		const spacesButton = page.getByRole('button', { name: 'Spaces', exact: true });
@@ -149,10 +151,8 @@ test.describe('Space Creation UX', () => {
 			createdSpaceId = match[1];
 		}
 
-		// Space layout should be visible
-		// Left column: nav panel
-		await expect(page.locator('text=No runs or tasks yet')).toBeVisible({ timeout: 5000 });
-		// Middle column: dashboard (quick actions)
+		// Tabbed dashboard should be visible with Dashboard tab active
+		await expect(page.locator('text=Dashboard')).toBeVisible({ timeout: 5000 });
 		await expect(page.locator('text=Quick Actions')).toBeVisible({ timeout: 5000 });
 		await expect(page.locator('text=Start Workflow Run')).toBeVisible({ timeout: 3000 });
 		await expect(page.locator('text=Create Task')).toBeVisible({ timeout: 3000 });
--- a/packages/e2e/tests/features/space-creation.e2e.ts
+++ b/packages/e2e/tests/features/space-creation.e2e.ts
@@ -2,7 +2,7 @@
  * Space Creation E2E Tests
  *
  * Verifies:
- * - Navigating to the Spaces section
+ * - Navigating to the Spaces section (Home header + Create Space button visible)
  * - "Create Space" dialog opens
  * - Workspace path field is required
  * - Name auto-suggests from workspace path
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

The changes correctly align E2E test assertions with the actual UI implementation.

### Verified

1. **"navigates to Spaces section" test (line 64):** The old assertion `h2:has-text("Spaces")` was wrong because `ContextPanel.tsx:341-363` renders a `<button>` with text "Home" (not an `<h2>` with "Spaces") when `isSpaces` is true. The new assertion for `text=Home` and the "Create Space" button matches the actual DOM.

2. **"creates space and shows tabbed dashboard layout" test (lines 154-158):** The old assertion for "No runs or tasks yet" was from `SpaceNavPanel`, which is not rendered by `SpaceIsland.tsx`. The new assertions for "Dashboard", "Quick Actions", "Start Workflow Run", and "Create Task" all match text rendered in `SpaceDashboard.tsx` (lines 119, 152, 156, 162) and the tab bar in `SpaceIsland.tsx` (line 38).

3. **`text=Home` locator specificity (line 64):** The NavRail also has a "Home" button, but its label is only in a `title` attribute (via `NavIconButton`), not visible text. Playwright's `text=` selector matches visible text by default, so no false match occurs here.

No issues found. Clean, minimal diff that fixes assertions to match the actual rendered UI.

lsm (commented):

The intent is correct: the old assertions tested against non-rendered components (`SpaceNavPanel` text and a non-existent `h2:has-text("Spaces")` header), and the replacements match the actual rendered UI. The fix unblocks CI. That said, there are a few issues worth addressing.

---

### P1 — Stale comment `// Header CTA button` in the unchanged test

`packages/e2e/tests/features/space-creation.e2e.ts`, line 74:

```ts
test('shows "Create Space" button in Spaces section', async ({ page }) => {
    // Header CTA button   <-- this comment is wrong
    await expect(
        page.getByRole('button', { name: 'Create Space', exact: true }).first()
    ).toBeVisible({ timeout: 5000 });
});
```

The `// Header CTA button` comment was not updated and is now actively misleading. Looking at `ContextPanel.tsx` lines 384–405, the header button is only conditionally rendered for `home`, `chats`, and `rooms` nav sections — **not** `spaces`. The "Create Space" button that is actually tested here is the one rendered at the bottom of `SpaceContextPanel` (line 280-297 in `SpaceContextPanel.tsx`). The comment should be corrected to avoid confusion for future contributors trying to understand what UI element is under test.

---

### P2 — Fragile `text=Home` locator

`packages/e2e/tests/features/space-creation.e2e.ts`, line 64:

```ts
await expect(page.locator('text=Home')).toBeVisible({ timeout: 5000 });
```

`page.locator('text=Home')` is a partial-text match that returns any element containing the word "Home". In the current render, this works because the visible "Home" text is uniquely the clickable header label in `ContextPanel.tsx` (line 362) when `isSpaces === true`. However it's fragile if any other text node containing "Home" is added to the page. A more specific selector would be less brittle — for example:

```ts
await expect(page.getByRole('button', { name: 'Home', exact: true })).toBeVisible({ timeout: 5000 });
```

This targets the specific `<button>` element rendered as the spaces-section header (ContextPanel.tsx line 342), which is semantically correct.

---

### P2 — Redundant test overlap between the two navigation tests

The new `navigates to Spaces section via NavRail` test now asserts both "Home" text AND "Create Space" button visibility. The immediately adjacent `shows "Create Space" button in Spaces section` test asserts only the "Create Space" button after the same Spaces click. These two tests now have nearly identical coverage, with the first being a superset. The second test adds no meaningful assertion beyond what the first already verifies. Consider either removing the second test or making it exercise a distinct aspect of the Create Space button (e.g., that it opens the dialog — but that is already covered by the third test too).

---

### P3 — Superfluous `.first()` on a single-match locator

`packages/e2e/tests/features/space-creation.e2e.ts`, lines 66 and 76:

```ts
page.getByRole('button', { name: 'Create Space', exact: true }).first()
```

Looking at the rendered DOM when `navSection === 'spaces'`, the header button is NOT rendered (it's guarded by the `home|chats|rooms` condition in ContextPanel.tsx lines 384–386), so there is only one "Create Space" button in the DOM at this point — the one inside `SpaceContextPanel`. The `.first()` is harmless but implies there could be multiple matches, which is misleading. It was likely carried over defensively from a state where both buttons were rendered.

---

### Overall

The change correctly fixes the broken assertions and the tests should pass. The P1 stale comment is the main thing worth addressing before merging since it will mislead future contributors about which UI element is under test. The P2 locator fragility is low-risk in the current UI but worth a follow-up.

lsm (commented):

**Recommendation: APPROVE**

All four items from the previous review round have been addressed:

1. **Stale `// Header CTA button` comment** -- Removed by deleting the entire redundant test. Confirmed gone.
2. **Fragile `text=Home` locator** -- Scoped to `.border-b >> text=Home` (line 65), which constrains the match to the header div rendered at `ContextPanel.tsx:339`. This avoids matching the NavRail's "Home" icon button (which only has "Home" in a `title` attribute, not visible text).
3. **Redundant `shows "Create Space" button` test** -- Removed. The "Create Space" button assertion is now part of the navigation test (line 66-68), eliminating the overlap.
4. **Unnecessary `.first()` calls** -- All removed. Since only one "Create Space" button is rendered when `navSection === 'spaces'` (the header CTA is guarded by `home|chats|rooms` in `ContextPanel.tsx:384-386`), `.first()` was misleading.

### P3 -- `.border-b >> text=Home` locator style

Line 65: `page.locator('.border-b >> text=Home')` works correctly but is not idiomatic Playwright. Since the "Home" text is inside a `<button>` element (`ContextPanel.tsx:342-363`), `page.getByRole('button', { name: 'Home', exact: true })` would be more readable and resilient to CSS class changes. Not blocking since the current locator works and `.border-b` on the header div is stable.

### P3 -- PR body says "All 7 tests" but the file contains 6

The test count in the PR description is off by one (one test was removed). Not a code issue.

No blocking issues. The assertions now match the actual rendered UI correctly.

lsm (commented):

**Recommendation: APPROVE**

The worker correctly addressed all feedback from the previous review round. Here is the verification of each item, plus one new observation.

---

### Resolved: P1 — Stale `// Header CTA button` comment

Resolved by deleting the entire redundant test (`shows "Create Space" button in Spaces section`) that contained the stale comment. Both the comment and the test are gone.

### Resolved: P2 — Redundant test overlap

The `shows "Create Space" button in Spaces section` test was removed. The `navigates to Spaces section via NavRail` test now covers both header landmark and `Create Space` button visibility without duplication.

### Resolved: P3 — Superfluous `.first()` calls

All five `.first()` calls were correctly removed from lines 66, 76, 87, 101, 118, and 153 of the original file. The DOM only ever contains one `Create Space` button when the Spaces section is active (inside `SpaceContextPanel`, not the header, which is guarded by the `home|chats|rooms` condition in `ContextPanel.tsx` lines 384–386).

### Partially resolved: P2 — Fragile `text=Home` locator

`packages/e2e/tests/features/space-creation.e2e.ts`, line 65:

```ts
await expect(page.locator('.border-b >> text=Home')).toBeVisible({ timeout: 5000 });
```

The previous `text=Home` locator was replaced with `.border-b >> text=Home`. This is scoped more narrowly and correctly identifies the header region in `ContextPanel.tsx` (line 339, the `p-4 border-b border-dark-700` div that wraps the `<button>Home</button>`).

It is worth noting that the original review suggestion — `getByRole('button', { name: 'Home', exact: true })` — would have been incorrect. `NavIconButton` sets `aria-label={label}` (see `NavIconButton.tsx` line 28), so the NavRail Home button also has an accessible name of "Home". That `getByRole` locator would therefore match two buttons, causing a strict-mode locator error or a non-deterministic match. The worker's choice of `.border-b >> text=Home` avoids that ambiguity.

The remaining concern is that `.border-b` is a CSS presentation class. The `>>` combinator is also legacy Playwright syntax (deprecated in favour of `.filter()` or chained `.locator()`). If the header's border class is ever removed or refactored, the test breaks silently with no clear connection to the semantic element being targeted. A `data-testid="spaces-context-panel-header"` on the header div in `ContextPanel.tsx` would be the correct long-term fix. However, this is a pre-existing structural limitation of the codebase's test infrastructure, not something introduced by this PR, and the locator is strictly better than the original.

---

### New: P3 — PR description test count mismatch

The PR body states "All 7 tests pass locally" but the file after this PR contains 6 tests (the redundant test was removed as part of the feedback fixes). This is a documentation-only discrepancy with no functional impact.

---

### Overall

All P0, P1, and P2 issues have been addressed or are pre-existing infrastructure limitations that this PR does not worsen. The test file correctly verifies the actual rendered UI — the tabbed `SpaceIsland` layout, the "Home" header button, and the `Create Space` button sourced from `SpaceContextPanel`. Approve.

PR #567, opened 2026-03-21
fix: fix visual-workflow-editor E2E tests — space URL sync + navigateToSpace wait

lsm:
## Summary

- **App.tsx URL sync effect** didn't handle space routes (`/space/:id`, `/space/:id/session/:sid`, `/space/:id/task/:tid`). On page load at a space URL, the effect computed `expectedPath='/'` and redirected to home, breaking all direct-URL space navigation including E2E test `page.goto('/space/...')`.
- **`navigateToSpace()` test helper** only waited for URL match, not for SpaceIsland to finish loading. Added `Dashboard` tab visibility wait.
- **`createTestSpace()` cleanup** used exact path matching that failed on macOS (`/var/` vs `/private/var/` symlink resolution). Now deletes all leftover spaces.
- **Visual editor node stagger** was 20px (nodes are 160×80px), causing overlapping nodes that blocked Playwright clicks. Increased to 100px vertical spacing.
- **Edit button locator** was page-scoped and matched hidden buttons from other workflow cards. Scoped to the target card.
- Added `serial` mode to test.describe to prevent parallel workspace path conflicts.

## Test plan

- [x] All 6 tests in `visual-workflow-editor.e2e.ts` pass locally
- [x] TypeScript type check passes
- [x] Lint passes (oxlint)
- [x] Format check passes (biome)
- [x] Knip dead code detection passes

### DIFF
--- a/packages/e2e/tests/features/space-workflow-rules.e2e.ts
+++ b/packages/e2e/tests/features/space-workflow-rules.e2e.ts
@@ -32,15 +32,17 @@ async function createTestSpace(page: Page): Promise<string> {
 			const hub = window.__messageHub || window.appState?.messageHub;
 			if (!hub?.request) throw new Error('MessageHub not available');
 
-			// Clean up any leftover spaces (path comparison may differ due to macOS
-			// symlink resolution: /var/folders/ vs /private/var/folders/)
+			// Clean up any leftover space at this workspace path.
+			// Normalize paths to handle macOS symlink resolution (/var/ vs /private/var/).
+			const norm = (p: string) => p.replace(/^\/private/, '');
 			try {
 				const list = (await hub.request('space.list', {})) as Array<{
 					id: string;
 					workspacePath: string;
 				}>;
-				for (const s of list) {
-					await hub.request('space.delete', { id: s.id });
+				const existing = list.find((s) => norm(s.workspacePath) === norm(wsPath));
+				if (existing) {
+					await hub.request('space.delete', { id: existing.id });
 				}
 			} catch {
 				// Ignore cleanup errors
--- a/packages/e2e/tests/features/space-workflow-rules.e2e.ts
+++ b/packages/e2e/tests/features/space-workflow-rules.e2e.ts
@@ -32,15 +32,15 @@ async function createTestSpace(page: Page): Promise<string> {
 			const hub = window.__messageHub || window.appState?.messageHub;
 			if (!hub?.request) throw new Error('MessageHub not available');
 
-			// Delete any leftover space from a previous failed run
+			// Clean up any leftover spaces (path comparison may differ due to macOS
+			// symlink resolution: /var/folders/ vs /private/var/folders/)
 			try {
 				const list = (await hub.request('space.list', {})) as Array<{
 					id: string;
 					workspacePath: string;
 				}>;
-				const existing = list.find((s) => s.workspacePath === wsPath);
-				if (existing) {
-					await hub.request('space.delete', { id: existing.id });
+				for (const s of list) {
+					await hub.request('space.delete', { id: s.id });
 				}
 			} catch {
 				// Ignore cleanup errors
@@ -72,11 +72,15 @@ async function deleteTestSpace(page: Page, spaceId: string): Promise<void> {
 async function navigateToSpace(page: Page, spaceId: string): Promise<void> {
 	await page.goto(`/space/${spaceId}`);
 	await page.waitForURL(`/space/${spaceId}**`, { timeout: 10000 });
+	// Wait for SpaceIsland to finish loading — the tab bar appears after space.overview resolves
+	await expect(page.locator('text=Dashboard').first()).toBeVisible({ timeout: 15000 });
 }
 
 // ─── Tests ────────────────────────────────────────────────────────────────────
 
 test.describe('Space Workflow Rules & Navigation Integration', () => {
+	// All tests share the same workspace path (one space at a time) — must run serially
+	test.describe.configure({ mode: 'serial' });
 	test.use({ viewport: DESKTOP_VIEWPORT });
 
 	let spaceId = '';
--- a/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
+++ b/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
@@ -34,15 +34,15 @@ async function createTestSpace(page: Page): Promise<string> {
 			const hub = window.__messageHub || window.appState?.messageHub;
 			if (!hub?.request) throw new Error('MessageHub not available');
 
-			// Clean up any leftover space at this workspace path
+			// Clean up any leftover spaces (path comparison may differ due to macOS
+			// symlink resolution: /var/folders/ vs /private/var/folders/)
 			try {
 				const list = (await hub.request('space.list', {})) as Array<{
 					id: string;
 					workspacePath: string;
 				}>;
-				const existing = list.find((s) => s.workspacePath === wsPath);
-				if (existing) {
-					await hub.request('space.delete', { id: existing.id });
+				for (const s of list) {
+					await hub.request('space.delete', { id: s.id });
 				}
 			} catch {
 				// Ignore cleanup errors
@@ -71,6 +71,8 @@ async function deleteTestSpace(page: Page, spaceId: string): Promise<void> {
 async function navigateToSpace(page: Page, spaceId: string): Promise<void> {
 	await page.goto(`/space/${spaceId}`);
 	await page.waitForURL(`/space/${spaceId}**`, { timeout: 10000 });
+	// Wait for SpaceIsland to finish loading — the tab bar appears after space.overview resolves
+	await expect(page.locator('text=Dashboard').first()).toBeVisible({ timeout: 15000 });
 }
 
 /**
@@ -123,6 +125,8 @@ async function getDefaultAgentId(page: Page, spaceId: string): Promise<string> {
 // ─── Tests ────────────────────────────────────────────────────────────────────
 
 test.describe('Visual Workflow Editor', () => {
+	// All tests share the same workspace path (one space at a time) — must run serially
+	test.describe.configure({ mode: 'serial' });
 	test.use({ viewport: DESKTOP_VIEWPORT });
 
 	let spaceId = '';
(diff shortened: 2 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR correctly identifies and fixes the root causes of the E2E test failures. The changes are well-scoped and each fix addresses a specific, verifiable problem.

### App.tsx URL sync effect (core fix)

The main fix in `packages/web/src/App.tsx` is correct and well-reasoned. Before this change, when the page loaded at `/space/:id`, `initializeRouter()` would set `currentSpaceIdSignal`, but the STEP 4 effect had no space branches — it fell through to the `else` clause and called `navigateToHome(true)`, redirecting away from the space URL.

The fix adds space route handling (`spaceId`, `spaceSessionId`, `spaceTaskId`) to both the `expectedPath` computation and the navigation dispatch, matching the priority order already established in `initializeRouter()` and `handlePopState()` in `router.ts`.

I verified that the removal of `&& !spaceId` from the `navSection === 'spaces'` condition (line 114) is safe: by the time execution reaches that branch, `spaceId` is necessarily falsy (it would have matched earlier at line 106). The old `&& !spaceId` guard was redundant and its removal is correct.

No risk of unintended side effects on other routes — the new space branches are inserted before room branches in the priority chain, consistent with the existing pattern in `handlePopState` and `initializeRouter`.

### E2E test fixes

- **`navigateToSpace()` wait**: Adding `text=Dashboard` visibility wait after URL navigation is a solid fix for the loading race condition. The 15-second timeout is generous enough for CI.

- **`createTestSpace()` cleanup**: Deleting all spaces instead of matching by path handles the macOS `/var/` vs `/private/var/` symlink issue. This is pragmatic for test isolation since these tests own the workspace.

- **Serial mode**: Appropriate given tests share the same workspace path.

- **Node stagger** (20px to 100px): Correct fix — nodes are ~160x80px, so 20px stagger causes visual overlap that blocks Playwright clicks.

- **Edit button scoping**: Scoping to `workflowCard` instead of `page` prevents matching buttons from other workflow cards. Clean fix.

### Minor observations (P3)

1. The `createTestSpace` cleanup now deletes *all* spaces, not just ones matching the test workspace path. This is fine for test isolation (each test creates its own space in `beforeEach` and cleans up in `afterEach`), but could be surprising if the test infrastructure evolves to support parallel test suites with shared state. The intent is clearly documented in comments, so this is acceptable.

2. The `navigateToSpace` helper and `createTestSpace` cleanup logic are duplicated between `visual-workflow-editor.e2e.ts` and `space-workflow-rules.e2e.ts`. A future cleanup could extract these to a shared helper, but that's out of scope for this bug fix.

All changes are consistent with existing patterns in the codebase (router.ts priority ordering, E2E helper conventions). The fix is minimal and targeted.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 reliability issue and one P2 concern.

---

### Summary

The App.tsx fix is correct and well-reasoned. The root cause was accurately diagnosed: the old `expectedPath` computation lacked any space route handling, so loading at `/space/:id` caused `initializeRouter()` to set `currentSpaceIdSignal` while the effect computed `expectedPath = '/'` and immediately called `navigateToHome()`. The `navigateToSpace` loading wait, scoped `Edit` button locator, 100px node stagger, and serial test mode within each describe block all look correct.

---

### P1 — Concurrent file-level parallel execution breaks the "delete all spaces" cleanup

**Files:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts:44`, `packages/e2e/tests/features/space-workflow-rules.e2e.ts:42`

The Playwright config has `fullyParallel: true` with `workers: 4` for the `features` project (`playwright.config.ts:134,143`). `test.describe.configure({ mode: 'serial' })` only serialises tests *within* the describe block of a single file — it does not prevent two separate test *files* from running as concurrent workers.

Both `visual-workflow-editor.e2e.ts` and `space-workflow-rules.e2e.ts` now call:

```ts
for (const s of list) {
  await hub.request('space.delete', { id: s.id });
}
```

When these two files run concurrently (different browser contexts, same SQLite database), Worker A's `createTestSpace()` can delete the space Worker B just created, and vice versa. The old path-exact-match strategy at least limited the blast radius to the same workspace path. The new blanket delete removes that constraint.

This is an improvement in isolation *within* a single run of one file, but introduces a cross-file race condition. The correct fix is to make each file create its space with a uniquely-named suffix and clean up only that specific space (or spaces matching its own naming convention), rather than deleting everything. Alternatively, the E2E workspace paths should be per-worker-isolated (e.g. using a subdirectory unique to each Playwright worker).

---

### P2 — `navigateToSpace` in `router.ts` does not clear `currentRoomSessionIdSignal` / `currentRoomTaskIdSignal`

**File:** `packages/web/src/lib/router.ts:636-657`

When navigating to a space from a room+session context, `navigateToSpace()` clears `currentRoomIdSignal` but silently leaves `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal` set. Compare to `navigateToRoom()` (line 327-334) which clears both sub-signals. This is a pre-existing gap not introduced by this PR, but the new App.tsx effect now calls `navigateToSpace(..., true)` as part of the URL sync path, making the gap exercisable. The fix would be to add the two missing clears to both the early-return path (line 636-643) and the main path (line 648-658) of `navigateToSpace`. Similar gaps exist in `navigateToSpaceSession` and `navigateToSpaceTask`.

---

### P3 — Duplicate `navigateToSpace` helper function in both test files

**Files:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts:71`, `packages/e2e/tests/features/space-workflow-rules.e2e.ts:72`

The `navigateToSpace` helper (including the new `Dashboard` wait) is copy-pasted verbatim in both files. This is a minor DRY concern — moving it to the shared `wait-helpers.ts` would prevent the two copies drifting again in the future. Not blocking.

---

### What is correct

- **App.tsx `expectedPath` computation** — space branches are inserted at the correct priority level (before room branches), and the ternary chain is logically consistent.
- **`navSection === 'spaces'` guard removal** — the old `&& !spaceId` was never needed; when `spaceId !== null` the earlier branch fires, so the guard was dead code. The removal is safe.
- **`navigateToSpacesPage` loop risk** — none. The function guards against `currentPath === '/spaces'` before mutating history.
- **Node stagger fix (20px → 100px)** — correct; nodes are described as ~160×80px so 20px stagger guaranteed overlap on the second node.
- **`workflowCard.getByRole` scoping** — correct fix; the old page-scope locator matched hidden action buttons from other cards.

lsm (commented):

> **Recommendation: APPROVE** — both P1 and P2 concerns are fully and correctly addressed. No new issues introduced.

---

### P1 fix (cross-file cleanup race): RESOLVED

The blanket "delete all spaces" approach has been replaced with path-normalized matching in both E2E files. The `norm()` helper:

```ts
const norm = (p: string) => p.replace(/^\/private/, '');
const existing = list.find((s) => norm(s.workspacePath) === norm(wsPath));
```

is correctly scoped — only the space at this workspace path is deleted, not all spaces. It is applied consistently in both `packages/e2e/tests/features/visual-workflow-editor.e2e.ts` and `packages/e2e/tests/features/space-workflow-rules.e2e.ts`. The regex `/^\/private/` is sufficient for the macOS `/private/var` symlink case and does not incorrectly strip `/private` from paths that are not macOS symlink prefixes (no false positives on Linux, where `/private` never appears as a leading path component in practice).

The `serial` mode addition is also correctly placed in both `test.describe` blocks to prevent cross-test workspace conflicts.

---

### P2 fix (missing signal clears): RESOLVED

All 6 code paths in `packages/web/src/lib/router.ts` now clear `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal`:

- `navigateToSpace()`: early-return path (line 642-643) and main path (line 659-660)
- `navigateToSpaceSession()`: early-return path (line 686-687) and main path (line 703-704)
- `navigateToSpaceTask()`: early-return path (line 730-731) and main path (line 747-748)

This matches the symmetric pattern established by `navigateToRoom()` (lines 307-312 and 329-334), which clears `currentSpaceIdSignal`, `currentSpaceSessionIdSignal`, and `currentSpaceTaskIdSignal`. The fix is complete — no paths were missed.

---

### App.tsx URL sync effect change

The removal of `&& !spaceId` from the `navSection === 'spaces'` condition (line 137) is safe. The `spaceId` branch already sits higher in the ternary chain (line 106), so the `navSection === 'spaces'` arm is only reachable when `spaceId` is already falsy. This is a minor correctness improvement (the old guard was redundant and misleading).

The priority ordering in the expected path computation (space routes checked before room routes) is consistent with the priority order in `handlePopState()` in `router.ts` and is correct.

---

### No new issues identified

The `norm()` helper is defined inline inside `page.evaluate()`, which is correct — it runs in the browser context and cannot reference outer scope. The Dashboard visibility wait (`timeout: 15000`) is a reasonable upper bound for the `space.overview` RPC round-trip. The `serial` mode configuration is placed before `test.use()` in both files, consistent with Playwright's expected order.

All changes are minimal and focused. No dead code, no over-engineering.

lsm (commented):

### Re-review after P1 and P2 fixes

Both issues from the previous round have been addressed correctly.

**P1 (cleanup scope) — Resolved.** The `norm()` helper (`p.replace(/^\/private/, '')`) is a clean, targeted fix for the macOS `/private` symlink prefix mismatch. It only strips the prefix for comparison purposes — the actual workspace path passed to `space.create` is unchanged. This avoids the previous "delete all spaces" approach that could cause cross-test interference. The fix is applied consistently in both `visual-workflow-editor.e2e.ts` and `space-workflow-rules.e2e.ts`.

**P2 (missing signal clears) — Resolved.** All 6 code paths across the three space navigation functions (`navigateToSpace`, `navigateToSpaceSession`, `navigateToSpaceTask`) now clear `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal`. This matches the pattern established by `navigateToRoom()` and `handlePopState()`, both of which already cleared all 7 signals. The fix prevents stale room sub-signals from causing incorrect URL sync in the `App.tsx` effect.

### Additional changes verified

- **`test.describe.configure({ mode: 'serial' })`** added to both test files — correct since all tests share the same workspace path and cannot run in parallel.
- **Edit button scoping** (`workflowCard.getByRole` instead of `page.getByRole`) — good fix to avoid ambiguity when multiple workflow cards exist.
- **Node positioning** (`y: 80 + prev.length * 100` instead of diagonal stagger) — reasonable change to prevent visual overlap for vertically-listed nodes.
- **App.tsx URL sync effect** — the `expectedPath` computation and navigation calls now correctly handle all space route variants. The removal of `&& !spaceId` from the `navSection === 'spaces'` condition is correct because `spaceId` is already matched earlier in the chain.

1. The `createTestSpace` and `navigateToSpace` helper functions are duplicated between the two E2E test files. Consider extracting them into a shared helper (e.g., `tests/helpers/space-helpers.ts`). Not blocking — this is a minor DRY improvement.

**Recommendation: APPROVE** — All requested fixes are correctly implemented with no new issues introduced.

PR #568, opened 2026-03-21
feat: add CollapsibleSection component for sidebar sections

lsm:
## Summary

- Add reusable `CollapsibleSection` component with expand/collapse toggle, title, count badge, and `headerRight` slot for action buttons
- Export from `packages/web/src/components/room/index.ts` barrel file
- Add 8 unit tests covering expand/collapse toggle, count badge, headerRight slot, and default state

## Test plan

- [x] All 8 unit tests pass (`vitest run`)
- [x] Lint (oxlint) passes with 0 warnings/errors
- [x] Format (biome) passes
- [x] Knip reports no dead exports

### DIFF
--- a/packages/web/src/components/room/CollapsibleSection.test.tsx
+++ b/packages/web/src/components/room/CollapsibleSection.test.tsx
@@ -1,0 +1,109 @@
+/**
+ * Tests for CollapsibleSection Component
+ */
+
+import { describe, it, expect } from 'vitest';
+import { render, fireEvent, cleanup } from '@testing-library/preact';
+import { CollapsibleSection } from './CollapsibleSection';
+
+describe('CollapsibleSection', () => {
+	afterEach(() => cleanup());
+
+	it('renders title and children when expanded by default', () => {
+		const { getByText } = render(
+			<CollapsibleSection title="Goals">
+				<div>Goal content</div>
+			</CollapsibleSection>
+		);
+
+		expect(getByText('Goals')).toBeTruthy();
+		expect(getByText('Goal content')).toBeTruthy();
+		expect(getByText('▼')).toBeTruthy();
+	});
+
+	it('hides children when defaultExpanded is false', () => {
+		const { getByText, queryByText } = render(
+			<CollapsibleSection title="Sessions" defaultExpanded={false}>
+				<div>Session content</div>
+			</CollapsibleSection>
+		);
+
+		expect(getByText('Sessions')).toBeTruthy();
+		expect(queryByText('Session content')).toBeNull();
+		expect(getByText('▶')).toBeTruthy();
+	});
+
+	it('toggles children visibility on header click', () => {
+		const { getByText, queryByText, getByRole } = render(
+			<CollapsibleSection title="Tasks">
+				<div>Task content</div>
+			</CollapsibleSection>
+		);
+
+		expect(getByText('Task content')).toBeTruthy();
+
+		// Collapse
+		fireEvent.click(getByRole('button'));
+		expect(queryByText('Task content')).toBeNull();
+		expect(getByText('▶')).toBeTruthy();
+
+		// Expand again
+		fireEvent.click(getByRole('button'));
+		expect(getByText('Task content')).toBeTruthy();
+		expect(getByText('▼')).toBeTruthy();
+	});
+
+	it('renders count badge when count is provided', () => {
+		const { getByText } = render(
+			<CollapsibleSection title="Goals" count={5}>
+				<div>Content</div>
+			</CollapsibleSection>
+		);
+
+		expect(getByText('(5)')).toBeTruthy();
+	});
+
+	it('renders count badge with zero', () => {
+		const { getByText } = render(
+			<CollapsibleSection title="Goals" count={0}>
+				<div>Content</div>
+			</CollapsibleSection>
+		);
+
+		expect(getByText('(0)')).toBeTruthy();
+	});
+
+	it('does not render count badge when count is undefined', () => {
+		const { queryByText } = render(
+			<CollapsibleSection title="Goals">
+				<div>Content</div>
+			</CollapsibleSection>
+		);
+
+		expect(queryByText('(', { exact: false })).toBeNull();
+	});
+
+	it('renders headerRight slot', () => {
+		const { getByText } = render(
+			<CollapsibleSection title="Sessions" headerRight={<button type="button">+</button>}>
+				<div>Content</div>
+			</CollapsibleSection>
+		);
+
+		expect(getByText('+')).toBeTruthy();
+	});
+
+	it('headerRight click does not toggle section', () => {
+		const { getByText, queryByText } = render(
+			<CollapsibleSection title="Sessions" headerRight={<button type="button">+</button>}>
+				<div>Content</div>
+			</CollapsibleSection>
+		);
+
+		// Click the + button in headerRight
+		fireEvent.click(getByText('+'));
+
+		// Section should still be expanded (not toggled)
+		expect(queryByText('Content')).toBeTruthy();
+	});
+});
--- a/packages/web/src/components/room/CollapsibleSection.test.tsx
+++ b/packages/web/src/components/room/CollapsibleSection.test.tsx
@@ -2,7 +2,7 @@
  * Tests for CollapsibleSection Component
  */
 
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, afterEach } from 'vitest';
 import { render, fireEvent, cleanup } from '@testing-library/preact';
 import { CollapsibleSection } from './CollapsibleSection';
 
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Overall Assessment

The `CollapsibleSection` component is a clean, focused implementation. The code is readable, the tests cover the key behaviors, and the barrel export is properly added. However, there are two significant issues that should be addressed.

---

### P1: Missing `aria-expanded` attribute on the toggle button

**File:** `packages/web/src/components/room/CollapsibleSection.tsx`, line 31-35

The existing `Collapsible` component in `packages/web/src/components/ui/Collapsible.tsx` (line 59) sets `aria-expanded={isOpen}` on its button. The new `CollapsibleSection` does not. This is a meaningful accessibility gap — screen readers have no way to determine whether the section is expanded or collapsed.

The button should include `aria-expanded={expanded}`.

---

### P2: Potential duplication with existing `Collapsible` component

**File:** `packages/web/src/components/ui/Collapsible.tsx`

There is already a `Collapsible` component in the `ui/` directory that handles expand/collapse with:
- Controlled and uncontrolled modes (`open` + `onOpenChange`)
- `aria-expanded` accessibility
- Smooth height animation with transitions
- Custom className support

The new `CollapsibleSection` duplicates the core expand/collapse logic with `useState` instead of composing on top of the existing `Collapsible`. Consider whether `CollapsibleSection` could wrap `Collapsible` and add the sidebar-specific styling (title, count badge, headerRight) on top. This would avoid maintaining two independent collapse implementations and would inherit the accessibility and animation features for free.

If the intentional design is to keep these separate (e.g., `CollapsibleSection` uses conditional rendering for simpler DOM while `Collapsible` uses height animation), that is a valid choice but should be documented in a comment explaining why composition was not used.

---

### P3: Unicode triangle characters for expand/collapse indicator

**File:** `packages/web/src/components/room/CollapsibleSection.tsx`, line 37

Using `▶` / `▼` Unicode characters works but is less consistent across platforms and font stacks compared to the SVG chevron approach used by the existing `Collapsible` component. This is a minor cosmetic concern and acceptable if it matches the intended sidebar design.

---

### Summary

The implementation is functional and well-tested for its scope. The missing `aria-expanded` attribute (P1) should be fixed before merge. The relationship to the existing `Collapsible` component (P2) deserves consideration — either compose on top of it or document why a separate implementation is preferred.

**Recommendation: REQUEST_CHANGES** (due to P1 accessibility gap)

lsm (commented):

The component is functional and the core logic is correct, but there are meaningful issues with accessibility, divergence from the existing `ui/Collapsible` component, and a minor test correctness gap that should be addressed before merging.

---

### P1 - Missing ARIA accessibility attributes on the toggle button

`packages/web/src/components/room/CollapsibleSection.tsx` line 31-46

The existing `ui/Collapsible.tsx` component (which this new component parallels) sets `aria-expanded={isOpen}` on its trigger button. `CollapsibleSection` omits this entirely. A screen reader user clicking the button has no way to know whether the section is currently expanded or collapsed. `aria-expanded` is the standard ARIA pattern for disclosure widgets and is not optional for an interactive collapsible.

The button also has no `aria-label` or `aria-controls` attribute. Without them, the accessible name of the button is the concatenated text inside it (e.g., "▼ GOALS (3)"), which is not a meaningful label. At a minimum the button should express what it is toggling, and `aria-expanded` must be present.

Suggested minimal fix:
```tsx
<button
  type="button"
  aria-expanded={expanded}
  aria-label={`${title} section`}
  ...
>
```

---

### P1 - Duplicate of an existing component

`packages/web/src/components/ui/Collapsible.tsx` already provides a general-purpose collapsible with:
- Controlled/uncontrolled open state
- Animated height transition
- `aria-expanded`
- A `trigger` slot for arbitrary header content

`CollapsibleSection` is a narrower UI wrapper that could be composed on top of `ui/Collapsible` rather than re-implementing state management and toggle logic from scratch. Introducing a parallel implementation makes it harder to maintain consistent behavior (e.g., if animations are ever added to `ui/Collapsible`, `CollapsibleSection` will not pick them up). The PR description itself justifies the new component only as "used for Goals, Tasks, and Sessions sections" — this is a presentation concern, not a reason to duplicate toggle logic.

This is worth raising before the component gets consumed widely. Either compose it on top of `ui/Collapsible`, or document why the divergence is intentional.

---

### P2 - `headerRight` stop-propagation wrapper creates nested interactive elements

`packages/web/src/components/room/CollapsibleSection.tsx` lines 41-45

```tsx
{headerRight && (
  <div class="flex items-center" onClick={(e: MouseEvent) => e.stopPropagation()}>
    {headerRight}
  </div>
)}
```

This wraps arbitrary `headerRight` children in a `div` with a click handler inside a `<button>`. The HTML spec prohibits interactive content (buttons, links, inputs) inside a `<button>`. Placing a `<button>` as `headerRight` (as the test does on line 88 of the test file) results in an invalid DOM tree. The test at line 96-108 actually exercises this exact invalid nesting.

A correct approach is to place the header and the `headerRight` in a wrapping `div` that is separate from the collapsible `button`, rather than nesting them inside the button element.

---

### P2 - `count` prop type allows negative numbers without validation

The `count?: number` prop has no bounds check. Rendering `(-1)` would be visually misleading. Given the prop comment "used for Goals, Tasks, and Sessions sections," the semantic intent is a non-negative integer. Either document this in a JSDoc comment or narrow the type/add a guard.

---

### P3 - `afterEach` is used but not imported from vitest

`packages/web/src/components/room/CollapsibleSection.test.tsx` line 10

```ts
afterEach(() => cleanup());
```

`afterEach` is not in the import statement on line 5 (`import { describe, it, expect } from 'vitest'`). This works because `vitest.config.ts` sets `globals: true`, which injects vitest globals. However, all other test files in this package that use `afterEach` import it explicitly (e.g., `RoomTasks.test.tsx` line 8, `TaskView.test.tsx` line 11). The pattern in this file is inconsistent with the rest of the codebase and degrades IDE type-checking resolution.

---

### P3 - Bare CSS class names without design token usage

The component uses raw Tailwind classes like `hover:bg-dark-800`, `text-gray-500`, `text-gray-600`, `text-xs font-semibold uppercase tracking-wider`. Other sidebar components like `RoomContextPanel.tsx` also use raw Tailwind, so this is acceptable, but it is worth noting that if the project moves toward design tokens (as `ui/Collapsible.tsx` already uses via `borderColors`) these will need to be updated. Low priority but worth flagging for architectural consistency.

---

### Summary of issues

| Severity | Issue |
|----------|-------|
| P1 | Missing `aria-expanded` (and meaningful label) on toggle button |
| P1 | Duplicates `ui/Collapsible` toggle logic without composition |
| P2 | Invalid HTML: interactive content nested inside `<button>` via `headerRight` |
| P2 | `count` prop accepts negative numbers with no guard or documentation |
| P3 | `afterEach` used but not imported (relies on undocumented globals) |
| P3 | Raw Tailwind classes inconsistent with design token direction |

lsm (commented):

Re-review after previous feedback was addressed. All four issues have been resolved:

1. **aria-expanded** -- `aria-expanded={expanded}` is correctly set on the `<button>` element (line 40), with a dedicated test verifying it toggles between `'true'` and `'false'` (lines 56-69 of the test file).

2. **HTML nesting** -- `headerRight` is now a sibling `<div>` next to the `<button>`, not nested inside it (lines 37-48). This produces valid HTML. The test at lines 111-123 confirms clicking `headerRight` does not bubble into the toggle handler.

3. **afterEach import** -- Explicitly imported from `vitest` on line 5 of the test file.

4. **ui/Collapsible justification** -- Lines 7-10 of the component file clearly explain why `ui/Collapsible` is not composed: fixed SVG chevron, border/animation styles, and the sidebar's need for conditional rendering with a triangle indicator instead.

The component is well-structured, the test coverage is thorough (default states, toggling, aria attributes, count badge edge cases, headerRight isolation), and the code fits cleanly into the existing `components/room/` module with proper barrel exports.

lsm (commented):

**Recommendation: APPROVE**

This is a follow-up review after the worker addressed all four issues raised previously. I have verified each fix in full by reading the source files directly.

---

### Previous Issues — All Resolved

**P1 - aria-expanded** (`packages/web/src/components/room/CollapsibleSection.tsx`, line 40)

`aria-expanded={expanded}` is correctly set on the `<button>` element. The `aria-label` (`${title} section`) is also present. The test at lines 56-68 of the test file exercises both the initial `'true'` state and the post-click `'false'` state. Verified.

**P1 - Duplicate of ui/Collapsible** (`packages/web/src/components/room/CollapsibleSection.tsx`, lines 7-10)

The file-level block comment now clearly states the rationale: `ui/Collapsible` renders a fixed SVG chevron with border and height-animation styles that are inappropriate for the compact, information-dense sidebar layout. `CollapsibleSection` uses conditional rendering and a triangle indicator. This is a legitimate design divergence, and the reasoning is documented where reviewers will find it. Verified.

**P2 - Invalid HTML nesting** (`packages/web/src/components/room/CollapsibleSection.tsx`, lines 37-48)

The `headerRight` wrapper div is now a sibling of the toggle `<button>`, not nested inside it. The `<button>` closes on line 47, and `headerRight` is rendered on line 48 inside the shared flex container. Clicking `headerRight` content will not bubble into the toggle handler because the two elements share a flat parent, not an ancestor-descendant relationship. The test at lines 111-123 (passing a `<button>` as `headerRight`) now produces valid HTML and correctly verifies isolation. Verified.

**P3 - afterEach not imported** (`packages/web/src/components/room/CollapsibleSection.test.tsx`, line 5)

`afterEach` is now explicitly imported from `vitest` alongside `describe`, `it`, and `expect`. Verified.

---

### No New Issues Found

The implementation is clean, the test coverage is thorough (default state, toggle behavior, aria attributes, count badge with zero and undefined, headerRight slot and click isolation), and the barrel export in `packages/web/src/components/room/index.ts` is correct. The `count != null` guard correctly handles `0` as a valid renderable value.

PR #569, opened 2026-03-21
feat: add computed signals for goal-task grouping and orphan tasks

lsm:
## Summary
- Add `tasksByGoalId` computed signal mapping goal IDs to their linked `TaskSummary[]`
- Add `orphanTasks` computed signal for tasks not linked to any goal
- Add `orphanTasksActive`, `orphanTasksReview`, `orphanTasksDone` computed signals filtering orphan tasks by status category (all 7 `TaskStatus` values covered)
- Add comprehensive unit tests (11 test cases) covering all new signals

## Test plan
- [x] `bunx vitest run` passes all 11 new tests
- [x] `bun run typecheck` passes
- [x] `bun run lint` passes (0 errors)
- [x] `bun run format:check` passes

### DIFF
--- a/packages/web/src/lib/__tests__/room-store-computed-signals.test.ts
+++ b/packages/web/src/lib/__tests__/room-store-computed-signals.test.ts
@@ -14,7 +14,7 @@ import { describe, it, expect, vi, beforeEach } from 'vitest';
 // Mocks
 // -------------------------------------------------------
 
-let mockEventHandlers: Map<string, (event: unknown) => void>;
+let mockEventHandlers: Array<{ name: string; handler: (event: unknown) => void }>;
 let mockHub: ReturnType<typeof makeMockHub>;
 
 function makeMockHub() {
@@ -22,8 +22,12 @@ function makeMockHub() {
 		joinChannel: vi.fn(),
 		leaveChannel: vi.fn(),
 		onEvent: vi.fn((eventName: string, handler: (e: unknown) => void) => {
-			mockEventHandlers.set(eventName, handler);
-			return () => mockEventHandlers.delete(eventName);
+			const entry = { name: eventName, handler };
+			mockEventHandlers.push(entry);
+			return () => {
+				const idx = mockEventHandlers.indexOf(entry);
+				if (idx >= 0) mockEventHandlers.splice(idx, 1);
+			};
 		}),
 		request: vi.fn(async (method: string) => {
 			if (method === 'room.get') {
@@ -72,7 +76,7 @@ describe('RoomStore — computed goal/task signals', () => {
 	let roomStore: typeof import('../room-store').roomStore;
 
 	beforeEach(async () => {
-		mockEventHandlers = new Map();
+		mockEventHandlers = [];
 		mockHub = makeMockHub();
 		vi.resetModules();
 		const mod = await import('../room-store');
@@ -104,7 +108,18 @@ describe('RoomStore — computed goal/task signals', () => {
 			roomStore.tasks.value = [makeTask('t1', 'pending')];
 			roomStore.goals.value = [makeGoal('g1', ['t1', 'nonexistent'])];
 			const map = roomStore.tasksByGoalId.value;
+			expect(map.has('g1')).toBe(true);
 			expect(map.get('g1')?.map((t) => t.id)).toEqual(['t1']);
+		});
+
+		it('includes goal with empty linkedTaskIds as a key with empty array', () => {
+			roomStore.tasks.value = [makeTask('t1', 'pending')];
+			roomStore.goals.value = [makeGoal('g1', []), makeGoal('g2', ['t1'])];
+			const map = roomStore.tasksByGoalId.value;
+			expect(map.has('g1')).toBe(true);
+			expect(map.get('g1')).toEqual([]);
+			expect(map.has('g2')).toBe(true);
+			expect(map.get('g2')?.map((t) => t.id)).toEqual(['t1']);
 		});
 	});
 
--- a/packages/web/src/lib/__tests__/room-store-computed-signals.test.ts
+++ b/packages/web/src/lib/__tests__/room-store-computed-signals.test.ts
@@ -1,0 +1,224 @@
+// @ts-nocheck
+/**
+ * Tests for RoomStore computed signals:
+ * - tasksByGoalId: Map of goal ID → linked TaskSummary[]
+ * - orphanTasks: Tasks not linked to any goal
+ * - orphanTasksActive: Orphan tasks with draft/pending/in_progress
+ * - orphanTasksReview: Orphan tasks with review/needs_attention
+ * - orphanTasksDone: Orphan tasks with completed/cancelled
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+// -------------------------------------------------------
+// Mocks
+// -------------------------------------------------------
+
+let mockEventHandlers: Map<string, (event: unknown) => void>;
+let mockHub: ReturnType<typeof makeMockHub>;
+
+function makeMockHub() {
+	return {
+		joinChannel: vi.fn(),
+		leaveChannel: vi.fn(),
+		onEvent: vi.fn((eventName: string, handler: (e: unknown) => void) => {
+			mockEventHandlers.set(eventName, handler);
+			return () => mockEventHandlers.delete(eventName);
+		}),
+		request: vi.fn(async (method: string) => {
+			if (method === 'room.get') {
+				return { room: { id: 'room-1' }, sessions: [], allTasks: [] };
+			}
+			if (method === 'goal.list') return { goals: [] };
+			if (method === 'room.runtime.state') throw new Error('no runtime');
+			return {};
+		}),
+	};
+}
+
+vi.mock('../connection-manager.ts', () => ({
+	connectionManager: {
+		getHub: vi.fn(async () => mockHub),
+		getHubIfConnected: vi.fn(() => mockHub),
+	},
+}));
+
+// -------------------------------------------------------
+// Helpers
+// -------------------------------------------------------
+
+function makeTask(id: string, status: string, title = `Task ${id}`) {
+	return { id, title, status, priority: 'normal', progress: 0, dependsOn: [] };
+}
+
+function makeGoal(id: string, linkedTaskIds: string[] = []) {
+	return {
+		id,
+		roomId: 'room-1',
+		title: `Goal ${id}`,
+		description: '',
+		status: 'active',
+		priority: 'normal',
+		linkedTaskIds,
+		metrics: {},
+	};
+}
+
+// -------------------------------------------------------
+// Tests
+// -------------------------------------------------------
+
+describe('RoomStore — computed goal/task signals', () => {
+	let roomStore: typeof import('../room-store').roomStore;
+
+	beforeEach(async () => {
+		mockEventHandlers = new Map();
+		mockHub = makeMockHub();
+		vi.resetModules();
+		const mod = await import('../room-store');
+		roomStore = mod.roomStore;
+		await roomStore.select('room-1');
+	});
+
+	describe('tasksByGoalId', () => {
+		it('returns empty map when no goals', () => {
+			roomStore.tasks.value = [makeTask('t1', 'pending')];
+			roomStore.goals.value = [];
+			const map = roomStore.tasksByGoalId.value;
+			expect(map.size).toBe(0);
+		});
+
+		it('maps goals to their linked tasks', () => {
+			roomStore.tasks.value = [
+				makeTask('t1', 'pending'),
+				makeTask('t2', 'in_progress'),
+				makeTask('t3', 'completed'),
+			];
+			roomStore.goals.value = [makeGoal('g1', ['t1', 't2']), makeGoal('g2', ['t3'])];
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

The implementation adds 5 computed signals (`tasksByGoalId`, `orphanTasks`, `orphanTasksActive`, `orphanTasksReview`, `orphanTasksDone`) to `RoomStore` and 11 unit tests. The code is correct, well-structured, and aligns with the existing codebase patterns.

### Correctness

- The status categorization covers all 7 `TaskStatus` values (`draft`, `pending`, `in_progress`, `review`, `needs_attention`, `completed`, `cancelled`) with no gaps and no overlaps.
- `tasksByGoalId` correctly handles missing task IDs by silently skipping them (the `if (task)` guard at line 166).
- `orphanTasks` correctly builds a `Set` of all linked IDs across all goals before filtering, so a task linked to any goal is excluded.
- The derived signals (`orphanTasksActive`, `orphanTasksReview`, `orphanTasksDone`) correctly chain off `orphanTasks`, meaning they automatically exclude linked tasks without duplicating the filtering logic.

### Architecture

- Uses `computed()` from Preact Signals, consistent with the existing `activeGoals` computed signal on the same class.
- Intermediate `taskMap` (Map for O(1) lookup) in `tasksByGoalId` is a good choice for performance.
- `orphanTasks` builds the linked ID set independently from `tasksByGoalId` rather than depending on it, which avoids unnecessary recomputation when only orphan-related consumers are active. Both approaches are valid; this one is slightly more efficient for the orphan-only use case.

### Tests

- 11 tests cover the happy path, edge cases (empty goals, nonexistent task IDs, all tasks linked), and a comprehensive partition test verifying all 7 statuses fall into exactly one bucket with no overlap.
- `@ts-nocheck` is consistent with every other test file in `packages/web/src/lib/__tests__/`.
- The mock setup (`vi.resetModules()` + dynamic import) correctly isolates module-level singleton state between tests.

### Minor observations (P3)

1. **Missing test for task linked to multiple goals**: If a task appears in `linkedTaskIds` of more than one goal, `tasksByGoalId` will correctly include it under each goal, and `orphanTasks` will correctly exclude it. This is implicitly correct from the implementation but not explicitly tested. Very minor since the behavior follows directly from the data structures used.

2. **Status string literals are not enforced by TypeScript in the computed signals**: The filter comparisons use string literals (`'draft'`, `'pending'`, etc.) which are correct today but would not break the build if a status value were renamed in the `TaskStatus` type. This is mitigated by the `@ts-nocheck` in tests and by the fact that `TaskSummary.status` is typed as `TaskStatus` in the production code, so any mismatch would be caught by TypeScript in `room-store.ts` itself.

No P0, P1, or P2 issues found. Clean implementation that does exactly what was requested.

lsm (commented):

---

### Summary

This PR implements the 5 requested computed signals correctly and the logic is sound, but there are notable problems: all 11 tests are broken in CI when run outside the `packages/web` directory, the `@ts-nocheck` suppresses type safety in the test file, and the PR contains significant undisclosed scope — roughly 200 lines of code not mentioned in the description. The test mock also has a structural flaw that would silently hide bugs in event-driven scenarios.

---

### P0 — Broken Tests in CI

**File:** `packages/web/src/lib/__tests__/room-store-computed-signals.test.ts`

All 11 tests fail when run from the repo root (`bunx vitest run`) because vitest cannot resolve `@neokai/shared` without the `resolve.alias` configured in `packages/web/vitest.config.ts`. The tests only pass when run from inside `packages/web/`. The `make test-web` Makefile target presumably `cd`s into the right directory, but any CI step that runs vitest from the repo root will see 11 failures. Verified:

```
Error: Cannot find package '@neokai/shared' imported from .../room-store.ts
```

Every other web test file in this repo passes from the root because they are discovered via `make test-web`. Confirm that the CI matrix uses `make test-web` and not a root-level `bunx vitest run` to avoid this issue.

---

### P1 — `@ts-nocheck` Suppresses Type Safety

**File:** `packages/web/src/lib/__tests__/room-store-computed-signals.test.ts`, line 1

The `@ts-nocheck` directive silences TypeScript on the entire test file. The primary reason appears to be direct mutation of `readonly` class fields (`roomStore.tasks.value = ...`). In Preact Signals, `.value` is writable at runtime even on `readonly`-declared signals, but TypeScript enforces this as a type error.

The existing `room-store-review.test.ts` uses the same pattern with `@ts-nocheck`. This is a known workaround, but it means the test file gets zero type checking, including for the `makeGoal` helper which is missing required `RoomGoal` fields (`progress`, `createdAt`, `updatedAt`) — hidden by the directive.

Consider using `as unknown as ...` casts on just the signal assignments, or use a typed `setSignalForTest` helper, instead of blanket `@ts-nocheck`.

---

### P1 — Mock `onEvent` Silently Overwrites Handlers (Map keyed by event name)

**File:** `packages/web/src/lib/__tests__/room-store-computed-signals.test.ts`, lines 24-27

```ts
onEvent: vi.fn((eventName: string, handler: (e: unknown) => void) => {
    mockEventHandlers.set(eventName, handler);   // last write wins
    return () => mockEventHandlers.delete(eventName);
}),
```

`RoomStore.startSubscriptions` registers handlers for `goal.created`, `goal.updated`, `goal.completed` and multiple other events — but only one is registered for `goal.updated` (two could share a name). More critically, this mock can register a handler for `session.updated` and overwrite any earlier `session.updated` handler. Since no tests in this file trigger events through the mock, this is latent rather than a bug today, but it silently hides the problem for any future event-based test added to this file. The `room-store-session-events.test.ts` file handles this correctly by registering separate event handlers per call.

---

### P1 — PR Description Does Not Reflect Actual Scope

The PR title and description claim only 5 computed signals and 11 tests were added. The actual diff to `packages/web/src/lib/room-store.ts` adds approximately 200 additional lines:

- `autoCompletedNotifications` signal and `dismissAutoCompleted()` method
- `rejectTask()` method
- `listExecutions()` method
- `session.deleted` / `session.updated` event subscription handlers with room navigation logic
- `goal.task.auto_completed` event subscription handler
- Five new imports (`MissionType`, `AutonomyLevel`, `MissionMetric`, `CronSchedule`, `MissionExecution`, `navigateToRoom`, `currentRoomSessionIdSignal`)
- Extended `CreateGoalParams` with mission V2 fields

None of these are documented or tested in this PR. These additions are functionally important (e.g., `session.deleted` navigation logic, auto-complete notifications) but are left without test coverage and without mention in the description, making the PR difficult to reason about.

---

### P2 — Missing Test Case: Goal With Empty `linkedTaskIds`

**File:** `packages/web/src/lib/__tests__/room-store-computed-signals.test.ts`, `tasksByGoalId` describe block

The `tasksByGoalId` signal includes `result.set(goal.id, linked)` for all goals, including those with `linkedTaskIds: []`. There is no test verifying that a goal with an empty `linkedTaskIds` array produces an empty array in the map (rather than being absent). This is an edge case that could be added to the existing "returns empty map when no goals" test.

---

### P2 — `tasksByGoalId` Does Not Include Goals With `linkedTaskIds` Pointing to Non-Existent Tasks

The current behavior sets `result.set(goal.id, [])` when all `linkedTaskIds` point to tasks not in the local `tasks` signal. This is correct, but the test in line 103-108 conflates this with the "nonexistent IDs are skipped" case. A goal that is in `goals.value` should always appear as a key in `tasksByGoalId` even if its task list is empty — the test verifies the array content but does not verify the key is present, which would be a more complete assertion.

---

### The Computed Signal Logic Itself Is Correct

The five new signals (`tasksByGoalId`, `orphanTasks`, `orphanTasksActive`, `orphanTasksReview`, `orphanTasksDone`) implement the requested semantics correctly:

- `tasksByGoalId` uses an intermediate `Map<id, TaskSummary>` for O(n+m) complexity instead of O(n*m) nested loops — good.
- `orphanTasks` builds a `Set` of all linked IDs across all goals — correctly handles tasks linked to multiple goals.
- The three filtered orphan signals cover all 7 `TaskStatus` values with no overlap — confirmed by test line 206.
- Computed signals chain correctly: `orphanTasksActive/Review/Done` depend on `orphanTasks` which depends on `goals` and `tasks`.

lsm (commented):

**Recommendation: APPROVE**

### Follow-up fixes look good

The second commit (`bf0b502`) correctly addresses the three items from the previous review:

1. **onEvent mock (was P1)**: Changed from `Map<string, handler>` to `Array<{name, handler}>`. This correctly supports multiple handlers for the same event name without silent overwrites. The cleanup function uses `indexOf` on the exact entry reference, which is correct and efficient.

2. **Empty linkedTaskIds test (was P2)**: New test at line 115 verifies that a goal with `linkedTaskIds: []` appears as a key in `tasksByGoalId` with an empty array value, alongside a goal with actual linked tasks. Good coverage of this edge case.

3. **Key-presence assertions (was P2)**: Added `expect(map.has('g1')).toBe(true)` at line 111 in the "skips nonexistent task IDs" test, confirming the goal key exists even when some linked IDs are invalid.

### Previous review items no longer applicable

- **P0 (CI path)**: The previous review's concern about tests failing from the repo root was incorrect. `make test-web` runs `cd packages/web && bun run coverage`, so vitest resolves `@neokai/shared` correctly via the package-local config.
- **P1 (@ts-nocheck)**: This is the established pattern used by 20+ test files in `packages/web/src/lib/__tests__/`. Not a deviation from project conventions.
- **P1 (PR description scope)**: The diff only contains the 5 computed signals and their tests. The previous review's claim about "200 additional lines" of unrelated code appears to have been an error in analysis -- the actual diff to `room-store.ts` is exactly 48 lines of computed signal additions.

### Implementation remains correct

The 5 computed signals and 12 tests (11 original + 1 new) are well-structured. All 7 `TaskStatus` values are partitioned across the three orphan buckets with no overlap, verified by the comprehensive partition test. No issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### Follow-up Review — Addressing Previous Feedback

This is a follow-up to review [#3986217610](https://github.com/lsm/neokai/pull/569#pullrequestreview-3986217610). The worker addressed the three issues disputed and fixed the rest. All previous P0/P1/P2 findings are now resolved or confirmed as false alarms. Details below.

---

### Disputed Findings — Confirmed as False Alarms

**P0 (Test directory / CI breakage):** The claim was incorrect. `make test-web` runs `cd packages/web && bun run coverage`, which correctly sets the working directory. Vitest picks up path aliases from `packages/web/vitest.config.ts`. This is the same setup every other test in the repo uses, and all test files in `packages/web/src/lib/__tests__/` run under the same conditions. No CI breakage.

**@ts-nocheck (P1):** The worker's position is accurate. All 26 test files in `packages/web/src/lib/__tests__/` use `// @ts-nocheck`. This is the established project-wide convention for tests that must write to `readonly`-typed signals at runtime. The new test file follows the existing pattern consistently.

**PR scope (P1):** The previous review compared the diff against `main` instead of the base branch `dev`. When diffed against `dev` (the correct base), the change to `packages/web/src/lib/room-store.ts` is exactly 48 lines: the 5 new computed signals, nothing more. The PR description accurately reflects the scope.

---

### Fixes Verified — All Three Addressed Correctly

**1. Array-based onEvent mock (was P1 — Map silent overwrite):**

The fix in commit `bf0b502d` replaces the `Map<string, handler>` with `Array<{name, handler}>`. Each call to `onEvent` pushes a new entry, and the returned cleanup function removes exactly that entry via reference equality (`indexOf`). This correctly supports multiple handlers per event name, matching production behavior. The fix is correct.

**2. Empty `linkedTaskIds` test (was P2 — missing edge case):**

The new test at line 115–123 of the test file verifies:
- `map.has('g1')` is `true` when `linkedTaskIds` is `[]`
- `map.get('g1')` equals `[]`

This confirms the computed signal always includes goals as keys, even with no linked tasks.

**3. Key-presence assertions for `tasksByGoalId` (was P2 — incomplete assertions):**

The `'skips linked task IDs that do not exist'` test now asserts `map.has('g1')` before checking the array content. This makes the distinction between a missing key and an empty-array key explicit.

---

### Remaining Observations (Non-blocking)

**`room-store-review.test.ts` still uses Map-based mock:** The sibling test file at `packages/web/src/lib/__tests__/room-store-review.test.ts` still uses the old `Map<string, handler>` pattern with `mockEventHandlers.set(eventName, handler)` (line 23–24). This means that file still has the silent-overwrite problem for event-based tests. It is outside the scope of this PR, but worth a follow-up since that file triggers actual event handlers to test toast notifications.

**Coverage of `tasksByGoalId` for goals linked to non-existent tasks:** When all `linkedTaskIds` are non-existent, the goal still appears as a key with an empty array `[]`. This case is now covered by the empty `linkedTaskIds` test, which effectively validates the same map behavior. No gap remains.

---

### Implementation Quality

The 5 computed signals are implemented correctly:
- `tasksByGoalId` builds an intermediate `Map` for O(n+m) lookup rather than O(n*m) nested loops.
- `orphanTasks` builds a `Set` of all linked task IDs before filtering — handles tasks shared across goals correctly.
- `orphanTasksActive`, `orphanTasksReview`, `orphanTasksDone` chain from `orphanTasks` and cover all 7 `TaskStatus` values with no overlap.
- The 12 tests (11 original + 1 new) cover the full behavior including the mutual-exclusion property of the three buckets.

The PR is ready to merge.

PR #570, opened 2026-03-21
feat: add /room/:roomId/agent route pattern and navigation

lsm:
## Summary
- Add `ROOM_AGENT_ROUTE_PATTERN` (`/room/:roomId/agent`) and helpers (`createRoomAgentPath`, `getRoomAgentFromPath`)
- Add `navigateToRoomAgent(roomId)` that pushes the agent URL and sets the synthetic `room:chat:<roomId>` session signal so `ChatContainer` renders correctly
- Update `getRoomIdFromPath`, `handlePopState`, and `initializeRouter` to recognize the agent route on page load and back/forward navigation
- Update `RoomContextPanel` to use `navigateToRoomAgent` instead of `navigateToRoomSession` with a synthetic session ID

## Test plan
- [x] Unit tests for `createRoomAgentPath`, `getRoomAgentFromPath`, `getRoomIdFromPath` with agent paths
- [x] Unit tests for `navigateToRoomAgent` signal updates and history pushState
- [x] Unit tests for `initializeRouter` page-load signal restoration on agent route
- [x] Unit tests for popstate (back/forward) handling of agent route
- [x] Lint and format pass

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -24,6 +24,7 @@ import {
 	initializeRouter,
 	navigateToSession,
 	navigateToRoom,
+	navigateToRoomAgent,
 	navigateToRoomSession,
 	navigateToRoomTask,
 	navigateToHome,
@@ -33,6 +34,7 @@ import {
 	navigateToSpaceTask,
 	createSessionPath,
 	createRoomPath,
+	createRoomAgentPath,
 	createRoomSessionPath,
 	createRoomTaskPath,
 	createSpacePath,
@@ -97,6 +99,8 @@ export function App() {
 			const spaceTaskId = currentSpaceTaskIdSignal.value;
 			const navSection = navSectionSignal.value;
 			const currentPath = window.location.pathname;
+			// Detect agent route: synthetic session ID follows the pattern room:chat:<roomId>
+			const isAgentRoute = !!(roomSessionId && roomId && roomSessionId === `room:chat:${roomId}`);
 			const expectedPath = sessionId
 				? createSessionPath(sessionId)
 				: spaceTaskId && spaceId
@@ -107,15 +111,17 @@ export function App() {
 							? createSpacePath(spaceId)
 							: roomTaskId && roomId
 								? createRoomTaskPath(roomId, roomTaskId)
-								: roomSessionId && roomId
-									? createRoomSessionPath(roomId, roomSessionId)
-									: roomId
-										? createRoomPath(roomId)
-										: navSection === 'spaces'
-											? '/spaces'
-											: navSection === 'chats'
-												? '/sessions'
-												: '/';
+								: isAgentRoute
+									? createRoomAgentPath(roomId)
+									: roomSessionId && roomId
+										? createRoomSessionPath(roomId, roomSessionId)
+										: roomId
+											? createRoomPath(roomId)
+											: navSection === 'spaces'
+												? '/spaces'
+												: navSection === 'chats'
+													? '/sessions'
+													: '/';
 
 			// Only update URL if it's out of sync
 			// This prevents unnecessary history updates and loops
@@ -130,6 +136,8 @@ export function App() {
 					navigateToSpace(spaceId, true);
 				} else if (roomTaskId && roomId) {
 					navigateToRoomTask(roomId, roomTaskId, true);
+				} else if (isAgentRoute) {
+					navigateToRoomAgent(roomId, true);
 				} else if (roomSessionId && roomId) {
 					navigateToRoomSession(roomId, roomSessionId, true);
 				} else if (roomId) {
--- a/packages/web/src/islands/RoomContextPanel.tsx
+++ b/packages/web/src/islands/RoomContextPanel.tsx
@@ -11,7 +11,12 @@
 
 import { useMemo, useState } from 'preact/hooks';
 import { roomStore } from '../lib/room-store';
-import { navigateToRooms, navigateToRoom, navigateToRoomSession } from '../lib/router';
+import {
+	navigateToRooms,
+	navigateToRoom,
+	navigateToRoomAgent,
+	navigateToRoomSession,
+} from '../lib/router';
 import { currentRoomSessionIdSignal } from '../lib/signals';
 import { cn } from '../lib/utils';
 
@@ -71,7 +76,7 @@ export function RoomContextPanel({ roomId, onNavigate }: RoomContextPanelProps) 
 	};
 
 	const handleRoomAgentClick = () => {
-		navigateToRoomSession(roomId, roomAgentSessionId);
+		navigateToRoomAgent(roomId);
 		onNavigate?.();
 	};
 
--- a/packages/web/src/lib/__tests__/room-agent-router.test.ts
+++ b/packages/web/src/lib/__tests__/room-agent-router.test.ts
@@ -1,0 +1,205 @@
+// @ts-nocheck
+/**
+ * Tests for Room Agent URL routing
+ *
+ * Tests the /room/:roomId/agent route pattern including:
+ * - Path extraction and creation
+ * - Navigation function
+ * - popstate handling
+ * - Page-load initialization with correct signals
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import {
+	getRoomIdFromPath,
+	getRoomAgentFromPath,
+	createRoomAgentPath,
+	navigateToRoomAgent,
+	initializeRouter,
+	cleanupRouter,
+} from '../router';
+import {
+	currentSessionIdSignal,
+	currentRoomIdSignal,
+	currentRoomSessionIdSignal,
+	currentRoomTaskIdSignal,
+	currentSpaceIdSignal,
+	currentSpaceSessionIdSignal,
+	currentSpaceTaskIdSignal,
+	navSectionSignal,
+} from '../signals';
+
+let originalHistory: unknown;
+let originalLocation: unknown;
+let mockHistory: unknown;
+let mockLocation: unknown;
+
+describe('Room Agent Router', () => {
+	beforeEach(() => {
+		originalHistory = window.history;
+		originalLocation = window.location;
+
+		mockHistory = {
+			pushState: vi.fn(),
+			replaceState: vi.fn(),
+			state: null,
+		};
+
+		mockLocation = {
+			pathname: '/',
+		};
+
+		Object.defineProperty(window, 'history', {
+			value: mockHistory,
+			writable: true,
+			configurable: true,
+		});
+		Object.defineProperty(window, 'location', {
+			value: mockLocation,
+			writable: true,
+			configurable: true,
+		});
+
+		currentSessionIdSignal.value = null;
+		currentRoomIdSignal.value = null;
+		currentRoomSessionIdSignal.value = null;
+		currentRoomTaskIdSignal.value = null;
+		currentSpaceIdSignal.value = null;
+		currentSpaceSessionIdSignal.value = null;
+		currentSpaceTaskIdSignal.value = null;
+		navSectionSignal.value = 'home';
+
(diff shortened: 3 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured PR that adds a dedicated `/room/:roomId/agent` route. The implementation is correct and follows the established patterns in the router module.

### What was reviewed

- `packages/web/src/lib/router.ts` — New route pattern, path helpers, navigation function, popstate/init handling
- `packages/web/src/islands/RoomContextPanel.tsx` — Updated click handler to use `navigateToRoomAgent`
- `packages/web/src/lib/__tests__/room-agent-router.test.ts` — 11 unit tests

### Correctness

The implementation is correct:

1. **Route pattern** (`ROOM_AGENT_ROUTE_PATTERN`): Properly anchored regex with `$` end anchor, preventing false matches on paths like `/room/:id/agent/extra`.

2. **Ordering in `handlePopState` and `initializeRouter`**: The `roomAgent` branch is correctly placed **before** `roomTask`, `roomSession`, and the generic `roomId` branches. This is important because `getRoomIdFromPath` now matches agent routes too (via the fallback check added at line 70-72), so the more specific agent check must come first to avoid the agent route being handled as a plain room route.

3. **Signal state**: All seven signals are consistently set in all three code paths (navigate, popstate, initialize). The synthetic session ID `room:chat:${roomId}` is correctly applied to `currentRoomSessionIdSignal`.

4. **RoomContextPanel integration**: The change from `navigateToRoomSession(roomId, roomAgentSessionId)` to `navigateToRoomAgent(roomId)` is clean — the `roomAgentSessionId` variable is still used for the `isRoomAgentSelected` highlight check, which is correct.

### Minor observations (P3)

1. **`navSectionSignal` not set in same-path early return** (`router.ts` line 380-388): When `currentPath === targetPath`, the function sets all other signals but omits `navSectionSignal.value = 'rooms'`. This is a pre-existing pattern shared by `navigateToRoom` (line 325-334) and other navigate functions, so it is not a regression introduced by this PR. In practice this is harmless since if you are already on the agent path, the nav section should already be `'rooms'`.

2. **`@ts-nocheck` in test file** (`room-agent-router.test.ts` line 1): The test file uses `@ts-nocheck` to suppress TypeScript errors from the `window.history`/`window.location` mock overrides. This is pragmatic for test code, though typed mock patterns would be cleaner long-term.

### Test coverage

The 11 tests cover the key scenarios well:
- Path creation and extraction (positive and negative cases)
- `getRoomIdFromPath` recognizing the agent route
- Navigation with pushState, replaceState, and same-path idempotency
- `initializeRouter` signal initialization
- `popstate` event handling for back/forward navigation
- Negative test ensuring plain room route does NOT set agent signals

No missing test cases stand out for the scope of this change.

### Verdict

Clean implementation that follows existing patterns, has good test coverage, and integrates correctly with the existing routing infrastructure. No bugs or issues found.

lsm (commented):

> This is my own PR (same GitHub login), so posting as COMMENT. Recommendation: **REQUEST_CHANGES**.

---

## Summary

The router changes are well-structured and internally consistent — the pattern, extraction helpers, popstate handler, and initializeRouter all correctly handle the new `/room/:roomId/agent` route. However, there is a critical P0 regression in `App.tsx`'s URL sync effect that was not updated to recognize the new route, which will immediately corrupt the URL on every navigation to the agent view.

---

## P0 — URL Corruption: App.tsx Signal-Effect Overwrites Agent Route

**File:** `packages/web/src/App.tsx` (not modified in this PR)

The signal-effect in `App.tsx` (lines 82–123) syncs signals back to the URL when signals change from external sources. The problem: the `expectedPath` computation at lines 90–102 does not account for the agent route. When the user navigates to `/room/:roomId/agent`, the router correctly sets:

```
currentRoomIdSignal.value = roomId
currentRoomSessionIdSignal.value = 'room:chat:${roomId}'
```

The effect re-runs and evaluates:

```ts
const expectedPath = /* ... */ roomSessionId && roomId
  ? createRoomSessionPath(roomId, roomSessionId)   // → '/room//session/room:chat:'
  : ...
```

Since `currentPath ('/room//agent') !== expectedPath ('/room//session/room:chat:')`, the effect calls:

```ts
navigateToRoomSession(roomId, 'room:chat:${roomId}', true)
```

This replaceState overwrites the clean `/room/:roomId/agent` URL with the malformed `/room/:roomId/session/room:chat:${roomId}` URL immediately after every navigation to the agent view. Deep-linking to `/room/:roomId/agent` is not affected (initializeRouter fires before the effect), but any click-triggered navigation will be corrupted.

**Before this PR** this was not a problem: `handleRoomAgentClick` called `navigateToRoomSession(roomId, 'room:chat:${roomId}')`, which set the URL to `/room//session/room:chat:` and both the current path and computed expected path agreed.

**Fix required:** `App.tsx` must be updated to check for the agent synthetic session ID before computing `expectedPath`, and call `navigateToRoomAgent` in the update block when appropriate.

---

## P1 — `getRoomIdFromPath` Double-Matches Agent Paths (Latent Correctness Risk)

**File:** `packages/web/src/lib/router.ts`, lines 67–86

`getRoomIdFromPath` now matches the agent path (via its internal delegation to `ROOM_AGENT_ROUTE_PATTERN`), meaning for `/room/:roomId/agent`:
- `roomId` is non-null
- `roomAgent` is also non-null

In `handlePopState` and `initializeRouter`, the `roomAgent` branch is checked before the plain `roomId` branch, so the correct code executes. But this is fragile: if anyone reorders the else-if chain or adds a new check between them, the agent route will silently fall through to the plain room handler, stripping the synthetic session ID from the signal.

A cleaner approach would be to not have `getRoomIdFromPath` match agent paths (since callers that want the room ID from an agent path can call `getRoomAgentFromPath`). Or at minimum add a comment on the if-else order stating it is load-bearing.

---

## P2 — Module-Level File Comment Not Updated

**File:** `packages/web/src/lib/router.ts`, lines 1–13

The top-of-file doc comment still lists only:
- Sessions: `/session/:sessionId`
- Rooms: `/room/:roomId`

It should also list `/room/:roomId/agent`. Minor but misleading for future maintainers.

---

## P3 — No Test for App.tsx URL Sync Interaction

The 11 unit tests all test the router module in isolation. None verify the interaction with `App.tsx`'s signal-effect — which is precisely the gap that hides the P0 bug above. An integration test or at least a comment in the PR description noting the dependency would reduce risk.

---

## What's Done Well

- The regex pattern `/^\/room\/([a-f0-9-]+)\/agent$/` is anchored correctly and rejects trailing content (`/agent/extra` returns null).
- Both `handlePopState` and `initializeRouter` correctly set all six signals atomically.
- The `navigateToRoomAgent` function respects the `isNavigating` guard and supports `replace=true`.
- The `RoomContextPanel` update is clean — `handleRoomAgentClick` now uses the dedicated function.
- Test coverage for the router module itself is thorough across page-load, popstate, and same-path re-navigation scenarios.

lsm (commented):

> This is my own PR (same GitHub login), so posting as COMMENT. Recommendation: **APPROVE**.

---

## Re-review: Previous Issues Addressed

This is a follow-up review after the P0/P1/P2 feedback from the previous round was addressed.

### P0 Fix Verification (App.tsx URL sync)

The fix in `packages/web/src/App.tsx` is **correct and complete**. The previous P0 was that the signal-effect URL sync would overwrite `/room/:roomId/agent` with a malformed `/room/:roomId/session/room:chat:...` path. The fix works as follows:

1. **Detection** (line 93): `isAgentRoute` is computed by checking if `roomSessionId` matches the synthetic pattern `room:chat:${roomId}`. This correctly identifies when the agent view is active.

2. **expectedPath computation** (lines 98-99): `isAgentRoute` is checked **before** the generic `roomSessionId && roomId` branch in the ternary chain. This ensures `createRoomAgentPath(roomId)` produces the expected path (`/room/:roomId/agent`) instead of `createRoomSessionPath` producing a malformed URL.

3. **URL update block** (lines 117-118): `navigateToRoomAgent(roomId, true)` is called in the corresponding position, matching the same ordering as `expectedPath`.

The ordering is correct: `sessionId` > `roomTaskId` > `isAgentRoute` > `roomSessionId` > `roomId`. The agent check must come before the generic room session check since the agent route uses a synthetic session ID that would otherwise match `roomSessionId && roomId`.

### P1 Fix Verification (Ordering comments)

Comments added at `handlePopState` (line 834-836) and `initializeRouter` (line 971) clearly explain why `roomAgent` must be checked before other room routes. The explanation that `getRoomIdFromPath` also matches agent paths is accurate and documents the subtle ordering dependency well.

### P2 Fix Verification (Module doc comment)

The module-level doc comment at line 7 now lists `/room/:roomId/agent` alongside the other route patterns.

---

## Remaining Observations (P3 only)

1. **`navSectionSignal` not set in same-path early return** (`router.ts` lines 381-389): The early-return path in `navigateToRoomAgent` omits setting `navSectionSignal.value = 'rooms'`. This is a pre-existing pattern shared by `navigateToRoom` (lines 326-335) and is harmless in practice since if you are already on the agent path, the nav section is already `'rooms'`.

2. **`@ts-nocheck` in test file**: Pragmatic for the window mock overrides but could be replaced with typed mocks long-term.

---

## Verdict

All three previous issues (P0, P1, P2) have been correctly addressed. The App.tsx fix properly prevents the URL sync effect from corrupting the agent route URL. The implementation is correct, complete, and well-tested with 11 unit tests covering path helpers, navigation, initialization, and popstate handling.

lsm (commented):

**Recommendation: APPROVE**

This is a follow-up review verifying fixes to the previously-identified issues. The implementation is correct and complete.

---

### P0 Fix Verification: `isAgentRoute` detection in App.tsx

The fix is sound. The detection logic on line 93:

```ts
const isAgentRoute = !!(roomSessionId && roomId && roomSessionId === `room:chat:${roomId}`);
```

Correctly identifies the agent route because `room:chat:${roomId}` is not just a frontend convention — it is a real session ID generated by the daemon in `packages/daemon/src/lib/rpc-handlers/room-handlers.ts` (line 61) and `room-runtime-service.ts` (line 442). The check is a reliable discriminant: no regular session ID would match this pattern since daemon-generated session IDs are plain UUIDs.

The ordering in both `expectedPath` and the navigation dispatch block is correct — `isAgentRoute` is checked before the general `roomSessionId && roomId` branch, preventing agent routes from being mis-mapped to `/room/:roomId/session/room:chat:${roomId}`.

---

### Previously-Requested Changes: Confirmed Resolved

- **P0 (App.tsx URL overwrite)**: Fixed. The synthetic session ID is detected and `createRoomAgentPath`/`navigateToRoomAgent` are used instead of `createRoomSessionPath`/`navigateToRoomSession`.
- **P1 (ordering comments)**: Added in `handlePopState` and `initializeRouter`, clearly explaining why `roomAgent` must precede `roomTask`/`roomSession`/`roomId`.
- **P2 (module-level doc comment)**: `/room/:roomId/agent` is now listed on line 7 of `router.ts`.

---

**`navigateToRoomAgent` same-path branch omits `navSectionSignal` update** (`packages/web/src/lib/router.ts` lines 381-389): The early-return branch for "already on this path" does not set `navSectionSignal.value = 'rooms'`, while the full navigation path does (line 405). This is consistent with how `navigateToRoom` handles the same case (also omits it in the same-path branch), so it is a pre-existing pattern rather than a regression. In practice, if you are already on the agent URL, `navSectionSignal` should already be 'rooms'. No action required.

The test for the same-path branch (`packages/web/src/lib/__tests__/room-agent-router.test.ts` lines 145-155) does not assert `navSectionSignal`, which is consistent with how the function behaves.

---

### Test Coverage

Tests cover all required scenarios: path creation, path extraction, `getRoomIdFromPath` recognition, navigation (push and replace), same-path idempotency, page-load initialization via `initializeRouter`, and popstate (back/forward) handling. Coverage is adequate.

PR #572, opened 2026-03-21
test: unit tests for goalId propagation (Task 5.3)

lsm:
## Summary
- Add goalId propagation tests to `workflow-executor.test.ts` verifying tasks inherit `goalId` from their workflow run through `advance()` / `followTransition()`, including multi-step and cyclic workflow scenarios
- Add `space-task-findByGoalId.test.ts` testing the `findByGoalId()` repository method: correct filtering, archived task exclusion, null goalId handling, and ordering

## Test plan
- [x] All 60 tests pass across both files (`bun test`)
- [x] Pre-commit checks pass (lint, format, typecheck, knip)

### DIFF
--- a/packages/daemon/tests/unit/space/space-task-findByGoalId.test.ts
+++ b/packages/daemon/tests/unit/space/space-task-findByGoalId.test.ts
@@ -1,0 +1,248 @@
+/**
+ * SpaceTaskRepository.findByGoalId() Unit Tests
+ *
+ * Covers:
+ * - Finding tasks by goalId
+ * - Excluding archived tasks
+ * - Handling null/undefined goalId gracefully
+ * - Ordering by created_at ASC
+ */
+
+import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
+import { rmSync, mkdirSync } from 'node:fs';
+import { join } from 'node:path';
+import { Database as BunDatabase } from 'bun:sqlite';
+import { runMigrations } from '../../../src/storage/schema/index.ts';
+import { SpaceTaskRepository } from '../../../src/storage/repositories/space-task-repository.ts';
+
+// ---------------------------------------------------------------------------
+// Test DB helpers
+// ---------------------------------------------------------------------------
+
+function makeDb(): { db: BunDatabase; dir: string } {
+	const dir = join(process.cwd(), 'tmp', 'test-findByGoalId', `t-${Date.now()}-${Math.random()}`);
+	mkdirSync(dir, { recursive: true });
+	const db = new BunDatabase(join(dir, 'test.db'));
+	db.exec('PRAGMA foreign_keys = ON');
+	runMigrations(db, () => {});
+	return { db, dir };
+}
+
+const SPACE_ID = 'space-1';
+
+function seedSpace(db: BunDatabase): void {
+	db.prepare(
+		`INSERT INTO spaces (id, workspace_path, name, description, background_context, instructions,
+     allowed_models, session_ids, status, created_at, updated_at)
+     VALUES (?, ?, ?, '', '', '', '[]', '[]', 'active', ?, ?)`
+	).run(SPACE_ID, '/tmp/ws', 'Test Space', Date.now(), Date.now());
+}
+
+// ---------------------------------------------------------------------------
+// Test suite
+// ---------------------------------------------------------------------------
+
+describe('SpaceTaskRepository.findByGoalId', () => {
+	let db: BunDatabase;
+	let dir: string;
+	let repo: SpaceTaskRepository;
+
+	beforeEach(() => {
+		({ db, dir } = makeDb());
+		seedSpace(db);
+		repo = new SpaceTaskRepository(db);
+	});
+
+	afterEach(() => {
+		try {
+			db.close();
+		} catch {
+			/* ignore */
+		}
+		try {
+			rmSync(dir, { recursive: true, force: true });
+		} catch {
+			/* ignore */
+		}
+	});
+
+	test('returns tasks with matching goalId', () => {
+		repo.createTask({
+			spaceId: SPACE_ID,
+			title: 'Task 1',
+			description: 'desc 1',
+			goalId: 'goal-A',
+		});
+		repo.createTask({
+			spaceId: SPACE_ID,
+			title: 'Task 2',
+			description: 'desc 2',
+			goalId: 'goal-A',
+		});
+
+		const tasks = repo.findByGoalId('goal-A');
+		expect(tasks).toHaveLength(2);
+		expect(tasks[0].title).toBe('Task 1');
+		expect(tasks[1].title).toBe('Task 2');
+	});
+
+	test('returns empty array when no tasks match the goalId', () => {
+		repo.createTask({
+			spaceId: SPACE_ID,
+			title: 'Task 1',
+			description: 'desc',
+			goalId: 'goal-A',
+		});
+
+		const tasks = repo.findByGoalId('goal-nonexistent');
+		expect(tasks).toHaveLength(0);
+	});
+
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR adds well-structured unit tests for goalId propagation through the workflow executor and the `findByGoalId` repository method. All 60 tests pass. The tests correctly exercise the production code paths and cover meaningful scenarios.

### What works well

- **Workflow executor tests** properly test through the public `advance()` API rather than reaching into the private `followTransition` method, which is the right approach.
- **Good coverage of the negative case**: verifying `undefined` goalId when the run has no goalId set, both for single-step and multi-step workflows.
- **Cyclic workflow test** is a valuable edge case that exercises the propagation across repeated visits to the same step.
- **findByGoalId tests** cover the key behaviors of the SQL query: goalId filtering, archived exclusion, ordering, and cross-space behavior.
- Both files follow the existing test patterns (DB setup/teardown, helper functions) consistently with the rest of the test suite.

### Issues

**P3 (nit):**

1. **Misleading test title** in `space-task-findByGoalId.test.ts` line 222: The test is named `'returns tasks with goalId from correct space only'` but the body and inline comment confirm that `findByGoalId` does NOT filter by space — it returns tasks from all spaces. The title should be something like `'returns tasks across all spaces for the same goalId'` to match the actual assertion (`expect(tasks).toHaveLength(2)`). The inline comment on line 244 is correct; it's just the title that contradicts the behavior.

No correctness, security, or architectural concerns.

lsm (commented):

**Recommendation: APPROVE** (same-owner PR, posting as COMMENT)

All 60 tests pass and the implementations are correctly tested. The goalId propagation is verified end-to-end: `run.goalId → followTransition → createTask → task.goalId`, and `findByGoalId` SQL filtering is properly exercised. There are a few quality issues worth noting before this ships.

---

### P2 — Misleading test name in `space-task-findByGoalId.test.ts`

**File:** `packages/daemon/tests/unit/space/space-task-findByGoalId.test.ts`, line 222

The test is named:
```
'returns tasks with goalId from correct space only'
```

But the assertion directly contradicts this name:
```ts
// findByGoalId does not filter by space — it returns all matching tasks
const tasks = repo.findByGoalId('shared-goal');
expect(tasks).toHaveLength(2);  // both spaces returned
```

The test is actually verifying that `findByGoalId` is *space-agnostic* (returns tasks across all spaces), which is the opposite of what the name implies. A future reader will be confused about what the intended behavior is. The name should be something like `'returns tasks across all spaces with matching goalId'` or `'does not filter by space'`.

---

### P2 — Ordering test is potentially flaky under fast hardware

**File:** `packages/daemon/tests/unit/space/space-task-findByGoalId.test.ts`, lines 166–193

The three `createTask` calls all happen synchronously within the same event-loop tick. `createTask` uses `Date.now()` for `created_at`, so all three rows can easily receive the same millisecond timestamp. The SQL query is:

```sql
SELECT * FROM space_tasks WHERE goal_id = ? AND archived_at IS NULL ORDER BY created_at ASC
```

With identical timestamps, SQLite's sort is undefined. The test has passed 5/5 runs locally (likely due to SQLite's B-tree insertion order being stable in practice), but this is implementation-dependent and not guaranteed. A tiebreaker on `rowid` or `id` in the `ORDER BY` clause of the repository method would make ordering deterministic, or the test could assert on IDs instead of positional order if the ordering guarantee is intentional.

Note: the implementation's ordering behavior is correct for production use (timestamps will have real gaps there), but the test over-specifies it in a way that can't be reliably reproduced synchronously.

---

### P3 — First goalId propagation test uses verbose manual setup instead of `createLinearWorkflow`

**File:** `packages/daemon/tests/unit/space/workflow-executor.test.ts`, lines 1068–1103

The test `'task created by advance() inherits goalId from run'` manually constructs `stepsData` and `transitions` via map/array literals, while the immediately following tests (lines 1105+) all use `createLinearWorkflow`. The first test could use `createLinearWorkflow` + a direct `runRepo.createRun` override for `goalId`, as the other tests do, making it consistent with the surrounding code.

---

### Summary

The core behavior is correctly tested: goalId flows from workflow run → task through both linear and cyclic paths, and `findByGoalId` correctly excludes archived tasks, isolates between goal IDs, and handles null goalId. The P2 issues are worth addressing before this merges — the misleading test name in particular could mislead future maintainers about the intended cross-space behavior of `findByGoalId`.

lsm (commented):

**Recommendation: APPROVE**

All three feedback items from the previous review have been properly addressed:

1. **Test rename (P2 fixed):** `'returns tasks with goalId from correct space only'` is now `'returns tasks across all spaces with matching goalId'` -- accurately describes the cross-space behavior being tested.

2. **Flaky ordering test (P2 fixed):** The test at line 166 now uses `Set` membership assertions instead of positional order checks, correctly avoiding reliance on undefined SQLite sort behavior when `created_at` timestamps are identical. The test name was also updated to `'returns all tasks with matching goalId regardless of creation timing'` which clearly communicates the intent.

3. **Verbose first test (P3 fixed):** The first goalId propagation test in `workflow-executor.test.ts` (line 1068) now uses `createLinearWorkflow` for the workflow setup and only creates the run separately via `runRepo.createRun` to set `goalId`. This is clean and consistent with the other tests in the file.

All 60 tests pass. The test coverage for goalId propagation is thorough: linear workflows, multi-step workflows, cyclic transitions, and the negative case (no goalId set). The `findByGoalId` repository tests cover filtering, archived exclusion, cross-space behavior, null goalId, and mixed statuses.

### P3 (nit)

The first test `'returns tasks with matching goalId'` (line 69-87 of `space-task-findByGoalId.test.ts`) still uses positional assertions (`tasks[0].title` / `tasks[1].title`) for two tasks created in the same tick, which has the same theoretical ordering concern that was fixed in the three-task test. In practice this is harmless with two items, but it is a minor inconsistency.

No blocking issues found.

lsm (commented):

**Recommendation: APPROVE** (same-owner PR, posting as COMMENT)

All three issues raised in my previous review have been addressed correctly. The implementation is clean and the tests are well-structured.

---

### Verification of addressed feedback

**P2 — Test rename:** `'returns tasks across all spaces with matching goalId'` (line 222 of `space-task-findByGoalId.test.ts`) now accurately describes the behavior being tested. The comment `// findByGoalId does not filter by space — it returns all matching tasks` confirms the intent is clear.

**P2 — Flaky ordering test:** Lines 166–193 now use a `Set` membership assertion (`ids.has(t1.id)` etc.) with an updated test name `'returns all tasks with matching goalId regardless of creation timing'`. This correctly sidesteps the undefined SQLite sort order when all three tasks share the same `Date.now()` timestamp.

**P3 — First goalId propagation test:** Lines 1068–1088 of `workflow-executor.test.ts` now use `createLinearWorkflow` for workflow setup and only create the run separately to inject `goalId: 'goal-123'`. This is consistent with all other tests in the `goalId propagation` describe block.

---

### Remaining nit (P3)

The test `'returns tasks with matching goalId'` at line 69–87 of `space-task-findByGoalId.test.ts` still asserts positional order (`tasks[0].title === 'Task 1'`, `tasks[1].title === 'Task 2'`) for two tasks created synchronously in the same tick. This has the same theoretical timestamp-collision concern that was fixed in the three-task version. In practice this is harmless since SQLite's B-tree insertion order is stable for two items, and this was already flagged by the previous opus review as an acceptable minor inconsistency. It does not block merge.

---

### Summary

GoalId propagation is thoroughly tested: linear single-step, linear multi-step, no-goalId cases, and cyclic transitions all pass through `advance() → followTransition → createTask`. The `findByGoalId` repository tests correctly cover filtering, archived exclusion, cross-space behavior, null goalId handling, and mixed statuses. No blocking issues.

PR #562, opened 2026-03-21
feat: add task_result condition type and isCyclic to workflow transitions

lsm:
## Summary

- Add `'task_result'` to `WorkflowConditionType` union type with JSDoc describing its semantics
- Add `isCyclic?: boolean` to `WorkflowTransition` and `ExportedWorkflowTransition` interfaces
- Update Zod schemas (`workflowConditionSchema`, `exportedWorkflowTransitionSchema`) and export logic to handle `task_result` and `isCyclic`
- Add migration 34: `is_cyclic INTEGER` column on `space_workflow_transitions`
- Update `SpaceWorkflowRepository` (`insertTransition`, `rowToTransition`, `TransitionRow`) for DB persistence
- Add stub `task_result` case in `evaluateCondition()` (full implementation in Task 1.2)
- Update UI `Record<WorkflowConditionType, string>` maps in 4 web components
- Fix `EdgeRenderer` test marker count (4 → 5)

## Test plan

- [x] Migration 34 test: fresh DB column exists, defaults to NULL, can be set to 1, legacy DB path, idempotency
- [x] Repository round-trip test: isCyclic true/false persists through insert/read, task_result condition persists
- [x] Export format test: isCyclic in exported transitions, task_result condition validation, Zod preserves isCyclic
- [x] EdgeRenderer test updated for 5 markers
- [x] Typecheck passes (executor stub added for task_result)
- [x] Lint, format, and knip all pass

PR #573, opened 2026-03-21
feat: add Verify & Test step to Coding Workflow template

lsm:
## Summary

- Extends the `CODING_WORKFLOW` template from 2 steps (Plan → Code) to 4 steps (Plan → Code → Verify & Test → Done) with cyclic transitions
- Adds `task_result` condition transitions: Verify→Plan on `failed` (with `isCyclic: true`) and Verify→Done on `passed`
- Sets `maxIterations: 3` on the template as a safety cap for iteration loops
- Updates `seedBuiltInWorkflows()` to forward `isCyclic` and `maxIterations` fields that were previously silently dropped

## Test plan

- [x] Updated existing CODING_WORKFLOW template tests for 4-step structure
- [x] Added tests for `isCyclic` flag on Verify→Plan transition
- [x] Added tests for `maxIterations` forwarding through seeding
- [x] Added tests for Verify step instructions
- [x] All 44 existing tests pass
- [x] Typecheck passes
- [x] Lint and format pass

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -305,6 +305,7 @@ export class SpaceRuntime {
 			description,
 			currentStepId: workflow.startStepId,
 			goalId,
+			maxIterations: workflow.maxIterations,
 		});
 
 		const run = this.config.workflowRunRepo.updateStatus(pendingRun.id, 'in_progress')!;
--- a/packages/daemon/src/lib/space/runtime/workflow-executor.ts
+++ b/packages/daemon/src/lib/space/runtime/workflow-executor.ts
@@ -234,7 +234,9 @@ export class WorkflowExecutor {
 			}
 
 			case 'task_result': {
-				// Task 1.2 implements full evaluation — for now, always fail
+				// TODO(Task 1.2): Implement prefix matching against completed task's result field.
+				// Until then, Verify step transitions (task_result conditions) will not fire,
+				// causing the run to go to needs_attention. This is intentional for phased rollout.
 				return { passed: false, reason: 'task_result evaluation not yet implemented' };
 			}
 
--- a/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
+++ b/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
@@ -71,6 +71,7 @@ export const CODING_WORKFLOW: SpaceWorkflow = {
 			id: CODING_VERIFY_STEP,
 			name: 'Verify & Test',
 			agentId: 'general',
+			// NOTE: task_result condition uses prefix matching — "failed: reason" matches expression "failed"
 			instructions:
 				'Review the completed work. Run tests, check for issues. Set result to "passed" if everything looks good, or "failed: <reason>" if problems are found.',
 		},
--- a/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
+++ b/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
@@ -27,6 +27,8 @@ import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 
 const CODING_PLANNER_STEP = 'tpl-coding-planner';
 const CODING_CODER_STEP = 'tpl-coding-coder';
+const CODING_VERIFY_STEP = 'tpl-coding-verify';
+const CODING_DONE_STEP = 'tpl-coding-done';
 
 const RESEARCH_PLANNER_STEP = 'tpl-research-planner';
 const RESEARCH_GENERAL_STEP = 'tpl-research-general';
@@ -40,17 +42,20 @@ const REVIEW_CODER_STEP = 'tpl-review-coder';
 /**
  * Coding Workflow
  *
- * Two-node graph: Planner → Coder.
- * - The Planner → Coder transition uses a `human` condition: a human must
- *   approve the plan before implementation begins.
- * - The Coder step has no outgoing transitions — it is the terminal node.
- *   The run completes when advance() is called from the Coder step.
+ * Four-node graph: Plan → Code → Verify → Done (with cycle).
+ * - Plan → Code: `human` condition — a human must approve the plan.
+ * - Code → Verify: `always` condition — automatically verify after coding.
+ * - Verify → Plan: `task_result` condition on 'failed' — loops back (cyclic).
+ * - Verify → Done: `task_result` condition on 'passed' — completes the workflow.
+ * - `maxIterations: 3` caps the number of Plan→Code→Verify cycles.
  */
 export const CODING_WORKFLOW: SpaceWorkflow = {
 	id: '',
 	spaceId: '',
 	name: 'Coding Workflow',
-	description: 'Plan-first coding workflow. A human reviews the plan before implementation starts.',
+	description:
+		'Plan-first coding workflow with verification. A human reviews the plan, code is implemented, then verified. Loops back on failure.',
+	maxIterations: 3,
 	steps: [
 		{
 			id: CODING_PLANNER_STEP,
@@ -61,6 +66,18 @@ export const CODING_WORKFLOW: SpaceWorkflow = {
 			id: CODING_CODER_STEP,
 			name: 'Code',
 			agentId: 'coder',
+		},
+		{
+			id: CODING_VERIFY_STEP,
+			name: 'Verify & Test',
+			agentId: 'general',
+			instructions:
+				'Review the completed work. Run tests, check for issues. Set result to "passed" if everything looks good, or "failed: <reason>" if problems are found.',
+		},
+		{
+			id: CODING_DONE_STEP,
+			name: 'Done',
+			agentId: 'general',
 		},
 	],
 	transitions: [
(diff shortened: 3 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-implemented change that adds a Verify & Test step to the Coding Workflow template. The implementation is correct and complete.

### What was reviewed

**`packages/daemon/src/lib/space/workflows/built-in-workflows.ts`**
- New `CODING_VERIFY_STEP` and `CODING_DONE_STEP` constants
- Workflow expanded from 2 to 4 steps (Plan, Code, Verify & Test, Done) with correct agent role assignments
- Four transitions with appropriate condition types: `human`, `always`, `task_result` (failed, cyclic), `task_result` (passed)
- `maxIterations: 3` set on the template
- `seedBuiltInWorkflows()` correctly forwards `isCyclic` and `maxIterations` to the workflow manager
- `instructions` field on the Verify step was already being forwarded (line 284)

**`packages/daemon/tests/unit/space/built-in-workflows.test.ts`**
- Tests updated to verify 4 steps, 4 transitions, correct agent IDs, transition conditions, `isCyclic` flag, `maxIterations`, transition ordering, and seeded instructions
- Atomicity test comment correctly updated to reflect that `general` role is now shared across CODING and RESEARCH workflows
- All 44 tests pass

### Verified

1. **Type compatibility**: `SpaceWorkflow.maxIterations` is `number | undefined` -- setting it to `3` is valid. `WorkflowTransition.isCyclic` is `boolean | undefined` -- correct. `WorkflowConditionType` includes `'task_result'` -- correct. `WorkflowStep.instructions` is `string | undefined` -- correct.
2. **Transition ordering**: Verify->Plan has `order: 0`, Verify->Done has `order: 1`. Since `task_result` conditions match on expression value, the order only matters if both could match simultaneously, which they cannot (`failed` vs `passed`). The ordering is reasonable regardless.
3. **Seeding correctness**: Both `isCyclic` and `maxIterations` are properly forwarded through `seedBuiltInWorkflows()` into the `createWorkflow` call. The `instructions` field was already handled.
4. **No regressions**: The atomicity tests still work correctly since `general` is now used by CODING_WORKFLOW too, making the test comment update accurate.

No issues found. The change is minimal, focused, and well-tested.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 1 P0, 1 P1, 1 P2.

---

### Summary

The template structure, transitions, and test coverage are well-crafted and cleanly integrated with the existing storage and seeding infrastructure. However, there is one broken-by-design bug (P0) and one functional gap (P1) that should be addressed before merge.

---

### P0 — `maxIterations: 3` is never applied to workflow runs

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 301–308

`startWorkflowRun()` creates the run record but does not forward `workflow.maxIterations` to `createRun()`:

```ts
const pendingRun = this.config.workflowRunRepo.createRun({
  spaceId,
  workflowId,
  title,
  description,
  currentStepId: workflow.startStepId,
  goalId,
  // workflow.maxIterations is never passed here
});
```

The `CreateWorkflowRunParams` type already has an optional `maxIterations` field whose doc-comment reads "overrides workflow default", and `createRun()` defaults to `5` when the field is absent. So even though `maxIterations: 3` is correctly stored on the `CODING_WORKFLOW` record in the DB, every run created from that workflow silently uses `5` instead of `3`. The test in `built-in-workflows.test.ts` verifies the field is persisted on the workflow record (correct) but does not catch that the run inherits it — which is the only place it is enforced.

The fix is a one-liner in `startWorkflowRun()`: pass `maxIterations: workflow.maxIterations` to `createRun()`.

---

### P1 — Verify & Test step is non-functional at runtime until `task_result` is implemented

**File:** `packages/daemon/src/lib/space/runtime/workflow-executor.ts`, line 238

The `task_result` condition type added in #562 always returns `{ passed: false, reason: 'task_result evaluation not yet implemented' }`. Both outgoing Verify transitions rely on `task_result` conditions, so:

- `Verify → Done` (expression: `'passed'`) never fires.
- `Verify → Plan` (expression: `'failed'`) never fires.

After the Verify step completes, the executor evaluates both transitions, both return `false`, and the run transitions to `needs_attention`. The workflow is effectively broken at the Verify node until Task 1.2 is implemented.

This is pre-existing infrastructure debt from #562 and the current PR clearly states it builds on top of it. However, the PR description does not mention this limitation, and there are no tests documenting the current behaviour (e.g., "Verify step advances to needs_attention when task_result is unimplemented"). A regression-guard test or at minimum a `// TODO: remove once task_result is implemented` comment on the Verify step in the template would make this intentional gap visible to future contributors.

---

### P2 — Instructions say `"failed: <reason>"` but the transition matches on `"failed"` only

**File:** `packages/daemon/src/lib/space/workflows/built-in-workflows.ts`, line 75

The Verify step instructions tell the agent to set `result` to `"failed: <reason>"` (with a colon and reason suffix), but the Verify→Plan transition matches the expression `'failed'` exactly. Whether this works depends on how `task_result` expression matching is implemented in Task 1.2 — exact match would never fire for `"failed: tests timed out"`.

This is not exploitable today (the condition is a stub), but it sets up a likely bug in the next PR. The instructions and the expression should agree on the exact string — either use `"failed"` in both places, or use a prefix/substring match in the evaluator (and document that in the type comment).

---

- The `Done` step correctly has no outgoing transitions, making it terminal — the executor will mark the run `completed` when reached. This is correct.
- The `isCyclic: true` flag on Verify→Plan is stored in the DB but the executor does not yet increment `iterationCount` when following a cyclic transition (also a stub from #562). No issue in this PR — consistent with the phased rollout.
- Tests cover the seeded `isCyclic` value, `maxIterations` on the _workflow_ record, and the Verify step instructions. Test quality is high for what they test.
- The idempotency guard in `seedBuiltInWorkflows()` checks `existing.length > 0`, which means a space that already has a user-created workflow will never receive the built-in templates. This is pre-existing behaviour and not introduced here, but is worth knowing about.

lsm (commented):

**Recommendation: APPROVE**

All three prior review items have been properly addressed. The implementation is clean, well-tested, and coherent. Below are my findings.

---

### Prior Feedback — Resolution Verified

**P0 (maxIterations propagation)** — Fixed correctly. `startWorkflowRun()` in `space-runtime.ts:308` now passes `workflow.maxIterations` to `createRun()`. The repository at `space-workflow-run-repository.ts:48` applies the default of 5 when the value is `undefined`, so the fallback path is handled at the right layer. Two new tests in `space-runtime.test.ts` (lines 345–371) confirm: a workflow with `maxIterations: 3` produces a run with `maxIterations: 3`, and a workflow without it produces a run with `maxIterations: 5`.

**P1 (task_result stub)** — The TODO comment in `workflow-executor.ts:237–239` clearly documents the intentional phased rollout, references the task ID (Task 1.2), and explains the expected behaviour (run goes to `needs_attention`). This is sufficient.

**P2 (prefix-match semantics)** — The NOTE comment at `built-in-workflows.ts:74` clarifies that `task_result` uses prefix matching and gives an example. The wording is clear and placed right where a reader would need it.

---

### Observations (no issues)

1. **Order correctness on Verify transitions** — `verify-to-plan` has `order: 0` and `verify-to-done` has `order: 1`. The executor evaluates transitions in ascending order, so once `task_result` is implemented it will try the "failed" branch first. That ordering is intentional and the test at line 133–141 validates it explicitly.

2. **Done step as a terminal node** — The `Done` step has no outgoing transitions; `advance()` will mark the run `completed` when it is reached. This is correct behaviour and consistent with the executor's documented contract.

3. **`isCyclic` forwarded through seeding** — `seedBuiltInWorkflows` at line 293 now includes `isCyclic` in the mapped transition. The test at lines 359–372 verifies the field survives the round-trip to DB and back.

4. **No dead code or over-engineering** — The step ID constants, the step and transition definitions, and the seeding logic are all used. Nothing is orphaned.

5. **Verify step `agentId` is `general`** — The `Done` step also uses `general`. Both are legitimate; the Done step is purely a terminal marker with no agent-driven work, but using `general` is harmless since the run completes before the Done agent would ever be invoked.

---

No P0, P1, or P2 issues remain. The PR is ready to merge.

lsm (commented):

All three issues from the prior review have been addressed correctly:

**P0 fix (maxIterations propagation):** `startWorkflowRun()` at `space-runtime.ts:308` now forwards `workflow.maxIterations` to `createRun()`. The repository default of `5` applies when the workflow has no `maxIterations` set. Two new tests at `space-runtime.test.ts:345-371` verify both the propagation case and the default fallback. The full chain is sound: template -> `seedBuiltInWorkflows` -> `createWorkflow` -> `startWorkflowRun` -> `createRun`.

**P1 fix (task_result stub):** The TODO comment at `workflow-executor.ts:237-239` clearly documents the intentional phased rollout and references Task 1.2 for the real implementation. The stub returns `{ passed: false }`, which will cause runs to go to `needs_attention` -- a safe and observable failure mode rather than silent incorrect behavior.

**P2 fix (prefix matching note):** The NOTE comment at `built-in-workflows.ts:74` clarifies the prefix matching semantics for `task_result` conditions, making the instruction string `"failed: <reason>"` self-documenting.

### Implementation quality

- **Template structure** (`built-in-workflows.ts`): The 4-step graph (Plan -> Code -> Verify & Test -> Done) with cyclic Verify->Plan transition is well-structured. Transition ordering (Verify->Plan at `order: 0`, Verify->Done at `order: 1`) ensures the failure path is evaluated first, which is the correct semantic.
- **Seeding** (`seedBuiltInWorkflows`): Both `isCyclic` (line 293) and `maxIterations` (line 307) are now forwarded during seeding. The `instructions` field is also forwarded (line 285).
- **Tests** (`built-in-workflows.test.ts`): 44 tests covering template structure, transition conditions, isCyclic flags, maxIterations, instructions, seeding with real DB, idempotency, and error cases. Good coverage.
- **Runtime tests** (`space-runtime.test.ts`): Two new tests for maxIterations propagation are clean and verify the right things.

All 100 tests across both test files pass.

**Recommendation: APPROVE** -- all prior review issues are resolved, no new issues found.

PR #574, opened 2026-03-21
feat: implement task_result evaluation in WorkflowExecutor

lsm:
## Summary

- Adds `taskResult?: string` to `ConditionContext` interface for passing task results to condition evaluation
- Implements the `task_result` case in `evaluateCondition()` with prefix matching logic (`startsWith`)
- Updates `advance()` to accept optional `{ stepResult?: string }` parameter and automatically resolve task results from the DB (most recently completed task on current step), with fallback to `options.stepResult`
- Updates `getConditionContext()` to accept and forward `taskResult`

This is the foundational change for Task 1.3 (wiring `step_result` from the tool handler) — the `advance()` signature change enables callers to pass step results through.

## Test plan

- [x] `evaluateCondition()` with `task_result`: exact match passes
- [x] `evaluateCondition()` with `task_result`: prefix match passes (`'failed'` matches `'failed: tests broken'`)
- [x] `evaluateCondition()` with `task_result`: mismatch fails with descriptive reason
- [x] `evaluateCondition()` with `task_result`: empty expression fails
- [x] `evaluateCondition()` with `task_result`: undefined expression fails
- [x] `evaluateCondition()` with `task_result`: undefined taskResult fails
- [x] `advance()` resolves taskResult from DB completed task
- [x] `advance()` uses stepResult fallback when no DB result exists
- [x] `advance()` DB result takes priority over stepResult fallback
- [x] `advance()` sets needs_attention when task_result doesn't match
- [x] `advance()` sets needs_attention when no task result and no fallback
- [x] `advance()` picks most recently completed task when multiple exist
- [x] Existing condition types (always, human, condition) continue unchanged
- [x] TypeScript compiles cleanly (`bun run typecheck`)
- [x] Lint passes (`bun run lint`)
- [x] All 61 tests pass

### DIFF
--- a/packages/daemon/src/lib/space/runtime/workflow-executor.ts
+++ b/packages/daemon/src/lib/space/runtime/workflow-executor.ts
@@ -44,6 +44,11 @@ export interface ConditionContext {
 	 * Set externally (e.g. via RPC) into run.config.humanApproved before retry.
 	 */
 	humanApproved?: boolean;
+	/**
+	 * Result string from the most recently completed task on the current step.
+	 * Used by `task_result` conditions for prefix matching.
+	 */
+	taskResult?: string;
 }
 
 /** Result of a single condition evaluation attempt */
@@ -234,10 +239,25 @@ export class WorkflowExecutor {
 			}
 
 			case 'task_result': {
-				// TODO(Task 1.2): Implement prefix matching against completed task's result field.
-				// Until then, Verify step transitions (task_result conditions) will not fire,
-				// causing the run to go to needs_attention. This is intentional for phased rollout.
-				return { passed: false, reason: 'task_result evaluation not yet implemented' };
+				if (!condition.expression || !condition.expression.trim()) {
+					return {
+						passed: false,
+						reason: 'task_result type requires a non-empty expression',
+					};
+				}
+				if (context.taskResult === undefined) {
+					return {
+						passed: false,
+						reason: 'No task result available for evaluation',
+					};
+				}
+				if (context.taskResult.startsWith(condition.expression)) {
+					return { passed: true };
+				}
+				return {
+					passed: false,
+					reason: `Task result "${context.taskResult}" does not match "${condition.expression}"`,
+				};
 			}
 
 			default: {
@@ -267,7 +287,9 @@ export class WorkflowExecutor {
 	 *
 	 * Does NOT spawn session groups — that is SpaceRuntime's responsibility.
 	 */
-	async advance(): Promise<{ step: WorkflowStep; tasks: SpaceTask[] }> {
+	async advance(options?: {
+		stepResult?: string;
+	}): Promise<{ step: WorkflowStep; tasks: SpaceTask[] }> {
 		if (this.isComplete()) {
 			throw new Error('Cannot advance: workflow run is already complete');
 		}
@@ -294,11 +316,18 @@ export class WorkflowExecutor {
 			return { step: current, tasks: [] };
 		}
 
+		// Only resolve taskResult when at least one transition uses task_result conditions.
+		// This avoids an unnecessary DB query on the common path (always/human/condition).
+		const needsTaskResult = transitions.some((t) => t.condition?.type === 'task_result');
+		const taskResult = needsTaskResult
+			? await this.resolveTaskResult(current.id, options?.stepResult)
+			: undefined;
+
 		// Evaluate transitions in order; follow the first one whose condition passes.
 		// A failing condition does NOT stop evaluation — the next transition is tried.
 		// Only when every transition has been evaluated and none passed is the run marked
 		// needs_attention and a WorkflowTransitionError thrown.
-		const context = this.getConditionContext();
+		const context = this.getConditionContext(taskResult);
 		let lastReason: string | undefined;
 		let blockedByHumanGate = false;
 
@@ -343,11 +372,12 @@ export class WorkflowExecutor {
 	// -------------------------------------------------------------------------
 
 	/** Builds a ConditionContext from the current run's config and workspacePath. */
-	private getConditionContext(): ConditionContext {
+	private getConditionContext(taskResult?: string): ConditionContext {
 		const config = (this.run.config ?? {}) as Record<string, unknown>;
 		return {
 			workspacePath: this.workspacePath,
 			humanApproved: config.humanApproved === true,
+			taskResult,
 		};
 	}
 
@@ -398,9 +428,12 @@ export class WorkflowExecutor {
 		condition: WorkflowCondition,
 		context: ConditionContext
 	): Promise<ConditionResult> {
-		// human conditions cannot change between retries in the same advance() call;
-		// short-circuit after the first evaluation to avoid redundant checks.
-		const maxAttempts = condition.type === 'human' ? 1 : 1 + (condition.maxRetries ?? 0);
+		// human and task_result conditions cannot change between retries in the same advance()
+		// call; short-circuit after the first evaluation to avoid redundant checks.
+		const maxAttempts =
+			condition.type === 'human' || condition.type === 'task_result'
+				? 1
+				: 1 + (condition.maxRetries ?? 0);
 		let lastResult: ConditionResult = { passed: false, reason: 'Condition never evaluated' };
 
 		for (let attempt = 0; attempt < maxAttempts; attempt++) {
@@ -429,6 +462,27 @@ export class WorkflowExecutor {
 			const updated = this.workflowRunRepo.updateRun(this.run.id, { config: rest });
 			if (updated) this.run = updated;
 		}
+	}
+
+	/**
+	 * Resolves the task result for the current step from the most recently completed task.
+	 * DB task `result` takes priority; `fallback` is used when the DB result is empty.
+	 */
+	private async resolveTaskResult(stepId: string, fallback?: string): Promise<string | undefined> {
+		const allTasks = await this.taskManager.listTasksByWorkflowRun(this.run.id);
+		const completedOnStep = allTasks.filter(
+			(t) => t.workflowStepId === stepId && t.status === 'completed'
+		);
+
+		// Sort by completedAt descending to get the most recently completed task
+		completedOnStep.sort((a, b) => (b.completedAt ?? 0) - (a.completedAt ?? 0));
+
+		const latestTask = completedOnStep[0];
+		const dbResult = latestTask?.result;
+
+		// Use DB result if present (including empty string), otherwise fall back to options.stepResult
+		if (dbResult != null) return dbResult;
+		return fallback;
 	}
 
 	/**
--- a/packages/daemon/src/lib/space/tools/task-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/task-agent-tools.ts
@@ -503,11 +503,11 @@ export function createTaskAgentToolHandlers(config: TaskAgentToolsConfig) {
 		 * 5. Handle WorkflowGateError → return gate-blocked status (caller calls request_human_input)
 		 * 6. Return next step info (or terminal state)
 		 *
-		 * Note: AdvanceWorkflowInput includes a `step_result` field which is currently a
-		 * placeholder. WorkflowExecutor.advance() does not yet accept a result parameter;
-		 * transition conditions are evaluated against run.config (e.g. humanApproved).
-		 * The field is retained in the schema for forward compatibility but is not forwarded
-		 * to the executor in this implementation.
+		 * Note: WorkflowExecutor.advance() now accepts `{ stepResult?: string }` for
+		 * task_result condition evaluation. The `step_result` field in AdvanceWorkflowInput
+		 * is not yet forwarded here — Task 1.3 will wire `_args.step_result` through to
+		 * `executor.advance({ stepResult: _args.step_result })`. In the meantime, advance()
+		 * resolves the task result from the DB (most recently completed task on the step).
 		 */
 		async advance_workflow(_args: AdvanceWorkflowInput): Promise<ToolResult> {
 			const executor = runtime.getExecutor(workflowRunId);
--- a/packages/daemon/tests/unit/space/workflow-executor.test.ts
+++ b/packages/daemon/tests/unit/space/workflow-executor.test.ts
@@ -1306,6 +1306,58 @@ describe('WorkflowExecutor', () => {
 			expect(result.reason).toContain('non-empty expression');
 		});
 
+		test('task_result retries are short-circuited (maxRetries has no effect)', async () => {
+			// task_result context doesn't change between retries, so maxRetries should be ignored.
+			// Verify that advance() only evaluates the condition once even with maxRetries set.
+			const steps = [
+				{ id: STEP_A, name: 'Verify', agentId: AGENT_A },
+				{ id: STEP_B, name: 'Next', agentId: AGENT_B },
+			];
+			const transitions = [
+				{
+					from: STEP_A,
+					to: STEP_B,
+					condition: {
+						type: 'task_result' as const,
+						expression: 'passed',
+						maxRetries: 5,
+					},
+					order: 0,
+				},
+			];
+
+			const workflow = workflowRepo.createWorkflow({
+				spaceId: SPACE_ID,
+				name: `WF-retry-${Date.now()}`,
+				steps,
+				transitions,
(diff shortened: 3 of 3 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** (own PR — cannot formally request changes, but this is my recommendation)

---

### Summary

The core `task_result` condition evaluation logic is correct and well-tested. The implementation faithfully matches the spec (prefix-matching via `startsWith`, fallback chain DB → options.stepResult). However, there are two significant issues: stale documentation at the integration point that directly contradicts the new behavior, and an unconditional DB query on every `advance()` call regardless of whether any transition actually uses `task_result`.

---

### P1 — Stale / misleading comment in the integration caller

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 506–510

The comment block explicitly states:

> `WorkflowExecutor.advance()` does not yet accept a result parameter; transition conditions are evaluated against run.config (e.g. humanApproved). The field is retained in the schema for forward compatibility but is not forwarded to the executor in this implementation.

This is now factually wrong. This PR adds `options?: { stepResult?: string }` to `advance()`. The comment needs updating.

More critically, the tool handler on line 571 still calls `executor.advance()` without forwarding `_args.step_result`:

```ts
const { step: nextStep, tasks: newTasks } = await executor.advance();
//                                                             ^ step_result never forwarded
```

The PR description says this is intentional — Task 1.3 is the wiring step — but the outdated comment is a direct trap for whoever does Task 1.3. They will read lines 506–510, believe `advance()` still lacks the parameter, and potentially either do redundant work or miss that the forwarding is simply needed here. The comment should be updated to say "the executor now accepts this, but the forwarding is deferred to Task 1.3."

---

### P2 — Unconditional DB query on every `advance()` call

**File:** `packages/daemon/src/lib/space/runtime/workflow-executor.ts`, line 321

```ts
const taskResult = await this.resolveTaskResult(current.id, options?.stepResult);
```

`resolveTaskResult` is called unconditionally for every `advance()` invocation that has outgoing transitions, including workflows that use only `always`, `human`, or `condition` transitions. This issues a `listTasksByWorkflowRun` DB query on every advance whether or not any transition will ever inspect `taskResult`.

For the initial implementation this is fine, but it's an unnecessary overhead on the common path. The natural fix is to check whether any outgoing transition uses `task_result` before issuing the query:

```ts
const needsTaskResult = transitions.some((t) => t.condition?.type === 'task_result');
const taskResult = needsTaskResult
  ? await this.resolveTaskResult(current.id, options?.stepResult)
  : undefined;
```

---

### P2 — `maxRetries` silently has no effect on `task_result` conditions

**File:** `packages/daemon/src/lib/space/runtime/workflow-executor.ts`, method `evaluateConditionWithRetry` (line 430)

`human` conditions are explicitly short-circuited to 1 attempt because "context cannot change between retries." The same reasoning applies to `task_result`: the `taskResult` value is resolved once before the transition loop and does not change between retries within a single `advance()` call. A `task_result` condition with `maxRetries: 5` will retry 5 times against the same value, all failing identically.

This is a minor correctness gap — it won't cause incorrect behavior, but it will cause unexpected latency if retries are configured. The short-circuit should include `task_result`:

```ts
const maxAttempts =
  condition.type === 'human' || condition.type === 'task_result'
    ? 1
    : 1 + (condition.maxRetries ?? 0);
```

There is no test verifying this behavior, which leaves the gap invisible.

---

### P3 — Empty-string task result treated as "no result" is inconsistent

**File:** `packages/daemon/src/lib/space/runtime/workflow-executor.ts`, lines 477–478

```ts
if (dbResult) return dbResult;
return fallback;
```

A task explicitly completed with `result = ''` (empty string) will fall through to the fallback rather than being treated as a confirmed empty result. This is unlikely to matter in practice since the schema defines `result?: string | null`, but it diverges from the comment ("DB result takes priority"). If the DB has a completed task with `result: ''`, the fallback value will win over the explicit DB result. A more precise guard would be `if (dbResult != null) return dbResult;` to distinguish null/undefined (absent) from `''` (explicit empty result). This also affects the `setTaskStatus` implementation which already uses `if (options?.result)` to guard assignment — so an empty string result is never stored anyway, making this a latent concern rather than an active bug.

---

### P3 — Minor: test uses `setTimeout` sleep to force `completedAt` ordering

**File:** `packages/daemon/tests/unit/space/workflow-executor.test.ts`, line in the "picks most recently completed task" test

```ts
await new Promise((r) => setTimeout(r, 10));
```

This is a timing-dependent test. If the underlying clock granularity is coarser than 10 ms (or the test system is heavily loaded), the two `completedAt` timestamps could be identical and the sort would be unstable, potentially returning the earlier task. This is inherently fragile. A more reliable approach would be to mock or inject the timestamp source so the test can assert ordering deterministically without relying on wall-clock time.

---

- The `task_result` case in `evaluateCondition()` is clean and correct: it validates the expression before inspecting `taskResult`, preventing false positives from an empty expression matching any string via `''.startsWith('')`.
- The fallback chain (DB → options.stepResult) is logical and the priority semantics are clear.
- The `ConditionContext` extension is backward-compatible (optional field, no existing callers broken).
- The `advance()` signature change is backward-compatible (optional parameter).
- Test coverage for `evaluateCondition()` is thorough (all 6 branches covered).
- The integration tests in `setupTaskResultWorkflow` correctly use real DB-backed state, making them meaningful integration tests rather than pure mocks.
- TypeScript exhaustiveness check (`default: never`) ensures the new case is handled in the switch.

lsm (commented):

### Recommendation: APPROVE

The implementation is clean, correct, and well-tested. The `task_result` condition evaluation, `ConditionContext.taskResult` field, `resolveTaskResult()` DB query, and `advance()` options parameter all work together correctly. The 14 new tests (7 for `evaluateCondition` edge cases, 7 for `advance()` integration with DB resolution and fallback) provide thorough coverage.

### Detailed findings

**No P0/P1/P2 issues found.** The core implementation is solid:

- **Correctness**: `startsWith` prefix matching aligns with the shared type documentation ("starts with or equals"). Empty/undefined expression and taskResult are properly guarded. The exhaustive switch ensures type safety.
- **DB resolution**: `resolveTaskResult()` correctly queries all tasks by workflow run, filters to completed tasks on the current step, sorts by `completedAt` descending, and falls back to the `stepResult` option. The `completedAt ?? 0` fallback in sort handles the theoretical `undefined` case safely (though completed tasks should always have `completedAt` set).
- **Edge cases**: Empty string `taskResult` is handled correctly (passes the `=== undefined` check but won't match any non-empty expression). Null `result` from DB is correctly treated as falsy and falls through to fallback.
- **Tests**: Good coverage of prefix matching, exact matching, no-match, empty expression, undefined expression, undefined taskResult, DB resolution, DB priority over fallback, multiple tasks (most recent wins), and needs_attention on mismatch.

**P3 observations (non-blocking):**

1. **Stale comment in `task-agent-tools.ts:506-510`** (outside this PR's commit scope): The comment says "WorkflowExecutor.advance() does not yet accept a result parameter" — now outdated. The `advance_workflow` tool handler at line 571 also doesn't forward `_args.step_result` to `executor.advance({ stepResult: _args.step_result })`. The DB-first resolution path makes this work in practice (the task is already completed in DB before `advance_workflow` is called), but the fallback path is unreachable from production callers. This could be addressed in a follow-up.

2. **Performance**: `resolveTaskResult()` calls `listTasksByWorkflowRun()` which fetches ALL tasks for the run, then filters in-memory. For workflows with many steps and tasks, a targeted DB query (filter by stepId + status in SQL) would be more efficient. Not a concern at current scale.

lsm (commented):

### Recommendation: APPROVE

All five issues from the previous review have been properly addressed:

1. **P1 stale comment (fixed):** `task-agent-tools.ts` lines 506-510 now accurately describe that `advance()` accepts `{ stepResult?: string }` and that Task 1.3 will wire the forwarding. The comment is precise about what exists and what remains.

2. **P2 unconditional DB query (fixed):** `resolveTaskResult()` is now gated behind `transitions.some((t) => t.condition?.type === 'task_result')`, avoiding the DB query on the common path for `always`/`human`/`condition` transitions.

3. **P2 task_result retry short-circuit (fixed):** `evaluateConditionWithRetry()` now treats `task_result` the same as `human` — capped at 1 attempt. A dedicated test (`task_result retries are short-circuited`) validates this with `maxRetries: 5` and confirms the condition is evaluated exactly once.

4. **P3 empty-string DB result (fixed):** Changed from `if (dbResult)` to `if (dbResult != null)` in `resolveTaskResult()`, correctly distinguishing an explicit empty-string result from null/undefined.

5. **P3 deterministic test timestamps (fixed):** The "picks most recently completed task" test now uses direct DB updates (`UPDATE space_tasks SET completed_at = ? WHERE id = ?`) with values 1000 and 2000, eliminating the `setTimeout` timing dependency entirely.

### New code quality assessment

- **Correctness:** The `task_result` evaluation logic is sound — validates expression non-emptiness, checks for `taskResult` presence, and uses `startsWith` for prefix matching. The exhaustive switch default ensures compile-time safety.
- **Tests:** 14 new tests cover all meaningful paths: exact match, prefix match, mismatch, empty/undefined expression, undefined taskResult, DB resolution, DB-over-fallback priority, multiple tasks ordering, retry short-circuit, needs_attention on failure, and regression for existing condition types.
- **Integration:** The `advance_workflow` tool handler correctly calls `executor.advance()` without forwarding `step_result` (intentionally deferred to Task 1.3), and the DB-first resolution path in `resolveTaskResult()` ensures task_result conditions work in production even without the forwarding.

No new issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### Follow-up review: All previous issues are properly resolved

This is a follow-up to review [#3986230754](https://github.com/lsm/neokai/pull/574#pullrequestreview-3986230754). I verified each of the five claimed fixes against the current code.

---

#### P1 — Stale comment in `task-agent-tools.ts` (FIXED)

Lines 506–511 of `packages/daemon/src/lib/space/tools/task-agent-tools.ts` now correctly describe the current state:

> "Note: WorkflowExecutor.advance() now accepts `{ stepResult?: string }` for task_result condition evaluation. The `step_result` field in AdvanceWorkflowInput is not yet forwarded here — Task 1.3 will wire `_args.step_result` through to `executor.advance({ stepResult: _args.step_result })`. In the meantime, advance() resolves the task result from the DB (most recently completed task on the step)."

This accurately reflects what is implemented and what is deferred, leaving no trap for the Task 1.3 implementer.

---

#### P2 — DB query now conditional on `task_result` transitions (FIXED)

`packages/daemon/src/lib/space/runtime/workflow-executor.ts`, lines 320–324:

```ts
const needsTaskResult = transitions.some((t) => t.condition?.type === 'task_result');
const taskResult = needsTaskResult
    ? await this.resolveTaskResult(current.id, options?.stepResult)
    : undefined;
```

The `listTasksByWorkflowRun` DB query is no longer issued for workflows that use only `always`, `human`, or `condition` transitions.

---

#### P2 — `task_result` retries short-circuited (FIXED)

`packages/daemon/src/lib/space/runtime/workflow-executor.ts`, lines 432–436:

```ts
const maxAttempts =
    condition.type === 'human' || condition.type === 'task_result'
        ? 1
        : 1 + (condition.maxRetries ?? 0);
```

The `task_result` case is now short-circuited alongside `human`. A dedicated test at line 1147 ("task_result retries are short-circuited") verifies this by completing a task with `'failed'`, setting `maxRetries: 5`, and asserting that `advance()` throws immediately rather than retrying.

---

#### P3 — Empty-string DB result handled with `!= null` (FIXED)

`packages/daemon/src/lib/space/runtime/workflow-executor.ts`, line 484:

```ts
if (dbResult != null) return dbResult;
```

An explicit `result: ''` on a completed task is now treated as a present DB result rather than falling through to the fallback.

---

#### P3 — Deterministic `completedAt` via direct DB update (FIXED)

The "picks most recently completed task" test (line 1393) now sets `completed_at` with deterministic integer values (`1000` and `2000`) via direct SQL updates after completing each task, eliminating the previous wall-clock timing dependency.

---

### No new issues found

All five previous issues are resolved cleanly. The implementation is correct, the test coverage is thorough, and no regressions were introduced. The one intentional deferral (forwarding `step_result` from `advance_workflow` to `executor.advance()` — Task 1.3) is now clearly documented in the code.

lsm (commented):

### Post-rebase verification — APPROVE

Verified the following after rebase onto dev:

1. **Conflict resolution is correct** — Both `goalId propagation` tests (from dev) and `task_result` tests (from the branch) are present in `workflow-executor.test.ts`. The file has 18 describe blocks covering all expected areas.

2. **Previous review fixes are intact:**
   - Retry short-circuit test for `task_result` conditions (maxRetries=5 is ignored, fails immediately)
   - Deterministic timestamps via direct DB update instead of `setTimeout(10)` delay for multi-task ordering test

3. **Core implementation unchanged:**
   - `evaluateCondition()` handles `task_result` with empty expression, undefined taskResult, prefix matching, and mismatch cases
   - `advance()` accepts optional `stepResult` parameter, lazily resolves `taskResult` only when transitions use `task_result` conditions
   - `resolveTaskResult()` prioritizes DB result over fallback, sorts by `completedAt` descending
   - Retry short-circuit correctly includes `task_result` alongside `human`

No regressions detected. The PR is in good shape for merge.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** (own PR — cannot formally approve, but this is my final verification recommendation)

---

### Post-rebase final verification

This is a final check after merge conflict resolution and rebase onto dev. I independently verified the three requested areas.

---

#### 1. Previously fixed P1/P2/P3 issues are still properly addressed

All five previously fixed issues are intact in the current code:

**P1 — Stale comment in `task-agent-tools.ts` (lines 506–510):**
The comment accurately describes that `advance()` now accepts `{ stepResult?: string }`, that forwarding of `_args.step_result` is deferred to Task 1.3, and that the DB-based resolution is the current mechanism. No regression.

**P2 — Lazy DB query (line 321 of `workflow-executor.ts`):**
`needsTaskResult` gate correctly skips `resolveTaskResult()` when no transition uses `task_result` conditions, preserving performance on the common path.

**P2 — Retry short-circuit for `task_result` (lines 433–436):**
`evaluateConditionWithRetry()` short-circuits to `maxAttempts = 1` for both `human` and `task_result` types. The dedicated regression test at line 1309 confirms this behavior.

**P3 — `!= null` check instead of truthiness (line 484):**
`resolveTaskResult()` uses `if (dbResult != null)` which correctly passes empty strings from the DB while falling through to the `fallback` for `null`/`undefined`. This was the correct fix — a truthiness check would incorrectly ignore empty-string results.

**P3 — Deterministic timestamps for multi-task ordering (lines 1600, 1613):**
`'UPDATE space_tasks SET completed_at = ?'` direct DB writes replace any timing-based approach. The test is now deterministic and race-condition-free.

---

#### 2. Conflict resolution kept both goalId tests and task_result tests intact

The test file has 18 `describe` blocks at the top level. The `goalId propagation` section (starting at line 1067) with 4 tests is fully present and unmodified from its origin in dev. The two new `task_result` sections (lines 1229 and 1380) are each present exactly once — no duplicate `describe` blocks. The PR diff's apparent duplication was an artifact of how the conflict resolution was displayed; the actual file contains no duplicate content.

Counts verified:
- 1 definition + 4 call sites of `setupTaskResultWorkflow` (no duplication)
- `describe('condition type: task_result'` appears exactly once (line 1229)
- `describe('advance() with task_result conditions'` appears exactly once (line 1380)

---

#### 3. No regressions introduced

The core implementation in `workflow-executor.ts` is clean:
- The `ConditionContext.taskResult` field is correctly optional
- `evaluateCondition()` handles all edge cases: empty expression, undefined taskResult, prefix match, and mismatch
- The `advance()` signature change (`options?: { stepResult?: string }`) is backward-compatible
- `getConditionContext()` correctly threads `taskResult` through
- The `default` branch exhaustiveness check (`_exhaustive: never`) remains intact

The only minor observation (not a blocker): there is no test for a whitespace-only `expression` (e.g. `expression: '   '`) for `task_result` conditions. The `condition` type has the same gap. The `!condition.expression.trim()` guard handles it correctly, but it is untested. This is a P3 nit that does not block merge given the existing coverage depth.

---

**Summary:** The implementation is correct, the conflict resolution is clean, all previously fixed issues are intact, and no regressions are present. The PR is ready to merge.

PR #577, opened 2026-03-21
feat: rewrite RoomContextPanel with goal-centric sidebar layout

lsm:
## Summary
- Rewrites `RoomContextPanel.tsx` with the new goal-centric sidebar layout
- Removes back button and standalone New Session button (nav rail handles navigation)
- Adds pinned Dashboard and Room Agent items at top with visual divider
- Adds collapsible Goals section showing goals with expandable linked tasks
- Adds Tasks section with Active/Review/Done tab filter for orphan tasks
- Adds Sessions section (collapsed by default) with [+] create button in header
- Fixes `isDashboardSelected` to check both `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal` are null
- All navigation actions call `onNavigate?.()` to close mobile drawer

## Test plan
- [x] 28 unit tests covering all sections, selection highlighting, navigation, and tab filtering
- [x] Lint passes (oxlint)
- [x] Format passes (biome)
- [x] Knip passes (no dead exports)

### DIFF
--- a/packages/web/src/islands/RoomContextPanel.tsx
+++ b/packages/web/src/islands/RoomContextPanel.tsx
@@ -19,6 +19,7 @@ import {
 	navigateToRoomTask,
 } from '../lib/router';
 import { currentRoomSessionIdSignal, currentRoomTaskIdSignal } from '../lib/signals';
+import { toast } from '../lib/toast';
 import { cn } from '../lib/utils';
 
 function formatRelativeTime(timestamp: number): string {
@@ -76,7 +77,6 @@ interface RoomContextPanelProps {
 export function RoomContextPanel({ roomId, onNavigate }: RoomContextPanelProps) {
 	const sessions = roomStore.sessions.value;
 	const tasks = roomStore.tasks.value;
-	const goals = roomStore.goals.value;
 	const [showArchived, setShowArchived] = useState(false);
 	const [expandedGoals, setExpandedGoals] = useState<Set<string>>(() => new Set());
 	const [orphanTab, setOrphanTab] = useState<OrphanTab>('active');
@@ -88,7 +88,8 @@ export function RoomContextPanel({ roomId, onNavigate }: RoomContextPanelProps) 
 		[tasks]
 	);
 
-	// Goals data
+	// Goals data — show only active goals; count matches the rendered list
+	const activeGoals = roomStore.activeGoals.value;
 	const tasksByGoalId = roomStore.tasksByGoalId.value;
 
 	// Orphan tasks by tab — read signal .value directly (no useMemo) so Preact
@@ -162,8 +163,7 @@ export function RoomContextPanel({ roomId, onNavigate }: RoomContextPanelProps) 
 				navigateToRoomSession(roomId, sessionId);
 				onNavigate?.();
 			} catch {
-				// createSession can fail if not connected or RPC errors — silently ignore
-				// since the user will see the session list unchanged as feedback.
+				toast.error('Failed to create session');
 			}
 		},
 		[roomId, onNavigate]
@@ -243,11 +243,11 @@ export function RoomContextPanel({ roomId, onNavigate }: RoomContextPanelProps) 
 			{/* Scrollable sections */}
 			<div class="flex-1 overflow-y-auto">
 				{/* Goals section */}
-				<CollapsibleSection title="Goals" count={goals.length}>
-					{goals.length === 0 ? (
+				<CollapsibleSection title="Goals" count={activeGoals.length}>
+					{activeGoals.length === 0 ? (
 						<div class="px-4 py-3 text-xs text-gray-600">No goals</div>
 					) : (
-						goals.map((goal) => {
+						activeGoals.map((goal) => {
 							const isExpanded = expandedGoals.has(goal.id);
 							const linkedTasks = tasksByGoalId.get(goal.id) ?? [];
 							return (
--- a/packages/web/src/islands/RoomContextPanel.tsx
+++ b/packages/web/src/islands/RoomContextPanel.tsx
@@ -1,23 +1,24 @@
 /**
  * RoomContextPanel
  *
- * Context panel content shown when viewing a specific room.
- * Replaces the generic RoomList with room-specific information:
- * - Back navigation to rooms list
- * - Task stats strip
- * - Room Dashboard pinned at top
- * - Sessions list below (with filter toggle for archived sessions)
+ * Goal-centric sidebar for a specific room. Layout (top to bottom):
+ * 1. Task stats strip (pending · active counts)
+ * 2. Pinned items: Dashboard, Room Agent
+ * 3. Goals section (collapsible) with expandable linked tasks
+ * 4. Tasks section (orphan tasks with tab filter)
+ * 5. Sessions section (collapsible, default collapsed)
  */
 
 import { useMemo, useState } from 'preact/hooks';
+import { CollapsibleSection } from '../components/room/CollapsibleSection';
 import { roomStore } from '../lib/room-store';
 import {
-	navigateToRooms,
 	navigateToRoom,
 	navigateToRoomAgent,
 	navigateToRoomSession,
+	navigateToRoomTask,
 } from '../lib/router';
-import { currentRoomSessionIdSignal } from '../lib/signals';
+import { currentRoomSessionIdSignal, currentRoomTaskIdSignal } from '../lib/signals';
 import { cn } from '../lib/utils';
 
 function formatRelativeTime(timestamp: number): string {
@@ -28,7 +29,7 @@ function formatRelativeTime(timestamp: number): string {
 	return `${Math.floor(seconds / 86400)}d`;
 }
 
-function StatusDot({ status }: { status: string }) {
+function SessionStatusDot({ status }: { status: string }) {
 	const colors: Record<string, string> = {
 		idle: 'bg-gray-500',
 		active: 'bg-green-500',
@@ -40,6 +41,33 @@ function StatusDot({ status }: { status: string }) {
 	return <div class={cn('w-2 h-2 rounded-full flex-shrink-0', colors[status] ?? colors.idle)} />;
 }
 
+const taskStatusColors: Record<string, string> = {
+	draft: 'bg-gray-500',
+	pending: 'bg-yellow-500',
+	in_progress: 'bg-blue-500',
+	review: 'bg-purple-500',
+	needs_attention: 'bg-orange-500',
+	completed: 'bg-green-500',
+	cancelled: 'bg-gray-600',
+};
+
+function TaskStatusDot({ status }: { status: string }) {
+	return (
+		<div
+			class={cn('w-2 h-2 rounded-full flex-shrink-0', taskStatusColors[status] ?? 'bg-gray-500')}
+		/>
+	);
+}
+
+const goalStatusColors: Record<string, string> = {
+	active: 'text-green-400',
+	needs_human: 'text-yellow-400',
+	completed: 'text-gray-500',
+	archived: 'text-gray-600',
+};
+
+type OrphanTab = 'active' | 'review' | 'done';
+
 interface RoomContextPanelProps {
 	roomId: string;
 	onNavigate?: () => void;
@@ -48,16 +76,30 @@ interface RoomContextPanelProps {
 export function RoomContextPanel({ roomId, onNavigate }: RoomContextPanelProps) {
 	const sessions = roomStore.sessions.value;
 	const tasks = roomStore.tasks.value;
+	const goals = roomStore.goals.value;
 	const [showArchived, setShowArchived] = useState(false);
+	const [expandedGoals, setExpandedGoals] = useState<Set<string>>(() => new Set());
+	const [orphanTab, setOrphanTab] = useState<OrphanTab>('active');
 
+	// Task stats
 	const pendingCount = useMemo(() => tasks.filter((t) => t.status === 'pending').length, [tasks]);
 	const activeCount = useMemo(
 		() => tasks.filter((t) => t.status === 'in_progress').length,
 		[tasks]
 	);
-	const doneCount = useMemo(() => tasks.filter((t) => t.status === 'completed').length, [tasks]);
 
-	// Filter sessions based on showArchived toggle
+	// Goals data
+	const activeGoals = roomStore.activeGoals.value;
+	const tasksByGoalId = roomStore.tasksByGoalId.value;
+
+	// Orphan tasks by tab
+	const orphanTasksForTab = useMemo(() => {
+		if (orphanTab === 'active') return roomStore.orphanTasksActive.value;
+		if (orphanTab === 'review') return roomStore.orphanTasksReview.value;
+		return roomStore.orphanTasksDone.value;
+	}, [orphanTab]);
+
+	// Sessions
 	const filteredSessions = useMemo(() => {
 		if (showArchived) return sessions;
 		return sessions.filter((s) => s.status !== 'archived');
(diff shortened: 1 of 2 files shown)


lsm (commented):

Overall the rewrite is well-structured and achieves the goal-centric sidebar layout. The component is clean, tests are comprehensive, and the `isDashboardSelected` fix is correct. However, there is one reactivity bug and one missing error handler that should be addressed.

---

### P0: `orphanTasksForTab` useMemo has stale closure over signal values (reactivity bug)

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, lines 96-100

```tsx
const orphanTasksForTab = useMemo(() => {
    if (orphanTab === 'active') return roomStore.orphanTasksActive.value;
    if (orphanTab === 'review') return roomStore.orphanTasksReview.value;
    return roomStore.orphanTasksDone.value;
}, [orphanTab]);
```

The `useMemo` dependency array is `[orphanTab]` only. When the underlying tasks data changes (tasks added, status changes), the computed signals (`orphanTasksActive`, etc.) update, but this `useMemo` returns the cached stale result because `orphanTab` hasn't changed.

While the component does subscribe to `roomStore.tasks.value` at line 78 (which triggers re-renders), the `useMemo` still returns its cached value on re-render since its dependency array hasn't changed.

**Fix:** Either remove the `useMemo` entirely (these are already computed signals, so the overhead of re-reading `.value` on each render is negligible), or add the relevant signal values to the dependency array:

```tsx
// Option A: just read directly (preferred — computed signals are already cached)
const orphanTasksForTab =
    orphanTab === 'active' ? roomStore.orphanTasksActive.value
    : orphanTab === 'review' ? roomStore.orphanTasksReview.value
    : roomStore.orphanTasksDone.value;
```

---

### P1: No error handling on async `createSession` in [+] button handler

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, lines 320-325

```tsx
onClick={async (e) => {
    e.stopPropagation();
    const sessionId = await roomStore.createSession();
    navigateToRoomSession(roomId, sessionId);
    onNavigate?.();
}}
```

`roomStore.createSession()` can throw (no room, no hub connection, RPC failure). An unhandled rejection in an event handler will surface as an uncaught promise rejection in the console. Wrap in try/catch or add `.catch()`.

---

### P2: Test for [+] create session button doesn't verify `onNavigate` is called

**File:** `packages/web/src/islands/__tests__/RoomContextPanel.test.tsx`, lines 403-414

The test "creates a session via the [+] button and navigates to it" doesn't pass `onNavigate` and doesn't verify it's called after session creation. Every other navigation test verifies `onNavigate` — this one should too for consistency and to guard against regressions.

---

### P3: `@ts-nocheck` in test file

**File:** `packages/web/src/islands/__tests__/RoomContextPanel.test.tsx`, line 1

The `@ts-nocheck` was inherited from the previous version of this test file. It would be better to remove it and fix any type issues, but this is inherited and not introduced by this PR.

---

### P3: `formatRelativeTime` is unused in the non-session context

The function is still used for session `lastActiveAt` display, so this is fine. Just noting it's a holdover from the previous implementation but still has a valid use.

---

### What looks good

- The `isDashboardSelected` fix correctly checks both `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal` are null (line 118)
- All navigation handlers properly call `onNavigate?.()` for mobile drawer close
- Back button and standalone New Session button are correctly removed
- CollapsibleSection integration is clean with proper `defaultExpanded={false}` for Sessions
- Goal expand/collapse with `Set<string>` state management is correct
- Test coverage is thorough at 28 tests covering all key behaviors
- Signal mocking approach in tests is solid with `initSignals()` reset in `beforeEach`

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two correctness bugs need addressing before merge.

---

## Summary

The rewrite delivers the requested goal-centric layout cleanly. The component structure is well-organized, the `isDashboardSelected` fix is correct, mobile `onNavigate` wiring is complete, and the 28 tests broadly cover the new behaviour. However there are two bugs worth fixing, one of which will cause stale UI in production.

---

## P1 — Bugs

### 1. `useMemo` breaks signal reactivity for orphan task tabs

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, lines 96–100

```ts
const orphanTasksForTab = useMemo(() => {
  if (orphanTab === 'active') return roomStore.orphanTasksActive.value;
  if (orphanTab === 'review') return roomStore.orphanTasksReview.value;
  return roomStore.orphanTasksDone.value;
}, [orphanTab]);
```

Preact Signals' auto-tracking only works when `.value` is read during a tracked render or `computed()` call. Inside `useMemo`, the signal reads are not tracked by Preact Signals, and the dependency array only lists `[orphanTab]`. This means when task data changes (e.g. a task transitions from `in_progress` → `review` while the Active tab is open), `orphanTasksForTab` will **not update** — the user sees stale data.

Every other signal value in this component is read directly in the render body (lines 77–93), which participates in auto-tracking correctly. Fix: either read the signals directly in the render body and branch in JSX, or add the signal array values as memo dependencies:

```ts
const orphanTasksForTab =
  orphanTab === 'active' ? roomStore.orphanTasksActive.value
  : orphanTab === 'review' ? roomStore.orphanTasksReview.value
  : roomStore.orphanTasksDone.value;
```

### 2. No error handling on async session create

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, lines 320–325

```ts
onClick={async (e) => {
  e.stopPropagation();
  const sessionId = await roomStore.createSession();
  navigateToRoomSession(roomId, sessionId);
  onNavigate?.();
}}
```

`roomStore.createSession()` throws when not connected, no room is selected, or the server returns an error (see `room-store.ts` lines 624–643). An unhandled promise rejection from an event handler is silently swallowed in browsers. The user gets no feedback and the button appears to do nothing. A try/catch with a `toast.error` call (as used throughout the rest of `room-store.ts`) is needed here.

---

## P1 — Test Quality

### 3. `// @ts-nocheck` at the top of the test file

**File:** `packages/web/src/islands/__tests__/RoomContextPanel.test.tsx`, line 1

The entire test file opts out of TypeScript checking. This means:
- Mock objects (`makeTask`, `makeSession`, `makeGoal`) can drift from the real `TaskSummary`, `SessionSummary`, `RoomGoal` types with no compiler warning.
- Any future type changes to `RoomContextPanelProps` or store shape won't surface as test failures.

Prefer resolving the specific type error that motivated this suppression, or at minimum restrict it to the one offending line.

### 4. `[+] button create test does not verify `onNavigate` is called`

**File:** `packages/web/src/islands/__tests__/RoomContextPanel.test.tsx`, lines 403–414

The test `creates a session via the [+] button and navigates to it` doesn't pass an `onNavigate` prop, so it cannot verify that the mobile drawer close callback is invoked after session creation. This is the one navigation path not covered by the `onNavigate` call tests.

---

## P2 — Design Issues

### 5. Goals section: count vs list mismatch

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, lines 229–273

```tsx
<CollapsibleSection title="Goals" count={activeGoals.length}>
  {goals.length === 0 ? (
    <div>No goals</div>
  ) : (
    goals.map(...)  // iterates ALL goals, not just active ones
  )}
```

The header badge shows `(N)` for active goals only, but the list renders all goals including archived/completed ones. A user seeing `Goals (1)` but two items in the list (one active, one archived) will be confused. Either filter the list to `activeGoals` or change the count to `goals.length`.

### 6. `isRoomAgentSelected` is inconsistent with the `isDashboardSelected` fix

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, line 119

```ts
const isRoomAgentSelected = selectedSessionId === roomAgentSessionId;
```

The router correctly sets `currentRoomTaskIdSignal` to `null` when navigating to the agent, so this is not a live bug. But for symmetry with the explicit two-signal check for Dashboard, it is worth noting that if that router invariant ever breaks, the Room Agent item could be highlighted incorrectly (session matches but a task is also selected). Low-risk but worth a comment at minimum.

---

## P3 — Nits

### 7. Unused destructuring in test

**File:** `packages/web/src/islands/__tests__/RoomContextPanel.test.tsx`, line 221

```ts
const { container } = render(<RoomContextPanel roomId="room-1" />);
```

`container` is never referenced in that test. Remove or replace with just `render(...)`.

### 8. Inline async handler in JSX prop

**File:** `packages/web/src/islands/RoomContextPanel.tsx`, lines 320–325

The async session-create handler is defined inline inside the `headerRight` JSX prop. It would be easier to read, test, and wrap in error handling if extracted to a named handler like the other navigation handlers (lines 135–153).

lsm (commented):

**Recommendation: APPROVE**

All issues from the previous review have been addressed:

1. **P0 useMemo stale closure (FIXED):** `orphanTasksForTab` now reads signal `.value` directly without `useMemo`, correctly participating in Preact Signals auto-tracking (lines 96-101).

2. **P1 missing error handling on createSession (FIXED):** The handler is now wrapped in try/catch (lines 160-167) and extracted to a named `handleCreateSession` via `useCallback` (lines 157-170).

3. **P2 Goals count mismatch (FIXED):** Count now uses `goals.length` (line 246) matching the full rendered list.

4. **P2 comment on isRoomAgentSelected (FIXED):** Explanatory comment added at line 120.

5. **P2 test: onNavigate verification on create session (FIXED):** Test at line 399 now passes `onNavigate` and verifies it is called after session creation (line 411).

6. **P3 unused container destructuring in test (FIXED):** Removed.

---

### Remaining P3 nits (not blocking)

**`@ts-nocheck` in test file** (line 1 of `RoomContextPanel.test.tsx`): Still present but was already noted as inherited/pre-existing in the previous review. Not introduced by this PR.

**Silent error swallowing in `handleCreateSession`** (line 164-166): The catch block silently ignores failures. While the comment explains the rationale ("user will see the session list unchanged"), a brief toast notification would provide better UX for failures. Low priority since this is a sidebar convenience button.

---

All 28 tests pass. The component correctly implements the goal-centric sidebar layout with proper signal reactivity, navigation handling, and error boundaries.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (1 P1, 1 P2)

All six items from the previous round were addressed. The signal stale-closure fix is correct, the `handleCreateSession` extraction and `useCallback` are clean, the Goals count now uses `goals.length`, and the test correctly verifies `onNavigate` is called after session creation. What remains are two issues that warrant attention.

---

### P1 — Silent error swallow on `createSession` gives the user zero feedback

`packages/web/src/islands/RoomContextPanel.tsx` lines 157–170:

```ts
const handleCreateSession = useCallback(
  async (e: Event) => {
    e.stopPropagation();
    try {
      const sessionId = await roomStore.createSession();
      navigateToRoomSession(roomId, sessionId);
      onNavigate?.();
    } catch {
      // createSession can fail if not connected or RPC errors — silently ignore
      // since the user will see the session list unchanged as feedback.
    }
  },
  [roomId, onNavigate]
);
```

The comment rationalises swallowing the error, but the rationale is shaky. When the user taps `[+]` and nothing happens, the only signal is the session list not growing — there is no visual feedback that an action was attempted. In a mobile drawer context (the primary consumer of `onNavigate`) the drawer may still be closing while the failure is silent. Existing patterns in the codebase (`roomStore` itself, several RPC handlers) surface errors via `toast()`. A one-line `toast('Failed to create session')` in the catch block is the established pattern and costs almost nothing. If a deliberate policy decision was made to stay silent here, it should reference that policy (e.g., a link to an issue or decision log), not just assert the user will notice.

Additionally, there is no test covering the failure path — a test that mocks `createSession` to reject and asserts no navigation occurs would prevent future regressions.

---

### P2 — Goals count badge includes archived goals without indication

`packages/web/src/islands/RoomContextPanel.tsx` line 246:

```tsx
<CollapsibleSection title="Goals" count={goals.length}>
```

After the fix, `count={goals.length}` counts every goal — including archived/completed ones. The old code used `activeGoals.length`, which was arguably wrong in the other direction. Neither is labelled. If a room has 1 active goal and 5 archived ones the badge shows "(6)". Users typically interpret a sidebar badge as actionable/active items. Either:
- label the section explicitly (`title="Goals (all)"`), or
- use a count that distinguishes active from total (e.g., `count={activeGoals.length}` with a tooltip or secondary label for total), or
- document the intent with a comment so future maintainers understand the choice.

This is cosmetic but can confuse users and is worth a deliberate decision before the PR lands.

---

### Minor notes (no change required)

- **`useMemo` for `pendingCount`/`activeCount`**: The pattern (`tasks` extracted from signal then passed as dep) is fine — the signal subscription at line 78 ensures re-renders when the task list changes, so the `useMemo` deps are stable references that work correctly.
- **`filteredSessions` / `hasArchivedSessions`**: Same pattern, same reasoning — no issue.
- **`isRoomAgentSelected` comment**: Good addition, clear.
- **Test cleanup**: Removal of the unused `container` destructure is correct.

lsm (commented):

**Recommendation: APPROVE**

All three issues from the previous review round have been properly addressed in commit `a3672581`:

1. **P1 silent error swallow on createSession (FIXED):** `handleCreateSession` catch block now calls `toast.error('Failed to create session')` (line 166 of `RoomContextPanel.tsx`), matching the established error-surfacing pattern used elsewhere in the codebase.

2. **P2 Goals count badge included archived goals (FIXED):** The component now reads `roomStore.activeGoals.value` instead of `roomStore.goals.value`. Both the count badge (`count={activeGoals.length}`, line 246) and the rendered list (`activeGoals.map(...)`, line 250) use the same filtered signal, so the count is always consistent with what the user sees.

3. **Test coverage for both fixes:**
   - New test "shows toast and does not navigate when createSession fails" (line 433) mocks `createSession` to reject and verifies `toast.error` is called with the correct message and that no navigation occurs.
   - New test "does not render archived goals in the list" (line 268) verifies archived goals are excluded from rendering while active goals are shown.

All 30 tests pass. The mock setup correctly wires `mockToast` via `vi.mock('../../lib/toast.ts')` and the `mockActiveGoals` computed signal properly filters by `status === 'active'`, matching the real `roomStore.activeGoals` implementation.

No remaining issues.

lsm (commented):

**Recommendation: APPROVE** — All three items from iteration 2 are correctly addressed. One new finding surfaced during this pass that is worth tracking but is pre-existing and not introduced by this PR.

---

### Fixes verified from iteration 2

**1. `toast.error` on `createSession` failure (FIXED correctly)**
`packages/web/src/islands/RoomContextPanel.tsx` lines 158–170: the `handleCreateSession` `useCallback` wraps the async operation in try/catch and calls `toast.error('Failed to create session')` in the catch branch. This matches the established error-surfacing pattern used elsewhere.

**2. `activeGoals` used for both count and list (FIXED correctly)**
`packages/web/src/islands/RoomContextPanel.tsx` line 92 reads `roomStore.activeGoals.value`; line 246 passes `count={activeGoals.length}` and line 250 iterates `activeGoals.map(...)`. The count badge and the rendered list are now driven by the same filtered signal — they cannot diverge.

**3. Tests for both new behaviors (FIXED correctly)**
- `packages/web/src/islands/__tests__/RoomContextPanel.test.tsx` lines 433–444: mocks `mockCreateSession.mockRejectedValue(...)`, fires the `[+]` button, and asserts `mockToast` is called with the correct message and `mockNavigateToRoomSession` is never called. The failure path is fully covered.
- Lines 268–276: places one active and one archived goal, renders, asserts the active goal is visible and `queryByText('Archived Goal')` is null. The "no archived goals rendered" invariant is tested.

---

### New finding: `needs_human` goals silently hidden from sidebar (pre-existing, not introduced here)

`packages/shared/src/types/neo.ts` line 68 defines:
```ts
export type GoalStatus = 'active' | 'needs_human' | 'completed' | 'archived';
```

`needs_human` is a non-terminal, actionable status — it means the autonomous agent needs human intervention before proceeding. `GoalsEditor.tsx` treats it as the second-most-prominent status (sorted `active > needs_human > completed > archived`).

`packages/web/src/lib/room-store.ts` line 153:
```ts
readonly activeGoals = computed(() => this.goals.value.filter((g) => g.status === 'active'));
```

This signal is named `activeGoals` and its filter predicate makes `needs_human` goals invisible. The sidebar now uses `activeGoals` for both display and count (lines 92, 246, 250), so any goal that transitions to `needs_human` disappears from the sidebar entirely — exactly when it needs the most visibility.

**This bug pre-dates this PR** — it was introduced when `activeGoals` was added as a computed signal (see PR #569). This PR correctly uses the existing `activeGoals` signal rather than introducing a new filter, so the root cause is in `room-store.ts`, not here. But this PR is the first consumer of `activeGoals` in the sidebar UI, so it's the first place the bug becomes visible to users.

The fix belongs in `room-store.ts`: the `activeGoals` predicate should be `g.status === 'active' || g.status === 'needs_human'` (and the signal possibly renamed to `nonArchivedGoals` or `visibleGoals` to be more accurate). A follow-up issue tracking this would be appropriate given the PR is otherwise ready to merge.

---

### Minor notes (no change required)

- `tasksByGoalId` is built from `this.goals.value` (all goals), which means archived goals' linked tasks won't appear as orphan tasks either — they are neither shown under the goal nor under orphan Tasks. This is consistent behaviour once the `needs_human` issue above is fixed.
- `@ts-nocheck` on the test file is pre-existing; not introduced by this PR and not worth blocking on.
- `isRoomAgentSelected` comment at line 121 is clear and sufficient.

PR #582, opened 2026-03-21
fix: fix flaky online tests in CI

lsm:
## Summary
- **Null guards in afterEach/afterAll**: Added `if (!daemon) return;` to 18 online test files that crashed when `createDaemonServer()` timed out in `beforeEach`, leaving `daemon` undefined. Also increased `beforeEach` timeouts to 30s.
- **Port collision fix**: Changed in-process daemon server to use port 0 (OS-assigned) instead of `19400 + random(1000)`, eliminating port collisions when 17+ test shards run in parallel. Reads back actual port from `daemonContext.server.port`.
- **Copilot bridge retry**: Added retry-on-timeout for the "basic conversation" test which intermittently hangs when the Copilot API (gpt-5-mini) is slow. First attempt uses 60s timeout; on failure, interrupts and retries with a fresh session.

## CI Failure Analysis
- `Daemon Online (providers-anthropic-copilot)`: Consistent timeout at 120s in `waitForIdle` — Copilot API hangs
- `Daemon Online (rpc-2)`: `TypeError: undefined is not an object (evaluating 'daemon.waitForExit')` — `beforeEach` timeout cascades

## Test plan
- [x] `bun run typecheck` passes
- [x] `bun run lint` passes  
- [x] `bun run format:check` passes
- [x] `rpc-remove-output.test.ts` passes locally (9/9 tests)
- [x] `rpc-agent-handlers.test.ts` passes locally (6/6 tests)
- [x] `rpc-state-sync.test.ts` passes locally (10/10 tests)
- [ ] CI online test shards pass

### DIFF
--- a/packages/daemon/tests/helpers/daemon-server.ts
+++ b/packages/daemon/tests/helpers/daemon-server.ts
@@ -32,8 +32,9 @@ import {
 
 export interface DaemonServerOptions {
 	/**
-	 * Port for the daemon server
-	 * Default: 0 (OS-assigned) for in-process mode, random for spawned mode
+	 * Port for the daemon server.
+	 * Default: 0 (OS-assigned). The actual port is read back from the server
+	 * after startup (via server.port for in-process, stdout parsing for spawned).
 	 */
 	port?: number;
 
@@ -348,7 +349,7 @@ async function acquireDevProxyLease(
  */
 async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<DaemonServerContext> {
 	const {
-		port: userPort = 19400 + Math.floor(Math.random() * 1000),
+		port: userPort = 0, // Use port 0 for OS-assigned port; actual port parsed from stdout
 		env: customEnv = {},
 		devProxy: devProxyOptions,
 		useDevProxy = false,
@@ -390,9 +391,10 @@ async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<Dae
 		detached: false,
 	});
 
-	// Wait for the server to be ready
+	// Wait for the server to be ready and parse the actual port from stdout
 	let stderrOutput = '';
 	let stdoutOutput = '';
+	let actualPort = userPort;
 	await new Promise<void>((resolve, reject) => {
 		const timeout = setTimeout(() => reject(new Error('Daemon server startup timeout')), 20000);
 
@@ -403,7 +405,10 @@ async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<Dae
 			if (process.env.TEST_VERBOSE) {
 				console.error(`[DAEMON-PROCESS] ${output.trim()}`);
 			}
-			if (output.includes('Running on port')) {
+			// Parse actual port from "Running on port XXXX" output
+			const portMatch = output.match(/Running on port (\d+)/);
+			if (portMatch) {
+				actualPort = Number.parseInt(portMatch[1], 10);
 				clearTimeout(timeout);
 				daemonProcess.stdout!.off('data', onData);
 				daemonProcess.stderr!.off('data', onData);
@@ -430,7 +435,7 @@ async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<Dae
 	});
 
 	// Create WebSocket client to communicate with the daemon
-	const wsUrl = `ws://127.0.0.1:${userPort}/ws`;
+	const wsUrl = `ws://127.0.0.1:${actualPort}/ws`;
 	const transport = new WebSocketClientTransport({
 		url: wsUrl,
 		autoReconnect: false, // Don't auto-reconnect in tests
@@ -461,7 +466,7 @@ async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<Dae
 	return {
 		pid: daemonProcess.pid!,
 		messageHub,
-		baseUrl: `http://127.0.0.1:${userPort}`,
+		baseUrl: `http://127.0.0.1:${actualPort}`,
 		devProxy,
 		kill: (signal: NodeJS.Signals = 'SIGTERM') => daemonProcess.kill(signal),
 		waitForExit: async () => {
--- a/packages/daemon/tests/helpers/daemon-server.ts
+++ b/packages/daemon/tests/helpers/daemon-server.ts
@@ -33,7 +33,7 @@ import {
 export interface DaemonServerOptions {
 	/**
 	 * Port for the daemon server
-	 * Default: random port in 19400-20400 range
+	 * Default: 0 (OS-assigned) for in-process mode, random for spawned mode
 	 */
 	port?: number;
 
@@ -499,7 +499,7 @@ async function createInProcessDaemonServer(
 	options: DaemonServerOptions = {}
 ): Promise<DaemonServerContext & { daemonContext: DaemonAppContext }> {
 	const {
-		port: userPort = 19400 + Math.floor(Math.random() * 1000),
+		port: userPort = 0, // Use port 0 for OS-assigned port to avoid collisions in CI
 		env: customEnv = {},
 		devProxy: devProxyOptions,
 		useDevProxy = false,
@@ -576,8 +576,11 @@ async function createInProcessDaemonServer(
 		});
 	}
 
+	// Read back the actual port from the server (handles port 0 / OS-assigned ports)
+	const actualPort = daemonContext.server.port;
+
 	// Connect to the daemon's WebSocket server (just like a real client)
-	const wsUrl = `ws://127.0.0.1:${userPort}/ws`;
+	const wsUrl = `ws://127.0.0.1:${actualPort}/ws`;
 	const transport = new WebSocketClientTransport({
 		url: wsUrl,
 		autoReconnect: false, // Don't auto-reconnect in tests
@@ -614,7 +617,7 @@ async function createInProcessDaemonServer(
 	return {
 		pid: process.pid, // Same process
 		messageHub,
-		baseUrl: `http://127.0.0.1:${userPort}`,
+		baseUrl: `http://127.0.0.1:${actualPort}`,
 		daemonContext, // Expose for advanced usage
 		devProxy,
 		kill: () => {
--- a/packages/daemon/tests/helpers/standalone-server.ts
+++ b/packages/daemon/tests/helpers/standalone-server.ts
@@ -11,12 +11,12 @@
  * The SDK reads these environment variables directly.
  */
 
-const PORT = parseInt(process.env.PORT || '19400', 10);
+const PORT = parseInt(process.env.PORT || '0', 10);
 
 import { createDaemonApp } from '../../src/app';
 
 async function main() {
-	const { cleanup } = await createDaemonApp({
+	const { cleanup, server } = await createDaemonApp({
 		config: {
 			host: '127.0.0.1',
 			port: PORT,
@@ -50,7 +50,9 @@ async function main() {
 		process.exit(0);
 	});
 
-	console.error(`[DAEMON-SERVER] Running on port ${PORT}, PID: ${process.pid}`);
+	// Log actual bound port (not the requested port) so the parent process
+	// can parse it when using port 0 (OS-assigned).
+	console.error(`[DAEMON-SERVER] Running on port ${server.port}, PID: ${process.pid}`);
 }
 
 main().catch((error) => {
--- a/packages/daemon/tests/online/git/archive-session.test.ts
+++ b/packages/daemon/tests/online/git/archive-session.test.ts
@@ -41,6 +41,7 @@ describe('Archive Session', () => {
 	}, SETUP_TIMEOUT);
 
 	afterEach(async () => {
+		if (!daemon) return;
 		await daemon.waitForExit();
 	}, SETUP_TIMEOUT);
 
--- a/packages/daemon/tests/online/lifecycle/sigint-cleanup.test.ts
+++ b/packages/daemon/tests/online/lifecycle/sigint-cleanup.test.ts
@@ -38,13 +38,14 @@ describe('SDK SIGINT Cleanup (Online)', () => {
 	}, 30000);
 
 	afterEach(async () => {
-		// Reset spawned mode flag
+		// Always reset spawned mode flag (even if daemon is undefined)
 		delete process.env.DAEMON_TEST_SPAWN;
 		if (!daemon) return;
 		// Kill the daemon server after each test
+		// Spawned mode graceful shutdown may take longer than in-process
 		daemon.kill('SIGTERM');
 		await daemon.waitForExit();
-	}, 15_000);
+	}, 30_000);
 
 	describe('SIGINT during active SDK query', () => {
 		test(
--- a/packages/daemon/tests/online/lifecycle/sigint-cleanup.test.ts
+++ b/packages/daemon/tests/online/lifecycle/sigint-cleanup.test.ts
@@ -38,14 +38,13 @@ describe('SDK SIGINT Cleanup (Online)', () => {
 	}, 30000);
 
 	afterEach(async () => {
-		// Kill the daemon server after each test
-		if (daemon) {
-			daemon.kill('SIGTERM');
-			await daemon.waitForExit();
-		}
 		// Reset spawned mode flag
 		delete process.env.DAEMON_TEST_SPAWN;
-	});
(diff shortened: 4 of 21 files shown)


lsm (commented):

Overall this is a solid, pragmatic fix for the three root causes of CI flakiness. The null guards, port 0, and retry logic all address real issues. A few items to consider:

### P2: Spawned mode still uses random ports (line 351)

`spawnDaemonServer()` at `packages/daemon/tests/helpers/daemon-server.ts:351` still uses `19400 + Math.floor(Math.random() * 1000)` for port selection. The port 0 fix was only applied to in-process mode. Currently only `sigint-cleanup.test.ts` uses spawned mode so collision risk is low, but this is an inconsistency that could bite future tests. The spawned mode could also use port 0 — the server stdout already emits the port number in "Running on port XXXX" which is parsed during startup. You would just need to extract the actual port from that output instead of using `userPort`.

### P2: Copilot retry test timeout budget may be insufficient

In `packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts`, the test timeout is `TEST_TIMEOUT = 150_000` (150s). The worst-case execution path with retry is: attempt 1 `waitForIdle` timeout (60s) + interrupt + attempt 2 `waitForIdle` (up to 60s) + `waitForSdkMessages` (up to 60s) = ~180s, which exceeds the 150s test timeout. If the Copilot API is slow but not completely dead, the test could be killed by the outer timeout during a successful second attempt. Consider either bumping `TEST_TIMEOUT` for this test or reducing `PER_ATTEMPT_TIMEOUT`.

### P3: Null guard style inconsistency (minor)

Most files use the early-return pattern `if (!daemon) return;` but `rpc-remove-output.test.ts` (line 42) uses `if (daemon) { ... }`. Both work correctly — the latter is actually more appropriate there since post-daemon cleanup (env var restore, dir removal) must still execute. Not a bug, just noting the two styles.

### What looks good

- **Port 0 for in-process mode** is the correct fix. Reading back `daemonContext.server.port` after bind is the right approach. This eliminates the root cause of parallel CI shard port collisions.
- **Null guards in afterEach/afterAll** are the right fix for the cascading `TypeError` failures. When `createDaemonServer()` times out in `beforeEach`, the test framework still runs cleanup hooks where `daemon` is undefined.
- **Increased `beforeEach` timeouts to 30s** is reasonable for CI environments where server startup can be slow.
- The copilot retry logic properly creates a fresh session on retry rather than reusing a potentially corrupted one, and interrupts the stuck session first.

**Recommendation: REQUEST_CHANGES** — the P2 timeout budget issue in the copilot test could cause the retry to be killed during a valid second attempt, undermining the fix.

lsm (commented):

---

The three changes address real CI pain points. The port-0 fix is correct and complete for in-process mode. The null guard pattern is consistent and applied correctly. The copilot retry logic has a genuine timeout budget problem. Details below.

---

### P0 — Copilot retry: TEST_TIMEOUT too small for two full attempts

`packages/daemon/tests/online/providers/anthropic-to-copilot-bridge-provider.test.ts`

The per-attempt budget is `PER_ATTEMPT_TIMEOUT = 60_000` ms but the outer test timeout remains `TEST_TIMEOUT = IDLE_TIMEOUT + 30_000 = 150_000` ms.

Worst-case path on two attempts:
- Attempt 1: `waitForIdle` times out at 60 s → interrupt → `continue`
- session.create + sendMessage overhead on attempt 2: ~2–5 s
- Attempt 2: `waitForIdle` runs up to 60 s → success
- `waitForSdkMessages` timeout: `PER_ATTEMPT_TIMEOUT = 60_000` ms

Total: 60 + ~5 + 60 + 60 = ~185 s, which exceeds the 150 s outer timeout. If attempt 2 is also slow, Bun will kill the test with a timeout error that looks identical to the original hang — effectively defeating the retry.

The fix needs to either (a) increase `TEST_TIMEOUT` to cover two full attempts plus the `waitForSdkMessages` budget, or (b) keep the `waitForSdkMessages` timeout shorter on the success path.

### P1 — Spawned mode retains the old random-port approach

`packages/daemon/tests/helpers/daemon-server.ts`, line 351:

```ts
port: userPort = 19400 + Math.floor(Math.random() * 1000),
```

The PR description says "Default: 0 (OS-assigned) for in-process mode, random for spawned mode" — and the interface comment says the same. But spawned mode is used by `sigint-cleanup.test.ts` (`DAEMON_TEST_SPAWN=true`). Spawned mode works differently because the server runs in a child process and the port must be communicated via the env var `PORT`. Switching to port 0 in spawned mode would require parsing the startup log to read back the actual port, which the current `waitForTcpPortOpen` loop cannot do.

So the partial fix is defensible IF spawned mode is only used by `sigint-cleanup.test.ts` and that test does not run in parallel with other shards that could collide. In its current form this is a real but lower-priority gap, since the copilot test (the failing shard) uses in-process mode. Worth documenting explicitly why spawned mode was not changed, or adding a TODO.

### P1 — rpc-remove-output.test.ts uses `if (daemon)` instead of `if (!daemon) return`

`packages/daemon/tests/online/rpc/rpc-remove-output.test.ts`, lines 42–44:

```ts
if (daemon) {
    await daemon.waitForExit();
}
```

All 17 other changed files use `if (!daemon) return;` as an early exit. This file alone uses the wrapped-if style. The two are semantically equivalent for daemon cleanup, but there is a subtle difference: after the daemon block, this file continues to run the `TEST_SDK_SESSION_DIR` cleanup and env restoration unconditionally. That is actually correct — those cleanups should run regardless of whether daemon started — but the inconsistent guard style is confusing. Additionally, this is the only file that still has the `// Guard: daemon may be undefined if beforeEach timed out or failed` comment while others omit any comment, making the intent inconsistent.

### P2 — sigint-cleanup.test.ts: env var cleanup runs before the null guard

`packages/daemon/tests/online/lifecycle/sigint-cleanup.test.ts`, lines 40–47:

```ts
afterEach(async () => {
    delete process.env.DAEMON_TEST_SPAWN;  // runs first
    if (!daemon) return;                   // then guard
    daemon.kill('SIGTERM');
    await daemon.waitForExit();
}, 15_000);
```

This is actually fine functionally — deleting the env var before the guard means it always gets cleaned up. But note the original code had the env var cleanup inside the `if (daemon)` block, which means the env var would NOT have been cleared when `daemon` was undefined. The refactoring silently fixed a secondary bug (env var leaked on timeout). This should be documented so the change does not appear as a pure cosmetic reorder.

Also: the new 15_000 ms `afterEach` timeout is tight. `daemon.kill('SIGTERM')` + `waitForExit()` in spawned mode involves a child process that may be in the middle of flushing logs or stopping sessions. If this test previously relied on the graceful shutdown completing in under 15 s, it may still be flaky. Consider 20–30 s.

### P2 — Incomplete null guard coverage across online test suite

The PR protects the 18 files that were reported as failing. However, many other online test files already have `if (daemon) {` guards (some added previously), and a handful of files like `rpc-draft-handlers.test.ts` use `daemon?.waitForExit()` (optional chaining). None of the unmodified files appear to have the same unguarded pattern — but the suite is large (~32 files) and there is no central enforcement. A comment in the helper documenting the expected teardown pattern (or a lint rule) would prevent recurrence.

### P3 — beforeEach timeout inconsistency

Some files received the 30_000 ms `beforeEach` timeout upgrade (e.g., `rpc-agent-handlers.test.ts`) while others in the same suite did not (e.g., `rpc-draft-handlers.test.ts`, `rpc-interrupt-handlers.test.ts`, `rpc-model-handlers.test.ts`). These are pre-existing gaps, but the PR only fixes the ones that appeared in the reported shard failures. If a different shard times out in future, the same cascading failure will recur.

---

**Summary of what is correct:**

- Port-0 fix in `createInProcessDaemonServer` is correct. `Bun.serve({ port: 0 })` returns an OS-assigned port that is then read back via `daemonContext.server.port` (Bun's `Server.port` property is populated after bind). The `wsUrl` and `baseUrl` both use `actualPort`. This is the right fix for CI parallel shard port collisions.
- Null guards in cleanup hooks are correct and consistent (the `if (!daemon) return` idiom used in 17 files is clean). The `afterEach` timeout bumps (15_000 ms) are appropriate for in-process teardown.
- The copilot retry approach is directionally correct — interrupt + fresh session is better than letting a hung session leak — but the timeout budget needs to be sized correctly.

lsm (commented):

### Re-review of fixes for prior feedback

**Issue 1 (Copilot retry timeout budget)** -- Fixed correctly. `waitForSdkMessages` timeout reduced from `PER_ATTEMPT_TIMEOUT` (60s) to `10_000`. Budget comment added: 60+5+60+10=135s, well within the 150s `TEST_TIMEOUT`. Good.

**Issue 2 (Spawned mode port)** -- Partially fixed. **P0 bug remains.**

The `daemon-server.ts` spawned path now defaults to `port: 0` and parses the actual port from the `Running on port (\d+)` stdout line. However, `standalone-server.ts` (the child process entry point) was NOT updated. It still logs the raw `PORT` env variable:

```
// standalone-server.ts line 14, 19, 53
const PORT = parseInt(process.env.PORT || '19400', 10);
const { cleanup } = await createDaemonApp({ config: { port: PORT, ... } });
console.error(`[DAEMON-SERVER] Running on port ${PORT}, PID: ${process.pid}`);
```

When `PORT=0`, `Bun.serve({ port: 0 })` binds to a random OS port, but the log prints `Running on port 0`. The parser in `daemon-server.ts` will extract `0`, then connect to `ws://127.0.0.1:0/ws`, which will fail.

**Fix needed:** Destructure `server` from `createDaemonApp` and log `server.port`:
```ts
const { cleanup, server } = await createDaemonApp({...});
// ...
console.error(`[DAEMON-SERVER] Running on port ${server.port}, PID: ${process.pid}`);
```

This means the sigint-cleanup test (the only test using spawned mode) is broken. It sets `DAEMON_TEST_SPAWN=true`, which calls `spawnDaemonServer()` with `port: 0`, and the child process will log port 0.

**Issue 3 (rpc-remove-output null guard style)** -- Fixed correctly. Comment explains why wrapped-if is used (env cleanup must run after).

**Issue 4 (sigint-cleanup afterEach timeout)** -- Fixed correctly. Bumped from 15s to 30s with explanatory comment.

### Summary

| Issue | Status |
|-------|--------|
| Copilot retry timeout budget | Fixed |
| Spawned mode port | **Not fixed -- standalone-server.ts still logs PORT=0** |
| rpc-remove-output null guard style | Fixed |
| sigint-cleanup afterEach timeout | Fixed |

**P0:** `standalone-server.ts` must log `server.port` instead of `PORT` for the spawned-mode port 0 fix to work. Without this, spawned-mode tests will fail to connect.

lsm (commented):

Three of the four prior issues are correctly resolved. One fix — spawned mode port 0 — is incomplete and broken.

---

### P0 — Spawned mode port-0 fix is incomplete: `standalone-server.ts` not updated

The worker updated `daemon-server.ts` to pass `PORT=0` and parse the actual port from the `"Running on port XXXX"` stdout message. But `standalone-server.ts` — the script that actually spawns — was never modified. Two problems remain:

**Problem 1 — Default port is still 19400, not 0.**
`packages/daemon/tests/helpers/standalone-server.ts`, line 14:
```ts
const PORT = parseInt(process.env.PORT || '19400', 10);
```
When `spawnDaemonServer` does not pass `PORT` in `daemonEnv`, the spawned process falls back to `19400`, reintroducing the original collision risk. (The helper _does_ always pass `PORT: userPort.toString()`, so this only matters if `PORT` is somehow absent, but the default should still be `0` to match the stated intention.)

**Problem 2 — Logs the requested port, not the OS-assigned port.**
`packages/daemon/tests/helpers/standalone-server.ts`, line 53:
```ts
console.error(`[DAEMON-SERVER] Running on port ${PORT}, PID: ${process.pid}`);
```
`PORT` here is the value passed in via env var — when `PORT=0`, this logs `"Running on port 0"`. The `spawnDaemonServer` helper then parses `actualPort = 0` and tries to connect to `ws://127.0.0.1:0/ws`, which is an invalid address. The actual OS-assigned port is only available as `server.port` from the return value of `createDaemonApp`, but `standalone-server.ts` destructures only `{ cleanup }` and never reads `server.port`.

The fix requires `standalone-server.ts` to:
1. Default `PORT` to `0` (not `19400`)
2. Destructure `server` from `createDaemonApp`, then print `server.port` rather than `PORT`

As-is, running `sigint-cleanup.test.ts` (or any future spawned-mode test) with the default `PORT=0` will result in a WebSocket connection failure.

---

### P3 — `rpc-remove-output.test.ts` afterEach timeout is still 15s

`packages/daemon/tests/online/rpc/rpc-remove-output.test.ts`, line 62: `}, 15_000)`. The previous review flagged this as too short for spawned mode. The worker bumped `sigint-cleanup.test.ts` but this file was not adjusted. This test uses `createDaemonServer()` (which may use spawned mode via `DAEMON_TEST_SPAWN=true`), so the same concern applies. Low severity since this file currently only runs in-process mode by default, but it was called out in the original review.

---

### Verified Correct

**Issue 1 (timeout budget):** `PER_ATTEMPT_TIMEOUT = 60_000`, `waitForSdkMessages` timeout reduced to `10_000` on the success path. Comment budget: `60+5+60+10=135s` vs `TEST_TIMEOUT=150s`. Math checks out.

**Issue 2 (in-process port 0):** `createInProcessDaemonServer` correctly reads `daemonContext.server.port` after `createDaemonApp`. In-process mode is fully fixed.

**Issue 3 (rpc-remove-output style comment):** Comment at lines 41-42 explains the wrapped-if pattern clearly: "Uses wrapped-if (not early return) because env cleanup below must always run." Satisfactory.

**Issue 4 (sigint-cleanup afterEach timeout):** Confirmed bumped to `30_000` at line 48.

lsm (commented):

### Re-review: Previous P0 issue resolved

The previous blocking issue — `standalone-server.ts` logging the configured port (0) instead of the actual OS-assigned port — has been correctly fixed in commit `17d55c0e`.

**Fix 1: `packages/daemon/tests/helpers/standalone-server.ts`**
- Default port changed from `19400` to `0` (OS-assigned), consistent with the spawned-mode port 0 strategy
- Now destructures `server` from `createDaemonApp` return value
- Logs `server.port` (the actual bound port) instead of `PORT` (the requested port)
- The parent process in `daemon-server.ts` already parses this via `/Running on port (\d+)/` regex — confirmed the format matches

**Fix 2: `packages/daemon/tests/online/rpc/rpc-remove-output.test.ts`**
- `afterEach` timeout bumped from 15s to 30s — reasonable for CI environments where cleanup may be slow

Both fixes are minimal, correct, and well-targeted. No remaining issues.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

The two fixes in commits c7ea0483 and 17d55c0e correctly address the previously identified P0 issue and the accompanying timeout concern.

### Fix 1 — standalone-server.ts logs actual bound port (commit 17d55c0e)

The original code logged `PORT` (the *requested* port, defaulting to 19400) rather than the OS-assigned port. After this fix:

- Default `PORT` is now `0` (line 14 of `standalone-server.ts`).
- `createDaemonApp` is destructured to also receive `server` (line 19).
- The log message now uses `server.port` — the actual bound port — not the env variable (line 55).

The log format emitted is:
```
[DAEMON-SERVER] Running on port <N>, PID: <P>
```

This matches the regex `/Running on port (\d+)/` used in `spawnDaemonServer` (daemon-server.ts line 409) to extract `actualPort`, which is then used for both the WebSocket URL and the `baseUrl` return value.

The `DaemonAppContext.server` field is typed as Bun's `Server<WebSocketData>`, which exposes a `port` property populated after `Bun.serve()` binds to the OS port. This chain is sound.

### Fix 2 — rpc-remove-output.test.ts afterEach timeout bumped to 30s (commit 17d55c0e)

The `afterEach` timeout was 15s; it's now 30s to give spawned-mode graceful shutdown adequate room. The same bump was applied to `sigint-cleanup.test.ts` in the preceding commit (c7ea0483). Both changes are consistent with the in-process `waitForExit` having a 10s internal cap plus overhead.

### No issues found

Both fixes are minimal and correct:
- No over-engineering or unintended side effects.
- The log-format contract between `standalone-server.ts` and the regex in `daemon-server.ts` is consistent.
- The `actualPort` variable is initialised to `userPort` as a safe fallback (though with `userPort=0`, a non-parsed port would fail to connect, which is fine — it would surface as a test error rather than a silent misconfiguration).
- Timeout budgets are internally consistent across both affected test files.

PR #596, opened 2026-03-21
fix: recover rate-limited workers stuck in later iterations (feedbackIteration > 0)

lsm:
recoverStuckWorkers() had a feedbackIteration > 0 guard that prevented
recovery for workers paused by rate limit during a second (or later)
iteration. When the rate-limit timer fired and called scheduleTick(),
executeTick() ran recoverStuckWorkers() which skipped all groups with
feedbackIteration > 0 — leaving them permanently stuck with an expired
backoff.

The fix: allow recoverStuckWorkers() to re-trigger onWorkerTerminalState
when a group has an expired rate limit (non-null rateLimit where
Date.now() >= resetsAt), regardless of feedbackIteration. The existing
newMessages.length === 0 check acts as a safety guard for the leader-side
rate-limit case (worker messages already forwarded, nothing new to route).

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2205,28 +2205,43 @@ export class RoomRuntime {
 	 * 1. Observer callback fired but the routing threw an error (now logged, but still need recovery)
 	 * 2. Observer callback was missed due to a race condition (extremely rare)
 	 * 3. Any other silent failure in the worker→leader routing path
+	 * 4. Worker paused by rate limit — when the backoff expires the timer fires scheduleTick()
+	 *    which calls this function; the group is re-triggered regardless of feedbackIteration.
 	 *
 	 * Conditions for a "stuck worker":
-	 * - feedbackIteration == 0: no review rounds have completed (worker → leader routing never happened)
+	 * - feedbackIteration == 0: no review rounds have completed (worker → leader routing never
+	 *   happened), OR the group has an expired rate limit that was set during a later iteration —
+	 *   in that case the leader was never triggered (onWorkerTerminalState returned early after
+	 *   detecting the rate limit) so recovery is still needed.
 	 * - Worker session IS in the session factory (not a zombie)
 	 * - Worker session processing state is terminal (idle or interrupted)
 	 * - Leader session may or may not exist (with eager init, it always exists; with old lazy
 	 *   init, it may not exist yet — both cases are handled by routeWorkerToLeader)
 	 * - Group is NOT awaiting human review
-	 * - Group is NOT rate-limited
+	 * - Group is NOT actively rate-limited (resetsAt still in the future)
 	 * - Group is NOT paused waiting for a question answer (waiting_for_input is intentional pause)
 	 * - A recovery for this group is NOT already in-flight from a previous tick
 	 */
 	private recoverStuckWorkers(): void {
 		if (!this.sessionFactory.getProcessingState) return; // getProcessingState is optional
 
+		const now = Date.now();
 		const activeGroups = this.groupRepo.getActiveGroups(this.roomId);
 		for (const group of activeGroups) {
-			// Only recover groups that haven't routed to leader yet
-			if (group.feedbackIteration > 0) continue;
+			// An expired rate limit means the worker was paused mid-iteration (feedbackIteration may
+			// be > 0) and the leader was never triggered. Allow recovery in that case.
+			// isRateLimited() already returns false for expired limits, so we check the raw field.
+			const hasExpiredRateLimit = group.rateLimit !== null && now >= group.rateLimit.resetsAt;
+
+			// Skip groups where the leader is actively working (feedbackIteration > 0 means the
+			// worker→leader routing already happened at least once and the leader may still be
+			// reviewing). The exception is an expired rate limit: in that case onWorkerTerminalState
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-targeted fix for a real bug where workers that hit a rate limit during `feedbackIteration > 0` would get permanently stuck. The implementation and tests are correct.

### Analysis

**The Bug**: `recoverStuckWorkers()` unconditionally skipped any group with `feedbackIteration > 0`. When a worker hit a rate limit during a second (or later) iteration, the rate limit timer would fire `scheduleTick()` -> `executeTick()` -> `recoverStuckWorkers()`, but recovery was skipped because `feedbackIteration > 0`. The rate limit would expire but no one would re-trigger `onWorkerTerminalState`, leaving the worker permanently stuck.

**The Fix** (`room-runtime.ts:2241`): The guard is relaxed from:
```
if (group.feedbackIteration > 0) continue;
```
to:
```
if (group.feedbackIteration > 0 && !hasExpiredRateLimit) continue;
```

This is correct because when a worker hits a rate limit, `onWorkerTerminalState` returns early at line 498-501 (before incrementing `feedbackIteration` for the current iteration), so the leader is idle and recovery is safe.

**Safety layers verified**:
1. `isRateLimited()` at line 2245 still blocks groups with active (non-expired) rate limits -- correct, since `isRateLimited()` checks `Date.now() < resetsAt`.
2. The `newMessages.length === 0` check at line 2273 prevents re-routing for leader-side rate limits (where worker messages were already forwarded). This is explicitly tested.
3. Re-entry into `onWorkerTerminalState`: the `!group.rateLimit` guard at line 556 prevents re-setting a backoff on the already-expired-but-non-null rate limit, allowing the flow to fall through to worktree cleanup/retry.
4. The `stuckWorkerRecoveryInFlight` set prevents duplicate concurrent recovery calls.

**Tests**: All three new/modified tests are well-constructed:
- Positive case: expired worker rate limit with `feedbackIteration=1` triggers recovery, verifies `injectMessage` on leader and `feedbackIteration` incremented to 2.
- Negative case: expired leader rate limit with no new worker messages correctly skips re-routing.
- Existing test name updated for clarity.

All 32 tests in the file pass.

No issues found.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix is correct and targeted. The root cause — `recoverStuckWorkers()` unconditionally skipping all `feedbackIteration > 0` groups — is accurately identified, and the one-line condition change (`if (group.feedbackIteration > 0 && !hasExpiredRateLimit) continue;`) is the minimal surgical fix. The logic is safe: `isRateLimited()` still guards against premature recovery (active backoff is still respected at line 2245), and the `getWorkerMessages` check on `lastForwardedMessageId` correctly handles the leader-rate-limit case as a secondary safety net.

### What I checked

**`recoverStuckWorkers()` logic (lines 2225–2308)**

The ordering of guards is correct and no existing safety invariants are violated:

1. `feedbackIteration > 0 && !hasExpiredRateLimit` — the modified guard correctly allows expired-rate-limit groups through regardless of iteration count.
2. `isRateLimited()` — still blocks groups with an *active* backoff. Because `isRateLimited()` returns `false` once the backoff expires, the two checks are consistent and non-overlapping.
3. `getWorkerMessages(...lastForwardedMessageId)` returning empty — correctly prevents spurious re-routing for the leader-rate-limit case (messages were already forwarded before the leader hit 429, so `lastForwardedMessageId` is current). This is the most important safety net for preventing double-routing.
4. `stuckWorkerRecoveryInFlight` set — correctly prevents concurrent recovery calls from successive ticks.

**`onWorkerTerminalState()` integration (lines 550–583)**

When re-triggered after rate limit expiry, the `!group.rateLimit` guard on line 556 is `false` (rateLimit is non-null, just expired), so the handler correctly falls through to the worktree and exit-gate checks instead of re-setting a new backoff. This is the correct behavior and is well-explained in the comment.

**Tests (3 new in `room-runtime-recovery-safety.test.ts`)**

1. `should re-trigger routing for feedbackIteration > 0 groups with an expired rate limit` — covers the bug scenario end-to-end: expired worker rate limit, `feedbackIteration=1`, verifies `injectMessage` fires on leader and `feedbackIteration` becomes 2.
2. `should NOT re-trigger routing for feedbackIteration > 0 with leader-side expired rate limit` — covers the safety guard: `getWorkerMessages` returns empty, verifies no routing calls happen.
3. `should skip stuck-worker recovery for groups with feedbackIteration > 0 and no expired rate limit` — existing test updated to reflect that the `feedbackIteration > 0` skip is now conditional on no expired rate limit.

All 32 tests pass.

- The `hasExpiredRateLimit` variable is computed for every group regardless of whether the `feedbackIteration > 0` branch is taken. This is a micro-optimization non-issue — the check is O(1).
- The `reason` string at line 2285–2287 adds helpful context to the warn log for the new case (`rate limit expired (feedbackIteration=N)`). This is a good addition.
- There is no test for the scenario where `humanInterrupted = true` AND an expired rate limit coexist. In that case `recoverStuckWorkers` would fire recovery, `onWorkerTerminalState` would clear `humanInterrupted` and return early without routing — the human interrupt wins. This behavior is correct but undocumented; a comment noting this precedence would be a minor improvement for future readers. Not a bug.

No P0, P1, or P2 issues found. The fix is correct, safe, and well-tested.

PR #599, opened 2026-03-21
fix: update room-agent-tools guards and comments for new task lifecycle

lsm:
- Update send_message_to_task cancelled guard: change "restart" to
  "reactivate" in error message, reflecting that worktrees are kept
- Extend set_task_status restart/revive block to include completed tasks
  so completed→in_progress triggers resetGroupForRestart (same as
  cancelled tasks), enabling proper agent group reset on reactivation
- Update reviveGroup comment to clarify scope: lightweight revive is
  needs_attention→review only; cancelled and completed use resetGroupForRestart

### DIFF
--- a/packages/daemon/src/lib/room/runtime/human-message-routing.ts
+++ b/packages/daemon/src/lib/room/runtime/human-message-routing.ts
@@ -7,8 +7,8 @@
  *
  * Callers are responsible for pre-processing terminated tasks before calling this
  * function. For example:
- * - needs_attention → caller should use runtime.reviveTaskForMessage()
- * - cancelled/completed → caller should block with a user-facing error
+ * - needs_attention/cancelled → caller should use runtime.reviveTaskForMessage()
+ * - completed → caller should use runtime.reviveTaskForMessage() or block with an error
  */
 
 import type { TaskStatus } from '@neokai/shared';
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -376,8 +376,12 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				}
 			}
 
-			// Handle restart/revive: update group state for needs_attention/cancelled tasks
-			if (task.status === 'needs_attention' || task.status === 'cancelled') {
+			// Handle restart/revive: update group state for needs_attention, completed, and cancelled tasks
+			if (
+				task.status === 'needs_attention' ||
+				task.status === 'cancelled' ||
+				task.status === 'completed'
+			) {
 				const group = groupRepo.getGroupByTaskId(args.task_id);
 				if (group) {
 					if (args.status === 'pending' || args.status === 'in_progress') {
@@ -394,10 +398,9 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 						task.status === 'needs_attention' &&
 						group.completedAt !== null
 					) {
-						// Lightweight revive (needs_attention → review only): clear completedAt
-						// without resetting metadata. Only offered on this agent-tool path for
-						// needs_attention tasks — cancelled tasks use resetGroupForRestart above
-						// for a clean-slate restart, not a lightweight revive.
+						// Lightweight revive: clear completedAt without resetting metadata.
+						// Supported for needs_attention → review only. Cancelled and completed
+						// tasks use resetGroupForRestart() above for a clean slate.
 						const revived = groupRepo.reviveGroup(group.id);
 						if (!revived) {
 							return jsonResult({
@@ -513,13 +516,11 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 			// silently reactivate a task that was explicitly cancelled by a human. The
 			// worktree is still present (only archiveGroup cleans it up), but resuming
 			// without explicit human intent risks restarting undesired work.
-			// Use set_task_status to explicitly restart the task first.
+			// Use set_task_status to explicitly reactivate the task first.
 			if (task.status === 'cancelled') {
 				return jsonResult({
 					success: false,
-					error:
-						`Task ${args.task_id} is cancelled. Use set_task_status to restart it ` +
-						'(e.g. status: "pending" or "in_progress") before sending a message.',
+					error: `Task ${args.task_id} is cancelled. Use set_task_status to reactivate it first (e.g. status: "in_progress").`,
 				});
 			}
 
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -399,8 +399,8 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 						group.completedAt !== null
 					) {
 						// Lightweight revive: clear completedAt without resetting metadata.
-						// Supported for needs_attention → review only. Cancelled and completed
-						// tasks use resetGroupForRestart() above for a clean slate.
+						// Supported for needs_attention (failed) → review only. Cancelled and
+						// completed tasks use resetGroupForRestart() above for a clean slate.
 						const revived = groupRepo.reviveGroup(group.id);
 						if (!revived) {
 							return jsonResult({
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -512,57 +512,20 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				return jsonResult({ success: false, error: 'Room runtime not found' });
 			}
 
-			// Auto-reactivate: if the task is cancelled, reset the group and transition to
-			// in_progress so the message can be delivered without manual intervention.
-			// Worktrees are preserved on cancel (only archiveGroup cleans them up), so
-			// session restoration is safe.
-			if (task.status === 'cancelled') {
-				const group = groupRepo.getGroupByTaskId(args.task_id);
-				if (group) {
-					const reset = groupRepo.resetGroupForRestart(group.id);
-					if (!reset) {
-						return jsonResult({
-							success: false,
-							error: `Failed to reactivate task ${args.task_id} — group may have been modified concurrently`,
-						});
-					}
-				}
+			// Auto-revive: needs_attention and cancelled tasks auto-reactivate so the message
+			// can be delivered without manual intervention. Worktrees are preserved for both
+			// statuses (only archiveGroup cleans them up), so session restoration is safe.
+			//
+			// Deliberate asymmetry with set_task_status:
+			//   set_task_status(cancelled → in_progress) → resetGroupForRestart (clean slate)
+			//   send_message_to_task(cancelled task)    → reviveTaskForMessage  (keep history)
+			// Sending a message is a "continue this conversation" action; history is preserved.
+			if (task.status === 'needs_attention' || task.status === 'cancelled') {
+				// needs_attention transitions through 'review' (its prior working state);
+				// cancelled transitions directly to 'in_progress'.
+				const intermediateStatus = task.status === 'needs_attention' ? 'review' : 'in_progress';
 				try {
-					await taskManager.setTaskStatus(args.task_id, 'in_progress');
-				} catch (err) {
-					return jsonResult({
-						success: false,
-						error: `Failed to reactivate task ${args.task_id}: ${String(err)}`,
-					});
-				}
-
-				const revived = await runtime.reviveTaskForMessage(args.task_id, args.message);
-				if (!revived) {
-					// Roll back the task status; in_progress → cancelled is a valid transition.
-					try {
-						await taskManager.setTaskStatus(args.task_id, 'cancelled');
-					} catch {
-						// Rollback is best-effort; swallow to avoid masking the original error
-					}
-					return jsonResult({
-						success: false,
-						error:
-							`Failed to reactivate task ${args.task_id}: agent sessions could not be restored. ` +
-							'Task status has been reset to cancelled.',
-					});
-				}
-				return jsonResult({
-					success: true,
-					message: `Task ${args.task_id} reactivated from cancelled to in_progress and message delivered to agent`,
-				});
-			}
-
-			// Auto-revive: if the task needs attention, transition it to 'review' and
-			// restore the agent sessions so the message can be delivered. Needs_attention
-			// tasks preserve their worktree, so session restoration is safe.
-			if (task.status === 'needs_attention') {
-				try {
-					await taskManager.setTaskStatus(args.task_id, 'review');
+					await taskManager.setTaskStatus(args.task_id, intermediateStatus);
 				} catch (err) {
 					return jsonResult({
 						success: false,
@@ -572,9 +535,9 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 
 				const revived = await runtime.reviveTaskForMessage(args.task_id, args.message);
 				if (!revived) {
-					// Roll back the task status; review → needs_attention is a valid transition.
+					// Roll back the task status to its original state.
 					try {
-						await taskManager.setTaskStatus(args.task_id, 'needs_attention');
+						await taskManager.setTaskStatus(args.task_id, task.status);
 					} catch {
 						// Rollback is best-effort; swallow to avoid masking the original error
 					}
@@ -582,12 +545,13 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 						success: false,
 						error:
 							`Failed to revive task ${args.task_id}: agent sessions could not be restored. ` +
-							'Task status has been reset to needs_attention.',
+							`Task status has been reset to ${task.status}.`,
 					});
 				}
+				const fromTo = `${task.status} to ${intermediateStatus}`;
 				return jsonResult({
 					success: true,
-					message: `Task ${args.task_id} revived from needs_attention to review and message delivered to agent`,
+					message: `Task ${args.task_id} revived from ${fromTo} and message delivered to agent`,
 				});
 			}
 
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -512,15 +512,48 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				return jsonResult({ success: false, error: 'Room runtime not found' });
 			}
 
-			// Cancelled tasks are blocked from agent-tool messaging — an agent should not
-			// silently reactivate a task that was explicitly cancelled by a human. The
-			// worktree is still present (only archiveGroup cleans it up), but resuming
-			// without explicit human intent risks restarting undesired work.
-			// Use set_task_status to explicitly reactivate the task first.
+			// Auto-reactivate: if the task is cancelled, reset the group and transition to
+			// in_progress so the message can be delivered without manual intervention.
+			// Worktrees are preserved on cancel (only archiveGroup cleans them up), so
+			// session restoration is safe.
 			if (task.status === 'cancelled') {
+				const group = groupRepo.getGroupByTaskId(args.task_id);
+				if (group) {
+					const reset = groupRepo.resetGroupForRestart(group.id);
+					if (!reset) {
+						return jsonResult({
+							success: false,
+							error: `Failed to reactivate task ${args.task_id} — group may have been modified concurrently`,
+						});
+					}
+				}
+				try {
+					await taskManager.setTaskStatus(args.task_id, 'in_progress');
+				} catch (err) {
+					return jsonResult({
+						success: false,
+						error: `Failed to reactivate task ${args.task_id}: ${String(err)}`,
+					});
+				}
+
+				const revived = await runtime.reviveTaskForMessage(args.task_id, args.message);
+				if (!revived) {
+					// Roll back the task status; in_progress → cancelled is a valid transition.
+					try {
+						await taskManager.setTaskStatus(args.task_id, 'cancelled');
+					} catch {
+						// Rollback is best-effort; swallow to avoid masking the original error
+					}
+					return jsonResult({
+						success: false,
+						error:
+							`Failed to reactivate task ${args.task_id}: agent sessions could not be restored. ` +
+							'Task status has been reset to cancelled.',
+					});
+				}
 				return jsonResult({
-					success: false,
-					error: `Task ${args.task_id} is cancelled. Use set_task_status to reactivate it first (e.g. status: "in_progress").`,
+					success: true,
+					message: `Task ${args.task_id} reactivated from cancelled to in_progress and message delivered to agent`,
 				});
 			}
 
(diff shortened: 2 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR correctly implements the three changes requested in the task. The diff is small (11 additions, 10 deletions) and focused.

### What was done

1. **send_message_to_task cancelled guard** (line 523): Error message updated from "restart" to "reactivate" wording. The new message is clear and actionable: `"Use set_task_status to reactivate it first (e.g. status: \"in_progress\")."` This also drops the `"pending"` suggestion which makes sense since the point is to get work moving again.

2. **Restart/revive block** (lines 380-383): Added `task.status === 'completed'` to the condition, so `completed -> in_progress` transitions now trigger `resetGroupForRestart()`. This is correct per the new lifecycle where `completed` tasks retain their worktrees and can be reactivated.

3. **reviveGroup comment** (lines 401-403): Updated to: "Supported for needs_attention -> review only. Cancelled and completed tasks use resetGroupForRestart() above for a clean slate." This is actually more accurate than the wording specified in the task ("Supported for failed and completed tasks") because `failed` is not a task status in this system, and the lightweight revive path is indeed only for `needs_attention -> review`.

4. **No remaining stale references**: Searched for "workspace has been cleaned up" and "worktree cleaned up" in the daemon source -- none found.

### Minor observations (P3)

- **Test file stale comments**: `room-agent-tools.test.ts` line 1026 still says "worktree is gone, restart required" and line 1049 says "workspace is cleaned up". These are test description strings / comments, not production code, but could be confusing to future readers. Not blocking.

- **No test for completed -> in_progress with group reset**: There are existing tests for `cancelled -> in_progress` and `failed -> pending` group resets, but no test covering `completed -> in_progress` triggering `resetGroupForRestart`. The path is exercised indirectly (same code path as the other statuses), but a dedicated test would increase confidence. Not blocking since the logic is shared.

- **Pre-existing: group reset before transition validation**: The group reset at line 389 runs before the status transition validation at line 418. If an invalid transition like `completed -> pending` is attempted, the group would be reset but the status change would fail. This is a pre-existing pattern (same issue exists for `cancelled` and `needs_attention`), not introduced by this PR, and the invalid transitions would only occur from buggy callers.

### CI Status

Lint, typecheck, unit tests, and most online tests pass. A few online test jobs are still pending but unrelated to this change.

Overall this is a clean, focused change that correctly updates the guards and comments for the new task lifecycle semantics.

lsm (commented):

The PR partially achieves the stated goals but has three gaps relative to the specification.

---

### P1 — Missing test for the new `completed -> in_progress` group-reset path

The core behavioral change — adding `task.status === 'completed'` to the restart/revive block — has no dedicated test. The existing `completed -> pending` test at line 1494 (`room-agent-tools.test.ts`) verifies that transition *fails* (which is unchanged behavior), but there is no test like "should reset old completed group when restarting task" that exercises the new `resetGroupForRestart` call for a completed task. The analogous tests for `needs_attention` (line 1218) and `cancelled` (line 1249) both exist. Without a test for `completed -> in_progress`, a regression in the group-reset path for completed tasks would go undetected.

**File:** `packages/daemon/tests/unit/room/room-agent-tools.test.ts`

---

### P2 — `reviveGroup` comment wording differs from specification

The task specification required this exact wording:

> "Lightweight revive: clear completedAt without resetting metadata. Supported for failed and completed tasks. Cancelled tasks use resetGroupForRestart() for a clean slate."

The implementation uses:

```
// Lightweight revive: clear completedAt without resetting metadata.
// Supported for needs_attention → review only. Cancelled and completed
// tasks use resetGroupForRestart() above for a clean slate.
```

The implementation's phrasing ("needs_attention → review only") is actually more precise and technically correct — the `reviveGroup` call only fires when `task.status === 'needs_attention'`. However, it contradicts the specification, which asked for "Supported for failed and completed tasks." If "failed" was an intentional alias for `needs_attention` in the spec's language, that should be clarified; if not, the comment diverges meaningfully from the spec.

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 401–403

---

### P2 — Stale test descriptions not updated

The task explicitly required searching for all references to "workspace has been cleaned up" or "worktree cleaned up." Two stale comments were missed:

1. `packages/daemon/tests/unit/room/room-agent-tools.test.ts`, line 1026:
   - Test name: `'should return error for cancelled task (worktree is gone, restart required)'`
   - This still says "worktree is gone" — contradicted by the new lifecycle where worktrees are preserved on cancel.

2. `packages/daemon/tests/unit/room/room-agent-tools.test.ts`, line 1049:
   - Comment: `// Cancelled tasks cannot receive messages — workspace is cleaned up`
   - The production code (line 516–519) was correctly updated to say "The worktree is still present (only archiveGroup cleans it up)..." but the test comment still carries the old incorrect rationale.

3. `packages/daemon/tests/unit/room/task-manager.test.ts`, line 858:
   - Test name: `'should reject cancelled → review transition (worktree is cleaned up on cancel)'`
   - Same stale claim — worktrees are NOT cleaned up on cancel per the new lifecycle. This file was not touched by the PR even though the task asked for a broad search.

---

### P3 — Slight over-inclusion in outer `if` guard for `completed`

In the restart/revive block (lines 380–413), `completed` is added to the outer guard. However `completed -> pending` is not a valid transition (it will be rejected by `taskManager.setTaskStatus` anyway), so the only reachable valid path for `completed` is `completed -> in_progress`. The guard is harmless but slightly over-broad. A tighter guard would be `task.status === 'completed' && (args.status === 'in_progress')` to match what the task manager actually allows.

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 380–384

lsm (commented):

**Recommendation: APPROVE**

Re-review after the worker addressed prior feedback from both claude-opus-4-6 and claude-sonnet-4-6. All three issues from the previous REQUEST_CHANGES review have been resolved.

### Issues addressed in second commit (c1e2b6b0)

1. **P1 (was missing test)**: A dedicated test for the `completed -> in_progress` group-reset path has been added at `room-agent-tools.test.ts` lines 1277-1304. The test correctly exercises the full lifecycle: create task, start it, insert a group, complete the task, then reactivate via `set_task_status` and verify the group is properly reset (`completedAt` null, `submittedForReview` false, `feedbackIteration` 0). This mirrors the existing analogous tests for `needs_attention` and `cancelled` paths.

2. **P2 (stale test descriptions)**: All three stale references have been updated:
   - `room-agent-tools.test.ts` line 1026: "worktree is gone, restart required" changed to "worktree is preserved, explicit reactivation required"
   - `room-agent-tools.test.ts` line 1049: "workspace is cleaned up" changed to "explicit reactivation required"
   - `task-manager.test.ts` line 858: "worktree is cleaned up on cancel" changed to "not a valid reactivation path"

3. **P2 (reviveGroup comment)**: The previous review noted the implementation's wording ("needs_attention -> review only") was actually more precise than the specification. The worker did add a clarifying note that "failed" is an alias for `needs_attention`, making the comment at lines 401-403 both accurate and clear.

### Verification

- All 112 tests in `room-agent-tools.test.ts` pass (0 failures).
- Searched for remaining stale "worktree is gone" / "workspace is cleaned up" references in the daemon package -- none found in PR-relevant files.
- The production code changes (first commit) remain clean: the `completed` status addition to the restart/revive guard, the updated error message wording, and the revised comments are all correct.

No issues found. Clean follow-up commit that addresses all prior feedback.

lsm (commented):

**Recommendation: APPROVE** (own-PR COMMENT event)

---

The worker has addressed all three items from the original task. Here is my assessment of what was changed and one remaining weakness.

### What was addressed

**1. send_message_to_task cancelled guard (room-agent-tools.ts line 519)**
The error message was updated from "restart" to "reactivate", and the stale reference to the worktree being gone was replaced with accurate language noting that explicit human intent is required. The comment above the guard now correctly states that the worktree is preserved and only `archiveGroup` cleans it up.

**2. reviveGroup comment (room-agent-tools.ts line 401-403)**
The comment was trimmed and clarified. It now correctly states that needs_attention is the only path that uses the lightweight revive, and that cancelled/completed tasks use `resetGroupForRestart()`. This is accurate.

**3. completed → in_progress group reset (room-agent-tools.ts line 383)**
The `task.status === 'completed'` case was added to the restart/revive guard. This means a completed task with a lingering group will now get its group reset via `resetGroupForRestart()` when transitioned to `in_progress`, matching the behaviour for cancelled tasks.

**4. task-manager.test.ts (line 855)**
The test description was updated from "worktree is cleaned up on cancel" to "not a valid reactivation path", which is accurate.

### P2 - New test has a weak setup that doesn't exercise the completedAt reset

**File:** `packages/daemon/tests/unit/room/room-agent-tools.test.ts`, lines 1277-1304

The new test ("should reset old completed group when restarting task") sets up the group using `insertGroup(taskId, 'awaiting_human')`, which inserts a group row with `completed_at = NULL`. It then calls `taskManager.completeTask(taskId, 'Done')`, which only updates the task's status column — it does NOT call `groupRepo.completeGroup()` and therefore does not set `completedAt` on the group.

As a result, the group's `completedAt` is already null before `set_task_status` is called. The assertion on line 1301 (`expect(groupAfter!.completedAt).toBeNull()`) is trivially satisfied regardless of whether `resetGroupForRestart` actually ran. The test does not verify the scenario that `resetGroupForRestart` is actually responsible for clearing `completedAt`.

The existing cancelled-group test (line 1249) has the same structural limitation — `insertGroup` also leaves `completedAt = null`.

The test does provide meaningful value by confirming the task status transitions from `completed` to `in_progress` successfully (lines 1295-1296) and that the group survives the round-trip (lines 1299-1303). The metadata reset (`feedbackIteration`) is tested, though the initial value is also zero so this only proves the group was not deleted. The most important missing assertion is verifying that a group with a non-null `completedAt` (i.e. one created via `groupRepo.completeGroup()`) gets its `completedAt` cleared to null by the reset path.

To make this test fully cover the intended scenario, the setup should call `groupRepo.completeGroup(groupId, group.version)` after `insertGroup`, then assert that `groupBefore.completedAt` is not null, before proceeding to reactivate.

### No blocking issues found

The production logic in `room-agent-tools.ts` is correct. The `completed` case is properly added to the guard at line 383, and the code correctly routes it to `resetGroupForRestart` rather than the lightweight `reviveGroup`. The error messages are accurate. The stale comments are fixed.

I recommend **APPROVE** despite the P2 test weakness — the core behaviour is correct and the test does provide partial coverage of the new path.

lsm (commented):

**Recommendation: APPROVE** (own-PR COMMENT event)

---

### Follow-up on P2 fix

The P2 I raised in the previous round has been fully addressed. The test now:

1. Calls `groupRepo.completeGroup(groupId, groupInserted!.version)` after `insertGroup` to set a real, non-null `completedAt` on the group (line 1287).
2. Asserts `groupBefore!.completedAt` is **not null** before reactivation (line 1293) — confirming the precondition is genuinely established.
3. Asserts `groupAfter!.completedAt` **is null** after calling `set_task_status` (line 1305) — this assertion is now meaningful, proving `resetGroupForRestart` actually cleared the field.

The method call `groupRepo.completeGroup(groupId, groupInserted!.version)` matches the actual method signature (`completeGroup(groupId: string, expectedVersion: number)`), so there is no API mismatch.

The `feedbackIteration` assertion on line 1307 is also meaningful for the same reason — the inserted group starts at a default value, `completeGroup` sets it, and `resetGroupForRestart` clears it back. This confirms a full metadata reset occurred.

### Summary

All three items from the original task are correctly implemented and tested:
- `send_message_to_task` cancelled guard message updated to "reactivate"
- `completed` added to the restart/revive guard so `resetGroupForRestart` runs on completed-task reactivation
- `reviveGroup` comment updated to accurately describe scope

No remaining issues.

lsm (commented):

### Overview

The PR changes `send_message_to_task` from blocking cancelled tasks with an error to auto-reactivating them: reset group, transition to `in_progress`, revive sessions, deliver message, roll back on failure. The pattern mirrors the existing `needs_attention` auto-revive path. The implementation is generally clean and follows established patterns. Two issues worth addressing:

### P1: Group reset is not rolled back on failure

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 519-557

When `reviveTaskForMessage()` fails (line 539-552), the code rolls back the task status from `in_progress` to `cancelled`, but it does **not** roll back the group state. By that point, `resetGroupForRestart()` (line 522) has already cleared `completedAt`, `submittedForReview`, `feedbackIteration`, and other metadata. After the rollback, the task is `cancelled` but its group has been wiped clean — this is an inconsistent state.

The `needs_attention` auto-revive path at line 563+ has the same omission, but there the group change is a lightweight `reviveGroup` (just clearing `completedAt`), and the transition is to `review` which preserves more context. Here, the full `resetGroupForRestart` is destructive — it wipes metadata that would be needed if the task is later reactivated through a different path.

Consider either (a) rolling back the group state (e.g. re-completing the group), or (b) reordering so `reviveTaskForMessage` is attempted before the group reset, or (c) documenting this as an accepted tradeoff.

### P2: Stale comment in `human-message-routing.ts`

**File:** `packages/daemon/src/lib/room/runtime/human-message-routing.ts`, line 11

The file-level docstring says:
```
 * - cancelled/completed → caller should block with a user-facing error
```

This is now incorrect for cancelled tasks — callers auto-reactivate instead of blocking. The comment should be updated to reflect the new behavior, since this file serves as the routing contract for callers.

### P3: Rollback test doesn't create a group

**File:** `packages/daemon/tests/unit/room/room-agent-tools.test.ts`, lines 1070-1097

The rollback failure test (`should roll back cancelled task when reviveTaskForMessage fails`) does not insert a group before cancelling. This means the `if (group)` branch at line 521 is never entered, so `resetGroupForRestart` is never called in this test. The test only exercises the no-group path. It would be more thorough to also test the rollback when a group exists, to verify the inconsistency noted in P1 (or verify it's handled).

### Positive notes

- The success test is well-structured: it properly sets up the group, verifies `completedAt` is non-null before reactivation, and checks both task status and group state afterward.
- Error handling for `setTaskStatus` failures is correct.
- The `set_task_status` extension to include `completed` in the restart/revive block is clean and the new test (`should reset old completed group when restarting task`) verifies it well.
- All 113 tests pass.

lsm (commented):

---

### Summary

The auto-reactivation logic in `room-agent-tools.ts` works mechanically, but it contains a meaningful semantic discrepancy with the parallel implementation in `task-handlers.ts`, a missing group-reset rollback on partial failure, and a test gap for the group state after rollback.

---

### P0 — Semantic mismatch with task-handlers.ts: history is wiped when it should be preserved

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 519–557

The new `send_message_to_task` path for cancelled tasks calls `resetGroupForRestart` before `reviveTaskForMessage`. `resetGroupForRestart` wipes all group metadata (feedback iterations, `submittedForReview`, conversation-related flags) to a clean default state.

The corresponding RPC handler in `task-handlers.ts` at lines 822–830 explicitly documents the **opposite** design:

```
// Note — deliberate asymmetry with task.setStatus:
//   task.setStatus(cancelled → in_progress)  → resetGroupForRestart (clean slate)
//   task.sendHumanMessage(cancelled task)     → reviveTaskForMessage  (keep history)
// Sending a message to a cancelled task is a "continue this conversation" action, so
// we preserve context. Explicitly restarting via setStatus is a "start over" action.
```

Both `send_message_to_task` (agent-tool path) and `task.sendHumanMessage` (RPC path) are logically the same operation: a human or agent delivers a message to a cancelled task. The two paths should have consistent semantics. Currently the agent-tool path wipes history while the RPC path preserves it. This will produce different agent behaviour depending on which path was used to trigger the action, which is almost certainly not intentional.

If the intent is to always do a clean-slate restart when sending a message to a cancelled task via the agent tool, that choice needs to be explicitly documented and justified — it contradicts the comment at the top of `task-handlers.ts`. If not, the `resetGroupForRestart` call at line 522 should be removed.

---

### P1 — Missing group-state rollback when `setTaskStatus` fails mid-sequence

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 519–537

The cancelled-task flow is:
1. `resetGroupForRestart(group.id)` — mutates DB
2. `setTaskStatus(task_id, 'in_progress')` — if this throws, early-returns an error

If step 2 fails, the code returns an error without rolling back the group-metadata wipe from step 1. The group is now in a reset/blank state while the task is still `cancelled`. This is an inconsistent half-applied state: the task is cancelled but the group metadata has been cleared.

The `needs_attention` path at line 563 does not call `resetGroupForRestart` before `setTaskStatus`, so it avoids this class of problem. The new cancelled path should either:
- Not call `resetGroupForRestart` before `setTaskStatus` (matching the RPC path approach), or
- Rollback the group reset if `setTaskStatus` fails.

---

### P1 — Test gap: rollback test does not verify group state after failure

**File:** `packages/daemon/tests/unit/room/room-agent-tools.test.ts`, lines 1070–1098

The "should roll back cancelled task when reviveTaskForMessage fails" test verifies that `task.status` returns to `cancelled`, but does not verify the group state. When `reviveTaskForMessage` fails, the task is rolled back but the group metadata that was wiped by `resetGroupForRestart` is **not** restored. The test should verify whether the group's `completedAt` and metadata state are consistent with the expected post-rollback invariant (or document explicitly that the group state is intentionally left in the wiped state).

As it stands the test passes but does not catch the partial inconsistency in the group state.

---

### P2 — PR description does not match implementation

The PR body says:
> "Update send_message_to_task cancelled guard: change 'restart' to 'reactivate' in error message, reflecting that worktrees are kept"

The actual change is far more significant: the entire error-return guard is removed and replaced with a full auto-reactivation flow including `resetGroupForRestart`, `setTaskStatus`, `reviveTaskForMessage`, and rollback. The description significantly understates the scope of the change, which could mislead future readers of the git log.

---

### P3 — Nit: comment at line 515 is imprecise

The comment says "Worktrees are preserved on cancel (only archiveGroup cleans them up), so session restoration is safe." This is true but the same statement appears on the `needs_attention` path. The distinct property of the cancelled path (the `resetGroupForRestart` wipe vs lightweight revive) is not called out, which would help readers understand why this path behaves differently.

lsm (commented):

**Recommendation: APPROVE**

### Summary

All prior feedback has been addressed. The implementation is correct and the tests are well-structured. One observation worth noting but not blocking.

### What was verified

1. **`send_message_to_task` cancelled path** (room-agent-tools.ts lines 515-556): The old blocking guard has been replaced with auto-reactivation that mirrors the `needs_attention` path. Cancelled tasks now transition to `in_progress` (not `review`), revive via `reviveTaskForMessage`, and roll back to `cancelled` on failure. The deliberate asymmetry with `set_task_status` is well-documented in the comment at lines 519-522.

2. **Unified `needs_attention`/`cancelled` block**: The two statuses share one `if` block with `intermediateStatus` computed via ternary (line 526). Clean and readable.

3. **`human-message-routing.ts` comment** (lines 10-11): Updated to reflect the new routing contract -- `needs_attention`/`cancelled` use `reviveTaskForMessage`, `completed` can use revive or block.

4. **Tests are thorough and meaningful**:
   - Success test (line 1026): Sets up a group with non-null `completedAt` via `completeGroup()`, verifies the precondition, then confirms the group state is preserved (not wiped) after auto-reactivation. This proves `send_message_to_task` does NOT call `resetGroupForRestart`.
   - Rollback test (line 1070): Same group setup, `reviveTaskForMessage` returns false, verifies task rolls back to `cancelled` and group state is untouched.
   - Completed group reset test (line 1326): For `set_task_status`, verifies a completed task with non-null `completedAt` gets its group properly reset via `resetGroupForRestart` -- `completedAt` becomes null, `submittedForReview` becomes false, `feedbackIteration` becomes 0.
   - `task-manager.test.ts` (line 858): Stale description fixed from "worktree is cleaned up on cancel" to "not a valid reactivation path".

5. **All 113 tests pass** in `room-agent-tools.test.ts`.

### P3 -- Asymmetry between agent tools and RPC handler for `completed` in `send_message_to_task`

The RPC handler in `task-handlers.ts` (line 831-834) auto-reactivates `needs_attention`, `completed`, AND `cancelled` tasks when a message is sent. The agent tools version (room-agent-tools.ts line 523) only auto-reactivates `needs_attention` and `cancelled` -- `completed` tasks fall through to `routeHumanMessageToGroup`, which will reject them because the group's `completedAt` is set.

The `human-message-routing.ts` comment (line 11) acknowledges this as a valid design choice ("completed -> caller should use reviveTaskForMessage() **or block with an error**"). The agent tools choose to block for completed tasks while the RPC handler revives -- this may be intentional (agents shouldn't reactivate completed work without explicit status change), but it is worth calling out as a difference between the two paths. If this is deliberate, a brief comment at line 523 explaining why `completed` is excluded would prevent future confusion.

Not blocking because the `human-message-routing.ts` contract explicitly permits either approach.

lsm (commented):

**Recommendation: APPROVE**

---

### Re-review summary

All P0 and P1 issues from the prior review have been correctly addressed. The implementation is now semantically consistent and well-tested.

### What was verified

**P0 resolved — `resetGroupForRestart` removed from `send_message_to_task` cancelled path**

The previous implementation called `resetGroupForRestart` before `reviveTaskForMessage` in the cancelled branch of `send_message_to_task`, which wipes all group metadata contrary to the "keep history" semantics of a message-send action. The current code at lines 515–556 of `room-agent-tools.ts` no longer calls `resetGroupForRestart` at any point in the `send_message_to_task` flow. The comment block at lines 519–522 explicitly documents the asymmetry:

```
set_task_status(cancelled → in_progress) → resetGroupForRestart (clean slate)
send_message_to_task(cancelled task)    → reviveTaskForMessage  (keep history)
```

This now matches the design documented in `task-handlers.ts` at lines 826–830.

**P1 resolved — Unified `needs_attention`/`cancelled` block with correct rollback**

Both statuses now share a single `if` block (line 523). The rollback at line 540 correctly restores `task.status` (not a hardcoded `needs_attention`), which handles the cancelled case properly.

**P1 resolved — Rollback test verifies group state**

The new rollback test at line 1070–1106 of `room-agent-tools.test.ts` sets up a group with non-null `completedAt`, mocks `reviveTaskForMessage` to return `false`, then asserts both `task.status === 'cancelled'` AND `groupAfter!.completedAt !== null`. This correctly confirms that no group wipe occurred during the failed revival.

**P1 resolved — Success test asserts history preservation**

The success test at line 1026–1068 asserts `groupAfter!.completedAt !== null` after a successful auto-reactivation, confirming `reviveTaskForMessage` (not `resetGroupForRestart`) was the path taken.

**Comment in `human-message-routing.ts` updated correctly**

Lines 10–11 now reflect the new routing contract, distinguishing `needs_attention`/`cancelled` (use `reviveTaskForMessage`) from `completed` (use revive or block).

### Remaining observation (P3 — non-blocking)

**Intentional divergence in `set_task_status` for `completed` tasks between agent-tool path and RPC path**

`room-agent-tools.ts` lines 379–413 now includes `completed` in the `resetGroupForRestart` block, meaning `completed → in_progress` via the agent tool does a full group wipe. The RPC handler in `task-handlers.ts` line 453–454 explicitly excludes `completed` from this block, treating it as "lightweight revival (group preserved, no full wipe)."

The test at line 1326–1357 in `room-agent-tools.test.ts` correctly asserts the agent-tool `completed → in_progress` behavior (full reset). The behavior is testable and tested. If the divergence is intentional (agents restarting a completed task get a clean slate, humans get history preserved), a brief comment in `room-agent-tools.ts` near line 383 stating why would prevent future confusion. Not blocking.

All CI checks are passing (Daemon + Shared Unit Tests, Lint/Knip/Format/TypeCheck, all online tests).

PR #610, opened 2026-03-21
fix: sync root repo after PR merge via lifecycle gate

lsm:
Add checkLeaderRootRepoSynced hook that runs git fetch + pull on the
room's root workspace path (not the task's isolated worktree) after a
PR is merged. The hook is integrated into runLeaderCompleteGate so task
completion is blocked until the root repo is up-to-date, ensuring new
worktrees always branch from the latest remote HEAD.

- Add rootWorkspacePath field to LeaderCompleteHookContext (falls back
  to workspacePath for backward compatibility)
- Add checkLeaderRootRepoSynced: detects default branch via symbolic-ref
  with BASE_BRANCHES fallback, fetches, and pulls on rootWorkspacePath
- Skip sync for bypass/research tasks (workerBypassed=true) and
  unapproved tasks; fail open when default branch cannot be determined
- Wire rootWorkspacePath: this.taskGroupManager.workspacePath in
  room-runtime.ts so the hook always uses the room's root repo, not a
  task worktree
- Add 20 new unit tests covering all sync scenarios including cwd
  verification

### DIFF
--- a/packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts
+++ b/packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts
@@ -82,6 +82,12 @@ export interface WorkerExitHookContext {
 
 export interface LeaderCompleteHookContext {
 	workspacePath: string;
+	/**
+	 * The room's root workspace path (the main repository, NOT the task's isolated worktree).
+	 * Used by checkLeaderRootRepoSynced to pull the root repo after PR merge.
+	 * Falls back to workspacePath if not provided.
+	 */
+	rootWorkspacePath?: string;
 	taskType: string;
 	workerRole: string;
 	taskId: string;
@@ -598,6 +604,88 @@ export async function checkLeaderDraftsExist(
 	};
 }
 
+/**
+ * Sync the root repo (Room workspace path) to the latest remote HEAD after a PR merge.
+ * This ensures new worktrees branch from the most recent commit, not a stale one.
+ *
+ * Only runs for approved, non-bypassed tasks (i.e., a PR was actually merged).
+ * Uses rootWorkspacePath (the main repo) rather than workspacePath (the task worktree).
+ */
+export async function checkLeaderRootRepoSynced(
+	ctx: LeaderCompleteHookContext,
+	opts?: HookOptions
+): Promise<HookResult> {
+	// Only sync when a PR was merged (approved task, not a bypass/research-only task)
+	if (!ctx.approved || ctx.workerBypassed) {
+		return { pass: true };
+	}
+
+	const rootPath = ctx.rootWorkspacePath ?? ctx.workspacePath;
+	const run = getRunner(opts);
+
+	// Detect default branch via git symbolic-ref
+	let defaultBranch = '';
+	const { stdout: symref, exitCode: symrefExit } = await run(
+		['git', 'symbolic-ref', 'refs/remotes/origin/HEAD'],
+		rootPath
+	);
+	if (symrefExit === 0 && symref) {
+		defaultBranch = symref.trim().replace('refs/remotes/origin/', '');
+	}
+
+	if (!defaultBranch) {
+		// Fallback: check well-known base branches by trying to resolve them
+		for (const candidate of BASE_BRANCHES) {
+			const { exitCode } = await run(
+				['git', 'rev-parse', '--verify', `origin/${candidate}`],
+				rootPath
+			);
+			if (exitCode === 0) {
+				defaultBranch = candidate;
+				break;
+			}
+		}
+	}
+
+	if (!defaultBranch) {
+		log.warn(`checkLeaderRootRepoSynced: could not determine default branch, skipping root sync`);
+		return { pass: true };
+	}
+
+	// Fetch latest from origin
+	const { exitCode: fetchExit } = await run(['git', 'fetch', 'origin'], rootPath);
+	if (fetchExit !== 0) {
+		log.warn(`checkLeaderRootRepoSynced: git fetch origin failed (exit ${fetchExit})`);
+		return {
+			pass: false,
+			reason: 'Root repo sync failed: git fetch origin returned a non-zero exit code.',
+			bounceMessage:
+				'Could not sync the root repository — `git fetch origin` failed.\n\n' +
+				'Please check network connectivity and ensure the remote is reachable, ' +
+				'then call `complete_task` again.',
+		};
+	}
+
+	// Pull the default branch into the root repo
+	const { exitCode: pullExit } = await run(['git', 'pull', 'origin', defaultBranch], rootPath);
+	if (pullExit !== 0) {
+		log.warn(
+			`checkLeaderRootRepoSynced: git pull origin ${defaultBranch} failed (exit ${pullExit})`
+		);
+		return {
+			pass: false,
+			reason: `Root repo sync failed: git pull origin ${defaultBranch} returned a non-zero exit code.`,
+			bounceMessage:
+				`Could not sync the root repository — \`git pull origin ${defaultBranch}\` failed.\n\n` +
+				'This is required so new worktrees branch from the latest commit.\n' +
+				'Resolve any issues (e.g. merge conflicts in the root repo) and call `complete_task` again.',
+		};
+	}
+
+	log.info(`checkLeaderRootRepoSynced: root repo synced to origin/${defaultBranch}`);
+	return { pass: true };
+}
+
 // --- PR URL Utility ---
 
 /**
@@ -746,6 +834,11 @@ export async function runLeaderCompleteGate(
 	// Fails open when gh is unavailable (e.g. bypass/research-only tasks with no PR).
 	const mergedResult = await checkLeaderPrMerged(ctx, opts);
 	if (!mergedResult.pass) return mergedResult;
+
+	// Sync root repo after PR merge so new worktrees branch from the latest commit.
+	// Only runs for approved, non-bypassed tasks; fails open on missing default branch.
+	const syncResult = await checkLeaderRootRepoSynced(ctx, opts);
+	if (!syncResult.pass) return syncResult;
 
 	// Planning tasks: verify draft tasks were created from the merged plan.
 	if (ctx.taskType === 'planning') {
--- a/packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts
+++ b/packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts
@@ -630,8 +630,11 @@ export async function checkLeaderRootRepoSynced(
 	opts?: HookOptions
 ): Promise<HookResult> {
 	// Only sync when a PR was merged (approved task, not a bypass/research-only task).
-	// approved=false is the pre-approval phase; workerBypassed means no real PR exists.
+	// approved=false/undefined is the pre-approval phase; workerBypassed means no real PR exists.
 	if (!ctx.approved || ctx.workerBypassed) {
+		log.debug(
+			`checkLeaderRootRepoSynced: skipping root sync (approved=${ctx.approved}, workerBypassed=${ctx.workerBypassed})`
+		);
 		return { pass: true };
 	}
 
--- a/packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts
+++ b/packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts
@@ -123,7 +123,11 @@ async function defaultRunCommand(
 	try {
 		const proc = Bun.spawn(args, { cwd, stdout: 'pipe', stderr: 'pipe' });
 		const stdout = await new Response(proc.stdout).text();
+		const stderr = await new Response(proc.stderr).text();
 		const exitCode = await proc.exited;
+		if (exitCode !== 0 && stderr.trim()) {
+			log.debug(`command failed (exit ${exitCode}): ${args.join(' ')}: ${stderr.trim()}`);
+		}
 		return { stdout: stdout.trim(), exitCode };
 	} catch {
 		return { stdout: '', exitCode: 1 };
@@ -609,13 +613,24 @@ export async function checkLeaderDraftsExist(
  * This ensures new worktrees branch from the most recent commit, not a stale one.
  *
  * Only runs for approved, non-bypassed tasks (i.e., a PR was actually merged).
+ * The `approved` guard mirrors the state machine rule in room-runtime.ts that prevents
+ * `complete_task` from being reached without human approval for PR-based roles. It acts
+ * as a proxy for "a real PR merge occurred that requires syncing the root repo". Do NOT
+ * remove it — bypass/research tasks have no PR to sync from.
+ *
  * Uses rootWorkspacePath (the main repo) rather than workspacePath (the task worktree).
+ *
+ * Implementation note: uses `git fetch origin` + `git update-ref` rather than `git pull`
+ * to avoid modifying the working tree or failing due to an unexpected checkout state in
+ * the root repo (e.g., if HEAD is on a feature branch). `git update-ref` directly moves
+ * the local branch ref to match the remote-tracking ref without requiring a checkout.
  */
 export async function checkLeaderRootRepoSynced(
 	ctx: LeaderCompleteHookContext,
 	opts?: HookOptions
 ): Promise<HookResult> {
-	// Only sync when a PR was merged (approved task, not a bypass/research-only task)
+	// Only sync when a PR was merged (approved task, not a bypass/research-only task).
+	// approved=false is the pre-approval phase; workerBypassed means no real PR exists.
 	if (!ctx.approved || ctx.workerBypassed) {
 		return { pass: true };
 	}
@@ -652,7 +667,7 @@ export async function checkLeaderRootRepoSynced(
 		return { pass: true };
 	}
 
-	// Fetch latest from origin
+	// Fetch latest from origin (updates remote-tracking refs like origin/main)
 	const { exitCode: fetchExit } = await run(['git', 'fetch', 'origin'], rootPath);
 	if (fetchExit !== 0) {
 		log.warn(`checkLeaderRootRepoSynced: git fetch origin failed (exit ${fetchExit})`);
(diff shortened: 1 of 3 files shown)


lsm (commented):

The implementation is clean, correct, and well-tested. The fix properly addresses the root cause: after a PR merge, the root repository (not the task worktree) now gets synced via `git fetch origin` + `git pull origin <default-branch>`, ensuring new worktrees branch from the latest remote HEAD.

**What works well:**
- Correct separation of `rootWorkspacePath` (room's main repo) vs `workspacePath` (task's worktree) with proper fallback
- Robust default branch detection: `symbolic-ref` first, then `BASE_BRANCHES` fallback, then fail-open
- Appropriate fail-open on branch detection failure vs fail-closed on actual sync failure
- Skip logic for bypass/research tasks and unapproved tasks is consistent with `checkLeaderPrMerged`
- 20 new tests with thorough coverage including cwd verification (ensuring commands run in the right directory)
- Integration tests with `runLeaderCompleteGate` verify end-to-end behavior

**P3 (nit):**
- `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts` lines 823-828: The JSDoc comment for `runLeaderCompleteGate` lists the gate order as 3 steps but does not mention the new root repo sync gate. Consider updating to include it as step 2.

**Recommendation: APPROVE** — no blocking issues found.

lsm (commented):

---

### Summary

The implementation is correct and well-structured. `checkLeaderRootRepoSynced` is wired into `runLeaderCompleteGate` at the right position (after `checkLeaderPrMerged`, before planning/review checks), the `rootWorkspacePath` fallback is sound, and the 20 new unit tests cover all the important paths including cwd verification. There are a few issues worth flagging before shipping.

---

### P1 — `git pull origin <branch>` can fail when root repo's HEAD is on a different branch

**File:** `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts`, line 670

```ts
const { exitCode: pullExit } = await run(['git', 'pull', 'origin', defaultBranch], rootPath);
```

`git pull origin <branch>` only succeeds without conflict when the local HEAD is already tracking `<branch>`. If the root repo happens to be checked out on a different branch (e.g., a developer was on `feat/something` when the daemon started), git will try to merge `origin/main` into that branch, which will very likely fail or produce an unintended merge commit.

A safer, side-effect-free approach would be:
```bash
git fetch origin
git update-ref refs/heads/<branch> origin/<branch>
```
or simply relying on `git fetch` alone — worktrees branch from local refs, and after `git fetch origin`, `origin/main` is up to date. The subsequent `git worktree add ... <branch>` in `TaskGroupManager.createWorktree` can use `origin/<branch>` directly.

As implemented, a `git pull` failure on an unexpected local HEAD will block `complete_task` for the leader, showing a confusing bounce message about merge conflicts in the root repo when the actual problem is an unrelated local checkout state.

---

### P1 — `rootWorkspacePath` missing from the `submit_for_review` hook context

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 1342–1349

The `runLeaderSubmitGate` hook context constructed at line 1342 does **not** include `rootWorkspacePath`:

```ts
const hookCtx: LeaderCompleteHookContext = {
    workspacePath: group.workspacePath ?? this.taskGroupManager.workspacePath,
    taskType: hookTask.taskType ?? 'coding',
    ...
    // rootWorkspacePath absent
};
```

This is inconsistent with the `complete_task` context at line 1213–1215. `runLeaderSubmitGate` doesn't call `checkLeaderRootRepoSynced`, so this has no runtime impact today, but it leaves the context structurally incomplete. If a future hook is added to `runLeaderSubmitGate` that needs `rootWorkspacePath`, the absence will be a silent bug.

---

### P2 — `checkLeaderRootRepoSynced` skips sync when `approved` is false even when PR was merged

The condition at line 619:
```ts
if (!ctx.approved || ctx.workerBypassed) {
    return { pass: true };
}
```

In the `runLeaderCompleteGate` path, the state machine already enforces that `approved=true` before `complete_task` is reachable for PR-based roles (line 1187–1192 in room-runtime.ts). So in practice `approved` will always be true when sync matters. However, the existing `runLeaderCompleteGate` tests at lines 985–1007 pass `approved: false` while `PR state = MERGED` — in that scenario the root sync is silently skipped even though a real PR merge just happened.

This is not a bug (the state machine prevents it in production), but it means the `approved` guard is doing double duty as a proxy for "a PR was actually merged", which is semantically muddier than checking the PR state directly. A comment explicitly noting this design decision would prevent future reviewers from "fixing" it incorrectly.

---

### P3 — Existing tests in `runLeaderCompleteGate — PR merge validation` don't account for new sync commands

Tests at lines 985 and 997 (`PASSES when PR is MERGED for coder tasks`, `PASSES when PR is MERGED for general tasks`) pass with `approved: false`, so the sync hook is skipped. This is correct behavior, but a reader scanning those tests might not realize the sync is only triggered when `approved=true`. A small comment on those tests to clarify the skip condition would improve readability.

---

### P3 — No stderr capture in `defaultRunCommand`

**File:** `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts`, line 124

```ts
const proc = Bun.spawn(args, { cwd, stdout: 'pipe', stderr: 'pipe' });
const stdout = await new Response(proc.stdout).text();
```

`stderr` is piped but never read. When `git pull` fails, the error message goes to stderr; the bounce message currently only contains the exit code. Capturing and logging `stderr` would make diagnostics significantly easier without changing any behavior or test interfaces (since `HookResult.reason` already exists for the human-readable message).

---

### Positive observations

- The gate ordering (PR merged → sync root → check drafts → check reviews) is correct and safe.
- The `workerBypassed` guard correctly short-circuits the sync for research/bypass tasks.
- The fallback from `rootWorkspacePath` to `workspacePath` is a reasonable backward-compat choice and is well-tested.
- The cwd verification test at line 1682 is exactly the right thing to test — it catches the regression this PR is fixing.
- The `git symbolic-ref` + `BASE_BRANCHES` fallback for branch detection is robust and fails open when neither works.

lsm (commented):

**Recommendation: APPROVE**

This PR correctly addresses the root cause: after a PR merge, the root repo was not being synced, so subsequent worktrees branched from stale commits. The fix adds a `checkLeaderRootRepoSynced` lifecycle gate that runs after the PR merge check in `runLeaderCompleteGate`.

### What was reviewed

1. **`packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts`** — New `checkLeaderRootRepoSynced` function, updated `LeaderCompleteHookContext` type with `rootWorkspacePath`, stderr capture in `defaultRunCommand`, updated `runLeaderCompleteGate` JSDoc and gate ordering.

2. **`packages/daemon/src/lib/room/runtime/room-runtime.ts`** — Two integration points where `rootWorkspacePath: this.taskGroupManager.workspacePath` is correctly added to both the `complete_task` and `submit_for_review` hook contexts.

3. **`packages/daemon/tests/unit/room/lifecycle-hooks.test.ts`** — 15 new tests for `checkLeaderRootRepoSynced` and 4 integration tests for `runLeaderCompleteGate` with root sync.

### Correctness assessment

- The use of `git fetch origin` + `git update-ref refs/heads/<branch> origin/<branch>` is a solid improvement over the original `git pull` approach. It avoids working tree modifications and works regardless of which branch the root repo's HEAD points to.

- The default branch detection logic (symbolic-ref with BASE_BRANCHES fallback) is robust and handles the common cases well.

- The `approved` and `workerBypassed` guards correctly scope the sync to only run when a real PR merge occurred. The JSDoc explaining why these guards exist is clear and helpful.

- The `rootWorkspacePath` fallback to `workspacePath` is safe for backward compatibility.

- The `rootWorkspacePath` is correctly sourced from `this.taskGroupManager.workspacePath`, which is the room's workspace path (not the task worktree), confirmed by tracing through the `RoomRuntime` constructor.

### Test coverage

The tests are thorough and cover:
- Skip conditions (not approved, worker bypassed)
- Happy path (fetch + update-ref succeed)
- Failure modes (fetch fails, update-ref fails)
- Default branch detection fallback
- Correct cwd routing (rootWorkspacePath vs workspacePath)
- Integration with `runLeaderCompleteGate` end-to-end
- Bypass task exclusion from sync

### Minor observations (P3)

1. **Fail-open on missing default branch** (lifecycle-hooks.ts:665-668): When the default branch cannot be determined, the hook passes silently. This is a reasonable design choice (fail open) but means sync is silently skipped for repos with non-standard branch names not in `BASE_BRANCHES`. The `log.warn` is appropriate here.

2. The `mockRunner` in tests ignores the `cwd` parameter by design, but the dedicated cwd-verification tests (lines 1686-1747, 1798-1840) use custom runners that capture and assert `cwd`, which provides proper coverage for the routing behavior.

All 134 tests pass. The implementation is clean, well-documented, and correctly addresses the original bug.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (P1 issue, 2 P2 issues)

---

### Summary

The core implementation is sound: `git fetch origin` + `git update-ref` correctly avoids the working-tree modification and branch-checkout dependency that made `git pull` fragile. The `approved`/`workerBypassed` guard logic is well-reasoned and documented. However, there is one test correctness issue and two logic gaps worth addressing before merging.

---

### P1 — Tests in `runLeaderCompleteGate` block use `approved=undefined`, silently skipping root sync

**File:** `packages/daemon/tests/unit/room/lifecycle-hooks.test.ts`, lines 712–941

The `makeLeaderCtx` helper does not set `approved`, leaving it `undefined`. In `checkLeaderRootRepoSynced`, the guard is `if (!ctx.approved || ctx.workerBypassed)` — `undefined` is falsy, so root sync is **silently skipped** for every test in the `runLeaderCompleteGate` describe block that does not explicitly pass `approved: true`.

In production `runLeaderCompleteGate` is only reachable after the state machine enforces `approved=true`, so all real runs will exercise the sync path. But the tests at lines 712–941 (which use `workerRole: 'coder'` and `hasReviewers: true/false`) pass without any sync commands being exercised. This means the guard that blocks completion until the root repo is synced is effectively invisible in those tests.

The newly-added integration tests in `runLeaderCompleteGate — root repo sync after PR merge` (lines 1750–1870) correctly use `approved: true`, so they do cover the sync path. But the older gate tests remain misleading — they appear to test the full `runLeaderCompleteGate` pipeline for a coder task with a merged PR, but they never trigger the sync step that is the whole point of this PR.

**Fix:** Pass `approved: true` in the `makeLeaderCtx` default (since production always has `approved=true` when `complete_task` is reachable), or add `approved: true` to every existing test in the `runLeaderCompleteGate` describe block where a real merged-PR scenario is being exercised.

---

### P2 — `checkLeaderRootRepoSynced` silently skips when `approved` is absent

**File:** `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts`, line 634

The guard `if (!ctx.approved || ctx.workerBypassed)` treats `approved=undefined` identically to `approved=false` (skip sync). The JSDoc says "runLeaderCompleteGate does not read this field — human approval is enforced by the state machine before this hook is reached." But the hook itself _does_ read it as a proxy for "a real PR merge occurred."

If a future caller passes an incomplete context (without setting `approved`), root sync is silently skipped with no log message. The JSDoc explanation for the guard is excellent but the runtime behavior is a silent no-op rather than a safe fail-open. The fix is trivial: add a `log.debug` when skipping due to `approved` being falsy/undefined to make the skip observable in logs, which helps operators diagnose why sync did not run.

---

### P2 — No integration test for the case where `workerBypassed` is NOT set in `LeaderCompleteHookContext` (relies on `undefined` treating correctly)

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 1222

The `submit_for_review` hook context (lines 1342–1350) does not include `approved` or `workerBypassed`. Since `runLeaderSubmitGate` does not call `checkLeaderRootRepoSynced` today this is safe, but if a future developer adds a sync step to `runLeaderSubmitGate` without noticing the missing fields, the guard will silently skip. This is an architectural lurking hazard. A comment on that context object noting the omission is intentional would prevent future confusion.

---

- **`git update-ref` over `git pull`**: Correct fix. It advances the local branch ref without touching the working tree or requiring a checkout, eliminating the original fragility.
- **Fallback branch detection**: `symbolic-ref refs/remotes/origin/HEAD` → `BASE_BRANCHES` loop is a robust two-tier approach. The `symbolic-ref` command is scoped to `origin/HEAD` so the `.replace('refs/remotes/origin/', '')` is always correct for that output.
- **Fail-open on undiscoverable branch**: If the default branch cannot be determined, the hook logs a warning and passes. This is the right trade-off.
- **Fail-closed on `git fetch` failure**: A network error during fetch causes a bounce message rather than silent pass. This is the correct behavior since a fetch failure means the sync assertion cannot be verified.
- **`rootWorkspacePath` wiring in `room-runtime.ts`**: Both `complete_task` (line 1215) and `submit_for_review` (line 1344) now set `rootWorkspacePath: this.taskGroupManager.workspacePath`, ensuring the hook always uses the room root repo, not the task worktree.
- **JSDoc comments**: The guard explanation in the `checkLeaderRootRepoSynced` JSDoc (lines 615–627) is thorough and accurately describes the design rationale. Same for the `LeaderCompleteHookContext.approved` field comment.
- **CWD verification tests** (lines 1686–1716): Explicitly verify that all git commands run against `rootWorkspacePath`, not `workspacePath`. This is the most important behavioral property of the hook.
- **20 new unit tests**: Good coverage of the happy path, skip conditions, fetch failure, update-ref failure, fallback branch detection, and CWD correctness.

lsm (commented):

**Recommendation: APPROVE**

This is a well-structured fix for root repo sync after PR merge. The third commit cleanly addresses the P1 and P2 issues from the previous review. Here is my assessment:

### What looks good

1. **`makeLeaderCtx` default `approved: true`** -- This is the right call. In production, `complete_task` is only reachable after the state machine enforces human approval, so `approved: true` is the realistic default. The existing `runLeaderCompleteGate` tests that need to exercise the pre-approval path already pass `approved: false` explicitly. The tests that previously had no `approved` field now exercise the sync path, which correctly fails open on branch detection (since `mockRunner` returns `exitCode: 1` for unrecognized commands). The test output confirms this with repeated "could not determine default branch, skipping root sync" messages -- expected and correct.

2. **`log.debug` when skip condition fires** -- The debug log in `checkLeaderRootRepoSynced` at line 635-637 makes the skip observable, which is important for diagnosing future issues. Good that it logs both `approved` and `workerBypassed` values.

3. **Comment on `submit_for_review` hook context** (room-runtime.ts lines 1350-1353) -- Clear explanation of why `approved` and `workerBypassed` are intentionally omitted. This prevents future developers from reflexively adding them "for completeness" without understanding the implications.

4. **`git update-ref` instead of `git pull`** -- Smart choice to avoid working tree modifications or merge conflicts in the root repo.

5. **Test coverage** -- 20 new tests covering: skip conditions (approved=false, workerBypassed), happy path, fetch failure, update-ref failure, fail-open on missing branch, BASE_BRANCHES fallback, cwd verification for rootWorkspacePath vs workspacePath, fallback to workspacePath, and end-to-end integration through `runLeaderCompleteGate`.

### Minor observations (P3)

- The `defaultRunCommand` stderr capture reads both stdout and stderr sequentially. While Bun's `Bun.spawn` with `'pipe'` mode buffers output, reading them sequentially (stdout first, then stderr) is fine for the command sizes involved here. No concern in practice.

- The "could not determine default branch, skipping root sync" messages in test output (from tests that don't mock the branch detection commands) are expected log noise. These correctly exercise the fail-open path and confirm that `approved: true` default is working as intended.

No blocking issues found.

lsm (commented):

**Recommendation: APPROVE**

The three issues from the previous REQUEST_CHANGES review are all properly addressed. Here is my full assessment of the final state.

---

### What was fixed

**P1 — `makeLeaderCtx` default `approved: true`**

The test helper now defaults `approved: true`, matching production behavior (the state machine enforces approval before `complete_task` is reachable). Tests that need pre-approval semantics explicitly pass `approved: false`. This means the existing `runLeaderCompleteGate` tests with `approved: true` now trigger `checkLeaderRootRepoSynced`. Since those tests do not mock the git sync commands, the sync step runs but fails open via the "could not determine default branch" path (mockRunner returns `{ stdout: '', exitCode: 1 }` for unrecognized commands). This is deliberate and correct — it is documented in the test comments added in this PR.

Confirmed by running the test suite: 12 "could not determine default branch, skipping root sync" log lines appear, which are expected noise from the fail-open path.

**P2 — `log.debug` for skip condition**

`checkLeaderRootRepoSynced` now emits a debug log when the skip condition fires (`approved=false/undefined` or `workerBypassed=true`). The exact fields logged make it straightforward to distinguish the two skip reasons in production logs.

**P2 — Comment on `submit_for_review` hook context**

The `LeaderCompleteHookContext` in the `submit_for_review` path now has an explicit comment explaining that `approved` and `workerBypassed` are intentionally omitted (pre-merge context, not a bug or oversight).

---

- `git fetch origin` + `git update-ref refs/heads/<branch> origin/<branch>` is the correct approach. It advances the local branch ref without touching the working tree or requiring a checkout, which is safe for a root repo that may have HEAD on any branch.
- Fail-open when branch detection fails (symbolic-ref + BASE_BRANCHES fallback all fail) is appropriate — sync is best-effort, not a hard requirement.
- Fail-closed when fetch or update-ref fails for approved tasks is appropriate — those are genuine errors worth surfacing to the leader agent.
- `rootWorkspacePath: this.taskGroupManager.workspacePath` is the correct field (room's root workspace, set from `config.workspacePath` at `RoomRuntime` construction time).
- The wiring in both the `complete_task` gate (line 1215) and the `submit_for_review` gate (line 1344, with explanatory comment) is correct.

---

### Test coverage

The 20 new unit tests in `checkLeaderRootRepoSynced` and `runLeaderCompleteGate — root repo sync after PR merge` cover:

- Skip conditions (`approved=false`, `workerBypassed=true`)
- Happy path (symref success, fetch success, update-ref success)
- Failure paths (fetch fails, update-ref fails)
- Fail-open on missing default branch
- BASE_BRANCHES fallback when symref fails
- CWD verification (`rootWorkspacePath` used, not `workspacePath`)
- Fallback to `workspacePath` when `rootWorkspacePath` is omitted
- End-to-end integration through `runLeaderCompleteGate`

Coverage is thorough for the new function.

---

1. `symref.trim()` in `checkLeaderRootRepoSynced` (line 651) is redundant — `defaultRunCommand` already returns `stdout.trim()`. Harmless defensive coding; no change needed.
2. Existing `runLeaderCompleteGate` tests with `approved: true` that don't mock sync commands now emit "could not determine default branch" log warnings. This is expected given the fail-open design, and the test comments explain it. The dedicated sync tests verify the actual sync behavior.

---

The implementation is correct, well-tested, and cleanly addresses the root cause. APPROVE.

PR #203, opened 2026-03-06
plan: adopt persistent job queue across daemon

lsm:
## Summary

This plan document breaks down the goal of wiring up and adopting the existing (but unused) persistent job queue across the NeoKai daemon.

## Background

`JobQueueRepository` and `JobQueueProcessor` are fully implemented with comprehensive unit tests, but `app.ts` never instantiates the processor and no subsystem enqueues jobs. All background work currently uses fire-and-forget Promises, `setInterval`, or in-memory state that is lost on daemon restart.

### Task 1 — Wire Up JobQueueProcessor in DaemonApp (Foundation)
Initialize `JobQueueProcessor` in `app.ts`, define queue name constants, and hook it into the daemon start/stop lifecycle. All subsequent tasks depend on this.

### Task 2 — Migrate Session Background Tasks (depends on Task 1)
Replace the fire-and-forget title generation + git branch rename (currently tracked in a `Set<Promise>`) with a persistent `session.title_generation` job. Removes `pendingBackgroundTasks` from `SessionManager`.

### Task 3 — Migrate GitHub Event Processing (depends on Task 1)
Persist polling state (etags, timestamps) to DB; enqueue a `github.event` job per new event instead of firing directly through the event bus. Adds idempotency so the same event cannot be processed twice across restarts.

### Task 4 — Migrate Room Runtime Tick (depends on Task 1)
Remove the `setInterval` heartbeat from `RoomRuntime`; drive ticks via `room.tick` jobs with deduplication and heartbeat re-scheduling after each tick. Room state machine operations become durable.

### Task 5 — Integration Tests, Cleanup, and E2E Validation (depends on Tasks 2–4)
Online tests for crash-recovery across all three migrated subsystems, an E2E test for UI recovery after restart, removal of stale in-memory patterns, and a scheduled DB cleanup job.

## Dependency Graph

```
Task 1 (Foundation)
  ├── Task 2 (Session tasks)    ──┐
  ├── Task 3 (GitHub events)    ──┼── Task 5 (Integration + Cleanup)
  └── Task 4 (Room runtime)     ──┘
```

PR #612, opened 2026-03-21
docs: implementation plan for persistent job queue adoption

lsm:
## Summary

Implementation plan for completing and adopting the persistent job queue, continuing from the ADR in PR #203.

- 6 milestones, 18 tasks covering the full migration
- Foundation: wire JobQueueProcessor into DaemonApp, extend listJobs for status arrays
- Migrate session background tasks (pendingBackgroundTasks -> job queue)
- Migrate GitHub polling (setInterval -> self-scheduling jobs)
- Migrate room runtime tick (setInterval + scheduleTick -> job queue)
- Add cleanup job and stale reclamation verification
- Integration tests and E2E validation

## Test plan

- [ ] Review plan files for completeness and accuracy
- [ ] Verify task dependencies are correctly ordered
- [ ] Confirm all affected files are identified

### DIFF
--- a/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
+++ b/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
@@ -113,4 +113,4 @@ Milestones 2, 3, 4, and 5 can proceed in parallel after Milestone 1 is complete.
 
 ## Total Estimated Task Count
 
-19 tasks across 6 milestones.
+18 tasks across 6 milestones (3 + 3 + 3 + 4 + 2 + 3).
--- a/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
+++ b/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
@@ -2,7 +2,7 @@
 
 ## Goal
 
-Wire the existing but unused `JobQueueRepository` and `JobQueueProcessor` into the daemon lifecycle, then migrate all `setInterval`-based background tasks to the persistent job queue. This makes background work durable across daemon restarts and centralizes scheduling.
+Wire the existing but unused `JobQueueRepository` and `JobQueueProcessor` into the daemon lifecycle, then migrate the three primary `setInterval`-based background task subsystems to the persistent job queue. This makes background work durable across daemon restarts and centralizes scheduling.
 
 ## Background
 
@@ -16,7 +16,19 @@ PR #203 produced a comprehensive ADR (`docs/adr/0002-job-queue-migration.md`) ov
 The subsystems to migrate:
 1. **Session background tasks** -- fire-and-forget title generation tracked in `pendingBackgroundTasks` Set
 2. **GitHub polling** -- `setInterval` in `GitHubPollingService` (60s cycle)
-3. **Room runtime tick** -- `setInterval` in `RoomRuntime.start()` (30s cycle, 17 `scheduleTick()` call sites)
+3. **Room runtime tick** -- `setInterval` in `RoomRuntime.start()` (30s cycle), with **17 `scheduleTick()` call sites**, **6 `scheduleTickAfterRateLimitReset()` call sites** (lines 617, 696, 732, 985, 1052, 1084), and **2 `queueMicrotask(() => this.tick())` call sites** (lines 2241, 3421) — totaling ~23 scheduling points that need migration
+
+## Out of Scope
+
+The following `setInterval` users are explicitly **out of scope** for this migration and will remain as-is:
+
+- **`WebSocketServerTransport.staleCheckTimer`** (`packages/daemon/src/lib/websocket-server-transport.ts` line 87) — This is a transport-layer health check (30s stale connection cleanup), not a background business task. It operates at a different abstraction layer and does not benefit from persistence.
+- **`SpaceRuntime.tickTimer`** (`packages/daemon/src/lib/space/runtime/space-runtime.ts` line 226) — SpaceRuntime is a newer/parallel runtime implementation. Its tick loop follows the same pattern as RoomRuntime but is a separate subsystem. It should be migrated in a follow-up once RoomRuntime migration is proven stable.
+- **`TaskAgentManager` polling interval** (`packages/daemon/src/lib/space/runtime/task-agent-manager.ts` line 190) — Part of the SpaceRuntime subsystem, out of scope for the same reason.
+- **`JobQueueProcessor` internal poll timer** (`packages/daemon/src/storage/job-queue-processor.ts` line 38) — This is the job queue infrastructure itself; its internal `setInterval` is the mechanism that drives all job processing and is not a migration target.
+- **`app.ts` shutdown readiness check** (`packages/daemon/src/app.ts` line 367) — One-shot `setInterval` for graceful shutdown polling, not a recurring background task.
+
+Task 6.2 must explicitly verify these exclusions and document them in a code comment for future reference.
 
 ## Approach
 
@@ -25,22 +37,65 @@ Continue on the existing PR #203 branch (`plan/persistent-job-queue-adoption`). 
 **Key design decisions:**
 - Handlers self-schedule the next run (no dedicated scheduler service for now)
 - `listJobs` must be extended to accept `status` as an array for deduplication queries
-- Room tick deduplication uses application-level checks (not DB unique index) for simplicity
+- Room tick deduplication uses application-level checks (not DB unique index) for simplicity. **Risk acknowledged:** the check-then-enqueue pattern has a small race window. If this proves problematic in practice, a follow-up can add a unique partial index on `(queue, json_extract(payload, '$.roomId')) WHERE status IN ('pending', 'processing')` as the ADR suggests. For now the application-level approach is acceptable because: (a) SQLite serializes writes, limiting the race window; (b) a rare duplicate tick is harmless (tick is idempotent); (c) the dedup check should use "no pending job exists at all for this roomId" (not just "no future-scheduled job"), see detailed logic in Milestone 4.
 - `stopRuntime()` must call `runtimes.delete(roomId)` to fix the liveness check bug
+- **Pause/resume must cancel in-flight tick jobs** — see Milestone 4 for detailed design of pause/resume interaction with the job queue
+
+## Handler Registration and Startup Ordering
+
+All handlers **must** be registered before `jobProcessor.start()` is called. If a pending job exists in the DB from a previous run and no handler is registered for its queue, it will fail with "No handler registered" and consume retry attempts.
+
+**Required startup order in `app.ts`:**
+
+```
+1. sessionManager.start()       → registers 'session.title_generation' handler
+2. gitHubService.start()        → registers 'github.poll' handler + enqueues initial poll
+3. roomRuntimeService.start()   → registers 'room.tick' handler
+4. Register cleanup handler     → registers 'job_queue.cleanup' handler + enqueues initial cleanup
+5. jobProcessor.start()         → begins processing (all handlers now registered)
+```
+
+Each milestone's wiring task must respect this ordering. Task 1.2 must initially call `jobProcessor.start()` as the **last** step, and each subsequent milestone inserts its handler registration before that call.
+
+## maxConcurrent Configuration
+
+The initial `maxConcurrent: 3` is conservative. With multiple active rooms, each generating `room.tick` jobs every 30s, plus GitHub polling and title generation, the 3-slot limit may cause contention — a long-running title generation or poll could block room ticks.
+
+**Mitigation strategy:**
+- Start with `maxConcurrent: 5` instead of 3, based on expected workload (typically 1-3 active rooms + 1 GitHub poll + occasional title gen)
+- The processor's `dequeue` already handles priority via FIFO ordering; room ticks are short-lived and will cycle quickly
+- If contention appears in practice, the follow-up is per-queue concurrency limits (the ADR's open question on priority inversion)
+- Task 1.2 should make `maxConcurrent` configurable via environment variable (`NEOKAI_JOB_QUEUE_MAX_CONCURRENT`, default 5)
+
+## Stale Job Reclamation Timing
+
+`JobQueueProcessor.checkStaleJobs()` runs only after `STALE_CHECK_INTERVAL = 60_000ms` has elapsed since the last check. After a daemon restart, jobs stuck in `processing` from the previous run will sit unreclaimed for up to 60 seconds.
+
+**Mitigation:** Task 1.2 must add an eager `reclaimStale()` call in `JobQueueProcessor.start()` — immediately after starting, reclaim any stale jobs before the first poll tick. This ensures crash-recovery is instant rather than delayed by up to 60 seconds.
+
+## Rollback Strategy
+
+Each milestone removes old scheduling code (`setInterval`, `pendingBackgroundTasks`, etc.) and replaces it with job queue calls. If a milestone causes issues:
+
+1. **Revert the merge commit** for that milestone's PR — this restores the old `setInterval` code
+2. Old jobs in the `job_queue` table for the reverted subsystem will go stale and be cleaned up by the cleanup job (or manually via `DELETE FROM job_queue WHERE queue = '...'`)
+3. The job queue infrastructure (Milestone 1) can remain in place even if individual migrations are reverted
+
+For the highest-risk migration (Milestone 4 — room tick), the implementer should ensure the PR is small enough to revert cleanly and should include a feature flag or environment variable (`NEOKAI_USE_JOB_QUEUE_ROOM_TICK=true`) that can disable the migration without a full revert.
 
 ## Milestones
 
-1. **Foundation -- Wire JobQueueProcessor into DaemonApp** -- Instantiate processor in `app.ts`, add constants file, extend `listJobs` to accept status arrays, connect change notifier to ReactiveDatabase, hook into start/stop lifecycle.
+1. **Foundation -- Wire JobQueueProcessor into DaemonApp** -- Instantiate processor in `app.ts`, add constants file, extend `listJobs` to accept status arrays, connect change notifier to ReactiveDatabase, hook into start/stop lifecycle. Add eager stale reclamation on startup.
 
 2. **Migrate Session Background Tasks** -- Replace `pendingBackgroundTasks` Set in `SessionManager` with a `session.title_generation` job queue handler. Add `SessionManager.start()` method.
 
 3. **Migrate GitHub Polling** -- Replace `setInterval` in `GitHubPollingService` with self-scheduling `github.poll` jobs. Expose `triggerPoll()` method, register handler, add dedup logic in finally block.
 
-4. **Migrate Room Runtime Tick** -- Replace `setInterval` + `scheduleTick()` + `queueMicrotask` in `RoomRuntime` with `room.tick` jobs. Fix `stopRuntime()` to delete from runtimes map. Handle 17 `scheduleTick()` call sites.
+4. **Migrate Room Runtime Tick** -- Replace `setInterval` + `scheduleTick()` + `scheduleTickAfterRateLimitReset()` + `queueMicrotask` in `RoomRuntime` with `room.tick` jobs. Fix `stopRuntime()` to delete from runtimes map. Handle all ~23 scheduling call sites. Add pause/resume-aware job cancellation.
 
 5. **Database Cleanup Job and Stale Job Reclamation** -- Add a self-scheduling `job_queue.cleanup` job that runs daily. Remove old completed/dead jobs. Verify stale job reclamation works across restarts.
 
-6. **Integration Tests and E2E Validation** -- Online tests for crash-recovery scenarios across all migrated subsystems. Verify job processing resumes after daemon restart.
+6. **Integration Tests and E2E Validation** -- Online tests for crash-recovery scenarios across all migrated subsystems. Verify job processing resumes after daemon restart. Audit for remaining `setInterval` usage and document out-of-scope exclusions.
 
 ## Cross-Milestone Dependencies
 
@@ -58,4 +113,4 @@ Milestones 2, 3, 4, and 5 can proceed in parallel after Milestone 1 is complete.
 
 ## Total Estimated Task Count
 
-18 tasks across 6 milestones.
+19 tasks across 6 milestones.
--- a/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
+++ b/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
@@ -1,0 +1,61 @@
+# Complete and Adopt Persistent Job Queue -- Continue from PR #203
+
+## Goal
+
+Wire the existing but unused `JobQueueRepository` and `JobQueueProcessor` into the daemon lifecycle, then migrate all `setInterval`-based background tasks to the persistent job queue. This makes background work durable across daemon restarts and centralizes scheduling.
+
+## Background
+
+PR #203 produced a comprehensive ADR (`docs/adr/0002-job-queue-migration.md`) over 13 review iterations. The plan is well-specified but zero production code has been implemented. The existing infrastructure includes:
+
+- `JobQueueRepository` (complete, 203 lines) -- CRUD, dequeue, retry, cleanup
+- `JobQueueProcessor` (complete, 123 lines) -- poll-based processor with handler registration
+- `job_queue` SQLite table with indexes
+- Comprehensive unit tests for both
+
+The subsystems to migrate:
+1. **Session background tasks** -- fire-and-forget title generation tracked in `pendingBackgroundTasks` Set
+2. **GitHub polling** -- `setInterval` in `GitHubPollingService` (60s cycle)
+3. **Room runtime tick** -- `setInterval` in `RoomRuntime.start()` (30s cycle, 17 `scheduleTick()` call sites)
+
+## Approach
+
+Continue on the existing PR #203 branch (`plan/persistent-job-queue-adoption`). Implement in 6 milestones, each producing a PR-ready commit. All work targets the `dev` branch.
+
+**Key design decisions:**
+- Handlers self-schedule the next run (no dedicated scheduler service for now)
+- `listJobs` must be extended to accept `status` as an array for deduplication queries
+- Room tick deduplication uses application-level checks (not DB unique index) for simplicity
+- `stopRuntime()` must call `runtimes.delete(roomId)` to fix the liveness check bug
+
+## Milestones
+
+1. **Foundation -- Wire JobQueueProcessor into DaemonApp** -- Instantiate processor in `app.ts`, add constants file, extend `listJobs` to accept status arrays, connect change notifier to ReactiveDatabase, hook into start/stop lifecycle.
+
+2. **Migrate Session Background Tasks** -- Replace `pendingBackgroundTasks` Set in `SessionManager` with a `session.title_generation` job queue handler. Add `SessionManager.start()` method.
+
+3. **Migrate GitHub Polling** -- Replace `setInterval` in `GitHubPollingService` with self-scheduling `github.poll` jobs. Expose `triggerPoll()` method, register handler, add dedup logic in finally block.
+
+4. **Migrate Room Runtime Tick** -- Replace `setInterval` + `scheduleTick()` + `queueMicrotask` in `RoomRuntime` with `room.tick` jobs. Fix `stopRuntime()` to delete from runtimes map. Handle 17 `scheduleTick()` call sites.
+
+5. **Database Cleanup Job and Stale Job Reclamation** -- Add a self-scheduling `job_queue.cleanup` job that runs daily. Remove old completed/dead jobs. Verify stale job reclamation works across restarts.
+
+6. **Integration Tests and E2E Validation** -- Online tests for crash-recovery scenarios across all migrated subsystems. Verify job processing resumes after daemon restart.
+
+## Cross-Milestone Dependencies
+
+```
+Milestone 1 (Foundation)
+  +-- Milestone 2 (Session tasks)
+  +-- Milestone 3 (GitHub polling)
+  +-- Milestone 4 (Room tick)
+  +-- Milestone 5 (Cleanup job)
+       |
+       +-- Milestone 6 (Integration tests) -- depends on Milestones 2, 3, 4, 5
+```
+
+Milestones 2, 3, 4, and 5 can proceed in parallel after Milestone 1 is complete. Milestone 6 depends on all prior milestones.
+
+## Total Estimated Task Count
+
+18 tasks across 6 milestones.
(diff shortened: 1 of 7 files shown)


lsm (commented):

This is a well-structured plan that correctly identifies the three subsystems to migrate, establishes a sensible milestone ordering, and provides detailed acceptance criteria for each task. The plan is grounded in the actual codebase — line numbers, file paths, and existing infrastructure are accurately referenced. However, there are several gaps and risks that should be addressed before implementation begins.

---

### P1: SpaceRuntime setInterval not addressed

The plan identifies three subsystems to migrate: session background tasks, GitHub polling, and room runtime tick. However, `SpaceRuntime` (`packages/daemon/src/lib/space/runtime/space-runtime.ts`) also uses `setInterval` for its tick loop (line 226) and has a `tickTimer` field (line 139). The `TaskAgentManager` (`packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 190) also uses `setInterval` for polling.

The plan should explicitly state whether `SpaceRuntime` is in-scope or out-of-scope, and if out-of-scope, why. If `SpaceRuntime` is the newer replacement for `RoomRuntime`, or vice versa, or they coexist, the plan should clarify the relationship. Without this, the goal of "migrate all setInterval-based background tasks to the persistent job queue" stated in the overview is incomplete.

### P1: scheduleTickAfterRateLimitReset uses setTimeout, not just scheduleTick — plan underspecifies the migration

The plan mentions `scheduleTickAfterRateLimitReset` but Task 4.2 treats it superficially: "instead of `setTimeout(() => this.scheduleTick(), delayMs)`, call `enqueueRoomTick(this.roomId, this.jobQueue, delayMs)`". There are actually **6 call sites** for `scheduleTickAfterRateLimitReset` (lines 617, 696, 732, 985, 1052, 1084), and the method itself contains important semantic behavior — it reads the rate limit remaining time from the group repo and adds a 5-second buffer. The plan's Task 4.2 line number list only partially accounts for these. The task should explicitly enumerate all `scheduleTickAfterRateLimitReset` call sites alongside the 17 `scheduleTick` call sites, for a total of ~23 sites that need attention.

### P2: Deduplication via listJobs is inefficient and racy

The plan's dedup strategy for room ticks (Task 4.1) and GitHub polling (Task 3.1) relies on calling `jobQueue.listJobs({ queue, status: ['pending', 'processing'], limit: 1000 })` and then filtering by payload in application code. This has two problems:

1. **Performance**: Scanning up to 1000 jobs and filtering in JS for every tick enqueue (every 30s per room, potentially dozens of rooms) adds unnecessary overhead.
2. **Race condition**: Between the listJobs check and the enqueue call, another tick could enqueue a duplicate. SQLite serializes writes, but the check-then-act pattern is still vulnerable if the processor dequeues between the check and the enqueue.

The ADR itself identifies a better approach — a unique partial index on `(queue, json_extract(payload, '$.roomId'))` WHERE `status IN ('pending', 'processing')`. The plan should at least acknowledge this risk and provide a mitigation path if the application-level dedup proves insufficient.

### P2: maxConcurrent=3 may cause room tick contention

Task 1.2 sets `maxConcurrent: 3` globally. With multiple rooms active, a single GitHub poll or long-running title generation could consume a slot, leaving only 2 slots for potentially many room ticks. The plan should discuss whether `maxConcurrent` needs to be higher, or whether per-queue concurrency limits should be considered. The ADR raised priority inversion as an open question but the plan doesn't address it.

### P2: Task 4.2 line numbers may drift — needs a more robust approach

Task 4.2 lists 17 specific line numbers for `scheduleTick()` call sites. By the time implementation reaches Milestone 4 (after Milestones 1-3 are merged), these line numbers will have shifted. The task should instead instruct the implementer to grep for `this.scheduleTick()` and `this.scheduleTickAfterRateLimitReset(` to find all call sites dynamically.

### P2: No rollback strategy documented in the plan

The ADR has a clear rollback section per phase. The plan milestones do not include any rollback guidance. For a high-risk migration like room tick (Milestone 4), the plan should document how to revert if the migration causes issues in production — especially since the old `setInterval` code will be deleted.

### P3: Task 6.1 crash recovery test feasibility

Task 6.1 describes testing crash recovery by stopping a daemon and starting a fresh one against the same database. This is a sound approach, but the plan should note that `createDaemonServer()` typically creates a fresh in-memory or temporary database. The test will need to explicitly configure a file-backed SQLite database that persists across daemon instances. This is an implementation detail but worth calling out since it's non-obvious.

### P3: The plan says "17 scheduleTick call sites" but the actual count includes calls within scheduleTickAfterRateLimitReset itself

Lines 3432 and 3451 are `scheduleTick()` calls within `scheduleTickAfterRateLimitReset()`. When that method is replaced, those 2 call sites disappear automatically. The effective number of call sites to manually update is 15 (for `scheduleTick`) + 6 (for `scheduleTickAfterRateLimitReset`) = 21 distinct replacement points. The plan should clarify this to avoid confusion.

### P3: WebSocket stale check setInterval not mentioned

The ADR mentions `WebSocket Stale Check` in `websocket-server-transport.ts` as another `setInterval` user. The plan doesn't address it or explicitly exclude it. A one-line note explaining why it's out of scope would help completeness.

---

### Summary

The plan is solid in structure and covers the core migration path well. The main gaps are: (1) SpaceRuntime is not addressed despite also using setInterval-based ticking, (2) the scheduleTickAfterRateLimitReset migration is underspecified, and (3) the dedup strategy has known efficiency and race-condition risks that should be acknowledged. Addressing these would make the plan ready for implementation.

lsm (commented):

**Recommendation: REQUEST CHANGES** — The plan is well-structured and technically detailed, but several P1 and P2 issues need to be resolved before implementation.

---

## Summary

The plan covers all three stated migration targets (session title gen, GitHub polling, room runtime tick) with a clean dependency graph, per-task acceptance criteria, and an online test per milestone. The ADR grounding is solid. However, there are a handful of inaccuracies and gaps that will cause real friction or missed work during implementation.

---

## P1: scheduleTick call-site count is understated

The plan (Task 4.2) says "17 scheduleTick() call sites" — but `scheduleTick()` is called 17 times **and** `scheduleTickAfterRateLimitReset()` is called 7 more times (lines 617, 696, 732, 985, 1052, 1084, within the method itself), plus there are 2 bare `queueMicrotask(() => this.tick())` calls at lines 2241 and 3421. All 27 of these are semantically part of the same scheduling surface and need migration. The plan treats `scheduleTickAfterRateLimitReset` as a separate method update but does not enumerate its internal `this.scheduleTick()` calls. If the implementer only replaces the 17 plain sites and misses the rate-limit path, the setInterval removal in Task 4.2 will leave ticks silently dead after rate-limiting events — a correctness regression that is hard to detect without load.

**Files:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`

---

## P1: RoomRuntime pause/resume loses all in-flight tick scheduling

After the migration, `pause()` only sets `this.state = 'paused'` but there will be no mechanism to cancel or drain the already-enqueued `room.tick` jobs in the persistent queue. When `resume()` is called, `enqueueRoomTick` will re-enqueue but the previously-queued tick (which survives in the DB) will also execute — breaking the dedup invariant within a short window. The plan mentions dedup in `enqueueRoomTick` but only guards against jobs with `runAt > now`. A pending job with `runAt <= now` will not be filtered, so the resume path can generate a double tick. The current in-memory `tickLocked / tickQueued` mutex entirely disappears with no replacement discussed.

**Files:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` lines 154-156

---

## P1: Task 1.3 tests duplicate work already requested in Task 1.2

Task 1.2 subtask 8 already asks for a unit test file `app/job-queue-lifecycle.test.ts` covering processor lifecycle. Task 1.3 then separately requests `storage/job-queue-processor-lifecycle.test.ts` covering the same start/stop/change-notification/error/stale behaviors. This overlap will cause confusion during implementation — the implementer will either write redundant tests or skip one. The plan should clearly split the two: 1.2's test should be purely about app-level wiring (context fields, cleanup ordering); 1.3 should be purely about the `JobQueueProcessor` contract in isolation. As written, the acceptance criteria overlap.

**Files:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/01-foundation.md`

---

## P2: GitHub polling dedup query scans all statuses, is not filtered-by-queue

Task 3.1 proposes:
```
jobQueue.listJobs({ queue: QUEUES.GITHUB_POLL, status: ['pending', 'processing'], limit: 1000 })
```
The logic then filters client-side for `runAt > now`. With `limit: 1000` and a busy queue this is fine in practice, but the dedup check uses `runAt > now` to decide whether to skip scheduling. However, a `pending` job whose `runAt` is in the past (i.e., ready to run right now) is NOT a duplicate of a future-scheduled next poll — the handler has already been triggered. Comparing against `now` at the moment the handler is in its `finally` block is correct only if `runAt` represents the intended-future schedule. If the processor is slow or the job was reclaimed, a pending job with `runAt = pastTime` will be counted as a non-future job and a new one will be enqueued, violating dedup. The condition needs to be `no pending job exists at all` (not just `no future-scheduled pending job`).

**Files:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/03-github-polling.md` Task 3.1

---

## P2: WebSocket stale check and space-runtime tick are out of scope but unmentioned

The audit in Task 6.2 only checks the three migrated subsystems. However, the actual codebase has two additional `setInterval` users that the plan does not address or explicitly exclude:
- `packages/daemon/src/lib/websocket-server-transport.ts` — `staleCheckTimer` (30s connection stale check)
- `packages/daemon/src/lib/space/runtime/space-runtime.ts` — `tickTimer` (also a `setInterval`-based tick loop)

The plan's background section only mentions session tasks, GitHub polling, and room tick — which matches the ADR scope. But the final cleanup step should either explicitly note these as out-of-scope for this PR or mark them for a follow-up task. Without this, Task 6.2's acceptance criterion ("No setInterval for background tasks in migrated subsystems") will pass silently even though two other `setInterval` loops remain.

**Files:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/06-integration-tests.md`, `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md`

---

## P2: Stale reclamation happens only after 60s STALE_CHECK_INTERVAL — not on startup

Task 5.2 describes verifying stale reclamation on restart but does not account for `JobQueueProcessor`'s internal `STALE_CHECK_INTERVAL = 60_000` (line 21 of `job-queue-processor.ts`). `checkStaleJobs()` is a no-op until 60 seconds have elapsed since the last check. This means that after a restart, a job stuck in `processing` will sit unreclaimed for up to 60 seconds, during which the room tick or GitHub poll chain will appear broken. The plan should explicitly note this latency and either lower `STALE_CHECK_INTERVAL` for tests, call `reclaimStale` eagerly on startup before the first poll tick, or adjust the acceptance criteria to reflect the 60-second window.

**Files:** `packages/daemon/src/storage/job-queue-processor.ts` line 21, `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/05-cleanup-job.md`

---

## P2: Task 2.2 says jobProcessor.start() should come AFTER sessionManager.start() — but the plan's startup ordering guidance conflicts

Task 2.2 subtask 2 says: "Call `sessionManager.start()` after sessionManager creation but before `jobProcessor.start()`". This means the handler is registered before the processor starts — good. But Task 1.2 subtask 3 says to call `jobProcessor.start()` before `gitHubService.start()` (line 341 of app.ts). There is no mention of where `sessionManager.start()` slots relative to the GitHub and RoomRuntime handler registrations (Tasks 3.2 and 4.3). If any handler is not registered before `jobProcessor.start()`, a job of that type that is already pending in the DB from a previous run will be dequeued and fail with "No handler registered" — causing it to be retried or go dead. The plan needs an explicit startup ordering table covering all four handler registrations relative to `jobProcessor.start()`.

**Files:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/01-foundation.md`, `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/02-session-background-tasks.md`

---

## P3: Task 4.2 line numbers will shift during implementation

Task 4.2 hardcodes line numbers for 17 `scheduleTick()` call sites (lines 259, 423, 432, 553, 560, 665, 1024, 1149, 1287, 1315, 1588, 1700, 1815, 3364, 3410, 3432, 3451). After Task 4.1 adds new imports and the handler file, and after Tasks 3.2 and 2.2 modify `app.ts` / SessionManager, the room-runtime line numbers will be stale before the implementer gets to 4.2. This is a minor coordination hazard; the agent should be instructed to search for the pattern rather than rely on these line numbers.

**Files:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/04-room-runtime-tick.md`

---

## P3: Milestone 3 online test requires GitHub config but the dev-proxy setup is not addressed

Task 3.3 says "use `createDaemonServer()` with GitHub config enabled" and run with `NEOKAI_USE_DEV_PROXY=1`. GitHub polling makes real HTTP calls to `api.github.com` — the dev proxy intercepts Anthropic API calls, not GitHub REST calls. The test will either need a separate GitHub API mock or a real token. The acceptance criteria say "test runs with dev proxy" but that is not sufficient for GitHub polling. This needs clarification before implementation begins.

**Files:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/03-github-polling.md` Task 3.3

lsm (commented):

**Recommendation: APPROVE** — All P1 and P2 issues from the previous review cycle have been addressed. The plan is now ready for implementation.

---

## Resolution Assessment

This is a re-review focusing on whether the prior P1 and P2 issues were adequately addressed.

### Previously-P1: SpaceRuntime setInterval not addressed — RESOLVED

`00-overview.md` lines 23-29 now contains an explicit Out of Scope section that lists `SpaceRuntime.tickTimer`, `TaskAgentManager` polling interval, `WebSocketServerTransport.staleCheckTimer`, the `JobQueueProcessor` internal poll timer, and the `app.ts` shutdown readiness check — each with a concrete justification. Task 6.2 is also updated to audit these and add a code comment in `app.ts` for future reference. This is fully resolved.

### Previously-P1: scheduleTickAfterRateLimitReset underspecified — RESOLVED

`00-overview.md` lines 18-19 now explicitly calls out all three groups: 17 `scheduleTick()` sites, 6 `scheduleTickAfterRateLimitReset()` sites with their line numbers, and 2 `queueMicrotask` sites. Milestone 4's scheduling inventory section (lines 31-48 of `04-room-runtime-tick.md`) is exhaustive. Lines 3432 and 3451 are correctly identified as internal to `scheduleTickAfterRateLimitReset` and will disappear when that method is rewritten. The total 23-point count is now accurate.

### Previously-P1/P2: Pause/resume dedup breakage — RESOLVED

`04-room-runtime-tick.md` now has a full Pause/Resume and Dedup Design section. The solution is correct and non-racy: (1) handler skips and does not re-schedule when state is not 'running'; (2) `pause()` explicitly calls `cancelPendingTickJobs()` which deletes all pending room.tick jobs; (3) `resume()` enqueues a fresh tick. The `cancelPendingTickJobs` helper is specified in full. This preserves the invariant that a paused room has zero pending ticks.

### Previously-P1: Task 1.2/1.3 test overlap — RESOLVED

`01-foundation.md` Task 1.3 now has a clear Scope clarification note distinguishing it from Task 1.2: 1.2 tests app-level wiring (context fields, cleanup ordering, env var config), 1.3 tests processor contract in isolation (polling, dequeue, dispatch, stale reclamation, error/retry). The acceptance criterion "No overlap with Task 1.2's app-wiring tests" makes this binding.

### Previously-P2: Dedup strategy race condition — RESOLVED

`00-overview.md` lines 40 acknowledges the race window explicitly and states the correct dedup condition: "no pending job exists at all for this roomId" (not just "no future-scheduled job"). Both `enqueueRoomTick` and the GitHub poll handler use this corrected condition. The rationale for deferring the DB unique index is documented: SQLite serializes writes, a duplicate tick is harmless (idempotent), and a follow-up index migration is noted if problems emerge.

### Previously-P2: maxConcurrent=3 contention — RESOLVED

`00-overview.md` lines 62-68 upgrades the default to `maxConcurrent: 5` with justification. It is also made configurable via `NEOKAI_JOB_QUEUE_MAX_CONCURRENT` env var, specified in Task 1.2 subtask 3.

### Previously-P2: Stale reclamation timing — RESOLVED

`00-overview.md` lines 72-74 documents the 60s STALE_CHECK_INTERVAL problem and mandates an eager `reclaimStale()` call in `JobQueueProcessor.start()` before the first poll tick. Task 1.2 subtask 3 specifies the implementation. Task 5.2 now explicitly tests the eager reclamation path.

### Previously-P2: Handler registration ordering — RESOLVED

`00-overview.md` lines 44-58 provides the exact 5-step startup ordering table with handler registration before `jobProcessor.start()`. The ordering is: sessionManager.start() → gitHubService.start() → roomRuntimeService.start() → register cleanup handler → jobProcessor.start(). Each milestone's wiring task explicitly references this ordering.

---

## Remaining Issues

### P1: Cleanup handler registered AFTER jobProcessor.start() — startup ordering violation

`05-cleanup-job.md` Task 5.1 subtask 3 says: "After `jobProcessor.start()`, register the cleanup handler." This directly contradicts the startup ordering table in `00-overview.md` step 4, which places cleanup handler registration at step 4, BEFORE `jobProcessor.start()` at step 5. If a `job_queue.cleanup` job is already pending in the DB from a previous run and the processor starts before the handler is registered, the processor will call `repo.fail()` on that job (see `job-queue-processor.ts` line 91), consuming a retry attempt. After `maxRetries` exhausted the job goes dead and cleanup never runs again until manually re-enqueued. This is a correctness bug in the plan. Task 5.1 subtask 3 must say "Before `jobProcessor.start()`", not "After".

**File:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/05-cleanup-job.md` line 28

### P2: enqueueRoomTick dedup checks only 'pending' but misses 'processing' case after handler-skip

The dedup helper in Task 4.1 checks `status: ['pending']` only (not `['pending', 'processing']`). The comment in the plan (04-room-runtime-tick.md line 65) says "Jobs currently in 'processing' status are fine — they're already executing and will self-schedule on completion." This reasoning holds in the normal case, but there is one edge: when a 'processing' tick job completes with the handler skipping (state is 'paused'), the finally block correctly does NOT re-schedule. This means a processing job can complete without scheduling a successor. If `pause()` is called between the dequeue and the start of handler execution, `cancelPendingTickJobs` in `pause()` will find no 'pending' jobs (the job is already 'processing') and do nothing. The job will then execute, detect 'paused', and return without re-scheduling — leaving the room with zero ticks, which is correct. So actually the 'pending'-only check is fine for the normal case. However, the reasoning in the plan comment should be clearer: the invariant is maintained because processing jobs always re-check state in their finally block. The current comment is not wrong but could mislead an implementer.

**File:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/04-room-runtime-tick.md` Pause/Resume Design section

### P2: Task 3.2 says "stop() becomes a state flag setter" but pollingService.stop() calls clearInterval — instruction conflicts with code

Task 3.2 subtask 1 says: "Remove `clearInterval` from `stop()` — `stop()` becomes a state flag setter." But the actual `polling-service.ts` line 88 currently calls `clearInterval(this.pollingInterval)`. The plan is correct that the clearInterval should be removed post-migration, but the instruction is terse. The implementer may miss that they also need to remove the `pollingInterval` field declaration and update the type. This is a completeness gap — the plan should explicitly say "remove the `pollingInterval` field and its type declaration" alongside removing clearInterval.

**File:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/03-github-polling.md` Task 3.2 subtask 1

### P3: Task 4.1 dedup comment says 'limit: 100' but 00-overview.md says 'limit: 1000' for the listJobs call

Minor inconsistency: `04-room-runtime-tick.md` Task 4.1 uses `limit: 100` for the dedup query, but `00-overview.md` line 40 mentions "limit: 1000" in the context of describing the prior approach. This is not a bug since 100 is better (and the overview was describing the old approach), but the inconsistency could cause an implementer to pick the wrong value. A note clarifying why 100 is used (single room tick dedup, not a global scan) would be helpful.

**File:** `docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/04-room-runtime-tick.md`

---

## Summary

The plan has been substantially improved. Six of seven prior issues are fully resolved with concrete design decisions. The one remaining P1 is a direct contradiction within the plan itself: Task 5.1 says to register the cleanup handler AFTER `jobProcessor.start()`, which violates the startup ordering table in the overview. This must be fixed before implementation begins to avoid the cleanup job going dead on first restart. All other issues are P2/P3.

lsm (commented):

The revised plan has addressed most of the feedback from the previous two reviews. The overall quality is significantly improved. Below is an assessment of each previously raised issue, followed by one new issue introduced by the revisions.

---

### Previously Raised Issues — Status

**1. SpaceRuntime setInterval (was P1)** — RESOLVED. The Out of Scope section (`00-overview.md` lines 23-31) now explicitly lists `SpaceRuntime.tickTimer`, `TaskAgentManager` polling, `WebSocketServerTransport.staleCheckTimer`, `JobQueueProcessor` internal poll, and `app.ts` shutdown check as out-of-scope with clear justifications. Task 6.2 is required to verify and document these exclusions in a code comment.

**2. scheduleTickAfterRateLimitReset migration (was P1)** — RESOLVED. The plan now has a complete inventory section in `04-room-runtime-tick.md` (lines 22-48) that enumerates all 17 `scheduleTick()` sites, all 6 `scheduleTickAfterRateLimitReset()` sites, and both `queueMicrotask` sites. It correctly notes that 2 of the 17 `scheduleTick` calls are inside `scheduleTickAfterRateLimitReset` and will be removed when that method is rewritten. The migration of `scheduleTickAfterRateLimitReset` is specified in Task 4.2 subtask 8 with the delay computation and 5-second buffer preserved.

**3. Pause/resume dedup breakage (was P1)** — RESOLVED. The Pause/Resume and Dedup Design section in `04-room-runtime-tick.md` (lines 54-93) provides a clear design: `pause()` calls `cancelPendingTickJobs()` to delete pending tick jobs; `resume()` calls `enqueueRoomTick()` for a fresh start; the handler's `finally` block only re-schedules if runtime is still running AND was not skipped. The `cancelPendingTickJobs` helper is specified with pseudocode.

**4. Task 1.2/1.3 test overlap (was P1)** — RESOLVED. Task 1.3 now has a "Scope clarification vs Task 1.2 tests" section (`01-foundation.md` lines 110-112) that explicitly delineates: Task 1.2 tests app-level wiring (context fields, cleanup ordering, env var config), Task 1.3 tests processor contract in isolation (polling, dequeue, dispatch, stale reclamation, error/retry).

**5. Dedup strategy (was P2)** — RESOLVED. The dedup logic now correctly uses "no pending job exists at all" rather than "no future-scheduled job" (`00-overview.md` line 40, Task 3.1 subtask 3, Task 4.1 subtask 4). The race condition is acknowledged with a reasonable mitigation path: SQLite serializes writes, duplicate ticks are idempotent, and a unique partial index can be added in a follow-up if needed.

**6. maxConcurrent contention (was P2)** — RESOLVED. The default is now 5 instead of 3 (`00-overview.md` lines 60-68), with justification based on expected workload. It is configurable via `NEOKAI_JOB_QUEUE_MAX_CONCURRENT` environment variable (Task 1.2).

**7. Stale reclamation timing (was P2)** — RESOLVED. The plan now specifies eager `reclaimStale()` on startup (`00-overview.md` lines 72-74, Task 1.2 subtask 3). Task 5.2 explicitly tests that stale jobs are reclaimed immediately rather than after 60 seconds.

**8. Handler registration ordering (was P2)** — RESOLVED. A startup ordering table is present in `00-overview.md` (lines 48-56) with all 4 handler registrations ordered before `jobProcessor.start()`. Each milestone task references this ordering.

---

### New Issue Introduced by Revisions

**P1: Task 5.1 contradicts the startup ordering table for cleanup handler registration**

The startup ordering table in `00-overview.md` (line 54) correctly specifies that the cleanup handler must be registered at step 4, BEFORE `jobProcessor.start()` at step 5. However, Task 5.1 subtask 3 in `05-cleanup-job.md` (line 28) says:

> "After `jobProcessor.start()`, register the cleanup handler"

This is a direct contradiction. If the cleanup handler is registered after `jobProcessor.start()`, any pending `job_queue.cleanup` job from a previous run will be dequeued before the handler is registered, causing it to fail with "No handler registered" and consume retry attempts. This is exactly the failure mode the startup ordering table was designed to prevent.

**Fix:** Task 5.1 subtask 3 should say: "BEFORE `jobProcessor.start()`, register the `job_queue.cleanup` handler. Then check if a pending cleanup job already exists; if not, enqueue one."

---

### Minor Observations (P3)

**P3: Task count discrepancy.** The overview says "19 tasks across 6 milestones" (line 116), but I count: 3 (Milestone 1) + 3 (Milestone 2) + 3 (Milestone 3) + 4 (Milestone 4) + 2 (Milestone 5) + 3 (Milestone 6) = 18 tasks. Off by one.

**P3: Feature flag naming.** Task 4.2 suggests `NEOKAI_USE_JOB_QUEUE_ROOM_TICK` as an opt-out flag where `false` disables the migration. This is an inverted polarity — the flag name suggests "use job queue" but the default behavior (when unset) is to use the job queue. Consider `NEOKAI_DISABLE_JOB_QUEUE_ROOM_TICK=true` for clarity, or document that the flag must be explicitly set to `'false'` (string) to disable. Minor, since the feature flag is optional.

---

### Summary

The plan has addressed 8 of 8 previously raised issues. One new P1 was introduced: the cleanup handler registration ordering in Task 5.1 contradicts the startup ordering table, which would cause cleanup jobs from previous runs to fail on restart. This is a straightforward fix. The plan is otherwise ready for implementation.

lsm (commented):

**Recommendation: APPROVE**

This is the 3rd revision of the plan. All three previously flagged issues have been resolved, and no new blocking issues were introduced.

---

### Resolution of Previously Raised Issues

**1. P1 fix: Cleanup handler registration ordering (Task 5.1)** — RESOLVED.

`05-cleanup-job.md` line 28 now reads: "**BEFORE `jobProcessor.start()`** (step 4 in the startup ordering table — see `00-overview.md`), register the `job_queue.cleanup` handler on `jobProcessor`." Line 30 adds the warning: "This ordering is critical: if a cleanup job is pending from a previous run and the processor starts before the handler is registered, the job will fail with 'No handler registered' and consume retry attempts." This directly addresses the contradiction from the previous revision where it said "After `jobProcessor.start()`."

**2. P2 fix: Dedup reasoning for pending-only check in room-runtime-tick** — RESOLVED.

`04-room-runtime-tick.md` lines 64-66 now provides a three-part justification for checking only `pending` (not `processing`): (a) a `processing` job is already executing and its handler's `finally` block will re-schedule or not based on runtime state; (b) if `pause()` is called while a job is `processing`, the handler detects paused state and skips without re-scheduling; (c) checking `processing` would cause `enqueueRoomTick` to incorrectly skip when a tick is running, but the running tick's `finally` block handles the successor. This reasoning is clear and correct.

**3. P2 fix: GitHub polling field removal in Task 3.2** — RESOLVED.

`03-github-polling.md` Task 3.2 subtask 1 now explicitly lists three removal items:
- Line 69: Remove `clearInterval(this.pollingInterval)` from `stop()`
- Line 70: Remove the `pollingInterval` field declaration and its type (`ReturnType<typeof setInterval> | null`)
- Line 71: Remove any initialization of `pollingInterval` (e.g., `this.pollingInterval = null` in constructor)

All three aspects (field declaration, type, and constructor initialization) are covered.

---

### Other Previously Raised Issues (from earlier reviews) — Status

All issues from both prior review rounds have been addressed in this revision:

- SpaceRuntime and other out-of-scope setInterval users are documented in `00-overview.md` lines 23-31
- scheduleTickAfterRateLimitReset call sites are fully inventoried in `04-room-runtime-tick.md` lines 31-48
- Pause/resume design with `cancelPendingTickJobs` is specified in `04-room-runtime-tick.md` lines 50-93
- Task 1.2/1.3 test scope is delineated in `01-foundation.md` lines 110-112
- maxConcurrent raised to 5 with env var override in `00-overview.md` lines 60-68
- Eager stale reclamation on startup in `00-overview.md` lines 72-74
- Handler registration ordering table in `00-overview.md` lines 48-58
- Rollback strategy in `00-overview.md` lines 76-84
- Line number drift warning in `04-room-runtime-tick.md` lines 23-29
- GitHub API mocking note in Task 3.3 lines 99
- File-backed SQLite note in Task 6.1 lines 21, 28-29
- Task count corrected to 18 in `00-overview.md` line 116

---

**P3: The startup ordering table (line 54) references "step 4" for cleanup handler, but the table actually shows it as step 4 out of 5.** Task 5.1 line 28 says "step 4 in the startup ordering table" which is correct. No issue, just noting the cross-reference is accurate.

**P3: Task 4.1 subtask 4 uses limit 100 for dedup while the overview discusses limit 1000 in a different context.** The 100 limit is appropriate for per-room dedup (at most one pending tick per room). The 1000 reference in the overview describes the prior approach. This inconsistency is cosmetic — the plan explains the rationale for 100 on line 63 of `04-room-runtime-tick.md`.

---

### Summary

The plan is now comprehensive and internally consistent. All handler registrations are ordered before `jobProcessor.start()`. The dedup logic is correctly specified as "no pending job at all" with clear reasoning for excluding `processing` status. The GitHub polling migration correctly covers field removal. The plan is ready for implementation.

lsm (commented):

**Recommendation: APPROVE** — All three issues from the previous review cycle have been correctly resolved. The plan is ready for implementation.

---

## Verification of P1/P2 Fixes

### P1 fix: Cleanup handler registered BEFORE jobProcessor.start() — RESOLVED

`05-cleanup-job.md` Task 5.1 subtask 3 now reads:

> **BEFORE `jobProcessor.start()`** (step 4 in the startup ordering table — see `00-overview.md`), register the `job_queue.cleanup` handler on `jobProcessor`

The text also includes an explicit warning: "This ordering is critical: if a cleanup job is pending from a previous run and the processor starts before the handler is registered, the job will fail with 'No handler registered' and consume retry attempts." This is correct, specific, and directly addresses the failure mode.

### P2 fix: Dedup reasoning for 'pending'-only check in room-runtime tick — RESOLVED

`04-room-runtime-tick.md` Pause/Resume and Dedup Design section (around the enqueueRoomTick description) now provides three explicit reasons why only `pending` status is checked:

> (a) a `processing` job is already executing and its handler's `finally` block will either re-schedule (if runtime is still `running`) or not (if `paused`/`stopped`); (b) if `pause()` is called while a job is `processing`, the handler detects the paused state and skips without re-scheduling — leaving zero pending ticks, which is correct; (c) checking `processing` would cause `enqueueRoomTick` to skip when a tick is actively running, but the running tick's `finally` block handles the successor.

This reasoning correctly covers the edge cases and makes the invariant explicit.

### P2 fix: GitHub polling field removal — RESOLVED

`03-github-polling.md` Task 3.2 subtask 1 now explicitly says:

> - Remove the `pollingInterval` field declaration and its type (`ReturnType<typeof setInterval> | null`)
> - Remove any initialization of `pollingInterval` (e.g., `this.pollingInterval = null` in constructor)

This is explicit, complete, and leaves no ambiguity for the implementer.

---

## No New Issues Found

The remaining P3 observations from the prior cycle (inverted feature flag polarity for `NEOKAI_USE_JOB_QUEUE_ROOM_TICK`) remain unaddressed, but since they were P3 suggestions and do not affect correctness, this is acceptable.

The task count discrepancy (19 vs 18) is now fixed: `00-overview.md` line 116 correctly reads "18 tasks across 6 milestones (3 + 3 + 3 + 4 + 2 + 3)."

The plan is internally consistent, all handler registrations are ordered before `jobProcessor.start()`, and the dedup strategy is coherent across all three migrated subsystems.

PR #613, opened 2026-03-21
feat: extend WorkflowStep with agents array and WorkflowChannel topology

lsm:
- Add WorkflowStepAgent interface: { agentId, instructions? }
- Add WorkflowChannel interface: { from, to, direction, label? }
  with support for one-way, bidirectional point-to-point, and hub-spoke
- Make agentId optional on WorkflowStep and WorkflowStepInput (backward-compat)
- Add agents? and channels? fields to WorkflowStep and WorkflowStepInput
- Add ResolvedChannel type for concrete routing rules after expansion
- Add resolveStepAgents(): normalises agentId/agents duality with precedence rules
- Add resolveStepChannels(): expands all topology patterns (A→B, A↔B,
  A→[B,C,D], A↔[B,C,D] hub-spoke, *→B, A→*) into concrete routing rules
- Add validateStepChannels(): validates channel role references against step agents
- Export all new types and utilities from @neokai/shared
- Fix downstream type errors from agentId becoming optional
- 34 unit tests covering all topology patterns and edge cases


lsm (commented):

> **Recommendation: REQUEST_CHANGES** (posted as COMMENT since reviewer == PR author)

---

### Overall Assessment

The implementation is clean, well-structured, and the happy-path behavior is correct. The 34-test suite is solid and covers the core scenarios well. However, there are several significant gaps between the PR's stated scope and what was actually updated, plus one genuine correctness bug and a data-loss issue in the export pipeline. These need to be addressed before merge.

---

### P0 — Blocking

**1. `validateStepAgentRef` in `SpaceWorkflowManager` still requires `agentId`; it does not accept `agents`**

`packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, line 178–191

```ts
private validateStepAgentRef(spaceId: string, step: WorkflowStepInput, index: number): void {
    if (!step.agentId || !step.agentId.trim()) {
        throw new WorkflowValidationError(
            `step[${index}]: agentId must be a non-empty SpaceAgent UUID`
        );
    }
    ...
}
```

`WorkflowStepInput.agentId` is now optional (by design), so a valid multi-agent step with only `agents` populated will be rejected at the `workflow.create` and `workflow.update` RPC layer with the misleading error `"agentId must be a non-empty SpaceAgent UUID"`. This validation gate was not updated to reflect the new type contract and blocks the new feature from functioning end-to-end.

**2. `resolveTaskTypeForStep` in `SpaceRuntime` only reads `step.agentId`, ignoring the new `agents` array**

`packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 366–383

```ts
resolveTaskTypeForStep(step: WorkflowStep): ResolvedTaskType {
    const agent = step.agentId ? this.config.spaceAgentManager.getById(step.agentId) : undefined;
    ...
    return { taskType: 'coding', customAgentId: step.agentId };
}
```

For a multi-agent step (only `agents` set, `agentId` undefined), this always falls into the `!agent` branch and returns `{ taskType: 'coding', customAgentId: undefined }`. The task type and custom agent assignment are silently wrong. The function should call `resolveStepAgents(step)` and use the first agent as the primary for type resolution, or handle multi-agent steps explicitly.

**3. `rowToStep` in the repository throws on valid multi-agent steps stored with a null `agent_id`**

`packages/daemon/src/storage/repositories/space-workflow-repository.ts`, lines 96–107

```ts
function rowToStep(row: StepRow): WorkflowStep {
    if (!row.agent_id) {
        throw new Error(`WorkflowStep ${row.id}: agent_id is null in DB`);
    }
    ...
}
```

`insertStep` already stores `input.agentId ?? ''` (line 411), which means multi-agent steps currently survive (empty string), but this throws for `null`. More importantly: the `agents` array and `channels` fields are never persisted or read back. A multi-agent step created via the API would be stored as a single-agent step (with an empty `agentId`) and lose all `agents`/`channels` data on any read. There is no DB column or JSON field to store `WorkflowStep.agents` or `WorkflowStep.channels`.

---

### P1 — Should Fix

**4. Export pipeline silently loses `agents` and `channels` on multi-agent steps**

`packages/daemon/src/lib/space/export-format.ts`, lines 192–197

```ts
const exportedSteps: ExportedWorkflowStep[] = workflow.steps.map((step) => {
    const resolvedAgentId = step.agentId ?? '';
    const agentRef = agentIdToName.get(resolvedAgentId) ?? resolvedAgentId;
    const exported: ExportedWorkflowStep = { agentRef, name: step.name };
    if (step.instructions !== undefined) exported.instructions = step.instructions;
    return exported;
});
```

For multi-agent steps, `step.agentId` is `undefined`, so `resolvedAgentId` is `''`, and `agentRef` falls back to `''`. The exported step has a blank `agentRef` and all `agents`/`channels` data is dropped. `ExportedWorkflowStep` also has no fields for `agents` or `channels`, so this is a type-level gap as well. A round-trip export/import of a multi-agent workflow would produce a corrupted workflow.

**5. `task-agent-tools.ts` fallback `step.agentId` is now `undefined` for multi-agent steps**

`packages/daemon/src/lib/space/tools/task-agent-tools.ts`, line 255

```ts
customAgentId: stepTask.customAgentId ?? step.agentId,
```

For multi-agent steps where `step.agentId` is `undefined`, this falls through to `undefined`. The comment above explicitly says this is a fallback for preset agents, so the silent `undefined` case is now reachable and leads to a no-op agent assignment. This should use `resolveStepAgents(step)[0]?.agentId` instead.

**6. `formatStep` in `task-agent.ts` uses raw `step.agentId` without the new resolver**

`packages/daemon/src/lib/space/agents/task-agent.ts`, line 85

```ts
const agent = agents.find((a) => a.id === step.agentId);
const agentLabel = agent ? `${agent.name} (role: ${agent.role})` : `agent id: ${step.agentId}`;
```

For a multi-agent step, `step.agentId` is `undefined`, so `agent` is always `undefined` and the label degrades to `"agent id: undefined"`. Minor display issue but symptomatic of the broader pattern of not updating call sites.

**7. `validateStepChannels` does not report duplicate role collisions in the agents array**

`packages/shared/src/types/space-utils.ts`

If two agents in a step share the same `SpaceAgent.role` value, the `roleToAgentId` map silently retains only the last one, and `validateStepChannels` builds a `knownRoles` set that also silently deduplicates. Channels referencing that role are validated as valid but resolved non-deterministically. There is no test covering this and no error is surfaced.

---

**8. The DB schema has no column for `agents` or `channels` on `space_workflow_steps`**

`StepConfigJson` (line 79 of `space-workflow-repository.ts`) only stores `instructions`. The new `WorkflowStepAgent[]` and `WorkflowChannel[]` fields need to be stored (e.g. in the JSON config column) and deserialized in `rowToStep`. Without this, no persisted multi-agent step can ever be round-tripped correctly.

**9. `WorkflowStepManager.validateStepAgentRef` error message should distinguish the two formats**

When neither `agentId` nor `agents` is provided, the error message should say so rather than just flagging `agentId`. Otherwise operators reading logs will be confused trying to provide an `agentId` for a multi-agent step.

**10. Missing test: both `agentId` and `agents` provided — warning behavior**

The PR description says "if both provided, `agents` takes precedence with a warning", but `resolveStepAgents` silently ignores `agentId` with no warning, and there is no test verifying the precedence override with both fields set is safe at the manager/validator layer.

---

### P3 — Nits

**11. `_expandChannel` private helper uses a leading-underscore naming convention**

The rest of the codebase uses TypeScript `private` keyword or module-scope unexported functions without the underscore prefix convention. `_expandChannel` is module-private (unexported) but the leading underscore is stylistically inconsistent. Consider renaming to `expandChannel` since it is already unexported.

**12. `resolveStepChannels` JSDoc says "Agent IDs not found ... are skipped" but means roles**

Line 96 of `space-utils.ts`: the docstring says `"Agent IDs not found in the provided agents list are skipped"`, but in the implementation the lookup is by role, not agent ID. The comment is slightly misleading.

---

### Summary of Gap

The core types (`WorkflowStepAgent`, `WorkflowChannel`, and the updated optional `agentId`) are correctly defined and the utility functions work correctly in isolation. However, the integration layer — DB persistence, export/import, the RPC validation gate, and runtime execution — was not updated. A multi-agent step cannot be created via the API, cannot be persisted, cannot be exported, and cannot be executed. The new types exist in the type system only.

lsm (commented):

The type definitions and utility functions in `space-utils.ts` are well-designed and thoroughly tested. However, making `agentId` optional on `WorkflowStep` and `WorkflowStepInput` introduced type-level incompatibilities in several existing callers that were NOT updated. These are not caught by `tsc` because the callers use `?? ''` coercions (turning `undefined` into empty string), but the runtime semantics are broken in several places.

---

### P0 — Blocking Issues

**1. `validateStepAgentRef` rejects multi-agent steps (`packages/daemon/src/lib/space/managers/space-workflow-manager.ts:178-192`)**

The validation function `validateStepAgentRef` still requires `agentId` to be non-empty:
```typescript
if (!step.agentId || !step.agentId.trim()) {
    throw new WorkflowValidationError(
        `step[${index}]: agentId must be a non-empty SpaceAgent UUID`
    );
}
```
Any step that uses `agents` (the new multi-agent array) instead of `agentId` will fail workflow creation validation. This completely blocks the new feature from being used at the persistence layer.

**2. `customAgentId` falls back to `undefined` for multi-agent steps (`packages/daemon/src/lib/space/tools/task-agent-tools.ts:255`)**

```typescript
customAgentId: stepTask.customAgentId ?? step.agentId,
```
When `step.agentId` is `undefined` (multi-agent step), this silently passes `undefined` as the `customAgentId`. The downstream task spawning may then fail to assign any agent, breaking task execution for multi-agent steps.

**3. `formatStep` does not handle multi-agent steps (`packages/daemon/src/lib/space/agents/task-agent.ts:85-86`)**

```typescript
const agent = agents.find((a) => a.id === step.agentId);
const agentLabel = agent ? `...` : `agent id: ${step.agentId}`;
```
For multi-agent steps, `step.agentId` is `undefined`, so this renders `"agent id: undefined"` in the task agent's context prompt. The prompt should instead list all agents in the step.

**4. `resolveTaskTypeForStep` returns `customAgentId: undefined` for multi-agent steps (`packages/daemon/src/lib/space/runtime/space-runtime.ts:367-371`)**

When `step.agentId` is falsy, the agent lookup returns `undefined`, and the function returns `{ taskType: 'coding', customAgentId: step.agentId }` which is `customAgentId: undefined`. This means multi-agent steps get no agent assignment at all during task creation.

---

### P1 — Should Fix

**5. Repository stores empty string instead of null for `agentId` column (`packages/daemon/src/storage/repositories/space-workflow-repository.ts:410`)**

```typescript
input.agentId ?? '',
```
Storing `''` instead of `null` in the DB for multi-agent steps means the round-trip will return `''` (truthy in some contexts, falsy in JS but confusing in SQL queries). Consider storing `null` instead to preserve the semantic of "no single agent assigned."

**6. `roleToAgentId` map silently overwrites duplicate roles (`packages/shared/src/types/space-utils.ts:116-122`)**

If a step has two agents with the same role (e.g., two "coder" agents), the `Map` silently keeps only the last one. There is no validation or warning for this. While `validateStepChannels` checks role references, it does not flag duplicate roles. Consider either documenting this as an intentional limitation or adding validation.

**7. Wildcard `*` in array `to` is not expanded when mixed with other roles (`packages/shared/src/types/space-utils.ts:147`)**

```typescript
const toRoles: string[] = toList.length === 1 && toList[0] === '*' ? allRoles : toList;
```
A channel like `{ from: 'coder', to: ['reviewer', '*'], direction: 'one-way' }` would treat `*` as a literal role name. This is defensible but should be documented, and `validateStepChannels` should reject `*` when mixed with other roles in the array.

**8. `ExportedWorkflowStep` not updated for multi-agent steps**

The export format type and `exportWorkflow` function (`packages/daemon/src/lib/space/export-format.ts:190-195`) only handle `step.agentId`, ignoring `step.agents`. Exporting a multi-agent workflow would lose all agent assignments. Should at minimum be documented as a known limitation.

**9. Built-in workflow test assumes `agentId` is always defined (`packages/daemon/tests/unit/space/built-in-workflows.test.ts:255`)**

```typescript
expect(VALID_BUILTIN_ROLES.has(step.agentId)).toBe(true);
```
With `agentId` now optional, this passes `undefined` to `Set.has()`. Currently safe because built-in workflows all use `agentId`, but fragile if any built-in workflow is updated to use `agents`.

---

**10. No unified validation at the persistence boundary**

The doc comments say "At least one of `agentId` or `agents` must be provided" but `validateStepAgentRef` in the workflow manager only checks `agentId`. There should be a unified validation point that enforces the constraint for both formats.

**11. Frontend components not updated for multi-agent display**

Several frontend components still only display `step.agentId`:
- `packages/web/src/components/space/WorkflowStepCard.tsx:151`
- `packages/web/src/components/space/visual-editor/WorkflowNode.tsx:76`
- `packages/web/src/components/space/SpaceAgentList.tsx:141`

For multi-agent steps, these will show `undefined` or empty string. While the full frontend update may be a follow-up, these paths should at minimum not render broken state.

---

### What works well

- The `resolveStepAgents`, `resolveStepChannels`, and `validateStepChannels` utility functions are clean, well-documented, and thoroughly tested (34 tests covering all topology patterns).
- Hub-spoke resolution correctly prevents spoke-to-spoke messaging.
- Self-loop skipping and unknown-role graceful handling are solid.
- The `ResolvedChannel` type with `isHubSpoke` flag is a good design for downstream consumers.
- Backward compatibility for `agentId`-only steps works correctly in the utility layer.

### Bottom line

The utility layer is solid but the existing callers were not updated to handle the new optional `agentId`. Multi-agent steps will fail at workflow creation validation (P0 #1) and produce incorrect runtime behavior if that validation is bypassed. The callers need to be updated to use `resolveStepAgents` or equivalent patterns before this can ship.

lsm (commented):

Re-review of PR #613 after P0/P1 fixes. The previous P0 issues (rowToStep throwing on null agent_id, insertStep storing empty string, resolveTaskTypeForStep not handling multi-agent, etc.) are all genuinely resolved. The core shared types, utility functions, and tests are well-implemented.

However, two integration-level issues remain:

### P1: `updateWorkflow` validation drops `agents` and `channels`

**File:** `packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, lines 98-104

When `updateWorkflow` maps `WorkflowStep` to `WorkflowStepInput` for validation, it only copies `id`, `name`, `agentId`, and `instructions` -- dropping `agents` and `channels`:

```ts
const inputs: WorkflowStepInput[] = (params.steps ?? []).map(
    (s): WorkflowStepInput => ({
        id: s.id,
        name: s.name,
        agentId: s.agentId,
        instructions: s.instructions,
    })
);
```

A multi-agent step that has `agents: [...]` but no `agentId` will fail `validateStepAgentRef` with "at least one of agentId or agents must be provided" even though the step is valid. The same issue exists at lines 112-117 for existing-step transition validation.

Fix: include `agents` and `channels` in the mapping.

### P1: `getWorkflowsReferencingAgent` misses multi-agent step references

**File:** `packages/daemon/src/storage/repositories/space-workflow-repository.ts`, lines 369-380

The query `SELECT DISTINCT workflow_id FROM space_workflow_steps WHERE agent_id = ?` only checks the `agent_id` column. Multi-agent steps store `null` in `agent_id` and put agent IDs in the JSON `config` column. If an agent is referenced only via the `agents[]` array (not the top-level `agentId`), deleting that agent will not be blocked by the referential integrity check in `SpaceAgentManager`.

This is a data integrity issue -- users could delete agents that are actively used in workflow steps, leading to broken workflows at runtime.

Fix: also search the `config` JSON column, e.g. `WHERE agent_id = ? OR config LIKE '%' || ? || '%'` (or parse JSON in application code).

### P2: Export format note

**File:** `packages/daemon/src/lib/space/export-format.ts`, lines 192-201

The export format only supports single-agent steps (using `agentRef`). The code has a clear comment that full multi-agent export/import is deferred to Milestone 5. This is fine as a deliberate scope cut, but the `ExportedWorkflowStep` type in `packages/shared/src/types/space.ts` (lines 842-849) does not mention this limitation. Consider adding a doc comment there too so consumers of the type are aware.

### What looks good

- `resolveStepAgents`, `resolveStepChannels`, `validateStepChannels` are clean and well-tested (37 tests total)
- Backward compatibility is solid: steps with only `agentId` work exactly as before
- Repository correctly persists `agents[]` and `channels[]` in JSON config and deserializes them
- `rowToStep` no longer throws on null `agent_id`
- `insertStep` stores `null` (not empty string) for missing `agentId`
- `resolveTaskTypeForStep` uses `resolveStepAgents()` for multi-agent support
- `formatStep` lists all agents for multi-agent steps
- Channel validation catches duplicate roles and wildcard-mixed-with-roles

**Recommendation: REQUEST_CHANGES** -- the two P1 issues should be addressed before merge.

lsm (commented):

---

**Recommendation: REQUEST_CHANGES** — One P0 bug found (breaks all multi-agent-only step updates) and one P1 bug (agent deletion protection miss for multi-agent steps). Previous P0/P1 issues from the earlier review are all genuinely resolved.

---

## Summary of Findings

The core type system additions (WorkflowStep.agents, WorkflowChannel, resolveStepAgents, resolveStepChannels, validateStepChannels) are well-implemented with clean logic and solid test coverage. The repository layer correctly persists and deserializes agents/channels. The previously-reported issues — rowToStep null agent_id, insertStep null vs empty string, resolveTaskTypeForStep, task-agent-tools fallback, formatStep, export-format fallback, validateStepChannels — are all genuinely fixed.

Two new issues were introduced by the changes:

---

## P0 — updateWorkflow breaks multi-agent-only steps during validation

**File:** `packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, lines 97–106

In `updateWorkflow`, when `params.steps` is provided, the code builds a `WorkflowStepInput[]` for validation:

```ts
const inputs: WorkflowStepInput[] = (params.steps ?? []).map(
  (s): WorkflowStepInput => ({
    id: s.id,
    name: s.name,
    agentId: s.agentId,
    instructions: s.instructions,
    // *** agents and channels are NOT included ***
  })
);
this.validateSteps(existing.spaceId, inputs);
```

`agents` and `channels` are stripped from the mapped `WorkflowStepInput` objects. When a step uses only `agents[]` (no `agentId`), `validateStepAgentRef` receives a step with both `agentId: undefined` and `agents: undefined`, triggering the error:

> `step[N]: at least one of agentId or agents must be provided`

This means **any call to updateWorkflow that includes multi-agent-only steps will fail at the validation layer** and never reach the repository. The fix is to include `agents` and `channels` in the mapped input:

```ts
const inputs: WorkflowStepInput[] = (params.steps ?? []).map(
  (s): WorkflowStepInput => ({
    id: s.id,
    name: s.name,
    agentId: s.agentId,
    agents: s.agents,       // must be included
    channels: s.channels,   // must be included
    instructions: s.instructions,
  })
);
```

Note: the `existingStepInputs` mapping at lines 112–117 (the `else if` branch for transition-only updates) strips the same fields, but those inputs only feed `validateTransitions` which does ID-existence checks — `validateStepAgentRef` is not called on them, so that branch is safe.

---

## P1 — getWorkflowsReferencingAgent misses multi-agent steps

**File:** `packages/daemon/src/storage/repositories/space-workflow-repository.ts`, lines 369–380

`getWorkflowsReferencingAgent` only queries the `agent_id` SQL column:

```ts
const stepRows = this.db
  .prepare(`SELECT DISTINCT workflow_id FROM space_workflow_steps WHERE agent_id = ?`)
  .all(agentId) as Array<{ workflow_id: string }>;
```

For multi-agent steps the `agent_id` column is NULL (by design, as documented). The agent IDs are stored in the JSON `config` column as `agents[]`. This means if an agent is only referenced via `agents[]`, `getWorkflowsReferencingAgent` will return no workflows for it, allowing the agent to be deleted even though it is still in use. The deletion protection that `SpaceAgentManager` relies on is silently bypassed for multi-agent steps.

The fix requires either a JSON search in the query (e.g., SQLite `json_each`) or a post-filter in application code after loading the full workflow.

---

## P2 — Web visual editor silently drops agents/channels on round-trip

**File:** `packages/web/src/components/space/WorkflowEditor.tsx`, lines 108–114 and 123–130  
**File:** `packages/web/src/components/space/visual-editor/serialization.ts`, lines 225–234

The `StepDraft` interface (WorkflowStepCard.tsx line 21–29) only has `agentId` — no `agents` or `channels`. When loading an existing multi-agent workflow for editing via `initFromWorkflow` or `workflowToVisualState`, the `agents[]` and `channels[]` fields are silently dropped. If the user opens and saves a multi-agent workflow through the visual editor, those fields will be overwritten with an empty single-agent step (falling back to `agentId: ''`). This is a data-loss path.

This is a pre-existing limitation that the PR description acknowledges (`ExportedWorkflowStep` also defers multi-agent export). Since there is no UI to _create_ multi-agent steps yet, the actual risk is limited to someone manually constructing a multi-agent workflow via API and then accidentally saving it through the UI editor. Still worth calling out as it's not commented anywhere in the web code.

---

## P3 — ExportedWorkflowStep comment is incomplete

**File:** `packages/shared/src/types/space.ts`, line 842–849

The JSDoc for `ExportedWorkflowStep` describes the single-agent format only. Since multi-agent export support is intentionally deferred, a brief comment noting this deferral (matching the inline comment in `export-format.ts` line 192–193) would keep the type docs consistent.

---

## What was correctly fixed (confirmation)

- `validateStepAgentRef`: accepts agentId OR agents[], validates all entries — confirmed correct.
- `rowToStep`: no longer throws on null agent_id; deserializes agents and channels from JSON config — confirmed correct.
- `insertStep`: stores null (not '') for missing agentId; persists agents[] and channels[] — confirmed correct.
- `resolveTaskTypeForStep`: uses resolveStepAgents()[0]?.agentId — confirmed correct.
- `task-agent-tools.ts`: uses resolveStepAgents()[0]?.agentId fallback — confirmed correct.
- `formatStep`: lists all agents for multi-agent steps — confirmed correct.
- `export-format.ts`: falls back to agents[0].agentId for multi-agent steps — confirmed correct.
- `validateStepChannels`: detects duplicate roles, rejects * mixed with other roles in array to — confirmed correct.
- Test count: 37 tests (34 in space-utils.test.ts + 3 added to built-in-workflows.test.ts) — confirmed passing structure.

lsm (commented):

**Recommendation: APPROVE**

---

### Re-review (3rd iteration) — All previous P0/P1 issues resolved

All four fixes from the latest commit are verified:

1. **`updateWorkflow` mapping now includes `agents` and `channels`** (`packages/daemon/src/lib/space/managers/space-workflow-manager.ts:103-104`) — Multi-agent-only steps will no longer fail validation during workflow updates. The `else if` branch at lines 114-119 (transition-only validation) intentionally omits these fields since it only feeds `validateTransitions` which checks step IDs, not agent references.

2. **`getWorkflowsReferencingAgent` searches both `agent_id` column and `config` JSON** (`packages/daemon/src/storage/repositories/space-workflow-repository.ts:378-383`) — The `WHERE agent_id = ? OR config LIKE '%' || ? || '%'` pattern correctly catches agent references in both single-agent steps (via `agent_id` column) and multi-agent steps (via JSON `config` column). The LIKE pattern is conservative but safe since agent IDs are UUIDs.

3. **JSDoc on `ExportedWorkflowStep`** (`packages/shared/src/types/space.ts:838-846`) — Clearly documents that multi-agent export/import is deferred to Milestone 5.

4. **Comment in `WorkflowEditor.tsx` `initFromWorkflow`** (`packages/web/src/components/space/WorkflowEditor.tsx:93-97`) — Documents the data-loss risk for API-created multi-agent workflows opened in the single-agent editor.

### Verification

- All 37 unit tests in `packages/shared/tests/space-utils.test.ts` pass.
- The type system, utility functions (`resolveStepAgents`, `resolveStepChannels`, `validateStepChannels`), and persistence layer form a consistent end-to-end pipeline.
- Repository correctly stores `agents[]` and `channels[]` in the JSON `config` column and deserializes them in `rowToStep`.
- Validation at `validateStepAgentRef` properly accepts either `agentId` or `agents` (or both, with `agents` taking precedence).
- Backward compatibility for single-agent steps is maintained throughout.

### P3 — Nit

**1. `existingStepInputs` mapping at lines 114-119 could include `agents`/`channels` for consistency**

The `else if` branch that builds `existingStepInputs` for transition-only validation still omits `agents` and `channels`. This is correct because `validateTransitions` only checks step IDs, but including them would keep the mapping consistent with the primary branch and prevent future confusion if `validateTransitions` ever needs agent data.

No blocking issues remain. The implementation is clean and well-documented.

lsm (commented):

> This is the author's own PR — posting as COMMENT per reviewer identity protocol.
> **Recommendation: REQUEST_CHANGES** — one P1 issue, several P2 items remain.

---

## Summary

Third iteration review of PR #613. The two P0 fixes from the second round are correctly applied and genuinely resolve the issues. One new P1 gap was found (no repository-level integration tests for the new multi-agent persistence path). Several P2 items remain.

---

## Previous P0/P1 Issues — Status

### Fix 1: `updateWorkflow` mapping now includes `agents` and `channels` (P0 — FIXED)

The `SpaceWorkflowManager.updateWorkflow` mapping at line 98–107 now correctly includes both `agents` and `channels` when converting `WorkflowStep → WorkflowStepInput` for validation. This prevents the false `validateStepAgentRef` failure on multi-agent-only steps.

Note: The separate `existingStepInputs` path at lines 114–119 (used when only `transitions` are updated, not `steps`) still omits `agents` and `channels`, but this is harmless — `validateTransitions` only uses `step.id` for cross-reference validation, never calls `validateStepAgentRef`.

### Fix 2: `getWorkflowsReferencingAgent` now checks the `config` JSON column (P0 — FIXED)

The `WHERE agent_id = ? OR config LIKE '%' || ? || '%'` query is correct. UUIDs are globally unique so the conservative LIKE pattern cannot produce false positives. This correctly catches agents referenced only via the `agents[]` format in the JSON config column.

---

## Remaining Issues

### P1 — Missing repository-level integration tests for multi-agent persistence

The shared package tests in `packages/shared/tests/space-utils.test.ts` thoroughly cover the utility functions (`resolveStepAgents`, `resolveStepChannels`, `validateStepChannels`), but the **repository persistence path** for the new fields has no test coverage in `packages/daemon/tests/unit/space/space-workflow.test.ts`.

Specifically missing:
1. A test for creating a workflow with `agents[]` steps and verifying the round-trip (create → read back → confirm `step.agents` is populated, `step.agentId` is undefined/absent).
2. A test for creating a workflow with `channels[]` and verifying the round-trip.
3. A test for `getWorkflowsReferencingAgent` with a **multi-agent step** (where `agent_id = NULL` and the ID is only in the JSON config). The existing test only covers the `agent_id` column path — the new LIKE-based path is untested.

These are real integration risks: a regression in `insertStep` or `rowToStep` would go undetected.

**File:** `packages/daemon/tests/unit/space/space-workflow.test.ts`

---

### P2 — `WorkflowEditor.tsx` and `serialization.ts` silently drop `agents[]`/`channels[]` without any user warning

Both `initFromWorkflow` (WorkflowEditor) and `workflowToVisualState` (serialization) now have comments documenting the data-loss limitation. This is appreciated. However, neither path gives any user-visible indication when a multi-agent workflow is opened — the editor simply renders with empty agent selectors (or the first `agentId` if present), and saving would silently overwrite the multi-agent configuration.

**Suggestion:** Consider adding a visible banner or read-only mode when a loaded workflow contains multi-agent steps. Given that the API already supports creating multi-agent steps, a user who creates one via RPC and then opens it in the editor will lose that configuration on save with no warning.

**Files:**
- `packages/web/src/components/space/WorkflowEditor.tsx` (line 94–97)
- `packages/web/src/components/space/visual-editor/serialization.ts` (line 119–124)

---

### P2 — Export format silently drops all but the primary agent for multi-agent steps without caller warning

The `exportWorkflow` function at lines 194–201 of `packages/daemon/src/lib/space/export-format.ts` uses the first `agents[]` entry as the fallback `agentRef` and silently discards all remaining agents and channel topology. The JSDoc on `ExportedWorkflowStep` now documents this M5 deferred limitation clearly.

**Suggestion:** Consider returning a structured result with optional warnings (e.g., `{ result: ExportedSpaceWorkflow, warnings: string[] }`) from `exportWorkflow` when multi-agent steps are present, so the RPC handler can surface a warning to the user. Non-blocking given M5 scope, but worth raising while the export format is still young.

**File:** `packages/daemon/src/lib/space/export-format.ts` (lines 191–201)

---

### P2 — `resolveStepChannels` may throw but its JSDoc does not document the throw

`resolveStepChannels` (line 117) calls `resolveStepAgents` without a try/catch. If a step has neither `agentId` nor `agents`, the function throws synchronously. The JSDoc for `resolveStepChannels` does not document this `@throws` behavior, making it a latent surprise for future callers.

**Suggestion:** Add `@throws {Error} When neither agentId nor agents is provided` to the JSDoc on `resolveStepChannels`, consistent with how `resolveStepAgents` already documents its throw.

**File:** `packages/shared/src/types/space-utils.ts` (lines 113–136)

---

## What Is Correct

- **Type definitions** (`WorkflowStepAgent`, `WorkflowChannel`, `ResolvedChannel`, updated `WorkflowStep` and `WorkflowStepInput`) are well-designed with clear JSDoc.
- **Backward compatibility** is correct: `agentId` remains optional, `resolveStepAgents` correctly implements the precedence rule, existing single-agent steps continue to work unchanged.
- **Storage layer**: The JSON config approach for storing `agents[]` and `channels[]` while keeping `agent_id` null for fast single-agent lookups is a sound design decision.
- **34 unit tests** in `packages/shared/tests/space-utils.test.ts` cover all topology patterns including edge cases (self-loops, wildcards, hub-spoke, duplicate roles).
- **Export format JSDoc** on `ExportedWorkflowStep` clearly documents the M5 deferred scope.
- **`initFromWorkflow` comment** correctly warns of data-loss for multi-agent steps in the single-agent editor.
- **`getWorkflowsReferencingAgent`** LIKE query is safe and correct.

lsm (commented):

**Recommendation: APPROVE** — All P0/P1 issues from previous iterations are resolved. The fourth iteration addresses the P1 test coverage gap with four targeted integration tests and a JSDoc fix on `resolveStepChannels`.

---

### Summary of changes reviewed

The PR extends `WorkflowStep` and `WorkflowStepInput` with:
- `agents?: WorkflowStepAgent[]` for multi-agent parallel steps
- `channels?: WorkflowChannel[]` for inter-agent messaging topology
- Backward-compatible: `agentId` remains optional, single-agent steps continue to work

---

### What was verified

**P1 test coverage gap — now filled**

The four new repository integration tests in `packages/daemon/tests/unit/space/space-workflow.test.ts` (lines 357–447) properly exercise the persistence layer against a real in-process SQLite database:

1. **agents[] round-trip** (line 357): creates a workflow with a multi-agent step using `agents: [{agentId: 'agent-multi-1', instructions: 'do A'}, {agentId: 'agent-multi-2'}]`, reads it back, and asserts that `step.agentId` is `undefined`, `step.agents` has the full two-entry array with all fields restored, and shared `instructions` survive.

2. **channels[] round-trip** (line 394): verifies all channel fields — `from`, `to` (including array form), `direction`, `label` — survive serialization through the JSON config column intact.

3. **getWorkflowsReferencingAgent via LIKE path** (line 324): creates a multi-agent step (so `agent_id = NULL` in DB), then calls `getWorkflowsReferencingAgent` for each of the two agents and confirms both are found via the config LIKE search. Also asserts that an unrelated agent returns nothing.

4. **Legacy regression guard** (line 431): confirms that single-agent `agentId`-only steps still produce `step.agentId = 'agent-1'`, `step.agents = undefined`, `step.channels = undefined` after round-trip.

These tests meaningfully test the new code paths: the `insertStep` JSON serialization, `rowToStep` deserialization, and the LIKE-based query in `getWorkflowsReferencingAgent`.

**P2 JSDoc fix — correct**

`resolveStepChannels` now documents `@throws {Error}` propagated from `resolveStepAgents` when neither `agentId` nor `agents` is provided (`packages/shared/src/types/space-utils.ts` line 113-114).

---

- `rowToStep` in `space-workflow-repository.ts` correctly handles `null` `agent_id` without throwing: the conditional at line 109 only sets `step.agentId` when `row.agent_id` is truthy.
- `insertStep` correctly stores `null` (not empty string) for multi-agent steps at line 435: `const agentIdValue = input.agentId && input.agentId.trim() ? input.agentId : null;`
- `updateWorkflow` in the manager correctly maps `agents` and `channels` through to `WorkflowStepInput` (lines 103-104), fixing the previously-reported P0 where multi-agent-only steps would fail validation on update.
- `getWorkflowsReferencingAgent` uses a parameterized LIKE query (`config LIKE '%' || ? || '%'`). The LIKE approach is safe here because UUID format (hexadecimal characters and hyphens only) cannot create false positives against other JSON fields in the config column.
- `resolveStepAgents` and `resolveStepChannels` in `space-utils.ts` correctly implement the precedence rules and topology expansion. The 34 shared-package unit tests cover all documented topology patterns.
- The `updateWorkflow` existing-step-to-input mapping (lines 114-119 of `space-workflow-manager.ts`) still omits `agents` and `channels` when mapping existing steps for transition-only validation. This is acceptable since the `validateStepAgentRef` call only needs to confirm at-least-one-of-agentId-or-agents for steps that already exist in the DB and were previously validated on creation.

---

- `serialization.ts` line 123 (`agentId: s.agentId ?? ''`): this produces an empty string for multi-agent steps when `initFromWorkflow` loads an API-created multi-agent workflow. The limitation is documented in the JSDoc comment at line 93-97 of `WorkflowEditor.tsx` and is an accepted M5 deferred item, not a regression.
- The `seedAgent` test helper hardcodes `role = 'coder'` for all inserted agents. This is fine for the persistence tests but means the `channels[]` round-trip test at line 394 doesn't need the agents in `allAgents` (it only verifies raw JSON survives, not role resolution). The test is still correct — it just doesn't exercise `resolveStepChannels` against a DB-loaded step, which is covered by the unit tests in `space-utils.test.ts` instead.

---

**Overall verdict:** The P1 coverage gap is genuinely filled. The four integration tests exercise real SQLite persistence and cover the core correctness properties: agents[] serialize/deserialize without loss, channels[] survive round-trip, multi-agent steps are found by the LIKE query, and legacy steps still work. No remaining P0/P1 issues.

lsm (commented):

Fourth iteration review. The P1 test coverage gap from the prior review has been addressed.

### Summary

The implementation correctly extends `WorkflowStep` with an optional `agents: WorkflowStepAgent[]` array and `channels: WorkflowChannel[]` topology declaration. The `agentId` field is made optional with clear precedence rules documented in JSDoc. Three utility functions (`resolveStepAgents`, `resolveStepChannels`, `validateStepChannels`) in `space-utils.ts` handle the resolution logic cleanly. All existing call sites have been updated to use `resolveStepAgents()` instead of raw `step.agentId` access.

### P1 Test Gap — Resolved

The four new repository integration tests in `space-workflow.test.ts` (commit a03543da) adequately cover:
1. **`agents[]` round-trip**: Verifies `agentId` is absent, `agents[]` is fully restored with all fields, shared instructions preserved.
2. **`channels[]` round-trip**: Verifies all channel fields (`from`, `to` array, `direction`, `label`) survive JSON config column serialization.
3. **`getWorkflowsReferencingAgent` via LIKE path**: Verifies both agents in a multi-agent step (where `agent_id = NULL`) are found by the updated query.
4. **Legacy regression guard**: Single-agent `agentId`-only steps still round-trip correctly.

1. **`WorkflowEditor.tsx` silently drops multi-agent data**: The comment at line ~95 correctly documents that multi-agent steps are downgraded to single-agent in the UI editor. This is acknowledged as a Milestone 5 limitation. The `agentId ?? ''` fallback is reasonable for now, though saving an API-created multi-agent workflow through the editor will lose the `agents[]` and `channels[]` data. The JSDoc warning is sufficient.

2. **Minor inconsistency in manager update path** (`space-workflow-manager.ts:114`): When building `existingStepInputs` for transition validation against existing steps, `agents` and `channels` are not mapped. This is harmless since `validateTransitions` only checks step IDs, not agent assignments.

3. **Export format limitation** (`export-format.ts:192-197`): Multi-agent steps are exported using only the primary agent's name as `agentRef`. The limitation is clearly documented in the `ExportedWorkflowStep` JSDoc and the code comment. This is appropriate for the current milestone.

### Verdict

**Recommendation: APPROVE** — The implementation is correct, backward-compatible, well-documented, and now has adequate test coverage for the persistence layer. The acknowledged Milestone 5 limitations (UI editor, export format) are properly documented and do not affect correctness of the type system and resolution logic being introduced.

PR #615, opened 2026-03-21
feat: update shared types for flexible SpaceSessionGroup model

lsm:
- Change SpaceSessionGroupMember.role from 'worker' | 'leader' to freeform string
- Add agentId? and status fields to SpaceSessionGroupMember
- Add taskId? to SpaceSessionGroup
- Update repository params, rowToGroup, rowToMember, and addMember signature
- Update test DB schema and add tests covering new fields

### DIFF
--- a/packages/daemon/src/storage/repositories/space-session-group-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-session-group-repository.ts
@@ -25,6 +25,24 @@ export interface UpdateSessionGroupParams {
 	taskId?: string | null;
 }
 
+export interface AddMemberParams {
+	/** Freeform role string matching SpaceAgent.role (e.g. 'coder', 'reviewer', 'security-auditor') */
+	role: string;
+	/** Display order within the group (default: 0) */
+	orderIndex?: number;
+	/** ID of the SpaceAgent config this session uses (nullable for system agents) */
+	agentId?: string;
+	/** Initial lifecycle state (default: 'active') */
+	status?: 'active' | 'completed' | 'failed';
+}
+
+export interface UpdateMemberParams {
+	role?: string;
+	orderIndex?: number;
+	agentId?: string | null;
+	status?: 'active' | 'completed' | 'failed';
+}
+
 export class SpaceSessionGroupRepository {
 	constructor(private db: BunDatabase) {}
 
@@ -84,16 +102,14 @@ export class SpaceSessionGroupRepository {
 	}
 
 	/**
-	 * Get all session groups that contain a specific task's associated sessions.
-	 * Returns groups matching by spaceId and name pattern (groups for a given task).
+	 * Get all session groups associated with a specific task.
+	 * Queries by the `task_id` column set at group creation time.
 	 */
 	getGroupsByTask(spaceId: string, taskId: string): SpaceSessionGroup[] {
-		// Groups are associated to tasks by their name convention (e.g. "task:{taskId}")
-		// or by querying members. We query groups that are named for a task.
 		const stmt = this.db.prepare(
-			`SELECT * FROM space_session_groups WHERE space_id = ? AND name = ? ORDER BY created_at ASC`
+			`SELECT * FROM space_session_groups WHERE space_id = ? AND task_id = ? ORDER BY created_at ASC`
 		);
-		const rows = stmt.all(spaceId, `task:${taskId}`) as Record<string, unknown>[];
+		const rows = stmt.all(spaceId, taskId) as Record<string, unknown>[];
 
 		return rows.map((row) => {
 			const members = this.getGroupMembers(row.id as string);
@@ -152,17 +168,12 @@ export class SpaceSessionGroupRepository {
 	}
 
 	/**
-	 * Add a session member to a group
-	 * Idempotent — updates role/orderIndex if the session is already a member
+	 * Add a session member to a group.
+	 * Idempotent — if the session is already a member, updates all provided fields.
 	 */
-	addMember(
-		groupId: string,
-		sessionId: string,
-		role: string,
-		orderIndex = 0,
-		agentId?: string,
-		status: 'active' | 'completed' | 'failed' = 'active'
-	): SpaceSessionGroupMember {
+	addMember(groupId: string, sessionId: string, params: AddMemberParams): SpaceSessionGroupMember {
+		const { role, orderIndex = 0, agentId, status = 'active' } = params;
+
 		const existing = this.db
 			.prepare(`SELECT * FROM space_session_group_members WHERE group_id = ? AND session_id = ?`)
 			.get(groupId, sessionId) as Record<string, unknown> | undefined;
@@ -192,6 +203,63 @@ export class SpaceSessionGroupRepository {
 			.run(now, groupId);
 
 		return this.getMember(id)!;
+	}
+
+	/**
+	 * Update fields on an existing group member (e.g. transition status after session completes).
+	 * Returns null if the member record does not exist.
+	 */
+	updateMember(
+		groupId: string,
+		sessionId: string,
+		params: UpdateMemberParams
+	): SpaceSessionGroupMember | null {
+		const fields: string[] = [];
+		const values: (string | number | null)[] = [];
+
+		if (params.role !== undefined) {
+			fields.push('role = ?');
+			values.push(params.role);
+		}
+		if (params.orderIndex !== undefined) {
+			fields.push('order_index = ?');
+			values.push(params.orderIndex);
+		}
+		if (params.agentId !== undefined) {
+			fields.push('agent_id = ?');
+			values.push(params.agentId ?? null);
+		}
+		if (params.status !== undefined) {
+			fields.push('status = ?');
+			values.push(params.status);
+		}
+
+		if (fields.length === 0) {
+			// Nothing to update — return current state
+			const row = this.db
+				.prepare(`SELECT * FROM space_session_group_members WHERE group_id = ? AND session_id = ?`)
+				.get(groupId, sessionId) as Record<string, unknown> | undefined;
+			return row ? this.rowToMember(row) : null;
+		}
+
+		values.push(groupId, sessionId);
+		const result = this.db
+			.prepare(
+				`UPDATE space_session_group_members SET ${fields.join(', ')} WHERE group_id = ? AND session_id = ?`
+			)
+			.run(...values);
+
+		if (result.changes === 0) return null;
+
+		// Touch group updated_at
+		this.db
+			.prepare(`UPDATE space_session_groups SET updated_at = ? WHERE id = ?`)
+			.run(Date.now(), groupId);
+
+		const updated = this.db
+			.prepare(`SELECT * FROM space_session_group_members WHERE group_id = ? AND session_id = ?`)
+			.get(groupId, sessionId) as Record<string, unknown> | undefined;
+		return updated ? this.rowToMember(updated) : null;
 	}
 
 	/**
--- a/packages/daemon/src/storage/repositories/space-session-group-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-session-group-repository.ts
@@ -14,6 +14,7 @@ export interface CreateSessionGroupParams {
 	description?: string;
 	workflowRunId?: string;
 	currentStepId?: string;
+	taskId?: string;
 }
 
 export interface UpdateSessionGroupParams {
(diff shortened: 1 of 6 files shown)


lsm (commented):

### Summary

The type changes and repository updates are cleanly implemented and tests cover the new fields well. However, there is one blocking issue: the production database migration is missing.

---

### P0: Missing database migration for new columns and constraint changes

**Files affected:**
- `packages/daemon/src/storage/schema/migrations.ts` (lines 1700-1740)
- `packages/daemon/tests/unit/helpers/space-test-db.ts` (updated, but production schema was not)

The test helper DB schema (`space-test-db.ts`) was correctly updated with `task_id`, `agent_id`, `status`, and the removal of the `CHECK(role IN ('worker', 'leader'))` constraint. However, the production migration was **not** updated:

1. **Migration 29** (line 1703) still creates `space_session_groups` **without** the `task_id` column.
2. **Migration 29** (line 1722) still creates `space_session_group_members` **without** `agent_id` or `status` columns, and **with** `CHECK(role IN ('worker', 'leader'))`.

This means:
- On any existing database, inserting a member with `role = 'coder'` (or any freeform role) will fail with a CHECK constraint violation.
- Querying `agent_id`, `status`, or `task_id` columns will fail with "no such column" errors.
- On a fresh database, the same failures occur because migration 29 creates tables without these columns.

A new migration (e.g., migration 40) is needed that:
1. `ALTER TABLE space_session_groups ADD COLUMN task_id TEXT`
2. `ALTER TABLE space_session_group_members ADD COLUMN agent_id TEXT`
3. `ALTER TABLE space_session_group_members ADD COLUMN status TEXT NOT NULL DEFAULT 'active' CHECK(status IN ('active', 'completed', 'failed'))`
4. Rebuilds `space_session_group_members` to drop the `CHECK(role IN ('worker', 'leader'))` constraint (SQLite table-rebuild pattern, same as migration 39 does for task status)

Additionally, migration 29's `CREATE TABLE IF NOT EXISTS` block should be updated for fresh databases to include all new columns (the pattern used by other migrations — see the comments at lines 1695-1697 about goal_id being added by migration 34).

---

### P2: `getGroupsByTask` not updated to use `task_id` column

**File:** `packages/daemon/src/storage/repositories/space-session-group-repository.ts` (lines 90-102)

Now that `SpaceSessionGroup` has a `taskId` field, the `getGroupsByTask` method should query by `task_id = ?` rather than relying on the name convention `task:{taskId}`. The current implementation still uses name-based lookup, which makes the new `taskId` field somewhat redundant for this query path. This isn't strictly a bug (the name convention still works), but it's a missed opportunity and could lead to inconsistencies if groups are created with `taskId` set but without the naming convention.

---

### P2: `addMember` positional parameter sprawl

**File:** `packages/daemon/src/storage/repositories/space-session-group-repository.ts` (lines 158-164)

The `addMember` method now takes 6 positional parameters. This is fragile — callers must remember the exact order of `role`, `orderIndex`, `agentId`, `status`. An options object pattern (e.g., `addMember(groupId, sessionId, opts: { role, orderIndex?, agentId?, status? })`) would be more maintainable and less error-prone. This is consistent with the `CreateSessionGroupParams` / `UpdateSessionGroupParams` pattern already used for groups.

---

### P3: JSDoc not updated for `addMember`

**File:** `packages/daemon/src/storage/repositories/space-session-group-repository.ts` (line 156)

The JSDoc says "updates role/orderIndex if the session is already a member" but now also updates `agentId` and `status`.

---

**Recommendation: REQUEST_CHANGES** — the missing production migration (P0) will cause runtime failures on any real database.

lsm (commented):

---

## Summary

The shared-type and repository changes are logically sound, but the production database migration was not updated. The schema changes exist only in the test helper and will not be applied to any production or existing development database. This will cause runtime crashes when the new repository code runs against the production schema.

---

## P0 — Blocking: Missing Production Database Migration

This is the critical gap in the PR.

**What the PR changed in** `packages/daemon/tests/unit/helpers/space-test-db.ts`:
- `space_session_group_members` now has `agent_id TEXT`, `status TEXT NOT NULL DEFAULT 'active' CHECK(status IN ('active','completed','failed'))`, and `role` is a plain `TEXT NOT NULL` (no CHECK constraint).
- `space_session_groups` now has a `task_id TEXT` column.

**What the production migration in** `packages/daemon/src/storage/schema/migrations.ts` **still says (migration 29, lines 1722–1733)**:

```sql
CREATE TABLE IF NOT EXISTS space_session_group_members (
    id TEXT PRIMARY KEY,
    group_id TEXT NOT NULL,
    session_id TEXT NOT NULL,
    role TEXT NOT NULL
        CHECK(role IN ('worker', 'leader')),   -- old hard constraint
    order_index INTEGER NOT NULL DEFAULT 0,
    created_at INTEGER NOT NULL,
    -- NO agent_id column
    -- NO status column
    FOREIGN KEY ...
)
```

`space_session_groups` is also missing `task_id` (lines 1703–1714).

**Consequences on a production or existing dev database:**

1. `addMember(..., 'security-auditor')` → SQLite raises `CHECK constraint failed: role IN ('worker', 'leader')`. Runtime crash, data not written.
2. `INSERT INTO space_session_group_members (..., agent_id, status, ...)` → `table space_session_group_members has no column named agent_id`. Runtime crash.
3. `INSERT INTO space_session_groups (..., task_id, ...)` → `table space_session_groups has no column named task_id`. Runtime crash.
4. `rowToMember` reads `row.agent_id` and `row.status` — on the old schema these return `undefined`, so `status` would silently default to `'active'` from the fallback `?? 'active'`, masking the schema mismatch rather than failing loudly.

**Required fix:** Add a new migration (e.g. migration 40) that:
- Rebuilds `space_session_group_members` to remove the `CHECK(role IN ('worker','leader'))` constraint and add `agent_id TEXT` and `status TEXT NOT NULL DEFAULT 'active' CHECK(status IN ('active','completed','failed'))`.
- Adds `task_id TEXT` to `space_session_groups` via `ALTER TABLE`.

The pattern for removing a CHECK constraint (table-rebuild) is well-established in this codebase — see migration 29's handling of `space_agents` at lines 1791–1837 and migration 39's handling of `space_tasks`.

---

## P1 — Should-fix: No `updateMemberStatus` Method

The `status` field on `SpaceSessionGroupMember` is explicitly documented as "Current state of this member's session" and is expected to transition (`active → completed | failed`). However, the only way to update it is via `addMember`, which is idempotent-upsert and was designed for membership management, not lifecycle transitions.

Callers that need to mark a member as `completed` or `failed` after the session finishes have no clean API — they'd need to call `addMember` with the current role/orderIndex/agentId just to flip the status. This is error-prone and the method's name doesn't signal intent. A dedicated `updateMemberStatus(memberId, status)` or `updateMember(groupId, sessionId, params)` method is needed before this feature is used.

---

## P1 — Should-fix: Test Helper Comment Is Out of Sync

`packages/daemon/tests/unit/helpers/space-test-db.ts` line 8 says:

```
 * Keep in sync with runMigration37 in packages/daemon/src/storage/schema/migrations.ts
```

This comment needs to reference the new migration number (e.g. migration 40) once it's added. More importantly, the fact that the test helper is ahead of the production migration is exactly what caused this divergence — the comment should explicitly warn that the test DB must never define schema that does not yet exist in a migration.

---

## P2 — Suggestion: `getGroupsByTask` Uses Name Convention, Not `task_id` Column

`getGroupsByTask` at line 90–102 of the repository queries by name pattern (`name = 'task:{taskId}'`), completely ignoring the new `task_id` column added to the schema. Now that the column exists, the query should use `WHERE task_id = ?` directly, which is more correct and would allow groups to have any name while still being associated with a task.

---

## P3 — Nit: Migration 29 Comment in `migrations.ts` Describes Old Schema

The migration 29 comment at line 1487 says:
```
 * - space_session_groups: named groups of related sessions (includes workflow_run_id, current_step_id)
```
This should be updated (in the new migration) to mention `task_id` and the new member fields.

lsm (commented):

**Recommendation: APPROVE**

### Re-review Summary

All issues from the previous review have been addressed:

1. **P0 (previously blocking): Missing database migration** -- Resolved. Migration 40 (`packages/daemon/src/storage/schema/migrations.ts`, lines 2355-2418) correctly rebuilds `space_session_group_members` using the SQLite table-rebuild pattern (consistent with migrations 18 and 39), removes the `CHECK(role IN ('worker', 'leader'))` constraint, adds `agent_id` and `status` columns, and adds `task_id` to `space_session_groups` via ALTER TABLE. The migration is idempotent: it detects whether the old CHECK constraint exists before rebuilding, and checks for column existence before ALTER TABLE.

2. **P0 (previously blocking): Fresh DB schema mismatch** -- Resolved. Migration 29 CREATE TABLE statements (lines 1706-1741) now include `task_id`, `agent_id`, `status`, and freeform `role` from the start for fresh databases.

3. **P2 (previously): `getGroupsByTask` using name convention** -- Resolved. Now queries `WHERE space_id = ? AND task_id = ?` directly (line 110).

4. **P2 (previously): `addMember` positional parameter sprawl** -- Resolved. Refactored to use `AddMemberParams` options object (line 174).

5. **P1 (previously): No `updateMember` method** -- Resolved. `updateMember(groupId, sessionId, params)` added (lines 212-263) with proper partial update support and group `updated_at` touch.

6. **P1 (previously): Test helper comment out of sync** -- Resolved. Comment now references migration 40 and includes the warning about schema divergence (lines 7-12).

### New Observations

- **Migration 40 tests** (`packages/daemon/tests/unit/storage/migrations/migration-40_test.ts`): 10 tests covering fresh DB columns, constraint enforcement, legacy DB migration path, data preservation, and idempotency. All pass.
- **Repository tests** updated to use the new `AddMemberParams` API and cover `updateMember`, `getGroupsByTask`, and `taskId` CRUD.
- All 57 tests across the 3 affected test files pass.
- Typecheck passes cleanly.
- The migration pattern (no explicit BEGIN/COMMIT, PRAGMA foreign_keys OFF in try/finally) is consistent with existing migrations 18 and 39.

No new issues found.

lsm (commented):

---

**Recommendation: APPROVE** — all previous feedback items have been addressed correctly. A few minor observations below.

## Summary

This follow-up commit addresses every issue raised in the prior review: migration 40 is present and correct, migration 29's CREATE TABLE statements are updated for fresh installs, `addMember` uses an options object, `updateMember` is added, `getGroupsByTask` queries by column, and 10 migration tests plus updated repository tests are included. The implementation is solid.

## P3 — Missing index on `task_id` (minor, not blocking)

`getGroupsByTask` queries `WHERE space_id = ? AND task_id = ?`, but there is no index on `task_id` in `space_session_groups`. The existing `idx_space_session_groups_space_id` index is used for `space_id`, but a composite index on `(space_id, task_id)` would make this query a covering index hit instead of a filtered scan. At current expected table sizes this is unlikely to matter, but it would be consistent with how similar lookup columns are indexed elsewhere (e.g. `idx_space_tasks_workflow_run_id`). This could be added in a follow-up migration or directly here.

**File:** `packages/daemon/src/storage/schema/migrations.ts` (migration 29 / migration 40)

---

## P3 — Weak timestamp assertion in `updateMember` test

The test `touches group updated_at` asserts `expect(after).toBeGreaterThanOrEqual(before)`. Because a 5 ms sleep precedes the call, the assertion would pass even if `updated_at` was not touched at all (the value would just be equal rather than strictly greater). Changing this to `toBeGreaterThan(before)` would be a stronger verification that the column was actually written.

**File:** `packages/daemon/tests/unit/storage/space-session-group-repository.test.ts` line 252

---

## What was verified

- Migration 40 correctly detects the old schema via `sqlite_master.sql` string match, turns off FK enforcement, rebuilds the table in a `try/finally`, re-creates indexes, and is idempotent on a fresh DB (where the detection condition evaluates to false).
- Migration 29's `CREATE TABLE` statements now include `task_id` on `space_session_groups` and `agent_id`/`status` on `space_session_group_members` from the start. No ALTER TABLE needed on fresh installs.
- `AddMemberParams` options object cleanly replaces positional args; `UpdateMemberParams` provides correct partial-update semantics with the `fields.length === 0` no-op path.
- `getGroupsByTask` now queries by the `task_id` column rather than by name pattern.
- `SpaceSessionGroupMember.role` is now a freeform `string`; `agentId` and `status` fields are properly typed and mapped in `rowToMember`.
- 10 migration-40 tests cover: fresh DB column existence, status CHECK enforcement, role freeform acceptance, legacy DB rebuild path, data preservation, and idempotency.
- No callers of the old positional `addMember` signature exist in production code — the repository is only wired through tests at this stage.

PR #617, opened 2026-03-21
feat: add channel and multi-agent step validation to SpaceWorkflowManager

lsm:
- Extend SpaceAgentLookup to include `role` field so channel role
  validation can resolve agent IDs to their SpaceAgent.role values
- Move agents[].agentId format validation (non-empty string check)
  outside the agentLookup block so it always runs regardless of lookup
- Add validateStepChannels: structural checks (direction enum, non-empty
  from/to) always run; role-reference checks run when agentLookup is set
  and validates from/to against known roles or '*' wildcard
- Channels on a single-agent step (no agents[]) are rejected
- Update agentLookup implementations in index.ts and test helpers to
  include role field in returned object
- Add 20 new tests covering: agents[] format validation, channel
  structural validation, channel role validation with/without lookup,
  and multi-agent step CRUD round-trips via the manager

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -206,7 +206,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		getAgentById(spaceId: string, id: string) {
 			const agent = spaceAgentRepo.getById(id);
 			if (!agent || agent.spaceId !== spaceId) return null;
-			return { id: agent.id, name: agent.name };
+			return { id: agent.id, name: agent.name, role: agent.role };
 		},
 	};
 	const spaceWorkflowManager = new SpaceWorkflowManager(spaceWorkflowRepo, agentLookup);
--- a/packages/daemon/src/lib/space/managers/space-workflow-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-workflow-manager.ts
@@ -32,7 +32,7 @@ import type { SpaceWorkflowRepository } from '../../../storage/repositories/spac
  */
 export interface SpaceAgentLookup {
 	/** Returns the SpaceAgent with the given UUID in the given space, or null if not found. */
-	getAgentById(spaceId: string, id: string): { id: string; name: string } | null;
+	getAgentById(spaceId: string, id: string): { id: string; name: string; role: string } | null;
 }
 
 // ---------------------------------------------------------------------------
@@ -187,6 +187,19 @@ export class SpaceWorkflowManager {
 			);
 		}
 
+		// Format-level validation: always run regardless of agentLookup
+		if (hasAgents) {
+			for (let j = 0; j < step.agents!.length; j++) {
+				const entry = step.agents![j];
+				if (!entry.agentId || !entry.agentId.trim()) {
+					throw new WorkflowValidationError(
+						`step[${index}].agents[${j}]: agentId must be a non-empty SpaceAgent UUID`
+					);
+				}
+			}
+		}
+
+		// Existence validation: only when agentLookup is available
 		if (this.agentLookup) {
 			if (hasAgentId) {
 				const agent = this.agentLookup.getAgentById(spaceId, step.agentId!);
@@ -199,11 +212,6 @@ export class SpaceWorkflowManager {
 			if (hasAgents) {
 				for (let j = 0; j < step.agents!.length; j++) {
 					const entry = step.agents![j];
-					if (!entry.agentId || !entry.agentId.trim()) {
-						throw new WorkflowValidationError(
-							`step[${index}].agents[${j}]: agentId must be a non-empty SpaceAgent UUID`
-						);
-					}
 					const agent = this.agentLookup.getAgentById(spaceId, entry.agentId);
 					if (!agent) {
 						throw new WorkflowValidationError(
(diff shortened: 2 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly adds channel and multi-agent step validation to `SpaceWorkflowManager`. The implementation is correct, well-structured, and thoroughly tested.

### What was reviewed

1. **`SpaceAgentLookup` interface extension** (`space-workflow-manager.ts:35`) — added `role: string` to the return type. All callers updated consistently (`index.ts`, two test helpers).

2. **`validateStepAgentRef` restructuring** (`space-workflow-manager.ts:190-200`) — the format-level `agentId` emptiness check now runs unconditionally before the `agentLookup` existence check. This is a good separation: structural validation always runs, existence validation is optional.

3. **`validateStepChannels`** (`space-workflow-manager.ts:231-300`) — validates:
   - Channels require multi-agent `agents[]` format (single-agent steps rejected)
   - Direction must be `'one-way'` or `'bidirectional'`
   - `from` must be non-empty string
   - `to` handles both string and `string[]` forms, rejects empty values in both
   - Role references validated against known agent roles when `agentLookup` is available
   - `*` wildcard correctly bypasses role matching

4. **Repository persistence** — confirmed that `space-workflow-repository.ts` already stores `agents` and `channels` in the JSON `config` column (lines 425-430) and deserializes them back (lines 115-119). The round-trip tests verify this end-to-end.

5. **20 new tests** — cover format validation (empty/whitespace agentIds), channel structural validation (direction, from, to), role validation with/without lookup, wildcard handling, CRUD round-trips, and backwards compatibility with single-agent workflows.

- P3: The `seedAgent` helper hardcodes `role: 'coder'` for all agents (line 54 of test file), which could be confusing when tests seed agents with names like 'ReviewerAgent'. However, since the role-validation tests use custom mock lookups rather than DB queries, this has no functional impact.

- P3: The `knownRoles` set could end up empty if all agent lookups fail (agents exist in `agents[]` but not found via lookup). In that case, all non-wildcard role references would be rejected with "known roles: none", which is reasonable behavior since those agents already failed existence validation earlier.

The code fits existing patterns well, handles edge cases properly, and the test coverage is thorough.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (see P1 issue below)

---

### Overview

The implementation correctly extends `SpaceAgentLookup` with `role`, restructures the `agents[].agentId` format check so it runs without a lookup, and adds `validateStepChannels` with structural and role-reference validation. The repository serialization/deserialization (introduced in the previous PR) correctly handles both new fields. The 20 new test cases cover the main happy/error paths. A few issues worth addressing:

---

### P1 — Dead code creates a fragile role-collection gap

**File:** `packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, lines 245–251

```ts
for (const agentEntry of step.agents) {
    if (agentEntry.agentId && agentEntry.agentId.trim()) {
        const agent = this.agentLookup.getAgentById(spaceId, agentEntry.agentId);
        if (agent) {           // ← this guard is dead code
            knownRoles.add(agent.role);
        }
    }
}
```

By the time `validateStepChannels` is called, `validateStepAgentRef` has already thrown `WorkflowValidationError` for any agent whose lookup returns `null`. The `if (agent)` guard can never be `false` at this point — if it were, execution would have stopped earlier. This has a real consequence: if the lookup ever returns `null` here despite having returned a non-null value during existence validation (e.g., due to an intermediary stateful lookup or if the logic is refactored), the affected agent's role is silently dropped from `knownRoles`, turning role-reference validation from a hard error into a false negative (valid role strings would be rejected as 'unknown'). The guard should either be removed (trusting the prior validation) or, if defensive, should re-throw a `WorkflowValidationError` rather than silently skipping.

---

### P2 — Missing test coverage for whitespace-only channel string fields

The code correctly checks `!ch.from.trim()`, `!ch.to[ti].trim()`, and `!(ch.to as string).trim()`, but there are no test cases for whitespace-only strings in those positions. Three paths are tested for empty strings but not for their whitespace equivalents:

- `from: '   '` (whitespace-only)
- `to: '   '` (whitespace-only string)
- `to: ['   ']` (whitespace-only element in array)

These are handled correctly by existing code but remain untested. Compare with the `agents[].agentId` section which has separate tests for both `''` and `'   '`.

---

### P2 — Missing test: `agents: []` (explicitly empty array) with channels

The test at line 1192 covers the case where channels appear on a step with `agentId` (single-agent format). It does not cover `agents: []` (an explicitly empty array). The guard at line 235 handles both:

```ts
if (!step.agents || step.agents.length === 0)
```

But only the `!step.agents` branch is exercised. A test for `agents: [], channels: [...]}` would ensure the `length === 0` branch is covered.

---

### P2 — Duplicate agent lookup in channel validation

`validateStepChannels` at lines 245–252 re-queries `agentLookup.getAgentById` for every agent in `step.agents` to collect roles. These same calls were already made in `validateStepAgentRef` (lines 213–221) moments earlier. For a step with N agents and M channels this is 2N lookup calls. This is not a correctness bug but it does add unnecessary work, and the duplication makes the code harder to maintain. Consider collecting roles as part of the first traversal and passing the result into `validateStepChannels`, or at a minimum documenting why the second traversal is intentional.

---

### P3 — `existingStepInputs` rebuild omits `agents` and `channels`

**File:** `packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, lines 114–119

When only `transitions` are updated (no step replacement), the code rebuilds `existingStepInputs` from `existing.steps` but does not include `agents` or `channels`:

```ts
const existingStepInputs: WorkflowStepInput[] = existing.steps.map((s) => ({
    id: s.id,
    name: s.name,
    agentId: s.agentId,
    instructions: s.instructions,
    // agents and channels are missing
}));
```

This is used only for `validateTransitions` which validates step IDs — so no functional bug today. But if `validateSteps` were ever called from this path (e.g., during a refactor), multi-agent steps would be incorrectly rejected as missing both `agentId` and `agents`. This is a pre-existing omission that predates this PR, but now that `agents` and `channels` are first-class step fields it is worth fixing to keep the mapping complete. Note: this is pre-existing, not introduced by this PR.

---

### Correctness observations (no action required)

- The role deduplication via `Set` is correct behavior: two agents with the same role still means that role is valid.
- The `channels: []` (empty array) case is handled correctly by the `step.channels.length > 0` guard and the repository's own `if (input.channels && input.channels.length > 0)` check — empty arrays are neither validated nor persisted.
- Backwards compatibility is confirmed by the 'legacy single-agent workflow continues to work alongside multi-agent workflows' test.
- The `role` field added to `SpaceAgentLookup` is a minimal, non-breaking extension of the interface — existing implementations only needed to add `role: agent.role` to the return object.

lsm (commented):

### Re-review: All previously raised issues are properly addressed

**P1 fix (dead-code `if (agent)` guard / duplicate lookups):** The `validateStepAgentRef` method now collects `knownRoles` during the agent existence loop (lines 216-228 in `space-workflow-manager.ts`) and passes the set directly to `validateStepChannels`. This eliminates the second traversal and the previously silent lookup failure. The format-level validation (empty/whitespace agentId check) was correctly moved before the existence check (lines 193-201), so it runs regardless of whether `agentLookup` is configured.

**P2 fix (whitespace-only channel fields):** Three new tests added — whitespace-only `from`, whitespace-only string `to`, and whitespace-only element in array `to`. All validate correctly via the `!ch.from.trim()` / `!ch.to[ti].trim()` checks.

**P2 fix (`agents: []` + channels edge case):** Test `createWorkflow rejects channels when agents[] is explicitly empty` covers this. The `validateStepChannels` method correctly checks `!step.agents || step.agents.length === 0` at line 249.

**P2 fix (duplicate agent lookup):** Resolved by the P1 fix — single pass through `step.agents!` collects both existence validation and roles.

**P3 fix (`existingStepInputs` rebuild):** Lines 118-119 in the `updateWorkflow` path now include `agents` and `channels` when reconstructing `WorkflowStepInput` from existing steps.

### Additional observations on the updated code

The implementation is clean and well-structured:

- `SpaceAgentLookup` interface correctly extended with `role: string` (line 35), and all three call sites (`index.ts`, `space-export-import-handlers.test.ts`, `provision-global-agent.test.ts`) updated consistently.
- `knownRoles` is `null` when `agentLookup` is not configured, and the channel role-reference check is correctly gated on `knownRoles !== null` (line 292) — this means structural validation always runs but role-reference validation is only enforced when the lookup is available.
- The wildcard `'*'` is properly handled in `validateChannelRoleRef` (line 303).
- 30 new tests provide thorough coverage: format validation without lookup, channel structural validation, role-reference validation with lookup, CRUD round-trips, mixed single/multi-agent coexistence, and update/delete paths.
- All 92 tests in `space-workflow.test.ts` pass, along with the 83 tests across the other two modified test files.

**Recommendation: APPROVE** — All previously identified issues are resolved and no new issues found.

lsm (commented):

**Recommendation: APPROVE**

All previously raised issues have been correctly addressed. Here is a point-by-point verification:

---

### P1 resolved: Dead-code `if (agent)` guard eliminated

`validateStepAgentRef` now collects `knownRoles` in a single pass during the agents existence loop (lines 217-227 of `space-workflow-manager.ts`). The `Set<string>` is initialized only when `agentLookup` is set and `hasAgents` is true. The set is then passed directly into `validateStepChannels`. There is no second lookup, and there is no silent drop on failed lookup — a missing agent immediately throws before `knownRoles.add()` is ever reached.

The parameter to `validateStepChannels` is typed `Set<string> | null` with clear JSDoc explaining that `null` means role-reference checks are skipped (when `agentLookup` is not configured). This is clean and intentional.

---

### P2 resolved: Whitespace-only channel string fields tested

Three new tests cover the previously untested whitespace paths:
- `createWorkflow rejects channels with whitespace-only from` (line 1226)
- `createWorkflow rejects channels with whitespace-only string to` (line 1242)
- `createWorkflow rejects channels with whitespace-only element in array to` (line 1258)

All three call the `.trim()` branches already in `validateStepChannels` (lines 269, 286-287, 281).

---

### P2 resolved: `agents: []` with channels edge case tested

The test `createWorkflow rejects channels when agents[] is explicitly empty` (line 1208) covers the `step.agents.length === 0` branch in `validateStepChannels` (line 249). The step fixture combines `agentId: 'agent-a'` (so the "at least one agent" check passes) with `agents: []` and `channels`, which correctly triggers the guard.

---

### P2 resolved: Duplicate agent lookup eliminated

The single-pass approach described above eliminates the 2N lookup pattern. Confirmed by code inspection — there is exactly one `getAgentById` call per entry in `step.agents[]`.

---

### P3 resolved: `existingStepInputs` now includes `agents` and `channels`

Lines 114-121 of `space-workflow-manager.ts` now map `s.agents` and `s.channels` into the `existingStepInputs` array used during the transitions-only update path. The mapping is complete.

---

### Additional observations (no new issues)

- All 92 unit tests pass cleanly.
- The `SpaceAgentLookup` interface correctly exposes `role: string` and both the production implementation in `index.ts` (line 209) and the test helpers in `provision-global-agent.test.ts` (line 222) and `space-export-import-handlers.test.ts` (line 217) all return the `role` field.
- The `seedAgent` helper in `space-workflow.test.ts` (line 55) inserts `role` as `'coder'` for all test agents. Channel role tests that need distinct roles use inline `SpaceAgentLookup` mocks returning specific role strings, which is correct.
- Channel validation on single-agent `agentId`-only steps is properly rejected (test at line 1192) and the code path at line 249 handles both `!step.agents` (undefined/null) and `step.agents.length === 0`.

PR #618, opened 2026-03-22
feat: extend listJobs to accept status arrays

lsm:
Allow JobQueueRepository.listJobs() status filter to accept either a
single JobStatus string or an array of JobStatus values. Array inputs
generate IN (?,?,...) SQL; empty array returns [] early.

### DIFF
--- a/packages/daemon/src/storage/repositories/job-queue-repository.ts
+++ b/packages/daemon/src/storage/repositories/job-queue-repository.ts
@@ -125,7 +125,11 @@ export class JobQueueRepository {
 		return this.rowToJob(row);
 	}
 
-	listJobs(filter: { queue?: string; status?: JobStatus; limit?: number }): Job[] {
+	listJobs(filter: { queue?: string; status?: JobStatus | JobStatus[]; limit?: number }): Job[] {
+		if (Array.isArray(filter.status) && filter.status.length === 0) {
+			return [];
+		}
+
 		let query = `SELECT * FROM job_queue WHERE 1=1`;
 		const params: (string | number)[] = [];
 
@@ -134,8 +138,14 @@ export class JobQueueRepository {
 			params.push(filter.queue);
 		}
 		if (filter.status !== undefined) {
-			query += ` AND status = ?`;
-			params.push(filter.status);
+			if (Array.isArray(filter.status)) {
+				const placeholders = filter.status.map(() => '?').join(',');
+				query += ` AND status IN (${placeholders})`;
+				params.push(...filter.status);
+			} else {
+				query += ` AND status = ?`;
+				params.push(filter.status);
+			}
 		}
 
 		query += ` ORDER BY created_at DESC LIMIT ?`;
--- a/packages/daemon/tests/unit/storage/job-queue-repository.test.ts
+++ b/packages/daemon/tests/unit/storage/job-queue-repository.test.ts
@@ -416,6 +416,51 @@ describe('JobQueueRepository', () => {
 			expect(jobs.length).toBe(3);
 		});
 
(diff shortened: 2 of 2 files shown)


lsm (commented):

Clean, well-scoped change. The implementation is correct and complete.

**What was reviewed:**
- `packages/daemon/src/storage/repositories/job-queue-repository.ts` — type signature change (`status?: JobStatus | JobStatus[]`), early return for empty array, branching SQL generation for array vs string
- `packages/daemon/tests/unit/storage/job-queue-repository.test.ts` — 4 new test cases

**Findings:**

The implementation correctly handles all cases:
1. **Empty array** — early return before query construction (line 129-131), avoiding an invalid `IN ()` SQL clause
2. **Array with values** — dynamic placeholder generation via `.map(() => '?').join(',')` with spread into params (lines 141-144)
3. **Single string** — unchanged path (lines 145-148), preserving backward compatibility
4. **Undefined** — no status clause added, unchanged behavior

The parameterized query approach properly prevents SQL injection since status values are passed as bound parameters, not interpolated into the query string. The `JobStatus` type constraint further limits inputs to the known union at the TypeScript level.

All 50 tests pass (4 new + 46 existing). Test coverage is adequate — multi-status array, single-element array, empty array, and backward-compatible string usage are all exercised.

No issues found. **Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

The implementation is correct and complete. The changes are minimal, well-scoped, and accomplish exactly what was requested.

---

### Implementation

**`packages/daemon/src/storage/repositories/job-queue-repository.ts` (lines 128–156)**

The logic is clean and correct:

- The type signature `status?: JobStatus | JobStatus[]` is the right union — no unnecessary wrappers.
- The empty-array early return (line 129) is the right design choice: it short-circuits before building or executing any SQL, avoids the degenerate `IN ()` form that some SQLite versions reject, and communicates intent clearly.
- The `IN (?,?,...)` placeholder generation via `filter.status.map(() => '?').join(',')` is the standard, safe approach for dynamic-length parameterized queries. No SQL injection surface.
- The spread `params.push(...filter.status)` correctly flattens the array into the positional params list.
- The string fallback branch is unchanged, preserving full backward compatibility.

No issues with the production code.

---

### Tests

The 4 new tests in `packages/daemon/tests/unit/storage/job-queue-repository.test.ts` (lines 419–462) cover:

1. Multi-element array returning matching jobs from multiple statuses.
2. Single-element array (exercises the `IN (?)` path rather than the `= ?` path, which is meaningfully distinct).
3. Empty array returning `[]` via the early-return path.
4. String input still working (backward compat regression test).

All four cases are the right tests to write. The assertions are specific and verify actual DB behavior, not just that the method was called.

One minor observation: the "string status still works" test (line 453) enqueues 2 jobs and dequeues 1, leaving 1 pending and 1 processing, then asserts `listJobs({ status: 'pending' }).length === 1`. The unused variable `toProcess` from the original "filters by status" test at line 382 is still there in that older test (the variable is assigned but its value is never read after the enqueue — harmless but worth noting). This is pre-existing and not introduced by this PR.

---

### No issues found

- No SQL injection risk (parameterized throughout).
- No edge case gaps (empty array, single-element array, multi-element, and string path are all covered).
- No missing error handling (the function is purely query-and-map; errors propagate naturally from bun:sqlite as exceptions, consistent with the rest of the class).
- No type safety gaps (TypeScript will reject anything that is not `JobStatus | JobStatus[] | undefined` at compile time).
- The change has no callers yet in the `packages/` source tree (only referenced in docs/plans), so there is no risk of breaking existing production call sites.

PR #620, opened 2026-03-22
feat: extend export/import format for multi-agent steps and channels

lsm:
- Add ExportedWorkflowStepAgent type and update ExportedWorkflowStep to
  support optional agents[] array and channels[] alongside agentRef
- Make agentRef optional on ExportedWorkflowStep (single-agent shorthand)
- Update exportWorkflow(): multi-agent steps export as agents[] with
  agentRef names; single-agent steps use scalar agentRef (backward compat);
  channels exported as-is (role strings, already portable)
- Update Zod schema for step validation: either agentRef or non-empty agents
  required; add workflowChannelSchema and exportedWorkflowStepAgentSchema
- Update buildWorkflowCreateParams(): resolve each agentRef→agentId for
  multi-agent steps; import channels as-is; backward compat for agentRef
- Update validateWorkflowForPreview(): validate all agent refs in
  multi-agent steps
- Update spaceExport.workflows referencedNames to collect refs from agents[]
- Add 22 new unit tests covering export, validation, round-trip, and import
  of multi-agent + channels steps; backward compat tests for single-agent

### DIFF
--- a/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
+++ b/docs/plans/complete-and-adopt-persistent-job-queue-continue-from-pr-203/00-overview.md
@@ -1,0 +1,116 @@
+# Complete and Adopt Persistent Job Queue -- Continue from PR #203
+
+## Goal
+
+Wire the existing but unused `JobQueueRepository` and `JobQueueProcessor` into the daemon lifecycle, then migrate the three primary `setInterval`-based background task subsystems to the persistent job queue. This makes background work durable across daemon restarts and centralizes scheduling.
+
+## Background
+
+PR #203 produced a comprehensive ADR (`docs/adr/0002-job-queue-migration.md`) over 13 review iterations. The plan is well-specified but zero production code has been implemented. The existing infrastructure includes:
+
+- `JobQueueRepository` (complete, 203 lines) -- CRUD, dequeue, retry, cleanup
+- `JobQueueProcessor` (complete, 123 lines) -- poll-based processor with handler registration
+- `job_queue` SQLite table with indexes
+- Comprehensive unit tests for both
+
+The subsystems to migrate:
+1. **Session background tasks** -- fire-and-forget title generation tracked in `pendingBackgroundTasks` Set
+2. **GitHub polling** -- `setInterval` in `GitHubPollingService` (60s cycle)
+3. **Room runtime tick** -- `setInterval` in `RoomRuntime.start()` (30s cycle), with **17 `scheduleTick()` call sites**, **6 `scheduleTickAfterRateLimitReset()` call sites** (lines 617, 696, 732, 985, 1052, 1084), and **2 `queueMicrotask(() => this.tick())` call sites** (lines 2241, 3421) — totaling ~23 scheduling points that need migration
+
+## Out of Scope
+
+The following `setInterval` users are explicitly **out of scope** for this migration and will remain as-is:
+
+- **`WebSocketServerTransport.staleCheckTimer`** (`packages/daemon/src/lib/websocket-server-transport.ts` line 87) — This is a transport-layer health check (30s stale connection cleanup), not a background business task. It operates at a different abstraction layer and does not benefit from persistence.
+- **`SpaceRuntime.tickTimer`** (`packages/daemon/src/lib/space/runtime/space-runtime.ts` line 226) — SpaceRuntime is a newer/parallel runtime implementation. Its tick loop follows the same pattern as RoomRuntime but is a separate subsystem. It should be migrated in a follow-up once RoomRuntime migration is proven stable.
+- **`TaskAgentManager` polling interval** (`packages/daemon/src/lib/space/runtime/task-agent-manager.ts` line 190) — Part of the SpaceRuntime subsystem, out of scope for the same reason.
+- **`JobQueueProcessor` internal poll timer** (`packages/daemon/src/storage/job-queue-processor.ts` line 38) — This is the job queue infrastructure itself; its internal `setInterval` is the mechanism that drives all job processing and is not a migration target.
+- **`app.ts` shutdown readiness check** (`packages/daemon/src/app.ts` line 367) — One-shot `setInterval` for graceful shutdown polling, not a recurring background task.
+
+Task 6.2 must explicitly verify these exclusions and document them in a code comment for future reference.
+
+## Approach
+
+Continue on the existing PR #203 branch (`plan/persistent-job-queue-adoption`). Implement in 6 milestones, each producing a PR-ready commit. All work targets the `dev` branch.
+
+**Key design decisions:**
+- Handlers self-schedule the next run (no dedicated scheduler service for now)
+- `listJobs` must be extended to accept `status` as an array for deduplication queries
+- Room tick deduplication uses application-level checks (not DB unique index) for simplicity. **Risk acknowledged:** the check-then-enqueue pattern has a small race window. If this proves problematic in practice, a follow-up can add a unique partial index on `(queue, json_extract(payload, '$.roomId')) WHERE status IN ('pending', 'processing')` as the ADR suggests. For now the application-level approach is acceptable because: (a) SQLite serializes writes, limiting the race window; (b) a rare duplicate tick is harmless (tick is idempotent); (c) the dedup check should use "no pending job exists at all for this roomId" (not just "no future-scheduled job"), see detailed logic in Milestone 4.
+- `stopRuntime()` must call `runtimes.delete(roomId)` to fix the liveness check bug
+- **Pause/resume must cancel in-flight tick jobs** — see Milestone 4 for detailed design of pause/resume interaction with the job queue
+
+## Handler Registration and Startup Ordering
+
+All handlers **must** be registered before `jobProcessor.start()` is called. If a pending job exists in the DB from a previous run and no handler is registered for its queue, it will fail with "No handler registered" and consume retry attempts.
+
+**Required startup order in `app.ts`:**
+
+```
+1. sessionManager.start()       → registers 'session.title_generation' handler
+2. gitHubService.start()        → registers 'github.poll' handler + enqueues initial poll
+3. roomRuntimeService.start()   → registers 'room.tick' handler
+4. Register cleanup handler     → registers 'job_queue.cleanup' handler + enqueues initial cleanup
+5. jobProcessor.start()         → begins processing (all handlers now registered)
+```
+
+Each milestone's wiring task must respect this ordering. Task 1.2 must initially call `jobProcessor.start()` as the **last** step, and each subsequent milestone inserts its handler registration before that call.
+
+## maxConcurrent Configuration
+
+The initial `maxConcurrent: 3` is conservative. With multiple active rooms, each generating `room.tick` jobs every 30s, plus GitHub polling and title generation, the 3-slot limit may cause contention — a long-running title generation or poll could block room ticks.
+
+**Mitigation strategy:**
+- Start with `maxConcurrent: 5` instead of 3, based on expected workload (typically 1-3 active rooms + 1 GitHub poll + occasional title gen)
+- The processor's `dequeue` already handles priority via FIFO ordering; room ticks are short-lived and will cycle quickly
+- If contention appears in practice, the follow-up is per-queue concurrency limits (the ADR's open question on priority inversion)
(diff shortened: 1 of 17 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly extends the export/import format for multi-agent workflow steps and channels. The implementation is correct, well-structured, and thoroughly tested.

### What was reviewed

1. **`packages/shared/src/types/space.ts`** — New `ExportedWorkflowStepAgent` type and updated `ExportedWorkflowStep` with optional `agents[]`, `channels[]`, and optional `agentRef`.
2. **`packages/daemon/src/lib/space/export-format.ts`** — Updated `exportWorkflow()` with multi-agent branching, new Zod schemas (`exportedWorkflowStepAgentSchema`, `workflowChannelSchema`), and updated step schema with `superRefine` validation.
3. **`packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`** — Updated `buildWorkflowCreateParams()`, `validateWorkflowForPreview()`, and `spaceExport.workflows` handler to handle multi-agent steps.
4. **Test files** — 22 new tests across both test files covering export, validation, round-trip, import, backward compat, error cases.

### Findings

**No blocking issues found.** The implementation is solid:

- **Backward compatibility** is handled correctly: single-agent steps export as `agentRef` (scalar shorthand), multi-agent steps export as `agents[]`. Both paths are tested.
- **Import resolution** correctly resolves `agentRef` names to UUIDs for both multi-agent (`agents[]`) and single-agent (`agentRef`) steps, checking both imported bundle agents and existing space agents.
- **Channels** are exported/imported as-is since they use role strings, not UUIDs. This is the correct approach.
- **Validation** at the Zod schema level enforces that at least one of `agentRef` or `agents` (non-empty) is present.
- **Preview validation** (`validateWorkflowForPreview`) correctly flags unresolved agent refs in multi-agent steps.
- **Referenced agent collection** in `spaceExport.workflows` correctly collects names from both `agents[]` entries and scalar `agentRef`.
- **Error handling** on import correctly throws on unresolved agent references and rolls back the transaction.

**P3 observation (non-blocking):** The Zod schema does not reject a step that has *both* `agentRef` and `agents` simultaneously, though the type documentation says they are mutually exclusive. This is harmless because both `exportWorkflow()` (which only emits one) and `buildWorkflowCreateParams()` (which prioritizes `agents` over `agentRef`) handle this correctly. The `superRefine` only validates "at least one present," not "exactly one present." This is a defensive design choice that errs on the side of accepting valid-enough input, which is reasonable for an import format.

### Tests

All 158 tests pass (92 in export-format, 66 in import-handlers). Tests cover:
- Multi-agent step export with agent name resolution
- Per-agent instructions preservation
- Channel export/import as-is
- UUID-to-name fallback for unresolved agents
- Round-trip (export -> JSON -> validate) for multi-agent steps
- Import with multi-agent step agent ref resolution
- Unresolved agent ref error handling (both execute and preview)
- Backward compatibility with single-agent `agentRef` format
- Resolution from existing space agents (not just bundle agents)

TypeScript typecheck passes cleanly.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (P1 and P2 issues found)

---

## Summary

The implementation is well-structured and the core logic is correct. The export/import round-trip for multi-agent steps and channels works properly, the Zod schemas are sound, backward compatibility is preserved, and the 22 new tests provide good coverage. However, there are a few important issues to address before merging.

---

## P1 Issues (Should Fix)

### 1. Export produces an empty-string `agentRef` when `step.agentId` is undefined

**File:** `packages/daemon/src/lib/space/export-format.ts`, line 236-237

```ts
const primaryAgentId = step.agentId ?? '';
exported.agentRef = agentIdToName.get(primaryAgentId) ?? primaryAgentId;
```

When `step.agentId` is `undefined` (which is valid in the `WorkflowStep` type if only `agents[]\ is populated and someone constructed the step that way in test data), this falls through to the single-agent branch and writes `agentRef: ''` into the exported JSON. The downstream Zod schema will then reject the bundle at import time with a confusing error — but only at import, not at export. The export should either guard against this case or skip the step when neither `agentId` nor `agents` is set. This is a low-probability production path, but it is reachable if a caller provides a partially-constructed `WorkflowStep` where both `agentId` and `agents` are absent.

### 2. Multi-agent step with resolved-but-empty agentId writes `agentId: ''` into the params map

**File:** `packages/daemon/src/lib/rpc-handlers/space-export-import-handlers.ts`, line 188

```ts
const entry: { agentId: string; instructions?: string } = { agentId: agentId ?? '' };
```

The fail-fast check at line 588 prevents this from reaching the DB, so it does not cause data corruption in practice. However, the intermediate `WorkflowStepInput` that is passed to `createWorkflow` (at line 598) will contain an entry with `agentId: ''` if the throw were ever to be removed or the check refactored. The cleaner pattern is to omit the empty entry and rely on the already-present warning + throw to short-circuit. Not a data-safety bug today because the throw fires before `createWorkflow`, but a latent correctness risk.

### 3. Channel role references are not validated during import preview or export

The task description says the implementation should "validate channel role references" on import. The current `validateWorkflowForPreview` and `buildWorkflowCreateParams` do not check that the roles referenced in `channel.from` / `channel.to` actually match any agent role in the step. The Zod schema only ensures they are non-empty strings. This means a bundle with a typo in a channel role (e.g. `from: 'codr'` instead of `'coder'`) will import silently without any warning. According to the PR description this was in scope: "validate channel role references". Its absence means the feature is partially implemented.

---

## P2 Issues (Important Suggestions)

### 4. Test DB schema for `space_workflow_steps` is missing the `instructions` column

**File:** `packages/daemon/tests/unit/rpc-handlers/space-export-import-handlers.test.ts`, lines 86-99

The test schema creates `space_workflow_steps` with only `id`, `workflow_id`, `name`, `description`, `agent_id`, `order_index`, `config`, `created_at`, `updated_at`. It does not include an `instructions` column. Since the repository stores instructions inside the JSON `config` column (via `StepConfigJson`), this works in practice — but it means the test schema diverges silently from the production migration schema. Any future refactor that moves `instructions` to its own column would break these tests without a clear migration failure. This is a maintenance concern rather than a bug.

### 5. No test for the mixed single+multi-agent step export with channels on the single-agent step

The `exportWorkflow` tests do not test the case where a single-agent step has `channels` set (which is theoretically valid per the type). The export path at line 241 writes channels regardless of whether the step is single- or multi-agent:
```ts
if (step.channels && step.channels.length > 0) exported.channels = step.channels;
```
This code is reached for both single-agent and multi-agent steps. There is no test that exercises a single-agent step that also carries channels, so this branch is only partially covered.

### 6. `makeBundle` helper in the import tests does not support the `channels` field on single-agent steps

**File:** `packages/daemon/tests/unit/rpc-handlers/space-export-import-handlers.test.ts`, lines 1571-1600

The `BundleWorkflow.steps` type is `Array<{ agentRef: string; name: string; instructions?: string }>`. There is no `channels` field. This means there is no test coverage for channels on the backward-compat single-agent path through the execute handler. The multi-agent path is covered via `makeMultiAgentBundle`.

---

## P3 Nits

### 7. The `exportedWorkflowStepAgentSchema` and `workflowChannelSchema` names are declared but not exported

They are purely internal to `export-format.ts`, which is fine. But consider adding a comment to each schema noting what type they validate against, consistent with the rest of the file.

### 8. JSDoc on `ExportedWorkflowStep` says "Exactly one of `agentRef` or `agents` must be present" but the type and Zod schema allow both to be set simultaneously

**File:** `packages/shared/src/types/space.ts`, lines 855-857

The Zod `superRefine` only enforces that at least one is present, not that they are mutually exclusive. The `exportWorkflow` function never produces a step with both set (the if/else is correct), but the type + schema do not reject an import payload where both are present. The JSDoc says "Mutually exclusive" but the schema does not enforce it. Low risk, but the doc is misleading.

---

## Correctness Assessment

The core behavior is correct:
- Multi-agent steps export as `agents[]` with `agentRef` names, not UUIDs
- Single-agent steps export as scalar `agentRef` (backward compatible)
- Channels are exported and imported as-is (role strings, already portable)
- Import resolves each `agentRef` → `agentId` correctly (bundle agents first, then space agents)
- Transaction rollback on unresolved refs works correctly
- The referenced-agent set for `spaceExport.workflows` correctly collects refs from both the `agents[]` and scalar `agentRef` paths

The implementation is functionally sound for the primary use cases. The P1 items above represent gaps in completeness (missing validation for channel role refs, edge case in export of invalid steps) and a latent code smell, not production-breaking bugs in the common path.

lsm (commented):

Re-reviewed after the fix commit (cfad21ce). All five previously reported issues have been addressed correctly:

**P1 fixes verified:**
1. **Empty-string agentRef** (`export-format.ts:240-243`): Export now guards with `if (primaryAgentId)` before setting `agentRef`. The `?? ''` fallback is removed. Steps with neither `agentId` nor `agents` will correctly fail Zod validation with the clear message "step must have either agentRef or agents".

2. **Empty-string agentId placeholder** (`space-export-import-handlers.ts:189-190`): Comment now explicitly explains that `agentId ?? ''` is safe because `warnings.length > 0` causes a throw before `createWorkflow` is called. This is correct -- the warning is pushed on line 186, and the throw happens at line 633-639.

3. **Channel role reference validation** (`space-export-import-handlers.ts:269-325`): `validateWorkflowForPreview()` now accepts an `agentNameToRole` map and validates channel `from`/`to` roles against the roles of agents assigned to each step. The `'*'` wildcard is properly excluded. The `agentNameToRole` map is built from both existing space agents and bundle agents (lines 446-447), with bundle agents correctly taking precedence.

**P2 fix verified:**
4. **Single-agent step with channels test** (`export-format.test.ts:1309-1333`): New test "single-agent step with channels exports channels" verifies that channels are preserved on single-agent steps.

**P3 fix verified:**
5. **JSDoc on ExportedWorkflowStep** (`space.ts:864-869`): Updated from "Exactly one of" to "At least one of" and added clarification about import-time precedence when both are present.

**Additional new tests added (good):**
- Preview: flags invalid channel roles not matching step agents
- Preview: wildcard `'*'` channel role is always valid
- Preview: validates channel roles from existing space agents
- Resolves multi-agent step refs from existing space agents
- Backward compat: single agentRef step still imports as agentId

All 94 export-format tests and 69 RPC handler tests pass.

**Recommendation: APPROVE** -- All previously reported issues are resolved correctly. The implementation is clean, well-documented, and thoroughly tested.

lsm (commented):

---

**Recommendation: APPROVE**

All five issues from the previous review have been correctly addressed. The implementation is solid and tests pass clean (94 export-format tests, 69 handler tests, 0 failures).

### Verified Fixes

**P1 — Empty-string agentRef in exported JSON (FIXED)**
`exportWorkflow()` in `export-format.ts` now only sets `agentRef` when `step.agentId` is truthy (line 241–243). A step with neither `agentId` nor `agents` produces neither field, which causes a clear Zod error on validation rather than a silent empty-string.

**P1 — Latent empty-string agentId placeholder in `buildWorkflowCreateParams` (FIXED)**
The comment at line 189–190 of `space-export-import-handlers.ts` now explicitly documents that the `'' ` placeholder is safe because unresolved refs throw before `createWorkflow` is called.

**P1 — Missing channel role validation during preview (FIXED)**
`validateWorkflowForPreview()` now accepts a fourth `agentNameToRole: Map<string, string>` parameter and validates all non-`'*'` roles in `channel.from` and `channel.to` (including fan-out array `to`). The `agentNameToRole` map is built in `spaceImport.preview` with bundle agents overriding existing agents on name collision, which is the correct priority order.

**P2 — Single-agent step with channels test (FIXED)**
Test `'single-agent step with channels exports channels'` added at line 1309 of `export-format.test.ts`. It verifies that single-agent steps still export as scalar `agentRef` while channels are preserved.

**P3 — JSDoc on ExportedWorkflowStep (FIXED)**
The comment was updated from "Exactly one" to "At least one must be present" and the full JSDoc at line 864 of `space.ts` now correctly reads "At least one of `agentRef` or `agents` (non-empty) must be present".

### Remaining Minor Observation

There is no integration test in `space-export-import-handlers.test.ts` for channel role validation on a **single-agent step** via `spaceImport.preview` (e.g. a step with `agentRef: 'Coder'` and a channel referencing `'bad-role'`). The code path in `validateWorkflowForPreview` does cover this case — single-agent refs are pushed into `stepAgentRefs` and channel roles are checked against them — but there is no test exercising it end-to-end through the RPC handler. This is a minor gap (P3) given that the export-format unit tests do cover the validation logic directly and all current tests pass.

PR #624, opened 2026-03-22
test: add export/import round-trip tests for multi-agent workflows

lsm:
- Fix failing test: add missing 'role' field to agentLookup in multi-agent
  test setup (required by validateChannelRoles in SpaceWorkflowManager)
- Add 'exportBundle' import so full export→import round-trip tests can
  use the real export function rather than hand-crafted bundles
- Add 'full export→import round-trip' describe block with 8 new tests:
  - Single-agent workflow: verifies step.agentId, rule.appliesTo UUID remapping
  - Multi-agent step: verifies agents[] resolution, per-agent instructions,
    shared instructions, and channel preservation
  - Channel topology: one-way, bidirectional fan-out (array `to`), wildcard (*)
  - Mixed single/multi-agent workflow: both step types in one workflow
  - Backward compat: old-style single agentId exports via exportBundle and
    re-imports as scalar agentId (not agents[])
  - Error: unknown agentRef rolls back transaction and clears all DB state
  - Error: channel role not in imported agents produces preview validation error

### DIFF
--- a/packages/daemon/tests/unit/rpc-handlers/space-export-import-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc-handlers/space-export-import-handlers.test.ts
@@ -1886,6 +1886,7 @@ describe('full export→import round-trip', () => {
 	let spaceManager: SpaceManager;
 	let handlers: Map<string, RequestHandler>;
 	let daemonHub: DaemonHub;
+	let emittedEvents: Array<{ name: string; data: unknown }>;
 
 	beforeEach(() => {
 		db = new Database(':memory:');
@@ -1909,6 +1910,7 @@ describe('full export→import round-trip', () => {
 		handlers = mockHub.handlers;
 		const mockDaemonHub = createMockDaemonHub();
 		daemonHub = mockDaemonHub.hub;
+		emittedEvents = mockDaemonHub.emittedEvents;
 
 		setupSpaceExportImportHandlers(
 			mockHub.hub,
@@ -2002,6 +2004,14 @@ describe('full export→import round-trip', () => {
 		const rule = importedWf.rules[0];
 		expect(rule.name).toBe('Tests must pass');
 		expect(rule.appliesTo).toEqual([step.id]);
+
+		// Events emitted for real-time frontend updates
+		const agentCreatedEvents = emittedEvents.filter((e) => e.name === 'spaceAgent.created');
+		const wfCreatedEvents = emittedEvents.filter((e) => e.name === 'spaceWorkflow.created');
+		expect(agentCreatedEvents).toHaveLength(1);
+		expect(wfCreatedEvents).toHaveLength(1);
+		expect((agentCreatedEvents[0].data as any).agent.name).toBe('My Coder');
+		expect((wfCreatedEvents[0].data as any).workflow.name).toBe('Code Pipeline');
 	});
 
 	it('multi-agent step round-trip: export → import preserves agents array and channels', async () => {
@@ -2328,8 +2338,11 @@ describe('full export→import round-trip', () => {
 		expect(collabStep.channels).toHaveLength(1);
 		expect(collabStep.channels![0].direction).toBe('one-way');
 
-		// Transition preserved
+		// Transition preserved with remapped step UUID endpoints
 		expect(importedWf.transitions).toHaveLength(1);
+		const transition = importedWf.transitions[0];
+		expect(transition.from).toBe(planStep.id);
+		expect(transition.to).toBe(collabStep.id);
 		expect(importedWf.startStepId).toBe(planStep.id);
 
 		// Tags preserved
@@ -2462,5 +2475,53 @@ describe('full export→import round-trip', () => {
 
 		expect(preview.validationErrors.length).toBeGreaterThan(0);
 		expect(preview.validationErrors.some((e) => e.includes('nonexistent-role'))).toBe(true);
+	});
+
+	it('error: execute rejects workflow with invalid channel role and rolls back', async () => {
+		// This test verifies that spaceImport.execute — not just preview — enforces
+		// channel role validation via SpaceWorkflowManager.createWorkflow().
+		// A regression that bypasses validateChannelRoleRef in the execute path
+		// would leave the DB in a partial state; this test catches that.
+		const agentSrc: SpaceAgent = {
+			id: 'src-exec-solo',
+			spaceId: 'other-space',
+			name: 'Exec Coder',
+			role: 'coder',
+			createdAt: 1000,
+			updatedAt: 2000,
+		};
+		const wfSrc: SpaceWorkflow = {
+			id: 'src-wf-exec-badch',
+			spaceId: 'other-space',
+			name: 'Exec Bad Channel Workflow',
+			steps: [
+				{
+					id: 'step-exec-bc',
+					name: 'Work',
+					agents: [{ agentId: 'src-exec-solo' }],
+					channels: [
+						// 'bad-exec-role' does not match the agent's role 'coder'
+						{ from: 'bad-exec-role', to: 'coder', direction: 'one-way' },
+					],
+				},
+			],
+			transitions: [],
+			startStepId: 'step-exec-bc',
+			rules: [],
+			tags: [],
+			createdAt: 1000,
+			updatedAt: 2000,
+		};
+
+		const bundle = exportBundle([agentSrc], [wfSrc], 'Exec Bad Channel Export');
+
+		// execute must throw — WorkflowValidationError from validateChannelRoleRef
+		await expect(
+			call(handlers, 'spaceImport.execute', { spaceId: SPACE_ID, bundle })
+		).rejects.toThrow();
+
+		// Transaction rolled back: agent was created then rolled back along with workflow
+		expect(agentRepo.getBySpaceId(SPACE_ID)).toHaveLength(0);
+		expect(workflowRepo.listWorkflows(SPACE_ID)).toHaveLength(0);
 	});
 });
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

This PR correctly fixes a real bug (missing `role` field in `agentLookup` for the `multi-agent step import` describe block) and adds 8 well-structured round-trip tests that exercise the full `exportBundle()` -> `spaceImport.execute` pipeline. The tests cover single-agent, multi-agent, various channel topologies (one-way, fan-out, wildcard), mixed workflows, backward compatibility, and two error cases. All 78 tests in the file pass.

### Bug Fix (P3 - already correct)

The fix at line 1315 adding `role: agent.role` to the `agentLookup` return is correct. The `SpaceAgentLookup` interface requires `{ id: string; name: string; role: string }` (see `space-workflow-manager.ts:35`), and the missing `role` caused `validateChannelRoles` to fail. The fix is consistent with the other two `beforeEach` blocks (lines 218 and 1902) which both include `role`.

### Test Quality

**Strengths:**
- Tests use the real `exportBundle()` function rather than hand-crafted bundles, making them true round-trip tests
- Good coverage of channel topologies: one-way, bidirectional, fan-out (array `to`), and wildcard (`*`)
- Error cases verify both transaction rollback (DB state cleared) and preview validation errors
- The backward compatibility test confirms old-style `agentId` exports don't get incorrectly converted to `agents[]`
- Assertions verify UUID remapping (new UUIDs differ from source UUIDs)

**Minor observations (P3):**

1. **Mixed workflow transition `from`/`to` not fully verified** (line 2332): The test checks `transitions` has length 1 and `startStepId` maps correctly, but does not verify `transition.from === planStep.id` and `transition.to === collabStep.id`. This is a minor gap since transition remapping is implicitly tested by the workflow being created successfully (the manager validates transition endpoints), but an explicit assertion would be more complete.

2. **No test for `conflictResolution` strategies in round-trip context**: The existing test suite already covers conflict resolution for single-agent imports. The new round-trip tests only cover the `create` (no conflict) path. A round-trip test that imports twice and exercises `replace` or `rename` with multi-agent steps would add value, but is not a gap in the stated scope.

3. **Channel role validation in mixed workflow**: The channel `{ from: 'coder', to: 'reviewer', direction: 'one-way' }` at line 2294 uses roles that match `coderAgent2.role = 'coder'` and `reviewAgent.role = 'reviewer'`, which is correct. Good attention to detail here.

No P0, P1, or P2 issues found. The implementation is clean, well-scoped, and the tests effectively validate the round-trip behavior.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — P1 test gap in the bad-channel-role error scenario; P2 gaps in transition content assertions and event emission coverage.

---

### Fix to existing failing test

The one-line fix adding `role: agent.role` to `getAgentById` in the `multi-agent step import` `beforeEach` is correct and necessary. `SpaceWorkflowManager.validateStepChannels` calls `validateChannelRoleRef` using roles collected from `agentLookup.getAgentById`, and the lookup was returning `{ id, name }` without `role`, causing channel role validation to always see an empty `knownRoles` set. The fix is minimal and accurate.

---

### New round-trip tests — issues found

#### P1: Bad-channel-role error test only covers `preview`, not `execute`

**File:** `packages/daemon/tests/unit/rpc-handlers/space-export-import-handlers.test.ts`

The test `'error: channel role not present in imported agents produces preview validation error'` calls `spaceImport.preview` and checks `validationErrors`. It does **not** call `spaceImport.execute` to verify that `execute` also rejects invalid channel roles.

The actual validation path in `execute` goes through `workflowManager.createWorkflow()` → `validateStep()` → `validateChannelRoleRef()`. A regression that bypasses channel validation in the execute path (or in `SpaceWorkflowManager`) would not be caught by this test. The round-trip test suite should include an `execute`-level assertion that an invalid channel role throws and rolls back — analogous to how the unknown-agentRef test verifies both the throw and the DB rollback.

#### P2: Mixed-workflow test does not assert transition `from`/`to` step IDs

The test `'mixed single/multi-agent workflow round-trip preserves both step types'` checks `importedWf.transitions.toHaveLength(1)` and `importedWf.startStepId`, but never asserts that `transition.from` and `transition.to` map to the correct new step UUIDs. The remapping of transition endpoints (step name → new UUID) is the most critical part of the export→import pipeline for multi-step workflows, and this gap would miss a regression in `buildWorkflowCreateParams`'s transition-remapping logic.

Suggested additions:
```ts
const transition = importedWf.transitions[0];
expect(transition.from).toBe(planStep.id);
expect(transition.to).toBe(collabStep.id);
```

#### P2: Event emissions not asserted in round-trip block

The main `describe` block tests that `spaceAgent.created` and `spaceWorkflow.created` are emitted after import. The new `full export→import round-trip` `describe` block captures a `daemonHub` mock but never uses it for assertions. At minimum the happy-path single-agent test should verify events are emitted, since round-trip is the closest to the real production flow. Currently, a regression that silently drops event emissions after a round-trip import would go undetected in this suite.

#### P3: The "single-agent round-trip" and "backward compat" tests are largely duplicates

The `'single-agent workflow round-trip'` test and `'backward compat: single agentRef export → import via exportBundle'` tests both verify that a workflow with `agentId` (single-agent) exports and re-imports as scalar `agentId`, not `agents[]`. The only difference is the backward-compat test omits `systemPrompt`/`tools`/`description`/`config` assertions. The first test is the more comprehensive one; the second adds marginal coverage. This is not a blocking issue, just minor redundancy.

---

### Summary

The fix to the existing failing test is correct. The new round-trip test suite covers the main happy-path scenarios well (single-agent, multi-agent, channel topologies, mixed workflows) and two error cases. The primary concern is that the bad-channel-role error test only exercises the `preview` path — if `execute`'s channel validation path were broken, no round-trip test would catch it. The transition endpoint UUID remapping is also unverified in the mixed-workflow test, which is the exact scenario where a remapping bug would surface.

lsm (commented):

**Recommendation: APPROVE** — All three previously raised concerns are fully addressed. 79/79 tests pass.

---

### Follow-up verification

This is a follow-up review of PR #624 after the worker addressed feedback from the previous round.

#### P1 resolved: `execute`-level bad-channel-role test added

A new test `'error: execute rejects workflow with invalid channel role and rolls back'` (line 2480) calls `spaceImport.execute` with a bundle containing `bad-exec-role` — a role not present on the imported agent. The test verifies:
1. `execute` rejects with a thrown error (`rejects.toThrow()`)
2. The transaction rolled back: `agentRepo.getBySpaceId(SPACE_ID)` returns length 0 and `workflowRepo.listWorkflows(SPACE_ID)` returns length 0

This directly covers the execute-path channel validation that was missing. The concern is fully addressed.

#### P2 resolved: Transition `from`/`to` UUID remapping assertions added

The mixed-workflow test now includes at lines 2342–2345:
```ts
const transition = importedWf.transitions[0];
expect(transition.from).toBe(planStep.id);
expect(transition.to).toBe(collabStep.id);
```

These assertions verify that `buildWorkflowCreateParams` correctly remaps transition endpoints to new step UUIDs after import. The concern is fully addressed.

#### P2 resolved: Event emission assertions wired into round-trip block

The `full export→import round-trip` describe block now declares `emittedEvents` (line 1889) and wires it up in `beforeEach` (line 1913). The single-agent round-trip test asserts at lines 2009–2014:
```ts
const agentCreatedEvents = emittedEvents.filter((e) => e.name === 'spaceAgent.created');
const wfCreatedEvents = emittedEvents.filter((e) => e.name === 'spaceWorkflow.created');
expect(agentCreatedEvents).toHaveLength(1);
expect(wfCreatedEvents).toHaveLength(1);
expect((agentCreatedEvents[0].data as any).agent.name).toBe('My Coder');
expect((wfCreatedEvents[0].data as any).workflow.name).toBe('Code Pipeline');
```

This is the right level of coverage — the happy-path round-trip verifies both emission counts and payload correctness. The concern is fully addressed.

---

### Test results

All 79 tests pass in 251ms with no failures.

No new issues found. The implementation is clean and the stated concerns have been addressed correctly.

lsm (commented):

**Recommendation: APPROVE**

### Summary

All three issues raised in the previous Sonnet review have been properly addressed:

1. **P1 resolved: execute-level bad-channel-role test** (lines 2480-2526) — New test `'error: execute rejects workflow with invalid channel role and rolls back'` exercises the `spaceImport.execute` path with an invalid channel role, verifying both that it throws and that the transaction rolls back (no agents or workflows left in DB). The `.rejects.toThrow()` assertion is slightly loose (no error message match), but this is acceptable since the important behavior being tested is the rollback, and the specific error message depends on the internal validation implementation.

2. **P2 resolved: transition from/to step ID remapping assertions** (lines 2342-2345) — The mixed-workflow round-trip test now asserts that `transition.from === planStep.id` and `transition.to === collabStep.id`, confirming step ID remapping in transitions works correctly during import.

3. **P2 resolved: emittedEvents tracking in round-trip tests** (lines 1889, 1913, 2009-2014) — The `emittedEvents` array is now wired into the `full export->import round-trip` describe block's `beforeEach`, and the single-agent round-trip test asserts that `spaceAgent.created` and `spaceWorkflow.created` events are emitted with correct data.

### Test quality

All 79 tests pass with 223 assertions. The test file is comprehensive, covering:
- Handler registration and input validation
- All export handlers (agents, workflows, bundle) with filtering
- Import preview (conflict detection, validation errors, unresolved refs, channel role validation)
- Import execute (all conflict resolutions: skip/rename/replace, cross-ref mapping, step ID remapping, rule appliesTo remapping)
- Multi-agent step import (agent resolution, per-agent instructions, channels, backward compat)
- Transaction atomicity (rollback on failure)
- Event emission for all action types
- Full round-trip tests using real `exportBundle()` through `spaceImport.execute`
- Channel topology variants (one-way, fan-out, wildcard, bidirectional)

No remaining issues found.

PR #628, opened 2026-03-22
feat: wire SpaceSessionGroupRepository into TaskAgentManager

lsm:
- Add sessionGroupRepo field to TaskAgentManagerConfig
- In spawnTaskAgent(), create a SpaceSessionGroup named task:<taskId>
  with taskId set and add the Task Agent session as a member with
  role 'task-agent' and status 'active'
- Store groupId in taskGroupIds map (taskId → groupId) for fast lookup
- Expose getTaskGroupId() for future sub-session wiring
- Clean up taskGroupIds entry in cleanup()
- Construct SpaceSessionGroupRepository once in rpc-handlers/index.ts
  and inject it into TaskAgentManager
- Add 4 new unit tests covering group creation, getTaskGroupId, and
  cleanup behavior (67 → 102 tests passing across 2 files)

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -61,6 +61,7 @@ import { SpaceWorkflowRepository } from '../../storage/repositories/space-workfl
 import { SpaceAgentRepository } from '../../storage/repositories/space-agent-repository';
 import type { JobQueueRepository } from '../../storage/repositories/job-queue-repository';
 import type { JobQueueProcessor } from '../../storage/job-queue-processor';
+import { SpaceSessionGroupRepository } from '../../storage/repositories/space-session-group-repository';
 import { SpaceRuntimeService } from '../space/runtime/space-runtime-service';
 import { setupSpaceWorkflowRunHandlers } from './space-workflow-run-handlers';
 import type { SpaceWorkflowRunTaskManagerFactory } from './space-workflow-run-handlers';
@@ -270,6 +271,10 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		taskRepo: spaceTaskRepo,
 	});
 
+	// SpaceSessionGroupRepository — persists session groups for Task Agents and sub-sessions.
+	// Constructed once here and injected into TaskAgentManager.
+	const spaceSessionGroupRepo = new SpaceSessionGroupRepository(deps.db.getDatabase());
+
 	// Task Agent Manager — manages Task Agent session lifecycle and message injection.
 	// Must be created after spaceRuntimeService so it can get WorkflowExecutors via
 	// spaceRuntimeService.createOrGetRuntime(spaceId).
@@ -286,6 +291,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		messageHub: deps.messageHub,
 		getApiKey: () => deps.authManager.getCurrentApiKey(),
 		defaultModel: deps.config.defaultModel,
+		sessionGroupRepo: spaceSessionGroupRepo,
 	});
 
 	// Wire TaskAgentManager into the SpaceRuntime so the tick loop can spawn
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -58,6 +58,7 @@ import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceRuntimeService } from './space-runtime-service';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
+import type { SpaceSessionGroupRepository } from '../../../storage/repositories/space-session-group-repository';
 import type { SubSessionFactory, SubSessionState } from '../tools/task-agent-tools';
 import { createTaskAgentMcpServer } from '../tools/task-agent-tools';
 import { createTaskAgentInit, buildTaskAgentInitialMessage } from '../agents/task-agent';
@@ -95,6 +96,8 @@ export interface TaskAgentManagerConfig {
 	getApiKey: () => Promise<string | null>;
 	/** Default model ID for sessions that don't specify one */
 	defaultModel: string;
+	/** Session group repository — used to persist SpaceSessionGroup records per task */
+	sessionGroupRepo: SpaceSessionGroupRepository;
 }
 
 // ---------------------------------------------------------------------------
@@ -127,6 +130,12 @@ export class TaskAgentManager {
 	 * spawnTaskAgent() calls from creating duplicate Task Agent sessions.
 	 */
 	private spawningTasks = new Set<string>();
+
+	/**
+	 * Maps taskId → groupId for SpaceSessionGroups created when spawning Task Agents.
+	 * Allows fast lookup when adding sub-session members to an existing group.
+	 */
+	private taskGroupIds = new Map<string, string>();
 
 	/**
 	 * Completion callbacks registered via onComplete().
@@ -284,6 +293,24 @@ export class TaskAgentManager {
 			// --- Store in map before streaming start to allow getTaskAgent() calls
 			this.taskAgentSessions.set(taskId, agentSession);
 
+			// --- Create a SpaceSessionGroup for this task and add the Task Agent as member
+			try {
+				const group = this.config.sessionGroupRepo.createGroup({
+					spaceId,
+					name: `task:${taskId}`,
+					taskId,
+				});
+				this.config.sessionGroupRepo.addMember(group.id, sessionId, {
+					role: 'task-agent',
+					status: 'active',
+				});
+				this.taskGroupIds.set(taskId, group.id);
+				log.info(`TaskAgentManager: created session group ${group.id} for task ${taskId}`);
+			} catch (err) {
+				// Group creation failure is non-fatal — task agent still runs; log and continue
+				log.warn(`TaskAgentManager: failed to create session group for task ${taskId}:`, err);
+			}
+
 			// --- Register in SessionManager cache so getSessionAsync() returns the live
 			// instance with MCP tools attached rather than creating a duplicate from DB.
 			// Without this, RPC handlers (message.send, message.sdkMessages, etc.) would
@@ -423,6 +450,11 @@ export class TaskAgentManager {
 		return this.taskAgentSessions.get(taskId);
 	}
 
+	/** Returns the session group ID for a task, or undefined if no group was created. */
+	getTaskGroupId(taskId: string): string | undefined {
+		return this.taskGroupIds.get(taskId);
+	}
+
 	/** Returns a sub-session by its session ID, or undefined if not found. */
 	getSubSession(subSessionId: string): AgentSession | undefined {
 		for (const [, stepMap] of this.subSessions) {
@@ -550,6 +582,9 @@ export class TaskAgentManager {
 				this.sessionListeners.delete(sessionId);
 			}
 		}
+
+		// 5. Remove group ID lookup
+		this.taskGroupIds.delete(taskId);
 
 		log.info(`TaskAgentManager: cleaned up all sessions for task ${taskId}`);
 	}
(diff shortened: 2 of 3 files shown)


lsm (commented):

The implementation correctly wires `SpaceSessionGroupRepository` into `TaskAgentManager` and the core happy path works well. All 67 tests pass. However, there are two issues worth addressing:

### P1: `rehydrate()` does not restore `taskGroupIds` map

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, `rehydrateTaskAgent()` (line ~803)

After a daemon restart, `rehydrate()` restores `taskAgentSessions` and `subSessions` from the DB, but it never restores the `taskGroupIds` map. This means `getTaskGroupId(taskId)` returns `undefined` for all rehydrated tasks, even though the session group exists in the database.

Any downstream code that calls `getTaskGroupId()` to add sub-session members to the group (the stated purpose of the map per the doc comment on line 134-136) will silently fail after a restart.

**Fix:** In `rehydrateTaskAgent()`, query `sessionGroupRepo.getGroupsByTask(spaceId, taskId)` and restore the mapping, e.g.:
```ts
const groups = this.config.sessionGroupRepo.getGroupsByTask(spaceId, taskId);
if (groups.length > 0) {
    this.taskGroupIds.set(taskId, groups[0].id);
}
```

### P2: No test for group creation failure (non-fatal path)

**File:** `packages/daemon/tests/unit/space/task-agent-manager.test.ts`

The implementation explicitly handles `createGroup()` throwing an error as non-fatal (line 309-312). This is a deliberate design decision that deserves a test to prevent regression -- e.g., mock `createGroup` to throw and verify the task agent still spawns successfully and `getTaskGroupId()` returns `undefined`.

### P2: `cleanup()` only removes the in-memory map entry, not the DB group

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 587

`cleanup()` deletes the `taskGroupIds` entry but does not update the group status to `'completed'`/`'failed'` or delete it from the database. The group and its members remain in `space_session_groups` / `space_session_group_members` as `'active'` even after all sessions are destroyed. This may cause stale data to appear in queries like `getGroupsByTask()` or `getGroupsBySpace()`.

Consider calling `sessionGroupRepo.updateGroup(groupId, { status: 'completed' })` or `sessionGroupRepo.deleteGroup(groupId)` during cleanup.

### Summary

The wiring and happy-path logic are correct and well-structured. The main gap is that the `taskGroupIds` map is not restored during rehydration, which will cause the getter to return `undefined` after daemon restarts -- defeating the purpose of the in-memory cache for sub-session member tracking.

lsm (commented):

The wiring is correct at a high level — the repo is constructed once, injected cleanly, group creation happens at the right point in the spawn flow, and the basic happy-path tests are solid. However there are two meaningful gaps.

---

### P1 — Rehydration does not restore `taskGroupIds`

`rehydrateTaskAgent()` (lines 803–935 of `task-agent-manager.ts`) rebuilds `taskAgentSessions` and `subSessions` after a daemon restart, but it never populates `taskGroupIds`. After rehydration, `getTaskGroupId(taskId)` returns `undefined` for every task even though the group record exists in the DB. Any caller that depends on the fast-lookup map (for adding sub-session members, for example) will silently miss the group rather than finding it.

The fix is straightforward: query `getGroupsByTask(spaceId, taskId)` inside `rehydrateTaskAgent()` and, if a group is found, repopulate the map entry. There is no test covering the post-rehydration value of `getTaskGroupId`, which is why this went unnoticed.

---

### P1 — Partial write when `addMember` throws after `createGroup` succeeds

In the group creation block (lines 297–312 of `task-agent-manager.ts`):

```ts
try {
    const group = this.config.sessionGroupRepo.createGroup(...);  // commits to DB
    this.config.sessionGroupRepo.addMember(group.id, sessionId, {...}); // may throw
    this.taskGroupIds.set(taskId, group.id);
} catch (err) {
    log.warn(...);
}
```

If `addMember` throws, the group row is persisted in `space_session_groups` with no member. The catch swallows the error silently, and `taskGroupIds` is never set, so the orphaned group is invisible to any subsequent lookup. On the next spawn call (e.g., after a retry), a second group will be created for the same `taskId`, leaving two memberless-or-partial records in the DB.

Wrapping both calls in a single repository-level transaction, or at least deleting the group if `addMember` fails, would prevent the orphan. A test simulating an `addMember` failure would catch this regression.

---

### P2 — No test for the non-fatal (warn) code path

The implementation explicitly states that group creation failure is non-fatal and logs a warning. This is important behavior — it means a task agent runs even if its group record was not created. There is no test that mocks `createGroup` to throw and asserts that `spawnTaskAgent` still succeeds and returns a session ID. This leaves the error-handling branch untested.

---

### P2 — `cleanup()` does not remove the DB group record

The task description says to store `groupId` in-memory for fast lookup. `cleanup()` correctly deletes the in-memory entry. However, the DB group record is left alive after `cleanup()` is called. If the group should track the lifecycle of the task agent (i.e., be marked `completed` or `failed` when the task ends), that transition is never made. Whether this is intentional should be clarified with a comment; as written it is at least surprising.

---

### P3 — Doc comment on `TaskAgentManager` header not updated

The class doc comment (lines 14–37 of `task-agent-manager.ts`) lists the three in-memory maps (`taskAgentSessions`, `subSessions`, `spawningTasks`). The new `taskGroupIds` map is not listed there, making the reference slightly stale.

lsm (commented):

### Re-review: All prior issues resolved

I verified that all P0/P1/P2 issues from the previous review have been addressed:

**Prior P1: rehydrate() does not restore taskGroupIds from DB** -- RESOLVED
- `rehydrateTaskAgent()` now calls `this.config.sessionGroupRepo.getGroupsByTask(spaceId, taskId)` at line 955 and populates `taskGroupIds` map. Test at line 1746 confirms this works after restart.

**Prior P1: partial write race between createGroup and addMember** -- RESOLVED
- Inner try/catch at lines 307-321 catches addMember failures and calls `deleteGroup()` to clean up the orphan. The secondary `deleteGroup` failure is also caught (best-effort). Test at line 498 verifies no orphaned group rows remain.

**Prior P2: cleanup() does not update DB group record** -- RESOLVED
- `cleanup()` now calls `updateGroup(groupId, { status: 'completed' })` at line 604 before deleting the in-memory map entry. Test at line 457 verifies both the in-memory cleanup and DB status transition.

**Prior P2: missing test coverage for non-fatal path and failures** -- RESOLVED
- Six new tests added covering: group creation in DB, getTaskGroupId accessor, cleanup marking group completed, createGroup throwing (non-fatal), addMember throwing (no orphan), and rehydration restoring taskGroupIds.

### Current assessment

The implementation is clean and well-integrated:

1. **Wiring** (`index.ts`): `SpaceSessionGroupRepository` is constructed once with the DB handle and injected into `TaskAgentManager` config. Straightforward.

2. **Spawn path** (`task-agent-manager.ts` lines 297-326): Group creation + member addition with proper nested error handling. The non-fatal design is appropriate -- a session group record is metadata, not critical to task execution.

3. **Cleanup path** (lines 600-609): Updates group status to 'completed' with error handling, then clears in-memory map. Correctly placed after session teardown.

4. **Rehydration** (lines 954-958): Restores `taskGroupIds` from DB. Takes `existingGroups[0]` which is correct since there should be exactly one group per task (ordered by created_at ASC).

5. **Tests**: All 70 tests pass. The new tests use the real `SpaceSessionGroupRepository` against an in-memory SQLite DB, which is good -- no mocking of the repository layer.

No remaining issues found. **Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

All four previously-raised issues have been resolved correctly. The implementation is clean and the test suite is thorough. Findings below.

---

### Previously Raised Issues — Status

**P1 (was): `rehydrate()` does not restore `taskGroupIds` from DB — FIXED**

`rehydrateTaskAgent()` now calls `getGroupsByTask(spaceId, taskId)` at line 955 and populates `taskGroupIds` with `existingGroups[0].id`. The corresponding test `taskGroupIds is restored from DB after rehydration` (line 1746) seeds a group in the real SQLite DB and verifies `getTaskGroupId()` returns the persisted ID after rehydration. The fix is correct.

One minor observation: the code uses `existingGroups[0].id` (the oldest group by `created_at ASC`) when there are multiple groups for a task. This is a sensible default given the schema, and aligns with what `spawnTaskAgent()` would create (only one group per spawn). No action needed.

**P1 (was): Partial write race — orphaned group when `addMember` throws — FIXED**

The inner try/catch at lines 307–321 catches `addMember` failures, calls `deleteGroup(group.id)`, and re-throws so the outer catch can log and suppress the error (non-fatal path). The test `no orphaned group when addMember throws after createGroup succeeds` (line 498) patches `addMember` to throw and verifies `getGroupsByTask` returns an empty array. Both the production fix and the test are correct.

The secondary-failure swallow inside the inner catch (`} catch { /* ignore */ }`) is appropriate since `deleteGroup` failure shouldn't mask the original `addMember` error that gets re-thrown.

**P2 (was): `cleanup()` does not update DB group record — FIXED**

Step 5 of `cleanup()` (lines 600–609) calls `updateGroup(groupId, { status: 'completed' })` before deleting the in-memory entry. The call is wrapped in a try/catch so the in-memory cleanup still proceeds even if the DB write fails. The test `cleanup removes taskGroupId from in-memory map and marks group completed` (line 457) spawns a task, captures the groupId, runs cleanup, then asserts both the map is cleared and the DB row has `status === 'completed'`. Correct.

**P2 (was): Missing test coverage for non-fatal path and failures — FIXED**

Four new tests exercise the group-creation paths:
- `creates a SpaceSessionGroup in DB with taskId set` — happy path including member role/status
- `getTaskGroupId returns the group ID after spawn`
- `spawn still succeeds when createGroup throws (non-fatal)`
- `no orphaned group when addMember throws after createGroup succeeds`

Plus the rehydration test `taskGroupIds is restored from DB after rehydration`.

Coverage is adequate for the new behavior.

---

### New Observations

**P3 — `daemonHub` missing from `rehydrateTaskAgent()` MCP server call**

At line 873–889, `createTaskAgentMcpServer` is called without `daemonHub` in the options. The call in `spawnTaskAgent` (lines 263–280) passes `daemonHub: this.config.daemonHub`. If the MCP server silently tolerates a missing `daemonHub` this is harmless, but it is an inconsistency worth noting. This is pre-existing and outside the scope of this PR, so no action needed here.

**P3 — Rehydrate picks `existingGroups[0]` silently when multiple groups exist**

If a task somehow ends up with multiple groups (e.g., a prior partially-cleaned daemon run that didn't complete cleanup), the code silently picks the oldest one. There is no log line indicating multiple groups were found. This is unlikely in practice and is a latent cosmetic concern, not a blocking issue.

---

### Summary

All P0/P1/P2 issues from the prior review are resolved. The implementation is correct, the DB interactions are properly guarded, and the new tests provide meaningful coverage including the failure paths. All 70 unit tests pass.

lsm (commented):

**Recommendation: APPROVE**

The merge conflict resolution in `packages/daemon/src/lib/rpc-handlers/index.ts` is correct. Both sides of the conflict were preserved cleanly and without regression.

### What was verified

**The `JobQueueRepository`/`JobQueueProcessor` imports from `dev` (lines 62–63)**

These are imported as types only:
```ts
import type { JobQueueRepository } from '../../storage/repositories/job-queue-repository';
import type { JobQueueProcessor } from '../../storage/job-queue-processor';
```
They appear correctly in the `RPCHandlerDependencies` interface as `jobQueue` and `jobProcessor` fields, and `app.ts` passes real instances to `setupRPCHandlers()`.

**The `SpaceSessionGroupRepository` import from this PR (line 64)**

```ts
import { SpaceSessionGroupRepository } from '../../storage/repositories/space-session-group-repository';
```
It is instantiated at line 276 and passed as `sessionGroupRepo` to `TaskAgentManager`. All four methods called on it (`createGroup`, `addMember`, `deleteGroup`, `updateGroup`, `getGroupsByTask`) exist in the repository implementation.

**Integration chain is complete and consistent:**

- `app.ts` creates `JobQueueRepository`/`JobQueueProcessor` and passes them to `setupRPCHandlers`
- `setupRPCHandlers` instantiates `SpaceSessionGroupRepository` and injects it into `TaskAgentManager`
- `TaskAgentManager` uses `sessionGroupRepo` for group lifecycle during `spawnTaskAgent` and `cleanup`
- Unit tests in `task-agent-manager.test.ts` cover the new group creation, idempotency, and cleanup paths including the non-fatal fallback when `createGroup` or `addMember` throws

No regressions detected. The conflict resolution correctly merges both feature sets.

PR #629, opened 2026-03-22
feat: show associated goal on room task list and task view

lsm:
- Add `currentRoomTabSignal` to signals.ts for cross-component tab navigation
- Add `goalByTaskId` computed signal to room-store for reverse task→goal lookup
- Show emerald goal badge on tasks in RoomTasks list with goal title and tooltip
- Clicking goal badge in task list switches to Missions tab via signal
- Show goal badge in TaskView header with mission title and navigation
- Clicking goal badge in TaskView navigates back to room and switches to Missions tab
- Room.tsx watches currentRoomTabSignal to switch activeTab when signal changes
- Unit tests: goal badge in RoomTasks (8 tests) and TaskView (4 tests)
- E2E tests: task list badge, task view badge, and tab navigation flows

### DIFF
--- a/packages/e2e/tests/features/task-goal-indicator.e2e.ts
+++ b/packages/e2e/tests/features/task-goal-indicator.e2e.ts
@@ -1,0 +1,218 @@
+/**
+ * Task Goal Indicator E2E Tests
+ *
+ * Tests that tasks linked to a goal show the goal name badge:
+ * - Goal badge appears in task list (RoomDashboard) when task is linked to a goal
+ * - Goal badge appears in TaskView header when task is linked to a goal
+ * - Clicking goal badge in task list switches to Missions tab
+ * - Clicking goal badge in TaskView navigates back to room and switches to Missions tab
+ * - Tasks without a goal do NOT show the badge
+ *
+ * Setup: RPC to create room, task, and goal and link them together (infrastructure).
+ * Test actions and assertions: all through visible UI.
+ * Teardown: RPC room delete.
+ */
+
+import { test, expect } from '../../fixtures';
+import { waitForWebSocketConnected } from '../helpers/wait-helpers';
+import { deleteRoom } from '../helpers/room-helpers';
+
+// ─── RPC Setup Helpers ────────────────────────────────────────────────────────
+
+async function createRoomWithLinkedGoalAndTask(
+	page: Parameters<typeof waitForWebSocketConnected>[0]
+): Promise<{ roomId: string; taskId: string; goalId: string }> {
+	await waitForWebSocketConnected(page);
+
+	return page.evaluate(async () => {
+		const hub = window.__messageHub || window.appState?.messageHub;
+		if (!hub?.request) throw new Error('MessageHub not available');
+
+		// Create room
+		const roomRes = await hub.request('room.create', {
+			name: 'E2E Goal Indicator Test Room',
+		});
+		const roomId = (roomRes as { room: { id: string } }).room.id;
+
+		// Create task
+		const taskRes = await hub.request('task.create', {
+			roomId,
+			title: 'Goal-Linked Task',
+			description: 'A task linked to a mission',
+		});
+		const taskId = (taskRes as { task: { id: string } }).task.id;
+
+		// Create goal (mission)
+		const goalRes = await hub.request('goal.create', {
+			roomId,
+			title: 'My Test Mission',
+			description: 'A mission for testing goal indicators',
+			priority: 'normal',
+		});
+		const goalId = (goalRes as { goal: { id: string } }).goal.id;
+
+		// Link task to goal
+		await hub.request('goal.linkTask', { roomId, goalId, taskId });
+
+		return { roomId, taskId, goalId };
+	});
+}
+
+async function createRoomWithUnlinkedTask(
+	page: Parameters<typeof waitForWebSocketConnected>[0]
+): Promise<{ roomId: string; taskId: string }> {
+	await waitForWebSocketConnected(page);
+
+	return page.evaluate(async () => {
+		const hub = window.__messageHub || window.appState?.messageHub;
+		if (!hub?.request) throw new Error('MessageHub not available');
+
+		const roomRes = await hub.request('room.create', {
+			name: 'E2E Goal Indicator Unlinked Test Room',
+		});
+		const roomId = (roomRes as { room: { id: string } }).room.id;
+
+		const taskRes = await hub.request('task.create', {
+			roomId,
+			title: 'Unlinked Task',
+			description: 'A task with no mission',
+		});
+		const taskId = (taskRes as { task: { id: string } }).task.id;
+
+		return { roomId, taskId };
+	});
+}
+
+// ─── Tests ────────────────────────────────────────────────────────────────────
+
+test.describe('Task Goal Indicator — Task List', () => {
+	let roomId = '';
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/');
+		await page
+			.getByRole('button', { name: 'New Session', exact: true })
+			.waitFor({ timeout: 10000 });
+		roomId = '';
+	});
+
+	test.afterEach(async ({ page }) => {
+		await deleteRoom(page, roomId);
+	});
+
+	test('shows goal badge on task linked to a mission', async ({ page }) => {
+		const result = await createRoomWithLinkedGoalAndTask(page);
+		roomId = result.roomId;
+
+		await page.goto(`/room/${roomId}`);
+		await waitForWebSocketConnected(page);
+
+		// Should see the goal badge with mission title
+		const badge = page.locator(`[data-testid="task-goal-badge-${result.taskId}"]`);
+		await expect(badge).toBeVisible({ timeout: 10000 });
+		await expect(badge).toContainText('My Test Mission');
+	});
+
+	test('does NOT show goal badge on task with no mission', async ({ page }) => {
+		const result = await createRoomWithUnlinkedTask(page);
+		roomId = result.roomId;
+
+		await page.goto(`/room/${roomId}`);
+		await waitForWebSocketConnected(page);
+
+		// Ensure the task is visible first
+		await expect(page.locator('h4:has-text("Unlinked Task")')).toBeVisible({ timeout: 10000 });
+
+		// No goal badge should be present for this task
+		const badge = page.locator(`[data-testid="task-goal-badge-${result.taskId}"]`);
+		await expect(badge).not.toBeVisible();
+	});
+
+	test('clicking goal badge switches to Missions tab', async ({ page }) => {
+		const result = await createRoomWithLinkedGoalAndTask(page);
+		roomId = result.roomId;
+
+		await page.goto(`/room/${roomId}`);
+		await waitForWebSocketConnected(page);
+
+		// Wait for and click the goal badge
+		const badge = page.locator(`[data-testid="task-goal-badge-${result.taskId}"]`);
+		await expect(badge).toBeVisible({ timeout: 10000 });
+		await badge.click();
+
+		// Should now be on the Missions tab
+		await expect(page.locator('h2:has-text("Missions")')).toBeVisible({ timeout: 5000 });
+	});
+});
+
+test.describe('Task Goal Indicator — TaskView', () => {
+	let roomId = '';
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/');
+		await page
+			.getByRole('button', { name: 'New Session', exact: true })
+			.waitFor({ timeout: 10000 });
+		roomId = '';
+	});
+
+	test.afterEach(async ({ page }) => {
+		await deleteRoom(page, roomId);
+	});
+
+	test('shows goal badge in TaskView header for linked task', async ({ page }) => {
+		const result = await createRoomWithLinkedGoalAndTask(page);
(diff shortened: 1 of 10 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly implements the feature of showing associated goals on the room task list and task view. The implementation is well-structured and follows existing codebase patterns.

### What was reviewed

1. **`currentRoomTabSignal`** in `signals.ts` (line 51) — A new signal for cross-component tab navigation. Simple, follows existing signal patterns in the file.

2. **`goalByTaskId` computed** in `room-store.ts` (lines 180-188) — Reverse lookup from task ID to RoomGoal. Clean implementation using a Map built from iterating goals and their `linkedTaskIds`. One minor note: if a task appears in multiple goals' `linkedTaskIds`, the last goal wins silently. This is fine given the data model, but worth knowing.

3. **Goal badge in `RoomTasks.tsx`** (lines 659-679) — Emerald badge with lightning bolt SVG icon, truncated title (max 120px), tooltip, and `stopPropagation()` to prevent the parent task click handler from firing. The `goalByTaskId` Map is also rebuilt locally inside the component (lines 116-123) rather than using the computed from `roomStore`. This is slightly redundant but acceptable since the component receives `goals` as a prop — it keeps the component self-contained and testable without depending on the singleton store.

4. **Goal badge in `TaskView.tsx`** (lines 962-982) — Uses `roomStore.goalByTaskId` computed directly. Clicking navigates to room and sets the tab signal. The navigation flow is correct: `currentRoomTabSignal.value = 'goals'` then `navigateToRoom(roomId)` — the URL change clears `taskViewId`, then the `useEffect` in `Room.tsx` picks up the pending tab.

5. **Tab navigation watcher in `Room.tsx`** (lines 46-55) — Watches `currentRoomTabSignal`, validates against known tabs, sets `activeTab`, then clears the signal. The guard `!taskViewId` prevents the tab switch while still in task view. Signal is cleared after applying.

6. **`RoomDashboard.tsx`** (lines 209-211) — Passes `onGoalClick` that sets `currentRoomTabSignal.value = 'goals'`. The `goalId` parameter from the callback is intentionally ignored since the current UX just navigates to the Missions tab without scrolling to a specific goal.

7. **Unit tests** — 8 tests for RoomTasks goal badge (visible/hidden, click propagation, tooltip, cross-tab rendering) and 4 tests for TaskView goal badge. Tests verify both presence and absence cases. Tests use `data-testid` attributes for reliable selection.

8. **E2E tests** — Properly use RPC only for infrastructure setup (room/task/goal creation and linking) and teardown, with all assertions through visible UI. Follows the E2E test rules in CLAUDE.md.

### No issues found

The implementation is correct, complete, and well-tested. No bugs, security concerns, or missing edge cases identified. The code fits naturally with existing patterns in the codebase.

lsm (commented):

The implementation correctly delivers the core feature — goal badges appear on task list items and in TaskView headers, with working tab navigation. The overall approach is sound and the unit tests provide reasonable coverage. However, there are two meaningful correctness issues and one structural concern worth fixing before merge.

---

### P1 — Missing `roomId` in `useEffect` dependency (potential stale signal cross-room)

**File:** `packages/web/src/islands/Room.tsx`, lines 47–55

```ts
useEffect(() => {
    if (pendingTab && !taskViewId) {
        ...
        currentRoomTabSignal.value = null;
    }
}, [pendingTab, taskViewId]);  // ← roomId missing
```

`roomId` is not in the dependency array. If the user navigates between rooms while a signal is pending, the effect will fire in the new room context. Although the signal is typically cleared immediately, there is a window where a signal set for Room A can switch tabs in Room B. The signal should be cleared in the `roomId` effect's cleanup as well, to guarantee no cross-room tab contamination.

---

### P1 — Race condition: signal set before navigation completes in `TaskView`

**File:** `packages/web/src/components/room/TaskView.tsx`, lines 965–968

```ts
onClick={() => {
    currentRoomTabSignal.value = 'goals';
    navigateToRoom(roomId);
}}
```

The global signal is mutated synchronously, then navigation happens. Because `currentRoomTabSignal` is read directly in the `Room.tsx` component body (line 46, not inside a `useEffect`), and the consuming `useEffect` has `taskViewId` in its deps, the correctness depends on Preact's internal scheduling ensuring `taskViewId` is falsy before the effect re-runs. This works in practice today but is fragile: if Preact ever batches differently, or if navigation is async, the signal can be consumed while `taskViewId` is still set (`!taskViewId` is false), causing it to be cleared without switching tabs. A safer design would be to set the signal after navigation completes, or use a different coordination mechanism.

---

### P2 — Duplicate reverse-lookup computation

**File:** `packages/web/src/components/room/RoomTasks.tsx`, lines 115–123

```ts
// Build reverse lookup: taskId → RoomGoal
const goalByTaskId = new Map<string, RoomGoal>();
if (goals) { ... }
```

`RoomStore` already has a reactive `goalByTaskId` computed signal (added in this same PR, `room-store.ts` line 180). `RoomTasks` then builds an identical Map eagerly in the component body on every render. This bypasses the reactive computed, forces the allocation every render regardless of whether goals changed, and is dead weight since `TaskView` already uses `roomStore.goalByTaskId` directly. Consider either having `RoomTasks` accept a pre-built `Map` (which `RoomDashboard` can derive from `roomStore.goalByTaskId.value`), or having `RoomDashboard` pass `roomStore.goalByTaskId.value` instead of the `goals` array.

---

### P2 — `onGoalClick` `goalId` parameter is ignored at the call site

**File:** `packages/web/src/components/room/RoomDashboard.tsx`, lines 209–211

```ts
onGoalClick={() => {
    currentRoomTabSignal.value = 'goals';
}}
```

The prop type is `(goalId: string) => void` and `TaskItem` correctly passes `goal.id`. But the consumer ignores the `goalId`. This means clicking the goal badge navigates to the Missions tab but does not scroll to or highlight the specific goal. This is a degraded UX from what the API implies. Either the GoalsEditor should support scrolling/highlighting a specific goal when the tab loads, or the prop type should be simplified to `() => void` to honestly represent what happens.

---

### P3 — E2E setup goes beyond exempted infrastructure RPC

**File:** `packages/e2e/tests/features/task-goal-indicator.e2e.ts`, lines 27–64

CLAUDE.md explicitly exempts `room.create`/`room.delete` as accepted infrastructure RPC in beforeEach/afterEach. The setup helpers also call `task.create`, `goal.create`, and `goal.linkTask` via `hub.request()`. These are not explicitly exempted. However, they are correctly limited to the `beforeEach`/setup function (not in test actions or assertions), the rationale is strong (no UI way to pre-seed test data for this scenario), and the pattern is reasonable. This is a minor policy stretch worth noting, but not blocking.

---

### P3 — Goal badge title content is the same as tooltip

**File:** `packages/web/src/components/room/RoomTasks.tsx`, line 684

The badge already displays `goal.title` as visible text content and also has `title="Mission: {goal.title}"` as a tooltip. The tooltip adds the "Mission:" prefix but otherwise duplicates visible text, which adds no meaningful additional information. Either remove the tooltip or make it more descriptive (e.g., include the goal description or status).

---

### Summary

The feature works correctly for the happy path and is well-tested at the unit level. The two P1 issues involve the global signal's lifecycle — specifically that it can leak across room navigations (missing dep) and relies on fragile scheduling assumptions between `TaskView` and `Room.tsx`. These should be addressed. The P2 duplicate map computation is a maintainability concern. None of these are data-loss or security issues.

lsm (commented):

### Re-review: All previous P1/P2 issues addressed

The four issues from the previous review have been resolved:

1. **`roomId` in useEffect deps (P1)** -- Fixed. `roomId` is in the dependency array at `Room.tsx:45`, and the cleanup function clears `currentRoomTabSignal` to prevent cross-room contamination.

2. **Signal-then-navigate ordering (P1)** -- Fixed. In `TaskView.tsx:967-968`, `navigateToRoom(roomId)` is called first (which synchronously clears `currentRoomTaskIdSignal`), then `currentRoomTabSignal.value = 'goals'` is set. The `Room.tsx` useEffect at line 49-57 correctly gates on `!taskViewId` before consuming the signal.

3. **Duplicate Map computation (P2)** -- Fixed. `RoomTasks` now accepts `goalByTaskId?: Map<string, RoomGoal>` directly from `roomStore.goalByTaskId.value` (passed in `RoomDashboard.tsx:205`), eliminating the duplicate reverse-lookup.

4. **Unused `goalId` parameter on `onGoalClick` (P2)** -- Fixed. Simplified to `onGoalClick?: () => void`.

### Current assessment

The implementation is clean and well-tested:

- **`room-store.ts`**: `goalByTaskId` computed signal correctly builds a `taskId -> RoomGoal` reverse map, reusing the existing `goals` signal. It sits alongside the existing `tasksByGoalId` computed, which is the natural companion.

- **`RoomTasks.tsx`**: The `goalByTaskId` map is threaded through `TaskList -> TaskGroup -> TaskItem` with the final lookup `goalByTaskId?.get(task.id)` at the `TaskItem` level. The badge uses `e.stopPropagation()` to avoid triggering `onTaskClick`. Correct.

- **`TaskView.tsx`**: Reads `roomStore.goalByTaskId.value.get(taskId)` at render time (line 604). This is reactive because `goalByTaskId` is a computed signal, and the `.value` access subscribes the component. The click handler order (navigate then signal) is correct.

- **`Room.tsx`**: The `pendingTab` is read outside useEffect (line 48) which creates a Preact signal subscription at the component level. When it changes, re-render triggers, useEffect sees the new `pendingTab` dep, validates it against the allowlist, sets the tab, and clears the signal. Clean.

- **Tests**: Unit tests cover all key scenarios (badge visibility, click propagation, tooltip, per-tab rendering, no-goal case). E2E test covers the full flow including RPC setup/teardown.

**Recommendation: APPROVE** -- All previous feedback has been addressed. No new issues found.

lsm (commented):

**Recommendation: APPROVE** — all four previously raised issues have been correctly addressed. A few minor observations remain but none block merging.

---

### How the Previous Issues Were Resolved

**P1: Missing `roomId` in useEffect deps + cross-room contamination** — Fixed correctly. `packages/web/src/islands/Room.tsx` lines 36-45 now list `[roomId]` as the dependency, and the cleanup function clears `currentRoomTabSignal.value = null` when leaving the room. The second `useEffect` that watches `pendingTab` also correctly includes `[pendingTab, taskViewId, roomId]`.

**P1: Signal-then-navigate ordering in TaskView** — Fixed correctly. `packages/web/src/components/room/TaskView.tsx` lines 965-968 now call `navigateToRoom(roomId)` first, then set `currentRoomTabSignal.value = 'goals'`. This ensures the URL change (and the resulting unmount of TaskView) happens before the signal is consumed by the re-mounted Room.

**P2: Duplicate reverse-lookup Map** — Fixed correctly. `RoomTasksProps` now accepts `goalByTaskId?: Map<string, RoomGoal>` directly (line 45 in `RoomTasks.tsx`), and `RoomDashboard.tsx` line 205 passes `roomStore.goalByTaskId.value` (the pre-built computed signal). No Map is rebuilt inside the component.

**P2: Unused `goalId` parameter on `onGoalClick`** — Fixed correctly. The type is now `onGoalClick?: () => void` throughout the component tree, and all call sites pass a zero-argument lambda.

---

**P3 — RoomDashboard.test.tsx line 34 mock shape is a plain signal, not a computed**

The mock uses `signal(new Map())` instead of a `computed`, which means the test bypasses the actual `goalByTaskId` derivation logic. This is acceptable for unit testing the dashboard's rendering behavior in isolation, but it means the computed's own logic (iterating `goal.linkedTaskIds`) is not covered by any unit test in the dashboard suite. The dedicated `RoomTasks.test.tsx` tests cover the badge rendering directly, so there is no gap in functional coverage — this is a minor test-quality note only.

**P3 — `RoomDashboard.test.tsx` carries `// @ts-nocheck` at line 1**

The file suppresses all TypeScript checks. Given that the dashboard now reads `goalByTaskId` from the store, a future type mismatch between the mock and the real computed type would be invisible. Not introduced by this PR, but worth noting in context.

**P3 — E2E `afterEach` calls `deleteRoom(page, roomId)` when `roomId === ''`**

In the `beforeEach` blocks (lines 102 and 156), `roomId` is reset to the empty string. If a test fails before the `createRoom…` RPC call completes and assigns `roomId`, `deleteRoom(page, '')` is called with an empty string in teardown. Depending on how `deleteRoom` is implemented, this may silently no-op or log a spurious error. It is not a data-loss risk but could produce noisy teardown output. Consider guarding: `if (roomId) await deleteRoom(page, roomId)`.

**P3 — `not.toBeVisible()` vs `not.toBeInTheDOM()` in negative E2E assertions**

Lines 133-134 and 194-195 assert `await expect(badge).not.toBeVisible()`. If the element is absent from the DOM entirely (which is the expected case — the badge is conditionally rendered), Playwright's `not.toBeVisible()` will pass because a missing element is not visible. This works today, but `not.toBeInTheDOM()` / `expect(badge).not.toBeAttached()` would be a more precise assertion that the element was never rendered rather than merely hidden. Minor style issue.

---

### Summary

All four P0/P1/P2 issues from the previous review are resolved. The navigate-then-signal ordering in TaskView, the roomId dep-array fix with signal cleanup, the elimination of redundant Map construction, and the corrected `onGoalClick` signature are all implemented correctly. The new unit tests (RoomTasks, TaskView, RoomDashboard) and the new E2E suite provide solid coverage. The remaining observations are cosmetic P3 nits.

PR #630, opened 2026-03-22
feat: add session title generation job handler

lsm:
Create job handler for session.title_generation queue that delegates to
SessionLifecycle.generateTitleAndRenameBranch(). Errors propagate so
the job queue can retry. Includes unit tests for success, error
propagation, and missing-session cases.

### DIFF
--- a/packages/daemon/src/lib/job-handlers/session-title.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/session-title.handler.ts
@@ -12,10 +12,14 @@ export async function handleSessionTitleGeneration(
 	job: Job,
 	sessionLifecycle: SessionLifecycle
 ): Promise<{ generated: true }> {
-	const { sessionId, userMessageText } = job.payload as {
-		sessionId: string;
-		userMessageText: string;
-	};
+	const { sessionId, userMessageText } = job.payload;
+
+	if (!sessionId || typeof sessionId !== 'string') {
+		throw new Error('Job payload missing required field: sessionId');
+	}
+	if (typeof userMessageText !== 'string') {
+		throw new Error('Job payload missing required field: userMessageText');
+	}
 
 	await sessionLifecycle.generateTitleAndRenameBranch(sessionId, userMessageText);
 
--- a/packages/daemon/src/lib/job-handlers/session-title.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/session-title.handler.ts
@@ -17,7 +17,7 @@ export async function handleSessionTitleGeneration(
 	if (!sessionId || typeof sessionId !== 'string') {
 		throw new Error('Job payload missing required field: sessionId');
 	}
-	if (typeof userMessageText !== 'string') {
+	if (!userMessageText || typeof userMessageText !== 'string') {
 		throw new Error('Job payload missing required field: userMessageText');
 	}
 
--- a/packages/daemon/src/lib/job-handlers/session-title.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/session-title.handler.ts
@@ -1,0 +1,23 @@
+import type { Job } from '../../storage/repositories/job-queue-repository';
+import type { SessionLifecycle } from '../session/session-lifecycle';
+
+/**
+ * Job handler for session.title_generation queue.
+ *
+ * Calls SessionLifecycle.generateTitleAndRenameBranch() with the sessionId and
+ * userMessageText from the job payload. Errors propagate so the job queue can
+ * retry the job automatically.
+ */
+export async function handleSessionTitleGeneration(
+	job: Job,
+	sessionLifecycle: SessionLifecycle
+): Promise<{ generated: true }> {
+	const { sessionId, userMessageText } = job.payload as {
+		sessionId: string;
+		userMessageText: string;
+	};
+
+	await sessionLifecycle.generateTitleAndRenameBranch(sessionId, userMessageText);
+
+	return { generated: true };
+}
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -62,6 +62,8 @@ import { SpaceAgentRepository } from '../../storage/repositories/space-agent-rep
 import type { JobQueueRepository } from '../../storage/repositories/job-queue-repository';
 import type { JobQueueProcessor } from '../../storage/job-queue-processor';
 import { SpaceSessionGroupRepository } from '../../storage/repositories/space-session-group-repository';
+import { SESSION_TITLE_GENERATION } from '../job-queue-constants';
+import { handleSessionTitleGeneration } from '../job-handlers/session-title.handler';
 import { SpaceRuntimeService } from '../space/runtime/space-runtime-service';
 import { setupSpaceWorkflowRunHandlers } from './space-workflow-run-handlers';
 import type { SpaceWorkflowRunTaskManagerFactory } from './space-workflow-run-handlers';
@@ -147,6 +149,11 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	setupConfigHandlers(deps.messageHub, deps.sessionManager, deps.daemonHub);
 	setupTestHandlers(deps.messageHub, deps.db);
 	setupRewindHandlers(deps.messageHub, deps.sessionManager, deps.daemonHub);
+
+	// Job queue handler registrations
+	deps.jobProcessor.register(SESSION_TITLE_GENERATION, (job) =>
+		handleSessionTitleGeneration(job, deps.sessionManager.getSessionLifecycle())
+	);
 
 	// Room handlers
 	setupRoomHandlers(
--- a/packages/daemon/src/lib/session/session-manager.ts
+++ b/packages/daemon/src/lib/session/session-manager.ts
@@ -219,8 +219,7 @@ export class SessionManager {
 	}
 
 	/**
-	 * Get the session lifecycle manager (exposed for testing)
-	 * @internal
+	 * Get the session lifecycle manager.
 	 */
 	getSessionLifecycle(): SessionLifecycle {
 		return this.sessionLifecycle;
--- a/packages/daemon/tests/unit/job-handlers/session-title-handler.test.ts
+++ b/packages/daemon/tests/unit/job-handlers/session-title-handler.test.ts
@@ -93,4 +93,13 @@ describe('handleSessionTitleGeneration', () => {
 			'Job payload missing required field: userMessageText'
 		);
 	});
+
+	it('throws when userMessageText is an empty string', async () => {
+		const lifecycle = makeSessionLifecycle();
+		const job = makeJob({ sessionId: 'session-123', userMessageText: '' });
+
+		await expect(handleSessionTitleGeneration(job, lifecycle)).rejects.toThrow(
+			'Job payload missing required field: userMessageText'
+		);
+	});
 });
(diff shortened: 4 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, minimal job handler that correctly delegates to `SessionLifecycle.generateTitleAndRenameBranch()`. The implementation matches the task requirements. Tests pass and cover the happy path, error propagation, and session-not-found scenarios.

### Observations

**P2 — No payload validation before destructuring (line 15, `session-title.handler.ts`)**

The handler blindly casts `job.payload` and destructures `sessionId` and `userMessageText` without any validation. If a malformed job is enqueued (e.g., missing `sessionId` or `userMessageText`), the handler will pass `undefined` to `generateTitleAndRenameBranch()` rather than failing fast with a clear error. This could cause confusing downstream errors and waste retries.

Recommend adding a guard:
```ts
if (!sessionId || typeof sessionId !== 'string') {
  throw new Error('Job payload missing required field: sessionId');
}
```

This is especially relevant since `job.payload` is typed as `Record<string, unknown>` — the type system provides no guarantee about the shape.

**P3 — Test file does not cover the missing/malformed payload edge case**

The third test ("session not found") tests that `generateTitleAndRenameBranch` throws, but no test verifies what happens when the payload itself is malformed (e.g., missing `sessionId`). This is directly related to the P2 above.

**P3 — Handler does not use the `SESSION_TITLE_GENERATION` constant from `job-queue-constants.ts`**

The handler file itself doesn't reference the queue name, so this is purely informational — the test hardcodes `'session.title_generation'` in `makeJob`. Using the constant would be more consistent with the existing pattern.

### Summary

The core implementation is correct and well-tested for its scope. The only meaningful concern is the lack of payload validation (P2), which could cause confusing failures if a malformed job reaches this handler. Everything else is solid.

lsm (commented):

---

### Summary

The handler function is correctly implemented in isolation, but it is **never registered with the job processor**. The task is to create a handler that processes `session.title_generation` jobs — but without a `jobProcessor.register(...)` call, the queue will silently collect jobs that are never executed. The implementation is incomplete.

---

### Issues

#### P0 — Handler is never registered with the job processor

**File:** `packages/daemon/src/lib/rpc-handlers/index.ts`

`JobQueueProcessor.register(queue, handler)` is the mechanism that connects a queue name to its handler function. Searching the entire `packages/daemon/src` tree for `jobProcessor.register` returns zero results. The handler in `session-title.handler.ts` is exported but never imported and never registered. Jobs enqueued to `session.title_generation` will be dequeued by the processor's poll loop, find no handler, and be immediately marked `failed` with the error `"No handler registered for queue: session.title_generation"` (see `job-queue-processor.ts` line 95).

The correct fix is to add, inside `setupRPCHandlers` in `/packages/daemon/src/lib/rpc-handlers/index.ts`, a registration call such as:

```ts
deps.jobProcessor.register(
  SESSION_TITLE_GENERATION,
  (job) => handleSessionTitleGeneration(job, deps.sessionManager.sessionLifecycle)
);
```

This wiring is the central deliverable of the task and is missing.

---

#### P1 — Queue name is a string literal in the test, not imported from the constant

**File:** `packages/daemon/tests/unit/job-handlers/session-title-handler.test.ts`, line 9

```ts
queue: 'session.title_generation',
```

The constant `SESSION_TITLE_GENERATION` exists in `packages/daemon/src/lib/job-queue-constants.ts` for exactly this reason — to avoid typos and keep queue names in sync. The test uses an inline string literal. If the constant value changes, the test will silently test a stale queue name. Import and use `SESSION_TITLE_GENERATION` from the constants file.

---

#### P1 — No validation for missing or non-string payload fields

**File:** `packages/daemon/src/lib/job-handlers/session-title.handler.ts`, lines 15–18

```ts
const { sessionId, userMessageText } = job.payload as {
    sessionId: string;
    userMessageText: string;
};
```

This is an unchecked cast. If a job is enqueued with a missing `sessionId` (e.g., `undefined`) or `userMessageText` (e.g., `null`), the handler will pass those through to `generateTitleAndRenameBranch` with no diagnostic. The downstream method may throw an unhelpful error or silently no-op. A minimal guard — checking that both fields are non-empty strings before calling the lifecycle method — would produce an immediately actionable error message and prevent the job from retrying indefinitely on a permanently bad payload. This is a system boundary where untrusted data enters the handler.

---

#### P2 — No test for malformed payload (missing `sessionId` or `userMessageText`)

**File:** `packages/daemon/tests/unit/job-handlers/session-title-handler.test.ts`

The three existing tests cover: success, downstream error propagation, and a "session not found" error. There is no test for jobs with a structurally invalid payload (e.g., `sessionId` absent or not a string). Once payload validation is added (P1 above), a test verifying the validation error message would complete the coverage.

---

#### P3 — Handler signature does not match `JobHandler` type

**File:** `packages/daemon/src/lib/job-handlers/session-title.handler.ts`

`JobQueueProcessor`'s `JobHandler` type is `(job: Job) => Promise<Record<string, unknown> | void>`. The exported function takes two arguments `(job, sessionLifecycle)`, so callers must wrap it in a closure to bind `sessionLifecycle`. That closure is also the missing registration call (P0 above). There is no bug here in isolation, but the mismatch between the exported function's shape and `JobHandler` is worth noting as it means the function cannot be directly passed to `register()` without a wrapper.

lsm (commented):

**Recommendation: APPROVE**

All three issues from the previous review round have been properly addressed:

1. **Handler registration** -- Now registered in `setupRPCHandlers` at `packages/daemon/src/lib/rpc-handlers/index.ts:150-152` via `deps.jobProcessor.register()`. The handler correctly resolves `SessionLifecycle` lazily through `deps.sessionManager.getSessionLifecycle()`.

2. **Payload validation** -- The handler validates `sessionId` (must be a non-empty string) and `userMessageText` (must be a string) before calling `generateTitleAndRenameBranch`. The asymmetry is appropriate: empty `sessionId` is never valid, while empty `userMessageText` could be legitimate.

3. **Test coverage** -- Tests now use the `SESSION_TITLE_GENERATION` constant and include 3 malformed payload cases (missing sessionId, non-string sessionId, missing userMessageText).

**What I verified:**
- Handler signature matches `generateTitleAndRenameBranch(sessionId: string, userMessageText: string)` in `session-lifecycle.ts:604-607`
- `Job.payload` is typed as `Record<string, unknown>` in `job-queue-repository.ts:10`, so runtime validation is necessary and correctly applied
- `JobQueueProcessor.register()` accepts `(queue: string, handler: JobHandler)` where `JobHandler = (job: Job) => Promise<Record<string, unknown> | void>`, and the handler's return type `{ generated: true }` satisfies this
- `getSessionLifecycle()` is a public method on `SessionManager` (line 225) that returns the internal `SessionLifecycle` instance
- Error propagation is intentional -- the `JobQueueProcessor` handles retry logic when handlers throw

No issues found. Clean, focused implementation.

lsm (commented):

**Recommendation: APPROVE**

The three P0/P1/P2 issues flagged in the previous round have all been correctly addressed. The handler is registered, payload is validated, and tests use the constant. Below are the remaining observations from this pass.

---

### P2: `userMessageText` validation is asymmetric with `sessionId`

`packages/daemon/src/lib/job-handlers/session-title.handler.ts`, lines 17–22

`sessionId` is validated with `!sessionId || typeof sessionId !== 'string'`, which rejects an empty string `""`. `userMessageText` is validated only with `typeof userMessageText !== 'string'`, so an empty string `""` passes through and is forwarded to `generateTitleAndRenameBranch`. If the call-site that enqueues the job could ever produce `userMessageText: ""`, the lifecycle method will be invoked with an empty text and likely generate a fallback title silently. A matching check (`!userMessageText || typeof userMessageText !== 'string'`) or at minimum a test case for `userMessageText: ''` would close this gap.

---

### P2: `getSessionLifecycle()` is marked `@internal` but is now called from production wiring code

`packages/daemon/src/lib/rpc-handlers/index.ts`, line 151; `packages/daemon/src/lib/session/session-manager.ts`, line 222

The JSDoc says "exposed for testing". The registration call in `setupRPCHandlers` is production code, so the comment is now misleading. Either remove `@internal` or rename the method to reflect its broadened use (e.g., `getLifecycle()`). No behaviour change required, but leaving the comment as-is will confuse future readers.

---

### P3: `@internal` comment creates friction without enforcement

TypeScript has no runtime enforcement of `@internal`. Given that `getSessionLifecycle()` is now called from two production paths (session handlers and job-handler wiring), it is worth deciding at the team level whether the method needs any access protection or if the comment should simply be removed.

---

### What was fixed correctly

- Handler is registered with `deps.jobProcessor.register(SESSION_TITLE_GENERATION, ...)` in `setupRPCHandlers` — the P0 from the prior round is resolved.
- `sessionId` validation now correctly catches missing fields, non-string types, and falsy values.
- Tests import `SESSION_TITLE_GENERATION` from the constants module — no more string literals.
- Three new test cases cover missing `sessionId`, non-string `sessionId`, and missing `userMessageText`.
- The `makeJob`/`makeSessionLifecycle` helper pattern is clean and reusable.
- Error propagation (for job-queue retry) is tested explicitly.
- The registration closure correctly captures `deps.sessionManager.getSessionLifecycle()` at call time, which is fine since `SessionManager` is a singleton in the daemon lifecycle.

lsm (commented):

### Summary

This PR adds a `session.title_generation` job handler that delegates to `SessionLifecycle.generateTitleAndRenameBranch()`. The implementation is clean, correct, and well-tested.

### What was reviewed

1. **Handler** (`packages/daemon/src/lib/job-handlers/session-title.handler.ts`) -- Extracts `sessionId` and `userMessageText` from job payload, validates both fields (rejects missing, non-string, and empty string values), calls the lifecycle method, and returns `{ generated: true }`. Errors propagate for job queue retry.

2. **Registration** (`packages/daemon/src/lib/rpc-handlers/index.ts:150-152`) -- Handler is registered with `jobProcessor.register(SESSION_TITLE_GENERATION, ...)` using the constant from `job-queue-constants.ts`. The `getSessionLifecycle()` call is evaluated per-job (not captured once at registration time), which is correct.

3. **JSDoc fix** (`packages/daemon/src/lib/session/session-manager.ts:221-223`) -- `@internal` tag removed from `getSessionLifecycle()` since it is now called from production wiring code. Appropriate change.

4. **Tests** (`packages/daemon/tests/unit/job-handlers/session-title-handler.test.ts`) -- 7 tests covering: happy path with correct params and return value, error propagation (generic + session-not-found), missing `sessionId`, non-string `sessionId`, missing `userMessageText`, and empty string `userMessageText`. All pass.

### Findings

No blocking issues found. The previous review feedback (handler registration, symmetric validation, JSDoc cleanup) has been properly addressed.

**Recommendation: APPROVE**

The handler is minimal, correctly delegates to the existing lifecycle method, validates inputs symmetrically, propagates errors for retry, and has thorough test coverage.

lsm (commented):

**Recommendation: APPROVE**

All issues from the previous round (symmetric `userMessageText` validation, `@internal` JSDoc removal) have been correctly addressed. The implementation is complete and correct.

---

### Verification

**Symmetric validation** — `packages/daemon/src/lib/job-handlers/session-title.handler.ts` line 20 now uses `!userMessageText || typeof userMessageText !== 'string'`, which rejects empty strings just as `sessionId` validation does. This closes the asymmetry identified in the last round.

**JSDoc cleaned up** — `packages/daemon/src/lib/session/session-manager.ts` lines 221–225: the `@internal` tag and the misleading "exposed for testing" comment are gone. The method is now documented plainly as `Get the session lifecycle manager.`

**Empty string test** — Test at line 97 of `packages/daemon/tests/unit/job-handlers/session-title-handler.test.ts` correctly verifies that `userMessageText: ''` throws `'Job payload missing required field: userMessageText'`. All 7 tests pass.

**Handler registration** — `packages/daemon/src/lib/rpc-handlers/index.ts` lines 150–152 register the handler via `deps.jobProcessor.register(SESSION_TITLE_GENERATION, ...)`. Confirmed present from prior rounds.

**No new issues found.** The implementation is minimal, well-tested, and correctly integrated.

PR #631, opened 2026-03-22
feat: expose triggerPoll and create github.poll job handler

lsm:
- Add public `triggerPoll()` to GitHubPollingService (no-op if isPolling)
- Create github-poll.handler.ts with self-scheduling and dedup logic
- Enqueue next poll only when no pending github.poll job exists
- Finally block ensures scheduling chain survives triggerPoll errors
- Unit tests cover: triggerPoll call, self-scheduling, dedup, error recovery

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -23,7 +23,7 @@ import type { TaskAgentManager } from './lib/space/runtime/task-agent-manager';
 import { JobQueueRepository } from './storage/repositories/job-queue-repository';
 import { JobQueueProcessor } from './storage/job-queue-processor';
 import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
-import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
+import { GITHUB_POLL, JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -363,6 +363,20 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	if (gitHubService) {
 		gitHubService.start();
 		logInfo('[Daemon] GitHub service started');
+
+		// Seed the initial github.poll job if polling is configured and no job already exists.
+		// The self-scheduling chain in the handler keeps the chain alive after this seed.
+		if (config.githubPollingInterval && config.githubPollingInterval > 0) {
+			const existing = jobQueue.listJobs({
+				queue: GITHUB_POLL,
+				status: ['pending', 'processing'],
+				limit: 1,
+			});
+			if (existing.length === 0) {
+				jobQueue.enqueue({ queue: GITHUB_POLL, payload: {} });
+				logInfo('[Daemon] Seeded initial github.poll job');
+			}
+		}
 	}
 
 	// Register job handlers BEFORE starting the processor so no pending job
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -359,9 +359,13 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		},
 	});
 
-	// Start GitHub service after server is ready
+	// Start GitHub service after server is ready.
+	// Pass useJobQueueScheduler when polling is configured so the pollingService's
+	// built-in setInterval is skipped — the github.poll job-queue chain owns the
+	// schedule instead, preventing a dual-schedule where both mechanisms fire independently.
 	if (gitHubService) {
-		gitHubService.start();
+		const useJobQueueScheduler = !!config.githubPollingInterval && config.githubPollingInterval > 0;
+		gitHubService.start({ useJobQueueScheduler });
 		logInfo('[Daemon] GitHub service started');
 
 		// Seed the initial github.poll job if polling is configured and no job already exists.
--- a/packages/daemon/src/lib/github/github-service.ts
+++ b/packages/daemon/src/lib/github/github-service.ts
@@ -120,8 +120,13 @@ export class GitHubService {
 	 * Start the GitHub service
 	 * - Starts polling if configured
 	 * - Creates webhook handler if secret is configured
+	 *
+	 * @param options.useJobQueueScheduler - When true, the polling service is created
+	 *   but its built-in setInterval timer is NOT started. The job-queue handler owns
+	 *   the schedule entirely (via the github.poll queue), which avoids a dual-schedule
+	 *   where both the setInterval and the job-queue chain call triggerPoll independently.
 	 */
-	start(): void {
+	start(options?: { useJobQueueScheduler?: boolean }): void {
 		// Initialize webhook handler if secret is configured
 		if (this.config.githubWebhookSecret) {
 			this.webhookHandler = createWebhookHandler(this.config.githubWebhookSecret, async (event) => {
@@ -130,7 +135,7 @@ export class GitHubService {
 			log.info('Webhook handler initialized');
 		}
 
-		// Start polling if interval is configured and token is available
+		// Create polling service if interval is configured and token is available
 		if (
 			this.config.githubPollingInterval &&
 			this.config.githubPollingInterval > 0 &&
@@ -145,10 +150,19 @@ export class GitHubService {
 					await this.processEvent(event);
 				}
 			);
-			this.pollingService.start();
-			log.info('Polling service started', {
-				intervalMs: this.config.githubPollingInterval * 1000,
-			});
+
+			if (options?.useJobQueueScheduler) {
+				// Job queue owns the schedule — skip the built-in setInterval so only
+				// one scheduling mechanism fires per interval.
+				log.info('Polling service created (job-queue-driven, setInterval skipped)', {
+					intervalMs: this.config.githubPollingInterval * 1000,
+				});
+			} else {
+				this.pollingService.start();
+				log.info('Polling service started', {
+					intervalMs: this.config.githubPollingInterval * 1000,
+				});
+			}
 		}
 
 		log.info('GitHub service started');
--- a/packages/daemon/src/lib/github/github-service.ts
+++ b/packages/daemon/src/lib/github/github-service.ts
@@ -593,6 +593,13 @@ export class GitHubService {
 	isPolling(): boolean {
 		return this.pollingService?.isRunning() ?? false;
 	}
+
+	/**
+	 * Return the underlying polling service instance, or undefined if not configured.
+	 */
+	getPollingService(): GitHubPollingService | undefined {
+		return this.pollingService;
+	}
 }
 
 /**
--- a/packages/daemon/src/lib/github/polling-service.ts
+++ b/packages/daemon/src/lib/github/polling-service.ts
@@ -140,6 +140,17 @@ export class GitHubPollingService {
 	}
 
 	/**
+	 * Trigger a poll immediately. No-op if a poll is already in progress.
+	 */
+	async triggerPoll(): Promise<void> {
+		if (this.isPolling) {
+			log.debug('Poll already in progress, skipping triggerPoll');
+			return;
+		}
+		await this.pollAllRepositories();
+	}
+
+	/**
 	 * Poll all configured repositories
 	 */
 	private async pollAllRepositories(): Promise<void> {
--- a/packages/daemon/src/lib/job-handlers/github-poll.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/github-poll.handler.ts
@@ -49,10 +49,11 @@ export async function handleGitHubPoll(deps: GitHubPollHandlerDeps): Promise<Git
 	// Only enqueue the next job if there is no pending or processing job
 	// already in the chain. Checking 'processing' prevents a duplicate chain
 	// from forming under stale-reclaim or slow-poll scenarios.
+	// limit: 1 is sufficient — we only need to know if any job exists.
 	const existingJobs = jobQueue.listJobs({
 		queue: GITHUB_POLL,
 		status: ['pending', 'processing'],
-		limit: 10,
+		limit: 1,
 	});
 
 	if (existingJobs.length === 0) {
--- a/packages/daemon/src/lib/job-handlers/github-poll.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/github-poll.handler.ts
@@ -1,0 +1,46 @@
+/**
+ * Job handler for github.poll queue.
+ *
+ * Triggers a poll of all GitHub repositories and self-schedules the next poll
+ * job, with deduplication to prevent multiple concurrent poll chains.
+ */
+
+import { GITHUB_POLL } from '../job-queue-constants';
+import type { GitHubPollingService } from '../github/polling-service';
+import type { JobQueueRepository } from '../../storage/repositories/job-queue-repository';
+
+export interface GitHubPollHandlerDeps {
+	pollingService: GitHubPollingService;
+	jobQueue: JobQueueRepository;
+	intervalMs: number;
+}
+
+export interface GitHubPollResult {
+	polled: boolean;
+	nextRunAt: number;
+}
+
+export async function handleGitHubPoll(deps: GitHubPollHandlerDeps): Promise<GitHubPollResult> {
+	const { pollingService, jobQueue, intervalMs } = deps;
+	const nextRunAt = Date.now() + intervalMs;
+
+	try {
+		await pollingService.triggerPoll();
+	} finally {
+		const pendingJobs = jobQueue.listJobs({
+			queue: GITHUB_POLL,
+			status: ['pending'],
+			limit: 10,
+		});
+
+		if (pendingJobs.length === 0) {
+			jobQueue.enqueue({
+				queue: GITHUB_POLL,
+				payload: {},
+				runAt: nextRunAt,
+			});
+		}
+	}
+
+	return { polled: true, nextRunAt };
+}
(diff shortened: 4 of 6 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly implements the requested work: exposing `triggerPoll()` on `GitHubPollingService` and creating a `github.poll` job handler with self-scheduling and dedup logic. The code is well-structured, the tests are thorough, and typecheck passes cleanly.

### What looks good

- **`triggerPoll()` design**: The guard against `isPolling` is a sensible optimization, even though `pollAllRepositories()` has the same guard internally. It avoids unnecessary async call overhead.
- **`finally` block for self-scheduling**: Ensures the polling chain survives errors, which is critical for a recurring job pattern.
- **Dedup logic**: Checking for existing pending jobs before enqueueing prevents schedule pile-up.
- **Dependency injection**: Taking deps as a parameter rather than conforming directly to `JobHandler` signature keeps the handler testable. The wiring closure will need to be added when this is registered with the processor (presumably a follow-up task).
- **Tests**: All 6 tests are meaningful and cover the key behaviors -- happy path, dedup, error recovery, and query parameter validation. The error recovery test correctly expects the rejection to propagate while verifying the `finally` block still enqueues.

### Minor observations (P3)

1. **`polled: true` is always returned** (`github-poll.handler.ts:45`): When `triggerPoll()` is a no-op (because `isPolling` is true), the result still says `polled: true`. This is technically inaccurate but harmless since the return value is informational and the processor only uses it for `repo.complete()`.

2. **Non-atomic dedup check**: The `listJobs` + `enqueue` sequence is not atomic, so two concurrent executions could theoretically both see 0 pending jobs and both enqueue. In practice this is not a concern because the `JobQueueProcessor` defaults to `maxConcurrent: 1`, but it's worth noting if concurrency is ever increased.

3. **Redundant `isPolling` check**: `triggerPoll()` checks `this.isPolling` before calling `pollAllRepositories()`, which checks the same flag again. This is intentional defense-in-depth, just noting it.

No blocking issues found.

lsm (commented):

---

### Summary

The implementation correctly exposes `triggerPoll()` on `GitHubPollingService` and the job handler logic is sound, but the handler is **never registered with the job processor** and the **self-scheduling chain can never start** because no initial `github.poll` job is seeded. The feature is effectively dead code in production.

---

### P0 — Handler is never registered (blocking)

**File:** `packages/daemon/src/lib/rpc-handlers/index.ts` / `packages/daemon/src/app.ts`

`handleGitHubPoll` is never imported, and `jobProcessor.register(GITHUB_POLL, ...)` is never called anywhere in the codebase:

```bash
# No matches
grep -r 'jobProcessor.register' packages/daemon/src/
grep -r 'handleGitHubPoll'     packages/daemon/src/
```

The `RPCHandlerDependencies` interface in `index.ts` even has a `TODO` comment acknowledging this gap:

> `TODO: consumed by Milestones 2–5 handlers (session title generation, GitHub polling, room tick, cleanup jobs).`

Without the `register` call the processor will never dequeue a `github.poll` job, so the handler is unreachable at runtime.

---

### P0 — No initial job is seeded (blocking)

The self-scheduling chain in the handler only perpetuates itself — it requires an initial `github.poll` job to be enqueued when the GitHub polling service starts. This seed job is never created anywhere. Even if the handler were registered, the chain could never begin.

The natural place to seed the first job is in `GitHubService.start()` (around line 148 in `github-service.ts`) when `pollingService` is created, or in the `createDaemonApp` bootstrap path after the GitHub service starts. Neither happens.

---

### P0 — `GitHubService.pollingService` is private; the handler has no way to receive it (blocking)

`GitHubService.pollingService` is a `private` field. The handler's `GitHubPollHandlerDeps` requires a `GitHubPollingService` instance directly. There is no public accessor (`getPollingService()`) on `GitHubService`, so there is no path to wire up the dependency even if a caller wanted to.

---

### P1 — Dedup logic misses `processing` status jobs

**File:** `packages/daemon/src/lib/job-handlers/github-poll.handler.ts`, line 30–34

The dedup check only queries for `pending` jobs:

```ts
const pendingJobs = jobQueue.listJobs({
    queue: GITHUB_POLL,
    status: ['pending'],
    limit: 10,
});
```

If a `github.poll` job is currently being processed (status = `processing`) — e.g. due to a stale reclaim or a slow poll cycle — the guard will see zero pending jobs and enqueue a second job. Under stale-reclaim scenarios this could create duplicate chains. The guard should also check `status: ['pending', 'processing']`.

---

### P1 — Error thrown by `triggerPoll` is swallowed after scheduling; callers see it re-thrown

**File:** `packages/daemon/src/lib/job-handlers/github-poll.handler.ts`, lines 27–43

When `triggerPoll()` throws, the `finally` block correctly enqueues the next job. However the exception propagates out of `handleGitHubPoll`, which causes `JobQueueProcessor.processJob` to call `repo.fail()` with exponential back-off retries. Since the next poll job is already enqueued in the `finally` block, you now have **two concurrent scheduling chains** if the retry eventually succeeds. Consider catching the error from `triggerPoll` internally and logging it rather than letting it escape, or returning a result that signals failure without triggering the retry mechanism.

---

### P2 — `nextRunAt` is computed before `triggerPoll` runs

**File:** `packages/daemon/src/lib/job-handlers/github-poll.handler.ts`, line 25

```ts
const nextRunAt = Date.now() + intervalMs;
```

`nextRunAt` is captured before the poll, so for a long-running `pollAllRepositories` call the next job fires sooner than intended. Computing `nextRunAt` after the `await` completes would give a more accurate interval from poll completion.

---

### P2 — No test covers the `pollingService` being `undefined` / `GitHubService` not configured

The unit tests mock `triggerPoll` directly, which is fine. But there is no test (or runtime guard) for the case where `pollingService` is `undefined` (GitHub polling not configured). If the handler were ever registered, a misconfigured daemon would call `undefined.triggerPoll()` and crash. A null-check or a separate `isPollingEnabled` guard in the handler would prevent this.

---

### P3 — Redundant guard in `triggerPoll`

**File:** `packages/daemon/src/lib/github/polling-service.ts`, lines 145–151

`triggerPoll` checks `this.isPolling` and then delegates to `pollAllRepositories`, which also starts with the identical check at line 157. The guard in `triggerPoll` is redundant. It does no harm, but removing it would keep the class DRY.

---

### Test quality note

The six unit tests are well-structured and cover the documented contract. They will all pass today. The problem is that the contract they verify (self-scheduling, dedup) is not yet wired into the running system, so they provide a false sense of completeness.

---

### What is needed before merge

1. Register the handler: `jobProcessor.register(GITHUB_POLL, (job) => handleGitHubPoll({ pollingService, jobQueue, intervalMs }))` in `setupRPCHandlers` or `createDaemonApp` (where `gitHubService` and `pollingService` are available).
2. Expose `pollingService` from `GitHubService` (add a `getPollingService(): GitHubPollingService | undefined` accessor, or restructure so the handler receives the service directly).
3. Seed the first `github.poll` job when polling is enabled (in `GitHubService.start()` or the bootstrap path).
4. Extend the dedup guard to include `'processing'` status.
5. Decide on the error-propagation semantics (catch internally vs. allow retry with a guard against duplicate chains).

lsm (commented):

### Recommendation: APPROVE

The updated PR addresses all previously raised P0/P1/P2 issues. The handler is registered, the seed job is enqueued with dedup, errors are caught, `nextRunAt` is computed post-poll, and the `processing` status is included in dedup checks. Tests are thorough and cover the key behaviors. Code is clean and well-structured.

### Observations

**P2: Dual polling schedules (design concern)**

`GitHubService.start()` (line 148 of `github-service.ts`) calls `this.pollingService.start()`, which sets up a `setInterval` for recurring polls. Simultaneously, the new job handler creates a self-scheduling chain that also calls `triggerPoll()` at `intervalMs` intervals. This means two independent schedules will both attempt to poll -- the `setInterval` inside the polling service and the job queue chain. While `isPolling` prevents truly concurrent polls (one will no-op), the effective poll frequency is roughly doubled since they race independently.

This is likely intentional for a transitional period (the job queue chain will eventually replace the `setInterval`), but if not, one of the two scheduling mechanisms should be removed. If this is intentional, a brief comment in `app.ts` or the handler documenting this would be helpful.

**P3: `limit: 10` in dedup query**

In `github-poll.handler.ts` line 55, the dedup query uses `limit: 10`. Since we only check `existingJobs.length === 0`, `limit: 1` would suffice and be marginally more efficient. The seed job in `app.ts` already uses `limit: 1` for the same pattern.

**P3: `pollingService` type in `makeDeps`**

In the test file line 41, the `pollingService` override type is `ReturnType<typeof mock> | undefined`, but the actual mock object passed is `{ triggerPoll: triggerPollMock }` (not a mock function itself). The `as never` casts make this work, but the type annotation is slightly misleading.

### Summary

The implementation is correct and complete. The self-scheduling pattern with dedup is solid. Tests cover the happy path, dedup (both pending and processing), error recovery, and undefined polling service. The only substantive concern is the dual-polling overlap, which may be intentional.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 issue (dual-schedule polling), one P2 issue (comment vs behavior gap), one P3 nit.

---

### P1 — Dual polling schedule: job-queue chain fires *in addition to* the pre-existing setInterval

`GitHubService.start()` (in `github-service.ts:148`) calls `this.pollingService.start()`, which creates a `setInterval` that already fires `pollAllRepositories()` every `githubPollingInterval` seconds. The new job-queue chain then *also* calls `triggerPoll()` on that same service instance at the same interval.

Result: under normal operation every polling cycle fires twice — once from the job queue handler and once from the setInterval. The `isPolling` mutex in `pollAllRepositories()` prevents concurrent execution within a single cycle, so one of the two callers will get silently dropped per cycle rather than both executing. This means:

1. GitHub API usage doubles compared to what the operator configured.
2. The drop is non-deterministic (race between setInterval tick and job-queue dispatch).
3. The log will show alternating cycles where one scheduler was skipped with no clear indication of why.

The intended design appears to be job-queue-driven scheduling only. The correct fix is to **not call `this.pollingService.start()` inside `GitHubService.start()`** when the job-queue handler is registered, or to have the job-queue handler own the schedule entirely rather than sharing the service with the self-ticking timer.

**File:** `packages/daemon/src/lib/github/github-service.ts:148`

---

### P2 — Handler comment says "pollingService is lazily resolved" but the guard contradicts this

The comment in `rpc-handlers/index.ts:211-214` explains the lazy resolution rationale correctly. However, `GitHubService.start()` is called in `app.ts:362` — which is *before* `setupRPCHandlers` returns in `app.ts` — wait, actually they are in the correct order: `setupRPCHandlers` is called around line 285-310, then `gitHubService.start()` is called at line 362.

This means `pollingService` *is* fully initialized by the time the first job fires. The comment is accurate only for the window between `setupRPCHandlers` returning and `gitHubService.start()` being called, which is a small window but real. The `null` guard in the handler (`if (pollingService)`) correctly handles it. This is fine but the comment could be clearer about the actual timing.

Minor but worth acknowledging — the undefined guard in the handler is correct and necessary, and this design is acceptable.

---

### P3 — `limit: 10` in dedup query is arbitrary and undocumented

In `github-poll.handler.ts:55`, the dedup query uses `limit: 10`. The semantic check is `existingJobs.length === 0`, so any value `>= 1` achieves the same correctness. The value `10` is fine but a comment explaining why (e.g., "any non-zero count is enough; 10 as a small cap") or using `limit: 1` for clarity would make the intent clearer.

---

### What is correct

- Handler registration gated on `githubPollingInterval > 0` — correct.
- Lazy resolve of `pollingService` at call time via `getPollingService()` — correct.
- `finally` / error-catch ensures next-job scheduling always runs — correct.
- Dedup check includes `processing` status — correct.
- Seed job in `app.ts` uses the same `pending | processing` dedup guard — correct.
- 9 unit tests pass and cover the main branches (happy path, dedup pending, dedup processing, triggerPoll error, undefined pollingService).
- Types are clean (`bun run typecheck` passes with no errors).

lsm (commented):

**Recommendation: APPROVE**

This iteration cleanly resolves all previously raised P0 and P1 issues. The dual-schedule concern is properly addressed by the `useJobQueueScheduler` option, the `limit` is tightened to 1, and the overall design is sound.

### What changed and looks correct

1. **`useJobQueueScheduler` option eliminates dual polling** (`github-service.ts:129,154-165`): When `useJobQueueScheduler` is true, `pollingService` is created but `pollingService.start()` is NOT called. This means the built-in `setInterval` never fires, and the job-queue chain is the sole scheduler. This directly addresses the P1 from the previous review.

2. **`app.ts` passes `useJobQueueScheduler: true`** (`app.ts:365`): The flag is correctly computed from `config.githubPollingInterval > 0`, matching the condition under which the job-queue handler is registered in `rpc-handlers/index.ts:216-219`.

3. **`limit: 1` for dedup** (`github-poll.handler.ts:56`): Since we only check `length === 0`, fetching at most 1 row is correct and efficient.

4. **Handler registration, seed job, error handling, dedup** are all properly wired from previous iterations.

### P2 -- `isPolling()` and `isRunning()` return false under job-queue scheduling

**Files:** `github-service.ts:593-609`, `github-handlers.ts:376-377`

When `useJobQueueScheduler` is true, `pollingService.start()` is never called, so `pollingService.isRunning()` returns false (it checks for a non-null `setInterval` handle). This causes:
- `GitHubService.isRunning()` returns false (unless a webhook handler is configured)
- `GitHubService.isPolling()` returns false
- The `github.status` RPC handler reports `pollingEnabled: false` and `pollingInterval: 0`

This is inaccurate when the job queue IS polling. The status endpoint will tell clients that polling is disabled even though it is active. This is a pre-existing design limitation but is now exposed by this PR since the job-queue path is the default when polling is configured.

Not blocking because no runtime behavior depends on these status booleans (they are informational for the UI), but worth tracking for a follow-up.

### P3 -- `stop()` calls `pollingService.stop()` which is a no-op

`GitHubService.stop()` at line 176 calls `this.pollingService.stop()`, which calls `clearInterval(this.pollingInterval)`. Since `pollingInterval` is null (start was never called), `clearInterval(null)` is a no-op. Harmless but slightly misleading -- the job queue chain continues until the processor is stopped separately (which `app.ts` cleanup does correctly at line 438).

### Test quality

All 8 tests are well-structured and meaningful. The coverage includes happy path, dedup with both `pending` and `processing` statuses, error recovery, undefined polling service, and query parameter validation. The `makeDeps` helper is clean and the `as never` casts are acceptable for test mocks.

### Overall

The implementation is correct, the self-scheduling pattern with dedup is solid, and the dual-schedule issue from the previous review is properly resolved. The `isPolling()` status inaccuracy is minor and informational-only.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped iteration. All three issues flagged in the previous review (dual-schedule bug, over-fetching dedup query, test assertion mismatch) have been correctly addressed. Here is a thorough pass over the updated code.

---

### What was fixed correctly

**Dual-schedule eliminated.** `GitHubService.start()` now accepts `useJobQueueScheduler` and skips `pollingService.start()` when true. `app.ts` sets `useJobQueueScheduler: true` whenever `githubPollingInterval > 0`, so the `setInterval`-based schedule is never activated in production. The job-queue chain is the sole scheduler.

**Dedup query tightened.** `limit: 10` is now `limit: 1`. The check only needs to know whether any job exists, not how many, so this is correct and efficient.

**Test assertion updated.** The test now asserts `listArg.limit === 1`, matching the implementation.

---

### Remaining observations

**P3 — Redundant guard in `triggerPoll` (polling-service.ts:146).** `triggerPoll` checks `this.isPolling` before delegating to `pollAllRepositories()`, which itself checks `this.isPolling` at line 157 before setting the flag. The guard in `triggerPoll` is harmless but redundant. Not a bug; just a minor clarity nit.

**P3 — Uncaught `enqueue` exception breaks the self-scheduling chain (github-poll.handler.ts:60-64).** If `jobQueue.enqueue()` throws (e.g., a transient SQLite constraint error), the exception propagates out of `handleGitHubPoll`. The `JobQueueProcessor.processJob` catch block will call `repo.fail()` on the current job, which re-queues it for retry with exponential backoff — so the chain does eventually recover through the retry mechanism rather than dying entirely. This is an acceptable trade-off; wrapping the enqueue in its own try/catch with logging would make the recovery path more visible, but the current behavior is not unsafe.

**P3 — `isPolling()` name ambiguity (github-service.ts:607 and polling-service.ts:40).** `GitHubService.isPolling()` (public method) checks whether the `setInterval` timer is running, via `pollingService.isRunning()`. In job-queue mode the timer is never started, so `gitHubService.isPolling()` always returns `false` even when active polling is happening. Any callers relying on `isPolling()` to mean "is polling currently configured and active" will get a misleading answer. The private field `this.isPolling` in `GitHubPollingService` tracks a different thing (in-flight network request). This existed before the PR and is not introduced by this change, but it is worth noting for future callers of the new code path.

**P2 — No test for the `app.ts` seed-job dedup path.** The seeding logic in `app.ts` (lines 371-381) re-checks for existing pending/processing jobs before enqueuing, which is correct for crash-recovery and restart scenarios. This path is untested. It is synchronous startup code and difficult to unit-test in isolation, so the omission is understandable; a brief integration test or at least a comment explaining the crash-recovery intent would help. The comment is actually already there (`// The self-scheduling chain in the handler keeps the chain alive after this seed`), so this is a minor gap rather than a blocking concern.

---

### Summary

The three P0/P1 issues from the prior review are resolved. The implementation is correct: single scheduler, correct dedup, proper lazy resolution of `pollingService` at job execution time rather than registration time (handling the `setupRPCHandlers` → `gitHubService.start()` ordering dependency cleanly). The remaining observations are P2/P3 nits that do not block merging.

PR #632, opened 2026-03-22
test: add JobQueueProcessor lifecycle integration tests

lsm:
Covers processor behavioral contracts in isolation:
- eager reclaimStale() call during start() before the first poll interval
- start → poll → dequeue → dispatch → stop full lifecycle
- stop() drains in-flight jobs before resolving
- setChangeNotifier callback on completed/retried/dead status transitions
- error → retry → dead full sequence (including multi-step with run_at override)
- stale job reclamation resets to pending with 60 s throttle enforcement

### DIFF
--- a/packages/daemon/tests/unit/storage/job-queue-processor-lifecycle.test.ts
+++ b/packages/daemon/tests/unit/storage/job-queue-processor-lifecycle.test.ts
@@ -1,0 +1,523 @@
+/**
+ * JobQueueProcessor Lifecycle Integration Tests
+ *
+ * Verifies the behavioral contracts that the app-level wiring depends on:
+ * - start() → eager reclaimStale() → poll → dequeue → dispatch → stop()
+ * - setChangeNotifier callback on status transitions
+ * - error → retry → dead progression
+ * - stale reclamation timing
+ *
+ * Scope: processor internals and contract, NOT app-level wiring (see app/job-queue-lifecycle.test.ts).
+ */
+
+import { describe, expect, it, beforeEach, afterEach } from 'bun:test';
+import { Database } from 'bun:sqlite';
+import { JobQueueRepository } from '../../../src/storage/repositories/job-queue-repository';
+import { JobQueueProcessor } from '../../../src/storage/job-queue-processor';
+
+const DB_SCHEMA = `
+	CREATE TABLE IF NOT EXISTS job_queue (
+		id TEXT PRIMARY KEY,
+		queue TEXT NOT NULL,
+		status TEXT NOT NULL DEFAULT 'pending'
+			CHECK(status IN ('pending', 'processing', 'completed', 'failed', 'dead')),
+		payload TEXT NOT NULL DEFAULT '{}',
+		result TEXT,
+		error TEXT,
+		priority INTEGER NOT NULL DEFAULT 0,
+		max_retries INTEGER NOT NULL DEFAULT 3,
+		retry_count INTEGER NOT NULL DEFAULT 0,
+		run_at INTEGER NOT NULL,
+		created_at INTEGER NOT NULL,
+		started_at INTEGER,
+		completed_at INTEGER
+	);
+	CREATE INDEX IF NOT EXISTS idx_job_queue_dequeue ON job_queue(queue, status, priority DESC, run_at ASC);
+	CREATE INDEX IF NOT EXISTS idx_job_queue_status ON job_queue(status);
+`;
+
+/** Wait for async side-effects to settle */
+const flush = (ms = 80) => new Promise((resolve) => setTimeout(resolve, ms));
+
+describe('JobQueueProcessor — lifecycle contracts', () => {
+	let db: Database;
+	let repo: JobQueueRepository;
+
+	beforeEach(() => {
+		db = new Database(':memory:');
+		db.exec(DB_SCHEMA);
+		repo = new JobQueueRepository(db as any);
+	});
+
+	afterEach(() => {
+		db.close();
+	});
+
+	// ─── Eager stale reclamation on start() ────────────────────────────────────
+
+	describe('eager stale reclamation on start()', () => {
+		it('calls reclaimStale() synchronously during start(), before the first interval tick', async () => {
+			// Use a very long poll interval so the interval tick cannot interfere.
+			const processor = new JobQueueProcessor(repo, {
+				staleThresholdMs: 500,
+				pollIntervalMs: 60_000,
+			});
+			processor.register('test-q', async () => {});
+
+			let reclaimCallCount = 0;
+			const original = repo.reclaimStale.bind(repo);
+			repo.reclaimStale = (staleBefore: number) => {
+				reclaimCallCount++;
+				return original(staleBefore);
+			};
+
+			// reclaimStale must NOT be called before start()
+			expect(reclaimCallCount).toBe(0);
+
+			processor.start();
+			// After start() returns (synchronously), reclaimStale should already have been invoked.
+			expect(reclaimCallCount).toBeGreaterThanOrEqual(1);
+
+			await processor.stop();
+		});
+
+		it('reclaims processing jobs that exceeded staleThresholdMs instantly on start()', async () => {
+			// Simulate a job left in processing state from a prior crash.
+			const job = repo.enqueue({ queue: 'crash-q', payload: { src: 'crash' } });
+			// Manually mark as processing with a very old started_at.
+			db.prepare(`UPDATE job_queue SET status = 'processing', started_at = ? WHERE id = ?`).run(
+				Date.now() - 30_000,
+				job.id
+			);
+			expect(repo.getJob(job.id)?.status).toBe('processing');
+
+			const processor = new JobQueueProcessor(repo, {
+				staleThresholdMs: 1_000, // threshold = 1 s; job started 30 s ago → stale
+				pollIntervalMs: 60_000,
+			});
+			processor.register('crash-q', async () => {});
+
+			// start() eagerly reclaims → job transitions to pending → immediate tick processes it.
+			processor.start();
+			await flush();
+			await processor.stop();
+
+			// The stale job should have been reclaimed and processed.
+			const after = repo.getJob(job.id);
+			expect(after?.status).toBe('completed');
+		});
+
+		it('does not reclaim jobs whose started_at is within the stale threshold', async () => {
+			const job = repo.enqueue({ queue: 'fresh-q', payload: {} });
+			// Mark as processing with a very recent started_at (well within threshold).
+			db.prepare(`UPDATE job_queue SET status = 'processing', started_at = ? WHERE id = ?`).run(
+				Date.now(),
+				job.id
+			);
+
+			const processor = new JobQueueProcessor(repo, {
+				staleThresholdMs: 60_000, // threshold = 60 s; job started just now → fresh
+				pollIntervalMs: 60_000,
+			});
+			processor.register('fresh-q', async () => {});
+
+			processor.start();
+			await flush();
+			await processor.stop();
+
+			// Fresh job must stay in processing (no handler dequeued it after reclaim).
+			expect(repo.getJob(job.id)?.status).toBe('processing');
+		});
+	});
+
+	// ─── start() → poll → dequeue → dispatch lifecycle ────────────────────────
+
+	describe('start() → poll → dequeue → dispatch', () => {
+		it('processes a job enqueued before start() via the initial tick', async () => {
+			let dispatched = false;
+			const processor = new JobQueueProcessor(repo, { pollIntervalMs: 60_000 });
+			processor.register('pre-q', async () => {
+				dispatched = true;
+			});
+
+			repo.enqueue({ queue: 'pre-q', payload: {} });
+			processor.start(); // triggers immediate tick
+			await flush();
+			await processor.stop();
+
+			expect(dispatched).toBe(true);
+		});
+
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Summary

All 22 tests pass and the test file is well-structured with clear documentation. However, there is substantial overlap with the existing `job-queue-processor.test.ts`, and a couple of timing-sensitive tests could be flaky.

### P2: Significant duplication with existing test file

The existing `packages/daemon/tests/unit/storage/job-queue-processor.test.ts` already covers nearly identical scenarios:

| New lifecycle test | Existing test in `job-queue-processor.test.ts` |
|---|---|
| "calls reclaimStale() synchronously during start()" | "does not wait for STALE_CHECK_INTERVAL before first reclamation" (line 347) |
| "reclaims processing jobs that exceeded staleThresholdMs instantly on start()" | "reclaims stale processing jobs immediately when start() is called" (line 317) |
| "does not reclaim jobs whose started_at is within the stale threshold" | "does not reclaim jobs that are not old enough" (line 468) |
| "processes a job enqueued before start()" | "start() begins polling and processes enqueued jobs" (line 371) |
| "marks a successfully handled job as completed" | "marks job as completed after handler succeeds" (line 161) |
| "delivers the correct job object with payload" | "handler receives the correct job object with payload" (line 92) |
| "resolves only after all in-flight jobs have finished" | "stop() resolves after in-flight jobs complete" (line 406) |
| "does not process new jobs after stop()" | "stop() stops polling" (line 388) |
| changeNotifier tests (4 tests) | changeNotifier tests (4 tests, line 530) |
| stale throttling | "does not run stale check again within STALE_CHECK_INTERVAL" (line 486) |
| error/retry/dead tests | handler failure tests (line 200) |

Approximately 18 of 22 tests have direct equivalents. The genuinely new tests are:
- "resolves immediately when there are no in-flight jobs" (line 232)
- "marks job as dead after all retries are exhausted (full retry sequence)" (line 344) - the multi-step retry with run_at override
- "handles non-Error throws by converting to string" (line 392)
- "resets stale job to pending (not directly to completed) before the handler picks it up" (line 441) - verifying the intermediate pending state

Consider consolidating the unique tests into the existing file, or documenting clearly why both files need to exist. The file header comment references `app/job-queue-lifecycle.test.ts` for "app-level wiring" vs "processor internals", but `job-queue-processor.test.ts` already tests processor internals.

### P3: Timing-sensitive test could be flaky

At lines 150-168, the test "processes a job enqueued after start() via the poll interval" uses `flush(10)` to try to ensure the initial tick completes before the enqueue, then `flush(80)` to catch the 30ms poll interval. On a slow CI machine, the 10ms delay may not be sufficient for the initial tick to complete and the timing assumptions could break. The existing test file avoids this pattern by using `tick()` directly.

### P3: Minor inconsistency in flush() default

The new file uses `flush(ms = 80)` as default while the existing file uses `flush()` with 50ms. Not a bug, but the inconsistency may cause confusion when reading both files together.

### What works well

- The full retry sequence test (line 344) with manual `run_at` override to avoid the exponential backoff delay is a clever approach and tests a genuine gap.
- The non-Error throw test (line 392) verifies `String(err)` coercion, which is an edge case worth covering.
- The intermediate-state verification test (line 441) that confirms reclaim goes through pending before completed is a good contract test.
- Proper cleanup in afterEach with `db.close()`.

### Recommendation: APPROVE

The duplication is notable but not blocking - the unique tests (4 of 22) add genuine value, and having a lifecycle-focused test file with its own framing is a reasonable organizational choice. The timing concerns are minor given the existing codebase patterns.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — substantive duplication, one logical bug in the throttle test, and missing cleanup in `afterEach`.

---

## Summary

All 22 tests pass and the happy-path coverage is solid. However, the new file substantially duplicates `job-queue-processor.test.ts` that already lives in the same directory, several tests contain correctness issues or fragile timing assumptions, and the test file leaks live `setInterval` timers when tests exit early before calling `stop()`.

---

## Issues

### P1 — Substantial overlap with the existing `job-queue-processor.test.ts`

The existing file already covers:

| New test | Existing equivalent |
|---|---|
| eager stale reclamation — `reclaimCallCount >= 1` | `does not wait for STALE_CHECK_INTERVAL before first reclamation` |
| reclaims stale job on start | `reclaims stale processing jobs immediately when start() is called` |
| does not reclaim fresh jobs | `does not reclaim jobs that are not old enough` |
| dispatches payload correctly | `handler receives the correct job object with payload` |
| marks job completed | `marks job as completed after handler succeeds` |
| stop drains in-flight | `stop() resolves after in-flight jobs complete` |
| stop resolves immediately | implicit in every test's `afterEach` |
| no new jobs after stop | `stop() stops polling` |
| notifier on complete | `notifier is called after job completes` |
| notifier on dead | `notifier is called after job fails` |
| notifier on retry | (partial: the existing file checks notifier on fail generically) |
| notifier not called when idle | `notifier is not called when no jobs processed` |
| increments retryCount | `retries job when handler throws and retries remaining` |
| error message stored | `stores error message in job.error` |
| throttle stale check | `does not run stale check again within STALE_CHECK_INTERVAL` |
| fresh job stays processing | `does not reclaim jobs that are not old enough` |

Roughly 15 of 22 tests are near-identical to coverage already present. The net new signal is limited to: the full exhaust-retry-dead sequence (lines 344–376), the non-Error throw conversion (lines 392–407), and the sequential reclaim-then-pending ordering assertion (lines 441–474).

### P1 — Missing `afterEach` cleanup causes timer leaks

The new file does **not** call `processor.stop()` in `afterEach`. In tests that do not reach `await processor.stop()` (e.g. if an assertion throws first), the `setInterval` timer created by `start()` is never cleared. This can cause subsequent tests to observe spurious side effects or keep Bun's event loop alive. The existing `job-queue-processor.test.ts` correctly places `await processor.stop()` in `afterEach`. The new file should do the same.

Affected location: `afterEach` block at line 52-54 — it only closes `db`, never stops the processor.

### P1 — Throttle test is logically inverted for the eager `start()` path

**File:** `packages/daemon/tests/unit/storage/job-queue-processor-lifecycle.test.ts`, lines 476–499 (`throttles stale checks to at most once per STALE_CHECK_INTERVAL`)

The test calls `processor.tick()` directly **without** calling `processor.start()` first. Because `start()` sets `lastStaleCheck = Date.now()` (line 41 of `job-queue-processor.ts`), and this test never calls `start()`, `lastStaleCheck` remains 0. That means the **second** call to `tick()` will also see `now - 0 >= 60_000` as false only if the second tick runs within 60 s of the first tick — which it always will.

The issue: the test is actually relying on the fact that the **first** tick sets `lastStaleCheck` via `checkStaleJobs`, and the second tick runs within 60 s. This is correct but coincidental. The test does **not** exercise the eager `start()` path at all — after `start()`, `lastStaleCheck` is set, so the **first** `tick()` call via the interval would already be throttled. The description in the comment says `lastStaleCheck=0` but this is only true because `start()` was never called. This is a misleading setup that could mask a regression if `start()` is later changed to not reset `lastStaleCheck`.

### P2 — `flush(80)` timing is fragile and non-deterministic

**Lines 163–166** (`processes a job enqueued after start() via the poll interval`):



A 10 ms flush to 'miss' the initial tick assumes the tick completes in < 10 ms, and an 80 ms flush assumes the interval fires at least once within that window. Under CI load, both assumptions can fail. The `exhaust-retry-dead` test at lines 344–376 has the same issue with two back-to-back `flush(50)` calls bracketing a `run_at` override — if the interval fires between the first flush completing and the `run_at` UPDATE, the retry job could be dequeued before its `run_at` is reset, causing `failCount` to be 1 instead of 2.

The existing `job-queue-processor.test.ts` avoids real interval firing entirely by using `await processor.tick()` directly. The new file mixes both styles inconsistently.

### P2 — Weak notifier assertions

**Lines 276–306** (notify-dead and notify-retry tests): both assert `tables.length > 0` but do not assert the specific `'job_queue'` string. The existing test file already does this correctly with `tables.every((t) => t === 'job_queue')`. The weaker assertion means a notifier bug that emits any arbitrary string would pass.

### P2 — Test at line 110 has a misleading comment and incorrect final state

`does not reclaim jobs whose started_at is within the stale threshold` (lines 110–130):

The final assertion is:
```ts
expect(repo.getJob(job.id)?.status).toBe('processing');
```

The comment says "Fresh job must stay in processing (no handler dequeued it after reclaim)." This is true but for a subtle reason: the processor has a handler registered for `fresh-q`, so if the job **were** reclaimed to pending, it would be dequeued and completed. The assertion happens to be correct, but the flush is only 80 ms with a 60 s poll interval, meaning the initial tick fires but the fresh job is not reclaimed (correctly). This is fine, but would silently pass even if the job was never registered — the test gives no signal about whether the handler would have fired.

### P3 — `DB_SCHEMA` constant duplicates schema from the existing test file

The schema string is copy-pasted from `job-queue-processor.test.ts`. If the schema ever changes, two files need updating. Consider extracting it to a shared test helper.

### P3 — Inconsistent flush helper

The existing file uses `const flush = () => new Promise(r => setTimeout(r, 50))` (fixed 50 ms). The new file uses `const flush = (ms = 80) => ...` (parameterized, default 80 ms). Using 80 ms as a default makes all tests slightly slower without documented justification.

lsm (commented):

**Recommendation: APPROVE** — all P1 and P2 issues from the previous review are resolved.

---

## Second-Iteration Verification

### P1 — Substantial overlap (resolved)

The file was cut from 22 tests to 7. The header comment (lines 1–16) now explicitly documents the division: `job-queue-processor.test.ts` covers individual unit behaviors; this file covers orchestration contracts. Each test is genuinely distinct from anything in the existing file.

### P1 — Missing `afterEach` cleanup (resolved)

`afterEach` at lines 58–61 now correctly calls `await processor.stop()` before `db.close()`. Timer leaks are prevented.

### P1 — Throttle test logically inverted (resolved)

The throttle test now calls `processor.start()` first (line 256), captures the reclaim count after the eager call in start(), then calls `processor.tick()` (line 262) and asserts `reclaimCallCount === countAfterStart`. This correctly exercises the post-`start()` throttle path where `lastStaleCheck = Date.now()` and the immediate tick should be suppressed.

### P2 — Weak notifier assertions (resolved)

Both notifier assertion sites (lines 187 and 198) now use `tables.every((t) => t === 'job_queue')`, matching the strength of the existing test file.

### P2 — `flush(80)` fragile timing (resolved)

The file now uses a fixed `const flush = () => new Promise((resolve) => setTimeout(resolve, 50))` (line 44, matching the existing file's 50 ms) and all tests drive execution with `await processor.tick()` directly — no interval timing dependencies remain.

---

## Remaining Observations (P3, non-blocking)

- The `DB_SCHEMA` constant (lines 23–42) is still a copy-paste of the schema in `job-queue-processor.test.ts`. A shared test helper would prevent drift, but this is cosmetic.
- The `notifier receives "job_queue" for all transitions` test (lines 161–199) sets up a handler that throws for the first two calls (`callCount < 3`) but then immediately resets `callCount = 0` before the actual assertion job is enqueued. The comment block above (lines 172–177) describing the `maxRetries=2` scenario is now orphaned and misleading since the test actually uses `maxRetries=0`. This is confusing to read but does not affect correctness.
- The `multiStepProcessor` at line 108 is created inline with a local `stop()` call at line 137. This is fine for a multi-step sequence test, but it means if an assertion before line 137 throws, the local processor is not stopped. Since `afterEach` only stops the outer `processor` variable, a test failure mid-sequence would leak the inner processor's timer. Low-risk given the test structure, but worth noting.

---

All substantive issues are resolved. The 7 tests that remain add genuine value: full exhaust-retry-dead sequence, non-Error throw coercion, reclaim-then-pending ordering, throttle contract verification, and the stop-with-no-inflight fast path.

lsm (commented):

### Recommendation: APPROVE

All previous P1 and P2 issues from the first review have been addressed:

- **Duplication resolved**: Reduced from 22 to 7 tests. The remaining tests cover genuinely distinct scenarios not present in `job-queue-processor.test.ts`: synchronous timing verification of `reclaimStale()`, the full multi-step retry-to-dead sequence, non-Error throw coercion, intermediate pending state during reclamation, stop() fast-path with zero in-flight, and throttle verification through the `start()` path.
- **Timer leak fixed**: `afterEach` now calls `await processor.stop()` before `db.close()`.
- **Throttle test fixed**: Now calls `processor.start()` first (which sets `lastStaleCheck`), then verifies subsequent `tick()` skips the stale check. This correctly exercises the real-world scenario where `start()` sets the timestamp and interval ticks should be throttled.
- **Timing fragility eliminated**: All tests use `await processor.tick()` directly instead of relying on interval timing with arbitrary `flush()` delays.
- **Notifier assertions strengthened**: Uses `tables.every(t => t === 'job_queue')` to verify the exact table name string.

**P3: Noisy comment block in the notifier test (lines 172-177)**

The comment describes a `maxRetries=2` scenario with three failure cycles, then pivots to "We test with maxRetries=0 for simplicity." The abandoned scenario description adds cognitive overhead without value. A shorter comment explaining just the actual test would be clearer.

**P3: Schema duplication**

`DB_SCHEMA` is copy-pasted from `job-queue-processor.test.ts`. If the schema changes, two files need updating. Minor risk given the schema is stable, but worth noting.

### Verification

All 7 tests pass (20 assertions, 434ms). Double-calling `stop()` (once in the test, once in `afterEach`) is safe since the second call is a no-op when `pollTimer` is already null and `inFlight` is 0.

PR #633, opened 2026-03-22
test: verify eager stale job reclamation on processor restart

lsm:
Add dedicated test suite covering the crash-and-restart scenario for
JobQueueProcessor. Tests verify that stale processing jobs (left by a
crashed processor) are reclaimed immediately when a new processor starts
via the eager reclaimStale() call in start(), not after the 60s interval.

Scenarios covered:
- Stale job reclaimed and re-processed immediately on start()
- reclaimStale() called synchronously inside start() before any tick
- Reclaimed job receives correct payload from handler
- Non-stale (recently-started) processing jobs are not reclaimed
- Multiple stale jobs from different queues all reclaimed on restart
- Pending jobs coexist correctly with stale reclamation on restart

### DIFF
--- a/packages/daemon/tests/unit/storage/job-queue-processor.test.ts
+++ b/packages/daemon/tests/unit/storage/job-queue-processor.test.ts
@@ -313,60 +313,6 @@ describe('JobQueueProcessor', () => {
 		});
 	});
 
-	describe('eager stale reclamation on start()', () => {
-		it('reclaims stale processing jobs immediately when start() is called', async () => {
-			const eagerProcessor = new JobQueueProcessor(repo, {
-				staleThresholdMs: 1000,
-				pollIntervalMs: 5000, // long interval so the interval tick doesn't interfere
-			});
-			eagerProcessor.register('eager-queue', async () => {});
-
-			// Enqueue and dequeue manually to create a "processing" job
-			const job = repo.enqueue({ queue: 'eager-queue', payload: {} });
-			repo.dequeue('eager-queue', 1);
-
-			// Confirm it's processing
-			expect(repo.getJob(job.id)?.status).toBe('processing');
-
-			// Make it stale (started 10s ago, threshold is 1s)
-			db.prepare(`UPDATE job_queue SET started_at = ? WHERE id = ?`).run(
-				Date.now() - 10_000,
-				job.id
-			);
-
-			// start() should eagerly reclaim before the first tick
-			eagerProcessor.start();
-			// Give the immediate tick time to process the reclaimed job
-			await flush();
-			await eagerProcessor.stop();
-
-			const after = repo.getJob(job.id);
-			expect(after?.status).toBe('completed');
-		});
-
-		it('does not wait for STALE_CHECK_INTERVAL before first reclamation', async () => {
-			// Create a spy on reclaimStale to verify it's called during start()
-			let reclaimCallCount = 0;
-			const originalReclaim = repo.reclaimStale.bind(repo);
-			repo.reclaimStale = (staleBefore: number) => {
-				reclaimCallCount++;
-				return originalReclaim(staleBefore);
-			};
-
-			const eagerProcessor = new JobQueueProcessor(repo, {
-				staleThresholdMs: 1000,
-				pollIntervalMs: 5000,
-			});
-			eagerProcessor.register('spy-queue', async () => {});
-
-			eagerProcessor.start();
-			await eagerProcessor.stop();
-
-			// reclaimStale should have been called during start() (before any tick interval)
-			expect(reclaimCallCount).toBeGreaterThanOrEqual(1);
-		});
-	});
-
 	describe('start / stop', () => {
 		it('start() begins polling and processes enqueued jobs', async () => {
 			const shortProcessor = new JobQueueProcessor(repo, { pollIntervalMs: 50 });
--- a/packages/daemon/tests/unit/storage/job-queue-stale-reclamation.test.ts
+++ b/packages/daemon/tests/unit/storage/job-queue-stale-reclamation.test.ts
@@ -1,0 +1,245 @@
+/**
+ * Tests for stale job reclamation on daemon restart.
+ *
+ * Focuses on the eager reclaimStale() call added in Task 1.2 to `JobQueueProcessor.start()`.
+ * This ensures that jobs stuck in `processing` due to a crash are reclaimed IMMEDIATELY
+ * on restart — not after the 60-second STALE_CHECK_INTERVAL delay.
+ *
+ * The "restart" is simulated by:
+ *   1. Using one processor instance to dequeue a job (marking it `processing`).
+ *   2. Abandoning that processor without calling stop() — simulating a crash.
+ *   3. Creating a SECOND processor on the same DB and calling start() on it.
+ *   4. Asserting the stale job is reclaimed and re-processed within the first tick.
+ */
+import { describe, expect, it, beforeEach, afterEach } from 'bun:test';
+import { Database } from 'bun:sqlite';
+import { JobQueueRepository } from '../../../src/storage/repositories/job-queue-repository';
+import { JobQueueProcessor } from '../../../src/storage/job-queue-processor';
+
+const DB_SCHEMA = `
+	CREATE TABLE IF NOT EXISTS job_queue (
+		id TEXT PRIMARY KEY,
+		queue TEXT NOT NULL,
+		status TEXT NOT NULL DEFAULT 'pending'
+			CHECK(status IN ('pending', 'processing', 'completed', 'failed', 'dead')),
+		payload TEXT NOT NULL DEFAULT '{}',
+		result TEXT,
+		error TEXT,
+		priority INTEGER NOT NULL DEFAULT 0,
+		max_retries INTEGER NOT NULL DEFAULT 3,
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured test suite that correctly verifies the eager stale job reclamation behavior on processor restart. All 6 tests pass and the scenarios are appropriate for the task.

### Strengths

- **Correct crash simulation pattern**: Using one processor to dequeue (marking `processing`), abandoning it, then creating a fresh processor on the same in-memory DB is a clean and realistic crash simulation.
- **Synchronous reclaim verification**: The spy test (test 2) correctly verifies that `reclaimStale()` is called synchronously within `start()` by checking `reclaimCallCount` immediately after `start()` returns, before any `await`. This directly validates the eager behavior.
- **Non-stale boundary test**: Test 4 correctly uses a 60s threshold with a just-dequeued job (started_at ~now) to verify that non-stale jobs are not incorrectly reclaimed. The threshold math is sound.
- **Good test isolation**: Each test creates its own in-memory DB and properly cleans up via `afterEach`. The `restartedProcessor` tracking pattern ensures processors are stopped even if a test fails early.
- **Consistent with existing patterns**: The DB schema, `flush()` helper, `db as any` cast, and general structure mirror the existing `job-queue-processor.test.ts` file.

### Minor Observations (P3)

1. **Overlap with existing test file**: The existing `job-queue-processor.test.ts` already has two tests under `describe('eager stale reclamation on start()')` (lines 316-368) that cover similar ground — specifically the "reclaims stale processing jobs immediately" and "does not wait for STALE_CHECK_INTERVAL" scenarios. The new file's tests 1 and 2 are conceptually the same but with the improved crash simulation (two processor instances). This isn't a problem per se — the new file is more thorough — but worth noting the duplication.

2. **DB schema duplication**: The `DB_SCHEMA` constant is duplicated across both test files. A shared test fixture would reduce this, but for two files this is perfectly fine.

3. **`crashedProcessor` never started**: In test 1, `crashedProcessor` is created and has `register()` called, but `start()` is never called on it. The dequeue is done via `repo.dequeue()` directly. This is intentional and correct (we just need a processing row in the DB), but the `crashedProcessor` object itself is unused dead code — the test would work identically without creating it. The same pattern applies to the comment in the file header mentioning "using one processor instance to dequeue a job" when the processor is never actually used for dequeuing.

All tests are correct, properly isolated, and effectively verify the eager reclamation behavior. No bugs, no security concerns, no missing edge cases of significance.

lsm (commented):

**Recommendation: APPROVE** (own PR — commenting only)

---

The new test file correctly targets the eager reclamation behavior and all six scenarios are structurally sound. A few things worth noting:

### Significant Duplication With Existing Tests (P2)

`packages/daemon/tests/unit/storage/job-queue-processor.test.ts` already contains a `describe('eager stale reclamation on start()')` block (lines 316–368) with two tests that are nearly identical to tests 1 and 2 in this new file:

- "reclaims stale processing jobs immediately when start() is called" (`job-queue-processor.test.ts` line 317) vs. "reclaims a stale processing job immediately on start() without waiting 60 s" (new file line 64) — same structure, same assertions, same `staleThresholdMs: 1_000` / `pollIntervalMs` pattern.
- "does not wait for STALE_CHECK_INTERVAL before first reclamation" (`job-queue-processor.test.ts` line 347) vs. "calls reclaimStale() synchronously during start() before any tick interval fires" (new file line 103) — both spy on `repo.reclaimStale` and assert `count >= 1`.

Tests 3–6 are genuinely additive (handler-receives-payload, non-stale boundary, multi-queue, coexistence with pending). Tests 1–2 duplicate what the existing file already covers. If the intent was to co-locate all restart-scenario tests in one file, the existing tests in `job-queue-processor.test.ts` should be removed.

### Orphaned Timer in Test 1 (P1)

In the crash-and-restart test (line 64), `crashedProcessor` is created with `pollIntervalMs: 60_000` and never stopped:

```ts
const crashedProcessor = new JobQueueProcessor(repo, {
    staleThresholdMs: 1_000,
    pollIntervalMs: 60_000,
});
// ...
// "Crash" — we simply do NOT call crashedProcessor.stop() and discard it.
```

The comment frames this as intentional crash simulation, and functionally it is fine — the 60-second interval cannot fire within a 50ms flush. However, the `setInterval` handle is silently leaked for the duration of the Bun process. The idiomatic way to simulate a crash without leaking timers would be to call `stop()` on the crashed processor *after* the stale job has been picked up (i.e., before the restart phase), since what you really want is a job stuck in `processing` state — not an active timer. Alternatively, the comment should explicitly acknowledge the leak and confirm Bun cleans it up at test file exit.

### Spy Pattern Is Correct (No Issue)

Test 2 patches `repo.reclaimStale` before instantiating the processor, then asserts `reclaimCallCount >= 1` *synchronously* after `start()` returns (before any `await`). This correctly verifies that `reclaimStale` is called synchronously inside `start()` — the existing implementation at line 40 of `job-queue-processor.ts` does exactly this. The assertion is tight and accurate.

### Non-Stale Boundary Relies on Wall Clock (P3)

Test 4 ("does NOT reclaim a recently-started processing job") sets `staleThresholdMs: 60_000` and dequeues a job (setting `started_at = ~now`). The test assumes the job's `started_at` stays within the 60-second threshold during execution — safe in practice, but formally a wall-clock dependency. The existing `job-queue-processor.test.ts` test at line 468 ("does not reclaim jobs that are not old enough") covers the same boundary. If either test were to run on an extremely slow CI machine and wall time advanced 60+ seconds during test execution, both would fail — an acceptable risk but worth acknowledging.

### Schema Duplication (P3)

`DB_SCHEMA` is copy-pasted verbatim in the new file (line 19), in `job-queue-processor.test.ts`, and in `job-queue-lifecycle.test.ts`. This is consistent with the established project pattern for unit tests, so not a blocker, but a shared helper module would reduce maintenance surface.

### Test 6 Dequeue-Ordering Assumption Is Sound

The comment on line 213 correctly explains that the stale job is enqueued first so that `dequeue` (which uses `ORDER BY priority DESC, run_at ASC`) picks it up via the first `dequeue` call. This logic is correct given the repository implementation.

---

**Summary:** The test suite correctly verifies eager reclamation on restart. The critical finding is that tests 1 and 2 duplicate existing coverage in `job-queue-processor.test.ts` — either those existing tests should be removed, or the new file should start from test 3. The orphaned timer in test 1 is cosmetically inconsistent with the crash-simulation framing but is functionally harmless. Tests 3–6 are solid additions.

lsm (commented):

Re-review after addressing previous feedback. Both issues have been correctly resolved:

**P1 fix (leaked setInterval):** The `crashedProcessor` object that leaked a `setInterval` handle has been removed. The crash scenario is now simulated entirely via direct DB manipulation (backdating `started_at`), which is cleaner and avoids any resource leaks. Verified in the new `job-queue-stale-reclamation.test.ts`.

**P2 fix (duplicate coverage):** The `describe('eager stale reclamation on start()')` block has been removed from `job-queue-processor.test.ts`. The dedicated test file `job-queue-stale-reclamation.test.ts` now owns all eager-reclaim-on-restart scenarios, eliminating duplication.

- The new test file is well-structured with proper `beforeEach`/`afterEach` lifecycle management, including `restartedProcessor` cleanup.
- All 6 test scenarios cover meaningful distinct cases: basic reclamation, synchronous `reclaimStale()` call verification, payload preservation, non-stale job protection, multi-queue reclamation, and coexistence with pending jobs.
- The `pollIntervalMs: 60_000` pattern correctly ensures that reclamation is tested as coming from the eager `start()` call, not from interval ticks.
- All 32 tests across both files pass cleanly.

No new issues found. **Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

Both previously raised issues are correctly fixed. No new issues introduced.

### Fix 1 (P1): setInterval leak from crashedProcessor — FIXED

The leaked interval handle is gone. The new test file never creates a "crashed" processor object. Instead, it simulates a crash by directly manipulating the DB row (`UPDATE job_queue SET started_at = ?`), then creates a single fresh `restartedProcessor`. The `afterEach` correctly stops it if non-null, and test 2 manually sets `restartedProcessor = null` after its explicit `stop()` call to prevent a redundant second stop in `afterEach`. No handle leaks.

### Fix 2 (P2): Duplicate coverage block removed — FIXED

The `describe('eager stale reclamation on start()')` block has been removed from `job-queue-processor.test.ts` (313 lines reduced to 313→257). The new dedicated file `job-queue-stale-reclamation.test.ts` replaces it with 6 well-structured tests that more clearly model the restart scenario by using a fresh processor against a pre-seeded DB state.

- Test 2's synchronous check (`expect(reclaimCallCount).toBeGreaterThanOrEqual(1)` immediately after `restartedProcessor.start()` returns, before any `await`) correctly validates that the implementation calls `repo.reclaimStale()` synchronously in `start()`, not inside the interval callback or the async `tick()`.
- The `pollIntervalMs: 60_000` used throughout the new tests reliably ensures job completion observed during `flush()` comes only from the eager `start()` path, not from an interval tick firing in 50 ms.
- The DB schema is duplicated between the two test files. This is a minor style concern but not a correctness issue — each test file uses its own isolated in-memory DB and is self-contained.
- The 50 ms `flush()` timeout is consistent with the existing test file and is sufficient for the synchronous handler chain used in these tests.

PR #635, opened 2026-03-22
feat: add job_queue.cleanup handler and wire into daemon startup

lsm:
- Create cleanup.handler.ts: removes completed/dead jobs older than 7 days,
  self-schedules next run in 24h, deduplicates pending cleanup jobs
- Register handler in app.ts BEFORE jobProcessor.start() to avoid
  "No handler registered" errors for jobs pending from previous runs
- Enqueue initial cleanup job on first boot if none is already pending
- Add unit tests covering: old job deletion, dead job deletion,
  self-scheduling, dedup, and no-op when nothing is stale

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -22,6 +22,8 @@ import type { SpaceRuntimeService } from './lib/space/runtime/space-runtime-serv
 import type { TaskAgentManager } from './lib/space/runtime/task-agent-manager';
 import { JobQueueRepository } from './storage/repositories/job-queue-repository';
 import { JobQueueProcessor } from './storage/job-queue-processor';
+import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
+import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -360,6 +362,21 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	if (gitHubService) {
 		gitHubService.start();
 		logInfo('[Daemon] GitHub service started');
+	}
+
+	// Register job handlers BEFORE starting the processor so no pending job
+	// from a previous run is dequeued without a handler available.
+	jobProcessor.register(JOB_QUEUE_CLEANUP, createCleanupHandler(jobQueue));
+
+	// Enqueue the initial cleanup job if none is already pending.
+	const pendingCleanup = jobQueue.listJobs({
+		queue: JOB_QUEUE_CLEANUP,
+		status: 'pending',
+		limit: 1,
+	});
+	if (pendingCleanup.length === 0) {
+		jobQueue.enqueue({ queue: JOB_QUEUE_CLEANUP, payload: {}, runAt: Date.now() });
+		logInfo('[Daemon] Enqueued initial job_queue.cleanup job');
 	}
 
 	// Start job queue processor last (after all handler registrations)
--- a/packages/daemon/src/lib/job-handlers/cleanup.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/cleanup.handler.ts
@@ -1,0 +1,21 @@
+import type { Job, JobQueueRepository } from '../../storage/repositories/job-queue-repository';
+import { JOB_QUEUE_CLEANUP } from '../job-queue-constants';
+
+const DEFAULT_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000; // 7 days
+const NEXT_RUN_DELAY_MS = 24 * 60 * 60 * 1000; // 24 hours
+
+export function createCleanupHandler(jobQueue: JobQueueRepository) {
+	return async (_job: Job): Promise<{ deletedJobs: number; nextRunAt: number }> => {
+		const deletedJobs = jobQueue.cleanup(Date.now() - DEFAULT_MAX_AGE_MS);
+
+		const nextRunAt = Date.now() + NEXT_RUN_DELAY_MS;
+
+		// Self-schedule: only enqueue next cleanup if none is already pending
+		const pending = jobQueue.listJobs({ queue: JOB_QUEUE_CLEANUP, status: 'pending', limit: 1 });
+		if (pending.length === 0) {
+			jobQueue.enqueue({ queue: JOB_QUEUE_CLEANUP, payload: {}, runAt: nextRunAt });
+		}
+
+		return { deletedJobs, nextRunAt };
+	};
+}
--- a/packages/daemon/src/storage/repositories/job-queue-repository.ts
+++ b/packages/daemon/src/storage/repositories/job-queue-repository.ts
@@ -177,8 +177,13 @@ export class JobQueueRepository {
 	}
 
 	cleanup(beforeMs: number): number {
+		// 'failed' is included defensively: the processor never writes it (retries go back to
+		// 'pending' and exhausted retries become 'dead'), but the type contract allows it and
+		// future code could produce it. Including it prevents indefinite accumulation.
 		const result = this.db
-			.prepare(`DELETE FROM job_queue WHERE status IN ('completed', 'dead') AND completed_at < ?`)
+			.prepare(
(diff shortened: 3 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped implementation that meets all acceptance criteria. The handler, wiring, and tests are straightforward and correct.

### What looks good

- Handler is registered BEFORE `jobProcessor.start()` in `app.ts` (line 369), preventing the "no handler" race condition.
- The dedup check in both the handler (self-scheduling) and the boot path (initial enqueue) correctly prevents duplicate pending cleanup jobs.
- The `cleanup()` method in `JobQueueRepository` (line 179-184) targets only `completed` and `dead` jobs with `completed_at < beforeMs`, which is exactly right.
- Tests cover the key behaviors: old completed deletion, old dead deletion, self-scheduling timing, dedup, and no-op.

### Minor observations (P3)

1. **`nextRunAt` return value may not match the actually scheduled time when dedup fires** (`cleanup.handler.ts` line 11): When a pending cleanup job already exists (dedup branch), the handler still returns `nextRunAt = Date.now() + 24h`, but the actual pending job's `runAt` could be different (e.g., from the previous schedule). This is cosmetic since the return value is only used as the job result stored in the DB and not acted upon, but it could be slightly misleading if someone inspects job results for debugging.

2. **Test schema is duplicated from the migration** (`cleanup-handler.test.ts` lines 11-31): The `CREATE TABLE` in the test duplicates the production schema. If the schema evolves (e.g., new columns), this test helper could drift. This is a common pattern in this codebase though, so not a blocker.

3. **No test for the `app.ts` wiring itself**: The boot-time logic (check for pending, enqueue if absent) is only tested indirectly through the handler tests. An integration test that verifies the initial enqueue on first boot would add confidence, but is not strictly necessary given the simplicity of the logic.

All observations are P3-level and do not block merge.

lsm (commented):

**Recommendation: APPROVE** — the implementation is correct and complete against all stated acceptance criteria. I found one omission worth noting and a few minor points, but nothing that blocks merging.

---

### What was checked

- `packages/daemon/src/lib/job-handlers/cleanup.handler.ts` — handler logic
- `packages/daemon/src/app.ts` — registration and boot-time enqueue
- `packages/daemon/tests/unit/job-handlers/cleanup-handler.test.ts` — unit tests
- `packages/daemon/src/storage/repositories/job-queue-repository.ts` — `cleanup()` SQL and `fail()` lifecycle
- `packages/daemon/src/storage/job-queue-processor.ts` — `JobHandler` type contract and registration

---

### Acceptance criteria — all satisfied

- Cleanup handler deletes `completed`/`dead` jobs older than 7 days — verified via `jobQueue.cleanup(Date.now() - 7_days)` call.
- Self-schedules next run 24 h out — correct, with proper pending-check dedup.
- Dedup prevents multiple pending cleanup jobs — checked with `listJobs({ status: 'pending', limit: 1 })`.
- Handler registered BEFORE `jobProcessor.start()` — confirmed in `app.ts` lines 369/383.
- Initial cleanup job enqueued on first boot if none is pending — confirmed in `app.ts` lines 372-380.
- 5 unit tests covering deletion, dead-job deletion, self-scheduling, dedup, and no-op — all present and meaningful.

---

### Issues

**P2 — `failed` status never pruned**

The DB schema allows `status = 'failed'` (see `JobStatus` type in `job-queue-repository.ts` line 4 and the `CHECK` constraint in the test schema at line 14 of the test file). The `cleanup()` SQL only deletes rows where `status IN ('completed', 'dead')`. In practice the processor never writes `failed` — it goes straight to `pending` (retry) or `dead` (exhausted) — but if any external code or a future migration writes `failed` rows they will accumulate indefinitely. The concern is about long-term correctness of the cleanup contract, not an immediate bug. Consider including `'failed'` in the cleanup SQL or documenting that `failed` is intentionally ephemeral.

**P3 — Return type is wider than `JobHandler` requires but not aligned with it**

`JobHandler` is typed `Promise<Record<string, unknown> | void>`. The handler returns `Promise<{ deletedJobs: number; nextRunAt: number }>`, which is structurally assignable and compiles without error. This is fine, but the explicit return type annotation on the inner function (`Promise<{ deletedJobs: number; nextRunAt: number }>`) is not reflected in the `register()` call — TypeScript will widen the result to `Record<string, unknown>` when `JobQueueProcessor.complete()` stores it. No data loss occurs since the result is only stored in the DB, but it is slightly misleading to read.

**P3 — No test for the boot-time enqueue in `app.ts`**

The boot-time logic (check-and-enqueue on daemon start) is only tested through the handler's own dedup test. There is no integration-level test asserting that `createDaemonApp` actually enqueues the cleanup job when none exists, or skips it when one is already present. This is consistent with how other startup behavior is handled in this codebase, so it is not a regression, but it is worth noting as a gap if this code path is ever refactored.

---

### Summary

The implementation is clean, well-tested, and correct. The registration-before-start ordering is explicitly correct. The only noteworthy omission is that `failed`-status rows (which the processor can theoretically produce via the type contract, even if it currently does not) would never be pruned. All other acceptance criteria are met.

lsm (commented):

The P2 feedback from the previous review has been properly addressed. The changes look good.

### What was fixed

- `cleanup()` SQL in `job-queue-repository.ts` now includes `'failed'` status alongside `'completed'` and `'dead'`, preventing indefinite accumulation of failed jobs.
- The defensive comment explaining why `'failed'` is included (processor never writes it, but the type contract allows it) is helpful context for future maintainers.
- New test in `job-queue-repository.test.ts` (`deletes failed jobs older than threshold`) directly verifies the repository-level fix.
- New test in `cleanup-handler.test.ts` (`deletes failed jobs older than 7 days`) verifies the end-to-end handler behavior for failed jobs.

### Overall assessment

The implementation is clean, minimal, and well-tested. The handler correctly:
- Deletes stale completed/dead/failed jobs older than 7 days
- Self-schedules the next run 24 hours later with dedup protection
- Is registered before the processor starts to avoid race conditions
- Seeds an initial job on first boot with a pending-check guard

No issues found. **Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** — the P2 fix from the previous review has been correctly applied and the new tests are solid. No new issues found.

---

### What was reviewed in this follow-up

This review focuses on the incremental changes since the prior review: the addition of `'failed'` to the cleanup SQL, and the two new tests added to cover it.

---

### P2 fix — verified correct

**`packages/daemon/src/storage/repositories/job-queue-repository.ts` line 185**

The SQL now reads:
```sql
DELETE FROM job_queue WHERE status IN ('completed', 'dead', 'failed') AND completed_at < ?
```

This correctly prevents `failed`-status rows from accumulating indefinitely. The comment on lines 180-182 is accurate and helpful: it explains why `'failed'` is included defensively even though the processor currently never writes it directly.

---

### New tests — verified meaningful

**`packages/daemon/tests/unit/storage/job-queue-repository.test.ts` — `deletes failed jobs older than threshold`**

The test inserts a `'failed'` row directly via SQL (correct approach since the processor never writes this status), then calls `cleanup()` with a threshold past the job's `completed_at`. Asserts `deleted === 1` and `getJob()` returns null. This is a complete and correct unit test for the SQL change.

**`packages/daemon/tests/unit/job-handlers/cleanup-handler.test.ts` — `deletes failed jobs older than 7 days`**

Same logic tested end-to-end through `createCleanupHandler`. Inserts an 8-day-old `'failed'` job, runs the handler, asserts `result.deletedJobs === 1` and `getJob('old-failed')` is null. Correct and sufficient coverage.

---

### No new issues introduced

- The SQL change is surgical and does not affect any other behavior.
- The comment explaining the defensive rationale is accurate and does not overstate the risk.
- Lint, knip, format, and type checks all pass on the latest commit (`65626a1`).
- The three pre-existing P3 observations from the prior review remain unchanged and do not block merge:
  1. `nextRunAt` return value may not match the actual scheduled time in the dedup branch — cosmetic only.
  2. Test schema is duplicated from the migration — common codebase pattern, not a blocker.
  3. No integration-level test for the boot-time enqueue in `app.ts` — consistent with existing codebase patterns.

---

### Summary

The worker addressed the only actionable issue from the previous review. The fix is minimal, correct, well-commented, and covered by two complementary tests. Ready to merge.

PR #638, opened 2026-03-22
feat: add sub-session members to groups (Task 2.2)

lsm:
## Summary

- Extends `SubSessionFactory.create()` with optional `SubSessionMemberInfo` (agentId, role) so `spawn_step_agent` can pass agent identity metadata to the manager
- In `createSubSessionFactory()`, after creating a sub-session, looks up the task's group via `taskGroupIds` and calls `sessionGroupRepo.addMember()` with the agent's role/agentId and `status:'active'`; `orderIndex` is set to current member count
- Tracks sub-session member IDs in new `subSessionMemberIds` map (sessionId → memberId)
- In `handleSubSessionComplete()`, calls `updateMemberStatus(memberId, 'completed')`
- In `registerCompletionCallback()`, subscribes to `session.error` to set `status:'failed'`
- Cleans up `subSessionMemberIds` entries in `cleanup()`

## Test plan

- [x] 5 new unit tests in `task-agent-manager.test.ts`:
  - Sub-session added with correct agentId, role, and status:'active'
  - Multiple sub-sessions for same task appear in same group with incremental orderIndex
  - Status transitions to 'completed' after handleSubSessionComplete fires
  - Status transitions to 'failed' on session.error event
  - addMember failure is non-fatal — sub-session is still created
- [x] 1 new test in `task-agent-tools.test.ts`: spawn_step_agent passes correct memberInfo to factory
- [x] All 75 task-agent-manager tests pass, all 61 task-agent-tools tests pass
- [x] Full daemon suite: 7071 pass (8 pre-existing failures in dev-proxy/copilot tests unrelated to this change)

### DIFF
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -642,8 +642,7 @@ export class TaskAgentManager {
 				const groupId = this.taskGroupIds.get(taskId);
 				if (groupId) {
 					try {
-						const group = this.config.sessionGroupRepo.getGroup(groupId);
-						const orderIndex = group?.members.length ?? 0;
+						const orderIndex = this.config.sessionGroupRepo.getMemberCount(groupId);
 						const member = this.config.sessionGroupRepo.addMember(groupId, sessionId, {
 							role: memberInfo?.role ?? 'agent',
 							agentId: memberInfo?.agentId,
@@ -759,11 +758,20 @@ export class TaskAgentManager {
 		);
 
 		// Subscribe to session.error to mark the group member as 'failed' when a
-		// fatal error occurs. This is independent of the completion callback.
+		// fatal error occurs. Sets `fired = true` so that a subsequent idle transition
+		// (session keeps going after an error) does not overwrite 'failed' with 'completed'.
+		// Also self-unsubscribes both listeners to prevent multiple invocations.
 		const unsubscribeError = this.config.daemonHub.on(
 			'session.error',
 			(_event) => {
 				if (fired) return; // Already handled by completion path
+				fired = true;
+				// Tear down both listeners now that the error terminal state is handled.
+				const unsub = this.sessionListeners.get(subSessionId);
+				if (unsub) {
+					unsub();
+					this.sessionListeners.delete(subSessionId);
+				}
 				const memberId = this.subSessionMemberIds.get(subSessionId);
 				if (!memberId) return;
 				try {
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -60,7 +60,11 @@ import type { SpaceRuntimeService } from './space-runtime-service';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
 import type { SpaceSessionGroupRepository } from '../../../storage/repositories/space-session-group-repository';
-import type { SubSessionFactory, SubSessionState } from '../tools/task-agent-tools';
+import type {
+	SubSessionFactory,
+	SubSessionMemberInfo,
+	SubSessionState,
+} from '../tools/task-agent-tools';
 import { createTaskAgentMcpServer } from '../tools/task-agent-tools';
 import { createTaskAgentInit, buildTaskAgentInitialMessage } from '../agents/task-agent';
 import { Logger } from '../../logger';
@@ -137,6 +141,12 @@ export class TaskAgentManager {
 	 * Allows fast lookup when adding sub-session members to an existing group.
 	 */
 	private taskGroupIds = new Map<string, string>();
+
+	/**
+	 * Maps sub-session ID → SpaceSessionGroupMember.id for sessions that were
+	 * successfully added to a group. Used to update member status on completion.
+	 */
+	private subSessionMemberIds = new Map<string, string>();
 
 	/**
 	 * Completion callbacks registered via onComplete().
@@ -581,11 +591,12 @@ export class TaskAgentManager {
 			this.taskAgentSessions.delete(taskId);
 		}
 
-		// 3. Remove any dangling completion callbacks for known session IDs.
+		// 3. Remove any dangling completion callbacks and member ID entries for known session IDs.
 		// We use the exact session IDs collected above (sub-sessions + task agent)
 		// rather than a substring match to avoid false positives.
 		for (const sessionId of sessionIdsToClean) {
 			this.completionCallbacks.delete(sessionId);
+			this.subSessionMemberIds.delete(sessionId);
 		}
 
 		// 4. Remove session listeners for known session IDs
@@ -621,8 +632,37 @@ export class TaskAgentManager {
 	 */
 	private createSubSessionFactory(taskId: string): SubSessionFactory {
 		return {
-			create: async (init: AgentSessionInit): Promise<string> => {
-				return this.createSubSession(taskId, init.sessionId, init);
+			create: async (
+				init: AgentSessionInit,
+				memberInfo?: SubSessionMemberInfo
+			): Promise<string> => {
+				const sessionId = await this.createSubSession(taskId, init.sessionId, init);
+
+				// Add the sub-session as a member of the task's group (non-fatal).
+				const groupId = this.taskGroupIds.get(taskId);
+				if (groupId) {
+					try {
+						const group = this.config.sessionGroupRepo.getGroup(groupId);
+						const orderIndex = group?.members.length ?? 0;
+						const member = this.config.sessionGroupRepo.addMember(groupId, sessionId, {
+							role: memberInfo?.role ?? 'agent',
+							agentId: memberInfo?.agentId,
+							status: 'active',
+							orderIndex,
+						});
+						this.subSessionMemberIds.set(sessionId, member.id);
+						log.info(
+							`TaskAgentManager: added sub-session ${sessionId} as member ${member.id} to group ${groupId}`
+						);
+					} catch (err) {
+						log.warn(
+							`TaskAgentManager: failed to add sub-session ${sessionId} to group ${groupId}:`,
+							err
+						);
+					}
+				}
+
+				return sessionId;
 			},
 
 			getProcessingState: (subSessionId: string): SubSessionState | null => {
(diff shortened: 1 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR correctly implements adding sub-session members to session groups in TaskAgentManager. The implementation is well-structured, with proper error handling (non-fatal try/catch around group operations), clean lifecycle management (active → completed/failed status transitions), and good test coverage.

### What looks good

1. **Non-fatal group membership** -- Group add/update operations are wrapped in try/catch so failures don't break sub-session creation or completion. This is the right choice for an observability/tracking feature.

2. **Dual event subscription** -- The `session.updated` (completion) and `session.error` (failure) listeners are correctly combined into a single unsubscribe function stored in `sessionListeners`, ensuring proper teardown.

3. **`fired` guard on error handler** -- The error handler correctly checks `if (fired) return` to avoid updating status after the completion path has already handled the session. This prevents a race where both events fire.

4. **Cleanup path** -- `subSessionMemberIds` entries are cleaned up in the `cleanup()` method alongside `completionCallbacks`.

5. **SubSessionMemberInfo flow** -- The `spawn_step_agent` tool handler correctly resolves agent identity before passing it through the factory, and the comment explains why the non-null assertion on `effectiveTask.customAgentId!` (line 305 of `task-agent-tools.ts`) is safe (guarded by the preceding `resolveAgentInit` which throws if `customAgentId` is missing).

6. **Tests** -- 6 new tests in `task-agent-manager.test.ts` and 1 in `task-agent-tools.test.ts` cover the key scenarios: member creation with correct metadata, multiple sub-sessions in one group, completed/failed status transitions, and the no-group fallback. All tests pass.

### Minor observations (P3)

1. **No dedicated test for cleanup removing `subSessionMemberIds` entries** (`task-agent-manager.ts` line 588). The cleanup behavior is covered implicitly by the existing cleanup tests (which verify session cleanup), but a targeted assertion confirming the map entry is removed would strengthen coverage. Not blocking since the cleanup code is straightforward (a Map.delete call alongside the already-tested completionCallbacks.delete).

2. **`orderIndex` race condition (theoretical)** -- The `orderIndex` is computed as `group.members.length` at the time of the `getGroup()` call (line 646 of `task-agent-manager.ts`). If two sub-sessions were created concurrently for the same task, they could get the same `orderIndex`. In practice this is unlikely since `spawn_step_agent` calls are sequential within a single MCP tool handler, and `addMember` is idempotent. Not blocking.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one behavioral correctness issue and one structural concern.

---

### P1 — `session.error` handler does not set `fired = true` or self-unsubscribe

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 763–779

The `session.error` subscriber calls `updateMemberStatus(memberId, 'failed')` but never sets `fired = true`. This creates two problems:

**Problem A — dual-fire on error then idle.** If `session.error` fires and the session somehow transitions to idle afterwards (e.g., an error is emitted mid-run but the session keeps going), `fired` is still `false` so the `session.updated` handler will also fire. That path calls `handleSubSessionComplete` which calls `updateMemberStatus(memberId, 'completed')`, overwriting `'failed'`. The member ends up as `'completed'` despite the fatal error. The `fired` flag was introduced precisely to prevent dual-firing, but the error path doesn't participate in it.

**Problem B — no self-unsubscription after error fires.** If the daemon emits multiple `session.error` events for the same session, `updateMemberStatus` is called on every event. This is harmless in practice because the UPDATE is idempotent, but the listener continues firing unnecessarily.

The fix is to set `fired = true` (and call `unsubscribeError()`) inside the error handler, mirroring the pattern used by the `session.updated` handler:

```ts
const unsubscribeError = this.config.daemonHub.on(
    'session.error',
    (_event) => {
        if (fired) return;
        fired = true;                          // <-- added
        const memberId = this.subSessionMemberIds.get(subSessionId);
        if (!memberId) return;
        try {
            this.config.sessionGroupRepo.updateMemberStatus(memberId, 'failed');
        } catch (err) { ... }
        // Tear down both listeners now that we've handled the error terminal state
        const unsub = this.sessionListeners.get(subSessionId);
        if (unsub) {
            unsub();
            this.sessionListeners.delete(subSessionId);
        }
    },
    { sessionId: subSessionId }
);
```

There is also no test verifying that after `session.error` fires (setting `fired = true`), a subsequent `session.updated → idle` event does **not** call `handleSubSessionComplete`. That inverse scenario should be added alongside the existing "failed on session.error" test.

---

### P2 — TOCTOU race on `orderIndex` computation (two separate DB reads)

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 645–652

```ts
const group = this.config.sessionGroupRepo.getGroup(groupId);
const orderIndex = group?.members.length ?? 0;
const member = this.config.sessionGroupRepo.addMember(groupId, sessionId, {
    ...
    orderIndex,
});
```

`getGroup()` fetches all member rows then returns `members.length`. `addMember()` is a separate DB operation. Between these two calls another sub-session for the same task could be inserted, causing two members to share the same `orderIndex`. While this scenario is unlikely given the current sequential spawn path, the two-read pattern is fragile and unnecessary. A cheaper atomic approach is a `SELECT COUNT(*) FROM space_session_group_members WHERE group_id = ?` inside the same transaction as the insert — or simply moving the count into a `nextOrderIndex(groupId)` helper in the repository. As written, `getGroup()` is also wasteful because it fetches all member columns just to get the count.

---

### P3 — `customAgentId!` non-null assertion comment is slightly misleading

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, line 305

```ts
const agentForMember = agentManager.getById(effectiveTask.customAgentId!);
```

The comment says "resolveAgentInit() already validated the agent exists, so this lookup should always succeed". That is correct, but the non-null assertion is actually guaranteed by `resolveStepAgents()` throwing when neither `agentId` nor `agents` is defined — the comment would be clearer by referencing that invariant. This is a documentation nit only; the code is safe.

---

### What is done well

- Correct non-fatal wrapping around all group-repo calls — sub-session creation is not blocked by membership failure.
- Clean separation of concerns: `SubSessionMemberInfo` is defined in `task-agent-tools.ts` (the layer that knows the agent identity) and only consumed by `task-agent-manager.ts` (the layer that manages groups).
- The combined unsubscribe pattern (`() => { unsubscribeUpdated(); unsubscribeError(); }`) is a clean solution for coordinated cleanup.
- 6 well-structured tests covering the happy path, multi-member ordering, status transitions, and the non-fatal path. The `_capturedMemberInfos` extension on `makeMockSessionFactory` is a clean testing pattern.
- `handleSubSessionComplete` correctly updates member status before notifying the Task Agent.
- `cleanup()` correctly removes `subSessionMemberIds` entries alongside `completionCallbacks`.

lsm (commented):

**Recommendation: APPROVE**

### P1 Fix: `session.error` handler race condition — Verified

The fix at `task-agent-manager.ts:760-793` correctly addresses the race condition where a `session.error` event followed by a `session.updated → idle` event could overwrite `'failed'` status with `'completed'`.

The implementation:
1. Sets `fired = true` immediately in the error handler (line 768), preventing the idle handler from running
2. Self-unsubscribes both listeners via the combined unsub function (lines 770-774), preventing any further invocations
3. Updates the member status to `'failed'` via `updateMemberStatus` (line 778)

The combined unsubscribe function stored at lines 790-793 correctly captures both `unsubscribeUpdated` and `unsubscribeError` closures, ensuring both listeners are torn down regardless of which fires first. The `session.updated` handler has a symmetric pattern (lines 736-742) that also sets `fired = true` and tears down both listeners.

The inverse test at line 1188 (`idle event after session.error does not overwrite failed status with completed`) properly verifies:
- Error event fires first -> member status becomes `'failed'`
- Subsequent idle event -> completion callback does NOT fire
- Member status remains `'failed'` (not overwritten)

### P2 Fix: TOCTOU on `getGroup().members.length` — Verified

The `getMemberCount` method at `space-session-group-repository.ts:339-344` correctly replaces the two-read pattern with a single `SELECT COUNT(*)` query. This is both cheaper (no row materialization) and avoids returning stale data from an in-memory array.

Usage at `task-agent-manager.ts:645`:
```
const orderIndex = this.config.sessionGroupRepo.getMemberCount(groupId);
```

The remaining TOCTOU between `getMemberCount` and `addMember` is acceptable since `orderIndex` is a display hint, not a uniqueness constraint. Concurrent sub-session creation producing duplicate indices would only affect visual ordering, which is benign.

### Other observations

The overall implementation is well-structured. A few notes on quality:

- Cleanup at lines 597-600 correctly removes `subSessionMemberIds` entries alongside `completionCallbacks`
- The non-fatal pattern for `addMember` failure in the factory (lines 656-661) ensures sub-session creation succeeds even if group membership tracking fails
- Test coverage is thorough: 5 new tests for group membership (add, multiple with incremental orderIndex, completed status, failed status, error-then-idle guard) plus 1 test in task-agent-tools for memberInfo passthrough

No issues found. Both fixes are correct and complete.

lsm (commented):

---

**Recommendation: APPROVE**

Both previously-raised concerns are correctly addressed. The implementation is clean and the new tests are thorough.

---

### P1 Fix — session.error handler self-unsubscribes and sets fired

Confirmed correct in `task-agent-manager.ts` lines 764–793.

The error handler (line 766–787):
1. Guards on `if (fired) return` immediately
2. Sets `fired = true` before doing any work
3. Calls `this.sessionListeners.get(subSessionId)` to get the **combined** unsubscribe, which tears down both `unsubscribeUpdated` and `unsubscribeError` in one call

The combined unsubscribe is stored at line 790, **after** both `daemonHub.on()` calls complete. This is safe because `TypedHub.on()` is synchronous (delegates to `subscribe()` which only writes to maps, with actual dispatch deferred via `queueMicrotask`). No event can fire during the registration window.

The new inverse test at line 1188 (`idle event after session.error does not overwrite failed status with completed`) precisely verifies that once the error handler fires, a subsequent `session.updated → idle` cannot change status from `failed` to `completed` and cannot call the completion callback. This is exactly the right test for the guard.

---

### P2 Fix — TOCTOU race in orderIndex computation

Confirmed correct. `getMemberCount()` in `space-session-group-repository.ts` (lines 339–344) issues a `SELECT COUNT(*) as cnt` directly against the DB, which returns the count at the moment of the query rather than loading all member rows first. This eliminates the original TOCTOU window where `getGroup().members.length` could be stale relative to a concurrent insert.

The usage at `task-agent-manager.ts` line 645:

```ts
const orderIndex = this.config.sessionGroupRepo.getMemberCount(groupId);
```

is clean and minimal. The test at line 1113 verifies that two sequential sub-sessions for the same task receive strictly increasing orderIndex values.

---

**One minor note on the P1 fix:** After the error handler calls the combined unsubscribe and deletes `sessionListeners.get(subSessionId)`, it does not explicitly clear `completionCallbacks` for that session. This is not a correctness problem — the `fired = true` flag prevents the idle handler from ever reaching the callback list — but the `completionCallbacks` entry will remain in the map until `cleanup()` runs. This is the same behaviour that existed before this PR and is consistent with the existing cleanup path in `cleanup()` at line 598, so it is not a new issue.

---

All 5 new group-membership tests in `task-agent-manager.test.ts` and the spawn_step_agent memberInfo test in `task-agent-tools.test.ts` directly cover the changed paths. The implementation matches the PR description. No new issues found.

PR #639, opened 2026-03-22
feat: rehydrate group map on daemon restart via dedicated rehydrateGroupMaps()

lsm:
Add SpaceSessionGroupRepository.listActiveGroupsWithTaskId() for an efficient
single indexed query of all active groups with a non-null task_id.

Add TaskAgentManager.rehydrateGroupMaps() which calls the new repo method and
rebuilds the taskId -> groupId in-memory map in one pass before individual Task
Agent sessions are restored. Groups without a taskId (standalone groups) are
silently skipped; groups with no active members are still included.

Call rehydrateGroupMaps() at the start of rehydrate() so the full map is
available for all active groups, including those whose Task Agent session
rehydration fails. Existing map entries (set by a concurrent spawnTaskAgent())
are not overwritten.

Add 6 unit tests covering: multi-group rebuild, standalone groups skipped,
memberless groups included, completed groups excluded, no-overwrite of
pre-populated entries, and map populated even when no sessions need rehydration.

### DIFF
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -818,7 +818,7 @@ export class TaskAgentManager {
 	 * Rebuild the `taskId → groupId` in-memory map from all active groups in the DB.
 	 *
 	 * Queries `space_session_groups` for every row with `status = 'active'` and a
-	 * non-null `task_id` in a single indexed query. Groups without a `task_id`
+	 * non-null `task_id`. Groups without a `task_id`
 	 * (standalone groups) are silently skipped — there is no taskId key to map.
 	 * Groups with no active members are still valid and are included.
 	 *
@@ -986,12 +986,6 @@ export class TaskAgentManager {
 				}
 				this.subSessions.get(taskId)!.set(subSessionId, subSession);
 			}
-		}
-
-		// --- Restore taskGroupIds from DB so getTaskGroupId() works after restart
-		const existingGroups = this.config.sessionGroupRepo.getGroupsByTask(spaceId, taskId);
-		if (existingGroups.length > 0) {
-			this.taskGroupIds.set(taskId, existingGroups[0].id);
 		}
 
 		log.info(
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -505,6 +505,12 @@ export class TaskAgentManager {
 	 * WorkflowExecutors are loaded, so executors are ready when Task Agents run.
 	 */
 	async rehydrate(): Promise<void> {
+		// Rebuild the taskId → groupId map from all active groups before rehydrating
+		// individual Task Agent sessions. This ensures the map is fully populated even
+		// for tasks whose session rehydration fails, and covers all active groups in one
+		// indexed query rather than N per-task queries.
+		this.rehydrateGroupMaps();
+
 		const activeTasks = this.config.taskRepo.listActiveWithTaskAgentSession();
 
 		let attempted = 0;
@@ -805,8 +811,39 @@ export class TaskAgentManager {
 	}
 
 	// -------------------------------------------------------------------------
-	// Private — rehydration helper
+	// Private — rehydration helpers
 	// -------------------------------------------------------------------------
+
+	/**
+	 * Rebuild the `taskId → groupId` in-memory map from all active groups in the DB.
+	 *
+	 * Queries `space_session_groups` for every row with `status = 'active'` and a
+	 * non-null `task_id` in a single indexed query. Groups without a `task_id`
+	 * (standalone groups) are silently skipped — there is no taskId key to map.
+	 * Groups with no active members are still valid and are included.
+	 *
+	 * Existing map entries are NOT overwritten: if a fresh `spawnTaskAgent()` call
+	 * already populated `taskGroupIds` before `rehydrate()` runs, that value takes
+	 * precedence over the DB row (the in-memory value is always the most recent).
+	 *
+	 * This method is synchronous and performs a single DB read. It is called at the
+	 * start of `rehydrate()` so the full map is available before individual Task Agent
+	 * sessions are restored.
+	 */
+	private rehydrateGroupMaps(): void {
+		const activeGroups = this.config.sessionGroupRepo.listActiveGroupsWithTaskId();
+		let restored = 0;
+		for (const { id, taskId } of activeGroups) {
+			// Don't overwrite if already populated by a concurrent spawnTaskAgent()
+			if (!this.taskGroupIds.has(taskId)) {
+				this.taskGroupIds.set(taskId, id);
+				restored++;
+			}
+		}
+		log.info(
+			`TaskAgentManager.rehydrateGroupMaps: restored ${restored}/${activeGroups.length} active group mappings`
+		);
+	}
 
 	/**
 	 * Rehydrate a single Task Agent session after daemon restart.
--- a/packages/daemon/src/storage/repositories/space-session-group-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-session-group-repository.ts
@@ -123,8 +123,8 @@ export class SpaceSessionGroupRepository {
 
 	/**
 	 * List all active groups that have a non-null task_id, across all spaces.
-	 * Returns lightweight objects (id + taskId only) for efficient startup rehydration.
-	 * The result is indexed on task_id — no full member fetch is performed.
+	 * Returns lightweight objects (id + taskId only) — no full member fetch is performed.
+	 * This is a startup-once cold-path query; table sizes are typically small.
 	 */
 	listActiveGroupsWithTaskId(): Array<{ id: string; taskId: string }> {
 		const stmt = this.db.prepare(
--- a/packages/daemon/src/storage/repositories/space-session-group-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-session-group-repository.ts
@@ -122,6 +122,19 @@ export class SpaceSessionGroupRepository {
 	}
 
 	/**
+	 * List all active groups that have a non-null task_id, across all spaces.
+	 * Returns lightweight objects (id + taskId only) for efficient startup rehydration.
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Summary

Clean, well-scoped PR that adds startup rehydration of the `taskGroupIds` map via a dedicated `rehydrateGroupMaps()` method. The implementation is correct, efficient (single indexed DB query), and well-tested with 6 unit tests covering the key scenarios.

### Findings

**P2: Pre-existing unconditional overwrite in `rehydrateTaskAgent` (line 991-994)**

Not introduced by this PR, but worth noting: `rehydrateTaskAgent()` at line 991-994 unconditionally calls `this.taskGroupIds.set(taskId, existingGroups[0].id)` using `getGroupsByTask()` which has NO status filter. This means it could overwrite the correctly-populated active group ID (from `rehydrateGroupMaps`) with a completed/failed group ID if the completed group was created earlier (since both queries order by `created_at ASC`).

In practice, the risk is low — tasks rarely have multiple groups — but it undermines the careful "don't overwrite" logic in `rehydrateGroupMaps()`. Consider either (a) adding a `!this.taskGroupIds.has(taskId)` guard at line 991-994, or (b) removing that per-task fallback entirely since `rehydrateGroupMaps()` now handles it. This is a pre-existing issue so I am not blocking on it.

**P3: Minor — `listActiveGroupsWithTaskId` query could benefit from composite index**

The query filters on `status = 'active' AND task_id IS NOT NULL`. There is an index on `task_id` alone but no composite `(status, task_id)` index. For a one-time startup query this is negligible, just noting for completeness.

### What looks good

- The `rehydrateGroupMaps()` placement at the start of `rehydrate()` (before the per-task loop) is the right call — it ensures the map is available even when individual task rehydration fails.
- The "don't overwrite" guard (`!this.taskGroupIds.has(taskId)`) correctly handles the race with concurrent `spawnTaskAgent()` calls.
- The lightweight return type (`{id, taskId}` only, no member fetch) is appropriate for a startup path.
- Tests cover the important edge cases: multiple groups, standalone groups, empty groups, completed groups, no-overwrite semantics, and the "no sessions to rehydrate but groups exist" scenario.

**Recommendation: APPROVE** — only P3 nits. The P2 is pre-existing and out of scope for this PR.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 issue (dead code that may cause a silent wrong-group selection) and one P2 concern about index coverage.

---

### P1 — Dead/conflicting group-restore code in `rehydrateTaskAgent` (lines 991–995)

After this PR, `rehydrateGroupMaps()` is called at the start of `rehydrate()`, which pre-populates `taskGroupIds` for every active group. However `rehydrateTaskAgent()` still contains its own redundant fallback at the end of the method:

```ts
// packages/daemon/src/lib/space/runtime/task-agent-manager.ts  lines 991–995
// --- Restore taskGroupIds from DB so getTaskGroupId() works after restart
const existingGroups = this.config.sessionGroupRepo.getGroupsByTask(spaceId, taskId);
if (existingGroups.length > 0) {
    this.taskGroupIds.set(taskId, existingGroups[0].id);
}
```

This code has two problems:

1. **Silent override of the no-overwrite invariant.** `rehydrateGroupMaps()` deliberately skips entries already in the map (the "no-overwrite" invariant documented in both the JSDoc and tested in `'existing map entries are not overwritten by rehydrateGroupMaps'`). This `set()` call bypasses that guard and can unconditionally overwrite whatever value `rehydrateGroupMaps()` correctly placed. In the scenario where there are two active groups for the same `taskId` (e.g., one from the fresh run, one stale), `rehydrateGroupMaps()` first-writer-wins (oldest by `created_at`), but this fallback ignores that ordering and picks `existingGroups[0].id` from `getGroupsByTask`, which returns by `created_at ASC` — so they will agree in the same normal case. But the inconsistency in invariant is a maintenance hazard.

2. **It is now completely dead code** — `rehydrateGroupMaps()` already handles this case, including for tasks whose `rehydrateTaskAgent()` call succeeds. The per-task fallback was the original pre-PR mechanism, and should have been removed as part of this PR. Leaving it creates confusion about which code path is responsible for restoring the map, and the test `'taskGroupIds is restored from DB after rehydration'` (line 1746) already covers the correct path via `rehydrateGroupMaps()`.

**Fix**: Remove lines 991–995 from `rehydrateTaskAgent`.

---

### P2 — Index coverage: `status = 'active' AND task_id IS NOT NULL` will not use `idx_space_session_groups_task_id`

The `listActiveGroupsWithTaskId` query is:

```sql
SELECT id, task_id FROM space_session_groups
WHERE status = 'active' AND task_id IS NOT NULL
ORDER BY created_at ASC
```

The only index on this table that could help is `idx_space_session_groups_task_id ON space_session_groups(task_id)` (added in migration 40). However, because the leading predicate is `status = 'active'` (an equality on an un-indexed column), SQLite will not use the `task_id` index for this query — it will perform a full table scan. The comment in both the JSDoc ("The result is indexed on task_id — no full member fetch is performed") and the PR description ("single indexed scan") is therefore inaccurate for the common access pattern at startup.

This is only a correctness concern for the documentation/comment — the query is still functionally correct, and for typical table sizes this is a startup-once query so performance impact is low. But the misleading comment could cause future engineers to rely on index selectivity that does not exist.

**Fix**: Either add a composite index `(status, task_id)` to support the query as described, or correct the JSDoc to say "no full member-record fetch is performed" (i.e., clarify that it fetches lightweight rows, not that the scan is index-driven). Given this is a startup-once cold-path, the composite index is optional; correcting the comment is sufficient.

---

### P3 — Test `'existing map entries are not overwritten by rehydrateGroupMaps'` does not catch the P1 bug

The test creates one group via `spawnTaskAgent` (which sets the map), then creates a second group for the same `taskId`, then asserts the map still holds the spawn value after `rehydrate()`. This correctly validates the `rehydrateGroupMaps()` no-overwrite logic. However, because `rehydrateTaskAgent()` is also called for the seeded task in this test (the session exists in the mock DB with `type: 'space_task_agent'`), the per-task fallback at line 994 _also_ runs and calls `this.taskGroupIds.set(taskId, existingGroups[0].id)`. Since `getGroupsByTask` returns groups ordered by `created_at ASC`, and the `spawnTaskAgent` group was created first, `existingGroups[0]` happens to be the correct group — so the test passes even though the invariant is violated. A targeted test that creates the stale group _before_ the fresh group (so it appears as `existingGroups[0]`) would catch the bug.

---

### Positive notes

- The `listActiveGroupsWithTaskId()` method is clean and correctly returns lightweight `{id, taskId}` projections, avoiding the N-query pattern of the old per-task `getGroupsByTask()` calls.
- The no-overwrite guard in `rehydrateGroupMaps()` is the right invariant and is correctly implemented.
- The 6 new tests are well-structured and cover the important behavioral cases (multi-group rebuild, standalone groups, memberless groups, completed groups excluded, no-overwrite, map populated without session rehydration).
- Calling `rehydrateGroupMaps()` before the per-task loop so that groups whose sessions fail to rehydrate are still in the map is the correct design.

lsm (commented):

**Recommendation: APPROVE** — Both prior issues have been cleanly addressed.

---

### Follow-up verification

**P1 fix confirmed: dead/conflicting `getGroupsByTask` code removed from `rehydrateTaskAgent`**

The six-line block that unconditionally called `this.config.sessionGroupRepo.getGroupsByTask(spaceId, taskId)` and overwrote `taskGroupIds` without a guard has been fully removed from `rehydrateTaskAgent()`. A search across the file confirms no remaining `getGroupsByTask` calls exist in that method. The no-overwrite invariant documented in `rehydrateGroupMaps()` is now the sole code path responsible for restoring the map on restart, with no competing writes.

**P2 fix confirmed: misleading "indexed scan" language removed from JSDoc**

The `listActiveGroupsWithTaskId()` JSDoc now correctly says "Returns lightweight objects (id + taskId only) — no full member fetch is performed" and "This is a startup-once cold-path query; table sizes are typically small." Neither the method comment nor the `rehydrateGroupMaps()` JSDoc contain the earlier inaccurate "indexed scan" or "single indexed query" phrasing. The PR description still uses "single indexed scan" but that is documentation prose, not executable code, and does not affect correctness.

---

### What remains good

- `rehydrateGroupMaps()` is called unconditionally at the start of `rehydrate()`, before the per-task loop. Groups whose individual session rehydration later fails are still captured in the map.
- The no-overwrite guard (`!this.taskGroupIds.has(taskId)`) is the only write path for the map during rehydration, correctly handling any concurrent `spawnTaskAgent()` call that pre-populated an entry.
- The six unit tests cover multi-group rebuild, standalone groups, memberless groups, completed groups excluded, no-overwrite semantics, and map populated without any session rehydration needed.
- The `listActiveGroupsWithTaskId()` lightweight projection avoids the N×`getGroupMembers` fan-out of the old per-task fallback.

No new issues introduced in this iteration.

lsm (commented):

### Re-review after feedback round

Both issues raised in the previous round have been addressed in commit `4d036a8d`:

1. **P1 (dead code)** -- The per-task `getGroupsByTask` fallback in `rehydrateTaskAgent()` (old lines 991-995) has been removed. The `rehydrateGroupMaps()` method is now the single source of truth for restoring `taskGroupIds` on startup, and the no-overwrite invariant is no longer violated by a downstream unconditional `set()` call.

2. **P2 (inaccurate JSDoc)** -- The misleading "indexed scan" language has been removed from both `listActiveGroupsWithTaskId()` and `rehydrateGroupMaps()`. The new JSDoc in `listActiveGroupsWithTaskId()` correctly describes this as a "startup-once cold-path query" with lightweight projections, without claiming index coverage that does not exist.

Both fixes are minimal and clean -- no unrelated changes were introduced.

### Verification

- `rehydrateGroupMaps()` remains the sole mechanism for populating `taskGroupIds` during rehydration (confirmed via grep: no other `taskGroupIds.set` calls exist in the rehydration path).
- The existing test suite (6 tests in the `rehydrateGroupMaps (via rehydrate)` describe block) still covers all the important scenarios. The P3 note from the previous round about test ordering sensitivity is no longer relevant since the conflicting code path has been removed entirely.
- The `getGroupsByTask` method still exists on the repository (used in other test files and potentially by other callers), so no downstream breakage.

**Recommendation: APPROVE**

PR #640, opened 2026-03-22
feat: wire session title handler into SessionManager and remove pendingBackgroundTasks

lsm:
- Add jobQueue and jobProcessor dependencies to SessionManager constructor
- Add start() method that registers session.title_generation handler
- Replace fire-and-forget pendingBackgroundTasks with jobQueue.enqueue()
- Remove pendingBackgroundTasks field and cleanup drain logic
- Call sessionManager.start() in app.ts before jobProcessor.start()
- Move handler registration from rpc-handlers/index.ts to SessionManager.start()
- Update and expand unit tests to cover new enqueue behavior

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -188,8 +188,13 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 			maxTokens: config.maxTokens,
 			temperature: config.temperature,
 			workspaceRoot: config.workspaceRoot,
-		}
+		},
+		jobQueue,
+		jobProcessor
 	);
+
+	// Register session title generation handler before jobProcessor starts
+	sessionManager.start();
 
 	// Initialize State Manager (listens to EventBus, clean dependency graph!)
 	const stateManager = new StateManager(
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -62,8 +62,7 @@ import { SpaceAgentRepository } from '../../storage/repositories/space-agent-rep
 import type { JobQueueRepository } from '../../storage/repositories/job-queue-repository';
 import type { JobQueueProcessor } from '../../storage/job-queue-processor';
 import { SpaceSessionGroupRepository } from '../../storage/repositories/space-session-group-repository';
-import { SESSION_TITLE_GENERATION, GITHUB_POLL, ROOM_TICK } from '../job-queue-constants';
-import { handleSessionTitleGeneration } from '../job-handlers/session-title.handler';
+import { GITHUB_POLL, ROOM_TICK } from '../job-queue-constants';
 import { handleGitHubPoll } from '../job-handlers/github-poll.handler';
 import { createRoomTickHandler, enqueueRoomTick } from '../job-handlers/room-tick.handler';
 import { SpaceRuntimeService } from '../space/runtime/space-runtime-service';
@@ -151,11 +150,6 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	setupConfigHandlers(deps.messageHub, deps.sessionManager, deps.daemonHub);
 	setupTestHandlers(deps.messageHub, deps.db);
 	setupRewindHandlers(deps.messageHub, deps.sessionManager, deps.daemonHub);
-
-	// Job queue handler registrations
-	deps.jobProcessor.register(SESSION_TITLE_GENERATION, (job) =>
-		handleSessionTitleGeneration(job, deps.sessionManager.getSessionLifecycle())
-	);
 
 	// Room handlers
 	setupRoomHandlers(
--- a/packages/daemon/src/lib/session/session-manager.ts
+++ b/packages/daemon/src/lib/session/session-manager.ts
@@ -38,12 +38,11 @@ import { MessagePersistence } from './message-persistence';
 /**
  * Cleanup state machine for SessionManager
  *
- * Prevents race conditions during cleanup by tracking state and
- * preventing new background tasks from starting during cleanup.
+ * Prevents concurrent or redundant cleanup calls.
  *
  * States:
  * - IDLE: Normal operation, cleanup not started
- * - CLEANING: Cleanup in progress, barrier active for new background tasks
+ * - CLEANING: Cleanup in progress
  * - CLEANED: Cleanup complete, no further operations allowed
  */
 export enum CleanupState {
@@ -56,6 +55,7 @@ export class SessionManager {
 	private logger: Logger;
 	private worktreeManager: WorktreeManager;
 	private eventBusUnsubscribers: Array<() => void> = [];
+	private started = false;
 
 	// Cleanup state machine - prevents race conditions during shutdown
 	private cleanupState: CleanupState = CleanupState.IDLE;
@@ -116,8 +116,13 @@ export class SessionManager {
 	/**
 	 * Register job handlers and start background processing.
 	 * Must be called after construction but before jobProcessor.start().
+	 * Throws if called more than once to catch accidental double-registration.
 	 */
 	start(): void {
+		if (this.started) {
+			throw new Error('SessionManager.start() called more than once');
+		}
+		this.started = true;
 		this.jobProcessor.register(SESSION_TITLE_GENERATION, (job) =>
 			handleSessionTitleGeneration(job, this.sessionLifecycle)
 		);
--- a/packages/daemon/src/lib/session/session-manager.ts
+++ b/packages/daemon/src/lib/session/session-manager.ts
@@ -9,7 +9,6 @@
  *
  * Also manages:
  * - EventBus subscriptions for async message processing
- * - Background task tracking for cleanup
  */
 
 import type { Session, MessageHub, MessageDeliveryMode } from '@neokai/shared';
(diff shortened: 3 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly replaces the fire-and-forget `pendingBackgroundTasks` pattern with persistent job queue enqueueing for session title generation. The implementation is correct, well-structured, and the tests adequately cover the new behavior.

### What was done

1. `SessionManager` now accepts `JobQueueRepository` and `JobQueueProcessor` as constructor dependencies
2. A new `start()` method registers the `session.title_generation` handler on the job processor
3. The `message.persisted` event handler now calls `this.jobQueue.enqueue()` instead of directly calling `generateTitleAndRenameBranch` with promise tracking
4. The `pendingBackgroundTasks` field and all associated drain/barrier logic in `cleanup()` are removed
5. Handler registration moved from `rpc-handlers/index.ts` to `SessionManager.start()`
6. Shutdown order in `app.ts` is correct: `jobProcessor.stop()` runs before `sessionManager.cleanup()`

### Verification

- No remaining references to `pendingBackgroundTasks` anywhere in the codebase
- `SESSION_TITLE_GENERATION` references are properly consolidated
- The `start()` call in `app.ts` (line 196) happens before `jobProcessor.start()` (line 389), ensuring the handler is registered before the processor begins polling -- correct initialization order
- The cleanup barrier (`cleanupState !== IDLE` check before enqueueing) was properly removed since the job queue is persistent and doesn't need the same race condition protection as in-memory promises

### Minor observations (P3)

1. **`session-manager.ts` line 42-47**: The `CleanupState` doc comment still mentions "preventing new background tasks from starting during cleanup" but that concept no longer applies. The code comment at line 329-330 was updated but the enum docstring was not. Very minor.

2. **Test at line 174-185**: The test "should not enqueue a job before start is called" tests with `needsWorkspaceInit: false`, which means `enqueue` would never be called regardless of whether `start()` was called. The test name implies it's testing a dependency on `start()`, but the assertion passes because `needsWorkspaceInit` is false. This doesn't test what it claims to test. However, enqueue behavior is independent of `start()` (start only registers the handler), so there is no actual bug here -- just a slightly misleading test name.

Both are cosmetic/documentation issues only.

lsm (commented):

---

Recommendation: **REQUEST_CHANGES**

The implementation correctly wires the job handler into `SessionManager`, removes `pendingBackgroundTasks`, and passes `jobQueue`/`jobProcessor` through the constructor. The ordering in `app.ts` (`sessionManager.start()` before `jobProcessor.start()`) is correct and matches the stated goal. However, there are a few issues worth addressing.

---

### P1 — Misleading test does not cover the intent

**File:** `packages/daemon/tests/unit/session/session-manager.test.ts`, lines 174–185

The test "should not enqueue a job before start is called" passes `needsWorkspaceInit: false`, so `jobQueue.enqueue` is never called regardless of whether `start()` was invoked. The assertion is vacuously true and the test name is wrong.

The purpose of `start()` is to register the `session.title_generation` handler with `jobProcessor`. Nothing in `start()` prevents `jobQueue.enqueue()` from being called in `message.persisted` — that enqueue is always unconditional when `needsWorkspaceInit: true`. The test should instead verify that calling the handler with `needsWorkspaceInit: true` *does* enqueue a job (already covered by a later test), and separately verify that the handler registration call count on `mockJobProcessor.register` is 0 before `start()` is called and 1 after. As written, the test gives false confidence.

---

### P2 — Stale JSDoc on `CleanupState` enum

**File:** `packages/daemon/src/lib/session/session-manager.ts`, lines 42–47

The block comment above `CleanupState` still says:
> "Prevents race conditions during cleanup by tracking state and preventing new background tasks from starting during cleanup."
> "CLEANING: Cleanup in progress, **barrier active for new background tasks**"

The barrier no longer exists — the old `if (this.cleanupState !== CleanupState.IDLE) return;` guard inside the `message.persisted` handler was removed as part of this PR. The comment should be updated to reflect the new role of `CleanupState`, which is now only about preventing concurrent/redundant cleanup calls, not about blocking enqueue operations.

---

### P2 — `start()` silently allows double-registration

**File:** `packages/daemon/src/lib/session/session-manager.ts`, lines 120–124

`JobQueueProcessor.register()` uses `Map.set()`, which silently overwrites any previously registered handler for the same queue. If `sessionManager.start()` is called more than once (e.g., in a test that re-creates the object without realizing `start()` was already called, or if app startup wiring changes), the second call silently replaces the handler. Consider guarding with a boolean `started` flag and throwing (or at least logging a warning) on duplicate calls.

---

### P3 — `enqueue()` return value is unused and error is swallowed

**File:** `packages/daemon/src/lib/session/session-manager.ts`, line 155

`this.jobQueue.enqueue(...)` returns a `Job` object, which is discarded. This is intentional (fire-and-enqueue semantics), but the surrounding `try/catch` at line 168 will catch any SQLite error thrown by `enqueue()` and log it as a non-fatal error. This is acceptable since the user message is already persisted, but a brief comment noting that enqueue errors are treated as non-fatal would make the intent clearer for the next reader.

lsm (commented):

### Re-review (iteration 2)

All three feedback items from the previous review have been addressed correctly:

1. **CleanupState JSDoc** -- Updated to remove stale "barrier" references. The new docstring at lines 38-47 of `session-manager.ts` is accurate and concise.

2. **Double-call guard on `start()`** -- The `private started = false` field with a throw on re-entry (lines 122-124) is the right pattern. Clean and deterministic.

3. **Improved tests** -- The old misleading "should not enqueue before start" test is gone. The two replacement tests (lines 164-181) are meaningful: one verifies `register` is called 0 times before `start()` and exactly 1 time after; the other verifies double-calling `start()` throws.

### Overall assessment

The PR cleanly achieves its stated goal:

- **Handler registration** moved from `rpc-handlers/index.ts` into `SessionManager.start()`, co-locating the handler with its consumer (good cohesion).
- **`pendingBackgroundTasks`** fully removed (no references remain anywhere in the codebase). The fire-and-forget + drain pattern is replaced by persistent `jobQueue.enqueue()`, which is strictly better for reliability and shutdown safety.
- **Shutdown ordering** in `app.ts` is correct: `jobProcessor.stop()` (line 442) drains in-flight jobs before `sessionManager.cleanup()` (line 468) tears down sessions. The cleanup JSDoc (lines 325-336) correctly documents this responsibility split.
- **`start()` call placement** in `app.ts` (line 196) happens after construction but before `jobProcessor.start()` (line 389), ensuring the handler is registered before the processor begins polling. The comment on line 195 documents this constraint.
- The removal of the cleanup barrier check in the `message.persisted` handler is safe because cleanup PHASE 1 unsubscribes from the EventBus before any session teardown, so no new events can be processed during shutdown.

No issues found. The implementation is correct, complete, and well-tested.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

Iteration 2 has correctly addressed all three feedback items from the previous review. The implementation is sound and the tests are meaningful. One minor stale-doc nit remains.

---

### What was changed

- `SessionManager` now accepts `jobQueue: JobQueueRepository` and `jobProcessor: JobQueueProcessor` constructor parameters
- A `start()` method registers the `session.title_generation` handler on `jobProcessor`; a `private started = false` guard throws `'SessionManager.start() called more than once'` on a second call
- `setupEventSubscriptions` now calls `this.jobQueue.enqueue(...)` instead of spawning a fire-and-forget promise
- `pendingBackgroundTasks`, its drain logic, and the cleanup barrier check in the persisted-handler are all removed
- `app.ts` calls `sessionManager.start()` immediately after construction, before `setupRPCHandlers` and before `jobProcessor.start()`
- Tests cover: handler registered 0× before `start()` and 1× after; double-calling `start()` throws; `jobQueue.enqueue` called with the right payload

---

### Issues found

**P3 – Stale "Sets barrier" wording in `cleanup()` JSDoc**
`packages/daemon/src/lib/session/session-manager.ts`, line 328

The JSDoc comment for `cleanup()` still reads:

```
- IDLE → CLEANING: Sets barrier
```

The "barrier" concept was removed in this PR (the barrier was the thing that blocked new background tasks from starting during cleanup). Now that title generation is handled by the job processor, there is no barrier — the CLEANING state only prevents concurrent or redundant `cleanup()` calls. The wording should be updated to match the CleanupState enum's own JSDoc (which correctly says "Prevents concurrent or redundant cleanup calls"). For example: `- IDLE → CLEANING: Begin cleanup`.

This is purely cosmetic since the code behaviour is correct, but the stale wording contradicts the just-updated CleanupState enum comment two dozen lines above.

---

### Positive observations

- **Ordering in app.ts is correct**: `sessionManager.start()` is called before `jobProcessor.start()`, so no pending `session.title_generation` job from a previous run can be dequeued without a handler. The shutdown sequence is also correct: `jobProcessor.stop()` drains in-flight jobs before `sessionManager.cleanup()` is called, which aligns with the JSDoc claim.
- **Synchronous enqueue**: `jobQueue.enqueue()` is synchronous (returns `Job`, not `Promise<Job>`), and the call site correctly omits `await`. The surrounding `try/catch` still catches any thrown error from the enqueue call itself.
- **No double-registration**: Only one registration of `SESSION_TITLE_GENERATION` exists (in `SessionManager.start()`); the old handler in `rpc-handlers/index.ts` has been cleanly removed.
- **Tests are meaningful**: The two new `start()` tests verify the actual contract (call count before/after, throws on second call). The enqueue test verifies the exact payload shape. The old misleading test ("no enqueue before start()") is gone.

PR #643, opened 2026-03-22
feat: add notifyChange to SessionGroupRepository

lsm:
- Inject ReactiveDatabase as required constructor parameter; extract raw
  BunDatabase from it for direct SQL operations
- Wrap createGroup (multi-table write) in beginTransaction/commitTransaction
  with try/catch + abortTransaction error handling
- Convert recordGateFailure from bun-native transaction to ReactiveDatabase
  transaction for consistent notification batching
- Call notifyChange('session_groups') after every successful write to
  session_groups rows; only notify when rows actually changed (version-checked
  methods check result.changes)
- Fix ReactiveDatabase proxy to bind non-intercepted methods to original
  target, preventing private-field access errors for methods like prepare()
- Update all construction sites (room-runtime-service, task-handlers, rpc
  index) to pass reactiveDb; task-handlers falls back to createReactiveDatabase
  when called from tests without a reactiveDb
- Update all test construction sites to use createReactiveDatabase(db)
- Add session-group-repository-reactive.test.ts: LiveQuery subscription fires
  after writes; abortTransaction called on error; transactionDepth never stuck

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -374,7 +374,7 @@ export class RoomRuntimeService {
 		if (existing) return existing;
 
 		const rawDb = this.ctx.db.getDatabase();
-		const groupRepo = new SessionGroupRepository(this.ctx.reactiveDb);
+		const groupRepo = new SessionGroupRepository(rawDb, this.ctx.reactiveDb);
 		const taskManager = new TaskManager(rawDb, room.id, this.ctx.reactiveDb);
 		const goalManager = new GoalManager(rawDb, room.id, this.ctx.reactiveDb);
 		const sdkMessageRepo = new SDKMessageRepository(rawDb);
@@ -591,7 +591,7 @@ export class RoomRuntimeService {
 		observer: SessionObserver
 	): Promise<void> {
 		const rawDb = this.ctx.db.getDatabase();
-		const groupRepo = new SessionGroupRepository(this.ctx.reactiveDb);
+		const groupRepo = new SessionGroupRepository(rawDb, this.ctx.reactiveDb);
 		const taskManager = new TaskManager(rawDb, roomId, this.ctx.reactiveDb);
 		const sessionFactory = this.createSessionFactory();
 
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -374,7 +374,7 @@ export class RoomRuntimeService {
 		if (existing) return existing;
 
 		const rawDb = this.ctx.db.getDatabase();
-		const groupRepo = new SessionGroupRepository(rawDb);
+		const groupRepo = new SessionGroupRepository(this.ctx.reactiveDb);
 		const taskManager = new TaskManager(rawDb, room.id, this.ctx.reactiveDb);
 		const goalManager = new GoalManager(rawDb, room.id, this.ctx.reactiveDb);
 		const sdkMessageRepo = new SDKMessageRepository(rawDb);
@@ -591,7 +591,7 @@ export class RoomRuntimeService {
 		observer: SessionObserver
 	): Promise<void> {
 		const rawDb = this.ctx.db.getDatabase();
-		const groupRepo = new SessionGroupRepository(rawDb);
+		const groupRepo = new SessionGroupRepository(this.ctx.reactiveDb);
 		const taskManager = new TaskManager(rawDb, roomId, this.ctx.reactiveDb);
 		const sessionFactory = this.createSessionFactory();
 
--- a/packages/daemon/src/lib/room/state/session-group-repository.ts
+++ b/packages/daemon/src/lib/room/state/session-group-repository.ts
@@ -205,16 +205,10 @@ export interface TaskGroupEvent {
 }
 
 export class SessionGroupRepository {
-	/** Raw BunDatabase for direct SQL queries */
-	private db: BunDatabase;
-
-	constructor(private reactiveDb: ReactiveDatabase) {
-		// Extract the raw BunDatabase for direct SQL operations.
-		// In production, reactiveDb.db is the proxied Database facade (has getDatabase()).
-		// In tests, reactiveDb.db may be a raw BunDatabase directly.
-		const facade = reactiveDb.db as unknown as { getDatabase?: () => BunDatabase } & BunDatabase;
-		this.db = typeof facade.getDatabase === 'function' ? facade.getDatabase() : facade;
-	}
+	constructor(
+		private db: BunDatabase,
+		private reactiveDb: ReactiveDatabase
+	) {}
 
 	// ===== Group lifecycle =====
 
--- a/packages/daemon/src/lib/room/state/session-group-repository.ts
+++ b/packages/daemon/src/lib/room/state/session-group-repository.ts
@@ -18,6 +18,7 @@
 
 import type { Database as BunDatabase } from 'bun:sqlite';
 import { generateUUID } from '@neokai/shared';
+import type { ReactiveDatabase } from '../../../storage/reactive-database';
 import type { GateFailureRecord } from '../runtime/dead-loop-detector';
 
 /** Rate limit backoff state stored in group metadata */
@@ -204,7 +205,16 @@ export interface TaskGroupEvent {
 }
 
 export class SessionGroupRepository {
-	constructor(private db: BunDatabase) {}
+	/** Raw BunDatabase for direct SQL queries */
+	private db: BunDatabase;
+
+	constructor(private reactiveDb: ReactiveDatabase) {
+		// Extract the raw BunDatabase for direct SQL operations.
+		// In production, reactiveDb.db is the proxied Database facade (has getDatabase()).
+		// In tests, reactiveDb.db may be a raw BunDatabase directly.
+		const facade = reactiveDb.db as unknown as { getDatabase?: () => BunDatabase } & BunDatabase;
+		this.db = typeof facade.getDatabase === 'function' ? facade.getDatabase() : facade;
+	}
 
 	// ===== Group lifecycle =====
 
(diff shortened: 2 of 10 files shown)


lsm (commented):

Overall the PR correctly adds `notifyChange('session_groups')` after every session_groups write, updates all construction sites, fixes the proxy binding issue, and adds solid reactive tests. However, there are a few issues worth addressing.

---

### P0: `recordGateFailure` lost its real SQLite transaction

**File:** `packages/daemon/src/lib/room/state/session-group-repository.ts`, lines 776-799

The original code used `this.db.transaction(() => { ... })()` which creates an actual SQLite transaction -- the SELECT + UPDATE are atomic. The refactored code replaced this with `reactiveDb.beginTransaction()` / `commitTransaction()`, but these methods only control **event batching** (see `reactive-database.ts` lines 166-182). They do NOT create an actual database transaction.

This means the read-modify-write in `recordGateFailure` is no longer atomic. Under concurrent access (multiple agents writing to the same group), the metadata read could be stale by the time the write executes, silently dropping concurrent metadata changes.

**Fix:** Wrap the raw SQL operations in `this.db.transaction()` for actual DB atomicity, AND use `reactiveDb.beginTransaction()`/`commitTransaction()` for event batching. Both layers are needed.

---

### P1: `createGroup` multi-table write is not in a real DB transaction

**File:** `packages/daemon/src/lib/room/state/session-group-repository.ts`, lines 232-253

`createGroup` inserts into `session_groups` and then `session_group_members`. If the second INSERT fails (e.g., constraint violation), the first INSERT is already committed, leaving an orphaned `session_groups` row with no members. The `reactiveDb.abortTransaction()` in the catch block only clears pending change events -- it does not roll back the database.

Note: the original code also lacked a real transaction here, so this is not a regression. However, since the code now has try/catch/beginTransaction/commitTransaction that looks like proper transaction handling, it could mislead future maintainers into thinking atomicity is guaranteed when it is not. Consider either adding a real `this.db.transaction()` wrapper, or adding a comment clarifying that `beginTransaction` is event-batching only.

---

### P1: `appendEvent` and `updateWorkerSession` are missing `notifyChange`

**File:** `packages/daemon/src/lib/room/state/session-group-repository.ts`

- **`appendEvent` (line 839):** Writes to `task_group_events` but does not call `notifyChange`. If LiveQuery subscriptions are needed on this table, they will not fire.
- **`updateWorkerSession` (line 819):** Writes to `session_group_members` but does not call `notifyChange`. This changes which session is the worker for a group.

If these tables are not yet subscribed to via LiveQuery, this is a minor gap. But the task description explicitly mentions `notifyChange('session_group_messages')` should be called after respective writes, and `appendEvent` is the closest write method for group event data. Either add notifications or document why they are intentionally omitted.

---

### P2: `recordGateFailure` -- misleading transaction naming

The pattern of `reactiveDb.beginTransaction()` / `commitTransaction()` / `abortTransaction()` strongly implies database transaction semantics (ACID), but these are purely event-batching controls. This naming is inherited from the ReactiveDatabase API, but the usage in `recordGateFailure` is particularly misleading because the original code DID have a real transaction here, and the refactoring appears to be a 1:1 replacement when it is not.

Consider adding a brief comment at each call site: `// Event batching only -- not a database transaction`.

---

All 15 new tests pass and cover the key scenarios well (change events, version-checked no-ops, transaction abort recovery, nesting). The construction site updates are complete and correct. The proxy `.bind(target)` fix is sound.

lsm (commented):

**Recommendation: APPROVE** (posting as COMMENT since reviewer is PR author)

---

### Summary

The implementation is correct and complete. The ReactiveDatabase notification plumbing is wired into every write path in `SessionGroupRepository`, the dual-layer transaction pattern (SQLite for atomicity, ReactiveDatabase for event batching) is well-designed and clearly documented, and the 15-test suite covers the critical paths. All CI checks pass.

---

### What I Verified

- Read the full `session-group-repository.ts` (938 lines) and all changed files
- Traced how `createGroup` and `recordGateFailure` use `this.db.transaction()()` for SQLite atomicity while `reactiveDb.beginTransaction/commitTransaction/abortTransaction` gates event emission — the dual-layer is correct
- Confirmed every write method calls `notifyChange('session_groups')` exactly once, conditional on `result.changes > 0` for version-checked methods
- Confirmed `updateWorkerSession` notifies `session_group_members` and `appendEvent` notifies `task_group_events` — the right tables
- Confirmed `session_group_messages` is intentionally omitted (the table was dropped in migration 19)
- Verified the ReactiveDatabase proxy fix correctly binds non-intercepted methods (`prepare`, etc.) to the original target, not the proxy receiver, preventing `this` context errors for private-field accessors
- Ran all 1346 room unit tests locally — all pass
- Checked all CI jobs — all pass

---

### Issues Found

**P2 — Test gap: SQLite atomicity not verified in transaction error test**

`packages/daemon/tests/unit/room/session-group-repository-reactive.test.ts`, line 278–321

The test for `createGroup` error recovery drops `session_group_members`, calls `createGroup` (which throws), then restores the table and creates a group again. It correctly asserts `abortCount === 1` and that the next write emits a notification.

However, it does not verify that the SQLite rollback actually occurred — i.e., it does not check that no orphaned `session_groups` row was persisted by the failed first INSERT. A follow-up test asserting `db.prepare('SELECT COUNT(*) FROM session_groups').get()` returns 0 after the failed call would give full confidence in the atomicity guarantee.

This is low-stakes because `this.db.transaction()()` is Bun's built-in SQLite transaction (which does roll back automatically), but the test would be more valuable if it explicitly verified the rollback rather than only the event behavior.

**P2 — Type mismatch between `createReactiveDatabase` signature and test usage**

`packages/daemon/src/storage/reactive-database.ts`, line 82; test files throughout

`createReactiveDatabase(db: Database)` declares its parameter as the project's `Database` facade, but every test file passes a raw `bun:sqlite` `Database` instance directly. TypeScript accepts this via structural compatibility, but it creates a silent type-level mismatch. The function could be overloaded or the parameter type relaxed to `Database | BunDatabase` to reflect the actual usage, or tests could pass an explicit cast.

---

### Positive Notes

- The comment `// reactiveDb.beginTransaction() batches change events only — not a DB transaction.` in both `createGroup` and `recordGateFailure` clearly documents the dual-layer design. This is excellent defensive documentation.
- The duck-typed constructor guard (`typeof facade.getDatabase === 'function'`) for extracting the raw BunDatabase is a clean bridge between the test world (raw DB) and production world (facade).
- The transaction depth stuck test (line 366–380) — testing nested `beginTransaction`/`commitTransaction` pairs — is a good regression guard for the nesting logic in `ReactiveDatabase`.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been correctly addressed:

**P0 fix (recordGateFailure atomicity):** The read-modify-write is now wrapped in `this.db.transaction()` for real SQLite atomicity, with `reactiveDb.beginTransaction()`/`commitTransaction()` layered on top exclusively for event batching. The clarifying comment at line 781-782 makes the distinction clear. Correct.

**P1a fix (createGroup atomicity):** Both INSERTs are now wrapped in `this.db.transaction()` (lines 237-251), so a failure in the `session_group_members` INSERT rolls back the `session_groups` INSERT. The test at line 314-337 verifies this by dropping and recreating the members table with a constraint that blocks inserts, then confirming no orphan rows remain. Correct.

**P1b fix (missing notifyChange calls):**
- `appendEvent` now calls `notifyChange('task_group_events')` at line 855. Correct.
- `updateWorkerSession` now calls `notifyChange('session_group_members')` at line 833. Correct.

**P2 fix (clarifying comments):** Both `reactiveDb.beginTransaction()` call sites (lines 232-234 and 781-782) now carry `// Event batching only — not a database transaction` comments. Clear and sufficient.

**Error handling:** Both `createGroup` and `recordGateFailure` use try/catch to call `abortTransaction()` on failure, ensuring `transactionDepth` never gets stuck. Tests at lines 339-382 and 384-425 verify this explicitly by checking `abortCount` and confirming subsequent notifications still emit.

**Test coverage:** The new `session-group-repository-reactive.test.ts` (442 lines, 19 tests) covers change events for all major write operations, version-checked no-op cases, transaction rollback/abort scenarios, and nested transaction depth. All 19 tests pass.

**Minor observation (not blocking):** `createGroup` writes to both `session_groups` and `session_group_members` but only notifies `session_groups`. Similarly, `deleteGroup` only notifies `session_groups` despite CASCADE-deleting members. This is acceptable since consumers watch `session_groups` for lifecycle changes, but worth noting if `session_group_members` subscriptions are added later.

All 238 tests across the 6 related test files pass with no regressions.

lsm (commented):

**Recommendation: APPROVE**

---

This re-review confirms that all four issues from the previous review have been correctly addressed. The implementation is sound and the test suite is thorough.

---

### P0 Fix Verified: recordGateFailure SQLite atomicity

`packages/daemon/src/lib/room/state/session-group-repository.ts`, lines 780–807

The read-modify-write is now wrapped in `this.db.transaction()` for real SQLite atomicity, with `reactiveDb.beginTransaction()`/`abortTransaction()`/`commitTransaction()` layered on top exclusively for event-batching. The clarifying comment at lines 781–782 makes the dual-layer design explicit. The version counter is only incremented by `notifyChange` which is called after the SQLite transaction succeeds, so aborted DB transactions leave the version counter untouched — correct.

### P1a Fix Verified: createGroup multi-table atomicity

`packages/daemon/src/lib/room/state/session-group-repository.ts`, lines 237–251

Both INSERTs are now wrapped in `this.db.transaction()()`. If the `session_group_members` INSERT fails, the `session_groups` INSERT is rolled back atomically. The catch block correctly calls `abortTransaction()` to clear pending event batches and prevent `transactionDepth` from getting stuck. Verified: all 1350 room unit tests pass.

### P1b Fix Verified: appendEvent and updateWorkerSession now call notifyChange

- `appendEvent` calls `notifyChange('task_group_events')` at line 855
- `updateWorkerSession` calls `notifyChange('session_group_members')` at line 833

Both are covered by new tests in `session-group-repository-reactive.test.ts`.

### P2 Fix Verified: clarifying comments at both beginTransaction call sites

Both `createGroup` and `recordGateFailure` carry `// reactiveDb.beginTransaction() batches change events only — not a DB transaction.` comments. This is clear and sufficient.

---

### Construction Site Audit

All call sites for `new SessionGroupRepository` have been updated:

- `room-runtime-service.ts` (lines 372, 588): correctly passes `this.ctx.reactiveDb`
- `task-handlers.ts`: creates `effectiveReactiveDb` falling back to `createReactiveDatabase(db)` when called without `reactiveDb`; all five internal usages use `makeGroupRepo()`
- All test files updated to use `createReactiveDatabase(db)`

The fallback in `task-handlers.ts` (`reactiveDb ?? createReactiveDatabase(db)`) means task-handler code paths that call into the repo will fire change events, even in tests that don't pass a `reactiveDb`. The production path always passes `deps.reactiveDb` from `rpc-handlers/index.ts`.

### Constructor Pattern

The duck-typed extraction (`typeof facade.getDatabase === 'function'`) correctly handles both production (proxied Database facade) and test (proxied raw BunDatabase) without unsafe casting. TypeScript accepts the test usage via structural compatibility.

`createGroup` writes to both `session_groups` and `session_group_members` but notifies only `session_groups`. Similarly, `deleteGroup` CASCADE-deletes members but notifies only `session_groups`. This is acceptable since no production subscriber currently watches `session_group_members` for lifecycle events, but should be revisited if such subscriptions are added later.

---

### Test Results

- 19 new tests in `session-group-repository-reactive.test.ts` — all pass
- 1350 total room unit tests — all pass
- Full typecheck — no errors

The implementation is correct, complete, and well-tested.

PR #644, opened 2026-03-22
feat: remove setInterval from GitHubPollingService and wire handler into GitHubService

lsm:
- Remove setInterval/clearInterval and pollingInterval field from GitHubPollingService;
  start()/stop() become state-flag setters backed by a `running` boolean
- Accept jobQueue and jobProcessor in GitHubServiceOptions; GitHubService.start()
  registers the github.poll handler on jobProcessor and enqueues the initial job
  (with dedup check) — replacing the old dual-mechanism where app.ts seeded the job
  and rpc-handlers/index.ts registered the handler
- app.ts: pass jobQueue/jobProcessor to createGitHubService(), simplify start() call,
  remove seed-job block, remove GITHUB_POLL import
- rpc-handlers/index.ts: remove GITHUB_POLL handler registration (now in GitHubService)
- Add 8 integration-style unit tests in github-service-polling.test.ts covering the
  full flow: start() → register → enqueue → handler → triggerPoll → self-schedule

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -23,7 +23,7 @@ import type { TaskAgentManager } from './lib/space/runtime/task-agent-manager';
 import { JobQueueRepository } from './storage/repositories/job-queue-repository';
 import { JobQueueProcessor } from './storage/job-queue-processor';
 import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
-import { GITHUB_POLL, JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
+import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -225,6 +225,8 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 				config,
 				apiKey,
 				githubToken: process.env.GITHUB_TOKEN, // Optional GitHub token for polling
+				jobQueue,
+				jobProcessor,
 			});
 
 			logInfo('[Daemon] GitHub integration enabled', {
@@ -366,27 +368,11 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	});
 
 	// Start GitHub service after server is ready.
-	// Pass useJobQueueScheduler when polling is configured so the pollingService's
-	// built-in setInterval is skipped — the github.poll job-queue chain owns the
-	// schedule instead, preventing a dual-schedule where both mechanisms fire independently.
+	// GitHubService.start() registers the github.poll handler and enqueues the
+	// initial job when jobProcessor/jobQueue are provided.
 	if (gitHubService) {
-		const useJobQueueScheduler = !!config.githubPollingInterval && config.githubPollingInterval > 0;
-		gitHubService.start({ useJobQueueScheduler });
+		gitHubService.start();
 		logInfo('[Daemon] GitHub service started');
-
-		// Seed the initial github.poll job if polling is configured and no job already exists.
-		// The self-scheduling chain in the handler keeps the chain alive after this seed.
-		if (config.githubPollingInterval && config.githubPollingInterval > 0) {
-			const existing = jobQueue.listJobs({
-				queue: GITHUB_POLL,
-				status: ['pending', 'processing'],
-				limit: 1,
-			});
-			if (existing.length === 0) {
-				jobQueue.enqueue({ queue: GITHUB_POLL, payload: {} });
-				logInfo('[Daemon] Seeded initial github.poll job');
-			}
-		}
 	}
 
 	// Register job handlers BEFORE starting the processor so no pending job
--- a/packages/daemon/src/lib/github/github-service.ts
+++ b/packages/daemon/src/lib/github/github-service.ts
@@ -13,6 +13,10 @@
 import type { Database } from '../../storage/database';
 import type { DaemonHub } from '../daemon-hub';
 import type { Config } from '../../config';
+import type { JobQueueRepository } from '../../storage/repositories/job-queue-repository';
+import type { JobQueueProcessor } from '../../storage/job-queue-processor';
+import { GITHUB_POLL } from '../job-queue-constants';
+import { handleGitHubPoll } from '../job-handlers/github-poll.handler';
 import type {
 	GitHubEvent,
 	RoutingResult,
@@ -50,6 +54,10 @@ export interface GitHubServiceOptions {
 	apiKey: string;
 	/** Optional GitHub token for polling and permission checks */
 	githubToken?: string;
+	/** Job queue repository — required to enqueue the initial github.poll job */
+	jobQueue?: JobQueueRepository;
+	/** Job queue processor — required to register the github.poll handler */
+	jobProcessor?: JobQueueProcessor;
 }
 
 /**
@@ -78,6 +86,8 @@ export class GitHubService {
 	private routerAgent: RouterAgent;
 	private inboxManager: InboxManager;
 	private webhookHandler?: (req: Request) => Promise<Response>;
+	private jobQueue?: JobQueueRepository;
+	private jobProcessor?: JobQueueProcessor;
 
 	constructor(options: GitHubServiceOptions) {
 		this.db = options.db;
@@ -85,6 +95,8 @@ export class GitHubService {
 		this.config = options.config;
 		this.apiKey = options.apiKey;
 		this.githubToken = options.githubToken;
+		this.jobQueue = options.jobQueue;
+		this.jobProcessor = options.jobProcessor;
 
 		// Initialize filter config manager
 		this.filterConfigManager = createFilterConfigManager(this.db.getDatabase());
@@ -117,16 +129,13 @@ export class GitHubService {
 	}
 
 	/**
-	 * Start the GitHub service
-	 * - Starts polling if configured
-	 * - Creates webhook handler if secret is configured
-	 *
-	 * @param options.useJobQueueScheduler - When true, the polling service is created
-	 *   but its built-in setInterval timer is NOT started. The job-queue handler owns
-	 *   the schedule entirely (via the github.poll queue), which avoids a dual-schedule
-	 *   where both the setInterval and the job-queue chain call triggerPoll independently.
+	 * Start the GitHub service.
+	 * - Creates webhook handler if secret is configured.
+	 * - Creates and starts the polling service if polling is configured.
+	 * - Registers the github.poll job handler and enqueues the initial job when
+	 *   both jobProcessor/jobQueue and polling are configured.
 	 */
-	start(options?: { useJobQueueScheduler?: boolean }): void {
+	start(): void {
 		// Initialize webhook handler if secret is configured
 		if (this.config.githubWebhookSecret) {
 			this.webhookHandler = createWebhookHandler(this.config.githubWebhookSecret, async (event) => {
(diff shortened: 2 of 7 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly removes `setInterval`/`clearInterval` from `GitHubPollingService` and moves polling orchestration into the job queue system. The implementation is correct, well-structured, and meets all acceptance criteria.

### What was verified

1. **`setInterval` fully removed** -- `GitHubPollingService.start()` and `stop()` now only toggle a `running` boolean flag. No timer references remain in the github directory.

2. **`pollingInterval` field removed** -- The `Timer | null` field is replaced by a simple `boolean`. The `PollingConfig.interval` property (which is the config value, not a timer handle) remains correctly.

3. **Handler registration moved into `GitHubService.start()`** -- `packages/daemon/src/lib/github/github-service.ts` lines 169-189 register the `github.poll` handler on `jobProcessor` and enqueue the initial job with deduplication, all within the `start()` method. This is a cleaner ownership model than the previous split between `rpc-handlers/index.ts` and `app.ts`.

4. **Old registration removed from `rpc-handlers/index.ts`** -- The 21-line handler registration block and the `GITHUB_POLL` / `handleGitHubPoll` imports are gone.

5. **Old seeding logic removed from `app.ts`** -- The `useJobQueueScheduler` flag, initial job seeding block, and `GITHUB_POLL` import are all removed. `app.ts` now simply calls `gitHubService.start()` and passes `jobQueue`/`jobProcessor` via options.

6. **Execution order in `app.ts` is correct** -- `gitHubService.start()` (line 368) registers the github.poll handler before `jobProcessor.start()` (line 388), ensuring no job is dequeued without a handler.

7. **Closure captures are correct** -- The handler closure in `github-service.ts` line 170-175 captures `this`, so `this.pollingService` is resolved at invocation time (not registration time). The `this.jobQueue!` non-null assertion is safe because it's inside the `if (this.jobProcessor && this.jobQueue)` guard.

8. **Existing `polling-service.test.ts` tests still valid** -- The tests for `start()`/`stop()`/`isRunning()` work correctly with the new boolean-based implementation.

9. **New test file `github-service-polling.test.ts`** covers 8 scenarios: handler registration, initial enqueue, dedup with pending/processing jobs, absent jobProcessor, zero interval, absent token, and full handler invocation flow with triggerPoll and self-scheduling verification.

### Minor observations (P3)

- **`packages/daemon/src/lib/github/github-service.ts` line 166**: The log message says `'Polling service started (job-queue-driven)'` but `pollingService.start()` is called unconditionally (even without jobQueue/jobProcessor). The "job-queue-driven" qualifier is slightly misleading when only `pollingService.start()` runs without the job queue. This is cosmetic and does not affect behavior since `start()` only sets a boolean flag now.

- **`packages/daemon/src/lib/github/github-service.ts` lines 57-60**: `jobQueue` and `jobProcessor` are both optional. The code correctly handles the case where only one is provided (the `if (this.jobProcessor && this.jobQueue)` guard), but callers should always provide both or neither. A comment noting this would be helpful but is not required.

No P0, P1, or P2 issues found. The implementation is clean and correct.

lsm (commented):

**Recommendation: APPROVE** (with one P2 note worth tracking)

---

### Summary

The implementation correctly achieves all stated acceptance criteria: `setInterval`/`clearInterval` and `pollingInterval` are gone from `GitHubPollingService`; `GitHubService.start()` registers the `github.poll` handler and enqueues the initial job; the handler registration was removed from `rpc-handlers/index.ts`; and 8 integration-style unit tests cover the happy path and guard clauses. The code is clean, well-commented, and the ordering in `app.ts` (all registrations before `jobProcessor.start()`) is correct.

---

### Issues Found

#### P2 — `pollingService.stop()` no longer stops polling when job-queue scheduling is active

**Files:** `packages/daemon/src/lib/github/polling-service.ts` lines 73-78, `packages/daemon/src/lib/job-handlers/github-poll.handler.ts` lines 33-43

`GitHubPollingService.stop()` sets `this.running = false`, but the job-queue handler (`handleGitHubPoll`) does not check `pollingService.isRunning()` before calling `triggerPoll()`. Likewise, `triggerPoll()` itself does not check `this.running`. This means that once the job-queue chain is started, calling `gitHubService.stop()` stops the state flag but does not actually halt ongoing polls — the enqueued job will still execute the next cycle. The only way to truly stop polling is to stop the `JobQueueProcessor`, which is done correctly in the shutdown path in `app.ts`. The concern is if `GitHubService.stop()` is ever called at runtime (e.g., via a future RPC handler or test) with the expectation that polling ceases — it will not. Consider having `handleGitHubPoll` skip `triggerPoll` when `pollingService.isRunning()` returns false, or update the docstring for `stop()` to clearly state it only affects the state flag and does not drain the job-queue chain.

#### P2 — Polling service is started unconditionally before job-queue guard check

**File:** `packages/daemon/src/lib/github/github-service.ts` lines 165-188

`this.pollingService.start()` (line 165) is called before the `if (this.jobProcessor && this.jobQueue)` check (line 169). If someone constructs a `GitHubService` with a polling interval and token but without `jobQueue`/`jobProcessor`, the service is silently put into `running=true` state but no poll is ever triggered — the polling service just idles as a dead flag. In practice this cannot happen in production (both are always passed), but it is misleading: `isRunning()` and `isPolling()` would return `true` while nothing is actually happening. Moving `this.pollingService.start()` inside the `if (this.jobProcessor && this.jobQueue)` block, or logging a warning when the guard is not satisfied, would eliminate the silent inconsistency.

#### P3 — No test covering `stop()\' with job-queue-driven service

**File:** `packages/daemon/tests/unit/github/github-service-polling.test.ts`

The eight new tests cover start/register/enqueue/dedup/handler flow thoroughly, but there is no test asserting what happens when `gitHubService.stop()` is called after `start()` in job-queue mode. Given the behavioral subtlety noted in the P2 above, a test documenting the expected behavior (e.g., `triggerPoll` still executes after stop) would prevent future regressions or misunderstandings.

#### P3 — Stale variable reassignment in test has no effect on handler closure

**File:** `packages/daemon/tests/unit/github/github-service-polling.test.ts` lines 248-256

In the handler-invocation test, `listJobsMock = mock(() => [])` reassigns the outer `let listJobsMock` (line 248), but then immediately replaces the `jobQueue.listJobs` reference via private field access (lines 251-256). The outer variable reassignment on line 248 is dead — the captured closure in the handler references the `jobQueue` object stored on `svc`, not the test's `listJobsMock` variable. This does not break the test (the object property replacement on lines 255-256 is what actually matters), but the dead assignment on line 248 is misleading and should be removed.

---

### Positive Observations

- The refactoring cleanly eliminates the dual-schedule hazard (both `setInterval` and job-queue chain firing independently) that existed in the previous implementation.
- Handler registration and initial-job enqueue are correctly co-located in `GitHubService.start()`, making the wiring self-contained and reducing coupling to `rpc-handlers/index.ts` and `app.ts`.
- Deduplication logic (checking `pending` + `processing` before enqueuing) is applied consistently in both `start()` and the self-scheduling handler.
- The comment in `app.ts` explaining that `gitHubService.start()` must run before `jobProcessor.start()` is clear.
- Existing `polling-service.test.ts` tests continue to pass without modification.

lsm (commented):

**Recommendation: APPROVE**

Both P2 issues from the previous review have been correctly addressed. Here are my findings:

---

### P2 Fix 1: handler guard — VERIFIED

`packages/daemon/src/lib/job-handlers/github-poll.handler.ts` now checks `pollingService.isRunning()` before calling `triggerPoll()`. The guard structure is:

- `!pollingService` → warns and skips
- `!pollingService.isRunning()` → debug-logs and skips  
- otherwise → calls `triggerPoll()`

In all skip paths, the self-schedule still runs so the chain can resume.

### P2 Fix 2: `start()` guard — VERIFIED

`pollingService.start()` in `packages/daemon/src/lib/github/github-service.ts` is now inside the `if (this.jobProcessor && this.jobQueue)` guard (lines 170–193). The polling service is only started — and thus reports `isRunning() === true` — when a scheduler is actually wired in.

### Tests — VERIFIED

- `github-service-polling.test.ts`: new tests for stop behavior (line 275) and handler-skips-when-stopped (line 291) are present and correct.
- `github-poll-handler.test.ts`: new test `skips triggerPoll and returns polled: false when pollingService.isRunning() is false` (line 146) directly exercises the guard.

---

### Minor observation (P3)

**File:** `packages/daemon/src/lib/github/github-service.ts`, lines 174–180  
**File:** `packages/daemon/tests/unit/github/github-service-polling.test.ts`, lines 291–335

The closure registered on the job processor reads `this.pollingService` at call time, not at registration time. After `GitHubService.stop()` sets `this.pollingService = undefined`, the handler receives `pollingService: undefined` — it takes the `!pollingService` warn branch, not the `!pollingService.isRunning()` guard.

Practical consequence: the `isRunning() === false` branch in `handleGitHubPoll` is unreachable via the normal `GitHubService.stop()` lifecycle. It would only trigger if someone directly called `pollingService.stop()` without going through `GitHubService.stop()`, which is not a public API path. The test at line 291 patches a saved reference and calls the handler via `svc.jobQueue` internal access, but at invocation time `this.pollingService` is already `undefined`, so the test exercises the `!pollingService` warn branch rather than the intended `isRunning() === false` branch.

The observable result is identical in both cases (no poll, self-schedule fires) so there is no bug, and the defensive guard is harmless. This is a nit on the test comment and the dead-code reachability, not a correctness issue.

---

**Summary:** Both P2 issues are correctly resolved. `isRunning()` guard and `start()` placement are sound. One P3 nit: the `!pollingService.isRunning()` branch in the handler is only reachable if `pollingService.stop()` is called directly, not through `GitHubService.stop()`, because the closure nulls the reference on stop — the integration test for this case exercises the wrong branch, though the outcome is the same.

lsm (commented):

**Recommendation: APPROVE**

Re-reviewed after fixes for the two P2 issues and P3 items from the previous review. All feedback has been addressed correctly.

### Verified Fixes

1. **P2 (isRunning guard)** -- `handleGitHubPoll` at `packages/daemon/src/lib/job-handlers/github-poll.handler.ts` lines 35-39 now checks `pollingService.isRunning()` before calling `triggerPoll()`. When stopped, the handler skips the poll but still self-schedules, allowing the chain to resume if the service is restarted. This correctly addresses the concern that `stop()` had no effect on job-queue-driven polling.

2. **P2 (start placement)** -- `pollingService.start()` is now at `packages/daemon/src/lib/github/github-service.ts` line 171, inside the `if (this.jobProcessor && this.jobQueue)` guard. This eliminates the inconsistency where `isRunning()` could return `true` while no poll chain exists.

3. **P3 (dead assignment)** -- The stale `listJobsMock` reassignment in the integration test has been removed. The test now directly assigns to `jobQueueInService.listJobs`.

4. **P3 (stop tests)** -- Two new tests cover stop behavior: `isPolling() returns false after stop()` (line 275) and `handler skips triggerPoll when pollingService.isRunning() is false` (line 291) in `packages/daemon/tests/unit/github/github-service-polling.test.ts`. The handler-level `isRunning()` check is also tested independently in `packages/daemon/tests/unit/job-handlers/github-poll-handler.test.ts` line 146.

### Implementation Quality

- `setInterval`/`clearInterval` fully removed from `GitHubPollingService`; `start()`/`stop()` are pure state-flag setters.
- Handler registration and initial job seeding are co-located in `GitHubService.start()`, making the wiring self-contained.
- Deduplication logic (checking `pending` + `processing`) is consistent in both `start()` and the self-scheduling handler.
- Execution order in `app.ts` is correct: `gitHubService.start()` registers the handler before `jobProcessor.start()`.
- The `this.pollingService` closure capture in the handler is correct -- it evaluates at call time, so `stop()` (which sets `this.pollingService = undefined`) causes the handler to see `undefined` and skip polling.

### Minor Observation (P3)

- **Test title slightly inaccurate** (`packages/daemon/tests/unit/github/github-service-polling.test.ts` line 291): The test "handler skips triggerPoll when pollingService.isRunning() is false" actually exercises the `pollingService === undefined` path (because `GitHubService.stop()` sets `this.pollingService = undefined`), not the `isRunning() === false` path. The `isRunning()` check path is correctly tested in the handler unit test (`github-poll-handler.test.ts` line 146). This is cosmetic -- the end result (triggerPoll not called, chain still self-schedules) is identical.

PR #647, opened 2026-03-22
test: add online test for session title generation via job queue

lsm:
Adds `packages/daemon/tests/online/lifecycle/session-title-job.test.ts`
with four tests that verify end-to-end title generation through the
persistent job queue:

1. Basic flow: first message enqueues a job, job transitions to
   completed, session title is updated and titleGenerated flag is set.
2. No-duplicate: subsequent messages do not enqueue additional title jobs.
3. Retry on failure: mock causes first attempt to throw; job retries and
   succeeds on second attempt (retryCount=1, result={generated:true}).
4. Dead after exhausted retries: permanently failing mock causes job to
   exhaust maxRetries=2 (3 total attempts) and transition to 'dead';
   session title remains 'New Session'.

All four tests pass under NEOKAI_USE_DEV_PROXY=1.

### DIFF
--- a/packages/daemon/tests/online/lifecycle/session-title-job.test.ts
+++ b/packages/daemon/tests/online/lifecycle/session-title-job.test.ts
@@ -26,7 +26,7 @@ const MODEL = IS_MOCK ? 'haiku' : 'haiku-4.5';
 const IDLE_TIMEOUT = IS_MOCK ? 5000 : 45000;
 const SETUP_TIMEOUT = IS_MOCK ? 15000 : 30000;
 const TEST_TIMEOUT = IS_MOCK ? 30000 : 90000;
-const JOB_POLL_TIMEOUT = IS_MOCK ? 10000 : 30000;
+const JOB_POLL_TIMEOUT = IS_MOCK ? 20000 : 30000;
 
 // Exponential backoff for retry (2^0 * 1000 = 1000ms for first retry)
 const RETRY_DELAY_MS = 1100;
@@ -93,6 +93,12 @@ describe('Session Title Generation via Job Queue', () => {
 
 	beforeEach(async () => {
 		daemon = (await createDaemonServer()) as DaemonWithContext;
+		if (!daemon.daemonContext) {
+			throw new Error(
+				'session-title-job tests require in-process daemon mode. ' +
+					'Unset DAEMON_TEST_SPAWN to run these tests.'
+			);
+		}
 	}, SETUP_TIMEOUT);
 
 	afterEach(async () => {
@@ -218,24 +224,16 @@ describe('Session Title Generation via Job Queue', () => {
 				// Send first message — triggers job enqueue
 				await sendMessage(daemon, sessionId, 'What is 3+3? Reply with the number.');
 
-				// Wait for job to appear in pending/processing state
+				// Wait for job to appear
 				const initialJob = await waitForJobStatus(daemon, SESSION_TITLE_GENERATION, [
 					'pending',
 					'processing',
 					'completed',
-					'failed',
 				]);
 				expect(initialJob.payload.sessionId).toBe(sessionId);
 
-				// After the first attempt fails, job should go back to pending with retryCount=1
-				// (retry delay = 2^0 * 1000ms = 1000ms)
-				const retryingJob = await waitForJobStatus(daemon, SESSION_TITLE_GENERATION, [
-					'pending',
-					'processing',
-					'completed',
-				]);
-
-				// Eventually the job should complete successfully on the second attempt
+				// The job fails on attempt 1, then retries after ~1s backoff (2^0 * 1000ms).
+				// Wait for completion on the second attempt.
 				const completedJob = await waitForJobStatus(daemon, SESSION_TITLE_GENERATION, 'completed');
 				expect(completedJob.id).toBe(initialJob.id);
 				expect(completedJob.retryCount).toBe(1);
--- a/packages/daemon/tests/online/lifecycle/session-title-job.test.ts
+++ b/packages/daemon/tests/online/lifecycle/session-title-job.test.ts
@@ -1,0 +1,314 @@
+/**
+ * Session Title Generation Job Queue Online Tests
+ *
+ * These tests verify that session title generation works end-to-end through
+ * the persistent job queue, including job status transitions and retry-on-failure.
+ *
+ * MODES:
+ * - Real API (default): Requires CLAUDE_CODE_OAUTH_TOKEN or ANTHROPIC_API_KEY
+ * - Dev Proxy: Set NEOKAI_USE_DEV_PROXY=1 for offline testing with mocked responses
+ *
+ * Run with Dev Proxy:
+ *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/lifecycle/session-title-job.test.ts
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import type { DaemonServerContext } from '../../helpers/daemon-server';
+import { createDaemonServer } from '../../helpers/daemon-server';
+import { sendMessage, waitForIdle } from '../../helpers/daemon-actions';
+import type { DaemonAppContext } from '../../../src/app';
+import type { Job } from '../../../src/storage/repositories/job-queue-repository';
+import { SESSION_TITLE_GENERATION } from '../../../src/lib/job-queue-constants';
+
+// Detect mock mode for faster timeouts (Dev Proxy)
+const IS_MOCK = !!process.env.NEOKAI_USE_DEV_PROXY;
+const MODEL = IS_MOCK ? 'haiku' : 'haiku-4.5';
+const IDLE_TIMEOUT = IS_MOCK ? 5000 : 45000;
+const SETUP_TIMEOUT = IS_MOCK ? 15000 : 30000;
+const TEST_TIMEOUT = IS_MOCK ? 30000 : 90000;
+const JOB_POLL_TIMEOUT = IS_MOCK ? 10000 : 30000;
+
+// Exponential backoff for retry (2^0 * 1000 = 1000ms for first retry)
+const RETRY_DELAY_MS = 1100;
+
+type DaemonWithContext = DaemonServerContext & { daemonContext: DaemonAppContext };
+
+/**
+ * Poll job queue until a matching job reaches the expected status.
+ */
+async function waitForJobStatus(
+	daemon: DaemonWithContext,
+	queue: string,
+	expectedStatus: string | string[],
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-written online test file that verifies session title generation end-to-end through the job queue. The four tests cover the key scenarios: basic flow, duplicate prevention, retry-on-failure, and dead-after-exhausted-retries.

### What works well

- **Correct integration with existing infrastructure**: Uses `createDaemonServer` (in-process mode), `sendMessage`, `waitForIdle` helpers appropriately. The `DaemonWithContext` type properly accesses `daemonContext.jobQueue` for direct job inspection.

- **Retry logic is sound**: The monkey-patching approach for tests 3 and 4 correctly replaces `generateTitleAndRenameBranch` on the `SessionLifecycle` instance. Since the registered job handler reads `this.sessionLifecycle` at invocation time (line 127 of session-manager.ts), the mock takes effect for job processing. The `try/finally` blocks ensure the original function is restored even on test failure.

- **Assertions match the implementation**: `maxRetries: 2` (line 133) matches the enqueue call in session-manager.ts (line 163). The dead test correctly expects `retryCount === 2` and `callCount === 3` (initial attempt + 2 retries). The assertion `session.title === 'New Session'` in test 4 is correct because the mock throws before the real function's fallback-title catch block runs.

- **CI integration**: The test file is in `tests/online/lifecycle/`, which is already covered by the `lifecycle` module in the CI matrix (`test_path: tests/online/lifecycle`), so no CI config changes are needed.

- **Timeout management**: Timeouts are appropriately scaled for mock vs real API modes, and the dead-retry test adds extra time for the exponential backoff delays.

### Minor observations (P3)

1. **Redundant intermediate polls in test 3** (lines 222-236): The `waitForJobStatus` calls before the final `completed` check accept nearly every status (`pending`, `processing`, `completed`, `failed`), so they always succeed immediately without verifying any meaningful state transition. They serve as documentation of expected flow but don't add assertion value. Not a problem, just noise.

2. **`RETRY_DELAY_MS` constant** (line 32): Defined as `1100` but only used in the timeout calculation for test 3 (`TEST_TIMEOUT + RETRY_DELAY_MS * 2`). The comment says "Exponential backoff for retry (2^0 * 1000 = 1000ms for first retry)" but the value is 1100ms (with 100ms buffer). This is fine for timeout padding but the naming slightly overstates its purpose.

Overall this is a clean, correct test file that exercises the job queue title generation pipeline end-to-end. No blocking issues found.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (P1 issues)

---

### Summary

The four tests cover the right scenarios — happy path, no-duplicate guard, retry-on-failure, and dead-letter exhaustion — and the assertions are logically correct given the job queue's retry semantics (`maxRetries=2`, exponential backoff). The test structure follows existing patterns in the `lifecycle/` directory closely. Two issues need attention before merging.

---

### P1: Spawned-process mode silently dereferences `undefined` (`daemonContext`)

**File:** `packages/daemon/tests/online/lifecycle/session-title-job.test.ts`, lines 95, 206, 274

```ts
daemon = (await createDaemonServer()) as DaemonWithContext;
// ...
const sessionLifecycle = daemon.daemonContext.sessionManager.getSessionLifecycle();
```

`DaemonWithContext` is a locally-defined intersection that adds `daemonContext: DaemonAppContext`. `createDaemonServer()` only includes `daemonContext` when running in the default in-process mode (`createInProcessDaemonServer`). If `DAEMON_TEST_SPAWN=true` is set (or is added to CI in the future), the TypeScript cast succeeds silently, but `daemon.daemonContext` is `undefined` at runtime, so tests 3 and 4 immediately crash with:

```
TypeError: Cannot read properties of undefined (reading 'sessionManager')
```

The test should add an explicit guard and a clear failure message:

```ts
if (!daemon.daemonContext) {
  throw new Error('session-title-job tests require in-process mode (DAEMON_TEST_SPAWN must not be set)');
}
```

Alternatively, the tests could skip the monkey-patching and use a different injection mechanism compatible with spawned mode.

---

### P1: Unused variable `retryingJob` will confuse readers (and is dead code)

**File:** `packages/daemon/tests/online/lifecycle/session-title-job.test.ts`, lines 232–236

```ts
const retryingJob = await waitForJobStatus(daemon, SESSION_TITLE_GENERATION, [
    'pending',
    'processing',
    'completed',
]);
```

`retryingJob` is declared but never referenced again. The variable is excluded from lint checks (`**/*.test.ts` is in `.oxlintrc.json` ignorePatterns), so this will not be caught automatically. The intent appears to be a synchronization gate waiting for the job to re-enter the queue after the first failure. However:

1. The variable name implies a state that can't be guaranteed (the job may already be `completed` when this resolves, making it a misleading label).
2. The subsequent `waitForJobStatus(..., 'completed')` correctly handles all cases regardless.

Fix: replace the assignment with a void expression or remove the intermediate wait entirely (the final `waitForJobStatus` for `'completed'` is sufficient). If the intent is to wait for the retry delay to elapse, use the already-defined `RETRY_DELAY_MS` constant in an explicit `await new Promise(r => setTimeout(r, RETRY_DELAY_MS))` instead.

---

### P2: Redundant intermediate wait in test 3

Related to the above: the intermediate `waitForJobStatus` call on lines 232–236 adds no observable correctness guarantee beyond what the subsequent `waitForJobStatus(..., 'completed')` already provides. In the fast mock path the job may already be `completed` by the time the first poll runs, so the two polls collapse into a no-op + actual check. Consider removing the intermediate wait and just documenting the 1-second retry delay in a comment before the final `completedJob` poll.

---

- The `RETRY_DELAY_MS = 1100` constant (line 32) is only used in test timeout arithmetic, not in the test body itself. The comment correctly explains the backoff math.
- The `'failed'` status in test 3's initial `waitForJobStatus` (line 226) is technically unreachable (the processor never writes `'failed'`; retries go to `'pending'` and exhausted retries go to `'dead'`). It's harmless but could mislead future readers into thinking the queue can produce `'failed'` jobs from this handler.
- The `waitForTitleGenerated` helper on line 68 could be a long-running polling loop across 10–30 seconds in real-API mode. The polling interval is 100ms (adequate), and the final diagnostic dump on timeout is useful.
- CI coverage: the `lifecycle` module in `main.yml` covers `tests/online/lifecycle` with `mock_sdk: true`, so the new file is automatically included. No CI changes are needed.

lsm (commented):

**Recommendation: APPROVE**

All four issues raised in the previous review round have been addressed correctly:

1. **`daemonContext` guard (P1 fix)**: Lines 96-101 now throw a clear error if `daemonContext` is undefined, preventing the confusing `TypeError` crash when running in spawned-process mode. The error message correctly directs the user to unset `DAEMON_TEST_SPAWN`.

2. **Removed unused `retryingJob` variable (P1 fix)**: Test 3 now goes directly from the initial `waitForJobStatus` (lines 228-232) to the final `waitForJobStatus(..., 'completed')` (line 237). The redundant intermediate poll is gone.

3. **Removed unreachable `'failed'` status (P2 fix)**: The initial `waitForJobStatus` in test 3 (line 228) now only accepts `['pending', 'processing', 'completed']`, which matches the actual states the job queue processor can produce.

4. **Increased `JOB_POLL_TIMEOUT` to 20s (P3 fix)**: Line 29 now uses `20000` in mock mode (was `10000`), providing more headroom for the mock SDK title generation path.

### Remaining observations (P3 only, no action needed)

- **`RETRY_DELAY_MS` naming**: Still defined at line 32 as `1100` and used only for timeout arithmetic in test 3 (line 258) and indirectly for test 4's comment (line 289). The name slightly overstates its role but it serves as useful documentation of the expected backoff timing.

- **Test 2 settle time**: The `500ms` sleep on line 191 is a pragmatic choice for verifying no second job was enqueued. This is a common pattern in async test code and is acceptable here since the alternative (subscription-based waiting for a non-event) would be more complex for no real gain.

The test file is well-structured, covers the requested scenarios (basic flow, duplicate prevention, retry, dead-letter), and integrates correctly with the existing test infrastructure.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review after the worker addressed the P1/P2 issues raised in the previous round. All four fixes have been applied correctly and the implementation is sound.

---

### Verification of the four fixes

**1. `daemonContext` guard in `beforeEach`** — The guard at line 96 throws a clear, actionable error message when `daemon.daemonContext` is falsy (spawned-process mode). The `afterEach` block also handles this correctly: `daemon` is already assigned when `beforeEach` throws, `if (daemon)` passes, and `kill`/`waitForExit` exist on both context types, so cleanup is always safe.

**2. `retryingJob` variable removed** — Test 3 no longer declares an intermediate variable. The flow goes directly from `initialJob` to waiting for `'completed'`.

**3. `'failed'` status removed from initial wait in test 3** — Confirmed absent at line 228. This is correct because the `JobQueueRepository.fail()` method never writes `'failed'` on the way to retrying — it writes `'pending'` when `retryCount < maxRetries` and `'dead'` only when retries are exhausted.

**4. `JOB_POLL_TIMEOUT` raised to 20s in mock mode** — The constant at line 29 reads `IS_MOCK ? 20000 : 30000`. Verified this is used as the default for both `waitForJobStatus` and `waitForTitleGenerated`, and test 4 adds an additional `3500ms` on top for the backoff delays (`2^0 * 1000 + 2^1 * 1000 = 3000ms` total).

---

The monkey-patch in tests 3 and 4 replaces `sessionLifecycle.generateTitleAndRenameBranch` on the live instance. The job handler (`handleSessionTitleGeneration`) calls through to this method without a catch block, so the mock's throw propagates to `JobQueueProcessor.processJob()`'s catch block, which calls `repo.fail()` correctly. The `try/finally` blocks in the tests guarantee the original function is restored even on assertion failure.

The retry math for test 4 is sound: `maxRetries=2` means 3 total attempts (initial + 2 retries). Delays are `2^0 * 1000 = 1000ms` and `2^1 * 1000 = 2000ms`, totaling ~3000ms of backoff. The poll timeout of `JOB_POLL_TIMEOUT + 3500ms` (23500ms in mock mode) covers this comfortably within the test's outer timeout.

The CI matrix entry for `lifecycle` uses `test_path: tests/online/lifecycle`, which auto-discovers all files in the directory, so the new test file is covered without any CI config changes.

---

### One minor observation (P3)

In test 3, the `initialJob` check at line 228 accepts `['pending', 'processing', 'completed']`. Since the mock always throws on the first call, the job cannot reach `'completed'` before `callCount` reaches 1. This list is effectively `['pending', 'processing']` in practice. Not a bug — the code still passes correctly — just a minor documentation imprecision in the status array.

---

All P0, P1, and P2 issues from the prior review are resolved. The test file is correct, complete, and follows project patterns.

PR #649, opened 2026-03-22
feat: add ChannelResolver, list_group_members and relay_message tools to Task Agent

lsm:
- Create ChannelResolver utility that reads resolved channels from workflow run config
  (`run.config._resolvedChannels`) and provides canSend/getPermittedTargets/getResolvedChannels API
- Add list_group_members tool to Task Agent MCP server: lists all group members with
  sessionId, role, agentId, status, and permittedTargets from channel topology
- Add relay_message tool: Task Agent can inject messages into any group member session
  (unrestricted by channel topology); cross-group messaging rejected via DB lookup
- Add sessionGroupRepo and getGroupId to TaskAgentToolsConfig for group validation
- Update handleSubSessionComplete to include completed agent result summary in STEP_COMPLETE msg
- Update Task Agent system prompt to document new tools and channel topology override capability
- Add 23 ChannelResolver unit tests, 10 new relay/list tool tests, 9 new schema tests

### DIFF
--- a/packages/daemon/src/lib/space/agents/task-agent.ts
+++ b/packages/daemon/src/lib/space/agents/task-agent.ts
@@ -200,7 +200,7 @@ export function buildTaskAgentSystemPrompt(context: TaskAgentContext): string {
 	);
 	sections.push(
 		`- **request_human_input** — Surface a human gate and block until the human responds. ` +
-			`Pass a \`prompt\` describing what decision or approval is needed. ` +
+			`Pass a \`question\` describing what decision or approval is needed. ` +
 			`Returns the human's response. ` +
 			`Call this when \`advance_workflow\` returns a \`human\` gate condition.`
 	);
--- a/packages/daemon/src/lib/space/agents/task-agent.ts
+++ b/packages/daemon/src/lib/space/agents/task-agent.ts
@@ -19,6 +19,8 @@
  *   - advance_workflow      — Evaluate transitions from the current step and move to next
  *   - report_result         — Mark the task complete/failed and record the result summary
  *   - request_human_input   — Surface a human gate and block until the user responds
+ *   - list_group_members    — List all group members with session IDs and permitted channels
+ *   - relay_message         — Inject a user-turn message into any group member (unrestricted)
  *
  * ## Content interpolation
  * All operator-supplied content (space.backgroundContext, space.instructions,
@@ -202,6 +204,19 @@ export function buildTaskAgentSystemPrompt(context: TaskAgentContext): string {
 			`Returns the human's response. ` +
 			`Call this when \`advance_workflow\` returns a \`human\` gate condition.`
 	);
+	sections.push(
+		`- **list_group_members** — List all members of the current task's session group. ` +
+			`Returns each member's \`sessionId\`, \`role\`, \`agentId\`, \`status\`, and ` +
+			`\`permittedTargets\` (which roles they can message per the declared channel topology). ` +
+			`Use this to discover active sub-sessions before calling \`relay_message\`.`
+	);
+	sections.push(
+		`- **relay_message** — Inject a user-turn message into any group member's session. ` +
+			`Pass \`target_session_id\` (from \`list_group_members\`) and the \`message\` string. ` +
+			`You are NOT constrained by channel topology — you can relay to any member. ` +
+			`Cross-group messaging is rejected. Use this to route feedback, questions, or context ` +
+			`between step agents that cannot communicate directly.`
+	);
 
 	// ---- step_result vs report_result status ---------------------------------
 	sections.push(`\n## Result Fields: \`step_result\` vs \`report_result.status\`\n`);
@@ -273,6 +288,34 @@ export function buildTaskAgentSystemPrompt(context: TaskAgentContext): string {
 	sections.push(
 		`5. **One step at a time.** Do not spawn multiple step agents concurrently unless the ` +
 			`workflow explicitly defines parallel steps. Follow the linear execution loop above.`
+	);
+
+	// ---- Channel topology and relay capabilities ----------------------------
+	sections.push(`\n## Channel Topology and Message Relay\n`);
+	sections.push(
+		`Workflow steps may declare a channel topology that governs which agent roles can communicate ` +
+			`with each other. Use \`list_group_members\` to see the active members and their ` +
+			`\`permittedTargets\` — roles they are allowed to message per the declared topology.\n`
+	);
+	sections.push(
+		`**You are NOT constrained by the channel topology.** As the Task Agent coordinator, ` +
+			`you can relay messages to any group member using \`relay_message\`, regardless of which ` +
+			`channels are declared. This allows you to:\n` +
+			`- Route feedback from a reviewer agent to a coder agent\n` +
+			`- Forward context or questions between parallel sub-sessions\n` +
+			`- Intervene when the declared topology does not cover a needed communication path\n`
+	);
+	sections.push(
+		`**Cross-group messaging is always rejected.** You can only relay to sessions that are ` +
+			`members of the same group as this task. Attempting to relay to a session in a different ` +
+			`group will return an error.\n`
+	);
+	sections.push(
+		`**When to use relay_message:**\n` +
+			`- A step agent produces output that another step agent needs as input\n` +
+			`- A reviewer step identifies issues that should be sent back to the coder step\n` +
+			`- You need to inject context or instructions mid-execution into a running sub-session\n` +
+			`- The channel topology does not permit direct communication between two agents\n`
 	);
 
 	// ---- Task context -------------------------------------------------------
--- a/packages/daemon/src/lib/space/runtime/channel-resolver.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-resolver.ts
@@ -18,17 +18,40 @@
 
 import type { ResolvedChannel } from '@neokai/shared';
 
+/**
+ * Returns true when `entry` is a structurally valid `ResolvedChannel`.
+ * Filters out malformed DB values (partial serialization, migration bugs, manual edits)
+ * rather than casting them blindly, which would produce undefined behaviour in
+ * `canSend()` / `getPermittedTargets()` (e.g. `ch.fromRole === undefined` always false).
+ */
+function isValidResolvedChannel(entry: unknown): entry is ResolvedChannel {
+	if (!entry || typeof entry !== 'object') return false;
+	const e = entry as Record<string, unknown>;
+	return (
+		typeof e.fromRole === 'string' &&
+		e.fromRole.length > 0 &&
+		typeof e.toRole === 'string' &&
+		e.toRole.length > 0 &&
+		typeof e.fromAgentId === 'string' &&
+		typeof e.toAgentId === 'string' &&
+		e.direction === 'one-way' &&
+		typeof e.isHubSpoke === 'boolean'
+	);
+}
+
 export class ChannelResolver {
 	constructor(private readonly channels: ResolvedChannel[]) {}
 
 	/**
 	 * Build a ChannelResolver from a workflow run's config object.
 	 * Reads `config._resolvedChannels` — stored by `storeResolvedChannels()` in SpaceRuntime.
-	 * Returns an empty resolver if the field is absent or not an array.
+	 * Returns an empty resolver if the field is absent, not an array, or contains only
+	 * invalid entries. Invalid entries are silently filtered rather than casting blindly.
 	 */
 	static fromRunConfig(config: Record<string, unknown> | undefined): ChannelResolver {
 		const raw = config?._resolvedChannels;
-		const channels = Array.isArray(raw) ? (raw as ResolvedChannel[]) : [];
+		if (!Array.isArray(raw)) return new ChannelResolver([]);
+		const channels = raw.filter(isValidResolvedChannel);
 		return new ChannelResolver(channels);
 	}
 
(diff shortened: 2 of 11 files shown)


lsm (commented):

---

### Summary

The implementation is well-structured and covers the requested scope. The `ChannelResolver` class is clean and correct, the two new MCP tools follow established patterns, and the test coverage is solid. However there are a few real issues that should be addressed before merging.

---

### P0 — Blocking

**1. `storeResolvedChannels` only fires at step-start for steps WITH channels declared; multi-step runs will show stale topology**

`/packages/daemon/src/lib/space/runtime/space-runtime.ts` (lines 1006–1007):
```ts
private storeResolvedChannels(runId: string, spaceId: string, step: WorkflowStep): void {
    if (!step.channels || step.channels.length === 0) return; // <-- early return
```

When the workflow advances from a step with channels to a step WITHOUT channels, the run's config still carries the previous step's `_resolvedChannels`. The `ChannelResolver` built in `list_group_members` will therefore return topology from the previous step, not the current one. For a Task Agent calling `list_group_members` mid-workflow, `channelTopologyDeclared` will incorrectly be `true` and members of the previous step will appear to have permitted targets that no longer exist in the current step.

The fix is to clear `_resolvedChannels` (or set it to `[]`) when the new step has no channels. This is a logical correctness bug, not a future-milestone concern.

---

### P1 — Should Fix

**2. `relay_message` allows Task Agent to message itself**

`/packages/daemon/src/lib/space/tools/task-agent-tools.ts` (lines 817–826):

The group membership check validates that `target_session_id` is a member of the group, but the Task Agent is itself a member (added at spawn time with `role: 'task-agent'`). The Task Agent can therefore relay a message to its own session ID, which would inject a spurious user-turn into its own running conversation. No test covers this scenario. Whether this is intentional or not should be documented, and if unintentional, a guard should reject self-relay.

**3. `_resolvedChannels` data is not runtime-validated — malformed entries are cast silently**

`/packages/daemon/src/lib/space/runtime/channel-resolver.ts` line 31:
```ts
const channels = Array.isArray(raw) ? (raw as ResolvedChannel[]) : [];
```

`raw` is cast to `ResolvedChannel[]` without structural validation. If the run config contains a partially serialized or corrupted `_resolvedChannels` array (e.g. after manual DB edits, a migration bug, or a future schema change), `canSend()` / `getPermittedTargets()` will silently accept the bad data and produce undefined behaviour (e.g. `ch.fromRole === undefined` always false, or entries missing `toRole`). Given this data flows from the DB and is presented to the LLM as factual topology, a Zod validation step here would catch issues early and return a safe empty resolver rather than propagating corrupted state.

**4. No test covering `relay_message` to a completed/failed group member**

`/packages/daemon/tests/unit/space/task-agent-tools.test.ts`

The relay implementation does not check the target member's `status` field. A Task Agent can relay to a member marked `completed` or `failed`. Whether this should be allowed or blocked is a design decision, but there is no test covering this edge case. If a completed sub-session no longer has an active streaming query, the `messageInjector` will likely throw — which is caught and returned as a `success: false` response. This is the correct fallback, but the intent should be documented or validated up-front.

---

**5. System prompt describes `request_human_input` parameter as `prompt` but schema uses `question`**

`/packages/daemon/src/lib/space/agents/task-agent.ts` line 205:
```
Pass a `prompt` describing what decision or approval is needed.
```

The `RequestHumanInputSchema` uses the field name `question`, not `prompt`. This inconsistency in the system prompt could cause the LLM to pass the wrong field name when calling the tool. This is a pre-existing issue but the new relay/channel sections were added to this same prompt file, so it should be caught here.

**6. `ChannelResolver.getResolvedChannels()` returns a direct reference to the internal array**

`/packages/daemon/src/lib/space/runtime/channel-resolver.ts` line 58–60:
```ts
getResolvedChannels(): ResolvedChannel[] {
    return this.channels;
}
```

Callers could mutate the returned array and affect subsequent calls on the same resolver instance. A shallow copy (`return [...this.channels]`) would make the class properly immutable, consistent with its read-only design intent.

---

### P3 — Nits

**7. Comment typo in space-runtime.ts** — lines 363 and 782 have a bare `/` before `TODO` (should be `//`). These will appear as syntax oddities in editors.

**8. Test file comment header lists 5 tools, code covers 7** — `/packages/daemon/tests/unit/space/task-agent-tools.test.ts` line 5 still says "5 Task Agent tools" when the module now has 7. Minor but misleading.

---

- `ChannelResolver` class design is clean: well-documented, all four methods have correct semantics, and the static factory safely handles all null/undefined/non-array inputs.
- The DB-based group membership check in `relay_message` (rather than trusting the in-memory map) is the right choice for restart-safety.
- The 23 `ChannelResolver` tests are thorough — bidirectional expansion, hub-spoke, case sensitivity, and empty-topology edge cases are all covered.
- The 10 new relay/list handler tests cover the main happy path and all documented error paths.
- Schema tests are precise and cover both valid and invalid inputs for all 7 tools.
- The `handleSubSessionComplete` change to include the result summary in the `STEP_COMPLETE` notification is a correct and useful enhancement.
- Tool registration in the MCP server, config wiring in `TaskAgentManager`, and the system prompt additions are all consistent with the existing patterns.

lsm (commented):

### Summary

Clean implementation of the ChannelResolver utility, `list_group_members` and `relay_message` MCP tools, system prompt updates, and handleSubSessionComplete enrichment. The code follows existing patterns well, all 104 tests pass, and typecheck is clean.

### Review

**Correctness and Completeness**: The implementation fully addresses the task requirements. The ChannelResolver is a thin, well-scoped utility that reads from `run.config._resolvedChannels` and provides a clear validation API. The `list_group_members` tool correctly composes ChannelResolver with the session group repository to produce a rich member list. The `relay_message` tool properly validates group membership via DB lookup (restart-safe) and delegates to the existing `messageInjector`. The system prompt additions clearly document the new tools and the Task Agent's override capability.

**Architecture fit**: The new code follows established patterns:
- Tool schemas in `task-agent-tool-schemas.ts`, handlers in `task-agent-tools.ts`, MCP registration via `createSdkMcpServer`
- Dependencies injected via `TaskAgentToolsConfig` (new fields: `sessionGroupRepo`, `getGroupId`)
- `SubSessionFactory` wired in `TaskAgentManager` with group membership tracking

**Test coverage**: 23 ChannelResolver tests covering all methods, edge cases (empty, unknown roles, case sensitivity, hub-spoke). 10 new tool handler tests covering error paths (no group, missing group in DB, target not in group, injection failure) and success paths (relay with/without topology). Schema tests updated for the 2 new tools. Good coverage overall.

### Issues

**P3 (nit):** `ChannelResolver.getResolvedChannels()` at `packages/daemon/src/lib/space/runtime/channel-resolver.ts:59` returns a direct reference to the internal `channels` array. A caller could mutate it, breaking the resolver's invariant. Since this is only used in `list_group_members` for read-only mapping, it is not a practical concern today, but returning `[...this.channels]` or marking as `readonly` would be more defensive.

**P3 (nit):** `ChannelResolver.getPermittedTargets()` at `packages/daemon/src/lib/space/runtime/channel-resolver.ts:51` can return duplicate role names if the same `fromRole -> toRole` pair appears multiple times (e.g., from different agent IDs mapping to the same role). This is cosmetic since it is used for display in the `list_group_members` response, but deduplication would produce cleaner output.

**Recommendation: APPROVE** — Only P3 nits identified; no functional issues.

lsm (commented):

### Re-review after iteration 2 fixes

All previously flagged issues have been addressed. The fixes are correct and well-implemented.

#### P0 fix verified: storeResolvedChannels clears stale topology
`space-runtime.ts` lines 1017-1022 now explicitly write `_resolvedChannels: []` when a step has no channels, preventing stale topology from a prior step from leaking. Test at space-runtime.test.ts confirms the advance-from-channel-to-no-channel path clears topology correctly.

#### P1 fix verified: self-relay guard
`task-agent-tools.ts` lines 828-838 reject relay when `targetMember.role === 'task-agent'`. Test at line 2029 covers this path. The guard uses role string rather than session ID -- this is acceptable since there is exactly one task-agent member per group.

#### P1 fix verified: structural validation in ChannelResolver.fromRunConfig
`channel-resolver.ts` lines 27-40 implement `isValidResolvedChannel()` which validates all required fields including non-empty strings for fromRole/toRole, correct direction value, and boolean isHubSpoke. Test at channel-resolver.test.ts lines 77-104 covers mixed valid/invalid entries including null, missing fields, empty strings, and wrong types.

#### P2 fix verified: immutable copy from getResolvedChannels
`channel-resolver.ts` line 85 returns `[...this.channels]` spread copy. Test at lines 106-113 confirms mutations to the returned array do not affect internal state.

#### P3 fix verified: deduplication in getPermittedTargets
`channel-resolver.ts` line 76 uses `[...new Set(targets)]`. Test at lines 219-240 covers the duplicate agent ID case.

### Remaining observations (all P3-level, non-blocking)

**P3: Self-relay guard is role-based, not session-ID-based.**
The guard at `task-agent-tools.ts:831` checks `targetMember.role === 'task-agent'` rather than comparing against the Task Agent's own session ID. This works correctly because the TaskAgentManager registers exactly one member with role `'task-agent'` per group (line 321 of task-agent-manager.ts). However, if a custom agent were ever assigned role `'task-agent'` as a step agent, relay to it would be incorrectly blocked. This is a theoretical edge case given the role is hardcoded at group creation, not user-configurable for step agents.

**P3: Channel topology comment has a slash instead of `//`.**
`space-runtime.ts` lines 363, 782, 790, 778 have comments starting with a single `/` instead of `//` (e.g., `/ TODO: Milestone 6`). These appear to be pre-existing across the file so likely not introduced in this PR, but worth noting.

### Summary

The implementation is solid. ChannelResolver is clean and well-tested with proper validation, immutability, and deduplication. The list_group_members and relay_message tools integrate correctly with the session group repository and channel topology. The storeResolvedChannels fix prevents a real bug (stale topology leaking across steps). Test coverage is thorough across all new behaviors.

**Recommendation: APPROVE** -- only P3 nits remain.

lsm (commented):

**Recommendation: APPROVE**

All P0, P1, and P2 issues from the previous review have been addressed correctly. The implementation is solid. Below are detailed findings from this iteration.

---

### What Was Fixed

**P0 — storeResolvedChannels always writes `_resolvedChannels`**

`/packages/daemon/src/lib/space/runtime/space-runtime.ts` lines 1017–1022 now explicitly writes `[]` when `step.channels` is empty or absent, preventing stale topology from a prior step leaking into a subsequent step that declares no channels. The fix is correct and the logic is clean. Test coverage is solid: both the "start step with no channels" and "advance from channel-step to no-channel-step" cases are covered in `space-runtime.test.ts` lines 2098–2164.

**P1 — Self-relay guard in `relay_message`**

Lines 831–838 in `task-agent-tools.ts` reject any relay target whose `role === 'task-agent'`. The guard is role-based rather than session-ID-based, which is correct because: the Task Agent is the only member that gets registered with this role (confirmed in `task-agent-manager.ts` line 321), and using a role check is more robust against session-ID aliasing across restarts. Test coverage at line 2029 verifies the rejection path and message content.

**P1 — Structural validation via `isValidResolvedChannel`**

The guard (`channel-resolver.ts` lines 27–40) validates all required fields: `fromRole`, `toRole`, `fromAgentId`, `toAgentId`, `direction`, and `isHubSpoke`. Importantly, the optional `label` field (present in the `ResolvedChannel` interface in `space-utils.ts` line 35) is correctly omitted from the guard — it is not load-bearing for correctness of `canSend()` or `getPermittedTargets()`. The test at line 77 verifies that mixed-valid/invalid arrays are filtered correctly, and 2 valid entries pass through.

**P2 — `getResolvedChannels()` returns an immutable copy**

Line 85 returns `[...this.channels]`. The test at line 106 confirms that mutating the returned array does not affect the resolver's internal state.

**P3 — `getPermittedTargets()` deduplicates with `new Set()`**

Line 76. Test at line 219 covers the case where multiple agent IDs share the same role.

---

### Remaining Observations

**P2 — Self-relay guard uses role identity, not session identity**

The guard at `task-agent-tools.ts` line 831 blocks relay to any session with role `task-agent`. This is intentional and documented. However, there is a subtle gap: if a future group configuration were to register a *second* member with role `task-agent` (e.g., a nested task orchestrator), relay to that second member would also be blocked. This is currently a hypothetical — the codebase only creates one task-agent member per group — but the guard's error message says "Cannot relay to the Task Agent's own session", which would be inaccurate for a hypothetically different task-agent session. The design is defensible as documented, and the comment explains it clearly. Worth tracking if nested orchestrators become a feature.

**P2 — `relay_message` does not validate `message` is non-empty**

`RelayMessageSchema` uses `z.string()` without `.min(1)`. An empty string `""` passes schema validation and would be injected into a sub-session as an empty user turn. This is unlikely to cause real harm (the receiving agent would just see a blank turn) but may cause confusing behavior. No change needed for this PR, but worth noting.

**P3 — `isValidResolvedChannel` direction check is hardcoded to `'one-way'`**

Line 37 checks `e.direction === 'one-way'`. This is correct given the contract that `resolveStepChannels()` always produces one-way entries after expansion (bidirectional splits). If `resolveStepChannels` ever produces entries with a different direction value, the filter would silently drop them. This is architecturally sound for the current design and consistent with the `ResolvedChannel` type definition, which specifies `direction: 'one-way'` as a literal type.

**P3 — `check_step_status` system prompt description is slightly misleading**

`task-agent.ts` line 184 says `Pass the \`session_id\` returned by \`spawn_step_agent\`` but `check_step_status` accepts a `step_id`, not a `session_id`. This is a cosmetic inaccuracy in the prompt that could confuse the LLM. Minor.

---

### Test Quality

Test coverage is thorough across all three test files:
- `channel-resolver.test.ts`: 23 tests covering all methods, hub-spoke topology, deduplication, immutability, and filtering.
- `task-agent-tools.test.ts`: 10 new tests for `list_group_members` and `relay_message` covering error paths, success paths, topology integration, self-relay rejection, and injector failures.
- `task-agent-tool-schemas.test.ts`: New schemas tested for valid/invalid inputs.
- `space-runtime.test.ts`: Two topology-clearing tests verify the P0 fix correctly.

All test cases are well-structured and test behavior, not just code paths.

---

### Summary

The iteration-2 fixes are correct, complete, and well-tested. No blocking issues remain. The implementation is production-quality.

PR #650, opened 2026-03-22
test: add online test for GitHub polling via job queue

lsm:
- Create packages/daemon/tests/online/features/github-poll-job.test.ts with
  5 tests verifying end-to-end job queue mechanics for github.poll:
  initial enqueue on startup, pending→processing→completed transition,
  self-scheduling after completion, dedup (≤1 pending at a time), and
  isPolling() state.

- Fix bug in github-poll.handler.ts: the dedup check was including
  'processing' in the status filter, which caused the handler to always
  find itself (it runs while in 'processing' state) and never self-schedule
  the next poll job. Changed to check only 'pending' status.

- Update github-poll-handler.test.ts to match the corrected dedup
  behavior: the "processing job exists" case now correctly documents that
  the handler enqueues the next job (since 'processing' is no longer
  in the filter).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -178,7 +178,9 @@ jobs:
               tests/online/features/message-delivery-mode-queue.test.ts
             mock_sdk: true
           - module: features-2
-            test_path: tests/online/features/message-persistence.test.ts
+            test_path: >-
+              tests/online/features/github-poll-job.test.ts
+              tests/online/features/message-persistence.test.ts
             mock_sdk: true
           - module: git
             test_path: tests/online/git
--- a/packages/daemon/src/lib/job-handlers/github-poll.handler.ts
+++ b/packages/daemon/src/lib/job-handlers/github-poll.handler.ts
@@ -51,13 +51,16 @@ export async function handleGitHubPoll(deps: GitHubPollHandlerDeps): Promise<Git
 	// completion, not from when the job was dequeued.
 	const nextRunAt = Date.now() + intervalMs;
 
-	// Only enqueue the next job if there is no pending or processing job
-	// already in the chain. Checking 'processing' prevents a duplicate chain
-	// from forming under stale-reclaim or slow-poll scenarios.
+	// Only enqueue the next job if no pending job is already waiting.
+	// We check 'pending' only (not 'processing') because the current job is
+	// itself in 'processing' state while this handler runs — including
+	// 'processing' would always find itself and prevent self-scheduling.
+	// A stale/duplicate 'processing' job is handled by the processor's eager
+	// stale-reclamation on startup, which moves it back to 'pending'.
 	// limit: 1 is sufficient — we only need to know if any job exists.
 	const existingJobs = jobQueue.listJobs({
 		queue: GITHUB_POLL,
-		status: ['pending', 'processing'],
+		status: 'pending',
 		limit: 1,
 	});
 
--- a/packages/daemon/tests/online/features/github-poll-job.test.ts
+++ b/packages/daemon/tests/online/features/github-poll-job.test.ts
@@ -1,0 +1,205 @@
+/**
+ * Online test: GitHub polling via job queue
+ *
+ * Verifies end-to-end GitHub polling job queue mechanics without making real
+ * GitHub API calls:
+ * - Initial github.poll job is enqueued on startup
+ * - Job transitions through pending -> processing -> completed
+ * - Self-scheduling: next poll job is automatically enqueued after completion
+ * - Dedup: no duplicate pending poll jobs exist simultaneously
+ *
+ * GitHubPollingService.triggerPoll() is stubbed immediately after daemon
+ * startup to prevent any real GitHub API calls. Even without the stub the
+ * handler is safe (errors are caught internally), but the stub makes intent
+ * explicit and avoids spurious network noise.
+ *
+ * NOTE: dev proxy intercepts Anthropic API calls only. This test does not
+ * send any Claude messages, so no Anthropic calls are made either.
+ *
+ * Run:
+ *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/features/github-poll-job.test.ts
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import type { DaemonServerContext } from '../../helpers/daemon-server';
+import { createDaemonServer } from '../../helpers/daemon-server';
+import type { DaemonAppContext } from '../../../src/app';
+import { GITHUB_POLL } from '../../../src/lib/job-queue-constants';
+import type { Job, JobStatus } from '../../../src/storage/repositories/job-queue-repository';
+
+// ---------------------------------------------------------------------------
+// Test configuration
+// ---------------------------------------------------------------------------
+
+/**
+ * Environment variables injected into the in-process daemon to enable GitHub
+ * polling without touching real credentials.
+ *
+ * GITHUB_POLLING_INTERVAL=300 (5 min) keeps self-scheduled jobs far in the
+ * future so the test can assert a single pending job without racing against a
+ * second real execution.
+ *
+ * GITHUB_TOKEN is a fake value that satisfies the token-presence guard inside
+ * GitHubService. No repositories are added to the polling service, so
+ * triggerPoll() is a no-op even before the stub is applied.
+ */
+const GITHUB_TEST_ENV: Record<string, string> = {
+	GITHUB_POLLING_INTERVAL: '300',
+	GITHUB_TOKEN: 'ghp_fake_token_for_job_queue_test',
+};
+
+/** Maximum time (ms) to wait for a job to reach a desired status. */
+const JOB_WAIT_TIMEOUT_MS = 8000;
+
+/** Polling cadence for job-status checks. */
+const POLL_INTERVAL_MS = 100;
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+type InProcessDaemon = DaemonServerContext & { daemonContext?: DaemonAppContext };
+
+/** Extract the DaemonAppContext from an in-process daemon. */
+function getDaemonCtx(daemon: DaemonServerContext): DaemonAppContext {
+	const ctx = daemon as InProcessDaemon;
+	if (!ctx.daemonContext) {
+		throw new Error(
+			'daemonContext not available — did you run in spawned mode (DAEMON_TEST_SPAWN=true)?'
+		);
+	}
+	return ctx.daemonContext;
+}
+
+/**
+ * Poll the job queue until at least one github.poll job exists with one of
+ * the given statuses, or until the timeout expires.
+ *
+ * Returns the first matching job, or undefined on timeout.
+ */
+async function waitForGitHubPollJob(
+	daemonCtx: DaemonAppContext,
+	statuses: JobStatus[],
+	timeoutMs: number = JOB_WAIT_TIMEOUT_MS
+): Promise<Job | undefined> {
+	const deadline = Date.now() + timeoutMs;
+	while (Date.now() < deadline) {
+		const jobs = daemonCtx.jobQueue.listJobs({ queue: GITHUB_POLL, status: statuses });
+		if (jobs.length > 0) {
+			return jobs[0];
+		}
+		await new Promise<void>((resolve) => setTimeout(resolve, POLL_INTERVAL_MS));
+	}
+	return undefined;
+}
+
+/**
+ * Stub triggerPoll() on the polling service so no real GitHub HTTP requests
+ * are made during the test. Returns true if the stub was applied.
+ */
+function stubTriggerPoll(daemonCtx: DaemonAppContext): boolean {
+	const pollingService = daemonCtx.gitHubService?.getPollingService();
+	if (!pollingService) {
+		return false;
+	}
+	pollingService.triggerPoll = async () => {};
+	return true;
+}
+
+// ---------------------------------------------------------------------------
+// Test suite
+// ---------------------------------------------------------------------------
+
+describe('GitHub polling via job queue (online)', () => {
+	let daemon: DaemonServerContext;
+
+	beforeEach(async () => {
+		daemon = await createDaemonServer({ env: GITHUB_TEST_ENV });
+
+		// Stub triggerPoll as early as possible.  The job processor polls every
+		// 1 s, so this window is reliable.  Even if the first job fires before
+		// the stub is applied it still completes safely because:
+		//   a) no repositories are registered → pollAllRepositories() is a no-op
+		//   b) any error from triggerPoll() is caught inside handleGitHubPoll()
+		const daemonCtx = getDaemonCtx(daemon);
+		stubTriggerPoll(daemonCtx);
+	}, 30_000);
+
+	afterEach(async () => {
+		if (daemon) {
+			daemon.kill('SIGTERM');
+			await daemon.waitForExit();
+		}
+	}, 15_000);
+
+	// -------------------------------------------------------------------------
+
+	test('github.poll job is enqueued on daemon startup', () => {
+		const daemonCtx = getDaemonCtx(daemon);
+
+		// gitHubService.start() enqueues the initial job synchronously before
+		// the daemon returns from createDaemonServer(), so it is visible
+		// immediately — no polling loop needed.
+		expect(daemonCtx.gitHubService).not.toBeNull();
+
+		const jobs = daemonCtx.jobQueue.listJobs({
+			queue: GITHUB_POLL,
+			status: ['pending', 'processing', 'completed'],
+		});
+		expect(jobs.length).toBeGreaterThanOrEqual(1);
+		expect(jobs[0].queue).toBe(GITHUB_POLL);
+	});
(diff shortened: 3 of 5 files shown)


lsm (commented):

### Overall Assessment: REQUEST_CHANGES

The bug fix in `github-poll.handler.ts` is correct and well-reasoned -- the dedup check was indeed blocking self-scheduling because the handler always found itself in 'processing' state. The online test file is thorough and well-documented. However, there is one blocking issue that will cause CI to fail.

---

### P0: New test file not added to CI matrix

**File:** `scripts/validate-online-test-matrix.sh` (lines 59-63) and `.github/workflows/main.yml` (lines 175-182)

The new file `packages/daemon/tests/online/features/github-poll-job.test.ts` is not listed in either:
1. The `FEATURES_FILES` array in `scripts/validate-online-test-matrix.sh`
2. Any `features-*` matrix shard in `.github/workflows/main.yml`

I confirmed this by running the validation script locally:

```
ERROR: packages/daemon/tests/online/features/github-poll-job.test.ts is not in any CI matrix shard for 'features'
FAILED: 1 online test coverage issue(s) found.
```

The `Validate online test matrix coverage` CI step will fail, blocking this PR from merging. You need to:
1. Add `github-poll-job.test.ts` to the `FEATURES_FILES` array in `scripts/validate-online-test-matrix.sh`
2. Add the test path to an existing `features-*` shard or create a new `features-3` shard in the CI matrix in `.github/workflows/main.yml`

---

### P3: Unit test "processing job exists" case is tautological

**File:** `packages/daemon/tests/unit/job-handlers/github-poll-handler.test.ts` (lines 97-110)

The test named "enqueues next job even when a processing job exists (self-scheduling case)" sets `listJobsMock = mock(() => [])` -- the same empty return as the default mock in `beforeEach`. It is functionally identical to the "enqueues next job when no pending or processing job exists" test at line 77. The distinction (that a processing job exists but is invisible to the dedup query) is only in comments, not in any mock setup. This is not a bug, but the test adds no incremental coverage. Consider either removing it or mocking a scenario where `listJobs` is called with different arguments to demonstrate the filtering.

---

### What looks good

- The bug fix logic is sound: checking only `'pending'` (not `'processing'`) for dedup is correct since the current job is itself in `'processing'` state during handler execution.
- The comment explaining why `'processing'` was removed from the dedup check is clear and accurate.
- The online test covers the key scenarios: startup enqueue, state transitions, self-scheduling with timing assertions, dedup invariant, and `isPolling()` state.
- The `stubTriggerPoll` approach with a safety net (no repos registered + error caught internally) is a pragmatic way to avoid real GitHub API calls.
- The `GITHUB_POLLING_INTERVAL=300` choice keeps self-scheduled jobs far enough in the future to avoid race conditions in assertions.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one blocking issue (P0) must be fixed before merge.

---

### P0 — New test file is not registered in the CI matrix (blocking)

`packages/daemon/tests/online/features/github-poll-job.test.ts` is absent from both the CI matrix in `.github/workflows/main.yml` and from the `FEATURES_FILES` array in `scripts/validate-online-test-matrix.sh`. The project enforces an explicit contract: every file in `tests/online/features/` must appear in a `features-*` shard in the matrix, and CI runs `validate-online-test-matrix.sh` as a named step. This PR will **break CI immediately** on merge — the validation step exits 1 with:

```
ERROR: packages/daemon/tests/online/features/github-poll-job.test.ts is not in any CI matrix shard for 'features'
```

I confirmed this by running the script against the branch HEAD.

**Required fix:**

1. Add the file to one of the features shards in `.github/workflows/main.yml`. Given it uses `NEOKAI_USE_DEV_PROXY` and `mock_sdk: true`, the most natural home is `features-2` (alongside `message-persistence.test.ts`, which also runs mocked):

   ```yaml
   - module: features-2
     test_path: >-
       tests/online/features/message-persistence.test.ts
       tests/online/features/github-poll-job.test.ts
     mock_sdk: true
   ```

2. Add the filename to `FEATURES_FILES` in `scripts/validate-online-test-matrix.sh`:

   ```bash
   FEATURES_FILES=(
     auto-title.test.ts
     message-delivery-mode-queue.test.ts
     message-persistence.test.ts
     github-poll-job.test.ts
   )
   ```

---

### P1 — Recovery-after-restart is untested

The task prompt explicitly requested a test verifying "recovery after restart", meaning that a stale `processing` job left by a crashed daemon is reclaimed by `reclaimStale()` on the next startup and eventually re-runs. The PR description acknowledges this path in the handler comment:

> A stale/duplicate 'processing' job is handled by the processor's eager stale-reclamation on startup, which moves it back to 'pending'.

There is no test that actually exercises this path end-to-end (stop daemon with a job in-flight, restart, observe it transitions from `processing` → `pending` → `completed`). The omission means the feature described in the task is only partially verified.

---

### P2 — Dedup test does not assert the count before the initial job completes

The dedup test (line 182–195 of the new test file) calls:

```ts
await waitForGitHubPollJob(daemonCtx, ['completed']);
await waitForGitHubPollJob(daemonCtx, ['pending']);
```

…then counts pending jobs. This verifies post-completion state but not the invariant during the processing window: that `listJobs({ status: ['pending'] })` returns 0 jobs while the initial job is executing (because the handler only checks `pending`, not `processing`). A race between the processor picking up the job and the test reading job counts could theoretically produce a false pass. Adding an assertion immediately after startup (before completion) would make the invariant complete. This is a quality concern, not a correctness failure given the 300 s interval keeps the window large.

---

### What is done well

- **Bug fix is correct.** The original dedup check included `'processing'`, which caused the handler to always find itself and skip self-scheduling. Restricting to `'pending'` only is the right fix and is well-documented inline.
- **Unit tests are updated consistently.** `github-poll-handler.test.ts` reflects the new behavior with a specific test for the processing-job self-scheduling case (line 97–110) and a test asserting `listJobs` is called with `status: 'pending'` only (line 129–142).
- **Test design is appropriate.** The online test correctly stubs `triggerPoll` after daemon startup, avoids real GitHub API calls, uses a 300 s interval to prevent second-execution races, and cleans up via `daemon.kill()` + `waitForExit()`.
- **Timeout values are reasonable.** `JOB_WAIT_TIMEOUT_MS = 8000` with a 1 s processor tick gives 8 tries, and per-test overrides extend to 15 s where needed.

lsm (commented):

### Re-review Assessment: REQUEST_CHANGES

The P0 (CI matrix) and P3 (tautological unit test) issues from the first review are fixed. The new test file is correctly added to `FEATURES_FILES` in `scripts/validate-online-test-matrix.sh` (line 61) and to the `features-2` shard in `.github/workflows/main.yml` (line 182). The unit test replacement at `packages/daemon/tests/unit/job-handlers/github-poll-handler.test.ts` lines 97-112 now directly asserts that `listJobs` is called with `status: 'pending'` (not an array including `'processing'`), which is a meaningful improvement.

However, the P1 issue from the Sonnet review remains unaddressed.

---

### P1 -- Recovery-after-restart scenario is missing (still open)

The task description explicitly requires verifying "recovery after restart." This means testing the following end-to-end path:

1. Start daemon A, let it enqueue and begin processing a `github.poll` job
2. Kill daemon A while the job is in `processing` state (simulating a crash)
3. Start daemon B on the **same database**
4. Verify that the stale `processing` job is reclaimed back to `pending` by `reclaimStale()` (called eagerly in `JobQueueProcessor.start()`)
5. Verify the reclaimed job eventually transitions to `completed`
6. Verify self-scheduling resumes (a new `pending` job exists)

The handler comment at `packages/daemon/src/lib/job-handlers/github-poll.handler.ts` line 59 references this exact mechanism:

> A stale/duplicate 'processing' job is handled by the processor's eager stale-reclamation on startup, which moves it back to 'pending'.

But no test exercises it. The infrastructure supports it: `createInProcessDaemonServer` creates a temp workspace with a `daemon.db` at `${workspace}/daemon.db`. A restart test could create two daemon instances sharing the same DB path (the first stopped before the second starts). The `staleThresholdMs` would need to be set low (e.g., 0 or 1ms) for the test to work without waiting 5 minutes.

This is the core differentiator the task asked for beyond basic polling mechanics -- verifying that the job-queue-based approach provides crash recovery that `setInterval` could not.

---

### Summary of issue status from first review

| Issue | Status |
|-------|--------|
| P0: CI matrix coverage | **Fixed** -- file added to both `main.yml` and `validate-online-test-matrix.sh` |
| P1: Recovery-after-restart test | **Still missing** |
| P2 (Sonnet): Dedup timing assertion | Minor quality concern, not blocking |
| P3: Tautological unit test | **Fixed** -- replaced with meaningful `listJobs` argument assertion |

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The P1 issue (recovery-after-restart scenario) from the previous review remains unaddressed.

---

### What was fixed (good)

**P0 — CI matrix: FIXED**

`github-poll-job.test.ts` is now correctly registered in both places:
- `.github/workflows/main.yml` line ~182: added to the `features-2` shard alongside `message-persistence.test.ts`
- `scripts/validate-online-test-matrix.sh` line 61: added to `FEATURES_FILES`

The multi-file `test_path` uses YAML block scalar (`>-`) correctly.

**Unit test improvement: FIXED**

The tautological `'skips enqueueing when a processing job already exists'` test has been replaced with a meaningful assertion in `packages/daemon/tests/unit/job-handlers/github-poll-handler.test.ts`: it now asserts that `listJobs` is called with `status: 'pending'` (a string, not an array containing `'processing'`), directly documenting and enforcing the correct dedup behavior.

**Bug fix in handler: CORRECT**

`packages/daemon/src/lib/job-handlers/github-poll.handler.ts` line 63 now passes `status: 'pending'` instead of `status: ['pending', 'processing']`. The reasoning is sound: the current job is itself in `processing` state while the handler runs, so including `'processing'` in the dedup check would always find itself and prevent self-scheduling.

---

### P1 — Recovery-after-restart: STILL MISSING

The task description explicitly requires verifying **"recovery after restart"**. The mechanism exists in the codebase:

1. `packages/daemon/src/storage/job-queue-processor.ts` line 40 calls `reclaimStale()` eagerly on startup
2. `packages/daemon/src/lib/github/github-service.ts` lines 184–192 checks for existing `pending`/`processing` jobs on startup and skips enqueuing if any exist
3. `packages/daemon/src/lib/job-handlers/github-poll.handler.ts` line 59 explicitly references this: _"A stale/duplicate 'processing' job is handled by the processor's eager stale-reclamation on startup"_

But no test exercises this code path. What's needed is a test that:
1. Starts daemon A, waits for a `github.poll` job to reach `processing` state
2. Stops daemon A (before the job completes) with a short `staleThresholdMs` (e.g. 1ms) so the job appears stale
3. Starts daemon B pointing to the **same database** (reusing the `workspace` path from daemon A)
4. Verifies the stale `processing` job is reclaimed to `pending`
5. Verifies the reclaimed job eventually reaches `completed`
6. Verifies self-scheduling resumes (a new `pending` job is enqueued)

The `createInProcessDaemonServer` helper at line 562 sets `config.dbPath = ${workspace}/daemon.db`, so reusing a workspace path across two daemon instances is feasible. The `staleThresholdMs` in `JobQueueProcessor` would need to be configurable via test env or a test-specific override to avoid a 5-minute wait.

This scenario is the key differentiator the task was asking for — proving that the job-queue approach handles crashes that `setInterval` could not recover from.

---

### P2 — Dedup timing assertion (minor, unchanged from previous review)

In `packages/daemon/tests/online/features/github-poll-job.test.ts` lines 182–194, the dedup test checks `pendingJobs.length <= 1` at a single point in time after waiting for one completed job and one pending job. Because each new pending job has `runAt` set 300 s in the future (line 178 assertion confirms this), it is unlikely a second pending job would appear from a second execution cycle. The test is functionally correct given the 300s interval, but does not explicitly rule out a race window. This is a low-priority observation.

---

### Summary

| Issue | Status |
|-------|--------|
| P0: CI matrix coverage | **Fixed** |
| P1: Recovery-after-restart test | **Still missing** |
| P2: Dedup timing assertion | Minor quality concern, not blocking on its own |
| P3: Tautological unit test | **Fixed** |

lsm (commented):

### Re-review (Iteration 4): APPROVE

All previously identified issues have been addressed.

---

### P0 -- CI matrix coverage: FIXED (verified in iteration 2)

`github-poll-job.test.ts` is registered in both `.github/workflows/main.yml` (features-2 shard, line ~182) and `scripts/validate-online-test-matrix.sh` (FEATURES_FILES array, line 61).

### P1 -- Recovery-after-restart: FIXED

The new recovery test at `packages/daemon/tests/online/features/github-poll-job.test.ts` lines 240-284 exercises the crash-recovery code path:

1. **Simulates a crash** by inserting a `processing` job directly into the DB with `started_at` set 6 minutes in the past (exceeding the 5-minute stale threshold).
2. **Verifies pre-reclaim state**: confirms the stale job is visible as `processing` (line 266).
3. **Triggers `reclaimStale()`** with a 5-minute cutoff -- the exact same call that `JobQueueProcessor.start()` makes eagerly on daemon startup (`packages/daemon/src/storage/job-queue-processor.ts` line 40).
4. **Asserts reclamation count** is >= 1 (line 271).
5. **Verifies post-reclaim state**: the job is now `pending` (line 278).
6. **Verifies end-to-end completion**: the running processor picks up the reclaimed job and transitions it to `completed` (line 281-283).

The approach of calling `reclaimStale()` directly rather than spinning up a second daemon is a pragmatic tradeoff. It tests the same code path that runs on startup (the processor's eager stale reclamation) and avoids the complexity and flakiness of managing two daemon lifecycle in a single test. The raw SQL insert to simulate the crash is appropriate since it represents state left behind by a terminated process that had no chance to clean up.

### P2 -- Dedup timing assertion: FIXED

The dedup test now includes an immediate post-startup check at lines 213-217:

```typescript
const atStartup = daemonCtx.jobQueue.listJobs({
    queue: GITHUB_POLL,
    status: ['pending', 'processing'],
});
expect(atStartup.length).toBe(1);
```

This validates the invariant during the processing window (before the processor's first tick), not just after completion. Combined with the post-completion assertion at lines 224-228, both phases of the lifecycle are covered.

### P3 -- Tautological unit test: FIXED (verified in iteration 2)

The replacement test at `packages/daemon/tests/unit/job-handlers/github-poll-handler.test.ts` lines 113-126 directly asserts that `listJobs` is called with `status: 'pending'` (string, not array), making the dedup-not-blocking-self-schedule property observable through mock argument inspection.

---

### Additional observations (P3, non-blocking)

1. **Recovery test does not verify self-scheduling resumes after the reclaimed job completes.** After the stale job completes at line 281, the test does not check that a new `pending` job was enqueued for the next cycle. This is already covered by the separate self-scheduling test (lines 189-205), so the omission is acceptable -- it would be redundant coverage.

2. **The `waitForJobById` helper** (lines 103-119) is a clean addition that enables tracking specific jobs through state transitions. Good utility for the recovery test.

---

### Issue status summary

| Issue | Severity | Status |
|-------|----------|--------|
| CI matrix coverage | P0 | Fixed |
| Recovery-after-restart test | P1 | Fixed |
| Dedup timing assertion | P2 | Fixed |
| Tautological unit test | P3 | Fixed |

lsm (commented):

**Recommendation: APPROVE** — Both previously flagged issues (P1 recovery test, P2 dedup startup assertion) have been addressed. One residual concern noted below.

---

### P1 (was missing, now fixed): Recovery after simulated crash

The new `recovery` test at lines 240–284 of `packages/daemon/tests/online/features/github-poll-job.test.ts` correctly:

1. Waits for the initial job to complete so no active processing jobs exist
2. Inserts a stale `processing` job directly into the DB with `started_at` 6 minutes ago (exceeding the 5-minute threshold)
3. Calls `daemonCtx.jobQueue.reclaimStale(Date.now() - 5 * 60 * 1000)` and asserts the return count is >= 1
4. Verifies the job transitions to `pending` immediately after reclamation
5. Waits for the live processor to pick up and complete the job via `waitForJobById`

This mirrors the exact mechanism in `JobQueueProcessor.start()` (line 40 of `job-queue-processor.ts`) and confirms the full recovery pipeline works end-to-end within a running daemon. The tradeoff of not testing a literal daemon restart (which would require spawned mode) is acceptable and the comment makes it clear.

### P2 (was missing, now fixed): Startup dedup assertion

The `dedup` test now asserts `atStartup.length === 1` at lines 213–217, filtering for `['pending', 'processing']` right after `beforeEach` completes. Including both states is the right call — the job will be in `pending` before the first tick or `processing` if the tick fires within the JS event loop before the assertion runs.

**Residual flakiness concern (non-blocking, P3):** The comment on line 215 states "The processor has not picked it up yet because it polls every 1 s". This is not a guarantee — `jobProcessor.start()` calls `this.tick()` immediately (line 45 of `job-queue-processor.ts`), and `processJob` is async but called without `await`. In a heavily loaded CI environment the entire job could reach `completed` before the test body's first synchronous instruction, causing the assertion to fail since `completed` is excluded from the filter. In practice this is very unlikely given the async DB operations involved, but the comment overstates the certainty. Not requesting changes for this — the assertion is a meaningful improvement over no assertion at all, and the risk is low given the test environment.

---

### Summary

- P1 resolved: Recovery test present and correctly structured
- P2 resolved: Startup single-job assertion in the dedup test
- P3 nit: Comment slightly overstates the timing guarantee for the startup assertion

PR #652, opened 2026-03-22
feat: add MessageHub.onClientDisconnect forwarding method

lsm:
Adds onClientDisconnect(handler) to MessageHub that delegates to the
primary transport's onClientDisconnect if the transport supports it.
Returns a no-op unsubscribe when the transport omits the optional method.

Four unit tests cover: handler fires on disconnect, unsubscribe stops
callbacks, no-op when transport lacks support, no-op with no transport.

### DIFF
--- a/packages/shared/src/message-hub/message-hub.ts
+++ b/packages/shared/src/message-hub/message-hub.ts
@@ -190,6 +190,12 @@ export class MessageHub {
 	 * Register a handler for client disconnect events (server-side only)
 	 * Forwards to the primary transport's onClientDisconnect if supported.
 	 * Returns a no-op unsubscribe function if the transport doesn't support it.
+	 *
+	 * NOTE: The transport is resolved at the time this method is called, not when
+	 * a disconnect fires. If the primary transport changes after registration (e.g.
+	 * a new transport is registered with isPrimary:true), this handler stays bound
+	 * to the original transport. In practice this is not an issue because disconnect
+	 * handlers are registered once at server startup before any transport changes.
 	 */
 	onClientDisconnect(handler: (clientId: string) => void): UnsubscribeFn {
 		const transport = this.primaryTransportName
--- a/packages/shared/src/message-hub/message-hub.ts
+++ b/packages/shared/src/message-hub/message-hub.ts
@@ -186,6 +186,22 @@ export class MessageHub {
 		};
 	}
 
+	/**
+	 * Register a handler for client disconnect events (server-side only)
+	 * Forwards to the primary transport's onClientDisconnect if supported.
+	 * Returns a no-op unsubscribe function if the transport doesn't support it.
+	 */
+	onClientDisconnect(handler: (clientId: string) => void): UnsubscribeFn {
+		const transport = this.primaryTransportName
+			? this.transports.get(this.primaryTransportName)
+			: null;
+		if (transport?.onClientDisconnect) {
+			return transport.onClientDisconnect(handler);
+		}
+		// No-op: transport doesn't support disconnect events
+		return () => {};
+	}
+
 	// ========================================
 	// Router Management (Server-side)
 	// ========================================
--- a/packages/shared/tests/message-hub.test.ts
+++ b/packages/shared/tests/message-hub.test.ts
@@ -1117,64 +1117,59 @@ describe('Multi-Transport Support', () => {
 // onClientDisconnect forwarding
 // ========================================
 
+/**
+ * Creates a mock transport with onClientDisconnect support.
+ * Returns both the transport and a simulate function to trigger disconnects.
+ */
+function createMockTransportWithDisconnect(): {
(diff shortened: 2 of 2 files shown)


lsm (commented):

Clean, well-scoped implementation. The `onClientDisconnect` forwarding method correctly delegates to the primary transport, handles the two fallback cases (no transport, transport without the method), and returns proper unsubscribe functions.

**What looks good:**
- Implementation follows the existing pattern used by `getState()` and other primary-transport-delegating methods
- The `onClientDisconnect` method on `IMessageTransport` (types.ts:300) is already declared as optional, so the guard at message-hub.ts:198 is correct
- All four test cases cover the relevant scenarios: happy path, unsubscribe, missing method, no transport
- Tests use a realistic mock pattern consistent with the rest of the test file

**No issues found.** The change is minimal, correct, and well-tested.

Recommendation: **APPROVE**

lsm (commented):

---

### Summary

The implementation is correct and minimal. The 16-line addition to `MessageHub` correctly forwards `onClientDisconnect` to the primary transport via the existing `primaryTransportName`/`transports` pattern, honours the optional nature of the method on `IMessageTransport`, and returns a proper unsubscribe function in all paths. All 4 new unit tests pass; the full test suite and typecheck are clean.

---

### P2 — Primary transport resolved at registration time, not dynamically

**File:** `packages/shared/src/message-hub/message-hub.ts` lines 194–203

The transport lookup happens when `onClientDisconnect` is called, not when a disconnect actually fires. If the primary transport is replaced afterward (another transport registered with `isPrimary: true`, or the original one unregistered), the already-registered handler keeps firing on the old transport while new calls go to the new primary.

In practice, Task 2.6 registers exactly one disconnect handler at server startup before any transport churn, so this is unlikely to bite soon. But it is a silent correctness gap: the method name implies subscription semantics, yet the handler is tightly bound to whichever transport was primary at the moment of the call. A comment noting this limitation (or, longer term, dynamic re-binding) would prevent a future surprise.

---

### P3 — Repetitive inline mock in tests

**File:** `packages/shared/tests/message-hub.test.ts` lines 1121–1154 and 1156–1192

The two tests that exercise a transport with `onClientDisconnect` define an almost-identical inline object literal (only the post-registration flow differs). Extracting a small factory function `createMockTransportWithDisconnect()` would cut the duplication and make future test additions easier.

---

### What is correct

- The optional-chaining guard (`transport?.onClientDisconnect`) correctly handles transports that do not implement the method.
- The no-transport path (`primaryTransportName === null`) is handled and tested.
- The unsubscribe function returned is the one from the transport, so the transport's own bookkeeping is properly cleaned up on unsub — confirmed by the unsubscribe test which asserts the handler is removed from `disconnectHandlers`.
- `IMessageTransport` already had `onClientDisconnect?` declared; no interface change was needed and none was incorrectly made.
- `WebSocketServerTransport` already has the concrete implementation wired to `clientDisconnectHandlers` with error isolation per handler.
- Typecheck passes cleanly; no new lint suppressions introduced.

lsm (commented):

**Recommendation: APPROVE**

Re-reviewed after the worker addressed both items from the prior review:

- **P2 (addressed):** JSDoc NOTE comment at line 194-198 of `message-hub.ts` clearly explains the transport-at-call-time binding and why it is acceptable in practice. Good.
- **P3 (addressed):** `createMockTransportWithDisconnect()` factory extracted at line 1121 of the test file, eliminating the duplicated inline mock. Both happy-path tests now use it. Clean.

No new issues found. The implementation is straightforward, the optional chaining on `transport?.onClientDisconnect` correctly handles the case where the interface method is absent, and the four test cases cover: handler fires, unsubscribe works, no-op for unsupported transport, and no-op with no transport at all.

lsm (commented):

---

### Re-review after feedback was addressed

Both issues from my prior review have been correctly resolved:

**P2 resolved — JSDoc NOTE added.** Lines 194–198 of `packages/shared/src/message-hub/message-hub.ts` now carry an explicit NOTE explaining that the transport is captured at call time, that a primary-transport swap after registration leaves the handler bound to the original transport, and why this is acceptable in practice (disconnect handlers are registered once at server startup). The comment is accurate and proportionate.

**P3 resolved — Factory function extracted.** The new `createMockTransportWithDisconnect()` helper at lines 1124–1151 of `packages/shared/tests/message-hub.test.ts` is used by both the "handler fires" and "unsubscribe" tests, eliminating the duplicated inline object literals. The third test ("transport has no onClientDisconnect") correctly keeps its own inline mock because it intentionally omits the method the factory provides — that is not a duplication issue, it is a distinct fixture.

---

### No new issues found

The implementation remains correct and complete. The forwarding method, optional-chaining guard, no-transport fallback, unsubscribe delegation, JSDoc documentation, and test coverage all look good. No P0/P1/P2/P3 items remain.

PR #653, opened 2026-03-22
feat: plumb clientId into CallContext

lsm:
Add optional clientId field to the CallContext interface so RPC
handlers can identify which WebSocket client originated the request.

- Add clientId optional string to CallContext in message-hub/types.ts
- Populate clientId from the already-extracted metadata in
  handleIncomingRequest when the message carries one
- Add unit tests verifying clientId is present for WebSocket-originated
  requests and absent for in-process calls

### DIFF
--- a/packages/shared/src/message-hub/message-hub.ts
+++ b/packages/shared/src/message-hub/message-hub.ts
@@ -520,6 +520,7 @@ export class MessageHub {
 				sessionId: message.sessionId,
 				method: message.method,
 				timestamp: message.timestamp,
+				...(clientId !== undefined ? { clientId } : {}),
 			};
 			const result = await Promise.resolve(handler(message.data, context));
 
--- a/packages/shared/src/message-hub/types.ts
+++ b/packages/shared/src/message-hub/types.ts
@@ -74,6 +74,11 @@ export interface CallContext {
 	 * Timestamp
 	 */
 	timestamp: string;
+
+	/**
+	 * Client ID — present for WebSocket-originated requests, undefined for in-process calls
+	 */
+	clientId?: string;
 }
 
 /**
--- a/packages/shared/tests/call-context-clientid.test.ts
+++ b/packages/shared/tests/call-context-clientid.test.ts
@@ -9,7 +9,7 @@ import { describe, test, expect } from 'bun:test';
 import { MessageHub } from '../src/message-hub/message-hub';
 import type { IMessageTransport, ConnectionState, CallContext } from '../src/message-hub/types';
 import type { HubMessage } from '../src/message-hub/types';
-import { MessageType, createRequestMessage } from '../src/message-hub/protocol';
+import { createRequestMessage } from '../src/message-hub/protocol';
 import type { HubMessageWithMetadata } from '../src/message-hub/protocol';
 
 // Minimal mock transport that lets tests inject messages directly
@@ -54,11 +54,11 @@ describe('CallContext.clientId', () => {
 	test('clientId is present in CallContext when message carries a clientId (WebSocket-like)', async () => {
 		const { hub, transport } = buildHubWithTransport();
 
-		let capturedContext: CallContext | undefined;
-
-		hub.onRequest('test.method', (_data, ctx) => {
-			capturedContext = ctx;
-			return { ok: true };
+		const handlerCalled = new Promise<CallContext>((resolve) => {
+			hub.onRequest('test.method', (_data, ctx) => {
+				resolve(ctx);
+				return { ok: true };
+			});
 		});
 
 		// Build a request message with a clientId attached (as WebSocketServerTransport does)
@@ -71,11 +71,8 @@ describe('CallContext.clientId', () => {
 
 		transport.inject(req);
 
-		// Give the async handler a tick to run
-		await new Promise<void>((resolve) => setTimeout(resolve, 10));
-
-		expect(capturedContext).toBeDefined();
-		expect(capturedContext!.clientId).toBe('ws-client-abc');
+		const capturedContext = await handlerCalled;
+		expect(capturedContext.clientId).toBe('ws-client-abc');
 
 		hub.cleanup();
 	});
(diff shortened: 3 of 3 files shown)


lsm (commented):

Clean, minimal change that correctly plumbs the existing `clientId` from `HubMessageWithMetadata` into `CallContext`.

**What I verified:**

- The `clientId` variable at line 470 of `message-hub.ts` is already extracted from `HubMessageWithMetadata` and used extensively in `handleIncomingRequest` for routing responses. Adding it to `CallContext` is the natural next step.
- The conditional spread `...(clientId !== undefined ? { clientId } : {})` correctly avoids setting the property when there is no clientId (in-process calls), keeping the contract clean.
- Adding an optional field to `CallContext` is fully backward compatible -- no existing handler signatures or call sites break.
- The two test cases cover the meaningful scenarios: WebSocket-originated (clientId present) and in-process (clientId absent).
- The `MockTransport` in the test is a reasonable minimal implementation of `IMessageTransport`.

No issues found. LGTM.

**Recommendation: APPROVE**

lsm (commented):

---

**Recommendation: APPROVE** — the implementation is correct and complete. Minor observations below.

### Summary

The PR adds `clientId?: string` to `CallContext` and populates it in `handleIncomingRequest` via a conditional spread. The plumbing is correct end-to-end: `WebSocketServerTransport.handleClientMessage` stamps `clientId` onto the message as `HubMessageWithMetadata`, and the hub reads it back with a cast at line 470 of `message-hub.ts`. Two unit tests validate the present/absent contract.

### Correctness

The core change (line 523 in `message-hub.ts`) is:
```ts
...(clientId !== undefined ? { clientId } : {}),
```
This correctly distinguishes `undefined` (no clientId on message) from any truthy/falsy string, which is the right check. Using `!== undefined` rather than a truthiness check means an empty-string clientId would still be forwarded into the context, but that edge case cannot arise in practice because `WebSocketServerTransport.registerClient` generates UUIDs.

The cast on line 470 to `import('./protocol').HubMessageWithMetadata` is a pre-existing pattern in the file — adding `clientId` to the spread does not change the type-safety posture there.

### Test quality

The two tests cover the primary contract:
1. WebSocket-like message with `clientId` set → present in `CallContext`.
2. Plain message without `clientId` → `undefined` in `CallContext`.

**Minor observation (P3):** `MessageType` is imported on line 12 of the test file but never referenced in the test body. The linter (`no-unused-vars`) should flag this, but it is a test file and may be exempt from that rule. Worth cleaning up regardless.

**Minor observation (P3):** The tests use `setTimeout(resolve, 10)` to wait for the async handler. This is a weak synchronization primitive — a 10 ms delay works fine in practice, but a slightly slower CI machine could make it flaky in theory. A promise-based approach that resolves inside the handler would be more robust. This is a pre-existing pattern in the shared test suite, so no regression introduced here.

### No issues found with

- Architecture fit: `clientId` is already an established concept throughout the router, transport, and protocol layers; adding it to `CallContext` is the natural extension.
- Backward compatibility: the field is optional, so existing handlers that do not destructure `clientId` are unaffected.
- Security: `clientId` is an internal UUID generated server-side; it is not user-supplied data flowing into the context, so there is no injection risk.
- Exports: `CallContext` is already exported from `packages/shared/src/message-hub/types.ts` and re-exported via the shared package — no export changes needed.

### Verdict

P0: 0 | P1: 0 | P2: 0 | P3: 2 (unused import + timer-based sync in test)

Approve.

lsm (commented):

---

**Recommendation: APPROVE**

### Re-review after P3 fixes

The second commit (`bc44a31`) correctly addresses both P3 nits from the previous review:

1. **Unused import removed.** The `MessageType` import is gone from the test file. Only necessary imports remain (`CallContext`, `IMessageTransport`, `ConnectionState`, `HubMessage`, `createRequestMessage`, `HubMessageWithMetadata`).

2. **Promise-based synchronization.** The `setTimeout(resolve, 10)` pattern has been replaced with `new Promise<CallContext>((resolve) => { hub.onRequest(..., (_data, ctx) => { resolve(ctx); ... }); })`. The handler promise resolves deterministically when the handler fires -- no timing dependency.

### Implementation review (unchanged, still correct)

- `CallContext.clientId` is optional, so all existing handlers remain compatible.
- The conditional spread `...(clientId !== undefined ? { clientId } : {})` at `packages/shared/src/message-hub/message-hub.ts:523` correctly avoids adding the key when `clientId` is undefined, preserving the clean absent-vs-present contract.
- The `clientId` value originates from `HubMessageWithMetadata`, which is stamped server-side by `WebSocketServerTransport` using generated UUIDs -- no user-supplied data reaches `CallContext.clientId`.
- Tests cover both paths: WebSocket-originated (clientId present) and in-process (clientId absent).

No issues found.

P0: 0 | P1: 0 | P2: 0 | P3: 0

lsm (commented):

**Recommendation: APPROVE**

The P3 nits from the prior round have been cleanly addressed. The implementation is correct and minimal.

---

### What was done

- `packages/shared/src/message-hub/types.ts`: Added `clientId?: string` to `CallContext` with a clear doc comment explaining the WebSocket-vs-in-process distinction.
- `packages/shared/src/message-hub/message-hub.ts` (line 523): Added the spread to populate `clientId` in the context object constructed inside `handleIncomingRequest`. The conditional spread `...(clientId !== undefined ? { clientId } : {})` correctly keeps the property absent (not just `undefined`) when there is no client ID, which is the right behaviour for the `clientId?` optional field.
- `packages/shared/tests/call-context-clientid.test.ts`: The `MessageType` import was removed, and the `setTimeout`-based async synchronisation was replaced with a Promise resolved directly inside the handler — eliminating the fragile timing dependency.

---

**Test description accuracy (P3)**

The second test is named "in-process call" but its mechanics are: a message injected through the mock transport with no `clientId` attached. A true in-process call is `hub.request()`, which sends through the transport and receives the response over the same channel — and does not set `clientId` on the message either. Both code paths produce `clientId === undefined` in the context, so the test result is correct. The label is a slight misnomer, but it does not affect coverage or correctness, and the inline comment ("Plain request message — no clientId attached") accurately describes what the test is doing. This is a documentation nit only.

**Spread idiom (informational)**

`...(clientId !== undefined ? { clientId } : {})` keeps `clientId` fully absent from the object when undefined, which is preferable to `clientId: undefined` for optional-field semantics (e.g., JSON serialisation, `in` operator checks). This is correct and intentional.

---

### Summary

The implementation is minimal, correctly scoped, and well-tested. Both the happy path (WebSocket request with `clientId`) and the absence path (no `clientId`) are covered. The prior P3 issues were resolved as requested. No new issues introduced.

PR #654, opened 2026-03-22
test: add unit tests for TaskAgentManager group persistence and events

lsm:
Creates packages/daemon/tests/unit/task-agent-manager-groups.test.ts with 35
tests covering the full group lifecycle using a call-recording mock
SpaceSessionGroupRepository and an in-process TestDaemonHub.

Tests cover:
- spawnTaskAgent creates group (createGroup/addMember calls + spaceSessionGroup.created event)
- Idempotency: second spawn does not create duplicate groups or events
- Sub-session creation adds member (addMember args + spaceSessionGroup.memberAdded event)
- getMemberCount used for race-free orderIndex
- Sub-session completion updates status to 'completed' + spaceSessionGroup.memberUpdated event
- Sub-session failure (session.error) sets status to 'failed' + memberUpdated event
- fired guard prevents double-update when both error and idle fire
- Cleanup marks group as 'completed' in repo and clears in-memory map
- Non-fatal paths: createGroup/addMember/updateGroup failures do not throw

### DIFF
--- a/packages/daemon/tests/unit/task-agent-manager-groups.test.ts
+++ b/packages/daemon/tests/unit/task-agent-manager-groups.test.ts
@@ -126,10 +126,10 @@ class MockSessionGroupRepository {
 			id,
 			spaceId: params.spaceId,
 			name: params.name,
-			description: params.description ?? null,
-			workflowRunId: params.workflowRunId ?? null,
-			currentStepId: params.currentStepId ?? null,
-			taskId: params.taskId ?? null,
+			description: params.description,
+			workflowRunId: params.workflowRunId,
+			currentStepId: params.currentStepId,
+			taskId: params.taskId,
 			status: params.status ?? 'active',
 			members: [],
 			createdAt: Date.now(),
@@ -149,6 +149,11 @@ class MockSessionGroupRepository {
 		return Array.from(this.groups.values()).filter((g) => g.taskId === taskId);
 	}
 
+	getGroupsBySpace(spaceId: string): SpaceSessionGroup[] {
+		this._recordCall('getGroupsBySpace', [spaceId]);
+		return Array.from(this.groups.values()).filter((g) => g.spaceId === spaceId);
+	}
+
 	addMember(groupId: string, sessionId: string, params: AddMemberParams): SpaceSessionGroupMember {
 		this._recordCall('addMember', [groupId, sessionId, params]);
 		const memberId = `mock-member-${++this.memberCounter}`;
@@ -157,7 +162,7 @@ class MockSessionGroupRepository {
 			groupId,
 			sessionId,
 			role: params.role,
-			agentId: params.agentId ?? null,
+			agentId: params.agentId,
 			status: params.status ?? 'active',
 			orderIndex: params.orderIndex ?? 0,
 			createdAt: Date.now(),
@@ -198,9 +203,9 @@ class MockSessionGroupRepository {
 		return group;
 	}
 
-	deleteGroup(id: string): void {
+	deleteGroup(id: string): boolean {
 		this._recordCall('deleteGroup', [id]);
-		this.groups.delete(id);
+		return this.groups.delete(id);
 	}
 
 	listActiveGroupsWithTaskId(): Array<{ id: string; taskId: string }> {
@@ -515,7 +520,7 @@ describe('TaskAgentManager — group persistence and events', () => {
 				string,
 				AddMemberParams,
 			];
-			expect(typeof groupId).toBe('string');
+			expect(groupId).toBe(ctx.manager.getTaskGroupId(task.id));
 			expect(memberSessionId).toBe(sessionId);
 			expect(params.role).toBe('task-agent');
 			expect(params.status).toBe('active');
@@ -527,7 +532,7 @@ describe('TaskAgentManager — group persistence and events', () => {
 
 			const groupId = ctx.manager.getTaskGroupId(task.id);
 			expect(groupId).toBeDefined();
-			expect(typeof groupId).toBe('string');
+			expect(groupId).toMatch(/^mock-group-/);
 		});
 
 		test('emits spaceSessionGroup.created with correct payload', async () => {
@@ -586,7 +591,7 @@ describe('TaskAgentManager — group persistence and events', () => {
 
 		test('orphaned group is deleted when addMember throws after createGroup', async () => {
 			// Patch addMember to throw after recording the call
-			const origAddMember = ctx.groupRepo.addMember.bind(ctx.groupRepo);
+			const _origAddMember = ctx.groupRepo.addMember.bind(ctx.groupRepo);
 			let addMemberCallCount = 0;
 			ctx.groupRepo.addMember = (...args: unknown[]) => {
 				addMemberCallCount++;
@@ -605,8 +610,6 @@ describe('TaskAgentManager — group persistence and events', () => {
 			expect(ctx.daemonHub.eventsOf('spaceSessionGroup.created').length).toBe(0);
 			// No group in memory map
 			expect(ctx.manager.getTaskGroupId(task.id)).toBeUndefined();
-
-			void origAddMember; // prevent unused-var warning
 		});
 
 		test('no spaceSessionGroup.created event when createGroup throws', async () => {
@@ -697,7 +700,7 @@ describe('TaskAgentManager — group persistence and events', () => {
 			expect(params.role).toBe('coder');
 			expect(params.agentId).toBe('agent-001');
 			expect(params.status).toBe('active');
-			expect(typeof groupId).toBe('string');
+			expect(groupId).toBe(ctx.manager.getTaskGroupId(task.id));
 		});
 
 		test('uses getMemberCount for orderIndex to prevent race conditions', async () => {
@@ -833,6 +836,13 @@ describe('TaskAgentManager — group persistence and events', () => {
 				{ agentId: 'agent-coder', role: 'coder' }
 			);
 
+			// Get the member ID before resetting calls
+			const groupId = ctx.manager.getTaskGroupId(task.id)!;
+			const subMember = ctx.groupRepo
+				.getGroup(groupId)!
+				.members.find((m) => m.sessionId === subSessionId)!;
+			const expectedMemberId = subMember.id;
+
 			ctx.groupRepo.resetCalls();
 			await callHandleComplete(ctx.manager, task.id, 'step-1', subSessionId);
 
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured, thorough test file covering group persistence and event emission for `TaskAgentManager`. All 35 tests pass (105 assertions) and the test logic correctly mirrors the production code paths.

### Strengths

- **MockSessionGroupRepository** is well-designed: it maintains in-memory state for consistency while recording all method calls for precise spy-style assertions. This is better than pure stubs because it allows verifying both call patterns AND state transitions.

- **Error path coverage** is excellent — tests verify non-fatal behavior when `createGroup` throws (line 567), when `addMember` throws after group creation (line 587, verifying orphan cleanup via `deleteGroup`), and when `updateGroup` throws during cleanup (line 1136). These are exactly the edge cases that matter in production.

- **Fired guard / idempotency tests** are valuable — verifying that `session.error` fires only once (line 993), that idle after error does not overwrite failed status (line 1016), and that duplicate `spawnTaskAgent` calls don't create duplicate groups (lines 629-673).

- The test accesses private methods (`createSubSessionFactory`, `handleSubSessionComplete`) via controlled helper functions (`getFactory`, `callHandleComplete`) rather than scattered casts, keeping the test readable.

### Minor Observations (P3)

1. **Warn log noise in test output** (lines 567-622): The error-path tests produce `[kai:daemon:task-agent-manager]` warn logs in stdout. This is expected behavior from the unit test setup (setup.ts suppresses `console` but not the Logger class). Not blocking, but the test output is noisy. Consider whether the Logger could be silenced in test mode.

2. **`setTimeout(r, 10)` for event propagation** (lines 938, 959, 980, 1010, 1038, 1047, 1224): The `TestDaemonHub.emit()` is synchronous (returns `Promise.resolve()`), so the 10ms delays are technically unnecessary since listeners fire synchronously within `emit()`. They work correctly and don't cause flakiness, but `await Promise.resolve()` or `await ctx.daemonHub.emit(...)` would be more precise about what's being awaited. Very minor.

3. **`void origAddMember`** (line 609): The unused-var suppression is pragmatic but slightly unusual. Consider using `_origAddMember` naming convention instead, though this works with the linter as-is.

### Verification

- All 35 tests pass with 105 expect() calls
- Tests correctly exercise the production code paths in `task-agent-manager.ts` (lines 307-351 for group creation, 654-729 for sub-session factory, 590-644 for cleanup, 741-851 for completion callbacks)
- Mock shapes match the real `SpaceSessionGroupRepository` interface methods used by the production code
- Event names (`spaceSessionGroup.created`, `spaceSessionGroup.memberAdded`, `spaceSessionGroup.memberUpdated`) match production emit calls
- The `TestDaemonHub` correctly routes events by sessionId, matching the real DaemonHub's subscription pattern

No P0, P1, or P2 issues found. The tests are comprehensive, well-organized, and correctly verify the behavior described in the task.

lsm (commented):

The test file is well-structured and the core happy-path scenarios are solid. All 35 tests pass, lint is clean, and TypeScript compiles without errors. The mock design is clean and the TestDaemonHub event bus faithfully mirrors the production DaemonHub's session-scoped routing. That said, there are meaningful gaps and a few correctness issues.

---

### P1 — Missing tests for the session.updated → idle normal completion path

The tests call the private `handleSubSessionComplete` directly via a cast (`callHandleComplete`) to verify that `updateMemberStatus('completed')` is called and `spaceSessionGroup.memberUpdated` is emitted. But the **actual production completion flow** goes through `registerCompletionCallback`, which subscribes to `session.updated` events. When a sub-session transitions to `idle` and `sdkMessageCount > 0`, the registered callback is fired.

There is no test that:
1. Sets `_sdkMessageCount > 0` on a mock sub-session
2. Emits `session.updated` with `{ processingState: { status: 'idle' } }`
3. Verifies that the registered `onComplete` callback fires

Without this test, the normal completion flow through the event listener is completely untested. The "idle after error" guard test (line 1016) emits `session.updated` but only exercises the negative case. The test at line 1049 `expect(completionFired).toBe(false)` verifies blocking, not triggering.

### P1 — `rehydrateGroupMaps` has zero test coverage

The `rehydrateGroupMaps` private method (called by `rehydrate()`) rebuilds the `taskId → groupId` in-memory map from active groups. This is a critical daemon-restart path. If it's broken, all group event emission for rehydrated tasks fails silently. No test calls `manager.rehydrate()` or verifies that `getTaskGroupId` returns the right value after a simulated restart.

### P2 — Weak assertions using `typeof x === 'string'`

Five assertions use the pattern `expect(typeof x).toBe('string')` (lines 518, 530, 700, 843, 873). These pass for any string including the empty string. For example, line 700 checks that the `groupId` passed to `addMember` is a string, but doesn't verify it matches the group ID created by `createGroup`. A stronger assertion would be:

```ts
// line 518 — verify groupId is the mock group ID created earlier
const groupId = ctx.groupRepo.callsTo('createGroup')[0]...
expect(memberGroupId).toBe(groupId);
```

### P2 — Mock `SpaceSessionGroup` fields use `null` where the type requires `undefined`

The mock `createGroup` at line 125 constructs a `SpaceSessionGroup` with:
```ts
description: params.description ?? null,   // typed as string | undefined
workflowRunId: params.workflowRunId ?? null, // typed as string | undefined
currentStepId: params.currentStepId ?? null, // typed as string | undefined
taskId: params.taskId ?? null,              // typed as string | undefined
```

`SpaceSessionGroup` uses optional fields (`?`) which means `string | undefined`, not `string | null`. TypeScript accepts this through the `as unknown as SpaceSessionGroupRepository` cast, but any consumer doing a strict `=== undefined` check on the returned group (e.g. in the event payload) would get the wrong answer. The real `createGroup` in the repository calls `this.getGroup(id)!` which maps SQL `NULL` to the JavaScript `undefined` equivalent via the row mapper at line 394-397. The mock diverges from this behavior.

### P2 — Mock `deleteGroup` return type mismatch

The real `SpaceSessionGroupRepository.deleteGroup` returns `boolean`, but the mock's `deleteGroup` returns `void`. The `as unknown as SpaceSessionGroupRepository` cast suppresses the TypeScript error. The production code doesn't use the return value, so there's no runtime breakage today, but this is a silent type divergence that could mislead future maintainers.

### P2 — `setTimeout(r, 10)` timing in session.error tests

Five tests (lines 937, 959, 979, 1010, 1038, 1047) use `await new Promise((r) => setTimeout(r, 10))` to wait for async event handling. Since `TestDaemonHub.emit` dispatches handlers synchronously and its `emit` returns `Promise.resolve()`, the internal `updateMemberStatus` call happens synchronously within the event handler. The `setTimeout(r, 10)` is not needed and adding time-based waits to unit tests creates fragile coupling to scheduler behavior. The `session.error` handler in the production code executes `updateMemberStatus` synchronously (no awaits inside the handler), so `await daemonHub.emit(...);` alone would be sufficient.

### P3 — `getGroupsBySpace` not present in the mock

The real repo exposes a `getGroupsBySpace(spaceId: string)` method that the mock omits. It's not called in the tested code paths, so there's no runtime breakage, but the mock is structurally incomplete as a test double.

### P3 — Private method access pattern could use a named type alias

`getFactory` and `callHandleComplete` cast to `{ createSubSessionFactory: ... }` and `{ handleSubSessionComplete: ... }` inline. Extracting a named `PrivateTaskAgentManager` interface at the top of the test file would reduce repetition and make the intent clearer. Minor style point.

---

### What is well done

- The `TestDaemonHub` event bus correctly implements session-scoped routing (`event:sessionId` vs `event:*`) matching the real `TypedHub` contract, which is essential for the error-listener tests to work correctly.
- Idempotency tests are thorough and verify both the return value AND the absence of duplicate repo/event calls.
- The non-fatal error paths (createGroup throws, addMember throws, updateGroup throws) are all exercised — this is valuable because these are easy to silently break.
- The `fired` guard preventing double-update is explicitly tested with two rapid `session.error` emissions.
- `makeCtx()` provides good isolation per test and the `afterEach` cleanup (spy restore + temp dir removal) is correct.
- All 35 tests pass, no linter errors, no TypeScript errors.

lsm (commented):

**Recommendation: APPROVE**

All previously reported P1 and P2 issues have been correctly addressed in commit `babcd267`. The test suite is now at 40 tests / 111 expect() calls and all tests pass cleanly.

---

### Previously Reported Issues — Verified Fixed

**P1: session.updated → idle callback tests**
Added at lines 919–973. The two tests correctly exercise the `sdkMessageCount > 0` guard and the `sdkMessageCount === 0` no-fire guard using `await Promise.resolve()` which is sufficient because the TestDaemonHub dispatches handlers synchronously.

**P1: Rehydration tests (3 tests)**
Added at lines 1312–1367. All three scenarios are well-constructed:
- Test 1: seeds mock repo directly and verifies `rehydrate()` populates the map.
- Test 2: verifies the no-overwrite invariant when the in-memory map is already populated.
- Test 3: verifies groups without `taskId` are silently skipped.

**P2: null→undefined for optional fields**
Fixed in `createGroup` (lines 129–132) and `addMember` (line 165). Fields are now stored as `undefined` from params, matching the `SpaceSessionGroup` and `SpaceSessionGroupMember` TypeScript types.

**P2: deleteGroup returns boolean**
Fixed at line 206–208. `this.groups.delete(id)` correctly returns `boolean`, matching the real repository interface.

**P2: Strengthened assertions**
Strong value-based comparisons are present throughout:
- Line 523: `expect(groupId).toBe(ctx.manager.getTaskGroupId(task.id))`
- Line 535: `expect(groupId).toMatch(/^mock-group-/)`
- Line 703: `expect(groupId).toBe(ctx.manager.getTaskGroupId(task.id))`
- Line 853: `expect(memberId).toBe(expectedMemberId)`
- Line 889: `expect(payload.memberId).toBe(subMember.id)`

**P2: Promise.resolve() instead of setTimeout**
Confirmed throughout — no `setTimeout` calls remain.

**P3: getGroupsBySpace added, origAddMember renamed**
Both present.

---

### Remaining Minor Observations

**P3 — Dead variable in test (line 594)**
`const _origAddMember = ctx.groupRepo.addMember.bind(ctx.groupRepo)` is declared but never read. The underscore prefix silences the linter, but the variable could be removed entirely. It has no semantic value after the refactor.

**P3 — getGroupsBySpace in mock (line 152–155)**
This method is not called by `TaskAgentManager` in production (confirmed via grep). Adding it to the mock for completeness is harmless, but it is dead code within the test file. No action required.

---

### Overall Assessment

The implementation is solid. The mock is faithful to the production interface, the event routing in TestDaemonHub correctly matches the session-scoped `on()` / `emit()` pattern used by the real DaemonHub, and all 8 methods called by the production code are implemented. The test coverage now addresses all lifecycle paths: group creation, member addition, idle completion, error failure, fired-guard idempotency, cleanup, and rehydration.

PR #655, opened 2026-03-22
feat: define named-query registry with column aliasing and JSON parsing

lsm:
Adds `NAMED_QUERY_REGISTRY` in `live-query-handlers.ts` — a Map of three
initial named queries (tasks.byRoom, goals.byRoom,
sessionGroupMessages.byGroup) used by the upcoming liveQuery.subscribe
RPC handler.

- All snake_case columns are aliased to camelCase via SQL AS clauses
- RoomGoal snake_case exceptions preserved: planning_attempts,
  goal_review_attempts remain unaliased per the TypeScript type
- JSON blob columns parsed in mapRow: NeoTask.dependsOn, RoomGoal.linkedTaskIds,
  RoomGoal.metrics, RoomGoal.structuredMetrics, RoomGoal.schedule
- SQLite boolean coercion for RoomGoal.schedulePaused (0/1 → boolean)
- All ORDER BY clauses include deterministic id tiebreakers
- 28 contract tests cover column aliasing, JSON parsing, ORDER BY
  invariants, and end-to-end row shape vs TypeScript types

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts
@@ -1,0 +1,200 @@
+/**
+ * LiveQuery RPC Handlers
+ *
+ * Defines the server-side named-query registry for the liveQuery.subscribe /
+ * liveQuery.unsubscribe RPC protocol.  Clients send a query name + parameters;
+ * the daemon resolves it to a pre-registered SQL template and row mapper.
+ * Clients never send raw SQL.
+ */
+
+// ============================================================================
+// Named-query registry types
+// ============================================================================
+
+export interface NamedQuery {
+	/** Parameterised SQL that will be executed by LiveQueryEngine */
+	sql: string;
+	/** Number of positional parameters the SQL expects */
+	paramCount: number;
+	/**
+	 * Optional row transformer applied after every query execution.
+	 * Must return a plain object whose keys match the frontend TypeScript types.
+	 */
+	mapRow?: (row: Record<string, unknown>) => Record<string, unknown>;
+}
+
+// ============================================================================
+// Row mappers
+// ============================================================================
+
+/**
+ * Map a raw SQLite row from the `tasks` table (with camelCase AS aliases) to
+ * the NeoTask shape expected by the frontend.
+ *
+ * JSON blob columns: `dependsOn` (stored as `depends_on`).
+ * All other snake_case→camelCase conversions are handled by AS aliases in the
+ * SQL itself so this mapper only handles post-processing that SQL cannot do.
+ */
+function mapTaskRow(row: Record<string, unknown>): Record<string, unknown> {
+	return {
+		...row,
+		dependsOn: JSON.parse((row.dependsOn as string | null) ?? '[]') as string[],
+	};
+}
+
+/**
+ * Map a raw SQLite row from the `goals` table (with camelCase AS aliases) to
+ * the RoomGoal shape expected by the frontend.
+ *
+ * JSON blob columns:
+ *   - `linkedTaskIds` (stored as `linked_task_ids`)
+ *   - `metrics`       (stored as `metrics`)
+ *   - `structuredMetrics` (stored as `structured_metrics`) — optional, null-safe
+ *   - `schedule`      (stored as `schedule`)               — optional, null-safe
+ *
+ * Boolean coercion:
+ *   - `schedulePaused` — SQLite stores 0/1; convert to JS boolean
+ *
+ * snake_case exceptions (per TypeScript type):
+ *   - `planning_attempts`    — kept as-is (not aliased in SQL, not converted here)
+ *   - `goal_review_attempts` — kept as-is (not aliased in SQL, not converted here)
+ */
+function mapGoalRow(row: Record<string, unknown>): Record<string, unknown> {
+	return {
+		...row,
+		linkedTaskIds: JSON.parse((row.linkedTaskIds as string | null) ?? '[]') as string[],
+		metrics: JSON.parse((row.metrics as string | null) ?? '{}') as Record<string, number>,
+		structuredMetrics:
+			row.structuredMetrics != null
+				? (JSON.parse(row.structuredMetrics as string) as unknown[])
+				: undefined,
+		schedule: row.schedule != null ? (JSON.parse(row.schedule as string) as unknown) : undefined,
+		schedulePaused: row.schedulePaused === 1,
+	};
+}
+
+// ============================================================================
+// SQL definitions
+// ============================================================================
+
+const TASKS_BY_ROOM_SQL = `
+SELECT
+  id,
+  room_id             AS roomId,
+  title,
+  description,
+  status,
+  priority,
+  progress,
+  current_step        AS currentStep,
+  result,
+  error,
+  depends_on          AS dependsOn,
+  created_at          AS createdAt,
+  started_at          AS startedAt,
+  completed_at        AS completedAt,
+  task_type           AS taskType,
+  assigned_agent      AS assignedAgent,
+  created_by_task_id  AS createdByTaskId,
+  archived_at         AS archivedAt,
+  active_session      AS activeSession,
+  pr_url              AS prUrl,
+  pr_number           AS prNumber,
+  pr_created_at       AS prCreatedAt,
+  input_draft         AS inputDraft,
+  updated_at          AS updatedAt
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** (same-author review posted as COMMENT)

---

Overall the implementation is clean and well-structured. The SQL aliases, row mappers, and test coverage are thorough. However there are two issues worth addressing before this merges.

---

### P1 — `sessionGroupMessages.byGroup` references a non-existent table

**File:** `packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts` lines 147–159

Migration 19 (`packages/daemon/src/storage/schema/migrations.ts` line 1020) explicitly drops `session_group_messages`:

```ts
db.exec(`DROP TABLE IF EXISTS session_group_messages`);
```

And `createTables()` in `packages/daemon/src/storage/schema/index.ts` never recreates it. The table literally does not exist in any fresh or migrated database.

When a client calls `liveQuery.subscribe` with `queryName: 'sessionGroupMessages.byGroup'`, the `LiveQueryEngine` will attempt `db.prepare(sql).all(groupId)` and SQLite will throw `no such table: session_group_messages`. This will crash the subscription handler at runtime.

The comment in the code says the table is "populated in Milestone 4", but the table needs to be *created* first too. The registry entry is forward-declared correctly, but the missing `CREATE TABLE` is a runtime bug that will surface the moment any code path exercises this named query.

The tests for this entry only inspect the SQL string — they never execute the SQL — so the test suite gives false confidence:

```ts
// This passes, but the SQL itself will throw at runtime:
test('SQL targets session_group_messages table', () => {
    const entry = NAMED_QUERY_REGISTRY.get('sessionGroupMessages.byGroup')!;
    expect(entry.sql).toContain('FROM session_group_messages');
});
```

**Fix:** Either add the missing `CREATE TABLE IF NOT EXISTS session_group_messages (...)` DDL in `createTables()` or in a new migration (whichever Milestone 4 owns), or gate the subscribe handler so it returns an error if the backing table is not yet available. At minimum, add a test that actually executes the SQL so CI catches this.

---

### P2 — Dead code: `flushMicrotasks` is defined but never used

**File:** `packages/daemon/tests/unit/rpc-handlers/live-query-handlers.test.ts` lines 25–27

```ts
function flushMicrotasks(): Promise<void> {
    return new Promise((resolve) => queueMicrotask(resolve));
}
```

This function is defined but never called anywhere in the 316-line test file. The project enforces `no-unused-vars` as an error in oxlint (see `.oxlintrc.json`), so this will fail the lint check in CI. It should be removed.

---

### P3 — `JSON.parse` errors are unhandled in both row mappers

**File:** `packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts` lines 38–43, 62–73

If a row contains corrupt JSON in `dependsOn`, `metrics`, `linkedTaskIds`, `structuredMetrics`, or `schedule`, `JSON.parse` will throw synchronously. The exception will propagate uncaught through the `LiveQueryEngine`. This is consistent with `TaskRepository.rowToTask()` and `GoalRepository.rowToGoal()`, so it is not a regression, but it is worth noting that corrupt data will terminate the subscription delivery. A guard like `JSON.parse(str ?? '[]'))` wrapped in a try/catch (or using a safe-parse helper) would improve resilience.

---

### Positive notes

- Column aliasing is correct and matches the `NeoTask` and `RoomGoal` TypeScript interfaces exactly, verified against `packages/shared/src/types/neo.ts`.
- The `planning_attempts` and `goal_review_attempts` snake_case exceptions are correctly preserved in both the SQL (no AS alias) and the mapper (no conversion), matching the intentional quirk in `RoomGoal`.
- `schedulePaused` boolean coercion (`=== 1`) is correct and consistent with `GoalRepository.rowToGoal()`.
- ORDER BY clauses all include deterministic `id` tiebreakers.
- The 26 active tests (excluding `sessionGroupMessages` execution tests that don't exist) cover the happy path well.
- `NAMED_QUERY_REGISTRY` is correctly exported as a `Map` for testability.

lsm (commented):

### Re-review Summary

**P1 (fixed): `session_group_messages` table recreation**
- Table is properly added to `createTables()` in `packages/daemon/src/storage/schema/index.ts` (lines ~303-318) with the correct schema: `id INTEGER PRIMARY KEY AUTOINCREMENT`, `group_id`, `session_id`, `role`, `message_type`, `content`, `created_at`.
- Migration 41 in `packages/daemon/src/storage/schema/migrations.ts` (lines 2441-2458) properly recreates the table for existing databases, with correct idempotency guard (`tableExists` check + `CREATE TABLE IF NOT EXISTS`).
- The index `idx_sgm_group ON session_group_messages(group_id, created_at, id)` is created in both `createTables()` and `runMigration41()`, matching the query's ORDER BY clause.
- The schema in migration 41 matches the schema in `createTables()` exactly.

**P2 (fixed): `flushMicrotasks` removed**
- Confirmed: no occurrences of `flushMicrotasks` in the test file.

**P3 (acknowledged): JSON parse resilience**
- No change, consistent with codebase pattern as noted.

### New Test Coverage

The 4 new execution tests for `sessionGroupMessages.byGroup` (lines 251-342) are well-structured:
- **`SQL executes without error against the real schema`** -- runs the actual SQL against an in-memory SQLite DB with `createTables()`, confirming the table exists.
- **`returns empty array when no messages exist`** -- validates empty result set.
- **`returns camelCase column aliases`** -- inserts a message and verifies all column aliases are correct (groupId, sessionId, messageType, createdAt) and snake_case originals are absent.
- **`null sessionId is preserved as null`** -- validates nullable column handling.

These tests genuinely execute SQL against the schema, not just string-matching. This is exactly what was needed.

### No New Issues Found

- The handler file (`live-query-handlers.ts`) is clean and unchanged from the first review except that it now has a valid backing table.
- Migration 41 has proper idempotency and documentation.
- The `createTables()` additions for mission tables, `session_group_messages`, and related indexes are all consistent with the existing schema patterns.

**Recommendation: APPROVE**

lsm (commented):

This is a follow-up review verifying the three issues from the previous round were addressed. All 32 tests pass against a real in-memory SQLite database.

---

**P1 (previously): `session_group_messages` table missing**

Fixed correctly. The table is now added in two places:

1. `createTables()` in `packages/daemon/src/storage/schema/index.ts` (lines 309-319) — fresh databases get the table on first run.
2. `runMigration41()` in `packages/daemon/src/storage/schema/migrations.ts` (lines 2441-2458) — existing databases that went through migration 19 (which dropped the old table) get it recreated.

The migration sequence is sound: migration 19 runs `DROP TABLE IF EXISTS session_group_messages`, then migration 41 (called after 40) runs `CREATE TABLE IF NOT EXISTS session_group_messages`. The `tableExists()` early-exit guard makes it idempotent. Schema between `createTables()` and `runMigration41()` is consistent (same columns, same defaults, same index name `idx_sgm_group`).

**P2 (previously): `flushMicrotasks` unused import**

Confirmed removed. `grep -n "flushMicrotasks"` in the test file returns no results.

**P3 (previously, acknowledged): JSON parse resilience**

Pattern is unchanged and consistent with the existing codebase. `task-repository.ts` and `goal-repository.ts` both use bare `JSON.parse()` without try/catch, so this mapper is consistent.

---

### New Execution Tests

The four new `sessionGroupMessages.byGroup` describe-block tests go beyond structural checks and actually execute SQL against an in-memory database:

- `SQL executes without error against the real schema (table exists)` — verifies the table exists and the query runs.
- `returns empty array when no messages exist for the group` — data-level assertion.
- `returns camelCase column aliases in result rows` — inserts a row and checks all aliased column names.
- `null sessionId is preserved as null` — edge-case for nullable FK column.

These are genuine execution tests, not just SQL string inspections.

---

### No New Issues Introduced

The implementation is clean. The `NamedQuery` interface, SQL templates, row mappers, and registry are all new files with no side effects. The `live-query-handlers.ts` file exports only `NamedQuery` and `NAMED_QUERY_REGISTRY` — no RPC handler registration, no global state mutation.

One minor observation (P3, no action needed): the `mapGoalRow` function always produces `metrics: {}` (an empty object) when the DB column is `'{}'`, even though `RoomGoal.metrics` is typed as optional (`metrics?: Record<string, number>`). This is consistent with how `goal-repository.ts` already behaves, so it is not a regression introduced by this PR.

---

**Recommendation: APPROVE** — All previously raised P1 and P2 issues are resolved. The new execution tests cover the critical path. No new issues found.

PR #657, opened 2026-03-22
fix: model switch now always restarts query when queryObject exists

lsm:
The previous condition `!queryObject || !transportReady` would skip the restart
when firstMessageReceived was false, even if a query was running. This caused
model switches to not take effect for room task sessions that were waiting for
messages.

Now the condition is simplified to `!queryObject` - if a query exists, we always
restart to apply the new model/provider configuration.

Also added:
- AgentSession.restart() method that calls lifecycleManager.restart() directly
- session.restart RPC handler for manual query restart that preserves SDK session
- Updated unit tests to verify the new behavior

### DIFF
--- a/packages/daemon/src/lib/agent/agent-session.ts
+++ b/packages/daemon/src/lib/agent/agent-session.ts
@@ -723,6 +723,20 @@ export class AgentSession
 		await this.lifecycleManager.restartQuery();
 	}
 
+	/**
+	 * Force-restart the query, preserving the SDK session if possible.
+	 *
+	 * Unlike restartQuery() which defers restart if the queue isn't running,
+	 * this method always stops and restarts the query immediately.
+	 * Preserves pending messages and attempts to resume the SDK session.
+	 *
+	 * Use case: Manual restart from UI to apply model/provider changes
+	 * while preserving conversation history.
+	 */
+	async restart(): Promise<void> {
+		await this.lifecycleManager.restart();
+	}
+
 	// ============================================================================
 	// Rewind Feature (delegated to RewindHandler)
 	// ============================================================================
--- a/packages/daemon/src/lib/agent/model-switch-handler.ts
+++ b/packages/daemon/src/lib/agent/model-switch-handler.ts
@@ -52,7 +52,6 @@ export interface ModelSwitchHandlerContext {
 
 	// SDK state
 	readonly queryObject: Query | null;
-	readonly firstMessageReceived: boolean;
 }
 
 /**
@@ -226,6 +225,15 @@ export class ModelSwitchHandler {
 
 				// Update context tracker model
 				contextTracker.setModel(resolvedModel);
+
+				// Emit session.updated event so state-manager and UI know the model changed
+				// This prevents stale model display during the restart window before
+				// the restarted query emits a fresh system:init with the new model
+				await daemonHub.emit('session.updated', {
+					sessionId: session.id,
+					source: 'model-switch',
+					session: { config: session.config },
+				});
 
 				// Restart the query via lifecycle manager
 				// This spawns a new SDK subprocess with the new model configuration
--- a/packages/daemon/src/lib/agent/model-switch-handler.ts
+++ b/packages/daemon/src/lib/agent/model-switch-handler.ts
@@ -102,7 +102,6 @@ export class ModelSwitchHandler {
 			logger,
 			lifecycleManager,
 			queryObject,
-			firstMessageReceived,
 		} = this.ctx;
 
 		try {
@@ -161,9 +160,6 @@ export class ModelSwitchHandler {
 				{ channel: `session:${session.id}` }
 			);
 
-			// Check if query is running AND ProcessTransport is ready
-			const transportReady = firstMessageReceived;
-
 			// Locate the provider instance for the new model.
 			// newProvider is a required string, so detectProviderForModel always receives
 			// an explicit provider — no heuristic fallback is needed.
@@ -179,8 +175,8 @@ export class ModelSwitchHandler {
 				return { success: false, model: session.config.model, error: errMsg };
 			}
 
-			if (!queryObject || !transportReady) {
-				// Query not started yet OR transport not ready - just update config
+			if (!queryObject) {
+				// Query hasn't been created yet - just update config, it will be used when query starts
 				session.config.model = resolvedModel;
 				// newProviderInstance is guaranteed non-null here (we returned early above).
 				session.config.provider = newProviderInstance.id as Provider;
@@ -204,7 +200,12 @@ export class ModelSwitchHandler {
 					session: { config: session.config },
 				});
 			} else {
-				// Query is running - restart it to ensure system:init is regenerated with new model
+				// Query exists - always restart to apply the new model/provider.
+				// We must restart even if firstMessageReceived is false because the SDK
+				// subprocess is already running with the old model. The restart spawns a new
+				// subprocess with the updated config and resumes the conversation if the
+				// SDK session file is still valid.
+				//
 				// FIX: SDK's setModel() doesn't update the cached system:init message,
 				// causing MessageInfoDropdown to show stale model info.
 				// Restarting forces SDK to emit fresh system:init with correct model.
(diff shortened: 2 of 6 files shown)


lsm (commented):

---

### Summary

The core bug fix is correct and well-reasoned: simplifying the restart condition from `!queryObject || !transportReady` to `!queryObject` ensures that when a query (SDK subprocess) exists but `firstMessageReceived` is false, the restart still fires to apply the new model/provider. The new `session.restart` RPC handler and `AgentSession.restart()` method are clean additions with good test coverage.

All 102 tests pass (27 model-switch-handler + 75 session-handlers), and typecheck is clean.

---

### P1: Missing `daemonHub.emit('session.updated')` in the restart (else) branch

**File:** `packages/daemon/src/lib/agent/model-switch-handler.ts`, lines 202-233

This is a pre-existing issue that this PR amplifies. The if-branch (no query) emits `session.updated` to notify `state-manager.ts` and other subscribers of the config change. The else-branch (query exists, restart) does not.

Previously, the case where `queryObject` exists but `transportReady` is false went into the if-branch and got the `session.updated` event. After this PR, that case now falls into the else-branch and no longer receives it. While `state-manager.ts` may eventually get an updated cache entry when the restarted query emits a new `system:init`, there is a window where the sidebar model display could be stale.

This is pre-existing for the `queryObject` + `transportReady=true` case (which always went to else), but the behavioral change for the previously-affected case makes it worth fixing now to avoid regressions.

The fix is straightforward: add `daemonHub.emit('session.updated', ...)` to the else-branch as well, or move it after the if/else to cover both paths.

### P2: Unused `firstMessageReceived` in `ModelSwitchHandlerContext`

**File:** `packages/daemon/src/lib/agent/model-switch-handler.ts`, line 55

The `firstMessageReceived` field is still declared in the `ModelSwitchHandlerContext` interface but is no longer destructured or used anywhere in the handler. Since the linter enforces `no-unused-vars`, this will trigger warnings if the interface is consumed elsewhere and the property is never read. Consider removing it from the interface to keep the contract tight and signal that the handler no longer depends on transport readiness.

### P3: No `daemonHub.emit` event on successful `session.restart`

**File:** `packages/daemon/src/lib/rpc-handlers/session-handlers.ts`, lines 769-787

The `session.resetQuery` handler (line 755) emits a `daemonHub.emit('agent.reset', ...)` event after the reset, but the new `session.restart` handler does not emit any event on success. For consistency with the existing pattern and to allow subscribers (like StateManager or UI components) to react to manual restarts, consider emitting an event. This is not blocking but is an asymmetry worth noting for a follow-up.

---

### Verdict

The bug fix is correct and the new RPC handler is well-tested. The P1 (missing event emission) is the most actionable issue -- it is a behavioral change introduced by this PR's restructuring, even though the root cause is pre-existing. I would recommend addressing P1 before merge and deferring P2/P3 to follow-ups.

lsm (commented):
## 🤖 Review by MiniMax-M2.7-highspeed (MiniMax)

> **Model:** MiniMax-M2.7-highspeed | **Client:** NeoKai | **Provider:** MiniMax

### Summary

This PR correctly fixes the model switch bug and adds a useful manual restart capability. The core bug fix is sound: the previous condition `!queryObject || !transportReady` would skip restarting when `queryObject` existed but `firstMessageReceived` was false (the SDK subprocess was running with the old model). Simplifying to `!queryObject` ensures a restart always happens when a query exists, which is the correct behavior.

### Changes Reviewed

1. **Bug fix in `model-switch-handler.ts` (lines 178-211)**: The condition change from `!queryObject || !transportReady` to `!queryObject` is correct. The removal of `firstMessageReceived` from the destructured context is clean since it is no longer needed in the method body. The `firstMessageReceived` field remains in the `ModelSwitchHandlerContext` interface because `AgentSession` implements it, and other context consumers (like `QueryLifecycleManager`) use it.

2. **New `AgentSession.restart()` method (lines 726-738 of `agent-session.ts`)**: Thin delegation to `lifecycleManager.restart()`, consistent with the existing facade pattern. The JSDoc correctly distinguishes it from `restartQuery()`, which has guard conditions (`isRunning()` check, deferred restart during `processing` state).

3. **New `session.restart` RPC handler (lines 769-787 of `session-handlers.ts`)**: Follows the existing pattern (verify session exists, call method, return result). Error handling wraps failures as `{ success: false, error }` rather than throwing, which is slightly inconsistent with the session-not-found case that throws -- but this matches the pattern used by `session.resetQuery` (compare lines 738-761).

4. **Tests**: Both new test cases in `model-switch-handler.test.ts` (lines 387-410) correctly verify the two scenarios: (a) restart when queryObject exists but transport not ready, and (b) no restart when queryObject is null. The three handler tests in `session-handlers.test.ts` (lines 1182-1220) cover success, failure, and session-not-found cases.

### Minor Observations (P2/P3)

- **P2: No event emission on `session.restart`**: The `session.resetQuery` handler emits `agent.reset` via `daemonHub` after calling `resetQuery()`. The new `session.restart` handler does not emit any event. While the `ProcessingStateManager.setState()` does emit `session.updated` events during state transitions (idle -> processing -> idle), a dedicated restart event might be useful for the UI to distinguish a manual restart from normal state transitions. This is a suggestion for future improvement, not a blocking issue.

- **P3: `firstMessageReceived` still in interface**: The field is no longer used by `ModelSwitchHandler` but remains in `ModelSwitchHandlerContext` (line 55). This is fine since `AgentSession` implements the interface and the field is used elsewhere, but a comment noting it is there for interface completeness would be helpful.

### Verification

- All 27 model-switch-handler tests pass
- All 75 session-handlers tests pass
- Lint (oxlint): 0 warnings, 0 errors
- Typecheck: passes
- Knip: no new dead exports

### Recommendation

**APPROVE** -- The bug fix is correct and well-tested. The new `session.restart` RPC handler is a clean addition that follows existing patterns.

lsm (commented):

### Summary

Re-reviewed PR #657 after the second commit (998513e) addressed all previous review feedback. The fixes are correct and complete. All tests pass, typecheck is clean, and lint is clean.

### Verified Fixes

**P1 fix (daemonHub.emit in else-branch):** Correctly addressed. The `daemonHub.emit('session.updated', ...)` call at line 232 of `packages/daemon/src/lib/agent/model-switch-handler.ts` now fires in the else-branch (when queryObject exists) before the restart. This prevents stale model display during the restart window. The event shape matches the `DaemonEventMap['session.updated']` type exactly.

**P2 fix (remove firstMessageReceived from interface):** Correctly addressed. The `ModelSwitchHandlerContext` interface at `packages/daemon/src/lib/agent/model-switch-handler.ts:42-55` no longer includes `firstMessageReceived`. The field was not used anywhere in the handler after the condition simplification. Production code in `AgentSession` still maintains the property for its own use (query-lifecycle-manager, query-runner, rewind-handler, sdk-runtime-config all use it), which is correct.

**P3 fix (agent.restart event + DaemonEventMap entry):** Correctly addressed. The `agent.restart` event is now emitted in both success and error paths of the `session.restart` RPC handler at `packages/daemon/src/lib/rpc-handlers/session-handlers.ts:783,788`. The `DaemonEventMap` at `packages/daemon/src/lib/daemon-hub.ts:143` has the matching type definition with the same shape as the existing `agent.reset` event.

### Tests

- `model-switch-handler.test.ts`: The new test at line 387 correctly verifies that restart IS called when queryObject exists but firstMessageReceived is false. The complementary test at line 400 verifies no restart when queryObject is null. Both tests pass.
- `session-handlers.test.ts`: Three new tests cover success, failure, and session-not-found cases for `session.restart`. They follow the same pattern as existing `session.resetQuery` tests. All 75 tests in the file pass.

### Minor Note (P3)

The test helper `createContext()` at `packages/daemon/tests/unit/agent/model-switch-handler.test.ts:229` still includes `firstMessageReceived: true` in the base object literal. Since the field was removed from `ModelSwitchHandlerContext`, this is dead data. It doesn't cause a test failure (tests tsconfig is excluded from typecheck), but it should be cleaned up for clarity. Not blocking.

### Verdict

All previous issues have been properly resolved. The implementation is clean, well-tested, and follows existing patterns. Ready to merge.

PR #662, opened 2026-03-22
feat: display active agents in SpaceTaskPane with real-time status badges

lsm:
Adds a Working Agents section to SpaceTaskPane that shows session group
members for the current task. Each member displays the agent name (looked
up from agents signal via agentId), role, and a status badge — pulsing
blue for active, green check for completed, red X for failed. Groups are
sorted newest-first and the section is hidden when no groups exist for the
task. Adds 11 new unit tests covering all status badges, agent name
resolution, multi-group display, and edge cases.

### DIFF
--- a/packages/web/src/components/space/SpaceTaskPane.tsx
+++ b/packages/web/src/components/space/SpaceTaskPane.tsx
@@ -249,7 +249,7 @@ export function SpaceTaskPane({ taskId, onClose }: SpaceTaskPaneProps) {
 	}
 
 	const task = tasks.find((t) => t.id === taskId);
-	const taskGroups = taskId ? (sessionGroupsByTask.get(taskId) ?? []) : [];
+	const taskGroups = sessionGroupsByTask.get(taskId) ?? [];
 
 	if (!task) {
 		return (
--- a/packages/web/src/components/space/SpaceTaskPane.tsx
+++ b/packages/web/src/components/space/SpaceTaskPane.tsx
@@ -8,7 +8,13 @@
 import { useState } from 'preact/hooks';
 import { spaceStore } from '../../lib/space-store';
 import { cn } from '../../lib/utils';
-import type { SpaceTask, SpaceTaskStatus, SpaceTaskPriority } from '@neokai/shared';
+import type {
+	SpaceTask,
+	SpaceTaskStatus,
+	SpaceTaskPriority,
+	SpaceSessionGroup,
+	SpaceSessionGroupMember,
+} from '@neokai/shared';
 
 interface SpaceTaskPaneProps {
 	taskId: string | null;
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly implements the requested feature: displaying active agents in SpaceTaskPane with real-time status badges sourced from session group data.

### What I verified

- **Type correctness**: `SpaceSessionGroupMember['status']` is a union of `'active' | 'completed' | 'failed'` in the shared types, and `MemberStatusBadge` exhaustively handles all three variants with the fallback branch covering `'failed'`.
- **Signal reactivity**: `.value` accesses in both `SpaceTaskPane` (lines 240-241) and `WorkingAgents` (line 181) occur inside component render bodies, so `@preact/signals` integration tracks these subscriptions correctly. Changes to `tasks`, `sessionGroupsByTask`, or `agents` will trigger re-renders.
- **Cross-task isolation**: The `sessionGroupsByTask` computed signal correctly partitions groups by `taskId`, and the component filters via `sessionGroupsByTask.get(taskId)`. The test at line 405 verifies groups from other tasks are excluded.
- **Agent name resolution**: Falls back to `member.role` when no `agentId` is set or when the agent lookup returns null. Shows `(role)` parenthetical only when an agent name is resolved (line 219).
- **Tests**: 11 new tests covering no-groups, group rendering, agent name resolution, all three status badges, multi-group display, cross-task isolation, and empty-members edge case. All 32 tests pass.
- **Codebase conventions**: Uses Preact (`class` not `className`), type-only imports, no `console` calls, no `any` types. Biome formatting with tabs is consistent.

### Minor observations (P3)

1. **Redundant guard** at line 252: `taskId ? (sessionGroupsByTask.get(taskId) ?? []) : []` — the `taskId` truthiness check is unnecessary because the early return at line 243 already handles the `null` case. Not a bug, just slightly redundant.

2. **`toLocaleTimeString` in tests**: The time rendering at line 201 uses `toLocaleTimeString` which is locale-dependent. The tests don't assert on time strings so this is fine in practice, but worth noting if locale-sensitive assertions are added later.

Overall this is a well-scoped, well-tested feature addition that fits the existing patterns.

lsm (commented):

---

### Summary

The implementation correctly adds a Working Agents section to `SpaceTaskPane` using `sessionGroupsByTask` from `spaceStore`. All 32 tests pass, linting is clean, formatting is conformant, and TypeScript type-checks without errors.

---

### What was verified

- `MemberStatusBadge` and `WorkingAgents` components are correctly implemented with Preact JSX conventions (hyphenated SVG attributes, `class` not `className`).
- Signal reactivity: `spaceStore.agents.value` inside `WorkingAgents` is correctly tracked by `@preact/preset-vite`'s signal transform, so agent name updates will re-render the component.
- `sessionGroupsByTask` is already computed in `SpaceStore` (line 127 of `space-store.ts`); the PR correctly reads from this existing computed signal rather than duplicating the logic.
- The test mock correctly reimplements `sessionGroupsByTask` as a `computed` signal in the mock factory to mirror the real store's behavior.
- All 11 new tests cover the specified scenarios: no-groups, group rendering, agent name resolution, all three status badges, multi-group display, cross-task isolation, empty-members.

---

### Issues found

**P2 — Redundant ternary guard on line 252 of `SpaceTaskPane.tsx`**

```ts
const taskGroups = taskId ? (sessionGroupsByTask.get(taskId) ?? []) : [];
```

The `taskId ?` check is dead code. The function already returned early on line 243–249 when `!taskId`, so `taskId` is always truthy at line 252. The simpler form `sessionGroupsByTask.get(taskId) ?? []` is sufficient. This is harmless but misleads readers into thinking `taskId` could be null at that point.

**P2 — Spurious `capabilities` field in `makeAgent()` test helper (line 88 of `SpaceTaskPane.test.tsx`)**

```ts
capabilities: [],   // line 88 — not a field on SpaceAgent
```

`SpaceAgent` has no `capabilities` field. This phantom field is silently tolerated because the test file has `// @ts-nocheck` at the top. The bogus field has no runtime effect (the field is simply ignored), but it's inaccurate documentation and will confuse future maintainers. It should be removed from the factory helper.

**P3 — Missing sort-order test**

The `renders multiple groups for the same task` test (line 379) verifies both groups are rendered but does not assert their relative DOM order. Since `WorkingAgents` sorts groups newest-first by `createdAt`, a test asserting that the newer group appears before the older one would complete the coverage of the documented sort behavior.

**P3 — "Working Agents" section appears for groups with zero members**

When `taskGroups` has entries but all groups have empty `members` arrays, the section header "Working Agents" renders with only "No members yet" inside each group card. The test at line 400 explicitly validates this, so it is intentional. However, the feature description says the pane shows "which agents are currently working on a task" — a group with no members doesn't convey useful information. Consider whether hiding the section when all groups have empty members is a better UX. (No change required — noting this as a design consideration.)

---

### No issues found

- No console calls introduced.
- No `any` types.
- No unused imports or exports.
- Formatting is Biome-compliant (tabs, single quotes, trailing commas).
- Preact patterns followed correctly (`class`, `preact/hooks`, hyphenated SVG attributes).
- The `computed` import added in the test file is used (line 35).
- The section is correctly hidden when `taskGroups.length === 0` — no empty-state rendering leak.

lsm (commented):

**Recommendation: APPROVE**

1. **P2 fixed**: Removed the redundant `taskId ?` ternary guard in `SpaceTaskPane.tsx` line 252. Since the early return on line 243 already handles `!taskId`, `taskId` is guaranteed non-null at that point.

2. **P2 fixed**: Removed the spurious `capabilities: []` field from the `makeAgent()` test helper. The `SpaceAgent` interface has no `capabilities` field.

3. **P3 fixed**: Added the `renders groups newest-first by createdAt` test that seeds an older and newer group and asserts DOM ordering via `indexOf` comparison on queried group name elements.

**Verification:**
- All 33 tests pass (including the new sort-order test).
- The `makeAgent` helper now matches the `SpaceAgent` interface exactly (id, spaceId, name, role, description, createdAt, updatedAt).
- The `WorkingAgents` component correctly reads from `spaceStore.agents.value` and `spaceStore.sessionGroupsByTask`, sorts groups by `createdAt` descending, and resolves agent names via `agentId` lookup.
- Test coverage is thorough: all three member statuses (active/completed/failed), agent name resolution with and without agentId, multi-group rendering, cross-task isolation, empty members state, and sort order.

No remaining issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### Follow-up Review — Addressed Feedback Verified

All three items from previous feedback have been correctly addressed. Here is the detailed verification.

**P2 fix — Redundant taskId ternary guard removed**

```ts
// Before
const taskGroups = taskId ? (sessionGroupsByTask.get(taskId) ?? []) : [];
// After
const taskGroups = sessionGroupsByTask.get(taskId) ?? [];
```

The guard was dead code because `taskId` is always a non-null `string` at that point — the early return for `!taskId` on line 243 eliminates the `null` case. The fix is correct.

**P2 fix — Spurious `capabilities` field removed from `makeAgent()`**

The field is not part of `SpaceAgent` in `packages/shared/src/types/space.ts`. It has been cleanly removed. The test helper now matches the actual type definition exactly.

**P3 fix — Sort-order test added**

The new test at line 389 of `SpaceTaskPane.test.tsx` seeds two groups with distinct `createdAt` timestamps (`now - 10000` vs `now`), renders the pane, then asserts `indexOf('Newer Group') < indexOf('Older Group')` in the DOM. This directly exercises the `[...groups].sort((a, b) => b.createdAt - a.createdAt)` logic in `WorkingAgents`.

One minor observation on the CSS selector used: `.text-xs.font-medium.text-gray-300.truncate` would also match member name spans (line 216, which has an additional `capitalize` class). However, because this test seeds empty `members: []` for both groups, no member spans are rendered, so the selector is unambiguous in this specific test. This is not a correctness issue but worth knowing if the selector is reused elsewhere.

**All 33 tests pass** — verified by running `bunx vitest run src/components/space/__tests__/SpaceTaskPane.test.tsx`.

---

The implementation is complete, correct, and well-tested. No remaining issues.

lsm:
Superseded by https://github.com/lsm/neokai/pull/663

PR #663, opened 2026-03-22
feat: display active agents in SpaceTaskPane

lsm:
Add a "Working Agents" section to SpaceTaskPane that reads
sessionGroupsByTask from spaceStore and renders each group's members
with role, agent name (looked up from the agents signal by agentId),
and a status badge (active with pulse, completed with checkmark, failed
with X). Multiple groups per task are shown sorted by creation time.
The section is hidden when no groups exist for the task.

Also extends SpaceTaskPane.test.tsx with 14 new tests covering all
Working Agents rendering paths.

### DIFF
--- a/packages/web/src/components/space/SpaceTaskPane.tsx
+++ b/packages/web/src/components/space/SpaceTaskPane.tsx
@@ -12,6 +12,7 @@ import type {
 	SpaceTask,
 	SpaceTaskStatus,
 	SpaceTaskPriority,
+	SpaceAgent,
 	SpaceSessionGroup,
 	SpaceSessionGroupMember,
 } from '@neokai/shared';
@@ -97,18 +98,18 @@ function MemberStatusBadge({ status }: { status: SpaceSessionGroupMember['status
 			{status === 'completed' && (
 				<svg class="w-2.5 h-2.5" viewBox="0 0 20 20" fill="currentColor">
 					<path
-						fill-rule="evenodd"
+						fillRule="evenodd"
 						d="M16.707 5.293a1 1 0 010 1.414l-8 8a1 1 0 01-1.414 0l-4-4a1 1 0 011.414-1.414L8 12.586l7.293-7.293a1 1 0 011.414 0z"
-						clip-rule="evenodd"
+						clipRule="evenodd"
 					/>
 				</svg>
 			)}
 			{status === 'failed' && (
 				<svg class="w-2.5 h-2.5" viewBox="0 0 20 20" fill="currentColor">
 					<path
-						fill-rule="evenodd"
+						fillRule="evenodd"
 						d="M4.293 4.293a1 1 0 011.414 0L10 8.586l4.293-4.293a1 1 0 111.414 1.414L11.414 10l4.293 4.293a1 1 0 01-1.414 1.414L10 11.414l-4.293 4.293a1 1 0 01-1.414-1.414L8.586 10 4.293 5.707a1 1 0 010-1.414z"
-						clip-rule="evenodd"
+						clipRule="evenodd"
 					/>
 				</svg>
 			)}
@@ -119,11 +120,10 @@ function MemberStatusBadge({ status }: { status: SpaceSessionGroupMember['status
 
 interface WorkingAgentsProps {
 	groups: SpaceSessionGroup[];
+	agents: SpaceAgent[];
 }
 
-function WorkingAgents({ groups }: WorkingAgentsProps) {
-	const agents = spaceStore.agents.value;
-
+function WorkingAgents({ groups, agents }: WorkingAgentsProps) {
 	// Show the most recent group first; if there are multiple, show them all
 	const sortedGroups = [...groups].sort((a, b) => b.createdAt - a.createdAt);
 
@@ -253,6 +253,7 @@ function HumanInputArea({ task }: HumanInputAreaProps) {
 
 export function SpaceTaskPane({ taskId, onClose }: SpaceTaskPaneProps) {
 	const tasks = spaceStore.tasks.value;
+	const agents = spaceStore.agents.value;
 	const sessionGroupsByTask = spaceStore.sessionGroupsByTask.value;
 
 	if (!taskId) {
@@ -264,7 +265,7 @@ export function SpaceTaskPane({ taskId, onClose }: SpaceTaskPaneProps) {
 	}
 
 	const task = tasks.find((t) => t.id === taskId);
-	const taskGroups = taskId ? (sessionGroupsByTask.get(taskId) ?? []) : [];
+	const taskGroups = sessionGroupsByTask.get(taskId) ?? [];
 
 	if (!task) {
 		return (
@@ -372,7 +373,7 @@ export function SpaceTaskPane({ taskId, onClose }: SpaceTaskPaneProps) {
 				)}
 
 				{/* Working Agents */}
-				{taskGroups.length > 0 && <WorkingAgents groups={taskGroups} />}
+				{taskGroups.length > 0 && <WorkingAgents groups={taskGroups} agents={agents} />}
 
 				{/* Result */}
 				{task.result && (
--- a/packages/web/src/components/space/SpaceTaskPane.tsx
+++ b/packages/web/src/components/space/SpaceTaskPane.tsx
@@ -8,7 +8,13 @@
 import { useState } from 'preact/hooks';
 import { spaceStore } from '../../lib/space-store';
 import { cn } from '../../lib/utils';
-import type { SpaceTask, SpaceTaskStatus, SpaceTaskPriority } from '@neokai/shared';
+import type {
+	SpaceTask,
+	SpaceTaskStatus,
+	SpaceTaskPriority,
+	SpaceSessionGroup,
+	SpaceSessionGroupMember,
+} from '@neokai/shared';
 
 interface SpaceTaskPaneProps {
 	taskId: string | null;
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-executed feature addition. The implementation is clean, follows existing patterns, and has thorough test coverage. A few minor observations:

### Observations

**P3: Redundant `taskId` guard (line 267 of SpaceTaskPane.tsx)**
```ts
const taskGroups = taskId ? (sessionGroupsByTask.get(taskId) ?? []) : [];
```
At this point, `taskId` is guaranteed non-null due to the early return at line 258. The ternary is harmless but unnecessary -- `sessionGroupsByTask.get(taskId) ?? []` would suffice.

**P3: `makeAgent` in tests uses `instructions` field not in `SpaceAgent` interface**
The test helper at line 114-125 of `SpaceTaskPane.test.tsx` includes `instructions: ''`, but `SpaceAgent` does not have a direct `instructions` field. This is masked by the pre-existing `// @ts-nocheck` directive on the test file, so it compiles fine and doesn't cause test failures, but it's worth noting for accuracy.

### What looks good

- The `MEMBER_STATUS_CLASSES` record correctly covers all three values of `SpaceSessionGroupMember['status']` (`active`, `completed`, `failed`), using the union type as the key for compile-time exhaustiveness.
- The mock in the test accurately mirrors the real `sessionGroupsByTask` computed signal implementation from `space-store.ts`.
- Agent name resolution logic is sound: checks `agentId` -> looks up in agents signal -> falls back to "Task Agent" for `task-agent` role -> falls back to raw `member.role`. The secondary role label is conditionally shown only when it differs from the displayed name.
- Sort order (most recent first) is correctly implemented with `b.createdAt - a.createdAt`.
- Signal access pattern (`.value` in render body) is consistent with the rest of the codebase.
- 14 new tests cover the key paths: presence/absence of groups, group filtering by taskId, member status badges, agent name lookup, multiple groups, and empty members.

lsm (commented):

**Recommendation: APPROVE** (same-author PR — posting as COMMENT)

All 4904 tests pass, TypeScript typechecks cleanly, oxlint reports 0 errors, and Biome format passes. The implementation is largely sound. Below are the findings by severity.

---

### P2 — Redundant null guard on `taskId` (dead code)

**File:** `packages/web/src/components/space/SpaceTaskPane.tsx`, line 267

```tsx
const taskGroups = taskId ? (sessionGroupsByTask.get(taskId) ?? []) : [];
```

By this point in the function, `taskId` is guaranteed to be a non-null `string` — the `if (!taskId) { return ... }` early exit on line 258 already handles the null case. The ternary `taskId ? ... : []` is dead code. It should be simplified to:

```tsx
const taskGroups = sessionGroupsByTask.get(taskId) ?? [];
```

---

### P2 — `WorkingAgents` accesses `spaceStore.agents` directly instead of via prop

**File:** `packages/web/src/components/space/SpaceTaskPane.tsx`, line 125

```tsx
function WorkingAgents({ groups }: WorkingAgentsProps) {
    const agents = spaceStore.agents.value;
```

The parent (`SpaceTaskPane`) extracts `sessionGroupsByTask` from the store and passes `groups` down, but `WorkingAgents` reaches into the store directly to read `agents`. This creates an implicit second store dependency in a child component that otherwise looks like a pure presentational component. With `@preact/signals` and `@preact/preset-vite`, this works at runtime (the component IS tracked as a subscriber of `agents`), so there is no reactivity bug. However, it breaks the component's contract — `WorkingAgentsProps` implies the component only needs `groups`, but it silently depends on the module-level `spaceStore` singleton. Passing `agents` as an additional prop (mirroring how the parent already extracts `sessionGroupsByTask`) would make the dependency explicit, improve testability, and align with how `SpaceTaskPane`'s own store reads are structured. This is the same pattern the codebase otherwise follows (e.g., `WorkflowEditor.tsx` line 188 reads `spaceStore.agents.value` in the top-level component, not deep in a helper).

---

### P2 — Sort order is not verified by any test

**File:** `packages/web/src/components/space/__tests__/SpaceTaskPane.test.tsx`, line 334

The multi-group test asserts that both group names are present in the DOM, but does not verify their order. The implementation comment explicitly states "most recent group first", yet the test would pass even if the sort were reversed or removed entirely. A DOM order assertion (e.g., comparing the `textContent` index of each group name in the rendered container) is missing.

---

### P3 — `fill-rule`/`clip-rule` SVG attributes are kebab-case rather than camelCase

**File:** `packages/web/src/components/space/SpaceTaskPane.tsx`, lines 100, 102, 109, 111

```tsx
<path fill-rule="evenodd" clip-rule="evenodd" ... />
```

The Preact JSX transform accepts both forms, and the browser handles them correctly, so this is not a functional bug. However, the project style is inconsistent — `ImportPreviewDialog.tsx` and other newer components use the camelCase `fillRule`/`clipRule` form (which TypeScript also prefers). The kebab-case form generates a TypeScript warning that is silently suppressed here only because the file is otherwise well-typed. This is a nit, but aligning with the project convention (`fillRule`/`clipRule`) would be cleaner.

---

### P3 — Missing test: non-`task-agent` role with unresolvable `agentId`

**File:** `packages/web/src/components/space/__tests__/SpaceTaskPane.test.tsx`

There is no test for the case where a member has an `agentId` set but the referenced agent is not present in the `agents` signal, and the member's role is not `'task-agent'`. In this case `agentName` becomes `null` and the fallback `{agentName ?? member.role}` displays the raw `role` string. This edge case (stale/deleted agent reference) is plausible in production and the behavior is correct — but it is untested. A test like "falls back to member.role when agentId does not match any agent and role is not task-agent" would provide a safety net.

---

### Summary

The implementation is correct, complete, and well-tested for the happy paths. The main actionable items are: (1) remove the dead `taskId ?` ternary guard on line 267, (2) consider verifying sort order in the multi-group test, and (3) consider passing `agents` as a prop to `WorkingAgents` to make its store dependency explicit. No blocking bugs or security issues found.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review round have been addressed:

1. **Dead null guard** -- Removed. `taskId` is now used directly without the redundant ternary.
2. **Hidden module-level dependency** -- Fixed. `WorkingAgents` now receives `groups` and `agents` as explicit props (`packages/web/src/components/space/SpaceTaskPane.tsx`, lines 121-123). The parent extracts `.value` from signals (lines 256-257) and passes data down (line 376).
3. **Sort order test** -- Added at test line 345. Verifies newer groups render before older ones by checking DOM order.
4. **Unresolvable agentId fallback test** -- Added at test line 369. Verifies fallback to `member.role` when `agentId` doesn't match any known agent.

### Remaining P3 nit (not blocking)

- `packages/web/src/components/space/__tests__/SpaceTaskPane.test.tsx`, line 119: `makeAgent` still includes `instructions: ''`, which is not a field on `SpaceAgent`. The `@ts-nocheck` directive suppresses the type error, but the stray property should be removed for accuracy. This was flagged in the previous round and not addressed.

No P0/P1/P2 issues found. The implementation is clean, well-tested, and properly integrated.

lsm (commented):

**Recommendation: APPROVE** (same-author PR — posting as COMMENT)

This is a follow-up review after the worker addressed three P2 issues raised in the previous round. All three have been correctly resolved. The implementation is solid.

---

### Previous P2 Issues — Verified Resolved

**P2 fix 1 — Dead `taskId ?` ternary removed**
`packages/web/src/components/space/SpaceTaskPane.tsx`, line 268 now reads:
```ts
const taskGroups = sessionGroupsByTask.get(taskId) ?? [];
```
The dead ternary is gone. Correct.

**P2 fix 2 — `WorkingAgents` now receives `agents` as an explicit prop**
`WorkingAgentsProps` was extended to include `agents: SpaceAgent[]`. The parent extracts `spaceStore.agents.value` at line 256 and passes it at line 376. The child component no longer has a hidden module-level dependency on the store singleton. Correct.

**P2 fix 3 — Sort order test added**
The new test "renders most recent group first (sort order)" at line 345 uses a DOM-position assertion (`newerIdx < olderIdx`) to verify the descending sort. The CSS selector `.text-gray-300.text-xs.font-medium` correctly targets group name spans (not member name spans, which lack `font-medium`). The test would fail if the sort were removed or reversed. Correct.

**Bonus — Unresolvable agentId fallback test added**
The test "falls back to member.role when agentId does not match any agent and role is not task-agent" at line 369 covers the previously untested edge case. The test accurately exercises the code path. Correct.

**P3 fix — SVG attributes corrected to camelCase**
`fillRule` and `clipRule` are now used consistently on both SVG path elements (lines 101, 103, 110, 112). Correct.

---

### Remaining Observations

**P3 — `makeAgent` factory includes a spurious `instructions` field**
`packages/web/src/components/space/__tests__/SpaceTaskPane.test.tsx`, line 119:
```ts
instructions: '',
```
`SpaceAgent` has no `instructions` field. This is suppressed by the file-level `// @ts-nocheck` and has zero runtime impact. Carried over from the previous round — still a nit worth cleaning up for accuracy when `@ts-nocheck` is eventually lifted.

**P3 — Secondary role label shown redundantly for task-agent with an unresolvable agentId**
In `WorkingAgents`, if a member has `agentId` set but the referenced agent is not in the `agents` signal, and `member.role === 'task-agent'`, the name resolution produces `agentName = 'Task Agent'`. The secondary role label condition at line 172 (`agentName && agentName !== member.role`) evaluates to `'Task Agent' !== 'task-agent'` = true, so both "Task Agent" (primary) and "task-agent" (secondary) are rendered. This is a display redundancy for a rare edge case. The test added in this PR (line 369) uses `role: 'security-auditor'` so it does not exercise this combination.

This is not a regression from the PR — the logic was introduced in the original commit — and the impact is cosmetic only. Not a blocker.

---

### Summary

All P2 issues from the previous review round have been cleanly resolved. The implementation is correct and well-tested (35 tests pass). CI shows lint, type check, and format passing. The two remaining notes are both P3 nits with no functional impact. This PR is ready to merge.

PR #664, opened 2026-03-22
test: add online test for room tick via job queue

lsm:
Adds packages/daemon/tests/online/room/room-tick-job.test.ts with 8 tests
covering the full room.tick job queue lifecycle:

- Tick job is enqueued when room runtime starts
- Job transitions through processing and reaches completed
- Handler re-schedules next tick after completion (30s delay)
- Dedup: at most one pending tick per room at any time
- Pause cancels pending tick jobs
- Resume enqueues a fresh immediate tick after pause
- Stop cancels pending ticks and prevents further scheduling
- Restart resumes tick scheduling with a fresh runtime

Also updates validate-online-test-matrix.sh to track room-tick-job.test.ts
in the ROOM_FILES registry (room tests are commented out in CI for resource
reasons but tracked for completeness).

### DIFF
--- a/packages/daemon/tests/online/room/room-tick-job.test.ts
+++ b/packages/daemon/tests/online/room/room-tick-job.test.ts
@@ -218,9 +218,10 @@ describe('room.tick via job queue (online)', () => {
 		const daemonCtx = getDaemonCtx(daemon);
 		const roomId = await createRoom(daemon, 'tick-dedup-test');
 
-		// Immediately after room creation: exactly 1 pending/processing tick exists.
-		// scheduleTick(delay=0) fires synchronously inside start(), so the job is
-		// visible right away — no race between the RPC returning and the enqueue.
+		// Wait for the tick to appear: room.created handlers fire via queueMicrotask
+		// after the RPC returns, so the enqueue is not guaranteed to have happened
+		// the moment createRoom() resolves. Poll until it does, then assert count=1.
+		await waitForTickJob(daemonCtx, roomId, ['pending', 'processing']);
 		const atStartup = listTickJobs(daemonCtx, roomId, ['pending', 'processing']);
 		expect(atStartup.length).toBe(1);
 
--- a/packages/daemon/tests/online/room/room-tick-job.test.ts
+++ b/packages/daemon/tests/online/room/room-tick-job.test.ts
@@ -11,7 +11,8 @@
  * - Pause cancels pending tick jobs
  * - Resume enqueues a fresh immediate tick
  * - Stopping a room cancels pending ticks and prevents further scheduling
- * - Restarting a room resumes tick scheduling
+ * - Restarting a room resumes tick scheduling (new job post-restart is identified)
+ * - Dead tick job (maxRetries=0) does not produce a ghost pending tick
  *
  * The room.tick handler calls runtime.tick() which is a no-op for an empty
  * room (no active task groups, no goals). No real AI calls are made.
@@ -73,7 +74,7 @@ function listTickJobs(daemonCtx: DaemonAppContext, roomId: string, statuses: Job
  * Poll the job queue until at least one room.tick job for the given room
  * exists with one of the specified statuses, or until the timeout expires.
  *
- * Returns the first matching job, or throws on timeout.
+ * Throws on timeout with a full diagnostic snapshot.
  */
 async function waitForTickJob(
 	daemonCtx: DaemonAppContext,
@@ -94,6 +95,29 @@ async function waitForTickJob(
 	throw new Error(
 		`Timeout waiting for room.tick job for room "${roomId}" with status [${statuses.join(',')}] after ${timeoutMs}ms. ` +
 			`All room.tick jobs: ${JSON.stringify(all.map((j) => ({ id: j.id, payload: j.payload, status: j.status, runAt: j.runAt })))}`
+	);
+}
+
+/**
+ * Poll until no room.tick jobs with the given statuses exist for the room,
+ * or throw on timeout.
+ */
+async function waitForNoTickJobs(
+	daemonCtx: DaemonAppContext,
+	roomId: string,
+	statuses: JobStatus[],
+	timeoutMs: number = JOB_WAIT_TIMEOUT_MS
+): Promise<void> {
+	const deadline = Date.now() + timeoutMs;
+	while (Date.now() < deadline) {
+		const jobs = listTickJobs(daemonCtx, roomId, statuses);
+		if (jobs.length === 0) return;
+		await new Promise<void>((resolve) => setTimeout(resolve, POLL_INTERVAL_MS));
+	}
+	const remaining = listTickJobs(daemonCtx, roomId, statuses);
+	throw new Error(
+		`Timeout waiting for room.tick jobs [${statuses.join(',')}] to clear for room "${roomId}" after ${timeoutMs}ms. ` +
+			`Remaining: ${JSON.stringify(remaining.map((j) => ({ id: j.id, status: j.status })))}`
 	);
 }
 
@@ -163,6 +187,8 @@ describe('room.tick via job queue (online)', () => {
 		expect(completed.status).toBe('completed');
 		expect(completed.completedAt).not.toBeNull();
 		expect((completed.payload as { roomId: string }).roomId).toBe(roomId);
+		// maxRetries=0 by design: tick failures are terminal (no silent retry loop).
+		expect(completed.maxRetries).toBe(0);
 	}, 15_000);
 
 	// -------------------------------------------------------------------------
@@ -173,13 +199,11 @@ describe('room.tick via job queue (online)', () => {
 		const daemonCtx = getDaemonCtx(daemon);
 		const roomId = await createRoom(daemon, 'tick-rescheduled-test');
 
-		// Wait for the initial tick to complete.
-		const firstCompleted = await waitForTickJob(daemonCtx, roomId, ['completed']);
-		expect(firstCompleted).toBeDefined();
+		// Wait for the initial tick to complete (discard result — waitForTickJob throws on timeout).
+		await waitForTickJob(daemonCtx, roomId, ['completed']);
 
 		// The handler's finally block enqueues the next tick with DEFAULT_TICK_INTERVAL_MS (30s).
 		const next = await waitForTickJob(daemonCtx, roomId, ['pending']);
-		expect(next).toBeDefined();
 		expect((next.payload as { roomId: string }).roomId).toBe(roomId);
 
 		// Next tick should be scheduled ~30 s in the future (at least 10 s to avoid flakiness).
@@ -187,24 +211,26 @@ describe('room.tick via job queue (online)', () => {
 	}, 15_000);
 
 	// -------------------------------------------------------------------------
-	// Test 4: Dedup — at most one pending tick per room
+	// Test 4: Dedup — exactly one pending tick per room
 	// -------------------------------------------------------------------------
 
-	test('dedup: at most one pending room.tick job per room', async () => {
+	test('dedup: exactly one pending room.tick job per room', async () => {
 		const daemonCtx = getDaemonCtx(daemon);
 		const roomId = await createRoom(daemon, 'tick-dedup-test');
 
-		// Check immediately after room creation: at most 1 pending/processing tick.
+		// Immediately after room creation: exactly 1 pending/processing tick exists.
+		// scheduleTick(delay=0) fires synchronously inside start(), so the job is
+		// visible right away — no race between the RPC returning and the enqueue.
 		const atStartup = listTickJobs(daemonCtx, roomId, ['pending', 'processing']);
-		expect(atStartup.length).toBeLessThanOrEqual(1);
+		expect(atStartup.length).toBe(1);
 
 		// Wait for the initial tick to complete and the next one to be enqueued.
 		await waitForTickJob(daemonCtx, roomId, ['completed']);
 		await waitForTickJob(daemonCtx, roomId, ['pending']);
 
-		// Still at most 1 pending tick for this room.
+		// After self-scheduling: still exactly 1 pending tick — dedup prevents extras.
 		const afterReschedule = listTickJobs(daemonCtx, roomId, ['pending']);
-		expect(afterReschedule.length).toBeLessThanOrEqual(1);
+		expect(afterReschedule.length).toBe(1);
 	}, 15_000);
 
 	// -------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

Well-structured online test file that covers the full room.tick job queue lifecycle. Each test gets a fresh daemon instance via `beforeEach`, providing proper isolation. The helper functions (`waitForTickJob`, `listTickJobs`) are well-designed with good timeout diagnostics.

### Observations

**P3 - Assertion weakness in resume/restart tests (tests 6 and 8)**

In `packages/daemon/tests/online/room/room-tick-job.test.ts` lines 257-263 and 314-320, `waitForTickJob` is called with `['pending', 'processing', 'completed']` after resume/restart. This could theoretically match the old completed job from before the pause/stop. It works correctly in practice because `listJobs` uses `ORDER BY created_at DESC` so the newest job is returned first, but narrowing the statuses to `['pending', 'processing']` would make the intent clearer and the test more robust against future ordering changes.

**P3 - Dedup test (test 4) is passive**

Test 4 verifies dedup by checking counts after natural scheduling, but doesn't actively attempt to create a duplicate (e.g., calling `enqueueRoomTick` directly or triggering a second event that would schedule a tick). This makes it more of a "verify single tick exists" test than a true dedup stress test. Acceptable for coverage purposes but worth noting.

**Positive notes:**

- Good use of unique room names per test to aid debugging
- Timeout error messages include full job queue state for diagnostics
- Proper guard against spawned daemon mode (test requires in-process access)
- 30s timeout buffer check in test 3 uses a 10s lower bound, which is reasonable flakiness protection
- Matrix validation script updated correctly

lsm (commented):

**Recommendation: APPROVE with minor suggestions**

The test file is well-structured, follows established project patterns (mirrors `github-poll-job.test.ts` closely), is correctly scoped to in-process mode only, and the 8 test cases cover the stated requirements. The code is reviewable and ready to merge. The issues below are mostly quality improvements rather than correctness blockers.

---

### P1 — Weak dedup assertion (Test 4, line 198)

```ts
// Test 4 — packages/daemon/tests/online/room/room-tick-job.test.ts:198
const atStartup = listTickJobs(daemonCtx, roomId, ['pending', 'processing']);
expect(atStartup.length).toBeLessThanOrEqual(1); // passes even when 0
```

This assertion would pass if `scheduleTick()` failed silently and no job was ever enqueued. The dedup test should use `toBe(1)` here, since the runtime's `start()` fires `scheduleTick(delay=0)` synchronously before the RPC returns — a job must exist immediately after room creation. The current `≤1` form doesn't distinguish between "one job, as required" and "no jobs at all".

---

### P1 — Test 8 doesn't assert the post-restart job is actually new

```ts
// packages/daemon/tests/online/room/room-tick-job.test.ts:314-320
const freshTick = await waitForTickJob(daemonCtx, roomId, ['pending', 'processing', 'completed']);
expect(freshTick).toBeDefined();
expect((freshTick.payload as { roomId: string }).roomId).toBe(roomId);
```

`listJobs` is ordered by `created_at DESC`, so the newest job comes first — this is fine for correctness. However, the assertions here don't verify the job was created *after* the restart, only that a job with this roomId exists. In practice this is unlikely to be an issue (the pre-restart job has an older `created_at`), but asserting `freshTick.createdAt > beforeRestartTimestamp` would make the intent explicit and guard against `listJobs` ordering changes.

---

### P2 — Redundant `expect(firstCompleted).toBeDefined()` in Test 3

```ts
// packages/daemon/tests/online/room/room-tick-job.test.ts:177-178
const firstCompleted = await waitForTickJob(daemonCtx, roomId, ['completed']);
expect(firstCompleted).toBeDefined(); // waitForTickJob already throws on timeout
```

`waitForTickJob` has return type `Promise<Job>` (non-optional) and throws if the timeout expires. The `toBeDefined()` assertion is dead code. Same pattern appears at line 184 (`expect(next).toBeDefined()`) and line 261 (`expect(freshTick).toBeDefined()`). These are harmless but add noise.

---

### P2 — Test 5/6 use fixed-duration sleep after pause instead of polling

```ts
// packages/daemon/tests/online/room/room-tick-job.test.ts:228
await new Promise<void>((resolve) => setTimeout(resolve, 200));
const afterPause = listTickJobs(daemonCtx, roomId, ['pending']);
expect(afterPause.length).toBe(0);
```

`cancelPendingTickJobs` in `pause()` is synchronous — the RPC call completes only after the pending tick is deleted. So 200 ms is not needed for the cancellation itself; the RPC returns with the pending tick already gone. However, if a tick is mid-flight (processing) at the moment of the pause call, the handler's `finally` block will re-check runtime state and correctly skip re-enqueuing. The sleep window is the grace period for that case. This is fine on fast machines, but a tighter pattern would be to poll with `waitForTickJob` configured to expect failure (`listTickJobs` returning empty within a short deadline), rather than a fixed 200 ms sleep.

---

### P2 — No test for the `failed`/`dead` recovery path

The task description mentions "recovery for stopped rooms". The tests cover `stop` and `restart` of a *healthy* room, but there is no test for what happens when the tick handler itself throws (which leads to a `dead` status, since `maxRetries: 0` in `enqueueRoomTick`). This is a gap given the recovery objective. A test that verifies the loop self-terminates on `dead` status (i.e., no automatic retry, no ghost pending job) would close this gap.

---

### P2 — Test 7 stop assertion relies on two separate fixed-duration sleeps

```ts
// packages/daemon/tests/online/room/room-tick-job.test.ts:280-290
await new Promise<void>((resolve) => setTimeout(resolve, 500));
const afterStop = listTickJobs(daemonCtx, roomId, ['pending']);
expect(afterStop.length).toBe(0);
await new Promise<void>((resolve) => setTimeout(resolve, 300));
const noNewPending = listTickJobs(daemonCtx, roomId, ['pending']);
expect(noNewPending.length).toBe(0);
```

This is the most timing-dependent test in the file. On a slow CI worker the 500 ms may not be enough for an in-flight tick to complete (the processor poll interval is 1 s). Consider waiting explicitly for any processing job to become completed/dead before asserting pending=0, to avoid a flaky false pass where the processing job finishes *after* the assertion window.

---

### P3 — Inconsistency with existing helper pattern

`waitForGitHubPollJob` (github-poll-job.test.ts) returns `undefined` on timeout while `waitForTickJob` throws. The throw-on-timeout approach is strictly better (no silent failures in test code), but the deviation makes the two helpers harder to swap and could confuse future authors maintaining both files. Neither approach is wrong; noting for awareness.

---

### P3 — `firstCompleted` unused variable in Test 3

```ts
const firstCompleted = await waitForTickJob(daemonCtx, roomId, ['completed']);
expect(firstCompleted).toBeDefined(); // only defensive check
```

The variable is only used for the redundant `toBeDefined()` check. Replacing with `await waitForTickJob(…)` (discarding the return value) would make the intent clearer.

---

### Summary

The core logic is sound: each test creates an isolated room with a fresh daemon, uses proper in-process introspection to check job queue state, and the  polling pattern is robust. The test correctly covers dedup, pause/resume, stop/restart lifecycle. The P1 issues (weak dedup assertion, missing post-restart identity check) are worth addressing before merge, but they don't represent correctness failures in the current implementation — they represent cases where the tests would fail to catch future regressions.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review after addressing previous feedback. All prior issues have been resolved cleanly:

- **Test 4 dedup**: `toBeLessThanOrEqual(1)` changed to `toBe(1)` -- correctly asserts exactly one pending job exists, no longer passes silently when zero jobs are enqueued.
- **Test 8 restart**: `beforeRestartTimestamp` added with `expect(freshTick.createdAt).toBeGreaterThanOrEqual(beforeRestartTimestamp)` -- properly verifies the post-restart job is genuinely new.
- **Redundant `.toBeDefined()` calls**: Removed across tests 3, 6, and 8, since `waitForTickJob` already throws on timeout.
- **Tests 5/6 sleep removal**: The 200ms sleep after pause is gone; the assertion runs immediately after the RPC returns, which is correct since `cancelPendingTickJobs` is synchronous within the pause handler.
- **Tests 7/8 fixed sleep replaced**: `waitForNoTickJobs(['processing'])` now polls for in-flight jobs to drain instead of using a fixed 500ms sleep. The remaining 300ms sleep in test 7 is an intentional safety window to confirm no ghost jobs appear, which is acceptable.
- **Test 9 (dead job)**: New test verifies that a dead tick job (maxRetries=0) does not produce ghost pending ticks. It inserts a dead job directly into the DB and confirms no pending tick materializes and the dead job remains in dead status. This closes the coverage gap for the failure/dead recovery path.
- **Test 6 narrowed statuses**: `waitForTickJob` after resume now uses `['pending', 'processing']` instead of including `'completed'`, avoiding ambiguity with pre-pause completed jobs.

**P3 -- Test 9 uses raw SQL insert to simulate dead job**

In `packages/daemon/tests/online/room/room-tick-job.test.ts` lines 365-374, the test inserts a dead job directly via `rawDb.prepare(...)`. This is the pragmatic approach since there is no public API to force a job into dead status, but it does couple the test to the `job_queue` table schema. If the schema changes, this test will break silently (the insert may succeed with wrong columns). This is acceptable for an online test -- just worth noting for future maintainers.

**P3 -- Dedup test (test 4) remains passive**

Test 4 still verifies dedup by observing natural scheduling behavior rather than actively attempting to create duplicates (e.g., calling `enqueueRoomTick` directly). This is fine for coverage purposes but remains a passive check rather than a stress test.

### Summary

All prior P1 and P2 issues have been addressed. The test file is well-structured with 9 comprehensive test cases covering job enqueue, lifecycle transitions, dedup, pause/resume, stop/restart, and dead job handling. Helper functions provide good diagnostic output on timeout. The matrix validation script is correctly updated. No blocking issues remain.

lsm (commented):

**Recommendation: APPROVE** — the previously identified P1/P2 issues are resolved. One remaining P2 flake risk is documented below.

---

## Summary of Changes vs. Previous Review

All previously reported P1/P2 issues were addressed:

- **Test 4 dedup**: `toBeLessThanOrEqual(1)` → `toBe(1)`. Fixed.
- **Test 8 restart**: `beforeRestartTimestamp` guard added to verify the post-restart job is genuinely new (`createdAt >= beforeRestartTimestamp`). The `createdAt` field is populated in `JobQueueRepository.enqueue()` via `Date.now()` and mapped in `rowToJob()`, so this assertion is valid. Fixed.
- **Tests 5/6**: Removed unnecessary 200ms sleep after pause. Fixed.
- **Tests 7/8**: Replaced fixed 500ms sleep with `waitForNoTickJobs(['processing'])`. Fixed.
- **Redundant `expect(x).toBeDefined()`**: Removed. Fixed.
- **Test 9 (new)**: Dead job verification via direct DB insert. Well-structured.

---

## Remaining Concern

### P2 — Test 4: synchrony assumption is incorrect at line 224

**File:** `packages/daemon/tests/online/room/room-tick-job.test.ts` (line 221–225)

The comment reads:

> `scheduleTick(delay=0) fires synchronously inside start(), so the job is visible right away — no race between the RPC returning and the enqueue.`

This is inaccurate. Tracing the actual code path:

1. `room.create` RPC handler calls `daemonHub.emit('room.created', ...).catch()` — **fire and forget, not awaited**.
2. The `room.created` handlers (both `RoomRuntimeService.subscribeToEvents` which calls `createOrGetRuntime → runtime.start() → scheduleTick()`, and `seedRoomTick` in `rpc-handlers/index.ts`) are invoked via `TypedHub.dispatchLocally`, which uses `queueMicrotask` internally.
3. The RPC handler returns `{ room }` **before** the `queueMicrotask` callbacks execute — the microtask queue drains after the current synchronous frame.

So when `await createRoom()` resolves, the tick enqueue has not necessarily happened yet, and the bare `listTickJobs` call on line 224 (no retry/polling) could theoretically return an empty array, causing a flaky assertion.

In practice this has not caused failures likely because: (a) the in-process WebSocket round-trip gives the microtasks enough time to run before the test continues, and (b) both the `RoomRuntimeService` subscriber and `seedRoomTick` fire in the same microtask batch. But this is timing-dependent, not guaranteed.

The fix is to use the existing `waitForTickJob` helper for the `atStartup` check as well, instead of asserting synchronously:

```ts
// Replace lines 224-225 with:
await waitForTickJob(daemonCtx, roomId, ['pending', 'processing']);
const atStartup = listTickJobs(daemonCtx, roomId, ['pending', 'processing']);
expect(atStartup.length).toBe(1);
```

This preserves the exact-count assertion (dedup correctness) while removing the race.

---

## Other Notes

- **Test 9 fixed sleep**: The 400ms sleep at line 378 to confirm no ghost tick appears is acceptable here since there is no event to poll on (we're verifying absence). A fixed wait is the standard pattern for negative-existence checks.
- **Matrix registration**: `room-tick-job.test.ts` is correctly added to `ROOM_FILES` in the validation script (commented-out CI section, consistent with other room tests).
- **Overall quality**: Test structure, diagnostics on timeout, and the `waitForNoTickJobs` helper are solid. The test suite provides meaningful coverage of the job-queue tick lifecycle.

lsm (commented):

**Recommendation: APPROVE**

This is a final follow-up review after the remaining P2 race condition in Test 4 was fixed.

### Fix verification

The change in Test 4 (dedup) is correct. Previously, `listTickJobs()` was called synchronously immediately after `createRoom()` resolved, but `room.created` handlers fire via `queueMicrotask`, meaning the enqueue could race with the assertion. The fix replaces this with:

1. `await waitForTickJob(daemonCtx, roomId, ['pending', 'processing'])` -- polls until the job exists
2. `listTickJobs(daemonCtx, roomId, ['pending', 'processing'])` -- then immediately checks count is exactly 1

This is the right pattern: wait for the async side-effect to land, then assert the invariant (exactly one job). The comment at lines 221-223 clearly explains why the poll is necessary.

### Overall assessment

All 9 tests use consistent, race-free patterns:
- Polling helpers (`waitForTickJob`, `waitForNoTickJobs`) for async expectations
- Synchronous `listTickJobs` only after a prior poll has confirmed the expected state
- Appropriate timeouts and diagnostic messages on failure

No remaining issues found.

lsm (commented):

**Recommendation: APPROVE**

This is the final follow-up review verifying the P2 race condition fix in Test 4.

---

## Fix Verification

The previously reported P2 (race condition in Test 4) is correctly resolved.

**The old code (racy):**
The bare `listTickJobs()` call executed immediately after `createRoom()` resolved — before the `room.created` event handlers had a chance to fire and enqueue the tick job. This was racy because `daemonHub.emit('room.created', ...)` in `room-handlers.ts` (line 86) is fire-and-forget (`.catch()` only, not awaited), and `TypedHub.dispatchLocally` invokes subscribers via `queueMicrotask`, which means the `createOrGetRuntime → runtime.start() → scheduleTick()` chain runs in a later microtask batch — after the RPC response is already on its way back to the test.

**The new code (correct):**

The `waitForTickJob` poll at line 224 blocks until the enqueue has actually happened, eliminating the race. The synchronous `listTickJobs` count assertion that follows is then safe — the job is already in the DB, so the count is stable.

**Comment accuracy:** The comment on lines 221-223 now correctly explains the queueMicrotask-based async delivery rather than the previous incorrect claim about synchronous firing. This is accurate per `TypedHub.dispatchLocally` in `packages/shared/src/message-hub/typed-hub.ts` (line 197-218).

---

## No Other Issues Found

- **Test structure**: All 8 tests remain well-structured with clear isolation (fresh daemon per test via `beforeEach`/`afterEach`).
- **Negative-existence sleeps**: The 300ms sleep in Test 7 (line 308) and 400ms sleep in Test 9 (line 379) are both for verifying absence of new jobs. These are appropriate since there is no event to poll on; a fixed wait is the standard pattern here.
- **Matrix registration**: `room-tick-job.test.ts` is correctly in `ROOM_FILES` in `scripts/validate-online-test-matrix.sh` (line 57), consistent with the other room tests that are tracked but not run in CI due to resource constraints.
- **Handler logic alignment**: The test's assumptions about `maxRetries=0`, `DEFAULT_TICK_INTERVAL_MS=30s` re-schedule delay, and state-based skip in the finally block all match the implementation in `packages/daemon/src/lib/job-handlers/room-tick.handler.ts`.
- **pause/stop synchrony**: Tests 5/6 rely on `cancelPendingTickJobs` being synchronous inside `pause()` — this is confirmed correct in `room-runtime.ts` line 431. No sleep needed after pause, which the tests correctly reflect.

The fix is minimal and targeted. The P2 race condition is fully resolved.

lsm:
Superseded by https://github.com/lsm/neokai/pull/670

PR #668, opened 2026-03-22

lsm:
Adds job-queue-crash-recovery.test.ts with 6 integration scenarios that
verify all job types survive a daemon crash and are correctly reclaimed
when the daemon restarts against the same file-backed SQLite database:

1. session.title_generation: stale processing job reclaimed and processed
2. github.poll: self-scheduling chain resumes after restart
3. room.tick: stale tick reclaimed and exits cleanly (no live runtime)
4. job_queue.cleanup: stale cleanup job runs and self-schedules next run
5. Eager reclamation: stale jobs are NOT stuck as processing after restart
6. File-backed DB: jobs written by daemon-1 are visible to daemon-2

Key implementation details:
- Each test uses a separate file-backed SQLite DB under tmp/test-crash-recovery/
- Stale processing rows are inserted directly via raw SQL to simulate crashes
- Tests verify eager reclaimStale() called synchronously in start(), not the
  60s polling cadence
- afterAll cleans up the tmp/test-crash-recovery/ directory
- CI matrix updated with new features-3 shard (mock_sdk: true)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -182,6 +182,9 @@ jobs:
               tests/online/features/github-poll-job.test.ts
               tests/online/features/message-persistence.test.ts
             mock_sdk: true
+          - module: features-3
+            test_path: tests/online/features/job-queue-crash-recovery.test.ts
+            mock_sdk: true
           - module: git
             test_path: tests/online/git
             timeout: 10
--- a/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+++ b/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
@@ -31,7 +31,7 @@
  *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
  */
 
-import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import { afterAll, afterEach, beforeEach, describe, expect, test } from 'bun:test';
 import * as fs from 'fs';
 import * as path from 'path';
 import { createDaemonApp, type DaemonAppContext } from '../../../src/app';
@@ -270,6 +270,16 @@ afterEach(async () => {
 		}
 	} catch {
 		// Directory may not exist yet or already deleted — not a failure.
+	}
+});
+
+afterAll(() => {
+	// Remove the TEST_DB_DIR directory itself once all tests have finished.
+	// Per-test afterEach handles individual files; afterAll handles the container.
+	try {
+		fs.rmSync(TEST_DB_DIR, { recursive: true, force: true });
+	} catch {
+		// Best-effort — a leftover empty directory is not a test failure.
 	}
 });
 
@@ -518,22 +528,38 @@ describe('Job queue crash/restart recovery (integration)', () => {
 		expect(completed).toBeDefined();
 		expect(completed!.status).toBe('completed');
 
-		// After completion, the cleanup handler self-schedules the next run.
-		// Accept 'pending' or 'processing' — the processor may have picked it up
-		// already by the time waitForJob polls.
-		const nextCleanup = await waitForJob(
-			h2.ctx,
-			JOB_QUEUE_CLEANUP,
-			['pending', 'processing'],
-			JOB_WAIT_TIMEOUT_MS
-		);
-		expect(nextCleanup).toBeDefined();
-
-		// The next cleanup job must be a *different* job (not the stale one that
-		// just ran) and must be scheduled ~24 h in the future.
-		expect(nextCleanup!.id).not.toBe(staleCleanupId);
-		const minExpected = Date.now() + 23 * 60 * 60 * 1000; // 23 h from now
-		expect(nextCleanup!.runAt).toBeGreaterThan(minExpected);
+		// After the stale cleanup job completes, the cleanup handler self-schedules
+		// a follow-up run ~24 h from now.
+		//
+		// IMPORTANT: app.ts also enqueues an immediate startup cleanup job when
+		// daemon-2 starts (because the stale job was 'processing', not 'pending',
+		// so the startup dedup check sees no pending jobs and adds one with
+		// runAt ≈ now).  Two cleanup jobs therefore race:
+		//   • stale (reclaimed, runAt ≈ now-6min)
+		//   • startup (runAt ≈ now)
+		//
+		// The cleanup handler only self-schedules if `pending.length === 0`.
+		// Whichever job runs last will find the other gone and will self-schedule
+		// the 24 h follow-up.  So in all interleavings exactly one 24 h job is
+		// eventually enqueued — but we must wait for it explicitly rather than
+		// grabbing the first pending/processing job (which may be the immediate
+		// startup job, not the 24 h one).
+		const minFutureRunAt = Date.now() + 23 * 60 * 60 * 1000; // 23 h from now
+		const scheduledFollowUp = await (async () => {
+			const deadline = Date.now() + JOB_WAIT_TIMEOUT_MS;
+			while (Date.now() < deadline) {
+				const jobs = h2.ctx.jobQueue.listJobs({
+					queue: JOB_QUEUE_CLEANUP,
+					status: ['pending', 'processing', 'completed'],
+				});
+				const match = jobs.find((j) => j.id !== staleCleanupId && j.runAt > minFutureRunAt);
+				if (match) return match;
+				await new Promise<void>((r) => setTimeout(r, POLL_INTERVAL_MS));
+			}
+			return undefined;
+		})();
+		expect(scheduledFollowUp).toBeDefined();
+		expect(scheduledFollowUp!.runAt).toBeGreaterThan(minFutureRunAt);
 
 		await stopDaemon(h2);
 	}, 30_000);
--- a/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+++ b/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
@@ -31,7 +31,7 @@
  *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
  */
 
-import { afterAll, afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
 import * as fs from 'fs';
 import * as path from 'path';
 import { createDaemonApp, type DaemonAppContext } from '../../../src/app';
@@ -67,24 +67,47 @@ const GITHUB_TEST_ENV: Record<string, string> = {
 // Lifecycle helpers
 // ---------------------------------------------------------------------------
 
-/** Track contexts created during a test so afterEach can clean them up. */
-let openContexts: DaemonAppContext[] = [];
+/**
+ * Per-daemon env snapshot captured by startDaemon.  Restored by stopDaemon
+ * so env vars do not leak between tests (and between daemon-1 / daemon-2
+ * within the same test).
+ */
+interface DaemonHandle {
+	ctx: DaemonAppContext;
+	/** DB file path — deleted per-test in afterEach. */
+	dbPath: string;
+	/** Keys whose original values were saved before mutation. */
+	savedEnv: Record<string, string | undefined>;
+}
+
+/** All handles created during the current test. */
+let openHandles: DaemonHandle[] = [];
 
 /**
  * Create a daemon with a file-backed SQLite DB.
  *
- * Each call re-uses the provided `dbPath` so the second daemon instance
- * inherits all rows written by the first one.
+ * Captures and saves all env vars before mutating them.  Call stopDaemon()
+ * (or rely on afterEach) to restore them.
  */
 async function startDaemon(
 	dbPath: string,
 	extraEnv: Record<string, string> = {}
-): Promise<DaemonAppContext> {
-	// Derive a temporary workspace from the DB path (each run gets its own dir
-	// so concurrent tests do not collide on the workspace root).
+): Promise<DaemonHandle> {
 	const workspace = path.dirname(dbPath);
 
-	// Apply env vars — reset after the test in afterEach.
+	// Keys that startDaemon will mutate — capture originals before any change.
+	const envKeys = [
+		'NEOKAI_WORKSPACE_PATH',
+		'NODE_ENV',
+		'NEOKAI_SDK_STARTUP_TIMEOUT_MS',
+		...Object.keys(extraEnv),
+	];
+	const savedEnv: Record<string, string | undefined> = {};
+	for (const k of envKeys) {
+		savedEnv[k] = process.env[k];
+	}
+
+	// Apply env vars.  Restored in stopDaemon / afterEach via restoreEnv().
 	process.env.NEOKAI_WORKSPACE_PATH = workspace;
 	process.env.NODE_ENV = 'test';
 	if (!process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS) {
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Recommendation: APPROVE

This is a well-structured, thorough integration test file that validates job queue crash/restart recovery across all four job types. The tests are clear in their intent, follow a consistent pattern, and use appropriate techniques to simulate crashes.

### Strengths

- **Correct crash simulation**: Inserting raw SQL rows with `startedAt` 6 minutes in the past is the right approach — it faithfully models what a crashed daemon leaves behind without requiring actual process kills.
- **Eager reclamation verified**: Test 5 correctly validates that `reclaimStale()` runs synchronously in `start()` before the 60s polling cadence, which is the key behavioral guarantee.
- **File-backed DB**: Each test uses a unique DB path with `Date.now()` suffix, avoiding collisions between parallel runs.
- **Chain continuity**: The GitHub poll test verifies both that the crashed job is recovered AND that the previously scheduled follow-up job survives across restart.
- **Cleanup**: `afterEach` drains all open contexts, `afterAll` removes the temp directory.

### Minor observations (P3)

1. **Environment variable pollution between tests** (`packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 82-87): `startDaemon` sets `process.env.NEOKAI_WORKSPACE_PATH` and `GITHUB_TOKEN` / `GITHUB_POLLING_INTERVAL` but never restores them afterward. Since all tests in this file run sequentially within a single `describe` block and each test creates its own daemon with its own config, this is unlikely to cause issues in practice. However, if a non-GitHub test runs after a GitHub test, the lingering `GITHUB_TOKEN` env var could cause the second daemon to initialize `gitHubService` unexpectedly. This is a theoretical concern, not a practical one given the current test ordering and isolation via separate DB files.

2. **`getJobById` scans up to 1000 rows** (`packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, line 146): This is fine for test workloads but slightly inelegant. The `JobQueueRepository.getJob(id)` method exists and could be used directly via `ctx.jobQueue.getJob(jobId)` — though this depends on whether `getJob` is exposed on the `DaemonAppContext.jobQueue` property. Not a real issue.

3. **No `timeout` field in CI matrix entry** (`.github/workflows/main.yml`, line 185): Other feature shards like `features-1` and `features-2` also lack explicit timeout, so this is consistent. The default timeout should be sufficient for 6 tests with 30s individual timeouts.

All acceptance criteria are met: four crash-recovery scenarios pass, jobs transition correctly, no lost jobs, self-scheduling chains resume, eager reclamation verified, and file-backed DB is used throughout.

lsm (commented):

Overall the implementation is solid in its happy-path coverage, but there are several correctness and reliability issues that should be addressed before merge.

---

### P0 — Environment variables are never restored after each test

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` — `startDaemon()` (lines 88–94)

`startDaemon()` mutates `process.env` for every test (`NEOKAI_WORKSPACE_PATH`, extra keys from `extraEnv` such as `GITHUB_POLLING_INTERVAL` and `GITHUB_TOKEN`), but no code ever restores those values. The `afterEach` only stops lingering daemon contexts; it never saves and restores the env snapshot. Compare the existing `createInProcessDaemonServer()` helper in `daemon-server.ts` (lines 526–530, 686–693), which explicitly captures originals and restores them in `waitForExit()`.

Consequences:
- Tests that set `GITHUB_POLLING_INTERVAL=600` and `GITHUB_TOKEN=...` will leak those values into subsequent tests in the same Bun worker process. If test 2 (github.poll) runs before test 3 (room.tick), daemon-3 unexpectedly tries to start a GitHub service, changing its job-queue behaviour and potentially causing flaky failures.
- `NEOKAI_WORKSPACE_PATH` leaks across tests, so daemon-2 in test N will use the workspace directory of test N-1 if `startDaemon()` is called before the env var is overwritten again.

Fix: capture originals before mutation inside `startDaemon()` and restore them in the returned context or in `stopDaemon()`.

---

### P1 — Test 2 (github.poll recovery) has a race-condition window in the triggerPoll stub

**File:** lines 303–307 and 344–345

`startDaemon(daemon1)` starts the job processor immediately (via `createDaemonApp()`). The `triggerPoll` stub is only applied *after* `startDaemon()` returns. Because `jobProcessor.start()` fires a synchronous `tick()` (line 45 of `job-queue-processor.ts`) and the initial github.poll job is already pending, there is a real window where the processor dequeues and invokes the real `triggerPoll` before the stub is installed. The same race exists in the existing `github-poll-job.test.ts` but that test's comments acknowledge it and note it is safe because no repos are registered. The new test must carry the same explanation, or stub before `gitHubService.start()` if the architecture allows it.

---

### P1 — Test 1 (session.title_generation) asserts reclamation synchronously with no await — timing assumption may not hold

**File:** lines 269–273

```ts
const reclaimed = daemon2.jobQueue.listJobs({
    queue: SESSION_TITLE_GENERATION,
    status: ['pending'],
});
expect(reclaimed.some((j) => j.id === staleJobId)).toBe(true);
```

`createDaemonApp()` resolves only after `jobProcessor.start()` is called (line 394 of `app.ts`). `start()` calls `reclaimStale()` synchronously, so the job is `pending` when `createDaemonApp()` returns. However, `start()` also immediately fires an async `tick()` in the background. By the time the next line executes, the processor may have already moved the job from `pending` to `processing` or even `failed`. The assertion will then fail with a cryptic "expected false to be true".

This is likely to be intermittent. The fix is to check for `['pending', 'processing', 'completed', 'failed', 'dead']` (i.e., "not stuck as old-stale processing") as test 5 correctly does, or to add a stabilisation step.

---

### P1 — Test 4 (cleanup) may fail when the self-scheduled follow-on job is already processing by the time waitForJob checks for it

**File:** lines 468–473

After the stale cleanup job completes, the cleanup handler self-schedules a new pending job. The test then calls `waitForJob(daemon2, JOB_QUEUE_CLEANUP, ['pending'])` and asserts `nextCleanup.runAt > Date.now() + 23h`. If the processor is fast and the self-scheduled job is already in `processing` by the time `waitForJob` polls, `waitForJob` will wait the full 10 s timeout and return `undefined`, causing the test to fail.

The assertion should also accept `processing` or completed state from the self-scheduled job, checking its `runAt` via `getJobById` rather than `waitForJob` with a `['pending']`-only filter.

---

### P2 — startDaemon does not use the shared daemon-server helper, bypassing env-restoration logic

**File:** lines 79–105

The existing `createDaemonServer()` helper in `packages/daemon/tests/helpers/daemon-server.ts` already handles workspace creation, env mutation/restoration, dev-proxy lifecycle, and `daemonContext` exposure. The new test reimplements this partially, omitting key pieces (env restoration above, worktree base dir setup). This is not just a code-duplication concern — the omissions are correctness gaps.

The test cannot directly reuse `createDaemonServer()` as-is because it needs to share a specific file-backed DB path across two daemon instances. The helper should be extended with a `dbPath` override option (it already sets `config.dbPath` on line 562 of `daemon-server.ts`), which would allow these tests to delegate env handling to the established helper. Alternatively, the test should at minimum copy the env save/restore pattern from that helper.

---

### P2 — Test 2 crash scenario is muddled: the inserted stale job is a third job, not the follow-up being processed

**File:** lines 330–339

After the initial github.poll job completes, a follow-up pending job is already in the queue. The test then inserts an *additional* stale processing job. The comment says "the follow-up arrived earlier and was being processed when daemon died", but the follow-up job is actually still `pending` — a brand-new unrelated stale job is inserted instead. The test passes because the assertions are loose enough, but the described crash scenario does not match the DB state being constructed. This misleads future readers about what is actually being verified.

---

### P3 — Stale comment says env vars are reset in afterEach but they are not

**File:** line 87

The comment `// Apply env vars — reset after the test in afterEach.` is incorrect. `afterEach` does not restore env vars. This creates a false sense of safety and should be removed or corrected.

---

### P3 — afterAll cleanup cannot run if the process crashes; per-test cleanup would be safer

**File:** lines 224–231

`TEST_DB_DIR` is cleaned in `afterAll`. If a test fails hard mid-run, directories accumulate under `tmp/test-crash-recovery/`. Each test creates its own timestamped DB, so repeated CI failures can accumulate disk usage. A `beforeEach` or per-test `afterEach` cleanup scoped to the individual DB path would bound the growth without affecting other tests.

---

### Summary

The core crash-simulation approach (direct SQL insert of stale `processing` rows, file-backed DB shared between two daemon instances) is sound and well-chosen. The eager `reclaimStale()` call in `start()` is correctly verified. The main problems are: (1) env vars are never restored, risking cross-test contamination (P0); (2) two assertions check for `pending` state synchronously when the job may already have advanced past it (P1, tests 1 and 4); (3) the `triggerPoll` stub has a startup race with no comment acknowledging it (P1, test 2). These are all fixable without restructuring the tests.

lsm (commented):

### Recommendation: APPROVE

The fix commit (`40d8fbee`) properly addresses the P3 feedback from the previous review.

#### P3 #1 — Environment variable pollution: **Fixed**

The new `DaemonHandle` interface with `savedEnv` snapshot and `restoreEnv()` cleanly solves env var leakage. Env vars are captured before mutation in `startDaemon()` and restored in both `stopDaemon()` (via `finally` block) and `afterEach` (for handles not explicitly stopped). This is a robust solution.

#### P3 #2 — `getJobById` scans 1000 rows: **Acknowledged, unchanged**

The stale comment was removed. The `listJobs` approach with `limit: 1000` is retained, which is perfectly fine for test workloads. No action needed.

#### Additional improvements in the fix commit

- **Per-test DB cleanup** (`afterEach` replaces `afterAll`): Better isolation — each test cleans its own `.db` / `.db-wal` / `.db-shm` files rather than bulk-deleting the directory at the end. Bounds disk usage even if the process crashes mid-run.
- **More robust assertions**: The session title recovery test (test 1) now checks "not stuck as stale processing" rather than asserting a specific intermediate status, which eliminates a race condition where the processor could dequeue the job before the assertion runs.
- **Simplified eager reclamation loop** (test 5): The inner polling loop no longer has a special case for "fresh processing" — it simply waits for terminal states, which is cleaner and less error-prone.
- **Unused import removed**: `afterAll` dropped from the import since it's no longer used.

No issues found. The test file is clean, well-documented, and correctly validates crash/restart recovery for all four job types.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (same reviewer as PR author — posting as COMMENT per policy)

---

This re-review covers the fix commit `40d8fbee`. Most of the original issues have been properly addressed, but one P1 remains with a real intermittent-failure risk.

---

### RESOLVED — P0: Env var restoration

The `DaemonHandle` interface with `savedEnv`, `restoreEnv()`, and the `finally` block in `stopDaemon()` are all correctly implemented. `afterEach` also calls `restoreEnv` on any lingering handles. This fully closes the original P0 concern.

---

### RESOLVED — P1 Test 1: Sync assertion race

Replaced with `isStillStale` check (lines 317–321). Correctly handles all states the job may be in by the time the check runs. Resolved.

---

### RESOLVED — P1 Test 2: `triggerPoll` stub race

Comment at lines 350–356 now explains why the pre-stub window is safe (no repos registered, errors caught internally). Matches the established pattern in `github-poll-job.test.ts`. Resolved.

---

### NOT FULLY RESOLVED — P1 Test 4: Cleanup self-schedule timing

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` lines 482–538

The `waitForJob` fix (now accepting `['pending', 'processing']`) only partially solves the original issue. There is a deeper race that the fix does not address.

When daemon-2 starts in this test, `app.ts` lines 382–390 check for pending cleanup jobs. The stale cleanup job is `processing` (not `pending`), so the check sees no pending cleanup jobs and enqueues a new immediate cleanup job (`runAt ≈ Date.now()`). The processor then runs both the stale cleanup job and this startup cleanup job concurrently.

The cleanup handler (in `cleanup.handler.ts` lines 13–17) only self-schedules the next 24-hour run `if (pending.length === 0)`. When the stale cleanup job finishes and checks for pending jobs, the startup cleanup job may still be `pending` — so `pending.length > 0` and no follow-up is self-scheduled.

As a result, `waitForJob(h2.ctx, JOB_QUEUE_CLEANUP, ['pending', 'processing'])` at line 524 returns the startup cleanup job (which has `runAt ≈ Date.now()`), and then `expect(nextCleanup!.runAt).toBeGreaterThan(Date.now() + 23h)` at line 536 fails.

This is a real intermittent failure path. The test would pass if the startup cleanup job is already `processing` when the stale job self-schedules, but fail if it is still `pending`.

To fix robustly: after `waitForJobById(staleCleanupId, ['completed'])`, explicitly look for a cleanup job with `runAt > now + 23h` using a dedicated poll (not `waitForJob` for any pending/processing job), and explain in comments that the startup cleanup job is expected to also run but has `runAt ≈ now`.

---

### RESOLVED — P2 Test 2 crash scenario comment

The updated comment at lines 382–388 now correctly describes the inserted stale job as a brand-new third job simulating an unscheduled webhook/manual-triggered poll. The mismatch between the described scenario and the DB state is resolved.

---

### RESOLVED — P3 misleading comment

Removed from `startDaemon`. Resolved.

---

### RESOLVED — P3 per-test DB file cleanup

`afterEach` now deletes `*.db`, `*.db-wal`, and `*.db-shm` files. Resolved.

---

### NEW P3 — `TEST_DB_DIR` directory never removed

**File:** lines 244–274

The original `afterAll` that deleted the entire `TEST_DB_DIR` was removed. Individual test files are now cleaned up in `afterEach`, but the directory itself (`tmp/test-crash-recovery/`) is never removed. This accumulates an empty directory across CI runs. Minor, but worth noting.

---

### Summary

5 of 6 original issues are properly resolved. The remaining P1 issue in Test 4 is a real intermittent failure path caused by the concurrent startup-enqueued cleanup job interfering with the stale job's self-schedule guard in `cleanup.handler.ts`. The `runAt > 23h` assertion will intermittently fail when the startup cleanup job is still `pending` when the stale job executes the self-schedule check.

lsm (commented):

### Final re-review (3rd pass)

The latest commit (`fbad57dc`) addresses the two remaining issues from the previous review. Both fixes are correct and well-implemented.

**P1 fix — Test 4 cleanup self-schedule race (lines 531-562):**
The previous `waitForJob` call could return the immediate startup cleanup job instead of the intended 24h follow-up. The new implementation uses a custom polling loop that filters by `j.id !== staleCleanupId && j.runAt > minFutureRunAt`, which correctly identifies only the 24h self-scheduled follow-up regardless of which cleanup job runs first. The detailed comment explaining the two-job race (stale reclaimed vs. startup dedup) is thorough and accurate.

**P3 fix — TEST_DB_DIR cleanup (lines 276-284):**
`afterAll` now removes the container directory with `fs.rmSync(TEST_DB_DIR, { recursive: true, force: true })`. Best-effort error handling is appropriate — a leftover empty dir should not fail tests.

lsm (commented):

**Recommendation: APPROVE** — both previously raised issues are correctly resolved.

---

### P1 — Test 4 cleanup self-schedule race: RESOLVED

The fix at lines 547-562 of `job-queue-crash-recovery.test.ts` is correct and robust.

The dedicated poll uses:
```ts
const match = jobs.find((j) => j.id !== staleCleanupId && j.runAt > minFutureRunAt);
```
where `minFutureRunAt = Date.now() + 23 * 60 * 60 * 1000`.

This correctly filters out both jobs that could cause false positives:
- The stale job itself (`j.id !== staleCleanupId`)
- The startup-enqueued immediate job that app.ts adds (its `runAt ≈ now`, which is far less than `now + 23h`)

The comment at lines 535-546 accurately explains the two-job interleaving: daemon-2's startup dedup check at app.ts lines 383-388 only looks for `status: 'pending'` jobs, so it finds none (the stale job is `'processing'`) and enqueues an immediate cleanup job. Both jobs race, and `cleanup.handler.ts` line 14 (`pending.length === 0` guard) ensures exactly one 24h follow-up is eventually scheduled regardless of interleave order. The test assertion matches this invariant precisely.

### P3 — TEST_DB_DIR cleanup: RESOLVED

`afterAll` at lines 276-284 calls `fs.rmSync(TEST_DB_DIR, { recursive: true, force: true })`. Combined with the `afterEach` per-file cleanup (lines 263-273), disk usage is bounded at both levels.

---

No new issues introduced. The implementation is correct.

lsm (commented):

**Recommendation: APPROVE**

### What this commit does

The latest commit (bf338cf0) adds `job-queue-crash-recovery.test.ts` to the `FEATURES_FILES` array in `scripts/validate-online-test-matrix.sh`, fixing a CI validation failure that would occur because the test file exists on disk but was not listed in the validator.

### Verification

The fix is correct and complete:

1. **File exists on disk**: `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` is present.

2. **`FEATURES_FILES` array now matches disk exactly**: All 5 files in the `features/` directory (`auto-title.test.ts`, `github-poll-job.test.ts`, `job-queue-crash-recovery.test.ts`, `message-delivery-mode-queue.test.ts`, `message-persistence.test.ts`) are listed in the validator. The check is bidirectional — every file on disk must be in the array AND every entry in the array must exist on disk — both now pass.

3. **CI matrix already has the entry**: `features-3` was added to `main.yml`'s `include:` block (commit cfcf217b) pointing to `tests/online/features/job-queue-crash-recovery.test.ts` with `mock_sdk: true`. Note that `features-3` is not in the base `module:` list, but GitHub Actions will still create a standalone job for it because `include` entries with no matching base matrix combination are added as new jobs. This is consistent with the CI running it.

4. **Validation script logic**: `check_split_module "features"` at line 112 of the script verifies the `features/` directory. The missing entry was the direct cause of the CI failure, and the fix addresses it precisely.

No issues found. The fix is minimal and correct.

lsm:
Superseded by https://github.com/lsm/neokai/pull/676

PR #669, opened 2026-03-22
test: online test for room tick via job queue

lsm:
## Summary

- Add `packages/daemon/tests/online/room/room-tick-job.test.ts` — 10 tests covering end-to-end `room.tick` job queue mechanics without real API calls (empty room tick is a no-op)
- Register the new file in `scripts/validate-online-test-matrix.sh` ROOM_FILES so the validator stays green

> **CI note:** All `room/*` shards are intentionally commented out in `.github/workflows/main.yml` (see lines 204–220) due to resource usage. This file is tracked by the validator but does **not** run automatically in CI. Run locally with `NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/room/room-tick-job.test.ts`, or re-enable the shard explicitly when needed.

## Tests

| Test | What it verifies |
|------|-----------------|
| job enqueued on runtime start | `runtime.start()` calls `enqueueRoomTick(roomId, jobQueue, 0)` |
| no duplicate pending tick jobs | at most one pending `room.tick` per room at any time |
| pause cancels pending ticks | `room.runtime.pause` → `cancelPendingTickJobs` removes all pending jobs |
| resume enqueues fresh tick | `room.runtime.resume` → `scheduleTick()` → new pending job appears |
| stop cancels pending ticks | `room.runtime.stop` cancels pending jobs |
| restart resumes scheduling | `room.runtime.start` creates fresh runtime and enqueues a new tick |
| job processes and re-schedules | tick transitions `pending → processing → completed`, then re-enqueues with ~30s delay |
| tick handler skips re-scheduling when stopped | handler returns `{skipped: true}` when runtime is gone; no new tick enqueued |
| each room gets its own independent tick job | pausing room1 does not affect room2's pending tick |
| daemon restart recovery | `initializeExistingRooms` re-enqueues ticks for all DB rooms on daemon2 startup |

## Test plan

- [x] All 10 tests pass locally with `NEOKAI_USE_DEV_PROXY=1`
- [x] `bash scripts/validate-online-test-matrix.sh` passes

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -200,6 +200,12 @@ jobs:
           - module: rewind
             test_path: tests/online/rewind
             mock_sdk: true
+          # room-tick-job uses Dev Proxy only (no real API calls) — enabled independently
+          # of the broader room test cohort which requires real credentials.
+          - module: room-tick-job
+            test_path: tests/online/room/room-tick-job.test.ts
+            mock_sdk: true
+            timeout: 10
           # Room tests temporarily disabled during refactoring
           # - module: room-1
           #   test_path: >-
--- a/packages/daemon/tests/online/room/room-tick-job.test.ts
+++ b/packages/daemon/tests/online/room/room-tick-job.test.ts
@@ -49,7 +49,12 @@ const JOB_WAIT_TIMEOUT_MS = 10_000;
 /** Polling interval for job-status checks. */
 const POLL_INTERVAL_MS = 50;
 
-/** How long to wait after a control operation before asserting no tick exists. */
+/**
+ * How long to wait after a control operation before asserting no tick exists.
+ * 300 ms is sufficient: the job processor's poll cycle is 1 s, so any pending
+ * enqueue triggered synchronously by pause/stop/resume will complete well
+ * within this window without waiting for a full processor cycle.
+ */
 const SETTLE_MS = 300;
 
 // ---------------------------------------------------------------------------
@@ -111,10 +116,17 @@ async function waitForTickJob(
 function acceleratePendingTick(daemonCtx: DaemonAppContext, roomId: string): void {
 	const pending = listTickJobs(daemonCtx, roomId, ['pending']);
 	for (const job of pending) {
+		// Direct DB delete bypasses any higher-level state machine, but
+		// JobQueueRepository.deleteJob() is a raw DELETE with no side effects —
+		// there is no state machine to bypass. The tradeoff is acceptable here
+		// because this is test-only code whose sole purpose is to accelerate the
+		// runAt timestamp, and deleteJob() is part of the public repository API.
 		daemonCtx.jobQueue.deleteJob(job.id);
 	}
 	// enqueueRoomTick with delay=0 schedules an immediate tick via the same
-	// production code path used by RoomRuntime.start() / resume().
+	// production code path used by RoomRuntime.start() / resume(). If
+	// RoomRuntime.scheduleTick() fires between the delete and re-enqueue,
+	// enqueueRoomTick's dedup guard absorbs the duplicate.
 	enqueueRoomTick(roomId, daemonCtx.jobQueue, 0);
 }
 
@@ -383,13 +395,15 @@ describe('room.tick via job queue (online)', () => {
 		// Start daemon2 using createDaemonApp() directly with the preserved DB.
 		// This exercises initializeExistingRooms() which reads rooms from the DB
 		// and re-enqueues tick jobs for each one.
-		const originalWorkspacePath = process.env.NEOKAI_WORKSPACE_PATH;
-		process.env.NEOKAI_WORKSPACE_PATH = workspacePath;
-
 		let daemonCtx2: DaemonAppContext | null = null;
 		let hub2: MessageHub | null = null;
 		let transport2: WebSocketClientTransport | null = null;
+		// originalWorkspacePath is captured inside the try block so the finally
+		// restoration is always paired with the mutation, even on sync failures.
+		let originalWorkspacePath: string | undefined;
 		try {
+			originalWorkspacePath = process.env.NEOKAI_WORKSPACE_PATH;
+			process.env.NEOKAI_WORKSPACE_PATH = workspacePath;
 			const config = getConfig();
 			config.port = 0;
 			config.dbPath = dbPath;
@@ -420,7 +434,7 @@ describe('room.tick via job queue (online)', () => {
 			// Verify the runtime is actively running in daemon2.
 			const stateResult = (await hub2.request('room.runtime.state', {
 				roomId,
-			})) as { state: string | null };
+			})) as { state: string };
 			expect(stateResult.state).toBe('running');
 		} finally {
 			// Restore environment.
--- a/packages/daemon/tests/online/room/room-tick-job.test.ts
+++ b/packages/daemon/tests/online/room/room-tick-job.test.ts
@@ -10,6 +10,10 @@
  * - Stopping a room cancels pending ticks and prevents further scheduling
  * - Restarting a room enqueues a new tick job
  * - Job processes and re-schedules (self-scheduling chain)
+ * - Tick handler skips re-scheduling when runtime is stopped
+ * - Each room gets its own independent tick job
+ * - Daemon restart: existing rooms get tick jobs re-enqueued via
+ *   initializeExistingRooms → recoverRoomRuntime → runtime.start()
  *
  * These tests exercise the persistent job queue mechanics (enqueueRoomTick,
  * cancelPendingTickJobs, createRoomTickHandler) introduced in Task 4.1–4.3.
@@ -23,9 +27,13 @@
  *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/room/room-tick-job.test.ts
  */
 
+import path from 'path';
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import { WebSocketClientTransport, MessageHub } from '@neokai/shared';
 import type { DaemonServerContext } from '../../helpers/daemon-server';
 import { createDaemonServer } from '../../helpers/daemon-server';
+import { createDaemonApp } from '../../../src/app';
+import { getConfig } from '../../../src/config';
 import type { DaemonAppContext } from '../../../src/app';
 import { ROOM_TICK } from '../../../src/lib/job-queue-constants';
 import { enqueueRoomTick } from '../../../src/lib/job-handlers/room-tick.handler';
@@ -287,8 +295,12 @@ describe('room.tick via job queue (online)', () => {
 	// -------------------------------------------------------------------------
 
 	test('tick handler skips re-scheduling when runtime is stopped', async () => {
-		// Stop the runtime first — this removes the runtime from the runtimes map
-		// (via runtimes.delete in stopRuntime) and cancels pending ticks.
+		// Wait for the runtime to be ready before stopping — room.created fires
+		// via queueMicrotask so the runtime may not be in the map yet when
+		// room.create resolves.
+		await waitForTickJob(daemonCtx, roomId, ['pending', 'processing', 'completed']);
+
+		// Stop the runtime — removes it from the runtimes map and cancels pending ticks.
 		await daemon.messageHub.request('room.runtime.stop', { roomId });
 		await new Promise<void>((resolve) => setTimeout(resolve, SETTLE_MS));
 		expect(listTickJobs(daemonCtx, roomId, ['pending']).length).toBe(0);
@@ -342,4 +354,103 @@ describe('room.tick via job queue (online)', () => {
 		expect(pendingRoom1.length).toBe(0);
 		expect(pendingRoom2.length).toBeGreaterThanOrEqual(1);
 	}, 20_000);
+
+	// -------------------------------------------------------------------------
+	// Daemon restart: existing rooms get tick jobs re-enqueued
+	// -------------------------------------------------------------------------
+
+	test('daemon restart: existing rooms get tick jobs re-enqueued via initializeExistingRooms', async () => {
+		// Confirm the initial tick appears so we know the room is fully set up.
+		const firstTick = await waitForTickJob(daemonCtx, roomId, [
+			'pending',
+			'processing',
+			'completed',
+		]);
+		expect(firstTick).toBeDefined();
+
+		// Preserve the DB path before shutdown.
+		const dbPath = daemonCtx.db.getDatabasePath();
+		const workspacePath = path.dirname(dbPath);
+
+		// Shut down daemon1 cleanly but WITHOUT deleting the workspace so the DB
+		// is preserved for daemon2. We call daemonCtx.cleanup() directly instead
+		// of daemon.waitForExit() (which would rm -rf the workspace).
+		daemon.kill('SIGTERM'); // no-op for in-process
+		await daemonCtx.cleanup();
+		// Prevent afterEach from double-calling cleanup.
+		daemon = null as unknown as DaemonServerContext;
+
+		// Start daemon2 using createDaemonApp() directly with the preserved DB.
+		// This exercises initializeExistingRooms() which reads rooms from the DB
+		// and re-enqueues tick jobs for each one.
+		const originalWorkspacePath = process.env.NEOKAI_WORKSPACE_PATH;
+		process.env.NEOKAI_WORKSPACE_PATH = workspacePath;
+
+		let daemonCtx2: DaemonAppContext | null = null;
+		let hub2: MessageHub | null = null;
+		let transport2: WebSocketClientTransport | null = null;
+		try {
+			const config = getConfig();
+			config.port = 0;
+			config.dbPath = dbPath;
+
+			daemonCtx2 = await createDaemonApp({ config, verbose: false, standalone: false });
+
+			// Connect a client so the daemon is ready.
+			const actualPort = daemonCtx2.server.port;
+			transport2 = new WebSocketClientTransport({
+				url: `ws://127.0.0.1:${actualPort}/ws`,
+				autoReconnect: false,
+			});
+			hub2 = new MessageHub({ defaultSessionId: 'global' });
+			hub2.registerTransport(transport2);
+			await transport2.initialize();
+
+			// initializeExistingRooms calls runtime.start() for every room in the DB,
+			// which calls scheduleTick() → enqueueRoomTick(roomId, jobQueue, 0).
+			const tickJob = await waitForTickJob(daemonCtx2, roomId, [
+				'pending',
+				'processing',
+				'completed',
+			]);
+			expect(tickJob).toBeDefined();
+			expect(tickJob!.queue).toBe(ROOM_TICK);
+			expect((tickJob!.payload as { roomId: string }).roomId).toBe(roomId);
+
+			// Verify the runtime is actively running in daemon2.
+			const stateResult = (await hub2.request('room.runtime.state', {
+				roomId,
+			})) as { state: string | null };
+			expect(stateResult.state).toBe('running');
+		} finally {
+			// Restore environment.
+			if (originalWorkspacePath === undefined) {
+				delete process.env.NEOKAI_WORKSPACE_PATH;
+			} else {
+				process.env.NEOKAI_WORKSPACE_PATH = originalWorkspacePath;
+			}
+
+			// Tear down daemon2: cleanup hub first (rejects pending calls), then transport.
+			if (hub2) {
+				try {
+					hub2.cleanup();
+				} catch {
+					// Already cleaned up
+				}
+			}
+			if (transport2) {
+				try {
+					await transport2.close();
+				} catch {
+					// Already closed
+				}
+			}
+			if (daemonCtx2) {
+				await daemonCtx2.cleanup();
+			}
+
+			// Remove the shared workspace now that both daemons are done.
+			await Bun.$`rm -rf ${workspacePath}`.quiet();
+		}
+	}, 30_000);
 });
--- a/packages/daemon/tests/online/room/room-tick-job.test.ts
+++ b/packages/daemon/tests/online/room/room-tick-job.test.ts
@@ -184,6 +184,9 @@ describe('room.tick via job queue (online)', () => {
 
 		// Attempt to enqueue additional ticks via the RPC resume path (which calls
 		// enqueueRoomTick internally). The dedup guard must absorb these.
+		// Sequence ends on resume so the runtime is running and exactly one pending
+		// tick should exist — asserting toBe(1) (not just <=1) proves the dedup
+		// guard absorbed the duplicates while still leaving one job standing.
 		await daemon.messageHub.request('room.runtime.pause', { roomId });
 		await daemon.messageHub.request('room.runtime.resume', { roomId });
 		await daemon.messageHub.request('room.runtime.pause', { roomId });
@@ -193,7 +196,9 @@ describe('room.tick via job queue (online)', () => {
 		await new Promise<void>((resolve) => setTimeout(resolve, SETTLE_MS));
 
 		const pendingJobs = listTickJobs(daemonCtx, roomId, ['pending']);
-		expect(pendingJobs.length).toBeLessThanOrEqual(1);
+		// Exactly 1: dedup absorbed the extra enqueue calls; the runtime is running
+		// so the one surviving tick was not cancelled.
+		expect(pendingJobs.length).toBe(1);
 	}, 15_000);
 
 	// -------------------------------------------------------------------------
@@ -408,6 +413,12 @@ describe('room.tick via job queue (online)', () => {
 			config.port = 0;
 			config.dbPath = dbPath;
 
+			// Note: createDaemonApp inherits the current process environment, which
+			// includes ANTHROPIC_BASE_URL and ANTHROPIC_API_KEY set by daemon1's
+			// createInProcessDaemonServer (Dev Proxy routing). Daemon2 will silently
+			// use the proxy for any LLM calls it makes. This is harmless here because
+			// the room tick handler makes no LLM calls, but future changes that add
+			// LLM paths to the tick handler should be aware of this implicit coupling.
 			daemonCtx2 = await createDaemonApp({ config, verbose: false, standalone: false });
 
 			// Connect a client so the daemon is ready.


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured online test file that thoroughly covers the room tick job queue lifecycle. The tests verify the key behaviors: initial enqueue on room creation, job completion and re-scheduling, dedup invariant, pause/resume/stop/restart lifecycle interactions with the job queue.

### Strengths

- **Good coverage of the specified scenarios**: All 7 tests map directly to the task requirements (enqueue, complete+reschedule, dedup, pause, resume, stop, restart).
- **Proper daemon lifecycle management**: Each test gets a fresh daemon via `beforeEach`/`afterEach`, avoiding cross-test contamination.
- **Appropriate use of in-process daemon**: Accessing `daemonContext.jobQueue` directly for assertions is the right approach for testing job queue internals, and the `getDaemonCtx` helper correctly guards against spawned mode.
- **Validate-matrix script updated**: The new file is properly registered in `ROOM_FILES` in `scripts/validate-online-test-matrix.sh`.

### Minor observations (P3)

1. **Timing sensitivity in the "pause cancels pending ticks" test** (line 186): The test calls `waitForRoomTickJob(daemonCtx, roomId, ['pending'])` after room creation. Since the initial tick is enqueued with delay=0 (`scheduleTick` calls `enqueueRoomTick(roomId, jobQueue, 0)`), the job processor could pick it up before the first poll iteration. In practice this is fine because after the initial tick completes, it re-schedules with a 30s delay, creating a pending job that will definitely be caught. But this means the test actually waits for the *re-scheduled* pending job rather than the *initial* one. The 10-second timeout is generous enough, and the behavior is correct -- just worth noting that the test implicitly depends on the self-rescheduling path, not just the initial enqueue.

2. **Repeated payload casting pattern** (lines 72, 85, 126, etc.): The `(j.payload as { roomId?: string }).roomId` cast is repeated in every helper and inline check. A tiny helper like `getRoomIdFromPayload(job: Job): string | undefined` could reduce repetition, but this is purely cosmetic and doesn't affect correctness.

3. **No explicit room cleanup**: Rooms created by `createRoom` are never explicitly deleted. This is acceptable since each test gets a fresh daemon instance that is torn down in `afterEach`, but it could be confusing to someone reading the test in isolation.

### Verification

- Imports resolve correctly: `ROOM_TICK` from `job-queue-constants`, `Job`/`JobStatus` from `job-queue-repository`, `DaemonAppContext` from `app.ts`, `DaemonServerContext` from helpers.
- The RPC names (`room.create`, `room.runtime.pause`, `room.runtime.resume`, `room.runtime.stop`, `room.runtime.start`, `room.runtime.state`) all match the registered handlers in `packages/daemon/src/lib/rpc-handlers/room-handlers.ts`.
- The `daemonContext` property is correctly exposed by `createInProcessDaemonServer` (line 626 of `daemon-server.ts`).
- The `listJobs` API usage matches the `JobQueueRepository` interface (queue filter + status array filter).

No bugs, no security issues, no missing error handling. Clean test code that exercises the intended functionality end-to-end.

lsm (commented):

The test file is well-structured and covers the right lifecycle behaviors, but there is one timing bug that will cause intermittent failures, one missing test case from the original task spec, and a few minor issues worth addressing.

---

### P0 — Race condition in test 1 (intermittent failure)

**File:** `packages/daemon/tests/online/room/room-tick-job.test.ts`, lines 116–130

The first test queries the job queue **synchronously** immediately after `createRoom()` resolves:

```ts
const roomId = await createRoom(daemon);
// No await / poll here — direct synchronous read:
const jobs = daemonCtx.jobQueue.listJobs({ ... });
```

The `room.create` RPC handler emits `room.created` **without awaiting**:

```ts
daemonHub.emit('room.created', { ... }).catch(...)   // fire-and-forget
return { room };                                     // returns immediately
```

`TypedHub.emit` dispatches handlers through `queueMicrotask` inside `dispatchLocally`. Even though the whole chain is `await`ed internally, the `room.create` handler returns **before** `emit` resolves, so the RPC response reaches the test before `RoomRuntimeService.createOrGetRuntime` has been called, and no tick job exists yet.

This makes the assertion `roomJobs.length >= 1` a race. On a slow or heavily loaded machine the list will be empty. The fix is to use `waitForRoomTickJob` (the same helper every other test uses) instead of the synchronous query:

```ts
const job = await waitForRoomTickJob(daemonCtx, roomId, ['pending', 'processing', 'completed']);
expect(job).toBeDefined();
expect(job!.queue).toBe(ROOM_TICK);
expect(job!.payload).toMatchObject({ roomId });
```

---

### P1 — Missing "recovery after daemon restart" test case

The original task asked for a test verifying **recovery for stopped rooms** (i.e., the scenario where the daemon is fully restarted and `initializeExistingRooms` re-enqueues a tick for every existing room). The 7 tests only exercise in-process lifecycle transitions (`room.runtime.stop` / `room.runtime.start` RPC). None restarts the daemon process to verify that existing rooms get tick jobs re-enqueued on startup.

This is a meaningful gap: the recovery path (`initializeExistingRooms → recoverRoomRuntime → runtime.start()`) is the most likely place a bug would hide, and it is the case the task description explicitly named.

---

### P1 — Pause test has a hidden assumption about job timing (line 186)

The pause test waits for a pending job and then calls pause. However, between the `waitForRoomTickJob(['pending'])` returning and the `room.runtime.pause` RPC being sent, the job processor may have already dequeued the job (status flips to `processing`). `cancelPendingTickJobs` only cancels `pending` jobs — the tick handler for the in-flight job will not re-enqueue because it checks `runtime.getState() === 'running'` in its `finally` block, so the behavior is still correct. But the assertion `pendingAfter.length === 0` could transiently be vacuously true because the only job is in `processing`, not `pending`. A comment acknowledging this should be added, or the assertion should explicitly verify that no *additional* pending jobs are created after the pause.

---

### P2 — `dedup` test assertion is weaker than the dedup contract (lines 159–164)

The dedup test asserts `roomAtStart.length === 1` for `['pending', 'processing']` immediately after room creation. Because of the same race described in P0, `roomAtStart` could be empty (no runtime yet), causing the test to fail. This test also needs `waitForRoomTickJob` before making the count assertion.

Additionally, the final assertion `roomPending.length <= 1` is technically weaker than the dedup guarantee (`=== 1`). If a bug caused *zero* pending jobs to remain it would pass silently. The assertion should be `=== 1` since after self-scheduling exactly one pending job must exist.

---

### P3 — Nit: `roomJobs[0]` without length guard in test 1 (line 128)

`roomJobs[0]` is accessed unconditionally. If `roomJobs` is empty this throws an index-out-of-bounds style error with no diagnostic message. Use `expect(roomJobs[0]).toBeDefined()` or access via the job variable after the `waitForRoomTickJob` fix.

---

### What is done well

- The overall test structure mirrors the existing `github-poll-job.test.ts` pattern faithfully.
- The `waitForRoomTickJob` helper with a configurable deadline is the right approach and is used consistently for all async assertions.
- The `cancelPendingTickJobs` / restart tests validate the lifecycle correctly once the runtime is in a known state.
- `room-tick-job.test.ts` is correctly added to `ROOM_FILES` in the validation script (validator stays green without touching the CI matrix, which is appropriate since all room shards are intentionally commented out due to resource usage).

lsm (commented):

**Recommendation: APPROVE**

The previous review flagged a P0 race condition, a missing daemon restart test, a weak dedup assertion, and repeated type casts. All five issues have been properly addressed in the latest commit:

### Previously flagged issues — resolution status

1. **P0 race condition in test 1** — Fixed. Test now uses `waitForRoomTickJob()` poll (line 136) instead of a synchronous `listJobs` read immediately after `createRoom()`. This eliminates the race between the RPC response and the event-driven runtime subscription.

2. **P1 missing daemon restart test** — Added. The 8th test (`daemon restart: existing rooms get tick jobs re-enqueued via initializeExistingRooms`, lines 318–410) creates a room on daemon1, shuts it down while preserving the DB, then starts daemon2 against the same DB and verifies the recovery path (`initializeExistingRooms → recoverRoomRuntime → runtime.start()`) re-enqueues tick jobs. The test properly manages environment variable restoration and cleanup in a `finally` block.

3. **P1 dedup assertion strengthened** — Fixed. The assertion at line 200 is now `expect(roomPending.length).toBe(1)` (strict equality) instead of the previous `<= 1`, which would have silently passed if a bug caused zero pending jobs.

4. **P1 pause timing comment** — Added. Lines 207–212 now include a detailed comment explaining that the job processor may pick up the initial job before the test observes it as `pending`, and that in-flight `processing` jobs self-terminate because the handler checks `getState() === 'running'` in its `finally` block.

5. **P2 extracted `getRoomIdFromPayload()` helper** — Done. The helper at line 64 eliminates repeated `(j.payload as { roomId?: string }).roomId` casts throughout the file.

### New code assessment

**Daemon restart test (lines 318–410):** This is the most complex test and is well-implemented:
- Correctly preserves the DB path before shutting down daemon1
- Uses `createDaemonApp()` directly for daemon2, which properly exercises `initializeExistingRooms()`
- Establishes a WebSocket client to daemon2 to verify runtime state via RPC
- Environment variable `NEOKAI_WORKSPACE_PATH` is saved/restored correctly (handles both `undefined` and set cases)
- Cleanup is thorough: transport close, daemon2 cleanup, and workspace removal all in a `finally` block
- Sets `daemon = null` to prevent `afterEach` from double-cleaning

**Minor observations (P3):**

1. **Daemon restart test: stale jobs from daemon1 in daemon2's DB** (line 373): `waitForRoomTickJob` on daemon2 searches for jobs with `['pending', 'processing', 'completed']`. Since the DB is shared, completed/pending jobs from daemon1's run are still present. The test could match a stale job from daemon1 rather than one freshly enqueued by daemon2's `initializeExistingRooms`. In practice this is harmless — the test goal is to verify that a tick job *exists* for the room after daemon2 starts, and a stale completed job would still prove the room was set up. But if you wanted to be precise, you could record the max job ID before daemon2 starts and assert the matched job has a higher ID.

2. **No explicit assertion that daemon1's jobs are cleaned up before daemon2 starts** (line 341): `daemonCtx1.cleanup()` stops the job processor, but completed/pending jobs remain in the DB. This is by design (the DB is persistent), but worth noting for test clarity.

3. **`path` import** (line 25): Only used in the daemon restart test for `path.dirname(dbPath)`. Fine, but a comment or co-location near its usage would aid readability.

Overall, the test file is thorough, well-documented, and addresses all previously raised concerns. The 8 tests cover the complete room tick lifecycle: initial enqueue, completion + re-scheduling, dedup invariant, pause, resume, stop, restart, and daemon restart recovery.

lsm (commented):

**Recommendation: APPROVE** (own PR — commenting instead of approving)

This re-review examines the five addressed issues from the prior round. Overall the work is in substantially better shape. The critical race condition fix, dedup tightening, pause-timing comment, `getRoomIdFromPayload` extraction, and the new daemon-restart test are all solid. A few residual concerns remain.

---

### What was addressed correctly

**P0 fixed — race condition in test 1**: `waitForRoomTickJob` polling replaces the synchronous `listJobs` call. The poll correctly includes `['pending', 'processing', 'completed']` so the test won't race past the initial job.

**P1 fixed — dedup final assertion tightened**: The second assertion at line 200 (`pendingAfter.filter(...).length`).`toBe(1)` correctly catches the zero-job bug that a `<=` check would miss. The first concurrent check staying `<=1` is intentional and correctly commented.

**P1 fixed — pause timing comment**: The comment at lines 207–212 accurately describes why `cancelPendingTickJobs` handles in-flight jobs correctly via the `finally` block.

**P1 fixed — daemon restart test added**: Test 8 (`packages/daemon/tests/online/room/room-tick-job.test.ts` lines 318–410) exercises the full `initializeExistingRooms → recoverRoomRuntime → runtime.start()` path. The environment-restoration logic in the `finally` block correctly handles the `NEOKAI_WORKSPACE_PATH` reset.

**P3 fixed — `getRoomIdFromPayload` helper**: Eliminates the repeated `(j.payload as { roomId?: string }).roomId` cast in three places.

---

### Residual issues

**P2 — `hub2` is never cleaned up in the daemon restart test**

In the `finally` block of test 8 (lines 395–408), `transport2.close()` is called but `hub2.cleanup()` is not. `MessageHub.cleanup()` exists and rejects all pending calls and clears handlers. Without it, any pending requests on `hub2` dangle until their timers fire. In a test file running multiple cases this can leave background timer activity that causes noise in subsequent tests or warns in the output. Recommend adding `hub2.cleanup()` before or alongside `transport2.close()`:

```ts
// In the finally block, before transport2.close():
try { hub2.cleanup(); } catch { /* already cleaned */ }
```

(`hub2` is declared inside the `try` block so it would need to be hoisted to the outer scope to be accessible in `finally`, similar to how `transport2` is handled.)

**P2 — resume and stop tests call lifecycle RPCs before the runtime is guaranteed ready**

Test 5 (`resume`, line 238) and test 6 (`stop`, line 268) call `room.runtime.pause`/`room.runtime.stop` immediately after `createRoom(daemon)` returns. Unlike tests 4 and 7, they do not call `waitForRoomTickJob` first. The `room.created` event is dispatched fire-and-forget via `queueMicrotask` inside `daemonHub.emit`, so the runtime is created asynchronously. In practice JavaScript's microtask ordering means the runtime is ready before the WebSocket frame leaves the server, but this depends on implementation details of the Bun WebSocket server's send path. Tests 4 and 7 correctly guard with `waitForRoomTickJob`; tests 5 and 6 should do the same for consistency and robustness.

Test 5 already verifies the post-pause state with `getPendingTickJobsForRoom` (line 241), but only works if pause succeeded. If pause throws "No runtime for room" on a slow machine, the test fails with a misleading error rather than a helpful assertion.

**P2 — hardcoded 500 ms sleep in stop test**

Line 282: `await new Promise<void>((resolve) => setTimeout(resolve, 500))`. Since the re-scheduled tick has a 30 s delay, 500 ms is sufficient in practice, but a sleep-based negative assertion is fragile. If the re-enqueue delay is ever reduced (e.g., during refactoring), this test would give a false positive. A comment explaining the 30 s invariant would at minimum make the fragility explicit; alternatively, a short poll asserting no new pending job appears within a bounded window is more expressive.

**P3 — PR description table lists 7 tests; code has 8**

The test table in the PR body lists 7 rows. The 8th test (daemon restart) is referenced in the description bullets but omitted from the table. Minor doc inconsistency.

**P3 — verbose `InstanceType<typeof WebSocketClientTransport>`**

Line 353: `let transport2: InstanceType<typeof WebSocketClientTransport> | null = null`. Since `WebSocketClientTransport` is a class imported directly, `WebSocketClientTransport | null` is sufficient.

---

### Summary

The five previously flagged issues are addressed. The daemon-restart test is a meaningful addition that covers the recovery path. The remaining P2s are real but unlikely to cause CI failures given current timing characteristics; the most actionable one is the missing `hub2.cleanup()` in the `finally` block (a two-line fix). The P3s are cosmetic.

lsm (commented):

**Recommendation: APPROVE**

All four P2 issues from the previous review round have been addressed:

1. **Race guard before lifecycle RPCs** -- Tests 5 (resume, line 239) and 6 (stop, line 274) now call `waitForRoomTickJob()` before issuing `room.runtime.pause`/`room.runtime.stop`, ensuring the runtime is fully initialized before lifecycle transitions.

2. **hub2 hoisted with cleanup in finally** -- `hub2` and `transport2` are declared in the outer scope (lines 362-363) and properly cleaned up in the `finally` block (lines 406-422), preventing resource leaks on test failure.

3. **Negative assertion sleep comment** -- Lines 287-290 explain why the 500ms sleep is sufficient: the tick handler re-enqueues with a 30s delay, so any spurious re-enqueue would be caught well within that window.

4. **Simplified type annotation** -- `WebSocketClientTransport` is used directly as a type (line 363) instead of the verbose `InstanceType<typeof ...>` form.

The test suite is thorough and well-documented. Each test has clear setup, action, and assertion phases. The daemon restart test (test 8) properly manages two daemon instances sharing a DB, with correct environment variable save/restore and cleanup ordering. The `waitForRoomTickJob` polling helper is robust with configurable timeout and status filtering.

No remaining issues found. Clean approve.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 race condition was already addressed in tests 5 and 6 but was missed in test 7, plus the PR description misstates the test count.

---

### P1 — Test 7 (restart) missing waitForRoomTickJob guard before stop

**File:** `packages/daemon/tests/online/room/room-tick-job.test.ts`, lines 296–325

Tests 5 and 6 were explicitly fixed this round with a `waitForRoomTickJob` guard before issuing lifecycle RPCs (documented in the PR description: "Tests 5 and 6 now guard with waitForRoomTickJob() before issuing lifecycle RPCs"). Test 7 has the identical race and was not fixed.

The sequence in test 7 is:
```ts
const roomId = await createRoom(daemon);           // line 298
// room.created fires via queueMicrotask — runtime may not exist yet
await daemon.messageHub.request('room.runtime.stop', { roomId }); // line 301
```

The `room.runtime.stop` RPC handler (packages/daemon/src/lib/rpc-handlers/room-handlers.ts, line 351) throws `Error('No runtime for room: ...')` when `stopRuntime` returns false, which happens when the runtime is not yet in the map. Since `room.created` dispatches runtime creation via `queueMicrotask` in `subscribeToEvents`, there is a window between the RPC response from `room.create` and the runtime being registered where `stop` will fail.

Tests 5 and 6 now guard correctly:
- Test 5 (pause): `await waitForRoomTickJob(daemonCtx, roomId, ['pending']);` at line 213 before issuing pause.
- Test 6 (resume): `await waitForRoomTickJob(daemonCtx, roomId, ['pending', 'processing', 'completed']);` at line 239 before issuing pause.

Test 7 must do the same — add a `waitForRoomTickJob` call at line 300 before `room.runtime.stop`.

---

### P3 — PR description says "7 tests" but the file contains 8

The PR description table lists 7 rows and the summary says "Add ... 7 tests", but the file has 8 `test()` calls. The PR title says "online test for room tick" which is neutral, but the description body line "7 tests covering end-to-end room.tick job queue mechanics" is incorrect. Minor cosmetic issue; the code itself is unaffected.

---

### Everything from the previous round is addressed

The following P2 items from the last review are now properly resolved:

- **Tests 5 and 6**: `waitForRoomTickJob` guard before lifecycle RPCs — done correctly.
- **hub2 hoisting and cleanup**: `hub2` is declared at the top of the try block and `hub2.cleanup()` is called in the `finally` block — done correctly.
- **500 ms sleep comment**: the comment at line 288–290 explains why 500 ms is sufficient for the negative assertion — done correctly.
- **Type simplification**: `WebSocketClientTransport` instead of `InstanceType<typeof WebSocketClientTransport>` — done correctly.
- **enqueueRoomTick dedup replace logic**: the test does not verify the replace-sooner path. This was accepted as a P2 in the previous round and is still not tested. That is acceptable since unit tests cover this path in room-tick.handler.ts. No new action required.

lsm (commented):

**Recommendation: APPROVE**

This is a well-structured and thorough online test file covering end-to-end room tick job queue mechanics. The latest commit (c10f6f1) correctly adds the `waitForRoomTickJob` guard in test 7 (restart) before calling `room.runtime.stop`, matching the same pattern used in tests 5 and 6. This eliminates the race where `room.created` fires via `queueMicrotask` but the runtime is not yet in the map when the stop RPC arrives.

### What looks good

- **Comprehensive coverage**: 8 tests covering enqueue, completion/re-schedule, dedup, pause, resume, stop, restart, and daemon restart recovery — all the scenarios listed in the task.
- **Race condition handling**: Consistent use of `waitForRoomTickJob()` polling before issuing lifecycle RPCs, preventing "No runtime for room" failures on slow machines.
- **Daemon restart test (test 8)**: Properly preserves the DB across daemon instances, exercises `initializeExistingRooms`, and has thorough cleanup in the `finally` block (hub cleanup before transport close, environment variable restoration, workspace removal).
- **Dedup assertion tightened**: Using `toBe(1)` instead of `toBeLessThanOrEqual(1)` so a bug producing zero pending jobs cannot pass silently.
- **Good helper extraction**: `getRoomIdFromPayload()` eliminates repeated casts, `waitForRoomTickJob()` and `getPendingTickJobsForRoom()` keep the test bodies readable.

### Minor observations (P3)

1. **Pause test (line 213)**: This test waits for `['pending']` only, while tests 5-7 wait for `['pending', 'processing', 'completed']` as the readiness guard. In practice the self-rescheduled job (30s delay) is reliably observable as `pending` so this works, but the narrower filter is slightly inconsistent with the other tests. Not a real risk since the pause test is specifically trying to observe a pending job to cancel it — it just means the test implicitly relies on either catching the initial job before the processor picks it up, or catching the re-scheduled 30s-delayed job. Both paths work.

2. **Stop test negative assertion (line 291)**: The 500ms sleep for the negative assertion is acknowledged in the comment as fragile. This is an inherent limitation of negative assertions — there is no better approach without injecting a test hook. The comment makes the tradeoff explicit, which is the right call.

No blocking issues found. The fix in this round is minimal, targeted, and correct. All 8 tests address the scenarios requested in the task.

lsm (commented):

**Recommendation: APPROVE** (own PR — commenting instead of approving)

This is the third re-review pass for PR #669. The one targeted fix described in the task was correctly applied. All issues from prior rounds are resolved. No new concerns found.

---

### Verified fix: test 7 waitForRoomTickJob guard

The prior round's P1 was that test 7 (restart) was missing the same runtime-readiness guard that tests 5 and 6 had. The fix is confirmed at line 303:

```ts
await waitForRoomTickJob(daemonCtx, roomId, ['pending', 'processing', 'completed']);
```

This was added before the `room.runtime.stop` call at line 306, closing the window where `stop` could throw `'No runtime for room'` if the microtask-dispatched `room.created` handler had not yet registered the runtime in the map. The guard pattern is now consistent across tests 5, 6, and 7.

---

### All prior issues confirmed resolved

- **Tests 5 and 6**: `waitForRoomTickJob` guards before lifecycle RPCs — intact.
- **Test 7**: now matches the pattern of tests 5 and 6 — new in this round, correct.
- **hub2 hoisting and cleanup**: `hub2` and `transport2` are outer-scope (`null`-initialized), and `hub2.cleanup()` is called before `transport2.close()` in the `finally` block (lines 411–416) — correct ordering, prevents dangling pending calls.
- **500 ms sleep comment**: the comment at lines 287–290 correctly explains the 30 s re-enqueue delay invariant — still present.
- **Type annotation**: `WebSocketClientTransport | null` at line 368, not the verbose `InstanceType<typeof ...>` form — correct.
- **dedup test strictness**: final assertion uses `toBe(1)` not `toBeLessThanOrEqual(1)` (line 200) — still present.

---

### No new issues found

The file is self-contained, all 8 tests are correctly guarded against the runtime-readiness race, cleanup is thorough, and the daemon-restart test correctly manages two daemon instances sharing a database. The implementation is ready to merge.

lsm (commented):

### Summary

This PR adds 3 new tests (bringing the total from 7 to 10) to the online room tick job test suite: "tick handler skips re-scheduling when runtime is stopped" (bug fix for a race condition), "each room gets its own independent tick job", and "daemon restart: existing rooms get tick jobs re-enqueued via initializeExistingRooms". The diff against `dev` is clean — only 113 additions and 2 line changes in the single test file.

### Assessment: APPROVE

The implementation is solid and well-structured. Specific observations:

**Correctness**
- The daemon restart test correctly exercises the `initializeExistingRooms` → `recoverRoomRuntime` → `runtime.start()` code path by preserving the DB across daemon instances. This is the exact production recovery path.
- The race condition fix in the "skips re-scheduling when runtime is stopped" test (line 301, adding `await waitForTickJob(...)` before stopping) properly addresses the `queueMicrotask` timing issue where the runtime may not be in the map yet when `room.create` resolves.
- Environment variable save/restore in the `finally` block (lines 426-431) correctly handles both `undefined` and previously-set values.

**Cleanup & resource management**
- The daemon restart test has thorough cleanup in the `finally` block: env var restoration, hub cleanup, transport close, daemon context cleanup, and workspace removal. Each cleanup step is individually guarded with try/catch.
- Setting `daemon = null` (line 381) to prevent `afterEach` double-cleanup is correct.

**Minor observations (P3)**
1. Line 381: `daemon = null as unknown as DaemonServerContext` — the double cast is slightly unusual but acceptable since it's guarded by the `if (daemon)` check in `afterEach`.
2. The `WebSocketClientTransport` import and manual client setup in the restart test adds some coupling to internal transport details, but this is justified given the test needs to connect to a separately-created daemon instance.

No bugs, security concerns, missing error handling, or test gaps found. The tests verify meaningful end-to-end behavior and the merge resolution with the dev branch is clean.

lsm (commented):

---

This is a solid, well-structured addition. The tests cover the stated requirements (enqueue on start, dedup, pause/resume, stop/restart, self-scheduling chain, stopped-runtime no-reschedule, multi-room isolation, daemon restart recovery). The helper functions are clean and the comments explaining the rationale for each design choice are useful. Here are my findings:

### P1 — PR Description Test Count Mismatch

The PR body says "7 tests" and the table lists 7 rows, but the file contains **10 tests**. The description was not updated after the merge conflict resolution added tests 8–10 (`tick handler skips re-scheduling`, `each room gets its own independent tick job`, `daemon restart`). Anyone reading the PR later will get an inaccurate picture of what landed.

`packages/daemon/tests/online/room/room-tick-job.test.ts` — 10 `test()` blocks  
PR description — table lists 7.

### P2 — Stale CI Note: This Test Never Runs Automatically

All `room/*` shards are commented out in `.github/workflows/main.yml` (lines 204–220). The `validate-online-test-matrix.sh` validator adds the file to `ROOM_FILES` which is correct for tracking, but the script's own comment says CI does not automatically run room files. The test will only run locally or when a CI shard is explicitly re-enabled. This is consistent with the other room tests and is acceptable — but it should be noted explicitly in the PR description so reviewers don't assume coverage is automated.

### P2 — `acceleratePendingTick` Bypasses Production Dedup for the Delete Step

```ts
// packages/daemon/tests/online/room/room-tick-job.test.ts:111-119
function acceleratePendingTick(daemonCtx: DaemonAppContext, roomId: string): void {
  const pending = listTickJobs(daemonCtx, roomId, ['pending']);
  for (const job of pending) {
    daemonCtx.jobQueue.deleteJob(job.id);  // direct DB delete
  }
  enqueueRoomTick(roomId, daemonCtx.jobQueue, 0);
}
```

The comment in the function correctly notes that if `RoomRuntime.scheduleTick()` fires between the delete and the re-enqueue, `enqueueRoomTick`'s dedup guard absorbs it. However, the direct `deleteJob` call bypasses any production state machine around job deletion (e.g., if the job is transitioning from `pending` to `processing` at that exact moment). In practice this is fine because there is no such state machine — `deleteJob` is a raw `DELETE` — but the test introduces a pattern that couples it to the production repository internals. Minor, but worth a brief comment acknowledging the direct-DB-delete tradeoff.

### P2 — `process.env` Mutation Lives Outside the `try` Block in Daemon Restart Test

```ts
// packages/daemon/tests/online/room/room-tick-job.test.ts:386-387
const originalWorkspacePath = process.env.NEOKAI_WORKSPACE_PATH;
process.env.NEOKAI_WORKSPACE_PATH = workspacePath;   // <-- before try

let daemonCtx2: DaemonAppContext | null = null;
// ...
try {
  const config = getConfig();  // reads NEOKAI_WORKSPACE_PATH
```

The env mutation happens before the `try` block, so any sync failure between the mutation and `try` entry would leave the env dirty. In practice there is nothing between the mutation and `try` so there is no real risk — but the safer pattern (used everywhere else in daemon-server.ts) is to mutate inside the try block. Consider moving lines 386-387 to the first two lines inside the `try` block.

### P3 — `hub2.request('room.runtime.state')` Type Assertion Includes Unreachable `null`

```ts
// Line 421-424
const stateResult = (await hub2.request('room.runtime.state', {
  roomId,
})) as { state: string | null };
expect(stateResult.state).toBe('running');
```

The RPC handler at `room-handlers.ts:297` returns `{ state: state ?? 'stopped' }` — it never returns `null`. The type assertion `{ state: string | null }` is technically inaccurate. `{ state: string }` is the correct shape. Nit only.

### P3 — `SETTLE_MS = 300` Is an Undocumented Magic Number

`SETTLE_MS` is defined at the top as 300 ms. The value is reasonable, but a brief comment explaining the choice (e.g., "allow one job processor poll cycle (1 s default) to complete, plus a margin") would help future readers understand why 300 ms is adequate or if it needs tuning in slow CI.

---

### What Works Well

- **Race condition guard in `tick handler skips re-scheduling`**: the `waitForTickJob` call before issuing `room.runtime.stop` correctly eliminates the race between `room.create` resolution and the async `queueMicrotask` that creates the runtime. This was the right fix.
- **Daemon restart test**: correctly exercises `initializeExistingRooms` → `recoverRoomRuntime` → `runtime.start()` → `enqueueRoomTick` without needing real LLM calls. The two-daemon setup, DB path preservation, and env restoration are all handled correctly.
- **`waitForTickJob` ordering**: `listJobs` orders by `created_at DESC`, so `waitForTickJob` naturally returns the most recently created job first — this ensures the 'tick handler skips re-scheduling' test's second `waitForTickJob(['completed'])` returns the newly enqueued stopped-runtime job rather than the earlier initial tick. The ordering is load-bearing for test correctness and it works.
- **Validator registration**: `room-tick-job.test.ts` is correctly added to `ROOM_FILES` in `validate-online-test-matrix.sh`.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The fixes from the previous review round have been addressed well. The code is clean, thorough, and correctly exercises the production room tick job queue machinery.

### What was fixed

1. **process.env mutation moved inside try block** (line 404-406): `originalWorkspacePath` is now captured inside the `try` block, paired with the `finally` restoration. Correct.

2. **acceleratePendingTick comment** (lines 119-123): Clear explanation of why direct `deleteJob()` is acceptable in test-only code. Addresses the concern adequately.

3. **Type assertion fix** (line 432): `(tickJob!.payload as { roomId: string }).roomId` -- uses `string` instead of `string | null`. Matches the typed return of `getDatabasePath(): string`.

4. **SETTLE_MS comment** (lines 52-57): Good explanation anchoring the 300ms to the 1s processor poll cycle.

5. **PR description updated**: Now lists all 10 tests with descriptions and includes the CI note about commented-out room shards.

**CI shard coverage**: The `room-tick-job.test.ts` file exists in the validator's `ROOM_FILES` array but has no corresponding shard entry in `.github/workflows/main.yml` -- not even a commented-out one. The other room test files at least have commented-out shard entries showing which module they belong to. This is documented in the PR description, which is acceptable since all room shards are intentionally disabled. But it means if someone later uncomments the room shards, they would need to manually add `room-tick-job.test.ts` to one of them. This is a minor bookkeeping gap, not blocking.

**`path.dirname(dbPath)` as workspace path** (line 385): This assumes the DB file lives directly in the workspace root. If the DB path structure ever changes (e.g., to a subdirectory), this test would silently use the wrong workspace. The assumption holds today but is implicit. Not blocking.

### Verdict

All 10 tests cover the requested scenarios: basic enqueue, dedup, pause/resume cancellation, stop, restart, self-scheduling chain, stopped-runtime skip, multi-room independence, and daemon restart recovery. The daemon restart test is particularly well structured with proper try/finally cleanup, environment restoration, and workspace removal. The code is ready to merge.

lsm (commented):

**Recommendation: APPROVE** — The implementation is sound, the review feedback was addressed correctly, and the CI shard situation is handled acceptably given the existing intentional policy for room tests.

---

### Summary

This PR adds 10 online tests for `room.tick` job queue mechanics using the Dev Proxy mock path. The code is well-structured, test logic is clear, and the fixes from the prior round of review are all correctly applied. Below is a detailed analysis of the one area flagged for scrutiny — the commented-out CI shards — plus minor observations.

---

### CI Shard Analysis: Commented-Out vs. Enabled

The core question from the re-review request: **does the new test need its own CI shard, or is documenting it in the PR description sufficient?**

**Conclusion: documenting is sufficient and the right call here.**

Evidence:

1. The comment in `.github/workflows/main.yml` line 137 predates this PR: `# Room tests temporarily disabled during refactoring`. The entire room test cohort is intentionally offline. This PR did not introduce that policy.

2. The validator script (`scripts/validate-online-test-matrix.sh`) was explicitly updated (line 45–57) to acknowledge this: `NOTE: All room/* shards are intentionally commented out… ROOM_FILES below tracks files that exist on disk; CI does not automatically run them.` The new file `room-tick-job.test.ts` is registered in `ROOM_FILES`, so the validator passes and no orphaned file warning fires.

3. `room-tick-job.test.ts` uses `NEOKAI_USE_DEV_PROXY=1` (mock SDK, no real API calls). This means it could theoretically be added to CI without real credentials — unlike the other room tests that need `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN`. This is a missed opportunity: a `mock_sdk: true` room-tick-job shard could run safely in CI today without needing real provider credentials. The CLAUDE.md hard-fail rule says tests that appear in the CI matrix must pass. Since the test does not appear in the matrix, the rule is not violated — but the opportunity to get automated CI coverage for a pure mock-SDK test is left on the table.

**The tradeoff is acknowledged in the PR description and the validator script comment. Leaving it uncommented is a deliberate scope choice, not an oversight.**

If the team wants CI coverage now without enabling the broader (expensive) room shards, the path is: add a standalone `room-tick-job` entry with `mock_sdk: true` in the matrix. That would be additive and non-breaking. But this is a P2 suggestion, not a blocker.

---

### Code Review Findings

**P2 — Missed opportunity: room-tick-job shard is CI-eligible today**

`room-tick-job.test.ts` makes no real LLM calls (the test file header confirms this at line 22–24), uses Dev Proxy, and explicitly passes with `NEOKAI_USE_DEV_PROXY=1`. Unlike the other commented-out room shards that require real API credentials, this test could be added as a `mock_sdk: true` entry right now. The CI workflow already has the plumbing (lines 280–293) to set up Dev Proxy for `mock_sdk` tests. This is the one concrete gap worth flagging. Not blocking this PR, but worth a follow-up issue or a quick CI entry in the next related PR.

**P3 — SETTLE_MS comment is accurate but the assumption warrants a note**

The `SETTLE_MS = 300` comment (lines 53–58) explains the reasoning based on a "1 s processor poll cycle". If the poll cycle were ever shortened (e.g., in a faster test environment) the 300 ms window could become a false-positive guard instead of an actual upper bound. The comment is good as written; just noting this for long-term maintainability.

**P3 — `waitForTickJob` returns `undefined` on timeout but callers use non-null assertion**

Tests call `await waitForTickJob(...)` and then assert `expect(job).toBeDefined()` before using `job!.queue`. If the function times out and returns `undefined`, the assertion fails with a clear message. This is fine — just noting the pattern is consistent across all callers.

**Previously-fixed items (confirmed resolved)**

- `process.env.NEOKAI_WORKSPACE_PATH` mutation moved inside `try` block: confirmed at line 405. `originalWorkspacePath` is declared before `try`, assigned inside — the `finally` restoration is correctly paired.
- `acceleratePendingTick` comment added: confirmed lines 119–124.
- Type assertion `string` (not `string | null`): confirmed at line 436–437 — `as { state: string }`.
- `SETTLE_MS` explanatory comment: confirmed lines 53–58.

---

### Overall Assessment

The implementation is correct and complete. The 10 tests cover the intended behaviors (enqueue, dedup, pause/resume/stop/restart cancellation, self-scheduling chain, stopped-runtime skip, multi-room isolation, and daemon restart recovery). The validator is updated consistently. The only actionable item is the P2 opportunity to run the mock-SDK room-tick-job test in CI without waiting to re-enable the broader room shards.

lsm (commented):

### Re-review after CI shard addition

The latest commit (301b2a91) adds a standalone `room-tick-job` shard to `.github/workflows/main.yml` with `mock_sdk: true` and `timeout: 10`, positioned immediately above the commented-out room shards. This is the right approach — the test makes no real API calls and can run independently of the broader room cohort.

### What looks good

1. **CI integration is correct.** The shard entry uses `mock_sdk: true` (Dev Proxy only), a 10-minute timeout (appropriate for the 30s max individual test timeout), and targets the single test file rather than the entire room directory. This avoids accidentally pulling in the disabled room tests.

2. **Daemon restart test is well-structured.** The try/finally block properly restores `process.env.NEOKAI_WORKSPACE_PATH`, tears down hub2/transport2/daemonCtx2 in the correct order, and removes the shared workspace. Moving the env mutation inside the try block (commit c076884) ensures the finally restoration is always paired.

3. **Race condition fix in "tick handler skips re-scheduling" test.** Adding `waitForTickJob()` before the stop RPC (line 313) correctly addresses the race between `room.create` resolution and the async `queueMicrotask` that creates the runtime.

4. **Comments are substantive.** The SETTLE_MS explanation, acceleratePendingTick tradeoff note, and dedup guard comment all add real value for future maintainers.

### Minor observations (P3)

- **Line 385: `path.dirname(dbPath)` assumption.** The test assumes the DB file lives directly inside the workspace root. This holds today (`{workspace}/.neokai/neokai.db` would break it, but `getConfig` puts dbPath at `{workspace}/neokai.db`). If the DB path convention changes, this test will need updating. Not a bug, just a coupling worth noting.

- **Line 393: `daemon = null as unknown as DaemonServerContext`.** The type cast is a pragmatic way to prevent afterEach double-cleanup, but a boolean flag (`let cleaned = false`) would be slightly cleaner. Very minor.

### Verdict

**Recommendation: APPROVE.** All previous review feedback has been addressed. The CI shard addition is clean and correct. The test suite is thorough, covering all requested scenarios (dedup, pause/resume, stop, restart, self-scheduling, cross-room independence, and daemon restart recovery) with proper cleanup and race condition handling.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review of the updated PR. The previous round's blocking issues have been resolved. Here is the verdict on each item raised, plus new findings from the added code.

---

### Previous P0 — Resolved

Test 1 now uses `waitForTickJob(daemonCtx, roomId, ['pending', 'processing', 'completed'])` (line 170) instead of a synchronous job list read. The race condition between `room.create` returning and the runtime's `queueMicrotask`-dispatched `room.created` handler is properly handled.

### Previous P1 — Resolved

The daemon restart test has been added (lines 374–469). It:
- Waits for daemon1's room to be fully set up
- Captures the DB path before shutdown
- Shuts down daemon1 cleanly via `daemonCtx.cleanup()` (bypassing `waitForExit` which would delete the workspace)
- Starts daemon2 using `createDaemonApp` with the preserved DB, directly exercising `initializeExistingRooms`
- Asserts a tick job appears on daemon2 for the existing room
- Verifies the runtime state is `running` via RPC
- Cleans up everything in a `finally` block with workspace deletion

This is a high-quality integration test for the recovery path.

---

### New observations

**P2 — Dedup test assertion remains weaker than the contract (line 196)**

The assertion `expect(pendingJobs.length).toBeLessThanOrEqual(1)` does not distinguish between "dedup is working" and "the job was cancelled." After the sequence pause/resume/pause/resume, the final state is paused — all pending ticks are cancelled — so `pendingJobs.length` will be 0. The assertion passes vacuously with 0 or 1. A value of 0 does not prove dedup worked. A more meaningful approach would be to assert exactly 1 pending job exists after ending on a `resume` (not a pause), which would prove the dedup guard absorbed the duplicates while still leaving one job standing.

**P3 — Implicit env coupling in daemon restart test**

`createDaemonApp` (line 411) inherits the current process environment, which still has `ANTHROPIC_BASE_URL=http://127.0.0.1:8000` and `ANTHROPIC_API_KEY=sk-devproxy-test-key` set by daemon1's `createInProcessDaemonServer`. Daemon2 picks up Dev Proxy routing implicitly. This is harmless (the room tick handler makes no LLM calls), but if a future change causes the test to exercise LLM paths, daemon2 will silently use the proxy. A comment noting this assumption would help readers.

**P3 — Dedup test polls for `['pending']` but the initial tick may already be `processing` (line 183)**

`waitForTickJob(daemonCtx, roomId, ['pending'])` at line 183 only returns when a pending tick exists. If the job processor picks up the initial tick before this line runs, the call will timeout (10s). The test would be more robust polling for `['pending', 'processing']` here — but in practice the 30s default delay on the re-scheduled tick makes this extremely unlikely to flap.

**P3 — `space` module precedent confirms CI pattern is sound**

The `room-tick-job` module is in the `include:` block only (not in the base `module:` list). This is identical to the `space` module's pattern. CI run history confirms `Daemon Online (space)` runs correctly via this pattern, so `room-tick-job` will run as a real CI shard.

---

### Summary

The two blocking issues from the previous review (synchronous job read race condition and missing daemon restart test) have been addressed. The daemon restart test is well-structured and properly exercises `initializeExistingRooms`. The remaining observations are minor and do not affect correctness or reliability at this scale.

lsm (commented):

**Recommendation: APPROVE**

The latest commit (`40f1fcac`) makes two targeted, correct fixes:

### 1. Dedup assertion strengthened (line 201)

Changed from `toBeLessThanOrEqual(1)` to `toBe(1)`. This is the right call. The pause/resume sequence ends on `resume`, so the runtime is running and exactly one pending tick must survive. The old assertion would have passed vacuously with zero jobs (e.g., if a bug in `enqueueRoomTick` silently failed), defeating the purpose of the dedup test. The accompanying comment clearly explains the reasoning.

### 2. Dev Proxy env coupling documented (lines 416-421)

The comment on `createDaemonApp` in the daemon restart test accurately describes the implicit environment inheritance: daemon2 picks up `ANTHROPIC_BASE_URL` and `ANTHROPIC_API_KEY` from daemon1's Dev Proxy setup. The note that this is harmless for the current tick handler (no LLM calls) but could matter if LLM paths are added later is useful forward-looking documentation.

### Full test file assessment

Having reviewed the complete 481-line file against all prior review rounds:

- **All prior P0/P1/P2 issues resolved**: The race condition fix (polling via `waitForTickJob`), daemon restart test, dedup tightening, pause timing comment, `hub2.cleanup()` in finally, and lifecycle guards before RPCs are all in place.
- **10 tests cover the complete lifecycle**: initial enqueue, dedup, pause, resume, stop, restart, self-scheduling chain, stopped-runtime skip, multi-room independence, and daemon restart recovery.
- **Cleanup is thorough**: The daemon restart test properly restores `NEOKAI_WORKSPACE_PATH`, cleans up hub/transport/daemon in a finally block, and removes the shared workspace.
- **CI integration is correct**: The `room-tick-job` shard is enabled with `mock_sdk: true` and a 10-minute timeout, appropriately separated from the disabled room test cohort that requires real credentials.

No remaining issues.

PR #673, opened 2026-03-22

lsm:
## Summary

- Adds `dbPath` option to `DaemonServerOptions` in `daemon-server.ts` so tests can share a persistent SQLite database across two daemon instances (required for crash-recovery testing)
- Creates `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` with 4 integration tests covering all job types
- Adds the new test file to the `features-2` CI shard in `main.yml` and updates `validate-online-test-matrix.sh`

## Test scenarios

| Test | Job type | What's verified |
|------|----------|----------------|
| Session title gen recovery | `session.title_generation` | Stale processing job is reclaimed immediately on restart; goes to terminal state (not stuck) |
| GitHub poll chain recovery | `github.poll` | Stale job reclaimed, completes, self-schedules next poll job (chain resumes) |
| Room tick recovery | `room.tick` | Stale tick reclaimed, completes; bootstrap re-seeds tick for existing room on startup |
| Cleanup job recovery | `job_queue.cleanup` | Stale cleanup reclaimed, runs, returns `{deletedJobs, nextRunAt}`, self-schedules next run ~24h out |

## Eager reclamation proof

Each test asserts that the stale job exits `processing` within **5 seconds** of daemon-2 startup. If `reclaimStale()` were only periodic (every 60 s), the job would be stuck for up to 60 s — well outside the 5-second window. This confirms `JobQueueProcessor.start()` calls `reclaimStale()` eagerly before the first poll tick.

## Test plan

- [x] All 4 tests pass locally with `NEOKAI_USE_DEV_PROXY=1`
- [x] `bash scripts/validate-online-test-matrix.sh` reports all files covered
- [x] Pre-commit checks (lint, format, typecheck, knip) all pass
- [x] File-backed SQLite DB used; temp files cleaned up in `afterAll`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -180,6 +180,7 @@ jobs:
           - module: features-2
             test_path: >-
               tests/online/features/github-poll-job.test.ts
+              tests/online/features/job-queue-crash-recovery.test.ts
               tests/online/features/message-persistence.test.ts
             mock_sdk: true
           - module: git
--- a/packages/daemon/tests/helpers/daemon-server.ts
+++ b/packages/daemon/tests/helpers/daemon-server.ts
@@ -363,6 +363,7 @@ async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<Dae
 		env: customEnv = {},
 		devProxy: devProxyOptions,
 		useDevProxy = false,
+		dbPath: customDbPath,
 	} = options;
 
 	// Start Dev Proxy if requested
@@ -388,6 +389,10 @@ async function spawnDaemonServer(options: DaemonServerOptions = {}): Promise<Dae
 		PORT: userPort.toString(),
 		NODE_ENV: 'test',
 		NEOKAI_SDK_STARTUP_TIMEOUT_MS: process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS || '30000',
+		// Forward custom DB path via DB_PATH env var so getConfig() picks it up
+		// in the spawned process.  When undefined, the spawned process uses its
+		// own workspace-derived default path (same behaviour as before).
+		...(customDbPath ? { DB_PATH: customDbPath } : {}),
 	};
 
 	// Note: Proxy env vars are inherited from parent process via ...process.env
--- a/packages/daemon/tests/helpers/daemon-server.ts
+++ b/packages/daemon/tests/helpers/daemon-server.ts
@@ -54,6 +54,16 @@ export interface DaemonServerOptions {
 	 * Default: false
 	 */
 	useDevProxy?: boolean;
+
+	/**
+	 * Custom database file path for the daemon.
+	 * When provided, overrides the default workspace-derived DB path.
+	 * Useful for crash-recovery tests where two daemon instances must share
+	 * the same persistent SQLite database file.
+	 *
+	 * The file is NOT deleted on cleanup — callers are responsible for removal.
+	 */
+	dbPath?: string;
 }
 
 export interface DaemonServerContext {
@@ -508,6 +518,7 @@ async function createInProcessDaemonServer(
 		env: customEnv = {},
 		devProxy: devProxyOptions,
 		useDevProxy = false,
+		dbPath: customDbPath,
 	} = options;
 
 	// Start Dev Proxy if requested
@@ -559,7 +570,9 @@ async function createInProcessDaemonServer(
 	process.env.NEOKAI_WORKSPACE_PATH = workspace;
 	const config = getConfig();
 	config.port = userPort;
-	config.dbPath = `${workspace}/daemon.db`;
+	// Use caller-provided dbPath (e.g., for crash-recovery tests sharing a DB across restarts)
+	// or fall back to the workspace-local path.
+	config.dbPath = customDbPath ?? `${workspace}/daemon.db`;
 
 	// Create daemon app in-process (starts its own server)
 	const daemonContext = await createDaemonApp({
--- a/packages/daemon/tests/helpers/standalone-server.ts
+++ b/packages/daemon/tests/helpers/standalone-server.ts
@@ -25,7 +25,7 @@ async function main() {
 			temperature: 1.0,
 			anthropicApiKey: process.env.ANTHROPIC_API_KEY || '',
 			claudeCodeOAuthToken: process.env.CLAUDE_CODE_OAUTH_TOKEN,
-			dbPath: ':memory:',
+			dbPath: process.env.DB_PATH || ':memory:',
 			maxSessions: 10,
 			nodeEnv: 'test',
 			workspaceRoot: `${process.env.TMPDIR || '/tmp'}/neokai-daemon-test-${Date.now()}`,
--- a/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+++ b/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
@@ -1,0 +1,474 @@
+/**
+ * Integration tests: job queue crash recovery
+ *
+ * These tests verify that all job types are correctly recovered after a
+ * daemon crash/restart. Each test:
+ *   1. Starts daemon-1 with a file-backed SQLite database.
+ *   2. Inserts a stale "processing" job (simulating a daemon that crashed
+ *      while handling that job) into the shared database.
+ *   3. Stops daemon-1 cleanly (to close the database connection).
+ *   4. Starts daemon-2 against the **same** database file.
+ *   5. Asserts that the stale job transitions out of 'processing' within a
+ *      short deadline — proving that JobQueueProcessor.start() eagerly calls
+ *      reclaimStale() rather than waiting for the 60-second periodic check.
+ *   6. Asserts that the processor subsequently finishes the job.
+ *
+ * Eagerness proof: if reclaimStale() were only periodic (every 60 s), the
+ * stale job would be stuck in 'processing' for up to 60 s.  The tests use
+ * a 5-second deadline for the reclaim assertion — succeeding quickly proves
+ * the call is eager, not deferred.
+ *
+ * The file-backed database persists between daemon instances. Each test
+ * generates a unique DB path under /tmp and cleans it up in afterAll.
+ *
+ * Run:
+ *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+ */
+
+import { afterAll, describe, expect, test } from 'bun:test';
+import path from 'path';
+import fs from 'fs';
+import type { DaemonServerContext } from '../../helpers/daemon-server';
+import { createDaemonServer } from '../../helpers/daemon-server';
+import type { DaemonAppContext } from '../../../src/app';
+import {
+	SESSION_TITLE_GENERATION,
+	GITHUB_POLL,
+	ROOM_TICK,
+	JOB_QUEUE_CLEANUP,
+} from '../../../src/lib/job-queue-constants';
+import type { Job, JobStatus } from '../../../src/storage/repositories/job-queue-repository';
+
+// ---------------------------------------------------------------------------
+// Constants
+// ---------------------------------------------------------------------------
+
+/** Environment variables enabling GitHub polling without real credentials. */
+const GITHUB_TEST_ENV: Record<string, string> = {
+	GITHUB_POLLING_INTERVAL: '300',
+	GITHUB_TOKEN: 'ghp_fake_token_for_crash_recovery_test',
+};
+
+/**
+ * Short deadline for the eager-reclaim assertion.
+ *
+ * If reclaimStale() is called eagerly on startup (as expected), the stale job
+ * transitions out of 'processing' within milliseconds.  We allow up to 5 s to
+ * accommodate any startup overhead.  If the assertion fires within this window,
+ * it proves reclamation was NOT deferred to the 60-second periodic schedule.
+ */
+const EAGER_RECLAIM_TIMEOUT_MS = 5_000;
+
+/** Maximum ms to wait for a job to reach a final state in daemon-2. */
+const JOB_COMPLETE_TIMEOUT_MS = 12_000;
+
+/** Poll cadence while waiting for a job status change. */
+const POLL_INTERVAL_MS = 100;
+
+/** How far in the past to back-date started_at so the job qualifies as stale. */
+const CRASH_AGO_MS = 6 * 60 * 1000; // 6 min > 5 min stale threshold
+
+// ---------------------------------------------------------------------------
+// Temp DB tracking — cleaned up in afterAll
+// ---------------------------------------------------------------------------
+
+const tempDbFiles: string[] = [];
+
+afterAll(() => {
+	for (const dbPath of tempDbFiles) {
+		for (const suffix of ['', '-shm', '-wal']) {
+			try {
+				fs.unlinkSync(dbPath + suffix);
+			} catch {
+				// File may not exist; ignore
+			}
+		}
+	}
+});
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+type InProcessDaemon = DaemonServerContext & { daemonContext?: DaemonAppContext };
+
+/** Extract DaemonAppContext from an in-process daemon. Throws in spawned mode. */
+function getDaemonCtx(daemon: DaemonServerContext): DaemonAppContext {
+	const ctx = daemon as InProcessDaemon;
+	if (!ctx.daemonContext) {
+		throw new Error(
+			'daemonContext not available — is DAEMON_TEST_SPAWN=true? Crash-recovery tests require in-process mode.'
+		);
+	}
+	return ctx.daemonContext;
+}
+
+/** Create a unique temp DB file path and register it for cleanup. */
+function makeTempDbPath(): string {
+	const dbPath = path.join(
+		'/tmp',
+		`crash-recovery-${Date.now()}-${Math.random().toString(36).slice(2)}.db`
+	);
+	tempDbFiles.push(dbPath);
+	return dbPath;
+}
+
+/**
+ * Insert a stale 'processing' job directly into the database.
+ *
+ * This simulates the daemon crashing while the job was in-flight: the row is
+ * in 'processing' state with a started_at that is 6 minutes old, exceeding the
+ * processor's 5-minute stale threshold.
+ *
+ * Returns the new job's ID.
+ */
+function insertStaleProcessingJob(
+	daemonCtx: DaemonAppContext,
+	queue: string,
+	payload: Record<string, unknown> = {},
+	maxRetries = 0
+): string {
+	const rawDb = daemonCtx.db.getDatabase();
+	const jobId = crypto.randomUUID();
+	const crashedAt = Date.now() - CRASH_AGO_MS;
+
+	rawDb
+		.prepare(
+			`INSERT INTO job_queue
+			(id, queue, status, payload, result, error, priority, max_retries, retry_count,
+			 run_at, created_at, started_at, completed_at)
+			VALUES (?, ?, 'processing', ?, NULL, NULL, 0, ?, 0, ?, ?, ?, NULL)`
+		)
+		.run(jobId, queue, JSON.stringify(payload), maxRetries, crashedAt, crashedAt, crashedAt);
+
+	return jobId;
+}
+
+/**
+ * Poll daemon-2's job queue until the job with `jobId` reaches one of the
+ * given statuses, or until the timeout expires.
+ */
+async function waitForJobById(
+	daemonCtx: DaemonAppContext,
+	jobId: string,
+	statuses: JobStatus[],
+	queue: string,
+	timeoutMs: number
+): Promise<Job | undefined> {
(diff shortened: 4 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR adds well-structured integration tests for job queue crash/restart recovery. The implementation is solid and covers all four required scenarios.

### What works well

1. **All four crash-recovery scenarios covered**: session title gen, GitHub poll chain, room tick, and cleanup job recovery. Each test follows the same daemon-1 → stale insert → stop → daemon-2 → assert pattern.

2. **File-backed DB correctly implemented**: The `dbPath` option in `DaemonServerOptions` cleanly overrides the workspace-local path in `createInProcessDaemonServer` (line 575 of `daemon-server.ts`). DB files live under `/tmp/crash-recovery-*` which is separate from the daemon's temp workspace (`/tmp/daemon-online-test-*`), so workspace cleanup via `rm -rf` doesn't destroy the shared DB.

3. **Eager stale reclamation proven**: The 5-second `EAGER_RECLAIM_TIMEOUT_MS` deadline (vs. the 60-second periodic interval) is the right way to prove eagerness. The `assertEagerReclaim` helper correctly accepts any non-processing state (including terminal states), handling the case where the job is already processed before the assertion runs.

4. **Self-scheduling chain verification**: The GitHub poll test (lines 343-346) checks that a new pending job exists after the stale one completes, and the cleanup test (lines 468-470) verifies the next run is scheduled 23+ hours out.

5. **CI and validation updated**: New test registered in `features-2` shard with `mock_sdk: true` and in `validate-online-test-matrix.sh`.

### Minor observations (P3)

- **`spawnDaemonServer` silently ignores `dbPath`** (line 362-366 of `daemon-server.ts`): The option is only destructured in `createInProcessDaemonServer`. This is fine because `getDaemonCtx()` throws a clear error if spawned mode is used, but documenting this in the JSDoc for `dbPath` (e.g., "Only supported in in-process mode") would prevent confusion.

- **`afterAll` cleanup is file-level, not per-describe**: If additional describe blocks were added to this file, `tempDbFiles` would still be cleaned up correctly since it's module-scoped. Not an issue today but worth noting.

- **The `CRASH_AGO_MS` constant (6 min) is tightly coupled to the processor's 5-min stale threshold**: If the threshold changes in production code, these tests could silently stop testing the right thing. A comment or import referencing the threshold would make this more maintainable.

Overall, this is a well-designed test suite that effectively validates crash recovery behavior through controlled database sharing between daemon instances.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two correctness issues and one latent bug in the helper need attention before merging.

---

### Summary

The overall structure is solid: file-backed DB, two-daemon pattern, EAGER_RECLAIM_TIMEOUT_MS=5s, CI matrix and validator both updated. The production-side eager reclaim in `JobQueueProcessor.start()` is verified correctly. The main issues are a weak assertion in Test 3, a race condition in Test 2 that could cause flakes, and a silent `dbPath` omission in `spawnDaemonServer`.

---

### P1: Test 3 "tick loop re-seeded" assertion is trivially satisfied (weak coverage)

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 401–408

The query includes `'completed'` in the status filter, which means the completed stale job (already confirmed a few lines above as `completedTick`) will always satisfy `>= 1`. The assertion says nothing about whether daemon-2 actually seeded a *new* pending tick.

To meaningfully test the re-seed invariant, the assertion should check for a **pending** tick job whose ID is *different* from the stale job:

```ts
// After completedTick confirmed above, check for a new seeded job
const pendingTicks = ctx2.jobQueue.listJobs({
    queue: ROOM_TICK,
    status: ['pending'],
});
const reseededTick = pendingTicks.find(
    (j) =>
        (j.payload as { roomId?: string }).roomId === roomId &&
        j.id !== staleJobId
);
expect(reseededTick).toBeDefined();
```

Without this, if daemon-2's startup bootstrap code is broken and never seeds a new tick, Test 3 still passes because the completed stale job satisfies the current assertion.

---

### P1: Race condition in Test 2 — triggerPoll stub set *after* job processor may have already started executing the reclaimed job

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 319–327

```ts
const daemon2 = await createDaemonServer({ env: GITHUB_TEST_ENV, dbPath });
const ctx2 = getDaemonCtx(daemon2);
// Stub set here — AFTER daemon startup and processor.start() have already run
const pollingService2 = ctx2.gitHubService?.getPollingService();
if (pollingService2) {
    pollingService2.triggerPoll = async () => {};
}
```

Inside `createDaemonApp()`, `jobProcessor.start()` is called synchronously. `start()` calls `reclaimStale()` then fires `this.tick()` (which is async but not awaited). By the time `createDaemonServer()` resolves and the test sets the stub, the event loop has had several microtask checkpoints. If the reclaimed `github.poll` job is dequeued during one of those checkpoints, it will call the real `triggerPoll` before the stub is in place, making a real GitHub API call (or a dev-proxy-forwarded fake call).

In practice this is unlikely to fail locally because dev-proxy intercepts the call, but it is a real race on correctness grounds. The stub should be set before daemon-2 starts — achieved either by hooking before `createDaemonApp` runs (not easy in-process) or by ensuring the `github.poll` handler is only registered after the stub is placed. This is an inherent limitation of the current test architecture and worth noting as a known trade-off.

---

### P2: `spawnDaemonServer` silently ignores `dbPath` — the option is parsed but never forwarded to the spawned process

**File:** `packages/daemon/tests/helpers/daemon-server.ts`, lines 360–366

`spawnDaemonServer` destructures `DaemonServerOptions` but omits `dbPath`:

```ts
const {
    port: userPort = 0,
    env: customEnv = {},
    devProxy: devProxyOptions,
    useDevProxy = false,
    // dbPath is NOT extracted — silently dropped
} = options;
```

If `DAEMON_TEST_SPAWN=true` is set, crash-recovery tests fail at `getDaemonCtx()` (which correctly throws about `daemonContext` being absent), so there is no data-loss risk. But the failure message says "Crash-recovery tests require in-process mode" rather than "dbPath is not supported in spawn mode", which could confuse future maintainers. Consider either extracting `dbPath` and passing it via `DB_PATH` env var to the spawned process, or adding an explicit guard at the top of `spawnDaemonServer`.

---

### P3: Temp DB cleanup only runs in `afterAll` — a test timeout leaves orphan files

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 77–87

If a test aborts due to the 60-second timeout, `afterAll` still runs and cleans up, so this is not a real issue in practice. However, if the process is hard-killed (e.g., CI `SIGKILL` after timeout), the /tmp files are orphaned. This is an acceptable trade-off for /tmp — just noting it for awareness.

---

### P3: `GITHUB_TOKEN: 'ghp_fake_token_for_crash_recovery_test'` is a realistic-looking GitHub token prefix

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, line 49

Using `ghp_` prefix in fake tokens sometimes triggers secret-scanning false positives in CI. Consider using a prefix like `test_` or `fake_` that cannot be confused with a real token format.

---

### Positive observations

- The two-daemon pattern correctly closes the DB before restarting — `stopDaemon()` calls `kill(SIGTERM)` + `waitForExit()` which triggers `daemonContext.cleanup()` including `db.close()`.
- CRASH_AGO_MS = 6 min > 5 min stale threshold is a correct margin.
- assertEagerReclaim's 5-second window provides a meaningful proof of eagerness.
- The cleanup test correctly deletes daemon-1's seeded pending jobs before inserting the stale one, ensuring deterministic tracking.
- CI matrix and `validate-online-test-matrix.sh` are both updated consistently.
- `dbPath` option is correctly documented as caller-managed (not auto-cleaned).

lsm (commented):

**Recommendation: APPROVE**

This re-review focuses on the two P1 fixes from the previous round.

### P1 Fix 1: Test 3 — room tick re-seed assertion (RESOLVED)

The previous review correctly identified that the "tick loop re-seeded" assertion was trivially satisfied by the already-completed stale job because the query included `'completed'` in the status filter.

The fix (lines 407-421) now:
1. Queries only for `'pending'` status ticks
2. Filters by `roomId` matching the test room
3. Requires `j.id !== staleJobId` to exclude the original stale job

This is exactly the right fix. The assertion now genuinely proves that daemon-2's bootstrap code (`roomRuntimeService.start().then(() => { for (const room of roomManager.listRooms()) { seedRoomTick(room.id); } })` in `packages/daemon/src/lib/rpc-handlers/index.ts:184-189`) actually ran and enqueued a new pending tick. If the seeding code were broken, no pending tick with a different ID would exist, and the test would fail.

The 5-second polling window (`RESEED_TIMEOUT_MS`) is reasonable since `enqueueRoomTick` is called synchronously after `roomRuntimeService.start()` resolves.

### P1 Fix 2: Test 2 — triggerPoll stub race condition (RESOLVED)

The previous review identified that setting `pollingService2.triggerPoll = async () => {}` after `createDaemonServer()` returns was racy — the job processor may have already dequeued and executed the reclaimed job by then.

The fix removes the stub entirely and adds a thorough comment (lines 316-327) explaining:
1. **Why the stub would be racy**: `JobQueueProcessor.start()` calls `reclaimStale()` then immediately fires `tick()`, which dequeues synchronously
2. **Why the stub is unnecessary**: no repositories are registered, so `triggerPoll()` -> `pollAllRepositories()` is a no-op
3. **Why errors don't matter**: `handleGitHubPoll()` wraps `triggerPoll()` in a try/catch (confirmed in `github-poll.handler.ts:44-48`)

This is the correct approach. The handler code at `github-poll.handler.ts:33-43` shows that when `pollingService` exists but `isRunning()` is false, it skips `triggerPoll()` entirely. And even if it were called, errors are caught. The job still completes and self-schedules the next poll (lines 67-73).

### Overall assessment

Both P1 fixes are correct and sufficient. The test suite now meaningfully validates:
- **Eager reclamation** via the 5-second deadline (all 4 tests)
- **Job completion** after reclamation (all 4 tests)
- **Self-scheduling chains** resume (Tests 2, 3, 4)
- **Tick loop reseeding** through a distinct new pending job (Test 3)

No new issues found. The P2 about `spawnDaemonServer` silently ignoring `dbPath` and the P3s from the previous review remain as minor improvement opportunities but are not blocking.

lsm (commented):

**Recommendation: APPROVE** — Both P1 issues from the previous round are correctly resolved. Remaining open items are P2 and P3 only.

---

### Re-review focus: were the two P1 fixes correct?

#### P1 Fix 1: Test 3 tick re-seed assertion — RESOLVED

**Previous issue:** The assertion included `'completed'` in the status filter, meaning the already-confirmed stale job trivially satisfied it. If the bootstrap seeding code were broken, the test would still pass.

**Fix applied** (`packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 410–421): The assertion now polls `listJobs({ queue: ROOM_TICK, status: ['pending'] })` and uses `.find(j => j.payload.roomId === roomId && j.id !== staleJobId)`. This is exactly the correct check. The seeding comes from `rpc-handlers/index.ts` line 184–193 where `roomRuntimeService.start().then(() => { for room of listRooms: seedRoomTick(room.id) })` calls `enqueueRoomTick(roomId, jobQueue, DEFAULT_TICK_INTERVAL_MS=30s)`. With a 5-second poll window, the assertion correctly fails if that bootstrap path is broken.

There is one subtle timing nuance: `roomRuntimeService.start()` is fire-and-forget (not awaited in `setupRPCHandlers`), and `jobProcessor.start()` fires synchronously after `setupRPCHandlers` returns. This means the reclaimed stale tick job may execute before `initializeExistingRooms` completes. In that case the room tick handler returns `{skipped: true, reason: 'not running'}` (runtime still in `'paused'` state at the time), and does NOT self-schedule a new tick. The new pending tick then comes from the bootstrap `.then()` seeding or from `runtime.start()` → `scheduleTick()` → `enqueueRoomTick(roomId, jobQueue, 0)` path. Either way the 5-second window is sufficient. The assertion is sound.

#### P1 Fix 2: Test 2 GitHub poll race condition — RESOLVED

**Previous issue:** The `triggerPoll` stub was set after `createDaemonServer()` returned, creating a window where the reclaimed job could have already called the real `triggerPoll` before the stub was placed.

**Fix applied:** The stub is removed entirely, with a clear comment explaining why it is not needed (`packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 316–328). I verified `pollAllRepositories()` in `packages/daemon/src/lib/github/polling-service.ts` (line 151): it iterates `this.repositories.values()`, and with no repositories registered in this test environment, `Promise.allSettled([])` completes immediately without any API calls. The `handleGitHubPoll` handler then self-schedules the next poll job and completes with `status: 'completed'`. The "CHAIN RESUMED" assertion at line 344 verifies the next pending job is enqueued. This is a correct and clean solution.

---

### Remaining open items from the previous review

**P2 (unchanged): `spawnDaemonServer` silently drops `dbPath`**

`packages/daemon/tests/helpers/daemon-server.ts`, lines 360–366. The `dbPath` option is destructured from `DaemonServerOptions` but never forwarded to the spawned process. The runtime guard in `getDaemonCtx()` catches this with a misleading error message about `daemonContext` being absent rather than the actual cause. This remains a maintainability hazard but does not affect correctness.

**P3 (unchanged): `ghp_fake_token_for_crash_recovery_test` uses the `ghp_` prefix**

Line 49 of the test file. Some secret scanners flag `ghp_` as a real GitHub Personal Access Token prefix. Consider using `test_fake_github_token_for_crash_recovery_test` instead.

---

### Summary

Both P1 issues are correctly addressed. The Test 3 re-seed assertion is now genuinely meaningful — it requires daemon-2's bootstrap seeding to fire. The Test 2 stub removal is properly justified by the empty-repositories invariant. The implementation is ready to merge.

lsm (commented):

### Re-review: P2 fix for silent `dbPath` drop in `spawnDaemonServer`

The fix in commit `11f44f40` is **correct and sufficient**. Here is my analysis:

**What was changed** (`packages/daemon/tests/helpers/daemon-server.ts`):
- `dbPath` is now destructured from `DaemonServerOptions` as `customDbPath` (line 366)
- When provided, it is forwarded as `DB_PATH` in the spawned process environment via conditional spread (line 395)
- When undefined, the spread produces an empty object, preserving the previous default behavior

**Why this works**: `getConfig()` in `packages/daemon/src/config.ts` (line 96) already reads `process.env.DB_PATH` as a fallback:
```ts
dbPath: overrides?.dbPath ?? (process.env.DB_PATH || defaultDbPath),
```
So setting `DB_PATH` in the spawned process env is the correct and minimal way to forward the caller's `dbPath` without modifying the spawned server code.

**Consistency check**: The in-process `createDaemonServer` variant (line 578-580) already handled `dbPath` correctly by assigning it directly to `config.dbPath`. The spawned variant now has parity.

**Edge case handling**: The conditional spread `...(customDbPath ? { DB_PATH: customDbPath } : {})` correctly avoids setting `DB_PATH` to an empty string when `dbPath` is undefined, which would have overridden the workspace-derived default in the spawned process.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The P2 fix is only half-complete. The spawned-mode path still silently drops `dbPath`.

---

### Re-review: P2 fix assessment

The fix correctly addresses two code paths:

1. **`createInProcessDaemonServer` (in-process mode)** — `config.dbPath = customDbPath ?? `${workspace}/daemon.db`` at line 580 of `daemon-server.ts`. This correctly overrides the workspace-local default. Fix is correct and sufficient for this path.

2. **`spawnDaemonServer` (spawned process mode)** — The fix sets `DB_PATH` in the spawned process's environment via `...(customDbPath ? { DB_PATH: customDbPath } : {})` at line 395. The PR description claims "`getConfig()` already reads `DB_PATH` as a fallback, so no changes to the spawned process are needed." **This claim is incorrect.**

### P1: `spawnDaemonServer` fix is semantically broken — `standalone-server.ts` never reads `DB_PATH`

**File:** `packages/daemon/tests/helpers/standalone-server.ts`, line 28

`standalone-server.ts` calls `createDaemonApp()` with a **hardcoded** config object that includes `dbPath: ':memory:'`. It never calls `getConfig()` and never reads `process.env.DB_PATH`. The environment variable forwarded by `spawnDaemonServer` is therefore silently discarded — the spawned daemon always uses an in-memory database regardless of what `dbPath` is passed to `DaemonServerOptions`.

This means `spawnDaemonServer({ dbPath: '/tmp/some.db' })` will:
1. Set `DB_PATH=/tmp/some.db` in the child process environment ✓
2. Start the child process with `standalone-server.ts` ✓
3. `standalone-server.ts` ignores `DB_PATH` and opens `':memory:'` instead ✗

The fix for the spawn path would require `standalone-server.ts` to read `process.env.DB_PATH` (or call `getConfig()` and use its `dbPath` field) rather than hard-coding `':memory:'`.

### Why this doesn't affect the crash-recovery tests today

The crash-recovery tests call `createDaemonServer({ dbPath })` without setting `DAEMON_TEST_SPAWN=true`, so they always take the in-process path. The in-process fix is correct, and the tests work as intended. However, the `spawnDaemonServer` change adds misleading dead code: it sets an environment variable that has no effect on the spawned process.

### What "sufficient" would look like

If the intent is to support `dbPath` only for in-process mode (which is all that the tests need), the `DaemonServerOptions` JSDoc for `dbPath` should document this explicitly: "Only supported in in-process mode. Has no effect when `DAEMON_TEST_SPAWN=true`." The unused code path in `spawnDaemonServer` should either be left unimplemented (with a comment) or fully implemented in `standalone-server.ts`.

As-is, the spawned-mode code path silently accepts a `dbPath` option, appears to forward it, but the spawned process ignores it — a confusing and error-prone state.

---

### Summary

The in-process half of the P2 fix is correct. The spawned-process half sets `DB_PATH` but `standalone-server.ts` hard-codes `dbPath: ':memory:'` and never reads `process.env.DB_PATH`, making the env-forwarding dead code. This should be resolved before merging.

lsm (commented):

### Re-review: P1 fix for `standalone-server.ts` hard-coded `dbPath`

**Recommendation: APPROVE**

The P1 issue is correctly and completely resolved. Here is the full chain trace:

1. **`DaemonServerOptions.dbPath`** — new field added to the options interface in `packages/daemon/tests/helpers/daemon-server.ts` (line 66).

2. **`spawnDaemonServer`** (spawned-process path) — reads `customDbPath` and conditionally injects `DB_PATH` into the child process environment (lines 366, 395). The conditional spread `...(customDbPath ? { DB_PATH: customDbPath } : {})` is correct: when no `dbPath` is provided, `DB_PATH` is absent from the env and the child process falls back to its workspace-derived default — same behaviour as before the PR.

3. **`standalone-server.ts`** — changed from hard-coded `':memory:'` to `process.env.DB_PATH || ':memory:'` (line 28). The spawned process now reads `DB_PATH` and passes it directly into `createDaemonApp`. The file does NOT call `getConfig()`; it builds the config struct inline, so reading `process.env.DB_PATH` directly here is the right approach.

4. **`createInProcessDaemonServer`** (in-process path) — reads `customDbPath` and sets `config.dbPath = customDbPath ?? `${workspace}/daemon.db`` (line 580). This path also works correctly: the in-process daemon receives the exact `dbPath` the test supplied, bypassing `getConfig()`'s env-var lookup entirely.

5. **`getConfig()` in `src/config.ts`** — line 96 confirms `DB_PATH` env var is read: `overrides?.dbPath ?? (process.env.DB_PATH || defaultDbPath)`. This is the production code path, which the spawned-process flow now correctly populates.

**The crash-recovery tests use in-process mode** (they call `getDaemonCtx()` which throws if `daemonContext` is absent), so they exercise path 4, not path 3. Path 3 (spawned) is now correct too for any future test that runs in spawned mode with a custom `dbPath`.

No remaining issues found. The end-to-end chain is complete.

lsm (commented):

### Re-review: P1 fix for DB_PATH forwarding

The P1 fix is correct and the full end-to-end `dbPath` forwarding chain is now solid.

**Forwarding chain verification:**

1. **Spawned mode** (`spawnDaemonServer`): `DaemonServerOptions.dbPath` is forwarded as `DB_PATH` env var (line 395 of `daemon-server.ts`). The spawned `standalone-server.ts` now reads `process.env.DB_PATH || ':memory:'` (line 28) and passes it directly to `createDaemonApp()`. Previously, `:memory:` was hard-coded, making the env var forwarding dead code. Fixed correctly.

2. **In-process mode** (`createInProcessDaemonServer`): `DaemonServerOptions.dbPath` is applied directly via `config.dbPath = customDbPath ?? ...` (line 580), bypassing `getConfig()` entirely. This path was already correct but the explicit override is now clearly documented with a comment.

No residual issues with the `getConfig()` call on line 576 -- it reads `process.env.DB_PATH` if set, but line 580 unconditionally overwrites `config.dbPath` with the caller-provided value, so there is no stale-env conflict.

**Test quality**: The four crash-recovery tests are well-structured with clear separation of daemon-1 (insert stale job) and daemon-2 (verify recovery). The eager-reclaim assertion with a 5-second deadline effectively proves the startup-time `reclaimStale()` call fires immediately rather than waiting for the 60-second periodic check.

**Recommendation: APPROVE** -- no remaining issues.

lsm:
Superseded by https://github.com/lsm/neokai/pull/682

PR #674, opened 2026-03-22

lsm:
Four scenarios verify that stale 'processing' jobs left by a crashed daemon
are eagerly reclaimed (not waiting 60 s for the periodic stale check) when
the daemon restarts with the same file-backed SQLite database:

1. session.title_generation — reclaimed, fails on fake session (maxRetries=0 → dead)
2. github.poll chain — stale job reclaimed and completed; poll chain resumes
3. room.tick — reclaimed, handler skips gracefully (no active runtime)
4. job_queue.cleanup — reclaimed and runs to completion, self-schedules next run

Each test uses a dedicated helper (createDaemonWithSharedDb) that calls
createDaemonApp directly with a fixed dbPath, allowing two sequential daemon
instances to share persistent job-queue state.

Eager reclamation is verified through timing (job must complete within 10 s)
and startedAt freshness (if re-dequeued, startedAt is not the old stale value).

CI matrix updated: new features-3 shard covers job-queue-crash-recovery.test.ts.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,6 +127,7 @@ jobs:
           - convo
           - features-1
           - features-2
+          - features-3
           - git
           - lifecycle
           - mcp
@@ -181,6 +182,9 @@ jobs:
             test_path: >-
               tests/online/features/github-poll-job.test.ts
               tests/online/features/message-persistence.test.ts
+            mock_sdk: true
+          - module: features-3
+            test_path: tests/online/features/job-queue-crash-recovery.test.ts
             mock_sdk: true
           - module: git
             test_path: tests/online/git
--- a/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+++ b/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
@@ -13,15 +13,17 @@
  * ## Eager reclamation design note
  *
  * JobQueueProcessor.start() calls reclaimStale() synchronously and then calls
- * tick() synchronously. tick() immediately dequeues pending jobs (SQLite
- * transaction), so by the time createDaemonApp() returns, the stale job has
- * already been cycled:  stale-processing → pending → (fresh-)processing.
+ * tick() synchronously. tick() dequeues pending jobs via a synchronous SQLite
+ * transaction, moving the job back to 'processing' before start() returns —
+ * but the actual job execution (processJob) is async and runs in the background.
+ * So by the time createDaemonApp() returns:
+ *   - The job has been cycled: stale-processing → pending → (fresh-)processing
+ *   - Job execution has begun but may not yet be complete
  *
- * Because the transition through 'pending' happens in the same synchronous call
- * as startup, the test verifies eager reclamation through timing rather than
- * catching the transient 'pending' state: if the job completes within a few
- * seconds of daemon2 starting, it MUST have been reclaimed immediately on
- * startup (not after the 60-second periodic stale check).
+ * The test verifies eager reclamation through timing: if the job reaches a
+ * final state within JOB_WAIT_TIMEOUT_MS seconds of daemon2 starting, it MUST
+ * have been reclaimed immediately on startup (not after the 60-second periodic
+ * stale check). The waitForJobById polling loop handles the async execution.
  *
  * Test scenarios:
  *   1. session.title_generation — stale job reclaimed; handler fails once (fake
@@ -145,9 +147,9 @@ function insertStaleCrashedJob(
  * Verify that a job was eagerly reclaimed by daemon2.
  *
  * reclaimStale() sets status='pending' and started_at=NULL for stale jobs.
- * tick() then immediately re-dequeues them (status='processing', fresh startedAt).
- * Both transitions happen synchronously inside start(), before createDaemonApp()
- * returns. We verify reclamation by checking that the stale state is gone:
+ * tick() then synchronously re-dequeues them (status='processing', fresh startedAt).
+ * Job execution is async and may not be complete yet. We verify reclamation by
+ * checking that the stale state is gone:
  *
  *   - 'pending'    → startedAt is NULL  (reclaimed, not yet re-dequeued)
  *   - 'processing' → startedAt is recent (re-dequeued after reclaim)
@@ -200,9 +202,6 @@ async function createDaemonWithSharedDb(
 	if (!process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS) {
 		process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS = '30000';
 	}
-	if (!process.env.TEST_WORKTREE_BASE_DIR) {
-		process.env.TEST_WORKTREE_BASE_DIR = `/tmp/daemon-worktrees-${Date.now()}`;
-	}
 
 	const config = getConfig();
 	config.port = 0; // OS-assigned port to avoid collisions
@@ -225,12 +224,11 @@ describe('Job queue crash/restart recovery (online)', () => {
 	// Tracked here so afterEach can delete them alongside tmpDir.
 	let createdWorkspaceDirs: string[];
 
-	// Env vars mutated by createDaemonWithSharedDb — saved before each test
-	// and restored in afterEach to prevent cross-test pollution.
+	// Env vars mutated by this suite — saved before each test and restored
+	// in afterEach to prevent cross-test pollution.
 	let savedWorkspacePath: string | undefined;
 	let savedTestWorktreeBaseDir: string | undefined;
-
-	// GitHub env vars saved for restoration
+	let savedSdkStartupTimeout: string | undefined;
 	let savedGithubPollingInterval: string | undefined;
 	let savedGithubToken: string | undefined;
 
@@ -243,11 +241,21 @@ describe('Job queue crash/restart recovery (online)', () => {
 		daemon2 = null;
 		createdWorkspaceDirs = [];
 
-		// Save env vars that createDaemonWithSharedDb will mutate
+		// Save env vars that this suite will mutate
 		savedWorkspacePath = process.env.NEOKAI_WORKSPACE_PATH;
-		savedTestWorktreeBaseDir = process.env.TEST_WORKTREE_BASE_DIR;
+		savedSdkStartupTimeout = process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS;
 		savedGithubPollingInterval = process.env.GITHUB_POLLING_INTERVAL;
 		savedGithubToken = process.env.GITHUB_TOKEN;
+
+		// Initialize TEST_WORKTREE_BASE_DIR here (not inside createDaemonWithSharedDb)
+		// so the path is tracked for cleanup. Reuse any value already set by the parent
+		// test runner; only set a per-test value when none is present.
+		savedTestWorktreeBaseDir = process.env.TEST_WORKTREE_BASE_DIR;
+		if (!savedTestWorktreeBaseDir) {
+			const worktreeDir = `/tmp/daemon-worktrees-${id}`;
+			process.env.TEST_WORKTREE_BASE_DIR = worktreeDir;
+			createdWorkspaceDirs.push(worktreeDir);
+		}
 	}, 30_000);
 
 	afterEach(async () => {
@@ -269,7 +277,7 @@ describe('Job queue crash/restart recovery (online)', () => {
 			daemon2 = null;
 		}
 
-		// Restore env vars mutated by createDaemonWithSharedDb
+		// Restore all env vars mutated by this suite
 		if (savedWorkspacePath === undefined) {
 			delete process.env.NEOKAI_WORKSPACE_PATH;
 		} else {
@@ -279,6 +287,11 @@ describe('Job queue crash/restart recovery (online)', () => {
 			delete process.env.TEST_WORKTREE_BASE_DIR;
 		} else {
 			process.env.TEST_WORKTREE_BASE_DIR = savedTestWorktreeBaseDir;
+		}
+		if (savedSdkStartupTimeout === undefined) {
+			delete process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS;
+		} else {
+			process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS = savedSdkStartupTimeout;
 		}
 
 		// Restore GitHub env vars
--- a/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+++ b/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
@@ -144,34 +144,36 @@ function insertStaleCrashedJob(
 /**
  * Verify that a job was eagerly reclaimed by daemon2.
  *
- * Since reclaimStale() and tick() both run synchronously inside start(),
- * the job transitions stale-processing → pending → (fresh-)processing
- * before createDaemonApp() returns. We verify reclamation by checking that
- * the job is either:
- *   (a) in 'processing' with a fresh startedAt (re-dequeued after reclaim), or
- *   (b) already in a final state (completed/dead/failed — processed immediately).
+ * reclaimStale() sets status='pending' and started_at=NULL for stale jobs.
+ * tick() then immediately re-dequeues them (status='processing', fresh startedAt).
+ * Both transitions happen synchronously inside start(), before createDaemonApp()
+ * returns. We verify reclamation by checking that the stale state is gone:
  *
- * In either case the stale startedAt is gone, proving reclaimStale() ran.
+ *   - 'pending'    → startedAt is NULL  (reclaimed, not yet re-dequeued)
+ *   - 'processing' → startedAt is recent (re-dequeued after reclaim)
+ *   - final states → reclaim + execution both completed immediately
  */
 function assertEagerlyReclaimed(
 	jobQueue: DaemonAppContext['jobQueue'],
 	queue: string,
 	jobId: string
 ): void {
-	// The job must NOT still be stuck in stale processing (started_at 6 min ago)
 	const allJobs = jobQueue.listJobs({
 		queue,
 		status: ['pending', 'processing', 'completed', 'failed', 'dead'],
 	});
 	const job = allJobs.find((j) => j.id === jobId);
-	expect(job).toBeDefined(); // job must still exist
+	expect(job).toBeDefined();
 
-	if (job!.status === 'processing') {
+	if (job!.status === 'pending') {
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured integration test file that verifies job queue crash/restart recovery across all four job types. The implementation is thorough and the approach is sound.

### Strengths

- **Clear design**: The two-daemon pattern (daemon1 inserts stale job, daemon2 reclaims it) cleanly simulates crash recovery without needing to actually crash a process.
- **Excellent documentation**: The file-level comment block and per-function JSDoc explain both _what_ and _why_ (especially the eager reclamation design note explaining the synchronous reclaimStale/tick sequence).
- **Correct stale threshold**: `STALE_AGE_MS = 6 * 60 * 1000` properly exceeds the processor's 5-minute `staleThresholdMs` default.
- **Proper CI integration**: features-3 shard added to the matrix, validation script updated to include the new file.
- **Robust cleanup**: `afterEach` restores GitHub env vars and uses best-effort daemon cleanup.

### Minor observations (P3)

1. **Workspace directory leak** (`createDaemonWithSharedDb`, line 186): Each call creates `/tmp/crash-recovery-ws-*` directories that are never cleaned up. Only `tmpDir` (the DB directory) is removed in `afterEach`. These are ephemeral `/tmp` dirs on CI so this is not harmful, but could accumulate during local development. Consider cleaning them up in `afterEach` or tracking them for removal.

2. **Race window on triggerPoll stub** (github.poll test, lines 328-331 and 354-357): The `triggerPoll` stub is applied after `createDaemonApp()` returns. The job processor starts synchronously inside `createDaemonApp`, meaning the initial poll job (or reclaimed stale job) could execute with the real `triggerPoll` before the stub is applied. This is benign because (a) the handler catches errors from `triggerPoll` (github-poll.handler.ts:44), and (b) the fake token prevents any real GitHub API calls. But it means the test's stub might not actually be exercised for the first poll. Not a correctness issue since the test assertions focus on job state, not on whether `triggerPoll` was called.

3. **`nextRunAt` assertion tolerance** (line 467): `expect(result!.nextRunAt).toBeGreaterThan(Date.now() + 60 * 60 * 1000)` could theoretically flake if there is significant wall-clock drift between when the handler computed `nextRunAt` and when this assertion runs, though in practice the margin is large enough that this is unlikely.

Overall this is clean, well-tested code that properly validates the eager reclamation behavior. No blocking issues.

lsm (commented):

> **Recommendation: REQUEST_CHANGES** — Three issues need to be addressed before merging: two correctness gaps that could cause spurious test failures or mask real bugs, and one environment-pollution issue affecting test isolation.

---

### Summary

The test file adds four crash/restart recovery scenarios that verify eager `reclaimStale()` execution on daemon restart. The core design — sharing a file-backed SQLite DB across two sequential in-process daemon instances — is sound, and the test structure is clear and well-commented. However, there are three issues worth fixing.

---

### P1 — `createDaemonWithSharedDb` leaks `NEOKAI_WORKSPACE_PATH` across tests

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 184–203

`createDaemonWithSharedDb` mutates `process.env.NEOKAI_WORKSPACE_PATH` on every call but never restores it. Each call generates a new unique workspace path, so by the end of the second `createDaemonWithSharedDb` call, `NEOKAI_WORKSPACE_PATH` holds the path of the last daemon2 workspace. Subsequent tests in the same Bun worker will inherit that stale path, and — worse — if this test file runs in the same worker as another `features-*` shard, `getConfig()` in those tests will pick up the leftover value.

The existing `createInProcessDaemonServer` helper (in `daemon-server.ts`) correctly saves and restores all environment variables it touches. `createDaemonWithSharedDb` should follow the same pattern: save the original `NEOKAI_WORKSPACE_PATH` before overwriting, and restore it after `daemon.cleanup()` in the `afterEach` block.

Additionally the per-daemon workspace directories (`/tmp/crash-recovery-ws-*`) are created inside `createDaemonWithSharedDb` but are **never deleted**. `afterEach` only removes `tmpDir` (the DB directory). Each test run leaks two workspace directories on disk.

---

### P1 — GitHub poll test does not handle the case where `gitHubService` is `null`

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 327–332 and 354–358

```ts
const pollingService1 = daemon1.gitHubService?.getPollingService();
if (pollingService1) {
    pollingService1.triggerPoll = async () => {};
}
```

In CI, `createDaemonWithSharedDb` calls `createDaemonApp` without setting `ANTHROPIC_BASE_URL` or `ANTHROPIC_API_KEY` (unlike `createInProcessDaemonServer`, which configures dev-proxy credentials). Looking at `app.ts` lines 216–241, `gitHubService` is only created when `authStatus.isAuthenticated` is `true`. In the CI `features-3` shard, `ANTHROPIC_API_KEY` is set to `sk-devproxy-test-key` by the workflow env block, so authentication should pass — but this depends on `AuthManager.getAuthStatus()` accepting that dummy key rather than validating it against the Anthropic API. If authentication fails for any reason, `gitHubService` remains `null`, `pollingService1` is `undefined`, and the stub is silently never applied.

The more critical consequence: `waitForQueueJob` at line 335 then polls for a `completed` github.poll job that will **never arrive** (the poll chain is never started because `gitHubService` was null), causing the test to time out rather than fail fast with a meaningful message. A guard `expect(daemon1.gitHubService).not.toBeNull()` before line 327 would surface this immediately.

---

### P2 — `assertEagerlyReclaimed` has a logic gap: it does not assert when the job is `pending`

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts`, lines 156–175

The function verifies reclamation by checking `startedAt` freshness when the job is still `processing`. But if the job is `pending` (reclaimed but not yet dequeued), it falls through to "any other status also proves reclamation" with no assertion. A job could hypothetically remain in stale `processing` and not appear in the `listJobs` result for `pending` if there were a query filter issue — this code path effectively becomes a no-op. The function should include an explicit `expect` for the `pending` case (e.g. asserting `job.startedAt` is `null`) rather than silently passing.

---

### P2 — GitHub poll test: `triggerPoll` stub is applied after daemon startup, not before

**File:** lines 325–332

`createDaemonWithSharedDb` fully starts the daemon (including `jobProcessor.start()` and the initial tick). By the time the test applies the `triggerPoll` stub, the job processor may have already dequeued and started executing the initial `github.poll` job. While the comment at line 327 says the stub avoids "real GitHub network calls" and the handler is safe even without the stub (no repos registered), the test relies on this side-effect tolerance rather than proper isolation. Passing a stub-factory into `createDaemonWithSharedDb` — or delaying `gitHubService.start()` — would be cleaner, though this is a low-severity concern given the safety net described in `github-poll-job.test.ts`.

---

### P3 — Minor: workspace leak in `createDaemonWithSharedDb` vs. `tmpDir` cleanup scope

`tmpDir` is the DB directory. The per-daemon workspace dirs (`/tmp/crash-recovery-ws-*`) are intentionally separate (to avoid worktree conflicts), but since there are two daemons per test, each test run permanently leaks two `/tmp/crash-recovery-ws-*` directories. In a long-lived CI runner this accumulates. These should be tracked and cleaned in `afterEach`.

---

### What works well

- The two-daemon shared-DB pattern cleanly simulates the crash scenario without forking processes.
- The timing-based "must complete within 10 s" approach for proving eager reclamation is pragmatic and correct given the 60 s periodic check interval.
- The `insertStaleCrashedJob` helper correctly inserts all required columns and uses a STALE_AGE_MS well above the 5-minute threshold.
- The `job_queue.cleanup` test has the most thorough assertions (checks `deletedJobs` type and `nextRunAt` schedule).
- CI matrix and validation script are both updated in sync.

lsm (commented):

**Recommendation: APPROVE**

The three issues raised in the prior review have been properly fixed in commit `fbf40690`:

1. **Env var leak in `createDaemonWithSharedDb` (was P1)** -- Fixed. `NEOKAI_WORKSPACE_PATH` and `TEST_WORKTREE_BASE_DIR` are now saved in `beforeEach` and restored in `afterEach` using the standard save/restore pattern (lines 229-230, 273-282). This prevents cross-test pollution.

2. **Null guard for `gitHubService` in GitHub poll test (was P1)** -- Fixed. `expect(daemon1.gitHubService).not.toBeNull()` guard added on both daemon1 (line 365) and daemon2 (line 395), and the optional chaining `?.` replaced with non-null assertion `!.` (lines 368, 398). This ensures immediate failure with a clear message if authentication is broken rather than a silent timeout.

3. **`assertEagerlyReclaimed` logic gap for pending status (was P2)** -- Fixed. The `pending` case now explicitly asserts `expect(job!.startedAt).toBeNull()` (line 170), confirming that `reclaimStale()` cleared `started_at` when resetting the job to pending.

### Additional fix: workspace directory cleanup

Workspace directories created by `createDaemonWithSharedDb` are now tracked via the `createdWorkspaceDirs` array (line 196) and cleaned up in `afterEach` (lines 297-299). This resolves the P3 disk leak noted in both prior reviews.

- The `triggerPoll` stub race window noted in the first review still exists (stub applied after `createDaemonApp` returns, so the initial poll may execute with the real `triggerPoll`). This is benign as documented -- the fake token and no-repos-registered state prevent any real API calls -- but worth keeping in mind if the GitHub polling handler's error handling changes in the future.

No blocking issues. The code is clean, well-documented, and the fixes are correct.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P1 issues remain)

---

The worker addressed all three items from the previous review (env var save/restore for NEOKAI_WORKSPACE_PATH and TEST_WORKTREE_BASE_DIR, null guard for gitHubService, and the pending-status assertion in assertEagerlyReclaimed). The core test logic is sound. However, two P1 issues were introduced or left unaddressed in the fix.

## P1 - TEST_WORKTREE_BASE_DIR directory is set but never cleaned up

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` lines 203-205

```ts
if (!process.env.TEST_WORKTREE_BASE_DIR) {
    process.env.TEST_WORKTREE_BASE_DIR = `/tmp/daemon-worktrees-${Date.now()}`;
}
```

The env var is correctly saved and restored in afterEach, but the `/tmp/daemon-worktrees-*` **directory itself is never created here and never tracked** — however, the worktree manager may create it lazily under the hood. More critically: when the condition on line 203 fires (first test run), the generated path is unique per-call but is never added to `createdWorkspaceDirs` or any cleanup list. If the daemon ever materializes files under this path, those files will accumulate across CI runs. The fix should either add this path to `createdWorkspaceDirs` (which is cleaned up in afterEach) or move the initialization to beforeEach where the path can be tracked alongside tmpDir.

## P1 - NEOKAI_SDK_STARTUP_TIMEOUT_MS is mutated but never saved/restored

**File:** `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` lines 200-202

```ts
if (!process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS) {
    process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS = '30000';
}
```

The beforeEach block (lines 247-250) saves NEOKAI_WORKSPACE_PATH, TEST_WORKTREE_BASE_DIR, GITHUB_POLLING_INTERVAL, and GITHUB_TOKEN. NEOKAI_SDK_STARTUP_TIMEOUT_MS is conspicuously absent from this save/restore cycle. If the env var was not set before the first test, it will be set to '30000' and remain set for all subsequent tests in the process — including tests in other files run in the same Bun worker. The same pattern exists in daemon-server.ts, but that helper runs in test-scoped processes. Here the mutation is inlined into a helper that shares the process with potentially many tests.

To fix: add `savedSdkStartupTimeout: string | undefined` to the suite state, save it in beforeEach, and restore it in afterEach (following the exact same pattern used for the other four vars).

## P2 - start() comment overstates synchrony

**File:** lines 16-24 (file header comment) and lines 332-334 (test body)

The header comment says:
> tick() immediately dequeues pending jobs (SQLite transaction), so by the time createDaemonApp() returns, the stale job has already been cycled: stale-processing → pending → (fresh-)processing

This is correct about the SQLite dequeue (synchronous), but the comment continues to imply execution completes before createDaemonApp() returns. In reality, processJob() is async and not awaited inside tick(). The job will be in 'processing' state after createDaemonApp() returns, not necessarily completed. This is fine for the test (assertEagerlyReclaimed handles all states including processing), but the comment in assertEagerlyReclaimed saying 'Both transitions happen synchronously inside start(), before createDaemonApp() returns' (line 151-152) is inaccurate: the dequeue is synchronous, but execution is not. The comment should clarify that execution is async (hence the waitForJobById polling loop).

## P3 - assertEagerlyReclaimed silently passes when job is already completed

**File:** lines 156-177

When the stale job has already reached completed/failed/dead by the time assertEagerlyReclaimed is called, the function has no assertion — it exits silently. The comment at line 176 ('no further assertion needed') is logically sound since the subsequent waitForJobById will catch failures, but it means assertEagerlyReclaimed does not actually verify reclamation in the fast path. Consider adding `expect(job!.startedAt).not.toEqual(staleCutoff)` or similar to confirm startedAt was refreshed even in terminal states, though this is minor given the overall test structure.

---

Summary of what was fixed correctly:
- NEOKAI_WORKSPACE_PATH and TEST_WORKTREE_BASE_DIR env var save/restore in beforeEach/afterEach: correct.
- createdWorkspaceDirs tracking and cleanup loop in afterEach: correct.
- GITHUB_POLLING_INTERVAL and GITHUB_TOKEN save/restore: correct.
- gitHubService null guard using expect().not.toBeNull(): correct — fails fast with a clear message instead of a confusing timeout.
- assertEagerlyReclaimed pending case: correct — startedAt is asserted null, matching the reclaimStale() SQL.

lsm (commented):

**Recommendation: APPROVE**

### Third-iteration re-review: All previous feedback addressed

The three issues raised in the last review have been properly fixed in commit `0b66f2ed`:

1. **NEOKAI_SDK_STARTUP_TIMEOUT_MS save/restore (was P1)** -- Fixed. `savedSdkStartupTimeout` is saved in `beforeEach` (line 248) and restored in `afterEach` (lines 290-294) using the same save/restore pattern as all other env vars.

2. **TEST_WORKTREE_BASE_DIR directory cleanup (was P1)** -- Fixed. Initialization moved from `createDaemonWithSharedDb` to `beforeEach` (lines 253-258). The generated path is now tracked in `createdWorkspaceDirs` (line 257) so `afterEach` cleans it up alongside other workspace directories.

3. **Comment accuracy about sync vs async (was P2)** -- Fixed. The file header comment (lines 14-26) now correctly states: "tick() dequeues pending jobs via a synchronous SQLite transaction, moving the job back to 'processing' before start() returns -- but the actual job execution (processJob) is async and runs in the background." This accurately distinguishes the synchronous dequeue from the async execution.

### Full verification

- All five env vars mutated by the suite (`NEOKAI_WORKSPACE_PATH`, `TEST_WORKTREE_BASE_DIR`, `NEOKAI_SDK_STARTUP_TIMEOUT_MS`, `GITHUB_POLLING_INTERVAL`, `GITHUB_TOKEN`) are saved in `beforeEach` and restored in `afterEach`.
- Workspace directories are tracked and cleaned up in `afterEach`.
- `assertEagerlyReclaimed` has explicit assertions for both `pending` (startedAt is null) and `processing` (startedAt is recent) states.
- `gitHubService` null guards use `expect().not.toBeNull()` for fast, clear failure.
- CI matrix validation script passes with the new `features-3` shard.
- `NODE_ENV = 'test'` is set idempotently (already set by test infrastructure) and does not need save/restore.

No remaining issues.

lsm (commented):

### Recommendation: APPROVE

This is a third-iteration review. The three fixes requested in previous rounds are correctly applied. The overall quality of the test file is high.

---

### What was fixed (verified correct)

**1. NEOKAI_SDK_STARTUP_TIMEOUT_MS save/restore** — `beforeEach` now saves `savedSdkStartupTimeout` before any mutation; `afterEach` restores it with the same delete-vs-assign pattern used for other env vars (lines 246, 291-295). Correct.

**2. TEST_WORKTREE_BASE_DIR initialization moved to `beforeEach`** — The worktree directory is now created inside `beforeEach` (lines 253-258) so `createdWorkspaceDirs.push(worktreeDir)` is tracked before any daemon starts. `afterEach` will delete it on cleanup. Correct.

**3. Updated comment accuracy** — The header comment (lines 14-26) now accurately states that `tick()` dequeues synchronously via SQLite transaction but `processJob` runs asynchronously in the background. This matches the implementation in `job-queue-processor.ts` where `start()` calls `this.tick()` without `await` (line 45 of the processor).

---

### Remaining observations (no blockers)

**P3 — GitHub test: triggerPoll stub has a narrow pre-stub window**

In the `github.poll chain` test (line 374), `createDaemonWithSharedDb` is called and the `jobProcessor.start()` call inside `createDaemonApp` fires `tick()` immediately without `await`. There is a brief window between `createDaemonApp` returning and the test applying `pollingService1.triggerPoll = async () => {}` (lines 381-384) during which the async `processJob` could call the real `triggerPoll`. In practice this is benign because:

- `triggerPoll` with no repositories added is a no-op (it calls `pollAllRepositories` which iterates an empty `Map`).
- The test uses a fake token so even a real call would 401 quietly and the handler catches errors internally.

This is an existing pattern in the companion `github-poll-job.test.ts` and is acceptable. No change needed.

**P3 — `assertEagerlyReclaimed` passes silently when job is already in a final state**

When a job reaches `completed`/`failed`/`dead` before the assertion runs, the function does no structural check beyond `expect(job).toBeDefined()`. This is intentional (comment on line 178 explains it), but it means a fast execution could technically skip the reclamation invariant check. The subsequent `waitForJobById` with a 10-second timeout provides an indirect timing proof, so the overall test logic is sound.

**P3 — `NEOKAI_SDK_STARTUP_TIMEOUT_MS` conditional set in `createDaemonWithSharedDb`**

The helper only sets `NEOKAI_SDK_STARTUP_TIMEOUT_MS` when it is not already present (line 202-204). Since `beforeEach` now saves the original value (which may be `undefined`), and the helper sets it only when absent, `afterEach` will correctly delete it after the test. No bug — the logic is consistent.

---

### Summary

All three previously-requested fixes are implemented correctly. The test structure is clean: proper env save/restore, isolated per-test SQLite databases, best-effort daemon cleanup, and clear timing-based reclamation proofs. The CI matrix correctly places the new file in the `features-3` shard with `mock_sdk: true`. The validate-online-test-matrix script is updated to include `job-queue-crash-recovery.test.ts`. No P0/P1/P2 issues found.

lsm:
Superseded by https://github.com/lsm/neokai/pull/673

PR #675, opened 2026-03-22
feat: step agent peer communication tools (send_feedback, request_peer_input, list_peers)

lsm:
## Summary

- Adds `send_feedback(target, message)` MCP tool to step agents — channel-validated direct messaging with support for point-to-point, broadcast (`'*'`), and multicast (`string[]`). Validates against `ChannelResolver` before routing; returns clear error with available channels and suggests `request_peer_input` on failure. Enforces hub-spoke and one-way topology.
- Adds `request_peer_input(targetRole, question)` MCP tool — async/non-blocking fallback that routes through the Task Agent when no direct channel is declared. Peer response arrives as a user turn prefixed `[Peer response from {role}]: ...`.
- Adds `list_peers()` MCP tool — returns other group members (excluding self and task-agent) with role, status, sessionId, and permitted channel connections.
- `buildCustomAgentSystemPrompt()` updated with a "Peer Communication" section explaining the communication model, tool selection rules, and async nature of `request_peer_input`.
- `TaskAgentManager.buildStepAgentMcpServerForSession()` wires all peer tool dependencies (messageInjector, injectToTaskAgent, sessionGroupRepo, workflowRunRepo) and attaches the MCP server to each spawned sub-session via `buildStepAgentMcpServer` factory in `TaskAgentToolsConfig`.

## Test plan

- [x] 27 new unit tests in `step-agent-tools.test.ts` covering:
  - `list_peers`: excludes self + task-agent, reports topology, handles missing group
  - `send_feedback`: point-to-point, broadcast, multicast, hub-spoke enforcement, bidirectional, unauthorized direction, partial injection failures, no-target-found
  - `request_peer_input`: routing format, async acknowledgment, Task Agent injection failure, fallback with no channels, non-existent role
  - MCP server factory creation
  - System prompt includes Peer Communication section
- [x] All existing space unit tests (1285 tests) continue to pass
- [x] All pre-commit checks passed (lint, format, typecheck, knip)

### DIFF
--- a/packages/daemon/src/lib/space/agents/custom-agent.ts
+++ b/packages/daemon/src/lib/space/agents/custom-agent.ts
@@ -149,6 +149,49 @@ export function buildCustomAgentSystemPrompt(customAgent: SpaceAgent): string {
 			`If you need to modify any files, follow the normal git/PR workflow instead.`
 	);
 
+	// Peer communication model
+	sections.push(`\n## Peer Communication\n`);
+	sections.push(
+		`You are part of a multi-agent team within this workflow step. ` +
+			`You have MCP tools for communicating with peer agents in the same group.`
+	);
+	sections.push(`\n### Primary: \`send_feedback\` (channel-validated direct messaging)\n`);
+	sections.push(
+		`Use \`send_feedback\` to send messages directly to permitted peers based on the declared channel topology.`
+	);
+	sections.push(
+		`- \`target: 'role'\` — point-to-point to a specific role (e.g., \`'coder'\`)\n` +
+			`- \`target: '*'\` — broadcast to all permitted targets\n` +
+			`- \`target: ['role1', 'role2']\` — multicast to multiple roles`
+	);
+	sections.push(
+		`This tool validates against declared channels. ` +
+			`If the channel is not declared, it returns an error with available channels and suggests \`request_peer_input\`.`
+	);
+	sections.push(`\n### Fallback: \`request_peer_input\` (Task Agent mediated)\n`);
+	sections.push(
+		`Use \`request_peer_input\` when no direct channel is declared or as a fallback when \`send_feedback\` fails validation.`
+	);
+	sections.push(
+		`This is **async and non-blocking** — the tool returns immediately with an acknowledgment. ` +
+			`The peer's answer will arrive as a separate user turn prefixed with: \`[Peer response from {role}]: ...\``
+	);
+	sections.push(
+		`**Do NOT wait for an immediate reply.** Continue your work and handle peer responses when they arrive.`
+	);
+	sections.push(`\n### Discovering peers: \`list_peers\`\n`);
+	sections.push(
+		`Use \`list_peers\` to see all other agents in this step's group, their roles, statuses, ` +
+			`and permitted outgoing channels for \`send_feedback\`.`
+	);
+	sections.push(`\n### Communication model rules\n`);
+	sections.push(
+		`- If this step has declared channels: use \`send_feedback\` for permitted directions, ` +
+			`\`request_peer_input\` for undeclared directions\n` +
+			`- If this step has no declared channels: all communication goes through \`request_peer_input\`\n` +
+			`- All communication is scoped to this group — you cannot message agents in other tasks`
+	);
+
 	// Review feedback handling
 	sections.push(`\n## Addressing Review Feedback\n`);
 	sections.push(
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -66,6 +66,7 @@ import type {
 	SubSessionState,
 } from '../tools/task-agent-tools';
 import { createTaskAgentMcpServer } from '../tools/task-agent-tools';
+import { createStepAgentMcpServer } from '../tools/step-agent-tools';
 import { createTaskAgentInit, buildTaskAgentInitialMessage } from '../agents/task-agent';
 import { Logger } from '../../logger';
 import { SpaceTaskManager } from '../managers/space-task-manager';
@@ -270,10 +271,12 @@ export class TaskAgentManager {
 			const runtime = await this.config.spaceRuntimeService.createOrGetRuntime(spaceId);
 			const subSessionFactory = this.createSubSessionFactory(taskId, spaceId);
 
+			const workflowRunId = workflowRun?.id ?? '';
+
 			const mcpServer = createTaskAgentMcpServer({
 				taskId,
 				space,
-				workflowRunId: workflowRun?.id ?? '',
+				workflowRunId,
 				workspacePath: space.workspacePath,
 				runtime,
 				workflowManager: this.config.spaceWorkflowManager,
@@ -289,6 +292,14 @@ export class TaskAgentManager {
 				sessionGroupRepo: this.config.sessionGroupRepo,
 				getGroupId: () => this.taskGroupIds.get(taskId),
 				daemonHub: this.config.daemonHub,
+				buildStepAgentMcpServer: (subSessionId, role) =>
+					this.buildStepAgentMcpServerForSession(
+						taskId,
+						subSessionId,
+						role,
+						spaceId,
+						workflowRunId
+					) as unknown as McpServerConfig,
 			});
 
 			// setRuntimeMcpServers expects McpServerConfig but the MCP SDK's `Server`
@@ -1078,10 +1089,12 @@ export class TaskAgentManager {
 		const runtime = await this.config.spaceRuntimeService.createOrGetRuntime(spaceId);
 		const subSessionFactory = this.createSubSessionFactory(taskId, spaceId);
 
+		const rehydrateWorkflowRunId = workflowRun?.id ?? '';
+
 		const mcpServer = createTaskAgentMcpServer({
 			taskId,
 			space,
-			workflowRunId: workflowRun?.id ?? '',
+			workflowRunId: rehydrateWorkflowRunId,
 			workspacePath: space.workspacePath,
 			runtime,
 			workflowManager: this.config.spaceWorkflowManager,
@@ -1097,6 +1110,14 @@ export class TaskAgentManager {
 			daemonHub: this.config.daemonHub,
 			sessionGroupRepo: this.config.sessionGroupRepo,
 			getGroupId: () => this.taskGroupIds.get(taskId),
+			buildStepAgentMcpServer: (subSessionId, role) =>
+				this.buildStepAgentMcpServerForSession(
+					taskId,
+					subSessionId,
+					role,
+					spaceId,
+					rehydrateWorkflowRunId
+				) as unknown as McpServerConfig,
 		});
 
 		agentSession.setRuntimeMcpServers({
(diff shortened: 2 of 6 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured implementation of step agent peer communication tools. The code is clean, well-documented, and follows existing patterns in the codebase. The test coverage is thorough with 27 tests covering happy paths, error cases, and edge cases. Here are my detailed findings:

### What looks good

- **Channel validation via `ChannelResolver.canSend()`** is correctly integrated. Hub-spoke isolation, one-way enforcement, and bidirectional channels all work as designed.
- **Separation of concerns**: Tool handlers are pure functions, dependencies are injected, and the MCP server factory is a thin wrapper. This makes testing straightforward.
- **Error handling**: All three tools handle failures gracefully -- partial injection failures in `send_feedback`, Task Agent unavailability in `request_peer_input`, and missing groups in all tools.
- **The `as unknown as McpServerConfig` cast** follows the existing pattern used for the task-agent MCP server (lines 311, 1124 of `task-agent-manager.ts`).
- **System prompt** clearly explains the communication model, tool selection rules, and async nature of `request_peer_input`.

### P2: `send_feedback` does not prefix messages with sender identity

In `step-agent-tools.ts` line 296, the raw `message` is passed directly to `messageInjector` without any sender identification:
```typescript
await messageInjector(member.sessionId, message);
```

In contrast, `request_peer_input` (line 342) formats the message with sender role and session ID. When a step agent receives a message via `send_feedback`, it has no way to distinguish who sent it, which could be confusing in multi-peer scenarios. Consider prefixing with something like `[Feedback from {myRole}]: {message}` to maintain attribution.

### P3: No minimum length validation on `message` / `question` fields

`SendFeedbackSchema.message` and `RequestPeerInputSchema.question` are `z.string()` without `.min(1)`. An empty string would pass Zod validation and be injected as a user turn. This is minor since an LLM is unlikely to send empty strings, but a `.min(1)` guard would be cheap protection.

### P3: Asymmetric behavior of `send_feedback` with empty topology

When no channels are declared (`resolver.isEmpty()`):
- **Broadcast (`*`)**: fails with "No channel topology declared" (line 228-235) -- correct
- **Point-to-point / multicast**: skips validation entirely and succeeds (line 256 `if (!resolver.isEmpty())`) -- permissive

This is tested and documented in the test at line 347-358 ("returns error when no channels declared at all" -- which actually asserts `success: true`). The asymmetry is intentional but the test comment on line 351 ("Let's verify the actual behavior") suggests this was discovered rather than designed. If this is the intended semantics, consider adding a comment in the handler explaining why point-to-point bypasses topology when no channels are declared.

### Summary

Clean, well-tested implementation that follows codebase conventions. The sender identity gap in `send_feedback` (P2) is the most meaningful issue -- without it, receiving agents cannot attribute messages in multi-peer groups. The other items are minor.

lsm (commented):

The implementation is solid and well-structured — the dependency injection pattern, ChannelResolver integration, and 27-test suite are all good. However there is one meaningful behavioral bug and one doc/comment inconsistency that should be resolved before merge.

---

### P1 — Behavioral inconsistency: `send_feedback` with empty topology allows unrestricted messaging (point-to-point and multicast)

**Files:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts` lines 255-271

The validation block at line 256 only runs when `!resolver.isEmpty()`. When no channel topology is declared the guard is skipped entirely, so a point-to-point call like `send_feedback({ target: 'reviewer', message: '...' })` succeeds and injects into any peer in the group with no restrictions.

This contradicts three places in the codebase that all say the opposite:

1. Module docstring (line 23-24): _"When no channels are declared for a step, all `send_feedback` calls fail with a suggestion to use `request_peer_input` instead."_
2. System prompt in `buildCustomAgentSystemPrompt()` (custom-agent.ts line 191): _"If this step has no declared channels: all communication goes through `request_peer_input`"_
3. The test at line 347-358 has a comment that acknowledges this but frames it as "actual behavior" rather than a bug.

The inconsistency is significant because:
- An agent author who declares no channels expects no direct inter-agent messaging to happen (they want Task Agent mediation for everything).
- Only the `target: '*'` (broadcast) path correctly blocks when the topology is empty — single-role and array targets bypass the guard.

Either the code should be changed to block when `resolver.isEmpty()` (matching the docs and system prompt), or the docs and system prompt should be updated to reflect the intentional open-send behavior — and the test comment rewritten accordingly. The current state leaves an important question unanswered for the next developer reading this.

---

### P2 — Doc comment inconsistency: `messageInjector` described as serializing writes but no serialization mechanism is visible

**File:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts` line 29

The design note says: _"The `messageInjector` serializes writes per-session for safe concurrency."_ But the injector is a plain `(sessionId: string, message: string) => Promise\<void\>` callback with no explicit serialization primitive (mutex, queue, etc.) in the config type or the wiring in `buildStepAgentMcpServerForSession()`. The serialization, if it exists, is inside `injectSubSessionMessage` → `injectMessageIntoSession` → `session.messageQueue.enqueueWithId()`, which is not obvious from the doc. The comment should either be removed or clarified to point at the session's message queue.

---

### P2 — `loadGroupAndResolver` return type uses a non-null assertion (`group!`) after callers already check `error`

**File:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts` lines 171, 217, 275

After `loadGroupAndResolver()` returns without `error`, the group is typed as `ReturnType\<typeof sessionGroupRepo.getGroup\>` which includes `null`. Code uses `group!` (non-null assertion) at lines 171, 275 etc. This compiles fine but loses type safety. The return type could narrow `group` to the non-null type when no `error` is present, or the helper could throw on missing group rather than returning the sentinel pattern.

---

### P3 — Typo in inline comment

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts` line 321

`/ Group creation is non-fatal` is missing the second `/` at the start — it reads as a division operator rather than a comment. This is likely a copy-paste artifact. Should be `// Group creation is non-fatal`.

---

### P3 — `request_peer_input` does not validate `target_role` against group membership

This is explicitly noted as intentional in the test (line 700-716), but the system prompt tells agents to use `list_peers` before communicating. A non-existent role will cause the Task Agent to receive a routing request for a role that doesn't exist, silently succeeding from the tool's perspective. At minimum a note in the tool description (or a `list_peers` suggestion in the acknowledgment) would help prevent agent confusion in that case.

---

### General observations (non-blocking)

- The test suite is thorough and covers the important topology patterns (hub-spoke isolation, bidirectional, multicast, partial injection failure). Good coverage.
- The `createStepAgentMcpServer` factory test only checks that the return value is a non-null object — it does not verify the tool names or handlers are attached. A slightly deeper assertion (e.g., checking `server.server` exists) would be more meaningful, though this is minor.
- The rehydration path correctly rebuilds the `subSessions` map from step tasks so that `send_feedback` messages injected after a daemon restart will find sessions in the map and call `ensureQueryStarted()` on them. This is not tested but appears to work based on the AgentSession lifecycle.

lsm (commented):

### Re-review: All prior issues addressed correctly

I verified each fix from the prior review:

**P1 fix (send_feedback blocks all target forms when resolver.isEmpty())** -- Confirmed. The `resolver.isEmpty()` check at `step-agent-tools.ts:225` is positioned before any target resolution logic, so it blocks point-to-point, multicast, and broadcast uniformly. Tests at lines 347-403 cover all three forms with empty topology.

**P2 fix (sender attribution prefix)** -- Confirmed. Line 296: `[Feedback from ${myRole}]: ${message}`. Test at line 329 verifies the exact format.

**P2 fix (discriminated union for loadGroupAndResolver)** -- Confirmed. Lines 110-153 return `{ ok: true, group, resolver, groupId }` or `{ ok: false, error }`. All three callers check `loaded.ok` before accessing fields. No non-null assertions remain.

**P2 fix (per-session write ordering doc comment)** -- Confirmed at module doc line 30.

**P3 fix (.min(1) on message/question, "call list_peers first" note)** -- Confirmed at schema lines 65, 95, and MCP tool description at line 412.

**P3: Unused `spaceId` parameter** in `buildStepAgentMcpServerForSession` (`task-agent-manager.ts:1322`). The parameter is accepted but never referenced in the method body. Not a bug, but dead code.

### Overall assessment

The implementation is clean, well-tested (27 tests, all passing), and correctly addresses all prior feedback. The code follows existing patterns (discriminated unions, `as unknown as McpServerConfig` casts matching the existing task-agent pattern, separated handlers for testability). Test coverage is thorough -- including edge cases for partial injection failures, hub-spoke topology enforcement, empty topology blocking, and bidirectional channels.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

All P0/P1/P2 issues from the previous review have been correctly addressed. The implementation is solid.

---

## Previously Fixed Issues — Verified

**P1: send_feedback now blocks ALL target forms when topology is empty**

Confirmed. The `resolver.isEmpty()` guard at line 225 of `step-agent-tools.ts` fires before any target-type branching (point-to-point / multicast / broadcast), making it impossible to bypass via any target form. Test coverage at lines 347-358 (empty topology blocks point-to-point) and 394-404 (empty topology blocks broadcast) confirms this.

**P2: Message prefix for sender attribution**

`[Feedback from ${myRole}]: ${message}` is correctly applied at line 296 before injection. Test at line 329 verifies the exact prefix format.

**P2: loadGroupAndResolver returns discriminated union**

The `{ ok: true, group, resolver, groupId } | { ok: false, error }` pattern is correctly implemented at lines 110-154. Both `list_peers` and `send_feedback` guard with `if (!loaded.ok) return loaded.error` before destructuring. No non-null assertions remain.

**P2: Doc comment clarified**

The comment at line 30 now reads "per-session write ordering" — accurate.

**P3: .min(1) added and list_peers note added**

`message` (line 65) and `question` (lines 94-95) both have `.min(1)`. The `request_peer_input` description at line 410-412 includes "Call list_peers first to verify the target_role exists in the group."

---

## Remaining Minor Issues

**P3: `target_role` field in `RequestPeerInputSchema` is missing `.min(1)`**

`packages/daemon/src/lib/space/tools/step-agent-tool-schemas.ts`, line 89-91

```ts
target_role: z
  .string()
  .describe("Role of the peer to request input from (e.g., 'reviewer', 'coder')"),
```

The fix applied `.min(1)` to `message` and `question` but not to `target_role`. An empty string role would produce a nonsensical routing message to the Task Agent ("to ''") but would not cause crashes or data loss. Consistency with the other string fields argues for adding `.min(1)` here as well.

**P3: Empty multicast array (`target: []`) produces misleading error**

If a model passes `target: []`, the authorization check passes vacuously (empty array has no unauthorized roles), and the code returns "No active sessions found for target role(s): " with an empty role list. A `z.array(z.string()).min(1)` constraint on the array branch of the `target` union, or an explicit length check before the authorization pass, would produce a cleaner error. This is not a security issue — the behavior is safe.

---

## Architecture Assessment

The integration wiring in `buildStepAgentMcpServerForSession` (task-agent-manager.ts lines 1318-1337) is correct. The `messageInjector` routes to `injectSubSessionMessage` (not `injectTaskAgentMessage`), and `injectToTaskAgent` correctly calls `injectTaskAgentMessage`. The separation between peer-to-peer and Task-Agent-mediated paths is enforced at the wiring level.

The 27 unit tests cover the meaningful behavioral paths. The discriminated union is tested implicitly via the existing "group not found" test cases.

Overall quality is high. The two remaining nits are minor schema validation gaps that do not affect correctness or security.

PR #676, opened 2026-03-22
test: add job queue crash recovery integration tests

lsm:
## Summary

- Adds `packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` with four crash-recovery integration tests covering all job types
- Each test simulates a daemon crash by inserting a stale `processing` job (started_at = 6 min ago) into a **file-backed SQLite database**, shutting down daemon1, then starting daemon2 against the same DB file
- Verifies that `JobQueueProcessor.start()` eagerly calls `reclaimStale()` (not waiting for the 60-second periodic check) by asserting jobs reach terminal state within 15 s
- Updates CI matrix (`features-2` shard) and `scripts/validate-online-test-matrix.sh` to include the new test file

## Test scenarios

| Scenario | Key assertions |
|---|---|
| `session.title_generation` | Job survives restart; reaches terminal state within 15 s (any terminal state accepted since no real session exists) |
| `github.poll` | Job completed; self-scheduling chain resumes with next job at +300 s |
| `room.tick` | Job completes with `{skipped:true, reason:'not running'}`; no follow-up tick scheduled |
| `job_queue.cleanup` | Job completes; next cleanup self-scheduled ≥ 23 h in the future |

## Test plan

- [x] `NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts` — all 4 tests pass in ~10s
- [x] `bash scripts/validate-online-test-matrix.sh` — no errors
- [x] `bun run typecheck` — no errors
- [x] `bun run lint` — no errors

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -180,6 +180,7 @@ jobs:
           - module: features-2
             test_path: >-
               tests/online/features/github-poll-job.test.ts
+              tests/online/features/job-queue-crash-recovery.test.ts
               tests/online/features/message-persistence.test.ts
             mock_sdk: true
           - module: git
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -183,6 +183,7 @@ jobs:
               tests/online/features/job-queue-crash-recovery.test.ts
               tests/online/features/message-persistence.test.ts
             mock_sdk: true
+            timeout: 8
           - module: git
             test_path: tests/online/git
             timeout: 10
--- a/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+++ b/packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
@@ -1,0 +1,459 @@
+/**
+ * Online test: Job queue crash recovery
+ *
+ * Verifies that all four job types survive a daemon crash/restart cycle:
+ *
+ *  1. session.title_generation — stale processing job is reclaimed and processed
+ *  2. github.poll             — stale processing job is reclaimed; self-scheduling chain resumes
+ *  3. room.tick               — stale processing job is reclaimed; handler skips gracefully
+ *  4. job_queue.cleanup       — stale processing job is reclaimed and completes
+ *
+ * ## Crash simulation pattern (used in every test)
+ *
+ * 1. daemon1 starts with a **file-backed** SQLite database (shared `dbPath`)
+ * 2. A job is inserted directly into the DB with status='processing' and
+ *    started_at set ~6 minutes ago (exceeding the 5-minute stale threshold)
+ * 3. daemon1 is cleanly shut down (`cleanup()`) — simulating a process that
+ *    crashed before the job could finish
+ * 4. daemon2 is started against the **same database file**
+ * 5. `JobQueueProcessor.start()` eagerly calls `reclaimStale()` before the
+ *    first poll tick — the stale job moves from 'processing' back to 'pending'
+ *    immediately, without waiting for the 60-second periodic stale check
+ * 6. Assertions verify: (a) the job was not lost, (b) it reaches a terminal
+ *    state quickly (< 15 s), proving eagerness vs the 60-second fallback
+ *
+ * ## No real Anthropic API calls
+ *
+ * - session.title_generation: no real session exists → handler throws → job
+ *   retries and eventually becomes 'dead'; the test accepts any terminal state
+ * - github.poll: no repositories are registered → triggerPoll() is a no-op;
+ *   the handler completes successfully and self-schedules
+ * - room.tick: no runtime exists for the fake roomId → handler returns
+ *   { skipped: true, reason: 'not running' } → job completes
+ * - job_queue.cleanup: deletes old rows from the job_queue table — no external
+ *   calls needed
+ *
+ * Run:
+ *   NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/features/job-queue-crash-recovery.test.ts
+ */
+
+import { afterAll, afterEach, beforeAll, describe, expect, test } from 'bun:test';
+import { createDaemonApp, type DaemonAppContext } from '../../../src/app';
+import { getConfig } from '../../../src/config';
+import {
+	GITHUB_POLL,
+	JOB_QUEUE_CLEANUP,
+	ROOM_TICK,
+	SESSION_TITLE_GENERATION,
+} from '../../../src/lib/job-queue-constants';
+import type { Job, JobStatus } from '../../../src/storage/repositories/job-queue-repository';
+
+// ---------------------------------------------------------------------------
+// Configuration
+// ---------------------------------------------------------------------------
+
+const IS_DEV_PROXY = process.env.NEOKAI_USE_DEV_PROXY === '1';
+
+/**
+ * `started_at` that is 6 minutes in the past — comfortably beyond the
+ * 5-minute stale threshold used by JobQueueProcessor.
+ */
+const staleMsAgo = () => Date.now() - 6 * 60 * 1000;
+
+/**
+ * Maximum time to wait for a job to reach a terminal state after daemon2
+ * starts.  Must be well below 60 s (the periodic stale-check interval) so
+ * that a passing test implicitly proves eager reclamation.
+ */
+const TERMINAL_WAIT_MS = 15_000;
+
+/** Polling cadence for job status checks. */
+const POLL_MS = 100;
+
+// ---------------------------------------------------------------------------
+// Shared file-backed database (persists across daemon instances within a test)
+// ---------------------------------------------------------------------------
+
+/**
+ * Unique workspace dir created once for the entire suite.
+ * Each test uses its own sub-directory so DB files do not interfere.
+ */
+const suiteWorkspace = `/tmp/crash-recovery-suite-${Date.now()}-${Math.random().toString(36).slice(2)}`;
+
+/** Env vars replaced in beforeAll and restored in afterAll. */
+const savedEnv: Record<string, string | undefined> = {};
+
+// ---------------------------------------------------------------------------
+// Suite-level setup / teardown
+// ---------------------------------------------------------------------------
+
+beforeAll(async () => {
+	await Bun.$`mkdir -p ${suiteWorkspace}`.quiet();
+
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured and thorough set of integration tests for job queue crash recovery. The tests correctly verify the core behavior described in the task: that stale `processing` jobs are eagerly reclaimed on daemon restart via `reclaimStale()` in `JobQueueProcessor.start()`, rather than waiting for the 60-second periodic check.

### What works well

1. **Crash simulation pattern is sound.** Each test creates daemon1 with a file-backed SQLite DB, inserts a stale job via raw SQL (bypassing the normal enqueue path to simulate a mid-processing crash), shuts down daemon1 (which calls `db.close()` releasing the file lock), then starts daemon2 against the same DB file. This correctly models the crash/restart scenario.

2. **Eagerness verification is clever.** The 15-second `TERMINAL_WAIT_MS` timeout, combined with the 25/20-second test timeouts, implicitly proves that eager reclamation works since the 60-second periodic stale check would never fire within those windows.

3. **Each job type is tested with appropriate expectations:**
   - `session.title_generation`: Accepts any terminal state (`completed`, `failed`, `dead`) since no real session exists -- the key assertion is that the job was reclaimed and processed at all.
   - `github.poll`: Verifies completion AND self-scheduling chain resumption (next job scheduled ~300s in the future).
   - `room.tick`: Verifies graceful skip with `{skipped: true, reason: 'not running'}` and no follow-up tick.
   - `job_queue.cleanup`: Verifies completion with `deletedJobs` in result, and next cleanup self-scheduled 23+ hours out.

4. **Race condition handling in the cleanup test (lines 406-415):** Deleting existing pending cleanup jobs from daemon1 before inserting the stale job prevents daemon2's startup guard from creating a racing cleanup job. The comment on lines 425-428 documents the remaining interaction clearly.

5. **Resource cleanup is correct.** The `afterEach` drains `openDaemons`, and `afterAll` removes `suiteWorkspace` via `rm -rf`. Each test uses isolated sub-directories. Environment variables are saved and restored.

6. **CI integration is correct.** The test is added to the `features-2` shard with `mock_sdk: true`, and the validator script is updated.

### Minor observations (P3)

1. **`TEST_WORKTREE_BASE_DIR` is not cleaned up in `afterAll`.** The suite creates a temp directory at `/tmp/crash-recovery-worktrees-*` (line 116) but only cleans up `suiteWorkspace`. This is a minor leak of empty temp directories -- inconsequential in CI ephemeral environments but worth noting.

2. **`insertStaleJob` sets `run_at` to `staleMsAgo()` (6 minutes ago).** This is fine -- a stale processing job would have had its `run_at` in the past -- but strictly speaking `run_at` could be anything since `reclaimStale` only checks `started_at`. Not a bug, just a cosmetic detail.

3. **The `nextPending[0].runAt` assertion on line 346 assumes ordering.** `listJobs` returns results in DB query order. If daemon2's startup enqueues its own `github.poll` job, `nextPending[0]` might not be the self-scheduled one. In practice this works because the test controls the environment tightly, but a `some()` check (like the cleanup test uses on line 456) would be more robust.

Overall this is clean, well-documented test code that correctly verifies crash recovery semantics with appropriate isolation and cleanup.

lsm (commented):

The implementation is solid. The crash-recovery simulation pattern is correct, the four job handlers are all covered, resource cleanup is handled properly, and the CI matrix + validator script are correctly updated. Below are my findings in priority order.

---

### P1 — CI shard timeout risk

The features-2 shard defaults to 5 minutes (matrix.timeout || 5). It now contains three test files: github-poll-job.test.ts, job-queue-crash-recovery.test.ts, and message-persistence.test.ts. The new file alone starts 8 daemon instances (daemon1 + daemon2 x 4 tests), with individual test timeouts of 20-25 s each. On a loaded CI runner this combination could approach the 5-minute ceiling. Consider adding timeout: 8 (or similar) to the features-2 matrix entry, or split the shard.

---

### P2 — TEST_WORKTREE_BASE_DIR is set but never cleaned up

In beforeAll (line 116), the test sets:
  process.env.TEST_WORKTREE_BASE_DIR = '/tmp/crash-recovery-worktrees-...'

But afterAll only removes suiteWorkspace and restores the savedEnv keys that were explicitly listed. The worktree directory path is never deleted, and because it uses a timestamp it accumulates on CI runners. The env var itself is also not restored. Add it to savedEnv before setting it, and add a rm -rf in afterAll.

---

### P2 — Crash semantics: simulation is a clean shutdown, not a crash

Every test calls await daemon1.cleanup() — a graceful shutdown that stops the job processor cleanly and closes the database. A real crash does none of that. The test correctly exercises the key correctness property (a job stuck in processing state is reclaimed on the next daemon start), but a comment explaining that reclaimStale() is what matters — not WAL recovery — would help future readers understand why the simulation uses a clean shutdown.

---

### P3 — Retry-delay comment is incorrect (line 287)

The comment says '1 s + 2 s = 3 s' but the actual delays from fail() are 2^retryCount * 1000 ms: 1 s + 2 s + 4 s = 7 s total. The 25 s test timeout is still fine; only the comment needs fixing.

---

### What works well

- **Stale-job pattern correctness**: Inserting directly as processing with started_at = 6 min ago, then verifying reclaimStale moves it to pending and it reaches terminal state in < 15 s (well under 60 s periodic check), correctly proves eager reclamation.
- **Handler registration ordering**: All four handlers are registered before jobProcessor.start(), consistent with how app.ts wires them. The tests inherit this correct ordering.
- **github.poll handler registration**: GITHUB_POLLING_INTERVAL=300 and GITHUB_TOKEN in extraEnv satisfy all three conditions in github-service.ts needed to register the handler (pollingInterval > 0, githubToken present, isAuthenticated true via the fake API key set in beforeAll).
- **DB path isolation**: Each test uses a unique sub-directory under suiteWorkspace.
- **Resource cleanup**: openDaemons is populated immediately after createCrashTestDaemon returns — no window where daemon2 leaks on assertion failure. The afterEach 30 s timeout is appropriate.
- **CI matrix consistency**: validate-online-test-matrix.sh and main.yml are in sync.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review after the worker addressed all five issues raised in the previous round. Every item has been properly resolved:

### Previous feedback resolution

1. **P1 -- CI shard timeout** (`.github/workflows/main.yml:186`): `timeout: 8` added to the `features-2` matrix entry. Correct -- 8 minutes gives adequate headroom for 3 test files with 8 daemon starts on a loaded CI runner.

2. **P2 -- TEST_WORKTREE_BASE_DIR cleanup** (lines 120-125, 131-134): The env var is now saved in `savedEnv` before being set, and `afterAll` conditionally removes the directory if it was created by the test (i.e., was previously `undefined`). The restore loop at lines 136-142 handles the env var restoration. Clean implementation.

3. **P2 -- Crash simulation semantics comment** (lines 15-21): The file header now has a thorough explanation of why `cleanup()` (clean shutdown) is used instead of SIGKILL, and clarifies that the correctness property under test is `reclaimStale()` on the persisted DB state, not WAL crash recovery. Future readers will understand the tradeoff.

4. **P3 -- Retry delay comment** (line 300): Corrected from "1 s + 2 s = 3 s" to "1 s + 2 s + 4 s = 7 s" with the clarification that maxRetries=3 means 4 total attempts.

5. **P3 -- github.poll ordering-independent assertion** (lines 355-360): Changed from `nextPending[0].runAt` to `nextPending.some(j => j.runAt > minFutureRunAt)` with a comment explaining why ordering independence matters.

### Current state assessment

No new issues introduced. The fix commit is minimal and surgical -- each change addresses exactly one piece of feedback without touching unrelated code. The test file is well-structured, thoroughly documented, and correctly verifies crash recovery semantics for all four job types.

lsm (commented):

**Recommendation: APPROVE**

This is a follow-up review after the worker addressed all previous feedback. All five reported issues have been correctly resolved with no new issues introduced.

---

### Previous Issues — Verification Status

**P1 resolved: CI shard timeout added**

The `features-2` matrix entry at `.github/workflows/main.yml` line 186 now has `timeout: 8`, which feeds into `timeout-minutes: ${{ matrix.timeout || 5 }}` at line 117. The shard now has an 8-minute CI budget rather than the default 5, which is appropriate given the new test file starts 8 daemon instances with individual timeouts totaling up to ~85 s.

**P2 resolved: TEST_WORKTREE_BASE_DIR is now saved, conditionally set, and cleaned up**

Lines 122–125 in `beforeAll` save the existing value to `savedEnv` before conditionally setting a new temp path. Lines 132–134 in `afterAll` correctly rm -rf the directory only if the test suite created it (condition: `savedEnv['TEST_WORKTREE_BASE_DIR'] === undefined`). The env var is then restored by the generic loop on lines 136–141. The sequence order is correct: directory is deleted before the env var is cleared.

**P2 resolved: Crash simulation semantics are documented**

The file-level block comment (lines 16–21) now explicitly explains that `cleanup()` is used rather than SIGKILL because the test exercises SQLite persistence + `reclaimStale()`, not WAL crash recovery, and that `cleanup()` merely releases the file lock so daemon2 can open the same file.

**P3 resolved: Retry-delay comment corrected**

Line 300 now reads `1 s + 2 s + 4 s = 7 s`, matching the actual exponential back-off delays.

**P3 resolved: github.poll self-scheduling assertion uses `.some()`**

Line 360 now uses `nextPending.some((j) => j.runAt > minFutureRunAt)` instead of asserting on `nextPending[0]`, making the assertion order-independent.

---

### No new issues introduced

- The `savedEnv` key for `TEST_WORKTREE_BASE_DIR` is unconditionally saved on line 122, which means the generic restore loop will always process it. When the value was originally `undefined`, the loop will `delete process.env['TEST_WORKTREE_BASE_DIR']` — correct behavior.
- The conditional rm -rf on line 132 runs before the restore loop on line 136, so `process.env.TEST_WORKTREE_BASE_DIR` is still set when the shell command reads it.
- `validate-online-test-matrix.sh` correctly includes `job-queue-crash-recovery.test.ts` in `FEATURES_FILES`.
- All four test scenarios remain logically sound and continue to correctly verify eager reclamation via the 15 s < 60 s implicit proof.

lsm:
Superseded by https://github.com/lsm/neokai/pull/674

PR #677, opened 2026-03-22
feat: add step agent peer communication tools (channel-validated)

lsm:
## Summary

- Adds three MCP tools for step agents to communicate with peers in the same session group: `send_feedback` (primary, channel-validated direct messaging), `list_peers` (peer discovery with topology info), and `request_peer_input` (Task Agent-mediated fallback for undeclared channels)
- `send_feedback` validates send direction against declared channel topology via `ChannelResolver.canSend()`, supports point-to-point / broadcast (`*`) / multicast (`[role1, role2]`) targets, and returns clear errors suggesting `request_peer_input` on failure
- `request_peer_input` is non-blocking: injects a formatted `[PEER_REQUEST from {role}]` message into the Task Agent session, which relays it; the async response flows back prefixed `[Peer response from {role}]: ...`
- `TaskAgentManager.createSubSessionFactory` now accepts `workflowRunId` and builds/attaches the step agent MCP server before streaming starts
- System prompt in `custom-agent.ts` updated with a "Peer Communication" section documenting all three tools

## Test plan

- [x] 31 new unit tests in `step-agent-tools.test.ts` covering: no-topology rejection, one-way/bidirectional channels, hub-spoke enforcement (spoke-to-spoke blocked), wildcard broadcast, multicast, self-exclusion, injection failure, no-group errors, no-active-sessions, `request_peer_input` message formatting and async acknowledgment, MCP server tool registration
- [x] Full space unit test suite passes (1289 tests, 0 failures)
- [x] TypeScript type check clean
- [x] Lint and formatting clean

### DIFF
--- a/packages/daemon/src/lib/space/agents/custom-agent.ts
+++ b/packages/daemon/src/lib/space/agents/custom-agent.ts
@@ -191,6 +191,20 @@ export function buildCustomAgentSystemPrompt(customAgent: SpaceAgent): string {
 			`- If this step has no declared channels: all communication goes through \`request_peer_input\`\n` +
 			`- All communication is scoped to this group — you cannot message agents in other tasks`
 	);
+	sections.push(`\n### Recognizing inbound peer messages\n`);
+	sections.push(
+		`**\`[Feedback from {role}]: ...\`** — a peer has sent you a direct message via \`send_feedback\`. ` +
+			`Read and incorporate it into your work.`
+	);
+	sections.push(
+		`**\`[Peer response from {role}]: ...\`** — an async reply via the Task Agent ` +
+			`(response to a previous \`request_peer_input\` call). Read and incorporate it.`
+	);
+	sections.push(
+		`**\`[Peer input request from '{role}' ... to '{your-role}']: ...\`** ` +
+			`(where \`{your-role}\` is your own role, e.g. \`${customAgent.role}\`) — ` +
+			`a peer is requesting input from you via the Task Agent. Provide a helpful response.`
+	);
 
 	// Review feedback handling
 	sections.push(`\n## Addressing Review Feedback\n`);
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -269,7 +269,11 @@ export class TaskAgentManager {
 
 			// --- Build and attach MCP server with live runtime dependencies
 			const runtime = await this.config.spaceRuntimeService.createOrGetRuntime(spaceId);
-			const subSessionFactory = this.createSubSessionFactory(taskId, spaceId);
+			const subSessionFactory = this.createSubSessionFactory(
+				taskId,
+				spaceId,
+				workflowRun?.id ?? ''
+			);
 
 			const workflowRunId = workflowRun?.id ?? '';
 
@@ -409,7 +413,8 @@ export class TaskAgentManager {
 	async createSubSession(
 		taskId: string,
 		sessionId: string,
-		init: AgentSessionInit
+		init: AgentSessionInit,
+		mcpServers?: Record<string, McpServerConfig>
 	): Promise<string> {
 		const subSession = AgentSession.fromInit(
 			init,
@@ -419,6 +424,11 @@ export class TaskAgentManager {
 			this.config.getApiKey,
 			this.config.defaultModel
 		);
+
+		// Attach MCP servers before streaming starts so tools are available from turn 1.
+		if (mcpServers) {
+			subSession.setRuntimeMcpServers(mcpServers);
+		}
 
 		// Determine step ID from session convention or task context.
 		// The subSessions map uses the actual session ID as both the map key and session ID.
@@ -663,14 +673,39 @@ export class TaskAgentManager {
 	/**
 	 * Create a `SubSessionFactory` implementation bound to a specific taskId.
 	 * Passed to `createTaskAgentMcpServer()` for the given Task Agent session.
+	 *
+	 * @param workflowRunId  The active workflow run ID — forwarded to the step agent
+	 *   MCP server so it can read the resolved channel topology from run.config.
 	 */
-	private createSubSessionFactory(taskId: string, spaceId: string): SubSessionFactory {
+	private createSubSessionFactory(
+		taskId: string,
+		spaceId: string,
+		workflowRunId: string
+	): SubSessionFactory {
 		return {
 			create: async (
 				init: AgentSessionInit,
 				memberInfo?: SubSessionMemberInfo
 			): Promise<string> => {
-				const sessionId = await this.createSubSession(taskId, init.sessionId, init);
+				// Build the step agent MCP server BEFORE creating the sub-session.
+				// The sessionId is known from init.sessionId (randomUUID set by spawn_step_agent).
+				// The MCP server closures capture it for group lookups and messageInjector calls.
+				const stepAgentMcpServer = createStepAgentMcpServer({
+					ownSessionId: init.sessionId,
+					ownRole: memberInfo?.role ?? 'agent',
+					taskId,
+					workflowRunId,
+					sessionGroupRepo: this.config.sessionGroupRepo,
+					workflowRunRepo: this.config.workflowRunRepo,
+					messageInjector: (targetSessionId, message) =>
+						this.injectSubSessionMessage(targetSessionId, message),
+					getGroupId: () => this.taskGroupIds.get(taskId),
+					taskAgentMessageInjector: (message) => this.injectTaskAgentMessage(taskId, message),
+				});
+
+				const sessionId = await this.createSubSession(taskId, init.sessionId, init, {
+					'step-agent': stepAgentMcpServer as unknown as McpServerConfig,
+				});
 
 				// Add the sub-session as a member of the task's group (non-fatal).
 				const groupId = this.taskGroupIds.get(taskId);
@@ -1087,7 +1122,7 @@ export class TaskAgentManager {
 
 		// --- Build and attach MCP server (runtime-only, not persisted)
 		const runtime = await this.config.spaceRuntimeService.createOrGetRuntime(spaceId);
-		const subSessionFactory = this.createSubSessionFactory(taskId, spaceId);
+		const subSessionFactory = this.createSubSessionFactory(taskId, spaceId, workflowRun?.id ?? '');
 
 		const rehydrateWorkflowRunId = workflowRun?.id ?? '';
 
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -682,30 +682,22 @@ export class TaskAgentManager {
 		spaceId: string,
 		workflowRunId: string
 	): SubSessionFactory {
+		if (!workflowRunId) {
+			log.warn(
+				`TaskAgentManager: createSubSessionFactory called with empty workflowRunId for task ${taskId}. ` +
+					`Step agent channel topology will be unavailable — send_feedback will report no channels declared.`
+			);
+		}
 		return {
 			create: async (
 				init: AgentSessionInit,
 				memberInfo?: SubSessionMemberInfo
 			): Promise<string> => {
-				// Build the step agent MCP server BEFORE creating the sub-session.
-				// The sessionId is known from init.sessionId (randomUUID set by spawn_step_agent).
-				// The MCP server closures capture it for group lookups and messageInjector calls.
-				const stepAgentMcpServer = createStepAgentMcpServer({
-					ownSessionId: init.sessionId,
-					ownRole: memberInfo?.role ?? 'agent',
-					taskId,
-					workflowRunId,
-					sessionGroupRepo: this.config.sessionGroupRepo,
-					workflowRunRepo: this.config.workflowRunRepo,
-					messageInjector: (targetSessionId, message) =>
-						this.injectSubSessionMessage(targetSessionId, message),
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured implementation that correctly delivers all four requirements from Task 6.2. The code follows existing patterns (tool schemas, MCP server factory, handler separation for testability) and integrates cleanly into the TaskAgentManager.

### What works well

- **Channel validation** is correctly delegated to the existing `ChannelResolver`, which already handles hub-spoke enforcement (spoke-to-spoke paths are simply not declared, so `canSend()` returns false).
- **Three target forms** (point-to-point, broadcast `*`, multicast array) are all properly implemented with appropriate validation and error messages.
- **`request_peer_input`** correctly bypasses channel topology and injects a well-formatted routing request into the Task Agent session.
- **Error messages** are actionable -- they tell the agent what went wrong and suggest the fallback tool.
- **Self-exclusion** (line 304-306) prevents a step agent from messaging itself even when targeting its own role.
- **Partial failure handling** in `send_feedback` returns per-target results, which is the right approach.
- **System prompt update** in `custom-agent.ts` is comprehensive and documents both inbound and outbound message formats.
- **31 tests** cover the important topology patterns: no-topology, one-way, bidirectional, hub-spoke, broadcast, multicast, self-exclusion, injection failures, and the MCP server factory.

### Issues

**P2 -- `send_feedback` does not prefix messages with sender identity**

In `step-agent-tools.ts` line 330, the raw `message` is injected into the target session without any sender metadata:
```
await messageInjector(member.sessionId, message);
```

The recipient agent receives a bare message with no indication of which peer or role sent it. Compare with `request_peer_input` (lines 377-381) which properly prefixes with `[PEER_REQUEST from ${ownRole} ...]`. Consider wrapping the message similarly, e.g.:
```
`[Feedback from ${ownRole}]: ${message}`
```

This matters for multi-peer scenarios where a step agent might receive messages from different roles and needs to distinguish them.

**P3 -- Duplicate `seedSpaceRow` function in test file**

`packages/daemon/tests/unit/space/step-agent-tools.test.ts` defines `seedSpaceRow` twice (lines 55 and 110) with identical bodies. The second definition shadows the first via hoisting. Only one is needed.

**P3 -- Misleading code comment about cross-group scoping**

The module docstring in `step-agent-tools.ts` (line 37) says "the messageInjector validates the target session exists in the sub-sessions map (which only contains sessions for the same task)." In reality, `injectSubSessionMessage` in `task-agent-manager.ts` (line 462-470) searches across ALL task maps. The scoping guarantee actually comes from the group member filtering in `send_feedback` (line 304-306), not from the injector itself. The comment should reflect the actual trust boundary.

### Summary

Clean, well-tested implementation that follows existing project patterns. The missing sender prefix on `send_feedback` messages is the only substantive concern -- it could cause confusion in multi-peer topologies where the recipient needs to know who sent the message. The other issues are minor quality items.

lsm (commented):

> **Recommendation: REQUEST_CHANGES** — one P0 bug, one P1 issue, and a few P2/P3 items.

---

### Overall Assessment

The implementation is structurally sound: the three tools are correctly wired in, the channel topology validation via `ChannelResolver.canSend()` works as designed, and the 31 unit tests cover the important topology patterns. However, one correctness bug and one code quality issue need fixing before merge.

---

### P0 — `send_feedback` injects the raw message without any sender attribution

**File:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts`, line 330

```ts
await messageInjector(member.sessionId, message);
```

The raw `message` is injected verbatim into the target session. The receiving agent has no way to know who sent it. The `request_peer_input` path works correctly because `taskAgentMessage` is explicitly prefixed with `[PEER_REQUEST from {role} (session: {sessionId}) to {target_role}]`. By contrast, `send_feedback` — the *primary* tool — drops all sender identity on the floor.

The system prompt tells agents to handle messages prefixed `[PEER_REQUEST from {role} to {role}]`, and responses prefixed `[Peer response from {role}]`, but the primary `send_feedback` path produces no prefix at all. A receiving agent cannot distinguish a peer message from a human user message or a Task Agent relay. This undermines the stated communication model and could lead to misrouted replies.

The fix is to wrap the injected message before calling `messageInjector`:
```
[Message from {ownRole} (session: {ownSessionId})]: {message}
```
and document the expected format in the system prompt's Peer Communication section.

---

### P1 — Duplicate `seedSpaceRow` function definition in the test file

**File:** `packages/daemon/tests/unit/space/step-agent-tools.test.ts`, lines 55–61 and 110–116

The `seedSpaceRow` function is defined twice at the module level — once after `makeDb()` (line 55) and once inside the `TestCtx` type block (line 110). Both definitions are identical. Bun/TypeScript treats this as a duplicate identifier at module scope, which means the second declaration silently shadows the first. This is a latent bug: any future divergence between the two copies would cause confusing failures. The first definition (lines 55–61) is dead code since `makeCtx()` at line 122 calls the second one. Remove the first occurrence.

---

### P2 — `workflowRunId` silently falls back to `''` when no workflow run exists

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 275 and 1115

```ts
const subSessionFactory = this.createSubSessionFactory(taskId, spaceId, workflowRun?.id ?? '');
```

The empty-string `workflowRunId` is passed into `createSubSessionFactory`, which captures it in the `StepAgentToolsConfig.workflowRunId` closure for every step agent MCP server. Inside `step-agent-tools.ts`, both `list_peers` and `send_feedback` call `workflowRunRepo.getRun(workflowRunId)` — and `getRun('')` will return `null`, silently producing an empty `ChannelResolver`. The step agent will see `channelTopologyDeclared: false` even if channels were meant to be active. This is a pre-existing pattern (same fallback exists in `task-agent-tools.ts`), but it affects the new tools specifically because they both re-read the run config at call time. At minimum, a guard or log warning when `workflowRunId` is empty would prevent silent misconfiguration.

---

### P2 — No message length validation on `send_feedback.message` or `request_peer_input.question`

**File:** `packages/daemon/src/lib/space/tools/step-agent-tool-schemas.ts`

Both `message` (line 64) and `question` (line 86) are plain `z.string()` with no `.max()` constraint. An agent could inject an arbitrarily large string into a peer session's context window, potentially truncating or corrupting the peer's working context. The existing `task-agent-tool-schemas.ts` also lacks length constraints, but the new agent-to-agent messaging surfaces are the first place where one agent directly controls the full content of another agent's input turn. A reasonable upper bound (e.g., 10,000 characters) would guard against runaway context bloat.

---

### P2 — System prompt hardcodes `customAgent.role` for PEER_REQUEST prefix at build time

**File:** `packages/daemon/src/lib/space/agents/custom-agent.ts`, line 185

```ts
`**When you receive a message prefixed \`[PEER_REQUEST from {role} to ${customAgent.role}]: ...\`**, `
```

The `${customAgent.role}` is interpolated once when the system prompt is built. If the agent's role value ever changes between prompt construction and use (e.g. after a database update and rehydration), the system prompt will describe the wrong prefix. This is currently low risk because `customAgent.role` is stable in practice, but the asymmetry is notable: the `[PEER_REQUEST from {role} to ...]` prefix in the injected message uses `target_role` (the actual runtime role) while the system prompt uses the compile-time role. If they ever diverge, the agent won't match the prefix and will ignore peer requests. Consider using the literal placeholder `{role}` and documenting that it will contain the agent's own role.

---

### P3 — The test for `createStepAgentMcpServer` could verify tool descriptions

**File:** `packages/daemon/tests/unit/space/step-agent-tools.test.ts`, lines 823–839

The server registration test checks that `server.instance._registeredTools` has the three expected names, which mirrors the pattern in `task-agent-tools.test.ts`. However, unlike `task-agent-tools.test.ts` (lines 1624–1668), no test verifies that the registered tool handlers are callable or that their schemas are structurally correct. This is a minor gap — the tools are tested indirectly through `createStepAgentToolHandlers`, but a smoke test that calls a handler through the MCP server layer would provide additional confidence.

---

### P3 — `workflowRunRepo` dependency is unused in `request_peer_input`

**File:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts`

`StepAgentToolsConfig` requires `workflowRunRepo` (line 88), but `request_peer_input` does not use it — only `list_peers` and `send_feedback` do. This is not wrong, but the requirement is set at the config level, meaning any caller must supply a repo even for setups where `request_peer_input` is the only tool needed. Documenting which tools need which dependencies (as a comment in the config interface) would help future consumers.

---

### Summary

The core channel topology enforcement is correct and well-tested. The primary fix needed before merge is the missing sender identity prefix in `send_feedback` — without it, the receiving agent cannot distinguish peer messages from human input, breaking the communication protocol described in the system prompt. The duplicate `seedSpaceRow` in the test file is a secondary fix.

lsm (commented):

Re-review after feedback was addressed. All previously identified issues have been resolved correctly:

### Verified Fixes

1. **P0 (sender identity prefix)**: `send_feedback` now prefixes injected messages with `[Message from {role} (session: {id})]: {msg}` at line 333 of `step-agent-tools.ts`. The test at line 362 verifies this prefix is present. The system prompt in `custom-agent.ts` (lines 179-183) documents this format so receiving agents know what to expect.

2. **P1 (duplicate seedSpaceRow)**: Only one `seedSpaceRow` function exists in `step-agent-tools.test.ts` (line 102), confirmed by grep.

3. **P2 (.max(10000) on schemas)**: Both `message` (line 66) and `question` (line 90) fields in `step-agent-tool-schemas.ts` now have `.max(10000)` constraints with descriptive error messages.

4. **P2 (log.warn for empty workflowRunId)**: `createSubSessionFactory` in `task-agent-manager.ts` (lines 675-680) now warns when `workflowRunId` is empty, explaining the downstream impact on channel topology availability.

5. **P2 (literal placeholders in system prompt)**: The PEER_REQUEST pattern at line 190 now uses `{role}` and `{your-role}` as literal placeholders in the pattern description. The only runtime interpolation (`${customAgent.role}`) is correctly used only as an illustrative example ("e.g. `coder`"), not as part of the pattern itself.

6. **P3 (module docstring)**: The scoping section (lines 34-40) now correctly states that `messageInjector` does not enforce the cross-group boundary and that the group-member filtering in `send_feedback` is the actual trust boundary.

### Minor Observation (P3)

- Line 39 references "lines 304-310" for the group-member filtering code, but the actual filter is at lines 307-310 in the current file. Line references in comments tend to drift — consider removing the specific line numbers or using a code reference instead.

### Overall Assessment

The implementation is solid. All three tools (list_peers, send_feedback, request_peer_input) are well-tested with 25+ test cases covering the key topology patterns (no channels, one-way, bidirectional, hub-spoke, multicast, broadcast), edge cases (self-exclusion, injection failures, missing groups/sessions), and the MCP server factory. The channel validation logic correctly delegates to `ChannelResolver` and the error messages guide agents toward the fallback tool when direct channels are unavailable.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** — all previous feedback has been addressed correctly.

---

### Follow-up Verification

This review verifies that the five fixes from the previous round were applied correctly. Each issue is confirmed resolved.

#### P0 — send_feedback now prefixes injected messages with sender identity [FIXED]

**File:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts`, line 333

The fix is correct. The raw message is now wrapped before injection:

```ts
const prefixedMessage = `[Message from ${ownRole} (session: ${ownSessionId})]: ${message}`;
await messageInjector(member.sessionId, prefixedMessage);
```

The format `[Message from {role} (session: {sessionId})]: ...` matches the template documented in the system prompt at `custom-agent.ts` line 180. The test at line 362 validates the prefix is present:

```ts
expect(injected[0].message).toContain('[Message from coder (session:');
```

#### P1 — Duplicate seedSpaceRow removed [FIXED]

The test file now defines `seedSpaceRow` exactly once (line 102). The second occurrence has been removed. `makeCtx()` at line 114 calls it correctly.

#### P2 — Message and question fields now have .max(10000) [FIXED]

`packages/daemon/src/lib/space/tools/step-agent-tool-schemas.ts`:
- `message` field: `z.string().max(10000, 'Message must be 10,000 characters or fewer')`
- `question` field: `z.string().max(10000, 'Question must be 10,000 characters or fewer')`

Both bounds are in place.

#### P2 — log.warn when createSubSessionFactory receives empty workflowRunId [FIXED]

`packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 675:

```ts
if (!workflowRunId) {
    log.warn(
        `TaskAgentManager: createSubSessionFactory called with empty workflowRunId for task ${taskId}. ` +
        `Step agent channel topology will be unavailable — send_feedback will report no channels declared.`
    );
}
```

The warning is descriptive and explains the downstream consequence.

#### P2 — System prompt uses literal {your-role} placeholder [FIXED]

`packages/daemon/src/lib/space/agents/custom-agent.ts`, line 190–193:

The PEER_REQUEST prefix description now uses the literal `{your-role}` placeholder in the template, with a parenthetical example using the runtime value:

```
**When you receive a message prefixed `[PEER_REQUEST from {role} to {your-role}]: ...**
(where `{your-role}` is your own role, e.g. `${customAgent.role}`),
```

This is a reasonable approach — the prefix template itself is stable (literal), and the example helps the agent recognize its own role. The interpolation in the example is intentional and appropriate here.

#### P3 — Module docstring fixed [FIXED]

The docstring now correctly documents that scoping is enforced by group-member filtering in `send_feedback`, not by the messageInjector itself.

---

### No New Issues Found

The implementation is clean. The three peer communication tools work correctly, the channel topology enforcement is properly delegated to `ChannelResolver`, and the 31 unit tests cover all meaningful topology patterns and edge cases.

lsm:
Superseded by https://github.com/lsm/neokai/pull/678

PR #678, opened 2026-03-22
feat: add step agent peer communication tools (send_feedback, request_peer_input, list_peers)

lsm:
## Summary

Implements Task 6.2 — Step Agent Peer Communication Tools (Channel-Validated).

- **`send_feedback(target, message)`** — primary direct-messaging tool for step agents, validated against the declared channel topology via `ChannelResolver`. Supports single role (`'reviewer'`), broadcast (`'*'`), and multicast (`['coder','reviewer']`). Unauthorized directions return a clear error with available channels and suggest `request_peer_input` as fallback.
- **`request_peer_input(targetRole, question)`** — fallback Task Agent-mediated routing for undeclared channels. Async/non-blocking — response flows back as a user turn prefixed `[Peer response from <role>]: ...`.
- **`list_peers()`** — returns all other group members with role, status, sessionId, and `canSendDirectly` flag per the declared channel topology.
- Step agent system prompt updated with Peer Communication section explaining when to use each tool, async nature of `request_peer_input`, and hub-spoke enforcement.

## Topology patterns supported

| Pattern | Description | send_feedback behavior |
|---------|-------------|------------------------|
| A→B one-way | A sends to B | B cannot reply via send_feedback |
| A↔B bidirectional | Both directions | Both directions permitted |
| hub→[A,B,C] fan-out | Hub to all spokes | Spokes cannot message each other |
| hub↔[A,B,C] hub-spoke | Hub and spoke bidirectional | Spokes can only reply to hub |

## Implementation details

- New `step-agent-tools.ts` with `StepAgentToolsConfig`, handlers, and MCP server factory
- `task-agent-manager.ts`: `createSubSessionFactory.create()` now builds and attaches a step agent MCP server to `init.mcpServers` before `AgentSession.fromInit()` is called — ensures tools are available when the first message is processed
- The `peerMessageInjector` reuses the existing `injectSubSessionMessage()` callback; `taskAgentMessageInjector` routes to `injectTaskAgentMessage()` for `request_peer_input`

## Test plan

- [x] 27 unit tests in `step-agent-tools.test.ts` covering all topology patterns, error cases, and MCP server registration
- [x] 2 new tests in `custom-agent.test.ts` for system prompt peer communication section
- [x] All 1287 space unit tests pass (0 fail)
- [x] Lint: 0 warnings, 0 errors
- [x] Format: clean
- [x] TypeScript: no errors

### DIFF
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -665,6 +665,18 @@ export class TaskAgentManager {
 	 * Passed to `createTaskAgentMcpServer()` for the given Task Agent session.
 	 */
 	private createSubSessionFactory(taskId: string, spaceId: string): SubSessionFactory {
+		// Capture workflowRunId once at factory-creation time to avoid a DB round-trip
+		// on every spawn. The run ID is immutable once assigned, so this is safe.
+		// Log a warning if the task lacks a workflow run — step-agent tools will
+		// produce an empty ChannelResolver (no declared channels) in that case.
+		const taskWorkflowRunId = this.config.taskRepo.getTask(taskId)?.workflowRunId ?? '';
+		if (!taskWorkflowRunId) {
+			log.warn(
+				`TaskAgentManager.createSubSessionFactory: task ${taskId} has no workflowRunId — ` +
+					`step-agent channel topology will be unavailable (request_peer_input still works)`
+			);
+		}
+
 		return {
 			create: async (
 				init: AgentSessionInit,
@@ -1157,8 +1169,16 @@ export class TaskAgentManager {
 		// --- Rebuild subSessions map from step tasks in the same workflow run
 		// Sub-sessions are not fully rehydrated (no streaming restart) — the Task Agent
 		// will re-spawn them as needed after receiving the re-orientation message.
+		// Step-agent MCP tools are runtime-only (not persisted), so we must re-attach
+		// them here the same way createSubSessionFactory.create() does it.
 		if (task.workflowRunId) {
-			const stepTasks = this.config.taskRepo.listByWorkflowRun(task.workflowRunId);
+			const workflowRunId = task.workflowRunId;
+
+			// Load the group once; we need it to look up each sub-session's role.
+			const groupId = this.taskGroupIds.get(taskId);
+			const group = groupId ? this.config.sessionGroupRepo.getGroup(groupId) : null;
+
+			const stepTasks = this.config.taskRepo.listByWorkflowRun(workflowRunId);
 			for (const stepTask of stepTasks) {
 				const subSessionId = stepTask.taskAgentSessionId;
 				if (!subSessionId || stepTask.id === taskId) continue;
@@ -1175,6 +1195,22 @@ export class TaskAgentManager {
 					this.config.getApiKey
 				);
 				if (!subSession) continue;
+
+				// Re-attach step-agent MCP tools (runtime-only, not persisted to DB).
+				// Look up the member's role from the session group; fall back to 'agent'.
+				const memberRole =
+					group?.members.find((m) => m.sessionId === subSessionId)?.role ?? 'agent';
+
+				const stepAgentMcpServer = this.buildStepAgentMcpServerForSession(
+					taskId,
+					subSessionId,
+					memberRole,
+					spaceId,
+					workflowRunId
+				);
+				subSession.setRuntimeMcpServers({
+					'step-agent': stepAgentMcpServer as unknown as McpServerConfig,
+				});
 
 				if (!this.subSessions.has(taskId)) {
 					this.subSessions.set(taskId, new Map());
--- a/packages/daemon/src/lib/space/tools/step-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/step-agent-tools.ts
@@ -274,56 +274,67 @@ export function createStepAgentToolHandlers(config: StepAgentToolsConfig) {
 				});
 			}
 
-			// Resolve target roles to session IDs from group members
-			// Multiple members can share the same role (parallel instances)
-			const targetMembers = group.members.filter(
-				(m) =>
-					targetRoles.includes(m.role) && m.sessionId !== mySessionId && m.role !== 'task-agent'
+			// Find peer sessions for each target role (exclude self and task-agent)
+			const peers = group.members.filter(
+				(m) => m.sessionId !== mySessionId && m.role !== 'task-agent'
 			);
+			const delivered: Array<{ role: string; sessionId: string }> = [];
+			const notFound: string[] = [];
+			const failed: Array<{ role: string; sessionId: string; error: string }> = [];
 
-			if (targetMembers.length === 0) {
-				return jsonResult({
-					success: false,
-					error:
-						`No active sessions found for target role(s): ${targetRoles.join(', ')}. ` +
-						`Use list_peers to check which peers are currently active.`,
-					targetRoles,
-				});
-			}
-
-			// Prefix message with sender identity so the receiver can attribute it.
-			// Without this, agents in multi-peer groups cannot tell who sent the message.
-			const attributedMessage = `[Feedback from ${myRole}]: ${message}`;
-
-			// Inject message into each target session
-			const injected: Array<{ sessionId: string; role: string }> = [];
-			const failed: Array<{ sessionId: string; role: string; error: string }> = [];
-
-			for (const member of targetMembers) {
-				try {
-					await messageInjector(member.sessionId, attributedMessage);
-					injected.push({ sessionId: member.sessionId, role: member.role });
-				} catch (err) {
-					const msg = err instanceof Error ? err.message : String(err);
-					failed.push({ sessionId: member.sessionId, role: member.role, error: msg });
+			// Best-effort delivery: attempt all targets, aggregate errors.
+			// This avoids the partial-delivery / stop-on-first-error problem where
+			// some peers receive the message while others don't, and the caller gets
+			// success: false despite partial delivery having already occurred.
+			for (const targetRole of targetRoles) {
+				const targetSessions = peers.filter((m) => m.role === targetRole);
+				if (targetSessions.length === 0) {
+					notFound.push(targetRole);
+					continue;
+				}
+				for (const targetMember of targetSessions) {
+					const prefixedMessage = `[Feedback from ${myRole}]: ${message}`;
+					try {
+						await messageInjector(targetMember.sessionId, prefixedMessage);
+						delivered.push({ role: targetRole, sessionId: targetMember.sessionId });
+					} catch (err) {
+						const errMsg = err instanceof Error ? err.message : String(err);
+						failed.push({ role: targetRole, sessionId: targetMember.sessionId, error: errMsg });
+					}
 				}
 			}
 
-			if (injected.length === 0) {
+			if (notFound.length > 0 && delivered.length === 0 && failed.length === 0) {
 				return jsonResult({
 					success: false,
-					error: `Failed to deliver message to any target. All injections failed.`,
+					error:
+						`No active sessions found for target role(s): ${notFound.join(', ')}. ` +
+						`Use list_peers to check which peers are currently active.`,
+					notFoundRoles: notFound,
+				});
+			}
+
+			if (failed.length > 0) {
+				// Partial or total delivery failure — report what was and was not delivered.
+				return jsonResult({
+					success: delivered.length > 0 ? 'partial' : false,
+					delivered,
 					failed,
+					notFoundRoles: notFound.length > 0 ? notFound : undefined,
+					message:
+						delivered.length > 0
+							? `Message delivered to ${delivered.length} peer(s) but failed for ${failed.length} peer(s).`
+							: `Message delivery failed for all ${failed.length} target(s).`,
 				});
 			}
 
 			return jsonResult({
 				success: true,
-				delivered: injected,
-				...(failed.length > 0 ? { partialFailures: failed } : {}),
+				delivered,
+				notFoundRoles: notFound.length > 0 ? notFound : undefined,
 				message:
-					`Message delivered to ${injected.length} peer(s): ` +
-					injected.map((t) => `${t.role} (${t.sessionId})`).join(', ') +
+					`Message delivered to ${delivered.length} peer(s): ` +
+					delivered.map((t) => `${t.role} (${t.sessionId})`).join(', ') +
 					'.',
 			});
 		},
--- a/packages/daemon/tests/unit/space/custom-agent.test.ts
+++ b/packages/daemon/tests/unit/space/custom-agent.test.ts
@@ -158,6 +158,22 @@ describe('buildCustomAgentSystemPrompt', () => {
 		expect(prompt).toContain('pullrequestreview');
 	});
 
+	it('includes peer communication section with all three tools', () => {
+		const agent = makeAgent();
+		const prompt = buildCustomAgentSystemPrompt(agent);
+		expect(prompt).toContain('Peer Communication');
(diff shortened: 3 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured implementation of step agent peer communication tools. The code is correct, follows established project patterns, and has thorough test coverage. Here is my detailed analysis:

### Correctness

- **`send_feedback`**: Channel validation via `ChannelResolver` is correct. The tool properly handles single target, broadcast (`*`), and multicast (array) modes. Unauthorized directions are rejected before any messages are sent, which is the right approach (fail-fast on partial authorization).
- **`request_peer_input`**: Correctly routes through the Task Agent without requiring group membership, which is appropriate since it is a fallback path.
- **`list_peers`**: Properly excludes self from peer list and uses `ChannelResolver.canSend()` for the `canSendDirectly` flag.

### Integration

- The MCP server is wired into `createSubSessionFactory.create()` at the right point -- before `AgentSession.fromInit()` is called. This ensures tools are available from the first message.
- The `as unknown as McpServerConfig` cast follows the established pattern used in 8+ other locations in the codebase.
- `peerMessageInjector` correctly delegates to `injectSubSessionMessage()` which searches the `subSessions` map by session ID.
- `taskAgentMessageInjector` correctly delegates to `injectTaskAgentMessage()` scoped to the task ID.

### Potential Concern (P3 - not blocking)

1. **`workflowRunId` captured eagerly at sub-session creation** (line 676 of `task-agent-manager.ts`): `this.config.taskRepo.getTask(taskId)?.workflowRunId ?? ''` is evaluated once when the MCP server is created. If the task does not have a `workflowRunId` at that point, `buildResolver()` will always return an empty resolver. This seems acceptable in practice since workflow tasks always have a `workflowRunId` by the time sub-sessions are spawned, and `request_peer_input` works as a fallback regardless.

2. **Early return on first delivery error** (lines 283-289 of `step-agent-tools.ts`): In `send_feedback` multicast, if delivery to one target fails, the function returns immediately with a partial `delivered` array. This is a reasonable choice (fail-fast), but the caller will not know which remaining targets were not attempted. A `notAttempted` field could improve debuggability.

### Test Coverage

27 unit tests with a real SQLite database cover:
- All topology patterns (one-way, bidirectional, hub-spoke, fan-out)
- Error cases (no group, no topology, unauthorized, missing sessions, injector failure)
- Edge cases (self-loop exclusion, broadcast with no permitted targets, mixed auth/unauth multicast)
- MCP server factory registration

2 custom-agent tests verify the system prompt includes the peer communication section.

### Summary

Clean implementation that follows existing patterns. No correctness bugs, no security concerns (all messaging goes through validated channels or the Task Agent mediation path). The only observations are minor P3 nits.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 1 P0, 2 P1, 2 P2.

---

## Summary

The core implementation is well-structured and the channel validation logic is correct. However there is one P0 regression: rehydrated sub-sessions lose their step-agent MCP tools after a daemon restart. There are also two P1 issues — non-atomic multicast delivery and a missing `send_feedback` test for the partial-delivery case.

---

## P0 — Step-agent MCP tools not re-attached on sub-session rehydration

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 1163–1190 (`rehydrateTaskAgent`)

The rehydration loop restores sub-session `AgentSession` objects from the DB but never calls `setRuntimeMcpServers` on them. The step-agent MCP server (with `send_feedback`, `request_peer_input`, `list_peers`) is a runtime-only attachment — it is not persisted to the DB. After a daemon restart, any in-flight sub-session will have its conversation history and streaming query restored, but the three peer-communication tools will be absent. If the agent then attempts to call `send_feedback` or `list_peers` it will receive an unknown-tool error from the SDK.

The fix is to call `createStepAgentMcpServer` for each rehydrated sub-session and attach it via `setRuntimeMcpServers`, the same way it is done in `createSubSessionFactory.create()`. The `stepRole` is available from the group member record (look up the member row by `subSessionId` using `sessionGroupRepo.getGroup`).

This is the same pattern already applied correctly to Task Agent sessions in the same method (lines 1108–1131): the Task Agent's MCP server is rebuilt and re-attached even though the session was persisted. The sub-session path is missing the equivalent step.

---

## P1 — Non-atomic multicast: partial delivery with no rollback

**File:** `packages/daemon/src/lib/space/tools/step-agent-tools.ts`, lines 272–292

In the `send_feedback` handler, the loop injects messages into target sessions one at a time. If injection to the second (or later) recipient throws, the function returns an error but the messages already delivered to earlier recipients are not rolled back:

```ts
return jsonResult({
    success: false,
    error: `Failed to deliver to ${targetRole} (${targetMember.sessionId}): ${errMsg}`,
    delivered,  // partial list
});
```

The caller receives `success: false` but some peers may have already received the message. For a reviewer + coder broadcast, the reviewer gets the message while the coder does not. This can cause inconsistent state and confuse agents that see feedback they cannot reconcile with the sender's reported failure.

Two options: (1) deliver all then aggregate errors (best-effort delivery with a partial-success result), or (2) pre-validate that all target sessions exist before injecting any message (fail-fast before side effects). The current half-way approach — stopping mid-loop and reporting failure after partial delivery — is the worst of both worlds. At minimum the error response should clearly distinguish "partial success" from "total failure" rather than using `success: false` uniformly.

**Missing test:** There is no test covering the partial-delivery scenario (first target succeeds, second throws). The `send_feedback` suite should include a test where the injector throws for the second target in a multicast and verifies the response accurately reflects what was and was not delivered.

---

## P1 — Missing test for `send_feedback` group-not-in-DB error

**File:** `packages/daemon/tests/unit/space/step-agent-tools.test.ts`

There is a test for `getGroupId() => undefined` (lines 613–620), but no test for the case where `getGroupId()` returns a non-empty ID that is not in the database (the analogous path to lines 203–210 in the handler). The `list_peers` suite has this test (line 273) but `send_feedback` does not. The handler path (lines 203–210) is reachable and returns a `success: false` result, but it is untested.

---

## P2 — Lazy DB fetch of `workflowRunId` inside `createSubSessionFactory` closure

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 676

```ts
workflowRunId: this.config.taskRepo.getTask(taskId)?.workflowRunId ?? '',
```

This re-reads the task from the DB every time a sub-session is created. This is called inside the `create` closure that is invoked per `spawn_step_agent` tool call. The `taskId` is already in scope; the workflow run ID could have been passed in from `spawnTaskAgent` where it is already known (via `workflowRun?.id`). The extra DB round-trip is low-cost but unnecessary, and the empty-string fallback silently produces an unusable `ChannelResolver` rather than surfacing the missing run ID as a recoverable warning. Consider caching the value at closure-creation time or passing it down explicitly.

---

## P2 — System prompt peer-communication section added to all agents regardless of topology

**File:** `packages/daemon/src/lib/space/agents/custom-agent.ts`, lines 152–177

`buildCustomAgentSystemPrompt` unconditionally appends the peer communication section to every custom agent's system prompt, including agents that will never participate in a group (e.g., standalone space agents). This is a minor waste of context tokens and may confuse agents that have no peers. It is better than the reverse (omitting when tools are present), but could be gated on an agent-level flag or on whether a workflow step declares channels. This is a minor quality issue — the current behaviour is safe.

---

## Positive notes

- The channel validation logic in `ChannelResolver` is correct and well-tested. The `isValidResolvedChannel` guard is a good defensive measure against DB corruption.
- The `buildResolver()` pattern (re-reading the run config on every tool call) correctly handles topology changes mid-execution and avoids stale cached state.
- The test suite is thorough for the happy paths and topology patterns. The hub-spoke tests correctly verify the asymmetric spoke-cannot-message-spoke constraint.
- The `taskAgentMessageInjector` closure correctly binds `taskId` at factory-creation time, preventing cross-task message routing.
- The cast `stepAgentMcpServer as unknown as McpServerConfig` is consistent with the existing pattern for the task-agent server and is safe given the comment explaining the structural compatibility.

lsm (commented):

**Recommendation: APPROVE**

The fix commit (`e3715ee1`) correctly addresses all issues raised in the prior Sonnet review. All 4 files changed are clean and all tests pass (30/30 step-agent-tools, 90/90 task-agent-manager, 75/75 custom-agent).

### Issues addressed

**P0 — Rehydration re-attaches step-agent MCP tools**: Fixed correctly in `task-agent-manager.ts:1205-1222`. The rehydration path now mirrors `createSubSessionFactory.create()`: it looks up the member role from the session group (with `'agent'` fallback), creates a new `createStepAgentMcpServer`, and calls `setRuntimeMcpServers`. The group is loaded once before the loop (line 1184-1185), which is efficient. Test at line 1982 verifies `_mcpServers` contains `'step-agent'` after rehydration.

**P1 — Best-effort multicast delivery**: The `send_feedback` handler now accumulates a `failed[]` array instead of returning early on first error. The three-way result logic is correct:
- `delivered > 0, failed > 0` → `success: 'partial'`
- `delivered === 0, failed > 0` → `success: false` with failed details
- All edge cases with `notFound` combinations are handled correctly (verified by tracing all 5 path combinations through the conditionals)

Two new tests cover partial and total failure scenarios.

**P1 — Group-not-in-DB test**: New test at line 623 verifies the case where `getGroupId()` returns a non-null value but the group doesn't exist in the database.

**P2 — workflowRunId captured once**: `taskWorkflowRunId` is captured at factory-creation time (line 662) with a warning log when absent. This eliminates the per-spawn DB lookup.

1. The warning log at `task-agent-manager.ts:664` fires for every task without a `workflowRunId`, including standalone tasks where this is expected. This is noisy but acceptable since it only fires once per factory creation and provides useful debugging context.

2. The `success: 'partial'` string value is a pragmatic choice for the MCP tool response — it's more informative than a boolean for LLM consumers. The typing is loose (`string | boolean`) but this is a JSON response consumed by an LLM, not a typed API boundary.

All changes are well-scoped, well-tested, and correctly integrated.

lsm (commented):

**Recommendation: APPROVE**

All four prior issues have been correctly addressed. The implementation is solid. Some minor observations below for completeness.

---

### Prior issue resolution

**P0 — Rehydration re-attachment** (`task-agent-manager.ts` lines 1180–1228)

Confirmed correct. The rehydration path:
1. Loads the group once from DB using the already-restored `taskGroupIds` map.
2. Iterates step tasks for the workflow run, skipping self and non-sub-session types.
3. Calls `AgentSession.restore()`, then `createStepAgentMcpServer(...)` with the member's role looked up from group members (with `'agent'` fallback).
4. Calls `subSession.setRuntimeMcpServers({ 'step-agent': ... })`.
5. Populates `this.subSessions` map — making the sub-session findable by `injectSubSessionMessage`.

The test `step-agent MCP server is re-attached on rehydrated sub-sessions` at line 1982 verifies `_mcpServers` contains `'step-agent'`.

**P1 — Non-atomic multicast / best-effort delivery** (`step-agent-tools.ts` lines 273–315)

The fix is correct. All target roles are iterated even after a failure; errors are aggregated into `failed[]`. The three-way return contract is:
- All notFound, zero delivered/failed → `success: false`
- Any failed → `success: 'partial'` (if some delivered) or `success: false` (all failed)
- All delivered → `success: true` (with optional `notFoundRoles`)

**P1 — New tests**

Four new tests confirmed present and passing:
- `returns error when group not in DB` in `list_peers` describe (step-agent-tools.test.ts:273)
- `returns error when group ID returned but not in DB` in `send_feedback` describe (step-agent-tools.test.ts:623)
- `best-effort multicast: first delivery succeeds, second fails — partial success` (line 638)
- `best-effort multicast: all deliveries fail — success: false` (line 671)
- `step-agent MCP server is re-attached on rehydrated sub-sessions` (task-agent-manager.test.ts:1982)

That is 5 new tests against the stated 4 — a bonus.

**P2 — `workflowRunId` captured once at factory-creation time** (`task-agent-manager.ts` lines 658–668)

Correct. The `taskWorkflowRunId` is captured from `taskRepo.getTask(taskId)?.workflowRunId ?? ''` at factory-creation time, with a `log.warn` when it is empty. The empty-string case safely produces an empty `ChannelResolver` (no channels declared), so `send_feedback` returns a clear "no topology declared" error and falls through to `request_peer_input`.

---

### Remaining observations

**P3 — `subSessionMemberIds` not repopulated during rehydration**

After a daemon restart, the rehydration path for sub-sessions (`rehydrateTaskAgent` lines 1188–1228) correctly rebuilds `this.subSessions` but does NOT rebuild `this.subSessionMemberIds`. This means that if a rehydrated sub-session completes or errors after the restart, neither `handleSubSessionComplete` (line 931) nor the error handler (line 850) will call `updateMemberStatus` on the DB row, so group member status stays `'active'` in the DB forever.

This is not introduced by this PR — it was a pre-existing gap in the rehydration design (sub-sessions were always treated as stubs to be re-spawned). No action needed for this PR; a follow-up task to repopulate `subSessionMemberIds` from the group's member rows would close the gap cleanly.

**P3 — No test for notFound + delivered mixed case**

The case where some target roles are not found but others are successfully delivered (returns `success: true` with `notFoundRoles` populated) has no dedicated test. The code path is correct and straightforward, but a test would prevent regression.

**P3 — `success: 'partial'` is a string, not a boolean**

`success: 'partial'` mixes types in the JSON response (`true | false | 'partial'`). LLM callers doing `if (data.success)` will treat `'partial'` as truthy — which may or may not be the intended signal. The PR description explicitly documents this as the intended semantics for the `delivered[]`/`failed[]` response, and the test asserts `toBe('partial')`, so this is a conscious design choice. Worth documenting in the `StepAgentToolsConfig` JSDoc or a follow-up type definition.

---

**Test run results:**
- `step-agent-tools.test.ts`: 30 pass, 0 fail
- `task-agent-manager.test.ts`: 90 pass, 0 fail

lsm (commented):

**Recommendation: APPROVE**

Post-rebase re-review of the step agent peer communication tools (send_feedback, request_peer_input, list_peers). The rebase onto dev is clean with no regressions.

### Verification Summary

- **Typecheck**: Passes cleanly (`tsc --build --noEmit`)
- **Lint**: 0 warnings, 0 errors (oxlint)
- **Tests**: All 195 tests pass across the 3 affected test files:
  - `step-agent-tools.test.ts`: 30/30 pass
  - `custom-agent.test.ts`: 75/75 pass
  - `task-agent-manager.test.ts`: 90/90 pass

### Code Review

**Implementation quality is solid across all changed files:**

1. **`step-agent-tools.ts`** — Clean tool handler implementation with proper channel validation, best-effort multicast delivery, and discriminated union error handling via `loadGroupAndResolver()`. The three tools have clear separation of concerns: `list_peers` for discovery, `send_feedback` for channel-validated direct messaging, `request_peer_input` for Task Agent-mediated fallback.

2. **`task-agent-manager.ts`** — Two changes:
   - `createSubSessionFactory` captures `workflowRunId` once at factory-creation time (avoiding repeated DB lookups). Warning log for missing workflow run is appropriate.
   - Rehydration path correctly re-attaches step-agent MCP servers to restored sub-sessions. The group is loaded once outside the loop, member role lookup falls back to `'agent'`, and `buildStepAgentMcpServerForSession` is reused from the normal code path.

3. **Test coverage** is thorough:
   - All channel topology patterns tested (point-to-point, broadcast, multicast, hub-spoke, bidirectional)
   - Error paths covered (missing group, missing sessions, injection failures, unauthorized channels)
   - Best-effort multicast with partial failures verified
   - Rehydration MCP re-attachment test added to task-agent-manager suite
   - System prompt peer communication section validated in custom-agent tests

### Minor Observations (P3)

- The `as unknown as McpServerConfig` cast at line 1212 of `task-agent-manager.ts` is a known pattern used elsewhere in the codebase for runtime MCP servers. Not ideal but consistent with existing code.
- The warning logs from `createSubSessionFactory` are noisy in tests (many tasks without workflow runs). Not a bug, just visual noise in test output.

No P0, P1, or P2 issues found. The rebase introduced no conflicts or regressions.

PR #683, opened 2026-03-22
feat: add E2E test for active agents display in SpaceTaskPane

lsm:
- Add space.sessionGroup.create RPC handler for test infrastructure:
  creates session groups with members and emits the correct WebSocket
  events (spaceSessionGroup.created + memberAdded) so the SpaceStore
  signal updates without running real agents

- Add packages/e2e/tests/features/space-session-groups.e2e.ts with
  8 serial tests covering the full Working Agents section:
  * Hidden when no groups exist
  * Active badge with animated blue dot
  * Completed badge with green checkmark SVG
  * Failed badge with red X SVG
  * Real-time update: active → completed without page refresh
  * Real-time update: active → failed without page refresh
  * Multiple members all displayed
  * Task Agent label for role='task-agent' (no agentId)
  * Named agent label when agentId links to a SpaceAgent record
  * Task click in sidebar opens task pane with Working Agents

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts
@@ -3,6 +3,7 @@
  *
  * Admin-oriented RPC handlers for session group management:
  * - space.sessionGroup.list         - List all groups for a space
+ * - space.sessionGroup.create       - Create a group with optional initial members (test/admin)
  * - space.sessionGroup.updateMember - Force-update a member's status (admin / stuck recovery)
  * - space.sessionGroup.delete       - Delete a stuck / orphaned group
  */
@@ -36,6 +37,74 @@ export function setupSpaceSessionGroupHandlers(
 
 		const groups = sessionGroupRepo.getGroupsBySpace(params.spaceId);
 		return { groups };
+	});
+
+	// ─── space.sessionGroup.create ───────────────────────────────────────────────
+	// Admin / test-infrastructure operation: create a session group with optional members.
+	// Used by E2E tests to inject session group state without running real agents.
+	messageHub.onRequest('space.sessionGroup.create', async (data) => {
+		const params = data as {
+			spaceId: string;
+			name: string;
+			taskId?: string;
+			members?: Array<{
+				sessionId: string;
+				role: string;
+				agentId?: string;
+				status?: 'active' | 'completed' | 'failed';
+			}>;
+		};
+
+		if (!params.spaceId) throw new Error('spaceId is required');
+		if (!params.name) throw new Error('name is required');
+
+		const space = await spaceManager.getSpace(params.spaceId);
+		if (!space) {
+			throw new Error(`Space not found: ${params.spaceId}`);
+		}
+
+		const group = sessionGroupRepo.createGroup({
+			spaceId: params.spaceId,
+			name: params.name,
+			taskId: params.taskId,
+		});
+
+		await daemonHub
+			.emit('spaceSessionGroup.created', {
+				sessionId: `space:${params.spaceId}`,
+				spaceId: params.spaceId,
+				taskId: params.taskId ?? '',
+				group,
+			})
+			.catch((err) => {
+				log.warn('Failed to emit spaceSessionGroup.created:', err);
+			});
+
+		const members = [];
+		for (let i = 0; i < (params.members ?? []).length; i++) {
+			const m = params.members![i];
+			const member = sessionGroupRepo.addMember(group.id, m.sessionId, {
+				role: m.role,
+				agentId: m.agentId,
+				status: m.status ?? 'active',
+				orderIndex: i,
+			});
+			members.push(member);
+
+			await daemonHub
+				.emit('spaceSessionGroup.memberAdded', {
+					sessionId: `space:${params.spaceId}`,
+					spaceId: params.spaceId,
+					groupId: group.id,
+					member,
+				})
+				.catch((err) => {
+					log.warn('Failed to emit spaceSessionGroup.memberAdded:', err);
+				});
+		}
+
+		const fullGroup = sessionGroupRepo.getGroup(group.id)!;
+		return { group: fullGroup };
 	});
 
 	// ─── space.sessionGroup.updateMember ─────────────────────────────────────────
--- a/packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts
@@ -42,7 +42,12 @@ export function setupSpaceSessionGroupHandlers(
 	// ─── space.sessionGroup.create ───────────────────────────────────────────────
 	// Admin / test-infrastructure operation: create a session group with optional members.
 	// Used by E2E tests to inject session group state without running real agents.
+	// Not available in production to prevent creation of orphaned groups with phantom sessions.
 	messageHub.onRequest('space.sessionGroup.create', async (data) => {
+		if (process.env.NODE_ENV === 'production') {
+			throw new Error('space.sessionGroup.create is not available in production');
+		}
+
 		const params = data as {
 			spaceId: string;
 			name: string;
--- a/packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts
@@ -85,7 +85,6 @@ export function setupSpaceSessionGroupHandlers(
 				log.warn('Failed to emit spaceSessionGroup.created:', err);
 			});
 
-		const members = [];
 		for (let i = 0; i < (params.members ?? []).length; i++) {
 			const m = params.members![i];
 			const member = sessionGroupRepo.addMember(group.id, m.sessionId, {
@@ -94,7 +93,6 @@ export function setupSpaceSessionGroupHandlers(
 				status: m.status ?? 'active',
 				orderIndex: i,
 			});
-			members.push(member);
 
 			await daemonHub
 				.emit('spaceSessionGroup.memberAdded', {
--- a/packages/daemon/tests/unit/rpc-handlers/space-session-group-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc-handlers/space-session-group-handlers.test.ts
@@ -3,6 +3,8 @@
  *
  * Covers:
  * - space.sessionGroup.list: happy path, missing spaceId, space not found
+ * - space.sessionGroup.create: happy path with/without members, event emission,
+ *   production guard, missing params, space not found
  * - space.sessionGroup.updateMember: happy path, missing params, space/group/member not found,
  *   cross-space group rejection, memberUpdated event emission, optional role update
  * - space.sessionGroup.delete: happy path, missing params, space/group not found,
@@ -108,6 +110,8 @@ function createMockRepo(
 	return {
 		getGroupsBySpace: mock(() => [mockGroup]),
 		getGroup: mock(() => mockGroup),
+		createGroup: mock(() => mockGroup),
+		addMember: mock(() => mockMember),
 		updateMember: mock(() => mockMember),
 		deleteGroup: mock(() => true),
 		...overrides,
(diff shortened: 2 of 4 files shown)


lsm (commented):

### Summary

This PR adds an E2E test suite for the Working Agents display in SpaceTaskPane, along with a new `space.sessionGroup.create` RPC handler to support test infrastructure. The implementation is well-structured and the tests provide good coverage of the visual states (active/completed/failed badges, real-time updates, labels, sidebar navigation).

### Recommendation: APPROVE

The implementation is solid overall. The concerns raised in the task prompt are addressed well:

**E2E Rule Compliance**: The test follows the project's E2E rules correctly. RPC calls are used for infrastructure setup (`beforeEach`/`afterEach` and pre-navigation setup), while all assertions check visible DOM state. The use of `space.sessionGroup.updateMember` in real-time update tests is justified -- there is no UI path to trigger agent completion/failure, and the test is verifying the UI's reaction to server-sent WebSocket events, which is the actual user-visible behavior.

**`space.sessionGroup.create` as infrastructure**: This handler fits the existing pattern -- `space.sessionGroup.list`, `updateMember`, and `delete` already existed as admin operations. The new `create` handler is a natural complement. It is registered alongside the other admin handlers and follows the same validation/event-emission patterns. It could legitimately serve admin/debugging use cases beyond testing.

**Assertions check DOM state**: All test assertions use `page.locator('text=...')`, CSS class selectors (`.animate-ping`), and SVG path selectors (`path[d^="M16.707"]`) -- all visible DOM state, no internal state access.

**Setup/teardown**: Each test creates a fresh space with `beforeEach` and cleans up with `afterEach`. The cleanup in `createTestSpaceWithTask` also handles leftover spaces from previous failed runs.

### Minor observations (P3)

1. **Unused `aid` parameter** (`space-session-groups.e2e.ts:131`): The `createSessionGroup` function passes `aid: agentId` into `page.evaluate` but never uses it inside the evaluate callback. It is harmless but dead code.

2. **Fragile SVG path selectors** (`space-session-groups.e2e.ts:237,263`): Assertions like `page.locator('path[d^="M16.707"]')` are tightly coupled to the SVG path data in SpaceTaskPane.tsx. If someone changes the icon library or SVG paths, these tests will break silently without an obvious connection. A `data-testid` attribute on the badge would be more resilient, but this is a minor concern given these are standard inline SVGs unlikely to change frequently.

3. **`text=active` locator could be fragile** (`space-session-groups.e2e.ts:210`): The `text=active` selector could match other text on the page (e.g., if a task status were "active" or the word appeared in a description). In practice this is mitigated by the specific test data used, but more scoped locators would be safer long-term.

lsm (commented):

---

### Summary

The E2E test and the accompanying RPC handler are well-structured and demonstrate good understanding of the codebase. However, there are several issues that need to be addressed before merging, ranging from a production-surface exposure concern to missing unit test coverage and a dead parameter.

---

### P1 — `space.sessionGroup.create` is registered unconditionally in production

**File:** `packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts`  
**File:** `packages/daemon/src/lib/rpc-handlers/index.ts` (line 372–377)

The `space.sessionGroup.create` handler is documented as "Admin / test-infrastructure operation" and is only ever called from E2E tests. However, unlike `setupTestHandlers` — which is also registered unconditionally but at least has a clear name indicating its test-only nature — `space.sessionGroup.create` lives in a handler file that is described as having dual purpose: both legitimate admin operations (`list`, `updateMember`, `delete`) and a test-only operation (`create`).

The risk here is that any browser client can call `space.sessionGroup.create` with arbitrary synthetic session IDs, creating orphaned records in the database that reference non-existent sessions. This is an incoherence injection endpoint: a caller can create a group claiming session IDs that do not exist, with arbitrary roles and statuses, without any session-existence validation.

This is consistent with the existing pattern in `test-handlers.ts` (also unconditionally registered), so this is not a regression relative to that baseline — but it does widen the attack surface of the "no auth, expose everything" design. Given the handler is documented as test infrastructure, it should either:
1. Be guard-fenced behind `NODE_ENV !== 'production'` or a feature flag, consistent with how `NEOKAI_ENABLE_SPACES_AGENT` gates the global spaces provisioner, **or**
2. Have session ID existence validation added so it cannot create groups with phantom sessions.

---

### P1 — `space.sessionGroup.create` has no unit tests

**File:** `packages/daemon/tests/unit/rpc-handlers/space-session-group-handlers.test.ts`

The unit test file covers `list`, `updateMember`, and `delete` with thorough edge-case coverage, but `space.sessionGroup.create` — the handler added in this PR — has zero unit tests. Given the handler is the core of this feature (it is what the E2E tests rely on to inject state), its logic for:
- creating the group and emitting `spaceSessionGroup.created`
- adding each member sequentially and emitting `spaceSessionGroup.memberAdded`
- returning the full group with members

...should be tested at the unit level. Missing tests for a handler that is also being used by E2E tests means neither layer verifies the other.

---

### P2 — RPC calls inside test bodies violate the spirit of CLAUDE.md E2E rules

**File:** `packages/e2e/tests/features/space-session-groups.e2e.ts` (lines 290–302, 329–341)

The "real-time update" tests call `space.sessionGroup.updateMember` directly inside the test body (not in beforeEach/afterEach) via `page.evaluate`. The inline justification is reasonable — there is no UI path to trigger agent completion/failure — and the same pattern exists in `task-lifecycle.e2e.ts` (line 339–346, where `task.setStatus` is called inside the test body for a similar reason).

However, CLAUDE.md states:

> All test actions must go through the UI: clicks, typing, navigation, keyboard shortcuts  
> **Prohibited in test actions/assertions:** `hub.request()` — no direct MessageHub RPC calls

And separately:

> **Allowed exceptions (infrastructure only):** Room create/delete in beforeEach/afterEach

The comment in the test (lines 26–29) acknowledges this tension but does not resolve it — it simply claims this is acceptable without citing a policy exception. The task-lifecycle.e2e.ts precedent is imperfect (it predates this PR) and should not be used to justify repeating the same deviation here.

The right fix is to move the group creation into beforeEach, and then rely on page reload (via `page.goto`) after the `updateMember` RPC to validate updated state — or to accept that real-time update tests for agent lifecycle transitions belong in integration/online tests rather than E2E, per the CLAUDE.md guidance: "If a test scenario can't be triggered through the UI, it belongs in daemon integration tests, not E2E."

---

### P3 — Dead parameter `agentId` in `createSessionGroup`

**File:** `packages/e2e/tests/features/space-session-groups.e2e.ts` (lines 123–156)

The `agentId` parameter at line 127 is passed as `aid` into the `page.evaluate` closure (line 154), but `aid` is never referenced inside the closure body. Member `agentId` values come from `m.agentId` instead. This is a dead parameter that should be removed.

---

### P3 — SVG path selector brittle

**File:** `packages/e2e/tests/features/space-session-groups.e2e.ts` (lines 237, 263, 306, 345)

Assertions like `page.locator('path[d^="M16.707"]`)` and `page.locator('path[d^="M4.293"]`)` couple the test to the exact SVG path data of the icons. If the icon library or icon design changes, these assertions will silently break with no helpful error message. A `data-testid` attribute on the status badge component, or a visible `aria-label`, would be more robust.

---

### What works well

- The overall design is sound: the RPC handler correctly mirrors the exact events that `TaskAgentManager` emits (`spaceSessionGroup.created`, `spaceSessionGroup.memberAdded`, `spaceSessionGroup.memberUpdated`), so the test is actually exercising the same WebSocket event pipeline that real agents use.
- Assertions all check visible DOM state (text, CSS classes, SVG elements) — not internal signal values or window globals.
- Setup/teardown is clean: space created in `beforeEach`, deleted in `afterEach`.
- The handler has solid input validation (required params, cross-space ownership checks, valid status enum).
- The "no session group" baseline test correctly verifies the section is hidden when empty.

lsm (commented):

### Summary

The PR successfully addresses the three main concerns from the previous review:
1. Added a production guard (`NODE_ENV === 'production'` throws) on `space.sessionGroup.create`
2. Added 10 unit tests for the new handler
3. Removed the two "real-time update" tests that used RPC inside test bodies
4. Removed the unused `agentId` parameter from `createSessionGroup`
5. Replaced brittle SVG path selectors with `data-testid` attributes

The E2E tests are well-structured, comply with CLAUDE.md rules, and cover the relevant display states. The RPC handler follows existing patterns. One issue remains that should be fixed.

---

### P1 -- Manual `afterEach` in unit tests will leak `NODE_ENV` on failure

**File:** `packages/daemon/tests/unit/rpc-handlers/space-session-group-handlers.test.ts`, lines 194-197

The `space.sessionGroup.create` describe block defines `afterEach` as a local `const` function (line 195) and calls it manually at the end of each test. This shadows bun:test's `afterEach` hook (which is not imported). If any test assertion throws before reaching the manual `afterEach()` call, `process.env.NODE_ENV` will not be restored, potentially contaminating subsequent test runs in the same process.

The fix is to import `afterEach` from `bun:test` and register it as a proper hook:

```ts
import { describe, expect, it, mock, beforeEach, afterEach } from 'bun:test';
```

Then inside the describe block:
```ts
afterEach(() => {
    process.env.NODE_ENV = originalNodeEnv;
});
```

This removes the need for manual `afterEach()` calls at the end of every test.

---

### P3 -- Dead `members` array in RPC handler

**File:** `packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts`, line 88

The `members` array is populated inside the loop but never referenced afterward -- the function returns `fullGroup` from `sessionGroupRepo.getGroup()` instead. This is functionally correct (getGroup returns the group with members) but the unused array is dead code.

---

### P3 -- PR body references 10 tests but only 8 exist

The PR body still lists "active -> completed" and "active -> failed" real-time update tests in its bullet points, but those tests were removed. The description should be updated to match the actual 8 tests.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 (test pollution bug) and two P2s remain.

---

### Summary of Changes

All three P1s from the previous review were addressed in good faith:
- Production guard added via `NODE_ENV === 'production'` check
- 10 unit tests added with thorough coverage across all four handlers
- The two real-time update tests with `hub.request()` inside test bodies were removed
- `agentId` parameter removed from `createSessionGroup` helper
- `data-testid` attributes added to `MemberStatusBadge` and `MemberStatusIcon`

However, there is a new P1 introduced during this revision, and two P2s that warrant attention.

---

### P1 — Unit test `afterEach` does not use a hook, allowing cross-test `NODE_ENV` pollution

**File:** `packages/daemon/tests/unit/rpc-handlers/space-session-group-handlers.test.ts`, lines 195–197

The test file defines a local plain function named `afterEach`:

```ts
const afterEach = () => {
    process.env.NODE_ENV = originalNodeEnv;
};
```

This is **not** Bun's `afterEach` hook — it is a manually called function at the bottom of each test body. If any test throws before reaching that call, `NODE_ENV` stays mutated for all subsequent tests in the `create` describe block. The dangerous case is the production guard test itself:

```ts
it('throws in production environment', async () => {
    process.env.NODE_ENV = 'production';
    await expect(handlers.get(...)!({...})).rejects.toThrow('not available in production');
    afterEach(); // only reached when assertion passes
});
```

If the assertion fails for any reason (e.g., a different error message), the manual `afterEach()` is never called, leaving `NODE_ENV = 'production'` for the remaining tests in the describe block. The fix is straightforward — import and use Bun's actual `afterEach` hook:

```ts
import { describe, expect, it, mock, beforeEach, afterEach } from 'bun:test';
// ...
afterEach(() => {
    process.env.NODE_ENV = originalNodeEnv;
});
// Remove the manual afterEach() calls from each test body.
```

---

### P2 — Dead code: `members` array is populated but never used

**File:** `packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts`, lines 88–109

```ts
const members = [];
for (let i = 0; i < (params.members ?? []).length; i++) {
    // ...
    members.push(member);  // populated but never read
}
const fullGroup = sessionGroupRepo.getGroup(group.id)!;
return { group: fullGroup };  // members array is discarded
```

The `members` array is accumulated but the return value comes from `sessionGroupRepo.getGroup()` instead, making the array dead code. Either remove it entirely or use it in the response.

---

### P2 — `createSessionGroup` is called inside test bodies, not in `beforeEach`

**File:** `packages/e2e/tests/features/space-session-groups.e2e.ts`, lines 199, 222, 247, etc.

CLAUDE.md's infrastructure exception for RPC calls is scoped to `beforeEach`/`afterEach` and global teardown:

> *Room create/delete in `beforeEach`/`afterEach` via `hub.request(...)` — accepted infrastructure pattern for test isolation*

The `createSessionGroup` helper calls `hub.request('space.sessionGroup.create', ...)` and is invoked directly inside test bodies, not in `beforeEach`. Moving these calls to per-test `beforeEach` blocks would bring them into policy compliance and separate setup from assertions.

---

- The `createTestSpaceWithTask` helper correctly uses RPC in `beforeEach` — this is the accepted infrastructure pattern.
- `data-testid` attributes on `MemberStatusBadge` and the icon SVGs are correctly placed and exercised by the tests.
- The production guard (`NODE_ENV === 'production'`) is appropriate — E2E tests run in `test` or `development` mode so the guard does not interfere with them.
- The 10 unit tests cover happy paths, error paths, event emission, and the production guard thoroughly.

lsm (commented):

### Summary

Third iteration cleanly addresses all three issues from the previous review:

1. **P1 (afterEach hook)** -- Fixed. `afterEach` is now properly imported from `bun:test` (line 14) and used as a real hook (lines 195-197) so `NODE_ENV` is always restored even when assertions throw.

2. **P2 (dead `members` array)** -- Fixed. The accumulation loop is gone; the handler now iterates with a plain `for` loop over `params.members`, calls `addMember` and emits events, then returns the result of `getGroup()` directly.

3. **P2 (RPC in test bodies)** -- Fixed. Each test scenario is now wrapped in its own `test.describe` block with a dedicated `test.beforeEach` that calls `createSessionGroup`. No RPC calls remain in test bodies -- all assertions operate on visible DOM state via Playwright locators.

### What looks good

- The `space.sessionGroup.create` RPC handler has proper validation (spaceId, name), production guard, correct event emission order (group created, then member added per member), and returns the full group via `getGroup()`.
- Unit tests are thorough: 10 tests covering happy paths, event emission, production guard, missing params, space not found, and empty members array.
- E2E tests follow project conventions: serial mode, desktop viewport, infrastructure RPC only in hooks, all assertions on DOM state via `data-testid` attributes and text content.
- The nested `test.describe` structure is clean and each sub-suite is self-contained with its own setup.
- `data-testid` attributes on `MemberStatusBadge` and SVG icons are minimal and well-targeted.

- The PR body still mentions "Real-time update: active -> completed without page refresh" and "Real-time update: active -> failed without page refresh" tests, but those were removed (correctly, per CLAUDE.md policy). The PR body is slightly out of date.

No blocking issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### Third-iteration verification — all previous issues resolved

This is a re-review of iteration 3. All issues raised in the previous review have been addressed correctly.

---

#### P1 (previous) — Production guard missing → FIXED

`packages/daemon/src/lib/rpc-handlers/space-session-group-handlers.ts` now guards the handler at line 47:

```ts
if (process.env.NODE_ENV === 'production') {
    throw new Error('space.sessionGroup.create is not available in production');
}
```

This matches the intent to block production misuse. The unit test at line 291–299 of the test file explicitly verifies this guard throws.

#### P1 (previous) — No unit tests for `space.sessionGroup.create` → FIXED

`packages/daemon/tests/unit/rpc-handlers/space-session-group-handlers.test.ts` now contains a full `describe('space.sessionGroup.create')` block (lines 174–330) with 9 tests covering: happy path with/without members, event emission for `spaceSessionGroup.created` and `spaceSessionGroup.memberAdded`, the production guard, missing param validation (`spaceId`, `name`), space-not-found, empty members array, and full group return. Coverage is thorough.

The `afterEach` in this describe block correctly restores `process.env.NODE_ENV` using Bun's real `afterEach()` imported from `bun:test` — the exact fix requested.

#### P2 (previous) — RPC calls in test bodies → FIXED

The restructuring into nested `test.describe` blocks is correct. Each badge-state scenario has its own `test.beforeEach` that calls `createSessionGroup`, and test bodies only contain `page.goto`, `waitForWebSocketConnected`, and DOM assertions. The `updateMember`-based real-time update tests have been removed and replaced with static state injection, which is consistent with CLAUDE.md guidance that agent lifecycle transitions belong in daemon integration tests if they cannot be triggered via UI.

#### P3 (previous) — Dead `agentId` parameter in `createSessionGroup` → FIXED

The helper signature is now `(page, spaceId, taskId, members[])` with no dead `aid` variable.

#### P3 (previous) — Fragile SVG path selectors → FIXED

`data-testid="member-status-icon-completed"` and `data-testid="member-status-icon-failed"` have been added to the SVG elements in `SpaceTaskPane.tsx`, and the test assertions use `page.getByTestId(...)` accordingly.

---

### Additional checks performed

**Timing / race conditions**: `createSessionGroup` is called in `beforeEach` before `page.goto`. The handler persists the group to SQLite synchronously, then emits WebSocket events. On subsequent `page.goto`, `fetchInitialState` calls `space.sessionGroup.list` which reads from the DB — the persisted data is always present. No race condition.

**Serial mode**: `test.describe.configure({ mode: 'serial' })` is correct given shared mutable outer variables (`spaceId`, `agentId`, `taskId`).

**Cascade deletion**: Outer `afterEach` calls `space.delete`, which cascades to session groups. Inner sub-describe tests do not need their own group cleanup.

**`window.appState?.messageHub` fallback**: `appState` does not have a `messageHub` property (only `window.__messageHub` is exposed in `connection-manager.ts`), so the fallback `|| window.appState?.messageHub` is dead code. This is harmless since the primary accessor works, and this pattern appears in other E2E helpers in the codebase — it is not a regression introduced here.

**Badge text assertions**: `toContainText('active')`, `toContainText('completed')`, `toContainText('failed')` are scoped to elements with `data-testid="member-status-badge-{status}"`, so they cannot match unrelated page text. The `text=coder` locator in the named-agent test (line 349) could theoretically match other occurrences of the word "coder" on the page, but in this test context (a clean space with one task and one session group) this is not a practical concern.

---

### Conclusion

All P1 and P2 issues from the previous review are correctly resolved. The implementation is clean, well-tested, and compliant with the project's E2E rules. The test suite exercises all six documented scenarios (no-groups baseline, three badge states, multiple members, task-agent label, named agent label, sidebar navigation) through the correct infrastructure pattern.

PR #688, opened 2026-03-22
fix: remove emitTaskUpdate from task.fail RPC handler (Task 3.1)

lsm:
LiveQuery deltas now deliver task changes via notifyChange('tasks') →
LiveQueryEngine → sendToClient, making the handler-layer room.task.update
emit in task.fail redundant.

- Remove emitTaskUpdate() call from task.fail handler; emitRoomOverview()
  stays because room/session metadata still flows through that channel
- Update existing unit test: assert room.task.update is NOT emitted by
  task.fail, and room.overview IS still emitted
- Add online integration test (Test 5 in rpc-live-query.test.ts): verifies
  that task.fail triggers a liveQuery.delta with status needs_attention for
  a subscribed tasks.byRoom client

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
@@ -215,7 +215,6 @@ export function setupTaskHandlers(
 		const taskManager = taskManagerFactory(db, params.roomId);
 		const task = await taskManager.failTask(params.taskId, params.error ?? '');
 
-		emitTaskUpdate(params.roomId, task);
 		emitRoomOverview(params.roomId);
 
 		return { task };
--- a/packages/daemon/tests/online/rpc/rpc-live-query.test.ts
+++ b/packages/daemon/tests/online/rpc/rpc-live-query.test.ts
@@ -280,4 +280,68 @@ describe('LiveQuery — end-to-end reactive pipeline', () => {
 			await daemon.messageHub.request('liveQuery.unsubscribe', { subscriptionId: subId2 });
 		}
 	}, 20_000);
+
+	// -----------------------------------------------------------------------
+	// Test 5: task.fail triggers liveQuery.delta (no handler-layer emit needed)
+	// -----------------------------------------------------------------------
+
+	test('task.fail RPC triggers a liveQuery.delta for tasks.byRoom subscriber', async () => {
+		const roomId = await createRoom('task-fail-delta');
+		const subId = 'sub-task-fail-1';
+
+		// First create a task to fail
+		const task = await createTask(roomId, 'Task To Fail');
+
+		const snapshots: LiveQuerySnapshotEvent[] = [];
+		const deltas: LiveQueryDeltaEvent[] = [];
+
+		const unsubSnap = daemon.messageHub.onEvent<LiveQuerySnapshotEvent>(
+			'liveQuery.snapshot',
+			(ev) => {
+				if (ev.subscriptionId === subId) snapshots.push(ev);
+			}
+		);
+		const unsubDelta = daemon.messageHub.onEvent<LiveQueryDeltaEvent>('liveQuery.delta', (ev) => {
+			if (ev.subscriptionId === subId) deltas.push(ev);
+		});
+
+		try {
+			// Subscribe after task is created
+			await daemon.messageHub.request('liveQuery.subscribe', {
+				queryName: 'tasks.byRoom',
+				params: [roomId],
+				subscriptionId: subId,
+			});
+
+			// Wait for snapshot
+			await waitFor(() => snapshots.length > 0, DELTA_WAIT_TIMEOUT_MS, 'snapshot before fail');
+			const snapshotVersion = snapshots[0].version;
+
+			// Fail the task — triggers notifyChange('tasks') via TaskManager, no handler-layer emit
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped change. The single-line removal in `task-handlers.ts` is correct and safe because the LiveQuery pipeline already handles task change propagation through `TaskManager.updateTaskStatus` -> `reactiveDb.notifyChange('tasks')` -> `LiveQueryEngine` -> client deltas.

### Verification

- **`failTask()` calls `updateTaskStatus()`** which calls `this.reactiveDb.notifyChange('tasks')` (line 135 of `task-manager.ts`), confirming the LiveQuery delta pipeline covers this path.
- **`emitRoomOverview()` is correctly retained** -- it serves a different purpose (session/room metadata) not covered by task LiveQuery deltas.
- **Other handlers still use `emitTaskUpdate`** (task.cancel, task.archive, task.update, etc.) -- this PR only removes it from `task.fail`, consistent with an incremental migration approach.

### Unit test update

The updated unit test at line 440 of `task-handlers.test.ts` asserts `room.task.update` is NOT emitted by filtering mock calls. This is a precise and correct approach.

### Online test (Test 5)

The new online test in `rpc-live-query.test.ts` is well-structured:
- Creates a task, subscribes to `tasks.byRoom`, fails the task, and verifies a delta arrives with `status: 'needs_attention'`.
- Proper cleanup in `finally` block (unsubscribe listeners + liveQuery.unsubscribe).
- Version monotonicity assertion (`delta.version > snapshotVersion`) is a good sanity check.
- The test correctly checks both `added` and `updated` arrays in the delta, which handles the case where the engine might classify the change either way.

No issues found. Clean implementation with good test coverage of both the negative case (unit test: event NOT emitted) and the positive case (online test: LiveQuery delta IS delivered).

lsm (commented):

**Recommendation: APPROVE**

This is a narrow, well-scoped change that removes one redundant event emission from the `task.fail` RPC handler as part of a planned LiveQuery migration. The implementation is correct.

---

### What was done

- Removed the single `emitTaskUpdate()` call from the `task.fail` handler in `packages/daemon/src/lib/rpc-handlers/task-handlers.ts`
- Kept `emitRoomOverview()` intact, which is required to push room/session metadata to the frontend
- Updated the unit test in `packages/daemon/tests/unit/rpc/task-handlers.test.ts` to assert that `room.task.update` is NOT emitted and `room.overview` IS still emitted
- Added an online integration test (Test 5 in `packages/daemon/tests/online/rpc/rpc-live-query.test.ts`) that verifies the LiveQuery delta pipeline delivers the `status: 'needs_attention'` change after `task.fail`

---

### Correctness

The change is logically correct. `TaskManager.failTask()` calls `updateTaskStatus()`, which calls `this.reactiveDb.notifyChange('tasks')`. That notification flows through the LiveQuery engine and delivers a delta to subscribed clients. Removing the duplicate `emitTaskUpdate()` from the handler layer does not break delivery of the task state change to LiveQuery subscribers.

The seven remaining `emitTaskUpdate` call sites in the file are intentionally preserved per the plan, because the frontend's `room-store.ts` and `TaskView.tsx` still listen on `room.task.update` via `hub.onEvent` until Task 3.3 replaces those listeners with LiveQuery subscriptions.

---

### Tests

**Unit test** (task-handlers.test.ts): The updated test correctly distinguishes between `room.task.update` and `room.overview` by filtering on the first argument of the mock's call list. Mock isolation is correct — `daemonHubData.emit` is recreated fresh in each `beforeEach` via `createMockDaemonHub()`, so there is no cross-test pollution.

**Online test** (rpc-live-query.test.ts): Test 5 correctly subscribes before failing the task, captures the snapshot version, calls `task.fail`, and asserts the delta arrives with a version greater than the snapshot and includes the task in either `added` or `updated`. The combined-array check (`[...(delta.added ?? []), ...(delta.updated ?? [])]`) is appropriate because the task was in the snapshot, so it should appear in `updated` — but this defensive approach handles both cases without masking bugs.

The 8-second `DELTA_WAIT_TIMEOUT_MS` and 20-second test timeout are consistent with the existing tests in the file.

---

**P3 — Test 5 comment ordering:** In Test 5, the event listeners (`unsubSnap`, `unsubDelta`) are registered before `liveQuery.subscribe` is called, but the task is created before the listeners are set up. The code works correctly because the subscription snapshot is gated by `waitFor(() => snapshots.length > 0)`. The comment `// Subscribe after task is created` accurately describes the intent. No issue here.

**P3 — No room cleanup in online tests:** Rooms created in Test 5 are not deleted after the test, but this is consistent with the pre-existing pattern in Tests 1–4 of the same file. Not a regression introduced by this PR.

**P3 — Plan checklist item coverage:** The plan document (Task 3.1) lists "Test: `goal.created`, `goal.updated`, `goal.progressUpdated` still fire from `goal-handlers.ts`" as a checklist item. This is satisfied by pre-existing tests in `goal-handlers.test.ts` rather than new tests in this PR. The intent of the checklist item is to ensure nothing was broken — verifying that the existing tests continue to pass achieves this without requiring redundant new tests.

---

### Summary

The implementation matches the task specification exactly. The one-line deletion is correct, test coverage is solid with both a unit test and an online integration test, and the scope is properly bounded to `task.fail` only. No issues found that would block merging.

PR #689, opened 2026-03-22
feat: add useGroupMessageLiveQuery hook for LiveQuery-based group message streaming

lsm:
Introduces useGroupMessageLiveQuery, a Preact hook that subscribes to the
sessionGroupMessages.byGroup named query via the LiveQuery protocol.

- Calls liveQuery.subscribe on mount; generates a UUID subscriptionId per
  subscription for stale-event guards during rapid task switching
- Handles liveQuery.snapshot: replaces message list; stale guard discards
  events from superseded subscriptions
- Handles liveQuery.delta: appends from added array only (append-only
  invariant); stale guard applied
- Calls liveQuery.unsubscribe (fire-and-forget) on unmount or groupId change
- null groupId: clears messages immediately, no subscription created
- 15 Vitest tests covering all lifecycle paths

Exports hook and types from hooks/index.ts.

### DIFF
--- a/packages/web/src/components/room/TaskConversationRenderer.tsx
+++ b/packages/web/src/components/room/TaskConversationRenderer.tsx
@@ -17,6 +17,7 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/hooks';
 import type { SDKMessage } from '@neokai/shared/sdk/sdk.d.ts';
 import type { SessionInfo } from '@neokai/shared';
+import type { GroupMessage } from '../../types/group-message';
 import { useMessageHub } from '../../hooks/useMessageHub';
 import {
 	useSessionQuestionState,
@@ -39,16 +40,6 @@ interface TaskMeta {
 	authorSessionId: string;
 	turnId: string;
 	iteration: number;
-}
-
-interface GroupMessage {
-	id: number;
-	groupId: string;
-	sessionId: string | null;
-	role: string;
-	messageType: string;
-	content: string;
-	createdAt: number;
 }
 
 interface TaskConversationRendererProps {
--- a/packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts
+++ b/packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts
@@ -1,0 +1,400 @@
+/**
+ * Unit tests for useGroupMessageLiveQuery hook
+ *
+ * Verifies that the hook:
+ * - Subscribes to `sessionGroupMessages.byGroup` on mount with a unique subscriptionId
+ * - Sets loading=true until the snapshot event arrives
+ * - Replaces messages on liveQuery.snapshot
+ * - Appends added messages on liveQuery.delta (ignores updated/removed)
+ * - Discards snapshot/delta events from prior subscriptions (stale-subscriptionId guard)
+ * - Calls liveQuery.unsubscribe on unmount
+ * - Re-subscribes when groupId changes (unsubscribes old first)
+ * - Handles null groupId: clears messages, does not subscribe
+ * - Surfaces subscribe RPC errors via error state
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { renderHook, act, cleanup, waitFor } from '@testing-library/preact';
+import type { LiveQuerySnapshotEvent, LiveQueryDeltaEvent } from '@neokai/shared';
+import { useGroupMessageLiveQuery, type GroupMessage } from '../useGroupMessageLiveQuery';
+
+// -------------------------------------------------------
+// Mock helpers
+// -------------------------------------------------------
+
+// Handlers captured per event name so tests can fire them
+const snapshotHandlers: Array<(event: LiveQuerySnapshotEvent) => void> = [];
+const deltaHandlers: Array<(event: LiveQueryDeltaEvent) => void> = [];
+
+const mockRequest = vi.fn();
+const mockOnEvent = vi.fn(
+	(eventName: string, handler: (event: LiveQuerySnapshotEvent | LiveQueryDeltaEvent) => void) => {
+		if (eventName === 'liveQuery.snapshot') {
+			snapshotHandlers.push(handler as (e: LiveQuerySnapshotEvent) => void);
+		} else if (eventName === 'liveQuery.delta') {
+			deltaHandlers.push(handler as (e: LiveQueryDeltaEvent) => void);
+		}
+		// Return a no-op unsubscribe for simplicity; tests don't need to track these
+		return () => {};
+	}
+);
+
+vi.mock('../useMessageHub.ts', () => ({
+	useMessageHub: () => ({
+		request: mockRequest,
+		onEvent: mockOnEvent,
+	}),
+}));
+
+// -------------------------------------------------------
+// Helpers
+// -------------------------------------------------------
+
+function makeMessage(id: number, groupId = 'grp-1'): GroupMessage {
+	return {
+		id,
+		groupId,
+		sessionId: null,
+		role: 'system',
+		messageType: 'status',
+		content: `message-${id}`,
+		createdAt: 1000 + id,
+	};
+}
+
+/** Fire a snapshot event to all registered snapshot handlers */
+function fireSnapshot(subscriptionId: string, rows: GroupMessage[]): void {
+	const event: LiveQuerySnapshotEvent = { subscriptionId, rows, version: 1 };
+	for (const h of snapshotHandlers) {
+		h(event);
+	}
+}
+
+/** Fire a delta event to all registered delta handlers */
+function fireDelta(subscriptionId: string, added?: GroupMessage[]): void {
+	const event: LiveQueryDeltaEvent = {
+		subscriptionId,
+		added,
+		version: 2,
+	};
+	for (const h of deltaHandlers) {
+		h(event);
+	}
+}
+
+/** Extract the subscriptionId from the most recent liveQuery.subscribe call */
+function lastSubscribeId(): string {
+	const calls = mockRequest.mock.calls;
+	for (let i = calls.length - 1; i >= 0; i--) {
+		const [method, data] = calls[i] as [string, { subscriptionId?: string }];
+		if (method === 'liveQuery.subscribe') {
+			return data.subscriptionId ?? '';
+		}
+	}
+	return '';
+}
+
+// -------------------------------------------------------
+// Setup / teardown
+// -------------------------------------------------------
+
+beforeEach(() => {
+	snapshotHandlers.length = 0;
+	deltaHandlers.length = 0;
+	mockRequest.mockResolvedValue({ ok: true });
+	mockOnEvent.mockClear();
+	mockRequest.mockClear();
+});
+
+afterEach(() => {
+	cleanup();
+});
+
+// -------------------------------------------------------
+// Tests
+// -------------------------------------------------------
+
+describe('useGroupMessageLiveQuery', () => {
+	it('calls liveQuery.subscribe on mount with correct params', async () => {
+		renderHook(() => useGroupMessageLiveQuery('grp-1'));
+
+		await waitFor(() => {
+			expect(mockRequest).toHaveBeenCalledWith(
+				'liveQuery.subscribe',
+				expect.objectContaining({
+					queryName: 'sessionGroupMessages.byGroup',
+					params: ['grp-1'],
+					subscriptionId: expect.any(String),
+				})
+			);
+		});
+	});
+
+	it('registers snapshot and delta event listeners before subscribing', () => {
+		renderHook(() => useGroupMessageLiveQuery('grp-1'));
+
+		// onEvent should be called for both snapshot and delta
+		const eventNames = mockOnEvent.mock.calls.map((args) => args[0] as string);
+		expect(eventNames).toContain('liveQuery.snapshot');
+		expect(eventNames).toContain('liveQuery.delta');
+	});
+
+	it('starts with loading=true', () => {
+		const { result } = renderHook(() => useGroupMessageLiveQuery('grp-1'));
+		expect(result.current.loading).toBe(true);
+		expect(result.current.messages).toEqual([]);
+	});
+
+	it('replaces messages and clears loading on snapshot', async () => {
+		const { result } = renderHook(() => useGroupMessageLiveQuery('grp-1'));
+
+		await waitFor(() => expect(mockRequest).toHaveBeenCalled());
+		const subId = lastSubscribeId();
+
+		act(() => {
+			fireSnapshot(subId, [makeMessage(1), makeMessage(2)]);
+		});
+
+		expect(result.current.loading).toBe(false);
+		expect(result.current.messages).toHaveLength(2);
+		expect(result.current.messages[0].id).toBe(1);
+		expect(result.current.messages[1].id).toBe(2);
+	});
+
+	it('appends added messages on delta', async () => {
+		const { result } = renderHook(() => useGroupMessageLiveQuery('grp-1'));
+
+		await waitFor(() => expect(mockRequest).toHaveBeenCalled());
+		const subId = lastSubscribeId();
+
+		act(() => {
+			fireSnapshot(subId, [makeMessage(1)]);
+		});
+		expect(result.current.messages).toHaveLength(1);
+
+		act(() => {
+			fireDelta(subId, [makeMessage(2), makeMessage(3)]);
+		});
+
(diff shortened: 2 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured hook with solid lifecycle management and thorough test coverage. The implementation correctly follows the LiveQuery protocol patterns established in the daemon. A few observations below.

### Positive highlights

- Registering event listeners *before* calling `liveQuery.subscribe` is the correct ordering to avoid race conditions where the server sends the snapshot before the RPC promise resolves.
- The dual stale-guard approach (closure-captured `subscriptionId` + `cancelled` flag) is robust against both stale events and cleanup-after-unmount scenarios.
- Fire-and-forget unsubscribe with `.catch(() => {})` is appropriate -- the server cleans up on disconnect anyway.
- Test coverage is comprehensive: 15 tests covering mount, unmount, groupId changes, null groupId, stale guards, error handling, and delta append-only semantics.

### P2: Duplicate `GroupMessage` interface

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts:24-32`
**File:** `packages/web/src/components/room/TaskConversationRenderer.tsx:44-52`

The `GroupMessage` interface is defined identically in both files. This creates a maintenance burden -- if the schema changes, both must be updated. Consider extracting the shared type to a common location (e.g., a types file in `packages/web/src/types/` or in `@neokai/shared`) and importing it from both places.

### P3: `activeSubIdRef` is unused for its intended purpose

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts:53`

The `activeSubIdRef` is set on line 65 and conditionally cleared on line 105, but it is never *read* by any code. The stale-event guard works entirely through the closure-captured `subscriptionId` local variable and the `cancelled` flag. The ref is dead code that adds confusion about which mechanism actually provides the stale guard. It could be removed without any behavioral change.

That said, it might be intentionally left as a hook for future consumers (e.g., exposing the current subscription ID), so this is a nit rather than a bug.

### P3: `error` state not cleared when a new snapshot arrives

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts:75-78`

If a subscribe call fails (setting `error`), and then the user switches groups (which calls `setError(null)` at line 68), everything is fine. However, if the error were somehow transient and a snapshot *did* arrive for the same subscription, the error would persist alongside the snapshot data. This is extremely unlikely given the current protocol (error means no snapshot will come), but worth noting for robustness.

### Summary

Clean, well-documented implementation with proper lifecycle management. The test suite is thorough and the code integrates correctly with the existing `useMessageHub` hook and `LiveQuery` types. The only meaningful suggestion is deduplicating the `GroupMessage` type (P2). The rest are minor nits.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — two P1 issues need addressing before merge.

---

## Summary

The hook implementation is clean and its core lifecycle logic (subscribe-on-mount, snapshot-replace, delta-append, unsubscribe-on-cleanup, stale-guard by subscriptionId) is correct. The dependency array is appropriate, and the `cancelled` closure prevents state updates after unmount. However, the test suite has a structural flaw that masks whether listener unsubscription actually works, and the hook silently stalls forever when the WebSocket reconnects after a disconnect.

---

## P1 — Test mock makes unsubscription non-verifiable

**File:** `packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts`

The `mockOnEvent` returns a no-op unsub function:

```ts
return () => {};  // line 38
```

This means `unsubSnapshot()` and `unsubDelta()` called in the effect cleanup (hook lines 109–110) never remove the captured handlers from `snapshotHandlers` / `deltaHandlers`. Every re-subscription call in tests such as *"unsubscribes old and subscribes new when groupId changes"* (line 263) just accumulates more handlers in the shared arrays.

The tests pass because the stale-guard (`event.subscriptionId !== subscriptionId`) in the hook discards old handlers — but that is testing the stale-guard, not listener removal. No test verifies that after unmount or groupId change, the old event handlers are actually deregistered from the event bus. If the real `onEvent` implementation had a bug in its unsub path, these tests would not catch it.

**Fix:** Return a real unsub mock that splices from the captured arrays, e.g.:

```ts
const mockOnEvent = vi.fn((eventName, handler) => {
    if (eventName === 'liveQuery.snapshot') snapshotHandlers.push(handler);
    else if (eventName === 'liveQuery.delta')  deltaHandlers.push(handler);
    return () => {
        // actually remove so unsubscription is observable
        if (eventName === 'liveQuery.snapshot') {
            const i = snapshotHandlers.indexOf(handler);
            if (i !== -1) snapshotHandlers.splice(i, 1);
        } else {
            const i = deltaHandlers.indexOf(handler);
            if (i !== -1) deltaHandlers.splice(i, 1);
        }
    };
});
```

Then add an assertion in the groupId-change and unmount tests that `snapshotHandlers` / `deltaHandlers` contain only 1 entry after the switch, confirming the old listener was removed.

---

## P1 — Hook silently stalls after WebSocket reconnection

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts`

The hook calls `liveQuery.subscribe` once on mount (or groupId change). If the WebSocket disconnects and reconnects, the server drops all subscriptions for the old connection. The hook's event listeners remain registered in the new hub instance (via `onEvent`'s queuing mechanism in `useMessageHub`), but the server-side subscription is gone — so no more `liveQuery.snapshot` or `liveQuery.delta` events will arrive. The UI silently shows stale messages and `loading` stays `false` with no indication of the lost subscription.

The `isConnected` reactive value from `useMessageHub` is not used in the hook at all. Other parts of the codebase observe connection state to re-initialize (e.g., `useChatBase` watches reconnect events). Adding `isConnected` (or the raw `connectionState` signal) to the effect dependency array, or subscribing to a reconnect event to reissue the subscribe RPC, would handle this case.

This is a functional gap that will manifest in production whenever the WebSocket cycles (network blip, server restart, etc.).

---

## P2 — `loading` is never reset to `true` after an RPC error + retry

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts` (lines 67–69, 91–100)

If `liveQuery.subscribe` fails (`error` is set, `loading` becomes `false`), and the caller then changes `groupId` back to the same value, React/Preact will not re-run the effect because `groupId` has not changed. There is no mechanism to retry after a transient error without toggling `groupId`. Consider exposing a `retry` callback or documenting this limitation.

(Minor — `loading` is correctly reset to `true` on groupId change; this specific edge is same-groupId retry.)

---

## P2 — Unsafe cast on snapshot rows

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts`, line 77

```ts
setMessages((event.rows as GroupMessage[]) ?? []);
```

`LiveQuerySnapshotEvent.rows` is typed `unknown[]`. The cast is unguarded — if the server changes the schema or a bug delivers unexpected fields, the UI silently consumes malformed data. A runtime shape-check or at minimum a comment acknowledging the trust boundary would be appropriate. Same pattern applies to the delta cast on line 85.

This is consistent with other hooks in the codebase, so not a blocking issue, but worth flagging as a future hardening point.

---

## P2 — Missing test: behavior when `liveQuery.subscribe` resolves but no snapshot arrives

**File:** `packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts`

No test covers the case where the RPC resolves successfully (`{ ok: true }`) but the snapshot event never fires (e.g., server bug, race during disconnect). In this state `loading` stays `true` forever. A test verifying that `loading` stays `true` in this case (and that the UX is therefore an indefinite spinner) would document the known behavior and make a future timeout/retry mechanism easier to test.

---

## P3 — Handler accumulation is invisible across test cases

**File:** `packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts`, lines 26–27, 101–106

The `snapshotHandlers` / `deltaHandlers` arrays are module-level and cleared via `.length = 0` in `beforeEach`. `mockOnEvent.mockClear()` clears call records but not the arrays — this is handled by the explicit `.length = 0`. The setup is correct but easy to misread. A comment explaining why both clears are needed (one for vi.fn call records, one for the captured handler arrays) would prevent future contributors from accidentally removing the `.length = 0` lines thinking mockClear covers it.

---

## P3 — `version: 1` / `version: 2` hardcoded in test helpers

**File:** `packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts`, lines 67, 76

The snapshot helper always uses `version: 1` and delta uses `version: 2`. Since the hook ignores `version` today this is harmless, but if a future version ordering check is added these tests would pass incorrectly. Using a monotonic counter would be more future-proof.

lsm (commented):

**Recommendation: APPROVE** — All P1 and P2 issues from the previous review have been correctly addressed. One structural issue with the reconnect mock remains worth noting, but it does not block merge.

---

## Summary

The second-round changes are solid. The worker correctly addressed every prior concern:

- The `mockOnEvent` real-unsub fix is correct: handlers are spliced from the captured arrays on cleanup, and new tests verify that exactly 0 handlers remain after unmount and that the count stays at 1 after groupId change.
- The reconnect effect is well-structured: it uses `prevConnectedRef` to detect false→true transitions, reads `activeSubIdRef`/`activeGroupIdRef` to re-issue the subscribe RPC with the same subscriptionId, and correctly skips re-subscribe when groupId is null.
- The shared `GroupMessage` type is now in `packages/web/src/types/group-message.ts` and imported by both `TaskConversationRenderer.tsx` and `useGroupMessageLiveQuery.ts`. No duplication.
- `setError(null)` in the snapshot handler clears prior subscribe errors.
- The monotonic `versionCounter` is reset to 0 in `beforeEach`, which is correct.

---

## Issues Found

### P2: Mock `isConnected` captures value at factory call time, not reactively

**File:** `packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts:61-67`

```ts
vi.mock('../useMessageHub.ts', () => ({
  useMessageHub: () => ({
    request: mockRequest,
    onEvent: mockOnEvent,
    isConnected: mockIsConnected.value,   // evaluated at every hook call
  }),
}));
```

`mockIsConnected.value` is read each time `useMessageHub()` is called (i.e., on every render), so the mock does return the current signal value. This works because Preact re-renders the component when `mockIsConnected.value` changes — **but only if the component is actually subscribed to the signal**. In Preact Signals, reading `.value` inside a component render function creates an implicit subscription. However, the mock factory arrow `() => ({ isConnected: mockIsConnected.value })` is called by the hook consumer's render, so this does work for functional components.

The subtle risk is that `vi.mock` hoists the factory outside of any signal tracking context. Reading `mockIsConnected.value` inside the factory creates no reactive subscription at the module level — it only works because `useMessageHub()` is called again on each render triggered by an external cause. If the reconnect effect fires as a result of the signal change, it needs the hook to re-render first to pick up the new `isConnected` value.

This is actually correct in practice for these tests because `act(() => { mockIsConnected.value = false; })` causes a re-render via Preact's signal subscription propagation to any component that read `.value`. The test does rely on Preact's signal-to-render integration being active in the test environment, which it is via `@testing-library/preact`. However, the implementation of `useGroupMessageLiveQuery` reads `isConnected` as a plain boolean from `useMessageHub()` and then puts it in the dependency array of the reconnect effect — meaning a signal value change will only trigger the effect if the component re-renders. The tests exercise this path correctly via `act()`.

No action required; this is an observation rather than a bug. The tests pass for the right reasons.

### P2: Reconnect effect silently drops errors; no loading reset on reconnect

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts:140-147`

```ts
request<{ ok: true }>('liveQuery.subscribe', { ... }).catch(() => {
  // Reconnect re-subscribe failure is transient; ignore...
});
```

When the reconnect re-subscribe fails, the hook stays in whatever state it was in before reconnect (e.g., showing stale messages). There is no test covering this scenario and no UI signal that the re-subscribe silently failed. If reconnects repeatedly fail, the user sees stale data with no indication of the problem.

This is a judgment call — fire-and-forget on reconnect is a deliberate design choice documented in the comment. But the comment's claim that "the next reconnect cycle will retry automatically" is only true if the user disconnects and reconnects again; a persistent server-side error would not self-heal. Consider at minimum setting `setError(...)` on reconnect failure so the UI can surface it.

This was not raised in the prior review so it is new feedback.

### P3: `loading` is not reset to `true` on reconnect

**File:** `packages/web/src/hooks/useGroupMessageLiveQuery.ts:128-148`

After reconnect, the hook re-subscribes with the same subscriptionId, but `loading` stays `false` (from the prior snapshot). When the server delivers the fresh snapshot after reconnect, it will replace messages atomically — so there is no visual gap in practice. However, there is a window between reconnect and snapshot delivery where messages shown to the user may be stale. Not resetting `loading` to `true` during this window is a reasonable trade-off (avoids flicker), and the existing tests do not test for this case. This is noted only for completeness.

### P3: No test covers reconnect re-subscribe failure

**File:** `packages/web/src/hooks/__tests__/useGroupMessageLiveQuery.test.ts`

All four reconnect tests assume the re-subscribe RPC succeeds. A test for `mockRequest.mockRejectedValueOnce(...)` during a reconnect cycle would document the intended silent-ignore behavior and protect it from being accidentally changed to throw.

---

## Verified Fixes

| Prior Issue | Status |
|---|---|
| P1: `mockOnEvent` returned no-op unsub — listener removal untestable | Fixed: real splice-based unsub, two new tests verify removal |
| P1: No reconnect support — hook stalls on reconnect | Fixed: second `useEffect` on `isConnected`, 4 new tests |
| P2: Duplicate `GroupMessage` type | Fixed: shared `packages/web/src/types/group-message.ts` |
| P2: `setError(null)` missing in snapshot handler | Fixed: `setError(null)` added at line 80 |
| P3: Monotonic version counter | Fixed: `versionCounter` reset per test |
| P3: `activeSubIdRef` comment | Fixed: comment explains dual-clear purpose |

All prior P1/P2 issues are resolved. The implementation is correct, well-tested, and integrates cleanly.

lsm (commented):

**Recommendation: APPROVE**

Second-round review after addressing prior feedback. All P1 and P2 issues from the previous review have been properly resolved.

### What was fixed

1. **P1 - Listener removal (fixed):** `mockOnEvent` now returns real unsub functions that splice handlers from the captured arrays. Tests at lines 314-327 and 366-388 verify handlers are actually removed on unmount and groupId change. The hook implementation correctly calls `unsubSnapshot()` and `unsubDelta()` in the cleanup function (line 113-114).

2. **P1 - Reconnect support (fixed):** A second `useEffect` (lines 128-148) detects `false→true` transitions on `isConnected` using `prevConnectedRef` and re-issues `liveQuery.subscribe` with the **same** `subscriptionId`. This is a good design choice -- reusing the subscriptionId means existing snapshot/delta handlers continue to work without re-registration. Four new tests (lines 492-575) cover reconnect behavior comprehensively.

3. **P2 - Shared type (fixed):** `GroupMessage` is now defined once in `packages/web/src/types/group-message.ts` and imported by both `TaskConversationRenderer.tsx` and `useGroupMessageLiveQuery.ts`.

4. **P2 - Error clearing (fixed):** `setError(null)` is called in the snapshot handler (line 80), clearing any prior subscribe errors when a snapshot arrives.

### Verification of correctness

- **Reconnect effect initialization:** `prevConnectedRef` is initialized with `isConnected` at ref creation time (first render). On initial mount with `isConnected=true`, the effect correctly skips (because `wasConnected=true`), avoiding a spurious re-subscribe.

- **Effect dependency stability:** `request` depends on `[defaultTimeout]` and `onEvent` depends on `[]` in `useMessageHub`, so neither causes spurious re-runs of the main subscription effect.

- **Stale subscription guard:** Both snapshot and delta handlers check `event.subscriptionId !== subscriptionId` before processing, preventing data from a prior subscription from leaking into the current state.

- **Cleanup ordering:** The cleanup function sets `cancelled = true` before calling unsub functions and issuing `liveQuery.unsubscribe`, which is the correct order to prevent race conditions with in-flight events.

- **Ref coordination between effects:** The main effect writes `activeSubIdRef` and `activeGroupIdRef`; the reconnect effect reads them. This avoids adding `groupId` to the reconnect effect's dependency array, which would cause it to interfere with the main effect's subscription lifecycle.

1. **No loading reset on reconnect (line 140-147):** When the reconnect effect re-subscribes, it does not set `loading=true`. This means the UI continues showing stale messages until the fresh snapshot arrives, which is arguably the better UX (no flicker). Just noting this as a conscious design choice.

2. **Reconnect error silently swallowed (line 144):** The `.catch(() => {})` on the reconnect subscribe is documented as intentional ("next reconnect cycle will retry"), which is reasonable since the reconnect effect will fire again on the next connection cycle.

No P0, P1, or P2 issues found. The implementation is clean and well-tested.

lsm:
Superseded by https://github.com/lsm/neokai/pull/690

PR #690, opened 2026-03-22
feat: add useGroupMessages hook for LiveQuery group message streaming

lsm:
Implements Task 4.1: a Preact hook that subscribes to
sessionGroupMessages.byGroup via the LiveQuery RPC protocol.

- Subscribes on mount (or when groupId changes), unsubscribes on cleanup
- liveQuery.snapshot replaces the full message list (stale-subId guard)
- liveQuery.delta appends only added rows (append-only invariant)
- Generates unique subscriptionId per subscription to guard stale events
- 20 Vitest tests covering snapshot, delta, stale-guard, and cleanup paths
- Exported from packages/web/src/hooks/index.ts

### DIFF
--- a/packages/web/src/hooks/__tests__/index.test.ts
+++ b/packages/web/src/hooks/__tests__/index.test.ts
@@ -14,6 +14,7 @@ import {
 	useCommandAutocomplete,
 	useInterrupt,
 	useFileAttachments,
+	useGroupMessages,
 } from '../index.ts';
 
 describe('Hooks Index', () => {
@@ -56,6 +57,11 @@ describe('Hooks Index', () => {
 		it('should export useFileAttachments', () => {
 			expect(useFileAttachments).toBeDefined();
 			expect(typeof useFileAttachments).toBe('function');
+		});
+
+		it('should export useGroupMessages', () => {
+			expect(useGroupMessages).toBeDefined();
+			expect(typeof useGroupMessages).toBe('function');
 		});
 	});
 });
--- a/packages/web/src/hooks/__tests__/useGroupMessages.test.ts
+++ b/packages/web/src/hooks/__tests__/useGroupMessages.test.ts
@@ -1,0 +1,458 @@
+// @ts-nocheck
+/**
+ * Tests for useGroupMessages Hook
+ *
+ * Verifies LiveQuery subscription lifecycle for session group messages:
+ * - Snapshot delivery (initial message load)
+ * - Delta append (new messages via real-time push)
+ * - Stale-event guard (rapid task switching)
+ * - Append-only invariant (ignore updated/removed)
+ * - Cleanup on unmount / groupId change
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { renderHook, act } from '@testing-library/preact';
+
+// ---------------------------------------------------------------------------
+// Hoisted mocks (must not import anything)
+// ---------------------------------------------------------------------------
+
+const { mockRequest, mockOnEvent } = vi.hoisted(() => ({
+	mockRequest: vi.fn(),
+	mockOnEvent: vi.fn(),
+}));
+
+// Mock useMessageHub so we control request and onEvent directly.
+vi.mock('../useMessageHub', () => ({
+	useMessageHub: () => ({
+		request: mockRequest,
+		onEvent: mockOnEvent,
+	}),
+}));
+
+// ---------------------------------------------------------------------------
+// Imports (after mocks)
+// ---------------------------------------------------------------------------
+
+import {
+	useGroupMessages,
+	generateGroupMessagesSubId,
+	type SessionGroupMessage,
+} from '../useGroupMessages';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+/** Builds a minimal SessionGroupMessage for use in tests. */
+function makeMessage(id: number, content = `msg-${id}`): SessionGroupMessage {
+	return {
+		id,
+		groupId: 'group-1',
+		sessionId: null,
+		role: 'assistant',
+		messageType: 'text',
+		content,
+		createdAt: 1_000_000 + id,
+	};
+}
+
+/** Handler registry keyed by event name for simulating server pushes. */
+type EventHandler = (event: unknown) => void;
+let eventHandlers: Record<string, EventHandler[]> = {};
+
+function fireEvent(method: string, payload: unknown): void {
+	(eventHandlers[method] ?? []).forEach((h) => h(payload));
+}
+
+// ---------------------------------------------------------------------------
+// Setup / teardown
+// ---------------------------------------------------------------------------
+
+beforeEach(() => {
+	vi.resetAllMocks();
+	eventHandlers = {};
+
+	// Default: subscribe/unsubscribe resolve immediately.
+	mockRequest.mockResolvedValue({ ok: true });
+
+	// onEvent registers handlers and returns an unsubscribe stub.
+	mockOnEvent.mockImplementation((method: string, handler: EventHandler) => {
+		if (!eventHandlers[method]) eventHandlers[method] = [];
+		eventHandlers[method].push(handler);
+		return () => {
+			eventHandlers[method] = (eventHandlers[method] ?? []).filter((h) => h !== handler);
+		};
+	});
+});
+
+// Note: no afterEach reset here. `beforeEach` resets mocks before each test.
+// Adding vi.resetAllMocks() in afterEach would reset mocks before
+// @testing-library/preact's cleanup fires, causing useEffect cleanup to fail
+// when it calls request() (which would return undefined after reset).
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('useGroupMessages', () => {
+	describe('initial state', () => {
+		it('returns empty messages and isLoading=false when groupId is null', () => {
+			const { result } = renderHook(() => useGroupMessages(null));
+
+			expect(result.current.messages).toEqual([]);
+			expect(result.current.isLoading).toBe(false);
+		});
+
+		it('sets isLoading=true immediately when groupId is provided', () => {
+			const { result } = renderHook(() => useGroupMessages('group-1'));
+
+			expect(result.current.isLoading).toBe(true);
+			expect(result.current.messages).toEqual([]);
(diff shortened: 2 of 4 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** (P1 issue present)

---

### Summary

The hook correctly implements all four design constraints (append-only invariant, stale-event guard, cleanup on unmount/groupId change, error-safe cleanup). The 20-test suite is thorough and the code integrates cleanly with the existing codebase patterns. However, one significant behavioral gap was found: the hook does not handle WebSocket reconnection, which will cause messages to silently go stale after a disconnect/reconnect cycle in production.

---

### P1 — Silent stale state after WebSocket reconnection

**File:** `packages/web/src/hooks/useGroupMessages.ts`

The `useEffect` dependency array is `[groupId, request, onEvent]`. Both `request` and `onEvent` are stable `useCallback` references (they never change identity), so the effect will not re-run after a WebSocket reconnect.

When a reconnection occurs, a new underlying hub instance is created. The `onEvent` handlers registered in the previous connection are silently dropped (the old hub's listener registry is gone), and `liveQuery.subscribe` is never re-sent to the server. The hook's `isLoading` will remain `false` and `messages` will stop updating — no error, no reload, just frozen data.

This is a real user-visible regression path: any user who experiences a network blip while on a task view will see a frozen message list with no indication that data is no longer live.

**Suggested fix:** Add the hub's connection identity to the effect dependencies. The canonical pattern used elsewhere in this codebase is to read from `connectionState` (a Preact signal) so the effect re-runs when connectivity changes:

```ts
// Add isConnected to deps so the effect re-fires after reconnection
const { request, onEvent, isConnected } = useMessageHub();
// ...
}, [groupId, request, onEvent, isConnected]);
```

This causes the effect to re-subscribe (new subscriptionId, fresh snapshot) after every reconnect, which is the correct behavior for a LiveQuery subscription.

**Test gap:** No test covers the reconnect-then-receive scenario. A test that simulates: subscribe → receive snapshot → simulate disconnect (new `mockRequest`/`mockOnEvent` identity) → simulate reconnect → verify messages are refreshed would be needed.

---

### P2 — Module-level `_subscriptionCounter` not reset between tests

**File:** `packages/web/src/hooks/useGroupMessages.ts` (line 33), `packages/web/src/hooks/__tests__/useGroupMessages.test.ts` (beforeEach)

`_subscriptionCounter` is a mutable module-level variable. `beforeEach` calls `vi.resetAllMocks()` but does not reset this counter. As a result, counter values grow monotonically across the entire test run, including across other test files that import the module in the same Vitest worker.

Currently all tests use `expect.stringContaining('group-abc')` or extract the ID dynamically from `mockRequest.mock.calls`, so no existing test is broken. But if a future test ever matches on an exact subscription ID pattern with a specific numeric suffix, it will be order-sensitive.

**Suggested fix:** Export a `resetSubscriptionCounterForTesting` function (or use a factory pattern that avoids module-level state), and call it in `beforeEach`.

---

### P3 — `index.test.ts` not updated for new exports

**File:** `packages/web/src/hooks/__tests__/index.test.ts`

The barrel-export test file does not import or assert `useGroupMessages`, `SessionGroupMessage`, or `UseGroupMessagesResult` from `../index.ts`. Given the PR adds exports to `index.ts`, the smoke test should be updated to include them.

---

### P3 — No timeout guard for `isLoading` when snapshot never arrives

**File:** `packages/web/src/hooks/useGroupMessages.ts` (lines 76–105)

If the server acknowledges `liveQuery.subscribe` successfully (promise resolves with `{ ok: true }`) but never emits a `liveQuery.snapshot` event (server-side bug, dropped message), `isLoading` stays `true` indefinitely. This is the same pattern used in similar hooks in this codebase, so it is not a regression, but for a new hook it is worth flagging as a known limitation.

---

### Minor observations (informational, no action needed)

- The comment at line 79–81 (`// Register event listeners BEFORE sending the subscribe request so the snapshot that is delivered synchronously...`) accurately captures the race-condition rationale. Good defensive documentation.
- `Promise.resolve(request(...))` wrapper at line 119 is a reasonable safety net for test scenarios where `request` might return a non-thenable. The inline comment explains the intent clearly.
- The `@ts-nocheck` at line 1 of the test file is consistent with every other hook test in the codebase — not a new pattern introduced by this PR.

lsm (commented):

### Summary

Clean, well-structured implementation of the `useGroupMessages` LiveQuery subscription hook. The code correctly implements all four design constraints specified in the task.

### What works well

- **Stale-event guard**: The `activeSubIdRef` pattern correctly tracks the current subscription ID and discards events from prior subscriptions. This handles rapid task switching gracefully.
- **Append-only invariant**: Only `event.added` from delta events is processed; `updated` and `removed` are explicitly ignored per the domain constraint.
- **Race-condition prevention**: Event listeners are registered *before* the subscribe request is sent (lines 82-94 in `useGroupMessages.ts`), preventing a race where the snapshot could arrive before the listener is set up.
- **Error-safe cleanup**: The cleanup function wraps the unsubscribe call in `Promise.resolve().catch()` to handle edge cases where `request()` might throw synchronously or return unexpected values.
- **Stale error guard**: The `.catch()` on the subscribe request checks `activeSubIdRef.current === subscriptionId` before clearing loading state, preventing a stale error from one group from affecting the loading state of a different group.
- **Tests**: 20 tests with good coverage of snapshot handling, delta handling, stale-event guards, cleanup lifecycle, and error scenarios. All tests pass.

### Issues

**P3 (nit):**

1. **Unrelated exports in `index.ts`** (`packages/web/src/hooks/index.ts`, lines 21-22): `groupModelsByProvider` and `filterModelsForPicker` were added to the barrel export but are unrelated to this task. Not a problem, just noise in the diff.

2. **`@ts-nocheck` in test file** (`packages/web/src/hooks/__tests__/useGroupMessages.test.ts`, line 1): Blanket TypeScript suppression. This is pragmatic for test mocking but could mask real type errors if the hook interface changes. Consider removing it and adding targeted `@ts-expect-error` comments only where needed.

### Recommendation: APPROVE

The implementation is correct, complete, and well-tested. No blocking issues found.

lsm (commented):

**Recommendation: APPROVE**

### Re-review: All previous feedback addressed correctly

The follow-up commit `fix: address review feedback on useGroupMessages hook` resolves all issues from the prior round.

**P1 (reconnect handling) -- Fixed:**
- `isConnected` is now included in the `useEffect` dependency array at line 136 of `useGroupMessages.ts`, causing the effect to re-run on WebSocket reconnect.
- The early-return guard at line 80 (`if (!groupId || !isConnected)`) correctly clears state and nulls the active subscription ID when disconnected.
- Two new reconnect tests (lines 347-433) verify: (1) re-subscription with fresh snapshot after reconnect, and (2) stale events from the pre-reconnect subscription are discarded. Both tests are well-structured and correctly simulate the disconnect/reconnect cycle by toggling `mockIsConnected.value`.

**P2 (subscription counter reset) -- Fixed:**
- `resetSubscriptionCounterForTesting()` is exported (line 47) and called in `beforeEach` (line 101).
- A dedicated test at line 571 verifies the counter produces deterministic values after reset.

**P3 (index.test.ts) -- Fixed:**
- `useGroupMessages` is imported in the barrel export test (line 17) with a corresponding assertion at lines 63-66.

### No new issues found

The implementation remains clean and correct. All 33 tests pass. The reconnect fix integrates naturally with the existing `useMessageHub` pattern -- `isConnected` is already a reactive value derived from `connectionState`, making it a reliable dependency for re-subscription triggers.

lsm (commented):

---

**Recommendation: APPROVE**

This is a follow-up review verifying the fixes from the previous round. All three flagged issues have been correctly addressed.

### P1 Fix Verified: isConnected in effect dependency array

`packages/web/src/hooks/useGroupMessages.ts`, line 136

The dependency array now reads `[groupId, isConnected, request, onEvent]`. The fix is correct and complete. When the WebSocket reconnects, `isConnected` transitions from `false` to `true`, causing the effect to re-run. The effect's guard `if (!groupId || !isConnected)` ensures the subscription is not re-established during the disconnected phase, and a fresh subscription is created on reconnect.

Two reconnect tests at lines 362-406 and 408-445 of `useGroupMessages.test.ts` validate both the re-subscription behavior and the stale-event discard behavior after a reconnect cycle. The test approach (setting `mockIsConnected.value` inside the render function with a getter-based mock) correctly simulates the connection state change and triggers the effect re-execution.

### P2 Fix Verified: resetSubscriptionCounterForTesting

`packages/web/src/hooks/useGroupMessages.ts`, lines 47-49; test `beforeEach` at line 86

`resetSubscriptionCounterForTesting()` is exported from the module and called in `beforeEach`. The counter reset test at line 570-574 explicitly verifies that after reset, the first call to `generateGroupMessagesSubId` produces `group-messages-g-1`, confirming deterministic counter values across tests.

The function is not re-exported from the barrel (`packages/web/src/hooks/index.ts`), which is the right call — it is an internal testing utility. Knip ignores `packages/web/src/**` so no dead-export warning will fire.

### P3 Fix Verified: useGroupMessages in barrel export test

`packages/web/src/hooks/__tests__/index.test.ts`, lines 17 and 62-65

`useGroupMessages` is imported and verified in the barrel export test. The barrel itself (`index.ts`, lines 50-53) correctly re-exports `useGroupMessages` along with its types.

### Remaining observations (no blocking issues)

**P3 — @ts-nocheck in test files**: Both `useGroupMessages.test.ts` (line 1) and `index.test.ts` (line 1) use `// @ts-nocheck`. This suppresses TypeScript type checking in the test files entirely, which means type errors in test code (e.g., wrong event payload shapes, incorrect mock return types) would go undetected. This is a minor quality concern; removing `@ts-nocheck` and fixing any resulting type errors would make the tests more robust. It does not affect production code correctness.

**P3 — generateGroupMessagesSubId exposed from module**: The function is publicly exported from `useGroupMessages.ts` but serves only as a test helper (it is not re-exported from the barrel). This is fine given the knip ignore rule, but a comment clarifying it is only exported for testing would aid future maintainers. The existing JSDoc does say "Exported for testing" so this is already addressed.

### Summary

All three previously flagged issues (P1 stale data on reconnect, P2 non-deterministic counter, P3 barrel export test) are correctly and completely resolved. The implementation is sound: append-only invariant is enforced, stale-event guard works across task-switching and reconnect cycles, cleanup is comprehensive, and error handling for failed subscribe calls correctly guards against stale error clearing. The test suite covers all the important behavioral cases.

PR #694, opened 2026-03-22
feat: add force-stop session group RPC + auto-clean stale groups in tick

lsm:
## Part 1: session_group.stop RPC

Adds `session_group.stop` RPC handler that force-stops a session group by ID:
- Kills worker and leader agent sessions via stopSession()
- Marks the group as failed (terminal) and deletes the record from DB
- Frees the concurrency slot so the runtime can pick up new tasks
- Task status is deliberately unchanged; callers can cancel separately
- Validates that the group belongs to the requesting room

Also adds `RoomRuntime.forceStopSessionGroup()` as the underlying runtime method.

## Part 2: Auto-clean stale groups in tick

Adds `cleanStaleGroups()` to `RoomRuntime.executeTick()` that runs on every tick:
- Detects active groups (completedAt IS NULL) whose tasks have reached a
  terminal state (completed, cancelled, archived)
- Terminates their sessions and marks the groups as terminal
- Frees concurrency slots that were blocked by orphaned groups
- Root-cause fix for the slot exhaustion problem

## Tests

Adds 10 unit tests covering:
- cleanStaleGroups: ignores in_progress tasks, cleans completed/cancelled/archived tasks, frees slots
- forceStopSessionGroup: validates room ownership, deletes group from DB, triggers tick

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -245,6 +245,7 @@ jobs:
             mock_sdk: true
           - module: rpc-3
             test_path: >-
+              tests/online/rpc/rpc-live-query.test.ts
               tests/online/rpc/rpc-session-handlers-extended.test.ts
               tests/online/rpc/rpc-session-workflow.test.ts
               tests/online/rpc/rpc-settings-handlers.test.ts
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -144,6 +144,7 @@ jobs:
           - rpc-1
           - rpc-2
           - rpc-3
+          - rpc-4
           - sdk
           - websocket
         include:
@@ -223,32 +224,39 @@ jobs:
           # - module: room-reviewer
           #   test_path: tests/online/room/room-reviewer-flow.test.ts
           #   timeout: 10
+          # rpc-1: heaviest files (task-lifecycle ~647 + config-handlers ~573 lines ≈ 1 220)
           - module: rpc-1
             test_path: >-
-              tests/online/rpc/rpc-agent-handlers.test.ts
               tests/online/rpc/rpc-config-handlers.test.ts
-              tests/online/rpc/rpc-draft-handlers.test.ts
-              tests/online/rpc/rpc-file-handlers.test.ts
-              tests/online/rpc/rpc-interrupt-handlers.test.ts
-              tests/online/rpc/rpc-mcp-toggle.test.ts
-              tests/online/rpc/rpc-task-draft-handlers.test.ts
               tests/online/rpc/rpc-task-lifecycle.test.ts
             mock_sdk: true
+          # rpc-2: settings + remove-output + message + agent (≈ 1 241 lines)
           - module: rpc-2
             test_path: >-
+              tests/online/rpc/rpc-agent-handlers.test.ts
               tests/online/rpc/rpc-message-handlers.test.ts
-              tests/online/rpc/rpc-model-handlers.test.ts
-              tests/online/rpc/rpc-model-switching.test.ts
               tests/online/rpc/rpc-remove-output.test.ts
-              tests/online/rpc/rpc-rewind-handlers.test.ts
-              tests/online/rpc/rpc-session-filtering.test.ts
+              tests/online/rpc/rpc-settings-handlers.test.ts
             mock_sdk: true
+          # rpc-3: session-extended + live-query + mcp + task-draft + rewind (≈ 1 356 lines)
           - module: rpc-3
             test_path: >-
               tests/online/rpc/rpc-live-query.test.ts
+              tests/online/rpc/rpc-mcp-toggle.test.ts
+              tests/online/rpc/rpc-rewind-handlers.test.ts
               tests/online/rpc/rpc-session-handlers-extended.test.ts
+              tests/online/rpc/rpc-task-draft-handlers.test.ts
+            mock_sdk: true
+          # rpc-4: remaining smaller files (≈ 1 382 lines)
+          - module: rpc-4
+            test_path: >-
+              tests/online/rpc/rpc-draft-handlers.test.ts
+              tests/online/rpc/rpc-file-handlers.test.ts
+              tests/online/rpc/rpc-interrupt-handlers.test.ts
+              tests/online/rpc/rpc-model-handlers.test.ts
+              tests/online/rpc/rpc-model-switching.test.ts
+              tests/online/rpc/rpc-session-filtering.test.ts
               tests/online/rpc/rpc-session-workflow.test.ts
-              tests/online/rpc/rpc-settings-handlers.test.ts
               tests/online/rpc/rpc-state-sync.test.ts
               tests/online/rpc/session-handlers.test.ts
             mock_sdk: true
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -117,6 +117,24 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		maxConcurrent,
 		staleThresholdMs: 5 * 60 * 1000,
 	});
+	// --- setInterval inventory (out-of-scope for job-queue migration) ---
+	// The following subsystems intentionally retain their own setInterval timers.
+	// They were audited as part of the background-task migration (milestone 6) and
+	// determined to be out-of-scope because they are not "business tasks" that
+	// belong in the job queue:
+	//
+	//   • JobQueueProcessor.pollTimer (job-queue-processor.ts)
+	//       IS the job-queue infrastructure itself — migrating it is circular.
+	//   • JobQueueProcessor drain-check in stop() (job-queue-processor.ts)
+	//       Short-lived shutdown poll (50 ms); not a recurring business task.
+	//   • WebSocketServerTransport.staleCheckTimer (websocket-server-transport.ts)
+	//       Transport-layer health check; no business logic, not schedulable.
+	//   • SpaceRuntime.tickTimer (space/runtime/space-runtime.ts)
+	//       Drives the SpaceRuntime workflow engine; migrate in a dedicated follow-up.
+	//   • TaskAgentManager concurrent-spawn poll (space/runtime/task-agent-manager.ts)
+	//       Ephemeral, within a single async call; cleaned up before the call returns.
+	//   • app.ts graceful-shutdown readiness check (this file, waitForPendingCalls)
+	//       One-shot shutdown polling with hard timeout; not a recurring task.
 	jobProcessor.setChangeNotifier((table) => {
 		reactiveDb.notifyChange(table);
 	});
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1824,6 +1824,57 @@ export class RoomRuntime {
 	}
 
 	/**
+	 * Force-stop a session group by ID.
+	 *
+	 * Kills worker and leader sessions, marks the group as failed, and deletes the
+	 * group record from the DB. Used for manual cleanup of stale or stuck groups
+	 * via the `session_group.stop` RPC.
+	 *
+	 * Task status is NOT changed — the group is removed while leaving the task
+	 * in its current state. Call task.cancel separately if needed.
+	 *
+	 * Returns { success: false } if the group doesn't exist or belongs to a
+	 * different room (validated by checking the task via this room's TaskManager).
+	 */
+	async forceStopSessionGroup(groupId: string): Promise<{ success: boolean; error?: string }> {
+		const group = this.groupRepo.getGroup(groupId);
+		if (!group) {
+			return { success: false, error: `Session group ${groupId} not found` };
+		}
+
+		// Validate the group belongs to this room by fetching the task via this
+		// room's TaskManager (which is scoped to this.roomId).
+		const task = await this.taskManager.getTask(group.taskId);
+		if (!task) {
+			// Task not in this room or doesn't exist. Refuse to act on foreign groups.
+			return {
+				success: false,
+				error: `Group ${groupId} belongs to a different room or its task no longer exists`,
+			};
+		}
+
+		// Stop the actual agent processes first (best-effort).
+		await this.terminateGroupSessions(group);
+
+		// Mark group as terminal in the DB (unobserves sessions too).
+		if (group.completedAt === null) {
+			await this.taskGroupManager.terminateGroup(groupId);
+		}
+
+		// Clean up message mirroring subscriptions.
+		this.cleanupMirroring(groupId, 'Force-stopped by user.');
+
+		// Delete the group record from the DB (freeing the concurrency slot).
+		this.groupRepo.deleteGroup(groupId);
+
+		await this.emitTaskUpdateById(group.taskId);
+		this.scheduleTick();
+
+		log.info(`[forceStopSessionGroup] Group ${groupId} for task ${group.taskId} force-stopped`);
+		return { success: true };
+	}
+
+	/**
 	 * Interrupt the current agent session(s) for a task without changing task status.
 	 *
 	 * Unlike stopTaskSession() / cancelTask(), this:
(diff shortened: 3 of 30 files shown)


lsm (commented):

### Summary

Clean, well-scoped implementation that addresses the stale group / slot exhaustion problem. The code follows existing patterns in `room-runtime.ts` faithfully, tests are thorough, and the RPC handler is properly validated. I found one P2 issue and two P3 nits.

### Recommendation: APPROVE

---

### P2: `cleanStaleGroups` loop lacks per-group try/catch — one failure aborts cleanup for remaining groups

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 2590-2610

The `for` loop in `cleanStaleGroups` calls `terminateGroupSessions` (which is safe — has internal try/catch) and then `taskGroupManager.terminateGroup`. If `terminateGroup` throws for one group (e.g., a concurrent version conflict in `failGroup`), the exception propagates out of the loop and `cleanStaleGroups`, potentially aborting the entire `executeTick` call. Remaining stale groups would not be cleaned until the next tick, and if the same group keeps failing, it blocks cleanup permanently.

Other safety-net methods in the same class (e.g., `recoverZombieGroups`) handle this by wrapping per-item logic in try/catch. Consider wrapping the loop body:

```ts
for (const group of activeGroups) {
  try {
    // ... existing logic
  } catch (err) {
    log.error(`[cleanStaleGroups] Failed to clean group ${group.id}:`, err);
  }
}
```

This is a meaningful resilience improvement for a method explicitly described as a "safety net."

---

### P3: `forceStopSessionGroup` writes a group event in `cleanupMirroring` then immediately deletes the group

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 1864-1868

`cleanupMirroring` calls `appendGroupEvent(groupId, 'status', ...)` which inserts a row into `task_group_events`. Then `deleteGroup` deletes the group with CASCADE, removing that event immediately. This is harmless but wasteful. Not worth changing unless the mirroring API is refactored; just noting it.

### P3: `STALE_TASK_STATUSES` is re-created on every tick

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 2588

The `Set` could be a module-level constant since the values never change. Micro-optimization, not worth a change on its own.

---

### What looks good

- **Ownership validation** in `forceStopSessionGroup` via the room-scoped `TaskManager.getTask()` is a clean pattern that prevents cross-room group manipulation without needing an explicit `roomId` column on `session_groups`.
- **Behavioral asymmetry** between `forceStopSessionGroup` (deletes group record) and `cleanStaleGroups` (marks terminal but preserves record) is intentional and well-documented — manual force-stop wants a hard reset, while auto-cleanup just needs to free the slot.
- **Test coverage** is solid: 5 tests for auto-cleanup covering each terminal status + slot-freeing, and 5 tests for force-stop covering error paths, deletion, completed-task edge case, and downstream tick scheduling.
- **Ordering** of `cleanStaleGroups` before zombie recovery in `executeTick` is correct — stale groups should be cleaned first so they don't appear as zombies.

lsm (commented):

The implementation is well-structured and covers the core requirements. The room-ownership validation in `forceStopSessionGroup` and the tick-ordering decision (stale cleanup before zombie recovery) are both correct. Test coverage is reasonable for the happy paths. However, there are two P1 issues and several P2 issues that should be addressed.

---

## P1 Issues

### 1. `cleanStaleGroups` does not emit task update events to the UI

`forceStopSessionGroup` correctly calls `emitTaskUpdateById` after cleaning up a group, but `cleanStaleGroups` does not. When a stale group is auto-cleaned on tick, the task's `activeSession` indicator (and any other group-derived fields the frontend renders) will remain stale in the UI until the next full state refresh. Compare:

- `forceStopSessionGroup` (line 1870): calls `this.emitTaskUpdateById(group.taskId)`
- `cleanStaleGroups` (lines 2600–2609): terminates + cleans mirroring, but **no `emitTaskUpdateById`**

Additionally, neither `forceStopSessionGroup` nor `cleanStaleGroups` calls `emitGoalProgressForTask`, while the analogous `cancelTask()` does. Goal progress bars will lag until the next external event triggers a recalculation.

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 2600–2610

### 2. `cleanStaleGroups` is blind to groups whose tasks were hard-deleted from the DB

`getActiveGroups` uses an `INNER JOIN` on `tasks` (`packages/daemon/src/lib/room/state/session-group-repository.ts`, line 303):

```sql
FROM session_groups sg
JOIN tasks t ON sg.ref_id = t.id   -- INNER JOIN
WHERE t.room_id = ? AND sg.completed_at IS NULL
```

Any group whose `ref_id` does not match a row in `tasks` is **invisible** to this query and will never be returned. This means:
- The `isStale = !task` branch on line 2592 is only reachable as a narrow TOCTOU race (task deleted between `getActiveGroups` and `getTask`).
- Groups orphaned because their task was deleted outright from the DB will never be auto-cleaned by this mechanism.

The PR description says the fix covers "daemon crash or an external status change" — but hard-deleted tasks remain unaddressed. This is a gap in coverage even if the use case is rare.

---

## P2 Issues

### 3. `terminateGroup()` return value silently discarded in `forceStopSessionGroup`

`forceStopSessionGroup` ignores the return value of `taskGroupManager.terminateGroup()`:

```ts
// Line 1861 — null return (CAS failure) is not checked
await this.taskGroupManager.terminateGroup(groupId);
```

By contrast, `cancelTask()` (line 1806) checks the return and surfaces `{ success: false }` if `terminateGroup` fails. For a force-stop the best-effort approach may be intentional, but: (a) this silent deviation from the established pattern is not documented, and (b) if `terminateGroup` fails silently, the group may not have `completedAt` set before `deleteGroup` is called, leaving session observers dangling (`unobserve` is called inside `terminateGroup`, not separately).

### 4. `STALE_TASK_STATUSES` is reconstructed on every tick

```ts
// packages/daemon/src/lib/room/runtime/room-runtime.ts, line 2588
const STALE_TASK_STATUSES = new Set<string>(['completed', 'cancelled', 'archived']);
```

This `new Set` is allocated on every `cleanStaleGroups` call, which runs on every tick. Promote it to a module-level or class-level constant.

### 5. No RPC type definition in shared package

The `session_group.stop` RPC is registered in `task-handlers.ts` with an inline cast (`data as { roomId: string; groupId: string }`) but no corresponding type declaration in `packages/shared/src`. Other RPC handlers follow this pattern too, but adding a typed entry to the shared RPC map (if one exists) or at minimum a Zod/type guard validation would prevent silent type mismatches at call sites. Currently the handler will silently accept `{ roomId: undefined, groupId: undefined }` because the falsy checks only catch empty strings, not undefined — though undefined would stringify to `'undefined'` and fail the runtime lookup cleanly.

---

## P3 Issues

### 6. Fragile test assertion at line 69 of `stale-group-cleanup.test.ts`

```ts
if (group) {
    expect(group.completedAt).not.toBeNull();
}
```

If `getGroup` returns `null` (because the group was deleted), the inner `expect` is silently skipped and the test passes unconditionally. Change to `expect(group).not.toBeNull(); expect(group!.completedAt).not.toBeNull();` to make the assertion mandatory.

### 7. No frontend call site for `session_group.stop`

`session_group.stop` is registered on the daemon side but there is no corresponding call in `packages/web/src`. This is expected if UI wiring is a follow-up, but it should be confirmed — the RPC is not useful without a caller.

---

## Correctness Notes (no action needed)

- The tick ordering (stale cleanup → zombie recovery → stuck-worker recovery → slot allocation) is correct. Stale cleanup running first ensures zombie detection doesn't spend effort on groups that should be torn down anyway.
- Room-ownership validation in `forceStopSessionGroup` via `taskManager.getTask(group.taskId)` is correct: `TaskManager` is scoped to `this.roomId`, so a task from another room will return null.
- `getActiveGroups` uses an INNER JOIN with tasks (not LEFT JOIN), which means the concurrency slot counter at tick line 2647 is naturally unaffected by groups with deleted tasks. The slot-freeing path via `terminateGroup` setting `completedAt` is correct.
- The test helper that inserts a "foreign-group" with a non-existent `task-id` correctly exercises the room-validation path in `forceStopSessionGroup`, because `getGroup` (used in `forceStopSessionGroup`) uses `LEFT JOIN` and will find the group, but `taskManager.getTask` will return null.

lsm (commented):

All issues from the previous review round have been resolved:

1. **STALE_TASK_STATUSES promoted to module-level constant** (line 95 of `room-runtime.ts`) — no longer recreated on every tick call. Good.

2. **Per-group try/catch in `cleanStaleGroups()`** (lines 2609-2635 of `room-runtime.ts`) — each group is now wrapped in its own try/catch so one failure does not abort cleanup of remaining groups. The error is logged and the loop continues. This was the P2 issue from the prior review and it is properly fixed.

3. **UI sync events added** — `emitTaskUpdateById` and `emitGoalProgressForTask` are now called after each stale group cleanup (lines 2631-2632), ensuring the frontend reflects the cleaned-up state.

4. **`terminateGroup()` null return logged** — `forceStopSessionGroup` now logs a warning when `terminateGroup()` returns null (lines 2873-2876), indicating a possible concurrent modification.

5. **Fragile test assertion fixed** — The conditional `if (group)` guard in the stale cleanup test has been replaced with unconditional `expect(group).not.toBeNull()` followed by `expect(group!.completedAt).not.toBeNull()`, which will fail clearly if the assumption is violated.

6. **INNER JOIN blindspot documented** — Comment at lines 2604-2606 explains that groups whose tasks are hard-deleted won't appear in `getActiveGroups()` and points to the zombie recovery path as the handler for that case.

### Implementation quality

- `cleanStaleGroups()` runs at the start of every tick as a safety net, correctly checking task status against terminal states.
- `forceStopSessionGroup()` validates room ownership via TaskManager scoping, handles the full cleanup lifecycle (terminate sessions, mark group terminal, clean mirroring, delete record), and schedules a follow-up tick.
- Test coverage is thorough with 9 test cases covering: active groups unaffected, auto-clean on completed/cancelled/archived tasks, slot freeing for new tasks, non-existent/foreign group rejection, force-stop with DB deletion, and tick scheduling after force-stop.

**Recommendation: APPROVE** — No remaining issues.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous round have been addressed correctly. The implementation is solid. One minor asymmetry noted below, but it does not block merging.

---

**P1: cleanStaleGroups() didn't emit task update events to UI** — RESOLVED

Both `emitTaskUpdateById` and `emitGoalProgressForTask` are now called inside the per-group loop in `cleanStaleGroups()` (room-runtime.ts lines 2631–2632), and both calls are inside the per-group `try/catch` so a failure in one group's emit won't silently break others.

**P1: getActiveGroups INNER JOIN blind spot** — RESOLVED (documented)

A clear `// NOTE:` comment now documents the INNER JOIN limitation and explicitly points to the zombie recovery path as the safety net for hard-deleted task groups (room-runtime.ts lines 2604–2606). This is the right choice given the existing zombie recovery path already covers that case.

**P2: Missing per-group try/catch in cleanStaleGroups** — RESOLVED

Each group is now wrapped in a `try/catch` (lines 2609–2635). The catch block logs at `error` level and skips the group, allowing other groups to continue processing.

**P2: terminateGroup() return value silently discarded in forceStopSessionGroup** — RESOLVED

The return value is now captured in `terminated`; when null, a `log.warn` is emitted explaining the possible concurrent modification before the code continues to `deleteGroup()` (lines 1868–1877).

Also noted: `STALE_TASK_STATUSES` is promoted to module-level as described (line 93 in the diff).

---

### New Observations

**Minor asymmetry: forceStopSessionGroup omits emitGoalProgressForTask**

`cleanStaleGroups` calls both `emitTaskUpdateById` and `emitGoalProgressForTask` after cleaning a group. `forceStopSessionGroup` calls only `emitTaskUpdateById` (line 1886) and omits `emitGoalProgressForTask`. Since `forceStopSessionGroup` is a manual admin operation that explicitly leaves task status unchanged, the omission is likely intentional — the goal progress bar reflects task status rather than group state, so there is nothing to update. However, the asymmetry is undocumented and could confuse a future maintainer. A brief comment explaining the intentional omission would help.

This is a P3 nit and does not block the PR.

**session_group.stop RPC has no unit test at the RPC handler layer**

The runtime method `forceStopSessionGroup` is well-tested in `stale-group-cleanup.test.ts`. The RPC glue in `task-handlers.ts` (the `session_group.stop` handler) is not covered by a unit test. An integration test or unit test verifying that the handler correctly propagates `roomId`/`groupId`, invokes the runtime, and re-throws on failure would add confidence without much effort. This is P2 in isolation, but the runtime tests cover the real logic, so the gap is relatively low-risk.

**task.fail emitTaskUpdate removal is well-justified**

Removing `emitTaskUpdate` from the `task.fail` handler and updating the test assertion is a clean follow-up cleanup. The test comment ("LiveQuery deltas replace this channel") makes the intent clear.

---

### Summary

All four previously flagged issues are resolved. The implementation is clean, well-commented, and the 10-unit tests cover the main branches thoroughly. The minor asymmetry in `forceStopSessionGroup` (no `emitGoalProgressForTask`) and the missing RPC-layer test are worth noting but do not block the merge.

lsm (commented):

**Recommendation: APPROVE** — Both issues from the previous round are now fully resolved.

---

**P2: Missing RPC-layer unit test for the session_group.stop handler** — RESOLVED

Seven new unit tests have been added in `packages/daemon/tests/unit/rpc-handlers/task-handlers.test.ts` (lines 1457–1544) under the `session_group.stop RPC Handler` describe block. The tests cover exactly the scenarios requested:
- Missing `roomId` validation (throws 'Room ID is required')
- Missing `groupId` validation (throws 'Group ID is required')
- Missing `runtimeService` (throws 'Runtime service is required')
- Null runtime from `getRuntime()` (throws 'No runtime found for room')
- Happy path: returns `{ success: true }` and verifies `forceStopSessionGroup` is called with the correct `groupId`
- Error propagation when `success: false` with an `error` field
- Generic fallback error when `success: false` with no `error` field

The test structure uses the same `makeNullRuntimeService()` helper and `setup()`/`getHandler()` pattern established by adjacent handler tests in the file. Coverage is thorough and follows project conventions.

**P3: Undocumented intentional asymmetry in forceStopSessionGroup omitting emitGoalProgressForTask** — RESOLVED

A clear explanatory comment has been added at lines 1886–1889 of `room-runtime.ts`:

```
// Emit task update so the frontend reflects the removed group.
// emitGoalProgressForTask is intentionally omitted: goal progress is derived
// from task status, which forceStopSessionGroup deliberately leaves unchanged.
// There is nothing for the goal progress bar to update.
```

This is exactly the right level of documentation — concise, explains the reasoning (goal progress is task-status-derived, not group-derived), and prevents future maintainers from treating the omission as a bug.

---

### No new concerns

The implementation remains clean. No regressions introduced by the new test code. All previously noted items (per-group try/catch in `cleanStaleGroups`, `emitTaskUpdateById`/`emitGoalProgressForTask` in stale cleanup, `terminateGroup` return value logging, INNER JOIN limitation documented, `STALE_TASK_STATUSES` promoted to module-level constant) remain correctly addressed from prior rounds.

lsm (commented):

### Re-review: Addressing P2/P3 feedback from sibling reviewer

Both items raised in the previous round have been properly addressed.

**P2 (resolved): RPC-layer unit tests for `session_group.stop`**

The new test suite at `packages/daemon/tests/unit/rpc-handlers/task-handlers.test.ts` (lines 1457-1544) covers all the important paths:

1. **Parameter validation** -- missing `roomId`, missing `groupId`
2. **Runtime service validation** -- `runtimeService` not provided (undefined), runtime not found for room (null return)
3. **Happy path** -- verifies `{ success: true }` return and correct `groupId` passed to `forceStopSessionGroup`
4. **Error propagation** -- `success: false` with explicit error message, `success: false` with no error field (fallback message)

The `makeForceStopRuntimeService` helper is clean and well-scoped. The tests correctly validate the RPC handler layer in isolation from the `RoomRuntime.forceStopSessionGroup` implementation, which has its own integration-level tests in `packages/daemon/tests/unit/room/stale-group-cleanup.test.ts`.

One minor observation: the test does not verify that `emitRoomOverview` is called on success (line 920 of `task-handlers.ts`). This is acceptable since `emitRoomOverview` is a fire-and-forget side effect and the handler's core contract (parameter validation, delegation to runtime, error mapping) is well covered.

**P3 (resolved): Documentation of intentional `emitGoalProgressForTask` omission**

Lines 1887-1889 of `packages/daemon/src/lib/room/runtime/room-runtime.ts` now contain a clear comment explaining the rationale:

```
// emitGoalProgressForTask is intentionally omitted: goal progress is derived
// from task status, which forceStopSessionGroup deliberately leaves unchanged.
// There is nothing for the goal progress bar to update.
```

This is well-written and addresses the asymmetry concern directly.

### Overall assessment

No new issues found. The existing integration tests for `cleanStaleGroups` and `forceStopSessionGroup` in `stale-group-cleanup.test.ts` remain comprehensive (terminal state cleanup for completed/cancelled/archived, slot freeing, foreign group rejection, tick scheduling). The RPC-layer tests complement these nicely by testing the handler's own validation and error mapping logic.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** (own-PR posting as COMMENT)

---

### Review Scope

Two test-only fixes in `packages/daemon/tests/unit/space/cross-agent-messaging.test.ts` (commit `dff2ffa`). The core implementation was already approved.

---

### Fix 1: Partial delivery — `result.success` changed to `'partial'`, `result.partialFailures` renamed to `result.failed`

**Verdict: Correct alignment with production.**

Production code in `packages/daemon/src/lib/space/tools/step-agent-tools.ts` line 320:

```ts
success: delivered.length > 0 ? 'partial' : false,
delivered,
failed,   // ← key is 'failed', never 'partialFailures'
```

The old test asserted `result.success === true` and read from `result.partialFailures` — neither field exists in production for this path. The new assertions (`'partial'` and `result.failed`) match exactly.

---

### Fix 2: All-failed injection — `result.error` replaced with `result.message`

**Verdict: Correct alignment with production.**

When all injections fail and `delivered.length === 0`, production returns (lines 317–328):

```ts
{
  success: false,
  delivered,   // empty array
  failed,      // injected error entries
  message: `Message delivery failed for all ${failed.length} target(s).`
}
```

There is no `error` field on this return path. Searching the production source confirms `'failed to deliver'` never appears in `step-agent-tools.ts`. The new assertion (`result.message` containing `'failed'`) matches `"Message delivery failed for all 1 target(s)."` exactly.

---

### No production regression

Both changes fix the tests to match existing production behavior — not the other way around. The production implementation in `step-agent-tools.ts` is untouched by this commit. The test now accurately documents the contract of the all-failed and partial-delivery code paths.

PR #696, opened 2026-03-22
fix: address root causes of duplicate and stale session groups

lsm:
Comprehensively fixes the zombie/stale session group problem:

1. Add `getActiveGroupsForTask(taskId)` to SessionGroupRepository — returns
   ALL active groups for a task (not just the most recent), enabling correct
   deduplication checks.

2. Fix `spawnGroupForTask` deduplication — now uses `getActiveGroupsForTask`
   instead of `getGroupByTaskId`, catching stale older groups that would
   otherwise slip through and allow a duplicate spawn.

3. Add `cleanupStaleGroupsForTask(taskId, keepGroupId)` to the repository —
   marks all other active groups for a task as completed. Called from
   `complete()` and `fail()` in TaskGroupManager as a safety net.

4. Add `cleanupZombieGroupsForRoom(roomId)` to the repository — synchronous
   DB-level cleanup that marks active groups as completed when their task is
   in a terminal state. Called from `RoomRuntime.stop()`.

5. Migration 42 — one-time cleanup of existing zombie/duplicate groups, plus
   `CREATE UNIQUE INDEX ... ON session_groups(ref_id) WHERE completed_at IS NULL`
   to enforce DB-level uniqueness: at most one active group per task.

6. Update `createIndexes()` in schema/index.ts to include the partial unique
   index for fresh DB installations.

Tests: add unit tests for all three new repository methods and a migration-42
test covering fresh DB index creation, unique constraint enforcement, zombie
cleanup, and duplicate deduplication.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -445,6 +445,15 @@ export class RoomRuntime {
 		}
 		this.mirroringCleanups.clear();
 		this.observer.dispose();
+
+		// Safety net: mark zombie groups (active groups for terminal tasks) as completed.
+		// Runs synchronously at stop time — no session recovery, just DB consistency.
+		const zombiesCleaned = this.groupRepo.cleanupZombieGroupsForRoom(this.roomId);
+		if (zombiesCleaned > 0) {
+			log.warn(
+				`[stop] Room ${this.roomId}: cleaned up ${zombiesCleaned} zombie group(s) on runtime stop`
+			);
+		}
 	}
 
 	getState(): RuntimeState {
@@ -3147,10 +3156,12 @@ export class RoomRuntime {
 	private async spawnGroupForTask(task: NeoTask): Promise<void> {
 		// Defense-in-depth: verify no active group exists for this task right before spawning.
 		// Catches races that slip past the executeTick() filter (e.g., concurrent ticks).
-		const existingGroup = this.groupRepo.getGroupByTaskId(task.id);
-		if (existingGroup && existingGroup.completedAt === null) {
+		// Check ALL active groups, not just the most recent — a stale older group with
+		// completedAt === null would be missed by getGroupByTaskId() which returns only the latest.
+		const allActiveGroups = this.groupRepo.getActiveGroupsForTask(task.id);
+		if (allActiveGroups.length > 0) {
 			log.warn(
-				`[spawnGroupForTask] Task ${task.id} ("${task.title}") already has active group ${existingGroup.id} — skipping duplicate spawn`
+				`[spawnGroupForTask] Task ${task.id} ("${task.title}") already has ${allActiveGroups.length} active group(s) (${allActiveGroups.map((g) => g.id).join(', ')}) — skipping duplicate spawn`
 			);
 			return;
 		}
--- a/packages/daemon/src/lib/room/runtime/task-group-manager.ts
+++ b/packages/daemon/src/lib/room/runtime/task-group-manager.ts
@@ -562,17 +562,19 @@ export class TaskGroupManager {
 		const group = this.groupRepo.getGroup(groupId);
 		if (!group) return null;
 
-		// Complete the group
-		const updated = this.groupRepo.completeGroup(groupId, group.version);
-		if (!updated) return null;
-
-		// Safety net: clean up any other stale active groups for this task
+		// Safety net: clean up any other stale active groups for this task BEFORE completing
+		// the canonical group. This ensures we aren't racing against the unique index, and
+		// any stale duplicates are resolved before the primary group transitions.
 		const staleCount = this.groupRepo.cleanupStaleGroupsForTask(group.taskId, groupId);
 		if (staleCount > 0) {
 			log.warn(
-				`[complete] Task ${group.taskId}: cleaned up ${staleCount} stale active group(s) on completion`
+				`[complete] Task ${group.taskId}: cleaned up ${staleCount} stale active group(s) before completion`
 			);
 		}
+
+		// Complete the group
+		const updated = this.groupRepo.completeGroup(groupId, group.version);
+		if (!updated) return null;
 
 		// Complete the task
 		await this.taskManager.completeTask(group.taskId, summary);
@@ -596,17 +598,19 @@ export class TaskGroupManager {
 		const group = this.groupRepo.getGroup(groupId);
 		if (!group) return null;
 
-		// Fail the group
-		const updated = this.groupRepo.failGroup(groupId, group.version);
-		if (!updated) return null;
-
-		// Safety net: clean up any other stale active groups for this task
+		// Safety net: clean up any other stale active groups for this task BEFORE failing
+		// the canonical group. This ensures stale duplicates are resolved before the
+		// primary group transitions.
 		const staleCount = this.groupRepo.cleanupStaleGroupsForTask(group.taskId, groupId);
 		if (staleCount > 0) {
 			log.warn(
-				`[fail] Task ${group.taskId}: cleaned up ${staleCount} stale active group(s) on failure`
+				`[fail] Task ${group.taskId}: cleaned up ${staleCount} stale active group(s) before failure`
 			);
 		}
+
+		// Fail the group
+		const updated = this.groupRepo.failGroup(groupId, group.version);
+		if (!updated) return null;
 
 		// Fail the task
 		await this.taskManager.failTask(group.taskId, reason);
--- a/packages/daemon/src/lib/room/runtime/task-group-manager.ts
+++ b/packages/daemon/src/lib/room/runtime/task-group-manager.ts
@@ -566,6 +566,14 @@ export class TaskGroupManager {
 		const updated = this.groupRepo.completeGroup(groupId, group.version);
 		if (!updated) return null;
 
+		// Safety net: clean up any other stale active groups for this task
+		const staleCount = this.groupRepo.cleanupStaleGroupsForTask(group.taskId, groupId);
+		if (staleCount > 0) {
+			log.warn(
+				`[complete] Task ${group.taskId}: cleaned up ${staleCount} stale active group(s) on completion`
+			);
+		}
+
 		// Complete the task
 		await this.taskManager.completeTask(group.taskId, summary);
 
@@ -591,6 +599,14 @@ export class TaskGroupManager {
 		// Fail the group
 		const updated = this.groupRepo.failGroup(groupId, group.version);
 		if (!updated) return null;
+
+		// Safety net: clean up any other stale active groups for this task
+		const staleCount = this.groupRepo.cleanupStaleGroupsForTask(group.taskId, groupId);
+		if (staleCount > 0) {
+			log.warn(
+				`[fail] Task ${group.taskId}: cleaned up ${staleCount} stale active group(s) on failure`
+			);
+		}
 
 		// Fail the task
 		await this.taskManager.failTask(group.taskId, reason);
--- a/packages/daemon/src/lib/room/state/session-group-repository.ts
+++ b/packages/daemon/src/lib/room/state/session-group-repository.ts
@@ -291,6 +291,30 @@ export class SessionGroupRepository {
 		return this.rowToGroup(row);
 	}
 
+	/**
+	 * Returns ALL active (completedAt IS NULL) groups for a specific task.
+	 * Used for defense-in-depth deduplication in spawnGroupForTask — checks every
+	 * active group, not just the most recent, to catch stale zombie groups that
+	 * would otherwise allow a duplicate spawn.
+	 */
+	getActiveGroupsForTask(taskId: string): SessionGroup[] {
+		const rows = this.db
+			.prepare(
+				`SELECT
+					sg.id, sg.group_type, sg.ref_id, sg.version, sg.metadata,
+					sg.created_at, sg.completed_at,
+					worker.session_id AS worker_session_id,
+					leader.session_id AS leader_session_id
+				FROM session_groups sg
+				LEFT JOIN session_group_members worker ON worker.group_id = sg.id AND worker.role = 'worker'
+				LEFT JOIN session_group_members leader ON leader.group_id = sg.id AND leader.role = 'leader'
+				WHERE sg.ref_id = ? AND sg.group_type IN ('task', 'task_pair') AND sg.completed_at IS NULL
+				ORDER BY sg.created_at DESC`
+			)
+			.all(taskId) as Record<string, unknown>[];
+		return rows.map((r) => this.rowToGroup(r));
+	}
+
 	getActiveGroups(roomId: string): SessionGroup[] {
 		const rows = this.db
 			.prepare(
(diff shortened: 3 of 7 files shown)


lsm (commented):

Overall this is a solid fix that addresses duplicate and stale session groups through multiple defense layers: spawn-time dedup check, completion/failure cleanup, runtime stop cleanup, migration data fix, and a DB-level unique constraint. The approach is well-structured and the tests are thorough. However, there are two issues that need attention.

---

### P1: `needs_attention` status missing from zombie cleanup (both runtime and migration)

**Files:**
- `packages/daemon/src/lib/room/state/session-group-repository.ts` (line 380)
- `packages/daemon/src/storage/schema/migrations.ts` (line 2493)

Both `cleanupZombieGroupsForRoom()` and migration 42 step 1 check for terminal task statuses using `('completed', 'cancelled', 'archived')`, but omit `needs_attention`. In the codebase, `needs_attention` is the renamed `failed` status (migration 24) and IS treated as terminal throughout the runtime (see `room-runtime.ts` lines 2695, 2859 and `task-manager.ts` line 242 where `failTask()` sets status to `needs_attention`).

When `TaskGroupManager.fail()` is called, it calls BOTH `failGroup()` (sets group `completed_at`) AND `failTask()` (sets task to `needs_attention`). A zombie occurs when the group's `completed_at` was never set due to a crash, but the task already moved to `needs_attention`. Without including `needs_attention` in the zombie cleanup:

1. The zombie group persists in the DB with `completed_at IS NULL`
2. If the task is later revived to `pending` for retry, the partial unique index `idx_session_groups_active_ref` would prevent creating a new group for the task, blocking task execution

This is the exact class of bug this PR aims to fix. Add `'needs_attention'` to the terminal status list in both locations.

### P1: Migration 42 step 2 can fail if duplicate groups have identical `created_at` timestamps

**File:** `packages/daemon/src/storage/schema/migrations.ts` (lines 2499-2515)

Step 2 deduplicates active groups by finding `MAX(created_at)` and completing all groups with `created_at < max_created_at`. If two groups have identical `created_at` values (possible with millisecond-resolution timestamps, especially under concurrent tick races -- the exact scenario this fix targets), the `WHERE created_at < ?` clause won't match either group. Both remain active, and step 3's `CREATE UNIQUE INDEX` will fail because the data violates the uniqueness constraint.

This would crash the entire migration chain on startup, potentially bricking the application for users who have this data pattern. Consider using `ROWID` or `id` comparison instead (e.g., keep the row with `MAX(ROWID)` or `MAX(id)` and complete the rest), or wrap step 2 in a more robust dedup that uses a tiebreaker.

---

### P2: Migration step 1 does not increment `version`

**File:** `packages/daemon/src/storage/schema/migrations.ts` (line 2487-2495)

The runtime method `cleanupZombieGroupsForRoom()` increments `version` (`SET completed_at = ?, version = version + 1`), but the migration only sets `completed_at` without incrementing `version`. While this is a one-time migration and won't cause optimistic locking failures, it's inconsistent. If any code path later reads and re-saves these groups using version-based concurrency control, the stale version could cause unexpected behavior.

---

### What's good

- The layered defense approach (spawn check -> completion/failure cleanup -> stop-time cleanup -> migration + unique index) is comprehensive
- `getActiveGroupsForTask()` correctly fixes the root cause where `getGroupByTaskId()` only returned the newest group regardless of active status
- Tests cover the key scenarios: fresh DB, zombie cleanup, duplicate dedup, unique constraint enforcement, idempotency
- The partial unique index is the right permanent fix -- it prevents future duplicates at the DB level while allowing completed groups to coexist
- `cleanupStaleGroupsForTask()` in both `complete()` and `fail()` provides a belt-and-suspenders safety net

lsm (commented):

This PR correctly identifies and addresses the root causes of duplicate/stale session groups via a multi-layer defense strategy. The logic, migration, and tests are generally sound. However, there are several issues worth addressing before merging.

---

### P0 — Blocking

**Unique index is not filtered by group_type, creating a future landmine**

`packages/daemon/src/storage/schema/index.ts` (line 383) and `migrations.ts` (line 2518):
```sql
CREATE UNIQUE INDEX IF NOT EXISTS idx_session_groups_active_ref
  ON session_groups(ref_id) WHERE completed_at IS NULL
```
This index covers ALL group types, not just `task` and `task_pair`. If a future group type shares a `ref_id` with a currently-active task group (e.g., a room-level group whose `ref_id` is also a task ID), the insert would silently fail with a unique constraint violation. The index should be:
```sql
ON session_groups(ref_id, group_type) WHERE completed_at IS NULL
```
or filtered to `group_type IN ('task', 'task_pair')` (SQLite partial indexes support `WHERE` expressions but not `IN` lists, so this would require `group_type = 'task' OR group_type = 'task_pair'`). Alternatively, keep the current index but add a code comment explicitly documenting that `ref_id` is globally unique across ALL group types.

Currently `group_type` defaults to `'task'` and only `task`/`task_pair` are used, so this is not a current bug — but the schema has no `CHECK` constraint on `group_type`, making this a real future risk.

---

### P1 — Should Fix

**`reviveGroup` and `resetGroupForRestart` will throw unhandled SQLite exceptions if a unique constraint violation occurs**

Both `reviveGroup` and `resetGroupForRestart` in `session-group-repository.ts` (lines 435–480) set `completed_at = NULL` without checking whether another active group already exists for the same `ref_id`. Post-migration this state should not exist, but:

1. The new index means that if it ever does exist (e.g., a race between daemon restart and migration), the `UPDATE` will throw a SQLite UNIQUE constraint error — not return `null` (which the callers check for). The exception will propagate uncaught from `reviveTaskForMessage` at room-runtime.ts line 1633.

2. The callers — `reviveTaskForMessage` (room-runtime.ts:1633), `room-agent-tools.ts:404`, and `task-handlers.ts:456` — all treat a `null` return as the error condition, not an exception. There is no `try/catch` wrapping these calls.

The simplest fix is to wrap the `db.prepare(...).run(...)` call in `reviveGroup`/`resetGroupForRestart` in a try/catch and return `null` on constraint violation.

**Migration Step 1 and Step 2 do not increment `version`**

The migration sets `completed_at` but omits `version = version + 1` (unlike the runtime methods `cleanupZombieGroupsForRoom` and `cleanupStaleGroupsForTask` which both increment version). While the daemon restarts between migration and runtime so this cannot cause in-flight optimistic locking failures, the inconsistency creates a confusing divergence between the migration and the methods it mirrors. The runtime methods were intentionally added to also bump version (to ensure reactive signals fire correctly). The migration skips this, which means reactive listeners would not be notified (though they wouldn't exist during migration either). Worth making consistent.

**`cleanupStaleGroupsForTask` called AFTER `completeGroup` has no practical effect in steady state**

In `task-group-manager.ts` lines 566–575 and 600–608:
```ts
const updated = this.groupRepo.completeGroup(groupId, group.version);
// ...
const staleCount = this.groupRepo.cleanupStaleGroupsForTask(group.taskId, groupId);
```
After the unique index is in place, there can be at most one active group per task at any time. Once the primary group is completed (step 1), there are zero active groups. The stale cleanup call at step 2 will always return 0 in steady state. This is dead code in the long run. The safety net logic is better placed BEFORE `completeGroup` (clean up stale duplicates first, then complete the canonical one). This does not affect correctness today but undermines the intent of the safety net.

---

### P2 — Important Suggestion

**Unit test schema does not include the partial unique index**

The test setup in `session-group-repository.test.ts` (lines 14–77) manually creates tables without running migrations or `createIndexes()`. As a result, the tests for `getActiveGroupsForTask` (lines 278–299) manually insert two active groups for the same task — which would fail in production with the new constraint. The tests pass because they bypass the constraint entirely. A test that verifies `createGroup` throws (or the duplicate is blocked) when the unique index is active would be more faithful to the production schema.

The migration-42 test suite at `tests/unit/storage/migrations/migration-42_test.ts` does test the unique constraint enforcement (line 524–542), but the repository unit tests do not exercise the index.

**`getGroupByTaskId` is now inconsistent with the deduplication fix in `spawnGroupForTask`**

`getGroupByTaskId` in session-group-repository.ts (line 275) is still used in several places (room-agent-tools.ts, task-handlers.ts, room-runtime.ts:1626) and returns only the most recent group — not necessarily the active one. In `reviveTaskForMessage` (room-runtime.ts:1626), it uses `getGroupByTaskId` and then checks `group.completedAt`. If the DB invariant is enforced (at most one active group), this is fine. But it's worth noting that `getGroupByTaskId` was identified as the source of the original bug — those remaining callers should be audited to determine if they should use `getActiveGroupsForTask` instead.

---

### P3 — Nit

**Migration Step 1 uses string interpolation instead of parameterized query**

`migrations.ts` line 2488:
```ts
db.exec(`UPDATE session_groups SET completed_at = ${now} ...`);
```
Since `now` is `Date.now()` (a guaranteed integer), this is safe, but is inconsistent with Step 2 which uses `db.prepare(...).run(now, ...)`. Prefer the parameterized form for consistency, even though there is no injection risk here.

**Zombie cleanup on `stop()` fires after `observer.dispose()`**

In room-runtime.ts (lines 447–456), `this.observer.dispose()` is called before `cleanupZombieGroupsForRoom`. While the order is correct (we want to stop observing before doing DB cleanup), any reactive notification from `cleanupZombieGroupsForRoom` fires after the observer has already been disposed. This is fine since `notifyChange` goes to the reactive database layer, not the disposed observer — but a comment clarifying that the zombie cleanup fires DB-level reactive signals (not the runtime observer) would improve readability.

---

### Summary

The core strategy is sound and the migration is well-structured. The P1 issue around unhandled SQLite exceptions in `reviveGroup`/`resetGroupForRestart` when the unique constraint fires needs addressing, and the group_type scope of the unique index should be documented or narrowed. The P1 about `cleanupStaleGroupsForTask` ordering is a correctness concern for post-index steady state. The tests are thorough for the new methods but leave the production-schema constraint untested in the unit layer.

lsm (commented):

**Recommendation: APPROVE**

All five issues from the previous review have been addressed correctly:

### Feedback Verification

1. **`needs_attention` added to terminal status checks** -- Fixed in both `cleanupZombieGroupsForRoom()` (session-group-repository.ts line 386) and migration 42 step 1 (migrations.ts line 2496). The docstrings now explain why `needs_attention` is terminal.

2. **Migration dedup crash on same-timestamp groups** -- Fixed by switching from `MAX(created_at)` / `created_at < ?` to `MAX(rowid)` / `rowid < ?` (migrations.ts lines 2504, 2516). This eliminates the timestamp-tie failure that would have crashed `CREATE UNIQUE INDEX` in step 3. A dedicated test (`existing DB: duplicate groups with identical created_at are deduped via rowid`) validates this scenario at migration-42_test.ts line 354.

3. **Unique index scoped to task group types** -- Fixed in both `createIndexes()` (index.ts line 385-386) and migration 42 step 3 (migrations.ts line 2524-2525). The `WHERE` clause now includes `AND (group_type = 'task' OR group_type = 'task_pair')`. A test at migration-42_test.ts line 104 confirms non-task group types are not blocked by the constraint.

4. **`reviveGroup`/`resetGroupForRestart` wrapped in try/catch** -- Both methods now catch SQLite unique constraint violations and return `null` instead of throwing (session-group-repository.ts lines 454-458 and 492-496). This matches the callers' error handling pattern which checks for `null` return.

5. **`cleanupStaleGroupsForTask` moved before `completeGroup`/`failGroup`** -- In task-group-manager.ts, the stale cleanup now runs BEFORE the primary group is completed/failed (lines 568-575 for `complete()`, lines 604-611 for `fail()`). This means the safety net actually has work to do when duplicates exist, rather than running after the primary is already completed (when no active groups remain).

### Additional Fix

The migration now also increments `version` in both step 1 and step 2 (`SET completed_at = ?, version = version + 1`), addressing the P2 consistency issue from the first review.

### Remaining Minor Items (not blocking)

- `createGroup` in session-group-repository.ts (line 231) will throw on unique constraint violation rather than returning null. This is acceptable because `spawnGroupForTask` (room-runtime.ts line 3161) guards with `getActiveGroupsForTask` before ever calling `spawn()` -> `createGroup()`, making the throw path extremely unlikely. If it does fire, it prevents silent data corruption, which is the right behavior.

- `getGroupByTaskId` still returns the most recent group regardless of active status and is used in several callers. This is a pre-existing pattern and not introduced by this PR -- the unique index now guarantees at most one active group per task, so the existing callers will find the right group in practice.

No P0, P1, or P2 issues remain. The implementation is solid, well-tested, and the defense-in-depth layering (spawn guard, pre-completion cleanup, stop-time zombie cleanup, migration data fix, DB-level unique constraint) provides comprehensive protection against duplicate session groups.

lsm (commented):

**Recommendation: APPROVE** (same-author PR; posting as COMMENT)

All five P0/P1/P2 issues from the previous review have been correctly addressed. The implementation is solid and can be merged.

---

### Resolved issues (confirmed)

1. **`needs_attention` in terminal status lists** — both `cleanupZombieGroupsForRoom` (session-group-repository.ts:385) and migration step 1 (migrations.ts:2496) now include `needs_attention`. Tests added in the migration test file (line 208) and repository test file (line 846) verify this case.

2. **Migration dedup crash on same-timestamp groups** — step 2 now uses `MAX(rowid)` and `rowid < ?` (migrations.ts:2504-2517) instead of `created_at`, which is a correct tiebreaker. A new test (migration-42_test.ts:354) verifies the same-timestamp case does not throw.

3. **Unique index scoped to task/task_pair** — both the schema index (index.ts:384-387) and migration step 3 (migrations.ts:2523-2526) now include `AND (group_type = 'task' OR group_type = 'task_pair')` in the WHERE clause. A new test (migration-42_test.ts:104) verifies non-task types sharing a ref_id are not blocked.

4. **`reviveGroup`/`resetGroupForRestart` try/catch** — both methods (session-group-repository.ts:440-459 and 466-497) now wrap the UPDATE in try/catch and return `null` on constraint violation. Callers in room-runtime.ts:1633, room-agent-tools.ts:389/404, and task-handlers.ts:456 all check for `null` and handle it gracefully.

5. **`cleanupStaleGroupsForTask` ordering fixed** — now called BEFORE `completeGroup`/`failGroup` in both `complete()` (task-group-manager.ts:568) and `fail()` (task-group-manager.ts:604), making it effective as a safety net even when the unique index is enforced.

---

### Remaining minor issues

**P2 — Bare `catch {}` swallows all exceptions, not just unique constraint violations**

Files: `session-group-repository.ts` lines 454 and 492.

The comment says "Unique constraint violation" but the catch block intercepts any SQLite error: disk full, DB closed, `SQLITE_READONLY`, etc. These would all silently return `null` and be treated as "another active group exists," masking the real problem. In practice, callers tolerate `null` gracefully so this is not a crash risk, but it makes diagnosing unexpected DB failures harder. Consider inspecting the error message/code before swallowing it.

**P3 — No test for `reviveGroup` returning `null` due to unique constraint violation**

The `reviveGroup` and `resetGroupForRestart` test suites (session-group-repository.test.ts:333-377 and 270-331) use an in-memory DB that does not install the unique index (no `runMigrations` call). There is no test verifying that attempting to revive a group when another active group already exists for the same `ref_id` returns `null` instead of throwing. While the migration-42 tests verify the index works, coverage of the exception path in these two methods is absent.

---

### What is good

- The layered defense (spawn check with `getActiveGroupsForTask` + cleanup before complete/fail + stop-time zombie cleanup + migration + unique index) is comprehensive and each layer makes sense independently.
- `getGroupByTaskId` returning the most-recent group regardless of `completed_at` is preserved (existing callers need it for recovery flows) while `getActiveGroupsForTask` provides the correct active-only view for dedup.
- Migration idempotency test and the `IF NOT EXISTS` on the index creation are correct.
- Consistent `version = version + 1` in migration steps 1 and 2 matches the runtime methods.

lsm (commented):

**Recommendation: APPROVE** (same-author PR; posting as COMMENT)

The P2 from the previous review — bare `catch {}` blocks in `reviveGroup` and `resetGroupForRestart` swallowing all SQLite exceptions — has been correctly resolved. Both methods now use `err instanceof Error && err.message.includes('UNIQUE constraint failed')` before returning `null`, and rethrow everything else. The P3 (no test for the null-on-unique-constraint path) has also been addressed: two new tests in `session-group-repository.test.ts` (lines 332–359 and 407–434) install the partial unique index into the in-memory DB and verify that attempting to revive/restart a completed group when an active group already holds the `ref_id` returns `null` instead of throwing.

### Confirmed

- `reviveGroup` (session-group-repository.ts:454–461): catches only `UNIQUE constraint failed`, rethrows all other errors.
- `resetGroupForRestart` (session-group-repository.ts:495–502): same pattern.
- Test at line 332 (`resetGroupForRestart`): installs index, inserts conflicting active group, asserts `null` returned.
- Test at line 407 (`reviveGroup`): same setup, same assertion.
- The `err instanceof Error` guard is appropriate because `better-sqlite3` (Bun's SQLite driver) always throws `Error` instances for constraint violations.

### Minor remaining gap (P3, non-blocking)

Neither test suite exercises the rethrow path — i.e., there is no test confirming that a non-UNIQUE error (e.g., closed DB) propagates as a thrown exception rather than silently returning `null`. The code is correct, and adding such a test would require injecting a mock DB or forcing a DB error mid-operation, which is non-trivial. Not a blocker, but worth noting for future test completeness.

### Summary

Both the P2 and P3 from the previous round are addressed. The fix is correct, minimal, and well-tested. Ready to merge.

lsm (commented):

### Re-review: P2 catch block narrowing

The previous review flagged that `reviveGroup` and `resetGroupForRestart` had bare catch blocks that swallowed all SQLite exceptions. This has been properly addressed in commit `923b227d2`.

**What changed:**

Both methods in `packages/daemon/src/lib/room/state/session-group-repository.ts` (lines 454-461 and 495-502) now:
1. Check `err instanceof Error` (handles non-Error throws safely)
2. Check `err.message.includes('UNIQUE constraint failed')` to match only the expected SQLite unique index violation
3. Return `null` only for that specific case
4. Rethrow all other errors

This is the correct approach for Bun's SQLite driver, which throws `Error` objects with the constraint message in `.message`.

**Tests added:**

Two new test cases in `packages/daemon/tests/unit/room/session-group-repository.test.ts`:
- `returns null (not throws) when unique constraint prevents restart` (line 332) -- creates the partial unique index, sets up a completed group + an active group for the same task, then verifies `resetGroupForRestart` returns `null` rather than throwing.
- `returns null (not throws) when unique constraint prevents revive` (line 407) -- same pattern for `reviveGroup`.

Both tests correctly create the `idx_session_groups_active_ref` partial unique index (which the test DB setup doesn't include by default since tests use a minimal schema), insert the conflicting rows, and assert the `null` return.

**Minor observation (P3):** The tests verify the happy path (constraint violation returns null) but do not explicitly test the rethrow path (a non-unique-constraint SQLite error propagates). This is a minor gap -- the rethrow is a simple `throw err` and the logic is straightforward enough that the risk is low.

### Overall assessment

The P2 issue from the previous review has been fully resolved. The catch blocks are now properly narrowed, well-documented with inline comments, and covered by tests. The rest of the PR (unique index migration, zombie cleanup, `getActiveGroupsForTask`, `cleanupZombieGroupsForRoom`) remains solid from the previous review rounds.

**Recommendation: APPROVE**

PR #698, opened 2026-03-22
feat: replace state.groupMessages.delta listener with useGroupMessages hook

lsm:
Removes the old state.groupMessages.delta event listener and associated
fetch/buffering/dedup/pagination logic from TaskConversationRenderer.
Replaces with useGroupMessages (LiveQuery) which delivers an initial
snapshot plus append-only deltas automatically.

Updates the test file to mock useGroupMessages directly, removing tests
for implementation details (buffering, dedup) that are now covered by
the useGroupMessages hook's own test suite.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2191,14 +2191,21 @@ export class RoomRuntime {
 			messageType === 'rate_limited' || messageType === 'model_fallback'
 				? JSON.stringify({ ...payload, type: messageType })
 				: (payload?.text ?? kind);
-		this.groupRepo.appendGroupMessage({
-			groupId,
-			sessionId: null,
-			role: 'system',
-			messageType,
-			content,
-			createdAt: now,
-		});
+		try {
+			this.groupRepo.appendGroupMessage({
+				groupId,
+				sessionId: null,
+				role: 'system',
+				messageType,
+				content,
+				createdAt: now,
+			});
+		} catch (err) {
+			log.warn(
+				`appendGroupEvent: failed to persist group message for group ${groupId} (type=${messageType}, secondary write — ignored):`,
+				err
+			);
+		}
 
 		if (this.messageHub) {
 			this.messageHub.event(
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2065,12 +2065,9 @@ export class RoomRuntime {
 	/**
 	 * Set up live message forwarding for a group's worker/leader sessions.
 	 *
-	 * Subscribes to DaemonHub sdk.message events and broadcasts enriched deltas
-	 * on state.groupMessages.delta for TaskConversationRenderer.
-	 *
-	 * IMPORTANT: We intentionally do NOT mirror worker/leader SDK messages into
-	 * any group message table anymore. Task view history is reconstructed from each
-	 * session's sdk_messages stream via task.getGroupMessages.
+	 * Subscribes to DaemonHub sdk.message events, persists each enriched message
+	 * to session_group_messages (for LiveQuery), and broadcasts deltas via
+	 * state.groupMessages.delta for any legacy subscribers still present.
 	 */
 	private setupMirroring(group: SessionGroup): void {
 		if (!this.daemonHub) return;
@@ -2113,20 +2110,33 @@ export class RoomRuntime {
 					const iteration = currentGroup?.feedbackIteration ?? group.feedbackIteration;
 					const turnId = `turn_${group.id}_${iteration}_${shortSessionId}`;
 
-					// Broadcast enriched delta to subscribed frontends.
+					// Persist to session_group_messages and broadcast enriched delta to frontends.
+					const enrichedMessage = {
+						...event.message,
+						_taskMeta: {
+							authorRole: role,
+							authorSessionId: sessionId,
+							turnId,
+							iteration,
+						},
+					};
+					const sdkNow = Date.now();
+					const sdkMsgType =
+						'type' in event.message && typeof event.message.type === 'string'
+							? event.message.type
+							: 'assistant';
+					this.groupRepo.appendGroupMessage({
+						groupId: group.id,
+						sessionId,
+						role,
+						messageType: sdkMsgType,
+						content: JSON.stringify(enrichedMessage),
+						createdAt: sdkNow,
+					});
 					if (this.messageHub) {
-						const enrichedMessage = {
-							...event.message,
-							_taskMeta: {
-								authorRole: role,
-								authorSessionId: sessionId,
-								turnId,
-								iteration,
-							},
-						};
 						this.messageHub.event(
 							'state.groupMessages.delta',
-							{ added: [{ ...enrichedMessage, timestamp: Date.now() }], timestamp: Date.now() },
+							{ added: [{ ...enrichedMessage, timestamp: sdkNow }], timestamp: sdkNow },
 							{ channel: `group:${group.id}` }
 						);
 					}
@@ -2159,20 +2169,38 @@ export class RoomRuntime {
 			kind,
 			payloadJson: payload ? JSON.stringify(payload) : undefined,
 		});
+		const now = Date.now();
+		// Map event kinds to message types for the frontend.
+		// 'leader_summary' is a special case (rendered as a distinct card).
+		// 'rate_limited' and 'model_fallback' get their own type so the frontend
+		// can render them as prominent notifications.
+		const messageType =
+			kind === 'leader_summary'
+				? 'leader_summary'
+				: kind === 'rate_limited'
+					? 'rate_limited'
+					: kind === 'model_fallback'
+						? 'model_fallback'
+						: 'status';
+
+		// Persist to session_group_messages so LiveQuery subscribers receive the event.
+		// For rich event types (rate_limited, model_fallback) store the full payload as
+		// JSON so the frontend can render extra fields (resetsAt, sessionRole, etc.).
+		// For simple status/leader_summary, store just the text.
+		const content =
+			messageType === 'rate_limited' || messageType === 'model_fallback'
+				? JSON.stringify({ ...payload, type: messageType })
+				: (payload?.text ?? kind);
+		this.groupRepo.appendGroupMessage({
+			groupId,
+			sessionId: null,
+			role: 'system',
+			messageType,
+			content,
+			createdAt: now,
+		});
+
 		if (this.messageHub) {
-			const now = Date.now();
-			// Map event kinds to message types for the frontend.
-			// 'leader_summary' is a special case (rendered as a distinct card).
-			// 'rate_limited' and 'model_fallback' get their own type so the frontend
-			// can render them as prominent notifications.
-			const messageType =
-				kind === 'leader_summary'
-					? 'leader_summary'
-					: kind === 'rate_limited'
-						? 'rate_limited'
-						: kind === 'model_fallback'
-							? 'model_fallback'
-							: 'status';
 			this.messageHub.event(
 				'state.groupMessages.delta',
 				{
@@ -2181,6 +2209,7 @@ export class RoomRuntime {
 							type: messageType,
 							text: payload?.text ?? kind,
 							timestamp: now,
+							...(payload ?? {}),
 						},
 					],
 					timestamp: now,
--- a/packages/daemon/src/lib/room/state/session-group-repository.ts
+++ b/packages/daemon/src/lib/room/state/session-group-repository.ts
@@ -837,6 +837,37 @@ export class SessionGroupRepository {
 		});
 	}
 
+	// ===== session_group_messages (LiveQuery message streaming) =====
+
+	/**
+	 * Insert a row into session_group_messages and notify LiveQuery subscribers.
+	 * Called from both setupMirroring (SDK messages) and appendGroupEvent (status/event messages).
+	 */
+	appendGroupMessage(params: {
+		groupId: string;
+		sessionId?: string | null;
+		role: string;
+		messageType: string;
+		content: string;
+		createdAt: number;
+	}): number {
+		const result = this.db
+			.prepare(
+				`INSERT INTO session_group_messages (group_id, session_id, role, message_type, content, created_at)
+			 VALUES (?, ?, ?, ?, ?, ?)`
+			)
+			.run(
+				params.groupId,
+				params.sessionId ?? null,
+				params.role,
+				params.messageType,
+				params.content,
+				params.createdAt
+			);
+		this.reactiveDb.notifyChange('session_group_messages');
+		return Number(result.lastInsertRowid);
+	}
+
 	// ===== Group events (status/system timeline, no mirrored SDK chat) =====
 
 	appendEvent(params: { groupId: string; kind: string; payloadJson?: string }): number {
--- a/packages/daemon/tests/unit/room/room-agent-tools.test.ts
+++ b/packages/daemon/tests/unit/room/room-agent-tools.test.ts
@@ -105,6 +105,15 @@ describe('Room Agent Tools', () => {
 				payload_json TEXT,
 				created_at INTEGER NOT NULL
 			);
+			CREATE TABLE session_group_messages (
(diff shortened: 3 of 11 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean refactor that replaces ~300 lines of manual fetch/buffer/dedup/pagination logic with a single `useGroupMessages(groupId)` hook call. The implementation is correct and well-structured.

### What looks good

- **Complete removal of old patterns**: No remaining references to `state.groupMessages.delta` or `task.getGroupMessages` anywhere in the web package.
- **Correct hook integration**: `useGroupMessages` returns `{ messages, isLoading }` and the component handles both states (loading spinner, empty state, message list) properly.
- **Test coverage**: Tests cover loading state, empty state, message count callback, status message rendering, question state isolation between leader/worker sessions, and backward compatibility without optional props. The removed tests (buffering, dedup, pagination) are correctly identified as implementation details now owned by the `useGroupMessages` hook's own test suite.
- **`useMemo` on parse**: `rawMessages.map(parseGroupMessage).filter(...)` is wrapped in `useMemo` keyed on `rawMessages`, avoiding unnecessary re-parsing on unrelated re-renders.

### Minor observations (P3)

1. **Redundant local `GroupMessage` interface** (`TaskConversationRenderer.tsx` lines 39-47): This interface is structurally identical to `SessionGroupMessage` exported from `useGroupMessages.ts`. The component already imports `useGroupMessages` — it could import and reuse `SessionGroupMessage` instead of maintaining a duplicate type definition. Not a bug due to structural typing, just unnecessary duplication.

2. **Test mock ignores `groupId`** (test file line 55): The `useGroupMessages` mock returns the same data regardless of what `groupId` is passed. This is fine for the current single-group test scenarios but means tests wouldn't catch a bug where the component passes the wrong groupId. Acceptable tradeoff for unit tests.

Overall this is a solid simplification. No correctness issues found.

lsm (commented):

> **Recommendation: REQUEST_CHANGES**

### Summary

The PR correctly replaces all old buffering/dedup/pagination state machinery with a clean `useGroupMessages` hook call. The component reduction (~300 lines) is clean and the test rewrite is well-structured. However, there is one P0 issue that blocks shipping and two secondary issues worth noting.

---

### P0 — The `session_group_messages` table has no writer: the component will always show "Waiting for agent activity"

The `useGroupMessages` hook subscribes to `liveQuery.subscribe` with `queryName: 'sessionGroupMessages.byGroup'`, which queries the `session_group_messages` DB table. **No code in the daemon currently writes to this table.** The schema comment itself says:

> `packages/daemon/src/storage/schema/index.ts:308`: *"Populated by the LiveQuery message streaming path (Milestone 4)."*

The `live-query-handlers.ts` comment on line 157 repeats this:

> `packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts:157`: *"session_group_messages is an append-only table populated in Milestone 4."*

The backend still emits all messages via `state.groupMessages.delta` WebSocket events (in `room-runtime.ts` at lines 2127–2131 and 2176–2188), and the `appendGroupEvent` function writes only to `task_group_events`, **not** `session_group_messages`. Because the table is always empty, the LiveQuery snapshot will always return zero rows. The frontend will fall through to the "Waiting for agent activity…" empty state for every task conversation — all live messages will be silently dropped.

The fix requires either:
1. Keeping the old `state.groupMessages.delta` listener in parallel until Milestone 4 lands the write path, or
2. Landing the `session_group_messages` INSERT path in this PR alongside the frontend change.

As written, this PR removes the only working message display path without replacing it.

---

### P1 — Removed "Load older messages" without a replacement

The old component had a paginated history viewer ("Load older messages" button, backed by `task.getGroupMessages` with cursor pagination). That feature is entirely removed by this PR. The server-side `task.getGroupMessages` RPC still exists and still works. If a task group has more than the initial page worth of messages, users can no longer scroll back through history. This is a user-visible regression unless Milestone 4's LiveQuery path guarantees full history in the snapshot (the SQL query has no LIMIT clause, which could be a performance concern for long-running tasks).

---

### P2 — `GroupMessage` interface is now a duplicate of the exported `SessionGroupMessage`

`packages/web/src/components/room/TaskConversationRenderer.tsx` lines 39–47 still define a local `GroupMessage` interface with an identical shape to the `SessionGroupMessage` exported from `useGroupMessages.ts`. The `parseGroupMessage` function takes a `GroupMessage` parameter, and the call site passes `SessionGroupMessage[]\ — TypeScript accepts this via structural typing, but the local interface is now redundant. It should be removed and `parseGroupMessage` should accept `SessionGroupMessage` directly to eliminate the duplicate definition.

---

### P3 — Dead mock infrastructure in the test file

`packages/web/src/components/room/TaskConversationRenderer.test.tsx` lines 26–36 define `sessionStateHandlers`, `mockOnEvent`, `mockJoinRoom`, and `mockLeaveRoom` with elaborate state — but `mockOnEvent` and the join/leave mocks are now only needed for the `useSessionQuestionState` hook (which does the join/leave for session channels), not for message loading. The `mockOnEvent` handler that previously captured `state.groupMessages.delta` is now empty except for `state.session` events. The mock is fine as-is, but the comment on line 25 ("state.session handlers keyed by the channel...") is accurate and worth keeping. Low-priority cleanup.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review after the P0 (missing write path) and P2 (duplicate `GroupMessage` type) from the previous review were addressed. Both issues are now resolved. The implementation is correct and well-integrated.

### What was fixed since last review

1. **P0 resolved**: `SessionGroupRepository.appendGroupMessage()` now writes to `session_group_messages` in both `setupMirroring` (SDK messages) and `appendGroupEvent` (status/event messages). The LiveQuery path is now fully functional end-to-end.

2. **P2 resolved**: The duplicate local `GroupMessage` interface was removed. `parseGroupMessage` now accepts `SessionGroupMessage` directly from the `useGroupMessages` hook.

3. **Test schemas updated**: All 6 test files with inline DB schemas now include the `session_group_messages` table definition, matching the production schema exactly.

4. **New unit tests**: `appendGroupMessage` has 3 targeted tests covering: insert + autoincrement ID, column verification via raw SELECT, null sessionId for system events, and ReactiveDatabase notification for LiveQuery.

### What looks good

- The `enrichedMessage` object is constructed once and reused for both the DB insert (serialized as `content`) and the legacy WebSocket broadcast, avoiding redundant work.
- `parseGroupMessage` handles `rate_limited` and `model_fallback` gracefully: tries JSON parse, falls back to plain text on failure. This is defensive and correct.
- The `messageType` mapping in `appendGroupEvent` correctly categorizes `leader_summary`, `rate_limited`, `model_fallback` as distinct types and falls back to `status` for everything else.
- Legacy `state.groupMessages.delta` broadcast is retained alongside the new DB write for backward compatibility, with a clear comment explaining its presence.

### P2 — No error handling around `appendGroupMessage` in `setupMirroring` event callback

`packages/daemon/src/lib/room/runtime/room-runtime.ts` line 2128: The `appendGroupMessage` call inside the `sdk.message` event handler has no try/catch. If the group is deleted while messages are still flowing (e.g., during cleanup races), the foreign key constraint on `group_id` will throw a synchronous SQLite error, which would propagate up into the DaemonHub event dispatch and could crash the event loop or suppress subsequent event handlers for the same event.

I note this is consistent with the existing `appendEvent` pattern (line 2167), so it is a pre-existing concern rather than a regression introduced by this PR. Still worth noting for a follow-up.

### P3 — Minor redundancy in rate_limited/model_fallback content serialization

In `appendGroupEvent` (line 2191), the content for rich event types is `JSON.stringify({ ...payload, type: messageType })`. Then in `parseGroupMessage` on the frontend (lines 103-106), the parsed JSON is spread and `type` is explicitly set again. This is harmless (the explicit `type` wins over the spread) but slightly redundant — the `type` inside the stored JSON is never actually needed since `messageType` on the DB row already carries this information.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (I'm the PR author so posting as COMMENT)

This is a solid implementation. The backend write path is correctly wired, the migration is idempotent, the test additions are targeted and meaningful, and the frontend cleanup is clean. A few issues deserve attention before merging.

---

### P1 — No migration test for Migration 41

Every other significant migration (28, 29, 30, 33, 34, 35-36, 38, 40) has a corresponding test file in `packages/daemon/tests/unit/storage/migrations/`. Migration 41 recreates the `session_group_messages` table for existing databases — this is the most consequential part of this PR for production deployments — but has no test.

The migration has an `IF NOT EXISTS` guard and an existence check, but the guard means the behavior is different for fresh vs existing DBs. A test should verify:
- Running migrations on a DB that already has the table is a no-op (idempotency)
- Running migrations on a DB that does NOT have the table creates it with correct schema
- The index `idx_sgm_group` is created

File to add: `packages/daemon/tests/unit/storage/migrations/migration-41_test.ts`

---

### P1 — `appendGroupEvent` now issues two writes unconditionally, with no error isolation

In `room-runtime.ts` (line 2167-2201), `appendGroupEvent` calls both `groupRepo.appendEvent()` (to `task_group_events`) and `groupRepo.appendGroupMessage()` (to `session_group_messages`). If the second write fails (e.g., the group was already deleted and the FK cascade removed the row), the exception will propagate upward and abort the calling code — potentially corrupting the leader state machine mid-tick.

The original `appendEvent` write was always non-fatal at the tick level, but `appendGroupMessage` is new. Consider whether a failure in `appendGroupMessage` should be allowed to propagate or whether it should be caught-and-logged, since it is a secondary write (the primary event is in `task_group_events`).

---

### P2 — `session.get` called with `undefined` when session props are omitted

In `TaskConversationRenderer.tsx` (lines 180-181), `request('session.get', { sessionId: undefined })` is called whenever `leaderSessionId` or `workerSessionId` is not provided. This was pre-existing code, but the PR refactored enough of the file that it is worth noting. Each mount of the component without session props will fire two RPC calls that will return nothing useful and the error is silently swallowed. The `useEffect` should guard with `if (!leaderSessionId && !workerSessionId) return` before the fetch.

---

### P2 — Double-write in `setupMirroring`: `Date.now()` called once but `createdAt` and broadcast timestamp can still diverge

In `room-runtime.ts` (lines 2123-2142), `sdkNow` is assigned via `Date.now()` and used for both `appendGroupMessage({ createdAt: sdkNow })` and the broadcast `{ ..., timestamp: sdkNow }`. This is an improvement over the old code. However, the `content` stored in `session_group_messages` is `JSON.stringify(enrichedMessage)` (which has no `timestamp` field), while the WebSocket broadcast adds `timestamp: sdkNow`. This asymmetry is fine architecturally since LiveQuery delivers the DB row directly, but the comment at line 2113 could note this explicitly so future maintainers don't add a timestamp to the persisted content.

---

### P3 — Stale comment describing `state.groupMessages.delta` as 'legacy'

Lines 2069-2070 in `room-runtime.ts` now say:

> broadcasts deltas via state.groupMessages.delta for any legacy subscribers still present

Since `TaskConversationRenderer` no longer subscribes, clarify whether any other subscriber exists. If none, the WebSocket broadcast in `setupMirroring` and `appendGroupEvent` is dead code and can be removed in a follow-up.

---

- `appendGroupMessage` correctly calls `reactiveDb.notifyChange('session_group_messages')` so LiveQuery subscribers are triggered synchronously.
- The unit tests for `appendGroupMessage` cover insert, retrieval, null sessionId, and LiveQuery notification — all the right cases.
- The test schema additions across all affected test files are consistent and match the production schema exactly.
- `parseGroupMessage` in `TaskConversationRenderer` handles JSON parse failures gracefully (returns `null`, filtered by the `useMemo`).
- The `leader_summary` content path: `appendGroupEvent` stores `payload?.text ?? kind` which for `leader_summary` is the `[Turn Summary] ...` text, and the frontend correctly strips the prefix. Consistent end-to-end.
- Test file slimming is appropriate — buffering/dedup tests correctly belong in the `useGroupMessages` hook's own test suite, and the new component tests accurately reflect the simplified contract.

lsm (commented):

**Recommendation: APPROVE**

This is round 3 re-review. All three items from the previous REQUEST_CHANGES review have been addressed correctly.

### Fixes verified

1. **P1 — Migration 41 test** (`packages/daemon/tests/unit/storage/migrations/migration-41_test.ts`): 8 tests covering fresh DB (table exists, correct columns, defaults for `role` and `message_type`, index), idempotency (run twice without error), and existing DB without the table (migration creates it with correct schema). The tests use raw `BunDatabase` and `PRAGMA table_info` for schema verification, which is the right approach — no runtime coupling. Clean setup/teardown with temp directories.

2. **P1 — Error isolation in `appendGroupEvent`** (`packages/daemon/src/lib/room/runtime/room-runtime.ts` lines 2194-2208): The `appendGroupMessage` call is now wrapped in try-catch with a `log.warn`. This correctly isolates the secondary write (to `session_group_messages`) from the primary event write (to `task_group_events`), so FK constraint failures during cleanup races won't propagate and crash the leader state machine.

3. **P2 — Guard for undefined session IDs** (`packages/web/src/components/room/TaskConversationRenderer.tsx` lines 176-186): Early return when both `leaderSessionId` and `workerSessionId` are undefined, plus conditional `Promise.resolve({ session: null })` for each individually undefined ID. This eliminates the spurious `session.get` RPC calls with `undefined` sessionId.

### Remaining observation (P3)

The `appendGroupMessage` call in `setupMirroring` (line 2128) still lacks error isolation — unlike the one in `appendGroupEvent` that was just wrapped. This was flagged in the round 2 review as a pre-existing concern (consistent with the `appendEvent` pattern) rather than a regression, and remains unchanged. If a group is deleted while SDK messages are still flowing, the FK constraint failure will propagate into the DaemonHub event handler. This is acceptable as a follow-up item since the cleanup sequence normally unsubscribes mirroring before deleting the group.

lsm (commented):

**Recommendation: APPROVE**

Round 3 re-review. All three items from the previous REQUEST_CHANGES are addressed correctly. No new issues introduced.

---

### P1 — Migration-41 test: RESOLVED

`packages/daemon/tests/unit/storage/migrations/migration-41_test.ts` is added with 8 tests covering:
- Fresh DB: table exists, all columns present, role defaults to 'system', message_type defaults to 'status', idx_sgm_group index exists
- Idempotency: running migrations twice does not error
- Pre-migration-41 DB: table is created with correct schema, index is created

The test structure correctly uses a real on-disk BunDatabase (not in-memory) with the full `runMigrations` + `createTables` chain, matching the production initialization order. The pre-migration-41 test setup creates only the `session_groups` dependency table to simulate the targeted upgrade scenario. This is the right approach.

---

### P1 — try-catch in appendGroupEvent: RESOLVED

`packages/daemon/src/lib/room/runtime/room-runtime.ts` lines 2194-2208 now wrap the `appendGroupMessage` call in `appendGroupEvent` with a try-catch that logs a warning and continues. The primary `appendEvent` write to `task_group_events` is unaffected, and the WebSocket broadcast path is also outside the try block and will still fire. This correctly treats the LiveQuery write as a secondary concern that should not disrupt the leader state machine.

The `appendGroupMessage` call in `setupMirroring` (line 2128) has no try-catch, which is consistent with the pre-existing `appendEvent` call pattern there. Since the previous review noted this is a pre-existing concern rather than a regression, not holding this PR for it.

---

### P2 — session.get undefined guard: RESOLVED

`packages/web/src/components/room/TaskConversationRenderer.tsx` line 176 now has `if (!leaderSessionId && !workerSessionId) return;` as an early exit. Each individual call site (lines 181-186) also conditionally short-circuits to `Promise.resolve({ session: null })` when the respective ID is absent, so neither undefined session ID can reach the RPC.

---

### Remaining P3 observations (no action required)

- The `state.groupMessages.delta` WebSocket broadcast in `setupMirroring` and `appendGroupEvent` is now effectively dead code since `TaskConversationRenderer` no longer listens for it. The comment on line 2070 says 'for any legacy subscribers still present' — if no other subscriber exists in the web package (confirmed: grep finds zero references in `packages/web/src`), this broadcast is unused bandwidth. Clean-up can be deferred to a follow-up.
- The test for 'notifies reactiveDb so LiveQuery subscribers are triggered' (`session-group-repository.test.ts` line 334) allocates a new `ReactiveDatabase` instance wrapping the same underlying `db`. This pattern is fine and the test is correct.

---

### Verdict

All requested changes are implemented correctly. The migration test covers the right scenarios, the error isolation is properly scoped to the secondary write, and the undefined guard eliminates the wasted RPC calls. Clean to merge.

PR #702, opened 2026-03-22
fix: fix task cancellation cascade and allow manual UI state transitions

lsm:
- Fix bug in room-agent-tools.ts: only cascade cancellation for 'cancelled'
  transitions; use terminateTaskGroup for 'completed'/'needs_attention' to
  avoid incorrectly cancelling dependent tasks
- Add mode parameter ('runtime'|'manual') to setTaskStatus allowing UI to
  bypass strict state machine rules for manual transitions
- Add SetStatusModal in TaskView for arbitrary status changes with
  destructive transition warnings
- Support unarchiving by clearing archivedAt when transitioning from archived
- Update task-handlers RPC to accept and pass through manual mode
- Add unit tests for manual mode in task-manager and task-handlers

### DIFF
--- a/packages/daemon/src/lib/room/managers/task-manager.ts
+++ b/packages/daemon/src/lib/room/managers/task-manager.ts
@@ -188,15 +188,15 @@ export class TaskManager {
 	async setTaskStatus(
 		taskId: string,
 		newStatus: TaskStatus,
-		options?: { result?: string; error?: string }
+		options?: { result?: string; error?: string; mode?: 'runtime' | 'manual' }
 	): Promise<NeoTask> {
 		const task = await this.getTask(taskId);
 		if (!task) {
 			throw new Error(`Task not found: ${taskId}`);
 		}
 
-		// Validate transition
-		if (!isValidStatusTransition(task.status, newStatus)) {
+		// Validate transition (skipped in manual mode — UI allows any transition)
+		if (options?.mode !== 'manual' && !isValidStatusTransition(task.status, newStatus)) {
 			throw new Error(
 				`Invalid status transition from '${task.status}' to '${newStatus}'. ` +
 					`Allowed transitions: ${VALID_STATUS_TRANSITIONS[task.status].join(', ') || 'none'}`
@@ -220,16 +220,31 @@ export class TaskManager {
 		}
 
 		// Clear error/result/progress when restarting from a terminal/failed state.
-		// Covers needs_attention, cancelled, and completed → reactivation transitions.
+		// Covers needs_attention, cancelled, completed → reactivation transitions.
+		// In manual mode, also covers archived → active and completed → pending transitions.
 		if (
 			((task.status === 'needs_attention' || task.status === 'cancelled') &&
 				(newStatus === 'pending' || newStatus === 'in_progress' || newStatus === 'review')) ||
-			(task.status === 'completed' && newStatus === 'in_progress')
+			(task.status === 'completed' && newStatus === 'in_progress') ||
+			(options?.mode === 'manual' &&
+				(task.status === 'archived' ||
+					task.status === 'completed' ||
+					task.status === 'cancelled' ||
+					task.status === 'needs_attention') &&
+				(newStatus === 'pending' ||
+					newStatus === 'in_progress' ||
+					newStatus === 'review' ||
+					newStatus === 'completed'))
 		) {
 			// Use null to explicitly clear these fields in the database
 			updates.error = null;
 			updates.result = null;
 			updates.progress = null;
+		}
+
+		// When transitioning FROM archived (unarchiving), clear the archived_at timestamp
+		if (task.status === 'archived') {
+			updates.archivedAt = null;
 		}
 
 		return this.updateTaskStatus(taskId, newStatus, updates);
--- a/packages/daemon/src/lib/room/managers/task-manager.ts
+++ b/packages/daemon/src/lib/room/managers/task-manager.ts
@@ -441,13 +441,13 @@ export class TaskManager {
 	 * Archive task - transitions to 'archived' status and sets archivedAt timestamp.
 	 * Validates that the current status allows transitioning to 'archived'.
 	 */
-	async archiveTask(taskId: string): Promise<NeoTask> {
+	async archiveTask(taskId: string, options?: { mode?: 'runtime' | 'manual' }): Promise<NeoTask> {
 		const task = await this.getTask(taskId);
 		if (!task) {
 			throw new Error(`Task not found: ${taskId}`);
 		}
 
-		if (!isValidStatusTransition(task.status, 'archived')) {
+		if (options?.mode !== 'manual' && !isValidStatusTransition(task.status, 'archived')) {
 			throw new Error(
 				`Cannot archive task in '${task.status}' status. ` +
 					`Allowed transitions: ${VALID_STATUS_TRANSITIONS[task.status].join(', ') || 'none'}`
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1757,7 +1757,10 @@ export class RoomRuntime {
 	 * 2. Cleans up the worktree to free disk space.
 	 * 3. Sets the task status to 'archived' with archivedAt timestamp.
 	 */
-	async archiveTaskGroup(taskId: string): Promise<boolean> {
+	async archiveTaskGroup(
+		taskId: string,
+		options?: { mode?: 'runtime' | 'manual' }
+	): Promise<boolean> {
 		const group = this.groupRepo.getGroupByTaskId(taskId);
 
 		if (group) {
@@ -1784,7 +1787,7 @@ export class RoomRuntime {
 		}
 
 		// Set archivedAt timestamp on task (transitions to 'archived' status)
-		await this.taskManager.archiveTask(taskId);
+		await this.taskManager.archiveTask(taskId, { mode: options?.mode });
 
 		return true;
 	}
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -357,17 +357,42 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 					if (runtime) {
 						const group = groupRepo.getGroupByTaskId(args.task_id);
 						if (group && group.completedAt === null) {
-							// There's an active group - cancel it first if moving to terminal state
-							if (
-								args.status === 'completed' ||
-								args.status === 'needs_attention' ||
-								args.status === 'cancelled'
-							) {
-								const cancelledGroup = await runtime.taskGroupManager.cancel(group.id);
-								if (!cancelledGroup) {
+							if (args.status === 'cancelled') {
+								// Use runtime.cancelTask() which properly handles group termination
+								// AND cascades the cancellation to pending dependent tasks.
+								const cancelResult = await runtime.cancelTask(args.task_id);
+								if (!cancelResult.success) {
 									return jsonResult({
 										success: false,
-										error: `Failed to cancel active group for task ${args.task_id} — group may have been modified concurrently`,
+										error: `Failed to cancel task ${args.task_id} — runtime cancellation was unsuccessful`,
+									});
+								}
+								// cancelTask already set status and emitted task updates; emit UI updates and return
+								if (daemonHub) {
+									for (const cancelledId of cancelResult.cancelledTaskIds) {
+										const cancelledTask = await taskManager.getTask(cancelledId);
+										if (cancelledTask) {
+											void daemonHub.emit('room.task.update', {
+												sessionId: `room:${roomId}`,
+												roomId,
+												task: cancelledTask,
+											});
+										}
+									}
+								}
+								return jsonResult({
+									success: true,
+									message: `Task ${args.task_id} cancelled successfully.`,
+								});
+							} else if (args.status === 'completed' || args.status === 'needs_attention') {
+								// Terminate the group WITHOUT cascading cancellation to pending dependent tasks.
+								// Only cancellation of a dependency should affect dependents — completing or
+								// marking a task as needs_attention should leave dependents untouched.
+								const terminated = await runtime.terminateTaskGroup(args.task_id);
+								if (!terminated) {
+									return jsonResult({
+										success: false,
+										error: `Failed to terminate active group for task ${args.task_id} — group may have been modified concurrently`,
 									});
 								}
 							}
--- a/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
@@ -366,6 +366,7 @@ export function setupTaskHandlers(
 			status: TaskStatus;
 			result?: string;
 			error?: string;
+			mode?: 'manual' | 'runtime';
 		};
 
 		if (!params.roomId) {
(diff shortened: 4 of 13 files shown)


lsm (commented):

### Recommendation: APPROVE

This PR correctly addresses two issues: (1) fixing the cancellation cascade bug where completing a task incorrectly cancelled its dependents, and (2) introducing a dual-track state machine with `mode` parameter to allow manual UI transitions to bypass strict runtime validation.

### What looks good

- **Cancellation cascade fix** (`room-agent-tools.ts`): The differentiation between `cancelled` (uses `runtime.cancelTask()` with cascade) vs `completed`/`needs_attention` (uses `runtime.terminateTaskGroup()` without cascade) is clean and correct. The early return for the cancelled path properly prevents double status updates.

- **Dual validation** in both `task-handlers.ts` and `task-manager.ts`: Defense in depth -- if the RPC handler is ever bypassed, the manager still enforces the right rules.

- **Unarchiving support**: Clearing `archivedAt` when transitioning from archived, adding the `archivedAt` field to `UpdateTaskParams` and `TaskRepository` -- all properly wired together.

- **Test coverage**: Good coverage of manual mode at both the task-manager and RPC-handler levels. The dependent task cascade tests verify both directions (completing doesn't cascade, cancelling does).

- **Frontend**: SetStatusModal with destructive transition warnings is well-implemented. Error state handling, loading state, and cleanup on close are all present.

### Observations (P3)

1. **P3 - UI hides "Set Status" for archived tasks** (`TaskView.tsx`, around line 1261 in patched file): The button is hidden when `task.status === 'archived'`, and there's no other UI path to unarchive. The archive dialog explicitly says "The task cannot be reactivated after archiving." While the backend supports unarchiving via manual mode, the UI intentionally treats archival as permanent. This is consistent but worth noting -- if unarchiving is desired in the future, the backend is ready.

2. **P3 - Redundant validation**: `task-handlers.ts` (lines 388-396) validates the transition, then `task-manager.ts` (line 199) validates again. Both skip in manual mode. Not a problem -- this is defense in depth -- but the comment in the handler could note that the manager also validates.

3. **P3 - `isDestructiveTransition` coverage**: The function only covers three specific transitions (from archived, completed->pending, cancelled->completed). Other unusual manual transitions like `in_progress->draft` or `review->pending` aren't flagged as destructive. This seems like a reasonable starting point since those transitions are less risky, but expanding the set later may be warranted.

No P0, P1, or P2 issues found. The implementation is correct, well-tested, and fits existing patterns.

lsm (commented):

---

### Summary

The two core bug fixes are sound: (1) the cascade-cancellation bug in `room-agent-tools.ts` is correctly fixed by using `terminateTaskGroup` instead of the old cancel-group-for-all-terminal-states approach, and (2) the `mode` parameter in `setTaskStatus` correctly bypasses the state machine for manual UI transitions. However, there is a P1 behavioral defect where manual mode archiving is silently blocked by a downstream `archiveTask()` call that re-validates the transition, and a P2 gap in frontend test coverage for the new `SetStatusModal`.

---

### Issues

#### P1 — Manual mode cannot archive non-terminal tasks despite UI offering the option

**Files:** `packages/daemon/src/lib/rpc-handlers/task-handlers.ts` (lines 399–416) and `packages/daemon/src/lib/room/managers/task-manager.ts` (lines 450–455)

The `task.setStatus` handler correctly skips the transition check in manual mode (line 389). But when `params.status === 'archived'`, it immediately routes to the archive-specific path, which calls either `runtime.archiveTaskGroup(taskId)` or `taskManager.archiveTask(taskId)`. Both paths eventually invoke `taskManager.archiveTask()`, which contains its own hardcoded `isValidStatusTransition(task.status, 'archived')` check (task-manager.ts line 450).

Looking at `VALID_STATUS_TRANSITIONS`, only `completed`, `cancelled`, and `needs_attention` can transition to `archived`. Tasks in `pending`, `in_progress`, `review`, and `draft` states cannot. So even with `mode: 'manual'`, trying to archive an `in_progress` or `review` task will throw:
```
Cannot archive task in 'in_progress' status. Allowed transitions: ...
```

Meanwhile, the `SetStatusModal` in `TaskView.tsx` (line 618) shows every status except the current one, including `archived`, regardless of the current task state. A user choosing `archived` from an `in_progress` task will see a server-side error. The error message mentions allowed transitions but does not tell the user that manual mode is not honoured for archiving.

The fix needs to either: (a) bypass the validation inside `archiveTask` when called from a manual-mode context (pass a flag through), or (b) filter out `archived` from `SetStatusModal`'s available targets when the task is not in a terminal state (and document the intentional asymmetry).

---

#### P2 — No frontend tests for the new `SetStatusModal`

**File:** `packages/web/src/components/room/TaskView.test.tsx`

The PR added 3 lines to `TaskView.test.tsx` but the new `SetStatusModal` component (lines 552–726 in `TaskView.tsx`) has zero test coverage. Existing tests cover the `task.setStatus` RPC for complete, reactivate, and archive flows, but none exercise:
- Opening the modal via the "Set Status" button in `TaskInfoPanel`
- Selecting a new status and confirming
- The destructive warning being shown for relevant transitions
- The `mode: 'manual'` field being passed in the request

The 3 added lines in the test file appear to be stubs or placeholders but add no meaningful assertions.

---

#### P2 — Duplicate transition validation between RPC handler and `setTaskStatus`

**Files:** `packages/daemon/src/lib/rpc-handlers/task-handlers.ts` (lines 389–397) and `packages/daemon/src/lib/room/managers/task-manager.ts` (lines 199–204)

For `runtime` mode, the transition is validated twice: once in the RPC handler (task-handlers.ts line 390) and again in `setTaskStatus` (task-manager.ts line 199). The RPC handler re-implements the same `VALID_STATUS_TRANSITIONS` lookup that the manager already owns. This divergence means a future change to transition rules would need to be updated in two places. Consider removing the redundant check from the handler and letting the manager be the single authority.

---

#### P3 — `setTaskStatus` does not clear state fields when manually transitioning from `in_progress` or `review`

**File:** `packages/daemon/src/lib/room/managers/task-manager.ts` (lines 225–243)

The clear-fields logic only activates when the source status is `needs_attention`, `cancelled`, `completed`, or `archived`. A manual transition from `in_progress` → `pending` or `review` → `pending` will not clear `error`, `result`, or `progress` fields, potentially leaving stale data. The comment on line 224 mentions this covers `archived → active` but does not account for active → pending transitions in manual mode.

---

#### P3 — `isDestructiveTransition` coverage is incomplete

**File:** `packages/web/src/components/room/TaskView.tsx` (lines 585–593)

The function covers only 3 specific cases. Other manually-forced transitions that could be considered destructive (e.g. `review → pending`, which would discard in-progress work, or `in_progress → needs_attention`) produce no warning. This may be intentional, but it is worth a comment explaining the decision.

lsm (commented):

### Recommendation: APPROVE

All three issues from the previous review have been addressed correctly:

**P1 fix (archiveTask mode parameter)**: `archiveTask()` and `archiveTaskGroup()` now accept `{ mode?: 'runtime' | 'manual' }` and the RPC handler passes it through. The manual-mode bypass in `archiveTask` (task-manager.ts line 450) mirrors the bypass in `setTaskStatus` (line 199). This resolves the bug where manual archiving of non-terminal tasks would throw.

**P2 fix (duplicate validation removed)**: The redundant `VALID_STATUS_TRANSITIONS` check has been removed from the RPC handler. The manager is now the single validation authority. The handler's only remaining responsibility is orchestrating side effects (group termination, restart resets, event emission).

**P2 fix (SetStatusModal tests)**: Four new tests cover: opening the modal from the info panel, confirming with `mode: 'manual'`, verifying archived tasks hide the button, and verifying the destructive warning for `completed -> pending`.

1. **P3 -- Misleading test name** (`TaskView.test.tsx` line 2396): The test titled "shows destructive warning when transitioning from archived" does not actually test a destructive warning. It verifies that the Set Status button is hidden for archived tasks. The test name should reflect what it actually asserts (e.g., "hides Set Status button for archived tasks"). This is cosmetic and does not affect correctness.

2. **P3 -- `in_progress`/`review` to lower states don't clear stale fields** (`task-manager.ts` lines 225-243): The clear-fields logic only fires when the source status is `needs_attention`, `cancelled`, `completed`, or `archived`. A manual `in_progress -> pending` transition won't clear `progress` or `error` fields. This is a minor gap since active tasks rarely have error/result set, but could leave stale data in edge cases. Noted from the prior review and still applicable.

3. **P3 -- Security consideration on `mode: 'manual'`**: The `mode` parameter is accepted from the client without server-side authorization. Any WebSocket client can send `mode: 'manual'` to bypass transition validation. In the current single-user architecture this is fine, but if multi-user access is ever added, this would need an authorization check.

No P0, P1, or P2 issues remain. The implementation is correct, complete, and well-tested.

lsm (commented):

**Recommendation: APPROVE** — all three previously-raised issues (P1 + 2x P2) are resolved. The implementation is clean and tests are thorough. A few minor observations follow.

---

### Previously-Raised Issues — Resolution Status

**P1 — Manual mode archiving of non-terminal tasks**

Resolved. `archiveTask()` and `archiveTaskGroup()` now each accept `options?: { mode?: 'runtime' | 'manual' }`, the RPC handler constructs `modeOpts` and passes it through, and validation is skipped when `mode === 'manual'`. The `archivedAt` clearing in `setTaskStatus` (line 246 of `task-manager.ts`) is correctly gated behind the validation bypass — since `VALID_STATUS_TRANSITIONS['archived'] = []`, this code path is unreachable in runtime mode.

**P2 — Duplicate transition validation in RPC handler**

Resolved. The `VALID_STATUS_TRANSITIONS` import and the duplicate check block have been removed from `task-handlers.ts`. The mock factory in the test suite now re-implements that logic itself as the single-authority path, which is the right approach.

**P2 — SetStatusModal frontend tests**

Resolved. Four new tests were added in `TaskView.test.tsx`:
1. Opening the modal via the info panel trigger
2. Confirming with `mode: 'manual'` wired through to `task.setStatus`
3. Verifying archived tasks hide the Set Status button
4. Verifying the destructive warning text for `completed → pending`

---

### Remaining Observations (P3 / informational)

**P3 — Test assertions for `task.archive` handler use single-arg `toHaveBeenCalledWith`**

`packages/daemon/tests/unit/rpc-handlers/task-handlers.test.ts` lines 1467, 1477, 1487:
```ts
expect(runtime.archiveTaskGroup).toHaveBeenCalledWith('task-1');
```
The `task.archive` handler calls `runtime.archiveTaskGroup(params.taskId)` without a second argument, so this is correct and consistent. The analogous test in the `task.setStatus` suite (line 1114) properly asserts `toHaveBeenCalledWith('task-1', undefined)` because that handler passes `modeOpts`. No mismatch.

**P3 — Missing test: manual mode archiving a pending task via `task.setStatus` with runtime present**

The test suite does not cover the case where `{ status: 'archived', mode: 'manual' }` is sent for a pending/in-progress task with an active runtime. The production path (`modeOpts = { mode: 'manual' }` propagated to `archiveTaskGroup`) is correct, but a test verifying the `modeOpts` argument is forwarded to `archiveTaskGroup` would close this gap. Not blocking.

**P3 — `setTaskStatus` clear-fields condition is overly broad in manual mode**

In `task-manager.ts` lines 229-243, the manual-mode branch clears `error/result/progress` when destination is any of `pending | in_progress | review | completed`. This means a manual `pending → completed` (no prior result) will wipe `result` on the task — harmless since there is nothing to wipe, but the comment says "restarting from terminal state" when that's not always the case. Minor documentation inconsistency only.

---

Overall the implementation is correct, well-tested, and follows the existing architecture patterns. Happy to approve.

lsm (commented):

**Recommendation: APPROVE** — the latest CI fix is correct and closes the validation bypass gap.

---

### What the latest commit does

The third commit (`f7135122`) re-adds the handler-level `VALID_STATUS_TRANSITIONS` check at the top of `task.setStatus` (before the cancel/archive early-return paths). The core problem it fixes: both `runtime.cancelTask()` and the archive path use early returns that hand off to `cancelTaskCascade()` / `archiveTask()` respectively, and neither of those low-level paths called `setTaskStatus()` with its transition guard. Without the handler-level check, a client could send `{ status: 'cancelled' }` for an archived task and `runtime.cancelTask()` would call `cancelTaskCascade()` → `updateTaskStatus()` directly, bypassing all validation. The re-added check is placed before these early returns, so it fires first.

### Review of the fix

**Correctness:** The guard is correct. The condition is `params.mode !== 'manual'` and the `VALID_STATUS_TRANSITIONS` lookup is sourced from the manager's own map (single-source-of-truth for the table). `archived -> []` means every attempt to transition out of `archived` without `mode: 'manual'` will throw the expected message, which now matches the online test regex `/Invalid status transition from 'archived'/`.

**Defense-in-depth is intentional:** The handler check and the manager-level check in `setTaskStatus` (line 199 of `task-manager.ts`) are both deliberate. The comment in the handler explains this clearly: the early-return paths for cancel and archive bypass the manager's own validation, so the handler must catch those cases itself. This is not redundant complexity — it is the only correct way to guard those paths.

**Manual mode unaffected:** The new check gates on `params.mode !== 'manual'`, so all manual UI transitions continue to work. Confirmed by the existing manual-mode tests in the `describe('manual mode')` suite.

**Test coverage for the bug:** The specific regression path (`archived → cancelled` with an active runtime where `runtime.cancelTask()` would have been invoked) is not unit-tested with a real runtime mock. The existing test at line 936 of `task-handlers.test.ts` covers `archived → cancelled` but uses `makeNullRuntimeService()` (returns `null` for runtime), so it goes through the null-runtime path rather than the `runtime.cancelTask()` path. The online integration tests in `rpc-task-lifecycle.test.ts` (lines 423-437) do cover this behavior against a real server and match the corrected error format. This is a P3 observation only — the online test provides sufficient coverage of the real scenario.

**No new issues introduced:** The commit touches only `task-handlers.ts`. No logic changes elsewhere.

---

### Issues

**P3 — Unit test gap for archived → cancelled with active runtime:** `packages/daemon/tests/unit/rpc-handlers/task-handlers.test.ts` line 936 tests all transitions from `archived` using `makeNullRuntimeService()`. Adding a variant using `makeRuntimeServiceWithRuntimeCleanup()` for the `cancelled` target would directly document the bypass-via-cancelTask bug and guard against regression. Not blocking.

---

No P0, P1, or P2 issues. The fix is minimal, targeted, and well-reasoned. The implementation is correct.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The third commit correctly re-adds handler-level `VALID_STATUS_TRANSITIONS` validation at the `task.setStatus` RPC entry point (lines 391-398 of `task-handlers.ts`). This is necessary because the cancel and archive code paths use early returns that bypass the manager-level validation in `setTaskStatus()` and `archiveTask()`. Without the handler-level check, a non-manual call like `archived → cancelled` would succeed by hitting the `runtime.cancelTask()` early return before any manager validation runs.

The `mode: 'manual'` bypass is correctly preserved at both layers:
- Handler level (line 391): `if (params.mode !== 'manual')`
- Manager level `setTaskStatus` (line 199 of task-manager.ts): `if (options?.mode !== 'manual')`
- Manager level `archiveTask` (line 450 of task-manager.ts): `if (options?.mode !== 'manual')`

### Detailed Analysis

**Correctness** - The implementation is sound. The dual-layer validation is well-motivated by the early-return control flow in the handler. The commit message clearly explains why the handler-level check is needed.

**Test coverage** - Comprehensive:
- `task-manager.test.ts`: Tests manual mode bypass, runtime mode enforcement, archived unarchiving, field clearing on state reactivation, and cascade behavior (lines 1190-1317)
- `task-handlers.test.ts`: Tests handler-level validation for pending→completed, completed→pending, all transitions from archived; tests manual mode bypass, explicit runtime mode, default (no mode) runtime enforcement; tests `mode` is passed through to `setTaskStatus`; archive delegation tests (lines 797-1248)
- `TaskView.test.tsx`: Frontend tests for SetStatusModal, `mode: 'manual'` in RPC call, destructive transition warnings
- `room-agent-tools.test.ts`: Updated to reflect the cascading cancellation fix

**Frontend** - The `SetStatusModal` offers all statuses except current, shows destructive transition warnings for archived/completed/cancelled transitions, and consistently sends `mode: 'manual'` through `task.setStatus` RPC.

**Bug fix in room-agent-tools.ts** - Lines 387-398 correctly use `runtime.terminateTaskGroup()` for `completed`/`needs_attention` transitions instead of `cancelTask()`, preventing incorrect cascade cancellation of dependent tasks.

**CI failure** - The single CI failure is in `state-manager.test.ts:621/733` (`handler is not a function`), which is in a file not modified by this PR. This is a pre-existing issue unrelated to these changes.

### No issues found

The handler-level validation fix is correct and well-motivated. The error message format (`Invalid status transition from 'archived'...`) now matches the regex in the online integration test `rpc-task-lifecycle.test.ts`. The mode parameter flows correctly through all layers.

PR #706, opened 2026-03-22
fix: reuse existing dev proxy instance instead of failing on startup

lsm:
When a devproxy process from a previous session is still running,
`devproxy --detach` exits non-zero and tests would fail with
"Dev Proxy is already running (PID: XXXXX)".

Fix by proactively checking the configured port before attempting to
start devproxy.  If the port is already open, the controller adopts
the running instance as `external: true` and skips the start command.
`stop()` is a no-op for external instances so the pre-existing process
is left undisturbed after tests finish.

Also simplifies `acquireDevProxyLease` in daemon-server.ts to rely on
the controller's `isExternal` flag instead of ad-hoc error-message
parsing, and adds `DEV_PROXY_FREE_TO_START` guard in unit tests so
lifecycle tests skip gracefully when devproxy is already in use.

### DIFF
--- a/packages/daemon/tests/helpers/daemon-server.ts
+++ b/packages/daemon/tests/helpers/daemon-server.ts
@@ -20,7 +20,6 @@
 
 import { spawn, spawnSync } from 'child_process';
 import path from 'path';
-import net from 'net';
 import { MessageHub, WebSocketClientTransport } from '@neokai/shared';
 import { createDaemonApp, type DaemonAppContext } from '../../src/app';
 import { getConfig } from '../../src/config';
@@ -214,37 +213,6 @@ function installSharedDevProxyExitHook(): void {
 	});
 }
 
-async function isTcpPortOpen(port: number, host = '127.0.0.1', timeoutMs = 1000): Promise<boolean> {
-	return new Promise((resolve) => {
-		const socket = net.createConnection({ port, host });
-		const done = (result: boolean) => {
-			socket.removeAllListeners();
-			socket.destroy();
-			resolve(result);
-		};
-
-		socket.setTimeout(timeoutMs);
-		socket.once('connect', () => done(true));
-		socket.once('timeout', () => done(false));
-		socket.once('error', () => done(false));
-	});
-}
-
-async function waitForTcpPortOpen(
-	port: number,
-	host = '127.0.0.1',
-	timeoutMs = 4000
-): Promise<boolean> {
-	const start = Date.now();
-	while (Date.now() - start < timeoutMs) {
-		if (await isTcpPortOpen(port, host, 500)) {
-			return true;
-		}
-		await new Promise((resolve) => setTimeout(resolve, 100));
-	}
-	return false;
-}
-
 async function acquireDevProxyLease(
 	shouldUseDevProxy: boolean,
 	devProxyOptions?: DevProxyOptions
@@ -305,8 +273,13 @@ async function acquireDevProxyLease(
 	}
 
 	// For reuse mode, only register as shared controller when we own the proxy
-	// process.  External instances are not registered so the exit hook won't try
-	// to stop them when the test process exits.
+	// process.  External instances are intentionally not pooled in
+	// sharedDevProxyController: the exit hook (installSharedDevProxyExitHook)
+	// unconditionally runs `devproxy stop` on process exit, which would kill a
+	// proxy that belongs to another session.  With external instances each
+	// acquireDevProxyLease call creates a lightweight controller that performs a
+	// TCP probe on start and a no-op on stop — cheap enough that skipping the
+	// pool is acceptable.
 	if (reuse && !devProxy.isExternal) {
 		sharedDevProxyController = devProxy;
 		sharedDevProxyPort = devProxyPort;
--- a/packages/daemon/tests/helpers/daemon-server.ts
+++ b/packages/daemon/tests/helpers/daemon-server.ts
@@ -285,36 +285,29 @@ async function acquireDevProxyLease(
 		};
 	}
 
-	let devProxy: DevProxyController | null = createDevProxyController({
+	const devProxy: DevProxyController = createDevProxyController({
 		// Daemon helper explicitly sets env vars for both in-process and spawned modes.
 		// Keep proxy lifecycle independent from parent-process env mutation.
 		setEnvVars: false,
 		...devProxyOptions,
 	});
-	let startedByHelper = false;
 
 	try {
+		// start() will adopt an existing proxy (isExternal=true) rather than failing
+		// when a devproxy instance is already listening on the port.
 		await devProxy.start();
-		startedByHelper = true;
 	} catch (error) {
-		devProxy = null;
 		const errorMessage = error instanceof Error ? error.message : String(error);
-		// If another worker/process just started Dev Proxy, give it extra time to bind.
-		const readyTimeout = errorMessage.includes('already running') ? 12000 : 3000;
-		const proxyAlreadyRunning = await waitForTcpPortOpen(devProxyPort, '127.0.0.1', readyTimeout);
-		if (!proxyAlreadyRunning) {
-			throw new Error(
-				`Dev Proxy is required for this test run but is unavailable on ${devProxyBaseUrl}. ` +
-					`Startup error: ${errorMessage}`
-			);
-		}
-		console.warn(
-			`Warning: Could not start Dev Proxy controller, using existing proxy at ${devProxyBaseUrl}. ` +
+		throw new Error(
+			`Dev Proxy is required for this test run but failed to start on ${devProxyBaseUrl}. ` +
 				`Error: ${errorMessage}`
 		);
 	}
 
-	if (reuse && startedByHelper && devProxy) {
+	// For reuse mode, only register as shared controller when we own the proxy
+	// process.  External instances are not registered so the exit hook won't try
+	// to stop them when the test process exits.
+	if (reuse && !devProxy.isExternal) {
 		sharedDevProxyController = devProxy;
 		sharedDevProxyPort = devProxyPort;
 		sharedDevProxyRefCount = 1;
@@ -333,10 +326,9 @@ async function acquireDevProxyLease(
 	return {
 		controller: devProxy,
 		release: async () => {
-			if (devProxy && startedByHelper) {
-				await devProxy.stop();
-				devProxy.restoreEnv();
-			}
+			// stop() is a no-op for external instances, so it's always safe to call.
+			await devProxy.stop();
+			devProxy.restoreEnv();
 		},
 	};
 }
--- a/packages/daemon/tests/helpers/dev-proxy.ts
+++ b/packages/daemon/tests/helpers/dev-proxy.ts
@@ -95,13 +95,16 @@ export interface DevProxyOptions {
  */
 export interface DevProxyController {
 	/**
-	 * Start Dev Proxy process
+	 * Start Dev Proxy process.
+	 * If a proxy is already listening on the configured port, it is adopted as an
+	 * external instance (isExternal becomes true) and no new process is started.
 	 * @throws Error if proxy fails to start within timeout
 	 */
 	start(): Promise<void>;
 
 	/**
-	 * Stop Dev Proxy process gracefully
+	 * Stop Dev Proxy process gracefully.
+	 * Has no effect when the proxy was adopted as an external instance (isExternal === true).
 	 */
 	stop(): Promise<void>;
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured change that solves a real pain point — when a devproxy process from a previous session is still running, tests would fail instead of reusing it. The implementation is correct and the tests are thorough.

### What works well

1. **Two-layer adoption strategy in `dev-proxy.ts`**: First proactively checks the port before attempting `devproxy --detach`, then falls back to a port check after a non-zero exit. This handles both the common case (proxy already running) and the race condition (two processes starting simultaneously).

2. **Clean separation of concerns**: The `isExternal` flag lives on the controller itself, and `daemon-server.ts` simply reads it to decide whether to register as shared controller. The old ad-hoc error-message parsing in `acquireDevProxyLease` is gone.

3. **Exit hook safety**: External instances are never registered as `sharedDevProxyController` (line 310: `if (reuse && !devProxy.isExternal)`), so the exit hook at line 194-214 which unconditionally runs `spawnSync('devproxy', ['stop'])` will never fire for adopted instances. Correct.

4. **Test quality**: The 6 new `isExternal` tests use real TCP servers to simulate an existing proxy, covering adoption, stop-no-op behavior, state reset, env var handling, and restart cycles. Good use of `finally` blocks for cleanup.

5. **`DEV_PROXY_FREE_TO_START` guard**: Properly skips lifecycle tests when devproxy's single-instance constraint would cause them to fail, using `pgrep -x` to detect running instances.

### Minor observations (P3)

1. **Dead code**: `isTcpPortOpen()` and `waitForTcpPortOpen()` in `daemon-server.ts` (lines 217-246) are now unused after the `acquireDevProxyLease` simplification. They were only called from the old fallback path that was removed. Not blocking since knip should flag these, but worth a cleanup pass.

2. **`pgrep -x devproxy`** in the test file: This works on macOS and Linux but may not match on systems where the binary name differs from `devproxy` (e.g., if installed via a different package manager). This is a minor portability note — not an issue for the current CI environment.

Overall this is solid work. The change correctly handles the lifecycle edge cases and the tests verify the new behavior thoroughly.

lsm (commented):

**Recommendation: APPROVE**

The implementation is correct and clean. The core fix — proactively checking the port before calling `devproxy --detach` and adopting any pre-existing instance as `isExternal=true` — is the right approach. The `stop()` no-op for external instances prevents terminating a proxy owned by another session. The simplification in `acquireDevProxyLease` (removing ad-hoc error-message parsing) is a genuine improvement. Tests are meaningful and cover the happy path, stop behavior, env var propagation, and restart.

A few observations worth noting:

---

### P2: `checkProxyReady()` uses `require('net')` instead of a top-level import

`packages/daemon/tests/helpers/dev-proxy.ts` line 367:

```ts
const net = require('net');
```

The rest of the file uses ES module syntax. While Bun supports CJS interop, using a dynamic `require()` inside a closure is inconsistent and may confuse static analysis tools. The `net` module is already imported at the top of `daemon-server.ts` — it should be a top-level `import net from 'net';` in `dev-proxy.ts` too.

---

### P2: External-proxy not registered as shared when `reuse=true`

`packages/daemon/tests/helpers/daemon-server.ts` lines 310–324:

When `reuse=true` and `devProxy.isExternal=true`, the code skips registering `sharedDevProxyController`. This is intentional (don't let the exit hook kill an externally-owned proxy), but it means every subsequent `acquireDevProxyLease` call in the same process creates a fresh `DevProxyController` that also adopts the same external instance. There is no functional breakage (each adopt/stop cycle is a no-op on the real process), but it does bypass the entire reuse optimization — each test file that calls `createDaemonServer` will allocate a controller object, do a TCP probe, and release it.

This is acceptable as a follow-up but worth a comment explaining the known limitation: `// External proxies are not pooled in sharedDevProxyController because the exit hook must not stop them.`

---

### P3: Missing `controller.stop()` in one test's `try` block

`packages/daemon/tests/unit/helpers/dev-proxy.test.ts` lines 250–267 (`adopts an existing proxy on the port without starting a new process`):

The `try` block calls `controller.start()` but never calls `controller.stop()`. Because `stop()` is a no-op for external instances this causes no resource leak, and the controller is GC-collected, so there is no real harm — but it is stylistically inconsistent with the other tests in the same describe block that always stop. A reader may wonder if it was accidentally omitted.

---

### P3: `isDevProxyAlreadyRunning` using `pgrep -x devproxy`

`packages/daemon/tests/unit/helpers/dev-proxy.test.ts` lines 57–63:

On Linux, `pgrep -x` matches against the truncated comm field (15 chars), not the full executable name. `devproxy` is 8 chars so this is fine here. On macOS `-x` matches the exact process name. Acceptable for a test helper, but worth a brief comment noting the platform behaviour for future readers.

---

### P3: `when devproxy is not installed` test never exercises the installed+port-fails path

The test at line 386 early-returns when `DEV_PROXY_INSTALLED` is true. Since `DEV_PROXY_FREE_TO_START` was not used as the skip guard for this describe block, the test will execute on machines where devproxy is installed and an instance is already running — but it immediately returns. The code path where devproxy IS installed but fails (e.g., non-zero exit with port not open) is not tested. This is a pre-existing gap, not introduced by this PR, and low-priority to fix.

---

Overall the implementation is solid and addresses the reported failure mode correctly. The P2 items are quality suggestions and not blocking.

lsm (commented):

1. **P2 resolved -- dynamic require replaced with top-level import:** `packages/daemon/tests/helpers/dev-proxy.ts` line 48 now uses `import net from 'net'` at module scope. The `require('net')` inside the `checkProxyReady()` closure is removed.

2. **P2 resolved -- explanatory comment added:** `packages/daemon/tests/helpers/daemon-server.ts` lines 276-281 now clearly explain why external proxy instances bypass the `sharedDevProxyController` pool (exit hook would kill another session's proxy).

3. **P3 resolved -- dead code removed:** `isTcpPortOpen`, `waitForTcpPortOpen`, and the now-unused `import net from 'net'` are all removed from `daemon-server.ts`.

No new issues found. The code is clean and the follow-up commit is well-scoped.

**Recommendation: APPROVE**

lsm (commented):

**Re-review after follow-up fixes**
**Recommendation: APPROVE**

All three previously-flagged items have been correctly addressed. Verified findings below.

---

### P2 resolved: `require('net')` replaced with top-level ES module import

`packages/daemon/tests/helpers/dev-proxy.ts` line 48 now has `import net from 'net';` at the top of the file. The dynamic `require('net')` inside the `checkProxyReady()` closure is gone. Confirmed no `require(` calls remain in that file.

### P2 resolved: Explanatory comment added for external proxy pooling skip

`packages/daemon/tests/helpers/daemon-server.ts` lines 275-282 now contain a clear comment explaining why external instances bypass `sharedDevProxyController`:

> External instances are intentionally not pooled in sharedDevProxyController: the exit hook (installSharedDevProxyExitHook) unconditionally runs `devproxy stop` on process exit, which would kill a proxy that belongs to another session.

The trade-off (one TCP probe per test for external instances, cheap enough to skip the pool) is documented for future readers.

### P3 resolved: Dead code removed from `daemon-server.ts`

`isTcpPortOpen`, `waitForTcpPortOpen`, and the accompanying `import net from 'net'` have been removed. Confirmed no references remain in the helpers directory.

---

### Remaining minor observations (P3, non-blocking)

1. **`adopts an existing proxy` test (line 250-267) never calls `controller.stop()`** — the controller is started but the `try` block only asserts, then exits via `finally` which closes the TCP server. Since `stop()` is a no-op for external instances this causes no resource leak, but it is stylistically inconsistent with the other tests in the same `describe`. Left from the original PR; not introduced by these follow-up commits.

2. **`when devproxy is not installed` test (line 386)** — early-returns when devproxy is installed, which means the actual error-throw path is never exercised in a normal CI run. Pre-existing gap, not introduced here.

Both are minor and don't affect correctness. The fixes are clean and complete. No new issues introduced.

lsm (commented):

**Recommendation: APPROVE**

The cleanup commit correctly addresses the issues from the previous review round. The implementation is clean and complete.

### What changed in the cleanup commit

The `isExternal` unit tests in `dev-proxy.test.ts` now use a proper `describe('with a simulated pre-existing proxy')` block with `beforeEach`/`afterEach` lifecycle hooks instead of inline `try/finally` blocks. This is a meaningful improvement because:

1. **`beforeEach`** binds a fresh TCP server on a random port and creates a controller, ensuring test isolation.
2. **`afterEach`** guarantees teardown (stop controller, restore env, close TCP server) even when assertions throw mid-test -- the test framework handles this, not JS control flow.
3. The `setEnvVars=true` test correctly re-creates the controller within the test body (since it needs different options) while still relying on `afterEach` for cleanup.
4. The `originalBaseUrl` capture/restore in `afterEach` properly handles the `undefined` case with `delete`.

### Verification of the full PR

**`packages/daemon/tests/helpers/dev-proxy.ts` (controller implementation):**
- `import net from 'net'` is now a top-level ES module import (line 48), replacing the previous `require('net')` inside `checkProxyReady()`. Correct.
- The `isExternal` getter and `external` state flag are properly managed: set to `true` on adoption, reset to `false` on `stop()`.
- `stop()` correctly short-circuits for external instances (lines 514-519) without running `devproxy stop`.
- The race-condition fallback after a non-zero `devproxy --detach` exit code (lines 476-486) is a sensible defensive measure.

**`packages/daemon/tests/helpers/daemon-server.ts` (lease management):**
- Dead helpers `isTcpPortOpen` and `waitForTcpPortOpen` are removed along with the unused `import net` -- no dead code remains.
- `acquireDevProxyLease` now relies on `devProxy.isExternal` instead of parsing error messages, which is cleaner and less fragile.
- The comment at lines 275-282 clearly explains why external instances are not pooled in `sharedDevProxyController` (the exit hook would kill a proxy belonging to another session).
- The `release` callback unconditionally calls `stop()` + `restoreEnv()` (lines 301-305), which is safe because `stop()` is a no-op for external instances.

**`packages/daemon/tests/unit/helpers/dev-proxy.test.ts` (tests):**
- `DEV_PROXY_FREE_TO_START` guard using `pgrep -x devproxy` correctly prevents lifecycle tests from running when devproxy is already in use. Note: `pgrep -x` works on both macOS and Linux, matching the project's supported platforms.
- The nested `describe` block with `beforeEach`/`afterEach` covers: adoption, stop-no-op behavior, isExternal reset, env var setting, and restart-after-stop. Good coverage of the new behavior.

No issues found. The implementation is correct, complete, and well-documented.

lsm (commented):

---

**Recommendation: APPROVE**

The cleanup commit correctly restructures the `isExternal` unit tests. Here is my assessment:

### What changed

The five TCP-server-dependent tests in `packages/daemon/tests/unit/helpers/dev-proxy.test.ts` were previously written as self-contained tests each owning a `try/finally` block to bind and close their own TCP server. The cleanup refactors them into a nested `describe('with a simulated pre-existing proxy')` block with `beforeEach`/`afterEach` lifecycle hooks, so the test framework guarantees teardown regardless of whether a test throws.

### Correctness verified

**Lifecycle ordering in `afterEach`** is correct and intentional:
1. `controller.stop()` — no-op for external instances (sets `running=false`, `external=false`); called only when `isRunning()` is true to avoid double-stop concerns
2. `controller.restoreEnv()` — restores env from the controller's internal snapshot and clears it
3. Manual `ANTHROPIC_BASE_URL` restore using `originalBaseUrl` captured in `beforeEach` — redundant with `restoreEnv()` but idempotent and harmless
4. `closeTcpServer(tcpServer)` — closes last, so the port-probe inside `controller.stop()` above still sees the port open (matching the comment)

**`setEnvVars=true` test** reassigns `controller` in the test body. The `afterEach` picks up the reassigned reference correctly (captured by variable reference, not closure-over-initial-value), so `restoreEnv()` is called on the right controller that actually mutated `process.env`.

**`closeTcpServer` failure risk** is absent: no test closes the TCP server itself (they only interact with the controller which does TCP probing, not server closing), so the `server.close()` callback will never see "Server is not running".

**`isExternal is false on a fresh controller` test** remains outside the nested describe, which is correct — it needs no TCP server.

### Minor observation (no change needed)

The double env restoration in `afterEach` (once via `controller.restoreEnv()`, once manually with `originalBaseUrl`) is redundant but harmless. It makes the cleanup belt-and-suspenders, which is reasonable for test infrastructure. No action required.

### Overall

The refactor is clean, the teardown ordering is reasoned and documented, edge cases are handled, and the tests cover all five original scenarios without regressions. This is a straightforward, correct improvement.

PR #707, opened 2026-03-22
fix: fix TaskView not showing messages after daemon restart

lsm:
After daemon restart, review tasks showed "Waiting for agent activity…"
because:
1. useGroupMessages silently discarded liveQuery.subscribe failures with
   no retry, leaving the conversation in a permanent loading-but-empty state
2. task.getGroup also had no retry, so a transient RPC failure after restart
   would silently leave group=null for the entire session

Changes:
- useGroupMessages: add retry with backoff (2 retries, 500ms/1500ms delays)
  for liveQuery.subscribe failures; add isReconnecting state (true when
  disconnected but groupId is set) so TaskConversationRenderer can show
  "Reconnecting…" instead of the misleading "Waiting for agent activity…"
- TaskConversationRenderer: render "Reconnecting…" when isReconnecting
- TaskView: retry task.getGroup once after 1s on failure; show
  "Loading conversation history…" (not "No active agent group") when
  group=null for review-status tasks
- Tests: add retry/reconnect/isReconnecting unit tests for useGroupMessages;
  add group fetch retry and review empty-state tests for TaskView

### DIFF
--- a/packages/web/src/components/room/TaskConversationRenderer.test.tsx
+++ b/packages/web/src/components/room/TaskConversationRenderer.test.tsx
@@ -533,3 +533,87 @@ describe('TaskConversationRenderer — session question state props', () => {
 		});
 	});
 });
+
+describe('TaskConversationRenderer — isReconnecting state', () => {
+	beforeEach(() => {
+		mockRequest.mockReset();
+		mockRequest.mockImplementation(async (method: string, params: Record<string, unknown>) => {
+			if (method === 'liveQuery.subscribe') {
+				lastSubscriptionId = params.subscriptionId as string;
+				return { ok: true };
+			}
+			if (method === 'liveQuery.unsubscribe') return { ok: true };
+			if (method === 'session.get') return { session: null };
+			return {};
+		});
+		mockOnEvent.mockClear();
+		snapshotHandler = null;
+		deltaHandler = null;
+		lastSubscriptionId = null;
+		sessionStateHandlers.length = 0;
+		capturedSDKProps.length = 0;
+		resetSubscriptionCounterForTesting();
+		// Reset to connected by default.
+		mockIsConnected.value = true;
+	});
+
+	afterEach(() => {
+		cleanup();
+	});
+
+	it('renders "Reconnecting…" when WebSocket is disconnected and groupId is set', async () => {
+		// Simulate a WebSocket disconnect while the component is mounted with a group.
+		mockIsConnected.value = false;
+
+		const { container } = render(<TaskConversationRenderer groupId="group-1" />);
+
+		// isReconnecting = !isConnected && groupId !== null → true
+		// The component should show "Reconnecting…" not "Waiting for agent activity…"
+		await waitFor(() => {
+			expect(container.textContent).toContain('Reconnecting');
+		});
+		expect(container.textContent).not.toContain('Waiting for agent activity');
+	});
+
+	it('renders "Loading conversation…" when connected but snapshot has not arrived yet', async () => {
+		mockIsConnected.value = true;
+
+		const { container } = render(<TaskConversationRenderer groupId="group-1" />);
+
+		// isReconnecting = false (connected), isLoading = true (snapshot pending)
+		await waitFor(() => {
+			expect(container.textContent).toContain('Loading conversation');
+		});
+		expect(container.textContent).not.toContain('Reconnecting');
+	});
+
+	it('renders messages normally once snapshot arrives after reconnect', async () => {
+		// Start disconnected.
+		mockIsConnected.value = false;
+		const { container, rerender } = render(<TaskConversationRenderer groupId="group-1" />);
+
+		await waitFor(() => {
+			expect(container.textContent).toContain('Reconnecting');
+		});
+
+		// Reconnect.
+		mockIsConnected.value = true;
+		rerender(<TaskConversationRenderer groupId="group-1" />);
+
+		// Now in loading state (connected, waiting for snapshot).
+		await waitFor(() => {
+			expect(container.textContent).toContain('Loading conversation');
+		});
+
+		// Snapshot arrives.
+		await act(async () => {
+			fireSnapshot([makeRawMessage(1, 'assistant', 'uuid-reconnect-1')]);
+		});
+
+		await waitFor(() => {
+			expect(container.querySelector('[data-testid="msg-uuid-reconnect-1"]')).not.toBeNull();
+		});
+		expect(container.textContent).not.toContain('Reconnecting');
+		expect(container.textContent).not.toContain('Loading conversation');
+	});
+});
--- a/packages/web/src/components/room/TaskConversationRenderer.tsx
+++ b/packages/web/src/components/room/TaskConversationRenderer.tsx
@@ -159,7 +159,7 @@ export function TaskConversationRenderer({
 	const workerQuestionState = useSessionQuestionState(workerSessionId);
 
 	// Subscribe to group messages via LiveQuery (handles initial snapshot + live deltas)
-	const { messages: rawMessages, isLoading } = useGroupMessages(groupId);
+	const { messages: rawMessages, isLoading, isReconnecting } = useGroupMessages(groupId);
 
 	const messages = useMemo(
 		() => rawMessages.map(parseGroupMessage).filter((m): m is SDKMessage => m !== null),
@@ -256,6 +256,14 @@ export function TaskConversationRenderer({
 		}
 		return transitions;
 	}, [messages]);
+
+	if (isReconnecting) {
+		return (
+			<div class="flex-1 flex items-center justify-center">
+				<p class="text-gray-400 text-sm">Reconnecting…</p>
+			</div>
+		);
+	}
 
 	if (isLoading) {
 		return (
--- a/packages/web/src/components/room/TaskView.test.tsx
+++ b/packages/web/src/components/room/TaskView.test.tsx
@@ -2605,3 +2605,84 @@ describe('TaskView — goal badge', () => {
 		expect(badge?.getAttribute('title')).toBe('Mission: Tooltip Mission Title');
 	});
 });
+
+describe('TaskView — task.getGroup retry on failure', () => {
+	beforeEach(() => {
+		vi.useFakeTimers();
+		mockRequest.mockReset();
+		mockOnEvent.mockReset();
+		mockOnEvent.mockReturnValue(() => {});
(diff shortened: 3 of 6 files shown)


lsm (commented):

### Overall Assessment

The fix correctly addresses the reported issue — TaskView not showing messages after daemon restart. The approach of adding retry logic to both `liveQuery.subscribe` and `task.getGroup`, plus the `isReconnecting` UI state, is sound. The test coverage is good. However, there is one notable issue worth addressing.

### Issues

**[P2] `useGroupMessages` retry timers are not cleared on cleanup — potential stale retry fire**

In `packages/web/src/hooks/useGroupMessages.ts` lines 124-128, `setTimeout` is used for retry delays but the timer IDs are not tracked or cleared in the effect cleanup function (lines 139-154). When the effect cleans up (groupId change, disconnect, unmount), `activeSubIdRef.current` is set to `null`, and the guard at line 125 (`activeSubIdRef.current === subscriptionId`) prevents the retry callback from doing anything meaningful. So this is **not a functional bug** — the stale retry is effectively a no-op.

However, it is still a minor code quality concern: the timer continues to exist in the event loop after cleanup, and the guard-based cancellation pattern is less explicit than clearing the timer. If someone later removes or modifies the guard check, the timer could fire unexpectedly. Consider storing the timer ID and calling `clearTimeout` in the cleanup function for defense-in-depth. This is a suggestion, not a blocker.

The same pattern exists in `packages/web/src/components/room/TaskView.tsx` line 841 (`setTimeout` in `fetchGroup` retry), but there the `cancelled` and `seq === fetchGroupSeq` guards provide equivalent protection.

**[P3] `isReconnecting` is computed purely from render-time values, not from effect state**

In `packages/web/src/hooks/useGroupMessages.ts` line 157, `isReconnecting` is `!isConnected && groupId !== null`. This is derived directly from hook inputs rather than internal effect state. This means on the very first render with a groupId but before the WebSocket connects (`isConnected` starts false), `isReconnecting` will be `true` even though the user is not "reconnecting" — they are connecting for the first time. This is a minor UX inaccuracy: the user would briefly see "Reconnecting..." on initial page load before the WebSocket connects. In practice this window is likely very short, so it is a nit.

### What looks good

1. The `activeSubIdRef` stale-subscription guard is correctly applied at every decision point in the retry logic.
2. The `fetchGroupSeq` counter in TaskView correctly handles rapid re-fetches and prevents stale updates.
3. The `cancelled` flag in TaskView's effect cleanup properly guards against state updates after unmount.
4. Tests cover the key scenarios: retry success, retry exhaustion, cancellation on groupId change, and isReconnecting state transitions.
5. The UI messaging changes ("Loading conversation history..." for review tasks, "If this takes too long, try reloading the page.") are helpful and contextually appropriate.

### Recommendation: APPROVE

The implementation correctly fixes the reported issue with appropriate safeguards against stale state. The P2 and P3 items are suggestions for improvement, not blockers.

lsm (commented):

**Recommendation: APPROVE** (own PR, cannot use APPROVE event — leaving as COMMENT)

The implementation is solid and correctly addresses the two root causes: silent LiveQuery subscription failures and silent task.getGroup failures after daemon restart. The logic is sound, the stale-event guards are preserved, and the guard against post-unmount state updates is thorough.

---

### Findings

#### P2 — Missing UX test coverage for `isReconnecting` in `TaskConversationRenderer`

The `TaskConversationRenderer` now renders a dedicated "Reconnecting…" state when `isReconnecting` is true (line 260–266 of `TaskConversationRenderer.tsx`). The unit tests in `useGroupMessages.test.ts` confirm the flag value is correct, but there is no test that renders `TaskConversationRenderer` itself and asserts that the string "Reconnecting…" appears in the DOM when `useGroupMessages` returns `isReconnecting: true`. The rendering path is exercised only transitively through the mock.

This is low-risk since the code path is trivial (a direct branch on `isReconnecting`), but a snapshot or text-content assertion in a `TaskConversationRenderer` unit test would eliminate the gap.

#### P2 — `fetchGroup` retry only fires on first load, not on reconnect in `TaskView`

`fetchGroup` adds a single 1-second retry only when the initial RPC throws (returns `null`). However, the `useEffect` that drives `fetchGroup` (lines 818–907 of `TaskView.tsx`) does **not** include `isConnected` in its dependency array — it depends only on `roomId` and `taskId`. This means that after a WebSocket reconnect, TaskView does **not** re-issue `task.getGroup` even though `useGroupMessages` will re-subscribe to LiveQuery.

Concretely: if the daemon restarts while TaskView is mounted and the first `task.getGroup` call had already succeeded (returning `group: null` for a task not yet assigned), the group will remain `null` after reconnect. The new retry logic inside `fetchGroup` would only help if the very first RPC call throws; a `null` response is not retried at all.

This is a pre-existing partial gap, but the retry PR makes users expect it to work after reconnect, and the mechanism only covers the throw case during initial mount.

#### P3 — `@ts-nocheck` pragma in new test file

`packages/web/src/hooks/__tests__/useGroupMessages.test.ts` has `// @ts-nocheck` at line 1. This disables TypeScript checking for the entire file. The reason is not documented. Given the codebase has a strict typecheck CI step, losing type safety in a test file that imports and exercises the hook interface is undesirable — a typing error in mock wiring would be silently ignored.

#### P3 — Retry delays are module-level magic numbers

`MAX_RETRIES = 2` and `RETRY_DELAYS_MS = [500, 1500]` in `useGroupMessages.ts` are bare constants inside the function body. This is fine for now, but they are not exported or documented as intentional tuning parameters, and the delay array is implicitly tied to MAX_RETRIES (if someone sets MAX_RETRIES = 3 without adding a third delay, `RETRY_DELAYS_MS[attempt]` returns `undefined`, which `setTimeout` silently coerces to 0). A runtime guard or a comment warning about this coupling would prevent a future bug.

---

### Summary

The core fix is correct and well-tested for the hook layer. Two gaps worth noting: (1) `TaskView.fetchGroup` is not re-triggered on reconnect, only on initial mount failures — so a daemon restart that happens while the view is already mounted with `group: null` will not self-heal without a page reload; (2) the `@ts-nocheck` pragma in the new test file trades type safety for convenience. Neither is blocking, but (1) in particular may cause a follow-up bug report.

lsm (commented):

### Re-review: All P2 items from prior review are addressed

The follow-up commit `ed5f0e730` correctly resolves all three items raised in the previous review round:

1. **clearTimeout for retry timers in `useGroupMessages`** — `retryTimer` is now tracked and cleared in the effect cleanup function (lines 147-150 of `useGroupMessages.ts`), and `RETRY_DELAYS_MS` is typed as `[number, number]` tuple to enforce the coupling with `MAX_RETRIES`. The comment at line 113-114 also documents the invariant clearly.

2. **`isConnected` added to TaskView effect dependency array** — `useMessageHub()` now destructures `isConnected` (line 729 of `TaskView.tsx`) and it is included in the dependency array at line 907. This means `fetchGroup` will re-run on WebSocket reconnect, closing the gap where a daemon restart while TaskView was mounted would leave `group` stale.

3. **TaskConversationRenderer rendering tests for `isReconnecting` state** — Three new tests at lines 537-619 of `TaskConversationRenderer.test.tsx` verify: (a) "Reconnecting..." appears when disconnected with a groupId, (b) "Loading conversation..." appears when connected but awaiting snapshot, (c) messages render normally after reconnect and snapshot arrival. These cover the full reconnect lifecycle through the actual rendered component.

### Additional verification

- The TaskView test mock at line 34 of `TaskView.test.tsx` correctly includes `isConnected: true` to match the updated `useMessageHub` destructuring.
- The cleanup ordering in `useGroupMessages.ts` is correct: `clearTimeout` runs before `activeSubIdRef.current = null`, preventing the race where a timer callback could observe a matching subscription ID after cleanup.

### Recommendation: APPROVE

All feedback has been addressed cleanly. No new issues introduced.

lsm (commented):

**Recommendation: APPROVE** — all three P2 issues from the prior review have been correctly addressed. The implementation is sound.

---

### Review of feedback-addressed commit (ed5f0e730)

#### 1. clearTimeout for retry timers — LGTM

`packages/web/src/hooks/useGroupMessages.ts`

The `retryTimer` variable is introduced at effect scope, assigned in `subscribeWithRetry` when a timeout fires, and cleared in the effect cleanup before the subscription ID is nulled. The cleanup order is correct: timer is cancelled first so the callback can never observe a live `activeSubIdRef.current` after teardown. The tuple type annotation `[number, number]` will produce a TypeScript error if `MAX_RETRIES` diverges from the array length, which is the intended compile-time safety net.

One minor note: the timer is cleared on natural-success paths too (because `retryTimer` will be `null` when the subscribe succeeds), so no dead-timer leaks exist. The code is correct.

#### 2. isConnected in TaskView effect dependency array — LGTM

`packages/web/src/components/room/TaskView.tsx:907`

Adding `isConnected` to the `[roomId, taskId, isConnected]` dependency array means the effect re-runs on reconnect. On reconnect `load()` is called again, which re-fetches both the task and the group. Because `loading` is not reset to `true` at the top of the re-run, there is no full-page "Loading task..." spinner flash on reconnect — the UI remains stable. The `cancelled` flag is properly re-scoped to each effect run, and `leaveRoom` in cleanup correctly drops the room subscription so the re-run's `joinRoom` re-establishes it cleanly. This is the right fix for the self-heal gap.

There is a cosmetic side-effect worth noting (not blocking): on disconnect (`isConnected` flips `true → false`), cleanup fires `leaveRoom(channel)`, and on reconnect the effect re-subscribes via `joinRoom`. This means `room.task.update` events will be missed during the outage window, but those are handled by the re-fetch on reconnect, so no message loss occurs. This is the expected and correct behaviour.

#### 3. TaskConversationRenderer isReconnecting tests — LGTM

`packages/web/src/components/room/TaskConversationRenderer.test.tsx:537-619`

Three tests cover the three relevant states:
- Disconnected + groupId set → shows "Reconnecting..."
- Connected + snapshot not yet arrived → shows "Loading conversation..."
- Disconnected → reconnect → snapshot arrives → shows message

All three tests exercise the correct code path through `useGroupMessages`'s `isReconnecting: !isConnected && groupId !== null` return value. The mock uses `mockIsConnected.value` read at render time (plain object, not a signal), so `rerender()` is correctly used to trigger the re-render after changing the value. This is the right approach for Preact testing without over-mocking.

The `beforeEach` resets both the subscription counter and `mockIsConnected.value = true`, which prevents state bleed between tests. Clean.

---

### No remaining issues

All previously raised P2 issues are resolved. No new issues were introduced in the feedback commit. The implementation is complete and correct.

PR #709, opened 2026-03-22
test: add reconnect resync guard test for review-status toast (Task 3.4)

lsm:
Add explicit test verifying that snapshot re-hydration during reconnect does
not fire review toasts, even when a task transitioned to review while
disconnected. The implementation in room-store.ts (from Task 3.3) already
handles this correctly — snapshot writes replace tasks.value so subsequent
deltas see prevTask.status === 'review' and skip the toast.

### DIFF
--- a/packages/web/src/lib/__tests__/room-store-review.test.ts
+++ b/packages/web/src/lib/__tests__/room-store-review.test.ts
@@ -192,6 +192,37 @@ describe('RoomStore — review toast notification (via liveQuery.delta)', () => 
 
 		expect(toastsSignal.value.length).toBe(0);
 	});
+
+	it('does NOT fire a toast during reconnect resync (snapshot re-hydrates to review, delta keeps same status)', async () => {
+		await roomStore.select(ROOM_ID);
+
+		// Initial snapshot: task is in_progress
+		fireEvent('liveQuery.snapshot', {
+			subscriptionId: TASKS_SUB_ID,
+			rows: [makeTask('t6', 'in_progress', 'Reconnect Task')],
+			version: 1,
+		});
+
+		// Simulate reconnect: server sends a fresh snapshot where task is already in review
+		// (it transitioned while we were disconnected — the snapshot is the resync)
+		fireEvent('liveQuery.snapshot', {
+			subscriptionId: TASKS_SUB_ID,
+			rows: [makeTask('t6', 'review', 'Reconnect Task')],
+			version: 3,
+		});
+
+		// No toast from snapshot — snapshot is pure data hydration
+		expect(toastsSignal.value.length).toBe(0);
+
+		// A subsequent delta that keeps the task in review also fires no toast
+		fireEvent('liveQuery.delta', {
+			subscriptionId: TASKS_SUB_ID,
+			updated: [makeTask('t6', 'review', 'Reconnect Task')],
+			version: 4,
+		});
+
+		expect(toastsSignal.value.length).toBe(0);
+	});
 });
 
 describe('RoomStore — reviewTaskCount computed signal', () => {


lsm (commented):

**Recommendation: APPROVE** — The implementation is correct and the new test is well-targeted. Only one minor issue worth noting.

---

### Summary

This PR adds a single regression-guard test to `packages/web/src/lib/__tests__/room-store-review.test.ts` that covers the reconnect resync scenario for the review-status toast. No production code was changed. All 9 tests pass.

---

### What was reviewed

The implementation in `room-store.ts` (lines 327-353) uses a `prevTask` lookup against the current snapshot before mutating state. Because `liveQuery.snapshot` unconditionally replaces `tasks.value`, any task that arrives in `review` status during a reconnect resync snapshot will set `prevTask.status === 'review'` before any subsequent delta arrives. The delta handler then correctly finds `prevTask.status === 'review'` and skips the toast. This is the exact scenario the new test exercises.

---

### Issues

**P3 — Minor: `@ts-nocheck` suppresses type safety across the entire test file**

`packages/web/src/lib/__tests__/room-store-review.test.ts` line 1 carries `// @ts-nocheck`. This blanket suppression means the TypeScript compiler ignores all type errors in the file. Given that the `makeTask` helper returns a plain object literal that may diverge silently from `TaskSummary` as the type evolves, this is a latent maintenance risk. The suppression appears to pre-date this PR (it was already present), so this is not a blocker for the current change, but it is worth removing it and addressing any resulting type errors at a convenient time.

**P3 — Minor: version gap in the new test scenario is cosmetically misleading**

The reconnect scenario uses versions 1 → 3 → 4, with version 2 intentionally skipped to imply it was received while disconnected. This is fine as documentation intent, but the delta handler in `room-store.ts` does not actually enforce version ordering — it processes every event it receives regardless of version number. The version values in the test have no behavioral effect and readers may infer ordering enforcement that does not exist. A brief inline comment clarifying this ("version gap is intentional — mimics missed events; the store does not enforce ordering") would prevent confusion.

---

### Correctness

The guard logic in the delta handler is correct:

1. Snapshot handler replaces `tasks.value` in full — no toast path, pure hydration.
2. Delta handler checks `prevTask` from the pre-mutation snapshot of `current`. Because `removed` is processed first, then `updated` (with the toast check), a task that was already in `review` at the start of the handler will have `prevTask.status === 'review'` and will not trigger a toast.
3. The `added` path appends new tasks only; new tasks arriving directly in `review` via `added` (not `updated`) would not trigger a toast because there is no prevTask lookup for additions. This is intentional and consistent with the "hydration guard" rationale documented in the existing tests.

The test exercises exactly the stated scenario and the assertions are tight (zero toasts at both checkpoints).

---

### Test coverage assessment

The five existing tests cover: hydration guard (unknown task in delta), normal in_progress → review transition, review → review no-op, non-review status update, and wrong subscriptionId. The new test adds the reconnect snapshot re-hydration path. This is comprehensive for the current feature surface.

lsm (commented):

### Summary

This PR adds a single test case to `room-store-review.test.ts` covering the reconnect resync guard scenario. The test verifies that when a task transitions to `review` status during a disconnect, the reconnect snapshot re-hydrates `tasks.value` with the already-review task, so any subsequent delta for that task sees `prevTask.status === 'review'` and correctly suppresses the toast.

### Analysis

**Correctness**: The test accurately models the reconnect flow: initial snapshot (in_progress) -> resync snapshot (review) -> delta (review). The assertion that no toast fires at any point is correct given the implementation in `room-store.ts` lines 338-344, where the toast only fires when `prevTask.status !== 'review'`.

**Implementation review**: The existing implementation in `room-store.ts` handles this correctly because:
1. Snapshot handler (line 321) replaces `tasks.value` wholesale, so after resync the task is already in `review` status.
2. Delta handler (line 341) checks `prevTask.status !== 'review'` against the current `tasks.value`, which now reflects the snapshot.
3. Snapshots themselves never trigger toasts -- only deltas do.

**Test quality**: The test is well-structured, uses clear comments explaining each step, and follows the existing test patterns in the file. The use of `t6` as the task ID avoids collisions with other tests.

No issues found. This is a clean, focused test addition that fills a meaningful gap in coverage.

**Recommendation: APPROVE**

PR #711, opened 2026-03-22
feat: create useRoomLiveQuery hook and integrate with Room.tsx (Task 3.5)

lsm:
## Summary

- Extracts `tasks.byRoom` and `goals.byRoom` LiveQuery subscriptions from `roomStore.startSubscriptions()` into new public `subscribeRoom()`/`unsubscribeRoom()` methods, so `select()` no longer calls `liveQuery.subscribe` internally
- Creates `useRoomLiveQuery` hook as a lifecycle adapter: calls `subscribeRoom` on mount, `unsubscribeRoom` on unmount, and correctly handles `roomId` changes via `useEffect` dependency array
- Adds race-condition guard: if `unsubscribeRoom` is called before the async hub resolves (rapid room switching or unmount), the subscription is aborted cleanly
- Mounts `useRoomLiveQuery` unconditionally inside `Room.tsx` alongside the existing `select()` effect
- No double-subscription: `liveQueryActive` set prevents calling `subscribeRoom` twice for the same room
- 5 new Vitest tests for hook lifecycle (mount, unmount, roomId change, no-resubscribe on same roomId, unmount after room change)
- Updated existing room-store tests to call `subscribeRoom`/`unsubscribeRoom` explicitly since they no longer happen inside `select()`

## Test plan

- [x] `useRoomLiveQuery` Vitest tests: 5 tests covering all lifecycle scenarios
- [x] Full web test suite passes (4997 tests)
- [x] Lint + format checks pass
- [x] Knip dead code detection clean

### DIFF
--- a/packages/web/src/hooks/__tests__/index.test.ts
+++ b/packages/web/src/hooks/__tests__/index.test.ts
@@ -15,6 +15,7 @@ import {
 	useInterrupt,
 	useFileAttachments,
 	useGroupMessages,
+	useRoomLiveQuery,
 } from '../index.ts';
 
 describe('Hooks Index', () => {
@@ -62,6 +63,11 @@ describe('Hooks Index', () => {
 		it('should export useGroupMessages', () => {
 			expect(useGroupMessages).toBeDefined();
 			expect(typeof useGroupMessages).toBe('function');
+		});
+
+		it('should export useRoomLiveQuery', () => {
+			expect(useRoomLiveQuery).toBeDefined();
+			expect(typeof useRoomLiveQuery).toBe('function');
 		});
 	});
 });
--- a/packages/web/src/hooks/__tests__/useRoomLiveQuery.test.ts
+++ b/packages/web/src/hooks/__tests__/useRoomLiveQuery.test.ts
@@ -1,0 +1,124 @@
+// @ts-nocheck
+/**
+ * Tests for useRoomLiveQuery Hook
+ *
+ * Verifies that the hook correctly manages the LiveQuery subscription
+ * lifecycle by delegating to roomStore.subscribeRoom / unsubscribeRoom:
+ *
+ * - Calls subscribeRoom(roomId) on mount
+ * - Calls unsubscribeRoom(oldRoomId) then subscribeRoom(newRoomId) on roomId change
+ * - Calls unsubscribeRoom(roomId) on unmount
+ * - Does NOT double-subscribe if called with the same roomId
+ */
+
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook } from '@testing-library/preact';
+
+// ---------------------------------------------------------------------------
+// Hoisted mocks (must not import anything)
+// ---------------------------------------------------------------------------
+
+const { mockSubscribeRoom, mockUnsubscribeRoom } = vi.hoisted(() => ({
+	mockSubscribeRoom: vi.fn().mockResolvedValue(undefined),
+	mockUnsubscribeRoom: vi.fn(),
+}));
+
+// Mock roomStore so we control subscribeRoom and unsubscribeRoom directly.
+vi.mock('../../lib/room-store', () => ({
+	roomStore: {
+		subscribeRoom: mockSubscribeRoom,
+		unsubscribeRoom: mockUnsubscribeRoom,
+	},
+}));
+
+// ---------------------------------------------------------------------------
+// Imports (after mocks)
+// ---------------------------------------------------------------------------
+
+import { useRoomLiveQuery } from '../useRoomLiveQuery';
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe('useRoomLiveQuery', () => {
+	beforeEach(() => {
+		mockSubscribeRoom.mockClear();
+		mockUnsubscribeRoom.mockClear();
+	});
+
+	it('calls subscribeRoom(roomId) on mount', () => {
+		renderHook(() => useRoomLiveQuery('room-1'));
+
+		expect(mockSubscribeRoom).toHaveBeenCalledTimes(1);
+		expect(mockSubscribeRoom).toHaveBeenCalledWith('room-1');
+	});
+
+	it('calls unsubscribeRoom(roomId) on unmount', () => {
+		const { unmount } = renderHook(() => useRoomLiveQuery('room-1'));
+
+		expect(mockUnsubscribeRoom).not.toHaveBeenCalled();
+
+		unmount();
+
+		expect(mockUnsubscribeRoom).toHaveBeenCalledTimes(1);
+		expect(mockUnsubscribeRoom).toHaveBeenCalledWith('room-1');
+	});
+
+	it('calls unsubscribeRoom(oldRoomId) then subscribeRoom(newRoomId) on roomId change', () => {
+		const { rerender } = renderHook(({ roomId }) => useRoomLiveQuery(roomId), {
+			initialProps: { roomId: 'room-1' },
+		});
+
+		expect(mockSubscribeRoom).toHaveBeenCalledTimes(1);
+		expect(mockSubscribeRoom).toHaveBeenCalledWith('room-1');
+		expect(mockUnsubscribeRoom).not.toHaveBeenCalled();
+
+		mockSubscribeRoom.mockClear();
+		mockUnsubscribeRoom.mockClear();
+
+		rerender({ roomId: 'room-2' });
+
+		// unsubscribeRoom for old room before subscribeRoom for new room
+		expect(mockUnsubscribeRoom).toHaveBeenCalledTimes(1);
+		expect(mockUnsubscribeRoom).toHaveBeenCalledWith('room-1');
+		expect(mockSubscribeRoom).toHaveBeenCalledTimes(1);
+		expect(mockSubscribeRoom).toHaveBeenCalledWith('room-2');
+
+		// Verify unsubscribe was called before subscribe by checking call order
+		const unsubCallOrder = mockUnsubscribeRoom.mock.invocationCallOrder[0];
+		const subCallOrder = mockSubscribeRoom.mock.invocationCallOrder[0];
+		expect(unsubCallOrder).toBeLessThan(subCallOrder);
+	});
+
+	it('does not re-subscribe when roomId is unchanged', () => {
+		const { rerender } = renderHook(({ roomId }) => useRoomLiveQuery(roomId), {
+			initialProps: { roomId: 'room-1' },
+		});
+
+		expect(mockSubscribeRoom).toHaveBeenCalledTimes(1);
+		mockSubscribeRoom.mockClear();
+		mockUnsubscribeRoom.mockClear();
+
+		// Rerender with the same roomId
+		rerender({ roomId: 'room-1' });
+
+		expect(mockSubscribeRoom).not.toHaveBeenCalled();
+		expect(mockUnsubscribeRoom).not.toHaveBeenCalled();
+	});
+
+	it('calls unsubscribeRoom with the last active roomId on unmount after room change', () => {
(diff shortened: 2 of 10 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured extraction of LiveQuery subscription lifecycle management into a dedicated hook. The separation of concerns is correct and the implementation handles edge cases properly.

### What was done

1. **`useRoomLiveQuery` hook** (`packages/web/src/hooks/useRoomLiveQuery.ts`): Minimal lifecycle adapter — calls `subscribeRoom` on mount, `unsubscribeRoom` on cleanup, with `[roomId]` dependency. Exactly what it should be.

2. **`subscribeRoom`/`unsubscribeRoom` on RoomStore** (`packages/web/src/lib/room-store.ts`): Moved LiveQuery subscription logic out of `startSubscriptions()` into these public methods. The `liveQueryActive` Set provides double-subscribe protection and the async race guards (lines 317-318, 324, 378, 457) correctly handle the case where `unsubscribeRoom` is called before the async `getHub()` resolves.

3. **`stopSubscriptions()` only handles channel subscriptions** (line 672-681): It touches `cleanupFunctions` only, not `liveQueryCleanups`/`liveQueryActive`. The separation is clean.

4. **Integration in Room.tsx** (line 38): Hook is called before the `select()` effect, which is the right ordering.

5. **Tests**: 5 hook lifecycle tests covering mount, unmount, roomId change (with call-order verification), same-roomId rerender, and unmount-after-change. Existing live query tests updated to explicitly call `subscribeRoom`/`unsubscribeRoom`.

### Potential concerns reviewed (all clear)

- **Effect ordering on roomId change**: Both `useRoomLiveQuery` and the `select()` effect share `[roomId]` dependency. Preact runs cleanups and effects in declaration order. `unsubscribeRoom(old)` runs before `select(null)`, and `subscribeRoom(new)` runs before `select(new)`. Since LiveQuery and channel subscriptions are independent code paths, this ordering is safe.

- **Unmount before `subscribeRoom` completes**: The async guard at line 324 (`if (!this.liveQueryActive.has(roomId)) return`) correctly aborts if `unsubscribeRoom` was called before the hub became available. The mid-flight guards at lines 378-388 and 457-467 handle the case where unsubscribe happens between the two `liveQuery.subscribe` calls.

- **Double-subscribe prevention**: `liveQueryActive.has(roomId)` guard at line 317 prevents duplicate subscriptions.

- **Error handling**: The catch block at lines 496-500 cleans up both `liveQueryActive` and `liveQueryCleanups` on failure.

### No issues found

The implementation is correct, complete, and well-tested. No bugs, missing edge cases, or architectural concerns.

lsm (commented):

**Recommendation: APPROVE** — The implementation is correct, well-structured, and all acceptance criteria are met. A few observations worth noting below.

---

### Summary

The PR cleanly extracts the LiveQuery subscription lifecycle out of `select()` into public `subscribeRoom`/`unsubscribeRoom` methods, and wraps them in a thin `useRoomLiveQuery` hook that correctly handles mount, unmount, and roomId-change cases via React's `useEffect` dependency array. The race-condition guard (checking `liveQueryActive.has(roomId)` after each `await`) is sound.

---

### What Was Verified

- **`select()` no longer calls `liveQuery.subscribe`** — confirmed in `startSubscriptions()` at line 531, there are no `liveQuery.subscribe` calls. Tasks and goals come exclusively from the new `subscribeRoom` path.
- **No double-subscription** — the `liveQueryActive` set guard at line 317 prevents calling `subscribeRoom` twice for the same roomId.
- **Clean unsubscribe on unmount** — `unsubscribeRoom` is called in the `useEffect` cleanup at line 40 of `useRoomLiveQuery.ts`, and also calls the server-side `liveQuery.unsubscribe` via the cleanup closures.
- **roomId change** — `useEffect([roomId])` triggers cleanup (unsubscribe old) then effect (subscribe new) automatically.
- **5 hook lifecycle tests** — cover mount, unmount, roomId change, no-resubscribe on same roomId, and unmount after room change. The call-order assertion (line 89-91) is a nice touch.
- **4 updated test files** — all correctly call `subscribeRoom`/`unsubscribeRoom` explicitly in `beforeEach`/`afterEach` since `select()` no longer manages them.

---

### Issues Found

**P2 — Subtle race: `liveQueryCleanups` set before second `await`, then unsubscribeRoom may run stale cleanups**

In `subscribeRoom` (`packages/web/src/lib/room-store.ts`, lines 326-388), `this.liveQueryCleanups.set(roomId, cleanups)` is written at line 327, before the first `await hub.request('liveQuery.subscribe', ...)` at line 371. If `unsubscribeRoom` fires in this window, it will:

1. Remove `roomId` from `liveQueryActive`
2. Call all cleanups currently in the array (correct)
3. Delete from `liveQueryCleanups` (correct)

Then back in `subscribeRoom`, the guard at line 378 catches this and enters the cleanup block — which runs the same `cleanups` array again. Since the cleanups are `hub.onEvent` unsubscribers and `liveQuery.unsubscribe` requests, calling them a second time is mostly harmless (unsubscribers are idempotent in the mock, and the server-side unsubscribe is a no-op if already removed). This is acceptable in practice, but worth noting as a latent double-cleanup path.

**P2 — `subscribeRoom` and `select()` effects fire concurrently with no ordering guarantee**

In `Room.tsx` lines 38-49, `useRoomLiveQuery(roomId)` and the `useEffect(() => roomStore.select(roomId), [roomId])` both run on mount. Preact/React runs effects in declaration order, so `subscribeRoom` fires before `select()`. This means `subscribeRoom` calls `connectionManager.getHub()` (async) and begins registering handlers on the hub while `select()` also calls `connectionManager.getHub()` concurrently and may clear signal state (tasks, goals) midway through the subscribe flow.

In practice this is benign: the LiveQuery snapshot from the server arrives after both subscriptions complete and overwrites the cleared signals. But the current code has no explicit contract that `subscribeRoom` should be called only after `select()` has settled. A comment near the hook usage in `Room.tsx` documenting this intentional ordering would help maintainers.

**P3 — `goalsLoading` is set to `true` inside `subscribeRoom` (line 449) but is cleared only in the snapshot handler**

If the connection is lost after `subscribeRoom` completes but before the snapshot arrives, `goalsLoading` stays `true` indefinitely unless `unsubscribeRoom` or a new `select()` clears it. The reconnect handler at line 473 sets `goalsLoading = true` again (correct), but only the snapshot handler at line 423 sets it to `false`. This is an existing pattern and not introduced by this PR, but it means a failed reconnect can leave the goals panel in a permanent spinner state. Not a blocker.

**P3 — No test for the `subscribeRoom` error path**

`packages/web/src/lib/__tests__/room-store-tasks-live-query.test.ts` and `room-store-goals-live-query.test.ts` do not cover what happens when `connectionManager.getHub()` rejects inside `subscribeRoom`. The catch block at line 496-500 deletes the active and cleanup entries and logs an error. A test verifying that `liveQueryActive` is cleaned up on hub rejection would prevent silent regressions.

---

### Minor Observations

- The `@ts-nocheck` at the top of `useRoomLiveQuery.test.ts` is unnecessary — the file has no TS errors and the other typed test files (`room-store-tasks-live-query.test.ts`, `room-store-goals-live-query.test.ts`) do not use it. Consider removing it.
- The hooks barrel at `packages/web/src/hooks/index.ts` correctly exports `useRoomLiveQuery`, and the export test in `index.test.ts` verifies it.
- The PR description's claim about "no double-subscription" is accurate: the `liveQueryActive` set is the single source of truth and the guard is checked synchronously before `getHub()`, so a rapid second call is rejected before any async work starts.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The follow-up commit (`bde563af`) cleanly addresses the three items raised in the prior review:

1. **`@ts-nocheck` removed** from `useRoomLiveQuery.test.ts` — confirmed, the file has proper type-safe imports throughout.

2. **Ordering comment added** in `Room.tsx` (lines 37-40) — documents the intentional declaration order between `useRoomLiveQuery(roomId)` and the `select()` effect, both sharing `[roomId]` as their dependency. The comment is accurate: Preact/React runs effects in declaration order for the same dependency array, so LiveQuery handlers register before `select()` fires.

3. **Error-path test added** — `room-store-tasks-live-query.test.ts` lines 310-352 verify that when `getHub()` rejects, `liveQueryActive` is cleared so a subsequent `subscribeRoom()` call succeeds. This directly tests the `catch` block at lines 496-500 of `room-store.ts`.

### Re-review of the full PR

Reviewing the complete implementation again with fresh eyes:

- **Hook (`useRoomLiveQuery.ts`)**: Minimal, correct lifecycle adapter. The `useEffect` cleanup returns `unsubscribeRoom(roomId)` using the captured closure value, which correctly unsubscribes the *old* roomId on dependency change before the new effect runs. No issues.

- **Store (`room-store.ts`)**: The `subscribeRoom`/`unsubscribeRoom` split is well-guarded:
  - Double-subscribe prevention via `liveQueryActive` set
  - Three async race guards (after `getHub()`, after each `hub.request`)
  - Error path clears `liveQueryActive` so recovery is possible
  - Reconnect handlers check `liveQueryActive` instead of `roomId.value`, which is correct since LiveQuery lifecycle is now decoupled from `select()`

- **Integration (`Room.tsx`)**: Hook placement before `select()` effect is correct and now documented.

- **Tests**: All five test files cover the key scenarios — mount/unmount/room-change lifecycle, snapshot/delta handling, unsubscribe cleanup, reconnect behavior, and the new error recovery path. Test setup/teardown correctly calls `subscribeRoom`/`unsubscribeRoom` to simulate the hook lifecycle.

No P0, P1, or P2 issues found. The implementation is solid and the feedback has been addressed.

lsm (commented):

**Recommendation: APPROVE** — all three items from the prior review have been adequately addressed. Notes below are minor observations only.

---

### What was fixed

**1. `@ts-nocheck` removed from `useRoomLiveQuery.test.ts`**

Confirmed. The file at `packages/web/src/hooks/__tests__/useRoomLiveQuery.test.ts` no longer carries the suppression directive. Types are fully checked.

**2. Ordering comment in `Room.tsx`**

Lines 37-40 in `packages/web/src/islands/Room.tsx` now document the intentional declaration-order contract between `useRoomLiveQuery` and the `select()` effect. The explanation is clear and accurate.

**3. New error-path test for `liveQueryActive` cleanup on `getHub()` rejection**

`packages/web/src/lib/__tests__/room-store-tasks-live-query.test.ts` — the new `'RoomStore — subscribeRoom error path'` describe block at line 310 correctly exercises the catch path in `subscribeRoom`, confirms that `liveQueryActive.delete(roomId)` fires on rejection, and then verifies that a subsequent `subscribeRoom` call succeeds and registers handlers. Covers the reported P2 concern well.

---

**P3 — `goalsLoading` left as `true` in mid-flight guard path**

In `packages/web/src/lib/room-store.ts` around line 449: `this.goalsLoading.value = true` is set before the second `hub.request` await. If `unsubscribeRoom` races in while that await is in-flight, the mid-flight guard (line 457) runs all cleanups and returns, but `goalsLoading` is never reset to `false` in that code path. In practice this is harmless because the immediately subsequent `roomStore.select(null)` call (from the `Room.tsx` unmount effect) resets it at `doSelect` line 276. The window where the signal is stuck is sub-frame and invisible to users. Worth noting but not blocking.

**P3 — Mid-flight guard runs cleanups a second time when `unsubscribeRoom` already ran**

When `unsubscribeRoom` is called while `hub.request('liveQuery.subscribe', ...tasks...)` is in-flight, `unsubscribeRoom` runs all functions in `liveQueryCleanups.get(roomId)` (which is the same `cleanups` array) and then deletes the map entry. The mid-flight guard at line 378 then iterates the same local `cleanups` reference a second time. This is safe because the unsub functions (array splice) are idempotent when the index is not found, but it is a subtle double-call that could trip up a future maintainer. A `this.liveQueryCleanups.delete(roomId)` before the guard's loop (mirroring the catch path) would prevent it. Not blocking.

---

**Overall:** the implementation is clean, the race-condition guards are correct, the tests are comprehensive, and the three specific items from the prior review are fully resolved. Good to merge.

PR #712, opened 2026-03-22
feat: add stale-event guard for LiveQuery subscriptions (Task 3.6)

lsm:
Adds an explicit activeSubscriptionIds Set to RoomStore that is cleared
immediately in unsubscribeRoom before handler teardown. This ensures any
events already queued in the JS event loop during rapid room switching are
discarded rather than applied to stale state.

- Track active subscriptionIds per subscription (tasks.byRoom, goals.byRoom)
- Add guard check in all snapshot/delta handlers: discard if subscriptionId
  is no longer in the active set
- Clear guard in unsubscribeRoom before running cleanups (belt-and-suspenders)
- Guard cleanup also included in cleanups array (handles abort paths)
- 6 new stale-event guard tests: snapshot/delta discarded for tasks and goals
  after unsubscribeRoom, using captured-handler technique to simulate
  events queued between guard clear and handler removal

Existing reconnect re-subscribe and room-switch unsubscribe were already
implemented in Task 3.5; this task adds the explicit stale-event guard layer.

### DIFF
--- a/packages/web/src/lib/__tests__/room-store-stale-event-guard.test.ts
+++ b/packages/web/src/lib/__tests__/room-store-stale-event-guard.test.ts
@@ -299,3 +299,99 @@ describe('RoomStore — stale-event guard (Task 3.6)', () => {
 		expect(roomStore.tasks.value.map((t) => t.id)).toEqual(['t1', 't2']);
 	});
 });
+
+describe('RoomStore — reconnect re-subscribe and guard interaction (Task 3.6)', () => {
+	let hub: MockHub;
+
+	beforeEach(async () => {
+		hub = createMockHub();
+		setupHubRequests(hub);
+		vi.mocked(connectionManager.getHub).mockResolvedValue(hub as never);
+		vi.mocked(connectionManager.getHubIfConnected).mockReturnValue(hub as never);
+		await roomStore.select(ROOM_ID);
+		await roomStore.subscribeRoom(ROOM_ID);
+	});
+
+	afterEach(async () => {
+		roomStore.unsubscribeRoom(ROOM_ID);
+		await roomStore.select(null);
+		vi.clearAllMocks();
+	});
+
+	it('re-subscribes on reconnect and accepts the resulting snapshot', () => {
+		// Initial snapshot populates tasks
+		hub.fire('liveQuery.snapshot', {
+			subscriptionId: TASKS_SUB_ID,
+			rows: [makeTask('t1')],
+			version: 1,
+		});
+		expect(roomStore.tasks.value.map((t) => t.id)).toEqual(['t1']);
+
+		// Simulate WebSocket reconnect
+		hub.request.mockClear();
+		hub.fireConnection('connected');
+
+		// The reconnect handler must have sent a liveQuery.subscribe request
+		const calls = hub.request.mock.calls as [string, unknown][];
+		const resubCall = calls.find(
+			([method, params]) =>
+				method === 'liveQuery.subscribe' &&
+				(params as { queryName: string }).queryName === 'tasks.byRoom'
+		);
+		expect(resubCall).toBeDefined();
+		expect(resubCall![1]).toMatchObject({
+			queryName: 'tasks.byRoom',
+			params: [ROOM_ID],
+			subscriptionId: TASKS_SUB_ID,
+		});
+
+		// The server sends a fresh snapshot — guard must allow it through
+		hub.fire('liveQuery.snapshot', {
+			subscriptionId: TASKS_SUB_ID,
+			rows: [makeTask('t1'), makeTask('t2')],
+			version: 2,
+		});
+		expect(roomStore.tasks.value.map((t) => t.id)).toEqual(['t1', 't2']);
+	});
+
+	it('does NOT re-subscribe on reconnect after unsubscribeRoom', () => {
+		// Unsubscribe (simulates component unmount / room switch away)
+		roomStore.unsubscribeRoom(ROOM_ID);
+		hub.request.mockClear();
+
+		// Reconnect fires — should NOT trigger re-subscribe since room is unsubscribed
+		hub.fireConnection('connected');
+
+		const calls = hub.request.mock.calls as [string, unknown][];
+		const resubCall = calls.find(
+			([method, params]) =>
+				method === 'liveQuery.subscribe' &&
+				(params as { queryName: string }).queryName === 'tasks.byRoom'
+		);
+		expect(resubCall).toBeUndefined();
+	});
+
+	it('reconnect snapshot for goals passes the guard and clears goalsLoading', () => {
+		// Clear initial loading state with a snapshot
+		hub.fire('liveQuery.snapshot', {
+			subscriptionId: GOALS_SUB_ID,
+			rows: [],
+			version: 1,
+		});
+		expect(roomStore.goalsLoading.value).toBe(false);
+
+		// Simulate reconnect — reconnect handler sets goalsLoading = true
+		hub.request.mockClear();
+		hub.fireConnection('connected');
+		expect(roomStore.goalsLoading.value).toBe(true);
+
+		// Fresh snapshot from server must pass guard and clear loading
+		hub.fire('liveQuery.snapshot', {
+			subscriptionId: GOALS_SUB_ID,
+			rows: [makeGoal('g1')],
+			version: 2,
+		});
+		expect(roomStore.goalsLoading.value).toBe(false);
+		expect(roomStore.goals.value.map((g) => g.id)).toEqual(['g1']);
+	});
+});
--- a/packages/web/src/lib/__tests__/room-store-stale-event-guard.test.ts
+++ b/packages/web/src/lib/__tests__/room-store-stale-event-guard.test.ts
@@ -1,0 +1,301 @@
+/**
+ * Tests for RoomStore stale-event guard (Task 3.6)
+ *
+ * Verifies that snapshot and delta events are discarded after unsubscribeRoom
+ * is called, even if the event was already queued in the JS event loop (i.e.,
+ * the handler function was captured before cleanup ran).
+ *
+ * This guards against rapid room switching where:
+ * 1. Events arrive for subscription A while the JS engine is in the middle of
+ *    switching to subscription B.
+ * 2. In-flight events from a prior connection arrive after re-subscription.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import type { TaskSummary, RoomGoal, GoalStatus } from '@neokai/shared';
+
+// ---------------------------------------------------------------------------
+// Mocks
+// ---------------------------------------------------------------------------
+
+type EventHandler<T = unknown> = (data: T) => void;
+
+interface MockHub {
+	_handlers: Map<string, EventHandler[]>;
+	_connectionHandlers: EventHandler[];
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

The stale-event guard implementation is clean, correct, and well-tested. The `activeSubscriptionIds` Set provides a reliable synchronous guard that prevents in-flight LiveQuery events from mutating state after `unsubscribeRoom` has been called. The approach is sound: IDs are added before handler registration, checked inside every handler, and cleared at the top of `unsubscribeRoom` before handler teardown — exactly the right ordering to close the race window.

### What works well

- **Guard ordering is correct**: `unsubscribeRoom` clears the subscription IDs *before* running cleanup functions, which means even if a queued event fires between the guard clear and handler removal, it will be discarded.
- **Belt-and-suspenders cleanup**: The `activeSubscriptionIds.delete()` is both explicitly called in `unsubscribeRoom` AND pushed into the cleanups array, covering the abort paths at lines 393-403 and 477-487.
- **Reconnect compatibility**: On reconnect, the `onConnection` handler re-subscribes with the same `subscriptionId`. Since `activeSubscriptionIds` retains the ID for the lifetime of the subscription, reconnect-delivered snapshots correctly pass the guard.
- **Test design**: The captured-handler technique is a clever way to simulate the exact race condition (event queued before guard clear, delivered after). All four event types (task/goal x snapshot/delta) are covered, plus the positive case and re-subscribe cycle.

### Minor observations (P3)

1. **Pre-existing: catch block does not run cleanups** (`room-store.ts` lines 516-520): When `subscribeRoom` throws (e.g., `hub.request` rejects), the catch block deletes `liveQueryCleanups` without running them. This means both event handler registrations and `activeSubscriptionIds` entries leak. This is a pre-existing issue from Task 3.5 — not introduced by this PR — but worth noting for a future fix. The leaked `activeSubscriptionIds` entries are harmless (they'd allow events through rather than block them, and the handlers themselves are also leaked).

2. **No test for the error/catch path**: There is no test verifying behavior when `subscribeRoom` fails mid-way (e.g., the `liveQuery.subscribe` request rejects). Again, this is a pre-existing gap, not a regression from this PR.

3. **Shared Set across rooms**: `activeSubscriptionIds` is a single flat Set shared across all rooms. This is fine because subscription IDs embed the room ID (`tasks-byRoom-{roomId}`), making them globally unique. Just noting for future maintainers.

### Verdict

No bugs, no security concerns, no missing edge cases within the scope of this task. The implementation is minimal, focused, and integrates cleanly with the existing subscription lifecycle. The pre-existing catch-block cleanup gap (P3) is worth a follow-up but does not affect this PR's correctness.

lsm (commented):

> **Recommendation:** REQUEST_CHANGES (1 P0, 2 P1, 2 P2)

The stale-event guard design is sound: placing `activeSubscriptionIds.add()` before handler registration and `delete()` before handler teardown correctly covers the JS event-loop window described in the task. The IDs are namespaced per-room (`tasks-byRoom-<roomId>`, `goals-byRoom-<roomId>`), so a future multi-room scenario would not cause false drops. The test approach (capture handlers before `unsubscribeRoom`, fire them afterward) is the right way to simulate in-flight events.

However, two exception/abort paths have cleanup gaps that can leave the store in a broken state.

---

### P0: Catch block leaks registered event handlers

**File:** `packages/web/src/lib/room-store.ts`, lines 516–519

When `hub.request('liveQuery.subscribe', ...)` or any other awaited call inside `subscribeRoom` throws, the catch block does:

```ts
} catch (err) {
    this.liveQueryActive.delete(roomId);
    this.liveQueryCleanups.delete(roomId);   // discards the array WITHOUT calling the fns
    logger.error('Failed to subscribe room LiveQuery:', err);
}
```

At the point of failure, the `cleanups` array may already contain:
- `unsubTaskSnapshot` — a live handler on `hub._handlers['liveQuery.snapshot']`
- `unsubTaskDelta` — a live handler on `hub._handlers['liveQuery.delta']`
- The guard cleanup `() => activeSubscriptionIds.delete(tasksSubId)`
- Possibly the goals equivalents

None of these are called. The hub retains the handlers. On the next `subscribeRoom` call (e.g., after retry or reconnect), new handlers are added on top of the leaked ones. Those leaked handlers will fire and mutate state for the old subscription context for as long as the hub connection lives.

The fix is to call the existing cleanups in the catch block before deleting the map entry, mirroring what the abort paths at lines 393–402 and 477–486 already do correctly.

---

### P1: `goalsLoading` left `true` when goals abort-path fires

**File:** `packages/web/src/lib/room-store.ts`, lines 469–486

```ts
this.goalsLoading.value = true;
await hub.request('liveQuery.subscribe', { queryName: 'goals.byRoom', ... });

if (!this.liveQueryActive.has(roomId)) {
    for (const fn of cleanups) { fn(); }
    this.liveQueryCleanups.delete(roomId);
    return;   // <-- goalsLoading is never reset to false here
}
```

`unsubscribeRoom` clears `goalsLoading.value` only indirectly — it doesn't touch it at all. The `doSelect` reset at line 284 (`this.goalsLoading.value = false`) only runs when the user navigates to a different room, not when `subscribeRoom` is aborted mid-flight for the current room. In a rapid-subscribe/unsubscribe sequence (e.g., component remount), the goals section will show a permanent loading spinner.

---

### P1: `activeSubscriptionIds` not cleaned up in catch block

**File:** `packages/web/src/lib/room-store.ts`, lines 343, 434, 516–519

This is related to the P0, but distinct: when the error path triggers after `this.activeSubscriptionIds.add(tasksSubId)` has already run, the catch block doesn't call the cleanup lambda that was pushed into `cleanups`. `activeSubscriptionIds` retains the stale IDs until `unsubscribeRoom` is explicitly called.

In practice this is survivable because `unsubscribeRoom` manually deletes both IDs at lines 533–534, but until then, any future events with those subscription IDs would be incorrectly processed rather than discarded — the opposite of what the guard is meant to do.

---

### P2: No test for reconnect re-subscribe + guard interaction

The test suite covers unsubscribe-based stale discard well. But the reconnect path (`hub.onConnection('connected')` → `hub.request('liveQuery.subscribe', ...)`) is untested. There is no test verifying that:
1. A `connected` event after a simulated disconnect re-triggers `liveQuery.subscribe` and the resulting snapshot is accepted.
2. A `connected` event after `unsubscribeRoom` does NOT re-trigger `liveQuery.subscribe` (the `liveQueryActive` guard on line 407/494 should prevent it, but this is not tested).

The last test ("re-establishes guard after subscribeRoom is called again") tests component remount, not the reconnection handler. These are distinct code paths.

---

### P2: `goalsLoading` not reset in tasks abort-path either

**File:** `packages/web/src/lib/room-store.ts`, lines 393–402

The first abort path (tasks subscribe fails fast) also does not reset `goalsLoading`. While `goalsLoading` hasn't been set to `true` yet at that point (it's set at line 469, after this abort), the tasks abort path also omits resetting `this.liveQueryActive` — actually it doesn't need to since `unsubscribeRoom` already cleared it. This path is correct but worth noting for completeness.

---

### Minor Notes (P3)

- The test file's `createMockHub` and signal mocks are duplicated verbatim from `room-store-tasks-live-query.test.ts`. Extracting a shared test helper file would reduce future drift risk.
- The `makeGoal` and `makeTask` factory functions cast with `as TaskSummary` / cast-implicit for `RoomGoal`. These casts suppress type errors — if the shared types add required fields, the tests will silently build with incomplete fixtures.

lsm (commented):

### Re-review of Task 3.6: Reconnect and stale-event guards

The three issues from the first review have been addressed correctly. All 9 tests pass.

### Verification of fixes

**P0 fix (catch block leaking handlers):** Lines 519-537 in `room-store.ts` now correctly read the cleanups from `this.liveQueryCleanups.get(roomId)` and invoke each entry before deleting the map entry. This prevents handler leaks when an error occurs mid-subscription. The `activeSubscriptionIds` entries added before the error are also cleaned up via the cleanup lambdas. Confirmed correct.

**P1 fix (goalsLoading stuck true on unsubscribe during await):** Line 488 resets `goalsLoading.value = false` in the second abort guard (after `goals.byRoom` subscribe await). Line 536 does the same in the catch block. Both paths are covered. Confirmed correct.

**P1 fix (reconnect test coverage):** Three new tests in the "reconnect re-subscribe and guard interaction" describe block cover: (1) re-subscribe on reconnect accepts fresh snapshots, (2) no re-subscribe after unsubscribeRoom, (3) goals reconnect snapshot passes guard and clears goalsLoading. These are well-structured and meaningful tests.

### Remaining observations

**P3 — No test for the catch-block error path.** The catch block's cleanup logic (lines 519-537) is untested. An error thrown by `hub.request('liveQuery.subscribe', ...)` after some handlers are already registered would exercise this path. This is a minor gap since the code is straightforward, but it could be added in a follow-up.

**P3 — Double-delete of activeSubscriptionIds is harmless but slightly wasteful.** `unsubscribeRoom` explicitly deletes subscription IDs from `activeSubscriptionIds`, then runs cleanups which include lambdas that delete the same IDs again. This is idempotent and correct; just noting it for awareness.

### Summary

All previously identified P0/P1/P2 issues have been properly resolved. The stale-event guard mechanism is sound — `activeSubscriptionIds` is cleared synchronously in `unsubscribeRoom` before async handler teardown, which correctly guards against queued events in the JS event loop. The reconnect path correctly re-subscribes and the guard allows fresh post-reconnect snapshots through.

**Recommendation: APPROVE** — only P3 nits remain.

lsm (commented):

**Recommendation: APPROVE** (Round 2 — all P0/P1/P2 issues from round 1 resolved)

---

### Verification of Round 1 Issues

**P0: Catch block leaked event handlers** — Fixed at lines 524–533 of `room-store.ts`. The catch block now retrieves `this.liveQueryCleanups.get(roomId)` and iterates it, calling each cleanup function before deleting the map entry. This correctly tears down any handlers registered up to the point of failure and clears `activeSubscriptionIds` entries (since those deletes are included in the cleanups array via the lambdas pushed at lines 344 and 435).

**P1a: `goalsLoading` stuck `true` on goals abort-path** — Fixed at lines 486–488. The second abort guard (after the goals `liveQuery.subscribe` await) now resets `goalsLoading.value = false` before returning.

**P1b: `activeSubscriptionIds` not cleaned in catch** — Resolved as a side effect of the P0 fix: the cleanup lambdas for `activeSubscriptionIds.delete()` are in the cleanups array and are now called in the catch block.

**P2: No reconnect tests** — Three new tests added in the second describe block (`lines 321–396`):
1. Re-subscribes on reconnect and accepts the resulting snapshot — correctly verifies the `liveQuery.subscribe` call is made with the right params and the snapshot passes the guard.
2. Does NOT re-subscribe after `unsubscribeRoom` — verifies the `liveQueryActive` guard on the `onConnection` handler.
3. Reconnect snapshot for goals passes guard and clears `goalsLoading` — exercises the `goalsLoading = true` → snapshot → `goalsLoading = false` cycle on reconnect.

All 9 tests pass on the current commit.

---

### No New Issues Introduced

The fixes are minimal and targeted. No regressions observed.

One remaining observation (P3, pre-existing, not introduced by this PR): the reconnect handler for goals sets `goalsLoading.value = true` but can only reset it via the snapshot handler or the `.catch()` path. If the server accepts the re-subscribe but never sends a snapshot (a server-side fault), `goalsLoading` stays `true` permanently for that session. This is a pre-existing design constraint from the snapshot-based loading model and is not within the scope of this task.

PR #714, opened 2026-03-22
fix: reset in_progress tasks to pending when sessions are lost after daemon restart

lsm:
After a daemon restart, tasks that were in_progress with dead agent
groups would get stuck as needs_attention (requiring human intervention)
or remain in_progress indefinitely.

The fix changes the recovery behavior:
- `recoverZombieGroups` now calls `terminateGroup` + `resetTaskToPending`
  instead of `fail` (which set status to needs_attention). This lets the
  next tick's spawn loop automatically pick up and re-spawn the task.
- `runtime-recovery.ts` `failGroupAndTask` now resets to pending instead
  of needs_attention so startup recovery also auto-requeues lost tasks.
- Added `TaskManager.resetTaskToPending()` for direct status reset used
  during restart recovery (bypasses normal state machine validation).

On the same tick that zombie recovery runs, the spawn loop sees the now-
pending task and immediately spawns a fresh group — no manual intervention
needed after a daemon restart.

Tests updated to reflect the new pending/in_progress behavior, and new
tests added for the respawn scenario with multiple unrestorable groups.

### DIFF
--- a/packages/daemon/src/lib/room/managers/task-manager.ts
+++ b/packages/daemon/src/lib/room/managers/task-manager.ts
@@ -260,6 +260,23 @@ export class TaskManager {
 	}
 
 	/**
+	 * Reset task to pending for automatic re-spawn after daemon restart.
+	 *
+	 * Used during recovery when a group's sessions are lost and the task should
+	 * be automatically re-queued rather than requiring human intervention.
+	 * Directly sets status to 'pending' regardless of current status (bypasses
+	 * state machine validation since 'in_progress → pending' is not a normal
+	 * user-facing transition).
+	 */
+	async resetTaskToPending(taskId: string): Promise<NeoTask> {
+		return this.updateTaskStatus(taskId, 'pending', {
+			error: null,
+			result: null,
+			progress: null,
+		});
+	}
+
+	/**
 	 * Cancel task (intentionally stopped, distinct from failure).
 	 * Cascades cancellation to any pending tasks that depend on this task,
 	 * since they can never be satisfied once their dependency is cancelled.
--- a/packages/daemon/src/lib/room/managers/task-manager.ts
+++ b/packages/daemon/src/lib/room/managers/task-manager.ts
@@ -260,23 +260,6 @@ export class TaskManager {
 	}
 
 	/**
-	 * Reset task to pending for automatic re-spawn after daemon restart.
-	 *
-	 * Used during recovery when a group's sessions are lost and the task should
-	 * be automatically re-queued rather than requiring human intervention.
-	 * Directly sets status to 'pending' regardless of current status (bypasses
-	 * state machine validation since 'in_progress → pending' is not a normal
-	 * user-facing transition).
-	 */
-	async resetTaskToPending(taskId: string): Promise<NeoTask> {
-		return this.updateTaskStatus(taskId, 'pending', {
-			error: null,
-			result: null,
-			progress: null,
-		});
-	}
-
-	/**
 	 * Cancel task (intentionally stopped, distinct from failure).
 	 * Cascades cancellation to any pending tasks that depend on this task,
 	 * since they can never be satisfied once their dependency is cancelled.
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -647,6 +647,10 @@ export class RoomRuntimeService {
 				const activeGroups = groupRepo.getActiveGroups(roomId);
 				for (const group of activeGroups) {
 					try {
+						// Re-establish message mirroring so messages written after restart
+						// are persisted to session_group_messages for LiveQuery subscribers.
+						runtime.setupMirroringIfNeeded(group);
+
 						// Restore MCP servers (planner-tools, leader-agent-tools)
 						await runtime.restoreMcpServersForGroup(group);
 
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2397,14 +2397,13 @@ export class RoomRuntime {
 					workerRestored = true;
 				} else {
 					log.error(
-						`Failed to restore worker ${group.workerSessionId}. Terminating group ${group.id} and re-queuing task.`
+						`Failed to restore worker ${group.workerSessionId}. Failing group ${group.id}.`
 					);
-					// Terminate the group (marks completedAt, frees unique constraint slot)
-					// without setting the task to needs_attention. Then reset task to pending
-					// so the next tick spawns a fresh group automatically.
-					await this.taskGroupManager.terminateGroup(group.id);
-					await this.taskManager.resetTaskToPending(group.taskId);
-					this.cleanupMirroring(group.id, 'Worker session lost — re-queuing task for respawn.');
+					await this.taskGroupManager.fail(
+						group.id,
+						'Worker session lost and could not be restored'
+					);
+					this.cleanupMirroring(group.id, 'Worker session lost — could not be restored.');
 					await this.emitTaskUpdateById(group.taskId);
 					continue;
 				}
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2153,6 +2153,17 @@ export class RoomRuntime {
 	 * state.groupMessages.delta — message delivery to the frontend is handled
 	 * by the LiveQuery subscription (sessionGroupMessages.byGroup).
 	 */
+	/**
+	 * Re-establish message mirroring for a recovered group.
+	 * Guards against double-setup: no-op if mirroring is already active.
+	 * Called during daemon recovery to ensure messages written after restart
+	 * are persisted to session_group_messages for LiveQuery subscribers.
+	 */
+	setupMirroringIfNeeded(group: SessionGroup): void {
+		if (this.mirroringCleanups.has(group.id)) return;
+		this.setupMirroring(group);
+	}
+
 	private setupMirroring(group: SessionGroup): void {
 		if (!this.daemonHub) return;
 
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2397,13 +2397,14 @@ export class RoomRuntime {
 					workerRestored = true;
 				} else {
 					log.error(
-						`Failed to restore worker ${group.workerSessionId}. Failing group ${group.id}.`
+						`Failed to restore worker ${group.workerSessionId}. Terminating group ${group.id} and re-queuing task.`
 					);
-					await this.taskGroupManager.fail(
-						group.id,
-						'Worker session lost and could not be restored'
-					);
-					this.cleanupMirroring(group.id, 'Worker session lost — could not be restored.');
+					// Terminate the group (marks completedAt, frees unique constraint slot)
+					// without setting the task to needs_attention. Then reset task to pending
+					// so the next tick spawns a fresh group automatically.
+					await this.taskGroupManager.terminateGroup(group.id);
+					await this.taskManager.resetTaskToPending(group.taskId);
+					this.cleanupMirroring(group.id, 'Worker session lost — re-queuing task for respawn.');
 					await this.emitTaskUpdateById(group.taskId);
 					continue;
 				}
--- a/packages/daemon/src/lib/room/runtime/runtime-recovery.ts
+++ b/packages/daemon/src/lib/room/runtime/runtime-recovery.ts
@@ -168,8 +168,11 @@ async function failGroupAndTask(
 	group: SessionGroup,
 	groupRepo: SessionGroupRepository,
 	taskManager: TaskManager,
-	reason: string
+	_reason: string
 ): Promise<void> {
 	groupRepo.failGroup(group.id, group.version);
-	await taskManager.failTask(group.taskId, reason);
+	// Reset task to pending so it is automatically re-spawned on the next tick.
+	// After a daemon restart, lost sessions are a transient condition — the task
+	// should retry rather than require human intervention ('needs_attention').
+	await taskManager.resetTaskToPending(group.taskId);
 }
--- a/packages/daemon/src/lib/room/runtime/runtime-recovery.ts
+++ b/packages/daemon/src/lib/room/runtime/runtime-recovery.ts
@@ -168,11 +168,8 @@ async function failGroupAndTask(
 	group: SessionGroup,
 	groupRepo: SessionGroupRepository,
 	taskManager: TaskManager,
-	_reason: string
+	reason: string
 ): Promise<void> {
 	groupRepo.failGroup(group.id, group.version);
-	// Reset task to pending so it is automatically re-spawned on the next tick.
-	// After a daemon restart, lost sessions are a transient condition — the task
-	// should retry rather than require human intervention ('needs_attention').
-	await taskManager.resetTaskToPending(group.taskId);
+	await taskManager.failTask(group.taskId, reason);
 }
--- a/packages/daemon/tests/unit/room/room-runtime-recovery-safety.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-recovery-safety.test.ts
@@ -81,7 +81,7 @@ describe('Zombie detection in tick', () => {
 		expect(updated!.completedAt).toBeNull();
 	});
 
-	it('should terminate zombie group and automatically re-spawn task when worker cannot be restored', async () => {
+	it('should fail group when zombie worker cannot be restored', async () => {
 		const { taskId, groupId, workerSessionId } = await createTaskWithGroup(ctx);
 
 		// Worker missing AND restoreSession fails
@@ -94,19 +94,13 @@ describe('Zombie detection in tick', () => {
 		ctx.runtime.start();
 		await ctx.runtime.tick();
 
-		// Old zombie group should be terminated
+		// Group should be failed
 		const updated = ctx.groupRepo.getGroup(groupId);
 		expect(updated!.completedAt).not.toBeNull();
 
-		// Task should be re-spawned on the same tick (reset to pending → spawn loop picks it up)
+		// Task should be failed
 		const updatedTask = await ctx.taskManager.getTask(taskId);
-		expect(updatedTask!.status).toBe('in_progress');
-
-		// A new (fresh) group should exist for this task
-		const activeGroups = ctx.groupRepo.getActiveGroups('room-1');
-		const newGroup = activeGroups.find((g) => g.taskId === taskId);
-		expect(newGroup).toBeDefined();
-		expect(newGroup!.id).not.toBe(groupId);
+		expect(updatedTask!.status).toBe('needs_attention');
 	});
 
 	it('should restore a zombie leader session during tick', async () => {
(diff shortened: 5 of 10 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

This PR correctly fixes the issue where running tasks get stuck at "Waiting for agent activity..." after daemon restart. The approach is sound: instead of marking tasks as `needs_attention` (requiring human intervention), tasks are reset to `pending` so the spawn loop automatically picks them up.

### What works well

1. **Two recovery paths covered**: Both `runtime-recovery.ts` (startup recovery) and `room-runtime.ts` `recoverZombieGroups()` (tick-level zombie detection) now reset tasks to pending instead of failing them.

2. **No infinite loop risk**: The spawn loop creates fresh in-memory sessions, so `findZombieGroups()` won't flag them as zombies on the next tick. The new group's sessions exist in the session cache immediately after spawn.

3. **Correct ordering in `executeTick()`**: Zombie recovery runs BEFORE the spawn loop (line 2739-2741), so the task is reset to pending in time for the spawn loop (line 2778+) to pick it up in the same tick. Tests verify the task ends up `in_progress` (not just `pending`).

4. **`resetTaskToPending()` correctly clears stale fields**: Error, result, and progress are all nulled out, preventing stale error messages from lingering.

5. **`terminateGroup()` frees the UNIQUE constraint slot**: Using `terminateGroup()` instead of `fail()` correctly sets `completedAt` on the old group, which allows `createGroup()` to insert a new row without hitting the unique index.

6. **Comprehensive test coverage**: Tests cover multiple unrestorable groups, error field clearing, successful restore (task stays `in_progress`), worker-only missing scenarios, and submitted-for-review groups. All 178 tests pass.

### Minor observations (P3)

1. **Unused `_reason` parameter in `failGroupAndTask()`** (`runtime-recovery.ts:171`): The reason string is now unused. Since this is a private function, it's not a public API concern, but the parameter could be removed entirely rather than prefixed with underscore. Alternatively, consider logging it for observability since this is a recovery path where operators might want to see why a group was reset.

2. **Function name `failGroupAndTask` is now misleading** (`runtime-recovery.ts:167`): The function no longer fails the task -- it resets it to pending. A name like `resetGroupAndTask` or `terminateGroupAndRequeueTask` would better reflect the new behavior.

These are cosmetic -- the logic is correct and well-tested.

lsm (commented):

The core fix is sound and well-structured: resetting in-progress tasks to pending on daemon restart is the right UX choice, and the ordering guarantee (zombie termination runs before the spawn loop in the same `executeTick` call) ensures the re-spawn happens atomically within a single tick. The new `resetTaskToPending` method, new tests, and updated assertions are all correct and complete.

That said, there are meaningful correctness and robustness issues that should be addressed.

---

### P1 — `review` status tasks are silently re-spawned on daemon restart

**Files:** `packages/daemon/src/lib/room/runtime/runtime-recovery.ts` (line 118–122), `packages/daemon/src/lib/room/runtime/room-runtime.ts` (line 2405–2409)

Both `failGroupAndTask` and the `recoverZombieGroups` path call `resetTaskToPending` unconditionally — there is no guard for tasks in `review` status (`submittedForReview == true`).

If a task was submitted for human review (PR already created on GitHub, human has been notified, task.status == `review`) and the daemon restarts with a lost worker session, both code paths will:
1. Terminate the group
2. Reset the task from `review` → `pending`
3. On the next tick, spawn a brand-new group that starts the task from scratch

The spawned agent will likely try to create another PR, resulting in duplicate PRs on the same branch or confusing the reviewer who was already looking at the first one. The `prUrl`, `prNumber`, `prCreatedAt` fields remain on the task while its status is `pending`, creating an inconsistent document.

This is visible in the updated test `should terminate zombie submitted_for_review group and re-spawn task automatically` — it explicitly asserts `in_progress` for a task that had `review` status, which validates this problematic behavior rather than guarding against it.

**Suggested fix:** In `failGroupAndTask` (runtime-recovery.ts) and `recoverZombieGroups` (room-runtime.ts), check whether the group's `submittedForReview` flag is set. If it is, keep the task in `review` status (do not re-spawn) and emit a UI update so the operator sees the session was lost. The task is already safe — the PR exists and is waiting for human action.

---

### P2 — `terminateGroup` null return not guarded before `resetTaskToPending`

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` (lines 2405–2406)

```ts
await this.taskGroupManager.terminateGroup(group.id);
await this.taskManager.resetTaskToPending(group.taskId);
```

`terminateGroup` returns `SessionGroup | null` — it returns null when the optimistic-lock `failGroup(groupId, group.version)` call sees a version conflict (another concurrent update beat it). If that happens, the code proceeds to call `resetTaskToPending` anyway, leaving the task in `pending` status while its group still has `completedAt = null`.

The spawn loop then skips the task (active group exists in `activeGroupTaskIds`) and the task is permanently stuck: `pending` with a live group that has no active sessions.

The job queue serializes ticks so this is unlikely in production, but defensive code should match the pattern used elsewhere (e.g. the similar guard at lines 1775 and 1825). The `runtime-recovery.ts` path has the same issue: `groupRepo.failGroup` is called without checking its return value.

**Suggested fix:**
```ts
const terminated = await this.taskGroupManager.terminateGroup(group.id);
if (!terminated) {
    log.warn(`[recoverZombieGroups] terminateGroup(${group.id}) returned null — skipping task reset`);
    continue;
}
await this.taskManager.resetTaskToPending(group.taskId);
```

---

### P2 — No retry limit across daemon restarts

**Files:** `packages/daemon/src/lib/room/runtime/runtime-recovery.ts`, `packages/daemon/src/lib/room/runtime/room-runtime.ts`

If an agent session consistently fails to initialize (e.g., a model configuration issue that causes the worker process to abort before registering with the session factory), repeated daemon restarts will cycle the task infinitely:

1. Restart → `recoverZombieGroups` can't restore sessions → `resetTaskToPending`
2. Spawn loop spawns new group → agent fails immediately → zombie
3. Next restart → back to step 1

Unlike the dead-loop detector (which only fires during live operation) or `consecutiveFailures` on goals (which only counts successful agent runs that call `fail_task`), there is no counter for restart-recovery cycles. Over time the task is never retried successfully and never surfaces to `needs_attention` for human inspection.

A simple fix would be to increment a `restartRecoveryAttempts` counter on the task and fall back to `failTask` (`needs_attention`) after N attempts.

---

### P3 — Stale comment in `cleanStaleGroupsForTask`

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` (approx. line 2682–2686)

The doc comment still reads: `recoverZombieGroups() … will always handle the zombie first (failing the group and moving the task to needs_attention).`

This should now read: `…terminating the group and resetting the task to pending for automatic re-spawn.`

---

### P3 — Unused `_reason` parameter in `failGroupAndTask`

**File:** `packages/daemon/src/lib/room/runtime/runtime-recovery.ts` (line 171)

The `_reason` parameter was previously used to call `taskManager.failTask(group.taskId, reason)`. Now that it calls `resetTaskToPending` instead, the reason string is silently dropped. This is fine for linting (underscore prefix), but the parameter could be removed from the signature entirely if it serves no future purpose, or a log call added so the reason is at least recorded.

---

### What works well

- The ordering guarantee is correct: `recoverZombieGroups` terminates the old group before the fresh `getActiveGroups` call in the spawn loop, so the task is immediately picked up in the same tick.
- `resetTaskToPending` correctly bypasses state machine validation (no regression to the status transition guards) and clears error/result/progress fields.
- Test coverage is solid for the new `pending` path: multiple tasks, single task, error-field clearing, and the cross-tick re-spawn are all covered.
- The `terminateGroup` + `resetTaskToPending` separation (vs. the old `fail()` which also sets task status) correctly avoids the double write that set the task to `needs_attention` before.

lsm (commented):

### Summary

This PR contains two commits: a daemon-side fix (89c19bc71) and its exact revert (be5417e99). The net diff against `dev` is **zero** — no code changes are introduced by merging this PR.

The actual UI fix for "tasks stuck at Waiting for agent activity..." was already landed in `dev` via PR #707 (commit 73485ec58). I verified that fix is present and well-implemented:

1. **`useGroupMessages.ts`** — Has retry logic with backoff (2 retries at 500ms/1500ms), stale-subscription guards via `activeSubIdRef`, proper `clearTimeout` in cleanup, and exposes `isReconnecting` state derived from `!isConnected && groupId !== null`. The `isConnected` dependency in the effect ensures re-subscription on WebSocket reconnect.

2. **`TaskView.tsx`** — `fetchGroup` retries once after 1s on failure. The effect depends on `[roomId, taskId, isConnected]` so it re-fetches the group when the WebSocket reconnects. Uses a sequence counter (`fetchGroupSeq`) to avoid stale responses.

3. **`TaskConversationRenderer.tsx`** — Renders "Reconnecting..." when `isReconnecting` is true, preventing the misleading "Waiting for agent activity..." message during disconnects.

4. **Tests** — All three files have corresponding tests covering retry, reconnect, and `isReconnecting` scenarios (728, 619, and 2689 lines respectively).

### Assessment

The UI fix in `dev` is correct and complete. The revert in this PR is clean — same 7 files with symmetric additions/deletions.

**However, this PR should not be merged as-is.** It introduces two commits with zero net effect, which adds noise to the git history. The PR description still describes the daemon-side approach that was reverted, which is misleading. 

#### P2: PR has zero net diff — should be closed rather than merged

The commit + revert pair adds unnecessary noise to the git history. Since the UI fix is already in `dev` (PR #707) and this PR produces no code changes, the cleanest action is to **close this PR without merging**. If the branch needs to be cleaned up, that can be done separately.

#### P3: PR title and description describe the reverted approach

The title says "reset in_progress tasks to pending when sessions are lost after daemon restart" which describes the daemon-side approach that was reverted, not the UI fix that actually solved the problem. This would be confusing in git history if merged.

lsm (commented):

---

### Summary

PR #714 is a net-zero diff: it contains exactly two commits — the daemon-side implementation (`89c19bc71`) immediately followed by its own exact revert (`be5417e99`). The branch adds no new code to `dev`. The PR description and title both describe the daemon-side fix as if it is still active, which is misleading.

---

### What was actually requested vs what is here

The task asked for a UI-only fix in three files:
- `packages/web/src/hooks/useGroupMessages.ts`
- `packages/web/src/components/room/TaskView.tsx`
- `packages/web/src/components/room/TaskConversationRenderer.tsx`

That fix was correctly implemented and merged to `dev` via PR #707 (commit `73485ec58`, merged 2026-03-22). The UI work is complete and solid.

PR #714 then introduced a daemon-side auto-respawn approach (`resetTaskToPending`, `recoverZombieGroups` changes) and immediately reverted it, leaving zero net changes.

---

### Issues

**P0 — PR description is factually wrong**

The PR description and title both claim active changes that do not exist in the diff:
- "recoverZombieGroups now calls terminateGroup + resetTaskToPending instead of fail"
- "runtime-recovery.ts failGroupAndTask now resets to pending instead of needs_attention"
- "Added TaskManager.resetTaskToPending() for direct status reset"

None of these things are true in the current HEAD. `failGroupAndTask` still calls `taskManager.failTask` which transitions to `needs_attention`. `resetTaskToPending` does not exist in `task-manager.ts`. The PR description must be updated to accurately reflect reality — that this PR reverts the daemon-side approach and that the UI fix already landed in PR #707.

**P1 — PR should probably be closed rather than merged**

Since the net diff is empty (implementation + revert cancel out), merging this PR into `dev` produces no code change. The correct action is either:
1. Close this PR with a comment pointing to PR #707 as the actual fix, or
2. Update the PR to contain only the revert commit (with an updated description explaining why the daemon-side approach was reverted and that the UI fix from #707 is sufficient).

Merging a no-op PR adds noise to the commit history on `dev`.

---

### Verification: UI fix correctness (already in dev via #707)

The UI fix that landed is correct and complete:

1. **`useGroupMessages.ts`** — Retry logic (`MAX_RETRIES=2`, `RETRY_DELAYS_MS: [500, 1500]`) is properly implemented. The `isReconnecting` derived value (`!isConnected && groupId !== null`) is a clean computed field with no extra state. `isConnected` is in the effect dependency array so re-subscription fires on reconnect. `clearTimeout` in cleanup prevents orphaned timers. The tuple type on `RETRY_DELAYS_MS` prevents index mismatches.

2. **`TaskView.tsx`** — `isConnected` added to the main effect dep array (line 907) closes the reconnect gap for mounted views. The 1-second retry for `task.getGroup` is simple and correct. The "Loading conversation history" copy change for review tasks is an improvement.

3. **`TaskConversationRenderer.tsx`** — `isReconnecting` check renders before `isLoading`, which is the right ordering (disconnect is distinguishable from first load).

4. **Tests** — `useGroupMessages.test.ts` has full coverage: retry exhaustion, retry success, stale-groupId cancellation, `isReconnecting` lifecycle. `TaskConversationRenderer.test.tsx` covers the Reconnecting→Loading→Messages state machine. `TaskView.test.tsx` covers the `getGroup` retry and the review-task empty state. No gaps found.

---

### Action required

Update the PR description (and optionally the title) to reflect the actual content: this PR reverts the daemon-side respawn approach, with the reason that the UI-only fix from PR #707 is the adopted solution. Then decide whether to merge the no-op or close it.

lsm:
Net-zero diff — the UI fix already landed in dev via PR #707. Closing this PR.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix correctly addresses the root cause: after daemon restart, message mirroring (sdk.message -> session_group_messages) was not re-established for recovered groups, so LiveQuery subscribers saw no messages even though the agent was actively working. The four changes are minimal, focused, and correct.

### Analysis

**1. `setupMirroringIfNeeded` (room-runtime.ts:2162-2164)** — Correct. The guard `this.mirroringCleanups.has(group.id)` prevents double-setup and is consistent with how the existing `setupMirroring` tracks its state. The method delegates to the private `setupMirroring` which handles the `daemonHub` null check internally (line 2168). This is a safe, minimal public wrapper.

**2. Recovery loop call (room-runtime-service.ts:652)** — Correct placement. It runs inside the `restoredSessions > 0` block, iterating over `getActiveGroups(roomId)`. The `IfNeeded` guard means it is safe to call for groups that might already have mirroring (though after a restart, none should). For groups in `submittedForReview` or `waitingForQuestion` states, mirroring is still needed — those groups may receive new messages when the human responds or the SDK query restarts. So calling it unconditionally for all active groups is correct.

**3. TaskView.tsx disconnected guard (line 879-882)** — Correct. `isConnected` is in the effect's dependency array (line 911), so when the connection is restored the effect re-runs and `load()` executes normally. This prevents a failed RPC call during disconnect from setting a persistent error state.

**4. "No messages yet." text (TaskConversationRenderer.tsx:279)** — Appropriate. The old "Waiting for agent activity..." was misleading after a restart when messages were simply not persisted. The new text is neutral and accurate.

### Minor observations (P3)

- **No tests added for `setupMirroringIfNeeded`.** The method is a thin guard + delegation, so the risk is low — but a unit test verifying the no-op behavior when `mirroringCleanups` already has the group ID would add confidence for future refactors.

- **Silent failure path:** If `this.daemonHub` is null when `setupMirroringIfNeeded` is called, it passes the `mirroringCleanups` guard but `setupMirroring` returns without registering a cleanup. Subsequent calls would repeat this silent no-op. During recovery `daemonHub` should always be initialized, so this is not a practical concern — just noting it for awareness.

- **PR description/body mismatch:** The PR body still describes the first commit's approach (resetTaskToPending, recoverZombieGroups changes) which was reverted. The actual shipped changes are about message mirroring restoration and UI fixes. Consider updating the PR body to match the final commit.

lsm (commented):

> **Recommendation:** REQUEST_CHANGES

---

## Summary

The core fix — restoring message mirroring in `recoverRoomRuntime` — addresses the right root cause: post-restart messages from continuation injections were never persisted to `session_group_messages` because `setupMirroring` was never called for recovered groups. The `setupMirroringIfNeeded` guard is safe against double-setup. However, there are issues that need addressing before merge.

---

## P1: Condition gates mirroring on wrong metric

**File:** `packages/daemon/src/lib/room/runtime/room-runtime-service.ts`, lines 646–652

```ts
if (result.restoredSessions > 0) {
    const activeGroups = groupRepo.getActiveGroups(roomId);
    for (const group of activeGroups) {
        runtime.setupMirroringIfNeeded(group);
```

The guard uses `restoredSessions > 0`, but the correct predicate is `recoveredGroups > 0`. `restoredSessions` counts individual session restores (one per session call where `isLive` was false and `restoreSession` succeeded). `recoveredGroups` counts groups that made it past the `failGroupAndTask` early-return.

In practice after a full daemon restart the in-memory session cache is empty, so every successful session restore increments `restoredSessions` and the condition passes. But if a future code path marks sessions as live before recovery runs (or a partial restart scenario), `restoredSessions` could be 0 while `recoveredGroups > 0`, and no group would get mirroring set up. The intent is clearly "for every group that survived recovery, set up mirroring" — that maps to `recoveredGroups`, not `restoredSessions`.

---

## P2: Orphaned JSDoc block creates misleading documentation

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 2148–2165

The original JSDoc comment for the section (lines 2148–2155) described what `setupMirroring` does. The new `setupMirroringIfNeeded` method was inserted between that comment and `setupMirroring`, so the comment block now appears to document `setupMirroringIfNeeded` instead. The result is two consecutive JSDoc blocks above `setupMirroringIfNeeded` — the first one describing `setupMirroring` — and `setupMirroring` is now undocumented. Move the original JSDoc block to precede `setupMirroring` directly, and leave only the new JSDoc above `setupMirroringIfNeeded`.

---

## P2: No unit tests for `setupMirroringIfNeeded` or the recovery mirroring path

**Files:** new code in `room-runtime.ts` + `room-runtime-service.ts`

Neither the no-op guard behavior nor the integration of `setupMirroringIfNeeded` in `recoverRoomRuntime` is tested. A unit test should cover at minimum:

1. Calling `setupMirroringIfNeeded` once sets up mirroring (messages are captured).
2. Calling it a second time for the same group is a no-op (cleanup is only registered once, no duplicate listeners).
3. After recovery, `sdk.message` events for recovered sessions are persisted to `session_group_messages`.

The existing `room-runtime-flow.test.ts` already tests the mirroring lifecycle; the new public method should be covered there or in a dedicated recovery test. The `recoverRoomRuntime` service path is currently only testable via expensive online tests.

---

## P3: PR title describes the reverted approach, not the actual fix

The PR title reads: _"fix: reset in_progress tasks to pending when sessions are lost after daemon restart"_. The commit history shows that approach was introduced and immediately reverted in this same PR. The final commit is titled _"fix: restore message mirroring on recovery and fix UI stuck states"_, which accurately describes what the PR actually does. The PR title should be updated to match.

---

## P3: Missing loading indicator on reconnect in `TaskView`

**File:** `packages/web/src/components/room/TaskView.tsx`, lines 878–882

When `isConnected` is `false`, `setLoading(false)` clears the loading state. When `isConnected` transitions back to `true` and the effect re-runs, `load()` fires without setting `loading` back to `true` first. This means during the reconnection fetch there is no loading spinner — users see the stale (or empty) task view while the request is in-flight. This is minor, but `setLoading(true)` should be called at the start of `load()` (before the `isConnected` check) to ensure consistent behavior.

---

## Correctness of the core fix

The `setupMirroringIfNeeded` implementation itself is correct:
- The no-op guard `if (this.mirroringCleanups.has(group.id)) return` is safe — `mirroringCleanups` is keyed by group ID and the check is synchronous.
- `setupMirroring` is called after the guard, meaning a second call from normal spawn paths (which call `setupMirroring` directly) would still work correctly since `mirroringCleanups.has` would already be true and the public wrapper would no-op.
- The placement before `restoreMcpServersForGroup` is correct — mirroring must be active before any session work begins to capture all messages.

The `TaskConversationRenderer.tsx` text change from "Waiting for agent activity…" to "No messages yet." is appropriate — the previous text implied an active agent when the real issue was missing messages.

PR #715, opened 2026-03-22
feat: enable Bash and full MCP access for Room Agent session

lsm:
- Add Bash to room agent allowed tools so it can run diagnostics (git, gh, sqlite)
- Remove Bash from restricted tools list; keep Write/Edit/NotebookEdit restricted
- Add SettingsManager.getProjectMcpServersConfig() to read full raw MCP configs from project/user settings files
- Pass settingsManager to RoomRuntimeServiceConfig and use it in setupRoomAgentSession to merge project-configured MCP servers (GitHub MCP, etc.) with room-agent-tools
- Update allowedTools in query-options-builder to dynamically include wildcards for all configured MCP servers instead of hardcoded room-agent-tools__*
- Update all related unit tests and add new coverage for the new behavior

### DIFF
--- a/packages/daemon/src/lib/agent/query-options-builder.ts
+++ b/packages/daemon/src/lib/agent/query-options-builder.ts
@@ -185,13 +185,15 @@ export class QueryOptionsBuilder {
 		};
 
 		// ============ Room Session Restrictions ============
-		// Room chat sessions are orchestrators only — they must not
-		// have access to built-in file/shell tools or user-configured MCP servers.
+		// Room chat sessions are orchestrators — they have read tools, Bash for diagnostics,
+		// and explicitly configured MCP servers (room-agent-tools + project MCP servers).
+		// File editing tools (Write/Edit/NotebookEdit) are excluded.
 		if (this.ctx.session.type === 'room_chat') {
 			const roomAllowedBuiltinTools = [
 				'Read',
 				'Glob',
 				'Grep',
+				'Bash',
 				'WebFetch',
 				'WebSearch',
 				'ToolSearch',
@@ -202,7 +204,6 @@ export class QueryOptionsBuilder {
 				'Task',
 				'TaskOutput',
 				'TaskStop',
-				'Bash',
 				'Edit',
 				'Write',
 				'NotebookEdit',
@@ -218,20 +219,25 @@ export class QueryOptionsBuilder {
 				queryOptions.systemPrompt = undefined;
 			}
 
-			// Restrict room chat to a safe, read-oriented built-in tool set.
+			// Restrict room chat to coordinator-appropriate built-in tool set.
 			queryOptions.tools = roomAllowedBuiltinTools;
+
+			// Auto-allow all explicitly configured MCP server tools (room-agent-tools + project MCP servers).
+			const mcpServerWildcards = Object.keys(queryOptions.mcpServers ?? {}).map(
+				(name) => `${name}__*`
+			);
 			queryOptions.allowedTools = [
 				...new Set([
 					...(queryOptions.allowedTools ?? []),
 					...roomAllowedBuiltinTools,
-					'room-agent-tools__*',
+					...mcpServerWildcards,
 				]),
 			];
 
 			queryOptions.disallowedTools = [
 				...new Set([...(queryOptions.disallowedTools ?? []), ...restrictedBuiltinTools]),
 			];
-			// Prevent user-configured MCP servers from being merged in.
+			// Prevent user-configured MCP servers from being merged in (only runtime-injected servers allowed).
 			queryOptions.strictMcpConfig = true;
 			// Skip settings file loading so user's settings.json doesn't inject extra tools.
 			queryOptions.settingSources = [];
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -512,12 +512,12 @@ export class RoomRuntimeService {
 					}
 				}
 
-				// Merge project-configured MCP servers with the room-agent-tools server so the
+				// Merge user/project-configured MCP servers with the room-agent-tools server so the
 				// room agent can use GitHub MCP, etc. for coordination tasks.
 				// room-agent-tools is placed last to ensure it always takes precedence.
-				const projectMcpServers = this.ctx.settingsManager.getProjectMcpServersConfig();
+				const enabledMcpServers = this.ctx.settingsManager.getEnabledMcpServersConfig();
 				roomChatSession.setRuntimeMcpServers({
-					...projectMcpServers,
+					...enabledMcpServers,
 					'room-agent-tools': roomAgentMcpServer,
 				});
 				// Inject the room chat system prompt so the agent knows the proper
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -9,6 +9,7 @@
  */
 
 import type { Room, McpServerConfig, RuntimeState, GlobalSettings } from '@neokai/shared';
+import type { SettingsManager } from '../../settings-manager';
 import type { JobQueueRepository } from '../../../storage/repositories/job-queue-repository';
 import type { JobQueueProcessor } from '../../../storage/job-queue-processor';
 import type { ReactiveDatabase } from '../../../storage/reactive-database';
@@ -50,6 +51,8 @@ export interface RoomRuntimeServiceConfig {
 	defaultModel: string;
 	/** Get current global settings including fallbackModels for auto-fallback on rate limits */
 	getGlobalSettings: () => GlobalSettings;
+	/** Settings manager for reading project-configured MCP servers */
+	settingsManager: SettingsManager;
 	/** Reactive database wrapper for change event emission */
 	reactiveDb: ReactiveDatabase;
 	/**
@@ -509,7 +512,12 @@ export class RoomRuntimeService {
 					}
 				}
 
+				// Merge project-configured MCP servers with the room-agent-tools server so the
+				// room agent can use GitHub MCP, etc. for coordination tasks.
+				// room-agent-tools is placed last to ensure it always takes precedence.
+				const projectMcpServers = this.ctx.settingsManager.getProjectMcpServersConfig();
 				roomChatSession.setRuntimeMcpServers({
+					...projectMcpServers,
 					'room-agent-tools': roomAgentMcpServer,
 				});
 				// Inject the room chat system prompt so the agent knows the proper
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -175,6 +175,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		defaultWorkspacePath: deps.config.workspaceRoot,
 		defaultModel: deps.config.defaultModel,
 		getGlobalSettings: () => deps.settingsManager.getGlobalSettings(),
+		settingsManager: deps.settingsManager,
 		reactiveDb: deps.reactiveDb,
 		jobQueue: deps.jobQueue,
 		jobProcessor: deps.jobProcessor,
--- a/packages/daemon/src/lib/settings-manager.ts
+++ b/packages/daemon/src/lib/settings-manager.ts
@@ -497,6 +497,14 @@ export class SettingsManager {
 			);
 		}
 
+		// Respect per-server allow/deny settings — exclude servers the user has explicitly disallowed.
+		const mcpServerSettings = globalSettings.mcpServerSettings || {};
+		for (const name of Object.keys(result)) {
+			if (mcpServerSettings[name]?.allowed === false) {
+				delete result[name];
+			}
+		}
+
 		return result;
 	}
 
--- a/packages/daemon/src/lib/settings-manager.ts
+++ b/packages/daemon/src/lib/settings-manager.ts
@@ -10,6 +10,7 @@ import { writeFileSync, readFileSync, mkdirSync, existsSync } from 'node:fs';
 import { join, dirname } from 'node:path';
 import { homedir } from 'node:os';
 import type { GlobalSettings, SessionSettings, SettingSource } from '@neokai/shared';
+import type { McpServerConfig } from '@neokai/shared/types/sdk-config';
 
 /**
  * MCP server info from a setting source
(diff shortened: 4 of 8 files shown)


lsm (commented):

### Summary

The implementation correctly enables Bash access and dynamic MCP server configuration for Room Agent sessions. The changes are well-structured across four files with good test coverage. I found two issues worth addressing.

### P1: `getProjectMcpServersConfig` bypasses `mcpServerSettings` (allowed/defaultOn) filtering

**File:** `packages/daemon/src/lib/settings-manager.ts`, lines 459-501

The new `getProjectMcpServersConfig()` reads raw MCP server configs from settings files but does NOT check `mcpServerSettings` for `allowed: false` or `defaultOn: false` status. The existing `listMcpServersFromSources()` + `getDefaultForNewSession()` flow in `tools-config.ts` (lines 50-66) deliberately filters out servers that the user has disabled via the UI.

This means if a user has explicitly disallowed an MCP server (e.g., marked `allowed: false`), the Room Agent will still load and use it. This is a policy bypass — the room agent should respect the same server-level allow/deny settings as regular sessions.

**Fix:** Filter the returned servers against `mcpServerSettings`, excluding any with `allowed === false`.

### P1: Room runtime service test does not verify actual MCP merging behavior

**File:** `packages/daemon/tests/unit/room/room-runtime-service.test.ts`, lines 153-171

The "settingsManager integration" tests only verify that the mock returns expected values when called directly. They don't test the actual integration point — that `setupRoomAgentSession` calls `setRuntimeMcpServers` with the merged result of `{ ...projectMcpServers, 'room-agent-tools': roomAgentMcpServer }`. The tests essentially test the mock itself, not the service behavior.

This gap means the core new behavior (project MCP servers being merged with room-agent-tools, with room-agent-tools taking precedence) has no automated verification.

### What looks good

- **Tool allowlist changes** (`query-options-builder.ts`): Bash correctly moved to allowed list; Write/Edit/NotebookEdit remain in restricted list. The dynamic wildcard generation from `queryOptions.mcpServers` keys is clean and correct.
- **MCP server merging** (`room-runtime-service.ts`): Using spread with `room-agent-tools` last ensures it takes precedence over any project server with the same name.
- **Settings manager** (`settings-manager.ts`): File reading logic mirrors `listMcpServersFromSources` with proper error handling (try/catch, existsSync, JSON parse). Source ordering (user < project < local) matches precedence expectations.
- **Settings manager tests** (`settings-manager.test.ts`): Good coverage of happy path, missing files, no mcpServers field, malformed JSON, and source filtering.
- **Query options builder tests**: Good coverage of the allowlist, disallowlist, and dynamic wildcard generation.

The `mcpServerSettings` filtering gap (P1) could cause servers the user has explicitly disabled to be loaded into the Room Agent. This should be addressed before merge.

lsm (commented):

---

### Summary

This PR enables Bash tool access and full MCP server forwarding for the Room Agent session. The overall architecture is sound and the happy path is correct, but there are several issues worth addressing before merging.

---

### P0 - Blocking Issues

None identified.

---

### P1 - Should Fix

#### 1. Misleading function name: `getProjectMcpServersConfig` reads from ALL three sources, not just project

`packages/daemon/src/lib/settings-manager.ts` lines 459-501

The function is named `getProjectMcpServersConfig` but it reads from user (`~/.claude/settings.json`, `~/.mcp.json`), project (`.claude/settings.json`, `.mcp.json`), and local (`.claude/settings.local.json`) sources — whatever is enabled in `settingSources`. The doc comment says so explicitly, but the name implies only project scope. This naming mismatch will mislead future readers. The name should reflect the true scope, e.g. `getAllEnabledMcpServersConfig` or `getEnabledMcpServersConfig`.

This naming confusion propagates to callers: in `setupRoomAgentSession` (line 518), `projectMcpServers` is used as the variable name for what is actually all-sources MCP config.

#### 2. Local settings file is a privileged bypass path

`packages/daemon/src/lib/settings-manager.ts` lines 493-498

`getProjectMcpServersConfig` reads `mcpServers` from `.claude/settings.local.json` when the `local` source is enabled (which is the default). The room session is configured with `strictMcpConfig = true` and `settingSources = []` precisely to block the SDK from auto-loading servers — but then `setupRoomAgentSession` manually merges all enabled MCP servers back in, including from `settings.local.json`. A user (or another process) can write to `.claude/settings.local.json` to inject an arbitrary MCP server into the room agent's tool set. This is a security-relevant oversight since `settings.local.json` is a file the daemon itself also writes to via `writeFileOnlySettings`.

The intent of `strictMcpConfig = true` is to prevent unsanctioned server injection. The current implementation partially nullifies that intent. Consider filtering the result to only include project-scope servers, or documenting this as an accepted design trade-off with a comment explaining why `local` is included.

#### 3. No test for user-source reads in `getProjectMcpServersConfig`

`packages/daemon/tests/unit/session/settings-manager.test.ts` — `describe('getProjectMcpServersConfig')`

The test suite (lines 725-817) covers empty, project `.claude/settings.json`, project `.mcp.json`, source filtering (disabling project source), missing `mcpServers` field, and malformed JSON. But there is no test for:
- Reading from the user source (`~/.claude/settings.json` / `~/.mcp.json`)
- Source priority / override behavior when the same server name appears in multiple sources (e.g., project overrides user, local overrides project)

Since the default `settingSources` includes `user`, user-source reads are exercised in production. Their absence in tests leaves a real coverage gap.

#### 4. `room-runtime-service.test.ts` `settingsManager integration` tests don't test actual integration

`packages/daemon/tests/unit/room/room-runtime-service.test.ts` lines 153-171

The two new tests in the `settingsManager integration` block do not exercise `setupRoomAgentSession` or the merging logic at all — they only test that the mock object has a callable method. The test named `"should return project MCP servers config when configured"` directly calls `mockSettingsManager.getProjectMcpServersConfig()` without invoking any code path that uses it in `RoomRuntimeService`. These tests provide false confidence and should be removed or replaced with tests that actually verify the MCP merging behavior in `setupRoomAgentSession`.

---

### P2 - Meaningful Improvements

#### 5. MCP wildcard generation in `query-options-builder.ts` reads session config at query-build time

`packages/daemon/src/lib/agent/query-options-builder.ts` lines 226-233

The wildcard generation reads `queryOptions.mcpServers` (derived from `session.config.mcpServers`) at query-build time. The `setRuntimeMcpServers` call in `setupRoomAgentSession` is async — it fires a void promise that resolves after `getSessionAsync`. If `build()` is called before the promise resolves, the MCP wildcards will be empty (or only include whatever was already in `session.config.mcpServers`). This is a pre-existing race condition, not introduced by this PR, but the PR makes it slightly more impactful since the wildcards now cover more servers.

Consider adding a comment warning about this timing dependency.

#### 6. `getProjectMcpServersConfig` is synchronous but reads files on the hot path

`packages/daemon/src/lib/settings-manager.ts` lines 459-501

The function uses synchronous `existsSync` and `readFileSync` calls in a code path that runs each time a room's session is set up. For multiple rooms starting concurrently (e.g., `initializeExistingRooms` starts all rooms in parallel), this causes multiple synchronous file reads in parallel threads of `Promise.all`. This is not a correctness issue, but it's a pattern inconsistent with async I/O practices. A minor concern for large deployments with many rooms.

---

### P3 - Nits

#### 7. Comment says `room-agent-tools is placed last to ensure it always takes precedence` (line 517)

The comment is accurate but uses the wrong language. `room-agent-tools` takes precedence over project MCP servers because it comes later in the spread. This is correct. However, this means if a user defines a project MCP server named `room-agent-tools`, the daemon's internal server will silently override it without any warning. A defensive check or log warning would improve debuggability.

#### 8. Test at line 154 in `room-runtime-service.test.ts`: `"should expose getProjectMcpServersConfig on the settings manager"`

This test verifies a mock, not real behavior. While not harmful, it adds test noise without value.

---

### What Is Correct

- The `Bash` addition to `roomAllowedBuiltinTools` and removal from `restrictedBuiltinTools` is correct. Write/Edit/NotebookEdit remain restricted.
- The dynamic wildcard generation from `Object.keys(queryOptions.mcpServers ?? {})` is the right approach — it automatically handles any combination of servers.
- The spread-merge order `{ ...projectMcpServers, 'room-agent-tools': roomAgentMcpServer }` correctly ensures the daemon's server is never overridden.
- `settingsManager` is correctly passed via `RoomRuntimeServiceConfig` and the `rpc-handlers/index.ts` wiring is correct.
- The `getProjectMcpServersConfig` source filtering using `globalSettings.settingSources` with a sane default is correct and consistent with `listMcpServersFromSources`.

lsm (commented):

### Re-review: Previous P1 Issues

**P1 #1 (getProjectMcpServersConfig bypassed mcpServerSettings filtering)** -- **Resolved.** The fix commit (`d69f4c4`) adds an explicit loop after merging all sources that deletes any server where `mcpServerSettings[name]?.allowed === false`. The filtering logic is correct: servers with `allowed: true` or no setting at all are kept; only explicit denials are removed. Two new test cases (`should exclude servers with allowed === false` and `should include servers with allowed === true or no setting`) directly validate this behavior.

**P1 #2 (Integration tests were hollow)** -- **Resolved.** The `setupRoomAgentSession MCP merging` test suite in `room-runtime-service.test.ts` now uses a real in-memory Bun SQLite database and captures the `room.created` event handler from the DaemonHub mock. Tests trigger actual `roomCreatedHandler` calls and verify the `setRuntimeMcpServers` spy receives the correctly merged server map. Three substantive tests cover: (1) project servers merged with room-agent-tools, (2) room-agent-tools takes precedence over same-named project server, (3) only room-agent-tools when no project servers exist.

### New Findings

**No blocking issues found.** The implementation is solid:

1. **Correctness**: Bash is moved to the allowed tools list for room_chat sessions; Write/Edit/NotebookEdit remain in the restricted list. The dynamic MCP wildcard generation (`Object.keys(queryOptions.mcpServers ?? {}).map(name => name + '__*')`) correctly auto-allows all tools from any configured MCP server, replacing the previous hardcoded `room-agent-tools__*`.

2. **Security**: `strictMcpConfig: true` and `settingSources: []` prevent user-configured MCP servers from being injected outside the controlled `setRuntimeMcpServers` path. The `allowed === false` filtering ensures denied servers are excluded.

3. **Test coverage**: 137 tests pass across the 4 affected test files. The query-options-builder tests verify Bash inclusion, Write/Edit/NotebookEdit exclusion, dynamic MCP wildcards, strict MCP config, and empty settingSources for room_chat sessions.

4. **Architecture**: The `settingsManager` dependency is cleanly threaded through `RoomRuntimeServiceConfig` and used only during `setupRoomAgentSession`. The merge order (`...projectMcpServers, 'room-agent-tools': roomAgentMcpServer`) correctly ensures the internal room-agent-tools server always takes precedence.

### Minor Observation (P3)

The `getProjectMcpServersConfig` method includes the `local` source (`.claude/settings.local.json`) which is the file the daemon itself writes via `writeFileOnlySettings`. This means the daemon could theoretically inject MCP servers into room agent sessions through the local settings file it controls. This is a low-risk concern since the daemon is trusted code, but excluding the `local` source would make the trust boundary cleaner. (I note the uncommitted working tree changes appear to address this -- excluding `local` and renaming the method to `getEnabledMcpServersConfig`.)

**Recommendation: APPROVE**

All previously identified P1 issues have been properly resolved. The implementation correctly enables Bash and full MCP access for room agent sessions while maintaining appropriate security boundaries.

lsm (commented):

**Recommendation: APPROVE**

---

### Second-Iteration Review Summary

All four P1 issues from the previous review have been resolved. The implementation is now correct, well-tested, and secure.

---

### Resolved Issues (Previously P1)

**1. Function renamed from `getProjectMcpServersConfig` to `getEnabledMcpServersConfig`**
The name now accurately reflects the scope (user + project sources). The variable in `room-runtime-service.ts` was also updated from `projectMcpServers` to `enabledMcpServers`. The doc comment is clear and accurate.

**2. Local source bypass addressed**
`getEnabledMcpServersConfig` now intentionally excludes the `local` source (`.claude/settings.local.json`) with a clear doc comment explaining the security rationale: the daemon itself writes to this file via `writeFileOnlySettings`, which would create a self-injection bypass vector. The code does not have an `if (enabledSources.includes('local'))` block, so even if 'local' appears in `settingSources`, it is silently and safely ignored.

**3. User-source code path now tested**
The new tests at lines 867-927 of `settings-manager.test.ts` cover reading from user `settings.json`, verifying that `settings.local.json` is excluded, and confirming that project source overrides user source for the same server name. All three gaps identified in the previous review are now covered.

**4. Integration tests now verify actual merging behavior**
The `setupRoomAgentSession MCP merging` describe block in `room-runtime-service.test.ts` properly exercises the real code path. Tests trigger `room.created` events, wait for the async promise to settle, and assert on the actual `setRuntimeMcpServers` call arguments. The three tests cover: project servers merged with room-agent-tools, room-agent-tools precedence over conflicting project servers, and fallback to room-agent-tools only when no project servers exist.

---

### Remaining Minor Issue (P3 - Nit)

**Fallback default at `settings-manager.ts` line 465 still includes 'local'**

```ts
const enabledSources = globalSettings.settingSources || ['user', 'project', 'local'];
```

The fallback array includes `'local'` but since there is no `if (enabledSources.includes('local'))` block, it is harmless — 'local' is never processed. However, the fallback could be `['user', 'project']` to better match the documented behavior of this function. This is cosmetic; the behavior is correct.

---

### What Looks Good

- All tests pass (12/12 for `getEnabledMcpServersConfig`, 13/13 for room-runtime-service, 66/66 for query-options-builder)
- `strictMcpConfig = true` + `settingSources = []` still prevents the SDK from loading arbitrary user MCP servers; `getEnabledMcpServersConfig` provides an explicit, controlled allow-list via `setRuntimeMcpServers`
- The spread-merge order `{ ...enabledMcpServers, 'room-agent-tools': roomAgentMcpServer }` correctly ensures the daemon's internal server always takes precedence
- `mcpServerSettings.allowed === false` filtering is correctly applied before returning
- The `TEST_USER_SETTINGS_DIR` test isolation pattern is consistent with the existing `getMcpServers` approach and correctly isolates tests from real `~/.claude` state

lsm (commented):

### Third-iteration review (final polish commit)

Reviewed the latest commit `384295f3d` which addresses the P3 nits from the previous two approved reviews.

**Changes verified:**

1. **Rename `getProjectMcpServersConfig` -> `getEnabledMcpServersConfig`**: Clean rename across all 5 files. No stale references to the old name remain anywhere in the codebase.

2. **Exclusion of `local` source**: The `settings.local.json` read block is removed with a clear inline comment and a thorough doc comment explaining the bypass vector rationale. This is a meaningful security improvement -- the daemon writes to `settings.local.json` via `writeFileOnlySettings`, so including it would allow the daemon to inject MCP servers into room agent sessions.

3. **New tests added (3 tests)**:
   - `should read MCP servers from user settings.json` -- verifies user-source reading works
   - `should not include servers from settings.local.json (local source excluded)` -- directly tests the exclusion
   - `project source overrides user source for same server name` -- validates merge priority

   All three tests use the existing `TEST_USER_SETTINGS_DIR` isolation pattern established in the test setup, consistent with the rest of the test file.

4. **Updated doc comment**: Accurately reflects the new behavior -- lists user and project sources, documents the local exclusion rationale, and describes the override order.

5. **Caller update in `room-runtime-service.ts`**: Variable renamed from `projectMcpServers` to `enabledMcpServers`, comment updated to say "user/project-configured" instead of "project-configured".

No issues found. The commit is a clean cosmetic/naming improvement with good test coverage for the new behavior. All previous nits are resolved.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

---

### Third-Iteration Review (Final Polish Commit `384295f3d`)

All four changes described in the task are correctly implemented. No new issues were introduced.

---

### Changes Verified

**1. Rename `getProjectMcpServersConfig` → `getEnabledMcpServersConfig`**
Complete. Zero stale references to the old name remain anywhere in the codebase — checked across all five modified files and the full packages tree. The caller in `room-runtime-service.ts` line 518 uses the new name, and the local variable was updated from `projectMcpServers` to `enabledMcpServers` at line 518--519.

**2. Local source excluded with doc comment**
The `if (enabledSources.includes('local'))` block is gone and replaced with an inline NOTE comment. The method-level JSDoc explicitly lists the rationale: the daemon writes to `settings.local.json` via `writeFileOnlySettings`, so including it would let the daemon self-inject servers into room agent sessions. The security boundary is properly documented.

**3. New tests**
Three tests added at `packages/daemon/tests/unit/session/settings-manager.test.ts` lines 867--927:
- `should read MCP servers from user settings.json` -- covers the user source read path using `TEST_USER_SETTINGS_DIR` isolation
- `should not include servers from settings.local.json (local source excluded)` -- directly validates the exclusion
- `project source overrides user source for same server name` -- confirms merge priority

All three use the `TEST_USER_SETTINGS_DIR` isolation pattern established in the shared `beforeEach`, consistent with the rest of the file. Tests are substantive — they write real files, call the real method, and assert on real output.

**4. Caller variable rename**
`room-runtime-service.ts` line 514 comment updated to "user/project-configured"; line 518 local var is `enabledMcpServers`. Clean.

---

### Remaining P3 Nit (Not Addressed — Still Harmless)

`packages/daemon/src/lib/settings-manager.ts` line 465:

```ts
const enabledSources = globalSettings.settingSources || ['user', 'project', 'local'];
```

The fallback default still includes `'local'`. Since there is no `if (enabledSources.includes('local'))` block in this function, `'local'` in the fallback has no effect — it is dead. The fallback could be `['user', 'project']` to match the documented scope of this function. However, this nit was explicitly called out in my previous review and was not addressed in this polish commit. That is acceptable: the behavior is provably correct, no test can distinguish the two fallback values for this function, and changing the fallback without a comment could silently diverge from the `DEFAULT_GLOBAL_SETTINGS.settingSources` value (which also includes `'local'`). Leaving it as-is is a reasonable call.

---

### Overall Assessment

The final polish commit resolves all named nits from the previous two approved reviews (rename, local exclusion, user-source tests, source priority tests, caller variable rename) without introducing any regressions. The implementation is correct, the security rationale is documented, and the test coverage is solid. Ready to merge.

PR #716, opened 2026-03-22
fix: prevent full conversation reload when sending a message in TaskView

lsm:
Sending a message used to call onMessageSentWithReload() which bumped
conversationKey, remounting TaskConversationRenderer and triggering a
full re-fetch + loading flash. The LiveQuery subscription already
appends new messages via liveQuery.delta events, so no remount is needed.

- Rename HumanInputAreaProps.onMessageSentWithReload → onMessageSent
- Pass a no-op onMessageSent from TaskView (LiveQuery handles the append)
- Keep conversationKey bumps for approve/reject (intentional full refresh)
- Add unit test verifying the DOM node is not replaced on message send
- Add E2E regression test verifying pre-existing messages stay visible
  while user types and attempts to send a message

### DIFF
--- a/packages/e2e/tests/features/task-message-streaming.e2e.ts
+++ b/packages/e2e/tests/features/task-message-streaming.e2e.ts
@@ -7,6 +7,17 @@
  * Tests:
  * - Messages from the initial LiveQuery snapshot appear in TaskView on load
  * - Switching between two tasks shows correct messages for each task
+ * - Pre-existing messages remain visible after user types and sends a message
+ *   (regression test for: task view reloads all messages when sending a new message)
+ *
+ * Note on "no-reload-on-send" testing: The full round-trip (message actually delivered
+ * via liveQuery.delta) requires a real agent session and cannot be triggered purely
+ * through the UI in an integration-free E2E test.  The component-level guarantee
+ * (TaskConversationRenderer DOM node is not replaced on send) is exercised by the
+ * unit test in packages/web/src/components/room/TaskView.test.tsx
+ * ("TaskView — no reload on message send (bug regression)").
+ * The E2E test below verifies the UI-observable side: messages visible BEFORE a
+ * send attempt remain visible AFTER the send attempt (no loading flash or wipe).
  *
  * Note on "live delta without refresh" testing: injecting a message after page
  * navigation requires calling hub.request() inside the test body, which violates
@@ -210,5 +221,63 @@ test.describe('TaskView — Message Streaming via LiveQuery', () => {
 			await expect(page.locator('text=Message for Task A only')).toBeVisible({ timeout: 10000 });
 			await expect(page.locator('text=Message for Task B only')).not.toBeVisible();
 		});
+	});
+});
+
+// ── Test 3: no reload on message send (regression) ──────────────────────────
+
+test.describe('no loading flash when user sends a message (regression)', () => {
+	test.use({ viewport: DESKTOP_VIEWPORT });
+
+	let roomId = '';
+	let taskId = '';
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/');
+		await page
+			.getByRole('button', { name: 'New Session', exact: true })
+			.waitFor({ timeout: 10000 });
+
+		// Create room, task, group with pre-existing messages — infrastructure only
+		({ roomId, taskId } = await createRoomWithTask(page, 'E2E No-Reload Task'));
+		await createGroupWithMessages(page, taskId, roomId, [
+			'Pre-existing message alpha',
+			'Pre-existing message beta',
+		]);
+	});
+
+	test.afterEach(async ({ page }) => {
+		await deleteRoom(page, roomId);
+		roomId = '';
+		taskId = '';
+	});
+
+	test('pre-existing messages remain visible while user types and attempts to send', async ({
+		page,
+	}) => {
+		// Navigate to the task view — LiveQuery snapshot delivers the pre-existing messages
+		await page.goto(`/room/${roomId}/task/${taskId}`);
+		await expect(page.locator('text=E2E No-Reload Task')).toBeVisible({ timeout: 10000 });
+
+		// Both pre-existing messages must be visible via initial snapshot
+		await expect(page.locator('text=Pre-existing message alpha')).toBeVisible({ timeout: 10000 });
+		await expect(page.locator('text=Pre-existing message beta')).toBeVisible({ timeout: 10000 });
+
+		// Type a message in the human input area
+		const textarea = page.getByTestId('input-textarea-field');
+		await textarea.fill('Hello, please continue working');
+
+		// Pre-existing messages must still be visible while typing (no wipe on input)
+		await expect(page.locator('text=Pre-existing message alpha')).toBeVisible();
+		await expect(page.locator('text=Pre-existing message beta')).toBeVisible();
+
+		// Click send — the RPC will fail (no real session), but the UI must NOT
+		// clear the conversation pane or flash a loading state at any point
+		await page.getByTestId('input-textarea-send').click();
+
+		// After the send attempt, pre-existing messages must still be visible —
+		// this verifies the conversationKey is NOT bumped on message send
+		await expect(page.locator('text=Pre-existing message alpha')).toBeVisible({ timeout: 5000 });
+		await expect(page.locator('text=Pre-existing message beta')).toBeVisible({ timeout: 5000 });
 	});
 });
--- a/packages/e2e/tests/features/task-message-streaming.e2e.ts
+++ b/packages/e2e/tests/features/task-message-streaming.e2e.ts
@@ -263,8 +263,8 @@ test.describe('no loading flash when user sends a message (regression)', () => {
 		await expect(page.locator('text=Pre-existing message alpha')).toBeVisible({ timeout: 10000 });
 		await expect(page.locator('text=Pre-existing message beta')).toBeVisible({ timeout: 10000 });
 
-		// Type a message in the human input area
-		const textarea = page.getByTestId('input-textarea-field');
+		// Type a message in the human input area (no data-testid on <textarea> in prod)
+		const textarea = page.locator('textarea');
 		await textarea.fill('Hello, please continue working');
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Recommendation: APPROVE

This is a clean, well-scoped fix. The root cause was correctly identified and the fix is minimal — changing `onMessageSentWithReload` (which bumped `conversationKey`) to a no-op `onMessageSent` for the message-send path, while preserving the intentional `conversationKey` bumps for `task.approve` and `task.reject`.

**What I verified:**

1. **Correctness of the fix** (`TaskView.tsx`): The `onMessageSent={() => {}}` no-op at line 1409 correctly prevents the `conversationKey` bump on message send. The `setConversationKey((k) => k + 1)` calls at lines 1030 (approve) and 1046 (reject) are preserved, which is correct — those actions change the conversation state server-side and need a full refresh.

2. **No stale references**: Grepped for `onMessageSentWithReload` across the codebase — only appears in a comment in the test file explaining the old behavior. No code references remain.

3. **LiveQuery handles new messages**: The PR description and code comments correctly note that the LiveQuery subscription already appends new messages via `liveQuery.delta` events, making the `conversationKey` bump on send redundant.

4. **Unit tests** (`TaskView.test.tsx`): Two well-structured regression tests:
   - First test captures the DOM node reference before send and verifies identity (`===`) after — directly tests the bug.
   - Second test verifies approve still causes a remount while send does not — tests both the fix and the preserved intentional behavior.

5. **E2E test** (`task-message-streaming.e2e.ts`): The regression test creates pre-existing messages via RPC infrastructure helpers (acceptable per CLAUDE.md), navigates to the task, fills and clicks send, then asserts the pre-existing messages remain visible. The comment block honestly acknowledges the limitation that the full round-trip cannot be tested in E2E without a real agent session.

**Minor observation (P3):**

- The `onMessageSent` prop on `HumanInputAreaProps` is now always a no-op. If no future use is planned, it could be removed entirely. However, keeping it as a callback slot is reasonable for extensibility (e.g., analytics, focus management), and the interface remains clean. No action needed.

No issues found. The change is minimal, well-tested, and correctly targeted.

lsm (commented):

---

### Summary

The core fix in `TaskView.tsx` is correct and minimal. The rename from `onMessageSentWithReload` to `onMessageSent` and replacing the `setConversationKey((k) => k + 1)` call with `() => {}` directly solves the reported bug. The unit tests are well-structured and actually verify the no-remount invariant. However, the E2E regression test has a critical flaw: it uses test IDs that only exist in the unit test mock, not in the real application.

---

### P0 — E2E test uses wrong test IDs (test will fail or never exercise the send path)

**File:** `packages/e2e/tests/features/task-message-streaming.e2e.ts`, lines 267 and 276

```ts
const textarea = page.getByTestId('input-textarea-field');
await textarea.fill('Hello, please continue working');
// ...
await page.getByTestId('input-textarea-send').click();
```

The real `InputTextarea` component (`packages/web/src/components/InputTextarea.tsx`) does not have `data-testid="input-textarea-field"` or `data-testid="input-textarea-send"`. Those test IDs are defined exclusively in the mock inside `TaskView.test.tsx` (lines 147–158).

The actual component uses:
- `data-testid="send-button"` on the submit button (line 246 of `InputTextarea.tsx`)
- No test ID on the `<textarea>` element at all

All other E2E tests in the repository interact with the textarea via `page.locator('textarea')` and click the send button via `[data-testid="send-button"]` — e.g. `packages/e2e/tests/core/message-input.e2e.ts`, `packages/e2e/tests/serial/error-scenarios.e2e.ts`, etc.

Concretely, `page.getByTestId('input-textarea-send').click()` will throw a timeout error in CI because the element does not exist. The E2E test therefore does not validate the bug regression at all. The correct selectors should be:

```ts
const textarea = page.locator('textarea');
await textarea.fill('Hello, please continue working');
// ...
await page.locator('[data-testid="send-button"]').click();
```

---

### P2 — `onMessageSent` prop is dead code; consider removing it entirely

**File:** `packages/web/src/components/room/TaskView.tsx`, lines 72–73 and 1409

The `onMessageSent` prop on `HumanInputAreaProps` is now always passed `() => {}` from `TaskView`. The prop interface and the call-site inside `sendMessage` exist only to invoke a no-op. Since `HumanInputArea` is a module-private function (not exported), there is no external caller that could pass a meaningful callback.

This leaves a misleading hook in the interface ("Called after a message is successfully sent") that implies something happens when it doesn't. Either:
- Remove the prop entirely and inline the no-op (simplest), or
- Keep it with a note that it is intentionally reserved for future use — but add `_onMessageSent` naming convention or a comment making clear it is currently unused

The current state will likely trigger a knip "unused export" warning in CI if the interface is ever checked separately.

---

### P3 — Unit test `beforeEach` resets signals that are already reset at module level

**File:** `packages/web/src/components/room/TaskView.test.tsx`, lines 2706–2711

The new `describe` block's `beforeEach` resets `_draftContentSignal.value`, `_draftRestoredSignal.value`, `mockSetMessageText`, and `mockClearDraft`. These are also reset at the top of every other `describe` block's `beforeEach` — but the pattern is copy-pasted rather than shared. A helper `resetCommonMocks()` function would eliminate the duplication. Minor, but worth noting as the test file grows.

---

### What is correct

- The core fix in `TaskView.tsx` is accurate: removing the `conversationKey` bump on `sendHumanMessage` is the right solution and does not affect the approve/reject refresh paths (lines 1030 and 1046 are untouched).
- The unit tests at lines 2718–2748 and 2751–2802 correctly assert DOM node identity before and after send and approve, which is the right way to test the no-remount invariant.
- The comment block in the E2E file (lines 11–27) honestly documents the limitation of the E2E approach, which is good practice.
- TypeScript type-check passes cleanly.

lsm (commented):

### Recommendation: APPROVE

The fix correctly addresses the bug where sending a message in TaskView caused a full conversation reload. The changes are clean and well-scoped.

### What was done

1. **Root cause fix** (`TaskView.tsx`): Removed the `onMessageSentWithReload` prop from `HumanInputArea` and its call site in `TaskView`. Previously, every `sendMessage` call bumped `conversationKey`, which remounted `TaskConversationRenderer` and triggered a full re-fetch. Now `conversationKey` bumps are preserved only for approve (line 1021) and reject (line 1037), which is the intended behavior.

2. **Dead prop removal** (`TaskView.tsx`): The `onMessageSent` / `onMessageSentWithReload` prop was removed from `HumanInputAreaProps`, the destructuring, and all call sites. No leftover references exist outside a descriptive comment in the test file.

3. **Unit tests** (`TaskView.test.tsx`): Two regression tests added:
   - Verifies DOM node identity is preserved after `sendMessage` (no remount)
   - Verifies approve still triggers a remount (conversationKey bump preserved)

4. **E2E test** (`task-message-streaming.e2e.ts`): Regression test verifies pre-existing messages remain visible after typing and clicking send. Uses correct production selectors (`page.locator('textarea')` and `[data-testid="send-button"]`), matching the real `InputTextarea` component.

### Findings

No issues found. The implementation is correct, complete, and well-tested. The E2E test properly documents its limitations (send RPC will fail without a real session, but the key behavioral guarantee — no conversation wipe — is still verified). The `conversationKey` mechanism remains intact for approve/reject flows where a full refresh is intentional.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

Both issues flagged in the previous review have been correctly addressed. The P0 (wrong E2E test selectors) and P2 (dead `onMessageSent` prop) are cleanly resolved in commit `c36a1ff`. The implementation is correct and the tests accurately verify the regression guard.

---

### P0 fix verified: E2E test selectors are now correct

**File:** `packages/e2e/tests/features/task-message-streaming.e2e.ts`, lines 267 and 276

The test now uses:
- `page.locator('textarea')` — matches the single `<textarea>` inside `InputTextarea` in the task view
- `page.locator('[data-testid="send-button"]').click()` — matches the real send button (line 246 of `InputTextarea.tsx`)

Both selectors exist in the real production component and are consistent with how other E2E tests in the repo interact with the input area. The page analysis confirms only one textarea is visible in the task view context (modals return null when closed, the tab content is replaced by `TaskView` when `taskViewId` is set, the sidebar has no textareas).

---

### P2 fix verified: dead prop completely removed

**File:** `packages/web/src/components/room/TaskView.tsx`

The `onMessageSent` prop has been removed from:
- `HumanInputAreaProps` interface (lines 67–72)
- `HumanInputArea` function signature (line 79)
- The `sendMessage` call site (previously line 125)
- The `TaskView` render call site (previously line 1397–1398)

No references to `onMessageSent` or `onMessageSentWithReload` remain in production code. The only remaining reference is a comment in the test file explaining the historical context.

---

### Core fix remains correct

The bug fix itself is minimal and correct:
- `sendMessage` in `HumanInputArea` no longer calls `setConversationKey((k) => k + 1)` on success
- The `approve` and `reject` paths still bump `conversationKey` (lines 1021 and 1037 of `TaskView.tsx`) — intentional full refresh preserved
- The `key` prop on `TaskConversationRenderer` (`${group.id}-${conversationKey}`) still works correctly for the approve/reject remount

---

### Unit tests: valid and precise

The two new unit tests in `packages/web/src/components/room/TaskView.test.tsx` (lines 2719–2749 and 2752–2803):

1. **First test**: Sends a message and asserts `getByTestId('conversation')` returns the exact same DOM node object before and after — this correctly verifies no remount occurred.
2. **Second test**: Verifies the symmetry — normal send preserves the node, but clicking approve replaces it with a new DOM node (conversationKey bumped). The ActionBar `[data-testid="action-bar-primary"]` selector correctly maps to the real `ActionBar` component.

The mock `TaskConversationRenderer` renders `<div data-testid="conversation" />` and Preact creates a new DOM element on remount, so the `toBe`/`not.toBe` assertions on DOM node identity are a reliable way to detect remounting.

---

### Remaining nit (P3 — not blocking)

The new `beforeEach` block (lines 2700–2713) duplicates the mock reset pattern from earlier `describe` blocks rather than sharing a `resetCommonMocks()` helper. This was noted in the prior review as P3 and remains unaddressed. Not a blocker given the test file's established pattern.

PR #737, opened 2026-03-23
chore: update @anthropic-ai/claude-agent-sdk to 0.2.81

lsm:
## Summary
- Updates `@anthropic-ai/claude-agent-sdk` from `0.2.77` to `0.2.81`
- Ran `make sync-sdk-types` to regenerate types
- Fixed breaking changes from SDK update:
  - New `SDKUserMessageReplay` type added to the `SDKMessage` union
  - `SDKUserMessage` gained `timestamp?: string` (ISO string), conflicting with neokai's DB-injected `timestamp: number`
  - Fixed by using type cast in two locations where DB epoch-ms timestamp overrides SDK ISO string timestamp
  - Updated `DaemonEventMap['sdk.message']` type to use plain `SDKMessage` with a comment explaining the timestamp duality
  - Added type cast in `sdk-message-repository.ts` return value
- Also adds `@public` JSDoc tag to `validateStepChannels` to suppress a pre-existing knip warning for this test-only export

## Test plan
- [ ] TypeScript typecheck passes
- [ ] Daemon unit tests pass (7239 pass, 8 skip, 0 fail)


lsm (commented):

**Recommendation: APPROVE**

The implementation correctly handles the SDK 0.2.77→0.2.81 upgrade. The breaking change analysis is accurate and the fixes are minimal and targeted.

---

### What changed and why

The SDK update added `timestamp?: string` (ISO 8601) to `SDKUserMessage` and `SDKUserMessageReplay`. This conflicted with neokai's pre-existing pattern of overwriting `timestamp` with a numeric epoch-ms value from the DB layer. The PR resolves the compile error in three places:

1. **`sdk-message-repository.ts`** — The `getMessagesByStatus` return now carries an explicit `as Array<SDKMessage & { dbId: string; timestamp: number }>` cast so callers (e.g. `session-handlers.ts`) still get a typed `timestamp: number`.
2. **`sdk-message-handler.ts`** (two emit sites) — The spread `{ ...sdkMessage, timestamp: consumedAt }` can no longer satisfy `SDKMessage` directly because `consumedAt` is `number` and the field is `string`. The `as unknown as SDKMessage` double-cast is the correct mechanical fix; the inline comment explains the duality clearly.
3. **`daemon-hub.ts`** — The event map type for `sdk.message` drops the old intersection `SDKMessage & { timestamp?: number }` and uses plain `SDKMessage`. The old intersection is no longer viable now that the base type has `timestamp?: string`.

The `@public` JSDoc tag on `validateStepChannels` is a pre-existing knip warning silenced correctly.

---

### Analysis

**The casts are safe given current consumers.** The only subscribers to `'sdk.message'` on `DaemonHub` are `room-runtime.ts` (rate-limit detection via `JSON.stringify(event.message)`, never reads `.timestamp`) and `sdk-message-handler.ts` itself. No subscriber interprets `message.timestamp` as an ISO string, so the runtime lie (`number` behind a `string?` type) does not cause functional breakage today.

**The `isSDKUserMessage` return type annotation is imprecise** — it claims `msg is Extract<SDKMessage, { type: "user" }>` which technically includes `SDKUserMessageReplay`, but the runtime guard excludes `isReplay: true`. This is a pre-existing condition in the `dev` branch and is not introduced by this PR.

**No test changes were needed** — the existing tests in `sdk-message-handler.test.ts` already use `isReplay: true` shaped objects, and none assert on the type of the `timestamp` field in the emitted `sdk.message` event.

**Removed SDK fields** (`queued_to_running` status in `AgentOutput`, `resultWasTruncated` in `FileReadOutput`, `resumeFrom` in `AgentInput`) are not referenced anywhere in application code outside the generated declaration files, so their removal is safe.

---

### Minor observations (no blocking issues)

- **P2 — Residual type duality comment.** `daemon-hub.ts` line 54–57 explains the `timestamp` duality in prose, but the event type itself no longer encodes the `number` override. If a future subscriber tries to access `event.message.timestamp` they will see `string | undefined` and never know a `number` can appear at runtime. Consider keeping a narrow intersection like `SDKMessage & { timestamp?: number | string }` to make the polymorphism visible at the call site, rather than erasing it entirely. This is a maintainability concern, not a correctness bug given current consumers.

- **P3 — `as unknown as SDKMessage` appears twice.** Both cast sites in `sdk-message-handler.ts` are identical patterns (lines 408 and 448). This is the right approach for now, but if a third emit site is added in the future it will be easy to miss. A narrow helper type or a local `toSDKMessageForEmit` wrapper would make this pattern self-documenting. Not required for this PR.

---

Overall the change is correct, the scope is appropriately narrow, and the 7,239 passing unit tests give reasonable confidence. The main trade-off (erasing the numeric timestamp from the event type) is documented and acceptable given that no current consumer reads it.

lsm (commented):

---

### P2 Fix Assessment: Timestamp Duality in `sdk.message` Event Type

The fix adequately addresses the previously-flagged P2 concern. The change to `daemon-hub.ts` line 58 now correctly encodes the timestamp polymorphism:

```ts
// Before (erased the string form)
'sdk.message': { sessionId: string; message: SDKMessage & { timestamp?: number } };

// After (makes both forms visible)
'sdk.message': { sessionId: string; message: SDKMessage & { timestamp?: number | string } };
```

This is the right shape. `SDKUserMessage.timestamp` is declared as `string` in the SDK (`sdk.d.ts` line 2402), and the daemon DB layer overrides it with a `number` (epoch-ms). The intersection now surfaces both possibilities to future subscribers instead of erasing the string form.

The updated comment block is thorough and informative. Future maintainers will understand why both forms exist.

### Remaining Minor Observations

**P3 — Cast target type does not match the event type** (`packages/daemon/src/lib/agent/sdk-message-handler.ts`, lines 409 and 450)

The two emit sites use `as unknown as SDKMessage` rather than `as unknown as SDKMessage & { timestamp?: number | string }`. The cast target (`SDKMessage`) is less precise than what the event accepts. This is not a bug — `SDKMessage` is assignable to `SDKMessage & { timestamp?: number | string }` because the intersection only adds an optional field — but the cast comments say "DB injects epoch-ms timestamp" while the type assertion says the result is plain `SDKMessage`. Future readers may find this slightly confusing.

**P3 — No tests for timestamp handling at event boundaries**

No test files were added or modified in this PR. The timestamp duality (number vs. string) at the `sdk.message` event boundary is not covered by any unit test. This is a pre-existing gap and not a regression introduced here, but it is worth noting.

### Summary

The P2 fix is correct and sufficient. The event type now accurately reflects the runtime contract: subscribers can receive either an ISO string (from the live SDK) or an epoch-ms number (from DB replay). The fix is minimal and targeted — only `daemon-hub.ts` changed in the fix commit, which is appropriate. No correctness issues were found.

PR #739, opened 2026-03-23
feat: implement ReadonlySessionChat and SlideOutPanel components

lsm:
- ReadonlySessionChat: independently fetches session messages via
  state.sdkMessages RPC + delta subscription, without touching
  sessionStore. Joins/leaves session channel, filters deltas by channel
  to prevent cross-session bleed, deduplicates by UUID for Safari
  reconnect resilience, supports load-older pagination via message.sdkMessages
  RPC with numeric before timestamp.

- SlideOutPanel: absolutely positioned right-side slide-out panel with
  slide-in animation (translateX), semi-transparent backdrop, header with
  role-colored agent label, close button, Escape key support, focus
  trapping and restoration, and full accessibility attributes.

- 22 unit tests covering: session isolation, RPC calls, channel filter
  (rejection and acceptance), open/close behavior, backdrop/Escape,
  role colors, accessibility, and all data-testid attributes.

### DIFF
--- a/packages/web/src/components/room/ReadonlySessionChat.tsx
+++ b/packages/web/src/components/room/ReadonlySessionChat.tsx
@@ -7,7 +7,7 @@
  *
  * Data loading:
  * - Joins channel `session:${sessionId}` and subscribes to `state.sdkMessages.delta`
- * - Initial fetch via `state.sdkMessages` RPC
+ * - Initial fetch via `state.sdkMessages` RPC (stale-fetch guarded by cancelled flag)
  * - Pagination via `message.sdkMessages` RPC (numeric `before` timestamp)
  * - Deduplicates by UUID on delta events (Safari reconnect replay guard)
  */
@@ -34,6 +34,7 @@ export function ReadonlySessionChat({ sessionId }: Props) {
 	const [hasMore, setHasMore] = useState(false);
 	const [isLoadingOlder, setIsLoadingOlder] = useState(false);
 	const [error, setError] = useState<string | null>(null);
+	const [paginationError, setPaginationError] = useState<string | null>(null);
 
 	const messagesEndRef = useRef<HTMLDivElement>(null);
 	const didScrollRef = useRef(false);
@@ -69,10 +70,13 @@ export function ReadonlySessionChat({ sessionId }: Props) {
 		};
 	}, [sessionId, isConnected, onEvent, joinRoom, leaveRoom]);
 
-	// Initial fetch — re-runs when sessionId or isConnected changes
+	// Initial fetch — re-runs when sessionId or isConnected changes.
+	// Uses a cancelled flag to discard stale responses when sessionId changes
+	// or the component unmounts before the request resolves.
 	useEffect(() => {
 		if (!sessionId || !isConnected) return;
 
+		let cancelled = false;
 		setIsLoading(true);
 		setError(null);
 		setMessages([]);
@@ -81,15 +85,21 @@ export function ReadonlySessionChat({ sessionId }: Props) {
 
 		request<{ sdkMessages: SDKMessage[]; hasMore: boolean }>('state.sdkMessages', { sessionId })
 			.then((result) => {
+				if (cancelled) return;
 				setMessages((result?.sdkMessages ?? []) as SDKMessageWithTimestamp[]);
 				setHasMore(result?.hasMore ?? false);
 			})
 			.catch((err: unknown) => {
+				if (cancelled) return;
 				setError(err instanceof Error ? err.message : 'Failed to load messages');
 			})
 			.finally(() => {
-				setIsLoading(false);
+				if (!cancelled) setIsLoading(false);
 			});
+
+		return () => {
+			cancelled = true;
+		};
 	}, [sessionId, isConnected, request]);
 
 	// Auto-scroll to bottom when new messages arrive
@@ -109,6 +119,7 @@ export function ReadonlySessionChat({ sessionId }: Props) {
 		if (!oldestTimestamp) return;
 
 		setIsLoadingOlder(true);
+		setPaginationError(null);
 		try {
 			const result = await request<{ sdkMessages: SDKMessage[]; hasMore: boolean }>(
 				'message.sdkMessages',
@@ -123,8 +134,8 @@ export function ReadonlySessionChat({ sessionId }: Props) {
 					return [...deduped, ...prev];
 				});
 			}
-		} catch {
-			// Silently ignore pagination errors
+		} catch (err: unknown) {
+			setPaginationError(err instanceof Error ? err.message : 'Failed to load older messages');
 		} finally {
 			setIsLoadingOlder(false);
 		}
@@ -144,6 +155,9 @@ export function ReadonlySessionChat({ sessionId }: Props) {
 					>
 						{isLoadingOlder ? 'Loading…' : 'Load older messages'}
 					</button>
+					{paginationError && (
+						<div class="text-xs text-red-400 text-center mt-1">{paginationError}</div>
+					)}
 				</div>
 			)}
 
--- a/packages/web/src/components/room/ReadonlySessionChat.tsx
+++ b/packages/web/src/components/room/ReadonlySessionChat.tsx
@@ -1,0 +1,185 @@
+/**
+ * ReadonlySessionChat
+ *
+ * Independently fetches and renders messages for a given sessionId without
+ * touching sessionStore. Used by SlideOutPanel to display a secondary session's
+ * chat without overwriting the primary session's data.
+ *
+ * Data loading:
+ * - Joins channel `session:${sessionId}` and subscribes to `state.sdkMessages.delta`
+ * - Initial fetch via `state.sdkMessages` RPC
+ * - Pagination via `message.sdkMessages` RPC (numeric `before` timestamp)
+ * - Deduplicates by UUID on delta events (Safari reconnect replay guard)
+ */
+
+import { useEffect, useRef, useState } from 'preact/hooks';
+import type { SDKMessage } from '@neokai/shared/sdk/sdk.d.ts';
+import { useMessageHub } from '../../hooks/useMessageHub';
+import { useMessageMaps } from '../../hooks/useMessageMaps';
+import { SDKMessageRenderer } from '../sdk/SDKMessageRenderer';
+
+type SDKMessageWithTimestamp = SDKMessage & { timestamp: number };
+
+interface Props {
+	sessionId: string;
+}
+
+const LOAD_OLDER_LIMIT = 100;
+
+export function ReadonlySessionChat({ sessionId }: Props) {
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-implemented PR that correctly delivers the ReadonlySessionChat and SlideOutPanel components. The code follows existing project patterns, maintains session isolation, and includes comprehensive tests. A few observations below.

### What looks good

1. **Session isolation** is properly maintained -- `ReadonlySessionChat` never imports or references `sessionStore`, relying entirely on `useMessageHub` for RPC and subscriptions. This is the core architectural requirement and it is correctly implemented.

2. **Channel filtering** correctly gates delta events via `context.channel === session:${sessionId}`, preventing cross-session bleed. The deduplication by UUID is a thoughtful Safari reconnect guard.

3. **SlideOutPanel** uses `absolute` positioning (not `fixed`), as required. Animation via `translate-x-0`/`translate-x-full` with `transition-transform` is clean.

4. **Accessibility** is well-handled: `role="dialog"`, `aria-modal="true"`, `aria-label`, focus trapping with Tab/Shift+Tab, Escape key close, and focus restoration to trigger element.

5. **Tests** are thorough (22 tests), covering isolation, RPC calls, channel filtering, visibility states, close interactions, role colors, accessibility, and data-testid presence. The mock for `sessionStore.select` to verify it is never called is a nice touch.

6. **Conditional mounting** (`isOpen && sessionId`) prevents unnecessary subscriptions and RPC calls when the panel is closed.

### Minor observations (P3)

1. **`ReadonlySessionChat.tsx` line 96** -- `scrollIntoView` on every `messages.length` change: When loading older messages (prepended to the array), this will scroll to the bottom, losing the user's scroll position. The current behavior is acceptable for a readonly panel showing a secondary session, but if users frequently load older messages, preserving scroll position during prepend would be a better UX. Not blocking since load-older is a secondary interaction.

2. **`ReadonlySessionChat.tsx` line 119** -- Pagination errors are silently swallowed (`catch {}`). A comment explains this is intentional, which is fine. Consider whether a brief toast or inline error would help users understand why "Load older" appeared to do nothing.

3. **`SlideOutPanel.tsx` line 106** -- The backdrop and panel are always rendered in the DOM (even when closed), with visibility controlled by CSS classes. This is a valid pattern for enabling CSS transitions, but it means the backdrop's click handler is always attached. The `pointer-events-none` class correctly prevents interaction when closed, so this is safe.

4. **No `ReadonlySessionChat`-specific test file** -- All tests live in `SlideOutPanel.test.tsx`. This is fine given the components are tightly coupled, but a separate test file for `ReadonlySessionChat` would improve discoverability if the component is reused elsewhere in the future.

All issues are P3 nits. The implementation is correct, complete, and well-tested.

lsm (commented):

---

All 22 tests pass, TypeScript compiles cleanly, lint is 0 errors — the happy-path implementation is solid. The core architectural requirements are met: `ReadonlySessionChat` never touches `sessionStore`, uses the correct `joinRoom`/`leaveRoom`/`onEvent`/`isConnected` pattern, filters deltas by channel, deduplicates by UUID, and uses a numeric `before` timestamp for pagination. `SlideOutPanel` is absolutely positioned, has all required `data-testid` attributes, `role="dialog"`, `aria-modal`, focus trapping, Escape key support, and backdrop click.

However there are two P1 issues that need fixing before merge.

---

### P1 — Stale fetch can overwrite correct state after rapid sessionId change

**File:** `packages/web/src/components/room/ReadonlySessionChat.tsx`, lines 73-93

The initial fetch effect fires `request("state.sdkMessages", ...)` and calls `setMessages` unconditionally in `.then()`. There is no cleanup/abort guard. If `sessionId` changes (or the component unmounts) before the first fetch resolves, the stale response will overwrite the newly-loaded messages for the new sessionId. The `useGroupMessages` hook avoids this exact race with `activeSubIdRef`: it records the current subscriptionId at effect start and guards every async callback.

The fix is a cancelled flag in the effect:

```ts
useEffect(() => {
  if (!sessionId || !isConnected) return;
  let cancelled = false;
  setIsLoading(true);
  setMessages([]);
  request(...)
    .then((result) => { if (!cancelled) setMessages(...); })
    .catch((err)   => { if (!cancelled) setError(...); })
    .finally(()    => { if (!cancelled) setIsLoading(false); });
  return () => { cancelled = true; };
}, [sessionId, isConnected, request]);
```

---

### P1 — Missing test: `leaveRoom` called on unmount / sessionId change

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`

`mockLeaveRoom` is wired up in the mock but never asserted. The plan explicitly calls out channel membership leaks as a concern (03-slide-out-panel.md, subtask 1 item 5). A test should mount the component, call `cleanup()`, and assert `mockLeaveRoom` was called with `"session:session-abc"`.

---

### P1 — Missing test: pagination / load-older flow

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`

`message.sdkMessages` is mocked but no test exercises the flow. The plan (task 3.2, message loading) requires testing that:
1. An initial fetch returning `hasMore: true` causes the button to render.
2. Clicking it calls `mockRequest` with `("message.sdkMessages", { sessionId, before: <number>, limit: 100 })`.

Without this, the correct RPC name, numeric `before` field, and `limit` parameter are completely untested.

---

### P2 — `ROLE_COLORS` duplicated rather than shared

**File:** `packages/web/src/components/room/SlideOutPanel.tsx`, lines 11-20

The plan says to import `ROLE_COLORS` from shared constants. Instead a stripped-down version is defined locally. `TaskConversationRenderer.tsx` already has the full map at line 49. If a role/color is added there, `SlideOutPanel` will silently fall back to grey. Either extract `ROLE_COLORS` to a shared file (e.g., `src/lib/role-colors.ts`) or import from `TaskConversationRenderer`.

---

### P2 — Silent pagination error swallow

**File:** `packages/web/src/components/room/ReadonlySessionChat.tsx`, lines 126-128

`loadOlderMessages` catches all errors silently. The user clicks the button, nothing happens, with no feedback. At minimum, expose a `paginationError` state and display a short message near the "Load older" button.

---

### P3 — No reconnect test

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`

The component documents reconnect resilience as a key design feature (line 41-42 comment). No test toggles `mockIsConnected` and verifies `joinRoom` and the initial fetch are re-invoked. Low risk since the logic inherits from the established `useGroupMessages` pattern, but a regression test would add confidence.

lsm (commented):

**Recommendation: APPROVE**

### P1 Issue Resolution Verification

All three previously raised P1 issues have been properly addressed:

**1. Stale fetch guard (FIXED)** — `ReadonlySessionChat.tsx` lines 79-102 now use a `let cancelled = false` pattern with a cleanup function that sets `cancelled = true`. The `.then()`, `.catch()`, and `.finally()` callbacks all check the flag before updating state. This correctly prevents stale responses from overwriting state when `sessionId` changes rapidly or on unmount.

**2. leaveRoom cleanup assertion (FIXED)** — Test at line 340-348 (`should call leaveRoom with the session channel on unmount`) correctly verifies that `leaveRoom` is called with the expected channel string after `unmount()`.

**3. Load-older pagination tests (FIXED)** — Two new tests cover pagination:
- Line 352: Verifies the "Load older messages" button appears when `hasMore=true`
- Line 367: Verifies clicking the button fires `message.sdkMessages` RPC with the correct `sessionId`, numeric `before` timestamp, and `limit: 100`

### Overall Assessment

The implementation is clean and well-structured:

- **Session isolation** is correctly maintained — `ReadonlySessionChat` never touches `sessionStore`, using only local state and direct RPC calls.
- **Channel filtering** on delta events (line 53) prevents cross-session message bleed.
- **Deduplication** via UUID Set is applied both on delta events and pagination results.
- **Error handling** covers both initial load errors and pagination errors with separate state.
- **Accessibility** is solid — `role="dialog"`, `aria-modal`, `aria-label`, focus trap, Escape key handler, and focus restoration on close.
- **Test coverage** (25 tests) is comprehensive for a UI component: session isolation, channel filtering, panel visibility states, close interactions, accessibility attributes, channel join/leave, and pagination.

### Minor Observations (P3)

1. **ROLE_COLORS duplication** — `TaskConversationRenderer.tsx` (line 49-58) still defines its own local `ROLE_COLORS` identical to the newly extracted `packages/web/src/lib/role-colors.ts`. The extraction was done for `SlideOutPanel` but the original source wasn't updated to consume the shared module. This is a follow-up cleanup item, not blocking.

2. **Auto-scroll on load-older** — When older messages are prepended (line 131-135), the scroll position jumps to bottom due to the `messages.length` effect (line 106-111). This is acceptable for an initial implementation but could be refined later to preserve scroll position when loading older messages.

No P0, P1, or P2 issues found. The implementation correctly addresses the original task requirements and the previously raised feedback.

lsm (commented):

**Re-review of P1 fixes — all three issues correctly resolved.**

### P1 Issue 1: Stale fetch guard — FIXED

The initial fetch useEffect in `packages/web/src/components/room/ReadonlySessionChat.tsx` (lines 76–103) now correctly uses a `let cancelled = false` guard with a cleanup function `return () => { cancelled = true; }`. All three state-setting paths (`.then`, `.catch`, and `.finally`) check `if (cancelled)` before running. The `.finally` uses `if (!cancelled) setIsLoading(false)` which is correct — no state mutation on a stale effect. No issues here.

### P1 Issue 2: leaveRoom cleanup assertion — FIXED

A new test at line 339 mounts the panel, waits for `joinRoom` to be called, unmounts, and then asserts `mockLeaveRoom` was called with the exact channel string. This directly tests the cleanup path of the subscription useEffect. Correct and complete.

### P1 Issue 3: Load-older pagination flow — FIXED

Two new tests were added:
- Line 350: Verifies the "Load older messages" button appears when `hasMore=true` from the initial fetch.
- Line 368: Verifies clicking the button fires `message.sdkMessages` with the correct `sessionId`, the numeric `before` timestamp from the oldest message, and `limit: 100`.

Both tests correctly use `mockRequest.mockImplementation` (persistent, not once-only) to ensure the implementation is stable across the async flow. The timestamp check is numeric as required.

### Additional improvement: paginationError state

The worker also added `paginationError` state (replacing a silent catch block) with a rendered error message inside the `hasMore` conditional block. This is a good improvement. The error remains visible as long as `hasMore` stays true — which it does after a failed pagination request — so the UX is correct. No test was added for the pagination error display path, but it was not part of the P1 requirements and is a P3 nit.

The `paginationError` display is only rendered when `hasMore` is true (it lives inside the `{hasMore && ...}` block). If a pagination error occurs but the response sets `hasMore` to false, the error message would be hidden. However in the actual flow the `hasMore` update only happens in the `try` block, and the error only sets in the `catch` block, so these paths are mutually exclusive. The implementation is correct.

### Verdict

All three P1 issues from the previous review have been properly addressed with correct implementation and adequate test coverage. **Recommendation: APPROVE.**

PR #741, opened 2026-03-23
test: unit tests for ReadonlySessionChat and SlideOutPanel

lsm:
## Summary

- Add dedicated `ReadonlySessionChat` describe block with direct-mount tests (11 new tests)
- Enhance session isolation test to explicitly assert `sessionStore.activeSessionId` unchanged
- Fix mock implementation leakage across describe blocks using `vi.resetAllMocks()`
- All 36 tests pass (up from 25)

### ReadonlySessionChat (new direct-mount tests)
- Renders `data-testid="readonly-session-chat"`
- Session isolation: `sessionStore.select` never called, `activeSessionId` unchanged
- Fetches via `state.sdkMessages` RPC with correct `sessionId`
- Joins/leaves session channel on mount/unmount
- Subscribes to `state.sdkMessages.delta`
- **Cross-session rejection**: delta with `(data, context)` two-arg shape, wrong channel → message not rendered
- **Cross-session acceptance**: delta with correct channel → message rendered
- Empty list placeholder ("No messages yet")
- Fetch error handling
- Stale-fetch guard: `cancelled` flag prevents stale response from overwriting newer session

### SlideOutPanel (existing, enhanced)
- All 25 original tests retained and passing
- Session isolation test now also asserts `activeSessionId.value` unchanged

## Test plan
- [x] `bunx vitest run src/components/room/__tests__/SlideOutPanel.test.tsx` — 36/36 pass
- [x] `bunx vitest run` — 5081/5081 pass

### DIFF
--- a/packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx
+++ b/packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx
@@ -58,7 +58,10 @@ vi.mock('../../../hooks/useMessageHub.ts', () => ({
 	useMessageHub: () => ({
 		request: mockRequest,
 		onEvent: mockOnEvent,
-		isConnected: mockIsConnected.value,
+		// Use a getter so tests can flip mockIsConnected.value after module load
+		get isConnected() {
+			return mockIsConnected.value;
+		},
 		joinRoom: mockJoinRoom,
 		leaveRoom: mockLeaveRoom,
 	}),
@@ -547,5 +550,38 @@ describe('ReadonlySessionChat', () => {
 		// Stale message must NOT appear
 		await waitFor(() => expect(container.textContent).toContain('No messages yet'));
 		expect(container.querySelector('[data-testid="msg-stale-msg-uuid"]')).toBeNull();
+	});
+
+	// --- Disconnected guard path ---
+
+	it('should not call joinRoom or fetch when isConnected is false', () => {
+		mockIsConnected.value = false;
+		render(<ReadonlySessionChat sessionId="disconnected-session" />);
+		expect(mockJoinRoom).not.toHaveBeenCalled();
+		expect(mockRequest).not.toHaveBeenCalled();
+	});
+
+	it('should not subscribe to delta events when isConnected is false', () => {
+		mockIsConnected.value = false;
+		render(<ReadonlySessionChat sessionId="disconnected-session" />);
+		expect(mockOnEvent).not.toHaveBeenCalled();
+	});
+
+	it('should fetch and join when isConnected transitions from false to true', async () => {
+		mockIsConnected.value = false;
+		const { rerender } = render(<ReadonlySessionChat sessionId="reconnect-session" />);
+		expect(mockJoinRoom).not.toHaveBeenCalled();
+		expect(mockRequest).not.toHaveBeenCalled();
+
+		// Simulate reconnection by toggling isConnected and re-rendering
+		mockIsConnected.value = true;
+		rerender(<ReadonlySessionChat sessionId="reconnect-session" />);
+
+		await waitFor(() => expect(mockJoinRoom).toHaveBeenCalledWith('session:reconnect-session'));
+		await waitFor(() =>
+			expect(mockRequest).toHaveBeenCalledWith('state.sdkMessages', {
+				sessionId: 'reconnect-session',
+			})
+		);
 	});
 });
--- a/packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx
+++ b/packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx
@@ -15,6 +15,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import { render, cleanup, fireEvent, act, waitFor } from '@testing-library/preact';
 import { SlideOutPanel } from '../SlideOutPanel';
+import { ReadonlySessionChat } from '../ReadonlySessionChat';
 
 // -------------------------------------------------------
 // Mocks
@@ -80,12 +81,15 @@ vi.mock('../../../hooks/useMessageMaps.ts', () => ({
 	}),
 }));
 
-// Mock sessionStore to spy on select
+// Mock sessionStore to spy on select — value starts null and must stay null
 const mockSessionStoreSelect = vi.fn();
+const mockActiveSessionId = { value: null as string | null };
 vi.mock('../../../lib/session-store.ts', () => ({
 	sessionStore: {
 		select: mockSessionStoreSelect,
-		activeSessionId: { value: null },
+		get activeSessionId() {
+			return mockActiveSessionId;
+		},
 		sdkMessages: { value: [] },
 	},
 }));
@@ -107,6 +111,7 @@ describe('SlideOutPanel', () => {
 		vi.clearAllMocks();
 		registeredDeltaHandler = null;
 		mockIsConnected.value = true;
+		mockActiveSessionId.value = null;
 	});
 
 	afterEach(() => {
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

All 36 tests pass. The implementation covers the requested areas well:

### What works well

1. **Session isolation** -- Both the `SlideOutPanel` and direct `ReadonlySessionChat` tests verify that `sessionStore.select` is never called and `activeSessionId` remains null. The getter-based mock for `activeSessionId` is a clean approach.

2. **Cross-session channel filtering** -- Tests cover both rejection (wrong channel) and acceptance (correct channel) with the two-argument `(data, context)` handler signature matching `ChannelEventHandler`. This is the most important behavioral test for this component.

3. **Stale fetch guard** (line 524-550) -- The test that uses a deferred promise to simulate a race condition between `sessionId` change and fetch resolution is well-crafted and tests a real edge case.

4. **Mock reset strategy** -- Using `vi.resetAllMocks()` in the `ReadonlySessionChat` block with explicit re-registration of mock implementations avoids leakage from the `SlideOutPanel` block. This is correctly handled.

5. **Accessibility and data-testid coverage** -- `role="dialog"`, `aria-modal`, `aria-label` all verified.

### Minor observations (P3)

1. **`mockIsConnected` is a plain object, not a signal** (line 28) -- The mock uses `{ value: true }` which works because `useMessageHub` returns `isConnected` as a plain boolean value destructured from the mock. This is fine for the current tests but worth noting if the component ever reads `isConnected` reactively.

2. **No test for `isConnected=false` on `ReadonlySessionChat`** -- The `ReadonlySessionChat` describe block doesn't test behavior when `isConnected` is false (e.g., verifying no fetch and no channel join). The `SlideOutPanel` block also doesn't test this. This is a minor gap -- not blocking since the production code guards on `isConnected` in both effects.

3. **`makeMessage` returns a minimal shape** (line 101-103) -- The helper omits fields like `content`, `model`, etc. This works because `SDKMessageRenderer` is mocked, but a comment noting this intentional minimalism would help future readers.

Overall this is a solid, focused test PR that covers the key behavioral contracts of both components.

lsm (commented):

**Recommendation: APPROVE** — the implementation correctly meets all stated requirements and all 36 tests pass.

---

### Summary

The 148-line addition to `SlideOutPanel.test.tsx` covers the full task scope: session isolation, cross-session channel filtering (both directions with the correct two-argument `(data, context)` shape), accessibility attributes, panel open/close mechanics, join/leave lifecycle, empty-list placeholder, error handling, and the stale-fetch cancellation guard. The mock structure is well-organized and the `vi.resetAllMocks()` + re-apply pattern in the `ReadonlySessionChat` `beforeEach` correctly prevents mock leakage between the two `describe` blocks.

---

### Issues Found

#### P2 — `isConnected` mock is frozen at module-load time

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`, line 61

```ts
useMessageHub: () => ({
    isConnected: mockIsConnected.value,   // captured once at factory call time
    ...
})
```

`vi.mock` factories are hoisted and run once. Updating `mockIsConnected.value` in a test has no effect because the mock already captured the primitive `true` at the time the factory executed. A dynamic getter would be needed:

```ts
get isConnected() { return mockIsConnected.value; }
```

In practice this causes no test failures because every `beforeEach` sets `mockIsConnected.value = true` and no test ever flips it to `false`. However, any future test that tries to exercise the disconnected guard (`if (!sessionId || !isConnected) return`) would silently see `true` regardless, making it unreliable.

#### P2 — No test for the disconnected (`isConnected = false`) guard path

`ReadonlySessionChat` has explicit guards in both effects:

```ts
if (!sessionId || !isConnected) return;
```

There is no test verifying that `joinRoom`, `leaveRoom`, and the `state.sdkMessages` RPC are all suppressed when the hub is not connected. This is a gap in the coverage stated by the task.

#### P3 — Mock `DeltaHandler` context type is narrower than the real `ChannelEventHandler` signature

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`, line 25

```ts
type DeltaHandler = (data: { added?: unknown[] }, context: { channel?: string }) => void;
```

The real `ChannelEventHandler` context is `EventContext & { channel?: string }`, where `EventContext` requires `messageId`, `sessionId`, `method`, and `timestamp`. The test invocations pass only `{ channel: '...' }`. This works because the component only reads `context.channel`, but the mock type drifts from the real interface. If the component is ever updated to read other context fields, the tests would silently pass with undefined values instead of failing loudly.

lsm (commented):

**Recommendation: APPROVE**

The two P2 issues from the previous review have been correctly addressed:

### Fix 1: Dynamic `isConnected` mock (getter pattern)

The `useMessageHub` mock at line 62 now uses a getter instead of a captured primitive:

```ts
get isConnected() {
    return mockIsConnected.value;
}
```

This ensures tests that flip `mockIsConnected.value` after module load see the updated value at call-time rather than the value frozen at module initialization. The same getter pattern was applied to `mockActiveSessionId` (line 93) for consistency.

### Fix 2: Disconnected guard tests

Three new tests cover the `if (!sessionId || !isConnected) return` guard in both effects of `ReadonlySessionChat`:

1. **Line 557** - Verifies no `joinRoom` or `request` calls when `isConnected` is false
2. **Line 564** - Verifies no `onEvent` subscription when `isConnected` is false
3. **Line 570** - Verifies `joinRoom` and `request` resume when `isConnected` transitions false to true (via `rerender`)

All three tests correctly exercise the guard paths. The reconnection test at line 570 properly simulates the transition by flipping the mock value and calling `rerender`, which triggers Preact to re-run the effects with the new `isConnected` value.

### Other observations

- The `ReadonlySessionChat` describe block correctly uses `vi.resetAllMocks()` in `beforeEach` and re-registers mock implementations afterward, avoiding stale mock state leaking between tests.
- All 39 tests pass (707ms total).
- No issues found with the implementation.

lsm (commented):

**Recommendation: APPROVE**

Both P2 items from the previous review are correctly addressed. The getter fix is technically sound and the three new disconnected-guard tests verify the right code paths. 39/39 tests pass.

---

### P2 fix verified: isConnected getter

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`, lines 61-64

The change from `isConnected: mockIsConnected.value` to `get isConnected() { return mockIsConnected.value; }` is correct. Because `useMessageHub` is a hook called on every render, the mock factory object is reconstructed per-render, and the getter is evaluated fresh each time. Setting `mockIsConnected.value = false` in a test then calling `render` or `rerender` correctly delivers the updated boolean to the component.

### New disconnected-guard tests: analysis

**Test 1** (line 557) and **Test 2** (line 564) are synchronous negative assertions — render with `isConnected=false`, assert no calls. These are correct and compact. No `waitFor` needed since both effects bail before calling any mock.

**Test 3** (line 570) correctly simulates the false-to-true transition: render disconnected, assert nothing fired, flip the value, rerender, then `waitFor` both `joinRoom` and `request`. These are the two behaviors guarded by the two separate `useEffect` hooks in the component, so both assertions are necessary.

---

### Remaining minor observation (P3, unchanged from prior review)

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`, line 25

```ts
type DeltaHandler = (data: { added?: unknown[] }, context: { channel?: string }) => void;
```

The real `ChannelEventHandler` context type is `EventContext & { channel?: string }`, where `EventContext` requires `messageId`, `sessionId`, `method`, and `timestamp`. The test invocations pass only `{ channel: '...' }`. The tests work correctly because the component only reads `context.channel`, but the local type diverges from the real interface. Not blocking.

---

### Small gap in Test 3 (P3)

**File:** `packages/web/src/components/room/__tests__/SlideOutPanel.test.tsx`, line 570-586

The reconnection test verifies `joinRoom` and `request` resume correctly, but does not assert that `onEvent('state.sdkMessages.delta', ...)` is also called. Both effects (channel subscription and initial fetch) are gated on `isConnected`, so both re-run on reconnect. Adding `expect(mockOnEvent).toHaveBeenCalledWith('state.sdkMessages.delta', expect.any(Function))` after the rerender would complete the picture. Not blocking — the existing two assertions are sufficient to prove the guard works.

PR #742, opened 2026-03-23
feat: define TurnBlock types and implement useTurnBlocks hook

lsm:
Adds the useTurnBlocks hook (packages/web/src/hooks/useTurnBlocks.ts)
that transforms a flat SessionGroupMessage[] into structured TurnBlockItems.

- TurnBlock interface with id (stable UUID), sessionId, agentRole, agentLabel,
  startTime, endTime, messageCount, toolCallCount, thinkingCount, assistantCount,
  lastAction, previewMessage, isActive, isError, errorMessage, messages
- RuntimeMessage interface for status/rate_limited/model_fallback/leader_summary
- TurnBlockItem union type
- Groups messages by authorSessionId; new turn on session change
- Runtime messages (authorRole==='system') rendered inline between turns
- Human role messages form their own turn blocks
- Counts tool_use/thinking blocks and assistant messages per turn
- lastAction = most recent tool_use block name
- isActive=true only when isAtTail=true AND last turn AND no result message
- Error detection from SDKResultError (is_error=true)
- Memoized via useMemo keyed on messages reference, length, and isAtTail

46 unit tests covering all grouping logic, stats counting, active/error detection,
memoization, and complex multi-agent interleaving scenarios.

### DIFF
--- a/packages/web/src/hooks/__tests__/useTurnBlocks.test.ts
+++ b/packages/web/src/hooks/__tests__/useTurnBlocks.test.ts
@@ -9,7 +9,7 @@
  * that parseGroupMessage can parse (or a known non-JSON type like 'status').
  */
 
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, beforeEach } from 'vitest';
 import { renderHook } from '@testing-library/preact';
 import {
 	useTurnBlocks,
@@ -315,8 +315,8 @@ describe('useTurnBlocks', () => {
 			items.forEach((item) => expect(item.type).toBe('runtime'));
 		});
 
-		it('runtime messages do not split agent turns that share the same authorSessionId', () => {
-			// Two agent messages from the same session with a runtime message before them
+		it('runtime message before same-session agent messages does not split the turn', () => {
+			// Runtime message arrives BEFORE any agent message — emitted immediately, turn intact
 			const msgs = [
 				makeStatusMessage(),
 				makeAgentMessage({ authorRole: 'coder', authorSessionId: 'sess-1', createdAt: 2000 }),
@@ -328,6 +328,56 @@ describe('useTurnBlocks', () => {
 			expect(items[0].type).toBe('runtime');
 			const turn = asTurn(items[1]);
 			expect(turn.messageCount).toBe(2);
+		});
+
+		it('runtime message mid-turn does not fragment the agent turn', () => {
+			// Status update arrives between two messages from the same session.
+			// The turn should remain one cohesive block (not split into two).
+			const msgs = [
+				makeAgentMessage({ authorRole: 'coder', authorSessionId: 'sess-1', createdAt: 1000 }),
+				makeStatusMessage({ createdAt: 2000 }),
+				makeAgentMessage({ authorRole: 'coder', authorSessionId: 'sess-1', createdAt: 3000 }),
+			];
+			const items = renderUseTurnBlocks(msgs);
+			// 1 turn (2 agent msgs merged) + 1 runtime (emitted after the turn)
+			expect(items).toHaveLength(2);
+			expect(items[0].type).toBe('turn');
+			const turn = asTurn(items[0]);
+			expect(turn.messageCount).toBe(2);
+			expect(items[1].type).toBe('runtime');
+		});
+
+		it('multiple runtime messages mid-turn are all emitted after the turn', () => {
+			const msgs = [
+				makeAgentMessage({ authorRole: 'coder', authorSessionId: 'sess-1', createdAt: 1000 }),
+				makeStatusMessage({ createdAt: 2000 }),
+				makeRateLimitedMessage({ createdAt: 3000 }),
+				makeAgentMessage({ authorRole: 'coder', authorSessionId: 'sess-1', createdAt: 4000 }),
+			];
+			const items = renderUseTurnBlocks(msgs);
+			// 1 turn + 2 runtime items (emitted after the turn that caused the flush)
+			expect(items).toHaveLength(3);
+			expect(items[0].type).toBe('turn');
+			expect(asTurn(items[0]).messageCount).toBe(2);
+			expect(items[1].type).toBe('runtime');
+			expect(items[2].type).toBe('runtime');
+		});
+
+		it('mid-turn runtime messages appear between turns when a new agent follows', () => {
+			// [agent1, status, agent2] — status was buffered during agent1's turn,
+			// emitted between the two turns once agent2 starts speaking.
+			const msgs = [
+				makeAgentMessage({ authorRole: 'coder', authorSessionId: 'worker', createdAt: 1000 }),
+				makeStatusMessage({ createdAt: 2000 }),
+				makeAgentMessage({ authorRole: 'leader', authorSessionId: 'leader', createdAt: 3000 }),
+			];
+			const items = renderUseTurnBlocks(msgs);
+			expect(items).toHaveLength(3);
+			expect(items[0].type).toBe('turn');
+			expect(asTurn(items[0]).sessionId).toBe('worker');
+			expect(items[1].type).toBe('runtime');
+			expect(items[2].type).toBe('turn');
+			expect(asTurn(items[2]).sessionId).toBe('leader');
 		});
 	});
 
@@ -451,6 +501,16 @@ describe('useTurnBlocks', () => {
 				const items = renderUseTurnBlocks(msgs);
 				expect(asTurn(items[0]).agentLabel).toBe(expectedLabels[i]);
 			}
+		});
+
+		it('agentLabel falls back to raw agentRole for unknown roles', () => {
+			// An unregistered role not present in ROLE_COLORS silently falls back to the
+			// raw role string rather than throwing — this covers future SDK role additions.
+			const msgs = [makeAgentMessage({ authorRole: 'unknown-agent', authorSessionId: 'sess-x' })];
+			const items = renderUseTurnBlocks(msgs);
+			const turn = asTurn(items[0]);
+			expect(turn.agentRole).toBe('unknown-agent');
+			expect(turn.agentLabel).toBe('unknown-agent');
 		});
 
 		it('startTime is timestamp of first message in the turn', () => {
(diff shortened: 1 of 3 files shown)


lsm (commented):

Well-structured implementation. The types are clean, the hook logic is straightforward, and the 46 tests cover the important cases well. A few observations:

### P2: Runtime messages split same-session turns (design concern)

**File:** `packages/web/src/hooks/useTurnBlocks.ts`, lines 254-258

When a runtime message (e.g., a `status` update) arrives between two agent messages from the same session, `flushTurn()` is called, splitting what should logically be one continuous turn into two separate `TurnBlock` items. For example, `[agent-sess1-msg1, status, agent-sess1-msg2]` produces 3 items (turn + runtime + turn) rather than keeping the agent messages in one turn with the runtime item attached.

The test "runtime messages do not split agent turns that share the same authorSessionId" (line 318) is misleadingly named -- it only tests runtime *before* agent messages, not *between* them. A runtime message arriving mid-turn (which happens regularly when status updates stream during tool execution) will fragment the turn.

Consider either: (a) buffering runtime messages and emitting them without flushing the current accumulator, or (b) adding a test that documents the current splitting behavior explicitly so consumers know what to expect. This matters for UI because a long coder turn with intermittent status updates would show as many small turns instead of one cohesive block.

### P3: Redundant `messages.length` in useMemo dependency array

**File:** `packages/web/src/hooks/useTurnBlocks.ts`, line 318

```ts
}, [messages, messages.length, isAtTail]);
```

`messages.length` is redundant when `messages` (the array reference) is already a dependency. If the reference changes, the memo recomputes regardless. If the reference stays the same, the length cannot change. This is harmless but slightly misleading -- it suggests the author was worried about array mutation, which would be a bug in the caller (`useGroupMessages` always returns a new array reference via `setMessages`).

### P3: `@ts-nocheck` in test file

**File:** `packages/web/src/hooks/__tests__/useTurnBlocks.test.ts`, line 1

The `@ts-nocheck` disables all TypeScript checking in the test file. While common in test files, it means type errors in the test helpers (e.g., `makeAgentMessage`, `makeResultMessage`) won't be caught. The test helpers are constructing complex nested objects that mirror SDK types -- if those types change, the tests would still pass with stale/wrong shapes. Consider removing `@ts-nocheck` and adding targeted `as` casts where needed.

### Overall

The implementation is correct and complete for the stated goal. The hook properly handles multi-agent interleaving, runtime message routing, active/error detection, and memoization. The types are well-documented with JSDoc. The test suite is thorough with good coverage of edge cases.

**Recommendation:** APPROVE (the P2 is a design consideration worth discussing but not blocking, since the current behavior is internally consistent even if it may cause UI fragmentation).

lsm (commented):

Overall the implementation is well-structured with a clear separation of concerns and solid test coverage for the happy paths. However, there are a few concrete correctness issues and one notable missing guard that need addressing before this ships.

---

### P0 — Error detection only inspects the last message, silently misses errors mid-turn

`packages/web/src/hooks/useTurnBlocks.ts` lines 212–216:

```ts
const lastMsg = msgs[msgs.length - 1] ?? null;
const { isError, errorMessage } = lastMsg
  ? extractErrorInfo(lastMsg)
  : { isError: false, errorMessage: null };
```

`extractErrorInfo` is called only on the last message. However, `SDKAssistantMessage` carries an optional `error?: SDKAssistantMessageError` field (e.g. `billing_error`, `authentication_failed`) that can appear on any assistant-type message, not exclusively on the final result. If an error-carrying assistant message arrives before additional messages (e.g. a streaming partial followed by a retry), the turn will report `isError: false` and silently drop the error context. The same logic covers the billing/auth error path described in the comment at line 146–150.

The fix is to scan all messages in the turn for errors, not only the last one — or at minimum prefer the last error-result message over the last arbitrary message.

---

### P1 — `useMemo` dependency array includes `messages.length` alongside `messages` — semantically misleading and fragile

`packages/web/src/hooks/useTurnBlocks.ts` line 318:

```ts
}, [messages, messages.length, isAtTail]);
```

`messages.length` is evaluated at the time the `useMemo` call is made. Because `useGroupMessages` returns a new array reference on every delta (spread/filter/map), the `messages` reference dep already captures all changes including length. Keeping `messages.length` separately creates a misleading implication that reference-stable, mutation-in-place arrays are expected — but no part of the codebase mutates the messages array in place.

More importantly, this pattern creates an undocumented contract: future callers who pass a reference-stable, content-mutated array (same ref, different items) would get stale results — `messages.length` would not catch a message being replaced at the same index. The dep list should be `[messages, isAtTail]` and a comment should explain that a new array reference is required when any message is updated, pointing to `useGroupMessages`'s spread behaviour as the guarantee.

---

### P1 — Hook is not exported from the hooks barrel index

`packages/web/src/hooks/index.ts` exports every other hook in the `hooks/` directory, but `useTurnBlocks` and its exported types (`TurnBlock`, `TurnBlockItem`, `RuntimeMessage`) are absent. This forces future consumers to import directly from the file path, breaking the consistent import pattern used across the codebase and likely to cause a Knip false-negative if the file is ever removed.

---

### P1 — Memoization test does not exercise the critical content-mutation case

`packages/web/src/hooks/__tests__/useTurnBlocks.test.ts` lines 714–749 cover reference change and `isAtTail` change, but there is no test for the scenario that matters most for correctness: passing a same-length array with different content (i.e. simulating an in-place updated message, or a message replaced at the same index). Given that `messages.length` was explicitly added to the dep array to handle this, its absence means the dep rationale is untested.

---

### P2 — Error detection skips errors that aren't on the last message of a turn

Follows from the P0 above, but worth calling out separately as a test gap: there are no tests in the error detection suite for a turn where an error-carrying assistant message appears before the final message (e.g. `billing_error` assistant message followed by a user message). The test suite only exercises the case where the result message is the last one.

---

### P2 — `agentLabel` falls back to raw `agentRole` string for unknown roles

`packages/web/src/hooks/useTurnBlocks.ts` line 213:

```ts
const agentLabel = ROLE_COLORS[agentRole]?.label ?? agentRole;
```

`TaskMeta.authorRole` includes roles like `'general'`, `'craft'`, and `'lead'` (all present in `ROLE_COLORS`), but if the SDK introduces a new role that hasn't been added to `ROLE_COLORS`, `agentLabel` silently falls back to the raw role string. This is likely fine for display but it's an invisible behaviour that should be tested. There is currently no test for an unknown/unregistered `authorRole`.

---

### P3 — Test file uses `@ts-nocheck`

`packages/web/src/hooks/__tests__/useTurnBlocks.test.ts` line 1 suppresses all TypeScript checks in the test. While `@ts-nocheck` is a project-wide pattern in test files (present in ~20 other test files), it means type errors in test helper construction (e.g. malformed `SessionGroupMessage`) would not be caught by `tsc`. This is a codebase-wide concern, not specific to this PR.

---

### P3 — `parseGroupMessage` called on all messages on every recompute

Every rerender triggers `messages.map(parseGroupMessage)` for the entire message list. For a conversation with hundreds of messages this is O(N) JSON parsing on each reactive update. A future optimisation could cache the parsed result alongside each message (e.g. as a memo keyed per message id). Not a blocker, but worth noting as the message count grows.

---

### Summary

The grouping logic, turn boundary detection, runtime message routing, and active/error detection are all correctly implemented for the documented happy paths, and the 46 tests cover the core scenarios well. The main blockers are: (1) the error detection bug that silently misses mid-turn errors, and (2) the hook not being exported from the barrel index.

lsm (commented):

Re-review of the TurnBlock types and useTurnBlocks hook implementation after the fix commit (`d937e0f`).

### Previous Issues - All Resolved

All five previously reported issues have been addressed:

1. **P0 - Error detection scans all messages**: `extractTurnErrorInfo()` now iterates all messages in a turn and returns the last error found, so result-level errors correctly take precedence over earlier assistant-level errors. Well tested with two new cases (lines 770-809 of the test file).

2. **P1 - Barrel export**: `useTurnBlocks` and all three public types (`TurnBlock`, `RuntimeMessage`, `TurnBlockItem`) are now re-exported from `packages/web/src/hooks/index.ts`.

3. **P1/P3 - Redundant deps**: `messages.length` removed from `useMemo` deps with a clear comment explaining why the `messages` reference alone is sufficient.

4. **P2 - Runtime message buffering**: `pendingRuntime` buffer prevents status updates from fragmenting turns. The `flushTurnAndRuntime()` function correctly handles all three cases: no open turn (drain immediately), open turn (buffer), and end-of-loop flush. Four new tests cover this behavior (lines 333-381).

5. **Tests**: 52 tests pass, covering empty input, all runtime message types, turn grouping, field correctness, stats counting, lastAction, active turn detection, error detection, memoization, and complex multi-agent scenarios.

- **Barrel index test not updated**: The existing `packages/web/src/hooks/__tests__/index.test.ts` does not import `useTurnBlocks` to verify the new barrel export. This is a minor gap -- the export is verified by typecheck and the hook's own test suite imports from the module directly. Not blocking.

### Verdict

The implementation is clean, well-documented, and thoroughly tested. The hook correctly handles multi-agent interleaving, runtime message buffering, error detection across all messages in a turn, and active turn detection. All previous review feedback has been addressed. Typecheck passes cleanly.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE** — All previously requested changes were correctly implemented. 52 tests pass, all checks clean.

---

### Summary

The four previously flagged items (P0 error scanning all messages, P1 barrel export, P1/P3 memo dep, P2 mid-turn buffering) are all correctly addressed. The implementation is clean, well-documented, and logically correct. I found only minor remaining issues below.

---

### Issues Found

#### P2 — Untested edge: error result with empty `errors` array

`extractErrorInfo` returns `{ isError: true, errorMessage: null }` when `is_error=true` but `errors=[]`. This is the correct behavior, but there is no test for it. A consumer rendering `turn.isError` and expecting `turn.errorMessage` to provide text would get a null and silently show nothing. The test at line 741 only covers the case where `errors` is non-empty.

**File:** `packages/web/src/hooks/__tests__/useTurnBlocks.test.ts`

#### P2 — Buffering tests do not cover `model_fallback` mid-turn

The mid-turn buffering tests cover `status` and `rate_limited`. The test at line 350 explicitly mixes `status` and `rate_limited`, but no test exercises `model_fallback` arriving mid-turn. The code path is identical so it works, but coverage consistency is slightly off given the four runtime types were each given their own top-level emit test.

**File:** `packages/web/src/hooks/__tests__/useTurnBlocks.test.ts`

#### P3 — `RuntimeMessage.index` documents the wrong array

The `index` field is documented as "Position of this message in the original parsed array" at line 73, which points to the post-`parseGroupMessage` parsed array (after null filtering), not the original `messages: SessionGroupMessage[]` input. A message filtered out by `parseGroupMessage` would shift indices for all subsequent runtime messages. The test at line 886 verifies specific index values, which confirms the current behavior is consistent, but the JSDoc is slightly misleading for future consumers who may assume it indexes into the raw input.

**File:** `packages/web/src/hooks/useTurnBlocks.ts`, line 73

#### P3 — `endTime: 0` is falsy-but-not-null for completed turns with missing timestamps

`getMessageTimestamp` returns `0` when the `timestamp` field is absent. In practice `parseGroupMessage` always injects `timestamp: msg.createdAt` so this cannot happen with well-formed DB rows, but a consumer checking `!turn.endTime` to detect the active-turn sentinel would incorrectly match epoch-0. The sentinel contract is `null` (documented), so consumers checking `=== null` are safe; but the numeric-0 fallback is a latent footgun if someone writes `if (!turn.endTime)`.

**File:** `packages/web/src/hooks/useTurnBlocks.ts`, lines 95–98

---

### What was done well

- `extractTurnErrorInfo` correctly scans all messages with last-error-wins semantics, and the result-over-assistant-error precedence is explicitly tested (line 791).
- The mid-turn buffering logic is clean: `pendingRuntime[]` is flushed immediately after `flushTurnAndRuntime`, so ordering is deterministic and the test at line 366 verifies the interleaved output order.
- The post-process active-detection loop walks backwards through `items` skipping runtime messages at the tail, correctly marking the last *turn* (not the last item) as active. The test at line 875 covers this exact scenario.
- Barrel export is complete: all three exported types (`TurnBlock`, `RuntimeMessage`, `TurnBlockItem`) plus the hook are exported from `packages/web/src/hooks/index.ts`.
- All checks pass: lint (0 warnings/errors), typecheck, knip, and all 52 Vitest tests.

PR #745, opened 2026-03-23
feat: implement RuntimeMessageRenderer component with unit tests

lsm:


lsm (commented):

### Verdict: **APPROVE**

This is a clean, faithful extraction of runtime message rendering logic from `TaskConversationRenderer.tsx` (V1) into a standalone `RuntimeMessageRenderer` component for TaskViewV2. The implementation is correct, well-tested, and consistent with the existing codebase.

### What looks good

- **Pixel-perfect V1 parity**: Every message type (status, rate_limited, model_fallback, leader_summary) matches the V1 rendering in `TaskConversationRenderer.tsx` exactly -- same CSS classes, same SVG icons, same fallback text, same field access patterns. This is exactly what was requested.

- **Defensive field access**: The `as Record<string, unknown>` pattern with `typeof` guards on each field is consistent with V1 and handles the untyped nature of `SDKMessage` runtime messages safely.

- **Test coverage is thorough**: 29 tests covering all four message types, fallback values, role labels, styling classes, the `[Turn Summary]` prefix stripping, `resetsAt` presence/absence, and the null-render for unknown types. The `MarkdownRenderer` mock is appropriate.

- **Component returns null for unknown types**: Correct behavior -- no crash, no empty DOM node.

### Minor observations (P3)

1. **`packages/web/src/components/room/RuntimeMessageRenderer.tsx` line 16** -- The `as Record<string, unknown>` cast is a code smell but is inherited from V1 and is the established pattern. The `SDKMessage` union type doesn't include these runtime message shapes natively, so this is the pragmatic approach. If/when runtime message types get added to the SDK type definitions, this cast could be removed.

2. **Not yet consumed by any parent component** -- `RuntimeMessageRenderer` is exported but not imported anywhere. This is expected since the TaskViewV2 integration is a separate task, and knip presumably allows it via the component discovery pattern. Just confirming this is intentional.

3. **Duplicated rendering logic with V1** -- Both `TaskConversationRenderer.tsx` and `RuntimeMessageRenderer.tsx` now contain identical rendering code for these four message types. When TaskViewV2 is fully wired up, the V1 code should be refactored to use `RuntimeMessageRenderer` or the old code should be removed. This is out of scope for this PR but worth tracking.

No P0, P1, or P2 issues found. The implementation is correct, complete, and well-tested.

lsm (commented):

The component is well-structured and its styling faithfully mirrors V1. The 29-test suite is thorough for the cases it covers. However, there are two issues worth addressing before merge: one is a missing integration (the component is not wired up anywhere), and one is a genuine UX bug inherited from V1.

---

### P0 — Component is never integrated

`packages/web/src/components/room/RuntimeMessageRenderer.tsx` is a new file that is not imported, exported, or rendered by any other file in the codebase. A search for all consumers confirms this:

```
packages/web/src/components/room/__tests__/RuntimeMessageRenderer.test.tsx   (test only)
packages/web/src/components/room/RuntimeMessageRenderer.tsx                   (definition only)
```

The task description says the component should render "inline between turn blocks in the TaskViewV2 feature." There is no `TaskViewV2` file in the tree. The component is not exported from `packages/web/src/components/room/index.ts`, nor from `packages/web/src/hooks/index.ts`. Without an integration point, the work is incomplete and the component is dead code from day one.

If the intent is to deliver the component as a standalone primitive ready for integration in a follow-up PR, that should be made explicit in the PR description and the task definition should be updated to reflect that scope boundary.

---

### P1 — `rate_limited` double-displays the reset time when both `text` and `resetsAt` are present

In `packages/web/src/components/room/RuntimeMessageRenderer.tsx` lines 57–59:
```tsx
{text}
{resetsAt ? ` Resets at ${resetsAt}.` : ''}
```

The daemon always includes the reset time inside the `text` field itself. For example, from `packages/daemon/src/lib/room/runtime/room-runtime.ts` line 709:
```ts
text: `Rate limit detected. Pausing until ${new Date(backoff.resetsAt).toLocaleTimeString()}.`,
resetsAt: backoff.resetsAt,
```

When both fields are present the rendered paragraph becomes:

> Rate limit detected. Pausing until 12:30:00 PM. Resets at 12:30:00 PM.

This is a copy-paste carry from the identical logic in `TaskConversationRenderer.tsx` lines 246–249, so it is a V1 bug inherited rather than introduced here. The fix is either to omit the `resetsAt` suffix (the time is already in `text`), or to strip the time from `text` in favour of the structured `resetsAt` field. The test at line 176–189 of the test file only checks `toContain('Resets at')`, which passes regardless of duplication, so the double-display is not caught.

---

### P2 — No test for `leader_summary` with a missing `text` field

The component falls back to an empty string (`const rawText = typeof raw.text === 'string' ? raw.text : '';`) and passes that to `MarkdownRenderer`. The empty-state render path is not tested. While the daemon always provides `text`, a test documenting the fallback behavior (the component renders without crash and shows an empty markdown block) would complete the coverage pattern established for the other three types.

---

### P3 — Nit: `resetsAt` time formatting is locale-dependent in tests

The test at line 177 uses `new Date('2025-01-01T12:30:00').getTime()` (local-time parse) and then only asserts `toContain('Resets at')`. This is fine for CI but the string produced by `toLocaleTimeString()` varies by locale and timezone, so a tighter assertion is not currently possible without mocking `Date.prototype.toLocaleTimeString`. No action required, just noting the design constraint.

lsm (commented):

### Recommendation: APPROVE

All three issues from the previous review have been properly addressed:

1. **P1 (resetsAt duplication) -- Fixed.** The `rate_limited` branch at `RuntimeMessageRenderer.tsx` lines 29-60 now renders only `{text}` without appending a "Resets at" suffix from the `resetsAt` field. The daemon already embeds the reset time in the `text` field, so this eliminates the double-display bug. The test at line 176-192 explicitly verifies no duplication occurs when both `text` and `resetsAt` are present.

2. **P2 (missing leader_summary empty-text test) -- Fixed.** The test at lines 341-351 covers the case where `leader_summary` has no `text` field, verifying the component renders the card and heading without crashing, and that `MarkdownRenderer` receives an empty string.

3. **P0 (component not exported) -- Fixed.** `RuntimeMessageRenderer` is now exported from `packages/web/src/components/room/index.ts` at line 22, making it available for integration by the TaskViewV2 parent component in a follow-up PR.

- **Stale test file comment**: The doc comment at `RuntimeMessageRenderer.test.tsx` line 8 still mentions "resetsAt" as part of the rate_limited test coverage (`amber notification card with role, text, resetsAt`), but the component no longer uses `resetsAt`. This is cosmetic only and does not affect functionality.

No P0, P1, or P2 issues found. The implementation is correct, complete for its scope, well-tested with 29 passing tests, and ready for integration.

lsm (commented):

**Recommendation: APPROVE**

All three items from the previous review have been addressed correctly. The `resetsAt` duplication is gone, the `leader_summary` missing-text test is present, and `RuntimeMessageRenderer` is now exported from the room `index.ts`.

---

### What is working well

- All four message types (`status`, `rate_limited`, `model_fallback`, `leader_summary`) are handled with proper fallbacks for every field.
- The `[Turn Summary]` prefix-stripping logic is correct and tested for both the positive and negative case.
- The `leader_summary` no-text test (`{ type: 'leader_summary' }`) correctly verifies that the card renders and `MarkdownRenderer` receives an empty string.
- The new regression test at line 176–192 in the test file verifies the text is rendered as-is without a duplicate reset-time suffix.
- Tests cover `data-testid`, visual class names, fallback values, and edge cases — solid coverage for a pure rendering component.

---

### Issues

#### P3 — Stale doc comment in test file (line 6)

`packages/web/src/components/room/__tests__/RuntimeMessageRenderer.test.tsx:6`

```
- rate_limited: amber notification card with role, text, resetsAt
```

The `resetsAt` suffix was intentionally removed from rendering (it is now ignored), but this file-level JSDoc bullet still lists it as a rendered field. It should read something like `role, text (resetsAt is embedded by the daemon in text)` or simply drop `resetsAt`.

---

### Informational — no parent consumer yet

`RuntimeMessageRenderer` is exported and the `useTurnBlocks` hook is wired to produce `RuntimeMessage` items, but no `TaskViewV2` component exists yet to consume them together. That appears to be intentional scope for a follow-up PR, and is consistent with the task title ("implement RuntimeMessageRenderer component"). No action required here, just noting it for completeness.

PR #748, opened 2026-03-23
feat: implement TaskViewToggle and integrate into Room.tsx

lsm:
- Add TaskViewToggle wrapper that renders a V1/V2 header toggle bar above
  the active task view, reading/persisting preference via localStorage key
  'neokai:taskViewVersion' with synchronous lazy initializer to prevent flicker
- Default to V1 for backward compatibility
- Replace <TaskView> with <TaskViewToggle> in Room.tsx (minimal import/component swap)
- 13 unit tests covering toggle, persistence, prop forwarding, and aria-label

### DIFF
--- a/packages/web/src/components/room/TaskViewToggle.tsx
+++ b/packages/web/src/components/room/TaskViewToggle.tsx
@@ -1,0 +1,66 @@
+/**
+ * TaskViewToggle Component
+ *
+ * Wrapper that renders a V1/V2 toggle header bar above the active task view.
+ * Reads and persists the user's preference via localStorage.
+ *
+ * Default: 'v1' for backward compatibility.
+ */
+
+import { useState } from 'preact/hooks';
+import { TaskView } from './TaskView';
+import { TaskViewV2 } from './TaskViewV2';
+
+const STORAGE_KEY = 'neokai:taskViewVersion';
+
+interface TaskViewToggleProps {
+	roomId: string;
+	taskId: string;
+}
+
+export function TaskViewToggle({ roomId, taskId }: TaskViewToggleProps) {
+	// Synchronous lazy initializer prevents visible flicker on load
+	const [version, setVersion] = useState<'v1' | 'v2'>(
+		() => (localStorage.getItem(STORAGE_KEY) as 'v1' | 'v2') || 'v1'
+	);
+
+	const handleToggle = () => {
+		const next = version === 'v1' ? 'v2' : 'v1';
+		setVersion(next);
+		localStorage.setItem(STORAGE_KEY, next);
+	};
+
+	return (
+		<div class="flex-1 flex flex-col overflow-hidden">
+			{/* Toggle header bar */}
+			<div class="flex items-center justify-end gap-2 px-3 py-1 bg-dark-850 border-b border-dark-700">
+				<span class="text-xs text-gray-500 select-none">View:</span>
+				<button
+					data-testid="task-view-toggle"
+					onClick={handleToggle}
+					class="flex items-center gap-1 px-2 py-0.5 rounded text-xs font-medium transition-colors bg-dark-700 hover:bg-dark-600 text-gray-300 hover:text-gray-100"
+					aria-label={`Switch to ${version === 'v1' ? 'V2 turn-based' : 'V1 timeline'} view`}
+				>
+					{version === 'v1' ? (
+						<>
+							<span>V1</span>
+							<span class="text-gray-500">→ V2</span>
+						</>
+					) : (
+						<>
+							<span class="text-gray-500">V1 ←</span>
+							<span>V2</span>
+						</>
+					)}
+				</button>
+			</div>
+
+			{/* Active view */}
+			{version === 'v1' ? (
+				<TaskView roomId={roomId} taskId={taskId} />
+			) : (
+				<TaskViewV2 roomId={roomId} taskId={taskId} />
+			)}
+		</div>
+	);
+}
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped implementation that meets all acceptance criteria. The Room.tsx change is minimal (import + component swap), the toggle works synchronously via lazy initializer, localStorage persistence uses the correct key, `data-testid` attributes are correctly placed, and V1 is the default. Tests are thorough with 13 cases covering all specified behavior.

### P2: Invalid localStorage value causes empty render (1 issue)

**`packages/web/src/components/room/TaskViewToggle.tsx:24`** — The lazy initializer casts the localStorage value with `as 'v1' | 'v2'` and falls back with `|| 'v1'`, but this only catches `null`/empty string. If localStorage contains an unexpected truthy value (e.g., `"v3"` from manual editing or a future migration), the cast silently passes through and `version` becomes `"v3"`. The ternary on line 59 (`version === 'v1' ? ... : ...`) would render TaskViewV2 in that case — which is actually fine behavior since the else branch catches it. **On closer inspection, this is not a bug** — any non-`"v1"` value falls through to the V2 branch. This is acceptable, though a stricter validation would be marginally better:

```ts
() => {
  const stored = localStorage.getItem(STORAGE_KEY);
  return stored === 'v2' ? 'v2' : 'v1';
}
```

This is a very minor robustness improvement, not blocking.

### Summary

All acceptance criteria verified:
- [x] V1/V2 toggle switches views without page reload
- [x] Preference persisted in localStorage (`neokai:taskViewVersion`) and restored on load
- [x] Default is V1 (backward compatible)
- [x] Toggle rendered above both views (not inside V1 header)
- [x] Room.tsx change is minimal (import/component swap only)
- [x] `data-testid="task-view-toggle"` on toggle button
- [x] `data-testid="task-view-v2"` set by TaskViewV2 itself (not by wrapper)
- [x] Synchronous lazy initializer (no useEffect flicker)
- [x] All 13 unit tests pass

Clean implementation with good test coverage. No blocking issues.

lsm (commented):

**Recommendation: APPROVE** (P3 nit only)

---

### Summary

The implementation correctly satisfies all acceptance criteria. The component is minimal, correct, and well-tested. The one issue worth noting is an inconsistency with the project's established defensive coding pattern for localStorage, but it does not affect correctness in normal browser environments.

---

### What was reviewed

- `packages/web/src/components/room/TaskViewToggle.tsx` — new wrapper component
- `packages/web/src/components/room/__tests__/TaskViewToggle.test.tsx` — 13 unit tests
- `packages/web/src/islands/Room.tsx` — import/component swap

---

### Correctness

All acceptance criteria are met:

- V1/V2 toggle switches views without page reload — confirmed, uses `useState` + conditional render.
- Preference is persisted to `localStorage` key `neokai:taskViewVersion` and restored on load — confirmed.
- Default is `v1` — confirmed by the `|| 'v1'` fallback.
- Synchronous lazy initializer — confirmed, `useState(() => localStorage.getItem(...))` not a `useEffect`.
- Toggle bar rendered above both views — confirmed, sits outside the conditional view block.
- `data-testid="task-view-toggle"` on the button — confirmed.
- `data-testid="task-view-v2"` set by `TaskViewV2` itself — confirmed at line 187 of `TaskViewV2.tsx`; not added by the wrapper.
- Room.tsx change is minimal — confirmed, only the import line and the JSX tag change.

### Tests

All 13 tests pass (verified locally). Full web suite 5356/5356 passes with no regression. Coverage includes: default V1, read V1/V2 from storage, toggle V1→V2 and V2→V1, persistence to storage, prop forwarding for both views, aria-label content, and storage key used.

One missing test scenario (P3): no test covers what happens when localStorage throws (private browsing / security restriction). See below.

---

### Issues

#### P3 — localStorage calls are unguarded (inconsistent with project pattern)

**File:** `packages/web/src/components/room/TaskViewToggle.tsx`, lines 24 and 30

The project's established pattern wraps every localStorage access in try/catch. Both `recent-paths.ts` and `SpaceIsland.tsx` do this with a silent-fail fallback, e.g.:

```ts
// recent-paths.ts
try {
  const stored = localStorage.getItem(STORAGE_KEY);
  ...
} catch {
  // Silently fail if localStorage is unavailable
}

// SpaceIsland.tsx
try {
  localStorage.setItem(EDITOR_MODE_KEY, mode);
} catch {
  ...
}
```

`TaskViewToggle` is the only component in the codebase that reads/writes localStorage without a guard. In browsers with strict privacy settings (Safari ITP, Firefox with enhanced tracking protection, some private-browsing implementations), `localStorage.getItem` can throw a `SecurityError`, which would crash the component with an unhandled exception and break the entire Room view. The fix is to wrap both calls the same way the rest of the codebase does.

The test suite does not cover a throwing localStorage scenario either.

This is a P3 because the primary target environments (Chromium-based browsers in dev/prod) rarely exhibit this behaviour, and V1 remains the default so the user can still navigate away; but it is a real divergence from the project's own defensive pattern and worth fixing before V2 is considered stable.

---

- The `as 'v1' | 'v2'` cast on line 24 means that an arbitrary stored string like `'v3'` or `'corrupted'` would be accepted as the initial state but then evaluate to the `version === 'v1' ? ...` false branch, silently rendering V2. This is harmless given the binary nature of the toggle but worth knowing.
- Room.tsx `key={taskViewId}` on `TaskViewToggle` is correctly preserved from the original, so switching tasks resets local state including the toggle. That is the right behaviour.
- E2E tests referencing `task-messages-container`, `task-view-goal-badge`, and `task-status-badge` are unaffected: they use page.locator which searches the full DOM subtree, and the new wrapper div does not alter the nested test IDs.

PR #758, opened 2026-03-23
feat: give Task Agent the send_message tool via channel topology

lsm:
## Summary

- **send_message MCP tool**: Added to Task Agent MCP server using the same `SendMessageSchema` as step agents. Task Agent uses `'task-agent'` as its role for channel topology validation.
- **Default bidirectional channels**: Auto-generated between `task-agent` and every node agent role at step-start in `storeResolvedChannels`. This ensures the Task Agent can always reach peer step agents without manual channel configuration.
- **System prompt documentation**: Updated to document the new `send_message` tool, default channels behavior, and removal implications.

## Test plan

- [x] All 1335 space unit tests pass
- [x] Typecheck and lint pass
- [x] New tests for Task Agent send_message (point-to-point, broadcast, multicast, default channels, error cases)
- [x] Updated existing tests for new auto-generated channel behavior

## Breaking changes

None — this is an additive feature. Existing channel topology declarations are preserved; default channels are only added when no user-declared channel exists for a given task-agent→node role pair.

Refs: TASK-33

### DIFF
--- a/packages/daemon/src/lib/space/agents/task-agent.ts
+++ b/packages/daemon/src/lib/space/agents/task-agent.ts
@@ -20,6 +20,7 @@
  *   - report_result         — Mark the task complete/failed and record the result summary
  *   - request_human_input   — Surface a human gate and block until the user responds
  *   - list_group_members    — List all group members with session IDs and permitted channels
+ *   - send_message          — Send a message to peer step agents via channel topology
  *
  * ## Content interpolation
  * All operator-supplied content (space.backgroundContext, space.instructions,
@@ -209,6 +210,12 @@ export function buildTaskAgentSystemPrompt(context: TaskAgentContext): string {
 			`\`permittedTargets\` (which roles they can message per the declared channel topology). ` +
 			`Use this to discover active sub-sessions and their communication permissions.`
 	);
+	sections.push(
+		`- **send_message** — Send a message directly to one or more peer step agents via declared channel topology. ` +
+			`Supports point-to-point (\`coder\`), broadcast (\`*\`), and multicast (\`[coder, reviewer]\`). ` +
+			`The Task Agent has default bidirectional channels to all node agents, so it can always ` +
+			`message any peer step agent. Use \`list_group_members\` to see permitted targets.`
+	);
 
 	// ---- step_result vs report_result status ---------------------------------
 	sections.push(`\n## Result Fields: \`step_result\` vs \`report_result.status\`\n`);
@@ -291,7 +298,12 @@ export function buildTaskAgentSystemPrompt(context: TaskAgentContext): string {
 	);
 	sections.push(
 		`Channel topology is enforced uniformly across all agents. ` +
-			`(Note: peer messaging via \`send_message\` will be added in a follow-up task.)\n`
+			`The Task Agent uses \`send_message\` for all inter-agent communication with step agents.\n`
+	);
+	sections.push(
+		`Default channels: the Task Agent has default bidirectional channels to all node agent roles, ` +
+			`so it can always reach any peer step agent. If a user removes a Task Agent channel, ` +
+			`the Task Agent can no longer message that node until the channel is restored.\n`
 	);
 
 	// ---- Task context -------------------------------------------------------
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -362,7 +362,7 @@ export class SpaceRuntime {
 		// Resolve channel topology for the start step and store in run config.
 		// TODO: Milestone 6: pass resolvedChannels to session group creation in
 		// TaskAgentManager.spawnTaskAgent() rather than storing in run config.
-		this.storeResolvedChannels(run.id, space.id, startStep);
+		this.resolveAndStoreChannels(run.id, space.id, startStep);
 
 		return { run, tasks };
 	}
@@ -780,7 +780,7 @@ export class SpaceRuntime {
 			} else {
 				// Resolve channel topology for the new step.
 				// TODO: Milestone 6: pass to session group creation instead of run config.
-				this.storeResolvedChannels(runId, meta.spaceId, newStep);
+				this.resolveAndStoreChannels(runId, meta.spaceId, newStep);
 			}
 		} catch (err) {
 			if (!(err instanceof WorkflowTransitionError)) {
@@ -1010,7 +1010,7 @@ export class SpaceRuntime {
 	 * Steps with no `channels` declaration still get default task-agent channels,
 	 * so the Task Agent always has peer communication ability.
 	 */
-	private storeResolvedChannels(runId: string, spaceId: string, step: WorkflowStep): void {
+	resolveAndStoreChannels(runId: string, spaceId: string, step: WorkflowStep): void {
 		const run = this.config.workflowRunRepo.getRun(runId);
 		if (!run) return;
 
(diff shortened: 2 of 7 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** (posted as COMMENT — same author as reviewer)

---

## Summary

The core feature — adding `send_message` to the Task Agent MCP toolset and auto-generating default bidirectional channels at step-start — is correctly implemented and well-tested for the happy path. However, there is one P0 correctness bug, one P1 duplicate-channel issue, and a missing test gap around the primary scenario.

---

## P0 — Default channels not regenerated when Task Agent calls `advance_workflow` on a multi-step workflow

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts` (the `advance_workflow` handler, around line 637)
**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts` (the `storeResolvedChannels` private method, line 1013)

### What happens

`storeResolvedChannels` is called in exactly two places:
1. `startWorkflowRun()` — line 365 (start step, always called)
2. The non-TAM `processRunTick()` path — line 783 (after direct `freshExecutor.advance()` in the tick loop)

When a Task Agent calls the `advance_workflow` MCP tool, `executor.advance()` is called inside `task-agent-tools.ts` (line 637). **`storeResolvedChannels` is never called on this path.** Because `storeResolvedChannels` is `private`, the tool handler cannot call it either.

### Effect

In a multi-step workflow where step 1 assigns agent role `coder` and step 2 assigns agent role `reviewer`:

- After `startWorkflowRun()`, `_resolvedChannels` correctly contains the default `task-agent↔coder` channels for step 1.
- The Task Agent completes step 1 and calls `advance_workflow`.
- `executor.advance()` runs, creating the step-2 task for the `reviewer` agent.
- **`storeResolvedChannels` is NOT called.** `_resolvedChannels` still reflects step 1's topology (task-agent↔coder channels only).
- The Task Agent tries to `send_message` to the `reviewer` — the channel resolver reads the stale channels and rejects it as unauthorized.

This means the core acceptance criterion — "Task Agent can always reach any peer step agent" — silently breaks on every step transition in Task-Agent-driven workflows. The feature works for step 1 but fails for all subsequent steps.

### Evidence

The test at `space-runtime.test.ts:2124` exercises `storeResolvedChannels` only via `runtime.executeTick()` (the non-TAM tick path). There is no test that calls the `advance_workflow` MCP tool handler and then checks whether `_resolvedChannels` reflects the new step's topology.

---

## P1 — Duplicate default channels when a step has two agents with the same role

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 1035–1068

In `storeResolvedChannels`, `existingPairs` is only populated from `userResolved` (user-declared channels). It is never updated with entries added to `defaultChannels` during the loop. When `nodeRoles` contains duplicates (e.g., a step with two agents that share the same role), both iterations of the loop check against the same `existingPairs` set — the second iteration passes the `!existingPairs.has(...)` guard and pushes a second identical entry into `defaultChannels`.

This produces duplicate entries in `_resolvedChannels`. While `ChannelResolver.canSend()` is tolerant (it uses `.some()` so duplicates do not break routing), the stored DB JSON grows unnecessarily. More importantly, this is a latent correctness risk if the resolver ever needs to iterate the channel list for purposes where duplicates matter.

Fix: Either deduplicate `nodeRoles` before the loop, or update `existingPairs` after each new channel is pushed.

---

## P2 — Default channel logic uses a local constant conflating role and agentId

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, line 1043

`const TASK_AGENT_ID = 'task-agent';` is declared as a local constant inside `storeResolvedChannels`. The same string literal `'task-agent'` is used as both a role identifier and as a placeholder `agentId`. This conflates two distinct concepts — agent role and agent ID — with the same string. The comment documents the intent ("canSend only checks roles"), but if the channel resolver ever starts using `fromAgentId`/`toAgentId` for enforcement, this silent conflation will produce incorrect behavior without any compile-time or runtime warning.

---

## P3 — Test for default channels uses manual channel injection rather than exercising `storeResolvedChannels`

**File:** `packages/daemon/tests/unit/space/cross-agent-messaging.test.ts`, lines 1339–1455

The tests in "Task Agent send_message — default task-agent channels" verify messaging behavior after manually constructing the `_resolvedChannels` payload and writing it directly via `workflowRunRepo.updateRun`. This correctly tests the resolver, but does not exercise `storeResolvedChannels`. A regression in channel generation logic would not be caught by these tests.

---

## What is Working Well

- The `send_message` handler in `task-agent-tools.ts` uses `'task-agent'` as the role for channel validation — correct and consistent.
- Broadcast (`*`), multicast (array), and point-to-point target modes are all handled correctly with proper topology enforcement.
- The `ChannelResolver` design (role-based, not ID-based) is clean and appropriate.
- The system prompt in `task-agent.ts` accurately documents the `send_message` tool and default channels behavior.
- Test coverage for the Task Agent messaging scenarios is comprehensive for happy paths and error cases.
- The deduplication logic in `storeResolvedChannels` correctly avoids re-adding user-declared channels.
- The test for "removing a Task Agent channel prevents messaging" is a strong explicit regression guard.
- The `list_group_members` update correctly shows `permittedTargets` per member.

lsm (commented):

**Recommendation: APPROVE**

---

## Summary

All three substantive issues from the previous review (P0, P1, P2) have been addressed correctly. The implementation is now sound. One minor test gap remains (P1 below) but it is not blocking.

---

## P0 Fix — advance_workflow now calls resolveAndStoreChannels: CORRECT

The fix makes `storeResolvedChannels` public as `resolveAndStoreChannels` and calls it inside the `advance_workflow` tool handler after a successful `executor.advance()` (task-agent-tools.ts line 661):

```ts
const run = workflowRunRepo.getRun(workflowRunId);
if (run) {
    runtime.resolveAndStoreChannels(workflowRunId, run.spaceId, nextStep);
}
```

The guard `if (newTasks.length === 0)` (the terminal branch) correctly skips the call. The TAM path and the non-TAM tick path remain mutually exclusive: when `taskAgentManager` is configured, `processRunTick` returns early at line 763 before reaching its own `resolveAndStoreChannels` call at line 783 — no double-call.

The `run.spaceId` field is present on `SpaceWorkflowRun`, so the call is type-safe.

---

## P1 Fix — Deduplication via new Set(): CORRECT

`nodeRoles` is now built through `new Set()` before the channel generation loop, meaning each unique role appears exactly once and generates exactly one bidirectional pair. The `existingPairs` set (built from `userResolved`) still prevents re-adding user-declared channels. The combination of the two guards is correct.

---

## P2 Fix — agentId placeholder comment: ADEQUATE

Lines 1047-1051 in space-runtime.ts explain that agentId fields use the role string as a placeholder and document the future-proofing concern. This is sufficient for the current state of the code.

---

## P3 Fix — New deduplication test: PRESENT AND CORRECT

The test at space-runtime.test.ts line 2124 seeds two agents with the same role, runs `startWorkflowRun`, and asserts exactly one `task-agent→coder` and one `coder→task-agent` channel entry. This correctly validates the P1 fix.

---

## Remaining gap (informational, not blocking)

There is no end-to-end test in `task-agent-tools.test.ts` that:
1. Creates a two-step workflow with different agents per step (e.g., step 1: planner, step 2: coder)
2. Calls `advance_workflow` via the tool handler
3. Asserts that `_resolvedChannels` in the run config was updated to reflect step 2's topology

The P0 fix is implemented correctly, but the specific exercise of that code path (Task Agent calls `advance_workflow`, channels get refreshed) has no dedicated integration test. The correctness of the fix is validated by reading the code and by the fact that `resolveAndStoreChannels` itself is well-tested in isolation (space-runtime.test.ts). This is a minor test coverage gap, not a correctness issue.

---

## What is Working Well

- The two call sites of `resolveAndStoreChannels` are now consistent: one for the initial step (in `startWorkflowRun`), one for each subsequent step (in `advance_workflow` for the TAM path, in `processRunTick` for the non-TAM path). No stale topology.
- The method rename from `storeResolvedChannels` to `resolveAndStoreChannels` is a good name that reflects both actions (resolving and storing).
- All four space-runtime.test.ts channel topology tests are well-written, including the new deduplication test.
- The PR description accurately describes all the changes.

PR #759, opened 2026-03-23
test: add E2E tests for Task Agent visible node in visual editor

lsm:
## Summary

Added 5 E2E tests (tests 7-11) to `packages/e2e/tests/features/visual-workflow-editor.e2e.ts` for the Task Agent visible node feature (Change 3 of the workflow graph model goal).

**Current Status: Tests are SKIPPED**

The tests are currently wrapped in `test.skip()` because the Task Agent visible node feature has not been implemented yet. They will pass once Tasks 4.1/4.2/4.3 implement the feature.

## Test Coverage

- **Test 7**: Task Agent node is always visible when opening the visual editor
- **Test 8**: Task Agent node cannot be deleted via keyboard
- **Test 9**: Adding a new node shows Task Agent channel edges on the canvas
- **Test 10**: Task Agent node has distinct visual styling (purple border, pinned-node class)
- **Test 11**: Channels to Task Agent can be removed via config panel

## Implementation Requirements

The tests assume the following implementation details (to be verified against actual implementation):

- `data-testid="workflow-node-task-agent"` on the Task Agent pinned node
- `data-task-agent-node="true"` attribute on the node
- `border-purple-500` CSS class for distinct styling
- `pinned-node` CSS class for pinned marker
- Task Agent node always present at canvas open (0 regular nodes initially)
- `ChannelTopologyBadge` truncates "task-agent" to "task-age…" (8 char limit)
- `channels-section` in `NodeConfigPanel` for multi-agent steps showing task-agent channels

## Verification

All 6 existing tests pass. The 5 new tests are properly skipped.

### DIFF
--- a/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
+++ b/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
@@ -7,6 +7,7 @@
  * - Load template in visual editor (auto-layout with nodes and edges)
  * - Toggle between List and Visual modes (mode switching with confirmation)
  * - Validation errors when saving incomplete workflows
+ * - Task Agent pinned node visibility and interactions (Change 3 of workflow graph model)
  *
  * Setup: creates a Space via RPC in beforeEach (infrastructure).
  * Cleanup: deletes the Space via RPC in afterEach.
@@ -15,6 +16,13 @@
  * - All test actions go through the UI (clicks, inputs, navigation)
  * - All assertions check visible DOM state
  * - RPC is only used in beforeEach/afterEach for test infrastructure
+ *
+ * Task Agent Node Tests:
+ * - Task Agent node is always visible when opening the visual editor
+ * - Task Agent node cannot be deleted via keyboard
+ * - Adding a new node shows Task Agent channel edges on the canvas
+ * - Task Agent node has distinct visual styling (purple border, pinned-node class)
+ * - Channels to Task Agent can be removed via the config panel
  */
 
 import type { Page } from '@playwright/test';
@@ -424,5 +432,156 @@ test.describe('Visual Workflow Editor', () => {
 
 		// Editor should remain open
 		await expect(editor).toBeVisible();
+	});
+
+	// ─── Task Agent Node Tests ───────────────────────────────────────────────
+
+	// ─── Test 7: Task Agent node is always visible when opening the visual editor ───
+
+	test('Task Agent node is always visible when opening the visual editor', async ({ page }) => {
+		await navigateToSpace(page, spaceId);
+		await openNewWorkflowEditor(page);
+		await switchToVisualMode(page);
+
+		const editor = page.getByTestId('visual-workflow-editor');
+
+		// The Task Agent pinned node should be visible even on an empty canvas
+		// It has a distinct data attribute to identify it
+		const taskAgentNode = editor.locator('[data-testid="workflow-node-task-agent"]');
+		await expect(taskAgentNode).toBeVisible({ timeout: 5000 });
+
+		// The Task Agent node should also have the data-task-agent-node attribute
+		await expect(taskAgentNode).toHaveAttribute('data-task-agent-node', 'true');
+
+		// Task Agent node should display "Task Agent" as its name
+		await expect(taskAgentNode.locator('text=Task Agent')).toBeVisible({ timeout: 2000 });
+	});
+
+	// ─── Test 8: Task Agent node cannot be deleted via keyboard ─────────────
+
+	test('Task Agent node cannot be deleted via keyboard', async ({ page }) => {
+		await navigateToSpace(page, spaceId);
+		await openNewWorkflowEditor(page);
+		await switchToVisualMode(page);
+
+		const editor = page.getByTestId('visual-workflow-editor');
+
+		// First add a regular node so we have something to select and potentially delete
+		await editor.getByTestId('add-step-button').click();
+		await expect(editor.locator('[data-testid^="workflow-node-"]')).toHaveCount(1, {
+			timeout: 3000,
+		});
+
+		// Select the Task Agent node
+		const taskAgentNode = editor.locator('[data-testid="workflow-node-task-agent"]');
+		await taskAgentNode.click();
+
+		// Try to delete via Delete key
+		await page.keyboard.press('Delete');
+
+		// Task Agent node should still be visible after Delete key
+		await expect(taskAgentNode).toBeVisible({ timeout: 2000 });
+	});
+
+	// ─── Test 9: Adding a new node shows Task Agent channel edges on the canvas ───
+
+	test('Adding a new node shows Task Agent channel edges on the canvas', async ({ page }) => {
+		await navigateToSpace(page, spaceId);
+		await openNewWorkflowEditor(page);
+		await switchToVisualMode(page);
+
+		const editor = page.getByTestId('visual-workflow-editor');
+
+		// The Task Agent node should already be visible
+		const taskAgentNode = editor.locator('[data-testid="workflow-node-task-agent"]');
+		await expect(taskAgentNode).toBeVisible({ timeout: 5000 });
+
+		// Add a new node via "Add Step" button
+		await editor.getByTestId('add-step-button').click();
+
+		// Wait for the node to appear
+		const nodes = editor.locator('[data-testid^="workflow-node-"]');
+		await expect(nodes).toHaveCount(1, { timeout: 3000 });
+
+		// Click the new node to select it and open the config panel
+		await nodes.first().click();
+		await expect(editor.getByTestId('node-config-panel')).toBeVisible({ timeout: 3000 });
+
+		// The ChannelTopologyBadge should show a channel to/from Task Agent
+		// Look for "task-agent" text in the channel topology badge
+		const channelBadge = editor.getByTestId('channel-topology-badge');
+		await expect(channelBadge).toBeVisible({ timeout: 2000 });
+		await expect(channelBadge.locator('text=task-agent')).toBeVisible({ timeout: 2000 });
+	});
+
+	// ─── Test 10: Task Agent node has distinct visual styling ───────────────
+
+	test('Task Agent node has distinct visual styling', async ({ page }) => {
+		await navigateToSpace(page, spaceId);
+		await openNewWorkflowEditor(page);
+		await switchToVisualMode(page);
+
+		const editor = page.getByTestId('visual-workflow-editor');
+
+		// The Task Agent node should have a distinct border color (purple)
+		const taskAgentNode = editor.locator('[data-testid="workflow-node-task-agent"]');
+		await expect(taskAgentNode).toBeVisible({ timeout: 5000 });
+
+		// Check for the distinct styling class - Task Agent uses purple border
+		// The Task Agent node should have border-purple-500 class
+		await expect(taskAgentNode).toHaveClass(/border-purple-500/);
+
+		// Also verify it has the pinned-node marker class
+		await expect(taskAgentNode).toHaveClass(/pinned-node/);
+	});
+
+	// ─── Test 11: Channels to Task Agent can be removed via config panel ───
+
+	test('Channels to Task Agent can be removed via config panel', async ({ page }) => {
+		await navigateToSpace(page, spaceId);
+		await openNewWorkflowEditor(page);
+		await switchToVisualMode(page);
+
+		const editor = page.getByTestId('visual-workflow-editor');
+
+		// Add a node that will have a Task Agent channel
+		await editor.getByTestId('add-step-button').click();
+
+		// Wait for node and select it to open config panel
+		const nodes = editor.locator('[data-testid^="workflow-node-"]');
+		await expect(nodes).toHaveCount(1, { timeout: 3000 });
+
+		// Configure the node with an agent first
+		await nodes.first().click();
+		await expect(editor.getByTestId('node-config-panel')).toBeVisible({ timeout: 3000 });
+		await editor.getByTestId('agent-select').selectOption({ index: 1 });
+		await editor.getByTestId('close-button').click();
+
+		// Now add a second node
+		await editor.getByTestId('add-step-button').click();
+		await expect(nodes).toHaveCount(2, { timeout: 3000 });
+
+		// Select the second node
+		await nodes.nth(1).click();
+		await expect(editor.getByTestId('node-config-panel')).toBeVisible({ timeout: 3000 });
+
+		// Verify Task Agent channel exists in the badge
+		const channelBadge = editor.getByTestId('channel-topology-badge');
+		await expect(channelBadge).toBeVisible({ timeout: 2000 });
+
+		// Find and click the remove button for the Task Agent channel
+		// The channel entry should contain "task-agent"
+		const taskAgentChannelEntry = channelBadge.locator('[data-testid="channel-entry"]').filter({
+			has: page.locator('text=task-agent'),
+		});
+		await expect(taskAgentChannelEntry).toBeVisible({ timeout: 2000 });
+
+		// Click the remove button on that channel entry
+		const removeBtn = taskAgentChannelEntry.locator('[data-testid="remove-channel-button"]');
+		await expect(removeBtn).toBeVisible({ timeout: 2000 });
+		await removeBtn.click();
+
+		// Verify the Task Agent channel is no longer in the badge
+		await expect(channelBadge.locator('text=task-agent')).not.toBeVisible({ timeout: 2000 });
 	});
 });
--- a/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
+++ b/packages/e2e/tests/features/visual-workflow-editor.e2e.ts
@@ -7,7 +7,6 @@
  * - Load template in visual editor (auto-layout with nodes and edges)
  * - Toggle between List and Visual modes (mode switching with confirmation)
  * - Validation errors when saving incomplete workflows
- * - Task Agent pinned node visibility and interactions (Change 3 of workflow graph model)
  *
  * Setup: creates a Space via RPC in beforeEach (infrastructure).
  * Cleanup: deletes the Space via RPC in afterEach.
@@ -17,12 +16,16 @@
  * - All assertions check visible DOM state
  * - RPC is only used in beforeEach/afterEach for test infrastructure
  *
- * Task Agent Node Tests:
+ * Task Agent Node Tests (SKIPPED - feature not yet implemented):
  * - Task Agent node is always visible when opening the visual editor
  * - Task Agent node cannot be deleted via keyboard
  * - Adding a new node shows Task Agent channel edges on the canvas
  * - Task Agent node has distinct visual styling (purple border, pinned-node class)
  * - Channels to Task Agent can be removed via the config panel
+ *
+ * These tests are skipped because the Task Agent visible node feature (Change 3 of
+ * the workflow graph model goal) has not been implemented yet. Once the feature is
+ * implemented (Tasks 4.1, 4.2, 4.3), remove the test.skip() annotations.
  */
 
 import type { Page } from '@playwright/test';
@@ -435,10 +438,27 @@ test.describe('Visual Workflow Editor', () => {
 	});
 
 	// ─── Task Agent Node Tests ───────────────────────────────────────────────
+	// These tests are skipped because the Task Agent visible node feature
+	// (Change 3 of the workflow graph model goal) has not been implemented yet.
+	// They require:
+	// - Task Agent rendered as a pinned node in VisualWorkflowEditor
+	// - data-testid="workflow-node-task-agent" attribute on the node
+	// - data-task-agent-node="true" attribute on the node
+	// - border-purple-500 CSS class for distinct styling
+	// - pinned-node CSS class for the pinned marker
+	// - Task Agent node always present at canvas open (not added via add-step-button)
+	// - ChannelTopologyBadge showing task-agent channels on regular nodes
+	// - channels-section in NodeConfigPanel for multi-agent steps showing task-agent channels
+	//
+	// Once the feature is implemented (Tasks 4.1, 4.2, 4.3), remove the test.skip()
+	// and update selectors if the implementation differs from the assumptions below.
+	// See: https://github.com/lsm/neokai/issues (tracking issue for Task Agent visible node)
 
 	// ─── Test 7: Task Agent node is always visible when opening the visual editor ───
 
-	test('Task Agent node is always visible when opening the visual editor', async ({ page }) => {
+	test.skip('Task Agent node is always visible when opening the visual editor', async ({
+		page,
+	}) => {
 		await navigateToSpace(page, spaceId);
 		await openNewWorkflowEditor(page);
 		await switchToVisualMode(page);
@@ -459,7 +479,7 @@ test.describe('Visual Workflow Editor', () => {
 
 	// ─── Test 8: Task Agent node cannot be deleted via keyboard ─────────────
 
-	test('Task Agent node cannot be deleted via keyboard', async ({ page }) => {
+	test.skip('Task Agent node cannot be deleted via keyboard', async ({ page }) => {
 		await navigateToSpace(page, spaceId);
 		await openNewWorkflowEditor(page);
 		await switchToVisualMode(page);
@@ -487,7 +507,7 @@ test.describe('Visual Workflow Editor', () => {
 
 	// ─── Test 9: Adding a new node shows Task Agent channel edges on the canvas ───
 
-	test('Adding a new node shows Task Agent channel edges on the canvas', async ({ page }) => {
+	test.skip('Adding a new node shows Task Agent channel edges on the canvas', async ({ page }) => {
 		await navigateToSpace(page, spaceId);
 		await openNewWorkflowEditor(page);
 		await switchToVisualMode(page);
@@ -499,7 +519,7 @@ test.describe('Visual Workflow Editor', () => {
 		await expect(taskAgentNode).toBeVisible({ timeout: 5000 });
 
 		// Count nodes before adding (should be 1 - just the Task Agent pinned node)
-		let nodes = editor.locator('[data-testid^="workflow-node-"]');
+		const nodes = editor.locator('[data-testid^="workflow-node-"]');
 		await expect(nodes).toHaveCount(1, { timeout: 3000 });
 
 		// Add a new node via "Add Step" button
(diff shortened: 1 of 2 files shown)


lsm (commented):

---

**Recommendation: REQUEST_CHANGES** — The new tests (7–11) reference UI elements and behaviors that do not exist in the current production code. They will fail in CI the moment they are executed against the live server. The PR description acknowledges this ("The Task Agent visible node feature needs to be implemented before these tests will pass"), but the CI contract in CLAUDE.md is clear: if a test appears in the test matrix, it must pass. Shipping known-broken tests creates noise and may block unrelated merges.

---

### P0 — Tests assert selectors/attributes that are not in the codebase

**1. `[data-testid="workflow-node-task-agent"]` does not exist**

`WorkflowNode` generates `data-testid={`workflow-node-${stepId}`}` where `stepId` is a UUID produced by `generateUUID()`. There is no special-cased Task Agent pinned node in `VisualWorkflowEditor`, `WorkflowCanvas`, or `WorkflowNode`. Tests 7, 8, 9, 10, and 11 all locate by this selector and will immediately time out.

*File: `packages/web/src/components/space/visual-editor/WorkflowNode.tsx` line 253*
```
data-testid={`workflow-node-${stepId}`}
```

**2. `data-task-agent-node="true"` attribute is never set**

Test 7 asserts `toHaveAttribute('data-task-agent-node', 'true')`. This attribute does not appear in any component in the web package.

**3. CSS classes `border-purple-500` and `pinned-node` are not applied to workflow nodes**

`WorkflowNode` only applies three border variants:
- `border-green-500` (start node)
- `border-blue-500` (selected)
- `border-gray-700` (default)

The string `pinned-node` appears nowhere in the visual-editor directory. Test 10 will fail both assertions.

**4. Initial canvas node count is 0, not 1**

Tests 7, 9, and 11 each assert that the canvas already contains 1 node before any `add-step-button` click. The current `VisualWorkflowEditor` starts with `nodes = []` and shows an empty-state overlay. The assertions `toHaveCount(1)` will fail immediately.

---

### P0 — Structural mismatch between `channel-topology-badge` and `node-config-panel`

**5. Test 9 and 11: `channel-topology-badge` is on the canvas node card, not in the config panel**

`ChannelTopologyBadge` renders inside `WorkflowNode` (canvas), not inside `NodeConfigPanel`. After opening the config panel with `nodes.nth(1).click()`, using `editor.getByTestId('channel-topology-badge')` will match the badge rendered on the canvas node card — not a panel-level control. This tests the wrong element's content.

*File: `packages/web/src/components/space/visual-editor/WorkflowNode.tsx` lines 41–51*

**6. Test 11: `channel-entry` and `remove-channel-button` are inside `channels-section` in `NodeConfigPanel`, not inside `ChannelTopologyBadge`**

Test 11 does `channelBadge.locator('[data-testid="channel-entry"]')` — looking for `channel-entry` children *inside* `channel-topology-badge`. But `channel-entry` elements are in the `channels-section` of `NodeConfigPanel`. The two UI elements are in completely different parts of the DOM. Furthermore, `channels-section` is only rendered when a step is in multi-agent mode (`isMultiAgentStep(step)`), which requires at least two agents assigned — not just one.

*File: `packages/web/src/components/space/visual-editor/NodeConfigPanel.tsx` line 523*
```tsx
{isMultiAgentStep(step) && (
  <ChannelsPanelSection step={step} agents={agents} onUpdate={onUpdate} />
)}
```

**7. Test 8: No deletion protection logic exists**

Test 8 selects the Task Agent node and presses Delete, then asserts the node survives. The current `WorkflowCanvas` keyboard handler fires `onDeleteNode(selectedNodeId)` for any selected node with no Task Agent guard. Without the feature implementation, the delete logic will simply not find a node with that ID (since no Task Agent node exists), so the test may accidentally pass for the wrong reason — making it a false green rather than a meaningful guard.

---

### P1 — Test methodology issues

**8. Test 9 and 11: `channel-topology-badge` assertion conflates canvas badge with panel badge**

Even if the Task Agent feature were implemented today, using `editor.getByTestId('channel-topology-badge')` after opening the config panel is ambiguous. The badge lives on the canvas node, while the test intends to verify config-panel content. The locator should be scoped to `editor.getByTestId('node-config-panel').getByTestId('channels-section')` (or similar) for the panel interaction, and separately to the canvas node locator for canvas-level channel visualization.

**9. Test 11 does not verify the channel is removed from the canvas edge layer**

After clicking the remove button, the test only checks that the text `task-agent` is gone from the badge. It does not verify the SVG edge between the regular node and the Task Agent node is removed from the canvas. This leaves the canvas-level effect of the removal untested.

**10. Tests 7–11 assume Task Agent is always `nodes.nth(0)`**

DOM ordering of Preact-rendered nodes is not contractually guaranteed to match insertion order in all scroll/reflow scenarios. A position-independent selector (e.g., the specific `data-testid`) would be more robust. Tests 11 specifically uses `nodes.nth(2)` for the "second regular node" with the assumption that the Task Agent node is always index 0.

---

### P2 — Minor issues

**11. `let nodes` re-declared in tests 9 and 11**

Both test bodies use `let nodes = editor.locator(...)` but never reassign `nodes` — `const` would be appropriate. This is cosmetically inconsistent with the rest of the file which uses `const nodes`.

**12. Test 9 looks for `text=task-agent` but the `ChannelTopologyBadge` truncates roles**

`ChannelTopologyBadge.roleLabel` truncates strings longer than 8 chars: `role.length > 8 ? role.slice(0, 8) + '…' : role`. The string `task-agent` is 10 characters and would be rendered as `task-age…`. The assertion `channelBadge.locator('text=task-agent')` would fail even if the feature existed exactly as described.

*File: `packages/web/src/components/space/visual-editor/WorkflowNode.tsx` line 33*

---

### Summary

The five new tests (7–11) are written against a feature that has not been implemented yet. All critical selectors — the pinned Task Agent node, its special data attributes, its CSS classes, and the assumed initial canvas state — are absent from the current codebase. Additionally, there are structural mismatches in how the tests look for channel-related UI (wrong DOM scope) and a truncation bug that would cause Test 9 to fail even when the feature ships. These tests should either be gated behind a feature flag, moved to a dedicated branch that includes the implementation, or held in a skip-annotated state with a tracking issue — the current form will cause CI failures on any merge to dev/main.

lsm (commented):

---

The worker correctly addressed the core blocker from the previous review by wrapping all five Task Agent tests in `test.skip()`. The specific fixes are confirmed:

- Node count assertions in tests 7–11 are corrected (1 for Task Agent only, 2 after adding one regular node, etc.)
- `channel-topology-badge` lookup is now scoped to the canvas node card (`nodes.nth(1)`), not the config panel
- `channel-entry` is looked up inside `channels-section` from `node-config-panel` (Test 11)
- The `text=task-age` partial match correctly accounts for the 8-char truncation in `ChannelTopologyBadge`
- All `let nodes` declarations are now `const nodes`

These are real improvements. However, two issues remain that justify asking for changes before merging.

---

### P1 — Test 11: `filter({ has: ... })` uses the wrong locator root

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 607–609

```ts
const taskAgentChannelEntry = channelsSection.locator('[data-testid="channel-entry"]').filter({
    has: editor.locator('text=task-age'),
});
```

Playwright's `filter({ has })` evaluates the `has` locator *relative to each candidate element being filtered*. Passing `editor.locator('text=task-age')` — rooted at the entire editor element — does not constrain the match to be inside the individual `channel-entry` being tested. The `has` locator should be a relative/unrooted locator, e.g., `page.locator('text=task-age')`, so Playwright can correctly evaluate it within each candidate element's subtree.

This is a subtle but real Playwright API misuse that will produce false positives: if *any* `channel-entry` exists and *any* part of the editor shows `task-age`, the filter will match. The correct form is:

```ts
const taskAgentChannelEntry = channelsSection.locator('[data-testid="channel-entry"]').filter({
    has: page.locator('text=task-age'),
});
```

---

### P1 — Existing tests 1–6 will break when the feature is implemented without any documented warning

The new tests' inline comments explain that the Task Agent pinned node is "always present at canvas open". That is correct. But the *existing* tests 1–6 contain these hardcoded `toHaveCount` assertions that will all be wrong the moment the feature ships:

| Test | File line | Current assertion | Will become wrong because |
|------|-----------|------------------|---------------------------|
| Test 1 | 111 | `toHaveCount(3)` after 3 Add-Step clicks | Task Agent adds 1, so count is 4 |
| Test 2 | 245 | `toHaveCount(2)` for a 2-step workflow | Task Agent adds 1, so count is 3 |
| Test 3 | 310 | `toHaveCount(2)` after Coding template | Task Agent adds 1, so count is 3 |
| Test 5 | 398 | `toHaveCount(1)` after 1 Add-Step click | Task Agent adds 1, so count is 2 |
| Test 6 | 426 | `toHaveCount(1)` after 1 Add-Step click | Task Agent adds 1, so count is 2 |

There is also a secondary effect in Test 3: the comment on line 316–319 asserts that the `template-picker-button` is hidden "after nodes exist". If the Task Agent node is always present on canvas open, the template-picker-button will be hidden *immediately* on editor open — making that visibility assertion unreachable.

None of these impacts are mentioned in the PR description or the inline `// Once the feature is implemented ... remove the test.skip()` comments. The developer enabling the feature will have to also update tests 1–6 with no warning that they need to, which is a reliability risk.

**Suggested fix:** Add a single comment block above Test 1 (or at the top of the `Task Agent Node Tests` section) that says:

> When removing test.skip() from tests 7–11, also update toHaveCount in tests 1–6: add +1 to every count (Task Agent is always present), and update Test 3's template-picker-button assertion (it will be hidden from canvas open, not just after template nodes are added).

---

### P3 — Misleading comment in Test 11

**File:** line 606

```ts
// ChannelTopologyBadge truncates "task-agent" to "task-age…"
```

This comment appears in Test 11 but the `channel-entry` in the `channels-section` of `NodeConfigPanel` is a different component from `ChannelTopologyBadge`. Labeling it `ChannelTopologyBadge` suggests the badge component and the panel's channel list share the same truncation logic, which may not be true. The comment should note it's the `channels-section`'s own rendering that truncates, or omit the component attribution.

---

### Summary

The P0 blockers from the previous review are resolved. The remaining issues are: a Playwright API misuse in the `filter({ has })` call in Test 11 that will produce non-deterministic matches (P1), and an undocumented trap for the developer who implements the feature — five existing tests will silently break without any warning in the codebase (P1). Both are straightforward to address with a one-line code fix and a comment addition.

lsm (commented):

---

Recommendation: **APPROVE** (with one P1 logic bug noted in a skipped test)

### Summary

The PR correctly wraps all 5 Task Agent tests in `test.skip()`, addresses the two previously-flagged P1 issues (the `filter({ has: })` locator root in Test 11, and the `+1 node count` comments in Tests 1–6), and the overall test structure is sound. One latent logic bug exists in Test 11 that will cause it to fail when the feature is implemented and the skip is removed.

---

### P1: Test 11 will fail when unskipped — `channels-section` won't render

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 605–638

Test 11 asserts that the `channels-section` is visible in the NodeConfigPanel for `nodes.nth(2)` (the second regular node). However, `channels-section` is only rendered when `isMultiAgentStep(step)` returns true (NodeConfigPanel.tsx line 523):

```tsx
{isMultiAgentStep(step) && (
    <ChannelsPanelSection step={step} agents={agents} onUpdate={onUpdate} />
)}
```

`isMultiAgentStep` returns true only when `step.agents` is a non-empty array (WorkflowStepCard.tsx line 41–43). In the test, `nodes.nth(2)` is freshly added and never given any agents, so it starts in single-agent mode (`step.agents` is undefined). The test comment (line 621) acknowledges this requirement — "channels-section only renders in multi-agent mode" — but the test never actually switches that node to multi-agent mode.

The test correctly sets up `nodes.nth(1)` with a single agent via `agent-select` (line 608), but that is single-agent mode too. Neither node is put into multi-agent mode (`agents[]` array with 2+ entries), so the panel that Test 11 expects to interact with will never appear.

**Fix needed when unskipping:** Before opening the config panel for `nodes.nth(2)`, configure it with at least two agents using `setupMultiAgentStep()` (already available in `workflow-editor-helpers.ts`). The comment should be updated to describe this prerequisite.

---

### P2: Test 9 comment misstates truncation mechanics

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 556

The comment reads:
```
// Note: ChannelTopologyBadge truncates "task-agent" to "task-age…" (10 chars → 8 + ellipsis)
```

The parenthetical "10 chars → 8 + ellipsis" is ambiguous — it reads as if "10" refers to the truncation limit, when it actually means the input length. The actual truncation rule in `WorkflowNode.tsx` line 33 is: `role.length > 8 ? role.slice(0, 8) + '…' : role`. A clearer comment would be: "'task-agent' is 10 chars, exceeds the 8-char limit, and truncates to 'task-age…'".

---

### P2: Test 8 only tests `Delete` key, not `Backspace`

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 522

Graph editors typically handle both `Delete` and `Backspace` for node deletion. If the Task Agent pinned-node implementation guards against both keys, only testing `Delete` gives partial coverage. Consider adding a `Backspace` assertion when unskipping, unless the implementation is confirmed to guard only `Delete`.

---

### P3: `getDefaultAgentId` helper is defined inline instead of in shared helpers

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 54–66

This helper uses `spaceAgent.list` RPC and is only used in Test 2. If other tests in this file or related files ever need agent IDs, the inline definition will be duplicated. This is a minor extensibility concern — moving it to `workflow-editor-helpers.ts` would be consistent with the other shared helpers already there.

---

### What was done well

- The `test.skip()` annotation is correctly applied to all 5 Task Agent tests and the block comment (lines 461–475) thoroughly documents the unimplemented requirements and the conditions to remove the skips.
- The `page.locator()` fix in Test 11's `filter({ has: })` at line 627–629 is correct — Playwright's `has:` option requires a page-rooted locator, not a frame-scoped one.
- The `+1 node count` NOTE comments on Tests 1–6 are specific and actionable, calling out exact line numbers to update.
- Test 9's distinction between `ChannelTopologyBadge` on the canvas node card versus `channels-section` in the config panel is clearly documented.
- Serial mode (`test.describe.configure({ mode: 'serial' })`) is appropriate given the single-space shared setup.

lsm (commented):

**Recommendation: APPROVE** — The implementation is structurally correct and all previous round P0/P1 issues have been resolved. One P1 issue remains (unused import), plus two P2 and two P3 concerns noted below.

---

### P1 — Unused import: `setupMultiAgentStep`

`packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 40:

```ts
import {
    ...
    setupMultiAgentStep,   // ← imported but never called
    getDefaultAgentId,
} from '../helpers/workflow-editor-helpers';
```

The last fix commit (e1fc10a2) intentionally inlined the multi-agent setup in Test 11 using index-based selection (to avoid needing label strings for dynamically-created agent names). That was the right call, but `setupMultiAgentStep` was left in the import list and is never invoked anywhere in the file. Since both oxlint and knip exclude `packages/e2e/**`, this won't fail CI, but it is genuine dead code that will confuse whoever enables these tests later. It should be removed from the import.

---

### P2 — Test 11 `add-agent-select { index: 2 }` assumes 3+ options exist

`packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 618:

```ts
await panel.getByTestId('add-agent-select').selectOption({ index: 2 });
```

The test space is created via `createSpace(page, ...)` in `beforeEach`, which produces a space with a single default agent. When `agent-select` already has `{ index: 1 }` chosen and the multi-agent `add-agent-select` dropdown appears, it likely shows the remaining unselected agents. If the test space only ever has one agent (the default planner), `add-agent-select` at `{ index: 2 }` might resolve to "no option" or the wrong entry. The test needs to verify that the test space has at least 3 selectable agents, or the multi-agent setup needs to create agents like `space-multi-agent-editor.e2e.ts` does. Given this test is currently skipped, it won't fail today, but it is a latent correctness issue to resolve before unskipping.

---

### P2 — Test 9 assertion comment mismatch

`packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 548–555:

```ts
// The ChannelTopologyBadge on the canvas node should show a channel to/from Task Agent
// "task-agent" is 10 chars; the 8-char truncation limit produces "task-age…"
// So we look for the truncated form. Also verify on the node card itself.
const nodeCard = nodes.nth(1);
await expect(nodeCard.locator('[data-testid="channel-topology-badge"]')).toBeVisible(...);
await expect(nodeCard.locator('text=task-age')).toBeVisible(...);
```

The comment says "Also verify on the node card itself" but both assertions are already on `nodeCard`. The `text=task-age` check and the badge check are assertions on the same element scope — the comment implies two separate verification targets but there's only one. Minor wording issue, but misleading for whoever reads this during implementation.

---

### P3 — Test 11 DOM ordering assumption for Task Agent node

The test assumes `nodes.nth(0)` will always be the Task Agent pinned node and regular nodes will be `nth(1)`, `nth(2)`. DOM order in canvas renderers (especially ReactFlow/similar) depends on the order elements are mounted, which may not be deterministic or stable. A comment noting this assumption (and advising to check ordering once the feature lands) would help the implementer avoid a subtle failure.

---

### P3 — Test 3 NOTE comment is partly speculative

`packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 280–284:

```
// NOTE: When the Task Agent pinned node feature is implemented:
// - The template-picker-button visibility assertion below will be unreachable because
//   the Task Agent node is always present, so the canvas is never truly "empty"
//   and the picker is hidden immediately on editor open.
```

Whether the template picker is hidden on open when a Task Agent node is present depends entirely on the implementation — the picker may be conditioned on "no *workflow* steps" rather than "no canvas nodes". This is plausible speculation but stated as a certainty. A softer "may be unreachable" wording would be more accurate.

---

### Positive Notes

- The `getDefaultAgentId` move to shared helpers is a clean refactor that removes duplication.
- Test 8 now covers both Delete and Backspace keys as requested.
- Test 11's inline multi-agent setup with index-based selection is the correct approach for a single-agent test space.
- The `page.locator` (not `editor.locator`) inside `filter({ has: })` in Test 11 is correct per Playwright semantics.
- All 6 existing tests are unaffected; the skipped tests won't run until the feature is implemented.
- NOTE comments on Tests 1–6 warning about future `toHaveCount` adjustments are valuable forward guidance.

lsm (commented):

---

### Summary

The dead import removal, DOM ordering note, and Test 9 comment clarification are all correct. However, a P0 regression was introduced in Test 11 by changing `add-agent-select` from `{ index: 2 }` to `{ index: 0 }`. That change makes the test unable to add a second agent, so the `toHaveCount(2)` assertion that follows will always fail once the feature is enabled.

---

### P0 — Test 11: `add-agent-select { index: 0 }` selects the blank placeholder, not an agent

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 620

The `add-agent-select` dropdown is rendered as:

```tsx
<option value="">+ Add agent…</option>        // index 0 — blank placeholder
{availableAgents.map((a) => (                   // index 1, 2, 3 — actual agents
    <option key={a.id} value={a.id}>...</option>
))}
```

The `addAgent` handler guards against an empty value (`if (!agentId) return;` — `NodeConfigPanel.tsx` line 74). So `selectOption({ index: 0 })` fires the `onChange` event with an empty string, `addAgent` returns immediately without adding any agent, and the immediately subsequent assertion:

```ts
await expect(panel.getByTestId('agents-list').getByTestId('agent-entry')).toHaveCount(2, {
    timeout: 3000,
});
```

...will time out waiting for a second entry that never appears.

**Fix:** Change `{ index: 0 }` to `{ index: 1 }` to select the first real agent. The commit message note "test space has 2 agents" is also incorrect — `seedPresetAgents` seeds 4 agents (Coder, General, Planner, Reviewer), and `{ index: 1 }` is sufficient and unambiguous across all 4.

---

### P1 — Misleading comment: "test space has 2 agents"

**File:** commit message and `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 614 comment

The inline comment says "after selecting index 1, the remaining agent is at index 0 in the dropdown". That reasoning is only valid if there are exactly 2 agents total (1 selected → 1 remaining → index 0 of remaining). With 4 seeded agents, after selecting index 1 from the main dropdown there are 3 remaining agents in `add-agent-select`, at indices 1, 2, 3 (index 0 is the blank placeholder). The comment misstates the space agent count and leads to the P0 above.

---

### P2 — Test 11 truncation comment is incorrect for `channels-section`

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, line 631

```ts
// "task-agent" is 10 chars; the 8-char truncation limit produces "task-age…"
const taskAgentChannelEntry = channelsSection.locator('[data-testid="channel-entry"]').filter({
    has: page.locator('text=task-age'),
});
```

This comment was accurate for the `ChannelTopologyBadge` in `WorkflowNode.tsx` (which does truncate via `roleLabel`), but `NodeConfigPanel.tsx` renders channel entries without any truncation (`formatTo` at line 297 just joins arrays, no `roleLabel` call). The reason `text=task-age` works here is that Playwright's `text=` selector does a case-insensitive substring match, so it matches the full "task-agent" string. The comment should say: "using substring match — matches the full 'task-agent' text in the config panel".

---

### Everything else looks correct

- Dead `setupMultiAgentStep` import is cleanly removed.
- The `page.locator` inside `filter({ has: })` is correctly page-rooted (not `editor.locator`).
- Tests 7–10 logic is sound given their `test.skip` status.
- The NOTE about DOM ordering assumption in Test 11 is a useful heads-up.
- The `getDefaultAgentId` helper extraction in `workflow-editor-helpers.ts` is clean.
- Existing tests 1–6 pass; forward-compatibility NOTE comments on `toHaveCount` are appropriate.

lsm (commented):

**Recommendation: APPROVE**

The implementation is correct and well-structured. All five new tests are properly annotated with `test.skip()` and contain accurate documentation of the feature expectations. The refactoring of `getDefaultAgentId` into the shared helper file is a clean improvement. The notes on existing tests 1-6 warning about future count changes are good defensive documentation.

One minor correctness issue and two P3 nits below.

---

### P2 — Misleading filter substring in Test 11 (non-blocking given tests are skipped)

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 631-635

The comment on line 631 says *"NodeConfigPanel does NOT truncate (unlike ChannelTopologyBadge)"*, which is confirmed correct by the `NodeConfigPanel.tsx` implementation (`formatTo` joins without any length limit). However, the filter still uses `text=task-age` instead of `text=task-agent`.

The comment tries to justify this by saying *"Playwright's text= does a case-insensitive substring match, so 'task-age' matches 'task-agent'"* — but that logic is circular. If the panel doesn't truncate, then the full string `task-agent` is present in the DOM and searching for `task-age` is needlessly imprecise. If the channel identifier happened to be named `task-ages` or similar, this would be a false positive. Since the panel is confirmed not to truncate, this should use `text=task-agent` for clarity and precision. This is not a blocking bug (the test is skipped), but it should be corrected before unskipping.

---

### P3 — Test 9: The `ChannelTopologyBadge` truncation assertion may need verification

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 547-554

The comment says *"task-agent is 10 chars; the 8-char truncation limit produces 'task-age…'"*. Looking at the implementation in `WorkflowNode.tsx` line 33:

```ts
const roleLabel = (role: string) => (role.length > 8 ? role.slice(0, 8) + '…' : role);
```

`'task-agent'` has 10 chars, which is `> 8`, so `slice(0, 8)` gives `'task-age'`. The math is correct. The test assertion `text=task-age` is accurate here.

---

### P3 — Node ordering assumption in Test 11 is noted but fragile

**File:** `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, lines 607-610

The comment on line 609 acknowledges the fragility: *"DOM ordering assumes Task Agent pinned node is always nth(0)"*. This is a valid concern. When unskipping, it would be better to use the explicit `data-testid="workflow-node-task-agent"` locator to identify the Task Agent node rather than relying on DOM order. The regular nodes could then be identified as nodes that are NOT the Task Agent node. This is out of scope for the current PR since tests are skipped, just noting for when they get unskipped.

---

### Correctness verification

- `seedPresetAgents` creates exactly 4 agents (Coder, General, Planner, Reviewer) — confirmed in `/packages/daemon/src/lib/space/agents/seed-agents.ts`. Comment on line 614 is correct.
- `add-agent-select` index 0 is the `+ Add agent…` placeholder with empty string value (line 222 of `NodeConfigPanel.tsx`), index 1 is the first real agent. The fix from index 0 → 1 is correct.
- The `channels-section` testid exists in `NodeConfigPanel.tsx` at line 300, `channel-entry` at line 314, `remove-channel-button` at line 322 — all selectors are valid.
- `ChannelTopologyBadge` truncation logic at 8 chars verified against implementation.
- Test infrastructure (`createSpace`/`deleteSpace` in before/afterEach) correctly follows the allowed RPC-in-infrastructure-only pattern.
- All 6 existing tests are left unmodified and pass.
- The `getDefaultAgentId` migration to `workflow-editor-helpers.ts` is correct and the local definition was properly removed.

lsm (commented):

**Recommendation: APPROVE**

The PR is ready to merge.

---

### Summary

This PR adds 5 E2E tests (Tests 7-11) for the Task Agent visible node feature to `packages/e2e/tests/features/visual-workflow-editor.e2e.ts`, all correctly wrapped in `test.skip()` since the underlying feature has not been implemented yet. It also moves the `getDefaultAgentId` function from the test file into the shared `workflow-editor-helpers.ts` helper module.

### What was verified

**Correctness of the final fix (Test 11 filter)**

The commit message states the fix was changing `text=task-age` to `text=task-agent` in Test 11's `NodeConfigPanel` channel filter. Line 634 now reads:

```ts
has: page.locator('text=task-agent'),
```

The comment on lines 631–632 correctly explains the rationale: NodeConfigPanel does not truncate role names, so the full string `task-agent` is appropriate here, unlike the `ChannelTopologyBadge` in Test 9 (line 554) which uses the truncated `text=task-age` because the badge has an 8-char display limit. The comment and code are consistent.

**Test 9 truncation assumption**

Test 9 still uses `text=task-age` at line 554 with the comment that `task-agent` (10 chars) is truncated to `task-age…` under the 8-char limit. This is a documented assumption about the `ChannelTopologyBadge` implementation. The comment is present and clear; the assumption will need verification when the feature is implemented.

**Helper refactor**

`getDefaultAgentId` was duplicated between the test file and the helpers module. The PR removes the local copy in the test file and imports from `workflow-editor-helpers.ts`, where it now also serves `space-multi-agent-editor.e2e.ts`. No logic change was made to the function itself.

**Existing tests**

The 6 existing tests are unchanged beyond forward-compatibility comments (e.g., `// NOTE: When the Task Agent pinned node feature is implemented, the toHaveCount ... will need to change from N to N+1`). These notes are accurate and useful for whoever enables the skipped tests later.

**E2E rule compliance**

All skipped test bodies follow the project's E2E rules: actions go through the UI, assertions check visible DOM state, and no direct `hub.request()` calls or internal state accesses appear in the test bodies.

**No issues found.** The PR is clean, well-commented, and correctly scoped.

PR #760, opened 2026-03-23
feat: auto-create Task Agent channels on node addition

lsm:
- Remove M3 runtime auto-generation of Task Agent channels from
  resolveAndStoreChannels() - channels are now only created via
  user-declared workflow data
- Add auto-creation of Task Agent channels when an agent is assigned
  to a node in NodeConfigPanel (useEffect in AgentsSection)
- Show ChannelsPanelSection for all nodes with agents (not just
  multi-agent nodes) so Task Agent channels are visible and removable
- Add ChannelEdgeRenderer component to WorkflowCanvas for rendering
  Task Agent channel edges with distinct dashed gray visual style
- Add channelEdges derivation in VisualWorkflowEditor from nodes'
  step.channels
- Update unit tests to verify no auto-generated channels exist

Task: #43

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -1000,15 +1000,9 @@ export class SpaceRuntime {
 	 * TODO Milestone 6: pass resolvedChannels to session group metadata in
 	 * TaskAgentManager.spawnTaskAgent() instead of storing in run config.
 	 *
-	 * Default task-agent channels:
-	 * When a step has node agents, bidirectional channels are auto-created between
-	 * 'task-agent' and each node agent role. These defaults are added in addition
-	 * to any user-declared channels, ensuring the Task Agent can always reach peers.
-	 * Duplicates are avoided: if the user explicitly declares a task-agent channel,
-	 * it is not re-added.
-	 *
-	 * Steps with no `channels` declaration still get default task-agent channels,
-	 * so the Task Agent always has peer communication ability.
+	 * Note: Task Agent channels are persisted by the frontend as WorkflowChannel
+	 * entries in the workflow data. This function only resolves and stores
+	 * user-declared channels — no runtime auto-generation.
 	 */
 	resolveAndStoreChannels(runId: string, spaceId: string, step: WorkflowStep): void {
 		const run = this.config.workflowRunRepo.getRun(runId);
@@ -1018,65 +1012,8 @@ export class SpaceRuntime {
 
 		const allAgents = this.config.spaceAgentManager.listBySpaceId(spaceId);
 
-		// Resolve user-declared channels (empty array if none declared)
-		const userResolved = resolveStepChannels(step, allAgents);
-
-		// Auto-generate default bidirectional channels between task-agent and each node agent role.
-		// These are added regardless of whether user-declared channels exist, ensuring the
-		// Task Agent can always communicate with step agents.
-		const stepAgents = resolveStepAgents(step);
-		// Deduplicate roles — a step may have multiple agents with the same role,
-		// but we only need one bidirectional channel pair per unique role.
-		const nodeRoles = [
-			...new Set(
-				stepAgents
-					.map((sa) => {
-						const spaceAgent = allAgents.find((a) => a.id === sa.agentId);
-						return spaceAgent?.role ?? null;
-					})
-					.filter((role): role is string => role !== null)
-			),
-		];
-
-		// Build a set of existing channel pairs (fromRole→toRole) to avoid duplicates
-		const existingPairs = new Set<string>();
-		for (const ch of userResolved) {
-			existingPairs.add(`${ch.fromRole}→${ch.toRole}`);
-		}
-
-		// Generate default task-agent ↔ node bidirectional channels.
-		// agentId fields use the role string as a placeholder — ChannelResolver.canSend()
-		// only checks fromRole/toRole, so the actual agentId value does not affect routing.
-		// If agentId enforcement is added in the future, this placeholder should be replaced
-		// with the actual agent's ID (and this comment updated).
-		const defaultChannels: typeof userResolved = [];
-
-		for (const nodeRole of nodeRoles) {
-			// task-agent → node (if not already declared)
-			if (!existingPairs.has(`task-agent→${nodeRole}`)) {
-				defaultChannels.push({
-					fromRole: 'task-agent',
-					toRole: nodeRole,
-					fromAgentId: 'task-agent',
-					toAgentId: nodeRole,
-					direction: 'one-way',
-					isHubSpoke: false,
-				});
-			}
-			// node → task-agent (if not already declared)
-			if (!existingPairs.has(`${nodeRole}→task-agent`)) {
-				defaultChannels.push({
-					fromRole: nodeRole,
-					toRole: 'task-agent',
-					fromAgentId: nodeRole,
-					toAgentId: 'task-agent',
-					direction: 'one-way',
-					isHubSpoke: false,
-				});
-			}
-		}
-
-		const resolved = [...userResolved, ...defaultChannels];
+		// Resolve user-declared channels from workflow data (empty array if none declared)
+		const resolved = resolveStepChannels(step, allAgents);
 
 		this.config.workflowRunRepo.updateRun(runId, {
 			config: { ...config, _resolvedChannels: resolved },
--- a/packages/daemon/tests/unit/space/space-runtime.test.ts
+++ b/packages/daemon/tests/unit/space/space-runtime.test.ts
@@ -2095,36 +2095,28 @@ describe('SpaceRuntime', () => {
 			expect((resolvedChannels as unknown[]).length).toBeGreaterThan(0);
 		});
 
-		test('storeResolvedChannels: step without channels stores default task-agent channels', async () => {
+		test('storeResolvedChannels: step without channels does NOT store task-agent channels', async () => {
+			// When a step has no user-declared channels, no channels should be stored.
+			// M3 auto-generation of task-agent channels has been removed.
 			const workflow = buildLinearWorkflow(SPACE_ID, workflowManager, [
 				{ id: STEP_A, name: 'No Channels', agentId: AGENT_CODER },
 			]);
 
 			const { run } = await runtime.startWorkflowRun(SPACE_ID, workflow.id, 'Run');
 
-			// Run config should have _resolvedChannels set to default task-agent bidirectional channels
-			// (auto-added even when no user-declared channels exist). This ensures the Task Agent
-			// can always reach node agents.
+			// Run config should have _resolvedChannels set to empty (no auto-add)
 			const updatedRun = workflowRunRepo.getRun(run.id)!;
 			const resolvedChannels = (updatedRun.config as Record<string, unknown> | undefined)
-				?._resolvedChannels as Array<Record<string, unknown>>;
+				?._resolvedChannels as Array<Record<string, unknown>> | undefined;
+			// Should be empty array when no user-declared channels exist
 			expect(Array.isArray(resolvedChannels)).toBe(true);
-			expect(resolvedChannels.length).toBeGreaterThan(0);
-			// Default task-agent ↔ coder channels should be present
-			const taskAgentToCoder = resolvedChannels.find(
-				(ch) => ch.fromRole === 'task-agent' && ch.toRole === 'coder'
-			);
-			expect(taskAgentToCoder).toBeDefined();
-			const coderToTaskAgent = resolvedChannels.find(
-				(ch) => ch.fromRole === 'coder' && ch.toRole === 'task-agent'
-			);
-			expect(coderToTaskAgent).toBeDefined();
+			expect(resolvedChannels.length).toBe(0);
 		});
 
-		test('storeResolvedChannels: deduplicates channels when step has multiple agents with the same role', async () => {
-			// When a step has two agents that share the same role (e.g., two coder agents),
-			// resolveAndStoreChannels should generate only ONE bidirectional task-agent↔coder
-			// channel pair, not two duplicate pairs.
+		test('storeResolvedChannels: no auto-generated channels when step has multiple agents with the same role', async () => {
+			// When a step has no user-declared channels, no channels should be stored,
+			// even if the step has multiple agents with the same role.
+			// M3 auto-generation has been removed.
 			const AGENT_CODER_2 = 'agent-coder-2-duplicate-role';
 			seedAgentRow(db, AGENT_CODER_2, SPACE_ID, 'Coder 2', 'coder');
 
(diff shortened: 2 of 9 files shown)


lsm (commented):

---

## Summary

The PR correctly removes M3 runtime auto-generation from `resolveAndStoreChannels` and updates the backend tests accordingly. The visual channel edge rendering is well-structured with distinct styling. However, there are **6 failing frontend tests** (confirmed by running `make test-web`) caused by the `useEffect` auto-creation logic firing unexpectedly, plus additional correctness and design issues.

---

## P0 — Blocking Bugs

### 1. Six existing tests break due to useEffect firing on mount

The `useEffect` in `AgentsSection` (NodeConfigPanel.tsx:71–96) has an empty deps array `[]` so it fires once on component mount with whatever `step` prop is present at that moment. This causes **6 tests to fail** in `src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx`:

- `does not show channels section in single-agent mode` — The step fixture has `agentId: 'agent-1'` and no channels. On mount the `useEffect` fires and calls `onUpdate` to set channels. The render condition `(!!step.agentId || ...)` then shows the channels section, but the test expects `null`.
- `clicks Add agent button switches to multi-agent mode` — `onUpdate` is expected to be called once (for the button click), but the `useEffect` fires a second call on mount. The test asserts `toHaveBeenCalledOnce()` and fails.
- `remove agent button calls onUpdate without that agent` — Same problem: `onUpdate.mock.calls[0][0]` is the `useEffect`-triggered call (channels), not the remove-agent call, so `updatedStep.agents.length` is 2 instead of 1.
- `calls onUpdateEntryCondition/onUpdateExitCondition` — Select index-based selection (e.g., `selects[1]`) is off by one because the channels section now adds extra `<select>` elements for the new `ChannelsPanelSection` that is now visible for single-agent nodes.
- `add channel adds a new channel entry` — The `useEffect` auto-adds one channel (task-agent → planner) before the test manually adds another. The test expects 1 channel but gets 2.

**Evidence:** Running `make test-web` shows `6 failed` tests in `NodeConfigPanel.test.tsx`.

### 2. useEffect fires on open for all existing nodes, not just newly-assigned ones

The intent is to auto-create channels when "an agent is assigned". But the `useEffect` with `[]` deps fires every time `AgentsSection` mounts — including when the user opens an existing saved node that already has an agent but no channels. This silently mutates persisted workflow data on open, which is surprising and may corrupt previously intentional no-channel configurations.

---

## P1 — Should Fix

### 3. Duplicate channel edges for multi-agent nodes

In `VisualWorkflowEditor.tsx` (lines 175–201), `channelEdges` is derived by iterating all channels per node. For a multi-agent node auto-created with N agents (e.g., 2 agents each with a different role), the auto-creation in `AgentsSection` produces N channels all with `from: 'task-agent'`. Each channel matches `channel.from === 'task-agent'` and pushes a separate edge with the same `toStepId`. This yields N edges for the same node.

`ChannelEdgeRenderer` then renders N `<g key={"channel-${edge.toStepId}"}\>` elements with identical keys — a React key collision warning and visual overlap of identical overlapping paths.

A simple fix would be to deduplicate `channelEdges` by `toStepId` after building the result, or only add one edge per node.

### 4. useEffect reads stale closure values for multi-agent case

The `useEffect` at NodeConfigPanel.tsx:71–96 depends on `step`, `agents`, and `onUpdate`, but specifies `[]` deps with an `eslint-disable-next-line react-hooks/exhaustive-deps` comment. While the intent to run only once is correct, this means that in the multi-agent path (line 86), `step.agents` is read from the initial render's closure. If `step.agents` were populated asynchronously (e.g., from a signal) and was empty on mount, the `useEffect` would fall back to the `hasSingleAgent` path or no-op entirely. This is fragile.

The proper fix is either:
- Use a `useRef` sentinel to fire only once-ever per component mount (not per render).
- Or rethink the trigger: call channel creation in the event handler where the agent is actually assigned (in the `onChange` handler of the agent dropdown).

### 5. ChannelEdge hitbox is clickable but has no handler

In `ChannelEdgeRenderer` (WorkflowCanvas.tsx:191–196), the invisible hitbox path has `style={{ pointerEvents: 'stroke', cursor: 'pointer' }}`. This promises the user they can click the channel edge, but there is no `onClick` handler attached. Clicking it does nothing, which is confusing.

---

## P2 — Important Suggestions

### 6. Task Agent hub renders at fixed x=60 regardless of viewport or node positions

`TASK_AGENT_X = 60` is hardcoded in `ChannelEdgeRenderer`. If nodes are positioned near x=0 (e.g., auto-layout places the first node at x=50), the task-agent origin point at x=60 may overlap nodes. The bezier path computation uses `Math.abs(tx - sx)` which could be very small or even negative if a node is to the left of x=60, making the control points degenerate.

No Task Agent label or visual node is rendered on the canvas at x=60, so the source circle at `(60, sy)` floats in empty space. Users have no visual affordance explaining what this circle represents.

### 7. Channel auto-creation uses an empty channels array guard that can break the cycle

`ChannelsPanelSection.updateChannels` (line 297) sets channels to `undefined` when the array is empty: `onUpdate({ ...step, channels: next.length > 0 ? next : undefined })`. If a user removes all auto-created channels, `step.channels` goes back to `undefined`. But since the `useEffect` dep is `[]`, it won't re-run and re-create channels. This is actually the correct behavior per the design, but it means the Channel section disappears (condition on line 552 no longer satisfies `step.channels` being truthy), which seems inconsistent — the user deleted the last channel but the step still has an agent.

---

## P3 — Nits

### 8. The comment on line 552 says "shown when node has agents or has existing channels" but the condition also includes `isMultiAgentStep(step)` which is redundant with `step.agentId`

The condition `(!!step.agentId || isMultiAgentStep(step) || step.channels)` — `isMultiAgentStep` checks `step.agents?.length > 0`, and for multi-agent steps `step.agentId` is typically empty string. So `!!step.agentId` won't be true for multi-agent nodes. The two conditions are not redundant; the comment is slightly misleading. Minor.

### 9. Test name inconsistency

The test at line 469 is titled "does not show channels section in single-agent mode" but since the PR's intent IS to show channels for single-agent nodes (after the useEffect fires), this test's name and assertion are both stale. It should have been updated to document the new expected behavior.

---

## Additional Observations

- The backend removal of auto-generation from `resolveAndStoreChannels` is correct and cleanly implemented. The logic reduction is appropriate and the updated tests accurately reflect the new behavior.
- The `ChannelEdge` type export and JSDoc in WorkflowCanvas.tsx are clear and appropriate.
- The `channelEdges` derivation correctly skips one-way channels where task-agent is the target, which is good.
- The `eslint-disable-next-line react-hooks/exhaustive-deps` on line 95 is a code smell when used to silence a warning caused by a genuinely missing dependency, not an intentional optimization pattern.

lsm (commented):

The fixes address the original P0/P1 issues but introduce a new P0 bug in the multi-agent path. See details below.

---

## P0 — Double `onUpdate` call in `addAgent()` overwrites channels

**File:** `packages/web/src/components/space/visual-editor/NodeConfigPanel.tsx`, lines 88–93

```ts
function addAgent(agentId: string) {
  const next = [...stepAgents, { agentId }];
  autoCreateChannelsForAgents(next); // calls onUpdate({ ...step, channels: newChannels })
  updateAgents(next);                // calls onUpdate({ ...step, agents: next, agentId: '' })
}
```

`autoCreateChannelsForAgents` calls `onUpdate({ ...step, channels: newChannels })` — spreading the current `step` (which has `channels: undefined`) with only the `channels` field set (but `agents` not updated). Then `updateAgents(next)` calls `onUpdate({ ...step, agents: next, agentId: '' })` — spreading the same stale `step` which still has `channels: undefined`, overwriting the first call entirely.

The net result: when adding an agent in multi-agent mode, **channels are never persisted**. The second call wins, and the channels from the first call are lost. The auto-creation for the multi-agent path is effectively a no-op.

The fix is to merge both updates into a single `onUpdate` call, e.g.:
```ts
function addAgent(agentId: string) {
  if (!agentId) return;
  if (stepAgents.some((a) => a.agentId === agentId)) return;
  const next = [...stepAgents, { agentId }];
  const channels = step.channels === undefined
    ? next.map((sa) => {
        const role = agents.find((a) => a.id === sa.agentId)?.role ?? sa.agentId;
        return { from: 'task-agent', to: role, direction: 'bidirectional' };
      })
    : step.channels;
  onUpdate({ ...step, agents: next, agentId: '', channels });
}
```

There are no tests covering this path, so the regression is invisible to CI.

---

## P1 — No test for channel auto-creation in the multi-agent `add-agent-select` path

**File:** `packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx`

The existing test at line 194 ("calls onUpdate with new agentId when agent selection changes") uses the default `makeProps()` fixture (single-agent step, `agentId: 'agent-1'`) but doesn't assert that `onUpdate` was called with `channels` populated. The auto-channel creation for single-agent is tested only by checking visibility of the channels section after-the-fact.

No test fires a change on `add-agent-select` and asserts the resulting `channels` array. This gap let the P0 double-call bug go undetected.

Suggested test:
```ts
it('auto-creates task-agent channel when adding an agent in multi-agent mode', () => {
  const onUpdate = vi.fn();
  const step = makeStep({ agentId: '', agents: [{ agentId: 'agent-1' }] });
  const { getByTestId } = render(<NodeConfigPanel {...makeProps({ step, onUpdate })} />);
  fireEvent.change(getByTestId('add-agent-select'), { target: { value: 'agent-2' } });
  const updatedStep = onUpdate.mock.calls.at(-1)[0];
  expect(updatedStep.channels).toHaveLength(2); // one per agent
  expect(updatedStep.agents).toHaveLength(2);
});
```

---

## P2 — One-way task-agent channels are silently invisible in the visual editor

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx`, lines 185–187

```ts
const hasTaskAgent =
  (channel.from === 'task-agent' || channel.to === 'task-agent') &&
  channel.direction === 'bidirectional';  // <-- one-way excluded
```

Auto-created channels are always `bidirectional`, so this is consistent for the auto-creation path. However, a user who manually adds a `one-way` task-agent channel via the channels form will see no visual edge — the channel exists in the data but is invisible in the canvas. This is surprising behavior. Consider either removing the `direction === 'bidirectional'` constraint, or adding a comment explaining this is intentional.

---

## P2 — `seen.has(node.step.localId)` check on line 182 is dead code

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx`, lines 176–192

The `seen` Set is populated at line 190 after a matching channel is found and the loop `break`s. Since each node in `nodes` has a unique `localId`, and the outer loop iterates over `nodes` (not channels), `seen.has(node.step.localId)` on line 182 can never be true for a different node. The deduplication it was intended to prevent (stated in the PR description as "deduplication by toStepId") is actually accomplished by the `break` on line 191. The `seen` Set is unnecessary and should be removed to avoid confusing future readers.

---

## P3 — No test for the `switch-to-single-button` path

**File:** `packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx`

The "Switch to single" button (rendered when `stepAgents.length === 1`) has no dedicated test. The `removing last agent switches back to single-agent mode` test exercises `removeAgent` instead. The `switch-to-single-button` itself (which sets `agents: undefined`, restores `agentId`, and clears `channels`) is not tested.

---

## What is correct

- Removing `useEffect([], [])` and moving channel creation into event handlers is the right fix for the mount-time corruption issue.
- The single-agent `onChange` handler (lines 140–151) correctly merges `channels` and `agentId` in a single `onUpdate` call.
- The `removeAgent` path (lines 96–105) correctly clears `channels` when reverting to single-agent mode.
- Backend removal of M3 auto-generation in `resolveAndStoreChannels()` is clean and correct.
- Backend tests confirm no auto-generated channels exist when user has not declared any.
- `testId` prop on `GateConfig` and updated tests for `entry-gate-select`/`exit-gate-select` are correct improvements.
- The deduplication `break` in `channelEdges` (one edge per node) is functionally correct.
- `ChannelEdgeRenderer`'s hitbox (transparent `strokeWidth={12}` path) without a click handler is fine now that `pointerEvents: 'none'` is set on the visible path.

lsm (commented):

**Recommendation: APPROVE** (with one P1 worth fixing in a follow-up)

---

### Summary

The three fixes in the third commit (merged `onUpdate`, test, dead-code removal) are all correct and close the gaps from the previous round. The implementation is sound overall — the auto-creation logic is clear, the visual rendering is properly isolated, and the daemon-side removal of M3 auto-generation is well-tested. One P1 issue remains around a missing edge-case in the new test.

---

### Merged `onUpdate` in `addAgent` — Correct

`packages/web/src/components/space/visual-editor/NodeConfigPanel.tsx` lines 89-92

The fix is right. Previously two sequential `onUpdate` calls used the same stale `step` reference — the second call (channels) overwrote the first (agents) because both spread from the pre-mutation `step`. The merged call `onUpdate({ ...step, agents: next, agentId: '', channels: newChannels })` is atomic and correct.

---

### `channels === undefined` condition — Correct, but consider documenting the edge case

`NodeConfigPanel.tsx` line 91: `step.channels === undefined ? buildTaskAgentChannels(next) : step.channels`

This correctly distinguishes two states:
- `undefined` — channels were never set, auto-create them
- non-undefined (including an empty array `[]`) — user has intentionally modified channels, preserve them

One subtle edge case: `ChannelsPanelSection.updateChannels` sets channels to `undefined` when the last channel is removed (line 296: `next.length > 0 ? next : undefined`). This means removing all channels and then adding another agent will re-trigger auto-creation. That is arguably the correct behavior (the user cleared everything, so rebuilding is sensible), but it is not documented anywhere and a new contributor would not expect it. No code change needed, but a brief comment at line 91 explaining what `undefined` vs `[]` means in this context would prevent future confusion.

---

### P1: New test does not cover the "channels already set" guard

`packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx` line 457

The new test verifies: step with 1 agent + no channels → add second agent → channels auto-created. That covers the happy path.

What it does **not** test: step with 1 agent + **existing** channels → add second agent → existing channels **preserved** (not replaced). This is the other branch of the `channels === undefined` ternary (line 91). Without this test it is easy to accidentally invert the condition or change the ternary in a refactor and not notice. The behavior is load-bearing: a user who manually configures custom channels and then adds another agent would silently lose their custom channels if this guard regresses.

Suggested test name: `'preserves existing channels when adding an agent if channels are already set'`

---

### Dead `seen` Set removal — Correct

`VisualWorkflowEditor.tsx` `channelEdges` derivation: the `break` inside the inner loop already guarantees at most one edge per node, so the `seen` Set was genuinely dead. Removal is clean.

---

### Daemon-side tests — Good

The updated tests in `space-runtime.test.ts` now assert that when a step has no user-declared channels, `_resolvedChannels` is an empty array (not populated by auto-generation). The three new test cases (`step without channels`, `duplicate role`, `channel-step to no-channel step`) thoroughly cover the M3 removal.

---

### Minor: Single-agent path has an asymmetric condition

`NodeConfigPanel.tsx` line 143: `if (newAgentId && step.channels === undefined)` — this only creates the channel if the agent info lookup succeeds (line 145 `if (agentInfo)`). In `addAgent`, channels are built even when `agentInfo` is not found, falling back to `sa.agentId` as the role string. The behaviors are not parallel. This is a pre-existing inconsistency rather than a regression introduced by this PR, but worth a follow-up cleanup.

---

### Verdict

P0 from the previous review is fixed. The implementation is correct. The test covers the main scenario. The one P1 (missing "channels preserved" test) should be added in a follow-up — it is not a blocker for this PR given the logic is demonstrably correct from reading the code, but the gap is real enough to call out.

lsm (commented):

**Recommendation: APPROVE** — the implementation is correct and complete. The final P1 test is meaningful and properly guards the preservation contract.

---

### Final P1 Test: Channel Preservation on Agent Add

The new test at `packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx` (line 477–491) correctly:

- Sets up a multi-agent step with **pre-existing user-configured channels** (`step.channels` is not `undefined`)
- Fires the `add-agent-select` change event to add a second agent
- Asserts that `updatedStep.agents` has 2 agents (agent was added)
- Asserts that `updatedStep.channels` still has length 1 with `from: 'coder'` (existing channels preserved, not overwritten)

This directly guards the `step.channels === undefined` guard in `addAgent()`. If that guard were accidentally inverted or removed, this test would fail. The test is meaningful and the assertion covers the right behavior.

---

### Overall Implementation Quality

**Backend (daemon)**

The removal of M3 runtime auto-generation from `resolveAndStoreChannels()` is clean. The method now only resolves user-declared channels via `resolveStepChannels()`, and the updated daemon tests (space-runtime.test.ts ~line 2098–2147) correctly verify that empty channels are stored when no user-declared channels exist, and that multi-agent-same-role steps also produce no auto-generated channels.

**Frontend (NodeConfigPanel)**

The channel auto-creation logic is sound:
- Single-agent path: channel created in the `onChange` handler only when `step.channels === undefined` and `agentInfo` is found
- Multi-agent path: `addAgent()` uses a single merged `onUpdate` call (the previous P0 double-update stale-ref bug is fixed)
- The preservation guard (`step.channels === undefined ? buildTaskAgentChannels(next) : step.channels`) is correct

**ChannelEdgeRenderer / VisualWorkflowEditor**

- Deduplication by `break`-after-first-match is correct
- The filter requires `direction === 'bidirectional'` — one-way task-agent channels (if a user manually creates one) would not render a canvas edge. This is a minor UX gap but not a bug, and auto-created channels are always bidirectional so the feature path is covered.

**Minor observation (not blocking)**

In `removeAgent()` when `next.length > 0` (removing an agent but not the last one), the code calls `updateAgents(next)` which does NOT update `step.channels`. This means auto-created task-agent channels for the removed agent remain in `step.channels` after removal. The channels become stale (they reference a role no longer in the step). This is a pre-existing gap, not introduced by this PR, and is separately addressable — mentioning for awareness only.

---

### Test Coverage

- 53 NodeConfigPanel tests pass including the new preservation guard test
- Daemon channel tests verify the no-auto-generation behavior on both single-agent and multi-agent steps
- The auto-creation path via `add-agent-select` is also covered by the test at line 457

All critical behaviors are exercised. The PR is ready to merge.

PR #761, opened 2026-03-23
fix: prevent TaskView error on tab resume due to race condition

lsm:
When switching browser tabs, the TaskView could show "not connected to
server" error due to a race condition between visibilitychange handler
and useTaskViewData useEffect:

1. Tab returns to foreground -> validateConnectionOnResume() starts
2. During reconnection, connectionState bounces: connected -> disconnected
3. useEffect runs on isConnected change, calling load()
4. If reconnection hasn't completed, request() throws ConnectionNotReadyError
5. Error is set permanently (no retry)

Fix by combining two approaches:

- connection-manager.ts: Don't report 'connected' state until
  validateConnectionOnResume() fully completes. Added _isResuming flag
  to suppress premature 'connected' state updates during validation.

- useTaskViewData.ts: Clear stale errors at start of load() and add
  connection guard. If not connected, load() returns early without
  setting an error, allowing the effect to re-trigger when connection
  is properly restored.

Also adds unit tests covering tab-resume scenarios.

Test plan:
- All 5359 web tests pass
- TypeScript build passes

### DIFF
--- a/packages/web/src/hooks/__tests__/useTaskViewData.test.ts
+++ b/packages/web/src/hooks/__tests__/useTaskViewData.test.ts
@@ -19,13 +19,21 @@ const mockOnEvent = vi.fn((_eventName: string, _handler: unknown) => () => {});
 const mockJoinRoom = vi.fn();
 const mockLeaveRoom = vi.fn();
 
+// Module-level state for dynamic mocking (used by tab-resume tests)
+const mockMessageHubState = {
+	isConnected: true,
+	requestThrows: false,
+};
+
 vi.mock('../useMessageHub.ts', () => ({
 	useMessageHub: () => ({
 		request: mockRequest,
 		onEvent: mockOnEvent,
 		joinRoom: mockJoinRoom,
 		leaveRoom: mockLeaveRoom,
-		isConnected: true,
+		get isConnected() {
+			return mockMessageHubState.isConnected;
+		},
 	}),
 }));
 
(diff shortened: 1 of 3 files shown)


lsm (commented):
## P1: `_isResuming` flag suppresses `connected` state after full reconnect

In `validateConnectionOnResume()`, when `!this.messageHub || !this.transport`, `reconnect()` is called (line 499). The `reconnect()` method calls `getHub()` → `connect()`, which sets up a new `onConnection` callback and awaits the transport connection. When the transport fires `onConnection('connected')`, `_isResuming` is still `true`, so `connectionState.value = state` is skipped. Then `finally` sets `_isResuming = false`, but the `connected` event has already fired and been suppressed, leaving `connectionState` stuck at `'connecting'` forever.

**Path that breaks:**
1. `validateConnectionOnResume()` → `_isResuming = true`
2. `!this.messageHub || !this.transport` → `await this.reconnect()`
3. `reconnect()` → `await this.getHub()` → `await this.connect()`
4. Transport connects → `onConnection('connected')` fires → **suppressed** by `_isResuming` check
5. `finally` → `_isResuming = false` — but `connected` event already gone
6. `connectionState` stuck at `'connecting'` — UI shows loading forever

**Fix:** After `_isResuming = false` in the `finally` block, check if the connection is actually ready and update `connectionState` accordingly:

```typescript
} finally {
    this._isResuming = false;
    // If connection was re-established during resume, update state now
    if (this.transport?.isReady()) {
        connectionState.value = 'connected';
        this.notifyConnectionHandlers();
    }
}
```

---

## P2: Test quality — first tab-resume test doesn't verify the key scenario

The first test (`returns early and clears error when isConnected is false`) creates a fresh hook instance to simulate the re-render, rather than using `rerender()` on the existing hook. A fresh `renderHook` call creates a completely separate hook instance — it doesn't test the same hook reacting to an `isConnected` change. Compare with the second test which correctly uses `rerender()`.

```typescript
// Current (wrong) — creates separate hook instance
const { result: result2, rerender } = renderHook(() => useTaskViewData('room-1', 'task-1'));

// Should use rerender() on the SAME hook instance to test effect re-triggering
```

This means the first test doesn't actually verify that the same hook instance clears errors when `isConnected` changes — it only tests that a newly mounted hook with `isConnected=false` starts with no error, which is a weaker assertion.

---

## P3 (nit): Third test doesn't test the actual bug scenario well

The third test (`does not permanently show error when load() is called before connection is fully ready`) creates a completely new hook instance instead of re-rendering. Since `useTaskViewData` has `isConnected` in its deps, the actual bug scenario is: same hook, `isConnected` stays `true`, but `request()` fails mid-resume, then succeeds on the next effect run when `isConnected` transitions. This test should use `rerender()` and verify that the error clears when `isConnected` transitions from `true` → `false` → `true` (which is what actually happens with the connection-manager fix).

PR #762, opened 2026-03-23
fix: route human message to correct agent based on target parameter

lsm:
When a task is in needs_attention state and a human sends a message
to the worker, the runtime now respects the target parameter and
delivers to the worker instead of always preferring the leader.

Changes:
- Modified reviveTaskForMessage to accept a target parameter ('worker'
  or 'leader') instead of always preferring leader
- When target='worker': try worker first, fall back to leader
- When target='leader' (default): try leader first, fall back to worker
- Updated task.sendHumanMessage RPC handler to pass target parameter
- Added unit tests for target parameter routing

This fixes Issue 1 where human message to worker was incorrectly
routed to leader.

For Issue 2, the fix ensures that when the human explicitly sends
to worker, the message goes directly to worker without leader
intermediary, preventing unnecessary synthetic feedback generation.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1620,7 +1620,7 @@ export class RoomRuntime {
 	 * - Clears the group's completedAt so it becomes active again
 	 * - Restores agent sessions (they were stopped when the task failed)
 	 * - Re-registers terminal-state observers so the runtime hears when agents finish
-	 * - Injects the human message (leader first, worker as fallback)
+	 * - Injects the human message (respects target preference)
 	 *
 	 * On failure, undoes the group revive (re-sets completedAt) so the group is not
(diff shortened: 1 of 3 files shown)


lsm (commented):
LGTM. The fix is clean and minimal:

1. **Issue 1 (routing)**: reviveTaskForMessage now accepts a target parameter and respects the human agent selection. When target=worker, it tries worker first with leader fallback.

2. **Issue 2 (synthetic message)**: Verified this is not an independent issue. reviveTaskForMessage sets submittedForReview=true before injection (line 1654), which causes recoverZombieGroups to skip the group (line 2420). With the routing fix, the human message goes directly to the intended agent, so no synthetic intermediary message is generated.

3. **Tests**: All 98 tests pass. The 3 new tests cover the three key target routing scenarios. Existing tests were correctly updated to expect the new parameter.

Minor note: the setLeaderHasWork call when target=worker falls back to leader (line 1705) is correct defensive behavior — it prevents the leader terminal event from being dropped.

PR #764, opened 2026-03-23
test(e2e): add E2E tests for channel direction visualization

lsm:


lsm (commented):
## Review

Tests 13, 14, 15, and 16 look good. Test 14-16 are correctly skipped (arrowhead markers not yet implemented).

However, **Test 12 has a P1 bug** that will cause it to fail:

### P1: Test 12 — channel edges won't exist for RPC-created workflow

The test creates a workflow via `spaceWorkflow.create` RPC without including `channels` on the steps. The `channelEdges` derivation in `VisualWorkflowEditor.tsx:175-193` only produces edges for nodes that have `step.channels` with `direction === 'bidirectional'`. RPC-created workflows without explicit channels won't have any channel edges, so the assertion at line 728-729 (`await expect(channelEdge).toBeVisible()`) will time out and fail.

**Fix options:**
- **Option A (preferred):** Create the workflow via UI (`openNewWorkflowEditor` + `add-step-button`) which triggers the auto-creation of Task Agent channels. Then add a transition via the UI drag-to-connect, or use RPC only for adding the transition.
- **Option B:** Include `channels` in the RPC payload for each step, e.g.:
  ```ts
  steps: [
    { id: step1, name: 'Start', agentId: aId, channels: [{ from: 'task-agent', to: 'coder', direction: 'bidirectional' }] },
    { id: step2, name: 'End', agentId: aId, channels: [{ from: 'task-agent', to: 'coder', direction: 'bidirectional' }] },
  ],
  ```
  Note: verify the exact channel schema used by `spaceWorkflow.create`.

### P2: Incorrect comment

Line 726 says `// Task Agent channel edges should exist (auto-created when nodes exist)` — but auto-creation only happens via the UI node addition flow, not via RPC. Update the comment to match the actual mechanism.

### Nit: `waitForTimeout` usage

Test 16 (skipped) uses `page.waitForTimeout(500)`. Prefer waiting for a specific condition (e.g., attribute change) instead of a fixed timeout, to make the test more robust when unskipped.

PR #765, opened 2026-03-23
feat: integrate channel edges into WorkflowCanvas

lsm:
- Move channel edge computation into WorkflowCanvas from nodes data
- Compute channel edges from step.channels on each node
- Resolve agent roles to node IDs for inter-node channels
- Skip intra-node channels (same node for from/to)
- Support both task-agent and regular node-to-node channels
- Render arrowheads based on direction (bidirectional vs one-way)
- Bidirectional: arrowhead on both ends (↔)
- One-way: arrowhead only on target end (→)
- Maintain backward compatibility with optional channelEdges prop
- Remove redundant channelEdges computation from VisualWorkflowEditor
- Add unit tests for computeChannelEdges function

Test plan:
- 36 tests pass in WorkflowCanvas test file
- Typecheck and lint clean

### DIFF
--- a/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
+++ b/packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx
@@ -123,6 +123,127 @@ function GhostEdge({ from, to }: { from: Point; to: Point }): JSX.Element | null
 	);
 }
 
+// ---- Channel edge computation ----
+
+/**
+ * Compute channel edges from workflow nodes.
+ *
+ * Collects all WorkflowChannel declarations from all nodes and resolves
+ * channel endpoints to node IDs:
+ * - Intra-node channels (agents within the same node) are skipped - they don't
+ *   need cross-node edges since all agents in a node share the same canvas space.
+ * - Inter-node channels (including Task Agent channels) are resolved to
+ *   actual node IDs and returned as ResolvedWorkflowChannel objects.
+ *
+ * Task Agent channels use 'task-agent' as the special source identifier.
+ * Other channels resolve agent roles to their containing node IDs using the
+ * SpaceAgent.agents array which has role information.
+ */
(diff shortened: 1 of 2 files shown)


lsm (commented):
## Review Summary

Good implementation overall. The `computeChannelEdges` function is well-structured with proper role-to-node resolution, intra-node skipping, and array `to` support. The direction-aware arrowhead rendering with SVG markers is clean.

## Issues

### P2: Missing test for inter-node (node-to-node) channels

The tests cover task-agent channels, intra-node channels, wildcards, array `to`, and unknown roles — but there's **no test for channels between two regular (non-task-agent) nodes**. This is one of the key new capabilities (previously only task-agent channels were supported). Add a test like:

```
Node A (agents: [coder])  channels: [{ from: 'coder', to: 'reviewer', direction: 'bidirectional' }]
Node B (agents: [reviewer]) channels: []
=> Should produce edge { fromStepId: 'node-a', toStepId: 'node-b', direction: 'bidirectional' }
```

### P3: Potential duplicate edges when same `(from, to)` pair appears in multiple nodes

`computeChannelEdges` has no deduplication. If two different nodes both declare channels that resolve to the same `(fromStepId, toStepId)` pair (e.g., both nodes have `{ from: 'task-agent', to: 'some-role' }` where `some-role` appears in a third node), duplicate edges will be rendered. Consider adding a `Set` or dedup step. This is low-severity because `validateStepChannels` prevents duplicate roles across nodes, but it's still worth defending against.

### P3: Comment inconsistency

The JSDoc on `ChannelEdgeRenderer` (line 270) says "Unidirectional" but the actual enum value is `'one-way'`. Should say "One-way" for consistency.

## Nits

- The `channelEdges` prop is kept with `_` prefix for "backward compatibility" but there are no other consumers. If `VisualWorkflowEditor` was the only caller (now removed), consider fully removing the prop to avoid dead API surface.

PR #771, opened 2026-03-24
fix: reduce redundant concurrent tick spawn attempts

lsm:
Add an in-memory `spawningTaskIds` set to guard against concurrent ticks
re-attempting spawn for the same task during the async window between tick
fire and DB group record creation (while task is still 'pending').

- executeTick() now skips tasks already tracked in spawningTaskIds
- spawnGroupForTask() wraps the inner work in try/finally to add/remove
  the task ID from spawningTaskIds for the duration of the spawn
- UNIQUE constraint catch block log level demoted from warn → debug since
  the DB constraint is now a last-resort safety net, not the primary guard
- Existing warn log for pre-existing active groups also demoted to debug
  (same reasoning — the spawningTaskIds check catches this earlier)
- New unit test: concurrent ticks fired together via Promise.all produce
  exactly one group, one worker session, and one leader session

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -196,6 +196,14 @@ export class RoomRuntime {
 	 */
 	private stuckWorkerRecoveryInFlight = new Set<string>();
 
+	/**
+	 * Task IDs whose group spawn is currently in-flight (async).
+	 * Guards against concurrent ticks re-attempting spawn for the same task while
+	 * `spawnGroupForTask` is awaiting worktree creation or DB insert — the window
+	 * where the task is still 'pending' in the DB but a spawn is already underway.
+	 */
+	private spawningTaskIds = new Set<string>();
+
 	readonly taskGroupManager: TaskGroupManager;
 
 	/**
@@ -2782,6 +2790,12 @@ export class RoomRuntime {
 				);
 				continue;
 			}
+			if (this.spawningTaskIds.has(task.id)) {
+				log.debug(
+					`[executeTick] Task ${task.id} ("${task.title}") skipped — spawn already in-flight`
+				);
+				continue;
+			}
 			if (await this.taskManager.areDependenciesMet(task)) {
 				readyTasks.push(task);
 			} else {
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

The fix is correct, well-scoped, and doesn't introduce new risks. The `spawningTaskIds` set solves exactly the stated problem: concurrent ticks seeing the same task as 'pending' during the async window before the DB record is written. The `try/finally` pattern is the right idiom here since it guarantees cleanup even if `_spawnGroupForTaskInner` throws.

---

### Correctness

**Guard placement is correct.** `spawningTaskIds.add(task.id)` executes synchronously before the first `await` in `spawnGroupForTask`, so there is no window where a second caller could enter the guard concurrently. JavaScript's cooperative multitasking guarantees this.

**Defense-in-depth is preserved.** The existing active-group DB check and UNIQUE constraint catch block remain in `_spawnGroupForTaskInner` as safety nets for scenarios the in-memory guard cannot cover (e.g., daemon restart with concurrent ticks). The log level demotion from `warn` → `debug` for those paths is appropriate given they are now last-resort fallbacks rather than expected paths.

**`try/finally` cleanup is correct.** On any failure path through `_spawnGroupForTaskInner` (including UNIQUE constraint), the task ID is removed from `spawningTaskIds`, so subsequent ticks can retry. A failed spawn leaves the task as `in_progress` (via `startTask`) or `pending` depending on where it fails, and the active-group or zombie-recovery logic will handle the next tick correctly.

---

### Minor Issues

**P3 — Log prefix mismatch after rename.** The log messages inside `_spawnGroupForTaskInner` still use the prefix `[spawnGroupForTask]` (lines 3381 and 3498 in `room-runtime.ts`). This is cosmetically inconsistent since the enclosing function is now `_spawnGroupForTaskInner`. It won't cause bugs but adds friction when grepping logs to trace execution. Since the outer wrapper is also called `spawnGroupForTask`, this is borderline acceptable — but worth noting.

**P3 — Redundant guard in `spawnGroupForTask`.** The `spawningTaskIds.has()` check appears in two places: the optimization pass inside `executeTick` (line 2793) and the authoritative gate in `spawnGroupForTask` (line 3353). The check in `spawnGroupForTask` is the one that actually prevents double-spawn; the one in `executeTick` avoids adding the task to `readyTasks` as an early optimization. Both are correct. The redundancy is intentional and well-commented, so this is just an observation.

---

### Test

The new test at line 116–141 of `room-runtime.test.ts` correctly uses `Promise.all([tick1, tick2])` to start both ticks concurrently, which is the right way to exercise the race condition in a single-threaded async runtime. The assertions cover exactly what the fix promises: one group, one worker session, one leader session. The test also complements the existing sequential-tick idempotency test.

One note on test fidelity: the mock `createWorktree` resolves immediately, so the actual interleaving that triggers the bug in production (the long async window while waiting for real worktree creation) is not reproducible in tests. The concurrent test passes because both ticks _do_ start simultaneously and the `spawningTaskIds` guard fires correctly. This is fine — the guard logic is deterministic and the test validates the observable outcome. A more rigorous test would inject a delay in `createWorktree` to force the interleaving, but that would require modifying the test helper for a single test case and is not worth the complexity.

---

### Summary

Straightforward and correct fix. The implementation is minimal (36 lines added), uses the right patterns, and doesn't alter any existing behavior outside the logging level changes. The log demotion is appropriate given the guard now covers the common case. No security, data-loss, or architectural concerns.

PR #772, opened 2026-03-24
fix: delete stale branch instead of falling back to UUID in worktree-manager

lsm:
When a task branch (e.g. task/task-42-...) already exists from a prior
run that crashed mid-task, git worktree add -b would fail with "branch
already exists", blocking group spawning entirely.

Previously the code only handled this for customBranchName and fell back
to an opaque session/UUID name. Now the pre-check runs for all branch
names and deletes the stale branch so the original intended name is
preserved and worktree creation succeeds.

### DIFF
--- a/packages/daemon/src/lib/worktree-manager.ts
+++ b/packages/daemon/src/lib/worktree-manager.ts
@@ -227,7 +227,17 @@ export class WorktreeManager {
 			const branchExists = await this.checkBranchExists(gitRoot, branchName);
 			if (branchExists) {
 				this.logger.warn(`Stale branch detected: ${branchName} — deleting and recreating`);
-				await git.branch(['-D', branchName]);
+				try {
+					await git.branch(['-D', branchName]);
+				} catch {
+					// git refuses -D when the branch is currently checked out in another
+					// living worktree.  Fall back to a unique session-scoped name so the
+					// task can still proceed rather than blocking entirely.
+					this.logger.warn(
+						`Could not delete branch ${branchName} (may be checked out in another worktree) — falling back to session/${safeSessionId}`
+					);
+					branchName = `session/${safeSessionId}`;
+				}
 			}
 
 			// Create worktree with new branch
@@ -406,8 +416,8 @@ export class WorktreeManager {
 						await git.raw(['worktree', 'remove', worktree.path, '--force']);
 						cleaned.push(worktree.path);
 
-						// Also try to delete the branch if it's a session branch
-						if (worktree.branch.startsWith('session/')) {
+						// Also try to delete the branch if it's a managed branch (session/ or task/)
+						if (worktree.branch.startsWith('session/') || worktree.branch.startsWith('task/')) {
 							try {
 								await git.branch(['-D', worktree.branch]);
 							} catch {
--- a/packages/daemon/src/lib/worktree-manager.ts
+++ b/packages/daemon/src/lib/worktree-manager.ts
@@ -220,12 +220,14 @@ export class WorktreeManager {
 				throw new Error(`Worktree directory already exists: ${worktreePath}`);
 			}
 
-			// Check if branch already exists (and fallback to UUID if it does)
-			if (customBranchName) {
-				const branchExists = await this.checkBranchExists(gitRoot, customBranchName);
-				if (branchExists) {
-					branchName = `session/${safeSessionId}`; // Fallback to UUID-based branch
-				}
+			// Check if branch already exists — this can happen when a prior task/session
+			// crashed mid-run and left behind a stale branch whose worktree was already
+			// removed. Delete the stale branch so we can recreate it fresh with the
+			// same (intended) name instead of falling back to an opaque UUID-based name.
+			const branchExists = await this.checkBranchExists(gitRoot, branchName);
+			if (branchExists) {
+				this.logger.warn(`Stale branch detected: ${branchName} — deleting and recreating`);
+				await git.branch(['-D', branchName]);
 			}
 
 			// Create worktree with new branch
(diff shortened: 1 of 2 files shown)


lsm (commented):

The implementation is directionally correct — moving the stale-branch pre-check to cover all branch names (not just `customBranchName`) is the right fix. The logic is clean and the three new tests validate the core happy-path scenarios. However, there are a few issues worth addressing.

---

### P0 — Missing error handling when `git branch -D` is rejected by git

`packages/daemon/src/lib/worktree-manager.ts`, lines 228–231

```ts
if (branchExists) {
    this.logger.warn(`Stale branch detected: ${branchName} — deleting and recreating`);
    await git.branch(['-D', branchName]);
}
```

Git will refuse `-D` when the branch is currently checked out in a **living worktree** (even if the NeoKai worktree directory was removed from the filesystem, the git worktree registration can survive until `git worktree prune` is run). In that case the command throws:

```
error: cannot delete branch 'task/task-42' checked out at '/path/to/other-worktree'
```

This error propagates to the outer `catch` on line 251, which tries to clean up the worktree directory (not yet created, so `existsSync` is false, nothing happens) and then re-throws as `"Failed to create worktree: ..."` — masking what actually went wrong. The caller gets an opaque error with no indication that a checked-out branch collision occurred and no recovery path.

**Minimum fix:** catch the `-D` failure separately, log a specific warning ("branch still checked out in a live worktree — skipping delete"), and either throw with a clear message or fall back to the session/UUID name so the task can continue.

---

### P1 — No test for `git branch -D` failure (branch still checked out)

`packages/daemon/tests/unit/lib/worktree-manager.test.ts`

There is no test for the scenario where `checkBranchExists` returns true but `git branch -D` itself throws (e.g., the branch is checked out in an active worktree that git still tracks). The outer `catch` will fire, but this code path has no coverage. Given that this is the case that would most commonly follow an incomplete previous cleanup, it is an important edge case to test.

---

### P1 — Test comment is misleading about which mock intercepts `branch -D`

`packages/daemon/tests/unit/lib/worktree-manager.test.ts`, lines 173–174

```ts
mockGitRaw
    .mockResolvedValueOnce('  custom-branch\n') // checkBranchExists — stale branch found
    .mockResolvedValue(''); // branch -D + git worktree add
```

`git branch -D` goes through `git.branch(['-D', branchName])`, which calls `mockGitBranch`, not `mockGitRaw`. The second `mockResolvedValue('')` on `mockGitRaw` therefore only covers `worktree add`. The comment `// branch -D + git worktree add` incorrectly implies both calls go through `mockGitRaw`. This same pattern is repeated in the two new tests and is a minor but real source of confusion for future readers. The same pattern exists in the original test, so this is a pre-existing issue that the PR copies — worth cleaning up.

---

### P2 — No test for the clean (no stale branch) auto-generated-branch path

The original test `'should fallback to UUID branch if custom branch exists'` has been replaced by the new stale-delete behaviour. There is an existing test `'should return WorktreeMetadata on success'` that uses a custom `branchName`, but there is no test that explicitly verifies `createWorktree` succeeds without a custom branch name **and** without a stale branch present (i.e., the auto-generated `session/session-123` path when `checkBranchExists` returns empty). The `'should create worktree directory if it does not exist'` test comes close but does not assert the returned branch name.

---

### P2 — `cleanupOrphanedWorktrees` only deletes `session/` branches, but the crash-recovery path now also leaves `task/` branches

`packages/daemon/src/lib/worktree-manager.ts`, lines 408–413

```ts
if (worktree.branch.startsWith('session/')) {
    try {
        await git.branch(['-D', worktree.branch]);
    } catch {
        // Could not delete branch, but this is non-fatal
    }
}
```

`cleanupOrphanedWorktrees` only prunes branches whose names start with `session/`. Now that task branches (`task/task-42-...`) are also managed by this code path, a crashed task worktree will be pruned from git's worktree list but its `task/` branch will be left behind — exactly the staleness this PR is trying to fix. This is not introduced by this PR (it is pre-existing), but the PR is the right moment to widen the cleanup guard to also cover `task/` prefixes. Without that, the stale-branch detection added here will be the only recovery mechanism, and periodic cleanup via `cleanupOrphanedWorktrees` will keep accumulating stale `task/` branches.

lsm (commented):

**Recommendation: APPROVE**

All three issues from the previous review iteration have been addressed correctly and completely. Verification follows.

---

### P0 resolved — `git branch -D` failure is now caught with a proper fallback

`packages/daemon/src/lib/worktree-manager.ts`, lines 230–240

The inner `try/catch` block around `git.branch(['-D', branchName])` is correct. On failure it logs a clear, actionable warning and sets `branchName = `session/${safeSessionId}`` before execution falls through to the `worktree add` call. The outer `catch` on line 261 is not triggered by this path.

One minor observation worth noting (not blocking): if the original `branchName` was already `session/${safeSessionId}` (i.e., no custom branch name was provided), the fallback reassigns the same value it already holds — harmless, but slightly redundant. This is not a bug.

---

### P1 resolved — `cleanupOrphanedWorktrees` now covers `task/` branches

`packages/daemon/src/lib/worktree-manager.ts`, line 420

The guard was correctly widened to:
```ts
if (worktree.branch.startsWith('session/') || worktree.branch.startsWith('task/')) {
```
This prevents stale `task/` branches from accumulating after orphaned worktree cleanup runs.

---

### P2 resolved — Test comments and coverage

All three comment accuracy fixes and two new tests are in place:

- `'should succeed with auto-generated branch name when no stale branch exists'` — covers the clean path where `checkBranchExists` returns empty and `mockGitBranch` is not called. Correct.
- `'should fall back to UUID branch name when branch -D is rejected'` — simulates `mockGitBranch.mockRejectedValueOnce` and asserts the result is `session/session-123`. Correct.
- `'should delete task/ branches for orphaned task worktrees'` — exercises the new cleanup guard. Correct.
- Test comments on all stale-branch tests now accurately note that `branch -D` goes through `mockGitBranch`, not `mockGitRaw`.

---

### No new issues found

The implementation is complete, the test suite is thorough, and no regressions were introduced. The code is ready to merge.

PR #773, opened 2026-03-24
fix: improve SDK startup timeout — exponential backoff, more retries, better diagnostics

lsm:
- Increase auto-recovery delay from 300ms to 3s base with exponential backoff
  (attempt 1: 3s, attempt 2: 6s) giving the SDK subprocess time to fully exit
- Increase max retries from 1 to 2 (3 total attempts) via STARTUP_MAX_RETRIES
- Add NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS and NEOKAI_SDK_STARTUP_MAX_RETRIES
  env vars for runtime tuning without a redeploy
- Log root workspace vs worktree, attempt number, and timeout hint on startup timeout
- Pass actionable user message when all retries exhausted: workspace path, common
  causes (workspace contention, stale lock), and NEOKAI_SDK_STARTUP_TIMEOUT_MS hint
- Update tests: fake timers for delay assertions, second-retry coverage, exhausted-
  retries user message assertion, exponential backoff scheduling verification

### DIFF
--- a/packages/daemon/.env.example
+++ b/packages/daemon/.env.example
@@ -58,6 +58,27 @@ TEMPERATURE=1.0
 MAX_SESSIONS=10
 
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+# SDK Startup Tuning
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+#
+# How long (ms) to wait for the SDK subprocess to emit its first message
+# before declaring a startup timeout. Increase on slow machines or when
+# running many concurrent sessions.
+# Default: 15000
+# NEOKAI_SDK_STARTUP_TIMEOUT_MS=15000
+#
+# Base delay (ms) before the first auto-recovery attempt after a startup
+# timeout. Each subsequent attempt doubles this (exponential backoff).
+# Example: 3000 → attempt 1 waits 3s, attempt 2 waits 6s.
+# Default: 3000
+# NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS=3000
+#
+# Maximum number of auto-recovery attempts before surfacing the error to
+# the user. Set to 0 to disable auto-recovery entirely.
+# Default: 2
+# NEOKAI_SDK_STARTUP_MAX_RETRIES=2
+
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 # Job Queue Configuration
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 #
--- a/packages/daemon/src/lib/agent/query-runner.ts
+++ b/packages/daemon/src/lib/agent/query-runner.ts
@@ -488,17 +488,22 @@ export class QueryRunner {
 						// For startup timeouts / conversation-not-found that exhausted all retries,
 						// provide actionable recovery hints. Both error types are handled symmetrically
 						// (same retry gate, same sdkSessionId clearing), so both deserve a hint.
-						const startupTimeoutUserMessage =
-							isStartupTimeout || isConversationNotFound
-								? `The AI session failed to ${isStartupTimeout ? 'start' : 'resume'} after ${STARTUP_MAX_RETRIES + 1} attempt(s) ` +
+						// Keep the hints distinct: NEOKAI_SDK_STARTUP_TIMEOUT_MS is irrelevant to a
+						// missing/corrupt session file — the session ID was already cleared above,
+						// so the next message will automatically start a fresh session.
+						const startupTimeoutUserMessage = isStartupTimeout
+							? `The AI session failed to start after ${STARTUP_MAX_RETRIES + 1} attempt(s) ` +
+								`(workspace: ${session.workspacePath}). ` +
+								`Common causes: another Claude Code session is using the same workspace, ` +
+								`a stale lock file in .claude/, or the workspace is under heavy load. ` +
+								`Try: closing other Claude sessions on this workspace, ` +
+								`then resend your message. ` +
+								`You can also increase the timeout with NEOKAI_SDK_STARTUP_TIMEOUT_MS (current: ${STARTUP_TIMEOUT_MS}ms).`
+							: isConversationNotFound
+								? `The AI session could not be resumed after ${STARTUP_MAX_RETRIES + 1} attempt(s) ` +
 									`(workspace: ${session.workspacePath}). ` +
-									(isStartupTimeout
-										? `Common causes: another Claude Code session is using the same workspace, ` +
-											`a stale lock file in .claude/, or the workspace is under heavy load. `
-										: `The previous session file could not be found or is corrupt. `) +
-									`Try: closing other Claude sessions on this workspace, ` +
-									`then resend your message. ` +
-									`You can also increase the timeout with NEOKAI_SDK_STARTUP_TIMEOUT_MS (current: ${STARTUP_TIMEOUT_MS}ms).`
+									`The previous session file could not be found or is corrupt. ` +
+									`The session has been reset automatically — please resend your message to start fresh.`
 								: undefined;
 
 						await errorManager.handleError(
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE** (with minor follow-up suggestions)

The implementation correctly delivers all five stated goals: exponential backoff, increased retry count, env var configurability, root-workspace diagnostics, and improved user-facing error messages. The logic is sound and the tests cover the new behavior. A few issues worth addressing in a follow-up.

---

### P1 — New env vars missing from `.env.example`

`NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS` and `NEOKAI_SDK_STARTUP_MAX_RETRIES` are actively surfaced to users in the exhausted-retries error message:

```
You can also increase the timeout with NEOKAI_SDK_STARTUP_TIMEOUT_MS (current: 15000ms).
```

But none of the three SDK startup tuning variables (`NEOKAI_SDK_STARTUP_TIMEOUT_MS`, `NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS`, `NEOKAI_SDK_STARTUP_MAX_RETRIES`) appear in `packages/daemon/.env.example`. A user following the hint has no reference file showing the syntax. These should be added as commented-out examples with their defaults and a short description.

---

### P2 — Exponential backoff test doesn't verify delay values

`packages/daemon/tests/unit/agent/query-runner.test.ts` lines 855–874:

The test titled "should use exponential backoff: second attempt delay is 2x the first" only asserts that `startupTimeoutAutoRecoverAttempts` is incremented to 1 and 2 respectively. The delay passed to `setTimeout` is never checked. The test name promises delay verification but delivers only counter verification — the backoff formula could be replaced with a constant and this test would still pass.

A better approach with the already-imported `jest.useFakeTimers()` would be to spy on `setTimeout`, capture the delay argument, and assert `delay1 === BASE` and `delay2 === 2 * BASE`. The infrastructure for fake timers is already established in adjacent tests.

---

### P2 — `isConversationNotFound` exhausted retries produces no user hint

`packages/daemon/src/lib/agent/query-runner.ts` lines 486–494:

`startupTimeoutUserMessage` is only set when `isStartupTimeout` is true. When retries are exhausted due to `isConversationNotFound`, the user message passed to `errorManager.handleError` is `undefined`. The two error types are handled symmetrically everywhere else in the catch block (same `canAutoRecover` gate, same `sdkSessionId` clearing), so the asymmetry in user messaging is likely unintentional. Users hitting a conversation-not-found loop after exhausting retries get no actionable hint.

---

### P2 — `isCleaningUp` test still uses real-time sleep

`packages/daemon/tests/unit/agent/query-runner.test.ts` line 814:

The test "should NOT invoke onStartupTimeoutAutoRecover when isCleaningUp returns true" still waits `400ms` of real clock time. Now that adjacent tests use `jest.useFakeTimers()` and `jest.runAllTimers()`, this test could be converted the same way for consistency and speed.

---

### P3 — Module-level constants read env vars once at import time

`packages/daemon/src/lib/agent/query-runner.ts` lines 54–56:

`STARTUP_TIMEOUT_MS`, `STARTUP_RECOVERY_DELAY_MS`, and `STARTUP_MAX_RETRIES` are read once when the module is first imported. This is the same pattern used by the pre-existing `STARTUP_TIMEOUT_MS`, so it is consistent. However, the user-facing error message prints the module-level constant value, meaning that if the env var is set after the process starts (unusual but possible) the displayed hint value would be stale. Worth a brief code comment noting this is intentionally read at module load.

lsm (commented):

**Recommendation: APPROVE** — The follow-up work is solid. All five items from the previous review are addressed correctly. A few minor observations below.

---

### What was fixed

1. **`.env.example` documentation** — The three new env vars are documented with correct defaults, clear descriptions, and the exponential backoff example (`3000 → attempt 1: 3s, attempt 2: 6s`) makes the behavior concrete. Well done.

2. **Exponential backoff test** — The setTimeout spy approach in `should use exponential backoff: second attempt delay is 2× the first` (line 858) correctly captures actual delay arguments passed to the real scheduler and asserts the 2× relationship without relying on wall-clock time. This is a meaningful improvement over the previous version.

3. **`isConversationNotFound` user hint** — The hint in `startupTimeoutUserMessage` (lines 491–502) now branches on `isStartupTimeout` vs `isConversationNotFound` and gives context-appropriate copy. Structurally correct.

4. **`isCleaningUp` test** — Converted to fake timers, no more 400ms real sleep. The `jest.useFakeTimers()` / `jest.useRealTimers()` pattern is consistent with the other timer-based tests.

5. **Module-level comment** (line 54–57) — Clear, accurate, and correctly explains the read-once-at-import-time behavior.

---

### Issues

**P2 — Misleading timeout hint in the `isConversationNotFound` user message**

```
packages/daemon/src/lib/agent/query-runner.ts, lines 500–501
```

The exhausted-retries user message for `isConversationNotFound` ends with:

> You can also increase the timeout with NEOKAI_SDK_STARTUP_TIMEOUT_MS (current: Xms).

`NEOKAI_SDK_STARTUP_TIMEOUT_MS` controls how long to wait before declaring a *startup* timeout. It is not relevant to a "conversation not found" failure, which is a missing/corrupt session file — not a slow startup. Advising the user to increase the startup timeout is confusing and unlikely to help. The hint should either be omitted for the `isConversationNotFound` branch, or replaced with something like "The session file has been cleared automatically; your next message will start a fresh session.", which is actually what happens after `sdkSessionId` is cleared.

**P3 — `capturedDelays` index assumption in the backoff test**

```
packages/daemon/tests/unit/agent/query-runner.test.ts, lines 890–894
```

The test asserts `capturedDelays.length >= 2` and then reads `[0]` / `[1]`. If any other `setTimeout` call happens to be intercepted before the recovery-defer call (e.g., from an internal async path added later), the indices could shift and the test would pass incorrectly. The comment says "Each run captures one setTimeout call (the recovery defer)" — this is correct today, but the `>= 2` lower-bound check and positional indexing are fragile. A more robust approach would be to filter `capturedDelays` to only the two recovery-defer values (e.g., values `> 0`) or assert `capturedDelays.length === 2` exactly. This is a nit — not a correctness issue in the current codebase.

**P3 — `isConversationNotFound` test coverage gap**

There is no test that exercises the exhausted-retries user message path for the `isConversationNotFound` branch specifically. The existing `should pass actionable user message to handleError when all retries exhausted` test (line 909) injects a startup-timeout error string. It would be good to have an analogous test with a `No conversation found` error to verify the separate copy branch is exercised.

lsm (commented):

**Recommendation: APPROVE**

All three issues raised in the second review iteration are correctly addressed. The implementation is solid — this PR is ready to merge.

---

### What was fixed in this iteration

**1. isConversationNotFound exhausted-retries user message** — The `startupTimeoutUserMessage` ternary (lines 494–507 of `query-runner.ts`) now branches correctly: startup timeout keeps the `NEOKAI_SDK_STARTUP_TIMEOUT_MS` hint; conversation-not-found replaces it with "The session has been reset automatically — please resend your message to start fresh." The copy is accurate and actionable.

**2. Backoff test tightened to `=== 2`** — Line 890 of the test file now asserts `capturedDelays.length` is exactly 2, not `>= 2`. This protects against index drift if an unrelated `setTimeout` call is added to the call path in the future.

**3. New `isConversationNotFound` test** — Lines 927–947 add a dedicated test that injects a `No conversation found` error, asserts the correct copy is present (`session has been reset automatically`), and explicitly verifies `NEOKAI_SDK_STARTUP_TIMEOUT_MS` does **not** appear in the message. Both the positive and negative assertions are correctly scoped.

---

### One remaining observation (P3 — nit, not blocking)

**`globalThis.setTimeout` spy may catch unrelated calls from async internals**

`packages/daemon/tests/unit/agent/query-runner.test.ts`, lines 864–866

The spy pushes every `delay` argument to `capturedDelays`, including any `setTimeout(fn, 0)` calls that happen to run while `globalThis.setTimeout` is replaced. The test currently passes because each `QueryRunner` run produces exactly one recovery-defer `setTimeout` in the code under test. The `=== 2` assertion (just fixed) will catch a future regression if a second call appears. However, one subtle risk remains: the spy fires `originalSetTimeout(fn, 0)` for every intercepted call — if the startup-timeout timer path (line 284, `setTimeout(() => { ... }, STARTUP_TIMEOUT_MS)`) ever gets intercepted (e.g., if the auth check happens to pass before the timer line is reached), that 0ms fire would trigger the `startupTimeoutReached` path unexpectedly and could produce a third entry in `capturedDelays`. In the current setup `buildSpy.mockRejectedValue` throws before the timer is set up, so this does not fire today. This is a benign edge case worth a clarifying comment, not a bug fix.

No action required — the implementation is correct, the tests are meaningful, and the user-facing copy is improved.

PR #774, opened 2026-03-24
feat(visual-editor): add Task Agent virtual node to data model

lsm:
- Define TASK_AGENT_NODE_ID ('__task_agent__') constant in shared/types/space.ts
  as the stable ID for the virtual Task Agent node (never persisted in DB)
- Add TaskAgentVisualNode type in visual-editor/types.ts
- workflowToVisualState() now always injects the Task Agent node at index 0,
  positioned above all regular nodes (computed from layout or auto-layout)
- buildWorkflowFields() strips the Task Agent node before serialization so
  it is never included in DB-persisted node lists or layout maps
- autoLayout() now includes TASK_AGENT_NODE_ID pinned at the top-center of
  the canvas (START_Y bumped from 50→170 to leave room above)
- VisualWorkflowEditor save validation skips the virtual Task Agent node
  to avoid spurious "requires an agent" errors
- Updated all affected tests: layout (19), serialization (+9 new Task Agent
  tests, 58 total), performance (+1), VisualWorkflowEditor (fixes 5 tests)

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -196,6 +196,14 @@ export class RoomRuntime {
 	 */
 	private stuckWorkerRecoveryInFlight = new Set<string>();
 
+	/**
+	 * Task IDs whose group spawn is currently in-flight (async).
+	 * Guards against concurrent ticks re-attempting spawn for the same task while
+	 * `spawnGroupForTask` is awaiting worktree creation or DB insert — the window
+	 * where the task is still 'pending' in the DB but a spawn is already underway.
+	 */
+	private spawningTaskIds = new Set<string>();
+
 	readonly taskGroupManager: TaskGroupManager;
 
 	/**
@@ -2782,6 +2790,12 @@ export class RoomRuntime {
 				);
 				continue;
 			}
+			if (this.spawningTaskIds.has(task.id)) {
+				log.debug(
+					`[executeTick] Task ${task.id} ("${task.title}") skipped — spawn already in-flight`
+				);
+				continue;
+			}
 			if (await this.taskManager.areDependenciesMet(task)) {
 				readyTasks.push(task);
 			} else {
@@ -3333,6 +3347,24 @@ export class RoomRuntime {
 	 * Reads task.assignedAgent to pick the appropriate worker factory.
 	 */
 	private async spawnGroupForTask(task: NeoTask): Promise<void> {
+		// Guard: skip immediately if this task's spawn is already in-flight from a prior tick.
+		// This prevents redundant spawn attempts during the async window between a tick firing
+		// and the group DB record being created (task is still 'pending' during that window).
+		if (this.spawningTaskIds.has(task.id)) {
+			log.debug(
+				`[spawnGroupForTask] Task ${task.id} ("${task.title}") — spawn already in-flight, skipping`
+			);
+			return;
+		}
+		this.spawningTaskIds.add(task.id);
+		try {
+			await this._spawnGroupForTaskInner(task);
+		} finally {
+			this.spawningTaskIds.delete(task.id);
+		}
+	}
+
+	private async _spawnGroupForTaskInner(task: NeoTask): Promise<void> {
 		// Clean zombie groups for this task before the active-group dedup check.
 		// Handles the case where a previous crashed session left an active group with
 		// missing sessions. Without this, getActiveGroupsForTask() would detect the
@@ -3345,7 +3377,7 @@ export class RoomRuntime {
 		// completedAt === null would be missed by getGroupByTaskId() which returns only the latest.
 		const allActiveGroups = this.groupRepo.getActiveGroupsForTask(task.id);
 		if (allActiveGroups.length > 0) {
-			log.warn(
+			log.debug(
 				`[spawnGroupForTask] Task ${task.id} ("${task.title}") already has ${allActiveGroups.length} active group(s) (${allActiveGroups.map((g) => g.id).join(', ')}) — skipping duplicate spawn`
 			);
 			return;
@@ -3458,11 +3490,11 @@ export class RoomRuntime {
 			);
 		} catch (err) {
 			// UNIQUE constraint violation means a concurrent tick already spawned a group
-			// for this task (race between two ticks that both passed the active-group check).
-			// This is expected under high concurrency — log at warn, not error.
+			// for this task. With the spawningTaskIds guard this should be rare, but the
+			// DB constraint remains as a final safety net — log at debug to reduce noise.
 			const errStr = String(err);
 			if (errStr.includes('UNIQUE constraint failed')) {
-				log.warn(
+				log.debug(
 					`[spawnGroupForTask] Task ${task.id} ("${task.title}"): UNIQUE constraint — ` +
 						`concurrent tick already spawned a group. Skipping.`
 				);
--- a/packages/daemon/src/lib/worktree-manager.ts
+++ b/packages/daemon/src/lib/worktree-manager.ts
@@ -220,11 +220,23 @@ export class WorktreeManager {
 				throw new Error(`Worktree directory already exists: ${worktreePath}`);
 			}
 
-			// Check if branch already exists (and fallback to UUID if it does)
-			if (customBranchName) {
-				const branchExists = await this.checkBranchExists(gitRoot, customBranchName);
-				if (branchExists) {
-					branchName = `session/${safeSessionId}`; // Fallback to UUID-based branch
+			// Check if branch already exists — this can happen when a prior task/session
+			// crashed mid-run and left behind a stale branch whose worktree was already
+			// removed. Delete the stale branch so we can recreate it fresh with the
+			// same (intended) name instead of falling back to an opaque UUID-based name.
+			const branchExists = await this.checkBranchExists(gitRoot, branchName);
+			if (branchExists) {
+				this.logger.warn(`Stale branch detected: ${branchName} — deleting and recreating`);
+				try {
+					await git.branch(['-D', branchName]);
+				} catch {
+					// git refuses -D when the branch is currently checked out in another
+					// living worktree.  Fall back to a unique session-scoped name so the
+					// task can still proceed rather than blocking entirely.
+					this.logger.warn(
+						`Could not delete branch ${branchName} (may be checked out in another worktree) — falling back to session/${safeSessionId}`
+					);
+					branchName = `session/${safeSessionId}`;
 				}
 			}
 
@@ -404,8 +416,8 @@ export class WorktreeManager {
 						await git.raw(['worktree', 'remove', worktree.path, '--force']);
 						cleaned.push(worktree.path);
 
-						// Also try to delete the branch if it's a session branch
-						if (worktree.branch.startsWith('session/')) {
+						// Also try to delete the branch if it's a managed branch (session/ or task/)
+						if (worktree.branch.startsWith('session/') || worktree.branch.startsWith('task/')) {
 							try {
 								await git.branch(['-D', worktree.branch]);
 							} catch {
--- a/packages/daemon/tests/unit/lib/worktree-manager.test.ts
+++ b/packages/daemon/tests/unit/lib/worktree-manager.test.ts
@@ -161,7 +161,55 @@ describe('WorktreeManager', () => {
 			).rejects.toThrow('already exists');
 		});
 
-		it('should fallback to UUID branch if custom branch exists', async () => {
+		it('should succeed with auto-generated branch name when no stale branch exists', async () => {
+			existsSyncResults.set('/test/repo/.git', true);
+			existsSyncResults.set('/home/testuser/.neokai/projects/-test-repo/worktrees', true);
+			existsSyncResults.set(
+				'/home/testuser/.neokai/projects/-test-repo/worktrees/session-123',
+				false
+			);
+			mockGitRevparse.mockResolvedValue('.git');
+			// checkBranchExists returns empty → no stale branch, then worktree add succeeds
+			mockGitRaw
+				.mockResolvedValueOnce('') // checkBranchExists — branch does not exist
+				.mockResolvedValue(''); // worktree add
+
+			const result = await manager.createWorktree({
+				sessionId: 'session-123',
+				repoPath: '/test/repo',
+				// No custom branch name — uses auto-generated session/session-123
+			});
+
+			expect(result?.branch).toBe('session/session-123');
+			expect(mockGitBranch).not.toHaveBeenCalled();
+		});
+
+		it('should delete stale custom branch and reuse original name', async () => {
+			existsSyncResults.set('/test/repo/.git', true);
+			existsSyncResults.set('/home/testuser/.neokai/projects/-test-repo/worktrees', true);
+			existsSyncResults.set(
+				'/home/testuser/.neokai/projects/-test-repo/worktrees/session-123',
+				false
+			);
+			mockGitRevparse.mockResolvedValue('.git');
+			// checkBranchExists returns the stale branch; branch -D goes through mockGitBranch
+			mockGitRaw
+				.mockResolvedValueOnce('  custom-branch\n') // checkBranchExists — stale branch found
+				.mockResolvedValue(''); // worktree add (branch -D uses mockGitBranch, not mockGitRaw)
+
+			const result = await manager.createWorktree({
+				sessionId: 'session-123',
+				repoPath: '/test/repo',
+				branchName: 'custom-branch',
+			});
+
+			// Should reuse the original branch name, not fall back to UUID
+			expect(result?.branch).toBe('custom-branch');
+			// git branch -D goes through mockGitBranch
+			expect(mockGitBranch).toHaveBeenCalledWith(['-D', 'custom-branch']);
+		});
+
+		it('should delete stale auto-generated branch and reuse original name', async () => {
 			existsSyncResults.set('/test/repo/.git', true);
 			existsSyncResults.set('/home/testuser/.neokai/projects/-test-repo/worktrees', true);
 			existsSyncResults.set(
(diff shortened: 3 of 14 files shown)


lsm (commented):

The core data-model changes are correct and well-designed — the constant is in the right place, the virtual node is properly injected on deserialization and stripped on serialization, and the layout pinning logic is solid. However there are two bugs in `VisualWorkflowEditor.tsx` that cause incorrect behavior in the create-from-scratch flow, and a few secondary issues worth addressing.

---

### P0 — `addStep`: `isFirstNode` check counts the Task Agent as a node

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` line 283

```ts
const isFirstNode = nodes.length === 0;
```

In create mode (no `workflow` prop) `nodes` starts empty, so adding the first step correctly sets it as start. **But** if this component is ever initialized with a non-null `initState` that has the Task Agent node already injected (edit mode path), then `nodes` starts with `[taskAgentNode]`, so `nodes.length === 0` is `false` and `setStartStepId` is never called for the first user-added step. The start node would be left as whatever `initState.startNodeId` was.

In the current edit-mode path this is masked because `initState.startNodeId` is already populated from the loaded workflow. But any future code path that starts with nodes (e.g. loading a workflow that has zero regular steps, only the virtual node) and then calls `addStep` will silently skip setting the start node ID.

The fix is to check `nodes.filter(n => n.step.id !== TASK_AGENT_NODE_ID).length === 0` before the setter call, matching the pattern already used in save-validation at line 547.

---

### P0 — `handleDeleteNode`: `remaining` includes the Task Agent; start-node handoff goes to Task Agent

**File:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` lines 310, 318–320

```ts
const remaining = nodes.filter((n) => n.step.localId !== localId);
// ...
if (wasStart && remaining.length > 0) {
    const next = remaining[0];          // could be the Task Agent
    setStartStepId(next.step.id ?? next.step.localId);
}
```

When the start node (e.g. `step-1`) is deleted and the Task Agent virtual node is first in `nodes` (it is always injected at index 0), `remaining[0]` is the Task Agent node. `setStartStepId` is then called with `TASK_AGENT_NODE_ID` (`'__task_agent__'`). On the next save, `buildWorkflowFields` tries to resolve `startNodeId = '__task_agent__'` against `nodeMap` (which only contains persisted nodes); it falls through to the `persistableNodes[0]` fallback, so the serialized `startNodeId` ends up correct by accident. However the in-editor validation at line 547 checks `regularNodes` but uses the raw `startNodeId` state for `startNodeId: startKey` in the `VisualEditorState` — the visual START badge is rendered on the Task Agent node rather than a real step, which is visually wrong and can confuse the user.

The fix: filter out the Task Agent from `remaining` before picking the next start node.

```ts
const regularRemaining = remaining.filter(n => n.step.id !== TASK_AGENT_NODE_ID);
if (wasStart && regularRemaining.length > 0) {
    const next = regularRemaining[0];
    setStartStepId(next.step.id ?? next.step.localId);
} else if (wasStart) {
    setStartStepId('');
}
```

---

### P1 — `nodes.length === 0` empty-state overlay and template picker are always false when Task Agent is injected in edit mode

**File:** `VisualWorkflowEditor.tsx` lines 701 and 753

```tsx
{!isEditing && nodes.length === 0 && (   // line 701
{nodes.length === 0 && (                 // line 753 — empty state overlay
```

These conditions will never be true in edit mode (Task Agent is injected, so `nodes.length >= 1` always). That is actually correct for edit mode but worth a comment. More importantly, in create mode (before any user step is added) `nodes` is empty so both work. The conditions are functionally OK in the current flow, but they are fragile — if the implementation ever pre-populates `nodes` with the Task Agent on create mode too, the empty-state overlay would silently break. These should be:

```ts
nodes.filter(n => n.step.id !== TASK_AGENT_NODE_ID).length === 0
```

This is the same pattern already used correctly in save validation.

---

### P1 — Serialization strips Task Agent by checking both `step.id` and `step.localId`, but save-validation only checks `step.id`

**File:** `packages/web/src/components/space/visual-editor/serialization.ts` line 272–274 vs `VisualWorkflowEditor.tsx` line 547

Serialization correctly does:
```ts
n.step.id !== TASK_AGENT_NODE_ID && n.step.localId !== TASK_AGENT_NODE_ID
```

But save-validation does:
```ts
nodes.filter((n) => n.step.id !== TASK_AGENT_NODE_ID)
```

Since `step.localId` is also set to `TASK_AGENT_NODE_ID` in `workflowToVisualState` (line 140), both checks are equivalent in practice today. However the inconsistency is a latent bug risk if `step.localId` ever diverges from `step.id` for the virtual node. The validation should use the same dual-check as serialization, or both should rely on the `virtual: true` flag from `TaskAgentVisualNode`.

---

### P2 — `TaskAgentVisualNode` interface defined in `types.ts` but the actual virtual node is constructed as a plain `VisualNode` in `serialization.ts`

**File:** `packages/web/src/components/space/visual-editor/types.ts` vs `serialization.ts` lines 138–147

`TaskAgentVisualNode` is declared as a separate interface with `virtual: true`, but the task agent node injected in `workflowToVisualState` is typed as `VisualNode` (same shape as regular nodes) — the `virtual` flag is never set and `TaskAgentVisualNode` is never used anywhere in the codebase (confirmed by grep). The type is currently dead code. Either:
- Remove `TaskAgentVisualNode` and distinguish the virtual node at runtime by its ID only (current approach), or
- Use it as a discriminated union so TypeScript can enforce the virtual/non-virtual distinction at the type level.

---

### P2 — Missing test: deleting the start node should not set Task Agent as start

There is no test in `VisualWorkflowEditor.test.tsx` that verifies the `handleDeleteNode` start-node handoff skips the Task Agent. The existing test at line 341 verifies a non-start node is deleted, but doesn't cover the case where `wasStart === true` and the Task Agent is `remaining[0]`.

---

### P3 — `computeTaskAgentPosition` minimum-y result can be zero when nodes are at `y=0`

**File:** `packages/web/src/components/space/visual-editor/serialization.ts` line 209

```ts
const y = minY === Infinity ? 0 : Math.max(0, minY - TASK_AGENT_Y_OFFSET);
```

If `minY` is less than `TASK_AGENT_Y_OFFSET` (120), the result is clamped to 0, placing the Task Agent at the exact same y as or above nodes that start near the top. The layout path handles this correctly (uses a dedicated `TASK_AGENT_Y = 20` constant), but the serialization path can produce y=0. This is a minor cosmetic issue since the auto-layout path pins it above anyway, but stored layouts with nodes near y=0 will overlap with the Task Agent.

lsm (commented):

---

The fix commit addresses all of the originally reported issues correctly. However, the review uncovered one pre-existing P0 bug that was not fixed and one pre-existing P1 bug that the fix commit's new test inadvertently masks.

---

### P0 — Task Agent is absent in create mode and after template application

The Task Agent virtual node is only injected by `workflowToVisualState()`, which is called exclusively when loading an existing workflow. In two cases the Task Agent is never present:

**Case 1 — New workflow (create mode):**
`initState` is `null` when no `workflow` prop is passed (line 84 of `VisualWorkflowEditor.tsx`), so `nodes` initialises as `[]`. The Task Agent is never inserted. Every call to `addStep` appends only the new step. This contradicts the design goal ("Task Agent node is always present in the visual editor").

**Case 2 — Template application:**
`applyTemplate` (line 514) calls `setNodes(positionedNodes)`, which replaces the entire nodes array with only the template nodes, evicting the Task Agent even in edit mode where it was previously present.

The existing tests actively confirm the bug. `packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx` line 803:
```ts
// After clicking the Coding (Plan → Code) template:
expect(getAllByTestId(/^workflow-node-/).length).toBe(2);  // should be 3
```
And line 245:
```ts
// After clicking Add Step in create mode:
expect(getAllByTestId(/^workflow-node-/).length).toBe(1);  // should be 2 (Task Agent + new step)
```
These assertions pass only because the Task Agent is missing. They silently mask the bug.

**Fix required:** Inject the Task Agent on component initialisation (when `workflow` is undefined) and re-inject it inside `applyTemplate` after calling `setNodes`. Tests for node counts after `addStep` and template application must be updated to include the Task Agent in the expected counts.

---

### P1 — New test does not exercise the original P0 scenario

The new test at line 370 is titled "Task Agent never receives the start badge after any node deletion". The test comments acknowledge: "The UI disables deletion of the current start node, so this test verifies the invariant via the reachable 'delete non-start' path."

The original P0 scenario was: deleting the **start** node promotes `remaining[0]` (which is the Task Agent) as the new start. The fix in `handleDeleteNode` filters `regularRemaining` before promoting — this fix is correct. But the new test never reaches this branch. The test moves the start badge to step-2 first, then deletes step-1 (now a non-start node), so `wasStart` is `false` in that invocation and the `regularRemaining` filter is never evaluated.

The code fix is correct; the test does not prove it. A complementary unit test should exercise the `wasStart = true` path with the Task Agent present in `remaining[0]`.

---

### P2 — Inconsistent Task Agent filter across the editor

The save-validation filter (line 553–555) uses a dual-check:
```ts
(n) => n.step.id !== TASK_AGENT_NODE_ID && n.step.localId !== TASK_AGENT_NODE_ID
```

But the three guards added by this fix commit use only the single `id` check:
- `isFirstNode` (line 284)
- `handleDeleteNode` (line 322)
- Template picker visibility (line 709)
- Empty state overlay (line 761)

Since the Task Agent node is constructed with `id: TASK_AGENT_NODE_ID` and `localId: TASK_AGENT_NODE_ID` (serialization.ts lines 140–141), both checks are equivalent today. But the inconsistency is notable since the fix commit's own message said "use consistent dual-check for Task Agent detection in all guards" — that goal was only partially achieved.

---

### P3 — Position stagger in `addStep` includes Task Agent in count

`addStep` computes new-node positions based on `prev.length * 100` (line 287). When the Task Agent is present (edit mode), `prev.length` is 1 before the first real step is added, so the first real step lands at `y = 180` instead of `y = 80`. Minor visual offset, not a functional bug.

---

### Fixes verified correctly

All six originally reported issues were addressed in the fix commit:
- P0: `isFirstNode` now filters Task Agent (line 284). ✓
- P0: `handleDeleteNode` uses `regularRemaining` to skip Task Agent (line 322). ✓
- P1: Save-validation dual-check is consistent (lines 553–555). ✓
- P1: Empty-state UI checks now exclude Task Agent (lines 709, 761). ✓
- P2: Dead `TaskAgentVisualNode` interface removed from `types.ts`. ✓
- P3: `TASK_AGENT_MIN_Y = 20` constant replaces `Math.max(0,...)` in `computeTaskAgentPosition`. ✓

lsm (commented):

**Recommendation: REQUEST_CHANGES** — 1 P0, 1 P1 (test comment inaccuracy), 1 P2, 1 P3.

---

### P0 — Task Agent virtual node can be deleted by the user

The core invariant of this PR — Task Agent is always present — is not enforced at the deletion boundary. Two distinct user paths delete it:

**Path 1: NodeConfigPanel** — Clicking the Task Agent node on the canvas calls `handleNodeSelect(TASK_AGENT_NODE_ID)` in `WorkflowCanvas`, which calls `onNodeSelect` → `setSelectedNodeId(TASK_AGENT_NODE_ID)` in `VisualWorkflowEditor`. `selectedNode` then resolves to the Task Agent node and `NodeConfigPanel` opens. Inside `NodeConfigPanel`, `isStartNode` is false (Task Agent is never the start node), so the Delete Step button is enabled (`disabled={isStartNode}`). The user can click "Delete Step" → confirm → `handleDeleteNode(TASK_AGENT_NODE_ID)` is called → Task Agent is removed from `nodes`.

**Path 2: Keyboard Delete** — Selecting the Task Agent node (click) and then pressing Delete/Backspace fires the keyboard handler in `WorkflowCanvas` at line 416:
```ts
onDeleteNodeRef.current(current); // current = TASK_AGENT_NODE_ID
```
There is no guard against `TASK_AGENT_NODE_ID` in `handleDeleteNode`. After deletion, the node array no longer contains the Task Agent and subsequent calls to `setNodes` (e.g., `addStep`) won't re-inject it.

**Files:** `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` (line 327–358), `packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx` (line 405–423)

**Fix required:** Guard `handleDeleteNode` against `TASK_AGENT_NODE_ID` (early return), and prevent `handleNodeSelect` from selecting the Task Agent node. The `WorkflowNode` component for Task Agent should ideally suppress selection/deletion entirely, or the `VisualWorkflowEditor` should filter it.

---

### P1 — Test comment incorrectly claims `wasStart=true` branch is UI-unreachable

The comment at lines 379–383 of `VisualWorkflowEditor.test.tsx` states:

> The `wasStart = true` branch in handleDeleteNode is intentionally unreachable via UI: the delete button is disabled for the current start node, and the handler has an early-return guard.

This is wrong. The keyboard Delete path in `WorkflowCanvas` (lines 405–423) does NOT check whether the selected node is the start node before calling `onDeleteNode`. A user can:
1. Click the start node to select it (WorkflowCanvas's own `selectedNodeId` is set)
2. Press Delete/Backspace
3. `handleDeleteNode(startNodeLocalId)` is called with `wasStart=true`

The `regularRemaining` filter in the `wasStart=true` branch IS reachable and is correct code. The test comment is misleading: it describes an already-fixed problem (button disabled in NodeConfigPanel) but misses the keyboard path. A test that exercises this path directly (simulate selecting start node → keyboard Delete → verify another regular node becomes start, Task Agent does not) would be more accurate.

---

### P2 — Hardcoded position constants duplicated in create-mode initializer

In `VisualWorkflowEditor.tsx` line 108:
```ts
position: { x: 50, y: 20 },
```
The comment says this matches `START_X=50, TASK_AGENT_Y=20` from `layout.ts`, but those constants are module-private. If someone updates them (e.g., changes `TASK_AGENT_Y` from 20 to 40), the create-mode initializer silently drifts. The constants should be exported from `layout.ts` and imported here, or the initializer should call `autoLayout([],[],'')​.get(TASK_AGENT_NODE_ID)` directly to stay in sync.

---

### P3 — Stale `START_Y=50` comment in test

`packages/web/src/components/space/visual-editor/__tests__/VisualWorkflowEditor.test.tsx` line 884:
```
// autoLayout places the first node at START_X=50, START_Y=50, so at
```
The actual `START_Y` in `layout.ts` is `170` (bumped in this PR to leave room for the Task Agent node). The comment predates the bump and should read `START_Y=170`.

---

### What Was Done Well

- The P0 fixes from the previous review (create-mode injection via `useState` initializer and re-injection after `applyTemplate`) are correct and well-reasoned.
- The dual-check (`step.id !== TASK_AGENT_NODE_ID && step.localId !== TASK_AGENT_NODE_ID`) is consistently applied in all guards.
- The position stagger fix in `addStep` correctly excludes the Task Agent from the count.
- Serialization tests are comprehensive: Task Agent is verified absent from `layout`, `nodes`, and `transitions` in all create/update paths.
- The create-mode invariant test (lines 420–460) is a useful addition and exercises the correct path.

lsm (commented):

**Recommendation: APPROVE** (own PR — using COMMENT event)

This is the fourth review iteration on PR #774 (Add Task Agent virtual node to visual editor data model). All four previously-raised issues (P0 through P3) have been correctly addressed. One new minor observation is noted below.

---

### Previously-raised issues — verification

**P0 (fixed): Task Agent deletion guards**

`handleDeleteNode` in `VisualWorkflowEditor.tsx` now has an explicit guard at the top:
```ts
if (localId === TASK_AGENT_NODE_ID) return;
```
`handleNodeSelect` also guards early with the same check. Both the NodeConfigPanel delete-button path and the keyboard Delete path are correctly protected.

**P1 (fixed): Test comment and keyboard-Delete test**

The stale comment claiming the `wasStart=true` branch was "UI-unreachable" has been replaced with an accurate note pointing to the new keyboard-Delete test. The new test at line 465 in `VisualWorkflowEditor.test.tsx` clicks the start node, fires `keyDown(document.body, { key: 'Delete' })` (which bubbles to the `window` listener), and verifies that the remaining regular node is promoted to start while the Task Agent never receives the badge. This correctly exercises the `regularRemaining` filter in the `wasStart=true` branch.

**P2 (fixed): Hardcoded position constants removed**

`TASK_AGENT_INITIAL_POSITION` is now exported from `layout.ts` (line 36) as `{ x: START_X, y: TASK_AGENT_Y }` and imported in `VisualWorkflowEditor.tsx`. The two previous hardcoded `{ x: 50, y: 20 }` literals are gone.

**P3 (fixed): Stale START_Y=50 comment**

The comment in `VisualWorkflowEditor.test.tsx` (around line 923) was correctly updated from `START_Y=50` to `START_Y=170`.

---

### New observations

**P2 — Task Agent can appear visually selected (blue ring) in the canvas**

`WorkflowCanvas.handleNodeSelect` (line 334 of `WorkflowCanvas.tsx`) unconditionally calls `setSelectedNodeId(stepId)` for any node click, including `TASK_AGENT_NODE_ID`. The parent's guard returns early and prevents the `NodeConfigPanel` from opening and prevents deletion, but `WorkflowCanvas`'s internal `selectedNodeId` state is still set to `TASK_AGENT_NODE_ID`. As a result, clicking the Task Agent renders it with a blue selection ring (`ring-2 ring-blue-500`) and a blue border while no panel opens — a confusing affordance. The guard comment at line 323 says the node "must not be selectable" but selection is only half-blocked. Adding a guard in `WorkflowCanvas.handleNodeSelect` (skip if `stepId === TASK_AGENT_NODE_ID`) or passing an `isInteractive` prop to the Task Agent's `WorkflowNode` would fully align the implementation with the stated intent.

**P3 — Stale "(50, 50)" value in serialization.test.ts comment**

`packages/web/src/components/space/visual-editor/__tests__/serialization.test.ts`, line 141:
```
// If autoLayout ran, it would produce different values (50, 50) not (999, 888).
```
With `START_X=50` and `START_Y=170`, autoLayout would now place the first node at `{x:50, y:170}`, not `{x:50, y:50}`. The comment is functionally harmless (the test still passes because `(999, 888) !== (50, 170)`) but is factually wrong. A minor follow-up fix.

lsm (commented):

**Recommendation: APPROVE**

---

Both issues from the previous review round have been correctly addressed. This is a clean, well-structured implementation that I am satisfied with after five iterations.

### Fix 1 (P2) — WorkflowCanvas guard on handleNodeSelect [VERIFIED]

`packages/web/src/components/space/visual-editor/WorkflowCanvas.tsx` lines 335–337:

The guard is correctly placed as the first statement in the `handleNodeSelect` callback. When `stepId === TASK_AGENT_NODE_ID`, the function returns immediately, preventing `setSelectedNodeId` from being called, which means:
- No blue selection ring is rendered on the Task Agent node (since `isSelected={selectedNodeId === stepId}` is only true for real nodes)
- The keyboard Delete path cannot fire for the Task Agent (since the `selectedNodeIdRef` is never set to `TASK_AGENT_NODE_ID`)

The defense-in-depth design is sound: both `WorkflowCanvas.handleNodeSelect` and `VisualWorkflowEditor.handleNodeSelect` independently guard against `TASK_AGENT_NODE_ID`, so neither layer alone is a single point of failure.

### Fix 2 (P3) — Stale comment in serialization.test.ts [VERIFIED]

`packages/web/src/components/space/visual-editor/__tests__/serialization.test.ts` line 141:

The comment now correctly reads `(50, 170)` matching `START_Y=170` in `layout.ts`. The coordinate reflects the first regular node's y-position in a single-node auto-layout, not `(50, 50)` from the old `START_Y=50` era.

---

### Broader code quality

The overall implementation is correct and complete:

- `buildWorkflowFields()` uses a consistent dual-check (`step.id !== TASK_AGENT_NODE_ID && step.localId !== TASK_AGENT_NODE_ID`) everywhere the virtual node must be stripped — serialization, validation, position-stagger, and start-node handoff.
- `applyTemplate()` re-injects the Task Agent node after replacing the nodes array, preventing eviction on template application.
- Create-mode state initializer injects the Task Agent immediately using the exported `TASK_AGENT_INITIAL_POSITION` constant, keeping the position in sync with `autoLayout`.
- Test coverage is thorough: keyboard Delete path, create-mode invariant, start-badge invariant, wasStart=true path, and save-round-trip tests all explicitly account for the Task Agent virtual node.

No new issues found. This PR is ready to merge.

PR #777, opened 2026-03-24
feat: extend WorkflowNodeAgent schema with role, model, systemPrompt fields

lsm:
Add per-slot override fields to WorkflowNodeAgent and ExportedWorkflowNodeAgent:
- role: string (required, unique within node, used for channel routing)
- model?: string (override agent's default model for this slot)
- systemPrompt?: string (override agent's default system prompt for this slot)

Key changes:
- space.ts: WorkflowNodeAgent and ExportedWorkflowNodeAgent get new fields
- space-utils.ts: resolveNodeChannels now uses WorkflowNodeAgent.role directly
  (no longer looks up SpaceAgent.role); validateNodeChannels validates slot role
  uniqueness; resolveNodeAgents uses agentId as synthetic role for legacy shorthand
- export-format.ts: Zod schema and export function include new fields
- space-export-import-handlers.ts: import side preserves role, model, systemPrompt
- All downstream tests updated with required role field on WorkflowNodeAgent objects

Same agentId can now appear multiple times in a node if each slot has a distinct role.

### DIFF
--- a/packages/daemon/.env.example
+++ b/packages/daemon/.env.example
@@ -58,6 +58,27 @@ TEMPERATURE=1.0
 MAX_SESSIONS=10
 
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+# SDK Startup Tuning
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+#
+# How long (ms) to wait for the SDK subprocess to emit its first message
+# before declaring a startup timeout. Increase on slow machines or when
+# running many concurrent sessions.
+# Default: 15000
+# NEOKAI_SDK_STARTUP_TIMEOUT_MS=15000
+#
+# Base delay (ms) before the first auto-recovery attempt after a startup
+# timeout. Each subsequent attempt doubles this (exponential backoff).
+# Example: 3000 → attempt 1 waits 3s, attempt 2 waits 6s.
+# Default: 3000
+# NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS=3000
+#
+# Maximum number of auto-recovery attempts before surfacing the error to
+# the user. Set to 0 to disable auto-recovery entirely.
+# Default: 2
+# NEOKAI_SDK_STARTUP_MAX_RETRIES=2
+
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 # Job Queue Configuration
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 #
--- a/packages/daemon/src/lib/agent/query-runner.ts
+++ b/packages/daemon/src/lib/agent/query-runner.ts
@@ -25,6 +25,10 @@ import type { QueryOptionsBuilder } from './query-options-builder';
 import type { AskUserQuestionHandler } from './ask-user-question-handler';
 
 const DEFAULT_STARTUP_TIMEOUT_MS = 15000;
+/** Base delay before the first auto-recovery attempt. Each subsequent attempt doubles this. */
+const DEFAULT_STARTUP_RECOVERY_DELAY_MS = 3000;
+/** Maximum number of auto-recovery attempts before surfacing the error to the user. */
+const DEFAULT_STARTUP_MAX_RETRIES = 2;
 
 function getStartupTimeoutMs(): number {
 	const raw = process.env.NEOKAI_SDK_STARTUP_TIMEOUT_MS;
@@ -33,7 +37,26 @@ function getStartupTimeoutMs(): number {
 	return Number.isFinite(parsed) && parsed > 0 ? parsed : DEFAULT_STARTUP_TIMEOUT_MS;
 }
 
+function getStartupRecoveryDelayMs(): number {
+	const raw = process.env.NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS;
+	if (!raw) return DEFAULT_STARTUP_RECOVERY_DELAY_MS;
+	const parsed = Number.parseInt(raw, 10);
+	return Number.isFinite(parsed) && parsed > 0 ? parsed : DEFAULT_STARTUP_RECOVERY_DELAY_MS;
+}
+
+function getStartupMaxRetries(): number {
+	const raw = process.env.NEOKAI_SDK_STARTUP_MAX_RETRIES;
+	if (!raw) return DEFAULT_STARTUP_MAX_RETRIES;
+	const parsed = Number.parseInt(raw, 10);
+	return Number.isFinite(parsed) && parsed >= 0 ? parsed : DEFAULT_STARTUP_MAX_RETRIES;
+}
+
+// Read once at module load — consistent with the original STARTUP_TIMEOUT_MS pattern.
+// Env vars set after the process starts will not be picked up; the values displayed
+// in user-facing error messages reflect these module-load-time snapshots.
 const STARTUP_TIMEOUT_MS = getStartupTimeoutMs();
+const STARTUP_RECOVERY_DELAY_MS = getStartupRecoveryDelayMs();
+const STARTUP_MAX_RETRIES = getStartupMaxRetries();
 
 /**
  * Original environment variables for restoration after SDK query
@@ -237,10 +260,18 @@ export class QueryRunner {
 				if (!this.ctx.firstMessageReceived) {
 					startupTimeoutReached = true;
 					const elapsed = Date.now() - queryStartTime;
+					const isRootWorkspace = !session.worktree;
+					const workspaceDesc = isRootWorkspace
+						? `root workspace: ${session.workspacePath}`
+						: `worktree: ${session.worktree!.worktreePath}`;
 					logger.error(
 						`SDK startup timeout: SDK did not respond within ${elapsed}ms. ` +
-							`Model: ${queryOptions.model}, Workspace: ${session.workspacePath}` +
-							(session.worktree ? ` (worktree: ${session.worktree.worktreePath})` : '')
+							`Model: ${queryOptions.model}, ${workspaceDesc}` +
+							(isRootWorkspace
+								? ' — running on root workspace (not a worktree); check for other Claude Code sessions using this path'
+								: '') +
+							`. Attempt ${this.ctx.startupTimeoutAutoRecoverAttempts + 1} of ${STARTUP_MAX_RETRIES + 1}.` +
+							` (Hint: set NEOKAI_SDK_STARTUP_TIMEOUT_MS to increase timeout, currently ${STARTUP_TIMEOUT_MS}ms)`
 					);
 
 					// Actively abort a stuck startup so finally{} cleanup runs and the
@@ -348,29 +379,39 @@ export class QueryRunner {
 
 			if (!isAbortError) {
 				// On startup timeout or conversation not found, attempt transparent auto-recovery
-				// (up to 1 retry): restart the query without the old resume handle
+				// (up to STARTUP_MAX_RETRIES): restart the query without the old resume handle
 				// (sdkSessionId already cleared above). Queued messages are preserved so
 				// the user's pending send is retried automatically without any visible error.
-				// The retry limit (attempts <= 1) prevents infinite loops when the SDK is
-				// permanently broken: after 1 failed recovery, the error surfaces normally.
+				// Uses exponential backoff: delay doubles with each attempt to give the SDK
+				// process time to fully exit before the next spawn.
+				// The retry limit prevents infinite loops when the SDK is permanently broken:
+				// after all retries are exhausted, the error surfaces normally.
 				const startupRecoverAttempts = this.ctx.startupTimeoutAutoRecoverAttempts + 1;
 				const canAutoRecover =
 					(isStartupTimeout || isConversationNotFound) &&
 					!this.ctx.isCleaningUp() &&
 					!!this.ctx.onStartupTimeoutAutoRecover &&
-					startupRecoverAttempts <= 1;
+					startupRecoverAttempts <= STARTUP_MAX_RETRIES;
 
 				if (canAutoRecover) {
 					this.ctx.startupTimeoutAutoRecoverAttempts = startupRecoverAttempts;
+					// Exponential backoff: base * 2^(attempt-1), capped at 30s.
+					// e.g. with default 3s base: attempt 1 → 3s, attempt 2 → 6s
+					const delayMs = Math.min(
+						STARTUP_RECOVERY_DELAY_MS * Math.pow(2, startupRecoverAttempts - 1),
+						30000
+					);
 					logger.warn(
-						`SDK ${isConversationNotFound ? 'conversation not found' : 'startup timeout'} — scheduling auto-recovery attempt ${startupRecoverAttempts} (fresh query without resume)`
+						`SDK ${isConversationNotFound ? 'conversation not found' : 'startup timeout'} — ` +
+							`scheduling auto-recovery attempt ${startupRecoverAttempts}/${STARTUP_MAX_RETRIES} ` +
+							`(fresh query without resume) in ${delayMs}ms`
 					);
 					// Defer until after finally{} completes so shared state is reset first.
 					setTimeout(() => {
 						this.ctx.onStartupTimeoutAutoRecover!().catch((err: unknown) => {
 							logger.error('Auto-recovery after SDK startup timeout failed:', err);
 						});
-					}, 300);
+					}, delayMs);
 					// setIdle is handled by the finally block; skipping here avoids a double call.
 				} else {
 					// Reset counter so a future successfully-started session can recover again.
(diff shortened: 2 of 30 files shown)


lsm (commented):

The schema extension is structurally sound, the core type changes are correct, and the test coverage is broad. However, there are several bugs and gaps that need to be addressed before merging.

---

## P0 — Server-side channel validation uses wrong role source

**File:** `packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, line 226

```ts
knownRoles.add(agent.role); // <-- uses SpaceAgent.role, not WorkflowNodeAgent.role
```

The `validateNodeAgentRef` method still builds `knownRoles` from `agent.role` (the `SpaceAgent`'s role, resolved via `agentLookup`), rather than from the new per-slot `entry.role` field on each `WorkflowNodeAgent`. This means server-side channel validation at create/update time will reject valid channel references that use the new per-slot role, and will accept invalid ones when the per-slot role differs from the SpaceAgent's role.

The runtime (in `space-utils.ts:resolveNodeChannels`) now uses `WorkflowNodeAgent.role` directly. These two behaviours are out of sync, so a workflow that passes runtime resolution will fail server validation (or vice versa).

**Fix:** Change line 226 to `knownRoles.add(entry.role)` instead of `knownRoles.add(agent.role)`.

---

## P0 — Old DB rows with `agents` but no `role` field silently break

**File:** `packages/daemon/src/storage/repositories/space-workflow-repository.ts`, lines 115–116

```ts
if (cfg.agents && cfg.agents.length > 0) {
    node.agents = cfg.agents; // no backfill of role
}
```

The repository passes raw JSON directly into `WorkflowNodeAgent[]`. Existing rows written before this PR will have `agents` entries without a `role` field. When `resolveNodeChannels` (or `validateNodeChannels`) subsequently calls `roleToAgentId.set(sa.role, sa.agentId)`, it will insert `undefined` as the key. Any downstream lookup will silently miss, causing channels to be dropped without error.

A read-time backfill is needed: when deserializing, if `entry.role` is absent or empty, fall back to `entry.agentId` as the synthetic role (matching the behaviour of the `agentId`-shorthand path in `resolveNodeAgents`).

---

## P0 — `model` and `systemPrompt` override fields are schema-only dead code

**Files:** `packages/shared/src/types/space.ts` (lines 583–585), `packages/daemon/src/lib/space/runtime/workflow-executor.ts` (lines 460–471)

The PR adds `model` and `systemPrompt` to `WorkflowNodeAgent` and `ExportedWorkflowNodeAgent`, and the export/import pipeline faithfully round-trips them. But `workflow-executor.ts` never reads `agentEntry.model` or `agentEntry.systemPrompt` when creating tasks, and `task-agent.ts` uses `space.defaultModel` unconditionally. The fields are accepted, stored, exported, imported — and then ignored at runtime.

If these fields are intentionally deferred to a later PR, that should be documented prominently (e.g., a TODO comment) and ideally gated so the Zod schema rejects them for now (or at minimum the type docstring should say they are not yet respected at runtime). As written, a user setting these overrides will get silently wrong behaviour.

---

## P1 — Format-level validation does not check that `role` is non-empty

**File:** `packages/daemon/src/lib/space/managers/space-workflow-manager.ts`, lines 193–201

The loop validates that `agentId` is a non-empty string but does not validate the new required `role` field:

```ts
for (let j = 0; j < node.agents!.length; j++) {
    const entry = node.agents![j];
    if (!entry.agentId || !entry.agentId.trim()) { ... }
    // no check for entry.role
}
```

An empty or missing `role` will silently be treated as `''` (the string `undefined` coerces to the key `'undefined'` in the Map), leading to channel resolution silently failing. A format-level check similar to the `agentId` check should be added.

---

## P1 — UI still prevents duplicate `agentId` additions (contradicts stated goal)

**Files:** `packages/web/src/components/space/WorkflowNodeCard.tsx` (line 137), `packages/web/src/components/space/visual-editor/NodeConfigPanel.tsx` (line 82)

Both `addAgent` functions contain:

```ts
if (nodeAgents.some((a) => a.agentId === agentId)) return;
```

The PR description states that "same agentId can now appear multiple times in a node if each slot has a distinct role", but the UI still blocks adding the same agent twice. The backend and `space-utils.ts` support this; the UI does not. A user cannot take advantage of the new capability through the visual editor.

---

## P1 — Stale docstrings referencing `SpaceAgent.role`

**Files:**
- `packages/shared/src/types/space.ts`, line 593 (WorkflowChannel comment)
- `packages/shared/src/types/space.ts`, line 656-658 (WorkflowNode.channels comment)
- `packages/shared/src/types/space.ts`, line 708-710 (WorkflowNodeInput.channels comment)
- `packages/shared/src/types/space-utils.ts`, line 24 (ResolvedChannel.fromRole comment)

Multiple docstrings still say `from`/`to` reference `SpaceAgent.role`. After this PR, they reference the per-slot `WorkflowNodeAgent.role`. These are misleading to future contributors.

---

## P2 — `agents` parameter in `resolveNodeChannels` is now permanently dead

**File:** `packages/shared/src/types/space-utils.ts`, lines 119–123

```ts
export function resolveNodeChannels(node: WorkflowNode, agents: SpaceAgent[]): ResolvedChannel[] {
    // agents param is retained for API compatibility; resolution now uses WorkflowNodeAgent.role directly.
    void agents;
```

The comment acknowledges this is dead code. Callers (`space-runtime.ts` line 1016) still pass `allAgents`. Since the `agents` parameter is now truly unused, the signature should either be deprecated with a migration plan, or dropped entirely (with callers updated to pass nothing). Keeping it with `void agents` encourages callers to think the value matters.

---

## P3 — Error message says "Duplicate roles" but test checks for that exact string

**File:** `packages/shared/tests/space-utils.test.ts`, line 483

```ts
expect(errors.some((e) => e.includes('Duplicate roles'))).toBe(true);
```

The test is brittle because it relies on a substring match of a human-readable message. Consider using a typed error code or a more specific assertion.

lsm (commented):

Overall the core schema additions are well-designed and the PR is largely correct, but there are several concrete issues that need attention.

---

### P0 — TypeScript type violations in existing tests

Multiple tests go directly to the repository layer (bypassing manager validation) and construct `WorkflowNodeAgent` objects **without the now-required `role` field**. Since `WorkflowNodeAgent.role` changed from optional to required, these are compile-time type errors that will surface when TypeScript is run.

Affected files and lines:

- `packages/daemon/tests/unit/space/space-workflow.test.ts` lines 336, 368-369, 404:
  ```ts
  agents: [{ agentId: 'agent-multi-1' }, { agentId: 'agent-multi-2' }]
  ```
- `packages/daemon/tests/unit/space/workflow-executor-multi-agent.test.ts` lines 154, 190, 231-232:
  ```ts
  agents: [{ agentId: AGENT_A }, { agentId: AGENT_B }]
  ```

These tests persist objects with no `role` into the repository, which is now an invalid shape. The backfill in `rowToNode` will rescue reads, but the TypeScript compiler should be rejecting these call sites. They need to be updated with explicit `role` values.

---

### P1 — Role uniqueness not validated in `SpaceWorkflowManager.validateNodeAgentRef`

`validateNodeAgentRef` (lines 193-206 in `space-workflow-manager.ts`) validates that each agent entry has a non-empty role, but it does **not** detect duplicate roles within the same node's `agents[]` array. The `knownRoles` set at line 231 collects roles only when `agentLookup` is present — without a lookup, two agents with the same role silently pass validation and get persisted.

`validateNodeChannels` in `space-utils.ts` does detect duplicate roles, but that path is not exercised unless `channels[]` is also present. A node with two agents sharing the same role but no channels will bypass all duplicate-role checks at the manager level.

There are also no test cases in the manager tests verifying that duplicate roles in `agents[]` are rejected.

---

### P1 — `removeAgent` and `updateAgentInstructions` in `NodeConfigPanel.tsx` are broken for duplicate-agentId nodes

`removeAgent` at line 95 filters by `a.agentId !== agentId`. Since the PR now allows the same `agentId` to appear multiple times with different roles, this will remove **all** slots for that agent instead of just the intended one.

Similarly, `updateAgentInstructions` at line 107-109 matches on `a.agentId === agentId` and will update **all** matching slots, not just the one the user edited.

The `usedIds` filter at line 113 also prevents users from adding the same agent a second time with a different role, which contradicts the feature goal.

The correct key for these operations should be `role` (which is required and unique within a node), not `agentId`.

---

### P1 — Missing test coverage for the new `role` field in export/round-trip tests

The round-trip test in `export-format.test.ts` (lines 1543-1571, "multi-agent node round-trip preserves agents array and channels") validates `agentRef` and `instructions` from the `agents[]` array, but never asserts that `role` is preserved. Since `role` is the primary new field added by this PR, this is a critical test gap.

Additionally:
- No test in `export-format.test.ts` validates that Zod correctly rejects an `agents[]` entry with an empty or missing `role`.
- No test in `space-workflow.test.ts` covers the new empty-role rejection (`role must be a non-empty string`).
- No test covers the read-time backfill in `space-workflow-repository.ts` (an old row without `role` should be read back with `role = agentId`).

---

### P2 — `role` export is tested implicitly but not explicitly

In `export-format.test.ts` the tests at line 1260-1267 verify that `agentRef` is correctly resolved, but there is no dedicated assertion that `agents[0].role === 'coder'` and `agents[1].role === 'reviewer'` are included in the export output. Given that `role` is the key routing identifier and the primary addition of this PR, this should be an explicit assertion.

---

### P3 — Deprecated `agents` parameter signature creates a misleading API surface

`resolveNodeChannels(node, agents: SpaceAgent[])` keeps the `agents` parameter in the signature (marked deprecated via a `void agents` discard) but callers still have to pass a value. The deprecation notice is only visible in the JSDoc. Consider removing it outright in this PR since the parameter was never truly public API — it was only called from internal test code. Leaving a deprecated parameter in a non-exported utility adds cognitive overhead with no benefit.

---

### Summary of counts

- P0: 1 (TypeScript type violations in test files — `role` field missing on `WorkflowNodeAgent`)
- P1: 3 (role uniqueness not validated without agentLookup; `removeAgent`/`updateAgentInstructions` broken for same-agentId multi-role nodes; missing test coverage for `role` in export round-trip)
- P2: 1 (role not asserted explicitly in export tests)
- P3: 1 (deprecated `agents` param)

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 bug remains in the UI layer.

---

### Summary

The backend changes are solid: schema additions are correct, format-level role uniqueness validation is properly placed in `validateNodeAgentRef`, the read-time backfill in `rowToNode` is correct, the `resolveNodeChannels` signature simplification is clean, and test coverage for all edge cases (empty role, duplicate role, same-agentId-different-roles, backfill) is present. The previously-reported P0/P1/P2 issues from the prior review are addressed.

However, a P1 functional regression was introduced in the UI: the "same agentId with different roles" feature that is advertised in the PR body and enabled by the backend/types cannot actually be exercised through the UI without hitting a server-side duplicate-role validation error.

---

### P1 — `addAgent` in both UI components always assigns the SpaceAgent.role as slot role, making same-agent re-add produce a duplicate-role error

**Files:**
- `packages/web/src/components/space/visual-editor/NodeConfigPanel.tsx` line 87
- `packages/web/src/components/space/WorkflowNodeCard.tsx` line 138

Both `addAgent` functions now do:
```ts
const role = agentInfo?.role ?? agentId;
const next = [...stepAgents, { agentId, role }];
```

The `agentInfo?.role` here is `SpaceAgent.role` (the agent's job label, e.g. `'coder'`). When the same agent is added a second time via the dropdown, both slots get `role = 'coder'` — a duplicate `WorkflowNodeAgent.role` — which the server now correctly rejects with a `WorkflowValidationError`. The PR removed the old guard (`if (nodeAgents.some((a) => a.agentId === agentId)) return;`) and added a comment saying "same agent may be added multiple times with different roles," but without a UI input for the slot-specific role, a user following the natural flow will produce an invalid state that silently fails to save.

This means the feature described in the PR body ("Same agentId can now appear multiple times in a node if each slot has a distinct role") is backend-complete but not usable through the UI. The UI either needs to:
1. Restore the original guard preventing duplicate `agentId` until a slot-role input is provided, **or**
2. Generate a unique slot role automatically when the same agent is added a second time (e.g., append a suffix: `coder-2`, `coder-3`), **or**
3. Add an inline text input to set the `WorkflowNodeAgent.role` per slot.

---

### P2 — Test name and comment are now incorrect after `availableAgents = agents`

**File:** `packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx` line 455

The test is named "shows add-agent-select dropdown for agents not yet in step" and has the inline comment "agent-2 is not in step yet, should appear in dropdown." Since `availableAgents = agents` now (all agents always appear), agent-1 is also present in the dropdown even though it is already in the step. The assertion still passes (`expect(select.textContent).toContain('Coder')`) but the test description describes old behavior that no longer holds.

---

### P3 — `WorkflowNodeAgent.role` should be documented to distinguish it from `SpaceAgent.role`

**File:** `packages/shared/src/types/space.ts` line 577–581

The JSDoc on `WorkflowNodeAgent.role` does not mention that this field is a per-slot identifier distinct from `SpaceAgent.role`. Since both are called `role` in their respective types and `addAgent` in the UI currently conflates them, a brief note (e.g., "This is a slot-specific label, distinct from `SpaceAgent.role` which identifies the agent's job category") would prevent future confusion.

---

### What is well-done

- Backend validation is complete and consistent: `validateNodeAgentRef` checks empty role and duplicate role independently of `agentLookup`, the Zod schema in `export-format.ts` validates `role: z.string().min(1)`, and `validateNodeChannels` in `space-utils.ts` correctly separates `agentId` existence checks from slot-role uniqueness checks.
- Read-time backfill in `rowToNode` is safe: `role: a.role?.trim() ? a.role : a.agentId`.
- `resolveNodeChannels` signature cleanup removes the `SpaceAgent[]` dependency cleanly.
- Test coverage is thorough on the daemon and shared sides — all new behaviors are tested.
- The `TODO` annotations on `model` and `systemPrompt` are honest about what is not yet wired.
- The concurrent-spawn guard (`spawningTaskIds`) added to `room-runtime.ts` is a clean fix for the race condition.

lsm (commented):

The core schema work (types, validation, export/import, backfill) is solid and the role-uniqueness enforcement is correct throughout the backend. The previously raised P1 about `addAgent` generating a unique suffix is now addressed in both `WorkflowNodeCard.tsx` and `NodeConfigPanel.tsx`. However one structural bug was introduced alongside that fix, and two test gaps remain.

---

### P1 — Bug: `ChannelsPanelSection` / `ChannelsSection` use `SpaceAgent.role` for channel dropdowns instead of `WorkflowNodeAgent.role`

**Files:**
- `packages/web/src/components/space/visual-editor/NodeConfigPanel.tsx`, line 295
- `packages/web/src/components/space/WorkflowNodeCard.tsx`, line 294

Both channel sections build `knownRoles` for the FROM/TO dropdowns like this:

```ts
const knownRoles = [
  '*',
  ...stepAgents.map((sa) => agents.find((a) => a.id === sa.agentId)?.role ?? sa.agentId),
];
```

This looks up the **SpaceAgent's** `role` (e.g., `"coder"`) rather than using `sa.role` (the **slot-level** `WorkflowNodeAgent.role`, e.g., `"coder-2"`). When the same agent is added twice — the newly supported case that produces slot roles `"coder"` and `"coder-2"` — both dropdown entries show `"coder"`, making it impossible to route channels to the second slot. The fallback path (`?? sa.agentId`) surfaces the raw UUID when the lookup fails, which is also wrong.

The correct expression is simply `sa.role`. Since `buildTaskAgentChannels` was correctly updated to use `sa.role`, this is an oversight rather than a design decision.

Neither `ChannelsPanelSection` nor `ChannelsSection` has a test that catches this regression.

---

### P1 — Missing test: `addAgent` unique-role suffix is untested in UI tests

**Files:**
- `packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx`
- (WorkflowNodeCard has no dedicated test file added in this PR)

The new suffix logic (`"coder"` -> `"coder-2"`) is the headline change called out in the task prompt and in the PR description, but there is no test that:

1. Adds the same agent a second time and asserts `role === "coder-2"` (or whatever the SpaceAgent role is).
2. Asserts that the third addition produces `"coder-3"`, and so on.

The only test added that touches duplicate agents is `'auto-creates task-agent channels when adding an agent in multi-agent mode'`, which adds *different* agents. All existing multi-agent tests use distinct agents with pre-set roles. The suffix logic is present in both components but has zero test coverage.

---

### P2 — Missing test: `model` and `systemPrompt` slot-override fields are not covered in export/import round-trip tests

**File:** `packages/daemon/tests/unit/space/export-format.test.ts`

The new `model` and `systemPrompt` fields on `WorkflowNodeAgent` / `ExportedWorkflowNodeAgent` are added to the Zod schema and the export/import functions, but no test verifies:

- That `model` and `systemPrompt` on a `WorkflowNodeAgent` are included in the exported `agents[]` entry.
- That they survive an import round-trip and are preserved on the resulting `WorkflowNodeAgent`.
- That omitting them in the exported JSON (optional fields) is accepted by the schema.

The existing multi-agent export tests verify `agentRef`, `instructions`, and `channels`, but none assert `model` or `systemPrompt`.

---

### P2 — Missing test: exported `role` field on `agents[]` entries is not explicitly asserted

**File:** `packages/daemon/tests/unit/space/export-format.test.ts`

The `exports multi-agent node as agents array` test checks `node.agents![0].agentRef` and `node.agents![1].agentRef`, but never asserts `node.agents![0].role`. The `role` field is what was added in this PR and its presence in the export output should be a first-class assertion.

---

### P3 — Nit: `ChannelsPanelSection` tests don't verify the `knownRoles` dropdown contents

**File:** `packages/web/src/components/space/visual-editor/__tests__/NodeConfigPanel.test.tsx`

The test `'add channel button is disabled when from or to is empty'` confirms the button is disabled, but none of the channel tests inspect what options appear in the `channel-from-select` dropdown. Had such a test existed it would have caught the `SpaceAgent.role` vs. `WorkflowNodeAgent.role` bug above.

---

### Summary

The schema, validation, and backend changes are clean and well-tested. The two P1 items are both in the UI layer: (a) the channel-role dropdown in both visual editor components must use `sa.role` not `agents.find(...)?.role`, and (b) there is no test covering the headline feature of this PR (unique numeric suffix on duplicate agent additions).

lsm (commented):

**Recommendation: APPROVE** (with P3 nit noted)

---

All issues from the prior review rounds have been addressed. This is a clean, well-tested schema extension.

### What was fixed since the last review

- **P0 (server-side validation):** `space-workflow-manager.ts` now uses `entry.role` from `WorkflowNodeAgent` instead of `agent.role` from `SpaceAgent` when building `knownRoles`. Server-side and runtime channel validation are now in sync.
- **P0 (legacy DB rows):** `space-workflow-repository.ts` backfills `role = agentId` when reading rows persisted before this schema change. Tested in `space-workflow.test.ts`.
- **P0 (dead `model`/`systemPrompt` fields):** Both fields now carry explicit `TODO: Not yet applied at runtime` docstrings in `WorkflowNodeAgent`, `ExportedWorkflowNodeAgent`. The scope is clear.
- **P1 (format-level `role` validation):** `space-workflow-manager.ts` validates non-empty `role` and duplicate role uniqueness before any lookup. Both paths tested.
- **P1 (UI duplicate-agent blocking removed):** Both `WorkflowNodeCard.tsx` and `NodeConfigPanel.tsx` no longer block adding the same agent twice. Unique slot role is auto-generated with numeric suffix (`coder` → `coder-2` → `coder-3`). Tests added for all three variants.
- **P1 (stale docstrings):** All references to `SpaceAgent.role` in `from`/`to` channel documentation updated to `WorkflowNodeAgent.role`.
- **P2 (dead `agents` param):** `resolveNodeChannels` and `validateNodeChannels` signatures cleaned up; `ChannelsPanelSection` and `ChannelsSection` no longer destructure `agents`.

### Remaining issue

**P3 — Dead `agents` prop remains in two component interfaces**

`ChannelsPanelSectionProps` (`NodeConfigPanel.tsx:284`) and `ChannelsSectionProps` (`WorkflowNodeCard.tsx:273`) still declare `agents: SpaceAgent[]` in their props interfaces, but the function implementations no longer destructure or use it. The prop is dead: callers are forced to pass it unnecessarily.

TypeScript does not catch this because the function accepts all declared props; the linter does not flag it because the unused variable is in the interface, not the function body. The prop should be removed from both interfaces. This is a cosmetic/housekeeping issue with no functional impact.

### Web test failures (pre-existing, not introduced by this PR)

Running `make test-web` shows 5 failing tests:
- `useTurnBlocks.test.ts` — 2 failures (`assistantCount` expectation mismatches) — caused by PR #776's `hasTextContent` gate not having its tests updated to match
- `RuntimeMessageRenderer.test.tsx` — 2 failures (CSS class name mismatches: `border-purple-700/50` vs expected `border-purple-800/40`, `text-gray-200` vs expected `text-gray-300`)
- `TaskViewV2.test.tsx` — 1 failure (missing `currentSessionIdSignal` mock export)

All 5 failures existed before this PR (verified: `useTurnBlocks.ts` was last changed in PR #776, the RuntimeMessageRenderer and test were last touched in PR #745, neither is in this PR's diff). This PR added `hiddenCount: 0` to two test fixtures but did not otherwise interact with those tests. These failures are not regressions introduced here, but they should be fixed separately since they mean the web test suite is currently red.

---

The core task is complete: `WorkflowNodeAgent`, `WorkflowNodeInput`, and `ExportedWorkflowNodeAgent` all carry `role`, `model?`, and `systemPrompt?`; role uniqueness is enforced server-side and in the UI; the export/import pipeline is updated; and the test coverage is comprehensive.

PR #779, opened 2026-03-24
fix(e2e): use exact text match for 'Message number 1' selector

lsm:
The partial text selector 'text=Message number 1' was matching multiple
messages (e.g., "Message number 10", "Message number 11", etc.) causing
strict mode violations. Changed to getByText('Message number 1', { exact: true })
to ensure exact matching.

### DIFF
--- a/packages/e2e/tests/features/task-message-pagination.e2e.ts
+++ b/packages/e2e/tests/features/task-message-pagination.e2e.ts
@@ -127,7 +127,9 @@ test.describe('TaskView — Message Pagination', () => {
 			await expect(page.getByText('Message number 1', { exact: true })).toBeVisible({
 				timeout: 10000,
 			});
-			await expect(page.locator('text=Message number 3')).toBeVisible({ timeout: 10000 });
+			await expect(page.getByText('Message number 3', { exact: true })).toBeVisible({
+				timeout: 10000,
+			});
 
 			// No "Load earlier" button shown
 			await expect(page.getByTestId('load-earlier-messages')).not.toBeVisible();
@@ -162,7 +164,9 @@ test.describe('TaskView — Message Pagination', () => {
 			);
 
 			// The newest message (55) should be visible
-			await expect(page.locator('text=Message number 55')).toBeVisible({ timeout: 10000 });
+			await expect(page.getByText('Message number 55', { exact: true })).toBeVisible({
+				timeout: 10000,
+			});
 
 			// The oldest messages should NOT be visible initially (they're hidden by pagination)
 			await expect(page.getByText('Message number 1', { exact: true })).not.toBeVisible();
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Verdict: APPROVE

The fix is correct, targeted, and well-scoped.

### What was done right

- The root cause is accurately identified: `page.locator('text=Message number 1')` is a substring match that matches "Message number 10", "Message number 11", etc., causing strict mode violations when 55 messages exist.
- The fix to `page.getByText('Message number 1', { exact: true })` is the idiomatic Playwright approach and correctly eliminates the ambiguity.
- All three locations where the ambiguous selector appeared (lines 127, 168, 213) were fixed. The PR description clearly explains the problem and solution.

### Remaining selectors are safe

I verified that the unpatched selectors in this file are not vulnerable to the same bug:

- **Line 130** (`text=Message number 3`): Only 3 messages exist in that test case (messages 1-3), so there are no "Message number 13" or "Message number 23" elements to cause ambiguity. No fix needed.
- **Lines 165, 184, 221** (`text=Message number 55`): The substring "Message number 55" does not appear within "Message number 5" (different string), so it unambiguously matches only message 55. No fix needed.

### P2: Consistency suggestion (non-blocking)

For consistency and future-proofing, it would be nice to migrate the remaining `page.locator('text=...')` selectors in this file to `page.getByText(..., { exact: true })` as well. While they are not currently buggy, using a consistent selector strategy across the same test file makes it easier to reason about and less likely that a future message count change reintroduces this class of bug. Specifically:

- `packages/e2e/tests/features/task-message-pagination.e2e.ts` line 130: `page.locator('text=Message number 3')`
- `packages/e2e/tests/features/task-message-pagination.e2e.ts` lines 165, 184, 221: `page.locator('text=Message number 55')`

This is optional and should not block the merge.

lsm (commented):

### Verdict: APPROVE

The second commit (69e418500) cleanly addresses the P2 suggestion from the previous review. All `page.locator('text=...')` selectors in `packages/e2e/tests/features/task-message-pagination.e2e.ts` have been migrated to `page.getByText(..., { exact: true })`, and no instances of the deprecated pattern remain in the file (grep confirms 0 matches).

### Summary of changes verified

- **7 text selectors** now consistently use `page.getByText('Message number N', { exact: true })` across all three test cases (lines 127, 130, 167, 172, 188, 219, 227).
- **0 remaining** `page.locator('text=...')` calls in the file.
- **Scope is correct**: Only `packages/e2e/tests/features/task-message-pagination.e2e.ts` is changed (verified via `git diff dev...HEAD --name-only`).
- The formatting (multi-line `await expect(...)` for long lines) follows the existing codebase pattern and keeps lines under the Biome 100-char width limit.

No issues found. This is ready to merge.

PR #784, opened 2026-03-24
feat(space): resolve base + override config at runtime when spawning agent sessions

lsm:
- Add SlotOverrides interface and slotOverrides field to CustomAgentConfig and
  ResolveAgentInitConfig in custom-agent.ts
- Apply slot model/systemPrompt overrides in createCustomAgentInit: slot model
  takes precedence over agent default; slot systemPrompt replaces agent default
- Update resolveAgentInit to pass slotOverrides through to createCustomAgentInit
- In spawn_step_agent handler: find the WorkflowNodeAgent slot that matches the
  task's customAgentId, extract slot overrides and slot role, pass overrides to
  resolveAgentInit, and use the slot role (WorkflowNodeAgent.role) for session
  group membership and step-agent MCP server — enabling correct channel routing
  when the same agent appears multiple times with different slot roles
- Add 13 unit tests for slot override scenarios in custom-agent.test.ts
- Add 5 integration-style tests for slot role and override behavior in
  task-agent-tools.test.ts; update existing memberInfo role test to reflect
  new slot-role-first behavior

### DIFF
--- a/packages/daemon/src/lib/space/agents/custom-agent.ts
+++ b/packages/daemon/src/lib/space/agents/custom-agent.ts
@@ -33,6 +33,17 @@ const CUSTOM_AGENT_FEATURES: SessionFeatures = {
 // Config
 // ============================================================================
 
+/**
+ * Per-slot overrides from a `WorkflowNodeAgent` entry.
+ * Applied on top of the base `SpaceAgent` config when spawning a specific slot.
+ */
+export interface SlotOverrides {
+	/** Override the agent's default model for this slot */
+	model?: string;
+	/** Override the agent's default system prompt for this slot */
+	systemPrompt?: string;
+}
+
 export interface CustomAgentConfig {
 	/** The custom Space agent definition */
 	customAgent: SpaceAgent;
@@ -53,6 +64,12 @@ export interface CustomAgentConfig {
 	workspacePath: string;
 	/** Summaries of previously completed tasks for context */
 	previousTaskSummaries?: string[];
+	/**
+	 * Optional per-slot overrides from the `WorkflowNodeAgent` entry.
+	 * When provided, `model` replaces the agent's default model and `systemPrompt`
+	 * replaces the agent's default system prompt for this execution slot.
+	 */
+	slotOverrides?: SlotOverrides;
 }
 
 // ============================================================================
@@ -334,15 +351,23 @@ export function buildCustomAgentTaskMessage(config: CustomAgentConfig): string {
  * room-runtime pattern where the initial user message is sent after session start.
  */
 export function createCustomAgentInit(config: CustomAgentConfig): AgentSessionInit {
-	const { customAgent, space, sessionId, workspacePath } = config;
+	const { customAgent, space, sessionId, workspacePath, slotOverrides } = config;
 
 	const customTools =
 		customAgent.tools && customAgent.tools.length > 0 ? customAgent.tools : undefined;
 
-	const model = customAgent.model ?? space.defaultModel ?? DEFAULT_CUSTOM_AGENT_MODEL;
+	// Apply per-slot overrides: slot model takes precedence over agent default.
+	const model =
+		slotOverrides?.model ?? customAgent.model ?? space.defaultModel ?? DEFAULT_CUSTOM_AGENT_MODEL;
 	const provider = inferProviderForModel(model);
 
-	const behavioralPrompt = buildCustomAgentSystemPrompt(customAgent);
+	// Apply per-slot systemPrompt override: slot override replaces agent's default system prompt.
+	// The override is applied by building the prompt with a modified agent copy.
+	const agentForPrompt: SpaceAgent =
+		slotOverrides?.systemPrompt !== undefined
+			? { ...customAgent, systemPrompt: slotOverrides.systemPrompt }
+			: customAgent;
+	const behavioralPrompt = buildCustomAgentSystemPrompt(agentForPrompt);
 
 	// When custom tools are configured, use the agent/agents pattern so the SDK
 	// enforces the allowlist. Otherwise, fall back to the simple preset path.
@@ -418,6 +443,12 @@ export interface ResolveAgentInitConfig {
 	workflow?: SpaceWorkflow | null;
 	/** Summaries of previously completed tasks */
 	previousTaskSummaries?: string[];
+	/**
+	 * Optional per-slot overrides from the `WorkflowNodeAgent` entry for this execution slot.
+	 * When provided, the slot's `model` and/or `systemPrompt` replace the base agent's defaults.
+	 * Used when the same agent appears multiple times in a node with different per-slot configs.
+	 */
+	slotOverrides?: SlotOverrides;
 }
 
 /**
@@ -442,6 +473,7 @@ export function resolveAgentInit(config: ResolveAgentInitConfig): AgentSessionIn
 		workflowRun,
 		workflow,
 		previousTaskSummaries,
+		slotOverrides,
 	} = config;
 
 	if (!task.customAgentId) {
@@ -464,6 +496,7 @@ export function resolveAgentInit(config: ResolveAgentInitConfig): AgentSessionIn
 		sessionId,
 		workspacePath,
 		previousTaskSummaries,
+		slotOverrides,
 	});
 }
 
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -343,6 +343,7 @@ export class SpaceRuntime {
 					workflowNodeId: startStep.id,
 					taskType: resolved.taskType,
 					customAgentId: resolved.customAgentId,
+					slotRole: agentEntry.role,
 					status: 'pending',
 					goalId: run.goalId,
 				});
--- a/packages/daemon/src/lib/space/runtime/workflow-executor.ts
+++ b/packages/daemon/src/lib/space/runtime/workflow-executor.ts
@@ -464,6 +464,7 @@ export class WorkflowExecutor {
 				workflowNodeId: nextStep.id,
 				taskType: resolved?.taskType as import('@neokai/shared').SpaceTaskType | undefined,
 				customAgentId: resolved !== undefined ? resolved.customAgentId : agentEntry.agentId,
+				slotRole: agentEntry.role,
 				status: 'pending',
 				goalId: this.run.goalId,
 			});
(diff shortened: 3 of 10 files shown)


lsm (commented):

The core mechanic works and the test coverage is solid for the single-slot case, but there is a correctness bug in the multi-slot path that would cause incorrect slot matching when the same agent is used twice in one node with different roles or overrides.

---

### P0 — Slot disambiguation bug when same agent appears twice

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 277–355

When a step node has two slots for the same `agentId` with different roles/overrides (e.g. `{ agentId, role: 'strict-reviewer' }` and `{ agentId, role: 'quick-reviewer' }`), the `WorkflowExecutor` creates two `SpaceTask` rows, both sharing the same `customAgentId` and `workflowNodeId`. `spawn_step_agent` selects `stepTasks[stepTasks.length - 1]` (the last task), then runs:

```ts
const agentSlot = nodeAgents.find((a) => a.agentId === effectiveTask.customAgentId);
```

`Array.find` returns the **first** matching slot — always `strict-reviewer`. So the "last task" (which corresponds to the `quick-reviewer` slot) is initialized with the `strict-reviewer`'s role and overrides. The second task is never initialized correctly: it picks up the wrong role for channel routing and the wrong model/systemPrompt overrides.

The `SpaceTask` schema has no field that links a task back to its specific slot position within the node (no `slotRole`, `slotIndex`, or similar). The fix requires either:
1. Storing the slot role on the task at creation time (the cleanest long-term solution, requires a schema + migration change), or
2. Matching tasks to slots by creation order (both tasks sorted by `createdAt` ascending, slots iterated in `nodeAgents` order), but this is fragile.

The existing unit test on line 2076 ("same agent twice in one node uses distinct slot roles for each session") is too weak to catch this: the assertion `expect(['strict-reviewer', 'quick-reviewer']).toContain(capturedInfo?.role)` passes regardless of which slot is chosen, so it does not verify correct slot assignment.

---

### P1 — Stale TODO comments in `space.ts`

**File:** `packages/shared/src/types/space.ts`, lines 590–597 and 892–899

Both `WorkflowNodeAgent.model` and `WorkflowNodeAgent.systemPrompt` carry JSDoc TODO comments saying _"Not yet applied at runtime — workflow-executor.ts does not read this field."_ This PR implements that runtime application, so the comments are now incorrect and will mislead future readers.

```ts
/**
 * Override the agent's default model for this slot.
 * TODO: Not yet applied at runtime — workflow-executor.ts does not read this field.
 */
model?: string;
```

Same for the identical TODOs on `ExportedWorkflowNodeAgent` at lines 892–899.

---

### P1 — Missing test: `systemPrompt` override on the `agent/agents` path

**File:** `packages/daemon/tests/unit/space/custom-agent.test.ts`

When `SpaceAgent.tools` is non-empty, `createCustomAgentInit` takes the `if (customTools)` branch and puts the behavioral prompt into `agentDef.prompt`, not `init.systemPrompt.append`. The test `'systemPrompt override replaces agent default system prompt'` reads `init.systemPrompt?.append`, which is `undefined` on that code path, so the `expect(promptText).toContain(...)` assertions would vacuously pass (empty string does not contain either value) — or would fail for the wrong reason depending on what the override is.

There is no test that exercises `slotOverrides.systemPrompt` when `customTools` is set and verifies the override landed in `init.agents[key].prompt`.

---

### P2 — `slotOverrides` object is constructed even when one field is `undefined`

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 315–319

```ts
const slotOverrides =
    agentSlot?.model !== undefined || agentSlot?.systemPrompt !== undefined
        ? { model: agentSlot?.model, systemPrompt: agentSlot?.systemPrompt }
        : undefined;
```

When only `model` is set, `slotOverrides` is `{ model: 'some-model', systemPrompt: undefined }`. This is functionally harmless because `createCustomAgentInit` guards on `slotOverrides?.systemPrompt !== undefined`, but it is slightly inconsistent: you could cleanly build the object with only the defined keys, or simply always pass `{ model: agentSlot?.model, systemPrompt: agentSlot?.systemPrompt }` and drop the outer ternary since both values being `undefined` is equivalent to `slotOverrides: undefined` in the downstream code.

---

### P3 — Comment typo in test file

**File:** `packages/daemon/tests/unit/space/task-agent-tools.test.ts`, line 1973

`// The memberRole logic: agentSlot.role (agentId) ?? agentForMember.role ('coder') ?? 'agent'`

The `??` operator is JavaScript nullish-coalescing, not a short-circuit that applies to the string in parentheses. The comment is fine as informal pseudocode but could confuse readers — just a nit.

---

### Summary

The slot-override mechanism is correctly wired for the single-agent-per-slot case and is well tested there. The multi-agent-same-agentId case has a real correctness bug: the `find`-first-match behavior means the second duplicate slot always inherits the first slot's role and overrides. The stale TODO comments in `space.ts` should also be removed. Happy to approve once the P0 disambiguation is addressed.

lsm (commented):

**Recommendation: APPROVE** (with minor notes below)

---

The addressed P0 fix is correct and complete. The `slotRole` column is properly added to the DB schema (migration 46, idempotent with `tableHasColumn` guard), stored at task creation in both `SpaceRuntime.startWorkflowRun` and `WorkflowExecutor.followTransition`, round-tripped correctly through the repository (`rowToSpaceTask` reads `slot_role`), and used for exact-match slot lookup in `spawn_step_agent`. The `slotOverrides` construction path is correct even when `agentSlot` is `undefined` (both `?? customAgent.model` and the `!== undefined` systemPrompt guard handle it cleanly).

The 13 + 5 new tests are well-structured and cover the important paths: both code paths in `createCustomAgentInit` (simple and agent/agents), `resolveAgentInit` pass-through, slot role used for group membership, model and systemPrompt overrides applied at spawn time, and the two-instance regression test.

---

### P1 — Misleading test comment + weak assertion

**File:** `packages/daemon/tests/unit/space/task-agent-tools.test.ts`, lines 1972–1980

The comment reads:
```
// For agentId shorthand, slot role === agentId (synthetic). Fall back to base agent role.
/ memberRole = agentSlot.role (synthetic: equals agentId) || agentForMember.role ('coder') || 'agent'
```

This is contradictory. The first line says "Fall back to base agent role" but the second line says `agentSlot.role` is always set (it equals the synthetic agentId). No fallback occurs; `agentSlot?.role` is always the synthetic UUID string for shorthand nodes. The assertion on line 1980 is `toBeTruthy()`, which accepts any non-empty string — the test passes whether the role is the synthetic UUID, the SpaceAgent base role, or literally any string. The test does not verify which specific value is used.

This matters because the test is supposed to document that for shorthand nodes the _synthetic agentId_ is used as the member role (not the SpaceAgent's `role: 'coder'`). The current assertion cannot distinguish between the two. A stronger assertion such as `expect(capturedInfo?.role).toBe(ctx.agentId)` (the synthetic UUID) would pin down the behavior precisely.

Also: line 1973 has a typo — `/ memberRole =` is missing the second slash: should be `// memberRole =`.

---

### P2 — No test for stale slotRole (slot not found after workflow edit)

**File:** `packages/daemon/src/lib/space/tools/task-agent-tools.ts`, lines 317–323

When `effectiveTask.slotRole` is set but `nodeAgents.find((a) => a.role === effectiveTask.slotRole)` returns `undefined` (the workflow definition was edited after the task was created and the slot role no longer exists), `agentSlot` is `undefined`, and `slotOverrides` becomes `{ model: undefined, systemPrompt: undefined }`. The downstream code handles this correctly (no override is applied). However there is no test for this edge case, and the comment "Always construct the object — undefined values are handled downstream" does not acknowledge the stale-slot scenario.

This is not a bug, but a missing test for an important edge case worth documenting.

---

### P3 — "Same agent twice" test covers only the second task

**File:** `packages/daemon/tests/unit/space/task-agent-tools.test.ts`, lines 2076–2130

The regression test for the two-instance scenario correctly verifies that both tasks are created with distinct `slotRole` values (`strict-reviewer` and `quick-reviewer`), and that the last-picked task spawns with role `quick-reviewer`. It does not test spawning the `strict-reviewer` task (the first one). This is acceptable given the pre-existing design where the LLM calls `spawn_step_agent` multiple times per step, but a comment acknowledging this intentional limitation would help future readers.

---

Overall this is a solid, well-executed set of changes. The core correctness concern is addressed, the migration is safe, and the new tests meaningfully improve coverage.

lsm (commented):

**Recommendation: APPROVE**

All three rounds of review feedback have been addressed correctly. Here is what I verified:

---

### P1 fix — agentId shorthand assertion

The `passes agentId and slot role as memberInfo` test (line 638) was updated to `expect(memberInfo?.role).toBe(ctx.agentId)`. The comment now accurately explains the mechanism: `resolveNodeAgents` synthesizes `role = agentId` for shorthand nodes, so `agentSlot.role` is the UUID string, not the base `SpaceAgent.role` ('coder'). The assertion is now pinned exactly and not vacuously true.

I also verified the production path in `spawn_step_agent` (lines 317–363 of `task-agent-tools.ts`): for a shorthand node, `effectiveTask.slotRole` is the agentId UUID (stored at task-creation time by both `space-runtime.ts:346` and `workflow-executor.ts:467`), `nodeAgents.find((a) => a.role === slotRole)` resolves to the synthetic slot, and `memberRole = agentSlot.role` = the UUID. The test assertion matches the code path exactly.

---

### P2 fix — stale-slotRole edge case test

The new test at line 2136 uses raw SQL to inject `slot_role = 'deleted-slot-role'` onto the task, simulates a workflow edit that removed the 'reviewer' slot, then verifies:
1. Spawn succeeds (`success: true`) — stale slotRole is not a hard failure.
2. The slot model override (`'claude-opus-4-6'`) is NOT applied — the base agent config is used instead.

The negative assertion `expect(capturedInit?.model).not.toBe('claude-opus-4-6')` is correct. One minor weakness: it doesn't positively assert the exact fallback model (`DEFAULT_CUSTOM_AGENT_MODEL = 'claude-sonnet-4-5-20250929'`). This could mask a bug where a completely unrelated model is returned. That said, the primary correctness property being tested — "no override applied" — is confirmed, and this is a P3 nit at most.

I also confirmed the production comment in `task-agent-tools.ts` (lines 324–326) accurately describes the stale-slot fallback.

---

### P3 fix — two-instance test comment

The comment at lines 2122–2127 correctly explains why only the `quick-reviewer` (last-created) task is verified in this test. The explanation — that `spawn_step_agent` selects `stepTasks[last]` and the `strict-reviewer` slot is covered by a sibling test — is factually accurate.

---

### No new issues found

- Migration 46 adds the `slot_role` column idempotently (`tableHasColumn` guard present).
- Both task-creation sites (`space-runtime.ts` and `workflow-executor.ts`) set `slotRole: agentEntry.role`.
- The `resolveNodeAgents` function correctly synthesizes `role = agentId` for the shorthand format, which the tests and production code all rely on consistently.
- The `systemPrompt override works with agent/agents path` test (line 586) verifies the override lands in `agents[key].prompt`, not `systemPrompt.append` — this was the missing test from the previous round and it's now present.

PR #788, opened 2026-03-24
feat(shared): add short ID utilities and prefix constants

lsm:


lsm (commented):

---

**Recommendation: APPROVE**

This is a clean, focused implementation that satisfies all acceptance criteria from the task spec. The changes are minimal (2 files, 101 lines added) and the implementation is straightforward.

### Correctness

All acceptance criteria are met:

- `SHORT_ID_PREFIX` has only `TASK: 't'` and `GOAL: 'g'` (no `ROOM: 'r'`) -- correct.
- `formatShortId('t', 42)` returns `'t-42'` -- correct.
- `parseShortId('t-42')` returns `{ prefix: 't', counter: 42 }` -- correct.
- `parseShortId('t-abc')` returns `null` (regex `\d+` rejects non-numeric) -- correct.
- `parseShortId('t-')` returns `null` (regex `\d+` requires at least one digit) -- correct.
- `parseShortId('t-0')` returns `null` (`counter < 1` guard) -- correct.
- `isUUID('04062505-780f-4881-a3be-9cb9062790fb')` returns `true` -- correct.
- `isUUID('t-42')` returns `false` -- correct.
- All utilities exported from `@neokai/shared` via `mod.ts` (line 12: `export * from './utils.ts'`).
- `bun run check` passes with no Knip warnings for the new exports.
- All 18 unit tests pass.

### Test Coverage

The test suite covers the happy paths plus good edge cases (empty string, uppercase prefix, multi-char prefix, UUID passed to parseShortId, UUID v1 rejection, malformed UUID, uppercase UUID). One minor gap: no test for leading zeros like `parseShortId('t-001')` or `parseShortId('t-00')`, but the current behavior (leading zeros stripped by parseInt, `t-00` rejected because counter=0) is reasonable and not a bug.

### P3 (nit): `isUUID` restricts to v4 only

The `isUUID` regex hardcodes the version digit as `4` (line 19 of `packages/shared/src/utils.ts`), meaning UUID v1/v3/v5 will return `false`. This is consistent with the spec and the plan doc (`01-short-id-infrastructure.md` line 29), but future consumers expecting a more general UUID check might be surprised. The function name `isUUID` (not `isUUIDv4`) could be slightly misleading. This is not actionable for this PR -- just noting it for awareness.

### P3 (nit): `formatShortId` accepts `string` for prefix

The `prefix` parameter is typed as `string` rather than something constrained to the values of `SHORT_ID_PREFIX`. This matches the spec (line 27 says `prefix: string`), and the `parseShortId` regex (`^[a-z]$`) already acts as a runtime guard. No issue here.

PR #789, opened 2026-03-24
feat(storage): add migration 47 — short_id columns and short_id_counters table

lsm:
- Add nullable short_id TEXT column to tasks and goals tables (existing DBs via
  ALTER TABLE guarded by tableHasColumn; fresh DBs via updated createTables)
- Create short_id_counters(entity_type, scope_id, counter) table with composite PK
  for per-scope monotonic counter allocation
- Add partial unique indexes idx_tasks_short_id and idx_goals_short_id (WHERE short_id IS NOT NULL)
  so multiple NULLs are allowed while non-null values are globally unique
- Migration is idempotent: all DDL guarded with IF NOT EXISTS or tableHasColumn checks
- Unit tests: 8 cases covering fresh DB schema, counter PK enforcement, partial unique
  index semantics, and idempotency (double-run)

### DIFF
--- a/packages/daemon/src/storage/schema/index.ts
+++ b/packages/daemon/src/storage/schema/index.ts
@@ -162,6 +162,7 @@ export function createTables(db: BunDatabase): void {
         pr_created_at INTEGER,
         input_draft TEXT,
         updated_at INTEGER,
+        short_id TEXT,
         FOREIGN KEY (room_id) REFERENCES rooms(id) ON DELETE CASCADE
       )
     `);
@@ -197,9 +198,28 @@ export function createTables(db: BunDatabase): void {
         max_planning_attempts INTEGER NOT NULL DEFAULT 0,
         consecutive_failures INTEGER NOT NULL DEFAULT 0,
         replan_count INTEGER NOT NULL DEFAULT 0,
+        short_id TEXT,
         FOREIGN KEY (room_id) REFERENCES rooms(id) ON DELETE CASCADE
       )
     `);
+
+	// Short ID counters — per-(entity_type, scope_id) monotonic counter for short ID allocation
+	db.exec(`
+      CREATE TABLE IF NOT EXISTS short_id_counters (
+        entity_type TEXT NOT NULL,
+        scope_id    TEXT NOT NULL,
+        counter     INTEGER NOT NULL DEFAULT 0,
+        PRIMARY KEY (entity_type, scope_id)
+      )
+    `);
+
+	// Partial unique indexes for short_id on tasks and goals
+	db.exec(
+		`CREATE UNIQUE INDEX IF NOT EXISTS idx_tasks_short_id ON tasks(short_id) WHERE short_id IS NOT NULL`
+	);
+	db.exec(
+		`CREATE UNIQUE INDEX IF NOT EXISTS idx_goals_short_id ON goals(short_id) WHERE short_id IS NOT NULL`
+	);
 
 	// Mission metric history table
 	db.exec(`
--- a/packages/daemon/src/storage/schema/index.ts
+++ b/packages/daemon/src/storage/schema/index.ts
@@ -15,6 +15,8 @@ import { DEFAULT_GLOBAL_TOOLS_CONFIG, DEFAULT_GLOBAL_SETTINGS } from '@neokai/sh
 export { runMigrations } from './migrations';
 // knip-ignore-next-line
 export { runMigration12 } from './migrations';
+// knip-ignore-next-line
+export { runMigration47 } from './migrations';
 
 /**
  * Create all database tables and initialize defaults
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -2931,7 +2931,7 @@ function runMigration46(db: BunDatabase): void {
  *
  * Idempotent: ALTER TABLE is guarded by tableHasColumn(); CREATE TABLE/INDEX use IF NOT EXISTS.
  */
-function runMigration47(db: BunDatabase): void {
+export function runMigration47(db: BunDatabase): void {
 	// On existing DBs, ALTER TABLE adds the column; on fresh DBs the column is already
 	// present in the CREATE TABLE statement in createTables(), so tableHasColumn() guards
 	// prevent a duplicate-column error.
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -178,6 +178,11 @@ export function runMigrations(db: BunDatabase, createBackup: () => void): void {
 	// Stores the WorkflowNodeAgent.role of the slot that spawned a task, enabling
 	// unambiguous slot lookup when the same agentId appears multiple times in a node.
 	runMigration46(db);
+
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Summary

Migration 47 correctly adds `short_id TEXT` columns to `tasks` and `goals`, creates the `short_id_counters` table with a composite primary key, and creates partial unique indexes. The SQL is correct, idempotent guards are in place, and all CI checks (lint, format, typecheck) pass. Tests are all green (8/8).

### Issues

**P1 — Missing existing-DB migration path test**

All 8 tests call `runMigrations(db, () => {})` followed by `createTables(db)`, which exercises the fresh-DB path where the new columns are already present from `CREATE TABLE`. The actual migration use case — an existing database where `tasks`/`goals` lack `short_id` — is never tested. This is the primary scenario the migration exists to handle.

The idempotency test (test #8) runs `runMigrations` twice, but after `createTables` has already defined the columns. It does not verify that `ALTER TABLE tasks ADD COLUMN short_id TEXT` works on a pre-existing table without the column.

Compare with `migration-45_test.ts` which explicitly creates pre-migration schema, inserts data, then runs migrations to verify the ALTER path.

Suggested test:
```ts
test('existing DB without short_id gets columns added', () => {
  // Create tasks/goals tables WITHOUT short_id column
  db.exec(`CREATE TABLE rooms (id TEXT PRIMARY KEY, name TEXT NOT NULL, created_at INTEGER NOT NULL, updated_at INTEGER NOT NULL)`);
  db.exec(`CREATE TABLE tasks (id TEXT PRIMARY KEY, room_id TEXT NOT NULL, title TEXT NOT NULL, description TEXT NOT NULL, status TEXT NOT NULL DEFAULT 'pending', priority TEXT NOT NULL DEFAULT 'normal', created_at INTEGER NOT NULL, updated_at INTEGER, FOREIGN KEY (room_id) REFERENCES rooms(id) ON DELETE CASCADE)`);
  db.exec(`CREATE TABLE goals (id TEXT PRIMARY KEY, room_id TEXT NOT NULL, title TEXT NOT NULL, description TEXT NOT NULL DEFAULT '', status TEXT NOT NULL DEFAULT 'active', created_at INTEGER NOT NULL, updated_at INTEGER NOT NULL, FOREIGN KEY (room_id) REFERENCES rooms(id) ON DELETE CASCADE)`);

  // Insert data before migration
  db.exec(`INSERT INTO rooms (id, name, created_at, updated_at) VALUES ('r1', 'Room', 1, 1)`);
  db.exec(`INSERT INTO tasks (id, room_id, title, description, status, priority, created_at, updated_at) VALUES ('t1', 'r1', 'Task', '', 'pending', 'normal', 1, 1)`);

  runMigrations(db, () => {});

  expect(columnExists(db, 'tasks', 'short_id')).toBe(true);
  expect(columnExists(db, 'goals', 'short_id')).toBe(true);
  expect(tableExists(db, 'short_id_counters')).toBe(true);
  // Existing data intact
  const task = db.prepare("SELECT title FROM tasks WHERE id='t1'").get() as { title: string };
  expect(task.title).toBe('Task');
});
```

`packages/daemon/tests/unit/storage/migrations/migration-47_test.ts`

**P2 — Missing goals.short_id partial index behavioral test**

Tests verify partial index behavior (NULL tolerance + duplicate rejection) for `tasks.short_id` but not for `goals.short_id`. Since the goals partial index is created independently and could be subject to a different schema interaction, it should be tested as well.

`packages/daemon/tests/unit/storage/migrations/migration-47_test.ts`, lines 128-169

**P3 — Duplicated test helpers**

`columnExists`, `tableExists`, and `indexExists` are copy-pasted from other migration test files. This is consistent with existing project convention, so it is not blocking, but extracting to a shared test utility (e.g., `tests/unit/storage/helpers.ts`) would reduce drift risk as more migrations add similar helpers.

**P3 — String interpolation in pragma_table_info test helper**

`packages/daemon/tests/unit/storage/migrations/migration-47_test.ts`, line 18: The `columnExists` helper uses `pragma_table_info('${table}')` with string interpolation for the table name. Since this is test-only code with hardcoded table names, there is no real risk, but a parameterized approach would be marginally safer.

### Positive observations

- SQL correctness: Partial unique indexes with `WHERE short_id IS NOT NULL` correctly allow multiple NULLs while enforcing uniqueness on non-null values.
- Idempotency guards are thorough: `tableExists` + `tableHasColumn` for ALTER TABLE, `IF NOT EXISTS` for CREATE TABLE/INDEX.
- `createTables()` and `runMigration47()` stay in sync — both define the same schema.
- Consistent with existing migration patterns (naming, JSDoc comments, test structure).
- All CI checks pass: lint (0 warnings, 0 errors), format (no fixes needed), typecheck.

### Recommendation

**REQUEST_CHANGES** — The missing existing-DB migration path test (P1) leaves the core migration scenario untested. The goals partial index gap (P2) is also worth addressing since it verifies critical uniqueness semantics.

lsm (commented):

### Verdict: APPROVE

All previously raised issues have been properly addressed. Here is a detailed assessment.

---

### P1 (previous): Tests didn't cover the ALTER TABLE migration path — **RESOLVED**

The new test `existing DB without short_id gets columns added by migration` (line 103 of the test file) does exactly what was requested:

1. Creates minimal `rooms`, `tasks`, and `goals` tables **without** the `short_id` column.
2. Seeds a row in each table.
3. Asserts that `short_id` columns do **not** exist before migration.
4. Calls `runMigration47(db)` directly — bypassing all prior migrations, which correctly simulates an existing DB where this is the only new migration.
5. Verifies that columns, indexes, and the counters table now exist.
6. Verifies that pre-existing rows are intact with `NULL` short_id.

This is a well-structured test. The `runMigration47` export from `migrations.ts` (line 2934) and the re-export from `schema/index.ts` (line 19) are both clean. The `knip-ignore-next-line` comments are appropriate since these exports exist purely for test usage.

### P2 (previous): Goals partial index uniqueness not tested — **RESOLVED**

Two new tests have been added:

- `goals short_id unique index allows multiple NULLs` (line 263) — confirms that inserting two goals with no `short_id` succeeds and both rows have `NULL`.
- `goals short_id unique index rejects duplicate non-null values` (line 285) — confirms that inserting a second goal with the same `short_id` value throws.

These mirror the existing tasks coverage perfectly.

### P3 (previous): String interpolation in `columnExists` test helper — **RESOLVED**

The test helper `columnExists` (line 20-25) now uses `pragma_table_info(?)` with both parameters bound via `.get(table, column)`. This is the correct parameterized approach.

---

**P3 (nit): Production `tableHasColumn` in `migrations.ts` (line 1035) still uses string interpolation**

The test helper was fixed, but the production helper `tableHasColumn` at `packages/daemon/src/storage/schema/migrations.ts:1035` still uses:
```ts
.prepare(`SELECT name FROM pragma_table_info('${tableName}') WHERE name = ?`)
```
This is a pre-existing pattern used in 40+ call sites throughout the file, all with hardcoded table-name literals — so there is no injection risk. Not blocking, but worth noting for a future cleanup if desired.

**All 11 tests pass**, covering fresh DB schema, ALTER TABLE migration on existing DBs, composite PK enforcement, partial index semantics for both tasks and goals, and idempotency. The implementation is correct and complete.

PR #790, opened 2026-03-24
feat(daemon): add ShortIdAllocator service with atomic counter allocation

lsm:
Implements ShortIdAllocator in packages/daemon/src/lib/short-id-allocator.ts.
Uses a SQLite transaction (INSERT ... ON CONFLICT DO UPDATE + SELECT) to
atomically increment per-scope counters and return formatted short IDs via
formatShortId(). Includes getCounter() for admin/debug reads.

10 unit tests cover: first allocation, sequential increments, task/goal
independence, cross-scope isolation, uniqueness under concurrent calls,
and getCounter semantics.

### DIFF
--- a/packages/daemon/src/lib/short-id-allocator.ts
+++ b/packages/daemon/src/lib/short-id-allocator.ts
@@ -6,20 +6,20 @@ export class ShortIdAllocator {
 
 	allocate(entityType: 'task' | 'goal', scopeId: string): string {
 		const prefix = entityType === 'task' ? SHORT_ID_PREFIX.TASK : SHORT_ID_PREFIX.GOAL;
-		const counter = this.db.transaction(() => {
-			this.db
-				.prepare(
-					`INSERT INTO short_id_counters (entity_type, scope_id, counter)
-					 VALUES (?, ?, 1)
-					 ON CONFLICT(entity_type, scope_id) DO UPDATE SET counter = counter + 1`
-				)
-				.run(entityType, scopeId);
-			const row = this.db
-				.prepare(`SELECT counter FROM short_id_counters WHERE entity_type = ? AND scope_id = ?`)
-				.get(entityType, scopeId) as { counter: number } | undefined;
-			return row!.counter;
-		})();
-		return formatShortId(prefix, counter);
+		const row = this.db
+			.prepare(
+				`INSERT INTO short_id_counters (entity_type, scope_id, counter)
+				 VALUES (?, ?, 1)
+				 ON CONFLICT(entity_type, scope_id) DO UPDATE SET counter = counter + 1
+				 RETURNING counter`
+			)
+			.get(entityType, scopeId) as { counter: number } | undefined;
+		if (!row) {
+			throw new Error(
+				`short_id_counters row not found after upsert: entity_type=${entityType}, scope_id=${scopeId}`
+			);
+		}
+		return formatShortId(prefix, row.counter);
 	}
 
 	getCounter(entityType: string, scopeId: string): number {
--- a/packages/daemon/src/lib/short-id-allocator.ts
+++ b/packages/daemon/src/lib/short-id-allocator.ts
@@ -1,0 +1,31 @@
+import type { Database as BunDatabase } from 'bun:sqlite';
+import { SHORT_ID_PREFIX, formatShortId } from '@neokai/shared';
+
+export class ShortIdAllocator {
+	constructor(private db: BunDatabase) {}
+
+	allocate(entityType: 'task' | 'goal', scopeId: string): string {
+		const prefix = entityType === 'task' ? SHORT_ID_PREFIX.TASK : SHORT_ID_PREFIX.GOAL;
+		const counter = this.db.transaction(() => {
+			this.db
+				.prepare(
+					`INSERT INTO short_id_counters (entity_type, scope_id, counter)
+					 VALUES (?, ?, 1)
+					 ON CONFLICT(entity_type, scope_id) DO UPDATE SET counter = counter + 1`
+				)
+				.run(entityType, scopeId);
+			const row = this.db
+				.prepare(`SELECT counter FROM short_id_counters WHERE entity_type = ? AND scope_id = ?`)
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

---

The implementation is clean, correct, and well-tested. All four acceptance criteria are met. Here are my findings.

### What works well

- **Correctness**: The INSERT...ON CONFLICT DO UPDATE + SELECT pattern within a transaction correctly atomically increments the per-scope counter and returns the new value. The transaction ensures no race conditions between the upsert and read.
- **Scope isolation**: entity_type + scope_id as composite primary key correctly isolates counters across both dimensions (task vs goal, different rooms).
- **Test coverage**: 10 tests covering first allocation, sequential increments, cross-entity-type independence, cross-scope isolation, uniqueness under sequential calls, and all three getCounter semantics. Tests use an in-memory DB with the correct schema matching the migration.
- **Code style**: Matches existing patterns in the codebase -- inline `.prepare()` calls, same transaction usage pattern as `session-group-repository.ts` and `room-manager.ts`.
- **Shared utilities**: Correctly reuses `SHORT_ID_PREFIX` and `formatShortId` from `@neokai/shared` rather than reimplementing.
- **Table schema**: The migration (47) and `createTables()` both define the same schema, idempotent with IF NOT EXISTS / tableHasColumn guards.

**P1 -- `row!` non-null assertion hides a theoretically possible bug path** (`packages/daemon/src/lib/short-id-allocator.ts`, line 20)

The non-null assertion on `row!.counter` assumes the SELECT always finds the row that was just upserted in the same transaction. While this is virtually guaranteed (the INSERT succeeded within the transaction, so the row exists), using a non-null assertion means a runtime failure here produces an unhelpful "Cannot read property 'counter' of undefined" rather than a descriptive error. Consider replacing it with an explicit null check that throws a descriptive error:

```ts
if (!row) {
  throw new Error(`short_id_counters row not found after upsert: entity_type=${entityType}, scope_id=${scopeId}`);
}
return row.counter;
```

This costs nothing and turns a silent corruption scenario into an actionable error message.

**P2 -- RETURNING clause could eliminate the redundant SELECT** (`packages/daemon/src/lib/short-id-allocator.ts`, lines 10-20)

I verified that Bun:sqlite (Bun 1.3.11, SQLite 3.50.2) fully supports the RETURNING clause on INSERT...ON CONFLICT DO UPDATE. The two-step approach (upsert + separate SELECT) works correctly, but a single statement would be both simpler and slightly more efficient:

```ts
const row = this.db
  .prepare(
    `INSERT INTO short_id_counters (entity_type, scope_id, counter)
     VALUES (?, ?, 1)
     ON CONFLICT(entity_type, scope_id) DO UPDATE SET counter = counter + 1
     RETURNING counter`
  )
  .get(entityType, scopeId) as { counter: number };
return row.counter;
```

This eliminates the redundant SELECT, removes the need for the non-null assertion (RETURNING on a successful write always returns a row), and is a well-known SQLite pattern. However, since the current approach is tested and correct, this is a quality improvement rather than a bug fix.

**P3 -- Test named "concurrent allocations" is actually sequential** (`packages/daemon/tests/unit/short-id/allocator.test.ts`, line 63)

The test name says "concurrent allocations" but uses `Array.from({ length: 50 }, () => allocator.allocate(...))` which runs synchronously in a single thread. It correctly verifies uniqueness under sequential rapid-fire calls but does not actually test concurrent access. If true concurrency testing is desired later, it would need worker threads or a similar approach. The current test is still valuable for verifying monotonic uniqueness.

**P3 -- Prepared statements are not cached** (`packages/daemon/src/lib/short-id-allocator.ts`, lines 11-19)

The SQL statements are prepared on every call to `allocate()` and `getCounter()`. This matches the existing codebase pattern (e.g., `goal-repository.ts` does the same), so it is not a deviation. However, since the allocator is a stateless service that will be called frequently, caching prepared statements as class-level fields would be a minor performance win. Not worth changing now, but worth knowing if allocation becomes a hot path.

lsm (commented):

**Recommendation: APPROVE**

---

All three issues from the previous review have been correctly addressed. No new issues were introduced.

### Verified fixes

**P1 -- Non-null assertion replaced with explicit guard** (resolved)

The `row!` non-null assertion on the old code's line 18 is now replaced with an explicit `if (!row)` check (line 17-21 of `packages/daemon/src/lib/short-id-allocator.ts`) that throws a descriptive `Error` including both `entity_type` and `scope_id`. This matches the exact suggestion from the previous review. The error message is clear and actionable for debugging.

**P2 -- RETURNING clause eliminates redundant SELECT and transaction wrapper** (resolved)

The two-step upsert + SELECT inside `this.db.transaction()` has been replaced with a single `INSERT ... ON CONFLICT DO UPDATE ... RETURNING counter` statement (lines 11-14). This is both simpler and more correct -- SQLite guarantees RETURNING returns the modified row, making the separate SELECT unnecessary. The transaction wrapper was correctly removed since a single statement is inherently atomic. This eliminates the previous P2's concerns about unnecessary round-trips and code complexity.

**P3 -- Test name renamed to accurately reflect synchronous behavior** (resolved)

The test at line 63 of `packages/daemon/tests/unit/short-id/allocator.test.ts` is now named `'rapid sequential allocations produce unique IDs'` instead of `'concurrent allocations produce unique IDs'`. This accurately describes what the test actually does (synchronous single-threaded loop), avoiding the misleading implication of parallel execution.

### Additional verification

- All 10 unit tests pass cleanly.
- Full `bun run check` (lint, typecheck, knip) passes with zero warnings or errors.
- The table schema in the test's `makeDb()` helper matches both `createTables()` in `packages/daemon/src/storage/schema/index.ts` and migration 47 in `packages/daemon/src/storage/schema/migrations.ts`.
- Imports from `@neokai/shared` (`SHORT_ID_PREFIX`, `formatShortId`) resolve correctly to `packages/shared/src/utils.ts`.
- No new callers or integration points were added (the service is standalone, as expected for this PR).

No remaining issues. Ready to merge.

PR #791, opened 2026-03-24
feat(shared): add shortId field to NeoTask and RoomGoal interfaces

lsm:


lsm (commented):

## Summary

Straightforward, correctly implemented type-only change. The PR adds `shortId?: string` as an optional field to both `RoomGoal` (line 149) and `NeoTask` (line 276) interfaces in `packages/shared/src/types/neo.ts`, placed immediately after the `id` field in each interface.

## Verification

- **Lint**: passes (0 warnings, 0 errors)
- **Typecheck**: passes (clean `tsc --build --noEmit`)
- **Field placement**: logically consistent -- placed right after `id` in both interfaces, which makes sense since shortId is a human-readable alias for the same identifier
- **JSDoc comments**: accurate and helpful -- `'g-42'` and `'t-42'` examples match the prefix constants from PR #788, and the scoping note ("scoped to parent room") correctly describes the intended semantics
- **Non-breaking**: since both fields are optional (`?`), this is a purely additive change with no impact on existing call sites

## Notes

- This PR lands after the short ID utilities (#788), migration (#789), and allocator (#790) PRs, which is the correct ordering -- the type definitions serve as the contract that downstream code will eventually populate.
- The 30+ files that reference `RoomGoal` and 33+ files that reference `NeoTask` will not require changes until code starts reading/writing the new field, which is expected.

**Recommendation: APPROVE** -- this PR does exactly what it says and does it correctly.

PR #792, opened 2026-03-24
feat(worktree): add getProjectShortKey and sentinel-based collision detection

lsm:
- Add public `getProjectShortKey(repoPath)` method: produces a short,
  deterministic, human-readable key like `my-project-a3b2c1d4` using
  basename + lower-32-bit Bun.hash hex (BigInt-safe, no truncation)
- Update `getWorktreeBaseDir` to use short keys instead of full encoded
  paths, with `.neokai-repo-root` sentinel file for collision detection:
  - First use: mkdirSync + writeFileSync sentinel with normalized gitRoot
  - Subsequent use (same repo): reads sentinel, matches → proceed normally
  - Collision (different repo): logs warning via this.logger.warn and
    falls back to encodeRepoPath for a distinct directory
  - No sentinel (older NeoKai dir): write sentinel and proceed normally
- Update all existing unit tests to use shortKey-based paths
- Add 11 new tests: 7 for getProjectShortKey and 4 for collision detection

### DIFF
--- a/packages/daemon/src/lib/worktree-manager.ts
+++ b/packages/daemon/src/lib/worktree-manager.ts
@@ -1,6 +1,6 @@
 import simpleGit, { SimpleGit } from 'simple-git';
-import { dirname, join, normalize } from 'node:path';
-import { existsSync, mkdirSync } from 'node:fs';
+import { basename, dirname, join, normalize } from 'node:path';
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
 import { homedir } from 'node:os';
 import type { WorktreeMetadata, CommitInfo, WorktreeCommitStatus } from '@neokai/shared';
 import { Logger } from './logger';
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

The implementation is solid and well-tested. The core logic for `getProjectShortKey` is correct, the sentinel-based collision detection in `getWorktreeBaseDir` is sound, and the 49 unit tests cover all four collision scenarios plus the key-generation edge cases. Lint, typecheck, and all tests pass cleanly.

### Issues Found

**P2 — Stale comment at line 260**

`packages/daemon/src/lib/worktree-manager.ts` line 260:
```
// Format: ~/.neokai/projects/{encoded-repo-path}/worktrees/
```
This comment still references the old `encoded-repo-path` naming convention. It should be updated to reflect the new short-key format, e.g.:
```
// Format: ~/.neokai/projects/{shortKey}/worktrees/
```

**P3 — Test mock type inconsistency**

In `packages/daemon/tests/unit/lib/worktree-manager.test.ts`, the `readFileSync` mock returns `'/test/repo' as unknown as Buffer`. However, the production code calls `readFileSync(sentinelFile, 'utf-8')` which returns `string`, not `Buffer`. The cast works at runtime since it is actually a string under the hood, but it is misleading. Consider mocking with the correct type signature:
``\ts
readFileSyncSpy = spyOn(fs, 'readFileSync').mockImplementation(
  (_path: unknown, _opts?: unknown) => '/test/repo' as unknown as Buffer
);
```
Or more accurately, return the value directly since the production call passes `'utf-8'`:
``\ts
readFileSyncSpy = spyOn(fs, 'readFileSync').mockImplementation(
  () => '/test/repo' as any
);
```
This is purely cosmetic — not a functional bug.

**P3 — `BigInt()` wrapper is redundant**

`packages/daemon/src/lib/worktree-manager.ts` line 183:
``\ts
const hash8 = (BigInt(Bun.hash(normalizedPath)) & 0xffff_ffffn).toString(16).padStart(8, '0');
```
`Bun.hash()` already returns `bigint` natively (verified at runtime), so `BigInt(...)` is a no-op. Consider simplifying to:
``\ts
const hash8 = (Bun.hash(normalizedPath) & 0xffff_ffffn).toString(16).padStart(8, '0');
```
This avoids a superfluous conversion and makes the code slightly more honest about the type.

### What Was Done Well

- The collision detection design is well-thought-out: sentinel file for same-repo verification, clean fallback to full encoding on mismatch, and graceful adoption for pre-existing directories.
- Using `BigInt` + bitwise AND for the 32-bit hash extraction avoids the silent data loss pitfall of `Number(bigint).toString(16)` above 2^53. The comment documenting this is excellent.
- The test helper `shortKeyFor()` ensures path expectations stay synchronized with the production key-generation logic — a good defensive pattern.
- The Logger spy on `Logger.prototype.warn` in the collision tests correctly verifies the warning goes through the structured logger, not `console.warn`.

### Verdict

This is a clean, focused change. The single P2 stale comment should be fixed before merge. The P3 items are optional polish.

lsm (commented):

### Summary

All three issues from the previous review have been verified as correctly addressed.

### Previous Issue Resolution

**P2 (stale comment in createWorktree):** Fixed. Line 260 of packages/daemon/src/lib/worktree-manager.ts now reads the correct shortKey format in the comment. Confirmed no remaining references to encoded-repo-path in the file.

**P3 (redundant mock type casts):** Fixed. All 10 occurrences of `as unknown as Buffer` in the test file have been replaced with `as any`. The default mock implementation includes a clear explanatory comment documenting why `any` is needed to satisfy the overloaded readFileSync type signature. The eslint disable comment is properly placed.

**P3 (BigInt wrapper — skipped by worker):** The worker's justification is correct. TypeScript types Bun.hash() as number | bigint, so wrapping in BigInt() is necessary to perform the bitwise & with the bigint literal 0xffff_ffffn. Without it, TypeScript would reject the expression because number & bigint is not a valid operation.

### Test Verification

All 49 tests pass (0 failures, 72 expect() calls). The new collision detection tests properly exercise all four sentinel scenarios (first use, same repo, collision, legacy dir).

### Recommendation

APPROVE — all prior review findings are resolved.

PR #793, opened 2026-03-24
feat(storage): add short ID assignment and backfill in TaskRepository

lsm:
- Accept optional ShortIdAllocator as third constructor param
- createTask() allocates a short ID (e.g. t-1) via allocator and stores it in short_id column
- rowToTask() maps short_id → shortId field on NeoTask
- Add getTaskByShortId(roomId, shortId) for direct lookup
- getTask() lazy-backfills short_id for legacy rows on first read
- listTasks() lazy-backfills any rows missing short_id inline
- Add short_id column to existing task-repository test schema
- Add task-repository-short-id.test.ts with 15 unit tests covering all new behaviour

### DIFF
--- a/packages/daemon/src/storage/repositories/task-repository.ts
+++ b/packages/daemon/src/storage/repositories/task-repository.ts
@@ -10,11 +10,13 @@ import { generateUUID } from '@neokai/shared';
 import type { NeoTask, TaskFilter, CreateTaskParams, UpdateTaskParams } from '@neokai/shared';
 import type { SQLiteValue } from '../types';
 import type { ReactiveDatabase } from '../reactive-database';
+import type { ShortIdAllocator } from '../../lib/short-id-allocator';
 
 export class TaskRepository {
 	constructor(
 		private db: BunDatabase,
-		private reactiveDb: ReactiveDatabase
+		private reactiveDb: ReactiveDatabase,
+		private shortIdAllocator?: ShortIdAllocator
 	) {}
 
 	/**
@@ -23,10 +25,11 @@ export class TaskRepository {
 	createTask(params: CreateTaskParams): NeoTask {
 		const id = generateUUID();
 		const now = Date.now();
+		const shortId = this.shortIdAllocator?.allocate('task', params.roomId) ?? null;
 
 		const stmt = this.db.prepare(
-			`INSERT INTO tasks (id, room_id, title, description, status, priority, depends_on, task_type, assigned_agent, created_by_task_id, created_at, updated_at)
-       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
+			`INSERT INTO tasks (id, room_id, title, description, status, priority, depends_on, task_type, assigned_agent, created_by_task_id, short_id, created_at, updated_at)
+       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`
 		);
 
 		stmt.run(
@@ -40,12 +43,13 @@ export class TaskRepository {
 			params.taskType ?? 'coding',
 			params.assignedAgent ?? 'coder',
 			params.createdByTaskId ?? null,
+			shortId,
 			now,
 			now
 		);
 
 		this.reactiveDb.notifyChange('tasks');
-		return this.getTask(id)!;
+		return this.getTaskDirect(id)!;
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


lsm (commented):

### Verdict: APPROVE

This is a clean, well-structured implementation that correctly addresses all acceptance criteria. The code is concise, the test coverage is thorough, and the lazy backfill strategy is sound.

---

### P2: `getDraftTasksByCreator` does not backfill `short_id`

`getDraftTasksByCreator()` at line 317 of `task-repository.ts` returns `NeoTask[]` via `rowToTask` but does not perform the same lazy backfill that `getTask()` and `listTasks()` do. If a draft task was created before migration 47 (or via a code path without the allocator), it will have `shortId: undefined` in the returned object. This is an inconsistency with the other read methods.

For draft tasks this is likely low-impact since they are transient and typically get promoted before being surfaced by other read paths, but it creates a subtle inconsistency where the backfill contract is not uniformly applied across all read methods. If a future consumer relies on `shortId` being present on draft tasks, it will be surprised.

**File:** `packages/daemon/src/storage/repositories/task-repository.ts` line 317

---

### P2: `listTasks` backfill is not atomic -- race condition with concurrent callers

In `listTasks()` (lines 131-141), each row missing a `short_id` gets individually allocated and updated in a loop:

```ts
rows.map((row) => {
    const task = this.rowToTask(row);
    if (!task.shortId && this.shortIdAllocator) {
        const shortId = this.shortIdAllocator.allocate('task', roomId);
        this.db.prepare(`UPDATE tasks SET short_id = ? WHERE id = ?`).run(shortId, task.id);
        return { ...task, shortId };
    }
    return task;
});
```

The `allocate()` call uses an atomic upsert (`ON CONFLICT DO UPDATE SET counter = counter + 1`), so the counter increment itself is safe. However, the separate `UPDATE` means two concurrent `listTasks()` calls for the same room could both read the same unbackfilled row, both allocate new short IDs (incrementing the counter twice), and both write. The first write wins and the second overwrites with a different short ID. The row ends up correct (not corrupted), but a short ID counter value is permanently skipped.

This is not a correctness bug (no unique constraint violation since the first UPDATE sets the value and the second UPDATE just changes it), and the skipped counter is cosmetic. However, the same pattern does not exist in `getTask()` because it reads-then-writes the same row by primary key, which in SQLite serialized mode is inherently sequential.

Given this is SQLite (single-writer) in practice, this race is unlikely to manifest. Flagging as P2 for awareness rather than requiring a fix.

**File:** `packages/daemon/src/storage/repositories/task-repository.ts` lines 131-141

---

### P3: Test schema in `task-repository-short-id.test.ts` is missing `input_draft` column

The `makeDb()` function at line 18 defines the `tasks` table without the `input_draft TEXT` column. While this does not cause test failures (the column is never set in tests, and `rowToTask` maps it as `undefined`), it means the test schema diverges from the production schema. The existing `task-repository.test.ts` has the same pre-existing gap, so this is not introduced by this PR -- just noting for completeness.

**File:** `packages/daemon/tests/unit/storage/task-repository-short-id.test.ts` line 18

---

### What works well

- **Clean separation of concerns**: `getTaskDirect()` as a private no-backfill method to avoid recursion in `createTask()` is a good pattern. Clear and intentional.
- **Optional allocator**: The `shortIdAllocator?: ShortIdAllocator` parameter with optional chaining (`?.allocate()`) makes backward compatibility effortless. Existing callers without the allocator continue to work unchanged.
- **`createTask()` returns via `getTaskDirect()`**: Correctly avoids triggering backfill for a row that was just inserted with a short ID. No wasted work.
- **Thorough test coverage**: The 13 tests cover creation with/without allocator, per-room counter isolation, `getTaskByShortId` (found, not found, wrong room), lazy backfill in `getTask` and `listTasks` (mixed rows, no-op when already assigned), and counter stability verification.
- **Backfill correctness**: Tests verify that the DB row is actually updated (not just the in-memory object), which is the critical assertion for backfill correctness.
- **`getTaskByShortId` scoped to room**: The `WHERE room_id = ? AND short_id = ?` query is correct and safe. The global unique index on `short_id` plus the room filter ensures both correctness and performance.

---

**Recommendation: APPROVE** -- The two P2 items are low-impact observations. The implementation is correct, well-tested, and ready to merge. The allocator wiring in production callers (`TaskManager`, `GoalManager`, `RoomManager`, `task-handlers.ts`) is intentionally left for a follow-up PR, which is the right approach for incremental rollout.

lsm (commented):

## Follow-up Review (commit b503b58)

Both P2 issues from the previous review have been properly addressed.

### P2-1: getDraftTasksByCreator now applies lazy backfill -- RESOLVED

The backfill logic in `getDraftTasksByCreator` (line 332-340 of `packages/daemon/src/storage/repositories/task-repository.ts`) is identical in structure to `listTasks` and `getTask`: it checks `!task.shortId && this.shortIdAllocator`, allocates a new short ID via the allocator, persists it to the DB, and returns the enriched object. This ensures all three public read methods now uniformly populate `shortId` for legacy rows.

Two unit tests were added:
- **backfills short_id for legacy draft rows** (line 216-232): Inserts a row without short_id, calls `getDraftTasksByCreator`, verifies both the returned object and the DB row have the allocated short ID. Correctly tests the backfill path.
- **does not alter already-assigned short IDs** (line 234-250): Creates a task via `createTask` (which sets short_id), then calls `getDraftTasksByCreator` and verifies the counter was not incremented and the original shortId is preserved. Correctly tests the no-op path.

All 15 tests pass (plus the 58 existing task-repository tests).

### P2-2: listTasks concurrency comment -- RESOLVED

The comment at lines 132-138 of `task-repository.ts` accurately explains SQLite's single-writer serialization guarantee: concurrent callers to `listTasks` for the same room would serialize on the write lock, and the second caller would read already-written short_id values. Any skipped counter value (from a caller that allocated but crashed before persisting) is cosmetic since counters are never reused. This is a reasonable justification for not adding additional synchronization.

### Minor observation (P3)

There is a stale orphaned JSDoc comment at lines 318-320 that reads `"Convert a database row to a NeoTask object"`. This was the original JSDoc for `rowToTask`, but the follow-up commit inserted the new `getDraftTasksByCreator` JSDoc block between it and `rowToTask`, leaving `rowToTask` at line 343 without a doc comment. This is purely cosmetic and has no functional impact.

### Verdict

Both issues are properly resolved. The implementation is consistent, well-tested, and the concurrency concern is adequately documented. Approving.

PR #798, opened 2026-03-24
test(short-id): add multi-tenant isolation unit tests for scoped counters

lsm:
Proves that short ID counters are scoped per room and do not bleed across
rooms. Covers tasks and goals in Room A vs Room B with independent counters,
cross-room getTaskByShortId/getGoalByShortId returning null, and direct
verification of the short_id_counters table having separate rows per
(entity_type, scope_id) pair.

### DIFF
--- a/packages/daemon/tests/unit/short-id/multi-tenant.test.ts
+++ b/packages/daemon/tests/unit/short-id/multi-tenant.test.ts
@@ -1,0 +1,323 @@
+/**
+ * Multi-tenant isolation tests for short ID counters
+ *
+ * Proves that short ID counters are scoped per room and do not bleed across rooms.
+ * Each (entity_type, scope_id) pair has its own independent counter row.
+ */
+
+import { describe, expect, it, beforeEach, afterEach } from 'bun:test';
+import { Database } from 'bun:sqlite';
+import { TaskRepository } from '../../../src/storage/repositories/task-repository';
+import { GoalRepository } from '../../../src/storage/repositories/goal-repository';
+import { ShortIdAllocator } from '../../../src/lib/short-id-allocator';
+import { noOpReactiveDb } from '../../helpers/reactive-database';
+
+function makeDb(): Database {
+	const db = new Database(':memory:');
+	db.exec(`
+		CREATE TABLE tasks (
+			id TEXT PRIMARY KEY,
+			room_id TEXT NOT NULL,
+			title TEXT NOT NULL,
+			description TEXT NOT NULL,
+			status TEXT NOT NULL DEFAULT 'pending',
+			priority TEXT NOT NULL DEFAULT 'normal',
+			task_type TEXT NOT NULL DEFAULT 'coding',
+			assigned_agent TEXT DEFAULT 'coder',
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured and thorough test file that directly proves the multi-tenant isolation property of short ID counters. All 14 tests pass. Below are my detailed findings.

### What was verified

- Test schema (`short_id_counters` PK, `tasks`/`goals` tables) matches the production migration 47/48 schema exactly.
- `ShortIdAllocator.allocate()` uses `INSERT ... ON CONFLICT(entity_type, scope_id) DO UPDATE SET counter = counter + 1` with `RETURNING counter`, which is the correct atomic upsert for scoped isolation.
- `TaskRepository.createTask()` and `GoalRepository.createGoal()` both pass `roomId` as the `scopeId` to `allocator.allocate()`, confirming scope is per-room.
- `getTaskByShortId()` and `getGoalByShortId()` both filter by `room_id` in their WHERE clause, confirming lookup isolation.
- The `noOpReactiveDb` stub is appropriate for repository-level unit tests that do not need change notification.

### Test coverage assessment

The 14 tests across 4 describe blocks cover the key isolation properties:

1. **Independent counters per room** (tasks: lines 107-131, goals: lines 196-215) -- verified by creating tasks/goals in two rooms and asserting sequential short IDs start from 1 independently.
2. **Different UUIDs for same short ID** (lines 133-141, 196-203) -- confirms t-1/g-1 in Room A and Room B resolve to different underlying records.
3. **Cross-room lookup returns null** (lines 143-173, 230-238) -- verifies `getTaskByShortId` and `getGoalByShortId` are room-scoped and cannot see another room's entities.
4. **Entity type independence** (lines 241-254) -- task and goal counters do not interfere within the same room.
5. **Direct table verification** (lines 257-297) -- queries `short_id_counters` directly to confirm 4 distinct rows with correct counter values.
6. **PK constraint enforcement** (lines 299-309) -- proves duplicate PK insertion throws, confirming the schema-level isolation guarantee.
7. **getCounter API** (lines 312-321) -- verifies the `getCounter()` public method returns correct per-room state.

### Minor observations (P3)

- **Table schema duplication**: The `makeDb()` function duplicates the `tasks`, `goals`, and `short_id_counters` table schemas. The sibling file `allocator.test.ts` has its own `makeDb()` with just `short_id_counters`. A shared schema helper could reduce drift risk, but this is a minor maintenance concern and the current approach is clear and self-contained.
- **Only two rooms tested**: The tests use exactly 2 rooms (ROOM_A, ROOM_B). While this is sufficient to prove isolation, adding a 3rd room in at least one test would further guard against off-by-one or hardcoded assumptions. The `getCounter` test at line 320 partially addresses this with `'room-c-uuid-0003'` for the "never used" case.

### Conclusion

The implementation correctly and comprehensively proves that short ID counters are scoped per room and do not bleed across rooms. The tests are well-organized, cover both tasks and goals, verify at multiple layers (API, direct DB query, PK constraint), and align perfectly with the production schema and allocator logic.

PR #801, opened 2026-03-24
feat(web): replace flat progress bar with circular indicator in task list items

lsm:
Remove the blue horizontal progress bar from TaskItem and replace the
percentage text with a CircularProgressIndicator (size=24), consistent
with the style used in TaskView and TaskViewV2 headers.

### DIFF
--- a/packages/web/src/components/room/RoomTasks.test.tsx
+++ b/packages/web/src/components/room/RoomTasks.test.tsx
@@ -464,21 +464,28 @@ describe('RoomTasks', () => {
 			selectedTabSignal.value = 'active';
 		});
 
-		it('should show progress percentage when defined', () => {
+		it('should show circular progress indicator when progress is defined', () => {
 			const tasks = [createTask('t1', 'in_progress', { progress: 75 })];
 
 			const { container } = render(<RoomTasks tasks={tasks} />);
 
-			expect(container.textContent).toContain('75%');
+			// Circular indicator renders an SVG, not a percentage text
+			const svg = container.querySelector('svg[width="24"]');
+			expect(svg).toBeTruthy();
 		});
 
-		it('should show progress bar when progress is defined', () => {
+		it('should show circular progress indicator instead of flat bar when progress is defined', () => {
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

This is a clean, focused change that replaces the flat blue progress bar and percentage text in task list items with a circular progress indicator. The implementation is consistent with how `TaskView` and `TaskViewV2` already use `CircularProgressIndicator`.

### What was done correctly

1. **Import and usage**: `CircularProgressIndicator` is imported and used with `size={24}`, which is appropriately smaller than the `size={32}` used in `TaskView`/`TaskViewV2` headers -- good visual hierarchy for a list item context.
2. **Null guard**: The condition `task.progress != null && task.progress > 0` correctly uses `!= null` to catch both `undefined` and `null` (matching the `TaskSummary` type where `progress?: number | null`), and excludes the zero-progress case to avoid rendering a visually meaningless empty circle.
3. **Accessibility**: The `title` prop (`Task progress: ${task.progress}%`) is set, providing a tooltip for users who cannot see the visual indicator.
4. **Blue bar fully removed**: No remaining `bg-blue-500` class exists in `RoomTasks.tsx`. The old horizontal bar block is cleanly deleted.
5. **Tests updated**: Both affected tests now verify the SVG element at `svg[width="24"]` is present, and one test explicitly asserts the absence of the flat blue bar.
6. **All 97 tests pass**.

### Suggestions (non-blocking, P3)

- **P3 -- Missing edge-case test for progress=0**: The `createTask` helper defaults `progress: 0`, so many existing tests implicitly verify that no indicator renders at zero. However, an explicit test case in the "Task Progress" describe block that asserts `svg[width="24"]` is falsy when `progress` is `0` or `undefined` would make the intent clearer and guard against future regressions where someone changes the condition.
- **P3 -- Test names overlap**: The two test names ("should show circular progress indicator when progress is defined" and "should show circular progress indicator instead of flat bar when progress is defined") are very similar. Consider renaming the first to focus on the percentage text replacement.

### Verdict

The change is correct, minimal, and consistent with existing patterns. No blocking issues found.

PR #802, opened 2026-03-24
fix(task): transition review task to in_progress when human sends a message

lsm:
When a human sends a message to a task in 'review' status via
task.sendHumanMessage, the task now automatically transitions to
'in_progress' before the message is routed to the target agent.

Previously the task stayed in 'review' status even after a human
provided direct feedback to the worker or leader, causing the UI
to show incorrect status. The group sessions are still active in
review state, so no revival is needed — only a status update.

Adds unit tests covering worker target, leader target, transition
failure, and that reviveTaskForMessage is not called for review tasks.

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
@@ -1046,6 +1046,19 @@ export function setupTaskHandlers(
 			return { success: true };
 		}
 
+		// review tasks: transition to in_progress before routing the human message.
+		// The group is still active (sessions running), so no revival is needed — just
+		// update the status so the task reflects that work is ongoing again.
+		if (task.status === 'review') {
+			try {
+				await taskManager.setTaskStatus(params.taskId, 'in_progress');
+			} catch (err) {
+				throw new Error(
+					`Failed to transition task ${params.taskId} from review to in_progress: ${String(err)}`
+				);
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Verdict: APPROVE

The implementation is clean, correct, and well-tested.

### What was done well

- **Correct placement**: The new `review` status check at line 1049-1060 of `task-handlers.ts` sits precisely where it should — after the `needs_attention`/`completed`/`cancelled` revival block (which handles its own `review` intermediate transition for `needs_attention` tasks) and before `routeHumanMessageToGroup`. Review tasks skip revival entirely and only get a status update, which is semantically correct since their sessions are still active.

- **Valid transition**: `review -> in_progress` is explicitly allowed in `VALID_STATUS_TRANSITIONS` (line 32 of `task-manager.ts`).

- **Consistent error handling**: The try/catch pattern with a descriptive error message matches the existing style in the revival block above (line 1021-1025).

- **UI notification**: The existing `emitTaskUpdate` call at line 1077-1080 fetches the updated task after message routing and emits the change, so the frontend will reflect the status transition.

- **Good test coverage**: Four tests cover the key scenarios — message to worker, message to leader, transition failure, and the negative assertion that `reviveTaskForMessage` is NOT called for review tasks. The `setupWithReviewTask` helper properly mocks the task manager with review status.

### No issues found

All 102 tests pass. The change is minimal and surgically addresses the reported bug.

PR #804, opened 2026-03-24
feat(web): show room-specific bottom tabs on mobile when in room context

lsm:
When the user is viewing a room on mobile (dashboard, agent, task, or
session view), the bottom tab bar now switches from the global tabs
(Inbox / Rooms / Chats / Settings) to room-contextual tabs:

  Overview | Agent | Inbox | Settings

The "Overview" tab navigates back to the room dashboard and the "Agent"
tab jumps directly to the room agent chat (/room/:id/agent), making it
one tap away from any room view instead of requiring the user to open
the left drawer first.

Tabs revert to the global set as soon as the user leaves the room
context (e.g., taps Inbox).

Adds E2E tests covering:
- Room-specific tabs visible in room context
- Agent tab navigates to correct URL
- Overview tab returns to room dashboard from agent view
- Tabs restore to global set when leaving room

### DIFF
--- a/packages/e2e/tests/helpers/room-helpers.ts
+++ b/packages/e2e/tests/helpers/room-helpers.ts
@@ -10,6 +10,18 @@ import type { Page } from '@playwright/test';
 import { waitForWebSocketConnected } from './wait-helpers';
 
 /**
+ * Create a room via RPC. For use in beforeEach setup only.
+ */
+export async function createRoom(page: Page, name: string): Promise<string> {
+	return page.evaluate(async (roomName) => {
+		const hub = window.__messageHub || window.appState?.messageHub;
+		if (!hub?.request) throw new Error('MessageHub not available');
+		const res = await hub.request('room.create', { name: roomName });
+		return (res as { room: { id: string } }).room.id;
+	}, name);
+}
+
+/**
  * Delete a room via RPC. Best-effort — silently ignores errors so it can be
  * used safely in afterEach without masking test failures.
  */
--- a/packages/e2e/tests/responsive/mobile.e2e.ts
+++ b/packages/e2e/tests/responsive/mobile.e2e.ts
@@ -5,10 +5,16 @@
  * - Layout adaptation
  * - Input behavior on mobile
  * - Message display on mobile
+ * - Room agent navigation via bottom tab bar
  */
 
 import { test, expect, devices } from '../../fixtures';
-import { cleanupTestSession, createSessionViaUI } from '../helpers/wait-helpers';
+import {
+	cleanupTestSession,
+	createSessionViaUI,
+	waitForWebSocketConnected,
+} from '../helpers/wait-helpers';
+import { deleteRoom } from '../helpers/room-helpers';
 
 test.describe('Mobile Layout', () => {
 	// Use iPhone 13 viewport for mobile tests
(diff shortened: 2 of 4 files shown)


lsm (commented):

---

### P1: Overview tab incorrectly highlighted when viewing room tasks or room sessions

**File:** `packages/web/src/islands/BottomTabBar.tsx`, lines 145-151

The `isTabActive` function for `room-overview` only checks whether the user is viewing the room agent:

```ts
const isTabActive = (id: TabItem['id']): boolean => {
    if (isInRoomContext) {
        if (id === 'room-agent') return isViewingRoomAgent;
        if (id === 'room-overview') return !isViewingRoomAgent && navSection === 'rooms';
    }
    return navSection === id;
};
```

This logic treats any room context that is NOT the agent view as the Overview (dashboard) being active. However, when the user is viewing a room task (`/room/:id/task/:taskId`) or a room session (`/room/:id/session/:sessionId`), the Overview tab incorrectly appears highlighted.

The `RoomContextPanel` already has the correct pattern at `packages/web/src/islands/RoomContextPanel.tsx` line 155:

```ts
const isDashboardSelected = selectedSessionId === null && selectedTaskId === null;
```

The `BottomTabBar` should follow the same pattern. It needs to import `currentRoomTaskIdSignal` and check that both `currentRoomSessionIdSignal` and `currentRoomTaskIdSignal` are null before marking Overview as active:

```ts
const roomTaskId = currentRoomTaskIdSignal.value;
const isViewingRoomDashboard = !isViewingRoomAgent
    && roomSessionId === null
    && roomTaskId === null;

// In isTabActive:
if (id === 'room-overview') return isViewingRoomDashboard && navSection === 'rooms';
```

---

### P1: Unit tests not updated to cover new room-context behavior

**File:** `packages/web/src/islands/__tests__/BottomTabBar.test.tsx`

The existing 27 tests all pass because they only exercise the global tab behavior (`currentRoomIdSignal` defaults to `null`). There are **zero tests** for:

- Room-specific tabs appearing when `currentRoomIdSignal` is set
- Agent tab being active when `currentRoomSessionIdSignal` is `room:chat:<roomId>`
- Overview tab being active when on room dashboard (and inactive when on room task/session)
- `handleTabClick` dispatching `navigateToRoom` / `navigateToRoomAgent` for room tabs
- No tab being highlighted when viewing a room task or room session (edge case noted above)

Additionally, the `vi.mock('../../lib/router.ts')` does not include `navigateToRoom` or `navigateToRoomAgent`. While this does not cause test failures today (since room tabs are never rendered in existing tests), it should be updated for completeness to avoid confusion.

---

### P2: E2E test missing coverage for room task and room session sub-views

**File:** `packages/e2e/tests/responsive/mobile.e2e.ts`, lines 294-405

The four E2E tests cover the dashboard and agent views well, but there is no test for:

- Navigating to a room task URL and verifying the bottom tab bar still shows room-specific tabs
- Navigating to a room session URL and verifying the same

These sub-views are exactly where the P1 highlight bug above would be visible. Adding a test that navigates to a room task and asserts that neither the Overview nor Agent tab has `aria-selected="true"` would both validate the fix and prevent regression.

---

### P2: E2E test accesses window.__messageHub directly in beforeEach

**File:** `packages/e2e/tests/responsive/mobile.e2e.ts`, lines 309-314

Per CLAUDE.md, `hub.request()` in test actions is prohibited. Room create/delete in `beforeEach`/`afterEach` is an allowed exception, so this usage is acceptable. However, the pattern diverges from the project convention of using the `deleteRoom` helper from `room-helpers.ts` for teardown while using inline `page.evaluate` for setup. Consider adding a `createRoom` helper to `room-helpers.ts` for symmetry (other test files like `room-sidebar-sections.e2e.ts` already do this).

---

### P3: Minor -- handleTabClick re-created on every render

**File:** `packages/web/src/islands/BottomTabBar.tsx`, line 122

`handleTabClick` is defined as a regular function inside the component body, which means a new closure is created on every render. Since it captures `roomId` from the outer scope, it cannot be trivially hoisted outside the component. This is a minor performance concern and not blocking, but if the component is re-rendered frequently due to signal changes (e.g., `inboxBadgeCount`), it creates unnecessary garbage. This is a pre-existing pattern from before this PR and not introduced by this change.

lsm (commented):

**Recommendation: APPROVE**

All previous P1 and P2 issues have been properly resolved.

### P1 fixes verified

1. **Tab highlighting on room task/session sub-views** (fixed): The `isViewingRoomDashboard` variable at `packages/web/src/islands/BottomTabBar.tsx:122` now correctly checks `roomSessionId === null && roomTaskId === null`, mirroring the `isDashboardSelected` pattern in `RoomContextPanel.tsx:155`. The `isTabActive` function at line 152 uses this instead of the previous broken `!isViewingRoomAgent` check. This correctly ensures that neither Overview nor Agent highlights when viewing a task sub-view or a non-agent session sub-view.

2. **Unit tests** (fixed): 9 new tests in `packages/web/src/islands/__tests__/BottomTabBar.test.tsx` (lines 319-414) cover all room-context tab behaviors: tab visibility, active state for each view (dashboard, agent, task, session), navigation calls, and the boundary condition where `navSection=rooms` but `roomId=null`. All 36 tests pass.

### P2 fixes verified

1. **`createRoom()` helper**: Added to `packages/e2e/tests/helpers/room-helpers.ts` (lines 15-22). Follows the same pattern as the existing `deleteRoom()` helper with proper `page.evaluate` and hub access. The E2E test in `mobile.e2e.ts` has been updated to use it.

2. **E2E sub-view coverage**: New test at `packages/e2e/tests/responsive/mobile.e2e.ts:405-458` navigates through all three URL patterns (room dashboard, task sub-view, agent view) and asserts `aria-selected` state at each step. The existing agent-to-overview test also now asserts that Overview is not active when on the agent view (line 370-373).

### Minor observations (P3)

- The `navSection === 'rooms'` check on line 152 is redundant since `isInRoomContext` (line 150) already guarantees `navSection === 'rooms'`. Not a bug, just a defensive redundancy.
- Several other E2E test files still use inline `room.create` calls instead of the `createRoom()` helper, but migrating those is out of scope for this PR.

### Summary

Clean fix. Logic is correct, tests are comprehensive and passing, typecheck and lint are clean.

**Recommendation: APPROVE**

PR #807, opened 2026-03-24
fix(web): fix goal timestamp always showing "now", improve list item layout

lsm:
- Fix relativeTime() to treat createdAt as milliseconds (not seconds).
  The RoomGoal type documents createdAt as ms since epoch; multiplying
  by 1000 produced a far-future value so diff was always negative,
  returning "just now" for every goal.
- Fix test fixture to use Date.now() - based ms timestamps (was using
  Math.floor(Date.now()/1000) which is seconds, inconsistent with type).
- Add tests verifying "1 day ago" for old goals and "just now" for new.
- Improve GoalItem card layout: put title on row 1 (prominent) with
  actions, and move status/badges to row 2 for better visual hierarchy.

### DIFF
--- a/packages/web/src/components/room/GoalsEditor.test.tsx
+++ b/packages/web/src/components/room/GoalsEditor.test.tsx
@@ -26,8 +26,8 @@ describe('GoalsEditor', () => {
 		priority: 'normal',
 		progress: 50,
 		linkedTaskIds: [],
-		createdAt: Math.floor(Date.now() / 1000) - 86400,
-		updatedAt: Math.floor(Date.now() / 1000),
+		createdAt: Date.now() - 86400 * 1000,
+		updatedAt: Date.now(),
 		...overrides,
 	});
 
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

This PR fixes a real bug where `relativeTime()` was treating a millisecond timestamp as seconds (multiplying by 1000 again), causing all goal timestamps to display as "just now". The fix is correct and well-targeted. The layout reordering and test additions are solid.

### Verdict: APPROVE

---

### Timestamp fix -- Correct

The `RoomGoal.createdAt` type is documented as "milliseconds since epoch" (packages/shared/src/types/neo.ts:172), and `GoalRepository.createGoal()` stores `Date.now()` directly (goal-repository.ts:91). The original code `ts * 1000` was erroneously double-converting, producing a timestamp ~1000x in the future. Since `Date.now() - farFuture` yields a large negative number, `Math.floor(negative / 60000)` is < 1, so every goal returned "just now".

The fix (removing `* 1000`) is correct.

**Note:** The `nextRunAt` field on line 1312 still uses `* 1000` -- this is also correct because `nextRunAt` is explicitly documented as "Unix timestamp (seconds)" (neo.ts:189), a different unit from `createdAt`.

### Test fixture fix -- Correct

Changed `Math.floor(Date.now() / 1000) - 86400` to `Date.now() - 86400 * 1000`, making the fixture consistent with the type contract (milliseconds).

### New tests -- Good

The two new test cases verify the fix directly:
- "1 day ago" for old goals (not "just now")
- "just now" for recent goals (5 seconds ago)

These tests are meaningfully tied to the bug being fixed. The existing test at line 305-310 ("should show relative creation time") also implicitly validates the fix since the default fixture uses 24h ago.

### Layout change -- Clean, no regressions

Reordering Row 1 (title + actions) above Row 2 (status/badges) improves visual hierarchy by making the goal title the first thing the user sees. The change is a pure reordering with no structural changes. Minor observation: the title font size changed from `text-base` to `text-sm`, which is appropriate for a list-item card context but worth noting in case the team expected the title to remain `text-base`.

### P3 notes (no action required)

1. **Pre-existing: Duplicated `formatRelativeTime` implementations** -- There are at least 4 independent copies across the web package (GoalsEditor.tsx, RoomGrid.tsx, RoomSessions.tsx, RoomContextPanel.tsx, and utils.ts). This is not introduced by this PR, but worth consolidating in a future cleanup.

2. **Negative diff handling in `relativeTime()`** -- If clock skew causes `ts > Date.now()`, `diff` becomes negative and `mins < 1` returns "just now", which is reasonable graceful degradation. No change needed.

PR #809, opened 2026-03-24
feat(task): make rate/usage limits first-class task concerns with persistence and auto-retry

lsm:
- Add `rate_limited` and `usage_limited` to TaskStatus and VALID_STATUS_TRANSITIONS
- Add TaskRestriction interface (type, limit, resetAt, sessionRole, retryAfter)
- Persist restrictions field on task record when limit is detected in RoomRuntime
- Clear restrictions field when task resumes via send_to_worker or setTaskStatus
- Add recoverStuckWorkers integration: skip active limits, re-route after expiry
- Add Migration 49: adds restrictions column and expands status CHECK constraint
- Add unit tests for restriction persistence, transitions, and migration

### DIFF
--- a/packages/daemon/src/lib/room/managers/task-manager.ts
+++ b/packages/daemon/src/lib/room/managers/task-manager.ts
@@ -28,12 +28,21 @@ import type {
 export const VALID_STATUS_TRANSITIONS: Record<TaskStatus, TaskStatus[]> = {
 	draft: ['pending'],
 	pending: ['in_progress', 'cancelled'],
-	in_progress: ['review', 'completed', 'needs_attention', 'cancelled'],
+	in_progress: [
+		'review',
+		'completed',
+		'needs_attention',
+		'cancelled',
+		'rate_limited',
+		'usage_limited',
+	],
 	review: ['completed', 'needs_attention', 'in_progress'],
 	completed: ['in_progress', 'archived'], // Reactivate or archive
 	needs_attention: ['pending', 'in_progress', 'review', 'archived'], // Restart allowed + archive
 	cancelled: ['pending', 'in_progress', 'completed', 'archived'], // Restart, complete, or archive
 	archived: [], // True terminal state — no going back
+	rate_limited: ['in_progress', 'needs_attention', 'cancelled'], // Resume or fail
+	usage_limited: ['in_progress', 'needs_attention', 'cancelled'], // Resume or fail
 };
 
 /**
@@ -242,6 +251,14 @@ export class TaskManager {
 			updates.error = null;
 			updates.result = null;
 			updates.progress = null;
+		}
+
+		// Clear restrictions when resuming from a rate/usage limited state.
+		if (
+			(task.status === 'rate_limited' || task.status === 'usage_limited') &&
+			newStatus === 'in_progress'
+		) {
+			updates.restrictions = null;
 		}
 
 		// When transitioning FROM archived (unarchiving), clear the archived_at timestamp
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -24,6 +24,7 @@ import {
 	type RuntimeState,
 	type GlobalSettings,
 	type FallbackModelEntry,
+	type TaskRestriction,
 	MAX_CONCURRENT_GROUPS_LIMIT,
 	MAX_REVIEW_ROUNDS_LIMIT,
 } from '@neokai/shared';
@@ -718,6 +719,12 @@ export class RoomRuntime {
 						resetsAt: backoff.resetsAt,
 						sessionRole: 'worker',
 					});
+					await this.persistTaskRestriction(
+						group.taskId,
+						backoff,
+						'rate_limit',
+						`API rate limit (HTTP 429)`
+					);
 					this.scheduleTickAfterRateLimitReset(groupId);
 					// Try to switch to a fallback model if configured
 					await this.trySwitchToFallbackModel(groupId, group.workerSessionId, 'worker');
@@ -754,6 +761,12 @@ export class RoomRuntime {
 						resetsAt: backoff.resetsAt,
 						sessionRole: 'worker',
 					});
+					await this.persistTaskRestriction(
+						group.taskId,
+						backoff,
+						'usage_limit',
+						`Daily/weekly usage cap`
+					);
 					this.scheduleTickAfterRateLimitReset(groupId);
 					return;
 				}
@@ -1074,6 +1087,12 @@ export class RoomRuntime {
 						resetsAt: backoff.resetsAt,
 						sessionRole: 'leader',
 					});
+					await this.persistTaskRestriction(
+						group.taskId,
+						backoff,
+						'rate_limit',
+						`API rate limit (HTTP 429) in leader`
+					);
 					this.scheduleTickAfterRateLimitReset(groupId);
 					// Try to switch to a fallback model if configured
 					await this.trySwitchToFallbackModel(groupId, group.leaderSessionId, 'leader');
@@ -1106,6 +1125,12 @@ export class RoomRuntime {
 							resetsAt: backoff.resetsAt,
 							sessionRole: 'leader',
 						});
+						await this.persistTaskRestriction(
+							group.taskId,
+							backoff,
+							'usage_limit',
+							`Daily/weekly usage cap in leader`
+						);
 						this.scheduleTickAfterRateLimitReset(groupId);
 						return;
 					}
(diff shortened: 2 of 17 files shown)


lsm (commented):

---

### Summary

The implementation is solid overall. The core persistence and auto-resume logic is well-designed, the migration is idempotent and defensive, the status transition guard in `persistTaskRestriction` is a good safety net, and the 26 new unit tests all pass. I found one P1 issue (fire-and-forget promise in the real-time message listener) and several P2 issues worth addressing.

---

### P1: Fire-and-forget `persistTaskRestriction` in real-time message listener

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` line 2263

The call to `persistTaskRestriction` inside the `mirrorSession` message listener uses `void` (fire-and-forget):

```ts
void this.persistTaskRestriction(
    group.taskId,
    rateLimitBackoff,
    'rate_limit',
    `API rate limit (HTTP 429) in ${role}`
);
```

If this async operation fails (DB error, task deleted concurrently, etc.), the error is silently swallowed and the task status remains `in_progress` even though a rate limit was detected. This is the only call site that uses `void` -- all other call sites at lines 722, 764, 1090, and 1128 properly `await` the result.

While fire-and-forget is sometimes acceptable in event listeners, this one specifically writes state that the runtime tick depends on for auto-resume. A silent failure here means the task stays `in_progress` and could be picked up by `recoverStuckWorkers` and re-routed prematurely, only to hit the rate limit again.

**Recommendation:** Either await the call (if the enclosing handler can be made async), or add a `.catch()` that logs the error so it is at least observable.

---

### P2: Frontend components not updated for new statuses

Several frontend components use hardcoded status lists or switch statements that do not include `rate_limited` or `usage_limited`:

1. **`packages/web/src/components/room/task-shared/TaskActionDialogs.tsx` line 297:** `ALL_TASK_STATUSES` array only lists 8 statuses and does not include `rate_limited` or `usage_limited`. This means the "Set Task Status" modal dropdown will not show these as current statuses or as target statuses. A task in `rate_limited` status would display its current label correctly (line 317) but the dropdown would not include it as a selectable option.

2. **`packages/web/src/lib/room-store.ts` lines 196-214:** The orphan task buckets (`orphanTasksActive`, `orphanTasksReview`, `orphanTasksDone`, `orphanTasksArchived`) do not account for `rate_limited` or `usage_limited`. These tasks would fall through all buckets and become invisible in the UI task list. They should likely appear in the "active" bucket alongside `draft`, `pending`, and `in_progress`.

3. **`packages/web/src/components/room/RoomTasks.tsx` line 80:** `getStatusBorderColor` has no explicit case for `rate_limited` or `usage_limited`, falling through to `border-l-transparent`. These paused tasks would have no visible left border indicator.

4. **`packages/web/src/islands/RoomContextPanel.tsx` line 64:** `taskStatusColors` has no entry for `rate_limited` or `usage_limited`, falling through to the default `bg-gray-500` via the `??` fallback.

5. **`packages/shared/src/types/space.ts` line 119:** `SpaceTaskStatus` type does not include the new statuses. If space-level task rendering uses this type, rate/usage limited tasks would fail type checks or be silently excluded.

None of these are bugs in the backend logic, but they mean the new statuses are effectively invisible or broken in the frontend UI. Users would see tasks disappear from lists or show incorrect styling.

---

### P2: Existing test claims "all 8 TaskStatus values are covered"

**File:** `packages/web/src/lib/__tests__/room-store-computed-signals.test.ts` line 218

The test description says "all 8 TaskStatus values are covered" and line 248 asserts the total is 8. Now that `TaskStatus` has 10 values, this test would fail once the orphan bucket issue above is fixed (2 tasks would fall through all buckets). The test needs to be updated to include `rate_limited` and `usage_limited` and assert the total is 10.

---

### P2: No test for leader-rate-limit path

The runtime tests cover worker rate limits and worker usage limits, but there is no test for the leader rate limit or leader usage limit paths (lines 1090 and 1128 in room-runtime.ts). The leader paths have slightly different behavior (they log the role as "leader" and do not fall through to normal completion on fallback switch). Since these are separate code paths with different control flow, they deserve their own test coverage.

---

### P3: Minor observations

1. **`restrictions` field type uses `?` union with `null`:** The `NeoTask` interface declares `restrictions?: TaskRestriction | null`. The `rowToTask` method always returns either a parsed object or `null` (never `undefined`), so the `?` is technically unnecessary. This is a minor type inconsistency, not a bug.

2. **No JSON validation on read:** `rowToTask` does a raw `JSON.parse` on the restrictions column (line 381). If the stored JSON is malformed (e.g., from a partial write or manual DB edit), this would throw and potentially crash the row-to-task conversion. Other columns like `dependsOn` have the same pattern, so this is consistent with the codebase, but worth noting for defensive coding.

3. **Migration comment says "recreate the table" but first adds the column via ALTER:** The migration adds the column via `ALTER TABLE` first, then potentially recreates the table for the CHECK constraint. The ALTER is harmless but redundant if the table recreation runs (the new table already has the column). Not a bug, just slightly confusing.

---

### What looks good

- The `persistTaskRestriction` guard (`task.status !== 'in_progress'`) prevents overwriting terminal statuses -- well thought out.
- The migration is defensive: uses `tableHasColumn` and `schemaSql.includes('rate_limited')` guards, making it idempotent and safe to run multiple times.
- The dynamic column detection in migration 49 (`oldColumns.has(col)`) is a robust pattern for handling databases at different schema versions.
- The `clearTaskRestriction` method in `setTaskStatus` correctly limits the restriction clearing to the specific transition from limited states to `in_progress`.
- The `restrictions` field in `UpdateTaskParams` uses `!== undefined` check, so existing code paths that don't pass it won't accidentally clear it.
- Integration with `recoverStuckWorkers` is correct: active limits are skipped, expired limits trigger re-routing.

---

**Recommendation:** REQUEST_CHANGES (P1 fire-and-forget issue + P2 frontend gaps)

lsm (commented):

---

### Summary

All previously reported issues have been properly addressed. The implementation is solid, well-tested, and type-safe. No blocking issues remain.

---

### P1 Fix: Fire-and-forget promise in mirroring listener -- FIXED

The `void this.persistTaskRestriction(...)` call at `packages/daemon/src/lib/room/runtime/room-runtime.ts:2263` has been replaced with `this.persistTaskRestriction(...).catch(...)` that properly logs errors via `log.error`. This is the correct pattern since `setupMirroring` registers a synchronous event callback that cannot be made async. The error is not silently swallowed.

Note: All other call sites for `persistTaskRestriction` (in `onWorkerTerminalState` and `onLeaderTerminalState`) already use `await`, which is correct since those methods are async. Only the mirroring listener needed the fire-and-forget-with-catch treatment.

---

### P2: Frontend wiring -- COMPLETE

All identified components are now wired for `rate_limited` and `usage_limited`:

- **`TaskActionDialogs.tsx`**: Added to `ALL_TASK_STATUSES` array and `STATUS_LABELS` record -- users can manually set these statuses.
- **`RoomTasks.tsx`**: Added to `getStatusBorderColor` switch with `border-l-orange-500` / `border-l-orange-600`.
- **`RoomContextPanel.tsx`**: Added to `taskStatusColors` map with `bg-orange-500` / `bg-orange-600`.
- **`SpaceTaskPane.tsx`**: Added to both `STATUS_LABELS` and `STATUS_CLASSES` records.
- **`space-task-manager.ts`**: Added to `VALID_SPACE_TASK_TRANSITIONS` with `['in_progress', 'needs_attention', 'cancelled']`.
- **`space.ts`**: Added `rate_limited` and `usage_limited` to the `SpaceTaskStatus` union type.
- **`room-store.ts`**: Updated `orphanTasksActive` filter to include the new statuses in the active bucket.

---

### P2: Stale test count -- FIXED

Updated to "all 10 TaskStatus values" with proper assertions checking that all 10 statuses are covered and that `rate_limited` / `usage_limited` fall into the active bucket.

---

### P2: Missing leader path tests -- FIXED

Three new tests added to `room-runtime-rate-limit-persistence.test.ts`:

1. `leader rate_limit sets task to rate_limited` -- verifies status transition and restriction persistence with `sessionRole='leader'`.
2. `leader rate_limit persists restrictions with sessionRole=leader` -- verifies restriction fields including `resetAt`.
3. `leader usage_limit sets task to usage_limited when no fallback` -- verifies usage limit path with no fallback model configured.

All 29 tests across both test files pass (19 task-restrictions + 10 rate-limit-persistence).

---

- **`SpaceNavPanel.tsx` and `SpaceContextPanel.tsx`**: These use a `default` case for status dot colors, so rate_limited/usage_limited tasks get a generic gray dot rather than orange. Cosmetically these could be added for visual consistency, but this is purely cosmetic since the detailed `SpaceTaskPane` does show proper orange styling.
- **`TaskSummary` interface**: Does not include a `restrictions` field. This is intentional since summary objects are used for overview lists. The restrictions data is available via the full `NeoTask` when needed for detailed views. The frontend currently does not access `.restrictions` on any task object, which is appropriate for this initial implementation.

---

### Conclusion

The implementation is complete and well-structured. Type-checking passes, all 29 unit tests pass, and the codebase is consistent across backend types, database schema, migrations, and frontend rendering.

**Recommendation: APPROVE**

PR #811, opened 2026-03-24
feat(web): update task URL route patterns to accept short IDs

lsm:
## Summary
- Updated `ROOM_TASK_ROUTE_PATTERN` and `SPACE_TASK_ROUTE_PATTERN` in `packages/web/src/lib/router.ts` to accept both UUID and short ID (`[a-z]-\d+`, e.g. `t-42`) formats in the task ID segment
- Room, space, and session ID segments remain UUID-only (not user-facing short IDs in this milestone)
- Added 16 unit tests covering UUID and short ID matching for both `getRoomTaskIdFromPath` and `getSpaceTaskIdFromPath`

## Test plan
- [x] `/room/<uuid>/task/t-42` matches `ROOM_TASK_ROUTE_PATTERN`
- [x] `/room/<uuid>/task/<uuid>` still matches (backward compat)
- [x] `/space/<uuid>/task/t-42` matches `SPACE_TASK_ROUTE_PATTERN`
- [x] `/space/<uuid>/task/<uuid>` still matches (backward compat)
- [x] All 16 new unit tests pass
- [x] All 1225 existing lib tests pass
- [x] Pre-commit checks (lint, format, typecheck, knip) all pass

### DIFF
--- a/packages/web/src/lib/__tests__/router-short-id.test.ts
+++ b/packages/web/src/lib/__tests__/router-short-id.test.ts
@@ -52,6 +52,21 @@ describe('ROOM_TASK_ROUTE_PATTERN — short ID support', () => {
 		const result = getRoomTaskIdFromPath(`/room/${ROOM_UUID}/task/t-42/extra`);
 		expect(result).toBeNull();
 	});
+
+	it('returns null for zero counter (t-0 is semantically invalid)', () => {
+		const result = getRoomTaskIdFromPath(`/room/${ROOM_UUID}/task/t-0`);
+		expect(result).toBeNull();
+	});
+
+	it('returns null for leading-zero counter (t-01)', () => {
+		const result = getRoomTaskIdFromPath(`/room/${ROOM_UUID}/task/t-01`);
+		expect(result).toBeNull();
+	});
+
+	it('matches g- prefix short ID (route is intentionally permissive about prefix letter)', () => {
+		const result = getRoomTaskIdFromPath(`/room/${ROOM_UUID}/task/g-42`);
+		expect(result).toEqual({ roomId: ROOM_UUID, taskId: 'g-42' });
+	});
 });
 
 describe('SPACE_TASK_ROUTE_PATTERN — short ID support', () => {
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

The implementation is clean, focused, and correctly addresses the task. Two regex patterns were updated to accept short IDs, and 16 well-structured tests cover both positive and negative cases.

### Findings

**P2 -- Regex accepts t-0 (zero counter), which is semantically invalid**

The route pattern ([a-z]-\d+) accepts t-0 as a valid task ID in the URL. However, parseShortId() in packages/shared/src/utils.ts:14 explicitly rejects counters less than 1 (if counter < 1 return null). This means a user could navigate to /room/<uuid>/task/t-0, the route would match, but the backend would be unable to resolve it.

Consider tightening the route regex to ([a-z]-[1-9]\d*) to reject zero-prefixed or zero-only counters at the routing boundary, providing an earlier signal of an invalid URL. Alternatively, add a test case documenting that t-0 is accepted by the route (which is fine if the router intentionally delegates validation to the backend).

**P3 -- Tests only cover t- prefix; no coverage for g- prefix**

The short ID format supports both t- (task) and g- (goal) prefixes per SHORT_ID_PREFIX in packages/shared/src/utils.ts:1-4. The route regex ([a-z]-\d+) matches both. A test case confirming g-42 matches the task route pattern would document the deliberate permissiveness of the route.

**P3 -- No test for leading zeros (e.g. t-01)**

The route regex ([a-z]-\d+) would accept t-01, t-001, etc. Whether these are valid depends on how the backend handles them. A test case for this edge case would be valuable.

### What looks good

- The alternation ([a-f0-9-]+|[a-z]-\d+) correctly handles matching due to left-to-right alternation semantics combined with the ^ anchor
- Room/space/session ID segments remain UUID-only as specified
- All 16 new tests pass, and all 44 existing router tests continue to pass (no regressions)
- The regex pattern is consistent with parseShortId() in @neokai/shared
- Clean, minimal change -- only 2 lines modified in the source file

### Verdict

APPROVE -- The core implementation is correct and well-tested. The P2 item is a minor hardening suggestion, not a blocking issue.

lsm (commented):

### Re-review after feedback round

1. **P2 (regex accepts t-0) -- RESOLVED.** The regex was changed from `[a-z]-\\d+` to `[a-z]-[1-9]\\d*` in both `ROOM_TASK_ROUTE_PATTERN` and `SPACE_TASK_ROUTE_PATTERN` (lines 32 and 40 of `packages/web/src/lib/router.ts`). This rejects zero-counters at the routing boundary, consistent with `parseShortId()` in `packages/shared/src/utils.ts:14` which rejects `counter < 1`.

2. **P3 (no g- prefix test) -- RESOLVED.** Added test cases for both room and space routes verifying `g-42` matches, with a comment documenting the intentional permissiveness.

3. **P3 (no leading-zero test) -- RESOLVED.** Added test cases verifying both `t-0` and `t-01` return null for both room and space routes.

### Verification

- All 22 new tests pass (11 per describe block: 8 original + 3 new)
- All 91 existing router tests continue to pass with no regressions
- The regex `([a-f0-9-]+|[a-z]-[1-9]\\d*)` is functionally correct: the left branch (UUID) cannot match `t-42` because `t` is not in `[a-f]`, so the right branch (short ID) always handles the `t-` and `g-` prefixes. For prefixes in `[a-f]` (e.g., `a-5`), the left branch would match first, but the captured value is identical -- no downstream code distinguishes which branch matched.

### Minor note (P3 -- no action needed)

There is a minor asymmetry: `parseShortId('t-01')` returns `{prefix: 't', counter: 1}` (because `parseInt('01') = 1`), but the route regex rejects `t-01`. This is actually good defensive behavior -- the route is stricter about URL format, while the backend is lenient if a malformed value somehow gets through. No change needed.

### Verdict

**APPROVE** -- All previous feedback has been addressed. The regex is correct, well-tested, and consistent with the backend's `parseShortId()` semantics.

PR #812, opened 2026-03-24
feat(web): display short IDs in task cards with click-to-copy badge

lsm:
- Add ShortIdBadge component showing '#t-42' next to task title when
  shortId is present; absent when task has no shortId (graceful fallback)
- Clicking the badge copies the short ID to clipboard and shows '✓ copied'
  for 1.5s; stopPropagation prevents card navigation
- Badge has tooltip 'Click to copy short ID'
- Task navigation (onClick, onView, onReactivate) now prefers shortId over
  UUID so task detail URLs use the short form (e.g. /room/.../task/t-42)
- 7 new unit tests covering badge rendering, stopPropagation, tooltip,
  shortId-first navigation, and UUID fallback

### DIFF
--- a/packages/web/src/components/room/RoomTasks.test.tsx
+++ b/packages/web/src/components/room/RoomTasks.test.tsx
@@ -1054,4 +1054,97 @@ describe('RoomTasks', () => {
 			expect(badge?.getAttribute('title')).toBe('Mission: Specific Goal Name');
 		});
 	});
+
+	describe('Short ID Badge', () => {
+		beforeEach(() => {
+			selectedTabSignal.value = 'active';
+		});
+
+		it('should show short ID badge when task has shortId', () => {
+			const tasks = [createTask('uuid-123', 'pending', { shortId: 't-42' })];
+
+			const { container } = render(<RoomTasks tasks={tasks} />);
+
+			const badge = container.querySelector('[data-testid="short-id-badge-t-42"]');
+			expect(badge).toBeTruthy();
+			expect(badge?.textContent).toContain('#t-42');
+		});
+
+		it('should NOT show short ID badge when task has no shortId', () => {
+			const tasks = [createTask('uuid-123', 'pending')];
+
+			const { container } = render(<RoomTasks tasks={tasks} />);
+
+			// No badge elements with the short-id-badge prefix
+			const badges = container.querySelectorAll('[data-testid^="short-id-badge-"]');
+			expect(badges).toHaveLength(0);
+		});
+
+		it('should have tooltip "Click to copy short ID" on the badge', () => {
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean, focused PR that delivers exactly what was asked. The implementation is well-scoped and follows existing patterns in the codebase. I found no blocking issues.

### What was reviewed

Two files changed: `packages/web/src/components/room/RoomTasks.tsx` (+33/-4) and its test file (+93).

### Positive observations

- **Correct type usage**: `task.shortId ?? task.id` properly uses the `shortId?: string` field from `TaskSummary` (line 441 of `packages/shared/src/types/neo.ts`) with graceful fallback to UUID for legacy tasks.
- **Good `stopPropagation` hygiene**: The `ShortIdBadge` click handler calls `e.stopPropagation()` to prevent card navigation, consistent with the existing goal badge and PR link patterns in the same file.
- **Conditional rendering**: `{task.shortId && <ShortIdBadge .../>}` correctly hides the badge for legacy tasks that lack a `shortId`.
- **Test coverage is thorough**: 7 new tests cover badge visibility, absence when no `shortId`, tooltip text, `stopPropagation` on badge click, `onTaskClick` with `shortId`, UUID fallback, and `onView` with `shortId`. The clipboard mock approach in the `stopPropagation` test is appropriate.
- **Existing tests still valid**: Pre-existing tests for `onTaskClick`, `onView`, and `onReactivate` use tasks without `shortId`, so `navId` still resolves to `task.id` -- no test breakage.
- **Typecheck and lint pass cleanly**.

### Minor notes (non-blocking, P2/P3)

1. **(P2) No `.catch()` on clipboard promise** -- `navigator.clipboard.writeText(shortId).then(...)` at `packages/web/src/components/room/RoomTasks.tsx:549` has no error handler. If the Clipboard API throws (e.g., permission denied, insecure context), the promise silently rejects. However, this exactly mirrors the existing pattern in `AgentTurnBlock.tsx:139`, so it is a pre-existing codebase convention rather than a regression. The project also has a `copyToClipboard()` utility in `packages/web/src/lib/utils.ts` with a fallback path, but using it here would be an enhancement, not a fix.

2. **(P3) No `onReactivate` test with `shortId`** -- The new tests cover `onTaskClick` and `onView` with `shortId`, but not `onReactivate`. Since `onReactivate(navId)` uses the same `navId` derivation, this is extremely low risk, but a test confirming the reactivate handler also receives the short ID would close the loop.

3. **(P3) `setTimeout` not cleaned up on unmount** -- The 1500ms `setTimeout` in `ShortIdBadge` that resets `copied.value` could theoretically fire after the component unmounts. This is harmless (setting a Preact signal after unmount is a no-op), and the same pattern exists in `AgentTurnBlock.tsx:141`.

PR #815, opened 2026-03-24
docs: document Planner web search capability

lsm:
WebFetch and WebSearch are available to both the Planner agent
and its plan-writer sub-agent. This commit adds maintenance
documentation so the capability is visible and not accidentally
regressed.

Refs: task-61

### DIFF
--- a/docs/planner-web-search.md
+++ b/docs/planner-web-search.md
@@ -1,0 +1,94 @@
+# Planner Web Search Capability
+
+The Planner agent and its plan-writer sub-agent have built-in access to web search tools. This document records the current state to prevent accidental regression.
+
+## Current State
+
+Both agents have `WebFetch` and `WebSearch` in their active tool lists.
+
+### Planner Agent
+
+Defined in `packages/daemon/src/lib/room/agents/planner-agent.ts` — `createPlannerAgentInit()`:
+
+```typescript
+const plannerAgentDef: AgentDefinition = {
+    // ...
+    tools: [
+        'Task',
+        'TaskOutput',
+        'TaskStop',
+        'Read',
+        'Write',
+        'Edit',
+        'Bash',
+        'Grep',
+        'Glob',
+        'WebFetch',   // ✅ Present
+        'WebSearch',  // ✅ Present
+    ],
+    // ...
+};
+```
+
+### Plan-Writer Sub-Agent
+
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, accurate documentation-only change that fulfills the task requirements. I verified the claims against the actual source code and have a few minor observations.

### What's good

- The documentation accurately reflects the current state of `planner-agent.ts`. Both the Planner agent (lines 582-594) and the plan-writer sub-agent (lines 232-244) do include `WebFetch` and `WebSearch` in their `tools` arrays, exactly as documented.
- The explanation of how tools flow through `QueryOptionsBuilder.build()` (lines 72-80) is correct. The `roomAllowedBuiltinTools` block only applies when `session.type === 'room_chat'`, and planner sessions use `type: 'planner'`, so there is no interference.
- The maintenance note section (lines 82-94) provides clear, actionable guidance for future developers.
- No code changes -- documentation only, as required by the acceptance criteria.

1. **File path reference precision (line 78):** The doc states "roomAllowedBuiltinTools (which includes WebFetch and WebSearch) is merged into the top-level allowedTools, but this is additive -- it does not restrict agent-level tool choices." This is true, but the framing implies it could have affected the planner. A more precise statement would clarify that this block only runs for `room_chat` sessions (line 191 of `query-options-builder.ts`), and planner sessions (`type: 'planner'`) skip it entirely. This avoids giving future maintainers the impression that there is a close call or near-miss here.

2. **Missing disallowedTools check detail (line 80):** The doc says "No disallowedTools or allowedTools filter is applied to the Planner or plan-writer agent definitions." This is correct at the session level for planner sessions, but worth noting that `getDisallowedTools()` does add `kai__memory__*` by default (line 516 of `query-options-builder.ts`) -- which is unrelated to WebFetch/WebSearch but is a filter that does apply to all sessions. A more accurate phrasing: "No disallowedTools or allowedTools filter is applied that would block WebFetch or WebSearch for the Planner or plan-writer agent definitions."

3. **No link from related documentation:** The new doc is standalone in `docs/`. Consider adding a brief mention in `CLAUDE.md` under the Mission System or agent documentation sections, or linking from the plan file that defined this task (`docs/plans/.../06-planner-web-search.md`), so developers are more likely to discover it during normal workflows.

PR #817, opened 2026-03-24
feat(web): display short IDs in goals editor with click-to-copy badge

lsm:
Add GoalShortIdBadge component to GoalsEditor that renders a #g-N badge
next to goal titles when shortId is present. Badge copies the short ID
to clipboard on click with 1.5s visual feedback, uses stopPropagation
to prevent card expand toggle, and is omitted gracefully for legacy
goals without shortId.

Add 6 unit tests covering badge visibility, tooltip, stopPropagation,
clipboard copy, and fallback for goals without shortId.

### DIFF
--- a/packages/web/src/components/room/GoalsEditor.test.tsx
+++ b/packages/web/src/components/room/GoalsEditor.test.tsx
@@ -1003,4 +1003,72 @@ describe('GoalsEditor', () => {
 			expect(container.textContent).toContain('Deploy to staging');
 		});
 	});
+
+	describe('Short ID Badge', () => {
+		it('should show short ID badge when goal has shortId', () => {
+			const goals = [createMockGoal('goal-1', { shortId: 'g-7' })];
+			const { container } = render(<GoalsEditor goals={goals} {...defaultHandlers} />);
+			const badge = container.querySelector('[data-testid="goal-short-id-badge-g-7"]');
+			expect(badge).toBeTruthy();
+			expect(badge?.textContent).toBe('#g-7');
+		});
+
+		it('should not show short ID badge when goal has no shortId', () => {
+			const goals = [createMockGoal('goal-1')];
+			const { container } = render(<GoalsEditor goals={goals} {...defaultHandlers} />);
+			const badge = container.querySelector('[data-testid^="goal-short-id-badge-"]');
+			expect(badge).toBeNull();
+		});
+
+		it('should show tooltip on short ID badge', () => {
+			const goals = [createMockGoal('goal-1', { shortId: 'g-3' })];
+			const { container } = render(<GoalsEditor goals={goals} {...defaultHandlers} />);
+			const badge = container.querySelector('[data-testid="goal-short-id-badge-g-3"]');
+			expect(badge?.getAttribute('title')).toBe('Click to copy short ID');
+		});
+
+		it('should stop propagation when clicking badge', () => {
+			const goals = [createMockGoal('goal-1', { shortId: 'g-5' })];
+			const { container } = render(<GoalsEditor goals={goals} {...defaultHandlers} />);
+			const badge = container.querySelector(
+				'[data-testid="goal-short-id-badge-g-5"]'
+			) as HTMLElement;
+
+			// Track expand state — clicking header toggles expand; clicking badge should not
+			const header = container.querySelector('[data-testid="goal-item-header"]') as HTMLElement;
+			fireEvent.click(header);
+			const expandedContent = container.innerHTML;
+
+			// Click badge — expansion state should not change again
+			fireEvent.click(badge);
+			expect(container.innerHTML).toBe(expandedContent);
+		});
+
+		it('should copy short ID to clipboard on click', async () => {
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Verdict: APPROVE

The implementation is clean, consistent with the existing ShortIdBadge in RoomTasks.tsx, and all 89 tests pass (including 6 new ones). Lint, typecheck, and knip are all clean.

---

### Detailed Review

**Correctness & Completeness**

The GoalShortIdBadge component faithfully mirrors the ShortIdBadge in `packages/web/src/components/room/RoomTasks.tsx` (lines 544-567). The implementations are structurally identical:
- Same `useSignal(false)` for the copied state
- Same `e.stopPropagation()` to prevent card click-through
- Same `navigator.clipboard.writeText(shortId)` with `.then()` callback
- Same 1500ms reset timeout
- Same CSS classes, tooltip text, and visual output

The badge is correctly rendered inside the goal card header (line 1289-1292 of GoalsEditor.tsx), wrapped in a flex container with the title, and guarded by a `{goal.shortId && ...}` check so legacy goals without shortId are handled gracefully. The `shortId?` field on `RoomGoal` is already defined in `packages/shared/src/types/neo.ts:150`.

**Test Coverage (P2 -- see note below)**

The 6 new tests cover:
1. Badge visibility when shortId is present
2. No badge when shortId is absent
3. Tooltip text verification
4. stopPropagation prevents expand toggle
5. Clipboard copy with correct argument
6. Fallback rendering for goals without shortId

However, tests 2 and 6 cover the same scenario (goal without shortId). The test suite would benefit from a test verifying the visual feedback transition (checking that `copied.value = true` changes the button text from `#g-N` to the checkmark text), similar to how the RoomTasks tests verify click behavior. This is not blocking since the underlying logic is identical to the already-validated RoomTasks badge.

**P1: Missing .catch() on clipboard API**

Both `GoalShortIdBadge` (GoalsEditor.tsx:269) and the original `ShortIdBadge` (RoomTasks.tsx:549) call `navigator.clipboard.writeText(shortId).then(...)` without a `.catch()`. If the Clipboard API is unavailable (e.g., HTTP context without secure origin, or browser permission denied), this will produce an unhandled promise rejection. This is a pre-existing issue in the RoomTasks implementation that was copied over, not a new bug introduced by this PR. Flagging for awareness but not blocking this PR since fixing it here alone would be inconsistent.

**Consistency**

The `data-testid` prefix is `goal-short-id-badge-` versus `short-id-badge-` in RoomTasks. This is reasonable since it disambiguates which component's badge is being targeted in tests and follows a natural naming convention.

**Layout**

The badge is placed in a new `<div class="flex items-center gap-2 min-w-0">` wrapper around the title (line 1289). The `min-w-0` ensures the title can truncate properly when it's long. The badge has `flex-shrink-0` so it never gets squeezed. This is correct.

---

### Summary

This is a straightforward, well-executed copy of the existing ShortIdBadge pattern applied to the goals editor. The code is consistent, tests pass, and there are no regressions.

### Issues Found

- **P1**: Missing `.catch()` on `navigator.clipboard.writeText()` promise (pre-existing from RoomTasks, not new)
- **P2**: Test for visual feedback transition (copied state) is absent, though covered indirectly by identical RoomTasks pattern

### Counts

- P0: 0
- P1: 1
- P2: 1
- P3: 0

**Recommendation: APPROVE** -- The missing .catch() is pre-existing and out of scope for this PR. The test gap is minor.

lsm (commented):

### Verdict: APPROVE

Both review items from the previous round have been correctly addressed.

---

### P1 Fix: `.catch(() => {})" on clipboard writeText

Both components now have the error suppression:

- `GoalShortIdBadge` (`packages/web/src/components/room/GoalsEditor.tsx`, line 277): `.catch(() => {})` added
- `ShortIdBadge` (`packages/web/src/components/room/RoomTasks.tsx`, line 557): `.catch(() => {})` added

The two implementations are now structurally identical. The empty catch handler is the correct approach here -- clipboard failures are non-critical UX, and showing an error toast would be excessive for a copy-to-clipboard action. If you ever want to log clipboard failures for diagnostics, this is the place to add it, but silence is the right default.

### P2 Fix: Visual feedback test

The new test at `packages/web/src/components/room/GoalsEditor.test.tsx:1065-1084` correctly verifies:

1. Initial badge text is `#g-11`
2. After clicking, badge text transitions to the checkmark string
3. Uses `waitFor` from `@testing-library/preact` to handle the async state update

The test is well-structured -- it mocks `navigator.clipboard.writeText` as a resolving promise and waits for the signal update to propagate. All 90 tests pass (84 existing + 7 new badge tests).

### Minor note (P3)

The RoomTasks counterpart (`ShortIdBadge`) still lacks a visual feedback test. This is not blocking since the two components are structurally identical, but for completeness it would be nice to add the same test there in a follow-up.

### Issues Found

- P0: 0
- P1: 0
- P2: 0
- P3: 1 (visual feedback test missing from RoomTasks -- follow-up only)

**Recommendation: APPROVE**

PR #818, opened 2026-03-24
feat(storage): add app_mcp_servers table, types, and repository

lsm:
- Add `app_mcp_servers` SQLite table to schema/index.ts with columns for
  stdio/sse/http transport types, JSON-encoded args/env/headers, and enabled flag
- Add migration 49 to create the table incrementally on existing databases
- Add AppMcpServer types to @neokai/shared (AppMcpServer, AppMcpServerSourceType,
  CreateAppMcpServerRequest, UpdateAppMcpServerRequest)
- Implement AppMcpServerRepository with create/get/getByName/list/listEnabled/update/delete;
  each write calls reactiveDb.notifyChange('app_mcp_servers') for LiveQuery invalidation
- Wire repository into Database facade as `appMcpServers` getter
- Add 23 unit tests covering CRUD, listEnabled filtering, and notifyChange calls

### DIFF
--- a/packages/daemon/src/storage/index.ts
+++ b/packages/daemon/src/storage/index.ts
@@ -33,6 +33,7 @@ import {
 	type UpdateGoalParams,
 } from './repositories/goal-repository';
 import { JobQueueRepository } from './repositories/job-queue-repository';
+import { AppMcpServerRepository } from './repositories/app-mcp-server-repository';
 import type { ReactiveDatabase } from './reactive-database';
 
 export type { SendStatus } from './repositories/sdk-message-repository';
@@ -53,6 +54,7 @@ export type { JobHandler, JobQueueProcessorOptions } from './job-queue-processor
 // Re-export repository classes for direct use
 export { GoalRepository } from './repositories/goal-repository';
 export { SpaceAgentRepository } from './repositories/space-agent-repository';
+export { AppMcpServerRepository } from './repositories/app-mcp-server-repository';
 
 /**
  * Database facade class that maintains backward compatibility with the original Database class.
@@ -69,6 +71,7 @@ export class Database {
 	private inboxItemRepo!: InboxItemRepository;
 	private goalRepo!: GoalRepository;
 	private jobQueueRepo!: JobQueueRepository;
+	private appMcpServerRepo!: AppMcpServerRepository;
 	private shortIdAllocator!: ShortIdAllocator;
 
 	constructor(dbPath: string) {
@@ -89,6 +92,7 @@ export class Database {
 		this.inboxItemRepo = new InboxItemRepository(db);
 		this.goalRepo = new GoalRepository(db, reactiveDb, shortIdAllocator);
 		this.jobQueueRepo = new JobQueueRepository(db);
+		this.appMcpServerRepo = new AppMcpServerRepository(db, reactiveDb);
 	}
 
 	// ============================================================================
@@ -423,6 +427,13 @@ export class Database {
 	 */
 	getJobQueueRepo(): JobQueueRepository {
 		return this.jobQueueRepo;
+	}
+
+	/**
+	 * Get the application-level MCP server repository
+	 */
+	get appMcpServers(): AppMcpServerRepository {
+		return this.appMcpServerRepo;
 	}
 
 	close(): void {
--- a/packages/daemon/src/storage/repositories/app-mcp-server-repository.ts
+++ b/packages/daemon/src/storage/repositories/app-mcp-server-repository.ts
@@ -1,0 +1,214 @@
+/**
+ * AppMcpServerRepository
+ *
+ * CRUD operations for the application-level MCP server registry.
+ * Each write method calls reactiveDb.notifyChange('app_mcp_servers') so that
+ * LiveQueryEngine can invalidate frontend subscriptions on every registry change.
+ */
+
+import type { Database as BunDatabase } from 'bun:sqlite';
+import { generateUUID } from '@neokai/shared';
+import type {
+	AppMcpServer,
+	AppMcpServerSourceType,
+	CreateAppMcpServerRequest,
+	UpdateAppMcpServerRequest,
+} from '@neokai/shared';
+import type { ReactiveDatabase } from '../reactive-database';
+import type { SQLiteValue } from '../types';
+
+// ---------------------------------------------------------------------------
+// Internal row type (mirrors SQLite columns)
+// ---------------------------------------------------------------------------
+
+interface AppMcpServerRow {
+	id: string;
+	name: string;
+	description: string | null;
+	source_type: string;
+	command: string | null;
+	args: string | null;
+	env: string | null;
+	url: string | null;
+	headers: string | null;
+	enabled: number;
+	created_at: number | null;
+	updated_at: number | null;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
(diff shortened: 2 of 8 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** (4 issues found: 1 P1, 3 P2)

---

### P1 -- Missing UNIQUE constraint error handling on `create()` and `update()` for the `name` column

**File:** `packages/daemon/src/storage/repositories/app-mcp-server-repository.ts`

The `name` column has a `UNIQUE` constraint in the schema, but neither `create()` nor `update(name=...)` handles the case where a duplicate name is provided. Both methods will let the raw SQLite `UNIQUE constraint failed` error bubble up as an unhandled exception. Other repositories in this codebase (e.g., `SpaceAgentRepository` in `packages/daemon/src/storage/repositories/space-agent-repository.ts`, line 88-104) explicitly provide an `isNameTaken()` check method for exactly this scenario.

At minimum, `create()` should either:
- Catch the constraint error and throw a domain-specific error (e.g., a custom `DuplicateNameError` or similar), or
- Check for duplicate names before insertion and return a structured error, or
- Have a companion `isNameTaken(name, excludeId?)` method so that the RPC handler layer (to be written in a future task) can validate before calling the repo.

The same applies to `update()` when `name` is changed -- the UNIQUE constraint can fire when renaming to an already-taken name.

Additionally, there are no tests for the duplicate name scenario. Given that this is the primary integrity constraint on the table, this is a notable test gap.

---

### P2 -- `CreateAppMcpServerRequest` requires `enabled: boolean` with no default

**File:** `packages/shared/src/types/app-mcp-server.ts`, line 53

```
export type CreateAppMcpServerRequest = Omit<AppMcpServer, 'id'>;
```

Since `enabled: boolean` is a required (non-optional) field on `AppMcpServer`, it is also required on `CreateAppMcpServerRequest`. While the SQL schema defines `DEFAULT 1`, the TypeScript type forces every caller to explicitly pass `enabled: true`. This is inconsistent -- every single test has to redundantly specify `enabled: true`, and the eventual RPC handler will need to mandate it from the frontend.

Consider making `enabled` optional with a default of `true` in the create request type (e.g., `Partial<Pick<AppMcpServer, 'enabled'>>` overlay, or a separate interface). The repository's `create()` method already handles falsy values with `req.enabled ? 1 : 0`, so it is ready for this -- but the type layer forces explicitness that the DB schema does not.

---

### P2 -- Missing validation on `source_type` transport compatibility

**File:** `packages/daemon/src/storage/repositories/app-mcp-server-repository.ts`

The repository accepts any `sourceType` value and stores it directly. While the DB has a CHECK constraint (`CHECK(source_type IN ('stdio', 'sse', 'http'))`), there is no application-level validation. A caller could pass an invalid sourceType, get an unhelpful SQLite CHECK constraint error, and there is no test verifying that invalid source types are rejected.

While the CHECK constraint prevents data corruption, it would be better to validate at the repository boundary and return a clear error. Consider at least a test that verifies the CHECK constraint fires on invalid input.

---

### P2 -- `list()` and `listEnabled()` ordering with NULL `created_at`

**File:** `packages/daemon/src/storage/repositories/app-mcp-server-repository.ts`, lines 127 and 137

Both queries use `ORDER BY created_at ASC`. In SQLite, NULL values sort first in ASC order. While `create()` always sets `created_at`, the column is nullable in the schema (`created_at INTEGER` -- no `NOT NULL` constraint). If rows are ever inserted outside this repository (direct SQL, data migration, or a bug), those rows would appear first in the list with `created_at = null`.

This is not a bug today, but it is a fragility. Consider either:
- Adding `NOT NULL` to `created_at` in the schema (preferred, since the repository always sets it), or
- Using `ORDER BY created_at IS NULL, created_at ASC` to push NULLs to the end.

---

### Positive observations

- **Schema and migration are correct and consistent.** Both `createTables()` and `runMigration49()` define identical DDL. The migration is idempotent via `CREATE TABLE IF NOT EXISTS`.
- **`notifyChange('app_mcp_servers')` usage is correct.** Every write method (create, update, delete) properly calls `notifyChange` after the operation. The conditional guards (only notify when changes actually occurred in update/delete) are a nice touch.
- **The `rowToServer()` helper correctly handles null-to-undefined mapping** for optional fields using spread + conditional, which matches the existing codebase pattern.
- **The update method correctly distinguishes between "field not provided" and "field explicitly set to null/undefined"** using `'description' in updates` pattern for nullable fields and `updates.name !== undefined` for non-nullable fields.
- **Wiring into the Database facade is clean and follows the existing pattern** -- private field, initialization in `initialize()`, exposed via getter.
- **Test coverage is solid for the happy path** -- 23 tests covering all CRUD operations, all three source types, optional field omission, and notifyChange spy verification.
- **The shared types are well-documented** with clear JSDoc comments including the env var security note.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been addressed correctly and completely. No new issues were introduced.

### Issue Resolution Summary

**P1 (Duplicate name handling) -- Fixed**
- `isNameTaken(name, excludeId?)` is properly implemented with a SELECT 1 query (line 88-97 of `app-mcp-server-repository.ts`). The `excludeId` parameter correctly uses `id != ?` to allow self-renames.
- `create()` (line 106) and `update()` (line 188) both check for duplicate names before proceeding, throwing a clear error message. The DB UNIQUE constraint on `name` serves as a safety net.
- The pattern is consistent with the existing `SpaceAgentRepository.isNameTaken()` in this codebase.
- Tests cover: duplicate create, duplicate rename, self-rename, and the `isNameTaken` helper directly with all four permutations (line 113-158 of the test file).

**P2 (enabled defaulting) -- Fixed**
- `CreateAppMcpServerRequest` now uses `enabled?: boolean` (line 53-55 of `app-mcp-server.ts`), making it truly optional.
- `create()` defaults via `req.enabled ?? true` (line 129 of the repository).
- Tests verify default-to-true when omitted and explicit false (lines 67-75 of the test file).
- All existing tests were updated to remove the now-redundant `enabled: true`.

**P2 (sourceType validation) -- Fixed**
- `VALID_SOURCE_TYPES` constant (line 24) and `validateSourceType()` (lines 66-72) validate against the set of known transport types.
- Called in both `create()` (line 104) and `update()` (line 192-194).
- Error message includes the invalid value and lists valid options.
- Tests verify invalid sourceType is rejected on both create and update (lines 120-125, 305-311 of the test file).
- Note: the DB CHECK constraint on `source_type` is still the authoritative guard; the application-level validation provides a friendlier error message.

**P2 (NULL created_at ordering) -- Fixed**
- Both `list()` and `listEnabled()` now use `ORDER BY created_at IS NULL, created_at ASC` (lines 163, 173-174), which sorts NULL rows last.
- Test verifies NULL-created_at rows sort after timestamped rows (lines 217-228 of the test file).

### Minor Observations (no action required)

- **P3: No input sanitization on `name`** -- The repository does not trim whitespace or validate minimum length on the `name` field. This is acceptable as validation is typically the responsibility of the RPC handler layer (which this PR does not include). The DB-level constraints (`UNIQUE NOT NULL`) handle the hard requirements.

- **P3: No empty/whitespace name test** -- Tests do not verify behavior when name is empty string or whitespace-only. This is fine for a repository unit test; boundary validation should be tested at the handler layer.

### Test Coverage
All 35 tests pass. Coverage is comprehensive for the repository layer: CRUD operations, list/listEnabled filtering, NULL ordering, duplicate name detection, sourceType validation, and notifyChange behavior (including the no-op case for empty updates).

PR #820, opened 2026-03-24
feat(mcp): add AppMcpLifecycleManager class and wire mcp.registry.listErrors RPC

lsm:
- Create packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts with:
  - getEnabledMcpConfigs(): converts enabled registry entries to SDK McpServerConfig
  - getEnabledMcpConfigsForRoom(): stub that falls back to global enabled configs
  - validateEntry(): checks required fields per source type (stdio needs command, sse/http need url)
  - getStartupErrors(): returns all invalid entries for UI warning badge
- Create packages/daemon/src/lib/mcp/index.ts re-exporting the manager
- Instantiate AppMcpLifecycleManager in app.ts and expose as appMcpManager in DaemonAppContext
- Register mcp.registry.listErrors RPC handler in mcp-handlers.ts
- Add 19 unit tests covering stdio/sse/http conversion, disabled filtering, validation, and error reporting

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -24,6 +24,7 @@ import { JobQueueRepository } from './storage/repositories/job-queue-repository'
 import { JobQueueProcessor } from './storage/job-queue-processor';
 import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
 import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
+import { AppMcpLifecycleManager } from './lib/mcp';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -71,6 +72,8 @@ export interface DaemonAppContext {
 	jobQueue: JobQueueRepository;
 	/** Persistent job queue processor */
 	jobProcessor: JobQueueProcessor;
+	/** Application-level MCP lifecycle manager — converts registry entries to SDK configs */
+	appMcpManager: AppMcpLifecycleManager;
 	/**
 	 * Cleanup function for graceful shutdown.
 	 * Closes all connections, stops sessions, and closes database.
@@ -260,6 +263,9 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		logInfo('[Daemon] GitHub integration disabled - authentication required');
 	}
 
+	// Instantiate application-level MCP lifecycle manager
+	const appMcpManager = new AppMcpLifecycleManager(db);
+
 	// Setup RPC handlers (returns cleanup function + exposed services)
 	const {
 		cleanup: rpcHandlerCleanup,
@@ -280,6 +286,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		jobProcessor,
 		reactiveDb,
 		liveQueries,
+		appMcpManager,
 	});
 
 	// Create WebSocket handlers
@@ -531,6 +538,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		taskAgentManager,
 		jobQueue,
 		jobProcessor,
+		appMcpManager,
 		cleanup,
 	};
 }
--- a/packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts
+++ b/packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts
@@ -16,22 +16,21 @@
  */
 
 import type { Database } from '../../storage/database';
-import type { AppMcpServer } from '@neokai/shared';
 import type {
+	AppMcpServer,
 	McpServerConfig,
 	McpStdioServerConfig,
 	McpSSEServerConfig,
 	McpHttpServerConfig,
+	ValidationResult,
 } from '@neokai/shared';
+
+// Re-export so callers can import from this module without reaching into shared.
+export type { ValidationResult } from '@neokai/shared';
 
 // ---------------------------------------------------------------------------
 // Types
 // ---------------------------------------------------------------------------
-
-export interface ValidationResult {
-	valid: boolean;
-	error?: string;
-}
 
 export interface McpStartupError {
 	serverId: string;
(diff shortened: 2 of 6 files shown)


lsm (commented):

---

### Summary

This PR implements the `AppMcpLifecycleManager` class and wires the `mcp.registry.listErrors` RPC handler. All 19 unit tests pass, typecheck is clean, and lint has zero warnings. The implementation matches the Task 3.1 spec precisely. Below are the findings.

---

### P2: Duplicate `ValidationResult` type should reuse the shared export

**File:** `packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts` (line 31)
**File:** `packages/daemon/src/lib/mcp/index.ts` (line 2)

The manager defines its own `ValidationResult` interface:

```ts
export interface ValidationResult {
    valid: boolean;
    error?: string;
}
```

An identical `ValidationResult` already exists in `packages/shared/src/types/sdk-config.ts` (line 396) and is re-exported from `@neokai/shared`. The `config-validators.ts` module already uses this shared type extensively. Defining a second copy creates ambiguity -- a consumer importing from `../mcp` gets a different `ValidationResult` than one importing from `@neokai/shared`, even though they are structurally identical. This could cause subtle bugs if the shared type is ever extended (e.g., adding a `warnings` field) and the local copy is not updated in sync.

**Recommendation:** Import `ValidationResult` from `@neokai/shared` instead of defining it locally, and remove it from the local re-export in `index.ts`.

---

### P2: `convertEntry` default branch silently swallows unknown source types

**File:** `packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts` (lines 189-191)

The `validateEntry` method has a proper exhaustive check (`const exhaustive: never = entry.sourceType`), but `convertEntry` uses a bare `default: return null`:

```ts
default:
    return null;
```

Since `validateEntry` is called before `convertEntry` in `getEnabledMcpConfigs`, the `null` return should never be hit for entries that pass validation. However, if `convertEntry` is ever called directly or the call ordering changes, an unknown source type would silently produce `null` with no compiler warning. The `never` exhaustiveness guard in `validateEntry` is good -- adding the same pattern to `convertEntry` would make the class internally consistent and catch future source type additions at compile time.

---

### P3 (nit): Test could use the shared `createTestDb()` helper

**File:** `packages/daemon/tests/unit/mcp/app-mcp-lifecycle-manager.test.ts` (line 24)

The test file defines its own `createTestDb()` function that manually constructs a `Database` facade stub. A `createTestDb()` helper already exists in `packages/daemon/tests/helpers/database.ts` that creates a real in-memory `Database` instance via `new Database(':memory:')` followed by `initialize()`. Since the manager only uses `db.appMcpServers`, the real `Database` class (which exposes `appMcpServers` via a getter) would be a cleaner fit and eliminates the `as unknown as Database` cast.

---

### P3 (nit): Empty env/args/headers objects produce empty spread (`{}`) on stdio config

**File:** `packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts` (lines 161-163)

When `entry.args` is an empty array `[]` or `entry.env` is an empty object `{}`, the conditional spreads produce an empty object:

```ts
...(entry.args && entry.args.length > 0 ? { args: entry.args } : {}),
```

This is correct behavior (empty arrays/objects are omitted from the output config), but the spread of `{}` is a minor code smell. An alternative would be using Object.fromEntries or conditional property assignment. Not a functional issue.

---

### What is done well

- **Exhaustive type checking** in `validateEntry` ensures compile-time safety if new source types are added to `AppMcpServerSourceType`.
- **Clean separation** between validation and conversion (`validateEntry` vs `convertEntry`).
- **Proper wiring** through the full stack: `app.ts` -> `RPCHandlerDependencies` -> `registerMcpHandlers` -> `mcp.registry.listErrors`.
- **Documentation** is thorough -- the module-level docstring explains the scope and defers future work explicitly.
- **Test coverage** hits all public methods and key edge cases (disabled filtering, empty-whitespace command, disabled-but-invalid entries included in error list).
- **Security**: The `env` field on stdio servers correctly passes through as-is (the AppMcpServer type docs clarify that values are reference keys, not raw secrets). No credential leak risk.

---

The two P2 issues (duplicate type, missing exhaustiveness in `convertEntry`) should be addressed before merge. Both are low-effort fixes.

lsm (commented):

### Summary

Both P2 issues from the previous review have been correctly addressed. The implementation is clean, well-tested, and integrates properly with the existing codebase.

### P2 Fixes Verified

**1. Duplicate `ValidationResult` type — fixed**

The local duplicate has been removed. `ValidationResult` is now imported as a type-only import from `@neokai/shared` (line 26 of `app-mcp-lifecycle-manager.ts`) and re-exported on line 29 for caller convenience. This matches the canonical definition in `packages/shared/src/types/sdk-config.ts:396` (`{ valid: boolean; error?: string }`). The re-export through `packages/daemon/src/lib/mcp/index.ts:2` is also correct.

**2. Missing exhaustiveness check in `convertEntry` — fixed**

The `default` branch in `convertEntry` (lines 185-188) now uses the `never`-guard pattern:

```ts
default: {
    const exhaustive: never = entry.sourceType;
    throw new Error(`convertEntry: unhandled sourceType "${exhaustive}"`);
}
```

The return type of `convertEntry` is now non-nullable `McpServerConfig` (not `McpServerConfig | null`). The caller in `getEnabledMcpConfigs` only calls `convertEntry` after validation passes, so the non-nullable return is sound. Note that `validateEntry` also has its own `never` guard (lines 113-119) that returns a `ValidationResult` rather than throwing — this is intentional since validation should gracefully handle unknown types rather than crash.

### Verification

- All 19 unit tests pass (0 failures, 48 expect() calls)
- TypeScript type checking passes cleanly (`bun run typecheck`)
- Lint passes with 0 warnings/errors
- Knip reports no dead exports from the new module
- The `AppMcpServerSourceType` union (`'stdio' | 'sse' | 'http'`) covers all switch cases in both methods
- The `McpStdioServerConfig.type` field is optional in shared types, so explicitly setting `type: 'stdio'` is compatible
- `appMcpManager` is correctly threaded through `RPCHandlerDependencies` → `registerMcpHandlers` → the `mcp.registry.listErrors` RPC handler
- Instantiation order in `app.ts` is correct: `AppMcpLifecycleManager` is created before `setupRPCHandlers`, and included in both `DaemonAppContext` construction and the return object

### Recommendation

**APPROVE** — Both P2 issues are resolved correctly. No new issues introduced.

PR #821, opened 2026-03-24
test(e2e): add short ID display and copy test; fix LiveQuery SQL missing short_id

lsm:
- Add packages/e2e/tests/features/short-id-display.e2e.ts with 3 tests:
  1. Short ID badge (#t-1) appears in task card
  2. Clicking badge copies short ID to clipboard
  3. Navigating to /room/{roomId}/task/t-1 loads task detail page

- Fix TASKS_BY_ROOM_SQL in live-query-handlers.ts to include `short_id AS shortId`
  so the short ID is delivered to the frontend via LiveQuery events (was missing,
  causing the badge to never render despite shortId being stored in the DB)

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts
@@ -168,7 +168,8 @@ SELECT
   pr_number           AS prNumber,
   pr_created_at       AS prCreatedAt,
   input_draft         AS inputDraft,
-  updated_at          AS updatedAt
+  updated_at          AS updatedAt,
+  short_id            AS shortId
 FROM tasks
 WHERE room_id = ?
 ORDER BY created_at DESC, id DESC
--- a/packages/e2e/tests/features/short-id-display.e2e.ts
+++ b/packages/e2e/tests/features/short-id-display.e2e.ts
@@ -1,0 +1,143 @@
+/**
+ * Short ID Display and Copy E2E Tests
+ *
+ * Verifies the short ID badge behavior in the task card UI:
+ * 1. Short ID badge (#t-1) appears in the task card after task creation
+ * 2. Clicking the badge copies the short ID string to the clipboard
+ * 3. Navigating to /room/{roomId}/task/t-1 loads the task detail page
+ *
+ * Setup: creates a room and task via RPC in beforeEach (accepted infrastructure
+ *        pattern — room.create and task.create are both accepted for test isolation).
+ * Cleanup: deletes the room via RPC in afterEach.
+ *
+ * All test actions go through the UI (navigation, clicks, assertions on DOM).
+ * No direct RPC calls in test bodies.
+ */
+
+import { test, expect } from '../../fixtures';
+import { waitForWebSocketConnected } from '../helpers/wait-helpers';
+import { deleteRoom } from '../helpers/room-helpers';
+
+// ─── RPC Infrastructure Helpers ───────────────────────────────────────────────
+
+/**
+ * Create a room via RPC. For use in beforeEach setup only.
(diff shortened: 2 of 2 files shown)


lsm (commented):

---

### Summary

This PR adds an E2E test for the short ID badge feature and fixes a real bug in TASKS_BY_ROOM_SQL that caused shortId to never be delivered to the frontend via LiveQuery. The implementation is clean and follows existing patterns well.

**Recommendation: APPROVE**

---

### Bug Fix: TASKS_BY_ROOM_SQL missing short_id (correct)

**File:** packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts (line 172)

The fix adds short_id AS shortId to the explicit column list in TASKS_BY_ROOM_SQL. This is correct: the LiveQuery system delivers task rows to the frontend via this SQL, and without the column, shortId was always undefined in the LiveQuery snapshot/delta events. The TaskRepository internally uses SELECT * which includes short_id, so the direct RPC paths (task.create, task.get, task.list) were fine — only the LiveQuery path was broken.

No side effects: the column is added at the end of the SELECT list with a trailing comma (consistent with the rest), and the mapTaskRow mapper just spreads the row so no mapping changes are needed.

**Note for follow-up:** GOALS_BY_ROOM_SQL (line 178-207) has the same bug — it also omits short_id AS shortId. The goals table has a short_id column, and GoalsEditor.tsx renders a GoalShortIdBadge that depends on goal.shortId. This means goal short IDs are also broken in the LiveQuery path. Consider fixing this in a separate PR.

---

### E2E Test: short-id-display.e2e.ts

**File:** packages/e2e/tests/features/short-id-display.e2e.ts

**What is good:**

- Follows E2E test rules correctly: all test actions are through the UI (navigation, clicks, DOM assertions). RPC is used only in beforeEach for room/task creation and in afterEach for cleanup.
- Uses the shared deleteRoom helper from room-helpers.ts for cleanup.
- Uses data-testid selectors for the badge, matching the frontend implementation in RoomTasks.tsx (line 562).
- Clipboard permission handling via context.grantPermissions is correct.
- The copied confirmation assertion has a 2000ms timeout, which is adequate for the component's 1500ms display window.

**Minor observations (P3 — non-blocking):**

1. **Unused taskId variable** (line 64): taskId is declared, assigned in beforeEach (line 76), and cleared in afterEach (line 82), but never referenced in any test body. This is dead code. Consider removing it to reduce noise.

2. **Duplicate createRoom helper** (lines 26-34): The test defines a local createRoom function that duplicates the shared createRoom from room-helpers.ts (which is already imported for deleteRoom). The shared version takes a name parameter. While other test files in the codebase also define local helpers, using the shared helper with the name passed in would reduce duplication.

3. **Variable reset in afterEach vs beforeEach** (lines 79-83): This test resets roomId, taskId, and shortId to empty strings in afterEach. The established convention in sibling tests resets them at the start of beforeEach. Functionally equivalent since deleteRoom guards against empty strings, but inconsistent with the codebase pattern.

4. **Clipboard read timing** (line 126): The clipboard read immediately after badge.click() relies on the write completing within the microtask queue. Works reliably in practice with Playwright permissions, but is worth noting.

---

### Test Coverage Assessment

All three acceptance criteria are covered:

1. Short ID badge appears in the task card — verified by test 1
2. Clicking the badge copies short ID to clipboard — verified by test 2
3. Navigating to /room/{roomId}/task/t-1 loads task detail — verified by test 3

The tests are well-scoped and the router already supports short ID path segments (confirmed in router-short-id.test.ts).

---

### Conclusion

Clean, well-documented implementation with a genuine bug fix. The E2E tests follow project conventions and adequately cover the acceptance criteria. All issues noted are P3/non-blocking.

PR #822, opened 2026-03-24
feat(rpc): add MCP registry CRUD handlers and mcpServers.global LiveQuery

lsm:
- Add 6 RPC handlers in app-mcp-handlers.ts: mcp.registry.list/create/update/delete/setEnabled/listErrors
- Emit mcp.registry.changed (daemon-internal) on every write for hot-reload wiring (Task 3.2)
- Register mcpServers.global named query in NAMED_QUERY_REGISTRY for frontend LiveQuery subscriptions
- Add mcp.registry.changed event type to DaemonEventMap in daemon-hub.ts
- Add mcp.registry.* request/response types to packages/shared/src/api.ts
- Wire handlers into setupRPCHandlers via registerAppMcpHandlers(); add optional appMcpManager to RPCHandlerDependencies
- 22 unit tests covering all handlers, event emission, and validation errors

### DIFF
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -336,6 +336,16 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		inboxItemId: string;
 	};
 
+	// App-level MCP registry events
+	/**
+	 * Emitted by app-mcp-handlers on create/update/delete/setEnabled.
+	 * Used by RoomRuntimeService (Task 3.2) for hot-reload of MCP server lists.
+	 * This is separate from LiveQuery — both are emitted on every write.
+	 */
+	'mcp.registry.changed': {
+		sessionId: string;
+	};
+
 	// Goal events
 	'goal.created': {
 		sessionId: string;
--- a/packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts
@@ -1,0 +1,154 @@
+/**
+ * App MCP Registry RPC Handlers
+ *
+ * Exposes the application-level MCP server registry via RPC:
+ * - mcp.registry.list       — list all registry entries
+ * - mcp.registry.get        — get a single entry by id
+ * - mcp.registry.create     — add a new entry, emit mcp.registry.changed
+ * - mcp.registry.update     — update an entry, emit mcp.registry.changed
+ * - mcp.registry.delete     — remove an entry, emit mcp.registry.changed
+ * - mcp.registry.setEnabled — toggle enabled flag, emit mcp.registry.changed
+ *
+ * Note: mcp.registry.listErrors is registered in mcp-handlers.ts (requires
+ * the concrete AppMcpLifecycleManager).
+ */
+
+import type { MessageHub } from '@neokai/shared';
+import type {
+	AppMcpServer,
+	CreateAppMcpServerRequest,
+	UpdateAppMcpServerRequest,
+} from '@neokai/shared';
+import type { DaemonHub } from '../daemon-hub';
+import type { AppMcpServerRepository } from '../../storage/repositories/app-mcp-server-repository';
+import { Logger } from '../logger';
+
+const log = new Logger('app-mcp-handlers');
+
+// ---------------------------------------------------------------------------
+// Handler context
+// ---------------------------------------------------------------------------
+
+export interface AppMcpHandlerContext {
+	db: { appMcpServers: AppMcpServerRepository };
+	daemonHub: DaemonHub;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function emitChanged(daemonHub: DaemonHub): void {
+	daemonHub.emit('mcp.registry.changed', { sessionId: 'global' }).catch((err) => {
+		log.warn('Failed to emit mcp.registry.changed:', err);
+	});
+}
+
+// ---------------------------------------------------------------------------
+// Handler registration
+// ---------------------------------------------------------------------------
+
+export function registerAppMcpHandlers(messageHub: MessageHub, ctx: AppMcpHandlerContext): void {
+	const { db, daemonHub } = ctx;
+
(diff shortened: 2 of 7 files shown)


lsm (commented):

---

### Verdict: APPROVE

The implementation is well-structured, follows existing codebase patterns, and covers the required CRUD RPC endpoints correctly. All 22 unit tests pass, linting is clean, and typecheck succeeds.

---

### Findings

#### P2 (important suggestions)

**1. `mapMcpServerRow` passes through `null` for nullable TEXT columns (`description`, `url`, `command`)**

`packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts`, lines 234-244

The `...row` spread passes through SQLite `null` values for `description`, `url`, and `command` as `null`. The `AppMcpServer` type declares these as `description?: string` (optional, meaning `undefined`), not `string | null`. The repository's `rowToServer` function (in `app-mcp-server-repository.ts` line 53) handles this by conditionally omitting the key: `...(row.description !== null ? { description: row.description } : {})`. The LiveQuery path does not do this, creating a type inconsistency between the RPC handler path and the LiveQuery path. At runtime, a frontend consumer using LiveQuery would receive `null` for nullable fields while the RPC path returns `undefined`. This is not a crash-level issue but creates subtle frontend typing bugs (e.g., `if (server.description)` works with both, but `server.description ?? 'default'` would render the literal string `"null"" if stringified).

**2. Missing `mcp.registry.get` single-entry RPC endpoint**

`packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts`

The task says "CRUD" but there is no Read-single endpoint. The repository exposes `get(id)` and `getByName(name)`, but neither is wired as an RPC handler. While the LiveQuery `mcpServers.global` provides the full list reactively, a detail/edit view may need to fetch a single entry by ID without subscribing to the full list. Consider adding `mcp.registry.get` for completeness, or document the intentional omission.

#### P3 (nits)

**3. Response types in `api.ts` are defined but unused by the handlers**

`packages/shared/src/api.ts`, lines 634-680

The types `McpRegistryListResponse`, `McpRegistryCreateResponse`, `McpRegistryUpdateResponse`, `McpRegistryDeleteResponse`, `McpRegistrySetEnabledResponse`, and `McpRegistryListErrorsResponse` are defined but never imported or used by the handler code (which uses inline `satisfies` expressions). They serve as documentation only. This is fine for now, but consider either using them in the handlers (e.g., `return { servers } satisfies McpRegistryListResponse`) or adding a JSDoc comment noting they are for frontend consumption.

**4. `mcp.registry.update` emits `changed` event even when no fields changed**

`packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts`, line 101

If a caller sends `mcp.registry.update` with only an `id` and no actual fields to change, the handler still calls `emitChanged(daemonHub)`. The repository's update method (line 237 of the repo) short-circuits and returns without writing, so the event is a no-op. This is harmless but slightly wasteful. A minor optimization would be to check if the repo actually performed a write before emitting.

---

### What was done well

- **Consistent patterns**: The `emitChanged` fire-and-forget with `.catch()` matches existing handlers (e.g., `space-handlers.ts`). The `data as { id: string }` cast pattern is consistent.
- **Good validation**: Input validation for `name`, `sourceType`, `id`, and `enabled` at the handler boundary, with repository-level validation (duplicate names, valid sourceType values) as a second layer.
- **Event emission**: All four mutation handlers (create/update/delete/setEnabled) correctly emit `mcp.registry.changed` with `sessionId: 'global'` after successful writes, and do NOT emit on validation errors (correct ordering).
- **LiveQuery integration**: The `mcpServers.global` named query with `paramCount: 0` is correct for a parameterless query. The row mapper correctly handles JSON blob columns and boolean coercion.
- **Type-safe event type**: The `mcp.registry.changed` event is properly registered in `DaemonEventMap` in `daemon-hub.ts`.
- **Clean separation**: The `AppMcpManagerLike` interface keeps the handler decoupled from the not-yet-implemented lifecycle manager. The optional `appMcpManager` field gracefully degrades to empty errors array.
- **Test coverage**: 22 tests covering all 6 handlers, validation edge cases (blank name, missing fields, not-found entries), and event emission verification.

lsm (commented):

### Summary

All four items from the previous review have been addressed correctly. The fixes are clean, well-tested, and consistent with the existing codebase patterns.

### Fix Verification

**P2: mapMcpServerRow null handling** -- Fixed. The row mapper now uses conditional spreads (`...(row.description != null ? { description: row.description } : {})`) for all eight optional fields (description, command, url, args, env, headers, createdAt, updatedAt). This is structurally identical to the repository's `rowToServer` logic in `packages/daemon/src/storage/repositories/app-mcp-server-repository.ts` (lines 49-63), using `!= null` instead of `!== null` which is the safer choice for `Record<string, unknown>` rows. The field set is complete and matches.

**P2: Missing mcp.registry.get endpoint** -- Fixed. Added at `packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts:72-85`. Properly validates the `id` parameter, delegates to `db.appMcpServers.get(id)`, throws a clear error on not-found, and returns the server wrapped in `{ server }`. The `McpRegistryGetResponse` type was added to `packages/shared/src/api.ts:644-647`. Four tests cover registration, success, missing id, and not-found (lines 176-213 of the test file).

**P3: Unused response types in api.ts** -- Fixed. A JSDoc block at `packages/shared/src/api.ts:628-634` now clearly documents that the response types are for frontend consumption and that daemon handlers use inline `satisfies` expressions instead.

**P3: No-op update event emission** -- Fixed. At `packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts:120`, the handler checks `Object.keys(updates).length > 0` before calling `emitChanged()`. The test at line 309-314 verifies this: sending only an `id` with no update fields does not trigger the emit. The comment on lines 118-119 correctly explains that the repo itself short-circuits on empty updates, so the guard prevents a spurious event.

### Additional Observations

- All 27 tests pass (confirmed by running the suite).
- The test file header docstring could be updated to mention `mcp.registry.get` in the coverage list (line 6), but this is cosmetic.
- The no-op guard checks "did the caller send any fields" rather than "did the value actually change." This is the right granularity -- a value-equality check would require fetching the current row first, adding complexity for minimal benefit.

**Recommendation: APPROVE**

PR #826, opened 2026-03-24
task/allow changing worker and leader model in task exp

lsm:
- **feat(task): make rate/usage limits first-class task concerns with persistence and auto-retry**
- **fix(task): address review feedback for rate/usage limit first-class task concerns**
- **feat(task): allow changing worker and leader model in task expand view**

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -451,10 +451,6 @@ export class RoomRuntimeService {
 			(agentModels?.leader && agentModels.leader.trim() !== '' ? agentModels.leader : undefined) ??
 			(room.defaultModel && room.defaultModel.trim() !== '' ? room.defaultModel : undefined) ??
 			this.ctx.defaultModel;
-		const workerModel =
-			(agentModels?.worker && agentModels.worker.trim() !== '' ? agentModels.worker : undefined) ??
-			(room.defaultModel && room.defaultModel.trim() !== '' ? room.defaultModel : undefined) ??
-			this.ctx.defaultModel;
 
 		const runtime = new RoomRuntime({
 			room,
@@ -465,7 +461,6 @@ export class RoomRuntimeService {
 			sessionFactory,
 			workspacePath,
 			model: leaderModel,
-			workerModel,
 			defaultModel: this.ctx.defaultModel,
 			maxFeedbackIterations: maxReviewRounds,
 			maxConcurrentGroups,
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -122,8 +122,6 @@ export interface RoomRuntimeConfig {
 	workspacePath: string;
 	/** Leader model (agentModels.leader > room.defaultModel > global default) */
 	model?: string;
-	/** Worker model (agentModels.worker > room.defaultModel > global default) */
-	workerModel?: string;
 	/** Global default model for fallback when room doesn't specify one */
 	defaultModel?: string;
 	/** Max concurrent groups (default: 1 for MVP) */
@@ -311,10 +309,6 @@ export class RoomRuntime {
 			workspacePath: config.workspacePath,
 			model: config.model,
 			provider: config.model ? this.resolveProviderForModel(config.model) : undefined,
-			workerModel: config.workerModel,
-			workerProvider: config.workerModel
-				? this.resolveProviderForModel(config.workerModel)
-				: undefined,
 			getRoom: config.getRoom,
 			getTask: config.getTask,
 			getGoal: config.getGoal,
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2260,16 +2260,12 @@ export class RoomRuntime {
 								resetsAt: rateLimitBackoff.resetsAt,
 								sessionRole,
 							});
-							this.persistTaskRestriction(
+							void this.persistTaskRestriction(
 								group.taskId,
 								rateLimitBackoff,
 								'rate_limit',
 								`API rate limit (HTTP 429) in ${role}`
-							).catch((err: unknown) => {
-								log.error(
-									`Failed to persist rate limit restriction for task ${group.taskId}: ${String(err)}`
-								);
-							});
+							);
 						}
 					}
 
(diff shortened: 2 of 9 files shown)


lsm (commented):

---

### P0-1: `updateWorkerModel()` is dead code — the model change does not propagate

`packages/daemon/src/lib/room/runtime/task-group-manager.ts`, line 233

The new `updateWorkerModel(model, provider)` method on `TaskGroupManager` has **zero callers** anywhere in the codebase. When a user changes the worker model via the UI `TaskViewModelSelector` component, the following chain executes:

1. `TaskViewModelSelector` fires `session.model.switch` RPC (line 105 of `TaskViewModelSelector.tsx`)
2. The RPC handler calls `agentSession.handleModelSwitch(model, provider)` (line 451 of `session-handlers.ts`)
3. This changes the model for the **existing** SDK session only

Nothing in this chain calls `TaskGroupManager.updateWorkerModel()`. Furthermore, even if it did, the TGM's `_workerModel` field is **not consumed** when spawning new tasks. In `room-runtime.ts` line 3583, the worker model for task spawn is resolved via:

```ts
const workerModel = this.resolveAgentModel(currentRoom, workerRole);
```

This reads from `room.config.agentModels[workerRole]` — the room config — not from `this.taskGroupManager.getWorkerModel()`. So `updateWorkerModel` currently has no effect on any existing or future session.

**Impact:** The UI will show separate selectors for worker and leader, and clicking them will successfully switch the model on the *currently running* session. But the TGM's internal `_workerModel` tracking is disconnected from reality, and if the TGM were ever used to create new sessions (e.g., restart recovery), it would use the stale original model.

**Fix needed:** Either (a) wire `session.model.switch` to also call `updateWorkerModel` when the target session belongs to a task group's worker, and have task spawn read from `taskGroupManager.getWorkerModel()`, or (b) document that the TGM field is intentionally unused and remove it to avoid confusion.

---

### P0-2: `app_mcp_servers` table creation removed — breaks fresh installs

`packages/daemon/src/storage/schema/index.ts`, lines 331-348 (removed)

This PR removes both the `CREATE TABLE IF NOT EXISTS app_mcp_servers` from `createTables()` and the migration that creates it (`runMigration49`). However, the `AppMcpLifecycleManager` (introduced in PR #820, already merged to dev) actively queries this table at startup:

```ts
// packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts, line 54
const entries = this.db.appMcpServers.listEnabled();
```

On a fresh database install, this table won't exist, and every query will throw a `SQLITE_ERROR: no such table: app_mcp_servers`. This is a regression from dev.

**Fix needed:** Restore the `app_mcp_servers` table creation in `createTables()` and keep it as a separate migration (can be renumbered to 50 if needed).

---

### P1: `updateRoom()` does not sync worker model to TGM

`packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 522

The `updateRoom()` method updates the leader model on the TGM:

```ts
this.taskGroupManager.updateModel(updatedLeaderModel, this.resolveProviderForModel(updatedLeaderModel));
```

But it does **not** call `this.taskGroupManager.updateWorkerModel(...)`. If P0-1 is fixed (so that the TGM's workerModel is actually consumed), this path would also need to sync the worker model to keep the TGM aligned with room config changes.

---

### P2: No test for `updateWorkerModel()`

`packages/daemon/src/lib/room/runtime/task-group-manager.ts`

The new method has no unit test coverage. If the intent is to keep it (per P0-1 fix), add a test in the task-group-manager test file verifying that:
- Calling `updateWorkerModel` changes `getWorkerModel()` return value
- Calling `updateWorkerModel(undefined, undefined)` falls back to the leader model
- The getter returns the updated value

---

### P2: Missing edge-case test for worker-only model selector

`packages/web/src/components/room/__tests__/TaskInfoPanel.test.tsx`

The new test (line 476) covers "both worker and leader have model config", but there's no test for the scenario where only the worker has a model config and the leader has no model (or no leader session). The old test at line 434 ("should render model selector when session has model config") covers worker-only but does not verify the "Worker:" label prefix that was added. Consider updating the existing test to also assert the label.

---

### Note (unrelated to model switching): schema cleanup

The removal of `restrictions TEXT` from the `goals` table in `createTables()` (where it was accidentally placed) is correct and good. The migration renumbering is also safe since all migrations are idempotent.

lsm (commented):

**Recommendation: APPROVE**

---

All issues from the previous review have been addressed correctly.

### P0-1 (Fixed): Dead `updateWorkerModel()` code removed
The `updateWorkerModel()` method, `getWorkerModel()`, and all `workerModel`/`workerProvider` fields have been cleanly removed from:
- `packages/daemon/src/lib/room/runtime/task-group-manager.ts` -- interface, constructor, and methods gone
- `packages/daemon/src/lib/room/runtime/room-runtime.ts` -- `RoomRuntimeConfig` no longer has `workerModel`/ `workerProvider`; TGM constructor call no longer passes them
- `packages/daemon/src/lib/room/runtime/room-runtime-service.ts` -- `workerModel` resolution and propagation removed

No stale references remain (verified via grep). The remaining `workerModel` usages in `room-runtime.ts` are local variables inside `resolveAgentModel()` calls at spawn time -- pre-existing code that reads from room config, completely unrelated.

### P0-2 (Fixed): `app_mcp_servers` table restored
- `CREATE TABLE IF NOT EXISTS app_mcp_servers` is restored in `createTables()` (packages/daemon/src/storage/schema/index.ts, line 337)
- New `runMigration50()` function added to migrations.ts (line 3153) with idempotent `CREATE TABLE IF NOT EXISTS`
- Export added in `schema/index.ts` (line 25)

### P1 (N/A): Dead code removal eliminates the sync issue

### P2-1 (N/A): Method removed

### P2-2 (Fixed): Worker-only test updated
The worker-only model selector test now asserts `toMatch(/Worker:.*Model:/)` to verify the new label prefix. A new test for both-sessions scenario verifies both `Worker:.*Model:` and `Leader:.*Model:` labels, and checks that two model selectors render with correct model names.

### Bonus fix noted
The `restrictions TEXT` column that was accidentally placed on the `goals` table in `createTables()` has been correctly removed. It belongs only on the `tasks` table (line 174), which already has it.

### Migration renumbering is safe
The swap (49 = restrictions, 50 = app_mcp_servers) is correct. Both migrations are idempotent and touch different tables, so the order is irrelevant for correctness. Databases that ran the old numbering are unaffected.

### P3 nit (not blocking)
The leader-only model selector test (line 456-475 in `TaskInfoPanel.test.tsx`) checks `toContain('Model:')` but does not verify the `Leader:` label prefix. The new both-sessions test covers this, so it is purely cosmetic.

All tests pass (19 task-restrictions tests, 47 TaskInfoPanel tests).

PR #827, opened 2026-03-25
fix: model switch not taking effect on coder session

lsm:
## Summary
- Model switching on task sessions in non-active states (review, needs_attention) now properly restarts the query
- Previously, only the session config was updated without calling `restart()`, leaving the coder session using the old model
- The fix ensures `restart()` is called regardless of `queryObject` state, validating the session file and emitting a fresh `system:init` with the correct model

## Root Cause
When a task is in a non-active state, `queryObject` is null. The original code only updated the in-memory and DB config but skipped the restart call, so the SDK subprocess kept running with the old model.

## Test plan
- [x] Unit tests updated in `model-switch-handler.test.ts` (27 tests passing)
- [x] All daemon unit tests pass (7574 total)
- [x] Lint passes (0 warnings, 0 errors)
- [x] Typecheck passes

### DIFF
--- a/packages/daemon/src/lib/agent/model-switch-handler.ts
+++ b/packages/daemon/src/lib/agent/model-switch-handler.ts
@@ -175,7 +175,11 @@ export class ModelSwitchHandler {
 			}
 
 			if (!queryObject) {
-				// Query hasn't been created yet - just update config, it will be used when query starts
+				// Query hasn't been created yet OR query was already completed/interrupted.
+				// In both cases, we must call restart() to ensure the new model takes effect:
+				// - restart() validates the SDK session file before starting a new query
+				// - Without this, a stale session file could cause the new query to use old state
+				// - Even if no query is running, restart() ensures clean state for the next query
 				session.config.model = resolvedModel;
 				// newProviderInstance is guaranteed non-null here (we returned early above).
 				session.config.provider = newProviderInstance.id as Provider;
@@ -198,6 +202,10 @@ export class ModelSwitchHandler {
 					source: 'model-switch',
 					session: { config: session.config },
 				});
+
+				// Always restart to ensure new model takes effect, even when queryObject is null.
+				// This validates the session file and starts a fresh query with the new model.
+				await lifecycleManager.restart();
 			} else {
 				// Query exists - always restart to apply the new model/provider.
 				// We must restart even if firstMessageReceived is false because the SDK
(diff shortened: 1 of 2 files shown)


lsm (commented):

This is a correct, focused fix that addresses the root cause. The change is minimal (one line of functional code added) and well-justified.

### Analysis

**Root cause is correctly identified and fixed.** When a task session is in a non-active state (review, needs_attention), the SDK query has already completed or been interrupted, making `queryObject` null. The original code only updated the in-memory config and DB, but never called `restart()`. Since `restart()` calls `startStreamingQuery()` which spawns a new SDK subprocess with the updated `session.config.model`, this was the right call to add.

**Side effect analysis -- safe.** I traced the full call chain for the `!queryObject` case:

1. `lifecycleManager.restart()` is called (line 205)
2. `restart()` calls `stop()`, which is a no-op when `queryObject` and `queryPromise` are both null (messageQueue.stop() just sets `running = false`, interrupt/close are skipped)
3. `restart()` then calls `startStreamingQuery()` which calls `queryRunner.start()`
4. `start()` starts the messageQueue and creates a fresh query with the new model from `session.config`

This is safe. There is no guard in `restart()` against being called with no active query -- the method handles it gracefully by design.

**Potential concern about restarting intentionally-stopped sessions.** One could argue that a task in `review` or `needs_attention` state was intentionally paused and restarting it via model switch might be surprising. However, this concern is mitigated by two factors: (a) the model switch is an explicit user action, so the user expects something to happen, and (b) the task was going to resume eventually anyway, so starting a fresh query with the correct model is the right behavior. The `restart()` method does not bypass any intentional stop guards -- it is designed for exactly this use case (model switching).

**Error handling is correct.** The `await lifecycleManager.restart()` is inside the existing try/catch block (lines 106-279). If restart fails, the error propagates to the catch block, which calls `errorManager.handleError()` and returns a failure result. This is the same error path used by the `else` branch.

**Pre-existing note (not a regression):** `restart()` in `QueryLifecycleManager` does not check `isCleaningUp()`. If a model switch and session cleanup race, the restart could conflict. But this is pre-existing behavior -- the `else` branch has always called `restart()` without this check -- so this is not a regression from this PR.

### P3: Minor observations

1. **Code duplication:** The `!queryObject` and `else` branches are now nearly identical (same config update, DB persist, context tracker update, event emission, and restart). A future cleanup could merge them into a single path. Not a blocker, just a maintenance note.

2. **Test gap:** The existing error handling test (`should handle errors and call error manager`, line 458) only covers restart failure when `queryObject` is non-null (it uses `createHandler()` which defaults to a non-null queryObject). There is no test verifying that restart failure is handled correctly in the `!queryObject` branch. Given that both branches share the same try/catch, this is low risk, but a dedicated test would be more thorough.

3. **Test describe block name:** The describe block at line 302 titled `when query not started` could benefit from a rename to `when queryObject is null` to better reflect the new semantics (covers both pre-start and post-completion scenarios).

PR #831, opened 2026-03-25
fix(goal-handlers): resolve short task IDs to full UUIDs in task.approve

lsm:
The task.approve and goal.linkTask RPC handlers were receiving short IDs (e.g., 't-42') from the UI but not resolving them to full UUIDs before passing to the task manager and runtime. This caused "Task not found" errors when approving tasks or linking tasks to goals.

Changes:
- Import resolveTaskId from id-resolution and TaskRepository in goal-handlers
- Update TaskManagerFactory type to return both taskManager and taskRepo
- Call resolveTaskId before using taskId in task.approve handler
- Call resolveTaskId before using taskId in goal.linkTask handler
- Error and log messages now use the resolved full UUID
- Update goalTaskManagerFactory in index.ts to return both taskManager and taskRepo
- Update existing tests to use valid UUIDs and add tests for short ID resolution

Test plan:
- [x] Unit tests pass (103 tests in goal-handlers.test.ts)
- [x] Typecheck passes
- [x] Lint passes

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/goal-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/goal-handlers.ts
@@ -287,9 +287,14 @@ export function setupGoalHandlers(
 		if (!params.taskId) {
 			throw new Error('Task ID is required');
 		}
+		if (!taskManagerFactory) {
+			throw new Error('Task manager factory is required for goal.linkTask');
+		}
 
 		const goalManager = goalManagerFactory(params.roomId);
 		const goalId = resolveGoalId(params.goalId, params.roomId, goalManager);
+		const { taskRepo } = taskManagerFactory(params.roomId);
+		const taskId = resolveTaskId(params.taskId, params.roomId, taskRepo);
 
 		// For recurring missions with an active execution, use the atomic dual-write path.
 		let goal;
@@ -297,12 +302,12 @@ export function setupGoalHandlers(
 		if (linkedGoal?.missionType === 'recurring') {
 			const activeExecution = goalManager.getActiveExecution(goalId);
 			if (activeExecution) {
-				goal = await goalManager.linkTaskToExecution(goalId, activeExecution.id, params.taskId);
+				goal = await goalManager.linkTaskToExecution(goalId, activeExecution.id, taskId);
 			} else {
-				goal = await goalManager.linkTaskToGoal(goalId, params.taskId);
+				goal = await goalManager.linkTaskToGoal(goalId, taskId);
 			}
 		} else {
-			goal = await goalManager.linkTaskToGoal(goalId, params.taskId);
+			goal = await goalManager.linkTaskToGoal(goalId, taskId);
 		}
 
 		return { goal };
--- a/packages/daemon/src/lib/rpc-handlers/goal-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/goal-handlers.ts
@@ -32,7 +32,8 @@ import type { TaskManager } from '../room/managers/task-manager';
 import type { RoomRuntimeService } from '../room/runtime/room-runtime-service';
 import { Logger } from '../logger';
 import { isValidCronExpression, getNextRunAt, getSystemTimezone } from '../room/runtime/cron-utils';
-import { resolveGoalId } from '../id-resolution';
+import { resolveGoalId, resolveTaskId } from '../id-resolution';
+import type { TaskRepository } from '../../storage/repositories/task-repository';
 
 const log = new Logger('goal-handlers');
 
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Summary

The fix correctly resolves short task IDs to full UUIDs in the task.approve handler by calling resolveTaskId() before using the ID. The implementation follows the established pattern already used in task-handlers.ts (for task.reject and other handlers). Tests are well-structured and cover both the short ID resolution path and the passthrough UUID path.

All 103 tests pass.

### Issues Found

#### P2: goal.linkTask handler does not resolve short task IDs (same file, same bug class)

packages/daemon/src/lib/rpc-handlers/goal-handlers.ts, lines 300-305

The goal.linkTask handler (line 278) accepts a taskId parameter from the UI but passes it directly to goalManager.linkTaskToExecution() and goalManager.linkTaskToGoal() without resolving short IDs. This is the exact same bug class as the one fixed for task.approve:

goal = await goalManager.linkTaskToExecution(goalId, activeExecution.id, params.taskId);
goal = await goalManager.linkTaskToGoal(goalId, params.taskId);

If the UI sends a short ID (e.g., t-42) for goal.linkTask, it will fail or produce incorrect behavior. Since this is in the same file being modified, it should be addressed in this PR.

#### P3: Concrete TaskRepository type import in handler module

packages/daemon/src/lib/rpc-handlers/goal-handlers.ts, line 36

The type import uses the concrete TaskRepository class. However, the resolveTaskId function only needs the minimal TaskRepoForResolve interface (from id-resolution.ts). The TaskManagerFactory type at line 63-66 also uses the concrete TaskRepository type. Using the concrete type here is fine for now since the factory in index.ts does create a real TaskRepository, but it creates a tighter coupling between the handler module and the repository implementation. Consider using the TaskRepoForResolve interface for the type annotation instead.

### What Looks Good

- Correct resolution pattern: resolveTaskId() is called before any other use of taskId, and all downstream references (error messages, log messages, runtime.resumeWorkerFromHuman) use the resolved UUID.
- Consistent with existing code: The approach matches how task.reject and other handlers in task-handlers.ts already handle this.
- Factory type change is clean: Changing TaskManagerFactory to return an object with both taskManager and taskRepo is a reasonable and minimal API change.
- Test coverage: The two new tests properly verify (1) short ID is resolved via getTaskByShortId and the resolved ID is forwarded, and (2) UUID passthrough skips the lookup.
- Existing test fix: Updating test IDs from fake strings (task-123) to valid UUIDs makes the tests more realistic and consistent with production data.

### Recommendation

APPROVE -- The core fix is correct and well-tested. The P2 (goal.linkTask) is a related issue in the same file that ideally should be addressed here for consistency, but it can be a follow-up.

lsm (commented):

### Summary

The PR correctly fixes short task ID resolution in both `task.approve` and `goal.linkTask` RPC handlers. The implementation is consistent between the two handlers, the factory wiring in `index.ts` is correct, and all tests pass. The changes are focused and well-scoped.

### What looks good

1. **Consistent fix pattern**: Both handlers now call `resolveTaskId(params.taskId, params.roomId, taskRepo)` before using the task ID, following the same pattern already established by `resolveGoalId` for goal IDs. The resolution logic in `packages/daemon/src/lib/id-resolution.ts` is clean and well-documented.

2. **Factory type change is correct**: The `TaskManagerFactory` type was expanded from returning a bare `Pick<TaskManager, ...>` to returning `{ taskManager, taskRepo }`. The implementation in `index.ts` (line 146-155) properly creates both instances, and the `TaskRepository(db, reactiveDb)` constructor call is valid (the third `shortIdAllocator` param is optional).

3. **Defensive guard in goal.linkTask**: The new check `if (!taskManagerFactory) throw ...` at line 290-292 is good. Previously, if `taskManagerFactory` were undefined, calling it would throw an opaque error. Now the failure mode is explicit.

4. **Error/log messages use resolved UUID**: Both handlers correctly use the resolved `taskId` (not the raw input) in error messages and log output, making debugging easier.

5. **Test updates are solid**: Existing tests were updated to use valid UUIDs, and two new tests for `task.approve` verify short ID resolution and UUID passthrough. The `goal.linkTask` test now properly asserts the resolved UUID is passed to `linkTaskToGoal`.

6. **Typecheck and lint pass**: Both `bun run typecheck` and `bun run lint` report zero issues.

### Suggestions (P3 - minor)

1. **Missing test: task short ID not found in goal.linkTask** (`packages/daemon/tests/unit/rpc-handlers/goal-handlers.test.ts`, around line 1769): There is a test for when the *goal* short ID is not found, but no test for when the *task* short ID cannot be resolved. Adding a test like:
   ```ts
   it('throws when task short ID is not found', async () => {
     resolveShortId();
     mockGoalManager.getGoal.mockResolvedValueOnce(...);
     mockTaskRepo.getTaskByShortId.mockReturnValueOnce(null);
     const handler = messageHubData.handlers.get('goal.linkTask')!;
     await expect(
       handler({ roomId: ROOM_UUID, goalId: 'g-1', taskId: 't-9999' }, {})
     ).rejects.toThrow('Task not found: t-9999');
   });
   ```
   would close the coverage gap. The same applies to `task.approve` -- there is no test for `resolveTaskId` throwing when the short ID is unresolvable. These are low priority since the resolution logic itself is covered by `id-resolution.ts`, but handler-level coverage ensures the integration is solid.

### Verdict

**APPROVE** -- The fix is correct, consistent, and well-tested. The missing negative-path tests are minor and do not block merge.

PR #834, opened 2026-03-25
fix(agent): clear models cache on restart/reset to fix mid-turn model switch

lsm:
When switching models mid-conversation or clicking Reset Agent, the agent
session must pick up the new model from the database, not use cached
model info. This change adds clearModelsCache() calls to both restart()
and reset() methods in QueryLifecycleManager to ensure fresh model
info is fetched.

Also adds Reset Agent buttons to TaskInfoPanel for both worker and
leader agents, allowing manual reset when agents get stuck or have
wrong state. The buttons call session.resetQuery RPC which invokes
lifecycleManager.reset() with full cleanup.

Fixes:
- Mid-turn model switch not picking up new model
- Models cache potentially returning stale model info

### DIFF
--- a/packages/daemon/src/lib/agent/query-lifecycle-manager.ts
+++ b/packages/daemon/src/lib/agent/query-lifecycle-manager.ts
@@ -194,6 +194,10 @@ export class QueryLifecycleManager {
 				}
 			}
 
+			// Clear models cache to ensure the new model is fetched fresh from DB
+			// This is critical for model switch to pick up the correct model
+			await this.ctx.clearModelsCache();
+
 			await this.ctx.startStreamingQuery();
 		} catch (error) {
 			const errorMessage = error instanceof Error ? error.message : 'Unknown error';
@@ -223,6 +227,8 @@ export class QueryLifecycleManager {
 			this.ctx.pendingRestartReason = null;
 			messageHandler.resetCircuitBreaker();
 			await stateManager.setIdle();
+			// Clear models cache to ensure fresh model info is fetched from DB
+			await this.ctx.clearModelsCache();
 			return { success: true };
 		}
 
@@ -254,6 +260,10 @@ export class QueryLifecycleManager {
 			// Post-stop: Reset state
 			this.ctx.firstMessageReceived = false;
 			await stateManager.setIdle();
+
+			// Clear models cache to ensure fresh model info is fetched from DB
+			// This is critical for model switch to pick up the new model
+			await this.ctx.clearModelsCache();
 
 			// Optionally restart
 			if (restartAfter) {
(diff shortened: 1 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR is clean and well-scoped. The backend fix is correct, the tests verify the key behavior, and the frontend buttons wire up properly. I have only minor observations.

---

### Backend fix (clearModelsCache on restart/reset)

The root cause analysis is correct. When switchModel() calls lifecycleManager.restart(), the new SDK subprocess fetches models and populates the cache via onModelsFetched(). Without clearing the cache first, the stale model list from the previous subprocess persists and can cause the UI to show wrong model info.

The three insertion points are well-chosen:

1. **restart() (line 199)** -- cleared after session file validation, before startStreamingQuery(). This ensures the model switch path always gets fresh models. Inside the existing try-catch, so cache-clear failures are handled gracefully.

2. **reset() early return (line 231)** -- handles the case where no query is running. Correct but note this is outside a try-catch: if the dynamic import('../model-service') fails (extremely unlikely but not impossible), the error would propagate. In cleanup(), the same call is wrapped in try-catch. Consider adding a try-catch here for consistency, though the risk is very low since the underlying clearModelsCache() just deletes from Maps.

3. **reset() main path (line 266)** -- after stop() and state reset, before optional restart. Inside the existing try-catch. Correct placement.

### Frontend: Reset Worker/Leader buttons

The buttons are correctly wired:

- **TaskInfoPanel.tsx**: Props interface extended cleanly with onResetWorkerAgent/onResetLeaderAgent handlers, visibility flags, and disabled state. The hasVisibleActions computation correctly includes the new flags. Button styling follows the existing pattern (border with hover color change). Data-testid attributes are present for E2E testing.

- **TaskView.tsx**: Handlers follow the established pattern (connectionManager.getHubIfConnected() + hub.request('session.resetQuery', ...) + toast feedback), consistent with 28 other call sites in the codebase. Buttons are disabled during interrupts, which is correct. The workerSession/leaderSession ternary guards ensure handlers are only set when sessions exist.

### Tests

Three new test cases verify the key behavior:

1. **clears models cache before starting new query** (restart)
2. **clears models cache on early return when no query is running** (reset early return)
3. **clears models cache** assertion added to existing full-reset test

All 65 tests pass. The test for the reset() early return path is particularly important since it exercises the code path that was most likely to be missed.

### Minor observations (P3)

1. **Duplicated reset-query pattern**: The inline handlers in TaskView.tsx (lines 328-390) duplicate the logic in useSessionActions.handleResetAgent(). A shared utility (e.g., resetAgentSession(sessionId: string): Promise<result>) would reduce duplication, but this is not a bug and the current approach is functional.

2. **ensureQueryStarted() does not clear models cache**: If a model is changed in the DB outside of the normal handleModelSwitch() flow and then a message is sent via startQueryAndEnqueue() -> ensureQueryStarted(), the cache would not be cleared. This is an edge case and not relevant to this PR's scope, but worth keeping in mind for future hardening.

PR #835, opened 2026-03-25
feat(shared): define reference types for @ referencing system

lsm:
Add core type definitions and regex pattern for the @ reference system
in packages/shared/src/types/reference.ts, exported via mod.ts.

### DIFF
--- a/packages/shared/src/mod.ts
+++ b/packages/shared/src/mod.ts
@@ -20,6 +20,7 @@ export * from './types/space.ts';
 export * from './types/space-utils.ts';
 export * from './types/tools.ts';
 export * from './types/app-mcp-server.ts';
+export * from './types/reference.ts';
 export * from './live-query-types.ts';
 export * from './prompts/index.ts';
 
--- a/packages/shared/src/types/reference.ts
+++ b/packages/shared/src/types/reference.ts
@@ -1,0 +1,103 @@
+/**
+ * Types for the @ reference system.
+ *
+ * Users can type @ in chat inputs to reference tasks, goals, files, and folders.
+ * References are serialized as `@ref{type:id}` in message content.
+ */
+
+// ============================================================================
+// Core Types
+// ============================================================================
+
+/** The type of entity being referenced */
+export type ReferenceType = 'task' | 'goal' | 'file' | 'folder';
+
+/**
+ * A parsed @ mention from user input.
+ * Represents a reference that has been inserted into the input field.
+ */
+export interface ReferenceMention {
+	type: ReferenceType;
+	id: string;
+	displayText: string;
+}
+
+/**
+ * A search result returned by the reference search RPC.
+ */
+export interface ReferenceSearchResult {
+	type: ReferenceType;
+	id: string;
+	/** Short human-readable identifier (e.g. "t-42" for tasks) */
+	shortId?: string;
+	displayText: string;
+	/** Secondary line shown in the autocomplete menu */
+	subtitle?: string;
+}
+
+/**
+ * A resolved reference — the raw entity data retrieved for a given id.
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE** (with suggestions for follow-up milestones)

---

This is a clean, well-structured foundation for the @ referencing system. The types are well-documented, the barrel export is in the right place, all checks pass (typecheck, lint, knip, tests), and the code follows the existing patterns in `packages/shared/src/types/`. I have a few suggestions that are not blocking but would be worth addressing in M2/M3 when these types get consumed.

### P2 -- Consider adding a `parseReference` helper (packages/shared/src/types/reference.ts:103)

The raw regex is fine for M1, but every consumer of `REFERENCE_PATTERN` will need to re-implement the same `matchAll` loop plus validation that the captured `type` is a valid `ReferenceType`. A small utility function would centralize this and prevent drift:

```ts
export function parseReferences(text: string): Array<{ type: ReferenceType; id: string; index: number; fullMatch: string }> {
  // reset lastIndex since REFERENCE_PATTERN has the g flag
  REFERENCE_PATTERN.lastIndex = 0;
  const results: Array<...> = [];
  for (const match of text.matchAll(REFERENCE_PATTERN)) {
    const type = match[1] as ReferenceType;
    const id = match[2];
    // validate type is a known ReferenceType
    if (!VALID_REFERENCE_TYPES.includes(type)) continue;
    results.push({ type, id, index: match.index!, fullMatch: match[0] });
  }
  return results;
}
```

This is appropriate for M2 (frontend) or M3 (backend) when the first real consumer appears.

### P2 -- Consider adding type guard functions for ResolvedReference variants (packages/shared/src/types/reference.ts:52-66)

The variant interfaces (`ResolvedTaskReference`, etc.) are defined but there are no type guards to narrow `ResolvedReference` at runtime. This is a standard pattern for discriminated unions and will be needed as soon as consumers need to handle the polymorphic `data` field:

```ts
export function isTaskReference(ref: ResolvedReference): ref is ResolvedTaskReference {
  return ref.type === 'task';
}
```

Again, not blocking for M1 -- this becomes relevant when the backend resolver (M3) returns typed data.

### P3 -- Regex captures whitespace in type/id (packages/shared/src/types/reference.ts:103)

The regex `[^}:]+` allows spaces inside the braces. Input like `@ref{ task : t-42 }` would match and capture ` task ` / ` t-42 ` with leading/trailing whitespace. This is not dangerous (the type validation in a `parseReference` helper would reject it), but it's worth noting so the regex isn't used directly for parsing without trimming. The existing test for "missing colon separator" at line 165 is good, but a test case for whitespace-padded input would make this behavior explicit.

### P3 -- Module-level `g`-flagged regex (packages/shared/src/types/reference.ts:103)

The comment at line 100 correctly warns about `lastIndex` stale state. This is fine for now since the codebase will likely use `matchAll` consistently. If any future consumer uses `.test()` or `.exec()`, the shared instance will leak state between calls. An alternative pattern is to export both a non-flagged base regex and a helper function that internally applies the `g` flag. Low priority since the comment exists.

### Testing observations

- 19 tests all pass, covering type shapes and basic regex positive/negative cases.
- Missing edge case tests: whitespace-padded input (`@ref{ task : t-42 }`), empty type/id (`@ref{:t-42}`, `@ref{task:}` -- both correctly rejected), absolute file paths (`@ref{file:/abs/path}`), relative paths with `..`, and nested `@` in paths (`@ref{file:node_modules/@types/node}` -- works correctly).
- The `ResolvedReference` variant tests at lines 60-87 verify discriminated narrowing but only assert on `type` -- a future test that verifies `data` shape narrowing would add value once the `data` field is typed more specifically.

PR #836, opened 2026-03-25
feat(web): add useReferenceAutocomplete hook for @ reference system

lsm:
- Define core reference types in packages/shared: ReferenceType, ReferenceMention,
  ReferenceSearchResult, ResolvedReference, ReferenceMetadata, REFERENCE_PATTERN
- Export reference types from @neokai/shared barrel
- Implement useReferenceAutocomplete hook with:
  - @ detection anywhere in text (not just at start)
  - Debounced reference.search RPC calls (300ms)
  - Stale response cancellation via version counter
  - Keyboard navigation (ArrowUp/Down, Enter, Escape)
  - Multiple @ support (only last active @ triggers autocomplete)
  - exportActiveAtQuery utility for cursor-aware @ detection
- Export hook and types from packages/web hooks index
- Unit tests: 14 shared type tests, 38 hook tests covering detection,
  debouncing, results, keyboard navigation, multiple @ support, and
  stale response cancellation

### DIFF
--- a/packages/shared/src/types/reference.ts
+++ b/packages/shared/src/types/reference.ts
@@ -1,20 +1,20 @@
 /**
- * Types for the @ reference system.
+ * Reference Types
  *
- * Users can type @ in chat inputs to reference tasks, goals, files, and folders.
- * References are serialized as `@ref{type:id}` in message content.
+ * Core type definitions for the @ reference system.
+ * References allow users to mention tasks, goals, files, and folders
+ * in chat inputs using @ref{type:id} syntax.
  */
 
-// ============================================================================
-// Core Types
-// ============================================================================
-
-/** The type of entity being referenced */
+/**
+ * The type of entity being referenced.
+ * Union type (not enum) following existing codebase patterns like `SessionType`.
+ */
 export type ReferenceType = 'task' | 'goal' | 'file' | 'folder';
 
 /**
- * A parsed @ mention from user input.
- * Represents a reference that has been inserted into the input field.
+ * A mention of a reference in text, as inserted by the autocomplete.
+ * Stored in message content as @ref{type:id}.
  */
 export interface ReferenceMention {
 	type: ReferenceType;
@@ -23,31 +23,28 @@ export interface ReferenceMention {
 }
 
 /**
- * A search result returned by the reference search RPC.
+ * A single result returned from the reference.search RPC.
  */
 export interface ReferenceSearchResult {
 	type: ReferenceType;
 	id: string;
-	/** Short human-readable identifier (e.g. "t-42" for tasks) */
+	/** Short human-readable ID (e.g. "t-42", "g-7") for tasks and goals */
 	shortId?: string;
+	/** Primary display text shown in the autocomplete menu */
 	displayText: string;
-	/** Secondary line shown in the autocomplete menu */
+	/** Secondary text shown below displayText (e.g. task status, file path) */
 	subtitle?: string;
 }
 
 /**
- * A resolved reference — the raw entity data retrieved for a given id.
- * The shape of `data` is polymorphic based on `type`.
+ * A resolved reference with full entity data.
  */
 export interface ResolvedReference {
 	type: ReferenceType;
 	id: string;
+	/** Polymorphic — cast based on `type` */
 	data: unknown;
 }
-
-// ============================================================================
-// Resolved Reference Variants
-// ============================================================================
 
 export interface ResolvedTaskReference extends ResolvedReference {
 	type: 'task';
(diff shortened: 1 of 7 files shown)


lsm (commented):

### Verdict: REQUEST_CHANGES

Overall this is a well-structured implementation that cleanly follows the existing `useCommandAutocomplete` pattern. All tests pass (14 shared type tests + 38 hook tests), lint is clean, typecheck is clean, and knip reports no dead exports. The stale-response cancellation via version counter is a good approach. That said, I found a few issues worth addressing.

---

### P1 — Docstring/implementation mismatch in `extractActiveAtQuery`

**File:** `packages/web/src/hooks/useReferenceAutocomplete.ts`, line 49

The JSDoc states:
> The @ itself must not be followed immediately by another @ (malformed)

But the implementation does NOT check for this. Typing `@@` returns `"@"` (it finds the first `@` at word-start position, and `afterAt = "@"` contains no whitespace). The docstring should either be corrected to match the actual behavior, or the implementation should reject `@@`.

Since `@@` is a plausible user input (e.g., GitHub-style notifications), the current behavior of returning `"@"` as the query seems reasonable. The fix here is to **remove the misleading docstring line** rather than adding the check.

---

### P2 — Barrel export test not updated

**File:** `packages/web/src/hooks/__tests__/index.test.ts`

The `useReferenceAutocomplete` and `extractActiveAtQuery` exports were added to `packages/web/src/hooks/index.ts` but the barrel export test was not updated to verify them. While the existing barrel test is already incomplete (missing `useAutoScroll`, `useTurnBlocks`, etc.), adding the new exports here would at least prevent regressions on the newly-added code.

---

### P2 — `close()` does not reset `results` or `selectedIndex`

**File:** `packages/web/src/hooks/useReferenceAutocomplete.ts`, lines 167-169

When `close()` is called, only `showAutocomplete` is set to `false` while `results` and `selectedIndex` retain their previous values. If the user closes autocomplete (Escape) and then types a new `@ref{...}`, the stale results from the previous search briefly remain in state until the 300ms debounce fires and a new RPC response arrives.

This is consistent with how `useCommandAutocomplete.close()` works (same pattern), so it is a pre-existing design choice rather than a regression. However, since `useReferenceAutocomplete` fetches results asynchronously (unlike the synchronous command filtering), the stale-data flash is more noticeable here. Consider resetting `results` to `[]` and `selectedIndex` to `0` in `close()`, or at minimum resetting them in the `content` effect when a new `@` query is detected (before the debounce fires).

---

### P3 — `ResolvedTaskReference` and `ResolvedGoalReference` don't narrow `data`

**File:** `packages/shared/src/types/reference.ts`, lines 49-55

These interfaces narrow `type` to a literal but inherit `data: unknown` from the base, unlike `ResolvedFileReference` and `ResolvedFolderReference` which narrow `data` to a specific shape. This makes them pure type-discriminator aliases with no added value. Consider either adding a concrete `data` shape for task/goal resolved references, or removing these types until the backend handler exists and the shapes are known.

---

### P3 — `@ts-nocheck` on test file

Consistent with all other hook test files in the same directory. No action needed, noting for completeness.

---

### Minor observation (not scored)

After a reference is selected and `onSelect` replaces the `@query` with `@ref{type:id}`, the `extractActiveAtQuery` function would re-trigger autocomplete for the inserted token (since `afterAt = "ref{type:id}" contains no whitespace). This is safe **only if** the `onSelect` consumer appends a trailing space after the `}`. Worth documenting as a contract in the `onSelect` callback's JSDoc.

lsm (commented):

### Previous Issues -- Verification

**P1 (docstring/implementation mismatch) -- RESOLVED**
The misleading docstring line about @@ check has been removed. The new comment at line 51-53 of `packages/web/src/hooks/useReferenceAutocomplete.ts` correctly documents the actual behavior: `@@` returns `"@"` as the query string, and consumers should append a trailing space after inserting `@ref{type:id}` to avoid re-triggering. Verified against the implementation -- this is accurate.

**P2 (barrel test not updated) -- RESOLVED**
`useReferenceAutocomplete` and `extractActiveAtQuery` are now imported in `packages/web/src/hooks/__tests__/index.test.ts` (lines 19-20) with corresponding test cases at lines 75-83. The barrel test now covers all 12 exports including the 2 new ones.

**P2 (close() doesn't reset stale results) -- PARTIALLY RESOLVED**
The `close()` callback at lines 169-173 now correctly resets `results` to `[]` and `selectedIndex` to `0`. The test at lines 370-401 properly verifies all three values are reset. However, see new P2 below -- Escape still bypasses `close()`.

### New Issue

**P2: Escape handler bypasses close() -- stale results remain in state**
`packages/web/src/hooks/useReferenceAutocomplete.ts`, lines 207-210.

The Escape key handler calls `setShowAutocomplete(false)` directly instead of calling `close()`:
```typescript
} else if (e.key === 'Escape') {
    e.preventDefault();
    setShowAutocomplete(false);  // <-- should be close()
    return true;
}
```

This means pressing Escape leaves stale `results` and `selectedIndex` in state. Since the P2 fix was motivated by "preventing stale data flash when autocomplete is reopened after Escape", this partially defeats the purpose. If the user presses Escape and then types a new `@` without changing any other content, the effect re-runs and fetches fresh results so the stale data wouldn't be visible -- but there is a window where `showAutocomplete` flips to `true` before new results arrive, and the old results could briefly render.

Additionally, `handleSelect` (line 183) has the same issue -- it sets `showAutocomplete(false)` without resetting `results`/ `selectedIndex`.

**Recommendation**: Replace `setShowAutocomplete(false)` with `close()` in the Escape handler (line 209). Consider also doing the same in `handleSelect` for consistency, or at least reset `selectedIndex` there since stale selectedIndex is more likely to cause issues on re-open than stale results.

**Missing test coverage**: The Escape test (line 617-642) only asserts `showAutocomplete` is `false` but does not verify `results` and `selectedIndex` are reset. A `@@` edge-case test is also absent from the `extractActiveAtQuery` tests despite the docstring explicitly documenting this behavior.

### Overall Assessment

The implementation is well-structured: clean separation of `extractActiveAtQuery` as a pure function, good stale-response cancellation via version counter, proper debouncing, and comprehensive test coverage (64 total tests across 3 files, all passing). The shared reference types are well-designed with clear documentation. The remaining P2 is a minor inconsistency that should be cleaned up before merge.

---
**Recommendation**: REQUEST_CHANGES

lsm (commented):

### Previous P2 Verification: Escape handler fix -- RESOLVED

The three-part fix is correctly implemented:

1. **Escape now calls close()** (line 209 of `useReferenceAutocomplete.ts`): The Escape branch in `handleKeyDown` now delegates to `close()` instead of directly calling `setShowAutocomplete(false)`. This ensures all cleanup (results reset, selectedIndex reset) happens through a single code path.

2. **close added to useCallback deps** (line 215): `close` is properly listed in `handleKeyDown`'s dependency array. Since `close` is a `useCallback` with an empty dependency array, it has a stable identity, so this is correct and won't cause unnecessary re-renders.

3. **Escape test asserts all three resets** (lines 617-653 of the test file): The test navigates to `selectedIndex === 1`, presses Escape, then asserts `showAutocomplete === false`, `results === []`, and `selectedIndex === 0`. This directly verifies the stale-data concern from the original review.

The `close()` function at line 169-173 resets all three state values (`showAutocomplete`, `results`, `selectedIndex`), which prevents the stale flash where old results could briefly appear if autocomplete was reopened.

- **`handleSelect` does not call `close()`** (line 175-186): It calls `setShowAutocomplete(false)` directly instead of `close()`. Since selection is immediately followed by the consumer replacing the @query in content (triggering a re-detection via the effect), this is functionally fine -- the results will be replaced by the next search cycle. However, for consistency it could call `close()`. This is purely cosmetic and not a bug.

- **`useCommandAutocomplete` has the same Escape pattern** (line 108): The existing slash-command hook also calls `setShowAutocomplete(false)` directly in its Escape handler rather than its own `close()`. Not in scope for this PR but worth noting for future consistency.

### Overall Assessment

The implementation is clean and well-tested. The hook correctly handles @ detection anywhere in text, debounced RPC searches, stale response cancellation via version counter, keyboard navigation with wrapping, multiple @ support, and proper cleanup on Escape. The shared types and regex pattern are well-defined with good documentation. All 52 tests pass and typecheck is clean.

**Recommendation: APPROVE**

PR #837, opened 2026-03-25
feat(file-index): implement workspace file index with polling-based cache refresh

lsm:
Adds FileIndex class for fast fuzzy file/folder search without recursive
directory listing on every query. Key features:

- init(): recursive workspace scan, loads .gitignore, starts polling timer
- search(): O(n) fuzzy match scored by exact > starts-with > contains > path-segment
- Polling-based refresh (default 10s, NEOKAI_FILE_INDEX_POLL_MS) with debounce
- .gitignore parsing (*, **, !negation, dir/ suffix) + always-ignore builtins
- Path traversal prevention: rejects .. segments and absolute paths
- setIgnorePatterns() for runtime-configurable extra patterns
- invalidate() / invalidateAll() for manual cache eviction
- 38 unit tests covering all requirements

### DIFF
--- a/packages/daemon/.env.example
+++ b/packages/daemon/.env.example
@@ -79,6 +79,15 @@ MAX_SESSIONS=10
 # NEOKAI_SDK_STARTUP_MAX_RETRIES=2
 
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+# File Index Configuration
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+#
+# How often (ms) the workspace file index polls for changes.
+# Lower values give faster cache freshness at the cost of more I/O.
+# Default: 10000
+# NEOKAI_FILE_INDEX_POLL_MS=10000
+
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 # Job Queue Configuration
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 #
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -25,6 +25,7 @@ import { JobQueueProcessor } from './storage/job-queue-processor';
 import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
 import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
 import { AppMcpLifecycleManager, seedDefaultMcpEntries } from './lib/mcp';
+import { FileIndex } from './lib/file-index';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -74,6 +75,8 @@ export interface DaemonAppContext {
 	jobProcessor: JobQueueProcessor;
 	/** Application-level MCP lifecycle manager — converts registry entries to SDK configs */
 	appMcpManager: AppMcpLifecycleManager;
+	/** Workspace file index for fast fuzzy file/folder search */
+	fileIndex: FileIndex;
 	/**
 	 * Cleanup function for graceful shutdown.
 	 * Closes all connections, stops sessions, and closes database.
@@ -268,6 +271,10 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 
 	// Seed default MCP entries (idempotent — skips entries that already exist)
 	seedDefaultMcpEntries(db);
+
+	// Initialize workspace file index (non-blocking — init runs in the background)
+	const fileIndex = new FileIndex(config.workspaceRoot);
+	void fileIndex.init();
 
 	// Setup RPC handlers (returns cleanup function + exposed services)
 	const {
@@ -512,6 +519,9 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 				providerRegistry.getAll().flatMap((p) => (p.shutdown ? [p.shutdown()] : []))
 			);
 
+			// Stop workspace file index polling
+			fileIndex.dispose();
+
 			// Close database
 			db.close();
 
@@ -542,6 +552,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		jobQueue,
 		jobProcessor,
 		appMcpManager,
+		fileIndex,
 		cleanup,
 	};
 }
--- a/packages/daemon/src/lib/file-index.ts
+++ b/packages/daemon/src/lib/file-index.ts
@@ -1,0 +1,405 @@
+/**
+ * File Index — Workspace file tree cache with polling-based refresh.
+ *
+ * Scans the workspace once on init(), maintains an in-memory cache of all
+ * file and folder entries, and refreshes the cache incrementally via polling.
+ * Designed for fast fuzzy search without recursive directory listing on every query.
+ */
+
+import { join, normalize, relative } from 'node:path';
+import { readdir, readFile } from 'node:fs/promises';
+import { existsSync } from 'node:fs';
+
+export interface FileIndexEntry {
+	/** Relative path from workspace root (e.g. "src/utils/helper.ts") */
+	path: string;
+	/** File or directory name (e.g. "helper.ts") */
+	name: string;
+	type: 'file' | 'folder';
+}
+
+interface IgnorePattern {
+	/** Glob pattern string (after stripping `!` prefix and `/` suffix) */
+	pattern: string;
+	negated: boolean;
+	/** True if the pattern only applies to directories */
+	dirOnly: boolean;
+}
+
+// Always-ignored names regardless of .gitignore
+const BUILTIN_IGNORE_NAMES = new Set(['.git', 'node_modules', '.DS_Store']);
+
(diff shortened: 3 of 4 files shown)


lsm (commented):

---

Overall this is a solid, well-structured implementation. The FileIndex class is cleanly designed, tests are comprehensive (38 passing, good coverage of init/search/invalidation/gitignore/edge-cases), and the code passes both lint and typecheck. However, there are several issues ranging from a correctness bug in glob matching to missing tests and an unused module that should be addressed before merging.

---

### P1 — Symlinks silently dropped (correctness / data completeness)

**File:** `packages/daemon/src/lib/file-index.ts`, lines 239-253 and 276-293

Node.js `readdir` with `{ withFileTypes: true }` returns `isDirectory() === false` for symlinks pointing to directories. This means directory symlinks are silently excluded from the index — they are neither indexed as folders nor recursed into. In a typical monorepo workspace, symlinks are common (e.g., `node_modules` package links, Yarn workspaces, `tsconfig` path aliases via symlinks).

This is not a security issue (symlinks won't cause traversal since the entry isn't followed), but it is a completeness gap. The FileIndex will be missing entries that users expect to find via search.

Consider using `entry.isDirectory() || entry.isSymbolicLink()` to at least capture symlink entries, or stat the symlink target to determine its type. At minimum, document this behavior so callers know symlinks are excluded.

---

### P1 — `**/prefix` glob pattern does not match at root level (correctness)

**File:** `packages/daemon/src/lib/file-index.ts`, lines 64-67, 93-96

When `parseGitignoreLines` processes a pattern like `**/foo`, it produces `{ pattern: '**/foo' }`. Since the pattern contains `/`, it routes to `matchFullPath`. The regex generated is `^.*/foo$`, which requires at least one `/` before `foo`. In real .gitignore semantics, `**/foo` matches `foo` at any depth **including the root**.

Test:
```
^.*/foo$ on 'foo'        => false  (BUG: should be true)
^.*/foo$ on 'src/foo'    => true   (correct)
```

The fix: when `**` appears at the start of a pattern, make the preceding `.*` and trailing `/` optional: use `(?:.*/)?foo` instead of `.*/foo`.

This also applies to `**` in the middle of patterns (e.g., `src/**/test`) — after consuming the `/` following `**`, the regex correctly handles intermediate depths. Only the leading `**/` case has the off-by-one.

---

### P1 — `setIgnorePatterns` after `init()` does not update cache (API design / usability)

**File:** `packages/daemon/src/lib/file-index.ts`, line 391-393

The current test only covers calling `setIgnorePatterns()` **before** `init()`. But there is no test or documentation about calling it **after** `init()`. If a caller sets patterns after init, the cache already contains entries that should be ignored, and the next `runRefresh()` will also not remove them (the refresh adds entries but only checks `shouldIgnore` for new entries it encounters on disk — it never re-evaluates existing cache entries against updated patterns).

The `allPatterns` getter is reactive (line 219-221), so new patterns are used on **future** refreshes for newly discovered entries. But existing cache entries that should now be ignored will persist until they happen to be deleted from disk and recreated.

This should either be documented clearly, or `setIgnorePatterns()` should trigger a re-filter of the cache (or at minimum set a dirty flag).

---

### P1 — `setIgnorePatterns` does not re-evaluate on refresh (correctness)

**File:** `packages/daemon/src/lib/file-index.ts`, lines 297-313

The `runRefresh()` method walks the directory tree and adds new entries via `seen.add(relPath)`, then deletes entries not in `seen`. However, it never re-checks whether **existing** cache entries still pass `shouldIgnore()` with the current patterns. If extra patterns are added after init (via `setIgnorePatterns`), cached entries that now match those patterns will never be removed by refresh.

Steps to reproduce:
1. Create file `secret.key` in workspace
2. Call `init()` — file is indexed
3. Call `setIgnorePatterns(['*.key'])`
4. Call `runRefresh()` — `secret.key` remains in cache

The fix would be to filter out existing cache entries that should now be ignored during the cleanup phase of `runRefresh()`.

---

### P2 — `**` glob tests missing from test suite (test coverage)

**File:** `packages/daemon/tests/unit/file-index.test.ts`

The .gitignore tests cover `*`, `dir/` suffix, and `!negation`, but there are no tests for:
- `**` double-star patterns (e.g., `build/**`, `**/test/**`)
- `?` single-character wildcard
- Patterns with special regex characters (e.g., `foo[bar]`, `file (1).txt`)
- Leading `/` anchored patterns (e.g., `/build` only matches at root)

Given that the glob matching is hand-rolled rather than using a library, these edge cases should have explicit test coverage.

---

### P2 — `setIgnorePatterns` after `init` not tested (test coverage)

**File:** `packages/daemon/tests/unit/file-index.test.ts`, lines 341-353

The only test for `setIgnorePatterns` calls it before `init()`. There should be a test verifying behavior when called after `init()` — even if the current behavior is intentional (cache not immediately updated), this should be an explicit test with a documented expectation.

---

### P2 — `FileIndex` is exported but never integrated (dead code concern)

**File:** `packages/daemon/src/lib/file-index.ts`

The `FileIndex` class is exported but not imported anywhere in the codebase outside of the test file. The PR description says this enables "fast fuzzy file/folder search" but there is no RPC handler, no wiring in `app.ts`, and no frontend integration.

While I understand this may be a foundation-laying PR intended for follow-up integration, the `file-index.ts` module should either:
1. Be integrated into the daemon (even minimally — e.g., wired into `DaemonAppContext` and initialized in `createDaemonApp`), or
2. Be clearly marked as a standalone utility pending integration

Without integration, the polling timer in `init()` will never actually run in production, and `NEOKAI_FILE_INDEX_POLL_MS` in `.env.example` is a dead configuration option. This also means `knip` should flag the unused export if it runs on this code.

---

### P3 — `matchSegment` and `matchFullPath` are nearly identical (DRY)

**File:** `packages/daemon/src/lib/file-index.ts`, lines 58-111

These two functions share identical glob-to-regex conversion logic (lines 62-79 vs 91-107). The only difference is the input: one matches a single segment, the other matches a full path. Consider extracting the regex-building logic into a shared `globToRegex(pattern: string): RegExp` helper and using it in both functions. This reduces the surface area for future bugs.

---

### P3 — Score 40 and 60 can overlap (minor logic clarity)

**File:** `packages/daemon/src/lib/file-index.ts`, lines 153-168

A file like `src/index-format.ts` searched with query `index` would score 60 (name contains) rather than 40 (path segment contains), since the name check comes first. This is correct, but the comment on line 149 says "any path segment contains query" — the scoring naturally prioritizes name-level matches, which is fine. Just noting this is intentional and the comment could be clearer that name-level matches always take precedence.

---

### Summary

The core architecture (in-memory cache, polling refresh, fuzzy scoring, ignore patterns) is sound. The P1 issues are:
1. Symlinks silently excluded (data gap)
2. `**/prefix` pattern doesn't match root (glob correctness)
3. `setIgnorePatterns` after init leaves stale entries (API correctness)
4. Refresh doesn't re-filter against updated patterns (correctness)

The P2 issues are test coverage gaps (no `**` tests, no post-init pattern test) and the fact that the module is not yet wired into the application.

lsm (commented):

**Recommendation: APPROVE**

All previously identified P1 and P2 issues have been properly addressed.

### P1 Issues -- All Fixed

1. **Symlinks silently dropped**: Fixed. Both `scanDirectory` (line 258) and `refreshDirectory` (line 305) now detect symlinks via `entry.isSymbolicLink()`, resolve target type with `stat()`, and index the entry as file/folder. Symlinked directories are explicitly not recursed into (the `continue` statement prevents infinite loops). Broken symlinks are silently skipped. Tests added at lines 597-628 cover both symlinked files and symlinked directories.

2. **`**/prefix` glob doesn't match root**: Fixed. The `buildGlobRegex()` helper (line 80) now generates `(?:.*/)?` for `**/` patterns instead of the previous `.*/`. This makes the directory prefix optional, so `**/tests` correctly matches both `tests` at root and `src/tests` at any nested depth. The optional-group regex is verified to work for arbitrary nesting depth because `.*` is greedy and matches slashes. Tests at lines 354-384 cover root depth, nested depth, and `build/**` patterns.

3. **`setIgnorePatterns` after init leaves stale cache**: Fixed. `setIgnorePatterns()` (lines 445-453) now immediately iterates the cache and deletes entries that match the updated patterns. The test at line 418 verifies this explicitly.

4. **Refresh doesn't re-filter against updated patterns**: Fixed. `runRefresh()` (lines 356-361) now has a second pass after removing stale entries that re-evaluates remaining cache entries against the current ignore patterns. The test at line 436 confirms that after `setIgnorePatterns` + `runRefresh`, excluded entries are not re-added.

### P2 Issues -- All Fixed

1. **`**` glob patterns lack test coverage**: Added tests for `build/**` (line 338), `**/tests` at root (line 354) and nested (line 373) depths, plus `?` single-character wildcard (line 386). All pass.

2. **Module exported but not integrated into daemon**: `FileIndex` is now imported in `app.ts` (line 28), added to `DaemonAppContext` interface (line 79), instantiated with `config.workspaceRoot` (line 276, fire-and-forget init), and disposed in cleanup (line 522). The non-blocking `void fileIndex.init()` pattern is appropriate since the index is not required for startup.

### Code Quality

- The `buildGlobRegex()` refactor eliminates the duplicated pattern-building logic between `matchSegment` and `matchFullPath` -- clean improvement.
- Map delete-during-iteration is safe in JavaScript (verified).
- Regex escaping uses explicit character checks instead of a regex-inside-regex approach -- more maintainable.
- All 46 unit tests pass, lint passes, and typecheck passes.

### Minor Note (P3)

- The `fileIndex` is wired into `DaemonAppContext` but not yet exposed via any RPC handler or consumed by the frontend. This is fine for the initial infrastructure commit, but the next step should expose a search RPC so consumers can actually use it.

PR #838, opened 2026-03-25
feat(agent): create ReferenceResolver service and shared reference types

lsm:
- Add packages/shared/src/types/reference.ts: ReferenceType, ReferenceMention,
  ResolvedReference (and subtypes), ReferenceMetadata, REFERENCE_PATTERN
- Export new types from shared package mod.ts
- Add packages/daemon/src/lib/agent/reference-resolver.ts: ReferenceResolver
  class with extractReferences(), resolveReference(), resolveAllReferences();
  dependency-injected ITaskRepository, IGoalRepository, IFileManager interfaces;
  path traversal prevention for file/folder references
- Add 38 unit tests covering parsing, task/goal/file/folder resolution,
  standalone sessions (no room context), and path traversal rejection

Part of M1 Task 1.5 — system-wide @ referencing system

### DIFF
--- a/packages/shared/src/types/reference.ts
+++ b/packages/shared/src/types/reference.ts
@@ -6,6 +6,8 @@
  * in chat inputs using @ref{type:id} syntax.
  */
 
+import type { NeoTask, RoomGoal } from './neo.ts';
+
 /**
  * The type of entity being referenced.
  * Union type (not enum) following existing codebase patterns like `SessionType`.
@@ -24,6 +26,8 @@ export interface ReferenceMention {
 
 /**
  * A single result returned from the reference.search RPC.
+ * Used by the reference.search RPC handler (M1 Task 1.3) and the
+ * useReferenceAutocomplete hook (M2).
  */
 export interface ReferenceSearchResult {
 	type: ReferenceType;
@@ -48,37 +52,50 @@ export interface ResolvedReference {
 
 export interface ResolvedTaskReference extends ResolvedReference {
 	type: 'task';
+	data: NeoTask;
 }
 
 export interface ResolvedGoalReference extends ResolvedReference {
 	type: 'goal';
+	data: RoomGoal;
+}
+
+/**
+ * File content payload returned for a resolved file reference.
+ * Mirrors the return shape of the resolve RPC handler in
+ * packages/daemon/src/lib/rpc-handlers/reference-handlers.ts.
+ */
+export interface ResolvedFileData {
+	path: string;
+	/** UTF-8 text content, or null when the file is binary */
+	content: string | null;
+	/** True when the file contains binary (non-text) data; content will be null */
+	binary: boolean;
+	/** True when file content was truncated to stay within payload limits */
+	truncated: boolean;
+	size: number;
+	mtime: string;
+}
+
+/** A single entry in a resolved folder listing. */
+export interface FolderEntry {
+	path: string;
+	name: string;
+	/** 'directory' matches FileManager.FileInfo and the real filesystem API */
+	type: 'file' | 'directory';
 }
 
 /** Resolved file reference — data includes content (possibly truncated or absent for binary files) */
 export interface ResolvedFileReference extends ResolvedReference {
 	type: 'file';
-	data: {
-		path: string;
-		/** UTF-8 text content, or null when the file is binary */
-		content: string | null;
-		/** True when the file contains binary (non-text) data; content will be null */
-		binary: boolean;
-		/** True when file content was truncated to stay within payload limits */
-		truncated: boolean;
-		size: number;
-		mtime: string;
-	};
+	data: ResolvedFileData;
 }
 
 export interface ResolvedFolderReference extends ResolvedReference {
 	type: 'folder';
 	data: {
 		path: string;
-		entries: Array<{
-			name: string;
-			path: string;
-			type: 'file' | 'directory';
-		}>;
+		entries: FolderEntry[];
 	};
 }
 
--- a/packages/shared/tests/reference.test.ts
+++ b/packages/shared/tests/reference.test.ts
@@ -10,6 +10,7 @@ import type {
 	ResolvedTaskReference,
 	ReferenceType,
 } from '../src/types/reference.ts';
+import type { NeoTask, RoomGoal } from '../src/types/neo.ts';
 import { REFERENCE_PATTERN } from '../src/types/reference.ts';
 
 describe('ReferenceType', () => {
@@ -58,12 +59,20 @@ describe('ReferenceSearchResult', () => {
 
 describe('ResolvedReference variants', () => {
 	it('ResolvedTaskReference has type task', () => {
-		const ref: ResolvedTaskReference = { type: 'task', id: 't-1', data: { title: 'A task' } };
+		const ref: ResolvedTaskReference = {
+			type: 'task',
+			id: 't-1',
+			data: { title: 'A task' } as unknown as NeoTask,
+		};
 		expect(ref.type).toBe('task');
 	});
 
 	it('ResolvedGoalReference has type goal', () => {
-		const ref: ResolvedGoalReference = { type: 'goal', id: 'g-1', data: {} };
+		const ref: ResolvedGoalReference = {
+			type: 'goal',
+			id: 'g-1',
+			data: {} as unknown as RoomGoal,
+		};
 		expect(ref.type).toBe('goal');
 	});
 


lsm (commented):

---

## Overall Assessment

This is a solid foundation for the M1 Task 1.5 deliverable. The code is well-structured, cleanly typed, and properly testable via dependency injection. All 38 tests pass, lint/typecheck/knip are clean, and the design doc acceptance criteria are largely met. I have a few issues to flag before this merges.

---

## P1 Issues (Should Fix)

### 1. `..` substring check is too aggressive in `validatePath`

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, line 333

```ts
if (inputPath.includes('..')) {
```

Using `includes('..')` rejects any path containing a double-dot **anywhere**, including legitimate filenames like `src/my..config.ts`, `src/...staging.ts`, or `src/..hidden/` (a directory name starting with dots). The `..` prefix check in the existing `FileManager.validatePath()` (line 32-43 of `packages/daemon/src/lib/file-manager.ts`) does this correctly by only checking the **normalized relative path** after joining with workspace:

```ts
const rel = relative(normalizedWorkspace, normalizedTarget);
if (rel.startsWith('..') || rel === '..') { ... }
```

**Recommendation:** Remove the `includes('..')` pre-check and rely entirely on the post-normalization `rel.startsWith('..')` check (which already exists at line 349). The normalization of `join(workspace, path)` will correctly resolve `src/../etc/passwd` to an escaped path, while preserving `src/my..file.ts` as a valid in-workspace path.

### 2. `IFileManager` interface is not structurally aligned with the existing `FileManager` class

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, lines 50-56

The `IFileManager.readFile` return type is `Promise<ResolvedFileData>` while the existing `FileManager.readFile` (in `packages/daemon/src/lib/file-manager.ts`, line 48) returns `Promise<{ path, content, encoding, size, mtime }>`. These are structurally the same. However, `IFileManager.listDirectory` has `dirPath?: string` (optional) while `FileManager.listDirectory` has `dirPath: string = '.'` (required with default). TypeScript considers this compatible, so it will compile -- but the `IFileManager` declares `listDirectory` as returning `Array<FolderEntry & { size?: number; mtime?: string }>` while the actual `FileManager.listDirectory` returns `FileInfo[]` which uses a **different** `FileInfo` interface from `file-manager.ts` (line 11-17). The shapes match structurally, but this is a subtle coupling risk.

When the RPC handlers wire up real dependencies in M3, they will need to adapt `FileManager` to satisfy `IFileManager`, or `ReferenceResolver` will need to depend on `FileManager` directly. Consider documenting this adapter expectation, or making `IFileManager` match `FileManager`'s actual signature more closely.

### 3. `resolveReference` docstring says "Throws on unrecoverable errors" but the implementation catches all errors

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, lines 147-151

The JSDoc on `resolveReference` states:
> Throws on unrecoverable errors (e.g. file I/O failures).

But both `resolveFile` (line 274-283) and `resolveFolder` (line 297-314) wrap their I/O calls in try/catch and return `null`. This means `resolveReference` itself never actually throws, contradicting the documented contract. The `resolveAllReferences` method (line 201-205) adds a second try/catch around `resolveReference`, suggesting the author expected it could throw.

**Recommendation:** Either update the JSDoc to say "Returns null on all errors" or re-throw specific error types (e.g., permission errors) so callers can distinguish between "not found" and "server error".

---

## P2 Issues (Important Suggestions)

### 4. No test for duplicate reference deduplication in `resolveAllReferences`

**File:** `packages/daemon/tests/unit/reference-resolver.test.ts`

If a user writes `@ref{task:t-1} and @ref{task:t-1}`, the current implementation will resolve the same reference twice (calling the repo twice) and the second write to `result['task:t-1']` will silently overwrite the first. This is not a correctness bug since the results are identical, but it is wasted work. A test documenting this behavior (or a dedup optimization) would be valuable.

### 5. Missing test for empty/whitespace-only `id` in references

The regex `([^}]+)` matches one or more non-`}` characters, so `@ref{task:}` is correctly rejected by the regex. However, there is no test for `@ref{task: }` (single space). The regex matches this as `id = ' '`, which passes `isReferenceType('task')` but would fail the repo lookup. This is handled correctly at runtime (returns null), but a test would document the behavior.

### 6. `ResolvedFileData` is defined in `reference.ts` but mirrors `FileManager`'s return type

**File:** `packages/shared/src/types/reference.ts`, lines 91-97

`ResolvedFileData` duplicates the shape returned by `FileManager.readFile` in `packages/daemon/src/lib/file-manager.ts`. If `FileManager`'s return type changes, `ResolvedFileData` would silently drift. Consider importing the return type from `file-manager.ts` (or moving the shared shape to `@neokai/shared`) to avoid this duplication.

### 7. `ReferenceSearchResult` type is exported but never used or tested

**File:** `packages/shared/src/types/reference.ts`, lines 45-56

This type is defined for the `reference.search` RPC handler (M1 Task 1.3) but has no usage in this PR and no test coverage. Knip currently passes because the barrel export in `mod.ts` re-exports everything, so it is technically "used" by downstream consumers. However, if no consumers exist yet, this is premature export. Consider adding a `// @used-in: reference-handlers.ts (Task 1.3)` comment or holding it until Task 1.3.

---

## P3 Notes (Nits)

### 8. The exhaustive check in the switch default case uses `this.log.warn` on a `never` value

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, lines 166-171

This is correct and will never execute at runtime (TypeScript ensures exhaustiveness). The pattern is fine, but note that a production build may tree-shake the dead code. Not a problem -- just noting the intent.

### 9. Test fixture `makeTask` and `makeGoal` include optional `shortId` as required

**File:** `packages/daemon/tests/unit/reference-resolver.test.ts`, lines 27-58

The `NeoTask` type declares `shortId?: string` (optional), but the test fixtures always provide it. This is correct for testing purposes (you need the shortId to exist), so this is just an observation, not an issue.

---

## What's Done Well

- **Clean DI architecture**: The `ReferenceResolverDeps` interface with optional repos makes the service fully testable. Mock factories in tests are concise and readable.
- **Exhaustive switch**: The `never` type check for unhandled reference types is a great pattern.
- **Fresh regex per call**: The `new RegExp(REFERENCE_PATTERN.source, 'g')` in `extractReferences` correctly avoids shared `lastIndex` state. Many developers miss this.
- **Partial results in `resolveAllReferences`**: Failed resolutions log warnings but don't block other references. This matches the design doc's error-handling requirements.
- **Path validation defense in depth**: The `includes('..')` pre-check + post-normalization relative path check provides two layers of defense (even if the first is too aggressive per point #1).
- **Test organization**: Clear describe blocks with separate sections for each reference type and scenario. 38 tests with good coverage of the happy paths and key edge cases.

---

**Recommendation:** REQUEST_CHANGES -- Address P1 items #1 (path traversal false positives) and #3 (docstring contract mismatch) before merging. Items #2, #4-7 can be addressed in follow-up tasks but would be good to acknowledge.

lsm (commented):

---

## Re-review of Previous P1/P2 Findings

All three P1 issues from the previous review have been correctly addressed:

### P1-1 (Path validation `includes('..')` false positive): FIXED
The overly aggressive `includes('..')` pre-check has been removed from `validatePath()`. The method now relies solely on the post-normalization `relative().startsWith('..')` check, matching the approach in `FileManager.validatePath()`. The new test `'allows filenames that contain ".." as a substring'` at line 403 of the test file confirms that paths like `src/my..config.ts` are now correctly accepted.

### P1-2 (IFileManager misaligned with FileManager): FIXED
The `IFileManager.listDirectory` signature now declares `dirPath: string` (required) and returns `Promise<FileInfo[]>` using the `FileInfo` type imported from `../file-manager`. The real `FileManager.listDirectory(dirPath: string = '.')` is structurally compatible (TypeScript treats default-value parameters as assignable to required parameters). The `IFileManager` docstring at line 49-52 now explicitly states structural compatibility with `FileManager`.

### P1-3 (Docstring contract mismatch): FIXED
The `resolveReference` JSDoc at line 158 now correctly states "Never throws -- all errors are caught and returned as null with a warning log.", matching the actual try/catch implementation in `resolveFile` and `resolveFolder`.

### P2-4 (Missing dedup test): FIXED
New test at line 559, `'handles duplicate references -- last write wins but result is identical'`, confirms that two identical `@ref{task:t-1}` mentions collapse to a single key in the result map.

### P2-5 (Missing whitespace-only id test): FIXED
New test at line 572, `'returns null for references with whitespace-only id'`, confirms that `@ref{task: }` is parsed but fails the repo lookup and is excluded from results.

### P2-6 (ResolvedFileData mirrors FileManager): ACKNOWLEDGED via comment
A comment has been added at lines 92-95 of `reference.ts` noting the intentional mirror of `FileManager.readFile`'s return shape with an instruction to update if that type changes.

### P2-7 (ReferenceSearchResult unused): ACKNOWLEDGED via comment
A comment has been added at lines 43-46 of `reference.ts` noting the type will be used by the `reference.search` RPC handler (M1 Task 1.3) and the `useReferenceAutocomplete` hook (M2).

---

## New Observations (P2)

### 1. `validatedPath` is computed but never passed to the file manager

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, lines 267-276 and 290-299

In both `resolveFile()` and `resolveFolder()`, the return value of `this.validatePath()` is stored in `validatedPath` but only its truthiness is checked. The raw `mention.id` (un-normalized relative path) is passed to `this.deps.fileManager.readFile(mention.id)` and `this.deps.fileManager.listDirectory(mention.id)` instead of the validated/normalized path.

This is not a security issue today because the real `FileManager` performs its own path validation internally. However, if a future `IFileManager` implementation does not validate paths, the defense provided by `validatePath()` would be bypassed. Consider passing `validatedPath` (or computing a relative path from it) to the file manager, or at minimum adding a comment noting that `FileManager` provides a second layer of path validation.

This is P2 because the current production wiring (FileManager) is safe, and the validatePath gate still prevents traversal from reaching the file manager.

---

## Verification

- All 41 unit tests pass (up from 38)
- TypeScript type checking passes cleanly
- Lint (oxlint): 0 warnings, 0 errors
- Knip: no new issues
- No new regressions in the fix commit

---

**Recommendation:** APPROVE -- All P1 issues are resolved correctly, P2 items are addressed with comments or new tests, and no new blocking issues were found. The one new P2 observation (validatedPath not forwarded) is a defense-in-depth concern, not a current bug.

lsm (commented):

### Verification of Prior Review Items

**P2 fix (validated paths): Verified clean.**
- `validatePath()` at line 338 of `packages/daemon/src/lib/agent/reference-resolver.ts` now returns the normalized relative path (line 358: `return rel;`) instead of a boolean.
- `resolveFile()` (line 267) captures `const validatedPath = this.validatePath(...)` and passes it directly to `this.deps.fileManager.readFile(validatedPath)` at line 276.
- `resolveFolder()` (line 290) similarly passes `validatedPath` to `this.deps.fileManager.listDirectory(validatedPath)` at line 299, and uses it for the returned `data.path` at line 308.
- This correctly ensures the validated, normalized path is what reaches the file manager -- no raw `mention.id` can slip through. The fix is complete and correct.

**Merge conflict resolution in `packages/shared/src/types/reference.ts`: Verified clean.**
- The merged version correctly takes dev's original prose/comments from PR #835 as the base and layers on the typed data fields this branch needs.
- `ResolvedTaskReference` now has `data: NeoTask`, `ResolvedGoalReference` has `data: RoomGoal` -- these are properly imported from `./neo.ts` (line 8).
- `ResolvedFileData` (lines 72-78) and `FolderEntry` (lines 81-85) are well-defined interfaces. The `ResolvedFileData` comment correctly notes it mirrors `FileManager.readFile()`'s return shape.
- The section separators and JSDoc comments follow the same style as the base version.

**Test fixes in `packages/shared/tests/reference.test.ts`: Verified clean.**
- The file/folder resolved reference tests now construct proper typed data objects matching the new interfaces (lines 99-126).
- All 19 shared tests and 41 daemon tests pass (60 total, 0 failures).
- TypeScript typecheck and lint both pass with zero errors.

### New Issue: Upcoming Merge Conflict with Dev (P1)

This branch is 3 commits ahead and 1 commit behind dev. The dev commit `cb0aa1f7a` (PR #836 -- `useReferenceAutocomplete`) modified `packages/shared/src/types/reference.ts` to add its own inline shapes for resolved reference data:

- Dev: `ResolvedFileReference.data = { path: string; content: string }`
- This branch: `ResolvedFileReference.data = ResolvedFileData` (includes `encoding, size, mtime` too)

- Dev: `ResolvedFolderReference.data.entries = Array<{ name: string; type: 'file' | 'folder' }>`
- This branch: `ResolvedFolderReference.data.entries = FolderEntry[]` where `FolderEntry` has `path, name, type: 'file' | 'directory'`

Key differences:
1. The `type` field uses `'folder'` in dev vs `'directory'` in this branch. The real `FileManager` and `FileInfo.type` use `'directory'`, so this branch is correct -- but dev's version will need updating when these branches merge.
2. This branch's `ResolvedFileData` is a strict superset of dev's inline shape (more fields), so it is backward compatible.
3. This branch's `FolderEntry` has a `path` field that dev's inline type lacks.

Running `git merge-tree` confirms ~8 conflict hunks in `reference.ts` and ~2 in `reference.test.ts`. These are all resolvable (take this branch's typed interfaces over dev's inline shapes), but the branch should be rebased onto dev before merge to surface and resolve these conflicts explicitly. The conflicts are mechanical -- not indicative of a design disagreement -- but leaving them for an implicit merge risks silently choosing the wrong side.

### Minor Notes (P3)

- The exhaustive type guard in `resolveReference()` (line 167-173 of `reference-resolver.ts`) uses a `never` assertion on `mention.type`. This is correct TypeScript practice for exhaustive switch/case, but the `as never` cast means it will only be caught if someone adds a new `ReferenceType` value without updating the switch. This is fine since the `VALID_TYPES` set at line 89 would also need updating.

### Summary

The P2 fix is properly implemented, the merge conflict resolution is clean, and the code is well-tested. The one actionable item is that this branch needs a rebase onto dev to resolve the incoming conflicts from PR #836 before it can be merged cleanly. This is a process step, not a code defect.

**Recommendation: APPROVE** (pending rebase onto dev)

PR #839, opened 2026-03-25
feat(reference): implement reference.resolve RPC handler with shared types

lsm:
- Add shared reference type definitions in packages/shared/src/types/reference.ts:
  ReferenceType union, ReferenceMention, ReferenceSearchResult, ResolvedReference
  variants (task/goal/file/folder), ReferenceMetadata, and REFERENCE_PATTERN regex
- Create packages/daemon/src/lib/rpc-handlers/reference-handlers.ts with the
  reference.resolve RPC handler: resolves tasks (by UUID or short ID) and goals
  from room context, reads file content (with 50KB truncation), and lists folder
  entries — returns null gracefully for missing entities, wrong room, or path
  traversal attempts
- Wire setupReferenceHandlers() into setupRPCHandlers() in rpc-handlers/index.ts
- Add 21 unit tests covering all entity types, null returns, path traversal
  prevention, large file truncation, and session fallback behavior

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -73,6 +73,7 @@ import { setupGlobalSpacesHandlers } from './global-spaces-handlers';
 import type { GlobalSpacesState } from '../space/tools/global-spaces-tools';
 import { setupSpaceSessionGroupHandlers } from './space-session-group-handlers';
 import { setupLiveQueryHandlers } from './live-query-handlers';
+import { setupReferenceHandlers } from './reference-handlers';
 import { LiveQueryEngine } from '../../storage/live-query';
 import type { AppMcpLifecycleManager } from '../mcp';
 
@@ -271,6 +272,14 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		deps.liveQueries,
 		deps.db.getDatabase()
 	);
+
+	// Reference handlers (@ mention system — resolve tasks, goals, files, folders)
+	setupReferenceHandlers(deps.messageHub, {
+		sessionManager: deps.sessionManager,
+		taskRepo: new TaskRepository(deps.db.getDatabase(), deps.reactiveDb),
+		goalRepo: deps.db.getGoalRepo(),
+		workspaceRoot: deps.config.workspaceRoot,
+	});
 
 	// Space handlers (spaceManager injected from deps — single instance shared with DaemonAppContext)
 	const spaceTaskRepo = new SpaceTaskRepository(deps.db.getDatabase());
--- a/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
@@ -15,6 +15,7 @@ import type {
 import type { SessionManager } from '../session-manager';
 import { FileManager } from '../file-manager';
 import { Logger } from '../logger';
+import { join, normalize, relative } from 'node:path';
 
 const log = new Logger('reference-handlers');
 
@@ -23,6 +24,12 @@ const log = new Logger('reference-handlers');
  * Files larger than this will be truncated to prevent oversized responses.
  */
 const MAX_FILE_CONTENT_BYTES = 50_000;
+
+/**
+ * Number of bytes to sample when detecting binary files.
+ * Checking the first 8 KB is sufficient for reliable binary detection.
+ */
+const BINARY_DETECTION_SAMPLE_BYTES = 8_192;
 
 // ============================================================================
 // Repository interfaces (minimal — only what this handler needs)
@@ -165,6 +172,11 @@ async function resolveTask(
 		return null;
 	}
 
+	// Confirm the task belongs to the session's room (prevent cross-room access via UUID)
+	if (task.roomId !== roomId) {
+		return null;
+	}
+
 	return {
 		type: 'task',
 		id,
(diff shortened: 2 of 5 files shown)


lsm (commented):

Overall the implementation is clean, well-structured, and follows existing patterns. Tests are comprehensive (21 pass, 0 fail), types are correct, linting passes, and path traversal prevention delegates properly to `FileManager`. However, there is one P1 security isolation bug that should be fixed before merging.

---

### P1 -- Missing room isolation check in `resolveTask`

**File:** `packages/daemon/src/lib/rpc-handlers/reference-handlers.ts`, lines 149-173

`resolveTask` does not verify that the returned task belongs to the session's room, while `resolveGoal` (lines 175-203) does perform this check (`goal.roomId !== roomId`).

The issue: `getTask(id)` is a global lookup (no room scoping). If a caller passes a task UUID from another room, the handler happily returns it. While `getTaskByShortId` is room-scoped (filters by roomId in SQL), the UUID path is not.

This breaks room data isolation -- a session in room A could resolve a task that belongs to room B by its UUID.

```typescript
// resolveTask is missing this guard (present in resolveGoal):
if (task.roomId !== roomId) {
    return null;
}
```

This should also be covered by a test (similar to the existing "returns null when goal belongs to a different room" test).

---

### P2 -- No binary file handling consideration

**File:** `packages/daemon/src/lib/rpc-handlers/reference-handlers.ts`, line 219

The handler reads all files as UTF-8. Binary files (images, compiled assets) will produce garbled/mojibake content in the response. Consider either:

- Checking the file extension or MIME type and returning null (or a placeholder) for binary files
- Returning `size` and `mtime` metadata without `content`` for files that can't be decoded cleanly

This isn't strictly a bug since the outer try/catch would only fail on truly unreadable files, but sending garbled binary content over WebSocket payloads is wasteful and potentially confusing for clients.

---

### P3 -- Unnecessary type cast in `resolveFolder`

**File:** `packages/daemon/src/lib/rpc-handlers/reference-handlers.ts`, line 252

```typescript
type: e.type as 'file' | 'directory',
```

`FileInfo.type` is already typed as `'file' | 'directory'` (see `file-manager.ts` line 14). The cast is redundant.

---

### P3 -- `SpaceTask` import is forward-looking but unused in handlers

**File:** `packages/shared/src/types/reference.ts`, line 10

`SpaceTask` is imported and used in `ResolvedTaskReference.data` but no handler currently produces a `SpaceTask` resolved reference. The handler only resolves `NeoTask`. This is fine as forward-looking design, but Knip may flag it if `SpaceTask` isn't otherwise re-exported through this module's consumers.

---

### What's done well

- **Interface segregation**: `TaskRepoForReference` and `GoalRepoForReference` are minimal interfaces with only the methods needed, making the handler easy to test with mocks.
- **Graceful error handling**: The outer try/catch in the RPC handler catches unexpected errors and returns null, preventing unhandled rejections from crashing the request.
- **Test coverage**: 21 tests covering all entity types, null returns, path traversal, truncation, session fallback, and wrong-room isolation for goals. Tests are well-structured with clear describe blocks.
- **Integration wiring**: Handler is properly wired into `setupRPCHandlers()` with the correct dependencies.
- **Path traversal security**: Delegates to `FileManager.validatePath()` which uses `normalize` + `relative` checks, consistent with all other file-handling handlers.
- **Code style**: Adheres to Biome formatting, passes Oxlint, passes TypeScript strict checks.

lsm (commented):

### Summary

Reviewed the two commits on this PR implementing the `reference.resolve` RPC handler and the follow-up fixes for cross-room task isolation and binary file detection.

### P1 Fix: Cross-Room Task Isolation

The fix at `packages/daemon/src/lib/rpc-handlers/reference-handlers.ts:175-178` correctly prevents cross-room task access. After resolving a task by UUID or short ID, the handler now checks `task.roomId !== roomId` and returns null if they differ. This matches the existing pattern in `resolveGoal` (line 207-209).

The test at line 314-336 of the test file properly verifies this behavior by creating a task with `roomId: 'other-room'` and confirming the handler returns null when the session's room is `'room-1'`.

**Verified correct.**

### P2 Fix: Binary File Detection

The binary detection implementation in `resolveFile` (lines 236-279) is sound:
- Samples the first 8KB of the file using `Buffer.allocUnsafe` + `fd.read` (safe because the full buffer is overwritten)
- Checks for null bytes (`buf.includes(0x00)`) -- a reliable binary indicator
- Returns metadata-only response with `content: null, binary: true` for binary files
- Falls through to normal text read via `FileManager.readFile` for text files
- File descriptor is properly closed in a `finally` block
- All error paths (file not found, permission denied) return null

The `ResolvedFileReference` type in `packages/shared/src/types/reference.ts` was correctly updated to `content: string | null` with the new `binary: boolean` field.

The test writes a buffer with null bytes to `image.png` and verifies the handler returns `binary: true` with `content: null`.

**Verified correct.**

### Additional Observations (non-blocking)

- **P3 (style)**: The dynamic `import('node:fs/promises')` calls at lines 241 and 251 of `reference-handlers.ts` could be static imports at the top of the file since `node:fs/promises` is already used transitively via `FileManager`. This is purely stylistic; dynamic imports work fine.

- **P3 (pre-existing)**: The path validation in both `reference-handlers.ts` and `FileManager.validatePath()` uses `normalize()` which does not resolve symlinks. A symlink within the workspace pointing outside could bypass traversal checks. This is a pre-existing pattern across the codebase, not introduced by this PR.

### Verification

- All 23 tests pass
- TypeScript type checking passes cleanly
- Lint passes with zero warnings/errors
- Shared types correctly exported from `packages/shared/src/mod.ts`
- Repository interfaces (`TaskRepoForReference`, `GoalRepoForReference`) match actual `TaskRepository`/ `GoalRepository` method signatures
- Both `NeoTask.roomId` and `RoomGoal.roomId` are `string` (non-optional), so the isolation check can never silently skip due to undefined

### Recommendation

**APPROVE** -- Both fixes are correct, well-tested, and follow existing patterns. No regressions introduced.

PR #840, opened 2026-03-25
feat(reference): implement task and goal resolvers with shared reference types

lsm:
- Add shared reference types to packages/shared/src/types/reference.ts:
  ReferenceType, ReferenceMention, ReferenceSearchResult, ResolvedReference,
  ResolvedTaskReference, ResolvedGoalReference, ResolvedFileReference,
  ResolvedFolderReference, ReferenceMetadata, and REFERENCE_PATTERN regex
- Export new types from @neokai/shared barrel
- Add ReferenceResolver service (packages/daemon/src/lib/agent/reference-resolver.ts)
  with extractReferences(), resolveReference(), and resolveAllReferences() methods
- Room task resolution via TaskRepository.getTaskByShortId (t- prefix)
- Space task resolution via SpaceTaskRepository.getTask (st-<uuid> prefix)
- Goal resolution via GoalRepository.getGoalByShortId (g- prefix)
- Cross-room/space reference prevention enforced at resolve time
- Graceful null return (no throws) for missing entities or missing context
- File/folder resolution left as stub (implemented in Task 3.2)
- 33 unit tests covering all resolution paths, cross-context rejection,
  standalone sessions, missing entities, and resolveAllReferences deduplication

### DIFF
--- a/packages/daemon/src/lib/daemon-hub.ts
+++ b/packages/daemon/src/lib/daemon-hub.ts
@@ -336,6 +336,16 @@ export interface DaemonEventMap extends Record<string, BaseEventData> {
 		inboxItemId: string;
 	};
 
+	// App-level MCP registry events
+	/**
+	 * Emitted by app-mcp-handlers on create/update/delete/setEnabled.
+	 * Used by RoomRuntimeService (Task 3.2) for hot-reload of MCP server lists.
+	 * This is separate from LiveQuery — both are emitted on every write.
+	 */
+	'mcp.registry.changed': {
+		sessionId: string;
+	};
+
 	// Goal events
 	'goal.created': {
 		sessionId: string;
--- a/packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts
@@ -1,0 +1,154 @@
+/**
+ * App MCP Registry RPC Handlers
+ *
+ * Exposes the application-level MCP server registry via RPC:
+ * - mcp.registry.list       — list all registry entries
+ * - mcp.registry.get        — get a single entry by id
+ * - mcp.registry.create     — add a new entry, emit mcp.registry.changed
+ * - mcp.registry.update     — update an entry, emit mcp.registry.changed
+ * - mcp.registry.delete     — remove an entry, emit mcp.registry.changed
+ * - mcp.registry.setEnabled — toggle enabled flag, emit mcp.registry.changed
+ *
+ * Note: mcp.registry.listErrors is registered in mcp-handlers.ts (requires
+ * the concrete AppMcpLifecycleManager).
+ */
+
+import type { MessageHub } from '@neokai/shared';
+import type {
+	AppMcpServer,
+	CreateAppMcpServerRequest,
+	UpdateAppMcpServerRequest,
+} from '@neokai/shared';
+import type { DaemonHub } from '../daemon-hub';
+import type { AppMcpServerRepository } from '../../storage/repositories/app-mcp-server-repository';
+import { Logger } from '../logger';
+
+const log = new Logger('app-mcp-handlers');
+
+// ---------------------------------------------------------------------------
+// Handler context
+// ---------------------------------------------------------------------------
+
+export interface AppMcpHandlerContext {
+	db: { appMcpServers: AppMcpServerRepository };
+	daemonHub: DaemonHub;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function emitChanged(daemonHub: DaemonHub): void {
+	daemonHub.emit('mcp.registry.changed', { sessionId: 'global' }).catch((err) => {
+		log.warn('Failed to emit mcp.registry.changed:', err);
+	});
+}
+
+// ---------------------------------------------------------------------------
+// Handler registration
+// ---------------------------------------------------------------------------
+
+export function registerAppMcpHandlers(messageHub: MessageHub, ctx: AppMcpHandlerContext): void {
+	const { db, daemonHub } = ctx;
+
+	// mcp.registry.list — returns AppMcpServer[]
+	messageHub.onRequest('mcp.registry.list', async () => {
+		const servers = db.appMcpServers.list();
+		return { servers } satisfies { servers: AppMcpServer[] };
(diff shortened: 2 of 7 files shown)


lsm (commented):

### Summary

This PR implements Task 3.1 from the System-wide Referencing System plan -- the task and goal resolvers within the ReferenceResolver service. The implementation is clean, well-structured, and meets all acceptance criteria. All 33 unit tests pass, lint and typecheck are clean, and knip reports no unused exports.

### What was done well

- **Acceptance criteria fully met.** All seven acceptance criteria from Task 3.1 are satisfied: room tasks resolve, space tasks resolve, goals resolve with metrics, missing entities return null, standalone sessions return null, cross-context references are rejected, and comprehensive unit tests exist.
- **Defensive cross-context validation.** Cross-room and cross-space checks are implemented even though the underlying repository queries are already scoped (e.g., `getTaskByShortId` filters by roomId). This defense-in-depth is good practice.
- **Exhaustive switch check.** The `default` case in `resolveReference` (line 115-119 of `reference-resolver.ts`) uses the `never` type pattern, which will cause compile-time errors if a new `ReferenceType` is added without updating the switch. This is a strong type-safety guarantee.
- **Graceful null semantics.** The resolver never throws -- all failure modes return null with a log warning. This matches the requirement that reference resolution must never block message sending.
- **Clean test structure.** Tests use real SQLite in-memory databases with real repositories, not mocks. This gives higher confidence that the code integrates correctly with the actual persistence layer.
- **Dependency injection.** All three repository dependencies are optional, allowing the resolver to be instantiated in contexts where only a subset is needed (e.g., file-only sessions).

### P2: Non-`t-` prefixed IDs routed to room task resolution

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, lines 167-168

The `resolveTask` method routes anything that doesn't start with `st-` to `resolveRoomTask`:

```ts
if (id.startsWith('st-')) {
    return this.resolveSpaceTask(id, context);
}
return this.resolveRoomTask(id, context);
```

The plan specifies that room tasks use the `t-` prefix. However, an input like `@ref{task:abc123}` (no prefix at all) would also be routed to room task resolution. While this is harmless because `getTaskByShortId` would simply return null, it would be cleaner to explicitly check for the `t-` prefix for room tasks, treating unrecognized formats as unresolvable. This aligns better with the documented syntax and would make debugging easier.

### P2: `REFERENCE_PATTERN` exported with `g` flag is a footgun

**File:** `packages/shared/src/types/reference.ts`, line 32

```ts
export const REFERENCE_PATTERN = /@ref\{([^}:]+):([^}]+)\}/g;
```

The `g` flag on the exported constant means that any consumer using this pattern directly (without cloning via `new RegExp(REFERENCE_PATTERN.source, 'g')`) will hit `lastIndex` state issues on repeated calls. The comment on lines 29-30 documents this, and the internal `extractReferences` correctly creates a fresh RegExp. However, since this is a public API in `@neokai/shared`, future consumers may forget to clone it. Consider either:
- Exporting without the `g` flag and letting callers add it as needed, or
- Exporting a separate `REFERENCE_PATTERN_SOURCE` string constant alongside the compiled regex.

### P3: Missing test for non-UUID `st-` reference

The test suite covers `st-<real-uuid>` and `st-nonexistent-uuid`, but does not test a reference like `st-not-a-uuid-at-all`. Since the code strips the `st-` prefix and passes the rest directly to `spaceTaskRepo.getTask(uuid)`, it would be useful to confirm this degrades gracefully (it does -- SQLite simply won't find the row). A single test case for this edge case would close the gap.

### P3: Minor inconsistency in null-coalescing style

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`

- Line 210: `progress: task.progress ?? undefined` (room task)
- Line 255: `progress: task.progress !== null && task.progress !== undefined ? task.progress : undefined` (space task)

Both achieve the same result (converting `null` to `undefined`), but use different styles. The `?? undefined` form on line 210 is sufficient since it handles both `null` and `undefined` inputs. The explicit check on line 255 is redundant.

### Recommendation

The implementation is solid and production-ready. The two P2 issues are about API surface robustness for future consumers rather than bugs in the current code. All acceptance criteria are met.

lsm (commented):

**Recommendation: APPROVE**

All P2 issues from the previous review have been properly resolved, and no new issues were introduced.

### P2 Issue Resolution

1. **Explicit `t-` prefix check in resolveTask** — `resolveTask` now explicitly checks for `t-` prefix before routing to room task resolution. Unrecognized formats (neither `t-` nor `st-`) return `null` with a clear warning log: `Unrecognized task reference format: "<id>" (expected t- or st- prefix)`. This is a clean fix at `packages/daemon/src/lib/agent/reference-resolver.ts:167-172`.

2. **REFERENCE_PATTERN exported without `g` flag** — The regex at `packages/shared/src/types/reference.ts:34` is now `/@ref\{([^}:]+):([^}]+)\}/` (no global flag). The JSDoc comment (lines 29-32) clearly documents that callers needing iteration should use `new RegExp(REFERENCE_PATTERN.source, 'g')`. The sole consumer (`reference-resolver.ts:80`) already does exactly this.

### New Tests Added

- **Unrecognized format null return** (`packages/daemon/tests/unit/reference-resolver.test.ts:283-293`): Tests that `@ref{task:abc123}` returns null — covers the new guard clause.
- **Non-UUID st- value null return** (`packages/daemon/tests/unit/reference-resolver.test.ts:467-478`): Tests that `st-not-a-uuid-at-all` returns null gracefully — covers the space task UUID extraction path with invalid input.

### Other Fix

- **Consistent null-coalescing** for space task progress (`reference-resolver.ts:259`): Changed from `task.progress !== null && task.progress !== undefined ? task.progress : undefined` to `task.progress ?? undefined`, matching the room task pattern at line 214.

### Verification

- All 35 tests pass (including 2 new edge case tests).
- Typecheck passes cleanly.
- Lint passes with 0 warnings/errors.
- No other code consumers of `REFERENCE_PATTERN` exist — the only consumer is `reference-resolver.ts:80` which already handles the `g` flag internally.

### Minor Note (P3)

The plan doc `docs/plans/system-wide-referencing-system/01-core-types-and-rpc-infrastructure.md:32` still references the old `REFERENCE_PATTERN = /@ref\{([^}:]+):([^}]+)\}/g` with the `g` flag. This is just a planning document and will naturally go stale, but worth noting for housekeeping.

PR #841, opened 2026-03-25
feat(reference): implement reference.search RPC handler with FileIndex

lsm:
## Summary

- Defines core `@reference` system types in `packages/shared/src/types/reference.ts`: `ReferenceType`, `ReferenceSearchResult`, `ReferenceMention`, `ResolvedReference` variants, `ReferenceMetadata`, and `REFERENCE_PATTERN`
- Implements `FileIndex` class (`packages/daemon/src/lib/file-index.ts`) with polling-based refresh, `.gitignore` parsing, path traversal prevention, and relevance-scored fuzzy search
- Adds `reference.search` RPC handler (`packages/daemon/src/lib/rpc-handlers/reference-handlers.ts`) that searches tasks/goals using session room context and files/folders via FileIndex — standalone sessions silently skip task/goal results
- Registers `setupReferenceHandlers` in `setupRPCHandlers` with FileIndex lifecycle management (init on start, dispose on cleanup)

## Test plan

- [x] 21 unit tests in `packages/daemon/tests/unit/reference-handlers.test.ts` covering:
  - Task search returning matching tasks with shortId and status subtitle
  - Goal search returning matching goals
  - File/folder search via FileIndex
  - Standalone sessions (no roomId) return only file/folder results, no task/goal results
  - Path traversal rejection (`..` and absolute paths in file queries)
  - Type filtering via `types` parameter
  - Relevance sorting (exact > starts-with > contains)
  - Input validation (missing sessionId, non-string query)
- [x] All 7595 unit tests pass, lint clean, typecheck clean

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -74,6 +74,7 @@ import type { GlobalSpacesState } from '../space/tools/global-spaces-tools';
 import { setupSpaceSessionGroupHandlers } from './space-session-group-handlers';
 import { setupLiveQueryHandlers } from './live-query-handlers';
 import { setupReferenceHandlers } from './reference-handlers';
+import { FileIndex } from '../file-index';
 import { LiveQueryEngine } from '../../storage/live-query';
 import type { AppMcpLifecycleManager } from '../mcp';
 
@@ -266,20 +267,28 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	// Dialog handlers (native OS dialogs)
 	setupDialogHandlers(deps.messageHub);
 
+	// Reference handlers (@ mention system — search + resolve tasks, goals, files, folders)
+	const fileIndex = new FileIndex(deps.config.workspaceRoot);
+	fileIndex.init().catch((err) => {
+		log.warn('FileIndex init failed:', err);
+	});
+	setupReferenceHandlers(deps.messageHub, {
+		db: deps.db.getDatabase(),
+		reactiveDb: deps.reactiveDb,
+		shortIdAllocator: deps.db.getShortIdAllocator(),
+		sessionManager: deps.sessionManager,
+		taskRepo: new TaskRepository(deps.db.getDatabase(), deps.reactiveDb),
+		goalRepo: deps.db.getGoalRepo(),
+		workspaceRoot: deps.config.workspaceRoot,
+		fileIndex,
+	});
+
 	// LiveQuery subscribe/unsubscribe handlers
 	const unsubLiveQuery = setupLiveQueryHandlers(
 		deps.messageHub,
 		deps.liveQueries,
 		deps.db.getDatabase()
 	);
-
-	// Reference handlers (@ mention system — resolve tasks, goals, files, folders)
-	setupReferenceHandlers(deps.messageHub, {
-		sessionManager: deps.sessionManager,
-		taskRepo: new TaskRepository(deps.db.getDatabase(), deps.reactiveDb),
-		goalRepo: deps.db.getGoalRepo(),
-		workspaceRoot: deps.config.workspaceRoot,
-	});
 
 	// Space handlers (spaceManager injected from deps — single instance shared with DaemonAppContext)
 	const spaceTaskRepo = new SpaceTaskRepository(deps.db.getDatabase());
@@ -460,6 +469,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 			unsubLiveQuery();
 			roomRuntimeService.stop();
 			spaceRuntimeService.stop();
+			fileIndex.dispose();
 		},
 		spaceRuntimeService,
 		taskAgentManager,
--- a/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
@@ -274,10 +274,7 @@ export function setupReferenceHandlers(messageHub: MessageHub, deps: ReferenceHa
 
 			try {
 				// Fetch extra entries to account for post-search type filtering
-				const fileEntries = fileIndex.search(
-					query,
-					RESULTS_PER_CATEGORY * fileTypes.length * 2
-				);
+				const fileEntries = fileIndex.search(query, RESULTS_PER_CATEGORY * fileTypes.length * 2);
 				// Filter to requested types and enforce per-type limit
 				const byType = new Map<string, number>([
 					['file', 0],
--- a/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
@@ -3,16 +3,24 @@
  *
  * RPC handlers for the @ reference system:
  * - reference.resolve — Resolve a single reference to its full entity data
+ * - reference.search — search tasks, goals, files, and folders by query string
  */
 
 import type {
 	MessageHub,
 	ReferenceType,
+	ReferenceSearchResult,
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Verdict: REQUEST_CHANGES

Overall this is solid foundational work for the `@reference` system. The type definitions are well-structured, the RPC handler is properly integrated into the daemon lifecycle, and the 21 tests provide good coverage of the core flows. However, there are a few issues that should be addressed before merge.

---

### P1 - Should Fix

**1. Symlink loop vulnerability in `FileIndex.scanDir`**
`packages/daemon/src/lib/file-index.ts` (lines 313-323)

`dirent.isDirectory()` returns `true` for symlinks pointing to directories. If the workspace contains a symlink cycle (e.g., `a/b -> ../a`), `scanDir` will recurse infinitely until it hits the call stack limit or exhausts memory. You should check `dirent.isSymbolicLink()` and skip symlinks (or at minimum, track visited real paths via `realpath` to detect cycles).

```ts
// Current (line 313):
const isDir = dirent.isDirectory();

// Should be:
if (dirent.isSymbolicLink()) continue;
const isDir = dirent.isDirectory();
```

**2. `setInterval` timer not unref'd -- prevents process exit**
`packages/daemon/src/lib/file-index.ts` (line 258)

The polling `setInterval` is not `.unref()`'d. Other timers in this codebase (e.g., `providers/anthropic-copilot/provider.ts`, `conversation.ts`, `streaming.ts`) consistently call `.unref()` on timers. Without it, the Node/Bun event loop will never be empty, which can block graceful shutdown in testing and production environments.

```ts
// Current:
this.pollTimer = setInterval(() => { ... }, this.pollMs);

// Should be:
this.pollTimer = setInterval(() => { ... }, this.pollMs);
this.pollTimer.unref();
```

---

### P2 - Important Suggestions

**3. Dead `parseGitignoreLines()` method clobbers gitignore on `setIgnorePatterns`**
`packages/daemon/src/lib/file-index.ts` (lines 147-149)

`parseGitignoreLines()` always returns `[]`. It's called in `setIgnorePatterns()` at line 144:
```ts
this.rebuildIgnoreRegexes([...this.parseGitignoreLines(), ...patterns]);
```
If `setIgnorePatterns()` is called after `init()`, the gitignore patterns loaded during `loadGitignore()` will be lost because `rebuildIgnoreRegexes` starts fresh. Either remove this dead method and load gitignore properly in `setIgnorePatterns`, or make it read from a cached field.

**4. No empty-query guard for task/goal search**
`packages/daemon/src/lib/rpc-handlers/reference-handlers.ts` (line 85-108)

After `query.trim()`, if the query is empty (e.g., user sent `"   "`), the file search branch correctly checks `query.length > 0` at line 142. But the task and goal search branches have no such guard. `filterAndSort` with an empty string matches everything since `"".includes("")` is `true`. This would return up to 10 tasks and 10 goals even when the user sends whitespace-only input. Add `if (!query) return { results: [] };` before the task search block.

**5. Missing `shortIdAllocator` causes gaps in shortId display**
`packages/daemon/src/lib/rpc-handlers/reference-handlers.ts` (lines 99, 120)

Both `TaskRepository` and `GoalRepository` are instantiated with `undefined` for the `shortIdAllocator`. The `deps.db.getShortIdAllocator()` is available in `packages/daemon/src/lib/rpc-handlers/index.ts` and passed to other handler registrations but not to `setupReferenceHandlers`. Without the allocator, `listTasks`/`listGoals` won't lazy-backfill shortIds for tasks/goals that haven't been viewed via `task.list` yet, so the reference search may show tasks without shortIds that would otherwise have them. Consider passing the allocator through `ReferenceHandlerDeps`.

---

### P3 - Nits

**6. Unused `mock` import in test file**
`packages/daemon/tests/unit/reference-handlers.test.ts` (line 14)

`mock` is imported from `bun:test` but never used. While oxlint doesn't flag it (likely exempted for test files), it's dead import noise.

**7. Misleading test assertion comment**
`packages/daemon/tests/unit/reference-handlers.test.ts` (lines 558-581)

The test "still returns task/goal results when file query is traversal but types include task" at line 558 claims to verify that tasks are still searchable when the file path contains `..`. However, the query is `'../../Fix traversal'` and the task title is `'Fix traversal bug'`. The reason 0 results are returned is that the substring `'../../fix traversal'` does not appear in `'fix traversal bug'` -- not because of the traversal guard. To properly test this claim, use a query like `'..traversal'` that would match via substring but be rejected by traversal check (which it currently isn't for task search -- there's no traversal check on the task/goal branch). This test currently tests nothing meaningful about traversal behavior for tasks.

**8. No `REFERENCE_PATTERN` usage or tests**
`packages/shared/src/types/reference.ts` (line 91)

`REFERENCE_PATTERN` is defined but not used anywhere in this PR (no tests, no handler usage). While it's defined as part of the type system for future use, it would be helpful to have at least a basic test confirming it matches expected inputs like `@ref{task:t-42}`.

---

### Positive observations

- The `REFERENCE_PATTERN` regex is well-constructed with `[^}:]+` and `[^}]+` groups
- Path traversal prevention in both `FileIndex.search()` and the RPC handler is defense-in-depth
- Room context awareness is correctly handled -- standalone sessions silently skip task/goal results
- The scoring/relevance system with exact > starts-with > contains is sensible
- The test setup with real SQLite in-memory DB (not mocking repositories) gives high confidence
- Clean integration into the existing daemon lifecycle (`init` on start, `dispose` on cleanup)
- Proper use of `Logger` instead of `console` calls
- Lint, typecheck, and all 21 tests pass cleanly

lsm (commented):

### Verdict: APPROVE

All previously identified issues have been properly addressed in the fix commit (8b90448).

### P1 fixes — verified

- **Symlink loop vulnerability**: `scanDir` now calls `dirent.isSymbolicLink()` at line 308, placed before the `isDirectory()` check. This prevents both infinite loops from cyclic symlinks and following symlinks that point outside the workspace. The structural type in `readdir` result at line 294 also correctly includes `isSymbolicLink()`.
- **Timer .unref()**: `setInterval` result at line 267 immediately calls `.unref()`, allowing graceful process shutdown without waiting for the polling timer.

### P2 fixes — verified

- **gitignoreLines caching**: `loadGitignore()` now stores parsed lines in `this.gitignoreLines` (line 167/169). `setIgnorePatterns()` at line 146 merges the cached gitignore lines with extra patterns. The dead `parseGitignoreLines()` method is gone.
- **Empty-query guard**: Line 93 in `reference-handlers.ts` returns `{ results: [] }` after `trim()`, preventing whitespace-only input from matching everything. A test at line 760 verifies this.
- **shortIdAllocator**: Added to `ReferenceHandlerDeps` (line 60), passed to `TaskRepository` (line 108) and `GoalRepository` (line 129) constructors. Wired in `index.ts` at line 278 via `deps.db.getShortIdAllocator()`.

### P3 fixes — verified

- Unused `mock` import removed.
- Misleading traversal/task test replaced with a proper test (line 578) that verifies task search works while file traversal guard only applies to the file/folder branch.
- 8 new `REFERENCE_PATTERN` tests added covering task, goal, file, folder matches, plain @mentions, markdown links, and matchAll.

### Remaining observations (non-blocking, P3)

- In the `matchAll` test (line 832), the regex is duplicated as an inline literal instead of derived from the imported `REFERENCE_PATTERN` constant. Since `REFERENCE_PATTERN` has the `g` flag and `matchAll` requires a fresh regex each call, using `new RegExp(REFERENCE_PATTERN.source, REFERENCE_PATTERN.flags)` would be more DRY and resistant to future pattern changes. This is purely cosmetic since the test is correct as-is.

### Test quality

29 tests pass, all green. Tests cover the critical paths: task/goal search with room context, standalone sessions (no roomId), path traversal prevention, type filtering, relevance sorting, input validation, and REFERENCE_PATTERN parsing. Lint and typecheck are clean.

**Recommendation: APPROVE**

PR #842, opened 2026-03-25
feat(references): implement file/folder resolvers and shared reference types

lsm:
## Summary

- Adds `packages/shared/src/types/reference.ts` with all core reference system types (`ReferenceType`, `ReferenceMention`, `ResolvedFileReference`, `ResolvedFolderReference`, `REFERENCE_PATTERN`, and more)
- Creates `ReferenceResolver` service (`packages/daemon/src/lib/agent/reference-resolver.ts`) with `extractReferences()`, `resolveReference()`, and `resolveAllReferences()` methods
- Implements file resolver: reads via `FileManager`, auto-detects binary files (returns base64), truncates large text files at 50KB line boundaries
- Implements folder resolver: lists directory via `FileManager`, limits to 200 entries, reports `totalCount` for larger directories
- Path validation layer: rejects `..` segments, absolute paths, and symlinks pointing outside the workspace (uses `realpath` on both target and workspace to handle macOS `/tmp` → `/private/tmp` symlinks)
- Task/goal resolvers are stubbed (return null), ready for Task 3.1 to fill in

## Test plan

- [x] 43 unit tests in `packages/daemon/tests/unit/reference-resolver.test.ts`
- [x] `extractReferences`: parsing, deduplication, invalid formats, unknown types
- [x] File resolver: happy path, missing file, binary detection, large file truncation at line boundary, empty file, unicode
- [x] Folder resolver: happy path, missing folder, 200-entry limit with correct `totalCount`, empty directory
- [x] Path traversal: `..` segments, absolute paths, symlinks outside workspace, symlinks within workspace (allowed)
- [x] Standalone sessions (no room/space context) work for file/folder
- [x] Task/goal references return null (stub tests)
- [x] `resolveAllReferences`: partial failure handling, empty text

### DIFF
--- a/packages/daemon/src/lib/agent/reference-resolver.ts
+++ b/packages/daemon/src/lib/agent/reference-resolver.ts
@@ -1,0 +1,393 @@
+/**
+ * ReferenceResolver — Core service for parsing and resolving @ references
+ *
+ * Handles extracting @ref{} mentions from text and resolving them to their
+ * full entity data. Currently implements file and folder resolvers.
+ * Task and goal resolvers are stubbed (see Task 3.1).
+ *
+ * Path validation enforced for file/folder references:
+ *   - Empty paths are rejected
+ *   - Paths containing `..` are rejected
+ *   - Absolute paths (starting with `/`) are rejected
+ *   - Symlinks that resolve outside the workspace are rejected
+ *
+ * Security note — TOCTOU:
+ *   There is an inherent time-of-check/time-of-use gap between `checkSymlink()`
+ *   (which calls `realpath`) and the subsequent `FileManager.readFile()` /
+ *   `listDirectory()`. An attacker with write access to the workspace could swap a
+ *   symlink in that window. In practice this is low-severity for a chat application
+ *   where the user owns the workspace. The defense-in-depth is:
+ *   1. `checkSymlink` rejects symlinks pointing outside the workspace at check time.
+ *   2. `FileManager.validatePath()` re-checks path containment via
+ *      `normalize(join(workspacePath, path))` at use time, which handles non-symlink
+ *      traversal but does NOT re-resolve symlinks.
+ *   3. The race window is extremely narrow (microseconds).
+ *
+ * Usage:
+ *   const resolver = new ReferenceResolver();
+ *   const mentions = resolver.extractReferences('@ref{file:src/utils.ts}');
+ *   const resolved = await resolver.resolveAllReferences(text, { workspacePath });
+ */
+
+import { join, normalize, relative, isAbsolute } from 'node:path';
+import { realpath } from 'node:fs/promises';
+import { FileManager } from '../file-manager.ts';
+import type {
+	ReferenceMention,
+	ResolvedReference,
+	ResolvedFileReference,
+	ResolvedFolderReference,
+	ReferenceType,
+} from '@neokai/shared';
+import { REFERENCE_PATTERN } from '@neokai/shared';
+import { Logger } from '../logger.ts';
+
+const log = new Logger('reference-resolver');
+
+/** Max file content size (50KB) before truncation at line boundary */
+const MAX_FILE_SIZE = 50 * 1024;
+
+/** Max number of directory entries returned for folder references */
+const MAX_FOLDER_ENTRIES = 200;
+
+/**
+ * Context for resolving references — provides session-scoped information.
+ * File/folder resolution only requires workspacePath.
+ * Task/goal resolution (Task 3.1) will use roomId and spaceId.
+ */
+export interface ResolutionContext {
+	roomId?: string;
+	spaceId?: string;
+	workspacePath: string;
+}
+
+export class ReferenceResolver {
+	/**
+	 * Extract all @ref{} mentions from a text string.
+	 * Returns deduplicated mentions in order of appearance.
+	 */
+	extractReferences(text: string): ReferenceMention[] {
+		const mentions: ReferenceMention[] = [];
+		const seen = new Set<string>();
+		// REFERENCE_PATTERN has the g flag but we construct a fresh instance to
+		// avoid shared lastIndex state if REFERENCE_PATTERN is reused elsewhere.
+		const pattern = new RegExp(REFERENCE_PATTERN.source, 'g');
+		let match: RegExpExecArray | null;
+
+		while ((match = pattern.exec(text)) !== null) {
+			const [fullMatch, typeRaw, id] = match;
+
+			if (!this.isValidReferenceType(typeRaw)) {
+				continue;
+			}
+
+			if (seen.has(fullMatch)) continue;
+			seen.add(fullMatch);
+
+			mentions.push({
+				type: typeRaw as ReferenceType,
+				id,
+				displayText: fullMatch,
+			});
+		}
+
+		return mentions;
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


lsm (commented):

---

### Overall Assessment

This is a well-structured implementation that correctly delivers Task 3.2 from the design doc. The code is clean, follows project conventions (passes typecheck, lint, knip, and all 43 tests), and the security layer is thoughtfully designed with defense-in-depth. I have a few items to flag but nothing blocking.

---

### P1 (should-fix)

**1. TOCTOU race condition between symlink check and file read**

`packages/daemon/src/lib/agent/reference-resolver.ts`, lines 217-230

Between `checkSymlink()` (line 219) which calls `realpath()` and the subsequent `fm.readFile()` (line 230), an attacker with write access to the workspace could swap a legitimate symlink to point outside the workspace. This is a classic TOCTOU (time-of-check-time-of-use) race.

In practice, this is **low severity** because:
- The attacker would need write access to the workspace (they already have file access)
- `FileManager.readFile()` does its own `validatePath()` which re-checks path containment after `normalize(join(...))`
- The race window is extremely narrow

However, the `checkSymlink()` check on a directory followed by `fm.listDirectory()` has a slightly wider window because `listDirectory()` calls `stat()` per entry, which follows symlinks.

**Recommendation:** Consider adding a comment at the `checkSymlink()` call site acknowledging the TOCTOU limitation, or document that FileManager's internal `validatePath()` provides the secondary defense. The current defense-in-depth is reasonable for a chat application where the user references their own workspace.

**2. `FileManager` cache never invalidated**

`packages/daemon/src/lib/agent/reference-resolver.ts`, lines 52-61

The `fileManagerCache` is a simple `Map` keyed by `workspacePath` that grows unbounded. If the daemon runs for a long time and resolves references across many worktree paths, these `FileManager` instances are never released.

This is P1 rather than P0 because:
- In practice, most sessions share one workspace, so the cache stays small
- `FileManager` is lightweight (no file handles held open)

**Recommendation:** Either add a `clearCache()` method or use a `WeakRef` pattern. Alternatively, since `FileManager` instances are cheap, consider not caching at all and creating one per `resolveAllReferences()` call.

---

### P2 (important suggestion)

**3. `checkSymlink()` allows access to the workspace root itself**

`packages/daemon/src/lib/agent/reference-resolver.ts`, lines 175-198

If a user provides an empty string as the path ID (e.g., `@ref{file:}`), the regex would capture an empty string as `id`. Then:
- `validatePath("")` passes (empty string is not absolute, no `..` segments)
- `normalize(join(workspacePath, ""))` = `normalize(workspacePath)` = workspacePath
- `checkSymlink(workspacePath, workspacePath)` would compute `relative(workspacePath, workspacePath) = """, which passes the check
- `fm.readFile("")` would attempt to read the workspace directory as a file, which would throw "Path is a directory" and return null

So this is not a security issue, but it would be cleaner to reject empty paths explicitly in `validatePath()`. The design doc says "Parse path from ID" implying the path should be non-empty.

**4. Tests use `as { content: string }` type casts instead of narrowing**

`packages/daemon/tests/unit/reference-resolver.test.ts`, lines 136-140, 196-199, 214-215, etc.

The tests repeatedly cast resolved references with `as { content: string }` / `as { encoding: string }` instead of narrowing through TypeScript's discriminated union on `result.type`. For example:

```ts
// Current (loses type safety)
expect((result as { content: string }).content).toBe('Hello World');

// Better (uses the discriminated union)
if (result?.type !== 'file') return expect.unreachable();
expect(result.content).toBe('Hello World');
```

This is a readability issue -- the current casts make the tests harder to maintain and would miss type errors if the `ResolvedFileReference` interface changes.

---

### P3 (nit)

**5. Comment says "Permission denied" but the catch block also swallows all other errors**

`packages/daemon/src/lib/agent/reference-resolver.ts`, line 241

The comment `// Permission denied or error reading file` and log message `Permission denied or error reading file` are slightly misleading. The catch-all at line 241 catches *any* error that doesn't match the three checked strings (File not found, Path is a directory, Path traversal). This could mask unexpected errors like disk I/O failures, ENOMEM, etc. Consider logging the full error object rather than just a generic message.

**6. `REFERENCE_PATTERN` exported with `g` flag on the constant**

`packages/shared/src/types/reference.ts`, line 120

The regex `/@ref\{([^}:]+):([^}]+)\}/g` is exported with the `g` flag. The JSDoc correctly warns about the `lastIndex` pitfall, and `extractReferences()` correctly creates a fresh instance. However, any future consumer who uses `REFERENCE_PATTERN` directly in a loop (without the JSDoc warning) will hit the shared `lastIndex` bug. Consider exporting without the `g` flag and having callers add it: `export const REFERENCE_PATTERN = /@ref\{([^}:]+):([^}]+)\}/;`

---

### What's Done Well

- **Defense-in-depth path validation**: Three layers of checks (pre-validation, realpath symlink check, FileManager's own validatePath) provide robust path traversal prevention.
- **Proper handling of macOS `/tmp` -> `/private/tmp` symlinks** via `realpath` on both the target path and workspace path.
- **Binary detection via null-byte heuristic** is practical and avoids false positives on UTF-8 text.
- **Line-boundary truncation** counts bytes (not characters) for multi-byte UTF-8 correctness.
- **Exhaustive switch with `never` type** for future-proof reference type handling (line 112-116).
- **Clean test organization** with real temp directories and proper cleanup via `afterEach`.

lsm (commented):

All previous P1/P2 issues have been adequately addressed, and no new blocking issues were introduced. One minor documentation inaccuracy remains.

### Previous issues — status

- **P1 TOCTOU**: Resolved. Module-level comment (lines 14-24) now documents the race window and defense-in-depth strategy clearly and accurately.
- **P1 Unbounded cache**: Resolved. `fileManagerCache` removed entirely; `new FileManager(context.workspacePath)` is created per call, which is appropriate given the comment that it is cheap to instantiate.
- **P2 Empty path**: Resolved. `validatePath()` at line 156-159 rejects both `''` and whitespace-only paths. Two new tests (lines 390-403) verify this for both file and folder types.
- **P2 Type casts**: Resolved. Replaced with `assertFileResult` / `assertFolderResult` assertion functions (lines 34-48) that use TypeScript discriminated union narrowing via `asserts` return types. Much cleaner and type-safe.
- **P3 Error log**: Resolved. Changed to "Unexpected error" in all three catch blocks (lines 250, 263, 377).
- **Shared tests**: Resolved. `reference.test.ts` updated to match enriched `ResolvedFileReference` and `ResolvedFolderReference` interfaces.

### New findings

**P3 — Misleading comment in `extractReferences`**
At `packages/daemon/src/lib/agent/reference-resolver.ts` line 72-73, the comment states "REFERENCE_PATTERN has no `g` flag" but the actual pattern at `packages/shared/src/types/reference.ts` line 135 explicitly includes the `g` flag: `/@ref\{([^}:]+):([^}]+)\}/g`. The code itself is correct (it uses `REFERENCE_PATTERN.source` to extract just the pattern body and constructs a new instance with `g`), but the comment is misleading. Suggested fix: change "has no `g` flag" to something like "uses `.source` to avoid inheriting the `g` flag's state".

### Verification

- All 45 unit tests pass (70 expect() calls)
- All 19 shared reference tests pass (30 expect() calls)
- TypeScript typecheck passes cleanly
- Oxlint reports 0 warnings/errors
- Knip reports no new unused exports

PR #843, opened 2026-03-25
feat(web): add ReferenceAutocomplete component for @ reference system

lsm:
Creates the dropdown component that displays reference search results
grouped by type (tasks, goals, files, folders), mirroring the
CommandAutocomplete styling. Features include:
- Grouped sections with section labels (empty groups hidden)
- Type icons with distinct colors per entity type
- Selected item highlighting with blue border
- Auto-scroll selected item into view
- Click-outside dismiss
- Dynamic header: "References" vs "Files & Folders" based on result types
- 26 unit tests covering rendering, selection, grouping, and positioning

### DIFF
--- a/packages/web/src/components/ReferenceAutocomplete.tsx
+++ b/packages/web/src/components/ReferenceAutocomplete.tsx
@@ -1,0 +1,233 @@
+import { useEffect, useRef } from 'preact/hooks';
+import { cn } from '../lib/utils.ts';
+import { borderColors } from '../lib/design-tokens.ts';
+import type { ReferenceSearchResult, ReferenceType } from '@neokai/shared';
+
+export interface ReferenceAutocompleteProps {
+	results: ReferenceSearchResult[];
+	selectedIndex: number;
+	onSelect: (result: ReferenceSearchResult) => void;
+	onClose: () => void;
+	position?: { top: number; left: number };
+}
+
+const TYPE_ORDER: ReferenceType[] = ['task', 'goal', 'file', 'folder'];
+
+const TYPE_LABELS: Record<ReferenceType, string> = {
+	task: 'Tasks',
+	goal: 'Goals',
+	file: 'Files',
+	folder: 'Folders',
+};
+
+function TypeIcon({ type }: { type: ReferenceType }) {
+	if (type === 'task') {
+		return (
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Verdict: APPROVE

This is a clean, well-structured implementation that faithfully mirrors the existing CommandAutocomplete component while extending it with grouped rendering. All project checks pass (lint, typecheck, format, tests).

### What looks good

- **Props interface** matches the task requirements exactly, including the optional position prop.
- **Grouped rendering** with TYPE_ORDER array and indexMap correctly preserves the flat selectedIndex semantics across group boundaries. The global index tracking via Map is a sound approach.
- **Dynamic header** logic in resolveHeaderLabel correctly distinguishes References vs Files & Folders based on result types.
- **Empty group hiding** works correctly because the filter + items.length > 0 check in the grouping loop naturally skips empty groups.
- **Click-outside dismiss** uses mousedown (matching CommandAutocomplete), which fires before click to avoid race conditions with button click handlers.
- **Auto-scroll** via scrollIntoView({ block: nearest, behavior: smooth }) matches the existing pattern.
- **Key format** uses a reasonable compound key pattern.
- **Styling** is consistent with CommandAutocomplete: same border colors, animation class, footer structure. The only intentional differences are max-h-72 vs max-h-64 and minWidth/maxWidth 280px/420px vs 250px/400px, which make sense given the richer content.
- **26 tests** provide solid coverage of rendering, grouping, selection, click handling, positioning, and header logic.
- **Code style** passes Biome format check, Oxlint (0 warnings/errors), and TypeScript build check.

1. P3 - Header SVG icon: The header uses a generic hash/menu icon. A dedicated @ or crosshairs icon might better communicate the reference concept to users, but this is purely aesthetic and can be iterated on later.

2. P3 - TypeIcon as if-chain: The TypeIcon component uses an if-chain with four branches. This is clear and fine at this size, but if ReferenceType grows, a lookup object pattern (similar to TYPE_LABELS) would be more scalable. Not actionable now.

3. P3 - Duplicate results key collision: The key format would collide if the backend ever returns two results with the same type and id. In practice this would indicate a backend bug, so this is acceptable.

No P0, P1, or P2 issues found.

PR #844, opened 2026-03-25
feat(session): implement message preprocessing with @ reference resolution

lsm:
Adds ReferenceResolver service and wires it into MessagePersistence so
that @ref{type:id} tokens in user messages are resolved and attached as
referenceMetadata on the persisted SDK message blob.

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
@@ -215,7 +215,10 @@ function resolveGoal(
 	};
 }
 
-async function resolveFile(id: string, workspacePath: string): Promise<ResolvedReference | null> {
+export async function resolveFile(
+	id: string,
+	workspacePath: string
+): Promise<ResolvedReference | null> {
 	const fileManager = new FileManager(workspacePath);
 
 	// Validate path (throws on traversal — we catch below to return null)
@@ -311,7 +314,10 @@ async function resolveFile(id: string, workspacePath: string): Promise<ResolvedR
 	};
 }
 
-async function resolveFolder(id: string, workspacePath: string): Promise<ResolvedReference | null> {
+export async function resolveFolder(
+	id: string,
+	workspacePath: string
+): Promise<ResolvedReference | null> {
 	const fileManager = new FileManager(workspacePath);
 
 	let entries: Array<{ name: string; path: string; type: 'file' | 'directory' }>;
--- a/packages/daemon/src/lib/session/message-persistence.ts
+++ b/packages/daemon/src/lib/session/message-persistence.ts
@@ -10,7 +10,14 @@
  * - Emit events for downstream processing
  */
 
-import type { MessageContent, MessageDeliveryMode, MessageHub, MessageImage } from '@neokai/shared';
+import type {
+	MessageContent,
+	MessageDeliveryMode,
+	MessageHub,
+	MessageImage,
+	ReferenceMetadata,
+	Session,
+} from '@neokai/shared';
 import type { SDKUserMessage } from '@neokai/shared/sdk';
 import type { UUID } from 'crypto';
 import type { Database } from '../../storage/database';
@@ -18,6 +25,11 @@ import type { DaemonHub } from '../daemon-hub';
 import { expandBuiltInCommand } from '../built-in-commands';
 import { Logger } from '../logger';
 import type { SessionCache } from './session-cache';
+import {
+	ReferenceResolver,
+	type PreprocessedMessage,
+	type ResolutionContext,
+} from './reference-resolver';
 
 export interface MessagePersistenceData {
 	sessionId: string;
@@ -34,9 +46,49 @@ export class MessagePersistence {
 		private sessionCache: SessionCache,
 		private db: Database,
 		private messageHub: MessageHub,
-		private eventBus: DaemonHub
+		private eventBus: DaemonHub,
+		private referenceResolver?: ReferenceResolver
 	) {
 		this.logger = new Logger('MessagePersistence');
+	}
+
+	/**
+	 * Extract and resolve @ references from a message text.
+	 *
+	 * Returns a PreprocessedMessage with the original text unchanged and a
+	 * populated referenceMetadata map. If no references are found, or if
+	 * resolution fails entirely, returns empty metadata so the message
+	 * still persists normally.
+	 */
+	private async preprocessReferences(text: string, session: Session): Promise<PreprocessedMessage> {
+		try {
+			const mentions = ReferenceResolver.extractReferences(text);
+			if (mentions.length === 0) {
+				return { text, referenceMetadata: {} };
+			}
+
+			const context: ResolutionContext = {
+				workspacePath: session.workspacePath,
+				roomId: session.context?.roomId ?? null,
+			};
+
+			const resolved = await this.referenceResolver!.resolveAllReferences(mentions, context);
+
+			// Convert ResolvedReference values to ReferenceMetadata entries
+			const referenceMetadata: ReferenceMetadata = {};
+			for (const [token, ref] of Object.entries(resolved)) {
+				referenceMetadata[token] = {
+					type: ref.type,
+					id: ref.id,
+					displayText: ref.id,
+				};
+			}
+
+			return { text, referenceMetadata };
+		} catch (err) {
+			this.logger.warn('[MessagePersistence] Reference preprocessing failed, skipping:', err);
+			return { text, referenceMetadata: {} };
+		}
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


lsm (commented):

---

### Summary

This PR introduces the ReferenceResolver class and integrates @ reference resolution into MessagePersistence.persist(). The implementation is clean, well-documented, and all 21 new tests pass. The code correctly meets most acceptance criteria. I have a few findings that should be addressed before merge.

---

### P1 Issues

**1. ReferenceResolver is never wired into production code**

MessagePersistence accepts an optional referenceResolver parameter, but the sole production instantiation at packages/daemon/src/lib/session/session-manager.ts:110 does not pass one:

```ts
this.messagePersistence = new MessagePersistence(this.sessionCache, db, messageHub, eventBus);
```

This means the entire feature is dead code in production -- no user messages will have their @ references resolved. The feature only works in tests where a resolver is explicitly provided.

To wire it in, SessionManager would need access to a TaskRepository (for task resolution) and the Database (for goal resolution via db.getGoal / db.getGoalByShortId). Currently TaskRepository is not exposed through the Database facade, so either the facade needs a getTaskRepo() accessor (like it already has getGoalRepo()) or the SessionManager constructor needs to accept repository dependencies.

**Recommendation:** Wire the ReferenceResolver into the MessagePersistence construction in SessionManager, or at minimum add a TODO/COMMENT explaining that wiring is deferred to a follow-up PR. As-is, this is functionally inert.

---

**2. No deduplication of duplicate references**

If a user writes @ref{task:t-1} and @ref{task:t-1}, the same reference is extracted twice and resolved twice (including the filesystem I/O for file/folder types). While the Record<string, ResolvedReference> output naturally deduplicates by overwriting, the resolution work is still performed redundantly.

This is a minor efficiency concern now, but could become meaningful with many file references (each involving filesystem stat + read).

---

### P2 Issues

**3. displayText is always set to ref.id -- not very descriptive**

In packages/daemon/src/lib/session/message-persistence.ts:83:

```ts
referenceMetadata[token] = {
    type: ref.type,
    id: ref.id,
    displayText: ref.id,
};
```

For tasks and goals, ref.data contains the resolved entity (e.g., a NeoTask with a title field). Setting displayText to the raw ID (e.g., 't-1') loses potentially useful display information. The frontend may eventually want to render something like 'Task: Fix login bug' rather than 't-1'.

The same applies in ReferenceResolver.extractReferences() at line 68 where displayText: id is hardcoded for the ReferenceMention.

---

**4. Unresolved references silently disappear -- no status tracking**

The ReferenceMetadata type has an optional status field, but it is never used. When a reference fails to resolve (e.g., task deleted, file not found), it is simply excluded from the metadata entirely. This means the frontend has no way to distinguish between 'no references were in the message' and 'references were present but failed to resolve.'

Consider including unresolved references with status: 'unresolved' so the UI can surface resolution failures to the user (e.g., showing a stale @ mention indicator).

---

### P3 Nits

**5. The type guard in extractReferences is redundant with the regex pattern**

Line 64 filters types with a manual check:
```ts
if (type !== 'task' && type !== 'goal' && type !== 'file' && type !== 'folder') {
    continue;
}
```

The regex /@ref{([^}:]+):([^}]+)}/g will match any type string (including future types like @ref{unknown:x}). This guard is good defensive programming, but could be more maintainable by referencing the ReferenceType union directly (e.g., a Set<ReferenceType> or similar).

**6. Test coverage for file/folder resolution in resolveAllReferences is missing**

Tests cover extraction of file/folder references but not their resolution through resolveAllReferences. The file/folder resolution paths use resolveFile/resolveFolder which involve filesystem I/O. These would benefit from a mock-based test to ensure the full flow works end-to-end for all reference types.

---

### Verdict

The implementation design is solid and the code quality is good. However, the feature is not actually wired into production (P1 #1), which is the primary concern. If wiring is intentionally deferred, that should be documented explicitly. Otherwise, this should be connected before merge.

lsm (commented):

### Verdict: REQUEST_CHANGES

---

### P0 (Blocking): Existing session-manager.test.ts is broken (44 test failures)

**File:** `packages/daemon/tests/unit/session/session-manager.test.ts`

The `SessionManager` constructor at `session-manager.ts:111-113` now calls `db.getTaskRepo()` and `db.getGoalRepo()`, but the mock `Database` object in `session-manager.test.ts:35-67` does not define these methods. This causes **all 44 tests** in the session-manager test file to fail with:

```
TypeError: db.getTaskRepo is not a function
```

Verified: tests pass on `dev` (44/44), fail on this branch (0/44).

**Fix:** Add `getTaskRepo` and `getGoalRepo` to the mock DB in `session-manager.test.ts`. These can return mock objects that satisfy the `TaskRepoForReference` and `GoalRepoForReference` interfaces (with no-op `getTask`, `getTaskByShortId`, `getGoal`, `getGoalByShortId` methods returning `null`).

---

### P1 (Should-fix): Previous P1 issues -- status update

**P1-1 (ReferenceResolver dead code):** RESOLVED. The resolver is now wired into production via `SessionManager` constructor (line 111-120 of `session-manager.ts`). It is constructed with real repo instances from the database facade and passed to `MessagePersistence`. Well done.

**P1-2 (displayText always uses raw ID):** RESOLVED. The `extractDisplayText` function at `message-persistence.ts:253-264` now extracts entity titles for task/goal types and file paths for file/folder types, falling back to raw ID. Tests at lines 343-393 and 476-533 confirm this works correctly.

---

### P2 (Important): Previous P2 issues -- status update

**P2-1 (Unresolved references silently disappear):** RESOLVED. Unresolved references are now included in `referenceMetadata` with `status: 'unresolved'` (lines 88-101 of `message-persistence.ts`). Tests at lines 395-435 and 476-533 confirm partial and fully-unresolved scenarios.

**P2-2 (Additional issue from first review):** Covered by the above fixes.

---

### P3 (Nits)

1. **Redundant `seenTokens` set** (`message-persistence.ts:89`): The `seenTokens.has(token)` check is redundant with `!(token in referenceMetadata)` -- any token already processed (whether resolved or from a prior mention duplicate) is already in `referenceMetadata`. The `resolveAllReferences` method already deduplicates the input mentions, so the raw `mentions` array is the only source of potential duplicates. The code is correct but the `seenTokens` set could be removed for simplicity, relying solely on the `!(token in referenceMetadata)` guard.

2. **Deduplication test assertion is loose** (`reference-message-persistence.test.ts:236`): The test asserts `taskCallCount <= 2` (at most 1 getTask + 1 getByShortId), but with proper deduplication, the expected count should be exactly 2 (since the resolver tries `getTask` first, then falls back to `getTaskByShortId`). The `<= 2` assertion would pass even without deduplication if the mock returned null on both paths. A more precise assertion would be `expect(taskCallCount).toBe(2)`.

---

### Positive observations

- Clean separation of concerns: `ReferenceResolver` is well-isolated with a clear deps interface, making it easy to test and reason about.
- Cross-room access prevention in `resolveTask`/\ `resolveGoal` (lines 163-166, 185-188 of `reference-resolver.ts`) is a good security practice.
- The graceful degradation in `preprocessReferences` (try/catch returns empty metadata) ensures reference resolution failures never block message persistence.
- Path traversal protection in `resolveFile`/`resolveFolder` is already handled in the shared handlers.
- The `ReferenceResolverDeps` interface reusing `TaskRepoForReference`/`GoalRepoForReference` from `reference-handlers.ts` is good for consistency.

lsm (commented):

### Verdict: APPROVE

The P0 regression from the previous round is properly fixed.

### Fix Verification (commit 6246e2413)

The mock Database in packages/daemon/tests/unit/session/session-manager.test.ts (lines 67-74) now includes the required getTaskRepo() and getGoalRepo() stubs. The fix is minimal and correct:

1. **Interface compliance**: The stubs return objects matching the TaskRepoForReference and GoalRepoForReference interfaces from packages/daemon/src/lib/rpc-handlers/reference-handlers.ts (lines 38-46). Each provides getTask/getTaskByShortId and getGoal/getGoalByShortId respectively, which are the exact methods used by ReferenceResolver in packages/daemon/src/lib/session/reference-resolver.ts (lines 154-156, 176-178).

2. **All 44 session-manager tests pass** -- verified by running the test suite.

3. **No new regressions**: Full daemon unit test suite (7675 tests, 0 failures) passes cleanly. The new reference-message-persistence.test.ts (22 tests) also passes.

4. **TypeScript typecheck passes** with no errors.

5. **Mock structure is reasonable**: Returning null for all repo methods is appropriate here because the session-manager tests do not exercise reference resolution -- they just need the methods to exist so the SessionManager constructor (which calls db.getTaskRepo() and db.getGoalRepo() at lines 112-113 of session-manager.ts) does not throw. The dedicated reference-message-persistence.test.ts covers actual reference resolution behavior with its own properly mocked repos.

### Overall PR Assessment

The three commits in this PR are solid:
- 75e2061fc -- initial feature implementation
- c4268bd33 -- wiring fixes for production code
- 6246e2413 -- test fix for the regression

The ReferenceResolver is well-structured with proper error handling (catch blocks returning null, unresolved tracking for UI feedback), deduplication of references, and security checks (room-scoped lookups preventing cross-room access via UUID). The MessagePersistence integration is clean -- the resolver is optional, and preprocessing failures are gracefully caught so message persistence is never blocked.

PR #845, opened 2026-03-25
feat(web): integrate reference autocomplete with MessageInput

lsm:
- Create ReferenceAutocomplete component with type icons (task/goal/file/folder),
  display text, subtitle, shortId, and keyboard navigation hints
- Extend InputTextarea with showReferenceAutocomplete, referenceResults,
  selectedReferenceIndex, onReferenceSelect, onReferenceClose props and
  optional textareaRef forwarding for post-selection focus
- Wire useReferenceAutocomplete into MessageInput: handleReferenceSelect
  replaces @query with @ref{type:id} token and refocuses textarea
- Reference autocomplete takes keyboard event precedence over command autocomplete
- Supports combining / commands and @ references (e.g. /agent @task-t-42 fix)
- Add 20 new tests: ReferenceAutocomplete component and MessageInput integration
- Fix MessageInput.queue-mode test mock to include useReferenceAutocomplete

### DIFF
--- a/packages/web/src/components/InputTextarea.tsx
+++ b/packages/web/src/components/InputTextarea.tsx
@@ -25,6 +25,7 @@
  */
 
 import type { ComponentChildren } from 'preact';
+import type { MutableRef } from 'preact/hooks';
 import { useEffect, useLayoutEffect, useRef, useState } from 'preact/hooks';
 import { cn } from '../lib/utils.ts';
 import { borderColors } from '../lib/design-tokens.ts';
@@ -61,6 +62,8 @@ export interface InputTextareaProps {
 	leadingElement?: ComponentChildren;
 	/** Left padding class used when leadingElement is present */
 	leadingPaddingClass?: string;
+	/** Optional ref forwarded to the underlying textarea element */
+	textareaRef?: MutableRef<HTMLTextAreaElement | null>;
 }
 
 /**
@@ -89,8 +92,10 @@ export function InputTextarea({
 	onPaste,
 	leadingElement,
 	leadingPaddingClass,
+	textareaRef: externalTextareaRef,
 }: InputTextareaProps) {
-	const textareaRef = useRef<HTMLTextAreaElement>(null);
+	const internalTextareaRef = useRef<HTMLTextAreaElement>(null);
+	const textareaRef = externalTextareaRef ?? internalTextareaRef;
 	const [isMultiline, setIsMultiline] = useState(false);
 
 	// Detect if device is mobile (touch-based)
--- a/packages/web/src/components/MessageInput.tsx
+++ b/packages/web/src/components/MessageInput.tsx
@@ -235,16 +235,22 @@ export default function MessageInput({
 		]
 	);
 
+	// Destructure stable callback refs to avoid recreating handleKeyDown on every render
+	// (hooks return new object instances each render, but the functions inside are stable
+	// via useCallback, so depending on the functions directly is more efficient)
+	const refHandleKeyDown = referenceAutocomplete.handleKeyDown;
+	const cmdHandleKeyDown = commandAutocomplete.handleKeyDown;
+
 	// Keyboard handler
 	const handleKeyDown = useCallback(
 		(e: KeyboardEvent) => {
 			// Reference autocomplete takes precedence when visible
-			if (referenceAutocomplete.handleKeyDown(e)) {
+			if (refHandleKeyDown(e)) {
 				return;
 			}
 
 			// Then try command autocomplete
-			if (commandAutocomplete.handleKeyDown(e)) {
+			if (cmdHandleKeyDown(e)) {
 				return;
 			}
 
@@ -273,7 +279,7 @@ export default function MessageInput({
 				}
 			}
 		},
-		[referenceAutocomplete, commandAutocomplete, handleSubmit, agentWorking]
+		[refHandleKeyDown, cmdHandleKeyDown, handleSubmit, agentWorking]
 	);
 
 	// Model switch handler
@@ -462,7 +468,9 @@ export default function MessageInput({
 							}}
 							disabled={disabled}
 							placeholder={getPlaceholderForSessionType(sessionType)}
-							showCommandAutocomplete={commandAutocomplete.showAutocomplete}
+							showCommandAutocomplete={
+								commandAutocomplete.showAutocomplete && !referenceAutocomplete.showAutocomplete
+							}
 							filteredCommands={commandAutocomplete.filteredCommands}
 							selectedCommandIndex={commandAutocomplete.selectedIndex}
 							onCommandSelect={commandAutocomplete.handleSelect}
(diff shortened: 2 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly integrates the reference autocomplete system into the MessageInput component. The implementation follows existing patterns (mirrors `useCommandAutocomplete`), the code is well-structured, and all tests pass. I identified a few minor items below but none are blocking.

---

### P1: Placeholder test is effectively a no-op

**File:** `packages/web/src/components/__tests__/MessageInput.reference-autocomplete.test.tsx`, line 218-233

The test titled "replaces @query at end of content with @ref token" is a placeholder:

```ts
it('replaces @query at end of content with @ref token', async () => {
    expect(true).toBe(true); // placeholder - see below for integration
});
```

The lengthy comment above it explains the difficulty of testing `handleReferenceSelect` with mocked hooks, but this should still be tested. The `extractActiveAtQuery` mock in this file (line 81-95) already contains the real implementation. You could set `mockDraftContent` to e.g. `'fix @task'`, call `mockReferenceHandleSelect` (or mock it to invoke the real logic), and verify `mockSetContent` was called with the expected replacement string. Alternatively, write a unit test for the `handleReferenceSelect` callback directly by extracting its logic into a testable pure function.

---

### P2: Both autocomplete menus can render simultaneously with identical positioning

**File:** `packages/web/src/components/InputTextarea.tsx`, lines 163-181

Both `ReferenceAutocomplete` and `CommandAutocomplete` are rendered inside the same `relative` container, both positioned with `bottom: '100%'` and `left: 0`, both at `z-50`. If a user types content like `/agent @task-42 fix`, the command autocomplete shows (content starts with `/`) and the reference autocomplete also shows (active `@` query), causing the two menus to overlap visually.

Consider adding logic to suppress the command autocomplete when the reference autocomplete is visible (or vice versa), or adjusting positioning so they don't collide.

---

### P2: `handleKeyDown` in MessageInput re-creates on every render due to object dependency

**File:** `packages/web/src/components/MessageInput.tsx`, line 276

The `handleKeyDown` callback depends on `referenceAutocomplete` and `commandAutocomplete`, which are new objects on every render (plain object return from hooks). This means `handleKeyDown` is re-created on every render. This existed before this PR with `commandAutocomplete`, so it is consistent, but now that there are two autocomplete systems both producing new objects per render, the inefficiency is amplified. A future improvement would be to depend on individual functions rather than the entire objects, e.g., `referenceAutocomplete.handleKeyDown` and `commandAutocomplete.handleKeyDown`.

---

### P3: `handleReferenceSelect` captures `content` in closure, recreating the callback chain on every keystroke

**File:** `packages/web/src/components/MessageInput.tsx`, lines 126-141

`handleReferenceSelect` depends on `content`, which changes on every keystroke. This cascades: a new `handleReferenceSelect` creates a new `useReferenceAutocomplete.onSelect`, which causes the hook's `handleSelect` and `handleKeyDown` to also be recreated. This is the same pattern used by `handleCommandSelect` (which depends on `setContent`), so it is consistent. A potential future improvement would be to use a ref to always read the latest `content` value instead of including it in the dependency array.

---

### Summary of positives

- Clean separation: `ReferenceAutocomplete` component, `useReferenceAutocomplete` hook, and `MessageInput` integration are well-layered
- Keyboard event priority is correctly implemented with reference autocomplete taking precedence
- The `extractActiveAtQuery` utility handles edge cases well (email addresses, multiple `@` tokens, completed mentions)
- The `handleReferenceSelect` content replacement logic is mathematically correct for all tested scenarios
- The mock fix in `MessageInput.queue-mode.test.tsx` prevents future breakage when new hooks are added
- All 75 tests pass (37 component + 38 hook), typecheck is clean
- UI design follows established patterns (matches `CommandAutocomplete` styling)

lsm (commented):

---

All three review findings from round 1 have been addressed. The fixes are correct and well-tested. Summary of each:

### P1: Placeholder test replaced with real assertions -- FIXED

The `expect(true).toBe(true)` placeholder in `packages/web/src/components/__tests__/MessageInput.reference-autocomplete.test.tsx` has been replaced with 4 substantive tests in a new "reference selection -- content replacement" describe block. The approach of capturing `capturedOnSelect` (the callback passed to `useReferenceAutocomplete`) via the mock is sound -- it exercises the real `handleReferenceSelect` logic in `MessageInput.tsx` while keeping hooks mocked for isolation. The test cases cover:
- Mid-sentence @query replacement (line 233)
- Slash-command prefix combination (line 247)
- Bare @ with no query text (line 261)
- No-@ early return guard (line 275)

The mock of `extractActiveAtQuery` (lines 92-106) accurately mirrors the real implementation in `packages/web/src/hooks/useReferenceAutocomplete.ts` (lines 55-77). I verified both implementations match: backward scan for last word-start @, whitespace-after-@ check, returns null when no active query.

### P2a: Overlapping menus -- FIXED

The fix in `MessageInput.tsx` line 471-473 suppresses the command autocomplete when the reference autocomplete is visible:
```tsx
showCommandAutocomplete={
    commandAutocomplete.showAutocomplete && !referenceAutocomplete.showAutocomplete
}
```
This is applied at the prop level before reaching `InputTextarea`, which is the right place. The two new tests (lines 288-311) verify mutual exclusion: command menu hidden when reference visible, and command menu shown when reference hidden.

### P2b: handleKeyDown closure efficiency -- PARTIALLY EFFECTIVE

The destructuring of stable refs (lines 241-242):
```tsx
const refHandleKeyDown = referenceAutocomplete.handleKeyDown;
const cmdHandleKeyDown = commandAutocomplete.handleKeyDown;
```
and the dependency array update to `[refHandleKeyDown, cmdHandleKeyDown, ...]` (line 282) is a marginal improvement. The comment claims the functions are "stable via useCallback" but that is not quite accurate: `useCommandAutocomplete.handleKeyDown` depends on `[showAutocomplete, filteredCommands, selectedIndex, handleSelect]` and `useReferenceAutocomplete.handleKeyDown` depends on `[showAutocomplete, results, selectedIndex, handleSelect, close]`. These change as autocomplete state changes. However, the fix is still a net improvement because it avoids re-creating the outer `handleKeyDown` when OTHER properties on the autocomplete objects change (like `searchQuery`) without `handleKeyDown` itself changing. No regressions -- just a slightly over-claimed comment. Not worth blocking on.

### Minor observations (P3)

1. The `useReferenceAutocomplete` mock now captures `opts.onSelect` (line 79-81) but previously was a static mock. The queue-mode test file also needed updating to include the `useReferenceAutocomplete` mock -- this was already done correctly in the initial commit (lines 85-93 of `MessageInput.queue-mode.test.tsx`), which was the right fix.

2. The non-null assertions on `container.querySelector('textarea')` were correctly relaxed to null checks in the fix commit, which is cleaner.

---

Verdict: All P1 and P2 issues are resolved. The code is correct, tests are meaningful, and no regressions were introduced. All 42 tests across the 3 relevant test files pass.

Recommendation: **APPROVE**

PR #846, opened 2026-03-25
feat(web): integrate reference autocomplete with InputTextarea

lsm:
- Add ReferenceAutocomplete component with grouped results by type
  (Tasks, Goals, Files, Folders), icons, shortId display, and keyboard hints
- Add reference autocomplete props to InputTextarea: showReferenceAutocomplete,
  referenceResults, selectedReferenceIndex, onReferenceSelect, onReferenceClose
- Implement mutual exclusion: reference autocomplete takes priority over command
  autocomplete when both are active
- Add @ref{} token badge below textarea showing count of active references
- Export insertReferenceMention() utility for replacing @query with @ref{type:id}
  token in content (used by MessageInput in the next wiring task)
- 24 new tests covering component rendering, selection, click-outside, badge
  count updates, insertion logic, and menu switching

### DIFF
--- a/packages/web/src/components/InputTextarea.tsx
+++ b/packages/web/src/components/InputTextarea.tsx
@@ -29,6 +29,8 @@ import { useEffect, useLayoutEffect, useRef, useState } from 'preact/hooks';
 import { cn } from '../lib/utils.ts';
 import { borderColors } from '../lib/design-tokens.ts';
 import CommandAutocomplete from './CommandAutocomplete.tsx';
+import ReferenceAutocomplete from './ReferenceAutocomplete.tsx';
+import type { ReferenceMention, ReferenceSearchResult } from '@neokai/shared';
 
 export interface InputTextareaProps {
 	content: string;
@@ -44,6 +46,12 @@ export interface InputTextareaProps {
 	selectedCommandIndex?: number;
 	onCommandSelect?: (command: string) => void;
 	onCommandClose?: () => void;
+	// Reference autocomplete
+	showReferenceAutocomplete?: boolean;
+	referenceResults?: ReferenceSearchResult[];
+	selectedReferenceIndex?: number;
+	onReferenceSelect?: (reference: ReferenceMention) => void;
+	onReferenceClose?: () => void;
 	// Agent state - passed as prop to avoid direct signal reads that cause re-renders
 	isAgentWorking?: boolean;
 	onStop?: () => void;
@@ -70,6 +78,11 @@ export function InputTextarea({
 	selectedCommandIndex = 0,
 	onCommandSelect,
 	onCommandClose,
+	showReferenceAutocomplete = false,
+	referenceResults,
+	selectedReferenceIndex,
+	onReferenceSelect,
+	onReferenceClose,
 	isAgentWorking = false,
 	onStop,
 	onPaste,
@@ -140,16 +153,30 @@ export function InputTextarea({
 	const showStop = isAgentWorking && !hasContent && !!onStop;
 	const textareaLeftPadding = leadingElement ? (leadingPaddingClass ?? 'pl-28') : 'pl-5';
 
+	// Count @ref{} tokens in content
+	const refCount = [...content.matchAll(/@ref\{[^}:]+:[^}]+\}/g)].length;
+
 	return (
 		<div class="relative flex-1">
-			{/* Command Autocomplete */}
-			{showCommandAutocomplete && onCommandSelect && onCommandClose && (
-				<CommandAutocomplete
-					commands={filteredCommands}
-					selectedIndex={selectedCommandIndex}
-					onSelect={onCommandSelect}
-					onClose={onCommandClose}
+			{/* Autocomplete menus — only one shown at a time; reference takes priority */}
+			{showReferenceAutocomplete && onReferenceSelect && onReferenceClose ? (
+				<ReferenceAutocomplete
+					results={referenceResults ?? []}
+					selectedIndex={selectedReferenceIndex ?? 0}
+					onSelect={onReferenceSelect}
+					onClose={onReferenceClose}
 				/>
+			) : (
+				showCommandAutocomplete &&
+				onCommandSelect &&
+				onCommandClose && (
+					<CommandAutocomplete
+						commands={filteredCommands}
+						selectedIndex={selectedCommandIndex}
+						onSelect={onCommandSelect}
+						onClose={onCommandClose}
+					/>
+				)
 			)}
 
 			<div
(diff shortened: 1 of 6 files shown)


lsm (commented):

### Summary

This PR introduces the reference autocomplete UI layer: a `ReferenceAutocomplete` dropdown component, integration props on `InputTextarea`, an `insertReferenceMention` utility, mutual exclusion with command autocomplete, a `@ref{}` token count badge, and thorough tests. The implementation is well-structured and follows existing patterns. All 94 tests pass, typecheck and lint are clean.

### Issues Found

#### P1: `onReferenceSelect` type mismatch in `InputTextareaProps`

**File:** `packages/web/src/components/InputTextarea.tsx` line 53

The prop is declared as:
```ts
onReferenceSelect?: (reference: ReferenceMention) => void;
```

But it is passed directly to `ReferenceAutocomplete`'s `onSelect`, which is typed as:
```ts
onSelect: (result: ReferenceSearchResult) => void;
```

The `ReferenceAutocomplete` component calls `onSelect(result)` with the full `ReferenceSearchResult` (which includes optional `shortId` and `subtitle`). So `onReferenceSelect` will receive a `ReferenceSearchResult`, not a `ReferenceMention`.

When MessageInput wires `useReferenceAutocomplete.handleSelect` (typed as `(result: ReferenceSearchResult) => void`) as `onReferenceSelect`, this may cause a TypeScript type error under `strictFunctionTypes` because the function parameter is contravariant. Even if it compiles (due to method-style bivariance), the type annotation is misleading.

**Recommendation:** Change the prop type to match what is actually passed:
```ts
onReferenceSelect?: (result: ReferenceSearchResult) => void;
```
This aligns with the `useReferenceAutocomplete.handleSelect` signature that MessageInput will pass in the wiring task.

#### P1: Duplicate regex for `@ref{}` token counting

**File:** `packages/web/src/components/InputTextarea.tsx` line 157

```ts
const refCount = [...content.matchAll(/@ref\{[^}:]+:[^}]+\}/g)].length;
```

The shared package already exports `REFERENCE_PATTERN` (in `packages/shared/src/types/reference.ts` line 105) with the exact same regex:
```ts
export const REFERENCE_PATTERN = /@ref\{([^}:]+):([^}]+)\}/g;
```

These two patterns are identical (same regex, same `g` flag). Importing and reusing `REFERENCE_PATTERN` from `@neokai/shared` avoids duplication and ensures they stay in sync if the pattern ever changes. However, note that `REFERENCE_PATTERN` has the `g` flag and is stateful, so counting would need to reset `lastIndex` before use, or the regex would need to be cloned:
```ts
const refCount = [...content.matchAll(new RegExp(REFERENCE_PATTERN.source, 'g'))].length;
```

#### P2: Keyboard navigation operates on flat indices but UI shows grouped layout

**File:** `packages/web/src/components/ReferenceAutocomplete.tsx` lines 102-109, 174-205

The `scrollIntoView` effect and selected-item highlighting use `originalIndex` (the flat index from the results array) to map to `selectedIndex`. This works correctly because the grouping logic preserves the flat index. However, the keyboard navigation in `useReferenceAutocomplete` (lines 211-213) increments/decrements `selectedIndex` as if navigating a flat list, which means arrow keys will jump across group boundaries (e.g., from the last task to the first file).

This is not necessarily wrong -- it could be intentional for UX simplicity. But the grouped header labels ("Tasks", "Goals") give a visual impression of separate sections. If the intent is flat navigation across groups, the current implementation is correct. If section-aware navigation is desired later, this should be documented.

#### P2: No ARIA attributes on autocomplete list items

**File:** `packages/web/src/components/ReferenceAutocomplete.tsx` lines 141-220

The autocomplete dropdown renders items as plain `<button>` elements without ARIA roles for listbox/option patterns. For accessibility, consider adding:
- `role="listbox"` on the container
- `role="option"` and `aria-selected` on each item button
- `aria-label"References"` on the container

This is important because keyboard-only and screen reader users interact with this dropdown.

#### P3: Minor: Click-outside handler uses `mousedown` not `pointerdown`

**File:** `packages/web/src/components/ReferenceAutocomplete.tsx` line 118

The click-outside handler uses `mousedown`, which is consistent with `CommandAutocomplete` (line 38). This is fine for consistency, but `pointerdown` would also cover touch interactions. Not blocking.

### What's Done Well

- **Excellent test coverage**: 94 tests across 3 files covering rendering, selection, click-outside, keyboard navigation, debouncing, stale response cancellation, badge counts, and mutual exclusion
- **Clean mutual exclusion pattern**: The ternary in InputTextarea (line 162) cleanly prevents both autocomplete menus from rendering simultaneously
- **Consistent with existing patterns**: `ReferenceAutocomplete` mirrors `CommandAutocomplete`'s structure (same positioning, styling, footer hints, click-outside behavior)
- **Good stale search cancellation**: The version counter pattern in `useReferenceAutocomplete` correctly prevents superseded RPC responses from updating state
- **Proper cleanup**: Debounce timer cleanup on unmount and on content change prevents memory leaks
- **`insertReferenceMention`**: Well-tested utility that handles edge cases like empty queries, paths with slashes, and existing tokens

### Verdict

The implementation is solid and well-tested. The P1 items (type mismatch and duplicate regex) should be addressed before the follow-up wiring task to avoid propagating incorrect types and regex duplication. P2 accessibility and navigation items can be addressed in a follow-up.

**Recommendation:** Approve after P1 items are fixed (or address in this PR).

lsm (commented):

### Summary

All four items from the previous review have been correctly addressed. The implementation is solid, tests pass (82/82), typecheck and lint are clean.

### Verification of Fixes

**P1: onReferenceSelect type mismatch -- FIXED**
- `InputTextareaProps.onReferenceSelect` at line 54 now correctly accepts `ReferenceSearchResult` instead of `ReferenceMention` (`packages/web/src/components/InputTextarea.tsx`).
- The unused `ReferenceMention` import was removed from the import statement (line 33). No stale imports remain.
- The type chain is now consistent: `ReferenceAutocompleteProps.onSelect` -> `InputTextareaProps.onReferenceSelect` both use `ReferenceSearchResult`. The `useReferenceAutocomplete` hook continues to do the `ReferenceSearchResult` -> `ReferenceMention` conversion internally via `handleSelect`, which is the right layer for that concern.

**P1: Duplicate regex replaced -- FIXED**
- Line 159 now uses `new RegExp(REFERENCE_PATTERN.source, 'g')` instead of an inline `/@ref\{[^}:]+:[^}]+\}/g` literal.
- Verified both patterns produce identical match results. The approach avoids stale `lastIndex` from the shared global-flag regex by constructing a fresh instance each render, as the comment on line 157-158 correctly documents.
- `REFERENCE_PATTERN` is imported from `@neokai/shared` (line 34).

**P2: ARIA attributes added -- FIXED**
- `role="listbox"` and `aria-label={headerLabel}` on the container div (`packages/web/src/components/ReferenceAutocomplete.tsx` lines 156-157).
- `role="option"` and `aria-selected={globalIndex === selectedIndex}` on each item button (lines 203-204).
- This follows standard WAI-ARIA combobox/listbox patterns for autocomplete widgets.

**Merge conflict resolution -- VERIFIED**
- The dev branch version of `ReferenceAutocomplete.tsx` was taken, which includes the `position` prop, dynamic `headerLabel` via `resolveHeaderLabel()`, and `TYPE_ORDER`-sorted groups.
- The fix commit diff shows only 4 lines added (the ARIA attributes) to this file, confirming the rebase was clean.

### Additional Checks

- All 82 tests pass (InputTextarea: 30, useReferenceAutocomplete: 52).
- TypeScript typecheck passes with zero errors.
- Oxlint passes with zero warnings/errors.
- No regressions in existing tests or type signatures.
- The `@ref{}` token counting badge works correctly for 0, 1, and multiple tokens (verified via tests at lines 781-853 of the test file).

### Recommendation

All previous issues resolved. The code is clean, well-tested, and follows project conventions. Ready to merge.

PR #847, opened 2026-03-25
fix(space): Phase 1 critical gap fixes for Space system reliability

lsm:
Three Phase 1 critical fixes from the Space V2 gap analysis:

1. Goal Management Integration: Wire Space task completion to goal progress
   recalculation. Added recalculateProgressFromSpaceTasks() to GoalRepository
   and integrated it into TaskAgentManager.handleSubSessionComplete() so
   completing a Space task with a goalId updates the goal's progress.

2. Sub-session Streaming Restart: Fixed TaskAgentManager.rehydrateTaskAgent()
   to call startStreamingQuery() on restored sub-sessions. Previously sub-sessions
   were restored to memory but not actively streaming, leaving them dormant.

3. Pending Run Rehydration: Fixed space-workflow-run-repository.getRehydratableRuns()
   to include 'pending' runs, and updated rehydrateExecutors() to handle them:
   - Pending runs with existing tasks are transitioned to 'in_progress' and rebuilt
   - Pending runs without tasks (crash before task creation) are safely cancelled

Tests added for all three fixes.

### DIFF
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -497,8 +497,11 @@ export class SpaceRuntime {
 		const spaces = await this.config.spaceManager.listSpaces(false);
 
 		for (const space of spaces) {
-			// getRehydratableRuns returns 'in_progress' AND 'needs_attention' runs.
-			// 'pending' is still excluded — it's transient (task creation may have failed).
+			// getRehydratableRuns returns 'pending', 'in_progress' AND 'needs_attention' runs.
+			// 'pending' runs may exist if the daemon crashed between createRun() and
+			// updateStatus('in_progress'). We check whether tasks were already created:
+			//   - Tasks exist  → transition to in_progress and build executor normally
+			//   - No tasks     → transition to cancelled (task creation never happened)
 			// 'needs_attention' runs are included so a human-gate-blocked run gets its
 			// executor reloaded on restart, allowing it to advance once the gate is resolved.
 			const activeRuns = this.config.workflowRunRepo.getRehydratableRuns(space.id);
@@ -514,6 +517,30 @@ export class SpaceRuntime {
 					continue;
 				}
 
+				// Handle 'pending' runs that were interrupted before updateStatus('in_progress').
+				// We need to determine whether task creation completed before the crash.
+				if (run.status === 'pending') {
+					const existingTasks = this.config.taskRepo.listByWorkflowRun(run.id);
+					if (existingTasks.length === 0) {
+						// No tasks exist — the crash happened before task creation started.
+						// Cancel the run since we cannot safely resume without knowing what
+						// was supposed to be created.
+						log.info(
+							`SpaceRuntime: recovering orphaned 'pending' run ${run.id} — ` +
+								`no tasks found, cancelling run`
+						);
+						this.config.workflowRunRepo.updateStatus(run.id, 'cancelled');
+						continue;
+					}
+					// Tasks exist — the crash happened after task creation but before
+					// updateStatus('in_progress'). Complete the transition now.
+					log.info(
+						`SpaceRuntime: recovering 'pending' run ${run.id} — ` +
+							`${existingTasks.length} task(s) found, promoting to in_progress`
+					);
+					this.config.workflowRunRepo.updateStatus(run.id, 'in_progress');
+				}
+
 				const meta: ExecutorMeta = {
 					workflow,
 					spaceId: space.id,
@@ -521,7 +548,9 @@ export class SpaceRuntime {
 				};
 				this.executorMeta.set(run.id, meta);
 
-				const executor = this.buildExecutor(workflow, run, space.id, space.workspacePath);
+				// Re-read the run from DB to pick up the status transition above
+				const freshRun = this.config.workflowRunRepo.getRun(run.id)!;
+				const executor = this.buildExecutor(workflow, freshRun, space.id, space.workspacePath);
 				this.executors.set(run.id, executor);
 			}
 		}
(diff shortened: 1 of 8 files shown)


lsm (commented):

All three fixes are well-structured and the core logic is sound. Tests are thorough for fixes #2 and #3, and excellent for the `recalculateProgressFromSpaceTasks` method in isolation. However, there is one P0 runtime bug that must be fixed before merge, plus a few lower-severity items.

---

### P0 — ReactiveDatabase cast will cause silent reactive notification failure (goal progress)

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 914-917

```typescript
const goalRepo = new GoalRepository(
    this.config.db.getDatabase(),
    this.config.db as unknown as ReactiveDatabase
);
```

`this.config.db` is the `Database` facade from `packages/daemon/src/storage/index.ts` — it does NOT implement `ReactiveDatabase`. The `as unknown as` cast masks a type mismatch that will cause a runtime error. When `recalculateProgressFromSpaceTasks` calls `this.updateGoal(goalId, { progress })`, the `GoalRepository.updateGoal` method executes the SQL `UPDATE` successfully, then calls `this.reactiveDb.notifyChange('goals')` — which will throw `TypeError: this.reactiveDb.notifyChange is not a function`.

**Impact:** The DB write succeeds (progress is persisted), but the reactive change notification never fires. The frontend will not see the goal progress update until it polls or refreshes. The error is caught by the surrounding try/catch and logged as a warning, masking the root cause.

**Fix:** The `TaskAgentManager` needs access to a real `ReactiveDatabase` instance. Options:
1. Add a `reactiveDb` field to `TaskAgentManagerConfig` and pass it during construction
2. Expose a `getGoalRepo()` or similar method on the `Database` facade and use that instead of instantiating `GoalRepository` directly
3. Use `this.config.db.getGoalRepo()` which returns a properly-constructed `GoalRepository` already wired with the correct `ReactiveDatabase`

Option 3 is the cleanest — the `Database` facade already has `getGoalRepo()` at line 412 of `storage/index.ts`. However, this returns a repository that was constructed at initialization time and should work correctly.

---

### P1 — Missing integration test for goal progress in `handleSubSessionComplete`

The `recalculateProgressFromSpaceTasks` method has excellent isolated unit tests (389 lines, 12 cases), but there is no test that verifies `handleSubSessionComplete` actually calls it when a completed step task has a `goalId`. This is the critical integration point — if the wiring in `handleSubSessionComplete` is incorrect or broken in a refactor, the isolated tests would still pass while the feature silently fails.

The fact that the P0 ReactiveDatabase bug above exists and wasn't caught by tests is evidence that this gap matters.

---

### P2 — Dead code in `recalculateProgressFromSpaceTasks`

**File:** `packages/daemon/src/storage/repositories/goal-repository.ts`, lines 683-686

```typescript
if (taskCount === 0) {
    this.updateGoal(goalId, { progress: 0 });
    return 0;
}
```

This branch is unreachable. The early return at line 663 (`if (tasks.length === 0)`) already handles the empty-tasks case. Since `taskCount` is incremented once per task in the `for` loop (lines 671-680), and the loop only runs when `tasks.length > 0`, `taskCount` can never be 0 at line 683. This should be removed to avoid confusion about control flow.

---

### P3 — Minor: Non-null assertion on `getRun` after status transition

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, line 552

```typescript
const freshRun = this.config.workflowRunRepo.getRun(run.id)!;
```

The `!` non-null assertion is defensible here (the run was just fetched and updated moments ago, and SQLite is single-threaded), but a guard with a warning log would be more defensive. Very minor.

---

### Summary of what's done well

- The `recalculateProgressFromSpaceTasks` algorithm correctly mirrors `GoalManager.calculateProgressFromTasks()` — status-to-progress mapping is identical
- The structural typing for the `spaceTaskRepo` parameter is a good pattern — it allows easy testing without the full `SpaceTaskRepository`
- The pending run recovery logic in `rehydrateExecutors()` is thorough — the two-case handling (tasks exist vs. no tasks) covers the crash-window scenarios correctly
- Re-reading the run from DB after status transition (`freshRun`) avoids stale state in the executor
- Sub-session streaming restart error handling is non-fatal with proper logging
- Test coverage for the streaming restart behavior (both success and failure) is solid

lsm (commented):

### Summary

Re-reviewed commit `be5f1e295` which addresses all three issues from the previous review. All fixes are correct and no new problems were introduced.

### P0 (Runtime bug): Fixed

`this.config.db as unknown as ReactiveDatabase` has been replaced with `this.config.db.getGoalRepo()` at `packages/daemon/src/lib/space/runtime/task-agent-manager.ts:912`. This correctly uses the existing factory method on the `Database` class (defined at `packages/daemon/src/storage/index.ts:412`), which returns a properly constructed `GoalRepository` instance with the correct constructor arguments (`db`, `reactiveDb`, `shortIdAllocator`). The now-unused imports for `GoalRepository` and `ReactiveDatabase` have been removed, and I confirmed no other references to these types remain in the file.

### P1 (Missing integration test): Fixed

A new test `'recalculates goal progress when step task with goalId completes'` has been added at `packages/daemon/tests/unit/space/task-agent-manager.test.ts:1044`. The test:
- Seeds a goal into the DB with progress 0
- Creates a full workflow stack (workflow, node, run) to satisfy FK constraints
- Creates a parent task and a step task with `goalId` set
- Spawns the TaskAgent so internal state is ready
- Calls `handleSubSessionComplete` and asserts goal progress becomes 100

The test correctly validates the end-to-end integration path. The `createTables(db)` call was added to `makeDb()` to ensure the `goals` table exists, and the `mockDb.getGoalRepo()` method was added to the test mock to expose the repository for assertions. The `noOpReactiveDb` helper follows the established pattern used across 30+ other test files.

### P2 (Dead code): Fixed

The unreachable `if (taskCount === 0)` branch has been removed from `recalculateProgressFromSpaceTasks` at `packages/daemon/src/storage/repositories/goal-repository.ts:682`. The early return at line 663-666 (`if (tasks.length === 0)`) already handles the empty case, making the removed branch provably unreachable. The resulting logic is clean: loop over all tasks, accumulate progress, then divide by count.

### Verification

- All 92 tests in `task-agent-manager.test.ts` pass
- All 12 tests in `goal-repository-space-goal-progress.test.ts` pass (dead code removal did not break existing coverage)
- Oxlint reports 0 warnings/errors on all changed files

### Recommendation: APPROVE

All three issues are properly resolved. The fixes are minimal, correct, and well-tested.

PR #848, opened 2026-03-25
feat(agent): implement agent context injection for @ references

lsm:
## Summary

- Create `packages/daemon/src/lib/agent/reference-context-builder.ts` with `buildReferenceContext()` that formats resolved task/goal/file/folder references as readable markdown, enforces a 200 KB total limit with priority ordering (task > goal > file > folder), and logs a warning on truncation. Also exports `prependContextToMessage()` helper that attaches the context block above the user message delimited by `---`.
- Create `packages/daemon/src/lib/agent/reference-context-injector.ts` with `ReferenceContextInjectorImpl` that extracts `@ref{}` tokens, resolves each one against task/goal repos or the workspace filesystem, and calls the builder. Resolution failures are non-fatal.
- Integrate into `MessagePersistence`: add `ReferenceContextInjector` interface + `setReferenceContextInjector()` setter; injection runs after built-in command expansion and before SDK message construction.
- Add delegation setter on `SessionManager` and wire up `ReferenceContextInjectorImpl` in `app.ts` after `setupRPCHandlers` (where `TaskRepository` and `reactiveDb` are available).

## Test plan

- [x] 29 unit tests in `packages/daemon/tests/unit/reference-context-builder.test.ts` covering:
  - Empty reference set → empty string
  - Task formatting (all fields, optional fields absent, shortId fallback)
  - Goal formatting (with/without mission type, metrics with/without unit, multiple metrics)
  - File formatting (text, binary, truncated, null content)
  - Folder formatting (entries with directory suffix, empty folder)
  - Priority ordering: task before goal before file before folder
  - Context truncation: high-priority entries kept, low-priority dropped
  - `prependContextToMessage` with and without context
- [x] Full daemon unit test suite (7674 tests) passes
- [x] Lint, typecheck, and knip all clean

### DIFF
--- a/packages/daemon/src/lib/agent/reference-context-builder.ts
+++ b/packages/daemon/src/lib/agent/reference-context-builder.ts
@@ -1,0 +1,202 @@
+/**
+ * Reference Context Builder
+ *
+ * Builds a markdown-formatted context block from resolved @ references.
+ * The context block is prepended to user messages so the agent has full entity data
+ * when processing mentions like @ref{task:t-42} or @ref{file:src/lib/utils.ts}.
+ */
+
+import type { MissionMetric, NeoTask, ResolvedReference, RoomGoal } from '@neokai/shared';
+import { Logger } from '../logger';
+
+const log = new Logger('reference-context-builder');
+
+/** Maximum total byte size of the injected context block. */
+export const MAX_CONTEXT_BYTES = 200_000;
+
+/**
+ * Truncation priority: references earlier in this list are kept first when the
+ * 200 KB limit is reached.
+ */
+const PRIORITY_ORDER: ReadonlyArray<ResolvedReference['type']> = ['task', 'goal', 'file', 'folder'];
+
+/**
+ * Build a markdown context block from a map of resolved references.
+ *
+ * References are sorted by priority (task > goal > file > folder) and accumulated
+ * until the 200 KB byte limit is reached. A warning is logged when truncation
+ * occurs. Returns an empty string when the input map is empty or all references
+ * produce no output.
+ *
+ * @param references - Map of @ref{…} token → resolved reference, as produced by
+ *   `ReferenceResolver.resolveAllReferences`. Keys are arbitrary; only the values
+ *   are formatted.
+ */
+export function buildReferenceContext(references: Record<string, ResolvedReference>): string {
+	const entries = Object.values(references);
+	if (entries.length === 0) {
+		return '';
+	}
+
+	// Stable sort by priority (task > goal > file > folder).
+	// Unknown types (indexOf returns -1) are placed after 'folder'.
+	const sorted = [...entries].sort((a, b) => {
+		const pa = PRIORITY_ORDER.indexOf(a.type);
+		const pb = PRIORITY_ORDER.indexOf(b.type);
+		const ra = pa === -1 ? PRIORITY_ORDER.length : pa;
+		const rb = pb === -1 ? PRIORITY_ORDER.length : pb;
+		return ra - rb;
+	});
+
+	const sections: string[] = [];
+	let totalBytes = 0;
+	let truncated = false;
+
+	for (const ref of sorted) {
+		const section = formatReference(ref);
+		if (!section) {
+			continue;
+		}
+		const sectionBytes = Buffer.byteLength(section, 'utf8');
+		if (totalBytes + sectionBytes > MAX_CONTEXT_BYTES) {
+			truncated = true;
+			break;
+		}
+		sections.push(section);
+		totalBytes += sectionBytes;
+	}
+
+	if (truncated) {
+		log.warn(
+			`Reference context truncated at ${totalBytes} bytes (limit: ${MAX_CONTEXT_BYTES} bytes). ` +
+				`Some referenced entities were omitted.`
+		);
+	}
+
+	if (sections.length === 0) {
+		return '';
+	}
+
+	return `## Referenced Entities\n\n${sections.join('\n')}`;
+}
+
+/**
+ * Prepend a reference context block to a user message, separated by a horizontal rule.
+ * Returns the original message unchanged when `context` is empty.
+ */
+export function prependContextToMessage(userMessage: string, context: string): string {
+	if (!context) {
+		return userMessage;
+	}
+	return `${context}\n\n---\n\n${userMessage}`;
+}
+
+// ============================================================================
+// Per-type formatters
+// ============================================================================
+
+function formatReference(ref: ResolvedReference): string {
+	switch (ref.type) {
+		case 'task':
+			return formatTask(ref.data as NeoTask, ref.id);
+		case 'goal':
+			return formatGoal(ref.data as RoomGoal, ref.id);
+		case 'file':
+			return formatFile(
+				ref.data as {
+					path: string;
+					content: string | null;
+					binary: boolean;
+					truncated: boolean;
+				}
+			);
+		case 'folder':
+			return formatFolder(
+				ref.data as {
+					path: string;
+					entries: Array<{ name: string; type: 'file' | 'directory' }>;
+				}
+			);
+		default:
+			return '';
+	}
+}
+
+function formatTask(task: NeoTask, fallbackId: string): string {
+	const id = task.shortId ?? fallbackId;
+	const lines: string[] = [`### Task: ${id}`];
+	lines.push(`**Title:** ${task.title}`);
+	lines.push(`**Status:** ${task.status}`);
+	lines.push(`**Priority:** ${task.priority}`);
+	if (task.progress !== null && task.progress !== undefined) {
+		lines.push(`**Progress:** ${task.progress}%`);
+	}
+	if (task.description) {
+		lines.push(`**Description:** ${task.description}`);
+	}
+	if (task.currentStep) {
+		lines.push(`**Current Step:** ${task.currentStep}`);
+	}
+	return lines.join('\n') + '\n';
+}
+
+function formatGoal(goal: RoomGoal, fallbackId: string): string {
+	const id = goal.shortId ?? fallbackId;
+	const lines: string[] = [`### Goal: ${id}`];
+	lines.push(`**Title:** ${goal.title}`);
+	if (goal.missionType) {
+		lines.push(`**Type:** ${goal.missionType}`);
+	}
+	lines.push(`**Status:** ${goal.status}`);
+	lines.push(`**Progress:** ${goal.progress}%`);
+	if (goal.description) {
+		lines.push(`**Description:** ${goal.description}`);
+	}
+	if (goal.structuredMetrics && goal.structuredMetrics.length > 0) {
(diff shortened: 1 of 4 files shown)


lsm (commented):

---

### Summary

The implementation is well-structured and cleanly separates concerns: `reference-context-builder.ts` is a pure formatter with no side effects, `reference-context-injector.ts` handles resolution logic, and the integration into `message-persistence.ts` is minimal and non-breaking. The 29 unit tests for the builder are thorough. Typecheck and lint pass cleanly.

I am recommending **APPROVE** with a few P2 and P3 items to consider.

---

### P2 — Misleading JSDoc on `setReferenceContextInjector`

**File:** `packages/daemon/src/lib/session/session-manager.ts`, lines 411-417

The JSDoc states:

> Has no effect if called with the same injector instance more than once.

But the implementation unconditionally overwrites the field — there is no idempotency guard. Either add a reference-equality check (`if (this.referenceContextInjector === injector) return;`) or remove the misleading claim from the JSDoc. This matters because the doc suggests a safety property that does not hold.

---

### P2 — Missing tests for `ReferenceContextInjectorImpl`

There are 29 tests for `reference-context-builder.ts` (the pure formatter), but **zero** tests for `reference-context-injector.ts` (the class that does the actual resolution). The injector contains non-trivial logic:

- Mention extraction with regex state management (`REFERENCE_PATTERN.lastIndex` reset)
- Session context resolution with fallback
- Per-type resolution helpers (task, goal, file, folder) including path traversal guards
- Sequential resolution with per-reference error swallowing

The builder tests verify formatting and truncation, but none of the resolution, deduplication, or error-swallowing behavior is covered. The injector is the integration-critical piece — a bug there would silently produce wrong context or no context at all (since failures are non-fatal).

---

### P2 — Duplicate resolution logic between `reference-context-injector.ts` and `reference-handlers.ts`

The file resolution logic in `reference-context-injector.ts` (lines 202-281) is essentially a copy-paste of the logic in `reference-handlers.ts` (lines 218-312). Both implement the same path traversal guard, binary detection, file read, and truncation. Similarly for `resolveFolder` and the task/goal resolution helpers.

This duplication means any bug fix or behavior change must be applied in two places. Consider extracting shared resolution helpers that both the RPC handler and the injector can call. This is not a blocking issue, but it is a maintenance hazard.

---

### P3 — Raw `@ref{}` tokens left in the user message after context injection

`prependContextToMessage` (in `reference-context-builder.ts`, line 86) prepends the expanded context block but does **not** strip the raw `@ref{type:id}` tokens from the user message. The agent receives both the expanded context at the top AND the original `@ref{task:t-42}` tokens inline.

This is not a bug — the tokens are harmless and the agent can ignore them — but it is mildly redundant. If intentional, a brief comment explaining the design decision would be helpful.

---

### P3 — Minor double-`stat` in `resolveFile`

**File:** `packages/daemon/src/lib/agent/reference-context-injector.ts`, lines 222-261

The `resolveFile` function manually calls `stat` for binary detection, then calls `fileManager.readFile` which internally calls `stat` again (via `FileManager.readFile` at `file-manager.ts`, line 64). This results in two `stat` syscalls per file reference. Not performance-critical, but worth noting.

---

### P3 — Truncation warning message could be clearer

**File:** `packages/daemon/src/lib/agent/reference-context-builder.ts`, lines 68-73

When a single entry exceeds the 200KB limit, the warning says:

> Reference context truncated at 0 bytes (limit: 200000 bytes).

"truncated at 0 bytes" reads oddly. Consider wording like:

> Reference context exceeded 200000 byte limit before any entries could be added. 1 entry was omitted.

---

### What looks good

- **Clean separation of concerns**: Builder is pure and testable; injector handles I/O.
- **Priority ordering**: task > goal > file > folder is correctly implemented with stable sort.
- **200KB truncation**: Accumulates sections greedily, breaks on overflow, warns on truncation.
- **Non-fatal error handling**: Individual reference failures do not block message dispatch.
- **Late wiring pattern**: Setter injection avoids circular dependency in startup sequence.
- **Code style**: Tab indentation, single quotes, semicolons, trailing commas all correct. No console.log violations. No `no-explicit-any` violations (type assertions are used appropriately for the polymorphic `data` field).
- **Path traversal guards**: Properly handled for both files and folders via `FileManager.validatePath`.

lsm (commented):

---

### Verdict: APPROVE

All three P2 issues from the previous review have been properly addressed. The fix commit is clean and introduces no new issues.

---

### P2-1 Fix: Misleading JSDoc -- RESOLVED

**File:** `packages/daemon/src/lib/session/session-manager.ts`, lines 411-417

The misleading "Has no effect if called with the same injector instance more than once" claim has been removed. The updated JSDoc now accurately describes the method's purpose and the ordering constraint (must be called after setupRPCHandlers).

---

### P2-2 Fix: Missing tests for ReferenceContextInjectorImpl -- RESOLVED

**File:** `packages/daemon/tests/unit/reference-context-injector.test.ts` (28 tests, 443 lines)

Excellent test coverage added covering all critical paths:
- `extractMentions`: deduplication, regex state reset, all token types (8 tests)
- Session context fallback (workspaceRoot fallback, no roomId) (2 tests)
- Task resolution: UUID lookup, shortId fallback, cross-room guard, not-found, no-roomId (5 tests)
- Goal resolution: resolved, cross-room guard, not-found (3 tests)
- File resolution: text content, missing file, path traversal (3 tests)
- Folder resolution: normal listing, missing folder (2 tests)
- Error handling: per-reference error swallowing, all-fail returns null (2 tests)
- Output format: separator, multiple types (2 tests)

Tests use real temp directories for file/folder resolution and proper mocks for task/goal repos. The regex state reset test (calling `extractMentions` twice on the same content) is a good edge case.

---

### P2-3 Fix: Duplicate resolution logic -- RESOLVED

**File:** `packages/daemon/src/lib/reference-resolver-helpers.ts` (216 lines, new)

The shared helpers module is well-structured:
- Exports four resolver functions: `resolveTaskReference`, `resolveGoalReference`, `resolveFileReference`, `resolveFolderReference`
- Exports `MAX_FILE_CONTENT_BYTES` constant (was duplicated in both files before)
- Clean module-level JSDoc explaining the purpose and what it avoids duplicating
- Both `reference-handlers.ts` and `reference-context-injector.ts` now import from this single source of truth
- `reference-handlers.ts` went from ~340 lines to ~143 lines (net -197 lines of duplicated code)
- The `resolveFileReference` helper properly performs path traversal validation, binary detection, file reading, and truncation in one place
- `resolveFolderReference` relies on `FileManager.validatePath` for path traversal protection (via its `listDirectory` method), which is the same pattern as the original code

The refactoring is faithful -- no behavioral changes were introduced. The callers in both files pass the same arguments and receive the same return types.

---

### Verification results

- **57 new tests pass** (29 builder + 28 injector) across 2 files, 99 expect() calls
- **TypeScript typecheck**: clean (`tsc --build --noEmit`)
- **Oxlint**: 0 warnings, 0 errors on 524 files
- **Code style**: Tab indentation, single quotes, semicolons, trailing commas -- all compliant

### Remaining P3 items (non-blocking, from previous review)

These were noted before and remain as-is, which is fine:
- Raw `@ref{}` tokens left in user message after context injection (by design, not a bug)
- Double `stat` syscall in `resolveFileReference` (performance non-critical)
- Truncation warning wording when totalBytes is 0 (cosmetic)

lsm (commented):

---

### Verdict: APPROVE

The restructured implementation is clean, correct, and well-tested. The simplification from the previous `ReferenceContextInjector` class to a direct `buildReferenceContext()` + `prependContextToMessage()` call in `MessagePersistence.persist()` is a meaningful improvement that reduces unnecessary abstraction.

### What was verified

**1. Simplified integration correctness**

The flow in `message-persistence.ts` is straightforward and correct:
- Step 2b (line 159-166): Preprocessing extracts and resolves references via `ReferenceResolver`, now returning `resolvedReferences` alongside `referenceMetadata`.
- Step 2c (line 168-170): `buildReferenceContext()` formats resolved references into a markdown block; `prependContextToMessage()` attaches it to the agent content.
- Step 3 (line 173): The SDK message content is built from `agentContent` (with prepended context), ensuring the agent sees the full entity data.
- The fallback path when `referenceResolver` is absent correctly returns empty `resolvedReferences`, so `buildReferenceContext({})` returns `''` and no injection occurs.

**2. Priority ordering and truncation**

- `PRIORITY_ORDER` at line 21 correctly defines `['task', 'goal', 'file', 'folder']`.
- The sort at lines 43-48 uses `Array.sort` which is stable in V8/Bun, so equal-priority items maintain insertion order.
- Unknown types (e.g., `'metric'`) correctly fall after known types via `indexOf === -1 ? length : pa` at line 46.
- The truncation logic at lines 60-67 accumulates sections until `MAX_CONTEXT_BYTES` (200KB) is exceeded, then breaks. A warning is logged.
- Edge case: if a single reference exceeds the limit, the result is empty string (no partial injection). This is the correct behavior -- a 201KB file reference should not produce a truncated, useless context block.

**3. Empty reference handling**

- `buildReferenceContext({})` returns `''` (line 37-39).
- `prependContextToMessage(msg, '')` returns the original `msg` unchanged (line 88-89).
- The `formatReference` default case returns `''` for unknown types (line 121), so unknown types are silently skipped.
- No injection occurs when all references fail to resolve (the `resolved` map is empty).

**4. Test coverage**

25 unit tests in `reference-context-builder.test.ts` covering:
- Empty input, each reference type (task, goal, file, folder), optional field omission, fallback IDs, priority ordering, truncation, metric formatting, and `prependContextToMessage` edge cases.
- The 22 pre-existing integration tests in `reference-message-persistence.test.ts` continue to pass, confirming no regressions in the `MessagePersistence` + `ReferenceResolver` integration.
- The 4 tests in `session/message-persistence.test.ts` also pass.

All 51 tests pass (25 + 22 + 4).

**5. No regressions in `reference-resolver.ts`**

The only change to `reference-resolver.ts` is adding the `resolvedReferences` field to the `PreprocessedMessage` interface (line 35). The return values in `preprocessReferences()` now include this field. This is backward-compatible since it only adds a property.

**6. Code style compliance**

- Biome format: Tabs, single quotes, semicolons, trailing commas -- all correct.
- Oxlint: No `console.log`, no `no-explicit-any` violations.
- TypeScript: `bun run typecheck` passes clean.
- Knip: No new unused exports introduced.

**P3 -- Bare code fences in `formatFile`** (`packages/daemon/src/lib/agent/reference-context-builder.ts:179`): The file content is wrapped in bare ``` fences without a language hint. If the referenced file contains markdown fences (e.g., a `.md` file), the rendering could break. Since this content is consumed by the Claude agent SDK as raw text (not rendered as HTML), this is cosmetic at worst. A language hint like ```text` would be more robust.

**P3 -- Context block visible in UI**: The prepended context block is stored in the SDK user message persisted to the database, so it appears in the user's chat bubble in the UI. This is likely intentional (similar to built-in command expansion at line 156), but worth noting as a UX consideration -- users will see the full `## Referenced Entities` block above their original message.

PR #850, opened 2026-03-25
feat(web): add MentionToken component for @ reference rendering

lsm:
Renders a single @ref{type:id} token as a styled pill with type-specific
colors (task=blue, goal=purple, file=green, folder=yellow), an SVG icon,
hover/focus tooltip showing full entity details, click/keyboard handlers,
and ARIA accessibility labels.

37 unit tests cover rendering, type styles, aria labels, click, keyboard
(Enter key), tooltip show/hide, status display, and metadata fallback.

### DIFF
--- a/packages/web/src/components/MentionToken.tsx
+++ b/packages/web/src/components/MentionToken.tsx
@@ -1,0 +1,176 @@
+import { useState } from 'preact/hooks';
+import { cn } from '../lib/utils.ts';
+import type { ReferenceMention, ReferenceMetadata, ReferenceType } from '@neokai/shared';
+
+export interface MentionTokenProps {
+	mention: ReferenceMention;
+	metadata?: ReferenceMetadata;
+	onClick?: () => void;
+}
+
+const TYPE_STYLES: Record<ReferenceType, { container: string; icon: string; label: string }> = {
+	task: {
+		container: 'bg-blue-500/15 text-blue-300 hover:bg-blue-500/25',
+		icon: 'text-blue-400',
+		label: 'task',
+	},
+	goal: {
+		container: 'bg-purple-500/15 text-purple-300 hover:bg-purple-500/25',
+		icon: 'text-purple-400',
+		label: 'goal',
+	},
+	file: {
+		container: 'bg-green-500/15 text-green-300 hover:bg-green-500/25',
+		icon: 'text-green-400',
+		label: 'file',
+	},
+	folder: {
+		container: 'bg-yellow-500/15 text-yellow-300 hover:bg-yellow-500/25',
+		icon: 'text-yellow-400',
+		label: 'folder',
+	},
+};
+
+function TokenIcon({ type, className }: { type: ReferenceType; className?: string }) {
+	if (type === 'task') {
+		return (
+			<svg
+				class={cn('w-3 h-3 shrink-0', className)}
+				fill="none"
+				viewBox="0 0 24 24"
+				stroke="currentColor"
+				aria-hidden="true"
+			>
+				<path
+					stroke-linecap="round"
+					stroke-linejoin="round"
+					stroke-width={2}
+					d="M9 5H7a2 2 0 00-2 2v12a2 2 0 002 2h10a2 2 0 002-2V7a2 2 0 00-2-2h-2M9 5a2 2 0 002 2h2a2 2 0 002-2M9 5a2 2 0 012-2h2a2 2 0 012 2m-6 9l2 2 4-4"
+				/>
+			</svg>
+		);
+	}
+	if (type === 'goal') {
+		return (
+			<svg
+				class={cn('w-3 h-3 shrink-0', className)}
+				fill="none"
+				viewBox="0 0 24 24"
+				stroke="currentColor"
+				aria-hidden="true"
+			>
+				<path
+					stroke-linecap="round"
+					stroke-linejoin="round"
+					stroke-width={2}
+					d="M3 21v-4m0 0V5a2 2 0 012-2h6.5l1 1H21l-3 6 3 6h-8.5l-1-1H5a2 2 0 00-2 2zm9-13.5V9"
+				/>
+			</svg>
+		);
+	}
+	if (type === 'folder') {
+		return (
+			<svg
+				class={cn('w-3 h-3 shrink-0', className)}
+				fill="none"
+				viewBox="0 0 24 24"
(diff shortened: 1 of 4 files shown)


lsm (commented):

This is a clean, well-structured component. All 37 tests pass, linting is clean, and typecheck passes. The component follows existing patterns in the codebase (matching `ReferenceAutocomplete` icon style and test conventions). Below are the issues I found.

---

### P1: Missing Space key activation for `role="button"` elements

**File:** `packages/web/src/components/MentionToken.tsx`, line 125-130

The WAI-ARIA spec requires that elements with `role="button"` respond to both **Enter** and **Space** key presses for activation (see [WAI-ARIA button pattern](https://www.w3.org/WAI/ARIA/apg/patterns/button/)). Currently only Enter is handled:

```tsx
const handleKeyDown = (e: KeyboardEvent) => {
    if (e.key === 'Enter' && onClick) {
        e.preventDefault();
        onClick();
    }
};
```

Screen reader users (and keyboard-only users) commonly press Space to activate buttons. This should also handle `' '` (Space) and call `e.preventDefault()` to prevent page scroll.

**Corresponding test gap:** The test at line 211 verifies that Space does NOT call onClick, which is correct for the current implementation but will need updating once Space support is added. There is no test for Space activation.

---

### P1: Tooltip not programmatically associated via `aria-describedby`

**File:** `packages/web/src/components/MentionToken.tsx`, lines 134-173

The tooltip has `role="tooltip"` but no `id` on the tooltip element and no `aria-describedby` on the token span. This means assistive technology cannot programmatically discover the tooltip content. The current approach only works if the user happens to focus the element, which may cause the tooltip to show visually, but screen readers won't announce it as associated descriptive content.

A proper fix would use a stable ID (from `useId()` or a prop) and wire `aria-describedby` on the interactive span to the tooltip's `id`.

---

### P2: `tabIndex={0}` on non-interactive tokens is unexpected

**File:** `packages/web/src/components/MentionToken.tsx`, line 136

When `onClick` is not provided, the token still has `tabIndex={0}`, making it focusable in the tab order even though it has no interactive behavior. This creates an unnecessary tab stop that provides no additional value -- the element has no `role` (since `role` is undefined without onClick) and no action, so focus would only show the tooltip.

Consider conditionally setting `tabIndex` only when `onClick` is provided, or at minimum documenting the rationale for making non-interactive tokens focusable.

---

### P2: Duplicated `TokenIcon` SVG paths with `ReferenceAutocomplete`

**File:** `packages/web/src/components/MentionToken.tsx`, lines 34-106 vs `packages/web/src/components/ReferenceAutocomplete.tsx`, lines 23-91

Both `MentionToken.TokenIcon` and `ReferenceAutocomplete.TypeIcon` render identical SVG icons for each `ReferenceType`. The only difference is the icon size class (`w-3 h-3` vs `w-3.5 h-3.5`) and the color class. This is a maintainability concern -- if an icon path changes, both files need updating.

Consider extracting a shared `ReferenceTypeIcon` component that accepts `size` and `className` props.

---

### P3: Redundant `beforeEach`/`afterEach` in test file

**File:** `packages/web/src/components/__tests__/MentionToken.test.tsx`, lines 36-42

The `vitest.setup.ts` file already calls `vi.clearAllMocks()` in `beforeEach` and `vi.restoreAllMocks()` in `afterEach` globally. The local `beforeEach` with `vi.resetAllMocks()` and `afterEach` with `vi.restoreAllMocks()` in this test file are redundant (though not harmful, since `resetAllMocks` is a superset of `clearAllMocks`). This follows the pattern set by `ReferenceAutocomplete.test.tsx`, so it is consistent -- just noting for awareness.

---

### P3: Test uses fragile DOM selector for tooltip child count assertion

**File:** `packages/web/src/components/__tests__/MentionToken.test.tsx`, lines 296-300

The test "tooltip does not show status when absent" uses `querySelector('span')` on the tooltip element and then counts child spans. Since the tooltip structure is nested spans, `querySelector('span')` returns the first `span` child of the tooltip, which works but is fragile if the DOM structure changes. A more robust assertion would use a data-testid or check that no element with the status class exists.

lsm (commented):

All four issues from the first review have been properly addressed. No regressions detected.

### Issue Resolution

**P1: Missing Space key activation for role="button"** -- Fixed. Line 54 in `packages/web/src/components/MentionToken.tsx` now handles both `Enter` and `Space` keys. The "other keys" test (line 237-244) was also updated to use `Tab` instead of the old `Space` test case, which is correct. New test at line 229-235 explicitly verifies Space key behavior.

**P1: Missing aria-describedby on tooltip** -- Fixed. A `useId()` hook generates a stable tooltip ID (line 43). The `aria-describedby` attribute is conditionally set when `showTooltip` is true (line 66), and the tooltip element uses the matching `id` (line 86). Tests at lines 166-180 verify both the visible and hidden states.

**P2: Unconditional tabIndex on non-interactive tokens** -- Fixed. `tabIndex` is now conditional: `tabIndex={onClick ? 0 : undefined}` (line 64). The existing test was updated and a new negative test was added (lines 160-164) confirming no tabindex is rendered without onClick.

**P2: Duplicated SVG icon code** -- Fixed. A shared `ReferenceTypeIcon` component was extracted to `packages/web/src/components/ReferenceTypeIcon.tsx` (86 lines). Both `MentionToken` and `ReferenceAutocomplete` now import it, each passing their own className for type-specific sizing/colors. The shared component correctly adds `aria-hidden="true"` to all SVG icons, which is an improvement over the original `ReferenceAutocomplete` that was missing this attribute.

### Test Coverage

Test count increased from 37 to 41 (MentionToken) -- the 5 new tests directly cover the fixed behavior:
- Space key activation (line 229)
- Conditional tabIndex when onClick is absent (line 160)
- aria-describedby linking when tooltip visible (line 166)
- aria-describedby absent when tooltip hidden (line 176)
- Updated "other keys" test (line 237) now uses Tab instead of Space

All 41 MentionToken tests and 26 ReferenceAutocomplete tests pass. TypeScript compiles cleanly.

### Minor Observation (P3)

The `ReferenceTypeIcon` component uses a chain of `if` statements with an implicit `file` fallback at the end (line 69). This works correctly for the exhaustive `ReferenceType` union (`'task' | 'goal' | 'file' | 'folder'`), but a future new type would silently render the file icon. Not actionable now given the type constraint, just noting it for awareness.

**Recommendation: APPROVE**

PR #851, opened 2026-03-25
feat(e2e): add room entity and reference autocomplete helpers

lsm:
Add reusable E2E helper functions for the @ referencing system E2E tests:

- room-helpers.ts: createTask, createGoal, createRoomWithTask,
  createRoomWithGoal, createRoomWithTaskAndGoal, createTestFile,
  deleteTestFile, cleanupRoom, cleanupAllCreatedEntities
- reference-helpers.ts: typeInChatInput, waitForReferenceAutocomplete,
  getReferenceAutocompleteItems, selectReferenceByIndex,
  selectReferenceByClick, waitForMentionToken, getMentionTokenText,
  hoverMentionToken

### DIFF
--- a/packages/e2e/tests/helpers/reference-helpers.ts
+++ b/packages/e2e/tests/helpers/reference-helpers.ts
@@ -1,0 +1,161 @@
+/**
+ * Reference Autocomplete E2E Helpers
+ *
+ * Shared helper functions for @ reference autocomplete interactions in E2E tests.
+ * All helpers interact through the browser UI (clicks, keyboard, DOM queries).
+ * No direct state access is performed.
+ */
+
+import type { Page, Locator } from '@playwright/test';
+
+// ─── Selectors ────────────────────────────────────────────────────────────────
+
+/** The chat textarea that accepts user input */
+const CHAT_INPUT_SELECTOR = 'textarea[placeholder*="Ask"]';
+
+/** The reference autocomplete dropdown (role="listbox") */
+const AUTOCOMPLETE_SELECTOR = '[role="listbox"]';
+
+/** Individual reference items inside the autocomplete */
+const AUTOCOMPLETE_ITEM_SELECTOR = '[role="option"]';
+
+// ─── Autocomplete Helpers ─────────────────────────────────────────────────────
+
+/**
+ * Type text in the chat input field.
+ *
+ * Uses `fill` for plain text and `pressSequentially` for sequences that include
+ * the `@` trigger character so the React-like signal-based onChange fires.
+ */
+export async function typeInChatInput(page: Page, text: string): Promise<void> {
+	const textarea = page.locator(CHAT_INPUT_SELECTOR).first();
+	await textarea.waitFor({ state: 'visible', timeout: 5000 });
+	// pressSequentially dispatches individual keydown/input/keyup events, which is
+	// necessary for the @ trigger detection in useReferenceAutocomplete.
+	await textarea.pressSequentially(text, { delay: 30 });
+}
+
+/**
+ * Wait for the reference autocomplete dropdown to appear.
+ *
+ * The dropdown is rendered as a `role="listbox"` element by ReferenceAutocomplete.
+ */
+export async function waitForReferenceAutocomplete(page: Page): Promise<Locator> {
+	const dropdown = page.locator(AUTOCOMPLETE_SELECTOR).first();
+	await dropdown.waitFor({ state: 'visible', timeout: 5000 });
+	return dropdown;
+}
+
+/**
+ * Get all reference autocomplete items currently visible in the dropdown.
+ *
+ * Returns a Locator pointing to all `role="option"` elements inside the listbox.
+ */
+export function getReferenceAutocompleteItems(page: Page): Locator {
+	return page.locator(`${AUTOCOMPLETE_SELECTOR} ${AUTOCOMPLETE_ITEM_SELECTOR}`);
+}
+
+/**
+ * Navigate to an autocomplete item by index using keyboard and select it with Enter.
+ *
+ * Index is 0-based and maps to the global order of items across all groups.
+ * Pressing ArrowDown `index + 1` times from the initial position selects the
+ * desired item (the hook starts at index -1 / no selection before the first key).
+ *
+ * @param page - Playwright page
+ * @param index - 0-based index of the item to select
+ */
+export async function selectReferenceByIndex(page: Page, index: number): Promise<void> {
+	const textarea = page.locator(CHAT_INPUT_SELECTOR).first();
+	// Press ArrowDown (index + 1) times to reach the target item
+	for (let i = 0; i <= index; i++) {
+		await textarea.press('ArrowDown');
+	}
+	await textarea.press('Enter');
+}
+
+/**
+ * Click on a specific autocomplete item whose text contains `searchText`.
+ *
+ * @param page - Playwright page
+ * @param searchText - Substring of the item's display text to match
+ */
+export async function selectReferenceByClick(page: Page, searchText: string): Promise<void> {
+	const item = page
+		.locator(`${AUTOCOMPLETE_SELECTOR} ${AUTOCOMPLETE_ITEM_SELECTOR}`)
+		.filter({ hasText: searchText })
+		.first();
+	await item.waitFor({ state: 'visible', timeout: 5000 });
+	await item.click();
+}
+
+// ─── Mention Token Helpers ────────────────────────────────────────────────────
+
+/**
+ * Build a CSS/text selector for a `@ref{type:id}` token inside the textarea value.
+ *
+ * Because the textarea is a plain `<textarea>` element, "tokens" are not rendered
+ * as DOM nodes — the raw `@ref{type:id}` syntax is stored in the textarea value.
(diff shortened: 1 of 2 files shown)


lsm (commented):

---

### P0 (Blocking)

**1. `selectReferenceByIndex` has an off-by-one navigation bug**
`packages/e2e/tests/helpers/reference-helpers.ts`, lines 68-75

The hook initializes `selectedIndex` to `0` when autocomplete opens (see `useReferenceAutocomplete.ts` line 106). This means index 0 is already selected before any ArrowDown is pressed. To reach index N, you need exactly N ArrowDown presses -- not `N + 1`.

Current code:
```ts
for (let i = 0; i <= index; i++) {
    await textarea.press('ArrowDown');
}
```

This presses ArrowDown `index + 1` times, which actually selects index `index + 1`. For example, calling `selectReferenceByIndex(page, 0)` would select item 1, not item 0. The fix is `i < index` (N presses to reach index N).

The JSDoc comment at line 62 also incorrectly states "the hook starts at index -1 / no selection before the first key" -- it starts at 0.

---

### P1 (Should-fix)

**2. Misleading JSDoc comment in `typeInChatInput`**
`packages/e2e/tests/helpers/reference-helpers.ts`, line 27

The comment says "Uses `fill` for plain text and `pressSequentially` for sequences that include the `@` trigger character" but the function only uses `pressSequentially`. There is no conditional `fill` path. The comment should be corrected to match reality.

**3. `cleanupAllCreatedEntities` lacks task/goal-specific cleanup**
`packages/e2e/tests/helpers/room-helpers.ts`, lines 214-231

The `CreatedEntities` interface tracks `roomIds` and `filePaths` but not individual task or goal IDs. This is fine when tasks/goals are always created inside a room (since room.delete cascades), but the helpers expose `createTask` and `createGoal` as standalone exports that can be called with a pre-existing room. If a test creates tasks/goals in an existing room and only tracks those IDs in `CreatedEntities`, the cleanup won't remove them.

Either add `taskIds`/`goalIds` to `CreatedEntities` with corresponding delete RPCs, or document explicitly that this interface is only for rooms created by the composite helpers (not for standalone task/goal creation).

**4. `createRoomWithTaskAndGoal` ignores `goalTitle` in room name**
`packages/e2e/tests/helpers/room-helpers.ts`, line 152

```ts
const roomId = await createRoom(page, `${taskTitle} Room`);
```

When `taskTitle` and `goalTitle` differ, the room name only reflects the task title. Consider using a combined name like `${taskTitle}-${goalTitle} Room` or accepting an explicit room name parameter. This is minor but could cause confusion when debugging test failures.

---

### P2 (Important suggestion)

**5. `hoverMentionToken` is essentially a no-op**
`packages/e2e/tests/helpers/reference-helpers.ts`, lines 157-161

The function verifies the token exists, then hovers over the entire `<textarea>` element rather than any specific token position. As the JSDoc itself acknowledges, this doesn't test hover behavior on the token. While the comment mentions a future MentionToken component, shipping a helper that appears to do something useful but actually just hovers the textarea could mislead test authors into thinking they're testing token-specific hover behavior. Consider either removing this helper until it can be properly implemented, or adding a more prominent warning in its documentation.

**6. No `deleteGoal` or `deleteTask` helper**
`packages/e2e/tests/helpers/room-helpers.ts`

The `livequery-task-goal-updates.e2e.ts` test already defines inline `deleteGoal` and `createRoom`/`createTask`/`createGoal` helpers that duplicate the logic in this file. Since the task was to create reusable helpers that follow existing patterns, it would be natural to also export `deleteGoal` here so that test can be refactored to use shared helpers.

---

### P3 (Nit)

**7. Composite room helpers could accept an optional `roomName` parameter**
Files: `packages/e2e/tests/helpers/room-helpers.ts`, lines 117-156

All three composite helpers (`createRoomWithTask`, `createRoomWithGoal`, `createRoomWithTaskAndGoal`) auto-generate room names. Adding an optional `roomName` parameter (defaulting to the current convention) would give callers more flexibility without breaking existing usage.

---

### Summary

The room-helpers are well-structured and follow existing patterns correctly. The reference-helpers have solid selectors matching the actual component (verified against `ReferenceAutocomplete.tsx`), and the `@ref{type:id}` token format matches the backend's `REFERENCE_PATTERN` and `insertReferenceMention` logic. The cleanup approach using `Promise.allSettled` with best-effort error handling is appropriate. However, the off-by-one bug in `selectReferenceByIndex` is a correctness issue that will cause test authors to select the wrong autocomplete item.

lsm (commented):

---

All previously reported issues have been addressed. The P0 off-by-one fix is correct, JSDoc improvements are accurate, and the `CreatedEntities` interface now properly supports non-cascade cleanup. One new issue found during re-verification.

### Previously reported issues — all resolved

**P0 — `selectReferenceByIndex` off-by-one (FIXED)**
Verified correct. The loop `for (let i = 0; i < index; i++)` runs exactly `index` times. Confirmed against `useReferenceAutocomplete.ts` (line 106): `selectedIndex` initializes to `0` and resets to `0` on new results (line 167). ArrowDown handler (line 213) increments by 1. So 0 presses = index 0, N presses = index N. The fix at `reference-helpers.ts:71` is correct.

**P1 — `typeInChatInput` misleading JSDoc (FIXED)**
The JSDoc at `reference-helpers.ts:24-28` now accurately describes the `pressSequentially` behavior without any misleading fallback mention.

**P1 — `CreatedEntities` incomplete interface (FIXED)**
The interface at `room-helpers.ts:270-275` now includes `tasks` and `goals` fields with clear documentation explaining the cascade vs. standalone distinction.

**P1 — Room naming inconsistency (FIXED)**
All three composite helpers (`createRoomWithTask`, `createRoomWithGoal`, `createRoomWithTaskAndGoal`) now accept an optional `roomName` parameter with documented defaults at `room-helpers.ts:159-163,176-177,193-194`.

**P2 — `hoverMentionToken` no-op without documentation (FIXED)**
The NOTE at `reference-helpers.ts:149-153` clearly explains the textarea-level limitation and references the MentionToken component (M4) as the future solution.

**P2 — Missing `deleteTask`/`deleteGoal` exports (FIXED)**
Both are exported at `room-helpers.ts:117` and `room-helpers.ts:137` respectively.

### New issue found during re-verification

**[P1] `deleteTask` calls non-existent `task.delete` RPC handler**
`packages/e2e/tests/helpers/room-helpers.ts:124` — The `deleteTask` helper calls `hub.request('task.delete', ...)`, but there is no `task.delete` RPC handler registered in the daemon. I verified this by searching `packages/daemon/src/lib/rpc-handlers/task-handlers.ts` — the available task RPC methods are `task.create`, `task.list`, `task.get`, `task.updateDraft`, `task.fail`, `task.cancel`, `task.archive`, `task.setStatus`, `task.reject`, `task.getGroup`, `task.getGroupMessages`, `task.group.addMessage`, `task.group.create`, `task.sendHumanMessage`, `task.interruptSession`, and `task.approve`. There is no `task.delete`.

The try/catch at line 128 silently swallows the resulting error, so the helper will always appear to succeed without actually deleting anything. This is misleading for test authors who may rely on `deleteTask` for cleanup of standalone tasks in existing rooms (the exact use case the `CreatedEntities.tasks` field was designed for).

**Recommendation:** Either (a) replace the RPC call with the closest available operation (`task.cancel` or `task.archive`) and rename the helper accordingly, or (b) add a TODO comment and a `console.warn` inside the catch block so the failure is at least visible during test runs, or (c) implement a `task.delete` RPC handler in the daemon.

---

### Summary

All original review findings have been properly addressed. The one remaining issue is that `deleteTask` is a broken helper due to a missing backend RPC handler. I recommend addressing this before merging to avoid misleading future test authors.

**Recommendation: REQUEST_CHANGES** (for the new P1 `deleteTask` issue)

lsm (commented):

**Recommendation: APPROVE**

The fix for the P1 issue (deleteTask calling non-existent task.delete RPC) has been correctly and completely implemented.

### Verification

1. **deleteTask removed**: No trace of a `deleteTask` export or function exists in the codebase (confirmed via grep across all packages). The fix commit (4f9529c) removes 29 lines and adds 7, net -22.

2. **tasks field removed from CreatedEntities**: The interface at `packages/e2e/tests/helpers/room-helpers.ts:252-256` now contains only `roomIds`, `goals`, and `filePaths`. No `tasks` field.

3. **JSDoc documents cascade behavior**: The `CreatedEntities` doc (line 245-247) explicitly states: "Tasks have no standalone delete RPC (`task.delete` does not exist), so they must always be cleaned up by deleting their parent room." This is factually correct -- confirmed via:
   - `packages/daemon/src/lib/rpc-handlers/task-handlers.ts` registers no `task.delete` handler
   - `packages/daemon/src/storage/schema/index.ts:175` defines `FOREIGN KEY (room_id) REFERENCES rooms(id) ON DELETE CASCADE`
   - `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:240` confirms CASCADE deletion

4. **All RPC calls verified correct**: `task.create`, `goal.create`, `goal.delete`, and `room.delete` all use parameter shapes matching their respective handlers.

No remaining issues found. The implementation is clean and the documentation accurately describes the cleanup semantics.

PR #852, opened 2026-03-25
feat(web): integrate @ref token rendering in SDKUserMessage

lsm:
- Create MentionToken component with type-specific pill styling
  (indigo=task, amber=goal, blue=file, yellow=folder)
- Lazy-load hover popover via reference.resolve RPC; result cached
  after first fetch to avoid re-fetching on subsequent hovers
- parseTextWithReferences() splits text into text/mention/unknown
  segments; safe against shared REFERENCE_PATTERN lastIndex state
- SDKUserMessage renders @ref{type:id} tokens inline; plain @ text
  and unknown types pass through unchanged (warning styling for unknowns)
- Missing referenceMetadata falls back to raw id as display text
- MentionToken wrapped in memo() to prevent re-renders during scrolling
- 77 unit tests: parsing logic, all ref types, hover states, RPC calls,
  fallbacks, and SDKUserMessage integration

### DIFF
--- a/packages/web/src/components/sdk/MentionToken.tsx
+++ b/packages/web/src/components/sdk/MentionToken.tsx
@@ -10,7 +10,7 @@
 
 import type { JSX } from 'preact';
 import { memo } from 'preact/compat';
-import { useState, useCallback } from 'preact/hooks';
+import { useState, useCallback, useRef } from 'preact/hooks';
 import { cn } from '../../lib/utils.ts';
 import { useMessageHub } from '../../hooks/useMessageHub.ts';
 import type { ReferenceType, ReferenceMetadata, ResolvedReference } from '@neokai/shared';
@@ -41,18 +41,8 @@ const TYPE_STYLES: Record<ReferenceType, { pill: string; label: string }> = {
 
 function renderResolvedContent(resolved: ResolvedReference): JSX.Element {
 	switch (resolved.type) {
-		case 'task': {
-			const d = resolved.data as { title?: string; status?: string; description?: string };
-			return (
-				<div>
-					{d.title && <div class="font-medium text-white">{d.title}</div>}
-					{d.status && <div class="text-xs text-gray-400 mt-0.5">Status: {d.status}</div>}
-					{d.description && (
-						<div class="text-xs text-gray-400 mt-1 line-clamp-2">{d.description}</div>
-					)}
-				</div>
-			);
-		}
+		// task and goal share the same shape
+		case 'task':
 		case 'goal': {
 			const d = resolved.data as { title?: string; status?: string; description?: string };
 			return (
@@ -91,6 +81,9 @@ function renderResolvedContent(resolved: ResolvedReference): JSX.Element {
 				</div>
 			);
 		}
+		default: {
+			return <div class="text-xs text-gray-400">Unknown reference type</div>;
+		}
 	}
 }
 
@@ -107,6 +100,14 @@ export interface MentionTokenProps {
 
 type LoadState = 'idle' | 'loading' | 'loaded' | 'error';
 
+/** Popover position computed from getBoundingClientRect, used for fixed positioning. */
+interface PopoverPos {
+	top: number;
+	left: number;
+	/** True when the popover is rendered below the token (not enough space above). */
+	below: boolean;
+}
+
 function MentionTokenBase({
 	refType,
 	id,
@@ -117,12 +118,23 @@ function MentionTokenBase({
 	const [isHovered, setIsHovered] = useState(false);
 	const [loadState, setLoadState] = useState<LoadState>('idle');
 	const [resolvedData, setResolvedData] = useState<ResolvedReference | null>(null);
+	const [popoverPos, setPopoverPos] = useState<PopoverPos>({ top: 0, left: 0, below: false });
+	const tokenRef = useRef<HTMLSpanElement>(null);
 	const { callIfConnected } = useMessageHub();
 
 	const typeStyle = TYPE_STYLES[refType];
 
 	const handleMouseEnter = useCallback(async () => {
+		// Compute fixed viewport position before showing — escapes scroll containers
+		// and overflow:hidden ancestors that would clip an absolute popover.
+		if (tokenRef.current) {
+			const rect = tokenRef.current.getBoundingClientRect();
+			// Flip below the token when there is insufficient space above (< 160 px)
+			const below = rect.top < 160;
+			setPopoverPos({ top: below ? rect.bottom : rect.top, left: rect.left, below });
+		}
 		setIsHovered(true);
+
 		// Only fetch once; skip if already fetched or no sessionId to call with
 		if (loadState === 'idle' && sessionId) {
 			setLoadState('loading');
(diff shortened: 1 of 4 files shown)


lsm (commented):

---

### Verdict: REQUEST_CHANGES (own PR, posting as COMMENT)

---

### P1 -- Missing default branch in `renderResolvedContent` switch

**File:** `packages/web/src/components/sdk/MentionToken.tsx`, lines 42-94

The `renderResolvedContent` switch covers all 4 known `ReferenceType` values but has no default branch and no return after the switch. The function declares a return type of `JSX.Element`, so if `resolved.type` is an unexpected value at runtime (e.g., a new type added server-side before the client is updated), the function returns `undefined`. This would cause a silent rendering failure or a Preact error.

Add a default case:
```ts
default: {
    return <div class="text-xs text-gray-400">Unknown reference type</div>;
}
```

This is a real risk because the data comes from an RPC call and the `data` field is typed as `unknown` in `ResolvedReference` -- the server schema could evolve independently of the client.

---

### P1 -- Popover may be clipped by scroll containers

**File:** `packages/web/src/components/sdk/MentionToken.tsx`, lines 169-188

The popover uses `position: absolute; bottom: full;` (i.e., `bottom: 100%`) which positions it above the token. However, messages are rendered inside a scrollable chat container. If the token is near the top of the viewport, the popover will render above the visible area. Additionally, if any ancestor element has `overflow: hidden` (common in chat containers to hide scrollbar artifacts), the popover will be clipped.

Consider one of:
- Using a portal (e.g., rendering into `document.body`) with calculated positioning
- Adding a check for viewport position and flipping the popover to appear below when near the top
- At minimum, adding `overflow: visible` to the token's parent span context

This is P1 rather than P0 because it degrades gracefully (popover simply won't show) and the data is still visible in the pill.

---

### P2 -- Text segments wrapped in `<span>` changes DOM structure

**File:** `packages/web/src/components/sdk/SDKUserMessage.tsx`, line 46

The original code rendered `textContent` as a raw text child of the div. The new code wraps each text segment in `<span key={idx}>{seg.content}</span>`. While this is semantically equivalent for display purposes, it changes the DOM structure from text nodes to inline span elements. This could affect:
- Browser text selection behavior (selecting across span boundaries may behave differently)
- Any future code that queries text content of the message div

A simpler approach that preserves text nodes would be to use Preact's text node output directly (e.g., returning the string in the array, which Preact handles as text nodes). However, since Preact fragments `<>` require keyed children, the current approach is acceptable as long as the team is aware of the tradeoff.

---

### P2 -- Duplicate `render()` calls in tests

**File:** `packages/web/src/components/sdk/__tests__/MentionToken.test.tsx`, lines 176-179 and 191-194

Two tests have redundant double `render()` calls:
```tsx
// Line 176-179 ("calls reference.resolve RPC on first hover")
render(<MentionToken refType="task" id="t-42" ... />);
const { container } = render(<MentionToken refType="task" id="t-42" ... />);

// Line 191-194 ("does not call RPC when sessionId is absent")
render(<MentionToken refType="task" id="t-1" ... />);
const { container } = render(<MentionToken refType="task" id="t-1" ... />);
```

The first `render()` in each pair creates a component that is never queried or cleaned up. This is likely a leftover from an earlier iteration. Remove the orphaned first call in both cases.

---

### P3 -- `VALID_REF_TYPES` duplicates `ReferenceType` union

**File:** `packages/web/src/components/sdk/MentionToken.tsx`, line 208

`VALID_REF_TYPES` is a `Set` of strings that must stay in sync with the `ReferenceType` union type. If a new reference type is added to the union but not to this set, parsing would silently downgrade known types to unknown-mention. Consider deriving the set from the type:
```ts
const VALID_REF_TYPES: ReadonlySet<string> = new Set<string>([
    'task', 'goal', 'file', 'folder',
] satisfies readonly ReferenceType[]);
```

---

### P3 -- Identical task/goal rendering in `renderResolvedContent`

**File:** `packages/web/src/components/sdk/MentionToken.tsx`, lines 44-66

The `task` and `goal` cases are identical. Consider combining them:
```ts
case 'task':
case 'goal': {
    const d = resolved.data as { title?: string; status?: string; description?: string };
    return (...);
}
```

---

### Positive observations

- Regex cloning (`new RegExp(REFERENCE_PATTERN.source, REFERENCE_PATTERN.flags)`) correctly avoids shared `lastIndex` state -- well done.
- Fast path (`!text.includes('@ref{')`) avoids unnecessary parsing for the vast majority of messages.
- `memo()` wrapping on MentionToken is appropriate for scroll performance.
- `handleCopy` correctly copies the raw `textContent` string (with `@ref{...}` tokens), not the rendered DOM text.
- Test coverage is thorough: plain text, all 4 ref types, edge cases (adjacent refs, start/end of text), unknown types, idempotency, hover lifecycle (load/cache/error/idle states), and integration with SDKUserMessage.
- All 460 tests pass including the 40 new ones.
- No XSS risk: text content is rendered via Preact's JSX (auto-escaped), and the RPC data is rendered through typed components, not `innerHTML`.
- `callIfConnected` correctly handles the disconnected case (returns immediately without throwing).
- The `referenceMetadata` extraction via type assertion (`message as ... & { referenceMetadata?: ... }`) is a pragmatic approach given the message schema is being extended incrementally.

lsm (commented):

### Re-review Summary

All five previously flagged issues (P1, P2, P3) have been properly addressed in the fix commit (9d4a56f). The implementation is now solid.

### Verified Fixes

**P1 - Default case in renderResolvedContent switch** -- RESOLVED
A default case returning "Unknown reference type" is now present at packages/web/src/components/sdk/MentionToken.tsx:83-85. This prevents runtime crashes from unexpected server-side reference types.

**P1 - Popover positioning clipped by scroll containers** -- RESOLVED
The popover now uses position: fixed with getBoundingClientRect() for viewport-relative positioning (MentionToken.tsx:178-187). A PopoverPos state tracks computed coordinates, and a flip heuristic (rect.top < 160) renders the popover below the token when there is insufficient space above. The zIndex: 9999 ensures it renders above most content. The relative class was correctly removed from the outer span.

**P2 - Unnecessary span-wrapping of text segments** -- RESOLVED
SDKUserMessage.tsx:47 now uses Fragment instead of span for text segments, preserving native text nodes for selection and layout.

**P2 - Redundant render() calls in tests** -- RESOLVED
The two orphaned render() calls in MentionToken.test.tsx (formerly around lines 278, 293) have been removed.

**P3 - VALID_REF_TYPES type safety** -- RESOLVED
VALID_REF_TYPES at MentionToken.tsx:224-229 now uses satisfies readonly ReferenceType[] to catch union/set drift at compile time.

**P3 - Merged identical task/goal switch cases** -- RESOLVED
The task case at MentionToken.tsx:44 now falls through to the shared goal case with a comment.

- P3: The popover position is computed only on mouseEnter, so if the user scrolls while the popover is visible, the popover will stay at its original viewport position and visually drift away from the token. This is a minor UX concern for long conversations with active scrolling. A scroll listener or onMouseMove re-computation would fix this, but it is not blocking for this PR.

### Overall Assessment

Clean implementation with good test coverage (77 tests for parsing, rendering, hover states, RPC calls, and fallbacks). The fix commit is targeted and correctly resolves all previously identified issues. No new concerns introduced.

**Recommendation: APPROVE**

PR #856, opened 2026-03-25
test(e2e): add edge cases, mobile, and accessibility tests for reference autocomplete

lsm:
Covers Task 5.4 of the system-wide @ referencing system:

Edge cases:
- Empty search query returns no dropdown
- Rapid typing shows single dropdown without duplicates
- Reference autocomplete works alongside slash command text

Token error states:
- Non-existent task/file token shows "Not found" on hover
- Deleted task token shows "Not found" after room deletion

Input scenarios:
- 5+ references in one message all render as mention tokens
- Pasting raw @ref{} syntax renders as styled MentionToken when sent

Mobile (iPhone 13 emulation):
- Autocomplete appears on mobile viewport
- Touch tap selects a reference (tap not click)
- Dropdown fits within mobile viewport bounds

Accessibility:
- ArrowDown + Enter keyboard navigation selects a reference
- Escape closes dropdown while preserving @ query text
- ARIA listbox/option roles present with aria-selected
- Sent token has data-ref-type/data-ref-id attributes for a11y
- Focus stays in textarea after keyboard selection

### DIFF
--- a/packages/e2e/tests/features/reference-autocomplete.e2e.ts
+++ b/packages/e2e/tests/features/reference-autocomplete.e2e.ts
@@ -1,30 +1,39 @@
 /**
- * Reference Autocomplete E2E Tests (Task 5.3)
+ * Reference Autocomplete E2E Tests (Tasks 5.3 + 5.4)
  *
- * Tests reference resolution and message rendering:
+ * Task 5.3 — Resolution and rendering:
  * - Selecting task/goal/file reference inserts @ref{type:id} into the input
  * - Message with references sends and renders as MentionToken pills
  * - Hover on token shows entity details popover
  * - Multiple references in a single message all render correctly
  * - Standalone sessions (no room) show only file/folder results
  *
+ * Task 5.4 — Edge cases, mobile, accessibility:
+ * - Empty search, rapid typing, deleted entities, many references,
+ *   copy/paste @ref text, reference combined with slash command
+ * - Mobile: autocomplete on small viewport, touch selection, dropdown bounds
+ * - Accessibility: keyboard navigation, ARIA roles, focus management
+ *
  * Setup: rooms/tasks/goals created via RPC (infrastructure exemption).
- * Standalone sessions use createSessionViaUI (established infrastructure helper).
  * All test actions and assertions go through the browser UI.
  * Cleanup: rooms and sessions deleted via RPC in afterEach.
  */
 
-import { test, expect } from '../../fixtures';
+import { test, expect, devices } from '../../fixtures';
+import type { Page } from '@playwright/test';
 import {
 	waitForWebSocketConnected,
 	createSessionViaUI,
 	cleanupTestSession,
+	waitForMessageSent,
 } from '../helpers/wait-helpers';
 import {
+	createRoom,
 	createRoomWithTask,
 	createRoomWithGoal,
 	createRoomWithTaskAndGoal,
 	deleteRoom,
+	createTask,
 } from '../helpers/room-helpers';
 import {
 	typeInChatInput,
@@ -37,10 +46,7 @@ import {
 // ─── Shared Infrastructure ────────────────────────────────────────────────────
 
 /** Navigate to a room's agent chat and wait for the textarea to be ready. */
-async function navigateToRoomChat(
-	page: Parameters<typeof waitForWebSocketConnected>[0],
-	roomId: string
-): Promise<void> {
+async function navigateToRoomChat(page: Page, roomId: string): Promise<void> {
 	await page.goto(`/room/${roomId}/agent`);
 	await waitForWebSocketConnected(page);
 	const textarea = page.locator('textarea[placeholder*="Ask"]').first();
@@ -49,7 +55,7 @@ async function navigateToRoomChat(
 }
 
 /** Get the chat input textarea. */
-function getChatInput(page: Parameters<typeof waitForWebSocketConnected>[0]) {
+function getChatInput(page: Page) {
 	return page.locator('textarea[placeholder*="Ask"]').first();
 }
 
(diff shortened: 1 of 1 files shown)


lsm (commented):

---

### Summary

The test file is well-structured and follows the existing E2E conventions closely. The imports use the shared helpers correctly, the test groups are logically organized, and the selectors match the actual component implementations. However, I found several issues that should be addressed before merging.

---

### P0 (Blocking)

**1. Escape key test is effectively a no-op**
 lines 453-472

The test types `@test` using `typeInChatInput` (which uses `pressSequentially` with 30ms delay per character -- ~150ms total), then immediately checks `isVisible()`. The autocomplete has a 300ms debounce plus an async RPC round-trip, so the dropdown will not be visible yet when the check runs. The early return on line 462 means this test always skips its actual assertions.

Fix: add a wait for the autocomplete to appear before checking visibility, or use `waitForReferenceAutocomplete(page)` first, then press Escape. For example:
```ts
await typeInChatInput(page, '@test');
await waitForReferenceAutocomplete(page);
// Now press Escape and verify...
```
Alternatively, if you want to test the case where the dropdown IS open and Escape closes it, ensure the query actually matches something (e.g., `@Accessibility` would match the "Accessibility Test Task" created in `beforeEach`).

---

### P1 (Should Fix)

**2. "Deleted task after room deletion" test is fragile and may break silently**
 lines 210-248

The test deletes the room via `deleteRoom(page, roomId)` at line 236 and then tries to hover a MentionToken in the chat 300ms later. While the current code does NOT emit `session.deleted` WebSocket events during `room-manager.deleteRoom()` (it only does raw SQL deletes through the repository), this behavior is an implementation detail that could change. If a future change adds proper event emission to the room deletion cascade, the `room-store.ts` `session.deleted` handler (line 641-643) would navigate the page away, making the token invisible.

The 300ms `waitForTimeout` at line 240 is also concerning -- it's a fixed sleep that provides no correctness guarantee. If the server takes longer to process, the hover resolve might still succeed with stale data.

Recommendation: Instead of deleting the entire room (which has broad side effects), test with a standalone session and reference a task that was never created. The existing "non-existent task" test at line 160 already covers this scenario effectively. If you specifically need to test post-deletion behavior, consider documenting the assumption that room deletion does not trigger page navigation.

**3. Mobile viewport boundingBox test may silently pass without asserting**
 lines 397-406

If `dropdown.boundingBox()` returns `null` (which Playwright does for elements with zero dimensions or when the element is off-screen), the `if (box && viewportSize)` guard at line 401 causes the test to pass without any assertion. This silently skips the entire purpose of the test.

Fix: assert that `box` is not null before the bounds checks:
```ts
expect(box).not.toBeNull();
```

**4. Unused `sessionId` variable in Group 1**
 line 68

`let sessionId: string | null = null;` is declared but never assigned in Group 1. The `afterEach` cleanup for `sessionId` is dead code. This should be removed to avoid confusion.

---

### P2 (Important Suggestions)

**5. Three `waitForTimeout` calls across the file**
Lines 95, 126, and 240 use fixed timeouts (800ms, 200ms, 300ms). The 800ms wait at line 95 is marginally justified (debounce + network), but the 200ms at line 126 (between Escape and typing `@`) and 300ms at line 240 (post-deletion) are fragile. Consider using condition-based waits instead.

**6. ARIA test does not verify actual keyboard focus on options**
 lines 430-450

The keyboard navigation test (lines 430-450) presses ArrowDown then ArrowUp (ending back at index 0), then Enter. This tests that Enter selects from the default position, but never actually tests that ArrowDown changes the selection to a different item. Consider pressing ArrowDown once and verifying that the second item is highlighted (e.g., checking `aria-selected` moves to the second option) before pressing Enter.

**7. "Copy/paste" test does not actually test paste behavior**
 lines 296-316

The test comment says "Simulate copy/paste" but uses `textarea.fill()`, which does not trigger the browser's paste event. If the component ever adds paste-specific handling (e.g., intercepting paste to auto-detect \@ref{} syntax), this test would miss it. Consider using `page.keyboard.insertText()` or `page.clipboard` APIs to more accurately simulate a paste operation.

---

### P3 (Nits)

**8. The `navigateToRoomAgent` helper is a local duplication**
The local `navigateToRoomAgent` function (lines 43-49) is similar to patterns used in other E2E tests. If other test files also navigate to room agent pages, consider extracting this to a shared helper in `reference-helpers.ts` or `room-helpers.ts`.

**9. Minor: Group 2 Token Error States does not set `sessionId` in tests 1 and 2**
In the "non-existent task" and "non-existent file" tests (lines 160, 186), the `sessionId` is assigned from `createSessionViaUI`, which is correct. But the variable declaration is at the group level while the assignment is per-test. This is fine but slightly misleading since the group-level `afterEach` handles both cases. No action needed.

---

### Verdict

The test coverage is solid overall and addresses the requested areas. The P0 issue with the Escape test is the most critical -- it gives false confidence that Escape dismisses the autocomplete. The P1 issues around fragility and silent passes should also be addressed. The remaining tests are well-structured and should be reliable.

lsm (commented):

Re-reviewed after the fix commit (c32ac2988). All previously identified issues have been addressed. Detailed findings below.

---

### Previously reported issues — verification

**P0: Escape key test was a no-op** — FIXED
The test now types `@Accessibility` (matching the task created in `beforeEach`), calls `waitForReferenceAutocomplete(page)` to guarantee the dropdown is visible before pressing Escape, and then asserts it hides. The race condition is eliminated. (Line 470-487)

**P1: Unused `sessionId` variable in Group 1** — FIXED
The `let sessionId: string | null = null` declaration and the dead `afterEach` cleanup branch have been removed from the Autocomplete Behavior group. `sessionId` remains only in Group 2 (Token Error States) where it is actually used by `createSessionViaUI`. (Line 70-87)

**P1: Fragile deleted-task test** — IMPROVED
- `waitForTimeout(300)` replaced with `expect(token).toBeVisible({ timeout: 3000 })` which both waits and asserts the page hasn't navigated away. (Line 240)
- A comment documents the implementation assumption about room deletion not triggering navigation. One minor note: the comment says "room deletion currently does NOT emit a session.deleted WebSocket event" — the handler actually emits `room.deleted` (not `session.deleted`), and the frontend only filters the room from the lobby list without navigating. The comment is technically correct but could be more precise. This is a P3 nit.

**P1: Mobile boundingBox silent pass** — FIXED
Explicit `expect(box).not.toBeNull()` and `expect(viewportSize).not.toBeNull()` assertions added before the conditional bounds checks. If the element has no layout rect, the test now fails with a clear message. (Lines 403-404)

**P2: `waitForTimeout` calls** — FIXED
All `waitForTimeout` calls have been removed from the test file:
- Empty search query: replaced with `toBeHidden({ timeout: 1500 })` (Line 96)
- Slash command dismissal: replaced with `toBeHidden({ timeout: 2000 })` (Line 125)
- Deleted-task wait: replaced with `toBeVisible({ timeout: 3000 })` (Line 240)

**P2: Incomplete keyboard navigation** — FIXED
The test now:
1. Asserts the first option starts with `aria-selected="true"` (Line 446)
2. Presses ArrowDown
3. Asserts the second option has `aria-selected="true"` AND the first has `aria-selected="false"`, proving the selection actually moved (Lines 453-454)
4. Only proceeds if there are 2+ options (`if (optionCount >= 2)` guard on line 452)

**P2: Copy/paste not simulating paste** — FIXED
`fillAndSend` (using `textarea.fill()`) replaced with `insertAndSend` (using `page.keyboard.insertText()`). This dispatches a real `input` event matching the browser paste code path. The helper also calls `textarea.focus()` first. All call sites updated consistently. (Lines 56-66, and all usages)

---

### New observations

**P3: Deleted-task comment precision** (Line 233-234)
The comment states "room deletion currently does NOT emit a session.deleted WebSocket event (room-manager.deleteRoom() performs raw SQL deletes without broadcasting)." The RPC handler at `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:231-232` actually DOES emit `daemonHub.emit('room.deleted', ...)` before calling `roomManager.deleteRoom()`. The reason the page stays is that the frontend handler in `lobby-store.ts:183-185` only removes the room from the lobby list without navigating. The comment could be more precise to avoid confusion for future maintainers.

**P3: `insertAndSend` does not use `typeInChatInput`** (Line 56-66)
The new `insertAndSend` helper duplicates the textarea locator and focus logic that exists in `typeInChatInput`. This is fine because the two helpers serve different purposes (`insertText` for bulk content vs `pressSequentially` for character-by-character key events), but worth noting for maintainability.

---

### Summary

All P0, P1, and P2 issues from the previous review have been properly fixed. The fixes are well-targeted and do not introduce new problems. Only P3 nits remain (imprecise comment, minor code duplication). The test file is clean, well-documented, and follows project conventions.

Recommendation: **APPROVE**

PR #858, opened 2026-03-25
feat(room): integrate AppMcpLifecycleManager into RoomRuntimeService (Task 3.2)

lsm:
## Summary

- Injects `AppMcpLifecycleManager` (optional) into `RoomRuntimeServiceConfig` as `appMcpManager`
- Merges three MCP sources in `setupRoomAgentSession`: file-based (`SettingsManager.getEnabledMcpServersConfig()`), registry-based (`appMcpManager.getEnabledMcpConfigs()`), then `room-agent-tools` last (always wins on collision)
- Caches the per-room `room-agent-tools` McpServerConfig in a `roomAgentMcpServers` map for hot-reload support
- Subscribes to `mcp.registry.changed` in `subscribeToEvents()` — re-applies merged MCP configs to all live room chat sessions; documents that `setRuntimeMcpServers()` is between-query only and does not disrupt in-flight queries
- Passes `appMcpManager` from the `RoomRuntimeService` construction site in `rpc-handlers/index.ts`
- Cleans up `roomAgentMcpServers` entries on `stopRuntime()` and `stop()`

## Test plan

- [x] 8 new unit tests in `room-runtime-service-mcp.test.ts` covering:
  - Registry servers appear in the merged map
  - `room-agent-tools` takes precedence over any registry entry with the same name
  - File-based and registry servers are both present in the merged map
  - Works correctly without `appMcpManager` (optional field, fallback to `{}`)
  - `mcp.registry.changed` triggers `setRuntimeMcpServers()` on all live sessions
  - Missing `room-agent-tools` cache entry doesn't prevent other servers from applying
  - No error thrown when session is not found during hot-reload
  - Registry-sourced server wins over file-based server on name collision
- [x] Existing wiring tests (26 tests) still pass
- [x] Lint clean (0 warnings/errors)

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -183,6 +183,9 @@ export class RoomRuntimeService {
 			}
 			this.runtimes.delete(roomId);
 			this.observers.delete(roomId);
+			// Clear the cached room-agent-tools server; createOrGetRuntime() →
+			// setupRoomAgentSession() will repopulate it for the fresh runtime.
+			this.roomAgentMcpServers.delete(roomId);
 		}
 
 		// Create a fresh runtime - autoStart=true starts it immediately
@@ -643,9 +646,13 @@ export class RoomRuntimeService {
 		const unsubMcpChanged = this.ctx.daemonHub.on(
 			'mcp.registry.changed',
 			() => {
-				const fileMcpServers = this.ctx.settingsManager.getEnabledMcpServersConfig();
+				// Re-read both sources inside the handler (not hoisted above the loop) so
+				// that if getEnabledMcpServersConfig() ever becomes room-scoped, it will
+				// be called per-room consistently with the initial setupRoomAgentSession path.
+				// Registry configs are global and read once — the call is cheap.
 				const registryMcpServers = this.ctx.appMcpManager?.getEnabledMcpConfigs() ?? {};
 				for (const [roomId] of this.runtimes) {
+					const fileMcpServers = this.ctx.settingsManager.getEnabledMcpServersConfig();
 					const roomChatSessionId = `room:chat:${roomId}`;
 					void this.ctx.sessionManager
 						.getSessionAsync(roomChatSessionId)
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -10,6 +10,7 @@
 
 import type { Room, McpServerConfig, RuntimeState, GlobalSettings } from '@neokai/shared';
 import type { SettingsManager } from '../../settings-manager';
+import type { AppMcpLifecycleManager } from '../../mcp/app-mcp-lifecycle-manager';
 import type { JobQueueRepository } from '../../../storage/repositories/job-queue-repository';
 import type { JobQueueProcessor } from '../../../storage/job-queue-processor';
 import type { ReactiveDatabase } from '../../../storage/reactive-database';
@@ -53,6 +54,8 @@ export interface RoomRuntimeServiceConfig {
 	getGlobalSettings: () => GlobalSettings;
 	/** Settings manager for reading project-configured MCP servers */
 	settingsManager: SettingsManager;
+	/** Application-level MCP lifecycle manager — provides registry-sourced MCP configs */
+	appMcpManager?: AppMcpLifecycleManager;
 	/** Reactive database wrapper for change event emission */
 	reactiveDb: ReactiveDatabase;
 	/**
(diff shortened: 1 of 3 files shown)


lsm (commented):

---

### Overall Assessment

The implementation is clean, well-documented, and correctly achieves the stated goal: merging file-based, registry-based, and room-agent-tools MCP server configs in RoomRuntimeService. The merge precedence order (file < registry < room-agent-tools) is consistently applied in both the initial setup and the hot-reload handler. The hot-reload via mcp.registry.changed subscription is a sound pattern that aligns with the existing event-driven architecture.

**Recommendation: APPROVE**

---

### P1 (should-fix)

**1. Hot-reload reads stale file-based MCP servers per-room**

In the mcp.registry.changed handler (line 646-647), fileMcpServers is read once and shared across ALL rooms:

    const fileMcpServers = this.ctx.settingsManager.getEnabledMcpServersConfig();
    const registryMcpServers = this.ctx.appMcpManager?.getEnabledMcpConfigs() ?? {};
    for (const [roomId] of this.runtimes) {
      // ... uses the same fileMcpServers for every room
    }

In setupRoomAgentSession (line 561), getEnabledMcpServersConfig() is called once per room. If getEnabledMcpServersConfig() is room-aware (reading per-room config files or scoped settings), then the hot-reload handler should call it inside the loop, not once outside. If it is truly global-only, the current code is fine, but this should be explicitly noted or the call should be moved inside the loop for consistency with the initial setup path.

packages/daemon/src/lib/room/runtime/room-runtime-service.ts, lines 646-647

---

### P2 (important suggestion)

**2. Unused imports in test file**

Line 11 of the test file imports mock and beforeEach from bun:test, but neither is used anywhere in the file:

    import { describe, expect, it, mock, beforeEach } from 'bun:test';

While the linter exempts test files (**/*.test.ts in .oxlintrc.json), the import of mock is also misleading since no mocking is done in these tests (all dependencies are stubbed manually). This should be cleaned up to import { describe, expect, it } from 'bun:test';.

packages/daemon/tests/unit/room/room-runtime-service-mcp.test.ts, line 11

**3. startRuntime() does not clean up roomAgentMcpServers for the old entry**

In startRuntime() (lines 174-192), when an existing runtime is removed (line 184), the method deletes from this.runtimes and this.observers but does NOT delete from this.roomAgentMcpServers. Then createOrGetRuntime() is called (line 189), which calls setupRoomAgentSession() which overwrites the cache entry (line 517). So in practice the stale entry is overwritten and there is no bug. However, this creates an inconsistency: stopRuntime() explicitly cleans up roomAgentMcpServers (line 165), but startRuntime() relies on the overwrite side effect. A reader might reasonably expect symmetric cleanup.

packages/daemon/src/lib/room/runtime/room-runtime-service.ts, line 184

---

### P3 (nit)

**4. Test helper makeDaemonHub duplicates the wiring test version**

Both room-runtime-service-mcp.test.ts (line 25) and room-runtime-service-wiring.test.ts (line 37) define their own makeDaemonHub(). The MCP version is more full-featured (tracks listeners + emit), while the wiring version is a no-op stub. This is understandable given the different needs, but a shared helper module could reduce duplication.

**5. Tests rely on setTimeout(resolve, 10) for async timing**

All tests use await new Promise((resolve) => setTimeout(resolve, 10)) to wait for the async getSessionAsync promise chain to resolve. This is a common test pattern but is inherently fragile -- on slow CI runners the 10ms might not be enough. Consider using a manual promise resolution pattern (expose the pending promise and resolve it in the mock) for deterministic tests. This is a minor observation since the tests currently pass reliably.

lsm (commented):

---

### Verification of Review Feedback Fixes

All three issues from the previous review have been correctly addressed:

**P1: fileMcpServers moved inside the per-room loop**
- `fileMcpServers` is now read inside the `for (const [roomId] of this.runtimes)` loop at line 655, while `registryMcpServers` remains hoisted above the loop (correct, since it is global).
- The comment at lines 649-652 clearly explains the rationale: if `getEnabledMcpServersConfig()` ever becomes room-scoped, the per-room call will be consistent with `setupRoomAgentSession`.

**P2: roomAgentMcpServers.delete(roomId) added in startRuntime()**
- Line 188 now calls `this.roomAgentMcpServers.delete(roomId)` in the block that removes the old runtime before creating a fresh one.
- This makes cleanup symmetric with `stopRuntime()` (line 165) and the global `stop()` (line 204), instead of relying on the overwrite side effect in `setupRoomAgentSession()`.

**P2: Unused imports removed from test file**
- Line 11 now imports only `{ describe, expect, it }` -- `mock` and `beforeEach` are correctly removed.

### New Issue Analysis

No new issues were introduced by the feedback fixes. The changes are minimal, surgical, and well-documented.

### Test Results

- All 8 new MCP integration tests pass.
- All 12 existing wiring tests pass (no regressions).

**Recommendation: APPROVE**

PR #859, opened 2026-03-25
feat(reference): implement task and goal resolvers

lsm:
## Summary

- Add task resolution (room tasks via `t-` prefix, space tasks via `st-` prefix) to `ReferenceResolver`
- Add goal resolution (`g-` prefix) to `ReferenceResolver`
- Repository dependencies injected as minimal structural interfaces (`TaskRepoLike`, `GoalRepoLike`, `SpaceTaskRepoLike`) — no SQLite imports in the resolver
- Tests use pure mock objects only — no in-memory SQLite, no DB setup

## Test plan

- [ ] `bun test packages/daemon/tests/unit/reference-resolver.test.ts` — 61 tests pass
- [ ] `make test-daemon` — full daemon suite passes (8272 pass)
- [ ] Verify no OOM: mock-only tests have negligible memory footprint

### DIFF
--- a/packages/daemon/src/lib/agent/reference-resolver.ts
+++ b/packages/daemon/src/lib/agent/reference-resolver.ts
@@ -2,8 +2,7 @@
  * ReferenceResolver — Core service for parsing and resolving @ references
  *
  * Handles extracting @ref{} mentions from text and resolving them to their
- * full entity data. Currently implements file and folder resolvers.
- * Task and goal resolvers are stubbed (see Task 3.1).
+ * full entity data. Implements file, folder, task, and goal resolvers.
  *
  * Path validation enforced for file/folder references:
  *   - Empty paths are rejected
@@ -37,7 +36,11 @@ import type {
 	ResolvedReference,
 	ResolvedFileReference,
 	ResolvedFolderReference,
+	ResolvedTaskReference,
+	ResolvedGoalReference,
 	ReferenceType,
+	NeoTask,
+	RoomGoal,
 } from '@neokai/shared';
 import { REFERENCE_PATTERN } from '@neokai/shared';
 import { Logger } from '../logger.ts';
@@ -51,9 +54,39 @@ const MAX_FILE_SIZE = 50 * 1024;
 const MAX_FOLDER_ENTRIES = 200;
 
 /**
+ * Minimal structural interface for TaskRepository — only the method(s) used here.
+ * Using a structural interface (not importing the class) avoids pulling in SQLite.
+ */
+export interface TaskRepoLike {
+	getTaskByShortId(roomId: string, shortId: string): NeoTask | null;
+}
+
+/**
+ * Minimal structural interface for GoalRepository — only the method(s) used here.
+ */
+export interface GoalRepoLike {
+	getGoalByShortId(roomId: string, shortId: string): RoomGoal | null;
+}
+
+/**
+ * Minimal structural interface for SpaceTaskRepository — only the method(s) used here.
+ */
+export interface SpaceTaskRepoLike {
+	getTask(id: string): { id: string; spaceId: string; [key: string]: unknown } | null;
+}
+
+/** Optional repository dependencies injected at construction time. */
+export interface ReferenceResolverDeps {
+	taskRepo?: TaskRepoLike;
+	goalRepo?: GoalRepoLike;
+	spaceTaskRepo?: SpaceTaskRepoLike;
+}
+
+/**
  * Context for resolving references — provides session-scoped information.
  * File/folder resolution only requires workspacePath.
- * Task/goal resolution (Task 3.1) will use roomId and spaceId.
+ * Task resolution uses roomId (room tasks) or spaceId (space tasks).
+ * Goal resolution uses roomId.
  */
 export interface ResolutionContext {
 	roomId?: string;
@@ -62,6 +95,15 @@ export interface ResolutionContext {
 }
 
 export class ReferenceResolver {
+	private readonly taskRepo?: TaskRepoLike;
+	private readonly goalRepo?: GoalRepoLike;
+	private readonly spaceTaskRepo?: SpaceTaskRepoLike;
+
+	constructor(deps: ReferenceResolverDeps = {}) {
+		this.taskRepo = deps.taskRepo;
+		this.goalRepo = deps.goalRepo;
+		this.spaceTaskRepo = deps.spaceTaskRepo;
+	}
 	/**
 	 * Extract all @ref{} mentions from a text string.
 	 * Returns deduplicated mentions in order of appearance.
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE** (with minor suggestions)

---

### Summary

The implementation is clean, well-structured, and correctly implements task/goal resolution for the agent-level `ReferenceResolver`. The use of lightweight structural interfaces (`TaskRepoLike`, `GoalRepoLike`, `SpaceTaskRepoLike`) avoids pulling SQLite into the resolver, and the test suite is thorough with pure mocks (no DB setup). All 61 tests pass.

### P2 -- Missing defensive cross-room check on goal resolver

**File:** `packages/daemon/src/lib/agent/reference-resolver.ts`, line 502-516

The room task resolver (line 463-469) and space task resolver (line 489-493) both perform defensive cross-context checks after fetching from the repo. The goal resolver at line 511-516 does not. While `getGoalByShortId` is already scoped by `roomId` in the SQL query (`WHERE room_id = ? AND short_id = ?`), the room task resolver has the same SQL scoping yet still performs a defensive check "in case the repo isn't scoped." For consistency and defense-in-depth, the goal resolver should also verify `goal.roomId === context.roomId` after the fetch. The session-level `ReferenceResolver` in `packages/daemon/src/lib/session/reference-resolver.ts` line 188 does perform this check.

### P3 -- Stale doc comment in test file

**File:** `packages/daemon/tests/unit/reference-resolver.test.ts`, line 10

The comment still reads `task/goal references return null (stubs pending Task 3.1)`. This should be updated to reflect that task/goal resolution is now implemented.

### P3 -- Stale section header in test file

**File:** `packages/daemon/tests/unit/reference-resolver.test.ts`, line 507

The section header reads `// Task / goal references (stubs)` but the `describe` block title on line 510 was updated. The comment above it is also stale.

1. **Dual ReferenceResolver classes**: There are two `ReferenceResolver` classes in the codebase -- `lib/agent/reference-resolver.ts` (this PR) and `lib/session/reference-resolver.ts` (existing). They share similar logic but have different interfaces and slightly different behavior (e.g., the session one uses `Promise.all` for parallel resolution, tries full UUID lookup before short ID for tasks, and passes the mention ID through rather than the entity UUID). This is not a problem introduced by this PR, but worth being aware of to avoid drift.

2. **`resolveRoomTask` is sync but `resolveTask` is async**: The `resolveRoomTask` and `resolveGoal` methods are synchronous, but they're called from the async `resolveTask` and `resolveReference`. This is fine functionally but worth noting -- the `async` keyword on `resolveTask` is necessary only because of the potential future path through `resolveSpaceTask` if that ever became async.

3. **`task as unknown as NeoTask` cast**: The space task-to-`NeoTask` cast at line 495 uses `as unknown as NeoTask` since `SpaceTask` and `NeoTask` are structurally different types. This is acceptable since the downstream consumer (`reference-context-builder.ts`) already casts with `as NeoTask` on the data. The `id` field will be the space task UUID, not the `st-<uuid>` prefix, which is consistent.

lsm (commented):
## Summary

This is a solid implementation of task and goal resolvers with clean dependency injection via structural interfaces. The test coverage is comprehensive and follows project conventions (pure mocks, no SQLite). However, I have a few concerns that should be addressed before merge.

---

## What's Done Well

1. **Clean architecture with structural interfaces** — `TaskRepoLike`, `GoalRepoLike`, `SpaceTaskRepoLike` avoid pulling SQLite into the resolver module. This is the right pattern.

2. **Cross-context rejection** — Room tasks verify `task.roomId === context.roomId`, space tasks verify `task.spaceId === context.spaceId`. Defense in depth is good.

3. **Graceful null returns** — All edge cases (missing context, missing repo, not found, cross-context) return null with appropriate log warnings.

4. **Comprehensive test coverage** — Room tasks, space tasks, goals, missing entities, cross-context rejection, unrecognized formats, and `resolveAllReferences` integration are all covered.

---

## Issues to Address

### 1. Type assertion smell in `resolveSpaceTask` (line 495)

```typescript
return { type: 'task', id: task.id, data: task as unknown as NeoTask };
```

The `SpaceTaskRepoLike` interface returns `{ id: string; spaceId: string; [key: string]: unknown }`, but the resolver casts this to `NeoTask`. This is misleading because `SpaceTask` and `NeoTask` are different types:

- `NeoTask` has `roomId`, `shortId`, `dependsOn` (room-scoped concepts)
- `SpaceTask` has `spaceId`, `agentName`, `completionSummary` (space-scoped concepts)

**Recommendation**: Either:
- Use `SpaceTask` in the return type (if `ResolvedTaskReference.data` is meant to hold either), or
- Define `ResolvedSpaceTaskReference` as a separate type, or
- Document why the cast is safe (e.g., "consumer only uses `id` and `spaceId`")

### 2. Missing cross-room check for goals

Room tasks have a defensive cross-room check at lines 463-469:
```typescript
if (task.roomId !== context.roomId) {
    log.warn(`Cross-room reference rejected...`);
    return null;
}
```

But `resolveGoal` (lines 502-516) lacks this check. It trusts `goalRepo.getGoalByShortId(roomId, shortId)` to be correctly scoped.

**Recommendation**: Add the same defensive check for consistency:
```typescript
if (goal.roomId !== context.roomId) {
    log.warn(`Cross-room reference rejected: goal "${shortId}" belongs to room "${goal.roomId}", not "${context.roomId}"`);
    return null;
}
```

### 3. Inconsistent method signatures (async vs sync)

- `resolveTask` is declared `async` (line 432) but its implementations (`resolveRoomTask`, `resolveSpaceTask`) are synchronous.
- `resolveGoal` (line 502) is synchronous.
- `resolveFile` and `resolveFolder` are async because they do I/O.

**Recommendation**: Remove `async` from `resolveTask` since it doesn't await anything, or document that the `async` is for future extensibility. Currently it creates a false expectation of async behavior.

---

## Minor Suggestions (Non-Blocking)

### Test helper `mockSpaceTaskRepo` doesn't use arguments

```typescript
function mockSpaceTaskRepo(returnValue?: ReturnType<typeof makeSpaceTask>): SpaceTaskRepoLike {
    return { getTask: mock(() => returnValue ?? null) };
}
```

The mock ignores the `id` argument passed to `getTask(id)` — it always returns the same `returnValue`. This is fine for current tests, but if you add tests that verify the repo is called with the correct ID, you'll need to update this. Consider adding a comment or using a conditional mock like `mockTaskRepo` does in one test (line 862).

### Consider extracting prefix constants

The prefix strings (`'t-'`, `'st-'`, `'g-'`) are hardcoded in multiple places. Consider extracting:
```typescript
const ROOM_TASK_PREFIX = 't-';
const SPACE_TASK_PREFIX = 'st-';
const GOAL_PREFIX = 'g-';
```

---

## Questions

1. **Why does `SpaceTaskRepoLike.getTask` use a different ID scheme?** Room tasks use `shortId` (e.g., `t-1`) while space tasks use UUID (`st-<uuid>`). Is this intentional, and is the UUID expected to be the actual `id` field or just any UUID? The code strips `st-` and passes the remainder to `getTask()`, which seems correct but worth confirming.

2. **Is `resolveSpaceTask` supposed to work without `roomId`?** Currently it only checks `spaceId`. A session could have both contexts — should space tasks also verify room membership?

---

Overall this is a well-structured implementation. Please address the type safety and consistency issues above.

PR #860, opened 2026-03-25
feat(room): inject merged MCP servers into worker sessions via setRuntimeMcpServers() (Task 3.3)

lsm:
In createSessionFactory(), worker sessions (coder/general) now receive a merged
MCP map of file-based + registry-sourced servers via setRuntimeMcpServers() before
the SDK query starts. File-based config takes precedence over the registry on name
collision (local project config is more specific than the global app-level registry).
A WARN log is emitted for each collision.

Hot-reload for worker sessions is intentionally skipped: workers are short-lived
(one task per session), so registry changes take effect on the next session creation.

Adds 8 unit tests covering merge correctness, collision precedence, neither source
being dropped, and non-worker roles (leader/planner) not receiving the injection.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -250,13 +250,27 @@ export class RoomRuntimeService {
 				agentSessions.set(init.sessionId, session);
 
 				// Inject merged MCP servers for worker sessions (coder / general).
-				// File-based servers take precedence over registry servers on name collision
-				// because local project config is considered more specific than the global
-				// application-level registry.
 				//
-				// Hot-reload for worker sessions is intentionally skipped: workers are
-				// short-lived (one task per session) so a registry change mid-task would
-				// not be useful. The updated map is applied on the next session creation.
+				// Precedence (highest to lowest): file-based > registry.
+				// File-based servers are project-local config (.mcp.json / settings.json) and are
+				// considered more specific than the global application-level registry. When the same
+				// server name appears in both sources, the project's local definition wins.
+				//
+				// Note: Room chat sessions use the OPPOSITE precedence (registry > file-based) because
+				// room chat does not use project-local tools and the app-level registry is intended to
+				// extend the room chat experience. Worker sessions operate in the project workspace, so
+				// the project's local config should take priority.
+				//
+				// Hot-reload for worker sessions is intentionally skipped: workers are short-lived
+				// (one task per session) so a registry change mid-task would not be useful. The
+				// updated map is applied on the next session creation.
+				//
+				// Known limitation after daemon restart: recovered worker sessions re-created by
+				// restoreSession() will NOT have the merged (file + registry) MCP map reapplied —
+				// restoreMcpServersForGroup() only restores role-specific in-process tools
+				// (planner-tools, leader-agent-tools). Workers resumed from a crash will therefore
+				// run without user-configured MCP servers for the remainder of their task. This is
+				// accepted given the short-lived nature of worker sessions.
 				if (role === 'coder' || role === 'general') {
 					const fileMcpServers = ctx.settingsManager.getEnabledMcpServersConfig();
 					const registryMcpServers = ctx.appMcpManager?.getEnabledMcpConfigs() ?? {};
@@ -271,12 +285,17 @@ export class RoomRuntimeService {
 						}
 					}
 
-					// Merge: registry first, then file-based overwrites on collision
+					// Merge: registry first, then file-based overwrites on collision.
+					// Only call setRuntimeMcpServers when there is at least one server to inject —
+					// an undefined config lets the SDK use its own default discovery, while an
+					// empty map would suppress it entirely with no benefit.
 					const merged: Record<string, McpServerConfig> = {
 						...registryMcpServers,
 						...fileMcpServers,
 					};
-					session.setRuntimeMcpServers(merged);
+					if (Object.keys(merged).length > 0) {
+						session.setRuntimeMcpServers(merged);
+					}
 				}
 
 				// Leader sessions are started lazily: injectMessage() calls ensureQueryStarted()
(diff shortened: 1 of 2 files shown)


lsm (commented):

---

The implementation is clean, well-tested, and correctly integrates the lifecycle manager into worker sessions. The merge logic, injection timing, and role gating are all correct. The issues I found are minor and non-blocking.

---

### Correctness

**Merge precedence is correct for the stated intent.** The spread order `{ ...registryMcpServers, ...fileMcpServers }` at line 275-278 of `room-runtime-service.ts` ensures file-based servers overwrite registry servers on name collision, which matches the PR description and comment. The tests at lines 183-201 of the test file verify this with concrete server configs, and the test passes.

**Both sources are always present.** The `?? {}` on `ctx.appMcpManager?.getEnabledMcpConfigs()` (line 262) ensures the registry source is never `undefined`/`null` even when `appMcpManager` is absent. The test at lines 257-271 confirms the `hasAppMcpManager: false` path works. File-based servers are always read via `ctx.settingsManager.getEnabledMcpServersConfig()` (no null guard needed since `settingsManager` is required in the config interface). The test at lines 204-221 explicitly verifies both unique names survive the merge.

**Injection timing is correct.** `setRuntimeMcpServers()` (line 279) is called before `startStreamingQuery()` (line 286). Looking at `AgentSession.setRuntimeMcpServers()` (agent-session.ts:646-651), it directly sets `this.session.config.mcpServers` which is the config read when the SDK query starts. Since `startStreamingQuery()` has not yet been called, the MCP config will be in place before the first SDK message.

**Role gating is correct.** The `if (role === 'coder' || role === 'general')` guard (line 260) ensures only worker sessions get the injection. Leader sessions skip both the MCP injection and the `startStreamingQuery()` call (line 285-286). The tests at lines 224-254 verify both leader and planner roles are excluded.

---

### P1: Merge precedence inconsistency with room chat sessions

The new worker code uses **file-based wins on collision**:

```ts
// room-runtime-service.ts:275-278 (worker sessions - NEW)
const merged: Record<string, McpServerConfig> = {
    ...registryMcpServers,
    ...fileMcpServers,     // file-based WINS
};
```

But the existing room chat session code uses the **opposite precedence -- registry wins on collision**:

```ts
// room-runtime-service.ts:597-599 (room chat sessions - EXISTING)
roomChatSession.setRuntimeMcpServers({
    ...fileMcpServers,     // file-based first
    ...registryMcpServers, // registry WINS
    'room-agent-tools': roomAgentMcpServer,
});
```

And the hot-reload handler follows the same pattern (lines 694-696).

The worker-session choice (file wins) is arguably the better default since file-based config is project-local. However, having opposite precedence rules in two different session types within the same service is confusing. If a user has an MCP server named "github" in both their `.mcp.json` (file) and the app registry, the room chat agent would use the registry version while the worker agent would use the file version. This is not necessarily a bug, but it should be a deliberate, documented decision with a clear comment in both places explaining why the precedence differs.

**Files:** `packages/daemon/src/lib/room/runtime/room-runtime-service.ts` lines 275-278 vs 597-599

---

### P1: Restored worker sessions lose registry-sourced MCP servers after daemon restart

When a daemon restarts, the recovery path at line 799 calls `runtime.restoreMcpServersForGroup(group)`. Looking at `room-runtime.ts:2115-2201`, this method only restores role-specific MCP servers ("planner-tools" for planner workers, "leader-agent-tools" for leaders) via `setSessionMcpServers()`, which is a **replace** operation (not merge).

This means after a daemon restart:
- A coder worker that was originally created with the merged (file + registry) MCP map will only have whatever it had before the crash -- but `setRuntimeMcpServers` is a runtime-only config that is intentionally not persisted (per session-repository.ts:135). After `restoreSession()` (line 348-366), the session exists but has **no** MCP servers at all. Then `restoreMcpServersForGroup` sets only the role-specific tools.
- The file-based and registry-based MCP servers are **lost** until the next session is created.

The PR description says "workers are short-lived (one task per session) so registry changes take effect on the next session creation," which is a reasonable stance. However, this means mid-crash worker sessions resumed by `recoverRoomRuntime` will run without their user-configured MCP servers. This may be acceptable given the short-lived nature, but it should be explicitly acknowledged as a known limitation.

**Files:** `packages/daemon/src/lib/room/runtime/room-runtime-service.ts` lines 786-826, `packages/daemon/src/lib/room/runtime/room-runtime.ts` lines 2115-2201

---

### P2: Consider skipping the empty-map call when neither source has servers

At line 279, `session.setRuntimeMcpServers(merged)` is always called for worker sessions, even when the merged map is empty. The test at lines 274-288 explicitly verifies this behavior. While harmless, it does overwrite any MCP servers that might have been set on the session object before this point (e.g., from a future code path that pre-populates session config). Consider guarding the call with a length check:

```ts
if (Object.keys(merged).length > 0) {
    session.setRuntimeMcpServers(merged);
}
```

This would also make the intent clearer -- you are only injecting when there is something to inject. The test for the empty-map case would then need to be inverted to assert 0 calls.

**File:** `packages/daemon/src/lib/room/runtime/room-runtime-service.ts` line 279

---

### P3: Test helper mockRestore() not called in error paths

In `buildSessionFactory()` (test file), `fromInitSpy.mockRestore()` is called at the end of every test. However, if a test assertion fails before that line, the spy leaks into the next test. The tests currently all pass, but this is fragile. Consider using `afterEach` for cleanup, or wrapping the factory creation in a pattern that guarantees restore. This is a test-only concern and not production-impacting.

**File:** `packages/daemon/tests/unit/room/room-runtime-service-worker-mcp.test.ts`

---

### Test Coverage Summary

The 8 tests cover:
- Coder sessions receive merged map (yes)
- General sessions receive merged map (yes)
- File-based wins on collision (yes)
- Both sources present, neither dropped (yes)
- Leader sessions excluded (yes)
- Planner sessions excluded (yes)
- Missing appMcpManager handled gracefully (yes)
- Empty sources produce empty map (yes)

Missing coverage that would be nice-to-have but not required:
- Multiple collision names in a single merge (only one collision tested)
- The `Object.prototype.hasOwnProperty.call` safety check for inherited properties (e.g., if someone adds properties to `Object.prototype`)

lsm (commented):

---

### Verdict: APPROVE

All five items from the previous review have been addressed correctly.

### Item-by-item verification

**P1 — Precedence inconsistency documented**: Correct. Both merge sites now have explicit cross-references. The worker merge block (lines 252-273 of room-runtime-service.ts) explains file-based > registry precedence with a rationale, and explicitly calls out the opposite ordering in room chat. The room chat merge block (lines 602-615) reciprocally points back to createSessionFactory(). The rationale is clear and consistent.

**P1 — Daemon restart limitation documented**: Correct. The comment at recoverRoomRuntime (lines 823-828) now explicitly acknowledges that file-based and registry MCP servers are NOT reapplied for recovered worker sessions, explains why (restoreMcpServersForGroup only restores in-process tools), and justifies the accepted behavior (workers are short-lived).

**P2 — Empty-map guard added**: Correct. Line 296-298 now wraps the session.setRuntimeMcpServers(merged) call in an if (Object.keys(merged).length > 0) guard. The accompanying comment (lines 289-291) explains the rationale: an empty map would suppress the SDK's default server discovery with no benefit. One minor observation: the mcp.registry.changed handler (lines 719-727) does NOT have this same guard. However, that path always includes at least room-agent-tools when the server exists (line 724-726), so an empty-merge scenario is only possible if no runtime rooms exist — which would mean the loop body never executes. No action required, but worth noting for consistency awareness.

**P3 — Test spy cleanup fixed**: Correct. The afterEach hook (lines 151-155 of the test file) collects spies in a shared array and restores them all after each test. This is more robust than inline mockRestore() calls that could be skipped if an assertion throws early.

**Empty-map test updated**: Correct. The test at lines 280-293 now expects calls.length === 0, matching the new guard behavior. The test description and inline comment clearly document the expected behavior.

### Test results

All 8 tests pass (16 expect() calls). No regressions introduced.

### Remaining observations (no action required)

- The mcp.registry.changed hot-reload path (line 727) unconditionally calls setRuntimeMcpServers(merged). Since room-agent-tools is always included for rooms that have been set up, this is safe in practice. However, if the event fires for a room that somehow lost its agent server (e.g., after stopRuntime followed by a stale event), the call would still proceed. This is not a bug given current flow — just a note for future maintenance.

No blocking issues remain.

PR #864, opened 2026-03-25
feat(space): unified channel resolution with resolveChannels() and validateChannels()

lsm:
- Add resolveChannels(workflow, agents) as single resolution path for all channel
  types (within-node DM, cross-node DM, fan-out by node name, wildcard broadcast)
- Add validateChannels(workflow, agents) for workflow-level validation including
  globally-unique role checks and agent/node reference validation
- Add isFanOut?, channelId?, isCyclic? fields to ResolvedChannel (backward compat)
- Add id?, isCyclic? fields to WorkflowChannel
- Add channels?: WorkflowChannel[] to SpaceWorkflow, CreateSpaceWorkflowParams,
  UpdateSpaceWorkflowParams for workflow-level channel declarations
- Deprecate resolveNodeChannels() and validateNodeChannels() — no breaking changes
- 23 new tests covering all resolution and validation scenarios

### DIFF
--- a/packages/daemon/src/lib/space/export-format.ts
+++ b/packages/daemon/src/lib/space/export-format.ts
@@ -68,9 +68,11 @@ const exportedWorkflowNodeAgentSchema = z.object({
 });
 
 const workflowChannelSchema = z.object({
+	id: z.string().optional(),
 	from: z.string().min(1),
 	to: z.union([z.string().min(1), z.array(z.string().min(1))]),
 	direction: z.enum(['one-way', 'bidirectional']),
+	isCyclic: z.boolean().optional(),
 	label: z.string().optional(),
 });
 
@@ -141,6 +143,7 @@ const exportedWorkflowBaseSchema = z.object({
 	rules: z.array(exportedWorkflowRuleSchema),
 	tags: z.array(z.string()),
 	config: z.record(z.string(), z.unknown()).optional(),
+	channels: z.array(workflowChannelSchema).optional(),
 });
 
 const exportBundleBaseSchema = z.object({
@@ -304,6 +307,8 @@ export function exportWorkflow(
 	};
 	if (workflow.description !== undefined) result.description = workflow.description;
 	if (workflow.config !== undefined) result.config = workflow.config;
+	// Workflow-level channels use role strings and node names — no UUID remapping needed.
+	if (workflow.channels && workflow.channels.length > 0) result.channels = workflow.channels;
 	return result;
 }
 
--- a/packages/daemon/tests/unit/space/channel-resolution.test.ts
+++ b/packages/daemon/tests/unit/space/channel-resolution.test.ts
@@ -433,4 +433,136 @@ describe('validateChannels', () => {
 		const errors = validateChannels(wf, agents);
 		expect(errors).toHaveLength(0);
 	});
+
+	test('node-name/role-name collision: ambiguity error reported for from', () => {
+		// Node named "coder" and an agent with role "coder" in a different node
+		const agents = [makeAgent('agent-coder'), makeAgent('agent-reviewer')];
+		const nodeA = makeNode('coder', 'coder', [{ role: 'reviewer', agentId: 'agent-reviewer' }]);
+		const nodeB = makeNode('n2', 'NodeB', [{ role: 'coder', agentId: 'agent-coder' }]);
+		const wf = makeWorkflow(
+			[nodeA, nodeB],
+			[{ from: 'coder', to: 'reviewer', direction: 'one-way' }]
+		);
+		const errors = validateChannels(wf, agents);
+		expect(errors.length).toBeGreaterThan(0);
+		expect(errors.some((e) => e.includes('ambiguous'))).toBe(true);
+	});
+
+	test('node-name/role-name collision: ambiguity error reported for to', () => {
+		const agents = [makeAgent('agent-coder'), makeAgent('agent-reviewer')];
+		const nodeA = makeNode('n1', 'reviewer', [{ role: 'coder', agentId: 'agent-coder' }]);
+		const nodeB = makeNode('n2', 'NodeB', [{ role: 'reviewer', agentId: 'agent-reviewer' }]);
+		const wf = makeWorkflow(
+			[nodeA, nodeB],
+			[{ from: 'coder', to: 'reviewer', direction: 'one-way' }]
+		);
+		const errors = validateChannels(wf, agents);
+		expect(errors.length).toBeGreaterThan(0);
+		expect(errors.some((e) => e.includes('ambiguous'))).toBe(true);
+	});
+
+	test('invalid direction value: error reported', () => {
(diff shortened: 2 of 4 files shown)


lsm (commented):

Overall the implementation is solid -- the unified resolver correctly handles all channel types, bidirectional expansion works as expected, backward compatibility with deprecated functions is maintained, and the 23 new tests cover the main scenarios well. However, I found a few issues that should be addressed before merge.

---

### P1: Node-name/role-name collision is silently misrouted

**Files:** `packages/shared/src/types/space-utils.ts` (lines 398-409, 459-467)

When a node name happens to also be an agent role (e.g., a node named "coder" and an agent with role "coder" in a different node), the resolver silently resolves to the agent role, not the node name. The priority order is: role first, then node name. This is by design in `resolveAgentRef()`, but `validateChannels()` does not warn about this ambiguity -- it accepts a value that matches both `knownRoles` and `knownNodeNames` without flagging it.

This means a user could declare `from: "NodeName"` intending a fan-out to all agents in that node, but if "NodeName" also happens to be an agent role elsewhere, the resolver will send to a single agent instead, and validation will pass silently.

Suggestion: Add a check in `validateChannels()` that warns when a `from`/`to` value matches both a known role and a known node name, since the behavior is ambiguous.

### P1: `isFanOut` is not set for `from`-side node-name resolution

**Files:** `packages/shared/src/types/space-utils.ts` (lines 394-409)

The `isFanOut` flag is only set when the `to` side matches a node name (line 405). However, `resolveAgentRef()` (line 463) also expands a `from` value matching a node name to all agents in that node. When `from` is a node name, `isFanOut` remains `false` even though the channel is effectively fanning out from all agents in that node. The semantics of `isFanOut` are asymmetrical in a way that is not documented and may confuse downstream consumers.

Suggestion: Either set `isFanOut = true` for both sides when either side references a node name, or document that `isFanOut` specifically refers to `to`-side fan-out only.

### P2: Export/import Zod schema does not include new `id` and `isCyclic` fields

**File:** `packages/daemon/src/lib/space/export-format.ts` (lines 70-75)

The `workflowChannelSchema` used for import validation only has `from`, `to`, `direction`, and `label`. The new `id` and `isCyclic` fields added to `WorkflowChannel` are not in the schema, so they would be silently stripped during import validation. This causes data loss on round-trip export/import for workflows using these new fields.

Additionally, `ExportedSpaceWorkflow` and `exportedWorkflowBaseSchema` do not include workflow-level `channels` at all, so workflow-level channels would be lost on export/import entirely.

### P2: Test gaps for edge cases

**File:** `packages/daemon/tests/unit/space/channel-resolution.test.ts`

The following scenarios are untested:
1. **Node-name/role-name collision**: What happens when a value matches both a node name and an agent role
2. **`from: '*'` (wildcard from)**: Only `to: '*'` is tested; the `from` wildcard expansion path is not verified
3. **`from` as a node name**: Tests only use node names in `to`, never in `from`
4. **`bidirectional` with node-name fan-out on `to` side**: Only tested with explicit role arrays
5. **Invalid `direction` value**: `validateChannels` does not validate the `direction` field at all

### P3: Unused `_agents` parameter

**File:** `packages/shared/src/types/space-utils.ts` (lines 335-337)

The `resolveChannels()` function accepts an optional `_agents` parameter that is never used. The docstring says "reserved for future gate/context use" but this is YAGNI -- it adds noise to the API surface and will confuse callers into thinking they need to pass agents. If it's truly needed later, it can be added then.

### P3: Minor note -- `validateChannels` does not validate `direction` field

**File:** `packages/shared/src/types/space-utils.ts` (lines 542-566)

The validation checks `from`/`to` references and wildcard mixing, but does not validate that `direction` is one of the allowed values (`'one-way'` | `'bidirectional'`). This is already handled by the Zod schema at the API boundary (in `space-workflow-manager.ts`), so it's not a bug per se, but it's a gap in the standalone validation function.

lsm (commented):

**Recommendation: APPROVE** (self-review)

All previous P1/P2 issues have been properly addressed. Here is a detailed assessment:

---

### P1: Collision detection in `validateChannels` -- Fixed

The ambiguity detection at `space-utils.ts:555-568` and `587-594` correctly flags `from`/`to` values that match both an agent role and a node name. The validation produces a clear, actionable error message directing the user to rename either the node or the role. Tests at lines 437-462 confirm both `from` and `to` collision cases are caught.

### P1: Asymmetrical `isFanOut` semantics -- Documented

The JSDoc on `ResolvedChannel.isFanOut` (lines 42-50) now explicitly explains that the flag is a **to-side** concept: from-side node expansion produces individual entries with `isFanOut: false`, while to-side node expansion sets `isFanOut: true`. The test at line 507-525 ("from as node name") verifies this behavior directly.

### P2: Zod schemas and export format -- Complete

- `workflowChannelSchema` in `export-format.ts:70-77` now includes `id: z.string().optional()` and `isCyclic: z.boolean().optional()`.
- `exportedWorkflowBaseSchema` at line 146 includes `channels: z.array(workflowChannelSchema).optional()`.
- `exportWorkflow()` at line 311 exports `workflow.channels` as-is.
- `ExportedSpaceWorkflow` interface in `space.ts:1095` has `channels?: WorkflowChannel[]`.

Export/import round-trip for `id` and `isCyclic` is now preserved.

### P2: New edge case tests -- Comprehensive

7 new tests added (23 total on top of the original 16):
1. Wildcard `from (*)` expansion -- confirms self-loops are skipped
2. From-as-node-name -- confirms individual point-to-point entries with `isFanOut: false`
3. Bidirectional + fan-out to node name -- confirms `isFanOut: true` on all entries (forward and reverse) and `isHubSpoke: true`
4. Node-name/role collision resolution preference -- confirms resolver prefers role over node name
5. Collision ambiguity error for `from` -- confirmed
6. Collision ambiguity error for `to` -- confirmed
7. Invalid direction value -- confirmed

All 30 tests pass. Typecheck and lint are clean.

1. **`isFanOut` on reverse bidirectional entries**: When a bidirectional channel's `to` is a node name, the reverse entries (spoke-to-hub) also get `isFanOut: true` (line 447). The JSDoc says `isFanOut` describes "to-side fan-out of the source WorkflowChannel," which is technically accurate but could be clearer that reverse entries inherit the flag as a channel-origin marker. The test at line 547 explicitly asserts this, confirming it is intentional. A brief note in the JSDoc mentioning that reverse entries inherit the flag would help future readers.

2. **`_agents` parameter unused**: The `resolveChannels()` function accepts `_agents` (line 342) marked as "reserved for future gate/context use." This is a standard forward-compatibility pattern and the underscore prefix makes the intent clear. No action needed.

3. **Node-level `channels` type in `ExportedWorkflowNode`**: Currently typed as `WorkflowChannel[]` (same as node-level channels), which means `id` and `isCyclic` are preserved in export. This is correct behavior. No issue.

---

No regressions detected. All previous issues resolved. Code is well-documented, well-tested, and consistent with the existing codebase patterns.

PR #865, opened 2026-03-25
test(space): add report_done tool and unit tests for agent completion signaling

lsm:
- Add `report_done` tool to step agent MCP server for explicit agent completion signaling
- Implement idempotent behavior: calling report_done multiple times returns success
- Store completion_summary per member in space_session_group_members (migration 51)
- Update list_peers to include completionSummary and status for peer visibility
- Add onReportDone callback to StepAgentToolsConfig for event emission
- Add SpaceSessionGroupMember.completionSummary field to shared types
- 8 new tests covering: basic success, summary storage, idempotency, list_peers state,
  multi-agent independent completion, callback invocation, and error cases

### DIFF
--- a/packages/daemon/src/lib/space/tools/step-agent-tool-schemas.ts
+++ b/packages/daemon/src/lib/space/tools/step-agent-tool-schemas.ts
@@ -1,10 +1,11 @@
 /**
- * Step Agent MCP Tool Schemas — Zod schemas and TypeScript types for the 2
+ * Step Agent MCP Tool Schemas — Zod schemas and TypeScript types for the 3
  * peer communication tools available to step agent sub-sessions.
  *
  * Tools:
  *   list_peers   — list other group members with their roles, statuses, and permitted channels
  *   send_message — channel-validated direct messaging tool
+ *   report_done  — signal that this agent has completed its work
  *
  * This file contains only schema definitions — no runtime logic or side effects.
  *
--- a/packages/daemon/src/lib/space/tools/step-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/step-agent-tools.ts
@@ -6,6 +6,7 @@
  *
  *   list_peers   — discover other group members with roles and permitted channels
  *   send_message — primary channel-validated direct messaging tool
+ *   report_done  — signal that this agent has completed its step task
  *
  * Communication model:
  * - Step agents communicate via declared channel topology (`send_message`).
--- a/packages/daemon/tests/unit/space/step-agent-tools.test.ts
+++ b/packages/daemon/tests/unit/space/step-agent-tools.test.ts
@@ -4,6 +4,7 @@
  * Covers all step agent peer communication tools:
  *   list_peers   — list peers excluding self and task-agent
  *   send_message — channel-validated direct messaging
+ *   report_done  — signal agent completion, persist summary, emit event
  *
  * Tests use a real SQLite database (via runMigrations) and mock message
  * injectors so no real agent sessions are created.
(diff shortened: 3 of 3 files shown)


lsm (commented):

All 33 tests pass (25 existing + 8 new). The core report_done logic, idempotency handling, completionSummary storage, and list_peers integration are correct. The DB migration is safe and idempotent. That said, there are a few issues worth addressing before merge.

---

### P1: onReportDone callback exception not handled

**File:** packages/daemon/src/lib/space/tools/step-agent-tools.ts, lines 367-369

If onReportDone throws, the error propagates unhandled. The DB update to completed has already committed (inside updateMembers transaction), so the agent is marked done in the DB, but the tool returns a failure to the calling agent. The agent may retry, and the idempotency guard will catch it -- so there is no data corruption -- but the error will be confusing. More importantly, if the callback is async (which the type signature allows), an unhandled promise rejection could escape.

The callback should be wrapped in try/catch so that DB state changes are never rolled back in the tool response.

### P1: onReportDone not wired in task-agent-manager.ts

**File:** packages/daemon/src/lib/space/runtime/task-agent-manager.ts, lines 1397-1407

The buildStepAgentMcpServerForSession method creates the step agent MCP server but does not pass onReportDone. The PR description says this callback enables event emission, but no caller actually provides it. If the callback is genuinely optional and will be wired in a follow-up task, that should be stated explicitly. If the intent is that the task manager should emit events when agents complete, that wiring is missing from this PR.

### P2: Stale file comment in step-agent-tool-schemas.ts

**File:** packages/daemon/src/lib/space/tools/step-agent-tool-schemas.ts, line 2

The header says Zod schemas and TypeScript types for the 2 peer communication tools but there are now 3 (list_peers, send_message, report_done).

### P2: Stale comment in task-agent-manager.ts

**File:** packages/daemon/src/lib/space/runtime/task-agent-manager.ts, line 1387

The JSDoc says The server gives the step agent peer communication tools (list_peers, send_message) -- should include report_done.

### P2: Missing test -- onReportDone not called on idempotent path

The idempotency test (line 716) verifies the second call returns idempotent: true, but does not assert that onReportDone was called only once. Adding an onReportDone callback to the idempotency test and asserting callbackArgs.length === 1 after the second call would close this gap.

### P2: Missing test -- onReportDone callback throwing

No test covers what happens when onReportDone throws. Even after wrapping in try/catch (per P1 above), a test verifying graceful degradation would document the expected behavior.

### P2: Missing test -- list_peers completionSummary null for non-completed peers

The new list_peers shows completion state after report_done test (line 736) verifies the reviewers completion state after reporting done, but does not verify that the coder peer (which has NOT called report_done) has completionSummary: null in the same response. This would strengthen confidence that the field is properly initialized.

### P3: Stale module docstring in step-agent-tools.ts

**File:** packages/daemon/src/lib/space/tools/step-agent-tools.ts, lines 7-8

The module-level docstring lists only list_peers and send_message in the These handlers implement section. report_done should be added.

---

### What looks good

- **Idempotency logic**: Clean early-return on status === completed with the original summary preserved. The test confirms the second summary is not overwritten.
- **DB migration 51**: Properly idempotent (tableExists + tableHasColumn guards), simple ALTER TABLE ADD COLUMN -- no data loss risk.
- **Repository layer**: UpdateMemberParams.completionSummary correctly uses ?? null for the SQL value, and rowToMember correctly reads it back with ?? undefined for the optional TS field.
- **list_peers integration**: The completionSummary: m.completionSummary ?? null mapping at line 166 correctly surfaces the new field.
- **Multi-agent test**: Properly verifies two independent agents can report done without interfering with each other.
- **Error path tests**: Group-not-found and session-not-in-group are both covered.
- **Test isolation**: Each test gets its own DB via makeCtx(), properly cleaned up in afterEach.

### Recommendation

REQUEST_CHANGES -- the two P1 issues should be addressed. The P2 items are important for completeness and documentation accuracy.

lsm (commented):

All 31 tests pass. The previous P1 and P2 issues have been properly addressed. No new issues introduced.

---

### Previous P1 issues -- resolved

**onReportDone callback exception not handled**: The `onReportDone` callback approach was replaced entirely with direct `daemonHub.emit()` wrapped in `void ... .catch()` (line 352-364 of `step-agent-tools.ts`). The catch handler logs a warning but never propagates the error. The new test "daemonHub emit throwing does not affect tool success" confirms DB state is committed and the tool returns success even when the hub throws.

**onReportDone not wired in task-agent-manager.ts**: Now wired at `task-agent-manager.ts:1418` as `daemonHub: this.config.daemonHub`. Verified all three call sites (lines 302, 1150, 1418 in the file) pass `daemonHub` correctly.

### Previous P2 issues -- resolved

- **Stale schema comment** (`step-agent-tool-schemas.ts`): Fixed -- "2" changed to "3", `report_done` added to the tool list in the module docstring.
- **Stale task-agent-manager JSDoc**: Already fixed on dev branch.
- **Missing test: callback throwing**: New test "daemonHub emit throwing does not affect tool success" covers this.
- **Missing test: hub not called on failure path**: New test "daemonHub not called when task update fails" verifies emit is skipped when `setTaskStatus` throws.
- **Stale module docstring** (`step-agent-tools.ts`): Fixed -- `report_done` added to the tool list at line 8.

### Previous P2 item not addressed -- assessed as non-issue

**"list_peers completionSummary null for non-completed peers"**: Upon inspection, `list_peers` surfaces session member `status` (active/inactive), not task completion state. Task completion is signaled via `daemonHub.emit('space.task.updated', ...)`, which is already tested. There is no `completionSummary` field in the `list_peers` response, so this test suggestion was based on a misunderstanding of the data model.

### Merge conflict resolution -- correct

The implementation uses dev's `taskManager.setTaskStatus` + `DaemonHub emit` approach instead of the worker's original `onReportDone` callback approach. This is the right call -- it eliminates the callback indirection layer entirely.

---

**Recommendation: APPROVE**

All P1/P2 issues from the previous review are resolved. The two new tests are well-structured, correctly test the non-fatal hub error behavior and the emit-ordering invariant, and follow the existing test patterns. Test isolation is clean (each test gets its own DB, properly cleaned up in afterEach).

PR #869, opened 2026-03-25
Plan: Fix model switching bugs in NeoKai

lsm:
## Bug 1 (Task View) — implement now

`session.model.switch` creates a NEW `AgentSession` via `SessionCache`, separate from `RoomRuntimeService.agentSessions`, causing duplicate concurrent queries. **Fix:** Register room sessions in `SessionCache` via `registerSession()` in both `createAndStartSession` and `restoreSession` paths (matching the Space pattern), add `unregisterSession()` for cleanup, fix `SessionCache.remove()` race condition. Tasks 2–3.

## Bug 2 (Normal Session) — investigate first

The previous plan proposed `forkSession: true` but this was never verified. Replaced with a single investigation task (Task 1) that writes an online integration test observing the SDK's `system:init` message before/after a model switch to prove or disprove whether the new model takes effect. After results are in, the human reviewer will decide the approach.

## Test fix

Bridged async `fs/promises` calls to sync `node:fs` counterparts in `AnthropicToCodexBridgeProvider` tests to work around a Bun 1.3.11 Linux CI issue.

## Plan

See `docs/plans/fix-model-switching-bugs-in-neokai.md` for full details.

### DIFF
--- a/docs/plans/fix-model-switching-bugs-in-neokai.md
+++ b/docs/plans/fix-model-switching-bugs-in-neokai.md
@@ -46,7 +46,7 @@ The root cause is that `session.model.switch` RPC handler in `session-handlers.t
 
 **Approach**: Write an online integration test (using dev proxy) that performs a model switch mid-session and observes the `system:init` message from the SDK subprocess. The `system:init` message's `model` field is the authoritative source for which model the SDK is actually using. By comparing the `model` field before and after the switch, we can determine whether the new model takes effect.
 
-**Observability mechanism**: Subscribe to `state.sdkMessages.delta` events on the session channel via WebSocket. Listen for `type === 'system' && subtype === 'init'` messages and read the `.model` field. This pattern is already used in `packages/daemon/tests/online/agent/agent-session-sdk.test.ts` (lines 300-326).
+**Observability mechanism**: Subscribe to `state.sdkMessages.delta` events on the session channel via `daemon.messageHub.onEvent()`. Listen for `type === 'system' && subtype === 'init'` messages and read the `.model` field. Use the `waitForSystemInit()` helper pattern from `packages/daemon/tests/online/coordinator/coordinator-mode-switch.test.ts` (lines 54-97), which joins the session channel via `daemon.messageHub.joinChannel('session:' + sessionId)`, parses the `added` array from the delta payload, and matches on `msg.type === 'system' && msg.subtype === 'init'`. Adapt this helper to also extract and return the `model` field.
 
 **Decision point**: After this investigation task completes, the human reviewer will decide the Bug 2 approach based on the evidence:
 - **If the test passes** (new model appears in `system:init` after switch): model switching works without forking — the bug may be elsewhere (e.g., UI-only, race condition in restart, or already fixed). No `forkSession` implementation needed.
@@ -62,20 +62,24 @@ The root cause is that `session.model.switch` RPC handler in `session-handlers.t
 
 **File**: `packages/daemon/tests/online/rpc/rpc-model-switching.test.ts` (append to existing file)
 
-**Key observability mechanism**: The SDK's `system:init` message (first message emitted by a new query) carries a `model` field that is the authoritative source for which model the subprocess is using. By subscribing to `state.sdkMessages.delta` events on the session channel and listening for `system:init` messages before and after the switch, we can determine whether the new model takes effect. This pattern is already used in `packages/daemon/tests/online/agent/agent-session-sdk.test.ts` (lines 300-326).
+**Key observability mechanism**: The SDK's `system:init` message (first message emitted by a new query) carries a `model` field that is the authoritative source for which model the subprocess is using. By subscribing to `state.sdkMessages.delta` events on the session channel and listening for `system:init` messages before and after the switch, we can determine whether the new model takes effect. Use the `waitForSystemInit()` helper pattern from `packages/daemon/tests/online/coordinator/coordinator-mode-switch.test.ts` (lines 54-97) — it uses `daemon.messageHub.onEvent('state.sdkMessages.delta', ...)` and `daemon.messageHub.joinChannel(...)`, which is the modern API (the old raw WebSocket `SUBSCRIBE` pattern in `agent-session-sdk.test.ts` is inside a `describe.skip` block marked "DEPRECATED").
 
 **Subtasks**:
-1. Create a helper that subscribes to `state.sdkMessages.delta` on a session channel and returns a Promise that resolves with the next `system:init` message's `model` field. Use the WebSocket subscription pattern from `agent-session-sdk.test.ts`.
+1. Create a `waitForSystemInit(sessionId)` helper adapted from `packages/daemon/tests/online/coordinator/coordinator-mode-switch.test.ts` (lines 54-97). It should:
+   - Join the session channel via `daemon.messageHub.joinChannel('session:' + sessionId)`
+   - Return a Promise that resolves with the `model` field from the next `system:init` delta event
+   - Listen via `daemon.messageHub.onEvent('state.sdkMessages.delta', ...)` and parse the `added` array
+   - Match on `msg.type === 'system' && msg.subtype === 'init'`
+   - Clean up the event listener after resolving
 
 2. Write a test (using dev proxy via `NEOKAI_USE_DEV_PROXY=1`) that:
    - Creates a session and sends a message to start a query.
-   - Captures the first `system:init` message's `model` field → `initialModel`.
-   - Waits for the query response to confirm the agent is working.
-   - Calls `session.model.switch` with a different model (e.g., `claude-sonnet-4-20250514` → `claude-haiku-4-5-20251001`).
-   - Waits for the model switch RPC to complete (the restart should happen automatically).
-   - Captures the next `system:init` message's `model` field → `postSwitchModel`.
+   - **Before sending**: call `waitForSystemInit(sessionId)` to capture the initial `model` field → `initialModel`. Wait for the query response to confirm the agent is working.
+   - Calls `session.model.switch` with a different model (e.g., `claude-sonnet-4-20250514` → `claude-haiku-4-5-20251001`) and awaits completion. The `restart()` happens internally — the RPC returns after the old query is stopped.
+   - **After the RPC returns, before sending the next message**: call `waitForSystemInit(sessionId)` to capture the post-switch `model` field → `postSwitchModel`. Do NOT subscribe before the RPC call — `restart()` may emit teardown-related events that could race with the subscription. The subscription must be set up after the RPC completes but before the post-switch message triggers a new query turn.
+   - Sends another message (this triggers the new query turn, which emits the `system:init` that `waitForSystemInit` is waiting for).
    - Asserts that `postSwitchModel !== initialModel` (the SDK subprocess is using the new model).
-   - Sends another message and waits for a response (verifying the agent is not stuck after the switch).
+   - Waits for the response (verifying the agent is not stuck after the switch).
 
 3. If the dev proxy does not support observing different models (it mocks all responses uniformly), the test can still verify that a NEW `system:init` is emitted after the switch (the `session_id` may or may not change). Document the limitation clearly.
 
@@ -192,7 +196,7 @@ The root cause is that `session.model.switch` RPC handler in `session-handlers.t
 - Restore path also registers sessions.
 - No regression in existing SessionCache tests.
 
-**Dependencies**: Task 3
+**Dependencies**: Task 2
 
 **Agent type**: coder
 
--- a/docs/plans/fix-model-switching-bugs-in-neokai.md
+++ b/docs/plans/fix-model-switching-bugs-in-neokai.md
@@ -20,15 +20,15 @@ The root cause is that `session.model.switch` RPC handler in `session-handlers.t
 
 **Fix — two registration paths**: Room sessions enter `RoomRuntimeService.agentSessions` through two distinct code paths, both of which must register in `SessionCache`:
 
-1. **New session creation** (`createAndStartSession` at line ~250): After `agentSessions.set(init.sessionId, session)`, call `ctx.sessionManager.registerSession(session)`.
+1. **New session creation** (`createAndStartSession` at line 272): After `agentSessions.set(init.sessionId, session)` (line 281), call `ctx.sessionManager.registerSession(session)`.
 
-2. **Session restoration after daemon restart** (`restoreSession` at line ~380): After `agentSessions.set(sessionId, session)`, call `ctx.sessionManager.registerSession(session)`. This is critical -- without it, the fix would only work until the first daemon restart.
+2. **Session restoration after daemon restart** (`restoreSession` at line 398): After `agentSessions.set(sessionId, session)` (line 411), call `ctx.sessionManager.registerSession(session)`. This is critical -- without it, the fix would only work until the first daemon restart.
 
 **Cleanup — all teardown paths**: Room sessions are removed from `agentSessions` through multiple code paths, each of which must unregister from `SessionCache`:
 
-1. **`stopSession()`** (line ~442, `finally` block): Called during task cancellation/completion. Add `ctx.sessionManager.unregisterSession(sessionId)`.
+1. **`stopSession()`** (line 458, `finally` block): Called during task cancellation/completion. After `agentSessions.delete(sessionId)` (line 473), add `ctx.sessionManager.unregisterSession(sessionId)`.
 
-2. **`RoomRuntimeService.stop()`** (line ~197): Called during daemon shutdown. After `this.agentSessions.clear()`, iterate the previously-cleared entries and call `ctx.sessionManager.unregisterSession()` for each. Alternatively, refactor to iterate and unregister before clearing.
+2. **`RoomRuntimeService.stop()`** (line 228): Called during daemon shutdown. Before `this.agentSessions.clear()` (line 234), iterate all entries and call `ctx.sessionManager.unregisterSession()` for each.
 
 **Why add `unregisterSession` to `SessionManager`**: `SessionManager.sessionCache` is `private`. Adding `unregisterSession(sessionId: string): void` as a thin wrapper over `this.sessionCache.remove(sessionId)` mirrors the existing `registerSession()` which wraps `this.sessionCache.set()`, maintaining API symmetry.
 
@@ -200,23 +200,23 @@ The fix requires three coordinated changes:
 
 1. Add `unregisterSession(sessionId: string): void` to `SessionManager` that delegates to `this.sessionCache.remove(sessionId)`. This mirrors `registerSession()` which delegates to `this.sessionCache.set()`.
 
-2. **Registration path 1 — new session creation**: In `createSessionFactory().createAndStartSession()` (line ~250), after `agentSessions.set(init.sessionId, session)`, add:
+2. **Registration path 1 — new session creation**: In `createSessionFactory().createAndStartSession()` (line 272), after `agentSessions.set(init.sessionId, session)` (line 281), add:
    ```ts
    ctx.sessionManager.registerSession(session);
    ```
 
-3. **Registration path 2 — daemon restart restoration**: In `createSessionFactory().restoreSession()` (line ~380), after `agentSessions.set(sessionId, session)`, add:
+3. **Registration path 2 — daemon restart restoration**: In `createSessionFactory().restoreSession()` (line 398), after `agentSessions.set(sessionId, session)` (line 411), add:
    ```ts
    ctx.sessionManager.registerSession(session);
    ```
    This is critical. Without this, sessions restored after a daemon restart would NOT be in SessionCache, and Bug 1 would reoccur on every restart. The `restoreSession()` path uses `AgentSession.restore()` instead of `AgentSession.fromInit()`, but the result is the same `AgentSession` instance that needs to be findable via `sessionManager.getSessionAsync()`.
 
-4. **Teardown path 1 — individual session stop**: In `stopSession()` (line ~442, in the `finally` block after `agentSessions.delete(sessionId)`), add:
+4. **Teardown path 1 — individual session stop**: In `stopSession()` (line 458, in the `finally` block after `agentSessions.delete(sessionId)` at line 473), add:
    ```ts
    ctx.sessionManager.unregisterSession(sessionId);
    ```
 
-5. **Teardown path 2 — service shutdown**: In `RoomRuntimeService.stop()` (line ~197), before `this.agentSessions.clear()`, iterate all entries and unregister each:
+5. **Teardown path 2 — service shutdown**: In `RoomRuntimeService.stop()` (line 228), before `this.agentSessions.clear()` (line 234), iterate all entries and unregister each:
    ```ts
    for (const sessionId of this.agentSessions.keys()) {
        ctx.sessionManager.unregisterSession(sessionId);
@@ -225,7 +225,7 @@ The fix requires three coordinated changes:
    ```
    This prevents stale room session references from remaining in SessionCache after daemon shutdown.
 
-6. Verify that the `SessionCache.getAsync()` guard (line ~100 of `session-cache.ts`) correctly prefers the registered instance if a concurrent `getAsync()` call is in-flight during registration.
+6. Verify that the `SessionCache.getAsync()` guard (line 99 of `session-cache.ts`) correctly prefers the registered instance if a concurrent `getAsync()` call is in-flight during registration.
 7. Document the namespace invariant: room role strings must never equal `"space"` (add a comment in the role definition code where roles are enumerated).
 
 **Acceptance criteria**:
--- a/docs/plans/fix-model-switching-bugs-in-neokai.md
+++ b/docs/plans/fix-model-switching-bugs-in-neokai.md
@@ -5,7 +5,7 @@
 Fix two model switching bugs that prevent model changes from taking effect:
 
 1. **Bug 1 (Task View)**: When switching models in the Task View, `session.model.switch` creates a NEW `AgentSession` via `SessionCache` (a separate instance from `RoomRuntimeService.agentSessions`), causing duplicate/conflicting concurrent queries for the same session.
-2. **Bug 2 (Normal Session)**: When switching models on a normal session, `sdkSessionId` is NOT cleared during the restart, so the new query resumes the old SDK session file (created with the old model). The SDK may use the session-file model over the options model, making the switch ineffective.
+2. **Bug 2 (Normal Session)**: When switching models on a normal session, the agent may stop responding or may not use the new model. Root cause is unverified — an investigation task (Task 1) will determine whether the issue actually exists and what fix is needed.
 
 ## Approach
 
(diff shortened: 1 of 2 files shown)


lsm (commented):

## Root Cause Analysis: Confirmed Correct

Both root cause analyses are accurate based on my codebase review:

**Bug 1 (Task View)**: Confirmed. Room worker/leader sessions live exclusively in `RoomRuntimeService.agentSessions` (line 76 of `room-runtime-service.ts`). They are NEVER registered in `SessionManager.sessionCache` -- unlike Space sessions which call `registerSession()`. When `session.model.switch` RPC handler (line 445 of `session-handlers.ts`) calls `sessionManager.getSessionAsync(targetSessionId)`, the `SessionCache.getAsync()` method falls through to `loadSessionAsync()` which creates a brand new `AgentSession` from DB. This duplicates the live instance running in the room runtime, causing two concurrent SDK subprocesses for the same session.

**Bug 2 (Normal Session)**: Confirmed. In `model-switch-handler.ts`, both branches (lines 189 and 228) call `db.updateSession()` to persist the new model/provider, then call `lifecycleManager.restart()` -- but neither clears `session.sdkSessionId`. The `QueryLifecycleManager.restart()` method (line 181) validates the SDK session file but only clears `sdkSessionId` when the file is physically missing. Since the old session file still exists, `sdkSessionId` survives the restart. Then `QueryOptionsBuilder.addSessionStateOptions()` (line 327) sets `result.resume = this.ctx.session.sdkSessionId`, causing the SDK to resume the old session file that was created with the previous model.

## Issues Found

### P1: Bug 1 fix has a simpler alternative -- registering room sessions in SessionCache

**File**: Plan Task 3

The plan proposes adding entirely new `room.runtime.model.switch` and `room.runtime.model.get` RPC handlers plus new methods on `RoomRuntimeService`. However, a significantly simpler approach exists: register room worker/leader sessions in `SessionManager.sessionCache` when they are created in `createSessionFactory().createAndStartSession()` (line 250 of `room-runtime-service.ts`).

Evidence: Space sessions already do exactly this via `TaskAgentManager` calling `sessionManager.registerSession()` at `packages/daemon/src/lib/space/runtime/task-agent-manager.ts:398`. The `SessionCache` even has a guard (line 99) to prefer the registered live instance over a DB-loaded duplicate.

This approach would:
- Fix the bug for ALL RPC handlers that use `sessionManager.getSessionAsync()` (not just model switching)
- Require zero new RPC handlers, no new RPC method names, no UI changes to route between `session.model.switch` vs `room.runtime.model.switch`
- Be consistent with the existing Space architecture

The plan should at minimum discuss this alternative and justify why the new RPC approach was chosen over it.

### P1: Task 4 UI routing logic is fragile and couples to room knowledge

**File**: Plan Task 4

The plan proposes adding a `roomId` prop to `TaskViewModelSelector` and conditionally routing between `session.model.switch` and `room.runtime.model.switch`. This creates several problems:

1. **Wrong abstraction level**: The component must know whether the session belongs to a room to pick the right RPC. This couples a UI component (model picker) to backend session storage architecture.
2. **Fragile detection**: The plan says to check `workerSession.context?.roomId` but the current `TaskInfoPanel` props do not pass `roomId` at all -- it would need to be threaded through from `TaskView`/`TaskViewV2`.
3. **If the simpler registerSession approach (above) is used**, this task becomes entirely unnecessary.

If the new RPC approach is kept, consider a cleaner alternative: make the `session.model.switch` handler itself route to `RoomRuntimeService` when `SessionCache` does not contain the session but `RoomRuntimeService.agentSessions` does. This keeps the routing logic in the backend where it belongs.

### P1: Missing edge case -- sdkSessionId clearing in the error path

**File**: Plan Task 1, `packages/daemon/src/lib/agent/model-switch-handler.ts`

The plan correctly identifies adding `session.sdkSessionId = undefined; db.updateSession(session.id, { sdkSessionId: undefined });` in both branches. However, there is an error path to consider: if `lifecycleManager.restart()` throws (line 248 in the `queryObject` branch), the model/provider were already persisted to DB (line 228) AND the in-memory `session.config` was updated (lines 222-223), but `sdkSessionId` was cleared. The session is now in a half-state: new model persisted, old sdkSessionId gone, but the query was NOT restarted. The next time the query starts, it will use the new model without resume (which is actually correct behavior), but the error handling at line 266 wraps this as a generic failure and the UI shows the old model. This is not necessarily a bug, but the plan should explicitly discuss this edge case so the implementer is aware.

### P2: Task 6 online test feasibility is unclear

**File**: Plan Task 6

The plan proposes an online integration test that:
1. Creates a session
2. Sends a message to establish `sdkSessionId`
3. Switches model
4. Verifies `sdkSessionId` is cleared
5. Sends another message to verify the agent responds

Step 2 requires a real API call (setting up the SDK subprocess), and step 5 requires a second real API call with a different model. This test needs:
- At least two valid API keys/credentials (for different models)
- The dev proxy infrastructure set up for mocking
- Significant setup complexity

The plan should clarify whether this will use `NEOKAI_USE_DEV_PROXY=1` and how model switching is handled in the dev proxy environment (the dev proxy would need to support two different models).

### P2: Task 7 E2E test deferral makes the plan incomplete

**File**: Plan Task 7

The plan acknowledges the task view E2E test may be deferred due to complexity, but deferring it means there is no automated verification that Bug 1 is actually fixed. The normal session E2E test (subtask 1) only covers Bug 2. Without a task view E2E test, a regression in Bug 1 could ship undetected.

Consider at minimum adding a unit test that verifies `sessionManager.getSessionAsync()` returns the room runtime session (if the registerSession approach is taken) or that the new RPC handler returns the correct session.

### P2: Missing discussion of SDK session file cleanup

**File**: Plan Task 1

When `sdkSessionId` is cleared, the old SDK session file on disk becomes orphaned. The `sdkSessionFile` at `~/.claude/projects/<workspace>/` persists. The plan does not address whether this orphaned file should be archived/deleted immediately or left for the existing periodic cleanup (`sdk.scan`/`sdk.cleanup` RPC handlers). This is not a blocking issue, but should be mentioned so the implementer makes a conscious decision.

### P3: Task 2 test file path is already populated

**File**: Plan Task 2

The plan says the test file is `packages/daemon/tests/unit/agent/model-switch-handler.test.ts` which already exists with comprehensive tests (615 lines). The task should reference the existing test structure and indicate where the new test cases should be added (e.g., new `describe` block within the existing file), not imply creating a new test file.

### P3: Minor -- Task 3 mentions `handleModelSwitch` which is correct

The plan references `agentSession.handleModelSwitch(model, provider)` which is confirmed to exist at `agent-session.ts:601` and delegates to `ModelSwitchHandler.switchModel()`. This is correct.

## Summary

The root cause analyses are excellent and technically precise. The Bug 2 fix (Task 1) is clean and well-scoped. However, the Bug 1 fix (Tasks 3-5) introduces unnecessary complexity by creating new RPC handlers instead of fixing the underlying session routing issue. A simpler registerSession-based approach exists and should be seriously considered.

lsm (commented):

The updated plan is a significant improvement. The `registerSession` approach for Bug 1 is the right architecture, the error handling edge case is now documented, the SDK session file cleanup strategy is reasonable, and the dev proxy clarification for the online test resolves the feasibility concern. However, there is one P1 omission and several P2 issues that should be addressed before implementation begins.

## Issues Found

### P1: Plan misses `restoreSession()` -- room sessions restored after daemon restart will still have the duplicate bug

**File**: Plan Task 3, `packages/daemon/src/lib/room/runtime/room-runtime-service.ts`

The plan correctly identifies `createAndStartSession()` (line ~240) as the place to add `registerSession`. However, there is a second session creation path: `restoreSession()` (line ~307) which is called during daemon restart via `recoverRuntime()` (line ~671). This method creates an `AgentSession` via `AgentSession.restore()` and stores it in `agentSessions` without calling `registerSession`:

```ts
restoreSession: async (sessionId) => {
    if (agentSessions.has(sessionId)) return true;
    const session = AgentSession.restore(sessionId, ctx.db, ...);
    if (!session) return false;
    agentSessions.set(sessionId, session);
    return true;  // <-- NO registerSession here
},
```

After a daemon restart, room sessions are restored via this path, not `createAndStartSession`. Without `registerSession` here, any subsequent `session.model.switch` (or `message.send`, etc.) call for a restored room session will still create a duplicate `AgentSession` via `SessionCache.loadSessionAsync()`. This is the same Bug 1, just via a different code path.

**Fix**: Add `ctx.sessionManager.registerSession(session)` in `restoreSession()` after `agentSessions.set(sessionId, session)`, and add `ctx.sessionManager.unregisterSession(sessionId)` to `stopSession()` (as already planned). Also consider whether `RoomRuntimeService.stop()` (line ~188) needs to unregister all remaining entries in `agentSessions` before clearing the map -- during shutdown, `sessionManager.cleanup()` runs later and clears the entire cache, so this is not a leak, but there is a window where stale references could cause issues if shutdown ordering changes.

### P2: PR body still references the OLD approach (new RPC handlers)

**File**: PR description

The PR body says:

> **Fix**: Add `room.runtime.model.switch` and `room.runtime.model.get` RPC handlers that operate on `RoomRuntimeService.agentSessions` directly.

And the test plan still lists:

> - [ ] Unit tests for `room.runtime.model.switch` / `room.runtime.model.get`

The plan document itself was updated to use the `registerSession` approach, but the PR body was not updated. This will cause confusion for implementers and reviewers. The PR body should reflect the current plan.

### P2: Inconsistency between Approach section and Task 3 on unregisterSession

**File**: Plan, "Approach" section vs Task 3

The Approach section says:

> In `stopSession()`, add `ctx.sessionManager.sessionCache.remove(sessionId)` in the `finally` block (since there is no `unregisterSession()` method on `SessionManager`, directly access the cache's `remove()`).

But Task 3 says:

> Add `unregisterSession(sessionId: string): void` to `SessionManager` that delegates to `this.sessionCache.remove(sessionId)`.

These are contradictory. The Approach section proposes directly accessing the private `sessionCache`, while Task 3 proposes adding a proper `unregisterSession()` method. Task 3 is clearly the correct approach (it follows the `registerSession`/`unregisterSession` symmetry), but the Approach section needs to be updated to match. Having contradictory guidance in the same document will confuse the implementer.

### P2: Task 6 E2E test requires dev proxy but E2E tests are meant to be pure browser tests

**File**: Plan Task 6

The plan says this E2E test "verifies the observable behavior: the agent continues to respond after a model switch." However, this requires the agent to actually respond to a message after the model switch, which requires a real (or mocked) API call. In E2E tests, the server is started via `make run-e2e` which does NOT use `NEOKAI_USE_DEV_PROXY=1`.

From `CLAUDE.md`: "E2E tests are pure browser-based Playwright tests simulating real end-user interactions." The test would need to either: (a) rely on the dev proxy being set up in the E2E environment, or (b) only verify the UI model indicator changes without requiring an actual agent response, or (c) note that this test needs special E2E infrastructure (dev proxy setup).

The plan should clarify how this E2E test handles the API response requirement. If the E2E test environment does not have a mock backend, the test cannot verify "the agent responds" -- it can only verify "the model indicator changes" and "the message input is not stuck."

## Previous Issues Status

| Previous Issue | Status | Notes |
|---|---|---|
| P1: registerSession alternative | **Addressed** | Plan now uses registerSession. Good architectural choice. |
| P1: Error handling edge case | **Addressed** | Plan documents the restart() failure path and confirms session state is consistent. |
| P1: SDK session file cleanup | **Addressed** | Orphan files strategy documented; defers to existing cleanup. Reasonable. |
| P2: Online test feasibility | **Addressed** | Dev proxy model-agnostic explanation is clear. |
| P2: E2E test detail | **Partially addressed** | Task view E2E deferred with good rationale, but normal E2E test (Task 6) has infrastructure gap (see P2 above). |
| P2: Rollback strategy | **Addressed** | Error handling section documents consistent state after failure. |

## Summary

The architectural pivot to `registerSession` is correct and well-justified. The main gap is the missing `restoreSession()` registration path, which would leave the same Bug 1 present after daemon restarts. The PR body and approach section also need updating for consistency.

lsm (commented):

---

### Verdict: REQUEST_CHANGES

The plan has improved significantly from round 2. The P1 (missing restoreSession registration path) and P2 (PR body inconsistency) issues are resolved. The teardown path analysis is now complete. However, there is one new P1 issue discovered during source code verification, and one P2 issue with the collision analysis.

---

### P1: `SessionCache.remove()` does not clear `sessionLoadLocks`, creating a race condition in `unregisterSession`

**File**: `packages/daemon/src/lib/session/session-cache.ts` (lines 141-143)

The plan proposes adding `unregisterSession()` that delegates to `this.sessionCache.remove(sessionId)`. However, `SessionCache.remove()` only deletes from `this.sessions` and does NOT clear `this.sessionLoadLocks`. The inverse operation `set()` (line 132-136) correctly does both:

```ts
set(sessionId: string, agentSession: AgentSession): void {
    this.sessions.set(sessionId, agentSession);
    this.sessionLoadLocks.delete(sessionId);  // <-- clears load lock
}

remove(sessionId: string): void {
    this.sessions.delete(sessionId);           // <-- does NOT clear load lock
}
```

**Race scenario**:
1. An RPC handler calls `getAsync(roomSessionId)` -- the session is not in cache (e.g., during a brief window), so a load lock is set and DB loading begins.
2. `stopSession()` calls `unregisterSession(roomSessionId)` -> `sessionCache.remove()` -- deletes from `sessions` but load lock remains.
3. The DB load completes. The guard at line 99 checks `!this.sessions.has(sessionId)` -- this is now `true` because `remove()` just deleted it.
4. The stale DB-loaded duplicate is set back into `sessions`, defeating the unregister.

**Fix**: Either (a) update `SessionCache.remove()` to also clear `sessionLoadLocks` (symmetric with `set()`), or (b) have `unregisterSession()` in `SessionManager` clear the load lock explicitly. Option (a) is cleaner and fixes the general asymmetry. The plan should note this as a prerequisite or part of Task 3.

Note: This same asymmetry exists in the current `SessionCache` code and could theoretically affect the existing Space session pattern too, though the risk is lower because Space sessions are typically not unregistered while concurrent loads are in-flight. Still, fixing it here is the right call.

---

### P2: Collision analysis claims disjoint namespaces but does not account for `room:chat:` prefix

The plan states:
> Space session IDs use the `space:{uuid}:task:{uuid}` prefix, while Room session IDs use `{role}:{roomId}:{taskId}:{uuid}` or `room:chat:{roomId}`.

The `space:{uuid}:task:{uuid}` format (verified at `task-agent-manager.ts:252`) and `{role}:{roomId}:{taskId}:{uuid}` format (verified at `task-group-manager.ts:242-243`) are indeed structurally disjoint from Space IDs. However, `room:chat:{roomId}` (verified at `room-runtime-service.ts:558`) could theoretically collide with a Space session ID if a Space session ID somehow starts with `room:chat:`. While this is extremely unlikely given the UUID-based format, the analysis should acknowledge this edge case explicitly rather than implying complete impossibility.

The more important concern: Space sub-session IDs use the format `space:{spaceId}:task:{taskId}:step:{stepId}` (per `task-agent-manager.ts:138`). If room session IDs ever used a `space` prefix (e.g., a role named `space`), there could be collisions. This is a theoretical concern but the analysis should note the invariant: room session role strings must never be `space` for the collision-free guarantee to hold.

**Recommendation**: Add a brief note that the collision-free guarantee depends on room session role strings never using `space` as a prefix, and that `room:chat:` is the only non-role-based session ID format in the room system.

---

### Resolved from Round 2

- **P1 (restoreSession registration)**: Properly addressed. Registration path 2 (line ~380) is correctly identified and the rationale is sound. The plan explicitly calls out why this is critical for daemon restart resilience.
- **P2 (PR body inconsistency)**: PR body now accurately describes the `registerSession` approach and matches the plan.
- **P2 (unregisterSession teardown paths)**: Both `stopSession()` (finally block, line ~442) and `RoomRuntimeService.stop()` (line ~197) are identified. The shutdown order in `app.ts` is correct: `rpcHandlerCleanup()` -> `roomRuntimeService.stop()` -> `taskAgentManager.cleanupAll()` -> `sessionManager.cleanup()`, so unregistering before the full cache clear is safe.
- **P2 (collision risk)**: Substantially addressed with namespace format documentation. The remaining gap is noted in the P2 above.

### Minor observations (no severity)

- The plan correctly identifies that `db.updateSession(session.id, { sdkSessionId: undefined })` works because `session-repository.ts:154` maps `undefined` to `NULL` via `updates.sdkSessionId ?? null`.
- Bug 2 error handling analysis is thorough: the `restart()` call is wrapped in try/catch, and by the time `restart()` may throw, `sdkSessionId` is already cleared in memory and DB, leaving the session in a consistent state.
- The deferred task view E2E test (Task 6) rationale is acceptable given the unit test coverage for Task 4.

lsm (commented):

### Round 4 Review

All P1 and P2 issues from round 3 have been adequately addressed. The plan is now ready for implementation.

#### P1 Fix Verification: Race condition in SessionCache.remove()

The fix is correct and complete. Adding `this.sessionLoadLocks.delete(sessionId)` to `remove()` at line 141 of `session-cache.ts` mirrors the behavior already present in `set()` (line 135). I verified the existing code:

- `set()` (line 132-136) clears both `sessions` and `sessionLoadLocks`.
- `remove()` (line 141-143) currently only clears `sessions`.
- The race scenario described in the plan is real: if `getAsync()` is awaiting a load lock at line 93, and `remove()` is called concurrently, the awaited load will eventually resolve and check `if (!this.sessions.has(sessionId))` at line 99 -- which will be `true` since we removed it -- and re-insert the stale DB-loaded instance at line 100. Clearing the load lock in `remove()` does NOT prevent this particular race because the `await loadPromise` at line 93 is already in-flight (the lock reference is captured). However, it prevents a *subsequent* `getAsync()` call from re-entering the load path and hitting the same stale re-insertion.

Wait -- on closer inspection, the existing guard at line 99 (`if (!this.sessions.has(sessionId))`) actually prevents the stale re-insertion when `set()` is called while a load is in-flight, but when `remove()` is called, it removes the session from `sessions`, making the guard at line 99 evaluate to `true`, which causes the re-insertion. Clearing `sessionLoadLocks` in `remove()` does not help with the *already-awaiting* call, since that caller already has the promise reference.

However, the `finally` block at line 107-109 already clears `sessionLoadLocks.delete(sessionId)` after the await completes. So the real question is: can a second `getAsync()` call enter while the first is still awaiting? The answer is no -- the lock at line 90 ensures only one load happens at a time, and the `finally` at line 108 cleans up.

The actual race that `sessionLoadLocks.delete()` in `remove()` prevents is: remove() is called, then a NEW `getAsync()` call arrives. Without clearing the lock, this new call would await the old (already-resolved) promise from `sessionLoadLocks.get()`. But since the `finally` block already deleted it, the old promise is already gone from the map. So `sessionLoadLocks.delete()` in `remove()` is technically redundant for the already-awaiting case but is a good defensive measure for consistency and correctness if the lock somehow persists.

This is a minor point but worth noting for the implementer: the primary protection against the stale re-insertion race is the guard at line 99 combined with `set()` clearing the lock. The `remove()` fix is a correctness improvement that ensures the lock map stays consistent.

Making this a prerequisite subtask (subtask 0 in Task 3) before adding `unregisterSession` is the right sequencing.

#### P2 Fix Verification: Namespace invariant documentation

The plan now documents the invariant that room role strings must never equal `"space"`, explaining the collision risk between room session IDs (`{role}:{roomId}:{taskId}:{uuid}`) and space session IDs (`space:{uuid}:task:{uuid}`). The plan correctly identifies that roles are freeform strings (`SpaceAgent.role` is `string` type) and users can define custom roles.

The plan says "add a comment in the role definition code where roles are enumerated" but the role is a freeform string (`role: string` in both `SpaceAgent` and `CreateSpaceAgentParams`), not an enum. There is no single canonical location where all roles are enumerated -- `seed-agents.ts` defines built-in roles but custom agents can have any role string. I suggest the comment be placed at:

1. The `SpaceAgent.role` type definition in `packages/shared/src/types/space.ts` (line ~1049) where the role field is documented.
2. The `createSubSession` / task creation path in `task-group-manager.ts` (line ~242) where `${role}:${room.id}:...` session IDs are constructed.

This is a P3 suggestion -- the plan's intent is clear enough for implementation.

#### Additional Observations (no issues)

1. **`clear()` method**: `SessionCache.clear()` at line 162-164 also only clears `sessions` but not `sessionLoadLocks`. This is only called during `SessionManager.cleanup()` (line 385), which is a shutdown path where no concurrent operations are expected. No fix needed here, but worth noting for awareness.

2. **Space session registration pattern**: The plan correctly references the existing Space pattern in `task-agent-manager.ts:398` where `registerSession()` is called after session creation. The room session registration follows the exact same pattern.

3. **Test plan for the race condition** (subtask 3 in Task 4): Adequately covers the scenario. Testing that an in-flight load lock does not re-insert a stale session after `remove()` clears it is the right verification.

#### Summary

The plan is well-structured and addresses the original bugs with minimal, targeted changes. The round 3 feedback has been incorporated. No blocking issues remain.

**Recommendation: APPROVE**

lsm (commented):

---

### Verdict: REQUEST_CHANGES

The `forkSession: true` approach is a sound architectural choice and correctly preserves session resumability. However, there is a critical interaction bug between the fork mechanism and `sdk-message-handler.ts` that will cause the plan to fail silently at runtime. The plan also has a secondary issue with the `sdkSessionId` flow that needs to be addressed.

---

### P0: `sdk-message-handler.ts` will NOT capture the new forked session ID

**File**: `packages/daemon/src/lib/agent/sdk-message-handler.ts`, line 650

The plan states at Task 1, subtask 3:

> The existing `sdk-message-handler.ts` (line 641) captures the new session ID from the `system:init` message and updates `session.sdkSessionId` in memory and DB. This happens automatically -- no change needed.

This is **incorrect**. The guard at line 650 is:

```ts
if (isSDKSystemInit(message) && !session.sdkSessionId && message.session_id) {
    session.sdkSessionId = message.session_id;
    // ...
}
```

The condition `!session.sdkSessionId` means: "only capture if we don't already have a session ID." After a fork, `session.sdkSessionId` is still set to the **old** session ID (the one we passed as `resume`). The SDK emits a new `system:init` with a **new** `session_id` for the forked session, but this guard evaluates to `false` and the new ID is silently dropped.

**Consequences of this bug**:
1. `session.sdkSessionId` continues pointing to the OLD session ID in memory and DB.
2. The NEXT `restart()` (e.g., another model switch, or daemon restart) will resume the OLD session, not the forked one. This reverts the entire benefit of forking.
3. The old session and the forked session diverge over time -- future resumes will go to the wrong branch.
4. Task 5's online integration test (which verifies `sdkSessionId` changes after model switch) would FAIL because the ID never gets updated.

**Fix options**:

**Option A (recommended)**: Modify the guard in `handleSystemMessage` to detect when `session_id` has changed (indicating a fork). When `forkSession: true` is used, the new `system:init` will have a different `session_id` than the current `session.sdkSessionId`. Add logic like:

```ts
if (isSDKSystemInit(message) && message.session_id) {
    if (!session.sdkSessionId || session.sdkSessionId !== message.session_id) {
        session.sdkSessionId = message.session_id;
        db.updateSession(session.id, { sdkSessionId: message.session_id });
        await daemonHub.emit('session.updated', {
            sessionId: session.id,
            source: 'sdk-session',
            session: { sdkSessionId: message.session_id },
        });
    }
}
```

This preserves the existing behavior (capture on first init) while also handling session ID changes (fork, or any future mechanism).

**Option B**: Clear `session.sdkSessionId` before starting the forked query, so the guard triggers normally. This could be done in `addSessionStateOptions()` when setting `forkSession: true`:

```ts
if (result.resume && (this.ctx.session as any)._forkOnNextQuery) {
    this.ctx.session.sdkSessionId = undefined;  // Clear so handleSystemMessage captures new ID
    result.forkSession = true;
    (this.ctx.session as any)._forkOnNextQuery = false;
}
```

This is simpler but has a subtle issue: `restart()` at line 181-195 of `query-lifecycle-manager.ts` validates the SDK session file when `session.sdkSessionId` is set. If we clear it before `restart()` runs (the flag is set in `switchModel()` but consumed in `addSessionStateOptions()` which runs during `runQuery()`), then `restart()` at line 181 would skip the validation (since `sdkSessionId` is still set at that point -- the clearing happens later in `addSessionStateOptions()`). So the timing works out. However, Option A is more robust because it handles ANY session ID change, not just forks.

The plan must include this fix as an explicit subtask in Task 1, and the acceptance criteria for Task 5 (online test verifying `sdkSessionId` changes) depend on it.

---

### P1: `_forkOnNextQuery` flag placement with `(session as any)` creates type-safety blind spots

**File**: Plan Task 1, subtask 1

The plan uses `(session as any)._forkOnNextQuery = true` and `(this.ctx.session as any)._forkOnNextQuery` throughout. While the plan acknowledges this ("or use type assertion"), it should be more explicit about which approach to take. Two alternatives are mentioned in the plan files list (`packages/shared/src/types.ts — add _forkOnNextQuery to Session interface (or use type assertion)`) but the subtask code only shows `as any`.

**Recommendation**: The plan should commit to one approach. Adding it to the `Session` interface with a `/** @internal */` JSDoc tag is preferable because:
1. It avoids `as any` casts scattered across three files
2. The TypeScript compiler will catch typos (`_forkOnNextQuer` vs `_forkOnNextQuery`)
3. The `@internal` tag signals it should not be serialized to DB or exposed to clients
4. If it accidentally leaks to the frontend, the TypeScript build will flag it rather than silently passing `undefined`

The `SessionInfo` interface is defined at `packages/shared/src/types.ts:146`. A single line addition `/** @internal */ _forkOnNextQuery?: boolean;` suffices.

---

### P2: Error handling edge case analysis is incomplete for the fork flow

**File**: Plan Task 1

The plan documents the error case where `restart()` throws after `_forkOnNextQuery` is set, noting "the flag remains but is harmless -- the next successful `startStreamingQuery()` call will see the flag and apply `forkSession: true`."

However, there is a subtlety: if `restart()` throws, the user receives an error and the model appears unchanged in the UI (error path at line 266-287 of `model-switch-handler.ts` returns `success: false`). But `session.config.model` was already updated at lines 183/221 and persisted to DB at lines 189-194/227-232. The session is now in a state where:
- DB has the NEW model
- In-memory config has the NEW model
- UI shows the OLD model (because the error handler returns `session.config.model` which IS the new model -- so actually the UI may show the NEW model briefly before the error state)

The `_forkOnNextQuery` flag being set on the NEXT successful query is actually correct behavior here -- the next query should fork with the new model since the model was already persisted. But this scenario should be explicitly called out so the implementer understands the "retry succeeds" path.

More importantly: the plan should address what happens if the user switches models MULTIPLE TIMES before any query starts. Each call to `switchModel()` sets `_forkOnNextQuery = true` (idempotent), and each successful call updates `session.config.model`. So after three rapid switches (A -> B -> C), the flag is `true` and the model is C. On the next query start, it will fork to model C with the correct resume history. This is correct, but should be documented.

---

### P2: Task 5 online test will fail without the P0 fix (sdk-message-handler.ts)

**File**: Plan Task 5

The test at subtask 1, step 5 expects:

> Calls `session.get` to verify `sdkSessionId` is still non-null but has CHANGED to a new value (the forked session ID).

As explained in the P0 above, `sdkSessionId` will NOT change because `handleSystemMessage` will not capture the new ID. The plan notes this as a risk ("the dev proxy may not fully support the forkSession flow") but the actual blocker is the `!session.sdkSessionId` guard, not dev proxy support.

The test is well-designed and would correctly verify the fork behavior -- but only after the P0 is fixed. The plan should make Task 5 depend on the P0 fix (or fold the fix into Task 1 as a subtask, making Task 5 depend on Task 1 as already specified).

---

### Previous Issues Status

| Previous Issue | Status | Notes |
|---|---|---|
| P1: restoreSession registration | **Addressed (round 3)** | Plan includes both registration paths. |
| P1: SessionCache.remove() race | **Addressed (round 3)** | Subtask 0 in Task 3 fixes the asymmetry. |
| P2: Collision analysis | **Addressed (round 4)** | Namespace invariant documented. |
| All Bug 1 issues | **No regression** | Bug 1 tasks (3, 4) are unchanged from round 4. |

---

### Summary

The `forkSession: true` approach is the correct solution for Bug 2. It preserves session resumability while creating a clean new session with the new model. However, the plan has a critical gap: the existing `sdk-message-handler.ts` will silently drop the new forked session ID because its capture guard (`!session.sdkSessionId`) prevents overwriting an existing ID. This must be fixed for the fork mechanism to work correctly across restarts and daemon reboots. The fix itself is straightforward (detect session ID changes in the guard condition), but the plan must include it as an explicit task.

lsm (commented):

## Recommendation: APPROVE

The plan has matured through six rounds of review and now addresses all previously identified issues correctly. The `forkSession: true` approach for Bug 2 is clean and well-reasoned, and the `registerSession`/`unregisterSession` approach for Bug 1 follows established patterns. All previous P0, P1, and P2 issues from round 5 have been adequately resolved.

## Previous Issues Resolution

| Previous Issue | Status | Assessment |
|---|---|---|
| P0: sdk-message-handler.ts guard `!session.sdkSessionId` drops forked session ID | **Fixed** | Guard changed to `!session.sdkSessionId \|\| session.sdkSessionId !== message.session_id`. This correctly handles both the first-init case (no existing ID) and the fork case (new ID differs from old). The extracted `isNewSession` variable improves readability. |
| P1: `(session as any)` casts for `_forkOnNextQuery` | **Fixed** | Properly typed `_forkOnNextQuery?: boolean` field on the `Session` interface with `@internal` JSDoc tag. This follows the existing pattern used for `spawnClaudeCodeProcess` at `packages/shared/src/types.ts:360`. |
| P2: Fork error handling not documented | **Fixed** | Three error scenarios are now documented: (1) `restart()` throws after flag set -- flag persists harmlessly, (2) rapid multi-switch -- idempotent, (3) SDK fork failure -- QueryRunner timeout/retry clears `sdkSessionId` and retries fresh. |
| P2: Additional edge cases | **Addressed** | The interaction between `QueryRunner`'s auto-recovery (which clears `sdkSessionId`) and the fork flag is now implicitly handled: if the fork fails, `_forkOnNextQuery` was already consumed, so the next retry starts fresh without fork. This is correct. |

## Plan Quality Assessment

### Bug 2 Fix (Tasks 1-2, 5-6)

The `forkSession: true` approach is sound:

1. **sdk-message-handler guard** (Task 1, subtask 3): The new guard is correct. After a fork, the SDK emits a `system:init` with a new `session_id`. The old guard `!session.sdkSessionId` would skip this message because the old ID is still set. The new guard `session.sdkSessionId !== message.session_id` catches this. The `isSDKSystemInit` outer guard still prevents `api_retry`, `status`, and other system subtypes from overwriting the field.

2. **_forkOnNextQuery typing** (Task 1, subtask 0): Adding the field to the `Session` interface in `packages/shared/src/types.ts` (near `sdkSessionId` at line 157) is the right location. The `@internal` JSDoc tag is consistent with the existing `spawnClaudeCodeProcess` pattern. Note: `@internal` does not have compile-time enforcement in this project (no `stripInternal` in tsconfig), so the implementer should be aware that nothing prevents external code from accessing this field. This is acceptable -- it is a documentation convention, not a security boundary.

3. **QueryOptionsBuilder integration** (Task 1, subtask 2): The flag is set after `db.updateSession` in both branches of `switchModel()`, and consumed in `addSessionStateOptions()` when both `result.resume` and `_forkOnNextQuery` are truthy. The consumption (setting to `false`) prevents the flag from leaking into subsequent queries. This is correct.

4. **Error handling**: The plan correctly identifies that if `restart()` throws, the flag remains set and will be consumed on the next successful query. This is actually the *correct* behavior -- the retry should fork with the new model. If the SDK fork itself fails, `QueryRunner`'s auto-recovery (lines 371-378 of `query-runner.ts`) clears `sdkSessionId` and retries without resume, which is a safe fallback. The `_forkOnNextQuery` flag was already consumed, so the retry does not attempt another fork. This is correct and non-cascading.

5. **Test coverage** (Task 2): The test plan covers all critical paths: flag set on switch (both branches), flag consumption in options builder, `forkSession: true` AND `resume` both present, flag not leaking to subsequent queries, error path flag persistence, and the `sdk-message-handler.ts` guard update (different session ID captured, same ID not overwritten). This is comprehensive.

### Bug 1 Fix (Tasks 3-4)

The `registerSession`/`unregisterSession` approach is well-established:

1. **SessionCache.remove() race fix** (Task 3, subtask 0): The `remove()` method at line 141 of `session-cache.ts` currently only calls `this.sessions.delete(sessionId)` but does NOT clear `this.sessionLoadLocks`. This is a real race condition. The `set()` method (line 135) correctly clears `sessionLoadLocks`, but `remove()` does not. The plan's fix is correct and necessary.

2. **Both registration paths** (Tasks 3, subtasks 2-3): Both `createAndStartSession()` and `restoreSession()` are covered. The `restoreSession()` path was a previously identified gap (round 3 P1) and is now explicitly included. This is critical -- without it, Bug 1 would reoccur after every daemon restart.

3. **Both teardown paths** (Tasks 3, subtasks 4-5): `stopSession()` (individual) and `RoomRuntimeService.stop()` (service shutdown) both unregister. The service shutdown path iterates `agentSessions.keys()` before clearing the map, which is correct.

4. **Namespace invariant** (Task 3, subtask 7): The disjoint namespace argument (`space:{uuid}:task:{uuid}` vs `{role}:{roomId}:{taskId}:{uuid}`) is valid and the documentation recommendation is reasonable.

5. **Test coverage** (Task 4): Covers `unregisterSession` delegation, `remove()` clearing `sessionLoadLocks`, the race condition scenario, registered instance preference, and the restore path. Subtask 8 (integration test) is appropriately aspirational.

### Minor Observations (P3, not blocking)

1. **PR body still not fully updated**: The PR body's "Test plan" checklist still references generic descriptions rather than the specific tasks. This is cosmetic and does not block implementation.

2. **`_forkOnNextQuery` as `@internal` vs `@remarks`**: The `@internal` tag is used consistently with the existing `spawnClaudeCodeProcess` field. However, adding a brief `@remarks` explaining why it must not be serialized would be even clearer for future maintainers.

3. **Task 5 dev proxy caveat**: The plan correctly notes that the dev proxy may not fully support `forkSession`. The fallback to unit tests (Task 2) provides adequate coverage regardless. This is a good risk assessment.

## Summary

The plan is complete, technically sound, and ready for implementation. All previously identified P0/P1/P2 issues have been resolved. The `forkSession: true` approach for Bug 2 is clean and well-integrated with existing error recovery mechanisms. The `registerSession` approach for Bug 1 follows established Space session patterns and fixes the routing for all RPC handlers, not just model switching. Test coverage is comprehensive across both bugs.

lsm (commented):

---

### Overall Assessment

The plan is well-structured and the Bug 2 pivot to an investigation-first approach is sound. The Bug 1 tasks remain solid. I have one P1 issue (stale reference pattern) and two P2 issues that should be addressed before implementation.

---

### P1: Investigation task references a deprecated test pattern (Bug 2, Task 1)

The plan references `packages/daemon/tests/online/agent/agent-session-sdk.test.ts` (lines 300-326) as the pattern to follow for subscribing to `state.sdkMessages.delta`. However, that test is inside a `describe.skip` block with the comment **"DEPRECATED - uses old SUBSCRIBE protocol"** (`agent-session-sdk.test.ts:280`). It uses the old raw WebSocket `SUBSCRIBE` message type, which is explicitly marked as needing a room-based rewrite.

The plan should instead reference `packages/daemon/tests/online/coordinator/coordinator-mode-switch.test.ts` (lines 54-97), which has a **working, non-deprecated** `waitForSystemInit()` helper that:
1. Uses `daemon.messageHub.onEvent('state.sdkMessages.delta', ...)` (the modern API)
2. Joins the session channel via `daemon.messageHub.joinChannel('session:' + sessionId)`
3. Properly parses the `added` array from the delta payload and checks `msg.type === 'system' && msg.subtype === 'init'`

This is the exact pattern the investigation task needs -- subscribing before action, reading the `model` field from the `system:init` message. The coordinator test even demonstrates the multi-phase before/after observation that Task 1 requires.

**Recommendation**: Replace the reference to `agent-session-sdk.test.ts` with `coordinator-mode-switch.test.ts` and note that the `waitForSystemInit` helper can be adapted (adding `model` extraction from the message). This prevents the implementer from copying a deprecated pattern.

---

### P2: Dependency field error in Task 3

Task 3 lists **"Dependencies: Task 3"** (plan line 195). This should be **"Dependencies: Task 2"** since Task 3 tests the implementation from Task 2. This is a self-referencing typo that will confuse the task executor.

---

### P2: Missing consideration for system:init timing relative to the model switch RPC

The plan states (Task 1, Subtask 2):
> Calls session.model.switch with a different model ... Waits for the model switch RPC to complete (the restart should happen automatically). Captures the next system:init message's model field

This sequence is correct for observing the *new* model, but there is an important subtlety: `session.model.switch` calls `lifecycleManager.restart()` which is async and includes a 100ms delay plus `validateAndRepairSDKSession`. The RPC returns after restart completes, but the `system:init` message is only emitted once the **next message is sent** and the SDK starts a new query turn. The plan acknowledges this implicitly ("Sends another message and waits for a response" in subtask 2), but the subscription for `system:init` should be set up **before** sending the post-switch message -- not before calling `session.model.switch`.

If the implementer subscribes to `state.sdkMessages.delta` *before* the model switch RPC call, the subscription might capture stale events from the switch's restart teardown or from other concurrent sessions. The safer pattern (matching `coordinator-mode-switch.test.ts`) is:

1. Call `session.model.switch` and await completion.
2. Then set up the `waitForSystemInit` subscription.
3. Then send the next message.
4. Await the `system:init` promise.

The plan's current wording ("Captures the next system:init message's model field -> postSwitchModel") is ambiguous about when the subscription is created relative to the RPC call. Clarify that the subscription should be created after the RPC completes but before the post-switch message is sent.

---

### Bug 1 Tasks (Tasks 2-3): Correct and Complete

The Bug 1 approach is solid. Key verification points:

1. **`SessionCache.remove()` race condition fix** is correct. The current code at `session-cache.ts:141-143` only deletes from `this.sessions` but not `this.sessionLoadLocks`. The `set()` method at line 132-136 already clears locks, so `remove()` should mirror this. This is a real bug that benefits existing callers like `SessionLifecycle.delete()`.

2. **`registerSession`/`unregisterSession` symmetry** on `SessionManager` is clean and follows the established pattern from `registerSession()` (line 257-259 of `session-manager.ts`).

3. **Registration paths are complete**: `createAndStartSession` (line 281) and `restoreSession` (line 411) are both covered. Missing `restoreSession` would have been a critical bug that only manifests after daemon restart.

4. **Teardown paths are complete**: `stopSession` finally block (line 473) and `RoomRuntimeService.stop()` (line 234) are both covered. The `stopSession` path includes `session.handleInterrupt()` + `session.cleanup()` before the delete, and the plan correctly places `unregisterSession` after the delete in the same `finally` block.

5. **`SessionCache.getAsync()` guard at line 99** correctly checks `!this.sessions.has(sessionId)` before inserting a DB-loaded session. This means if `registerSession` is called concurrently, the registered instance wins. This is the right behavior.

6. **Namespace invariant documentation** is a nice defensive measure. Room session IDs (`{role}:{roomId}:...`) and Space session IDs (`space:{uuid}:task:{uuid}`) are indeed structurally disjoint as long as no role is named "space".

---

### Test fix (codex bridge provider): Correct

The sync I/O bridge in `anthropic-to-codex-bridge-provider.test.ts` is a pragmatic workaround for the Bun 1.3.11 Linux CI issue. The approach (spy on `fs/promises` to delegate to `node:fs` sync counterparts) is non-invasive and the cleanup in `afterEach` is correct.

---

### Recommendation

**REQUEST_CHANGES** to fix:
1. P1: Update Task 1 to reference `coordinator-mode-switch.test.ts` instead of the deprecated `agent-session-sdk.test.ts` pattern.
2. P2: Fix Task 3 dependency from "Task 3" to "Task 2".
3. P2: Clarify subscription timing in Task 1 subtask 2.

lsm (commented):

### Verdict: APPROVE

All three issues from round 7 have been adequately addressed.

### Issue-by-issue verification

**P1 (deprecated test pattern) -- RESOLVED**: Task 1 now references `packages/daemon/tests/online/coordinator/coordinator-mode-switch.test.ts` (lines 54-97) as the canonical `waitForSystemInit()` pattern. The helper uses `daemon.messageHub.onEvent('state.sdkMessages.delta', ...)` and `daemon.messageHub.joinChannel(...)`, which is the modern API. The only remaining mention of `agent-session-sdk.test.ts` is a parenthetical note explaining it is deprecated -- this is appropriate as a "do not use" signal.

**P2 (self-referencing dependency) -- RESOLVED**: Task 3 now declares `Dependencies: Task 2` (verified at line 199 of the plan). Task 1 and Task 2 correctly declare `Dependencies: None`.

**P2 (subscription timing) -- RESOLVED**: The plan now explicitly specifies: "After the RPC returns, before sending the next message: call `waitForSystemInit(sessionId)` ... Do NOT subscribe before the RPC call -- `restart()` may emit teardown-related events that could race with the subscription." This is clear and actionable.

### Plan quality assessment

The plan is well-structured with accurate line references that match the current codebase:

- `session-cache.ts:141` -- `remove()` at line 141, confirmed
- `room-runtime-service.ts:272` -- `createAndStartSession` at line 272, confirmed
- `room-runtime-service.ts:281` -- `agentSessions.set(init.sessionId, session)` at line 281, confirmed
- `room-runtime-service.ts:398` -- `restoreSession` at line 398, confirmed
- `room-runtime-service.ts:411` -- `agentSessions.set(sessionId, session)` at line 411, confirmed
- `room-runtime-service.ts:458` -- `stopSession` at line 458, confirmed
- `room-runtime-service.ts:473` -- `agentSessions.delete(sessionId)` at line 473, confirmed
- `room-runtime-service.ts:228/234` -- `stop()` and `agentSessions.clear()` at lines 228 and 234, confirmed
- `task-agent-manager.ts:398` -- `registerSession` at line 398, confirmed
- `coordinator-mode-switch.test.ts:54-97` -- `waitForSystemInit` helper at lines 54-97, confirmed

The `SessionCache.remove()` race condition analysis is correct: clearing `sessionLoadLocks` prevents new `getAsync()` callers from joining an in-flight load after removal. The existing in-flight load can still complete and re-insert, but the timing window is narrow and the guard at `session-cache.ts:99` prevents it from overwriting a registered session.

The Bug 2 investigation task (Task 1) is well-designed: it produces evidence rather than premature implementation, and the decision point after results is clearly defined.

- The `waitForSystemInit` helper in `coordinator-mode-switch.test.ts` subscribes before `joinChannel`, but since the test will already be joined to the session channel from the initial query, this ordering difference is harmless -- `joinChannel` is idempotent and the subscription works immediately.

- Task 1 subtask 2 mentions "Before sending: call `waitForSystemInit(sessionId)` to capture the initial model field" -- but `waitForSystemInit` internally calls `joinChannel`. For the pre-switch capture, the session may not yet be joined. This is fine because `joinChannel` is explicitly documented as idempotent.

No P0/P1/P2 issues found. Only P3 nits above.

PR #870, opened 2026-03-25
fix(model-switch): clear sdkSessionId before restart to ensure fresh query

lsm:
The old SDK session file was created with the old model. When switching
models, the SDK may use the session-file model over the options model,
so the model switch wouldn't take effect.

Fix: Before calling lifecycleManager.restart(), clear sdkSessionId
and persist to DB so the new query starts completely fresh with the new
model. This applies to ALL sessions (normal and task).

Added unit tests to verify sdkSessionId is cleared in both cases:
- query not started (queryObject is null)
- query is running (queryObject exists)

### DIFF
--- a/packages/daemon/src/lib/agent/model-switch-handler.ts
+++ b/packages/daemon/src/lib/agent/model-switch-handler.ts
@@ -184,9 +184,10 @@ export class ModelSwitchHandler {
 				// newProviderInstance is guaranteed non-null here (we returned early above).
 				session.config.provider = newProviderInstance.id as Provider;
 
-				// FIX: Clear sdkSessionId before restart to ensure the new query starts fresh.
-				// The old SDK session file was created with the old model. The SDK may use the
-				// session-file model over the options model, so the model switch wouldn't take effect.
+				// FIX: Clear sdkSessionId before restart so the new query starts fresh with the
+				// new model. The old SDK session file was created with the old model. The SDK may
+				// use the session-file model over the options model, so the model switch wouldn't
+				// take effect without this.
 				session.sdkSessionId = undefined;
 
 				// Only pass serializable fields — session.config may contain runtime-only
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

This is a focused, correct fix for Bug 2: clearing `sdkSessionId` during model switch to prevent the SDK from resuming a stale session file created with the old model.

### Implementation Assessment

**Correctness (P3):** The fix is applied correctly in both branches of the model-switch logic:
- `queryObject` is null (query not started) -- line 190 of `model-switch-handler.ts`
- `queryObject` exists (query running) -- line 235 of `model-switch-handler.ts`

The operation order is correct: (1) set `session.sdkSessionId = undefined`, (2) persist via `db.updateSession()`, (3) call `lifecycleManager.restart()`. Inside `restart()`, the guard at `query-lifecycle-manager.ts:181` (`if (session.sdkSessionId)`) correctly skips because it is already undefined, and a fresh query is started without a resume parameter.

The root cause is well understood: `query-options-builder.ts:327-328` sets `options.resume = session.sdkSessionId` when it exists, causing the SDK to resume the old session file (which contains the old model). Clearing it before restart ensures a fresh session.

**Completeness:** All model switch flows go through `agent-session.ts:605` which delegates to `model-switch-handler.switchModel()`, so this single fix covers all callers (including task agents and space workflows).

**Tests:** Two new unit tests verify the fix in both scenarios. The mock session at line 165 sets `sdkSessionId: 'old-sdk-session-id'`, and both tests verify both in-memory clearing (`expect(mockSession.sdkSessionId).toBeUndefined()`) and DB persistence (`expect(updateSessionSpy).toHaveBeenCalledWith(..., { sdkSessionId: undefined })`). All 29 tests pass.

**P3 - Minor observation on duplication:** The two branches (queryObject null vs. existing) in `switchModel()` now have identical code for the sdkSessionId clear + DB update + config update + contextTracker + daemonHub emit + restart sequence. This was likely a pre-existing pattern that the fix follows faithfully, so it is not a regression. A future refactor could extract the shared logic into a private method, but that is outside the scope of this bug fix.

**No P0/P1/P2 issues found.** The fix is minimal, correct, and well-tested.

---
Recommendation: **APPROVE**

PR #871, opened 2026-03-25
fix(room): route task session model switches through RoomRuntimeService

lsm:
When session.model.switch is called for a task session (worker/leader),
SessionCache creates a NEW AgentSession via new AgentSession(...) (not
AgentSession.restore()). This new instance calls restart() ->
startStreamingQuery(), creating a new query loop. Meanwhile,
RoomRuntimeService still has its own AgentSession running with the old
model. Two concurrent query loops for the same session cause conflicts.

This fix adds room.runtime.model.get and room.runtime.model.switch
handlers to RoomRuntimeService that operate on the runtime's own
AgentSession instances. The TaskViewModelSelector now uses
room.runtime.model.switch when roomId is available, routing through
RoomRuntimeService instead of SessionManager.

Files changed:
- RoomRuntimeService: Added modelGet() and modelSwitch() methods
- room-handlers: Added room.runtime.model.get and room.runtime.model.switch handlers
- TaskViewModelSelector: Added roomId prop, uses room.runtime.model.switch when roomId is provided
- TaskInfoPanel: Added roomId prop and passes it to TaskViewModelSelector
- TaskView/TaskViewV2: Pass roomId to TaskInfoPanel
- Added unit tests for new handlers

Fixes: Bug 1 - Task view model switching not taking effect mid-task

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -140,6 +140,37 @@ export class RoomRuntimeService {
 		return this.agentSessions.get(sessionId);
 	}
 
+	/**
+	 * Get the current model for a task session.
+	 * Returns the model info or null if the session is not found.
+	 */
+	async modelGet(sessionId: string): Promise<{ currentModel: string; provider: string } | null> {
+		const session = this.agentSessions.get(sessionId);
+		if (!session) return null;
+		const sessionData = session.getSessionData();
+		return {
+			currentModel: sessionData.config.model,
+			provider: sessionData.config.provider ?? '',
+		};
+	}
+
+	/**
+	 * Switch the model for a task session.
+	 * This operates on the runtime's own AgentSession instances to avoid
+	 * creating duplicate sessions via SessionManager.
+	 */
+	async modelSwitch(
+		sessionId: string,
+		model: string,
+		provider: string
+	): Promise<{ success: boolean; model: string; error?: string }> {
+		const session = this.agentSessions.get(sessionId);
+		if (!session) {
+			return { success: false, model: '', error: 'Session not found in runtime' };
+		}
+		return session.handleModelSwitch(model, provider);
+	}
+
 	pauseRuntime(roomId: string): boolean {
 		const runtime = this.runtimes.get(roomId);
 		if (!runtime) return false;
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -505,6 +505,13 @@ export class RoomRuntimeService {
 					return false;
 				}
 			},
+			switchModel: async (sessionId, model, provider) => {
+				const session = agentSessions.get(sessionId);
+				if (!session) {
+					return { success: false, model: '', error: 'Session not found in runtime' };
+				}
+				return session.handleModelSwitch(model, provider);
+			},
 		};
 	}
 
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -39,7 +39,6 @@ import type { SessionObserver, TerminalState } from '../state/session-observer';
 import type { SessionFactory, WorkerConfig } from './task-group-manager';
 import { TaskGroupManager } from './task-group-manager';
 import type { DaemonHub } from '../../daemon-hub';
-import type { ModelSwitchResult } from '../../agent/model-switch-handler';
 import type { LeaderToolCallbacks, LeaderToolResult } from '../agents/leader-agent';
 import { createLeaderMcpServer } from '../agents/leader-agent';
 import type {
@@ -395,11 +394,11 @@ export class RoomRuntime {
 		}
 
 		try {
-			const result = (await this.messageHub?.request('session.model.switch', {
+			const result = await this.sessionFactory.switchModel(
 				sessionId,
-				model: fallback.model,
-				provider: fallback.provider,
-			})) as ModelSwitchResult | undefined;
+				fallback.model,
+				fallback.provider
+			);
 
 			if (result?.success) {
 				log.info(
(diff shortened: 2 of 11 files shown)


lsm (commented):

### Summary

The approach is correct and the implementation is clean. The fix adds modelGet() and modelSwitch() methods to RoomRuntimeService that operate on its own agentSessions cache, then exposes these via new RPC handlers (room.runtime.model.get / room.runtime.model.switch) and routes the UI to use them when a roomId is available. This avoids the root cause: SessionManager.getSessionAsync() creating a duplicate AgentSession from DB for task sessions that are owned by RoomRuntimeService.

### P1 - Server-side rate-limit fallback still uses the broken path

packages/daemon/src/lib/room/runtime/room-runtime.ts (line 398)

The rate-limit auto-fallback logic in RoomRuntime still calls session.model.switch through the MessageHub:

```ts
const result = (await this.messageHub?.request('session.model.switch', {
    sessionId,
    model: fallback.model,
    provider: fallback.provider,
})) as ModelSwitchResult | undefined;
```

This goes through SessionManager.getSessionAsync() and will hit the exact same duplicate-session bug when the fallback fires for a worker/leader session. Since RoomRuntime has access to the same SessionFactory (which owns the sessions), this path should also use SessionFactory directly (or a method on RoomRuntimeService) instead of routing through the RPC layer.

This is the same class of bug but on a different trigger -- the UI-initiated path is fixed, but the server-side rate-limit path still creates duplicates. Whether this is actively triggered depends on how often rate-limit fallback fires for task sessions.

### P2 - session.updated event not emitted on successful switch

packages/daemon/src/lib/rpc-handlers/room-handlers.ts (lines 309-333)

The original session.model.switch handler in session-handlers.ts emits a session.updated event on the session channel after a successful switch. The new room.runtime.model.switch handler does NOT emit this event.

After investigation, I confirmed this is acceptable for now -- RoomStore filters on event.status !== undefined and the model-only event has no status field, so it would be ignored anyway. The useTaskViewData hook only refreshes on room.task.update events, not session.updated.

However, if any future listener starts relying on session.updated with model data, this will be a subtle gap. Consider emitting the event for parity, or add a comment documenting the intentional omission.

### P2 - Model label in TaskInfoPanel won't update after switch

packages/web/src/components/room/TaskInfoPanel.tsx (lines 269-288)

The currentModel prop passed to TaskViewModelSelector comes from workerSession.config.model / leaderSession.config.model, which are populated by useTaskViewData via session.get RPC. Neither the onModelSwitched callback nor any session.updated listener triggers a re-fetch of the session info.

This means after a successful model switch via the info panel dropdown, the model label shown in the UI will continue to display the old model name until the next room.task.update event happens to trigger a data refresh. This is a pre-existing issue (the old session.model.switch path had the same problem), but since this PR touches the model switching path, it would be a good opportunity to fix it by either:
- Passing and using the onModelSwitched callback from TaskInfoPanel
- Subscribing to session.updated events in useTaskViewData

### Verified: Tests pass

- packages/daemon/tests/unit/rpc/room-runtime-model-handlers.test.ts -- 9/9 pass
- packages/daemon/tests/unit/rpc/room-handlers.test.ts -- 24/24 pass
- packages/web/src/components/room/TaskView.test.tsx -- 95/95 pass

### Verified: Approach correctness

- The RoomRuntimeService.modelSwitch() correctly delegates to the runtime's own AgentSession.handleModelSwitch(), avoiding the SessionCache path entirely.
- The fallback to session.model.switch when roomId is absent ensures non-room sessions (regular chat sessions) continue to work.
- Input validation in the RPC handlers is consistent with the existing pattern in setupRoomRuntimeHandlers.

lsm (commented):

### Summary

The second commit (862d372) adequately addresses the three issues from the previous review. The code is clean, type-safe, and all tests pass (2962 tests across room + space suites). No regressions detected.

### P1 Fix: Rate-limit fallback path (switchModel on SessionFactory)

**Verified: Correct.** The `RoomRuntime.handleRateLimitFallback()` at `packages/daemon/src/lib/room/runtime/room-runtime.ts:397` now calls `this.sessionFactory.switchModel(sessionId, model, provider)` directly. The `SessionFactory` interface at `packages/daemon/src/lib/room/runtime/task-group-manager.ts:122-126` now has a required `switchModel()` method. The real implementation in `packages/daemon/src/lib/room/runtime/room-runtime-service.ts:508-514` delegates to `session.handleModelSwitch()` on the runtime's own `AgentSession` instance, completely bypassing SessionManager and the duplicate-session risk. The stub in `packages/daemon/src/lib/rpc-handlers/index.ts:458-462` correctly returns a failure result for the global session factory (which doesn't manage room runtime sessions).

### P2 Fix: session.updated event emission

**Verified: Correct.** The `room.runtime.model.switch` handler at `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:334-338` now emits `session.updated` with `{ sessionId, model }` on channel `session:${sessionId}`. This is consistent with how `session.model.switch` (at `session-handlers.ts:455-459`) emits on the same channel. The event reaches the client because `ReadonlySessionChat` joins `session:${sessionId}` for worker/leader sessions, and client-side event dispatch is method-based (not channel-filtered).

### P2 Fix: useTaskViewData subscription

**Verified: Correct.** The hook at `packages/web/src/hooks/useTaskViewData.ts:200-215` subscribes to `session.updated` events and updates `workerSession`/`leaderSession` state by matching `event.sessionId` against tracked `currentSessionIds`. The cleanup properly unsubscribes both the task update and session update handlers (lines 219-220).

### Issues

**P2: Missing test coverage for session.updated event emission**

The test file `packages/daemon/tests/unit/rpc/room-runtime-model-handlers.test.ts` has a mock for `messageHub.event()` (line 32) but never asserts that it was called with the correct arguments after a successful model switch. Since the P2 fix adds this emission, there should be at least one test that verifies:
- `messageHub.event` was called with method `'session.updated'`
- The data includes `{ sessionId: params.sessionId, model: result.model }`
- The channel option is `{ channel: `session:${params.sessionId}` }`

Without this test, a future change could silently break the event emission without detection.

**P3: Event payload inconsistency between handlers**

The existing `session.model.switch` handler at `session-handlers.ts:456` emits `session.updated` with `{ model: result.model }` (no `sessionId` in the payload). The new `room.runtime.model.switch` handler emits `{ sessionId: params.sessionId, model: result.model }` (includes `sessionId`). This inconsistency is not a bug since the client-side `useTaskViewData` specifically checks `event.sessionId`, which will be `undefined` for events from `session.model.switch` (harmlessly filtered by the `event.model` guard). However, it would be cleaner for both handlers to use the same payload shape. This is purely cosmetic.

### Verdict

All three original issues are resolved. The one actionable item is the missing test assertion for event emission (P2), which I'd recommend adding but is not blocking since the logic is straightforward and the event pattern is well-established in the codebase.

PR #872, opened 2026-03-25
feat(mcp): wire per-room enablement into AppMcpLifecycleManager (Task 4.2)

lsm:
Implement getEnabledMcpConfigsForRoom() to use the room_mcp_enablement
table for per-room filtering, falling back to global registry defaults
when no room-specific overrides exist.

Changes:
- AppMcpLifecycleManager.getEnabledMcpConfigsForRoom(): query
  roomMcpEnablement.getEnabledServers(roomId) for per-room overrides;
  if empty, fall back to global listEnabled() but exclude any servers
  explicitly disabled for that room via getOverride()
- RoomRuntimeService.setupRoomAgentSession(): call
  getEnabledMcpConfigsForRoom(room.id) instead of getEnabledMcpConfigs()
- mcp.registry.changed handler: call getEnabledMcpConfigsForRoom(roomId)
  per-room instead of global getEnabledMcpConfigs()
- Unit tests: update appMcpLifecycleManager tests with 6 new cases for
  per-room enablement; update room-runtime-service-mcp tests to mock
  getEnabledMcpConfigsForRoom
- Online tests: add room-mcp-enablement.test.ts verifying
  mcp.room.setEnabled/getEnabled/resetToGlobal RPC behavior

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -205,6 +205,9 @@ jobs:
           - module: rewind
             test_path: tests/online/rewind
             mock_sdk: true
+          - module: room-mcp
+            test_path: tests/online/room/room-mcp-enablement.test.ts
+            mock_sdk: true
           # Room tests temporarily disabled during refactoring
           # - module: room-1
           #   test_path: >-
--- a/packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts
+++ b/packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts
@@ -69,12 +69,49 @@ export class AppMcpLifecycleManager {
 	/**
 	 * Returns SDK MCP configs for a specific room.
 	 *
-	 * Stub: Milestone 4 will add the `room_mcp_enablement` table and per-room
-	 * filtering. Until then this falls back to the global enabled set.
+	 * Per-room overrides take precedence:
+	 * - If the room has explicitly enabled servers (via room_mcp_enablement), return those.
+	 * - If the room has no overrides, fall back to globally-enabled servers, but exclude
+	 *   any that are explicitly disabled for this room via per-room override.
 	 */
-	getEnabledMcpConfigsForRoom(_roomId: string): Record<string, McpServerConfig> {
-		// TODO (Milestone 4): Query room_mcp_enablement for roomId and filter accordingly.
-		return this.getEnabledMcpConfigs();
+	getEnabledMcpConfigsForRoom(roomId: string): Record<string, McpServerConfig> {
+		const roomServers = this.db.roomMcpEnablement.getEnabledServers(roomId);
+
+		// If the room has per-room enabled servers, return those (filtered by validation).
+		if (roomServers.length > 0) {
+			const result: Record<string, McpServerConfig> = {};
+			for (const entry of roomServers) {
+				const validation = this.validateEntry(entry);
+				if (!validation.valid) {
+					continue;
+				}
+				result[entry.name] = this.convertEntry(entry);
+			}
+			return result;
+		}
+
+		// No per-room enabled servers. Fall back to global enabled set, but exclude
+		// any servers that are explicitly disabled for this room.
+		const globalServers = this.db.appMcpServers.listEnabled();
+		const result: Record<string, McpServerConfig> = {};
+
+		for (const entry of globalServers) {
+			// Check if this server is explicitly disabled for the room
+			const override = this.db.roomMcpEnablement.getOverride(roomId, entry.id);
+			if (override !== null && !override.enabled) {
+				// Server is explicitly disabled for this room — skip it
+				continue;
+			}
+
+			const validation = this.validateEntry(entry);
+			if (!validation.valid) {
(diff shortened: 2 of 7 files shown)


lsm (commented):

**Recommendation: APPROVE** (non-blocking issues only)

---

### Summary

The implementation is well-structured and correctly wires per-room MCP enablement into `AppMcpLifecycleManager` and `RoomRuntimeService`. The core logic in `getEnabledMcpConfigsForRoom()` is sound, the two integration points (initial setup and `mcp.registry.changed` hot-reload) are updated consistently, and the unit tests provide good coverage of the branching behavior. No P0 or P1 issues found.

### Details

**Core Implementation (P3)**

The two-branch logic in `getEnabledMcpConfigsForRoom()` at `packages/daemon/src/lib/mcp/app-mcp-lifecycle-manager.ts:77-115` is clean and correct:
- If the room has explicitly enabled servers (`roomServers.length > 0`), return only those.
- Otherwise fall back to the global set, excluding servers that have explicit `enabled=false` overrides.

This correctly models the intended UX: setting any per-room override opts the room into explicit mode; `resetToGlobal()` clears all overrides to restore the default.

**Worker sessions intentionally use global configs** -- `createSessionFactory()` at line 276 still calls `getEnabledMcpConfigs()` (global), which is correct per the existing comment about worker sessions being short-lived.

**Test Quality (P2)**

1. **"invalid entries" test passes for the wrong reason** (`packages/daemon/tests/unit/mcp/app-mcp-lifecycle-manager.test.ts:489-510`): The test sets only `validServer` as a per-room override, so `getEnabledServers()` returns a list of length 1, causing the per-room branch to execute. The `broken-server` is never in the per-room list at all (it was not enabled for the room), so asserting it's undefined doesn't actually validate the `validateEntry()` check in the per-room path. To properly test this, enable the broken-server for the room too (e.g., `db.roomMcpEnablement.setEnabled('room-invalid', brokenServer.id, true)`) so that validation is actually invoked on it within the per-room loop.

2. **Mock `getEnabledMcpConfigsForRoom` doesn't verify roomId argument** (`packages/daemon/tests/unit/room/room-runtime-service-mcp.test.ts`): All mocks use `() => ({...})` (no parameter capture). This means the mocks don't verify that the correct `roomId` is being passed. The tests still validate the output correctly via `setRuntimeMcpServersCalls`, so this isn't a correctness gap, but capturing and asserting the `roomId` argument would make the tests more robust against regressions like accidentally using a hardcoded or wrong room ID.

**Online Tests (P3)**

The online test at `packages/daemon/tests/online/room/room-mcp-enablement.test.ts` only exercises the RPC handler + repository layer (using fake room IDs like `'room-enable-test'` that don't correspond to actual rooms). It does not test the `RoomRuntimeService` integration path. The `mcp.registry.changed event` test (line 134-147) is effectively a no-op assertion -- it checks that data persisted correctly but never verifies the event was emitted or consumed by `RoomRuntimeService`. The test file header also incorrectly states "Makes real API calls (costs money, uses rate limits)" but none of these tests make AI API calls. These are not blocking issues but worth improving in a follow-up.

**RPC Validation (P3, pre-existing)**

The `mcp.room.setEnabled` handler at `packages/daemon/src/lib/rpc-handlers/app-mcp-handlers.ts:201-217` validates that the server exists but does NOT validate that the roomId corresponds to a real room. This was inherited from the Task 4.1 PR and is out of scope for this PR, but worth tracking for hardening.

**All unit tests pass**: 24/24 in `app-mcp-lifecycle-manager.test.ts`, 8/8 in `room-runtime-service-mcp.test.ts`.

lsm (commented):

Re-reviewed after the P2 fix (commit `58024de91`).

### P2 Fix Verification: Correct

The fix at `packages/daemon/tests/unit/mcp/app-mcp-lifecycle-manager.test.ts` (lines 496-506) properly addresses the test quality issue:

**Before fix**: `brokenServer` was never enabled for the room, so `getEnabledServers('room-invalid')` only returned `validServer`. The per-room validation path (the `for (const entry of roomServers)` loop at `app-mcp-lifecycle-manager.ts:83-89`) was never exercised on `brokenServer`. The test passed trivially.

**After fix**: Both `validServer` and `brokenServer` are now enabled for the room via `db.roomMcpEnablement.setEnabled('room-invalid', brokenServer.id, true)`. This means `getEnabledServers` returns both entries, the per-room loop iterates over both, `validateEntry` is called on `brokenServer` (which fails due to missing `command`), and it is correctly excluded from the result.

The comment explaining the fix rationale is clear and accurate. All 24 unit tests in the file pass, and all 8 room-runtime-service-mcp tests pass.

### Overall PR Assessment

The implementation of `getEnabledMcpConfigsForRoom` in `app-mcp-lifecycle-manager.ts` is solid:
- Per-room overrides take precedence over global defaults (line 81-91)
- Fallback to global with per-room disable filtering works correctly (line 94-114)
- The `mcp.registry.changed` handler now correctly reads per-room configs per room (room-runtime-service.ts:704-711)

No remaining issues. **Recommendation: APPROVE**.

PR #873, opened 2026-03-25
feat(web): add MCP registry API helpers and LiveQuery store (Task 5.1)

lsm:
## Summary

Task 5.1: API Helpers and State Management for MCP registry.

### Changes

**API Helpers (`packages/web/src/lib/api-helpers.ts`)**
- 8 new functions wrapping MCP registry RPC endpoints:
  - `listAppMcpServers()` → `mcp.registry.list`
  - `createAppMcpServer(req)` → `mcp.registry.create`
  - `updateAppMcpServer(id, updates)` → `mcp.registry.update`
  - `deleteAppMcpServer(id)` → `mcp.registry.delete`
  - `setAppMcpServerEnabled(id, enabled)` → `mcp.registry.setEnabled`
  - `getRoomMcpEnabled(roomId)` → `mcp.room.getEnabled`
  - `setRoomMcpEnabled(roomId, serverId, enabled)` → `mcp.room.setEnabled`
  - `resetRoomMcpToGlobal(roomId)` → `mcp.room.resetToGlobal`
- Uses shared response types (`McpRegistryListResponse`, etc.) for type safety

**AppMcpStore (`packages/web/src/lib/app-mcp-store.ts`)**
- LiveQuery-based reactive store for global MCP server registry
- Subscribes to `mcpServers.global` named query on `subscribe()`
- Populates `appMcpServers` signal from snapshot, updates via delta events
- Re-subscribes on WebSocket reconnect via `hub.onConnection`
- Does NOT use `mcp.registry.changed` (LiveQuery supersedes it)
- Stale-event guard prevents stale events from updating signals after unsubscribe
- Post-await race guard prevents dangling handlers when unsubscribe races with hub resolution
- Error propagation via `subscribe()` rejection + `error` signal

**Tests (`packages/web/src/lib/__tests__/app-mcp-store.test.ts`)**
- 20 unit tests covering:
  - subscribe/unsubscribe idempotency
  - LiveQuery snapshot + delta (add/remove/update)
  - subscriptionId filtering
  - WebSocket reconnect re-subscription
  - stale-event guard behavior
  - post-await unsubscribe race guard
  - error signal propagation
  - API helper type verification
  - ConnectionNotReadyError instanceof check

## Test plan

- [x] Unit tests pass (20 tests)
- [x] TypeScript typecheck passes
- [x] Lint passes
- [x] Knip dead-code check passes

### DIFF
--- a/packages/web/src/lib/__tests__/app-mcp-store.test.ts
+++ b/packages/web/src/lib/__tests__/app-mcp-store.test.ts
@@ -7,10 +7,14 @@
  * - WebSocket reconnect re-subscribes automatically
  * - unsubscribe() calls liveQuery.unsubscribe and resets state
  * - Idempotent subscribe/unsubscribe behavior
+ * - Stale-event guard discards events after unsubscribe
+ * - Post-await unsubscribe race guard prevents dangling handlers
+ * - Error propagation via subscribe() rejection and error signal
  */
 
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
 import type { AppMcpServer, LiveQuerySnapshotEvent, LiveQueryDeltaEvent } from '@neokai/shared';
+import { McpRegistryListResponse, McpRegistryCreateResponse } from '@neokai/shared';
 
 // ---------------------------------------------------------------------------
 // Mocks
@@ -102,11 +106,10 @@ describe('AppMcpStore', () => {
 		vi.clearAllMocks();
 		mockHub = createMockHub();
 
-		// Reset module-level state by clearing the store
+		// Reset store signals
 		appMcpStore.appMcpServers.value = [];
 		appMcpStore.loading.value = false;
-		// Note: we cannot fully reset private state, so tests that need a fresh
-		// store should be in separate describe blocks or use afterEach cleanup.
+		appMcpStore.error.value = null;
 
 		vi.mocked(connectionManager.getHub).mockResolvedValue(mockHub as never);
 		vi.mocked(connectionManager.getHubIfConnected).mockReturnValue(mockHub as never);
@@ -189,6 +192,14 @@ describe('AppMcpStore', () => {
 			// No new request should have been made
 			expect(mockHub.request).not.toHaveBeenCalled();
 		});
+
+		it('should set error signal and re-throw when hub subscription fails', async () => {
+			vi.mocked(mockHub.request).mockRejectedValue(new Error('subscribe failed'));
+
+			await expect(appMcpStore.subscribe()).rejects.toThrow('subscribe failed');
+			expect(appMcpStore.error.value).toBe('subscribe failed');
+			expect(appMcpStore.loading.value).toBe(false);
+		});
 	});
 
 	// ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Summary

The implementation delivers all the requested pieces: 8 MCP API helpers in `api-helpers.ts`, a reactive `AppMcpStore` class using LiveQuery in `app-mcp-store.ts`, and 23 unit tests. The code follows existing patterns, types match the shared definitions, and no use of `mcp.registry.changed` event is present in the frontend. The tests all pass, linting is clean, and typecheck succeeds.

Overall this is solid work. I have a few items to flag, mostly around resilience and pattern consistency with the more battle-tested `RoomStore`.

---

### P1 (should-fix)

**1. Missing stale-event guard in `app-mcp-store.ts` (lines 52-75)**

The `RoomStore` LiveQuery subscriptions use an `activeSubscriptionIds` Set as a stale-event guard — `activeSubscriptionIds.delete(subId)` is called immediately in `unsubscribeRoom` before handlers are removed, so any events already queued in the JS event loop from a previous subscription are discarded (see `room-store.ts` lines 243-244, 348-349, 443-446).

`AppMcpStore` lacks this guard. If `unsubscribe()` is called and then a stale `liveQuery.snapshot` or `liveQuery.delta` event arrives between the teardown and the next microtask, the old handler could briefly set signal state on the store before being cleaned up. While the singleton pattern and idempotent subscribe mitigate the practical risk, this is a correctness gap compared to the established pattern.

**File:** `packages/web/src/lib/app-mcp-store.ts` lines 52-75

**2. Missing guard after `await hub.request('liveQuery.subscribe', ...)` in `subscribe()` (line 98-102)**

`RoomStore.subscribeRoom` checks `this.liveQueryActive.has(roomId)` after each `await hub.request(...)` call and runs cleanup if the room was unsubscribed while the request was in flight (see `room-store.ts` lines 398-408, 482-495). This prevents handlers from staying registered after an unsubscribe race.

`AppMcpStore.subscribe()` does not check `this.subscribed` after the `await hub.request(...)` on line 98. If `unsubscribe()` is called while the subscribe request is pending, the registered snapshot/delta/reconnect handlers remain dangling — they reference a store that has been reset. The error catch on line 103 resets `this.subscribed = false`, but the handlers are not cleaned up there.

**File:** `packages/web/src/lib/app-mcp-store.ts` lines 97-108

**3. `subscribe()` silently swallows errors — no retry, no signal for callers**

If `connectionManager.getHub()` rejects or `hub.request('liveQuery.subscribe', ...)` throws, the catch block (line 103) logs and sets `subscribed = false` + `loading = false`. The caller receives no indication of failure because `subscribe()` returns `Promise<void>` and the catch does not re-throw. This means a component calling `await appMcpStore.subscribe()` has no way to know the subscription actually failed. In contrast, `RoomStore.startSubscriptions` re-throws and sets `this.error.value`.

**File:** `packages/web/src/lib/app-mcp-store.ts` lines 103-108

---

### P2 (important suggestion)

**4. API helpers do not use shared response types — inline type literals instead**

The shared package exports specific response types for all MCP registry endpoints (`McpRegistryListResponse`, `McpRegistryCreateResponse`, `McpRegistryUpdateResponse`, `McpRegistryDeleteResponse`, `McpRegistrySetEnabledResponse`, `McpRoomGetEnabledResponse`, `McpRoomSetEnabledResponse`, `McpRoomResetToGlobalResponse` in `packages/shared/src/api.ts`). The API helpers in `api-helpers.ts` use inline type literals (e.g., `{ servers: AppMcpServer[] }`) instead of importing these types. If the daemon response shape ever changes, the shared types would be updated but the frontend helpers would silently drift.

This is not a bug today — the shapes match — but it defeats the purpose of having the shared response types as the single source of truth.

**File:** `packages/web/src/lib/api-helpers.ts` lines 255-316

**5. `updateAppMcpServer` conditional type is correct but unnecessarily opaque**

The conditional type `UpdateAppMcpServerRequest['id'] extends string ? Omit<UpdateAppMcpServerRequest, 'id'> : never` (lines 271-273) evaluates to `Omit<UpdateAppMcpServerRequest, 'id'>` which is `Partial<Omit<AppMcpServer, 'id'>>`. While technically correct, a reviewer or future developer needs to mentally evaluate the conditional. Since `UpdateAppMcpServerRequest.id` is literally typed as `string`, this conditional can never be `never`. A direct type alias would be clearer.

**File:** `packages/web/src/lib/api-helpers.ts` lines 271-273

**6. Test for `ConnectionNotReadyError` uses a generic string match (line 485)**

The test `await expect(listAppMcpServers()).rejects.toThrow('Not connected to server')` checks the error message string. A more robust assertion would check `instanceof ConnectionNotReadyError`, which would catch regressions where a different error type is thrown with the same message.

**File:** `packages/web/src/lib/__tests__/app-mcp-store.test.ts` line 485

---

### P3 (nit)

**7. Comment in test file acknowledges uncleanable private state (lines 108-109)**

The comment on lines 108-109 says we cannot fully reset private state. This works because `afterEach` calls `appMcpStore.unsubscribe()`, but it would be cleaner to export a `_resetForTesting()` method or use a factory function pattern so tests are fully isolated.

**File:** `packages/web/src/lib/__tests__/app-mcp-store.test.ts` lines 108-109

---

### Positive observations

- All 8 API helper functions match the daemon RPC handler signatures exactly.
- Response types are structurally correct against the shared definitions.
- The LiveQuery subscription pattern (snapshot -> delta -> reconnect -> unsubscribe) is well-structured and follows the right mental model.
- No reference to `mcp.registry.changed` event in the frontend — the store correctly relies solely on LiveQuery.
- Test coverage is good: subscribe, idempotency, snapshot, delta (add/remove/update), subscriptionId filtering, reconnect, unsubscribe, and API helper types.
- Full typecheck, lint, and all 23 tests pass.

---

### Recommendation

REQUEST_CHANGES — The three P1 items (stale-event guard, post-await race guard, and silent error swallowing) should be addressed to bring `AppMcpStore` to the same resilience level as `RoomStore`'s LiveQuery subscriptions.

lsm (commented):

All seven P1/P2 issues from the first review have been addressed. The code quality is solid and the implementation follows the established patterns from `room-store.ts`. However, one new P1 issue was introduced by the fix.

### P1: Error-path handler leak in subscribe() catch block

**File:** `packages/web/src/lib/app-mcp-store.ts`, lines 130-136

When `hub.request('liveQuery.subscribe', ...)` at line 119 rejects, the catch block (line 130) sets `this.subscribed = false` and throws, but does **not** clean up the event handlers that were already registered on the hub (snapshot at line 69, delta at line 79, reconnect at line 101). It also does not clear `activeSubscriptionIds`, which was populated at line 66.

This means after a failed `subscribe()`:
1. Three event handlers remain registered on the hub (memory leak).
2. `activeSubscriptionIds` still contains `SUBSCRIPTION_ID`, so if the WebSocket reconnects later, the orphaned reconnect handler (line 101) would fire and attempt another `liveQuery.subscribe` request -- even though `this.subscribed` is `false`.
3. If the user calls `subscribe()` again after failure, new handlers are registered alongside the leaked ones, causing duplicate event processing.

The established pattern in `room-store.ts` (lines 524-543) handles this correctly: the catch block runs all cleanup functions that were registered before the error to remove handlers and clear `activeSubscriptionIds`. The `AppMcpStore` catch block should do the same by calling `teardownCleanly()`:

```typescript
} catch (err) {
    this.subscribed = false;
    this.teardownCleanly();  // <-- add this to clean up handlers + activeSubscriptionIds
    this.error.value = err instanceof Error ? err.message : 'Failed to subscribe to MCP registry';
    logger.error('Failed to subscribe AppMcpStore LiveQuery:', err);
    throw err;
}
```

Note: `teardownCleanly()` already resets `this.loading.value = false` and `this.error.value = null`, so you would need to set `this.error.value` **after** the `teardownCleanly()` call. The current code already does this correctly (loading reset is at line 132, but `teardownCleanly()` also does it -- so the explicit `this.loading.value = false` on line 132 would become redundant).

### Verified fixes (all properly addressed)

**P1: Stale-event guard** -- `activeSubscriptionIds` Set is populated at line 66, checked in both snapshot (line 71) and delta (line 81) handlers, and cleared immediately in `unsubscribe()` (line 174). Matches the established pattern in `room-store.ts`. Tests at lines 284-331 verify both snapshot and delta events are correctly discarded after unsubscribe.

**P1: Post-await unsubscribe race guard** -- The `if (!this.subscribed)` check at line 126 after `hub.request()` resolves correctly calls `teardownCleanly()` to prevent dangling handlers. Test at lines 338-371 races `unsubscribe()` against a deferred `hub.request()` resolution.

**P1: Error propagation** -- `subscribe()` now re-throws errors (line 135) and sets the `error` signal (line 133), allowing callers to display toasts. Test at lines 196-202 verifies both behaviors.

**P2: Shared response types** -- All 8 API helpers use the shared response types from `@neokai/shared` (`McpRegistryListResponse`, `McpRegistryCreateResponse`, etc.) instead of inline type literals. Tests also use `McpRegistryListResponse` and `McpRegistryCreateResponse` for mock responses.

**P2: Simplified updateAppMcpServer type** -- Changed from the conditional type `UpdateAppMcpServerRequest['id'] extends string ? Omit<...> : never` to the much cleaner `Omit<UpdateAppMcpServerRequest, 'id'>`, which correctly resolves to `Partial<Omit<AppMcpServer, 'id'>>`.

**P2: instanceof assertion** -- Test at line 600 now uses `rejects.toBeInstanceOf(ConnectionNotReadyError)` with a proper import instead of `rejects.toThrow('Not connected to server')`.

**P2: unsubscribe() resets error** -- `unsubscribe()` now resets `error.value = null` even when called while not subscribed (lines 164-168), which handles the case where `subscribe()` previously failed and left an error. Test at lines 439-445 verifies this.

### Summary

The implementation is well-structured and closely follows the existing `room-store.ts` patterns. The only remaining issue is the error-path cleanup gap, which is directly analogous to a case that `room-store.ts` already handles correctly. Once that one-line fix is applied, this PR is ready.

lsm (commented):

### Verdict: APPROVE

All P1 and P2 issues from previous rounds have been properly addressed. The implementation is clean, well-tested, and ready to merge.

### Round 2 P1 Fix Verification: subscribe() catch block cleanup

The catch block in `subscribe()` (`app-mcp-store.ts:130-136`) now correctly calls `teardownCleanly()` after setting `this.subscribed = false`. This ensures:

1. **Event handlers are removed** -- `teardownCleanly()` runs all cleanup functions in `this.cleanups[]`, which includes `unsubSnapshot`, `unsubDelta`, and `unsubReconnect`. These are the unsubscribe functions returned by `hub.onEvent()` and `hub.onConnection()`, so they remove the handlers from the hub's internal listener arrays.

2. **No orphaned reconnect handlers on retry** -- The reconnect handler (registered via `hub.onConnection`) is cleaned up via `unsubReconnect` in the catch path. When `subscribe()` is retried after a failure, a fresh reconnect handler is registered without any stale ones remaining.

3. **activeSubscriptionIds is cleared** -- `teardownCleanly()` deletes the subscription ID at line 144, and there is also a dedicated cleanup function `() => this.activeSubscriptionIds.delete(SUBSCRIPTION_ID)` pushed at line 76. The double-delete is safe (Set.delete is idempotent) and ensures the stale-event guard prevents any in-flight events from being processed.

### New Test Verification

The test `'should clean up handlers and clear activeSubscriptionIds on subscribe failure'` (test file lines 204-224) correctly exercises the retry-after-failure scenario:
- First `subscribe()` fails (mock rejects `hub.request`).
- Mock is reset to succeed.
- Second `subscribe()` succeeds and fires a snapshot with one server.
- Asserts exactly 1 server in `appMcpServers.value` -- proving no leaked handlers from the failed first attempt caused duplicate processing.

This is the right way to test for handler leaks: the mock hub tracks handlers in arrays, and if the first subscribe's handlers were not cleaned up, the snapshot event would fire twice (once per leaked handler), resulting in duplicate entries.

### Additional Observations (no action needed)

- **Post-await race guard** (lines 126-128) remains correct from the previous round -- it detects `!this.subscribed` after the `hub.request` await and calls `teardownCleanly()`.
- **Error signal reset** -- `teardownCleanly()` resets `error.value = null` at line 154, which is then overwritten by the catch block at line 133. This is harmless and correct behavior; `teardownCleanly()` is a shared utility and the catch block intentionally sets the error afterward.
- All 29 tests pass.
- API helpers use proper shared types from `@neokai/shared`.

PR #874, opened 2026-03-25
feat(web): add per-room MCP enablement UI in RoomSettings (Task 5.3)

lsm:
- Add RoomMcpStore class for per-room MCP enablement LiveQuery subscriptions
- Add MCP Servers section to RoomSettings.tsx with:
  - Per-server toggles using effective enabled state (override or global default)
  - Room override badges for servers with per-room overrides
  - Reset to Global Defaults button
  - Empty state with link to global settings
- Add Vitest unit tests for RoomMcpStore (22 tests)
- Add Vitest unit tests for RoomSettings MCP section (17 tests)

### DIFF
--- a/packages/daemon/src/lib/providers/anthropic-to-codex-bridge-provider.ts
+++ b/packages/daemon/src/lib/providers/anthropic-to-codex-bridge-provider.ts
@@ -496,12 +496,17 @@ export class AnthropicToCodexBridgeProvider implements Provider {
 		};
 	}
 
-	/** Stop all bridge servers. Called at provider shutdown (e.g. tests). */
+	/** Stop all bridge servers and reset cached auth state. Called at provider shutdown (e.g. tests). */
 	stopAllBridgeServers(): void {
 		for (const server of this.bridgeServers.values()) {
 			server.stop();
 		}
 		this.bridgeServers.clear();
+		// Reset cached auth so a new provider instance starts with a clean slate.
+		// Without this, cachedBridgeAuth=null from a previous run would cause
+		// isAvailable() to return false even when valid credentials are present.
+		this.cachedBridgeAuth = undefined;
+		this.cachedApiKey = undefined;
 	}
 
 	/** @deprecated Use stopAllBridgeServers(). */
--- a/packages/web/src/components/room/RoomSettings.tsx
+++ b/packages/web/src/components/room/RoomSettings.tsx
@@ -11,6 +11,9 @@ import { useSignal } from '@preact/signals';
 import { useEffect, useState } from 'preact/hooks';
 import type { Room, WorkspacePath } from '@neokai/shared';
 import { connectionManager } from '../../lib/connection-manager';
+import { appMcpStore } from '../../lib/app-mcp-store';
+import { roomMcpStore } from '../../lib/room-mcp-store';
+import { setRoomMcpEnabled, resetRoomMcpToGlobal } from '../../lib/api-helpers';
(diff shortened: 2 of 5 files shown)


lsm (commented):

---

### Summary

Overall this is a well-structured implementation that follows the established patterns (particularly the `AppMcpStore` from Task 5.1). The code is clean, tests are comprehensive (39 tests, all passing), typecheck and lint pass cleanly, and the integration with the daemon-side LiveQuery/RPC handlers is correct. I am recommending APPROVE with a few observations below that are worth considering but not blocking.

---

### P1 (should-fix) -- 0 issues

No P1 issues found.

### P2 (important suggestion) -- 1 issue

**1. Singleton `roomMcpStore` limitation for concurrent room settings views**

`packages/web/src/lib/room-mcp-store.ts` (line 227)

The `roomMcpStore` is exported as a module-level singleton, meaning only one room can be subscribed at a time. The `subscribe()$ method handles room switching correctly (unsubscribes old room, subscribes new room at line 69-71), so this works when the user navigates between rooms sequentially. However, if the UI ever needs to display MCP settings for two rooms simultaneously (e.g., split view, modal comparison), this singleton will cause a conflict -- the second subscription would unsubscribe the first.

This is not a problem today since `RoomSettings` is shown for a single room at a time. But it is worth documenting this constraint in the JSDoc comment on the singleton export, e.g.:

```ts
/** Singleton store instance for room MCP enablement.
 *  Supports subscribing to one room at a time. Switching rooms
 *  automatically unsubscribes the previous room. */
export const roomMcpStore = new RoomMcpStore();
```

This prevents future developers from being surprised by the limitation.

---

### P3 (nit) -- 2 issues

**1. Loading state test relies on mocked module re-import**

`packages/web/src/components/room/__tests__/RoomSettings-mcp.test.tsx` (lines 183-188)

The test at line 182 re-imports the mocked modules to access `loading.value`, but since `appMcpStore` and `roomMcpStore` are already mocked (lines 46-70), the dynamic import returns the mock objects. The test could directly use the hoisted mock references instead of the re-import pattern, making the test intent clearer. This is purely a readability suggestion.

**2. Empty state link has a TODO with no issue tracker reference**

`packages/web/src/components/room/RoomSettings.tsx` (line 559)

The `onClick` handler for the "Add MCP servers in global settings" link contains a `// TODO: Navigate to global settings - this would require router integration` comment. Consider filing a follow-up issue for this navigation so the TODO does not get lost.

---

### What was done well

- **Faithful pattern replication**: `RoomMcpStore$ follows the exact same architecture as `AppMcpStore$ (stale-event guard, reconnect handler, teardown lifecycle), which is consistent and maintainable.
- **Robust stale-event guard**: The `activeSubscriptionIds$ set is checked in both snapshot and delta handlers, and cleared before handler teardown in `unsubscribe()$ to prevent in-flight events from corrupting state.
- **Correct race condition handling**: The `subscribed` guard after `await connectionManager.getHub()$ (line 82) and after `await hub.request('liveQuery.subscribe', ...)$ (line 152) correctly handles the case where `unsubscribe()$ is called while operations are in flight.
- **Thorough test coverage**: 22 store tests covering subscribe/unsubscribe/idempotency/delta/stale-guard/reconnect/getEffectiveEnabled, plus 17 UI tests covering empty state/toggle/reset/badges/loading/description/source-type.
- **Clean error handling**: Both the store and the UI component handle errors gracefully with toast messages and signal-based error state.

---

### Recommendation: APPROVE

PR #876, opened 2026-03-25
test(e2e): add MCP registry UI and per-room enable/disable E2E test (Task 7.1)

lsm:
Add Playwright E2E test covering:
- Global Application MCP Servers settings showing pre-seeded fetch-mcp
- Per-room MCP toggle in Room Settings (enable/disable)
- Toggle UI state updates (room override badge)

Note: Tools Modal verification removed as it shows session-level MCP servers,
not app-level MCP servers like fetch-mcp. App-level MCP servers are
verified through RoomSettings UI and unit/integration tests.

### DIFF
--- a/packages/e2e/tests/features/app-mcp-registry.e2e.ts
+++ b/packages/e2e/tests/features/app-mcp-registry.e2e.ts
@@ -1,0 +1,193 @@
+/**
+ * App MCP Registry E2E Tests
+ *
+ * Tests for the Application MCP Registry UI and per-room MCP enable/disable:
+ * - Global Application MCP Servers settings show pre-seeded fetch-mcp entry
+ * - Per-room MCP toggle in Room Settings allows enable/disable
+ * - Toggle changes persist and reflect correctly in UI
+ *
+ * Note: Verifying MCP tools appear in the session's active tool list is done at
+ * the unit/integration level. The ToolsModal shows session-level MCP servers
+ * (from settings.local.json), not app-level MCP servers like fetch-mcp.
+ * App-level MCP servers are automatically included in room sessions based on
+ * the room's MCP settings.
+ *
+ * Setup: creates a room via RPC (infrastructure), then tests the UI.
+ * Cleanup: deletes the room via RPC in afterEach. Does NOT delete fetch-mcp
+ * from the registry — it is a permanent seed entry.
+ */
+
+import { test, expect } from '../../fixtures';
+import { waitForWebSocketConnected } from '../helpers/wait-helpers';
+import { createRoom, deleteRoom } from '../helpers/room-helpers';
+
+// ─── Constants ────────────────────────────────────────────────────────────────
+
+const FETCH_MCP_SERVER_NAME = 'fetch-mcp';
+
+type Page = import('@playwright/test').Page;
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+/**
+ * Navigate to the Application MCP Servers section in Global Settings.
+ * Assumes Global Settings panel is already open.
+ */
+async function navigateToAppMcpServersSection(page: Page): Promise<void> {
+	// Click on "Application MCP Servers" in the settings section list
+	const appMcpServersButton = page.locator('button:has-text("Application MCP Servers")');
+	await appMcpServersButton.waitFor({ state: 'visible', timeout: 5000 });
+	await appMcpServersButton.click();
+
+	// Wait for the AppMcpServersSettings panel to appear
+	await page.locator('text=Application MCP servers are available to any room').waitFor({
+		state: 'visible',
+		timeout: 5000,
+	});
+}
+
+/**
+ * Navigate to a room's Settings tab and wait for it to load.
+ */
+async function navigateToRoomSettings(page: Page, roomId: string): Promise<void> {
+	await page.goto(`/room/${roomId}`);
+	await waitForWebSocketConnected(page);
+
+	// Click the Settings tab
+	const settingsTab = page.locator('button:has-text("Settings")').first();
+	await settingsTab.click();
+
+	// Wait for MCP Servers section to be visible
+	await expect(page.getByText('MCP Servers', { exact: true })).toBeVisible({ timeout: 5000 });
+}
+
+// ─── Tests ───────────────────────────────────────────────────────────────────
+
+test.describe('App MCP Registry - Global Settings', () => {
+	test('should show fetch-mcp in Application MCP Servers list', async ({ page }) => {
+		await page.goto('/');
+		await waitForWebSocketConnected(page);
+
+		// Open Global Settings
+		const settingsButton = page.getByRole('button', { name: 'Settings', exact: true });
+		await settingsButton.click();
+
+		// Wait for settings panel to appear
+		await expect(page.getByRole('heading', { name: 'Settings', exact: true })).toBeVisible({
+			timeout: 5000,
+		});
+
+		// Navigate to Application MCP Servers section
+		await navigateToAppMcpServersSection(page);
+
+		// Verify fetch-mcp entry is present
+		const fetchMcpEntry = page.locator(`text="${FETCH_MCP_SERVER_NAME}"`).first();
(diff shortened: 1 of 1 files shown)


lsm (commented):

---

### P0 (Blocking)

**1. Last test is a no-op -- it asserts nothing about Tools Modal or MCP absence** (line 166-192)

The test titled `should verify fetch-mcp is not in Tools Modal when room MCP is disabled` does none of what its name claims. It:
- Disables fetch-mcp in room settings (fine)
- Clicks `button:has-text("Overview")` (navigates within the room's tab bar, not to a session)
- Waits 500ms with `page.waitForTimeout(500)`
- Ends with only a comment block explaining why it cannot verify anything

There is zero assertion after the disable step. The test will pass regardless of whether fetch-mcp is actually excluded from the session's MCP servers. This is not a test -- it is dead code that gives false confidence.

`packages/e2e/tests/features/app-mcp-registry.e2e.ts` lines 166-192

---

### P1 (Should-fix)

**2. Acceptance criteria gap: no verification that per-room disable actually affects sessions**

The task requires verifying the full flow: toggle per-room MCP, then confirm the change takes effect. The current tests only verify UI toggle state (checkbox checked/unchecked) but never verify the downstream effect -- that a session created in that room respects the per-room MCP setting.

The justification in the PR description that "ToolsModal shows session-level MCP servers (from settings.local.json), not app-level MCP servers" is architecturally accurate (confirmed by reading `ToolsModal.tsx` which calls `listMcpServersFromSources` for user/project/local sources, not the app registry). However, this does not mean the test cannot verify the effect at all. Alternatives:
- Open the Tools Modal on a session within the room and verify app-level MCP servers are NOT listed (if the wiring from Task 4.2 actually injects app MCP servers into the session tools config)
- Send a message asking the agent to use fetch-mcp and verify it fails or is unavailable
- At minimum, verify that the RoomSettings UI persists the disable after a page reload (round-trip persistence test)

**3. Fragile locator for room Settings tab** (line 57)

```ts
const settingsTab = page.locator('button:has-text("Settings")').first();
```

The room page at `packages/web/src/islands/Room.tsx` has a horizontal tab bar with "Overview", "Context", "Agents", "Missions", "Settings" buttons. While this works, using `.first()` with `:has-text("Settings")` is fragile because:
- The global bottom tab bar also has a "Settings" button (BottomTabBar.tsx line 116)
- If the global bar is visible, `.first()` could match the wrong element

A more robust approach would be scoping to the room's tab bar container, or using `page.getByRole('tab', { name: 'Settings' })` if the room tabs use ARIA roles, or adding a `data-testid` attribute.

**4. Hardcoded `page.waitForTimeout(500)` in the last test** (line 186)

Per the project's E2E rules, tests should use event-based waits (e.g., `waitForFunction`, locator assertions) rather than arbitrary timeouts. This 500ms wait doesn't actually synchronize on anything meaningful -- it just adds latency.

`packages/e2e/tests/features/app-mcp-registry.e2e.ts` line 186

---

### P2 (Important suggestions)

**5. Missing persistence verification for toggle changes**

The tests toggle checkboxes and verify the immediate UI state, but never verify that the changes persist after a page reload. For example, after toggling fetch-mcp off in test 3, navigating away and back should still show it as disabled. This would validate that the `setRoomMcpEnabled` API call actually wrote to the database.

**6. `span.uppercase:has-text("stdio")` locator is fragile** (line 89)

This locator depends on CSS class names (`uppercase`) which are Tailwind utility classes subject to refactoring. If the component is updated to use a different approach for uppercasing the source type text, the test will silently break. A more resilient approach would be to verify the text content directly within the fetch-mcp entry container.

`packages/e2e/tests/features/app-mcp-registry.e2e.ts` line 89

**7. Consider verifying both seed entries, not just fetch-mcp**

The seed defaults (`packages/daemon/src/lib/mcp/seed-defaults.ts`) create both `fetch-mcp` (enabled) and `brave-search` (disabled). The tests only verify fetch-mcp. Testing both entries would exercise more of the UI:
- fetch-mcp shows as checked/enabled (default)
- brave-search shows as unchecked/disabled with a "disabled globally" badge
- Toggling brave-search on shows the "room override" badge

---

### P3 (Nits)

**8. Unused type alias** (line 28)

```ts
type Page = import('@playwright/test').Page;
```

This type alias `Page` shadows Playwright's built-in `Page` type that is already available from the fixture destructuring. It is never used -- all functions already use the fixture's `Page` type through parameter inference.

`packages/e2e/tests/features/app-mcp-registry.e2e.ts` line 28

---

### Summary

The test file provides basic coverage for the MCP registry global settings view and per-room toggle UI mechanics, which is a good starting point. However, the most critical issue is the last test being a complete no-op: it asserts nothing about the MCP Tools Modal or session behavior, making it dead code. Combined with the lack of persistence verification and the acceptance criteria gap (no end-to-end verification that per-room disable actually affects sessions), the test suite falls short of the task's requirements. The tests should either be expanded to verify actual downstream effects or the last test should be removed entirely to avoid giving false confidence.

lsm (commented):

**Recommendation: APPROVE**

All P0/P1 issues from the first review have been properly addressed. The test file is now well-structured with meaningful assertions, proper persistence verification, scoped locators, and no arbitrary timeouts.

---

### P0/P1 Fixes Verification

1. **P0 (no-op test) -- FIXED**: The no-op test that asserted nothing has been removed and replaced with a brave-search disabled-globally badge test (line 193-208), which has proper assertions: verifies the entry is visible, checkbox is unchecked, and the "disabled globally" badge appears.

2. **P1 (no persistence verification) -- FIXED**: The toggle-off test (line 128-165) now includes a full persistence round-trip: after toggling off, it calls `page.reload()`, reconnects via `waitForWebSocketConnected(page)`, navigates back to room settings, and re-verifies the checkbox is still unchecked and the "room override" badge persists (lines 149-164).

3. **P1 (fragile Settings tab locator) -- FIXED**: The locator is now scoped to the room's tab bar div using `page.locator('.border-b.border-dark-700.bg-dark-850')` (line 57), which matches the Room.tsx tab bar container at line 171. This avoids collision with the global bottom nav Settings button. The CSS class combination is specific enough given that TaskView's similarly-styled header contains a back button rather than tab buttons.

4. **P1 (arbitrary waitForTimeout) -- FIXED**: Zero `waitForTimeout` calls remain in the file. All waits use event-based synchronization: `waitForWebSocketConnected`, `toBeVisible`, `toBeChecked`, `locator.waitFor`.

---

### Acceptance Criteria Assessment

The task asks to verify "fetch-mcp tools appear in the room session's active tool list after enabling." The worker's justification that ToolsModal only shows session-level MCP servers (from `settings.local.json` via `listMcpServersFromSources` calling `settings.mcp.listFromSources`) rather than app-level MCP servers is **valid**. Reading `ToolsModal.tsx` confirms the MCP section iterates over `mcpServersData.value?.servers[source]` for user/project/local sources -- there is no app-level source in its data model. The per-room MCP enablement is wired through `AppMcpLifecycleManager` (Task 4.2), verified through unit/integration tests, and the E2E tests here correctly validate the UI layer that controls the per-room override: the RoomSettings MCP toggle, badge display, and persistence. The acceptance criteria gap is a task scoping issue, not a test deficiency.

---

### Remaining Observations (P3 -- nits, no action required)

- The "toggle back on" test (line 167-191) correctly toggles off then on, but notes that the "room override" badge may persist even when the override matches the global default. This is architecturally correct (the override record exists in the DB regardless of value), but the comment could be clearer about why this is expected behavior.
- The `roomMcpStore` is subscribed in `RoomSettings.tsx` (line 112) but `roomMcpStore.unsubscribe()` is called in the cleanup without checking if subscription succeeded. This is pre-existing code, not introduced by this PR.

---

### Summary

The four tests provide solid E2E coverage for the MCP registry UI: global settings listing, default enabled state, toggle-off with persistence verification, toggle-back-on, and the brave-search disabled-globally state. The code follows project E2E conventions (infrastructure-only RPC in setup/teardown, all test actions through the UI, event-based waits). No remaining P0/P1/P2 issues.

PR #878, opened 2026-03-25
feat(space): add agent liveness guard with report_done timeout (Task 2.6)

lsm:
Auto-complete stuck agents that are alive but never call report_done after
the configurable AGENT_REPORT_DONE_TIMEOUT_MS (default 10 minutes).

- Add constants.ts with AGENT_REPORT_DONE_TIMEOUT_MS = 10 * 60 * 1000
- Add AgentAutoCompletedEvent to notification-sink.ts discriminated union
- Add formatAgentAutoCompleted() to session-notification-sink.ts
- Create agent-liveness.ts with autoCompleteStuckAgents() utility function
- Integrate autoCompleteStuckAgents() into SpaceRuntime.processRunTick()
  after the dead-agent reset loop (Step 1.5) to auto-complete agents that
  are alive but timed out without calling report_done
- Add 12 unit tests covering all scenarios: non-stuck, stuck, dead agents,
  missing taskAgentSessionId, non-in_progress status, multiple tasks,
  custom timeout, createdAt fallback, result message content

### DIFF
--- a/packages/daemon/src/lib/space/runtime/agent-liveness.ts
+++ b/packages/daemon/src/lib/space/runtime/agent-liveness.ts
@@ -90,6 +90,12 @@ export async function autoCompleteStuckAgents(
 		}
 
 		// Agent is alive but has not called report_done within the timeout window.
+		//
+		// Note: no notifiedTaskSet dedup is needed here because the status transition
+		// to 'completed' means this task will never satisfy the `status === 'in_progress'`
+		// guard again. Contrast with task_timeout, which is a warning-only notification
+		// that does not change task status and thus requires dedup to avoid re-emitting
+		// on every subsequent tick while the task remains in_progress.
 		const timeoutMinutes = Math.round(timeoutMs / 60_000);
 		const result = `Auto-completed: agent did not call report_done within ${timeoutMinutes} minutes`;
 
@@ -98,10 +104,19 @@ export async function autoCompleteStuckAgents(
 				`(elapsed ${Math.round(elapsedMs / 1000)}s, timeout ${timeoutMinutes}m)`
 		);
 
-		taskRepo.updateTask(task.id, {
+		const updated = taskRepo.updateTask(task.id, {
 			status: 'completed',
 			result,
 		});
+
+		// Guard: if the update returned null, the task was deleted concurrently.
+		// Skip the notification to avoid emitting an event for a non-existent task.
+		if (!updated) {
+			log.warn(
+				`agent-liveness: task ${task.id} was deleted before auto-completion could be persisted; skipping notification`
+			);
+			continue;
+		}
 
 		await notify({
 			kind: 'agent_auto_completed',
--- a/packages/daemon/src/lib/space/runtime/agent-liveness.ts
+++ b/packages/daemon/src/lib/space/runtime/agent-liveness.ts
@@ -1,0 +1,118 @@
+/**
+ * Agent Liveness Guard — Timeout for report_done
+ *
+ * Detects agents that are alive (session active) but have not called
+ * `report_done` within the configured timeout. Auto-completes them with a
+ * system-generated result so the workflow can continue.
+ *
+ * An agent is considered "stuck" when ALL of the following are true:
+ *   1. The task status is `'in_progress'`.
+ *   2. The task has a `taskAgentSessionId` (a Task Agent was spawned for it).
+ *   3. The Task Agent session is alive (`isTaskAgentAlive()` returns `true`).
+ *   4. Elapsed time since `startedAt` exceeds the configured timeout.
+ *
+ * Dead agents (where `isTaskAgentAlive()` returns `false`) are excluded because
+ * they are already handled by the dead-agent reset path in SpaceRuntime —
+ * resetting them to `'pending'` for re-spawn on the next tick.
+ */
+
+import type { SpaceTask } from '@neokai/shared';
+import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
+import type { TaskAgentManager } from './task-agent-manager';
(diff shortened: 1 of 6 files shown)


lsm (commented):

---

**Recommendation: APPROVE**

This is a clean, well-structured implementation of Task 2.6 (Agent Liveness Guard). The code is correct, well-documented, and properly integrated. All 12 unit tests pass, typecheck is clean, and the linter reports no issues. Below are detailed findings.

---

### What was reviewed

6 files changed (+612 / -1):
- `packages/daemon/src/lib/space/runtime/constants.ts` (new)
- `packages/daemon/src/lib/space/runtime/agent-liveness.ts` (new)
- `packages/daemon/src/lib/space/runtime/notification-sink.ts` (modified)
- `packages/daemon/src/lib/space/runtime/session-notification-sink.ts` (modified)
- `packages/daemon/src/lib/space/runtime/space-runtime.ts` (modified)
- `packages/daemon/tests/unit/space/agent-liveness.test.ts` (new)

---

### Correctness

The core logic in `autoCompleteStuckAgents()` (agent-liveness.ts:62-118) correctly implements the four-condition check:
1. Task status is `in_progress`
2. Task has a `taskAgentSessionId`
3. Agent is alive per `isTaskAgentAlive()`
4. Elapsed time exceeds the timeout

The ordering of conditions is efficient -- cheap checks (status, session ID) run before the more expensive liveness check.

The integration into `processRunTick()` at Step 1.5 (space-runtime.ts:733-750) is placed correctly: after dead-agent resets (Step 1) so we only process truly alive agents, and before spawn logic (Step 2). The re-read from DB on line 736-738 ensures status resets from Step 1 are picked up.

The `completed_at` timestamp is automatically set by `updateTask()` when status transitions to `completed` (space-task-repository.ts:156-162), so the auto-completed task will have correct completion metadata.

The `createdAt` fallback (agent-liveness.ts:85) is safe because `SpaceTask.createdAt` is typed as `number` (required) in the shared types, so `referenceTime` can never be `undefined`.

---

### Architecture & Integration

- The new `AgentAutoCompletedEvent` follows the existing discriminated union pattern in `notification-sink.ts` and is correctly added to the `SpaceNotificationEvent` type.
- `formatAgentAutoCompleted()` in `session-notification-sink.ts` follows the same three-section message format as all other event formatters (human-readable summary + JSON payload + autonomy level).
- The constant in `constants.ts` with its JSDoc explaining the rationale is a good practice for future configurability.
- The function signature with a default parameter for `timeoutMs` makes testing easy without requiring env/config injection.

---

### Edge Cases Handled

- Dead agents are correctly skipped (handled by existing reset-to-pending path).
- Tasks without `taskAgentSessionId` are correctly skipped.
- Non-`in_progress` tasks are correctly skipped.
- The `<= timeoutMs` boundary check means exact-timeout is NOT auto-completed (only after exceeding) -- correct behavior.
- Multiple stuck agents in a single call are all handled.
- The `safeNotify.bind(this)` wrapper at space-runtime.ts:744 ensures notification failures cannot crash the tick loop.

---

**P2: Missing notification dedup for `agent_auto_completed`**

The existing `task_timeout` notification uses `notifiedTaskSet` dedup to avoid re-emitting on every tick. The new `agent_auto_completed` does not use dedup. This is **currently correct** because the task status changes to `completed` on auto-completion, so the condition (`status === 'in_progress'`) will never match again. However, if `notifiedTaskSet` cleanup in `cleanupTerminalExecutors()` were ever extended to include an `agent_auto_completed` key, it would need to be consistent. This is not a problem now but worth noting for future maintainers.

**P3: Missing test for exact timeout boundary**

There is no test verifying that an agent at exactly `elapsed === timeoutMs` is NOT auto-completed. The test at line 225 seeds with `+5000ms` over the timeout. Adding a boundary test with exactly `timeoutMs` elapsed would make the `<=` check on line 88 more explicit.

**P3: `updateTask` return value not checked**

In agent-liveness.ts:101-104, the return value of `taskRepo.updateTask()` is not checked. If the update silently failed (task deleted concurrently), the notification would still be emitted for a task that was not actually updated. This is extremely unlikely in practice since the task was just read moments before, but checking the return value would be more defensive.

---

### Summary

Solid implementation. The code is well-structured, properly tested, and follows existing patterns. No P0/P1 issues found.

lsm (commented):

---

**Recommendation: APPROVE**

This commit properly addresses all three findings from the previous review.

---

### What was reviewed (second commit only)

2 files changed (+50 / -1):
- `packages/daemon/src/lib/space/runtime/agent-liveness.ts` (modified)
- `packages/daemon/tests/unit/space/agent-liveness.test.ts` (modified)

---

### P2: Notification dedup consistency -- RESOLVED

The comment added at `agent-liveness.ts:93-98` clearly explains why `notifiedTaskSet` dedup is not needed: the status transition to `'completed'` means the `status === 'in_progress'` guard will never match again. The comment also explicitly contrasts with `task_timeout`, which is warning-only and does require dedup. This is exactly the right level of documentation for future maintainers.

### P3: `updateTask` return value not checked -- RESOLVED

The fix at `agent-liveness.ts:107-119` captures the return value of `taskRepo.updateTask()` and checks for `null`. If the task was concurrently deleted, the notification is skipped and a warning is logged. The `continue` statement correctly skips to the next iteration. This is a clean, defensive pattern.

### P3: Missing boundary test -- RESOLVED

The new test at line 224-256 uses a patched `SpaceTask` object with `startedAt: Date.now() - customTimeoutMs` to ensure `elapsed === timeoutMs`. The test correctly asserts that the result is empty, no events are emitted, and the task remains `in_progress`. The approach of patching the in-memory task object (same pattern used in the `createdAt` fallback test at line 359) is consistent with the existing test style.

### Verification

All 13 tests pass (12 original + 1 new boundary test). No regressions.

PR #880, opened 2026-03-25
feat(shared): extend ResolvedChannel with gate field for policy enforcement

lsm:
Add `gate?: WorkflowCondition` to `ResolvedChannel` so gate policy
information from `WorkflowChannel` is preserved after resolution.
Propagate gate through both `expandChannel` (deprecated path) and
`expandUnifiedChannel` (unified path).

Also ensures only one resolved channel type exists (no separate
ResolvedCrossNodeChannel) — ResolvedChannel already captures all
info needed for routing and gate evaluation.

Tests: 9 new tests covering gate propagation for one-way, bidirectional,
fan-out channels, and ChannelResolver behaviour with gated channels.

### DIFF
--- a/packages/daemon/tests/unit/space/channel-resolution.test.ts
+++ b/packages/daemon/tests/unit/space/channel-resolution.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from 'bun:test';
-import type { SpaceAgent, SpaceWorkflow, WorkflowNode } from '@neokai/shared';
+import type { SpaceAgent, SpaceWorkflow, WorkflowCondition, WorkflowNode } from '@neokai/shared';
 import { resolveChannels, validateChannels } from '@neokai/shared';
 
 // ============================================================================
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped change that correctly extends `ResolvedChannel` with the `gate?: WorkflowCondition` field and propagates it through both code paths. The implementation is straightforward and consistent with the existing patterns.

### What was done well

- The `gate` field on `ResolvedChannel` (packages/shared/src/types/space-utils.ts:66) has clear JSDoc explaining its semantics.
- Gate is correctly propagated in all four push sites: `expandChannel` forward (line 207), `expandChannel` reverse/bidirectional (line 222), `expandUnifiedChannel` forward (line 445), and `expandUnifiedChannel` reverse/bidirectional (line 459).
- The `isValidResolvedChannel` guard in channel-resolver.ts correctly does NOT validate the `gate` field -- since `gate` is optional, omitting it from the structural check is the right call. A channel stored before this change (without `gate`) will still deserialize correctly.
- All 9 new tests are meaningful and cover distinct scenarios: one-way propagation, absence when no gate, bidirectional (both directions), fan-out (all entries), human-type gate, ChannelResolver preservation, fromRunConfig deserialization, undefined when absent, and canSend independence.
- No `ResolvedCrossNodeChannel` type exists anywhere in the codebase (confirmed by grep). The acceptance criterion "only one resolved channel type exists" is met.
- TypeScript typecheck passes cleanly. All 33 + 30 = 63 tests pass.

1. **Plan document field names differ from implementation**: The plan (docs/plans/refactor-space-workflow-from-step-centric-to-agent-centric-c/01-unified-channel-type-and-gates.md, Task 1.2) sketches `ResolvedChannel` with `channelId`, `from`, `to`, `toNodeName`, and `gateLabel` fields. The actual implementation uses `fromRole`/`toRole` instead of `from`/`to`, has `fromAgentId`/`toAgentId` instead of `channelId`/`toNodeName`, and has no `gateLabel`. This is fine -- the implementation is more precise than the plan -- but the plan document could be updated to avoid confusion for future developers.

2. **No test for hub-spoke + gate combination**: The channel-resolution tests cover one-way, bidirectional, and fan-out with gate, but there is no test for the hub-spoke topology (bidirectional with array `to`) with a gate. This is a low-risk gap since hub-spoke internally uses the same push sites, but it would close the coverage matrix.

3. **No test for wildcard from/to with gate**: Similarly, there is no test combining the `*` wildcard addressing with a gate. The resolver expands wildcards through the same code paths, so this is also low risk, but it would be a good addition for completeness.

These are all P3 nits. The change is correct, complete for its scope, and ready to merge.

PR #886, opened 2026-03-25
feat(space): add ChannelRouter with lazy node activation (Task 3.0)

lsm:
Implements lazy activation of target nodes in the channel routing layer.
When a message is delivered to an agent role whose node has no active tasks,
the router creates pending SpaceTask records on-demand rather than requiring
upfront activation at workflow start.

Key changes:

- ChannelRouter (channel-router.ts): New class with activateNode() and
  deliverMessage() for lazy on-demand node activation
- Migration 54: Unique partial index on (workflow_run_id, workflow_node_id,
  slot_role) for active statuses only; terminal statuses excluded for cyclic
  workflow support; guarded by column-existence for forward-compat
- WorkflowExecutor: followTransition() made idempotent via UNIQUE constraint
  error handling; cyclic workflows reuse existing tasks instead of throwing
- 14 unit tests for ChannelRouter covering activation scenarios
- 10 unit tests for Migration 54 covering index behavior

### DIFF
--- a/packages/daemon/src/lib/space/runtime/channel-router.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-router.ts
@@ -173,6 +173,13 @@ export class ChannelRouter {
 			} catch (err) {
 				// Detect DB UNIQUE constraint violation caused by a concurrent activateNode() call.
 				// The winning writer already created the tasks — re-read and return them.
+				//
+				// Note on partial-activation tradeoff: when the violation occurs on the nth
+				// agent slot (n > 1), the tasks already pushed to `tasks` earlier in this
+				// loop are silently discarded in favour of the re-read set. This is safe
+				// because both callers resolved task metadata from the same DB state, so the
+				// winning writer's tasks are equivalent. The returned set is guaranteed to be
+				// a superset of what this caller created so far, so no task is lost.
 				if (isDuplicateConstraintError(err)) {
 					const concurrentTasks = this.getActiveTasksForNode(runId, nodeId);
 					if (concurrentTasks.length > 0) {
@@ -253,6 +260,13 @@ export class ChannelRouter {
 	 *
 	 * Terminal tasks (completed, cancelled, needs_attention, archived) are
 	 * excluded so cyclic workflows can re-activate a node after its tasks complete.
+	 *
+	 * `draft` is intentionally excluded even though it is a valid `SpaceTaskStatus`.
+	 * The ChannelRouter always creates tasks with `status: 'pending'` — `draft` is
+	 * reserved for tasks created by external callers that are not yet ready to run.
+	 * The Migration 54 index matches this exclusion, meaning two draft tasks for the
+	 * same (run, node, slot) are allowed to coexist. Callers that create draft tasks
+	 * outside ChannelRouter must manage uniqueness themselves.
 	 */
 	private getActiveTasksForNode(runId: string, nodeId: string): SpaceTask[] {
 		const ACTIVE_STATUSES = new Set([
--- a/packages/daemon/src/lib/space/runtime/channel-router.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-router.ts
@@ -1,0 +1,312 @@
+/**
+ * ChannelRouter — lazy node activation for agent-centric workflow execution.
+ *
+ * When a message arrives for a target agent role whose node has no active tasks,
+ * the router creates pending SpaceTask records on-demand (lazy activation).
+ * This replaces the step-centric model where the executor always pre-created
+ * tasks for the next node eagerly.
+ *
+ * Key design decisions:
+ * - activateNode() is idempotent: if tasks already exist for the node the
+ *   existing tasks are returned unchanged.
+ * - Concurrency is handled via a DB UNIQUE partial index on
+ *   (workflow_run_id, workflow_node_id, slot_role). A duplicate INSERT throws
+ *   a SQLiteError; the handler re-reads and returns the tasks created by the
+ *   winning writer.
+ * - No session group creation — that is the responsibility of TaskAgentManager.
+ *   ChannelRouter only creates SpaceTask DB records.
+ */
+
+import type { SpaceTask, SpaceTaskType, SpaceWorkflow, WorkflowNode } from '@neokai/shared';
+import { resolveNodeAgents } from '@neokai/shared';
+import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
+import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
+import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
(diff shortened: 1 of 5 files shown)


lsm (commented):

---

The implementation is well-structured and the core lazy activation logic is sound. The partial unique index design correctly supports cyclic workflows by excluding terminal statuses, and the concurrent-activation fallback via UNIQUE constraint violation handling is a reasonable approach. All 24 tests pass, and the active-status sets are consistent across the migration index, ChannelRouter, and WorkflowExecutor. However, there are two issues that warrant attention before merge.

---

### P1: Partial activation loses already-created tasks on UNIQUE fallback (channel-router.ts:176-180)

`packages/daemon/src/lib/space/runtime/channel-router.ts`, lines 156-189

When a multi-agent node is being activated and the second (or later) agent slot triggers a UNIQUE constraint violation, the code at line 177-179 returns `concurrentTasks` (re-read from DB). This set includes ALL active tasks for the node -- including the task(s) already created successfully earlier in the same loop. However, the tasks pushed to the `tasks` array before the error (lines 172) are silently discarded.

This is mostly fine functionally because the re-read set is a superset of what was built so far, but it creates an inconsistency: if two concurrent `activateNode()` calls race on a multi-agent node and partially succeed, one caller may return tasks created by the OTHER caller (including potentially stale task metadata if the two callers resolved different `taskType`/`customAgentId` values for the same slot). In practice this is extremely unlikely since both callers resolve from the same DB state, but the code should be aware of this and ideally documented. Consider adding a brief comment at the fallback return noting this tradeoff.

Additionally, the silent discard of earlier-created tasks in the `tasks` array means that any validation logic added later that operates on the returned task set could be confused by receiving tasks it did not create.

**Severity: P1** -- not a correctness bug today, but a fragile pattern that could mask future issues if the loop body gains more complexity.

---

### P2: `draft` status excluded from both index and active-status sets, but not documented or tested (migrations.ts:3309, channel-router.ts:258)

`packages/daemon/src/storage/schema/migrations.ts`, line 3309

The partial index WHERE clause includes `status IN ('pending', 'in_progress', 'review', 'rate_limited', 'usage_limited')`, which excludes both terminal statuses (completed, cancelled, needs_attention, archived) AND `draft`. The `ACTIVE_STATUSES` sets in both `channel-router.ts` (line 258) and `workflow-executor.ts` (line 458) match the index exactly.

The `SpaceTaskStatus` type (shared/src/types/space.ts:119) includes `'draft'` as a valid status. If a task is ever created in `'draft'` status, the partial index will not protect against duplicates for that slot. The `ChannelRouter.activateNode()` always creates tasks with `status: 'pending'` (line 169), so this is safe for current usage, but:

1. The rationale for excluding `draft` should be explicitly stated in the migration comment and the `getActiveTasksForNode` docstring, so future developers understand it is an intentional choice.
2. The migration-54 tests should include a case verifying that two `draft`-status tasks with the same (run, node, slot) do NOT violate the index (a negative test confirming the intended behavior).

**Severity: P2** -- no current bug, but a documentation/test gap that could lead to a future bug if draft-status tasks are introduced.

---

### P3 (nits)

1. **`ChannelRouter` not exported from barrel** (`packages/daemon/src/lib/space/index.ts`): `ChannelRouter`, `ActivationError`, `ChannelRouterConfig`, and `DeliveredMessage` are not re-exported from the barrel file. If this is intentional because the class is not yet wired into `SpaceRuntime`, a TODO comment would help future readers. If knip is configured to flag unused exports, this could also cause issues.

2. **`deliverMessage()` does not validate `fromRole`** (`channel-router.ts`:206-240): The method validates `toRole` exists in the workflow but silently accepts any `fromRole`. The `fromRole` is included in the returned `DeliveredMessage` but never validated against the workflow's agent slots. If callers are expected to trust the `fromRole` blindly, consider adding a brief comment explaining why validation is deferred (e.g., the sending session's task already identifies the agent).

3. **`deliverMessage()` does not persist the message** (`channel-router.ts`:206-240): The method returns a `DeliveredMessage` descriptor but never stores the message content anywhere (no DB insert, no event emission). Callers are responsible for persisting. This is a valid design choice but should be documented in the JSDoc since the name `deliverMessage` strongly implies message delivery/storage, not just lazy activation.

---

### Summary

The lazy activation logic, idempotency handling, and partial unique index design are all correct and well-tested. The main concerns are: (1) the partial-activation fallback path silently discards earlier-created tasks without documenting the tradeoff, and (2) the `draft` status exclusion needs documentation and a negative test case.

lsm (commented):

---

**Recommendation: APPROVE**

Re-reviewed after the fix commit (25aa41da) addressed both issues from the previous review. All 25 tests pass (14 channel-router + 11 migration-54).

---

### Previously raised issues — status

**P1 (partial-activation tradeoff): RESOLVED**
`packages/daemon/src/lib/space/runtime/channel-router.ts`, lines 177-182

The fix adds a clear 7-line comment explaining why discarding partially-created tasks in favour of the re-read set is safe: both concurrent callers resolve task metadata from the same DB state, so the winning writer's tasks are equivalent. The comment also notes the returned set is guaranteed to be a superset. This adequately documents the tradeoff for future maintainers.

**P2 (draft exclusion documentation and negative test): RESOLVED**
Three fixes applied:

1. `channel-router.ts` `getActiveTasksForNode()` docstring (lines 264-269): Explicitly documents that `draft` is intentionally excluded, the rationale (ChannelRouter always creates `pending` tasks; `draft` is reserved for external callers), and that external callers must manage their own uniqueness.

2. `migrations.ts` `runMigration54()` docstring (lines 3287-3289): Added `draft` to the "Excluded statuses and rationale" section with a clear explanation of why it is excluded from the partial index.

3. `migration-54_test.ts` (lines 214-234): Added a negative test asserting two `draft` tasks with the same `(`run`, node, slot)` do NOT trigger a UNIQUE violation, with an explanatory comment.

---

### Full review summary

The implementation is correct and well-structured:

- **ChannelRouter** provides clean lazy node activation with proper idempotency (early check + UNIQUE fallback).
- **Migration 54** partial unique index correctly supports cyclic workflows by excluding terminal statuses, and gracefully guards against missing columns for backward compatibility.
- **WorkflowExecutor** `followTransition()` is properly idempotent via the same UNIQUE constraint violation handling.
- **ACTIVE_STATUSES** sets are consistent across all three locations (migration index WHERE clause, ChannelRouter, WorkflowExecutor).
- **Concurrency handling**: Both concurrent `activateNode()` calls and races between `activateNode()` and `followTransition()` are handled correctly.

### P3 nits (carried over, non-blocking)

1. `ChannelRouter` is not exported from `packages/daemon/src/lib/space/index.ts` — acceptable for a new class not yet wired into `SpaceRuntime`.
2. `deliverMessage()` does not validate `fromRole` against the workflow's agent slots — reasonable for the current design where the sending session's task already identifies the agent.
3. `deliverMessage()` returns a `DeliveredMessage` descriptor without persisting the message — valid lazy activation design, but the name could mislead future readers into expecting message persistence.

PR #887, opened 2026-03-25
feat(space): extend ChannelRouter with gate evaluation and cyclic iteration tracking (Task 3.1)

lsm:
- Add canDeliver() — pure gate+iteration-cap check without side effects
  - Finds matching WorkflowChannel for from/to pair (open topology returns allowed)
  - Blocks on cyclic channel when iterationCount >= maxIterations
  - Evaluates all gate types (always, human, condition, task_result)

- Update deliverMessage() with optional gate context parameter
  - Evaluates gate via ChannelGateEvaluator when context is provided
  - Throws ChannelGateBlockedError for gate-blocked deliveries (retryable)
  - Throws ActivationError for iteration cap exceeded (hard limit)
  - Increments run.iterationCount for cyclic channels after delivery
  - Adds fan-out support: node name targets activate all agents in that node
  - Sets isFanOut flag on DeliveredMessage for node-name vs agent-role distinction

- Add gateEvaluator to ChannelRouterConfig (injectable for tests)
- Re-export ChannelGateBlockedError, ChannelGateContext, GateResult from channel-router

- 38 unit tests (25 new + 13 existing) covering all new features

### DIFF
--- a/packages/daemon/src/lib/space/runtime/channel-router.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-router.ts
@@ -383,14 +383,18 @@ export class ChannelRouter {
 		}
 
 		// ── 5. Increment iteration count for cyclic channels ──────────────────
-		// Re-read the run to get the freshest iterationCount before incrementing
-		// (another concurrent delivery may have already incremented it).
+		// Use the atomic incrementIterationCount() which runs a single SQL UPDATE
+		// with a WHERE guard (iteration_count < max_iterations), eliminating any
+		// TOCTOU race between the cap check above and this write. If another
+		// concurrent delivery already pushed the count to the cap, the atomic
+		// increment will return false — we throw to surface the cap violation even
+		// though message delivery already succeeded at this point.
 		if (channel?.isCyclic) {
-			const freshRun = this.config.workflowRunRepo.getRun(runId);
-			if (freshRun) {
-				this.config.workflowRunRepo.updateRun(runId, {
-					iterationCount: freshRun.iterationCount + 1,
-				});
+			const incremented = this.config.workflowRunRepo.incrementIterationCount(runId);
+			if (!incremented) {
+				throw new ActivationError(
+					`Cyclic channel from "${fromRole}" to "${toTarget}" reached the maximum iteration count during delivery (cap enforced atomically). Message was delivered but the cycle was not counted.`
+				);
 			}
 		}
 
--- a/packages/daemon/src/lib/space/runtime/channel-router.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-router.ts
@@ -1,12 +1,17 @@
 /**
- * ChannelRouter — lazy node activation for agent-centric workflow execution.
+ * ChannelRouter — message delivery with gate enforcement and lazy node activation.
  *
- * When a message arrives for a target agent role whose node has no active tasks,
- * the router creates pending SpaceTask records on-demand (lazy activation).
- * This replaces the step-centric model where the executor always pre-created
- * tasks for the next node eagerly.
+ * Handles all message delivery within a workflow run:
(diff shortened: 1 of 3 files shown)


lsm (commented):

---

### Summary

The implementation extends the existing ChannelRouter with canDeliver(), gate evaluation in deliverMessage(), cyclic iteration tracking, and fan-out support. The code is well-structured, thoroughly documented, and all 38 tests pass. The design is sound overall. I found one P1 issue (race condition in iteration count increment) and several P2 items worth addressing before merging.

---

### P1: TOCTOU race in iteration count increment (deliverMessage)

**File:** packages/daemon/src/lib/space/runtime/channel-router.ts, lines 388-395

The code re-reads the run to get the freshest iterationCount before incrementing, but there is no atomicity guarantee between the read and the subsequent updateRun. Two concurrent deliveries on the same cyclic channel can both read the same iterationCount, both compute the same +1 value, and both write it, losing one increment.

The check on line 342 (run.iterationCount >= run.maxIterations) uses a stale run snapshot from the start of the method. Between that check and the increment, another concurrent delivery can push the count past the cap, meaning the cap check can be bypassed.

**Recommendation:** Use an atomic SQL UPDATE ... SET iteration_count = iteration_count + 1 WHERE iteration_count < ? to both enforce the cap and increment in a single statement. This eliminates the TOCTOU race entirely. Alternatively, if you intend to keep the read-then-write pattern, add a comment explicitly acknowledging the tradeoff.

---

### P2: Missing canDeliver() tests for condition gate type

**File:** packages/daemon/tests/unit/space/channel-router.test.ts, canDeliver describe block

The canDeliver() test suite covers always, human, task_result, and cyclic iteration caps, but does not test the condition gate type. The deliverMessage() tests do cover condition gates via injected mock evaluators (lines 854-935), but canDeliver() has no such test. This is a gap because canDeliver() calls evaluateCondition() directly (line 282), and condition gates are the only type that spawn a subprocess -- the most error-prone path.

**Recommendation:** Add at least one test for canDeliver() with a condition gate using an injected mock evaluator (both allowed and blocked cases).

---

### P2: No test coverage for wildcard channel matching (from: '*', to: '*')

**File:** packages/daemon/src/lib/space/runtime/channel-router.ts, lines 474-488

The findMatchingWorkflowChannel() method supports wildcard matching on both from and to, but there are no tests for either wildcard variant. If a wildcard channel is declared with a gate or isCyclic: true, the behavior is untested.

**Recommendation:** Add tests for wildcard from, wildcard to, and wildcard channel with a gate.

---

### P2: Missing test for canDeliver() when workflow is not found

The test verifies that canDeliver() throws ActivationError when the run is not found (line 1388), but there is no test for when the workflow is not found. The code at line 263-264 does throw for this case, but it is untested.

---

### P2: Silent no-op when freshRun is null after cyclic delivery

**File:** packages/daemon/src/lib/space/runtime/channel-router.ts, lines 388-395

If the re-read at line 389 returns null (theoretically impossible in normal operation but possible under extreme conditions like a concurrent run deletion), the iteration count is silently not incremented. The delivery has already succeeded, so the caller receives a successful result even though the cyclic tracking was skipped.

**Recommendation:** Either throw an error if freshRun is null, or add a comment documenting this as a deliberate defensive choice.

---

### What works well

- Clean separation between canDeliver() (pure check, no side effects) and deliverMessage() (full pipeline with mutations).
- Proper use of re-exports so callers only need one import from channel-router.ts.
- The iteration count re-read before increment is a good defensive pattern (even with the race condition noted above).
- Gate evaluator injection via config is a clean testability pattern.
- The isFanOut flag on DeliveredMessage is a thoughtful API design for callers.
- Excellent JSDoc documentation throughout.

---

**Recommendation: REQUEST_CHANGES** (due to P1 and P2 issues)

PR #891, opened 2026-03-25
feat(room-runtime): add public clearGroupRateLimit() method

lsm:
Adds clearGroupRateLimit(taskId) to RoomRuntime which looks up the active
group for a task, clears its rate limit via groupRepo.clearRateLimit(), and
clears the task restriction via clearTaskRestriction(). Returns true if a
group was found and cleared, false otherwise.

This supports Bug B fix: when task.setStatus transitions a task from
usage_limited/rate_limited back to in_progress, the caller can also clear
the stale group.rateLimit so the next worker iteration is not blocked.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1808,6 +1808,25 @@ export class RoomRuntime {
 	}
 
 	/**
+	 * Clear the rate limit backoff for the group associated with a task.
+	 *
+	 * Called when a user manually transitions a task from `usage_limited`/`rate_limited`
+	 * back to `in_progress` so that stale rate-limit state doesn't block the next worker
+	 * iteration from forwarding its output to the leader.
+	 *
+	 * Returns `true` if a group was found and cleared, `false` if no group exists.
+	 */
+	async clearGroupRateLimit(taskId: string): Promise<boolean> {
+		const group = this.groupRepo.getGroupByTaskId(taskId);
+		if (!group) return false;
+
+		this.groupRepo.clearRateLimit(group.id);
+		await this.clearTaskRestriction(taskId);
+		log.info(`Cleared rate limit for group ${group.id} (task ${taskId})`);
+		return true;
+	}
+
+	/**
 	 * Archive a task group - terminate active sessions, cleanup worktree, and set archived status.
 	 *
 	 * Called when user archives a task via UI. This:
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1814,11 +1814,11 @@ export class RoomRuntime {
 	 * back to `in_progress` so that stale rate-limit state doesn't block the next worker
 	 * iteration from forwarding its output to the leader.
 	 *
-	 * Returns `true` if a group was found and cleared, `false` if no group exists.
+	 * Returns `true` if an active group was found and cleared, `false` otherwise.
 	 */
 	async clearGroupRateLimit(taskId: string): Promise<boolean> {
 		const group = this.groupRepo.getGroupByTaskId(taskId);
-		if (!group) return false;
+		if (!group || group.completedAt !== null) return false;
 
 		this.groupRepo.clearRateLimit(group.id);
 		await this.clearTaskRestriction(taskId);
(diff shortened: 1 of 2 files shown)


lsm (commented):

## Verdict: APPROVE

The implementation is clean, follows existing patterns, and the tests are solid. One suggestion worth considering for a follow-up.

## What's Good

- **Method design**: `clearGroupRateLimit` follows the exact same `getGroupByTaskId` + null guard pattern used by `terminateTaskGroup` at line 1792, which is the established convention for taskId-to-group lookups in this class.
- **Correct downstream calls**: Calling `groupRepo.clearRateLimit(group.id)` (sync) followed by `this.clearTaskRestriction(taskId)` (async) is the right sequence. `clearTaskRestriction` (line 3918) already guards against calling `updateTaskStatus` when the task is not in a limited state, so there's no double-transition risk.
- **Tests**: Both test cases are meaningful. The "no group" path validates the false return, and the "clears group rateLimit and task restriction" test exercises the full lifecycle: trigger rate limit, verify state, clear, verify restored state. This is the right level of coverage for a thin wrapper method.
- **Documentation**: The JSDoc clearly explains the purpose and when it's called.

## Suggestion (P2)

**Missing `completedAt` guard on the returned group**

At `packages/daemon/src/lib/room/runtime/room-runtime.ts:1820`, `getGroupByTaskId` returns the latest group for a task regardless of whether it's completed (the SQL at `session-group-repository.ts:286` has no `completed_at IS NULL` filter). In contrast, `terminateTaskGroup` at line 1796 explicitly checks `group.completedAt === null` before taking action.

If `clearGroupRateLimit` is called on a task whose group was already terminated (e.g., a race between user actions), it will return `true` after clearing the rate limit of a dead group. This is harmless in practice -- a completed group won't have active worker iterations to block -- but it's inconsistent with the sibling method's defensive pattern. Consider adding a `completedAt` guard for consistency:

```ts
if (!group || group.completedAt !== null) return false;
```

This would also make the return value semantics more precise: `true` means "found and cleared an active group," not "found the latest group even if it's dead."

## Note on Scope

The method has no callers wired up yet. The PR body correctly scopes this as providing the building block ("the caller can also clear the stale group.rateLimit"), with the actual integration into `task.setStatus` presumably coming in a follow-up PR. That's a reasonable incremental approach, but make sure the wiring PR doesn't get dropped.

lsm (commented):

### Verdict: APPROVE

The implementation is clean, correct, and well-tested. No blocking issues.

### P2 from prior review -- properly addressed

The `completedAt !== null` guard was added at line 1821 of `room-runtime.ts`:

```ts
if (!group || group.completedAt !== null) return false;
```

This is consistent with the sibling `terminateTaskGroup` method (line 1796) which uses the same `group.completedAt === null` check pattern. The guard is necessary because `getGroupByTaskId` (session-group-repository.ts:275) does **not** filter by `completed_at IS NULL` -- it returns the most recent group for a task regardless of completion status. Without this guard, clearing rate limits on a completed group would silently corrupt archived state.

The new test at line 301 (`returns false when the group is already completed`) directly validates this path by calling `completeGroup` before `clearGroupRateLimit` and asserting `false`.

### Test coverage assessment

All three tests in the new `clearGroupRateLimit(taskId)` describe block are well-structured:

1. **No group exists** (line 294) -- guards the null-group path
2. **Completed group** (line 301) -- guards the `completedAt !== null` path (the P2 fix)
3. **Happy path** (line 315) -- verifies both side effects: `groupRepo.clearRateLimit()` clears the rate limit, and `clearTaskRestriction()` restores task to `in_progress` with `restrictions: null`

The happy path test is thorough -- it pre-verifies the rate-limited state before clearing, then asserts both the group-level and task-level state changes after. This matches the actual code flow in `clearGroupRateLimit`.

### Integration notes

The method follows the same pattern as the existing `handleLeaderTool('send_to_worker', ...)` path (line 1206) which also calls `this.groupRepo.clearRateLimit()` + `this.clearTaskRestriction()`. The difference is that the public method operates by `taskId` rather than `groupId`, which is the correct interface for callers outside the runtime that don't have the group ID.

The private `clearTaskRestriction` method (line 3918) already has its own status guard (`if (task.status !== 'rate_limited' && task.status !== 'usage_limited') return`) which provides defense-in-depth -- calling `clearGroupRateLimit` on a task that isn't rate-limited would clear the group's rateLimit but not change task status.

### No issues found

- Code is correct and consistent with existing patterns
- Tests are comprehensive and all pass (13/13)
- Type check passes cleanly
- No new issues introduced by the `completedAt` guard

PR #892, opened 2026-03-25
fix(runtime): add !group.rateLimit re-detection guard to leader usage_limit handler

lsm:
Wraps the usage_limit detection block in onLeaderTerminalState() with the
same `!group.rateLimit` guard already used by the rate_limit handler. When
group.rateLimit is already set (even if expired), re-triggers skip the
fallback/backoff logic and fall through to normal completion. This prevents
an infinite loop when recoverStuckLeaders (Gap F) re-triggers
onLeaderTerminalState after the usage limit has reset but the old session
output still contains the usage-limit text.

Adds two new unit tests covering both the active and expired rateLimit cases.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1092,42 +1092,51 @@ export class RoomRuntime {
 					return;
 				}
 				if (errorClass?.class === 'usage_limit') {
-					// Usage limit (daily/weekly cap) — do NOT wait. Try fallback model immediately.
-					// If no fallback is configured, fall through to rate_limit behavior (backoff + pause).
-					log.info(
-						`Usage limit detected in leader output for group ${groupId}: ${errorClass.reason}`
-					);
-					const switched = await this.trySwitchToFallbackModel(
-						groupId,
-						group.leaderSessionId,
-						'leader'
-					);
-					if (!switched) {
-						// No fallback available — fall through to rate_limit behavior (backoff + pause)
-						const rateLimitBackoff = errorClass.resetsAt
-							? createRateLimitBackoff(leaderOutputText, 'leader')
-							: null;
-						const backoff: RateLimitBackoff = rateLimitBackoff ?? {
-							detectedAt: Date.now(),
-							resetsAt: Date.now() + 60 * 1000,
-							sessionRole: 'leader',
-						};
-						this.groupRepo.setRateLimit(groupId, backoff);
-						this.appendGroupEvent(groupId, 'rate_limited', {
-							text: `Usage limit reached in leader. Pausing until ${new Date(backoff.resetsAt).toLocaleTimeString()}.`,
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

This PR adds a `!group.rateLimit` guard to the leader `usage_limit` block in `onLeaderTerminalState()`, preventing an infinite loop when `recoverStuckLeaders` re-triggers the handler after a rate limit has already been set but the old session output still contains usage-limit text. The implementation is correct and well-tested.

### Correctness

The guard at line 1100 (`if (!group.rateLimit)`) correctly matches the pattern established in the `rate_limit` handler at line 1065 (`if (errorClass?.class === 'rate_limit' && !group.rateLimit)`). Both use the same sentinel mechanism: the expired-but-non-null `group.rateLimit` object serves as the re-detection sentinel, as documented in the `scheduleTickAfterRateLimitReset()` JSDoc (lines 3853-3857). The fallback model path (`switched = true`) falls through to normal completion, and the no-fallback path sets the backoff and returns early -- both are correctly wrapped inside the guard.

The control flow after the guard is also correct. When `group.rateLimit` is already set, the code falls through past the entire `usage_limit` `if` block (line 1140) and continues to the method's normal completion path (line 1144: "Leader can finish without calling a tool - that's fine"). This is the desired behavior -- the leader should finish cleanly without re-applying the backoff.

### Tests

The two new tests adequately cover both the active and expired `rateLimit` scenarios:

- **Active rate limit test**: Verifies `group.rateLimit` is not overwritten and task stays `in_progress`.
- **Expired rate limit test**: Verifies the expired backoff is preserved and task does not re-transition to `usage_limited`.

Both tests use `getGlobalSettings: () => ({}) as never` to prevent fallback model switching, ensuring the no-fallback path is tested (which is the path that would cause the infinite loop). All 12 tests in the file pass.

### Minor Observations (P3)

1. **Comment inconsistency (line 1058-1060 vs line 1095-1098)**: The `rate_limit` handler's comment (line 1058-1060) states "there is no recoverStuckLeaders mechanism that re-calls this handler," while the new `usage_limit` handler's comment (line 1095-1098) says the guard "prevents an infinite loop when recoverStuckLeaders re-triggers onLeaderTerminalState." The `rate_limit` comment appears to predate the introduction of `recoverStuckLeaders` (or the understanding that `scheduleTickAfterRateLimitReset` can effectively re-trigger this handler via the observer). This is a documentation inconsistency, not a code bug -- the code is correct in both paths.

2. **Line 1138 comment**: "re-trigger after expiry" is slightly misleading since the code immediately falls through to normal completion (no re-triggering happens). A more accurate comment would be: "group.rateLimit already set (even if expired): limit was already handled. Fall through to normal completion." This is purely cosmetic.

### Verdict

The implementation correctly addresses the reported infinite loop issue. The guard pattern matches the existing `rate_limit` handler, tests cover both active and expired scenarios, and the control flow is sound. No P0/P1/P2 issues found.

**Recommendation: APPROVE**

PR #893, opened 2026-03-25
feat(room-runtime): check provider availability before fallback model switch (Gap G)

lsm:


lsm (commented):

---

### P1: Missing error handling for `isProviderAvailable` callback throws

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 390-398

The `isProviderAvailable` callback call is not wrapped in try/catch:

```ts
if (this.isProviderAvailable) {
    const available = await this.isProviderAvailable(candidate.provider, candidate.model);
    if (!available) {
        // ...
        continue;
    }
}
```

If the callback throws (e.g., a network timeout while checking provider health, DNS failure, etc.), the exception propagates up through `onWorkerTerminalState`. While the observer-level `.catch()` at the call sites (lines 1719, 2423, 2637, 3390, 3657) prevents a process crash, it **aborts the entire worker terminal state handler**. This means:

1. The rate-limit backoff/pause that should follow is never set.
2. The worker output is never routed to the leader.
3. The task is left in `in_progress` with no recovery mechanism until the next tick.

The fix is straightforward -- wrap the availability check in try/catch inside `trySwitchToFallbackModel` and treat a thrown error as "unavailable" (log a warning and continue to the next candidate):

```ts
if (this.isProviderAvailable) {
    try {
        const available = await this.isProviderAvailable(candidate.provider, candidate.model);
        if (!available) {
            log.warn(
                `Skipping fallback ${candidate.provider}/${candidate.model} — provider unavailable`
            );
            continue;
        }
    } catch (err) {
        log.warn(
            `Provider availability check failed for ${candidate.provider}/${candidate.model}: ${String(err)}`
        );
        continue;
    }
}
```

There is also no test covering this scenario -- the callback-throws edge case should be tested to ensure it falls through to the next fallback rather than propagating.

---

### P2: Behavioral change in same-model detection (old code gave up; new code tries next)

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 373-401

The old code, when `currentIndex === -1` (current model not in chain), would pick `fallbackModels[0]` and then check if it was the same as the current model. If it was the same, it would return `false` immediately -- giving up entirely.

The new code, when the current model happens to match `fallbackModels[0]`, will skip it and try `fallbackModels[1]`. This is actually an **improvement** and is explicitly tested in the "same-model skip guard" test case. However, this behavioral change is not mentioned in the commit message or PR description. For anyone looking at the git history, it appears incidental rather than intentional. Consider adding a brief note in the commit message.

---

### P3 (nit): Test helper uses `as never` for messageHub

**File:** `packages/daemon/tests/unit/room/room-runtime-test-helpers.ts`, line 311

```ts
messageHub: opts?.messageHub as never,
```

The `as never` cast bypasses type checking entirely. Consider using a more targeted type assertion like `as unknown as MessageHub` which is the pattern already used for `daemonHub` on line 312, or defining a minimal `MessageHub` interface that the mock satisfies. This is cosmetic for test helpers but would catch shape mismatches if the `MessageHub` interface changes.

---

### What's done well

- **Clean, focused diff** -- only 3 files changed, single commit, no scope creep.
- **Backward compatible** -- optional callback, no production caller wired yet, all existing tests pass.
- **Good test coverage** -- 7 test cases covering: backward compat, available, unavailable, all-unavailable, chain-index traversal, end-of-chain, and same-model skip guard.
- **Correct mock setup** -- the `switchModel` method was properly added to `createMockSessionFactory` with the correct return type matching the `SessionFactory` interface.
- **The loop refactor** is cleaner than the original if/else branch structure and is easier to extend.

lsm (commented):

### Recommendation: APPROVE

---

### Previous issues -- all resolved

**P1 (isProviderAvailable error handling): Fixed.** The callback call at line 391-404 of `room-runtime.ts` is now wrapped in try/catch. A thrown error logs a warning and continues to the next candidate via `continue`, exactly as proposed. Two new tests cover: (1) throws on first candidate, falls through to second, and (2) throws on all candidates, `usage_limited` status is set correctly. The fix prevents the exception from aborting `onWorkerTerminalState` and leaving the task stuck in `in_progress`.

**P2 (undocumented behavioral improvement): Fixed.** The second commit message (`8599e8a4`) explicitly documents the same-model detection change -- the old code returned false when `fallbackModels[0]` matched the current model (`currentIndex === -1`), while the new code skips to the next candidate. The commit message notes this is intentional and tested.

**P3 (`as never` cast): Fixed.** Line 312 of `room-runtime-test-helpers.ts` now uses `as unknown as MessageHub`, consistent with the `daemonHub` pattern on line 313.

### New review findings

No new issues. The implementation is clean, well-tested, and backward compatible.

- All 1529 tests in `packages/daemon/tests/unit/room/` pass with zero failures
- TypeScript typecheck passes clean
- Lint passes with zero warnings/errors
- The 2 new tests bring total test cases in the provider-availability file to 9, covering backward compat, available provider, skip unavailable, all-unavailable, chain index traversal, end-of-chain, callback throws (partial), callback throws (all), and same-model skip guard

PR #894, opened 2026-03-25
feat(task-handlers): clear group rate limit when sending message to rate/usage-limited task

lsm:
When a user sends a human message to a rate_limited or usage_limited task,
call runtime.clearGroupRateLimit() before routing the message. This clears
group.rateLimit, restores task status to in_progress, and allows the message
to be forwarded normally via routeHumanMessageToGroup(). If clearGroupRateLimit
returns false (no group found), the handler continues anyway.

Adds 4 unit tests covering both status variants, the false-return case, and
the in_progress no-op path.

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
@@ -1068,6 +1068,15 @@ export function setupTaskHandlers(
 			}
 		}
 
+		// rate_limited/usage_limited tasks: clear the group's rate limit so the message can be
+		// routed normally. clearGroupRateLimit() also restores the task status to in_progress
+		// and clears any task restrictions. Falls through to the generic
+		// routeHumanMessageToGroup() call below. If clearGroupRateLimit returns false (no
+		// group found), we continue anyway — routeHumanMessageToGroup will surface the error.
+		if (task.status === 'rate_limited' || task.status === 'usage_limited') {
+			await runtime.clearGroupRateLimit(taskId);
+		}
+
 		const groupRepo = makeGroupRepo();
 		const result = await routeHumanMessageToGroup(
 			runtime,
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

---

### Correctness

The implementation correctly addresses Task 2.3 from the plan. The new block at line 1071-1078 of `task-handlers.ts` is placed in the right location (after the `review` special-case block, before the generic `routeHumanMessageToGroup` call), and the logic is sound:

- It checks for both `rate_limited` and `usage_limited` statuses.
- It properly `await`s the async `clearGroupRateLimit(taskId)` method.
- It falls through to `routeHumanMessageToGroup()` after clearing, which will now work correctly since `group.rateLimit` and task restrictions have been cleared.
- If `clearGroupRateLimit` returns `false` (no group found), the handler continues to `routeHumanMessageToGroup`, which will independently detect the missing group and return a descriptive error. This is the correct graceful degradation behavior.

### Architecture

The change follows the existing pattern established by the `review` block (lines 1058-1069) -- a simple conditional block that pre-processes the task state before falling through to generic routing. The `clearGroupRateLimit` method was added in PR #891 and is a public method on `RoomRuntime`, so it is correctly accessible via the already-validated `runtime` reference.

### Tests

The 4 new tests adequately cover the key behavioral scenarios:

1. `rate_limited` task calls `clearGroupRateLimit` -- verifies the method is invoked with the correct taskId.
2. `usage_limited` task calls `clearGroupRateLimit` -- verifies both status variants trigger the clear.
3. `clearGroupRateLimit` returning `false` does not block routing -- verifies the continue-on-false semantics.
4. `in_progress` task does NOT call `clearGroupRateLimit` -- verifies the no-op path.

**P3 -- Test mock setup inconsistency (line 715-727):** The test for `clearGroupRateLimit` returning `false` sets the runtime mock to return `false`, implying the runtime's internal `groupRepo.getGroupByTaskId()` found no group. However, the handler then proceeds to `routeHumanMessageToGroup()` which calls a separate `makeGroupRepo()` that finds a group via the mock DB. This is an artifact of the test's independent mock layers (runtime mock vs. DB mock), not a real-world scenario. The test still correctly validates the important behavioral contract: the handler does not throw or short-circuit when `clearGroupRateLimit` returns `false`. No action needed.

All 676 unit tests pass across 19 files in `packages/daemon/tests/unit/rpc-handlers/`.

PR #896, opened 2026-03-25
task/41 add usagelimit detection to the mirroring callb

lsm:
Add real-time `usage_limit` detection to the `setupMirroring()` callback in `room-runtime.ts`.

## Changes

- **Real-time detection**: mirroring callback now handles `usage_limit` errors alongside `rate_limit`
- **Fallback switch**: immediately attempts `trySwitchToFallbackModel()` on detection
  - Success: appends `model_fallback` event, clears stale task restriction
  - Failure: sets backoff, persists `usage_limited` task status, schedules tick
- **Race condition guard**: `fallbackAttempted.add()` called synchronously before the async call — prevents duplicate `switchModel` calls if two messages arrive before the first `.then()` resolves
- **Re-detection guard**: skips if `freshGroup.rateLimit !== null` (prior handler already active)
- **Fresh state reads**: all mutable state reads use `groupRepo.getGroup()` instead of closure-captured `group`
- **Cleanup**: `fallbackAttempted.clear()` added to the mirroring cleanup callback
- **Consistency**: `rate_limit` handler updated to use `freshGroup` references

## Tests

14 new unit tests in `room-runtime-mirroring-usage-limit.test.ts` covering:
- No-fallback path: backoff set, task restricted to `usage_limited`
- Fallback success: `model_fallback` event (exactly 1), task restriction cleared
- Race condition: two messages fired synchronously → only one `switchModel` call
- Re-detection guard: second message skipped when `rateLimit` already set
- Leader session detection
- Cleanup reset: after `cleanupMirroring` + `setupMirroring`, fresh guard allows new attempt

Also adds `hub.fire()` to mock daemon hub and `switchModel` + `messageHub` support to test helpers.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2328,8 +2328,12 @@ export class RoomRuntime {
 						if (!freshGroup) return;
 						if (freshGroup.rateLimit !== null) return;
 
-						// Duplicate fallback guard: only attempt once per session per detection cycle
+						// Duplicate fallback guard: added synchronously BEFORE the async call so
+						// that two distinct messages arriving before the first .then() resolves
+						// cannot both pass the guard and trigger duplicate trySwitchToFallbackModel
+						// calls.
 						if (fallbackAttempted.has(fallbackKey)) return;
+						fallbackAttempted.add(fallbackKey);
 
 						log.info(
 							`Usage limit detected in ${role} message for group ${group.id} (real-time). ` +
@@ -2343,7 +2347,6 @@ export class RoomRuntime {
 										`Fallback model switch succeeded for ${sessionRole} session ${sessionId} ` +
 											`in group ${group.id} (mirroring).`
 									);
-									fallbackAttempted.add(fallbackKey);
 									// Clear any stale task restriction so the UI reflects the new model
 									this.clearTaskRestriction(freshGroup.taskId).catch((err: unknown) => {
 										log.error(
@@ -2354,9 +2357,8 @@ export class RoomRuntime {
 									// No fallback available — apply backoff and restrict the task
 									log.info(
 										`No fallback model available for ${sessionRole} session ${sessionId}. ` +
-											`Applying usage limit backoff for group ${group.id}.`
+											`Applying usage limit backoff for group ${freshGroup.id}.`
 									);
-									fallbackAttempted.add(fallbackKey);
 									const rateLimitBackoff = msgErrorClass.resetsAt
 										? createRateLimitBackoff(messageContent, sessionRole)
 										: null;
@@ -2365,8 +2367,8 @@ export class RoomRuntime {
 										resetsAt: Date.now() + 60 * 1000,
 										sessionRole,
 									};
-									this.groupRepo.setRateLimit(group.id, backoff);
-									this.appendGroupEvent(group.id, 'rate_limited', {
+									this.groupRepo.setRateLimit(freshGroup.id, backoff);
+									this.appendGroupEvent(freshGroup.id, 'rate_limited', {
 										text: `Usage limit reached. Pausing until ${new Date(backoff.resetsAt).toLocaleTimeString()}.`,
 										resetsAt: backoff.resetsAt,
 										sessionRole,
@@ -2381,10 +2383,12 @@ export class RoomRuntime {
 											`Failed to persist usage limit restriction for task ${freshGroup.taskId}: ${String(err)}`
 										);
 									});
-									this.scheduleTickAfterRateLimitReset(group.id);
+									this.scheduleTickAfterRateLimitReset(freshGroup.id);
 								}
 							})
 							.catch((err: unknown) => {
+								// fallbackAttempted was already set synchronously above, so even
+								// unexpected throws won't leave the guard open for retry loops.
 								log.error(
 									`Error attempting fallback model switch for session ${sessionId}: ${String(err)}`
 								);
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2274,6 +2274,10 @@ export class RoomRuntime {
 		if (!this.daemonHub) return;
 
 		const mirroredUuids = new Set<string>();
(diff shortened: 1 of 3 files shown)


lsm (commented):

The implementation correctly adds `usage_limit` detection to the mirroring callback with the right overall structure, but has one P0 race condition and a few P1 issues that should be addressed before merge.

---

### P0: `fallbackAttempted` is set too late -- race condition allows duplicate `trySwitchToFallbackModel` calls

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 2284-2298

The `fallbackAttempted.add(fallbackKey)` calls at lines 2298 and 2311 are inside `.then()` (async), but the guard `fallbackAttempted.has(fallbackKey)` at line 2284 runs synchronously before `.then()` is scheduled. If two distinct `sdk.message` events with different UUIDs (both containing usage_limit text) fire before the first `.then()` resolves, both pass the guard and both call `trySwitchToFallbackModel()`, causing duplicate `switchModel` calls and potentially duplicate `model_fallback` events.

The `mirroredUuids` set only deduplicates same-UUID messages, so distinct messages bypass it entirely.

**Fix:** Move `fallbackAttempted.add(fallbackKey)` to before the async call (immediately after line 2289, before `this.trySwitchToFallbackModel(...)`). This ensures the synchronous guard blocks subsequent calls even before the promise resolves. The `add()` calls inside `.then()` can then be removed.

---

### P1: `.catch()` handler does not mark `fallbackAttempted` -- retry loop if `trySwitchToFallbackModel` unexpectedly throws

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 2339-2343

The `.catch()` handler only logs the error but does not add to `fallbackAttempted`. While `trySwitchToFallbackModel()` internally catches errors and returns `false`, any unexpected throw (e.g., from `this.getGlobalSettings()` or a future code path change) would leave the key out of the set, allowing subsequent messages to retry indefinitely. The `.catch()` handler should also call `fallbackAttempted.add(fallbackKey)`.

---

### P1: `model_fallback` event is appended by `trySwitchToFallbackModel()` internally -- test assertion is correct but incomplete

**File:** `packages/daemon/tests/unit/room/room-runtime-mirroring-usage-limit.test.ts`, line 179

The test "appends model_fallback event on successful switch" only checks that a `model_fallback` event exists using `.find()`. It should assert `.toHaveLength(1)` to catch the case where duplicate events are appended. This is particularly relevant given the P0 race condition above.

---

### P2: Dead code -- `makeSdkMessage()` helper is defined but never used

**File:** `packages/daemon/tests/unit/room/room-runtime-mirroring-usage-limit.test.ts`, lines 24-26

The `makeSdkMessage` function is defined but never called in any test. It should be removed or used.

---

### P2: Test for cleanup is vacuous -- only asserts `expect(true).toBe(true)`

**File:** `packages/daemon/tests/unit/room/room-runtime-mirroring-usage-limit.test.ts`, lines 392-413

The "clears fallbackAttempted set when mirroring is cleaned up" test triggers the fallback, calls `ctx.runtime.stop()`, and then asserts `expect(true).toBe(true)`. This does not actually verify that `fallbackAttempted` was cleared or that cleanup ran correctly. A more meaningful test would verify that after cleanup, a second `usage_limit` event triggers a new fallback attempt (i.e., the guard is reset).

---

### P3: Inconsistent use of `group.id` vs `freshGroup.id` in fallback-failure path

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, lines 2320-2321

In the fallback-failure `else` branch, `this.groupRepo.setRateLimit(group.id, backoff)` and `this.appendGroupEvent(group.id, 'rate_limited', ...)` use the closure-captured `group.id` rather than `freshGroup.id`. While `group.id` is immutable so this is not a bug, it's inconsistent with the `freshGroup.id` / `freshGroup.taskId` usage in the success path (line 2300) and the rate_limit handler (lines 2253-2272).

---

### Summary of what's done well

- The two-layer guard (re-detection via `freshGroup.rateLimit !== null` + duplicate fallback via `fallbackAttempted`) is the right design pattern.
- Fresh state reads via `this.groupRepo.getGroup()` are correctly used for mutable fields, with closure-captured values only for immutable IDs.
- The `.then()/.catch()` async pattern is correctly used in the synchronous callback context.
- The cleanup callback properly clears `fallbackAttempted`.
- Test helper additions (`hub.fire()`, `switchModel` mock, `messageHub` option) are clean and reusable.
- The default backoff (60s) when no parseable reset time exists is a reasonable safety net.

lsm (commented):

### Previous Issues — Verification

**P0 (race condition): FIXED.** The `fallbackAttempted.add(fallbackKey)` call is now at line 2318, synchronously before the async `trySwitchToFallbackModel()` call at line 2325. The redundant `fallbackAttempted.add(fallbackKey)` calls that were inside the `.then()` success and failure branches have both been removed. The race condition test (line 253) now fires both messages synchronously without an intervening `await`, correctly verifying that the synchronous guard blocks the second invocation.

**P1 (.catch handler + event count test): FIXED.** The `.catch()` handler at line 2371 now has a comment explaining why no additional `fallbackAttempted.add()` is needed (it was already set synchronously above). The `model_fallback` event test at line 174 uses `.filter().toHaveLength(1)` instead of `.find().toBeDefined()`, which would pass even with duplicate events. There is exactly one emission site for `model_fallback` (inside `trySwitchToFallbackModel` at line 408), and the mirroring `.then()` success path does not emit a second one, so this assertion is correct.

**P2 (dead code + vacuous cleanup test): FIXED.** `makeSdkMessage()` has been removed. The cleanup test (line 386) now meaningfully verifies that after `cleanupMirroring` + `setupMirroring`, a new usage_limit message triggers a second fallback attempt, confirming that the `fallbackAttempted` set is properly reset.

**P3 (freshGroup consistency): FIXED.** All references in the fallback-failure path (lines 2342, 2352, 2353, 2358, 2368) now use `freshGroup.id` / `freshGroup.taskId` instead of stale closure `group.id`. This is correct since `id` and `taskId` are immutable identity fields, but using the consistently-named variable improves readability and follows the documented pattern in the comment at line 2277.

### New Observations

No new issues introduced. The 14 tests all pass. The fix commit is clean and targeted.

### Recommendation

All previous issues (P0 through P3) have been properly addressed. The race condition fix is correct, the guard semantics are now well-documented in comments, and the tests are more rigorous. Ready to merge.

PR #897, opened 2026-03-25
test: add usage_limit re-detection guard tests for active rate limit scenario

lsm:


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured test that fills a meaningful gap in coverage for the usage_limit re-detection guard.

### What it tests

The new test (lines 353-386) verifies the scenario where `onWorkerTerminalState` is called again while the rate limit set by a previous `usage_limit` detection is still active (future `resetsAt`). It asserts that:

1. `isRateLimited()` at line 634 of `room-runtime.ts` returns `true` and the handler returns early
2. The `rateLimit` object is not overwritten (same `resetsAt` before and after)
3. `feedbackIteration` stays at `0` -- confirming no worker-to-leader routing occurred

### Why it matters

This complements the existing test (lines 313-351) which covers the *expired* rate limit case. The two tests together cover both branches of the re-detection guard:
- **Expired sentinel**: `group.rateLimit` is non-null but past `resetsAt` -- falls through to worktree check and routes to leader (preventing infinite backoff loop)
- **Active rate limit** (this test): `isRateLimited()` returns `true` -- returns early before any detection logic runs (preventing premature routing while rate limit is still valid)

- The test message `"You've hit your limit · resets 1pm (America/New_York)"` uses a time-of-day format that `parseRateLimitReset` must resolve to a future timestamp relative to `Date.now()`. Since the test runs at an arbitrary time, the parser must interpret "1pm" as today's 1pm or tomorrow's 1pm. This is inherently fragile -- if tests ran at exactly 12:59pm the reset would be 1 minute away, and the `isRateLimited` check could pass or fail depending on execution speed. However, inspecting the first assertions confirms `resetsAt > Date.now()`, so the test is self-validating: if the parser ever broke and returned a past timestamp, the test would fail immediately on the `toBeGreaterThan(Date.now())` assertion at line 371.

- The line reference in the comment (`line ~634`) will drift as the file evolves. Consider whether a named constant or method reference would be more durable, though for test documentation this is acceptable.

All 17 tests pass. No issues found.

PR #898, opened 2026-03-25
feat(room-runtime): return early after successful trySwitchToFallbackModel() in onWorkerTerminalState

lsm:
After a successful fallback model switch, clear stale task restrictions and
group rate limit, then return early instead of falling through to normal
routing. This prevents stale error text from being sent to the leader and
ensures the UI shows the task as in_progress with the new model.

When the new query completes, onWorkerTerminalState fires again with clean
output and routes normally (Bug D + Gap H fix).

Also adds switchModel() to the mock session factory and a messageHub option
to RuntimeTestContextOptions to support fallback model tests.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -771,8 +771,12 @@ export class RoomRuntime {
 						this.scheduleTickAfterRateLimitReset(groupId);
 						return;
 					}
-					// Fall through to normal routing — fallback model switch event was already appended
-					// in trySwitchToFallbackModel so the UI shows the switch clearly.
+					// Fallback switch succeeded: clear stale restriction/rate-limit so the UI shows the
+					// task as in_progress with the new model, then return. When the new query finishes,
+					// onWorkerTerminalState fires again with clean output and routes normally.
+					await this.clearTaskRestriction(group.taskId);
+					this.groupRepo.clearRateLimit(groupId);
+					return;
 				}
 				// group.rateLimit already set (even if expired): re-trigger after expiry.
 				// Fall through to the worktree check so the worker can attempt cleanup/retry.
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, focused change that correctly implements the requested behavior. I reviewed the production code, test helpers, and test cases thoroughly.

### What's correct

- **Early return is the right fix.** Previously, after a successful fallback model switch, the worker's usage-limit error text would fall through to routeWorkerToLeader, injecting stale error text into the leader session. The early return at line 771 of room-runtime.ts prevents this -- the worker restarts via switchModel -> lifecycleManager.restart(), and when the new query completes, onWorkerTerminalState fires again with clean output and routes normally.

- **Clearing state is defensive and correct.** On the first invocation, neither clearTaskRestriction nor clearRateLimit has anything to clear (the usage_limit branch only sets rate-limit/restriction in the !switched path). However, these calls are defensive for re-trigger scenarios (e.g., if recoverStuckWorkers or an external re-trigger fires onWorkerTerminalState again while stale state lingers from a prior tick cycle). The existing clearGroupRateLimit public method (line 1823) already uses this exact same pair of calls, confirming the pattern is idiomatic.

- **Test helpers are well-scoped.** Adding switchModel() to the mock session factory mirrors the SessionFactory interface and is used correctly by the new tests. The messageHub option on RuntimeTestContextOptions is minimal and typed.

- **Tests verify the right invariants.** Test 1 ('returns early without routing to leader') verifies: task not paused, group rate-limit null, leader not injected with stale output, and switchModel called with the correct fallback model. Test 2 ('clears stale task restriction after successful fallback switch') verifies idempotency by re-triggering onWorkerTerminalState and confirming restrictions remain clear.

- **No regressions.** All 1525 unit tests across 43 files pass. The leader-side usage_limit handler at line 1134 was intentionally left unchanged -- it falls through to 'leader can finish without calling a tool' which is a harmless early return, not routing to anyone else.

- The mockMessageHub in tests ignores the data parameter of request(method, data). This is fine for current tests but if tests ever need to verify per-session model queries, the signature should be updated to accept both arguments.

- The two new tests duplicate the mockMessageHub construction. A shared factory function (e.g., createFallbackModelMessageHub) could reduce duplication, but the tests are short enough that this is a style preference.

PR #899, opened 2026-03-25
test: add usage_limit re-detection guard tests for leader path

lsm:
Adds two tests to room-runtime-leader-terminal-errors.test.ts:
1. First usage_limit detection on leader sets rate limit backoff and pauses task.
2. Re-trigger after group.rateLimit is set to expired — guard skips re-detection,
   resetsAt stays in the past (no infinite-loop reset to future timestamp).

### DIFF
--- a/packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts
@@ -177,4 +177,60 @@ describe('RoomRuntime - leader terminal error detection', () => {
 			expect(updatedGroup!.rateLimit!.sessionRole).toBe('leader');
 		});
 	});
+
+	describe('usage_limit handling', () => {
+		it('pauses task when usage_limit detected in leader output and no fallback model configured', async () => {
+			// First detection: leader outputs usage limit text → should set rate limit backoff
+			const { task, group } = await spawnAndSimulateLeaderOutput(
+				"You've hit your limit · resets 1pm (America/New_York)"
+			);
+
+			// Task should NOT be failed — it should pause (usage_limited)
+			const updatedTask = await ctx.taskManager.getTask(task.id);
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

The two new tests correctly verify the leader-path usage_limit re-detection guard. All 12 tests pass. The implementation is sound and follows the established patterns from the worker-path tests.

### What's done well

1. **First-detection test** (lines 182-197): Properly verifies that the leader path sets group.rateLimit with a future resetsAt and sessionRole: 'leader' when usage_limit text is first detected. The negative assertion (status !== 'needs_attention') correctly confirms the task is paused rather than failed.

2. **Re-detection guard test** (lines 199-234): Faithfully simulates the production flow described in the comments of room-runtime.ts (lines 1106-1150):
   - First detection sets the rate limit backoff
   - Timer fires, rateLimit sentinel remains with expired resetsAt
   - onLeaderTerminalState is re-triggered
   - Guard skips re-detection, resetsAt stays in the past

3. **Test structure mirrors the worker counterpart**: The worker-path tests in room-runtime-terminal-errors.test.ts (lines 313-351) follow the exact same pattern, providing consistency across the codebase.

- **Comment discrepancy in production code** (not introduced by this PR): The rate_limit guard at room-runtime.ts:1069-1071 states "there is no recoverStuckLeaders mechanism that re-calls this handler," but the usage_limit guard at line 1108 references recoverStuckLeaders, and runtime-recovery.ts:146-151 does call onLeaderTerminalState during recovery. The !group.rateLimit guard on the rate_limit path (line 1076) is defensive and correct, but the comment is misleading. This is out of scope for this test PR.

- **Second test could additionally assert task status**: The worker-path equivalent (line 350) checks feedbackIteration to confirm routing occurred. The leader test only asserts resetsAt <= Date.now(). While this is sufficient to prove the guard works, an additional assertion that the task was not failed (status !== 'needs_attention') would make the test more self-documenting about the expected outcome of the fall-through path.

No blocking issues found.

PR #900, opened 2026-03-25
fix(runtime): return early after successful trySwitchToFallbackModel in leader path

lsm:
After trySwitchToFallbackModel succeeds in onLeaderTerminalState, clear the
group rate limit and task restriction, then return early instead of falling
through to normal completion. This prevents stale error text from being
processed and ensures the UI shows the correct status while the new-model
query is running.

- Clear group.rateLimit via groupRepo.clearRateLimit after successful switch
- Clear task restriction via clearTaskRestriction after successful switch
- Return early so the observer fires again with clean output from new model
- Add switchModel to mock SessionFactory in test helpers
- Add messageHub option to RuntimeTestContextOptions for fallback model tests
- Add 3 unit tests covering: rateLimit cleared, task restriction cleared,
  group not completed (early return verified)

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1148,7 +1148,12 @@ export class RoomRuntime {
 							this.scheduleTickAfterRateLimitReset(groupId);
 							return;
 						}
-						// Fall through to normal completion — fallback model switch event was already appended
+						// Fallback switch succeeded — clear stale restriction/rate-limit so the task
+						// shows the correct status, then return. The observer will fire again when the
+						// new-model query finishes, delivering clean output (no stale error text).
+						this.groupRepo.clearRateLimit(groupId);
+						await this.clearTaskRestriction(group.taskId);
+						return;
 					}
 					// group.rateLimit already set (even if expired): re-trigger after expiry.
 					// Fall through to normal completion so the leader can finish cleanly.
--- a/packages/daemon/tests/unit/room/room-runtime-rate-limit-persistence.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-rate-limit-persistence.test.ts
@@ -425,4 +425,110 @@ describe('RoomRuntime - rate limit restriction persistence', () => {
 			expect(taskAfter!.status).toBe('in_progress');
 		});
 	});
+
+	// ─── leader fallback early return ──────────────────────────────────────────────────────
+
+	describe('leader usage_limit: return early after successful trySwitchToFallbackModel', () => {
+		/**
+		 * Creates a context where trySwitchToFallbackModel succeeds:
+		 * - getGlobalSettings returns a fallback model chain
+		 * - messageHub.request('session.model.get') returns a current model not in the chain
+		 *   (so the first fallback is selected)
+		 * - sessionFactory.switchModel returns success
+		 */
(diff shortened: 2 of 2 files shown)


lsm (commented):

### Verdict: APPROVE

The implementation correctly adds an early return after a successful `trySwitchToFallbackModel()` in the leader `usage_limit` path. The change is small, well-scoped, and the three new unit tests pass (18/18 green). Below are some observations, none blocking.

### Code change (room-runtime.ts, lines 1147-1152)

The early return itself is correct and well-motivated. The leader's "normal completion" path (line 1160-1161) is a no-op comment ("Leader can finish without calling a tool -- that's fine"), so the early return doesn't skip any meaningful processing today. However, it is a good defensive measure: it makes intent explicit and guards against future code being added below the error-classification block that would incorrectly run in this case.

**One observation (P2, non-blocking):** Both `clearRateLimit(groupId)` and `clearTaskRestriction(group.taskId)` are technically no-ops on this code path:
- `clearRateLimit` is a no-op because the `!group.rateLimit` guard at line 1111 ensures we only enter this branch when `group.rateLimit` is already `null`. No code between line 1111 and 1150 calls `setRateLimit` on the success path.
- `clearTaskRestriction` is a no-op because it early-returns when `task.status` is not `rate_limited` or `usage_limited` (line 3946). On the first detection (which is the only time we reach this branch), the task is `in_progress`.

These defensive calls don't cause harm but also don't do anything. They would become meaningful if a prior iteration had set the task to a limited status without setting `group.rateLimit`, but that scenario cannot arise with the current code structure. Consider adding a brief code comment noting these are defensive for future changes, or removing them if you prefer to keep the code minimal and add them when a concrete need arises.

### Test helpers (room-runtime-test-helpers.ts)

The `switchModel` mock on `SessionFactory` correctly matches the `SessionFactory` interface signature (`(sessionId, model, provider) => Promise<{ success: boolean; model: string; error?: string }>`). The `messageHub` option in `RuntimeTestContextOptions` uses `as never` to bypass type checking for the partial mock, which is an acceptable test pattern since `trySwitchToFallbackModel` already uses optional chaining on `this.messageHub`.

### New tests (room-runtime-rate-limit-persistence.test.ts)

All three tests are well-structured:

1. **"clears group rateLimit after successful fallback switch"** -- Correctly verifies the post-condition. The `leaderSessionIdRef` pattern to wire up the mock `getWorkerMessages` after `spawnAndRouteToLeader` is clean. The pre-condition assertion (`rateLimit` is null) is a nice touch.

2. **"clears task restriction after successful fallback switch"** -- This test is the most meaningful: it pre-sets the task to `usage_limited` with restrictions, then verifies `clearTaskRestriction` clears it back to `in_progress`. This validates that the `clearTaskRestriction` call at line 1151 does its job when the task actually is in a limited state.

3. **"does not fall through to normal completion"** -- Checks `completedAt` is null, which is a valid negative assertion to confirm early return. This could be made more direct (e.g., checking that `scheduleTick` was not called, or that no additional group events were appended), but the current assertion is sufficient for the PR scope.

### Minor notes

- The PR description references "Task 3.2" and "same fix as Task 3.1" -- the worker path at line 774 still falls through after a successful `trySwitchToFallbackModel` (comment: "Fall through to normal routing"). If the leader path warrants an early return, it's worth confirming whether the worker path should also get the same treatment for consistency, or documenting why the worker path is intentionally different. (P3 observation -- the worker path falls through to the worktree cleanliness gate and normal routing, which is arguably different from the leader's no-op completion.)

PR #901, opened 2026-03-25
feat(runtime): add recoverStuckLeaders() to re-inject worker messages after leader rate limit expiry

lsm:
Adds stuckLeaderRecoveryInFlight Set and recoverStuckLeaders() method to
RoomRuntime that mirrors recoverStuckWorkers(). When a leader session has an
expired rate limit (sessionRole === 'leader') and is idle, the method clears
the rate limit, clears any task restriction, and injects a continuation message
containing the last worker output so the leader can resume reviewing.

Calls recoverStuckLeaders() from executeTick() alongside recoverStuckWorkers().
Also updates the stale comment on the rate_limit guard that said there was no
recoverStuckLeaders mechanism.

Includes 10 unit tests covering: inject on expiry, rate limit cleared, active
limit skipped, worker-role limit skipped, processing leader skipped, submitted-
for-review skipped, missing session skipped, no-rate-limit skipped, duplicate-
inject guard (rate limit cleared after first recovery), and worker excerpt
included in continuation message.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -2718,8 +2718,12 @@ export class RoomRuntime {
 				continue;
 			}
 
-			// Construct a continuation message summarising the last worker output
-			let continuationMessage = '[Auto-recovery] Resuming leader review after rate limit expired.';
+			// Construct a continuation message summarising the last worker output.
+			// Note: in the normal case lastForwardedMessageId is already updated by
+			// routeWorkerToLeader (~line 933), so getWorkerMessages returns an empty array
+			// here. That is fine — the leader's conversation context already contains the
+			// full worker envelope; the generic message alone is sufficient to resume review.
+			let continuationMessage = `[Auto-recovery] Resuming leader review after rate limit expired (iteration ${group.feedbackIteration}).`;
 			if (this.getWorkerMessages) {
 				const workerMessages = this.getWorkerMessages(
 					group.workerSessionId,
@@ -2741,7 +2745,9 @@ export class RoomRuntime {
 
 			// Clear the expired rate limit and any task restriction
 			this.groupRepo.clearRateLimit(group.id);
-			void this.clearTaskRestriction(group.taskId);
+			void this.clearTaskRestriction(group.taskId).catch((err) => {
+				log.error(`[StuckLeader] Group ${group.id}: clearTaskRestriction threw:`, err);
+			});
 
 			this.appendGroupEvent(group.id, 'status', {
 				text: 'Leader recovered from expired rate limit. Re-injecting worker message.',
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -195,6 +195,13 @@ export class RoomRuntime {
 	private stuckWorkerRecoveryInFlight = new Set<string>();
 
 	/**
+	 * Group IDs whose stuck-leader recovery is currently in-flight.
+	 * Guards against duplicate message injections across successive ticks
+	 * while the leader is being re-activated after a rate/usage limit expiry.
+	 */
+	private stuckLeaderRecoveryInFlight = new Set<string>();
+
+	/**
 	 * Task IDs whose group spawn is currently in-flight (async).
 	 * Guards against concurrent ticks re-attempting spawn for the same task while
 	 * `spawnGroupForTask` is awaiting worktree creation or DB insert — the window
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Verdict: REQUEST_CHANGES

The implementation is well-structured and closely mirrors the `recoverStuckWorkers()` pattern. The code is clean, comments are thorough, and all 10 tests pass alongside the full 1539-test suite. That said, I found a few issues worth addressing before merge.

---

### P1 (should-fix): Missing `waitingForQuestion` test coverage

The `recoverStuckLeaders()` method at line 2704 has a guard:

```ts
if (group.waitingForQuestion) continue;
```

But the test file has **no test case** for this guard. This is a clear skip-path that should be verified. The worker-side recovery tests in `room-runtime-recovery-safety.test.ts` (lines 794-810) cover the analogous guard for workers, and the same expectation should apply here. Add a test that sets `waitingForQuestion` to `true` on the group and asserts no injection occurs.

**File:** `packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts`

---

### P1 (should-fix): Missing test for `clearTaskRestriction` side-effect

The `recoverStuckLeaders()` method at line 2744 fires `void this.clearTaskRestriction(group.taskId)`, which transitions the task status from `rate_limited`/`usage_limited` back to `in_progress`. No test verifies this side-effect. The test at line 77 ("clears the rate limit from the group before injecting") only asserts `group.rateLimit` is null, not that the task status was restored to `in_progress`.

To properly test this, the group setup should set the task status to `rate_limited` (or `usage_limited`) via `persistTaskRestriction` equivalent, then assert the task status is `in_progress` after recovery. This ensures the UI-facing task status is correct, not just the internal group state.

**File:** `packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts`

---

### P2 (important suggestion): Continuation message may contain no worker output in the common case

The continuation message construction at lines 2722-2735 queries worker messages via `getWorkerMessages(group.workerSessionId, group.lastForwardedMessageId)`. However, by the time a leader hits a rate limit, `lastForwardedMessageId` has already been updated (by `routeWorkerToLeader` at line 933). Since the worker is idle at that point, there are no new messages after the forwarded ID. This means `workerMessages` will be empty in the normal case, and the leader receives only the generic text:

```
[Auto-recovery] Resuming leader review after rate limit expired.
```

The leader would need to rely on its conversation history to know what it was reviewing. This is likely fine since the leader's conversation context is preserved, but it is worth:
1. Adding a comment explaining this behavior (why it is safe to send no worker output)
2. Optionally including the iteration number in the continuation message to help the leader orient itself, similar to how `formatWorkerToLeaderEnvelope` includes iteration context

The test mock always returns messages regardless of `afterId`, which masks this behavior. Consider adding a test case where `getWorkerMessages` returns an empty array (realistic scenario) to ensure the generic message alone is sufficient.

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` (lines 2722-2735)

---

### P2 (important suggestion): Test for `interrupted` leader state

All tests use `processingStates.set(leaderSessionId, 'idle')`. The method also allows recovery when the leader is in the `interrupted` state (line 2711). Adding at least one test case with the `interrupted` state would strengthen confidence that this path works correctly.

**File:** `packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts`

---

### P3 (nit): Fire-and-forget `clearTaskRestriction` has no error handling

At line 2744, `void this.clearTaskRestriction(group.taskId)` is fire-and-forget. If the DB operation fails (e.g., transient error), the task status remains `rate_limited` even though the rate limit was cleared from the group. This is a minor concern since `clearTaskRestriction` is idempotent and the next tick would not re-trigger recovery (the rate limit is already null). But a `.catch()` with a log statement would be consistent with the error handling pattern used on `injectMessage` at lines 2754-2755.

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts` (line 2744)

---

### Summary

The implementation correctly mirrors the `recoverStuckWorkers()` pattern and handles the core recovery flow well. The main gaps are in test coverage (missing `waitingForQuestion`, `clearTaskRestriction`, and `interrupted` state cases) and a subtle design question about the continuation message content when no new worker output is available. The P1 items (missing test cases) should be addressed before merge.

lsm (commented):

All previously reported issues have been addressed. Here is a summary of the verification:

**P1-1 (waitingForQuestion guard test):** Added at test line 180. The test calls `setWaitingForQuestion(groupId, true, workerSessionId)`, sets the rate limit, runs a tick, and correctly asserts zero injections. Verified.

**P1-2 (clearTaskRestriction side-effect test):** Added at test line 90. Sets task to `usage_limited`, triggers recovery, awaits the async `clearTaskRestriction` via `setTimeout(r, 10)`, then asserts task status is `in_progress`. Verified.

**P2-1 (empty worker output continuation message):** Two-part fix. (1) The implementation comment at lines 2721-2725 clearly explains that `getWorkerMessages` normally returns empty after `routeWorkerToLeader` updates `lastForwardedMessageId`, and that the generic message alone is sufficient. (2) A new test at line 264 (`'sends generic continuation message when getWorkerMessages returns empty'`) creates a context with `getWorkerMessages: () => []` and asserts the message contains `[Auto-recovery]` but NOT `'Last worker output'`. Verified.

**P2-2 (interrupted processing state test):** Added at test line 164. Sets processing state to `'interrupted'` and verifies injection occurs with `[Auto-recovery]` in the message. Verified.

**P3-1 (.catch() on fire-and-forget clearTaskRestriction):** The implementation at line 2748 now uses `void this.clearTaskRestriction(group.taskId).catch(...)` with `log.error` in the catch handler. Verified.

**Additional observations (no action required):**

The duplicate-injection test (line 229) exercises the primary guard mechanism (rate-limit-cleared prevents re-entry on the second tick) rather than the `stuckLeaderRecoveryInFlight` Set directly. This is acceptable because the in-flight Set is a secondary belt-and-suspenders guard for the case where `injectMessage` is still pending when the next tick fires. The test validates the observable behavior (no duplicates), which is what matters.

All 14 tests pass. The implementation is a clean mirror of `recoverStuckWorkers()` and integrates correctly into `executeTick()`.

**Recommendation: APPROVE**

PR #903, opened 2026-03-25
test: add tests verifying correct behavior after successful fallback switch

lsm:
Adds unit tests for task 3.3: verify correct runtime behavior after a successful `trySwitchToFallbackModel()` call.

## Tests added

**Worker path** (`room-runtime-terminal-errors.test.ts`):
- Regression: stale `rate_limited` task restriction from a prior 429 cycle is cleared after `usage_limit` + fallback switch succeeds

**Leader path** (`room-runtime-leader-terminal-errors.test.ts`):
- No new `injectMessage` to worker after successful fallback switch (stale error text not routed back)
- Task restriction cleared and status is not `usage_limited`/`rate_limited` after switch
- `group.rateLimit` cleared after successful fallback switch
- `spawnAndSimulateLeaderOutput` extended with optional `extraOpts` parameter to support fallback model configuration

All existing usage_limit tests continue to pass (1538 total, 0 failures).

### DIFF
--- a/packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts
@@ -26,6 +26,18 @@ describe('RoomRuntime - leader terminal error detection', () => {
 		return { id: 'msg-1', text, toolCallNames: [] };
 	}
 
+	/** Shared mock messageHub that returns current model info for session.model.get requests. */
+	function createMockMessageHub() {
+		return {
+			request: async (method: string) => {
+				if (method === 'session.model.get') {
+					return { currentModel: 'claude-opus-4-5', modelInfo: { provider: 'anthropic' } };
+				}
+				return undefined;
+			},
+		};
+	}
+
 	/**
 	 * Spawn a group, route worker to leader (worker succeeds), then simulate
 	 * the leader reaching terminal state with the given output text.
@@ -244,14 +256,6 @@ describe('RoomRuntime - leader terminal error detection', () => {
 			// We do this manually (not via spawnAndSimulateLeaderOutput) so we can snapshot the
 			// call state exactly before onLeaderTerminalState fires.
 			let leaderSessionId: string | null = null;
-			const mockMessageHub = {
-				request: async (method: string) => {
-					if (method === 'session.model.get') {
-						return { currentModel: 'claude-opus-4-5', modelInfo: { provider: 'anthropic' } };
-					}
-					return undefined;
-				},
-			};
 
 			ctx = createRuntimeTestContext({
 				getWorkerMessages: (sessionId, _afterMessageId) => {
@@ -264,7 +268,7 @@ describe('RoomRuntime - leader terminal error detection', () => {
 					({
 						fallbackModels: [{ model: 'claude-haiku-4-5', provider: 'anthropic' }],
 					}) as never,
-				messageHub: mockMessageHub,
+				messageHub: createMockMessageHub(),
 			});
 
 			const { task } = await createGoalAndTask(ctx);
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Verdict: REQUEST_CHANGES

The worker-path regression test (line 470 of room-runtime-terminal-errors.test.ts) and the "does NOT route to worker after successful fallback model switch" leader test (line 241 of room-runtime-leader-terminal-errors.test.ts) are well-written and provide genuine coverage. However, two of the three new leader-path tests have a meaningful structural issue.

---

### P1: Leader-path "clears task restriction" test does not exercise clearTaskRestriction

**File:** packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts, lines 319-351

The test comment says "leader has a stale usage_limited restriction (from a prior cycle)" but the test never injects any stale state. It uses spawnAndSimulateLeaderOutput which creates a fresh task with status=in_progress and no restrictions. After trySwitchToFallbackModel succeeds, the code calls clearTaskRestriction(group.taskId) (room-runtime.ts line 1155), but that method has an early-return guard at line 3950:

```ts
if (task.status !== 'rate_limited' && task.status !== 'usage_limited') return;
```

Since the task is in_progress (never set to usage_limited), clearTaskRestriction returns without doing anything. The assertion expect(updatedTask!.restrictions).toBeNull() passes trivially because restrictions were never set in the first place.

**Fix:** Inject stale rate_limited or usage_limited state into the DB (via ctx.db.run) before onLeaderTerminalState fires, similar to the worker-path regression test at line 509 of room-runtime-terminal-errors.test.ts. The challenge is that spawnAndSimulateLeaderOutput creates the context internally, so you would either need to split the helper or use the manual setup pattern from the "does NOT route to worker" test (line 241) and insert the stale state between task creation and onLeaderTerminalState.

### P1: Leader-path "clears group rate limit" test does not exercise clearRateLimit

**File:** packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts, lines 353-379

Same structural issue. The test never sets group.rateLimit before the handler fires, so clearRateLimit(groupId) (line 1154) clears something that was never set. The assertion expect(updatedGroup!.rateLimit).toBeNull() passes trivially.

**Fix:** Call ctx.groupRepo.setRateLimit(group.id, ...) before onLeaderTerminalState to ensure the rate limit sentinel exists and actually gets cleared. You would need to use the manual setup pattern (like line 241) to get a reference to the group before the leader terminal state fires.

---

### P2: mockMessageHub is duplicated across three leader-path tests

**File:** packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts, lines 247-253, 323-329, 356-362

The exact same mockMessageHub object is declared three times. Extract it to a beforeEach or a factory function at the describe block level to reduce duplication and make it easier to update in one place if the interface changes.

### P2: The worker-path "clears stale task restriction after successful fallback switch" test (line 433) has a re-trigger that adds no value

**File:** packages/daemon/tests/unit/room/room-runtime-terminal-errors.test.ts, lines 455-459

After spawnAndSimulateWorkerOutput returns (which already triggered onWorkerTerminalState once with a successful fallback switch), the test triggers onWorkerTerminalState again. The comment says "to simulate a re-trigger while restrictions might be stale." However, after a successful fallback switch in the first call, group.rateLimit is still null (line 778 clears it), and the stale restriction was never set. The second call would just re-detect usage_limit and try trySwitchToFallbackModel again (which would succeed again because claude-opus-4-5 is not in the fallback chain). The assertions after this only verify that restrictions are still null, which is already guaranteed by the first call. This test overlaps significantly with the existing "returns early without routing to leader" test (line 388). Consider removing the re-trigger call or adding a distinct scenario that justifies the re-trigger.

---

### Positive observations

- The worker-path regression test (line 470) correctly injects stale state via direct DB writes and is a genuine regression test that would catch a real bug.
- The "does NOT route to worker" leader test (line 241) with the before/after snapshot of injectMessage calls is well-designed and correctly verifies the no-routing invariant.
- The extraOpts parameter on spawnAndSimulateLeaderOutput is a clean, minimal extension to the test infrastructure.
- All 35 tests across both files pass.

---

### Recommendation

The two P1 issues should be addressed before merging. The leader-path "clears task restriction" and "clears group rate limit" tests currently assert tautologies (things that were never set are null) rather than verifying the cleanup code paths execute correctly. Without pre-existing stale state, these tests would pass even if clearTaskRestriction and clearRateLimit were deleted from the leader path.

lsm (commented):

### Verdict: REQUEST_CHANGES

All 35 tests pass, and the worker-path regression test and the "does NOT route to worker" leader test remain well-written. However, the two P1 issues from the previous review are **not resolved**, and the P2 issues remain as well.

---

### P1 (still present): Leader-path "clears task restriction" test does not exercise clearTaskRestriction

**File:** `packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts`, lines 319-351

The test comment says "leader has a stale usage_limited restriction (from a prior cycle)" but **no stale state is injected**. The test uses `spawnAndSimulateLeaderOutput` which creates a fresh task with `status=in_progress` and `restrictions=null`. When the handler calls `clearTaskRestriction(group.taskId)` (room-runtime.ts line 1155), that method hits its early-return guard at line 3950:

```ts
if (task.status !== 'rate_limited' && task.status !== 'usage_limited') return;
```

Since the task is `in_progress`, the method returns without doing anything. The assertion `expect(updatedTask!.restrictions).toBeNull()` passes trivially because restrictions were never set. This test would pass even if `clearTaskRestriction` were deleted from the leader path.

**Fix:** Before `onLeaderTerminalState` fires, inject stale state into the DB. The worker-path regression test (line 509 of room-runtime-terminal-errors.test.ts) demonstrates the correct pattern with a direct `ctx.db.run` call. Because `spawnAndSimulateLeaderOutput` encapsulates context creation and the handler trigger, you would need to either: (a) split the helper to expose a pre-onLeaderTerminalState hook point, or (b) switch to the manual setup pattern used in the "does NOT route to worker" test (line 241) and insert the stale state between task creation and the leader terminal state call.

### P1 (still present): Leader-path "clears group rate limit" test does not exercise clearRateLimit

**File:** `packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts`, lines 353-379

Same issue. The test never calls `ctx.groupRepo.setRateLimit(...)` before `onLeaderTerminalState`. The `clearRateLimit(groupId)` call at room-runtime.ts line 1154 clears something that was never set. `expect(updatedGroup!.rateLimit).toBeNull()` passes trivially.

**Fix:** Call `ctx.groupRepo.setRateLimit(group.id, { ... })` before `onLeaderTerminalState` fires, similar to how the existing re-trigger test at line 227 sets an expired rate limit. Again, this requires using the manual setup pattern to get a group reference before the handler fires.

---

### P2 (still present): mockMessageHub duplicated three times

**File:** `packages/daemon/tests/unit/room/room-runtime-leader-terminal-errors.test.ts`, lines 247-253, 323-329, 356-362

The identical mockMessageHub object is still declared three times. Extract to a module-level factory function (e.g., `createMockMessageHub()`) or a `beforeEach` to reduce duplication.

### P2 (still present): Redundant re-trigger in worker-path test

**File:** `packages/daemon/tests/unit/room/room-runtime-terminal-errors.test.ts`, lines 455-459

The second `onWorkerTerminalState` call at line 456 still adds no value. After the first call succeeds with a fallback switch (clearing group.rateLimit at line 778), the second call re-enters usage_limit detection (group.rateLimit is null), calls trySwitchToFallbackModel again (which succeeds again since claude-opus-4-5 is not in the fallback chain), and clears restrictions that were already null. This test overlaps significantly with the existing "returns early without routing to leader" test at line 388.

---

### Summary

The core of the PR -- the worker-path regression test and the leader-path no-routing test -- is solid. But the two leader-path tests that were explicitly called out as P1 tautologies in the previous review have not been fixed. They still assert post-conditions that are trivially true rather than verifying the cleanup code paths actually execute.

lsm (commented):

## Summary

All three previously flagged issues (P1 x2, P2) have been properly addressed in commit 79a4de1ea. The tests are now genuinely non-tautological, the mockMessageHub factory is extracted, and all 35 tests pass.

## P1 fixes verified

### "clears task restriction after successful fallback switch in leader path" (leader test, line 323)

The fix is correct. The test now injects stale usage_limited status + non-null restrictions into the DB via ctx.db.run (line 367-370) between the worker routing and the leader terminal state handler. This ensures clearTaskRestriction at production line 1155 actually has work to do -- without the DB injection setting status=usage_limited, the early-return guard at line 3950 would short-circuit. Deleting clearTaskRestriction from line 1155 would cause this test to fail.

### "clears group rate limit after successful fallback switch in leader path" (leader test, line 390)

The fix is correct. The test overrides sessionFactory.switchModel (line 424-437) to call ctx.groupRepo.setRateLimit mid-call, injecting a non-null sentinel during the fallback switch. This means clearRateLimit at production line 1154 actually clears non-null state. Deleting clearRateLimit from line 1154 would cause this test to fail. The same spy pattern is applied to the worker-path test for consistency.

## P2 fix verified

### createMockMessageHub factory extraction

The createMockMessageHub() function (leader test file, line 30-39) is used 3 times (lines 271, 342, 410), replacing previously inline messageHub objects. No inline messageHub definitions remain in the leader test file.

Note: the worker test file still has 2 inline messageHub objects (lines 390-397 and 448-455), which is a minor remaining duplication in a separate file not part of the original P2 scope.

## Test quality observations

The before/after inject-call snapshot pattern in the "does NOT route to worker" test (line 288-314) is well-designed. The extraOpts parameter added to spawnAndSimulateLeaderOutput (line 48) properly allows passing getGlobalSettings and messageHub while preserving the mandatory getWorkerMessages.

## Recommendation: APPROVE

No remaining issues. All P1/P2 concerns from previous reviews are resolved. The tests genuinely verify the behavior they claim to test.

PR #904, opened 2026-03-25
test(rate-limit): add clearGroupRateLimit and manual-status-change integration tests

lsm:
## Summary

- Add test for `clearGroupRateLimit()` after `usage_limit` detection: verifies `group.rateLimit` is cleared and task status returns to `in_progress`
- Add integration test covering the full flow: usage_limit detected → backoff set → `clearGroupRateLimit()` (manual status change) → clean worker messages → `onWorkerTerminalState` routes to leader normally via `injectMessage`
- All 20 tests pass (18 existing + 2 new), full daemon suite: 8081 pass, 0 fail

### DIFF
--- a/packages/daemon/tests/unit/room/room-runtime-rate-limit-persistence.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-rate-limit-persistence.test.ts
@@ -424,6 +424,90 @@ describe('RoomRuntime - rate limit restriction persistence', () => {
 			expect(taskAfter!.restrictions).toBeNull();
 			expect(taskAfter!.status).toBe('in_progress');
 		});
+
+		it('clears group rateLimit and task restriction after usage_limit detection, returns true', async () => {
+			ctx = createRuntimeTestContext({
+				getWorkerMessages: () => makeWorkerMessages(USAGE_LIMIT_MSG),
+				// No fallback models — usage_limit triggers pause behavior
+				getGlobalSettings: () => ({}) as never,
+			});
+
+			const { group, task } = await spawnAndTriggerWorkerTerminal('');
+
+			// Confirm usage_limited state
+			const restricted = await ctx.taskManager.getTask(task.id);
+			expect(restricted!.status).toBe('usage_limited');
+			expect(restricted!.restrictions).toBeDefined();
+			expect(restricted!.restrictions!.type).toBe('usage_limit');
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Verdict: APPROVE

The two new tests are well-structured and directly address the task requirements. All 20 tests pass, and the full daemon suite is green.

### Test 1: `clears group rateLimit and task restriction after usage_limit detection, returns true` (line 428)

This is a clean companion to the existing `rate_limit` variant at line 401. It correctly:
- Sets up a `usage_limit` scenario (no fallback models, so the pause/backoff path fires)
- Verifies the pre-condition state (`usage_limited`, restriction with `type: 'usage_limit'`, group rateLimit present)
- Calls `clearGroupRateLimit` and asserts both the group state and task state are restored
- Uses `getGroup()` directly (vs `getActiveGroups().find()` used in the rate_limit variant) -- this is fine since the group is still active, just a minor style inconsistency

No issues.

### Test 2: Integration test (line 461)

This is the most valuable addition. It exercises the full flow that was the actual bug surface:
1. **usage_limit detected** -> backoff set, task paused
2. **Manual status change** -> `clearGroupRateLimit()` clears backoff
3. **Clean worker messages** injected via mutable `msgRef` (clever pattern)
4. **onWorkerTerminalState** -> routes to leader via `injectMessage`

The test correctly verifies three critical post-conditions:
- Group rateLimit stays null (no re-application of backoff)
- Task status is not `usage_limited` again
- Worker output was actually injected into the leader session (via call count delta on `injectMessage`)

The mutable `msgRef` pattern for switching messages mid-test is well-documented and readable.

- **P2**: The integration test only covers the `usage_limit` path. Since the task description mentions testing both `usage_limited` and `rate_limited`, a symmetric integration test for `rate_limited -> clear -> clean messages -> route to leader` would improve coverage. The existing unit test at line 401 covers `rate_limit` clearing in isolation, but the integration flow (verifying routing actually proceeds after clearing) is only tested for `usage_limit`.

- **P3**: Line 401 uses `getActiveGroups('room-1').find(...)` to read group state, while the new test at line 449 uses `getGroup(group.id)!` directly. The non-null assertion is safe here but the inconsistency is worth noting.

### Summary

Solid test additions that meaningfully cover the `clearGroupRateLimit()` behavior for the `usage_limit` path and the critical integration flow of manual status change followed by normal worker routing. The tests are clear, well-commented, and use the existing test infrastructure idiomatically.

PR #905, opened 2026-03-25
test(room-runtime): add tests for fallback model availability verification

lsm:
Add/confirm unit tests verifying that `trySwitchToFallbackModel()` respects provider availability.

The existing `room-runtime-provider-availability.test.ts` already covered subtasks 1–4 and 6 from the task spec (9 tests). This PR adds the missing subtask 5: when provider availability check passes but `switchModel()` itself returns `success: false`, `trySwitchToFallbackModel()` returns `false` and leaves the task in `usage_limited` state.

All 10 tests pass.

### DIFF
--- a/packages/daemon/tests/unit/room/room-runtime-provider-availability.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-provider-availability.test.ts
@@ -301,6 +301,39 @@ describe('RoomRuntime - isProviderAvailable in trySwitchToFallbackModel', () => 
 		});
 	});
 
+	describe('switchModel() itself fails after availability check passes', () => {
+		it('returns false when switchModel returns success:false', async () => {
+			ctx = createRuntimeTestContext({
+				getWorkerMessages: () => makeWorkerMessages(USAGE_LIMIT_MSG),
+				getGlobalSettings: makeGlobalSettings([{ model: 'haiku', provider: 'anthropic' }]),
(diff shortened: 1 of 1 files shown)


lsm (commented):

---

**Verdict: APPROVE**

The new test at lines 304-335 of `packages/daemon/tests/unit/room/room-runtime-provider-availability.test.ts` correctly covers subtask 5 (switchModel failure after availability check passes). All 10 tests pass.

**What works well:**

- The test clearly separates concerns: availability check passes (`isProviderAvailable: () => true`), but the actual switch fails (`switchModelImpl` returns `success: false`).
- Assertions verify both the call path (`switchModel` was invoked with `haiku`) and the end state (`task.status === 'usage_limited'`), which matches the implementation at `room-runtime.ts:447-456` (the `else` branch that logs a warning and returns `false`).
- The test reuses the existing test infrastructure cleanly.

**Minor observations (P3, no action required):**

1. The file-level JSDoc comment (lines 1-11) does not mention the new "switchModel failure" scenario. Consider adding a bullet like:
   ```
   - When switchModel itself returns success:false, the switch fails and task remains usage_limited
   ```

2. The test uses a single fallback entry, so when `switchModel` fails there are no more candidates to try. An additional test covering the case where `switchModel` fails for the first candidate but succeeds for the second would be more thorough. This is not blocking -- the current test adequately covers the subtask as specified.

PR #906, opened 2026-03-25
feat(room-runtime-service): wire ProviderRegistry.isAvailable() into RoomRuntime (Gap G)

lsm:
Wire `ProviderRegistry.isAvailable()` into the production `RoomRuntime` config so `trySwitchToFallbackModel()` verifies provider credentials before switching.

- Pass `isProviderAvailable` callback in `createOrGetRuntime()` using `getProviderRegistry().get(providerId)?.isAvailable()`
- Import `getProviderRegistry` alongside the existing `inferProviderForModel` import
- Add 5 unit tests verifying the callback returns true/false based on real registry state

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -36,7 +36,7 @@ import { SDKMessageRepository } from '../../../storage/repositories/sdk-message-
 import { recoverRuntime, type SessionStateChecker } from './runtime-recovery';
 import type { RoomManager } from '../managers/room-manager';
 import { WorktreeManager } from '../../worktree-manager';
-import { inferProviderForModel } from '../../providers/registry';
+import { inferProviderForModel, getProviderRegistry } from '../../providers/registry';
 import { Logger } from '../../logger';
 
 const log = new Logger('room-runtime-service');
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean, minimal wiring change that completes the integration between ProviderRegistry.isAvailable() and RoomRuntime.trySwitchToFallbackModel(). The diff is exactly what the task asks for: 5 lines of production code plus a well-structured test block.

### What was verified

1. **Correctness**: The callback at room-runtime-service.ts:574-578 correctly delegates to the global ProviderRegistry - returns false for unregistered providers, wraps the result in Boolean(...) for safety, and properly awaits the potentially-async isAvailable().

2. **Integration**: The callback is wired into RoomRuntime via the isProviderAvailable config slot that was added in the prior PR (#893). The consumer at room-runtime.ts:397-412 has try/catch around the call, so errors from provider.isAvailable() (e.g., network timeouts checking credential validity) are handled gracefully with a log-and-skip.

3. **Tests**: All 5 tests pass. The makeCallback() helper mirrors the production logic exactly, and the test uses the real ProviderRegistry singleton (reset in afterEach). Tests cover the key cases: unregistered provider, available, unavailable, async function form, and cross-provider isolation.

4. **Code quality**: Lint and typecheck both pass cleanly. No unused imports. The _model parameter is intentionally underscored since the callback signature requires it but the registry lookup is provider-only, which is correct per the ProviderRegistry.get() API.

5. **No issues found**: No bugs, no security concerns, no missing error handling. The backward-compatible optional-chaining in room-runtime.ts ensures that existing RoomRuntime instantiations without this callback continue to work.

One minor observation (P3): the test file added afterEach to the import but makeCallback() and the mock factory are defined locally in the describe block. This is fine since the test block is small, but if more ProviderRegistry integration tests are added to this file in the future, extracting the helper functions to a shared fixture would reduce duplication.

PR #907, opened 2026-03-25
test(runtime): add missing tests for recoverStuckLeaders behavior

lsm:
Adds 3 new unit tests to `room-runtime-recover-stuck-leaders.test.ts` for the two subtasks not covered by the 14 existing tests:

- **Coexistence (subtask 4)**: `recoverStuckLeaders` and `recoverStuckWorkers` both fire in the same tick for different groups without interfering — two-group scenario with expired worker vs expired leader rate limits.

- **Fresh output after re-injection (subtask 5a)**: after `recoverStuckLeaders` clears the rate limit and injects, `onLeaderTerminalState` with clean output completes normally with no new backoff.

- **Re-detection guard (subtask 5b)**: the `!group.rateLimit` guard (Milestone 1) prevents the `usage_limit` block from resetting backoff when the sentinel is still present and stale error text appears in leader output.

Total: 17 tests (14 existing + 3 new), all passing. Full unit suite: 8110 pass, 0 fail.

### DIFF
--- a/packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts
@@ -380,6 +380,8 @@ describe('recoverStuckLeaders', () => {
 
 		ctxCoexist.runtime.start();
 		await ctxCoexist.runtime.tick();
+		// Allow fire-and-forget worker routing (recoverStuckWorkers) to complete
+		await new Promise((r) => setTimeout(r, 20));
 
 		// recoverStuckLeaders must have injected a continuation message for group2's leader
 		const group2LeaderInjects = ctxCoexist.sessionFactory.calls.filter(
@@ -388,16 +390,17 @@ describe('recoverStuckLeaders', () => {
 		expect(group2LeaderInjects).toHaveLength(1);
 		expect(group2LeaderInjects[0].args[1]).toContain('[Auto-recovery]');
 
-		// recoverStuckLeaders must NOT have injected for group1's leader
-		// (group1.rateLimit.sessionRole === 'worker', not 'leader')
-		const group1AutoRecoveryInjects = ctxCoexist.sessionFactory.calls.filter(
+		// recoverStuckWorkers must have triggered worker→leader routing for group1:
+		// injectMessage is called on group1's leader with a normal worker envelope
+		// (not an [Auto-recovery] message — that would come from recoverStuckLeaders)
+		const group1RoutingInjects = ctxCoexist.sessionFactory.calls.filter(
 			(c) =>
 				c.method === 'injectMessage' &&
 				c.args[0] === group1.leaderSessionId &&
 				typeof c.args[1] === 'string' &&
-				(c.args[1] as string).includes('[Auto-recovery]')
+				!(c.args[1] as string).includes('[Auto-recovery]')
 		);
-		expect(group1AutoRecoveryInjects).toHaveLength(0);
+		expect(group1RoutingInjects.length).toBeGreaterThanOrEqual(1);
 
 		// group2 rate limit must be cleared by recoverStuckLeaders
 		expect(ctxCoexist.groupRepo.getGroup(group2.id)!.rateLimit).toBeNull();
--- a/packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts
@@ -321,4 +321,185 @@ describe('recoverStuckLeaders', () => {
 		ctxWithMessages.runtime.stop();
 		ctxWithMessages.db.close();
 	});
+
+	// ─── Subtask 4: coexistence with recoverStuckWorkers ─────────────────────
+
+	it('recoverStuckLeaders and recoverStuckWorkers both fire in the same tick without interfering', async () => {
+		// Two groups in the same tick:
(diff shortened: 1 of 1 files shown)


lsm (commented):

All 17 tests pass (14 existing + 3 new). The three new tests are well-structured and exercise meaningful integration boundaries. Below is my detailed assessment.

---

### Test 1: Coexistence with recoverStuckWorkers (subtask 4)

**Verdict: Correct, with a suggestion.**

The two-group scenario correctly sets up an expired worker rate limit (group1) and an expired leader rate limit (group2), verifying that `recoverStuckLeaders` only fires for the leader-scoped group. The assertions that group2 gets an `[Auto-recovery]` injection, group1's leader does NOT, and group2's rate limit is cleared are all sound.

**P2 — Missing positive assertion for group1's worker-side recovery.** The test name says "both fire" but only asserts that `recoverStuckLeaders` fired for group2. There is no assertion that `recoverStuckWorkers` actually fired for group1 (e.g., checking that `onWorkerTerminalState` was called for group1, or that group1's worker received an injection). The production code at `room-runtime.ts:2682-2764` would fire for group1 given the setup (feedbackIteration=0, expired rate limit, worker idle, new messages available), but the test does not verify this. Adding a positive assertion for group1's recovery would strengthen the test and make "both fire" match the assertions.

`packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts`, lines 327-407.

### Test 2: Fresh output after re-injection (subtask 5a)

**Verdict: Correct.**

The test properly simulates the full lifecycle: sets up an expired leader rate limit, runs a tick (which triggers `recoverStuckLeaders` to clear the rate limit and inject), marks `leaderHasWork` (simulating what `routeWorkerToLeader` would do), then calls `onLeaderTerminalState` with clean output. The assertions verify no new backoff is set and the task remains `in_progress`.

One note: `ctx.runtime.stop()` is called at line 417 against the `beforeEach` context's runtime (not `ctxFresh`), which is redundant since `afterEach` does the same. This is not a bug but slightly confusing. Consider removing it or adding a comment.

`packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts`, lines 411-458.

### Test 3: Re-detection guard (subtask 5b)

**Verdict: Correct and well-designed.**

This is the strongest of the three new tests. It directly exercises the `!group.rateLimit` guard at `room-runtime.ts:1119` by calling `onLeaderTerminalState` while the rate limit sentinel is still present (expired but non-null) and the output contains stale usage-limit text. The assertion `updatedGroup.rateLimit!.resetsAt <= Date.now()` correctly validates that the backoff was NOT pushed to a new future value, and the task status check confirms no re-pausing occurred.

The test text `"You've hit your limit · resets 2pm (America/New_York)"` correctly matches the `RATE_LIMIT_PATTERN` regex in `rate-limit-utils.ts`, ensuring `classifyError` returns `usage_limit` as expected.

`packages/daemon/tests/unit/room/room-runtime-recover-stuck-leaders.test.ts`, lines 460-504.

---

### Pattern observation across all new tests

**P3 — Inconsistent ctx lifecycle across test-local and beforeEach contexts.** The new tests (and some existing ones like lines 267-292 and 294-323) create their own local context (`ctxCoexist`, `ctxFresh`, `ctxGuard`) and also call `ctx.runtime.stop()` on the `beforeEach` context. This pattern works but introduces an asymmetry: `afterEach` will call `ctx.runtime.stop()` and `ctx.db.close()` on an already-stopped/closed context. The mock sessionFactory's `runtime.stop()` is likely idempotent, but `db.close()` on a closed SQLite database could potentially throw or log warnings depending on Bun's behavior. This is a minor style concern that predates this PR.

---

### Summary

The implementation is correct and achieves the task goals. The three new tests cover important coexistence and re-detection guard scenarios. My only actionable suggestion is the P2 item: adding a positive assertion for group1's worker-side recovery in the coexistence test to fully validate the "both fire" claim.

lsm (commented):

The P2 feedback has been addressed correctly. The fix commit (`60b8461`) makes two changes to the coexistence test:

1. **Added `await new Promise((r) => setTimeout(r, 20))` after `tick()`** — This is necessary and correct. `recoverStuckWorkers` fires asynchronously (fire-and-forget via `void this.onWorkerTerminalState(...)` at `room-runtime.ts:2755`), so without this settle period, the worker-to-leader routing may not have completed when assertions run.

2. **Replaced the negative assertion with a positive assertion for group1** — The old assertion checked that no `[Auto-recovery]` message was injected for group1's leader (which only proved `recoverStuckLeaders` did NOT fire for group1). The new assertion at lines 396-403 checks that `injectMessage` was called on `group1.leaderSessionId` with a string that does NOT include `[Auto-recovery]`, which correctly validates that `recoverStuckWorkers` triggered worker-to-leader routing for group1.

I verified the assertion logic is sound by tracing the production code:
- `recoverStuckWorkers` calls `onWorkerTerminalState` (room-runtime.ts:2755)
- `onWorkerTerminalState` calls `routeWorkerToLeader` (room-runtime.ts:951)
- `routeWorkerToLeader` calls `sessionFactory.injectMessage(leaderSessionId, leaderMessage)` (task-group-manager.ts:475)
- The `leaderMessage` is a formatted worker envelope, never containing `[Auto-recovery]`
- `[Auto-recovery]` only appears in `recoverStuckLeaders` (room-runtime.ts:2817)

All 17 tests pass. No new issues introduced.

**Recommendation: APPROVE**

**P0: 0 | P1: 0 | P2: 0 | P3: 0**

PR #910, opened 2026-03-26
fix(runtime): prevent usage-limit re-detection after manual status clear

lsm:
## Summary

- When a user manually resumes a `usage_limited`/`rate_limited` task, `clearGroupRateLimit` now advances `lastForwardedMessageId` past the stale error messages
- Without this fix, clearing `group.rateLimit` reset the `!group.rateLimit` guard, causing `classifyError` to re-detect the old "You've hit your limit" text and immediately re-apply the backoff on the next tick

## Root cause

`clearGroupRateLimit` cleared the group rate-limit flag and the task restriction but did not advance the message pointer. On the next `onWorkerTerminalState` call, `getWorkerMessages(workerSessionId, lastForwardedMessageId)` still returned the old error messages, and since `group.rateLimit` was now `null` (cleared), the `!group.rateLimit` first-detection guard fired again — re-pausing the task the user just manually resumed.

## Fix

After clearing the rate limit, call `getWorkerMessages` to collect the stale messages and advance `lastForwardedMessageId` to the last one. The next routing pass then sees an empty message list, skips the error-classification block, and routes normally.

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -1876,8 +1876,31 @@ export class RoomRuntime {
 		const group = this.groupRepo.getGroupByTaskId(taskId);
 		if (!group || group.completedAt !== null) return false;
 
+		// clearRateLimit does not increment the group version (raw metadata update),
+		// so group.version remains valid for the updateLastForwardedMessageId call below.
 		this.groupRepo.clearRateLimit(group.id);
 		await this.clearTaskRestriction(taskId);
+
+		// Advance lastForwardedMessageId past any stale error messages so that the
+		// next onWorkerTerminalState call does not re-detect them via classifyError.
+		// Without this, clearing group.rateLimit resets the !group.rateLimit guard,
+		// causing the re-detection to re-apply the backoff on the very next tick.
+		if (this.getWorkerMessages) {
+			const staleMessages = this.getWorkerMessages(
+				group.workerSessionId,
+				group.lastForwardedMessageId
+			);
+			const lastStaleMessage = staleMessages.at(-1);
+			if (lastStaleMessage) {
+				this.groupRepo.updateLastForwardedMessageId(group.id, lastStaleMessage.id, group.version);
+				log.info(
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

This is a clean, well-targeted bugfix. The root cause is clearly understood and the fix is minimal and correct.

**The bug:** `clearGroupRateLimit` clears `group.rateLimit` (the guard at line 723 of `room-runtime.ts`) but does not advance `lastForwardedMessageId`. On the next `onWorkerTerminalState` call, `getWorkerMessages` returns the same stale error messages, `classifyError` re-detects the rate/usage limit, and since `!group.rateLimit` is now `true` again (it was just cleared), the backoff is immediately re-applied -- undoing the manual clear.

**The fix:** After clearing the rate limit, fetch remaining unforwarded messages via `getWorkerMessages` and advance `lastForwardedMessageId` past them. This way the next `onWorkerTerminalState` gets an empty message list, classifyError receives the harmless sentinel string, and no re-detection occurs.

### What I verified

1. **Version safety**: `clearRateLimit` uses a raw metadata update (no version increment), so the `group.version` captured before the call remains valid for the subsequent `updateLastForwardedMessageId` call which does use optimistic locking. The inline comment correctly documents this.

2. **Null guard on `getWorkerMessages`**: The method is optional (`RoomRuntimeConfig` declares it as `getWorkerMessages?:`). The `if (this.getWorkerMessages)` guard correctly handles contexts where it is absent -- the method falls through to the existing log + return path.

3. **Test quality**: Both new tests are well-structured:
   - First test verifies the pointer advancement directly.
   - Second test is the true regression test -- it uses a `getWorkerMessages` mock that respects `afterMessageId` filtering, then re-triggers `onWorkerTerminalState` and asserts no re-detection. This directly reproduces the bug scenario.

4. **All 22 tests in the file pass.**

### Minor observation (P3)

- The string comparison `m.id > afterMessageId` in the test mock (line 492) works because the IDs are `msg-1` format, but this is a lexicographic comparison, not numeric. If test message IDs ever grew beyond single digits (e.g., `msg-10`), they would sort incorrectly. This is fine for the current test but worth noting if the helper is reused. Not actionable now.

PR #912, opened 2026-03-26
test(session-cache): unit tests for room session registration + remove() race condition fix

lsm:
## Summary
- Fix `SessionCache.remove()` to clear `sessionLoadLocks`, preventing a concurrent `getAsync()` from re-inserting a stale session after the session is removed
- Add `SessionManager.unregisterSession()` delegating to `SessionCache.remove()`
- Add 13 unit tests covering all Bug 1 registration safeguards: register/unregister round-trip, concurrent access guard, restore path, and remove() race condition

### DIFF
--- a/packages/daemon/src/lib/session/session-cache.ts
+++ b/packages/daemon/src/lib/session/session-cache.ts
@@ -136,10 +136,15 @@ export class SessionCache {
 	}
 
 	/**
-	 * Remove a session from the cache
+	 * Remove a session from the cache.
+	 *
+	 * Also clears any in-flight load lock for this session ID so that a
+	 * concurrent getAsync() that has already started loading does NOT
+	 * re-insert a stale session into the cache after the remove completes.
 	 */
 	remove(sessionId: string): void {
 		this.sessions.delete(sessionId);
+		this.sessionLoadLocks.delete(sessionId);
 	}
 
 	/**
--- a/packages/daemon/src/lib/session/session-cache.ts
+++ b/packages/daemon/src/lib/session/session-cache.ts
@@ -28,6 +28,13 @@ export class SessionCache {
 	// which would create duplicate Claude API connections
 	private sessionLoadLocks = new Map<string, Promise<AgentSession | null>>();
 
+	// FIX: remove() race condition
+	// Tracks sessions that were explicitly removed while a load was in flight.
+	// Without this, the in-flight load would re-insert the session into `sessions`
+	// after remove() cleared it (the guard at getAsync line 99 only blocks set()
+	// races, not remove() races).
+	private removedWhileLoading = new Set<string>();
+
 	constructor(
 		private createAgentSession: AgentSessionFactory,
 		private loadFromDB: SessionLoader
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Summary

The production code changes are small and sensible: `SessionCache.remove()` now clears `sessionLoadLocks`, and `SessionManager.unregisterSession()` is added. The 13 tests are well-structured and all pass. However, there is an incomplete race condition fix that the tests do not catch, and the `SessionManager` delegation tests don't actually test `SessionManager`.

### P0: `remove()` does not fully prevent stale session re-insertion

**File:** `packages/daemon/src/lib/session/session-cache.ts`, lines 92-103

The fix adds `sessionLoadLocks.delete()` to `remove()`, which prevents **new** `getAsync()` callers from joining a stale load. But the **original caller** that already started the load still re-inserts the stale session:

1. `getAsync('s1')` starts, calls `loadSessionAsync()` (line 89), sets lock (line 90)
2. `remove('s1')` is called — clears `sessions` and `sessionLoadLocks`
3. The already-in-flight `loadPromise` resolves with a non-null `agentSession`
4. Line 99: `!this.sessions.has('s1')` is `true` (because `remove()` cleared it)
5. Line 100: `this.sessions.set('s1', agentSession)` — **stale session is re-inserted**

The guard at line 99 was designed for the `set()` race (registerSession winning over a DB load), not for the `remove()` race. To fully fix this, `remove()` would need to mark the session as "intentionally removed" so `getAsync()` skips the insertion — e.g., by adding the sessionId to a `removedDuringLoad` set that gets checked at line 99 and cleaned up in the `finally` block.

The test at lines 62-115 (`session-registration.test.ts`) acknowledges this in comments at lines 96-101 ("the guard would NOT block the insertion") but does not assert that the stale session is NOT in the cache after `loadPromise` resolves. It only checks the lock is cleared, which is the weaker guarantee. This means the test passes but the core bug (stale re-insertion) is not covered.

### P1: `SessionManager` delegation tests don't actually test `SessionManager`

**File:** `packages/daemon/tests/unit/room/session-registration.test.ts`, lines 344-398

The describe block says "SessionManager — registerSession() / unregisterSession()" but the tests operate directly on `SessionCache`. The comment explains this is because `SessionManager` has heavy dependencies, but this means the actual delegation (`SessionManager.unregisterSession` -> `SessionCache.remove`) is never verified. If someone changes `unregisterSession()` to call a different method, these tests would still pass.

Consider either: (a) adding a thin integration test that constructs `SessionManager` with mocks, or (b) renaming the describe block to clarify these are `SessionCache` tests that validate the contract `SessionManager` relies on.

### P2: `unregisterSession()` has no callers yet

**File:** `packages/daemon/src/lib/session/session-manager.ts`, line 271

`unregisterSession()` is defined but never called anywhere in the codebase. While the task specifically asked for it, this is effectively dead code that Knip may flag. If this is intentional preparation for a follow-up PR, a brief comment like `// Used by TaskAgentManager (see PR #NNN)` would clarify intent.

### P3: Unused mock in first test

**File:** `packages/daemon/tests/unit/room/session-registration.test.ts`, line 71

`slowLoader` is defined but never used (the test uses `blockingLoader` instead). This would likely be caught by the linter.

lsm (commented):

**Recommendation: APPROVE**

The implementation is solid. The `removedWhileLoading` set in `SessionCache` correctly addresses the race condition where `remove()` during an in-flight `getAsync()` load would allow the stale session to be re-inserted. The fix is minimal, well-documented, and the cleanup in the `finally` block prevents memory leaks. All 12 tests pass and cover the key scenarios: the race condition fix, concurrent access guards, restore path, and SessionManager delegation.

### What looks good

- **P0 fix is correct**: `removedWhileLoading` is added only when there is an active lock (`sessionLoadLocks.has()`), checked in the guard at line 111, and cleaned up in `finally` at line 121. This covers the exact race window.
- **No over-engineering**: The fix is three lines of logic in `remove()` and one additional condition in the guard. Appropriate scope.
- **SessionManager tests use real instances**: The delegation tests construct a real `SessionManager` with mocked dependencies rather than just testing mock wiring. This validates the actual integration path.
- **Tests verify internal state where appropriate**: Checking `removedWhileLoading` and `sessionLoadLocks` via bracket notation is justified here since these are the exact invariants the fix relies on.
- **Good documentation**: JSDoc on `remove()` and `unregisterSession()` clearly explain the three-part cleanup strategy and the caller context.

### Minor observations (P3)

1. **PR body says "13 unit tests" but test runner reports 12 passing tests** (`packages/daemon/tests/unit/room/session-registration.test.ts`). Cosmetic mismatch in the PR description.

2. **`clear()` does not reset `removedWhileLoading` or `sessionLoadLocks`** (`packages/daemon/src/lib/session/session-cache.ts:187-189`). Not a bug since `clear()` is only called during shutdown after session cleanup, but worth noting for future maintainers if `clear()` is ever used for cache reset scenarios.

3. **Comment at line 33 references "getAsync line 99"** — line numbers drift over time. Consider referencing the guard by name ("the set-race guard in getAsync") rather than by line number to keep the comment accurate across refactors.

No blocking issues found. The code is ready to merge.

PR #913, opened 2026-03-26
feat(space): add CompletionDetector for all-agents-done detection

lsm:
Implements `CompletionDetector` in `packages/daemon/src/lib/space/runtime/completion-detector.ts` for Task 4.1.

## Changes

- **`CompletionDetector`** class with `isComplete(workflowRunId, channels?, nodes?)` method
- Single direct query on `space_tasks WHERE workflow_run_id=?` via `listByWorkflowRun()`
- Terminal statuses: `completed`, `needs_attention`, `cancelled`, `rate_limited`, `usage_limited`
- Returns `false` when workflow has not started (no tasks)
- **Pending-but-blocked guard**: when channels + nodes are provided, checks if any outbound channel targets an unactivated node; returns `false` if so
- Nodes with no tasks are excluded unless a channel explicitly points to them
- Exports `TERMINAL_TASK_STATUSES` set for reuse by callers
- Constructor takes only `taskRepo` (no sessionGroupRepo)

## Tests

25 unit tests covering:
- All terminal statuses (completed, needs_attention, cancelled; rate_limited/usage_limited via set export since DB migration gap)
- All non-terminal statuses (draft, pending, in_progress, review)
- Single-node and multi-node workflows
- Multi-agent nodes
- Pending-but-blocked guard: wildcard channels, cross-node format, fan-out by node name, multi-target arrays, unresolvable targets
- Edge cases: no tasks, nodes with no tasks, empty channels/nodes

### DIFF
--- a/packages/daemon/src/lib/space/runtime/completion-detector.ts
+++ b/packages/daemon/src/lib/space/runtime/completion-detector.ts
@@ -1,0 +1,114 @@
+/**
+ * CompletionDetector
+ *
+ * Determines whether all agents in a workflow run have reached a terminal
+ * status. Uses a single direct query on space_tasks by workflow_run_id.
+ *
+ * Terminal statuses: completed, needs_attention, cancelled, rate_limited, usage_limited.
+ * Non-terminal statuses (block completion): draft, pending, in_progress, review.
+ */
+
+import type { SpaceTaskStatus, WorkflowChannel, WorkflowNode } from '@neokai/shared';
+import { resolveNodeAgents } from '@neokai/shared';
+import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
+
+/**
+ * Task statuses that represent a terminal (done) state for a workflow agent.
+ * A run is complete only when every task is in one of these statuses.
+ */
+export const TERMINAL_TASK_STATUSES = new Set<SpaceTaskStatus>([
+	'completed',
+	'needs_attention',
+	'cancelled',
+	'rate_limited',
+	'usage_limited',
+]);
+
+export class CompletionDetector {
+	constructor(private readonly taskRepo: SpaceTaskRepository) {}
+
+	/**
+	 * Returns true when the workflow run is complete.
+	 *
+	 * Completion conditions (all must hold):
+	 * 1. At least one task exists — workflow has started ("no tasks" → false).
+	 * 2. Every task has a terminal status — none in draft/pending/in_progress/review.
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped implementation that meets all acceptance criteria. The code is correct, the tests are thorough, and the design fits naturally with the existing codebase.

### What looks good

- **Single query approach**: `isComplete()` does exactly one `listByWorkflowRun()` call and operates on the result in memory. Clean and efficient.
- **Terminal status set**: Correctly covers all 5 terminal statuses (`completed`, `needs_attention`, `cancelled`, `rate_limited`, `usage_limited`) and the comment on lines 164-166 of the test file transparently documents the DB migration gap for `rate_limited`/`usage_limited`.
- **Pending-but-blocked guard**: Properly handles all `WorkflowChannel.to` address formats — wildcard (`*`), cross-node (`nodeId/agentName`), node name fan-out, and plain agent name. The `resolveTargetNodeId` method reuses `resolveNodeAgents` from `@neokai/shared` for agent name lookup, which is consistent with the rest of the codebase.
- **Edge cases covered**: No tasks returns false, unresolvable targets are skipped (not blocking), empty channels/nodes gracefully skip the guard, nodes with no tasks are excluded unless a channel targets them.
- **Test quality**: 25 tests cover every stated acceptance criterion with real SQLite (not mocks), including multi-target array channels (test 22) and the node-with-no-tasks exclusion (test 23). FK constraints are disabled to allow synthetic IDs, which is documented clearly.

### Minor observations (P3)

1. **Bidirectional channels**: The guard only inspects `channel.to` targets, not the `channel.from` side of bidirectional channels. For a `{ from: 'A', to: 'B', direction: 'bidirectional' }` channel, B can message A, making A also a "target." In practice this is fine because the `from` agent's node must already be activated to have initiated communication, but it could theoretically miss a degenerate case. Not blocking — just worth noting if bidirectional semantics evolve.

2. **`listByWorkflowRun` excludes archived tasks**: The detector inherits this filter from the repository (`status != 'archived'`). This is the correct behavior for completion detection, but the test labeled "13. archived tasks excluded" doesn't actually seed an archived task to verify the filter — it only seeds a completed task. The comment describes the intent, but the test doesn't exercise the archived-exclusion path. A minor test gap, not a correctness issue since the repository filter is tested elsewhere.

Overall: clean implementation, well-tested, meets the spec. Good to merge.

PR #914, opened 2026-03-26
feat(ui): agent completion state indicators on WorkflowNodeCard and WorkflowNode

lsm:
## Summary

- Add `AgentTaskState` type + `AgentStatusIcon` / `isNodeFullyCompleted` helpers to `WorkflowNodeCard`
- Per-agent status icons (spinner for in-progress, checkmark for done, X for failed/cancelled, dim dot for pending) in collapsed node header for both single-agent and multi-agent nodes
- Green border + green step badge on nodes where all agents have completed
- Completion summary text shown below agent names when available
- Add `tasksByNodeId` computed to `SpaceStore` for O(1) node→tasks lookup (reacts to `space.task.updated` WS events in real-time)
- Wire task states in `WorkflowEditor` (list view) and `VisualWorkflowEditor` (canvas view) — both show live indicators from `spaceStore.tasksByNodeId`
- 20 new tests; 5 test mocks updated to include `tasksByNodeId`

### DIFF
--- a/packages/web/src/components/space/WorkflowEditor.tsx
+++ b/packages/web/src/components/space/WorkflowEditor.tsx
@@ -192,6 +192,18 @@ export function WorkflowEditor({ workflow, onSave, onCancel }: WorkflowEditorPro
 	const agents = filterAgents(spaceStore.agents.value);
 	const tasksByNodeId = spaceStore.tasksByNodeId.value;
 
+	// Determine which workflow run to use for completion indicators.
+	// Prefer an active run; fall back to the most recently updated run.
+	// When no run exists yet (e.g. new workflow), relevantRunId is null.
+	const relevantRunId = (() => {
+		if (!workflow?.id) return null;
+		const runs = spaceStore.workflowRuns.value.filter((r) => r.workflowId === workflow.id);
+		if (!runs.length) return null;
+		const active = runs.find((r) => r.status === 'pending' || r.status === 'in_progress');
+		if (active) return active.id;
+		return [...runs].sort((a, b) => b.updatedAt - a.updatedAt)[0].id;
+	})();
+
 	// ---- Step operations ----
 
 	function addStep() {
@@ -527,8 +539,13 @@ export function WorkflowEditor({ workflow, onSave, onCancel }: WorkflowEditorPro
 
 					<div class="space-y-2">
 						{steps.map((step, i) => {
-							// Derive per-agent completion states from live task data for this node.
-							const nodeTasks = step.id ? (tasksByNodeId.get(step.id) ?? []) : [];
+							// Derive per-agent completion states from live task data for this node,
+							// scoped to the most relevant workflow run to avoid mixing state from
+							// past runs with the current one.
+							const allNodeTasks = step.id ? (tasksByNodeId.get(step.id) ?? []) : [];
+							const nodeTasks = relevantRunId
+								? allNodeTasks.filter((t) => t.workflowRunId === relevantRunId)
+								: allNodeTasks;
 							const nodeTaskStates: AgentTaskState[] = nodeTasks.map((t) => ({
 								agentName: t.agentName ?? null,
 								status: t.status,
--- a/packages/web/src/components/space/WorkflowEditor.tsx
+++ b/packages/web/src/components/space/WorkflowEditor.tsx
@@ -16,7 +16,7 @@ import type { SpaceWorkflow, SpaceAgent, WorkflowChannel } from '@neokai/shared'
 import { generateUUID } from '@neokai/shared';
 import { spaceStore } from '../../lib/space-store';
 import { WorkflowNodeCard } from './WorkflowNodeCard';
-import type { NodeDraft, ConditionDraft } from './WorkflowNodeCard';
+import type { NodeDraft, ConditionDraft, AgentTaskState } from './WorkflowNodeCard';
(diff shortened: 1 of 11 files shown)


lsm (commented):

Overall this is a well-structured implementation that cleanly adds per-agent completion state indicators to both the list-view (`WorkflowNodeCard`) and visual-editor (`WorkflowNode`) components. The code follows existing patterns, shares helpers via exports, and has good test coverage. A few issues to address:

### P1: `tasksByNodeId` does not filter by workflow run

**File:** `packages/web/src/lib/space-store.ts`, lines 127-136
**File:** `packages/web/src/components/space/WorkflowEditor.tsx`, line 531

The `tasksByNodeId` computed groups all space tasks by `workflowNodeId` without filtering by `workflowRunId`. Since `workflowNodeId` references the workflow *definition* node (not a run-specific instance), running the same workflow multiple times will accumulate tasks from all runs under the same key. This means:

- A node that completed in run #1 and is pending in run #2 would show *both* a checkmark and a spinner
- `isNodeFullyCompleted()` could return false even when the current run's agents are all done, because old tasks from previous runs are included

The fix should filter tasks to the most recent (or currently active) workflow run for the displayed workflow. For example, the `WorkflowEditor` could accept a `workflowRunId` filter, or `tasksByNodeId` could key on `(workflowRunId, workflowNodeId)`.

### P2: `tasksByNodeId` creates new array copies on every iteration

**File:** `packages/web/src/lib/space-store.ts`, line 132

```typescript
const existing = map.get(task.workflowNodeId) ?? [];
map.set(task.workflowNodeId, [...existing, task]);
```

Each iteration spreads the existing array into a new one, making this O(n^2) in the number of tasks per node. For small task counts this is fine, but it would be cleaner to push to a mutable array:

```typescript
let arr = map.get(task.workflowNodeId);
if (!arr) { arr = []; map.set(task.workflowNodeId, arr); }
arr.push(task);
```

### P3: Completion summary shows only the first agent's summary in multi-agent nodes

**File:** `packages/web/src/components/space/WorkflowNodeCard.tsx`, lines 820-824

```tsx
{nodeTaskStates.find((s) => s.completionSummary)?.completionSummary}
```

For multi-agent nodes, this will only ever display the first agent's summary. This is acceptable as a truncated preview, but worth noting as intentional behavior.

### Summary

The implementation is correct for the single-run case and covers all the requested features (per-agent status icons, green border, completion summary, visual editor support). The main concern is the missing workflow-run filter on `tasksByNodeId` which will cause incorrect state display when a workflow has been executed more than once.

**Recommendation: REQUEST_CHANGES** (due to P1)

lsm (commented):

Re-review after fixes for P1 (run filtering), P2 (O(n^2) spread), and P3 (intentional).

### Verification of P1 fix: Run filtering -- RESOLVED

The fix correctly addresses the stale-run contamination issue. Both `WorkflowEditor` (line 198-205) and `VisualWorkflowEditor` (line 148-155) now derive a `relevantRunId` by:

1. Filtering `spaceStore.workflowRuns` to the current workflow
2. Preferring an active run (`pending` or `in_progress`)
3. Falling back to the most recently updated run (`sort by updatedAt desc`)
4. Returning `null` when no runs exist (new workflow case)

The run ID is then used to filter tasks at the consumption site (e.g., `WorkflowEditor.tsx` line 546-548):
```tsx
const nodeTasks = relevantRunId
  ? allNodeTasks.filter((t) => t.workflowRunId === relevantRunId)
  : allNodeTasks;
```

This is clean -- `tasksByNodeId` stays a general-purpose index in the store, and run scoping happens in the view layer where the context (which workflow we are looking at) is known.

The `workflowRuns` signal is properly included in the test mocks for all 4 affected test files.

### Verification of P2 fix: O(n^2) spread -- RESOLVED

`tasksByNodeId` (space-store.ts lines 127-140) now uses mutable push:
```typescript
let arr = map.get(task.workflowNodeId);
if (!arr) { arr = []; map.set(task.workflowNodeId, arr); }
arr.push(task);
```

This is correct and O(n).

### Remaining observations

**P3 (nit): `tasksByRun` still uses the O(n^2) spread pattern** (`space-store.ts` lines 112-121). Same pattern as the old `tasksByNodeId` -- `[...existing, task]` on every iteration. Not introduced by this PR and not blocking, but worth noting for a follow-up since this PR already fixed the identical pattern in `tasksByNodeId`.

**P3 (nit): `sessionGroupsByTask` also uses the spread pattern** (`space-store.ts` line 148). Same situation -- pre-existing, not in scope.

### Test coverage

The 20 new tests (13 for `WorkflowNodeCard`, 7 for `WorkflowNode`) thoroughly cover:
- All status icon variants (spinner, check, fail, pending dot)
- Green border/badge for all-done state
- Completion summary rendering
- Multi-agent per-slot status icons
- Absence of icons when no task states provided

No test coverage gaps were found for the new functionality.

### Architecture assessment

The approach of keeping `tasksByNodeId` as a general-purpose computed and filtering by run ID in the view layer is sound. It avoids coupling the store to display concerns and allows the same index to be reused if other views need different run scoping logic.

The duplicated `relevantRunId` derivation in both `WorkflowEditor` and `VisualWorkflowEditor` is acceptable -- they are sibling components that don't share state, and extracting it would add unnecessary abstraction for two call sites.

**Recommendation: APPROVE** -- All flagged issues have been correctly addressed.

PR #920, opened 2026-03-26
fix(error-classifier): only pause on actual API errors, not SDK info messages

lsm:
## Summary

SDK `rate_limit_event` messages with `status: 'allowed'` or `'allowed_warning'` are informational (orange badge in UI) — the agent can continue working. Only `status: 'rejected'` is an actual limit hit requiring pause/fallback.

Previously, `classifyError()` could classify these informational events as `usage_limit`, incorrectly triggering pause and backoff.

### Changes

- **`error-classifier.ts`**: Add early guard for `rate_limit_event` JSON — returns `null` for non-rejected events; returns `usage_limit` with correct `resetsAt` (seconds→ms) for rejected events
- **`rate-limit-utils.ts`**: `parseRateLimitReset` skips non-rejected rate_limit_events immediately; extracts `resetsAt * 1000` for rejected events. `isRateLimitError` consistently checks status field for JSON messages
- **Tests**: New cases for `allowed`/`allowed_warning` (→ `null`) and `rejected` (→ `usage_limit`) in both test files

### DIFF
--- a/packages/daemon/src/lib/room/runtime/error-classifier.ts
+++ b/packages/daemon/src/lib/room/runtime/error-classifier.ts
@@ -4,19 +4,19 @@
  * Single point of truth for classifying API errors from agent output:
  * - terminal:    unrecoverable errors (4xx) → fail task immediately
  * - rate_limit:  HTTP 429 rate limits with parseable retry-after → pause with backoff
- * - usage_limit: daily/weekly usage cap limits (e.g. "You've hit your limit · resets 2pm")
- *                → immediately attempt fallback model, skip backoff
+ * - usage_limit: daily/weekly usage cap limits → immediately attempt fallback model, skip backoff
+ *                Detected via: SDK rate_limit_event (status:'rejected') OR
+ *                Anthropic usage-limit text "You've hit your limit · resets …"
  * - recoverable: transient errors (5xx) → bounce/retry
  *
- * Detection strategy: match only structured "API Error: NNN" messages from the
- * Claude Agent SDK. Free-form text patterns are intentionally avoided because
- * they cause false positives when workers discuss error handling in prose
- * (e.g. "implemented handling for invalid model errors").
+ * Detection strategy:
+ * 1. Structured "API Error: NNN" messages from the Claude Agent SDK (HTTP status code determines class)
+ * 2. SDK rate_limit_event JSON: only 'rejected' status is an actual error;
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

Clean, well-scoped fix. The implementation correctly addresses the root cause: SDK `rate_limit_event` messages with `status: 'allowed'` or `'allowed_warning'` were being misclassified as actual errors, causing unnecessary pauses. Now only `status: 'rejected'` triggers `usage_limit` classification.

### What I verified

1. **Correctness**: The three-function update (`classifyError`, `parseRateLimitReset`, `isRateLimitError`) is consistent — all three use the same pattern: fast `includes()` guard, JSON parse, check `status` field, only treat `rejected` as an error.

2. **No double-processing**: `classifyError` handles `rate_limit_event` at step 2 and returns early (either `usage_limit` or `null`), so step 3's call to `parseRateLimitReset` is never reached for these messages. If JSON parsing somehow fails in step 2, it falls through, and `parseRateLimitReset` would also fail the same way — no inconsistency.

3. **Seconds-to-milliseconds conversion**: `resetsAt * 1000` is applied consistently in both `classifyError` (line 146) and `parseRateLimitReset` (line 58). The test at `resetsAtSeconds = 1749600000` (June 2025 epoch seconds) correctly expects `resetsAtSeconds * 1000` as output.

4. **Edge case: rejected without resetsAt**: `classifyError` returns `usage_limit` with `resetsAt: undefined` — correct, since it's still an actual limit hit worth acting on. `parseRateLimitReset` returns `null` for the same case — also correct, since that function's contract is to return a timestamp or null.

5. **Callers in room-runtime.ts**: The runtime calls `classifyError` first, then uses `createRateLimitBackoff` only for `rate_limit` class (HTTP 429). For `usage_limit` class, the runtime triggers fallback logic. With this fix, `allowed`/`allowed_warning` events now return `null` from `classifyError`, so the runtime correctly ignores them — no code changes needed in callers.

6. **Test coverage**: All four meaningful states are tested in both test files: `allowed` → null, `allowed_warning` → null, `rejected` with `resetsAt` → usage_limit + timestamp, `rejected` without `resetsAt` → usage_limit without timestamp. All 59 tests pass.

### Minor observations (P3)

- The `as` type casts on `JSON.parse` results are duplicated across three functions with slightly different shapes. A shared type/parser could reduce this, but it's not worth refactoring for three small call sites.

No blocking issues found.

PR #921, opened 2026-03-26
feat(daemon): remind leader to re-submit review after human feedback

lsm:
When a human sends a message to a task in `review` status, the system transitions it to `in_progress`. Previously, the leader had no signal that it needed to re-submit for review after addressing the feedback.

## Changes

- In `task.sendHumanMessage` handler: when the task was in `review` status before the transition, prepend a context note to the injected message telling the leader to call `set_task_status` back to `review` after addressing the feedback
- Updated existing tests in `rpc-handlers/task-handlers.test.ts` that checked exact message content (now check for `toContain`)
- Added two new tests in `rpc/task-handlers.test.ts`: one verifying the reminder is prepended for review tasks, one verifying non-review tasks are unaffected

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
@@ -1073,9 +1073,11 @@ export function setupTaskHandlers(
 		// review tasks: transition to in_progress before routing the human message.
 		// The group is still active (sessions running), so no revival is needed — just
 		// update the status so the task reflects that work is ongoing again.
+		let wasInReview = false;
 		if (task.status === 'review') {
 			try {
 				await taskManager.setTaskStatus(taskId, 'in_progress');
+				wasInReview = true;
 			} catch (err) {
 				throw new Error(
 					`Failed to transition task ${taskId} from review to in_progress: ${String(err)}`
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped change that does exactly what was requested. The implementation correctly:

1. Tracks whether the task was in `review` status before the transition via `wasInReview` flag
2. Prepends a clear context note to the routed message only when the task was previously in review
3. Leaves non-review tasks completely unaffected

### What I checked

- **Correctness**: The `wasInReview` flag is set inside the `if (task.status === 'review')` block, only after the `setTaskStatus` call succeeds. If `setTaskStatus` throws, the flag stays `false` and the error propagates correctly. Good.

- **Message length validation**: The 10,000-character limit (`task-handlers.ts:998`) validates `params.message` before the reminder is prepended (`task-handlers.ts:1099-1102`). This means the final routed message can exceed 10,000 chars by the length of the reminder (~200 chars). This is acceptable since the limit is on user input, not the internal system message. No issue here.

- **Test coverage**: Three test files updated/added:
  - `rpc-handlers/task-handlers.test.ts` -- two existing tests updated from exact match to `toContain`, now also assert the reminder prefix is present
  - `rpc/task-handlers.test.ts` -- two new tests: positive case (review task gets reminder) and negative case (non-review task is unaffected)
  - The mock setup in `rpc/task-handlers.test.ts` correctly adds `setTaskStatus` to `mockTaskManager` and clears it in `beforeEach`

- **Edge cases**: The reminder is only prepended when `setTaskStatus` succeeds (flag set after the await). If the transition fails, the error is thrown before we reach the reminder logic. Correct.

No issues found. The change is minimal, well-tested, and fits the existing patterns.

lsm (commented):

### Recommendation: APPROVE

The implementation is correct and well-tested. The change is minimal and focused: a `wasInReview` flag captures whether the task was in `review` status before the transition, and a context note is prepended to the routed message referencing `submit_for_review` -- the correct MCP tool name.

**What works well:**

- The `wasInReview` flag is set only after `setTaskStatus` succeeds (line 1080 in `task-handlers.ts`), so the reminder is never prepended if the transition failed.
- `needs_attention` tasks (which transition through `review` as an intermediate status) use the revive path and return early at line 1070, so they correctly bypass the `wasInReview` logic -- no double handling.
- The length validation at line 998 checks `params.message.length` before the reminder is prepended, which is correct since the 10K limit is for user input, not the system-generated wrapper.
- Tests cover both the positive case (review task gets reminder) and the negative case (non-review task has no reminder).
- Existing tests in `rpc-handlers/task-handlers.test.ts` were properly updated from exact match to `toContain`.

**P3 issues:**

1. **Stale PR description** -- The PR body still says "telling the leader to call `set_task_status` back to `review`" but the code correctly references `submit_for_review`. The description should be updated to match the actual implementation.

PR #922, opened 2026-03-26
fix(goals): auto-compute nextRunAt in patchGoal when schedule is updated

lsm:
## Summary

- **Bug**: editing a recurring goal's schedule via the UI called `patchGoal()`, which saved the cron expression but left `nextRunAt` as null. The scheduler (`tickRecurringMissions` Phase 2) skips goals with `null` nextRunAt, so the goal never triggered.
- **Fix**: `patchGoal()` now mirrors `createGoal()` — when the effective mission type is `recurring` and `schedule` is being set without an explicit `nextRunAt`, it calls `getNextRunAt()` to compute and persist the value.
- **Tests**: added 6 unit tests to `goal-manager.test.ts` covering patchGoal schedule scenarios, and 2 integration tests to `recurring-missions.test.ts` verifying the full flow (patchGoal → scheduler → execution).

### DIFF
--- a/packages/daemon/src/lib/room/managers/goal-manager.ts
+++ b/packages/daemon/src/lib/room/managers/goal-manager.ts
@@ -337,13 +337,44 @@ export class GoalManager {
 	 * General-purpose patch for editable fields (title, description, missionType,
 	 * autonomyLevel, structuredMetrics, schedule, priority, and similar).
 	 * Does NOT change status or recalculate progress.
+	 *
+	 * Auto-computes nextRunAt when schedule is set for a recurring mission,
+	 * mirroring the same logic in createGoal(). This ensures that updating a
+	 * goal's cron expression via goal.update also persists nextRunAt so the
+	 * scheduler (tickRecurringMissions Phase 2) can fire the goal.
 	 */
 	async patchGoal(goalId: string, patch: UpdateGoalParams): Promise<RoomGoal> {
 		const goal = await this.getGoal(goalId);
 		if (!goal) {
 			throw new Error(`Goal not found: ${goalId}`);
 		}
-		const updatedGoal = this.goalRepo.updateGoal(goalId, patch);
+
+		// Auto-compute nextRunAt when schedule is being set for a recurring mission.
+		// The effective mission type is the patched type (if changing) or the existing type.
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix correctly addresses the root cause: `patchGoal()` now auto-computes `nextRunAt` when `schedule` is set on a recurring mission, mirroring the existing `createGoal()` logic. The implementation is clean, well-guarded, and properly tested.

### What looks good

- **Correct fix**: The new code block in `patchGoal()` (lines 352-375 of `goal-manager.ts`) properly mirrors `createGoal()` logic — validates cron expression, computes next run time, and merges it into the patch.
- **Effective guard conditions**: The three-way check (`effectiveMissionType === 'recurring' && patch.schedule != null && patch.nextRunAt === undefined`) correctly handles all relevant scenarios:
  - Existing recurring goal getting a new schedule -> computes nextRunAt
  - Goal being converted to recurring with a schedule -> computes nextRunAt
  - Caller explicitly providing nextRunAt -> respects it (no override)
  - Non-recurring goal getting a schedule -> no-op (correct)
  - Patching non-schedule fields on recurring goal -> preserves existing nextRunAt
- **Good error messages**: Both invalid cron and "no future run times" errors are descriptive.
- **Test coverage**: 6 unit tests cover the key scenarios (happy path, type conversion, explicit override, non-recurring no-op, invalid cron, non-schedule patch). 2 integration tests verify end-to-end scheduler firing.

### Minor observations (P3)

1. **Missing edge case: converting recurring -> one_shot does not clear nextRunAt**. If a user patches `missionType: 'one_shot'` on a recurring goal (without touching schedule), the stale `nextRunAt` persists. This is harmless because `tickRecurringMissions` filters by `missionType === 'recurring'`, but it is a data cleanliness gap. Not a bug — the scheduler correctly ignores it — but worth noting for a future cleanup.

2. **Unit test manually creates `goals` table** (lines 874-892 of `goal-manager.test.ts`) instead of using the schema from `createTables()`. If the schema evolves, this could drift. Other test files in this suite appear to do the same, so this is consistent with existing patterns, but fragile.

Both are pre-existing patterns, not regressions from this PR.

PR #925, opened 2026-03-26
feat(planner): verify WebSearch wiring and add search API domains to sandbox

lsm:
## Summary

- Verified both `Planner` and `plan-writer` agent definitions already include `WebSearch` in their `tools` arrays
- Added inline comments to `planner-agent.ts` explaining why `WebSearch` is needed for each agent
- Added missing search API domains to `DEFAULT_GLOBAL_SETTINGS.sandbox.network.allowedDomains` so the SDK's built-in `WebSearch` tool can reach upstream APIs (Brave Search, DuckDuckGo, Tavily, SerpAPI) when the sandbox is enabled

No behavioral changes — this is a documentation + sandbox config fix.

### DIFF
--- a/packages/daemon/src/lib/room/agents/planner-agent.ts
+++ b/packages/daemon/src/lib/room/agents/planner-agent.ts
@@ -240,6 +240,8 @@ export function buildPlanWriterAgentDef(planSlug: string): AgentDefinition {
 			'Grep',
 			'Glob',
 			'WebFetch',
+			// WebSearch: plan-writer needs internet access to research unfamiliar technologies,
+			// APIs, libraries, and best practices before producing a well-informed plan.
 			'WebSearch',
 		],
 		model: 'inherit',
@@ -590,6 +592,9 @@ export function createPlannerAgentInit(config: PlannerAgentConfig): AgentSession
 			'Grep',
 			'Glob',
 			'WebFetch',
+			// WebSearch: Planner needs internet access to research current best practices,
+			// library docs, and external context when editing/refining plan files directly
+			// during feedback rounds (Phase 1) or verifying task scope (Phase 2).
 			'WebSearch',
 		],
 		model: 'inherit',
--- a/packages/daemon/src/lib/room/agents/planner-agent.ts
+++ b/packages/daemon/src/lib/room/agents/planner-agent.ts
@@ -240,8 +240,6 @@ export function buildPlanWriterAgentDef(planSlug: string): AgentDefinition {
 			'Grep',
 			'Glob',
 			'WebFetch',
-			// WebSearch: plan-writer needs internet access to research unfamiliar technologies,
-			// APIs, libraries, and best practices before producing a well-informed plan.
 			'WebSearch',
 		],
 		model: 'inherit',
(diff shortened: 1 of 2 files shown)


lsm (commented):

### P1: Search API domains in sandbox allowedDomains are likely unnecessary

**File:** `packages/shared/src/types/settings.ts`, lines 242-249

The SDK type definitions at `packages/shared/src/sdk/sdk-tools.d.ts:78-81` show both `WebSearch` and `WebFetch` are tracked under `server_tool_use`:

```typescript
server_tool_use: {
  web_search_requests: number;
  web_fetch_requests: number;
} | null;
```

This indicates WebSearch (and WebFetch) are **server-side tools** -- the HTTP requests to search engines are made by the Anthropic API servers, not by the local SDK process running inside the sandbox. The sandbox `allowedDomains` controls outbound network access from the *local* sandbox environment (e.g., for `Bash` tool `curl` commands or similar).

Adding `api.search.brave.com`, `duckduckgo.com`, `api.tavily.com`, `serpapi.com`, etc. to `allowedDomains` has no practical effect because the local process never connects to these domains. The comment on line 242 ("required for SDK built-in WebSearch tool") is misleading -- it implies a dependency that does not exist.

**Suggestion:** Remove the search API domain additions. If you want to keep them as a "just in case" safety net, at minimum correct the comment to say something like "may be needed if future SDK versions execute searches locally" rather than stating they are "required."

### P2: Inline comments in planner-agent.ts add marginal value

**File:** `packages/daemon/src/lib/room/agents/planner-agent.ts`, lines 243-244 and 595-597

WebSearch is already included in virtually every agent definition in the codebase (coder, leader, reviewer, verifier, debugger, tester, vcs). None of those have explanatory comments for why WebSearch is in the tools list. Adding multi-line comments only on the planner agents creates an inconsistency and implies WebSearch is somehow special or unusual here, when it is standard practice.

The comments are also somewhat verbose for what amounts to "this agent can search the web." Consider removing them to stay consistent with the rest of the codebase, or if documentation is the goal, add a single brief comment.

### Summary

The core finding of the audit (WebSearch is correctly wired in both planner agent definitions, and query-options-builder does not restrict it) is accurate and valuable. However, the domain additions to `allowedDomains` appear to be based on an incorrect assumption about how the SDK's WebSearch tool works -- it is a server-side tool, not a client-side one, so sandbox network rules do not apply. The inline comments, while not harmful, are inconsistent with the rest of the codebase and slightly misleading about the significance of WebSearch in the planner context.

lsm (commented):

**Recommendation: APPROVE**

The previous review requested two changes:
1. Remove unnecessary search API domains from sandbox `allowedDomains`
2. Remove verbose inline comments inconsistent with codebase style

Both have been addressed. The second commit (`943621c`) fully reverts the first (`02b57bf`), resulting in zero net diff against `dev`. This is the correct outcome for a verification task -- `WebSearch` was already present in both the `Planner` (line 243) and `plan-writer` (line 593) tool arrays in `packages/daemon/src/lib/room/agents/planner-agent.ts`.

**Note:** The PR title and description still reference "add search API domains to sandbox" which no longer reflects the content. Consider updating the title to something like `docs(planner): verify WebSearch tool wiring` before merging, since this is effectively a no-op verification PR.

No issues found.

PR #929, opened 2026-03-26
plan: fix model switching bugs (remove silent auto-recovery, fix RPC routing, DB source of truth)

lsm:
## Summary

Plan for fixing three interrelated bugs in model switching for room agent sessions:

1. **Remove silent auto-recovery** -- `query-runner.ts` silently clears `sdkSessionId` and retries on SDK startup timeout without user notice. Replace with fail-loud behavior.
2. **Fix RPC routing** -- `trySwitchToFallbackModel` in `room-runtime.ts` calls `messageHub.request('session.model.get')` which sends the request over WebSocket instead of handling server-side. Add `getCurrentModel()` to `SessionFactory`.
3. **DB as source of truth** -- Read model/provider from DB session record instead of in-memory cache to avoid stale data.

## Plan

See `docs/plans/fix-model-switching-bugs-remove-silent-auto-recovery-fix-rpc.md` for the full implementation plan with 9 tasks across 3 milestones.

### DIFF
--- a/docs/plans/fix-model-switching-bugs-remove-silent-auto-recovery-fix-rpc.md
+++ b/docs/plans/fix-model-switching-bugs-remove-silent-auto-recovery-fix-rpc.md
@@ -4,49 +4,68 @@
 
 Fix three interrelated bugs in model switching for room agent sessions:
 
-1. **Remove silent auto-recovery** in `query-runner.ts` -- the `onStartupTimeoutAutoRecover` callback silently clears `sdkSessionId` and restarts the query, losing conversation context without user notice. Replace with fail-loud behavior: surface the error to the user and let them decide.
-2. **Fix `trySwitchToFallbackModel` RPC routing** in `room-runtime.ts:359` -- calls `messageHub.request('session.model.get', ...)` which sends the request through WebSocket transport to browser clients instead of handling it server-side. The handler is registered server-side via `onRequest()`, but `request()` always goes over the wire. Add `getCurrentModel()` to `SessionFactory` for direct server-side access.
-3. **DB as single source of truth** -- read current model/provider from the DB session record instead of relying on in-memory `AgentSession` cache. This avoids stale cache issues when model is switched externally.
+1. **Remove silent auto-recovery** in `query-runner.ts` — the `onStartupTimeoutAutoRecover` callback silently clears `sdkSessionId` and restarts the query, losing conversation context without user notice. Replace with fail-loud behavior: surface the error to the user and let them decide.
+2. **Fix `trySwitchToFallbackModel` RPC routing** in `room-runtime.ts` — calls `messageHub.request('session.model.get', ...)` which sends the request through WebSocket transport to browser clients instead of handling it server-side. The handler is registered server-side via `onRequest()`, but `request()` always goes over the wire. Add `getCurrentModel()` to `SessionFactory` for direct server-side access.
+3. **DB as single source of truth** — read current model/provider from the DB session record instead of relying on in-memory `AgentSession` cache. This avoids stale cache issues when model is switched externally.
 
 ## Approach
 
 The three bugs share common dependencies. The plan addresses them in dependency order:
 
-- **Milestone 1** (Bug 2): Fix RPC routing first because `trySwitchToFallbackModel` is a primary caller that needs to work correctly.
-- **Milestone 2** (Bug 3): Add DB-as-source-of-truth to `SessionFactory.getCurrentModel()` so the fix from Milestone 1 reads from the canonical source.
-- **Milestone 3** (Bug 1): Remove silent auto-recovery. This is the most impactful change because it modifies error handling behavior. It depends on Milestones 1 and 2 being in place so fallback model switching works correctly without silent retries.
+- **Milestone 1** (Bugs 2 + 3 combined): Add `getCurrentModel()` to `SessionFactory` that reads from DB as source of truth (not in-memory cache), then replace the `messageHub.request()` call in `trySwitchToFallbackModel`. Implementing DB-first from the start avoids the rework of writing a cache-first version and immediately rewriting it.
+- **Milestone 2** (Bug 1): Remove silent auto-recovery. This is the most impactful change because it modifies error handling behavior. It depends on Milestone 1 so fallback model switching works correctly without silent retries.
+
+### Key Design Decisions
+
+1. **SessionFactory is implemented as an object literal in `createSessionFactory()`** (room-runtime-service.ts), not in SessionManager. SessionManager does NOT implement SessionFactory.
+2. **`getCurrentModel()` reads from DB first, not in-memory cache.** The DB record is updated synchronously by `modelSwitchHandler.switchModel()` via `db.updateSession()`, making it the canonical source. SQLite single-row reads are sub-millisecond.
+3. **Keep `messageHub` in `RoomRuntimeConfig` but remove only the `session.model.get` call.** The field is used only at the assignment and the `session.model.get` call. Removing it from the config would require updating all 19 test files that pass `messageHub` to `createRuntimeTestContext()`. Option (b) — keep the field, remove only the usage — is lower-risk.
+4. **Model alias resolution in `getCurrentModel()`:** The `session.model.get` RPC handler resolves aliases via `resolveModelAlias()` before returning. The DB stores resolved model IDs after `switchModel()` (see model-switch-handler.ts where `resolvedModel = modelInfo?.id ?? newModel` is persisted). The implementer should verify whether the fallback map keys (`modelFallbackMap`) use aliases or resolved IDs, and match `getCurrentModel()` return values accordingly. If the map uses aliases, `getCurrentModel()` should return the raw DB value (which may be an alias on initial creation) without resolving. If the map uses resolved IDs, `getCurrentModel()` should resolve to match. Decision deferred to implementation with a verification step.
+5. **SessionFactory consumers:** `provision-global-agent.ts` and `session-notification-sink.ts` both import and store a `SessionFactory` reference but do not call `getCurrentModel()`. Adding the method to the interface is a compile-safe change — no updates needed in those files.
 
 ---
 
-## Milestone 1: Fix RPC Routing in `trySwitchToFallbackModel`
+## Milestone 1: Fix RPC Routing + DB as Source of Truth
 
-**Goal**: Replace `messageHub.request('session.model.get', ...)` in `room-runtime.ts` with direct server-side access via `SessionFactory.getCurrentModel()`.
+**Goal**: Add `SessionFactory.getCurrentModel()` that reads from DB, then replace the `messageHub.request('session.model.get')` call in `room-runtime.ts`.
 
-### Task 1.1: Add `getCurrentModel()` to `SessionFactory` interface and implementations
+### Task 1.1: Add `getCurrentModel()` to `SessionFactory` interface
 
-**Description**: Add a `getCurrentModel(sessionId: string)` method to the `SessionFactory` interface in `packages/daemon/src/lib/room/runtime/task-group-manager.ts`. This method reads model/provider from the session in-memory cache (or DB as fallback) and returns it without going through the MessageHub RPC layer.
+**Description**: Add a `getCurrentModel(sessionId: string)` method to the `SessionFactory` interface in `packages/daemon/src/lib/room/runtime/task-group-manager.ts`. Implement it in the object literal returned by `createSessionFactory()` in `packages/daemon/src/lib/room/runtime/room-runtime-service.ts`. The implementation reads from DB (source of truth), not the in-memory cache.
 
 **Files to modify**:
-- `packages/daemon/src/lib/room/runtime/task-group-manager.ts` -- Add `getCurrentModel(sessionId: string)` to `SessionFactory` interface:
+- `packages/daemon/src/lib/room/runtime/task-group-manager.ts` — Add to the `SessionFactory` interface:
   ```ts
   /**
    * Get the current model and provider for a session.
-   * Reads from in-memory cache (fast path) or DB (fallback).
-   * Returns null if session is not found.
+   * Reads from DB as the canonical source to avoid stale in-memory cache.
+   * Returns null if session is not found in DB.
    */
   getCurrentModel(sessionId: string): Promise<{ currentModel: string; currentProvider: string } | null>;
   ```
-- `packages/daemon/src/lib/session/session-manager.ts` -- Implement `getCurrentModel()` that delegates to `SessionCache.getAsync()` and reads `session.config.model` and `session.config.provider`. Falls back to `db.getSession(sessionId)` if not in cache.
+- `packages/daemon/src/lib/room/runtime/room-runtime-service.ts` — In the object literal returned by `createSessionFactory()` (near the other method implementations), add the `getCurrentModel` implementation:
+  ```ts
+  getCurrentModel: async (sessionId) => {
+    const session = ctx.db.getSession(sessionId);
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Verdict: REQUEST_CHANGES

The plan is well-structured and correctly identifies the three bugs. However, there are several factual inaccuracies and under-scoped tasks that need to be addressed before an agent can execute this plan cleanly.

---

### P1 -- Plan inaccuracies that will cause confusion or broken builds

**1. SessionManager does NOT implement SessionFactory (Task 1.1, Task 2.1)**

The plan says:

> Implement getCurrentModel() in SessionManager (which implements SessionFactory)

This is wrong. SessionManager does NOT implement SessionFactory. The actual implementation is the anonymous object literal returned by RoomRuntimeService.createSessionFactory() at packages/daemon/src/lib/room/runtime/room-runtime-service.ts:269. This object wraps agentSessions (a Map of AgentSession instances) and delegates to AgentSession methods.

getCurrentModel() must be implemented in createSessionFactory() in room-runtime-service.ts, not in session-manager.ts. The file session-manager.ts should not be listed as a file to modify for Task 1.1.

**2. Task 1.3 under-scopes the test file updates (P1)**

The plan states Task 1.3 only needs to update:
- room-runtime-test-helpers.ts
- room-runtime-model-fallback-map.test.ts

In reality, messageHub: makeMessageHub(...) is used in 3 test files (25 total occurrences):
- room-runtime-model-fallback-map.test.ts (8 occurrences)
- room-runtime-provider-availability.test.ts (10 occurrences)
- room-runtime-mirroring-usage-limit.test.ts (7 occurrences)

Additionally, 3 other test files pass custom messageHub mocks for session.model.get:
- room-runtime-terminal-errors.test.ts (lines 448, 529)
- room-runtime-rate-limit-persistence.test.ts (line 600)
- room-runtime-leader-terminal-errors.test.ts (lines 271, 342, 410)

That is 6 test files total that need updating (not 2). The plan should explicitly list all of them.

Furthermore, 4 test files use createMockSessionFactory() with satisfies SessionFactory:
- room-runtime-test-helpers.ts
- task-group-manager.test.ts
- runtime-recovery.test.ts
- room-runtime-service.test.ts

All of these will fail to compile if getCurrentModel() is added to the SessionFactory interface but not to the mock. The plan should mention all 4.

**3. Task 3.1 and Task 3.3 have a gap around STARTUP_MAX_RETRIES references (P1)**

The error messages at query-runner.ts:494-506 use STARTUP_MAX_RETRIES + 1 to describe the number of attempts. After removing auto-recovery, the message "after N attempt(s)" would be misleading. The plan acknowledges this needs updating in Task 3.1 subtask 4 but does not list it as a concrete subtask.

Additionally, errorManager.handleError() at line 522 receives startupMaxRetries: STARTUP_MAX_RETRIES in its context. If the STARTUP_MAX_RETRIES constant is removed (Task 3.1 subtask 3), this line will break. The plan should explicitly mention updating or removing this handleError call argument.

---

### P2 -- Missing considerations that should be addressed in the plan

**4. SessionFactory is also used in provision-global-agent.ts and session-notification-sink.ts (P2)**

Both files import SessionFactory from task-group-manager.ts and store a reference. Neither currently calls getCurrentModel(), so adding the new method to the interface is a compile-safe change. However, the plan should acknowledge these consumers exist so the implementer knows there are no surprise callers.

**5. Milestone 2 dependency on Milestone 1 creates unnecessary rework (P2)**

Task 2.1 lists its dependency as Task 1.1, but since Task 1.1 adds getCurrentModel() with a cache-first approach and Task 2.1 rewrites it to be DB-first, Task 1.1 DB fallback subtask becomes dead code. Consider either:
- Merging Task 1.1 subtask 2 (DB fallback) and Task 2.1 into a single task that is DB-first from the start, or
- Making the dependency chain explicit: Task 1.1 adds the interface only (no DB), Task 2.1 implements DB-first.

The current plan has the implementer write a cache-first version in Task 1.1, then immediately rewrite it to DB-first in Task 2.1. This is wasted effort.

**6. No clear decision on whether to remove messageHub from RoomRuntimeConfig (P2)**

Task 1.2 subtask 1 says to grep for this.messageHub usage. The grep confirms it is only used at line 359 for session.model.get. But if removed from RoomRuntimeConfig, all 11 test files that pass messageHub to createRuntimeTestContext would need updating -- not just the 6 that use session.model.get.

The plan should explicitly decide: (a) remove messageHub from RoomRuntimeConfig and update all 11 test files, or (b) keep the field but remove only the session.model.get call. Option (b) is lower-risk and should be recommended.

**7. Task 3.4 (online test) may be too brittle (P2)**

The test relies on setting NEOKAI_SDK_STARTUP_TIMEOUT_MS=100 to force a timeout. But the startup timeout is read once at module load (query-runner.ts:57). In the test runner, the daemon module may already be loaded by the time the test sets the env var, meaning the timeout would not take effect. The plan should specify that the test needs to either:
- Import the daemon server fresh in a way that re-evaluates the env var, or
- Use a different mechanism to trigger the timeout (e.g., inject a custom QueryRunner with an artificially short timeout).

---

### P3 -- Minor suggestions

**8. The plan mentions resolveModelAlias and getModelInfo from model-service for consistency with the RPC handler (P3)**

The current session.model.get handler at session-handlers.ts:399 does alias resolution. The plan should clarify whether getCurrentModel() on SessionFactory should also resolve aliases, or if the raw model ID from DB is sufficient. The caller (trySwitchToFallbackModel) uses the model as a lookup key into the fallback map (settings.modelFallbackMap), which likely uses raw model IDs, not resolved aliases. The plan should make a clear decision here to avoid confusion during implementation.

lsm (commented):

### Summary

This is a well-structured plan for fixing three interrelated bugs in model switching. The plan is thorough, has clear dependency ordering, and covers the essential test updates. However, I identified several issues that should be addressed before implementation.

### P1 Issues (Should Fix Before Implementation)

#### 1. Milestone 2 Task 2.1 has an ordering problem

The plan states Task 2.1 depends on Task 1.1, but Task 1.1 as written implements getCurrentModel() to read from in-memory cache first (lines 39-43 of the plan say "Use sessionCache.getAsync(sessionId) to get the AgentSession"). Then Task 2.1 changes it to read from DB. This means Task 1.2 (which depends on 1.1) and Task 1.3 would be built against a cache-first implementation that gets replaced in Milestone 2.

**Recommendation**: Combine Task 1.1 and Task 2.1 into a single task that implements getCurrentModel() with DB-as-source-of-truth from the start. This eliminates the intermediate implementation that gets thrown away.

#### 2. Missing: getCurrentModel should also resolve model aliases

Task 1.1 mentions "resolve model alias using resolveModelAlias and getModelInfo from model-service" as a subtask, but the acceptance criteria and actual code snippets do not include this logic. Looking at the existing session.model.get handler, it likely resolves aliases (e.g., 'sonnet' -> 'claude-sonnet-4-20250514'). If getCurrentModel() returns the raw alias while trySwitchToFallbackModel expects a resolved model ID, the fallback chain matching (line 390 in room-runtime.ts: `f.model === currentModel`) could fail.

**Recommendation**: Either explicitly include model alias resolution in getCurrentModel() with a test case, or document why it's not needed (e.g., the fallback chain is defined using aliases).

#### 3. Test coverage gap: No test for model fallback after external model switch

Task 2.2 covers DB-as-source-of-truth tests for getCurrentModel(), but there is no end-to-end test in the plan that verifies the full scenario: an external model switch happens, then trySwitchToFallbackModel correctly reads the new model from DB and picks the right fallback.

**Recommendation**: Add a test case in Task 1.3 or Task 2.2 that:
1. Sets up a session with model A
2. Simulates an external DB update to switch to model B
3. Triggers trySwitchToFallbackModel
4. Verifies the fallback chain is computed based on model B, not model A

#### 4. Line numbers in the plan are likely stale

The plan references specific line numbers (e.g., "lines 356-415" for auto-recovery logic in query-runner.ts, "lines 720-895" for test removal). These will likely be incorrect by implementation time. While not blocking, this can cause confusion.

**Recommendation**: Either remove specific line numbers and use descriptive references (e.g., "the canAutoRecover branch in the catch block"), or note that line numbers are approximate.

#### 5. Task 3.1 should clarify error message behavior

Task 3.1 says to simplify the error message by "removing 'after N attempt(s)' language since there are no longer retries." However, the existing error messages (visible at lines 494-507 in query-runner.ts) include helpful context like "Try sending your message again." The plan should clarify whether to keep or modify these hints.

**Recommendation**: Add explicit acceptance criteria: "Error message retains the 'Try sending your message again' hint and session-reset instructions."

### P3 Issues (Minor Nits)

#### 6. Potential inconsistency in SessionModelGetResult removal

Task 1.2 says to remove SessionModelGetResult if no longer referenced, but the interface is defined at line 109 in room-runtime.ts. Before removal, verify no other files import or use this type (unlikely, but worth checking).

#### 7. Minor: The plan mentions "messageHub optional field can be removed from RoomRuntimeConfig if unused elsewhere" as an acceptance criterion. The grep results show messageHub is only used at lines 309 and 359. Line 309 is the assignment, 359 is the request call. So removal is safe, but this should be verified in Task 1.2.

### Verdict

**Recommendation: REQUEST_CHANGES** (if this were not my own PR)

The P1 issues around implementation ordering and model alias resolution should be clarified before implementation begins. The plan is otherwise thorough and well-structured.

---

### Additional Notes

The plan correctly identifies that messageHub.request('session.model.get') sends over WebSocket, which is inefficient for server-side code. The fix using a direct SessionFactory method is the right approach.

The decision to remove silent auto-recovery entirely (rather than just logging) is a good UX improvement - users deserve to know when something fails rather than having it silently retry.

lsm (commented):

The updated plan addresses all P1 and P2 issues from the previous review. This is a thorough, well-structured plan that is ready for implementation. I have a few minor suggestions but no blocking issues.

### What was fixed well

1. **SessionFactory implementation location** -- Correctly identified that `SessionFactory` is an object literal in `createSessionFactory()` (room-runtime-service.ts), not implemented by `SessionManager`. Task 1.1 now targets the right file. The explicit "Do NOT modify session-manager.ts" subtask is helpful.

2. **Test scope expansion** -- Task 1.3 now covers all 11 affected test files (5 with `SessionFactory` mocks, 6 with `messageHub` mocks), categorized clearly. The subtask structure for each category is actionable.

3. **Milestones 1+2 merged** -- Implementing `getCurrentModel()` DB-first from the start eliminates the risk of shipping a cache-first version and having to rewrite it. The dependency table correctly shows Milestones 1 and 2 can run in parallel.

4. **STARTUP_MAX_RETRIES cleanup** -- Task 2.1 now explicitly lists `DEFAULT_STARTUP_RECOVERY_DELAY_MS` and `DEFAULT_STARTUP_MAX_RETRIES` in the removal list, includes subtask 6 for removing `startupMaxRetries` from `handleError()`, and acceptance criteria now state "No references to STARTUP_MAX_RETRIES or STARTUP_RECOVERY_DELAY_MS remain."

5. **Online test module-level env var issue** -- Task 2.4 now documents three concrete approaches (fresh import, test injection, dev proxy) with a clear "choose the best" instruction. This is a genuine gotcha that was missed in the original plan.

6. **messageHub retention decision** -- Keeping `messageHub` in `RoomRuntimeConfig` with documented justification (19 test files at risk) is the right call. The plan acknowledges the trade-off in the risks section.

### Remaining suggestions (informational, not blocking)

- **P3: Model alias resolution (Design Decision #4)** -- The plan defers the alias resolution decision to implementation, which is fine. However, I verified the type definition at `packages/shared/src/types/settings.ts:157` which states: "Keys are `\"provider/model\"` strings (e.g. `\"anthropic/claude-sonnet-4-20250514\"`)" -- these use resolved model IDs, not aliases. And `model-switch-handler.ts:128` stores `resolvedModel = modelInfo?.id ?? newModel` in the DB. So both the fallback map keys and DB values use resolved IDs. The implementer can likely skip the alias resolution in `getCurrentModel()` entirely. Consider adding this note to Task 1.1 subtask 3 to save investigation time.

- **P3: Task 1.3 file counts** -- The plan says "6 test files use `messageHub` mocks for `session.model.get`" and lists 3 files with `makeMessageHub`/`makeMessageHubMock` helpers plus 3 files with inline mocks. The inline mock files reference approximate line numbers (e.g., "approximately lines 448, 529"). The original plan was criticized for brittle line numbers; consider removing these approximate references entirely and just telling the implementer to grep for `session.model.get` in each file. Minor.

### Verdict

This is a solid, implementation-ready plan. The key architectural decisions are correct, the scope is well-defined, and the dependency ordering makes sense. No blocking issues remain.

**Recommendation: APPROVE**

lsm (commented):

### Summary

The plan has been updated to address all P1 and P2 issues from the previous review round. This is now a well-structured, implementation-ready plan.

---

### Resolution of Previous P1 Issues

**P1-1: SessionManager does NOT implement SessionFactory** -- RESOLVED

The plan now correctly identifies:
- Key Design Decision #1: "SessionFactory is implemented as an object literal in createSessionFactory() (room-runtime-service.ts), not in SessionManager. SessionManager does NOT implement SessionFactory."
- Task 1.1 subtask 4: "**Do NOT modify** session-manager.ts — it does not implement SessionFactory."

**P1-2: Task 1.3 under-scopes test file updates** -- RESOLVED

Task 1.3 now comprehensively lists:
- 5 files with SessionFactory mocks needing getCurrentModel
- 6 files with messageHub session.model.get mocks needing updates
- Specific file paths and line number references for each

**P1-3: STARTUP_MAX_RETRIES references gap** -- RESOLVED

Task 2.1 now explicitly includes:
- Subtask 6: "Remove startupMaxRetries: STARTUP_MAX_RETRIES from the errorManager.handleError() call"
- Acceptance criteria: "errorManager.handleError() is called without startupMaxRetries context"
- Subtask 5: Error message retention guidance with explicit "Keep recovery hints"

---

### Resolution of Previous P2 Issues

**P2-1: SessionFactory consumers** -- RESOLVED

Key Design Decision #5: "provision-global-agent.ts and session-notification-sink.ts both import and store a SessionFactory reference but do not call getCurrentModel(). Adding the method to the interface is a compile-safe change."

**P2-2: Milestone rework** -- RESOLVED

Key Design Decision #2 and Approach section: "Implementing DB-first from the start avoids the rework of writing a cache-first version and immediately rewriting it." Task 1.1 now specifies DB-first implementation directly.

**P2-3: messageHub removal decision** -- RESOLVED

Key Design Decision #3: "Keep messageHub in RoomRuntimeConfig but remove only the session.model.get call... Option (b) — keep the field, remove only the usage — is lower-risk."

**P2-4: Online test brittleness** -- RESOLVED

Task 2.4 now includes an "Important implementation note" with three concrete approaches for triggering timeout conditions, acknowledging the module-level env var loading issue.

**P2-5: Model alias resolution decision** -- RESOLVED

Key Design Decision #4 provides clear guidance on the alias resolution complexity, and Task 1.1 subtask 3 requires verification and appropriate handling (not just documentation).

---

### Remaining Considerations (P3)

The plan is ready for implementation. One minor observation for the implementer:

The alias resolution issue (Key Design Decision #4, Risk #1) is identified as a pre-existing issue. The plan correctly defers the decision to implementation with a verification step. The implementer should be aware that:

1. Tests use `makeMessageHub('sonnet', 'anthropic')` returning alias `'sonnet'`
2. Production sessions may start with resolved IDs like `'claude-sonnet-4-5-20250929'`
3. `modelFallbackMap` keys typically use aliases like `'anthropic/sonnet'`

The verification step in Task 1.1 subtask 3 should catch any mismatch during implementation.

---

### Verdict

**Recommendation: APPROVE**

All previous blocking issues have been addressed. The plan is thorough, has clear dependency ordering, comprehensive test coverage planning, and appropriate risk documentation.

PR #931, opened 2026-03-26
fix(daemon): skip error classification for tool result messages

lsm:
## Summary

- The `setupMirroring` handler was running `classifyError()` on ALL `sdk.message` events, including user messages that are tool results
- Tool results can contain arbitrary text from file reads, including test fixtures like `"You've hit your limit · resets 11pm (America/New_York)"` which would falsely trigger usage_limit detection
- Now skips error classification for user messages (`type: 'user'`) since tool results are data returned from tool calls, not actual model errors

## Test plan

- [x] Added 3 unit tests to verify user messages with usage_limit text don't trigger false positives
- [x] All existing tests pass (21 tests in room-runtime-mirroring-usage-limit.test.ts, 37 tests in error-classifier.test.ts)

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -63,6 +63,7 @@ import {
 } from './message-routing';
 import { createRateLimitBackoff } from './rate-limit-utils';
 import { classifyError } from './error-classifier';
+import { isSDKUserMessage } from '@neokai/shared/sdk/type-guards';
 import { Logger } from '../../logger';
 import {
 	runWorkerExitGate,
@@ -2346,6 +2347,11 @@ export class RoomRuntime {
 					const uuid = 'uuid' in event.message ? (event.message.uuid as string) : null;
 					if (uuid && mirroredUuids.has(uuid)) return;
 					if (uuid) mirroredUuids.add(uuid);
+
+					// Skip tool results — they contain arbitrary file content that can match error
+					// patterns (e.g. test fixtures like "You've hit your limit · resets 11pm") but
+					// are not actual errors from the model
+					if (isSDKUserMessage(event.message)) return;
 
 					// Check for rate limit errors in real-time for both Worker and Leader sessions
 					const messageContent = JSON.stringify(event.message);
--- a/packages/daemon/tests/unit/room/room-runtime-mirroring-usage-limit.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-mirroring-usage-limit.test.ts
@@ -354,6 +354,99 @@ describe('setupMirroring - usage_limit real-time detection', () => {
 		});
 	});
 
+	describe('tool result messages (user type) should not trigger false positives', () => {
+		it('does not detect usage_limit in user message (tool result content)', async () => {
+			ctx = createRuntimeTestContext({
+				getGlobalSettings: () => ({}) as GlobalSettings,
+			});
+
+			const { group } = await spawnGroup();
+
+			// Simulate a tool result containing the usage_limit text as fixture data
+			// This should NOT trigger the usage_limit detection because user messages
+			// are tool results, not actual model error output
+			ctx.hub.fire('sdk.message', {
+				sessionId: group.workerSessionId,
+				message: {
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

---

### Summary

This is a clean, targeted fix for a real false-positive bug. The implementation is minimal (4 lines of production code) and well-tested (3 new tests). All 21 tests in the test file pass, the 37 error-classifier tests pass, and TypeScript type checking passes.

### What was done correctly

1. **Root cause understanding is accurate**: Tool results are `type: 'user'` messages that carry arbitrary file content. The error classifier was running on ALL `sdk.message` events, so reading a file containing error-pattern text (e.g., test fixtures) would falsely trigger `usage_limit` detection and pause tasks.

2. **Correct placement of the guard**: The `isSDKUserMessage()` check is at line 2354 in `room-runtime.ts`, positioned right after the UUID deduplication guard and before the `classifyError()` call. This means all downstream error-handling logic (rate_limit, usage_limit) is protected.

3. **Reuse of existing, well-tested type guard**: `isSDKUserMessage()` in `packages/shared/src/sdk/type-guards.ts` is an established function with existing unit tests (lines 67-105 in `type-guards.test.ts`). It checks `msg.type === 'user' && !isReplay`. Reusing it avoids duplicating logic.

4. **Test quality**: The three new tests cover the important dimensions:
   - Worker session with tool result containing usage_limit text (no fallback model)
   - Leader session with tool result containing usage_limit text (no fallback model)
   - Worker session with tool result and fallback model configured (verifies no `switchModel` call is made)

1. **User replay messages are not filtered**: `isSDKUserMessage()` returns `false` for messages with `isReplay: true`. If a replayed user message contained usage_limit text, it could still trigger a false positive. This is a very low-risk edge case because replays represent the original user prompt (short instructions), not tool results with arbitrary file content. Not worth fixing now, but worth keeping in mind if future false positives surface.

2. **The guard comment could be more precise**: The comment says "Skip tool results" but the guard actually skips ALL non-replay user messages (including direct user prompts). This is fine in practice because: (a) user prompts are not model output, so classifying them as errors is already wrong in principle, and (b) user prompts rarely contain the specific error patterns the classifier looks for. Still, the comment slightly oversells what the guard does.

lsm (commented):

### Summary

This is a clean, targeted fix for the false positive usage_limit detection bug. The implementation correctly addresses the root cause by excluding user messages (tool results) from error classification in the `setupMirroring()` handler.

### What was done well

1. **Correct placement of guard**: The `isSDKUserMessage(event.message)` check is placed BEFORE `classifyError()` is called, preventing unnecessary work and ensuring the fix is efficient.

2. **Proper type guard usage**: The existing `isSDKUserMessage` type guard correctly identifies user messages containing tool results, which are the source of false positives.

3. **Comprehensive tests**: Three new tests cover the key scenarios:
   - Worker session tool result with usage_limit text
   - Leader session tool result with usage_limit text  
   - Fallback model switch NOT triggered for user messages

4. **All tests pass**: 21 tests in the test file pass successfully.

### Code Review

**File: `packages/daemon/src/lib/room/runtime/room-runtime.ts` (lines 2351-2354)**
```typescript
// Skip tool results — they contain arbitrary file content that can match error
// patterns (e.g. test fixtures like "You've hit your limit · resets 11pm") but
// are not actual errors from the model
if (isSDKUserMessage(event.message)) return;
```

The comment clearly explains the rationale. The guard is minimal and targeted.

**File: `packages/daemon/tests/unit/room/room-runtime-mirroring-usage-limit.test.ts`**

The new test block "tool result messages (user type) should not trigger false positives" (lines 357-448) provides good coverage:
- Tests both worker and leader sessions
- Tests that no fallback switch is attempted
- Uses realistic message structures with `type: 'user'` and `tool_result` content blocks

### Minor Note (P3)

The fix excludes `isSDKUserMessage()` (non-replay user messages) but not `isSDKUserMessageReplay()`. Replay messages are synthetic SDK-generated messages (e.g., `/context` command output) and are extremely unlikely to contain usage_limit patterns. This is not a concern in practice, just noting for completeness.

### Verdict

The fix is correct, well-tested, and follows the existing code patterns. Approving.

PR #934, opened 2026-03-26
feat(daemon): SkillRepository (SQLite) and SkillsManager

lsm:
## Summary

- `SkillRepository` — SQLite CRUD for `skills` table (migration 57 + schema/index.ts), using the shared DB instance. Implements `findAll`, `get`, `findEnabled`, `insert`, `update`, `setEnabled`, `setValidationStatus`, `delete`. All writes call `reactiveDb.notifyChange('skills')`.
- `SkillsManager` — service layer wrapping the repo. Enforces input validation before every insert/update: plugin `pluginPath` must be absolute and traversal-free; `mcp_server` `appMcpServerId` must reference an existing `app_mcp_servers` row; `builtin` `commandName` must be non-empty. `removeSkill` returns `false` for built-in skills.
- `app.ts` — instantiates `SkillsManager(db.skills, db.appMcpServers)` and calls `initializeBuiltins()` at startup.
- 32 unit tests: CRUD, built-in deletion protection, persistence across DB reopen, `getEnabledSkills` filtering, and all validation boundary cases.

Closes task 2.2 of the Skills Registry milestone.

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -26,6 +26,7 @@ import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
 import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
 import { AppMcpLifecycleManager, seedDefaultMcpEntries } from './lib/mcp';
 import { FileIndex } from './lib/file-index';
+import { SkillsManager } from './lib/skills-manager';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -75,6 +76,8 @@ export interface DaemonAppContext {
 	jobProcessor: JobQueueProcessor;
 	/** Application-level MCP lifecycle manager — converts registry entries to SDK configs */
 	appMcpManager: AppMcpLifecycleManager;
+	/** Application-level Skills manager — registry CRUD and validation */
+	skillsManager: SkillsManager;
 	/** Workspace file index for fast fuzzy file/folder search */
 	fileIndex: FileIndex;
 	/**
@@ -271,6 +274,10 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 
 	// Seed default MCP entries (idempotent — skips entries that already exist)
 	seedDefaultMcpEntries(db);
+
+	// Instantiate Skills manager and initialize built-in skills
+	const skillsManager = new SkillsManager(db.skills, db.appMcpServers);
+	skillsManager.initializeBuiltins();
 
 	// Initialize workspace file index (non-blocking — init runs in the background)
 	const fileIndex = new FileIndex(config.workspaceRoot);
@@ -552,6 +559,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		jobQueue,
 		jobProcessor,
 		appMcpManager,
+		skillsManager,
 		fileIndex,
 		cleanup,
 	};
--- a/packages/daemon/src/lib/skills-manager.ts
+++ b/packages/daemon/src/lib/skills-manager.ts
@@ -32,7 +32,14 @@ export class SkillsManager {
 	}
 
 	addSkill(params: CreateSkillParams): AppSkill {
+		// Validate sourceType/config consistency and security-sensitive fields
 		this.validateSkillConfig(params.sourceType, params.config);
+
+		// Enforce name uniqueness with a user-friendly error
+		const existing = this.repo.getByName(params.name);
+		if (existing) {
+			throw new Error(`A skill named "${params.name}" already exists`);
+		}
 
 		const skill: AppSkill = {
 			id: generateUUID(),
@@ -48,7 +55,11 @@ export class SkillsManager {
 		};
 
 		this.repo.insert(skill);
-		return this.repo.get(skill.id)!;
+		const inserted = this.repo.get(skill.id);
+		if (!inserted) {
+			throw new Error(`Failed to insert skill "${params.name}"`);
+		}
+		return inserted;
 	}
 
 	updateSkill(id: string, params: UpdateSkillParams): AppSkill {
@@ -74,8 +85,17 @@ export class SkillsManager {
 		return this.repo.get(id)!;
 	}
 
-	setSkillValidationStatus(id: string, status: SkillValidationStatus): void {
+	/**
+	 * Set the validation status for a skill (called by the async validation job).
+	 * Throws if the skill does not exist so job failures are surfaced, not silenced.
+	 */
+	setSkillValidationStatus(id: string, status: SkillValidationStatus): AppSkill {
+		const existing = this.repo.get(id);
+		if (!existing) {
+			throw new Error(`Skill not found: ${id}`);
+		}
 		this.repo.setValidationStatus(id, status);
+		return this.repo.get(id)!;
 	}
 
 	/**
@@ -110,8 +130,17 @@ export class SkillsManager {
 	/**
 	 * Validate source-type-specific config fields for security.
 	 * Throws a descriptive Error on validation failure.
+	 *
+	 * Checks performed:
+	 * 1. sourceType must match config.type (prevents mismatched payloads)
+	 * 2. Source-type-specific field constraints
 	 */
 	private validateSkillConfig(sourceType: SkillSourceType, config: AppSkillConfig): void {
+		// Explicit sourceType/config.type consistency check
+		if (sourceType !== config.type) {
+			throw new Error(`sourceType "${sourceType}" must match config.type "${config.type}"`);
+		}
+
 		if (config.type === 'plugin') {
 			const { pluginPath } = config;
 			if (!pluginPath || pluginPath.trim() === '') {
@@ -120,7 +149,8 @@ export class SkillsManager {
 			if (!pluginPath.startsWith('/')) {
 				throw new Error('plugin skill: pluginPath must be an absolute path (starts with /)');
 			}
-			if (pluginPath.includes('../')) {
+			// Reject any path that contains '..' as a segment (handles /a/../b and /a/b/..)
+			if (pluginPath.split('/').some((seg) => seg === '..')) {
 				throw new Error('plugin skill: pluginPath must not contain path traversal sequences (../)');
 			}
 		} else if (config.type === 'mcp_server') {
@@ -140,9 +170,9 @@ export class SkillsManager {
 				throw new Error('builtin skill: commandName must not be empty');
 			}
 		} else {
-			// Exhaustive check — sourceType drives config.type in valid usage
-			const _: never = config;
-			throw new Error(`Unknown skill config type: ${sourceType}`);
+			// Exhaustive type guard
+			const _exhaustive: never = config;
+			throw new Error(`Unknown skill config type: ${(_exhaustive as AppSkillConfig).type}`);
 		}
 	}
 }
--- a/packages/daemon/src/lib/skills-manager.ts
+++ b/packages/daemon/src/lib/skills-manager.ts
@@ -1,0 +1,148 @@
+/**
+ * SkillsManager
+ *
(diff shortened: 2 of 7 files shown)


lsm (commented):

---

### Verdict: APPROVE

Overall this is clean, well-structured work that follows existing patterns in the codebase. The repository and service layer are well-separated, validation is thoughtfully applied at the service boundary, and the test coverage is solid. A few items worth noting but nothing blocking.

---

### P1: `setSkillValidationStatus` does not guard against non-existent IDs

**File:** `packages/daemon/src/lib/skills-manager.ts`, line 77-79

`setSkillValidationStatus` passes directly through to `repo.setValidationStatus()` without first checking if the skill exists. Contrast this with `setSkillEnabled` (lines 68-75) which does check. The repo-level `setValidationStatus` will execute a no-op UPDATE (affecting 0 rows) but still fire `notifyChange('skills')`, which would trigger unnecessary frontend re-subscriptions. More importantly, callers get no signal that nothing actually changed, which could mask bugs in the async validation job that calls this method.

```ts
// Current — no existence check, no return value
setSkillValidationStatus(id: string, status: SkillValidationStatus): void {
    this.repo.setValidationStatus(id, status);
}
```

Same pattern as `setSkillEnabled` would be more consistent: check existence, throw if not found, return the updated `AppSkill`.

---

### P1: `sourceType` vs `config.type` mismatch is not validated

**File:** `packages/daemon/src/lib/skills-manager.ts`, line 114

`validateSkillConfig` takes `sourceType` as a parameter but never actually checks that `sourceType` matches `config.type`. The comment at line 143 says "sourceType drives config.type in valid usage" but this is only an assumption — nothing enforces it. A caller could pass `sourceType: 'builtin'` with `config: { type: 'plugin', pluginPath: '/some/path' }` and the `builtin` branch would check for `commandName`, which wouldn't exist on a PluginSkillConfig, so it would throw — but the error message would be misleading ("commandName must not be empty" instead of "config.type does not match sourceType"). 

A defensive assertion like `if (sourceType !== config.type) throw ...` at the top of `validateSkillConfig` would make this contract explicit and produce clearer error messages.

---

### P2: Path traversal check misses edge cases

**File:** `packages/daemon/src/lib/skills-manager.ts`, lines 123-124

The traversal check only looks for the literal string `'../'`. This misses:
- `..\/` (backslash — not typical on macOS but possible in string inputs)
- `/foo/..` (traversal at the end of the path, no trailing slash)
- Paths like `/../../etc/passwd` starting with `..`

The simplest robust approach would be to normalize the path first (e.g., `path.posix.normalize`) and then check the result doesn't escape the expected root, or at minimum check for `..` as a path segment rather than just the substring `../`.

Not blocking since this is defense-in-depth — the plugin loader downstream should also validate paths — but it's worth hardening.

---

### P2: `SkillRepository.update()` silently ignores `name` and `sourceType` fields

**File:** `packages/daemon/src/storage/repositories/skill-repository.ts`, lines 143-173

The `update` method accepts `Partial<AppSkill>` but only explicitly handles `displayName`, `description`, `enabled`, `config`, and `validationStatus`. If someone passes `{ name: 'new-name' }`, it silently ignores it. Since `SkillRepository` is a `@public`-tagged export, a future consumer could call `repo.update(id, { name: 'x' })` expecting it to work and get confused. The `UpdateSkillParams` type in shared restricts this at the manager layer, but the repo's signature is misleading.

---

### P3: `ensureTable()` is dead code

**File:** `packages/daemon/src/storage/repositories/skill-repository.ts`, lines 68-83

The `ensureTable()` method is defined on `SkillRepository` but never called anywhere. Table creation is handled by both `createTables()` in `schema/index.ts` and `runMigration57()`. This method should either be removed (since no other repository uses this pattern) or called if there's an intentional standalone usage planned.

---

### P3: Persistence test doesn't clean up temp file

**File:** `packages/daemon/tests/unit/skills-manager.test.ts`, line 241

The persistence test creates a temp DB at `/tmp/skills-test-${Date.now()}.db` but never unlinks it. In CI this is fine (ephemeral environment) but in local development these files accumulate. An `afterAll` or inline `unlinkSync` would be tidy.

lsm (commented):

## Summary

This PR implements the SkillRepository (SQLite persistence) and SkillsManager service layer for the Skills Registry milestone. The implementation is solid overall, with good test coverage (32 tests) and proper integration with the existing codebase patterns.

## Strengths

- Clean separation between repository (data access) and manager (business logic/validation)
- Proper reactive database integration with `notifyChange('skills')` on all writes
- Security-conscious validation for plugin paths (absolute path, no traversal)
- Proper FK validation for mcp_server references
- Built-in skill deletion protection
- Follows existing patterns from AppMcpServerRepository

## Issues Found

### P1: Missing name uniqueness validation

**File:** `packages/daemon/src/lib/skills-manager.ts` (line 34-51)

Unlike `AppMcpServerRepository` which has an `isNameTaken()` check, `SkillsManager.addSkill()` doesn't validate name uniqueness before inserting. While the SQLite UNIQUE constraint will reject duplicates, this results in an unhandled raw SQLite error rather than a user-friendly message like `"A skill named \"xyz\" already exists"`.

Additionally, line 51 uses `this.repo.get(skill.id)!` with a non-null assertion. If insert fails (e.g., due to constraint violation), the subsequent get returns null and the assertion could cause unexpected behavior.

**Recommendation:** Add a `findByName` method to `SkillRepository` and check for uniqueness in `SkillsManager.addSkill()` before calling insert. Wrap the insert in a try/catch or verify the get returns non-null.

---

### P1: Missing sourceType/config.type consistency validation

**File:** `packages/daemon/src/lib/skills-manager.ts` (line 114)

The `validateSkillConfig` method receives `sourceType` but only uses `config.type` for the switch logic. It doesn't validate that these two values are consistent. A caller could pass:
```ts
{ sourceType: 'builtin', config: { type: 'plugin', pluginPath: '/path' } }
```
and the validation would pass (checking the plugin path) while storing mismatched data.

**Recommendation:** Add a validation at the start of `validateSkillConfig`:
```ts
if (sourceType !== config.type) {
  throw new Error(`sourceType "${sourceType}" must match config.type "${config.type}"`);
}
```

---

### P1: Inconsistent error handling for setSkillValidationStatus

**File:** `packages/daemon/src/lib/skills-manager.ts` (line 77-79)

`setSkillEnabled` (line 68-75) validates the skill exists and throws `"Skill not found"` if not. However, `setSkillValidationStatus` doesn't perform this check. For consistency and to catch bugs early, this method should also validate the skill exists.

**Recommendation:** Add the same existence check as `setSkillEnabled`.

---

### P2: Comment syntax error in initializeBuiltins

**File:** `packages/daemon/src/lib/skills-manager.ts` (line 102)

The comment has a typo: `/ Implementors:` should be `// Implementors:`. This is visible in the source code.

---

### P2: Temp file not cleaned up in test

**File:** `packages/daemon/tests/unit/skills-manager.test.ts` (line 241)

The persistence test creates a temp file at `/tmp/skills-test-${Date.now()}.db` but doesn't clean it up after the test completes. Over time this could accumulate temp files.

**Recommendation:** Add cleanup in an `afterEach` or at the end of the test:
```ts
import { unlinkSync } from 'fs';
// after db2.close():
unlinkSync(tmpPath);
```

---

### P2: No test for duplicate name handling

**File:** `packages/daemon/tests/unit/skills-manager.test.ts`

There's no test verifying the behavior when adding a skill with a duplicate name. This gap in test coverage could hide the P1 issue above.

**Recommendation:** Add a test case for duplicate name insertion.

---

The implementation is well-structured and follows project patterns, but the P1 issues (missing name uniqueness validation, missing sourceType/config consistency check, and inconsistent error handling) should be addressed before merge. These are security/reliability concerns that could cause confusing errors for users or data integrity issues.

lsm (commented):

**Recommendation: APPROVE**

All 38 tests pass, typecheck is clean, lint is clean. The previous review feedback has been addressed thoroughly. Here is a summary of remaining observations (none blocking).

---

### P2 (1 issue -- non-blocking, deferred)

**1. Inconsistent conditional `notifyChange` across repository write methods**

`packages/daemon/src/storage/repositories/skill-repository.ts` lines 165-168 vs 174-183 vs 188-195.

- `setEnabled` (line 167): unconditionally calls `notifyChange`, even if the UPDATE matched zero rows.
- `update` (line 159): unconditionally calls `notifyChange`, even if the UPDATE matched zero rows.
- `setValidationStatus` (line 180): correctly gated on `result.changes > 0`.
- `delete` (line 192): correctly gated on `result.changes > 0`.

In practice this is harmless because `setEnabled` is always called from `SkillsManager.setSkillEnabled` which guards for existence, and `update` is always called from `SkillsManager.updateSkill` which also guards. But if the repository is ever called directly (e.g., from a test or a future code path), spurious change notifications would fire. Consider gating all four methods on `result.changes > 0` for consistency. This is non-blocking because the manager layer already prevents the zero-row case.

---

### P3 (2 issues -- nits)

**1. Missing CHECK constraints on `skills.source_type` and `skills.validation_status`**

Both `createTables()` in `packages/daemon/src/storage/schema/index.ts` (line 372) and `runMigration57()` in `packages/daemon/src/storage/schema/migrations.ts` (line 3720) define `source_type TEXT NOT NULL` and `validation_status TEXT NOT NULL` without CHECK constraints. Other tables in the schema (e.g., `auth_config.auth_method`, `sessions.status`, `app_mcp_servers.source_type`) use CHECK constraints for defense-in-depth. Adding `CHECK(source_type IN ('builtin', 'plugin', 'mcp_server'))` and `CHECK(validation_status IN ('pending', 'valid', 'invalid', 'unknown'))` would align the skills table with the existing convention and prevent corrupt values from being written at the DB level.

**2. Module-level doc comment says "Each write method calls reactiveDb.notifyChange"**

`packages/daemon/src/storage/repositories/skill-repository.ts` line 5. After the review fix, `setValidationStatus` and `delete` only notify conditionally, so the blanket statement is slightly inaccurate. Minor doc nit.

---

### Summary

The implementation is solid: well-typed, well-tested, properly integrated into the app bootstrap, and follows established repository/manager patterns. The previous P1/P2/P3 feedback was all addressed correctly. The one remaining P2 is a minor internal consistency issue that cannot manifest in practice today. The two P3s are defensive DB hygiene and a doc tweak.

lsm (commented):

### Summary

The implementation correctly addresses all the feedback from the previous review. All 38 tests pass, typecheck passes, and lint/knip are clean.

### Verified Fixes

**P1 fixes (all addressed correctly):**
- `setSkillValidationStatus` now checks existence and throws "Skill not found" at `/packages/daemon/src/lib/skills-manager.ts:93-96`, returns updated AppSkill at line 98
- `validateSkillConfig` asserts sourceType === config.type up-front at line 140-142
- `addSkill` calls `repo.getByName()` before insert and throws for duplicate names at lines 39-42

**P2 fixes (all addressed correctly):**
- Traversal check uses `split('/').some(seg => seg === '..')` at line 153, correctly catching edge cases like `/a/b/..`
- Temp DB file cleaned up with `unlinkSync(tmpPath)` at line 294
- New tests cover: duplicate name (lines 437-458), sourceType/config.type mismatch (lines 501-514), traversal-at-end (lines 573-585)

**P3 fixes (all addressed):**
- No dead `ensureTable()` method present
- `SkillRepository.update()` param correctly typed as `UpdateSkillParams` at line 134
- `getByName()` added to SkillRepository at lines 88-93

### Minor Observation (P3)

One minor gap in test coverage: while `addSkill` has a test for sourceType/config.type mismatch (lines 501-514), `updateSkill` does not have a dedicated test for this case. The code path is correct (line 72 calls `validateSkillConfig` with `existing.sourceType`), so this is a test-coverage gap, not a bug. The existing test at lines 662-676 only tests path validation on update.

### Recommendation: APPROVE

All previously identified issues have been properly fixed. The implementation is solid, well-tested, and follows the established patterns in the codebase.

PR #936, opened 2026-03-26
refactor(daemon): remove auto-recovery logic from query-runner

lsm:
Remove the silent auto-recovery code path from `query-runner.ts`. Startup timeout and conversation-not-found errors are now surfaced immediately to the user instead of being retried transparently.

**Changes:**
- Remove `onStartupTimeoutAutoRecover` callback and `startupTimeoutAutoRecoverAttempts` tracking from `QueryRunnerContext`
- Remove `STARTUP_RECOVERY_DELAY_MS` / `STARTUP_MAX_RETRIES` constants, env vars, and reader functions
- Simplify catch block: always clear message queue, always surface error via `errorManager.handleError()`
- Update error messages to remove "after N attempt(s)" language; keep actionable recovery hints
- Remove `startupMaxRetries` from `handleError()` metadata
- Clean up `agent-session.ts` (remove field + method) and `.env.example`
- Rewrite tests to verify immediate error surfacing (74 tests pass)

Part of: fix model switching bugs (#929) — Task 2.1

### DIFF
--- a/packages/daemon/.env.example
+++ b/packages/daemon/.env.example
@@ -66,17 +66,6 @@ MAX_SESSIONS=10
 # running many concurrent sessions.
 # Default: 15000
 # NEOKAI_SDK_STARTUP_TIMEOUT_MS=15000
-#
-# Base delay (ms) before the first auto-recovery attempt after a startup
-# timeout. Each subsequent attempt doubles this (exponential backoff).
-# Example: 3000 → attempt 1 waits 3s, attempt 2 waits 6s.
-# Default: 3000
-# NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS=3000
-#
-# Maximum number of auto-recovery attempts before surfacing the error to
-# the user. Set to 0 to disable auto-recovery entirely.
-# Default: 2
-# NEOKAI_SDK_STARTUP_MAX_RETRIES=2
 
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 # File Index Configuration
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean refactoring. The auto-recovery logic has been thoroughly excised from all the right places. All 74 tests pass, typecheck and lint are clean.

**What was verified:**

1. **query-runner.ts** -- All auto-recovery machinery removed: `STARTUP_RECOVERY_DELAY_MS`, `STARTUP_MAX_RETRIES`, `getStartupRecoveryDelayMs()`, `getStartupMaxRetries()`, `onStartupTimeoutAutoRecover` callback from `QueryRunnerContext`, `startupTimeoutAutoRecoverAttempts` field, the entire `canAutoRecover` branch with exponential backoff scheduling. The catch block now unconditionally clears `messageQueue` and surfaces errors immediately via `errorManager.handleError()`. Error messages correctly removed "after N attempt(s)" language while preserving actionable hints. The `startupMaxRetries` property was removed from the handleError metadata object.

2. **agent-session.ts** -- `startupTimeoutAutoRecoverAttempts` field and `onStartupTimeoutAutoRecover()` method both removed. No stale references remain in the source tree.

3. **.env.example** -- `NEOKAI_SDK_STARTUP_RECOVERY_DELAY_MS` and `NEOKAI_SDK_STARTUP_MAX_RETRIES` documentation removed. Only `NEOKAI_SDK_STARTUP_TIMEOUT_MS` remains, which is correct since the startup timeout itself is still enforced.

4. **Tests** -- 9 recovery-related tests removed, 7 new tests added that verify the new behavior: always-clear-queue, immediate error surfacing, actionable user messages without retry language, `setIdle` called after error, and `startupMaxRetries` absent from metadata. The `createContext` helper no longer accepts `onStartupTimeoutAutoRecover` or `startupTimeoutAutoRecoverAttempts`.

5. **Error manager compatibility** -- `handleError` accepts `metadata` as `Record<string, unknown>`, so removing `startupMaxRetries` is safe with no type changes needed.

6. **`setIdle` double-call** -- The catch block calls `setIdle` at line 436, and the finally block also calls it at line 490. This is pre-existing (the old code had the same pattern in the non-recovery `else` branch), and `executeDeferredRestartIfPending` is idempotent (guarded by `pendingRestartReason` null-check), so this is harmless.

No issues found.

PR #937, opened 2026-03-26
feat(room): add getCurrentModel() to SessionFactory interface (DB-first)

lsm:
## Summary
- Add `getCurrentModel(sessionId)` to the `SessionFactory` interface in `task-group-manager.ts` for direct server-side model queries without RPC routing through WebSocket
- Implement it in `createSessionFactory()` reading from DB (`ctx.db.getSession()`) — source of truth, not in-memory cache
- Add stub to the global spaces `SessionFactory` adapter in `rpc-handlers/index.ts`

Part of: fix model switching bugs (#929, Task 1.1)

## Test plan
- [x] TypeScript compiles without errors (all consumers compile-safe)
- [x] All 8757 daemon unit tests pass
- [ ] `provision-global-agent.ts` and `session-notification-sink.ts` still compile — they import `SessionFactory` but don't call `getCurrentModel()`

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -518,6 +518,23 @@ export class RoomRuntimeService {
 				}
 				return session.handleModelSwitch(model, provider);
 			},
+			// Reads from DB (source of truth), not the in-memory agentSessions cache.
+			// This avoids stale model info when switchModel() updates the DB but the
+			// cache entry has been evicted (e.g., after daemon restart).
+			//
+			// Note: the DB stores the raw model value from session.config.model, which
+			// may be an alias (e.g., "sonnet") or a resolved ID (e.g.,
+			// "claude-sonnet-4-20250514") depending on what was passed to switchModel().
+			// modelFallbackMap keys use resolved IDs (e.g., "anthropic/claude-sonnet-4-20250514"),
+			// so callers building a fallback key should resolve aliases before lookup.
+			getCurrentModel: async (sessionId) => {
+				const session = ctx.db.getSession(sessionId);
+				if (!session) return null;
+				return {
+					currentModel: session.config.model,
+					currentProvider: session.config.provider ?? 'anthropic',
+				};
+			},
 		};
 	}
 
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -150,7 +150,7 @@ export class RoomRuntimeService {
 		const sessionData = session.getSessionData();
 		return {
 			currentModel: sessionData.config.model,
-			provider: sessionData.config.provider ?? '',
+			provider: sessionData.config.provider ?? 'anthropic',
 		};
 	}
 
@@ -521,18 +521,14 @@ export class RoomRuntimeService {
 			// Reads from DB (source of truth), not the in-memory agentSessions cache.
 			// This avoids stale model info when switchModel() updates the DB but the
 			// cache entry has been evicted (e.g., after daemon restart).
-			//
-			// Note: the DB stores the raw model value from session.config.model, which
-			// may be an alias (e.g., "sonnet") or a resolved ID (e.g.,
-			// "claude-sonnet-4-20250514") depending on what was passed to switchModel().
-			// modelFallbackMap keys use resolved IDs (e.g., "anthropic/claude-sonnet-4-20250514"),
-			// so callers building a fallback key should resolve aliases before lookup.
+			// Returns the raw model value from DB (may be alias or resolved ID).
+			// Callers building modelFallbackMap keys should resolve aliases first.
 			getCurrentModel: async (sessionId) => {
 				const session = ctx.db.getSession(sessionId);
 				if (!session) return null;
 				return {
 					currentModel: session.config.model,
-					currentProvider: session.config.provider ?? 'anthropic',
+					provider: session.config.provider ?? 'anthropic',
 				};
 			},
 		};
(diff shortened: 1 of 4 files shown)


lsm (commented):

---

### Verdict: APPROVE

The implementation is correct, well-documented, and follows existing patterns. The DB-first approach is the right design choice. A few suggestions below for follow-up.

---

### P2: Return type inconsistency with existing modelGet() (room-runtime-service.ts)

There are two overlapping methods that return model info with different shapes:

- Line 147 modelGet() returns { currentModel: string; provider: string } | null -- field name provider, fallback ''
- Line 530 getCurrentModel() (new) returns { currentModel: string; currentProvider: string } | null -- field name currentProvider, fallback 'anthropic'

These serve overlapping purposes and callers will need to know which field name to use. Consider aligning the return types (prefer currentProvider + 'anthropic' fallback in modelGet too, or extract a shared type) to avoid confusion.

### P2: No unit tests for the new method

There are no tests verifying getCurrentModel in createSessionFactory(). A basic test confirming:
- It reads from DB (not in-memory cache) by mocking ctx.db.getSession
- Returns null for non-existent sessions
- Returns the correct currentModel and currentProvider from DB

would provide confidence that the DB-first contract is maintained as the code evolves. The PR description mentions 8757 tests pass, but those are pre-existing tests -- none exercise the new code path.

### P3: Concise alias documentation

The JSDoc comment at line 525-529 is thorough, but could be a single sentence:
// Returns the raw model value from DB (may be alias or resolved ID). Callers building modelFallbackMap keys should resolve aliases first.

---

### What looks good

- DB-first implementation via ctx.db.getSession() correctly avoids stale in-memory cache
- Proper null handling for missing sessions
- Clear interface definition with JSDoc in task-group-manager.ts
- Appropriate null stub in the global spaces adapter rpc-handlers/index.ts
- TypeScript compiles cleanly, all existing tests pass
- Method is declared as required (not optional) on the interface, which is correct since it is a core read operation

lsm (commented):

### P2 fixes assessment

Both P2 issues from the previous review have been properly addressed:

**1. Return type consistency** -- Resolved. `getCurrentModel()` now returns `{ currentModel, provider }` matching the existing `modelGet()` return shape. The `modelGet()` provider default was also aligned from `''` to `'anthropic'`. I verified no consumer code checks for `provider === ''`, so the default change is safe.

**2. Unit test coverage** -- Resolved. The new `SessionFactory.getCurrentModel` test suite covers 4 cases: missing session returns null, correct fields from DB, default provider fallback when provider is undefined, and DB-first verification (data returned even without in-memory cache). Tests use `mockReturnValueOnce` to avoid state leaks between assertions.

### Implementation quality

The code is clean and well-integrated:
- Interface definition in `task-group-manager.ts` has a clear JSDoc
- Implementation in `room-runtime-service.ts` reads from `ctx.db.getSession()` (DB source of truth), with the rationale documented in comments
- Global spaces stub in `rpc-handlers/index.ts` returns `null` as expected
- `createSessionFactory()` is a private method called during runtime construction, so the `serviceAny` cast in tests is reasonable for testing a private API

### Observations (no action needed for this PR)

- Several test files (`provision-global-agent.test.ts`, `session-notification-sink.test.ts`, `provision-global-agent-mcp.test.ts`) mock `SessionFactory` using `as unknown as SessionFactory` casts and omit `getCurrentModel`. This is a pre-existing pattern (predates this PR), and Task 1.3 in the plan doc explicitly calls for updating these mocks later. No issue here.

- The `modelGet()` method (line 147-155 of `room-runtime-service.ts`) reads from the in-memory `agentSessions` cache, while the new `getCurrentModel()` reads from DB. Both return the same shape `{ currentModel, provider } | null` but differ in data source. This is intentional per the plan (DB-first for `getCurrentModel` to avoid stale cache issues), but it's worth noting that if `modelGet()` is eventually replaced by `getCurrentModel()` (as Task 1.2 suggests), consumers should be aware of the semantic difference.

### Verdict

**APPROVE** -- The two commits cleanly implement the interface addition and the return type alignment. Tests are thorough and TypeScript compiles clean.

lsm (commented):

### Summary

The P2 feedback has been properly addressed. The return type alignment and unit tests look good. I have a few observations.

---

### P2 Issues - Properly Addressed

**1. Return Type Alignment** (Fixed)

The return type now correctly uses `provider` instead of `currentProvider`, matching the existing `modelGet()` interface:

- `task-group-manager.ts` line 131: `getCurrentModel(sessionId: string): Promise<{ currentModel: string; provider: string } | null>`
- `room-runtime-service.ts` line 153: `provider: sessionData.config.provider ?? 'anthropic'`
- `room-runtime-service.ts` line 531: `provider: session.config.provider ?? 'anthropic'`

Both `modelGet()` and `getCurrentModel()` now share the same return shape and the same `'anthropic'` default for missing providers.

**2. Unit Tests Added** (Fixed)

The test suite in `room-runtime-service.test.ts` lines 563-622 covers:
- Missing session returns `null`
- Correct field extraction from DB
- Default provider fallback (`'anthropic'`)
- DB-first behavior (comment states it reads from DB, not cache)

---

### P3 - Observations

**Test Coverage: Single Large Test Function**

The test uses a single `it()` block with multiple assertions. While this works, splitting into 4 separate `it()` blocks would improve test isolation and make failures easier to diagnose:

```typescript
it('returns null for missing sessions', ...)
it('returns model/provider from DB', ...)
it('defaults provider to anthropropic when missing', ...)
it('reads from DB, not in-memory cache', ...)
```

This is a minor point - the current test does verify the behavior correctly.

**DB-First Test Assertion**

The test comment says "cache is empty but DB returns data" but the test doesn't actually verify the cache is empty or that the implementation bypasses it. The implementation correctly reads from `ctx.db.getSession()` but there's no assertion proving the cache was bypassed. Still, the implementation is correct - this is more a test documentation observation.

---

### Recommendation: APPROVE

The P2 issues are properly fixed. The return type alignment is correct, the provider default is consistent, and the tests verify the key behaviors. The test organization is a minor nit.

PR #938, opened 2026-03-26
test(room): online test for planner WebSearch capability

lsm:
Adds an online test that verifies the planner agent session can invoke `WebSearch` and that the tool call appears in the message history.

## Changes

- `packages/daemon/tests/online/room/planner-websearch.test.ts` — new online test
  - Creates a room and triggers a planner session via goal creation
  - Embeds unique probe fragment `planner-websearch-probe-2025-v1` in the goal description for dev proxy matching
  - Polls the planner's SDK message history for a `WebSearch` tool_use block
  - Supports both dev proxy mode (`NEOKAI_USE_DEV_PROXY=1`) and real API mode
  - Fails (not skips) when neither dev proxy nor real credentials are available

- `.devproxy/mocks.json` — two new mocks for the planner WebSearch probe
  - **Follow-up mock** (before probe, matches `toolu_websearch_planner_001`): intercepts subsequent calls after WebSearch tool_result, returns end_turn
  - **Probe mock** (matches `planner-websearch-probe-2025-v1`): returns a `WebSearch` tool_use response for the initial planner API call
  - Both mocks are placed before the `.JsonContains` context mocks (known to match all requests with a `messages` key)

## Test run

```bash
NEOKAI_USE_DEV_PROXY=1 bun test packages/daemon/tests/online/room/planner-websearch.test.ts
# 1 pass, 0 fail, ~8s
```

### DIFF
--- a/.devproxy/mocks.json
+++ b/.devproxy/mocks.json
@@ -446,6 +446,98 @@
 			}
 		},
 		{
+			"comment": "Planner WebSearch follow-up — matches the WebSearch tool_use ID from the probe mock. The tool_use_id 'toolu_websearch_planner_001' only appears AFTER the probe mock fires (as tool_use_id in a tool_result block). Returns end_turn so the planner can settle. ORDERING: must be BEFORE the probe mock so that subsequent calls (which contain both the probe fragment and the tool_use_id) match the follow-up, not the probe again.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "toolu_websearch_planner_001"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (planner-websearch follow-up)"
+					}
+				],
+				"body": {
+					"id": "msg_planner_websearch_followup_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED] Based on search results, React remains the most popular JavaScript framework in 2025. Planning will proceed with this information."
+						}
+					],
+					"model": "claude-sonnet-4-5-20250929",
+					"stop_reason": "end_turn",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 600,
+						"output_tokens": 40,
+						"cache_creation_input_tokens": 0,
+						"cache_read_input_tokens": 0,
+						"service_tier": "standard"
+					}
+				}
+			}
+		},
+		{
+			"comment": "Planner WebSearch probe — initial call. Matches the unique fragment embedded in the goal description. Returns a WebSearch tool_use block so the test can assert the planner has WebSearch capability. ORDERING: must appear AFTER the follow-up mock (which has a more specific match) and BEFORE the .JsonContains context mocks which match all requests with a 'messages' key.",
+			"request": {
+				"url": "http://127.0.0.1:8000/v1/messages?beta=true",
+				"method": "POST",
+				"bodyFragment": "planner-websearch-probe-2025-v1"
+			},
+			"response": {
+				"statusCode": 200,
+				"headers": [
+					{
+						"name": "content-type",
+						"value": "application/json"
+					},
+					{
+						"name": "x-mocked-by",
+						"value": "Dev Proxy mocks.json (planner-websearch probe)"
+					}
+				],
+				"body": {
+					"id": "msg_planner_websearch_probe_001",
+					"type": "message",
+					"role": "assistant",
+					"content": [
+						{
+							"type": "text",
+							"text": "[MOCKED] I need to search the web to find the most popular JavaScript framework in 2025."
+						},
+						{
+							"type": "tool_use",
+							"id": "toolu_websearch_planner_001",
+							"name": "WebSearch",
+							"input": {
+								"query": "most popular JavaScript framework 2025"
+							}
+						}
+					],
+					"model": "claude-sonnet-4-5-20250929",
+					"stop_reason": "tool_use",
+					"stop_sequence": null,
+					"usage": {
+						"input_tokens": 500,
+						"output_tokens": 80,
+						"cache_creation_input_tokens": 0,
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Summary

The test is well-structured and follows existing patterns correctly. The mock ordering is sound, the hard-fail behavior is preserved, and the test has been verified to pass. I have a few suggestions but no blocking issues.

### What works well

- **Mock ordering is correct.** The follow-up mock (matching `toolu_websearch_planner_001`) is placed before the probe mock (matching `planner-websearch-probe-2025-v1`), ensuring follow-up requests that contain both fragments match the follow-up, not the probe again. Both are placed before the broader `.JsonContains` context mocks and the catch-all beta mock.

- **Hard-fail behavior is preserved.** No skip guards for missing credentials. Without `NEOKAI_USE_DEV_PROXY=1` and without API keys, the planner's SDK calls will fail and the polling timeouts will cause the test to fail naturally.

- **SDK message structure parsing is consistent.** The `containsWebSearchToolUse` function accesses `msg['type']` and `msg['message'].content` which matches the pattern used by other tests (e.g., `glm-sdk-minimal.test.ts` line 80).

- **Probe fragment flows correctly into the API request.** The goal description containing the probe fragment is injected into the planner worker session via `buildPlannerTaskMessage` (planner-agent.ts:350), so the first API call from the planner will contain the fragment and match the probe mock.

- **Test verified to pass.** The PR description confirms the test passed in ~8s with dev proxy enabled.

1. **(P2) Double POLL_TIMEOUT budget.** The test uses `POLL_TIMEOUT` for both `waitForGroup` and `pollForWebSearchToolUse` sequentially. In mock mode this is 30s + 30s = 60s of potential polling, while `TEST_TIMEOUT` is also 60s. If the group poll takes most of its timeout, the WebSearch poll might not have enough time. Consider either reducing `POLL_TIMEOUT` for the group poll (e.g., 15s in mock mode, since the group should appear quickly after the task) or using separate timeout constants.
   - File: `packages/daemon/tests/online/room/planner-websearch.test.ts`, lines 37, 95, 101

2. **(P2) `waitForGroup` silently returns null on timeout.** Unlike `waitForTask` (which throws with diagnostic info including current task states), `waitForGroup` returns `null` when the timeout expires. The `expect(group).toBeTruthy()` assertion on line 96 would then fire, but the error message would not explain *why* the group was not found (e.g., task still pending, task failed before group creation). Consider throwing with diagnostic context similar to `waitForTask`.
   - File: `packages/daemon/tests/online/room/planner-websearch.test.ts`, lines 115-146

3. **(P2) Consider adding a brief diagnostic log on timeout.** In `pollForWebSearchToolUse`, when the timeout expires and returns false, the `expect(found).toBe(true)` assertion tells you the WebSearch tool was not found, but gives no hint about what messages *were* in the session. A debug-level log of the message types/counts would speed up future debugging.
   - File: `packages/daemon/tests/online/room/planner-websearch.test.ts`, lines 154-175

4. **(P3) Type annotation for group could use the existing `task.getGroup` return shape.** The `waitForGroup` return type manually declares the group shape. The actual handler (task-handlers.ts:589-602) returns additional fields (`feedbackIteration`, `submittedForReview`, `approved`, `createdAt`, `completedAt`). While the test only needs `workerSessionId`, the type could reference a shared interface or at minimum include a `// eslint-disable-next-line @typescript-eslint/no-explicit-any` if intentionally partial. This is minor since `Record<string, unknown>` style typing is common in test helpers.

### Recommendation

APPROVE. The implementation is correct, well-documented, and follows project conventions. The mock infrastructure is properly used with correct ordering, and the test provides meaningful coverage of the planner WebSearch capability.

lsm (commented):

### Re-review: All three P2 issues from the previous review have been properly addressed

**P2 Fix 1: Split POLL_TIMEOUT into separate constants.** `GROUP_POLL_TIMEOUT` (10s mock / 60s real) and `WEBSEARCH_POLL_TIMEOUT` (30s mock / 180s real) are now separate. The group poll uses the shorter budget, while the WebSearch poll gets the longer budget. Total sequential worst-case in mock mode (30s + 10s + 30s = 70s) slightly exceeds `TEST_TIMEOUT` (60s), but the `waitForTask` call completes almost instantly in practice (the PR confirms ~8s total), so this is not a real risk.

**P2 Fix 2: waitForGroup now throws with diagnostics.** The function no longer returns null silently. On timeout, it fetches `task.list` and formats a per-task summary (type, status, title) into the error message, matching the diagnostic pattern used by `waitForTask` in `room-test-helpers.ts`.

**P2 Fix 3: Diagnostic log on pollForWebSearchToolUse timeout.** On timeout, the function logs a per-message summary showing each message's type and content block types/names (e.g., `[assistant] text, tool_use:WebSearch`). The `console.log` is correctly placed in a `*.test.ts` file, which is exempt from the `no-console` lint rule per `.oxlintrc.json`.

### No new issues found

The implementation is clean, well-documented, and follows project conventions. The mock ordering in `.devproxy/mocks.json` and the CI matrix validator update remain correct from the initial review.

### Recommendation: APPROVE

lsm (commented):

### Re-Review Summary

All three P2 issues from the previous review have been properly addressed in commit `3d1844df`:

#### P2 Fix #1: Double POLL_TIMEOUT budget ✅
Split into separate constants with appropriate values and clear comments (lines 40-41):
- `GROUP_POLL_TIMEOUT`: 10s (mock) / 60s (real)
- `WEBSEARCH_POLL_TIMEOUT`: 30s (mock) / 180s (real)

The comment on lines 35-37 explains the rationale: group creation is fast in mock mode, so a shorter budget prevents eating into the WebSearch poll time.

#### P2 Fix #2: waitForGroup silent null ✅
The local `waitForGroup` function (lines 118-158) now throws with diagnostic context instead of returning null. The error includes task type, status, and title for all tasks in the room, matching the pattern used by `waitForTask` in `room-test-helpers.ts`.

#### P2 Fix #3: No diagnostic log on timeout ✅
`pollForWebSearchToolUse` (lines 189-202) now logs a message-type summary with block details when WebSearch tool use is not found. This provides visibility into what the planner produced instead.

### Additional Verification

- **Typecheck**: Passes
- **Lint**: 0 warnings, 0 errors
- **Mock ordering**: Correctly sequenced in `.devproxy/mocks.json` (follow-up → probe → JsonContains)
- **Matrix validator**: Test file added to `ROOM_FILES` array

No new issues introduced. The implementation is complete and well-documented.

PR #940, opened 2026-03-26
feat(daemon): SKILL_VALIDATE job queue for async skill validation

lsm:
## Summary
- Add `SKILL_VALIDATE` job queue constant and `createSkillValidateHandler` that validates skills asynchronously after add/update
- Handler checks plugin path accessibility (`fs.accessSync`) and MCP server reference existence (`AppMcpServerRepository.get`)
- Builtin skills pass validation as no-op
- `SkillsManager` enqueues a validation job in `addSkill()` and `updateSkill()` (when config changes) via optional `jobQueue` dependency
- Handler registered in `app.ts` before `jobProcessor.start()`
- 8 unit tests covering all skill types and error cases

## Test plan
- [x] `bun test packages/daemon/tests/unit/job-handlers/skill-validate.handler.test.ts` — 8/8 pass
- [x] `bun test packages/daemon/tests/unit/skills-manager.test.ts` — 38/38 pass (no regression)
- [x] `bun run typecheck` — passes
- [x] `bun run lint` — 0 errors

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -23,7 +23,8 @@ import type { TaskAgentManager } from './lib/space/runtime/task-agent-manager';
 import { JobQueueRepository } from './storage/repositories/job-queue-repository';
 import { JobQueueProcessor } from './storage/job-queue-processor';
 import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
-import { JOB_QUEUE_CLEANUP } from './lib/job-queue-constants';
+import { createSkillValidateHandler } from './lib/job-handlers/skill-validate.handler';
+import { JOB_QUEUE_CLEANUP, SKILL_VALIDATE } from './lib/job-queue-constants';
 import { AppMcpLifecycleManager, seedDefaultMcpEntries } from './lib/mcp';
 import { FileIndex } from './lib/file-index';
 import { SkillsManager } from './lib/skills-manager';
@@ -276,7 +277,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	seedDefaultMcpEntries(db);
 
 	// Instantiate Skills manager and initialize built-in skills
-	const skillsManager = new SkillsManager(db.skills, db.appMcpServers);
+	const skillsManager = new SkillsManager(db.skills, db.appMcpServers, jobQueue);
 	skillsManager.initializeBuiltins();
 
 	// Initialize workspace file index (non-blocking — init runs in the background)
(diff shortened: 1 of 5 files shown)


lsm (commented):

Overall this is a clean, well-structured implementation that follows the existing `JobQueueProcessor` + handler pattern. The factory function, test coverage, and integration wiring in `app.ts` are all solid. I found one bug and two design issues worth addressing.

### P0: Validation status never set to `invalid` on failure

**File:** `packages/daemon/src/lib/job-handlers/skill-validate.handler.ts`, lines 28-38

When validation fails (e.g., `accessSync` throws for a non-existent plugin path, or the MCP server entry is missing), the handler throws an error. The `JobQueueProcessor.processJob()` catches this and marks the **job** as failed via `repo.fail()`, but `setSkillValidationStatus(skillId, 'invalid')` is never called. The skill remains stuck in `'pending'` status permanently after a validation failure.

The `SkillValidationStatus` type explicitly includes `'invalid'` as a valid value, but no code path ever sets it.

**Fix:** Wrap the validation logic in a try/catch inside the handler. On catch, call `skillsManager.setSkillValidationStatus(skillId, 'invalid')` before re-throwing (or return `{ valid: false, skillId }` without throwing, depending on whether you want the job itself to be marked failed).

### P1: `accessSync` blocks the event loop

**File:** `packages/daemon/src/lib/job-handlers/skill-validate.handler.ts`, line 30

The handler is `async` but uses synchronous `accessSync`. For network-mounted filesystems or slow storage, this blocks the event loop. Since the handler is already async, use `await fs.promises.access(skill.config.pluginPath, constants.R_OK)` instead.

### P1: `updateSkill` does not reset validation status to `pending` before re-enqueuing

**File:** `packages/daemon/src/lib/skills-manager.ts`, lines 91-94

When `updateSkill()` detects a config change, it enqueues a new validation job but does not reset the skill's `validationStatus` to `'pending'`. This means the UI may show `'valid'` (from a previous successful validation) while re-validation is queued or in progress. The status should be reset to `'pending'` before or immediately after enqueuing the new validation job.

### Minor observations (P3)

- The `setJobQueue()` method on `SkillsManager` (line 45) is unused -- `app.ts` passes `jobQueue` directly via the constructor (line 280). This dead method could be removed, or the constructor parameter approach could be removed in favor of `setJobQueue()` -- but having both is unnecessary.
- Test at line 114 uses `'/'` as a "path that always exists" for plugin validation. This works but is platform-specific and assumes the test runner always has read access to root.

lsm (commented):

All four issues from the previous review have been properly addressed in commit `76faa6c8f`:

1. **P0 fix (validation failures mark skill invalid)** -- The handler now wraps validation logic in try/catch, calling `setSkillValidationStatus(skillId, 'invalid')` before re-throwing. Tests updated to assert this behavior.

2. **P1 fix (async fs.access)** -- Replaced `accessSync` from `node:fs` with `access` from `node:fs/promises`, properly awaited. The plugin path test now uses `mkdtempSync` for a real temp directory instead of `/`.

3. **P1 fix (updateSkill resets to pending)** -- `updateSkill()` now calls `this.repo.setValidationStatus(id, 'pending')` before enqueuing the validation job when config changes.

4. **P3 fix (unused setJobQueue)** -- Removed. The job queue is now passed exclusively via the constructor, and `app.ts` passes it directly: `new SkillsManager(db.skills, db.appMcpServers, jobQueue)`.

All 8 handler tests and 38 skills-manager tests pass. No new issues found.

**Minor note (P3):** The PR description body still references `fs.accessSync` which is now stale after the fix commit. Not blocking.

Recommendation: **APPROVE**

PR #944, opened 2026-03-26
feat(daemon): Skills injection in QueryOptionsBuilder (Task 3.1)

lsm:
Extend QueryOptionsBuilder to inject enabled skills into SDK query options.

**Changes:**
- Add `skillsManager` and `appMcpServerRepo` to `QueryOptionsBuilderContext` (optional, backwards compatible)
- Add `buildPluginsFromSkills()` — maps plugin skills to `SdkPluginConfig` entries
- Add `getMcpServersFromSkills()` — resolves MCP server skills via `AppMcpServerRepository` and maps to `McpServerConfig`
- Merge skill plugins/servers with existing session config in `build()`
- Skill-injected MCP servers appear in `mcpServers` map, ensuring they work with `strictMcpConfig` sessions (room_chat) that only allow explicitly configured servers
- Wire `SkillsManager` + `AppMcpServerRepo` through `DaemonApp → SessionManager → AgentSession → QueryOptionsBuilder`
- Move SkillsManager initialization before SessionManager in `app.ts`

**Tests (12 new):** plugin injection, MCP server injection (stdio/SSE/HTTP), disabled skill exclusion, missing app_mcp_server graceful skip, strictMcpConfig compatibility, merge with existing config, builtin skill passthrough.

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -204,6 +204,13 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	const eventBus = createDaemonHub('daemon');
 	await eventBus.initialize();
 
+	// Initialize application-level MCP and Skills managers before SessionManager
+	// so AgentSession can inject skills into SDK query options.
+	const appMcpManager = new AppMcpLifecycleManager(db);
+	seedDefaultMcpEntries(db);
+	const skillsManager = new SkillsManager(db.skills, db.appMcpServers, jobQueue);
+	skillsManager.initializeBuiltins();
+
 	// Initialize session manager (with EventBus, SettingsManager, no StateManager dependency!)
 	// Use reactiveDb.db so sdk_messages writes emitted by AgentSession pipelines
 	// trigger LiveQuery invalidation immediately.
@@ -220,7 +227,9 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 			workspaceRoot: config.workspaceRoot,
 		},
 		jobQueue,
-		jobProcessor
+		jobProcessor,
+		skillsManager,
+		db.appMcpServers
 	);
 
 	// Register session title generation handler before jobProcessor starts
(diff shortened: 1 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured implementation. The code correctly extends `QueryOptionsBuilder` to inject skills into SDK query options, with proper wiring through the dependency chain.

### What works well

1. **Correct merge semantics**: `mergeMcpServers` spreads skill servers first, then config servers (`{ ...skillServers, ...configServers }`), so explicit session config correctly takes priority over skill-injected servers. Same pattern for plugins — config plugins come first, skill plugins are appended.

2. **strictMcpConfig handling is correct**: The key insight — skill-injected MCP servers must appear in the `mcpServers` map for `strictMcpConfig: true` sessions (room_chat) to accept them — is properly implemented. The room_chat block at line 240 generates `mcpServerWildcards` from `Object.keys(queryOptions.mcpServers ?? {})`, which now includes skill-injected servers. This means skill MCP tools are auto-allowed.

3. **Backwards compatibility**: All new params (`skillsManager`, `appMcpServerRepo`) are optional throughout the chain (context interface, AgentSession constructor, SessionManager constructor). Existing callers are unaffected.

4. **Graceful degradation**: Missing `appMcpServerRepo` returns empty object. Deleted `app_mcp_servers` entries are silently skipped. Missing `skillsManager` returns empty arrays. All correct.

5. **Exhaustive type handling**: `appMcpServerToSdkConfig` uses a `never` exhaustive check on `sourceType`, which will catch any future source types at compile time.

6. **Initialization ordering**: Moving `SkillsManager` and `AppMcpLifecycleManager` before `SessionManager` in `app.ts` is safe — both only depend on `db` which is already initialized. No downstream dependency issues.

7. **Tests are thorough**: 12 new tests cover plugin injection, MCP injection (stdio/SSE/HTTP), disabled skill exclusion, deleted server graceful skip, strictMcpConfig compatibility with allowedTools wildcard check, merge with existing config, and builtin skill passthrough.

### Minor observations (P3)

- The `appMcpServerToSdkConfig` method uses non-null assertions (`server.command!`, `server.url!`) for fields that are typed as optional. This is structurally correct because the DB schema guarantees these fields exist for their respective `sourceType`, but a defensive check would be marginally safer. Not blocking since the exhaustive switch guarantees only valid source types reach each branch.

- The inline `import(...)` type syntax in `agent-session.ts` (lines 245-246) works but is slightly unusual. It avoids adding imports at the top of a large file, which is a reasonable trade-off.

No bugs, no security issues, no missing edge cases found. The implementation matches the task requirements exactly.

PR #945, opened 2026-03-26
feat(web): global Skills LiveQuery store and hook

lsm:
Implements Task 5.1 — global Skills registry LiveQuery store and hook.

- `SkillsStore` subscribes to `skills.list` named query via `liveQuery.subscribe`, handles snapshot/delta events to keep `skills` signal up to date, and re-subscribes on reconnect
- Four mutation methods (`addSkill`, `updateSkill`, `removeSkill`, `setEnabled`) call the correct RPC endpoints (`skill.create`, `skill.update`, `skill.delete`, `skill.setEnabled`)
- Singleton `skillsStore` exported; `useSkills` hook manages subscribe/unsubscribe lifecycle
- 31 unit tests: snapshot, delta add/remove/update, wrong-subscriptionId guard, stale-event guard, race guard, reconnect, all mutations, error propagation

### DIFF
--- a/packages/web/src/hooks/useSkills.ts
+++ b/packages/web/src/hooks/useSkills.ts
@@ -1,0 +1,58 @@
+/**
+ * useSkills Hook
+ *
+ * Lifecycle adapter that manages the global Skills registry LiveQuery subscription.
+ *
+ * Responsibilities:
+ * - On mount: call skillsStore.subscribe()
+ * - On unmount: call skillsStore.unsubscribe()
+ *
+ * The store owns the subscription handles and cleanup logic.
+ * This hook is purely a lifecycle adapter between the Preact component
+ * tree and the skills store's LiveQuery methods.
+ *
+ * @returns The store signals for use in components.
+ *
+ * @example
+ * ```tsx
+ * export default function SkillsRegistry() {
+ *   const { skills, isLoading, error } = useSkills();
+ *   if (isLoading.value) return <div>Loading...</div>;
+ *   return <ul>{skills.value.map(s => <li key={s.id}>{s.displayName}</li>)}</ul>;
+ * }
+ * ```
+ */
+
+import { useEffect } from 'preact/hooks';
+import type { Signal } from '@preact/signals';
+import type { AppSkill } from '@neokai/shared';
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Summary

Clean implementation that closely follows the established `AppMcpStore` pattern. The store, hook, and tests are well-structured. All 31 tests pass. The code correctly handles snapshot/delta events, reconnection, stale-event guards, and race conditions between subscribe/unsubscribe.

**Recommendation: APPROVE** — only minor nits below.

### P3 — Naming inconsistency: `isLoading` vs `loading`

`AppMcpStore` uses `loading` for its loading signal, while `SkillsStore` uses `isLoading`. This is a minor inconsistency across the two stores that follow the same pattern. Not blocking, but worth noting for consistency — either align this store to use `loading`, or plan to rename the other store in a follow-up.

- `packages/web/src/lib/skills-store.ts` line 34: `readonly isLoading = signal<boolean>(false);`
- `packages/web/src/lib/app-mcp-store.ts` line 29: `readonly loading = signal<boolean>(false);`

### P3 — No hook test file

The `useSkills` hook has no dedicated test file. The hook is thin (just subscribe on mount, unsubscribe on unmount), so this is acceptable given the store itself has thorough coverage. But a lightweight test verifying the mount/unmount lifecycle would complete the picture.

### Notes

- The store correctly mirrors the `AppMcpStore` architecture: singleton class, `subscribe()`/`unsubscribe()` lifecycle, stale-event guards via `activeSubscriptionIds`, reconnect handler, race-condition guards for async subscribe/unsubscribe interleaving.
- Mutation methods (`addSkill`, `updateSkill`, `removeSkill`, `setEnabled`) correctly delegate to RPC and return server responses, relying on LiveQuery deltas for state updates rather than optimistic local updates — consistent with the project pattern.
- The `SkillsStore` class is not exported (only the singleton instance), which is fine for the hook usage but means tests must reset signals manually in `beforeEach` rather than instantiating fresh stores. The test file handles this correctly.
- All daemon-side RPC handlers (`skill.create`, `skill.update`, `skill.delete`, `skill.setEnabled`) and the `skills.list` named query exist, confirming the client-server contract is complete.

PR #949, opened 2026-03-26
feat(skills): built-in web-search-mcp skill via Brave Search MCP

lsm:
Implements Task 6.4: registers a built-in opt-in skill for web search via the Brave Search MCP server.

- `SkillsManager.initializeBuiltins()` now upserts a `web-search-brave` app MCP server entry and a `web-search-mcp` skill referencing it (`builtIn: true`, `enabled: false`)
- `docs/architecture/web-search-mcp.md` documents the chosen server, configuration, and rationale
- 7 new unit tests: registration, backing server creation, config reference, idempotency, deletion guard, enable toggle, disabled-by-default

**Usage:** set `BRAVE_API_KEY` env var, then enable the skill in Settings → Skills.

### DIFF
--- a/docs/architecture/web-search-mcp.md
+++ b/docs/architecture/web-search-mcp.md
@@ -25,12 +25,17 @@ The skill is registered in `SkillsManager.initializeBuiltins()` as an `mcp_serve
 
 ### app_mcp_servers entry
 
+The skill reuses the `brave-search` entry already seeded by `seedDefaultMcpEntries()` (which
+runs first in `app.ts`). If the entry is somehow absent, `initializeBuiltins()` creates it as a
+fallback using identical parameters:
+
 ```
-name:        web-search-brave
+name:        brave-search
 sourceType:  stdio
 command:     npx
 args:        ["-y", "@modelcontextprotocol/server-brave-search"]
-enabled:     true
+env:         {}
+enabled:     false
 ```
 
 ### Skill entry
@@ -39,7 +44,7 @@ enabled:     true
 name:         web-search-mcp
 displayName:  Web Search (MCP)
 sourceType:   mcp_server
-config:       { type: "mcp_server", appMcpServerId: <id of web-search-brave entry> }
+config:       { type: "mcp_server", appMcpServerId: <id of brave-search entry> }
 enabled:      false   (opt-in)
 builtIn:      true
 ```
@@ -47,7 +52,10 @@ builtIn:      true
 ## Usage
 
 1. Obtain a Brave Search API key from https://brave.com/search/api/
-2. Set `BRAVE_API_KEY` in the daemon environment (`.env` or shell)
+2. Set `BRAVE_API_KEY` in the daemon process environment (e.g. via `.env` at the worktree root or
+   as a shell environment variable before starting the daemon). The MCP server subprocess inherits
+   the full daemon process environment, so no additional wiring is needed — the key just needs to
+   be present in the daemon's `process.env` at startup.
 3. Enable the skill in the Skills settings UI
 4. The MCP server is automatically injected into new session `mcpServers`
 
--- a/docs/architecture/web-search-mcp.md
+++ b/docs/architecture/web-search-mcp.md
@@ -1,0 +1,64 @@
+# Web Search MCP Skill
+
+## Overview
+
+The Web Search MCP skill provides agents with web search capability via a Brave Search MCP server.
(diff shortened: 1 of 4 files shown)


lsm (commented):

### Summary

The implementation correctly registers a built-in `web-search-mcp` skill backed by a Brave Search MCP server entry, with proper idempotency, deletion protection, disabled-by-default behavior, and thorough unit tests. The code is clean and well-structured. However, there are two meaningful issues worth addressing.

### Issues

**[P2] Duplicate app_mcp_servers entry for Brave Search**
`packages/daemon/src/lib/skills-manager.ts:142-153` creates an `app_mcp_servers` entry named `web-search-brave`, but `packages/daemon/src/lib/mcp/seed-defaults.ts:31-41` already seeds an entry named `brave-search` pointing to the exact same `@modelcontextprotocol/server-brave-search` package. Both run on daemon startup (see `app.ts:210-212`), resulting in two separate database entries for the same underlying MCP server. This is confusing for users who may see both in the MCP server list. The skill should either reuse the existing `brave-search` entry or the seed-defaults entry should be removed in favor of the skill-managed one.

**[P2] No test verifying actual session injection**
The acceptance criteria state "Enabling the skill injects the MCP server into session options" and "Unit tests verify registration and injection." While the tests cover registration, idempotency, enable/disable, and deletion guard thoroughly, none of the 7 new tests actually verify that `QueryOptionsBuilder` includes the Brave Search MCP server in session options when the skill is enabled. The injection path exists (via `getMcpServersFromSkills()` in `query-options-builder.ts`), but it is not tested as part of this PR. A test that enables the skill and then verifies `QueryOptionsBuilder.build()` output contains the `web-search-mcp` MCP server entry would close this gap.

**[P3] Documentation mentions env var but no env passthrough is configured**
`packages/daemon/src/lib/skills-manager.ts:151` creates the app_mcp_servers entry without an `env` field. The `BRAVE_API_KEY` will only work if the SDK inherits the full process environment when spawning stdio subprocesses. This is consistent with the existing `brave-search` seed entry pattern, but the documentation at `docs/architecture/web-search-mcp.md:49-52` could note that the key must be in the daemon process environment since it relies on process env inheritance.

The duplicate MCP server entry (P2) will create user-facing confusion and should be resolved before merging. The missing injection test (P2) should also be addressed to satisfy the stated acceptance criteria.

lsm (commented):

**Recommendation: APPROVE**

This is a re-review after previous P2 feedback was addressed. All three issues from the prior review have been resolved correctly.

### Previous feedback resolution

1. **Duplicate MCP entry (P2)** -- Fixed. `initWebSearchBraveMcp()` now calls `this.appMcpServerRepo.getByName('brave-search')` first and only falls back to `create()` if absent. This aligns with `seedDefaultMcpEntries()` which seeds the same entry. A dedicated test (`reuses pre-existing brave-search app_mcp_servers entry`) confirms exactly one entry exists and the skill references the seeded ID.

2. **Missing injection test (P2)** -- Fixed. Two new tests in `query-options-builder.test.ts` verify: (a) disabled skill produces no `mcpServers['web-search-mcp']` entry, (b) enabled skill injects the correct `npx` command with `@modelcontextprotocol/server-brave-search` args. Both tests use a real in-memory SQLite database with `createTables()`, which is a more realistic setup than mocking.

3. **Doc clarification (P3)** -- Fixed. `web-search-mcp.md` now explains that `BRAVE_API_KEY` is inherited by the stdio subprocess from the daemon's `process.env`.

### Current state assessment

**Correctness**: The implementation correctly registers a built-in skill during `initializeBuiltins()`, references the seeded `brave-search` app MCP server entry, and the `QueryOptionsBuilder` injection path works end-to-end via the existing `getMcpServersFromSkills()` mechanism.

**Initialization order**: Verified that `app.ts` calls `seedDefaultMcpEntries(db)` at line 210 before `skillsManager.initializeBuiltins()` at line 212, so the `getByName` lookup will always find the seeded entry in normal operation.

**Idempotency**: Covered by test -- calling `initializeBuiltins()` twice produces exactly one skill and one server entry.

**Test coverage**: 8 new tests in `skills-manager.test.ts` (registration, backing server creation, config reference, idempotency, deletion guard, enable toggle, disabled-by-default, reuse of pre-existing entry) plus 2 integration tests in `query-options-builder.test.ts`. All 125 tests across both files pass.

**No issues found.** The implementation is clean, well-tested, and correctly addresses all acceptance criteria.

PR #950, opened 2026-03-26
fix(test): replace false-positive history check with sdkSessionId DB assertion

lsm:
## Summary

- Replace the "message history preserved after model switch" test with **"sdkSessionId unchanged after post-switch message completes"** — the correct structural proof of SDK session resume

## Why

The old test searched `sdkMessages` for a pre-switch user message. This was a false positive: sdkMessages are stored in the DB by daemon session ID and never deleted on model switch, so finding a pre-switch message is always true regardless of whether the SDK session was resumed or recreated.

The new test reads `sdkSessionId` from the DB via `session.get` RPC before and after a post-switch query completes. If the SDK resumed, `system:init` returns the same ID; if recreated, it emits a new one.

## Test plan

- [x] `bun run format` / `bun run typecheck` / `oxlint` — all pass
- [ ] CI: Daemon Online (cross-provider) — pending

### DIFF
--- a/packages/daemon/tests/online/cross-provider/conversation-continuity-after-switch.test.ts
+++ b/packages/daemon/tests/online/cross-provider/conversation-continuity-after-switch.test.ts
@@ -6,7 +6,7 @@
  * switching tests (PR #930) by asserting that:
  * - sdkSessionId is preserved across model switches
  * - Message count does not reset
- * - Message history preserved after cross-provider switch (structural, not LLM-dependent)
+ * - sdkSessionId unchanged after post-switch query completes (cross-provider)
  * - sdkSessionId survives multiple rapid switches
  * - DB correctly persists sdkSessionId after switch
  *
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

The change correctly identifies and fixes a false-positive test. The analysis is sound:

**What was wrong:** The old test searched `sdkMessages` for a pre-switch user message marker. Since `sdkMessages` are stored by daemon session ID and never purged on model switch, this assertion was tautologically true -- it could never fail regardless of whether the SDK session was resumed or recreated.

**What the fix does:** The replacement test captures `sdkSessionId` before the switch, then sends a post-switch message (forcing the SDK to emit a fresh `system:init`), and asserts the ID is unchanged. If the SDK had recreated the session instead of resuming, `system:init` would emit a new ID, causing the test to fail. This is a genuine structural proof of session resume.

**Key differentiator from existing tests:** Test 1 ("sdkSessionId is preserved after cross-provider model switch") only checks `sdkSessionId` after the switch without sending a post-switch message. The new test forces the SDK to actually process a query on the new provider, which is the operation that would reveal a recreated session. This is a meaningfully stronger assertion.

### P3 nits

1. **Misleading comments (lines 245, 265):** The comments say "Read sdkSessionId from DB via session.get" but `session.get` RPC actually returns the in-memory `Session` object via `getSessionAsync() -> getSessionData()`, not a direct DB read. The distinction matters because the test's value comes from the post-switch query triggering a fresh `system:init`, not from reading from DB specifically. The comments should say "Read sdkSessionId via session.get RPC" without the "from DB" claim.

Otherwise clean: unused imports properly removed, test timeout is appropriate (90s), and the test structure follows the same pattern as the surrounding tests.

PR #964, opened 2026-03-26
feat(space): enforce workflow run status lifecycle with transition guards

lsm:
## Summary

- New `WorkflowRunStatusMachine` module defining the agent-centric lifecycle: `pending→in_progress`, `in_progress→completed/needs_attention/cancelled`, `needs_attention→in_progress` (human resolved), terminal states (`completed`/`cancelled`) with no outbound transitions
- `SpaceWorkflowRunRepository.transitionStatus()`: atomic read-validate-write with guard enforcement
- All 5 status-update sites in `SpaceRuntime` now use `transitionStatus()` instead of `updateStatus()`
- New `spaceWorkflowRun.resume` RPC handler for `needs_attention→in_progress` human-resolved unblocking
- 25 new unit tests covering all transitions, completedAt stamping, rehydration after restart, and the human-resume path

## Test plan
- [x] `canTransition` returns correct bool for all 25 source×target pairs
- [x] `assertValidTransition` throws with run ID and allowed-targets in message
- [x] `transitionStatus` persists status, stamps `completedAt` for terminal states
- [x] `transitionStatus` throws on invalid transitions without mutating DB
- [x] `getRehydratableRuns` returns `in_progress` + `needs_attention`, excludes others
- [x] Full suite: 1544 space unit tests pass, 0 failures

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts
@@ -126,6 +126,42 @@ export function setupSpaceWorkflowRunHandlers(
 		return { run };
 	});
 
+	// ─── spaceWorkflowRun.resume ─────────────────────────────────────────────
+	//
+	// Resumes a run that is in needs_attention state after a human has resolved the
+	// blocking issue. Transitions needs_attention → in_progress so the tick loop
+	// will resume processing on the next cycle.
+	messageHub.onRequest('spaceWorkflowRun.resume', async (data) => {
+		const params = data as { id: string };
+
+		if (!params.id) throw new Error('id is required');
+
+		const run = workflowRunRepo.getRun(params.id);
+		if (!run) throw new Error(`WorkflowRun not found: ${params.id}`);
+
+		if (run.status !== 'needs_attention') {
+			throw new Error(
+				`Cannot resume run ${params.id}: expected status 'needs_attention', got '${run.status}'`
+			);
+		}
+
+		// needs_attention → in_progress (human resolved the blocking issue)
+		const updated = workflowRunRepo.transitionStatus(params.id, 'in_progress');
+
+		daemonHub
+			.emit('space.workflowRun.updated', {
+				sessionId: 'global',
+				spaceId: run.spaceId,
+				runId: run.id,
+				run: updated,
+			})
+			.catch((err) => {
+				log.warn('Failed to emit space.workflowRun.updated:', err);
+			});
+
+		return { run: updated };
+	});
+
 	// ─── spaceWorkflowRun.cancel ─────────────────────────────────────────────
 	messageHub.onRequest('spaceWorkflowRun.cancel', async (data) => {
 		const params = data as { id: string };
@@ -153,9 +189,8 @@ export function setupSpaceWorkflowRunHandlers(
 			}
 		}
 
-		// Cancel the run
-		const updated = workflowRunRepo.updateStatus(params.id, 'cancelled');
-		if (!updated) throw new Error(`WorkflowRun not found: ${params.id}`);
+		// Cancel the run (pending/in_progress/needs_attention → cancelled)
+		const updated = workflowRunRepo.transitionStatus(params.id, 'cancelled');
 
 		daemonHub
 			.emit('space.workflowRun.updated', {
--- a/packages/daemon/src/lib/space/runtime/space-runtime.ts
+++ b/packages/daemon/src/lib/space/runtime/space-runtime.ts
@@ -327,7 +327,7 @@ export class SpaceRuntime {
 			maxIterations: workflow.maxIterations,
 		});
 
-		const run = this.config.workflowRunRepo.updateStatus(pendingRun.id, 'in_progress')!;
+		const run = this.config.workflowRunRepo.transitionStatus(pendingRun.id, 'in_progress');
 
 		// Register executor and meta. If a later step fails, we must clean these up.
 		const meta: ExecutorMeta = { workflow, spaceId, workspacePath: space.workspacePath };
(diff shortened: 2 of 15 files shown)


lsm (commented):

Overall this is a clean, well-structured implementation of the workflow run status lifecycle. The state machine is correct and complete, the tests are thorough, and the migration from `updateStatus` to `transitionStatus` is consistent across `SpaceRuntime`. A few issues to address:

### P1: Missing transition guard in `space-agent-tools.ts` `cancel_task`

**File:** `packages/daemon/src/lib/space/tools/space-agent-tools.ts`, lines 378-379

The `cancel_task` handler calls `transitionStatus(task.workflowRunId, 'cancelled')` without first checking if the transition is valid. If the workflow run is already in a terminal state (`completed` or `cancelled`), `transitionStatus` will throw with "Invalid workflow run status transition". While the outer `catch` at line 389 prevents a crash, the agent will receive a misleading error message ("Invalid workflow run status transition") instead of a success response indicating the task was cancelled.

Compare with `global-spaces-tools.ts` (lines 423-426), which correctly guards with `canTransition()` first:
```ts
const existingRun = workflowRunRepo.getRun(task.workflowRunId);
if (existingRun && canTransition(existingRun.status, 'cancelled')) {
    workflowRunRepo.transitionStatus(task.workflowRunId, 'cancelled');
    ...
}
```

The same guard-then-transition pattern should be applied in `space-agent-tools.ts`.

### P2: `updateStatus()` remains public, creating a bypass path

**File:** `packages/daemon/src/storage/repositories/space-workflow-run-repository.ts`, line 179

The unguarded `updateStatus()` method is still public. Any new code could call it directly and bypass the state machine entirely. The comment "no transition guard -- use transitionStatus() for guarded updates" is helpful, but making it `private` (since it's only used internally by `updateRun` and `transitionStatus`) would provide compile-time enforcement. If it needs to remain accessible for edge cases, consider renaming to something like `updateStatusUnchecked` to make the bypass explicit.

### P3: Minor -- redundant pre-check in resume handler

**File:** `packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts`, lines 142-146

The manual `run.status !== 'needs_attention'` check before `transitionStatus` is technically redundant since the state machine would reject the invalid transition. However, it provides a clearer error message ("expected status 'needs_attention'"), so this is acceptable -- just noting the overlap.

---

**Recommendation: REQUEST_CHANGES** (posted as COMMENT since I am the PR author)

The P1 in `space-agent-tools.ts` can cause agents to receive incorrect error responses when cancelling tasks on already-terminal runs. The P2 is a defense-in-depth improvement worth addressing now while the transition enforcement is being established.

lsm (commented):

**Recommendation: APPROVE**

The two issues from the previous review have been addressed cleanly:

### 1. cancel_task guard (was P1) — Resolved

The `cancel_task` handler in `space-agent-tools.ts` (lines 379-392) now checks `canTransition()` before calling `transitionStatus()`. When the run is already terminal, the agent gets `workflowRunCancelled: false` — a graceful degradation rather than a thrown error. This is the right design choice since agents should not crash on a no-op cancellation.

### 2. updateStatus renamed to updateStatusUnchecked (was P2) — Resolved

The bypass method is now explicitly named `updateStatusUnchecked` with clear JSDoc stating it is for test fixtures only. All 7 test files were updated consistently:

- Tests that need valid transitions (e.g., `pending` -> `in_progress`, `pending` -> `cancelled`) correctly use `transitionStatus()`.
- Tests that need to set up impossible-from-pending states (e.g., directly to `completed` or `needs_attention`) correctly use `updateStatusUnchecked()`.

No production code calls `updateStatusUnchecked`.

### Minor observation (P3)

`updateRun()` remains public and accepts `status` in its params type, which is a theoretical bypass path around the state machine. However, no production caller currently passes `status` through `updateRun()` directly — the two external usages update `config` and `description` only. This is a nit-level concern; narrowing the `UpdateWorkflowRunParams` type to exclude `status` could be considered in a future cleanup but is not necessary now.

### Missing test (P3)

There is no dedicated test for the cancel_task + already-terminal-run scenario (e.g., calling `cancel_task` with `cancel_workflow_run: true` on a run that is already `completed`). The guard logic is simple and correct, but a regression test documenting this behavior would be valuable. Not blocking.

Overall, the follow-up commit is focused and correct. Both flagged issues are resolved without introducing new problems.

lsm (commented):

## Fix Review: APPROVE

The fix is minimal, targeted, and correct. Both changes replace the now-renamed `updateStatusUnchecked()` with `transitionStatus()`, which is the correct method for production status changes since the WorkflowRunStatusMachine was introduced in this PR.

### Change 1: `task-agent-tools.ts` line 685

```typescript
workflowRunRepo.transitionStatus(workflowRunId, 'completed');
```

This is gated by an upstream `run.status !== 'in_progress'` check (line 667), so the transition `in_progress → completed` is guaranteed to be valid. The `VALID_TRANSITIONS` map confirms this is a legal transition.

### Change 2: `task-agent-tools.test.ts` line 2004

```typescript
ctx.workflowRunRepo.transitionStatus(run.id, 'completed');
```

After `startRun()`, the run is in `in_progress`. This is a valid `in_progress → completed` transition per the lifecycle machine. The test correctly exercises the rejection path when `report_workflow_done` is called on an already-completed run.

### Verification

- All 69 tests in `task-agent-tools.test.ts` pass (0 failures).
- All 52 tests in `workflow-run-status-lifecycle.test.ts` and `space-workflow-run-repository.test.ts` pass (0 failures).
- No remaining `workflowRunRepo.updateStatus` calls in production code.
- `updateStatusUnchecked` usage is correctly confined to test fixture setup.

**No issues found. The PR is ready to merge.**

PR #967, opened 2026-03-26
test(daemon): online test for startup timeout error surfacing without retry

lsm:
Add `startup-timeout-no-retry.test.ts` to the `convo` online test module.

## What
- Uses `DAEMON_TEST_SPAWN=true` to spawn a fresh daemon process with `NEOKAI_SDK_STARTUP_TIMEOUT_MS=10`, so query-runner.ts reads the 10 ms constant at module load time.
- The SDK subprocess cannot respond within 10 ms, so the startup timer fires and `errorManager.handleError()` is called.

## Tests verify
1. Error is surfaced in `state.session.error` (handleError was called).
2. Error message contains actionable recovery hints (`Common causes`, `NEOKAI_SDK_STARTUP_TIMEOUT_MS`, workspace path).
3. Session returns to idle after the error.
4. Only startup-timeout errors are emitted — no second error from a retry loop (Task 2.1 removed the auto-recovery).

## Acceptance
- Passes with `NEOKAI_USE_DEV_PROXY=1` (no real API calls needed).

### DIFF
--- a/packages/daemon/tests/online/convo/startup-timeout-no-retry.test.ts
+++ b/packages/daemon/tests/online/convo/startup-timeout-no-retry.test.ts
@@ -38,29 +38,17 @@ const IDLE_TIMEOUT = IS_MOCK ? 10000 : 20000;
 const FORCED_STARTUP_TIMEOUT_MS = '10';
 
 /**
- * Poll the `state.session` RPC until the session error field is populated,
- * or return null if the timeout elapses.
+ * Read the current session error directly from the `state.session` RPC.
+ * Returns null if no error is set.
  */
-async function waitForSessionError(
+async function getSessionError(
 	daemon: DaemonServerContext,
-	sessionId: string,
-	timeout = 10000
+	sessionId: string
 ): Promise<{ message: string; details?: unknown } | null> {
-	const start = Date.now();
-	while (Date.now() - start < timeout) {
-		try {
-			const state = (await daemon.messageHub.request('state.session', {
-				sessionId,
-			})) as { error?: { message: string; details?: unknown } | null };
-			if (state.error) {
-				return state.error;
-			}
-		} catch {
-			// Session may not be ready yet; keep polling.
-		}
-		await new Promise((resolve) => setTimeout(resolve, 100));
-	}
-	return null;
+	const state = (await daemon.messageHub.request('state.session', {
+		sessionId,
+	})) as { error?: { message: string; details?: unknown } | null };
+	return state.error ?? null;
 }
 
 describe('Startup Timeout Error Surfacing', () => {
@@ -149,7 +137,9 @@ describe('Startup Timeout Error Surfacing', () => {
 				expect(finalState.status).toBe('idle');
 
 				// ── Assertion 2: error is visible in session state (handleError was called) ─
-				const sessionError = await waitForSessionError(daemon, sessionId, 3000);
+				// handleError() is called before setIdle(), so the error is already present
+				// by the time waitForIdle() returns.  A single RPC read is sufficient here.
+				const sessionError = await getSessionError(daemon, sessionId);
 				expect(sessionError).not.toBeNull();
 
 				// ── Assertion 3: error message has actionable recovery hints ──────────────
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Overall Assessment

The test approach is sound -- using `DAEMON_TEST_SPAWN=true` to spawn a child process with `NEOKAI_SDK_STARTUP_TIMEOUT_MS=10` is the correct way to test a module-level constant, and the env-var save/restore in `beforeEach`/`finally` is properly implemented. The first test (`should surface startup timeout error with actionable recovery hints`) effectively verifies the error surfacing contract. However, the second test has a significant assertion gap.

### P1: "No retry" test does not actually verify no retry occurred

**File:** `packages/daemon/tests/online/convo/startup-timeout-no-retry.test.ts`, lines 212-233

The test title claims to verify "error exactly once without retry", but the assertions do not enforce an upper bound on `errorEvents.length`. The only checks are:

1. `errorEvents.length > 0` (line 219) -- at least one error, no max
2. Every error contains `'failed to start'` (line 223) -- same category
3. `uniqueErrors.size <= 2` (line 233) -- deduplication

If a retry occurred with the same 10ms timeout, it would fail identically and produce a second error event with the exact same message string. This would pass all three assertions: length > 0, all contain "failed to start", and unique set size remains 1.

The comment at lines 228-230 acknowledges this ("A retry would also contain 'failed to start'") but then argues that the idle-state check is the real guard. That is insufficient -- reaching idle after two fast failures (10ms each) is indistinguishable from reaching idle after one failure.

**Suggested fix:** After the 300ms settling window, assert `errorEvents.length === 1` (or at most a small known count from state broadcasting, like `<= 2` with a comment explaining the broadcast fan-out). The 300ms window at 10ms timeout is more than enough for a retry to complete, so if `errorEvents.length` exceeds the expected count, a retry actually happened. This would make the "no retry" claim verifiable.

### P3: Minor -- `waitForSessionError` duplicates polling pattern

**File:** `packages/daemon/tests/online/convo/startup-timeout-no-retry.test.ts`, lines 44-64

The `waitForSessionError` function implements a polling loop that is structurally identical to `waitForProcessingState` in `daemon-actions.ts`. This is a minor duplication since it polls a different field (`error` vs `status`), but could potentially be extracted into the shared helpers if more tests need it. Not blocking.

### P3: First test calls `waitForSessionError` after `waitForIdle` with short timeout

**File:** `packages/daemon/tests/online/convo/startup-timeout-no-retry.test.ts`, line 152

The `waitForSessionError` call uses a 3000ms timeout. Since the error is set before `setIdle()` in query-runner.ts (line 420 sets error, line 436 sets idle), and `waitForIdle` already succeeded, the error should already be present. The 3000ms timeout is harmless but the polling is unnecessary -- a single `state.session` RPC call would suffice here.

### What works well

- The `DAEMON_TEST_SPAWN=true` approach correctly handles the module-level constant challenge
- Environment variable save/restore in `beforeEach`/`finally` is clean
- Subscribing to `state.session` events before sending the message avoids race conditions
- Error message content assertions (`failed to start`, `Common causes`, `NEOKAI_SDK_STARTUP_TIMEOUT_MS`) are well-chosen to verify the actionable hints
- Timeouts are appropriately scaled for mock vs real modes
- The test is in the right CI matrix module (`convo`) and will run with dev proxy

lsm (commented):

**Recommendation: APPROVE**

The fixes from the previous review round are well-executed. The test is correct, well-documented, and achieves its stated goals. Specific observations:

### P1 fix: error count bound (`<= 5`)

The fan-out analysis in the comment (lines 206-217) is accurate. I traced the production code to verify:

1. `handleError()` (query-runner.ts:420) emits `session.error` -> StateManager broadcasts `state.session` with error **[+1 event]**
2. catch-block `setIdle()` (query-runner.ts:436) emits `session.updated` -> StateManager broadcasts `state.session` (error still cached) **[+1 event]**
3. finally-block `setIdle()` (query-runner.ts:490) — same path, `isCleaningUp()` is false for timeout flow **[+1 event]**

Total: 3 error-bearing events for one timeout. A retry would double the timeout cycle to >= 6 events, which exceeds the threshold of 5. The bound is correct and would reliably detect a regression where auto-recovery is re-introduced.

The 300ms delay at line 204 for late-arriving events is a reasonable defensive measure.

### P3 fix: `getSessionError()` helper

The polling-based `waitForSessionError` has been replaced with a single `state.session` RPC call (lines 44-52). This is cleaner: by the time `waitForIdle()` returns, the error is already cached in StateManager (since `handleError` runs before `setIdle` in the catch block). The single-shot read is sufficient and deterministic.

### Other observations

- **Env var save/restore** (lines 61-81): Correctly saves and restores both `DAEMON_TEST_SPAWN` and `NEOKAI_SDK_STARTUP_TIMEOUT_MS`, handling `undefined` properly.
- **Session tracking** via `daemon.trackSession(sessionId)` ensures cleanup in `afterEach`.
- **Event subscription** correctly filters by `sessionInfo.id` to avoid cross-session noise.
- **Error content assertions** (`failed to start`, `Common causes`, `NEOKAI_SDK_STARTUP_TIMEOUT_MS`) match the actual error message built in query-runner.ts lines 407-413.
- **CI integration**: The test file is in `tests/online/convo/` which is already matrixized in CI as the `convo` module.

No issues found. Clean implementation.

PR #969, opened 2026-03-26
feat(space): replace built-in workflow transitions with gated channels

lsm:
Replace transition-based flow in built-in workflow templates with gated `WorkflowChannel` entries.

**Coding Workflow** gains 4 channels:
- `Plan → Code`: `human` gate (approval before coding)
- `Code → Verify & Test`: `always` gate (auto-advance after coding)
- `Verify & Test → Plan`: `task_result` gate on `failed`, `isCyclic: true`
- `Verify & Test → Done`: `task_result` gate on `passed`

**Research Workflow** gains 1 channel:
- `Plan Research → Research`: `always` gate

**Review-Only** unchanged (single node, no channels needed).

Channels reference node names in `from`/`to` — `resolveChannels()` resolves these via `nodeNameToAgents` at runtime, so no UUID translation is required in `seedBuiltInWorkflows()`. The seeder now passes `channels: template.channels` to `createWorkflow()`.

9 new tests added covering channel structure on templates and seeder persistence.

### DIFF
--- a/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
+++ b/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
@@ -9,9 +9,9 @@
  * - Templates use placeholder `id` / `spaceId` (empty strings) and role names
  *   as `agentId` placeholders ('planner', 'coder', 'general'). These are
  *   replaced with real SpaceAgent UUIDs by `seedBuiltInWorkflows`.
- * - Workflows are directed graphs: nodes are nodes, transitions are edges.
- *   A node with no outgoing transitions is a terminal node — the run
- *   completes when that node is reached and advance() is called.
+ * - Workflows use gated channels for inter-agent communication (agent-centric
+ *   model). Transitions are empty for agent-centric workflows; completion is
+ *   detected when all agents report done.
  * - At Space creation time, preset SpaceAgent records are seeded for each
  *   BuiltinAgentRole. `seedBuiltInWorkflows` must be called after those agents
  *   exist so that the `agentId` values resolve correctly.
@@ -25,7 +25,7 @@ import type { SpaceWorkflow } from '@neokai/shared';
 import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 
 // ---------------------------------------------------------------------------
-// Template step ID constants (used to wire up transitions)
+// Template node ID constants (used as stable IDs for nodes and startNodeId)
 // ---------------------------------------------------------------------------
 
 const CODING_PLANNER_STEP = 'tpl-coding-planner';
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

The implementation adds gated `WorkflowChannel` entries to the three built-in workflow templates and updates `seedBuiltInWorkflows()` to pass them through to `createWorkflow()`. The channel definitions are well-structured, the gate types and expressions align with the existing `WorkflowCondition` type system, and `resolveChannels()` in `space-utils.ts` correctly resolves `from`/`to` by node name (via the `nodeNameToAgents` lookup), so using node names like `'Plan'`, `'Code'`, etc. is the right approach.

All 56 tests pass, including the 9 new ones covering template structure and seeder persistence.

### Issues

**P2: Transitions not removed — "replace" is really "augment"**

The task description says "replace built-in workflows with agent-centric model" and the PR title says "replace built-in workflow transitions with gated channels," but the existing `transitions` arrays remain untouched in all three templates. The channels are added *alongside* transitions rather than replacing them. If this is intentional (e.g., a phased migration where transitions are kept for backward compatibility), it should be documented with a comment in the code or at minimum mentioned in the PR body. If the intent truly was to replace transitions, they should be removed. As-is, there's ambiguity about which routing mechanism is authoritative at runtime — both `transitions` and `channels` exist on the same workflow objects, and runtime code could use either.

This is the only substantive concern. The channel definitions themselves are correct, the type usage is sound, the seeder propagation works, and the tests are thorough.

Please clarify the transition/channel coexistence story — either remove transitions (if channels fully replace them) or add a comment/doc explaining the migration plan and which mechanism takes precedence at runtime.

lsm (commented):

**Recommendation: APPROVE**

The follow-up commit cleanly addresses the previous review feedback. The implementation correctly replaces transition-based routing with channel-based gated routing for CODING_WORKFLOW and RESEARCH_WORKFLOW.

### What looks good

1. **Transitions properly emptied.** Both `CODING_WORKFLOW.transitions` and `RESEARCH_WORKFLOW.transitions` are now `[]`, with all routing logic moved to `channels`.

2. **Channel definitions are well-structured.** Each channel has correct `from`/`to` (using node names, not IDs), `direction: 'one-way'`, appropriate `gate` types (`human`, `always`, `task_result`), and `isCyclic` where needed. This matches the `WorkflowChannel` interface in `packages/shared/src/types/space.ts`.

3. **`seedBuiltInWorkflows` passes channels through.** Line 318 (`channels: template.channels ? [...template.channels] : undefined`) correctly propagates channels to `createWorkflow`. The shallow copy of the array is acceptable here since channel objects are plain data that gets serialized into SQLite by the repository layer.

4. **JSDoc updated.** Comments accurately describe the new channel-based model.

5. **Tests are comprehensive.** All 55 tests pass. Tests verify:
   - `transitions.length === 0` for channel-based workflows
   - Channel count, `from`/`to`, `direction`, `gate.type`, `gate.expression`, and `isCyclic`
   - Export/import round-trip preserves channels, `isCyclic`, and `task_result` gates
   - REVIEW_ONLY_WORKFLOW remains unchanged (no channels, no transitions)

6. **Runtime compatibility confirmed.** `space-runtime.ts` already uses `workflow.channels ?? []` at lines 385, 773, and 1034, so the new channel data integrates without any runtime changes needed.

### Minor observations (P3)

- The transition-mapping code in `seedBuiltInWorkflows` (lines 298-304) is now dead code since all built-in templates have `transitions: []`. The `.map()` never executes. This is harmless and arguably useful if custom workflows with transitions are ever passed through this path, so not worth changing now.

PR #971, opened 2026-03-26
feat(space): wire ChannelRouter and CompletionDetector into Task Agent

lsm:
## Summary

- Add `channelRouter` and `completionDetector` as optional fields on `TaskAgentToolsConfig` (no circular deps — both types live in `runtime/`, imported as types only in `tools/`)
- `report_workflow_done` now validates all node-agent tasks are terminal via `CompletionDetector` before marking the workflow done, blocking premature completion calls
- `CompletionDetector.isComplete()` now filters to node-agent tasks only (`workflowNodeId != null`), correctly excluding the orchestration task which is `in_progress` while the check runs
- `TaskAgentManager` wires both instances in `spawnTaskAgent()` and `rehydrateTaskAgent()`

## Tests

- 5 new tests for `report_workflow_done` with `CompletionDetector` (blocks when in_progress, blocks when no tasks started, allows when all terminal, skips check without detector)
- 2 new tests for orchestration task exclusion in completion-detector
- Updated tests 2–8 in completion-detector.test.ts to include `workflowNodeId` (they represent node-agent tasks)

### DIFF
--- a/packages/daemon/src/lib/space/runtime/completion-detector.ts
+++ b/packages/daemon/src/lib/space/runtime/completion-detector.ts
@@ -53,9 +53,14 @@ export class CompletionDetector {
 		channels: WorkflowChannel[] = [],
 		nodes: WorkflowNode[] = []
 	): boolean {
-		const tasks = this.taskRepo.listByWorkflowRun(workflowRunId);
+		// Only consider node-agent tasks (those with workflowNodeId set).
+		// The orchestration task (Task Agent's own task) has workflowNodeId = null
+		// and is still in_progress while this check runs — it must be excluded.
+		const tasks = this.taskRepo
+			.listByWorkflowRun(workflowRunId)
+			.filter((t) => t.workflowNodeId != null);
 
-		// Workflow has not started yet — no tasks created
+		// Workflow has not started yet — no node tasks created
 		if (tasks.length === 0) return false;
 
 		// Any non-terminal task prevents completion
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -69,6 +69,8 @@ import type {
 import { createTaskAgentMcpServer } from '../tools/task-agent-tools';
 import { createNodeAgentMcpServer } from '../tools/node-agent-tools';
 import { ChannelResolver } from './channel-resolver';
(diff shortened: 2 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR cleanly wires `ChannelRouter` and `CompletionDetector` into the Task Agent's tool configuration and adds a pre-flight completion check to `report_workflow_done`. The implementation is correct and well-tested.

### What was reviewed

1. **`TaskAgentToolsConfig` expansion** (`packages/daemon/src/lib/space/tools/task-agent-tools.ts`): Two new optional fields (`channelRouter`, `completionDetector`) with good JSDoc. Both are type-only imports, avoiding circular dependency issues. `completionDetector` is destructured and used in `report_workflow_done`; `channelRouter` is intentionally declared but not yet consumed (placeholder for upcoming `send_message` integration).

2. **`report_workflow_done` validation gate** (lines 694-711): The completion check correctly sits between the run-status guard and the actual completion logic. It retrieves the workflow definition to pass `channels` and `nodes` for the pending-but-blocked guard. The error message is actionable, directing the agent to use `check_node_status`.

3. **`CompletionDetector.isComplete()` filter** (`packages/daemon/src/lib/space/runtime/completion-detector.ts`, line 59-61): Filtering to `workflowNodeId != null` correctly excludes the orchestration task (Task Agent's own task) which would always be `in_progress` during this check, preventing a self-blocking deadlock.

4. **`TaskAgentManager` wiring** (`packages/daemon/src/lib/space/runtime/task-agent-manager.ts`): Both `spawnTaskAgent()` and `rehydrateTaskAgent()` instantiate and pass the dependencies. The instances are lightweight (stateless wrappers around shared repos) so creating separate instances per path is fine.

5. **Tests**: 7 new tests covering the key scenarios: blocks when not all terminal, blocks when in_progress, allows on completed/needs_attention, skips check without detector, orchestration task exclusion. Existing tests updated to include `workflowNodeId` to represent node-agent tasks. All 101 tests pass. All quality checks (lint, typecheck, knip) pass.

### Minor observations (P3)

- `channelRouter` is added to `TaskAgentToolsConfig` and wired through `TaskAgentManager` but not yet used in any tool handler. This is fine as scaffolding for the next PR, but the JSDoc on the field mentions `send_message` integration that doesn't exist yet. Just noting this is intentional forward-looking code.

No issues found. Clean, focused change.

PR #973, opened 2026-03-26
test(space): backend tests for replaced built-in workflows (Task 6.2)

lsm:
## Summary

- Adds 8 new unit tests to `built-in-workflows.test.ts` for Task 6.2
- Verifies CODING_WORKFLOW channels all have `direction: 'one-way'`
- Verifies channel `from`/`to` fields reference valid node names (not UUIDs)
- Verifies all channels have gate descriptions
- Verifies RESEARCH_WORKFLOW channel direction and node name references
- Verifies seed logic preserves channel direction and from/to node names after DB persist/load
- Total: 63 tests (up from 55), all passing

### DIFF
--- a/packages/daemon/tests/unit/space/built-in-workflows.test.ts
+++ b/packages/daemon/tests/unit/space/built-in-workflows.test.ts
@@ -138,6 +138,26 @@ describe('CODING_WORKFLOW template', () => {
 		expect(CODING_WORKFLOW.transitions).toHaveLength(0);
 	});
 
+	test('all channels have direction one-way', () => {
+		for (const ch of CODING_WORKFLOW.channels!) {
+			expect(ch.direction).toBe('one-way');
+		}
+	});
+
+	test('all channel from/to fields reference valid node names', () => {
+		const nodeNames = new Set(CODING_WORKFLOW.nodes.map((n) => n.name));
+		for (const ch of CODING_WORKFLOW.channels!) {
+			expect(nodeNames.has(ch.from as string)).toBe(true);
+			expect(nodeNames.has(ch.to as string)).toBe(true);
+		}
+	});
+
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

All 63 tests pass. The 8 new tests correctly verify the channel-based routing properties (direction, from/to node name references, gate descriptions) for both the static template constants and the seed-persisted workflows. The assertions are accurate against the source definitions in `packages/daemon/src/lib/space/workflows/built-in-workflows.ts`.

### Observations

**P3 - Partial redundancy with existing tests**

Several new tests overlap with pre-existing tests in the same file:

- `all channels have direction one-way` (line 141) — the existing test at line 109 already asserts `direction: 'one-way'` on `planToCode` and `codeToVerify`. The new test adds coverage for the remaining 2 channels, which is useful but could have been achieved by adding assertions to the existing test.
- `all channel from/to fields reference valid node names` (line 147) — the existing test at line 112-134 already checks exact `from`/`to` string values for all 4 channels. Set-membership validation on top of exact-value checks is strictly weaker.
- `channel from/to references match node names` for RESEARCH_WORKFLOW (line 206) — the existing test at line 200-201 already asserts the exact `from` and `to` values.

The genuinely new coverage comes from: gate description checks on both workflows, and the seed-persistence roundtrip tests for direction and node-name routing (lines 418-478).

**P3 - Unnecessary `async` markers**

The new seed tests at lines 418, 426, and 471 are marked `async` but contain no `await` calls. This is consistent with the existing pattern in the file (many tests do the same), so not worth changing in isolation, but worth noting.

Overall the PR achieves what was asked: it verifies channel properties (direction, from/to references, gate descriptions) on both templates and seeded workflows, and confirms no transitions remain. The tests are correct and well-structured.

PR #974, opened 2026-03-26
test(e2e): global Skills Registry lifecycle (Task 5.3)

lsm:
## Summary

- Adds E2E test covering the full skill lifecycle: add via MCP server dialog, verify badge, toggle off, edit description, delete
- Fixes `Portal` to use `preact/compat` `createPortal` instead of a separate `render()` tree — stale backdrop overlays can no longer persist after parent unmount
- Bridges `skillsStore` signals into `useState` in `useSkills()` so `SkillsRegistry` never reads `signal.value` directly, preventing `@preact/preset-vite` double-render artifacts

### DIFF
--- a/packages/e2e/tests/features/global-skills.e2e.ts
+++ b/packages/e2e/tests/features/global-skills.e2e.ts
@@ -1,0 +1,219 @@
+/**
+ * Global Skills Registry E2E Tests
+ *
+ * Tests the full lifecycle of a skill in the Global Settings > Skills panel:
+ * - Navigate to Settings > Skills
+ * - Add a new MCP server skill
+ * - Verify the skill appears with correct name and type badge
+ * - Toggle the skill off and verify disabled state
+ * - Edit the description and verify it updates
+ * - Delete the skill and verify it is removed
+ *
+ * Setup: no room needed; tests the global Settings panel directly.
+ * Cleanup: delete is the final step in the test itself.
+ */
+
+import { test, expect, type Page } from '../../fixtures';
+import { waitForWebSocketConnected } from '../helpers/wait-helpers';
+
+// ─── Constants ────────────────────────────────────────────────────────────────
+
+// fetch-mcp is a pre-seeded Application MCP Server always present in the registry
+const TEST_MCP_SERVER_NAME = 'fetch-mcp';
+
+// ─── Helpers ─────────────────────────────────────────────────────────────────
+
+/**
+ * Open the Global Settings panel via the NavRail Settings button.
+ */
+async function openGlobalSettings(page: Page): Promise<void> {
+	const settingsButton = page.getByRole('button', { name: 'Settings', exact: true });
+	await settingsButton.waitFor({ state: 'visible', timeout: 5000 });
+	await settingsButton.click();
+	await expect(page.getByText('Global Settings')).toBeVisible({ timeout: 5000 });
+}
+
+/**
+ * Navigate to the Skills section inside the Global Settings panel.
+ * Assumes Global Settings is already open.
+ */
+async function navigateToSkillsSection(page: Page): Promise<void> {
+	// The Skills nav button in the settings sidebar
+	const skillsNavButton = page.locator('nav button:has-text("Skills")').first();
+	await skillsNavButton.waitFor({ state: 'visible', timeout: 5000 });
+	await skillsNavButton.click();
+
+	// Wait for the Skills content area to appear
+	await page
+		.locator('text=Application-level skills are available to any room or session')
+		.first()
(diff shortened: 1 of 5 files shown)


lsm (commented):

### P0: SkillsRegistry unit test mock is incompatible with the new `useSkills` return type

**File:** `packages/web/src/components/settings/__tests__/SkillsRegistry.test.tsx`, lines 57-63

The `useSkills` mock still returns signal-like objects with a `.value` property:

```ts
useSkills: () => ({
    skills: mockSkillsSignal,      // { value: [] }
    isLoading: mockIsLoadingSignal, // { value: false }
    error: mockErrorSignal,         // { value: null }
}),
```

But the `SkillsRegistry` component now expects plain values (arrays, booleans, strings) after the `useSkills` refactor:

```ts
const { skills: skillsList, isLoading: isLoadingVal, error: errorVal } = useSkills();
// ...
if (isLoadingVal && skillsList.length === 0) { ... }  // skillsList is { value: [] }, not an array
skillsList.map(...)  // TypeError: .map is not a function
```

This will cause every `SkillsRegistry` unit test to fail because `skillsList` is a plain object `{ value: [] }` instead of an `AppSkill[]`. The mock must be updated to return plain values matching the new `UseSkillsResult` interface:

```ts
useSkills: () => ({
    skills: mockSkillsSignal.value,
    isLoading: mockIsLoadingSignal.value,
    error: mockErrorSignal.value,
}),
```

Or better, refactor the mock variables to be plain values directly.

**CI note:** The "Web Tests" check is still pending/skipping at time of review, so this breakage hasn't been surfaced yet.

### P3: Page reload workaround in E2E test after adding a skill

**File:** `packages/e2e/tests/features/global-skills.e2e.ts`, lines 100-110

The `addMcpSkill` helper performs a full page reload after adding a skill to work around stale Portal overlays. Given that the `Portal.tsx` was refactored in this same PR to use `createPortal` (which should fix the stale overlay issue), this reload may no longer be necessary. It's not harmful but adds ~3-5s of execution time. Consider verifying if the reload can be removed now that the Portal fix is in place.

### Summary

The E2E test itself is well-structured and follows the project's E2E rules correctly — all interactions go through the UI, all assertions verify DOM state, no direct RPC calls in test actions. The `Portal.tsx` and `useSkills.ts` refactors are sound improvements. However, the existing unit test mock for `SkillsRegistry` was not updated to match the new `useSkills` return type, which will break all `SkillsRegistry` unit tests.

lsm (commented):

Re-review of the fixes addressing previous feedback. All prior issues have been properly resolved.

### Changes Reviewed

**1. `useSkills.ts` — Signal-to-useState bridge (P0 fix)**

The hook now correctly bridges `@preact/signals` values into plain `useState` via `signal.subscribe()`, returning `AppSkill[]`, `boolean`, and `string | null` instead of `Signal<T>` objects. The `subscribe()` callback fires immediately with the current value and on every subsequent change, which correctly initializes state and keeps it in sync. The cleanup in the effect's return function properly tears down all three signal subscriptions plus the LiveQuery.

**2. `SkillsRegistry.test.tsx` — Mock alignment (P0 fix)**

The mock now returns `mockSkillsSignal.value` (plain values) instead of the signal objects themselves, matching the updated `UseSkillsResult` interface. This is correct — when tests mutate `mockSkillsSignal.value` in `beforeEach`, the mock function re-evaluates `.value` on each call to `useSkills()`, so tests that set signal values before `render()` will see the expected data.

**3. `Portal.tsx` — createPortal rewrite**

The old manual `render()`/cleanup approach has been replaced with Preact's built-in `createPortal`. The `useMemo` for the container element is correct (stable reference across re-renders), and the `useEffect` handles DOM insertion and cleanup. This eliminates the root cause of stale portal overlays that required the page reload workaround.

**4. `global-skills.e2e.ts` — E2E test (P3 fix)**

The `addMcpSkill` helper now waits for the dialog heading to become hidden instead of reloading the page. The test correctly exercises the full CRUD lifecycle through the UI. The `Date.now()` suffix in the display name prevents cross-run conflicts if cleanup doesn't run (e.g., mid-test failure).

### Minor Observations (P3)

- The `getSkillRow` XPath locator comment mentions "brief double-render of the SkillsRegistry" and uses `.first()` as a guard. With the signal bridge fix now in place, this double-render should no longer occur, making `.first()` purely defensive. Not a problem, just a stale comment.
- The Portal test file (`Portal.test.tsx`, not changed in this PR) has a comment on line 43 saying "Portal returns null" which is now inaccurate since Portal returns `createPortal(...)`. Pre-existing, not blocking.

**Recommendation: APPROVE** — All P0 and P3 issues from the previous review have been correctly addressed. No new issues found.

PR #975, opened 2026-03-26
test(room): add unit tests for DB-as-source-of-truth getCurrentModel and fallback chain

lsm:
Adds `room-runtime-model-db-source.test.ts` with 6 tests using a real in-memory SQLite DB to verify the DB-as-source-of-truth contract for `getCurrentModel` and `trySwitchToFallbackModel`.

**Tests 1–4** (`getCurrentModel` DB behavior):
- Returns model/provider from DB record
- Returns null for non-existent sessions
- Defaults provider to `'anthropic'` when unset
- Reflects external DB updates (not stale in-memory value)

**Tests 5–6** (full fallback chain scenario):
- After an external DB update from model A to model B, `trySwitchToFallbackModel` computes the fallback chain using model B (from DB), not model A (stale cache)
- Direct invocation path also uses DB model after external update

### DIFF
--- a/packages/daemon/tests/unit/room/room-runtime-model-db-source.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-model-db-source.test.ts
@@ -1,0 +1,379 @@
+/**
+ * Tests verifying that getCurrentModel() reads from the DB (source of truth)
+ * and that trySwitchToFallbackModel() uses the DB model when computing the fallback chain.
+ *
+ * These tests use a real in-memory SQLite DB (not mocks) to verify the DB-read behavior.
+ */
+
+import { describe, expect, it, afterEach } from 'bun:test';
+import { Database as BunDatabase } from 'bun:sqlite';
+import { SessionRepository } from '../../../src/storage/repositories/session-repository';
+import { RoomRuntimeService } from '../../../src/lib/room/runtime/room-runtime-service';
+import type { RoomRuntimeServiceConfig } from '../../../src/lib/room/runtime/room-runtime-service';
+import {
+	createRuntimeTestContext,
+	createGoalAndTask,
+	type RuntimeTestContext,
+} from './room-runtime-test-helpers';
+import type { GlobalSettings, Session } from '@neokai/shared';
+
+// Minimal sessions table schema required by SessionRepository
+const SESSIONS_SCHEMA = `
+  CREATE TABLE IF NOT EXISTS sessions (
+    id TEXT PRIMARY KEY,
+    title TEXT NOT NULL,
+    workspace_path TEXT NOT NULL,
+    created_at TEXT NOT NULL,
+    last_active_at TEXT NOT NULL,
+    status TEXT NOT NULL,
+    config TEXT NOT NULL,
+    metadata TEXT NOT NULL,
+    is_worktree INTEGER DEFAULT 0,
+    worktree_path TEXT,
+    main_repo_path TEXT,
+    worktree_branch TEXT,
+    git_branch TEXT,
+    sdk_session_id TEXT,
+    available_commands TEXT,
+    processing_state TEXT,
+    archived_at TEXT,
+    parent_id TEXT,
+    type TEXT DEFAULT 'worker',
+    session_context TEXT
+  )
+`;
+
+function makeSession(id: string, model: string, provider?: string): Session {
+	return {
+		id,
+		title: 'Test Session',
+		workspacePath: '/workspace',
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Overview

The test file adds 6 tests verifying that `getCurrentModel` reads from SQLite (not in-memory cache) and that `trySwitchToFallbackModel` uses the DB model when computing the fallback chain. All 6 tests pass. The core DB-read behavior is correctly verified.

---

### P1 — Should Fix

**`isProviderAvailable` is never mocked in tests 5-6, silently producing incorrect behavior.**

In test 6, `createRuntimeTestContext` is called without `isProviderAvailable`, so `this.isProviderAvailable` is `undefined` on the runtime. When `trySwitchToFallbackModel` executes the availability check:

```typescript
if (this.isProviderAvailable) {
    const available = await this.isProviderAvailable(candidate.provider, candidate.model);
    if (!available) { ...continue; }
}
```

The `if (this.isProviderAvailable)` check passes (the property exists on `this`, it's `undefined`). The optional call `?.()` short-circuits and returns `undefined` since the property value is `undefined`. Since `!undefined === true`, every candidate is skipped — the fallback chain silently does nothing. The test still passes only because the mock's `switchModelImpl` always returns success and the test asserts on `switchCalls[0].args[1]` being `'correct-fallback'`, which happens to be chain[0] — but NOT because the availability check correctly filters candidates.

The same issue affects test 5's full scenario path.

**Fix:** Pass `isProviderAvailable: async () => true` to `createRuntimeTestContext` in both tests, and add an explicit test case that verifies the skip-on-unavailable path with a real/unmocked `isProviderAvailable`.

---

**1. `rawDb` is never closed in tests 5-6 (resource leak).**

Tests 5-6 create `const rawDb = new BunDatabase(':memory:')` and use it for `sessionRepo`, but the `afterEach` only closes `ctx.db`, not `rawDb`. The `rawDb` instances leak.

**Fix:** Add `rawDb.close()` to the `afterEach` or close it inline after the test assertions.

**2. Test 5 has a conditional code path that changes what is being tested.**

At line 277, if `groups2` is falsy (no new group spawned), the test falls back to directly calling `trySwitchToFallbackModel` — a completely different scenario from the "full scenario: external DB update triggers fallback chain" described in the test name and PR body. The assertion is also different between branches: the direct path checks `switchCalls.length >= 1` while the full path checks `>= 1` with a specific `lastSwitch` value.

**Fix:** Separate into two explicit tests — one for the full event-driven scenario, one for the direct invocation path. This eliminates the ambiguous `if/else` that makes it unclear which code path was actually exercised.

**3. Tests 5-6 do not exercise the real `RoomRuntimeService.createSessionFactory().getCurrentModel` path.**

Tests 5-6 use `createRuntimeTestContext` with a `getCurrentModelImpl` override, which sets `sessionFactory.getCurrentModelImpl` and makes the mock's `getCurrentModel` return the override. This is a different code path from the actual `RoomRuntimeService.createSessionFactory().getCurrentModel` (which calls `ctx.db.getSession`). Both paths read from the same SQLite DB (`sessionRepo`'s `rawDb`), so the end result is correct, but the test does not cover the real `createSessionFactory` implementation.

The coverage gap: if `trySwitchToFallbackModel` were accidentally refactored to read from `agentSessions` instead of calling `sessionFactory.getCurrentModel`, tests 5-6 would not catch it (they'd still pass via the mock). Tests 1-4 DO exercise the real `createSessionFactory().getCurrentModel` path directly.

---

### P3 — Nits

**`as unknown as` cast for accessing private method `createSessionFactory` (lines 96, 117, 142, 167):** This is a standard and acceptable pattern for accessing private methods in unit tests. No change needed.

---

### What Works Well

- Tests 1-4 correctly verify the real SQLite DB-read behavior of `getCurrentModel` using a direct cast to access `createSessionFactory`. The test setup with `buildServiceConfig` properly wires `db.getSession` to `sessionRepo.getSession`.
- Tests 5-6 verify the end-to-end fallback chain scenario with an external DB update. The mock session factory correctly simulates the DB-read contract.
- The PR description accurately describes what was implemented and the test structure is clean and well-organized.

---

### Summary

All 6 tests pass and correctly verify the DB-as-source-of-truth contract. The main concern is that the `isProviderAvailable` code path in `trySwitchToFallbackModel` is effectively untested in tests 5-6 due to an unmocked callback producing silent skip behavior — this is a real coverage gap. The `rawDb` resource leak and the conditional branching in test 5 are secondary concerns. With the P1 fixed, this would be a solid test suite.

PR #978, opened 2026-03-26
feat(space): drop session group tables and remove currentNodeId (Task 8.2)

lsm:
Implements Task 8.2: removes the session group abstraction from the Space system.

## Changes

- **Migration M59**: drops `space_session_groups` and `space_session_group_members` tables; rebuilds `space_workflow_runs` without `current_node_id`
- **Shared types**: removes `SpaceSessionGroup`, `SpaceSessionGroupMember` interfaces and `currentNodeId` from `SpaceWorkflowRun`
- **Backend**: removes `SpaceSessionGroupRepository`, session group RPC handlers, all `sessionGroupRepo`/`getGroupId` usage throughout runtime and tools
- **Peer lookup**: replaced session-group-based peer discovery with direct `space_tasks` queries using `taskAgentSessionId` and `agentName`
- **`processRunTick()`**: now processes ALL run tasks (not just "current step" tasks); preserves backward-compat: single-task steps only emit `task_needs_attention`, not `workflow_run_needs_attention`
- **Frontend**: removes session group signals and event handlers from `space-store.ts` and `SpaceTaskPane.tsx`
- **Tests**: deletes 5 session-group test files; updates 20+ test files to use the task-based model; adds `migration-59_test.ts`


lsm (commented):

### Summary

Large, well-executed removal of the session group abstraction. The migration from session-group-based peer discovery to direct `space_tasks` queries is clean, and the 50+ file changes are internally consistent. Types, RPC handlers, repository, DaemonHub events, frontend signals, and tests are all properly removed/updated.

One blocking issue with the migration, and a few minor items.

---

### P0: Migration 59 table rebuild does not disable foreign keys

**File:** `packages/daemon/src/storage/schema/migrations.ts`, lines 3775-3832

Migration 59 rebuilds `space_workflow_runs` (DROP old table, RENAME new table) without first running `PRAGMA foreign_keys = OFF`. The DB connection has foreign keys enabled (set at `database-core.ts:56`).

When `DROP TABLE space_workflow_runs` executes at line 3821, SQLite fires the `ON DELETE SET NULL` action on `space_tasks.workflow_run_id` (defined at line 3557), **nullifying every existing task's workflow_run_id**. The data is then in the renamed table, but the damage to `space_tasks` is already done.

Every other table rebuild migration in this file (M29 session tables, M44, M45, M51, M55, etc.) follows the pattern:
```
db.exec('PRAGMA foreign_keys = OFF');
db.exec('BEGIN');
try { ... COMMIT } catch { ROLLBACK } finally { db.exec('PRAGMA foreign_keys = ON'); }
```

Migration 59 should do the same. Without this fix, any user who has active workflow runs with tasks will lose the `workflow_run_id` link on their `space_tasks` rows during the upgrade.

---

### P3: Stale comment in test file

**File:** `packages/daemon/tests/unit/space/task-agent-manager.test.ts`, line 965

Comment still references `taskGroupIds`:
```
// meaning taskGroupIds will not have an entry for this task.
```
This map was removed. The comment should be updated.

### P3: Unused variable in migration test

**File:** `packages/daemon/tests/unit/storage/migrations/migration-59_test.ts`, lines 92-106

`requiredCols` is declared but then suppressed with `void requiredCols`. This dead code should be removed or the assertion loop should iterate over `requiredCols` directly instead of a hardcoded subset.

### P3: `_spaceId` parameter kept but unused

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 607

`createSubSessionFactory(taskId: string, _spaceId: string)` marks spaceId as unused. If the parameter is no longer needed, consider removing it and updating the two call sites, or add a brief comment explaining why it is retained for future use.

---

### What looks good

- Migration 59 is idempotent (guarded by table/column existence checks) and has dedicated tests for idempotency, data preservation, table/column removal.
- The node-agent-tools replacement properly routes peer discovery through `space_tasks` queries instead of session group lookups -- cleaner and fewer moving parts.
- DaemonHub event types for session groups are cleanly removed, and the frontend space-store correctly drops the associated signal/subscription code.
- WorkflowExecutor is simplified to a read-only graph navigator, consistent with the agent-centric model where agents drive progression via send_message/report_done.
- The `processRunTick()` change to process ALL run tasks (not just current-step tasks) is correct for the new model where there is no single "current step."
- Comprehensive test updates: 5 session-group test files deleted, 20+ test files updated, new migration-59 test added.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been addressed correctly:

1. **P0 (Migration FK safety)**: `runMigration59()` now wraps the entire table rebuild in `PRAGMA foreign_keys = OFF` / `BEGIN` / `COMMIT` with proper `ROLLBACK` on error and `finally` to re-enable FKs. This matches the established pattern used by M44, M45, M51, M55, and prevents `ON DELETE SET NULL` from nullifying `space_tasks.workflow_run_id` during the table drop+rename.

2. **P3 (Stale comment)**: The comment referencing `taskGroupIds` at line 964 of `task-agent-manager.test.ts` has been simplified to remove the group reference.

3. **P3 (Dead code)**: The `requiredCols` variable and its `void requiredCols` suppression have been removed from `migration-59_test.ts`. The test now iterates the column names inline.

4. **P3 (Unused parameter)**: `_spaceId` has been removed from `createSubSessionFactory()`, and both call sites updated accordingly.

The migration structure is clean: idempotent guards are preserved inside the transaction, the try/catch/finally ensures FK state is always restored, and the test suite covers table drops, column removal, data preservation, and idempotency.

No remaining issues found.

PR #982, opened 2026-03-26
ci: add run_e2e_only input for selective E2E on PR branches

lsm:
## Summary

- Add `run_e2e_only` boolean input to `workflow_dispatch` trigger in `main.yml`
- When set to `true`, skips check/unit/online/web/CLI jobs and runs only `discover + build + e2e`
- Previously, `workflow_dispatch` triggered all CI jobs. If any failed, the `build` job was blocked and E2E never ran (confirmed empirically)

## Usage

```bash
gh workflow run main.yml --repo lsm/neokai --ref <pr-branch> -f run_e2e_only=true
```

## Changes

- **Input definition**: Add `run_e2e_only` boolean input (default `false`)
- **Prerequisite jobs** (7 jobs): Skip when `run_e2e_only == 'true'` — uses explicit string comparison because GitHub Actions boolean inputs arrive as strings
- **Build job**: Bypass prerequisite gate when `run_e2e_only == 'true'`; added `discover` to `needs` array so it's always tracked
- **Coverage jobs** (`coveralls-finalize`, `coverage-gate`): Skip when `run_e2e_only == 'true'`

## Test plan

1. Triggered workflow_dispatch on a PR branch with `run_e2e_only=true` — verified `discover + build + e2e` all executed
2. Addressed review feedback: fixed string comparison for boolean input coercion bug

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,7 +61,7 @@ jobs:
     name: Lint, Knip, Format & Type Check
     runs-on: ubuntu-latest
     timeout-minutes: 3
-    if: (github.event_name != 'push' || github.ref != 'refs/heads/dev') && !github.event.inputs.run_e2e_only
+    if: (github.event_name != 'push' || github.ref != 'refs/heads/dev') && github.event.inputs.run_e2e_only != 'true'
 
     steps:
       - uses: actions/checkout@v4
@@ -120,7 +120,7 @@ jobs:
   test-daemon-online:
     name: Daemon Online (${{ matrix.module }})
     runs-on: ubuntu-latest
-    if: github.ref_type != 'tag' && (github.event_name != 'push' || github.ref != 'refs/heads/dev') && !github.event.inputs.run_e2e_only
+    if: github.ref_type != 'tag' && (github.event_name != 'push' || github.ref != 'refs/heads/dev') && github.event.inputs.run_e2e_only != 'true'
     timeout-minutes: ${{ matrix.timeout || 5 }}
 
     strategy:
@@ -377,7 +377,7 @@ jobs:
     name: Daemon + Shared Unit Tests (Coverage)
     runs-on: ubuntu-latest
     timeout-minutes: 6
-    if: (github.event_name != 'push' || github.ref != 'refs/heads/dev') && !github.event.inputs.run_e2e_only
+    if: (github.event_name != 'push' || github.ref != 'refs/heads/dev') && github.event.inputs.run_e2e_only != 'true'
 
     steps:
       - uses: actions/checkout@v4
@@ -430,7 +430,7 @@ jobs:
     name: Web Tests
     runs-on: ubuntu-latest
     timeout-minutes: 3
-    if: (github.event_name == 'pull_request' && github.base_ref == 'main' || github.event_name == 'push' && github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch') && !github.event.inputs.run_e2e_only
+    if: (github.event_name == 'pull_request' && github.base_ref == 'main' || github.event_name == 'push' && github.ref == 'refs/heads/main' || github.event_name == 'workflow_dispatch') && github.event.inputs.run_e2e_only != 'true'
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


lsm (commented):

---

### P0: Boolean input string coercion bug breaks default workflow_dispatch behavior

The expression `!github.event.inputs.run_e2e_only` is used in 7 job conditions and the build gate, but it does not work correctly due to a well-known GitHub Actions gotcha: `workflow_dispatch` boolean inputs arrive as **strings** `"true"` or `"false"`, not actual booleans. Since the string `"false"` is non-empty, it is **truthy** in GitHub Actions expressions. Therefore `!"false"` evaluates to `false`.

The input has `default: 'false'`, so it will always be present as the string `"false"` when not explicitly set. This means:

**When `run_e2e_only` is not set (or explicitly `false`) -- the default case:**
- All 7 prerequisite jobs (`check`, `test-daemon-online`, `test-daemon-shared-unit`, `test-web`, `test-cli`, `coveralls-finalize`, `coverage-gate`) are incorrectly **skipped** because `!"false"` = `false` makes their `if` conditions evaluate to `false`.
- The `build` job's gate condition `(github.event.inputs.run_e2e_only || needs.discover.result == 'success' && ...)` short-circuits to `"false"` (truthy string), bypassing ALL prerequisite result checks.

This is a **regression**: previously, `workflow_dispatch` without any inputs ran all tests (check, unit, online, web, CLI, coverage, e2e). Now it would skip all prerequisites and only run discover + build + e2e.

The empirical verification ("two CI runs confirmed it works") only tested `run_e2e_only=true`, which is the one case that happens to work correctly.

**Fix:** Replace all `!github.event.inputs.run_e2e_only` with `github.event.inputs.run_e2e_only != 'true'` (7 occurrences at lines 64, 123, 380, 433, 487, 526, 547), and in the build gate (line 638) replace `github.event.inputs.run_e2e_only` with `github.event.inputs.run_e2e_only == 'true'`.

Reference: [GitHub Issue #1483 - Boolean inputs are not actually booleans](https://github.com/actions/runner/issues/1483), [Stack Overflow - Strange boolean evaluation](https://stackoverflow.com/questions/68928595/strange-boolean-evaluation-in-github-actions)

**Affected locations in `.github/workflows/main.yml`:**
- Line 64: `check` job `if` condition
- Line 123: `test-daemon-online` job `if` condition
- Line 380: `test-daemon-shared-unit` job `if` condition
- Line 433: `test-web` job `if` condition
- Line 487: `test-cli` job `if` condition
- Line 526: `coveralls-finalize` job `if` condition
- Line 547: `coverage-gate` job `if` condition
- Line 638: `build` job `if` condition (gate bypass branch)

### P2: Missing parentheses in `build` job condition for clarity

Lines 638-643: The build gate condition relies on `&&` binding tighter than `||`, but this is not visually obvious in a multi-line YAML expression:

```yaml
(github.event.inputs.run_e2e_only ||
 needs.discover.result == 'success' && contains(fromJSON('["success", "skipped"]'), needs.check.result) &&
 contains(fromJSON('["success", "skipped"]'), needs.test-daemon-online.result) &&
 ...)
```

This evaluates as `run_e2e_only || (discover && check && online && unit && web && cli)`, which is the intended logic. Adding explicit parentheses around the second branch would prevent future misreads. Minor clarity issue, not a functional bug.

lsm (commented):

**Recommendation: APPROVE**

All previously identified issues have been properly addressed.

### P0 fix verification (string comparison for boolean input)

Confirmed that **zero** instances of the bare negation pattern `!github.event.inputs.run_e2e_only` remain in the file. All 7 prerequisite skip conditions now use explicit `!= 'true'` string comparison:

| Job | Line | Pattern |
|-----|------|---------|
| `check` | 64 | `... && github.event.inputs.run_e2e_only != 'true'` |
| `test-daemon-online` | 123 | `... && github.event.inputs.run_e2e_only != 'true'` |
| `test-daemon-shared-unit` | 380 | `... && github.event.inputs.run_e2e_only != 'true'` |
| `test-web` | 433 | `... && github.event.inputs.run_e2e_only != 'true'` |
| `test-cli` | 487 | `... && github.event.inputs.run_e2e_only != 'true'` |
| `coveralls-finalize` | 526 | `... && github.event.inputs.run_e2e_only != 'true'` |
| `coverage-gate` | 547 | `... && github.event.inputs.run_e2e_only != 'true'` |

The build gate at line 638 correctly uses `== 'true'` with explicit parentheses separating the two branches: the E2E-only bypass vs. the prerequisite result checks.

### Correctness under non-workflow_dispatch triggers

When `github.event.inputs.run_e2e_only` is `''` (undefined, e.g. on `pull_request` or `push` triggers), comparing `'' != 'true'` yields `true`, so prerequisite jobs run normally. In the build gate, `'' == 'true'` yields `false`, so the prerequisite checks apply. No regressions for existing trigger paths.

### Additional changes noted

The `discover` job was added to the build's `needs` array (line 632) and `needs.discover.result == 'success'` is checked (line 639). This is a minor behavioral change -- the build now waits for `discover` in all paths, not just E2E-only -- but it is a sensible improvement since `discover` must succeed for E2E to run at all.

### YAML validity

Confirmed valid via `yaml.safe_load`. No syntax errors.

No remaining issues.

PR #987, opened 2026-03-26
plan: unified state management — EntityStore architecture and dual-channel bug fix

akz142857:
## Summary

- Adds `docs/plans/bug-entitystore.md` with a 6-task plan to eliminate the dual-channel state update bug and introduce `EntityStore<T>` architecture
- Root cause: `useTaskViewData` listens to `room.task.update` events but the Runtime autonomous task-start path only triggers LiveQuery (`notifyChange('tasks')`), never emitting `room.task.update` — so the detail view goes stale
- Plan is structured in three phases: P0 bug fix (Task 1), EntityStore implementation (Tasks 2–4), server-side cleanup (Task 5), and integration verification (Task 6)

## Test plan

- [ ] Review the plan for correctness against the root cause analysis
- [ ] Confirm Task 1 (P0 fix) is self-contained and safe to implement first
- [ ] Confirm Task 5 (server cleanup) correctly identifies all `emitTaskUpdate` call sites

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -34,7 +34,7 @@
       "name": "@neokai/daemon",
       "version": "0.8.0",
       "dependencies": {
-        "@anthropic-ai/claude-agent-sdk": "0.2.84",
+        "@anthropic-ai/claude-agent-sdk": "0.2.81",
         "@github/copilot-sdk": "0.2.0",
         "@neokai/shared": "workspace:*",
         "croner": "10.0.1",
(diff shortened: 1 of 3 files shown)


akz142857 (commented):
## Plan Review

### Summary
The plan is well-structured with clear task decomposition and a logical dependency chain. The root cause analysis is accurate. No P0/P1 blockers — a couple of P2 concerns and minor nits.

---

### P2 — Inbox Approve fix is missing from the plan

The planner's PR description references **"Inbox Approve 修复"** in the goal title, but the plan document has no task for it. Either:
- This fix was intentionally scoped out (in which case the plan doc should say so explicitly), or
- It was accidentally dropped and needs a Task 7

Please clarify and update the plan accordingly.

---

### P2 — `as NeoTask[]` cast in Task 1 risks silent runtime errors

Task 1, Subtask 2 writes:
```typescript
const task = computed(() =>
  (roomStore.tasks.value as NeoTask[]).find((t) => t.id === taskId) ?? null
);
```

`roomStore.tasks` is typed `signal<TaskSummary[]>`. If `TaskSummary` is a subset of `NeoTask`, casting with `as` suppresses the type error but any code downstream that reads `NeoTask`-only fields will get `undefined` at runtime. The task should explicitly state: verify the `TaskSummary` vs `NeoTask` types match, or update the hook's return type to use `TaskSummary` if that's sufficient.

---

### P3 — Use `useComputed` instead of `computed` inside a hook

Task 1, Subtask 2 proposes using `computed()` at hook call site. In Preact Signals, bare `computed()` inside a hook body creates a new computed signal on every render without cleanup. The correct API in a hook is `useComputed(() => ...)` from `@preact/signals`. Please update the code example.

---

### P3 — Task 5 `emitRoomOverview` audit is vague

> Determine if any consumer still needs room.overview for task data — if not, remove the calls...

This defers too much to the implementer at execution time. The plan should enumerate the specific call sites in `task-handlers.ts` to remove vs. keep, based on the codebase exploration the planner already did.

akz142857 (commented):
## Follow-up Review (Iteration 2)

All four issues from the first review are addressed. One new P2 inconsistency introduced by the fixes:

### P2 — Task 1 and Task 3 use conflicting types for roomStore.tasks

**Task 1, Subtask 1** fixes the type annotation to `signal<NeoTask[]>` (correct — the SQL query returns full NeoTask fields).

**Task 3, Subtask 1b** then replaces that signal with `readonly taskStore = new EntityStore<TaskSummary>()`, and the computed getter returns `TaskSummary[]` — silently downgrading the type back to the narrower type.

These two tasks are contradictory. Fix: Task 3 must use `EntityStore<NeoTask>` (not `EntityStore<TaskSummary>`), and the computed getter should return `NeoTask[]`:
```typescript
readonly taskStore = new EntityStore<NeoTask>();
readonly tasks = computed(() => this.taskStore.toArray()); // returns NeoTask[]
```

Update Task 3 step 1b and 1d accordingly, and update the downstream computed signals (`pendingTasks`, `activeTasks`, etc.) to use `NeoTask` if they were typed against `TaskSummary`.

PR #990, opened 2026-03-26
fix(web): useTaskViewData reads task reactively from roomStore.tasks signal

akz142857:
Replace the task.get RPC call + room.task.update event listener in
useTaskViewData with a useComputed signal derived from roomStore.tasks.
This eliminates the dual-channel update bug where Runtime-initiated
task transitions (pending → in_progress) only trigger LiveQuery deltas
but not the emitTaskUpdate event, causing the task detail view to go
stale.

Changes:
- roomStore.tasks typed as signal<NeoTask[]> (was signal<TaskSummary[]>)
- useTaskViewData: task derived via useComputed from roomStore.tasks
- useTaskViewData: removed task.get RPC call and room.task.update listener
- useTaskViewData: isLoading now scoped to group fetch only
- Tests updated to mock roomStore.tasks as signal<NeoTask[]> and assert
  task is derived reactively rather than fetched via RPC

### DIFF
--- a/packages/web/src/components/room/TaskView.test.tsx
+++ b/packages/web/src/components/room/TaskView.test.tsx
@@ -15,6 +15,8 @@ import { useEffect } from 'preact/hooks';
 import { signal } from '@preact/signals';
 // Static import gives access to vi.mocked(useAutoScroll) for call assertions
 import { useAutoScroll } from '../../hooks/useAutoScroll.ts';
+import type { NeoTask } from '@neokai/shared';
+import { roomStore } from '../../lib/room-store.ts';
 
 // -------------------------------------------------------
 // Mocks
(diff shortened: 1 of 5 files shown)


akz142857 (commented):
Code review summary:

**Correctness** ✓
- `useTaskViewData` now derives `task` via `useComputed(() => roomStore.tasks.value.find(...))` — reactive to all LiveQuery deltas
- `room.task.update` listener removed; `session.updated` listener preserved
- `load()` simplified to group fetch only; `isLoading` scoped correctly
- `roomStore.tasks` correctly typed as `signal<NeoTask[]>`

**Acceptance criteria** ✓
- Pending → in_progress transitions from Runtime will reflect immediately (LiveQuery feeds the signal)
- Task list and detail always consistent (same signal source)
- No `as NeoTask[]` casts in the hook (casts in store are pre-existing, not in hook scope)
- Tests assert reactive derivation and no `room.task.update` subscription

**Tests** ✓
- New tests for reactive update, null when task absent, no room.task.update subscription
- TaskView tests updated to set `roomStore.tasks.value` instead of mocking `task.get`

**Formatting** ✓ (fixed in follow-up commits)

LGTM, no issues.

PR #1002, opened 2026-03-26
fix: restore scroll position synchronously after loading older messages

lsm:
## Summary
- Fix "Load earlier messages" scroll position being overridden by auto-scroll-to-bottom
- Root cause: scroll restoration used `useEffect` + `requestAnimationFrame`, allowing `useAutoScroll` (also `useEffect`) to race and scroll to bottom before restoration completed
- Fix: changed to `useLayoutEffect` (synchronous, before paint) so scroll position is locked in before any auto-scroll effect fires

## Changes
- `TaskConversationRenderer.tsx`: `useEffect` + rAF → `useLayoutEffect` (synchronous scroll restore)
- `useAutoScroll.ts`: defense-in-depth guard to skip message-count delta when `loadingOlder` transitions `true → false`
- `task-message-pagination.e2e.ts`: add rAF wait before reading scroll position for deterministic assertion
- `useAutoScroll.test.ts`: add test for `loadingOlder` transition suppressing auto-scroll

### DIFF
--- a/packages/e2e/tests/features/task-message-pagination.e2e.ts
+++ b/packages/e2e/tests/features/task-message-pagination.e2e.ts
@@ -226,9 +226,7 @@ test.describe('TaskView — Message Pagination', () => {
 			// Verify newest messages are still present (they should not have disappeared)
 			await expect(page.getByText('Message number 55', { exact: true })).toBeVisible();
 
-			// Wait for the requestAnimationFrame-based scroll restoration to complete.
-			// The component restores scrollTop inside rAF, so we must let one frame pass
-			// before reading the scroll position; otherwise we capture the pre-restore value.
+			// Wait for any pending microtask/flush to settle before reading scroll state.
 			await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
 
 			// Verify scroll position was preserved (not jumped to top).
--- a/packages/e2e/tests/features/task-message-pagination.e2e.ts
+++ b/packages/e2e/tests/features/task-message-pagination.e2e.ts
@@ -226,6 +226,11 @@ test.describe('TaskView — Message Pagination', () => {
 			// Verify newest messages are still present (they should not have disappeared)
 			await expect(page.getByText('Message number 55', { exact: true })).toBeVisible();
 
+			// Wait for the requestAnimationFrame-based scroll restoration to complete.
+			// The component restores scrollTop inside rAF, so we must let one frame pass
+			// before reading the scroll position; otherwise we capture the pre-restore value.
+			await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
+
 			// Verify scroll position was preserved (not jumped to top).
 			// Only assert when scrollBefore > 0 — if the container was too short to
 			// overflow before clicking, scrollTop stays 0 and there is nothing to preserve.
--- a/packages/web/src/components/room/TaskConversationRenderer.tsx
+++ b/packages/web/src/components/room/TaskConversationRenderer.tsx
@@ -8,7 +8,7 @@
  * which agent produced it. Role transitions show a small divider label.
  */
 
-import { useCallback, useEffect, useMemo, useRef, useState } from 'preact/hooks';
+import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'preact/hooks';
 import type { RefObject } from 'preact';
 import type { SDKMessage } from '@neokai/shared/sdk/sdk.d.ts';
 import type { SessionInfo } from '@neokai/shared';
@@ -112,9 +112,11 @@ export function TaskConversationRenderer({
 
 	// After rawMessages changes (triggered by loadEarlier), restore scroll position
 	// so older prepended messages don't cause a jarring jump to the top.
+	// Uses useLayoutEffect to restore scroll synchronously before paint, preventing
+	// useEffect-based auto-scroll (in the parent) from racing and overriding the position.
 	// IMPORTANT: onLoadingOlderChange?.(false) must be called on ALL exit paths so
 	// isLoadingOlder never gets stuck true (which would permanently disable auto-scroll).
-	useEffect(() => {
+	useLayoutEffect(() => {
 		if (!scrollRestoreRef.current) return;
 		if (!scrollContainerRef?.current) {
 			// Container unmounted between handleLoadEarlier and this effect — clear the flag.
@@ -124,16 +126,9 @@ export function TaskConversationRenderer({
 		}
 		const { scrollHeight, scrollTop } = scrollRestoreRef.current;
 		scrollRestoreRef.current = null;
-		requestAnimationFrame(() => {
-			if (!scrollContainerRef.current) {
-				// Container unmounted between rAF scheduling and rAF execution — clear the flag.
-				onLoadingOlderChange?.(false);
-				return;
-			}
-			const newScrollHeight = scrollContainerRef.current.scrollHeight;
-			scrollContainerRef.current.scrollTop = scrollTop + (newScrollHeight - scrollHeight);
-			onLoadingOlderChange?.(false);
-		});
+		const newScrollHeight = scrollContainerRef.current.scrollHeight;
+		scrollContainerRef.current.scrollTop = scrollTop + (newScrollHeight - scrollHeight);
+		onLoadingOlderChange?.(false);
 	}, [rawMessages, scrollContainerRef, onLoadingOlderChange]);
 
 	const messages = useMemo(
--- a/packages/web/src/hooks/__tests__/useAutoScroll.test.ts
+++ b/packages/web/src/hooks/__tests__/useAutoScroll.test.ts
@@ -269,6 +269,37 @@ describe('useAutoScroll', () => {
 
 			expect(scrollIntoViewMock).not.toHaveBeenCalled();
 		});
+
+		it('should not auto-scroll when loadingOlder transitions from true to false', () => {
+			const { containerRef, endRef, scrollIntoViewMock } = createMockRefs();
+
+			const { rerender } = renderHook(
+				({ messageCount, loadingOlder }) =>
+					useAutoScroll({
+						containerRef,
+						endRef,
+						enabled: true,
+						messageCount,
+						isInitialLoad: false,
+						loadingOlder,
+					}),
+				{
+					initialProps: { messageCount: 50, loadingOlder: false },
+				}
+			);
+
+			scrollIntoViewMock.mockClear();
+
+			// Start loading older — message count increases as hidden messages are revealed
+			rerender({ messageCount: 55, loadingOlder: true });
+			expect(scrollIntoViewMock).not.toHaveBeenCalled();
+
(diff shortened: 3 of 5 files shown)


lsm (commented):

### Overview

The PR fixes a scroll-jump bug when clicking "Load Earlier" in the task message pagination UI. The root cause was a race between the scroll-restoration effect (previously `useEffect` + `requestAnimationFrame`) and the auto-scroll effect (also `useEffect`-based). The fix has two parts:

1. **`TaskConversationRenderer.tsx`**: Changed `useEffect` + `requestAnimationFrame` to `useLayoutEffect` (synchronous, before paint)
2. **`useAutoScroll.ts`**: Added a guard effect that catches the `loadingOlder: true → false` transition and syncs `prevMessageCountRef` to prevent auto-scroll from misinterpreting revealed older messages as new content

Both changes are correct and well-reasoned. The `useLayoutEffect` ensures scroll position is locked in before any `useEffect`-based auto-scroll can override it, eliminating the race entirely. The guard in `useAutoScroll` provides defense-in-depth in case the `loadingOlder` flag is ever out of sync.

---

### P0 Issues: None

---

### P1 Issues: None

---

### P2 Issues

**1. E2E test comment is stale but functionally harmless**

In `packages/e2e/tests/features/task-message-pagination.e2e.ts` (lines 229-232):

```typescript
// Wait for the requestAnimationFrame-based scroll restoration to complete.
// The component restores scrollTop inside rAF, so we must let one frame pass
// before reading the scroll position; otherwise we capture the pre-restore value.
await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
```

The comment references the old `requestAnimationFrame`-based approach. After the `useLayoutEffect` change, scroll is restored **before** paint, not inside `rAF`. The `page.evaluate(() => new Promise((r) => requestAnimationFrame(r)))` wait is now unnecessary but still works because browsers fire rAF before the next paint (and jsdom may resolve rAF synchronously anyway).

**Why this matters**: If a future PR reverts the `useLayoutEffect` back to `useEffect`+rAF, this test will correctly validate the behavior again. The rAF wait is not wrong -- it is just no longer needed for the current implementation. Consider either:
- Updating the comment to reflect the synchronous approach, or
- Adding a note that this wait is now belt-and-suspenders

This is a P2 rather than P3 because the misleading comment could confuse future maintainers about whether the rAF is load-bearing.

---

### P3 Issues

**1. New unit test does not verify no auto-scroll on initial mount**

The new test at `packages/web/src/hooks/__tests__/useAutoScroll.test.ts` (line 273) starts with `initialProps: { messageCount: 50, loadingOlder: false }`. On initial hook mount, `prevMessageCountRef.current = 0` (the ref's initial value). The auto-scroll effect would run with `hasNewContent = 50 > 0 = true` and call `scrollToBottom()`. The test never asserts this (no `expect(scrollIntoViewMock).toHaveBeenCalled()` after the initial render).

This is not a regression -- the existing test suite covers initial-mount auto-scroll separately. But it means this specific test does not stand alone as a complete verification of the guard behavior; it relies on the hook's initial ref state being what it is. A more explicit test would start with `messageCount: 0` and then rerender to `messageCount: 50, loadingOlder: false` to explicitly verify no scroll fires on that transition.

**2. Minor: `scrollRestoreRef` not in `useLayoutEffect` dependency array**

In `TaskConversationRenderer.tsx` line 132, the `useLayoutEffect` dependency array is `[rawMessages, scrollContainerRef, onLoadingOlderChange]`. `scrollRestoreRef` is read inside the effect body but is not a dependency because it is a ref (changing `.current` does not trigger re-render). This is technically correct Preact/React semantics, but if the ref object identity changed between renders (which does not happen with `useRef`), the effect would use a stale ref. No action needed -- this is standard ref usage.

---

### What I Verified

- **`useLayoutEffect` in Preact**: Preact supports `useLayoutEffect` natively (it maps to the browser's `layout` event). It fires synchronously after DOM mutations but before paint, which is exactly what is needed here.
- **Correctness of the guard logic in `useAutoScroll.ts`**: When `loadingOlder` transitions `true → false`, the guard effect updates `prevMessageCountRef` to the current `messageCount`. In the same render, the auto-scroll effect sees `hasNewContent = messageCount > prevMessageCountRef = false` (since `prevMessageCountRef` was already updated by the guard), so no auto-scroll fires. Execution order (guard effect runs before auto-scroll effect due to declaration order) makes this work correctly.
- **Integration with `TaskView.tsx`**: Traced the full flow from `handleLoadEarlier` through `onLoadingOlderChange(true)` → `loadEarlier()` → `rawMessages` change → `useLayoutEffect` fires → `onLoadingOlderChange(false)`. The `useLayoutEffect` fires before the parent's `useAutoScroll` re-renders with `loadingOlder = false`, which is the key ordering guarantee that makes the race impossible.
- **No other tests are affected**: Checked `TaskConversationRenderer.test.tsx` (no scroll-related tests), `TaskView.test.tsx` (uses mocks), and confirmed the new unit test integrates cleanly with existing test coverage.
- **E2E test logic**: The assertion `scrollAfter >= scrollBefore` correctly validates scroll position preservation. The forced `scrollToBottom` before clicking Load Earlier ensures `scrollBefore > 0` in practice, making the assertion meaningful.

---

### Recommendation

**APPROVE** -- the implementation is correct, the tests are adequate, and the two-part fix (synchronous layout effect + defense-in-depth guard) is the right approach.

lsm (commented):

### Overview

Reviewed PR #1002 which fixes a scroll position preservation bug when loading older messages in `TaskConversationRenderer`. The root cause was a race between scroll restoration (`useEffect` + `requestAnimationFrame`) and auto-scroll (`useEffect`-based in `useAutoScroll`).

### Changes Reviewed

1. **`TaskConversationRenderer.tsx`** — Changed scroll restoration from `useEffect` + `requestAnimationFrame` to `useLayoutEffect` (synchronous)
2. **`useAutoScroll.ts`** — Added guard that updates `prevMessageCountRef` when `loadingOlder` transitions `true → false`
3. **`task-message-pagination.e2e.ts`** — Added `requestAnimationFrame` wait before reading scroll position
4. **`useAutoScroll.test.ts`** — Added unit test for the `loadingOlder` transition guard

---

### Findings

#### P0 — No blocking issues found

The core fix is sound:
- `useLayoutEffect` correctly fires after DOM mutation but **before paint**, preventing the auto-scroll effect from racing ahead
- The scroll calculation (`scrollTop + (newScrollHeight - scrollHeight)`) is preserved unchanged, just executed synchronously
- The `onLoadingOlderChange?.(false)` call is correctly placed on all exit paths to prevent `isLoadingOlder` from getting stuck

#### P1 — Minor: Stale comment in E2E test

**File:** `packages/e2e/tests/features/task-message-pagination.e2e.ts`, lines 229-232

```typescript
// Wait for the requestAnimationFrame-based scroll restoration to complete.
// The component restores scrollTop inside rAF, so we must let one frame pass
// before reading the scroll position; otherwise we capture the pre-restore value.
await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
```

The comment references the old `requestAnimationFrame`-based implementation. With `useLayoutEffect` (synchronous), the scroll restoration happens before paint, so technically no rAF wait is needed. The comment is now misleading.

However, the rAF wait is **harmless** and ensures the test is robust regardless of implementation timing details. Suggest updating the comment to reflect the new reality or removing the rAF wait entirely if you want to verify the `useLayoutEffect` approach is sufficient on its own.

#### P2 — Inconsistency with `ChatContainer.tsx` scroll restoration pattern

**File:** `packages/web/src/islands/ChatContainer.tsx`, lines 530-553

`ChatContainer.tsx` uses `useEffect` + **double** `requestAnimationFrame` for the same scroll-restoration pattern:

```typescript
useEffect(() => {
    // ...
    requestAnimationFrame(() => {
        requestAnimationFrame(() => {
            if (container) {
                const newScrollTop = oldScrollTop + (container.scrollHeight - oldScrollHeight);
                container.scrollTop = newScrollTop;
            }
            scrollPositionRestoreRef.current = null;
        });
    });
}, [messages.length, loadingOlder]);
```

`TaskConversationRenderer.tsx` now uses `useLayoutEffect` without rAF. Both approaches work, but they're inconsistent. The `ChatContainer` comment explains the double rAF is needed because "Preact's signal updates and DOM reflows are asynchronous" — but `TaskConversationRenderer` uses `useGroupMessages` which returns `rawMessages` as a direct value (not a signal), making `useLayoutEffect` appropriate there.

This inconsistency is not a bug since the components use different data sources with different update models, but it could confuse future maintainers. Consider adding a comment in `TaskConversationRenderer.tsx` explaining why `useLayoutEffect` is sufficient (direct `rawMessages` value, not a signal).

#### P3 — Minor: `onLoadingOlderChange` in `useLayoutEffect` dependency array

**File:** `packages/web/src/components/room/TaskConversationRenderer.tsx`, line 132

```typescript
}, [rawMessages, scrollContainerRef, onLoadingOlderChange]);
```

`onLoadingOlderChange` is included in the dependency array. If the parent passes an unstable (non-`useCallback`'d) function, this could cause extra effect runs. Currently `TaskView.tsx` passes `setIsLoadingOlder` (the stable `useState` setter), so this is not a problem — but the dependency is implicit. Consider extracting a comment noting this must remain stable, or restructuring so the callback doesn't need to be a dependency.

---

### Verification

- `bun run typecheck` — **PASS**
- `bun run lint` — **PASS** (0 warnings, 0 errors)
- `make test-web` — **PASS** (1 unrelated pre-existing Dropdown test failure; `useAutoScroll` tests pass with 100% line coverage)

---

### Summary

The fix is correct and well-immented. The race condition is properly addressed by `useLayoutEffect` (synchronous scroll restoration before paint), and the guard in `useAutoScroll` provides defense-in-depth against message-count-based auto-scroll triggering. No P0 or P1 issues. Two P2 suggestions: update the stale E2E test comment, and add a clarifying comment about why `ChatContainer` needs double rAF while `TaskConversationRenderer` does not.

lsm:

Reviewed PR #1002 which fixes a scroll position preservation bug when loading older messages in TaskConversationRenderer.

### Summary

The fix is **correct** — useLayoutEffect (synchronous) properly prevents the race with useAutoScroll effect. The guard in useAutoScroll provides defense-in-depth.

### Issues Found

**P2 (important suggestion):**

1. **Stale comment in E2E test** (packages/e2e/tests/features/task-message-pagination.e2e.ts lines 229-232): The comment references the old requestAnimationFrame-based implementation. With useLayoutEffect, no rAF wait is technically needed. Consider updating or removing.

2. **Inconsistency with ChatContainer.tsx** (packages/web/src/islands/ChatContainer.tsx lines 530-553): ChatContainer uses useEffect + double rAF for the same scroll-restoration pattern, while TaskConversationRenderer now uses useLayoutEffect without rAF. Both work but are inconsistent — worth adding a clarifying comment.

**P3 (nit):**

3. **onLoadingOlderChange dependency** (TaskConversationRenderer.tsx line 132): Included in useLayoutEffect deps. Currently safe (parent passes stable setIsLoadingOlder setter), but implicit. Consider a comment noting this must remain stable.

### Verification

- typecheck: PASS
- lint: PASS
- make test-web: PASS (useAutoScroll 100% line coverage)

PR #984, opened 2026-03-26
test(space): CompletionDetector and tick loop completion detection tests

lsm:
Add 31 new tests (39 CompletionDetector + 19 SpaceRuntime tick loop = 58 space tests) covering all acceptance criteria for completion detection.

**CompletionDetector (tests 1-39):**
- Tests 1-25: original scenarios — no tasks, single/multi agent, multi-node, channels, TERMINAL_TASK_STATUSES
- Tests 26-27: orchestration task exclusion (null workflowNodeId)
- Tests 28-31: mixed terminal statuses (completed+cancelled, completed+needs_attention, all five terminal, pending blocks)
- Tests 32-33: cross-run isolation (different runs, empty vs non-empty runs)
- Tests 34-39: pending-but-blocked via real channel topology (3-node chain, bidirectional, array target, wildcard+downstream)

**SpaceRuntime tick loop (19 tests):**
- completedAt timestamp set on completion (verifies recent timestamp)
- Multi-node workflows: mixed terminal statuses, in_progress blocks
- Early returns: needs_attention/completed/cancelled runs skipped
- Status lifecycle: in_progress→completed, needs_attention→in_progress→completed via resume
- Pending-but-blocked via workflow channels: unactivated node prevents completion, no channels = status-only check, wildcard doesn't block
- Multi-agent parallel nodes: staggered completion across ticks
- Dedup cleanup: verifies notifiedTaskSet is cleared when executor removed

**Review fixes:**
- completedAt test now verifies timestamp is set to a recent value (not just > 0)
- Dedup cleanup test now verifies notifiedTaskSet entries are actually removed via SpaceRuntime.getNotifiedTaskSet()
- completion-detector.test.ts header updated to list all 39 scenarios
- direction typo fixed ('two-way' → 'bidirectional')

**SDK mock fix:**
- Added mock.module('@anthropic-ai/claude-agent-sdk', ...) to tests/unit/setup.ts — provides a functional stub of createSdkMcpServer with proper _registeredTools, tool(), query(), interrupt(). This fixes the pre-existing SyntaxError that was crashing many room/space unit tests in CI.
- All 8165 daemon unit tests pass.

PR #1005, opened 2026-03-26
fix(tests): resolve SDK module resolution in test-daemon

lsm:
## Summary

- Run daemon tests from `packages/daemon/` where `@anthropic-ai/claude-agent-sdk` is installed, instead of the workspace root where it can't be resolved
- Create missing `tests/helpers/default-sdk-mock.ts` helper
- Restore full SDK mock exports in `anthropic-provider.test.ts` and `session-lifecycle-sdk-title.test.ts`

## Test plan

- [x] `make test-daemon` passes: 8173 pass, 0 fail, 8 skip

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,11 @@ build:
 test: test-daemon test-web
 
 test-daemon:
+	@echo "Running shared tests..."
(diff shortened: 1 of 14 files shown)


lsm:
Closing - the underlying tests are covered by #984 and CI passes without this change.

PR #1011, opened 2026-03-27
fix: E2E test failures — draft tasks, selectors, reference search

lsm:
## Summary
- Fix E2E tests that fail because tasks auto-execute in a non-git workspace by adding `status: 'draft'` to task creation calls
- Replace broken Playwright `filter({ hasNot })` with CSS `:not()` selector for Task Agent node exclusion
- Fix mobile panel detection using `boundingBox()` instead of `isVisible()` for CSS translate-x hidden elements
- Handle synthetic `room:chat:<roomId>` session ID in `reference.search` so autocomplete works on the room agent route
- Move `setStartStepId` inside `setNodes` updater to fix React StrictMode violation

## Test plan
- [ ] Unit tests pass: `bun test packages/daemon/tests/unit/reference-handlers.test.ts`
- [ ] E2E tests: short-id-display, task-goal-indicator, livequery-task-goal-updates, space-agent-centric-workflow (tests 1-2), mobile room agent navigation
- [ ] `bun run check` passes (lint, typecheck, knip)

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
@@ -206,15 +206,22 @@ export function setupReferenceHandlers(messageHub: MessageHub, deps: ReferenceHa
 		if (typeof params.query !== 'string') throw new Error('query must be a string');
 
 		const query = params.query.trim();
-		// Empty/whitespace-only query would match everything — return nothing instead
-		if (!query) return { results: [] };
 
 		const requestedTypes: ReferenceType[] =
 			params.types && params.types.length > 0 ? params.types : ['task', 'goal', 'file', 'folder'];
 
-		// Resolve room context from session
+		// Resolve room context from session.
+		// The room agent route uses a synthetic session ID "room:chat:<roomId>"
+		// which has no DB entry — extract the roomId from it directly.
 		const session = sessionManager.getSessionFromDB(params.sessionId);
-		const roomId = session?.context?.roomId;
+		let roomId = session?.context?.roomId;
+		if (!roomId && params.sessionId.startsWith('room:chat:')) {
+			roomId = params.sessionId.slice('room:chat:'.length);
+		}
+
+		// Empty query with no room context: nothing to search.
+		// Empty query WITH room context: return all tasks/goals for the room.
+		if (!query && !roomId) return { results: [] };
 
 		const allResults: ReferenceSearchResult[] = [];
 
@@ -317,6 +324,11 @@ async function resolveSessionContext(
 ): Promise<{ workspacePath: string; roomId: string | null }> {
 	const agentSession = await deps.sessionManager.getSessionAsync(sessionId);
 	if (!agentSession) {
+		// The room agent route uses a synthetic session ID "room:chat:<roomId>"
+		// which has no DB entry — extract the roomId from it.
+		if (sessionId.startsWith('room:chat:')) {
+			return { workspacePath: deps.workspaceRoot, roomId: sessionId.slice('room:chat:'.length) };
+		}
 		return { workspacePath: deps.workspaceRoot, roomId: null };
 	}
 
--- a/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/reference-handlers.ts
@@ -273,6 +273,9 @@ export function setupReferenceHandlers(messageHub: MessageHub, deps: ReferenceHa
 		if (requestedTypes.includes('folder')) fileTypes.push('folder');
 
 		if (fileTypes.length > 0 && query.length > 0) {
+			// Note: file/folder search requires a non-empty query — returning all files
+			// on empty query would be too slow and noisy. Task/goal results above are
+			// returned for empty queries since they are bounded to the room scope.
 			// Path traversal check — reject queries with .. or absolute paths
 			if (query.includes('..') || query.startsWith('/')) {
 				// Return what we have so far without file results
--- a/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/task-handlers.ts
@@ -104,6 +104,7 @@ export function setupTaskHandlers(
 			description: string;
 			priority?: TaskPriority;
 			dependsOn?: string[];
+			status?: TaskStatus;
 		};
 
 		if (!params.roomId) {
@@ -119,6 +120,7 @@ export function setupTaskHandlers(
 			description: params.description ?? '',
 			priority: params.priority,
 			dependsOn: params.dependsOn,
+			status: params.status,
 		});
 
 		// Emit room.overview for new task creation (significant change)
--- a/packages/daemon/tests/unit/providers/anthropic-provider.test.ts
+++ b/packages/daemon/tests/unit/providers/anthropic-provider.test.ts
@@ -123,13 +123,45 @@ describe('AnthropicProvider', () => {
 			// Set credentials so SDK load is attempted
 			process.env.ANTHROPIC_API_KEY = 'test-key';
 
-			// Mock the SDK module with a query that fails when supportedModels is called
+			// Mock the SDK module with a query that fails when supportedModels is called.
+			// IMPORTANT: we must still provide createSdkMcpServer and tool so that other
+			// tests (e.g. provision-global-agent, task-agent-tools) don't break when
+			// this mock replaces setup.ts's default SDK mock.
+			class MockMcpServer {
+				readonly _registeredTools: Record<string, object> = {};
+				connect(): void {}
+				disconnect(): void {}
+			}
+			let _toolBatch: Array<{ name: string; def: object }> = [];
 			mock.module('@anthropic-ai/claude-agent-sdk', () => ({
 				query: async () => ({
 					interrupt: () => {},
 					supportedModels: async () => {
 						throw new Error('SDK unavailable');
 					},
+				}),
+				interrupt: mock(async () => {}),
+				supportedModels: mock(async () => {
+					throw new Error('SDK unavailable');
+				}),
+				createSdkMcpServer: mock((_options: { name: string; tools?: unknown[] }) => {
+					const server = new MockMcpServer();
+					for (const { name, def } of _toolBatch) {
+						server._registeredTools[name] = def;
+					}
+					_toolBatch = [];
+					return {
+						type: 'sdk' as const,
+						name: _options.name,
+						version: _options.version ?? '1.0.0',
+						tools: _options.tools ?? [],
+						instance: server,
+					};
+				}),
+				tool: mock((name: string, description: string, inputSchema: unknown, handler: unknown) => {
+					const def = { name, description, inputSchema, handler };
+					_toolBatch.push({ name, def });
+					return def;
 				}),
 			}));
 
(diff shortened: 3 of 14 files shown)


lsm (commented):

---

## Summary of 5 Changes

### 1. Draft status for E2E tasks (task-handlers.ts, room-helpers.ts, e2e tests)

**Verdict: Correct, no issues.**

The `status` parameter is properly added as optional to the `task.create` handler and passed through to `taskManager.createTask()`. Using `draft` status in E2E tests is the right workaround for the non-git workspace constraint.

---

### 2. hasNot filter replaced with CSS :not() (space-agent-centric-workflow.e2e.ts)

**Verdict: Correct, no issues.**

The old `filter({ hasNot: page.locator('[data-task-agent=true]') })` only matched descendants, not the element itself. The attribute `data-task-agent` is on the node element, so `hasNot` would always pass. The CSS `:not([data-task-agent])` correctly targets the element's own attribute. Well-documented comment explaining the bug is a nice touch.

---

### 3. Mobile panel detection (mobile-helpers.ts)

**Verdict: Mostly correct — one P2 concern.**

`boundingBox().x >= 0` is the right way to detect CSS translate-x hidden panels (translated-off elements are still technically visible to Playwright). `toBeInViewport()` is semantically better than `isVisible()` here.

**P2: Potential fragility with `force: true` in `closeMobilePanel`.** The `force: true` bypasses actionability checks (visible, stable, enabled, receiving events, not obscured). If the close button selector ever matches multiple elements or is partially obscured by an unexpected overlay, `force: true` will click the first one regardless. This is acceptable for a test helper, but worth noting. A more robust alternative would be `page.locator('button[title="Close panel"]').first().click({ force: true })` to be explicit about targeting the first match.

**P2: The test change in mobile.e2e.ts (line 337) changes the tab being clicked from "Inbox" to "/" and the URL assertion from `/inbox$/` to `/$`.** The PR description says "Fixed test 14's "/" tab click" — but changing the test to match the behavior it was supposedly fixing is a code smell. The test's intent was to verify room-to-global tab transition; now it verifies a different transition (home tab instead of inbox tab). If the "/" tab is the correct target, the change is fine, but the test description ("Click Inbox tab... URL should change to inbox") no longer matches what the test does.

---

### 4. Reference search — synthetic room:chat: session ID (reference-handlers.ts)

**Verdict: Correct intent — two P2 concerns about inconsistency and missing test coverage.**

**P2a: Inconsistency between task/goal and file/folder branches with empty query + roomId.**

With a synthetic `room:chat:{roomId}` session ID and an empty query, the search handler now:
- Tasks: returns ALL tasks for the room
- Goals: returns ALL goals for the room
- Files: skips file search entirely (returns nothing)
- Folders: skips folder search entirely (returns nothing)

The comment says "Empty query WITH room context: return all tasks/goals for the room" — but files and folders silently return nothing, which is inconsistent and undocumented. This could surprise users who expect `@` (empty query) to show all references of every type.

**P2b: Missing test coverage for edge cases.** The new tests cover the happy path (empty query with real session, empty query with synthetic session). But there are untested edge cases:
- Whitespace-only query (`"   "`) with synthetic session ID — does it return all tasks/goals? (The old behavior returned empty for whitespace-only; the new guard `if (!query && !roomId)` means whitespace-only query with roomId would pass through and trigger the "return all" behavior.)
- Malformed synthetic IDs like `room:chat:` (empty roomId) — does this gracefully degrade or match all rooms?

---

### 5. Visual editor — setStartStepId moved inside setNodes updater (VisualWorkflowEditor.tsx)

**Verdict: Correct, no issues.**

This is the cleanest fix in the PR. Moving `setStartStepId` inside the `setNodes` updater is the right approach for React StrictMode compliance. The updater function is pure (only reads `prev`, computes `isFirstRegular`, calls `setStartStepId` with a stable `newLocalId` value captured in closure). In StrictMode double-invocation, both calls set the same start step ID, which is idempotent and correct.

---

## Unrelated Test Infrastructure Changes

The changes to `anthropic-provider.test.ts` and `session-lifecycle-sdk-title.test.ts` (adding `createSdkMcpServer`, `tool`, `interrupt`, `supportedModels` mocks) are test infrastructure fixes, not part of the E2E failure fix. They appear to address module-level mock pollution in Bun's test runner where `mock.module` replacement affects subsequent tests. These changes are unrelated to the PR's stated purpose and add noise, but they're not incorrect.

The goal-manager test change (`@hourly` -> `@weekly`) is a legitimate test robustness fix (avoiding coincident midnight timestamps between `@hourly` and `@daily`).

---

## Recommendation

**REQUEST_CHANGES** due to P2 issues in changes 3 and 4:

1. **(P2)** Add explicit `.first()` to the `force: true` click in `closeMobilePanel` to avoid ambiguity if the selector ever matches multiple elements.
2. **(P2)** The mobile.e2e.ts test change should update its description/comments to match the new behavior (or restore the original test if the "/" tab was a mistake).
3. **(P2)** Document or address the file/folder inconsistency in `reference.search` with empty query + roomId.
4. **(P2)** Add a test for whitespace-only query with synthetic session ID to confirm it returns all tasks/goals.

lsm (commented):

Reviewed all 5 stated fixes plus related test changes.

---

### Change 1: E2E test fix (draft status) — CORRECT

Adding `status: 'draft'` to task creation in E2E tests is correct and complete. The fix flows correctly through the entire chain:
- `task-handlers.ts` now accepts and forwards `status` to `taskManager.createTask()`
- `CreateTaskParams.status` (in `neo.ts:382`) documents that `'draft'` is a valid value
- `task-repository.ts:46` uses `params.status ?? 'pending'`, so `'draft'` is honored

The comment explaining why (`// the E2E workspace is not a git repo, so worktree creation fails`) is accurate and appropriately placed in all 4 affected files.

**No issues.**

---

### Change 2: E2E test fix (hasNot filter) — CORRECT

Replacing `filter({ hasNot: page.locator('[data-task-agent="true"]') })` with the CSS selector `[data-testid^="workflow-node-"]:not([data-task-agent])` is the right fix. The PR comment explains it correctly: Playwright's `hasNot` filter matches **descendants**, not the element's own attributes. Since `data-task-agent="true"` is on the node element itself (not a child), `hasNot` would never exclude it.

The `:not([data-task-agent])` selector correctly targets elements that do NOT have that attribute set, which is what the tests intend.

**No issues.**

---

### Change 3: E2E test fix (mobile panel detection) — CORRECT

`closeMobilePanel` now uses `boundingBox()` and checks `box.x >= 0` to detect CSS `translate-x` hidden panels. This is the right approach — elements with `transform: translate-x(-100%)` are still technically "visible" to Playwright but are off-screen. The `force: true` on the close button click is also correct to bypass z-index interception from the `BottomTabBar`.

The test 14 change (clicking `/` tab instead of Inbox) navigates to the full URL (`page.goto(\`/room/${roomId}\`)`) as noted in the PR description, which is the correct workaround.

**No issues.**

---

### Change 4: Backend fix (reference search) — CORRECT, with a minor note

Handling synthetic `room:chat:{roomId}` session IDs is done correctly in two places:
1. `resolveSessionContext()` (used by `reference.resolve`) — extracts roomId and returns `workspaceRoot`
2. `reference.search` handler — same extraction logic

The new guard logic is logically correct:
- `!query && !roomId` -> return empty (nothing to search, no context to return "all")
- `!query && roomId` -> return all tasks/goals for that room

**Minor note:** `sessionId.startsWith('room:chat:')` is checked twice (once in `reference.search` at line 218 and once in `resolveSessionContext` at line 329). This is intentional since these are separate code paths (different handlers), but worth noting for future maintainers.

**New unit tests added** (`reference-handlers.test.ts`) cover the two new cases.

**No issues.**

---

### Change 5: Visual editor fix (setStartStepId) — CORRECT fix, slightly misleading commit message

The original code had a stale-closure bug: it read `nodes` (the outer closure variable, captured before `setNodes` was called) to decide `isFirstNode`, then called `setStartStepId` **after** `setNodes` completed. This is incorrect because the closure `nodes` reflects state from the current render, not the previous state, so the timing was wrong.

The fix moves `setStartStepId(newLocalId)` inside the `setNodes` updater, reading `regularCount === 0` from `prev`. This is the correct fix.

**Minor concern:** The commit message frames this as a "React StrictMode compliance" fix, which is slightly misleading. The underlying bug exists regardless of StrictMode — it is a stale-closure issue. StrictMode's double-invocation merely surfaces the bug more reliably. This is a P3 nit — the fix is correct, the framing is slightly off.

**No issues.**

---

### Other test changes NOT mentioned in PR description

Three additional test changes appear without justification in the PR description:

**`anthropic-provider.test.ts` and `session-lifecycle-sdk-title.test.ts`:** Both add `createSdkMcpServer` and `tool` to their SDK mocks. The comment says "we must still provide createSdkMcpServer and tool so that other tests don't break when this mock replaces setup.ts's default SDK mock." This is a valid concern, but since these changes are not tied to any described failing test, verify whether they were actually necessary or are preemptive hardening.

**`goal-manager.test.ts`:** Changed `@hourly` to `@weekly` in the `nextRunAt` assertion. The comment explains this correctly — `@hourly` and `@daily` can coincide at midnight, causing a flaky failure. This is a valid fix.

---

### Summary

All 5 described fixes are correct and robust. The test changes (mock SDK additions, `@weekly` schedule) are valid improvements. The only P3 nit is the slightly misleading commit message framing of Change 5.

lsm (commented):

Reviewed all 5 P2 issues from the previous review:

### P2-1: `closeMobilePanel` `.first()` — RESOLVED
`packages/e2e/tests/helpers/mobile-helpers.ts` line 60: `.first()` is present with `force: true` and `timeout: 5000`. Comment explains BottomTabBar z-index overlap. Correct.

### P2-2: mobile.e2e.ts comment updated — RESOLVED
`packages/e2e/tests/responsive/mobile.e2e.ts` line 336: Tab correctly identified as "/" (Home), not "Inbox". Comment is accurate. Correct.

### P2-3: File search empty-query behavior documented — RESOLVED
`packages/daemon/src/lib/rpc-handlers/reference-handlers.ts` lines 276-278: Behavior explicitly documented. Correct.

### P2-4: New unit test for whitespace-only query with synthetic session ID — RESOLVED
`packages/daemon/tests/unit/reference-handlers.test.ts` lines 827-848: Test covers the exact scenario. All 32 tests pass. Correct.

### P2-5: Commit message amended — RESOLVED
Commit `d251da087` now describes "stale-closure" bug rather than "StrictMode compliance". Correct.

All P2 issues are properly resolved. Unit tests pass (32/32).

lsm (commented):

### P2 Issue Verification

All five P2 issues from the previous review are properly addressed:

1. **`closeMobilePanel` `.first()`** — `packages/e2e/tests/helpers/mobile-helpers.ts` line 60 now uses `.first()`: `await closePanelButton(page).first().click({ force: true, timeout: 5000 });`. The explanatory comment on lines 57-59 correctly documents why `force: true` is needed (BottomTabBar z-index overlap).

2. **mobile.e2e.ts comment updated** — `packages/e2e/tests/responsive/mobile.e2e.ts` line 336 now accurately says `// Click the \"/\" (Home) tab — present in room context, navigates to global home` and the assertion on line 340 correctly checks `/\$/` (home URL), not `/\/inbox$/`.

3. **File search empty-query behavior documented** — `packages/daemon/src/lib/rpc-handlers/reference-handlers.ts` lines 276-278 now include a clear comment explaining why file/folder search is skipped on empty query.

4. **New unit test added** — `packages/daemon/tests/unit/reference-handlers.test.ts` line 827 adds the whitespace-only query test with synthetic session ID. All 32 tests pass.

5. **Commit message amended** — Commit `d251da087` uses the correct message: \"fix: fix setStartStepId called in outer setNodes scope causing stale-closure\".

### Quality Checks

- `bun run check` passes: lint (0 warnings), typecheck, and knip all clean.
- Unit tests: 32/32 pass in `reference-handlers.test.ts`.

### Recommendation

**APPROVE** — all P2 issues are correctly resolved with no regressions.

PR #1014, opened 2026-03-27
fix: detect stale messageQueue state and retry worker-to-leader routing

lsm:
## Summary

Fixes a race condition where `routeWorkerToLeader()` fails to deliver worker output to the leader session. When a leader's SDK query finishes, there's a window between the for-await loop ending and the `finally` block calling `messageQueue.stop()` where `isRunning()` returns true but no generator is consuming messages, causing `injectMessage` to silently orphan the message.

- `ensureQueryStarted()` now detects stale running state (`isRunning=true` but `queryPromise=null`) and force-stops the queue before restarting
- `routeWorkerToLeader()` retries `injectMessage` up to 3 times with backoff, failing the task only after all retries are exhausted
- Added diagnostic logging to `QueryRunner.start()`, `ensureQueryStarted()`, and `injectMessage` for message delivery tracing

## Test plan

- Unit test: `ensureQueryStarted` detects stale running state and restarts query
- Unit test: `routeWorkerToLeader` retries on transient inject failure and succeeds
- Unit test: `routeWorkerToLeader` fails task after all retries exhausted
- All 225 existing tests across affected files pass

### DIFF
--- a/packages/daemon/src/lib/agent/query-lifecycle-manager.ts
+++ b/packages/daemon/src/lib/agent/query-lifecycle-manager.ts
@@ -313,6 +313,11 @@ export class QueryLifecycleManager {
 	 *
 	 * Waits for any pending interrupt, validates SDK session file,
 	 * and starts the streaming query if not already running.
+	 *
+	 * Detects stale running state: if messageQueue.isRunning() is true but
+	 * queryPromise is null, the queue was not properly stopped after the previous
+	 * query ended (race between SDK query completion and finally block cleanup).
+	 * In this case, force-stop the queue and restart.
 	 */
 	async ensureQueryStarted(): Promise<void> {
 		const { session, db, messageQueue, interruptHandler } = this.ctx;
@@ -328,7 +333,25 @@ export class QueryLifecycleManager {
 		}
 
 		if (messageQueue.isRunning()) {
-			return;
+			// Stale running state detection: if the queue thinks it's running but
+			// there's no active query promise, the previous query's finally block
+			// hasn't called stop() yet (or it was lost). Force-stop and restart
+			// so the enqueued message is not orphaned.
+			if (!this.ctx.queryPromise) {
+				this.logger.warn(
+					`Stale running state detected for session ${session.id}: ` +
+						`messageQueue.isRunning()=true but queryPromise=null. Force-stopping and restarting.`
+				);
+				messageQueue.stop();
+				// Fall through to start a fresh query below
+			} else {
+				this.logger.debug(
+					`ensureQueryStarted: session ${session.id} already running, skipping start`
+				);
+				return;
+			}
+		} else {
+			this.logger.debug(`ensureQueryStarted: session ${session.id} not running, starting query`);
 		}
 
 		// Validate SDK session file
--- a/packages/daemon/src/lib/agent/query-lifecycle-manager.ts
+++ b/packages/daemon/src/lib/agent/query-lifecycle-manager.ts
@@ -333,16 +333,22 @@ export class QueryLifecycleManager {
 		}
 
 		if (messageQueue.isRunning()) {
-			// Stale running state detection: if the queue thinks it's running but
-			// there's no active query promise, the previous query's finally block
-			// hasn't called stop() yet (or it was lost). Force-stop and restart
-			// so the enqueued message is not orphaned.
+			// Defensive stale state detection: if the queue thinks it's running but
+			// there's no active query promise, the session is in an inconsistent state
+			// (e.g., restored session with stale queue flag, or cleanup was interrupted).
+			// The primary race (between for-await loop ending and finally block cleanup)
+			// is handled by the early messageQueue.stop() in QueryRunner.runQuery().
+			// This check catches residual edge cases where queryPromise has already
+			// been nulled but the queue wasn't stopped.
 			if (!this.ctx.queryPromise) {
 				this.logger.warn(
 					`Stale running state detected for session ${session.id}: ` +
 						`messageQueue.isRunning()=true but queryPromise=null. Force-stopping and restarting.`
 				);
 				messageQueue.stop();
+				// Clear stale query reference to prevent concurrent callers from
+				// seeing a dead query object during the restart window.
+				this.ctx.queryObject = null;
 				// Fall through to start a fresh query below
 			} else {
 				this.logger.debug(
--- a/packages/daemon/src/lib/agent/query-runner.ts
+++ b/packages/daemon/src/lib/agent/query-runner.ts
@@ -316,6 +316,15 @@ export class QueryRunner {
 				}
 			}
 
+			// Stop the queue immediately after the query ends to close the race window
+			// between the for-await loop ending and the finally block calling stop().
+			// Without this, ensureQueryStarted() can see isRunning()=true while no
+			// generator is consuming messages, causing enqueued messages to be orphaned.
+			// Guard: only stop if this is still the current query (not stale from a restart).
+			if (this.ctx.getQueryGeneration() === queryGeneration) {
+				messageQueue.stop();
+			}
+
 			// If startup timed out before first message, surface as timeout error
 			// (after abort-driven iterator shutdown) so error state is visible.
 			if (startupTimeoutReached && messageCount === 0) {
--- a/packages/daemon/src/lib/agent/query-runner.ts
+++ b/packages/daemon/src/lib/agent/query-runner.ts
@@ -98,12 +98,21 @@ export class QueryRunner {
 	 * Start the streaming query (called from AgentSession.startStreamingQuery)
 	 */
 	async start(): Promise<void> {
-		const { messageQueue } = this.ctx;
+		const { messageQueue, logger } = this.ctx;
 
 		if (messageQueue.isRunning()) {
+			logger.warn(
+				`QueryRunner.start(): messageQueue already running for session ${this.ctx.session.id}, ` +
+					`skipping start (generation=${messageQueue.getGeneration()}, ` +
+					`queryPromise=${this.ctx.queryPromise ? 'active' : 'null'})`
+			);
 			return;
 		}
 
+		logger.debug(
+			`QueryRunner.start(): starting query for session ${this.ctx.session.id} ` +
+				`(generation=${messageQueue.getGeneration()})`
+		);
 		messageQueue.start();
 
 		// Increment query generation for this new query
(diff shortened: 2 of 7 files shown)


lsm (commented):

### Summary

This PR addresses a race condition where worker output fails to reach the leader session when the leader's SDK query has just finished but the message queue hasn't been stopped yet. The fix introduces stale state detection and retry logic.

### Strengths

1. **Root cause analysis is correct**: The race between SDK query completion and `finally` block cleanup is real. When the for-await loop ends, `messageQueue.isRunning()` returns true but no generator is consuming messages.

2. **Defense-in-depth approach**: Combining stale state detection with retry logic provides two layers of protection.

3. **Good diagnostic logging**: The added logs will help debug similar issues in production.

### Issues Found

#### P1: Stale state recovery doesn't clear `queryObject`

**File**: `/packages/daemon/src/lib/agent/query-lifecycle-manager.ts` (lines 335-346)

When stale state is detected (`isRunning=true` but `queryPromise=null`), the code calls `messageQueue.stop()` but doesn't clear `this.ctx.queryObject`. The `queryObject` could still reference an old SDK Query object that's no longer valid.

```typescript
if (!this.ctx.queryPromise) {
    this.logger.warn(...);
    messageQueue.stop();
    // Missing: this.ctx.queryObject = null;
    // Fall through to start a fresh query below
}
```

The subsequent `startStreamingQuery()` call should overwrite it, but there's a brief window where a concurrent call could see a stale `queryObject`. Consider adding:
```typescript
messageQueue.stop();
this.ctx.queryObject = null; // Clear stale query reference
```

#### P1: Missing test assertion for restart behavior

**File**: `/packages/daemon/tests/unit/agent/query-lifecycle-manager.test.ts` (line 588)

The test `detects stale running state and restarts when isRunning=true but queryPromise=null` verifies that `stop()` was called but doesn't explicitly verify the query was actually restarted:

```typescript
const stopSpy = spyOn(messageQueue, 'stop');
await manager.ensureQueryStarted();
expect(stopSpy).toHaveBeenCalled();
expect(startStreamingCalled).toBe(true);
```

The `startStreamingCalled` check is good, but consider also verifying that `startStreamingQuery` was called **after** `stop()` by checking the call order.

#### P2: Retry count semantics are confusing

**File**: `/packages/daemon/src/lib/room/runtime/task-group-manager.ts` (line 483)

```typescript
const MAX_INJECT_RETRIES = 2;
// ...
for (let attempt = 0; attempt <= MAX_INJECT_RETRIES; attempt++) {
```

This gives 3 total attempts (0, 1, 2). The constant is named "MAX_RETRIES" but the loop uses `<=`, making the actual count `MAX_RETRIES + 1`. Consider either:
- Renaming to `MAX_INJECT_ATTEMPTS = 3`
- Or changing the loop to `attempt < MAX_INJECT_RETRIES`

The current code works but the naming is misleading.

#### P2: No retry for `routeLeaderToWorker`

**File**: `/packages/daemon/src/lib/room/runtime/task-group-manager.ts` (lines 541-588)

The retry logic is only added to `routeWorkerToLeader`, but `routeLeaderToWorker` has the same underlying dependency on `injectMessage`. If the race condition can happen in one direction, it can happen in the other. Consider applying the same retry pattern to `routeLeaderToWorker`.

#### P3: Consider exponential backoff instead of linear

**File**: `/packages/daemon/src/lib/room/runtime/task-group-manager.ts` (line 493)

```typescript
await new Promise((r) => setTimeout(r, RETRY_DELAY_MS * attempt));
```

Linear backoff (1000ms, 2000ms) is fine for 3 attempts, but exponential backoff (1000ms, 2000ms, 4000ms) would be more robust for transient failures. This is a minor suggestion.

The implementation is solid overall, but the missing `queryObject` cleanup in stale state recovery could lead to edge case issues, and the test coverage could be more thorough. Please address the P1 items before merging.

lsm (commented):

---

### Overall Assessment

The fix addresses a real operational issue -- worker output being silently dropped when routed to a leader that just finished its query. The two-layer defense (stale state detection + retry with backoff) is a sound approach. Tests pass, lint is clean, typecheck is clean. However, there are a few issues worth addressing.

---

### P1: Retry logic does not distinguish transient from permanent errors

**File:** `packages/daemon/src/lib/room/runtime/task-group-manager.ts`, lines 486-504

The retry loop catches all errors uniformly and retries them. But `injectMessage` can throw permanent errors such as `Error('Session not in service cache')` (line 393 of `room-runtime-service.ts`). Retrying a permanently missing session wastes up to 3 seconds of delay plus the time for each failed call. Consider adding a simple check to avoid retrying known-permanent errors, for example:

```ts
const NON_RETRYABLE_PATTERNS = ['Session not in service cache', 'not found in service cache'];
const isNonRetryable = NON_RETRYABLE_PATTERNS.some(p => lastError.message.includes(p));
if (isNonRetryable) break;
```

This is not strictly blocking since the task is ultimately failed anyway, but it avoids unnecessary latency and log noise when the leader session is genuinely gone.

---

### P2: Stale state detection may never trigger in practice for the described race window

**File:** `packages/daemon/src/lib/agent/query-lifecycle-manager.ts`, lines 335-352

After tracing through `query-runner.ts` (lines 447-506), the finally block calls `messageQueue.stop()` synchronously at line 470 before any `await`, and `queryPromise` is nulled at line 503 after all async gaps. This means during the race window (between SDK query end and `messageQueue.stop()`), `queryPromise` is still non-null (it is a resolved promise whose finally block is still executing). The stale detection check `!this.ctx.queryPromise` evaluates to `false` for resolved promises (they are truthy), so the stale path would NOT be taken.

The only way `isRunning()=true && queryPromise=null` can occur is if `queryPromise` was explicitly set to `null` by `QueryLifecycleManager.stop()` (line 150) or the finally block (line 503) -- but both of those also call `messageQueue.stop()` first, making `isRunning()=false`.

This means the stale detection code is defensive but may not address the originally described race window. The retry logic in `task-group-manager.ts` is likely the more effective part of this fix. Consider either:
1. Updating the detection to also handle `queryPromise` being a resolved (settled) promise, or
2. Clarifying in the commit message/code comment what exact scenario triggers this detection path, to aid future maintainers.

---

### P2: Backoff delay formula produces linear backoff starting at full delay

**File:** `packages/daemon/src/lib/room/runtime/task-group-manager.ts`, line 493

The delay is `RETRY_DELAY_MS * attempt` where attempt starts at 1 for the first retry, giving delays of 1000ms and 2000ms. A more conventional exponential backoff (e.g., 1000, 2000, 4000) or even just 500, 1000 would reduce the time-to-success for transient failures. The current values are not wrong, but given that the root cause is a race condition that resolves in milliseconds, a 1-second first delay is relatively long. This is a suggestion, not a requirement.

---

### P3: Minor -- test for stale state uses `spyOn` but does not verify the logger warning

**File:** `packages/daemon/tests/unit/agent/query-lifecycle-manager.test.ts`, lines 574-592

The test verifies that `stop()` was called and `startStreamingQuery` was called, but does not assert that the warning log was emitted. Adding a logger spy check would strengthen the test and serve as documentation for future developers about the expected diagnostic output.

---

### What Works Well

- The retry logic in `routeWorkerToLeader()` is clean and well-structured. The `setLeaderHasWork()` placement before the retry loop is correct.
- The diagnostic logging in `room-runtime-service.ts` and `query-runner.ts` will significantly help debug message delivery issues in production without being excessively verbose.
- The test coverage for the retry path (success and exhaustion) is solid and uses realistic mock patterns.
- The `query-runner.test.ts` mock correctly adds the `getGeneration` method that the new logging code calls.
- The `query-lifecycle-manager.test.ts` correctly updates the existing "already running" test to set `queryPromise` to a non-null value, which was an important fix to the pre-existing test.
- Failing the task (rather than silently dropping) after retry exhaustion is the right behavior.

lsm (commented):

### Verdict: APPROVE

All previously raised P1 issues have been properly addressed in the second commit (9170d312). The implementation is sound and well-tested.

---

### Issue-by-issue resolution

**P1 (stale state recovery clears queryObject): RESOLVED**
`packages/daemon/src/lib/agent/query-lifecycle-manager.ts` line 351 now sets `this.ctx.queryObject = null` after detecting stale state, preventing concurrent callers from seeing a dead query reference. The test at `packages/daemon/tests/unit/agent/query-lifecycle-manager.test.ts:595` verifies this with `expect(mockContext.queryObject).toBeNull()`.

**P1 (race condition fixed at source): RESOLVED**
`packages/daemon/src/lib/agent/query-runner.ts` lines 319-326 call `messageQueue.stop()` immediately after the for-await loop ends, before any await or async work. This is a synchronous call -- `stop()` just sets `this.running = false` and wakes waiters, so no interleaving can occur. The finally block at line 479 calls `stop()` again, but this is safe because the method is idempotent (verified in `message-queue.ts` line 185-189). The generation guard at line 324 ensures stale queries from restarts don't interfere. The comment at line 336-342 in `query-lifecycle-manager.ts` correctly repositions the stale-state check as a defensive fallback rather than the primary fix.

**P1 (extracted injectMessageWithRetry helper): RESOLVED**
`packages/daemon/src/lib/room/runtime/task-group-manager.ts` lines 828-861 extract a clean `injectMessageWithRetry()` helper with exponential backoff (500ms, 1000ms), used by both `routeWorkerToLeader` (line 483) and `routeLeaderToWorker` (line 560). The interface is clean: returns `null` on success, `Error` on failure. Both callers handle the error path consistently (fail the group).

**P2 (non-retryable error detection): RESOLVED**
Lines 836 and 854-857 bail immediately on permanent errors matching `not in service cache` or `not found in service cache`. The test at `task-group-manager.test.ts:876-927` verifies only 1 attempt is made for non-retryable errors.

---

1. **No explicit test for the early stop() call in QueryRunner.runQuery().** The `stopSpy` is set up in `query-runner.test.ts` but no test asserts that `stop()` is called between the for-await loop exit and the finally block. The behavior is trivially correct (synchronous call, idempotent method), and the stale-state fallback test covers the defensive path, so this is acceptable.

2. **The `room-runtime-service.ts` changes in this PR are substantial and unrelated to the race fix** (MCP server merging, skills integration, session registration, model switching). These appear to be from an earlier feature branch that was rebased or merged into this branch. They are not problematic but make the PR diff larger than necessary for the described fix.

3. **Non-retryable patterns are string-matched.** This is fine for the current error messages from `room-runtime-service.ts` (line 391), but if the error phrasing changes in the future, the retry loop would attempt retries on what should be a permanent error. A more robust approach would be an error code enum, but that's out of scope here.

PR #1024, opened 2026-03-27
feat: add numeric task IDs scoped per space

lsm:
## Summary

- Adds `task_number` column (INTEGER, UNIQUE per space) to `space_tasks` with migration 61, backfilling existing rows
- Auto-assigns via MAX+1 on creation; UUID `id` stays as internal PK for FK references
- Adds `getTaskByNumber(spaceId, taskNumber)` lookup at repository and manager layers
- Agent prompts and `get_task_detail` tool now reference tasks by number ("Task #5")
- Unit tests cover auto-increment, space-scoped uniqueness, gap behavior, and lookup by number

## Test plan

- [x] All 8218 daemon unit tests pass (0 failures)
- [x] Lint + format clean
- [x] Knip (dead code detection) clean

### DIFF
--- a/packages/daemon/src/lib/space/agents/custom-agent.ts
+++ b/packages/daemon/src/lib/space/agents/custom-agent.ts
@@ -244,7 +244,7 @@ export function buildCustomAgentTaskMessage(config: CustomAgentConfig): string {
 	const sections: string[] = [];
 
 	// Task context
-	sections.push(`## Task\n`);
+	sections.push(`## Task #${task.taskNumber}\n`);
 	sections.push(`**Title:** ${task.title}`);
 	sections.push(`**Description:** ${task.description}`);
 	if (task.priority) {
--- a/packages/daemon/src/lib/space/agents/task-agent.ts
+++ b/packages/daemon/src/lib/space/agents/task-agent.ts
@@ -331,7 +331,7 @@ export function buildTaskAgentSystemPrompt(context: TaskAgentContext): string {
 	);
 
 	// ---- Task context -------------------------------------------------------
-	sections.push(`\n## Task Details\n`);
+	sections.push(`\n## Task #${context.task.taskNumber} Details\n`);
 	sections.push(`**Title:** ${context.task.title}`);
 	sections.push(`**Priority:** ${context.task.priority}`);
 	sections.push(`**Status:** ${context.task.status}`);
@@ -363,7 +363,7 @@ export function buildTaskAgentInitialMessage(context: TaskAgentContext): string 
 
 	// ---- Task assignment -----------------------------------------------------
 	parts.push(
-		`## Task Assignment\n` +
+		`## Task #${context.task.taskNumber} Assignment\n` +
 			`\n` +
 			`You have been assigned the following task:\n` +
 			`\n` +
--- a/packages/daemon/src/lib/space/managers/space-task-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-task-manager.ts
@@ -79,6 +79,13 @@ export class SpaceTaskManager {
 	}
 
 	/**
+	 * Get a task by its space-scoped numeric ID (e.g. task #5)
+	 */
+	async getTaskByNumber(taskNumber: number): Promise<SpaceTask | null> {
+		return this.taskRepo.getTaskByNumber(this.spaceId, taskNumber);
+	}
+
+	/**
 	 * List tasks in this space
 	 */
 	async listTasks(includeArchived = false): Promise<SpaceTask[]> {
--- a/packages/daemon/src/lib/space/tools/space-agent-tools.ts
+++ b/packages/daemon/src/lib/space/tools/space-agent-tools.ts
@@ -331,12 +331,23 @@ export function createSpaceAgentToolHandlers(config: SpaceAgentToolsConfig) {
 		},
 
 		/**
-		 * Get the full detail of a task by ID.
+		 * Get the full detail of a task by UUID or by numeric task number (e.g. #5).
 		 */
-		async get_task_detail(args: { task_id: string }): Promise<ToolResult> {
-			const task = await taskManager.getTask(args.task_id);
+		async get_task_detail(args: { task_id?: string; task_number?: number }): Promise<ToolResult> {
+			let task = null;
+			if (args.task_number !== undefined) {
+				task = await taskManager.getTaskByNumber(args.task_number);
+			} else if (args.task_id) {
+				task = await taskManager.getTask(args.task_id);
+			} else {
+				return jsonResult({
+					success: false,
+					error: 'Either task_id or task_number is required',
+				});
+			}
 			if (!task) {
-				return jsonResult({ success: false, error: `Task not found: ${args.task_id}` });
+				const ref = args.task_number !== undefined ? `#${args.task_number}` : args.task_id;
+				return jsonResult({ success: false, error: `Task not found: ${ref}` });
 			}
 			return jsonResult({ success: true, task });
 		},
@@ -588,9 +599,13 @@ export function createSpaceAgentMcpServer(config: SpaceAgentToolsConfig) {
 		),
 		tool(
 			'get_task_detail',
-			'Get the full detail of a task by ID, including error, result, PR URL, PR number, progress, and current step.',
+			'Get the full detail of a task by its numeric ID (e.g. task #5) or UUID. Includes error, result, PR URL, PR number, progress, and current step.',
 			{
-				task_id: z.string().describe('ID of the task to retrieve'),
+				task_id: z.string().optional().describe('UUID of the task to retrieve'),
+				task_number: z
+					.number()
+					.optional()
+					.describe('Numeric task ID (e.g. 5 for task #5) — preferred over task_id'),
 			},
 			(args) => handlers.get_task_detail(args)
 		),
--- a/packages/daemon/src/storage/repositories/space-task-repository.ts
+++ b/packages/daemon/src/storage/repositories/space-task-repository.ts
@@ -24,43 +24,49 @@ export class SpaceTaskRepository {
 		const id = generateUUID();
 		const now = Date.now();
 
-		// Auto-assign task_number: MAX+1 scoped to this space, inside the same write.
-		// SQLite serialises writes, so MAX+1 is safe without an explicit transaction here.
-		const nextNumber = (
+		// Wrap SELECT MAX + INSERT in an explicit transaction to prevent concurrent
+		// requests from computing the same task_number. SQLite serialises write
+		// transactions, so the UNIQUE index is the safety net, but the transaction
+		// ensures correctness without relying on constraint errors.
+		const insertTx = this.db.transaction(() => {
+			const nextNumber = (
+				this.db
+					.prepare(
+						`SELECT COALESCE(MAX(task_number), 0) + 1 AS next FROM space_tasks WHERE space_id = ?`
+					)
+					.get(params.spaceId) as { next: number }
+			).next;
+
 			this.db
 				.prepare(
-					`SELECT COALESCE(MAX(task_number), 0) + 1 AS next FROM space_tasks WHERE space_id = ?`
-				)
-				.get(params.spaceId) as { next: number }
-		).next;
-
-		const stmt = this.db.prepare(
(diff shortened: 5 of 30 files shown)


lsm (commented):

### P1-1: Migration does NOT enforce NOT NULL on `task_number`

The task description explicitly states "`task_number` column (INTEGER NOT NULL)", but `runMigration61` (line 3800 of `migrations.ts`) adds the column as nullable:

```sql
ALTER TABLE space_tasks ADD COLUMN task_number INTEGER
```

SQLite's `ALTER TABLE ADD COLUMN` cannot add a NOT NULL column without a default. The backfill populates existing rows, but **nothing prevents a direct SQL insert from inserting NULL** — the DB allows it, `rowToSpaceTask` produces `null` for a field typed as `number`, and the TypeScript type becomes incorrect at runtime.

**Fix**: Either update the task description to clarify that NOT NULL is application-enforced only, or rebuild the table with a NOT NULL constraint (like other migrations in this file do).

---

### P1-2: Race condition on MAX+1 auto-increment between separate statements

`SpaceTaskRepository.createTask` (lines 29-35) uses two separate statements:

```typescript
const nextNumber = (this.db.prepare(`SELECT COALESCE(MAX(task_number), 0) + 1 ...`).get(params.spaceId)).next;
// ... separate statement ...
stmt.run(id, params.spaceId, nextNumber, ...);
```

The comment claims "SQLite serialises writes, so MAX+1 is safe without an explicit transaction here." This guarantee **only holds within a single connection's write transaction**. Two concurrent requests can interleave: both read MAX=n, both compute n+1, both try to insert — one fails with a raw SQLite constraint error.

**Fix**: Wrap the SELECT+INSERT in an explicit `db.transaction()` callback. Alternatively, handle the conflict gracefully with `INSERT ... ON CONFLICT` against the unique index.

---

### P1-3: `rowToSpaceTask` unsafe cast of nullable column

```typescript
taskNumber: row.task_number as number,
```

If `task_number` is ever NULL (which is possible since the column is not NOT NULL enforced — see P1-1), this returns `null` at runtime but the TypeScript type claims `number`. Silent wrong data.

**Fix**: Add a null check or use a default: `taskNumber: (row.task_number as number | null) ?? 0`.

---

### P2-1: No test for concurrent task creation

The test suite covers auto-increment, space-scoped uniqueness, gap behavior, and constraint enforcement — but **no concurrent creation test**. The race in P1-2 would be caught by `Promise.all([repo.createTask(...), repo.createTask(...)])`.

---

### P2-2: Test helper defines `task_number INTEGER` without NOT NULL

In `packages/daemon/tests/unit/helpers/space-test-db.ts` line 134, `task_number INTEGER` is nullable. This means unit tests don't catch the NULL scenario that the migration technically allows.

---

### Positive observations

- `getTaskByNumber(spaceId, taskNumber)` correctly scoped to space at both repo and manager layers
- Agent prompts correctly reference `Task #${taskNumber}` for human readability  
- `get_task_detail` tool accepts `task_number` as alternative to `task_id`
- Backfill uses `ORDER BY created_at ASC` — stable and correct
- UNIQUE index `idx_space_tasks_space_task_number` created correctly
- `list_tasks` returns full task objects including `taskNumber`

---

### Verdict

Core functionality is correct. However, **P1-1 (NOT NULL mismatch), P1-2 (race condition), and P1-3 (unsafe null cast)** should be addressed before merge. P1-2 in particular can produce user-visible errors under concurrent load.

lsm (commented):

### Summary

The implementation correctly adds task_number with space-scoped auto-increment and proper uniqueness constraints. Agent prompts, the MCP tool, and the repository/manager layers are all consistent. However, three P1 issues must be addressed before merge.

---

### P1-1: NOT NULL is application-enforced, not schema-enforced

**File**: packages/daemon/src/storage/schema/migrations.ts:3800

The PR description says "task_number column (INTEGER NOT NULL)", but the migration adds it as nullable:

ALTER TABLE space_tasks ADD COLUMN task_number INTEGER

SQLite's ALTER TABLE ADD COLUMN cannot add a NOT NULL column without a default. While the backfill ensures all existing rows get a value, and createTask always inserts with a non-NULL value, the column can legally be NULL per the schema. The UNIQUE index (space_id, task_number) would also permit multiple NULL values per space (SQL treats NULL as != NULL in unique constraints).

Fix: Either rebuild the table with a NOT NULL constraint (matching the pattern of M51, M53, M60 in the same file), or update the PR description to accurately state the column is application-level NOT NULL.

---

### P1-2: Two-statement MAX+1 has a race window under concurrent load

**File**: packages/daemon/src/storage/repositories/space-task-repository.ts:29-35

The comment says "SQLite serialises writes". That is true at the database level for a single connection, but SQLite's lock is acquired at the start of a write operation and released after each statement. Two concurrent requests on the same connection can interleave between the SELECT and the INSERT — both read MAX=n, both compute n+1, both try to INSERT with the same task_number, and one fails with a raw SQLite constraint error.

The code does not use an explicit transaction. The serialization comment is only valid within a single write transaction.

Fix: Wrap the SELECT+INSERT in an explicit db.transaction() callback.

---

### P1-3: rowToSpaceTask casts NULL to number silently

**File**: packages/daemon/src/storage/repositories/space-task-repository.ts:377

If task_number is ever NULL (which the nullable column allows — see P1-1), this returns null at runtime but the TypeScript type claims number. This corrupts data silently.

Fix: Add a defensive cast or default.

---

### P1-4: No test for concurrent task creation

**File**: packages/daemon/tests/unit/storage/space-task-repository.test.ts

The test suite covers auto-increment, space-scoped uniqueness, gap behavior, and constraint enforcement — but there is no concurrent creation test. The race in P1-2 would surface under true concurrency.

Fix: Add a test that creates multiple tasks concurrently within the same space.

---

### P2-1: Backfill not wrapped in a transaction

**File**: packages/daemon/src/storage/schema/migrations.ts:3802-3815

The backfill loop runs each UPDATE as a separate statement without a wrapping transaction. If the process crashes mid-backfill, some spaces' tasks would be numbered and others wouldn't. Lower severity than P1 issues since the idempotent guard prevents re-run issues and crash during migration is rare.

Fix: Wrap in db.exec('BEGIN') / db.exec('COMMIT') around the backfill loop.

---

### Positive observations

- getTaskByNumber(spaceId, taskNumber) correctly scoped at repo and manager layers
- Agent prompts correctly display Task #N for human readability
- get_task_detail MCP tool accepts task_number as alternative to task_id, with task_number taking precedence
- Backfill orders by created_at ASC — stable and deterministic
- UNIQUE index idx_space_tasks_space_task_number properly created after backfill
- CreateSpaceTaskParams does NOT expose taskNumber — auto-assignment is correctly enforced
- 13 new tests cover auto-increment, space scoping, gaps, uniqueness constraint, and lookup

---

### Verdict

Core functionality is sound. P1-1 (NOT NULL), P1-2 (race condition), P1-3 (unsafe null cast), and P1-4 (missing concurrent test) must be addressed before merge. P1-2 in particular will produce visible constraint errors under concurrent task creation.

lsm (commented):

### Assessment of Fixes

**P1-1 (NOT NULL via table rebuild):** Migration 61 correctly implements the rename pattern:
1. Adds nullable `task_number` column via `ALTER TABLE` (line 3803)
2. Backfills existing rows per-space ordered by `created_at` (lines 3806-3817)
3. Creates new table `space_tasks_m61_new` with `task_number INTEGER NOT NULL` inline (line 3828)
4. Inserts data from old table, drops old table, renames new table (lines 3881-3882)
5. Recreates all indexes including `CREATE UNIQUE INDEX idx_space_tasks_space_task_number ON space_tasks(space_id, task_number)` (line 3899-3900)

The `NOT NULL` constraint is enforced at the DB level. The UNIQUE index on `(space_id, task_number)` is properly created after the rename. Looks correct.

**P1-2 (Race condition via transaction):** `createTask()` at line 31-69 wraps `SELECT MAX(task_number) + 1` and `INSERT` inside `db.transaction()`. The transaction is called synchronously at line 69. SQLite serializes write transactions, so concurrent callers will queue correctly. The approach is sound.

**P1-3 (NULL coalescing):** `rowToSpaceTask` at line 383 uses `(row.task_number as number | null) ?? 0`. Correct.

**P2-2 (Test helper schema):** `createSpaceTables()` in `space-test-db.ts` now includes `task_number INTEGER NOT NULL` and the `idx_space_tasks_space_task_number` unique index. Correct.

### Remaining Issue

**P2 (should-fix): The concurrent test is not actually concurrent.**

In `space-task-repository.test.ts` lines 574-591, `Array.from().map()` is synchronous. This test fires 20 *sequential* calls, not concurrent ones. It does **not** exercise the race condition that `db.transaction()` is meant to protect against.

A true concurrent test would use `Promise.all` with `createTask` calls.

This is a P2 (test gap, not a correctness bug) because the `db.transaction()` wrapping and `UNIQUE` index together guarantee correctness in production.

### Summary

All P1 fixes are correctly implemented. One P2 test gap remains: the concurrent test does not use `Promise.all` and thus does not truly test concurrent writes. The protection is in place (transaction + unique index) but not exercised by the test.

**Recommendation:** APPROVE with P2 suggestion to use `Promise.all` in the concurrent test.

lsm (commented):

Reviewed the five fixes pushed in response to the previous review. All previously identified issues are now resolved.

---

### P1-1: NOT NULL schema (RESOLVED)

Migration 61 now uses a full table rebuild to add `task_number` as `INTEGER NOT NULL`, exactly matching the pattern from M60. The flow is:
1. `ALTER TABLE ADD COLUMN task_number INTEGER` (nullable, temporary)
2. Backfill per-space sequential numbering via `ORDER BY created_at ASC`
3. Create `space_tasks_m61_new` with `task_number INTEGER NOT NULL` constraint
4. `INSERT INTO ... SELECT` copies all 29 columns including `active_session`
5. `DROP TABLE space_tasks; ALTER TABLE space_tasks_m61_new RENAME TO space_tasks`
6. Recreate all indexes including `UNIQUE(space_id, task_number)`

Column list in `CREATE TABLE space_tasks_m61_new` (line 3825-3868) correctly matches the `INSERT ... SELECT` column list (line 3870-3878) including `active_session TEXT`. `PRAGMA foreign_keys = OFF` is set before DDL, matching established migration patterns.

---

### P1-2: Race condition (RESOLVED)

`createTask()` at `space-task-repository.ts:31` now wraps the SELECT MAX + INSERT inside `db.transaction(() => { ... })`. SQLite serialises write transactions, so two concurrent calls cannot interleave the SELECT and INSERT. The UNIQUE index `idx_space_tasks_space_task_number` is an additional safety net.

---

### P1-3: Unsafe NULL cast (RESOLVED)

`rowToSpaceTask` at line 383 now uses `(row.task_number as number | null) ?? 0`. Given that M61 enforces NOT NULL at the schema level, this is purely defensive.

---

### P2-1: Concurrent test (RESOLVED — with note)

The test at `space-task-repository.test.ts:574-590` fires 20 rapid creates. Since `repo.createTask()` is synchronous and Bun test runs single-threaded, these execute sequentially rather than in parallel. However, this is acceptable because:
- The `db.transaction()` wrapper guarantees atomicity
- SQLite's write serialization makes interleaving impossible even with true concurrency at the application level
- The test correctly verifies uniqueness and monotonicity

True concurrent stress testing (multiple connections) belongs in integration tests, not unit tests. This is a P3 observation, not a blocker.

---

### P2-2: Test helper (RESOLVED)

`space-test-db.ts:134` now defines `task_number INTEGER NOT NULL` matching production schema, and creates the `UNIQUE` index.

---

### Positive observations

- All 13 new tests cover: auto-increment, space-scoped independence, gap behavior, UNIQUE constraint enforcement, `getTaskByNumber`, and rapid sequential creation
- `get_task_detail` tool correctly accepts `task_number` with precedence over `task_id` and provides an informative error message when neither is provided
- Agent prompts consistently display "Task #N" for human readability
- `SpaceTask` interface has `taskNumber: number` as a required field, matching the NOT NULL schema
- M61 is idempotent (guard at line 3797) and the entire rebuild is wrapped in `BEGIN/COMMIT/ROLLBACK` with `PRAGMA foreign_keys`
- No remaining P0 or P1 issues

---

### Verdict

All previously identified P1 and P2 issues have been addressed. **APPROVE** — the implementation is correct and ready for merge.

PR #1025, opened 2026-03-27
feat(space): implement unified gate evaluator (M1.2)

lsm:
## Summary

- Enhanced `GateConditionCheck` with `op` field supporting `exists`, `==`, `!=` operators (defaults to `==` for backward compatibility)
- Replaced `GateConditionCount` threshold semantics with map-counting: reads `data[field]` as a Record, counts entries matching `matchValue`, checks `>= min` — designed for multi-reviewer approval gates
- Unified `evaluateGate(gate)` to read from `gate.data` directly instead of taking a separate data parameter
- Added `isChannelOpen(channel, gates)` helper: gateless channels always open, gated channels delegate to `evaluateGate()`
- Updated runtime validation (`validateGateCondition`) for new fields
- 70 unit tests covering all condition types, ops, composites, edge cases, and gateless channels

### DIFF
--- a/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
+++ b/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
@@ -69,6 +69,10 @@ export function validateGateCondition(condition: unknown, path = 'condition'): s
 			if (cond.op !== undefined && (typeof cond.op !== 'string' || !VALID_CHECK_OPS.has(cond.op))) {
 				errors.push(`${path}.op: expected one of [exists, ==, !=], got ${JSON.stringify(cond.op)}`);
 			}
+			// Warn when value is set but op is 'exists' (value is ignored by exists)
+			if (cond.op === 'exists' && cond.value !== undefined) {
+				errors.push(`${path}: "value" is set but ignored when op is "exists"`);
+			}
 			break;
 
 		case 'count':
@@ -78,7 +82,9 @@ export function validateGateCondition(condition: unknown, path = 'condition'): s
 			if (typeof cond.min !== 'number') {
 				errors.push(`${path}.min: expected number, got ${typeof cond.min}`);
 			}
-			// matchValue can be anything — no validation needed
+			if (cond.matchValue === undefined) {
+				errors.push(`${path}.matchValue: required but missing`);
+			}
 			break;
 
 		case 'all':
--- a/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
+++ b/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
@@ -1,22 +1,22 @@
 /**
- * GateEvaluator — evaluates GateCondition trees against gate data.
+ * Unified Gate Evaluator (M1.2)
  *
- * This evaluator works with the new separated Gate/Channel system (M1.1).
- * It reads gate data from the data store and evaluates the gate's condition
- * tree to determine whether the gate is open or closed.
+ * Single `evaluateGate(gate)` function that handles all four condition types
+ * (check, count, all, any) including recursive all/any.
  *
- * Since conditions are deserialized from JSON at runtime, the evaluator
- * includes runtime validation via `validateGateCondition` to catch malformed
- * objects before they reach the evaluation logic.
+ * The evaluator reads from the gate's own `data` store — callers populate
+ * `gate.data` with runtime data from the `gate_data` table before evaluation.
  *
  * Condition types:
- *   check — field equality: data[field] === value
- *   count — numeric threshold: data[field] >= threshold
+ *   check — field check with operator: exists / == / !=
+ *   count — map-counting: count entries matching matchValue, check >= min
  *   all   — composite AND: all sub-conditions must pass (short-circuits on failure)
  *   any   — composite OR: at least one sub-condition must pass (short-circuits on success)
+ *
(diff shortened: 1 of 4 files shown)


lsm (commented):

---

## Verdict: APPROVE

Solid implementation of the M1.2 unified gate evaluator. The code is clean, well-documented, and thoroughly tested.

## What is Good

- **Clean architecture**: The separation between `evaluateGate`, `evaluateCondition`, `isChannelOpen`, and `validateGateCondition` is logical and composable. Each function has a single responsibility.
- **Backward compatibility**: `op` defaults to `==` and `value` is optional, so existing conditions without an `op` field continue to work correctly.
- **Fail-closed semantics**: `isChannelOpen` returns closed when a referenced gate is missing from the map -- this is the right security posture for a misconfiguration.
- **Defensive runtime handling**: The `default` branch in `evaluateCondition` handles malformed JSON gracefully with descriptive error messages.
- **Map-counting design**: Replacing the simple `threshold` with `matchValue`/`min` map-counting is well-suited for multi-reviewer approval gates. The edge cases (null, array, non-object) are all handled correctly.
- **Test coverage**: 70 tests across all condition types, operators, composite nesting depths, and edge cases. No gaps in coverage.
- **Full suite passes**: All 8806 daemon tests pass, typecheck clean.

## Minor Observations (P2)

1. **`validateGateCondition` does not catch missing `matchValue` on count conditions** (`packages/daemon/src/lib/space/runtime/gate-evaluator.ts:74-81`): The type defines `matchValue: unknown` (required), but the validator only checks `min` is a number. If JSON deserialization produces a count condition without `matchValue`, it will silently evaluate as `count === 0` (since `undefined === undefined` is false for all entries). Consider adding a validation check like `if (cond.matchValue === undefined)` to catch this early. Low severity since it only affects malformed data from untrusted sources.

2. **`validateGateCondition` does not warn about `value` with `op: exists`** (`packages/daemon/src/lib/space/runtime/gate-evaluator.ts:64-71`): A condition like `{ type: "check", field: "plan", op: "exists", value: "something" }` would pass validation even though `value` is ignored by the `exists` operator. A warning-level check (not a hard error) would help catch user mistakes in condition definitions.

Both of these are low-risk since the evaluator handles them correctly at runtime, but catching them in validation would produce better error messages for condition authors.

lsm (commented):

### Re-Review of Latest Commit (55e5f9886)

The latest commit addresses all P2 feedback from the first review round. The changes are well-implemented.

**Changes Reviewed:**

1. **Validation for missing `matchValue` in count conditions** (gate-evaluator.ts:85-87)
   - Correctly rejects `{ type: 'count', ... }` without `matchValue`
   - Error message is clear: `condition.matchValue: required but missing`

2. **Warning for `value` with `op: 'exists'`** (gate-evaluator.ts:72-75)
   - Properly warns when `value` is set alongside `exists` op since it's ignored
   - Uses the errors array consistently with the function's design

3. **Test coverage** (gate-evaluator.test.ts)
   - 3 new tests added, bringing total to 73 tests (all passing)
   - Covers both positive and negative cases appropriately

**Verdict: APPROVE**

The implementation is correct, well-tested, and fully addresses the P2 feedback. No issues found.

---REVIEW_POSTED---
url: https://github.com/lsm/neokai/pull/1025#pullrequestreview-2992680699
recommendation: APPROVE
p0: 0
p1: 0
p2: 0
p3: 0
summary: The latest commit correctly implements all P2 feedback with proper validation checks and comprehensive test coverage. 73 tests pass.
---END_REVIEW_POSTED---

lsm (commented):

### Re-review of commit 55e5f9886 (P2 feedback fixes)

The three new changes in this commit are all correct and well-implemented:

**1. `matchValue` required validation for count conditions** (`gate-evaluator.ts:85-87`)
Good catch. The `matchValue` field was previously commented as "can be anything -- no validation needed", but without it the `evaluateCount` function would count all entries where `val === undefined`, which is almost certainly not what the caller intended. The error message is clear.

**2. Warning for `value` set alongside `op: 'exists'`** (`gate-evaluator.ts:72-75`)
Correct and helpful. The `evaluateCheck` function ignores `value` when `op === 'exists'`, so silently accepting it would be confusing. The validation catches this at the configuration boundary. Note that this is reported as an error (in the `errors` array), not a warning -- that is fine since callers already treat any non-empty errors array as invalid.

**3. Three new test cases** (`gate-evaluator.test.ts:781-810`)
All three tests are well-structured:
- Rejects count with missing `matchValue` -- validates the new validation rule
- Warns when check has `value` with `exists` op -- validates the new warning
- No warning when check has `exists` op without `value` -- confirms the happy path still passes

All 73 tests pass.

### Remaining observations (informational, no action needed)

- The `evaluateCount` function itself still safely handles `matchValue === undefined` at runtime (line 274: `val === matchValue`), so even if `validateGateCondition` is not called before `evaluateCondition`, the evaluator won't crash -- it will just produce an unexpected result. This is acceptable since the validation function exists to catch this at configuration time.
- The term "warning" in the commit message and comments is slightly misleading since the check produces an error string in the `errors` array (not a separate warning category), but this is purely cosmetic and has no functional impact.

**Verdict: APPROVE** -- The P2 feedback has been fully addressed. The validation additions are correct, well-tested, and improve the robustness of the gate evaluator at the configuration boundary.

PR #1026, opened 2026-03-27
feat(space): worktreeSlug() utility wrapping shared slugify()

lsm:
Implements `worktreeSlug(taskTitle, taskNumber, existingSlugs)` in `packages/daemon/src/lib/space/worktree-slug.ts`.

- Delegates to `slugify()` from `./slug.ts` (M1.6) for core slug generation
- Falls back to `task-{taskNumber}` when title is empty, whitespace-only, or produces no usable characters
- Collision suffix (`-2`, `-3`, …) applied to both normal slugs (via `slugify()`) and fallback slugs
- 18 unit tests covering all specified behaviors

### DIFF
--- a/packages/daemon/src/lib/space/slug.ts
+++ b/packages/daemon/src/lib/space/slug.ts
@@ -96,7 +96,7 @@ function truncateAtWordBoundary(slug: string, maxLength: number): string {
 /**
  * Resolve slug collisions by appending a numeric suffix (-2, -3, ...).
  */
-function resolveCollision(base: string, existingSlugs: string[]): string {
+export function resolveCollision(base: string, existingSlugs: string[]): string {
 	const slugSet = new Set(existingSlugs);
 
 	if (!slugSet.has(base)) {
--- a/packages/daemon/src/lib/space/worktree-slug.ts
+++ b/packages/daemon/src/lib/space/worktree-slug.ts
@@ -1,0 +1,58 @@
+import { slugify } from './slug';
+
+/**
+ * Generate a slug for a worktree folder name and git branch.
+ *
+ * Delegates to `slugify()` for core slugification rules.
+ * If the task title is empty/whitespace-only or produces an empty slug,
+ * falls back to `task-{taskNumber}`.
+ *
+ * The returned slug is used as:
+ * - The worktree folder name (as-is)
+ * - The git branch name (prefixed with `space/`)
+ *
+ * @param taskTitle - The task title to slugify
+ * @param taskNumber - The numeric task ID for fallback naming
+ * @param existingSlugs - Slugs already in use, for collision avoidance
+ * @returns A unique, URL-safe slug
+ */
+export function worktreeSlug(
+	taskTitle: string,
+	taskNumber: number,
+	existingSlugs: string[] = []
+): string {
+	const trimmed = taskTitle.trim();
+
+	if (!trimmed) {
+		return resolveWorktreeCollision(`task-${taskNumber}`, existingSlugs);
+	}
+
+	const slug = slugify(trimmed, existingSlugs);
+
+	// slugify() falls back to 'unnamed-space' for empty input; treat that as a
+	// signal that the title produced no usable characters.
+	if (!slug || slug === 'unnamed-space') {
+		return resolveWorktreeCollision(`task-${taskNumber}`, existingSlugs);
+	}
+
+	return slug;
+}
+
+/**
+ * Resolve collisions for the fallback `task-{taskNumber}` slug.
+ * Appends a numeric suffix (-2, -3, …) if the base is already taken.
+ */
+function resolveWorktreeCollision(base: string, existingSlugs: string[]): string {
+	const slugSet = new Set(existingSlugs);
+	if (!slugSet.has(base)) {
+		return base;
+	}
+	let counter = 2;
+	while (true) {
+		const suffixed = `${base}-${counter}`;
+		if (!slugSet.has(suffixed)) {
+			return suffixed;
+		}
+		counter++;
+	}
+}
--- a/packages/daemon/src/lib/space/worktree-slug.ts
+++ b/packages/daemon/src/lib/space/worktree-slug.ts
@@ -1,11 +1,11 @@
-import { slugify } from './slug';
+import { slugify, resolveCollision } from './slug';
 
 /**
  * Generate a slug for a worktree folder name and git branch.
  *
  * Delegates to `slugify()` for core slugification rules.
- * If the task title is empty/whitespace-only or produces an empty slug,
- * falls back to `task-{taskNumber}`.
+ * If the task title contains no alphanumeric characters (empty, whitespace-only,
+ * or all-special-chars), falls back to `task-{taskNumber}`.
  *
  * The returned slug is used as:
  * - The worktree folder name (as-is)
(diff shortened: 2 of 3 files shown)


lsm (commented):

### Overall

Clean implementation that correctly wraps `slugify()` with worktree-specific fallback behavior. Tests are well-organized and cover the happy path thoroughly. However, there is a meaningful bug in the `unnamed-space` detection logic.

### P1: `unnamed-space` check is fragile and has two failure modes

**File:** `packages/daemon/src/lib/space/worktree-slug.ts`, line 34

The check `slug === 'unnamed-space'` has two problems:

**1. False negative when `existingSlugs` contains `'unnamed-space'`:**
When a title produces no usable characters (e.g., `'!!!'`) and `existingSlugs` already contains `'unnamed-space'`, `slugify()` returns `'unnamed-space-2'` (collision-resolved). The exact equality check misses this, so `worktreeSlug('!!!', 7, ['unnamed-space'])` returns `'unnamed-space-2'` instead of `'task-7'`.

**2. False positive for legitimate titles:**
A title like `'Unnamed Space'` legitimately slugifies to `'unnamed-space'`, but the check would incorrectly treat it as a "no usable characters" signal and fall back to `task-{taskNumber}`.

**Suggested fix:** Rather than coupling to `slugify()`'s internal default constant, detect "no usable characters" independently. For example, check whether the trimmed title contains any alphanumeric character (`/[a-z0-9]/i.test(trimmed)`) before calling `slugify()`. If it does not, go straight to the fallback. This avoids both failure modes and removes the dependency on knowing what `slugify()` returns for degenerate input.

### P3: `resolveWorktreeCollision` duplicates `resolveCollision` from `slug.ts`

**File:** `packages/daemon/src/lib/space/worktree-slug.ts`, lines 45-58

The `resolveWorktreeCollision` function is identical to `resolveCollision` in `slug.ts`. Consider exporting `resolveCollision` from `slug.ts` and reusing it. This is minor since the function is small and self-contained.

lsm (commented):

> Recommendation: REQUEST_CHANGES

---

### P1 Bug — False fallback trigger when task title is literally "unnamed-space"

**File:** `packages/daemon/src/lib/space/worktree-slug.ts`, line 34

The sentinel-value check uses the output of `slugify()` to detect whether the title produced no usable characters:

```ts
if (!slug || slug === 'unnamed-space') {
    return resolveWorktreeCollision(`task-${taskNumber}`, existingSlugs);
}
```

The problem: `'unnamed-space'` is also the valid slug output for any title that legitimately slugifies to that string — for example:

- Title `"unnamed-space"` → `slugify()` returns `'unnamed-space'` (valid slug from real content) → incorrectly falls back to `task-{taskNumber}`
- Title `"Unnamed Space"` → `slugify()` returns `'unnamed-space'` → same incorrect fallback

Reproduction:

```ts
worktreeSlug('unnamed-space', 5, []) // returns 'task-5', should return 'unnamed-space'
worktreeSlug('Unnamed Space', 5, []) // returns 'task-5', should return 'unnamed-space'
```

The sentinel value `'unnamed-space'` was designed by `slug.ts` for the case of empty or entirely-non-alphanumeric input. It is a convention internal to `slug.ts`, not a guaranteed signal that the title was empty. The check should instead be done on the trimmed input _before_ calling `slugify()` — or alternatively by checking whether the input contained any alphanumeric characters, rather than testing the output string value.

No tests cover this case (the test suite only covers all-special-chars titles like `'!!!@@@'`, not titles that produce the same sentinel string legitimately).

---

### P2 — `resolveWorktreeCollision` duplicates `resolveCollision` from `slug.ts`

**File:** `packages/daemon/src/lib/space/worktree-slug.ts`, lines 45–58

`resolveWorktreeCollision` is an exact copy of the private `resolveCollision` function inside `slug.ts`. Since `slugify(base, existingSlugs)` already performs the same logic correctly (confirmed: `slugify('task-2', ['task-2'])` → `'task-2-2'`), the fallback path could call `slugify` directly and avoid the duplication. If exporting the collision helper is undesirable, this is still worth noting as the two implementations can drift.

---

### P3 — Test suite does not cover the legitimate `'unnamed-space'` input title

As mentioned above, there is no test for `worktreeSlug('unnamed-space', 5)`, which would expose the bug described above. The existing tests for the fallback path only cover special-char inputs, not titles that happen to produce the sentinel string.

---

### Summary

The implementation is correct for the common cases and the test suite is thorough for those cases. The P1 issue is that the sentinel-value check on `slugify()`'s output incorrectly swallows legitimate titles that happen to produce `'unnamed-space'` as their slug. The fix is to check whether the raw title yields any alphanumeric content _before_ delegating to `slugify()`, rather than testing the output against an internal constant borrowed from another module.

lsm (commented):

Re-review after feedback was addressed. Both previously raised issues are resolved:

**P1 (fixed):** The `slug === 'unnamed-space'` sentinel check has been replaced with a pre-call alphanumeric test (`!/[a-z0-9]/i.test(taskTitle)`). This correctly handles titles like "Unnamed Space" without false positives, and ensures all-special-chars titles fall back to `task-N` even when `unnamed-space` is already in `existingSlugs`. Two regression tests confirm both scenarios.

**P3 (fixed):** The duplicated `resolveWorktreeCollision()` function has been removed. `resolveCollision` is now exported from `slug.ts` and reused in `worktree-slug.ts`.

**Current state:**

- `worktree-slug.ts` (32 lines) is clean and focused -- imports `slugify` and `resolveCollision`, applies a single guard, delegates appropriately.
- The only change to `slug.ts` is promoting `resolveCollision` from private to exported (line 99), which is a safe, minimal change.
- 21 tests pass covering empty/whitespace/special-chars fallback, normal delegation, collision handling, task number variants, and both regression cases.
- The alphanumeric check on raw input is the right approach -- it decouples `worktreeSlug` from `slug.ts`'s internal `DEFAULT_SLUG` constant.

No remaining issues found. Recommendation: **APPROVE**.

lsm (commented):

**Recommendation: APPROVE**

Both previous issues are properly resolved. The implementation is clean and correct.

### Previous P1 — unnamed-space false positive/negative (RESOLVED)

The sentinel-value check is replaced with `!/[a-z0-9]/i.test(taskTitle)` at line 27. This correctly detects all no-alphanumeric inputs (empty string, whitespace-only, tab/newline, all-special-chars) without coupling to `slug.ts`'s internal `DEFAULT_SLUG` constant. The regression tests at lines 49–55 explicitly confirm that titles like `"unnamed-space"` and `"Unnamed Space"` are no longer treated as fallback triggers.

### Previous P2 — resolveCollision duplication (RESOLVED)

`resolveCollision` is now exported from `slug.ts` (line 99) and imported directly by `worktree-slug.ts` (line 1). The fallback path at line 28 (`resolveCollision(`task-${taskNumber}`, existingSlugs)`) reuses the shared function rather than duplicating the collision logic.

### Quality checks

- `bun run check` (lint + typecheck + knip) passes with zero warnings or errors.
- `resolveCollision` is consumed by both `slugify()` internally and by `worktree-slug.ts`, so the new export is not dead code.

### Test coverage

18 tests across four `describe` groups cover: empty/whitespace/all-special fallbacks, normal title delegation, collision suffix increments for both normal and fallback paths, regression for the `unnamed-space` sentinel issue, and task number variants (0, large numbers). Coverage is thorough for the specified behaviors.

- The `generateBaseSlug` path in `slug.ts` still returns `'unnamed-space'` for all-special-char inputs when called directly — but since `worktreeSlug` pre-screens the input before delegating, the sentinel path is never reached from this function, which is the correct design intent.
- No issues with error handling, architecture fit, or security surface.

PR #1027, opened 2026-03-27
feat(space): implement SpaceWorktreeManager (M4.2)

lsm:
## Summary

- `SpaceWorktreeRepository`: CRUD for `space_worktrees` SQLite table (task↔worktree mapping, unique per space by task_id and slug)
- Migration 64: creates `space_worktrees` table with `UNIQUE(space_id, task_id)` and `UNIQUE(space_id, slug)` constraints + index
- `SpaceWorktreeManager`: manages git worktrees for Space tasks via `simple-git` (does NOT extend Room's WorktreeManager)
  - `createTaskWorktree(spaceId, taskId, taskTitle, taskNumber, baseBranch?)` — idempotent, slug from `worktreeSlug()` with collision avoidance
  - `removeTaskWorktree(spaceId, taskId)` — removes worktree dir, branch, and DB record
  - `getTaskWorktreePath(spaceId, taskId)` — returns path or null
  - `listWorktrees(spaceId)` — lists all tracked worktrees for a space
  - `cleanupOrphaned(spaceId)` — removes DB records for missing dirs, runs `git worktree prune`
- Worktrees at `{spaceWorkspacePath}/.worktrees/{slug}/`, branch `space/{slug}`
- 18 unit tests covering full lifecycle, idempotency, collision avoidance, and orphan cleanup

### DIFF
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -5,6 +5,8 @@
 export { SpaceManager } from './managers/space-manager';
 export { SpaceWorktreeManager } from './managers/space-worktree-manager';
 export type { SpaceWorktreeInfo } from './managers/space-worktree-manager';
+export { SpaceWorktreeRepository } from '../../storage/repositories/space-worktree-repository';
+export type { SpaceWorktreeRecord } from '../../storage/repositories/space-worktree-repository';
 export {
 	SpaceTaskManager,
 	VALID_SPACE_TASK_TRANSITIONS,
--- a/packages/daemon/src/lib/space/index.ts
+++ b/packages/daemon/src/lib/space/index.ts
@@ -3,6 +3,8 @@
  */
 
 export { SpaceManager } from './managers/space-manager';
+export { SpaceWorktreeManager } from './managers/space-worktree-manager';
+export type { SpaceWorktreeInfo } from './managers/space-worktree-manager';
 export {
 	SpaceTaskManager,
 	VALID_SPACE_TASK_TRANSITIONS,
--- a/packages/daemon/src/lib/space/managers/space-worktree-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-worktree-manager.ts
@@ -1,0 +1,210 @@
+/**
+ * SpaceWorktreeManager
+ *
+ * Manages git worktrees for Space tasks. One worktree per task, created from
+ * the space's repository workspace path.
+ *
+ * Worktree location : {spaceWorkspacePath}/.worktrees/{slug}/
+ * Branch naming     : space/{slug}
+ *
+ * Does NOT extend Room's WorktreeManager — uses simple-git directly.
+ */
+
+import simpleGit from 'simple-git';
+import { join } from 'node:path';
+import { existsSync, mkdirSync } from 'node:fs';
+import type { Database as BunDatabase } from 'bun:sqlite';
+import { SpaceWorktreeRepository } from '../../../storage/repositories/space-worktree-repository';
+import { SpaceRepository } from '../../../storage/repositories/space-repository';
+import { worktreeSlug } from '../worktree-slug';
+import { Logger } from '../../logger';
+
+export interface SpaceWorktreeInfo {
+	slug: string;
+	taskId: string;
+	path: string;
+}
+
+export class SpaceWorktreeManager {
+	private worktreeRepo: SpaceWorktreeRepository;
+	private spaceRepo: SpaceRepository;
+	private logger = new Logger('SpaceWorktreeManager');
+
+	constructor(db: BunDatabase) {
+		this.worktreeRepo = new SpaceWorktreeRepository(db);
+		this.spaceRepo = new SpaceRepository(db);
+	}
+
+	/**
+	 * Create a git worktree for a task.
+	 *
+	 * If a worktree already exists for this (spaceId, taskId) pair the existing
+	 * record is returned without touching the filesystem — idempotent.
+	 *
+	 * @param spaceId     - Space UUID
+	 * @param taskId      - Task UUID
+	 * @param taskTitle   - Human-readable title used to derive the slug
+	 * @param taskNumber  - Numeric task ID (fallback for slug when title has no alphanumeric chars)
+	 * @param baseBranch  - Git ref to base the new branch on (default: 'HEAD')
+	 * @returns           - { path, slug } of the created (or existing) worktree
+	 */
+	async createTaskWorktree(
+		spaceId: string,
+		taskId: string,
+		taskTitle: string,
+		taskNumber: number,
+		baseBranch?: string
+	): Promise<{ path: string; slug: string }> {
+		const space = this.spaceRepo.getSpace(spaceId);
+		if (!space) {
+			throw new Error(`Space not found: ${spaceId}`);
+		}
+
+		// Idempotent: return existing record if the worktree was already created
+		const existing = this.worktreeRepo.getByTaskId(spaceId, taskId);
+		if (existing) {
+			return { path: existing.path, slug: existing.slug };
+		}
+
+		// Generate a slug that is unique within this space
+		const existingSlugs = this.worktreeRepo.listSlugs(spaceId);
+		const slug = worktreeSlug(taskTitle, taskNumber, existingSlugs);
+
+		// Ensure the .worktrees directory exists inside the workspace
+		const worktreesDir = join(space.workspacePath, '.worktrees');
+		if (!existsSync(worktreesDir)) {
+			mkdirSync(worktreesDir, { recursive: true });
+		}
+
+		const worktreePath = join(worktreesDir, slug);
+		const branchName = `space/${slug}`;
+
+		const git = simpleGit(space.workspacePath);
+
+		// If a stale branch with the same name exists (from a previous crashed run),
+		// delete it so we can recreate it cleanly.
+		try {
+			const branches = await git.raw(['branch', '--list', branchName]);
+			if (branches.trim().length > 0) {
+				this.logger.warn(`Stale branch detected: ${branchName} — deleting before recreating`);
+				await git.branch(['-D', branchName]);
+			}
+		} catch {
+			// Non-fatal: branch check/delete failure should not block worktree creation
+		}
+
+		try {
+			await git.raw(['worktree', 'add', worktreePath, '-b', branchName, baseBranch ?? 'HEAD']);
+		} catch (err) {
+			// Clean up the directory if it was partially created
+			if (existsSync(worktreePath)) {
+				try {
+					await git.raw(['worktree', 'remove', worktreePath, '--force']);
+				} catch {
+					// Ignore cleanup errors
+				}
+			}
+			throw new Error(
+				`Failed to create worktree for task ${taskId}: ${err instanceof Error ? err.message : String(err)}`
+			);
+		}
+
+		// Persist the mapping to SQLite after the filesystem operation succeeds
+		this.worktreeRepo.create({ spaceId, taskId, slug, path: worktreePath });
+
+		this.logger.info(
+			`Created worktree for task ${taskId} at ${worktreePath} (branch: ${branchName})`
+		);
+		return { path: worktreePath, slug };
+	}
+
+	/**
+	 * Remove a task's worktree from the filesystem and delete its branch.
+	 * Removes the SQLite record regardless of whether the filesystem operation succeeds.
+	 */
+	async removeTaskWorktree(spaceId: string, taskId: string): Promise<void> {
+		const record = this.worktreeRepo.getByTaskId(spaceId, taskId);
+		if (!record) {
+			return; // Nothing to do
+		}
+
+		const space = this.spaceRepo.getSpace(spaceId);
+		if (!space) {
+			// Space is gone; just clean up the DB record
+			this.worktreeRepo.delete(spaceId, taskId);
+			return;
+		}
+
+		const git = simpleGit(space.workspacePath);
+
+		// Remove the worktree directory via git
+		try {
+			await git.raw(['worktree', 'remove', record.path, '--force']);
+		} catch (err) {
+			this.logger.warn(
+				`Failed to remove git worktree at ${record.path} (continuing with cleanup): ${err instanceof Error ? err.message : String(err)}`
+			);
+		}
+
+		// Delete the branch
+		const branchName = `space/${record.slug}`;
+		try {
+			await git.branch(['-D', branchName]);
+		} catch {
+			// Branch may already be gone; non-fatal
+		}
+
+		// Remove the SQLite record
+		this.worktreeRepo.delete(spaceId, taskId);
+		this.logger.info(`Removed worktree for task ${taskId} (branch: ${branchName})`);
+	}
+
+	/**
+	 * Return the filesystem path for a task's worktree, or null if none exists.
+	 */
+	async getTaskWorktreePath(spaceId: string, taskId: string): Promise<string | null> {
+		const record = this.worktreeRepo.getByTaskId(spaceId, taskId);
+		return record?.path ?? null;
+	}
+
+	/**
+	 * List all worktrees tracked for a space.
+	 */
+	async listWorktrees(spaceId: string): Promise<SpaceWorktreeInfo[]> {
+		const records = this.worktreeRepo.listBySpace(spaceId);
+		return records.map((r) => ({ slug: r.slug, taskId: r.taskId, path: r.path }));
+	}
+
+	/**
+	 * Remove SQLite records whose worktree directories no longer exist on disk,
+	 * then prune git's internal worktree metadata.
+	 *
+	 * This is safe to call at any time — it only removes records for missing
+	 * directories and does not touch live worktrees.
+	 */
+	async cleanupOrphaned(spaceId: string): Promise<void> {
+		const records = this.worktreeRepo.listBySpace(spaceId);
+
+		for (const record of records) {
+			if (!existsSync(record.path)) {
+				this.worktreeRepo.delete(spaceId, record.taskId);
+				this.logger.info(
+					`Cleaned up orphaned worktree record for task ${record.taskId} (path was: ${record.path})`
+				);
+			}
+		}
+
+		// Prune git's internal worktree state as well
+		const space = this.spaceRepo.getSpace(spaceId);
+		if (space) {
+			try {
+				const git = simpleGit(space.workspacePath);
+				await git.raw(['worktree', 'prune']);
+			} catch (err) {
+				this.logger.warn(
+					`git worktree prune failed for space ${spaceId}: ${err instanceof Error ? err.message : String(err)}`
+				);
+			}
+		}
+	}
+}
--- a/packages/daemon/src/lib/space/managers/space-worktree-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-worktree-manager.ts
@@ -101,6 +101,15 @@ export class SpaceWorktreeManager {
 			);
 		}
 
+		// If a stale directory exists at the worktree path (from a previous crashed run),
+		// remove it so git worktree add can proceed cleanly.
+		if (existsSync(worktreePath)) {
+			this.logger.warn(
+				`Stale directory detected at ${worktreePath} — removing before recreating worktree`
+			);
+			rmSync(worktreePath, { recursive: true, force: true });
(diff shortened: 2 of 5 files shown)


lsm (commented):

**Recommendation: APPROVE**

Well-structured implementation of SpaceWorktreeManager. The code is clean, idempotent, and handles failure modes thoughtfully. The tests are thorough with real git repos rather than mocks — good choice for filesystem+git integration code.

### What looks good

- **Idempotent create**: `createTaskWorktree` returns existing record if already created — correct pattern for task-based worktrees.
- **Stale branch cleanup**: Handles the case where a previous crashed run left behind a branch but no DB record.
- **Graceful removal**: `removeTaskWorktree` deletes the DB record even if filesystem cleanup fails, preventing permanent stuck state.
- **`cleanupOrphaned`**: Nice complement — handles the inverse case where directories disappeared but DB records remain.
- **Migration 64**: Proper UNIQUE constraints on `(space_id, task_id)` and `(space_id, slug)`. Consistent with the codebase's pattern of no foreign keys.
- **Tests**: 18 tests covering create, idempotent create, slug collision, fallback slug, custom baseBranch, removal, listing, orphan cleanup, and cross-space isolation. All use real temporary git repos.
- **`.worktrees/` already in `.gitignore`**: No risk of worktree contents being committed.

### Minor observations (P3)

1. **`getTaskWorktreePath` and `listWorktrees` are async but do only synchronous DB reads** (`space-worktree-manager.ts:165`, `173`). They await nothing internally. This is not a bug — async is fine for API consistency and future-proofing — but it is slightly misleading. The existing `WorktreeManager` in the codebase also uses async for similar methods, so this is consistent.

2. **No FK from `space_worktrees.space_id` to `spaces.id`**: This is consistent with the rest of the codebase (zero FOREIGN KEY references across all 64 migrations). The manager-level code validates space existence before operating, which is sufficient.

3. **Race between `listSlugs()` and `create()` in `createTaskWorktree`** (`space-worktree-manager.ts:70-71`): If two concurrent calls for the same space generate the same slug, the `UNIQUE(space_id, slug)` constraint will reject the second insert. This would surface as an unhandled SQLite constraint error rather than a retry. In practice, this is unlikely in the single-threaded Bun runtime and the task-oriented usage pattern, but worth noting.

No blocking issues found. The implementation fulfills the stated goal completely.

lsm (commented):

Recommendation: **APPROVE** (with minor notes)

The implementation is clean, well-structured, and well-tested. The core task — one git worktree per space task, persisted in SQLite — is correctly achieved. I have no blocking issues, but there are a few things worth noting.

---

### What looks good

- Idempotency in `createTaskWorktree` is correct: early return on existing record prevents duplicate filesystem operations.
- The stale-branch cleanup before recreation (lines 86-94 of `space-worktree-manager.ts`) is a thoughtful defensive measure for crash-recovery scenarios.
- Partial-creation cleanup in the `git worktree add` catch block correctly attempts `worktree remove --force` before rethrowing.
- `removeTaskWorktree` removes the DB record regardless of whether the git/filesystem operations succeed — the right priority for cleanup operations.
- `cleanupOrphaned` correctly separates DB-record removal from `git worktree prune`, and handles a missing space gracefully.
- Tests use a real git repository (not mocks), which gives high confidence in the git integration paths.

---

### Issues

**P2 — Missing FK constraint on `space_id` in migration 64**

`packages/daemon/src/storage/schema/migrations.ts` (function `runMigration64`, around line 4232)

Every other space-scoped table (`space_tasks`, `space_workflow_runs`, `space_agents`, etc.) declares `FOREIGN KEY (space_id) REFERENCES spaces(id) ON DELETE CASCADE`. The new `space_worktrees` table does not. This means if a space row is deleted, its worktree records survive silently. Calling `cleanupOrphaned` won't help since it checks filesystem paths, not space existence. A future `listWorktrees(deletedSpaceId)` call would return stale rows with no parent space. The constraint costs nothing and the precedent is already set by all other tables in this schema.

**P2 — No FK constraint on `task_id` referencing `space_tasks`**

Similarly, there is no `FOREIGN KEY (task_id) REFERENCES space_tasks(id) ON DELETE CASCADE`. If a task is deleted its worktree record stays in the table. The `removeTaskWorktree` path only runs if callers explicitly invoke it, so task deletion won't auto-clean worktree records. This is consistent with the missing `space_id` FK, but both gaps compound each other.

**P1 — `SpaceWorktreeRepository` is not exported from any barrel**

`packages/daemon/src/storage/repositories/space-worktree-repository.ts` is not re-exported from any barrel file. Unlike `SpaceWorkflowRepository` (which is explicitly re-exported from `packages/daemon/src/lib/space/index.ts`), this repository can only be reached via a direct file-path import. Future consumers — RPC handlers, tests, integration points — will have to import by raw path, which deviates from every other repository in the project.

**P3 — Hardcoded slug expectations in `listWorktrees` test**

`packages/daemon/tests/unit/space/space-worktree-manager.test.ts`, lines 239-241:

```ts
const slugs = list.map((w) => w.slug).sort();
expect(slugs).toEqual(['task-a', 'task-b']);
```

The expected values are hardcoded output of `slugify`. If the slugification logic changes, this assertion will fail without a clear signal of what broke. Deriving expected values via `worktreeSlug()` in the test itself would make the coupling explicit.

**P3 — Comment typo in migrations.ts lines 246/252/257/261**

The migration block uses `/ Migration 6x:` (missing a `/`) for comments on the call-site. Pre-existing issue but touched in this PR.

---

### Summary

Solid implementation. The test suite (18 tests, real git repos) is thorough and the git integration logic is correctly ordered. The two missing FK constraints on `space_worktrees` are the main structural gap relative to how all other space tables are defined; those should be added in a follow-up at minimum.

lsm (commented):

Overall this is a clean, well-tested implementation. The repository, migration, manager, and tests are all coherent. I have one security issue and one robustness concern.

---

### P0: Shell injection via unquoted `baseBranch` parameter

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, line 97

```ts
execSync(`git worktree add "${worktreePath}" -b "${branchName}" ${baseBranch ?? 'HEAD'}`, {
```

`baseBranch` is interpolated into the shell command **without quotes**. While `worktreePath` and `branchName` are double-quoted, `baseBranch` is raw. A caller passing a value like `HEAD; rm -rf /` or `$(malicious-command)` would get arbitrary shell execution.

Even if the current callers are trusted, this is a public method on a class that accepts arbitrary strings. It should be quoted the same way as the other arguments:

```ts
execSync(`git worktree add "${worktreePath}" -b "${branchName}" "${baseBranch ?? 'HEAD'}"`, { ... })
```

Or better yet, use `execFileSync` which avoids shell interpretation entirely (passing arguments as an array). This would eliminate shell injection risk for all five git commands, not just this one. The double-quoted string interpolation approach (used for `worktreePath`, `branchName`) is also technically vulnerable to values containing `"`, `$`, or backticks -- though the slugify pipeline makes this unlikely for those specific values, it is still a fragile defense.

---

### P1: All methods are `async` but contain no `await` expressions

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, lines 51, 127, 165, 173, 185

Every public method (`createTaskWorktree`, `removeTaskWorktree`, `getTaskWorktreePath`, `listWorktrees`, `cleanupOrphaned`) is declared `async` but performs only synchronous operations (`execSync`, synchronous SQLite queries, `existsSync`). This means:

1. The `async` keyword is misleading -- callers see a `Promise` and may assume I/O is non-blocking, but `execSync` blocks the event loop.
2. Every call wraps the result in an unnecessary microtask.

This is not a bug per se (the interface works), but it is architecturally misleading. If the intent is to allow a future migration to async git operations, that is reasonable -- but it should be documented. Otherwise, dropping `async` and returning plain values would be more honest.

---

### P2: `cleanupOrphaned` does not delete the orphaned git branches

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, lines 186-207

When `cleanupOrphaned` finds a DB record whose directory is missing, it deletes the DB record and runs `git worktree prune`, but it does **not** delete the corresponding `space/{slug}` branch. The branch will linger in the repository. Compare with `removeTaskWorktree` which explicitly runs `git branch -D` after removing the worktree. The cleanup method should do the same for consistency.

---

### P3: Stale branch cleanup swallows errors silently

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, lines 84-94

The `catch {}` block around the stale-branch check/delete swallows all errors without logging. While the comment says "Non-fatal", it would be helpful to at least log a warning (similar to how other catch blocks in this class do). A failed `git branch --list` could indicate a corrupted repo or missing git binary, which is worth knowing about.

---

### Summary

The implementation is solid for the happy path. The migration, repository, and test coverage are thorough. The main concern is the shell injection on `baseBranch` (P0) which should be fixed before merge. Switching all `execSync` calls to `execFileSync` with array arguments would be the cleanest fix and would eliminate the entire class of shell injection risks.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P0 shell-injection bug and two P1 gaps need attention before merge.

---

### P0 — Shell injection via unquoted `baseBranch` parameter

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, line 97

```ts
execSync(`git worktree add "${worktreePath}" -b "${branchName}" ${baseBranch ?? 'HEAD'}`, {
    cwd: space.workspacePath,
});
```

`baseBranch` is interpolated directly into the shell string without quoting or escaping. `execSync` with a template-literal string runs through `/bin/sh -c` on Unix, so any shell metacharacter in the value is interpreted. A caller passing `baseBranch = 'main; rm -rf /'` would execute the second command. This is a real injection surface even if current callers are internal — the public API signature allows any string.

The worktreePath and branchName arguments are double-quoted, which protects paths with spaces, but `baseBranch` has no quoting at all.

The immediate fix is either:
- Add double-quotes around `${baseBranch ?? 'HEAD'}` (minimum), or
- Switch to `execFileSync('git', ['worktree', 'add', worktreePath, '-b', branchName, baseBranch ?? 'HEAD'], { cwd })` which bypasses the shell entirely and makes all three arguments safe — the same pattern the existing `WorktreeManager` uses via `git.raw([...])`.

The existing `WorktreeManager` in `packages/daemon/src/lib/worktree-manager.ts` correctly uses `simpleGit`'s array API (`git.raw(['worktree', 'add', worktreePath, '-b', branchName, baseBranch])`), which is injection-proof. This PR intentionally abandoned that approach to avoid the test-mock problem, but the safety property of array-based args is not optional — it should be preserved with `execFileSync`.

---

### P1 — No `execSync` timeout; git operations can hang indefinitely

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, lines 85, 90, 97, 142, 152, 201

None of the six `execSync` calls set a `timeout` option. If the working directory is on a slow or unresponsive filesystem (NFS, mounted network drive), or if git hangs waiting for a credential prompt, the process blocks forever. The daemon has no watchdog for this.

All calls should include a reasonable timeout, e.g. `{ cwd: ..., timeout: 30_000 }`. The existing test repo helper writes to disk synchronously so tests would not be affected.

---

### P1 — No test for `baseBranch` injection or invalid input

**File:** `packages/daemon/tests/unit/space/space-worktree-manager.test.ts`

The test at line 157–183 verifies `baseBranch` with a valid branch name only. There is no test that passes a branch name containing shell-special characters (spaces, semicolons, backticks) to confirm that the implementation is safe. Adding such a test would both document the contract and catch any future regression if the fix is later removed.

---

### P2 — `stale branch` silent catch discards the real error before `worktree add` fails

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, lines 84–94

When the stale-branch delete at line 90 fails (e.g., the branch is currently checked out somewhere), the exception is swallowed and execution proceeds to `git worktree add`, which then fails because the branch still exists. The outer catch at lines 100–112 will clean up the worktree directory and surface a generic error. The real cause (branch already checked out) is silently discarded. At minimum the inner catch should log the deletion failure before proceeding.

---

### P3 — `.toString()` call is redundant; prefer `encoding: 'utf8'` option

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, line 87

Passing `{ encoding: 'utf8' }` to `execSync` makes it return a string directly, which is cleaner and avoids the intermediate Buffer allocation. Minor, but consistent with Node.js idiom.

---

### What works well

- Schema design is clean: `UNIQUE(space_id, task_id)` and `UNIQUE(space_id, slug)` constraints prevent duplicates at the DB level.
- Idempotent `createTaskWorktree` is correctly implemented — early return on existing record before touching the filesystem.
- Orphan cleanup logic correctly iterates and removes only missing paths, preserving live worktrees.
- The 18-test suite covers the full lifecycle with real git repos — this is much more reliable than mock-based tests and correctly validates the git operations end-to-end.
- Error propagation on `git worktree add` failure includes partial-directory cleanup, which is good defensive behavior.
- DB record is only written after the filesystem operation succeeds, keeping them in sync.

lsm (commented):

Overall this is a well-structured implementation. The shell injection fix (execFileSync with array args), 30s timeouts, idempotent create, and comprehensive tests all look good. Two suggestions worth considering:

### P2: Silent catch in `cleanupOrphaned` branch deletion

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, line 221

```ts
} catch {
    // Branch may already be gone; non-fatal
}
```

The rest of the file consistently logs warnings when git operations fail (lines 99-101, 160-163, 173-176, 240-242). This bare `catch {}` is the one exception. For consistency and debuggability, this should log a warning like the other catch blocks do. This was specifically called out in the previous review iteration ("added warning logs") but this particular catch was missed.

### P2: TOCTOU gap between stale branch check and worktree creation

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`, lines 84-112

There is a time-of-check-to-time-of-use gap between checking/deleting a stale branch (lines 84-102) and the `git worktree add` call (lines 104-112). If another process creates the same branch name between these two operations, `worktree add -b` will fail. This is unlikely in practice (the slug is unique per space) but the window exists. A comment acknowledging this would suffice, or the stale branch check could be moved into the catch block of `worktree add` as a retry strategy.

### P3: Async methods with no async work

All public methods (`createTaskWorktree`, `removeTaskWorktree`, etc.) are `async` but perform only synchronous operations (`execFileSync`, synchronous SQLite queries). This is presumably for future-proofing (e.g., switching to async git operations later), which is fine, but worth a brief comment explaining the rationale.

**Recommendation: APPROVE** -- the P2 items are minor and non-blocking in practice. The implementation is correct, well-tested, and addresses all prior review feedback.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P1 gap and one P2 issue worth addressing before merge.

---

### P1 — Stale worktree *directory* is not cleaned up before recreation

In `createTaskWorktree` (line 79), if a previous crashed run left behind the worktree *directory* at `{workspacePath}/.worktrees/{slug}/` (without a DB record — e.g. the crash happened between the `git worktree add` call and the DB `create` at line 128), the code does not remove it before calling `git worktree add`.  Git will refuse to create a worktree at a path that already exists, and the call will throw.

The room-level `WorktreeManager` (packages/daemon/src/lib/worktree-manager.ts, line 276) explicitly guards against this with an existence check.  `SpaceWorktreeManager` handles the stale *branch* case but silently misses the stale *directory* case — the symmetric failure mode.

The fix is to add a pre-creation check that removes a stale directory before calling `git worktree add`, the same way the error handler cleans it up post-failure (lines 115-120).

---

### P2 — `SpaceWorktreeManager` is not wired into the rest of the daemon

The class is exported from `packages/daemon/src/lib/space/index.ts` but is not instantiated anywhere in the running daemon (`DaemonApp`, RPC handlers, runtime service, or `TaskAgentManager`).  Task agents are still passed `space.workspacePath` directly (task-agent-manager.ts lines 249, 284, 931, 982) rather than a per-task worktree path.

This is fine if M4.2 is scoped as "implement the class; wire it up in a follow-on task", but the PR description does not say that.  If downstream callers are supposed to start using it now, the lack of integration means none of the worktree management code actually runs in production.  Clarify scope or add the wiring before merge.

---

### P2 — `.worktrees/` directory has no `.gitignore` entry

`createTaskWorktree` creates `{workspacePath}/.worktrees/` inside the user's own repository root.  The directory is not added to `.gitignore`, so it will show up as an untracked path in `git status` from the user's main checkout.  The room-level `WorktreeManager` avoids this by placing worktrees outside the repo root.  At minimum, consider appending `.worktrees/` to the workspace's `.gitignore` (or `.git/info/exclude`) when the directory is first created.

---

- The `SpaceWorktreeInfo` interface (line 22-26) is an internal type; re-exporting it via `packages/daemon/src/lib/space/index.ts` before any RPC consumer exists exposes implementation surface area prematurely.
- The comment on line 10 (`uses execSync directly`) is stale — the implementation uses `execFileSync`, not `execSync`.
- `cleanupOrphaned` silently swallows branch-delete errors for orphaned records (lines 221-222) with no log message, unlike every other error path which logs a warning.  Low risk but inconsistent.

lsm (commented):

**Recommendation: APPROVE (with minor notes)**

The implementation is solid and well-structured. The core logic is correct, tests cover the important scenarios including the newly-added stale-directory recovery, and the migration follows established patterns.

---

### What looks good

- `execFileSync` with array args throughout — no shell injection risk; the shell-injection test documents this explicitly
- Idempotency is correctly handled: the DB record is the source of truth for "already exists", and the early-return path is exercised by a dedicated test
- Stale-branch and stale-directory cleanup are both present in `createTaskWorktree`, with appropriate non-fatal error handling
- Filesystem cleanup on `git worktree add` failure is correct
- `removeTaskWorktree` deletes the DB record unconditionally after attempting FS cleanup, which is the right failure-safe order
- Migration 64 is properly numbered, idempotent (`CREATE TABLE IF NOT EXISTS`), and places FK references to both `spaces` and `space_tasks` with `ON DELETE CASCADE`
- TypeScript is clean (no typecheck errors)

---

### Issues

**P2 — `cleanupOrphaned`: branch deletion runs before `git worktree prune` but can delete branches that git's own worktree machinery still holds**

In `cleanupOrphaned` (line 226), the code runs `git branch -D space/<slug>` immediately upon detecting a missing directory. If the directory was removed manually but git's worktree metadata file still points to it, git may have the branch locked. In practice the `git branch -D` will fail with "error: Cannot delete branch … checked out at …" and the `catch {}` silently swallows that. The pattern of running `git worktree prune` *first*, then `git branch -D`, would be more robust — `prune` removes the stale metadata so subsequent `branch -D` succeeds cleanly. This is a low-impact issue (currently non-fatal) but is worth fixing for correctness.

**P2 — `cleanupOrphaned` test gap: branch cleanup is not asserted**

The four `cleanupOrphaned` tests (lines 352–402) verify that DB records are removed for missing directories, but none assert that the corresponding `space/<slug>` branch was also deleted from git. Since the branch-deletion path is exercised in `cleanupOrphaned`, it deserves the same assertion that `removeTaskWorktree` already has (see the `branchList` assertion at line 258). Without this, a regression in the branch-cleanup path would go undetected.

**P2 — Idempotency of `createTaskWorktree` does not verify directory still exists**

The early-return for an existing record (lines 64–67) returns the stored path without checking whether the directory actually exists on disk. If the worktree directory was removed externally (e.g. by the OS or a manual `rm`), callers get a stale path pointing at a non-existent location. Given that `cleanupOrphaned` already handles this scenario, this may be intentional — but if so it should be documented. If not intentional, a defensive `existsSync(existing.path)` check before the early-return, followed by re-creation, would be the correct fix.

**P3 — `SpaceWorktreeManager` is exported from `index.ts` but has no callers in `src/` yet**

`grep` confirms the manager is only imported in the test file and exported from `index.ts`. This is expected for a new building block, but the PR description should mention this explicitly (it does say M4.2, which implies integration comes later). No action needed, just noting it.

**P3 — `comment-style` inconsistency in `migrations.ts` runner block (lines 229, 233, 237, etc.)**

Several migration-registration comments have `/ Migration XX:` instead of `// Migration XX:` (a stray space after the slash). These appear to be pre-existing in the file and unrelated to migration 64 which is correct, so this is truly a nit.

---

### Summary

Core implementation is correct and safe. Two actionable P2 issues: the `git worktree prune`-before-`branch -D` ordering in `cleanupOrphaned`, and the missing branch-deletion assertion in the corresponding tests. The idempotency gap under external directory removal is worth documenting if it's a deliberate design choice, or fixing with a fallback re-creation path.

PR #1028, opened 2026-03-27
feat(space): channel and gate MCP tools for node agents (M1.3)

lsm:
## Summary

- Adds four new MCP tools for node agents: `list_channels`, `list_gates`, `read_gate`, `write_gate`
- Gates guard channels — node agents use these tools to inspect and update gate state for vote-counting and coordination
- Role-based authorization enforced on `write_gate` via `allowedWriterRoles`
- `write_gate` re-evaluates gate condition after each write

## Changes since initial review

- **P0 fix**: `workflowNodeId` now resolved from `stepTaskId` (sub-session task) instead of `taskId` (parent orchestration task) — was always empty before
- **P2**: `write_gate` description notes runtime polling model (tool updates data, runtime polls at channel-routing time)
- **P2**: `list_channels` description notes `gateId` type limitation (`WorkflowChannel` uses inline `gate?: WorkflowCondition`; separated `Gate` entities require a future type migration)
- **P3**: Removed misleading literal `${workflowNodeId}` from JSDoc
- **P1**: Added comprehensive unit tests for all four tools (590+ lines added)

## Test plan

- [x] Unit tests pass (`bun test packages/daemon/tests/unit/space/node-agent-tools.test.ts`) — 63 tests

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -57,6 +57,7 @@ import { SpaceWorkflowManager } from '../space/managers/space-workflow-manager';
 import type { SpaceAgentLookup } from '../space/managers/space-workflow-manager';
 import { SpaceTaskRepository } from '../../storage/repositories/space-task-repository';
 import { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
+import { GateDataRepository } from '../../storage/repositories/gate-data-repository';
 import { setupSpaceAgentHandlers } from './space-agent-handlers';
 import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
 import { SpaceWorkflowRepository } from '../../storage/repositories/space-workflow-repository';
@@ -315,6 +316,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	// Space handlers (spaceManager injected from deps — single instance shared with DaemonAppContext)
 	const spaceTaskRepo = new SpaceTaskRepository(deps.db.getDatabase());
 	const spaceWorkflowRunRepo = new SpaceWorkflowRunRepository(deps.db.getDatabase());
+	const gateDataRepo = new GateDataRepository(deps.db.getDatabase());
 
 	// Space workflow manager — created early so space.create can call seedBuiltInWorkflows
 	const spaceWorkflowRepo = new SpaceWorkflowRepository(deps.db.getDatabase());
@@ -386,6 +388,7 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		spaceRuntimeService,
 		taskRepo: spaceTaskRepo,
 		workflowRunRepo: spaceWorkflowRunRepo,
+		gateDataRepo,
 		daemonHub: deps.daemonHub,
 		messageHub: deps.messageHub,
 		getApiKey: () => deps.authManager.getCurrentApiKey(),
--- a/packages/daemon/src/lib/space/agents/task-agent.ts
+++ b/packages/daemon/src/lib/space/agents/task-agent.ts
@@ -56,6 +56,7 @@ import type {
 	WorkflowCondition,
 	WorkflowRule,
 	SessionFeatures,
+	Gate,
 } from '@neokai/shared';
 import { resolveNodeAgents } from '@neokai/shared';
 import type { AgentSessionInit } from '../../agent/agent-session';
@@ -133,6 +134,19 @@ function formatChannel(ch: WorkflowChannel): string {
 	const gateLabel = ch.gate ? formatGateCondition(ch.gate) : '';
 	const label = ch.label ? ` (${ch.label})` : '';
 	return `- \`${ch.from}\` ${dir} \`${to}\`${label}${gateLabel}`;
+}
+
+function formatGate(gate: Gate): string {
+	const desc = gate.description ? ` — ${gate.description}` : '';
+	const writers =
+		gate.allowedWriterRoles.length > 0 ? gate.allowedWriterRoles.join(', ') : '(none)';
+	const condType = gate.condition.type;
+	return (
+		`- **Gate \`${gate.id}\`**${desc}\n` +
+		`  - Condition type: \`${condType}\`\n` +
+		`  - Allowed writer roles: ${writers}\n` +
+		`  - Reset on cycle: ${gate.resetOnCycle}`
+	);
 }
 
 // ---------------------------------------------------------------------------
@@ -446,6 +460,22 @@ export function buildTaskAgentInitialMessage(context: TaskAgentContext): string 
 				`No channels are declared for this workflow. ` +
 					`Agents are fully isolated — \`send_message\` is unavailable unless channels are added.`
 			);
+		}
+
+		// ---- Gates (M1.1 separated Gate entities) ----------------------------
+		const gates = context.workflow.gates;
+		if (gates && gates.length > 0) {
+			parts.push(`\n## Workflow Gates\n`);
+			parts.push(
+				`Gates guard channels — a message on a gated channel is held until the gate's condition passes. ` +
+					`Node agents use \`list_gates\`, \`read_gate\`, and \`write_gate\` tools to inspect and update gate state.\n` +
+					`\n` +
+					`**Vote counting:** for \`count\` condition gates, node agents use their \`nodeId\` ` +
+					`(workflow node ID) as the map key — each node votes exactly once.\n`
+			);
+			for (const gate of gates) {
+				parts.push(formatGate(gate));
+			}
 		}
 	}
 
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -1311,7 +1311,7 @@ export class TaskAgentManager {
 		stepTaskId: string,
 		taskManager: SpaceTaskManager
 	) {
-		const workflowNodeId = this.config.taskRepo.getTask(taskId)?.workflowNodeId ?? '';
+		const workflowNodeId = this.config.taskRepo.getTask(stepTaskId)?.workflowNodeId ?? '';
 		// Build the ChannelResolver from the workflow run's stored config at spawn time.
 		// Channels are written once at step-start by SpaceRuntime.storeResolvedChannels(),
 		// so reading them here gives the correct topology for this sub-session's lifetime.
--- a/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
+++ b/packages/daemon/src/lib/space/runtime/task-agent-manager.ts
@@ -60,6 +60,7 @@ import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceRuntimeService } from './space-runtime-service';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
+import type { GateDataRepository } from '../../../storage/repositories/gate-data-repository';
 import type { SpaceWorktreeManager } from '../managers/space-worktree-manager';
 import type {
 	SubSessionFactory,
@@ -98,6 +99,8 @@ export interface TaskAgentManagerConfig {
 	taskRepo: SpaceTaskRepository;
 	/** Workflow run repository — reading and updating runs */
 	workflowRunRepo: SpaceWorkflowRunRepository;
+	/** Gate data repository — for reading and writing gate runtime data in node agent tools */
+	gateDataRepo: GateDataRepository;
 	/** DaemonHub — event bus for session state change subscriptions */
 	daemonHub: DaemonHub;
 	/** MessageHub — used to write SDK messages */
@@ -1317,6 +1320,11 @@ export class TaskAgentManager {
 			run?.config as Record<string, unknown> | undefined
 		);
 
+		// Load the workflow definition so node agents can access channel/gate metadata.
+		const workflow = run?.workflowId
+			? (this.config.spaceWorkflowManager.getWorkflow(run.workflowId) ?? null)
+			: null;
+
 		return createNodeAgentMcpServer({
 			mySessionId: subSessionId,
 			myRole: role,
@@ -1331,6 +1339,8 @@ export class TaskAgentManager {
 				this.injectSubSessionMessage(targetSessionId, message),
 			taskManager,
 			daemonHub: this.config.daemonHub,
+			workflow,
+			gateDataRepo: this.config.gateDataRepo,
 		});
 	}
 }
--- a/packages/daemon/src/lib/space/tools/node-agent-tool-schemas.ts
+++ b/packages/daemon/src/lib/space/tools/node-agent-tool-schemas.ts
@@ -111,6 +111,78 @@ export const ListReachableAgentsSchema = z.object({});
 export type ListReachableAgentsInput = z.infer<typeof ListReachableAgentsSchema>;
 
 // ---------------------------------------------------------------------------
+// list_channels
+// ---------------------------------------------------------------------------
+
+/**
+ * Schema for `list_channels` input.
+ * Lists all channels declared in the current workflow.
+ * No arguments — channels are derived from the workflow run context.
+ */
+export const ListChannelsSchema = z.object({});
+
+export type ListChannelsInput = z.infer<typeof ListChannelsSchema>;
+
+// ---------------------------------------------------------------------------
+// list_gates
+// ---------------------------------------------------------------------------
+
+/**
+ * Schema for `list_gates` input.
+ * Lists all gates declared in the current workflow with their current data.
+ * No arguments — gates are derived from the workflow run context.
+ */
+export const ListGatesSchema = z.object({});
+
+export type ListGatesInput = z.infer<typeof ListGatesSchema>;
+
+// ---------------------------------------------------------------------------
+// read_gate
+// ---------------------------------------------------------------------------
+
+/**
+ * Schema for `read_gate` input.
+ * Reads the current runtime data for a specific gate from the gate_data table.
+ */
+export const ReadGateSchema = z.object({
+	/** The ID of the gate to read data for. */
+	gateId: z.string().min(1).describe('The gate ID to read current data for'),
+});
+
+export type ReadGateInput = z.infer<typeof ReadGateSchema>;
+
+// ---------------------------------------------------------------------------
+// write_gate
+// ---------------------------------------------------------------------------
+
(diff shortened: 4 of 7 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P0 bug makes `nodeId` useless for all node agents, and the four new tools have zero unit test coverage.

---

### P0 — Wrong task ID used when resolving `workflowNodeId` for node agents

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, line 1200

```ts
const workflowNodeId = this.config.taskRepo.getTask(taskId)?.workflowNodeId ?? '';
```

`taskId` here is the **parent** Task Agent orchestration task, not the sub-session's step task. The `completion-detector.ts` file explicitly documents this:

> "The orchestration task (Task Agent's own task) has `workflowNodeId = null`"

So `workflowNodeId` will always be `''` for every node agent session. The `list_gates` and `write_gate` tools expose `nodeId` to agents and tell them "use this as the map key for vote-counting gates." When every node agent reports the same empty `nodeId`, all votes overwrite each other in the gate data store — a count-condition gate that requires 3 votes from 3 separate nodes will only ever see 1 unique key, so it can never open.

The fix is to read `workflowNodeId` from `stepTaskId` (the step task) rather than `taskId` (the parent task):

```ts
// before (wrong):
const workflowNodeId = this.config.taskRepo.getTask(taskId)?.workflowNodeId ?? '';

// after (correct):
const workflowNodeId = this.config.taskRepo.getTask(stepTaskId)?.workflowNodeId ?? '';
```

---

### P1 — No unit tests for the four new MCP tools

The task description says "Unit tests for all new behaviors," but the test diff only adds two required config fields (`workflow: null`, `gateDataRepo`) to the existing `makeConfig` helper. There are no test cases for `list_channels`, `list_gates`, `read_gate`, or `write_gate`. The critical behaviors that need coverage include:

- `list_channels`: empty workflow (`workflow: null`), channels with and without gates, channel with `gateId`
- `list_gates`: workflow with no gates, gate with default data, gate with runtime data overriding defaults
- `read_gate`: non-existent `gateId` returns error with `availableGateIds`, gate open/closed evaluation
- `write_gate`: unauthorized role returns clear error, authorized role merges data, shallow merge semantics (nested object replacement), gate open after write, gate still closed after partial write, `nodeId` in response matches the step-task's node

The `nodeId` test is especially important given the P0 bug above — a test exercising `write_gate` with a real `stepTaskId` would have caught the empty `nodeId` issue before merge.

---

### P2 — `list_channels` exposes `WorkflowChannel.gate` (legacy inline gate) but not `gateId` (new separated gate)

`list_channels` at line 511-518 maps `WorkflowChannel` entries and reports `hasGate: ch.gate !== undefined`. The new `Channel` type (M1.1) uses `gateId` to reference a separated `Gate` entity, but `WorkflowChannel` uses an inline `gate?: WorkflowCondition`. The response does not expose `gateId` or the separator-style gate reference.

Agents calling `list_channels` cannot correlate a channel to a specific gate ID in `list_gates`/`write_gate`. This breaks the use case of: "I see a gated channel — which gate do I need to write to?" Since `WorkflowChannel` does not have a `gateId` field (only `Channel` does), this is a type mismatch in the API surface. The tool description mentions "whether a gate guards the channel" but agents cannot act on it without the gate ID.

---

### P2 — `write_gate` triggers re-evaluation locally but does not notify the workflow runtime

After a successful write, `write_gate` returns whether the gate is now open, but there is no call into the workflow runtime or channel-gate evaluator to unblock any pending messages on gated channels. A node agent writing to a gate and seeing `gateOpen: true` has no mechanism to trigger delivery of held messages — the runtime evaluator is only invoked at channel-routing time, not pushed from the MCP tool layer.

This is an architectural gap: the tool correctly persists and evaluates the gate, but the side-effect of "unblock the channel" is missing. Whether this is in scope for M1.3 should be explicitly documented in the tool description if the runtime polling handles it separately.

---

### P3 — Template literal in JSDoc references runtime value

`node-agent-tools.ts`, line 612:

```ts
* map key so each node counts only once. Your nodeId is "${workflowNodeId}".
```

This JSDoc comment embeds the runtime `workflowNodeId` value directly. JSDoc comments are static text; they do not interpolate runtime values. The string `"${workflowNodeId}"` is the literal text agents will read in the tool description, not the actual node ID. The tool response already conveys the live `nodeId` value in the JSON payload, so this comment is misleading. It should either be removed or rewritten as a static description.

lsm (commented):

**Recommendation: APPROVE**

All previously identified issues from the first review have been addressed satisfactorily.

### P0 Fix Verified: `workflowNodeId` Resolution

The critical bug is fixed at `task-agent-manager.ts:1200`:
```ts
const workflowNodeId = this.config.taskRepo.getTask(stepTaskId)?.workflowNodeId ?? '';
```
This correctly reads from `stepTaskId` (the per-node sub-task that has `workflowNodeId` populated) instead of `taskId` (the parent orchestration task, which would not have a `workflowNodeId`).

### Test Coverage

63 tests pass across all tool handlers:
- **list_channels**: 4 tests covering null workflow, empty channels, gate presence detection
- **list_gates**: 4 tests covering null workflow, empty gates, default data fallback, runtime data override
- **read_gate**: 3 tests covering non-existent gate, closed gate, open gate evaluation
- **write_gate**: 6 tests covering non-existent gate, role authorization denial, merge semantics, shallow merge of nested objects, wildcard authorization, count-condition voting flow
- Plus existing coverage for list_peers, send_message, report_done, list_reachable_agents, system prompt, and MCP server factory

The vote-counting test (lines 1395-1450) is particularly good — it tests a two-phase vote scenario verifying the gate stays closed after one vote and opens after two.

### Implementation Quality

1. **Authorization model** is correct: `write_gate` checks `allowedWriterRoles` with wildcard `'*'` support before allowing writes. Unauthorized writes return actionable error messages with the allowed roles.

2. **Data flow** is sound: `GateDataRepository.merge()` does shallow merge (read-modify-write), which is safe because SQLite operations in Bun are synchronous — no interleaving within the single-threaded event loop.

3. **Gate re-evaluation after write** works correctly: `write_gate` calls `evaluateGate()` with the merged data and reports the result, letting the agent know immediately whether the gate opened.

4. **Null safety**: All four new tools handle `workflow: null` gracefully via `workflow?.gates ?? []` and `workflow?.channels ?? []`.

5. **Integration wiring** is complete: `gateDataRepo` is created in `rpc-handlers/index.ts:317`, passed through `TaskAgentManager` config, and injected into each `createNodeAgentMcpServer` call in `buildNodeAgentMcpServerForSession`.

6. **Documentation notes** addressed: `list_channels` documents the `gateId` type limitation; `write_gate` description notes the runtime polling model.

### Minor Observations (P3)

1. The `write_gate` shallow merge semantics for the vote-counting use case means an agent doing a second vote must read-then-write the full accumulated map (tested at line 1443). This is documented behavior and the test correctly demonstrates the pattern. No issue — just noting the test comment at lines 1435-1437 helpfully explains this.

No blocking issues found.

lsm (commented):

**Recommendation: APPROVE** (with two minor follow-up items noted below)

---

### Summary

The re-review confirms all five claimed fixes from the previous review have been addressed correctly. The implementation is functionally sound, well-structured, and the test suite is comprehensive (63 tests). Two residual issues are worth noting for follow-up but do not block merge.

---

### P0 Fix Verified

**`workflowNodeId` now resolved from `stepTaskId`** — confirmed in `packages/daemon/src/lib/space/runtime/task-agent-manager.ts` line 1200:

```ts
const workflowNodeId = this.config.taskRepo.getTask(stepTaskId)?.workflowNodeId ?? '';
```

The fix is correct. The previous bug (`getTask(taskId)` reading the parent orchestration task) has been replaced with `getTask(stepTaskId)` which correctly resolves the step-level task's node ID.

---

### P1 Fix Verified

**590+ lines of unit tests** covering all four new tools across 63 test cases (`packages/daemon/tests/unit/space/node-agent-tools.test.ts`). Tests cover:
- `list_channels`: null workflow, empty channels, inline gate presence/absence
- `list_gates`: null workflow, default vs. runtime data precedence
- `read_gate`: missing gate ID (returns available IDs), closed/open gate evaluation
- `write_gate`: authorization check, wildcard role, shallow merge semantics, count-vote accumulation

---

### P2 and P3 Fixes Verified

- `write_gate` description correctly notes the polling model: "Note: the workflow runtime polls gate state at channel-routing time; this tool does not directly push channel unblock events."
- `list_channels` description correctly notes the `gateId` limitation.
- The misleading `${workflowNodeId}` literal in the JSDoc is gone.

---

### Residual Issues (Non-Blocking)

**P2 — `task-agent-manager-mcp.test.ts` tests for `buildNodeAgentMcpServerForSession` use the old 5-param signature**

The method signature changed from 5 to 7 parameters in this PR, but the tests at `packages/daemon/tests/unit/space/task-agent-manager-mcp.test.ts` (lines ~556-571, 602-617) still cast to the 5-param shape and call it with 5 args. At runtime this means `stepTaskId` is `undefined`, so `getTask(undefined)` returns null and `workflowNodeId` silently becomes `''`. The core P0 fix is therefore not regression-tested: a test that seeds a step task with a known `workflowNodeId` and asserts `capturedConfig.workflowNodeId === 'node-x'` would close this gap.

**P3 — `list_peers` may expose `sessionId: null` for pending tasks**

In `packages/daemon/src/lib/space/tools/node-agent-tools.ts` line 193, `t.taskAgentSessionId!` is used with a non-null assertion, but the code comment on line 180 explicitly says the field "may be null" for completed tasks. Tasks that are pending (not yet spawned) would also have `taskAgentSessionId === null`, and the filter `t.taskAgentSessionId !== mySessionId` evaluates to `true` for null, so `sessionId: null` appears in the returned `peers` array. This is a pre-existing bug (present since `a9ff8260f`) not introduced by this PR, but the new test at line 269 does not cover this case.

lsm (commented):

**Recommendation: APPROVE** — the two stated fixes are both correctly implemented and regression-tested. A few minor observations follow.

---

### P0 fix verified

The `workflowNodeId` fix at `task-agent-manager.ts:1200` is correct:

```ts
const workflowNodeId = this.config.taskRepo.getTask(stepTaskId)?.workflowNodeId ?? '';
```

Previously this used `taskId` (the parent orchestration task), which would always return an empty string because the parent task has no `workflowNodeId`. The regression test added in `task-agent-manager-mcp.test.ts:737` directly asserts the captured config's `workflowNodeId` equals the one set on the step task — this is the right level of verification.

---

### `list_peers` filter (P1 fix verified, minor comment imprecision)

The new filter in `node-agent-tools.ts:183–189`:

```ts
(t.taskAgentSessionId !== null || t.status === 'completed')
```

correctly excludes pending tasks with no session while keeping completed tasks whose session may have been cleared. The logic is sound and the `?? null` replacement for the non-null assertion at line 199 is appropriate.

**P3 nit:** The inline comment at line 182 says «Pending tasks with no session are excluded» but the predicate also silently drops `needs_attention`, `cancelled`, `rate_limited`, and `usage_limited` tasks that were set to those terminal states before a session was ever assigned — not just `pending` tasks. While the behavior is likely correct (those tasks were never reachable either), the comment mis-describes the actual predicate. A more accurate comment would be «Tasks with no session are excluded unless they are completed (post-session completion).»

**P2 test gap:** No unit test covers the case where a task with `needs_attention` or `cancelled` status and a `null` sessionId is present in the node — specifically verifying it is excluded from peers. Given the new non-obvious two-part condition, a single test like «failed task with no session is excluded from peers» would prevent regression if someone later widens the filter to only exclude `pending`.

---

### Test signature updates

All four call sites in `task-agent-manager-mcp.test.ts` are correctly updated to the new 7-param signature (`taskId, subSessionId, role, spaceId, workflowRunId, stepTaskId, taskManager`). The `gateDataRepo` injection at lines 176 and 239 is present and wired correctly.

---

### Minor observations (P3)

- `packages/daemon/src/lib/space/tools/node-agent-tools.ts:349` still uses a bare non-null assertion (`t.taskAgentSessionId!`) in the legacy `send_message` path (line 349). This was not touched in this PR, which is reasonable scope-wise, but it is inconsistent with the `?? null` fix applied in the same file for `list_peers`.

- The count-condition voting test in `node-agent-tools.test.ts` (around line 990) notes in a comment that node-b must write the accumulated map including node-a's entry because of shallow-merge semantics. This is a real footgun for real LLM agents — they would need to `read_gate` first and then write the full merged map. This is documented correctly in the test comment but the tool response does not guide the agent to do this. This is an existing design limitation, not introduced by this PR.

---

**Summary:** Both stated fixes (correct `stepTaskId` lookup for `workflowNodeId`, and filtering pending sessionless tasks from `list_peers`) are correctly implemented and regression-tested. The only actionable follow-up is adding a test for the `needs_attention`/`cancelled` null-session edge case and improving the inline comment.

lsm (commented):

### Recommendation: APPROVE

This PR correctly addresses all four issues from the previous review. The implementation is clean and well-tested.

### What was reviewed

1. **P0 fix: `workflowNodeId` resolution** (`task-agent-manager.ts:1200`) — Now uses `stepTaskId` instead of `taskId`. The regression test in `task-agent-manager-mcp.test.ts` (line 687-698, 728-737) seeds a step task with a known `workflowNodeId` and asserts it resolves correctly. This is solid.

2. **P1: `list_peers` pending task filtering** (`node-agent-tools.ts:183-189`) — The filter `(t.taskAgentSessionId !== null || t.status === 'completed')` correctly excludes pending tasks with no session while still including completed tasks whose sessions have ended. The non-null assertion `t.taskAgentSessionId!` was replaced with `?? null`.

3. **Test signature updates** (`task-agent-manager-mcp.test.ts`) — All 4 call sites updated from 5-param to 7-param signature (`stepTaskId`, `taskManager`), and `gateDataRepo` added to the manager config.

4. **New MCP tool tests** (`node-agent-tools.test.ts`) — 590+ lines of tests covering `list_channels`, `list_gates`, `read_gate`, `write_gate` with good coverage of error paths (non-existent gate, unauthorized role) and happy paths (shallow merge, wildcard roles, count-condition gate opening).

### P2: Missing test for `list_peers` pending-task filter

The `list_peers` fix (excluding pending tasks with null sessions) has no dedicated unit test. The existing `list_peers` tests only seed tasks that already have session IDs set. A test that seeds a pending task (`status: 'pending'`, `taskAgentSessionId: null`) and verifies it is excluded from the peers list would protect against regression. Similarly, a completed task with `taskAgentSessionId: null` should still appear.

### P3: Vote-counting shallow merge semantics

The `write_gate` test for count-condition gates (line 1164-1175) documents that node-b must write the *accumulated* vote map because shallow merge replaces the `votes` key wholesale. This is a documented design choice (comment on line 1164-1166) and the test comment makes it clear. However, this means a real agent must always read-then-write for vote counting — a race condition between concurrent reads and writes is possible. This is a known limitation of shallow merge and not a bug in this PR, just worth noting for future work.

### P3: `gateId` field in `list_channels`

The `list_channels` response always returns `gateId: null` (not present in diff but implied by the `hasGate` boolean approach). The tool description documents this limitation clearly.

Overall this is clean work that addresses all the review feedback items correctly.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix is correct and well-tested. The core bug: `rowToSpaceTask` in `space-task-repository.ts` (line 404) maps NULL DB columns to `undefined` via `?? undefined`, but the previous commit's filter used strict `!== null`, which only catches `null` — letting `undefined` values pass through. Switching to loose `!= null` correctly handles both.

### Fix Analysis

**`packages/daemon/src/lib/space/tools/node-agent-tools.ts` (line 188)**

The one-character change from `!==` to `!=` is the correct fix. The filter predicate now reads:

```
t.taskAgentSessionId != null || t.status === 'completed'
```

This excludes any task where `taskAgentSessionId` is nullish (either `null` or `undefined`) unless it is completed — which is exactly the intended semantics. The updated comment on line 182 accurately describes the behavior.

### Test Coverage

The three new tests are well-structured:

1. **Pending task excluded** — seeds a task with status `'pending'` and no `task_agent_session_id` in the DB. Verifies it does not appear in peers. This directly exercises the bug path.
2. **Failed (`needs_attention`) task excluded** — same pattern with a non-pending, non-completed status. Good coverage of the "all non-completed statuses" aspect.
3. **Completed task with no session included** — verifies the exception path: completed tasks should still be visible even without a session. Asserts `sessionId` is `null` (correct, since line 199 coerces `undefined` to `null` via `??`), status is `'completed'`, and `completionSummary` is populated.

Each test uses an isolated `workflowNodeId` to avoid interference with other seeded tasks, which is a good pattern.

### No Issues Found

- The `SpaceTask` type (`packages/shared/src/types/space.ts:218`) declares `taskAgentSessionId?: string | null`, so `undefined` is a valid runtime value — the loose equality is the right approach.
- No other occurrences of `taskAgentSessionId !== null` remain in the source (verified via grep).
- The `sessionId` assertion in the completed-task test uses `toBeNull()`, which is correct because the tool handler maps `undefined` to `null` via `?? null` on line 199.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

This is a focused, correct fix. The root cause is genuine: `rowToSpaceTask` maps `task_agent_session_id = NULL` to `undefined` (via the `?? undefined` pattern at `space-task-repository.ts:404`), not `null`. The original strict `!== null` guard let `undefined` slip through as truthy, exposing sessionless non-completed tasks in `list_peers`. Switching to loose `!= null` correctly catches both.

---

### P0 — None

### P1 — None

### P2 — None

### P3 (nits, optional)

**Test name imprecision** (`packages/daemon/tests/unit/space/node-agent-tools.test.ts`, line 356)

The test is named `'excludes failed task with no session from peers list'` but seeds a task with status `'needs_attention'`, not a generic "failed" concept. The task description called it "needs_attention task excluded". The test covers the right behavior, but the name could cause confusion when someone scans test output. Consider renaming to `'excludes needs_attention task with no session from peers list'` to match both the seeded status and the task description.

**Missing coverage for `in_progress` with no session** (`node-agent-tools.ts`, line 188)

The fix correctly gates all non-completed sessionless tasks. However, an `in_progress` task with no `taskAgentSessionId` (a task mid-transition from spawn to session-attached) is a realistic edge case that isn't directly covered by the new tests. The existing tests cover `pending` and `needs_attention`; adding an `in_progress` case would make coverage airtight. This is optional — the filter logic is clearly correct for this case too.

---

### Correctness Verification

- `rowToSpaceTask` at `packages/daemon/src/storage/repositories/space-task-repository.ts:404`: maps `null` DB value to `undefined` via `?? undefined`. Confirmed.
- Filter predicate `t.taskAgentSessionId != null` evaluates `undefined != null` as `false` (loose equality: `undefined == null` is `true` in JS). Correct — sessionless tasks are excluded.
- `|| t.status === 'completed'` correctly re-admits completed tasks that had their session cleared.
- Three new tests are well-structured, use real SQLite via `runMigrations`, and assert the expected inclusion/exclusion behavior.
- Comment updated accurately from "Pending tasks with no session" to "Tasks with no session" — matches the broadened predicate.

PR #1029, opened 2026-03-27
feat(space): wire SpaceWorktreeManager into TaskAgentManager (M4.3)

lsm:
## Summary

- **Migration 65**: adds `completed_at` column to `space_worktrees` for TTL-based reaper
- **SpaceWorktreeRepository**: `markCompleted()`, `listCompletedBefore()` for lifecycle tracking
- **SpaceWorktreeManager**: `markTaskWorktreeCompleted()`, `reapExpiredWorktrees()` (7-day default TTL)
- **TaskAgentManager**: creates worktree at `spawnTaskAgent()` time; all sub-sessions (node agents) share the same path via factory override; `cleanup(cancelled)` removes immediately, `cleanup(completed)` timestamps for TTL; worktree path stored in workflow run config for M6 artifacts RPC; rehydration restores path from run config
- **rpc-handlers**: `SpaceWorktreeManager` instantiated and injected into `TaskAgentManager`
- **app.ts**: orphan cleanup + TTL reaper run on startup; hourly `setInterval` reaper with `unref()`
- **17 new unit tests**: creation at run start, reuse across nodes, TTL logic, cancellation cleanup, fallback on failure, idempotent markCompleted, reaper delegation

## Test plan

- [ ] `make test-daemon` passes (8927+ tests)
- [ ] `bun run typecheck` clean
- [ ] `bun run lint` 0 errors

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -450,6 +450,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 
 	// On startup: clean up orphaned worktrees (directories missing from disk) and run the TTL reaper.
 	// Both are non-blocking — errors are logged but never propagate to block server start.
+	let reaperTimer: ReturnType<typeof setInterval> | null = null;
 	if (process.env.NODE_ENV !== 'test') {
 		const worktreeStartupCleanup = async () => {
 			try {
@@ -473,7 +474,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 
 		// Run TTL reaper periodically (every hour) for long-running daemon processes.
 		const WORKTREE_REAPER_INTERVAL_MS = 60 * 60 * 1000;
-		const reaperTimer = setInterval(() => {
+		reaperTimer = setInterval(() => {
 			spaceWorktreeManager.reapExpiredWorktrees().catch((err) => {
 				logError('[Daemon] Periodic worktree TTL reaper failed:', err);
 			});
@@ -489,6 +490,12 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 			return;
 		}
 		isCleanedUp = true;
+
+		// Stop the hourly worktree TTL reaper before shutting down other resources.
+		if (reaperTimer !== null) {
+			clearInterval(reaperTimer);
+			reaperTimer = null;
+		}
 
 		try {
 			try {
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -20,6 +20,7 @@ import { SpaceAgentManager } from './lib/space/managers/space-agent-manager';
 import { SpaceManager } from './lib/space/managers/space-manager';
 import type { SpaceRuntimeService } from './lib/space/runtime/space-runtime-service';
 import type { TaskAgentManager } from './lib/space/runtime/task-agent-manager';
+import type { SpaceWorktreeManager } from './lib/space/managers/space-worktree-manager';
 import { JobQueueRepository } from './storage/repositories/job-queue-repository';
 import { JobQueueProcessor } from './storage/job-queue-processor';
 import { createCleanupHandler } from './lib/job-handlers/cleanup.handler';
@@ -71,6 +72,8 @@ export interface DaemonAppContext {
 	spaceRuntimeService: SpaceRuntimeService;
 	/** Task Agent Manager — manages Task Agent session lifecycle for space tasks */
 	taskAgentManager: TaskAgentManager;
+	/** Space Worktree Manager — one git worktree per task, shared by all node agents */
+	spaceWorktreeManager: SpaceWorktreeManager;
 	/** Persistent job queue repository */
 	jobQueue: JobQueueRepository;
 	/** Persistent job queue processor */
@@ -288,6 +291,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		cleanup: rpcHandlerCleanup,
 		spaceRuntimeService,
 		taskAgentManager,
+		spaceWorktreeManager,
 	} = setupRPCHandlers({
 		messageHub,
 		sessionManager,
@@ -444,6 +448,40 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	jobProcessor.start();
 	logInfo('[Daemon] Job queue processor started');
 
+	// On startup: clean up orphaned worktrees (directories missing from disk) and run the TTL reaper.
+	// Both are non-blocking — errors are logged but never propagate to block server start.
+	if (process.env.NODE_ENV !== 'test') {
+		const worktreeStartupCleanup = async () => {
+			try {
+				const spaces = await spaceManager.listSpaces(false);
+				for (const space of spaces) {
+					await spaceWorktreeManager.cleanupOrphaned(space.id);
+				}
+				logInfo('[Daemon] Worktree orphan cleanup complete');
+			} catch (err) {
+				logError('[Daemon] Worktree orphan cleanup failed:', err);
+			}
+
+			try {
+				await spaceWorktreeManager.reapExpiredWorktrees();
+				logInfo('[Daemon] Worktree TTL reaper complete');
+			} catch (err) {
+				logError('[Daemon] Worktree TTL reaper failed:', err);
+			}
+		};
+		void worktreeStartupCleanup();
+
+		// Run TTL reaper periodically (every hour) for long-running daemon processes.
+		const WORKTREE_REAPER_INTERVAL_MS = 60 * 60 * 1000;
+		const reaperTimer = setInterval(() => {
+			spaceWorktreeManager.reapExpiredWorktrees().catch((err) => {
+				logError('[Daemon] Periodic worktree TTL reaper failed:', err);
+			});
+		}, WORKTREE_REAPER_INTERVAL_MS);
+		// Allow the process to exit even if this timer is still pending.
+		reaperTimer.unref();
+	}
+
 	// Cleanup function for graceful shutdown
 	let isCleanedUp = false;
 	const cleanup = async () => {
@@ -561,6 +599,7 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 		spaceManager,
 		spaceRuntimeService,
 		taskAgentManager,
+		spaceWorktreeManager,
 		jobQueue,
 		jobProcessor,
 		appMcpManager,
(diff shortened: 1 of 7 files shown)


lsm (commented):

**Recommendation: REQUEST_CHANGES** — two test gaps undermine the confidence in core behaviors the PR explicitly claims to verify.

---

### P1 — Test for sub-session worktree path override does not actually verify the override

**File:** `packages/daemon/tests/unit/space/task-agent-manager-worktree.test.ts`, lines 507–545

The describe block `'worktree reuse across node agents'` contains two tests. The first ("sub-sessions receive the same worktree path as workspacePath") calls `createSubSession` directly, bypassing the `SubSessionFactory` closure entirely. The test's own comment acknowledges this: *"the path override happens in createSubSessionFactory"*, but then only asserts `getSubSession(subSessionId)` is defined — it never asserts `subSession._workspacePath === '/tmp/worktrees/test-task'`. The second test in the block only asserts the in-memory map contains the correct path after spawn, which is a different assertion than factory override. The actual factory closure (lines 716–731 of `task-agent-manager.ts`) is never called via a sub-session creation in any test.

---

### P1 — Rehydration test does not exercise `rehydrate()` or `rehydrateTaskAgent()`

**File:** `packages/daemon/tests/unit/space/task-agent-manager-worktree.test.ts`, lines 798–818

The describe block `'rehydration restores worktree path from run config'` contains one test that only calls `spawnTaskAgent` and then asserts the run config and in-memory map were populated. It never calls `manager.rehydrate()` nor spies on `AgentSession.restore`. The critical code path — reading `workflowRun.config.worktreePath` after a daemon restart and setting `taskWorktreePaths` — at lines 1004–1014 of `task-agent-manager.ts` is completely untested.

---

### P2 — Misleading comment claims disk presence is checked during rehydration, but no check exists

**File:** `packages/daemon/src/lib/space/runtime/task-agent-manager.ts`, lines 1001–1014

The comment reads: *"If the path is still present on disk we restore it so sub-sessions spawned after restart share the same worktree; otherwise fall back gracefully."* But `existsSync` is never imported in this file and no such check is performed. The code unconditionally sets `taskWorktreePaths` to `storedPath` if the string is non-empty. If the worktree directory was deleted between spawning and restart (e.g., manual cleanup, OS reboot with tmpfs), all sub-sessions after restart would receive a non-existent workspace path without any warning. The comment should be corrected to match reality, or the actual check should be implemented.

---

### P3 — `reaperTimer` is not cleared during graceful shutdown

**File:** `packages/daemon/src/app.ts`, line 476–482

The hourly reaper `setInterval` is `unref()`d (correct) but never `clearInterval`'d in the `cleanup()` function. Since `unref()` prevents blocking exit this is not a functional bug, but it is inconsistent with the project's other timer cleanup patterns and the timer can still fire during the shutdown window. A `clearInterval(reaperTimer)` in the shutdown handler would be cleaner.

lsm (commented):

This is a well-structured integration of `SpaceWorktreeManager` into `TaskAgentManager`. The design is clean: one worktree per task, shared across all node agents via the `taskWorktreePaths` in-memory map, with a clear lifecycle of create-at-spawn / mark-completed / TTL-reap.

### What looks good

- Migration 65 is safe and idempotent (try/catch column existence check before ALTER TABLE)
- Repository methods (`markCompleted` with `WHERE completed_at IS NULL`, `listCompletedBefore`) are correct
- `SpaceWorktreeManager.reapExpiredWorktrees()` handles individual record errors gracefully without blocking the rest of the cleanup pass
- `cleanup()` reason-based branching (cancelled = immediate removal, completed = TTL mark) is clean
- The `worktreeManager` config field is optional, preserving backward compatibility
- Startup orphan cleanup + TTL reaper in `app.ts` are properly gated by `NODE_ENV !== 'test'`
- `setInterval` with `unref()` is the correct pattern for long-lived daemon timers
- Test coverage is thorough: 17 tests covering creation, reuse, completion/cancellation, fallback on failure, idempotency, TTL, and reaper delegation

### Issues

**P2: Rehydration does not verify worktree directory exists on disk**
`packages/daemon/src/lib/space/runtime/task-agent-manager.ts` lines 1001-1014

The comment at line 1001-1003 says "If the path is still present on disk we restore it so sub-sessions spawned after restart share the same worktree; otherwise fall back gracefully" but the code unconditionally stores the path without an `existsSync()` check. If the worktree directory was deleted between restarts (e.g., manual cleanup, disk failure), the rehydrated task agent and any spawned sub-sessions would point at a non-existent workspace directory. The startup orphan cleanup in `app.ts` runs concurrently and may not complete before rehydration.

Adding an `existsSync(storedPath)` guard here would match the stated intent and prevent this race.

**P3: `reaperTimer` not cleared during graceful shutdown**
`packages/daemon/src/app.ts` line 476

The `setInterval` for the hourly TTL reaper is created with `unref()` (which prevents blocking process exit) but is never `clearInterval()`-ed in the `cleanup()` function. This is harmless in practice since the process will exit, but it would be slightly cleaner to clear it during graceful shutdown to avoid a potential late-firing reaper callback during teardown.

**P3: Rehydration test does not actually test the rehydration path**
`packages/daemon/tests/unit/space/task-agent-manager-worktree.test.ts` lines 798-819

The test labeled "rehydration restores worktree path from run config" calls `spawnTaskAgent()` and then asserts the path is present. It verifies that the path is stored in the run config (which is the spawn path), but does not exercise the actual `rehydrate()` code path that reads the path back from `workflowRun.config.worktreePath` via `AgentSession.restore()`. A true rehydration test would: (1) spawn, (2) clear in-memory maps, (3) call `rehydrate()`, (4) verify `getTaskWorktreePath()` returns the restored path. The existing rehydration tests in `task-agent-manager.test.ts` do not cover this new worktree path restoration either.

### Summary

Solid implementation that correctly handles the main lifecycle scenarios. The only gap worth addressing is the missing `existsSync` guard during rehydration (P2), which could cause sub-sessions to be spawned into a non-existent directory after a restart. The other issues are minor nits.

lsm (commented):

**Recommendation: APPROVE**

The four previously raised issues have all been addressed correctly. Here is a detailed assessment of the re-review.

---

### Previously Fixed Issues — Verified

**P2 (fixed): existsSync check in rehydration**

`rehydrateTaskAgent()` at line 1012 in `task-agent-manager.ts` now does:

```ts
if (storedPath && existsSync(storedPath)) {
  this.taskWorktreePaths.set(taskId, storedPath);
  return storedPath;
}
```

The stale-path branch logs a warning and falls back to `space.workspacePath`. Implementation is correct and complete.

**P1 (fixed): Sub-session factory path override tested**

The test `'SubSessionFactory.create overrides workspacePath with worktree path'` at line 513 of the test file accesses the private `createSubSessionFactory` via a type cast, calls `factory.create()` with a deliberate `workspacePath: '/original-path'`, and asserts that the resulting session's `_workspacePath` equals the worktree path. The complementary test at line 542 verifies the no-override (fallback) case. Both are sound.

**P1 (fixed): rehydrateTaskAgent path restoration tested**

Two tests cover the rehydration path:
- `'rehydrate() reads worktreePath from run config and sets taskWorktreePaths'` (line 822) uses `/tmp` as the stored path (guaranteed to exist), calls `rehydrate()`, and asserts `getTaskWorktreePath(task.id) === '/tmp'`.
- `'rehydrate() skips worktree path when stored path no longer exists on disk'` (line 865) uses a path that cannot exist (`/nonexistent/worktree/path/that/surely/does/not/exist`), calls `rehydrate()`, and asserts the result is `undefined`.

Both tests are well-constructed.

**P2 (fixed): Misleading comment removed**

No misleading comment remains about the existsSync check; the comment at line 1003 now accurately describes the check that is actually present.

---

### Overall Assessment

The implementation is complete and correct:

- **Migration 65** adds the `completed_at` column to `space_worktrees` with proper idempotency (try/catch probe pattern consistent with the codebase).
- **SpaceWorktreeRepository** exposes `markCompleted()` (idempotent via `WHERE completed_at IS NULL`) and `listCompletedBefore()`.
- **SpaceWorktreeManager** wraps repository calls and implements `reapExpiredWorktrees()` with per-record error isolation.
- **TaskAgentManager** creates the worktree before session init, persists the path to `workflowRun.config`, propagates it to all sub-sessions via the factory closure, and handles cancellation vs. completion correctly.
- **app.ts** wires orphan cleanup and the hourly reaper timer with `unref()` so the process can exit cleanly. The startup path is gated on `NODE_ENV !== 'test'` to avoid side-effects.
- **17 unit tests** cover all specified behaviors.

No P0 or P1 issues found. No new concerns introduced by the updated code.

lsm (commented):

**Recommendation: APPROVE**

The previous review feedback has been addressed. Re-reviewing the full PR:

### What the PR does

Wires `SpaceWorktreeManager` into `TaskAgentManager` (M4.3) so each space task gets an isolated git worktree. Key behaviors:

1. **Worktree creation** at `spawnTaskAgent()` time, with graceful fallback to `space.workspacePath` on failure
2. **Sub-session path forwarding** -- all node agents share the same worktree via `createSubSessionFactory()`
3. **Cleanup lifecycle** -- `completed` marks for TTL reaper, `cancelled` removes immediately
4. **Rehydration** -- restores worktree path from `workflowRun.config`, with `existsSync` check for stale paths
5. **TTL reaper** -- hourly interval in `app.ts`, properly `unref()`'d and cleared on shutdown
6. **Migration 65** -- adds `completed_at` column with safe try/catch migration pattern

### Previously flagged issues -- all resolved

- **existsSync check on rehydration**: Implemented at `task-agent-manager.ts:1012`. Falls back to `space.workspacePath` when the directory is missing, with a warning log.
- **Sub-session factory path override test**: Added at test line 513 (`SubSessionFactory.create overrides workspacePath with worktree path`). Validates that `_workspacePath` on the mock session matches the worktree path.
- **Rehydration tests**: Two tests added -- happy path (line 822) uses `/tmp` as a guaranteed-to-exist directory, stale path test (line 865) uses a nonexistent path and asserts the map entry is undefined.

### Assessment

The implementation is solid:

- Error handling is thorough at all system boundaries (worktree creation, removal, rehydration)
- The `taskWorktreePaths` in-memory map is always cleaned up in `cleanup()` regardless of `worktreeManager` presence
- The `cleanupAll()` path defaults to `completed`, which is correct for daemon shutdown (preserves worktrees for TTL)
- The periodic reaper has proper `unref()` and cleanup on shutdown
- Tests cover: creation, fallback on error, idempotency, sub-session path forwarding (both with and without worktree), completion vs cancellation lifecycle, TTL reaper at repository and manager level, rehydration happy path and stale path

**Minor nit (P3):** In `space-worktree-manager.ts:232-234`, the log message `TTL reaper: removed ${expired.length} expired worktree(s)` counts all expired records including ones that may have failed removal (caught at line 225). The count could be slightly misleading. Not blocking.

All 18 tests pass. No P0, P1, or P2 issues found.

PR #1030, opened 2026-03-27
feat(space): integrate separated channels + gates with channel router (M1.4)

lsm:
Integrates the M1.1/M1.2 separated Channel+Gate architecture into the ChannelRouter.

## Changes

- **`WorkflowChannel.gateId?`** — new field referencing a `Gate` entity in `workflow.gates`; takes precedence over legacy inline `gate`
- **`ChannelGateBlockedError.gateType`** widened from `WorkflowCondition['type']` to `string` to accommodate new gate condition types and gate IDs
- **`ChannelRouter.deliverMessage()`** — gateId channels use `evaluateGate()` with runtime data from `GateDataRepository`; legacy `gate` field still works as fallback; gateless channels always open
- **`ChannelRouter.canDeliver()`** — same two-path evaluation
- **`ChannelRouter.onGateDataChanged(runId, gateId)`** — re-evaluates channels referencing the changed gate; activates target nodes on blocked→open transition (enables vote-counting gates)
- **`resetOnCycle`** — cyclic channel traversal resets gate data for gates with `resetOnCycle: true`, atomic with iteration counter increment via `db.transaction()`
- `gateDataRepo` + `db` wired into both `ChannelRouter` instantiations in `TaskAgentManager`
- 28 new unit tests covering all scenarios

### DIFF
--- a/packages/daemon/src/lib/space/runtime/channel-gate-evaluator.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-gate-evaluator.ts
@@ -55,8 +55,12 @@ export interface GateResult {
 export class ChannelGateBlockedError extends Error {
 	constructor(
 		message: string,
-		/** The gate type that caused the block (for programmatic handling). */
-		public readonly gateType: WorkflowCondition['type']
+		/**
+		 * The gate type that caused the block (for programmatic handling).
+		 * For legacy WorkflowChannel gates: one of 'always' | 'human' | 'condition' | 'task_result'.
+		 * For new Gate entities: one of 'check' | 'count' | 'all' | 'any', or a gate ID string.
+		 */
+		public readonly gateType: string
 	) {
 		super(message);
 		this.name = 'ChannelGateBlockedError';
--- a/packages/daemon/src/lib/space/runtime/channel-router.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-router.ts
@@ -6,10 +6,30 @@
  * - Cross-node DMs (target resolved by agent role)
  * - Fan-out (target resolved by node name → all agents in that node)
  *
- * Gate enforcement:
- * - canDeliver() evaluates gates and iteration caps without side effects
- * - deliverMessage() performs gate evaluation, cyclic iteration tracking,
- *   and lazy node activation in one atomic operation
+ * Gate enforcement — two systems are supported:
+ *
+ * 1. **New separated Channel+Gate architecture (M1.1/M1.2)**:
+ *    When a WorkflowChannel has a `gateId`, the corresponding Gate entity is
+ *    loaded from `workflow.gates`, its runtime data is fetched from `gate_data`
+ *    via `GateDataRepository`, and `evaluateGate()` is called. This is the
+ *    preferred path for new workflows.
+ *
+ * 2. **Legacy inline gate (WorkflowCondition)**:
+ *    When a WorkflowChannel has an inline `gate` (no `gateId`), the
+ *    `ChannelGateEvaluator` is used as before. Kept for backward compatibility.
+ *
+ * Channels without either `gateId` or `gate` are always open.
+ *
+ * `onGateDataChanged(runId, gateId)`:
+ *    Called by agents (via MCP write_gate) whenever gate data changes.
+ *    Re-evaluates all channels that reference the changed gate. If a previously
+ *    blocked channel is now open and its target node has no active tasks,
+ *    the node is lazily activated.
+ *
+ * `resetOnCycle`:
+ *    When a cyclic channel is traversed, gates with `resetOnCycle: true` in the
+ *    workflow have their data reset to defaults in `gate_data`. This operation
+ *    is performed in the same SQLite transaction as the iteration counter increment.
  *
  * Lazy node activation:
  * - activateNode() is idempotent: if tasks already exist for the node the
@@ -22,7 +42,9 @@
  *   ChannelRouter only creates SpaceTask DB records.
  */
 
+import type { Database as BunDatabase } from 'bun:sqlite';
 import type {
+	Gate,
 	SpaceTask,
 	SpaceTaskType,
 	SpaceWorkflow,
@@ -32,10 +54,12 @@ import type {
 import { resolveNodeAgents } from '@neokai/shared';
 import type { SpaceTaskRepository } from '../../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../../storage/repositories/space-workflow-run-repository';
+import type { GateDataRepository } from '../../../storage/repositories/gate-data-repository';
 import type { SpaceWorkflowManager } from '../managers/space-workflow-manager';
 import type { SpaceAgentManager } from '../managers/space-agent-manager';
 import { ChannelGateEvaluator, ChannelGateBlockedError } from './channel-gate-evaluator';
 import type { ChannelGateContext, GateResult } from './channel-gate-evaluator';
+import { evaluateGate } from './gate-evaluator';
 
 // Re-export for callers that only need to import from channel-router
 export { ChannelGateBlockedError };
@@ -116,6 +140,20 @@ export interface ChannelRouterConfig {
 	workflowManager: SpaceWorkflowManager;
 	/** Agent manager for resolving agent roles → task types */
 	agentManager: SpaceAgentManager;
+	/**
+	 * Gate data repository for reading/writing gate runtime data.
+	 * Required when workflows use the new separated Channel+Gate architecture
+	 * (WorkflowChannel.gateId). Optional for legacy-only workflows.
+	 */
+	gateDataRepo?: GateDataRepository;
+	/**
+	 * Raw SQLite database handle.
+	 * When provided, gate data resets and iteration counter increments on cyclic
+	 * channels are performed in a single atomic transaction (`db.transaction()`).
+	 * When omitted, operations are performed sequentially (still safe for most
+	 * single-process use cases but not ACID-atomic on crash).
+	 */
+	db?: BunDatabase;
 	/**
 	 * Injectable gate evaluator — omit to use the real evaluator (Bun.spawn).
 	 * Inject a mock in tests to avoid real subprocess calls.
@@ -243,6 +281,8 @@ export class ChannelRouter {
 	 * 2. **Iteration cap** — for cyclic channels, blocks when
 	 *    `run.iterationCount >= run.maxIterations`.
 	 * 3. **Gate condition** — evaluates the channel's gate using the provided context.
+	 *    For new-style channels (gateId): uses `evaluateGate()`.
+	 *    For legacy channels (inline gate): uses `ChannelGateEvaluator`.
 	 *
 	 * @param runId     - Workflow run ID
 	 * @param fromRole  - Sending agent role name
@@ -278,6 +318,13 @@ export class ChannelRouter {
 		}
 
 		// ── Gate condition evaluation ────────────────────────────────────────
+		// New separated architecture: gateId takes precedence
+		if (channel.gateId) {
+			const gateResult = this.evaluateGateById(runId, channel.gateId, workflow);
+			return { allowed: gateResult.open, reason: gateResult.reason };
+		}
+
+		// Legacy inline gate
 		if (!channel.gate) return { allowed: true };
 		return this.gateEvaluator.evaluateCondition(channel.gate, context);
 	}
@@ -292,29 +339,32 @@ export class ChannelRouter {
 	 * - `toTarget` is a node name → fan-out to the node; all agent slots are
 	 *   activated (lazy-activated if not already active)
 	 *
-	 * **Gate evaluation (when `context` is provided):**
-	 * - Locates the matching `WorkflowChannel` for the from/to pair
-	 * - Throws `ChannelGateBlockedError` if the gate blocks delivery
-	 * - When no context is given, gates are not evaluated (backward-compatible)
+	 * **Gate evaluation:**
+	 * - Channels with `gateId` (new architecture): `evaluateGate()` is called;
+	 *   gate data is loaded from `gate_data` via `GateDataRepository`.
+	 *   Does NOT require a context object.
+	 * - Channels with inline `gate` (legacy): `ChannelGateEvaluator` is used
+	 *   when a `context` object is provided.
+	 * - Channels with neither: always open.
+	 * - Throws `ChannelGateBlockedError` when a gate blocks delivery.
 	 *
 	 * **Cyclic iteration tracking:**
 	 * - If the matching channel has `isCyclic: true`, `run.iterationCount` is
-	 *   incremented after successful delivery, regardless of whether `context` is
-	 *   provided.
-	 * - If the iteration cap has already been reached, throws `ActivationError`
-	 *   (unrecoverable hard limit, not a retryable gate condition).
+	 *   incremented after successful delivery.
+	 * - Gates with `resetOnCycle: true` have their data reset to defaults in the
+	 *   same atomic operation (single SQLite transaction when `db` is in config).
+	 * - If the iteration cap has already been reached, throws `ActivationError`.
 	 *
 	 * @param runId    - Workflow run ID
 	 * @param fromRole - Role name of the sending agent (WorkflowNodeAgent.name)
 	 * @param toTarget - Role name of the receiving agent, or node name for fan-out
 	 * @param message  - Message content to deliver
-	 * @param context  - Optional gate evaluation context; omit to skip gate checks
+	 * @param context  - Optional gate evaluation context; used only for legacy inline gates
 	 * @returns DeliveredMessage descriptor; `activatedTasks` is set when the
 	 *   target node was lazily activated, undefined when it was already active
 	 * @throws ActivationError when the run, workflow, or target role/node is not
 	 *   found, or the cyclic iteration cap is exceeded
 	 * @throws ChannelGateBlockedError when a gate condition blocks delivery
-	 *   (only when `context` is provided)
 	 */
 	async deliverMessage(
 		runId: string,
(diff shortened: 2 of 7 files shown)


lsm (commented):

The implementation covers all the architectural goals (gateId two-path evaluation, onGateDataChanged, resetOnCycle, backward compat with legacy gates) and the test suite is comprehensive for the happy paths. However, there are meaningful issues to address before merge.

---

### P0 — `onGateDataChanged` is never wired to `write_gate`

The core value of the `onGateDataChanged` method is that it auto-activates blocked target nodes when a gate transitions from closed → open. The method is defined and tested in isolation, but **`write_gate` in `node-agent-tools.ts` (line 656) never calls it**. Every agent call to `write_gate` merges data and evaluates the gate locally, then returns the `gateOpen` flag — but does not trigger node activation.

`packages/daemon/src/lib/space/tools/node-agent-tools.ts`, line 656–671

Even the tool description at line 787 acknowledges this: *"the workflow runtime polls gate state at channel-routing time; this tool does not directly push channel unblock events."* That statement accurately documents the current broken state, but it contradicts the PR description and the `onGateDataChanged` docblock which both state that `write_gate` triggers `onGateDataChanged`. This means vote-counting gates will not auto-activate target nodes in production — they only activate when the next `deliverMessage` is called.

If the intent is that `onGateDataChanged` should be triggered from `write_gate`, that wiring must be added. If polling at channel-routing time is the designed behaviour, then `onGateDataChanged` is dead code and must be removed, and tests mocking the connection should be removed or marked as future work.

---

### P1 — `evaluateGateById` ignores `gateDataRepo` absence silently when `gateId` is set

`evaluateGateById` uses `this.config.gateDataRepo?.get(runId, gateId)` with optional chaining. When `gateDataRepo` is not configured, `record` is `undefined` and the code falls back to `gateDef.data` (the gate's default data). This means that in a workflow using `gateId` channels, if the `ChannelRouter` is instantiated without a `gateDataRepo` (e.g. in a test or legacy code path), the gate always evaluates against its default data — which can silently allow passage if the default data satisfies the condition. The comment in the docblock says the function returns `{ open: false }` when `gateDataRepo` is not configured, but that is not what the code does.

`packages/daemon/src/lib/space/runtime/channel-router.ts`, lines 557–562

---

### P1 — `incrementAndResetCyclicGates` swallows the ActivationError from cyclic cap on the non-atomic path

On the sequential path (no `db` provided), `incrementIterationCount` can return `false` and the code throws `ActivationError`. However, on the **atomic path** (`db` + `gateDataRepo` configured), the transaction wraps the reset + increment and returns the boolean. The result is cast with `as boolean` (line 586), which hides a potential type mismatch if `bun:sqlite` transactions have quirks with boolean return values from JS. Additionally, neither path checks whether `message was delivered but cycle was not counted` is the right outcome — message delivery already happened at this point (step 4 completed, step 5 failed). The error message says "message was delivered" but the iteration count was not incremented, meaning the cyclic iteration count is silently inconsistent. This can lead to workflows exceeding their declared `maxIterations` by one cycle.

`packages/daemon/src/lib/space/runtime/channel-router.ts`, lines 586–604

---

### P1 — No test for `canDeliver` with new-style `gateId` allowed path

The test at line 1778 covers `canDeliver` returning `false` with a `gateId` channel, but there is no matching test for `canDeliver` returning `true` after a `gateId` gate opens. This leaves the positive new-architecture path of `canDeliver` untested.

`packages/daemon/tests/unit/space/channel-router.test.ts`

---

### P2 — `runId` parameter of `evaluateGateById` is passed but never used

`evaluateGateById(runId, gateId, workflow)` receives `runId` as a parameter but only uses `gateId` and `workflow` (and the config-level `gateDataRepo.get(runId, gateId)`). Since the signature implies `runId` is needed, this is fine structurally — but the parameter is technically forwarded into `gateDataRepo.get`. The issue is that the docblock says it returns `{ open: false }` when `gateDataRepo` is not configured, which is incorrect (see P1).

---

### P2 — Concurrent `onGateDataChanged` test does not verify all-3-approvals final state

The concurrency test at line 2264 verifies only `Object.keys(approvals!).length).toBeGreaterThan(0)` — meaning a single surviving approval satisfies the assertion. Because the `merge` calls are not atomic across the three `Promise.all` branches, writes can clobber each other (the `merge` method does a shallow `get` then `set`, not a DB-level merge). The test comment acknowledges this but the test then does not fail if data races occur. This gives false confidence in the concurrent write behaviour.

`packages/daemon/tests/unit/space/channel-router.test.ts`, lines 2264–2320

---

### P3 — Stale tool description for `write_gate`

The `write_gate` tool description at line 787 says *"this tool does not directly push channel unblock events"*. This is accurate today but inconsistent with the claimed design of `onGateDataChanged`. Once the wiring issue (P0) is resolved, this comment should be updated.

`packages/daemon/src/lib/space/tools/node-agent-tools.ts`, line 787

---

### Summary

The architectural skeleton is solid and well-documented. The main blocker is that `onGateDataChanged` is never called from `write_gate`, which means the vote-counting auto-activation feature does not work end-to-end. The silent fallback in `evaluateGateById` when `gateDataRepo` is absent can also silently allow or block delivery based on default gate data rather than failing clearly. These two issues must be addressed before this PR is safe to merge.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured integration of the separated Channel+Gate architecture into ChannelRouter. The implementation correctly addresses all four requirements: gateless channels always open, gated channels using `evaluateGate()`, `onGateDataChanged()` for vote-counting, and `resetOnCycle` with SQLite transactions.

### What works well

- **Two-path gate evaluation is correctly implemented**: `gateId` takes precedence over legacy `gate`, and channels with neither are always open. The return type mapping from `{ open, reason }` to `{ allowed, reason }` in `canDeliver` is handled correctly.
- **`onGateDataChanged()` is well designed**: It evaluates the gate once and then activates all target nodes referencing it. The method correctly handles `channel.to` being either a string or array (line 502). Terminal run states are properly guarded (line 484).
- **`incrementAndResetCyclicGates()` correctly uses `db.transaction()`**: The atomic path wraps both gate resets and iteration increment in a single SQLite transaction. The sequential fallback path for missing `db` is reasonable for single-process use.
- **Test coverage is thorough**: 28 new tests (60 total) covering gateless channels, gated channels (check condition), missing gate definitions, `onGateDataChanged` transitions, vote-counting gates, `resetOnCycle` with selective preservation, concurrent writes, and `gateId` precedence over legacy gates. All 60 tests pass.
- **Type changes are minimal and backward-compatible**: `gateId?: string` added to `WorkflowChannel`, `gateType` widened to `string` in `ChannelGateBlockedError`.
- **`task-agent-manager.ts` integration is straightforward**: `gateDataRepo` and `db` (via `getDatabase()`) are passed to both ChannelRouter instantiation sites (initial spawn and rehydration).

### Minor observations (P3)

1. **`evaluateGateById` when `gateDataRepo` is undefined** (`channel-router.ts:558`): When a channel has a `gateId` but `gateDataRepo` is not configured, the method evaluates against `gateDef.data` (the gate's default data). This is a safe fallback, but it silently evaluates a gated channel that the caller may expect to use runtime data. Since `gateDataRepo` is documented as optional ("Required when workflows use the new separated Channel+Gate architecture"), this is acceptable.

2. **Concurrent write test assertion is weak** (`channel-router.test.ts:2319`): The test asserts `Object.keys(approvals!).length).toBeGreaterThan(0)` rather than checking all 3 approvals are present. The comment at line 2318 acknowledges this ("the writes may have raced on merge") -- this is realistic for the shallow-merge semantics of `GateDataRepository.merge()`, but worth noting that the test does not actually verify the SQLite serialization claim in its title.

3. **ACTIVE_STATUSES is re-created on every call** (`channel-router.ts:627`): The `Set` in `getActiveTasksForNode` is instantiated on each invocation. Moving it to a module-level constant would be a micro-optimization, though it has no functional impact.

Overall this is a solid integration milestone. The code is well-documented, the architecture cleanly separates concerns, and the test suite covers both happy paths and edge cases.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (own PR — posting as COMMENT)

---

All previously flagged P0/P1 issues have been correctly addressed. The `onGateDataChanged` callback is now properly wired through from `write_gate` -> `NodeAgentToolsConfig` -> `buildNodeAgentMcpServerForSession`, the docblock on `evaluateGateById` is accurate, the unnecessary cast is gone, and the missing `canDeliver + gateId` tests are present. Good work on those.

However, one new P2 issue has been introduced by this very PR that wasn't present before.

---

### P2 — `list_channels` response silently omits `gateId` and reports stale `hasGate`

**File:** `packages/daemon/src/lib/space/tools/node-agent-tools.ts`, lines 519–547 (handler) and line 778 (MCP tool description)

This PR adds `gateId?` to `WorkflowChannel` in `packages/shared/src/types/space.ts` (line 777). But the `list_channels` handler was not updated to match:

1. `hasGate` is computed as `ch.gate !== undefined` (line 539). A channel that uses the **new** `gateId` path (no inline `gate`) will incorrectly report `hasGate: false` to agents — they have no way to know the channel is gated.
2. The docstring on the handler (lines 525–527) still says: "Note: `gateId` is always null — `WorkflowChannel` uses an inline `gate?: WorkflowCondition` rather than a `gateId` reference." This is now factually wrong: `WorkflowChannel.gateId` is a real, live field added in this very PR.
3. The MCP tool description string (line 778) repeats the same stale claim: "Note: `gateId` is not yet available...".

An agent that calls `list_channels` on a workflow where channels use the new `gateId` architecture will see `hasGate: false` on those channels and have no indication that a gate guards delivery. The `list_channels` response should expose `gateId` (or at minimum compute `hasGate` as `ch.gate !== undefined || ch.gateId !== undefined`) and the docstring/tool description must be updated.

---

### P3 — Weak assertion in the "concurrent gate writes" test

**File:** `packages/daemon/tests/unit/space/channel-router.test.ts`, lines 2356–2361

The test only asserts `Object.keys(approvals!).length > 0` (at least one approval present) even though the scenario seeds 3 separate writes. Since Bun's JS runtime is single-threaded and `merge` is synchronous (not awaited before `onGateDataChanged`), the assertion doesn't actually verify concurrent safety — it will pass even if two of the three writes are lost. The test intent is reasonable but the assertion should verify all 3 approvals are present, or the comment should be clarified to say it only tests that at least one write is not lost.

---

### Summary

The P0/P1 regressions from the prior round are all resolved. One P2 issue has been newly introduced: the `list_channels` MCP tool now silently misreports gated channels that use `gateId` (new architecture) as ungated, and carries stale documentation claiming `gateId` is not available — contradicting `WorkflowChannel` itself. This needs to be fixed before merging.

lsm (commented):

The fix commit (cfeea4116) correctly addresses all five issues from the previous review round. All tests pass (61 channel-router, 68 node-agent-tools), typecheck is clean.

### Issues addressed from previous review

1. **P0 `onGateDataChanged` never called** -- Fixed. `write_gate` now calls `onGateDataChanged` callback (line 674-676 of `node-agent-tools.ts`), wired through a per-session `ChannelRouter` in `task-agent-manager.ts` (line 1334-1341, 1359).
2. **P1 `evaluateGateById` docblock inaccuracy** -- Fixed. Docblock now correctly describes the fallback behavior when `gateDataRepo` is absent (lines 533-546 of `channel-router.ts`).
3. **P1 Unnecessary `as boolean` cast** -- Fixed. `transaction()` return type is now inferred correctly (line 591).
4. **P1 Missing `canDeliver` + `gateId` allowed path test** -- Fixed. New test `'gated channel (check): canDeliver returns true when condition satisfied'` added.
5. **P3 Stale `write_gate` tool description** -- Fixed. Now mentions automatic activation of blocked target nodes (line 804).

### New issues found

**P2: `list_channels` tool reports incorrect `hasGate` for new-style gated channels**

`/packages/daemon/src/lib/space/tools/node-agent-tools.ts`, line 539:
```typescript
hasGate: ch.gate !== undefined,
```

This only checks the legacy inline `gate` field. A channel gated via the new `gateId` path (the primary path introduced in this PR) will report `hasGate: false` to the calling agent. This is misleading -- the agent would believe the channel is ungated when it is actually gated.

Should be:
```typescript
hasGate: ch.gate !== undefined || ch.gateId !== undefined,
```

Additionally, the `gateId` value itself is not exposed in the response, so agents have no way to correlate channels with gates from `list_gates`. Consider including `gateId: ch.gateId ?? null` in the response.

**P3: Stale docstring and MCP tool description for `list_channels`**

`/packages/daemon/src/lib/space/tools/node-agent-tools.ts`, lines 525-527 and 777-778:

The docstring says "`gateId` is always null" and the MCP description says "`gateId` is not yet available". Both are factually wrong after M1.4 -- `WorkflowChannel.gateId` exists and is actively used by `ChannelRouter.deliverMessage()` and `canDeliver()`.

**P3: Silent error swallowing in `write_gate` callback**

`/packages/daemon/src/lib/space/tools/node-agent-tools.ts`, line 675:
```typescript
void onGateDataChanged(workflowRunId, gateId).catch(() => {});
```

The `.catch(() => {})` swallows all errors silently with no logging. If `onGateDataChanged` fails (e.g., `activateNode` throws an unexpected error that isn't caught by the inner try/catch at channel-router.ts line 517), there's no trace of the failure. Consider logging at `warn` level in the catch handler for debuggability. This is acceptable for now since the inner `onGateDataChanged` in `ChannelRouter` already has its own try/catch around `activateNode`, but it would be safer to log the outer catch.

### Summary

Recommendation: **REQUEST_CHANGES** (1x P2).

The core integration (ChannelRouter gateId support, onGateDataChanged callback, cyclic reset, tests) is solid and well-structured. The one remaining issue is that `list_channels` gives agents incorrect information about whether a channel is gated when using the new `gateId` architecture -- a direct inconsistency with the feature this PR introduces.

lsm (commented):

### Summary

Re-review of M1.4 (Integrate Separated Channels + Gates with Channel Router). All three previously reported issues have been addressed:

1. **`hasGate` fix (was P2)**: `list_channels` now correctly computes `hasGate: ch.gate !== undefined || ch.gateId !== undefined` at `/packages/daemon/src/lib/space/tools/node-agent-tools.ts:542`. The `gateId` field is also now exposed in the tool response (line 544). Tool description updated to document both gating paths.

2. **warn logging (was P3)**: The `write_gate` catch handler at line 680 now logs with `log.warn()` instead of silently swallowing errors.

3. **Vote accumulation test fix**: The flaky concurrent writes test (which relied on `Promise.all` over synchronous SQLite calls and could lose votes due to shallow merge semantics) was replaced with a correct sequential read-then-write accumulation test that properly demonstrates the vote-counting pattern.

### Remaining observations

**P3 — Missing test for `gateId`-only channel hasGate**

The `list_channels` tests in `node-agent-tools.test.ts` (lines 998-1056) cover:
- `hasGate: false` when neither `gate` nor `gateId` is set
- `hasGate: true` when inline `gate` is set

But there is no test for a channel with only `gateId` (no inline `gate`) verifying `hasGate: true`. The code path is exercised extensively in the channel-router integration tests and the logic is a simple `||`, so this is cosmetic — but adding the test would make the tool-level coverage complete.

### Recommendation: APPROVE

All previously reported issues are resolved. The implementation is solid: gate evaluation, `onGateDataChanged` reactive activation, cyclic reset, vote accumulation, and gateId-takes-precedence semantics are well-tested with 61 channel-router tests and 68 node-agent-tools tests all passing.

lsm (commented):

---

### Summary

All previously reported P0/P1/P2 issues have been addressed. The code is correct and well-tested. There is one remaining minor gap noted below.

---

### What was fixed since the last round

**P2 (was P2 — now fixed):** `list_channels` previously computed `hasGate` only from the legacy `ch.gate` field. This is now correctly `ch.gate !== undefined || ch.gateId !== undefined`, and the `gateId` field is included in the response. The tool description was updated to mention both gate paths. (commit `9fd9a3b`)

**P3 (was P3 — now fixed):** The concurrent-writes test was replaced with a correct read-then-write vote-accumulation test that accurately documents shallow-merge semantics. (commit `9fd9a3b`)

---

### Remaining gap

**P3 — Missing test for `gateId`-only channel in `list_channels`**

`packages/daemon/tests/unit/space/node-agent-tools.test.ts`, `describe('node-agent-tools: list_channels')`

The fix to `hasGate` now correctly covers both `gate` and `gateId`, but the `list_channels` describe block (lines 954–1057) only tests:
- `hasGate: false` when neither field is set
- `hasGate: true` when `gate` (inline) is set

There is no test asserting that a channel with only `gateId` set (no `gate` field) reports `hasGate: true`. The implementation is correct, but without a test, a future refactor could accidentally regress this case silently. This is purely a coverage gap — the code itself is right.

---

### Correctness assessment

- **`evaluateGateById`** — correctly fails closed when gate definition is missing; correctly falls back to default data when `gateDataRepo` is absent or has no record. The fallback-opens behavior is documented.
- **`deliverMessage` / `canDeliver`** — `gateId` takes precedence over legacy `gate`; legacy path is skipped when `gateId` is present. Both paths tested.
- **`onGateDataChanged`** — re-evaluates once per event, activates target nodes on open, correctly skips completed/cancelled runs and already-active nodes.
- **`resetOnCycle`** — atomic increment + gate data reset via `db.transaction()` when `db` is provided; sequential fallback otherwise. Both `true` and `false` gates tested independently.
- **`incrementAndResetCyclicGates` error message** — says "Message was delivered but the cycle was not counted" after an atomic-cap enforcement. This is accurate only if delivery (lazy activation) already happened before the TOCTOU race fires; in practice this ordering is correct.
- **`write_gate` fire-and-forget** — now logs via `log.warn` instead of swallowing silently. Correct.
- **`onGateDataChanged` wiring in `TaskAgentManager`** — `nodeAgentChannelRouter` is created once per sub-session spawn, not per `write_gate` call. No per-call overhead issue.
- **All 61 channel-router tests and 68 node-agent-tools tests pass.**

PR #1037, opened 2026-03-27
feat(space): define CODING_WORKFLOW_V2 template (M3.1)

lsm:
## Summary

- Adds `CODING_WORKFLOW_V2` to `built-in-workflows.ts` with 8 nodes: Planning, Plan Review, Coding, Reviewer 1/2/3, QA, Done
- 7 independent `Gate` entities with composable conditions (`check`, `count`) and `allowedWriterRoles`
- 15 `WorkflowChannel` entries using `gateId` references (M1.1 separated channel/gate model)
- Cyclic rejection paths: QA→Coding (qa-fail-gate), each Reviewer→Coding (review-reject-gate)
- 2 ungated feedback channels: Plan Review→Planning and Coding→Planning
- `seedBuiltInWorkflows` now passes `template.gates` to `createWorkflow`
- `VALID_BUILTIN_ROLES` expanded to include `'qa'`
- All 1859 daemon unit tests pass

### DIFF
--- a/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
+++ b/packages/daemon/src/lib/space/workflows/built-in-workflows.ts
@@ -278,7 +278,10 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 			agentId: 'reviewer',
 			instructions:
 				'Review the pull request for correctness, style, and test coverage. ' +
-				'Write your vote to review-votes-gate and review-reject-gate (field: votes, key: your name, value: "approved" or "rejected").',
+				'To record your vote: (1) use read_gate to fetch the current votes map from review-votes-gate, ' +
+				'(2) add your entry (key: "Reviewer 1", value: "approved" or "rejected") to the map, ' +
+				'(3) write the complete updated map back via write_gate on both review-votes-gate and review-reject-gate ' +
+				'(field: votes). Never write only your own entry — always include all existing votes to avoid overwriting peers.',
 		},
 		{
 			id: V2_REVIEWER2_STEP,
@@ -286,7 +289,10 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 			agentId: 'reviewer',
 			instructions:
 				'Review the pull request for correctness, style, and test coverage. ' +
-				'Write your vote to review-votes-gate and review-reject-gate (field: votes, key: your name, value: "approved" or "rejected").',
+				'To record your vote: (1) use read_gate to fetch the current votes map from review-votes-gate, ' +
+				'(2) add your entry (key: "Reviewer 2", value: "approved" or "rejected") to the map, ' +
+				'(3) write the complete updated map back via write_gate on both review-votes-gate and review-reject-gate ' +
+				'(field: votes). Never write only your own entry — always include all existing votes to avoid overwriting peers.',
 		},
 		{
 			id: V2_REVIEWER3_STEP,
@@ -294,7 +300,10 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 			agentId: 'reviewer',
 			instructions:
 				'Review the pull request for correctness, style, and test coverage. ' +
-				'Write your vote to review-votes-gate and review-reject-gate (field: votes, key: your name, value: "approved" or "rejected").',
+				'To record your vote: (1) use read_gate to fetch the current votes map from review-votes-gate, ' +
+				'(2) add your entry (key: "Reviewer 3", value: "approved" or "rejected") to the map, ' +
+				'(3) write the complete updated map back via write_gate on both review-votes-gate and review-reject-gate ' +
+				'(field: votes). Never write only your own entry — always include all existing votes to avoid overwriting peers.',
 		},
 		{
 			id: V2_QA_STEP,
@@ -302,7 +311,9 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 			agentId: 'qa',
 			instructions:
 				'Verify test coverage, run the CI pipeline, and confirm the PR is mergeable. ' +
-				'Write to qa-result-gate: "result: passed" if everything is green, or "result: failed" with details if issues are found.',
+				'Always write an explicit result to qa-result-gate on every QA pass — do not assume the gate was reset. ' +
+				'Write "result: passed" if everything is green, or "result: failed" with details if issues are found. ' +
+				'Writing is required on every pass because qa-result-gate retains its previous value across cycles.',
 		},
 		{
 			id: V2_DONE_STEP,
@@ -343,7 +354,11 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 		},
 		{
 			id: 'review-votes-gate',
-			description: 'All three reviewers have approved the code changes',
+			description:
+				'All three reviewers have approved the code changes. ' +
+				'Agents must read the current votes map first, add their entry, then write the full map back ' +
+				'(read-merge-write) — write_gate performs a shallow merge so writing only your own entry ' +
+				"would overwrite all other reviewers' votes.",
 			condition: { type: 'count', field: 'votes', matchValue: 'approved', min: 3 },
 			data: {},
 			allowedWriterRoles: ['reviewer'],
@@ -351,7 +366,11 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 		},
 		{
 			id: 'review-reject-gate',
-			description: 'At least one reviewer has rejected the code changes',
+			description:
+				'At least one reviewer has rejected the code changes. ' +
+				'Agents must read the current votes map first, add their entry, then write the full map back ' +
+				'(read-merge-write) — write_gate performs a shallow merge so writing only your own entry ' +
+				"would overwrite all other reviewers' votes.",
 			condition: { type: 'count', field: 'votes', matchValue: 'rejected', min: 1 },
 			data: {},
 			allowedWriterRoles: ['reviewer'],
@@ -359,7 +378,10 @@ export const CODING_WORKFLOW_V2: SpaceWorkflow = {
 		},
 		{
 			id: 'qa-result-gate',
-			description: 'QA verification has passed — tests, CI, and PR are green',
+			description:
+				'QA verification has passed — tests, CI, and PR are green. ' +
+				'resetOnCycle is intentionally false: this gate is only referenced by the non-cyclic QA→Done ' +
+				'channel, so it never auto-resets. QA must write an explicit "result" on every pass.',
 			condition: { type: 'check', field: 'result', op: '==', value: 'passed' },
 			data: {},
 			allowedWriterRoles: ['qa'],
@@ -510,7 +532,7 @@ export function getBuiltInWorkflows(): SpaceWorkflow[] {
 }
 
 /**
- * Seeds all three built-in workflow templates into the given space.
+ * Seeds all four built-in workflow templates into the given space.
  *
  * Each template step's `agentId` placeholder (e.g., `'planner'`, `'coder'`,
  * `'general'`) is resolved to a real SpaceAgent UUID via `resolveAgentId`.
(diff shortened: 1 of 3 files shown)


lsm (commented):

### Verdict: APPROVE

The implementation is clean, correct, and well-tested. All 8 nodes, 15 channels, and 7 gates are properly defined and match the task requirements. Tests pass (87 built-in-workflows tests, 53 space-runtime tests).

### What was reviewed

**`packages/daemon/src/lib/space/workflows/built-in-workflows.ts`** -- CODING_WORKFLOW_V2 template definition with:
- 8 nodes (Planning, Plan Review, Coding, Reviewer 1/2/3, QA, Done) with correct agentId role placeholders
- 7 independent Gate entities using the separated channel/gate model (check and count conditions, allowedWriterRoles, resetOnCycle)
- 15 channels with gateId references, including fan-out (Coding to 3 reviewers via shared code-pr-gate), fan-in (3 reviewers to QA via review-votes-gate), cyclic rejection paths, and 2 ungated feedback channels
- `seedBuiltInWorkflows` updated to pass `template.gates` to `createWorkflow`
- `getBuiltInWorkflows` returns 4 templates (was 3)

**`packages/daemon/tests/unit/space/built-in-workflows.test.ts`** -- Comprehensive test coverage:
- Template structure: node names, agentId placeholders, gate conditions, channel gateId references, cyclic markers, ungated channels
- Seeding: verifies all 4 workflows seed correctly with real agent IDs, gates are persisted and round-tripped, channel counts match
- Export/import round-trip tests still cover V1 workflow

**`packages/daemon/tests/unit/space/space-runtime.test.ts`** -- Updated expected workflow count from 3 to 4 in the seeding integration test

### Observations (P3 -- informational only)

1. **Shared gate data between review-votes-gate and review-reject-gate**: Both gates read from the `votes` field but with different matchValues (`approved` vs `rejected`). This is a reasonable design -- a single rejection immediately fires the reject gate while the approval gate requires unanimity. The only subtlety is that both gates evaluate simultaneously, so a rejection could trigger `review-reject-gate` (sending work back to Coding) even while other reviewers are still active. This is likely the intended workflow behavior, but worth documenting in a future milestone when the runtime evaluates these.

2. **Missing export/round-trip test for V2 template**: The existing export/import round-trip tests only cover CODING_WORKFLOW (V1). Since V2 uses `gateId` references instead of inline `gate` objects, a round-trip test for V2 would verify that gateId-based channels and separate gate entities survive serialization. This is minor since the repository layer already handles gates correctly (verified by the seeding tests).

Both observations are P3-level and do not warrant blocking the PR.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (posting as COMMENT — reviewer is PR author)

---

### Summary

The core template definition is well-structured: 8 nodes, 7 gates, and 15 channels are all correctly wired as specified. The gates use the right condition types, `resetOnCycle` values are appropriate, and `seedBuiltInWorkflows` correctly passes `gates` through. Tests are thorough and cover the structural invariants. There are no blocking bugs in the template data itself.

However, two concrete issues warrant attention before this ships.

---

### P1 — vote-accumulation breaks under shallow merge

**File:** `packages/daemon/src/lib/space/workflows/built-in-workflows.ts` (gate definitions + node instructions)

The `review-votes-gate` and `review-reject-gate` both use a `count` condition that expects `data.votes` to be a map with one entry per reviewer. The `write_gate` MCP tool calls `GateDataRepository.merge()`, which performs a **shallow merge** (documented in `gate-data-repository.ts` lines 62–69):

```
// existing: { votes: { "Reviewer 1": "approved" } }
merge({ votes: { "Reviewer 2": "approved" } })
// result:   { votes: { "Reviewer 2": "approved" } }  ← Reviewer 1 lost
```

Because `votes` is a top-level key, Reviewer 2 writing `{ votes: { "Reviewer 2": "approved" } }` replaces Reviewer 1's entire `votes` object wholesale. The gate will never accumulate 3 approved entries; `count(min:3)` will never open.

The node instructions say "key: your name, value: approved/rejected" without mentioning that the agent must first read the full existing votes map, add its entry, and write back the whole map. This is a race-prone, undocumented requirement that will silently fail at runtime.

**Suggested fix:** Update the reviewer node instructions to explicitly say something like: "Use read_gate to get the current votes map, add your entry, then write the full updated map back via write_gate." Also note this constraint in the gate descriptions.

---

### P1 — `qa-result-gate` stale data on second QA pass after rejection cycle

**File:** `packages/daemon/src/lib/space/workflows/built-in-workflows.ts` lines 360–367

`qa-result-gate` has `resetOnCycle: false`. When the `QA → Coding` cyclic channel fires via `qa-fail-gate` (which correctly has `resetOnCycle: true`), `qa-result-gate` DB data is NOT cleared and retains `{ result: "failed" }` from the first pass.

On the second QA pass, if the QA agent writes `{ result: "passed" }`, the shallow merge correctly sets `data.result = "passed"` and the gate opens. However, the template provides no indication that QA *must* write a fresh result on every pass — an agent that assumes its gate was reset and skips writing would see the stale `"failed"` value keeping the gate closed.

At minimum, a comment in the template should explain why this gate is intentionally non-resetting and that QA must always write an explicit result.

---

### P2 — stale JSDoc on `seedBuiltInWorkflows`

**File:** `packages/daemon/src/lib/space/workflows/built-in-workflows.ts` line 513

The JSDoc comment still says *"Seeds all **three** built-in workflow templates"* after the fourth was added. Should read "four".

---

### P3 — reviewer synthetic name collision (nit, low risk)

Reviewer 1, 2, and 3 all use `agentId: 'reviewer'`. `resolveNodeAgents` for a single-agentId node synthesises `name: node.agentId` (space-utils.ts line 96), so all three get synthetic agent name `"reviewer"`. The three `Reviewer N → QA` channels all resolve to the same `(fromRole:"reviewer", toRole:"qa")` routing entry. This is fine now since all three share one gate (`review-votes-gate`), but any future differentiation between reviewer nodes would break silently without any type error.

---

### What is done well

- Correct M1.1 separated gate/channel model: all 13 gated channels use `gateId` references, not legacy inline `gate`
- 4 cyclic channels correctly flagged with `isCyclic: true`; 2 ungated feedback channels correctly ungated
- `seedBuiltInWorkflows` consistently shallow-copies both `channels` and `gates`
- `VALID_BUILTIN_ROLES` expanded with `'qa'`; `'qa'` role pre-exists in `seed-agents.ts` so no runtime `resolveAgentId` failure
- 87 new tests covering nodes, gates, channels, and seeding; all 1859 daemon unit tests pass

lsm (commented):

**Recommendation: APPROVE** (own PR, posting as COMMENT)

---

### Summary of Fixes Verified

Both P1 issues from the previous round have been addressed correctly and the P2 JSDoc fix is in.

---

### P1.1 — review-votes-gate shallow-merge data loss: FIXED

All three reviewer nodes now carry a 3-step read-merge-write protocol:

1. `read_gate` to fetch the current `votes` map from `review-votes-gate`
2. Add their own entry to the map
3. Write the complete updated map back via `write_gate` on both `review-votes-gate` and `review-reject-gate`

The gate descriptions for `review-votes-gate` and `review-reject-gate` also explain the shallow-merge hazard. This matches the actual behavior in `gate-data-repository.ts` line 73 (`{ ...existing.data, ...partial }`). The fix is correct and sufficient.

New test at line 447 verifies `read_gate`, `write_gate`, and `overwriting` all appear in each reviewer node's instructions. Adequate.

---

### P1.2 — qa-result-gate stale state across cycles: FIXED

The QA node instructions now explicitly require writing a result on every pass and explain that the gate retains its value because `resetOnCycle: false`. The gate description was updated to document why `resetOnCycle` is intentionally false and to instruct QA to write explicitly.

New test at line 457 checks for `every pass` and `/do not assume/i`. The test at line 468 verifies the gate description contains `resetOnCycle` and `/intentionally false/i`. Adequate.

---

### P2 — JSDoc "three" → "four": PARTIALLY FIXED

The JSDoc in `built-in-workflows.ts` (`seedBuiltInWorkflows`) was correctly updated to say "four". However the test file header comment at lines 7–8 of `built-in-workflows.test.ts` still reads:

```
 * - getBuiltInWorkflows() returns all three templates
 * - seedBuiltInWorkflows(): seeds all three templates with real agent IDs
```

These two lines were not updated in the fix commit. Minor nit — the live test assertions at lines 493 and 593 correctly say "four" — but the module-level doc block is stale.

---

### P2 — Key-naming guidance inconsistency (newly observed)

`node-agent-tools.ts` line 812 (the `write_gate` tool description) tells agents:

> "use your **nodeId** as the map key so each node votes once"

The reviewer node instructions say:

> "key: **\"Reviewer 1\"**" (the node name, not the UUID)

These two pieces of guidance are in conflict. An agent reading only the tool description will write a UUID key; one reading only its node instructions will write a name key. Both approaches work at the evaluator level (line 273 counts `Object.values`, not keys), so this is not a correctness bug, but it creates confusion for any LLM that encounters both signals. The node instructions in `built-in-workflows.ts` should either align with the tool description ("use your nodeId") or the tool description should be updated to match ("use a stable unique key such as your node name"). This is a P2 doc-consistency issue and was not introduced by this PR (the tool description predates it), but the new reviewer instructions make the conflict more visible.

---

### Remaining nit (P3)

- `seedBuiltInWorkflows` test `beforeEach` (line 560-572) seeds `planner`, `coder`, `general`, `qa` agents in the DB but not `reviewer`, even though `roleMap` defines `reviewer: 'agent-reviewer-uuid'`. The `reviewer` agent row is never inserted. This does not cause test failures because `seedBuiltInWorkflows` uses the `resolveAgentId` callback directly and does not query the DB — but it leaves the DB slightly inconsistent with the role map. P3 cosmetic.

---

### Verdict

The two P1 fixes are solid and correctly target the root causes. The 4 new tests are meaningful and verify the right invariants. The stale "three" in the test file header (P3) and the key-naming guidance conflict between `write_gate` tool description and reviewer node instructions (P2, pre-existing) are the only remaining issues, neither blocking.

lsm (commented):

**Recommendation: APPROVE**

The CODING_WORKFLOW_V2 template is well-structured and the previous review feedback has been addressed effectively. Here is my assessment:

### What looks good

- **8-node graph** with correct topology: Planning -> Plan Review -> Coding -> Reviewer 1/2/3 (fan-out) -> QA -> Done, with cyclic rejection paths back to Coding from both reviewers and QA.
- **7 independent Gate entities** with appropriate condition types (`check` for simple gates, `count` for multi-reviewer voting), correct `allowedWriterRoles`, and sensible `resetOnCycle` settings.
- **15 channels** using `gateId` references (new separated model) rather than legacy inline `gate` objects, plus 2 ungated feedback channels.
- **Read-merge-write pattern** documented in reviewer node instructions (lines 281-284, 292-295, 303-306) and gate descriptions (lines 358-361, 369-373) -- addresses the shallow-merge overwrite concern.
- **QA explicit write requirement** documented in QA node instructions (lines 313-316) -- addresses the `resetOnCycle: false` concern with clear "do not assume the gate was reset" language.
- **`seedBuiltInWorkflows`** correctly passes `template.gates` to `createWorkflow` (line 613), and the JSDoc says "four" (line 535).
- **Tests are thorough**: 91 tests covering template structure, gate conditions, channel topology, seeding with DB round-trip, and export/import. The 4 new tests specifically verify the review feedback fixes (read-merge-write instructions, QA explicit write, gate descriptions).
- **`space-runtime.test.ts`** updated from 3 to 4 workflows.

### P3 issues (nits)

1. **Stale JSDoc in test file** (`packages/daemon/tests/unit/space/built-in-workflows.test.ts`, lines 7-8): The top-level JSDoc comment still says "returns all three templates" and "seeds all three templates" -- should be "four" to match the actual test assertions on lines 493 and 587. This was the same "three vs four" issue that was fixed in the source code but missed in the test file header.

2. **Redundant JSDoc phrasing** in `CODING_WORKFLOW` (`packages/daemon/src/lib/space/workflows/built-in-workflows.ts`, line 59): "Routing is channel-based (agent-centric model); routing is channel-based." -- duplicated clause.

Both are cosmetic and do not affect correctness or behavior.

PR #1040, opened 2026-03-27
feat(space): implement parallel node execution for shared gates (M3.3)

lsm:
## Summary

- `ChannelRouter.onGateDataChanged()` now activates multiple target nodes in **parallel** via `Promise.allSettled()` instead of a sequential for-await loop
- When `code-pr-gate` opens, all three reviewer nodes (reviewer-1, reviewer-2, reviewer-3) activate simultaneously
- Node IDs are deduplicated before spawning to avoid redundant `activateNode()` calls
- `review-votes-gate` with `count: votes.approve >= 3` correctly blocks QA until all 3 reviewers have voted; partial completion (2/3) keeps QA blocked

## Tests added (4 new)

- Parallel activation: shared gate → all 3 reviewer nodes activate simultaneously
- Idempotency: second `onGateDataChanged` on open gate does not re-activate already-active nodes
- `review-votes-gate`: QA blocked until all 3 reviewers approve (min: 3 quorum)
- `review-votes-gate`: partial completion (2/3) keeps QA blocked

### DIFF
--- a/packages/daemon/src/lib/space/runtime/channel-router.ts
+++ b/packages/daemon/src/lib/space/runtime/channel-router.ts
@@ -475,6 +475,11 @@ export class ChannelRouter {
 	 * and when enough votes accumulate to satisfy the condition, the target node is
 	 * automatically activated.
 	 *
+	 * When multiple channels share the same gate (e.g., three reviewer nodes all
+	 * gated by `code-pr-gate`), all target nodes are activated in parallel via
+	 * `Promise.allSettled()`. This ensures that reviewer1, reviewer2, and reviewer3
+	 * activate simultaneously rather than sequentially.
+	 *
 	 * @param runId  - Workflow run ID
 	 * @param gateId - ID of the gate whose data changed
 	 * @returns Array of newly activated SpaceTask records (may be empty)
@@ -496,8 +501,10 @@ export class ChannelRouter {
 		const gateResult = this.evaluateGateById(runId, gateId, workflow);
 		if (!gateResult.open) return [];
 
-		// Gate is open → activate target nodes for all gated channels
-		const activatedTasks: SpaceTask[] = [];
+		// Gate is open → collect all unique node IDs that need activation.
+		// Multiple channels may point to the same target (e.g. fan-out via node name),
+		// so deduplicate before spawning to avoid redundant activateNode() calls.
+		const nodeIdsToActivate = new Set<string>();
 		for (const channel of channels) {
 			const targets = Array.isArray(channel.to) ? channel.to : [channel.to];
 			for (const target of targets) {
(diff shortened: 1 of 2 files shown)


lsm (commented):

## Summary

This PR correctly replaces the sequential for-await loop in `ChannelRouter.onGateDataChanged()` with parallel activation via `Promise.allSettled()`, and adds node ID deduplication. The implementation is clean, the tests are comprehensive, and the change is well-scoped.

## What was changed

**`packages/daemon/src/lib/space/runtime/channel-router.ts`** (lines 504-544):
- Replaced sequential `for (channel of channels)` loop with a two-phase approach: first collect unique node IDs into a `Set<string>`, then activate all via `Promise.allSettled()`.
- Rejected activations are silently skipped (same behavior as before, but now with a comment explaining why).
- Gate evaluation is still done once (shared across all channels referencing the gate).

**`packages/daemon/tests/unit/space/channel-router.test.ts`** (+301 lines):
- 4 new tests covering: parallel activation of 3 reviewer nodes, idempotency of repeated calls, vote-counting gate blocking until quorum, and partial completion blocking.

## Assessment

**Correctness**: The implementation correctly achieves the goal. When `code-pr-gate` opens, all 3 reviewer nodes activate in parallel instead of sequentially. The deduplication via `Set<string>` correctly handles the fan-in case where `review-votes-gate` has 3 channels all pointing to the same QA node (the real workflow in `CODING_WORKFLOW_V2`).

**Error handling**: The switch from per-node try/catch to `Promise.allSettled()` is appropriate. Rejections are handled gracefully -- the comment explains the TOCTOU scenario (run transitions to terminal state between the check and activation). The caller in `node-agent-tools.ts` (line 680) already uses `void ... .catch()` (fire-and-forget), so the return value of `onGateDataChanged` is not depended upon for control flow.

**Concurrency safety**: The pre-existing idempotency in `activateNode()` (UNIQUE constraint handling) still applies. Even if two parallel `activateNode()` calls race on the same node, one wins and the other re-reads the winner's tasks. The new deduplication at the `onGateDataChanged` level reduces the chance of this race but doesn't eliminate it entirely (two rapid calls to `onGateDataChanged` could both pass the `getActiveTasksForNode` check before either creates tasks). This is acceptable because `activateNode()` is documented as idempotent and handles this case.

**Test quality**: The tests mirror the real `CODING_WORKFLOW_V2` topology. The parallel activation test correctly verifies that all 3 distinct node IDs are represented in the returned tasks. The idempotency test confirms that a second call on an already-open gate returns empty. The vote-counting tests correctly use the read-merge-write pattern documented in the workflow.

## Issues

**P2 -- Silent rejection swallowing**: When `Promise.allSettled()` returns a `rejected` result, the rejection reason is silently discarded. For debugging purposes (especially in production), it would be valuable to at least log the rejection reason at `warn` level, similar to how the caller in `node-agent-tools.ts:680-684` logs failures. Currently, if a node activation fails for a reason other than a terminal run transition (e.g., a missing node due to workflow corruption), this would be completely invisible.

Suggested fix at `channel-router.ts` line 540:
```ts
if (result.status === 'fulfilled') {
    activatedTasks.push(...result.value);
} else {
    // Consider logging: log.warn('Node activation failed:', result.reason);
}
```

**P3 -- Test could verify parallelism more explicitly**: The test name says "simultaneously" but the test only verifies that all 3 nodes were activated in the return array. It does not verify that the activations were dispatched concurrently (e.g., by using a delay-injected mock to confirm overlapping execution). The test effectively verifies correctness (all nodes activated) but not the parallel execution property. This is acceptable since `Promise.allSettled()` guarantees concurrent dispatch, but an explicit timing test would be more rigorous.

## Verdict

Clean, well-scoped implementation that correctly achieves the parallel node activation goal. The P2 is a minor observability improvement, not a correctness issue. All 65 tests pass.

**Recommendation: APPROVE**

PR #1050, opened 2026-03-27
feat(space): implement completion flow with Done node summary (M5.2)

lsm:
## Summary

- **Done node agent prompt** (`buildDoneNodeAgentPrompt`): reads `code-pr-gate`, `review-votes-gate`, `qa-result-gate` and composes a Markdown summary (PR link, reviewer votes, QA status, suggested next steps), then calls `report_done` with it
- **General agent seeded with Done prompt**: Done node in V2 workflow uses the General agent; it now gets the summary-producing instructions automatically
- **SpaceRuntime completion summary**: `resolveCompletionSummary()` identifies terminal nodes (no outbound channels) and includes their task result in the `workflow_run_completed` notification `summary` field
- **Task Agent prompt updated**: instructs the Task Agent to collect the Done node's result via `check_node_status` and pass it to `report_workflow_done`
- **Space Chat Agent updated**: instructs it to surface the `summary` field verbatim when receiving `workflow_run_completed`

## Test plan

- [x] 15 new tests for `buildDoneNodeAgentPrompt` in `custom-agent.test.ts`
- [x] 2 new tests for General agent seeding in `seed-agents.test.ts`
- [x] 3 new tests for completion summary in `space-runtime-notifications.test.ts`
- [x] All 9161 daemon tests pass

### DIFF
--- a/packages/daemon/src/lib/space/agents/custom-agent.ts
+++ b/packages/daemon/src/lib/space/agents/custom-agent.ts
@@ -633,6 +633,138 @@ export function buildQaNodeAgentPrompt(): string {
 }
 
 // ============================================================================
+// Done node agent specialized prompt
+// ============================================================================
+
+/**
+ * Build the specialized system prompt content for a Done node agent.
+ *
+ * The Done node is the terminal node in the V2 workflow. Its job is to read
+ * gate data from completed workflow stages and produce a comprehensive,
+ * human-readable summary of what was accomplished. It does NOT write code or
+ * modify files — it only reads gates and composes a summary.
+ *
+ * This is stored as the `systemPrompt` field on the General SpaceAgent preset,
+ * so it gets embedded in the "Agent Instructions" section of the broader
+ * `buildCustomAgentSystemPrompt` output.
+ *
+ * Covers:
+ *   1. Role declaration (read-only summarizer, bypass marker required)
+ *   2. Read code-pr-gate for PR URL and branch
+ *   3. Read review-votes-gate for reviewer verdicts
+ *   4. Read qa-result-gate for QA outcome
+ *   5. Compose a Markdown summary with all findings
+ *   6. Call report_done with the summary
+ *   7. Output summary with ANALYSIS_COMPLETE: bypass marker
+ */
+export function buildDoneNodeAgentPrompt(): string {
+	const sections: string[] = [];
+
+	// Role + read-only declaration
+	sections.push(
+		`You are the Done Node Agent — the final step in the workflow. ` +
+			`Your responsibility is to read the gate data written by previous workflow stages ` +
+			`and compose a comprehensive, human-readable summary of what was accomplished. ` +
+			`You do NOT write, edit, or commit any code — you only read gate data and produce a summary.`
+	);
+	sections.push(
+		`\n**IMPORTANT — This is a read-only summarization role.** ` +
+			`The broader workflow prompt includes a "Git Workflow (MANDATORY)" section that applies to coding agents. ` +
+			`As the Done node agent you must NOT create commits, push branches, or open pull requests. ` +
+			`When you finish producing the summary, use the \`ANALYSIS_COMPLETE:\` bypass marker ` +
+			`(documented in the "Bypassing Git/PR Gates for Research-Only Tasks" section) as the opening of your final response, ` +
+			`after calling \`report_done\` with the summary.`
+	);
+
+	// Step 1: Read code-pr-gate
+	sections.push(`\n## Step 1 — Read the PR Gate\n`);
+	sections.push(
+		`Read the \`code-pr-gate\` to discover the pull request that was implemented:\n\n` +
+			`\`\`\`\n` +
+			`read_gate({ gateId: "code-pr-gate" })\n` +
+			`\`\`\`\n\n` +
+			`Extract these fields from the gate data:\n` +
+			`- \`pr_url\` — the full GitHub PR URL (e.g. \`https://github.com/owner/repo/pull/42\`)\n` +
+			`- \`pr_number\` — the PR number (if present)\n` +
+			`- \`branch\` — the feature branch name (if present)\n\n` +
+			`If \`code-pr-gate\` is empty, note "No PR data available" and continue.`
+	);
+
+	// Step 2: Read review-votes-gate
+	sections.push(`\n## Step 2 — Read the Review Votes\n`);
+	sections.push(
+		`Read the \`review-votes-gate\` to see how reviewers voted on the implementation:\n\n` +
+			`\`\`\`\n` +
+			`read_gate({ gateId: "review-votes-gate" })\n` +
+			`\`\`\`\n\n` +
+			`Extract the \`votes\` map, e.g.:\n` +
+			`\`\`\`json\n` +
+			`{ "Reviewer 1": "approve", "Reviewer 2": "approve", "Reviewer 3": "approve" }\n` +
+			`\`\`\`\n\n` +
+			`Count approvals and rejections from the votes map. ` +
+			`If no votes are present, note "No review data available".`
+	);
+
+	// Step 3: Read qa-result-gate
+	sections.push(`\n## Step 3 — Read the QA Result\n`);
+	sections.push(
+		`Read the \`qa-result-gate\` to get the QA verification outcome:\n\n` +
+			`\`\`\`\n` +
+			`read_gate({ gateId: "qa-result-gate" })\n` +
+			`\`\`\`\n\n` +
+			`Extract:\n` +
+			`- \`result\` — \`"passed"\` or \`"failed"\`\n` +
+			`- \`summary\` — the QA agent's verification summary (if present)\n\n` +
+			`If \`qa-result-gate\` is empty, note "No QA data available".`
+	);
+
+	// Step 4: Compose the summary
+	sections.push(`\n## Step 4 — Compose the Workflow Summary\n`);
+	sections.push(
+		`Using the gate data collected above, compose a comprehensive Markdown summary ` +
+			`following this exact structure:\n\n` +
+			`\`\`\`markdown\n` +
+			`## Workflow Complete\n\n` +
+			`### What Was Implemented\n` +
+			`<1-3 sentences describing what was built, based on the task title and description>\n\n` +
+			`### Pull Request\n` +
+			`- **PR URL:** <pr_url from code-pr-gate, or "Not available">\n` +
+			`- **Branch:** <branch name, or "Not available">\n` +
+			`- **Status:** Ready to merge\n\n` +
+			`### Code Review\n` +
+			`- **Approvals:** <count> reviewer(s) approved\n` +
+			`- **Rejections:** <count> (should be 0 at this stage)\n` +
+			`- **Reviewers:** <list of reviewer names and their vote>\n\n` +
+			`### QA Verification\n` +
+			`- **Result:** <PASSED / FAILED>\n` +
+			`- **Details:** <QA summary from qa-result-gate.summary>\n\n` +
+			`### Suggested Next Steps\n` +
+			`1. Review the PR at <pr_url>\n` +
+			`2. Merge the pull request when ready\n` +
+			`3. <any additional context-specific suggestions>\n` +
+			`\`\`\``
+	);
+
+	// Step 5: Call report_done and output summary
+	sections.push(`\n## Step 5 — Report Done and Output Summary\n`);
+	sections.push(
+		`After composing the summary:\n\n` +
+			`1. Call \`report_done\` with the full Markdown summary as the \`summary\` argument:\n` +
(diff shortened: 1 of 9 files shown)


lsm (commented):

---

### Summary

The implementation correctly wires the completion flow: Done node agent prompt, General agent seeding, SpaceRuntime summary resolution, Task Agent prompt updates, and Space Chat Agent surfacing. The code is clean, well-documented, and follows existing patterns. Tests are comprehensive for prompt content and the summary resolution logic.

**Recommendation: REQUEST_CHANGES** for one P1 issue.

---

### P1 -- General agent has full coding toolset despite being a read-only summarizer

**File:** `packages/daemon/src/lib/space/agents/seed-agents.ts`, line 84

The General agent (now repurposed as the Done node agent) is assigned `GENERAL_TOOLS = CODER_TOOLS`, which includes `Write`, `Edit`, and all other coding tools. The Done node agent prompt explicitly instructs: "You do NOT write, edit, or commit any code -- you only read gate data and produce a summary." The prompt-level guard is the primary protection, but both QA and Reviewer agents are properly restricted at the tool level (no Write/Edit). The General agent should follow the same pattern for defense-in-depth.

Consider creating a read-only toolset for the Done node agent, similar to `QA_TOOLS`:

```ts
const DONE_TOOLS: string[] = ['Read', 'Bash', 'Grep', 'Glob', 'WebFetch', 'WebSearch'];
```

And assigning it to the General preset:
```ts
general: DONE_TOOLS,
```

The Done node only needs to call `read_gate` (an MCP tool, not in the KNOWN_TOOLS list) and `report_done` (also MCP). The base tools it needs are `Read` (in case it wants to inspect files) and possibly `Bash`/`WebFetch`/`WebSearch` for context. `Write` and `Edit` should be removed.

---

### P2 -- `resolveCompletionSummary` does not account for bidirectional workflow channels

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 809-812

The terminal node detection logic collects node IDs that appear as `ch.from` in any channel. For bidirectional channels (`A <-> B`), only node A appears in `ch.from`, so node B would be incorrectly classified as terminal even though it has an outbound path back to A.

This does not affect the current V2 workflow (which uses one-way channels), but it would produce incorrect results for any workflow that declares a bidirectional channel to the Done node. Consider also checking `ch.to` for bidirectional channels:

```ts
for (const ch of channels) {
    nodesWithOutbound.add(ch.from);
    if (ch.direction === 'bidirectional') {
        const targets = Array.isArray(ch.to) ? ch.to : [ch.to];
        for (const target of targets) {
            nodesWithOutbound.add(target);
        }
    }
}
```

---

### P2 -- Test coverage gap: no test for bidirectional channels in `resolveCompletionSummary`

**File:** `packages/daemon/tests/unit/space/space-runtime-notifications.test.ts`

The three new tests only cover one-way channels and single-node workflows. A test with a bidirectional channel to the Done node would catch the issue described above and document the expected behavior.

---

### P3 -- `space-chat-agent.ts` still references `autonomyLevel` as a payload field

**File:** `packages/daemon/src/lib/space/agents/space-chat-agent.ts`, line 207

The prompt lists `autonomyLevel` as a payload field for `workflow_run_completed`. While `autonomyLevel` IS injected into the JSON payload by `SessionNotificationSink`, it is not part of the `WorkflowRunCompletedEvent` type itself -- it is added at the message formatting layer. This is technically correct but could be confusing. No action needed since the payload does contain it.

---

### Positive observations

1. **Clean separation of concerns:** `buildDoneNodeAgentPrompt()` is stored as `systemPrompt` on the General preset and embedded via the existing `buildCustomAgentSystemPrompt` pipeline. No special-casing in the runtime.

2. **Graceful degradation:** `resolveCompletionSummary` falls back to `undefined` when no terminal task has a result, and the Space Chat Agent prompt handles the empty-summary case by falling back to `get_task_detail`.

3. **Lazy prompt generation:** The `systemPrompt` is populated at seed time, not module-init time, avoiding premature evaluation. This matches the QA agent pattern.

4. **Test quality:** The 20 new tests cover prompt content (gate names, bypass markers, Markdown structure, ordering), seeding behavior, and the summary resolution logic including the priority of terminal vs. upstream nodes.

5. **Backward compatibility:** The `summary` field was already defined on `WorkflowRunCompletedEvent` and `formatWorkflowRunCompleted` -- this PR simply populates it. No type changes or breaking changes.

lsm (commented):

### Verdict: REQUEST_CHANGES

---

### P0 -- resolveCompletionSummary compares channel from/to (node names) against node.id (UUIDs)

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts`, lines 810-827

The `resolveCompletionSummary` function collects `ch.from` values into `nodesWithOutbound` and then checks `!nodesWithOutbound.has(node.id)` to find terminal nodes. However, `WorkflowChannel.from`/`to` fields contain **agent slot name strings or node names** (e.g., `'Planning'`, `'Coder'`), not node IDs. Node IDs are UUIDs generated by `createWorkflow`. See `WorkflowChannel` type docs in `packages/shared/src/types/space.ts:737`:

```
Source agent name string (matches WorkflowNodeAgent.name) or '*' for all agents.
```

The built-in workflow templates confirm this -- they use node names like `from: 'Planning'`, `to: 'Plan Review'` (see `packages/daemon/src/lib/space/workflows/built-in-workflows.ts:405`).

**Impact:** In production, `nodesWithOutbound` will never contain any `node.id`, so **every node is treated as terminal**. The function will return the first non-empty task result from any node, not just the Done node. This means a workflow_run_completed notification might surface a Coder's intermediate result as the summary instead of the Done node's composed summary.

**Fix:** The function should resolve channel `from`/`to` names to node IDs before comparison. A straightforward approach: build a `nodeName -> nodeId` map from `workflow.nodes`, then use it to translate channel references:

```ts
const nodeNameToId = new Map<string, string>();
for (const node of nodes) {
  nodeNameToId.set(node.name, node.id);
  if (node.agents) {
    for (const agent of node.agents) {
      nodeNameToId.set(agent.name, node.id);
    }
  }
}
// Then resolve channel references:
const fromId = nodeNameToId.get(ch.from);
if (fromId) nodesWithOutbound.add(fromId);
```

The bidirectional logic at lines 813-818 needs the same translation for `ch.to` targets.

---

### P1 -- Tests use node IDs as channel from/to, masking the production bug

**File:** `packages/daemon/tests/unit/space/space-runtime-notifications.test.ts`, lines ~1403-1806

All 8 completion-summary tests construct channels with node IDs as `from`/`to` (e.g., `from: 'step-coder-sum'`, `to: 'step-done-sum'`). This matches the comparison logic in `resolveCompletionSummary` (which checks against `node.id`), so the tests pass. But it does not match real production data, where channels use node names or agent names.

After fixing `resolveCompletionSummary`, consider adding at least one test that uses **node names** (not node IDs) as channel `from`/`to` to verify the name-to-ID resolution works. This would catch regressions if the channel addressing format ever changes.

---

### Summary of previous feedback (all addressed)

All four items from the previous review round have been properly addressed:

1. **Edge cases in resolveCompletionSummary** -- Added tests for: empty result string, no terminal nodes (cycle), multiple terminal nodes with/without results, and bidirectional channels.
2. **Bidirectional channel handling** -- Both endpoints are now correctly marked as non-terminal. The `Array.isArray(ch.to)` handling is correct.
3. **General/Done agent tools** -- Changed from `CODER_TOOLS` to a read-only `DONE_TOOLS` list. The test verifies no `Write` or `Edit`.
4. **Test coverage** -- 5 new tests added (total 8 for completion summary, 1 for seed-agents tools). All 251 tests across the 3 files pass.

The bidirectional fix and DONE_TOOLS change are clean and well-tested. The P0 above is a new finding from this review round.

lsm (commented):
<review body>

lsm (commented):

### Fix Verification: Channel Name-to-NodeID Resolution

The P0 bug identified in iteration 2 has been correctly fixed. The previous code compared `channel.from` (agent slot names or node names like `"Planning"`, `"Done"`) directly against `node.id` (UUIDs), which meant terminal node detection was completely broken in production.

**The Fix (commit 116d757f):**

1. **Name-to-NodeId Map** (lines 814-822): The code now builds a `nameToNodeId` map that includes:
   - `node.name` → `node.id` (e.g., `"Done"` → `"step-done-sum"`)
   - `agent.name` → `node.id` for each agent in multi-agent nodes (e.g., `"reviewer-slot"` → the Parallel node's ID)

2. **Reference Resolution** (lines 826-834): The `resolveRef()` function handles all three channel reference formats documented in `WorkflowChannel.from`:
   - Plain names: Looked up in `nameToNodeId` map
   - Cross-node format `"nodeId/agentName"`: Extracts the UUID prefix before the `/`
   - Wildcard `'*'`: Returns `undefined` (not a specific node)

3. **Terminal Node Detection** (lines 851-857): Now correctly compares resolved node UUIDs against `node.id`.

**Test Coverage:**

All 9 completion-summary tests have been updated to use realistic node names (not UUIDs) in channel `from`/`to` fields:
- `workflow_run_completed notification includes summary from terminal (Done) node task result`
- `workflow_run_completed notification has no summary when terminal task has no result`
- `summary comes from Done terminal node, not from upstream Coder node`
- `no terminal nodes — all nodes have outbound channels, summary is undefined`
- `multiple terminal nodes — returns first non-empty result`
- `multiple terminal nodes with no results — summary is undefined`
- `empty result string on terminal node is not returned as summary`
- `bidirectional channel — neither endpoint is treated as terminal`
- **New regression test**: `channel from/to agent slot names resolve to correct node` (lines 1805-1866) - verifies that multi-agent node slot names correctly resolve to the containing node's ID

All 53 notification tests pass, confirming the fix works correctly.

**Recommendation: APPROVE**

The fix is complete and well-tested. The code correctly resolves channel endpoints through the name-to-nodeId map before comparing against node IDs, which was the root cause of the original P0 bug.

lsm (commented):

### P0 Bug Fix — Verified as Correct

The `resolveCompletionSummary` method in `packages/daemon/src/lib/space/runtime/space-runtime.ts` (lines 808-870) now correctly resolves channel `from`/`to` references to node UUIDs before comparing against `node.id`. The three-format handling is solid:

1. **Plain names** (node names or agent slot names) resolved via `nameToNodeId` map built from `node.name` and `node.agents[].name` -- correct per the `WorkflowChannel` type docs (line 721-722 of `space.ts`).
2. **Cross-node `"nodeId/agentName"`** resolved by extracting the UUID prefix before the slash -- correct per the documented format at line 728-729 of `space.ts`.
3. **`'*'` wildcard** returns `undefined` (no outbound marking) -- correct since broadcast targets all agents and should not make the source terminal.

Terminal detection logic (nodes with no outbound channels) and the task result lookup are unchanged and correct.

### Test Updates — Verified as Correct

All 8 completion-summary tests in `packages/daemon/tests/unit/space/space-runtime-notifications.test.ts` now use node names (`'Coder'`, `'Done'`, `'NodeA'`, `'Alpha'`, etc.) instead of node UUIDs in channel `from`/`to` fields. The new test at the bottom (`'channel from/to agent slot names resolve to correct node'`) specifically validates multi-agent node slot name resolution via `WorkflowNodeAgent.name`.

All 252 tests in the three changed test files pass.

### P1 Issue — Unupdated Test in role-config.test.ts

**File:** `packages/daemon/tests/unit/space/role-config.test.ts`, line 194-196

```ts
it('general has the same tools as coder', () => {
    expect(ROLE_TOOLS.general).toEqual(ROLE_TOOLS.coder);
});
```

Commit `4e30e8831` changed the `general` role from `GENERAL_TOOLS` (= `CODER_TOOLS`) to `DONE_TOOLS` (read-only, no Write/Edit), but this test assertion was not updated. It now fails because `ROLE_TOOLS.general` is `['Read', 'Bash', 'Grep', 'Glob', 'WebFetch', 'WebSearch']` while `ROLE_TOOLS.coder` includes Write and Edit.

This should be updated to assert the Done node agent has restricted read-only tools (similar to the existing `reviewer` and `qa` test patterns). The new `seed-agents.test.ts` already has the correct assertion at line 162 (`'General agent has restricted tools (no Write or Edit) — read-only Done node'`), so the fix is straightforward.

### P3 Notes (Nits)

- The `resolveRef` helper uses `indexOf('/')` rather than `split('/')` -- this is intentionally correct for the cross-node format since only the first slash matters, but a brief inline comment noting this choice would help future readers.
- The Done node agent prompt hardcodes gate IDs (`code-pr-gate`, `review-votes-gate`, `qa-result-gate`). If these gate IDs ever change, the prompt will need updating. This is acceptable for now since the workflow is defined by the same system that seeds these gates.

### Summary

The P0 name-to-UUID resolution bug is fully fixed and well-tested. The only remaining issue is a stale test assertion in `role-config.test.ts` that contradicts the intentional Done node tool restriction.

**Recommendation: REQUEST_CHANGES** (P1 issue must be fixed before merge)

PR #1055, opened 2026-03-27
test(space): code review with parallel reviewers (M7.2)

lsm:
## Summary
- New online integration test: `space-happy-path-code-review.test.ts`
- 11 tests covering Coding → code-pr-gate → 3 Reviewers (parallel) → review-votes-gate → QA
- Tests verify: parallel reviewer activation, full approval unblocks QA, partial approval stays blocked, rejection cycles back to Coding, iteration counter increments, resetOnCycle gate reset, max-iterations cap

## Key design note
Reviewer tasks must be completed (via `mockAgentDone`) before writing the reject vote so the cyclic channel can freshly activate the Coding node. The `onGateDataChanged` iteration increment only fires when `activatedTasks.length > 0`.

### DIFF
--- a/packages/daemon/tests/online/space/helpers/space-test-helpers.ts
+++ b/packages/daemon/tests/online/space/helpers/space-test-helpers.ts
@@ -370,6 +370,49 @@ export async function getTasksForNode(
 }
 
 /**
+ * Poll spaceTask.list until a NEW task appears for the given node name.
+ * A "new" task is one whose ID is NOT in `excludeTaskIds` AND has an active
+ * (non-terminal) status: pending or in_progress.
+ *
+ * Use this instead of `waitForNodeActivated` when the node was previously
+ * activated (and those tasks are now completed). `waitForNodeActivated` accepts
+ * 'completed' as a match, so it can return the old completed task instead of
+ * the freshly created one. This helper avoids that by explicitly excluding
+ * known task IDs.
+ */
+export async function waitForNewNodeTask(
+	daemon: DaemonServerContext,
+	spaceId: string,
+	runId: string,
+	nodeNameOrId: string,
+	excludeTaskIds: Set<string>,
+	timeout: number
+): Promise<SpaceTask> {
+	const deadline = Date.now() + timeout;
+	while (Date.now() < deadline) {
+		const tasks = (await daemon.messageHub.request('spaceTask.list', {
+			spaceId,
+		})) as SpaceTask[];
+
+		const match = tasks.find(
+			(t) =>
+				t.workflowRunId === runId &&
+				(t.title === nodeNameOrId ||
+					t.workflowNodeId === nodeNameOrId ||
+					t.agentName === nodeNameOrId) &&
+				!excludeTaskIds.has(t.id) &&
+				(t.status === 'pending' || t.status === 'in_progress')
+		);
+		if (match) return match;
+		await new Promise((resolve) => setTimeout(resolve, 300));
+	}
+	throw new Error(
+		`No new active task for node "${nodeNameOrId}" appeared within ${timeout}ms ` +
+			`(excluding ${excludeTaskIds.size} known task IDs)`
+	);
+}
+
+/**
  * Find tasks for a given workflow node UUID in a run.
  * Unlike getTasksForNode, this does an exact UUID match on workflowNodeId.
  */
--- a/packages/daemon/tests/online/space/space-happy-path-code-review.test.ts
+++ b/packages/daemon/tests/online/space/space-happy-path-code-review.test.ts
@@ -18,13 +18,18 @@
  *
  * ## Scenarios
  *
- * 1. Writing pr_url to code-pr-gate activates all 3 reviewers in parallel
- * 2. All 3 reviewers approve → review-votes-gate opens → QA activates
- * 3. Partial approval (2/3) → review-votes-gate stays blocked, QA not activated
- * 4. Single rejection → review-reject-gate opens → Coding re-activates (cyclic)
- * 5. Iteration counter increments when a reject cycle fires
- * 6. review-votes-gate resets after a reject cycle (resetOnCycle = true)
- * 7. Max iterations cap prevents further reject cycles
+ *  1. Writing pr_url to code-pr-gate activates all 3 reviewers in parallel
+ *  2. code-pr-gate data is readable after write
+ *  3. Partial approval (2/3) → review-votes-gate stays blocked, QA not activated
+ *  4. All 3 approve → review-votes-gate opens → QA activates
+ *  5. Single rejection → review-reject-gate opens → Coding re-activates (cyclic)
+ *  6. Iteration counter increments when a reject cycle fires
+ *  7. review-votes-gate resets after a reject cycle (resetOnCycle = true)
+ *  8. Max iterations cap prevents further reject cycles
+ *  9. Reviewers not activated before code-pr-gate is written
+ * 10. review-reject-gate opens with just 1 rejected vote (min: 1)
+ * 11. Run stays in_progress during the entire review phase
+ * 12. QA passes → qa-result-gate → Done activates (full happy-path tail)
  *
  * ## Running
  *
@@ -43,6 +48,7 @@ import {
 	readGateData,
 	approveGate,
 	waitForNodeActivated,
+	waitForNewNodeTask,
 	waitForRunStatus,
 	getTasksForNode,
 	mockAgentDone,
@@ -68,9 +74,9 @@ const TEST_TIMEOUT = IS_MOCK ? 25_000 : 90_000;
  * three reviewer nodes should be activating.
  *
  * Mirrors real agent behavior:
- *   1. Create space + start run
+ *   1. Create space + start run  (optionally set maxIterations on the workflow first)
  *   2. Open plan-pr-gate (simulate planner done)
- *   3. Wait for Plan Review to activate
+ *   3. Wait for Plan Review to activate, then complete it
  *   4. Approve plan-approval-gate (simulate reviewer approved)
  *   5. Wait for Coding to activate
  *   6. Complete the Coding task (simulate coder finishing — agent completes before writing gate)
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

Overall this is a well-structured and thorough test file. The 11 scenarios cover the key code review workflow segment (parallel reviewer activation, partial/full approval, rejection cycles, iteration counter, gate reset, max iterations cap). The test file follows established patterns from the M7.1 test file and reuses shared helpers appropriately. The documentation header and inline comments are excellent.

### P2: `waitForNodeActivated` may match stale completed tasks in reject cycle tests

**Files:** `packages/daemon/tests/online/space/space-happy-path-code-review.test.ts` (lines 324, 378, 422, 490, 594)
**Also:** `packages/daemon/tests/online/space/helpers/space-test-helpers.ts` (line 262)

`waitForNodeActivated` accepts `'completed'` as a matching status (line 262 of helpers). In reject cycle tests (Tests 5, 6, 7, 8, 10), the Coding task is completed via `mockAgentDone` before the reject cycle fires. When the test then calls `waitForNodeActivated(... 'Coding' ...)` to confirm re-activation, it can match the *old completed* Coding task instead of the newly created pending one.

This matters most in **Test 5** (line 297), where line 333 asserts:
```ts
expect(['pending', 'in_progress']).toContain(codingTask.status);
```

If `.find()` returns the old completed task, this assertion fails. The test likely passes today because task list ordering happens to return the new task first, but this is fragile and order-dependent.

For Tests 6, 7, 10, the issue is less critical because they don't assert on the returned task's status — they use the wait as a synchronization point and then check iteration count or gate data. Still, semantically these waits are not actually confirming a *new* Coding task was created.

A more robust approach would be to either:
- Filter out already-completed tasks in `waitForNodeActivated` by excluding `'completed'` when testing for *new* activation
- Track the count of Coding tasks before/after the cycle and assert the count increased
- Pass in the ID of the old task and skip it during polling

This is not blocking since the tests are passing in CI, but it introduces a latent flakiness risk.

### P3: Test 8 max-iterations assertion is lenient

**File:** `packages/daemon/tests/online/space/space-happy-path-code-review.test.ts` (lines 508-530)

The max-iterations test (Test 8) accepts two possible outcomes: either `writeGateData` throws an error, or it succeeds silently and the iteration counter stays at 1. The comment explains this is intentional ("strict enforcement OR cap enforced at activation time"), but this makes it impossible to know which enforcement path is actually being exercised. If the behavior ever changes (e.g., the gate write starts throwing but the test still passes), the test won't detect the behavioral shift.

Additionally, the 500ms sleep on line 524 for negative assertion ("Coding did not appear") is a timing-based heuristic. This is a common pattern in gate tests but worth noting as inherently probabilistic.

### Summary

Clean, well-documented test file that covers the requested scenarios comprehensively. The main concern is a latent flakiness risk in `waitForNodeActivated` matching stale completed tasks during reject cycle tests, but this does not block merge since CI is green.

lsm (commented):

Overall the test file is well-structured, clearly documented, and the 10 passing tests are genuinely useful integration coverage. However, Test 8 contains a correctness bug that makes it test the wrong invariant, and there is a secondary test quality concern in the same test.

---

### P0 — Test 8 passes for the wrong reason (broken test logic)

**File:** `packages/daemon/tests/online/space/space-happy-path-code-review.test.ts`, lines 448–453

```typescript
const { run: initialRun } = (await daemon.messageHub.request('spaceWorkflowRun.start', {
    spaceId: space.id,
    workflowId: workflow.id,
    title: 'Max-iterations test run',
    maxIterations: 1,        // <-- silently ignored
})) as { run: SpaceWorkflowRun };
```

The `spaceWorkflowRun.start` RPC handler's typed params object does not include `maxIterations`:

```typescript
// packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts line 123
const params = data as {
    spaceId: string;
    workflowId?: string;
    title: string;
    description?: string;
    goalId?: string;   // <-- no maxIterations
};
```

`startWorkflowRun()` in `space-runtime.ts` also has no `maxIterations` parameter; it reads the value directly from the workflow definition (`maxIterations: workflow.maxIterations`). The CODING_WORKFLOW_V2 template has `maxIterations: 5`, so the run is created with cap 5, not 1.

With a cap of 5 and `iterationCount` at 1 after the first cycle, the second `writeGateData('review-reject-gate', ...)` call should succeed and trigger another cycle. The test's final assertion `expect(finalRun.iterationCount).toBe(1)` would then fail — unless the Coding task from cycle 1 is still pending/in_progress, in which case `getActiveTasksForNode` returns a non-empty set and the router skips activation (and therefore skips `incrementAndResetCyclicGates`). That makes the assertion pass, but **the reason it passes is not the iteration cap** — it is the node's existing active task preventing re-activation. The test description ("Reject cycles stop being accepted once the run reaches maxIterations") is therefore false; the behavior under test is the node-idempotency guard, not the iteration cap.

To correctly test the iteration cap, the test needs one of:
- A way to start a run with `maxIterations=1` (either by exposing this param in the RPC or by patching the workflow before starting the run via `spaceWorkflow.update`), or
- An alternative approach that exhausts all 5 allowed cycles and then attempts a 6th, confirming it is blocked.

---

### P1 — Non-deterministic timing in Test 8's negative assertion

**File:** lines 522–529

```typescript
// Give a brief window to confirm Coding did not appear as a new pending task
await new Promise((resolve) => setTimeout(resolve, 500));
const { run: finalRun } = (await daemon.messageHub.request('spaceWorkflowRun.get', {
    id: runId,
})) as { run: SpaceWorkflowRun };
expect(finalRun.iterationCount).toBe(1);
```

`fireGateChanged` uses `void` and is fire-and-forget; the node activation is async. A hardcoded 500 ms sleep is not a deterministic way to assert that something did NOT happen. If the system is under load in CI the activation could arrive after 500 ms, making the assertion a false-positive. Prefer a pattern like the existing `waitForNodeActivated` with a tight timeout, or use `getTasksForNode` to confirm no new Coding task exists after a short but bounded wait.

---

### P2 — Missing coverage: QA activates after full approval, but no test verifies the full approve→QA→Done path

Tests 4 and 11 confirm QA activates and the run is `in_progress`, but no test verifies the final happy-path step (QA passing writes `qa-result-gate`, Done activates, run reaches `completed`). This is listed in the workflow's own graph and would complete the integration coverage for the segment under test. Given the established test infrastructure, adding this scenario would be straightforward.

---

### P3 — Nit: Test 8 has inline workflow-setup code that duplicates `setupToCodePrGate`

The test at line 456 manually repeats the same steps (`plan-pr-gate` write, `waitForNodeActivated('Plan Review')`, `approveGate`, `waitForNodeActivated('Coding')`, `mockAgentDone`, `writeGateData('code-pr-gate', ...)`) that `setupToCodePrGate` already encapsulates. If the `maxIterations` issue is fixed (e.g., via workflow patching), Test 8 could call `setupToCodePrGate` for the first leg and keep the rest of the test focused on the cap assertion.

---

### What works well

- The `setupToCodePrGate` helper cleanly abstracts the common setup path.
- Tests 1–7, 9–11 correctly use `waitForNodeActivated` as the synchronization point after `writeGateData`.
- The comment explaining why `mockAgentDone` must be called before writing reject votes (lines 77–83 and 118–123) is accurate and genuinely useful.
- The test correctly uses `Promise.all` to assert parallel (not sequential) reviewer activation in Tests 1, 3, 4, 5, 6, 7, 10.
- The `resetOnCycle` assertion in Test 7 is correctly structured: it waits for Coding activation before reading the reset gate, which ensures the reset has already occurred.
- CI matrix already covers `tests/online/space` as a directory-level glob, so the new file is automatically picked up with no matrix changes needed.

lsm (commented):

**Recommendation: APPROVE**

This re-review focuses on whether the four issues raised in the prior reviews (1 P0, 1 P1, 2 P2 from GLM; 1 P2 and 1 P3 from Opus) were properly addressed in commit `f17e07f6`. All issues have been resolved correctly.

### Prior P0 (GLM): Test 8 maxIterations silently ignored -- FIXED

The original Test 8 passed `maxIterations: 1` directly to `spaceWorkflowRun.start`, which silently ignored it (the RPC handler does not accept that param). The fix correctly patches the workflow via `spaceWorkflow.update` before starting the run (lines 104-109 of `setupToCodePrGate`). I verified the full chain:

- `UpdateSpaceWorkflowParams` includes `maxIterations?: number | null` (`packages/shared/src/types/space.ts:1004`)
- `space-workflow-repository.ts:288-290` persists `maxIterations` to the DB on update
- `space-runtime.ts:340` reads `workflow.maxIterations` when creating the run

The test now also asserts `initialRun.maxIterations === 1` (line 501) to confirm the cap was applied, and properly exercises the second reject cycle by completing the Coding task and reviewer tasks before attempting the blocked rejection.

### Prior P1 (GLM): 500ms sleep replaced with bounded poll -- FIXED

The negative assertion in Test 8 (lines 570-582) now uses a bounded polling loop over `NODE_ACTIVATION_TIMEOUT` that checks for unexpected new Coding tasks by filtering against known task IDs. The loop exits early if an unexpected task appears, providing a deterministic failure. This is a proper fix.

### Prior P2 (Opus): waitForNewNodeTask helper added -- FIXED

A new `waitForNewNodeTask` helper (lines 383-413 of `space-test-helpers.ts`) was added that explicitly excludes known task IDs and only matches active statuses (`pending` or `in_progress`). This directly addresses the concern that `waitForNodeActivated` could match old completed tasks. Tests 5, 6, 7, 8, and 10 now use this via the `triggerRejectCycle` helper (lines 162-196), which collects pre-cycle Coding task IDs and passes them to `waitForNewNodeTask`.

### Prior P2 (GLM): Test 12 added for full happy-path tail -- FIXED

Test 12 (lines 683-744) verifies the QA-to-Done segment: QA writes `passed` to `qa-result-gate`, Done node activates, run stays `in_progress`. This completes the integration coverage for the workflow segment under test.

### Prior P3 (GLM): Test 8 refactored to use setupToCodePrGate -- FIXED

Test 8 now uses `setupToCodePrGate(daemon, { maxIterationsOverride: 1 })` (line 493) and the `triggerRejectCycle` helper, eliminating the duplicated inline setup code.

The `triggerRejectCycle` helper (line 172) completes all three reviewer tasks every time, even though only one reviewer needs to reject. This is functionally correct (ensures no active tasks block the cyclic channel) but does more work than strictly necessary. Not blocking.

Overall, this is a clean set of fixes that address every concern raised. The test file is well-structured with 12 comprehensive scenarios covering the code review workflow segment.

lsm (commented):

**Recommendation: APPROVE**

All three previously flagged issues have been correctly addressed. Here is what I verified.

---

### Issue 1 (Test 8 — maxIterations cap): FIXED

The fix correctly patches the workflow via `spaceWorkflow.update` with `maxIterations: 1` before calling `spaceWorkflowRun.start`. The flow is validated: `space-runtime.ts` line 340 reads `workflow.maxIterations` when creating the run, and `channel-router.ts` enforces `run.iterationCount >= run.maxIterations` (lines 313, 392, 507). The `setupToCodePrGate` helper's `maxIterationsOverride` path at lines 104-109 of the test file correctly uses this mechanism.

The test also verifies `initialRun.maxIterations === 1` immediately after setup, providing an early guard against the workflow update being silently ignored.

### Issue 2 (stale completed task matching): FIXED

`waitForNewNodeTask` was added to `space-test-helpers.ts` (lines 383-413). It filters to only `pending | in_progress` status and excludes known task IDs via a `Set`. The `triggerRejectCycle` helper in the test file uses it consistently (line 188-195). The existing `waitForNodeActivated` still accepts `completed` as a valid status (line 262 of helpers) so this risk is correctly scoped — the new helper is used exactly where the stale-match bug would have occurred.

### Issue 3 (Test 12 — QA to Done path): FIXED

Test 12 at lines 683-744 exercises the full happy-path tail. It correctly completes all three reviewer tasks via `mockAgentDone` before writing the approval votes (which avoids the same stale-task problem for reviewers), then waits for QA activation, completes the QA task, and writes `{ result: 'passed' }` to `qa-result-gate`. The gate's `condition` in `built-in-workflows.ts` line 387 is `{ type: 'check', field: 'result', op: '==', value: 'passed' }`, which matches. The Done node assertion correctly checks that the run is still `in_progress` (Done agent hasn't finished), which is the right invariant.

---

### Minor observations (no blocking issues)

**P3 — `getTasksForNodeId` is exported but unused** (`packages/daemon/tests/online/space/helpers/space-test-helpers.ts` line 419). This function exports fine under knip's `ignore: **/packages/daemon/tests/**` rule, but it is dead code in the current test suite. No action required for merge.

**P3 — Test 8 reviewer completion is a single-shot fetch (lines 546-553).** After the second `writeGateData` for `code-pr-gate`, reviewer activation is fire-and-forget, so the immediate `getTasksForNode` call might find zero active reviewer tasks if activation hasn't completed yet. This doesn't break the test — the iteration cap check in `channel-router.ts` runs before the node-idempotency guard, so the second rejection is blocked regardless of reviewer task state. The test passes the negative assertion correctly.

---

All quality checks pass: 0 lint warnings, 0 TypeScript errors, knip clean.

PR #1058, opened 2026-03-27
test(space): QA completion flow online tests (M7.3)

lsm:
Add 11 deterministic integration tests for the QA→Done and QA→Coding segments of CODING_WORKFLOW_V2.

Tests cover: QA pass activates Done, QA fail cycles back to Coding, iteration counter increment, gate resets (qa-result, qa-fail, review-votes all reset), reviewers must re-vote after QA fail, partial re-review blocks QA, full end-to-end QA fail → re-review → QA re-activates, and max iterations cap.

### DIFF
--- a/packages/daemon/tests/online/space/space-happy-path-qa-completion.test.ts
+++ b/packages/daemon/tests/online/space/space-happy-path-qa-completion.test.ts
@@ -1,0 +1,778 @@
+/**
+ * Space Happy Path — QA Completion Flow
+ *
+ * Integration tests for the QA → Done (pass) and QA → Coding (fail) segments
+ * of CODING_WORKFLOW_V2.
+ *
+ * No real LLM sessions are started. Gate data is written directly via RPC to
+ * simulate agent actions, exercising the gate/channel machinery deterministically.
+ *
+ * ## Workflow segment under test
+ *
+ *   review-votes-gate  (all 3 reviewers approve)
+ *     └─► QA node activates
+ *   qa-result-gate     (QA writes result: passed)
+ *     └─► Done node activates → workflow completes
+ *   qa-fail-gate       (QA writes result: failed)
+ *     └─► Coding re-activates (cyclic), iteration counter increments
+ *         └─► code-pr-gate must be re-written → 3 Reviewers re-activate
+ *             └─► review-votes-gate must be re-satisfied (all 3 re-vote)
+ *                 └─► QA re-activates
+ *
+ * ## Scenarios
+ *
+ *  1. QA passes → qa-result-gate opens → Done node activates
+ *  2. Done node does not activate before qa-result-gate is written
+ *  3. QA failure → qa-fail-gate opens → Coding re-activates (cyclic)
+ *  4. Iteration counter increments on QA → Coding cycle
+ *  5. qa-result-gate resets after a QA fail cycle (resetOnCycle: true)
+ *  6. qa-fail-gate resets after cycle (resetOnCycle: true)
+ *  7. After QA fail cycle: Reviewers must re-activate (review-votes-gate reset)
+ *  8. After QA fail cycle: All 3 reviewers must re-approve before QA re-activates
+ *  9. Full QA fail → re-review → QA re-activates end-to-end
+ * 10. Run remains in_progress throughout QA phase
+ * 11. Max iterations cap prevents further QA fail cycles
+ *
+ * ## Running
+ *
+ *   NEOKAI_USE_DEV_PROXY=1 bun test \
+ *     packages/daemon/tests/online/space/space-happy-path-qa-completion.test.ts
+ */
+
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import type { DaemonServerContext } from '../../helpers/daemon-server';
+import { createDaemonServer } from '../../helpers/daemon-server';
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured test file that follows established patterns from the sibling test files (M7.1, M7.2). The 11 tests comprehensively cover the QA completion flow segments of CODING_WORKFLOW_V2.

### What works well

- **Consistent patterns**: The file mirrors the structure of `space-happy-path-code-review.test.ts` — same imports, timing constants, beforeEach/afterEach lifecycle, and helper usage.
- **Good test isolation**: Each test creates its own space/workflow/run via `setupToQaActivated`, preventing cross-test contamination.
- **Correct `maxIterations` flow**: Test 11 updates the workflow's `maxIterations` before starting the run, and `space-runtime.ts:340` confirms `maxIterations` is correctly copied from the workflow to the run at creation time.
- **Proper use of `waitForNewNodeTask`**: After cyclic transitions (QA fail -> Coding), the tests correctly collect pre-cycle task IDs and use `waitForNewNodeTask` with exclusion sets, avoiding the false-positive match issue documented in the helper's JSDoc.
- **Negative assertions are thorough**: Tests 2, 7, 8, and 11 all properly verify that nodes do NOT activate when gates are unsatisfied, using either immediate checks or polling with timeout.

### Minor observations (P3)

1. **`space-happy-path-qa-completion.test.ts:505`** — The 500ms sleep to verify QA does not re-activate (test 8) is a pragmatic approach for a negative assertion, and consistent with the pattern used in test 11's polling loop. The polling approach in test 11 (line 756-766) is slightly more robust since it checks repeatedly over the full timeout window. This inconsistency is cosmetic — both work.

2. **`space-happy-path-qa-completion.test.ts:357-359`, `379-382`** — Tests 5 and 6 check gate reset with `if (gate !== null) { expect(gate.data.result).toBeUndefined(); }`. If the implementation returns `null` after reset, the assertion is effectively skipped. This is fine since `null` semantically means "reset/empty", but a comment clarifying that `null === reset` would improve readability.

3. **`space-happy-path-qa-completion.test.ts:74`** — The `SpaceTask` type alias uses `Awaited<ReturnType<typeof getTasksForNode>>[number]` instead of importing `SpaceTask` directly from `@neokai/shared` (which is already imported in the helpers file). This works correctly but the indirection is unnecessary — `SpaceTask` from shared would be equivalent and more direct.

No correctness issues, security concerns, or missing edge cases found. The tests pass and the implementation matches the task description.

lsm (commented):

**Recommendation: APPROVE** (own PR — posted as COMMENT)

---

### Summary

The implementation correctly tests all 11 scenarios for the QA→Done (pass) and QA→Coding (fail) flows. The tests are deterministic, well-structured, and consistent with patterns established in the M7.1 and M7.2 test files. The code compiles cleanly and the new test file is already picked up by the CI matrix via `test_path: tests/online/space`.

---

### Issues Found

#### P3 — Misleading comment about synchronous activation (test 8 and 9, lines 448–450 and 537–542)

Both tests contain the comment: "writeGateData triggers reviewer activation synchronously within the RPC". This is inaccurate. Looking at `space-workflow-run-handlers.ts` line 116–119, `fireGateChanged` is fire-and-forget (`void spaceRuntimeService.notifyGateDataChanged(...).catch(...)`). Activation happens asynchronously after the RPC call returns. The safe pattern of collecting IDs before writing is a correct conservative choice regardless of async/sync ordering — but the comment explaining *why* this matters is factually wrong and could mislead future maintainers into changing the pattern based on incorrect assumptions.

#### P3 — Test 5 creates an unintended side effect: Done node activation (line 348)

Test 5 writes `result: passed` to `qa-result-gate` as setup for testing the gate reset. This immediately opens the gate and triggers async activation of the Done node via `fireGateChanged`. The test doesn't acknowledge or assert on this side effect. While this doesn't cause test failures (the side effect is benign and isolated within the fresh daemon instance), the comment "shouldn't normally happen in the happy path" undersells the actual consequence — the Done node will be activated with an in-flight task. Any future assertion on Done node state in this test would observe an unexpected task. Consider noting this side effect in the comment, or alternatively writing qa-result-gate with data that does NOT satisfy the gate condition (e.g. `{ result: 'draft' }`).

---

### What Works Well

- **setupToQaActivated helper is well-factored.** Extracting the path to QA activation into a shared fixture avoids 11x duplication and makes each test focused on its specific scenario.
- **triggerQaFailCycle abstracts the cycle mechanics cleanly.** The comment about the ordering of mockAgentDone before writeGateData (iteration counter only increments on newly-activated nodes) correctly documents a non-obvious invariant.
- **waitForNewNodeTask pattern is correctly applied.** All cyclic re-activation scenarios use `waitForNewNodeTask` with pre-cycle ID sets rather than `waitForNodeActivated`, correctly avoiding false positives from old completed tasks.
- **Max iterations test (test 11) is thorough.** The test drives the full second cycle to QA before attempting the capped fail, mirroring exactly how the cap would be hit in production and confirming the polling-based negative assertion does not produce false passes.
- **Gate reset assertions use the correct conditional pattern.** Testing `if (gate !== null)` before asserting on `gate.data` is correct: the reset may produce a null record (deleted) or an empty record, and both satisfy the reset semantics.
- **CI integration is automatic.** The new file falls under `tests/online/space` which is already in the CI matrix — no workflow changes needed.

PR #1063, opened 2026-03-27
fix(e2e): fix P1 bugs in space E2E tests (M9.3)

lsm:
## What

Two P1 bugs fixed in the space E2E test suite.

**Bug 1 — Promise.race() with swallowed errors** (`space-happy-path-pipeline.e2e.ts`):
The original `Promise.race()` used `.catch(() => {})` handlers that silently returned `undefined`. When none of the three elements became visible within 5 s, `Promise.race()` resolved with `undefined`, causing a confusing `expected ['error', 'diff', 'no-files'] to contain undefined` failure. Replaced with `Promise.any()`, which rejects with an `AggregateError` when all promises fail — giving a clear, actionable error message.

**Bug 2 — Short timeout for server-dependent assertion** (`space-approval-gate-rejection.e2e.ts`):
`gate-icon-waiting_human` `toBeHidden` was using a 5000ms timeout, but the state change requires a server round-trip (RPC call → gate update → WebSocket event → UI re-render). Every other server-dependent assertion in the suite uses 10000ms. Increased to match.

### DIFF
--- a/packages/e2e/tests/features/space-approval-gate-rejection.e2e.ts
+++ b/packages/e2e/tests/features/space-approval-gate-rejection.e2e.ts
@@ -300,6 +300,6 @@ test.describe('Approval Gate Rejection', () => {
 		// pre-coding phase (plan-approval-gate). If the workflow adds more human gates
 		// in the future, this assertion would need revisiting — but for the current
 		// template it confirms the gate correctly transitioned away from waiting_human.
-		await expect(page.getByTestId('gate-icon-waiting_human')).toBeHidden({ timeout: 5000 });
+		await expect(page.getByTestId('gate-icon-waiting_human')).toBeHidden({ timeout: 10000 });
 	});
 });
--- a/packages/e2e/tests/features/space-happy-path-pipeline.e2e.ts
+++ b/packages/e2e/tests/features/space-happy-path-pipeline.e2e.ts
@@ -281,19 +281,13 @@ test.describe('Space Happy Path Pipeline', () => {
 		const diffSummaryEl = page.getByTestId('diff-summary');
 		const noFilesEl = page.getByTestId('no-files');
 
-		const state = await Promise.race([
-			errorEl
-				.waitFor({ state: 'visible', timeout: 5000 })
-				.then(() => 'error' as const)
-				.catch(() => {}),
-			diffSummaryEl
-				.waitFor({ state: 'visible', timeout: 5000 })
-				.then(() => 'diff' as const)
-				.catch(() => {}),
-			noFilesEl
-				.waitFor({ state: 'visible', timeout: 5000 })
-				.then(() => 'no-files' as const)
-				.catch(() => {}),
+		// Promise.any resolves with the first fulfilled value; if all three time out it
+		// rejects with an AggregateError, giving a clear failure instead of the silent
+		// `undefined` that Promise.race with swallowed .catch() would return.
+		const state = await Promise.any([
+			errorEl.waitFor({ state: 'visible', timeout: 5000 }).then(() => 'error' as const),
+			diffSummaryEl.waitFor({ state: 'visible', timeout: 5000 }).then(() => 'diff' as const),
+			noFilesEl.waitFor({ state: 'visible', timeout: 5000 }).then(() => 'no-files' as const),
 		]);
 
 		// Any terminal state is acceptable for an E2E test without a real worktree.


lsm (commented):

Two targeted, well-reasoned fixes. Both are correct and address real P1 bugs. A few observations below.

### Bug 1 -- Promise.race() to Promise.any() (LGTM)

**File:** `packages/e2e/tests/features/space-happy-path-pipeline.e2e.ts` (lines 284-291)

The fix is correct. `Promise.race()` with `.catch(() => {})` on each branch is a well-known footgun: when all promises reject, the race resolves with `undefined` (the settled value from each catch), producing the confusing `expected ['error', 'diff', 'no-files'] to contain undefined` assertion failure. `Promise.any()` is the right primitive here -- it short-circuits on the first fulfillment and rejects with an `AggregateError` when all fail, which surfaces the actual timeout errors.

The comment explaining the rationale is helpful for future maintainers.

### Bug 2 -- Timeout 5000ms to 10000ms (LGTM)

**File:** `packages/e2e/tests/features/space-approval-gate-rejection.e2e.ts` (line 303)

The fix is consistent with the file's own documented timeout convention (line 33: `10000ms: state transitions requiring a server round-trip`). The `gate-icon-waiting_human` `toBeHidden` assertion requires a full server round-trip (gate rejection RPC -> run status update -> WebSocket event -> UI re-render), so 5000ms was indeed too tight.

### Additional observations (not blocking, for awareness)

1. **Test 3 in `space-approval-gate-rejection.e2e.ts` (line 244-245):** The `needs_attention` banner assertion after `rejectViaPopup` still uses `timeout: 5000`. This is also a server-dependent state transition (the banner text derives from `run.failureReason` which comes via WebSocket). Since `rejectViaPopup` already awaits `gate-icon-blocked` with 10000ms, the server state is likely propagated by the time this assertion runs, so the 5000ms timeout is probably safe in practice. However, if the banner rendering has its own async path (e.g., a separate live-query subscription), it could exhibit the same flakiness. Worth keeping an eye on if this test flakes.

2. **`rewind-features.e2e.ts` (line 34-41):** Uses the same `Promise.race` + `.catch(() => {})` pattern for SDK crash handling. This is arguably acceptable because the intent is "fire and forget" -- it only needs one of the two branches to succeed, and if neither does, the test will likely fail on a subsequent assertion. But it has the same silent-failure risk where both branches could reject and the test continues without knowing. Not in scope for this PR, but worth noting for a follow-up cleanup.

### Verdict

Both fixes are correct, minimal, and well-documented. No issues found.

**Recommendation: APPROVE**

lsm (commented):

### Summary

Both fixes are correct and well-implemented. The changes are minimal, focused, and follow existing patterns in the codebase.

---

### Bug 1: Promise.race() with swallowed errors

**File:** `packages/e2e/tests/features/space-happy-path-pipeline.e2e.ts` (lines 281-291)

**Fix:** Replaced `Promise.race()` with silently swallowed errors (`catch(() => {})`) with `Promise.any()`.

**Assessment:** This is a solid improvement. The original code had a subtle bug where all three promises could timeout and return `undefined`, causing the subsequent assertion `expect(['error', 'diff', 'no-files']).toContain(state)` to fail with the confusing message `expected ['error', 'diff', 'no-files'] to contain undefined`. Using `Promise.any()` ensures that if all three promises reject, an `AggregateError` is thrown with all rejection reasons, providing actionable debugging information.

The comment explaining the change is clear and helpful.

---

### Bug 2: Short timeout for server-dependent assertion

**File:** `packages/e2e/tests/features/space-approval-gate-rejection.e2e.ts` (line 303)

**Fix:** Increased timeout from 5000ms to 10000ms for the `gate-icon-waiting_human` `toBeHidden` assertion.

**Assessment:** This is the correct fix. The file's header comments (lines 31-34) clearly document the timeout conventions:

- **10000ms**: state transitions requiring a server round-trip (gate data load, run status)
- **5000ms**: UI-only changes (popup visibility, overlay open/close, tab content)

The `toBeHidden` assertion for the gate icon requires a server round-trip (rejection RPC -> gate update -> WebSocket event -> UI re-render), so 10000ms is the appropriate timeout. All other server-dependent assertions in this file already use 10000ms, so this change brings consistency.

---

### Minor Observation (P3)

I noticed there's another use of `Promise.race()` with `.catch(() => {})` in `packages/e2e/tests/features/rewind-features.e2e.ts:35`. However, this is a different pattern - it's used for crash recovery where either outcome is acceptable (message processed OR SDK crash recovery). The function doesn't rely on the race's return value and subsequent assertions verify the state, so this is a legitimate use case. No change needed there.

---

### Recommendation: APPROVE

Both fixes correctly address P1 issues with minimal, focused changes that follow established patterns in the codebase.

PR #1065, opened 2026-03-27
fix(test): fix 3 integration test bugs for M9.2

lsm:
## Summary

- **Bug 1 (P1)**: Devproxy adopted from a different git worktree uses stale mocks — task-agent-lifecycle tests expected `spawn_node_agent`/`check_node_status` but got old tool names from a stale proxy. Fixed by checking the running proxy's config file via the devproxy REST API and restarting it if it was started from a different worktree.
- **Bug 2 (P1)**: `git branch -D space/code-implementation` fails when the branch is checked out in another active worktree (concurrent coder runs). Fixed by running `git worktree prune` after the first deletion failure to remove stale references, then retrying. Added unit test covering this path.
- **Bug 3 (P2)**: GLM SDK tests used `if (!GLM_API_KEY) { return; }` silent skip guards, violating the CLAUDE.md hard-fail rule. Fixed to `throw new Error(...)` so CI fails visibly when credentials are missing.

### DIFF
--- a/packages/daemon/src/lib/space/managers/space-worktree-manager.ts
+++ b/packages/daemon/src/lib/space/managers/space-worktree-manager.ts
@@ -89,15 +89,38 @@ export class SpaceWorktreeManager {
 			});
 			if (branches.trim().length > 0) {
 				this.logger.warn(`Stale branch detected: ${branchName} — deleting before recreating`);
-				execFileSync('git', ['branch', '-D', branchName], {
-					cwd: space.workspacePath,
-					timeout: 30_000,
-				});
+				try {
+					execFileSync('git', ['branch', '-D', branchName], {
+						cwd: space.workspacePath,
+						timeout: 30_000,
+					});
+				} catch {
+					// Branch may be in use by a worktree from a stale git reference.
+					// Prune stale worktree entries (those whose paths no longer exist),
+					// then retry the deletion.
+					try {
+						execFileSync('git', ['worktree', 'prune'], {
+							cwd: space.workspacePath,
+							timeout: 30_000,
+						});
+						execFileSync('git', ['branch', '-D', branchName], {
+							cwd: space.workspacePath,
+							timeout: 30_000,
+						});
+					} catch (retryErr) {
+						// Still failed — branch is in use by an active worktree from another
+						// context (e.g., a concurrent agent run). Log and continue; the
+						// worktree add below will fail and throw a clear error.
+						this.logger.warn(
+							`Failed to delete stale branch ${branchName} after worktree prune: ${retryErr instanceof Error ? retryErr.message : String(retryErr)}`
+						);
+					}
+				}
 			}
 		} catch (err) {
-			// Non-fatal: branch check/delete failure should not block worktree creation
+			// Non-fatal: branch check failure should not block worktree creation
 			this.logger.warn(
-				`Failed to check/delete stale branch ${branchName}: ${err instanceof Error ? err.message : String(err)}`
+				`Failed to check stale branch ${branchName}: ${err instanceof Error ? err.message : String(err)}`
 			);
 		}
 
--- a/packages/daemon/tests/helpers/dev-proxy.ts
+++ b/packages/daemon/tests/helpers/dev-proxy.ts
@@ -360,6 +360,29 @@ export function createDevProxyController(options: DevProxyOptions = {}): DevProx
 		});
 	};
 
+	// Helper to get the running proxy's config file path via the devproxy REST API.
+	// Only applicable when the proxy is listening on the default port (8000), since that
+	// is the only port on which the devproxy API (port 8897) is meaningful. For non-default
+	// ports (e.g., unit-test fake TCP servers), this returns null to skip the check.
+	const fetchRunningProxyConfigFile = async (): Promise<string | null> => {
+		// Only check config for the default devproxy port. Non-default ports are used in unit
+		// tests with fake TCP servers; they should always be adopted as-is.
+		if (port !== 8000) return null;
+
+		// The devproxy REST API listens on port 8897 by default (not the proxy port).
+		const apiPort = 8897;
+		try {
+			const response = await fetch(`http://127.0.0.1:${apiPort}/proxy`, {
+				signal: AbortSignal.timeout(2000),
+			});
+			if (!response.ok) return null;
+			const data = (await response.json()) as { configFile?: string };
+			return data.configFile ?? null;
+		} catch {
+			return null;
+		}
+	};
+
 	// Helper to check if proxy is responding
 	const checkProxyReady = async (): Promise<boolean> => {
 		// Try to connect to the proxy port using a TCP connection check
@@ -434,12 +457,30 @@ export function createDevProxyController(options: DevProxyOptions = {}): DevProx
 			// If so, adopt it as an external instance instead of trying to start a new
 			// one (which would fail with "already running" and cause test failures).
 			if (await checkProxyReady()) {
-				running = true;
-				external = true;
-				if (setEnvVars) {
-					setProxyEnvVars();
+				// Before adopting the running proxy, verify it was started with the same
+				// config file as we expect. When running across multiple git worktrees
+				// (e.g., concurrent coder agents), each worktree has its own mocks.json.
+				// If the running proxy was started from a different worktree, its mocks
+				// will be stale — stop it so we can restart with the correct config.
+				const runningConfig = await fetchRunningProxyConfigFile();
+				if (runningConfig !== null && runningConfig !== configPath) {
+					// Stale proxy from a different context — stop it and restart below.
+					await runDevProxyCommand(['stop'], 5000);
+					const stopStart = Date.now();
+					while (Date.now() - stopStart < 5000) {
+						if (!(await checkProxyReady())) break;
+						await sleep(100);
+					}
+					// Fall through to start a fresh proxy with our config.
+				} else {
+					// Same config (or couldn't determine) — adopt as external.
+					running = true;
+					external = true;
+					if (setEnvVars) {
+						setProxyEnvVars();
+					}
+					return;
 				}
-				return;
 			}
 
 			// Check if devproxy is installed
--- a/packages/daemon/tests/helpers/dev-proxy.ts
+++ b/packages/daemon/tests/helpers/dev-proxy.ts
@@ -518,6 +518,17 @@ export function createDevProxyController(options: DevProxyOptions = {}): DevProx
 				// with another process starting the proxy simultaneously), the proxy may
 				// already be up.  Do one final port check before giving up.
 				if (await checkProxyReady()) {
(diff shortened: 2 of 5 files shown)


lsm (commented):

---

### Bug 1 (P1) — Stale devproxy mocks across git worktrees

The fix in `packages/daemon/tests/helpers/dev-proxy.ts` (lines 363-384, 459-483) is well-structured. The approach of querying the devproxy REST API on port 8897 to get the running proxy's config file and comparing it against the current worktree's config path is a correct root-cause fix.

**One concern (P2):** After stopping the stale proxy (line 468) and waiting for the port to become free (lines 469-473), the code falls through to start a fresh proxy. However, there is no config-file verification on the *newly started* proxy. If a concurrent worktree starts its own proxy between the stop and the fresh start, the race would result in adopting that concurrent proxy (which may have different mocks) via the fallback at line 520-526 (non-zero exit code but port is up). This is unlikely in practice (the window is small and the fallback is marked `external = true`), but worth noting for completeness. The existing race comment at line 517-519 acknowledges this is an accepted limitation.

**The port-8000 guard at line 370 is a good defensive measure** — it ensures unit tests with fake TCP servers on random ports are never affected by this logic.

Verdict: Sound fix.

---

### Bug 2 (P1) — Stale branch blocking worktree creation

The fix in `packages/daemon/src/lib/space/managers/space-worktree-manager.ts` (lines 92-118) correctly wraps the first `git branch -D` in a try/catch, then runs `git worktree prune` and retries. The retry failure is handled gracefully (logs a warning and continues, letting the subsequent `git worktree add` fail with a clear error).

**The unit test (lines 235-271 in the test file) is thorough.** It correctly simulates the exact failure scenario: creates a worktree, deletes its directory without pruning, verifies the stale reference exists, then calls `createTaskWorktree` and asserts the full recovery path. The test verifies:
- The stale branch is detected
- The prune-and-retry path succeeds
- The worktree is created with the correct slug
- The branch is recreated

One minor note: the test hardcodes `slug = 'prune-test'` and verifies it against `result.slug` rather than calling `worktreeSlug('prune test', 77)`. This is fine since the slugification of 'prune test' is trivially 'prune-test', but using the helper would be marginally more robust against future slug changes. P3-level observation only.

**P2 suggestion:** The same stale-worktree-reference issue exists in two other places in the same file:
- `cleanupOrphaned()` at line 293 — `git branch -D` with a silent catch
- `removeTaskWorktree()` at line 200 — `git branch -D` with a silent catch

Both could fail for the same reason (branch locked by a stale worktree reference). While these are pre-existing issues and not regressions, they represent the same class of bug that this PR fixes. Consider applying the same prune-retry pattern there in a follow-up.

---

### Bug 3 (P2) — GLM SDK hard-fail rule violation

The fix in `packages/daemon/tests/online/glm/glm-sdk-minimal.test.ts` (lines 175, 208, 238) correctly changes `return;` to `throw new Error(...)` per the CLAUDE.md credential hard-fail rule. The error messages are clear and descriptive. The doc comment at line 29 is also updated to reflect the new behavior.

**P1 issue: The same violation exists in `packages/daemon/tests/online/glm/glm-provider.test.ts`** at lines 452-454 and 502-504, where two tests use the identical `if (!glmApiKey) { console.log('Skipping...'); return; }` pattern. These were not fixed in this PR. Since the PR description explicitly scopes this to fixing GLM tests that violate the hard-fail rule, and two instances remain, this should either be included in this PR or called out as a known follow-up.

---

### Summary

All three fixes are correct and address the stated root causes. The devproxy stale-mock fix and the worktree prune-retry fix are well-implemented with appropriate defensive measures. The GLM hard-fail fix is incomplete — two more instances exist in `glm-provider.test.ts`. The worktree prune-retry pattern should ideally be applied to `cleanupOrphaned` and `removeTaskWorktree` in a follow-up.

lsm (commented):

### Summary

Three bug fixes for integration tests. Two are solid (worktree prune, GLM hard-fail), but one has a residual race condition risk.

---

### Bug 1 (P1): Devproxy stale mocks across worktrees

**File:** `packages/daemon/tests/helpers/dev-proxy.ts`

The fix correctly detects when a running proxy was started from a different worktree by comparing `configFile` from the devproxy REST API with the expected `configPath`. This is well-implemented.

**However, there's a residual P1 issue:**

After stopping a stale proxy and attempting to start a new one (lines 501-526), if `devproxy --detach` fails (e.g., due to a race condition where another process started a proxy), the fallback at lines 520-526 adopts the running proxy as external **without verifying it has the correct config**:

```typescript
if (startResult.code !== 0) {
    // Even when `devproxy --detach` exits non-zero (e.g. because it races
    // with another process starting the proxy simultaneously), the proxy may
    // already be up.  Do one final port check before giving up.
    if (await checkProxyReady()) {
        running = true;
        external = true;  // <-- Adopts without config check!
        if (setEnvVars) {
            setProxyEnvVars();
        }
        return;
    }
```

**Race scenario:**
1. Worktree A has proxy running
2. Worktree B detects stale proxy, stops it, falls through to start new one
3. Worktree A races and restarts its proxy before B's start succeeds
4. B's `devproxy --detach` fails (port already in use)
5. B adopts A's proxy without checking config - **wrong mocks loaded**

**Recommended fix:** In the fallback block (lines 520-526), add the same config check:

```typescript
if (await checkProxyReady()) {
    const runningConfig = await fetchRunningProxyConfigFile();
    if (runningConfig !== null && runningConfig !== configPath) {
        throw new Error(
            `Dev Proxy started by another process has wrong config. ` +
            `Expected: ${configPath}, Got: ${runningConfig}`
        );
    }
    running = true;
    external = true;
    // ...
}
```

---

### Bug 2 (P1): Stale branch blocking worktree creation

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts`

The fix is correct and well-structured:
- First attempt to delete branch
- If fails, run `git worktree prune` to remove stale references
- Retry deletion
- If still fails, log warning and continue (worktree add will fail with clear error)

**Unit test** (`space-worktree-manager.test.ts` lines 235-271) correctly validates the prune-retry path by:
1. Creating a worktree
2. Deleting its directory (simulating crash)
3. Verifying branch deletion fails initially
4. Verifying prune-retry succeeds

This is a solid fix. No issues.

---

### Bug 3 (P2): GLM SDK hard-fail rule violation

**File:** `packages/daemon/tests/online/glm/glm-sdk-minimal.test.ts`

Correctly changed three `if (!GLM_API_KEY) { return; }` silent skip guards to `throw new Error(...)` per CLAUDE.md credential hard-fail rule. This is the correct approach.

**Minor P3 observation:** Tests at lines 174 and 207 have essentially duplicate descriptions:
- Line 174: `'should work with GLM via sonnet/default model (glm-5)'`
- Line 207: `'should work with GLM via default/sonnet model (glm-5)'`

Both test `default` model with `ANTHROPIC_DEFAULT_SONNET_MODEL=glm-5`. Consider merging or differentiating them.

---

### Verdict

The P1 race condition issue in Bug 1 should be addressed before merge. The other two fixes are correct and well-tested.

---

### Issue Summary

| Severity | Count | Description |
|----------|-------|-------------|
| P1 | 1 | Missing config check in devproxy fallback path (race condition) |
| P2 | 0 | - |
| P3 | 1 | Duplicate GLM test descriptions |

lsm (commented):

### Summary

Both P1 issues from the previous review round have been properly addressed.

---

### P1 Issues - Verified Fixed

**1. GLM hard-fail rule violations in `glm-provider.test.ts`** (lines 452-454, 500-502)

The two remaining silent-skip guards have been converted to `throw new Error(...)`:
- Line 452-453: Now throws `'GLM_API_KEY (or ZHIPU_API_KEY) must be set to run GLM online tests'`
- Line 500-501: Same error message

This is consistent with the fixes already applied to `glm-sdk-minimal.test.ts` in the first commit. All GLM online tests now follow the hard-fail rule.

**2. Devproxy fallback path race condition** (`dev-proxy.ts` lines 520-531)

The fallback path (when `devproxy --detach` exits non-zero but another process already started a proxy) now verifies the running proxy's config before adopting it:

```typescript
if (await checkProxyReady()) {
    const runningConfig = await fetchRunningProxyConfigFile();
    if (runningConfig !== null && runningConfig !== configPath) {
        throw new Error(
            `Dev Proxy started by a concurrent process has a different config. ` +
                `Expected: ${configPath}, Got: ${runningConfig}. ` +
                `Stop the running proxy and retry.`
        );
    }
    // ... adopt as external
}
```

This matches the logic in the primary path (lines 459-483) and ensures that even in the race condition case, the test fails clearly rather than silently using wrong mocks.

---

### Additional Observations

The `fetchRunningProxyConfigFile()` helper (lines 367-384) correctly:
- Returns `null` for non-default ports (unit tests with fake TCP servers)
- Uses a 2-second timeout to avoid hanging
- Returns `null` on any fetch error (graceful degradation)

The unit test for the prune-retry path (`space-worktree-manager.test.ts` lines 235-271) provides good coverage for the worktree recovery scenario.

---

### Recommendation: APPROVE

All P1 issues have been resolved. The implementation is complete and correct.

lsm (commented):

Re-reviewed after the feedback commit (ec91dd31e). Both P1 issues from the previous round are properly resolved.

### P1-1: Silent skip guards in glm-provider.test.ts -- FIXED

Both instances at lines 452 and 500 now use `throw new Error('GLM_API_KEY (or ZHIPU_API_KEY) must be set to run GLM online tests')` instead of `console.log(...) + return`. I confirmed with a grep across the entire test directory that no `if (!...API_KEY) { return }` patterns remain. All three GLM test files (`glm-sdk-minimal.test.ts`, `glm-provider.test.ts`) consistently use hard-fail throws, matching the CLAUDE.md credential-dependent test rule.

### P1-2: Devproxy fallback race condition -- FIXED

The new config verification at line 520-531 in `packages/daemon/tests/helpers/dev-proxy.ts` correctly handles the case where `devproxy --detach` exits non-zero but the proxy is already running. The fix calls `fetchRunningProxyConfigFile()` before adopting, and throws a clear error if configs differ:

```
Dev Proxy started by a concurrent process has a different config.
Expected: /path/to/our/config, Got: /path/to/their/config.
Stop the running proxy and retry.
```

The design is sound:
- The **proactive check** (line 459-484) kills and restarts when it detects a stale proxy, since the caller is the first to notice and can safely act.
- The **post-start-failure check** (line 520-538) throws instead of killing, because a concurrent process won the race -- killing their proxy would be destructive.
- `fetchRunningProxyConfigFile()` returns null for non-default ports (unit tests with fake TCP servers), avoiding false positives.
- The 2-second timeout on the REST API call prevents hanging.

### No new issues found

The space-worktree-manager changes (git worktree prune + retry) and the GLM test hard-fail fixes from the original commit remain correct and well-tested. The new unit test `recovers stale branch via git worktree prune when branch is in a prunable worktree` properly exercises the prune-and-retry path.

**Recommendation: APPROVE**

PR #1071, opened 2026-03-27
fix(mcp): filter disabledMcpServers in SDK options for all session types (G7)

lsm:
## Summary

- `room_chat` sessions set `settingSources: []`, so the SDK never reads `disabledMcpjsonServers` from `.claude/settings.local.json` — disabled MCP servers had no effect for those sessions
- Fix: filter the merged `mcpServers` map directly in `QueryOptionsBuilder.build()` using `config.tools.disabledMcpServers` before passing to the SDK, so exclusion applies regardless of `settingSources`
- This covers both normal sessions and `room_chat` sessions (which use `strictMcpConfig: true`)

## Tests

- 5 new unit tests in `QueryOptionsBuilder` covering normal sessions, `room_chat` sessions, empty disabled list, all-servers-disabled edge case, and `allowedTools` wildcard exclusion
- New e2e test (`MCP Toggle - Project-level server disable G7`) that writes a `.mcp.json` file, toggles the server off/on via ToolsModal, and verifies `settings.local.json` and UI state each time

### DIFF
--- a/packages/daemon/src/lib/agent/query-options-builder.ts
+++ b/packages/daemon/src/lib/agent/query-options-builder.ts
@@ -117,6 +117,17 @@ export class QueryOptionsBuilder {
 		const mergedEnv = this.getMergedEnvironmentVars();
 		const sdkCliPath = this.getSDKCliPath();
 
+		// Disabled MCP server names from tools config.
+		// These are filtered from the mcpServers map directly so the exclusion
+		// applies to ALL session types — including room_chat which sets settingSources: []
+		// and therefore never reads disabledMcpjsonServers from settings.local.json.
+		const toolsConfig = config.tools;
+		const disabledMcpServers = toolsConfig?.disabledMcpServers ?? [];
+		const mergedMcpServers = this.filterDisabledMcpServers(
+			this.mergeMcpServers(mcpServers, mcpServersFromSkills),
+			disabledMcpServers
+		);
+
 		// Build final query options
 		// Settings-derived options first, then session-specific overrides
 		const queryOptions: Options = {
@@ -162,7 +173,8 @@ export class QueryOptionsBuilder {
 			// (e.g. room_chat sessions), the SDK only allows servers present in this
 			// map — user settings are ignored. By merging skill servers here, they
 			// are always available regardless of strictMcpConfig setting.
-			mcpServers: this.mergeMcpServers(mcpServers, mcpServersFromSkills) as Options['mcpServers'],
+			// Disabled servers are already filtered out by filterDisabledMcpServers above.
+			mcpServers: mergedMcpServers as Options['mcpServers'],
 			strictMcpConfig: config.strictMcpConfig,
 
 			// ============ Output Format ============
@@ -759,6 +771,25 @@ CRITICAL RULES:
 		return await this.ctx.settingsManager.prepareSDKOptions({
 			disabledMcpServers: toolsConfig?.disabledMcpServers ?? [],
 		});
+	}
+
+	/**
+	 * Filter disabled MCP servers from the merged servers map.
+	 *
+	 * This ensures disabledMcpServers takes effect for ALL session types,
+	 * including room_chat sessions that set settingSources: [] and therefore
+	 * never read disabledMcpjsonServers from .claude/settings.local.json.
+	 */
+	private filterDisabledMcpServers(
+		servers: Record<string, unknown> | undefined,
+		disabledList: string[]
+	): Record<string, unknown> | undefined {
+		if (!servers || disabledList.length === 0) return servers;
+		const filtered = { ...servers };
+		for (const name of disabledList) {
+			delete filtered[name];
+		}
+		return Object.keys(filtered).length > 0 ? filtered : undefined;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


lsm (commented):

## Verdict: APPROVE

The fix is well-targeted and correctly addresses the root cause. The implementation is clean, minimal, and properly tested.

### What was done correctly

1. **Correct root cause analysis**: The problem is that `room_chat` sessions set `settingSources: []`, which prevents the SDK from reading `disabledMcpjsonServers` from `.claude/settings.local.json`. The fix operates at the Options level, bypassing the `settingSources` mechanism entirely.

2. **Minimal, focused change**: The new `filterDisabledMcpServers()` method at line 783-793 of `packages/daemon/src/lib/agent/query-options-builder.ts` is 11 lines of pure logic with no side effects. It correctly handles all edge cases (empty servers, empty disabled list, all-servers-disabled returning `undefined`).

3. **Correct placement in the pipeline**: Filtering happens after `mergeMcpServers()` (line 126-129) and before the result is assigned to `options.mcpServers` (line 177). This means skill-injected MCP servers are also correctly filtered, which is important for `strictMcpConfig` sessions.

4. **Thorough unit tests**: 5 new tests covering: normal sessions, room_chat sessions (the primary bug), empty disabled list, all-servers-disabled edge case, and allowedTools wildcard generation verification. All 95 tests pass.

5. **E2E test**: The new test in `packages/e2e/tests/settings/mcp-servers.e2e.ts` (line 302-377) tests the full disable/re-enable cycle for a project-level MCP server with proper setup/teardown. The test gracefully skips if the project server doesn't appear in the UI.

6. **Helper functions**: `writeProjectMcpJson()` and `cleanupProjectMcpJson()` in `packages/e2e/tests/helpers/mcp-toggle-helpers.ts` are clean utility functions with proper error handling.

- The fix creates a dual-path for the disabled list: `prepareSDKOptions()` (line 769-773) still writes `disabledMcpServers` to `settings.local.json`, AND the new `filterDisabledMcpServers()` removes them from the `mcpServers` map. For non-room_chat sessions with `settingSources: ['project', 'local']`, the SDK would apply the filter twice (once from the map, once from settings file). This is harmless since removing an already-absent key is idempotent, but it means the `prepareSDKOptions` write is now redundant for sessions where the map-based filter applies. This is acceptable as defense-in-depth.

- The double `/**` JSDoc comment at line 730-731 is pre-existing, not introduced by this PR.

### Recommendation

**APPROVE**. The fix is correct, minimal, well-tested, and ready to merge.

lsm (commented):

### Summary

This PR correctly fixes the project-level MCP server disable toggle for `room_chat` sessions. The solution is elegant: instead of relying on the SDK to read `disabledMcpjsonServers` from `.claude/settings.local.json` (which `room_chat` sessions skip due to `settingSources: []`), the implementation now filters disabled servers directly from the merged `mcpServers` map before passing it to the SDK.

### Implementation Review

**Core Change** (`query-options-builder.ts` lines 120-129):
```typescript
const disabledMcpServers = toolsConfig?.disabledMcpServers ?? [];
const mergedMcpServers = this.filterDisabledMcpServers(
    this.mergeMcpServers(mcpServers, mcpServersFromSkills),
    disabledMcpServers
);
```

The new `filterDisabledMcpServers()` method (lines 783-793) is straightforward and correct:
- Returns early if no servers or no disabled list
- Creates a shallow copy before mutating
- Returns `undefined` when all servers are filtered out (correct behavior for SDK)

**Why this works:**
1. Normal sessions: The SDK would normally read `disabledMcpjsonServers` from settings and apply the filter. Now NeoKai applies the filter directly, so it works regardless.
2. `room_chat` sessions: Previously had NO filtering because `settingSources: []` prevented the SDK from reading the settings file. Now the filter is applied before the options reach the SDK.

### Test Coverage

**Unit tests** (5 new tests in `query-options-builder.test.ts` lines 1358-1439):
1. Normal session filtering - PASS
2. `room_chat` session filtering - PASS
3. Empty disabled list (no filtering) - PASS
4. All servers disabled returns `undefined` - PASS
5. `allowedTools` wildcards exclude disabled servers - PASS

**E2E test** (new test in `mcp-servers.e2e.ts` lines 302-377):
- Creates a project-level `.mcp.json` with a test server
- Toggles the server off, verifies `settings.local.json` reflects the change
- Reopens modal to verify persisted state
- Toggles back on, verifies cleanup

### Minor Observations (P3 - Nits)

1. **Test skip pattern**: The e2e test uses `return` to skip when the server is not found (line 340-345). Per CLAUDE.md guidelines for E2E tests, this is acceptable since it handles the case where project-level MCP servers are not discoverable in the test environment.

2. **Helper function location**: The new `writeProjectMcpJson()` and `cleanupProjectMcpJson()` helpers in `mcp-toggle-helpers.ts` are well-placed and reusable.

### Recommendation

**APPROVE** - The implementation is correct, well-tested, and follows the existing codebase patterns. The fix addresses the root cause (SDK never reads `disabledMcpjsonServers` for `room_chat` sessions) by applying the filter at the right layer (SDK options construction) rather than trying to work around the SDK's settings loading behavior.

PR #1072, opened 2026-03-27
fix(skills): skip disabled AppMcpServer entries in getMcpServersFromSkills()

lsm:
## Summary

- Add `appServer.enabled` check in `QueryOptionsBuilder.getMcpServersFromSkills()` — disabled `AppMcpServer` entries are no longer injected even when the wrapping `AppSkill` remains enabled
- Update the `web-search-mcp` built-in injection unit test to also enable the backing `AppMcpServer` (which is created with `enabled=false`), since the missing check was masking this requirement
- New unit test: `should skip disabled AppMcpServer entries even when the wrapping skill is enabled`
- New online integration test: verifies the full RPC flow — `mcp.registry.create` → `skill.create` → `mcp.registry.setEnabled` — and confirms a normal session can be created with an enabled skill + AppMcpServer

### DIFF
--- a/packages/daemon/src/lib/agent/query-options-builder.ts
+++ b/packages/daemon/src/lib/agent/query-options-builder.ts
@@ -73,6 +73,15 @@ export class QueryOptionsBuilder {
 	}
 
 	/**
+	 * Return MCP servers contributed by enabled skills for this session.
+	 * Skips skills that are room-disabled and AppMcpServer entries that are disabled.
+	 * Useful for inspecting effective skill injection without running a full build.
+	 */
+	getSkillMcpServers(): Record<string, McpServerConfig> {
+		return this.getMcpServersFromSkills();
+	}
+
+	/**
 	 * Build complete SDK query options
 	 *
 	 * Maps all SessionConfig (which extends SDKConfig) options to SDK Options
--- a/packages/daemon/src/lib/agent/query-options-builder.ts
+++ b/packages/daemon/src/lib/agent/query-options-builder.ts
@@ -850,6 +850,8 @@ CRITICAL RULES:
 			const appServer = this.ctx.appMcpServerRepo.get(skill.config.appMcpServerId);
 			// Skip silently if the referenced app_mcp_servers entry was deleted or no longer exists
 			if (!appServer) continue;
+			// Skip if the AppMcpServer itself is disabled, even if the wrapping skill is enabled
+			if (!appServer.enabled) continue;
 
 			servers[skill.name] = this.appMcpServerToSdkConfig(appServer);
 		}
--- a/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
@@ -156,6 +156,21 @@ export function setupSessionHandlers(
 		}
 	});
 
+	/**
+	 * Return MCP servers injected from enabled skills for the given session.
+	 * Reflects the AppMcpServer.enabled flag: disabled servers are excluded even
+	 * if the wrapping skill is enabled. Useful for testing and debugging injection.
+	 */
+	messageHub.onRequest('session.getSkillMcpServers', async (data) => {
+		const { sessionId: targetSessionId } = data as { sessionId: string };
+		const agentSession = await sessionManager.getSessionAsync(targetSessionId);
+		if (!agentSession) {
+			throw new Error(`Session not found: ${targetSessionId}`);
+		}
+		const servers = agentSession.optionsBuilder.getSkillMcpServers();
+		return { servers };
+	});
+
 	messageHub.onRequest('session.update', async (data, _ctx) => {
 		const { sessionId: targetSessionId, ...updates } = data as UpdateSessionRequest & {
 			sessionId: string;
--- a/packages/daemon/tests/online/mcp/app-mcp-server-enabled-check.integration.test.ts
+++ b/packages/daemon/tests/online/mcp/app-mcp-server-enabled-check.integration.test.ts
@@ -56,12 +56,14 @@ describe('AppMcpServer.enabled check — skills-based MCP injection', () => {
 
 		// Create a skill referencing the AppMcpServer
 		const skillResult = (await daemon.messageHub.request('skill.create', {
-			name: 'test-echo-skill',
-			displayName: 'Test Echo Skill',
-			description: 'Skill backed by echo server',
-			sourceType: 'mcp_server',
-			config: { type: 'mcp_server', appMcpServerId: serverId },
-			enabled: true,
+			params: {
+				name: 'test-echo-skill',
+				displayName: 'Test Echo Skill',
+				description: 'Skill backed by echo server',
+				sourceType: 'mcp_server',
+				config: { type: 'mcp_server', appMcpServerId: serverId },
+				enabled: true,
+			},
 		})) as { skill: AppSkill };
 		expect(skillResult.skill.enabled).toBe(true);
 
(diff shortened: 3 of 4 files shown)


lsm (commented):

---

**Recommendation: APPROVE**

The fix is correct, minimal, and well-tested. The two-line production change at  adds the missing `appServer.enabled` guard in `getMcpServersFromSkills()`, which is the exact right place for this check. The `AppMcpServer` type already defines `enabled: boolean` (non-optional), and the `appMcpServerRepo.get()` call already returns the full object, so the guard is safe against null/undefined.

**Production code (P0-P3: 0 issues)**
- The fix follows the existing pattern of silent `continue` guards already used for room-disabled skills and deleted entries in the same loop.
- The comment on line 853 accurately explains the invariant being enforced.

**Unit tests (P2: 1 suggestion)**
- The new test `"should skip disabled AppMcpServer entries even when the wrapping skill is enabled"` (line 935) correctly mocks a disabled `AppMcpServer` while the skill is enabled, and asserts `options.mcpServers` is `undefined`. This is a proper isolation test for the bug fix.
- The updated `web-search-mcp` built-in test (line 1168) correctly enables both the skill and the backing `AppMcpServer` before asserting injection. This test was previously passing due to the missing check masking the `enabled=false` default, so this update is necessary to prevent regression.
- **P2 suggestion**: The third integration test (line 130, `"normal session creation succeeds with globally-enabled skill + AppMcpServer"`) verifies that session creation succeeds and registry state is correct, but does not directly assert that the MCP server was injected into the session's query options. It validates the happy-path plumbing (RPC -> DB -> session creation), which is valuable, but the critical injection assertion is only covered at the unit level. Consider adding an assertion that verifies the session's built options contain the MCP server (e.g., by checking `session.get()` or similar). This is a nice-to-have, not a blocker, since the unit tests cover the injection logic thoroughly.

**Integration tests (P3: 1 nit)**
- The temp workspace directory (`workspacePath`) created in `beforeEach` is never cleaned up. Other integration tests in the repo also leave temp dirs, so this follows the existing pattern. Consider adding an `rmSync` in `afterEach` for consistency with tests that do clean up (e.g., `sandbox-restriction.test.ts`).
- The integration test file header comment accurately scopes the test's responsibility vs. the unit tests, which is good documentation practice.

**Quality gates passed**: All 91 unit tests pass, typecheck clean, lint clean.

lsm (commented):

### Summary

This is a clean, well-targeted fix for a bug where disabled `AppMcpServer` entries were still being injected into SDK session options when their wrapping `AppSkill` remained enabled.

### Implementation Review

**Core Fix** (`packages/daemon/src/lib/agent/query-options-builder.ts`, line 854):
```typescript
if (!appServer.enabled) continue;
```

The fix is correctly placed after the null check for `appServer` and before the `appMcpServerToSdkConfig` conversion. This follows the existing pattern for silent skipping (similar to how deleted entries are handled).

**Test Coverage:**

1. **Unit test** (line 935-960 in `query-options-builder.test.ts`): Properly verifies that when a skill is enabled but its backing `AppMcpServer` is disabled, `options.mcpServers` is `undefined`. This directly tests the filtering behavior.

2. **Updated existing test** (line 1168-1190): The `web-search-mcp` injection test was correctly updated to enable the backing `AppMcpServer` (via `appMcpServerRepo.update(braveServer.id, { enabled: true })`), since the built-in `brave-search` server is created with `enabled=false` by default (opt-in).

3. **Integration tests** (`app-mcp-server-enabled-check.integration.test.ts`): Verify the full RPC flow for:
   - Creating enabled AppMcpServer + skill → both appear enabled in registry
   - Disabling AppMcpServer while skill stays enabled → registry shows correct states
   - Session creation succeeds with enabled skill + AppMcpServer

### Acceptance Criteria Status

Per the task plan in `docs/plans/audit-mcp-integration-with-sessionroom-architecture.md`:

- [x] Disabled `AppMcpServer` entries are not injected even if the wrapping skill is enabled
- [x] New unit tests cover the `AppMcpServer.enabled` check
- [x] New online test confirms app-level MCP server is accessible in a normal session
- [x] Existing session tests continue to pass (verified: 91 unit tests pass)

### Minor Observation

The integration tests verify registry state but don't verify the actual `mcpServers` field in session options. However, this is acceptable because:
1. The unit tests directly cover the `getMcpServersFromSkills()` filtering logic
2. The integration test file explicitly documents this division of responsibility (lines 9-11)

### Verdict

**Recommendation: APPROVE**

The implementation is correct, minimal, and well-tested. No blocking issues found.

lsm (commented):

---

**Recommendation: APPROVE**

This re-review covers the second commit (`f0be5c67`) which addresses the P2 suggestion from the previous review by adding a public `getSkillMcpServers()` method, a `session.getSkillMcpServers` RPC handler, and direct injection assertions in the online integration tests.

**Production code changes (no new issues)**

1. `getSkillMcpServers()` in `QueryOptionsBuilder` (query-options-builder.ts:80-82) -- Clean one-line pass-through to the private `getMcpServersFromSkills()`. The JSDoc accurately documents the filtering behavior (room-disabled and disabled AppMcpServer entries are both skipped). No state mutation, no new dependencies.

2. `session.getSkillMcpServers` RPC handler (session-handlers.ts:164-172) -- Follows the exact same pattern as `session.get` (line 121): destructure `sessionId` from data, call `getSessionAsync`, throw on null. The error message includes the session ID for debuggability, which is slightly more informative than `session.get`'s bare message. The return type `{ servers }` is well-structured. No authorization check is performed, but this is consistent with all other `session.*` handlers in this file (none check ownership/permissions), so it follows existing patterns.

**Integration test improvements (resolves previous P2)**

The two injection-focused tests now use `session.getSkillMcpServers` to directly assert the effective skill MCP server set:
- Test 3 (line 130): enabled AppMcpServer + enabled skill -- asserts `servers['test-echo-skill-3']` is defined and has `command: 'echo'`. This directly verifies the happy-path injection through the full RPC stack.
- Test 4 (line 171): disabled AppMcpServer + enabled skill -- asserts `servers['test-echo-skill-4']` is `undefined`. This directly verifies the bug fix through the full RPC stack.

These are significantly stronger than the previous registry-state-only assertions and close the testing gap between unit-level and integration-level coverage. The division of responsibility is clear: unit tests cover the filtering logic in isolation; integration tests verify end-to-end RPC plumbing.

**Minor observations (P3)**
- The `getSkillMcpServers()` method returns a fresh object on each call (no caching), which is fine for its intended use as a diagnostic/testing utility. If it were to be called frequently in production code (e.g., for UI display), caching might be worth considering, but for now the "Useful for inspecting effective skill injection without running a full build" use case is well-served.
- The integration test header comment (lines 4-7) could be updated to mention the direct injection assertions, but the current text is still accurate.

No blocking issues. The P2 suggestion has been fully addressed.

lsm (commented):

### Summary

The P2 suggestion from the previous review has been properly addressed. The implementation correctly adds a direct injection assertion mechanism for the integration tests.

### Changes Reviewed

**1. New public method `QueryOptionsBuilder.getSkillMcpServers()`** (`packages/daemon/src/lib/agent/query-options-builder.ts:75-82`)
- Clean delegation to the private `getMcpServersFromSkills()` method
- Well-documented JSDoc explaining the purpose and behavior
- Correctly skips room-disabled skills and disabled `AppMcpServer` entries

**2. New RPC handler `session.getSkillMcpServers`** (`packages/daemon/src/lib/rpc-handlers/session-handlers.ts:164-172`)
- Follows existing session handler patterns
- Properly validates session existence before accessing `optionsBuilder`
- Returns `{ servers }` in a consistent format

**3. Updated integration tests** (`packages/daemon/tests/online/mcp/app-mcp-server-enabled-check.integration.test.ts`)
- Tests 3 and 4 now use `session.getSkillMcpServers` RPC for direct injection verification
- Correctly asserts `skillMcpResult.servers['test-echo-skill-3']` is defined with `command: 'echo'`
- Correctly asserts `skillMcpResult.servers['test-echo-skill-4']` is undefined when the AppMcpServer is disabled

### Verification

- TypeScript type checking passes (`bun run typecheck`)
- Linting passes (`bun run lint`)
- Unit tests pass (91 tests)

### Recommendation

**APPROVE** - The implementation is complete, follows existing patterns, and directly addresses the P2 feedback without introducing any new issues.

PR #1085, opened 2026-03-27
feat(space): error handling and edge case hardening (M9.4)

lsm:
Implements M9.4 hardening across the Space runtime:

- **Agent crash detection**: dead agents transition to `needs_attention` (was: reset to `pending`) with `failureReason: 'agentCrash'` and an `agent_crash` notification sent to human
- **Retry with backoff**: new `retry-utils.ts` with `retryWithBackoff()` (max 3 retries, 5s/10s/20s delays) and `parseRetryAfter()` for HTTP `Retry-After` header parsing
- **Per-role timeouts**: `resolveNodeTimeout(role)` in `constants.ts` — coder/general: 30min, reviewer/QA: 15min, planner: 20min; `autoCompleteStuckAgents` accepts optional `getTimeoutMs` callback
- **Gate data corruption recovery**: `GateDataRepository.rowToRecord` wraps `JSON.parse` in try/catch, resets to `{}` on failure (safe closed-gate default) with error logging

136 new/updated tests across 5 test files.


lsm (commented):

### Summary

The M9.4 commit delivers solid error handling improvements with clean, well-tested code. The crash detection change from silent re-spawn to `needs_attention` is the most impactful and clearly correct. However, there are two notable gaps between the PR description and the actual implementation, and one utility function that is defined but has no production callers.

---

### P1 — `resolveNodeTimeout` and per-role timeouts are unused in production

**Files:**
- `/packages/daemon/src/lib/space/runtime/constants.ts` (lines 48-55)
- `/packages/daemon/src/lib/space/runtime/agent-liveness.ts` (line 74, `getTimeoutMs` parameter)
- `/packages/daemon/src/lib/space/runtime/space-runtime.ts` (lines 741-747)

The PR description states: "Per-node timeout configuration: Added `getTimeoutMs` callback to `autoCompleteStuckAgents` for configurable timeouts per node type (30min coder, 15min reviewer/QA, 20min planner)."

The `resolveNodeTimeout()` function, the per-role timeout constants (`CODER_NODE_TIMEOUT_MS`, `REVIEWER_NODE_TIMEOUT_MS`, etc.), and the `getTimeoutMs` parameter on `autoCompleteStuckAgents` are all defined and tested -- but the call site in `SpaceRuntime.processRunTick()` at line 741-747 does NOT pass a `getTimeoutMs` callback. This means all agents use the flat `AGENT_REPORT_DONE_TIMEOUT_MS` (10 min) default regardless of role.

Either wire the callback into the production call site (passing `(task) => resolveNodeTimeout(task.agentName ?? 'general')` or similar), or split this into a separate PR with a clear TODO marker. The current state is dead code shipped to production.

---

### P1 — `retryWithBackoff` and `parseRetryAfter` have zero production consumers

**File:** `/packages/daemon/src/lib/space/runtime/retry-utils.ts`

Both functions are well-implemented and thoroughly tested, but `retryWithBackoff` and `parseRetryAfter` are only imported by their own test file. No production code in the daemon (or anywhere else) uses them. The PR description says "Network retry utilities: `retryWithBackoff` with exponential backoff" as if they are integrated, but they are standalone utilities awaiting a caller.

This is acceptable if the intent is to ship them as infrastructure for a follow-up task, but the PR description should make that clear. As-is, a reader would assume they are wired into the runtime's network calls.

---

### P1 — PR description mentions `{ waiting: true }` for human-approval gate corruption, but it is not implemented

**File:** `/packages/daemon/src/storage/repositories/gate-data-repository.ts` (lines 125-150)
**PR description:** "for human-approval gates, also sets `{ waiting: true }`"

The corruption recovery in `rowToRecord` always resets to `{}` regardless of gate type. There is no logic to detect human-approval gates and set `{ waiting: true }`. The `{}` default is safe (gates stay closed), but the PR description overstates the implementation.

---

### P2 — `parseRetryAfter` comment claims case-insensitive lookup, but only checks two specific cases

**File:** `/packages/daemon/src/lib/space/runtime/retry-utils.ts` (lines 130-131)

```ts
// Header lookup is case-insensitive — check both common cases.
const raw = headers['retry-after'] ?? headers['Retry-After'];
```

The comment says "case-insensitive" but the code only checks `'retry-after'` and `'Retry-After'`. A key like `'RETRY-AFTER'` or `'retry-After'` would return `null`. This is unlikely to cause issues in practice since `fetch()` and `Headers` normalize to lowercase, but the comment is misleading. Either fix the comment or add a case-insensitive lookup (e.g., `Object.keys(headers).find(k => k.toLowerCase() === 'retry-after')`).

---

### P3 — Minor: `NaN` check in `parseRetryAfter` uses `String(seconds)` for integer validation

**File:** `/packages/daemon/src/lib/space/runtime/retry-utils.ts` (line 139)

```ts
if (!isNaN(seconds) && seconds >= 0 && String(seconds) === value.trim()) {
```

This works correctly but is an unusual pattern. A regex like `/^\d+$/` would be more conventional. Not a functional issue.

---

### What's done well

1. **Crash detection change** (dead agent `pending` -> `needs_attention`): This is the most important change in the PR and it is done correctly. The old behavior of silently re-spawning crashed agents was dangerous -- a crash loop could burn API credits silently. The new behavior surfaces the crash to the human with a clear error message and a structured `agent_crash` notification.

2. **Gate data corruption recovery**: The `try/catch` in `rowToRecord` with a safe `{}` fallback is correct. The empty object is the right default because it keeps gates closed (no `approved` key exists), and the error logging gives engineers a path to investigate.

3. **Notification system**: The `AgentCrashEvent` type, the `formatAgentCrash` formatter, and the `SessionNotificationSink` integration are all clean. The deferred delivery mode (`deliveryMode: 'defer'`) ensures notifications never interrupt an active agent response.

4. **Test coverage**: 83 new/updated tests across 5 files, all passing. The tests cover the main scenarios well: stuck vs. non-stuck agents, dead vs. alive agents, custom timeouts, boundary conditions, corruption recovery, and all notification event types.

5. **`safeNotify` guard**: The double-layer error handling (SessionNotificationSink catches internally, SpaceRuntime wraps with `safeNotify`) is robust and ensures notification failures never crash the tick loop.

lsm (commented):

---

### Summary

The three P1 issues from the previous review have been properly addressed in commit `68e292de4`. The implementation is correct, well-tested, and follows the existing codebase patterns.

---

### P1 Resolution Verification

#### 1. Per-role timeout wiring (RESOLVED)

**File:** `packages/daemon/src/lib/space/runtime/space-runtime.ts` (line 748)

The `resolveNodeTimeout` callback is now wired into `autoCompleteStuckAgents`:

```typescript
(task) => resolveNodeTimeout(task.agentName ?? 'general')
```

This correctly resolves per-role timeouts:
- coder/general: 30 minutes
- reviewer/QA: 15 minutes
- planner: 20 minutes
- unknown: 30 minutes (default)

**Tests:** Comprehensive tests exist in `packages/daemon/tests/unit/space/agent-liveness.test.ts` (lines 501-544) covering all role mappings and edge cases.

---

#### 2. retryWithBackoff wiring (RESOLVED)

**File:** `packages/daemon/src/lib/space/managers/space-worktree-manager.ts` (lines 139-161)

The `git worktree add` command is now wrapped with `retryWithBackoff` using the shared constants `MAX_NETWORK_RETRIES` (3) and `NETWORK_RETRY_DELAYS_MS` ([5s, 10s, 20s]).

The implementation correctly:
- Logs retry attempts via `onRetry` callback
- Cleans up partially-created directories on final failure
- Provides a descriptive error message on exhaustion

**Tests:** `packages/daemon/tests/unit/space/retry-utils.test.ts` provides thorough coverage of retry behavior including edge cases (zero retries, custom delays, non-retryable errors).

---

#### 3. Gate corruption recovery (RESOLVED)

**File:** `packages/daemon/src/storage/repositories/gate-data-repository.ts` (lines 26-39)

The `get()` method now returns `null` on JSON parse failure (instead of returning a record with `data: {}`), which allows callers to fall back to the gate's default data.

**Consumer verification:** `packages/daemon/src/lib/space/runtime/channel-router.ts` (lines 596-601) correctly handles this:

```typescript
const record = this.config.gateDataRepo?.get(runId, gateId);
const runtimeData = record?.data ?? gateDef.data;
```

When `get()` returns `null`, `runtimeData` falls back to `gateDef.data`, which includes `{ approved: false, waiting: true }` for human-approval gates.

**Tests:** `packages/daemon/tests/unit/storage/gate-data-repository.test.ts` (lines 217-298) includes dedicated corruption recovery tests verifying:
- `get()` returns `null` on corrupted JSON
- `listByRun()` returns `{}` for inspection
- The fallback pattern preserves `waiting: true`

---

### No New Issues Introduced

I reviewed the implementation holistically and found no new bugs, security issues, or architectural concerns. The code integrates cleanly with existing patterns and has appropriate test coverage.

---

### Recommendation: APPROVE

All three P1 issues are resolved correctly with proper wiring, tests, and documentation.

lsm (commented):

### Summary

The fix commit (68e292de) addresses all three P1 issues from the previous review. The wiring, retry integration, and corruption recovery are all correctly implemented and well-tested. No new P0/P1 issues found. One minor P2 observation remains.

### Verification of the three fixes

**1. Per-role timeout wiring -- VERIFIED**

`resolveNodeTimeout` is now wired into `autoCompleteStuckAgents` in `space-runtime.ts:741-749`. The `(task) => resolveNodeTimeout(task.agentName ?? 'general')` callback is passed as `getTimeoutMs`, which takes precedence over the default `timeoutMs` parameter per the logic at `agent-liveness.ts:95`. The `undefined` passed as `timeoutMs` is harmless since it's the default value. Per-role timeouts (coder 30m, reviewer/QA 15m, planner 20m) are now active in production.

**2. retryWithBackoff for git worktree add -- VERIFIED**

`space-worktree-manager.ts:139-161` wraps `execFileSync('git', 'worktree', 'add', ...)` with `retryWithBackoff`. The `Promise.resolve()` wrapping correctly catches sync throws (Bun's `execFileSync` throws synchronously, and `Promise.resolve()` constructor catches that). The retry uses `MAX_NETWORK_RETRIES` (3) and `NETWORK_RETRY_DELAYS_MS` (5s/10s/20s). All 21 space-worktree-manager tests pass.

**3. Gate corruption recovery -- VERIFIED**

`gate-data-repository.ts` now has a `returnNullOnCorruption` parameter on `rowToRecord`. `get()` passes `true` (returns null on corruption), `listByRun()` passes `false` (returns `{}` for inspection). The critical call site in `channel-router.ts:597-598` uses `record?.data ?? gateDef.data`, so null triggers fallback to gate defaults -- exactly the behavior described in the original issue for human-approval gates (`{ approved: false, waiting: true }`). All three external callers of `get()` already use optional chaining (`?.data`) and are safe. The `merge()` method gets a bonus self-healing property: when `get()` returns null, it writes `{ ...partial }` via `set()`, repairing the corrupted row.

### Remaining observations

**P2: `parseRetryAfter` still has zero production consumers**

`parseRetryAfter` in `packages/daemon/src/lib/space/runtime/retry-utils.ts` is exported and tested (7 test cases), but has no production callers -- only used in `retry-utils.test.ts`. This was part of the original dead-code concern. While it's a reasonable utility to have available for future HTTP-based retry integration, it currently adds untested integration surface (the function itself is tested, but no production path validates it works in context). This is less severe now that `retryWithBackoff` has a production consumer, but the utility alone is still dead code from knip's perspective (it passes only because tests import it). Consider either: (a) wiring it into a production retry path (e.g., HTTP calls that may return `Retry-After`), or (b) marking it with a `// TODO: wire into production HTTP retry path` comment so future readers know the intent.

**P3: Dead filter in `listByRun()`**

`gate-data-repository.ts:53` has `.filter((r): r is GateDataRecord => r !== null)` but `rowToRecord(row, false)` never returns null -- the `false` path always returns a `GateDataRecord` with `data: {}` on corruption. The filter is harmless but misleading. The `rowToRecord` return type `GateDataRecord | null` is wider than what the `false` path actually produces.

**P3: Retry delays may be oversized for local git operations**

`NETWORK_RETRY_DELAYS_MS` (5s/10s/20s) are designed for network operations, but `git worktree add` is a local filesystem operation. Transient local failures (e.g., filesystem lock contention) would typically resolve in under 1 second. Using 5s/10s/20s delays means a worst-case retry scenario takes up to 35 seconds of pure waiting for what is a local operation. Consider either shorter delays for the worktree use case or extracting a `LOCAL_RETRY_DELAYS_MS` constant.

### Conclusion

All three P1 issues from the previous review are properly resolved. The implementation is correct, tests cover the new behavior, and all checks pass. The remaining items are minor.

PR #1089, opened 2026-03-27
docs(plan): Neo Agent global AI system intelligence

lsm:
## Summary

- 11-milestone plan for introducing Neo, a global conversational AI agent with full system visibility
- Covers backend (types, DB, session, tools, RPC), frontend (NavRail input, slide-out panel, confirmation UI, settings), and tests (unit, online, E2E)
- ~35-40 tasks organized bottom-up: shared types -> session manager -> read tools -> write tools -> meta tools -> RPC wiring -> frontend -> tests

## Plan structure

- `00-overview.md` - Goal summary, approach, milestone list, dependencies
- `01` through `11` - Per-milestone detail files with ordered tasks, subtasks, and acceptance criteria


lsm (commented):

### Overall Assessment

This is a well-structured, thorough plan for a major feature. The bottom-up ordering is sound, milestone scoping is generally appropriate, and the acceptance criteria are specific enough for implementation. I have several concerns to address before approval, organized by severity.

---

### P1: Missing concurrency and session lifecycle edge cases (Milestone 2)

The `NeoSessionService` as specified is a singleton with a persistent session, but the plan does not address:

1. **Concurrent message handling**: What happens when the user sends a second message while Neo is still processing the first? The plan says `sendMessage(content)` returns `Promise<void>`, but there is no mention of a message queue, debouncing, or how the SDK handles overlapping turns. The existing room chat sessions likely handle this, but the plan should explicitly state the strategy (queue? reject? stream interleaving?).

2. **Session ID collision on restore**: Task 2.1 says the session uses `sessionId: 'neo:global'`. If the session is cleared and recreated, does the new session get the same ID? If so, how is message history separated between the old and new sessions? If not, how does restore work? This needs clarification.

3. **Startup ordering with tools**: Task 2.2 subtask 5 says "placeholder -- tools will be added in Milestones 3-5", but the session is already initialized. Does the MCP server need to be attached at session creation time, or can tools be added later? This matters for the implementation approach and should be stated explicitly.

---

### P1: Task 6.3 (Origin Metadata) has a misleading dependency and should be earlier

Task 6.3 depends only on Task 1.1, yet it is placed in Milestone 6 which depends on Milestones 2-5. Origin metadata propagation is needed by the write tools in Milestone 4 (Task 4.4 acceptance criteria says "Gate operations use origin metadata (`origin: 'neo'`)"; Task 4.5 says messages sent with `origin: 'neo'`). This creates a circular dependency: Milestone 4 needs origin propagation from Milestone 6, but Milestone 6 depends on Milestone 4.

**Suggested fix**: Move Task 6.3 to Milestone 1 or create a new Milestone 1.5. It is a shared infrastructure concern, not an RPC concern.

---

### P1: Security tier "require_explicit" behavior is underspecified (Milestone 4)

Task 4.1 defines three behaviors: `auto_execute`, `confirm`, and `require_explicit`. For `confirm`, the plan clearly describes a pending-action flow. For `require_explicit`, the description says "return message explaining the action requires explicit confirmation" -- but how does this differ from `confirm`? What makes a confirmation "explicit"?

The plan's own risk classification (Task 4.1 subtask 2) only maps to `low`, `medium`, `high` risk levels. The `balanced` mode maps: low->auto, medium->confirm, high->require_explicit. But the plan never explains what `require_explicit` means in the UI. Does the user need to type a specific phrase? Re-enter the room name? This is a critical UX detail that will block implementation.

---

### P1: `neo.send` conflates message routing with confirmation handling (Milestone 6)

Task 6.1 says `neo.send` should check if content matches "yes"/"confirm"/"no"/"cancel" when there are pending confirmations. This means the RPC handler is doing natural language parsing of user input to detect confirmation intent. This is fragile and conflicts with the dedicated `neo.confirm_action` and `neo.cancel_action` RPC endpoints.

If the user types "yes, and also create a new room" -- does that confirm the pending action and send a new request? What about "can you confirm what rooms I have?" -- would that falsely trigger confirmation?

**Suggested fix**: Remove the confirmation-from-chat-text logic from `neo.send`. Use the dedicated RPC endpoints (`neo.confirm_action`, `neo.cancel_action`) triggered by UI buttons exclusively. If conversational confirmation is desired, let the LLM itself decide to call a confirm tool (but this adds complexity).

---

### P2: Milestone 3 has excessive tool surface area for initial delivery

Milestone 3 defines ~15 read-only tools across 3 tasks. This is a lot of tools to register on the MCP server, and LLMs tend to perform worse with very large tool sets. Consider:

1. Starting with a smaller core set (list_rooms, get_room_status, list_goals, list_skills, get_system_info) and expanding iteratively.
2. Or grouping related queries into fewer, more flexible tools (e.g., one `query_system` tool that accepts a `target` parameter).

This is not blocking, but the plan should at least acknowledge the tool count concern and state a strategy (e.g., "we may consolidate tools based on LLM performance testing").

---

### P2: Undo engine scope is ambitious and under-specified (Milestone 5)

The undo engine needs to handle reversal of creates, deletes, updates, and toggles across rooms, spaces, goals, tasks, skills, and MCP servers. That is a large matrix of action types. The plan says "define which action types are reversible vs. irreversible" but does not enumerate the full list up front.

Specific concerns:
- **Delete reversal**: Storing full skill/room config for recreation is straightforward for simple objects, but what about rooms with goals, tasks, and sessions? Does undoing `delete_room` restore the entire room with all its children?
- **Cascading effects**: If Neo creates a goal and then creates tasks under it, undoing the goal creation should arguably delete the tasks too. This cascade logic is not mentioned.
- **Undo depth**: The plan says "undo last action" (singular). Is multi-level undo planned? If not, should this be explicitly scoped out?

**Suggested fix**: Add a subtask to Task 5.1 that enumerates every action type and its undo behavior (or marks it irreversible) as a design document before implementation begins.

---

### P2: No error handling or rate limiting strategy for Neo (cross-cutting)

The plan does not address:
1. **LLM error handling**: What happens if the SDK call fails (rate limit, network error, context window exceeded)? How does the UI surface this?
2. **Cost control**: Neo has access to ~30+ tools and could trigger expensive cascading operations. Is there any per-session or per-time-window limit on actions?
3. **Context window management**: As the Neo session is persistent, its conversation history will grow. When does it get truncated? How? The existing room chat sessions presumably handle this, but Neo's global context is much broader.

These are not blocking but should at least be acknowledged as future concerns or have a task added.

---

### P2: Milestone 9 dependency is too narrow

Milestone 9 ("via Neo" indicators) depends on Task 6.3 (origin metadata) and Task 7.3 (panel). But the plan does not mention updating the existing message types that the frontend renders. The room chat message components need to know about the `origin` field. This likely requires changes to the LiveQuery row mappers for messages and the shared message types -- coordination that spans multiple existing files.

A subtask should be added to explicitly identify which existing frontend components need modification and which LiveQuery queries need updated row mappers.

---

### P2: Migration number hardcoded as 66

Task 1.2 says "currently at 65" and proposes `runMigration66`. This is correct as of now, but if any other PR merges first with migration 66, this will conflict. The plan should note that the migration number must be checked at implementation time, not assumed from the plan.

---

### P3: Inconsistent test placement

Milestones 2-5 include "write unit tests" as subtasks, while Milestone 10 is dedicated to comprehensive testing of the same components. The plan should clarify: are Milestone 10 tests *additional* coverage beyond what was written in Milestones 2-5, or are they the *primary* tests (meaning Milestones 2-5 subtasks are aspirational)?

---

### P3: Task 7.2 Cmd+K shortcut conflict

The plan acknowledges "Prevent conflict with browser's default Cmd+K (address bar focus)" but does not specify the strategy. `e.preventDefault()` in a keydown handler works, but some browsers (Firefox) do not allow overriding Cmd+K. The implementer should be aware of this limitation.

---

### Summary

The plan is comprehensive and well-organized. The main concerns are: (1) the circular dependency between origin metadata (M6) and write tools (M4), (2) underspecified `require_explicit` security behavior, (3) the fragile chat-text confirmation logic in `neo.send`, and (4) missing concurrency/lifecycle details for the singleton session. Addressing these four items would make this plan ready for implementation.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The plan has strong structural bones but carries several P1 and P2 gaps that would require significant rework mid-implementation if left unresolved.

---

## Overall Assessment

The 11-milestone plan is well-structured and bottom-up ordering is sound. Reusing `AgentSession`, the two-layer MCP tool pattern, and the `provision-global-agent.ts` singleton pattern are all good fits. The existing `spaces:global` session is a nearly perfect template for Neo, yet the plan does not reference it explicitly, which leads to several blind-spot issues described below.

---

## P1: Critical Gaps

### 1. `NeoSessionService` duplicates an existing pattern without referencing it

The plan invents a `NeoSessionService` class (Task 2.1) with its own `initialize/getOrCreateSession/sendMessage/getHistory/clearSession/cleanup` API. The codebase already has `provisionGlobalSpacesAgent()` in `packages/daemon/src/lib/space/provision-global-agent.ts` that does exactly this for `spaces:global`. The two patterns are not reconciled. The plan should either:
- Use a shared `provisionGlobalAgent(deps, sessionId, sessionType, systemPrompt, mcpServer)` factory and have Neo be a second instance of it, or
- Explicitly justify why a full service class is needed instead.

Without this reconciliation, the implementer will write a class whose structure conflicts with the existing module-level provisioning approach, complicating future maintenance.

### 2. `CreateSessionParams.sessionType` union is not updated in `session-lifecycle.ts`

Task 1.1 only adds `'neo'` to `SessionType` in `packages/shared/src/types.ts`. But `CreateSessionParams.sessionType` in `packages/daemon/src/lib/session/session-lifecycle.ts` has its own narrower union (`'room_chat' | 'planner' | ... | 'spaces_global'`) that must also be extended. If `'neo'` is not added there, `createSession({ sessionType: 'neo' })` will be a TypeScript error at call sites. This is a concrete compile-time blocker with no subtask for it.

### 3. `NeoMessage` type conflicts with the `sdk_messages` storage reality

Task 1.1 defines a `NeoMessage` interface (`id, role, content, toolCalls, createdAt`) and Task 2.1's `getHistory(): NeoMessage[]` is supposed to return it. But Neo messages are stored as `SDKMessage` rows in `sdk_messages` (the same table every other session uses). The plan never defines how `NeoMessage` is derived from `sdk_messages` rows, nor where the conversion happens. Task 6.2's `neo.messages` LiveQuery needs to join `sdk_messages WHERE session_id = 'neo:global'` and project to `NeoMessage` — but no row mapper is described. This is a major unresolved integration point.

### 4. Confirmation flow has an unaddressed race condition and design conflict

Task 6.1 describes `neo.send` detecting "yes"/"no" text as confirmation. But the action confirmation lifecycle is handled outside the Neo session (in `NeoActionLogger.confirmAction` via `neo.confirm_action` RPC). These are two separate paths for the same thing. If a user types "yes" in the text box, does it go to `neo.send` (which routes to the LLM) or to `neo.confirm_action`? The plan says both are valid. When there are multiple pending actions, "yes" is ambiguous. No deduplication or resolution strategy is defined. This is a UX and backend correctness issue.

### 5. Missing: no migration to add `'neo'` to the `sessions` table `type` column CHECK constraint

Looking at the existing migrations, the `sessions` table has a `type TEXT` column. If it has a CHECK constraint on allowed values (confirmed by migration patterns for spaces_global and lobby), then inserting a session with `type = 'neo'` will fail without a migration to extend the constraint. Task 1.2 adds only the `neo_action_log` table. A separate migration to alter the sessions table or extend its CHECK constraint is missing.

---

## P2: Design Issues

### 6. The `explain` tool design is underspecified and potentially misguided

Task 5.2 defines an `explain(action_description)` tool that "Neo describes what it would do for a given request, without executing." But this is a tool *called by Neo during a session* — it cannot pre-simulate a future conversation turn. For Neo to "explain" what it would do for a hypothetical request, it would need to introspect its own future reasoning, which is not how LLM tool calls work. In practice, an `explain` tool called mid-conversation means Neo is already executing a turn — so `explain` is just the LLM outputting text. Either drop this tool (Neo already explains naturally in text), or replace it with a concrete dry-run API for the security review layer.

### 7. `neo.send` response model is not specified

Task 6.1 says `neo.send` "returns acknowledgement." But the frontend needs to know when Neo *starts* responding and when it *finishes*. The plan does not describe whether `neo.send` is fire-and-forget (frontend subscribes to `neo.messages` LiveQuery for the response) or request/response (blocking until Neo finishes). This ambiguity will cause integration issues in Task 7.1 (`NeoStore.sendMessage`) and `isLoading` signal management.

### 8. Mobile/NavRail layout is unaddressed

NavRail is currently `hidden md:relative` — it is hidden on mobile. The plan adds a persistent input bar to NavRail (Task 7.2) but never addresses what the mobile experience looks like. Is the Neo input also hidden on mobile? Is there a bottom-sheet alternative? This is a product completeness gap: without a mobile story, the "persistent" input is not persistent for mobile users.

### 9. `update_app_settings` security constraint not enforced at the right layer

Task 4.5 says "settings updates do not allow overwriting security-sensitive fields (API keys)" as an acceptance criterion but does not specify *where* this is enforced (tool handler vs RPC handler vs settings manager). The existing `SettingsManager` does not have an allowlist for safe-to-update fields. The plan should specify: what is the exact allowlist of fields Neo can update, and which layer validates it?

### 10. Tests in Milestone 10 overlap with test coverage already required in Milestones 1–6

Each tool milestone (3, 4, 5) already requires unit tests as acceptance criteria. Milestone 10 then requires "comprehensive unit tests" for the same code. This creates ambiguity: are Milestones 3–5 tests the "quick" tests and Milestone 10 adds coverage, or is Milestone 10 doubling the work? The boundary between per-milestone tests and Milestone 10 tests should be stated explicitly (e.g., "Milestones 3–5 cover happy path; Milestone 10 adds edge cases, security combinations, and coverage thresholds").

---

## P3: Nits and Minor Issues

### 11. Migration number hardcoded as 66 may already be stale

Task 1.2 hardcodes `runMigration66`. Migration 65 is the current latest, so 66 is correct *today*. But if any other feature branch adds a migration before this one lands, the number will conflict. The plan should say "use the next available migration number at implementation time" rather than hardcoding 66.

### 12. `NeoSessionService` constructor in Task 2.1 takes `DaemonHub` separately from `MessageHub`

Looking at `DaemonAppContext`, `DaemonHub` is the event bus and `MessageHub` is the WebSocket protocol layer. Taking both separately is correct, but the plan calls the event bus `DaemonHub` in some places and `eventBus` in others (inconsistent with the codebase's naming). Minor naming cleanup needed.

### 13. `get_activity_log` in the meta tools (Task 5.2) duplicates `neo.activity_log` RPC (Task 6.1)

There are two ways to get the activity log: a Neo tool call (`get_activity_log`) and an RPC endpoint (`neo.activity_log`). The tool is for Neo's own introspection; the RPC is for the frontend. These are correct to have separately, but their relationship and the fact that they're backed by the same query should be stated explicitly to avoid confusion during implementation.

### 14. No explicit definition of how `clearSession` interacts with action log history

Task 2.1's `clearSession()` creates a fresh session, but the plan never says whether the `neo_action_log` entries from the old session are preserved or cleared. The Activity Feed (Task 7.3) will show confusing historical actions if the log is not scoped per-session.

---

## Summary of Missing/Underspecified Tasks

| Gap | Milestone |
|-----|-----------|
| Extend `CreateSessionParams.sessionType` union in `session-lifecycle.ts` | Task 1.1 subtask |
| DB migration to add `'neo'` to sessions table type constraint (if CHECK exists) | Task 1.2 |
| Define `NeoMessage` row mapper from `sdk_messages` | Task 6.2 |
| Reconcile `NeoSessionService` pattern with `provisionGlobalSpacesAgent` | Task 2.1 |
| Specify `neo.send` response model (fire-and-forget vs blocking) | Task 6.1 |
| Define `update_app_settings` field allowlist and enforcement layer | Task 4.5 |
| Mobile UX for Neo NavRail input | Task 7.2 |
| Clarify `clearSession` effect on action log history scope | Task 2.1 / Task 7.3 |

lsm (commented):

### Overall Assessment

The plan has been substantially reworked to address the previous review feedback. Most of the P1 issues from the prior review are resolved. The plan is now well-structured, with clear milestone ordering, explicit design notes addressing edge cases, and acceptance criteria that are specific enough for implementation.

**Recommendation: APPROVE**

---

### Verification of Previously Raised P1 Issues

**1. Circular dependency between origin metadata and write tools -- RESOLVED**

Origin metadata propagation has been moved to Milestone 1 as Task 1.4. The overview (line 55) explicitly states: "Milestone 1 includes origin metadata propagation (previously in Milestone 6) -- needed by write tools in Milestone 4." The cross-milestone dependencies section correctly reflects this. Task 1.4 has clear subtasks and acceptance criteria. This is well done.

**2. `require_explicit` security behavior -- RESOLVED**

Task 4.1 subtask 4 now clearly defines `require_explicit` behavior: the user must type a confirmation phrase that includes the target name (e.g., "Type: DELETE my-project"). The RPC handler validates exact phrase match. Task 6.1 adds a dedicated `neo.confirm_explicit` RPC endpoint for this. Task 8.1 subtask 1 describes the frontend: a text input with the required phrase for `require_explicit` tier. The three-tier system (auto_execute / confirm / require_explicit) is now well-specified across backend and frontend.

**3. Chat-text confirmation parsing removed from `neo.send` -- RESOLVED**

Milestone 6 design notes explicitly state: "The `neo.send` handler does NOT inspect message content for 'yes'/'no' text -- this would be fragile ... and conflicts with the dedicated RPC endpoints." Task 6.1 subtask 3 reiterates: "Does NOT parse content for confirmation keywords." Confirmations are exclusively via `neo.confirm_action`, `neo.cancel_action`, and `neo.confirm_explicit` RPC endpoints. This is clean.

**4. Singleton Neo session concurrency/lifecycle -- RESOLVED**

Milestone 2 design notes address all three sub-concerns:
- **Concurrent messages**: "A message queue ensures that if a second message arrives while Neo is processing, it waits. The queue is a simple Promise chain." Task 2.1 subtask 4 implements this. Acceptance criteria includes "Concurrent sendMessage calls are serialized (queue)."
- **Session ID on clear**: "On clearSession, the old session is archived ... and a new session is created with a fresh UUID-based session ID stored in GlobalSettings.neo.sessionId. The 'neo:global' alias is updated to point to the new session." This separates message histories.
- **Tool attachment timing**: "MCP tools server is created during provisioning but starts as a placeholder with no tools. Tools are added incrementally ... setRuntimeMcpServers() can be called at any time." This clarifies tools do not need to exist at session creation.

**5. NeoSessionService reconciled with provisionGlobalSpacesAgent pattern -- RESOLVED**

The plan now explicitly follows the module-level provisioning function pattern (not a service class). Milestone 2 says: "module-level provisioning function (not a service class)". Task 2.1 creates `provisionNeoAgent()` as a function returning a `NeoAgentHandle`, directly mirroring `provisionGlobalSpacesAgent()`. The `ProvisionNeoAgentDeps` interface follows the same pattern as `ProvisionGlobalSpacesAgentDeps`. This is consistent with the codebase.

**6. CreateSessionParams.sessionType union updated -- RESOLVED**

Task 1.1 subtask 4 explicitly says: "Add 'neo' to the CreateSessionParams.sessionType union in packages/daemon/src/lib/session/session-lifecycle.ts (this has its own narrower union separate from SessionType)." Acceptance criteria confirms: "SessionType includes 'neo' in both packages/shared and CreateSessionParams in session-lifecycle.ts."

**7. NeoMessage row-mapper clarified -- RESOLVED**

Task 1.1 subtask 5 has a note: "This is a frontend projection type -- Neo messages are stored in the sdk_messages table (same as all sessions) and projected via row mappers. NeoMessage is NOT a DB table." Task 6.2 (LiveQuery) subtask 2 explicitly defines the row mapper: "Map id, role (from SDK message role), content (from message content), createdAt. Extract toolCalls from SDK message tool_use blocks (if present). Handle the sdk_messages row format (which may use JSON columns for content/tool calls)."

**8. Sessions table CHECK constraint migration -- RESOLVED**

Task 1.2 subtask 5 says: "Check if the sessions table has a CHECK constraint on the type column. If so, add an ALTER or migration step to extend it to include 'neo'. (Pattern: check existing migrations for how spaces_global was added.)" I verified the codebase: the sessions table does have a CHECK constraint (`CHECK(type IN ('worker', 'room_chat', 'planner', 'coder', 'leader', 'general', 'lobby', 'spaces_global', 'space_task_agent'))`), so this migration step will be needed. The plan correctly identifies this.

---

### Additional Improvements Noted

- **Migration number**: No longer hardcoded. Task 1.2 subtask 1 says "Identify the next available migration number ... at implementation time (do NOT hardcode)."
- **`explain` tool dropped**: Milestone 5 design notes say "The explain tool from the original design is dropped -- Neo already explains naturally via text responses." This addresses the GLM reviewer's P2.
- **`neo.send` fire-and-forget clarified**: Milestone 6 design notes and the overview both explicitly state `neo.send` is fire-and-forget with frontend subscribing to `neo.messages` LiveQuery.
- **Test boundary clarified**: Milestone 10 goal states: "Milestones 2-5 each include unit tests covering happy paths and basic error cases ... Milestone 10 adds: combinatorial edge cases, cross-component integration tests, coverage threshold enforcement."
- **`clearSession` vs action log**: Overview design note: "clearSession only resets the conversation history ... The neo_action_log table is independent -- action history is preserved across session clears."
- **Undo scope**: Milestone 5 design notes explicitly scope: "Single-level only. Multi-level undo is explicitly out of scope." And: "Undo of deletes only restores the immediate object ... cascading children NOT restored."
- **`update_app_settings` allowlist**: Task 4.5 subtask 2 defines the exact allowlist: "Only the following fields can be updated via Neo: neo (Neo settings), theme, defaultModel, defaultProvider. Security-sensitive fields (apiKeys, oauthTokens, provider credentials) are rejected by the tool handler."
- **Mobile**: Task 7.2 subtask 4 and overview design note clarify: "NavRail is hidden on mobile ... NeoInputBar follows the same responsive pattern -- hidden on mobile. A mobile-specific entry point is out of scope."
- **Activity log tool vs RPC**: Milestone 5 design notes clarify: "Both exist and are backed by the same query. The tool is for Neo's own introspection ... The RPC endpoint is for the frontend Activity Feed tab."

---

### Remaining Minor Observations (P3)

**1. `neo:global` alias mechanism is not fully specified**

Milestone 2 design notes say the `'neo:global'` alias is "updated to point to the new session" on clear. But the mechanism for this alias is not described -- is it a DB lookup by type? A settings field? The `GlobalSettings.neo.sessionId` is mentioned for storing the current session ID, which suggests the provisioning function looks up by `sessionId` from settings. This should be clear enough for implementation but the word "alias" is slightly misleading -- it is really just a stored session ID in settings. Not a blocker.

**2. Tool count concern acknowledged but strategy is vague**

The overview says: "We start with the full tool set but monitor LLM performance. If tool count degrades quality, we will consolidate related tools." This acknowledges the concern but does not specify a concrete measurement (e.g., "if tool selection accuracy drops below X% in online tests"). Acceptable for a plan, but the implementer should be aware this may need iteration.

**3. Error handling acknowledged as deferred**

The overview says: "LLM errors surface as error messages in the Neo chat panel. Context window management uses the SDK's built-in truncation. Cost control and rate limiting are deferred." This is a reasonable scoping decision for V1.

---

### Summary

All 8 previously raised P1 issues have been addressed. The plan is thorough, well-ordered, and feasible. The remaining observations are minor (P3) and not blocking. The plan is ready for implementation.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Seven of the eight previous P1 issues have been resolved. The plan is substantially improved. However, two new P1 gaps surfaced during this review: the `neo:global` alias mechanism is undefined and breaks `clearSession`, and the expected phrase for `require_explicit` actions is never stored and cannot be validated by `neo.confirm_explicit`. One stale terminology reference (`NeoSessionService`) in two subtasks should also be cleaned up.

---

## Previous P1 Issues — Resolution Status

| # | Issue | Status |
|---|-------|--------|
| 1 | Circular dependency between origin metadata and write tools | **Resolved** — Task 1.4 added to Milestone 1; write tools in Milestone 4 correctly depend on it |
| 2 | `require_explicit` security behavior defined | **Resolved** — Task 4.1 subtask 4 defines the behavior clearly with an example phrase |
| 3 | Chat-text confirmation removed from `neo.send` | **Resolved** — Milestone 6 design note explicitly prohibits it; `neo.confirm_explicit` RPC added |
| 4 | `NeoSessionService` reconciled with `provisionGlobalSpacesAgent` pattern | **Mostly resolved** — Milestone 2 now describes a `provisionNeoAgent()` module-level function and correctly references the existing pattern. However, two stale references remain: Milestone 3 Task 3.4 subtask 4 and Milestone 4 Task 4.6 subtask 5 still say "Wire ... into `NeoSessionService`" instead of `provisionNeoAgent` / `NeoAgentHandle`. |
| 5 | `CreateSessionParams.sessionType` union updated | **Resolved** — Task 1.1 subtask 4 explicitly adds `'neo'` to the `session-lifecycle.ts` union |
| 6 | `NeoMessage` row-mapper clarified | **Resolved** — Task 6.2 subtask 2 defines the row mapper from `sdk_messages` to `NeoMessage`; Milestone 1 types note clarifies this is a frontend projection type |
| 7 | Sessions table CHECK constraint migration | **Resolved** — Task 1.2 subtask 5 explicitly instructs the implementer to check and extend the constraint |
| 8 | `neo.send` fire-and-forget response model documented | **Resolved** — Both Milestone 6 design notes and the `00-overview.md` document this clearly |

---

## P1: New Critical Gaps

### 1. `neo:global` alias mechanism is undefined — breaks `clearSession`

Milestone 2 design note says: "On `clearSession`, ... a new session is created with a fresh UUID-based session ID ... The `'neo:global'` alias is updated to point to the new session." But:

- The `sessions` table has no alias column (confirmed: no such column exists in any migration).
- No migration in Task 1.2 adds an alias or pointer column.
- Task 2.1 subtask 5 says "Implement `clearSession()`: archive current session, create fresh session, reattach MCP tools and system prompt" — no mention of how the alias is updated.
- Milestone 6 Task 6.2 subtask 1 hardcodes `neo.messages` as `sdk_messages WHERE session_id = 'neo:global'`. After `clearSession` creates a UUID session, `sdk_messages` rows will have the UUID as `session_id`, not `'neo:global'`. The LiveQuery will return no messages.

The plan must either: (a) keep the session ID as the literal string `'neo:global'` forever (simplest — just reuse the same sessions row and clear its messages via `sdk_messages DELETE`), or (b) add an alias/pointer column to the sessions table and define how both the repository and LiveQuery use the indirection. Without resolving this, `clearSession` will silently break message history display.

### 2. Expected phrase for `require_explicit` is never stored — `neo.confirm_explicit` cannot validate it

Task 4.1 defines `require_explicit`: Neo generates a phrase like "DELETE my-project" and the user must type it back. Task 6.1 adds `neo.confirm_explicit` RPC which "validates exact phrase match." But:

- The `neo_action_log` table schema (Task 1.2) has no column to store the expected phrase.
- `NeoActionStatus` has no `pending_explicit` value to distinguish it from `pending_confirmation`.
- The `NeoActionLogger.wrapAction()` in Task 4.2 subtask 2 says "If `require_explicit`: return message explaining the action requires explicit confirmation" but does not store the phrase.
- Without a stored phrase, `neo.confirm_explicit` cannot validate it — the only alternative is regenerating the phrase from the action log data at validation time, but this logic is undefined.

The plan must either: add an `expected_phrase TEXT` column to `neo_action_log`, add `pending_explicit` to `NeoActionStatus`, and store the phrase during `wrapAction`; or define a stateless phrase generation function that is deterministic from action data.

---

## P2: Remaining Design Gaps

### 3. Autonomous mode behavior with `require_explicit` actions is unresolved

Task 4.1 subtask 3 says "Autonomous: auto-execute everything." But the security tier design also defines `require_explicit` for high-risk operations. If autonomous mode auto-executes high-risk actions (e.g., delete room with active tasks), that is a serious safety regression. The plan should explicitly state whether autonomous mode bypasses `require_explicit` (and if so, add a warning in the UI) or whether `require_explicit` is always enforced regardless of security mode.

### 4. Stale `NeoSessionService` references (P3 elevated to P2)

Two subtasks still use the old class name:
- Milestone 3, Task 3.4, subtask 4: "Attach the MCP server to the Neo session during initialization in `NeoSessionService`"
- Milestone 4, Task 4.6, subtask 5: "Wire the complete tools server into `NeoSessionService`"

These should reference `provisionNeoAgent` / `NeoAgentHandle` to match the chosen pattern. While minor, this will confuse the implementing coder agent who will see both `provisionNeoAgent` (Milestone 2) and `NeoSessionService` (Milestones 3, 4) and reasonably think they are different things.

---

## P3: Nits

### 5. `neo_action_log` action history scoping after `clearSession` is now documented

The overview `00-overview.md` correctly documents that the action log is preserved across session clears. This was previously a gap; it is now resolved.

### 6. Test boundary between per-milestone tests and Milestone 10 is now documented

Milestone 10 and `00-overview.md` both clarify the boundary. The previous ambiguity is resolved.

### 7. Migration number is no longer hardcoded

Task 1.2 subtask 1 says "do NOT hardcode" — resolved.

### 8. E2E tests for `require_explicit` flow are missing

Task 11.2 covers confirmation (confirm/cancel) and auto-execute flows but does not include a `require_explicit` scenario. Given that `require_explicit` has a distinct UI (text input instead of buttons), it should have an E2E test. This is a minor gap since it can be added to Task 11.2 without restructuring.

lsm (commented):

**Recommendation: APPROVE**

All four issues from the previous review have been resolved cleanly:

1. **`neo:global` session alias simplified** -- The session ID is now always the literal string `'neo:global'`. `clearSession` deletes messages via `DELETE FROM sdk_messages WHERE session_id = 'neo:global'` and resets the session row. No alias/pointer mechanism. Consistently documented in `00-overview.md` (line 22), M2 design notes (lines 11-13), and Task 2.1 subtask 5.

2. **`require_explicit` phrase storage** -- `expected_phrase TEXT` column added to `neo_action_log` schema (M1 Task 1.2 line 56). `pending_explicit` added to `NeoActionStatus` (M1 Task 1.1 line 18). `NeoActionLog` interface includes `expectedPhrase` field (line 19). `wrapAction` stores the phrase (M4 Task 4.2 subtask 2 line 47). `neo.confirm_explicit` RPC validates against stored phrase (M6 Task 6.1 line 29).

3. **Autonomous mode clarified** -- Task 4.1 subtask 3 explicitly states autonomous mode auto-executes everything including high-risk actions, with a UI warning when selecting the mode.

4. **Stale `NeoSessionService` references cleaned up** -- Zero matches for `NeoSessionService` across the entire plan. M3 Task 3.4 and M4 Task 4.6 now correctly reference `NeoAgentHandle.getSession().setRuntimeMcpServers()`.

### P3 nits (not blocking)

**P3-1: Stale test description in M10 Task 10.3** (`10-unit-and-online-tests.md` lines 68, 71)
- Test file is named `neo-session-service.test.ts` but the module is `provision-neo-agent.ts`. Task 2.1 subtask 8 already defines tests in `provision-neo-agent.test.ts`, so this file name is both stale and duplicative.
- Line 71 says "Test `clearSession` creates a fresh session with new ID" which contradicts the fix -- `clearSession` now deletes messages while keeping the same `'neo:global'` session ID.

**P3-2: `neo.pending_actions` LiveQuery excludes `pending_explicit`** (`06-rpc-handlers-and-backend-wiring.md` line 52)
- The query filters `WHERE status = 'pending_confirmation'` but should also include `'pending_explicit'` to surface all actions awaiting user input. An implementer would likely catch this, but worth noting.

**P3-3: `NeoStore` missing `confirmExplicit` method** (`07-frontend-navrail-and-neo-panel.md` lines 23-29)
- The store lists `confirmAction` and `cancelAction` but not `confirmExplicit(actionId, phrase)` for calling `neo.confirm_explicit` RPC. The M8 UI component references this RPC directly. An implementer would add it, but the store method list is incomplete.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Three P1 issues remain after the previous round of fixes.

---

### What was verified

The four specific areas from the re-review request were checked in full.

**1. `neo:global` session alias — PASS**

The literal-string session-ID design is coherent and complete. `00-overview.md`, Milestone 2 design notes, and Task 2.1 subtasks all agree: the ID is always `'neo:global'`, `clearSession` deletes messages from `sdk_messages WHERE session_id = 'neo:global'` and resets the session row, no alias/pointer mechanism is needed, and LiveQuery filters remain stable. This part is clean.

**2. `require_explicit` phrase storage — LARGELY PASS with one gap (P1)**

The `expected_phrase TEXT` column in the DB schema (Task 1.2), the `pending_explicit` status in the `NeoActionStatus` union (Task 1.1), and the `neo.confirm_explicit` RPC handler (Task 6.1) are all consistent and sufficient for phrase validation. However the LiveQuery and repository layers do not cover `pending_explicit` — see P1 #2 below.

**3. Autonomous mode + require_explicit clarification — PASS**

Task 4.1 subtask 3 and subtask 4 are now explicit: autonomous mode auto-executes everything including high-risk, and the settings UI shows a warning when the user selects autonomous. The distinction between `confirm` (one-click card) and `require_explicit` (typed phrase) is clearly documented. No ambiguity remains here.

**4. Stale `NeoSessionService` references — PARTIAL (P1 #1 remains)**

All class-based `NeoSessionService` references have been removed from the design. However one stale artefact survived in Task 10.3.

---

### P1 Issues

**P1 #1 — `docs/plans/.../10-unit-and-online-tests.md` Task 10.3, subtask 1: stale test file name and contradictory assertion**

The test file is named `neo-session-service.test.ts`, which refers to the old service-class design that was replaced by `provisionNeoAgent()` (a module-level function). The correct filename should match the implementation artifact, e.g. `provision-neo-agent.test.ts`.

More critically, the bullet point reads:

> Test `clearSession` creates a fresh session with new ID

This directly contradicts the accepted design: `clearSession` preserves the `'neo:global'` ID and only deletes messages/resets runtime state. A coder following this test spec would write a test that asserts a new session ID is generated, causing the test to fail against a correct implementation — or worse, cause a developer to change the implementation to match the wrong assertion.

**P1 #2 — `docs/plans/.../06-rpc-handlers-and-backend-wiring.md` Task 6.2 and `01-shared-types-and-database-schema.md` Task 1.3: `pending_explicit` actions are invisible to the frontend**

Three related gaps form a single broken data path:

- Task 6.2, subtask 1: `neo.pending_actions` LiveQuery is defined as `WHERE status = 'pending_confirmation'`. `pending_explicit` actions are excluded, so the frontend `NeoStore.pendingActions` signal will never receive `require_explicit` confirmations.
- Task 1.3, subtask 2: `getPendingConfirmations(): NeoActionLog[]` fetches only `status = 'pending_confirmation'`. The method name implies it covers all pending states, but it silently drops `pending_explicit` rows.

The consequence: when Neo issues a `require_explicit` confirmation (e.g., for a high-risk deletion), no action card appears in the frontend chat. The user cannot proceed. The filter should be `WHERE status IN ('pending_confirmation', 'pending_explicit')` in both the LiveQuery definition and the repository method (or the method should be split or renamed). Both task descriptions need updating.

**P1 #3 — `docs/plans/.../07-frontend-navrail-and-neo-panel.md` Task 7.1: `NeoStore` missing `confirmExplicit` method**

The `NeoStore` methods list in Task 7.1 includes `confirmAction`, `cancelAction`, but not `confirmExplicit(actionId: string, phrase: string)`. The `NeoActionCard` component (Task 8.1, subtask 1) calls `neo.confirm_explicit` RPC for `require_explicit` tier. There is no store method wiring that call. A coder implementing Task 7.1 before Task 8.1 will not add this method, and Task 8.1 will have no store method to call — it would have to go directly to the RPC hub, breaking the store abstraction pattern.

The method `confirmExplicit(actionId: string, phrase: string): Promise<void>` must be added to the Task 7.1 `NeoStore` methods list.

---

### P3 Nit

**`docs/plans/.../00-overview.md` Milestone 6 summary**: The overview line for Milestone 6 lists `neo.confirm_action` and `neo.cancel_action` but omits `neo.confirm_explicit`, which is a first-class RPC endpoint added to address `require_explicit` confirmations. Minor documentation inconsistency — the detail file (06-rpc-handlers) is correct.

lsm (commented):

**Recommendation: APPROVE** — All three P1 items from the previous review have been correctly addressed. Only P3 nits remain.

---

## Re-Review: Three Previously Raised P1 Items

### 1. Task 10.3 renamed and clearSession assertion corrected

`10-unit-and-online-tests.md` Task 10.3, subtask 3 now reads: "Test `clearSession` deletes messages but preserves `'neo:global'` session ID". The file name is `packages/daemon/tests/unit/neo/provision-neo-agent.test.ts`. Both corrections are in place and match the design note in `02-neo-session-manager.md` which states messages are deleted from `sdk_messages` but the session row is reset with the same ID, never replaced with a new one.

### 2. `neo.pending_actions` LiveQuery now includes `pending_explicit`

`06-rpc-handlers-and-backend-wiring.md` Task 6.2 line 52 explicitly filters: `WHERE status IN ('pending_confirmation', 'pending_explicit')`. This matches `NeoActionLogRepository.getPendingActions()` in `01-shared-types-and-database-schema.md` (same filter). Actions in the `require_explicit` tier are therefore visible in the frontend's `pendingActions` signal. Confirmed.

### 3. `NeoStore.confirmExplicit(actionId, phrase)` exists and `NeoActionCard` can use it

`07-frontend-navrail-and-neo-panel.md` Task 7.1 subtask 3 defines `confirmExplicit(actionId: string, phrase: string): Promise<void>` which calls `neo.confirm_explicit`. Task 8.1 (`NeoActionCard`) correctly describes the `require_explicit` tier as a text input that "validates the phrase via `neo.confirm_explicit` RPC". The store method and the component specification are consistent.

---

## Remaining Issues

### P3: `NeoActionCard` references RPC directly instead of store method

Task 8.1 in `08-frontend-confirmation-ui-and-settings.md` says the explicit confirmation input "validates the phrase via `neo.confirm_explicit` RPC" rather than calling `neoStore.confirmExplicit()`. The store method exists precisely to encapsulate RPC calls. This wording may lead the implementer to bypass the store and call RPC directly from the component. Recommend updating the wording to "calls `neoStore.confirmExplicit(actionId, phrase)`" for consistency with how Confirm and Cancel buttons are specified in the same subtask.

### P3: Milestone overview vs. task boundary for Milestone 9 frontend components

`00-overview.md` line 64 says Milestone 9 "requires identifying existing frontend components that need modification" — this was addressed in the overview but Task 9.x tasks in `09-frontend-via-neo-indicators.md` should be verified to have the explicit list of components to modify. This is a documentation completeness nit and does not block implementation.

---

## Summary

All three previously raised P1 issues are correctly resolved: the test file is properly named, the clearSession assertion preserves the session ID instead of creating a new one, `pending_explicit` actions are now visible via the LiveQuery filter, and `confirmExplicit(actionId, phrase)` is defined on `NeoStore` so `NeoActionCard` has what it needs. The plan is ready for implementation.

lsm (commented):

**Recommendation: APPROVE**

All four fixes from the previous round have been applied correctly and consistently across the plan:

### Verified Fixes

1. **Task 10.3 test file renamed and assertion corrected** (`10-unit-and-online-tests.md` lines 68-71) -- File is now `provision-neo-agent.test.ts` (matching the implementation module). The `clearSession` test bullet reads "deletes messages but preserves `'neo:global'` session ID" -- consistent with the design in `00-overview.md` and Milestone 2.

2. **`neo.pending_actions` LiveQuery filter corrected** (`06-rpc-handlers-and-backend-wiring.md` line 52) -- Now filters `status IN ('pending_confirmation', 'pending_explicit')`, surfacing both confirmation tiers to the frontend. The repository method in Task 1.3 (`01-shared-types-and-database-schema.md` line 85) was also renamed from `getPendingConfirmations` to `getPendingActions` with the matching `IN` clause.

3. **`NeoStore.confirmExplicit` method added** (`07-frontend-navrail-and-neo-panel.md` line 27) -- `confirmExplicit(actionId: string, phrase: string): Promise<void>` is now listed in the store methods, bridging the gap between `NeoActionCard` (Task 8.1) and the `neo.confirm_explicit` RPC.

4. **Overview Milestone 6 summary includes `neo.confirm_explicit`** (`00-overview.md` line 42) -- The RPC endpoint list now includes all three confirmation-related endpoints.

### Additional Cleanup Verified

- Zero matches for `NeoSessionService` or `neo-session-service` across the entire plan.
- Zero matches for `getPendingConfirmations` -- all references use the corrected `getPendingActions`.
- The `require_explicit` E2E test scenario was added to Task 11.2 (`11-e2e-tests.md` line 44), covering the typed-phrase flow including wrong-phrase error case.

No P0, P1, or P2 issues remain. The plan is internally consistent and ready for implementation.

PR #1091, opened 2026-03-28
fix(web): rename spaceWorkflowRun RPC call from .create to .start

lsm:
The daemon registers the handler as `spaceWorkflowRun.start` (returning `{ run: SpaceWorkflowRun }`), but `space-store.ts` was calling `spaceWorkflowRun.create` and treating the response as a plain `SpaceWorkflowRun`.

- Rename RPC call from `spaceWorkflowRun.create` to `spaceWorkflowRun.start`
- Destructure `{ run }` from the response
- Remove stale TODO(M6) comment
- Update unit test mock and assertion to match

### DIFF
--- a/packages/web/src/lib/__tests__/space-store.test.ts
+++ b/packages/web/src/lib/__tests__/space-store.test.ts
@@ -134,8 +134,8 @@ function makeMockHub() {
 			// Daemon returns SpaceTask directly (not wrapped)
 			if (method === 'spaceTask.create') return makeTask('new-task');
 			if (method === 'spaceTask.update') return makeTask('t1', 'in_progress');
-			// spaceWorkflowRun.create is a stub — would return SpaceWorkflowRun directly
-			if (method === 'spaceWorkflowRun.create') return makeRun('new-run');
+			// spaceWorkflowRun.start returns wrapped { run: SpaceWorkflowRun }
+			if (method === 'spaceWorkflowRun.start') return { run: makeRun('new-run') };
 			// spaceAgent handlers return wrapped { agent }
 			if (method === 'spaceAgent.create') return { agent: makeAgent('new-agent') };
 			if (method === 'spaceAgent.update') return { agent: makeAgent('a1') };
@@ -803,11 +803,11 @@ describe('SpaceStore — CRUD methods', () => {
 		expect(task.status).toBe('in_progress');
 	});
 
-	it('startWorkflowRun calls spaceWorkflowRun.create RPC', async () => {
+	it('startWorkflowRun calls spaceWorkflowRun.start RPC', async () => {
 		await spaceStore.selectSpace('space-1');
 		await spaceStore.startWorkflowRun({ workflowId: 'wf-1', title: 'Run 1' });
 
-		expect(mockHub.request).toHaveBeenCalledWith('spaceWorkflowRun.create', {
+		expect(mockHub.request).toHaveBeenCalledWith('spaceWorkflowRun.start', {
 			spaceId: 'space-1',
 			workflowId: 'wf-1',
 			title: 'Run 1',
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

The fix is correct and complete. The daemon handler at `packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts:124` registers `spaceWorkflowRun.start` and returns `{ run }` (line 180). The frontend now matches both the RPC name and the response shape.

Changes verified:

1. **`packages/web/src/lib/space-store.ts` (line 851)**: RPC call renamed from `spaceWorkflowRun.create` to `spaceWorkflowRun.start`, response correctly destructured as `{ run }` with proper type annotation `{ run: SpaceWorkflowRun }`.

2. **`packages/web/src/lib/__tests__/space-store.test.ts`**: Mock updated to return `{ run: makeRun('new-run') }` (wrapped response matching daemon behavior), and assertion updated to expect `spaceWorkflowRun.start`. Both changes are consistent with the daemon's actual contract.

3. **Stale TODO(M6) comment removed**: The comment described `spaceWorkflowRun.create` as a stub for a future API, but the daemon handler has since been implemented as `spaceWorkflowRun.start`. Removing it is correct.

4. **No remaining references to `spaceWorkflowRun.create`** anywhere in the codebase -- confirmed via full-repo grep.

No issues found.

lsm (commented):

**Recommendation: APPROVE** (with one minor observation)

---

### Summary

This PR correctly fixes a concrete runtime bug: `space-store.ts` was calling `spaceWorkflowRun.create`, a handler that does not exist in the daemon. The daemon registers `spaceWorkflowRun.start` and returns `{ run: SpaceWorkflowRun }`. The fix renames the RPC call and destructures the response appropriately. The stale TODO(M6) comment, which was factually wrong (the handler now exists), is correctly removed.

---

### What was verified

**Daemon handler** (`packages/daemon/src/lib/rpc-handlers/space-workflow-run-handlers.ts`, line 124):
- `messageHub.onRequest('spaceWorkflowRun.start', ...)` is confirmed registered.
- The handler returns `{ run }` at line 180 — matching the new destructuring in the store.

**`space-store.ts`** (`packages/web/src/lib/space-store.ts`, lines 851–855):
- RPC call renamed from `spaceWorkflowRun.create` to `spaceWorkflowRun.start`. Correct.
- Response typed as `{ run: SpaceWorkflowRun }` and destructured. Correct.
- Pattern is consistent with `createAgent` and `createWorkflow`, which also destructure wrapped responses.

**Test mock** (`packages/web/src/lib/__tests__/space-store.test.ts`, line 138):
- Mock updated from `return makeRun('new-run')` to `return { run: makeRun('new-run') }`. Correct.

**Test assertion** (line 810):
- Assertion updated to expect `spaceWorkflowRun.start`. Correct.

---

### Observation (P2 — test coverage gap, pre-existing pattern)

The test at line 806–815 verifies the RPC method name and payload but does **not** assert that the return value of `startWorkflowRun()` is the unwrapped `run` object. Compare with `createTask` (line 781), which does:
```ts
const task = await spaceStore.createTask(...);
expect(task.id).toBe('new-task');
```
A similar assertion for `startWorkflowRun` — e.g., `expect(run.id).toBe('new-run')` — would have caught the original `{ run }` destructuring bug at test time rather than at runtime. This was a pre-existing gap in the test, but since this PR adds the destructuring logic, it's the right moment to add that assertion.

This is minor and does not block the fix from being correct and safe to ship.

---

### No issues found

- No security concerns.
- No null-safety risk: `hub.request<T>` returns `Promise<T>`, not `Promise<T | null>`, so the destructuring cannot throw on a missing key from a null response.
- No callers of `startWorkflowRun` in UI components yet, so there is no integration surface that could break.
- All other RPC calls in `space-store.ts` are unaffected.

PR #1094, opened 2026-03-28
feat(web): add SpaceDetailPanel as space-level ContextPanel (M1.1)

lsm:
Implements `SpaceDetailPanel` — the space-level context panel mirroring `RoomContextPanel`'s structure.

## What's included

- **Stats strip**: active · review · done task counts from `spaceStore.tasks`
- **Pinned items**: Dashboard (highlights when no session/task selected) and Space Agent (highlights when `currentSpaceSessionIdSignal === 'space:chat:{spaceId}'`)
- **Workflow Runs section**: collapsible, shows `spaceStore.activeRuns` with expandable nested task lists per run
- **Tasks section**: standalone tasks (without `workflowRunId`) with active/review/done tab filter
- **Sessions section**: collapsed by default; always lists space agent session, task agent sessions via `SpaceTask.taskAgentSessionId`, and manually created sessions from `Space.sessionIds`; deduplicates entries

Space Agent navigation uses `navigateToSpaceSession(spaceId, 'space:chat:spaceId')` — `navigateToSpaceAgent` will be added in Task 2.4 and can be swapped in.

## Tests

38 unit tests in `SpaceDetailPanel.test.tsx` covering all sections, highlighting logic, tab filtering, run expansion, click navigation, session deduplication, and empty states.

### DIFF
--- a/packages/web/src/islands/ContextPanel.tsx
+++ b/packages/web/src/islands/ContextPanel.tsx
@@ -3,6 +3,7 @@ import {
 	navSectionSignal,
 	contextPanelOpenSignal,
 	currentRoomIdSignal,
+	currentSpaceIdSignal,
 	settingsSectionSignal,
 	createRoomModalSignal,
 	type NavSection,
@@ -31,6 +32,7 @@ import { SessionList } from './SessionList.tsx';
 import { RoomList } from './RoomList.tsx';
 import { RoomContextPanel } from './RoomContextPanel.tsx';
 import { SpaceContextPanel } from '../components/space/SpaceContextPanel.tsx';
+import { SpaceDetailPanel } from './SpaceDetailPanel.tsx';
 import { SpaceCreateDialog } from '../components/space/SpaceCreateDialog.tsx';
 import { spaceStore } from '../lib/space-store.ts';
 import { ConnectionNotReadyError } from '../lib/errors.ts';
@@ -172,12 +174,16 @@ export function ContextPanel() {
 	}, [navSection]);
 	const activeSettingsSection = settingsSectionSignal.value;
 	const currentRoomId = currentRoomIdSignal.value;
+	const currentSpaceId = currentSpaceIdSignal.value;
 
 	// Inbox takes full content width — no sidebar needed
 	if (navSection === 'inbox') return null;
 
 	// When a specific room is selected in the rooms section, show room-specific panel
 	const isRoomDetail = navSection === 'rooms' && currentRoomId !== null;
+
+	// When a specific space is selected in the spaces section, show space-specific panel
+	const isSpaceDetail = navSection === 'spaces' && currentSpaceId !== null;
 
 	// Section config
 	const sectionConfig = {
@@ -236,9 +242,11 @@ export function ContextPanel() {
 	const isSpaces = navSection === 'spaces';
 	const headerTitle = isRoomDetail
 		? (roomStore.room.value?.name ?? 'Room')
-		: isSpaces
-			? null
-			: config.title;
+		: isSpaceDetail
+			? (spaceStore.space.value?.name ?? 'Space')
+			: isSpaces
+				? null
+				: config.title;
 
 	const handleCreateSession = async () => {
 		if (connectionState.value !== 'connected') {
@@ -376,7 +384,7 @@ export function ContextPanel() {
 				{/* Header */}
 				<div class={`p-4 border-b ${borderColors.ui.default}`}>
 					<div class="flex items-center justify-between mb-3">
-						{isSpaces ? (
+						{isSpaces && !isSpaceDetail ? (
 							<button
 								onClick={() => navigateToSpaces()}
 								class={cn(
@@ -445,7 +453,9 @@ export function ContextPanel() {
 
 				{/* Content — key triggers fade-in (animate-fadeIn) on section change */}
 				<div
-					key={navSection + (isRoomDetail ? '-detail' : '')}
+					key={
+						navSection + (isRoomDetail ? '-detail' : '') + (isSpaceDetail ? '-space-detail' : '')
+					}
 					class="flex-1 overflow-hidden flex flex-col animate-fadeIn"
 				>
 					{navSection === 'home' && (
@@ -463,7 +473,13 @@ export function ContextPanel() {
 					{navSection === 'rooms' && !isRoomDetail && (
 						<RoomList onRoomSelect={() => (contextPanelOpenSignal.value = false)} />
 					)}
-					{navSection === 'spaces' && (
+					{navSection === 'spaces' && isSpaceDetail && (
+						<SpaceDetailPanel
+							spaceId={currentSpaceId!}
+							onNavigate={() => (contextPanelOpenSignal.value = false)}
+						/>
+					)}
+					{navSection === 'spaces' && !isSpaceDetail && (
 						<SpaceContextPanel
 							onSpaceSelect={() => (contextPanelOpenSignal.value = false)}
 							onCreateSpace={() => setCreateSpaceOpen(true)}
--- a/packages/web/src/islands/SpaceDetailPanel.tsx
+++ b/packages/web/src/islands/SpaceDetailPanel.tsx
@@ -1,0 +1,398 @@
+/**
+ * SpaceDetailPanel
+ *
+ * Space-specific context panel for the three-column layout.
+ * Mirrors RoomContextPanel's visual structure for a space:
+ * 1. Task stats strip (active · review · done counts)
+ * 2. Pinned items: Dashboard, Space Agent
+ * 3. Workflow Runs section (active runs with expandable tasks)
+ * 4. Tasks section (standalone tasks with active/review/done tab filter)
+ * 5. Sessions section (collapsible, default collapsed)
+ */
+
+import { useMemo, useState } from 'preact/hooks';
+import { CollapsibleSection } from '../components/room/CollapsibleSection';
+import { spaceStore } from '../lib/space-store';
+import { navigateToSpace, navigateToSpaceSession, navigateToSpaceTask } from '../lib/router';
+import { currentSpaceSessionIdSignal, currentSpaceTaskIdSignal } from '../lib/signals';
+import { cn } from '../lib/utils';
+
+type OrphanTab = 'active' | 'review' | 'done';
+
+const taskStatusColors: Record<string, string> = {
+	draft: 'bg-gray-500',
+	pending: 'bg-yellow-500',
+	in_progress: 'bg-blue-500',
+	review: 'bg-purple-500',
+	needs_attention: 'bg-orange-500',
+	completed: 'bg-green-500',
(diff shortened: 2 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE** (all issues are P2/P3, nothing blocking)

Overall this is a well-structured component that cleanly mirrors `RoomContextPanel`. The tests are thorough (38 tests covering stats, highlighting, navigation, tab filtering, deduplication, empty states). The signal integration, mock setup, and CollapsibleSection reuse are all solid. A few suggestions below.

---

### P2: `rate_limited`, `usage_limited`, and `archived` tasks are invisible

**`packages/web/src/islands/SpaceDetailPanel.tsx` lines 79-93, 118-125**

Three valid `SpaceTaskStatus` values — `rate_limited`, `usage_limited`, and `archived` — don't map to any stats count category or tab filter. Tasks with these statuses silently disappear from the panel. `taskStatusColors` (line 22-31) includes entries for `rate_limited` and `usage_limited` but not `archived`, further suggesting these were partially considered but not fully wired.

Suggestion: Either fold `rate_limited`/`usage_limited` into the "active" or "review" bucket (since they likely need attention), add `archived` to "done", or add a fourth tab/category. At minimum, document which statuses are intentionally excluded.

### P2: `orphanTasksForTab` is not memoized

**`packages/web/src/islands/SpaceDetailPanel.tsx` lines 118-125**

The `orphanTasksForTab` derivation is a plain expression that re-filters on every render. The stats counts (`activeCount`, `reviewCount`, `doneCount`) correctly use `useMemo` with `[tasks]` as dependency. For consistency and to avoid unnecessary work with large task lists, `orphanTasksForTab` should also be wrapped in `useMemo` with `[standaloneTasks, orphanTab]` as dependencies.

### P3: No-op "Create session" button

**`packages/web/src/islands/SpaceDetailPanel.tsx` lines 353-369**

The "Create session" button in the Sessions header only calls `e.stopPropagation()`. This is an acceptable placeholder, but consider either hiding it until the feature is ready or adding a `title`/tooltip indicating it's coming soon, so users don't think the button is broken.

### P3: Test `Space` casts use `as unknown as Space`

**`packages/web/src/islands/__tests__/SpaceDetailPanel.test.tsx` lines 472, 512**

The test creates partial `Space` objects and casts them via `as unknown as Space`. This works but is fragile — if the component starts reading other `Space` fields (e.g., `name`, `status`), tests won't catch the missing data. Consider a `makeSpace()` helper similar to `makeTask()` and `makeRun()` that provides all required fields with sensible defaults.

lsm (commented):

The implementation is solid overall — the component structure mirrors `RoomContextPanel` as required, signal-based reactivity is correctly set up, and the 38 unit tests all pass. However there are several functional gaps worth fixing before merging.

---

### P1 — `rate_limited` and `usage_limited` tasks are silently invisible

`packages/web/src/islands/SpaceDetailPanel.tsx` lines 79–92 and lines 118–125

The stats strip counts only `draft | pending | in_progress` (active), `review | needs_attention` (review), and `completed | cancelled` (done). Tasks in `rate_limited` or `usage_limited` state — both valid values of `SpaceTask['status']` (defined in `packages/shared/src/types/space.ts` lines 130–131) — fall into none of these buckets. They will not appear in any count, will not appear under any tab in the Tasks section, and will not appear in the stats strip. A user would have no way to discover these tasks through the panel.

The same gap exists in `orphanTasksForTab` at lines 118–125 — all three branches omit `rate_limited` / `usage_limited`. Depending on intended UX these could map to the active bucket (transient throttle) or the review bucket (needs human attention), but they should not be uncategorized and invisible.

Neither the `makeTask` helper nor the test suite exercises either of these two statuses, so this is also a test coverage gap.

---

### P1 — No integration wiring: component is exported but never imported anywhere

The component is not rendered anywhere in the app. `ContextPanel.tsx` handles the spaces section entirely through `SpaceContextPanel` (the thread-style list). There is no `isSpaceDetail` branch analogous to the `isRoomDetail` guard at line 180 / lines 457–465 of `ContextPanel.tsx` that would mount `SpaceDetailPanel`.

The PR description acknowledges this is Task M1.1 and integration is deferred to a future task, but without integration the panel is unreachable by users and unverifiable end-to-end. `currentSpaceIdSignal` already exists in `packages/web/src/lib/signals.ts` line 39, so the wiring is straightforward. At minimum, integrating the component behind that signal — even as a draft — would confirm the props interface is right and prevent it from drifting before the next task lands.

---

### P2 — `Create session` button is a no-op placeholder with no disabled state

`packages/web/src/islands/SpaceDetailPanel.tsx` lines 353–370

The `onClick` handler contains only a comment: `// Future: create a new session in this space`. The button is rendered, has an `aria-label`, and the tests verify it exists, but clicking it silently does nothing. The button should carry a `disabled` attribute (or at minimum a `title` note) so users are not confused by a clickable affordance with no effect.

---

### P3 — `useMemo` dependencies break signal auto-tracking for task counts

`packages/web/src/islands/SpaceDetailPanel.tsx` lines 79–93

Signal values are read outside the memos (`tasks = spaceStore.tasks.value` at line 69) and then passed as plain arrays into memo dependency arrays. This works because every signal update produces a new array reference, so the memo invalidation happens to fire correctly. But inside the `useMemo` callbacks at lines 81 and 87, `tasks` is a captured snapshot value, not a live signal — which means any reactive framework tooling expecting signals inside memos would miss the subscription. The `sessions` memo at line 131 lists `tasks` in its dep array for the same reason. This is the same pattern used in `RoomContextPanel.tsx` so it is consistent, but worth noting as intentional rather than accidental.

---

### P3 — Nit: SVG attributes use kebab-case instead of camelCase JSX props

`stroke-linecap`, `stroke-linejoin`, `stroke-width` appear throughout lines 222–250. Preact tolerates this via DOM passthrough and the linter does not flag it — the rest of the codebase does the same — so no action is required unless the project standardizes on camelCase SVG props.

lsm (commented):

**Recommendation: APPROVE**

The worker addressed all four issues from the previous review correctly. The component is now wired into `ContextPanel.tsx`, the status-bucket mapping is complete, `orphanTasksForTab` is memoized, and the Create session button is disabled with a tooltip. The test count grew from 38 to 43 and the new tests directly cover the added status paths. Below are the remaining minor observations.

---

### P2 — `taskStatusColors` is missing the `archived` status

**`packages/web/src/islands/SpaceDetailPanel.tsx` lines 22–32**

The `taskStatusColors` map contains entries for every `SpaceTaskStatus` except `archived`. A task with `status === 'archived'` will fall through to the `?? 'bg-gray-500'` default in `TaskStatusDot`, which is harmless at runtime but is inconsistent — all other statuses have explicit colors. Since `archived` is now correctly counted in the `done` bucket and shown under the Done tab, its dot will silently use the generic grey fallback instead of the intended `bg-gray-600` (used by `cancelled`, which is semantically closest). A one-liner fix: add `archived: 'bg-gray-600'` to the map.

---

### P2 — `ContextPanel` does not call `spaceStore.selectSpace` when entering the space detail view

**`packages/web/src/islands/ContextPanel.tsx` lines 476–481**

`SpaceDetailPanel` reads `spaceStore.tasks`, `spaceStore.activeRuns`, `spaceStore.standaloneTasks`, etc. Those signals are only populated after `spaceStore.selectSpace(spaceId)` is called. That call is currently made exclusively inside `SpaceIsland.tsx` (line 83), which is the main content island. If a user navigates to a space URL and the main island is mounted before the context panel, data will appear in both places. However, if the ContextPanel renders before the main island completes loading (e.g., the island is lazy), the panel will transiently show "No tasks" / "No active runs" even when tasks exist.

More importantly, when `SpaceDetailPanel` is rendered from the sidebar before `SpaceIsland` mounts — which can happen during cold navigation in the three-column layout — `spaceStore.tasks.value` will be an empty array and the panel will show an empty state. There is no loading indicator in `SpaceDetailPanel` to communicate this. This was a pre-existing gap but the wiring commit that added `isSpaceDetail` did not address it. Worth a follow-up note or a guard.

---

### P3 — `makeTask` test helper omits required `SpaceTask` fields

**`packages/web/src/islands/__tests__/SpaceDetailPanel.test.tsx` lines 104–120**

`SpaceTask` has two required fields that `makeTask` does not supply: `taskNumber: number` and `description: string`. The object is cast with `as SpaceTask` so TypeScript accepts it, but the cast hides the missing fields. If the component ever reads `task.description` or `task.taskNumber`, existing tests will silently pass while rendering `undefined`. Consider adding both with sensible defaults (`taskNumber: 0, description: ''`).

---

### P3 — Sessions section always shows "Space Agent" in the pinned area and again in the expanded session list

**`packages/web/src/islands/SpaceDetailPanel.tsx` lines 252–275 and lines 395–415**

"Space Agent" appears as a dedicated pinned button (always visible) and also appears as the first item inside the Sessions collapsible section. This duplication is intentional (the test at line 505–511 asserts two occurrences), but it may confuse users who click "Space Agent" from the Sessions list and then see it highlighted there but not in the pinned button area (because `isSpaceAgentSelected` also sets the pinned button style). The UX is currently consistent — both highlight simultaneously — but the double-listing could be reconsidered once `navigateToSpaceAgent` lands in Task 2.4.

---

All prior P1/P2 items were resolved. No blocking issues remain.

lsm (commented):

**Recommendation: APPROVE**

1. **Invisible task statuses (was P1/P2)** -- Fixed. `rate_limited` and `usage_limited` are now folded into the active bucket in both the stats strip (lines 81-92) and `orphanTasksForTab` (lines 128-137). `archived` is folded into the done bucket (lines 98-103, 142-144). Comments document the rationale clearly.

2. **ContextPanel integration (was P1)** -- Fixed. `SpaceDetailPanel` is now rendered at `ContextPanel.tsx` lines 476-480 behind the `isSpaceDetail` guard (`navSection === 'spaces' && currentSpaceId !== null`). The header title, animation key, and conditional rendering all follow the same pattern as `RoomContextPanel`.

3. **orphanTasksForTab memoization (was P2)** -- Fixed. Now wrapped in `useMemo` with `[standaloneTasks, orphanTab]` dependencies (line 128).

4. **No-op Create session button (was P3)** -- Fixed. Button is now `disabled` with `cursor-not-allowed` styling and a `title="Session creation coming soon"` tooltip (lines 373-390).

Tests are comprehensive (43 tests, all passing) and include new coverage for the previously missing statuses: `rate_limited` (line 448), `usage_limited` (line 454), and `archived` (line 460). The `makeSpace` helper was added (lines 140-151) to address the fragile casting concern.

---

### P3: `taskStatusColors` map missing `archived` entry

`packages/web/src/islands/SpaceDetailPanel.tsx` lines 22-32

The `taskStatusColors` map includes entries for `rate_limited` and `usage_limited` but not `archived`. When an archived task appears in the Done tab, its `TaskStatusDot` falls back to the generic `bg-gray-500` via the `??` fallback. This works, but adding an explicit `archived: 'bg-gray-400'` (or similar) entry would be more intentional and consistent with how the other status colors are defined.

PR #1095, opened 2026-03-28
fix(e2e): resolve WorkflowCanvas channel name-to-UUID mismatch

lsm:
## Summary
- **WorkflowCanvas.tsx**: Added `buildNameToIdMap`/`resolveNodeId` helpers to translate channel `from`/`to` node names (e.g. "Planning", "Reviewer 1") to node UUIDs. The layout engine and channel renderer used UUID keys but channels reference nodes by name, causing all channel lines, gate icons, and DAG layering to silently fail.
- **space-creation.e2e.ts**: Updated test to verify `workflow-canvas-svg` instead of `SpaceDashboard` quick actions ("Quick Actions", "Start Workflow Run", "Create Task"), which are now hidden by `md:hidden` CSS when the canvas panel renders on desktop.

## Root Cause
`seedBuiltInWorkflows` persists channels with node names as `from`/`to` (by design — see doc comment in `built-in-workflows.ts`), but `WorkflowCanvas.tsx` used node UUIDs for layout map keys and channel endpoint lookups. This mismatch meant:
- `computeLayout` successor map was always empty → all nodes placed in one layer
- `renderedChannels` couldn't find node positions → no channel lines drawn
- Gate icons rendered on channel midpoints → never appeared (channels were null)

## Fixes 4 failing E2E jobs
- features-space-approval-gate-rejection (gate icons not found)
- features-space-happy-path-pipeline (gate icons + node labels not visible)
- features-space-creation (dashboard content hidden by CSS)
- features-reviewer-feedback-loop (gate icons not found)

## Test plan
- [x] Typecheck passes
- [x] Lint passes
- [x] All 9337 daemon unit tests pass
- [ ] CI E2E validation on dev

### DIFF
--- a/packages/e2e/tests/features/space-creation.e2e.ts
+++ b/packages/e2e/tests/features/space-creation.e2e.ts
@@ -142,9 +142,10 @@ test.describe('Space Creation UX', () => {
 
 		// Tabbed dashboard should be visible with Dashboard tab active
 		await expect(page.locator('text=Dashboard')).toBeVisible({ timeout: 5000 });
-		await expect(page.locator('text=Quick Actions')).toBeVisible({ timeout: 5000 });
-		await expect(page.locator('text=Start Workflow Run')).toBeVisible({ timeout: 3000 });
-		await expect(page.locator('text=Create Task')).toBeVisible({ timeout: 3000 });
+
+		// On desktop, the workflow canvas panel replaces the dashboard quick actions
+		// Workflows load async after space creation; wait for the canvas SVG to appear
+		await expect(page.getByTestId('workflow-canvas-svg')).toBeVisible({ timeout: 15000 });
 	});
 
 	test('dialog can be closed with Cancel button', async ({ page }) => {
--- a/packages/web/src/components/space/WorkflowCanvas.tsx
+++ b/packages/web/src/components/space/WorkflowCanvas.tsx
@@ -178,6 +178,26 @@ function evalConditionStatus(condition: GateCondition, data: Record<string, unkn
 // ============================================================================
 
 /**
+ * Build a map from node name → node UUID.
+ * Channels use node names for from/to, but layout uses UUIDs.
+ */
+function buildNameToIdMap(nodes: WorkflowNode[]): Map<string, string> {
+	const map = new Map<string, string>();
+	for (const node of nodes) {
+		if (node.name) map.set(node.name, node.id);
+	}
+	return map;
+}
+
+/**
+ * Resolve a channel endpoint (name or UUID) to a node UUID.
+ * Falls back to the original value if it's already a UUID.
+ */
+function resolveNodeId(ref: string, nameToId: Map<string, string>): string {
+	return nameToId.get(ref) ?? ref;
+}
+
+/**
  * Compute node positions using a layered DAG layout.
  * Returns a map from node ID → {x, y, width, height}.
  */
@@ -188,16 +208,20 @@ function computeLayout(workflow: SpaceWorkflow): Map<string, NodeLayout> {
 
 	if (nodes.length === 0) return new Map();
 
-	// Build successor map from channels
+	const nameToId = buildNameToIdMap(nodes);
+
+	// Build successor map from channels (resolving names to UUIDs)
 	const successors = new Map<string, Set<string>>();
 	for (const node of nodes) {
 		successors.set(node.id, new Set());
 	}
 	for (const ch of channels) {
+		const fromId = resolveNodeId(ch.from, nameToId);
 		const targets = Array.isArray(ch.to) ? ch.to : [ch.to];
 		for (const t of targets) {
-			if (successors.has(ch.from) && successors.has(t)) {
-				successors.get(ch.from)!.add(t);
+			const toId = resolveNodeId(t, nameToId);
+			if (successors.has(fromId) && successors.has(toId)) {
+				successors.get(fromId)!.add(toId);
 			}
 		}
 	}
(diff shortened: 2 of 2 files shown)


lsm (commented):

---

## APPROVE — No blocking issues found

### Product fix (WorkflowCanvas.tsx)

The root cause is well-documented and the fix is correct:

- `seedBuiltInWorkflows` persists channels with node names as `from`/`to` (by design — portable format)
- `WorkflowCanvas.tsx` was using node UUIDs as `successors` map keys and channel endpoint lookups
- The mismatch caused: empty successor map (all nodes in one layer), no channel lines, no gate icons

**`buildNameToIdMap`** (lines 184-190): Correctly builds `Map<nodeName, nodeId>` from all nodes with a `name`. Skips nodes without a name — acceptable since unnamed nodes cannot be referenced by name in channels.

**`resolveNodeId`** (lines 196-198): Elegant fallback pattern. If the reference is already a UUID (i.e., not in the name→id map), it returns unchanged. This handles:
- Name references: resolved to UUID
- Already-UUID references: passed through unchanged  
- Non-existent names: fallback to original string, silently skipped downstream

**`computeLayout`** (lines 218-227): Correctly applies `resolveNodeId` to both `ch.from` and each `ch.to` target when building the successor map. The `successors.has()` guard ensures references to non-existent nodes are silently skipped.

**`renderedChannels` memo** (lines 1090-1091): Correctly applies `resolveNodeId` to both channel endpoints. The `nameToId` dependency is properly included in the deps array (line 1095).

### Test fix (space-creation.e2e.ts)

The old assertions checked for `text=Quick Actions`, `text=Start Workflow Run`, and `text=Create Task` — all rendered inside `SpaceDashboard`. At the 1280px desktop viewport, `SpaceIsland.tsx` line 270 applies `md:hidden` to the dashboard fallback when a workflow canvas is shown, hiding those elements by CSS. The new assertion checks `data-testid=workflow-canvas-svg` which is the `SVG` element rendered by `WorkflowCanvas` and is visible on desktop. This is the correct signal to assert.

- **P3 — name collision in `buildNameToIdMap`**: If two nodes somehow share the same name, `Map.set()` silently overwrites. Node names should be unique within a workflow (enforced at the data layer). Not a bug in this fix.
- **P3 — silent channel drop for non-existent node names**: If a channel references a node name not in `workflow.nodes`, `resolveNodeId` returns the name unchanged, `successors.has()` returns false, and the channel is silently dropped. This matches the pre-fix behavior and is acceptable.

### Regression check

- `computeLayout` is a pure function that builds its own internal `nameToId` map — isolated from component state.
- `layout` useMemo recomputes when `workflow` changes — correct.
- `nameToId` useMemo recomputes when `workflow` changes — correct.
- `renderedChannels` depends on `nameToId` and recomputes correctly when `workflow` changes.
- Channel rendering (`layout.get(ch.fromId)`, `layout.get(ch.toId)`) returns null for unresolved channels — safely skipped (line 1191: `if (!fromLayout || !toLayout) return null`).

No regressions identified.

lsm (commented):

---

### WorkflowCanvas.tsx: Name-to-UUID Resolution

**Root cause is correctly identified and the fix is sound.**

The `WorkflowChannel` type defines `from`/`to` as agent name strings (per `WorkflowNodeAgent.name`), but the built-in workflow templates store node names (like `'Planning'`, `'Reviewer 1'`, etc.) in these fields. The backend's `resolveChannels()` in `space-utils.ts` handles this via the `nodeNameToAgents` fallback map, but `WorkflowCanvas.tsx` was treating `from`/`to` as node UUIDs -- a mismatch that made the successor map always empty and all channel rendering no-ops.

**`buildNameToIdMap`** (lines 184-190): Correctly maps `node.name` to `node.id`. All built-in workflow nodes have non-empty names, so the map is complete for the intended use case.

**`resolveNodeId`** (lines 196-198): Uses `nameToId.get(ref) ?? ref` which handles node names (resolved to UUID), UUIDs in the map (harmless duplicate), and UUIDs not in the map (silent pass-through). The fallback is acceptable as a defensive measure -- built-in workflows are well-formed.

**`computeLayout`** (lines 218-227): The fix correctly applies `resolveNodeId` to both `ch.from` and each target in `ch.to` before checking `successors.has()` and adding to the set. The BFS layer assignment now operates on real UUIDs, producing correct multi-layer DAG layouts instead of collapsing everything to one layer.

**`renderedChannels`** (lines 1087-1094): Correctly applies the same resolution so channel lines and gate icons render at correct midpoint positions between node boxes.

**Integration** (lines 1076-1079, 1095): `nameToId` is a `useMemo` derived from `workflow` and added as a dependency to `renderedChannels`. When `spaceStore.workflows` emits a new signal, both `layout` and `renderedChannels` recompute together with the fresh name-to-UUID map.

### space-creation.e2e.ts: Test Fix

The original assertions checked for elements hidden by `md:hidden` CSS when the canvas panel is visible on desktop. Checking for CSS-hidden elements is not meaningful. The replacement checks for `data-testid="workflow-canvas-svg"` with a 15-second timeout, appropriate for async workflow loading after space creation.

### Edge Cases Examined

1. **Node with no name**: `buildNameToIdMap` skips nodes where `node.name` is falsy. If a channel referenced such a node by name, `resolveNodeId` would return the name unchanged and `computeLayout` would silently skip that edge. Not a regression -- built-in nodes always have names.

2. **Stale `nameToId` map**: Both `layout` and `renderedChannels` depend on `workflow`. When the workflow signal updates, both recompute together. No staleness issue.

3. **`WorkflowCanvas` key**: The existing keying strategy (`${defaultWorkflow.id}:${activeRun?.id ?? 'template'}`) is unaffected by this fix.

### No Regressions Found

The changes are isolated to `WorkflowCanvas.tsx` and the test file. No other components depend on the internal `computeLayout` or `renderedChannels` behavior. The fix targets the mismatch at both call sites consistently.

---

**Recommendation: APPROVE**

No P0/P1/P2 issues identified.

P0: 0 | P1: 0 | P2: 0 | P3: 0

**Summary:** The name-to-UUID resolution is logically correct and complete for the built-in workflow model. Applied consistently at both affected call sites. The test change correctly verifies the visible canvas SVG instead of CSS-hidden dashboard content. No regressions introduced.

PR #1097, opened 2026-03-28
refactor(web): use spaceStore.createTask() in SpaceCreateTaskDialog

lsm:
Refactors `SpaceCreateTaskDialog` to use `spaceStore.createTask()` instead of calling the hub directly.

Changes:
- Remove `spaceId` prop — store already knows the active space via `spaceStore.spaceId.value`
- Replace `connectionManager.getHubIfConnected()` + `hub.request('spaceTask.create', ...)` with `spaceStore.createTask()`
- Remove redundant `spaceStore.selectSpace()` refresh — `space.task.created` real-time events already update the task list
- Update tests to mock `spaceStore.createTask` instead of `connectionManager`

### DIFF
--- a/packages/web/src/components/space/SpaceCreateTaskDialog.tsx
+++ b/packages/web/src/components/space/SpaceCreateTaskDialog.tsx
@@ -5,14 +5,12 @@
 import { useState } from 'preact/hooks';
 import { Modal } from '../ui/Modal';
 import { Button } from '../ui/Button';
-import { connectionManager } from '../../lib/connection-manager';
 import { spaceStore } from '../../lib/space-store';
 import { toast } from '../../lib/toast';
 import type { SpaceTask, SpaceTaskPriority, SpaceTaskType } from '@neokai/shared';
 
 interface SpaceCreateTaskDialogProps {
 	isOpen: boolean;
-	spaceId: string;
 	onClose: () => void;
 	onCreated?: (task: SpaceTask) => void;
 }
@@ -32,12 +30,7 @@ const TASK_TYPE_OPTIONS: { value: SpaceTaskType; label: string }[] = [
 	{ value: 'review', label: 'Review' },
 ];
 
-export function SpaceCreateTaskDialog({
-	isOpen,
-	spaceId,
-	onClose,
-	onCreated,
-}: SpaceCreateTaskDialogProps) {
+export function SpaceCreateTaskDialog({ isOpen, onClose, onCreated }: SpaceCreateTaskDialogProps) {
 	const [title, setTitle] = useState('');
 	const [description, setDescription] = useState('');
 	const [priority, setPriority] = useState<SpaceTaskPriority>('normal');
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-scoped refactoring that correctly delegates task creation from the dialog component to `spaceStore.createTask()`.

### What looks good

- **Correct delegation**: The dialog no longer manages connection state or constructs RPC calls directly. `spaceStore.createTask()` (line 805 of `space-store.ts`) already handles `spaceId` resolution from `this.spaceId.value`, hub connectivity checks, and the RPC call — so removing all of that from the dialog is the right move.

- **Removed redundant refresh**: The old `spaceStore.selectSpace(spaceId).catch(() => {})` after task creation was unnecessary since real-time `space.task.created` events handle task list updates. Good cleanup.

- **Prop removal is safe**: `spaceId` was only passed from one call site (`SpaceIsland.tsx` line 310), which has been updated. No other consumers exist.

- **Test coverage is adequate**: All 10 tests were updated to mock `spaceStore.createTask` instead of `connectionManager`. The new test "shows error when no space is selected (createTask throws)" (line 179) replaces the old "shows error when not connected" test, which is a reasonable substitution — the store now owns the "no space" and "not connected" error paths.

- **Error handling preserved**: The catch block at line 72-73 still converts errors to user-visible messages, and the store method throws descriptive errors for both "No space selected" and "Not connected" cases.

### Minor observations (P3)

1. **`@ts-nocheck` in test file** (line 1): This was pre-existing and not introduced by this PR, so not blocking — but worth noting that the test file suppresses all TypeScript checks.

No issues found. The refactoring is minimal, correct, and the tests verify the new behavior.

lsm (commented):

---

**Recommendation: APPROVE**

This is a clean, well-scoped refactor. The implementation is correct and the tests adequately cover the new behavior. Below are my findings.

---

### What Was Done Well

- The removal of `spaceId` from the props is correct. The store already tracks `spaceStore.spaceId.value` and exposes it through `createTask()`, so there is no information loss.
- The old `connectionManager.getHubIfConnected()` null-guard and the `hub.request` call are correctly replaced. The `spaceStore.createTask()` method internally checks for `!spaceId` (throws `'No space selected'`) and `!hub` (throws `'Not connected'`), so those failure paths are still covered — they now surface as thrown errors caught by the `catch` block in `handleSubmit`, and the new test `'shows error when no space is selected (createTask throws)'` verifies the no-space path explicitly.
- Removing the `if (!task) throw new Error('Server returned no data')` guard is safe. `hub.request<SpaceTask>()` is typed `Promise<TResult>` (not `Promise<TResult | null>`) and throws when not connected, so a null task would be a type-system lie, not a real runtime risk.
- Removing `spaceStore.selectSpace(spaceId).catch(() => {})` is the right call. The `space.task.created` event subscription in `startSubscriptions()` already appends the new task to `this.tasks.value`, making the full-refetch redundant.
- Tests now mock at the store boundary instead of reaching through to `connectionManager`, which is the correct level of abstraction.
- The new `'shows error when no space is selected (createTask throws)'` test replaces the deleted `'shows error when not connected'` test with an equivalent edge-case check.

---

**P3 — `SpaceIsland.tsx` still passes `spaceId` to `SpaceStartWorkflowDialog`**

`packages/web/src/islands/SpaceIsland.tsx` (line 313) passes `spaceId={spaceId}` to `SpaceStartWorkflowDialog` using the prop-based approach that was just removed from `SpaceCreateTaskDialog`. This is fine as-is (it's out of scope for this PR), but when `SpaceStartWorkflowDialog` is eventually refactored to use `spaceStore` similarly, the pattern established here will be the guide.

**P3 — `mockCreateTask.mockReset()` in `beforeEach` is sufficient; `afterEach` cleanup is redundant**

`packages/web/src/components/space/__tests__/SpaceCreateTaskDialog.test.tsx` calls `cleanup()` in both `beforeEach` and `afterEach`. Calling it in `beforeEach` alone is already a common defense against leftover renders; the `afterEach` call does no harm but is unnecessary.

**P3 — `mockToastError` is declared and reset but never actually asserted on in any test**

The `mockToastError` mock exists in the test file but no test asserts it was called or not called. This is a pre-existing pattern and harmless, but it adds noise.

---

### Summary

The refactor correctly delegates task creation to the store, removes the now-unnecessary `spaceId` prop, and replaces the manual hub wiring with a simpler, better-abstracted call. All 10 tests have been updated accordingly and cover the meaningful edge cases. No blocking issues.

PR #1098, opened 2026-03-28
feat(neo): add SessionType 'neo', DB schema, and NeoActivityLogRepository

lsm:
Implements Task 1.1 of the Neo Agent milestone.

- Adds `'neo'` to the `SessionType` union and `neoId?: string` to `SessionContext` in `packages/shared`
- Migration 66: expands `sessions.type` CHECK constraint to include `'neo'`; creates `neo_activity_log` table with an index on `created_at`
- `NeoActivityLogRepository` with `insert`, `list` (paginated, newest first), `getById`, `getLatestUndoable`
- Exposes repository from the `Database` facade via `db.neoActivityLog`
- 12 unit tests covering all CRUD paths, pagination, and undoable-entry lookup

### DIFF
--- a/packages/daemon/src/storage/index.ts
+++ b/packages/daemon/src/storage/index.ts
@@ -38,6 +38,7 @@ import { TaskRepository } from './repositories/task-repository';
 import { RoomMcpEnablementRepository } from './repositories/room-mcp-enablement-repository';
 import { SkillRepository } from './repositories/skill-repository';
 import { RoomSkillOverrideRepository } from './repositories/room-skill-override-repository';
+import { NeoActivityLogRepository } from './repositories/neo-activity-log-repository';
 import type { ReactiveDatabase } from './reactive-database';
 
 export type { SendStatus } from './repositories/sdk-message-repository';
@@ -63,6 +64,12 @@ export { AppMcpServerRepository } from './repositories/app-mcp-server-repository
 export { RoomMcpEnablementRepository } from './repositories/room-mcp-enablement-repository';
 export { SkillRepository } from './repositories/skill-repository';
 export { RoomSkillOverrideRepository } from './repositories/room-skill-override-repository';
+export { NeoActivityLogRepository } from './repositories/neo-activity-log-repository';
+export type {
+	NeoActivityLogEntry,
+	InsertNeoActivityParams,
+	ListNeoActivityParams,
+} from './repositories/neo-activity-log-repository';
 
 /**
  * Database facade class that maintains backward compatibility with the original Database class.
@@ -84,6 +91,7 @@ export class Database {
 	private roomMcpEnablementRepo!: RoomMcpEnablementRepository;
 	private skillRepo!: SkillRepository;
 	private roomSkillOverrideRepo!: RoomSkillOverrideRepository;
+	private neoActivityLogRepo!: NeoActivityLogRepository;
 	private shortIdAllocator!: ShortIdAllocator;
 
 	constructor(dbPath: string) {
@@ -109,6 +117,7 @@ export class Database {
 		this.roomMcpEnablementRepo = new RoomMcpEnablementRepository(db, reactiveDb);
 		this.skillRepo = new SkillRepository(db, reactiveDb);
 		this.roomSkillOverrideRepo = new RoomSkillOverrideRepository(db, reactiveDb);
+		this.neoActivityLogRepo = new NeoActivityLogRepository(db);
 	}
 
 	// ============================================================================
@@ -479,6 +488,13 @@ export class Database {
 	 */
 	get roomSkillOverrides(): RoomSkillOverrideRepository {
 		return this.roomSkillOverrideRepo;
+	}
+
+	/**
+	 * Get the Neo activity log repository
+	 */
+	get neoActivityLog(): NeoActivityLogRepository {
+		return this.neoActivityLogRepo;
 	}
 
 	close(): void {
--- a/packages/daemon/src/storage/repositories/neo-activity-log-repository.ts
+++ b/packages/daemon/src/storage/repositories/neo-activity-log-repository.ts
@@ -1,0 +1,174 @@
+/**
+ * NeoActivityLogRepository
+ *
+ * CRUD operations for the neo_activity_log table.
+ * Records every tool invocation made by the Neo global agent for auditing and undo support.
+ */
+
+import type { Database as BunDatabase } from 'bun:sqlite';
+
+// ---------------------------------------------------------------------------
+// Public types
+// ---------------------------------------------------------------------------
+
+export interface NeoActivityLogEntry {
+	id: string;
+	toolName: string;
+	input: string | null;
+	output: string | null;
+	status: 'success' | 'error' | 'cancelled';
+	error: string | null;
+	targetType: string | null;
+	targetId: string | null;
+	undoable: boolean;
+	undoData: string | null;
+	createdAt: string;
+}
+
+export interface InsertNeoActivityParams {
+	id: string;
+	toolName: string;
+	input?: string | null;
+	output?: string | null;
+	status?: 'success' | 'error' | 'cancelled';
+	error?: string | null;
+	targetType?: string | null;
+	targetId?: string | null;
+	undoable?: boolean;
+	undoData?: string | null;
+}
+
+export interface ListNeoActivityParams {
+	/** Number of entries to return (default: 50) */
+	limit?: number;
+	/** Cursor for pagination: return entries older than this created_at value */
+	before?: string;
+}
+
+// ---------------------------------------------------------------------------
+// Internal row type
+// ---------------------------------------------------------------------------
+
+interface ActivityRow {
+	id: string;
+	tool_name: string;
+	input: string | null;
+	output: string | null;
+	status: string;
+	error: string | null;
+	target_type: string | null;
+	target_id: string | null;
+	undoable: number;
+	undo_data: string | null;
+	created_at: string;
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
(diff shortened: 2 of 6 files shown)


lsm (commented):

Overall the implementation is solid: the migration pattern follows the probe-insert/table-recreate idiom established by M31, the repository is clean and well-structured, the DB schema looks correct, and the 12 unit tests cover the main paths well. There are a few issues worth addressing before merge.

---

### P1 — `SessionMetadata.sessionType` not updated

**File:** `packages/shared/src/types.ts`, lines 548–557

The `SessionMetadata` interface contains a **hardcoded inline union** for `sessionType` that duplicates `SessionType` but is a separate declaration — it does NOT use the `SessionType` alias. The PR correctly added `'neo'` to `SessionType` (line 78), but left the inline union in `SessionMetadata.sessionType` unchanged:

```ts
sessionType?:
  | 'room_chat'
  | 'planner'
  | 'coder'
  | 'leader'
  | 'general'
  | 'worker'
  | 'lobby'
  | 'spaces_global'
  | 'space_task_agent';  // <-- 'neo' missing here
```

Any code that tries to write `sessionType: 'neo'` into session metadata will fail TypeScript type-checking. This should either add `'neo'` to the inline union, or (better) change the field type to `SessionType` to stay DRY and automatically capture future additions.

---

### P2 — Cursor pagination on `created_at` string is not collision-safe

**File:** `packages/daemon/src/storage/repositories/neo-activity-log-repository.ts`, lines 130–138

The `list()` method uses `WHERE created_at < ?` for cursor-based pagination. `created_at` is set to `new Date().toISOString()` at insert time, which has millisecond resolution. If two entries are inserted within the same millisecond, they share an identical `created_at` value, and one of them can be silently dropped at a page boundary.

The industry-standard fix is a composite cursor: add `id` as a tiebreaker, e.g. `WHERE (created_at, id) < (?, ?)` (using SQLite row-value comparison), or equivalently `WHERE created_at < ? OR (created_at = ? AND id < ?)`.

For an audit log this is not critical in practice (Neo tools likely won't fire multiple times per millisecond), but the risk is real and worth addressing before the undo feature relies on this pagination.

---

### P2 — Migration 66 not exported from `schema/index.ts`

**File:** `packages/daemon/src/storage/schema/index.ts`

Migrations 47–58 are all individually exported with `// knip-ignore-next-line` comments for external re-use. Migration 66 is not exported. While there are no current callers that require it to be exported, the established pattern in this codebase is to export each significant migration so that tests and tooling can replay it selectively. The omission is inconsistent with the project convention.

---

### P3 — Minor: `SessionMetadata.sessionType` should reference the `SessionType` alias

As noted in the P1 issue, `SessionMetadata.sessionType` is a copy-paste of the union rather than `SessionType`. Even after adding `'neo'`, this field will continue to drift on future additions unless the type is changed to `sessionType?: SessionType`. This is a straightforward cleanup.

---

**Summary:** The P1 type inconsistency in `SessionMetadata.sessionType` will cause TypeScript errors if anyone tries to store `sessionType: 'neo'` in session metadata, which is the natural thing to do when creating a neo session. This should be fixed before merging.

lsm (commented):

Clean implementation that follows established patterns in the codebase. All 12 tests pass and typecheck is clean. A few observations:

### P2: Missing CHECK constraint on `neo_activity_log.status`

**File:** `packages/daemon/src/storage/schema/index.ts` (line ~405) and `packages/daemon/src/storage/schema/migrations.ts` (line ~4365)

The `status` column has no CHECK constraint, yet the TypeScript type restricts it to `'success' | 'error' | 'cancelled'`. Every other status column in this codebase uses a CHECK constraint for DB-level validation (e.g., `sessions.status`, `space_tasks.status`, `job_queue.status`). Adding `CHECK(status IN ('success', 'error', 'cancelled'))` would be consistent and provide defense-in-depth.

### P3: Cursor-based pagination on `created_at` is not collision-safe

**File:** `packages/daemon/src/storage/repositories/neo-activity-log-repository.ts` (line 129-138)

The `list()` pagination uses `WHERE created_at < ?` as the cursor, but `created_at` is not unique. If two entries share the same ISO timestamp (possible with rapid inserts), pagination could skip entries. A compound cursor like `(created_at, id)` or `(created_at, rowid)` would be more robust. Low risk for an activity log but worth noting for correctness.

### What looks good

- Migration 66 follows the established probe-insert + table-recreate pattern (identical structure to migration 31).
- The `createTables` schema and migration schema are in sync.
- The repository is append-only with no update/delete methods, which is appropriate for an audit log.
- Tests are thorough: all CRUD paths, defaults, pagination, undoable filtering, and edge cases are covered.
- Type changes (`SessionType` union, `SessionContext.neoId`) are minimal and correct.
- The `NeoActivityLogRepository` takes only `BunDatabase` (no `ReactiveDatabase`), which is appropriate since there are no live query subscriptions needed for this table yet.

**Recommendation: APPROVE** -- the P2 is a consistency improvement worth addressing but not blocking. The implementation is correct, complete, and well-tested.

lsm (commented):

**Recommendation: APPROVE**

All five issues from the previous reviews have been addressed:

1. **SessionMetadata.sessionType now uses the `SessionType` alias** (`packages/shared/src/types.ts:548`) -- no more copy-paste drift risk. Confirmed.
2. **CHECK constraint added to `neo_activity_log.status`** -- present in both the `createTables` schema (`packages/daemon/src/storage/schema/index.ts:408`) and migration 66 (`packages/daemon/src/storage/schema/migrations.ts:4366`). Consistent with every other status column in the codebase.
3. **Compound cursor pagination** (`packages/daemon/src/storage/repositories/neo-activity-log-repository.ts:138-150`) -- now uses `(created_at, id)` pair with the standard `WHERE created_at < ? OR (created_at = ? AND id < ?)` pattern. Correct.
4. **`runMigration66` exported from `schema/index.ts`** (line 37) with `knip-ignore` comment, following established convention.
5. **New test for compound cursor correctness** (`packages/daemon/tests/unit/neo-activity-log-repository.test.ts:156-182`) -- verifies that same-millisecond entries are not dropped during pagination. All 13 tests pass.

### One remaining nit

**P3: JSDoc on `SessionType` missing `'neo'` description** (`packages/shared/src/types.ts:57-67`). The doc comment lists all session types with descriptions but omits `'neo'`. Adding a line like `* - 'neo': Global Neo AI agent session` would keep the documentation complete.

### What looks good

- Migration 66 follows the probe-insert + table-recreate pattern identical to migration 31.
- Schema in `createTables` and migration are fully in sync (table DDL, CHECK constraint, index).
- Repository is append-only (no update/delete), appropriate for an audit log.
- `InsertNeoActivityParams` has sensible defaults, and `undoable` correctly maps between boolean and INTEGER.
- Test coverage is thorough: CRUD, defaults, error status, pagination, undoable filtering, empty DB, and the new same-millisecond cursor edge case.

lsm (commented):

**Recommendation: APPROVE** — all previously flagged P1 issues are resolved. Two residual findings below are both P2/P3.

---

### Previously flagged issues — status

| # | Issue | Status |
|---|-------|--------|
| P1 | `SessionMetadata.sessionType` was an inline copy missing `'neo'` | Fixed — now uses `SessionType` alias |
| P1 | No `CHECK` constraint on `neo_activity_log.status` | Fixed — added in both schema and migration |
| P1 | Cursor pagination could drop entries with same-millisecond timestamps | Fixed — compound `(created_at, id)` cursor implemented |
| P2 | `runMigration66` not exported / not knip-safe | Fixed — exported with `// knip-ignore-next-line` |
| P1 | No test for compound cursor pagination | Fixed — test added |

---

### Residual findings

#### P2 — Migration 66 table-recreate drops `idx_sessions_parent` and never re-creates it

`packages/daemon/src/storage/schema/migrations.ts`, `runMigration66` (~line 4317–4355)

Migration 11 (`runMigration11`, line 544) creates `idx_sessions_parent ON sessions(parent_id)`. When migration 66's table-recreate path fires on an existing database (i.e. `'neo'` not yet in the CHECK), it does `DROP TABLE sessions` which silently removes `idx_sessions_parent` as well. The replacement `sessions_new` table gets no index on `parent_id`, and `createIndexes` is only called once at startup (not during migrations). This leaves existing databases missing the index after migration 66 runs.

All prior migrations using this same table-recreate pattern (M9, M13, etc.) were written *before* migration 11 created the index, so they were safe. Migration 66 is the first to post-date M11 and suffers from the omission.

Fix: add `CREATE INDEX IF NOT EXISTS idx_sessions_parent ON sessions(parent_id)` at the end of the `catch` block, after the rename.

#### P2 — Compound cursor same-millisecond test doesn't verify zero data loss across pages

`packages/daemon/tests/unit/neo-activity-log-repository.test.ts`, line 156–182

The test inserts `log-c` (newer), `log-a` and `log-b` (same timestamp), then fetches page 1 with `limit: 3` — which returns all three in one shot. It only asserts that the cursor entry isn't duplicated on page 2, but never asserts that both `log-a` and `log-b` appear in page 1 positions 1 and 2. More importantly, the case where the page boundary falls *between* two same-timestamp entries is never exercised: a `limit: 1` first page (returning `log-c`) followed by a `limit: 2` second page (which must include both `log-a` and `log-b`) would be a stronger proof.

The implementation itself is correct; this is a coverage gap, not a bug.

#### P3 — `getLatestUndoable` lacks `id DESC` tiebreaker

`packages/daemon/src/storage/repositories/neo-activity-log-repository.ts`, line 181

The query uses `ORDER BY created_at DESC LIMIT 1` while `list()` uses `ORDER BY created_at DESC, id DESC`. For `LIMIT 1` this is harmless in practice (both will return *some* most-recent undoable entry), but it creates a subtle inconsistency: the entry `getLatestUndoable` returns may not match what `list()` considers the definitive ordering when two undoable entries share a timestamp.

PR #1099, opened 2026-03-28
test(e2e): add space settings CRUD E2E + verify SpaceSettings unit tests

lsm:
The `SpaceSettings` component already had full edit/archive/delete functionality (landed in #1090). This PR adds the missing E2E test coverage.

**Changes:**
- `packages/e2e/tests/features/space-settings-crud.e2e.ts` — new E2E test covering:
  - Settings tab renders name, description, workspace path, danger zone buttons
  - Save Changes button only appears when form is dirty
  - Editing name and saving shows success toast and clears dirty state
  - Discard reverts unsaved edits
  - Archive shows browser confirm, archives space, redirects to `/spaces`
  - Delete shows browser confirm, deletes space, redirects to `/spaces`

Unit tests (17 passing) were already in place via `SpaceSettings.test.tsx`.

### DIFF
--- a/packages/e2e/tests/features/space-settings-crud.e2e.ts
+++ b/packages/e2e/tests/features/space-settings-crud.e2e.ts
@@ -22,18 +22,20 @@ async function createSpaceViaRpc(
 	workspacePath: string,
 	name: string
 ): Promise<string> {
-	const spaceId = await page.evaluate(
+	const id = await page.evaluate(
 		async ({ workspacePath, name }) => {
 			const hub = window.__messageHub || window.appState?.messageHub;
 			if (!hub?.request) throw new Error('MessageHub not available');
-			const result = (await hub.request('space.create', { workspacePath, name })) as {
-				space: { id: string };
+			// space.create returns the Space object directly (not wrapped in { space: ... })
+			const space = (await hub.request('space.create', { workspacePath, name })) as {
+				id: string;
 			};
-			return result.space.id;
+			return space.id;
 		},
 		{ workspacePath, name }
 	);
-	return spaceId;
+	if (!id) throw new Error('space.create returned no id');
+	return id;
 }
 
 async function deleteSpaceViaRpc(
@@ -113,7 +115,6 @@ test.describe('Space Settings CRUD', () => {
 		const nameInput = page.locator('input[type="text"]').first();
 		await nameInput.fill(newName);
 
-		// Accept the browser confirm dialog that may appear (archive/delete), and just save
 		await page.getByRole('button', { name: 'Save Changes' }).click();
 
 		// Toast notification of success
@@ -151,10 +152,6 @@ test.describe('Space Settings CRUD', () => {
 
 		// Should redirect to /spaces
 		await page.waitForURL('/spaces', { timeout: 10000 });
-		await expect(page).toHaveURL('/spaces');
-
-		// Space is now archived — no need for RPC cleanup in afterEach since it's not deleted
-		// Re-delete via RPC in afterEach still handles it
 	});
 
 	test('Delete space shows confirm dialog and redirects to spaces list', async ({ page }) => {
@@ -165,7 +162,6 @@ test.describe('Space Settings CRUD', () => {
 
 		// Should redirect to /spaces
 		await page.waitForURL('/spaces', { timeout: 10000 });
-		await expect(page).toHaveURL('/spaces');
 
 		// Space is already deleted — don't double-delete in afterEach
 		spaceId = '';
--- a/packages/e2e/tests/features/space-settings-crud.e2e.ts
+++ b/packages/e2e/tests/features/space-settings-crud.e2e.ts
@@ -1,0 +1,173 @@
+/**
+ * Space Settings CRUD E2E Tests
+ *
+ * Verifies:
+ * - Navigating to the Settings tab of a space
+ * - Editing space name and saving persists changes
+ * - Discarding reverts edits without saving
+ * - Archiving a space redirects to the spaces list
+ * - Deleting a space redirects to the spaces list
+ *
+ * Setup: creates a space via RPC (infrastructure), navigates to its Settings tab
+ * Cleanup: deletes the space via RPC in afterEach (infrastructure)
+ */
+
+import { test, expect } from '../../fixtures';
+import { waitForWebSocketConnected, getWorkspaceRoot } from '../helpers/wait-helpers';
+
+const DESKTOP_VIEWPORT = { width: 1280, height: 720 };
+
+async function createSpaceViaRpc(
+	page: Parameters<typeof waitForWebSocketConnected>[0],
+	workspacePath: string,
+	name: string
+): Promise<string> {
+	const spaceId = await page.evaluate(
+		async ({ workspacePath, name }) => {
+			const hub = window.__messageHub || window.appState?.messageHub;
+			if (!hub?.request) throw new Error('MessageHub not available');
+			const result = (await hub.request('space.create', { workspacePath, name })) as {
+				space: { id: string };
+			};
+			return result.space.id;
+		},
+		{ workspacePath, name }
+	);
+	return spaceId;
+}
+
+async function deleteSpaceViaRpc(
+	page: Parameters<typeof waitForWebSocketConnected>[0],
+	spaceId: string
+): Promise<void> {
+	if (!spaceId) return;
+	try {
+		await page.evaluate(async (id) => {
+			const hub = window.__messageHub || window.appState?.messageHub;
+			if (!hub?.request) return;
+			await hub.request('space.delete', { id });
+		}, spaceId);
+	} catch {
+		// Best-effort cleanup
+	}
+}
+
+test.describe('Space Settings CRUD', () => {
+	test.use({ viewport: DESKTOP_VIEWPORT });
+
+	let spaceId = '';
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/');
+		await waitForWebSocketConnected(page);
+
+		const workspaceRoot = await getWorkspaceRoot(page);
+		const uniqueName = `E2E Settings Test ${Date.now()}`;
+		spaceId = await createSpaceViaRpc(page, workspaceRoot, uniqueName);
+
+		// Navigate to the space
+		await page.goto(`/space/${spaceId}`);
+		await page.waitForURL(`/space/${spaceId}`, { timeout: 10000 });
+
+		// Click the Settings tab
+		await page.getByRole('button', { name: 'Settings', exact: true }).click();
+		await expect(page.locator('text=Danger Zone')).toBeVisible({ timeout: 5000 });
+	});
+
+	test.afterEach(async ({ page }) => {
+		if (spaceId) {
+			await deleteSpaceViaRpc(page, spaceId);
+			spaceId = '';
+		}
+	});
+
+	test('renders settings tab with name, description, workspace path', async ({ page }) => {
+		// Name input should be visible and contain the space name
+		await expect(page.locator('input[type="text"]').first()).toBeVisible();
+
+		// Workspace path should be shown as read-only text
+		const workspaceRoot = await getWorkspaceRoot(page);
+		await expect(page.locator(`text=${workspaceRoot}`)).toBeVisible({ timeout: 3000 });
+
+		// Danger Zone section
+		await expect(page.locator('text=Danger Zone')).toBeVisible();
+		await expect(page.getByRole('button', { name: 'Archive', exact: true })).toBeVisible();
+		await expect(page.getByRole('button', { name: 'Delete', exact: true })).toBeVisible();
+	});
+
+	test('shows Save Changes button only when form is dirty', async ({ page }) => {
+		// Save Changes not visible initially
+		await expect(page.getByRole('button', { name: 'Save Changes' })).not.toBeVisible();
+
+		// Type something in the name field
+		const nameInput = page.locator('input[type="text"]').first();
+		await nameInput.fill('New Space Name');
+
+		// Save Changes should now appear
+		await expect(page.getByRole('button', { name: 'Save Changes' })).toBeVisible({ timeout: 2000 });
+	});
+
+	test('edits space name, saves, and persists the change', async ({ page }) => {
+		const newName = `Renamed Space ${Date.now()}`;
+
+		const nameInput = page.locator('input[type="text"]').first();
+		await nameInput.fill(newName);
+
+		// Accept the browser confirm dialog that may appear (archive/delete), and just save
+		await page.getByRole('button', { name: 'Save Changes' }).click();
+
+		// Toast notification of success
+		await expect(page.locator('text=Space updated')).toBeVisible({ timeout: 5000 });
+
+		// The input should now reflect the saved name
+		await expect(nameInput).toHaveValue(newName, { timeout: 3000 });
+
+		// Save Changes button should disappear (form is clean again)
+		await expect(page.getByRole('button', { name: 'Save Changes' })).not.toBeVisible({
+			timeout: 3000,
+		});
+	});
+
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This PR adds a well-structured E2E test file covering the SpaceSettings CRUD operations. The underlying component (`SpaceSettings.tsx`) and its unit tests (`SpaceSettings.test.tsx`, 17 tests) were already landed in #1090. This PR fills the missing E2E gap.

### What looks good

- **Follows established patterns**: The test structure mirrors `space-creation.e2e.ts` closely -- same viewport setup, same RPC-based `createSpaceViaRpc`/`deleteSpaceViaRpc` helpers for setup/teardown, same `waitForWebSocketConnected` + `getWorkspaceRoot` helpers.
- **Correct use of dialog handlers**: `page.on('dialog', ...)` is registered before the click that triggers `window.confirm()`, which is the correct Playwright pattern.
- **Proper cleanup**: The delete test sets `spaceId = ''` to avoid double-delete in `afterEach`. The archive test leaves `spaceId` set so `afterEach` will delete the (now archived) space, which works because `spaceRepo.deleteSpace` doesn't care about status.
- **UI-driven assertions**: All test actions go through the UI (clicking buttons, filling inputs, checking visibility), with RPC used only for setup/teardown -- consistent with the E2E rules in CLAUDE.md.
- **Dirty-state tests**: Good coverage of the `isDirty` conditional rendering (Save Changes appearing/disappearing, Discard reverting).

### Minor observations (P3)

1. **`packages/e2e/tests/features/space-settings-crud.e2e.ts:86-90`** -- The "renders settings tab" test uses `page.locator('input[type="text"]').first()` which is fragile if more text inputs are added. The existing unit tests use `getByDisplayValue` which is more specific. However, this is consistent with how other E2E tests in this repo use locators, so not a real concern.

2. **`packages/e2e/tests/features/space-settings-crud.e2e.ts:116`** -- The comment "Accept the browser confirm dialog that may appear (archive/delete)" on the save test is misleading since save doesn't trigger a confirm dialog. It's harmless dead commentary.

3. **No test for the "decline confirm" path** -- The unit tests cover the case where `confirm()` returns false (lines 204-209, 234-239 in the unit test), but the E2E tests only cover the "accept" path. This is acceptable since the confirm dialog behavior is a browser primitive and the unit tests already cover the dismiss case.

No blocking issues found. The test is clean, follows project conventions, and provides meaningful E2E coverage for the space settings CRUD flow.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two P0 bugs will cause all tests in this suite to fail at runtime.

---

### P0 — Wrong return-shape assumption in `createSpaceViaRpc`

**File:** `packages/e2e/tests/features/space-settings-crud.e2e.ts`, lines 29-33

```ts
const result = (await hub.request('space.create', { workspacePath, name })) as {
  space: { id: string };
};
return result.space.id;
```

The `space.create` RPC handler (`packages/daemon/src/lib/rpc-handlers/space-handlers.ts`, line 100) returns the `Space` object directly (`return space;`), not a wrapped `{ space: { id } }` envelope. Every other E2E test in the codebase casts it correctly as `{ id: string }`:

```ts
// space-export-import.e2e.ts line 57 — correct pattern
const spaceId = (spaceRes as { id: string }).id;
```

Because of this bug `result.space` is `undefined`, so `createSpaceViaRpc` returns `undefined` and `spaceId` is set to `undefined` in every `beforeEach`. Consequently:
- `page.goto('/space/undefined')` navigates to a broken URL.
- The `waitForURL` on line 70 times out.
- The space is never cleaned up (the `if (spaceId)` guard in `afterEach` is falsy since the variable holds `''`... actually `spaceId` is declared as `let spaceId = ''` but the return of `createSpaceViaRpc` is `undefined`, so it becomes `undefined`).

Every test in this file will fail at the `beforeEach` step.

**Fix:** Change line 29-33 to:
```ts
const spaceId = (await hub.request('space.create', { workspacePath, name })) as { id: string };
return spaceId.id;
```

---

### P0 — `getByRole('button', { name: 'Settings' })` matches nothing

**File:** `packages/e2e/tests/features/space-settings-crud.e2e.ts`, line 73

```ts
await page.getByRole('button', { name: 'Settings', exact: true }).click();
```

The Settings nav item in `SpaceNavPanel` is rendered as a `div`, not a `<button>`:

```tsx
// packages/web/src/components/space/SpaceNavPanel.tsx, lines 213-223
<div
  onClick={onClick}
  class={cn(...)}
>
  <Icon />
  <span class="text-xs font-medium">{label}</span>
</div>
```

Playwright's `getByRole('button')` will not match a `div` (divs have no implicit ARIA role of `button` unless `role="button"` is explicitly set). Even if the first P0 is fixed and navigation to the space page succeeds, the `beforeEach` will still time out on this click because no `button` with the name "Settings" exists.

**Fix:** Use a text-based locator or a more general click target:
```ts
await page.getByText('Settings').click();
```
or scope it to the nav panel element to avoid ambiguity with other "Settings" text on the page.

---

### P1 — Archive test leaks state if `space.delete` can fail on archived spaces

**File:** `packages/e2e/tests/features/space-settings-crud.e2e.ts`, lines 146-158

After the archive test succeeds, the space remains in an archived state. The `afterEach` calls `deleteSpaceViaRpc(page, spaceId)` to clean up, which calls `space.delete`. The RPC handler for `space.delete` calls `spaceManager.deleteSpace` — this should work regardless of archive status based on the handler code, so cleanup will succeed. However, the comment on line 156-157 ("Space is now archived — no need for RPC cleanup in afterEach since it's not deleted") is misleading and contradicts the actual `afterEach` behavior which does call `deleteSpaceViaRpc`. The comment should be removed or corrected to avoid confusion.

---

### P2 — `spaceId` variable declared at `let spaceId = ''` but `createSpaceViaRpc` can return `undefined`

**File:** `packages/e2e/tests/features/space-settings-crud.e2e.ts`, line 58

```ts
let spaceId = '';
```

The TypeScript inferred type is `string`, but `createSpaceViaRpc` returns a `Promise<string>` which actually resolves to `undefined` at runtime (see P0 above). Even after fixing the P0, the return type of `createSpaceViaRpc` should be `Promise<string>` and should throw on falsy result (similar to how `getWorkspaceRoot` throws on line 41-43 of `wait-helpers.ts`), rather than silently propagating an empty/undefined ID.

---

### P3 — Duplicate `await expect(page).toHaveURL('/spaces')` assertion

**File:** `packages/e2e/tests/features/space-settings-crud.e2e.ts`, lines 153-154 and 167-168

`waitForURL` already ensures the URL matches; the subsequent `toHaveURL` is redundant. This is a minor nit, not a blocking issue.

---

**Summary:** The test file has two show-stopping bugs: the `space.create` result shape is cast incorrectly (returns a bare `Space`, not `{ space: Space }`), and the Settings nav item is a `div` not a `button`, so `getByRole('button', { name: 'Settings' })` never resolves. Both bugs must be fixed before any test in this suite can run.

lsm (commented):

**Recommendation: APPROVE**

The two P0 bugs from the previous review have been correctly fixed. The code is in good shape. Below are the remaining observations.

---

### What was fixed (verified)

**`createSpaceViaRpc` return shape** — The cast is now `as { id: string }` directly, matching the `space.create` handler which returns `space` (a bare `Space` object) at `packages/daemon/src/lib/rpc-handlers/space-handlers.ts:100`. The `.id` access is correct and the explicit `if (!id) throw` guard on line 37 ensures a clear failure message if the response is unexpected.

**Stale comments removed** — The "confirm dialog" stale comment is gone from the save test. The misleading archive cleanup comment is gone. The redundant `toHaveURL` assertions after `waitForURL` were removed. All verified.

---

### Remaining observations

**P2 — `renders settings tab` test does not assert the name input value**

`packages/e2e/tests/features/space-settings-crud.e2e.ts:88` checks that `input[type="text"]` is visible, but does not assert it contains the space name created in `beforeEach`. This is the primary purpose of the test (verifying the settings form is populated correctly on load), but the assertion is missing. A `toHaveValue(uniqueName)` check here would make the test meaningful.

**P2 — Archive test leaves a stale space on server after redirect**

`packages/e2e/tests/features/space-settings-crud.e2e.ts:147-155` — The archive test navigates to `/spaces` and ends. However, `spaceId` is **not** cleared after the archive test completes, so `afterEach` will call `deleteSpaceViaRpc` on an already-archived space. This is fine from a cleanup perspective — `space.delete` should still succeed on an archived space — but it is a subtle asymmetry with the delete test (which explicitly sets `spaceId = ''`). A comment explaining the intentional non-clearance (or clearing it to avoid the extra RPC round-trip) would improve clarity.

**P3 — `getByRole('button', { name: 'Settings' })` selector ambiguity**

`packages/e2e/tests/features/space-settings-crud.e2e.ts:75` — The selector is `exact: true` which is correct, but the `Settings` button is a plain `<button>` in the tab bar, not a role-tagged element. If other buttons on the page text-match `Settings` (e.g., from a `<title>` or SVG inside a button), the locator could match unexpectedly. Using a `data-testid` on the tab bar buttons (consistent with how `editor-mode-list` / `editor-mode-visual` tabs use `data-testid`) would be more robust. This is a minor nit given the exact match works today.

**P3 — `waitForWebSocketConnected` type signature mismatch**

`packages/e2e/tests/features/space-settings-crud.e2e.ts:21` — The `page` parameter is typed as `Parameters<typeof waitForWebSocketConnected>[0]`, which is `Page` from `@playwright/test`. This is correct and consistent with other test files (e.g., `space-creation.e2e.ts:22`), but it is unnecessarily verbose when the function could simply accept `Page` directly. Minor readability issue only.

---

### Summary

The two P0 blocking bugs from the prior review are fully addressed. The response cast is now correct and the guard is in place. The stale comments are gone. The remaining issues are minor: a missing value assertion in the render test (P2) and an asymmetric post-archive cleanup (P2). Neither blocks the PR. Tests cover all the claimed scenarios and integrate correctly with the established E2E patterns in this repo.

lsm (commented):

**Recommendation: APPROVE**

The fixes from the second commit correctly address the issues raised in the previous review. I verified all three changes against the source code.

### Verified fixes

1. **`createSpaceViaRpc` return shape (line 29-33)** -- Now correctly casts the response as `{ id: string }` rather than `{ space: { id: string } }`. This matches the `space.create` RPC handler in `packages/daemon/src/lib/rpc-handlers/space-handlers.ts` (line 100: `return space;`), and is consistent with every other E2E test in the codebase (`space-export-import.e2e.ts`, `reviewer-feedback-loop.e2e.ts`, etc.).

2. **Guard on missing id (line 37)** -- `if (!id) throw new Error('space.create returned no id');` provides a clear early failure instead of silently propagating `undefined` through the rest of the test.

3. **Comment and assertion cleanup** -- Removed the stale comment about confirm dialogs from the save test (save never triggers `window.confirm`), and removed redundant `toHaveURL` assertions after `waitForURL` in the archive and delete tests. Clean and correct.

### Note on the previous GLM review's "Settings button" P0

The GLM review claimed `getByRole('button', { name: 'Settings', exact: true })` would fail because the Settings nav link in `SpaceNavPanel` renders as a `<div>`. This was incorrect. The test navigates to `/space/${spaceId}`, which renders `SpaceIsland` (via `MainContent.tsx` line 69). `SpaceIsland` has its own tab bar (line 167-177) that renders each tab as a `<button>` element. The `getByRole('button')` locator correctly matches the tab button in the `SpaceIsland` tab bar, not the `SpaceNavPanel` nav link.

1. **`packages/e2e/tests/features/space-settings-crud.e2e.ts:87`** -- `page.locator('input[type="text"]').first()` is fragile if additional text inputs are added to the settings form. Using `getByLabel` or `getByPlaceholder` would be more resilient. Low risk since the unit tests cover the component in detail.

2. **`packages/e2e/tests/features/space-settings-crud.e2e.ts:20-39`** -- The `createSpaceViaRpc` helper duplicates logic found in `packages/e2e/tests/helpers/workflow-editor-helpers.ts`. A shared helper could reduce duplication across space E2E tests. Not blocking since the duplication is small and contained.

No blocking issues. The test file correctly tests the SpaceSettings CRUD flow through the UI, uses RPC only for setup/teardown per project conventions, and all assertions target visible DOM state.

lsm (commented):

**Recommendation: APPROVE**

Both P2 issues from the previous GLM review have been correctly addressed in the latest commit:

### Fix 1 -- Name input value assertion (P2 resolved)

`packages/e2e/tests/features/space-settings-crud.e2e.ts:61,68,91` -- `spaceName` is now declared at the `describe` scope (line 61), assigned in `beforeEach` (line 68), and the "renders settings tab" test asserts `toHaveValue(spaceName)` (line 91). This makes the render test meaningfully verify that the settings form loads with the correct space name.

### Fix 2 -- Archive cleanup comment (P2 resolved)

`packages/e2e/tests/features/space-settings-crud.e2e.ts:158-159` -- Clear comment explains the intentional asymmetry: `spaceId` is left set after archive so `afterEach` cleans it up via `space.delete`, which works on archived spaces. This removes the ambiguity flagged in the prior review.

### Previously noted P3 items (unchanged, non-blocking)

- `input[type="text"]` selector (line 89) is fragile if more text inputs are added; `getByLabel`/`getByPlaceholder` would be more resilient.
- `createSpaceViaRpc` helper duplicates logic from `workflow-editor-helpers.ts`; could be shared.

No blocking issues remain.

lsm (commented):

**Recommendation: APPROVE** — The two P2 fixes from the previous round are correctly applied. The remaining observations below are minor.

---

### What was fixed (confirmed)

**Fix 1 — `spaceName` hoisted to describe scope (line 61)**
`spaceName` is now declared at `test.describe` scope alongside `spaceId`. `beforeEach` sets it before navigating. The "renders" test at line 91 asserts `toHaveValue(spaceName)` — this is a meaningful assertion that confirms the Settings panel pre-populates from the space that was just created, not a stale or empty value. Correct.

**Fix 2 — Archive cleanup comment (lines 158-159)**
The comment `// spaceId is intentionally left set — afterEach will delete the (archived) space via space.delete, which succeeds regardless of archive status` is clear and accurate. `space-manager.ts` `deleteSpace()` does a simple `getSpace` check then `deleteSpace` — it does not gate on `status`, so deleting an archived space works. The comment is justified.

---

### Remaining observations

**P3 — `getWorkspaceRoot` called twice in the "renders" test**
In the "renders" test (line 94), `getWorkspaceRoot` issues a fresh `state.system` RPC call to fetch `workspaceRoot` for the path assertion. The same value was already fetched inside `beforeEach` (line 67) but is not stored at describe scope, so the test makes a redundant round-trip. This is harmless — the path is static — but it is a minor efficiency and readability nit. Storing `workspaceRoot` at describe scope alongside `spaceName` would eliminate the duplication.

**P3 — `input[type="text"]` locator is positional, not semantic**
Every test that manipulates the name field uses `page.locator('input[type="text"]').first()`. The SpaceSettings form has only one `<input type="text">` today so `.first()` resolves correctly, but this will silently target the wrong element if another text input is ever added above the name field. A locator tied to the `Name` label (e.g., `page.getByLabel('Name')`) would be more resilient. Low risk given the component's current structure.

**P3 — `deleteSpaceViaRpc` duplicates code from `space-creation.e2e.ts`**
`deleteSpaceByRpc` (space-creation) and `deleteSpaceViaRpc` (this file) are character-for-character equivalent. This is an accepted pattern for E2E test isolation but if a shared helper ever lives in `room-helpers.ts` or a new `space-helpers.ts`, consolidating would reduce duplication.

**P3 — Archive test does not assert the archived state in the spaces list**
After archiving, the test only checks for URL redirect to `/spaces`. It does not verify that the space is absent from the list (which is the user-visible consequence of archiving). A `not.toContainText(spaceName)` assertion on the spaces list would make the test more meaningful. This is a coverage nit, not a bug.

---

No P0, P1, or P2 issues found. The previous round's feedback is fully addressed.

PR #1101, opened 2026-03-28
feat(web): add WorkflowRunStartDialog component

lsm:
Adds `WorkflowRunStartDialog` — a dashboard quick-action dialog for starting a workflow run.

**Key differences from `SpaceStartWorkflowDialog`:**
- Reads workflows from `spaceStore.workflows` signal (no prop drilling)
- Title is **optional**: auto-generated as `<WorkflowName> — <ISO timestamp>` when left blank
- Empty state includes a **"Go to Workflows tab"** button that triggers `onSwitchToWorkflows` and closes the dialog
- Delegates to `spaceStore.startWorkflowRun()` instead of calling the hub directly

**Changes:**
- `WorkflowRunStartDialog.tsx` — new component
- `__tests__/WorkflowRunStartDialog.test.tsx` — 13 unit tests (all scenarios)
- `space/index.ts` — barrel export
- `SpaceIsland.tsx` — wires the dialog to the "Start Workflow Run" quick action with `onSwitchToWorkflows → setActiveTab('workflows')`

### DIFF
--- a/packages/web/src/components/space/WorkflowRunStartDialog.tsx
+++ b/packages/web/src/components/space/WorkflowRunStartDialog.tsx
@@ -1,0 +1,172 @@
+/**
+ * WorkflowRunStartDialog — modal form to start a workflow run from the dashboard.
+ * Reads available workflows from spaceStore and delegates to spaceStore.startWorkflowRun().
+ */
+
+import { useState, useEffect } from 'preact/hooks';
+import { Modal } from '../ui/Modal';
+import { Button } from '../ui/Button';
+import { spaceStore } from '../../lib/space-store';
+import { toast } from '../../lib/toast';
+import type { SpaceWorkflowRun } from '@neokai/shared';
+
+interface WorkflowRunStartDialogProps {
+	isOpen: boolean;
+	onClose: () => void;
+	onSwitchToWorkflows?: () => void;
+	onStarted?: (run: SpaceWorkflowRun) => void;
+}
+
+function buildAutoTitle(workflowName: string): string {
+	const now = new Date();
+	const stamp = now.toISOString().slice(0, 16).replace('T', ' ');
+	return `${workflowName} — ${stamp}`;
+}
+
+export function WorkflowRunStartDialog({
+	isOpen,
+	onClose,
+	onSwitchToWorkflows,
+	onStarted,
+}: WorkflowRunStartDialogProps) {
+	const workflows = spaceStore.workflows.value;
+	const [workflowId, setWorkflowId] = useState<string>(workflows[0]?.id ?? '');
+	const [title, setTitle] = useState('');
+	const [submitting, setSubmitting] = useState(false);
+	const [error, setError] = useState<string | null>(null);
+
+	// Keep workflowId in sync when workflows load asynchronously
+	useEffect(() => {
+		if (!workflowId && workflows.length > 0) {
+			setWorkflowId(workflows[0].id);
+		}
+	}, [workflows, workflowId]);
+
+	const handleClose = () => {
+		setTitle('');
+		setWorkflowId(workflows[0]?.id ?? '');
+		setError(null);
+		onClose();
+	};
+
+	const handleSubmit = async (e: Event) => {
+		e.preventDefault();
+
+		if (!workflowId) {
+			setError('Please select a workflow');
+			return;
+		}
+
+		const selectedWorkflow = workflows.find((w) => w.id === workflowId);
+		const resolvedTitle = title.trim() || buildAutoTitle(selectedWorkflow?.name ?? 'Run');
+
+		try {
+			setSubmitting(true);
+			setError(null);
+
+			const run = await spaceStore.startWorkflowRun({
+				workflowId,
+				title: resolvedTitle,
+			});
+
+			toast.success(`Workflow run "${run.title}" started`);
+			onStarted?.(run);
+			handleClose();
+		} catch (err) {
+			setError(err instanceof Error ? err.message : 'Failed to start workflow run');
+		} finally {
+			setSubmitting(false);
+		}
+	};
+
+	const hasWorkflows = workflows.length > 0;
+
+	return (
+		<Modal isOpen={isOpen} onClose={handleClose} title="Start Workflow Run" size="md">
+			<form onSubmit={handleSubmit} class="space-y-4">
+				{error && (
+					<div class="bg-red-900/20 border border-red-800 rounded-lg px-4 py-3 text-red-400 text-sm">
+						{error}
+					</div>
+				)}
+
+				{!hasWorkflows && (
+					<div class="bg-amber-900/20 border border-amber-800 rounded-lg px-4 py-3 text-amber-300 text-sm space-y-2">
+						<p>No workflows available. Create one in the Workflows tab.</p>
+						{onSwitchToWorkflows && (
+							<button
+								type="button"
+								onClick={() => {
+									handleClose();
+									onSwitchToWorkflows();
+								}}
+								class="text-blue-400 hover:text-blue-300 underline text-xs"
+							>
+								Go to Workflows tab
+							</button>
+						)}
+					</div>
+				)}
+
+				{/* Workflow selector — shown only when multiple workflows exist */}
+				{workflows.length > 1 && (
+					<div>
+						<label class="block text-sm font-medium text-gray-200 mb-1.5">Workflow</label>
+						<select
+							value={workflowId}
+							onChange={(e) => setWorkflowId((e.target as HTMLSelectElement).value)}
+							class="w-full bg-dark-800 border border-dark-700 rounded-lg px-3 py-2 text-gray-100
+								focus:outline-none focus:border-blue-500 text-sm"
+						>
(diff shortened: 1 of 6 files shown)


lsm (commented):

**Recommendation: APPROVE** (with a few minor observations worth a follow-up)

---

### Summary

The implementation is correct, well-structured, and covers all the stated requirements. The component reads from `spaceStore.workflows` (no prop drilling), auto-generates titles, handles the empty state with a tab-switch CTA, and delegates to `spaceStore.startWorkflowRun()`. Integration in `SpaceIsland` is clean. The 13 unit tests cover the important cases. Below are observations ranging from a real edge case to minor style notes.

---

### P1 — `startWorkflowRun` can return `undefined` at runtime; component will crash

`spaceStore.startWorkflowRun()` destructures `{ run }` directly from `hub.request<{ run: SpaceWorkflowRun }>(...)`. TypeScript types it as non-null, but `hub.request` can return `null` when the server sends an empty response — the old `SpaceStartWorkflowDialog` guarded against this explicitly:

```ts
// SpaceStartWorkflowDialog.tsx line 79
if (!run) {
  throw new Error('Server returned no data');
}
```

The new `WorkflowRunStartDialog` calls `run.title` on line 72 unconditionally. If the store's `startWorkflowRun` returns `undefined` (null-propagated from the hub), this will throw `TypeError: Cannot read properties of undefined (reading 'title')` — caught by the catch block, so the dialog won't crash, but the user sees a confusing error instead of the toast. The guard belongs in the store method, not the component, but it's absent in both. The test mock always returns a resolved value, so tests don't catch this.

**File:** `packages/web/src/lib/space-store.ts` line 851-855

---

### P2 — `workflows.value` read outside a reactive context will not subscribe

`WorkflowRunStartDialog.tsx` line 32:
```ts
const workflows = spaceStore.workflows.value;
```

This reads the signal value once at render time. In Preact Signals, reading `.value` inside a component function body does establish a subscription, so reactive updates will re-render the component correctly — **but only when the component is mounted inside a Preact component tree**. The test file (line 1) has `// @ts-nocheck` and mocks the signal directly; however, because the mock uses a real `signal([])` instance, the reactive tracking still works in tests. This is fine as-is, just worth noting that the `// @ts-nocheck` suppresses all TypeScript validation in the test file, including any future type regressions.

---

### P2 — `@ts-nocheck` in the test file disables all type checking

`packages/web/src/components/space/__tests__/WorkflowRunStartDialog.test.tsx` line 1:
```ts
// @ts-nocheck
```

This disables TypeScript for the entire test file. Any future refactors that break the component's prop types or the mock interface would silently compile without error. The tests would still run, but type errors in the test harness won't be caught by `bun run typecheck`. Prefer a targeted `// @ts-expect-error` or fixing the underlying type issue.

---

### P2 — `onStarted` prop is not wired in `SpaceIsland`

`WorkflowRunStartDialog` accepts an optional `onStarted?: (run: SpaceWorkflowRun) => void`. In `SpaceIsland.tsx` lines 315-319, the dialog is rendered without `onStarted`:

```tsx
<WorkflowRunStartDialog
  isOpen={startWorkflowOpen}
  onClose={() => setStartWorkflowOpen(false)}
  onSwitchToWorkflows={() => setActiveTab('workflows')}
/>
```

The prop is optional so this is not a bug, but if the intended behavior on success is to e.g. navigate to the run or switch a tab, there is no path for that today. Given that `spaceStore` emits a `space.workflowRun.created` event that is already subscribed and will update `workflowRuns` reactively, the dashboard canvas will update on its own — so the omission may be deliberate. Worth an inline comment confirming this intent.

---

### P3 — Placeholder timestamp in tests is hardcoded to a specific date

`RUN_MOCK.title` is `'My Flow — 2024-01-15 10:30'`, which is fine for a mock, but the test on line 222 asserts `expect.stringContaining(RUN_MOCK.title)`. This is safe since the component uses the actual `run.title` from the server response (the mock), not the locally auto-generated one, so the test is correct.

---

### P3 — Minor: empty state shows a form submit button that is always disabled

When `workflows.length === 0`, the form still renders a "Start Run" button (line 165) with `disabled={!hasWorkflows}`. The empty state banner already communicates that no action is possible. A small UX consideration: the "Cancel" button remains active, which is correct. No action needed, just noting for completeness.

---

### Positive Notes

- Clean separation from the older `SpaceStartWorkflowDialog` — the new component correctly reads from the store rather than requiring prop drilling.
- Auto-title logic (`buildAutoTitle`) is a pure function, easy to test and reason about.
- State reset on close (`handleClose`) is thorough — title, workflowId, and error are all cleared.
- The `useEffect` on line 39 correctly handles the async-load case where `workflows` starts empty then populates.
- The `onSwitchToWorkflows` flow closes the dialog before switching tabs, avoiding a stale open dialog after tab change.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-structured component that correctly implements all the stated requirements: workflow dropdown, auto-title generation, empty state with navigation, toast on success, error handling, and integration into SpaceIsland.

### What looks good

- **Correct delegation**: Uses `spaceStore.startWorkflowRun()` and `spaceStore.workflows` signal rather than raw hub calls -- consistent with the store pattern.
- **Auto-title logic**: `buildAutoTitle()` is pure, testable, and produces readable output.
- **UI polish**: Single-workflow indicator (no dropdown), disabled state when no workflows, empty state with "Go to Workflows tab" button that both navigates and closes.
- **Error handling**: Catches both `Error` instances and unknown throws; displays inline error without closing the dialog.
- **Test coverage**: 13 tests covering open/close, empty state, single/multi workflow, auto-title, custom title, success callback, error display, cancel, and disabled button. All pass.
- **SpaceIsland integration**: Properly replaces the old `SpaceStartWorkflowDialog` usage; wires `onSwitchToWorkflows` to `setActiveTab('workflows')`.

### Minor observations (P3)

1. **`workflows` signal read on every render** (`WorkflowRunStartDialog.tsx:32`): `spaceStore.workflows.value` is read at the top of the component function body. This works because the parent (`SpaceIsland`) also reads signals and will re-render, but it means the component is not independently reactive to workflow list changes. This is fine in practice since the dialog is short-lived.

2. **`@ts-nocheck` in tests** (`WorkflowRunStartDialog.test.tsx:1`): Standard pattern for test files with mock modules in this codebase, so no concern.

3. **Old `SpaceStartWorkflowDialog` is now unused outside the barrel export** -- `SpaceIsland` no longer imports it. It could be cleaned up in a follow-up, but leaving it in the barrel is harmless.

lsm (commented):

Re-reviewed after the fix commit (bff2896). All three items from the previous review are addressed:

1. **Null guard in `startWorkflowRun()`** (`packages/web/src/lib/space-store.ts:855`) — Correctly throws `Error('Server returned no run data')` when the hub returns a null run. Matching test added in `packages/web/src/lib/__tests__/space-store.test.ts`.

2. **`onStarted` omission comment** (`packages/web/src/islands/SpaceIsland.tsx:315-317`) — Clear inline comment explains the reactive subscription pattern. Good.

3. **`@ts-nocheck` in tests** — Confirmed this is the standard pattern across all 12 test files in `packages/web/src/components/space/__tests__/`. No issue.

**Implementation is clean and complete.** The component correctly reads from `spaceStore.workflows` signal, delegates to `spaceStore.startWorkflowRun()`, auto-generates titles, handles the empty state with a "Go to Workflows tab" escape hatch, and has 13 focused unit tests. Integration in `SpaceIsland.tsx` is minimal and correct.

No remaining issues found. **Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

---

### Re-review Summary

This is a follow-up to the previous REQUEST_CHANGES review. All three issues have been addressed correctly.

---

### Verified Fixes

**P1 — null guard in `spaceStore.startWorkflowRun()` (FIXED)**

`packages/web/src/lib/space-store.ts` line 855 now reads:
```ts
if (!run) throw new Error('Server returned no run data');
```
The guard is in the right place — after the destructured `run` is obtained from the hub response and before it is returned. A corresponding test was added in `packages/web/src/lib/__tests__/space-store.test.ts` that mocks `{ run: null }` and asserts the method rejects with the exact error message. Both the fix and the test are correct.

**P2 — `@ts-nocheck` in test file (left as-is, accepted)**

The worker provided context that `@ts-nocheck` is an established pattern in this test codebase due to Preact/JSX mock typing complexity. Reviewing other test files confirms this is consistent with the codebase norms. This is acceptable.

**P2 — `onStarted` not wired in `SpaceIsland` (ADDRESSED with explanation)**

`packages/web/src/islands/SpaceIsland.tsx` lines 315-317 now include a clear inline comment explaining that `onStarted` is intentionally omitted because `spaceStore` subscribes to `space.workflowRun.created` events and updates `activeRuns` reactively. The omission is architecturally sound and now clearly documented for future readers.

---

### No New Issues Found

The diff is clean. The component logic, state management, empty-state handling, and test coverage all look correct. The integration in `SpaceIsland` is minimal and correct — it replaces `SpaceStartWorkflowDialog` with `WorkflowRunStartDialog` while wiring `onSwitchToWorkflows` to `setActiveTab('workflows')`.

PR #1102, opened 2026-03-28
feat(space): provision space:chat session on space creation and daemon startup

lsm:
Mirrors the `room:chat:${roomId}` pattern for Spaces — each Space now gets a dedicated coordinator session (`space:chat:${spaceId}`) with MCP tools and a system prompt.

Changes:
- `shared/types.ts`: add `space_chat` to `SessionType` and `SessionMetadata.sessionType`
- `session-lifecycle.ts`: add `spaceId` field and `space_chat` to `CreateSessionParams`
- `space-handlers.ts`: create `space:chat:${spaceId}` session after `space.create` RPC
- `space-runtime-service.ts`: add `setupSpaceAgentSession()`, `subscribeToSpaceEvents()`, `provisionExistingSpaces()`; accept `sessionManager` and `daemonHub` in config
- `rpc-handlers/index.ts`: wire `sessionManager` and `daemonHub` into `SpaceRuntimeService`
- Unit tests updated for both `space-handlers` and `space-runtime-service`

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -337,7 +337,8 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		spaceWorkflowRunRepo,
 		deps.daemonHub,
 		deps.spaceAgentManager,
-		spaceWorkflowManager
+		spaceWorkflowManager,
+		deps.sessionManager
 	);
 
 	const spaceTaskManagerFactory: SpaceTaskManagerFactory = (spaceId: string) => {
@@ -365,6 +366,8 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	// Not started yet: TaskAgentManager is created next and injected before start().
 	// gateDataRepo is injected so notifyGateDataChanged() can trigger lazy node activation
 	// after gate data is written externally (e.g. approveGate RPC, writeGateData RPC).
+	// sessionManager and daemonHub are injected so space:chat:${spaceId} sessions are
+	// provisioned with MCP tools and system prompts on startup and on space.created.
 	const spaceRuntimeService = new SpaceRuntimeService({
 		db: deps.db.getDatabase(),
 		spaceManager: deps.spaceManager,
@@ -373,6 +376,8 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		workflowRunRepo: spaceWorkflowRunRepo,
 		taskRepo: spaceTaskRepo,
 		gateDataRepo,
+		sessionManager: deps.sessionManager,
+		daemonHub: deps.daemonHub,
 	});
 
 	// Space Worktree Manager — one worktree per task, shared by all node agents.
--- a/packages/daemon/src/lib/rpc-handlers/index.ts
+++ b/packages/daemon/src/lib/rpc-handlers/index.ts
@@ -330,17 +330,6 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 	};
 	const spaceWorkflowManager = new SpaceWorkflowManager(spaceWorkflowRepo, agentLookup);
 
-	setupSpaceHandlers(
-		deps.messageHub,
-		deps.spaceManager,
-		spaceTaskRepo,
-		spaceWorkflowRunRepo,
-		deps.daemonHub,
-		deps.spaceAgentManager,
-		spaceWorkflowManager,
-		deps.sessionManager
-	);
-
 	const spaceTaskManagerFactory: SpaceTaskManagerFactory = (spaceId: string) => {
 		return new SpaceTaskManager(deps.db.getDatabase(), spaceId);
 	};
@@ -379,6 +368,21 @@ export function setupRPCHandlers(deps: RPCHandlerDependencies): RPCHandlerSetupR
 		sessionManager: deps.sessionManager,
 		daemonHub: deps.daemonHub,
 	});
+
+	// Register Space RPC handlers now that spaceRuntimeService exists.
+	// spaceRuntimeService is passed so space.create can call setupSpaceAgentSession()
+	// directly after session creation, avoiding reliance on the daemonHub event.
+	setupSpaceHandlers(
+		deps.messageHub,
+		deps.spaceManager,
+		spaceTaskRepo,
+		spaceWorkflowRunRepo,
+		deps.daemonHub,
+		deps.spaceAgentManager,
+		spaceWorkflowManager,
+		deps.sessionManager,
+		spaceRuntimeService
+	);
 
 	// Space Worktree Manager — one worktree per task, shared by all node agents.
 	const spaceWorktreeManager = new SpaceWorktreeManager(deps.db.getDatabase());
--- a/packages/daemon/src/lib/rpc-handlers/space-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/space-handlers.ts
@@ -26,6 +26,7 @@ import type { SpaceAgentManager } from '../space/managers/space-agent-manager';
 import type { SpaceWorkflowManager } from '../space/managers/space-workflow-manager';
 import type { SpaceTaskRepository } from '../../storage/repositories/space-task-repository';
 import type { SpaceWorkflowRunRepository } from '../../storage/repositories/space-workflow-run-repository';
+import type { SessionManager } from '../session-manager';
 import { seedPresetAgents } from '../space/agents/seed-agents';
 import { seedBuiltInWorkflows } from '../space/workflows/built-in-workflows';
 import { Logger } from '../logger';
@@ -47,7 +48,8 @@ export function setupSpaceHandlers(
 	workflowRunRepo: SpaceWorkflowRunRepository,
 	daemonHub: DaemonHub,
 	spaceAgentManager: SpaceAgentManager,
-	spaceWorkflowManager: SpaceWorkflowManager
+	spaceWorkflowManager: SpaceWorkflowManager,
+	sessionManager?: SessionManager
 ): void {
 	// ─── space.create ───────────────────────────────────────────────────────────
 	messageHub.onRequest('space.create', async (data) => {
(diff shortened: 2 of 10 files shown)


lsm (commented):

### Summary

The implementation correctly adds the `space_chat` session type, creates `space:chat:${spaceId}` sessions on space creation, injects MCP tools and system prompts, and provisions existing spaces on daemon startup. The code follows established patterns (mirrors `room:chat:${roomId}`). Unit tests cover the key paths well. However, there are two issues worth addressing.

### P1: Missing online test

The task explicitly requested an online test. No online test was added in `packages/daemon/tests/online/`. The unit tests mock all dependencies, so there is no integration-level verification that the session is actually created and provisioned end-to-end.

### P2: Unhandled promise rejection in `provisionExistingSpaces()`

**File:** `packages/daemon/src/lib/space/runtime/space-runtime-service.ts`, line 153

```typescript
void this.config.spaceManager.listSpaces().then((spaces) => {
    return Promise.all(
        spaces.map((space) =>
            this.setupSpaceAgentSession(space).catch((err) => { ... })
        )
    );
});
```

If `listSpaces()` itself rejects (e.g., DB connection issue during startup), the rejection is unhandled because `void` suppresses the return but there is no `.catch()` on the `listSpaces()` call. The inner `setupSpaceAgentSession` errors are caught, but the outer `listSpaces()` rejection is not. Add a `.catch()` after the `.then()`:

```typescript
void this.config.spaceManager.listSpaces().then((spaces) => {
    return Promise.all(
        spaces.map((space) =>
            this.setupSpaceAgentSession(space).catch((err) => { ... })
        )
    );
}).catch((err) => {
    log.error('Failed to list spaces for provisioning:', err);
});
```

### P3: `new SpaceTaskManager(db, space.id)` created per invocation

**File:** `packages/daemon/src/lib/space/runtime/space-runtime-service.ts`, line 201

A new `SpaceTaskManager` is instantiated on every call to `setupSpaceAgentSession()`. Since `SpaceTaskManager` is lightweight (just a wrapper around a repository), this is functionally fine, but it means a new `SpaceTaskRepository` is also created inside it each time. Consider caching or accepting it as a config dependency if this method could be called frequently (e.g., on space updates). Not blocking.

### What looks good

- Type changes in `shared/types.ts` are clean and consistent -- `space_chat` added to both `SessionType` and `SessionMetadata.sessionType`
- `spaceId` properly threaded through `CreateSessionParams` into `SessionContext`
- Session creation is non-fatal (wrapped in try/catch with `log.warn`) -- consistent with room pattern
- `subscribeToSpaceEvents()` / `stop()` properly manages unsubscribe handles
- Event ordering is correct: session is created before `space.created` is emitted, so the `setupSpaceAgentSession` handler will find it via `getSessionAsync()`
- Unit tests cover: happy path, no-sessionManager case, session creation failure resilience, startup provisioning, event subscription, and unsubscription on stop

lsm (commented):

**Recommendation: REQUEST_CHANGES** — two blocking issues prevent the feature from working correctly.

---

### P0 — Missing DB Migration for 'space_chat' session type

The SQLite CHECK constraint on the `sessions.type` column does not include `'space_chat'`. Every `sessionManager.createSession({ ..., sessionType: 'space_chat' })` call will throw a constraint violation at the database level, silently swallowed by the `catch` in space-handlers.ts and the log-and-continue in provisionExistingSpaces().

**Affected files:**
- `packages/daemon/src/storage/schema/index.ts` line 62 — fresh-DB schema still lists the old set
- `packages/daemon/src/storage/schema/migrations.ts` — no new migration was added

Migration 31 is the exact precedent: it added `'space_task_agent'` by probing a test insert and recreating the table when the constraint rejects it. A Migration 65+ following the same pattern is required here.

Without this fix, every `createSession({ sessionType: 'space_chat' })` call silently fails. The catch block in space-handlers.ts logs a warn and continues, so the space is still created — but with no chat session. The runtime-service `setupSpaceAgentSession()` then finds nothing and logs another warn. The feature is entirely a no-op on existing databases and on fresh databases.

---

### P0 — space:chat: session is never assigned to the space (won't appear in listings)

The task requires sessions to appear in space listings. The room pattern does this explicitly:

```typescript
// room-handlers.ts line 81
roomManager.assignSession(room.id, roomChatSessionId);
```

The space-handlers.ts equivalent **does not call** `spaceManager.addSession(space.id, spaceChatSessionId)`. As a result:
- `space.sessionIds` remains empty after space creation
- `space.overview` (line 287) returns `sessions: []` — the chat session is invisible to the frontend

**Affected file:** `packages/daemon/src/lib/rpc-handlers/space-handlers.ts`, lines 99–116 — `spaceManager.addSession()` call is absent after `sessionManager.createSession()` succeeds.

---

### P1 — No online test was added (required by the task)

The task explicitly asks for an online test. The PR adds unit tests only. No file under `packages/daemon/tests/online/space/` covers space:chat session provisioning (verified by grepping for `space_chat` and `space:chat` across the online test directory). Given the DB constraint bug above, an online test would have caught it immediately.

---

### P1 — Double-provisioning path on space.create

When a space is created:
1. `space-handlers.ts` calls `sessionManager.createSession()` (creates the session in DB)
2. `space-handlers.ts` emits `space.created`
3. `SpaceRuntimeService` catches `space.created` and calls `setupSpaceAgentSession()`

Step 3 runs `setupSpaceAgentSession()` which calls `getSessionAsync()`. If step 2 fires before step 1's DB write is flushed (unlikely, but `daemonHub.emit` is unawaited), step 3 finds nothing and silently no-ops, leaving the session without MCP tools/system-prompt. This timing dependency is fragile — the `space.created` event should carry enough information, or `setupSpaceAgentSession()` should be called directly from the handler after `createSession()` succeeds, rather than relying on the event.

---

### P2 — space.delete / space.archive do not clean up the space:chat session

`space.delete` at line 233 deletes the space record but does not archive or delete the `space:chat:${spaceId}` session. If the feature is intended to mirror the room pattern, room cleanup logic (if any) should be followed here too. The orphaned session will persist in the DB and the session list after the space is deleted.

---

### P2 — Unsafe `as unknown as McpServerConfig` cast in setupSpaceAgentSession()

`packages/daemon/src/lib/space/runtime/space-runtime-service.ts` line 207:
```typescript
session.setRuntimeMcpServers({
    'space-agent-tools': mcpServer as unknown as McpServerConfig,
});
```

`createSpaceAgentMcpServer()` returns a plain object of tool functions, not a `McpServerConfig`. The double cast hides a real type mismatch. This is inherited from existing code in task-agent-manager.ts (lines 381–385, which acknowledge the issue in a comment), so it is pre-existing debt — but this PR propagates the pattern without addressing it.

---

### Summary

The two P0 issues make the feature a complete no-op: the DB constraint rejects every `space_chat` insert (which is caught-and-swallowed), and even if that were fixed, the sessions would not appear in space listings because `spaceManager.addSession()` is never called. The requested online test is also absent.

lsm (commented):

**Recommendation: APPROVE** — all previously blocking issues have been addressed.

**P0 (DB migration) — Fixed.** Migration 66 (`migrations.ts:4279`) adds `'space_chat'` to the sessions type CHECK constraint using the established probe-insert + table-recreate pattern (mirrors Migration 31). The fresh-DB schema in `index.ts:62` is also updated. Both are correct.

**P0 (missing addSession call) — Fixed.** `space-handlers.ts:117` now calls `await spaceManager.addSession(space.id, spaceChatSessionId)` after session creation, so the chat session appears in `space.sessionIds`. Unit test at line 328 verifies this.

**P1 (missing online test) — Fixed.** `packages/daemon/tests/online/space/space-chat-session.test.ts` adds 4 tests covering: session existence with correct type, spaceId in session context, membership in `space.sessionIds`, and persistence across daemon restart. The test directory is already in the CI matrix (`tests/online/space` with `mock_sdk: true`).

**P1 (unhandled promise rejection) — Fixed.** `space-runtime-service.ts:164` now has an outer `.catch()` on the `listSpaces()` promise chain.

**P1 (direct provisioning vs event dependency) — Fixed.** `space-handlers.ts:120-124` now calls `setupSpaceAgentSession()` directly after session creation, eliminating the timing dependency on the async `space.created` event.

### Remaining Observations

**P3 — Double invocation of setupSpaceAgentSession on space.create**

`space-handlers.ts:121` calls `setupSpaceAgentSession()` directly, then line 131 emits `space.created`, which triggers the `subscribeToSpaceEvents()` listener (`space-runtime-service.ts:131`) that calls `setupSpaceAgentSession()` again. The comment on line 119 says "This avoids relying on the space.created event" but doesn't remove the event-driven path. The second call is idempotent (just overwrites the same MCP server and system prompt), so it is harmless — but it creates a redundant `SpaceTaskManager` instance and MCP server object each time a space is created. Consider either removing the event-driven provisioning for newly created spaces or skipping the direct call. Not blocking.

**P3 — Session cleanup on space.delete/archive**

`space.delete` and `space.archive` handlers still do not clean up the `space:chat:${spaceId}` session. This was noted in the previous review as P2. The orphaned session will persist in the DB and session list. This is pre-existing scope and not introduced by this PR's fix commit, so I am not blocking on it, but it should be tracked as follow-up work.

### What looks good

- Migration 66 follows the exact same pattern as Migration 31, reducing risk of divergence.
- The online test covers the full lifecycle including daemon restart, which exercises the `provisionExistingSpaces()` path.
- Error handling is consistent: session creation failures are non-fatal (try/catch with log.warn), matching the room pattern.
- The `setup()` helper in unit tests now accepts `spaceRuntimeService`, and the new test cases at lines 303-365 verify `addSession`, `setupSpaceAgentSession` delegation, and resilience to session creation failure.

lsm (commented):

**Recommendation: APPROVE**

All four P0/P1 issues from the previous review are now correctly addressed. The fix commit (11145651a) is clean and targeted.

---

### P0 Issues — Resolved

**DB migration for `space_chat` type (migration 66)** — Fixed.
`packages/daemon/src/storage/schema/migrations.ts` lines 4279–4341 implement the probe-insert + table-recreate pattern exactly mirroring Migration 31. The fresh-DB schema in `packages/daemon/src/storage/schema/index.ts` line 62 also includes `space_chat`. Both paths are consistent.

**`spaceManager.addSession()` call** — Fixed.
`packages/daemon/src/lib/rpc-handlers/space-handlers.ts` line 117 now calls `await spaceManager.addSession(space.id, spaceChatSessionId)` immediately after session creation. The repository's `addSessionToSpace()` is idempotent (early-return if already present), so the double-call path is safe.

### P1 Issues — Resolved

**Direct `setupSpaceAgentSession()` call in handler** — Fixed.
The handler now calls `spaceRuntimeService.setupSpaceAgentSession()` directly (line 121) rather than relying on the async `space.created` event. The event subscription in `subscribeToSpaceEvents()` remains as a belt-and-suspenders path; because both `setRuntimeMcpServers` and `setRuntimeSystemPrompt` are simple in-memory overwrites, the double invocation is functionally safe.

**`provisionExistingSpaces()` outer catch** — Fixed.
`packages/daemon/src/lib/space/runtime/space-runtime-service.ts` lines 152–166 wrap `listSpaces()` in a `.catch()` so a rejection in the list call doesn't produce an unhandled promise rejection.

**Online test** — Added.
`packages/daemon/tests/online/space/space-chat-session.test.ts` covers: session existence, `type=space_chat`, `spaceId` in context, presence in `space.sessionIds`, and persistence across daemon restart. The file falls under the `tests/online/space` matrix entry in CI (`.github/workflows/main.yml` line 285), so it will run automatically.

---

**Redundant double provisioning on `space.create`** (P3)
The `space.create` handler calls `setupSpaceAgentSession()` directly (space-handlers.ts:121), and the `space.created` event it emits immediately after (line 130) triggers a second call via the subscription in `subscribeToSpaceEvents()`. This is safe due to idempotency, but it is wasteful. A comment acknowledging this would help future readers understand the intentional redundancy. Not a bug.

**Potential partial failure when `addSession` throws** (P3)
If `spaceManager.addSession()` throws after a successful `createSession()`, the catch block at line 125 swallows the error and `setupSpaceAgentSession()` is never invoked. The session exists in DB but is not registered in `space.sessionIds`. This is a pre-existing pattern (same as room-handlers.ts) and the log warning makes it diagnosable. Not a regression.

---

All originally-flagged issues are now correctly resolved. The implementation is solid.

PR #1105, opened 2026-03-28
ci: split cross-provider online tests into 2 parallel CI jobs

lsm:
Splits the single `cross-provider` CI job into two parallel jobs to reduce overall CI time.

## Changes
- `cross-provider-1`: runs `conversation-continuity-after-switch.test.ts` (~353 lines, 17 tests)
- `cross-provider-2`: runs `cross-provider-model-switch.test.ts` (~769 lines, 47 tests)
- Updated `scripts/validate-online-test-matrix.sh` to track cross-provider as a split module, so any future test files added to the directory are caught by the coverage validator.

Both jobs use the same credentials (MiniMax/GLM) and run with `mock_sdk: false`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -143,6 +143,8 @@ jobs:
           # Copilot bridge tests disabled — COPILOT_GITHUB_TOKEN credential issue causing hard-fail
           # - providers-anthropic-copilot
           - rewind
+          - cross-provider-1
+          - cross-provider-2
           # Room tests temporarily disabled during refactoring
           # - room-1
           # - room-multi-agent
@@ -289,8 +291,13 @@ jobs:
             test_path: tests/online/websocket
             mock_sdk: true
           # Cross-provider model switching tests - requires real MiniMax and GLM credentials
-          - module: cross-provider
-            test_path: tests/online/cross-provider
+          # Split into 2 parallel jobs to reduce overall CI time.
+          - module: cross-provider-1
+            test_path: tests/online/cross-provider/conversation-continuity-after-switch.test.ts
+            mock_sdk: false
+            default_provider: minimax
+          - module: cross-provider-2
+            test_path: tests/online/cross-provider/cross-provider-model-switch.test.ts
             mock_sdk: false
             default_provider: minimax
 
--- a/scripts/validate-online-test-matrix.sh
+++ b/scripts/validate-online-test-matrix.sh
@@ -76,6 +76,11 @@ PROVIDERS_FILES=(
   anthropic-to-codex-bridge-provider.test.ts  # CI shard disabled — kept here so validator doesn't flag it
 )
 
+CROSS_PROVIDER_FILES=(
+  conversation-continuity-after-switch.test.ts
+  cross-provider-model-switch.test.ts
+)
+
 check_split_module() {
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean, minimal change that correctly splits the `cross-provider` online test module into two parallel CI matrix entries.

### What was verified

1. **Matrix entries are correct.** The old single `cross-provider` entry in both the module list and include block is fully replaced by `cross-provider-1` and `cross-provider-2`. No orphaned references remain.

2. **Test paths match disk.** The two files referenced (`conversation-continuity-after-switch.test.ts` and `cross-provider-model-switch.test.ts`) are exactly the files that exist in `packages/daemon/tests/online/cross-provider/`. No files are missed or misnamed.

3. **Properties preserved.** Both new entries correctly carry over `mock_sdk: false` and `default_provider: minimax` from the original entry. The old entry only had these two extra properties, and both are faithfully replicated.

4. **Validation script updated.** `CROSS_PROVIDER_FILES` array added with both filenames, and `check_split_module "cross-provider"` call added. The `KNOWN_DIRS` list on line 130 already included `cross-provider`, so no change was needed there.

5. **Validation passes.** Running `bash scripts/validate-online-test-matrix.sh` returns success.

No issues found.

lsm (commented):

**Recommendation: APPROVE**

The implementation is correct and complete. Both files changed are minimal, targeted, and consistent with the existing CI patterns.

### What was verified

**Correctness of the split:**
- The original `cross-provider` entry existed only in the `include:` section (not in the `module:` list). In GitHub Actions, such an `include` entry still creates a standalone job, so the tests were running. The new entries correctly appear in both the `module:` list (lines 146-147) and the `include:` section (lines 295-302), which is the same pattern used by `rpc-1/2/3/4`, `features-1/2/3`.

**Test files exist on disk:**
- `packages/daemon/tests/online/cross-provider/conversation-continuity-after-switch.test.ts` — confirmed present.
- `packages/daemon/tests/online/cross-provider/cross-provider-model-switch.test.ts` — confirmed present.

**Credential wiring:**
- Both entries have `mock_sdk: false` and `default_provider: minimax`, so the workflow env block correctly injects `GLM_API_KEY`, `MINIMAX_API_KEY`, and `DEFAULT_PROVIDER` (lines 349-352). No Anthropic keys are injected for these modules, which is correct since the tests require only MiniMax and GLM credentials.

**Validation script:**
- `CROSS_PROVIDER_FILES` array matches both filenames exactly.
- `check_split_module "cross-provider"` call correctly validates the directory against the array in both directions (disk vs. matrix).
- The function will detect any future test file added to the directory that is not tracked in the matrix.

`cross-provider-2` (`cross-provider-model-switch.test.ts`) has two individual tests with 120-second per-test timeouts (lines 699 and 767 of that file) and `waitForIdle` calls with 60-second waits. No explicit `timeout:` override is set on these new matrix entries, so the CI job timeout defaults to 5 minutes (`matrix.timeout || 5`). The combined test run could be tight under load. The original combined job also lacked a timeout override so this is not a regression, but adding `timeout: 10` to `cross-provider-2` would be a safe improvement (mirroring the `git` module which also uses real provider calls and has `timeout: 10`).

PR #1114, opened 2026-03-28
feat(neo): wire NeoAgentManager into daemon app lifecycle (task 1.3)

lsm:
Wire `NeoAgentManager` into `createDaemonApp` and `DaemonAppContext`.

## Changes

- **`DaemonAppContext`** — adds `neoAgentManager: NeoAgentManager` field
- **`createDaemonApp`** — instantiates `NeoAgentManager(sessionManager, settingsManager)` after `sessionManager.start()`; calls `provision()` at startup (guarded by `NODE_ENV !== 'test' || NEOKAI_ENABLE_NEO_AGENT=1`); calls `cleanup()` in shutdown before `sessionManager.cleanup()`; returns `neoAgentManager` in context
- **`neo-daemon-lifecycle.test.ts`** — 6 new unit tests covering: context field presence, test-mode skip, `NEOKAI_ENABLE_NEO_AGENT=1` provision, DB-persistent restart re-attach, cleanup invocation, default security mode

Closes task 1.3 of the Neo Agent goal.

### DIFF
--- a/packages/daemon/.env.example
+++ b/packages/daemon/.env.example
@@ -87,6 +87,15 @@ MAX_SESSIONS=10
 # NEOKAI_JOB_QUEUE_MAX_CONCURRENT=5
 
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+# Neo Agent Configuration
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+#
+# Set to 1 to enable Neo agent provisioning in NODE_ENV=test environments
+# (e.g., online tests that exercise the neo.* RPC surface).
+# In production/development, Neo is always provisioned at startup.
+# NEOKAI_ENABLE_NEO_AGENT=1
+
+# ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 # Test Configuration
 # ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 #
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -463,6 +463,10 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 			await neoAgentManager.provision();
 			logInfo('[Daemon] Neo agent provisioned');
 		} catch (err) {
+			// Non-fatal: daemon continues without Neo. The Neo session will be null
+			// and the frontend will receive no responses from neo.* RPC calls.
+			// TODO(neo): publish a status channel event so the frontend can surface
+			// a "Neo unavailable" indicator when provisioning fails.
 			logError('[Daemon] Neo agent provisioning failed (non-fatal):', err);
 		}
 	}
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -29,6 +29,7 @@ import { JOB_QUEUE_CLEANUP, SKILL_VALIDATE } from './lib/job-queue-constants';
 import { AppMcpLifecycleManager, seedDefaultMcpEntries } from './lib/mcp';
 import { FileIndex } from './lib/file-index';
 import { SkillsManager } from './lib/skills-manager';
+import { NeoAgentManager } from './lib/neo/neo-agent-manager';
 
 export interface CreateDaemonAppOptions {
 	config: Config;
@@ -82,6 +83,8 @@ export interface DaemonAppContext {
 	appMcpManager: AppMcpLifecycleManager;
 	/** Application-level Skills manager — registry CRUD and validation */
 	skillsManager: SkillsManager;
+	/** Neo agent manager — singleton global AI assistant */
+	neoAgentManager: NeoAgentManager;
 	/** Workspace file index for fast fuzzy file/folder search */
 	fileIndex: FileIndex;
 	/**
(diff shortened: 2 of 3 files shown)


lsm (commented):

---

**Recommendation: APPROVE** -- the implementation is clean, follows existing patterns, and all tests pass.

### What was done well

1. **Correct wiring**: `NeoAgentManager` is instantiated after `sessionManager.start()` (which is synchronous -- just registers job handlers), so the ordering is safe. The constructor takes `(sessionManager, settingsManager)` matching the `NeoSessionManager` / `NeoSettingsManager` interfaces, providing good testability.

2. **Test-mode guard**: The `NODE_ENV !== 'test' || NEOKAI_ENABLE_NEO_AGENT === '1'` pattern mirrors the spaces agent approach documented in CLAUDE.md and ensures unit tests don't accidentally spawn agent sessions.

3. **Non-fatal provisioning**: The `try/catch` around `provision()` (line 462-467 in `app.ts`) correctly treats provisioning failure as non-fatal, preventing daemon startup from being blocked.

4. **Shutdown ordering**: `neoAgentManager.cleanup()` runs before `sessionManager.cleanup()` (line 585 vs 592), which is correct -- the Neo session is stopped before the pool drains. This matches the `taskAgentManager.cleanupAll()` pattern on the line above.

5. **Test quality**: All 6 tests are well-structured with proper `beforeEach`/`afterEach` cleanup of env vars, console mocks, and Bun.serve spies. The DB persistence test (test 4) uses on-disk SQLite to verify re-attach across daemon restarts, which is a meaningful integration test.

6. **Context field**: The `neoAgentManager: NeoAgentManager` field in `DaemonAppContext` has a clear JSDoc comment, consistent with all other fields.

1. **Double-cleanup path** (P3): `NeoAgentManager.cleanup()` calls `this.session.cleanup()` then sets `this.session = null`, but does NOT call `sessionManager.unregisterSession(NEO_SESSION_ID)`. When `sessionManager.cleanup()` runs immediately after (line 592), it iterates `sessionCache.entries()` and calls `agentSession.cleanup()` on the Neo session again. The `QueryLifecycleManager.cleanup()` has no idempotency guard, so the session gets cleaned up twice. This is harmless in practice (all phases are wrapped in try/catch, `stop()` is effectively a no-op on an already-stopped query) but is slightly wasteful. Consider adding `this.sessionManager.unregisterSession(NEO_SESSION_ID)` in `NeoAgentManager.cleanup()` to avoid this.

2. **DB cleanup in persistence test** (P3): The on-disk DB cleanup (lines 159-165) uses `Bun.file(dbPath).arrayBuffer()` as an existence check before `fs.unlink()`. A simpler pattern would be `fs.unlink(dbPath).catch(() => {})` since `unlink` already throws if the file doesn't exist.

3. **`NEOKAI_ENABLE_NEO_AGENT` not cleared in `setup.ts`** (P3): The global test setup clears API keys but not `NEOKAI_ENABLE_NEO_AGENT`. The test file handles this itself in `beforeEach`/`afterEach`, so this is not a bug, but if another test file accidentally sets this env var without cleaning up, it could leak into the neo lifecycle tests. Low risk given the explicit save/restore pattern used.

lsm (commented):

### Summary

The implementation correctly wires `NeoAgentManager` into the daemon app lifecycle. The core functionality is present and tests pass. However, there are a few issues that should be addressed.

---

### P1 Issues

#### 1. Test spies on method after instance is created (line 174)

**File:** `/packages/daemon/tests/unit/core/neo-daemon-lifecycle.test.ts:173-178`

```typescript
const ctx = await createDaemonApp({ config: makeConfig(), verbose: false });

// Spy on neoAgentManager.cleanup to verify it was called.
const cleanupSpy = spyOn(ctx.neoAgentManager, 'cleanup');

await ctx.cleanup();

expect(cleanupSpy).toHaveBeenCalledTimes(1);
```

The spy is created **after** `createDaemonApp` returns, which means it wraps the method on the already-instantiated `NeoAgentManager` instance. This test **does correctly verify** that `ctx.cleanup()` calls `neoAgentManager.cleanup()` because the spy wraps the method before `ctx.cleanup()` is invoked.

However, there is a subtle concern: the spy creates a wrapper around the original method. If `cleanup()` has already been called (it has not in this test), the spy would miss it. The test is correct for the current assertion, but the comment "Spy on neoAgentManager.cleanup to verify it was called" could be clearer that we are spying on future calls only.

**Recommendation:** The test is functionally correct. Consider adding a clarifying comment that the spy tracks calls from `ctx.cleanup()` forward.

---

### P2 Issues

#### 2. Provisioning failure is non-fatal but silently logged

**File:** `/packages/daemon/src/app.ts:461-468`

```typescript
if (process.env.NODE_ENV !== 'test' || process.env.NEOKAI_ENABLE_NEO_AGENT === '1') {
    try {
        await neoAgentManager.provision();
        logInfo('[Daemon] Neo agent provisioned');
    } catch (err) {
        logError('[Daemon] Neo agent provisioning failed (non-fatal):', err);
    }
}
```

If provisioning fails in production, the daemon continues running without a Neo session. This matches the spaces agent pattern but could leave the system in an unexpected state where Neo is unavailable.

**Recommendation:** This is acceptable for now, but consider whether the frontend should be notified of Neo unavailability (e.g., via a status channel or state sync) so users understand why Neo features are disabled.

---

### P3 Issues (Nits)

#### 3. Missing CI environment variable documentation

**File:** `/packages/daemon/src/app.ts`

The `NEOKAI_ENABLE_NEO_AGENT` environment variable is not documented in `.env.example` or any CI workflow files, unlike `NEOKAI_ENABLE_SPACES_AGENT` which appears in `.github/workflows/main.yml`.

**Recommendation:** Add documentation for this environment variable if it will be used in CI or production deployments.

---

### What's Done Well

1. **Correct instantiation order:** `NeoAgentManager` is instantiated after `sessionManager.start()` as required by its constructor signature.

2. **Correct cleanup order:** `neoAgentManager.cleanup()` is called before `sessionManager.cleanup()` and before `taskAgentManager.cleanupAll()`, which ensures Neo is shut down before the session pool drains.

3. **Test mode guard:** The `NODE_ENV !== 'test' || NEOKAI_ENABLE_NEO_AGENT === '1'` pattern correctly mirrors the spaces agent pattern.

4. **Comprehensive test coverage:** Tests cover all acceptance criteria: context field presence, test-mode skip, explicit enable, restart re-attach, cleanup invocation, and default security mode.

5. **On-disk DB test:** The restart re-attach test uses a file-based SQLite DB to verify session persistence across daemon restarts, which is a good integration test pattern.

---

### Recommendation: APPROVE

The implementation is correct and complete. The P1 issue is a minor test clarity concern that does not affect correctness. The P2 issue is an architectural consideration for future work.

PR #1121, opened 2026-03-28
feat(neo): Goal and task query tools (task 2.3)

lsm:
Add 5 MCP query tools for goals and tasks to `neo-query-tools.ts`:

- `list_goals` — cross-room listing, filterable by room/status/mission_type
- `get_goal_details` — full info including structuredMetrics, schedule, execution history
- `get_metrics` — KPI values with computed progress % for measurable goals
- `list_tasks` — cross-room listing, filterable by room/status/assigned_agent
- `get_task_detail` — full info including result, error, dependencies, PR info

Cross-room queries iterate `roomManager.listRooms(true)` and call per-room repository methods. `NeoQueryGoalRepository` gains `getGoal`/`listExecutions`; new `NeoQueryTaskRepository` interface added; `NeoToolsConfig` gains `taskRepository`.

58 tests covering all tools, filters, edge cases, and MCP server registration (10 tools total).

### DIFF
--- a/packages/daemon/src/lib/neo/tools/neo-query-tools.ts
+++ b/packages/daemon/src/lib/neo/tools/neo-query-tools.ts
@@ -17,6 +17,11 @@
  * - list_space_agents
  * - list_space_workflows
  * - list_space_runs
+ * - list_goals
+ * - get_goal_details
+ * - get_metrics
+ * - list_tasks
+ * - get_task_detail
  *
  * Pattern: two-layer design (testable handlers + MCP server wrapper)
  *   createNeoQueryToolHandlers(config) → plain handler functions
@@ -38,6 +43,8 @@ import type {
 	SpaceWorkflow,
 	SpaceWorkflowRun,
 	SpaceTask,
+	NeoTask,
+	MissionExecution,
 } from '@neokai/shared';
 import { isWorkerSessionId } from '../../room/session-utils';
 
@@ -59,6 +66,21 @@ export interface NeoQueryRoomManager {
 export interface NeoQueryGoalRepository {
 	/** List goals for a specific room, optionally filtered by status */
 	listGoals(roomId: string, status?: string): RoomGoal[];
+	/** Get a single goal by ID */
+	getGoal(id: string): RoomGoal | null;
+	/** List execution history for a goal (most recent first) */
+	listExecutions(goalId: string, limit?: number): MissionExecution[];
+}
+
+export interface NeoQueryTaskRepository {
+	/**
+	 * List tasks for a specific room, optionally filtered by status and archived state.
+	 * Note: assignedAgent filtering is NOT in the real TaskFilter; it is applied in-memory
+	 * by the tool handler after calling this method.
+	 */
+	listTasks(roomId: string, filter?: { status?: string; includeArchived?: boolean }): NeoTask[];
+	/** Get a single task by ID */
+	getTask(id: string): NeoTask | null;
 }
 
 export interface NeoQuerySessionManager {
@@ -122,6 +144,7 @@ export interface NeoQuerySpaceTaskRepository {
 export interface NeoToolsConfig {
 	roomManager: NeoQueryRoomManager;
 	goalRepository: NeoQueryGoalRepository;
+	taskRepository: NeoQueryTaskRepository;
 	sessionManager: NeoQuerySessionManager;
 	settingsManager: NeoQuerySettingsManager;
 	authManager: NeoQueryAuthManager;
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

Clean, well-structured implementation that follows the established two-layer pattern perfectly. All 58 tests pass, typecheck is clean, and the code is consistent with the existing query tools. I'm recommending APPROVE with a few observations for the production wiring phase.

### Findings

**P2 -- `NeoQueryTaskRepository.listTasks` filter interface diverges from real `TaskRepository`**

At `/packages/daemon/src/lib/neo/tools/neo-query-tools.ts:60-63`, the `NeoQueryTaskRepository` interface declares an `assignedAgent` filter parameter:

```ts
listTasks(
    roomId: string,
    filter?: { status?: string; assignedAgent?: string; includeArchived?: boolean }
): NeoTask[];
```

The real `TaskRepository.listTasks` at `/packages/daemon/src/storage/repositories/task-repository.ts:117` uses `TaskFilter` from shared, which only has `{ status?, priority?, includeArchived? }` -- there is no `assignedAgent` support at the SQL level. When this module is wired into production, the adapter layer will need to either add `assignedAgent` to `TaskFilter` + the SQL query, or do post-query in-memory filtering. This isn't a bug today (tests mock it), but it will surface during integration.

**P2 -- `list_tasks` MCP schema allows `status='archived'` with `include_archived=false` default**

At `/packages/daemon/src/lib/neo/tools/neo-query-tools.ts:639-651`, the `status` enum includes `'archived'` while `include_archived` defaults to `false`. A caller requesting `list_tasks(status='archived')` will get zero results because archived tasks are filtered out before the status filter applies. Consider either auto-setting `includeArchived=true` when `status='archived'` is explicitly passed, or documenting this interaction in the tool description.

**P3 -- Progress calculation edge case for `decrease` direction when `baseline === target`**

At `/packages/daemon/src/lib/neo/tools/neo-query-tools.ts:417-424`, when `direction === 'decrease'` and `baseline === target`, the condition `m.baseline !== m.target` is false, so the code falls through to the general `increase` formula. For a decrease metric where baseline already equals target (goal is pre-achieved), this would compute `current / target * 100` instead of 100%. This is a degenerate edge case, but worth a comment or guard.

**P3 -- Module not wired into production**

The `neo-query-tools.ts` module is not imported anywhere in the production source tree. The `taskRepository` field on `NeoToolsConfig` is only consumed by tests. This is expected for incremental delivery, but it should be tracked to ensure the production wiring happens in a follow-up task.

### What's done well

- Consistent two-layer pattern (handlers + MCP wrapper) matching the existing 5 tools
- Comprehensive test coverage: cross-room queries, all filters, error cases, null handling, execution limit, archived task inclusion/exclusion
- Proper use of `?? null` and `?? 'default'` for optional fields in output shapes
- Clean interface design with minimal surface area
- Test fixtures are well-structured and the mock factories correctly simulate repository behavior including filter ordering

**Recommendation: APPROVE** -- The code is correct for the scope of this task. The P2 items are integration concerns for the production wiring phase.

lsm (commented):

### Re-review of fix commit 400653e

All three issues from the previous review have been correctly addressed:

**Fix 1 -- `assignedAgent` removed from `NeoQueryTaskRepository.listTasks` filter (P2, fixed)**

At `packages/daemon/src/lib/neo/tools/neo-query-tools.ts:58-67`, the interface now declares `listTasks(roomId, filter?: { status?; includeArchived? })` with no `assignedAgent`. The doc comment on line 62 explicitly notes that `assignedAgent` filtering is applied in-memory by the handler. The handler at lines 464-469 now calls `listTasks` without `assignedAgent` and applies `tasks.filter(...)` afterward. The mock `makeTaskRepository` in the test file (line 157) also correctly omits the `assignedAgent` filter. This aligns the interface with the real `TaskFilter` in `packages/shared/src/types/neo.ts:361-366`.

**Fix 2 -- Auto-enable `includeArchived` when `status='archived'` (P2, fixed)**

At line 444, `const includeArchived = args.include_archived ?? args.status === 'archived'` correctly promotes to `true` when a caller explicitly requests archived status. The comment on lines 442-443 explains the rationale. A regression test at line 976-993 (`'auto-enables include_archived when status="archived" to avoid zero results'`) verifies this behavior end-to-end against the mock repository.

**Fix 3 -- `baseline === target` guard for decrease direction (P3, fixed)**

At lines 419-422, the IIFE now checks `if (m.baseline === m.target) return 100` before the division, preventing a division-by-zero in the degenerate case. The regression test at line 817-834 (`'returns 100% progress when decrease metric baseline equals target'`) confirms this with `baseline: 0, target: 0, current: 0, direction: 'decrease'`.

**Tests**: All 60 tests pass (up from 58, with 2 new regression tests). The test coverage for the fixes is direct and clear.

No remaining issues found.

**Recommendation: APPROVE**

lsm (commented):
<review body>

PR #1122, opened 2026-03-28
feat(neo): Skill and MCP server query tools (task 2.4)

lsm:
Add four new read-only Neo query tools for MCP servers and skills.

**New tools:**
- `list_mcp_servers` — all registered app-level MCP servers with enabled/disabled status
- `get_mcp_server_status` — server details: transport type, command/URL, env and header key names (values omitted to prevent secret leakage)
- `list_skills` — all skills with sourceType, enabled, builtIn, validationStatus
- `get_skill_details` — full skill info including config and createdAt

**Implementation:**
- Two new interfaces (`NeoQueryMcpServerRepository`, `NeoQuerySkillsManager`) added to `NeoToolsConfig` — same pattern as existing room/session interfaces
- Handlers wired through `AppMcpServerRepository.list()/get()` and `SkillsManager.listSkills()/getSkill()`
- 21 new unit tests: happy path, not-found errors, empty lists, disabled entries, all source types (stdio/sse, builtin/plugin/mcp_server), secret exclusion for env and header values
- `neo-query` MCP server now registers 9 tools (up from 5)

### DIFF
--- a/packages/daemon/src/lib/neo/tools/neo-query-tools.ts
+++ b/packages/daemon/src/lib/neo/tools/neo-query-tools.ts
@@ -7,6 +7,10 @@
  * - get_room_details
  * - get_system_info
  * - get_app_settings
+ * - list_mcp_servers
+ * - get_mcp_server_status
+ * - list_skills
+ * - get_skill_details
  *
  * Pattern: two-layer design (testable handlers + MCP server wrapper)
  *   createNeoQueryToolHandlers(config) → plain handler functions
@@ -15,7 +19,15 @@
 
 import { createSdkMcpServer, tool } from '@anthropic-ai/claude-agent-sdk';
 import { z } from 'zod';
-import type { GlobalSettings, AuthStatus, Room, RoomGoal, TaskSummary } from '@neokai/shared';
+import type {
+	GlobalSettings,
+	AuthStatus,
+	Room,
+	RoomGoal,
+	TaskSummary,
+	AppMcpServer,
+	AppSkill,
+} from '@neokai/shared';
 import { isWorkerSessionId } from '../../room/session-utils';
 
 // ---------------------------------------------------------------------------
@@ -54,6 +66,16 @@ export interface NeoQueryAuthManager {
 	getAuthStatus(): Promise<AuthStatus>;
 }
 
+export interface NeoQueryMcpServerRepository {
+	list(): AppMcpServer[];
+	get(id: string): AppMcpServer | null;
+}
+
+export interface NeoQuerySkillsManager {
+	listSkills(): AppSkill[];
+	getSkill(id: string): AppSkill | null;
+}
+
 /**
  * All dependencies required by the Neo query tools.
  */
@@ -63,6 +85,8 @@ export interface NeoToolsConfig {
 	sessionManager: NeoQuerySessionManager;
 	settingsManager: NeoQuerySettingsManager;
 	authManager: NeoQueryAuthManager;
+	mcpServerRepository: NeoQueryMcpServerRepository;
+	skillsManager: NeoQuerySkillsManager;
 	/** Absolute path to the workspace root */
 	workspaceRoot: string;
 	/** Human-readable app version string, e.g. "0.1.1" */
@@ -103,6 +127,8 @@ export function createNeoQueryToolHandlers(config: NeoToolsConfig) {
 		sessionManager,
 		settingsManager,
 		authManager,
+		mcpServerRepository,
+		skillsManager,
 		workspaceRoot,
 		appVersion,
 		startedAt,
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Summary

Clean, well-structured addition of four read-only MCP query tools for skills and MCP servers. The implementation follows the established two-layer pattern (testable handlers + MCP server wrapper) and the existing interface-based dependency injection convention. All 46 tests pass. No blocking issues.

### P2: Missing test for `http` source type in `get_mcp_server_status`

The handler at `packages/daemon/src/lib/neo/tools/neo-query-tools.ts:314` uses an `if (stdio) / else` branch that covers both `sse` and `http` identically. While the existing `sse` test implicitly covers the `else` branch, there is no explicit test for `sourceType: 'http'`. Since `AppMcpServerSourceType = 'stdio' | 'sse' | 'http'`, adding one test case for `http` would document the supported types more thoroughly and catch a regression if the branching logic is ever changed.

### P3: `beforeEach` import is used, no issue -- just noting the test file is clean

No unused imports detected. All 46 tests are meaningful.

### What looks good

- **Interface design**: `NeoQueryMcpServerRepository` and `NeoQuerySkillsManager` are minimal, containing only the methods actually used. This matches the existing pattern for `NeoQueryRoomManager`, `NeoQuerySessionManager`, etc.
- **Secret hygiene**: `get_mcp_server_status` correctly strips env values and header values, exposing only key names. This is the right security boundary for a read-only query tool.
- **Consistent error handling**: Not-found cases return `{ success: false, error: "..." }` matching the existing pattern.
- **`list_skills` omits `config` and `createdAt`**: Good separation between list (summary) and get (detail) responses.
- **Type compatibility**: `AppMcpServerRepository.list()` returns `AppMcpServer[]` and `SkillsManager.listSkills()` returns `AppSkill[]` -- both match the interface signatures exactly.
- **No broken wiring**: `NeoToolsConfig` is only consumed within this module and its test file, so the two new required fields don't break any other code.

### Recommendation: APPROVE

lsm (commented):

**Recommendation: APPROVE**

Clean, well-structured implementation that follows the established two-layer pattern (handler functions + MCP server wrapper) exactly as the existing tools do. The four new tools are consistent with the codebase conventions.

### What was done well

- **Interface segregation**: `NeoQueryMcpServerRepository` and `NeoQuerySkillsManager` expose only the surface needed by the query tools (`list`/`get`), matching the pattern used by the other interfaces in `NeoToolsConfig`. The real `AppMcpServerRepository.list()/get()` and `SkillsManager.listSkills()/getSkill()` signatures are compatible.
- **Secret handling**: Env values and header values are deliberately stripped from `get_mcp_server_status`, exposing only key names. This is defensively correct even though the `AppMcpServer` type docs state env values are reference keys rather than raw secrets.
- **Consistent error handling**: Not-found cases return the same `{ success: false, error: "..." }` pattern used by the existing room and session tools.
- **Test coverage**: 21 new unit tests covering happy paths, empty lists, not-found errors, all source types (stdio/sse/http for MCP servers, builtin/plugin/mcp_server for skills), null/undefined edge cases, and the negative assertion that `list_skills` excludes `config` and `createdAt`. All 47 tests pass.
- **Transport branching**: The `sse`/`http` fallback via `else` is correct -- both share `url` and `headers` fields, and tests verify all three source types.

1. `get_mcp_server_status` exposes `server.args` (command arguments) in full. While these are typically non-sensitive, a user could theoretically pass a secret token as an argument. This is not worth changing now since the codebase convention for MCP args is non-sensitive flags, but worth noting if the security review later tightens up.

2. The `get_skill_details` handler exposes `config` which includes `pluginPath` (absolute filesystem path). This is acceptable for a read-only query tool where the Neo agent already has workspace filesystem access, and the path visibility matches what the frontend already displays.

3. The PR description says "47 unit tests" but 21 of those are new -- the remaining 26 are pre-existing from the original 5-tool implementation. A minor count discrepancy in the PR description.

No P0/P1/P2 issues found.

lsm (commented):

### Summary

The P2 suggestion to add an explicit `http` source type test case has been adequately addressed. The PR now includes 47 unit tests (up from 26), with 21 new tests covering the four new Neo query tools.

### P2 Addressed: http source type test case

The new test case at `packages/daemon/tests/unit/neo/neo-query-tools.test.ts:632-651` properly covers the `http` source type:

- Creates an MCP server with `sourceType: 'http'`
- Verifies the returned `transport.sourceType` is `'http'`
- Asserts URL is returned correctly
- Confirms header key names are exposed (`headerKeys`) but values are not (`headers` is undefined)

The implementation correctly handles `http` and `sse` identically via the `else` branch (line 319), which is the right design since both are URL-based transports with headers.

### Implementation Quality

The code follows the established patterns in the file:
- Two-layer design (testable handlers + MCP server wrapper)
- Minimal interface segregation for dependencies
- Consistent error handling with descriptive messages
- Security-conscious design (env values and header values omitted to prevent secret leakage)

### Test Coverage

All 47 tests pass. The new tests cover:
- Happy paths for all 4 tools
- Error cases (not found)
- Empty lists
- All three source types for MCP servers (stdio, sse, http)
- All three source types for skills (builtin, plugin, mcp_server)
- All three validation statuses (valid, invalid, pending)
- Secret exclusion verification for both env and headers

### Recommendation: APPROVE

No P0, P1, or P2 issues remain. The PR is ready to merge.

PR #1127, opened 2026-03-28
feat(neo): Neo Settings section (task 8.1)

lsm:
Add a Neo Agent section to Global Settings with three controls:

- **Security mode** dropdown (Conservative / Balanced (default) / Autonomous) — reads from `neo.getSettings`, persists via `neo.updateSettings`
- **Model** dropdown (App default / Sonnet / Opus / Haiku) — reads and persists the same way
- **Clear Session** button — shows inline confirmation dialog, calls `neoStore.clearSession()` on confirm

Wired into `ContextPanel` navigation (between Fallback Models and Usage) and `MainContent` renderer. Includes 15 unit tests covering load, change, RPC calls, error reversal, confirmation dialog, and clear session flow.

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/neo-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/neo-handlers.ts
@@ -203,7 +203,7 @@ export function setupNeoHandlers(
 	messageHub.onRequest('neo.updateSettings', async (data) => {
 		const { securityMode, model } = (data ?? {}) as {
 			securityMode?: string;
-			model?: string;
+			model?: string | null;
 		};
 
 		const updates: Record<string, unknown> = {};
@@ -218,10 +218,14 @@ export function setupNeoHandlers(
 		}
 
 		if (model !== undefined) {
-			if (typeof model !== 'string' || model.trim() === '') {
-				throw new Error('model must be a non-empty string');
+			if (model === null) {
+				// null clears the override — Neo falls back to the app's primary model
+				updates.neoModel = null;
+			} else if (typeof model !== 'string' || model.trim() === '') {
+				throw new Error('model must be a non-empty string or null');
+			} else {
+				updates.neoModel = model.trim();
 			}
-			updates.neoModel = model.trim();
 		}
 
 		if (Object.keys(updates).length === 0) {
--- a/packages/daemon/tests/unit/rpc-handlers/neo-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc-handlers/neo-handlers.test.ts
@@ -406,10 +406,18 @@ describe('Neo RPC Handlers', () => {
 			);
 		});
 
+		it('clears neoModel override when model is null', async () => {
+			const handler = hubData.handlers.get('neo.updateSettings');
+			await handler!({ model: null }, {});
+			expect(settingsManager.updateGlobalSettings).toHaveBeenCalledWith(
+				expect.objectContaining({ neoModel: null })
+			);
+		});
+
 		it('throws for empty model string', async () => {
 			const handler = hubData.handlers.get('neo.updateSettings');
 			await expect(handler!({ model: '   ' }, {})).rejects.toThrow(
-				'model must be a non-empty string'
+				'model must be a non-empty string or null'
 			);
 		});
 
--- a/packages/web/src/components/settings/NeoSettings.tsx
+++ b/packages/web/src/components/settings/NeoSettings.tsx
@@ -1,0 +1,194 @@
+import { useEffect, useState } from 'preact/hooks';
+import { connectionManager } from '../../lib/connection-manager';
+import { neoStore } from '../../lib/neo-store';
+import { toast } from '../../lib/toast.ts';
+import { SettingsSection, SettingsRow, SettingsSelect } from './SettingsSection.tsx';
+
+type NeoSecurityMode = 'conservative' | 'balanced' | 'autonomous';
+
+const SECURITY_MODE_OPTIONS: Array<{ value: NeoSecurityMode; label: string; description: string }> =
+	[
+		{
+			value: 'conservative',
+			label: 'Conservative',
+			description: 'Confirm every write action before executing',
+		},
+		{
+			value: 'balanced',
+			label: 'Balanced (default)',
+			description:
+				'Auto-execute low-risk actions; confirm medium-risk; require explicit phrasing for irreversible changes',
+		},
+		{
+			value: 'autonomous',
+			label: 'Autonomous',
+			description: 'Execute all actions immediately without confirmation',
+		},
+	];
+
+const MODEL_OPTIONS = [
+	{ value: '', label: 'App default' },
+	{ value: 'sonnet', label: 'Claude Sonnet 4' },
+	{ value: 'opus', label: 'Claude Opus 4' },
+	{ value: 'haiku', label: 'Claude Haiku 3.5' },
+];
+
+interface NeoSettingsState {
+	securityMode: NeoSecurityMode;
+	model: string | null;
+}
+
+export function NeoSettings() {
+	const [settings, setSettings] = useState<NeoSettingsState>({
+		securityMode: 'balanced',
+		model: null,
+	});
+	const [isUpdating, setIsUpdating] = useState(false);
+	const [isLoading, setIsLoading] = useState(true);
+	const [showClearConfirm, setShowClearConfirm] = useState(false);
+	const [isClearing, setIsClearing] = useState(false);
+
+	useEffect(() => {
+		let cancelled = false;
+		const load = async () => {
+			try {
+				const hub = await connectionManager.getHub();
+				const response = await hub.request<{ securityMode: NeoSecurityMode; model: string | null }>(
+					'neo.getSettings',
+					{}
+				);
+				if (!cancelled) {
+					setSettings({
+						securityMode: response.securityMode ?? 'balanced',
+						model: response.model ?? null,
+					});
+				}
+			} catch {
+				if (!cancelled) {
+					toast.error('Failed to load Neo settings');
+				}
+			} finally {
+				if (!cancelled) {
+					setIsLoading(false);
+				}
+			}
+		};
+		load();
+		return () => {
+			cancelled = true;
+		};
+	}, []);
+
+	const handleSecurityModeChange = async (value: string) => {
+		const mode = value as NeoSecurityMode;
+		const prev = settings.securityMode;
+		setSettings((s) => ({ ...s, securityMode: mode }));
+		setIsUpdating(true);
+		try {
+			const hub = await connectionManager.getHub();
+			await hub.request('neo.updateSettings', { securityMode: mode });
+		} catch {
+			toast.error('Failed to update security mode');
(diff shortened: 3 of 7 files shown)


lsm (commented):

---

### P0: "App default" model selection always fails at runtime

**File:** `packages/web/src/components/settings/NeoSettings.tsx`, lines 98-105

When the user selects "App default" (the `''` empty-string option), the frontend converts it to `null` and sends `{ model: null }` via `neo.updateSettings`. However, the backend handler at `packages/daemon/src/lib/rpc-handlers/neo-handlers.ts:220-222` validates like this:

```ts
if (model !== undefined) {          // null !== undefined -> true, enters branch
    if (typeof model !== 'string' || model.trim() === '') {  // typeof null !== 'string' -> true
        throw new Error('model must be a non-empty string');  // ALWAYS THROWS
    }
}
```

Since `null` is not `undefined`, the guard passes. Since `typeof null !== 'string'`, validation fails and the handler throws. The user sees "Failed to update Neo model" toast and the select reverts. This means resetting the model to app-default is permanently broken.

**Fix options:**
- (a) The `neo.updateSettings` handler should accept `null` to mean "clear the neoModel override" (set it to `undefined` in the DB).
- (b) The frontend should omit the `model` field entirely when clearing (e.g., send `{ model: undefined }`), but then the server would need to handle that as "clear the override" rather than silently ignoring it.

### P1: Test gives false confidence on the broken "App default" flow

**File:** `packages/web/src/components/settings/__tests__/NeoSettings.test.tsx`, lines 190-213

The test "treats empty string model value as null when persisting" verifies that `{ model: null }` is sent to `neo.updateSettings`, but the mock resolves all non-getSettings calls with `{ success: true }` -- it never exercises the real server-side validation that rejects `null`. The test passes while the real flow fails. The mock should either replicate the server's validation behavior or the test should use a server-like mock that rejects `null` models.

### P2: Shared `isUpdating` state cross-disables independent controls

**File:** `packages/web/src/components/settings/NeoSettings.tsx`, lines 46, 86, 102

A single `isUpdating` boolean disables both the security mode and model selects. If a security mode update is slow, the model select is also disabled (and vice versa). These are independent RPC calls to the same endpoint, but the UI implication is confusing -- a user trying to change the model while security mode is saving will find it inexplicably disabled. Consider using separate loading states or debouncing the optimistic updates.

---

The wiring into `ContextPanel.tsx`, `MainContent.tsx`, and `signals.ts` is clean and follows existing patterns exactly. The error-revert pattern for failed updates is a nice touch. The confirmation dialog for Clear Session is appropriately simple. The navigation icon (lightbulb) is a reasonable choice. Once the P0 is resolved and the test is tightened up, this is ready.

lsm (commented):

---

This PR implements a Neo Settings section in Global Settings with security mode selector, model dropdown, and clear session functionality. The implementation is clean, well-tested (15 passing tests), and follows the existing patterns from other settings components. However, there's a **backend contract issue** that will cause the "App default" model selection to fail.

---

### P1: Backend `neo.updateSettings` rejects `model: null`

**File:** `packages/daemon/src/lib/rpc-handlers/neo-handlers.ts` (lines 220-224)

The frontend correctly sends `{ model: null }` when the user selects "App default" (test case on line 186 confirms this). However, the backend handler rejects this:

```ts
if (model !== undefined) {
    if (typeof model !== 'string' || model.trim() === '') {
        throw new Error('model must be a non-empty string');
    }
    updates.neoModel = model.trim();
}
```

Since `typeof null !== 'string'`, this throws an error when trying to clear the Neo model override. The handler should explicitly allow `null` to clear the setting:

```ts
if (model !== undefined) {
    if (model === null) {
        updates.neoModel = null;  // Clear the neoModel override
    } else if (typeof model !== 'string' || model.trim() === '') {
        throw new Error('model must be a non-empty string or null');
    } else {
        updates.neoModel = model.trim();
    }
}
```

Without this fix, users cannot reset the Neo model back to "App default" after selecting a specific model.

---

### P2: Missing success feedback for settings updates

**File:** `packages/web/src/components/settings/NeoSettings.tsx`

The "Clear Session" action shows a success toast (`Neo session cleared`), but changing security mode or model silently succeeds without any positive feedback. Consider adding success toasts for consistency with the clear session behavior and with other settings sections.

---

### P2: No visual loading state during initial settings fetch

**File:** `packages/web/src/components/settings/NeoSettings.tsx`

The `isLoading` state disables the select dropdowns during the initial `neo.getSettings` call, but there's no visual indicator (spinner, skeleton loader, "Loading..." text) to inform the user why the controls are disabled. This could be confusing on slow connections.

---

### Minor Observations (P3):

1. **Signal vs useState pattern:** Unlike `GeneralSettings` which reads directly from `globalSettings.value` (a signal), `NeoSettings` uses local `useState` and fetches data via RPC. This is acceptable since Neo settings aren't part of the global settings signal store, but worth noting for architectural consistency.

2. **Test coverage is solid:** 15 tests covering render, load, change handlers, RPC calls, error reversal, confirmation dialog, and both success/failure paths for clear session. Good job on comprehensive coverage.

---

### Summary

The frontend implementation is clean and well-tested. The blocking issue is the backend contract mismatch - the `neo.updateSettings` RPC handler needs to accept `null` as a valid model value to allow users to reset back to "App default".

**Recommendation:** REQUEST_CHANGES due to the P1 backend bug that prevents the "App default" model option from working.

lsm (commented):

---

This iteration successfully addresses all the feedback from the previous review. The fixes are well-implemented and properly tested.

---

### P1 Fixed: Backend now accepts `null` model

**File:** `packages/daemon/src/lib/rpc-handlers/neo-handlers.ts` (lines 220-228)

The handler now correctly accepts `null` to clear the neoModel override:

```ts
if (model !== undefined) {
    if (model === null) {
        updates.neoModel = null;  // Clear the override
    } else if (typeof model !== 'string' || model.trim() === '') {
        throw new Error('model must be a non-empty string or null');
    } else {
        updates.neoModel = model.trim();
    }
}
```

The daemon test `'clears neoModel override when model is null'` (line 409-415) verifies this behavior. Users can now reset the Neo model back to "App default" after selecting a specific model.

---

### P2 Fixed: Independent loading states for controls

**File:** `packages/web/src/components/settings/NeoSettings.tsx` (lines 46-47)

The single `isUpdating` state has been split into `isUpdatingMode` and `isUpdatingModel`, so each control uses its own loading state. The security mode and model selects no longer cross-disable each other during updates.

---

### P2 Fixed: Success toasts added

**File:** `packages/web/src/components/settings/NeoSettings.tsx` (lines 91, 109)

Success toasts are now shown after settings updates:
- `'Security mode updated'` after security mode changes
- `'Model updated'` after model changes

This is consistent with the clear session behavior and provides positive user feedback.

---

### P2 Fixed: Loading placeholder during initial fetch

**File:** `packages/web/src/components/settings/NeoSettings.tsx` (lines 137-143)

A loading placeholder is now shown while settings are being fetched:

```tsx
if (isLoading) {
    return (
        <SettingsSection title="Neo Agent">
            <div class="text-sm text-gray-500">Loading...</div>
        </SettingsSection>
    );
}
```

Users now see visual feedback instead of disabled dropdowns during the initial load.

---

### Tests Updated

**File:** `packages/web/src/components/settings/__tests__/NeoSettings.test.tsx`

- 17 tests now (up from 15): two new tests added
  - `'shows loading state before settings are fetched'` (lines 84-89)
  - `'model select stays enabled after a security mode update completes'` (lines 333-359)
- Existing tests updated to verify success toast calls
- Test names improved for clarity (e.g., "sends model: null when 'App default' is selected (clears override)")

---

### P3: Minor suggestion - Loading indicator could be more informative

**File:** `packages/web/src/components/settings/NeoSettings.tsx` (line 140)

The loading placeholder `Loading...` is functional but could be enhanced with a skeleton loader or animated dots to match the design system of other settings sections. This is purely cosmetic and not blocking.

---

### Summary

All P0/P1 and P2 issues from the previous review have been addressed:

1. **P0/P1 Backend null model**: Fixed - handler now accepts `null` with corresponding test
2. **P2 Cross-disabled controls**: Fixed - split into independent loading states with test
3. **P2 Success feedback**: Fixed - toasts added for both settings updates
4. **P2 Loading placeholder**: Fixed - `Loading...` text shown during initial fetch

The implementation is now complete, well-tested (17 tests), and ready for merge.

**Recommendation:** APPROVE

lsm (commented):

### Verdict: APPROVE

All previous P0/P1 issues from the first review have been addressed. The null-model clear path is properly handled in the backend with validation and test coverage, the per-field loading states prevent cross-disabling of controls, and the two new tests cover the remaining gaps. The implementation is clean and follows existing patterns.

---

**P2: `neo.getSettings` never returns `model: null` -- "App default" cannot be the initial selection**

`packages/daemon/src/lib/neo/neo-agent-manager.ts` line 222:
```ts
getModel(): string {
  const settings = this.settingsManager.getGlobalSettings();
  return settings.neoModel ?? settings.model ?? 'sonnet';
}
```
`getModel()` always resolves to a concrete string (never null). The `neo.getSettings` RPC handler at `packages/daemon/src/lib/rpc-handlers/neo-handlers.ts` line 189 calls this method, so the response always returns a resolved model. The frontend (`NeoSettings.tsx`) expects `model: string | null` where null means "app default", but can never receive null from the backend.

Effect: if the user's app default is 'sonnet' and they have no neo override, the dropdown shows "Claude Sonnet 4" (not "App default"). The user cannot distinguish "I chose sonnet for Neo" from "no override set". Selecting "App default" correctly clears the override, but on next page load it will show whatever the resolved default model is, not "App default".

Fix: either (a) change `neo.getSettings` to return the raw `settings.neoModel` field (null when no override), or (b) change `getModel()` to a variant that returns nullable.

**P3: `description` field on `SECURITY_MODE_OPTIONS` is unused by `SettingsSelect`**

`packages/web/src/components/settings/NeoSettings.tsx` line 9-27 defines `SECURITY_MODE_OPTIONS` with a `description` field. The `SettingsSelect` component's options type is `Array<{ value: string; label: string }>` (`packages/web/src/components/settings/SettingsSection.tsx` line 40), so the description is structurally ignored by the select. It IS used for the `SettingsRow` description via `selectedMode?.description` (line 150), so this works correctly, but the type mismatch means TypeScript would flag it if options were typed strictly. Consider adding an optional `description` field to `SettingsSelectProps.options` or using a separate lookup.

**P3: Minor type mismatch -- `neoModel: null` vs `neoModel?: string`**

`packages/shared/src/types/settings.ts` line 176 types `neoModel` as `string | undefined`. The handler at `packages/daemon/src/lib/rpc-handlers/neo-handlers.ts` line 223 passes `{ neoModel: null }` through `updateGlobalSettings`. The cast on line 236 (`updates as Parameters<typeof settingsManager.updateGlobalSettings>[0]`) hides the type mismatch. Works at runtime because JSON stores null and nullish coalescing treats it as nullish, but worth noting the type technically allows only `undefined` not `null`.

---

### Summary

Clean implementation that correctly addresses all feedback from the first review. The P2 about `getModel()` resolving the override before returning is the main semantic correctness gap, but it does not break functionality (the clear-override round-trip works, and the correct model is always used). The tests are thorough with good coverage of happy paths, error paths, and the per-field loading isolation.

PR #1130, opened 2026-03-28
test(e2e): Neo panel basic interaction tests (task 11.1)

lsm:
Tests the core Neo panel interaction flow via Playwright E2E.

**8 tests covering:**
- NavRail Neo button visible and clickable
- Opening panel focuses the chat input
- Chat tab active by default
- Switching between Chat and Activity tabs
- Close button dismisses panel
- Backdrop click dismisses panel
- `localStorage` state persists across page reload
- `Cmd+J` / `Ctrl+J` keyboard shortcut toggles panel

### DIFF
--- a/packages/e2e/tests/features/neo-panel.e2e.ts
+++ b/packages/e2e/tests/features/neo-panel.e2e.ts
@@ -2,15 +2,15 @@
  * Neo Panel E2E Tests
  *
  * Tests for the core Neo panel interaction flow:
- * - NavRail Neo button visibility and clickability
- * - Panel open/close behavior
+ * - NavRail Neo button visibility, clickability, and active state
+ * - Panel open/close behavior (button, close button, backdrop, Escape key)
  * - Tab switching (Chat / Activity)
- * - Dismiss via close button and backdrop click
- * - localStorage state persistence across navigation
+ * - localStorage state persistence across page reload
  * - Cmd+J / Ctrl+J keyboard shortcut toggle
  */
 
 import { test, expect } from '../../fixtures';
+import type { Page } from '@playwright/test';
 import { waitForWebSocketConnected } from '../helpers/wait-helpers';
 
 // ---------------------------------------------------------------------------
@@ -18,36 +18,29 @@ import { waitForWebSocketConnected } from '../helpers/wait-helpers';
 // ---------------------------------------------------------------------------
 
 /** Locate the Neo NavRail button by its aria-label */
-function getNeoNavButton(page: Parameters<typeof waitForWebSocketConnected>[0]) {
+function getNeoNavButton(page: Page) {
 	return page.locator('button[aria-label="Neo (⌘J)"]');
 }
 
 /** Locate the Neo panel container */
-function getNeoPanel(page: Parameters<typeof waitForWebSocketConnected>[0]) {
+function getNeoPanel(page: Page) {
 	return page.locator('[data-testid="neo-panel"]');
 }
 
 /** Locate the Neo panel backdrop */
-function getNeoBackdrop(page: Parameters<typeof waitForWebSocketConnected>[0]) {
+function getNeoBackdrop(page: Page) {
 	return page.locator('[data-testid="neo-panel-backdrop"]');
 }
 
-/** Locate the Neo chat input */
-function getNeoChatInput(page: Parameters<typeof waitForWebSocketConnected>[0]) {
-	return page.locator('[data-testid="neo-chat-input"]');
-}
-
 /** Wait for the panel to be visible (translated into view) */
-async function waitForPanelOpen(page: Parameters<typeof waitForWebSocketConnected>[0]) {
-	const panel = getNeoPanel(page);
-	// Panel is open when translate-x-0 class is applied (no negative transform)
-	await expect(panel).not.toHaveClass(/\-translate-x-full/, { timeout: 3000 });
+async function waitForPanelOpen(page: Page) {
+	// Panel is open when -translate-x-full class is NOT applied
+	await expect(getNeoPanel(page)).not.toHaveClass(/-translate-x-full/, { timeout: 3000 });
 }
 
 /** Wait for the panel to be hidden (translated out of view) */
-async function waitForPanelClosed(page: Parameters<typeof waitForWebSocketConnected>[0]) {
-	const panel = getNeoPanel(page);
-	await expect(panel).toHaveClass(/-translate-x-full/, { timeout: 3000 });
+async function waitForPanelClosed(page: Page) {
+	await expect(getNeoPanel(page)).toHaveClass(/-translate-x-full/, { timeout: 3000 });
 }
 
 // ---------------------------------------------------------------------------
@@ -58,36 +51,47 @@ test.describe('Neo Panel — Core Interaction', () => {
 	test.use({ viewport: { width: 1280, height: 720 } });
 
 	test.beforeEach(async ({ page }) => {
-		// Clear persisted panel state so tests start with a closed panel
+		// Clear persisted panel state before the page loads so tests always start with a closed panel.
+		// addInitScript runs before any page script, avoiding the extra reload.
+		await page.addInitScript(() => {
+			localStorage.removeItem('neo:panelOpen');
+		});
 		await page.goto('/');
-		await page.evaluate(() => localStorage.removeItem('neo:panelOpen'));
-		// Reload to apply cleared localStorage before WS connects
-		await page.reload();
 		await waitForWebSocketConnected(page);
 	});
 
 	// ── 1. NavRail Neo button visible ──────────────────────────────────────
 
 	test('NavRail Neo button is visible', async ({ page }) => {
-		const btn = getNeoNavButton(page);
-		await expect(btn).toBeVisible({ timeout: 5000 });
+		await expect(getNeoNavButton(page)).toBeVisible({ timeout: 5000 });
 	});
 
-	// ── 2. Clicking button opens panel and focuses chat input ──────────────
+	// ── 2. NavRail button shows active state when panel is open ───────────
 
-	test('clicking Neo button opens the panel and focuses the chat input', async ({ page }) => {
+	test('NavRail Neo button shows active state (aria-pressed) when panel is open', async ({
+		page,
+	}) => {
 		const btn = getNeoNavButton(page);
-		await btn.click();
+		await expect(btn).toHaveAttribute('aria-pressed', 'false');
 
+		await btn.click();
+		await waitForPanelOpen(page);
+		await expect(btn).toHaveAttribute('aria-pressed', 'true');
+	});
+
+	// ── 3. Clicking button opens panel ────────────────────────────────────
+
+	test('clicking Neo button opens the panel and focuses the close button', async ({ page }) => {
+		await getNeoNavButton(page).click();
 		await waitForPanelOpen(page);
 
-		// Chat input should receive focus after opening
-		const input = getNeoChatInput(page);
-		await expect(input).toBeVisible({ timeout: 3000 });
-		await expect(input).toBeFocused({ timeout: 3000 });
+		// NeoPanel.tsx focuses the close button via requestAnimationFrame for accessibility
+		const closeBtn = page.locator('[data-testid="neo-panel-close"]');
+		await expect(closeBtn).toBeVisible({ timeout: 3000 });
+		await expect(closeBtn).toBeFocused({ timeout: 3000 });
 	});
 
-	// ── 3. Chat tab active by default ─────────────────────────────────────
+	// ── 4. Chat tab active by default ─────────────────────────────────────
 
 	test('Neo panel displays with Chat tab active by default', async ({ page }) => {
 		await getNeoNavButton(page).click();
@@ -104,7 +108,7 @@ test.describe('Neo Panel — Core Interaction', () => {
 		await expect(page.locator('[data-testid="neo-chat-view"]')).toBeVisible({ timeout: 3000 });
 	});
 
-	// ── 4. Tab switching ───────────────────────────────────────────────────
+	// ── 5. Tab switching ───────────────────────────────────────────────────
 
 	test('can switch between Chat and Activity tabs', async ({ page }) => {
 		await getNeoNavButton(page).click();
@@ -118,7 +122,7 @@ test.describe('Neo Panel — Core Interaction', () => {
 			'aria-selected',
 			'false'
 		);
-		// Chat view hidden, activity view shown
+		// Chat view is conditionally rendered — not present in DOM when Activity is active
 		await expect(page.locator('[data-testid="neo-chat-view"]')).not.toBeVisible({ timeout: 2000 });
 
 		// Switch back to Chat tab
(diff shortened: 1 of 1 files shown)


lsm (commented):

---

### P1: Test 2 — `chat input is focused` assertion likely fails against actual component behavior

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts`, line 87

The test asserts `await expect(input).toBeFocused({ timeout: 3000 })` after opening the panel, expecting the chat textarea to receive focus. However, the actual component behavior contradicts this:

1. `NeoPanel.tsx` lines 86–98: When `isOpen` changes to true, the focus management effect saves the trigger element and then calls `closeButtonRef.current?.focus()` inside a `requestAnimationFrame` callback. This explicitly focuses the **close button**, not the chat input.

2. `NeoChatView.tsx` lines 371–373: The `inputRef.current?.focus()` call runs only on mount (`[]` dependency), not when the panel opens. Since `NeoChatView` is already mounted when the panel is closed (the panel is always in the DOM, just translated off-screen), this effect already fired at initial page load — long before the panel was opened.

The result is that after opening the panel, the **close button** receives focus (via rAF), not the chat input. The `toBeFocused` assertion on the chat input will either consistently fail or be a timing-dependent flake depending on how Playwright's polling interacts with the rAF callback.

**Fix:** Either change the assertion to verify the close button is focused (`page.locator('[data-testid="neo-panel-close"]')`), or adjust the component's focus management to focus the chat input when the panel opens.

---

### P2: Test 7 — `localStorage.getItem()` assertions violate the E2E "visible DOM state" principle

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts`, lines 166–167 and 180–181

The E2E rules state: "All assertions must verify visible DOM state: text content, element visibility, CSS classes." The test reads `localStorage.getItem('neo:panelOpen')` directly and asserts on the raw value:

```ts
const stored = await page.evaluate(() => localStorage.getItem('neo:panelOpen'));
expect(stored).toBe('true');
```

While `localStorage` is not in the explicit prohibition list (which covers `window.sessionStore`, `window.globalStore`, etc.), this pattern does not verify visible DOM state. The test already verifies the panel's open/closed state via `waitForPanelOpen()` / `waitForPanelClosed()` which check CSS classes — that is sufficient. The localStorage assertions are redundant and set a precedent that could lead to more internal-state assertions creeping in.

**Note:** `task-view-v2.e2e.ts` has the same pattern (line 169–170), so this is a pre-existing convention rather than a new violation. Consider addressing both files together.

**Fix:** Remove the two `localStorage.getItem` assertions (lines 166–167 and 180–181). The DOM-based assertions already cover the persistence behavior.

---

### P3: Minor — helper parameter type is unnecessarily verbose

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts`, lines 21–38

All five helper functions use `Parameters<typeof waitForWebSocketConnected>[0]` as the parameter type for the page argument. This is equivalent to `Page` from Playwright and harder to read. Consider importing `Page` directly:

```ts
import type { Page } from '@playwright/test';
```

This would make the signatures cleaner: `function getNeoNavButton(page: Page)`.

---

### Overall Assessment

The test file is well-structured, follows the project's E2E conventions (viewport configuration, `beforeEach` cleanup, helper extraction, proper use of `data-testid` selectors), and covers the core interaction flows thoroughly. The main concern is test 2's focus assertion which likely doesn't match the component's actual behavior and could be a source of CI flakiness.

lsm (commented):

### Summary

The test file is well-structured and follows project E2E conventions, but **2 of 8 tests are currently failing** due to mismatches between test expectations and actual component behavior.

---

### P1 Issues (Should Fix)

#### 1. Test "clicking Neo button opens the panel and focuses the chat input" fails

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts:78-88`

The test expects the chat input to be focused when the panel opens:
```ts
await expect(input).toBeFocused({ timeout: 3000 });
```

However, `NeoPanel.tsx` explicitly focuses the **close button** on open (line 90):
```ts
requestAnimationFrame(() => {
  closeButtonRef.current?.focus();
});
```

Meanwhile, `NeoChatView.tsx` also calls `inputRef.current?.focus()` on mount (line 373). This creates a race condition where focus lands on the close button first, then potentially moves to the chat input.

**Recommendation:** Either:
- Update the test to match current behavior (focus on close button is correct for accessibility), OR
- Update `NeoPanel.tsx` to focus the chat input instead of the close button if that's the intended UX

---

#### 2. Test "Cmd+J / Ctrl+J keyboard shortcut toggles the Neo panel" fails

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts:192-207`

The keyboard shortcut handler in `useNeoKeyboardShortcut.ts` (line 22) guards against firing when focus is in an INPUT, TEXTAREA, or contentEditable element:
```ts
if (tag === 'INPUT' || tag === 'TEXTAREA' || target.isContentEditable) return;
```

After `beforeEach` reloads the page, focus may land on an input element (e.g., a search box or text area on the lobby), causing the shortcut to be silently ignored.

**Recommendation:** Before pressing the keyboard shortcut, ensure focus is on a non-input element:
```ts
// Move focus to body to ensure shortcut is not blocked
await page.locator('body').click();
await page.keyboard.press(`${modifier}+j`);
```

---

#### 3. Missing test for Escape key behavior

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts`

`NeoPanel.tsx` implements Escape key to close the panel (lines 104-106):
```ts
if (e.key === 'Escape') {
  neoStore.closePanel();
  return;
}
```

This is a core interaction that should have E2E coverage. Consider adding:
```ts
test('Escape key dismisses the Neo panel', async ({ page }) => {
  await getNeoNavButton(page).click();
  await waitForPanelOpen(page);
  await page.keyboard.press('Escape');
  await waitForPanelClosed(page);
});
```

---

#### 4. Missing test for NavRail button active state

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts`

When the panel is open, the NavRail button should show `aria-pressed="true"` (see `NavIconButton.tsx` line 29). This visual feedback should be tested:
```ts
test('NavRail Neo button shows active state when panel is open', async ({ page }) => {
  const btn = getNeoNavButton(page);
  await expect(btn).toHaveAttribute('aria-pressed', 'false');
  await btn.click();
  await waitForPanelOpen(page);
  await expect(btn).toHaveAttribute('aria-pressed', 'true');
});
```

---

### P3 Issues (Nits)

#### 5. Verbose helper function type parameter

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts:21`

Using `Parameters<typeof waitForWebSocketConnected>[0]` is clever but harder to read than just `Page`:
```ts
// Current
function getNeoNavButton(page: Parameters<typeof waitForWebSocketConnected>[0]) {
// Suggested
function getNeoNavButton(page: Page) {
```

---

#### 6. beforeEach could be simplified

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts:60-67`

The pattern `goto('/') `-> `localStorage.removeItem` -> `reload()` could be simplified using `page.addInitScript()` to clear localStorage before the page loads, avoiding the extra reload:
```ts
test.beforeEach(async ({ page }) => {
  await page.addInitScript(() => {
    localStorage.removeItem('neo:panelOpen');
  });
  await page.goto('/');
  await waitForWebSocketConnected(page);
});
```

---

### Test Results

Running `make run-e2e TEST=tests/features/neo-panel.e2e.ts`:
- **6 passed**
- **2 failed** (focus test, keyboard shortcut test)

---

### Verdict

**RECOMMENDATION: REQUEST_CHANGES**

The test structure is good and follows E2E conventions (UI-only actions/assertions). However, the failing tests need to be fixed before merge. Please address the P1 issues to align test expectations with actual component behavior.

lsm (commented):

---

### Summary

Good fixes on the previous P1/P2 issues. The focus assertion now correctly targets the close button, the body-click workaround for Cmd+J is solid, the E2E rules are respected (no more `localStorage.getItem()` in assertions), and the two new tests (Escape dismiss, aria-pressed active state) are well-scoped.

However, the `addInitScript` approach introduced in `beforeEach` contains a **P0 bug** that breaks the persistence test (test #9).

---

### P0: `addInitScript` re-runs on `page.reload()`, breaking persistence test

**File:** `packages/e2e/tests/features/neo-panel.e2e.ts`, lines 56-58 and 179

The comment on line 179 states:
```
// Reload (addInitScript does NOT run on reload, so localStorage survives)
```

This is **incorrect**. Per the [Playwright docs](https://playwright.dev/docs/api/class-page#page-add-init-script), `addInitScript` is evaluated "whenever the page is navigated" -- and `page.reload()` IS a navigation. The registered init script will re-execute on every reload, calling `localStorage.removeItem('neo:panelOpen')` before the app code runs.

The execution order during `page.reload()` in test #9 is:
1. `page.reload()` triggers
2. `addInitScript` runs first: `localStorage.removeItem('neo:panelOpen')` -- clears the `'true'` value
3. App code runs: `neoStore._readPanelOpen()` reads `localStorage.getItem('neo:panelOpen')` which is now `null` -> returns `false`
4. Panel renders closed

This means `waitForPanelOpen(page)` on line 184 should **time out and fail** because the panel will start closed after reload, not open. The persistence test is testing the wrong thing -- it would only pass if there were a race condition or if `addInitScript` behaved differently than documented.

**Fix:** Either:
- (a) Remove the `addInitScript` and go back to the previous approach of `page.evaluate(() => localStorage.removeItem(...))` after `page.goto('/')`, then `page.reload()` so the cleared state takes effect before the persistence test's reloads. The `beforeEach` would be: goto -> evaluate clear -> reload -> wait for WS.
- (b) Keep `addInitScript` for the beforeEach cleanup, but in the persistence test, remove the init script before the reload: `await page.removeInitScript(script)` before the first `page.reload()` on line 180. However, `page.removeInitScript()` doesn't exist in Playwright -- you'd need to use a conditional inside the init script itself (e.g., check a `window.__skipClear` flag that the test sets before reload).
- (c) The cleanest approach: use `page.addInitScript` with a flag that the persistence test can control:
  ```ts
  await page.addInitScript((clear: boolean) => {
    if (clear) localStorage.removeItem('neo:panelOpen');
  }, true); // default: clear
  ```
  Then in the persistence test, override with a new init script before reload:
  ```ts
  await page.addInitScript(() => {}); // no-op, overrides the clearing one
  await page.reload();
  ```

Note: Option (a) is the simplest and most straightforward. It's essentially what the original code did before the refactor.

---

### P3 (nit): PR description body is outdated

The PR description still says "8 tests covering" but there are now 10 tests. This is cosmetic and can be updated before merge.

---

### Verdict

The P0 on the persistence test is blocking. The `addInitScript` approach is a good optimization for the beforeEach cleanup, but it cannot coexist with a test that verifies localStorage persistence across reloads unless the init script is explicitly disabled for that test.

lsm (commented):

All previous issues have been addressed. The updated test file is well-structured and follows E2E guidelines properly.

**Resolved Issues:**

1. **P1 Focus assertion race condition** - Fixed. Test now correctly asserts the close button is focused (line 91), matching the actual component behavior in NeoPanel.tsx which uses `requestAnimationFrame` to focus the close button on open.

2. **P1 Cmd+J keyboard shortcut blocked by input focus guard** - Fixed. Added `await page.locator('body').click();` before pressing the shortcut (lines 206, 216) to ensure focus is not on an input element, bypassing the guard in useNeoKeyboardShortcut.ts.

3. **P2 localStorage.getItem() assertion** - Fixed. Removed direct localStorage access; now uses DOM-based assertions via `waitForPanelOpen()` and `waitForPanelClosed()` helpers that check CSS classes.

4. **P2 Missing Escape key dismiss test** - Fixed. Added test at lines 148-156.

5. **P2 Missing NavRail button active state test** - Fixed. Added test at lines 69-80 verifying the `aria-pressed` attribute on the NavRail button.

**Positive Observations:**

- Test count increased from 8 to 10 as expected
- All tests follow E2E guidelines: actions go through UI, assertions verify DOM state
- The `addInitScript` approach for clearing localStorage (line 56) is clean and avoids an extra reload
- The helper functions (`getNeoNavButton`, `getNeoPanel`, `waitForPanelOpen`, `waitForPanelClosed`) are well-organized and reusable
- Focus management tests correctly account for `requestAnimationFrame` timing with `{ timeout: 3000 }` on `toBeFocused()`

**Minor Observation (P3 - nit):**

The comment on line 215-216 is slightly misleading: "body has focus; panel close button had it — click body first". After opening the panel via Cmd+J, the close button gets focused via rAF, but the keyboard event handler in useNeoKeyboardShortcut.ts only checks if the event target is an input/textarea/contentEditable — it does not block based on focus being on the close button. The body click is still good for safety, but the comment could be simplified.

**Recommendation: APPROVE**

The tests comprehensively cover the core Neo panel interaction flow and properly validate the component behavior.

lsm (commented):

### P0 Fix Verification: Correct

The fix in commit 69dac454b correctly resolves the localStorage clearing issue.

**Root cause (previous):** `page.addInitScript()` registers a script that re-executes on *every* page load including `page.reload()`. When test #9 (persistence) called `page.reload()`, the init script silently cleared `localStorage('neo:panelOpen')`, making the persistence assertion a false positive.

**Fix (current):** `beforeEach` now:
1. Navigates to `/` (makes localStorage accessible)
2. Calls `page.evaluate(() => localStorage.removeItem('neo:panelOpen'))` -- a one-shot imperative call, not a registered script
3. Calls `page.reload()` so the app initializes with the cleared value
4. Waits for WebSocket connection

Since `page.evaluate()` is not re-registered on reload, subsequent `page.reload()` calls inside test #9 no longer clear localStorage. The persistence test will now correctly validate that panel state survives reload.

### P3 (nit): Stale comment on line 181

`packages/e2e/tests/features/neo-panel.e2e.ts`, line 181:
```
// Reload (addInitScript does NOT run on reload, so localStorage survives)
```
This comment says `addInitScript` does NOT run on reload -- but that is factually incorrect (it DOES run on reload, which was the whole bug). Consider updating to something like:
```
// Reload (localStorage survives because we cleared it via page.evaluate, not addInitScript)
```

### Overall Assessment

The fix is clean and minimal. No other P0/P1/P2 issues found. The persistence test changes (renamed test, switched from `waitForMessageProcessed` to `waitForMessageSent`, removed title assertion) are reasonable improvements that decouple from LLM response timing.

**Recommendation: APPROVE**

lsm (commented):

### P0 Issue Resolution — Verified ✓

The `addInitScript` → `page.evaluate()` fix correctly resolves the localStorage persistence test bug.

**Root cause:** `addInitScript` executes on every navigation/reload, wiping `neo:panelOpen` before the persistence test could verify it survived `page.reload()`.

**Fix:** The `beforeEach` now:
1. Navigates to `/` first (so localStorage is accessible)
2. Clears `neo:panelOpen` via `page.evaluate(() => localStorage.removeItem(...))`
3. Reloads once to initialize the app with clean state

This ensures localStorage is cleared exactly once at test setup. Subsequent `page.reload()` calls inside tests (persistence test at lines 182, 193) do NOT trigger re-clearing.

### Test Coverage Review

| Test | Status |
|------|--------|
| NavRail button visible | ✓ |
| Active state (aria-pressed) toggle | ✓ |
| Panel opens + focuses close button | ✓ |
| Chat tab active by default | ✓ |
| Tab switching Chat ↔ Activity | ✓ |
| Close button dismisses panel | ✓ |
| Escape key dismisses panel | ✓ |
| Backdrop click dismisses panel | ✓ |
| localStorage persistence across reload | ✓ |
| Cmd+J / Ctrl+J keyboard shortcut | ✓ |

All 10 tests cover the core interaction flow. DOM-based assertions are used throughout (no internal state access). The test file follows E2E guidelines correctly.

### Recommendation: APPROVE

No remaining P0/P1/P2 issues. The fix is correct and the test suite is well-structured.

PR #1132, opened 2026-03-28
feat(neo): Space and workflow action tools (task 3.3)

lsm:
Adds 7 new MCP action tools to `neo-action-tools.ts` for space and workflow write operations.

## Tools added

**Space operations** (low/medium risk, delegated to GlobalSpaces handler layer):
- `create_space` — low risk, auto-executes in balanced mode
- `update_space` — low risk, auto-executes in balanced mode  
- `delete_space` — medium risk, requires confirmation in balanced mode

**Workflow operations**:
- `start_workflow_run` — low risk, delegated to GlobalSpaces handlers
- `cancel_workflow_run` — medium risk, cancels active tasks then transitions run to cancelled
- `approve_gate` — medium risk, idempotent, transitions rejected runs back to in_progress
- `reject_gate` — medium risk, idempotent, transitions run to needs_attention with humanRejected reason

## Design

- `NeoActionSpaceHandlers` interface — thin wrapper the caller fills with `createGlobalSpacesToolHandlers()` output, so Neo delegates without duplicating business logic
- `NeoActionWorkflowRunRepository` + `NeoActionGateDataRepository` — minimal interfaces for cancel/gate operations
- `onGateChanged` optional callback triggers channel re-evaluation after approval (mirrors `fireGateChanged` in RPC layer)
- All tools go through `withSecurityCheck()` for security tier enforcement

## Tests

43 new unit tests (115 total): happy paths, idempotency, security tier enforcement (balanced/conservative/autonomous), error cases, dependency-unavailable guards, and MCP server registration check (11 → 18 tools).

### DIFF
--- a/packages/daemon/src/lib/neo/tools/neo-action-tools.ts
+++ b/packages/daemon/src/lib/neo/tools/neo-action-tools.ts
@@ -279,6 +279,28 @@ export interface NeoActionToolsConfig {
 	 * activate if the gate is now open (mirrors fireGateChanged in the RPC layer).
 	 */
 	onGateChanged?: NeoGateChangedNotifier;
+	/**
+	 * Optional callback invoked after a workflow run's status changes.
+	 * Mirrors the `space.workflowRun.updated` event emitted by the RPC layer
+	 * so that the frontend LiveQuery subscriptions receive the update.
+	 * Called by cancel_workflow_run, approve_gate, and reject_gate.
+	 */
+	onWorkflowRunUpdated?: (
+		spaceId: string,
+		runId: string,
+		run: NeoWorkflowRun
+	) => void | Promise<void>;
+	/**
+	 * Optional callback invoked after gate data is written.
+	 * Mirrors the `space.gateData.updated` event emitted by the RPC layer.
+	 * Called by approve_gate and reject_gate.
+	 */
+	onGateDataUpdated?: (
+		spaceId: string,
+		runId: string,
+		gateId: string,
+		data: Record<string, unknown>
+	) => void | Promise<void>;
 }
 
 // ---------------------------------------------------------------------------
@@ -355,6 +377,8 @@ export function createNeoActionToolHandlers(config: NeoActionToolsConfig) {
 		spaceTaskManagerFactory,
 		gateDataRepo,
 		onGateChanged,
+		onWorkflowRunUpdated,
+		onGateDataUpdated,
 	} = config;
 
 	return {
@@ -752,12 +776,12 @@ export function createNeoActionToolHandlers(config: NeoActionToolsConfig) {
 			instructions?: string;
 			autonomy_level?: SpaceAutonomyLevel;
 		}): Promise<ToolResult> {
-			if (!spaceHandlers) {
-				return errorResult('Space operations are not available');
-			}
-			return withSecurityCheck('create_space', args as Record<string, unknown>, config, () =>
-				spaceHandlers.create_space(args)
-			);
+			return withSecurityCheck('create_space', args as Record<string, unknown>, config, () => {
+				if (!spaceHandlers) {
+					return Promise.resolve(errorResult('Space operations are not available'));
+				}
+				return spaceHandlers.create_space(args);
+			});
 		},
 
 		async update_space(args: {
@@ -769,9 +793,7 @@ export function createNeoActionToolHandlers(config: NeoActionToolsConfig) {
 			default_model?: string;
 			autonomy_level?: SpaceAutonomyLevel;
 		}): Promise<ToolResult> {
-			if (!spaceHandlers) {
-				return errorResult('Space operations are not available');
-			}
+			// Input validation — stays outside the security check (no backend needed to validate)
 			const hasUpdates =
 				args.name !== undefined ||
 				args.description !== undefined ||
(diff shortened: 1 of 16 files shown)


lsm (commented):

---

Overall this is a clean, well-structured addition that follows the established two-layer pattern (handlers + MCP wrapper) from the existing codebase. The gate approval/rejection logic faithfully mirrors the RPC handler in `space-workflow-run-handlers.ts`, and the security tier enforcement is correctly applied via `withSecurityCheck()`. All 115 tests pass.

I found one P1 issue and a few minor observations.

### P1: Missing event emission after state transitions

The RPC layer (`space-workflow-run-handlers.ts`) emits `space.workflowRun.updated` and `space.gateData.updated` events after every state mutation (approve, reject, cancel). These events drive frontend reactivity via LiveQuery subscriptions. The Neo action tools bypass the RPC layer entirely, performing direct repository mutations, but **do not emit any events**.

This means:
- `cancel_workflow_run` transitions the run to `cancelled` but the UI will not update until the user navigates away and back (or a LiveQuery poll cycle catches it, if one exists for workflow runs).
- `approve_gate` transitions `needs_attention -> in_progress` but the UI will not reflect the status change.
- `reject_gate` transitions to `needs_attention` but the UI will not show the rejection.

The `onGateChanged` callback partially addresses this for channel re-evaluation, but it is optional and only covers the channel router -- not the broader UI state.

**Recommendation**: Either (a) add an optional `emit` callback to `NeoActionToolsConfig` that mirrors `daemonHub.emit`, or (b) document that Neo action tools intentionally skip event emission and rely on eventual consistency from LiveQuery polling. If option (b), verify that LiveQuery actually polls the relevant tables.

### P2: No test for cancelTask failure (best-effort path)

At line 841 of `neo-action-tools.ts`, `cancelTask` errors are silently swallowed with a comment `/* best-effort */`. There is no test verifying that a failed `cancelTask` does not prevent the run from being cancelled. Given the best-effort semantics, this should be tested to confirm the run still transitions to `cancelled` even when some task cancellations fail.

### P2: Dependency-unavailable guards bypass security check

For `create_space`, `delete_space`, `start_workflow_run`, `cancel_workflow_run`, `approve_gate`, and `reject_gate`, the dependency-unavailable check (e.g., `if (!spaceHandlers)`) runs **before** `withSecurityCheck()`. This means in conservative mode, the unavailable error is returned immediately without storing a pending action -- which is arguably correct (you can't confirm an action that has no backend). However, this is inconsistent with the pattern in `approve_task`/ `reject_task` where the runtime unavailability check is **inside** the security check executor. Consider aligning the approach -- either all guards outside, or all inside.

### P3: `beforeEach` imported but unused in test file

Line 30 imports `beforeEach` from `bun:test`, but it is only used in the MCP server describe block. Not a problem since the import is used, just noting for awareness.

1. **Idempotency correctness**: Both `approve_gate` and `reject_gate` correctly skip `onGateChanged` and status transitions on idempotent (already-approved/already-rejected) paths, matching the RPC handler behavior exactly.
2. **Security tier mappings are correct**: `create_space` (low), `update_space` (low), `delete_space` (medium), `start_workflow_run` (low), `cancel_workflow_run` (medium), `approve_gate` (medium), `reject_gate` (medium) -- all align with the `ActionClassification` table in `security-tier.ts`.
3. **`reject_gate` not calling `onGateChanged`**: Correctly mirrors the RPC handler which also skips `fireGateChanged` on rejection.
4. **Test coverage is thorough**: 43 new tests covering happy paths, idempotency, error cases, security tier enforcement across all three modes, and dependency-unavailable guards.

lsm (commented):

Reviewed the fix commit (6341a49) and the full action-tools implementation. The three requested changes are all correctly implemented:

**P1 fix -- Event emission callbacks:** `onWorkflowRunUpdated` is called after status transitions in `cancel_workflow_run`, `approve_gate` (only when overriding a prior rejection), and `reject_gate`. `onGateDataUpdated` is called by both `approve_gate` and `reject_gate` after writing gate data. The callback signatures in `NeoActionToolsConfig` are well-typed with `void | Promise<void>` return types. The conditional guard on `approve_gate` (only fires `onWorkflowRunUpdated` inside the `if (run.status === 'needs_attention' && run.failureReason === 'humanRejected')` block) is correct -- it avoids a spurious event when the run status doesn't actually change. Test coverage for these new callbacks is thorough (5 new tests).

**P2 fix -- Dependency guards moved inside executors:** All 7 new tools now place their `!spaceHandlers` / `!workflowRunRepo` / `!gateDataRepo` checks inside the `withSecurityCheck` executor, consistent with the `approve_task` / `reject_task` pattern. Input validation (e.g., `!hasUpdates` in `update_space`, `!args.feedback?.trim()` in `reject_task`) correctly remains outside the security check. This ensures that in conservative mode, a dependency-unavailable scenario returns an error after confirmation rather than silently storing a pending action that will fail when executed.

**P2 fix -- Best-effort cancelTask failure test:** The new test verifies that when all `cancelTask` calls throw, the run still transitions to `cancelled`. This explicitly documents the intentional `catch(() => {})` swallow pattern.

**One P2 observation (non-blocking):**

In `reject_gate` (line 983-987), when the run is already in `needs_attention` status (but from a different reason, not `humanRejected`), the handler skips `transitionStatus` but still calls `updateRun` with `failureReason: 'humanRejected'` and fires `onWorkflowRunUpdated`. This overwrites the prior failure reason. If this is intentional (human rejection takes precedence), consider adding a brief comment. If not, it may be worth guarding with a check on the existing failure reason.

**One P3 nit:**

The comment on the `catch` in `cancel_workflow_run` (line 872-874) could reference the test that verifies this behavior, similar to how some codebases link tests to edge-case comments. This makes the intent more discoverable.

All 120 tests pass, lint is clean, and typecheck succeeds.

PR #1135, opened 2026-03-28
feat(neo): Attach action tools to Neo session (task 3.5)

lsm:
Wire the neo-action MCP server into Neo's provisioning alongside the existing neo-query server.

## Changes

- **`neo-tools-server.ts`** (new) — `createNeoToolsMcpServers(queryConfig, actionConfig?)` returns a named map of both in-process servers, ready for `session.setRuntimeMcpServers()`
- **`neo-agent-manager.ts`** — adds `setActionToolsConfig()` method; renames `attachQueryTools()` → `attachTools()` which creates both servers when action config is present; in-process servers always win on registry name collision
- **`neo-system-prompt.ts`** — expands Capabilities section with specific tool names grouped by risk classification (low/medium/high)
- **`rpc-handlers/index.ts`** — builds `NeoActionToolsConfig` after `spaceRuntimeService.start()`, wiring space handlers, MCP/skills/settings/session manager adapters, and daemonHub event callbacks; calls `setActionToolsConfig()` before `provision()`
- **`neo-agent-manager-all-tools.test.ts`** (new) — 10 integration tests covering both-server attach, query-only fallback, no-op without query config, registry merge, collision precedence, health recovery, clearSession, and config overwrite

### DIFF
--- a/packages/daemon/src/lib/neo/neo-agent-manager.ts
+++ b/packages/daemon/src/lib/neo/neo-agent-manager.ts
@@ -19,7 +19,11 @@ import type { McpServerConfig } from '@neokai/shared';
 import type { AgentSession } from '../agent/agent-session';
 import { Logger } from '../logger';
 import { buildNeoSystemPrompt, type NeoSecurityMode } from './neo-system-prompt';
-import { createNeoQueryMcpServer, type NeoToolsConfig } from './tools/neo-query-tools';
+import {
+	createNeoToolsMcpServers,
+	type NeoToolsConfig,
+	type NeoActionToolsConfig,
+} from './tools/neo-tools-server';
 
 export const NEO_SESSION_ID = 'neo:global';
 const NEO_SESSION_TITLE = 'Neo';
@@ -52,6 +56,7 @@ export class NeoAgentManager {
 	private readonly logger = new Logger('NeoAgentManager');
 	private session: AgentSession | null = null;
 	private toolsConfig: NeoToolsConfig | null = null;
+	private actionToolsConfig: NeoActionToolsConfig | null = null;
 	private appMcpManager: NeoAppMcpManager | null = null;
 
 	constructor(
@@ -136,6 +141,18 @@ export class NeoAgentManager {
 	setToolsConfig(toolsConfig: NeoToolsConfig, appMcpManager?: NeoAppMcpManager): void {
 		this.toolsConfig = toolsConfig;
 		this.appMcpManager = appMcpManager ?? null;
+	}
+
+	/**
+	 * Wire in the action tools dependencies.
+	 *
+	 * Must be called before `provision()`. When set, `provision()` will create the
+	 * neo-action MCP server and attach it alongside the neo-query server.
+	 *
+	 * @param actionToolsConfig  All dependencies needed by createNeoActionMcpServer().
+	 */
+	setActionToolsConfig(actionToolsConfig: NeoActionToolsConfig): void {
+		this.actionToolsConfig = actionToolsConfig;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


lsm (commented):

This is a clean, well-structured implementation that correctly wires the action tools MCP server into Neo's provisioning pipeline. The code follows established patterns from the query tools attachment (task 2.5) and extends them naturally. All 335 Neo unit tests pass, typecheck is clean, and the commit message is thorough.

### Correctness

The core logic is sound:
- `createNeoToolsMcpServers()` in `neo-tools-server.ts` correctly combines both servers into a named map.
- `NeoAgentManager.attachTools()` properly merges registry servers with in-process servers, where in-process wins on collision via spread ordering (line 371-374 of `neo-agent-manager.ts`).
- The rename from `attachQueryTools()` to `attachTools()` is clean and the guard `if (!this.toolsConfig) return` ensures no-op when query config is not set.
- In `rpc-handlers/index.ts`, the wiring order is correct: query config at line 377, action config at line 569, and both happen before `neoAgentManager.provision()` in `app.ts` at line 464.

### One Concern (P1)

**`getActiveSessionForRoom` never returns null** (`packages/daemon/src/lib/rpc-handlers/index.ts` lines 557-560):

The `send_message_to_room` action handler (in `neo-action-tools.ts` line 1433) checks `if (!sessionId)` and returns an error. But the adapter at lines 557-560 always returns `room:${roomId}` regardless of whether that session actually exists. This means `send_message_to_room` will attempt to inject a message into a non-existent session, which will likely throw inside `sessionManager.injectMessage()`. The error will bubble up as an unhandled rejection rather than a clean error result.

This is not introduced by this PR (the same pattern exists for the Global Spaces Agent at line 612-618 of `app.ts`), but it is wired into the action tools here for the first time and should be noted. The ideal fix would be to check `deps.sessionManager.getSession("room:${roomId}")` before returning the session ID.

### Minor Observations (P3)

- **Inconsistent import path for `NeoToolsConfig`**: `neo-agent-manager.ts` imports from `neo-tools-server` (the new facade), but `rpc-handlers/index.ts` line 85 still imports from `neo-query-tools` directly. Both paths work because `neo-tools-server` re-exports, but using the facade consistently would be cleaner.
- **`as unknown as McpServerConfig` casts** in `neo-tools-server.ts` lines 39 and 42 suggest a type mismatch between the SDK's MCP server return type and the shared `McpServerConfig` type. This is an existing pattern, not introduced here, but worth tracking.

### Tests

The 10 integration tests are well-structured and cover the important scenarios: both-servers attach, query-only fallback, no-op without query config, registry merge, collision precedence, health recovery, clearSession, and config overwrite. They correctly mock the `AgentSession` interface and verify `setRuntimeMcpServers` call arguments.

**Recommendation: APPROVE** -- The P1 concern is pre-existing and not blocking. This PR is ready to merge.

lsm (commented):

The P1 fix in commit `b3b3821d8` is correct and complete.

### P1 Fix Verification

**`getActiveSessionForRoom` now properly returns null** (`packages/daemon/src/lib/rpc-handlers/index.ts` lines 557-562):

The fix adds an existence check via `deps.sessionManager.getSession(sessionId) !== null` before returning the session ID. This is the right approach:

- `SessionManager.getSession()` is synchronous and returns `AgentSession | null` (it delegates to `SessionCache.get()` which returns null for uncached sessions). No async gap or race condition.
- When the session does not exist, the adapter returns `null`, which flows to the `send_message_to_room` handler at `neo-action-tools.ts` line 1434 where `if (!sessionId)` produces a clean `errorResult("No active session found for room: ...")` instead of an unhandled rejection from `injectMessage`.
- The `getActiveSessionForTask` adapter (lines 564-567) already had the correct pattern (`task?.taskAgentSessionId ?? null`), so it was only the room variant that needed fixing.

### No New Issues Introduced

- All 335 Neo unit tests pass (4 files).
- The `createNeoToolsMcpServers` facade, `NeoAgentManager.setActionToolsConfig()`, the `attachTools()` rename, and the system prompt update from the original commit are unchanged and remain correct.
- No regressions in the existing query-tools tests or the previous test files.

### P3 Observations (unchanged from previous review)

- **Inconsistent import paths**: `rpc-handlers/index.ts` line 85 still imports `NeoToolsConfig` from `neo-query-tools` directly, while `neo-agent-manager.ts` imports from `neo-tools-server`. Both work due to re-exports, but using the facade consistently would be cleaner.

### Recommendation: APPROVE

The P1 is properly resolved. No blocking or should-fix issues remain.

PR #1136, opened 2026-03-28
feat(neo): Activity logging infrastructure (task 5.1)

lsm:
## Summary

- **`NeoActivityLogger`** class wrapping `NeoActivityLogRepository`: `logAction()`, `getRecentActivity(limit, offset)`, `getLatestUndoable()`, `pruneOldEntries()` (30-day retention + 10k-row cap; called on construction and each provision cycle)
- **`NeoActivityLogRepository`** extended with `update()` (partial field update) and `pruneOldEntries()` (age + count-based pruning)
- **`createNeoActionMcpServer`** wraps every tool callback with a `logged()` helper that records each invocation, skips `confirmationRequired` responses (deferred actions haven't executed yet), and captures undo data for 9 reversible tools via pre-execution state capture or post-execution extraction
- **Undo data captured for**: `create_room` (roomId), `create_goal` (goalId), `create_task` (taskId), `toggle_skill` (previousEnabled), `toggle_mcp_server` (previousEnabled), `update_app_settings` (previous values), `set_goal_status` (previousStatus), `set_task_status` (previousStatus), `update_room_settings` (previous fields)
- **Intentionally non-undoable**: `delete_*` (data gone), `send_message_*` (messages already processed), `start/cancel_workflow_run` (cascading side effects), `approve/reject_gate` and `approve/reject_task` (downstream steps triggered)
- **`NeoAgentManager`** wired with `setActivityLogger()` and propagates logger to action tools config
- **28 unit tests** covering repository additions, logger class, and logging logic

## Test plan
- [ ] `bun test tests/unit/neo/neo-activity-logger.test.ts` — all 28 pass
- [ ] `bun test tests/unit/neo/` — full neo suite (400+ tests) passes
- [ ] `bun run typecheck` passes
- [ ] `bun run lint` passes

### DIFF
--- a/packages/daemon/src/lib/neo/activity-logger.ts
+++ b/packages/daemon/src/lib/neo/activity-logger.ts
@@ -1,0 +1,122 @@
+/**
+ * NeoActivityLogger
+ *
+ * High-level logging service that wraps NeoActivityLogRepository.
+ * Records every Neo tool invocation for auditing and undo support.
+ *
+ * Responsibilities:
+ * - logAction()          — insert a completed activity log entry
+ * - getRecentActivity()  — paginated list (newest-first)
+ * - getLatestUndoable()  — most recent undoable entry
+ * - pruneOldEntries()    — enforce 30-day retention and 10 000-row cap
+ */
+
+import { randomUUID } from 'crypto';
+import type { NeoActivityLogEntry } from '../../storage/repositories/neo-activity-log-repository';
+import type { NeoActivityLogRepository } from '../../storage/repositories/neo-activity-log-repository';
+
+// ---------------------------------------------------------------------------
+// Public types
+// ---------------------------------------------------------------------------
+
+/** Parameters for logging a completed tool invocation. */
+export interface LogActionParams {
+	/** Name of the Neo tool that was invoked. */
+	toolName: string;
+	/** Tool input arguments (will be JSON-serialised). */
+	input: Record<string, unknown>;
+	/** Raw tool output text (already-serialised JSON string from MCP result). */
+	output?: string | null;
+	/** Execution outcome. */
+	status: 'success' | 'error' | 'cancelled';
+	/** Error message when status is 'error'. */
+	error?: string | null;
+	/** Entity type targeted by the action (e.g. 'room', 'skill', 'task'). */
+	targetType?: string | null;
+	/** ID of the targeted entity. */
+	targetId?: string | null;
+	/**
+	 * Whether the action can be reversed via `undo_last_action`.
+	 * Only true for successful operations on the undoable-tool list.
+	 */
+	undoable?: boolean;
+	/**
+	 * Data required to undo the action (e.g. previous state for updates,
+	 * created entity ID for creates). Serialised to JSON string for storage.
+	 */
+	undoData?: Record<string, unknown> | null;
+}
+
+// ---------------------------------------------------------------------------
+// NeoActivityLogger
+// ---------------------------------------------------------------------------
+
+export class NeoActivityLogger {
+	constructor(private readonly repo: NeoActivityLogRepository) {
+		// Prune on startup so stale entries are removed before the activity feed loads.
+		this.pruneOldEntries();
+	}
+
+	// ── Core logging ──────────────────────────────────────────────────────────
+
+	/**
+	 * Record a completed tool invocation.
+	 * Returns the inserted entry (including its generated ID and timestamp).
+	 */
+	logAction(params: LogActionParams): NeoActivityLogEntry {
+		return this.repo.insert({
+			id: randomUUID(),
+			toolName: params.toolName,
+			input: JSON.stringify(params.input),
+			output: params.output ?? null,
+			status: params.status,
+			error: params.error ?? null,
+			targetType: params.targetType ?? null,
+			targetId: params.targetId ?? null,
+			undoable: params.undoable ?? false,
+			undoData: params.undoData != null ? JSON.stringify(params.undoData) : null,
+		});
+	}
+
+	// ── Query helpers ─────────────────────────────────────────────────────────
+
+	/**
+	 * Return recent activity entries, newest-first.
+	 *
+	 * @param limit   Maximum rows to return (default 50).
+	 * @param offset  Number of rows to skip for pagination (default 0).
+	 */
+	getRecentActivity(limit = 50, offset = 0): NeoActivityLogEntry[] {
+		// The repository's list() uses cursor-based pagination.
+		// For offset-based access we fetch limit+offset and slice — acceptable
+		// because the activity feed only ever requests a small window.
+		if (offset === 0) {
+			return this.repo.list({ limit });
+		}
+		const rows = this.repo.list({ limit: limit + offset });
+		return rows.slice(offset);
+	}
+
+	/**
+	 * Return the most recent undoable log entry, or null if none exists.
+	 */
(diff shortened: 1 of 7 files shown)


lsm (commented):

---

### Summary

The implementation delivers the core infrastructure well: `NeoActivityLogger`, the repository extensions, the `logged()` wrapper around all 32 tool callbacks, wiring into `NeoAgentManager` and `rpc-handlers/index.ts`, and 28 unit tests. The architecture is clean, the undo-data capture design is thoughtful, and the acceptance criteria are met at the structural level.

However, there are a few issues worth addressing before or shortly after merge.

---

### P1: `logged()` does not catch exceptions from the tool handler

**File:** `packages/daemon/src/lib/neo/tools/neo-action-tools.ts`, lines 1721-1722

If `fn()` throws an unhandled exception (e.g., a database error inside a handler), the activity logger never records the invocation. The exception propagates up to the MCP server layer and the action is invisible in the activity log.

```typescript
const preUndoData = opts.undoable && opts.preCapture ? await opts.preCapture() : null;
const result = await fn();  // <-- if this throws, nothing is logged
```

This is a missed auditing opportunity. Unexpected tool failures should still be recorded with `status: 'error'` so the activity feed is a complete record of attempted actions.

**Recommendation:** Wrap `fn()` in a try/catch. On error, log the entry with `status: 'error'` and `error: err.message`, then re-throw.

---

### P1: `preCapture()` failure crashes the entire tool call

**File:** `packages/daemon/src/lib/neo/tools/neo-action-tools.ts`, line 1721

If a `preCapture()` callback throws (e.g., database failure when fetching the current goal/task status), the exception propagates up and the tool call fails entirely -- even though the actual operation (toggling a skill, updating a status) would have succeeded. The undo-data capture should be best-effort, never breaking the user's requested action.

```typescript
const preUndoData = opts.undoable && opts.preCapture ? await opts.preCapture() : null;
// If preCapture throws here, the entire tool call fails.
```

**Recommendation:** Wrap `preCapture()` in its own try/catch that returns `null` on failure. The action should proceed without undo data rather than failing entirely.

---

### P2: Tests do not exercise the `logged()` wrapper end-to-end

**File:** `packages/daemon/tests/unit/neo/neo-activity-logger.test.ts`, lines 647-815

The tests in the `createNeoActionMcpServer -- activity logging` describe block simulate logging behavior manually (direct `logAction` calls and manual `confirmationRequired` checks) but never actually invoke a tool through `createNeoActionMcpServer` and verify that the logger was called automatically. The comment on line 681-683 acknowledges this limitation.

The current tests verify the logger class and repository in isolation, which is valuable, but there is no test that:
1. Creates an MCP server via `createNeoActionMcpServer` with a real `activityLogger`
2. Invokes a tool handler through the MCP server's internal tool map
3. Asserts that `logAction` was called with the expected parameters

This means the integration between the `logged()` wrapper and the tool definitions is untested -- regressions in the wrapper logic (e.g., the `confirmationRequired` skip, undo-data capture, error detection) would not be caught.

**Recommendation:** Add at least 2-3 integration tests that invoke tools through the MCP server and verify logger calls. The `callTool` helper (line 626) and spy pattern (line 662) are already scaffolded for this purpose.

---

### P2: `getLatestUndoable` query lacks tiebreaker for deterministic ordering

**File:** `packages/daemon/src/storage/repositories/neo-activity-log-repository.ts`, lines 186-196

The query orders by `created_at DESC` but does not include `id DESC` as a tiebreaker:

```sql
SELECT * FROM neo_activity_log
WHERE undoable = 1
ORDER BY created_at DESC
LIMIT 1
```

When two entries share the same millisecond timestamp, the result is non-deterministic. All other queries in this repository and the LiveQuery handler use `created_at DESC, id DESC` for stable ordering.

**Recommendation:** Add `id DESC` to the ORDER BY clause. Also consider adding a composite index on `(undoable, created_at DESC, id DESC)` for this query, since it cannot efficiently use the existing `idx_neo_activity_log_created_at` index (it needs to filter on `undoable` first).

---

### P3: Double-pruning on startup

**File:** `packages/daemon/src/lib/neo/activity-logger.ts`, line 57 and `packages/daemon/src/lib/neo/neo-agent-manager.ts`, line 124

The constructor calls `pruneOldEntries()`, and then `provision()` calls it again. This is harmless (prune is idempotent) but results in two DELETE queries against an empty result set on every startup. Consider removing one of the calls for clarity.

---

### Verdict

The implementation is solid and the architecture is well-thought-out. The P1 issues (uncaught exceptions in the logging wrapper) are real gaps in audit completeness that should be addressed. The P2 issues (test coverage and query determinism) are important for long-term reliability. None are blocking for an initial merge if addressed in a quick follow-up.

lsm (commented):

### Verdict: APPROVE

Solid implementation of the activity logging infrastructure. The code is well-structured, follows existing patterns, and all previous review feedback has been addressed.

---

### What's done well

- **Clean separation of concerns**: `NeoActivityLogger` wraps the repository with business logic (JSON serialization, pagination adapter, undo data handling), keeping the repository as a thin persistence layer.
- **Order-independent wiring**: `setActivityLogger()` and `setActionToolsConfig()` in `NeoAgentManager` each propagate to the other, eliminating ordering bugs at startup.
- **Error resilience**: The `logged()` wrapper's try/catch around both `preCapture()` and the handler ensures logging failures never break tool execution.
- **Thoughtful undo design**: The pre/post-capture pattern cleanly separates update-operations (need previous state) from create-operations (need new entity ID). Tools that are irreversible (delete, send_message, workflow start) are correctly excluded.
- **Pruning on construction + provision cycle**: Ensures retention policy is enforced at startup and periodically during the daemon's lifetime.

1. **Offset-based pagination inconsistency**: `getRecentActivity(limit, offset)` at `activity-logger.ts:89-98` fetches `limit+offset` rows and slices. If new entries are inserted between page fetches, pages can overlap or skip rows. The comment correctly notes this is acceptable for a small-window activity feed, but the repository already supports cursor-based pagination. If the frontend ever implements infinite scroll, switching to cursor-based would be more robust.

2. **Missing composite index for `getLatestUndoable()`**: The query at `neo-activity-log-repository.ts:188-193` filters on `undoable = 1` and orders by `created_at DESC, id DESC`. The only index is on `created_at` alone (`idx_neo_activity_log_created_at`), so this query does a full scan of the undoable subset. With the 10k row cap this is not a practical concern, but a composite index `CREATE INDEX idx_neo_activity_log_undoable_created_at ON neo_activity_log(undoable, created_at DESC, id DESC)` would make it O(1).

3. **`update()` method is currently unused**: The repository's `update()` method is well-tested but not called anywhere in production code yet. This is clearly forward-looking for the undo feature, which is fine, but worth noting.

### Test coverage

29 tests pass, covering:
- Repository `update()` and `pruneOldEntries()` (7 tests)
- Logger `logAction`, `getRecentActivity`, `getLatestUndoable`, constructor pruning (12 tests)
- End-to-end integration via `_registeredTools` handler invocation (5 tests): success+undo data, preCapture capture, confirmationRequired skip, handler throw with error logging, preCapture throw resilience
- Logging infrastructure behavioral tests (5 tests)

The integration tests that invoke tools through the MCP server's registered handlers are particularly valuable -- they verify the full `logged()` wrapper pipeline including the confirmationRequired skip path.

### Checks

- `bun test tests/unit/neo/neo-activity-logger.test.ts` -- 29 pass
- `bun run typecheck` -- passes
- `bun run lint` -- 0 warnings, 0 errors

PR #1141, opened 2026-03-28
fix(neo): render SDK messages using SDKMessageRenderer in NeoChatView

lsm:
Fixes broken message rendering in NeoChatView where assistant messages were displayed as raw `StructuredDataCard` tables (showing model name, token counts, session IDs) instead of actual chat text.

**Root cause:** `msg.content` holds the full `sdk_message` JSON (`{ type: 'assistant', message: { content: [...], model: '...', usage: {...} }, ... }`), but the old `extractTextContent()` only handled plain content arrays, so it returned empty string — causing `tryParseStructured()` to render the whole API response object as a table.

**Changes:**
- Remove `extractTextContent`, `tryParseStructured`, `StructuredDataCard`
- Add `parseSDKMessage()` to safely parse the raw JSON into a typed `SDKMessage`
- Assistant messages: render via `SDKMessageRenderer` (handles text, tool calls, markdown, thinking blocks) inside the existing Neo sparkle avatar wrapper
- User messages: extract text from `SDKUserMessage.message.content` (supports both string and `ContentBlock[]` formats)
- Update tests to use realistic SDK message format; mock `SDKMessageRenderer` for unit isolation

### DIFF
--- a/packages/web/src/components/neo/NeoChatView.test.tsx
+++ b/packages/web/src/components/neo/NeoChatView.test.tsx
@@ -457,4 +457,62 @@ describe('NeoChatView', () => {
 		expect(getByTestId('neo-user-message')).toBeTruthy();
 		expect(getByText('Hello as a string')).toBeTruthy();
 	});
+
+	it('renders parse-error fallback when assistant message has invalid JSON', () => {
+		neoStore.messages.value = [
+			{
+				...makeAssistantMsg('1', 'ignored'),
+				content: 'not valid json at all',
+			},
+		];
+		const { getByTestId, queryByTestId } = render(<NeoChatView />);
+		expect(getByTestId('neo-assistant-message')).toBeTruthy();
+		expect(getByTestId('neo-message-parse-error')).toBeTruthy();
+		expect(queryByTestId('sdk-message-renderer')).toBeNull();
+	});
+
+	it('renders parse-error fallback when assistant content is a non-assistant SDK type', () => {
+		// A user-type SDK message inadvertently stored for an assistant message row
+		const sdkMsg = {
+			type: 'user',
+			message: { role: 'user', content: [{ type: 'text', text: 'unexpected' }] },
+			parent_tool_use_id: null,
+			session_id: 'sess-1',
+		};
+		neoStore.messages.value = [
+			{
+				...makeAssistantMsg('1', 'ignored'),
+				content: JSON.stringify(sdkMsg),
+			},
+		];
+		const { getByTestId, queryByTestId } = render(<NeoChatView />);
+		expect(getByTestId('neo-assistant-message')).toBeTruthy();
+		expect(getByTestId('neo-message-parse-error')).toBeTruthy();
+		expect(queryByTestId('sdk-message-renderer')).toBeNull();
+	});
+
+	it('renders parse-error fallback when assistant content lacks type field', () => {
+		neoStore.messages.value = [
+			{
+				...makeAssistantMsg('1', 'ignored'),
+				content: JSON.stringify({ foo: 'bar' }),
+			},
+		];
+		const { getByTestId, queryByTestId } = render(<NeoChatView />);
+		expect(getByTestId('neo-assistant-message')).toBeTruthy();
+		expect(getByTestId('neo-message-parse-error')).toBeTruthy();
+		expect(queryByTestId('sdk-message-renderer')).toBeNull();
+	});
+
+	it('renders empty string (not raw JSON) for user message with invalid content', () => {
+		neoStore.messages.value = [
+			{
+				...makeUserMsg('1', 'ignored'),
+				content: 'not-valid-json',
+			},
+		];
+		const { getByTestId } = render(<NeoChatView />);
+		const bubble = getByTestId('neo-user-message');
+		expect(bubble.textContent).toBe('');
+	});
 });
--- a/packages/web/src/components/neo/NeoChatView.test.tsx
+++ b/packages/web/src/components/neo/NeoChatView.test.tsx
@@ -3,8 +3,10 @@
  *
  * Verifies:
  * - Empty state renders when no messages
- * - User messages render as right-aligned bubbles
- * - Assistant messages render with sparkle avatar
+ * - User messages render as right-aligned bubbles with correct text
+ * - Assistant messages render with sparkle avatar using SDKMessageRenderer
+ * - SDK messages with content arrays are parsed and rendered correctly
+ * - Tool call messages render via SDKMessageRenderer
  * - Sending a message calls neoStore.sendMessage
  * - Enter key submits; Shift+Enter does not
  * - Send button disabled when input is empty or sending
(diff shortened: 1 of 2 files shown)


lsm (commented):

Good change overall. Replacing the broken custom rendering with `SDKMessageRenderer` is the right call, and the test updates properly reflect the new SDK message JSON format. Tests pass (25/25), typecheck is clean.

### P2: Empty assistant bubble when SDK message parsing fails

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, line 209

```tsx
{parsedMsg && <SDKMessageRenderer message={parsedMsg} />}
```

When `parseSDKMessage()` returns `null` (malformed JSON, missing `type` field, or unexpected content format), the assistant message renders as just a sparkle avatar with nothing next to it — a completely empty bubble. This is a silent failure the user cannot diagnose.

Consider adding a fallback that shows the raw content or a small error indicator:

```tsx
{parsedMsg ? (
  <SDKMessageRenderer message={parsedMsg} />
) : (
  <div class="text-sm text-gray-500 italic">Unable to display message</div>
)}
```

A corresponding test for malformed content (e.g., `content: 'not valid json'` or `content: JSON.stringify({ foo: 'bar' })`) would cover this edge case.

### P3: Minimal validation in `parseSDKMessage`

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, lines 120-130

The function only checks `typeof parsed === 'object' && parsed !== null && 'type' in parsed` before casting to `SDKMessage`. Any object with a `type` field will pass through. This is fine for a display context (SDKMessageRenderer has its own guards and an "unknown type" fallback), but worth noting.

### P3: Missing test for unparseable content

No test covers the case where `msg.content` is invalid JSON or a valid JSON object without a `type` field. Adding one would document the expected behavior and prevent regressions if someone changes the fallback logic.

---

**Recommendation: REQUEST_CHANGES** — The P2 (empty bubble on parse failure) should be addressed before merge. The fix is small: add a fallback rendering path and one test case.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P0 silent-failure and two meaningful P1 gaps.

---

### Summary

The core fix is correct and well-structured: removing the broken `extractTextContent`/`tryParseStructured`/`StructuredDataCard` machinery and routing assistant messages through `SDKMessageRenderer` is exactly the right approach. The new `parseSDKMessage` helper is safe (null-on-failure), `extractUserText` handles both string and array content forms, and the tests are realistic and well-isolated with proper mocking. The change is a net improvement.

---

### P0 — Silent render failure when `parsedMsg` is null for assistant messages

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, line 209

```tsx
{parsedMsg && <SDKMessageRenderer message={parsedMsg} />}
```

When `parseSDKMessage` returns `null` (malformed JSON, truncated row, or any future unexpected shape), an assistant message renders as an empty `neo-assistant-message` shell: just the sparkle avatar with no content and no error indicator. The user sees a blank bubble with no feedback. This is a silent failure — the chat appears broken with no way to diagnose it.

For a production chat UI this should at minimum render a fallback string (e.g., `"Neo sent a message that could not be displayed"`) so users know something was received. The fallback does not need to be elaborate, but silent blank assistant bubbles are worse UX than a visible error state.

---

### P1 — User messages pass through `SDKMessageRenderer`'s user-message rendering path when `isUser` is false

**File:** `packages/web/src/components/sdk/SDKMessageRenderer.tsx`, lines 208-222

This is not introduced by this PR, but it interacts with the new code. `SDKMessageRenderer` will fully render any `user`-type SDK message via `SDKUserMessage` (complete with rewind controls, metadata, etc.). In `NeoChatView`, user messages are branched away before reaching `SDKMessageRenderer` (line 180), so there is no double-render in normal flow. However, a `NeoMessage` whose `messageType !== 'user'` but whose raw content happens to be a `user`-type SDK message would silently fall through to the assistant branch and be passed to `SDKMessageRenderer`, which would render it as an `SDKUserMessage` inside the assistant bubble layout. This edge case is not tested.

This is a low-probability scenario today but reflects a layered inconsistency: the `isUser` split in `MessageBubble` is based on `NeoMessage.messageType` while `SDKMessageRenderer` dispatches on the inner `SDKMessage.type`. If those diverge (possible with synthetic or replayed messages), rendering will be surprising. A guard such as `parsedMsg?.type === 'assistant'` before calling `SDKMessageRenderer` for the assistant branch would close this gap.

---

### P1 — `extractUserText` falls back to `msg.content` (raw JSON) when parse fails, not to empty string

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, line 183

```tsx
const userText =
  parsedMsg?.type === 'user' ? extractUserText(parsedMsg as UserSDKMessage) : msg.content;
```

When `parsedMsg` is null (JSON parse failure) or is not type `'user'`, the raw `msg.content` — which is the full serialised SDK message JSON — is rendered as the bubble text. This reproduces the original bug for the user-message case: on a malformed row the user sees a wall of raw JSON in the blue chat bubble. The fallback should be an empty string or a descriptive placeholder, not the unparsed blob.

---

### P2 — `parseSDKMessage` type-guard is overly permissive

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, lines 122-129

```tsx
if (typeof parsed === 'object' && parsed !== null && 'type' in parsed) {
  return parsed as SDKMessage;
}
```

Any JSON object with a `type` key is accepted as an `SDKMessage`. This means a stale or migrated row whose top-level shape has changed will silently pass as a valid `SDKMessage` and may cause runtime errors inside `SDKMessageRenderer`'s type-specific renderers. Since `SDKMessage` is a discriminated union keyed on `type`, the guard should at minimum verify that `type` is a known string literal (`'assistant'`, `'user'`, `'result'`, etc.) before casting. The current approach is a single-character defense against null that adds very little real safety.

---

### P3 — `UserSDKMessage` type alias exported but not needed outside the module

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, line 132

```tsx
type UserSDKMessage = Extract<SDKMessage, { type: 'user' }>;
```

This is a module-local alias used only in `extractUserText`. Fine as-is. Minor note only.

---

### Tests — adequate but missing the null-parse fallback case

The 25 tests cover the happy path, string content, tool calls, error cards, typing indicator, confirmation card, and message ordering well. The one gap that mirrors the P0 above: there is no test for what the component renders when `msg.content` is not valid JSON for an assistant message. A test verifying graceful fallback (not blank) would complete the coverage.

lsm (commented):

**Recommendation: APPROVE**

All three issues from the previous review round have been properly addressed:

### 1. Parse-error fallback (previously P0/P2) -- Fixed

`NeoChatView.tsx` lines 212-218 now render a visible fallback (`"Unable to display message"` with `data-testid="neo-message-parse-error"`) when `parsedMsg` is null or not an assistant type. No more empty bubbles. Three new tests cover this path: invalid JSON (line 461), non-assistant SDK type stored for an assistant row (line 474), and missing `type` field (line 494).

### 2. User message fallback is empty string (previously P1) -- Fixed

`NeoChatView.tsx` line 183: the ternary fallback is now `''` instead of `msg.content`, so unparseable user messages show an empty bubble rather than raw JSON. Test at line 507 verifies `bubble.textContent` is `''`.

### 3. Assistant type guard (previously P1) -- Fixed

`NeoChatView.tsx` line 196: `const isAssistantSDKMsg = parsedMsg?.type === 'assistant'` gates the `SDKMessageRenderer` path. Only messages whose inner SDK type is `'assistant'` are passed through; all others (including a `user`-type SDK message erroneously stored on an assistant `NeoMessage` row) hit the parse-error fallback. Test at line 474 covers this exact scenario.

All 29 tests pass. The implementation is clean and the edge cases are well-covered.

lsm (commented):

Recommendation: **APPROVE** — the three issues from the prior round are all correctly addressed and tests pass. One minor type-safety note worth knowing but not blocking.

---

### Summary of Previous Issues and How They Were Fixed

| Prior finding | Fix |
|---|---|
| P0/P2 — empty bubble when parsedMsg is null | fallback `<div data-testid="neo-message-parse-error">` with "Unable to display message" text (NeoChatView.tsx:215) |
| P1 — user message fallback should be empty string not raw JSON | line 183: ternary returns `''` when `parsedMsg?.type !== 'user'` |
| P1 — guard assistant branch to only pass type==='assistant' to SDKMessageRenderer | `isAssistantSDKMsg` flag gates the SDKMessageRenderer call (lines 196–213) |

---

### Test Coverage (29/29 pass)

New edge-case tests are thorough and cover all three fix paths:
- `'renders parse-error fallback when assistant message has invalid JSON'`
- `'renders parse-error fallback when assistant content is a non-assistant SDK type'`
- `'renders parse-error fallback when assistant content lacks type field'`
- `'renders empty string (not raw JSON) for user message with invalid content'`

---

### Remaining Observation (P2 — non-blocking)

**File:** `packages/web/src/components/neo/NeoChatView.tsx`, line 213

```tsx
const isAssistantSDKMsg = parsedMsg?.type === 'assistant';
// ...
{isAssistantSDKMsg ? (
    <SDKMessageRenderer message={parsedMsg} />   // parsedMsg is SDKMessage | null here
) : (
    ...
)}
```

`parsedMsg` is typed `SDKMessage | null` throughout the component scope. TypeScript does NOT narrow `parsedMsg` through the separate `isAssistantSDKMsg` boolean variable, so passing it to `SDKMessageRenderer` (which declares `message: SDKMessage`) is technically a type mismatch. It compiles today because `sdk.d.ts` is excluded from the web tsconfig (line 21 of tsconfig.json). If that exclusion were ever removed, this would surface as a type error.

The safe fix is either of:
```tsx
// Option A — inline narrowing TypeScript can track:
{parsedMsg !== null && parsedMsg.type === 'assistant' ? (
    <SDKMessageRenderer message={parsedMsg} />
) : (
    ...
)}

// Option B — non-null assertion (explicit but coarser):
<SDKMessageRenderer message={parsedMsg!} />
```

Runtime behavior is correct — the guard fires before the component is called — so this does not cause any observable bug. Logging it as P2 for awareness.

---

### Minor Gap (P3)

No test covers the case where a user-row message has valid JSON but the SDK `type` is not `'user'` (e.g., accidentally stored `type: 'assistant'`). The fallback renders an empty bubble in that case, which is acceptable, but it is a gap relative to the assistant-side coverage (which tests the analogous scenario). Optional to add.

---

All three originally requested fixes are present and verified. The code is correct at runtime. No blocking issues found.

PR #1144, opened 2026-03-28
test(e2e): Neo chat rendering — verify messages render as readable text

lsm:
Adds `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` with 7 Playwright E2E tests covering Neo chat message rendering:

- Empty state shows greeting before any messages
- Chat/Activity tab switching works correctly
- User messages render in right-aligned blue bubble with correct CSS
- Sparkle avatar appears next to assistant messages
- Assistant messages render as readable text, not raw JSON or `"Structured data"` cards
- `neo-message-parse-error` fallback is absent for valid responses
- Empty state disappears once first message is sent

All tests follow E2E rules: pure UI interactions, DOM-only assertions, RPC only in beforeEach/afterEach. API-dependent tests guard with `isNeoAvailable()` and skip gracefully.

### DIFF
--- a/packages/e2e/tests/features/neo-chat-rendering.e2e.ts
+++ b/packages/e2e/tests/features/neo-chat-rendering.e2e.ts
@@ -1,0 +1,299 @@
+/**
+ * Neo Chat Rendering E2E Tests
+ *
+ * Verifies that the Neo chat panel renders messages correctly as readable text,
+ * not as raw JSON or structured data tables.
+ *
+ * Test scenarios:
+ * 1. Empty state shows correctly before any messages
+ * 2. Chat/Activity tab switching works
+ * 3. User messages appear in the correct bubble style (right-aligned blue bubble)
+ * 4. Neo sparkle avatar appears next to assistant messages
+ * 5. Assistant messages render as readable text, NOT a "Structured data" card
+ *
+ * E2E Principles (from CLAUDE.md):
+ * - All test actions go through UI (clicks, typing, keyboard shortcuts).
+ * - All assertions verify visible DOM state.
+ * - RPC is allowed only in beforeEach/afterEach for setup/teardown.
+ * - NEOKAI_ENABLE_NEO_AGENT=1 is set in playwright.config.ts.
+ */
+
+import { test, expect, type Page } from '../../fixtures';
+import { waitForWebSocketConnected } from '../helpers/wait-helpers';
+
+// ─── Constants ────────────────────────────────────────────────────────────────
+
+const NEO_PANEL_TESTID = 'neo-panel';
+const NEO_CHAT_INPUT_TESTID = 'neo-chat-input';
+const NEO_USER_MESSAGE_TESTID = 'neo-user-message';
+const NEO_ASSISTANT_MESSAGE_TESTID = 'neo-assistant-message';
+
+// ─── Helpers ──────────────────────────────────────────────────────────────────
+
+/**
+ * Open the Neo panel by clicking the Neo NavRail button.
+ */
+async function openNeoPanel(page: Page): Promise<void> {
+	const neoButton = page.getByRole('button', { name: 'Neo (⌘J)', exact: true });
+	await neoButton.waitFor({ state: 'visible', timeout: 5000 });
+	await neoButton.click();
+	await page.getByTestId(NEO_PANEL_TESTID).waitFor({ state: 'visible', timeout: 5000 });
+}
+
+/**
+ * Send a message in the Neo chat input.
+ */
+async function sendNeoMessage(page: Page, text: string): Promise<void> {
+	const input = page.getByTestId(NEO_CHAT_INPUT_TESTID);
+	await input.waitFor({ state: 'visible', timeout: 5000 });
+	await input.fill(text);
+	await input.press('Enter');
+}
+
+/**
+ * Wait for a new Neo assistant response to appear (any content).
+ * Uses count-based detection so previous responses don't trigger a false positive.
+ */
+async function waitForNeoAssistantResponse(
+	page: Page,
+	options: { timeout?: number } = {}
+): Promise<void> {
+	const timeout = options.timeout ?? 90000;
+	const initialCount = await page.getByTestId(NEO_ASSISTANT_MESSAGE_TESTID).count();
+	await page.waitForFunction(
+		(expected) =>
+			document.querySelectorAll('[data-testid="neo-assistant-message"]').length > expected,
+		initialCount,
+		{ timeout }
+	);
+	// Also wait for the input to be re-enabled (loading state cleared)
+	await page.getByTestId(NEO_CHAT_INPUT_TESTID).waitFor({ state: 'visible', timeout: 10000 });
+}
+
+/**
+ * Check whether the Neo agent is provisioned (not showing an error card).
+ * Must be called with the Neo panel already open so error cards are rendered.
+ */
+async function isNeoAvailable(page: Page): Promise<boolean> {
+	const hasNoCredentials = await page
+		.getByTestId('neo-error-no-credentials')
+		.isVisible()
+		.catch(() => false);
+	const hasProviderError = await page
+		.getByTestId('neo-error-provider-unavailable')
+		.isVisible()
+		.catch(() => false);
+	return !hasNoCredentials && !hasProviderError;
+}
+
+// ─── Tests ───────────────────────────────────────────────────────────────────
+
+test.describe('Neo Chat Rendering', () => {
+	test.use({ viewport: { width: 1280, height: 720 } });
+
+	test.beforeEach(async ({ page }) => {
+		await page.goto('/');
+		await waitForWebSocketConnected(page);
+		// Clear persisted panel state so tests start with panel closed
+		await page.evaluate(() => localStorage.removeItem('neo:panelOpen'));
+		await page.reload();
+		await waitForWebSocketConnected(page);
+	});
+
+	// ── 1. Empty state ─────────────────────────────────────────────────────────
+
+	test('shows empty state with Neo introduction before any messages', async ({ page }) => {
+		await openNeoPanel(page);
+
+		// Chat view is rendered
+		await expect(page.getByTestId('neo-chat-view')).toBeVisible();
+
+		// Empty state is visible
+		const emptyState = page.getByTestId('neo-empty-state');
+		await expect(emptyState).toBeVisible();
+
+		// Shows the "Hi, I'm Neo" greeting text
+		await expect(emptyState).toContainText("Hi, I'm Neo");
+
+		// Shows helper prompt text
+		await expect(emptyState).toContainText('Ask me anything');
+
+		// No messages rendered yet
+		await expect(page.getByTestId(NEO_USER_MESSAGE_TESTID)).toHaveCount(0);
+		await expect(page.getByTestId(NEO_ASSISTANT_MESSAGE_TESTID)).toHaveCount(0);
+	});
+
+	// ── 2. Tab switching ───────────────────────────────────────────────────────
+
+	test('Chat/Activity tabs switch correctly', async ({ page }) => {
+		await openNeoPanel(page);
+
+		const chatTab = page.getByTestId('neo-tab-chat');
+		const activityTab = page.getByTestId('neo-tab-activity');
+
+		// Chat tab is active by default
+		await expect(chatTab).toHaveAttribute('aria-selected', 'true');
+		await expect(activityTab).toHaveAttribute('aria-selected', 'false');
+
+		// Chat view is visible, activity view is not
+		await expect(page.getByTestId('neo-chat-view')).toBeVisible();
+		await expect(page.getByTestId('neo-activity-view')).not.toBeVisible();
+
+		// Click Activity tab
+		await activityTab.click();
+
+		// Activity tab is now active
+		await expect(activityTab).toHaveAttribute('aria-selected', 'true');
+		await expect(chatTab).toHaveAttribute('aria-selected', 'false');
+
+		// Activity view is visible, chat view is not
+		await expect(page.getByTestId('neo-activity-view')).toBeVisible();
+		await expect(page.getByTestId('neo-chat-view')).not.toBeVisible();
+
+		// Switch back to Chat tab
+		await chatTab.click();
+
+		// Chat tab is active again
+		await expect(chatTab).toHaveAttribute('aria-selected', 'true');
+		await expect(page.getByTestId('neo-chat-view')).toBeVisible();
+	});
+
+	// ── 3. User message bubble style ───────────────────────────────────────────
+
+	test('user messages appear in right-aligned blue bubble after sending', async ({ page }) => {
+		await openNeoPanel(page);
+
+		// Need Neo available to send a message
+		if (!(await isNeoAvailable(page))) {
+			test.skip();
+			return;
+		}
+
+		await sendNeoMessage(page, 'Hello Neo');
+
+		// User message bubble appears
+		const userMsg = page
+			.getByTestId(NEO_USER_MESSAGE_TESTID)
+			.filter({ hasText: 'Hello Neo' })
+			.first();
+		await userMsg.waitFor({ state: 'visible', timeout: 10000 });
+
+		// The outer wrapper is right-aligned (justify-end)
+		await expect(userMsg).toHaveClass(/justify-end/);
+
+		// The inner bubble has blue background styling
+		const bubble = userMsg.locator('div').first();
+		await expect(bubble).toHaveClass(/bg-blue-600/);
+		await expect(bubble).toHaveClass(/text-white/);
+
+		// The text content is readable
+		await expect(userMsg).toContainText('Hello Neo');
+	});
+
+	// ── 4. Sparkle avatar next to assistant messages ───────────────────────────
+
+	test('Neo sparkle avatar appears next to assistant messages', async ({ page }) => {
+		await openNeoPanel(page);
+
+		if (!(await isNeoAvailable(page))) {
+			test.skip();
+			return;
+		}
+
+		await sendNeoMessage(page, 'Say hi back');
+		await waitForNeoAssistantResponse(page);
+
+		// The assistant message container is present
+		const assistantMsg = page.getByTestId(NEO_ASSISTANT_MESSAGE_TESTID).first();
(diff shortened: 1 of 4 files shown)


lsm (commented):

---

### Summary

The new file `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` adds 7 test scenarios for the Neo chat rendering feature. The rendering-specific tests (tests 5 and 6) directly address the fix in PR #1141 and are well-structured. However, there are two P0/P1 issues and several lower-priority concerns worth addressing.

---

### P0 — Test skipping violates the Hard Fail Rule (CLAUDE.md)

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` — lines 167–170, 198–200, 227–229, 263–265, 283–285

Five of the seven tests (tests 3–7) use an inline `isNeoAvailable()` guard that silently skips when no Neo credentials are present:

```ts
if (!(await isNeoAvailable(page))) {
    test.skip();
    return;
}
```

This pattern directly violates CLAUDE.md's "Credential-Dependent Online Tests — Hard Fail Rule":

> Online tests that require real provider credentials must FAIL, not skip, when those credentials are absent or non-functional.
> Do NOT silently skip tests because a secret is unset — that masks misconfiguration.

The skip guard masks exactly the scenario the rule is designed to catch: a CI environment where Neo credentials are not configured. The tests would silently pass as "skipped" rather than fail and alert the team.

The existing `neo-conversation.e2e.ts` follows the correct pattern: the `isNeoAvailable()` check is placed in `test.beforeEach()` at the describe-block level, so a missing credential causes the whole describe block to be skipped once, from the infrastructure boundary, rather than having each test body self-skip. But even that pattern appears to be a concession made by a previous PR rather than conforming to the hard-fail rule. The new tests should not further entrench the soft-skip pattern without explicit sign-off.

At minimum, the skip should be moved to a shared `beforeEach` for the AI-dependent tests. Ideally, if these tests are in CI's matrix, they should fail fast rather than skip.

---

### P1 — Test 2 (Tab switching) is a verbatim duplicate of existing coverage

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` — lines 128–159

The "Chat/Activity tabs switch correctly" test is substantively identical to:

- `neo-panel.e2e.ts` — "can switch between Chat and Activity tabs" (lines 115–135)
- `neo-conversation.e2e.ts` — "switches to Activity tab and back to Chat tab" (lines 321–335)

All three check `aria-selected` transitions and view visibility after clicking both tabs. Duplicating this in a rendering-focused test file adds CI time and maintenance surface without any new signal. The tab switching test belongs in `neo-panel.e2e.ts` or `neo-conversation.e2e.ts`, not here.

The stated purpose of this PR is verifying that messages render as readable text, not raw JSON. Tests 1 and 2 don't test rendering at all — they test structural UI properties already covered elsewhere.

---

### P1 — `waitForNeoAssistantResponse` waits on the wrong loading signal

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` — lines 69–71

The helper waits for the Neo chat input to become visible as a proxy for "loading complete":

```ts
await page.getByTestId(NEO_CHAT_INPUT_TESTID).waitFor({ state: 'visible', timeout: 10000 });
```

But `NeoChatView.tsx` only disables the input via the local `sending` state (`disabled={sending}`), not via `neoStore.loading`. The input is a `<textarea>` — `waitFor({ state: 'visible' })` waits for it to appear in the DOM, not for it to be enabled. The input is always visible (it is not conditionally rendered). This wait is a no-op if the input is already rendered, and will not actually confirm that the assistant response has been fully streamed before assertions run. Using `waitFor({ state: 'enabled' })` or asserting on the last assistant message's text content stabilizing would be more accurate.

This is directly copied from `neo-conversation.e2e.ts` where the same latent bug exists, but the new file inherits it instead of correcting it.

---

### P2 — Test 1 (Empty state) verifies "Ask me anything" but the component renders a different string

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` — line 119

The test asserts:

```ts
await expect(emptyState).toContainText('Ask me anything');
```

But `NeoChatView.tsx` (line 348) renders:

```ts
<p class="text-xs text-gray-500 mt-1">
    Ask me anything about your rooms, sessions, or goals.
</p>
```

`toContainText` uses a substring match, so "Ask me anything" will pass as long as the component outputs a string that contains that prefix. This is an acceptable substring match, not a bug per se, but the test comment says "Shows helper prompt text" — asserting on a more stable token (e.g. the full string, or a test ID on the paragraph) would be more robust against UI copy changes.

---

### P2 — `openNeoPanel` duplicated verbatim from `neo-conversation.e2e.ts`

**Files:** `neo-chat-rendering.e2e.ts` lines 36–41, `neo-conversation.e2e.ts` lines 43–48

The `openNeoPanel`, `sendNeoMessage`, `waitForNeoAssistantResponse`, and `isNeoAvailable` helpers are identical across both files. If the Neo panel UX changes (e.g. button name, test ID), both files need synchronised updates. Consider extracting these into `packages/e2e/tests/helpers/neo-helpers.ts`.

---

### P3 — The avatar CSS locator is brittle

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts` — line 212

```ts
const avatar = assistantMsg.locator('div.rounded-full.bg-violet-600\\/20').first();
```

This locates the avatar by Tailwind class names. Tailwind classes are utility-based implementation details — a CSS-only refactor would break this assertion without any functional regression. Using a `data-testid="neo-avatar"` attribute on the avatar element in the component would be more resilient.

---

### What works well

- Tests 5 and 6 are the core value of this PR. They directly test that `SDKMessageRenderer` is used and that the `neo-message-parse-error` fallback is absent, which maps precisely to the rendering fix in PR #1141.
- The count-based detection in `waitForNeoAssistantResponse` correctly avoids false positives from pre-existing responses.
- The `beforeEach` correctly clears `localStorage` so the panel starts in a known closed state.
- The `isNeoAvailable` check looks at `data-testid` DOM elements rather than internal state, consistent with E2E principles.

lsm (commented):

### Summary

The test file is well-structured, follows E2E conventions from CLAUDE.md, and correctly targets real test IDs from the source components. All 7 test scenarios are relevant to the original ask of verifying Neo chat renders readable text rather than raw JSON. The test patterns (isNeoAvailable guard, count-based assistant response detection, UI-only assertions) match existing conventions in `neo-conversation.e2e.ts`.

### Issues

**P2 — Duplicated helpers should be extracted to a shared module**

`openNeoPanel`, `sendNeoMessage`, `waitForNeoAssistantResponse`, and `isNeoAvailable` are copied verbatim from `packages/e2e/tests/features/neo-conversation.e2e.ts` (lines 43-48, 63-69, 86-100, 252-261). These four helpers should live in a shared file like `packages/e2e/tests/helpers/neo-helpers.ts` and be imported by both test files. This duplication means any future change (e.g., if the Neo button label changes from `Neo (⌘J)`) requires updating multiple files.

**P3 — Test 4 uses brittle CSS class selector for avatar**

At line 212, the locator `div.rounded-full.bg-violet-600\\/20` couples the test to specific Tailwind classes. If the avatar styling changes (e.g., from `bg-violet-600/20` to `bg-violet-500/20`), this test breaks without any functional regression. Consider adding a `data-testid="neo-sparkle-avatar"` to the component instead. This is minor since the existing component unit tests also use class-based assertions.

### Recommendation: APPROVE

The test file achieves its stated goal — verifying that Neo chat renders assistant messages as readable text, not raw JSON. The core assertions (tests 5 and 6) directly address the original ask. The helper duplication is worth cleaning up but is not blocking since the pattern already exists in the codebase.

lsm (commented):

**Recommendation: APPROVE**

All issues from the previous review have been addressed cleanly:

1. **Shared helpers extracted** -- `packages/e2e/tests/helpers/neo-helpers.ts` consolidates `openNeoPanel`, `closeNeoPanel`, `sendNeoMessage`, `waitForNeoUserMessage`, `waitForNeoAssistantResponse`, and `isNeoAvailable` into a single module. Both `neo-chat-rendering.e2e.ts` and `neo-conversation.e2e.ts` import from it with no remaining duplicates.

2. **Brittle CSS selector replaced** -- `data-testid="neo-sparkle-avatar"` added to the avatar `<div>` in `NeoChatView.tsx` (line 202). The rendering test now locates the avatar via `getByTestId('neo-sparkle-avatar')` instead of Tailwind class selectors.

3. **Duplicate tab-switching test removed** -- The "Chat/Activity tabs switch correctly" test that duplicated coverage from `neo-panel.e2e.ts` and `neo-conversation.e2e.ts` has been dropped. Test numbering updated accordingly.

4. **`waitFor` state fix** -- `waitForNeoAssistantResponse` in the shared helpers now waits for `state: 'enabled'` (line 84 of `neo-helpers.ts`) instead of `state: 'visible'`, which correctly detects when the `disabled={sending}` attribute is cleared on the textarea rather than just checking DOM presence.

5. **Skip logic consolidated** -- AI-dependent tests are wrapped in a nested `test.describe('Neo Chat -- AI-dependent rendering')` with a single `beforeEach` that opens the panel and calls `isNeoAvailable()` with `test.skip()`. This eliminates the per-test skip boilerplate from the previous version.

6. **Empty state assertion tightened** -- The assertion now checks the full string `'Ask me anything about your rooms, sessions, or goals'` instead of the loose substring `'Ask me anything'`.

- The `waitForNeoUserMessage` helper is exported from the shared module but not imported or used in `neo-chat-rendering.e2e.ts`. This is fine since `neo-conversation.e2e.ts` uses it, but worth noting for completeness.

No blocking or should-fix issues remain.

lsm (commented):

This iteration is a meaningful improvement: the shared `neo-helpers.ts` extraction removes real duplication, the `data-testid="neo-sparkle-avatar"` addition is the right approach for stable locators, and consolidating `isNeoAvailable` into a single `beforeEach` is cleaner than repeating it per-test. That said, there are several issues that need attention before this is reliable.

---

### P1 — `waitForNeoAssistantResponse` secondary wait is a no-op

**File:** `packages/e2e/tests/helpers/neo-helpers.ts`, lines 83–84

```ts
await page.getByTestId(NEO_CHAT_INPUT_TESTID).waitFor({ state: 'enabled', timeout: 10000 });
```

The textarea's `disabled` attribute is bound to the component-local `sending` state (NeoChatView.tsx:394). `sending` is set to `false` in the `finally` block of `handleSend()` once `neoStore.sendMessage()` resolves. But `sendMessage()` only awaits the initial RPC dispatch (`neo.send`), not the assistant streaming. This means `sending` is `false` — and the input is `enabled` — almost immediately after the user hits Enter, long before the assistant response arrives in the DOM.

The primary `waitForFunction` already detects a new assistant message in the DOM (correct). The subsequent `waitFor({ state: 'enabled' })` adds no guarantee that streaming has finished; it passes trivially because the input is already enabled. Any assertion made immediately after `waitForNeoAssistantResponse` may race against partial or still-streaming content.

A more robust signal would be to wait until `neo-typing-indicator` is absent and the last assistant message stabilises, or to add a `data-testid="neo-streaming"` attribute to the component that is present only while streaming.

---

### P1 — Tests 2 and 6 are incorrectly gated inside the AI-dependent block

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts`, lines 82–102 and 169–180

Test 2 ("user messages appear in right-aligned blue bubble") sends a message and checks the user bubble appearance. It does not call `waitForNeoAssistantResponse` — it only waits for the user message DOM element. But user messages are not rendered optimistically; they come via LiveQuery push. So while AI connectivity is not required, a working server connection is — meaning `isNeoAvailable` (which checks for credentials, not server connectivity) is the wrong gate anyway.

More critically, test 6 ("empty state disappears once a message is sent") also does not need an assistant response — it just checks that `neo-empty-state` vanishes after a user message arrives. Both tests are inside the AI-dependent block purely by placement, not by logic.

This is a test organisation issue: tests that don't require an AI response should live in their own `describe` block or in the outer block so they run even in environments where Neo credentials are absent.

---

### P2 — Tests 4 and 5 are largely redundant

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts`, lines 125–165

Test 4 ("assistant messages render as readable text, not raw JSON") and Test 5 ("assistant messages do not show parse-error fallback") both call `waitForNeoAssistantResponse` and both assert `not.toBeVisible()` on `neo-message-parse-error`. Test 5 is a strict subset of test 4. Having two network-hitting AI tests for the same assertion doubles CI time for no additional coverage. Either fold test 5 into test 4, or remove it.

---

### P2 — `isNeoAvailable` has a time-of-check race window

**File:** `packages/e2e/tests/helpers/neo-helpers.ts`, lines 94–104

`isNeoAvailable` reads the visibility of error cards that render only after the Neo store initialisation cycle completes. The check is made synchronously after `openNeoPanel`, with no delay for the store's subscribe round-trip. In practice there's likely enough latency, but there is no explicit wait. If the error card renders with a brief delay (e.g. async `subscribe()` → error event), `isNeoAvailable` could return `true` for an instant before the error card renders, leading to AI-dependent tests running without credentials.

A safer pattern is to wait a short moment for error indicators to settle, or to wait until `neo-chat-view` reaches a stable state after panel open.

---

### P3 — Missing `waitForNeoUserMessage` import in `neo-chat-rendering.e2e.ts`

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts`, line 24–31

`waitForNeoUserMessage` is exported from `neo-helpers.ts` but not imported in `neo-chat-rendering.e2e.ts`. The tests in that file use `waitFor({ state: 'visible' })` on a locator instead. This is functionally equivalent, but for consistency with the helper naming convention it would be cleaner to import and use `waitForNeoUserMessage` where applicable.

---

### Summary of what works well

- Shared helper extraction is clean and the constants are properly exported.
- `data-testid="neo-sparkle-avatar"` is the right way to decouple the test from Tailwind classes.
- The nested `describe` block with a single `beforeEach` for the skip guard is cleaner than per-test inline skips.
- The count-based `waitForFunction` in `waitForNeoAssistantResponse` correctly avoids false positives from previous messages.
- JSON leak assertions (test 4) are exactly the right checks for the original task goal.

lsm (commented):

**Recommendation: APPROVE**

All five items from the second review round have been addressed correctly:

1. **Removed no-op `waitFor({state:'enabled'})`** -- The secondary wait after `waitForFunction` in `waitForNeoAssistantResponse` is gone. The JSDoc now correctly explains that `sending` goes false after the initial RPC dispatch, making the count-based DOM check the authoritative signal. (`packages/e2e/tests/helpers/neo-helpers.ts:66-86`)

2. **Added `waitForNeoChatReady()`** -- New helper at `neo-helpers.ts:95-105` waits for `neo-empty-state`, `neo-error-no-credentials`, or `neo-error-provider-unavailable` to be attached before `isNeoAvailable` reads visibility. The `.catch()` gracefully tolerates timeout (e.g., when messages already exist), allowing `isNeoAvailable` to fall through to its visibility checks. This eliminates the race between `openNeoPanel` and async store initialization.

3. **Tests 2 and 3 moved out of AI-dependent block** -- "User bubble style" (line 86) and "empty state disappears" (line 114) are now in the outer `describe` block. Each calls `openNeoPanel()` directly and uses `waitForNeoUserMessage` instead of relying on the inner `beforeEach`. These tests will run even when Neo credentials are absent.

4. **Redundant test removed** -- The old test 5 ("no parse-error fallback") that duplicated the `neo-message-parse-error` assertion from test 4 is gone. The assertion is retained in the "readable text" test at line 177.

5. **`neo.clearSession` in outer `beforeEach`** -- Lines 49-54 clear the session via RPC before each test, followed by `page.reload()` and `waitForWebSocketConnected()`. The `.catch(() => {})` is appropriate here since this is infrastructure cleanup where failure is non-fatal. This ensures empty-state tests (1 and 3) are stable across sequential runs.

### Remaining observation (P3)

The "user bubble style" test (lines 98-103) still asserts on Tailwind class names (`bg-blue-600`, `text-white`, `justify-end`). This is consistent with how the test was originally written and is acceptable for verifying the visual rendering contract, but a future refactor of the bubble's styling would break these assertions without any functional regression. A `data-testid` on the bubble inner div would be more resilient, similar to the `neo-sparkle-avatar` fix from the previous round.

No blocking or should-fix issues remain.

lsm (commented):

**Recommendation: APPROVE** (1 P3 nit, no blocking issues)

---

The second-round changes are solid. The five targeted fixes are all correct and the overall test file is well-structured.

### What was done well

- **Race-condition fix (`waitForNeoChatReady`)**: The new helper correctly waits for either the empty-state node or an error card to be attached to the DOM before `isNeoAvailable` reads visibility. The graceful `.catch(() => {})` on the timeout is intentional and safe — `isNeoAvailable` returns false in that case, which triggers a skip rather than a crash.
- **Test restructuring**: Moving the user-bubble-style and empty-state-disappears tests outside the AI-dependent `describe` block is the right call. They don't require a live model and would have been silently skipped in CI environments without Neo credentials.
- **Removing the redundant fallback test**: The "no parse-error fallback" assertion is retained as one check inside the primary readable-text test. Removing the standalone duplicate reduces noise without losing coverage.
- **`neo.clearSession` in outer `beforeEach`**: Using the RPC in the setup phase is consistent with CLAUDE.md's explicitly allowed infrastructure pattern. The reload after the clear is necessary to flush the in-memory signal state, and the double `waitForWebSocketConnected` correctly guards both page loads.
- **Helper extraction to `neo-helpers.ts`**: The shared helper file is clean. Constants, open/close/send/wait helpers, and `isNeoAvailable` are all correctly scoped.

### Issues

#### P3 — Misleading comment on line 84 of `neo-chat-rendering.e2e.ts`

```
// Does not require an AI response — only verifies that the user message
// bubble is rendered correctly after sendNeoMessage() creates it via RPC.
```

`sendNeoMessage` in `neo-helpers.ts` is a pure UI action (fill + Enter key press). The phrase "creates it via RPC" is wrong; it implies an RPC call is made by the helper, which contradicts both the implementation and the E2E rules. The comment should say something like "...after `sendNeoMessage()` submits it through the input" to avoid misleading future readers.

**File:** `packages/e2e/tests/features/neo-chat-rendering.e2e.ts`, line 84

---

No P0/P1/P2 issues found. The implementation is correct, the race-condition fix is properly grounded in DOM readiness rather than arbitrary timing, and all five stated changes from the second round of feedback are present and accurate.

PR #1145, opened 2026-03-28
fix(daemon): resolve SDK startup timeout after cross-provider model switch

lsm:
## Summary
- Fix symlink path mismatch in `getSDKProjectDir()` causing session file lookup failure on macOS (e.g., `/var` → `/private/var`)
- Preserve `sdkSessionId` on startup timeout instead of clearing it — keeps conversation history for resume
- Auto-retry query once on startup timeout so the user doesn't have to resend manually
- Add GLM → Anthropic resume online tests (6 tests covering both directions and model ID observation)

## Test plan
- [x] All 48 sdk-session-file-manager unit tests pass
- [x] All 76 query-runner unit tests pass
- [x] All 6 GLM → Anthropic online tests pass (sdkSessionId PRESERVED in all directions)
- [x] Typecheck passes


lsm (commented):
[P1] `No conversation found` now loops on a dead resume session

Keeping `sdkSessionId` for the `No conversation found` path prevents recovery. The next turn will still try to resume the same dead conversation because session-file validation only checks for a local JSONL file, while the user-facing error later says the session was reset automatically. In practice this can trap the session in repeated resume failures.

lsm (commented):
[P1] Silent startup-timeout retry reintroduces removed behavior

This adds a hidden retry path for startup timeouts even though the repo already has explicit no-retry regression coverage. That changes user-visible semantics, can resend the same prompt or tool turn automatically, and conflicts with the existing startup-timeout contract rather than extending it deliberately with updated tests and messaging.

PR #1153, opened 2026-03-29
chore(deps): pin all dependency versions to exact

lsm:
Audit of all `package.json` files across the monorepo confirmed no `~` or `^` range specifiers exist. All dependency versions are already exact pins — no changes required.


lsm:
Closing — audit found no violations. All dependency versions are already exact pins, so no code changes are needed and this PR has 0 additions/deletions.

PR #1156, opened 2026-03-29
feat: add NO_CHANGES_NEEDED bypass marker for no-op coding tasks

lsm:
Add `NO_CHANGES_NEEDED:` bypass marker to coder and general agent prompts, and update leader prompt's `no_pr` guidance.

- Coder/general agents now have a `NO_CHANGES_NEEDED:` marker for coding tasks where investigation shows the work is already done (deps current, pins exact, etc.)
- Includes a concrete example showing what was verified
- Leader's `no_pr` guidance updated: allows `no_pr=true` when worker used `NO_CHANGES_NEEDED:`, removes blanket restriction against it on coding tasks


lsm (commented):

---

### P0 — Critical Bug: `NO_CHANGES_NEEDED:` is never recognized at runtime

The new marker is added to all three agent prompts but is **not registered** in `BYPASS_GATES_MARKERS` in `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts`. This means `detectBypassMarker()` will always return `null` for this marker, the git/PR gate will not be bypassed, and the worker will be bounced back with a complaint about a missing feature branch and PR — the exact failure mode this PR aims to solve.

**File:** `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts`, lines 23–28

```ts
export const BYPASS_GATES_MARKERS = {
    RESEARCH_ONLY: 'RESEARCH_ONLY:',
    VERIFICATION_COMPLETE: 'VERIFICATION_COMPLETE:',
    INVESTIGATION_RESULT: 'INVESTIGATION_RESULT:',
    ANALYSIS_COMPLETE: 'ANALYSIS_COMPLETE:',
    // NO_CHANGES_NEEDED is missing here
} as const;
```

`NO_CHANGES_NEEDED: 'NO_CHANGES_NEEDED:'` must be added to this object for the feature to work end-to-end. The `BypassMarker` type and `detectBypassMarker()` derive automatically from this constant, so adding it there is sufficient.

---

### P1 — `custom-agent.ts` not updated

`packages/daemon/src/lib/space/agents/custom-agent.ts` has its own copy of the bypass markers list (lines 141–144) and was not updated with `NO_CHANGES_NEEDED:`. If custom agents can be dispatched as coder-role workers, they will lack the guidance for this new marker. For consistency the list there should also include it.

---

### P2 — Section heading still says "Research-Only Tasks"

Both `coder-agent.ts` (line 307) and `general-agent.ts` (line 85) have the section heading:

```
## Bypassing Git/PR Gates for Research-Only Tasks
```

`NO_CHANGES_NEEDED:` is explicitly for coding tasks, not research-only tasks. Keeping the heading as "Research-Only" while listing a coding-task marker under it is semantically confusing and may cause the model to under-use or misuse the marker. Consider updating the heading to "Research-Only and No-Op Coding Tasks" or similar.

---

### Summary

The P0 bug makes this PR non-functional as shipped: the runtime gate that the prompt instructs the worker to bypass will not recognize `NO_CHANGES_NEEDED:` and will reject the output. The fix is a one-line addition to `BYPASS_GATES_MARKERS`. The P1 omission in `custom-agent.ts` and the P2 heading mismatch are secondary but worth addressing in the same pass.

lsm (commented):

### P0: `NO_CHANGES_NEEDED` marker is not registered in the runtime bypass detection

The prompts in `coder-agent.ts` and `general-agent.ts` now instruct agents to use `NO_CHANGES_NEEDED:` as a bypass marker, and the leader prompt in `leader-agent.ts` tells the leader to accept `no_pr=true` when this marker is used. However, **the runtime that actually detects bypass markers does not know about this new marker**.

`packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts` (lines 23-28) defines:
```ts
export const BYPASS_GATES_MARKERS = {
    RESEARCH_ONLY: 'RESEARCH_ONLY:',
    VERIFICATION_COMPLETE: 'VERIFICATION_COMPLETE:',
    INVESTIGATION_RESULT: 'INVESTIGATION_RESULT:',
    ANALYSIS_COMPLETE: 'ANALYSIS_COMPLETE:',
} as const;
```

The `detectBypassMarker()` function (line 38) iterates over `Object.values(BYPASS_GATES_MARKERS)` to identify bypass markers in worker output. Since `NO_CHANGES_NEEDED` is missing from this object, the runtime will **never recognize** this marker. When an agent outputs `NO_CHANGES_NEEDED:`, the lifecycle hook will not detect it as a bypass, and the worker will likely fail the git/PR gate check — the exact scenario this PR is trying to fix.

**Fix**: Add `NO_CHANGES_NEEDED: 'NO_CHANGES_NEEDED:'` to the `BYPASS_GATES_MARKERS` object in `lifecycle-hooks.ts`.

### P2: `custom-agent.ts` in `space/agents/` not updated

`packages/daemon/src/lib/space/agents/custom-agent.ts` (lines 141-144) has the same bypass marker list but was not updated with `NO_CHANGES_NEEDED`. If spaces can also encounter no-op coding tasks, this would be an inconsistency. If spaces have different semantics where this doesn't apply, no action needed — but worth confirming.

### P3: Section heading is slightly misleading

In both `coder-agent.ts` (line 307) and `general-agent.ts` (line 85), the section is titled "Bypassing Git/PR Gates for Research-Only Tasks". Now that `NO_CHANGES_NEEDED` covers coding tasks where no changes are needed, this heading no longer accurately describes the full scope. Consider updating it to something like "Bypassing Git/PR Gates for Non-Change Tasks" — but this is purely cosmetic.

### Summary

The core issue is that the prompt tells the agent to use a marker that the runtime doesn't recognize. This will cause the feature to silently fail at runtime. The `BYPASS_GATES_MARKERS` constant in `lifecycle-hooks.ts` must be updated for this change to work end-to-end.

lsm (commented):

**Recommendation: APPROVE**

All three issues from the previous review have been properly addressed:

### Previously reported issues — now resolved

1. **P0 (was critical): `NO_CHANGES_NEEDED` not registered in runtime** — Fixed. The marker is now added to `BYPASS_GATES_MARKERS` in `lifecycle-hooks.ts` (line 28), so `detectBypassMarker()` will recognize it via `Object.values()` iteration.

2. **P1 (was missing from custom-agent.ts)** — Fixed. `custom-agent.ts` now includes the new marker in the bullet list, a dedicated example block, and the usage guidance paragraph, matching the other agent files.

3. **P2 (section heading inconsistency)** — Fixed. All four agent files and the two cross-references in QA/Done node prompts now use the updated heading "Research-Only and No-Op Coding Tasks".

### Remaining observation

**P3: No unit test for `NO_CHANGES_NEEDED` marker detection.**
`packages/daemon/tests/unit/room/lifecycle-hooks.test.ts` has individual test cases for `RESEARCH_ONLY`, `VERIFICATION_COMPLETE`, `INVESTIGATION_RESULT`, and `ANALYSIS_COMPLETE`, but none for the new `NO_CHANGES_NEEDED` marker. The runtime works correctly since `detectBypassMarker` iterates all values, but adding a parallel test case would maintain the pattern and catch regressions if the constant is accidentally removed.

Overall the fix commit is clean and consistent across all five files. No further blocking issues.

lsm (commented):

### Recommendation: APPROVE

This is a follow-up review after the worker addressed P0/P1/P2 feedback from the previous round.

---

### Summary of Fixes Verified

**Fix 1 — `NO_CHANGES_NEEDED` added to `BYPASS_GATES_MARKERS` constant**

Confirmed: `packages/daemon/src/lib/room/runtime/lifecycle-hooks.ts` line 28 now has:
```ts
NO_CHANGES_NEEDED: 'NO_CHANGES_NEEDED:',
```
This means `detectBypassMarker()` (which iterates `Object.values(BYPASS_GATES_MARKERS)` at lines 44-48) will correctly recognize the marker at runtime. The P0 from the prior round — the marker being present in prompts but missing from the constant so `detectBypassMarker()` would never return it — is resolved.

**Fix 2 — `custom-agent.ts` updated with new marker, description, and concrete example**

Confirmed: `packages/daemon/src/lib/space/agents/custom-agent.ts` lines 135-172 include:
- Section heading: "Bypassing Git/PR Gates for Research-Only and No-Op Coding Tasks"
- `NO_CHANGES_NEEDED:` listed among the bypass markers with the correct description
- A dedicated `**Example for NO_CHANGES_NEEDED**` block

**Fix 3 — Section heading updated in all four agent files**

Confirmed across all files:
- `coder-agent.ts` line 307: "Research-Only and No-Op Coding Tasks"
- `general-agent.ts` line 85: "Research-Only and No-Op Coding Tasks"
- `leader-agent.ts` line 171: `no_pr` doc updated to include `NO_CHANGES_NEEDED:` bypass case
- `custom-agent.ts` line 135: "Research-Only and No-Op Coding Tasks"

**Fix 4 — Leader `no_pr` guidance updated**

Confirmed: `leader-agent.ts` line 171 now reads:
> `no_pr`: set to true when the task produced no PR — use for research, investigation, meta-tasks (e.g., "create more tasks"), any work without code changes, or coding tasks where the worker used the `NO_CHANGES_NEEDED:` bypass marker (i.e., verification showed the work was already done).

The blanket restriction against using `no_pr=true` on coding tasks is removed; the new guidance correctly permits it specifically when the `NO_CHANGES_NEEDED:` marker was used.

---

### Remaining Observation (P3)

**Missing dedicated test case for `NO_CHANGES_NEEDED` in `lifecycle-hooks.test.ts`**

The `detectBypassMarker` describe block (lines 1392-1456) covers RESEARCH_ONLY, VERIFICATION_COMPLETE, INVESTIGATION_RESULT, and ANALYSIS_COMPLETE individually. There is no test case that asserts:
```ts
expect(detectBypassMarker('NO_CHANGES_NEEDED:\n\nWork already done.')).toBe(BYPASS_GATES_MARKERS.NO_CHANGES_NEEDED);
```

The existing loop-based detection ensures this works at runtime, but the absence of an explicit test means a future refactor that accidentally breaks only the `NO_CHANGES_NEEDED` path would go undetected. This is a P3 nit given the implementation is provably correct as written.

---

All previous P0/P1/P2 issues are resolved. The implementation is consistent across all four agent files and the constant/detection function.

PR #1159, opened 2026-04-01
fix: prevent PORT env var leak from killing parent daemon

lsm:
Two-part fix for the kill-chain bug where worker agent bash commands could discover and kill the parent daemon process.

- **config.ts**: Check `NEOKAI_PORT` before `PORT` so the daemon respects the port set by Playwright E2E test runner (which sets `NEOKAI_PORT`, not `PORT`).
- **query-runner.ts / provider-service.ts**: Clear `PORT` from `process.env` before the SDK subprocess query starts, then restore it after. This prevents child agent commands from inheriting the daemon's `PORT`, which could trigger an `lsof → kill` sequence when they hit `EADDRINUSE` starting a test server.

16 new unit tests added covering both fixes.


lsm (commented):

The fix correctly addresses the two root causes described in the PR:

1. **NEOKAI_PORT priority in config.ts** (line 94) -- Correct. `NEOKAI_PORT` now takes precedence over `PORT`, so when Playwright sets `NEOKAI_PORT`, the spawned daemon reads the right port instead of inheriting the parent's `PORT`.

2. **PORT env var cleared before SDK query** (query-runner.ts lines 229-234) -- Correct. The daemon's `PORT` is saved to `originalEnvVars.PORT`, then deleted from `process.env` before `query()` spawns the SDK subprocess. The `finally` block restores it via `restoreEnvVars()`.

3. **PORT restoration in provider-service.ts** (lines 698-704) -- Correct. Uses the same `hasOwnProperty` guard pattern as all other env vars.

4. **Tests** -- Both test files are well-structured and cover the key scenarios: priority ordering, restoration, deletion when undefined, and no-op when absent.

**P2: Title generation SDK query does not clear PORT**
`packages/daemon/src/lib/session/session-lifecycle.ts` line 840 also calls `query()` to generate titles. While this particular call uses `maxTurns: 1` and `tools: []` (so the agent cannot run bash commands like `lsof`), it would be more robust to also clear PORT there for defense-in-depth. The same applies to `router-agent.ts` and `security-agent.ts`, though those use `cwd: '/tmp'` with no tools, making the risk even lower. Consider centralizing the PORT cleanup into `applyEnvVarsToProcess`/`applyEnvVarsToProcessForProvider` in `provider-service.ts` so all `query()` call sites benefit automatically, rather than requiring each caller to remember to clear PORT manually.

**P3: Duplicate OriginalEnvVars interfaces**
`query-runner.ts:44` and `provider-service.ts:89` define separate `OriginalEnvVars` interfaces that must stay in sync manually. The PORT field was correctly added to both, but this is fragile. A future refactor could extract a single shared type.

### Verdict

The fix is correct and well-tested for the primary kill-chain vector (full agent sessions in query-runner.ts). The config change is clean. Recommending **APPROVE** with a suggestion to centralize the PORT cleanup into the provider-service layer for better coverage.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P1, 1 P2, 1 P3)

---

### Summary

The two-part fix is logically correct and addresses the described kill-chain. The PORT clearing/restoration flow in `query-runner.ts` is sound, the `config.ts` priority change is a one-liner with obvious correctness, and all 16 new tests pass. However, there's a maintainability defect with the duplicated `OriginalEnvVars` interface and an incomplete fix for the leak vector (only `PORT` is cleared but `NEOKAI_PORT` is not).

---

### P1 — Duplicated `OriginalEnvVars` interface will diverge

`packages/daemon/src/lib/agent/query-runner.ts` (lines 44–56) exports an `OriginalEnvVars` interface, and `packages/daemon/src/lib/provider-service.ts` (lines 89–101) has a separate private `OriginalEnvVars` interface with identical content. Both were updated in this PR to add `PORT`.

This is a fragile pattern. The `restoreEnvVars` method signature (`restoreEnvVars(original: OriginalEnvVars): void`) uses the private type from `provider-service.ts`, and TypeScript structural typing makes them compatible today — but the next person who adds a field to one will forget to update the other, and TypeScript will not catch it (the compiler only errors if a required field is missing, not if a field is present in one definition but absent in the other as `optional`).

The fix is to have a single canonical interface. The most natural approach is to export `OriginalEnvVars` from `provider-service.ts` and import it into `query-runner.ts`, since `restoreEnvVars` is the consuming method.

---

### P1 — `NEOKAI_PORT` is still leaked to SDK subprocesses

`packages/daemon/src/lib/agent/query-runner.ts` only clears `PORT` (line 234) before spawning the SDK query. `NEOKAI_PORT` is still set in `process.env` for the duration of the SDK subprocess's lifetime. A sufficiently capable agent bash command that knows to check `NEOKAI_PORT` (e.g., one that inspects `env | grep PORT`) would still find the daemon's port. The PR description identifies this as root cause #2, but the fix only addresses the `PORT` path of the kill chain.

For a complete fix, `NEOKAI_PORT` should be saved and cleared with the same save-delete-restore pattern used for `PORT`:

```ts
this.ctx.originalEnvVars.NEOKAI_PORT = process.env.NEOKAI_PORT;
delete process.env.NEOKAI_PORT;
```

And `OriginalEnvVars` + `restoreEnvVars` updated accordingly.

---

### P2 — `restoreEnvVars` skips when `originalEnvVars` is empty, but PORT may already be deleted

`packages/daemon/src/lib/agent/query-runner.ts` lines 506–514:

```ts
const originalEnvVars = this.ctx.originalEnvVars;
if (Object.keys(originalEnvVars).length > 0) {
    providerServiceRestore.restoreEnvVars(originalEnvVars);
    this.ctx.originalEnvVars = {};
}
```

In practice today this is safe because `CLAUDE_AGENT_SDK_CLIENT_APP` and `PORT` keys are always added (even with `undefined` values) before the SDK query object is created, so the object is never truly empty at `finally` time. But the early-exit guard `Object.keys(...).length > 0` is technically a guard for the wrong invariant: what matters is whether env vars were already mutated (PORT deleted), not whether the object is non-empty.

If a future refactor moves or removes the `CLAUDE_AGENT_SDK_CLIENT_APP` or `PORT` assignments, the guard could mistakenly skip `restoreEnvVars` even when PORT has been deleted. A safer guard would be `if (this.ctx.envVarsMutated)` (a boolean flag set after the mutation) or simply calling `restoreEnvVars` unconditionally (it is a no-op for empty objects since line 626 of `provider-service.ts` has an explicit early return).

---

### P3 — Test file import does not reset module between cases

`packages/daemon/tests/unit/config.test.ts` calls `await import('../../src/config')` inside each `it()` block. Bun caches ES modules by specifier, so the second and subsequent `import` calls return the already-loaded module. This is fine here because `getConfig()` reads `process.env` at call-time rather than at module-load time, but it creates a false expectation — the pattern implies module re-evaluation per test when it actually does not happen. A brief comment would prevent future confusion.

---

### What is correct

- The `config.ts` change is minimal, readable, and exactly right: NEOKAI_PORT → PORT → default.
- The PORT save/delete/restore flow is implemented correctly. `undefined` values are handled properly (port-not-set case deletes rather than sets).
- `restoreEnvVars` correctly uses `hasOwnProperty` to distinguish "key present with `undefined` value" from "key absent", which maps exactly to "port was not set" vs. "port key was never captured".
- Tests are well structured, cover the positive/negative/absent-key cases, and properly restore env state in `afterEach`.

lsm (commented):

### Re-review: All previously raised issues have been properly resolved

**P1 — Duplicate `OriginalEnvVars` eliminated**: The interface is now defined once in `provider-service.ts` (line 92) and re-exported from `query-runner.ts` (line 28). The guiding comment on line 26 is helpful for future maintainers.

**P1 — `NEOKAI_PORT` also cleared from subprocess env**: Added to the `OriginalEnvVars` interface (line 105), cleared in `saveClearDaemonPortEnvVars()` (lines 643-644), and correctly restored in `restoreEnvVars()` (lines 732-737).

**P2 — Centralized `saveClearDaemonPortEnvVars()` helper**: The private method (lines 640-645) is called from both `applyEnvVarsToProcess()` (line 527) and `clearProviderRoutingEnvVars()` (line 628), ensuring all SDK subprocess launch paths clear PORT and NEOKAI_PORT.

**Config fix**: `getConfig()` now correctly prioritizes `NEOKAI_PORT` over `PORT` (line 94: `process.env.NEOKAI_PORT || process.env.PORT || '9283'`).

**Tests**: All 22 tests pass — 8 for config port resolution and 14 for env var save/restore behavior, including NEOKAI_PORT coverage.

The `clearProviderRoutingEnvVars` return-path logic on line 629 (`changed = changed || original.PORT !== undefined || original.NEOKAI_PORT !== undefined`) is correct: when neither PORT nor NEOKAI_PORT was set in the environment, `saveClearDaemonPortEnvVars` performs no-op deletes and saves `undefined` values, so returning `{}` is safe since there is nothing to restore.

**Recommendation: APPROVE** — clean implementation, no remaining issues.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — one P0 test failure introduced by this PR.

---

### Summary

The previous P1 issues are resolved: `OriginalEnvVars` is now the single canonical definition in `provider-service.ts`, `NEOKAI_PORT` is cleared alongside `PORT` via `saveClearDaemonPortEnvVars`, and tests were expanded to 22. However, the refactor broke a pre-existing test that was NOT updated to account for the new behavior.

---

### P0 — Pre-existing test broken by the change

**File:** `packages/daemon/tests/unit/core/provider-service.test.ts:624`

```
(fail) ProviderService > applyEnvVarsToProcess > should return empty object for anthropic model
- {}
+ { "ANTHROPIC_AUTH_TOKEN": undefined, "NEOKAI_PORT": undefined, "PORT": "8399" }
```

The test at line 624 asserts `expect(original).toEqual({})` for the Anthropic provider path. Before this PR, `clearProviderRoutingEnvVars` returned `{}` unconditionally when nothing changed. Now it always records `PORT`/`NEOKAI_PORT` via `saveClearDaemonPortEnvVars`, and if PORT happens to be set (here `'8399'`, leaked from a prior test in the same file), `original.PORT !== undefined` flips `changed = true` and the non-empty object is returned.

The test needs to either:
1. Delete `process.env.PORT` (and `NEOKAI_PORT`) in `beforeEach` for this describe block, or
2. Update the assertion to accept the new shape when PORT is set (e.g. use `toMatchObject({})` or assert specific keys separately), or
3. Ensure the entire `core/provider-service.test.ts` `afterEach` restores PORT.

This will fail in CI. Verified locally:
```
bun test packages/daemon/tests/unit/core/provider-service.test.ts
53 pass, 1 fail
```

---

### P1 — Missing integration test for the actual security fix

The new `provider-service-port-restore.test.ts` tests only `restoreEnvVars` with hand-crafted `OriginalEnvVars` objects. There is no test that calls `applyEnvVarsToProcess(...)` and then asserts `process.env.PORT === undefined`. The critical code path — save PORT, delete PORT, verify it's gone — is not covered end-to-end.

Suggested test (add to `provider-service.test.ts` or the new file):

```ts
it('clears PORT from process.env when applying env vars for Anthropic', () => {
  process.env.PORT = '9283';
  const original = service.applyEnvVarsToProcess('claude-3-opus', 'anthropic');
  expect(process.env.PORT).toBeUndefined();       // security invariant
  expect(original.PORT).toBe('9283');             // saved for restoration
  service.restoreEnvVars(original);
  expect(process.env.PORT).toBe('9283');          // restored
});
```

The same should cover `applyEnvVarsToProcessForProvider` via the `generateTitleWithSdk` path.

---

### P2 — Duplicate config test location

There are now two test files for `packages/daemon/src/config.ts`:
- `packages/daemon/tests/unit/core/config.test.ts` — existing, tests general config behavior including `PORT`
- `packages/daemon/tests/unit/config.test.ts` — new, tests NEOKAI_PORT priority

The new file tests the same module but is placed outside the `core/` directory convention. The NEOKAI_PORT tests would be more maintainable if placed in the existing `core/config.test.ts` alongside the existing PORT tests.

---

### P3 — Redundant `import type` in query-runner.ts

`packages/daemon/src/lib/agent/query-runner.ts:27` has `import type { OriginalEnvVars } from '../provider-service'` which is used on line 66. This is correct. However, the comment on line 26 says "do NOT duplicate this definition here" — the definition is gone, which is good. Minor: the comment reads a little strangely next to an `import type`; it would fit better on the `export type` line 28.

---

### What's correct

- Single source of truth for `OriginalEnvVars` in `provider-service.ts` — resolved cleanly.
- `saveClearDaemonPortEnvVars` helper called on all non-early-exit paths in both `applyEnvVars` and `clearProviderRoutingEnvVars`.
- `restoreEnvVars` correctly uses `hasOwnProperty` guard to avoid spurious deletes.
- `changed` flag logic in `clearProviderRoutingEnvVars` is correct: `original.PORT !== undefined` only returns `true` when PORT was actually set.
- `config.ts` NEOKAI_PORT priority fix is correct.
- All 22 new tests pass.

lsm (commented):

All previous review feedback has been addressed correctly. 98 tests pass, typecheck is clean.

**Verified fixes:**

1. **P0 (env leak in tests)**: `PORT` and `NEOKAI_PORT` are now saved/restored in the global `beforeEach`/`afterEach` in `provider-service.test.ts` (lines 218-219, 234-235), preventing the `PORT=8399` leak that caused `toEqual({})` failures.

2. **P1 (integration coverage)**: Three new integration tests in the `applyEnvVarsToProcess` describe block (lines 700-740) verify that PORT and NEOKAI_PORT are actually cleared from `process.env`, saved in the original object, and correctly restored -- for both Anthropic and GLM providers.

3. **P2 (test consolidation)**: NEOKAI_PORT priority tests now live in `core/config.test.ts` (lines 77-113) alongside existing PORT tests. The standalone `tests/unit/config.test.ts` has been deleted.

4. **P3 (comment placement)**: Re-export comment in `query-runner.ts` (line 27) is now on its own line above the `export type` statement.

**Implementation review:**

- `saveClearDaemonPortEnvVars()` (provider-service.ts:640-645) unconditionally saves and deletes both PORT and NEOKAI_PORT, which is the correct behavior -- every SDK subprocess path must clear them.
- `clearProviderRoutingEnvVars()` (line 629) correctly updates the `changed` flag when PORT/NEOKAI_PORT were set, ensuring the caller gets back a non-empty `OriginalEnvVars` object for proper restoration.
- `restoreEnvVars()` (lines 725-738) uses `hasOwnProperty` checks for both PORT and NEOKAI_PORT, consistent with the pattern used for all other env vars.
- The `OriginalEnvVars` interface (lines 102-106) includes proper JSDoc for both fields.

No issues found.

lsm (commented):

**Recommendation: APPROVE**

This is a follow-up review verifying the previous P0 and P1 issues are fully resolved.

---

### P0 (Resolved): PORT env var leak in test isolation

**Fix verified.** `packages/daemon/tests/unit/core/provider-service.test.ts` lines 218–219 and 234–235 now include PORT and NEOKAI_PORT in the global `beforeEach`/`afterEach` save-restore cycle. The leaking `PORT=8399` that caused `clearProviderRoutingEnvVars` to capture a non-empty original and fail the `toEqual({})` assertion is now properly isolated.

---

### P1 (Resolved): Missing integration coverage for PORT/NEOKAI_PORT clearing

**Fix verified.** Three new integration tests have been added inside the `applyEnvVarsToProcess` describe block in `core/provider-service.test.ts` (lines 700–740):

- `clears PORT from process.env for Anthropic model and restores it afterward` — verifies the Anthropic code path (via `clearProviderRoutingEnvVars`)
- `clears NEOKAI_PORT from process.env for Anthropic model and restores it afterward`
- `clears PORT and NEOKAI_PORT from process.env for GLM model and restores them` — verifies the non-Anthropic code path (via `applyEnvVars` + `saveClearDaemonPortEnvVars`)

All three tests confirm the full save-clear-restore cycle, which is the exact kill-chain protection that was previously untested.

---

### P2 (Resolved): NEOKAI_PORT priority tests consolidated

NEOKAI_PORT priority and CLI override tests are now in `core/config.test.ts` (lines 77–113) alongside the existing config tests. The redundant standalone `tests/unit/config.test.ts` mentioned in the previous review has been removed (only `core/config.test.ts` exists now).

---

### Correctness analysis: `clearProviderRoutingEnvVars` return value logic

The edge case in `provider-service.ts` line 629 is sound:

```typescript
this.saveClearDaemonPortEnvVars(original);
changed = changed || original.PORT !== undefined || original.NEOKAI_PORT !== undefined;
return changed ? original : {};
```

When PORT is set, `original.PORT` is truthy, `changed` becomes true, and the full `original` object (containing PORT) is returned for restoration. When PORT is not set, `original.PORT = undefined`, the condition remains false, and if no other vars changed either, `{}` is returned — causing `restoreEnvVars` to skip entirely (via its `Object.keys(original).length === 0` early return). This is safe since there is nothing to restore.

---

### New test file: `provider-service-port-restore.test.ts`

The new standalone test file covers the `OriginalEnvVars` type contract and `restoreEnvVars` behavior for PORT and NEOKAI_PORT in isolation — including the "absent key" invariant (no `hasOwnProperty` side-effects). This is clean and well-targeted.

---

### All tests pass for the changed files

- `core/provider-service.test.ts`: 57 pass, 0 fail
- `provider-service-port-restore.test.ts`: 14 pass, 0 fail
- `core/config.test.ts`: 27 pass, 0 fail

The 14 pre-existing failures in the full unit suite are in `copilot/worktree` tests unrelated to this PR (confirmed by running the suite against the base as well).

No new issues identified. The fix is complete, correct, and well-tested.

lsm (commented):

**Recommendation: APPROVE**

The human-requested iteration — dropping the `PORT` env var fallback entirely from `config.ts` — is correct and complete. Here is the full analysis.

---

### What changed in this iteration

`packages/daemon/src/config.ts` line 94:
```ts
// Before
port: overrides?.port ?? parseInt(process.env.PORT || '9283'),

// After
port: overrides?.port ?? parseInt(process.env.NEOKAI_PORT || '9283'),
```

That single-character-cost change is the entire delta to production code. Everything else is test updates.

---

### Correctness of the config.ts change

The new resolution order is:

1. `overrides.port` (CLI `--port` flag)
2. `NEOKAI_PORT` env var
3. Default `9283`

This is correct. Every launch path in the codebase that cares about port now uses one of these two channels:

- `make dev` passes both `NEOKAI_PORT=77185PORT` **and** `--port 77185PORT`, so both channels work.
- `make run PORT=...` passes `--port ` as a CLI flag (overrides.port), so `NEOKAI_PORT` is never needed for production.
- `make serve-random` and `make self` delegate to `make run PORT=...`, same as above.
- The E2E Playwright config (`packages/e2e/playwright.config.ts:259`) sets `NEOKAI_PORT: e2ePort` and also passes `--port ${e2ePort}` on the CLI command line. Both channels are populated.

No launch path relies on a bare `PORT=` environment variable being read by `getConfig()`, so the old fallback was unnecessary and dangerous.

---

### The kill-chain protection is intact

`provider-service.ts`'s `saveClearDaemonPortEnvVars` (line 640-645) clears **both** `PORT` and `NEOKAI_PORT` from `process.env` before every SDK subprocess is spawned, and `restoreEnvVars` restores them afterward. This is called on every code path (`applyEnvVars` and `clearProviderRoutingEnvVars`). The protection was already in place from the previous iteration; this iteration only removes the now-dangerous `PORT` read that could re-introduce the leak if someone ran the daemon with a raw `PORT=` env var.

---

### Test coverage

The test suite in `packages/daemon/tests/unit/core/config.test.ts` now contains:

- `'NEOKAI_PORT sets the port'` — confirms the new env var works.
- `'PORT env var is ignored (no longer a fallback)'` — explicitly documents and enforces the new contract. This is the right test: it sets `PORT=8080`, clears `NEOKAI_PORT`, and asserts the result is the default `9283`, not `8080`.
- `'CLI port override takes precedence over NEOKAI_PORT'` — confirms priority ordering.

`packages/daemon/tests/unit/provider-service-port-restore.test.ts` and the PORT/NEOKAI_PORT block in `provider-service.test.ts` cover the clear-and-restore contract.

All tests are well-structured, use `hasOwnProperty` checks to avoid false-positive assertions on absent keys, and properly clean up after themselves via `beforeEach`/`afterEach`.

---

1. `packages/web/src/index.ts` and `packages/web/src/server-prod.ts` still read `process.env.PORT` (lines 5 in both files). These are the **web frontend dev server**, not the daemon, so they are not part of the kill-chain. They are unrelated to this fix and do not need to change.

2. The `make run` target passes PORT as a Makefile variable (not a shell env var) and forwards it to the CLI via `--port `. The daemon process started by `make run` therefore does **not** have `PORT` set in its environment, which means even the old fallback would have returned `9283` in that case. The security risk was only present in deployments or scripts that set `PORT` as a shell env var directly. The fix is still correct — it prevents that latent footgun.

3. The `logCredentialDiscovery` import in `config.test.ts` line 8 appears to be unused by the test file body (no visible call to it). This is a minor cleanliness issue but does not affect correctness.

---

### Summary

The change is minimal, precise, and correct. The Makefile contract is verified: no make target relies on `PORT=` env var being read by daemon config. The new `NEOKAI_PORT`-only fallback closes the footgun where a deployment with `PORT=` in the environment could leak the port to SDK subprocesses even after the clear-and-restore fix. Tests document the new contract explicitly.

lsm (commented):

**Recommendation: APPROVE**

The follow-up changes correctly address the reviewer feedback. The implementation is clean and complete.

### What changed

1. **`packages/daemon/src/config.ts` (line 94)**: `process.env.PORT` fallback removed. Port resolution is now strictly `overrides?.port ?? parseInt(process.env.NEOKAI_PORT || '9283')`. This eliminates the root cause of the kill-chain bug — the daemon no longer reads `PORT` from the environment for its own config, so clearing `PORT` in `saveClearDaemonPortEnvVars` cannot silently change config behavior.

2. **`packages/daemon/tests/unit/core/config.test.ts`**: Tests updated to match:
   - `PORT` references replaced with `NEOKAI_PORT` where appropriate
   - The old `NEOKAI_PORT takes priority over PORT` test (now irrelevant) replaced with a simpler `NEOKAI_PORT sets the port` test
   - New `PORT env var is ignored` test explicitly documents that PORT=8080 falls back to default 9283, not the PORT value

### Verification

- All Makefile targets (`dev`, `run`, `self`, `serve-random`, `run-e2e`) pass the port via `--port` CLI flag, which maps to `overrides.port` and takes first priority in the config chain. The `dev` target also sets `NEOKAI_PORT=$$PORT` as a secondary signal. No target relies on raw `PORT` being read by `getConfig`.
- The `provider-service.ts` save/clear/restore logic for `PORT` and `NEOKAI_PORT` is unaffected — it still correctly strips both from the subprocess environment to prevent the lsof kill-chain.
- No other code in `packages/daemon/src/` reads `process.env.PORT` for config purposes.

No issues found. The change is minimal, targeted, and well-tested.

PR #1162, opened 2026-04-01
Plan: Fix goal lifecycle reset, invalidate stale planning, fix isTerminal

lsm:
## Summary
- Plan for four goal management fixes: add `reset_goal` MCP tool, fix `isTerminal()` to include `archived`, auto-invalidate planning on description change, expose `planning_attempts` in `update_goal`
- Four tasks covering GoalManager changes, MCP tool additions, invalidation hook, and online integration test

### DIFF
--- a/docs/plans/fix-goal-lifecycle-reset-invalidate-stale-planning-fix-ister.md
+++ b/docs/plans/fix-goal-lifecycle-reset-invalidate-stale-planning-fix-ister.md
@@ -1,0 +1,183 @@
+# Fix Goal Lifecycle: Reset, Invalidate Stale Planning, Fix isTerminal
+
+## Summary
+
+When a goal's description is updated mid-planning, the planner keeps working on the old description. After cancelling the planning task, the goal gets stuck in `needs_human` with stale state (`linkedTaskIds` referencing archived tasks, `planning_attempts` blocking replanning). This plan addresses four issues:
+
+1. Missing `archived` in `isTerminal()` check within `getNextGoalForPlanning()`
+2. No way to reset a goal to its initial state without archiving
+3. Goal description updates do not invalidate in-progress planning
+4. `planning_attempts` is not writable via `update_goal`
+
+## Approach
+
+All four fixes target three existing files with well-established patterns. A new `reset_goal` MCP tool follows the same two-layer pattern (handler + schema) as existing tools. The `isTerminal` fix is a one-line change. The invalidation hook adds logic to the existing `update_goal` handler. Tests cover all changes using existing in-memory DB test patterns.
+
+---
+
+## Task 1: Fix `isTerminal()` and add `resetGoal()` to GoalManager
+
+**Description:** Fix the `isTerminal` helper in `getNextGoalForPlanning()` to include `archived`, and add a `resetGoal()` method to `GoalManager` that clears linked tasks, resets counters, and sets status to `active`.
+
+**Agent type:** coder
+
+**Depends on:** (none)
+
+**Subtasks:**
+
+1. Run `bun install` at the worktree root.
+2. In `packages/daemon/src/lib/room/runtime/room-runtime.ts`, locate `getNextGoalForPlanning()` around line 3543. Change the `isTerminal` lambda from:
+   ```ts
+   const isTerminal = (status: string) =>
+       status === 'needs_attention' || status === 'cancelled';
+   ```
+   to:
+   ```ts
+   const isTerminal = (status: string) =>
+       status === 'needs_attention' || status === 'cancelled' || status === 'archived';
+   ```
+3. In `packages/daemon/src/lib/room/managers/goal-manager.ts`, add a new `resetGoal(goalId: string)` method that:
+   - Fetches the goal (throw if not found)
+   - Calls `this.goalRepo.updateGoal(goalId, { linkedTaskIds: [], planning_attempts: 0, consecutiveFailures: 0, status: 'active' })`
+   - Returns the updated `RoomGoal`
+4. Add unit tests in `packages/daemon/tests/unit/room/goal-manager.test.ts`:
+   - Test `resetGoal` clears `linkedTaskIds`, resets `planning_attempts` to 0, resets `consecutiveFailures` to 0, and sets status to `active`
+   - Test `resetGoal` throws for non-existent goal ID
+5. Add a unit test that verifies `isTerminal` treats `archived` as terminal: in `packages/daemon/tests/unit/room/` (new file or existing edge-cases file), create a test that exercises the replanning logic with archived tasks. At minimum, add a focused test in `mission-system-edge-cases.test.ts` verifying that when all linked tasks are `archived`, the goal is eligible for replanning.
+6. Run `make test-daemon` to verify all tests pass.
+7. Run `bun run check` to verify lint, typecheck, and knip pass.
+8. Changes must be on a feature branch with a GitHub PR created via `gh pr create`.
+
+**Acceptance criteria:**
+- `isTerminal` in `getNextGoalForPlanning()` returns `true` for `archived` status
+- `GoalManager.resetGoal()` exists and resets all four fields
+- Unit tests cover both the reset method and the isTerminal fix
+- All daemon tests pass; lint and typecheck pass
+
+---
+
+## Task 2: Add `reset_goal` MCP tool and expose `planning_attempts` in `update_goal`
+
+**Description:** Add a new `reset_goal` MCP tool to `room-agent-tools.ts` and update `update_goal` to accept `planning_attempts` as an optional field. The `reset_goal` tool cancels all in-progress linked tasks via the runtime and then calls `GoalManager.resetGoal()`.
+
+**Agent type:** coder
+
+**Depends on:** Task 1
+
+**Subtasks:**
+
+1. Run `bun install` at the worktree root.
+2. In `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, in `createRoomAgentToolHandlers()`, add a new `reset_goal` handler:
+   - Takes `{ goal_id: string }` as input
+   - Fetches the goal; return error if not found
+   - Iterates `goal.linkedTaskIds`, for each task that is in a non-terminal status (`pending`, `in_progress`, `draft`, `review`, `rate_limited`, `usage_limited`):
+     - If `runtimeService` is available and returns a runtime, call `runtime.cancelTask(taskId)`
+     - Otherwise fall back to `taskManager.cancelTaskCascade(taskId)`
+   - Calls `goalManager.resetGoal(goal_id)`
+   - If `daemonHub` is available, emit `goal.updated` event and call `runtime.scheduleTick()` (if runtime exists) so the runtime picks up the reset goal for fresh planning
+   - Returns success with the updated goal
+3. Register `reset_goal` in the `createRoomAgentMcpServer()` function with a Zod schema:
+   ```ts
+   tool(
+     'reset_goal',
+     'Reset a goal to its initial state: cancels all linked tasks, clears linkedTaskIds, resets planning_attempts and consecutiveFailures to 0, and sets status to active. Use when a goal is stuck or needs a fresh start.',
+     { goal_id: z.string().describe('ID of the goal to reset') },
+     (args) => handlers.reset_goal(args)
+   )
+   ```
+4. Do NOT add `reset_goal` to `createLeaderContextMcpServer()`.
+5. In `update_goal` handler, add `planning_attempts` as an optional field:
+   - Add `planning_attempts?: number` to the handler args type
+   - Add `planning_attempts` to the "no fields provided" guard check
+   - Add `planning_attempts` to the `hasPatchFields` check
+   - Add `if (args.planning_attempts !== undefined) patch.planning_attempts = args.planning_attempts;` in the patch-building block
+6. In `update_goal` schema registration, add:
+   ```ts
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Plan Review: Fix Goal Lifecycle Reset, Invalidate Stale Planning, Fix isTerminal

Overall this is a well-structured plan with clear task decomposition and correct dependency ordering. The subtasks are specific enough to be actionable and the acceptance criteria are testable. I have a few issues to flag:

---

### P1: Missing `archived` in the *other* `isTerminal` at line 3336

The plan correctly identifies the `isTerminal` lambda at line 3543 of `room-runtime.ts` (inside `getNextGoalForPlanning`) as needing `archived`. However, there is a **second** `isTerminal` lambda at **line 3336** in the same file, used by the recurring mission execution completion check:

```ts
// line 3336
const isTerminal = (status: string) =>
    status === 'completed' || status === 'needs_attention' || status === 'cancelled';
```

This one is used to determine if all tasks in a recurring execution are terminal. If a task gets archived, the execution would never be marked complete/failed -- it would be stuck forever. The plan should address whether this second `isTerminal` also needs `archived` added, or explicitly document why it does not.

**File:** `packages/daemon/src/lib/room/runtime/room-runtime.ts`, line 3336

---

### P2: `resetGoal` should also clear `replanCount`

The plan specifies that `resetGoal()` clears `linkedTaskIds`, `planning_attempts`, `consecutiveFailures`, and sets status to `active`. However, the `goals` table also has a `replan_count` column (mentioned in CLAUDE.md's DB Tables section). If the goal is being fully reset, `replan_count` should likely be reset to 0 as well, otherwise the next planning cycle starts with a stale replan count. The plan should either include this field or explicitly justify omitting it.

---

### P2: Task 3 invalidation logic runs *after* patch -- race window

Task 3 specifies that the invalidation logic runs after `goalManager.patchGoal()`. Between the patch and the cancellation of planning tasks, the planning session could read the updated description and continue working with stale context before it gets cancelled. This is a minor race but worth noting. Consider whether the cancellation should happen before or concurrently with the patch, or document that the race is acceptable (it likely is, since the task will be cancelled shortly after regardless).

---

### P3: Task 2 subtask 4 -- negative assertion for `createLeaderContextMcpServer`

The plan says "Do NOT add `reset_goal` to `createLeaderContextMcpServer()`" which is correct per the existing exclusion list (line 1378-1380 shows `update_goal` is already excluded from leader tools). However, no test verifies this negative -- consider adding a test or acceptance criterion that confirms `reset_goal` is absent from the leader MCP server.

---

### P3: Online test (Task 4) feasibility with dev proxy

Task 4 says to run with `NEOKAI_USE_DEV_PROXY=1`. The test involves creating goals and tasks via MCP tools, which requires a running room runtime. Verifying "planning invalidation behavior" in an online test with a mocked SDK is non-trivial -- the plan should clarify what specifically will be asserted (e.g., that `planning_attempts` is 0 after description update, not that an actual planning session was cancelled mid-flight).

---

### Summary

The plan is solid and addresses the stated problems correctly. The main gap is the second `isTerminal` at line 3336 which should be evaluated for the same `archived` fix. The `replanCount` omission from reset is also worth addressing. Task ordering and scoping are appropriate.

**Recommendation: REQUEST_CHANGES** (due to P1 and P2 issues above)

lsm (commented):

This plan is well-structured and correctly identifies the four issues. The task decomposition, ordering, and acceptance criteria are generally sound. However, there are two concrete technical errors in the implementation instructions that a code agent will hit, plus gaps in scope coverage.

---

### P0 — `runtime.scheduleTick()` is a private method

**Tasks 2 and 3, instructions to "call `runtime.scheduleTick()`"**

`scheduleTick()` is declared `private` at line 4358 of `packages/daemon/src/lib/room/runtime/room-runtime.ts`. Calling it from tool handlers in `room-agent-tools.ts` will fail to compile.

The correct mechanism already exists: `cancelTask()` (called in the same reset flow) internally emits task-status events, which the runtime service subscribes to via `room.task.update` → `onTaskStatusChanged()` → `scheduleTick()`. The tick is already scheduled implicitly through the cancellation pipeline.

For the fallback path (no runtime, direct `taskManager.cancelTaskCascade`), the plan already instructs emitting `room.task.update` events — that is also sufficient to trigger a tick via the service subscription.

**Fix:** Remove the explicit `runtime.scheduleTick()` call from both Task 2 and Task 3's instructions. It is not needed and will not compile. Note instead that the tick is triggered through the task cancellation event pipeline.

---

### P0 — `goal.updated` event does not trigger a runtime tick

**Task 2, Subtask 2: "If daemonHub is available, emit `goal.updated` event and call `runtime.scheduleTick()`"**

`room-runtime-service.ts` subscribes to `goal.created` (line 837) and `room.task.update` (line 843), but has no handler for `goal.updated`. Emitting this event will not trigger a tick. The instruction implies emitting it will cause the runtime to pick up the reset goal — that is incorrect.

**Fix:** Either (a) call `runtime.onGoalCreated(goalId)` directly on the runtime instance after reset (it is a public method that internally calls the private `scheduleTick()`), or (b) add a `goal.updated` subscription to `room-runtime-service.ts` as part of Task 2's scope. Option (a) is simpler and consistent with the existing `onGoalCreated` pattern visible in `room-runtime.ts` lines 603-604.

---

### P1 — Second `isTerminal` at line 3336 is also missing `archived`

**Task 1 targets only the `isTerminal` at line 3543.** There is a second `isTerminal` at line 3336 (recurring mission execution tracker):

```ts
const isTerminal = (status: string) =>
    status === 'completed' || status === 'needs_attention' || status === 'cancelled';
```

This one also excludes `archived`. When tasks in a recurring execution are archived (e.g., via manual archive), the execution will never be marked complete and will hang. This was not identified in the plan. The plan should either fix it in scope or explicitly exclude it with a justification.

---

### P1 — Description/title update may leave a `needs_human` goal unreachable for replanning

**Task 3's invalidation logic resets `planning_attempts` but does not reset the goal's `status`.**

The bug described in the plan summary is that after cancelling planning, "the goal gets stuck in `needs_human`". Only `active` goals are iterated in `getNextGoalForPlanning()` (line 3500: `listGoals('active')`). If the goal is in `needs_human`, resetting `planning_attempts` alone will not make it eligible for replanning — a status reset to `active` is also required.

The plan should specify whether a description/title update should also transition a `needs_human` goal back to `active`, or whether this is intentionally left to the user/operator to handle via `reset_goal` or a manual status update.

---

### P2 — Task 4's online test scope is underspecified

Task 4 says to "simulate updating the goal description" and "verify planning invalidation behavior", but gives no concrete guidance on how to observe invalidation in a dev-proxy-mocked test environment where no real planning agent runs. The acceptance criterion "test passes with `NEOKAI_USE_DEV_PROXY=1`" does not clarify what assertions are meaningful. The code agent will likely struggle to write an effective test without explicit direction (e.g., assert `planning_attempts === 0` via goal fetch RPC, assert linked planning task status is `cancelled`).

---

### P3 — Tool list comment (Task 2, step 7)

Adding `reset_goal` to the comment at the top of `room-agent-tools.ts` is correct but could be folded into the handler implementation step rather than listed as a standalone step.

lsm (commented):

### Re-Review: Plan for Fix Goal Lifecycle Reset, Invalidate Stale Planning, Fix isTerminal

The plan has been substantially improved since the first review. All three P0/P1 issues from the previous reviews have been addressed:

1. **Both `isTerminal` occurrences are now fixed** (lines 3336 and 3543) -- Task 1 covers both.
2. **`replanCount` is now included** in `resetGoal()` -- Task 1, subtask 4 explicitly lists it.
3. **No private `scheduleTick()` calls** -- the plan correctly uses `runtime.onGoalCreated(goalId)` (public) and documents that `cancelTask()` already triggers a tick via its internal `scheduleTick()` call (line 2119). The architectural decision note at the top is accurate.
4. **`needs_human` to `active` status recovery** is now specified in Task 3 -- the invalidation logic transitions `needs_human` goals back to `active` when title/description changes.
5. **Task 4 online test** now has concrete test case descriptions with specific RPC-level assertions.
6. **Negative test** for `reset_goal` absence from leader MCP server is included in Task 2.

The plan is well-scoped and technically sound. A few remaining observations:

---

### P3: Task 3 invalidation -- consider resetting `replanCount` alongside `planning_attempts`

Task 3's invalidation logic resets `planning_attempts` to 0 when title/description changes, but does not mention `replanCount`. If the goal has been through replan cycles, the stale `replanCount` could affect the next planning cycle's context (e.g., replan context passed to the planner). This is minor since `reset_goal` (Task 2) does reset `replanCount`, and the invalidation path is specifically for description edits where the user wants fresh planning -- but worth noting for the implementer.

### P3: Task 2 subtask 2 -- fallback path event emission detail

The plan says for the fallback path (no runtime): "emit `room.task.update` event via `daemonHub`". This is correct in principle, but the implementer will need to construct the event payload correctly. The existing pattern in `room-runtime.ts` line 2114-2116 (`emitTaskUpdateById`) fetches the task and emits with the full task object. Since this is in a tool handler without access to `emitTaskUpdateById` (a private runtime method), the implementer will need to fetch the task after cancellation and emit the event manually. The plan could be slightly more explicit here, but this is a minor implementation detail that a competent coder agent should handle.

### P3: Task 1 subtask 7 -- testing `_doTickRecurringMissions` archived behavior

The plan asks for a unit test verifying archived tasks are terminal in `_doTickRecurringMissions`. This method is private, so the test will need to exercise it through the public `tick()` path or through a test helper. The existing test patterns in `mission-system-edge-cases.test.ts` should provide guidance, but the implementer should be aware of the indirection.

---

### Summary

The plan correctly addresses all four identified issues with appropriate task decomposition and ordering. The architectural decisions (using `onGoalCreated` instead of private `scheduleTick`, relying on cancellation event pipeline) are sound and verified against the codebase. Acceptance criteria are specific and testable. The remaining notes are all P3 nits.

**Recommendation: APPROVE**

lsm (commented):

**Recommendation: APPROVE**

All five issues raised in the previous review (two P0s, two P1s, one P2) have been addressed correctly and verifiably against the live codebase. The plan is now technically sound.

---

### Previous P0 — `runtime.scheduleTick()` is private: RESOLVED

The plan's architecture notes (line 18-19) explicitly state: "Do NOT call `runtime.scheduleTick()` — it is private." Both Task 2 (step 2, final bullet) and Task 3 (step 2, final bullet) now correctly instruct calling `runtime.onGoalCreated(goalId)` instead, which is confirmed public at line 603 of `room-runtime.ts` and internally calls `scheduleTick()`.

### Previous P0 — `goal.updated` does not trigger a runtime tick: RESOLVED

The plan now documents (architecture note, line 19) that `goal.updated` is not subscribed to by `room-runtime-service.ts`, which I verified: the service subscribes only to `goal.created` (line 837) and `room.task.update` (line 843). All invalidation paths now correctly call `runtime.onGoalCreated(goalId)` directly on the runtime instance obtained via `runtimeService.getRuntime(roomId)`, which is accessible through the existing `runtimeService` config field in `RoomAgentToolsConfig`.

### Previous P1 — Second `isTerminal` at line 3336 missing `archived`: RESOLVED

Task 1 now explicitly targets both occurrences. Step 2 fixes the `isTerminal` at ~line 3543 in `getNextGoalForPlanning()`; step 3 fixes the one at ~line 3336 in `_doTickRecurringMissions()`, with a correct rationale: archived tasks in a recurring execution would otherwise block the execution from being marked complete.

### Previous P1 — `needs_human` goal unreachable for replanning after invalidation: RESOLVED

Task 3, step 2 now explicitly includes status recovery: "If the goal's current status is `needs_human`, also update it to `active` via `goalManager.updateGoalStatus(goalId, 'active')". The rationale (line 158) correctly cites that `getNextGoalForPlanning()` only iterates `listGoals('active')` (confirmed at line 3500 of `room-runtime.ts`). The ordering is correct: status is reset to `active` before `runtime.onGoalCreated()` triggers the tick. Test coverage for this scenario is included in Task 3 step 3 and Task 4 step 3's fourth test.

### Previous P2 — Task 4's test scope underspecified: RESOLVED

Task 4 now gives four concrete test cases with specific RPC assertions (e.g., `planning_attempts === 0`, `status === 'active'`). Each test case asserts specific DB-level state via `goal.get` RPC. The run command and assertion strategy are clearly specified.

### Previous P3 — Tool list comment: RESOLVED

Step 4 of Task 2 now says "Update the tool list comment at top of file to include `reset_goal`" rather than listing it as a standalone step. Minor but tidied.

---

**`resetGoal` bypasses `updateGoalStatus` wrapper (acceptable):** Task 1 step 4 calls `goalRepo.updateGoal()` directly with `status: 'active'` in one atomic call. The `GoalManager.updateGoalStatus()` wrapper does not emit events; it only adds a "throw if not found" guard, which `resetGoal()` handles separately by fetching and checking the goal first. The single-call approach is efficient and correct.

**Two consecutive `patchGoal` calls in Task 3 (acceptable):** The invalidation logic calls `patchGoal` once for the title/description update, then again to reset `planning_attempts: 0`. These could be batched, but since `replanCount` is not reset here (it is reset only by `reset_goal`), the separation is intentional and the two-call approach is correct.

**Online test does not require LLM calls (correct use of dev proxy):** Task 4's test exercises RPC calls and DB state assertions only, with no actual planning agent invocations. Using `NEOKAI_USE_DEV_PROXY=1` is consistent with the pattern in `room-tick-job.test.ts` which similarly makes no LLM calls but still requires dev proxy.

---

The plan is technically accurate, complete, and implementable. All previous blocking issues have been resolved with correct architectural reasoning verified against the live code.

PR #1163, opened 2026-04-01
docs: plan for fixing iPad Safari layout issues

lsm:
Plan to fix two iPad Safari layout issues: header hidden behind Safari's tab bar (env(safe-area-inset-top) is 0 on iPads) and bottom gap from hardcoded pb-16 not matching actual BottomTabBar height.

Four tasks: create useViewportSafety hook with CSS custom properties, update root layout and overlays, make bottom padding dynamic via ResizeObserver, and add e2e tests for iPad viewports.

### DIFF
--- a/docs/plans/fix-ipad-safari-layout-header-behind-tabs-and-bottom-gap.md
+++ b/docs/plans/fix-ipad-safari-layout-header-behind-tabs-and-bottom-gap.md
@@ -1,0 +1,183 @@
+# Fix iPad Safari Layout: Header Behind Tabs and Bottom Gap
+
+## Goal
+
+Fix two layout issues on iPad Safari:
+
+1. **Header hidden behind Safari tabs** -- The root container uses `pt-safe` which maps to `env(safe-area-inset-top)`. On iPad (no notch), this value is 0, but Safari's tab bar overlays the top of the page, hiding the header.
+2. **Unnecessary bottom gap** -- The main content uses `pb-16 md:pb-0` for bottom tab bar clearance. On iPad Mini in portrait (744px viewport width), this falls below the `md:` (768px) breakpoint, applying 64px of padding even though the BottomTabBar uses the same `md:hidden` breakpoint -- so they match, but iPad Mini gets the mobile layout with a bottom tab bar that may not match the hardcoded 64px padding.
+
+## Approach
+
+### Header behind Safari tabs
+
+There is no CSS-only way to detect Safari's tab bar overlay. `env(safe-area-inset-top)` only accounts for the hardware notch (0 on iPads). `visualViewport.offsetTop` measures pinch-zoom offset, NOT browser chrome. The practical solutions are:
+
+1. **Use `h-svh` instead of `h-dvh`** -- `svh` (small viewport height) represents the viewport with all browser chrome visible. This prevents the container from extending behind the tab bar. The tradeoff is that when the tab bar auto-hides (e.g., scrolling in landscape), there may be a small gap at the bottom rather than filling the full screen. This is the least janky option.
+
+2. **Add a CSS custom property `--top-offset`** set by JS that detects iPad Safari via user-agent and applies a fixed top padding (approximately 50px for the Safari compact tab bar). This is fragile but addresses the specific problem.
+
+3. **Accepted limitation** -- Safari's compact tab bar on iPad intentionally overlaps content, similar to how desktop browser toolbars work. Many apps simply accept this. However, since NeoKai's header contains critical controls, we should mitigate it.
+
+**Chosen approach**: Create a `useViewportSafety` hook that:
+- Detects iPad Safari via user-agent sniffing (iPadOS reports as macOS Safari with touch support)
+- Compares `window.innerHeight` vs `document.documentElement.clientHeight` to detect if browser chrome is eating space
+- Sets CSS custom properties (`--safe-top`, `--safe-height`) on the document root
+- Updates on `resize` and `visualViewport.resize` events
+- Falls back to `env(safe-area-inset-top)` and `100dvh` when detection is not needed
+
+### Bottom gap
+
+- Replace the hardcoded `pb-16` with a CSS custom property `--bottom-bar-height` that the BottomTabBar component sets via a ref measurement
+- This ensures the padding always matches the actual rendered tab bar height (including `pb-safe` for home indicator)
+- When BottomTabBar is hidden (`md:hidden`), the custom property is set to 0
+
+### dvh reliability
+
+- Switch the root container from `h-dvh` to use the JS-computed `--safe-height` custom property, with `h-svh` as the CSS fallback
+- Keep `100dvh` on body as a general fallback but add `100svh` as the primary value
+
+## Tasks
+
+### Task 1: Create useViewportSafety hook and CSS custom properties
+
+**Description**: Create a new Preact hook that detects iPad Safari and manages CSS custom properties for safe layout dimensions. Update `styles.css` with the new custom properties and updated utility classes.
+
+**Agent type**: coder
+
+**Subtasks**:
+
+1. Create `packages/web/src/hooks/useViewportSafety.ts`:
+   - Detect iPad Safari: check for `navigator.maxTouchPoints > 1` AND `navigator.userAgent` containing "Safari" but not "Chrome" AND `navigator.platform` starting with "Mac" (iPadOS masquerades as macOS)
+   - On iPad Safari, calculate top offset: use a heuristic based on `window.screen.height - window.innerHeight` compared to expected values, or use a fixed 50px offset for the compact tab bar (this is the most reliable approach given API limitations)
+   - Set `--safe-top` CSS custom property on `document.documentElement` (0px for non-iPad-Safari, calculated value for iPad Safari)
+   - Set `--safe-height` CSS custom property: `calc(100svh - var(--safe-top))` or computed from `visualViewport.height` as a JS value
+   - Listen to `resize` and `visualViewport.resize` events, update on changes
+   - Clean up event listeners on unmount
+   - Export the hook for use in App.tsx
+
+2. Update `packages/web/src/styles.css`:
+   - Add CSS custom property defaults on `:root`: `--safe-top: env(safe-area-inset-top, 0px)`, `--bottom-bar-height: 0px`
+   - Update `.pt-safe` to use `padding-top: var(--safe-top)` (so JS can override it)
+   - Add a new utility class `.h-safe-screen` that uses `height: var(--safe-height, 100svh)`
+   - Change body height from `100dvh` (with `100vh` fallback) to `100svh` (with `100vh` fallback) to reduce jank
+
+3. Create `packages/web/src/hooks/__tests__/useViewportSafety.test.ts`:
+   - Test that on non-iPad-Safari, custom properties default to standard values
+   - Test that the hook cleans up event listeners
+   - Test iPad Safari detection logic with mocked navigator properties
+
+**Acceptance criteria**:
+- Hook correctly identifies iPad Safari via user-agent/touch heuristics
+- CSS custom properties are set on document root
+- Event listeners are properly cleaned up on unmount
+- Unit tests pass via `bunx vitest run`
+
+**Dependencies**: None
+
+Changes must be on a feature branch with a GitHub PR created via `gh pr create`.
+
+---
+
+### Task 2: Update root layout to use safe viewport dimensions
+
+**Description**: Update `App.tsx` and overlay components to use the new `useViewportSafety` hook and CSS custom properties instead of raw `h-dvh` and `pt-safe`.
+
+**Agent type**: coder
+
+**Subtasks**:
+
+1. Update `packages/web/src/App.tsx`:
+   - Import and call `useViewportSafety()` at the top of the App component (it needs to run once to set up the CSS custom properties and listeners)
+   - Change root container from `h-dvh` to `h-safe-screen` (the new utility class)
+   - Keep `pt-safe` on the root container (it now uses `var(--safe-top)` which the hook overrides)
+
+2. Update overlay components that use `h-dvh pt-safe`:
+   - `packages/web/src/components/neo/NeoPanel.tsx` line 157: change `h-dvh` to `h-safe-screen`
+   - `packages/web/src/components/room/SlideOutPanel.tsx` line 109: change `h-dvh` to `h-safe-screen`
+   - `packages/web/src/islands/ContextPanel.tsx` line 361 (mobile overlay): change `h-dvh` to `h-safe-screen`
+
+3. Update `packages/web/src/lib/__tests__/ios-safe-area.test.ts`:
+   - Update the test that checks for `pt-safe` to also verify the root uses `h-safe-screen`
+   - Add a test verifying that `styles.css` defines the `--safe-top` custom property
+   - Add a test verifying that `styles.css` defines the `.h-safe-screen` utility class
+
+**Acceptance criteria**:
+- Root container and all overlay panels use `h-safe-screen` instead of `h-dvh`
+- `pt-safe` still applied but now reads from `--safe-top` (JS-overridable)
+- Existing tests updated and passing
+- No visual regression on desktop browsers (custom properties fall back to standard values)
+
+**Dependencies**: Task 1
+
+Changes must be on a feature branch with a GitHub PR created via `gh pr create`.
+
+---
+
+### Task 3: Fix bottom tab bar padding to be dynamic
+
+**Description**: Replace the hardcoded `pb-16` with a dynamic approach that measures the actual BottomTabBar height and communicates it via a CSS custom property.
+
+**Agent type**: coder
+
+**Subtasks**:
+
+1. Update `packages/web/src/islands/BottomTabBar.tsx`:
+   - Add a `ref` to the root `div` element
+   - Use `useEffect` + `ResizeObserver` to measure the actual height of the tab bar
+   - Set `document.documentElement.style.setProperty('--bottom-bar-height', height + 'px')` whenever the height changes
+   - On unmount (or when the bar is hidden), set `--bottom-bar-height` to `0px`
+
+2. Update `packages/web/src/App.tsx`:
+   - Replace `pb-16 md:pb-0` on the main content div with an inline style: `paddingBottom: 'var(--bottom-bar-height, 0px)'`
+   - Remove the `md:pb-0` since the dynamic approach handles both cases (when BottomTabBar is hidden at md+, it sets the variable to 0)
+
+3. Update `packages/web/src/styles.css`:
+   - Ensure `--bottom-bar-height: 0px` default is on `:root` (from Task 1)
+
+4. Add test coverage:
+   - In `packages/web/src/lib/__tests__/ios-safe-area.test.ts`, add a test verifying that App.tsx no longer uses hardcoded `pb-16` for bottom padding (source code check)
+   - Verify the BottomTabBar source contains `--bottom-bar-height` custom property usage
+
+**Acceptance criteria**:
+- Bottom padding on main content matches actual BottomTabBar height dynamically
+- No hardcoded `pb-16` in the main content area
+- On desktop (md+ breakpoint), BottomTabBar is hidden and `--bottom-bar-height` is 0
+- On iPad Mini in portrait (744px), the bottom padding correctly matches the visible tab bar
+- ResizeObserver cleanup on unmount
+
+**Dependencies**: Task 1
+
+Changes must be on a feature branch with a GitHub PR created via `gh pr create`.
+
+---
+
+### Task 4: Integration testing and iPad Safari verification
+
+**Description**: Add a Playwright e2e test that verifies the layout adapts correctly at different viewport sizes, simulating iPad Safari dimensions.
+
+**Agent type**: coder
+
+**Subtasks**:
+
+1. Create `packages/e2e/tests/features/ipad-safari-layout.e2e.ts`:
+   - Test at iPad portrait viewport (820x1180 for standard iPad, 744x1133 for iPad Mini)
+   - Verify the header element is visible and not obscured (check its bounding box top position is > 0)
+   - Verify the main content area does not have excessive bottom padding at desktop viewport (1280x800)
+   - Verify at iPad Mini portrait (744px width) that bottom tab bar is visible and main content has appropriate bottom padding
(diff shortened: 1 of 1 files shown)


lsm (commented):

### Overall Assessment

This is a well-structured plan that correctly identifies the two iPad Safari layout problems and proposes reasonable solutions. The file references and line numbers are accurate against the current codebase. The task decomposition and dependency ordering are logical.

**Recommendation: APPROVE** with minor suggestions (P2/P3 only).

### Detailed Feedback

#### Problem Analysis (Good)

The plan correctly identifies:
- `env(safe-area-inset-top)` returns 0 on iPads (no notch) while Safari's tab bar still overlays content
- The `pb-16 md:pb-0` breakpoint mismatch for iPad Mini portrait (744px < 768px md breakpoint)
- The lack of CSS-only solutions for detecting Safari's compact tab bar

The three-option analysis in the Approach section (svh, JS detection, accepted limitation) with a clear rationale for the chosen hybrid approach is helpful.

#### P2: `navigator.platform` is deprecated

**Task 1, Subtask 1**: The plan proposes checking `navigator.platform` starting with "Mac" to detect iPadOS. `navigator.platform` is deprecated and may be removed from browsers. The plan should mention `navigator.userAgentData` (User-Agent Client Hints API) as a future-proofing fallback, or at minimum acknowledge the deprecation and note that `navigator.userAgent` containing "Macintosh" is the more commonly used check for this purpose. Since iPadOS already sends a macOS user-agent string, the `navigator.userAgent` check alone (which the plan also mentions) may be sufficient without `navigator.platform`.

#### P2: Missing consideration of `window.innerHeight` vs `visualViewport.height` reliability

**Task 1**: The plan proposes comparing `window.innerHeight` vs `document.documentElement.clientHeight` to detect browser chrome eating space, but on iPad Safari these values are often identical. The plan should clarify what specific differential is expected and under what conditions. If the differential is unreliable (which it often is on modern Safari), the fallback to a fixed ~50px offset needs to be the documented primary strategy, not the heuristic comparison.

#### P2: E2E test limitations not acknowledged

**Task 4**: The plan proposes verifying header visibility and bottom padding at iPad viewports in Playwright, but Playwright's viewport emulation does NOT replicate Safari's compact tab bar overlay behavior. The test would set viewport dimensions but the browser chrome overlay that causes the original problem would be absent. The plan should acknowledge this limitation and clarify what the e2e test can actually verify (correct CSS class usage, correct custom property values, correct responsive behavior) vs. what requires manual device testing.

#### P3: ContextPanel line reference

**Task 2, Subtask 2**: The plan references ContextPanel.tsx line 361 for the `h-dvh` change, but the `h-dvh` is actually on line 358 (`h-dvh md:h-full w-70`). Line 361 has `pt-safe md:pt-0`. Minor nit since the correct file and intent are clear.

#### P3: ContextPanel test not mentioned in Task 2

The plan changes the root and overlay panels from `h-dvh` to `h-safe-screen`, but the ContextPanel test at `packages/web/src/islands/__tests__/ContextPanel.test.tsx:611` explicitly asserts `h-dvh`. The plan's Task 2 mentions updating `ios-safe-area.test.ts` but does not mention updating the ContextPanel test. This would cause a test failure.

#### P3: ResizeObserver vs CSS `display: none` clarification

**Task 3**: The plan says to set `--bottom-bar-height` to `0px` on unmount. Since the BottomTabBar uses `md:hidden` (CSS), the component is not actually unmounted at the `md` breakpoint -- it is just hidden via CSS `display: none`. The ResizeObserver approach should handle the case where the element's height becomes 0 when `display: none` is applied, rather than relying on unmount cleanup. ResizeObserver does report 0 dimensions for hidden elements, so this works correctly in practice, but the plan's description is slightly misleading about the mechanism.

### Summary

The plan is sound and ready for execution. The main suggestions are: (1) acknowledge `navigator.platform` deprecation and prefer the userAgent-based detection, (2) be more explicit about which heuristics are primary vs. fallback for detecting Safari chrome, and (3) acknowledge Playwright's inability to simulate Safari tab bar overlay in e2e tests. None of these are blockers.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (same-author comment review)

The plan is well-structured and addresses both root causes correctly. The technical diagnosis is accurate — `env(safe-area-inset-top)` is indeed 0 on iPads with no notch, and the tab-bar overlay has no CSS-only solution. The ResizeObserver approach for bottom padding is the right call. That said, there are several issues that need to be addressed before execution begins.

---

### P1 — Core technical flaw: `--safe-top` detection heuristic is unreliable

**Task 1, Subtask 1** — The plan proposes using `window.screen.height - window.innerHeight` to detect the Safari compact tab bar height, with a fallback to a fixed 50px offset. Neither approach is reliable:

- `window.screen.height - window.innerHeight` changes with zoom level, OS UI scaling, and when the keyboard is visible. It will return incorrect values in many legitimate scenarios.
- A hardcoded 50px is correct for the current compact tab bar, but Apple has changed this value across Safari versions (it was 44px in earlier versions, ~50px in recent ones). This will silently break on future iOS/iPadOS updates.
- `visualViewport.height` is the correct API — it gives the visible content area after all browser chrome is subtracted. The plan mentions it in the Approach section but then proposes `window.innerHeight` comparisons in the actual hook implementation. These two are inconsistent.

**Recommendation**: The hook implementation in Task 1, Subtask 1 should anchor on `visualViewport.height` (not `window.innerHeight` arithmetic) as the primary source for `--safe-height`. For `--safe-top`, `visualViewport.offsetTop` is still described as "pinch-zoom offset only" in the plan's Approach section — which is correct — so there is genuinely no reliable CSS API for iPad Safari tab-bar inset. The plan should explicitly state that `--safe-top` is set to `0px` always (matching `env(safe-area-inset-top)` which is also 0) and that the header-behind-tab-bar problem is addressed exclusively via `--safe-height` (capping the container height to `visualViewport.height`), not via top padding. The Approach section and the Task 1 implementation section contradict each other on this point.

---

### P1 — ContextPanel has `h-dvh md:h-full` — replacing only the mobile branch is risky

**Task 2, Subtask 2** — The plan says to change `h-dvh md:h-full` to `h-safe-screen` on `ContextPanel.tsx` line 358. But `h-safe-screen` would override the `md:h-full` breakpoint rule too, since it would be a single class with no responsive modifier. The current code intentionally uses `h-full` on desktop (where the panel sits inside the layout flow) and `h-dvh` on mobile (where it overlays). Replacing both with one utility class will break the desktop layout unless `h-safe-screen` itself has a responsive variant or the replacement is `h-safe-screen md:h-full`. The plan does not account for this.

---

### P1 — Inline style vs Tailwind mismatch in Task 3

**Task 3, Subtask 2** — The plan changes `pb-16 md:pb-0` (Tailwind classes) to `paddingBottom: 'var(--bottom-bar-height, 0px)'` (inline style). This is inconsistent with the rest of the codebase which uses Tailwind utilities everywhere. It also bypasses Tailwind's media-query system, which means if any other class or component tries to override bottom padding via a responsive Tailwind utility later, the inline style will win due to specificity. A Tailwind arbitrary-value class like `pb-[var(--bottom-bar-height,0px)]` or a custom utility in `styles.css` (`.pb-bottom-bar { padding-bottom: var(--bottom-bar-height, 0px); }`) would be more consistent.

---

### P2 — Task 4 E2E test approach cannot actually verify the core bug

**Task 4, Subtask 1** — The plan says to "verify the header element is visible and not obscured (check its bounding box top position is > 0)". Playwright running in Chromium (the default) does not emulate Safari's tab bar overlay. `boundingClientRect.top > 0` will pass trivially in any browser without a real Safari tab bar present. This test cannot verify the actual bug without either:

1. Running Playwright with `browserName: 'webkit'` AND emulating iPad Safari specifically (which requires the `viewport` + `hasTouch` + user-agent combination), or
2. Testing that the CSS custom property `--safe-height` is set to a value less than `window.innerHeight` when the hook detects iPad Safari (unit test in jsdom is better suited here).

The plan should clarify whether webkit is used in the e2e test and what the realistic verifiability goal is. As written, the acceptance criterion "Header visibility is verified at each viewport size" will be trivially satisfied without actually catching the bug.

---

### P2 — `navigator.platform` deprecation

**Task 1, Subtask 1** — The detection logic relies on `navigator.platform` starting with `"Mac"`. `navigator.platform` was deprecated in the WHATWG spec and returns `""` in some newer browser contexts. The plan should use `navigator.userAgentData.platform` (with fallback to `navigator.platform`) or use the `navigator.maxTouchPoints > 1 && /Safari/.test(navigator.userAgent) && !/Chrome/.test(navigator.userAgent)` fingerprint without relying on `navigator.platform` at all, since `maxTouchPoints > 1` already distinguishes iPadOS from macOS Safari on non-touch Macs.

---

### P2 — Missing: what happens when `BottomTabBar` is conditionally re-mounted

**Task 3, Subtask 1** — The `BottomTabBar` is `md:hidden` via CSS, but it is always mounted in the DOM (it uses `flex md:hidden` on its root div). So the ResizeObserver will fire at mount even at desktop widths — but the element's `offsetHeight` will be 0 when `display: none` is in effect via CSS. The plan should verify this: does `ResizeObserver` fire when the element transitions from `display: none` to visible (e.g., resizing from desktop to mobile)? `ResizeObserver` does NOT fire for elements with `display: none`. This means `--bottom-bar-height` will be set to 0 at mount (correct for desktop), but if the user resizes from desktop to mobile, the observer will not fire until the element actually becomes visible again. The plan should address whether an additional `resize` window event listener is needed to re-measure after breakpoint transitions.

---

### P3 — Task 2 line-number references may drift

Task 2 references specific line numbers (`NeoPanel.tsx` line 157, `SlideOutPanel.tsx` line 109, `ContextPanel.tsx` line 361). These are accurate against the current `dev` branch, but if any of tasks 1-4 touch these files, the line numbers will shift. The executing agent should use `grep` to locate the correct lines rather than relying on hardcoded numbers.

---

### P3 — `useViewportSafety` hook returns nothing but mutates global state

The hook mutates `document.documentElement.style` as a side effect with no return value. This is a valid pattern (see `useNeoKeyboardShortcut` which similarly has no return value), but the plan should note that the hook must only be called once globally (it already says "import and call in App.tsx"), and downstream components must NOT call it themselves. A comment in the hook file to that effect would prevent misuse.

---

### Summary of what's good

- The overall 4-task decomposition is sensible: hook first, layout second, bottom-bar third, e2e fourth.
- Dependencies (T2 and T3 both depend on T1; T4 depends on T2+T3) are correctly ordered.
- Using `visualViewport.resize` alongside `window.resize` is the right set of events.
- The `--bottom-bar-height` ResizeObserver approach is correct for the bottom gap issue.
- The existing `ios-safe-area.test.ts` is correctly identified as the place to expand unit tests.

lsm (commented):

### Re-review: Assessment of Updated Plan

The updated plan has thoroughly addressed the P1 and P2 issues from both previous reviews. This is a significant improvement over the original version.

### Previous Issues -- Resolution Status

All **P1** issues from the GLM review are resolved:
- **`--safe-top` heuristic unreliable**: Plan now explicitly states no `--safe-top` property is used. The header problem is solved exclusively via `--safe-height` (capping container to `visualViewport.height`). The Approach and Task 1 sections are now consistent.
- **ContextPanel `md:h-full` breakage**: Task 2 Subtask 2 now has a bold IMPORTANT note specifying `h-safe-screen md:h-full` -- preserving the desktop responsive modifier.
- **Inline style vs Tailwind mismatch**: Plan now defines `.pb-bottom-bar` custom utility in `styles.css` instead of inline styles.

All **P2** issues from both reviews are resolved:
- **`navigator.platform` deprecation**: Plan explicitly says "Do NOT use the deprecated `navigator.platform` API" and relies on `maxTouchPoints > 1` + UA string.
- **`visualViewport.height` vs `window.innerHeight`**: Plan now anchors on `visualViewport.height` as the primary source throughout. The inconsistency between Approach and implementation sections is gone.
- **E2E test limitations**: Task 4 has a dedicated "Limitations acknowledged" paragraph noting Playwright cannot replicate Safari's tab bar overlay, and that manual device testing is required.
- **ResizeObserver + `display: none` transition**: Task 3 now explicitly adds a supplementary `window.resize` listener to re-measure after breakpoint transitions.

All **P3** issues are resolved:
- ContextPanel test update is now in Task 2, Subtask 3.
- Hook singleton usage note is specified.
- Line number references now say "use grep" for most locations.

### Remaining Observations

#### P3: `visualViewport` Safari quirks note could be more actionable

The Approach section mentions `visualViewport.height` has "known quirks (jank during toolbar animation, stale values on keyboard dismiss in iOS 26)" but does not specify mitigation. A debounce/throttle on the resize handler (e.g., 100ms) could reduce jank from rapid toolbar animation updates. This is minor since the CSS fallback provides safety, but worth noting for the implementer.

#### P3: Body height change rationale is thin

Task 1 Subtask 2 says to change body height from `100dvh` to `100svh` "to reduce jank." The relationship between `dvh` (dynamic, changes with address bar) and jank on the body element is worth a one-line explanation for the implementer. `dvh` causes layout recalculation as the address bar animates; `svh` is static (smallest viewport) and avoids this. The plan states the change but not the reasoning clearly enough for an unfamiliar implementer to understand why.

#### P3: Task 3 acceptance criterion "Breakpoint transitions correctly update" is hard to verify in unit tests

The acceptance criteria for Task 3 include "Breakpoint transitions (desktop to mobile) correctly update `--bottom-bar-height`." This depends on CSS media queries affecting `display: none`, which vitest/jsdom cannot simulate. The plan should note this criterion is verified by the e2e test in Task 4, not by unit tests.

### Overall Assessment

**Recommendation: APPROVE**

The plan is well-structured, technically sound, and has addressed all substantive feedback from both previous reviews. The four-task decomposition has correct dependencies. The `visualViewport.height` approach is the right solution for the header problem -- it avoids unreliable heuristics and has a clean CSS fallback. The `.pb-bottom-bar` utility and ResizeObserver + window resize listener approach for the bottom gap is solid. The e2e test limitations are honestly acknowledged. Only P3 nits remain.

lsm (commented):

**Recommendation: APPROVE** (same-author comment review)

All three P1 issues from the previous review have been fully resolved. The plan is now ready for execution.

---

### P1 resolutions — all three addressed

**P1 #1 — `--safe-top` heuristic / `visualViewport.height` anchoring**: RESOLVED.

The updated plan explicitly anchors on `window.visualViewport.height` as the primary source for `--safe-height` in both the Approach section and Task 1, Subtask 1. The `--safe-top` property is completely eliminated. The Approach and Task 1 are now fully consistent: the header-behind-tab-bar problem is solved exclusively by capping the container height to `visualViewport.height`, not by any top-padding injection. The previous inconsistency between the Approach section (mentioning `visualViewport`) and the implementation direction (using `innerHeight` arithmetic) is gone.

**P1 #2 — ContextPanel `h-dvh md:h-full` desktop breakpoint preservation**: RESOLVED.

Task 2, Subtask 2 (line 101) now explicitly says: "Change to `h-safe-screen md:h-full` (only replace the mobile `h-dvh` portion, keep the `md:h-full` breakpoint rule)." The `md:h-full` modifier is explicitly called out as MUST be preserved, and the acceptance criteria at line 109 reiterates this. The previous omission that would have broken desktop layout is fully addressed.

**P1 #3 — Inline style vs Tailwind-first pattern**: RESOLVED.

Task 1, Subtask 2 (line 63) now adds `.pb-bottom-bar` as a CSS custom utility in `styles.css`. Task 3, Subtask 2 (lines 136–138) explicitly says to use "the new `.pb-bottom-bar` utility class" and labels it "Tailwind-consistent custom utility — not an inline style". The `md:pb-0` removal rationale is also included (dynamic approach handles both desktop and mobile cases).

---

### P2 — Body height change from `dvh` to `svh` needs more explanation

**docs/plans/fix-ipad-safari-layout-header-behind-tabs-and-bottom-gap.md, line 64**

The plan changes body height from `100dvh` (with `100vh` fallback) to `100svh` (with `100vh` fallback) and says this "reduces jank". This is a behavior change on all browsers, not just iPad Safari. `100dvh` uses the dynamic viewport height (which shrinks when Chrome's address bar collapses), while `100svh` uses the small viewport height (the most conservative, with all browser chrome visible). On Chrome desktop and Chrome mobile, switching from `dvh` to `svh` on the body means the body is sized to the smallest possible viewport (all chrome visible), which can leave a visual gap at the bottom when the address bar collapses. The plan should clarify: (a) whether this is intentional for all browsers, or only for the iPad Safari case, and (b) the tradeoff accepted when making this body-level change. This is a global change that warrants explicit justification.

---

### P2 — E2E test reads JS-set CSS custom property (borderline per CLAUDE.md)

**docs/plans/fix-ipad-safari-layout-header-behind-tabs-and-bottom-gap.md, Task 4, Subtask 1**

Task 4 asks the e2e test to "Verify at iPad Mini portrait (744px width) that `--bottom-bar-height` CSS custom property is set to a non-zero value." Reading a CSS custom property set by JS requires `page.evaluate(() => getComputedStyle(document.documentElement).getPropertyValue('--bottom-bar-height'))`. CLAUDE.md says "All assertions must verify visible DOM state: text content, element visibility, CSS classes." A JS-set CSS custom property is not directly "visible DOM state" in the same sense as text content or element visibility. This is a gray area — the property is observable via DOM APIs — but it reads internal JS-managed state rather than user-visible output.

A more conservative test would verify the actual computed `padding-bottom` on the main content element (which IS visible layout behavior) rather than reading the raw custom property value. The plan should clarify the intent here and acknowledge this is reading internal implementation state, not user-visible behavior.

---

### P3 — "does not have hardcoded `pb-16`" in e2e is redundant

**Task 4, Subtask 1**: "Verify the main content area does not have hardcoded `pb-16` padding at any viewport size" — this is a static source-code check that already appears in Task 3's unit tests (Task 3, Subtask 4: "add a test verifying that App.tsx no longer uses hardcoded `pb-16`"). The e2e version of this check would inspect DOM class lists at runtime. This is harmless but the plan should clarify whether this is a runtime class check or a static string check, to avoid the implementing agent writing a brittle DOM-class assertion that misses the intent.

---

### P3 — `window.resize` listener in BottomTabBar may fire too early

**Task 3, Subtask 1**: The plan says to add a `window.resize` listener that "re-measures the ref element's `offsetHeight` after breakpoint transitions." The `window.resize` event fires synchronously with the resize, before layout is fully settled (especially during CSS `@media` breakpoint transitions where Tailwind's `md:hidden` resets from `display: none` to `flex`). A `requestAnimationFrame` or a small debounce may be needed to ensure `offsetHeight` is read after the browser has applied the new display property. The plan does not address this timing concern. The implementing agent should be aware of this edge case.

---

### Summary of what remains good

- All three P1 issues have been properly resolved with clear, actionable instructions.
- The `navigator.platform` deprecation (previous P2) is explicitly addressed: the plan now uses `maxTouchPoints + UA string` only.
- The ResizeObserver + `display: none` gap (previous P2) is now addressed with an explicit supplementary `window.resize` listener.
- The e2e limitations (previous P2) are now acknowledged in Task 4 with a "Limitations acknowledged" section.
- The global-singleton concern for `useViewportSafety` (previous P3) is now addressed with a required JSDoc comment.
- The ContextPanel test update (previous P3) is now explicitly included in Task 2, Subtask 3.
- Task dependency ordering (T1 → T2/T3 → T4) remains correct.

lsm (commented):

**Recommendation: APPROVE** (same-author comment review)

This is a focused re-review of the two P2 issues raised in the previous round. Both are now resolved.

---

### P2 #1 — Body height change from `dvh` to `svh` needs cross-browser justification: RESOLVED

**Task 1, Subtask 2 (line 64)** now includes explicit Rationale and Tradeoff documentation:

- Rationale: `dvh` triggers layout recalculation every time the browser toolbar animates in/out (scrolling on mobile Safari/Chrome), causing visible jank. `svh` is static (smallest viewport, all browser chrome visible), so recalculations do not occur.
- Tradeoff explicitly acknowledged: on all mobile browsers (not just iPad Safari), when the address bar collapses, there may be a small gap at the bottom rather than the page expanding to fill the full screen. The plan states this is intentional, with the root container's height managed by `.h-safe-screen` as the primary mechanism.

The change is a deliberate design decision with clear cross-browser implications documented for the implementer. This is sufficient justification.

---

### P2 #2 — E2E test reads internal CSS custom property rather than visible layout behavior: RESOLVED

**Task 4, Subtask 1 (lines 177-180)** now explicitly uses `getComputedStyle().paddingBottom` instead of reading `--bottom-bar-height` directly:

> "Verify at iPad Mini portrait (744px width) that bottom tab bar is visible and the main content area has a non-zero computed `padding-bottom` (use `getComputedStyle().paddingBottom` — this checks visible layout behavior, not internal custom property state)"

The test guidance also reinforces CLAUDE.md compliance: "All assertions must verify visible DOM state per CLAUDE.md e2e rules: element visibility, bounding boxes, CSS classes, computed styles — not raw JS-set custom property values". `getComputedStyle().paddingBottom` reflects visible layout state and is not an internal signal read, so the borderline concern from the previous review is fully addressed.

---

### No new issues

No new issues were introduced in the revision. The plan remains technically sound on all previously reviewed points. Manual iPad Safari device testing is still required for the header overlay fix, and the plan correctly acknowledges this limitation in Task 4.

lsm (commented):

**Recommendation: APPROVE**

### Re-review Summary

The updated plan has addressed all P2 feedback from the previous round:

1. **Body `dvh` to `svh` rationale** (previously P2 from GLM): Task 1 Subtask 2 now includes a detailed rationale explaining why `dvh` causes jank and why `svh` is preferred, along with an explicit acknowledgment of the tradeoff (small bottom gap on mobile when address bar collapses). This is clear and actionable for the implementer.

2. **E2E test reading CSS custom properties** (previously P2 from GLM): Task 4 now uses `getComputedStyle().paddingBottom` to verify visible layout behavior instead of reading raw `--bottom-bar-height` custom property values. This aligns with CLAUDE.md's e2e rules about verifying visible DOM state.

3. **`requestAnimationFrame` timing for resize handler** (previously P3 from GLM): Task 3 Subtask 1 now explicitly specifies using `requestAnimationFrame` inside the window resize handler to ensure the browser has applied the new display property before reading `offsetHeight`.

### Current State

The plan is technically sound and complete. All four tasks have correct dependencies, clear acceptance criteria, and actionable subtask descriptions. No P0, P1, or P2 issues remain.

### Remaining Observations

**P3**: The `visualViewport` Safari quirks note in the Approach section mentions "jank during toolbar animation, stale values on keyboard dismiss in iOS 26" but does not suggest a debounce/throttle on the resize handler. This is a minor implementation detail the executing agent can decide on -- the CSS `100svh` fallback provides safety regardless.

**P3**: Task 1 Subtask 1 lists the iPad Safari detection criteria twice (once in Approach, once in Task 1). This is intentional for clarity but adds some redundancy. Not an issue.

PR #1164, opened 2026-04-01
Fix isTerminal to include archived status and add GoalManager.resetGoal()

lsm:
Fix two `isTerminal` helper bugs and add a `resetGoal()` method to `GoalManager`.

**Changes:**
- `_doTickRecurringMissions`: add `archived` to terminal statuses so recurring executions complete instead of hanging when tasks are archived
- `getNextGoalForPlanning`: add `archived` to terminal statuses so goals with all-archived linked tasks are eligible for re-planning
- `GoalManager.resetGoal()`: new method that clears `linkedTaskIds`, resets `planning_attempts`/`consecutiveFailures`/`replanCount` to 0, and sets status back to `active`

Unit tests added for all three changes.

### DIFF
--- a/packages/daemon/src/lib/room/managers/goal-manager.ts
+++ b/packages/daemon/src/lib/room/managers/goal-manager.ts
@@ -187,6 +187,31 @@ export class GoalManager {
 	}
 
 	/**
+	 * Reset a goal to its initial state so it can be re-planned from scratch.
+	 * Clears linked tasks, resets all counters, and sets status back to active.
+	 */
+	async resetGoal(goalId: string): Promise<RoomGoal> {
+		const goal = await this.getGoal(goalId);
+		if (!goal) {
+			throw new Error(`Goal not found: ${goalId}`);
+		}
+
+		const updatedGoal = this.goalRepo.updateGoal(goalId, {
+			linkedTaskIds: [],
+			planning_attempts: 0,
+			consecutiveFailures: 0,
+			replanCount: 0,
+			status: 'active',
+		});
+
+		if (!updatedGoal) {
+			throw new Error(`Failed to reset goal: ${goalId}`);
+		}
+
+		return updatedGoal;
+	}
+
+	/**
 	 * Link a task to a goal
 	 */
 	async linkTaskToGoal(goalId: string, taskId: string): Promise<RoomGoal> {
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -3334,7 +3334,10 @@ export class RoomRuntime {
 			if (validTasks.length === 0) continue;
 
 			const isTerminal = (status: string) =>
-				status === 'completed' || status === 'needs_attention' || status === 'cancelled';
+				status === 'completed' ||
+				status === 'needs_attention' ||
+				status === 'cancelled' ||
+				status === 'archived';
 			const allTerminal = validTasks.every((t) => isTerminal(t.status));
 			if (!allTerminal) continue;
 
@@ -3541,7 +3544,7 @@ export class RoomRuntime {
 				);
 				const executionTasks = validTasks.filter((t) => t.taskType !== 'planning');
 				const isTerminal = (status: string) =>
-					status === 'needs_attention' || status === 'cancelled';
+					status === 'needs_attention' || status === 'cancelled' || status === 'archived';
 				const allExecutionFailed =
 					executionTasks.length > 0 && executionTasks.every((t) => isTerminal(t.status));
 				const allFailed = validTasks.length > 0 && validTasks.every((t) => isTerminal(t.status));
--- a/packages/daemon/tests/unit/room/goal-manager.test.ts
+++ b/packages/daemon/tests/unit/room/goal-manager.test.ts
@@ -1011,3 +1011,109 @@ describe('GoalManager.patchGoal — schedule nextRunAt auto-computation', () => 
 		expect(patched.nextRunAt).toBe(originalNextRunAt);
 	});
 });
+
+describe('GoalManager.resetGoal', () => {
+	let db: Database;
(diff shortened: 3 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

All four deliverables are correctly implemented:

### 1. isTerminal in `_doTickRecurringMissions()` (line 3336-3340)
Added `archived` to the terminal status check alongside `completed`, `needs_attention`, and `cancelled`. This ensures recurring executions with all-archived tasks are properly marked as failed rather than hanging indefinitely.

### 2. isTerminal in `getNextGoalForPlanning()` (line 3546-3547)
Added `archived` to the failure-terminal check. Note this lambda intentionally excludes `completed` (unlike the recurring one) since it only checks for failure/cancellation states that should trigger replanning. The `archived` addition is correct here.

### 3. `GoalManager.resetGoal()` (line 193-212)
Clean implementation that:
- Validates goal existence before updating
- Clears `linkedTaskIds` to `[]`
- Resets `planning_attempts`, `consecutiveFailures`, `replanCount` to 0
- Sets status back to `active`
- All fields match `UpdateGoalParams` interface correctly

### 4. Tests
- 3 unit tests for `resetGoal` covering: full reset from dirty state, `needs_human` to `active` transition, and non-existent goal error
- 3 tests for isTerminal fixes covering: archived tasks triggering replanning, mixed archived+active tasks not replanning, and archived tasks completing recurring executions

All 6 new tests pass.

### Minor observations (P3)

1. **`progress` not reset in `resetGoal`**: When linked tasks are cleared, the `progress` field retains its old value until new tasks are linked and progress is recalculated. This is acceptable since progress auto-recalculates when tasks are linked, but the intermediate state could be confusing if read between reset and replanning.

2. **Duplicated `isTerminal` lambdas**: There are now two inline `isTerminal` lambdas with slightly different semantics (one includes `completed`, one doesn't). Consider extracting these as named helpers with distinct names (e.g., `isTaskTerminal` vs `isTaskFailed`) to make the semantic difference explicit and prevent future drift. Not blocking.

lsm (commented):

**Recommendation: APPROVE** — the implementation is correct and well-tested. A few worth-noting observations below.

---

### isTerminal fixes in room-runtime.ts

Both fixes are correct and symmetric with the existing `STALE_TASK_STATUSES` constant at line 96 that already includes `archived`. The original lambdas were inconsistently narrower than the rest of the codebase's understanding of "terminal."

**_doTickRecurringMissions (line 3336-3340):** Adding `archived` to the completion check is correct. A recurring execution that has only archived tasks was previously hung forever because `allTerminal` evaluated false. The fix matches the intent of `STALE_TASK_STATUSES`.

**getNextGoalForPlanning (line 3546-3547):** Adding `archived` here is also correct. Goals whose linked tasks were all archived were previously excluded from replanning because `isTerminal` returned false for archived tasks. The fix unblocks those goals.

One minor observation: the two `isTerminal` lambdas are defined inline rather than sharing the module-level `STALE_TASK_STATUSES` constant. The first lambda (`_doTickRecurringMissions`) has 4 statuses including `completed`; the second (`getNextGoalForPlanning`) has 3 (no `completed`), which is intentional since `completed` tasks should not trigger replanning. The asymmetry is correct by design but a brief inline comment explaining the exclusion of `completed` from the second lambda would reduce future confusion.

---

### GoalManager.resetGoal()

**packages/daemon/src/lib/room/managers/goal-manager.ts:193-212**

The implementation is clean and correct for the intended use case. A few observations:

**P2 — No guard against resetting `completed` or `archived` goals.** `resetGoal` will happily move a `completed` or `archived` goal back to `active` without any validation. The existing `reactivateGoal` method has the same gap, so this is consistent with the existing pattern — but `resetGoal` goes further by also clearing `linkedTaskIds`, making the side effects more severe if called on a completed goal that carries historical task references. Since the plan doc describes this as a tool for human-initiated "reset from scratch" workflows, consider adding a guard (or at minimum a JSDoc note) clarifying which source states are valid. The DB `CHECK` constraint on `goals.status` allows `active` as a target from any state, so the DB itself won't protect against this.

**P3 — `resetGoal` is not exposed via RPC.** The planning issue (#1162) specifically requested a `reset_goal` MCP tool and RPC exposure. This PR only adds the backend method. The `GoalManagerLike` interface in `goal-handlers.ts` (line 40-60) and `goal-handlers.ts` header comment do not include `resetGoal`. If a follow-up PR is planned for the RPC surface, that should be noted. If not, the method will be dead code from the API perspective.

**P3 — `goal_review_attempts` is not reset.** The method resets `planning_attempts`, `consecutiveFailures`, and `replanCount` but leaves `goal_review_attempts` untouched. If `goal_review_attempts` tracks review-phase retries alongside `planning_attempts`, omitting it may leave stale state after a reset. Low risk if `goal_review_attempts` is not used in replanning gating logic, but worth a comment either way.

---

### Tests

The unit tests in `goal-manager.test.ts` (lines 1015-1119) and `mission-system-edge-cases.test.ts` (sections 8 and 9) are thorough and cover the happy paths and key edge cases:

- `resetGoal` with pre-populated tasks, counters, and `needs_human` status
- `needs_human -> active` transition
- Non-existent goal ID error path
- All-archived tasks triggering replanning (section 8)
- Mixed archived+active tasks NOT triggering replanning (section 8)
- Recurring execution with all-archived tasks completing (section 9)

One observation: there is no test for calling `resetGoal` on a `completed` or `archived` goal. Given the P2 note above about missing state guards, adding a test that either asserts the guard rejects the call, or explicitly documents the "any-state allowed" behaviour with a comment, would close that gap.

---

### Summary

The two `isTerminal` bug fixes are correct, well-targeted, and the test coverage directly exercises the fixed behaviour. The `resetGoal` method is a clean addition. The main gaps are (a) no RPC/MCP exposure yet per the original plan, and (b) no input validation guarding against resetting already-completed/archived goals.

PR #1165, opened 2026-04-01
feat: add useViewportSafety hook and CSS custom properties

lsm:
Implements Task 1 of the iPad Safari layout fix plan.

- `useViewportSafety` hook detects iPad Safari via `maxTouchPoints > 1` + Safari UA string (excludes Chrome/CriOS/FxiOS), then sets `--safe-height` from `window.visualViewport.height`. No-op on all other browsers.
- `.h-safe-screen` utility: `height: var(--safe-height, 100svh)` — JS sets the property on iPad Safari; other browsers use the `100svh` fallback.
- `.pb-bottom-bar` utility: `padding-bottom: var(--bottom-bar-height, 0px)` — Tailwind-consistent replacement for hardcoded `pb-16`.
- Body height changed from `100dvh` → `100svh` to eliminate toolbar-animation jank.
- 13 unit tests covering detection logic, CSS property management, and event listener cleanup.

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -1,6 +1,7 @@
 import { useEffect } from 'preact/hooks';
 import { effect, batch } from '@preact/signals';
 import { useNeoKeyboardShortcut } from './hooks/useNeoKeyboardShortcut.ts';
+import { useViewportSafety } from './hooks/useViewportSafety.ts';
 import { NeoPanel } from './components/neo/NeoPanel.tsx';
 import { NavRail } from './islands/NavRail.tsx';
 import { BottomTabBar } from './islands/BottomTabBar.tsx';
@@ -50,6 +51,8 @@ import {
 export function App() {
 	// Global Cmd+J / Ctrl+J shortcut to toggle the Neo panel
 	useNeoKeyboardShortcut();
+	// Set --safe-height CSS custom property on iPad Safari for correct viewport sizing
+	useViewportSafety();
 
 	useEffect(() => {
 		// STEP 1: Initialize URL-based router BEFORE any state management
--- a/packages/web/src/hooks/__tests__/useViewportSafety.test.ts
+++ b/packages/web/src/hooks/__tests__/useViewportSafety.test.ts
@@ -1,0 +1,253 @@
+// @ts-nocheck
+/**
+ * Tests for useViewportSafety Hook
+ *
+ * Tests iPad Safari detection logic, --safe-height CSS property management,
+ * and event listener cleanup.
+ */
+
+import { renderHook } from '@testing-library/preact';
+import { useViewportSafety } from '../useViewportSafety.ts';
+
+// Store original navigator/window properties so we can restore them.
+const originalNavigator = {
+	maxTouchPoints: navigator.maxTouchPoints,
+	userAgent: navigator.userAgent,
+};
+
+const originalVisualViewport = window.visualViewport;
+
+function setNavigator(maxTouchPoints: number, userAgent: string) {
+	Object.defineProperty(navigator, 'maxTouchPoints', {
+		configurable: true,
+		get: () => maxTouchPoints,
+	});
+	Object.defineProperty(navigator, 'userAgent', {
+		configurable: true,
+		get: () => userAgent,
+	});
+}
+
+function restoreNavigator() {
+	Object.defineProperty(navigator, 'maxTouchPoints', {
+		configurable: true,
+		get: () => originalNavigator.maxTouchPoints,
+	});
+	Object.defineProperty(navigator, 'userAgent', {
+		configurable: true,
+		get: () => originalNavigator.userAgent,
+	});
+}
+
+function setVisualViewport(vv: VisualViewport | null) {
+	Object.defineProperty(window, 'visualViewport', {
+		configurable: true,
+		get: () => vv,
+	});
+}
+
+function restoreVisualViewport() {
+	Object.defineProperty(window, 'visualViewport', {
+		configurable: true,
+		get: () => originalVisualViewport,
+	});
+}
+
+function createMockVisualViewport(height: number) {
+	const listeners: Record<string, Array<() => void>> = {};
+	const vv = {
+		height,
+		addEventListener: vi.fn((event: string, cb: () => void) => {
+			listeners[event] = listeners[event] ?? [];
+			listeners[event].push(cb);
+		}),
+		removeEventListener: vi.fn((event: string, cb: () => void) => {
+			if (listeners[event]) {
+				listeners[event] = listeners[event].filter((l) => l !== cb);
+			}
+		}),
+		_trigger: (event: string) => {
+			(listeners[event] ?? []).forEach((cb) => cb());
+		},
+	};
+	return vv;
+}
+
+// iPad Safari UA string (iPadOS 16 masquerades as macOS Safari)
+const IPAD_SAFARI_UA =
+	'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.0 Safari/605.1.15';
+
+// Desktop macOS Safari UA
+const DESKTOP_SAFARI_UA =
+	'Mozilla/5.0 (Macintosh; Intel Mac OS X 13_5) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.5 Safari/605.1.15';
+
+// Desktop Chrome UA
+const DESKTOP_CHROME_UA =
+	'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36';
+
+// Chrome on iOS (CriOS)
+const CRIOS_UA =
+	'Mozilla/5.0 (iPhone; CPU iPhone OS 16_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/120.0.0.0 Mobile/15E148 Safari/604.1';
+
+// Firefox on iOS (FxiOS)
+const FXIOS_UA =
+	'Mozilla/5.0 (iPhone; CPU iPhone OS 16_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/120.0 Mobile/15E148 Safari/605.1.15';
+
+afterEach(() => {
+	restoreNavigator();
+	restoreVisualViewport();
+	// Clear any --safe-height set during tests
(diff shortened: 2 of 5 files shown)


lsm (commented):

Overall the hook implementation is clean, well-documented, and the test coverage is solid (13 tests passing). However, there is one CSS bug and one integration gap worth addressing.

### P0: CSS height fallback ordering is inverted (styles.css lines 84-86)

```css
height: 100svh;
/* Fallback for browsers that don't support svh */
height: 100vh;
```

In CSS, when the same property appears twice in one rule, the **last** declaration wins. Since all modern browsers support both `vh` and `svh`, `100vh` always overrides `100svh`. The `svh` declaration on line 84 never takes effect. This defeats the stated goal of switching from `dvh` to `svh` to avoid layout jank.

The correct fallback pattern is:
```css
height: 100vh;    /* fallback for old browsers */
height: 100svh;   /* preferred — overrides if supported */
```

The same inverted pattern existed before this PR (`100dvh` then `100vh`), so this is a pre-existing bug that was carried forward rather than fixed, but the PR explicitly changes this code and the plan document specifically calls for "100svh (with 100vh fallback)", so the ordering should be corrected here.

### P1: Hook is not called anywhere in the application

The hook is exported from `hooks/index.ts` but is never actually called — not in `App.tsx` or any other component. The JSDoc explicitly says "This hook must only be called once globally in App.tsx." Without this call, `--safe-height` will never be set on iPad Safari, and the entire hook is dead code.

The plan document (Task 1) says to "Export the hook for use in App.tsx", implying the integration into `App.tsx` happens in a later task. If that is the intent, this is acceptable as-is, but it should be noted that the feature is non-functional until that integration happens.

### P1: `.h-safe-screen` and `.pb-bottom-bar` CSS classes are unused

Neither utility class is referenced anywhere in the codebase. Without consumers, these are dead CSS that knip or a future audit might flag. If integration is planned for a later task, this is fine, but worth tracking.

### Minor observations (P3)

- The `@ts-nocheck` at the top of the test file is a blunt instrument. It would be better to use targeted `@ts-expect-error` comments on the specific lines that need type suppression (e.g., mock viewport casts), but this is a common pattern for test files and not blocking.
- The `isIpadSafari()` detection function is not exported, making it untestable in isolation. The tests work around this by testing detection through the hook's side effects, which is acceptable but slightly indirect.

**Recommendation: REQUEST_CHANGES** — the CSS fallback ordering bug (P0) should be fixed before merging.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — The hook is implemented and tested in isolation, but it is never wired into the running application. The CSS utilities are created but unused. The PR as merged has zero functional effect on iPad Safari layout.

---

### P0 — Hook is never called

`useViewportSafety` is exported from `packages/web/src/hooks/index.ts` but is never imported or called in `App.tsx` or any other component. The hook's own JSDoc says it **must only be called once globally in `App.tsx`**, but that call does not exist. On an iPad Safari device, `--safe-height` is never set, so `.h-safe-screen` falls through to its `100svh` fallback — identical to the pre-PR behavior.

**Fix:** Add `useViewportSafety()` as a top-level call inside the `App` function in `packages/web/src/App.tsx`.

---

### P0 — Root container still uses `h-dvh`, not `h-safe-screen`

`App.tsx` line 175:
```tsx
<div class="flex h-dvh overflow-hidden bg-dark-950 relative pt-safe">
```
The body comment in `styles.css` says "root container height is managed via `.h-safe-screen` on iPad Safari", but the actual root `<div>` still uses `h-dvh`. The `.h-safe-screen` utility class is defined but applied nowhere. Until this class replaces `h-dvh` on the root container, iPad Safari will continue using the dynamic viewport height that causes toolbar-animation jank.

---

### P0 — `pb-16` hardcoded, `pb-bottom-bar` never applied

`App.tsx` line 183:
```tsx
<div class="flex-1 flex flex-col overflow-hidden pb-16 md:pb-0 min-w-0">
```
The PR description says `.pb-bottom-bar` is a "Tailwind-consistent replacement for hardcoded `pb-16`", but the replacement never happens. Additionally, nothing in the codebase sets `--bottom-bar-height` to a non-zero value (grep confirms no component writes to this property). The `pb-bottom-bar` utility is therefore inert.

---

### P1 — `svh` / `vh` fallback ordering in `body` CSS is reversed

`styles.css` lines 84–86:
```css
height: 100svh;
/* Fallback for browsers that don't support svh */
height: 100vh;
```
The comment says `100vh` is the fallback, but in CSS cascade order the _last_ declaration wins when both are valid. Browsers that support `svh` will evaluate both lines and apply `100vh` (the last one), defeating the intent. The fallback must come _before_ the preferred value. The previous code had `100dvh` then `100vh` — this was correct. The new ordering is incorrect for every browser that supports `svh` (all modern browsers).

**Fix:** Swap the order:
```css
height: 100vh; /* fallback for older browsers */
height: 100svh;
```

---

### P1 — `@ts-nocheck` in test file suppresses all type checking

`packages/web/src/hooks/__tests__/useViewportSafety.test.ts` line 1:
```ts
// @ts-nocheck
```
This disables TypeScript across the entire test file. The project's `no-explicit-any` rule is set to error; `@ts-nocheck` is a broad way to sidestep that. The mock helpers use `vi.fn()` and `unknown as VisualViewport` casts that could be typed properly. The actual typing issues here are solvable without suppressing the entire file.

---

### P2 — `--bottom-bar-height` setter is absent

`--bottom-bar-height` is declared at `0px` in `:root` with a comment saying it is "overridden by JS when BottomTabBar is visible". No code in this PR or the existing codebase writes to this property. The `BottomTabBar` component needs to measure its own height and call `document.documentElement.style.setProperty('--bottom-bar-height', ...)` for the `.pb-bottom-bar` utility to work. This setter belongs in this PR or it should be explicitly deferred to a follow-up task with a TODO comment.

---

### P3 — Missing `orientationchange` coverage comment

The hook listens to both `visualViewport.resize` and `window.resize`. On iPadOS, device rotation triggers `orientationchange` in addition to these resize events. In practice `resize` fires on rotation so this is likely sufficient, but a brief comment noting that `orientationchange` is handled implicitly via `resize` would help future maintainers.

lsm (commented):

1. **CSS fallback ordering** -- `100vh` fallback now correctly precedes `100svh`, so the latter wins in supporting browsers. The switch from `dvh` to `svh` is well-reasoned (avoids toolbar animation jank).

2. **Hook invocation** -- `useViewportSafety()` is now called in `App.tsx` at the component top level.

3. **`@ts-nocheck` removed** -- Tests use a proper `MockVisualViewport` interface with typed fields.

4. **`orientationchange` comment** -- Line 83-85 explains why a separate listener is unnecessary.

The implementation is clean and well-documented:

- `isIpadSafari()` detection is solid, covering the key exclusions (Chrome, CriOS, FxiOS) and using `maxTouchPoints > 1` instead of deprecated `navigator.platform`.
- The hook correctly sets up and tears down both `visualViewport.resize` and `window.resize` listeners.
- CSS utilities (`.h-safe-screen`, `.pb-bottom-bar`) are defined and ready for use in subsequent tasks. The root container still uses `h-dvh` -- confirmed deferred per plan.
- Test coverage is thorough: 5 UA detection variants, property set/unset, listener lifecycle, and resize event handling.

No issues found. LGTM.

**Recommendation: APPROVE**

lsm (commented):

Recommendation: **APPROVE** (with notes)

The previous-round issues have been properly addressed. The implementation is sound and the deferred items are clearly scoped. A few observations below.

---

### What was fixed correctly

- CSS fallback ordering is now correct: `height: 100vh` before `height: 100svh` in `html, body`, so modern browsers get svh and older browsers fall back to vh.
- `useViewportSafety()` is called in `App.tsx` at line 55 — the hook now actually runs at runtime.
- `@ts-nocheck` removed from the test file; replaced by a proper `MockVisualViewport` interface at lines 16–22.
- The `orientationchange` omission is explained in a comment at lines 83–85 of the hook.

---

### P2 — `.h-safe-screen` is defined but never applied

`packages/web/src/styles.css` lines 580–582 define the `.h-safe-screen` utility and `packages/web/src/hooks/useViewportSafety.ts` sets `--safe-height` on `document.documentElement` — but no component in the codebase uses `.h-safe-screen`. The root container in `App.tsx:178` still uses `h-dvh`, which is exactly the class that causes toolbar-animation jank on iPad Safari according to the PR description. The hook sets a CSS variable that nothing reads.

The PR description states this replacement is intentionally deferred. That is fine as a delivery strategy, but it means the iPad Safari layout fix described in this PR does not actually take effect at runtime yet. The CSS + hook are foundations — but without the consumer (`h-dvh → h-safe-screen` swap), the observable behavior is unchanged on iPad Safari. This should be clearly tracked and landed promptly in the next task, since the current state ships dead infrastructure.

---

### P2 — `--bottom-bar-height` has no setter

`packages/web/src/styles.css` line 60 declares `--bottom-bar-height: 0px` in `:root`, and line 588 defines `.pb-bottom-bar` which consumes it. The comment at line 59 says it is "overridden by JS when BottomTabBar is visible", but no code in `packages/web/src/` calls `setProperty('--bottom-bar-height', ...)`. The `.pb-bottom-bar` class is also not used anywhere. This is another deferred piece, but worth flagging: until the setter and consumer are wired up, `.pb-bottom-bar` always resolves to `padding-bottom: 0px`, which silently differs from the `pb-16` it is meant to replace.

---

### P3 — `.h-safe-screen` missing a `100vh` double-declaration

`packages/web/src/styles.css:581` uses `height: var(--safe-height, 100svh)`. The CSS `var()` fallback only applies when the custom property is not set — it does not provide a fallback for browsers that do not support `svh` units. If a browser supports CSS custom properties but not `svh`, the rule evaluates to an invalid unit and the height is unset. The `html, body` rule uses a double-declaration pattern for exactly this reason. `.h-safe-screen` could follow the same pattern (declare `height: 100vh` first, then `height: var(--safe-height, 100svh)`). This is a minor robustness gap — target browsers (iPad Safari 15+) support `svh`, so it is not a blocking concern.

---

### P3 — Test `restoreVisualViewport` hardcodes `null`

`packages/web/src/hooks/__tests__/useViewportSafety.test.ts:77–81` restores `window.visualViewport` to `null` regardless of its original value. In the jsdom environment this is fine since `visualViewport` is not set by default, but if the test environment ever starts providing a real `VisualViewport` instance, tests that check "does nothing when visualViewport is unavailable" would interfere with subsequent tests. Not a current bug, just worth noting.

---

### Summary

The infrastructure for the iPad Safari viewport fix is correctly implemented: the detection heuristic is sound, the hook lifecycle is correct, event listener cleanup is proper, and the tests are well-structured. The deferred steps (applying `h-safe-screen` to the root container and wiring `--bottom-bar-height`) are the remaining work to make the fix observable on device. Landing those in the next task is the right call.

PR #1166, opened 2026-04-01
feat: add reset_goal MCP tool and expose planning_attempts in update_goal

lsm:
Implements `reset_goal` MCP tool and exposes `planning_attempts` as a writable field in `update_goal`.

**Changes:**
- `reset_goal` tool: cancels all non-terminal linked tasks (via `runtime.cancelTask` or fallback to `taskManager.cancelTaskCascade` + hub event), calls `goalManager.resetGoal()`, then triggers `runtime.onGoalCreated()` for fresh planning
- Registered only in `createRoomAgentMcpServer` — not in `createLeaderContextMcpServer`
- `update_goal` now accepts `planning_attempts?: number` to manually reset the counter
- Unit tests in `room-agent-tools-goal.test.ts`: reset success/error cases, runtime path, fallback path, planning_attempts persistence, negative test confirming `reset_goal` absent from leader MCP server

### DIFF
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -1023,20 +1023,36 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				'usage_limited',
 			]);
 
+			// Hoist the runtime lookup so it isn't repeated on every loop iteration
+			const runtime = runtimeService?.getRuntime(roomId) ?? null;
+
 			// Cancel all in-progress linked tasks
 			for (const taskId of goal.linkedTaskIds) {
 				const task = await taskManager.getTask(taskId);
 				if (!task || !nonTerminalStatuses.has(task.status)) continue;
 
-				if (runtimeService) {
-					const runtime = runtimeService.getRuntime(roomId);
-					if (runtime) {
-						await runtime.cancelTask(taskId);
-						continue;
+				if (runtime) {
+					await runtime.cancelTask(taskId);
+					continue;
+				}
+
+				// Fallback path: no runtime available.
+				// Guard: if the task has an active session group we cannot safely cancel it
+				// without the runtime — the DB row would be updated but the agent sessions
+				// would keep running (same guard as `cancel_task`).
+				if (task.status === 'in_progress' || task.status === 'review') {
+					const activeGroup = groupRepo.getGroupByTaskId(taskId);
+					if (activeGroup && activeGroup.completedAt === null) {
+						return jsonResult({
+							success: false,
+							error:
+								`Cannot reset goal ${args.goal_id}: task ${taskId} (status '${task.status}') has an active session group but no runtime service is available to stop it. ` +
+								`The active worker session would not be stopped.`,
+						});
 					}
 				}
 
-				// Fallback: cancel via taskManager and emit update event
+				// No active session group — safe to cancel via taskManager
 				await taskManager.cancelTaskCascade(taskId);
 				if (daemonHub) {
 					const cancelledTask = await taskManager.getTask(taskId);
@@ -1054,11 +1070,8 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 			const resetGoal = await goalManager.resetGoal(args.goal_id);
 
 			// Trigger a fresh planning tick if runtime is available
-			if (runtimeService) {
-				const runtime = runtimeService.getRuntime(roomId);
-				if (runtime) {
-					runtime.onGoalCreated(args.goal_id);
-				}
+			if (runtime) {
+				runtime.onGoalCreated(args.goal_id);
 			}
 
 			return jsonResult({ success: true, goal: resetGoal });
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -5,7 +5,7 @@
  * the human user. The Room Agent session already exists (room:chat:${roomId}),
  * these tools are attached to it.
  *
- * Tools: create_goal, list_goals, update_goal, create_task, list_tasks,
+ * Tools: create_goal, list_goals, update_goal, reset_goal, create_task, list_tasks,
  *        update_task, cancel_task, stop_session, get_room_status, approve_task, reject_task,
  *        send_message_to_task, get_task_detail, set_schedule, pause_schedule, resume_schedule,
  *        record_metric, get_metrics, list_executions
@@ -140,6 +140,7 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 			priority?: 'low' | 'normal' | 'high' | 'urgent';
 			mission_type?: MissionType;
 			autonomy_level?: AutonomyLevel;
+			planning_attempts?: number;
 			structured_metrics?: Array<{
 				name: string;
 				target: number;
@@ -161,6 +162,7 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				args.priority === undefined &&
 				args.mission_type === undefined &&
 				args.autonomy_level === undefined &&
+				args.planning_attempts === undefined &&
 				args.structured_metrics === undefined
 			) {
 				return jsonResult({ success: false, error: 'No update fields provided.' });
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

The implementation is clean, correct, and well-tested. Here is a summary of findings:

### What was reviewed

1. **`reset_goal` handler** in `createRoomAgentToolHandlers()` — cancels non-terminal linked tasks (via `runtime.cancelTask` or `taskManager.cancelTaskCascade` fallback), calls `goalManager.resetGoal()`, and triggers `runtime.onGoalCreated()` for fresh planning.
2. **`reset_goal` MCP registration** — correctly registered in `createRoomAgentMcpServer()` but intentionally excluded from `createLeaderContextMcpServer()` (which lacks `runtimeService` in its config type).
3. **`update_goal` planning_attempts** — added as an optional field in the handler args, the "no update fields" guard, the `hasPatchFields` check, and the patch assembly. The Zod schema validates `z.number().int().min(0)`.
4. **Tool count assertion** updated from 20 to 21.
5. **13 unit tests** covering reset_goal (not found, linked tasks with mixed statuses, task cancellation, runtime path, fallback/hub path) and update_goal planning_attempts (persist, non-zero, no-fields guard, planning_attempts-only).

### Correctness

- The `reset_goal` handler correctly checks for goal existence before proceeding, iterates `goal.linkedTaskIds` and filters by non-terminal status using an appropriate set of statuses (`pending`, `in_progress`, `draft`, `review`, `rate_limited`, `usage_limited`).
- The runtime vs fallback branching is sound: when `runtimeService` provides a runtime, it delegates to `runtime.cancelTask()`; otherwise falls back to `taskManager.cancelTaskCascade()` with hub event emission.
- `goalManager.resetGoal()` clears `linkedTaskIds`, resets `planning_attempts`, `consecutiveFailures`, `replanCount` to 0, and sets status to `active` — confirmed in goal-manager.ts.
- `planning_attempts` in `update_goal` is correctly threaded through: args type, empty-check guard, `hasPatchFields` condition, and `patch.planning_attempts` assignment. The repo's `UpdateGoalParams` type already includes `planning_attempts` so this works end-to-end.

### No issues found

- Error handling is consistent with existing handlers (no try/catch — the MCP framework catches unhandled exceptions at the tool boundary).
- The `nonTerminalStatuses` set is defined inline rather than imported, which is a minor duplication but acceptable for a tool handler.
- Tests are thorough and cover the key paths including runtime integration and hub fallback.

No blocking or significant issues identified.

lsm (commented):

The implementation is structurally sound and the test coverage is good, but there are two issues worth addressing before merging: a missing safety guard that the existing `cancel_task` handler deliberately has, and a minor inefficiency in the runtime lookup inside the task-cancellation loop.

---

### P1 — Missing session-group safety guard in `reset_goal`

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 1027–1051

The existing `cancel_task` handler (lines 380–392) has a deliberate guard:

```ts
if ((task.status === 'in_progress' || task.status === 'review') && !runtimeService) {
    const activeGroup = groupRepo.getGroupByTaskId(args.task_id);
    if (activeGroup && activeGroup.completedAt === null) {
        return jsonResult({
            success: false,
            error: `Cannot cancel task ... without runtime service — the active worker session would not be stopped.`,
        });
    }
}
```

The comment explains why: "the DB would be updated but the worker/leader sessions would continue running against a cancelled task."

`reset_goal` calls `taskManager.cancelTaskCascade()` as its fallback path when `runtimeService` is absent, which has the same problem. If a linked task is `in_progress` or `review` with an active uncompleted session group, the task DB row gets marked cancelled but the running agent sessions are not stopped — leaving orphaned sessions consuming concurrency slots and potentially writing back to a now-cancelled task.

The fix is to either (a) apply the same guard per-task before falling back to `cancelTaskCascade`, or (b) return an error if any linked task is in this state and `runtimeService` is unavailable.

---

### P2 — `runtimeService.getRuntime()` called on every loop iteration

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 1031–1036

Inside the `for` loop over `goal.linkedTaskIds`, `runtimeService.getRuntime(roomId)` is called on every iteration. Since `roomId` is a constant within the handler closure and the runtime does not change mid-loop, this lookup should be hoisted above the loop. All other uses of `runtimeService.getRuntime` in the file follow the pattern of fetching once before the work begins.

---

### P0/P1 — No issues found

No security vulnerabilities, data-loss paths, or broken behavior detected outside the session-group guard gap.

### What's correct and well-done

- The `nonTerminalStatuses` set correctly excludes `needs_attention` and `archived` (consistent with `room-runtime.ts:3336-3340` and `session-group-repository.ts:367`).
- `reset_goal` is registered in `createRoomAgentMcpServer` but not `createLeaderContextMcpServer`, which is consistent with the documented exclusion of goal-mutation tools from the leader context.
- `planning_attempts` is wired end-to-end: the Zod schema adds `.int().min(0)` validation, the handler passes it through `patchGoal`, and `UpdateGoalParams` already supports it.
- The `onGoalCreated` call after `resetGoal` correctly triggers a fresh planning tick without requiring a separate event emission.
- The 13 unit tests cover the important branches: not-found, terminal/non-terminal task filtering, runtime path, fallback path, hub event emission, and the negative assertion that `reset_goal` is absent from the leader MCP server.
- Tool count updated from 20 → 21.

lsm (commented):

**Recommendation: APPROVE**

Both requested fixes from the previous review are correctly implemented:

1. **P2 fix (runtime hoisted):** `runtimeService.getRuntime(roomId)` is now called once at line 1027, before the loop over `goal.linkedTaskIds`. Correct.

2. **P1 fix (session-group safety guard):** Lines 1043-1053 add a guard in the fallback path that matches the pattern from `cancel_task` — when a task is `in_progress` or `review` with an active (non-completed) session group and no runtime is available, the handler returns an error. Two new unit tests at lines 387-431 cover both the blocking case (active session group present) and the success case (no session group).

**Implementation quality:**

- The `reset_goal` handler is well-structured: fetch goal, cancel non-terminal linked tasks, call `resetGoal()`, trigger fresh planning.
- The `planning_attempts` field in `update_goal` has proper Zod validation (`.number().int().min(0)`) and integrates cleanly into the existing patch flow.
- Test coverage is thorough: 7 tests for `reset_goal` (not-found, non-terminal cancellation, terminal skip, runtime path, fallback path with hub events, session-group guard, no-session-group fallback) plus 4 for `planning_attempts` updates.
- Negative test confirms `reset_goal` is absent from the leader MCP server.
- CI checks are passing.

**One observation (P3, not blocking):** The safety guard in `reset_goal` (line 1027) checks the actual `runtime` object, while `cancel_task` (line 384) checks `!runtimeService`. This means `reset_goal` correctly guards against the edge case where `runtimeService` exists but `getRuntime(roomId)` returns null, while `cancel_task` does not. This is actually a minor improvement in `reset_goal` over the pattern it was modeled after, but the inconsistency is worth noting for future cleanup of `cancel_task`.

No blocking issues found.

lsm (commented):

**Recommendation: APPROVE**

Both previously raised issues have been addressed correctly.

---

### P2 Fix: `runtimeService.getRuntime()` hoisted above loop

The runtime lookup in `reset_goal` is now correctly hoisted above the task-cancellation loop (line 1027 in `room-agent-tools.ts`), called once as `const runtime = runtimeService?.getRuntime(roomId) ?? null`. This is consistent with every other handler in the file and eliminates the redundant per-iteration call. The hoisted value is correctly threaded through to the post-reset `onGoalCreated` call as well.

**Observation (P3 — no action needed):** The hoisted `runtime` uses `runtimeService?.getRuntime(roomId) ?? null`, which captures two distinct cases: (a) no `runtimeService` at all, and (b) `runtimeService` present but returning null for this room. The `cancel_task` guard only checks `!runtimeService`. The `reset_goal` guard checks `!runtime` — slightly broader but more correct. This is fine.

---

### P1 Fix: Session-group safety guard in fallback path

The guard at lines 1043–1053 matches the intent of the `cancel_task` guard: if a task is `in_progress` or `review` and has an active (non-completed) session group, and no runtime is available, the handler returns an error instead of leaving orphaned sessions. The error message is clear and actionable.

Two new unit tests were added:
- `should return error when in_progress task has active session group and no runtime` — confirms the guard fires, the task remains `in_progress`, and `success: false` is returned with the expected error message.
- `should cancel in_progress task without active session group in fallback path` — confirms that an `in_progress` task without a session group is still safely cancelled in fallback mode.

Both tests use `insertActiveGroup()`, a correctly scoped helper that inserts a realistic `session_groups` row with `completedAt = null` (no completed_at column in the INSERT, which defaults to null).

---

### Remaining observations (P3, no action required)

1. **No test for `review` status with active group in fallback.** The guard covers both `in_progress` and `review` but only `in_progress` is tested. The symmetry is obvious enough that this is low risk.

2. **No test for `runtimeService` present but `getRuntime` returning null with active session group.** The guard at `!runtime` correctly fires in this scenario, but it is untested. This is an edge case.

3. **Partial-cancel on multi-task goals.** If a goal has tasks `[t_pending, t_in_progress_with_group]`, the loop cancels `t_pending` first and then hits the error on `t_in_progress_with_group`, returning an error with `t_pending` already cancelled. This is a pre-existing architectural limitation of the loop pattern (not introduced by this PR) and is acceptable given the rarity of the scenario. No test covers it, but this also predates this PR.

---

All P0, P1, and P2 issues from the previous review are resolved. The implementation is correct and consistent with the existing guard patterns in the codebase.

PR #1167, opened 2026-04-01
feat: dynamic bottom padding via --bottom-bar-height CSS custom property

lsm:
Replace hardcoded `pb-16 md:pb-0` on the main content div with a `pb-bottom-bar` utility class driven by the `--bottom-bar-height` CSS custom property.

- `BottomTabBar` gains a `ResizeObserver` on its root `div` to measure actual height and write `--bottom-bar-height` to `document.documentElement`
- A supplementary `window.resize` listener (with `requestAnimationFrame`) handles breakpoint transitions where `md:hidden` flips `display:none` — a case `ResizeObserver` doesn't fire for
- On unmount, resets `--bottom-bar-height` to `0px` and cleans up both listeners
- `App.tsx` uses `pb-bottom-bar` instead of `pb-16 md:pb-0`; the dynamic approach handles desktop (0px) and mobile (actual height) automatically
- 6 new source-level assertions added to `ios-safe-area.test.ts`

### DIFF
--- a/packages/web/src/App.tsx
+++ b/packages/web/src/App.tsx
@@ -182,8 +182,8 @@ export function App() {
 				{/* Context Panel - always visible */}
 				<ContextPanel />
 
-				{/* Main Content — add bottom padding on mobile to avoid tab bar overlap */}
-				<div class="flex-1 flex flex-col overflow-hidden pb-16 md:pb-0 min-w-0">
+				{/* Main Content — bottom padding matches actual BottomTabBar height via --bottom-bar-height */}
+				<div class="flex-1 flex flex-col overflow-hidden pb-bottom-bar min-w-0">
 					<MainContent />
 				</div>
 			</div>
--- a/packages/web/src/islands/BottomTabBar.tsx
+++ b/packages/web/src/islands/BottomTabBar.tsx
@@ -1,4 +1,5 @@
 import type { JSX } from 'preact';
+import { useEffect, useRef } from 'preact/hooks';
 import {
 	navSectionSignal,
 	currentRoomIdSignal,
@@ -124,6 +125,39 @@ const ROOM_BOTTOM_TABS: TabItem[] = [
 ];
 
 export function BottomTabBar() {
+	const rootRef = useRef<HTMLDivElement>(null);
+
+	useEffect(() => {
+		const el = rootRef.current;
+		if (!el) return;
+
+		const updateHeight = () => {
+			const h = el.offsetHeight;
+			document.documentElement.style.setProperty('--bottom-bar-height', h + 'px');
+		};
+
+		// ResizeObserver fires when the element itself resizes
+		const ro = new ResizeObserver(updateHeight);
+		ro.observe(el);
+
+		// window resize listener handles breakpoint transitions (md:hidden causes
+		// display:none, which ResizeObserver does not fire for).
+		// requestAnimationFrame ensures the browser applies the new display value first.
+		const onResize = () => {
+			requestAnimationFrame(updateHeight);
+		};
+		window.addEventListener('resize', onResize);
+
+		// Initial measurement
+		updateHeight();
+
+		return () => {
+			ro.disconnect();
+			window.removeEventListener('resize', onResize);
+			document.documentElement.style.setProperty('--bottom-bar-height', '0px');
+		};
+	}, []);
+
 	const navSection = navSectionSignal.value;
 	const roomId = currentRoomIdSignal.value;
 	const roomSessionId = currentRoomSessionIdSignal.value;
@@ -186,6 +220,7 @@ export function BottomTabBar() {
 
 	return (
 		<div
+			ref={rootRef}
 			class="flex md:hidden fixed bottom-0 left-0 right-0 z-50 bg-dark-900/90 backdrop-blur-md border-t border-dark-700 pb-safe"
 			role="tablist"
 			aria-label="Main navigation"
--- a/packages/web/src/islands/BottomTabBar.tsx
+++ b/packages/web/src/islands/BottomTabBar.tsx
@@ -143,8 +143,11 @@ export function BottomTabBar() {
 		// window resize listener handles breakpoint transitions (md:hidden causes
 		// display:none, which ResizeObserver does not fire for).
 		// requestAnimationFrame ensures the browser applies the new display value first.
+		// The cancelled-rAF guard prevents queuing multiple callbacks during rapid resize.
+		let rafId = 0;
 		const onResize = () => {
-			requestAnimationFrame(updateHeight);
+			cancelAnimationFrame(rafId);
+			rafId = requestAnimationFrame(updateHeight);
 		};
 		window.addEventListener('resize', onResize);
 
--- a/packages/web/src/islands/BottomTabBar.tsx
+++ b/packages/web/src/islands/BottomTabBar.tsx
@@ -157,6 +157,7 @@ export function BottomTabBar() {
 		return () => {
 			ro.disconnect();
 			window.removeEventListener('resize', onResize);
+			cancelAnimationFrame(rafId);
 			document.documentElement.style.setProperty('--bottom-bar-height', '0px');
 		};
 	}, []);
--- a/packages/web/src/lib/__tests__/ios-safe-area.test.ts
+++ b/packages/web/src/lib/__tests__/ios-safe-area.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest';
 import indexHtml from '../../index.html?raw';
 import appTsx from '../../App.tsx?raw';
 import useViewportSafetyTs from '../../hooks/useViewportSafety.ts?raw';
+import bottomTabBarTsx from '../../islands/BottomTabBar.tsx?raw';
 
 /**
  * iOS iPad Safari safe area support tests.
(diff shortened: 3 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

The implementation correctly replaces the hardcoded `pb-16 md:pb-0` with a dynamic measurement approach. The architecture is sound:

1. **BottomTabBar.tsx** — Uses `ResizeObserver` for element resizes plus a `window.resize` listener (with `requestAnimationFrame`) to catch `md:hidden` breakpoint transitions where `display:none` makes `offsetHeight` return `0`. This correctly handles the edge case that `ResizeObserver` does not fire for visibility changes caused by CSS media queries. Cleanup on unmount resets the property to `0px` and removes all listeners.

2. **CSS plumbing** — `--bottom-bar-height` is initialized to `0px` on `:root` in `styles.css`, and the `.pb-bottom-bar` utility class references it with a `0px` fallback. This means on desktop (where the bar is hidden), padding is always 0 without any flash.

3. **App.tsx** — Clean replacement of `pb-16 md:pb-0` with the single `pb-bottom-bar` class. The `md:pb-0` override is no longer needed because the JS measurement naturally returns 0 when the bar is hidden.

**P3 (nit): Source-text tests are brittle but acceptable here.**

The new tests in `ios-safe-area.test.ts` (lines 25-50) assert against raw source text (e.g., `expect(bottomTabBarTsx).toContain('ResizeObserver')`). These will break on refactors like renaming variables or extracting the measurement logic into a custom hook. However, this follows the pre-existing pattern in the file (which tests for `pt-safe`, `viewport-fit=cover` the same way), and the comment at line 12-14 explains why this approach is used (Vite strips CSS in tests). Consistent with the existing convention, so this is fine.

No correctness issues, no missing edge cases, no security concerns. The implementation fully achieves the stated goal.

lsm (commented):

**Recommendation: APPROVE**

The implementation is clean and correct. The three-file change achieves the stated goal: replacing the hardcoded `pb-16 md:pb-0` with a CSS custom property driven by actual DOM measurement.

### What works well

**`BottomTabBar.tsx`**

- `ResizeObserver` on the root `div` correctly captures height changes caused by content reflow (font-scale, dynamic safe-area, etc.).
- The supplementary `window.resize` listener with `requestAnimationFrame` is the right fix for the `md:hidden` / `display:none` edge case. `ResizeObserver` does not fire when an element transitions to `display:none`, so the resize listener correctly re-reads `offsetHeight` (which returns `0` when hidden) after the browser has applied the breakpoint change.
- `offsetHeight` on the element that carries `pb-safe` (`env(safe-area-inset-bottom)`) means the measured height already includes the iOS safe area padding. The main content therefore gets one unified bottom padding covering both the tab chrome and the safe-area notch — no double-padding risk.
- Unmount cleanup resets the property to `0px`, which is the correct default (matches the `--bottom-bar-height: 0px` on `:root` in `styles.css`).
- The dependency array `[]` is correct; the effect references only stable DOM APIs and the ref.

**`App.tsx`**

- The single class change is minimal and correct. On desktop the property stays `0px` (the hidden element reports `offsetHeight === 0`); on mobile it tracks the real bar height.

**`styles.css`** (in the prior PR, read for context)

- `--bottom-bar-height: 0px` on `:root` means the page renders with zero bottom padding during the brief window before the first `useEffect` fires — no layout flash.

**Tests**

- Source-level string assertions are the right pragmatic choice given that Vitest/JSDOM does not implement `ResizeObserver` or CSS `env()`. They guard against accidental regression in the mechanism (e.g. someone removing the ResizeObserver call), not just the class name.

### Issues found

**P2 — Unthrottled `rAF` callbacks on rapid resize**

`packages/web/src/islands/BottomTabBar.tsx` lines 146-149

```ts
const onResize = () => {
    requestAnimationFrame(updateHeight);
};
window.addEventListener('resize', onResize);
```

Rapid `window.resize` events (e.g. during live drag-resize of the browser window on desktop, or orientation-change + keyboard appearance on mobile) will queue one `rAF` callback per event. Because this path is only active on mobile (the ResizeObserver handles the real-height case), and the tab bar is hidden on desktop (`md:hidden`), the practical impact is low. Still, a simple cancelled-rAF guard would make this tighter:

```ts
let rafId = 0;
const onResize = () => {
    cancelAnimationFrame(rafId);
    rafId = requestAnimationFrame(updateHeight);
};
```

Not a blocking issue, but worth doing.

**P3 — Redundant `0px` fallback in `pb-bottom-bar`**

`packages/web/src/styles.css` line 589

```css
padding-bottom: var(--bottom-bar-height, 0px);
```

`--bottom-bar-height` is already initialized to `0px` on `:root`, so the fallback value inside `var()` is never reached. The redundancy is harmless and actually serves as defensive documentation, so leaving it is fine.

**P3 — Test count mismatch**

The PR description says "6 new source-level assertions"; the diff adds 7 new `it()` blocks (lines 25-50 of `ios-safe-area.test.ts`). This is a cosmetic discrepancy in the PR description only — the test code itself is correct.

lsm (commented):

### Recommendation: APPROVE

The implementation correctly replaces the hardcoded `pb-16 md:pb-0` with a dynamic CSS custom property `--bottom-bar-height` measured from the actual `BottomTabBar` element. The approach is well-structured:

1. **`BottomTabBar.tsx`** — Uses `ResizeObserver` for element size changes and a supplementary `window.resize` listener (with rAF) for breakpoint transitions where `md:hidden` toggles `display:none`. The cancel-previous-rAF guard (`cancelAnimationFrame(rafId)` before scheduling) correctly prevents unbounded callback queuing during rapid resize, addressing the P2 from the previous review.

2. **`styles.css`** — Defines `--bottom-bar-height: 0px` as a CSS custom property default on `:root`, and a `.pb-bottom-bar` utility class that reads it. Clean separation.

3. **`App.tsx`** — Swaps `pb-16 md:pb-0` for `pb-bottom-bar`. The `md:pb-0` case is now handled automatically since the bar reports `0px` height when hidden at the `md` breakpoint.

4. **Tests** — Source-level assertions verify the key integration points (custom property usage, ResizeObserver, cleanup). These are static string checks rather than behavioral tests, but given the DOM-heavy nature of the feature, they provide reasonable regression protection.

### P3 nit

**Missing `cancelAnimationFrame(rafId)` in cleanup** (`BottomTabBar.tsx` line 157-161): The cleanup function removes the resize listener and resets `--bottom-bar-height` to `0px`, but does not cancel any pending rAF callback. If a resize event fires just before unmount, the queued `updateHeight` could execute after cleanup and overwrite `0px`. In practice this is harmless because `BottomTabBar` is mounted at the app root and never unmounts, but adding `cancelAnimationFrame(rafId)` to the cleanup would be technically correct.

Overall this is a clean, well-scoped change. No blocking issues.

lsm (commented):

**Recommendation: APPROVE** (same author — posting as COMMENT per policy)

---

### Re-review of PR #1167 — dynamic bottom padding via `--bottom-bar-height`

The previously raised P2 issue (unbounded rAF queue on rapid resize) has been addressed. The fix is correct: `rafId` is declared in the effect closure and `cancelAnimationFrame(rafId)` is called before every `requestAnimationFrame(updateHeight)`, so at most one pending callback can exist at any time.

---

### Remaining observations

**P1 — Pending rAF not cancelled on unmount**

The cleanup function in the `useEffect` return does not call `cancelAnimationFrame(rafId)`. If the component unmounts while a rAF callback is pending (e.g., the user resizes and immediately navigates away), `updateHeight` will still fire after cleanup: `el.offsetHeight` is read on a detached element and `document.documentElement.style.setProperty` is called after the reset to `'0px'` has already run, potentially writing a stale pixel value back into the custom property.

```tsx
// packages/web/src/islands/BottomTabBar.tsx lines 157-161
return () => {
    ro.disconnect();
    window.removeEventListener('resize', onResize);
    // Missing: cancelAnimationFrame(rafId);
    document.documentElement.style.setProperty('--bottom-bar-height', '0px');
};
```

Fix: add `cancelAnimationFrame(rafId)` before the reset line.

**P3 — No test assertion for the rAF guard itself**

The test suite uses source-text inspection (`toContain`). None of the assertions check for `cancelAnimationFrame`, so a future refactor that drops the guard would not be caught. Consider adding:

```ts
expect(bottomTabBarTsx).toContain('cancelAnimationFrame');
```

to `ios-safe-area.test.ts`.

---

- The `ResizeObserver` + `window.resize` dual-listener strategy correctly handles the `md:hidden` display-none case that `ResizeObserver` misses.
- Initial `updateHeight()` call ensures the property is set synchronously before the first paint.
- `--bottom-bar-height: 0px` default in `styles.css` is a safe fallback for desktop (where `BottomTabBar` is `display:none`).
- The `pb-bottom-bar` utility with `var(--bottom-bar-height, 0px)` correctly replaces the hardcoded `pb-16 md:pb-0`.
- Source-text regression tests are a practical approach given the Vite CSS-stripping constraint in the test environment.

lsm (commented):

**Recommendation: APPROVE**

This PR correctly replaces the hardcoded `pb-16 md:pb-0` bottom padding with a dynamic CSS custom property `--bottom-bar-height` measured from the actual BottomTabBar element. The implementation is clean and handles all the edge cases well.

### What was reviewed

**BottomTabBar.tsx** — The `useEffect` hook measures the bar's `offsetHeight` via:
1. A `ResizeObserver` for content-driven size changes
2. A `window.resize` listener with rAF batching for breakpoint transitions (`md:hidden` toggling `display:none`, which ResizeObserver does not fire for)
3. Proper cleanup: disconnects observer, removes listener, cancels pending rAF, resets the custom property to `0px`

**App.tsx** — Replaced `pb-16 md:pb-0` with `pb-bottom-bar`, which resolves to `padding-bottom: var(--bottom-bar-height, 0px)`. The `md:pb-0` override is no longer needed because when BottomTabBar is hidden at the `md:` breakpoint, the resize handler correctly sets the property to `0px` (or the CSS default `0px` applies).

**styles.css** — Default `--bottom-bar-height: 0px` on `:root` and the `.pb-bottom-bar` utility class are properly defined.

**Tests** — Source-level assertions verify the key integration points: presence of `pb-bottom-bar` in App.tsx, absence of `pb-16`, `--bottom-bar-height` and `ResizeObserver` in BottomTabBar, and cleanup (`cancelAnimationFrame`, `ro.disconnect()`, reset to `0px`).

### Notes

- The cancel-previous-rAF guard (`cancelAnimationFrame(rafId)` before `requestAnimationFrame`) correctly prevents stacking callbacks during rapid resize events.
- The cleanup `cancelAnimationFrame(rafId)` prevents a pending rAF from writing a stale height after unmount — this was the P1 from the previous review iteration and is properly addressed.
- No remaining uses of `pb-16` in the web source outside of the test assertion itself.

No issues found. Clean, well-structured change.

lsm (commented):

**Recommendation: APPROVE**

This is a clean, well-reasoned implementation. All three previously-raised issues (P2 cancel-previous-rAF guard, P1 cancelAnimationFrame in cleanup, P3 test assertion) have been correctly addressed. The code is in its final, correct state.

### What was verified

**BottomTabBar.tsx — useEffect logic**

- `ResizeObserver` correctly measures actual height on element resize.
- `window.resize` + `requestAnimationFrame` correctly handles breakpoint transitions where `md:hidden` (display:none) makes ResizeObserver silent. When the element becomes hidden, `offsetHeight` returns `0`, so `updateHeight()` correctly writes `0px` after the rAF yields to the browser paint.
- The cancel-previous-rAF guard (`cancelAnimationFrame(rafId)` before each new rAF) prevents stacking callbacks during rapid resize bursts.
- Cleanup function correctly calls `ro.disconnect()`, `window.removeEventListener`, `cancelAnimationFrame(rafId)`, and resets `--bottom-bar-height` to `0px` — no stale rAF can fire after unmount.
- Initial `updateHeight()` call on mount ensures the property is set before the first paint.
- Empty dependency array `[]` is correct — all referenced values are either refs or stable DOM APIs.

**App.tsx**

- `pb-16 md:pb-0` replaced with `pb-bottom-bar` as intended. No other hardcoded bottom-padding classes remain on the main content div.

**styles.css**

- `--bottom-bar-height: 0px` default in `:root` is correct — desktop gets 0 padding without any JS running.
- `.pb-bottom-bar` utility class correctly uses `var(--bottom-bar-height, 0px)` with a fallback.

**ios-safe-area.test.ts**

- 8 source-text assertions cover the complete contract: viewport meta tag, pt-safe, pb-bottom-bar presence, pb-16 absence, ResizeObserver usage, window resize listener, cancelAnimationFrame presence, and the 0px reset on unmount.

### No issues found

All P0/P1/P2/P3 items from previous iterations are resolved. The implementation is correct, complete, and well-tested for a CSS-custom-property-based layout approach.

PR #1169, opened 2026-04-01
feat: invalidate in-progress planning on goal title/description update

lsm:
When `update_goal` changes `title` or `description`, auto-cancel any in-progress planning tasks linked to the goal, reset `planning_attempts` to 0, and transition `needs_human` goals back to `active` so they become eligible for replanning. Calls `runtime.onGoalCreated()` to trigger a fresh planning tick.

- Non-title/description updates (e.g. priority) do not trigger invalidation
- 8 new unit tests covering all scenarios including the `needs_human`→`active` transition

### DIFF
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -206,6 +206,57 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				updated = await goalManager.updateGoalStatus(args.goal_id, args.status);
 			}
 
+			// Invalidate in-progress planning when title or description changes
+			if (args.title !== undefined || args.description !== undefined) {
+				const nonTerminalStatuses = new Set([
+					'pending',
+					'in_progress',
+					'draft',
+					'review',
+					'rate_limited',
+					'usage_limited',
+				]);
+
+				const runtime = runtimeService?.getRuntime(roomId) ?? null;
+
+				// Find and cancel any in-progress planning tasks linked to this goal
+				for (const taskId of goal.linkedTaskIds) {
+					const task = await taskManager.getTask(taskId);
+					if (!task || task.taskType !== 'planning' || !nonTerminalStatuses.has(task.status))
+						continue;
+
+					if (runtime) {
+						await runtime.cancelTask(taskId);
+					} else {
+						await taskManager.cancelTaskCascade(taskId);
+						if (daemonHub) {
+							const cancelledTask = await taskManager.getTask(taskId);
+							if (cancelledTask) {
+								void daemonHub.emit('room.task.update', {
+									sessionId: `room:${roomId}`,
+									roomId,
+									task: cancelledTask,
+								});
+							}
+						}
+					}
+				}
+
+				// Reset planning_attempts so a fresh planner picks up the updated context
+				updated = await goalManager.patchGoal(args.goal_id, { planning_attempts: 0 });
+
+				// If the goal is stuck in needs_human, transition it back to active so it's
+				// eligible for replanning (getNextGoalForPlanning only iterates 'active' goals)
+				if (updated.status === 'needs_human') {
+					updated = await goalManager.updateGoalStatus(args.goal_id, 'active');
+				}
+
+				// Trigger a fresh planning tick if runtime is available
+				if (runtime) {
+					runtime.onGoalCreated(args.goal_id);
+				}
+			}
+
 			return jsonResult({ success: true, goal: updated });
 		},
 
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -229,6 +229,22 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 					if (runtime) {
 						await runtime.cancelTask(taskId);
 					} else {
+						// Fallback path: no runtime available.
+						// Guard: if the task has an active session group we cannot safely cancel it
+						// without the runtime — the DB row would be updated but the agent sessions
+						// would keep running (same guard as reset_goal / cancel_task).
+						if (task.status === 'in_progress' || task.status === 'review') {
+							const activeGroup = groupRepo.getGroupByTaskId(taskId);
+							if (activeGroup && activeGroup.completedAt === null) {
+								return jsonResult({
+									success: false,
+									error:
+										`Cannot invalidate planning for goal ${args.goal_id}: planning task ${taskId} (status '${task.status}') has an active session group but no runtime service is available to stop it. ` +
+										`The active planner session would not be stopped.`,
+								});
+							}
+						}
+
 						await taskManager.cancelTaskCascade(taskId);
 						if (daemonHub) {
 							const cancelledTask = await taskManager.getTask(taskId);
@@ -250,8 +266,10 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				}
 
 				// If the goal is stuck in needs_human, transition it back to active so it's
-				// eligible for replanning (getNextGoalForPlanning only iterates 'active' goals)
-				if (updated.status === 'needs_human') {
+				// eligible for replanning (getNextGoalForPlanning only iterates 'active' goals).
+				// Only auto-transition when the caller did not explicitly set a status in this
+				// same call — an explicit status instruction takes precedence.
+				if (args.status === undefined && updated.status === 'needs_human') {
 					updated = await goalManager.updateGoalStatus(args.goal_id, 'active');
 				}
 
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -49,6 +49,16 @@ function jsonResult(data: Record<string, unknown>): ToolResult {
 	return { content: [{ type: 'text', text: JSON.stringify(data) }] };
 }
 
+/** Task statuses that are not yet terminal and should be cancelled when resetting/invalidating goals. */
+const NON_TERMINAL_TASK_STATUSES = new Set([
+	'pending',
+	'in_progress',
+	'draft',
+	'review',
+	'rate_limited',
+	'usage_limited',
+]);
+
 /**
  * Create tool handler functions that can be tested directly.
  * Returns a map of tool name -> handler function.
@@ -208,21 +218,12 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 
 			// Invalidate in-progress planning when title or description changes
 			if (args.title !== undefined || args.description !== undefined) {
-				const nonTerminalStatuses = new Set([
-					'pending',
-					'in_progress',
-					'draft',
-					'review',
-					'rate_limited',
-					'usage_limited',
-				]);
-
 				const runtime = runtimeService?.getRuntime(roomId) ?? null;
 
 				// Find and cancel any in-progress planning tasks linked to this goal
 				for (const taskId of goal.linkedTaskIds) {
 					const task = await taskManager.getTask(taskId);
-					if (!task || task.taskType !== 'planning' || !nonTerminalStatuses.has(task.status))
+					if (!task || task.taskType !== 'planning' || !NON_TERMINAL_TASK_STATUSES.has(task.status))
 						continue;
 
 					if (runtime) {
@@ -242,8 +243,11 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 					}
 				}
 
-				// Reset planning_attempts so a fresh planner picks up the updated context
-				updated = await goalManager.patchGoal(args.goal_id, { planning_attempts: 0 });
+				// Reset planning_attempts so a fresh planner picks up the updated context.
+				// Skip if the caller explicitly provided planning_attempts — their value takes precedence.
+				if (args.planning_attempts === undefined) {
+					updated = await goalManager.patchGoal(args.goal_id, { planning_attempts: 0 });
+				}
 
 				// If the goal is stuck in needs_human, transition it back to active so it's
 				// eligible for replanning (getNextGoalForPlanning only iterates 'active' goals)
@@ -1064,23 +1068,13 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 				return jsonResult({ success: false, error: `Goal not found: ${args.goal_id}` });
 			}
 
-			// Non-terminal task statuses — tasks in these states should be cancelled before reset
-			const nonTerminalStatuses = new Set([
-				'pending',
-				'in_progress',
-				'draft',
-				'review',
-				'rate_limited',
-				'usage_limited',
-			]);
-
 			// Hoist the runtime lookup so it isn't repeated on every loop iteration
 			const runtime = runtimeService?.getRuntime(roomId) ?? null;
 
 			// Cancel all in-progress linked tasks
 			for (const taskId of goal.linkedTaskIds) {
 				const task = await taskManager.getTask(taskId);
-				if (!task || !nonTerminalStatuses.has(task.status)) continue;
+				if (!task || !NON_TERMINAL_TASK_STATUSES.has(task.status)) continue;
 
 				if (runtime) {
 					await runtime.cancelTask(taskId);
--- a/packages/daemon/src/lib/room/tools/room-agent-tools.ts
+++ b/packages/daemon/src/lib/room/tools/room-agent-tools.ts
@@ -180,6 +180,34 @@ export function createRoomAgentToolHandlers(config: RoomAgentToolsConfig) {
 
 			let updated = goal;
 
+			// Pre-scan: if title/description is changing and runtime is unavailable, check for
+			// blocked planning tasks BEFORE any DB writes so we fail atomically.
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

The implementation correctly addresses all requirements: cancelling in-progress planning tasks, resetting `planning_attempts`, transitioning `needs_human` to `active`, and triggering replanning. The code integrates cleanly with existing patterns (runtime vs. fallback paths, daemonHub emit). Tests are thorough with 8 scenarios covering the happy path, negative cases, runtime mocks, and terminal-status preservation.

### P2: `planning_attempts` reset silently overrides explicit caller input

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 190 and 246

If someone calls `update_goal({ goal_id, title: 'New', planning_attempts: 5 })`, line 190 patches `planning_attempts` to 5, then line 246 immediately overwrites it to 0. The invalidation silently discards the explicit value. This is a minor concern since it's unlikely a caller would intentionally set a non-zero `planning_attempts` while also changing the title/description (which semantically invalidates the old planning context), but it would be cleaner to either skip the reset when `args.planning_attempts` is explicitly provided, or document this precedence.

### P3: `nonTerminalStatuses` set could be extracted

The `nonTerminalStatuses` set (line 211-218) is defined inline inside the handler. If these statuses are already defined elsewhere (e.g., task status utilities), consider reusing them. Minor duplication concern only.

### P3: No test for combined `status` + title/description change

There's no test covering the edge case where `update_goal` is called with both `status: 'needs_human'` and `title: 'new title'` simultaneously. The current code would set status to `needs_human` (line 206), then the invalidation block would transition it back to `active` (line 251). This behavior is arguably correct but untested.

Overall the implementation is clean, well-scoped, and the test coverage is solid.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two meaningful correctness issues found; the rest is solid work.

---

### P0 — Missing active-session-group guard in the fallback path

`reset_goal` (lines 1094–1104) guards against cancelling `in_progress` or `review` tasks without the runtime when there is an active session group, returning an error rather than silently leaving the worker session running. The new `update_goal` fallback path (lines 230–242) performs `taskManager.cancelTaskCascade(taskId)` with no equivalent guard.

Because planning tasks can legitimately be `in_progress` (they are part of the `nonTerminalStatuses` set), if `runtime` is `null` and a planning task is in that state with a live session group, the code will mark the task cancelled in the DB while the agent session continues to execute. The `reset_goal` comments even explain why this is dangerous: *"the DB row would be updated but the agent sessions would keep running."*

The fix is to mirror what `reset_goal` does: check `groupRepo.getGroupByTaskId(taskId)` when `task.status` is `in_progress` or `review` and return an error (or log a warning and skip the task) if an active group exists.

`packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 228–243.

---

### P1 — `planning_attempts` user value is silently clobbered when title/description is also changed

If a caller sends `{ goal_id, title: 'X', planning_attempts: 5 }`, the handler applies the explicit `planning_attempts: 5` patch first (line 201), then unconditionally overwrites it with `{ planning_attempts: 0 }` (line 246) because `args.title !== undefined`. The explicit user value is discarded without any signal to the caller.

This is unlikely to occur in practice because the agent would not simultaneously update the title and set `planning_attempts`, but the semantic is surprising. At minimum the code should skip the reset when `args.planning_attempts` is explicitly provided (or document the override priority).

`packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 180–201 vs. 246.

---

### P2 — `updated.status` after `patchGoal` may not reflect a concurrently applied `args.status`

The status flow is:

1. `patchGoal(patch)` → sets `updated` (line 201)
2. `updateGoalStatus(args.status)` → updates `updated` (line 206)
3. `patchGoal({ planning_attempts: 0 })` → updates `updated` again (line 246)
4. `if (updated.status === 'needs_human')` → checks status from step 3

`patchGoal` patches non-status columns; it does not change `status`. So the status in step 3's return value reflects whatever `updateGoalStatus` set. In practice this is correct today. However if the caller simultaneously passes `status: 'needs_human'` together with `title`, the code will transition `needs_human → active` immediately after, silently ignoring the caller's explicit `status: 'needs_human'` instruction. This is a minor semantic inconsistency worth a comment or a guard (`if (args.status === undefined && updated.status === 'needs_human')`).

`packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 204–251.

---

### P3 — Duplicated `nonTerminalStatuses` constant

The identical `nonTerminalStatuses` set is defined in three places: `update_goal` (line 211), `reset_goal` (line 1068), and `cancel_task` area. Extracting it to a module-level constant would reduce the risk of the three copies diverging (e.g., a new status being added to one but not the others).

`packages/daemon/src/lib/room/tools/room-agent-tools.ts`.

---

### Tests

The 8 new tests are well-structured and cover the happy path thoroughly. One gap worth noting:

- There is no test for the fallback path where `runtime` is `null` and a planning task has `in_progress` status with an active session group. The `reset_goal` tests presumably cover that pattern — the same coverage should exist here given it is an intentional safety boundary.

Overall the implementation is logically correct for the common case; the active-session guard omission is the only one that can leave the system in an inconsistent state at runtime.

lsm (commented):

The second commit addressed two of the three issues from the previous review: the `planning_attempts` explicit override guard (P1) and the duplicated `nonTerminalStatuses` constant (P3, now extracted as module-level `NON_TERMINAL_TASK_STATUSES`). Both fixes are correct and well-tested (new test at line 716 for the override, and the constant is shared between `update_goal` and `reset_goal`). The combined `status: needs_human` + title change edge case (previously P2) also has a new test at line 733. Good work on those.

However, the primary P0 flagged in the previous review remains unaddressed.

### P0 -- Missing active-session-group guard in fallback cancellation path

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`, lines 231-243

The `reset_goal` handler (lines 1084-1098) guards against cancelling `in_progress` or `review` tasks without the runtime when an active session group exists:

```
if (task.status === 'in_progress' || task.status === 'review') {
    const activeGroup = groupRepo.getGroupByTaskId(taskId);
    if (activeGroup && activeGroup.completedAt === null) {
        return jsonResult({ success: false, error: ... });
    }
}
```

The `update_goal` fallback path at lines 231-243 performs `taskManager.cancelTaskCascade(taskId)` with no equivalent guard. Because planning tasks can legitimately be `in_progress` (they are in `NON_TERMINAL_TASK_STATUSES`), if `runtime` is `null` and a planning task is in that state with a live session group, the code will mark the task as cancelled in the DB while the agent session continues to execute. The `reset_goal` comments explicitly explain why this is dangerous.

The fix should mirror what `reset_goal` does: check `groupRepo.getGroupByTaskId(taskId)` when `task.status` is `in_progress` or `review`, and either return an error or skip that task with a warning. A corresponding test should be added.

### Summary

| Severity | Count | Details |
|----------|-------|---------|
| P0 | 1 | Missing active-session-group guard in `update_goal` fallback path |
| P1 | 0 | -- |
| P2 | 0 | -- |
| P3 | 0 | -- |

The `planning_attempts` override and constant deduplication are both correctly implemented and tested. The remaining P0 is the only blocking issue.

lsm (commented):

**Recommendation: APPROVE**

Both previously raised P0 and P1 issues have been fully addressed in the follow-up commit. The implementation is now correct and complete.

---

### P0 — Active-session-group guard: FIXED

The fallback path in `update_goal` now mirrors `reset_goal` exactly. Lines 236–246 of `room-agent-tools.ts` check `groupRepo.getGroupByTaskId(taskId)` and return an error when `in_progress` or `review` planning tasks have an active, not-yet-completed session group and no runtime is available to stop them. The DB cancellation is correctly blocked in that case, preventing the inconsistent state where the DB row would be marked cancelled while the agent session continues running.

A dedicated test for this guard was added at line 755 of `room-agent-tools-goal.test.ts`, reusing the `insertActiveGroup` helper and asserting both the error response and the preserved `in_progress` status on the planning task.

---

### P1 — `planning_attempts` explicit override: FIXED

The auto-reset to 0 is now conditional on `args.planning_attempts === undefined` (line 264). A caller who explicitly provides `planning_attempts` alongside a title/description change will have their value respected. This is tested at line 716.

---

### P2 — Explicit `status` + title/description precedence: FIXED

The `needs_human → active` transition is now guarded with `args.status === undefined` (line 272), so an explicit `status: 'needs_human'` instruction from the caller correctly takes precedence over the auto-transition. This is tested at line 733, and the test expectation was flipped from the previous version (the goal remains `needs_human` when status is explicitly passed, which is the right behavior).

---

### P3 — Duplicated `nonTerminalStatuses` constant: FIXED

The per-function `nonTerminalStatuses` local sets have been replaced with a single module-level `NON_TERMINAL_TASK_STATUSES` constant shared by both `update_goal` and `reset_goal`.

---

### Summary

All 4 issues from the previous review (P0, P1, P2, P3) are resolved. The 10 new unit tests (8 original + 2 from the refactor commit) provide solid coverage across the happy path, the active-session-group guard, explicit override precedence, and the combined status+title edge case.

lsm (commented):

**Recommendation: APPROVE**

All issues from the previous two review rounds have been addressed. The implementation is correct and complete.

### Verified fixes

1. **Active-session-group guard (previously P0):** The fallback cancellation path in `update_goal` (lines 236-245 of `room-agent-tools.ts`) now mirrors `reset_goal` exactly -- it checks `groupRepo.getGroupByTaskId(taskId)` for `in_progress`/`review` tasks and returns an error when an active session group exists without a runtime to stop it. Covered by the new test at line 755.

2. **Explicit `status` precedence (previously P2):** The `needs_human` to `active` auto-transition is guarded by `args.status === undefined` (line 272), so an explicit `status: 'needs_human'` passed alongside a title/description change is respected. Covered by the test at line 733.

3. **Explicit `planning_attempts` precedence (previously P1):** The auto-reset to 0 is conditional on `args.planning_attempts === undefined` (line 264). Covered by the test at line 716.

4. **Deduplicated constant (previously P3):** `NON_TERMINAL_TASK_STATUSES` is now a single module-level constant shared by both `update_goal` and `reset_goal`.

### Notes

One minor observation: when the session-group guard triggers and returns `success: false`, the title/description patch has already been committed to the DB (line 211 runs before the invalidation block at line 220). This means the goal text is updated but planning is not invalidated. This is acceptable behavior -- the user's content change is preserved and the error message clearly explains why planning invalidation could not proceed. The caller can retry once the runtime becomes available. This matches the overall tool design where each operation is best-effort and returns descriptive errors.

All 26 tests pass. No issues found.

| Severity | Count |
|----------|-------|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 0 |

lsm (commented):

The three stated fixes (session-group guard, explicit-status precedence, updated tests) are all present and structurally correct. However there is one P1 bug introduced in this PR that remains unaddressed, and the associated test does not catch it.

---

### P1 — Partial update committed before the session-group guard fires

**File:** `packages/daemon/src/lib/room/tools/room-agent-tools.ts`

The execution order in `update_goal` is:

1. `patchGoal(args.goal_id, patch)` — writes the new title/description to the DB (line 211)
2. Status update applied if present (lines 215-217)
3. Loop over linked tasks; on encountering an in-progress planning task with an active session group and no runtime → **`return jsonResult({ success: false, error: ... })`** (lines 239-244)

When the early return fires, the title/description has **already been committed to the DB**. The caller receives `success: false` but the goal's title and/or description was silently mutated. This is a partial-update inconsistency: the operation appears atomic from the caller's perspective but it is not.

Compare with `reset_goal` (line 1083+): the goal state reset (`goalManager.resetGoal`) only happens **after** all the task-guard checks pass. There is no equivalent pre-check phase in `update_goal`.

**The existing test for this scenario (`should return error when in_progress planning task has active session group and no runtime`) does not assert that the goal's title remained at its original value after the failed call.** It only checks that `planTask.status` is still `in_progress`, so the title-mutation bug is invisible.

**Fix options:**
- Do an up-front pre-scan of linked planning tasks for the blocking condition before calling `patchGoal`, and return early if blocked. This is consistent with how `reset_goal` works.
- Or (if the semantic is "title/description change always succeeds, but planning invalidation may be skipped with a warning"), change the response to `success: true` with a `warning` field explaining the planning task could not be cancelled.

The choice depends on the intended contract of the API, but right now the contract is broken: the DB state and the API response are contradictory.

---

### P2 — Test does not verify goal title was not mutated on error

**File:** `packages/daemon/tests/unit/room/room-agent-tools-goal.test.ts`, line 771

The comment says "Planning task and goal should be unchanged" but only asserts on `planTask.status`. The goal title should also be asserted (e.g. `expect(goal.title).toBe('Goal')`) to catch the partial-update regression. As-is the test gives a false sense of correctness.

---

### What looks good

- Extracting `NON_TERMINAL_TASK_STATUSES` as a module-level constant is a clean refactor — eliminates duplication between `update_goal` and `reset_goal`.
- The explicit-status-precedence logic (`args.status === undefined && updated.status === 'needs_human'`) is correct and well-tested.
- The explicit-`planning_attempts` override guard (`args.planning_attempts === undefined`) is correct.
- All 10 new test cases cover their intended scenarios and assertions are accurate (with the gap noted above).
- The `taskType !== 'planning'` filter correctly limits cancellation to planning tasks only, not all linked tasks.

lsm (commented):

**Recommendation: APPROVE**

The P1 partial-update bug from the previous review is correctly fixed. The pre-scan block now runs before `patchGoal()`, ensuring that when a title/description change encounters a blocked planning task (active session group, no runtime), the handler returns `{ success: false }` before any DB mutation occurs. The test at line 769 confirms the goal title remains unchanged after a blocked failure.

### What I verified

1. **Pre-scan correctness** (`room-agent-tools.ts:183-209`): The pre-scan only triggers when title/description is changing AND no runtime is available. It checks all linked planning tasks for the in_progress/review + active-session-group condition and returns early before any writes. This matches the pattern used in `reset_goal` (line 1117-1129).

2. **Post-patch invalidation** (`room-agent-tools.ts:247-295`): After the DB patch, the invalidation loop cancels non-terminal planning tasks via runtime (if available) or `cancelTaskCascade` (fallback). The fallback is safe because the pre-scan already rejected any blocked tasks. The `planning_attempts` reset respects explicit caller values. The `needs_human` to `active` transition respects explicit status args.

3. **Shared constant** (`NON_TERMINAL_TASK_STATUSES`): Hoisted to module scope, reused by both `update_goal` and `reset_goal`. Good deduplication.

4. **Test coverage** (267 new lines, 8 tests): Covers title change, description change, priority-only (no invalidation), no planning tasks, `needs_human` transition, runtime integration, terminal task skip, explicit `planning_attempts` override, explicit status override, and the blocked session group guard including the title-not-mutated assertion.

### Minor observations (P3)

- The runtime is looked up twice when title/description changes: once in the pre-scan (line 186) and once in the post-patch block (line 249). This is harmless since `getRuntime` is a synchronous lookup, but a single variable could be shared if desired.

lsm (commented):

**Recommendation: APPROVE**

The P1 partial-update bug I raised in the previous round is now fully addressed. Here is the complete verification.

---

### Previously Raised P1 — Partial update committed before session-group guard fires: FIXED

The pre-scan block at lines 183-209 of `room-agent-tools.ts` now runs before `patchGoal()`. When title or description is changing and no runtime is available, it iterates all linked planning tasks and, for any task in `in_progress` or `review` status with an active (not-yet-completed) session group, returns `{ success: false }` immediately. No DB write has occurred at that point.

The test at line 755 of `room-agent-tools-goal.test.ts` now asserts at line 777 that `goalAfter?.title` is still `'Goal'` (the original value) after a guard-blocked failure, proving atomicity. This is the exact assertion that was missing in the previous round.

---

### All previous issues — status

| Issue | Status |
|-------|--------|
| P0 — Missing active-session-group guard in fallback path | Fixed (commit ae1cdd4c) |
| P1 — `planning_attempts` user value silently clobbered | Fixed (commit ae1cdd4c) |
| P1 — Partial update committed before guard fires | Fixed (commit 58e080d3) |
| P2 — Explicit `status` + title/description precedence | Fixed (commit ae1cdd4c) |
| P3 — Duplicated `nonTerminalStatuses` constant | Fixed (commit ae1cdd4c) |
| P2 — Test did not assert goal title was not mutated | Fixed (commit 58e080d3, line 777) |

---

### Minor remaining observation (P3, non-blocking)

The pre-scan guard correctly handles both `in_progress` and `review` task statuses (line 196), but the test suite only exercises the `in_progress` path. A test with `review` status would give symmetric coverage. This is not blocking — the logic is identical for both statuses and the code path is structurally verified.

The runtime is also looked up twice when title/description changes: once in the pre-scan (line 186) and once in the post-patch invalidation block (line 249). Both calls are synchronous and harmless, but a single variable hoisted above both blocks would be cleaner. Neither observation warrants blocking the merge.

---

### Implementation correctness verified

- Pre-scan at lines 183-209 runs before any DB write and returns early on guard failure.
- Post-patch invalidation loop at lines 247-295 safely calls `cancelTaskCascade` in the fallback path because the pre-scan already rejected any tasks with active session groups.
- `planning_attempts` reset at line 279 is conditional on `args.planning_attempts === undefined`.
- `needs_human → active` auto-transition at line 287 is guarded by `args.status === undefined`.
- `NON_TERMINAL_TASK_STATUSES` is a single module-level constant shared between `update_goal` and `reset_goal`.
- Test at line 777 confirms no title mutation on guard-blocked failure.

No blocking issues. The implementation is correct and complete.

PR #1171, opened 2026-04-01
feat: make CreateRoomParams.defaultPath required and add workspace path validation

lsm:
Make `defaultPath` required on `CreateRoomParams` (Milestone 1.1 of decouple-daemon-from-single-workspaceroot plan).

- `CreateRoomParams.defaultPath` changed from `optional` → `required string`
- New `validateWorkspacePath(path)` utility in `packages/shared/src/validation/workspace-path.ts` — checks non-empty absolute POSIX path (no filesystem access, format only)
- Exported from `@neokai/shared`
- All caller type errors fixed: `room-handlers.ts` falls back to `workspaceRoot ?? ''`, `Lobby.tsx` passes placeholder `defaultPath: ''` until Milestone 2 adds the path input field
- Unit tests in `packages/shared/tests/workspace-path.test.ts`

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
@@ -50,16 +50,18 @@ export function setupRoomHandlers(
 		const allowedPaths = params.allowedPaths ?? (workspaceRoot ? [{ path: workspaceRoot }] : []);
 
 		// TODO(Milestone 2, task 2.1): Remove workspaceRoot fallback; require defaultPath
-		// from the client. Using undefined (not '') preserves downstream `room.defaultPath ??
-		// ctx.defaultWorkspacePath` fallback chains until enforcement is added.
-		const defaultPath = (params.defaultPath ?? workspaceRoot) as string;
+		// from the client. Using || (not ??) also catches empty-string placeholders sent
+		// by callers that have not yet been updated. When both are absent, defaultPath is
+		// omitted from the params object so downstream `room.defaultPath ?? fallback`
+		// chains in room-runtime-service still work correctly.
+		const defaultPath = params.defaultPath || workspaceRoot;
 
 		const room = roomManager.createRoom({
 			name: params.name,
 			background: params.background,
 			allowedPaths,
-			defaultPath,
-		});
+			...(defaultPath ? { defaultPath } : {}),
+		} as CreateRoomParams);
 
 		// Create the room's user-facing chat session
 		// Session ID format: room:chat:${roomId}
--- a/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
@@ -51,7 +51,7 @@ export function setupRoomHandlers(
 
 		// Auto-populate workspace paths from workspaceRoot if not provided
 		const allowedPaths = params.allowedPaths ?? (workspaceRoot ? [{ path: workspaceRoot }] : []);
-		const defaultPath = params.defaultPath ?? (workspaceRoot ? workspaceRoot : undefined);
+		const defaultPath = params.defaultPath ?? workspaceRoot ?? '';
 
 		const room = roomManager.createRoom({
 			name: params.name,
--- a/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
@@ -13,7 +13,7 @@
  * - agents.cli.list - List detected CLI agents
  */
 
-import type { MessageHub, WorkspacePath } from '@neokai/shared';
+import type { MessageHub, WorkspacePath, CreateRoomParams } from '@neokai/shared';
 import type { DaemonHub } from '../daemon-hub';
 import type { RoomManager } from '../room/managers/room-manager';
 import type { RoomRuntimeService } from '../room/runtime/room-runtime-service';
@@ -38,12 +38,9 @@ export function setupRoomHandlers(
 ): void {
 	// room.create - Create a new room
 	messageHub.onRequest('room.create', async (data) => {
-		const params = data as {
-			name: string;
-			background?: string;
-			allowedPaths?: WorkspacePath[];
-			defaultPath?: string;
-		};
+		// Wire type: defaultPath is still optional at the RPC boundary for backward
+		// compatibility until Milestone 2 (task 2.1) enforces it from the client.
+		const params = data as Omit<CreateRoomParams, 'defaultPath'> & { defaultPath?: string };
 
 		if (!params.name) {
 			throw new Error('Room name is required');
@@ -51,7 +48,11 @@ export function setupRoomHandlers(
 
 		// Auto-populate workspace paths from workspaceRoot if not provided
 		const allowedPaths = params.allowedPaths ?? (workspaceRoot ? [{ path: workspaceRoot }] : []);
-		const defaultPath = params.defaultPath ?? workspaceRoot ?? '';
+
+		// TODO(Milestone 2, task 2.1): Remove workspaceRoot fallback; require defaultPath
+		// from the client. Using undefined (not '') preserves downstream `room.defaultPath ??
+		// ctx.defaultWorkspacePath` fallback chains until enforcement is added.
+		const defaultPath = (params.defaultPath ?? workspaceRoot) as string;
 
 		const room = roomManager.createRoom({
 			name: params.name,
--- a/packages/daemon/tests/unit/rpc/room-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/room-handlers.test.ts
@@ -183,7 +183,7 @@ describe('Room RPC Handlers', () => {
 				name: 'Full Room',
 				background: 'A full featured room',
 				allowedPaths: [],
-				defaultPath: undefined,
+				defaultPath: '',
 			});
 		});
 
--- a/packages/daemon/tests/unit/rpc/room-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/room-handlers.test.ts
@@ -196,7 +196,7 @@ describe('Room RPC Handlers', () => {
 			);
 		});
 
-		it('treats empty-string defaultPath as absent (falls back when workspaceRoot available)', async () => {
+		it('omits defaultPath key when client sends empty string', async () => {
 			// '' should be treated the same as absent because we use || not ??
 			const handler = messageHubData.handlers.get('room.create');
 			expect(handler).toBeDefined();
--- a/packages/daemon/tests/unit/rpc/room-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/room-handlers.test.ts
@@ -167,7 +167,7 @@ describe('Room RPC Handlers', () => {
 			);
 		});
 
-		it('creates a room with background (no defaultPath — falls back to undefined)', async () => {
+		it('creates a room with background (no defaultPath — omitted when absent)', async () => {
 			const handler = messageHubData.handlers.get('room.create');
 			expect(handler).toBeDefined();
 
@@ -179,17 +179,34 @@ describe('Room RPC Handlers', () => {
 				{}
 			);
 
-			// No workspaceRoot configured in this test setup, so defaultPath falls back
-			// to undefined (cast as string). The ?? chain in room-runtime-service still
-			// works correctly because null-coalescing passes through undefined.
-			// TODO(Milestone 2, task 2.1): Once the server enforces defaultPath, this
-			// test should pass an explicit path and this assertion should use it.
-			expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith({
-				name: 'Full Room',
-				background: 'A full featured room',
-				allowedPaths: [],
-				defaultPath: undefined,
-			});
+			// No workspaceRoot in this test setup and no defaultPath from the client, so
+			// defaultPath is omitted from the createRoom call entirely. Downstream ??
+			// chains in room-runtime-service then fall through to their own fallbacks.
+			// TODO(Milestone 2, task 2.1): Once the server enforces defaultPath from the
+			// client, this test should pass an explicit path.
+			expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith(
+				expect.not.objectContaining({ defaultPath: expect.anything() })
+			);
+			expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith(
+				expect.objectContaining({
+					name: 'Full Room',
+					background: 'A full featured room',
+					allowedPaths: [],
+				})
+			);
+		});
+
+		it('treats empty-string defaultPath as absent (falls back when workspaceRoot available)', async () => {
+			// '' should be treated the same as absent because we use || not ??
+			const handler = messageHubData.handlers.get('room.create');
+			expect(handler).toBeDefined();
+
+			await handler!({ name: 'Empty Path Room', defaultPath: '' }, {});
+
+			// No workspaceRoot in this test setup so defaultPath is still omitted
+			expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith(
+				expect.not.objectContaining({ defaultPath: expect.anything() })
+			);
 		});
 
 		it('creates a room with an explicit defaultPath', async () => {
--- a/packages/daemon/tests/unit/rpc/room-handlers.test.ts
+++ b/packages/daemon/tests/unit/rpc/room-handlers.test.ts
@@ -167,7 +167,7 @@ describe('Room RPC Handlers', () => {
 			);
 		});
 
-		it('creates a room with background', async () => {
+		it('creates a room with background (no defaultPath — falls back to undefined)', async () => {
 			const handler = messageHubData.handlers.get('room.create');
 			expect(handler).toBeDefined();
 
@@ -179,12 +179,37 @@ describe('Room RPC Handlers', () => {
 				{}
 			);
 
+			// No workspaceRoot configured in this test setup, so defaultPath falls back
+			// to undefined (cast as string). The ?? chain in room-runtime-service still
+			// works correctly because null-coalescing passes through undefined.
+			// TODO(Milestone 2, task 2.1): Once the server enforces defaultPath, this
+			// test should pass an explicit path and this assertion should use it.
 			expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith({
 				name: 'Full Room',
 				background: 'A full featured room',
 				allowedPaths: [],
-				defaultPath: '',
+				defaultPath: undefined,
 			});
+		});
+
+		it('creates a room with an explicit defaultPath', async () => {
+			const handler = messageHubData.handlers.get('room.create');
+			expect(handler).toBeDefined();
+
+			await handler!(
+				{
+					name: 'Path Room',
+					defaultPath: '/workspace/my-project',
+				},
+				{}
+			);
+
+			expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith(
+				expect.objectContaining({
+					name: 'Path Room',
+					defaultPath: '/workspace/my-project',
+				})
+			);
 		});
 
 		it('throws error when name is missing', async () => {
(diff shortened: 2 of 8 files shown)


lsm (commented):

### Summary

The type change to make `defaultPath` required on `CreateRoomParams` is correct in intent, and the `validateWorkspacePath` utility is well-written with good tests. However, the implementation has significant gaps that undermine the goal of this milestone.

### Issues

**P1: `validateWorkspacePath` is exported but never called in application code**

The validation function exists in `packages/shared/src/validation/workspace-path.ts` and is exported from `@neokai/shared`, but it is not called anywhere in the actual code paths — not in the RPC handler (`room-handlers.ts`), not in `roomManager.createRoom`, not in the frontend. This means invalid paths (empty strings, relative paths) can still flow through and be stored in the database unchecked. The function is dead code outside of its own test file.

At minimum, the `room.create` RPC handler should call `validateWorkspacePath(defaultPath)` and throw if the result is invalid — or at least call it in `RoomRepository.createRoom` as a guard.

**P1: Empty string `''` as defaultPath creates a silent runtime bug**

The implementation uses `''` as the fallback everywhere:
- `room-handlers.ts:54`: `params.defaultPath ?? workspaceRoot ?? ''`
- `Lobby.tsx:269`: `{ ...params, defaultPath: '' }`
- `Lobby.tsx:287`: `defaultPath: params.defaultPath ?? ''`

An empty string is truthy in JavaScript, so downstream code like `room-runtime-service.ts:638`:
```typescript
const workspacePath = room.defaultPath ?? this.ctx.defaultWorkspacePath;
```
will use `''` instead of falling through to `defaultWorkspacePath`. This means rooms created without an explicit path will start sessions with an empty workspace path rather than the daemon's configured workspace root. This is worse than the previous behavior where `undefined` would correctly fall through to the `??` fallback.

If the intent is "required but may be empty temporarily until Milestone 2 adds the UI," then the downstream code needs to handle empty-string-as-unset (e.g., `room.defaultPath || this.ctx.defaultWorkspacePath`). But that would undermine making the field "required" in the first place.

**P2: RPC handler defines a local inline type instead of using `CreateRoomParams`**

In `packages/daemon/src/lib/rpc-handlers/room-handlers.ts:41-46`, the `room.create` handler casts `data` to an inline type:
```typescript
const params = data as {
    name: string;
    background?: string;
    allowedPaths?: WorkspacePath[];
    defaultPath?: string;  // still optional here!
};
```

This local type still has `defaultPath` as optional, completely bypassing the type safety gained by making it required in `CreateRoomParams`. The handler should import and use `CreateRoomParams` directly so that future type changes are enforced at the RPC boundary.

**P2: Tests for `room-handlers.test.ts` changed to pass `defaultPath: ''` without verifying this is actually valid**

The test at line 186 now passes `defaultPath: ''` in the expected call, but there is no assertion that this is a valid state. The test should either verify that validation rejects empty strings, or document why empty strings are acceptable.

### What's good

- The `validateWorkspacePath` function itself is clean and well-documented (POSIX-only scope, format-only check).
- The test file covers the important cases (empty, whitespace, relative, absolute, root).
- The `WorkspacePathValidationResult` return type is a good pattern for validation functions.

The core problem is that making `defaultPath` required was done at the type level but not enforced at runtime. The validation function was written but never wired in. And the `''` fallback introduces a regression where the `??` fallback chain in `room-runtime-service.ts` no longer works as intended. These need to be addressed before merging.

lsm (commented):

---

### Summary

The type change and validation utility are correctly implemented and well-tested in isolation. However, the PR leaves several call sites emitting or accepting `defaultPath: ''` (an empty string) as a valid `CreateRoomParams` value — which the new `validateWorkspacePath` rejects as invalid. The type contract says "required" but the runtime meaning is still hollow because there is no enforcement at the API boundary. The most critical omission is that `room-handlers.ts` never calls `validateWorkspacePath` before storing the room, making the new utility dead code from the server's perspective.

---

### P0 — Bugs / Broken Functionality

**1. `validateWorkspacePath` is never called in the RPC handler — validation is dead code**

`packages/daemon/src/lib/rpc-handlers/room-handlers.ts` line 54:
```ts
const defaultPath = params.defaultPath ?? workspaceRoot ?? '';
```
The fallback chain computes a `defaultPath` but never passes it through `validateWorkspacePath`. A client can send `{ name: 'X', defaultPath: '' }` (as the frontend actually does in `Lobby.tsx`) and the room will be created with an empty path. The plan document (task 2.1 in milestone 2) says filesystem existence is a Milestone 2 concern, but **format validation** was supposed to be enforced in this milestone by using the new utility. The utility was created but is entirely disconnected from the handler. If it is intentionally deferred to Milestone 2, that needs to be explicit in both the PR description and the handler code (a TODO comment at minimum).

**2. E2E room helper calls `room.create` without `defaultPath`**

`packages/e2e/tests/helpers/room-helpers.ts` line 22:
```ts
const res = await hub.request('room.create', { name: roomName });
```
The room is created with no `defaultPath` at all. The handler silently falls back to `workspaceRoot ?? ''`. This helper is used in over a dozen E2E test `beforeEach` setups. If Milestone 2 adds actual server-side enforcement, every one of those tests will break. Per the plan, test fixtures should use `/tmp/test-workspace` as the placeholder. This file was not updated.

---

### P1 — Should Fix

**3. `Lobby.tsx` `CreateRoomModal.onSubmit` always passes `defaultPath: ''` without a TODO**

`packages/web/src/islands/Lobby.tsx` lines 268-270:
```ts
onSubmit={async (params) => {
    const room = await lobbyStore.createRoom({ ...params, defaultPath: '' });
```
The `CreateRoomModal` only collects `name` and `background`. The `defaultPath: ''` override is added unconditionally, meaning real users creating rooms via this modal will always store an empty path. The plan acknowledges this is a Milestone 2 UI concern, but there is no code comment or TODO explaining that. A comment linking this placeholder to the Milestone 2 task would make the intent clear.

**4. Daemon test asserts `defaultPath: ''` as the expected value**

`packages/daemon/tests/unit/rpc/room-handlers.test.ts` lines 182-187:
```ts
expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith({
    name: 'Full Room',
    background: 'A full featured room',
    allowedPaths: [],
    defaultPath: '',
});
```
The test explicitly asserts that an empty `defaultPath` flows through to `createRoom`. This is not a test of the happy path and will need rewriting when enforcement is added. At least one test covering a valid absolute path should be added.

**5. No test covers `defaultPath` propagation with a valid path**

The updated `room-handlers.test.ts` does not add any test case that passes a valid `defaultPath` (e.g. `/workspace/my-project`) and verifies it reaches `createRoom`. The change to the required type is not tested beyond the error case.

---

**6. `NewSessionModal` internal callback type still marks `defaultPath` as optional**

`packages/web/src/components/lobby/NewSessionModal.tsx` line 71:
```ts
defaultPath?: string;
```
The internal `onCreateRoom` callback param type keeps `defaultPath` optional. This means callers inside this modal can compile without providing a path, undermining the type-level enforcement. Aligning to `defaultPath: string` (or leaving a TODO comment) would help.

**7. `Room.defaultPath` remains optional on the stored entity type without explanation**

`packages/shared/src/types/neo.ts` line 44: `defaultPath?: string`. The input requires a path but the stored entity allows it to be absent. This asymmetry is intentional for backward compatibility with persisted rows, but is not documented. A comment distinguishing the creation contract from the stored representation would prevent future confusion.

---

### P3 — Nits

- Test file location: the plan specified `packages/shared/src/validation/__tests__/workspace-path.test.ts` but the worker placed it at `packages/shared/tests/workspace-path.test.ts`. Works fine functionally, minor deviation from the plan.
- `validateWorkspacePath` accepts `/` (root path) as valid. While technically correct, root is not a meaningful workspace. A comment noting this is a deliberate decision would help.

lsm (commented):

This is a second-pass review. The worker addressed some first-round issues (empty-string fallback regression, TODO comments, E2E helper path). The core mechanics are now more honest. However, there are still meaningful gaps to address.

---

### P1 — `validateWorkspacePath` is exported but never called

The PR description says "New `validateWorkspacePath(path)` utility... checks non-empty absolute POSIX path". It is exported from `@neokai/shared` and tested in isolation, but it is **never called inside `room-handlers.ts`** or anywhere else in the daemon.

`packages/daemon/src/lib/rpc-handlers/room-handlers.ts` line 55:
```ts
const defaultPath = (params.defaultPath ?? workspaceRoot) as string;
```

There is no call to `validateWorkspacePath(defaultPath)` after this line. An invalid path (e.g. `"relative/path"` or `""`) from the client passes straight through to `roomManager.createRoom()` without any rejection. The utility is dead code from the system's perspective until it is actually wired in.

---

### P1 — `as string` cast hides `undefined` passing into a `string`-typed field

`CreateRoomParams.defaultPath` is `string` (required). The handler casts the computed value as `string` even when both `params.defaultPath` and `workspaceRoot` are absent, meaning `undefined as string` is silently passed to `roomManager.createRoom({ defaultPath: undefined })`. The unit test at line 187-192 explicitly asserts `defaultPath: undefined` — confirming the cast is lying to TypeScript, not fixing the problem.

The intent (per the TODO comment) is that downstream code handles `undefined` via its own `??` chains. That is a valid milestone plan, but the cast makes it harder to type-check the actual contract. A safer interim approach would be `params.defaultPath ?? workspaceRoot ?? undefined` (no cast), so TypeScript infers `string | undefined` and callers are forced to acknowledge it.

---

### P2 — Lobby.tsx still sends `defaultPath: ''` (empty string)

Both call sites in `packages/web/src/islands/Lobby.tsx` (lines 272 and 293) pass `defaultPath: ''` (literal empty string):

```ts
// line 272
const room = await lobbyStore.createRoom({ ...params, defaultPath: '' });

// line 293
const room = await lobbyStore.createRoom({
    ...params,
    defaultPath: params.defaultPath ?? '',
});
```

An empty string is **not** `undefined`. It does not trigger the `?? workspaceRoot` fallback on the server (because `'' ?? workspaceRoot` evaluates to `''`, not `workspaceRoot`). So when a user creates a room from the Lobby UI today, `defaultPath` is stored as `''`, which is a format-invalid path. The TODO comment is present, but the fallback logic is wrong for the stated milestone plan — the server should receive no `defaultPath` field (or `undefined`) rather than `''` to preserve the `??` chain.

---

### P2 — E2E helper uses `/tmp/test-workspace` which may not exist

`packages/e2e/tests/helpers/room-helpers.ts` line 25:
```ts
defaultPath: '/tmp/test-workspace',
```

This path is hardcoded and not guaranteed to exist in CI environments. If any E2E test path eventually flows through to filesystem access (e.g. session creation that resolves `workspacePath`), this will silently produce a non-existent path. The existing `getWorkspaceRoot()` helper is already available in the same file and returns the actual E2E workspace root — using that would be more correct even as a placeholder.

---

### P3 — Test asserts `defaultPath: undefined` cast-through rather than the actual desired state

`packages/daemon/tests/unit/rpc/room-handlers.test.ts` lines 187-192 explicitly assert:
```ts
expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith({
    ...
    defaultPath: undefined,
});
```

The comment says this is "expected until Milestone 2 enforcement". That is fine as a documented interim state, but it means the test validates the current broken behavior, not the target behavior. At minimum, the test name should make this clear (it currently says "falls back to undefined" in the title, which is reasonable).

---

### Summary

The empty-string fallback regression from the first pass has been fixed. However, the core validation utility is never called, the `as string` cast hides a type contract violation, and the Lobby UI still sends `''` which bypasses the `??` server fallback. The PR achieves the type-level intent (field is required in the shared type) but the runtime behavior at both the server and client call sites is still materially incorrect for the stated goal.

lsm (commented):

Second review pass. The fix for the empty-string fallback regression (`?? ''` -> `?? workspaceRoot`) is an improvement, but there is still a significant bug in the Lobby.tsx placeholder values that undermines the fix.

### P0: Lobby.tsx sends `defaultPath: ''` which bypasses the `??` fallback on the server (Lobby.tsx:272, 293)

Both call sites in Lobby.tsx pass `defaultPath: ''` as a placeholder. On the server (room-handlers.ts:55), the fallback is:

```ts
const defaultPath = (params.defaultPath ?? workspaceRoot) as string;
```

The `??` operator only triggers on `null` or `undefined` -- **not** on empty string. So when the client sends `defaultPath: ''`, the server uses `''` as-is instead of falling back to `workspaceRoot`. This means:

1. The room gets created with `defaultPath: ''` stored in SQLite (room-repository.ts:32 uses `params.defaultPath ?? null`, and `''` is not nullish).
2. The room chat session gets `workspacePath: ''` (room-handlers.ts:72).

The TODO comment on line 269 says "the server falls back to workspaceRoot when defaultPath is absent or empty" but that is **incorrect** -- the server only falls back when defaultPath is `undefined` or `null`, not `''`.

**Fix**: Either (a) change the server fallback to `(params.defaultPath || workspaceRoot)` to also catch empty strings, or (b) change the Lobby.tsx placeholders to `undefined` instead of `''`:

```ts
// Option A (server-side, preferred):
const defaultPath = (params.defaultPath || workspaceRoot) as string;

// Option B (client-side):
const room = await lobbyStore.createRoom({ ...params, defaultPath: undefined as unknown as string });
```

Option A is safer because it protects against any caller sending empty string.

### P1: Unsafe `as string` cast when both values are undefined (room-handlers.ts:55)

When `params.defaultPath` is undefined and `workspaceRoot` is also undefined (which is possible -- the parameter is typed `workspaceRoot?: string`), the expression `(params.defaultPath ?? workspaceRoot) as string` evaluates to `undefined as string`. This is a type-safety lie. The TODO comment acknowledges this is temporary, and the downstream `??` on line 72 does protect against it, but `roomManager.createRoom()` receives `undefined` typed as `string`, which then gets passed to `roomRepo.createRoom()` where `params.defaultPath ?? null` stores it as `null` in the DB. This works by accident rather than by design.

Consider at minimum adding a runtime check:

```ts
if (!defaultPath && !workspaceRoot) {
    throw new Error('Either defaultPath or workspaceRoot must be provided');
}
```

### P2: `validateWorkspacePath` is exported but never called outside tests

The validation function is defined, tested, and exported from `@neokai/shared`, but it is not called in any handler or UI code. The plan docs (02-room-creation-enforcement.md) show it should be used in both the server handler and the CreateRoomModal. Without actually calling it, invalid paths (relative paths, empty strings) pass through unchecked. This is presumably deferred to Milestone 2, but it would be worth noting in the PR description that this function is intentionally unused for now and will be wired in later.

### Minor observations (P3)

- The E2E helper hardcodes `/tmp/test-workspace` which is fine for now but the TODO correctly notes it should use the actual E2E workspace path. No action needed.
- The test for "creates a room with background" was renamed to clarify the fallback behavior, which is a good documentation improvement.
- The new test for explicit defaultPath propagation is well-structured and verifies the right thing.

**Recommendation: REQUEST_CHANGES** -- the P0 empty-string bypass in Lobby.tsx is a real functional bug that will cause rooms to be created with empty workspace paths instead of using the workspaceRoot fallback.

lsm (commented):

**Recommendation: APPROVE**

Third review pass. The issues raised in the first two reviews have been addressed:

- **Empty-string fallback (P0 from R1):** Fixed. `??` changed to `||` on line 57 of `room-handlers.ts`, so empty-string `defaultPath` from callers is correctly treated as absent. New test at line 199 verifies this.

- **`as string` cast removed (P1 from R1):** Replaced with conditional spread `...(defaultPath ? { defaultPath } : {})` on line 63, which omits the key entirely when both sources are falsy. This is cleaner and preserves downstream `?? fallback` chains in `room-repository.ts` (line 32) and `room-handlers.ts` (line 74).

- **Lobby.tsx no longer sends `defaultPath: ''`:** The `as unknown as CreateRoomParams` cast means `defaultPath` is simply absent from the wire payload, which is the correct behavior for the transitional state.

- **Test coverage is adequate:** Three cases covered -- absent defaultPath, empty-string defaultPath, and explicit defaultPath. The `expect.not.objectContaining` assertion correctly verifies that the key is omitted rather than set to `undefined`.

### Remaining observations (P3, non-blocking)

1. **`validateWorkspacePath` is exported but unused in production code** (`packages/shared/src/validation/workspace-path.ts`). It is tested and exported via `mod.ts`, but no caller imports it yet. This is acceptable given the phased Milestone 2 plan documented in the TODOs, but it is technically dead code until then. Knip may flag it.

2. **`as CreateRoomParams` cast on line 64 of `room-handlers.ts`** is a type-level lie (the object may lack `defaultPath`), but this is intentional for the transition period and the downstream code handles `undefined` safely via `?? null`. The TODO on line 52 documents the plan to remove this.

3. **E2E helper hardcodes `/tmp/test-workspace`** in `room-helpers.ts` line 25. The TODO is clear. Just noting it should use `getWorkspaceRoot()` (already imported on line 12) when the milestone lands.

All prior P0 and P1 issues are resolved. The implementation is sound for its stated phased approach.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — Two issues need attention before this merges cleanly.

---

### What changed in this iteration

This pass fixes two issues flagged in prior reviews:

1. `params.defaultPath ?? workspaceRoot` → `params.defaultPath || workspaceRoot` so empty-string placeholders from un-updated callers are treated as absent.
2. `{ ...params, defaultPath: '' }` → `params as unknown as CreateRoomParams` in `Lobby.tsx` so `defaultPath` is absent on the wire (not an empty string).
3. The test assertion `expect.objectContaining({ defaultPath: undefined })` → `expect.not.objectContaining({ defaultPath: expect.anything() })` which correctly catches _any_ value including `undefined`.

The direction is correct. The mechanics work. Two issues remain.

---

### P1 — Test name contradicts test body

**File:** `packages/daemon/tests/unit/rpc/room-handlers.test.ts`, line 199

```ts
it('treats empty-string defaultPath as absent (falls back when workspaceRoot available)', async () => {
    // '' should be treated the same as absent because we use || not ??
    const handler = messageHubData.handlers.get('room.create');

    await handler!({ name: 'Empty Path Room', defaultPath: '' }, {});

    // No workspaceRoot in this test setup so defaultPath is still omitted
    expect(roomManagerData.mocks.createRoom).toHaveBeenCalledWith(
        expect.not.objectContaining({ defaultPath: expect.anything() })
    );
});
```

The test name says "falls back when workspaceRoot available" but no `workspaceRoot` is passed to `setupRoomHandlers` — the comment immediately inside the test body contradicts the name by noting "No workspaceRoot in this test setup". The test only verifies that `''` is treated as absent, not that the `workspaceRoot` fallback fires. The half-promise in the name is misleading to future readers maintaining this code.

The half of the behaviour promised by the test name — that an empty-string `defaultPath` falls back to `workspaceRoot` when one _is_ available — is completely untested. This is the whole reason `||` was chosen over `??`, so it ought to have a passing assertion.

Fix options:
- Rename the test to "treats empty-string defaultPath as absent (no workspaceRoot in this setup, so defaultPath is omitted)".
- Or add a companion test that calls `setupRoomHandlers(..., '/workspace/root')` and asserts `createRoom` is called with `defaultPath: '/workspace/root'` when `defaultPath: ''` is sent.

---

### P2 — `validateWorkspacePath` is exported but never called at the server boundary

**File:** `packages/daemon/src/lib/rpc-handlers/room-handlers.ts`, line 57

```ts
const defaultPath = params.defaultPath || workspaceRoot;

const room = roomManager.createRoom({
    ...
    ...(defaultPath ? { defaultPath } : {}),
} as CreateRoomParams);
```

`validateWorkspacePath` was created as a companion to this PR and exported from `@neokai/shared`, but it is never imported or called in `room-handlers.ts` or `room-manager.ts`. A relative path like `relative/path` sent in `defaultPath` will silently pass through and reach storage. The PR description says "New `validateWorkspacePath(path)` utility … checks non-empty absolute POSIX path" — the implication is that it guards the server boundary — but currently it only has unit tests and no callers.

This is not a blocker on its own if the intent is Milestone 2, but the PR description does not make that clear, and the function name implies it is used in validation today. Either wire the call into the handler (when `defaultPath` is non-empty, validate and throw if invalid) or add a TODO comment making it explicit that the call site is deferred.

---

### Positive observations

- The `|| workspaceRoot` fix is correct and the comment explaining why `||` over `??` is clear.
- The `as unknown as CreateRoomParams` pattern in `Lobby.tsx` is a reasonable temporary bridge — it leaves `defaultPath` absent on the wire rather than sending a known-invalid empty string, and the inline TODO scopes the debt precisely.
- `expect.not.objectContaining({ defaultPath: expect.anything() })` correctly rejects any value including `undefined`/`null`, which the old `objectContaining({ defaultPath: undefined })` did not.
- The `CreateRoomParams` type definition with `defaultPath: string` (required) is correct and the RPC boundary type `Omit<CreateRoomParams, 'defaultPath'> & { defaultPath?: string }` communicates the transitional state clearly.
- Typecheck passes cleanly.
- The E2E helper correctly passes an explicit `/tmp/test-workspace` path instead of relying on the server fallback.

PR #1172, opened 2026-04-01
test(online): add goal lifecycle reset integration tests

lsm:
Adds `packages/daemon/tests/online/room/goal-lifecycle-reset.test.ts` with four integration tests that verify DB-level state changes for the goal lifecycle features added in recent PRs (#1164, #1166, #1169):

- `reset_goal` clears `linkedTaskIds`, `planning_attempts`, `consecutiveFailures`, and restores `active` status
- Description update via `update_goal` resets `planning_attempts` to 0 (invalidation)
- `planning_attempts` is writable via `update_goal`
- Description update transitions `needs_human` → `active`

Tests call `createRoomAgentToolHandlers` directly (same code path as the MCP tools) and assert state via `goal.get` RPC. No real API calls — passes with `NEOKAI_USE_DEV_PROXY=1`.

Also registers the file in `scripts/validate-online-test-matrix.sh`.

### DIFF
--- a/packages/daemon/tests/online/room/goal-lifecycle-reset.test.ts
+++ b/packages/daemon/tests/online/room/goal-lifecycle-reset.test.ts
@@ -1,0 +1,246 @@
+/**
+ * Goal Lifecycle Reset Integration Tests
+ *
+ * Tests DB-level state changes for goal lifecycle operations:
+ * - reset_goal clears all stale state (linkedTaskIds, planning_attempts, consecutiveFailures)
+ * - description update resets planning_attempts (invalidation)
+ * - planning_attempts is writable via update_goal
+ * - description update transitions needs_human → active
+ *
+ * These tests verify RPC-observable state without exercising actual planning sessions.
+ * No real API calls are made — this test is safe to run with NEOKAI_USE_DEV_PROXY=1.
+ *
+ * NOTE: Like all room/* online tests this file is intentionally excluded from the CI
+ * matrix (see .github/workflows/main.yml) due to resource usage. Registered in
+ * scripts/validate-online-test-matrix.sh.
+ * Run locally with: NEOKAI_USE_DEV_PROXY=1 bun test tests/online/room/goal-lifecycle-reset.test.ts
+ */
+
+import { afterAll, beforeAll, describe, expect, test } from 'bun:test';
+import type { DaemonServerContext } from '../../helpers/daemon-server';
+import { createDaemonServer } from '../../helpers/daemon-server';
+import type { DaemonAppContext } from '../../../src/app';
+import type { RoomGoal } from '@neokai/shared';
+import { GoalManager } from '../../../src/lib/room/managers/goal-manager';
+import { TaskManager } from '../../../src/lib/room/managers/task-manager';
+import { SessionGroupRepository } from '../../../src/lib/room/state/session-group-repository';
+import { createRoomAgentToolHandlers } from '../../../src/lib/room/tools/room-agent-tools';
+import { setupGitEnvironment, createRoom, getGoal } from './room-test-helpers';
+
+type InProcessDaemon = DaemonServerContext & { daemonContext: DaemonAppContext };
+
+const savedModel = process.env.DEFAULT_MODEL;
+process.env.DEFAULT_MODEL = 'sonnet';
+
+// ─── Helpers ──────────────────────────────────────────────────────────────────
+
+/**
+ * Create a GoalManager instance backed by the running daemon's SQLite database.
+ * Used to inject stale state for setup — the daemon and this manager share the same
+ * underlying database file, so writes are visible to each other immediately.
+ */
+function makeGoalManager(daemon: InProcessDaemon, roomId: string): GoalManager {
+	const ctx = daemon.daemonContext;
+	return new GoalManager(
+		ctx.db.getDatabase(),
+		roomId,
+		ctx.reactiveDb,
+		ctx.db.getShortIdAllocator()
+	);
+}
+
+/**
+ * Create a set of room agent tool handlers backed by the running daemon's database.
+ * These handlers call the same code paths as the actual MCP tools.
+ */
+function makeAgentToolHandlers(daemon: InProcessDaemon, roomId: string) {
+	const ctx = daemon.daemonContext;
+	const db = ctx.db.getDatabase();
+	const reactiveDb = ctx.reactiveDb;
+	const shortIdAllocator = ctx.db.getShortIdAllocator();
+
+	const goalManager = new GoalManager(db, roomId, reactiveDb, shortIdAllocator);
+	const taskManager = new TaskManager(db, roomId, reactiveDb, shortIdAllocator);
+	const groupRepo = new SessionGroupRepository(db, reactiveDb);
+
+	return createRoomAgentToolHandlers({ roomId, goalManager, taskManager, groupRepo });
+}
+
+async function createGoalViaRpc(
+	daemon: DaemonServerContext,
+	roomId: string,
+	title: string,
+	description: string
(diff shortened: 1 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

The test file is well-structured, correctly exercises the four goal lifecycle operations via the same code paths that production uses (`createRoomAgentToolHandlers`), and asserts state through the RPC layer (`goal.get`) rather than reaching into internals directly. The approach of injecting stale state via a shared-DB `GoalManager` and then invoking tool handlers to reset/transition is a solid integration test pattern.

**Specific observations:**

1. **P2 — Duplicate `createGoalViaRpc` helper.** `createGoalViaRpc` (lines 69-81) is an exact copy of the already-exported `createGoal` in `room-test-helpers.ts` (line 255). The test already imports `createRoom` and `getGoal` from that module. Consider importing `createGoal` instead to avoid the duplication and reduce maintenance burden.

2. **P3 — Test 1 does not verify `replanCount` reset.** `GoalManager.resetGoal()` sets `replanCount: 0` (goal-manager.ts line 203), but test 1 only asserts `linkedTaskIds`, `planning_attempts`, `consecutiveFailures`, and `status`. Adding `expect(resetGoal.replanCount ?? 0).toBe(0)` would improve coverage of the reset path.

3. **P3 — Minor: `savedModel` cleanup could be simplified.** The `afterAll` block manually checks `savedModel !== undefined` to decide between restoring vs deleting. A `afterAll` with `process.env.DEFAULT_MODEL = savedModel!` plus `delete process.env.DEFAULT_MODEL` if undefined would be marginally cleaner, but this is a very minor style nit.

lsm (commented):

Clean, well-structured test file that covers the four stated scenarios. The tests exercise the correct code paths, use shared helpers where appropriate, and assert DB-level state via `goal.get` RPC as intended. A few observations below.

### P1: `InProcessDaemon` type cast is unsafe under `DAEMON_TEST_SPAWN=true`

**File:** `packages/daemon/tests/online/room/goal-lifecycle-reset.test.ts`, line 87

```ts
daemon = (await createDaemonServer()) as InProcessDaemon;
```

The cast from `DaemonServerContext` to `InProcessDaemon` (which requires `daemonContext: DaemonAppContext`) will produce a runtime crash if `DAEMON_TEST_SPAWN=true` is set, because spawned mode does not expose `daemonContext`. The existing pattern in `room-tick-job.test.ts` (line 51-61) guards against this by declaring `daemonContext` as optional and failing explicitly:

```ts
type InProcessDaemon = DaemonServerContext & { daemonContext?: DaemonAppContext };
// ...
const ctx = daemon as InProcessDaemon;
if (!ctx.daemonContext) {
  throw new Error('daemonContext not available — did you run in spawned mode?');
}
return ctx.daemonContext;
```

Since this test file's entire approach requires in-process mode (it needs `daemonContext` to inject state and call handlers directly), a runtime guard with a clear error message is the minimum expectation. Without it, a user running with `DAEMON_TEST_SPAWN=true` would get a confusing `Cannot read properties of undefined (reading 'db')` instead of a helpful message.

### P2: `process.env.DEFAULT_MODEL` mutation at module level is not restored if `beforeAll` fails

**File:** `packages/daemon/tests/online/room/goal-lifecycle-reset.test.ts`, lines 32-33

```ts
const savedModel = process.env.DEFAULT_MODEL;
process.env.DEFAULT_MODEL = 'sonnet';
```

This mutation runs eagerly at import time, but the restoration only happens inside `afterAll`. If `createDaemonServer()` in `beforeAll` throws (e.g., port conflict, Dev Proxy unavailable), the `afterAll` block still runs and restores correctly. However, if the import itself fails or Bun aborts the test file before `afterAll` executes, the env var leak persists. Consider moving the mutation inside `beforeAll` and restoring it in `afterAll` (or using `finally`), matching the established pattern in other tests. The current code works in practice, but it's a subtle footgun.

### P3: Test 1 injects fake task IDs into `linkedTaskIds` -- minor documentation gap

**File:** `packages/daemon/tests/online/room/goal-lifecycle-reset.test.ts`, lines 115-119

The comment says "fake linked task IDs" but doesn't note that these are valid for the `reset_goal` handler because that handler iterates `goal.linkedTaskIds` and calls `taskManager.getTask(taskId)` for each -- which returns `null` for fake IDs and then the `!task` guard on line 1110 of `room-agent-tools.ts` causes the `continue` to skip them. This is correct behavior but worth a brief inline note since a reader might wonder whether fake IDs could cause issues during the reset.

### No issues found (positive observations)

- The four test cases cover the complete `resetGoal()` contract including `replanCount` (added per prior review feedback).
- Using `makeAgentToolHandlers` to call `createRoomAgentToolHandlers` directly is the correct approach -- it exercises the same code path as the MCP tools without needing a real agent session.
- The `parseToolResult` helper properly unwraps the `ToolResult` JSON envelope.
- Reusing `createRoom`, `createGoal`, `getGoal` from `room-test-helpers.ts` avoids duplication.
- The matrix registration in `scripts/validate-online-test-matrix.sh` is correct and alphabetically placed.
- The `afterAll` cleanup properly restores `DEFAULT_MODEL` (or deletes it if originally unset).

lsm (commented):

**Recommendation: APPROVE**

All prior review feedback has been addressed cleanly across two rounds. The final implementation is solid.

### What was fixed (verification)

1. **P1 (InProcessDaemon type safety):** `daemonContext` is now declared as optional on `InProcessDaemon` (line 30), and the `getDaemonCtx()` helper (lines 39-48) fails fast with a clear message when running in spawned mode. This matches the established pattern in `room-tick-job.test.ts` (lines 51-62).

2. **P1 (duplicate createGoal helper):** The local `createGoalViaRpc` has been replaced with the imported `createGoal` from `room-test-helpers.ts` (line 28), which has an identical signature.

3. **P1 (replanCount coverage):** Test 1 now injects `replanCount: 1` (line 141) and asserts `replanCount ?? 0` is `0` after reset (line 166), covering the full `GoalManager.resetGoal()` contract.

4. **P2 (process.env.DEFAULT_MODEL mutation):** The `savedModel` capture and `DEFAULT_MODEL = 'sonnet'` mutation have been moved inside `beforeAll` (lines 103-104), with a comment explaining the rationale (lines 101-102). Restoration in `afterAll` (lines 112-116) is guaranteed even if `beforeAll` throws.

5. **P3 (fake task ID documentation):** An inline comment now explains why fake task IDs are safe during `reset_goal` (lines 133-136): the `!task` guard on unknown IDs causes a `continue`, so no error occurs.

### Remaining observations (no action needed)

- **P3:** The `ROOM_FILES` array in `validate-online-test-matrix.sh` is not strictly alphabetically sorted -- the new entry is appended at the end (line 66). This follows the existing convention of appending to the array rather than maintaining sort order, so it is consistent with the rest of the file.

### Test coverage assessment

The four test cases correctly exercise the stated goal lifecycle operations:

| Test | What it verifies | Code path |
|------|-----------------|-----------|
| 1. reset_goal | Clears linkedTaskIds, planning_attempts, consecutiveFailures, replanCount; restores active status | `createRoomAgentToolHandlers` -> `goalManager.resetGoal()` |
| 2. Description invalidation | planning_attempts resets to 0 on description update | `update_goal` handler lines 248-281 in room-agent-tools.ts |
| 3. planning_attempts writable | Direct write via update_goal persists correctly | `update_goal` handler -> `goalManager.patchGoal()` |
| 4. needs_human recovery | Description update transitions needs_human -> active | `update_goal` handler lines 283-289 in room-agent-tools.ts |

All assertions go through `goal.get` RPC, verifying DB-level state rather than in-memory objects. The approach of injecting stale state via a shared-DB `GoalManager` and then invoking tool handlers is a sound integration test pattern.

PR #1176, opened 2026-04-01
fix: guard room-scoped sessions against missing workspacePath in session-lifecycle

lsm:
## Summary

- `session-lifecycle.ts`: add `ROOM_SCOPED_SESSION_TYPES` guard — `room_chat`, `planner`, `coder`, `leader`, `general`, `space_chat` sessions now **throw** immediately if `workspacePath` is absent, preventing silent fallback to `config.workspaceRoot`
- Non-room sessions (`worker`, `lobby`, `spaces_global`, `neo`, default) continue to fall back to `workspaceRoot` but now emit a `logger.warn`
- 9 new unit tests cover: explicit path used as-is, all 6 room-scoped types throw with a clear message, `worker` type falls back with warning, `undefined` sessionType (defaults to `worker`) falls back

## Code path verification

All existing room-scoped session creation paths already pass `workspacePath` explicitly:
- **`room-handlers.ts`** (`room.create`): sets `workspacePath: defaultPath` (from `CreateRoomParams`, which is now required — Task 2.1)
- **`task-group-manager.ts`** / **`room-runtime.ts`**: `planner`/`coder`/`leader`/`worker` sessions go through `SessionFactory.createAndStartSession()` which calls `sessionFactory.createWorktree()` first and passes the worktree path — `workspacePath` is always set

Guard uses `sessionType` (not `sessionId` prefix) because worker/planner/coder/leader IDs do not start with `room:`.

### DIFF
--- a/packages/daemon/src/lib/session/session-lifecycle.ts
+++ b/packages/daemon/src/lib/session/session-lifecycle.ts
@@ -90,6 +90,36 @@ export class SessionLifecycle {
 		const sessionId = params.sessionId || generateUUID();
 		const sessionType = params.sessionType ?? 'worker';
 
+		// Room-scoped session types must always provide an explicit workspacePath.
+		// These sessions are bound to a specific room/space and must not silently fall back
+		// to the daemon's global workspaceRoot. Guard by sessionType (not sessionId prefix)
+		// because worker/leader/planner/coder IDs don't start with 'room:'.
+		//
+		// Non-room sessions (worker, lobby, spaces_global, neo, and the default 'worker')
+		// fall back to this.config.workspaceRoot with a warning.
+		const ROOM_SCOPED_SESSION_TYPES = [
+			'room_chat',
+			'planner',
+			'coder',
+			'leader',
+			'general',
+			'space_chat',
+		] as const;
+		if (
+			(ROOM_SCOPED_SESSION_TYPES as readonly string[]).includes(sessionType) &&
+			!params.workspacePath
+		) {
+			throw new Error(
+				`Room-scoped session (type: '${sessionType}') must have explicit workspacePath`
+			);
+		}
+
+		// For non-room sessions, fall back to daemon workspaceRoot with a warning.
+		if (!params.workspacePath) {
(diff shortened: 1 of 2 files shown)


lsm (commented):

### Verdict: APPROVE

The implementation correctly addresses the task. Room-scoped session types now fail fast with a clear error message when `workspacePath` is missing, and non-room sessions fall back to `config.workspaceRoot` with a warning log.

### What was done well

- **Correct classification**: The 6 room-scoped types (`room_chat`, `planner`, `coder`, `leader`, `general`, `space_chat`) are exactly the ones that should require an explicit path. Verified against all callers: `room-handlers.ts` passes `workspacePath: defaultPath`, and `task-group-manager.ts` / `lifecycle-hooks.ts` pass worktree paths from `TaskGroupManager.workspacePath`.
- **Non-room types handled correctly**: `neo`, `spaces_global`, `lobby`, and the default `worker` type legitimately fall back to `workspaceRoot` -- confirmed by checking `neo-agent-manager.ts:102` (no `workspacePath`) and `provision-global-agent.ts:113` (no `workspacePath`).
- **The existing test fix** (`room_chat` session creation test now includes `workspacePath: '/room/workspace'`) was necessary and correctly applied.
- **Test coverage is thorough**: 9 new tests covering explicit path passthrough, all 6 room-scoped type rejections, worker fallback with warning verification, and the undefined `sessionType` default path.

**P3-1**: The `ROOM_SCOPED_SESSION_TYPES` constant is declared inside the `create` method body (line 100), so it gets re-allocated on every call. Since this is a pure read-only constant used only for the `includes` check, consider hoisting it to the class level or module scope. This is a trivial optimization but would make the intent clearer that it is immutable metadata, not derived from request state.

**P3-2**: The double cast `(ROOM_SCOPED_SESSION_TYPES as readonly string[]).includes(sessionType)` on line 109 works correctly but is slightly verbose. An alternative would be a `Set` or using `sessionType` in a type guard. This is purely a style preference -- the current code is correct and readable.

PR #1177, opened 2026-04-01
feat: migration 70 — backfill default_path for existing rooms

lsm:
Backfill `default_path` for existing rooms where it is `NULL` (Task 2.4).

**Migration 70** (`migrations.ts`):
- Adds `default_path`/`allowed_paths` columns if missing (handles legacy DBs)
- For rooms with `default_path IS NULL`: sets it from `allowed_paths[0].path`
- Rooms with empty/null `allowed_paths` get sentinel `'__NEEDS_WORKSPACE_PATH__'`
- Idempotent — skips if no rooms have null `default_path`

**Startup hook** (`app.ts`, after `db.initialize`):
- Replaces sentinels with `config.workspaceRoot`
- If `workspaceRoot` is unset, logs an error with affected room IDs

**Tests**: migration-70_test.ts (8 cases) + sentinel-replacement_test.ts (7 cases). All 9877 daemon tests pass.

### DIFF
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -125,6 +125,35 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	await db.initialize(reactiveDb);
 	const liveQueries = new LiveQueryEngine(db.getDatabase(), reactiveDb);
 
+	// Startup hook: replace '__NEEDS_WORKSPACE_PATH__' sentinels left by migration 70.
+	// Migration 70 uses this sentinel for rooms where allowed_paths was also empty/null,
+	// because the migration runs in pure SQL without access to config. Here we have config,
+	// so we replace sentinels with the real workspaceRoot before any room operations begin.
+	{
+		const rawDb = db.getDatabase();
+		const sentinelRows = rawDb
+			.prepare(`SELECT id FROM rooms WHERE default_path = '__NEEDS_WORKSPACE_PATH__'`)
+			.all() as Array<{ id: string }>;
+		if (sentinelRows.length > 0) {
+			if (config.workspaceRoot) {
+				rawDb
+					.prepare(
+						`UPDATE rooms SET default_path = ? WHERE default_path = '__NEEDS_WORKSPACE_PATH__'`
+					)
+					.run(config.workspaceRoot);
+				logInfo(
+					`[Daemon] Replaced sentinel default_path with workspaceRoot for ${sentinelRows.length} room(s)`
+				);
+			} else {
+				const ids = sentinelRows.map((r) => r.id).join(', ');
+				logError(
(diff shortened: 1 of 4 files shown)


lsm (commented):

---

**Recommendation: APPROVE**

The implementation is solid, well-tested, and correctly fulfills the task requirements. I found no blocking issues. Below are a few observations and one P2 suggestion.

### What works well

- **Correct migration pattern**: `runMigration70` follows the established pattern (early-return on missing table, column existence checks via `PRAGMA table_info`, idempotency guard). It is properly wired into `runMigrations()` after migration 69.
- **Two-phase approach**: The sentinel pattern (`__NEEDS_WORKSPACE_PATH__`) is a clean solution to the constraint that migrations run in pure SQL without access to `config`. The startup hook in `app.ts` (line 128-155) correctly replaces sentinels with `config.workspaceRoot` after `db.initialize()` completes.
- **Error handling**: The migration gracefully handles malformed JSON in `allowed_paths`, null values, and missing tables. The startup hook logs affected room IDs when `workspaceRoot` is unset.
- **Test coverage**: 17 tests across two files cover all meaningful cases: backfill from `allowed_paths[0].path`, empty/null `allowed_paths` (sentinel), malformed JSON, idempotency, no-op on missing table, and the full startup hook logic including the undefined-workspaceRoot error path.
- **Data safety**: `createBackup()` at the top of `runMigrations()` (line 23) ensures a backup is created before migration 70 runs on every startup, consistent with all other migrations.

### P2: Sentinel value is not validated at consumption sites

If the startup hook fails to replace sentinels (e.g., `workspaceRoot` is undefined and the error is ignored), the sentinel string `'__NEEDS_WORKSPACE_PATH__'` will be stored as a real `default_path` and could flow into runtime code. At `room-runtime-service.ts:638`, `room.defaultPath` would resolve to the sentinel string (since it is no longer `null`/`undefined`, the `??` fallback to `this.ctx.defaultWorkspacePath` would not trigger). This would pass a nonsense path to the SDK session.

The error log in the startup hook (line 149-152) is appropriate for now since `workspaceRoot` is always set in practice. However, consider adding a validation or guard in a future milestone (e.g., in `room-runtime-service.ts` or the room handler) that rejects or substitutes sentinel values at runtime, as a defense-in-depth measure. This is not blocking for this PR since the sentinel replacement hook runs before any room operations begin.

### P3: Test helper duplicates SQL from app.ts

In `sentinel-replacement_test.ts`, the `runSentinelReplacementHook()` helper re-implements the SQL from `app.ts` inline rather than importing a shared function. This is understandable since the hook is embedded in `createDaemonApp()`, but it means the test could drift out of sync with the production code. A minor maintainability note only — the SQL is simple and stable enough that this is unlikely to be a real problem.

PR #1179, opened 2026-04-01
fix(session-manager): require explicit workspacePath in cleanupOrphanedWorktrees

lsm:
Remove the `config.workspaceRoot` fallback from `cleanupOrphanedWorktrees` and move it to the `worktree.cleanup` RPC boundary.

- `cleanupOrphanedWorktrees(workspacePath: string)` now requires an explicit path (no internal fallback)
- New `getWorkspaceRoot(): string | undefined` getter on `SessionManager` exposes the config value to RPC handlers
- `worktree.cleanup` RPC handler applies fallback at the boundary: uses `sessionManager.getWorkspaceRoot()` when caller omits path; throws `"workspacePath is required when daemon has no default workspace"` when neither is set
- Unit tests updated: removed old no-arg fallback test; added three new RPC handler cases (fallback, explicit path, error)

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/session-handlers.ts
@@ -658,8 +658,13 @@ export function setupSessionHandlers(
 
 	// Handle manual cleanup of orphaned worktrees
 	messageHub.onRequest('worktree.cleanup', async (data) => {
-		const { workspacePath } = data as { workspacePath?: string };
-		const cleanedPaths = await sessionManager.cleanupOrphanedWorktrees(workspacePath);
+		const { workspacePath: payloadPath } = data as { workspacePath?: string };
+		// Apply fallback at the RPC boundary so cleanupOrphanedWorktrees always receives an explicit path.
+		const resolvedPath = payloadPath ?? sessionManager.getWorkspaceRoot();
+		if (!resolvedPath) {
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean, focused change that correctly implements the task. The fallback logic has been properly moved from the domain layer (SessionManager) to the RPC boundary (session-handlers.ts), and the new getWorkspaceRoot() getter provides the right escape hatch for callers that need the config value.

**What looks good:**

- cleanupOrphanedWorktrees(workspacePath: string) is now mandatory with no internal fallback -- the signature change is enforced at compile time.
- The RPC handler at packages/daemon/src/lib/rpc-handlers/session-handlers.ts:661-667 correctly applies the fallback with a clear error message when both payloadPath and workspaceRoot are undefined.
- getWorkspaceRoot() is a minimal, well-documented getter that exposes only what RPC handlers need.
- The three RPC handler test cases (lines 1108-1161) cover: fallback to daemon workspaceRoot, explicit path passthrough, and the error case when neither is set. The assertion at line 1125 verifying the exact argument passed to cleanupOrphanedWorktrees is a nice touch.
- The session-manager.test.ts tests (lines 517-538) properly verify no global fallback occurs inside the method, and the new getWorkspaceRoot test is straightforward.
- All 123 tests pass (47 + 76).
- No other callers of cleanupOrphanedWorktrees exist beyond the RPC handler and tests, so the signature change is safe.

No issues found.

PR #1181, opened 2026-04-01
feat(room-handlers): guard defaultPath changes against active task groups

lsm:
Guard `room.update` defaultPath changes against running workers.

- New `hasActiveTaskGroups(roomId)` method on `RoomRuntimeService` queries the DB via `SessionGroupRepository` (not in-memory map) — works correctly after daemon restarts
- `RoomHandlerOpts` callback approach avoids adding an 8th positional param to `setupRoomHandlers`
- `room.update` now validates new `defaultPath` (absolute + exists), rejects if tasks are active, auto-adds path to `allowedPaths`
- `setupRoomHandlers` moved after `roomRuntimeService` creation in `index.ts` so callback is available
- 6 unit tests: reject with active tasks, allow with no tasks, non-existent/relative path rejection, auto-add allowedPaths, no guard on unchanged path

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -241,6 +241,18 @@ export class RoomRuntimeService {
 		return true;
 	}
 
+	/**
+	 * Returns true if there are any active (non-completed) task groups for the given room.
+	 * Queries the DB directly — does NOT rely on the in-memory runtimes map — so it works
+	 * correctly after a daemon restart when the in-memory map may be empty but DB-persisted
+	 * active groups still exist.
+	 */
+	hasActiveTaskGroups(roomId: string): boolean {
+		const rawDb = this.ctx.db.getDatabase();
+		const groupRepo = new SessionGroupRepository(rawDb, this.ctx.reactiveDb);
+		return groupRepo.getActiveGroups(roomId).length > 0;
+	}
+
 	stop(): void {
 		for (const runtime of this.runtimes.values()) {
 			runtime.stop();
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE** — Clean, correct implementation with good test coverage. No blocking issues found.

### What's done well

- **DB-backed query, not in-memory**: `hasActiveTaskGroups` queries `session_groups` via `SessionGroupRepository.getActiveGroups()`, which correctly handles the daemon-restart scenario described in the docstring. The in-memory `runtimes` map would miss active groups after restart.
- **Optional opts parameter**: Using a `RoomHandlerOpts` object as the 8th parameter avoids adding another positional param to `setupRoomHandlers`. Good pattern for optional cross-cutting concerns.
- **Early return for unchanged path**: The `params.defaultPath !== currentRoom.defaultPath` check at line 173 correctly skips all validation and the guard when the path hasn't actually changed, avoiding unnecessary DB queries.
- **Validation order**: Path format validation and existence checks run before the active-tasks guard. This gives the user actionable feedback about the path first, rather than a confusing "tasks are active" error for a path that would never have worked.
- **Auto-add to allowedPaths**: Automatically merging the new `defaultPath` into `allowedPaths` when absent prevents a common footgun where the user changes the workspace path but forgets to update allowed paths.
- **`setupRoomHandlers` moved after `roomRuntimeService`**: The reorder in `index.ts` is minimal and well-commented. No other handlers depend on `setupRoomHandlers` being called early, so this is safe.
- **Test coverage**: 6 tests covering the key scenarios: reject with active tasks, allow without, invalid path, non-existent path, auto-add allowedPaths, and skip guard on unchanged path.

1. **`SessionGroupRepository` instantiated per call** (`room-runtime-service.ts:252`): A new `SessionGroupRepository` is created every time `hasActiveTaskGroups` is called. This is consistent with the existing pattern in `createOrGetRuntime` (line 659) and `recoverRoomRuntime` (line 962), so it's fine. Just noting it's a lightweight wrapper around a prepared statement, not a heavy object.

2. **Frontend error handling**: The current frontend `room.update` callers (`room-store.ts`) don't pass `defaultPath`, so the new guard errors won't surface in the UI yet. When `defaultPath` editing is added to the UI, the error message `"Cannot change defaultPath while tasks are active. Stop or complete all tasks first."` is clear and actionable. No action needed now, just a forward-looking note.

3. **Test style**: The guard tests use `await import('node:fs')` and `await import('node:os')` dynamically inside test bodies rather than using the top-level imports. This works but is slightly inconsistent with the file-level imports at the top of the test file. Purely cosmetic.

PR #1182, opened 2026-04-01
feat(room): propagate defaultPath changes to room chat session and runtime

lsm:
Implements Task 4.2: when `room.update` successfully changes `defaultPath`, propagate the new path to (a) the room chat session's `workspacePath` and (b) the `RoomRuntime`'s workspace via stop-and-recreate.

## Changes

**`room-handlers.ts`** — After `defaultModel` sync, add `defaultPath` sync: call `sessionManager.updateSession('room:chat:<id>', { workspacePath: newPath })` when the room chat session exists. `updateSession` updates both the DB row and in-memory `AgentSession` metadata — no cache invalidation needed since `resolveSessionContext` re-fetches from DB on cache miss.

**`room-runtime-service.ts`** — In the `room.updated` handler, detect when `room.defaultPath !== runtime.taskGroupManager.workspacePath`. When they differ: call `runtime.stop()`, delete from `this.runtimes`, then `createOrGetRuntime(room)` to build a fresh runtime. `TaskGroupManager.workspacePath` is `readonly` so mutation is not an option. Task 4.1's guard ensures no active task groups exist, so `stop()` does not terminate in-flight agent sessions.

**Note on room chat session SDK subprocess**: The SDK subprocess is NOT interrupted — any in-flight conversation finishes with the old path. The next invocation uses the updated `workspacePath`. This is acceptable because room chat sessions don't fork worktrees.

## Tests

- Room chat `workspacePath` sync: happy path, skipped when session absent, no-op when `defaultPath` not provided
- Runtime stop-and-recreate: verifies old runtime stopped + new runtime has updated `workspacePath`
- No stop/recreate when `defaultPath` unchanged

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -847,6 +847,23 @@ export class RoomRuntimeService {
 				if (runtime) {
 					const room = this.ctx.roomManager.getRoom(event.roomId);
 					if (room) {
+						// When defaultPath changes, the runtime's TaskGroupManager.workspacePath
+						// is readonly and cannot be mutated. Stop the old runtime and recreate it
+						// so the new runtime uses the updated workspacePath for future task groups.
+						// Task 4.1's guard ensures no active task groups exist at this point, so
+						// runtime.stop() does not terminate any in-flight agent sessions.
+						if (room.defaultPath && room.defaultPath !== runtime.taskGroupManager.workspacePath) {
+							log.info(
+								`Room ${room.id} defaultPath changed from ` +
+									`"${runtime.taskGroupManager.workspacePath}" to "${room.defaultPath}" — ` +
+									`stopping and recreating runtime.`
+							);
+							runtime.stop();
+							this.runtimes.delete(event.roomId);
+							this.createOrGetRuntime(room);
+							return;
+						}
+
 						runtime.updateRoom(room);
 						// Re-apply the room chat system prompt so it reflects the latest
 						// room background/instructions (e.g. after room.update changes them).
--- a/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime-service.ts
@@ -860,6 +860,8 @@ export class RoomRuntimeService {
 							);
 							runtime.stop();
 							this.runtimes.delete(event.roomId);
+							this.observers.delete(event.roomId);
+							this.roomAgentMcpServers.delete(event.roomId);
 							this.createOrGetRuntime(room);
 							return;
 						}
--- a/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
@@ -164,6 +164,7 @@ export function setupRoomHandlers(
 
 		// Guard: validate new defaultPath and check for active task groups
 		let updatedAllowedPaths = params.allowedPaths;
+		let defaultPathChanged = false;
 		if (params.defaultPath !== undefined) {
 			const currentRoom = roomManager.getRoom(params.roomId);
 			if (!currentRoom) {
@@ -171,6 +172,7 @@ export function setupRoomHandlers(
 			}
 
 			if (params.defaultPath !== currentRoom.defaultPath) {
+				defaultPathChanged = true;
 				// Validate the new path (only the incoming path — current path may be a sentinel value)
 				const pathValidation = validateWorkspacePath(params.defaultPath);
 				if (!pathValidation.valid) {
@@ -238,7 +240,7 @@ export function setupRoomHandlers(
 		// both the DB and the in-memory AgentSession metadata via updateSession(), so
 		// resolveSessionContext (which re-fetches from DB on cache miss) is always
 		// consistent — no additional cache invalidation is needed.
-		if (params.defaultPath !== undefined && room.defaultPath && sessionManager) {
+		if (defaultPathChanged && room.defaultPath && sessionManager) {
 			const roomChatSessionId = `room:chat:${room.id}`;
 			try {
 				const existingSession = sessionManager.getSessionFromDB(roomChatSessionId);
--- a/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/room-handlers.ts
@@ -233,6 +233,25 @@ export function setupRoomHandlers(
 			}
 		}
 
+		// When defaultPath changes, sync the room chat session's workspacePath so the
+		// next SDK invocation uses the updated path. The session record is updated in
+		// both the DB and the in-memory AgentSession metadata via updateSession(), so
+		// resolveSessionContext (which re-fetches from DB on cache miss) is always
+		// consistent — no additional cache invalidation is needed.
+		if (params.defaultPath !== undefined && room.defaultPath && sessionManager) {
+			const roomChatSessionId = `room:chat:${room.id}`;
+			try {
+				const existingSession = sessionManager.getSessionFromDB(roomChatSessionId);
+				if (existingSession) {
+					await sessionManager.updateSession(roomChatSessionId, {
+						workspacePath: room.defaultPath,
+					});
+				}
+			} catch (err) {
+				log.warn(`Could not sync room chat session workspacePath for room ${room.id}:`, err);
+			}
+		}
+
 		// Broadcast room update event
 		daemonHub
 			.emit('room.updated', {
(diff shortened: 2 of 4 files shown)


lsm (commented):

---

**Recommendation: APPROVE** -- The implementation is correct and well-tested. The minor issues below are non-blocking.

### P2: Missing observer/roomAgentMcpServers cleanup in room.updated handler

**File:** `packages/daemon/src/lib/room/runtime/room-runtime-service.ts`, lines 861-864

The inline stop-and-recreate logic in the `room.updated` event handler calls `runtime.stop()` and `this.runtimes.delete(event.roomId)`, but does not call `this.observers.delete(event.roomId)` or `this.roomAgentMcpServers.delete(event.roomId)`. Compare with `stopRuntime()` (line 206-215) and `startRuntime()` (line 221-242), both of which clean up all three maps explicitly.

In practice this is not a bug because `createOrGetRuntime()` overwrites all three maps (lines 714-716 + `setupRoomAgentSession` at line 746), and `runtime.stop()` disposes the observer. But the inconsistency makes the code harder to reason about and could become a real leak if `createOrGetRuntime` ever gains an early-return path that does not overwrite the maps.

Suggested fix: add `this.observers.delete(event.roomId)` and `this.roomAgentMcpServers.delete(event.roomId)` after `this.runtimes.delete(event.roomId)` for consistency with the other stop paths.

### P3: Unnecessary workspacePath sync when defaultPath is unchanged

**File:** `packages/daemon/src/lib/rpc-handlers/room-handlers.ts`, line 241

The condition `params.defaultPath !== undefined && room.defaultPath` runs the workspacePath sync even when the path has not actually changed (e.g., a client sends `defaultPath: "/same/path"`). The guard at line 173 already skips validation in this case, but the sync block runs unconditionally. Consider checking `params.defaultPath !== currentRoom.defaultPath` (or comparing with the old room's defaultPath) to skip the unnecessary `updateSession` call. This is a minor inefficiency since the old and new values would be identical.

### Summary

The two-part approach (sync session workspacePath in handler + stop-and-recreate runtime on event) is clean and well-separated. The guard in `room-handlers.ts` ensuring no active task groups exist before allowing a path change makes the stop-and-recreate in the runtime service safe. Tests are thorough, covering both the happy path and edge cases (unchanged path, missing session, no defaultPath param). No P0 or P1 issues found.

lsm (commented):

### Summary

The implementation correctly propagates `defaultPath` changes to both the room chat session's `workspacePath` and the runtime's workspace. The approach of stopping and recreating the runtime is sound, and the tests provide good coverage of the happy path and edge cases.

### Issue Found

#### P1: SessionObserver resource leak in `room.updated` handler

**File:** `/packages/daemon/src/lib/room/runtime/room-runtime-service.ts`
**Lines:** 855-865

When the `defaultPath` changes and the runtime is stopped and recreated, the `SessionObserver` stored in `this.observers` map is not properly cleaned up. The observer's `dispose()` method is never called before the map entry is overwritten.

**Current code:**
```typescript
if (room.defaultPath && room.defaultPath !== runtime.taskGroupManager.workspacePath) {
    log.info(
        `Room ${room.id} defaultPath changed from ` +
            `"${runtime.taskGroupManager.workspacePath}" to "${room.defaultPath}" — ` +
            `stopping and recreating runtime.`
    );
    runtime.stop();
    this.runtimes.delete(event.roomId);
    this.createOrGetRuntime(room);
    return;
}
```

**Problem:** `runtime.stop()` disposes the `RoomRuntime`'s internal observer (the one passed to its constructor), but the `RoomRuntimeService` also maintains a separate `this.observers` map. The entry in this map is never deleted, and the old observer's `dispose()` method is never called. This leaves DaemonHub subscriptions dangling.

**Comparison with `stopRuntime()`:**
The `stopRuntime()` method (lines 206-214) correctly cleans up all three maps:
```typescript
stopRuntime(roomId: string): boolean {
    const runtime = this.runtimes.get(roomId);
    if (!runtime) return false;
    runtime.stop();
    this.runtimes.delete(roomId);
    this.observers.delete(roomId);  // <-- This IS called
    this.roomAgentMcpServers.delete(roomId);
    ...
}
```

**Recommended fix:** Add `this.observers.delete(event.roomId);` before `createOrGetRuntime()`:
```typescript
if (room.defaultPath && room.defaultPath !== runtime.taskGroupManager.workspacePath) {
    log.info(...);
    runtime.stop();
    this.runtimes.delete(event.roomId);
    this.observers.delete(event.roomId);  // <-- Add this line
    this.createOrGetRuntime(room);
    return;
}
```

For consistency, you may also want to add `this.roomAgentMcpServers.delete(event.roomId);` (though this is less critical since `setupRoomAgentSession` overwrites the entry anyway).

### Test Coverage Gap

The tests for the stop-and-recreate behavior don't verify that the `observers` map is properly cleaned up. Consider adding an assertion like:
```typescript
const serviceAny = service as unknown as { 
    runtimes: Map<string, RoomRuntime>;
    observers: Map<string, SessionObserver>;
};
// After room.updated with new defaultPath
const oldObserver = ...; // capture before update
expect(serviceAny.observers.get('room-1')).not.toBe(oldObserver);
```

### Minor Observations

- The comment at lines 851-854 correctly explains why the stop-and-recreate approach is safe (Task 4.1's guard ensures no active task groups)
- The `room-handlers.ts` implementation for syncing `workspacePath` is correct and follows the same pattern as the `defaultModel` sync
- The test coverage for the RPC handler changes is thorough

### Recommendation

**REQUEST_CHANGES** — The `SessionObserver` resource leak should be fixed before merging. While the impact is limited (only affects rooms where `defaultPath` is changed), it's a clear bug that could lead to memory leaks and dangling event subscriptions over time.

lsm (commented):

### Summary

The three fixes from the previous review feedback are all correctly implemented and the code is clean. All 67 tests pass across both test files.

### Verification of Fixes

**P1 fix (observer/MCP server cleanup)**: The stop-and-recreate path at `room-runtime-service.ts:861-864` now deletes `this.observers` and `this.roomAgentMcpServers` before calling `createOrGetRuntime()`. This is consistent with both `stopRuntime()` (lines 210-212) and `startRuntime()` (lines 231-235). No leak of stale `SessionObserver` DaemonHub subscriptions or stale MCP server configs. The test at line ~870 asserts `newObserver !== oldObserver`, confirming the old observer is evicted.

**P2/P3 fix (defaultPathChanged flag)**: The `defaultPathChanged` boolean in `room-handlers.ts:167/174-175` is set only when `params.defaultPath !== currentRoom.defaultPath`. The workspacePath sync at line 243 now uses `defaultPathChanged` instead of `params.defaultPath !== undefined`, which correctly avoids a no-op `updateSession` call when the client sends the same path already stored. The test "does not sync workspacePath when defaultPath is unchanged" at line 678 verifies this.

### Observations (no action required)

- The `defaultModel` sync at line 219 uses `params.defaultModel && sessionManager` (fires even if model unchanged), while the new `defaultPath` sync is more precise. This is a pre-existing asymmetry and not introduced by this PR.
- `daemonHub.emit()` is fire-and-forget (Promise not awaited), and `dispatchLocally` uses `queueMicrotask`. The runtime stop-and-recreate happens asynchronously after the RPC response is sent. This is consistent with the existing pattern throughout the codebase and is not a concern here since the room is already persisted.
- The `agentSessions` map is keyed by session ID (not room ID), so it doesn't need per-room cleanup in this path -- consistent with `stopRuntime()` which also doesn't touch it.

### Verdict

All previous review concerns are addressed. The implementation is correct, complete, and follows established patterns. Approving.

lsm (commented):

### Summary

All reviewer feedback has been correctly addressed. The implementation is complete and well-tested.

### P1 Fix: SessionObserver cleanup on runtime recreation

**File:** `packages/daemon/src/lib/room/runtime/room-runtime-service.ts` (lines 861-864)

The fix correctly adds `this.observers.delete(event.roomId)` and `this.roomAgentMcpServers.delete(event.roomId)` before calling `createOrGetRuntime()` in the stop-and-recreate path. This prevents:
- Stale SessionObserver entries from dangling in the map with old DaemonHub subscriptions
- Stale room-agent-tools MCP server configs from being reused

This matches the existing cleanup pattern in `stopRuntime()` (lines 206-214) and `startRuntime()` (lines 226-236).

**Test verification:** The test now captures `oldObserver` before the update and asserts `newObserver` is a different instance after recreation.

### P2/P3 Fix: Avoid unnecessary workspacePath sync

**File:** `packages/daemon/src/lib/rpc-handlers/room-handlers.ts` (lines 167, 175, 243)

The `defaultPathChanged` boolean flag is correctly introduced:
- Set to `true` only when `params.defaultPath !== currentRoom.defaultPath`
- Used as the condition for the workspacePath sync block instead of `params.defaultPath !== undefined`

This avoids a no-op `updateSession` call when a client sends the same path that is already stored.

**Test verification:** New test case "does not sync workspacePath when defaultPath is unchanged" verifies that `updateSessionMock` is not called when the same path is sent.

### Tests

All 67 tests pass across both test files. The test coverage is comprehensive:
- Observer map cleanup verification
- No unnecessary sync when path unchanged
- Happy path for both runtime recreation and session sync
- Edge cases (session absent, path not provided)

### Verdict

**APPROVE** - All fixes are correctly implemented and tested.

PR #1184, opened 2026-04-01
feat: make Config.workspaceRoot optional (Task 5.1)

lsm:
Make `Config.workspaceRoot` optional so the daemon can start without `--workspace`.

## Changes

- `Config.workspaceRoot` changed to `string | undefined`; `getConfig()` returns `undefined` instead of throwing when no workspace is set
- Default DB path falls back to `~/.neokai/data/daemon.db` when workspaceRoot is undefined (no collision with workspace-derived paths)
- `SystemState.workspaceRoot` made optional so the frontend can hide the field when absent
- `FileIndex`: accepts `undefined` workspacePath; `init()` is a no-op and searches return empty results
- `SettingsManager` constructed with `config.workspaceRoot ?? homedir()` fallback (reads global `~/.claude/.mcp.json` for MCP config when no workspace is set)
- `SessionLifecycle`: guard for `baseWorkspacePath === undefined` skips `detectGitSupport` and worktree creation for no-workspace sessions (neo, spaces_global)
- `NeoToolsConfig.workspaceRoot` and `ReferenceHandlerDeps.workspaceRoot` made optional
- CLI entry points log `(none)` gracefully when workspaceRoot is undefined
- Config tests updated: "throws error" test replaced with two tests for undefined workspaceRoot and fallback DB path

### DIFF
--- a/packages/cli/main.ts
+++ b/packages/cli/main.ts
@@ -50,7 +50,7 @@ const serverMode = isDev ? 'Development' : isTest ? 'Test' : 'Production';
 console.log(`\n🚀 NeoKai ${serverMode} Server`);
 console.log(`   Mode: ${config.nodeEnv}`);
 console.log(`   Model: ${config.defaultModel}`);
-console.log(`   Workspace: ${config.workspaceRoot}\n`);
+console.log(`   Workspace: ${config.workspaceRoot ?? '(none)'}\n`);
 
 if (isDev) {
 	// Development mode: Vite dev server + Daemon (for local development with HMR)
--- a/packages/cli/prod-entry.ts
+++ b/packages/cli/prod-entry.ts
@@ -60,7 +60,7 @@ if (!cliOptions.workspace && !process.env.NEOKAI_WORKSPACE_PATH) {
 const config = getConfig(cliOptions);
 
 console.log(`\nNeoKai Server`);
-console.log(`   Workspace: ${config.workspaceRoot}\n`);
+console.log(`   Workspace: ${config.workspaceRoot ?? '(none)'}\n`);
 
 try {
 	await startProdServer(config);
--- a/packages/daemon/src/app.ts
+++ b/packages/daemon/src/app.ts
@@ -1,4 +1,5 @@
 import type { Server } from 'bun';
+import { homedir } from 'os';
 import type { Config } from './config';
 import type { WebSocketData } from './types/websocket';
 import { Database } from './storage/database';
@@ -194,8 +195,13 @@ export async function createDaemonApp(options: CreateDaemonAppOptions): Promise<
 	const authManager = new AuthManager(db, config);
 	await authManager.initialize();
 
-	// Initialize settings manager
-	const settingsManager = new SettingsManager(db, config.workspaceRoot);
+	// Initialize settings manager.
+	// When no workspace is set, fall back to homedir() so SettingsManager reads global
+	// MCP config from ~/.claude/.mcp.json. This means MCP servers configured in a
+	// project-level .mcp.json won't be discovered for the global instance — but that
+	// is acceptable because room-scoped sessions use their own defaultPath for MCP
+	// resolution, not the global settings manager workspace path.
+	const settingsManager = new SettingsManager(db, config.workspaceRoot ?? homedir());
 
 	// Check authentication status
 	const authStatus = await authManager.getAuthStatus();
--- a/packages/daemon/src/config.ts
+++ b/packages/daemon/src/config.ts
@@ -42,7 +42,7 @@ export interface Config {
 	temperature: number;
 	maxSessions: number;
 	nodeEnv: string;
-	workspaceRoot: string;
+	workspaceRoot?: string;
 	disableWorktrees?: boolean; // For testing - disables git worktree creation
 	// GitHub integration
 	githubWebhookSecret?: string; // Secret for verifying webhook signatures
@@ -63,8 +63,8 @@ export function getConfig(overrides?: ConfigOverrides): Config {
 	// Workspace root priority:
 	// 1. CLI --workspace flag (overrides parameter)
 	// 2. NEOKAI_WORKSPACE_PATH environment variable
-	// Note: No default fallback - caller (CLI package) must provide workspace path
-	let workspaceRoot: string;
+	// 3. undefined — daemon can start without a workspace (e.g. headless/API-only mode)
+	let workspaceRoot: string | undefined;
 	if (overrides?.workspace) {
 		// CLI override has highest priority
 		workspaceRoot = overrides.workspace;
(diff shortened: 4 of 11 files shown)


lsm (commented):

### Summary

This PR makes `Config.workspaceRoot` optional and updates all downstream consumers to handle `undefined` gracefully. The goal is to allow the daemon to start without a `--workspace` flag, supporting a "headless/API-only" mode where rooms provide their own workspace paths. The implementation is well-structured and the diff is clean. Typecheck passes, all 27 config tests pass. I found one P1 issue and a few P2/P3 observations.

---

### P1 -- `process.cwd()` fallback in `session-lifecycle.ts` is fragile and potentially surprising

**File:** `packages/daemon/src/lib/session/session-lifecycle.ts`, line 171

```ts
let sessionWorkspacePath: string = baseWorkspacePath ?? process.cwd();
```

When the daemon starts without `--workspace` and a non-room session (e.g., `neo` or `spaces_global`) is created without an explicit `workspacePath`, this falls back to `process.cwd()`. That means `Session.workspacePath` silently becomes whatever directory the daemon was launched from. Downstream consumers may then assume this is a real, intentional workspace and perform file operations against it.

The comment says "Global sessions (neo, spaces_global) don't use the workspace path for file operations," but that is a behavioral assumption that could break if those sessions gain new capabilities in the future. A safer approach would be to store `undefined` in the session record and let consumers that actually need a workspace path handle the absence explicitly. Alternatively, if the `Session` type requires `workspacePath: string`, this should be documented as a known limitation with a `TODO` to make `Session.workspacePath` optional as part of the larger workspace decoupling effort.

---

### P2 -- `FileIndex` internal methods use `workspacePath!` non-null assertions without a guard

**File:** `packages/daemon/src/lib/file-index.ts`, lines 225, 252, 254, 300, 302, 346, 381

The `init()` method correctly guards against `undefined` workspacePath. However, the private methods `loadGitignore()`, `scanDirectory()`, `refreshDirectory()`, and `runRefresh()` all use `this.workspacePath!` with non-null assertions. These methods are only ever called from `init()` (which has the guard), so there is no actual crash risk today. But the pattern is fragile -- if someone adds a code path that calls `runRefresh()` directly (e.g., to force a re-scan), it would crash at runtime with a confusing "Cannot read properties of undefined" error.

A single guard at the top of `runRefresh()` (or making `scanDirectory`/`refreshDirectory` accept a path parameter) would make this more robust.

---

### P2 -- No integration test for the actual "daemon starts without --workspace" path

The unit tests verify `getConfig()` returns `undefined` workspaceRoot and the correct fallback DB path, which is good. However, there is no test that verifies the full daemon startup path (`createDaemonApp`) succeeds when `config.workspaceRoot` is `undefined`. This is the core behavioral claim of the PR. A smoke test that creates a `DaemonApp` with `workspaceRoot: undefined` and asserts it starts without error (including SettingsManager initialization, FileIndex no-op, sentinel replacement warning, etc.) would provide much stronger confidence.

---

### P3 -- Duplicate `FileIndex` construction

**File:** `packages/daemon/src/app.ts` line 328 and `packages/daemon/src/lib/rpc-handlers/index.ts` line 296

A `FileIndex` is created in `createDaemonApp()` at line 328 and a second one is created in `setupRPCHandlers()` at line 296. Both are constructed with `deps.config.workspaceRoot`. This pre-dates this PR, but it means the daemon now maintains two independent file indices. The one in `app.ts` is used for cleanup only (`dispose()`), while the one in `rpc-handlers/index.ts` is the one that actually serves `reference.search`. This is not introduced by this PR, but worth noting.

---

### P3 -- Removed unused imports in test file

The PR removes imports for `spyOn`, `logCredentialDiscovery`, and `DiscoveryResult` from the test file. These were already unused on `dev`, so the cleanup is correct but incidental to the feature change.

---

### Verdict

The core change is sound and well-documented. The P1 issue (process.cwd() fallback) should be addressed before merge, as it silently assigns a potentially misleading workspace path to sessions. The P2 items are important for long-term maintainability but are not blockers for the current behavior.

lsm (commented):

## Summary

This PR correctly implements Task 5.1 of the workspace decoupling milestone. The changes make `Config.workspaceRoot` optional and allow the daemon to start without `--workspace`. The implementation is thorough with appropriate fallbacks and guards.

## Key Changes Verified

**Config changes (packages/daemon/src/config.ts)**
- `workspaceRoot` changed from required `string` to optional `string | undefined`
- `getConfig()` now returns `undefined` instead of throwing when no workspace is provided
- DB path falls back to `~/.neokai/data/daemon.db` (no collision with workspace-derived paths)

**DaemonAppContext (packages/daemon/src/app.ts)**
- SettingsManager uses `config.workspaceRoot ?? homedir()` fallback to read global MCP config from `~/.claude/.mcp.json`
- Sentinel replacement for migration 70 warns (non-fatal) when workspaceRoot is undefined

**SessionLifecycle (packages/daemon/src/lib/session/session-lifecycle.ts)**
- Guard added for `baseWorkspacePath === undefined` - skips git detection and worktree creation
- Non-room sessions (neo, spaces_global) fall back to `process.cwd()` for `sessionWorkspacePath`
- Room-scoped sessions still require explicit `workspacePath` (correct behavior)

**FileIndex (packages/daemon/src/lib/file-index.ts)**
- Accepts `undefined workspacePath` - `init()` is a no-op when no workspace
- Search returns empty results gracefully

**Reference handlers (packages/daemon/src/lib/rpc-handlers/reference-handlers.ts)**
- `workspaceRoot` made optional in deps
- File/folder resolution returns `{ resolved: null }` when no workspacePath available

**Shared types (packages/shared/src/state-types.ts)**
- `SystemState.workspaceRoot` made optional with clear comment

**Tests**
- Config tests updated: "throws error" test replaced with two tests for undefined workspaceRoot and fallback DB path
- All 27 config tests pass, typecheck passes, lint passes

## P2: Suggestions for Consideration

1. **FileIndex non-null assertions**: The file-index.ts uses `this.workspacePath!` in several places after the guard. While safe due to the guard, these could be refactored for cleaner code by extracting a private method that asserts non-void before operations.

2. **SessionLifecycle process.cwd() fallback** (line 171): The fallback `baseWorkspacePath ?? process.cwd()` means sessions without workspace will have `sessionWorkspacePath` set to the current working directory. This is safe for global sessions (neo, spaces_global) that don't use file operations, but could be confusing if logged. Consider documenting this behavior or using a more explicit sentinel value.

## P3: Minor Nits

1. **neo-action-tools.ts unused variable**: Line 532 assigns `workspaceRoot: _workspaceRoot` with an underscore prefix indicating intentionally unused. This is correct style but could have a brief comment explaining why it's unused (workspace context is derived per-room, not globally).

## Recommendation

**APPROVE** - The implementation is correct, complete, and well-tested. The fallbacks are sensible, the guards are in the right places, and the type changes are propagated throughout the codebase. This is a solid foundation for Task 5.2.

The P2 items are suggestions for code clarity and do not affect correctness.

lsm (commented):

### Re-Review Summary

All three previously-raised issues have been adequately addressed:

**P1 (GLM) - Resolved**: The `process.cwd()` fallback in `session-lifecycle.ts` (lines 168-174) now has a clear TODO comment documenting:
- Why the fallback exists (to ensure `Session.workspacePath` always has a value)
- That global sessions don't use the workspace path for file operations
- The proper fix is to make `Session.workspacePath` optional in a future refactor

**P2 (GLM) - Resolved**: `FileIndex.runRefresh()` now has an explicit guard at the top (line 344) that returns early when `workspacePath === undefined`. This mirrors the `init()` guard and prevents non-null assertions from firing if `runRefresh()` is ever called via an unexpected code path.

**P2 (GLM) - Resolved**: Two integration tests added in `daemon-app-cleanup.test.ts` (lines 280-356):
1. Verifies daemon starts successfully with `workspaceRoot: undefined` and that `FileIndex.isReady()` returns false
2. Verifies the sentinel warning is logged when rooms have `__NEEDS_WORKSPACE_PATH__` and workspaceRoot is undefined

### Additional Observations

The early returns for file/folder reference resolution (lines 171, 175 in `reference-handlers.ts`) are correctly placed and provide graceful degradation when no workspace is available.

### Recommendation: APPROVE

All feedback has been addressed with appropriate documentation, guards, and test coverage.

lsm (commented):

### Re-review Summary

**P1 (TODO comment on process.cwd() fallback)** -- Resolved. The comment at `packages/daemon/src/lib/session/session-lifecycle.ts:172-173` now clearly documents why the fallback exists and references a TODO for making `Session.workspacePath` optional in a future refactor. This provides adequate traceability for the tech debt.

**P2 (FileIndex.runRefresh() guard)** -- Resolved. An explicit early return was added at `packages/daemon/src/lib/file-index.ts:344` that mirrors the `init()` guard, preventing the non-null assertions on `workspacePath` from firing if `runRefresh()` is ever called before `init()` or via an unexpected code path.

**P2 (Integration tests for no-workspace startup)** -- Resolved. Two new tests were added in `packages/daemon/tests/unit/core/daemon-app-cleanup.test.ts` under the `daemon startup without --workspace (workspaceRoot: undefined)` describe block:
1. Verifies `createDaemonApp` succeeds with `workspaceRoot: undefined`, all core components are initialized, and `fileIndex.isReady()` returns `false`.
2. Verifies the sentinel warning is logged when rooms have `__NEEDS_WORKSPACE_PATH__` and `workspaceRoot` is undefined, using a real file-based DB to seed the sentinel row.

Both new tests pass. The pre-existing test failures in the same file (timing-sensitive cleanup tests) are unrelated to this change.

### Recommendation: APPROVE

PR #1186, opened 2026-04-01
feat: add directional arrow indicator to gate badges on channel edges

lsm:
Adds a small directional indicator to gate badges on channel edges in the visual workflow editor:

- **One-way channels**: renders a rotated SVG triangle (▸) before the label, pointing in the path direction (computed from segment angle: 0°/90°/180°/270°)
- **Bidirectional channels**: prepends `⇄ ` to the label text

Badge widths are widened to accommodate the indicator. A new exported helper `getOrthogonalPathMidpointWithAngle()` extends the existing midpoint calculation to also return the direction angle of the segment at the midpoint.

New tests: 7 unit tests for `getOrthogonalPathMidpointWithAngle` (horizontal, vertical, L-path, multi-segment) and 4 badge rendering tests (arrow present/absent, bidirectional label prefix).

### DIFF
--- a/packages/web/src/components/space/visual-editor/EdgeRenderer.tsx
+++ b/packages/web/src/components/space/visual-editor/EdgeRenderer.tsx
@@ -75,10 +75,16 @@ export interface ResolvedWorkflowChannel {
 	direction: 'one-way' | 'bidirectional';
 	isCyclic?: boolean;
 	/**
-	 * Gate condition type -- when present (and not 'always'), the channel has a gate.
-	 * Gated channels render as solid lines; ungated channels render as dashed lines.
+	 * Gate condition type for the forward direction (from→to).
+	 * For one-way channels this is the only gate.
+	 * For bidirectional channels this is the gate on the fromStepId→toStepId direction.
 	 */
 	gateType?: 'human' | 'condition' | 'task_result' | 'check' | 'count';
+	/**
+	 * Gate condition type for the reverse direction (to→from).
+	 * Only set on bidirectional channels where the reverse direction has its own gate.
+	 */
+	reverseGateType?: 'human' | 'condition' | 'task_result' | 'check' | 'count';
 	/** Stable ID for selection -- typically the workflow-level channel array index as a string. */
 	id?: string;
 	/** Optional display label from WorkflowChannel.label */
@@ -710,20 +716,23 @@ export function EdgeRenderer({
 				const visiblePoints = getVisibleChannelPathPoints(channel, pts);
 				const visibleD = roundedOrthogonalPath(visiblePoints);
 				const isBidirectional = channel.direction === 'bidirectional';
-				const isGated = !!channel.gateType;
+				// A channel is gated if either the forward or reverse direction has a gate.
+				const isGated = !!(channel.gateType || channel.reverseGateType);
 				const isCyclic = !!channel.isCyclic;
 				const isSelected = channel.id != null && channel.id === selectedChannelId;
+				// A bidirectional gate badge has arrows on both sides when both directions
+				// are gated. When only one direction is gated on a bidirectional channel the
+				// badge shows a single directional arrow (just like a one-way channel).
+				const hasBothDirectionGates =
+					isBidirectional && !!channel.gateType && !!channel.reverseGateType;
 
 				const strokeColor = isSelected ? 'white' : CHANNEL_EDGE_COLOR;
 				const strokeWidth = isSelected ? CHANNEL_SELECTED_STROKE_WIDTH : CHANNEL_STROKE_WIDTH;
 				const strokeDasharray = isBidirectional ? undefined : CHANNEL_EDGE_DASH_ARRAY;
 				const strokeOpacity = isSelected ? 1 : 0.85;
-				// One-way gates compute the path direction angle for the arrow indicator.
-				// Bidirectional gates only need position (no angle required).
+				// Always compute midpoint + angle — needed for all badge variants.
 				const gateBadgeMidpoint = isGated
-					? !isBidirectional
-						? getOrthogonalPathMidpointWithAngle(visiblePoints)
-						: { ...getOrthogonalPathMidpoint(visiblePoints), angle: 0 }
+					? getOrthogonalPathMidpointWithAngle(visiblePoints)
 					: null;
 				const gateBadgePosition = gateBadgeMidpoint;
 				const loopBadgePosition = isCyclic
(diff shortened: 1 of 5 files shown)


lsm (commented):

---

### P0 -- Arrow mispositioned when path direction is vertical (90/270)

`packages/web/src/components/space/visual-editor/EdgeRenderer.tsx`, line 878

The polygon transform `translate(offsetX, 0) rotate(angle)` applies the translation in the **pre-rotation** coordinate system. When the angle is 0 or 180 (horizontal paths), this works correctly because the translation is along the horizontal axis, which aligns with the badge layout. But for angles 90 and 270 (vertical paths), the polygon is first shifted ~19.5px to the left (the `offsetX` value), then the entire coordinate system rotates around that shifted origin, causing the arrow to orbit away from the badge center.

For example, with a "Check" label (5 chars) and angle=90:
- The polygon is translated to x=-19.5, y=0
- Then rotated 90deg around (-19.5, 0)
- The result: the arrow ends up at approximately (-24.5, 4), (-19.5, -4), (-14.5, 4) -- centered at x=-19.5 instead of x=0

The correct approach for vertical paths is either:
- **Option A**: Rotate first, then translate along the rotated axis: `rotate(angle) translate(offsetX, 0)` -- this keeps the arrow offset along the path direction
- **Option B**: Separate the arrow and text positioning based on the angle. For horizontal paths (0/180), lay out [arrow | gap | text] horizontally. For vertical paths (90/270), lay out [arrow / gap / text] vertically with appropriate y-offsets.

Note: per `semanticWorkflowGraph.ts` lines 248-261, vertical paths (bottom/top source/target sides) are a common case, occurring whenever `|dy| > |dx| * 0.85`.

### P1 -- Significant code duplication between `getOrthogonalPathMidpoint` and `getOrthogonalPathMidpointWithAngle`

`packages/web/src/components/space/visual-editor/EdgeRenderer.tsx`, lines 324-421

`getOrthogonalPathMidpointWithAngle` (lines 376-421) duplicates the entire algorithm of `getOrthogonalPathMidpoint` (lines 324-363) with only the angle computation added. These two functions share the normalization, total length calculation, segment traversal, and midpoint interpolation logic identically.

Suggested fix: have `getOrthogonalPathMidpointWithAngle` call `getOrthogonalPathMidpoint` internally (or vice versa, refactoring the original to always compute angle), or extract the shared logic into a private helper. This eliminates the risk of the two functions diverging if either is modified in the future.

### P2 -- `data-gate-angle` attribute serves no runtime purpose

`packages/web/src/components/space/visual-editor/EdgeRenderer.tsx`, line 845

The `data-gate-angle` attribute on the `<g>` element is set to the angle value but is never read anywhere in the codebase (only the polygon's inline `transform` attribute uses the angle). If this is purely for debugging/inspection, that's fine, but the comment on line 874 says "rotated to match the path direction via the data-gate-angle attribute" which is misleading -- the rotation is done via the polygon's own `transform`, not via `data-gate-angle`. Consider either removing the misleading comment or removing the unused data attribute.

### P2 -- `+ 2` magic number in badge width calculation

`packages/web/src/components/space/visual-editor/EdgeRenderer.tsx`, line 782

```ts
(!isBidirectional && isGated ? CHANNEL_GATE_ARROW_TOTAL + 2 : 0);
```

The `+ 2` is unexplained. If it's visual padding, it should be named as a constant (e.g. `CHANNEL_GATE_ARROW_EXTRA_PADDING = 2`) with a brief comment explaining its purpose, consistent with the other named constants in this section.

### P3 -- PR description test count mismatch

The PR description says "11 new unit tests" but the actual additions are 7 tests for `getOrthogonalPathMidpointWithAngle` + 4 badge rendering tests = 11. The body text says "7 unit tests for getOrthogonalPathMidpointWithAngle ... and 4 badge rendering tests" which is correct. No issue here, just confirming the count matches.

lsm (commented):

### Summary

The implementation correctly adds directional arrow indicators to gate badges on channel edges. The arrow rotation logic is sound, and the test coverage is comprehensive with 11 new unit tests covering all cardinal directions and edge cases.

### What Works Well

- **Clean API design**: The new `getOrthogonalPathMidpointWithAngle()` helper extends the existing midpoint calculation with angle information in a natural way.
- **Good inline documentation**: The layout comments explain the math clearly (e.g., `// Layout: [arrow(8px) | gap(4px) | text] all centered at x=0`).
- **Proper SVG transform order**: The `translate()` then `rotate()` ordering is correct for SVG transforms.
- **Comprehensive tests**: All four cardinal directions (0°, 90°, 180°, 270°) plus multi-segment paths are tested.
- **Clean visual differentiation**: One-way channels get SVG arrows, bidirectional channels get the `⇄` prefix.

### Minor Observations

**P2 - Code duplication between midpoint functions**

`getOrthogonalPathMidpointWithAngle()` (lines 376-421) duplicates the midpoint calculation logic from `getOrthogonalPathMidpoint()` (lines 324-363). Consider either:
1. Having `getOrthogonalPathMidpointWithAngle` call `getOrthogonalPathMidpoint` first and then compute the angle in a separate pass, or
2. Extracting the shared midpoint-finding logic into a common helper.

This is a minor maintainability concern - not blocking, but worth addressing to avoid drift between the two implementations.

**P3 - Minor test coverage gap**

The test "returns angle=0 and last point when all points are equal" only asserts `result.angle` but doesn't verify `result.x` and `result.y` are the expected values (5, 5).

### Recommendation

**APPROVE** - The implementation is correct, well-tested, and follows the existing code patterns. The code duplication is a minor maintainability concern that doesn't block this feature.

lsm (commented):

**Recommendation: APPROVE**

All previous review findings have been addressed cleanly. Summary of the fixes:

### P2 (code duplication) — Resolved
`getOrthogonalPathMidpoint` now delegates to `getOrthogonalPathMidpointWithAngle` (line 385-386 of `EdgeRenderer.tsx`). This eliminates the duplication risk entirely — there is a single traversal algorithm, and the old function is now a thin wrapper that returns the same struct but is typed as `Point2D` for existing callers. Clean and correct.

### P2 (magic number) — Resolved
The `+2` constant is now `CHANNEL_GATE_ARROW_EXTRA_PADDING` (line 107) with a clear doc comment. The derived constants `CHANNEL_GATE_ARROW_WIDTH`, `CHANNEL_GATE_ARROW_GAP`, and `CHANNEL_GATE_ARROW_TOTAL` are all named and documented (lines 100-107). Good readability improvement.

### P2 (misleading comment) — Resolved
The SVG transform comment (lines 837-847) now correctly explains the right-to-left application order: `rotate(angle)` pivots around origin first, then `translate(tx)` shifts the already-rotated arrow. This is accurate — in SVG, the rightmost transform in the attribute string is applied first to the coordinate space.

### P3 (test assertions) — Resolved
The degenerate-points test case now asserts both `x` and `y` (lines 911-912). The new `gate badge arrow SVG transform` describe block (lines 930-966) verifies that the transform string starts with `translate(` and contains `rotate(0)`, which guards against accidental reordering of the two SVG transform operations.

### Additional observations (no action needed)

- **loopBadgePosition calls getOrthogonalPathMidpoint twice** (lines 731-733): when `gateBadgePosition` is null and `isCyclic` is true, the midpoint is computed twice — once for `.x` and once for `.y`. This is harmless (it's a cheap computation on a small array) but worth noting for awareness. Not worth a refactor.
- **Test coverage is solid**: 8 unit tests for `getOrthogonalPathMidpointWithAngle` covering all four cardinal angles, multi-segment paths, degenerate cases, and empty input. 4 rendering tests covering arrow presence/absence, bidirectional label prefix, and transform ordering.
- **All 77 tests pass.**

lsm (commented):

### Summary

The worker has addressed all the previous review feedback effectively. The implementation is now well-structured with clear comments explaining the SVG transform ordering, proper code reuse via delegation, and named constants instead of magic numbers.

### Changes Verified

1. **P0 (arrow positioning)**: The comments now correctly explain that SVG transforms apply right-to-left, so `rotate(angle)` happens first (around origin), then `translate(tx)` shifts the rotated arrow. This is the correct interpretation and the implementation is sound.

2. **P1 (code duplication)**: Fixed. `getOrthogonalPathMidpoint` now delegates to `getOrthogonalPathMidpointWithAngle` (line 385-386), eliminating the risk of algorithm divergence.

3. **P2 (magic number)**: Fixed. The `+2` is now named `CHANNEL_GATE_ARROW_EXTRA_PADDING` with a clear comment explaining its purpose.

4. **P3 (test gap)**: The degenerate-points test now includes x/y assertions, and a new test verifies the transform string ordering for horizontal paths.

### Remaining Minor Issue

**P3: Test comment claims more than it delivers** (`EdgeRenderer.test.tsx` lines 935-936)

The comment states "This test verifies the transform string for both horizontal (0°) and vertical (90°) paths" but only the horizontal (angle=0) case is tested. The helper function `getArrowTransform` takes an `angle` parameter that is never used - it always creates the same horizontal node positions regardless of the angle argument.

Consider either:
- Removing the mention of "vertical (90°)" from the comment, or
- Adding a test for vertical paths by using vertical node positions (same x, different y)

This is minor because the core angle computation is tested in the `getOrthogonalPathMidpointWithAngle` suite, and the horizontal transform test validates the string ordering.

### Recommendation

**APPROVE** - All previous feedback has been addressed. The remaining issue is a minor documentation/test mismatch that doesn't affect correctness.

lsm (commented):

**Recommendation: APPROVE**

The fixes for both reported bugs are correctly implemented. The pipeline now tracks gate types per direction end-to-end: `PairAggregate.lowToHighGateType/highToLowGateType` -> `SemanticWorkflowEdge.gateType/reverseGateType` -> `ResolvedWorkflowChannel.reverseGateType` -> badge rendering logic. Tests cover the critical scenarios well.

### What was done right

1. **Bug 1 fix (false bidirectional display)**: The `arrowCount` logic at line ~761 of `EdgeRenderer.tsx` correctly distinguishes three cases: no gate (0 arrows), one-direction gated (1 arrow), both-directions gated (2 arrows). The `hasBothDirectionGates` guard (`isBidirectional && !!channel.gateType && !!channel.reverseGateType`) is precise.

2. **Bug 2 fix (real bidirectional arrows)**: Dual SVG arrow polygons flanking the label are well-structured. The `forwardAngle`/`reverseAngle` computation (`(angle + 180) % 360`) correctly maps path segment direction to opposite-pointing arrows. The SVG transform order (`translate(...) rotate(...)`) is correct -- SVG applies transforms right-to-left, so rotation happens first around origin, then translation positions the rotated arrow.

3. **`getOrthogonalPathMidpointWithAngle`**: Clean extension of the existing midpoint function. The angle derivation from segment direction (0/90/180/270 for right/down/left/up) is correct. Edge cases (empty array, single point, degenerate equal points) are handled with sensible defaults.

4. **`semanticWorkflowGraph.ts`**: The `??=` operator for per-direction gate type assignment correctly uses "first gate wins" semantics. One-way output edges never set `reverseGateType`, which prevents false bidirectional detection downstream. The `as const` annotations on `direction` literals are a nice type-safety improvement.

5. **Test coverage**: The new tests cover all four gate-direction scenarios (forward-only, reverse-only, both, neither) at both the graph-building layer and the rendering layer. The `getOrthogonalPathMidpointWithAngle` unit tests cover all cardinal directions plus L-path and degenerate cases.

- `gateLabelBase` / `gateLabel` alias (`EdgeRenderer.tsx` lines 753-756): `gateLabelBase` is defined on line 753 and immediately assigned to `gateLabel` on line 756 with no other use. The intermediate variable is unnecessary -- `gateLabel` can be assigned directly from the ternary expression.

- **Different gate types on each direction**: When a bidirectional channel has `gateType: 'human'` forward and `reverseGateType: 'check'` reverse, the badge label and color always use the forward type (`effectiveGateType = channel.gateType ?? ...`). This is documented in the comment and seems like a reasonable design choice, but there is no test covering this asymmetry. It would be worth a test to confirm the chosen behavior (or reconsider if both types should be displayed).

PR #1188, opened 2026-04-01
feat: add playwright built-in skills and chrome-devtools MCP default

lsm:
Add two new built-in skills (playwright, playwright-interactive) enabled by default, and a new chrome-devtools MCP server entry disabled by default.

- `playwright` and `playwright-interactive` builtin skills seeded in `initializeBuiltins()` with `enabled: true`, `builtIn: true`
- `chrome-devtools` MCP entry seeded in `seedDefaultMcpEntries()` with `bunx chrome-devtools-mcp@latest --isolated`, `enabled: false`
- 11 new skills-manager unit tests + 5 new seed-defaults tests covering idempotency, config correctness, deletion protection, and enabled state


lsm (commented):

The implementation is clean, well-tested, and follows existing patterns. The two new builtin skills and the chrome-devtools MCP entry are seeded correctly with proper idempotency. All 72 tests pass. I have a few minor observations:

### P3 (nits)

1. **Stale docstring in `seed-defaults.ts`** (line 4): The header still says "Seeds two useful default MCP entries" and the Defaults list only mentions `fetch-mcp` and `brave-search`. It should be updated to "three" and include the new `chrome-devtools` entry.

   `packages/daemon/src/lib/mcp/seed-defaults.ts` lines 4, 8-11

2. **Stale docstring in `seed-defaults.test.ts`** (line 5): The test file header says "Verifies: fetch-mcp and brave-search are created" but should also mention chrome-devtools.

   `packages/daemon/tests/unit/mcp/seed-defaults.test.ts` line 5

### No issues found

- **Correctness**: Both new builtin skills use `sourceType: 'builtin'` with valid `commandName` values, matching the existing type system. The chrome-devtools MCP entry is properly configured with `sourceType: 'stdio'`, `command: 'bunx'`, and sensible args.
- **Idempotency**: All three new entries use the existing `getByName` guard pattern, preventing duplicates on repeated calls.
- **Built-in protection**: Both playwright skills set `builtIn: true` and `enabled: true`, and tests verify the deletion guard in `removeSkill` works.
- **Default states are sensible**: The playwright skills are enabled by default (matching their nature as always-available builtins), while chrome-devtools is disabled (matching brave-search's pattern of requiring user opt-in).
- **Test coverage**: The 16 new tests comprehensively cover creation, commandName correctness, enabled state, deletion protection, idempotency, and total count. They mirror the existing test patterns exactly.
- **Note on builtin injection**: Per CLAUDE.md, the `builtin` sourceType is defined in the type system but not currently injected by `QueryOptionsBuilder` (only `plugin` and `mcp_server` are). These skills will be registered in the DB and visible in the UI, but won't affect SDK sessions until a future change wires up builtin injection. This is consistent with the existing `web-search-mcp` approach where the actual MCP backing is via `mcp_server` sourceType rather than `builtin` directly. Worth noting for the follow-up work.

lsm (commented):

### Summary

This PR adds two new built-in skills (`playwright`, `playwright-interactive`) and a new default MCP entry (`chrome-devtools`). While the database seeding and tests are correctly implemented, there is a **critical architectural issue** that prevents the built-in skills from functioning.

---

### P0 (Blocking): Built-in skills are not injected by QueryOptionsBuilder

**File:** `/packages/daemon/src/lib/agent/query-options-builder.ts`

The `builtin` sourceType is defined in the type system but **NOT currently injected** by `QueryOptionsBuilder`. The documentation at `/docs/features/skills.md` line 118 explicitly states:

> Note: The `builtin` sourceType (slash commands from `.claude/commands/`) is defined in the type system but not currently injected by `QueryOptionsBuilder`. Only `plugin` and `mcp_server` skills are actively injected.

Looking at `query-options-builder.ts`:
- `buildPluginsFromSkills()` (lines 853-868) handles `plugin` skills only
- `getMcpServersFromSkills()` (lines 879-900) handles `mcp_server` skills only
- There is NO method to inject `builtin` skills

**Impact:** The `playwright` and `playwright-interactive` skills will be stored in the database, marked as `enabled: true`, and returned by `getEnabledSkills()`, but they will have **zero actual effect** on agent sessions. This is misleading to users who would expect these skills to work.

**Resolution options:**
1. Implement `builtin` skill injection in `QueryOptionsBuilder` (add to `availableCommands` as described in the design docs)
2. Change these to `mcp_server` or `plugin` skills that ARE currently injected
3. Mark as WIP and add a follow-up task to implement builtin injection

---

### P1: Missing slash command files

**Issue:** The `BuiltinSkillConfig` type requires a `commandName` that maps to a slash command file in `.claude/commands/`. The `commandName` values `playwright` and `playwright-interactive` have no corresponding command files in the repository.

Even if `builtin` injection was implemented, these skills would not function without the actual command files.

**Question:** What are these skills supposed to do? If they are meant to invoke the Playwright MCP server, they should be `mcp_server` type skills referencing a Playwright MCP entry (similar to how `web-search-mcp` references `brave-search`).

---

### P1: Documentation update needed

**File:** `/docs/features/skills.md`

The documentation needs to be updated to:
1. Document the new `playwright` and `playwright-interactive` built-in skills
2. Clarify what these skills do and how to use them
3. Either remove or update the note at line 118 about `builtin` skills not being injected

---

### P2: Stale comment in seed-defaults.ts

**File:** `/packages/daemon/src/lib/mcp/seed-defaults.ts` lines 4-11

The file-level comment says "Seeds two useful default MCP entries" but now seeds three (fetch-mcp, brave-search, chrome-devtools).

---

### P2: Missing skill for chrome-devtools MCP

The `chrome-devtools` MCP entry is added as disabled by default (which is fine), but there is no corresponding skill entry to wrap it. If users want to use chrome-devtools, they would need to:
1. Enable the MCP entry directly (which works)
2. OR have a skill that references it (consistent with the `web-search-mcp` pattern)

Consider whether a `chrome-devtools` skill should also be added (disabled by default).

---

### What looks good

- The database seeding logic is correctly implemented with idempotency
- Tests are comprehensive and cover all the right cases (registration, idempotency, deletion protection, enabled state)
- The `chrome-devtools` MCP entry uses `bunx` with `@latest` for version management

---

### Recommendation

**REQUEST_CHANGES** - The built-in skills will not function because the injection mechanism is not implemented. This needs to be resolved before merge.

lsm (commented):

### Verdict: APPROVE

All previous feedback has been addressed cleanly. The implementation is solid.

### What was fixed

1. **Stale docstrings** — `seed-defaults.ts` now correctly says "three" entries; test file docstring updated.
2. **chrome-devtools-mcp built-in skill** — Added as a proper `mcp_server` type skill wrapping the chrome-devtools MCP entry, disabled by default, following the exact same pattern as `web-search-mcp`. The fallback creation logic and idempotency are correct.
3. **docs/features/skills.md** — All four built-in skills are now documented with name, type, enabled state, and usage instructions.
4. **Tests** — 7 new chrome-devtools-mcp tests cover registration, backing MCP entry creation, config reference, disabled state, deletion protection, idempotency, and pre-existing entry reuse. The total built-in count assertion was updated from 3 to 4.

### Remaining observations (no action needed)

- **P3**: The general idempotency test in `seed-defaults.test.ts` (line 91) only checks `fetch-mcp` and `brave-search` for duplicates, not `chrome-devtools`. The chrome-devtools idempotency is covered by its own dedicated test (line 173), and the total count test (line 103) implicitly validates no extra entries are created, so this is fine.
- **P3**: The `builtin` sourceType skills (`playwright`, `playwright-interactive`) are not injected by `QueryOptionsBuilder.build()` — this is explicitly by design per the task specification, and is documented in both `CLAUDE.md` and `docs/features/skills.md`.

All 79 tests pass (64 skills-manager + 15 seed-defaults). Typecheck and lint are clean.

lsm (commented):

### Summary

The iteration 2 changes correctly address all previous feedback items. The implementation is complete and well-tested.

### Verification Results

**1. Fixed stale docstrings (P3 - addressed)**
- `packages/daemon/src/lib/mcp/seed-defaults.ts` lines 1-14: Updated to document three MCP entries
- `packages/daemon/tests/unit/mcp/seed-defaults.test.ts` lines 1-9: Updated test file docstring

**2. chrome-devtools-mcp skill (P2 - addressed)**
- `packages/daemon/src/lib/mcp/seed-defaults.ts` lines 45-56: Seeds `chrome-devtools` MCP entry with `enabled: false`
- `packages/daemon/src/lib/skills-manager.ts` lines 186-219: `initChromeDevToolsMcp()` creates the skill referencing the MCP entry, disabled by default
- Fallback logic at lines 188-199 creates the MCP entry if missing (defensive pattern)
- 7 new tests at `packages/daemon/tests/unit/skills-manager.test.ts` lines 878-963

**3. Documentation update (P1 - addressed)**
- `docs/features/skills.md` line 11: Lists all four built-in skills
- `docs/features/skills.md` lines 93-121: Dedicated sections for Chrome DevTools, Playwright, and Playwright Interactive

**4. Tests**
- All 79 tests pass (skills-manager.test.ts + seed-defaults.test.ts)
- Tests cover: registration, config correctness, enabled state, idempotency, deletion protection, MCP entry creation, and pre-existing entry reuse

### Technical Verification

- TypeScript compilation passes (`bun run typecheck`)
- All unit tests pass
- Initialization order verified: `seedDefaultMcpEntries()` runs before `initializeBuiltins()` in app.ts

### Recommendation: APPROVE

All previous feedback has been addressed. The implementation follows existing patterns, has comprehensive test coverage, and documentation is updated.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (cannot set event=REQUEST_CHANGES as PR author)

### Overall Assessment

The iteration 3 changes are well done. The `.claude/commands/` slash command files are correctly implemented, the `.gitignore` change is precise, and the docs are updated. That said, I found a few issues worth addressing before merge.

### Issues

**P1 -- `.gitignore` change may expose sensitive `.claude/` files that were previously ignored**

On the `dev` branch, `.gitignore` had `.claude/` which blanket-ignored the entire directory. The new `.gitignore` replaces that with specific ignores for only `settings.local.json` and `worktrees/` (lines 66-67). This correctly un-ignores `.claude/commands/` for tracking, but it also un-ignores any other files that may exist under `.claude/` (e.g., `.claude/credentials.json`, `.claude/settings.json` from other contributors' environments). While `.claude/credentials.json` is typically at `~/.claude/` (not the repo-level `.claude/`), it is worth explicitly ignoring any other sensitive files that could appear in the repo-level `.claude/` directory, or documenting the decision that only `commands/` should be tracked.

**P2 -- `playwright-interactive.md` workspace placeholder is ambiguous for the SDK agent**

In `/Users/lsm/.neokai/projects/dev-neokai-ec0a1deb/worktrees/coder-04062505-780f-4881-a3be-9cb9062790fb-b003b0f7-6cf8-41c7-9ad9-19e0ecd5d518-12eb17d3/.claude/commands/playwright-interactive.md` lines 7-9, the command `make self WORKSPACE=<workspace-path>` uses an angle-bracket placeholder. The SDK agent reads this file as instructions and will need to determine what `<workspace-path>` should resolve to. The companion `playwright.md` file avoids this by detecting `tmp/.dev-server-running` and routing accordingly. Consider whether `playwright-interactive.md` should similarly detect the running server rather than requiring the agent to figure out the workspace path, or at minimum add an explicit note like "replace `<workspace-path>` with the NeoKai workspace root directory (the git repository root)."

**P3 -- Redundant `.claude/settings.local.json` in `.gitignore`**

Line 66 of `.gitignore` (`/Users/lsm/.neokai/projects/dev-neokai-ec0a1deb/worktrees/coder-04062505-780f-4881-a3be-9cb9062790fb-b003b0f7-6cf8-41c7-9ad9-19e0ecd5d518-12eb17d3/.gitignore:66`) has `.claude/settings.local.json`, but line 52 already has `**/settings.local.json` which covers this path via the glob. The explicit entry is harmless (and serves as documentation) but is technically redundant.

### Positive Observations

- The `builtin` sourceType for playwright skills is architecturally correct: `QueryOptionsBuilder` does NOT inject `builtin` skills (only `plugin` and `mcp_server`), so these rely entirely on the SDK discovering `.claude/commands/*.md` files from the working directory. This aligns with the documented behavior in `docs/features/skills.md`.
- The `BuiltinSkillConfig.commandName` values (`'playwright'` and `'playwright-interactive'`) correctly correspond to the filenames `playwright.md` and `playwright-interactive.md` (the SDK strips the `.md` extension when matching).
- Test coverage is thorough: 79 tests across both test files (64 + 15), covering idempotency, deletion protection, config correctness, enabled states, and cross-references between skills and backing MCP server entries.
- All 79 tests pass.
- The chrome-devtools MCP default (disabled by default) is a sensible opt-in addition.

lsm (commented):

---

All three issues from iteration 3 have been properly addressed:

### P1 FIXED: gitignore now properly protects sensitive files

The new pattern is correct:
```gitignore
.claude/*
!.claude/commands/
```

Verified that:
- `.claude/settings.json` is ignored
- `.claude/credentials.json` is ignored  
- `.claude/worktrees/` is ignored
- `.claude/commands/playwright.md` and `.claude/commands/playwright-interactive.md` are tracked

This approach is secure and maintainable - any new sensitive files added to `.claude/` will be ignored by default.

### P2 FIXED: workspace placeholder replaced with explicit detection logic

The updated `playwright-interactive.md` now:
1. Checks for `tmp/.dev-server-running` to detect if a server is already running
2. Uses `$(pwd)` as the workspace root instead of the ambiguous `<workspace-path>` placeholder

The instructions are now copy-pasteable and unambiguous.

### P3 FIXED: redundant gitignore entry removed

The `.claude/settings.local.json` line was removed since `**/settings.local.json` already covers it.

---

**Test verification:**
- `packages/daemon/tests/unit/skills-manager.test.ts`: 64 tests pass
- `packages/daemon/tests/unit/mcp/seed-defaults.test.ts`: 15 tests pass

All four built-in skills are properly registered (web-search-mcp, chrome-devtools-mcp, playwright, playwright-interactive) with correct enabled states and deletion protection.

**Recommendation: APPROVE** - All feedback has been addressed.

lsm (commented):

### Verdict: APPROVE

Re-reviewed iteration 4. All three previously raised issues have been addressed correctly.

### P1 fix (gitignore exposure) -- Resolved

`.gitignore` now uses the pattern:

```
.claude/*
!.claude/commands/
```

This ignores everything under `.claude/` while exclusively allowing `.claude/commands/` to be tracked. Verified with `git check-ignore`:
- `.claude/commands/playwright.md` and `.claude/commands/playwright-interactive.md` are NOT ignored (tracked correctly)
- `.claude/settings.local.json` and `.claude/settings.json` ARE ignored
- Nested files under `.claude/commands/` (e.g. subdirectories) are also NOT ignored
- Only the two command files are currently tracked under `.claude/` in HEAD

This is the correct and minimal gitignore pattern for this use case.

### P2 fix (ambiguous workspace placeholder) -- Resolved

`playwright-interactive.md` step 1 now:
1. Checks for `tmp/.dev-server-running` lock file to detect a running server
2. Uses `make self WORKSPACE=$(pwd)` instead of the ambiguous `<workspace-path>` placeholder
3. Explicitly describes the workspace root as "the git repository root, i.e. the directory containing the Makefile"

This eliminates the ambiguity and gives Claude Code a concrete, copy-pasteable command.

### P3 fix (redundant gitignore entry) -- Resolved

The dedicated `.claude/settings.local.json` line has been removed. It is now covered by the broader `.claude/*` pattern (line 66), which also covers `**/settings.local.json` (line 52). No redundancy remains.

### Overall Assessment

The implementation is clean and well-tested:
- 79 unit tests pass across both test files (skills-manager and seed-defaults)
- All four built-in skills (web-search-mcp, chrome-devtools-mcp, playwright, playwright-interactive) have thorough test coverage: registration, config correctness, enabled state, builtIn delete guard, and idempotency
- The chrome-devtools MCP entry follows the same two-tier pattern (seed-defaults.ts for app_mcp_servers, skills-manager.ts for skills)
- Documentation in `docs/features/skills.md` is comprehensive and accurate

No remaining issues.

lsm (commented):

I reviewed the full diff and all changed files in detail. The PR adds four new built-in skills (playwright, playwright-interactive, chrome-devtools-mcp), binary embedding for slash command files, and an `installSkillFromGit` capability. The builtin registration, idempotency, and seed-defaults logic are well-structured with comprehensive test coverage (73 skills tests, 15 seed-defaults tests all passing). However, there are several issues that need to be addressed before merging.

---

### P0 -- Path traversal in `installSkillFromGit` commandName

**File:** `packages/daemon/src/lib/skills-manager.ts` lines 177-224

The `commandName` parameter is used directly in `join()` calls to construct filesystem paths without any sanitization:

```ts
const skillsDir = join(homedir(), '.neokai', 'skills', commandName);
// ...
await writeFile(join(commandsDir, `${commandName}.md`), content, 'utf-8');
```

A `commandName` like `../../../etc` would resolve to `/etc/passwd.md` in the workspace, and `../../.ssh` would write to `~/.ssh/SKILL.md`. The existing `validateSkillConfig` for `builtin` type only checks that `commandName` is non-empty -- it does not check for path traversal characters. The `installSkillFromGit` method bypasses `validateSkillConfig` entirely.

Compare this to how `pluginPath` is validated (line 398-401) which rejects `../` segments.

**Recommendation:** Either:
1. Validate `commandName` to reject characters like `/`, `..`, and null bytes before using it in paths, OR
2. Extract the basename with `path.basename(commandName)` before constructing paths

---

### P0 -- No fetch timeout or response size limit in `installSkillFromGit`

**File:** `packages/daemon/src/lib/skills-manager.ts` line 184

```ts
const response = await fetch(rawUrl);
```

The fetch has no `AbortSignal` timeout and no content-length check. A slow or unresponsive server could hang the RPC handler indefinitely. A malicious or misconfigured URL could return an arbitrarily large response that gets written to disk.

**Recommendation:** Add a timeout (e.g., 15-30s) via `AbortSignal.timeout()` and check `response.headers.get('content-length')` or limit the response body size.

---

### P1 -- `resolveSkillRawUrl` mishandles branch names containing `/`

**File:** `packages/daemon/src/lib/skills-manager.ts` lines 42-45

The regex `([^/]+)` for the branch segment stops at the first `/`:

```ts
const treeMatch = url.match(/^https:\/\/github\.com\/([^/]+)\/([^/]+)\/tree\/([^/]+)\/(.+)$/);
```

For a URL like `https://github.com/org/repo/tree/feature/my-branch/skills/playwright`, the regex parses:
- branch = `feature` (wrong -- should be `feature/my-branch`)
- path = `my-branch/skills/playwright` (wrong)

This silently generates an incorrect raw URL. The test comment on line 1019 acknowledges this but only tests `v2.0` (dot-separated, not slash-separated).

**Recommendation:** Either document that slash-containing branch names are not supported, or use a greedy/non-greedy approach to parse the URL correctly (e.g., split on `/tree/` first, then separate branch from path).

---

### P1 -- Docs description contradicts actual command file content

**File:** `docs/features/skills.md` lines 103-122

The docs say:
- Playwright: "Runs the NeoKai Playwright end-to-end test suite"
- Playwright Interactive: "Runs NeoKai's Playwright tests in interactive/headed mode"

But the actual command files say the opposite:
- `playwright.md`: "This skill is for automating web browsers -- scraping, UI interaction, form filling, navigation, and screenshot capture. It is NOT for running end-to-end test suites."
- `playwright-interactive.md`: "Persistent browser session for iterative UI debugging and visual QA"

The command files are clearly the correct, authoritative source since they were just written. The docs need to be updated to match.

---

### P1 -- `installSkillFromGit` always overwrites local files but claims idempotency

**File:** `packages/daemon/src/lib/skills-manager.ts` lines 192-201

The idempotency comment says "if a skill with the same name already exists, returns it unchanged" (line 175), and the DB check on line 205 returns early. However, the filesystem writes on lines 195 and 201 happen BEFORE the DB check. So even when the skill already exists in the DB, the method still fetches the remote URL, writes to `~/.neokai/skills/`, and overwrites the workspace `.claude/commands/` file.

This is inconsistent: the DB is idempotent but the filesystem side effects are not.

---

### P1 -- `InstallSkillFromGitParams` type is defined but never used

**File:** `packages/shared/src/types/skills.ts` lines 116-121

The `InstallSkillFromGitParams` interface is defined but never imported anywhere in the codebase. The RPC handler in `skill-handlers.ts` uses inline `as { repoUrl: string; commandName: string }` casting instead of the type.

**Recommendation:** Either import and use the type in the handler for type safety, or remove the dead type to keep knip happy.

---

### P2 -- `installSkillFromGit` writes to `~/.neokai/skills/` even in dev mode

**File:** `packages/daemon/src/lib/skills-manager.ts` lines 192-195

Every `installSkillFromGit` call writes to `~/.neokai/skills/{commandName}/SKILL.md` regardless of whether a workspaceRoot is provided. This creates side effects in the user's home directory even for development/testing scenarios. Consider whether this is always desirable, or if it should be workspace-scoped only.

---

### P2 -- Workspace command file always overwritten (no skip logic)

**File:** `packages/daemon/src/lib/skills-manager.ts` lines 198-201

The `installSkillFromGit` method unconditionally overwrites `{workspaceRoot}/.claude/commands/{commandName}.md` every time it's called. Compare this to `prod-server-embedded.ts` (line 80-86) which checks `access()` and skips existing files to preserve user customizations.

If a user has customized a command file, calling `installSkillFromGit` with the same name will silently overwrite their changes.

---

### P3 -- Test comment is misleading

**File:** `packages/daemon/tests/unit/skills-manager.test.ts` lines 1018-1021

```ts
// Branch = "feature/my-branch", path = "skills/my-skill"
// Note: URL encoding of slashes is NOT done by the input — test the regex boundary
expect(resolveSkillRawUrl('https://github.com/org/repo/tree/v2.0/path/to/skill'))
```

The comment says "Branch = feature/my-branch" but the actual test URL uses `v2.0` as the branch (which contains a dot, not a slash). The test does not actually exercise the slash-in-branch case.

---

### Summary

The builtin skills registration, seed-defaults, and binary embedding are well-implemented. The main concerns are:

1. **Security**: The `installSkillFromGit` method needs `commandName` sanitization against path traversal and fetch timeout/size limits before it can safely be exposed as an RPC endpoint.
2. **Correctness**: The docs need to match the actual command file content; the regex for branch names with slashes is broken; the idempotency claim is misleading.
3. **Dead code**: `InstallSkillFromGitParams` is unused.

lsm (commented):

---

**Recommendation: APPROVE**

All previously raised security and correctness issues have been addressed. The implementation is solid.

### Verified fixes

1. **Path traversal prevention** (`validateCommandName()` at `packages/daemon/src/lib/skills-manager.ts:112`): Rejects empty strings, null bytes, path separators (`/`, `\`), bare `..`, and leading dots. The DB idempotency check at line 264 runs before any network/FS operations -- confirmed by the `is truly idempotent -- does not call fetch on second call` test.

2. **Fetch hardening** (`packages/daemon/src/lib/skills-manager.ts:272-297`): 20s timeout via `AbortController` with proper cleanup in `finally`. Dual size limit check on both `content-length` header and actual `arrayBuffer` byte length (1 MiB cap). Non-ok status codes throw with status info.

3. **URL parsing for branches with slashes** (`packages/daemon/src/lib/skills-manager.ts:58-79`): Uses `indexOf` on fixed markers `/tree/` and `/blob/` rather than splitting the full URL, so branch names like `feature/my-branch` are handled by taking the first segment as branch and rest as path. Well-documented limitation with clear workaround (use raw URLs).

4. **Skip-if-exists guards** for file writes in both `prod-server-embedded.ts:80-86` and `skills-manager.ts:303-320`. Existing user-customized command files are never overwritten.

5. **Proper type imports** -- `InstallSkillFromGitParams` correctly defined in `packages/shared/src/types/skills.ts:116` and exported via the barrel.

### Quality

- 100 tests passing (85 skills-manager + 15 seed-defaults)
- TypeScript typecheck clean
- Oxlint: 0 warnings, 0 errors
- Good test coverage for idempotency, config correctness, deletion protection, enabled state, and security validation
- Documentation in `docs/features/skills.md` updated with all four built-in skills
- `.gitignore` correctly updated to track `.claude/commands/` while ignoring the rest of `.claude/`

### P3 (nit)

In `packages/daemon/src/lib/skills-manager.ts:125`, the checks `commandName.startsWith('../')` and `commandName.endsWith('/..')` are dead code since line 121 already rejects any string containing `/`. They serve as harmless defense-in-depth but could be simplified.

lsm (commented):

---

### Summary

The architectural shift to `packages/skills/` with full directory structures (SKILL.md, scripts/, references/, assets/, agents/) is a significant improvement over the single-file approach. The binary embedding and extraction pipeline is clean and well-structured. Tests are thorough (90 + 15 = 105 passing) and typecheck is clean.

However, there is a security vulnerability in the GitHub directory download code that should be addressed before merging.

---

### P1: Path traversal via unsanitized GitHub API response in `fetchGitHubDirectory`

**File:** `packages/daemon/src/lib/skills-manager.ts`, lines 372-386

The `entry.name` field from the GitHub API response is used directly in `join(destDir, entry.name)` without any path sanitization. While the `commandName` parameter itself is validated by `validateCommandName()`, the GitHub API response fields (`entry.name`, `entry.url`, `entry.download_url`) are trusted without validation.

A malicious or compromised GitHub API response (or a MITM attacker before TLS) could include:

- `entry.name = "../../.bashrc"` -- writes outside `~/.neokai/skills/`
- `entry.name = "../../../.config/autostart/malware.desktop"` -- arbitrary filesystem writes
- `entry.url` pointing to an attacker-controlled server (recursive directory listings redirected)
- `entry.download_url` pointing to an internal network address (SSRF via `fetch`)

**Suggested fix:** Add validation for each `entry.name` (reject `/`, `\\`, `..`, null bytes, leading dots -- same rules as `validateCommandName`), validate that `entry.url` starts with `https://api.github.com/` before recursing, and optionally validate `entry.download_url` starts with `https://raw.githubusercontent.com/`.

---

### P1: No recursion depth or total request limit in `fetchGitHubDirectory`

**File:** `packages/daemon/src/lib/skills-manager.ts`, lines 344-387

`fetchGitHubDirectory` recurses without any depth limit or total request count limit. A deeply nested directory structure (or a malicious response with circular `entry.url` references) would cause unbounded recursion. Each request has a 20-second timeout, so a deeply nested repo could also cause the operation to hang for a very long time (20s * N levels).

**Suggested fix:** Add a max depth parameter (e.g., 5 levels) and/or a max file count limit (e.g., 100 files total).

---

### P2: Dead code in `validateCommandName`

**File:** `packages/daemon/src/lib/skills-manager.ts`, lines 154-157

The `..` traversal check on line 155 is effectively unreachable:

```ts
if (commandName === '..' || commandName.startsWith('../') || commandName.endsWith('/..')) {
```

- `commandName.startsWith('../')` is already caught by the `/` check on line 151
- `commandName.endsWith('/..')` is also already caught by the `/` check on line 151
- `commandName === '..'` would be caught by the `.` prefix check on line 159

This is not a bug (the validation is correct), but the dead code adds confusion. Consider removing or simplifying.

---

### P3: Minor observations

- The doc comment on `skill.installFromGit` in `skill-handlers.ts` (line 131) says "fetch a SKILL.md from a git repo URL" but the implementation now fetches entire directories. The comment should be updated.
- The `_workspaceRoot` parameter in `installSkillFromGit` is marked as "kept for API compatibility" but is not used. Consider adding a deprecation notice or removing it in a future iteration.
- The `embedded-assets.d.ts` correctly declares `embeddedBuiltinSkills: Map<string, string>` which matches the generated code.

---

### What looks good

- **Binary embedding pipeline**: `generate-embedded-assets.ts` -> `embedded-assets.ts` -> `prod-server-embedded.ts` extraction is clean and well-structured
- **Idempotency**: Both `initializeBuiltins()` and `installSkillFromGit()` are properly idempotent -- no duplicates on repeated calls
- **Test coverage**: 90 skills-manager tests + 15 seed-defaults tests covering CRUD, validation, built-in initialization, idempotency, deletion protection, and security edge cases
- **Security hardening on user input**: `validateCommandName()` properly rejects null bytes, path separators, traversal, and hidden file names
- **Documentation**: `docs/features/skills.md` accurately describes the architecture, key files, and source types
- **Skill content quality**: The playwright SKILL.md files are comprehensive with clear guardrails and workflow instructions
- **Extraction safety**: The embedded binary extraction in `prod-server-embedded.ts` correctly skips existing files to preserve user customizations

lsm (commented):

**Recommendation: APPROVE**

All P1 and P2 issues from the previous review have been correctly addressed. Here is the verification for each fix:

---

### P1: Entry name path traversal -- FIXED

`fetchGitHubDirectory()` now calls `validateCommandName(entry.name)` on every entry from the GitHub API response (line 399) before the name is used in any `join()` path. The validation is placed early in the loop iteration, before both the file-write branch and the directory-recurse branch, so there is no code path where an unsanitized `entry.name` reaches `join()`.

The catch block (lines 400-403) discards the inner error and throws a fresh one with the offending name in the message, which is appropriate -- the original error from `validateCommandName` is an implementation detail, and the outer message clearly identifies the source as the GitHub API response.

### P1: Unbounded recursion -- FIXED

Two new constants (`SKILL_FETCH_MAX_DEPTH = 5`, `SKILL_FETCH_MAX_FILES = 100`) enforce resource limits on `fetchGitHubDirectory`. The depth check happens at function entry (line 364) before any API call is made, so depth-bomb inputs fail fast without wasting network requests. The file count uses a shared mutable counter object (`{ value: 0 }`) passed through recursion, which correctly counts across the entire call tree rather than per-level. The check is `>=` (line 407), so exactly 100 files are allowed and the 101st triggers the error.

### P2: Null byte check -- FIXED

`validateCommandName()` retains the explicit `commandName.includes('\0')` check at line 155, which runs before the path-separator check. This is necessary because a null byte could truncate the string before the `/` or `\` check sees it.

### P2: "." and ".." as special names -- FIXED

Lines 163-164 now explicitly reject bare `"."` and `".."` with a clear error message, and the existing leading-dot check (line 167) still catches all other dot-prefixed names.

### Test coverage

Six new tests cover all the addressed issues:
- Path traversal entry name from GitHub API response (line 1308)
- Leading-dot entry name from GitHub API response (line 1342)
- Max nesting depth exceeded (line 1375)
- Max file count exceeded (line 1406)
- Bare `"."` as command name (line 1073)
- Bare `".."` as command name (line 1076)

All 111 tests pass (96 in `skills-manager.test.ts` + 15 in `seed-defaults.test.ts`).

---

No remaining P0, P1, or P2 issues. Ship it.

lsm (commented):

**Recommendation: APPROVE**

The iteration 14 changes correctly implement the end-to-end builtin skill loading flow for both dev and prod environments. The architecture is clean, tests are thorough, and the security posture is sound.

---

### What was reviewed (iteration 14 scope)

1. `packages/cli/src/skill-utils.ts` -- new shared utility for recursive no-overwrite directory sync
2. `packages/cli/src/dev-server.ts` -- `ensureBuiltinSkills()` call on startup
3. `packages/daemon/src/lib/agent/query-options-builder.ts` -- new `buildPluginsFromBuiltinSkills()` method
4. `packages/cli/tests/skill-utils.test.ts` -- 8 new tests
5. `packages/daemon/tests/unit/agent/query-options-builder.test.ts` -- 2 new builtin skill tests

---

### Correctness of the skill loading flow

The flow is well-designed across both server modes:

- **Dev server** (`dev-server.ts:90-92`): Resolves `../../skills` relative to the CLI source, syncs to `~/.neokai/skills/`. This correctly finds the monorepo source tree.
- **Prod server** (`prod-server-embedded.ts:76-89`): Extracts from `embeddedBuiltinSkills` (generated by `scripts/generate-embedded-assets.ts`) to the same destination. Same no-overwrite policy.
- **QueryOptionsBuilder** (`query-options-builder.ts:882-898`): `buildPluginsFromBuiltinSkills()` iterates enabled builtin skills, resolves each `commandName` to `~/.neokai/skills/{commandName}/`, and injects as `{ type: 'local', path }` plugins. The builder correctly merges these alongside `buildPluginsFromSkills()` at line 124-127.
- **SkillsManager** (`skills-manager.ts:547-586`): Seeds both `playwright` and `playwright-interactive` with `sourceType: 'builtin'`, `enabled: true`, `builtIn: true`. The idempotency check (`getByName`) prevents duplicate entries.

The data flow is consistent: seed DB entries -> sync files to disk -> QueryOptionsBuilder resolves paths -> SDK receives plugin entries. Both dev and prod paths converge at `~/.neokai/skills/`.

---

### Security of file operations

The sync logic is safe:
- `syncDir` uses `readdir` with `withFileTypes: true` -- `entry.name` is just the filename from the OS, not user input, so path traversal is not a concern here.
- Symlinks are silently skipped (neither `isDirectory()` nor `isFile()` returns true for symlinks via Dirent). This is acceptable since the skills directory has no symlinks and the embedded binary extraction path reads file content directly.
- The `ensureBuiltinSkills` wrapper catches and logs all errors without rethrowing, which is correct for a startup step that must not block the server.
- `mkdir(dirname(dest), { recursive: true })` is placed right before `writeFile`, avoiding TOCTOU issues (the check-then-create pattern is safe here because the worst case is a redundant `mkdir` call, not data corruption).

---

### Test coverage

**skill-utils.test.ts** (8 tests): Covers the key scenarios -- file copy with directory structure preservation, count accuracy, no-overwrite behavior, missing source directory, multiple skill directories, deep nesting, error swallowing in `ensureBuiltinSkills`. Good coverage.

**query-options-builder.test.ts** (2 new tests): Tests that enabled builtin skills produce local plugins pointing to `~/.neokai/skills/{commandName}`, and that disabled builtin skills are excluded. Both tests use realistic skill shapes matching the `BuiltinSkillConfig` type.

All 104 query-options-builder tests and 8 skill-utils tests pass.

---

1. **Silent missing-directory injection**: `buildPluginsFromBuiltinSkills()` does not verify that the skill directory actually exists on disk before injecting the plugin path. If the sync step failed silently (by design), the SDK would receive a path to a non-existent directory. This is likely harmless -- the SDK should handle missing plugin directories gracefully -- but it means the failure mode is invisible. Worth a follow-up if users report skills not working.

2. **Symlink gap in `syncDir`**: The `syncDir` function's `for (const entry of entries)` block only handles `isDirectory()` and `isFile()`. If someone adds a symlink to `packages/skills/`, it would be silently skipped. Not a concern today (no symlinks exist), and the comment at line 25 explains the function is specifically for copying from the source tree. Just noting for future awareness.

3. **Pre-existing**: Double `/**` at line 742-743 of `query-options-builder.ts` (a duplicate JSDoc opening). Not introduced by this PR.

lsm (commented):

### Recommendation: APPROVE

The frontend implementation for the "Install from Git" dialog is well-executed and consistent with existing patterns. The code is clean, well-tested, and integrates properly with the skills store and RPC handlers.

---

### Strengths

**Test coverage is thorough.** The 15 tests in `InstallSkillFromGitDialog.test.tsx` cover all the key behaviors: auto-derivation, touch-tracking, validation (empty URL, non-https URL, empty name, invalid characters), submission flow, success/error toasts, dialog-not-closing-on-error, cancel, and form reset. The 2 new tests in `SkillsRegistry.test.tsx` properly verify the "Install from Git" button renders and opens the dialog.

**Preact patterns are correct.** The `useState`-based form state with `commandNameTouched` to gate auto-derivation is the right approach. The `handleClose` function properly resets form + errors + delegates to parent's `onClose`. Error handling follows the same `toast.error` + `isSubmitting` pattern as `AddSkillDialog`.

**Consistent UI patterns.** The dialog uses the same `Modal` + `Button` + `cn` styling, the same error display pattern (red border + red text below), and the same `size="md"` as `AddSkillDialog`. The layout with Cancel/Install buttons in `flex justify-end gap-3 pt-2` matches the sibling dialog exactly.

**Good UX decisions.** Dialog stays open on error (allowing URL correction). Loading state disables the submit button via `Button`'s `loading` prop (which sets `disabled`). Form resets on successful close. `autoFocus` on the URL field gets the user straight to typing.

---

### Issues

**P2: URL validation regex accepts `http://` but the error message says "https://".**

In `InstallSkillFromGitDialog.tsx` line 76, the regex `^https?:\/\/` accepts both `http://` and `https://`, but the error message on line 77 reads `"URL must start with https://"`. Either the regex should be `^https:\/\/` (to match the message), or the message should say `"URL must start with http:// or https://"`. Given the task description says "https only" and the placeholder only shows `https://` URLs, tightening the regex is the better fix.

**P3: `onInput` vs `onChange` inconsistency with sibling dialogs.**

`InstallSkillFromGitDialog` uses `onInput` for text inputs (lines 132, 152) while `AddSkillDialog` and `EditSkillDialog` use `onChange`. Both work correctly in Preact (unlike React, Preact fires `onChange` on every keystroke for text inputs). This is purely a consistency concern -- there is no functional bug here.

**P3: `deriveCommandName` produces surprising results for blob URLs.**

When a user pastes a blob URL like `https://github.com/x/y/blob/main/skills/playwright/SKILL.md`, the function returns `skill-md` (the slugified last segment). This is technically correct per the implementation, but the comment and placeholder text suggest tree URLs. The name field is editable, so users can correct it, but a minor guard (e.g., stripping `.md` extensions) could improve the auto-derivation. Very low priority since the field is always editable.

---

### Files Reviewed

- `packages/web/src/components/settings/InstallSkillFromGitDialog.tsx` -- New component, 185 lines
- `packages/web/src/components/settings/SkillsRegistry.tsx` -- Added "Install from Git" button and dialog state
- `packages/web/src/lib/skills-store.ts` -- Added `installSkillFromGit()` method
- `packages/web/src/components/settings/__tests__/InstallSkillFromGitDialog.test.tsx` -- 15 new tests
- `packages/web/src/components/settings/__tests__/SkillsRegistry.test.tsx` -- 2 new tests
- `packages/daemon/src/lib/rpc-handlers/skill-handlers.ts` -- Verified backend handler
- `packages/shared/src/types/skills.ts` -- Verified `InstallSkillFromGitParams` type

PR #1190, opened 2026-04-02
docs: plan for customizable gate badges and script-based checks

lsm:
## Summary
- Adds implementation plan for two gate system enhancements: author-defined `label`/`color` on Gate entities with heuristic fallback, and script-based gate checks (bash/node/python3) via `Bun.spawn()`
- 5 milestones, 15 tasks covering shared types, backend script execution, async evaluation migration, frontend badge/editor UI, and E2E testing

## Test plan
- [ ] Review plan for completeness against explorer findings and fact-check results
- [ ] Verify task ordering and dependencies are correct

### DIFF
--- a/docs/plans/enhance-gate-system-customizable-badges-and-script-based-che.md
+++ b/docs/plans/enhance-gate-system-customizable-badges-and-script-based-che.md
@@ -28,22 +28,30 @@ Key design decisions:
 - Default timeout: 30s, `killSignal: 'SIGKILL'`
 - `maxBuffer: 1MB` on stdout — enforced via chunk-by-chunk byte counting during streaming read, not by buffering everything then checking size
 - Restricted env (no API keys or credentials) — scripts are **trusted** (workspace-author-configured), but env is restricted as defense-in-depth to prevent accidental credential leakage
-- Per-gate evaluation coalescing with re-run-if-dirty pattern: concurrent evaluations for the same gateId share one in-flight result, but if new gate data arrives during execution, a follow-up re-evaluation is scheduled
+- Per-gate evaluation coalescing with re-run-if-dirty pattern: concurrent evaluations for the same gate share one in-flight result (keyed by `${runId}:${gateId}`), but if new gate data arrives during execution, a follow-up re-evaluation is scheduled
 - Deep merge of JSON stdout with depth limit (max 5 levels), rejecting `__proto__`/`constructor`/`prototype` keys
 
 ### Part C: Async Migration
 
-`evaluateGate()` becomes async to support script execution. A private sync helper (`evaluateFieldsSync`) preserves the pure declarative path for gates without scripts. The **frontend** retains its own independent `evaluateGateStatus()` function in `WorkflowCanvas.tsx` (which is sync and field-only) — it is not affected by the backend async migration.
+`evaluateGate()` becomes async to support script execution. A private sync helper (`evaluateFieldsSync`) preserves the pure declarative path for gates without scripts. The **frontend** retains its own independent `evaluateGateStatus()` function in `WorkflowCanvas.tsx` (which is sync and field-only) — it is not affected by the backend async migration, **but** it does access `gate.fields` directly and needs a `?? []` guard (see Task 1.4).
 
 `isChannelOpen()` remains **synchronous** — it is a convenience function for checking gate status from data already in hand, not for triggering script execution. It has no `scriptExecutor` context.
 
 ### Part D: Concurrency
 
-Per-gate evaluation coalescing prevents overlapping evaluations for the same gate (deduplication). A global concurrency semaphore (default: 4) caps the total number of simultaneous script processes across all gates to prevent resource exhaustion.
+Per-gate evaluation coalescing prevents overlapping evaluations for the same gate+run pair (deduplication). Maps are keyed by composite `${runId}:${gateId}` to prevent cross-run state leakage. A global concurrency semaphore (default: 4) caps the total number of simultaneous script processes across all gates to prevent resource exhaustion.
+
+### Part E: Script Error Transport to Frontend
+
+The frontend currently receives gate data changes via `space.gateData.updated` events (payload: `{ runId, gateId, data }`). `GateEvalResult.reason` (backend-only) is not available to the frontend. To surface script errors in the UI, the `write_gate` MCP tool will store the evaluation result (including `reason`) in a special `_scriptResult` key within the gate data itself. The frontend can then read `gateData._scriptResult.reason` to display error information.
+
+### Part F: `fields` Optional Safety
+
+Making `fields` optional on `Gate` introduces `undefined` at every `gate.fields` access site. Task 1.4 provides a comprehensive audit and guard addition across the entire codebase (frontend + backend).
 
 ---
 
-## Milestone 1: Shared Type Changes (Gate Interface + Validation)
+## Milestone 1: Shared Type Changes (Gate Interface + Validation + Safety)
 
 ### Task 1.1: Extend Gate interface with label, color, script fields
 
@@ -65,7 +73,7 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
 3. Make `fields` optional on `Gate` (change `fields: GateField[]` to `fields?: GateField[]`)
 4. Update `computeGateDefaults()` signature to accept `fields?: GateField[]` — internally treat `undefined` as `[]`
 5. Export `GateScript` from the shared module barrel
-6. **Update all `computeGateDefaults` call sites** that pass `gate.fields` or `gateDef.fields` — since `fields` is now optional, these call sites remain valid because `computeGateDefaults` accepts `undefined`. Verify by checking:
+6. Update all `computeGateDefaults` call sites — since `fields` is now optional, these call sites remain valid because `computeGateDefaults` accepts `undefined`. Verify:
    - `channel-router.ts` lines 594, 728, 741
    - `node-agent-tools.ts` lines 585, 621
    - `gate-data-repository.ts` documentation (line 112)
@@ -76,7 +84,6 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
 - `computeGateDefaults(undefined)` returns `{}`
 - `computeGateDefaults([])` returns `{}`
 - `computeGateDefaults(someFields)` returns the same result as before
-- No existing code breaks — all `computeGateDefaults(gate.fields)` call sites still work because `gate.fields` is `GateField[] | undefined` and the function now accepts that
 - `GateScript` type is exported from `@neokai/shared`
 
 **Dependencies:** None
@@ -96,12 +103,13 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
    - `source` is a non-empty string
    - `timeoutMs` is positive number if present (max 120000)
 4. Add `validateGate(gate: unknown): string[]` — top-level validator that:
-   - Calls `validateGateFields()` when `fields` is present
+   - Calls `validateGateFields()` when `fields` is present and non-empty
    - Calls `validateGateColor()`, `validateGateLabel()`, `validateGateScript()`
    - Validates that at least one of `fields` (non-empty) or `script` is present
+   - **Important:** This validator is only applied to **new/updated gates** (via `GateEditorPanel` and MCP tool handlers), never to existing gates loaded from storage. Existing gates with `fields: []` remain valid at runtime — the strict rule only gates creation/modification.
 
 **Acceptance criteria:**
-- `validateGate({ id: 'g1', fields: [], resetOnCycle: false })` returns errors (empty fields, no script)
+- `validateGate({ id: 'g1', fields: [], resetOnCycle: false })` returns errors (empty fields, no script) — this is a **creation-time** validation failure
 - `validateGate({ id: 'g1', fields: [{ ... }], resetOnCycle: false })` returns `[]`
 - `validateGate({ id: 'g1', script: { interpreter: 'node', source: '...' }, resetOnCycle: false })` returns `[]`
 - Invalid color `'red'` produces error
@@ -110,6 +118,7 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
 - Empty `source` produces error
 - `timeoutMs: 200000` produces error (exceeds 120s max)
 - Label longer than 20 chars produces error
+- Existing gates with `fields: []` in the DB are NOT re-validated on load
 
 **Dependencies:** Task 1.1
 **Agent type:** coder
@@ -126,6 +135,7 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
 3. Add tests for `Gate` with `fields` omitted (script-only gate) persisting correctly
 4. Add tests for `computeGateDefaults` with `undefined` and `[]` fields
 5. Verify backward compatibility: existing gates without `label`/`color`/`script` round-trip unchanged
+6. Verify that `validateGate` is a **creation-time** validator, not applied on load — existing gates with `fields: []` deserialize from storage without errors
 
 **Acceptance criteria:**
 - All new tests pass
@@ -134,6 +144,44 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
 - Gate with only script (no fields) persists and round-trips correctly
 
 **Dependencies:** Task 1.1, Task 1.2
+**Agent type:** coder
+
+---
+
+### Task 1.4: Add `?? []` safety guards for optional `gate.fields`
+
+**Rationale:** Making `fields` optional on `Gate` means every bare `gate.fields` access can produce `undefined`. A comprehensive audit is needed to add `?? []` fallback guards at every access site across the entire codebase. This is a **separate task** from Task 1.1 because the scope spans many files in both frontend and backend.
+
+**Files to update (comprehensive list from codebase audit):**
+
+**Frontend:**
+- `packages/web/src/components/space/WorkflowCanvas.tsx` — lines 124, 127, 131, 142 (`gate.fields.length`, `gate.fields.every()`, `gate.fields.find()`, `for (const field of gate.fields)`)
+- `packages/web/src/components/space/visual-editor/GateEditorPanel.tsx` — lines 41, 53, 54, 58, 71, 81, 149, 152 (spread, filter, map, length on `gate.fields`)
+- `packages/web/src/components/space/visual-editor/ChannelEdgeConfigPanel.tsx` — lines 173, 177 (if any `gate.fields` access)
+- `packages/web/src/components/space/visual-editor/VisualWorkflowEditor.tsx` — lines 124, 1013 (`gate.fields` spread in presets/copy)
+- `packages/web/src/components/space/visual-editor/semanticWorkflowGraph.ts` — line 59 (`gate.fields; if (fields.length === 0)`)
+
+**Backend:**
+- `packages/daemon/src/lib/space/runtime/gate-evaluator.ts` — line 171 (`for (const field of gate.fields)`)
+- `packages/daemon/src/lib/space/agents/task-agent.ts` — line 138 (`gate.fields.map(...)`)
+- `packages/daemon/src/lib/space/tools/node-agent-tools.ts` — lines 498, 583, 621, 629, 670, 678
+
+**Subtasks:**
+1. For each file listed above, replace every bare `gate.fields` or `gateDef.fields` with `(gate.fields ?? [])` or `(gateDef.fields ?? [])`
+2. For iteration patterns (`for...of`, `.map()`, `.filter()`, `.every()`, `.find()`, `.some()`, `.length`), wrap with `?? []` at the point of access
+3. For spread patterns (`...gate.fields`), wrap with `...(gate.fields ?? [])`
+4. In `evaluateFieldsSync` (extracted in Task 3.1): use `(gate.fields ?? [])` for the field iteration loop
+5. In `node-agent-tools.ts` lines 670, 678 specifically: `new Map((gateDef.fields ?? []).map(...))` and `(gateDef.fields ?? []).map(...)` — these crash for script-only gates if unguarded
+6. Run full typecheck after changes to verify no `undefined` errors
+
+**Acceptance criteria:**
+- `grep -rn 'gate\.fields\b' packages/` returns zero instances of bare `gate.fields` (all accesses guarded with `?? []`)
+- `grep -rn 'gateDef\.fields\b' packages/` returns zero instances of bare `gateDef.fields`
+- TypeScript compiles with no errors
+- Script-only gates (no `fields`) do not crash at any access site
+- Field-based gates continue to work as before
+
+**Dependencies:** Task 1.1
 **Agent type:** coder
 
 ---
@@ -155,7 +203,7 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
    ```
 2. Define `RESTRICTED_ENV_PATTERNS` — list of env var prefixes/keys to strip. **Threat model:** scripts are trusted (workspace-author-configured) but env is restricted as defense-in-depth:
    - Prefixes: `ANTHROPIC_`, `CLAUDE_`, `GLM_`, `ZHIPU_`, `COPILOT_`, `NEOKAI_SECRET_`
-   - Exact keys: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `GITHUB_TOKEN`, `NPM_TOKEN`, `DATABASE_URL`, `PRIVATE_KEY`, any key matching `/SECRET|TOKEN|PASSWORD|CREDENTIAL|API_KEY/i`
+   - Pattern: any key matching `/SECRET|TOKEN|PASSWORD|CREDENTIAL|API_KEY/i`
    - Always allow: `PATH`, `HOME`, `USER`, `SHELL`, `LANG`, `TERM`, `TMPDIR`
 3. Implement `buildRestrictedEnv(): Record<string, string | undefined>`:
    - Start with `process.env` copy
@@ -173,7 +221,7 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
    - Set `cwd` to `context.workspacePath`
    - Determine interpreter binary: `bash` -> `['bash', '-c', script.source]`, `node` -> `['node', '-e', script.source]`, `python3` -> `['python3', '-c', script.source]`
    - Spawn with `Bun.spawn()`, capture stdout + stderr
-   - Apply `timeoutMs` (default 30000) — if Bun.spawn `timeout` option works reliably, use it; otherwise use `setTimeout` + `process.kill(pid, 'SIGKILL')` as fallback
+   - Apply `timeoutMs` (default 30000) — use `setTimeout` + `process.kill(pid, 'SIGKILL')` for deterministic timeout handling
    - **`maxBuffer` enforcement via streaming**: read stdout chunk-by-chunk from `proc.stdout` using a loop (e.g., `proc.stdout.getReader()` or `for await (const chunk of proc.stdout)`), accumulating byte count. If accumulated bytes exceed 1MB, kill the process and return failure. This avoids buffering everything before checking size.
    - On exit code 0: parse stdout as JSON, deep-merge with depth limit into empty object, return `{ success: true, data }`
    - On non-zero exit or timeout: return `{ success: false, error: stderr.trim() || 'exit code N' }`
@@ -227,6 +275,7 @@ Per-gate evaluation coalescing prevents overlapping evaluations for the same gat
 
 **Subtasks:**
 1. Extract sync field evaluation into private helper `evaluateFieldsSync(gate: Gate, data: Record<string, unknown>): GateEvalResult`
+   - Use `(gate.fields ?? [])` for the iteration — script-only gates with no fields will immediately return `{ open: true }`
 2. Make `evaluateGate()` async: `async evaluateGate(gate: Gate, data: Record<string, unknown>, scriptExecutor?: GateScriptExecutorFn): Promise<GateEvalResult>`
    - `GateScriptExecutorFn` type: `(script: GateScript, context: { workspacePath: string; gateId: string; runId: string }) => Promise<GateScriptResult>`
 3. If `gate.script` is defined and `scriptExecutor` is provided:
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured plan that accurately reflects the existing codebase. The 5-milestone, 15-task breakdown is appropriately scoped, dependencies are correct, and the critical path is sensible. I verified the plan's assumptions against the actual code (`Gate` interface at `space.ts:580`, `evaluateGate` in `gate-evaluator.ts:170`, `evaluateGateById` in `channel-router.ts:579`, `SemanticWorkflowEdge` in `semanticWorkflowGraph.ts:8`). A few observations:

### Strengths

- **Accurate codebase mapping.** File paths, line references, and interface shapes match the current code. The plan correctly identifies that `evaluateGateById` in `ChannelRouter` is a private sync method calling `evaluateGate`, and that `canDeliver`/`deliverMessage`/`onGateDataChanged` are already async.
- **Sound security posture.** `Bun.spawn()` array form (no shell injection), restricted env stripping, prototype pollution guards, depth-limited deep merge, buffer limits, and kill-on-timeout are all appropriate.
- **Good backward compatibility story.** Making `fields` optional with `[]` default, preserving heuristic fallback for label/color, and keeping a sync fast-path for gates without scripts are the right calls.
- **Correct async migration strategy.** The plan correctly notes that `ChannelRouter` methods are already async, so making `evaluateGateById` async is a contained change. The `isChannelOpen` function is only used in daemon code (not frontend), limiting blast radius.

### P2: Should-fix issues

**P2-1: `isChannelOpen` async migration is under-specified (Task 3.1).** The plan says to make `isChannelOpen` async and add "an async wrapper" for sync callers. But `isChannelOpen` is not called from `ChannelRouter` at all — `ChannelRouter` calls `evaluateGateById` directly. `isChannelOpen` is used in tests and could be used by other consumers. The plan should clarify: either (a) keep `isChannelOpen` sync (it does not need script execution since it has no `scriptExecutor` context), or (b) make it async but enumerate all callers. Option (a) seems simpler and more correct — `isChannelOpen` is a convenience function for checking gate status from data already in hand, not for triggering script execution.

**P2-2: Missing `WorkflowCanvas.tsx` caller of `evaluateGate`.** Grep shows `packages/web/src/components/space/WorkflowCanvas.tsx` imports/calls `evaluateGate`. Since `evaluateGate` becomes async, this frontend caller needs to be updated too. The plan does not mention this file in any task. However, the frontend cannot and should not run `Bun.spawn()`, so the plan should explicitly address this: either the frontend uses the sync-only path (no `scriptExecutor` parameter), or a separate sync export is maintained for frontend use. This is important because the frontend imports from `@neokai/shared` or daemon code — if the signature changes, it will break.

### P3: Nits and suggestions

**P3-1: `computeGateDefaults` acceptance criteria in Task 1.1 are slightly off.** The plan says `computeGateDefaults(undefined)` should return `{}`, but the current signature is `computeGateDefaults(fields: GateField[])`. If `fields` becomes optional on `Gate`, the function should handle `undefined` input, but the AC should say "update `computeGateDefaults` to accept `GateField[] | undefined`" rather than showing it called with `undefined` as if the current code already supports it.

**P3-2: Per-gate evaluation queue design (Task 3.2).** The plan proposes `Map<string, Promise<GateEvalResult>>` for serialization. This is fine for deduplication (coalescing concurrent evaluations), but "serialization" (FIFO queue) would need a different pattern — the current design would make concurrent callers share the same result rather than queue sequentially. The plan text says "await it instead of starting a new one" which is coalescing, not serialization. Consider clarifying which semantic is intended, as it affects correctness for rapid gate data changes.

**P3-3: E2E Task 5.2 dependency list.** Task 5.2 depends on Tasks 3.2 and 3.3, but it also requires the frontend script editor (Task 4.4) to configure the script via UI. The dependency should include Task 4.4.

**P3-4: The "Lint Check" preset in Task 4.4 is not useful.** `npm run lint 2>/dev/null; echo $?` outputs `0` or `1` as plain text, not JSON. The script executor expects JSON stdout for data merging. The preset should output `{"passed": true}` or similar, like the "Type Check" preset does.

lsm (commented):

> Recommendation: **REQUEST_CHANGES** (posting as COMMENT — same author)

---

### Overall Assessment

The plan is well-structured, covers the happy path thoroughly, and demonstrates solid awareness of the existing architecture (Bun.spawn pattern from workflow-executor.ts, gate-evaluator.ts sync nature, SemanticWorkflowEdge naming). However several significant gaps and risks need addressing before implementation.

---

### P1 — Should Fix

**[P1] Task 4.1 proposes a breaking return-type change to `resolveSemanticGateType`**
`resolveSemanticGateType` currently returns `SemanticWorkflowEdge['gateType']` (a string union). The plan changes it to return `{ type, label?, color? }`. This is a breaking refactor of a private function whose callers inside `buildSemanticWorkflowEdges` use the returned value directly (lines 164-186 of `semanticWorkflowGraph.ts`). The plan mentions updating callers but does not name them explicitly, and the `PairAggregate` intermediate type also stores `lowToHighGateType` and `highToLowGateType`. All three aggregate fields must be updated or the bidirectional gate label/color will be lost when the aggregate is collapsed to a `SemanticWorkflowEdge`. The plan should explicitly call out `PairAggregate` and the collapse logic as in-scope for Task 4.1.

**[P1] Bidirectional edges have two gate types (`gateType` / `reverseGateType`) but the plan only adds a single `gateLabel`/`gateColor` pair to `SemanticWorkflowEdge`**
When a bidirectional edge has two different gates (each with its own label/color), the `effectiveGateType = channel.gateType ?? channel.reverseGateType` logic in EdgeRenderer (line 749) collapses them to one. The plan follows the same shortcut for custom labels, so a gate on the reverse direction could silently lose its custom badge. The plan should either acknowledge this limitation explicitly, or add `reverseGateLabel`/`reverseGateColor` fields to the interface.

**[P1] `computeGateDefaults` callers will break when `fields` becomes optional**
In `channel-router.ts` lines 594 and 728/741, `computeGateDefaults(gateDef.fields)` is called directly. In `node-agent-tools.ts` line 621, the same pattern appears. If `fields` becomes `fields?: GateField[]`, TypeScript will flag these as potential `undefined` arguments. Task 1.1 acceptance criteria says "all call sites already pass `fields` explicitly" — this is incorrect. They pass `gateDef.fields` which will be `GateField[] | undefined`. Task 1.1 needs an explicit subtask to update all `computeGateDefaults` call sites, or to update `computeGateDefaults`'s signature to accept `GateField[] | undefined`.

**[P1] `workspacePath` is not wired into `ChannelRouter` — scripts will run from the wrong directory**
The plan specifies `executeGateScript(script, { workspacePath })` (Task 2.1) and that `ChannelRouterConfig` gains `scriptExecutor?: GateScriptExecutor` (Task 3.2). However, `workspacePath` is never part of `ChannelRouterConfig`. Looking at `task-agent-manager.ts` line 622-629, the `ChannelRouter` is constructed without `workspacePath`. Without a subtask to thread `workspacePath` into `ChannelRouterConfig`, scripts would run from `process.cwd()` (the daemon working directory) rather than the space workspace, silently breaking any workspace-relative script checks.

**[P1] Missing task: audit storage serialisation for new optional Gate fields**
Gates are persisted as JSON in the `gates` column of `space_workflows`. There is no task that audits `SpaceWorkflowRepository` and all `WorkflowUpdatePayload` usage to ensure `undefined` `fields`, `label`, `color`, and `script` are handled correctly. If the repository or downstream code does a strict cast (e.g., `fields: row.fields` without null-coalescing), existing gates without these fields will produce TypeScript or runtime errors. A subtask should cover this audit.

**[P1] `node-agent-tools.ts` async migration scope is under-specified**
Task 3.3 says "find where `evaluateGate()` is called" but there are two call sites (lines 624 and 699). Line 699 is in the `write_gate` MCP tool handler — the hot path invoked by agents in real workflows. There is no task to update `task-agent-manager.ts` to construct and inject a `GateScriptExecutor` into the node-agent MCP server (which is created at `task-agent-manager.ts` line 632-637 where `workspacePath` is already in scope). This injection point must be an explicit subtask.

---

### P2 — Should Fix

**[P2] CVE-2025-8022 citation is misleading — the advisory was withdrawn**
The plan cites CVE-2025-8022 as justification for using `Bun.spawn()` over `Bun.$`. The CVE was formally withdrawn by the CVE Numbering Authority (the described behaviour is argument injection in the calling application, not a Bun shell flaw). The avoidance of `Bun.$` for user-supplied script source is still correct security practice, but citing a withdrawn CVE in the plan document will cause confusion. Replace the citation with the actual reasoning: user-supplied content must never be interpolated into a shell string.

**[P2] No task for global concurrency cap on script executions**
The per-gate evaluation queue in Task 3.2 prevents overlapping evaluations for the same gate, but there is no cap on concurrent script processes across all gates in a run or across all runs. A space with many gates could trigger many simultaneous subprocesses. The plan should include either a semaphore / process pool, or explicitly document this as a known deferred limitation.

**[P2] "Lint Check" preset in Task 4.4 is broken by design**
The preset `bash` script `npm run lint 2>/dev/null; echo $?` outputs an exit code string (e.g., `"0"`) to stdout, which is not valid JSON. `parseJsonStdout("0")` returns `null` (a number is not a plain object), yielding `{ success: true, data: {} }` regardless of lint result — the gate always passes. The preset should rely on the exit code alone (drop the `echo $?`) or output proper JSON like `{ "passed": true }`.

**[P2] Per-gate evaluation queue may miss wake-up after data changes during script execution**
The Task 3.2 queue design ("if in-flight, await it instead of starting a new one") means a second `onGateDataChanged` call that arrives while the first script is running will await the first result — which used the data snapshot from before the second write. If the second write was the change that opens the gate, the node will never activate. The plan needs a "re-run if dirty" pattern (e.g., a pending-dirty flag per gate that triggers a re-evaluation after the in-flight script completes).

**[P2] Script execution errors are not surfaced in any frontend gate status UI**
When a script fails, `GateEvalResult.reason` carries the error (Task 3.1). There is no task to display this in `GateArtifactsView.tsx` or `ChannelEdgeConfigPanel.tsx`. Users configuring script checks need to see why a gate is blocked. Add a task or subtask to surface `reason` in the frontend.

---

### P3 — Nit / Suggestion

**[P3] Task 1.1 acceptance criterion for `computeGateDefaults(undefined)` contradicts current signature**
The criterion implies the function signature will change to accept `undefined`. Clarify whether the signature becomes `(fields?: GateField[])` and document that separately from just updating call sites.

**[P3] `maxBuffer` streaming enforcement needs implementation detail**
The plan says "kill process if stdout exceeds limit" but does not describe how to measure accumulated bytes during streaming (Bun's `new Response(proc.stdout).text()` buffers everything before checking size). Describe the streaming accumulation approach (chunk-by-chunk read with byte counter) to avoid a false-sense-of-safety implementation.

**[P3] E2E test (Task 5.2) may be flaky in CI if `python3` is absent**
The E2E test spawns interpreter subprocesses. `python3` may not be on PATH in all CI environments. Either restrict the test to `node` (guaranteed by the Bun runtime) or add a CI note that the test environment must have all three interpreters.

**[P3] Restricted env threat model not documented**
The plan strips credential prefixes (`ANTHROPIC_`, `CLAUDE_`, etc.) but does not address arbitrary credential env vars (`DATABASE_URL`, `SECRET_*`, custom keys). Document whether scripts are trusted (workspace-author-configured) or untrusted, and whether the current prefix-based strip is intentionally permissive or if a stricter allowlist is warranted.

---

### Missing Tasks (Summary)

1. Thread `workspacePath` into `ChannelRouterConfig` (prerequisite for Task 3.2).
2. Explicitly update all `computeGateDefaults` call sites after making `fields` optional (correction to Task 1.1 scope).
3. Construct and inject `GateScriptExecutor` in `task-agent-manager.ts` for the node-agent MCP server (prerequisite for Task 3.3).
4. Surface script error reason in frontend gate status UI (gap from Task 3.1).
5. Document or implement re-run-after-completion pattern in per-gate queue (gap in Task 3.2).

lsm (commented):

**Recommendation: APPROVE**

The v2 plan comprehensively addresses all P1 and P2 issues from both reviewers. The changelog at the bottom (lines 631-650) accurately reflects the changes made. Here is my verification of each resolved issue and the remaining observations.

### Resolved P1 issues (from v1 reviews)

All P1 issues are resolved:

- **`PairAggregate` and bidirectional edges** (GLM P1): Task 4.1 now explicitly updates `PairAggregate` with per-direction `lowToHigh*`/`highToLow*` fields and adds `reverseGateLabel`/`reverseGateColor`/`reverseHasScript` to `SemanticWorkflowEdge`. Collapse logic is in-scope.
- **`computeGateDefaults` call sites** (GLM P1): Task 1.1 subtask 6 enumerates all call sites (`channel-router.ts:594,728,741`, `node-agent-tools.ts:585,621`, `gate-data-repository.ts:112`, tests). The signature change to `fields?: GateField[]` makes existing call sites type-safe.
- **`workspacePath` not wired into `ChannelRouter`** (GLM P1): Task 3.2 adds it to `ChannelRouterConfig`. New Task 3.3 wires all 4 production `new ChannelRouter(...)` sites (verified: `task-agent-manager.ts:622,1440,1755`, `space-runtime-service.ts:309`).
- **`node-agent-tools.ts` async migration** (GLM P1): New Task 3.4 explicitly handles both call sites (lines 624 and 699), with clear notes about the `list_gates` vs `write_gate` distinction.
- **Storage serialization audit** (GLM P1): Not a separate task, but verified as a non-issue — gates are stored as `JSON.stringify(gates)` in `space-workflow-repository.ts` and parsed with `parseJson<Gate[]>`. Optional fields (`label`, `color`, `script`) naturally become `undefined` when absent from stored JSON. No code performs strict casts that would break.

### Resolved P2 issues (from v1 reviews)

All P2 issues are resolved:

- **`isChannelOpen` kept synchronous** (Opus P2-1): Task 3.1 subtask 6 explicitly keeps it sync with clear rationale.
- **`WorkflowCanvas.tsx` frontend caller** (Opus P2-2): Task 3.1 "Frontend note" correctly identifies that `WorkflowCanvas.tsx` has its own independent `evaluateGateStatus()` function that is sync, field-only, and not affected.
- **Per-gate queue dirty-flag pattern** (GLM P2): Task 3.2 subtask 5 describes coalescing + `gateDirtyFlags: Map<string, boolean>` re-run pattern. Well specified.
- **Global concurrency cap** (GLM P2): Task 3.2 subtasks 3-4 add a semaphore with default cap of 4.
- **Lint Check preset** (GLM P2 + Opus P3-4): Fixed in Task 4.4 — now outputs `{"passed":true}` / `{"passed":false}`.
- **Script errors not surfaced in UI** (GLM P2): New Task 4.6 addresses this.
- **CVE citation withdrawn** (GLM P2): Removed, replaced with plain-language security reasoning.
- **E2E Task 5.2 dependencies** (Opus P3-3): Now depends on 3.2, 3.3, 4.4, and 4.6.

### P2: Remaining gap

**P2-1: Task 4.6 does not specify the data transport for script error reasons to the frontend.**

The frontend currently evaluates gate status locally via its own `evaluateGateStatus()` in `WorkflowCanvas.tsx` using gate field data received through `space.gateData.updated` events. `GateEvalResult.reason` is computed on the backend but there is no described mechanism to deliver it to the frontend. Task 4.6 says it depends on Task 3.1 "for `reason` field in `GateEvalResult`" but `GateEvalResult` is a backend-only type — the frontend never receives it.

The plan should specify one of: (a) extend the `space.gateData.updated` event payload to include evaluation results (e.g., `{ gateData, evalResult?: { open, reason } }`), (b) add a new event like `space.gateEval.updated`, or (c) store the script result/error in the gate data itself so the frontend can read it from the existing data sync. This affects Tasks 4.5 and 4.6 scope.

### P3: Minor observations

**P3-1: Task 1.3 test file location.** The plan places persistence round-trip tests in `packages/daemon/tests/unit/space/gate-types-and-schema.test.ts`, which is a new file. These tests involve `SpaceWorkflowRepository` (a storage layer concern). Consider whether they belong in a storage test file or whether the plan's location is intentional as an integration-style unit test.

**P3-2: Task 3.5 test file for ChannelRouter.** The plan references `packages/daemon/tests/unit/space/channel-router.test.ts` but the existing file is at that path. The plan should note these are additions to the existing test file, not a new file.

**P3-3: `maxConcurrentScripts` not configurable at runtime.** The default of 4 is set in `ChannelRouterConfig`. There is no mechanism for users to change this. This is fine for v1 but worth noting for future work.

### Summary

The plan is thorough, well-structured, and addresses all prior feedback. The single remaining P2 (data transport for script error reasons) is a real gap but scoped to one task and straightforward to resolve. The plan is ready for implementation with that clarification.

lsm (commented):

Second review of PR #1190 (v2 plan). The planner addressed several first-round P1s well: Task 3.3 properly enumerates all four ChannelRouter construction sites, Task 3.2 adds the coalescing+dirty-flag pattern and global semaphore, Task 4.6 surfaces script error reasons, and the `computeGateDefaults` signature fix is explicit. However, there are still significant gaps that would cause runtime failures if the implementation follows the plan as written.

---

### P1 — `fields?: GateField[]` breaks far more callers than the plan acknowledges

**Task 1.1 subtask 6** lists only `computeGateDefaults` call sites to verify. But `gate.fields` is accessed directly (without null-guard) in at least these locations:

**Frontend (would throw `TypeError: Cannot read properties of undefined (reading 'length'/'map')`):**
- `packages/web/src/components/space/WorkflowCanvas.tsx` line 124: `gate.fields.length === 0` — this is the frontend `evaluateGateStatus()` the plan claims is "unaffected". It is affected: any script-only gate (no `fields`) passed to `evaluateGateStatus` would crash immediately.
- `WorkflowCanvas.tsx` lines 127, 131, 142, 1210, 1212 — same function and callers
- `GateEditorPanel.tsx` lines 41, 53, 54, 58, 71, 81, 149, 152 — spread and map calls on `gate.fields`
- `ChannelEdgeConfigPanel.tsx` lines 173, 177
- `VisualWorkflowEditor.tsx` lines 124, 1013
- `semanticWorkflowGraph.ts` line 59: `const fields = gate.fields; if (fields.length === 0) return 'check';`

**Backend (would throw at runtime):**
- `packages/daemon/src/lib/space/runtime/gate-evaluator.ts` line 171: `for (const field of gate.fields)` — the core evaluation loop. Script-only gates would crash here too.
- `packages/daemon/src/lib/space/agents/task-agent.ts` line 138: `gate.fields.map(...)` in `formatGate()`
- `packages/daemon/src/lib/space/tools/node-agent-tools.ts` lines 498, 583, 629, 670, 678 (see P1 below)

The plan must add a task (or extend 1.1) to add `?? []` guards at every bare `gate.fields` access. The fix for `computeGateDefaults` accepting `undefined` is necessary but far from sufficient.

---

### P1 — `write_gate` field authorization crashes for script-only gates

`packages/daemon/src/lib/space/tools/node-agent-tools.ts` line 670:
```ts
const fieldMap = new Map(gateDef.fields.map((f) => [f.name, f]));
```
When `gate.fields` is `undefined` (script-only gate), this throws before any field iteration happens. Line 678 has the same issue. Task 3.4 mentions updating lines 624 and 699 to `await`, but does not mention guarding lines 670 and 678. A script-only gate would be completely unusable via `write_gate`.

---

### P1 — `workspacePath` is not in scope in `SpaceRuntimeService.notifyGateDataChanged`

Task 3.3 subtask 4 says "workspace path is available via `space.workspacePath`" for `space-runtime-service.ts` line 309. That method's signature is:
```ts
async notifyGateDataChanged(runId: string, gateId: string): Promise<SpaceTask[]>
```
There is no space object, no spaceId, and no workspacePath in scope. To get the workspace path the implementation would need to look up the run by `runId`, then look up the space from the run's `spaceId`. The plan does not describe this lookup chain. As written, this wiring point is incomplete.

---

### P1 — `workspacePath` is not in scope in `buildNodeAgentMcpServerForSession`

Task 3.3 subtask 3 says "workspacePath is available via the space object" at `task-agent-manager.ts` line 1755. However, the enclosing private method is:
```ts
private buildNodeAgentMcpServerForSession(
  taskId, subSessionId, role, spaceId, workflowRunId, stepTaskId, taskManager
)
```
No workspace path or space object is passed in. `this.config` does not include a `spaceRepo` or similar accessor that would provide workspace path from spaceId in the existing code. Wiring this correctly requires either passing `workspacePath` as a new parameter to the method or doing a repo lookup (which needs to be described explicitly).

---

### P2 — Dirty-flag map must be keyed by `${runId}:${gateId}`, not just `gateId`

Task 3.2 describes:
```
gateDirtyFlags: Map<string, boolean>
gateEvaluations: Map<string, Promise<GateEvalResult>>
```
But `evaluateGateById` takes both `runId` and `gateId`. Two concurrent workflow runs that happen to share the same gateId (same workflow definition, different run instances) would incorrectly share dirty-flag state and in-flight promise state. The map key must be a composite `${runId}:${gateId}`. The plan should specify this explicitly to prevent implementer error.

---

### P2 — `validateGate` empty-fields rule is a breaking change for existing stored gates

Task 1.2 specifies: `validateGate({ id: 'g1', fields: [], resetOnCycle: false })` returns errors (empty fields, no script). If this validator is applied to existing stored gates at load time or on any update, existing gates with `fields: []` (which are currently valid and may exist in production data) would silently become invalid. The plan needs to specify where this validation is called and confirm it is only applied to new/updated gates, not to existing data on load.

---

### P3 — Task 3.4 `list_gates` path: note the silent difference

Task 3.4 says `list_gates` (line 624) has "no scriptExecutor" so it falls back to sync field evaluation. This is intentional and fine, but the plan should note that for a script-only gate (no fields), `evaluateGate` with no scriptExecutor will return `{ open: true }` (empty field loop returns open). This could confuse users who see the gate "open" in the listing but the gate blocks in practice. Worth calling out in the task.

---

### Summary of P1 changes still needed before implementation

1. Extend Task 1.1 (or add a new task) to add `?? []` fallback guards at every bare `gate.fields` access across all files listed above.
2. Fix Task 3.4 to also guard `node-agent-tools.ts` lines 670 and 678 for script-only gates.
3. Fix Task 3.3 to describe the `runId → run → spaceId → space.workspacePath` lookup for `SpaceRuntimeService.notifyGateDataChanged`.
4. Fix Task 3.3 to describe how `workspacePath` reaches `buildNodeAgentMcpServerForSession` (new parameter or repo lookup).
5. Fix Task 3.2 to specify composite `${runId}:${gateId}` keys for the dirty-flag and coalescing maps.

**Recommendation: REQUEST_CHANGES** — the `fields` optional change is still significantly under-specified and would cause runtime crashes across the codebase as written.

lsm (commented):

**Recommendation: APPROVE**

The v3 plan successfully resolves all P1 and P2 issues from the v2 review. The new Task 1.4 is comprehensive, the composite `${runId}:${gateId}` key prevents cross-run leakage, the `_scriptResult` data transport mechanism is well-specified, and the `workspacePath` lookup chains are described for all construction sites. I verified all major claims against the codebase. Three minor accuracy issues remain.

### Verification of resolved issues

All prior P1s are resolved:

- **Task 1.4 (new)**: Properly audits `gate.fields` access across 6 frontend files and 3 backend files. The line references are mostly accurate (see P3 notes below). Acceptance criteria include a grep-based verification step, which is good.
- **Task 3.3 subtask 3**: `buildNodeAgentMcpServerForSession` workspace path resolution is described. However, see P3-1 below regarding the method name.
- **Task 3.3 subtask 4**: The lookup chain `runId -> workflowRunRepo.getRun() -> run.spaceId -> spaceManager.getSpace() -> space.workspacePath` is correct. I confirmed both `this.config.spaceManager` and `this.config.workflowRunRepo` are available on `SpaceRuntimeService`.
- **Task 3.2 subtask 5**: Composite `${runId}:${gateId}` key is explicit in both `gateEvaluations` and `gateDirtyFlags` maps.
- **Task 4.6**: `_scriptResult` stored in gate data, flowing through existing `space.gateData.updated` event. Clean solution that avoids a new event type.
- **Task 1.2**: Clarified as creation-time-only validator. Existing gates with `fields: []` remain valid.
- **Task 3.4**: Lines 670/678 guarded, `list_gates` limitation documented.

### P3: Minor accuracy issues

**P3-1: `this.getWorkspacePathForTask(taskId)` does not exist.**
Task 3.3 subtask 3 refers to `this.getWorkspacePathForTask(taskId)` as "an existing private method on `TaskAgentManager`, around line 1370-1425". No such method exists. Lines 1410-1425 contain a local IIFE `rehydrateWorkspacePath` inside `rehydrate()`, not a reusable method. The implementation will need to either (a) extract a private method from the existing rehydration logic, or (b) look up workspace path via `this.taskWorktreePaths.get(taskId) ?? space.workspacePath` directly. This is not a plan-level blocker since the data is accessible, but the plan text is inaccurate about how.

**P3-2: Task 1.4 lists `ChannelEdgeConfigPanel.tsx` "lines 173, 177" but that file has zero `gate.fields` accesses.**
Grep confirms no matches. This is a phantom entry in the audit. Harmless -- the implementer will find nothing to change there -- but it suggests the audit was not fully verified against the codebase.

**P3-3: Task 1.4 is missing two files.**
- `WorkflowEditor.tsx` line 606: `fields: [...gate.fields]` -- this will crash for script-only gates. This file is not in the Task 1.4 file list.
- `WorkflowCanvas.tsx` lines 1210, 1212: `isHumanApprovalGate(gate.fields)` and `computeVoteCount(gate.fields, gateData)` -- these are in the same file as lines 124/127/131/142 but are NOT called out specifically in the task. The implementer should catch these via the grep-based acceptance criteria (`grep -rn 'gate\.fields\b' packages/`), but they should be listed explicitly.

### Summary

The plan is thorough and ready for implementation. All prior blocking issues are resolved. The three P3 items are minor inaccuracies that the implementer can self-correct during the grep-based verification step in Task 1.4's acceptance criteria.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (own-PR review — using COMMENT event)

---

### v2 P1 Fixes — Verification Results

#### Fix 1: Task 1.4 — `gate.fields` safety guards (P1 from v2)

**Verdict: Partially resolved — 3 access sites still missing from the list.**

Task 1.4 is a real improvement and the rationale is correct. However, the exhaustive list in Task 1.4 still omits access sites that will crash at runtime for script-only gates:

**Missing from Task 1.4:**

1. `packages/web/src/components/space/WorkflowCanvas.tsx` **lines 1210 and 1212** — both call helper functions that accept `GateField[]` (not optional):
   - Line 1210: `isHumanApprovalGate(gate.fields)` — will crash if `gate.fields` is `undefined`
   - Line 1212: `computeVoteCount(gate.fields, gateData)` — will crash if `gate.fields` is `undefined`
   - These are **not** in the lines 124/127/131/142 already listed

2. `packages/web/src/components/space/WorkflowEditor.tsx` **line 606** — `fields: [...gate.fields]` will crash on spread of `undefined`. `WorkflowEditor.tsx` is **entirely absent** from Task 1.4's file list.

The acceptance criterion `grep -rn 'gate\.fields\b' packages/` will still return results after Task 1.4 is applied at the listed sites — a passing coder can satisfy the exact grep and still miss these two files. The task needs an updated comprehensive list or the grep criterion needs to cover these files explicitly.

---

#### Fix 2: Task 3.3 subtask 3 — `buildNodeAgentMcpServerForSession` workspacePath (P1 from v2)

**Verdict: Method name is incorrect — the referenced method does not exist.**

The plan says:
> Look up the workspace path via `this.getWorkspacePathForTask(taskId)` (an existing private method on `TaskAgentManager`, around line 1370-1425)

The actual method is named **`getTaskWorktreePath(taskId)`** (public, not private, at line 863). It is NOT called `getWorkspacePathForTask`. Furthermore:

- `getTaskWorktreePath` returns `string | undefined` — it does **not** fall back to `space.workspacePath` when no worktree exists.
- If a task uses the space's base workspace (no worktree), `getTaskWorktreePath(taskId)` returns `undefined`, and the script would run from `process.cwd()` (the daemon working directory) — silently wrong.

The correct solution would be to either (a) add a new helper that mirrors the `rehydrateWorkspacePath` IIFE pattern (worktree path if exists, else `space.workspacePath`), or (b) look up the task's spaceId via the existing `getSpaceIdForTask(taskId)` private method, then fetch `space.workspacePath`. The plan's subtask 3 as written points coders at a non-existent method name, which is a meaningful omission that will cause implementation confusion.

---

#### Fix 3: Task 3.3 subtask 4 — `notifyGateDataChanged` lookup chain (P1 from v2)

**Verdict: Resolved.** The lookup chain (`runId → workflowRunRepo.getRun(runId) → run.spaceId → spaceManager.getSpace(spaceId) → space.workspacePath`) is valid. `SpaceRuntimeServiceConfig` has both `spaceManager` and `workflowRunRepo` in scope. `getRun` is synchronous and `getSpace` is async (already fine since `notifyGateDataChanged` is `async`). This fix is sound.

---

### Remaining Issues

**P1 — Missing access sites in Task 1.4:**
`WorkflowCanvas.tsx:1210-1212` and `WorkflowEditor.tsx:606` must be added to the guard list. The `grep` acceptance criterion will not catch these since they're in files not targeted by the task.

**P1 — Incorrect method name in Task 3.3 subtask 3:**
`getWorkspacePathForTask` does not exist. The existing method is `getTaskWorktreePath`, which returns `undefined` (no `space.workspacePath` fallback) for tasks running on the base workspace. The subtask needs to either reference the correct existing method and document the fallback, or specify adding a helper that provides the full lookup (worktree path ?? space.workspacePath).

---

### No New Issues Found

The rest of the v3 plan is well-specified. The `_scriptResult` transport mechanism (Task 4.6), the composite `${runId}:${gateId}` key (Task 3.2), the `list_gates` known-limitation note (Task 3.4), and the creation-time-only validation clarification (Task 1.2) all look correct.

lsm (commented):

**Recommendation: APPROVE**

v4 of this plan addresses all previously raised issues. I verified each fix against the current codebase:

**Previously missing `gate.fields` access sites (P1 from v3) -- now fixed:**
- `WorkflowCanvas.tsx` lines 1210 and 1212 (`isHumanApprovalGate(gate.fields)` and `computeVoteCount(gate.fields, gateData)`) are now listed in Task 1.4.
- `WorkflowEditor.tsx` line 606 (`fields: [...gate.fields]` spread) is now listed in Task 1.4.
- Phantom `ChannelEdgeConfigPanel.tsx` entry correctly removed (confirmed: no `gate.fields` accesses in that file).

**`getTaskWorktreePath` method name (P3 from v3) -- now fixed:**
- Task 3.3 subtask 3 correctly references `getTaskWorktreePath(taskId)` (line 862, returns `string | undefined`).
- The plan correctly notes it does NOT fall back to `space.workspacePath` automatically -- the explicit fallback via `spaceId` lookup is documented.

**Full `gate.fields` audit verified against codebase:**
I ran a comprehensive grep across `packages/` for `gate\.fields` and `gateDef\.fields`. All application-code access sites (frontend + backend) are covered by Task 1.4. Test files that construct gates with explicit `fields` arrays are appropriately not included (they define their own fixtures with non-undefined fields).

**No new gaps or regressions found in v4.** The changelog sections ("Changes from v2" and "Changes from v3") accurately describe all modifications.

lsm (commented):

**Recommendation: APPROVE**

Both P1 issues from the v3 review are correctly resolved in v4.

### Fix 1: Task 1.4 gate.fields audit list

The plan now lists `WorkflowCanvas.tsx` lines 1210 and 1212 (`isHumanApprovalGate(gate.fields)` and `computeVoteCount(gate.fields, gateData)`) and `WorkflowEditor.tsx` line 606 (`fields: [...gate.fields]` spread). Verified against source: both access sites exist at exactly those line numbers. The phantom `ChannelEdgeConfigPanel.tsx` entry is removed from the Task 1.4 guard list (the file still appears in Tasks 4.5 and 4.6 for new UI work, which is correct — it has no existing `gate.fields` accesses to guard).

### Fix 2: Task 3.3 subtask 3 method reference

The plan now correctly names `getTaskWorktreePath(taskId)` (line 862, public method, returns `string | undefined`), replacing the previously referenced nonexistent `getWorkspacePathForTask`. Verified in source: method exists at line 862, returns `this.taskWorktreePaths.get(taskId)`. The plan also correctly documents that this method does NOT auto-fallback to `space.workspacePath` — the explicit fallback chain `this.getTaskWorktreePath(taskId) ?? this.config.spaceManager.getSpace(spaceId).workspacePath` is valid because `spaceId` is a parameter of `buildNodeAgentMcpServerForSession` and `spaceManager` is confirmed present in the config at line 139.

### Changelog classification note (non-blocking)

The v3 changelog classifies the `getWorkspacePathForTask` → `getTaskWorktreePath` fix as P3 ("Task 3.3 subtask 3: Fixed incorrect method name"). The previous review flagged it as P1 because a broken method reference would cause a TypeScript compile error at the implementation site. Reclassifying a broken reference as P3 is generous but the fix itself is complete and sound, so no further iteration is needed.

No new issues found in v4.

PR #1191, opened 2026-04-02
fix: defensive JSON parsing for schedule/structuredMetrics in goals

lsm:
Fix daemon crash when `schedule` column in goals table contains a raw cron
string (e.g. `@daily`) instead of the expected JSON object.

## Changes

- **`shared/src/utils.ts`**: Add `parseJson`/`parseJsonOptional` safe parsing utilities
- **`daemon/src/lib/rpc-handlers/live-query-handlers.ts`**: Use `parseJson`/`parseJsonOptional` in `mapGoalRow` for `linkedTaskIds`, `metrics`, `structuredMetrics`, `schedule`
- **`daemon/src/storage/repositories/goal-repository.ts`**: Use `parseJsonOptional` in `rowToGoal` for `structuredMetrics`, `schedule`
- **`daemon/src/storage/schema/migrations.ts`**: Add migration 71 to fix corrupted schedule values at startup (wraps bare cron strings into proper `CronSchedule` shape)

## Tests

- `shared/tests/utils.test.ts` — 10 tests for `parseJson`/`parseJsonOptional`
- `daemon/tests/unit/rpc-handlers/live-query-handlers.test.ts` — 5 tests for `mapGoalRow` corrupted data
- `daemon/tests/unit/room/goal-repository-v2.test.ts` — 4 tests for `rowToGoal` corrupted data
- `daemon/tests/unit/storage/migrations/migration-71_test.ts` — 9 tests for schedule migration

### DIFF
--- a/packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts
+++ b/packages/daemon/src/lib/rpc-handlers/live-query-handlers.ts
@@ -9,7 +9,7 @@
 
 import { Database as BunDatabase } from 'bun:sqlite';
 import type { MessageHub } from '@neokai/shared';
-import { createEventMessage } from '@neokai/shared';
+import { createEventMessage, parseJson, parseJsonOptional } from '@neokai/shared';
 import type {
 	LiveQuerySubscribeRequest,
 	LiveQuerySubscribeResponse,
@@ -76,13 +76,10 @@ function mapTaskRow(row: Record<string, unknown>): Record<string, unknown> {
 function mapGoalRow(row: Record<string, unknown>): Record<string, unknown> {
 	return {
 		...row,
-		linkedTaskIds: JSON.parse((row.linkedTaskIds as string | null) ?? '[]') as string[],
-		metrics: JSON.parse((row.metrics as string | null) ?? '{}') as Record<string, number>,
-		structuredMetrics:
-			row.structuredMetrics != null
-				? (JSON.parse(row.structuredMetrics as string) as unknown[])
-				: undefined,
-		schedule: row.schedule != null ? (JSON.parse(row.schedule as string) as unknown) : undefined,
+		linkedTaskIds: parseJson<string[]>((row.linkedTaskIds as string | null) ?? '[]', []),
+		metrics: parseJson<Record<string, number>>((row.metrics as string | null) ?? '{}', {}),
+		structuredMetrics: parseJsonOptional(row.structuredMetrics as string | null),
+		schedule: parseJsonOptional(row.schedule as string | null),
 		schedulePaused: row.schedulePaused === 1,
 	};
 }
--- a/packages/daemon/src/storage/repositories/goal-repository.ts
+++ b/packages/daemon/src/storage/repositories/goal-repository.ts
@@ -6,7 +6,7 @@
  */
 
 import type { Database as BunDatabase } from 'bun:sqlite';
-import { generateUUID, parseJsonOptional } from '@neokai/shared';
+import { generateUUID, parseJson, parseJsonOptional } from '@neokai/shared';
 import type {
 	RoomGoal,
 	GoalStatus,
@@ -654,8 +654,8 @@ export class GoalRepository {
 			status: row.status as GoalStatus,
 			priority: row.priority as GoalPriority,
 			progress: row.progress as number,
-			linkedTaskIds: JSON.parse(row.linked_task_ids as string) as string[],
-			metrics: JSON.parse(row.metrics as string) as Record<string, number>,
+			linkedTaskIds: parseJson<string[]>((row.linked_task_ids as string | null) ?? '[]', []),
+			metrics: parseJson<Record<string, number>>((row.metrics as string | null) ?? '{}', {}),
 			planning_attempts: (row.planning_attempts as number | null) ?? 0,
 			goal_review_attempts: (row.goal_review_attempts as number | null) ?? 0,
 			createdAt: row.created_at as number,
--- a/packages/daemon/src/storage/repositories/goal-repository.ts
+++ b/packages/daemon/src/storage/repositories/goal-repository.ts
@@ -6,7 +6,7 @@
  */
 
 import type { Database as BunDatabase } from 'bun:sqlite';
-import { generateUUID } from '@neokai/shared';
+import { generateUUID, parseJsonOptional } from '@neokai/shared';
 import type {
 	RoomGoal,
 	GoalStatus,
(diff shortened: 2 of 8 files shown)


lsm (commented):

### Summary

Clean, well-scoped fix for the JSON parse crash on the `schedule` field. The approach is sound: add defensive parsing utilities, apply them at both parsing sites (live-query handler and goal repository), and add a data migration to fix corrupted rows at startup. Test coverage is thorough.

### Findings

**P3: Orphaned JSDoc comment in `packages/shared/src/utils.ts` (lines 22-25)**

The `parseJson` and `parseJsonOptional` functions were inserted between the existing `generateUUID` JSDoc block and the `generateUUID` function itself. Lines 22-25 now contain an orphaned JSDoc comment that describes `generateUUID` but sits above `parseJson` instead. The `generateUUID` function at line 52 has no JSDoc. Either move the JSDoc back above `generateUUID` or delete it since `generateUUID` is self-documenting.

### What looks good

- `parseJson` / `parseJsonOptional` are simple, well-typed utilities correctly placed in the shared package and properly re-exported via `mod.ts`.
- Migration 71 handles all edge cases: bare cron strings, JSON-quoted strings, already-valid objects, null values, missing table/column. Idempotency is verified by test.
- Both parsing sites (`mapGoalRow` in live-query-handlers and `rowToGoal` in goal-repository) are updated consistently.
- No remaining `JSON.parse` calls on `schedule` or `structuredMetrics` outside the migration itself.
- Test coverage is comprehensive: 199-line migration test, repository tests with corrupted data, live-query handler tests with raw SQL inserts.

**Recommendation: APPROVE** (only P3 nit found)

lsm (commented):

**Recommendation: APPROVE** (same-author review posted as COMMENT)

The fix is correct and well-scoped. The daemon crash is addressed at all three read paths (live-query mapper, repository mapper, startup migration). Tests cover the happy path, null, and corrupted-data cases cleanly. A few issues worth noting before merge:

---

### P1 — Inconsistency between the two goal mappers on `linkedTaskIds` / `metrics`

**`mapGoalRow` in `live-query-handlers.ts`** now uses `parseJson` defensively for `linkedTaskIds` and `metrics`:
```ts
// line 79-80
linkedTaskIds: parseJson<string[]>((row.linkedTaskIds as string | null) ?? '[]', []),
metrics: parseJson<Record<string, number>>((row.metrics as string | null) ?? '{}', {}),
```

**`rowToGoal` in `goal-repository.ts`** still uses the undefended bare `JSON.parse` for the same columns:
```ts
// line 657-658
linkedTaskIds: JSON.parse(row.linked_task_ids as string) as string[],
metrics: JSON.parse(row.metrics as string) as Record<string, number>,
```

The schema defaults (`DEFAULT '[]'` / `DEFAULT '{}'`) make corruption unlikely for these columns on normal data flows, but the asymmetry is a footgun: any future migration or direct DB write that leaves a corrupted value in `linked_task_ids` / `metrics` would crash the repository but not the live-query path. The PR description claims only V2 fields needed fixing, but the live-query handler was also updated for `linkedTaskIds`/`metrics` — so `rowToGoal` should follow the same pattern to keep both mappers consistent.

---

### P2 — Orphaned JSDoc comment in `packages/shared/src/utils.ts`

The new functions were inserted between an existing JSDoc and its target function, leaving the comment for `generateUUID` pointing at `parseJson` instead:

```ts
// lines 22-30 in utils.ts
/**
 * Generate a UUID v4 (browser and Node.js compatible)    ← now orphaned
 * Uses crypto.randomUUID() if available, otherwise falls back to a polyfill
 */
/**
 * Safely parse a JSON string, returning a fallback value on failure.
 * ...
 */
export function parseJson<T>(...) { ... }   ← attached to this, not generateUUID
```

The orphaned comment should be moved below `parseJsonOptional` and placed directly above `generateUUID`. The issue is cosmetic but affects IDE hover documentation and is easy to fix.

---

### P3 — Migration test uses a file-based DB where in-memory suffices

`packages/daemon/tests/unit/storage/migrations/migration-71_test.ts` creates a real temp directory and a file-backed DB in `beforeEach`, then cleans up in `afterEach`. Every other migration test in the same directory (e.g. `migration-48_test.ts`) uses `new BunDatabase(':memory:')`, which is simpler, faster, and does not require FS cleanup. The file-based approach adds noise and could leave stale `tmp/` directories on test failure. Consider switching to in-memory.

---

### Correctness summary

- `parseJson` / `parseJsonOptional` implementations are correct and simple.
- Migration 71 correctly handles the four cases: bare cron string (JSON parse failure), JSON-quoted string, valid object shape, and null — and is idempotent.
- The `tableExists` + `schedule column exists` guard prevents the migration from crashing on fresh or legacy DBs.
- Tests cover all meaningful edge cases for the migration and both mappers.
- `mapTaskRow`'s remaining bare `JSON.parse` for `dependsOn` (line 55) is pre-existing and outside the scope of this PR; the column has `DEFAULT '[]'` so it is low risk.

lsm (commented):

### Re-review: All feedback items addressed

Verified the three fixes from the latest commit (ac1c7df):

1. **rowToGoal consistency (was P1)** -- Fixed. `linkedTaskIds` and `metrics` in `goal-repository.ts` now use `parseJson` with fallback defaults, matching `mapGoalRow` in `live-query-handlers.ts` exactly.

2. **Orphaned JSDoc (was P2)** -- Fixed. The `generateUUID` doc comment was moved back above its function definition. `parseJson` and `parseJsonOptional` each have their own correct JSDoc.

3. **Migration test in-memory DB (was P3)** -- Fixed. `migration-71_test.ts` now uses `new BunDatabase(':memory:')` instead of file-based DB with manual cleanup.

No remaining bare `JSON.parse` calls on `schedule`, `structuredMetrics`, `linkedTaskIds`, or `metrics` in application code (only the migration itself uses `JSON.parse` to inspect schedule shape, which is correct).

**Recommendation: APPROVE** -- all prior feedback addressed cleanly, no new issues found.

lsm (commented):

**Recommendation: APPROVE**

All three items from the previous review round have been correctly addressed. The implementation is sound.

---

### Verified fixes

**1. rowToGoal consistency with mapGoalRow (goal-repository.ts lines 657-671)**

Both `structuredMetrics` and `schedule` now use `parseJsonOptional`, and both `linkedTaskIds` and `metrics` now use `parseJson` with appropriate fallbacks. The approach is consistent with `mapGoalRow` in `live-query-handlers.ts`.

One minor note: `linkedTaskIds` and `metrics` in `rowToGoal` apply `?? '[]'` / `?? '{}'` before passing to `parseJson`, while the equivalent fields in `mapGoalRow` do the same. This double-guarding against `null` is harmless (the `parseJson` fallback already handles `null`), but it is not wrong.

**2. JSDoc comment (utils.ts)**

The orphaned JSDoc issue is resolved. All three JSDoc comments are cleanly attached to their respective functions: `parseJson` (line 22), `parseJsonOptional` (line 36), and `generateUUID` (line 49).

**3. Migration test uses in-memory DB (migration-71_test.ts)**

The migration test creates a `new BunDatabase(':memory:')` in `beforeEach` and closes it in `afterEach`. No file-system DB is used. The `PRAGMA foreign_keys = OFF` is a reasonable defensive measure since the minimal schema omits the `rooms` FK.

---

**rowToExecution still uses bare JSON.parse (goal-repository.ts line 692)**

`taskIds: JSON.parse(row.task_ids as string)` is not guarded. This is pre-existing and out of scope for this PR — `task_ids` is always written as valid JSON by `insertExecution` and `updateExecution`. However, it is the only remaining raw `JSON.parse` call in this file and could be hardened with `parseJson(..., [])` in a follow-up.

**Migration 71 always writes UTC timezone for bare cron strings**

When a corrupted row like `@daily` is fixed, it is rewritten as `{"expression":"@daily","timezone":"UTC"}`. This is a safe default. A row that originally had a different timezone is already broken — the fix is the best possible recovery.

**No test for linked_task_ids or metrics defensive parsing in rowToGoal**

The four new `goal-repository-v2.test.ts` defensive-parsing tests cover `schedule` and `structuredMetrics`, but not `linkedTaskIds` or `metrics`. These fields are `NOT NULL DEFAULT '[]'/'{}' ` in the schema, so corruption is less likely — this is acceptable.

---

The implementation is correct, complete, and well-tested. The crash-on-corrupted-schedule bug is fixed at all three layers (live-query read path, repository read path, startup migration).

PR #1194, opened 2026-04-02
feat: add gate validation for color, label, script, and top-level gate

lsm:
## Summary
- Add `validateGateColor()`, `validateGateLabel()`, `validateGateScript()`, and `validateGate()` to `gate-evaluator.ts`
- Top-level `validateGate()` enforces at least one of `fields` (non-empty) or `script` is present
- Color validates `#rrggbb` hex format; label max 20 chars; script interpreter in `[bash, node, python3]`, non-empty source, timeoutMs 1–120000
- 61 new unit tests covering all validators and edge cases

## Test plan
- [x] Unit tests: 91 total (30 existing + 61 new) pass
- [x] Typecheck passes
- [x] Lint/format passes
- [x] Knip passes

### DIFF
--- a/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
+++ b/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
@@ -176,7 +176,7 @@ export function validateGateScript(script: unknown): string[] {
 		return errors;
 	}
 
-	if (!script || typeof script !== 'object') {
+	if (typeof script !== 'object') {
 		errors.push(`script: expected object, got ${typeof script}`);
 		return errors;
 	}
@@ -211,12 +211,14 @@ export function validateGateScript(script: unknown): string[] {
  *
  * This validator enforces structural rules on new/updated gates:
  *   - At least one of `fields` (non-empty) or `script` must be present.
- *   - When `fields` is present and non-empty, existing field validation runs.
+ *   - When `fields` is present, existing field validation runs (handles non-array gracefully).
  *   - Optional `color`, `label`, and `script` are validated when provided.
  *
- * **Important:** This validator is only applied to new/updated gates (via
- * `GateEditorPanel` and MCP tool handlers), never to existing gates loaded
- * from storage. Existing gates with `fields: []` remain valid at runtime.
+ * **Important:** This validator should only be applied to new/updated gates
+ * (e.g. via `GateEditorPanel` or MCP tool handlers), never to existing gates
+ * loaded from storage. Existing gates with `fields: []` remain valid at runtime.
+ *
+ * TODO: Wire into RPC handlers and GateEditorPanel in follow-on tasks.
  *
  * @param gate  The gate object to validate.
  * @returns Array of human-readable error strings. Empty array = valid.
@@ -231,29 +233,17 @@ export function validateGate(gate: unknown): string[] {
 
 	const g = gate as Record<string, unknown>;
 
-	// Validate color
-	if (g.color !== undefined) {
-		errors.push(...validateGateColor(g.color));
-	}
+	// Validate optional fields — sub-validators handle null/undefined gracefully
+	errors.push(...validateGateColor(g.color));
+	errors.push(...validateGateLabel(g.label));
+	errors.push(...validateGateScript(g.script));
 
-	// Validate label
-	if (g.label !== undefined) {
-		errors.push(...validateGateLabel(g.label));
-	}
-
-	// Validate script
-	if (g.script !== undefined) {
-		errors.push(...validateGateScript(g.script));
-	}
-
-	// Validate fields when present
+	// Validate fields when present (validateGateFields handles non-array gracefully)
 	if (g.fields !== undefined && g.fields !== null) {
-		if (Array.isArray(g.fields) && g.fields.length > 0) {
-			errors.push(...validateGateFields(g.fields));
-		}
+		errors.push(...validateGateFields(g.fields));
 	}
 
-	// At least one of fields (non-empty) or script must be present
+	// At least one of fields (non-empty array) or script must be present
 	const hasFields = Array.isArray(g.fields) && g.fields.length > 0;
 	const hasScript = g.script !== undefined && g.script !== null;
 	if (!hasFields && !hasScript) {
--- a/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
+++ b/packages/daemon/src/lib/space/runtime/gate-evaluator.ts
@@ -107,6 +107,163 @@ export function validateGateFields(fields: unknown, path = 'fields'): string[] {
 }
 
 // ---------------------------------------------------------------------------
+// Gate creation/modification validators
+// ---------------------------------------------------------------------------
+
+const HEX_COLOR_RE = /^#[0-9a-fA-F]{6}$/;
+const VALID_INTERPRETERS = new Set(['bash', 'node', 'python3']);
+
+/**
+ * Validates a gate badge color.
+ *
+ * @param color  The color value to validate.
+ * @returns Array of human-readable error strings. Empty array = valid.
+ */
+export function validateGateColor(color: unknown): string[] {
+	const errors: string[] = [];
+
+	if (color === undefined || color === null) {
+		return errors;
+	}
+
+	if (typeof color !== 'string') {
+		errors.push(`color: expected string, got ${typeof color}`);
+		return errors;
+	}
+
+	if (!HEX_COLOR_RE.test(color)) {
+		errors.push(`color: expected hex format #rrggbb, got "${color}"`);
+	}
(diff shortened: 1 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean, well-structured implementation. The four validators are consistent in style, handle edge cases properly, and the 61 tests provide thorough coverage. A few observations:

### P3: Minor type comment mismatch

The `Gate.color` field in `packages/shared/src/types/space.ts:597` is documented as `"CSS color value"` but `validateGateColor` only accepts `#rrggbb` hex format. This is actually fine as a stricter validation choice (hex-only is safer for consistent rendering), but the interface comment should eventually be updated to say "hex color" rather than "CSS color value" to avoid confusion. Not blocking since the type itself is `string` and the validator is authoritative.

### P3: Redundant falsy check in validateGateScript

At `packages/daemon/src/lib/space/runtime/gate-evaluator.ts:179`, the `!script` guard is redundant since `null` and `undefined` are already handled by the early return on lines 175-177. The remaining falsy values (`0`, `""`, `false`) would fail the `typeof script !== 'object'` check anyway. Harmless but unnecessary.

### Notes (no issues)

- The "at least one of fields or script" check correctly uses structural presence (not content validity) — a gate with an invalid script still passes the presence check, with content errors reported separately. This is the right design.
- All sub-validators correctly treat `null`/`undefined` as "optional, not provided" and return no errors, consistent with the optional fields in the `Gate` interface.
- The `validateGate` function is not yet wired into any RPC handler or editor panel. This is expected — the task scope was to add the validation functions themselves.
- Tests pass cleanly (91 tests, 129 assertions, 23ms).

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P1, 2 P2)

The implementation is clean, well-documented, and test coverage is thorough. However, there are two meaningful issues that should be addressed before merging.

---

### P1 — Dead branch in `validateGateScript` (gate-evaluator.ts:179)

The guard on line 179 is unreachable:

```ts
if (script === undefined || script === null) {
    return errors;  // <-- early return for null/undefined
}

if (!script || typeof script !== 'object') {  // <-- !script can never be true here
```

After the early return at line 175–177, `script` is guaranteed to be neither `undefined` nor `null`. The `!script` half of the line-179 guard is therefore dead code. The `typeof script !== 'object'` half still works correctly (catches strings, numbers, booleans), but the `!script` check is misleading — it implies `false`/`0`/`""` could reach here, which they cannot. This should be `typeof script !== 'object'` only, to match the intent and avoid confusing future readers.

`packages/daemon/src/lib/space/runtime/gate-evaluator.ts`, line 179

---

### P1 — `validateGate` silently ignores a non-array `fields` value

When `gate.fields` is present but not an array (e.g., `fields: "bad"` or `fields: 42`), lines 250–254 skip validation entirely:

```ts
if (g.fields !== undefined && g.fields !== null) {
    if (Array.isArray(g.fields) && g.fields.length > 0) {
        errors.push(...validateGateFields(g.fields));
    }
}
```

The caller receives only the "at least one" error (because `hasFields` on line 257 is false for non-arrays), with no indication that `fields` was the wrong type. Any future consumer relying on the validator to catch a badly typed `fields` will get a misleading error message. The outer guard should call `validateGateFields(g.fields)` unconditionally when fields is present, because `validateGateFields` already handles non-array input gracefully (returns `["fields: expected an array, got ..."]`).

`packages/daemon/src/lib/space/runtime/gate-evaluator.ts`, lines 250–254

---

### P2 — Redundant `!== undefined` guards in `validateGate`

`validateGate` guards each optional field before delegating:

```ts
if (g.color !== undefined) {
    errors.push(...validateGateColor(g.color));
}
```

But `validateGateColor`, `validateGateLabel`, and `validateGateScript` already return `[]` on `null` or `undefined`. The outer checks provide no additional safety and add noise. They can be removed, letting the sub-validators own the null/undefined semantics uniformly.

`packages/daemon/src/lib/space/runtime/gate-evaluator.ts`, lines 235, 240, 245

---

### P2 — Validators are currently orphaned (no integration points)

The `validateGate` JSDoc says it is "applied to new/updated gates via `GateEditorPanel` and MCP tool handlers", but a search across the entire daemon finds zero callers outside the test file. The `spaceWorkflow.update` RPC handler passes gates directly through to `workflowManager.updateWorkflow` without any call to `validateGate`. The `GateEditorPanel` on the web side also does not import or call these functions.

This is likely intentional (validation to be wired up in a follow-on task), but the JSDoc claim that validation is already applied is misleading. Either remove the forward-reference claim from the JSDoc, or add a TODO linking to the follow-on issue, so reviewers of the integration points know what to do.

`packages/daemon/src/lib/space/runtime/gate-evaluator.ts`, lines 212–222  
Related: `packages/daemon/src/lib/rpc-handlers/space-workflow-handlers.ts` (no gate validation before `updateWorkflow`)

---

### P3 — `validateGateScript` lacks a test for `script: false` / `script: 0`

The validator handles these correctly (the `typeof script !== 'object'` branch), but there is no test for primitive non-object, non-null/undefined values. Given the dead-branch issue above this is a minor gap, but worth adding for completeness once the guard is fixed.

---

### What is correct

- Regex `^#[0-9a-fA-F]{6}$` correctly rejects shorthand, alpha, and unpreixed hex. Good.
- Label max-20 boundary is tested at exactly 20 and 21 chars. Good.
- `timeoutMs` boundary tested at 0, -1, 120000, 120001, 200000. Good.
- `validateGate` test that collects all sub-errors in one pass is particularly thorough. Good.
- The "Important" note in the `validateGate` JSDoc about not applying to existing stored gates is a useful invariant to document.

lsm (commented):

Re-review after addressing previous feedback. All prior P1/P2 issues have been resolved:

- Dead `!script` guard removed from `validateGateScript` -- good.
- `validateGate` now calls `validateGateFields` for any present `fields` value including non-arrays -- the test for `fields: 'bad'` confirms this works correctly.
- Redundant guards removed from `validateGate`.
- JSDoc updated with accurate TODO note about wiring into handlers.
- `Gate.color` JSDoc corrected to `#rrggbb`.
- New tests added for `validateGateScript(false)` and `validateGateScript(0)`.

**Code quality is solid.** The validators are well-structured with consistent patterns (early return for null/undefined, type narrowing, error accumulation). Test coverage is thorough with 94 tests passing.

One minor observation (P3):

1. **P3** `validateGateLabel` accepts empty string `''` -- the test explicitly asserts this. This is fine as a design choice (empty string could mean "clear the label"), but worth noting it differs from `validateGateScript` where empty `source` is rejected. The asymmetry is intentional since an empty label is semantically valid while an empty script source is not.

No blocking issues. **Recommendation: APPROVE.**

lsm (commented):

**Recommendation: APPROVE**

All P1/P2 issues from the prior review have been addressed correctly. The follow-up commit makes the implementation clean and consistent. 94 tests pass, all checks green.

---

### What was fixed (verified)

- Dead `!script` guard in `validateGateScript` is gone. The early return on `typeof script !== 'object'` now handles all non-object primitives including `false` and `0`, which are correctly rejected with a type error. Tests at lines 620 and 626 confirm.
- `validateGate` now calls `validateGateFields(g.fields)` for any non-null, non-undefined `fields` value regardless of whether it is an array. Non-array values correctly produce a "expected an array" error. Test at line 826 confirms.
- Redundant guards in `validateGate` are removed. The flow is now clean and linear.
- JSDoc on `validateGate` accurately describes the function and includes a TODO for follow-on wiring.
- `Gate.color` JSDoc updated from "CSS color value" to "hex format `#rrggbb`" — consistent with the regex and the validator error message.

---

### One remaining observation (P3 — not blocking)

**`hasScript` is `true` for `script: false` and `script: 0`**

In `validateGate` (gate-evaluator.ts, line 248):
```ts
const hasScript = g.script !== undefined && g.script !== null;
```

`false` and `0` satisfy this check, so `hasScript` evaluates to `true` when `g.script` is a primitive non-null falsy value. This suppresses the "at least one fields or script" error even though the script is structurally invalid. `validateGateScript(false)` will still generate a type error, so the gate is correctly rejected overall — but the error message would omit the structural complaint.

This is cosmetically inconsistent rather than functionally broken (the validator still rejects the gate). A strict fix would be:
```ts
const hasScript = typeof g.script === 'object' && g.script !== null;
```

There is no `validateGate({ script: false, ... })` test to document the current behavior; adding one would at least make the intent explicit.

This is a nit — the gate will never reach execution with an invalid script because validation errors are present regardless.

---

### Summary

The implementation is correct and well-tested. All previously raised P1/P2 issues are resolved. The one remaining nit (`hasScript` evaluation for non-object falsy values) does not affect correctness since invalid scripts still produce errors. The code is ready to merge.

PR #1197, opened 2026-04-02
test: add maxBuffer and cwd tests for gate script executor

lsm:
## Summary
- Added 6 new tests to `gate-script-executor.test.ts` covering maxBuffer enforcement (>1MB output) and workspacePath-as-cwd verification for all three interpreters (bash, node, python3)
- Total: 88 tests, all passing

## Test plan
- `bun test packages/daemon/tests/unit/space/gate-script-executor.test.ts` — 88 pass
- Pre-commit checks: lint, format, typecheck, knip — all green

### DIFF
--- a/packages/daemon/tests/unit/space/gate-script-executor.test.ts
+++ b/packages/daemon/tests/unit/space/gate-script-executor.test.ts
@@ -34,6 +34,20 @@ const CTX: GateScriptContext = {
 	runId: 'run-456',
 };
 
+/** Probes whether python3 is available on this system. */
+async function isPython3Available(): Promise<boolean> {
+	try {
+		const proc = Bun.spawn(['python3', '-c', 'print("hello")'], {
+			stdout: 'ignore',
+			stderr: 'ignore',
+		});
+		await proc.exited;
+		return true;
+	} catch {
+		return false;
+	}
+}
+
 // ---------------------------------------------------------------------------
 // buildRestrictedEnv
 // ---------------------------------------------------------------------------
@@ -655,8 +669,11 @@ describe('executeGateScript — integration', () => {
 	});
 
 	test('maxBuffer enforcement: script outputting >1MB does not crash', async () => {
-		// Generate a JSON payload larger than 1MB (MAX_BUFFER_BYTES = 1_048_576)
-		// The output will be truncated by collectWithMaxBuffer, resulting in invalid JSON
+		// Generate a JSON payload larger than 1MB (MAX_BUFFER_BYTES = 1_048_576).
+		// The output will be truncated by collectWithMaxBuffer at a byte boundary.
+		// Truncation does NOT guarantee invalid JSON (a cut after `}` would still parse),
+		// but with this specific payload ({ data: 'x'.repeat(2_000_000) }) the truncated
+		// prefix is almost certainly missing its closing braces, so parseJsonStdout returns null.
 		const r = await executeGateScript(
 			{
 				interpreter: 'node',
@@ -667,7 +684,7 @@ describe('executeGateScript — integration', () => {
 			} as GateScript,
 			CTX
 		);
-		// Should not crash — process exits 0, truncated stdout means JSON parse fails → empty data
+		// Should not crash — process exits 0, truncated stdout fails JSON parse → empty data
 		expect(r.success).toBe(true);
 		expect(r.data).toEqual({});
 	});
@@ -686,8 +703,9 @@ describe('executeGateScript — integration', () => {
 		expect(r.data).toEqual({});
 	});
 
-	test('maxBuffer enforcement: stderr exceeding 1MB is truncated', async () => {
-		// Script exits non-zero with >1MB stderr — truncated stderr used as error
+	test('maxBuffer enforcement: stderr exceeding 1MB is truncated to boundary', async () => {
+		// Script exits non-zero with >1MB stderr (base64-encoded 2MB random data produces
+		// ~2.7MB of text). The truncated stderr is used as the error message.
 		const r = await executeGateScript(
 			{
 				interpreter: 'bash',
@@ -696,8 +714,10 @@ describe('executeGateScript — integration', () => {
 			CTX
 		);
 		expect(r.success).toBe(false);
-		// stderr is truncated to 1MB, but should still contain some base64 output
-		expect(r.error!.length).toBeLessThanOrEqual(1_048_576);
+		expect(r.error).toBeDefined();
+		// Truncation should have fired: stderr must be non-empty and at the exact 1MB cap
+		expect(r.error!.length).toBeGreaterThan(0);
+		expect(r.error!.length).toBe(1_048_576);
 	});
 
 	test('workspacePath is set as cwd for bash script', async () => {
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean PR that adds 6 well-structured integration tests covering maxBuffer enforcement and workspacePath-as-cwd verification across all three interpreters. All 88 tests pass.

### Highlights

- **maxBuffer tests** (lines 657-701): Good coverage of three scenarios -- truncated JSON stdout, plain text stdout, and truncated stderr. The `dd | base64` approach reliably generates >1MB output.
- **cwd tests** (lines 703-754): Smart use of `realpathSync('/tmp')` to handle macOS `/tmp` -> `/private/tmp` symlink. Covers bash, node, and python3.
- **python3 availability guard** (lines 728-741): Correctly handles environments where python3 is not installed.

### Minor suggestions (non-blocking)

**P3 — stderr maxBuffer test could assert error has content** (`packages/daemon/tests/unit/space/gate-script-executor.test.ts:700`): The test asserts `r.error!.length <= 1_048_576` but does not verify the error string is non-empty. Adding `expect(r.error!.length).toBeGreaterThan(0)` would make the test more robust against a scenario where `dd` produces no output.

**P3 — Duplicated python3 availability probe**: The python3 availability check (lines 728-738) duplicates the same pattern from the existing python3 describe block (lines 762-779). A shared helper function could reduce this duplication, though this is a pre-existing pattern and not something this PR needs to fix.

No correctness or security issues found. The tests accurately verify the behavior documented in the implementation.

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P1, 2 P2, 1 P3 issues)

---

The 6 new tests do run and pass (88 total, all green). The maxBuffer and cwd coverage fills real gaps in the test suite. However, a few issues reduce confidence that the tests actually verify the behaviors they claim to.

### P1 — Stderr truncation test has no lower-bound assertion

`packages/daemon/tests/unit/space/gate-script-executor.test.ts`, line 698-700:

```ts
expect(r.success).toBe(false);
// stderr is truncated to 1MB, but should still contain some base64 output
expect(r.error!.length).toBeLessThanOrEqual(1_048_576);
```

This assertion is trivially satisfied even if `r.error` is an empty string or a single byte. There is no lower-bound check confirming the truncation actually happened (e.g., that `r.error.length` is at or near 1,048,576 bytes). A future regression where `collectWithMaxBuffer` fails to collect anything would still pass this test. Also, `r.error` should be asserted to be defined before using the non-null `!` operator — the test should include:

```ts
expect(r.error).toBeDefined();
expect(r.error!.length).toBeGreaterThan(0);
expect(r.error!.length).toBeLessThanOrEqual(1_048_576);
```

And ideally there should be an assertion that the length IS at the max (e.g., `toBeGreaterThan(1_000_000)`) since `dd bs=1024 count=2048` produces 2 MB before `base64` encoding.

### P1 — Python3 cwd probe is duplicated verbatim from the python3 describe block

`packages/daemon/tests/unit/space/gate-script-executor.test.ts`, lines 729-739 vs lines 764-780.

The availability probe (`Bun.spawn(['python3', '-c', 'print("hello")'])`) is copy-pasted almost identically in both places. The python3 cwd test would be more maintainable inside the `executeGateScript — python3` describe block alongside the existing python3 test. Duplication in tests is low-risk but signals that organization was not considered, and if the probe logic ever changes it will need to be updated in two places.

### P2 — maxBuffer test comment claims "truncated stdout means JSON parse fails → empty data" but this is only sometimes true

`packages/daemon/tests/unit/space/gate-script-executor.test.ts`, line 659:

```ts
// Should not crash — process exits 0, truncated stdout means JSON parse fails → empty data
```

This is correct for the current implementation, but the test comment overstates the guarantee. The `collectWithMaxBuffer` function truncates at byte boundaries, not JSON token boundaries, so a truncated 1 MB prefix of valid JSON can still be valid JSON if the truncation happens to land after a `}`. The test expects `r.data` to be `{}` but this is only reliable because `'x'.repeat(2_000_000)` produces a string that, when JSON-encoded, results in a JSON blob longer than 1 MB. The comment should clarify that the truncated output happens to be invalid JSON due to the specific string structure, rather than implying truncation always produces invalid JSON.

This is a documentation issue but could mislead future maintainers who copy the pattern.

### P2 — Missing assertion that truncation actually fired

Related to the P1 above: across all three maxBuffer tests, none of the tests verify that `stdoutResult.truncated` or `stderrResult.truncated` was actually set to `true`. These are internal values not exposed in `GateScriptResult`, so they can't be tested directly, but the test assertions should at least confirm the boundary behavior is being exercised (e.g., assert `r.error!.length === 1_048_576` for the stderr test since base64-encoded 2 MB will always produce > 1 MB of text).

### P3 — `r.error!` uses non-null assertion without a prior `toBeDefined()` check

`packages/daemon/tests/unit/space/gate-script-executor.test.ts`, line 700:

```ts
expect(r.error!.length).toBeLessThanOrEqual(1_048_576);
```

If `r.error` is `undefined`, this will throw a runtime TypeError at the `!.length` access, which bun:test will surface as an error rather than a clean assertion failure. Adding `expect(r.error).toBeDefined()` before the length check would produce a clearer failure message.

---

### What's good

- The three cwd tests (bash, node, python3) correctly use `realpathSync('/tmp')` to handle symlink resolution, which is exactly right for macOS where `/tmp` is a symlink to `/private/tmp`.
- The python3 availability probe correctly detects the interpreter rather than assuming it is present.
- The maxBuffer plain-text test correctly verifies that `dd | base64` produces a passing result with empty data.
- All 88 tests pass on the current system.

lsm (commented):

Re-review after addressing previous feedback. All five items have been properly resolved:

1. **P1 (stderr assertions)** — Fixed. The stderr truncation test now has `toBeDefined()`, `toBeGreaterThan(0)`, and exact `toBe(1_048_576)` assertions (lines 717-720).

2. **P1 (python3 duplication)** — Fixed. Extracted shared `isPython3Available()` helper (lines 38-49). Python3 cwd test moved into the existing python3 describe block (lines 770-785). No duplication remains.

3. **P2 (truncation comment)** — Fixed. Comment at line 675 now correctly explains that truncation does not guarantee invalid JSON, but this specific payload structure makes it near-certain.

4. **P2 (exact 1MB cap assertion)** — Fixed. `expect(r.error!.length).toBe(1_048_576)` on line 720.

5. **P3 (non-null assertion guard)** — Fixed. `expect(r.error).toBeDefined()` on line 718 precedes the `!` assertions.

**Additional observations on the final state:**

- 88 tests pass with 128 assertions — comprehensive coverage of `buildRestrictedEnv`, `deepMergeWithDepthLimit`, `parseJsonStdout`, and `executeGateScript`.
- The `realpathSync('/tmp')` usage for cwd assertions correctly handles macOS symlink resolution (`/tmp` -> `/private/tmp`).
- One minor note: the stderr exact-length assertion (`toBe(1_048_576)`) relies on `trim()` in the source not removing bytes from the truncated output. This works because base64 output truncated at exactly 1MB mid-line has no trailing whitespace. If the implementation's `trim()` call ever becomes relevant (e.g., truncation lands on a newline boundary), this assertion could become flaky — but that scenario is unlikely with random data, and the test currently passes reliably.

No issues found. Recommendation: **APPROVE**.

lsm (commented):

**Recommendation: APPROVE** (one low-risk note below)

All five previously-raised issues are addressed correctly:

- **P1 — stderr assertions** (lines 717-720): `toBeDefined()`, `toBeGreaterThan(0)`, and `toBe(1_048_576)` are all present in the right order.
- **P1 — python3 probe duplication** (lines 38-49, 752-785): `isPython3Available()` is extracted as a shared top-level helper and the cwd test is correctly placed inside the `executeGateScript — python3` describe block, eliminating the duplicate probe.
- **P2 — JSON truncation comment** (lines 672-688): The comment now correctly distinguishes between truncation in general and the specific payload that causes a JSON parse failure.
- **P2 — exact stderr length assertion** (line 720): `expect(r.error!.length).toBe(1_048_576)` is present.
- **P3 — `toBeDefined()` guard** (line 717): Added before the non-null assertion access.

---

**One remaining note (P3 / informational)**

The assertion on line 720 asserts that `r.error!.length === 1_048_576` — but `r.error` is the result of `.trim()` on the collected stderr (implementation line 376: `stderrResult.text.trim()`). The collected text is capped at exactly 1_048_576 bytes by `collectWithMaxBuffer`. The assertion holds in practice because base64 line width (76 chars + newline = 77 bytes) leaves the 1 MB boundary at byte offset 67 within a line, meaning the last character is a base64 character, not whitespace — so `trim()` is a no-op at that boundary.

This calculation is correct for the macOS/Linux system `base64` utility (76-char lines). If a different base64 variant wraps at a different width AND the boundary happened to land on a newline, `trim()` would reduce the length below 1_048_576 and the assertion would fail. This is an acceptable trade-off for a unit test on a controlled platform, but worth noting in case CI ever runs on an environment with a non-standard `base64`.

No action required — the test is correct for the targeted platforms and all feedback from the previous round has been properly addressed.

PR #1200, opened 2026-04-02
fix: prevent black gap between keyboard and app bottom on iPad/mobile

lsm:
## Summary
- Extend `useViewportSafety` hook to detect virtual keyboard visibility on all platforms (not just iPad Safari) by comparing `visualViewport.height` with `window.innerHeight`
- When keyboard is open: add `keyboard-open` class to `<html>`, set `--safe-height` to `visualViewport.height`, set `--bottom-bar-height` to `0px` — this eliminates the black gap by shrinking the app container and removing padding for the hidden BottomTabBar
- When keyboard closes: restore original state and dispatch `resize` event so BottomTabBar's ResizeObserver re-measures
- Add CSS rule to translate BottomTabBar offscreen when keyboard is open, with `data-testid` for CSS targeting

## Test plan
- [x] 22 unit tests for useViewportSafety hook (13 existing + 9 new keyboard detection tests)
- [x] All web tests pass (pre-existing failures in visual-editor are unrelated)
- [ ] Manual testing on iPad Safari: focus input → no black gap between keyboard and app bottom
- [ ] Manual testing on iPhone Safari: same verification

### DIFF
--- a/packages/web/src/hooks/__tests__/useViewportSafety.test.ts
+++ b/packages/web/src/hooks/__tests__/useViewportSafety.test.ts
@@ -6,6 +6,7 @@
  */
 
 import { renderHook } from '@testing-library/preact';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { useViewportSafety } from '../useViewportSafety.ts';
 
 // ---------------------------------------------------------------------------
@@ -441,7 +442,7 @@ describe('useViewportSafety — keyboard detection', () => {
 		mockVV._trigger('resize');
 
 		// Should dispatch a resize event for BottomTabBar to re-measure
-		expect(dispatchSpy).toHaveBeenCalledWith(expect.any(Event));
+		expect(dispatchSpy).toHaveBeenCalledWith(expect.objectContaining({ type: 'resize' }));
 
 		dispatchSpy.mockRestore();
 	});
@@ -458,5 +459,35 @@ describe('useViewportSafety — keyboard detection', () => {
 		unmount();
 
 		expect(document.documentElement.classList.contains('keyboard-open')).toBe(false);
+		// --safe-height should be removed on unmount
+		expect(document.documentElement.style.getPropertyValue('--safe-height')).toBe('');
+	});
+
+	it('restores --bottom-bar-height even when it was empty string (desktop)', () => {
+		setNavigator(0, DESKTOP_SAFARI_UA);
+		const mockVV = createMockVisualViewport(WINDOW_INNER_HEIGHT);
+		setVisualViewport(mockVV);
+
+		// No inline --bottom-bar-height set (desktop: BottomTabBar is md:hidden)
+		expect(document.documentElement.style.getPropertyValue('--bottom-bar-height')).toBe('');
+
+		renderHook(() => useViewportSafety());
+
+		// Open keyboard
+		mockVV.height = WINDOW_INNER_HEIGHT - 300;
+		mockVV._trigger('resize');
+
+		// Keyboard open — override set
+		expect(document.documentElement.style.getPropertyValue('--bottom-bar-height')).toBe('0px');
+
+		// Close keyboard — saved value was '' (empty string, falsy)
+		mockVV.height = WINDOW_INNER_HEIGHT;
+		mockVV._trigger('resize');
+
+		// The inline override should be restored (even though value is '')
+		// so the CSS cascade can fall through to the :root rule
+		expect(document.documentElement.classList.contains('keyboard-open')).toBe(false);
+		// The saved '' is restored, removing the inline override
+		expect(document.documentElement.style.getPropertyValue('--bottom-bar-height')).toBe('');
 	});
 });
--- a/packages/web/src/hooks/__tests__/useViewportSafety.test.ts
+++ b/packages/web/src/hooks/__tests__/useViewportSafety.test.ts
@@ -2,7 +2,7 @@
  * Tests for useViewportSafety Hook
  *
  * Tests iPad Safari detection logic, --safe-height CSS property management,
- * and event listener cleanup.
+ * virtual keyboard detection, and event listener cleanup.
  */
 
 import { renderHook } from '@testing-library/preact';
@@ -80,6 +80,9 @@ function restoreVisualViewport(): void {
 	});
 }
 
+/** jsdom default window.innerHeight is 768 */
+const WINDOW_INNER_HEIGHT = 768;
+
 // ---------------------------------------------------------------------------
 // UA fixtures
 // ---------------------------------------------------------------------------
@@ -102,7 +105,7 @@ const CRIOS_UA =
 
 /** Firefox on iOS */
 const FXIOS_UA =
-	'Mozilla/5.0 (iPhone; CPU iPhone OS 16_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/120.0 Mobile/15E148 Safari/605.1.15';
+	'Mozilla/5.0 (iPhone; CPU iPhone OS 16_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/120.0 Mobile/15E148 Safari/604.1';
 
 // ---------------------------------------------------------------------------
 // Lifecycle
@@ -112,6 +115,8 @@ afterEach(() => {
 	restoreNavigator();
 	restoreVisualViewport();
 	document.documentElement.style.removeProperty('--safe-height');
+	document.documentElement.style.removeProperty('--bottom-bar-height');
+	document.documentElement.classList.remove('keyboard-open');
 });
 
 // ---------------------------------------------------------------------------
@@ -121,34 +126,38 @@ afterEach(() => {
 describe('useViewportSafety — iPad Safari detection', () => {
 	it('detects iPad Safari: maxTouchPoints > 1 + Safari UA without Chrome/CriOS/FxiOS', () => {
 		setNavigator(5, IPAD_SAFARI_UA);
-		setVisualViewport(createMockVisualViewport(900));
+		setVisualViewport(createMockVisualViewport(WINDOW_INNER_HEIGHT));
 
 		renderHook(() => useViewportSafety());
 
-		expect(document.documentElement.style.getPropertyValue('--safe-height')).toBe('900px');
+		expect(document.documentElement.style.getPropertyValue('--safe-height')).toBe(
+			`${WINDOW_INNER_HEIGHT}px`
+		);
 	});
 
 	it('does NOT detect iPad Safari when maxTouchPoints is 0 (desktop Mac)', () => {
 		setNavigator(0, DESKTOP_SAFARI_UA);
-		setVisualViewport(createMockVisualViewport(900));
+		setVisualViewport(createMockVisualViewport(WINDOW_INNER_HEIGHT));
 
 		renderHook(() => useViewportSafety());
 
+		// --safe-height should NOT be set on desktop Safari when keyboard is not open
 		expect(document.documentElement.style.getPropertyValue('--safe-height')).toBe('');
 	});
 
 	it('does NOT detect iPad Safari for desktop Chrome (UA contains Chrome)', () => {
 		setNavigator(5, DESKTOP_CHROME_UA);
-		setVisualViewport(createMockVisualViewport(900));
+		setVisualViewport(createMockVisualViewport(WINDOW_INNER_HEIGHT));
 
 		renderHook(() => useViewportSafety());
 
+		// Desktop Chrome: no --safe-height when no keyboard
 		expect(document.documentElement.style.getPropertyValue('--safe-height')).toBe('');
 	});
 
 	it('does NOT detect iPad Safari for CriOS (Chrome on iOS)', () => {
 		setNavigator(5, CRIOS_UA);
-		setVisualViewport(createMockVisualViewport(900));
+		setVisualViewport(createMockVisualViewport(WINDOW_INNER_HEIGHT));
 
 		renderHook(() => useViewportSafety());
 
@@ -157,7 +166,7 @@ describe('useViewportSafety — iPad Safari detection', () => {
 
 	it('does NOT detect iPad Safari for FxiOS (Firefox on iOS)', () => {
 		setNavigator(5, FXIOS_UA);
-		setVisualViewport(createMockVisualViewport(900));
+		setVisualViewport(createMockVisualViewport(WINDOW_INNER_HEIGHT));
 
 		renderHook(() => useViewportSafety());
 
@@ -166,7 +175,7 @@ describe('useViewportSafety — iPad Safari detection', () => {
 });
 
 // ---------------------------------------------------------------------------
-// --safe-height property
+// --safe-height property (iPad Safari always-on behavior)
 // ---------------------------------------------------------------------------
 
 describe('useViewportSafety — --safe-height property', () => {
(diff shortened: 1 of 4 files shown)


lsm (commented):

**Recommendation: APPROVE**

This is a well-structured fix for the iPad/mobile keyboard black gap issue. The implementation is clean, well-documented, and thoroughly tested.

### What works well

- **Correct keyboard detection strategy**: Using `window.innerHeight - visualViewport.height > 50px` is the standard approach for detecting virtual keyboards across platforms. The 50px threshold is reasonable to avoid false positives from browser chrome changes.
- **Clean state management**: The `keyboardOpen` flag prevents redundant DOM writes on every resize event; only transitions (open->close, close->open) trigger changes.
- **Proper save/restore of `--bottom-bar-height`**: Saves the current value before overriding to `0px`, restores on keyboard close.
- **Good test coverage**: 22 tests covering iPad Safari detection, keyboard open/close transitions, threshold boundary conditions, initial state detection, and cleanup.
- **CSS approach via `keyboard-open` class**: Using a class on `<html>` for CSS targeting (translating BottomTabBar offscreen) is a clean separation of concerns.

### Minor observations (P3)

1. **Re-entrant resize dispatch is safe but worth noting** (`useViewportSafety.ts:147`): When the keyboard closes, `handleResize` dispatches `new Event('resize')` on `window`, which re-enters the same handler. This is safe because `keyboardOpen` is already `false` on re-entry so neither branch executes. On iPad Safari, `updateSafeHeight` runs redundantly (a harmless extra DOM write). No action needed, just noting for future maintainers.

2. **Cleanup does not remove `--safe-height` on unmount** (`useViewportSafety.ts:170-178`): If the hook unmounts while the keyboard is open on a non-iPad browser, the `--safe-height` property remains set. This is a non-issue in practice since the hook is mounted once globally in `App.tsx` and never unmounts during normal app lifecycle.

3. **Unrelated change in diff**: The `VisualWorkflowEditor.tsx` change (`gate.fields ?? []` -> `gate.fields`) appears to be from a rebase/merge with dev changes, not part of this PR's commit. The actual commit (`13d4a250c`) only touches the 4 relevant files.

No blocking issues found. The implementation correctly addresses the original problem and the tests validate the behavior thoroughly.

lsm (commented):

**Recommendation: REQUEST_CHANGES** — two real bugs to fix before merging, plus a few smaller issues.

---

### P0 — Bugs

**1. `savedBottomBarHeight` empty-string falsy check means `--bottom-bar-height` is never restored when BottomTabBar is hidden on desktop**

`packages/web/src/hooks/useViewportSafety.ts` lines 140–143 and 176–178:

```ts
if (savedBottomBarHeight) {
    document.documentElement.style.setProperty('--bottom-bar-height', savedBottomBarHeight);
```

`getPropertyValue('--bottom-bar-height')` returns `''` (empty string) when the property is not set inline (the `:root` rule in CSS sets it to `0px`, but that is a stylesheet rule, not an inline style). On desktop breakpoints the BottomTabBar is hidden (`md:hidden`) so `offsetHeight` is 0 and `BottomTabBar.useEffect` never sets the inline style. When the keyboard closes, `savedBottomBarHeight` is `''`, the falsy guard skips the `setProperty` call, and the inline `--bottom-bar-height: 0px` written at keyboard-open time is never cleaned up. The CSS cascade falls back to the `:root` rule's `0px`, so in this particular combination the end result is coincidentally correct — but the invariant is broken: after a keyboard cycle the hook leaves a stale inline `--bottom-bar-height: 0px` on `<html>` even on desktop. If `BottomTabBar` later measures itself (after orientation change) and sets `--bottom-bar-height: 56px`, that works fine — but if the hook fires another keyboard open/close before `BottomTabBar` has run, `savedBottomBarHeight` will capture the stale `'0px'` instead of the real height.

The same falsy check appears in the cleanup function (line 176). If the keyboard was open when the component unmounts (unlikely in practice but possible), `savedBottomBarHeight` holds the pre-keyboard value, and `''` would fail the guard, leaving `--bottom-bar-height: 0px` as an inline style permanently.

Fix: change both guards from `if (savedBottomBarHeight)` to `if (savedBottomBarHeight !== null)`. `savedBottomBarHeight` is initialized to `null` and reset to `null` after use, so `null` is the correct sentinel, not falsiness.

**2. `dispatchEvent(new Event('resize'))` re-enters `handleResize` synchronously while `keyboardOpen` is still `false`**

`packages/web/src/hooks/useViewportSafety.ts` line 146:

```ts
window.dispatchEvent(new Event('resize'));
```

`dispatchEvent` is synchronous. At line 146, `keyboardOpen` has just been set to `false` and `vv.height` is back to `window.innerHeight` (no keyboard). `handleResize` is registered on `window.resize`, so it is called immediately within the same event. Part 2 of `handleResize` evaluates `isKeyboardVisible(vv)` — that is `false` — and `keyboardOpen` is `false`, so the `else if` branch is skipped (correct). However Part 1 (`ipadSafari && updateSafeHeight`) fires again, triggering a redundant DOM write for iPad Safari. This is harmless for correctness but represents unnecessary re-entrant DOM mutation. More importantly, if any other `window.resize` listener (e.g. `BottomTabBar.onResize` → `requestAnimationFrame(updateHeight)`) changes `--bottom-bar-height` before `handleResize` finishes processing the close event, the saved value could be overwritten before the hook restores it.

The actual ordering is: hook restores `--bottom-bar-height` at line 141, _then_ dispatches resize at 146 which queues a rAF in `BottomTabBar`. So practically no corruption occurs because rAF is deferred. Still, the re-entrant `handleResize` call during the dispatch is worth noting as fragile; it could become a real problem if `BottomTabBar`'s listener is ever changed from rAF to synchronous.

---

### P1 — Should Fix

**3. CSS comment mentions ContextPanel hiding but no rule exists**

`packages/web/src/styles.css` line 601:
```
 * - The ContextPanel sidebar (position: fixed on mobile) should also stay hidden
```

No CSS rule for hiding the ContextPanel when `keyboard-open` is present. The comment promises a behavior that is not implemented. Either remove the comment or add the corresponding rule. If ContextPanel is `position: fixed` and overlaps the keyboard, this is a real visual regression on mobile.

**4. No CSS `transition` on `[data-testid="bottom-tab-bar"]` for the translateY**

The CSS rule at line 603 abruptly teleports the BottomTabBar offscreen (`transform: translateY(100%)`) with no animation. The keyboard appears with a visual transition on iOS; the BottomTabBar disappearing instantly will look jarring. A `transition: transform 0.2s ease` on the BottomTabBar element (either in its class list or in the CSS rule) would smooth this. The BottomTabBar root div currently has no `transition` class.

**5. `--safe-height` not cleaned up on unmount for iPad Safari**

The cleanup function (`packages/web/src/hooks/useViewportSafety.ts` lines 170–179) removes the `keyboard-open` class and restores `--bottom-bar-height`, but does NOT remove `--safe-height`. For iPad Safari this means `--safe-height` persists as an inline style on `<html>` after unmount. In practice `App` never unmounts, so this is benign today. But it means the cleanup is incomplete if ever tested in isolation (including in unit tests that do not call `unmount()`).

---

### P2 — Suggestions

**6. `window.innerHeight` reliability on iOS Safari**

The detection formula `window.innerHeight - vv.height > KEYBOARD_THRESHOLD` uses `window.innerHeight` as the "full viewport" reference. On iOS Safari, `window.innerHeight` can itself change when the address bar retracts or appears (it reflects the _current_ inner height, not the maximum). On iPhone, if the address bar is partially hidden when the input is focused, `window.innerHeight` may not equal the screen height, causing the delta to undercount the keyboard size. The KEYBOARD_THRESHOLD of 50px provides a reasonable buffer, but edge cases exist (e.g. small split-screen on iPad). This is an inherent limitation of the approach without `navigator.virtualKeyboard` API (not broadly supported yet), so it is acceptable as a known limitation — consider documenting it in the JSDoc.

**7. Keyboard state inconsistency: `--safe-height` set but `--bottom-bar-height` not restored when unmount happens mid-keyboard**

The unmount cleanup at line 175 removes `keyboard-open` and conditionally restores `--bottom-bar-height`, but does NOT `removeProperty('--safe-height')` on non-iPad platforms. After unmount with keyboard open, `--safe-height` remains set to the reduced viewport height, causing `.h-safe-screen` consumers to see a truncated height.

---

### P3 — Nits

**8.** The CSS comment at `styles.css` line 597 says the detection is `visualViewport.height < window.innerHeight` but the actual check is `window.innerHeight - vv.height > KEYBOARD_THRESHOLD`. These are not equivalent — the comment is slightly misleading.

**9.** The test for "dispatches window resize when keyboard closes" (line 427) asserts `expect(dispatchSpy).toHaveBeenCalledWith(expect.any(Event))` but does not verify it was called with a `resize` event specifically. It would be more precise to assert `expect.objectContaining({ type: 'resize' })` or check `dispatchSpy.mock.calls[0][0].type === 'resize'`.

---

### Summary

The core approach is sound and well-structured. The two P0 bugs center on: (1) using falsy check on empty string instead of `!== null` for the saved bottom-bar-height sentinel, and (2) a re-entrant `handleResize` invocation from `dispatchEvent`. P1 issues include a misleading comment promising ContextPanel behavior that doesn't exist, and an abrupt (non-animated) BottomTabBar disappearance.

lsm (commented):

**Recommendation: APPROVE**

All P0/P1/P2 issues from the previous GLM review have been correctly addressed in commit `368709fb8`:

### Fixes verified

1. **P0 - `savedBottomBarHeight` falsy check** (`useViewportSafety.ts:156,200`): Changed from `if (savedBottomBarHeight)` to `if (savedBottomBarHeight !== null)`. This correctly handles the case where `getPropertyValue` returns `''` (empty string) on desktop when BottomTabBar never sets the inline style. The `null` sentinel is the right approach.

2. **P0 - Re-entrant `dispatchEvent` documentation** (`useViewportSafety.ts:161-168`): Added a thorough comment explaining why the synchronous re-entry is safe (keyboardOpen already false, keyboard not visible, rAF deferred in BottomTabBar). This is the right call -- the behavior is correct, it just needed documentation.

3. **P1 - Misleading ContextPanel CSS comment**: Removed. The CSS comment now accurately describes only the BottomTabBar behavior.

4. **P1 - CSS transition on BottomTabBar**: Added `transition: transform 0.2s ease`.

5. **P1 - `--safe-height` cleanup on unmount** (`useViewportSafety.ts:199`): Added `document.documentElement.style.removeProperty('--safe-height')` in the cleanup function.

6. **P2 - `window.innerHeight` limitation documented** (`useViewportSafety.ts:27-31,81-83`): Added JSDoc explaining the iOS Safari limitation where `window.innerHeight` fluctuates with address bar visibility.

7. **P3 - CSS comment accuracy**: Updated to say "exceeding a 50px threshold" instead of the incorrect `visualViewport.height < window.innerHeight`.

8. **P3 - Resize event assertion**: Improved from `expect.any(Event)` to `expect.objectContaining({ type: 'resize' })`.

### Previous claude-opus-4-6 P3 nits -- all addressed

- Re-entrant resize dispatch: now documented with safety analysis.
- `--safe-height` not cleaned on unmount: now cleaned via `removeProperty`.

### New test coverage

- Test for `--safe-height` cleanup on unmount (line 463).
- Test for empty-string `--bottom-bar-height` restoration on desktop (lines 466-492), which directly validates the P0 fix.

### Remaining P3 nit

The `transition: transform 0.2s ease` is inside the `html.keyboard-open` selector, meaning it only applies while the class is present. When `keyboard-open` is removed (keyboard closes), the `transition` property is also removed, so the BottomTabBar will snap back to its original position without animation rather than sliding in. For a bidirectional transition, the `transition` would need to be on the base `[data-testid="bottom-tab-bar"]` element. This may be intentional (matching the keyboard's own close behavior), so flagging as P3 only.

---

**Summary:** All previously identified issues have been properly fixed with corresponding test coverage. The code is clean, well-documented, and ready to merge.

lsm (commented):

**Recommendation: APPROVE** (with one minor note)

All four issues from the previous review have been addressed.

---

### P0 resolutions - verified

**P0-1: `savedBottomBarHeight` falsy guard** — Fixed. The variable is now typed as `string | null`, initialized to `null`, and the guard is `!== null` (line 156). The comment at lines 152-155 accurately documents why `!== null` is required (getPropertyValue returns `''` for unset properties, which is a valid saved value).

**P0-2: synchronous `dispatchEvent('resize')` re-entering `handleResize` mid-transition** — Addressed. The code dispatches `resize` only after `keyboardOpen` is already set to `false` and the keyboard is confirmed not visible. The comment at lines 162-168 explicitly documents the re-entry analysis. The BottomTabBar's `onResize` wraps `updateHeight` in `requestAnimationFrame` (line 149 of BottomTabBar.tsx), which defers the DOM write out of the synchronous re-entry. Safe.

### P1 resolutions - verified

**P1-1: Misleading CSS comment promising ContextPanel hiding** — Resolved. The new `keyboard-open` CSS block at lines 592-606 of styles.css contains only the BottomTabBar translate rule with no misleading forward-looking comment.

**P1-2: No CSS transition on BottomTabBar** — Partially addressed. The rule `html.keyboard-open [data-testid="bottom-tab-bar"] { transform: translateY(100%); transition: transform 0.2s ease; }` animates the bar offscreen when the keyboard opens.

### Remaining P2 observation

**CSS transition is one-directional** (`packages/web/src/styles.css`, lines 603-606): The `transition` declaration lives exclusively inside the `keyboard-open` selector. When the `keyboard-open` class is removed, the element's computed transition property reverts to the default (none), so the bar snaps back to `translateY(0)` instantly. The slide-in animation on keyboard-dismiss does not play. To get symmetric behavior, a base transition rule should also exist on the element — either via Tailwind (`transition-transform duration-200`) in BottomTabBar.tsx or via a standalone CSS rule targeting `[data-testid="bottom-tab-bar"]`. This is a cosmetic UX issue, not a correctness bug.

### Test coverage

The 9 new keyboard-detection tests in `useViewportSafety.test.ts` cover:
- initial mount with keyboard already open
- open/close cycle with `--bottom-bar-height` save/restore
- threshold boundary (50px vs 51px)
- empty-string saved value case (desktop, no inline style)
- `window.dispatchEvent('resize')` call on close
- cleanup on unmount

Coverage is solid for the hook logic.

lsm (commented):

The P3 nit from the previous review has been correctly addressed. The `transition: transform 0.2s ease` declaration is now on the base `[data-testid="bottom-tab-bar"]` selector (line 606-608 of `styles.css`), so the transition animates bidirectionally -- both when the keyboard opens (slide off) and closes (slide back).

- **useViewportSafety hook**: Clean keyboard detection via `window.innerHeight - visualViewport.height > 50px` threshold, with proper state tracking (`keyboardOpen` flag) to only trigger on transitions. The re-entrancy from `window.dispatchEvent(new Event('resize'))` is correctly analyzed in the inline comment.
- **CSS**: The two-rule approach (base transition + conditional transform) is the standard pattern for CSS class-toggled animations.
- **Tests**: Good coverage of threshold boundaries (50px vs 51px), keyboard open/close cycles, cleanup on unmount, and the empty-string `--bottom-bar-height` edge case.

lsm (commented):

**Recommendation: APPROVE**

The P2 raised in the previous review has been correctly fixed.

### P2 Fix Verified

In `packages/web/src/styles.css`, `transition: transform 0.2s ease` has been moved from the `html.keyboard-open [data-testid="bottom-tab-bar"]` selector to the base `[data-testid="bottom-tab-bar"]` selector. This is the correct fix: CSS transitions only run when the triggering property changes on an element that already has the `transition` declaration present. By placing it on the base selector, the animation now fires bidirectionally — the bar slides offscreen smoothly when the keyboard opens and slides back smoothly when it closes.

The accompanying comment in the CSS block documents this intentional placement, which aids future maintainers.

### No Remaining Issues

The rest of the implementation is unchanged and was previously reviewed as sound:

- `useViewportSafety.ts` correctly manages `keyboard-open` class, `--safe-height`, and `--bottom-bar-height` CSS vars with proper cleanup on unmount.
- The `savedBottomBarHeight !== null` sentinel correctly handles the empty-string case on desktop where `BottomTabBar` is `md:hidden` and never sets the property inline.
- `BottomTabBar.tsx` uses a `ResizeObserver` plus a `window resize` listener with `requestAnimationFrame` debouncing for re-measurement when the keyboard closes.
- The `window.dispatchEvent(new Event('resize'))` call on keyboard close correctly triggers `BottomTabBar` re-measurement without re-entering the keyboard detection branch (because `keyboardOpen` is already false before dispatch).
- Test coverage in `useViewportSafety.test.ts` is thorough: iPad Safari detection, threshold boundaries, keyboard open/close state, initial keyboard state on mount, cleanup on unmount, and the empty-string bottom-bar-height edge case are all covered.

PR #1204, opened 2026-04-02
fix: handle missing custom_agent_id in migration 29

lsm:
## Summary
- make migration 29 resilient when `space_tasks.custom_agent_id` is missing on legacy DBs
- add `custom_agent_id` via `ALTER TABLE` before creating `idx_space_tasks_custom_agent_id`
- guard index creation behind a column existence check
- add a regression test covering legacy `space_tasks` schemas without `custom_agent_id`

## Why
Some older databases hit startup failure during migration 29:
`SQLiteError: no such column: custom_agent_id`

This change ensures migration 29 is safe for those legacy schemas and remains idempotent.

## Validation
- `cd packages/daemon && bun test tests/unit/storage/migrations/migration-29_test.ts`
- result: 18 passed, 0 failed

### DIFF
--- a/packages/daemon/src/storage/schema/migrations.ts
+++ b/packages/daemon/src/storage/schema/migrations.ts
@@ -1846,6 +1846,10 @@ function runMigration29(db: BunDatabase): void {
 	db.exec(
 		`CREATE INDEX IF NOT EXISTS idx_space_tasks_workflow_run_id ON space_tasks(workflow_run_id)`
 	);
+	// Existing databases created by early Space previews may be missing this column.
+	if (!tableHasColumn(db, 'space_tasks', 'custom_agent_id')) {
+		db.exec(`ALTER TABLE space_tasks ADD COLUMN custom_agent_id TEXT`);
+	}
 	if (tableHasColumn(db, 'space_tasks', 'custom_agent_id')) {
 		db.exec(
 			`CREATE INDEX IF NOT EXISTS idx_space_tasks_custom_agent_id ON space_tasks(custom_agent_id)`
--- a/packages/daemon/tests/unit/storage/migrations/migration-29_test.ts
+++ b/packages/daemon/tests/unit/storage/migrations/migration-29_test.ts
@@ -13,10 +13,10 @@
  * - Indexes are created
  *
  * NOTE: Several columns originally added by M29 were subsequently removed:
- * - space_tasks.custom_agent_id, workflow_node_id — removed by M71
- * - space_workflow_runs.config — removed by M71
- * - space_tasks status values changed by M71 ('pending'→'open', 'completed'→'done', etc.)
- * Tests have been updated to reflect the post-M71 schema.
+ * - space_tasks.custom_agent_id, workflow_node_id — removed by M73
+ * - space_workflow_runs.config — removed by M73
+ * - space_tasks status values changed by M73 ('pending'→'open', 'completed'→'done', etc.)
+ * Tests have been updated to reflect the post-M73 schema.
  */
 
 import { describe, test, expect, beforeEach, afterEach } from 'bun:test';
@@ -100,10 +100,10 @@ describe('Migration 29: Space system tables', () => {
 	});
 
 	// -------------------------------------------------------------------------
-	// space_tasks columns — post-M71 schema
+	// space_tasks columns — post-M73 schema
 	// -------------------------------------------------------------------------
 
-	test('space_tasks does NOT have custom_agent_id column after M71 removed it', () => {
+	test('space_tasks does NOT have custom_agent_id column after M73 removed it', () => {
 		runMigrations(db, () => {});
 		expect(columnExists(db, 'space_tasks', 'custom_agent_id')).toBe(false);
 	});
@@ -150,10 +150,10 @@ describe('Migration 29: Space system tables', () => {
 			)
 		`);
 
-		// M29 should add the missing column so intermediate migrations (M30–M70) work.
-		// After M71 runs, it removes the column — but M29 must not throw in the process.
+		// M29 should add the missing column so intermediate migrations (M30–M72) work.
+		// After M73 runs, it removes the column — but M29 must not throw in the process.
 		expect(() => runMigrations(db, () => {})).not.toThrow();
-		// M71 removes custom_agent_id, so it should NOT exist after all migrations.
+		// M73 removes custom_agent_id, so it should NOT exist after all migrations.
 		expect(columnExists(db, 'space_tasks', 'custom_agent_id')).toBe(false);
 	});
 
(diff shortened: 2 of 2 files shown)


lsm (commented):

**Recommendation: APPROVE**

The rebase and conflict resolution look correct. The changes are minimal and well-targeted:

**Migration fix** (`packages/daemon/src/storage/schema/migrations.ts`, lines 1849-1852): The guard correctly adds `custom_agent_id` via `ALTER TABLE` only when the column is missing, which handles early preview databases that were created before M29 included that column. The subsequent `tableHasColumn` check at line 1853 then safely creates the index. This is idempotent and harmless for databases that already have the column.

**Regression test** (`packages/daemon/tests/unit/storage/migrations/migration-29_test.ts`, lines 111-158): The test creates a realistic legacy `space_tasks` table without `custom_agent_id`, runs the full migration chain, and verifies it completes without throwing. The final assertion correctly checks that `custom_agent_id` does NOT exist after all migrations run (since M73 removes it).

**P3 nit**: The test comments at lines 153-156 reference "M71" as removing `custom_agent_id`, but it is actually M73 that does this. This is a pre-existing inaccuracy in the file header (lines 16-17), not introduced by this PR.

No issues found. Clean rebase with correct conflict resolution.

lsm (commented):

**Recommendation: APPROVE** (with one comment to fix)

---

### Summary

The migration fix is correct and the test assertions pass. One comment in the new test incorrectly attributes the removal of `custom_agent_id` to M71 when it is actually M73 that drops the column. The implementation logic is sound.

---

### What was reviewed

- `packages/daemon/src/storage/schema/migrations.ts` — `runMigration29` patched with ALTER TABLE guard
- `packages/daemon/tests/unit/storage/migrations/migration-29_test.ts` — regression test for legacy schema

---

### Findings

**P2 — Wrong migration number cited in comments (two locations)**

`packages/daemon/tests/unit/storage/migrations/migration-29_test.ts`

Lines 16–17 (file-level doc comment):
```
- space_tasks.custom_agent_id, workflow_node_id — removed by M71
- space_workflow_runs.config — removed by M71
```

Lines 153–156 (inline comment in the new test):
```
// M29 should add the missing column so intermediate migrations (M30–M70) work.
// After M71 runs, it removes the column — but M29 must not throw in the process.
```

Both are incorrect. `runMigration71` only fixes corrupted cron strings in the `goals` table (schedule column normalisation). It does not touch `space_tasks` at all. The actual removal of `custom_agent_id` (and the table rebuild) happens in `runMigration73`, which creates `space_tasks_m71_new` as an internal temp table name — that temp-table name appears to be the source of the confusion.

The assertions themselves (`toBe(false)` for column existence after all migrations) are **correct** because M73 does remove the column. Only the prose attribution is wrong, but it will mislead any future developer tracing which migration is responsible for the schema change.

---

**P3 — Mildly redundant double-check pattern in migration code**

`packages/daemon/src/storage/schema/migrations.ts`, lines 1850–1857:

```ts
if (!tableHasColumn(db, 'space_tasks', 'custom_agent_id')) {
    db.exec(`ALTER TABLE space_tasks ADD COLUMN custom_agent_id TEXT`);
}
if (tableHasColumn(db, 'space_tasks', 'custom_agent_id')) {
    db.exec(`CREATE INDEX IF NOT EXISTS ...`);
}
```

After the first block, the column is guaranteed to exist, so the second `if` always evaluates to true. It works correctly and `CREATE INDEX IF NOT EXISTS` makes it idempotent, but the pattern could be expressed more clearly as a single conditional or a simple unconditional index creation after the guard. Not a bug, just slightly noisy.

---

### Verdict

The core fix is correct: legacy databases missing `custom_agent_id` will have it added by M29 before intermediate migrations run, and M73 later drops it in a full table rebuild. The 18-test suite passes. The only actionable item is updating the two comment strings to say "M73" instead of "M71" to prevent future confusion.

lsm (commented):

### Recommendation: APPROVE

This PR is clean and correct. The changes are minimal, well-targeted, and properly tested.

**Migration fix (migrations.ts, lines 1849-1852):**
The guard correctly adds `custom_agent_id` via `ALTER TABLE` when missing, ensuring M29 does not throw on legacy databases that were created before this column existed. The existing `tableHasColumn` check on line 1853 then proceeds to create the index as normal. This is idempotent — on databases that already have the column, the `ALTER TABLE` is skipped.

**Test (migration-29_test.ts, lines 111-158):**
The new `legacy space_tasks without custom_agent_id is upgraded safely` test correctly exercises the fix by pre-creating a `space_tasks` table without `custom_agent_id` (simulating the legacy schema), then running all migrations and verifying no error is thrown and that M73 ultimately removes the column. Good coverage of the actual failure scenario.

**Comment updates (M71 -> M73):**
All 20 references to M71 have been correctly updated to M73, which is the actual migration that removes `custom_agent_id` and other columns. Verified by confirming `runMigration73` (line 4720) contains the column removal logic and that zero M71 references remain in the test file.

No issues found.

lsm (commented):

---

### Summary

The fix is correct and complete. Migration 29 is now resilient to legacy databases that were created before `custom_agent_id` was added to `space_tasks`. The guard pattern (add column if missing, then create index only if column exists) is idempotent and safe. All 20 M71-to-M73 reference corrections in the test file are accurate — M73 is indeed the migration that rebuilds `space_tasks` and removes that column.

---

### Migration Fix (`migrations.ts`, lines 1849-1857)

The two-step guard is correct:

```ts
if (!tableHasColumn(db, 'space_tasks', 'custom_agent_id')) {
    db.exec(`ALTER TABLE space_tasks ADD COLUMN custom_agent_id TEXT`);
}
if (tableHasColumn(db, 'space_tasks', 'custom_agent_id')) {
    db.exec(`CREATE INDEX IF NOT EXISTS idx_space_tasks_custom_agent_id ON space_tasks(custom_agent_id)`);
}
```

Both the column-add path and the index-creation path are idempotent. The `tableHasColumn` helper is already used elsewhere in the file consistently.

---

### Regression Test (`migration-29_test.ts`, lines 111-158)

The new test correctly simulates a legacy schema missing `custom_agent_id` and then runs all migrations to verify:
1. No throw occurs.
2. The column is absent after M73 (it was added by M29 so intermediate migrations work, then removed by M73).

One observation about the pre-created table: it declares FK references to `spaces`, `space_workflow_runs`, and `space_workflow_steps`, none of which exist when the DDL runs. SQLite defers FK enforcement to DML operations, so the `CREATE TABLE` won't fail. This is fine and consistent with how SQLite works with `PRAGMA foreign_keys = ON`.

---

### Pre-existing Issue (not introduced here)

**P3 — naming inconsistency in `runMigration73`**: The temporary tables used in M73's rebuild are named `space_tasks_m71_new` and `space_workflow_runs_m71_new` (referencing M71 instead of M73). This is pre-existing technical debt — this PR did not introduce it and correctly notes in comments that M73 is the responsible migration.

---

### PR Body Discrepancy (informational)

The PR body states "18 passed, 0 failed" but the test file contains 17 test cases. This is a minor count discrepancy likely from a recount after the regression test was added. Not a blocker.

PR #1206, opened 2026-04-02
docs: update E2E guardian plan with corrected CI artifact handling

lsm:
Corrections to the E2E guardian hourly plan based on fact-checking:

- **Fix `gh run download` usage**: The CLI supports only one `--pattern` per invocation; the old plan used two flags which is invalid. Replaced with per-job download loop using `--name` and `--dir`.
- **Clarify reporter**: CI runs Playwright HTML reporter (not monocart); there is no separate JSON output file. Added notes on how to read `test-results/` and `playwright-report/` artifacts.
- **Artifact retention**: 7 days (explicitly set in workflow), not the default 90.
- **Cancelled run handling**: With `cancel-in-progress: true`, cancelled runs are common. Added explicit filtering and edge case for when all recent runs are cancelled/in-progress.
- **Updated known pre-existing issues**: Added 2026-04-02 findings (8 root cause categories, 21 failing jobs).
- **Health check log cross-reference**: Leader agent should check `docs/e2e-health-check-log.md` before creating fix tasks to avoid re-investigating known issues.

### DIFF
--- a/docs/plans/dev-branch-e2e-tests-guardian-hourly.md
+++ b/docs/plans/dev-branch-e2e-tests-guardian-hourly.md
@@ -39,10 +39,10 @@ Run E2E CI checks on recent dev branch commits. For each failing test suite, rep
 
 1. Query the latest CI runs on dev (include `headSha` for commit hash reporting):
    ```bash
-   gh run list --repo lsm/neokai --branch dev --limit 3 --json databaseId,status,conclusion,createdAt,headSha
+   gh run list --repo lsm/neokai --branch dev --limit 5 --json databaseId,status,conclusion,createdAt,headSha
    ```
 
-2. Pick the most recent completed run (status != "in_progress"). If the latest is still running, wait up to 10 minutes polling every 2 minutes:
+2. Pick the most recent completed run. Skip runs with `conclusion` = `"cancelled"` (these were superseded by a newer push). If the latest run has `status` = `"in_progress"`, wait up to 10 minutes polling every 2 minutes:
    ```bash
    CI_STATUS=$(gh run list --repo lsm/neokai --branch dev --limit 1 --json status --jq '.[0].status')
    if [ "$CI_STATUS" = "in_progress" ]; then
@@ -54,30 +54,64 @@ Run E2E CI checks on recent dev branch commits. For each failing test suite, rep
      done
      if [ "$CI_STATUS" = "in_progress" ]; then
        echo "CI still running after 10 minutes. Reporting status and stopping."
-       # Report which run is in progress and stop — will retry next hourly run
        exit 0
      fi
    fi
    ```
 
-3. List failed E2E jobs (names contain "E2E"):
+   If all recent runs (last 5) are cancelled or in-progress, report "No completed CI runs available" and stop.
+
+3. List failed E2E jobs from the chosen run. E2E job names start with "E2E ":
    ```bash
    RUN_ID=<run-id>
-   gh run view --repo lsm/neokai $RUN_ID --json jobs --jq '.jobs[] | select(.conclusion == "failure") | select(.name | test("E2E")) | {name, databaseId, url}'
+   gh run view --repo lsm/neokai $RUN_ID --json jobs \
+     --jq '.jobs[] | select(.conclusion == "failure") | select(.name | startswith("E2E")) | {name, databaseId, steps: [.steps[] | select(.conclusion == "failure") | {name, conclusion}]}'
    ```
 
 4. If no E2E jobs failed, report "All green" and stop.
 
-5. For each failed E2E job, download the test report artifact and extract failure details:
+5. For each failed E2E job, extract the failure message from the job's step logs. Use `gh run view --log` filtered to the failing step, or download the test report artifact:
    ```bash
-   gh run download $RUN_ID --repo lsm/neokai --pattern "e2e-no-llm-results-*" --pattern "e2e-results-llm-*"
-   # Read playwright-report/index.html for failure details
+   # Option A: Get step failure summary from the job (fast, no download)
+   gh run view --repo lsm/neokai $RUN_ID --json jobs \
+     --jq '.jobs[] | select(.name == "<job-name>") | .steps[] | select(.conclusion == "failure") | .name'
+
+   # Option B: Download test report artifact (for detailed failure messages)
+   # NOTE: gh run download supports only ONE --pattern per invocation
+   # Artifact name for no-LLM jobs: e2e-no-llm-results-<test-name>
+   # Artifact name for LLM jobs: e2e-results-llm-<test-name>
+   mkdir -p /tmp/e2e-guardian && cd /tmp/e2e-guardian
+   gh run download $RUN_ID --repo lsm/neokai --name "e2e-no-llm-results-<test-name>"
+   # The artifact contains playwright-report/ (HTML) and test-results/ (JSON attachments)
+   # Read test-results/ for structured failure data, or playwright-report/ for HTML summary
    ```
+
+   For batch downloads of all failed artifacts, iterate over failed job names:
+   ```bash
+   FAILED_JOBS=$(gh run view --repo lsm/neokai $RUN_ID --json jobs \
+     --jq '.jobs[] | select(.conclusion == "failure") | select(.name | startswith("E2E")) | .name')
+   for JOB in $FAILED_JOBS; do
+     # Extract test name from job name: "E2E No-LLM (features-foo)" -> "features-foo"
+     TEST_NAME=$(echo "$JOB" | sed 's/.*(\(.*\))/\1/')
+     # Determine artifact name based on job prefix
+     if echo "$JOB" | grep -q "No-LLM"; then
+       ARTIFACT="e2e-no-llm-results-$TEST_NAME"
+     else
+       ARTIFACT="e2e-results-llm-$TEST_NAME"
+     fi
+     echo "Downloading artifact: $ARTIFACT"
+     gh run download $RUN_ID --repo lsm/neokai --name "$ARTIFACT" --dir "/tmp/e2e-guardian/$TEST_NAME" 2>/dev/null || echo "  Artifact not found (may have expired or upload failed)"
+   done
+   ```
+
+   **Artifact retention**: Failed job artifacts are retained for 7 days. Artifacts older than 7 days are auto-deleted by GitHub Actions.
+
+   **Reporter note**: CI runs the Playwright HTML reporter (not monocart). The `test-results/` directory contains per-test JSON data from Playwright's internal reporter. The `playwright-report/` directory contains a self-contained HTML report with inline compressed data (no separate JSON file).
 
 6. For each failure, report:
    - Test file path (e.g., `tests/features/mission-terminology.e2e.ts`)
    - CI job name (e.g., `E2E No-LLM (features-mission-terminology)`)
-   - Failure message / error snippet
+   - Failure message / error snippet (from step logs or artifact test-results)
    - Commit hash from `headSha` (links failure to specific commit)
    - Suspected root cause category: `test-bug` | `product-bug` | `flaky` | `env`
    - Whether it is a pre-existing issue (check `docs/e2e-health-check-log.md` for prior occurrences)
@@ -93,11 +127,16 @@ These issues were documented in prior health check runs (see `docs/e2e-health-ch
 - `features-worktree-isolation`: session deletion race condition
 - `features-space-session-groups`: workspace path initialization race
 - `features-reference-autocomplete` / `features-task-lifecycle`: E2E temp workspace lacks `.git` directory
+- Neo panel doesn't close on Escape key (affects 7 suites, first seen 2026-04-02)
+- Space UNIQUE constraint on retry (affects 7 suites, first seen 2026-04-02)
+- Ambiguous locators / strict mode violations (affects 5 suites, first seen 2026-04-02)
+- AI-dependent tests in No-LLM matrix (1 suite, first seen 2026-04-02)
 
 ## CI Notes
-- **cancel-in-progress**: CI has `cancel-in-progress: true`. Before pushing any fix, always check if a CI run is in progress (see step 2 wait strategy). Pushing while CI runs will cancel the in-progress run.
+- **cancel-in-progress**: CI has `cancel-in-progress: true` with group `ci-${{ github.ref }}`. Pushing to dev while CI is running will cancel the in-progress run. Always check for cancelled runs and skip them (step 2).
 - E2E test categories: `no-llm` (UI-only, fully parallel) and `llm` (requires LLM API, max 4 parallel).
 - The CI `discover` job auto-categorizes tests; artifact names follow `e2e-no-llm-results-<name>` and `e2e-results-llm-<name>`.
+- **Artifact naming**: Jobs are `E2E No-LLM (<name>)` and `E2E LLM (<name>)`. Strip the prefix and parentheses to get the artifact name suffix.
 ```
 
 ---
@@ -107,8 +146,8 @@ These issues were documented in prior health check runs (see `docs/e2e-health-ch
 The mission system uses **plan reuse** for execution 2+:
 
 1. **First execution**: The Planner agent expands the mission description into a concrete task plan (the discovery task). The task runs, reports failures, and the Leader creates fix tasks.
-2. **Second execution onwards**: The mission runtime **clones** the discovery task from the first execution (same prompt, fresh run). The cloned task checks the latest CI state at that point in time — it does not repeat old results.
-3. **Fix tasks are NOT cloned** — each execution's fix tasks are created fresh by the Leader based on what the discovery task finds in that specific run.
+2. **Second execution onwards**: The mission runtime **clones** the discovery task from the first execution (same prompt, fresh run). The cloned task checks the latest CI state at that point in time -- it does not repeat old results.
+3. **Fix tasks are NOT cloned** -- each execution's fix tasks are created fresh by the Leader based on what the discovery task finds in that specific run.
 
 This means: if execution #1 finds 3 failures and the Leader creates 3 fix tasks, execution #2 will only create new fix tasks for failures that still exist (or new failures introduced since the last run). Fixed tests should no longer appear in the discovery output.
 
@@ -134,7 +173,7 @@ For subsequent hourly executions, the mission system reuses the plan from the fi
 The mission uses `semi_autonomous` autonomy, which means:
 - Fix tasks **must create a PR** via `gh pr create` targeting `dev` (the standard supervised workflow for code changes).
 - The Leader can **merge approved PRs** without waiting for human confirmation.
-- This does NOT mean direct push to `dev` — all code changes go through PRs for traceability.
+- This does NOT mean direct push to `dev` -- all code changes go through PRs for traceability.
 
 > **Why PRs instead of direct push**: Even though `semi_autonomous` allows skipping human review on merge, PRs provide a commit history trail, enable CI verification before merge, and avoid the `cancel-in-progress` race condition that direct pushes to dev trigger.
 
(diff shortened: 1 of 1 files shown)


lsm (commented):

## Plan Review: Dev Branch E2E Tests Guardian (Hourly)

Overall: **APPROVE** with the following issues to address before or shortly after the first execution.

---

### P1 — Missing `build`/`discover` prerequisite check

**File:** Mission description, step 3

The plan queries for failed E2E jobs (`conclusion == "failure"`) but does not first verify that the `build` and `discover` jobs succeeded. In the CI workflow, E2E jobs have `needs: [build, discover]` and a conditional `if: always() && needs.build.result == 'success' && needs.discover.result == 'success'`. If `discover` fails, the E2E matrix job is **skipped** (not failed), so `gh run view --json jobs` will simply not include it in the list at all. The plan would then reach step 4 ("no E2E jobs failed") and report "All green" — even though no tests ran.

**Fix:** Before step 3, add a check equivalent to the old plan's step 3:
```bash
gh run view --repo lsm/neokai $RUN_ID --json jobs \
  --jq '.jobs[] | select(.name == "Discover Tests" or .name == "build") | {name, conclusion}'
# If Discover Tests or build is not "success", report "E2E tests did not run (build/discover failure)" and stop.
```

---

### P1 — Silent `gh run download` failures hide non-404 errors

**File:** Mission description, step 5 (batch download loop)

```bash
gh run download ... --dir "/tmp/e2e-guardian/$TEST_NAME" 2>/dev/null || echo "  Artifact not found"
```

`2>/dev/null` suppresses **all** stderr — including authentication failures, network timeouts, and server errors. The plan only handles the 404 case ("artifact not found"), but if `gh run download` fails for any other reason, the agent proceeds without any failure data and with no indication anything went wrong.

**Fix:** Remove `2>/dev/null` and let the `|| echo` handle the non-zero exit code more explicitly:
```bash
if ! gh run download ... --dir "/tmp/e2e-guardian/$TEST_NAME" 2>&1; then
  echo "  Artifact download failed (may have expired)"
fi
```

---

### P1 — Polling loop can exit 0 on network errors

**File:** Mission description, step 2

The polling script exits 0 if it determines CI is still in-progress after 5 waits. But if `gh run list` fails due to network or auth issues at any point, `$CI_STATUS` will be empty, the `for` loop breaks, and the outer `if [ "$CI_STATUS" = "in_progress" ]` is false — so execution continues to step 3 with an empty `$CI_STATUS`. This is treated as "CI completed" rather than "CI status unknown."

**Fix:** Validate that `$CI_STATUS` is non-empty:
```bash
if [ -z "$CI_STATUS" ]; then
  echo "Failed to query CI status. Will retry next hour."
  exit 0
fi
```

---

### P2 — "All recent runs cancelled" is under-specified

**File:** Mission description, step 2

The 5-run limit is somewhat arbitrary. If the most recent **completed** run is 4 hours old, the agent will still report on it — which may be confusing for a mission billed as "hourly." Consider checking more runs (e.g., `--limit 10`) or document that a stale result is acceptable for a guardian mission.

---

### P2 — No handling of concurrent runs on same commit

**File:** Mission description, step 2

A new push during the polling window could invalidate results. Consider adding a `headSha` check after polling to detect new pushes.

---

### P2 — Known pre-existing issues list may become stale

**File:** Mission description, "Known Pre-existing Issues"

The static list will diverge from `docs/e2e-health-check-log.md` over time. Consider removing the hard-coded list and relying solely on the health check log as the source of truth.

---

### P3 — Excluded tests section is redundant with CI

**File:** Mission description, "Excluded Tests"

`features/space-export-import` and `features/space-workflow-rules` are already excluded from CI's `discover` job. They will never appear as failures. The exclusion in the plan is harmless but unnecessary.

---

### What the plan gets right

- **Artifact naming is correct:** CI uses `e2e-no-llm-results-<safe_name>` and `e2e-results-llm-<safe_name>` where safe_name = test path with `/` replaced by `-`. Verified against workflow lines 791-792 and 908.
- **Artifact retention is correctly stated as 7 days** (confirmed in workflow at lines 912 and 1033 with `retention-days: 7`).
- **Cancelled run handling is correct** (`cancel-in-progress: true` at workflow line 34, cancelled runs skipped).
- **`gh run download` limitation (one `--name` per invocation) is correctly handled** with the per-job loop.
- **Artifact contents are correctly described** (`test-results/` JSON and `playwright-report/` HTML, not monocart).
- **Fix task creation guidance is sound:** one task per distinct failure, clear acceptance criteria, `coder` vs `general` type, PR workflow, health check log cross-reference.
- **Plan reuse (task cloning) mechanism is correctly explained.**

---

### Summary

The plan is well-structured and the corrections to the previous version (CI artifact handling, reporter type, `gh run download` limitation, cancelled run handling) are all accurate and verified against `.github/workflows/main.yml`. The three P1 issues should be fixed before the first execution to avoid false "all green" reports or silent data loss.

lsm (commented):

### Overall Assessment

The plan is well-researched and the CI artifact corrections (fixing `gh run download` usage, clarifying the monocart reporter, artifact retention) are accurate and important. The adaptive discovery-then-fix approach is a sound simplification over the previous rigid multi-task plan. However, there are several gaps and inaccuracies that need to be addressed before this plan is actionable.

---

### P1 Issues (Blocking)

**1. Missing build/discover verification before investigating E2E failures**

The plan's Step 3 picks the most recent completed run and immediately starts inspecting E2E job failures. It never checks whether `build` and `discover` jobs succeeded. The original health check plan (`docs/plans/dev-branch-e2e-tests-health-check.md`, Task 1, step 3) explicitly verifies this because if `build` or `discover` failed, E2E tests never ran at all. The current plan's step 3 jq filter would simply return no failed E2E jobs in that case and report "All green" — a false negative.

The plan should add a check analogous to the original:
```bash
gh run view --repo lsm/neokai $RUN_ID --json jobs \
  --jq '.jobs[] | select(.name == \"build\" or .name == \"discover\") | \"\(.name): \(.conclusion)\"'
```
If either is not `"success"`, report "Build/discover failed — E2E tests did not run" and stop.

**2. Incorrect `test-results/.last-run.json` reference**

The plan's Step 5 Option B says: "Read test-results/ for structured failure data, or playwright-report/ for HTML summary." This is fine. However, the introductory paragraph before the mission description also says: "download the test report artifact: ... Read test-results/ for structured failure data." The health check log correctly identifies that CI uses **monocart-reporter**, not the Playwright HTML reporter.

The monocart-reporter creates `test-results/<test-name>.data.browser-context.test-results.json` files, not a `.last-run.json` at the root of test-results/. There is no centralized JSON summary file. The plan should either:
- Name the actual monocart output files (e.g., `test-results/<name>.data.browser-context.test-results.json`)
- Or clarify that agents should use `playwright-report/index.html` (the HTML report with inline data) for human-readable output and the per-test JSON files for programmatic parsing

**3. No guidance on how the Leader reads discovery output to create fix tasks**

The "Fix Task Creation Guidance" section (lines 156-228) describes what the Leader should do but never explains the actual mechanics: how does the Leader agent, after the discovery task completes, read the failure report? Specifically:
- Is the discovery task output stored in the execution record, and does the Leader have access to it?
- Or does the Leader read `docs/e2e-health-check-log.md`?
- The plan mentions the Leader calls `mcp__planner-tools__create_task` but provides no details about what parameters to pass (task type, description format, depends_on linking, etc.)

Without this, the fix task creation guidance is aspirational but not executable by an autonomous agent.

---

### P2 Issues (Should-Fix)

**4. Artifact download loop silently ignores all failures**

Step 5 Option B iterates over failed job names downloading artifacts with `gh run download ... 2>/dev/null || echo "  Artifact not found"`. If `gh run download` fails for every job (e.g., a bug in the artifact naming extraction, or all artifacts expired), the loop silently continues and the agent reports "All green" or produces an empty failure report. There should be a guard: if no artifacts were successfully downloaded and failures were expected, report a warning rather than silently producing an empty report.

**5. `sed` extraction assumes consistent job naming**

The plan uses `TEST_NAME=$(echo "$JOB" | sed 's/.*(\(.*\))/\1/')` to extract artifact names. For job names like `E2E No-LLM (features-mission-terminology)`, this correctly yields `features-mission-terminology`. However, if a job name has nested parentheses or a different format, this silently extracts the wrong value and the `gh run download --name` fails (since `--name` requires an exact match). The plan should validate the extraction or use a more robust parsing approach (e.g., JSON-based extraction instead of text-based sed).

**6. Known pre-existing issues list may be stale**

The plan lists "Neo panel doesn't close on Escape key (affects 7 suites, first seen 2026-04-02)" and other issues from the 2026-04-02 health check. If this plan is used for recurring hourly runs, the known issues list will grow stale. The plan should include a note that agents should re-check the health check log at `docs/e2e-health-check-log.md` rather than relying on the snapshot in the mission description, and that the mission description should be updated periodically (e.g., when issues are resolved or new ones are added).

**7. No criteria for when to stop polling and report an in-progress CI run**

Step 2 polls up to 10 minutes. After that, it says `echo "CI still running after 10 minutes. Reporting status and stopping."` and exits 0. But what does "reporting status" mean in practice? It should append to the health check log that the run was in-progress and no data was available, rather than exiting silently. An empty failure report could be misread as "all green."

---

### P3 Issues (Nit)

- The `sed` substitution on line 95 would produce incorrect results if a job name contains multiple parentheses (e.g., `E2E No-LLM (features-foo (special)))` — unlikely but possible.
- The artifact retention note (7 days) is accurate based on the workflow, but the plan doesn't explain what happens when a failure is found for an artifact that has already expired — the agent should still report the failure but note that artifact logs are no longer available for deep inspection.

---

### What Works Well

- The cancelled run handling (skipping `conclusion: "cancelled"`, polling in-progress, all-5-cancelled edge case) is correct and well-thought-out.
- Limiting to the last 5 runs is a reasonable balance between thoroughness and API usage.
- The categorization guidance (`test-bug`, `product-bug`, `flaky`, `env`) and the cross-reference with the health check log are solid.
- The exclusion of `space-export-import` and `space-workflow-rules` is correctly documented.
- The `semi_autonomous` autonomy level and PR-based workflow for fix tasks are appropriate.
- The plan reuse / task cloning mechanism for subsequent executions is clearly explained.

---

### Summary

The plan is a significant improvement over the previous multi-task approach. The corrections to `gh run download` usage, the clarification of the monocart reporter, and the cancelled/in-progress edge case handling are all accurate. The main gaps are: (1) missing build/discover verification that could cause false negatives, (2) the non-existent `.last-run.json` reference suggesting confusion between monocart and Playwright HTML reporter output structures, and (3) no mechanical explanation of how the Leader reads discovery output to create fix tasks. Fixing these three issues will make the plan fully actionable for an autonomous agent.

lsm:
## P1 Fixes

### P1-1: Missing build/discover prerequisite check
Added **step 3** — a new prerequisite verification step that runs before checking for E2E failures. It:
- Checks `build` and `discover` job statuses via `gh run view`
- Counts E2E jobs that actually ran vs. were skipped
- Reports "CI infrastructure issue" (not "all green") when E2E jobs were skipped due to upstream failures
- Updated acceptance criteria to include this check

### P1-2: test-results/.last-run.json confusion
Replaced the vague artifact description with:
- Confirmed artifact directory structure (from CI workflow `upload-artifact@v4`)
- Concrete `.last-run.json` path and `jq` extraction command for structured error messages
- Explained that CI uses Playwright HTML reporter (not monocart), and `test-results/` contains per-test Playwright internal output
- Split into step 6a (fast job log query) and step 6b (artifact download with parsing)

### P1-3: Discovery output → Leader data flow
Replaced the abstract mission system description with:
- ASCII flow diagram: discovery output → Leader reads → create_task() → dispatch
- Table of expected failure report fields with examples
- Step-by-step Leader actions: skip pre-existing, group by root cause, create tasks with specific `title`/`agent`/`priority` values
- Deduplication logic: skip failures already covered by in-progress fix tasks from this or prior executions

## P2 Fixes
- **Network error handling**: Empty `$CI_RESULT` check with explicit error message and exit 1
- **`2>/dev/null` removed**: Artifact download errors now shown; graceful fallback to step 6a
- **5-run limit explained**: Documents cancel-in-progress burst scenario
- **Job name extraction**: Added guard for unexpected formats with warning message

lsm (commented):

Reviewed the updated plan at `docs/plans/dev-branch-e2e-tests-guardian-hourly.md`. All previously flagged P1 issues are resolved.

### P1-1 (E2E job verification): RESOLVED
Step 3 now properly verifies that E2E jobs actually ran before checking for failures:
- Checks upstream jobs (`Build Binary (linux-x64)` and `Discover Tests`) via explicit jq filtering
- Computes `E2E_RAN = E2E_TOTAL - E2E_SKIPPED`
- Reports CI infrastructure issue if `E2E_RAN -eq 0` (not "all green")
- Correctly distinguishes skipped-due-to-gating from genuinely passing

### P1-2 (artifact parsing): RESOLVED
Critical fixes implemented:
- Replaced invalid `--pattern` dual-flag (not supported by `gh run download`) with per-job download loop using `--name` and `--dir`
- Artifact directory structure clearly documented: `test-results/<test-name>/.last-run.json`
- jq command uses proper optional chaining (`?.`) with `// empty` fallback for missing paths
- Artifact retention correctly documented as 7 days (not the 90-day default)
- Batch download loop correctly extracts test names via `sed -n 's/.*(\(.*\)).*/\1/p'`

### P1-3 (Leader data flow): RESOLVED
The section is now clear and actionable:
- Data flow diagram: Discovery task -> failure report -> Leader reads output -> create_task() calls
- Exact failure report format specified (table with 6 fields: test path, job name, error message, commit hash, root cause, pre-existing flag)
- Leader action rules with skip/group/create logic, agent type mapping (`coder` for test-bug/product-bug, `general` for flaky/env), and priority mapping
- Explicit cross-reference requirement to `docs/e2e-health-check-log.md` before creating fix tasks

### P2 issues: All addressed
- **Network error handling**: `if [ -z "$CI_RESULT" ]` check before parsing, polling continues on WARNING (not exit) for transient failures
- **Polling improvements**: Checks both `status` and `conclusion` fields, proper loop exit with `[ "$CI_STATUS" != "in_progress" ] && break`
- **Cancelled run handling**: Explicit filtering logic with up to 5-run fallback, clear message when all runs are cancelled/in-progress
- **Artifact naming convention**: "Strip the prefix and parentheses to get the artifact name suffix" is explicitly documented

### Minor observations (P3)
- Step 2 polling loop uses `continue` on network warnings but the outer `for` loop will still sleep before the next attempt -- this is correct behavior but the comment could say "Retrying next iteration" more clearly (already does)
- The `sed` extraction pattern `s/.*(\(.*\)).*/\1/p` assumes exactly one set of parentheses -- holds for current naming convention

**Recommendation: APPROVE**

The plan is now solid and ready for implementation. All P1 issues are genuinely resolved with correct jq syntax, proper CLI usage, and clear agent workflow documentation.

lsm (commented):

Reviewed against the three P1 issues flagged in the prior review.

---

### P1-1 (E2E job verification): RESOLVED

Step 3 now properly verifies that E2E jobs actually ran:
- Counts total E2E jobs (`E2E_TOTAL`), skipped E2E jobs (`E2E_SKIPPED`), and computes `E2E_RAN = E2E_TOTAL - E2E_SKIPPED`
- If `E2E_RAN` is 0, reports a CI infrastructure warning and exits with code 0
- Also prints upstream job status (`Build Binary (linux-x64)` and `Discover Tests`) for diagnostics

The logic correctly detects the build/discover gating case. The `if [ "$E2E_RAN" -eq 0 ]` guard fires before the "all green" path, preventing false negatives. Job names match the actual CI workflow (`Build Binary (linux-x64)` / `Discover Tests`). This is solid.

---

### P1-2 (artifact parsing): NOT FULLY RESOLVED (P2 concern)

The plan correctly designates step 6a (job logs) as "fast, recommended first" and step 6b (artifact) as secondary detail extraction. However, the `.last-run.json` extraction has accuracy problems:

1. **Directory structure mismatch**: The plan describes `test-results/<test-name>/.last-run.json`, but Playwright's actual default `testResultsDir` structure uses hash-based directory names inside `test-results/`, not a `<test-name>` subdirectory. The artifact uploads `packages/e2e/test-results/` (no subdirectory), which when downloaded becomes `test-results/` with Playwright's internal naming.

2. **jq extraction path may not match**: The path `.suites[]?.specs[]?.tests[]?.results[]?.error // empty | .message // empty` is unlikely to match Playwright's actual `.last-run.json` schema. The `// empty` operator with `jq -r` will silently output nothing for non-matching paths rather than surfacing errors.

**Why this is P2 not P1**: Step 6a (job logs) is the primary extraction method and works independently of the artifact structure. The agent would fall back to job logs if step 6b produces no output. The batch download loop also has an explicit fallback message. The plan is designed to succeed even when artifact parsing fails.

---

### P1-3 (Leader data flow): RESOLVED

The new "How the Leader Receives Discovery Output" section (lines 232-269) is clear and actionable:
- Data flow diagram: discovery task output -> Leader reads task output -> `create_task()` per failure
- Field table with exact expected fields (test file path, CI job name, failure message, commit hash, root cause category, pre-existing flag)
- Explicit Leader action items: skip pre-existing, group by root cause, call `create_task()` with title/description/agent/priority
- Excluded tests are explicitly skipped

The mechanism is now unambiguous. The critical path (discovery task runs -> failure report produced -> Leader reads output -> fix tasks created) is fully documented.

---

### P2 Fixes: All Adequately Addressed

- **Network error handling in polling**: The polling loop now checks for empty output (`if [ -z "$POLL_RESULT" ]`) and continues to the next iteration with a warning. The initial query also checks for empty `CI_RESULT` and exits with error code 1. This is correct.
- **Polling improvements**: Progress messages are informative ("Waiting 2 minutes (attempt $i/5)..."). Clear exit states: "CI still running after 10 minutes. Reporting current status and stopping."
- **sed pattern fix**: Changed from `sed 's/.*(\(.*\))/\1/'` to `sed -n 's/.*(\(.*\)).*/\1/p'` -- correctly anchors the pattern and avoids empty output on mismatch.
- **Artifact download fallback**: `2>&1` redirects stderr so the failure message is visible, with a clear "use step 6a logs instead" guidance.
- **Upstream job check**: `UPSTREAM` variable is computed and echoed but not programmatically acted upon -- however, the `E2E_RAN = 0` check catches the practical case where upstream failures cause E2E to be skipped.

---

### Remaining P3 nits (non-blocking)

- The `exit 0` for "no E2E ran" and "CI still in progress" means these cases don't appear as failures in mission execution tracking, only in the report text. This is a design choice (not flooding with alerts for transient CI states) rather than a bug.
- The acceptance criteria (line 337) is missing a check that the upstream job names in the script exactly match the actual CI job names (they do, verified against `.github/workflows/main.yml`).

---

### Recommendation

**APPROVE**. All three P1 issues are resolved: E2E gating is properly detected, artifact parsing works via the correct primary path (job logs), and the Leader data flow is fully documented. The P2 artifact structure concern does not block the plan's primary operation.

lsm:
## CI Failure Analysis

### Summary
The CI failures are **pre-existing flaky test issues**, unrelated to this docs-only PR.

### Failing test
- **Job**: `Daemon Online (cross-provider-1)`
- **Test**: `Cross-Provider Conversation Continuity After Model Switch > sdkSessionId unchanged after post-switch message completes (cross-provider)`
- **Error**: `Timeout waiting for processing state "idle" after 60000ms`
- **File**: `packages/daemon/tests/online/cross-provider/conversation-continuity-after-switch.test.ts`

### Evidence this is pre-existing
1. **Passed on a different PR** (`task/db-migrations`, run #23907532115) just 3 minutes before our run started — all 3 cross-provider jobs passed there.
2. **Failed on rerun**: Re-ran the failed jobs and `cross-provider-1` failed again with the exact same test and timeout error. This is a **flaky online test** — it depends on external API response times (MiniMax + GLM providers).
3. **5 of 6 tests pass**: The failing test (`sdkSessionId unchanged after post-switch message completes`) times out waiting for the "idle" processing state, which is a race condition with external API latency. The other 5 tests in the same file pass fine.

### Conclusion
This is a **flaky cross-provider online test** — not a regression. It fails intermittently due to external provider API latency exceeding the 60s timeout. This PR only changes `docs/plans/dev-branch-e2e-tests-guardian-hourly.md` and cannot affect daemon online tests.

**Recommendation**: This flaky test should be tracked as a pre-existing issue in the E2E guardian's health check log (or the online test equivalent). The PR is safe to approve/merge.

PR #1207, opened 2026-04-02
fix: update migration tests for M74 column drops

lsm:
Update 5 migration test files to reflect M74 schema changes:

- **migration-29**: Fix role assertion (dropped by M74), remove idx_space_workflow_nodes_order from expected indexes, remove role/order_index from INSERT statements
- **migration-33**: Update fresh DB tests to verify inject_workflow_context/role are absent after M74, remove dropped columns from INSERT/SELECT statements
- **migration-35-36**: Change max_iterations assertion on space_workflows to verify column is absent (M74 drops it)
- **migration-45**: Change idx_space_workflow_nodes_order assertion to not.toContain, remove max_iterations from SELECT in preserves-columns test
- **migration-53**: Replace config column check with tags column check, remove config from post-migration assertions


lsm (commented):

---

### P0 — Schema bug: NOT NULL + ON DELETE SET NULL on node_executions.agent_id

**File:** `packages/daemon/src/storage/schema/migrations.ts` lines 4955 and 4965

The `node_executions` table declares:
```sql
agent_id TEXT NOT NULL,
...
FOREIGN KEY (agent_id) REFERENCES space_agents(id) ON DELETE SET NULL
```

These two constraints are incompatible. When a `space_agents` row is deleted, SQLite will attempt to SET `node_executions.agent_id = NULL`, which immediately violates the NOT NULL constraint. The result is:

```
Error: NOT NULL constraint failed: node_executions.agent_id
```

This means any call to delete a space agent that has associated node executions will fail at the database level. The fix is either:
- Change `agent_id` to `agent_id TEXT` (nullable), or
- Change the FK action to `ON DELETE CASCADE` or `ON DELETE RESTRICT`, depending on the intended semantics.

This was verified by running the schema in isolation under Bun SQLite.

---

### P1 — node_executions creation is not atomic

**File:** `packages/daemon/src/storage/schema/migrations.ts` lines 4947-4974

The CREATE TABLE and two CREATE INDEX calls for `node_executions` are separate, non-transactional `db.exec` calls with no BEGIN/COMMIT wrapper. If the first index creation succeeds but the second fails (or the process crashes in between), the table exists but is missing the second index. The idempotency guard (`if (!tableExists(...))`) then skips the entire block on the next run, leaving the table permanently without that index.

The other three table rebuilds in M74 (space_workflows, space_workflow_nodes, space_agents) all correctly use BEGIN/COMMIT; node_executions should too.

---

### P1 — No test covers the agent-deletion FK behavior

**File:** `packages/daemon/tests/unit/storage/migrations/migration-74_test.ts`

The test suite verifies the status CHECK constraint (line 262) but has no test that deletes a `space_agents` row while `node_executions` rows reference it. Because of the NOT NULL + ON DELETE SET NULL bug described above, such a test would fail. A test covering referential behavior on agent deletion is needed to catch this class of regression going forward.

---

### P2 — Tags extraction uses N UPDATE queries instead of a single-pass INSERT

**File:** `packages/daemon/src/storage/schema/migrations.ts` lines 5019-5039

The INSERT on line 5024 writes `'[]'` for all rows, then the loop on lines 5037-5038 runs one UPDATE per workflow. For a database with many workflow rows this is N+1 queries instead of 1. A more efficient approach would use the `tagsMap` values directly in a parameterized INSERT `... VALUES (?, ?, ...)?` or merge the tag extraction logic into the SELECT via a `json_extract(config, '$.tags')` expression. This is a correctness-preserving refactor that reduces write amplification.

---

### Summary of task description vs actual implementation

The reviewer task description stated item #1 as: "node_executions backfilled from space_tasks WHERE workflow_node_id IS NOT NULL". This claim is incorrect in the implementation — no such backfill exists, and none is possible: M73 already dropped the `workflow_node_id` column from `space_tasks`. The table is created empty. The implementation is correct on this point; the task description was misleading. The code, tests, and in-file docblock all consistently describe an empty table creation.

Item #6 ("space_tasks rebuild with dropped columns and remapped status values") was implemented in M73, not M74. M74 does not touch space_tasks at all. The test file and the migration's docblock correctly describe M74's actual scope.

---

### Minor notes

- The `_index` parameter in `SpaceWorkflowRepository.insertNode` (line 333) is now unused following the removal of `order_index`. This is a cosmetic leftover; the leading underscore signals intentional non-use.
- All three table-rebuild sub-migrations correctly use `PRAGMA foreign_keys = OFF` + transaction; the pattern is consistent with earlier migrations in the file.
- Idempotency guards for the three table rebuilds (checking a representative dropped column) are sound given that the column sets were always added together in prior migrations.

lsm (commented):

Overall this is a solid, well-structured migration. The table rebuilds follow the correct SQLite pattern (PRAGMA foreign_keys OFF, BEGIN, create new table, INSERT...SELECT, DROP old, RENAME, recreate indexes, COMMIT, PRAGMA ON). Idempotency guards are in place for every sub-migration. Tests are comprehensive (19 tests, all passing) and cover schema changes, data preservation, JSON migration, and idempotency. All 217 migration tests and 1930 space unit tests pass.

### P0: `node_executions.agent_id` is NOT NULL but FK action is SET NULL

**File:** `packages/daemon/src/storage/schema/migrations.ts` line 4955/4965
**File:** `packages/daemon/tests/unit/helpers/space-test-db.ts` line 151/161

```sql
agent_id TEXT NOT NULL,
...
FOREIGN KEY (agent_id) REFERENCES space_agents(id) ON DELETE SET NULL
```

`agent_id` is declared `NOT NULL` but the foreign key specifies `ON DELETE SET NULL`. If a `space_agents` row is deleted while `node_executions` rows reference it, SQLite will attempt to set `agent_id = NULL`, which violates the NOT NULL constraint and raises a runtime error.

Fix: either make `agent_id` nullable (`TEXT` without NOT NULL) or change the FK action to `ON DELETE CASCADE` or `ON DELETE RESTRICT`, depending on the desired behavior when agents are deleted.

### P3: PR title is misleading

The PR title is "fix: update migration tests for M74 column drops" which describes only the first commit. The primary payload is the M74 migration itself (`feat(daemon): add M74 migration`). Consider updating to reflect the full scope, e.g., `feat(daemon): add M74 migration — schema cleanup for workflow completion`.

---

**Recommendation:** REQUEST_CHANGES due to the P0 NOT NULL / SET NULL conflict on `node_executions.agent_id`.

lsm (commented):

**Recommendation: APPROVE**

Re-review after fixes for the two issues raised in the previous review.

### P0 fix: `node_executions.agent_id` made nullable — VERIFIED

`migrations.ts` line 4957 now declares `agent_id TEXT` (no `NOT NULL`), which is compatible with the `ON DELETE SET NULL` foreign key action on line 4967. The new test at line 301 (`deleting space_agent sets agent_id to NULL`) correctly inserts a row with a non-null `agent_id`, deletes the referenced agent, and asserts the column becomes `NULL`. All 20 migration-74 tests pass.

### P1 fix: `node_executions` table creation wrapped in transaction — VERIFIED

Lines 4949-4981 now wrap the `CREATE TABLE`, both `CREATE INDEX` statements, and `COMMIT` inside a `try/catch` with `ROLLBACK` on error. The idempotency guard (`tableExists` check) sits outside the transaction, which is correct — if the table already exists, the block is skipped entirely; if creation fails midway, indexes won't be left behind.

No remaining issues found. Clean implementation.

lsm (commented):

**Recommendation: APPROVE**

Re-review of PR #1207 — both previously-raised issues have been correctly addressed in commit b53cbc8.

---

### P0 fix — agent_id made nullable: CORRECT

`packages/daemon/src/storage/schema/migrations.ts` line 4957:
```sql
agent_id TEXT,
```
Changed from `TEXT NOT NULL` to `TEXT` (nullable). This is the necessary fix — `ON DELETE SET NULL` requires the column to accept NULL, otherwise SQLite would raise a constraint violation when the referenced agent row is deleted.

The companion test in `migration-74_test.ts` (`node_executions: deleting space_agent sets agent_id to NULL`) correctly validates the behavior end-to-end: foreign keys are enabled in `beforeEach`, a node execution is created with a live agent_id, the agent is deleted, and the test asserts `agent_id` becomes null without an error.

The test helper schema in `packages/daemon/tests/unit/helpers/space-test-db.ts` line 151 was also updated to `agent_id TEXT` — consistent with the migration.

---

### P1 fix — node_executions creation wrapped in transaction: CORRECT

`packages/daemon/src/storage/schema/migrations.ts` lines 4949–4980: The `CREATE TABLE` and both `CREATE INDEX` statements are now wrapped in `BEGIN` / `COMMIT` with a `ROLLBACK` on error inside the `if (!tableExists(...))` guard. This matches the pattern used by the other sub-migrations in M74 (space_workflows at line 5006, space_workflow_nodes at line 5063, space_agents at line 5104).

One minor observation: the `runMigrations` runner does not wrap individual migration calls in a transaction, so each migration (including M74) is responsible for its own atomicity. The `try/catch/ROLLBACK` here is the correct pattern for this codebase. No outer transaction conflict is possible since `runMigrations` itself does not `BEGIN` a transaction.

---

### No regressions observed

- The `space-test-db.ts` helper schema is consistent with the migration schema.
- The idempotency guard (`if (!tableExists(db, 'node_executions'))`) is intact, so re-running the migration is safe.
- The full migration chain test (`node_executions is created via full migration chain`) passes through `runMigrations` and confirms the table exists.

No new issues found.

lsm (commented):

**Recommendation: APPROVE**

The fix in `channel-router-async.test.ts` is correct and consistent with the M74 migration changes.

### What changed

The `seedAgent` helper function (line 58-69) was updated to:
1. Remove `role` and `config` from the INSERT statement — these columns were dropped from `space_agents` by M74
2. Make the `role` parameter optional (`_role?`) and prefix with underscore to indicate it is unused — this avoids changing all call sites while correctly reflecting the column no longer exists

### Verification

- The `space_agents` schema in `space-test-db.ts` (lines 43-57) confirms `role` and `config` are absent post-M74
- All call sites (lines 133, 134, 513, 582, 862, 914) still compile — the parameter is accepted but ignored
- The INSERT columns (`id, space_id, name, description, model, tools, system_prompt, created_at, updated_at`) match the post-M74 schema exactly
- The approach is consistent with `seedAgent` fixes in the other test files in this PR (e.g., `channel-router.test.ts`)

No issues found.

lsm (commented):

**Recommendation: APPROVE**

The CI fix in `channel-router-async.test.ts` is correct. Here is what was verified:

**Schema alignment**

M74 drops `role`, `config`, and `inject_workflow_context` from `space_agents` by recreating the table without those columns (see `migrations.ts` lines 5101-5144). The fixed INSERT statement matches the post-M74 schema exactly:

```sql
INSERT INTO space_agents (id, space_id, name, description, model, tools, system_prompt,
     created_at, updated_at)
     VALUES (?, ?, ?, '', null, '[]', '', ?, ?)
```

This is identical to the version already used in the sibling `channel-router.test.ts` (line 79).

**Parameter change**

The `role` parameter was renamed to `_role?` (optional, underscore-prefixed). All six call sites still compile because the parameter is now optional — callers that pass `'coder'` or `'planner'` simply have the value silently ignored. The underscore prefix correctly signals intentional non-use. Since `**/*.test.ts` files are in Oxlint's `ignorePatterns`, the unused parameter will not trigger a lint error.

**No issues found.** The fix is minimal, targeted, and consistent with the rest of the M74 test cleanup already done in the branch.

PR #1208, opened 2026-04-02
test: add edge-case tests for gate badge customization

lsm:
## Summary
- Add 7 edge-case tests for gate `label`/`color`/`hasScript` propagation in `semanticWorkflowGraph.test.ts` (19 → 26 tests)
- Covers multiple gates per direction (first-wins `??=` for label/color, any-true for `hasScript`), hybrid gates (fields + script), bidirectional edges with multiple gates per direction, and script-only gates resolving to `check` type
- All existing 19 tests continue to pass — no regressions

### DIFF
--- a/packages/web/src/components/space/visual-editor/__tests__/semanticWorkflowGraph.test.ts
+++ b/packages/web/src/components/space/visual-editor/__tests__/semanticWorkflowGraph.test.ts
@@ -459,4 +459,163 @@ describe('gate label/color/hasScript propagation', () => {
 			gateColor: '#998877',
 		});
 	});
+
+	it('uses first gate label/color when multiple gates exist on the same direction', () => {
+		// Two channels from Planning→Review (different agents), each with a gate.
+		// Implementation uses ??= (nullish coalesce assignment) so first gate wins.
+		const channels: WorkflowChannel[] = [
+			{ from: 'Planning', to: 'Reviewer 1', direction: 'one-way', gateId: 'gate-a' },
+			{ from: 'Planning', to: 'Reviewer 2', direction: 'one-way', gateId: 'gate-b' },
+		];
+		const gates = [
+			humanGate('gate-a', { label: 'First', color: '#aa0000' }),
+			humanGate('gate-b', { label: 'Second', color: '#00bb00' }),
+		];
+
+		const result = buildSemanticWorkflowEdges(NODES, channels, gates);
+		expect(result).toHaveLength(1);
+		expect(result[0]).toMatchObject({
+			direction: 'one-way',
+			channelCount: 2,
+			// First gate's label/color wins (??= behavior)
+			gateLabel: 'First',
+			gateColor: '#aa0000',
+		});
+	});
+
+	it('sets hasScript to true when any gate on the same direction has a script', () => {
+		// Two channels same direction: first gate has no script, second gate has script.
+		// Implementation uses `if (gateInfo.hasScript)` which is true if ANY gate has a script.
+		const channels: WorkflowChannel[] = [
+			{ from: 'Planning', to: 'Reviewer 1', direction: 'one-way', gateId: 'gate-no-script' },
+			{ from: 'Planning', to: 'Reviewer 2', direction: 'one-way', gateId: 'gate-with-script' },
+		];
+		const gates = [
+			humanGate('gate-no-script', { label: 'No Script' }),
+			humanGate('gate-with-script', { script: { interpreter: 'bash', source: 'exit 0' } }),
+		];
+
+		const result = buildSemanticWorkflowEdges(NODES, channels, gates);
+		expect(result[0]).toMatchObject({
+			hasScript: true,
+		});
+	});
+
+	it('gate with both fields and script propagates both gateType and hasScript', () => {
+		// A human gate (with approved field) that also has a script.
+		// Should produce gateType='human' AND hasScript=true.
+		const channels: WorkflowChannel[] = [
+			{ from: 'Planning', to: 'Code Review', direction: 'one-way', gateId: 'g1' },
+		];
+		const gates = [
+			humanGate('g1', {
+				label: 'Hybrid Gate',
+				color: '#112233',
+				script: { interpreter: 'python3', source: 'exit(0)' },
+			}),
+		];
+
+		const result = buildSemanticWorkflowEdges(NODES, channels, gates);
+		expect(result[0]).toMatchObject({
+			gateType: 'human',
+			gateLabel: 'Hybrid Gate',
+			gateColor: '#112233',
+			hasScript: true,
+		});
+	});
+
+	it('first gate label/color wins on bidirectional edge with multiple gates per direction', () => {
+		// Multiple channels in each direction of a bidirectional edge.
+		// Forward: two gates (gate-fwd-1 with label/color, gate-fwd-2 with different label/color)
+		// Reverse: one gate with its own label/color
+		const channels: WorkflowChannel[] = [
+			{ from: 'Planning', to: 'Reviewer 1', direction: 'one-way', gateId: 'gate-fwd-1' },
+			{ from: 'Planning', to: 'Reviewer 2', direction: 'one-way', gateId: 'gate-fwd-2' },
+			{ from: 'Reviewer 3', to: 'Planning', direction: 'one-way', gateId: 'gate-rev' },
+		];
+		const gates = [
+			humanGate('gate-fwd-1', { label: 'Fwd 1', color: '#111111' }),
+			humanGate('gate-fwd-2', { label: 'Fwd 2', color: '#222222' }),
+			checkGate('gate-rev', { label: 'Rev', color: '#333333' }),
+		];
+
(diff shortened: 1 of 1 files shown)


lsm (commented):

**Recommendation: APPROVE**

All 7 new tests are well-written and correctly verify the edge cases for gate `label`/`color`/`hasScript` propagation in `buildSemanticWorkflowEdges`. I verified each test against the implementation in `semanticWorkflowGraph.ts` and confirmed the assertions match the actual code paths:

- **First-wins (`??=`) semantics** (lines 463-484, 527-554): Correctly tests that the first gate's `label`/`color` wins when multiple gates exist on the same direction, matching the `??=` assignments on lines 223-224, 233-234 of the implementation.
- **Any-true `hasScript`** (lines 486-502, 556-575): Correctly tests that `hasScript` becomes `true` if ANY gate in the direction has a script, matching the `if (gateInfo.hasScript)` pattern on lines 225, 235.
- **Hybrid gate** (lines 504-525): Good coverage of a gate with both `fields` and `script`, verifying both `gateType` and `hasScript` are set simultaneously.
- **Script-only gate** (lines 577-595): Verifies the `fields.length === 0` branch (line 96-97) correctly resolves to `'check'` while still setting `hasScript: true`.
- **Multi-agent node** (lines 597-620): Uses `Reviewer 1`/`Reviewer 2` from the `review` node to exercise the `endpointLookup` mapping, confirming different agents on the same step collapse correctly.

No issues found. The tests are thorough, correctly mirror the implementation behavior, and use the existing `humanGate`/`checkGate` helpers consistently.

lsm (commented):

---

**Recommendation: APPROVE** (self-PR — posting as COMMENT per policy)

### What the tests cover

All 7 new tests in the `gate label/color/hasScript propagation` describe block correctly target the implementation in `semanticWorkflowGraph.ts`. I traced every assertion against the implementation and found no incorrect expectations.

Key behaviors verified correctly:

- `??=` first-wins semantics for `label`/`color` across multiple gates on the same direction (lines 463, 527, 597)
- `if (gateInfo.hasScript) aggregate.lowToHighHasScript = true` any-gate logic (lines 486, 556)
- `humanGate` with `script` produces both `gateType: 'human'` and `hasScript: true` (line 504)
- Script-only `checkGate` (no fields) still resolves to `gateType: 'check'` because `fields.length === 0` path (line 577)
- Bidirectional multi-gate per direction with correct forward/reverse split (line 527)
- Multi-agent node sub-agents collapsing to same semantic edge pair (line 597)

The `humanGate` and `checkGate` test helpers are appropriate and reusable.

---

### P2 — Semantic gap in "first gate wins" tests

**Files:** `packages/web/src/components/space/visual-editor/__tests__/semanticWorkflowGraph.test.ts`, lines 463–484 and 527–554

The comments in these tests say "first gate's label/color wins (??= behavior)". This is accurate only when the first gate has a non-null label/color. If the first gate has `label: undefined`, JavaScript's `??=` will propagate the second gate's label (since `undefined ??= 'something'` assigns). None of the tests cover this case — a variant where the first gate has no label and the second does would demonstrate a subtly different outcome. This scenario could surprise a future developer relying on the "first wins" mental model.

Suggested addition (not blocking, but would strengthen documentation value):

```ts
it('second gate label wins when first gate has no label', () => {
  const channels = [
    { from: 'Planning', to: 'Reviewer 1', direction: 'one-way', gateId: 'gate-a' },
    { from: 'Planning', to: 'Reviewer 2', direction: 'one-way', gateId: 'gate-b' },
  ];
  const gates = [
    humanGate('gate-a'),  // no label
    humanGate('gate-b', { label: 'Second' }),
  ];
  const result = buildSemanticWorkflowEdges(NODES, channels, gates);
  expect(result[0].gateLabel).toBe('Second'); // ??= propagates second when first is undefined
});
```

---

### P2 — `toMatchObject({ key: undefined })` is weaker than it appears

**File:** line 589

The test at line 589 uses:

```ts
expect(result[0]).toMatchObject({
  gateType: 'check',
  gateLabel: undefined,
  gateColor: undefined,
  hasScript: true,
});
```

In Vitest, `toMatchObject` with explicit `undefined` values does not distinguish between a key being absent and a key being present with value `undefined`. Both cases pass this assertion. If the intent is to confirm the fields are truly absent/undefined (which it is), using `expect(result[0].gateLabel).toBeUndefined()` or `expect(result[0]).not.toHaveProperty('gateLabel')` would be more explicit. This is consistent with how existing tests in this file already handle similar checks (lines 257–259), so there is precedent for the explicit form.

---

### P3 — `hasScript: true` test does not assert label propagation

**File:** line 486

The test for "any gate with script sets hasScript" has `gate-no-script` with `label: 'No Script'`. Since it is processed first, `gateLabel` in the result will be `'No Script'`. The test only asserts `hasScript: true` which is the intended focus, but the gate name `gate-no-script` implies it only has "no script" — readers may not realize it does carry a label that propagates. A nit: renaming the gate ID to `gate-1` or `gate-without-script` (keeping just the identifier) or dropping the `label` from that gate helper call avoids any potential confusion.

---

### Summary

The implementation is correct, the tests are accurate, and all 7 new edge-case tests meaningfully exercise the actual `??=` / any-gate logic paths in the implementation. The P2 items are improvement suggestions — neither constitutes a bug or incorrect test.

lsm (commented):

**Recommendation: APPROVE**

The 8 new tests (19 → 27 total) are well-structured and directly exercise the gate badge propagation logic in `buildSemanticWorkflowEdges`. The previous review feedback has been addressed:

1. `toMatchObject({ key: undefined })` replaced with explicit `toBeUndefined()` calls in 3 places (lines 257-259, 268-271, 280-283) -- correct fix since `toMatchObject` with `undefined` values passes even when the key is missing entirely, while `toBeUndefined()` actually reads the property and asserts its value.

2. The `??= undefined propagation` test (line 482) correctly validates that when the first gate has no label, the second gate's label wins -- this is the complementary case to the "first-wins" test and exercises a real edge of the `??=` operator.

3. Gate IDs use descriptive names (`gate-fwd-1`/`gate-fwd-2`, `gate-fwd-a`/`gate-fwd-b`, `gate-a`/`gate-b`) that clearly indicate their role in each test scenario.

The tests accurately reflect the implementation patterns:
- `??=` for label/color (first non-nullish wins)
- `if (gateInfo.hasScript)` for script (any-true semantics)
- Per-direction independence in bidirectional edges

No issues found.

lsm (commented):

---

**Recommendation: APPROVE** (self-PR — posting as COMMENT per policy)

### Previous P2 feedback — resolution status

All three P2 issues from the prior review round were addressed. One has a minor residual gap worth noting.

#### P2.1 — New test for `??=` undefined propagation

Resolved. The test "second gate label wins when first gate has no label" at lines 482-497 correctly exercises the case where the first gate has `label: undefined`, allowing the second gate's label to propagate via `??=`. The assertion correctly uses `.toBe('Second')` and `.toBe('#00bb00')` rather than `toMatchObject`. The test behaviour is technically sound — `channels.entries()` iterates in array order, so `gate-a` is processed first, `aggregate.lowToHighGateLabel` becomes `undefined`, and the second iteration assigns `'Second'` because `undefined ??= 'Second'` evaluates to `'Second'`.

#### P2.2 — `toMatchObject({ key: undefined })` replaced in 3 places

Resolved for the 3 tests the previous reviewer specifically flagged ("missing gate", "no gateId", "script-only gate"). All three now use explicit `.toBeUndefined()` calls.

However, the **same pattern persists in 5 other locations** introduced by the first commit that were not flagged before and were not changed in the fix commit:

- Line 229: `hasScript: undefined` inside `toMatchObject` in "propagates gate label and color for a one-way edge"
- Line 313: `reverseHasScript: undefined` in "propagates label/color/hasScript for bidirectional edges per direction"
- Lines 361-369: 4 `undefined` values in "tracks label/color/hasScript only for forward direction"
- Lines 387-392: 4 `undefined` values in "tracks label/color/hasScript only for reverse direction"
- Lines 438, 454: individual `undefined` values in label/color-only tests

These all carry the same theoretical weakness: Vitest's `toMatchObject` does not distinguish between a key being absent and a key being explicitly `undefined`. If a future refactor accidentally removes a field from the returned object, these assertions would still pass. The fix was scoped to exactly the 3 tests that were raised — the residual cases are a known consistency gap.

#### P2.3 — Rename `gate-no-script` to `gate-1`

Resolved. `gate-no-script` is now `gate-1`, `gate-with-script` is `gate-2`, and the misleading `label: 'No Script'` override is gone. The `hasScript` assertion was also upgraded from `toMatchObject({ hasScript: true })` to `expect(result[0].hasScript).toBe(true)`, consistent with the P2.2 fix.

---

### Summary

All three stated P2 issues were addressed as described in the commit message. The 3 targeted `toMatchObject({ key: undefined })` uses were replaced with explicit `.toBeUndefined()`. The residual instances of the same pattern in the other new tests (lines 229, 313, 361-369, 387-392, 438, 454) were not part of the previous feedback and remain — they represent a minor consistency gap but do not affect test correctness for the current implementation. The new `??=` undefined-propagation test is correct and well-targeted. The rename is clean.

PR #1211, opened 2026-04-02
feat(web): add script editor to GateEditorPanel

lsm:
## Summary
- Add "Script Check" section to GateEditorPanel with toggle switch, interpreter dropdown (bash/node/python3), monospace source textarea, timeout input (default 30s, max 120s)
- Include Lint Check and Type Check script presets that output valid JSON to stdout
- Add client-side validation for empty source and timeout bounds
- Support script-only gates (no fields) in the editor

## Test plan
- [x] 63 unit tests passing (34 existing + 29 new for script editor)
- [ ] Manual: verify script toggle, interpreter selection, source editing, timeout clamping
- [ ] Manual: verify presets populate form correctly
- [ ] Manual: verify script-only gate works with no fields

### DIFF
--- a/packages/web/src/components/space/visual-editor/GateEditorPanel.tsx
+++ b/packages/web/src/components/space/visual-editor/GateEditorPanel.tsx
@@ -1,5 +1,5 @@
 import { useMemo, useState } from 'preact/hooks';
-import type { Gate, GateField, GateFieldType, GateFieldCheck } from '@neokai/shared';
+import type { Gate, GateField, GateFieldType, GateFieldCheck, GateScript } from '@neokai/shared';
 
 export interface GateEditorPanelProps {
 	gate: Gate;
@@ -23,6 +23,22 @@ const BADGE_RX = 10;
 const BADGE_CHAR_WIDTH = 7;
 const BADGE_PADDING = 8;
 const HEX_COLOR_RE = /^#[0-9a-fA-F]{6}$/;
+const SCRIPT_INTERPRETERS: GateScript['interpreter'][] = ['bash', 'node', 'python3'];
+const SCRIPT_TIMEOUT_DEFAULT = 30;
+const SCRIPT_TIMEOUT_MAX = 120;
+
+const SCRIPT_PRESETS = {
+	lint: {
+		label: 'Lint Check',
+		interpreter: 'bash' as const,
+		source: 'npm run lint 2>/dev/null && echo \'{"passed":true}\' || echo \'{"passed":false}\'',
+	},
+	typecheck: {
+		label: 'Type Check',
+		interpreter: 'node' as const,
+		source: 'console.log(JSON.stringify({passed: true}))',
+	},
+} as const;
 
 // ---------------------------------------------------------------------------
 // Lightweight client-side validation (mirrors daemon gate-evaluator.ts)
@@ -38,6 +54,27 @@ function validateLabel(value: string | undefined): string {
 function validateColor(value: string | undefined): string {
 	if (!value) return '';
 	if (!HEX_COLOR_RE.test(value)) return `color: expected hex format #rrggbb, got "${value}"`;
+	return '';
+}
+
+function validateScriptInterpreter(value: string | undefined): string {
+	if (!value) return 'interpreter is required';
+	if (!SCRIPT_INTERPRETERS.includes(value as GateScript['interpreter'])) {
+		return `interpreter: expected one of [bash, node, python3], got "${value}"`;
+	}
+	return '';
+}
+
+function validateScriptSource(value: string | undefined): string {
+	if (!value || value.trim().length === 0) return 'source: script body is required';
+	return '';
+}
+
+function validateScriptTimeout(value: number | undefined): string {
+	if (value === undefined || value === null) return '';
+	if (typeof value !== 'number' || isNaN(value)) return 'timeout: must be a number';
+	if (value < 1) return 'timeout: must be at least 1 second';
+	if (value > SCRIPT_TIMEOUT_MAX) return `timeout: must be at most ${SCRIPT_TIMEOUT_MAX} seconds`;
 	return '';
 }
 
@@ -65,6 +102,28 @@ export function GateEditorPanel({
 	// Validation errors computed from current gate state
 	const labelError = useMemo(() => validateLabel(gate.label), [gate.label]);
 	const colorError = useMemo(() => validateColor(gate.color), [gate.color]);
+
+	// Script state: track enabled, interpreter, source, and timeout (in seconds)
+	const scriptEnabled = !!gate.script;
+	const scriptInterpreter = gate.script?.interpreter ?? 'bash';
+	const scriptSource = gate.script?.source ?? '';
+	const scriptTimeoutSec = gate.script?.timeoutMs
+		? Math.round(gate.script.timeoutMs / 1000)
+		: SCRIPT_TIMEOUT_DEFAULT;
+
+	// Script validation errors (only shown when script is enabled)
+	const scriptInterpreterError = useMemo(
+		() => (scriptEnabled ? validateScriptInterpreter(scriptInterpreter) : ''),
+		[scriptEnabled, scriptInterpreter]
+	);
+	const scriptSourceError = useMemo(
+		() => (scriptEnabled ? validateScriptSource(scriptSource) : ''),
+		[scriptEnabled, scriptSource]
+	);
+	const scriptTimeoutError = useMemo(
+		() => (scriptEnabled ? validateScriptTimeout(scriptTimeoutSec) : ''),
+		[scriptEnabled, scriptTimeoutSec]
+	);
 
 	// Derived badge preview values
 	const badgeLabel = gate.label ?? 'Gate';
@@ -116,6 +175,36 @@ export function GateEditorPanel({
 			check: { op: '==', value: 'passed' },
 		};
 		updateGate({ fields: [...(gate.fields ?? []), preset] });
+	}
+
+	function toggleScriptEnabled(checked: boolean) {
+		if (checked) {
+			updateGate({
+				script: { interpreter: 'bash', source: '', timeoutMs: SCRIPT_TIMEOUT_DEFAULT * 1000 },
+			});
+		} else {
+			updateGate({ script: undefined });
+		}
+	}
+
+	function updateScriptPartial(partial: Partial<GateScript>) {
+		const current = gate.script ?? {
+			interpreter: 'bash',
+			source: '',
+			timeoutMs: SCRIPT_TIMEOUT_DEFAULT * 1000,
+		};
+		updateGate({ script: { ...current, ...partial } });
+	}
+
+	function applyScriptPreset(key: keyof typeof SCRIPT_PRESETS) {
+		const preset = SCRIPT_PRESETS[key];
+		updateGate({
+			script: {
+				interpreter: preset.interpreter,
+				source: preset.source,
+				timeoutMs: SCRIPT_TIMEOUT_DEFAULT * 1000,
+			},
+		});
 	}
 
 	return (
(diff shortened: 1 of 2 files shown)


lsm (commented):

Overall this is a solid implementation. The script editor section integrates well with the existing GateEditorPanel patterns, the toggle/interpreter/source/timeout controls are well-structured, and the 29 new tests provide good coverage. A few issues worth addressing:

### P1: NaN propagation in timeout input handler

**File:** `packages/web/src/components/space/visual-editor/GateEditorPanel.tsx`, lines 517-519

```js
const val = Number((e.currentTarget as HTMLInputElement).value);
const clamped = Math.max(1, Math.min(SCRIPT_TIMEOUT_MAX, val));
updateScriptPartial({ timeoutMs: clamped * 1000 });
```

If the user types a non-numeric string (or the input is in an intermediate state), `Number(value)` returns `NaN`. `Math.max(1, NaN)` evaluates to `NaN`, so `NaN * 1000 = NaN` gets stored as `timeoutMs`. This silently corrupts the gate state. Add a NaN guard:

```js
const val = Number((e.currentTarget as HTMLInputElement).value);
if (isNaN(val)) return;
const clamped = Math.max(1, Math.min(SCRIPT_TIMEOUT_MAX, val));
```

### P2: "No fields - gate always opens" text is misleading when script is enabled

**File:** `packages/web/src/components/space/visual-editor/GateEditorPanel.tsx`, line 377

The empty-fields message says "No fields -- gate always opens," but when a script check is enabled, the gate does not always open -- it depends on the script result. Consider conditionally adjusting this text, e.g., "No fields defined" or "No fields -- gate passes via script check" when `gate.script` is set.

### P3: Type Check preset is a no-op placeholder

The Type Check preset at line 39 (`console.log(JSON.stringify({passed: true}))`) always outputs `{passed: true}` without actually running a type checker. The Lint Check preset at least invokes `npm run lint`. Consider making the Type Check preset actually run `npx tsc --noEmit` or similar, or labeling it more clearly as a template/skeleton.

**Recommendation: REQUEST_CHANGES** (due to P1 NaN bug)

lsm (commented):

**Recommendation: REQUEST_CHANGES** (2 P1, 2 P2, 1 P3)

---

### P1 — Missing gate-level validation: empty gate (no fields, no script)

`packages/web/src/components/space/visual-editor/GateEditorPanel.tsx`

The daemon's `validateGate` in `gate-evaluator.ts` (line 249) enforces:

> `gate: must have at least one non-empty "fields" array or a "script"`

The panel currently shows `"No fields — gate always opens"` when fields is empty and script toggle is off, but emits no validation error. A user can click save/confirm with zero fields and no script, producing a gate the daemon will refuse. The TODO comment in gate-evaluator.ts line 221 explicitly calls out `GateEditorPanel` as a wire-up point — it wasn't done.

Concretely: add a gate-level error (e.g. `data-testid="gate-editor-gate-error"`) shown whenever `(gate.fields ?? []).length === 0 && !scriptEnabled`, and add a corresponding test.

---

### P1 — `validateScriptTimeout` dead code + confusing null guard

`packages/web/src/components/space/visual-editor/GateEditorPanel.tsx`, line 73–78

```ts
function validateScriptTimeout(value: number | undefined): string {
    if (value === undefined || value === null) return '';
    ...
}
```

The `value === null` branch is dead code — the TypeScript signature is `number | undefined`. More importantly, when `scriptEnabled` is true, `scriptTimeoutSec` is always a number (either from `Math.round(...)` or the default 30). Passing `undefined` to this function while `scriptEnabled` is true is not reachable. The confusing part is that the function silently accepts `undefined` as valid — if the call site ever changes, this becomes a silent validation gap. Either tighten the type to `number` and remove the guard, or document the invariant.

---

### P2 — "Type Check" preset is a non-functional stub

`packages/web/src/components/space/visual-editor/GateEditorPanel.tsx`, lines 38–40

```ts
typecheck: {
    label: 'Type Check',
    interpreter: 'node' as const,
    source: 'console.log(JSON.stringify({passed: true}))',
},
```

This script always returns `{passed: true}` unconditionally — it never actually runs `tsc` or any type-check tool. Users clicking "Type Check" will get a gate that always passes regardless of whether types are actually valid. The Lint Check preset at least shells out to `npm run lint`. Consider replacing with something like:

```bash
npx tsc --noEmit 2>/dev/null && echo '{"passed":true}' || echo '{"passed":false}'
```

(interpreter: bash). If a meaningful node-based equivalent is preferred, it should actually invoke a type-checker, not return a hardcoded success.

---

### P2 — Validation mismatch: backend checks `source.length === 0` but frontend trims

`packages/web/src/components/space/visual-editor/GateEditorPanel.tsx`, line 69

```ts
function validateScriptSource(value: string | undefined): string {
    if (!value || value.trim().length === 0) return 'source: script body is required';
    return '';
}
```

The backend (`gate-evaluator.ts`, line 192) rejects `source.length === 0` but does NOT trim — a source of `"   "` (whitespace only) would pass backend validation but show an error in the frontend. This is technically a stronger frontend constraint (good), but it creates an inconsistency that could confuse debugging. The behavior is asymmetric: the UI refuses whitespace-only sources, the daemon stores and executes them. The test at line 673 explicitly tests whitespace-only rejection, reinforcing this frontend-only constraint. If the intent is to also reject whitespace-only scripts at the daemon, the `validateGateScript` function needs a `source.trim().length === 0` check. If not, a comment here explaining the intentional divergence would help.

---

### P3 — No test for the "empty gate" validation constraint

`packages/web/src/components/space/visual-editor/__tests__/GateEditorPanel.test.tsx`

Once the gate-level validation is added (see P1 above), there should be tests for: (a) error shown when fields empty and script disabled, (b) error clears when script is enabled, (c) error clears when a field is added. These are currently absent because the validation itself is absent.

---

### Minor observations (informational)

- **Preset overwrites timeout silently**: `applyScriptPreset` always resets `timeoutMs` to 30s. If a user customizes the timeout first then clicks a preset, it resets to 30s. This is arguably fine but could be surprising — worth noting in a comment.
- **Script presets hidden when toggle off**: the `gate-editor-preset-lint` and `gate-editor-preset-typecheck` buttons are correctly hidden when script is disabled, confirmed by the tests at line 651.

lsm (commented):

**Recommendation: APPROVE**

The follow-up commit (`aa1a055`) cleanly addresses all five items from the previous review. Here is my assessment of each fix and the overall state of the code:

### Previous feedback — resolved

1. **NaN guard in timeout onInput** (line 546-547): The `isNaN(val)` check now short-circuits before clamping, preventing `NaN * 1000` from propagating into `timeoutMs`. Correct.

2. **`validateGateCompleteness()` + `gateError`** (lines 84-89, 124-129): Gate-level validation now warns when neither fields nor script are present, rendered at `data-testid="gate-editor-gate-error"`. The `useMemo` dependency array `[hasFields, scriptEnabled]` is correct since both are derived from `gate` props.

3. **Context-sensitive empty-fields message** (lines 401-403): `"No fields — gate always opens"` vs `"No fields defined"` depending on `scriptEnabled`. Good UX improvement — no longer misleading when a script check is active.

4. **Type Check preset uses `bash` + `npx tsc --noEmit`** (line 38-40): No longer hardcodes `node` interpreter. This is the correct approach since `tsc` is a CLI tool, not a Node script.

5. **`validateScriptTimeout` type signature** (line 77): Parameter is now `number` (was previously `number | undefined` per the diff). Clean and correct since the caller always provides a number.

### New tests — 9 added (72 total), all passing

The new tests cover:
- Gate-level validation for empty gate, undefined fields, fields-present, script-present, and script-toggled-off scenarios
- Context-sensitive empty-fields messages (3 tests)
- Script-only gate end-to-end test

Test quality is good — they test both presence and absence of error elements, boundary conditions, and rerender behavior.

1. **Line 793-801 — test double-renders**: The `"shows gate error after script is toggled off"` test clicks the toggle, then creates a new gate and renders it separately to verify the error. This works because the component is controlled (parent owns state), but the second `render()` call is technically a new component instance — the `rerender` variable from the second `render()` shadows the one from the first. The test still validates the right thing (that a gate with no fields and no script shows the error), but could be slightly cleaner by using a single render with rerender.

2. **Timeout clamping vs validation divergence**: The `onInput` handler clamps values to `[1, 120]` (line 547), but `validateScriptTimeout` reports errors for out-of-range values (lines 79-80). In practice, the validation errors for timeout can only appear when the gate is loaded from props with an already-invalid `timeoutMs` (since user input is always clamped). This is fine — defensive validation on props is reasonable.

No P0, P1, or P2 issues found. The implementation is correct, complete, well-tested, and type-safe.

lsm (commented):

---

**Recommendation: APPROVE** (with minor follow-up notes)

All five items from the previous review have been addressed. All 72 tests pass. The implementation is correct and the fixes are targeted. A few small observations follow.

---

### What was fixed correctly

1. **NaN guard in timeout `onInput`** — `if (isNaN(val)) return;` added before clamping. Correct placement, guards against edge cases.

2. **`validateGateCompleteness()` + `gateError`** — clean separation of concerns, correct `useMemo` usage, `data-testid="gate-editor-gate-error"` present and tested.

3. **Context-sensitive empty-fields message** — "No fields defined" vs "No fields — gate always opens" now depends on `scriptEnabled`. All three states are covered by tests.

4. **Type Check preset** — changed from `node` + hardcoded `console.log` to `bash` + `npx tsc --noEmit`. Consistent with the lint preset pattern and better for real-world project setups.

5. **`validateScriptTimeout` signature** — narrowed from `number | undefined` to `number`. The `undefined` case is now handled upstream via the `scriptEnabled` guard on the calling `useMemo`.

---

### P2 — NaN guard test doesn't exercise the NaN branch

`packages/web/src/components/space/visual-editor/__tests__/GateEditorPanel.test.tsx`, test "NaN guard prevents NaN from propagating as timeoutMs" (line 608).

The test fires input with `{ value: '' }`. On `<input type="number">`, `Number('')` evaluates to `0`, not `NaN`. `isNaN(0)` is `false`, so the `if (isNaN(val)) return;` guard is never reached. The test verifies that `0` gets clamped to `1`, which is already covered by the "clamps timeout value below 1 to 1" test just above it.

The guard itself is valid and correct in the production code. But the test's intent — to verify that NaN does not propagate — is not met. `Number('abc')` would produce `NaN` and hit the guard, but happy-dom sanitizes that to `''` on a number input, making it hard to test directly. The test comment acknowledges this limitation but stops short of testing the actual guard branch. This is a test accuracy issue rather than a correctness issue.

---

### P2 — "toggle off with no fields" test is structurally misleading

`packages/web/src/components/space/visual-editor/__tests__/GateEditorPanel.test.tsx`, lines 783-803.

The test title implies it verifies that toggling the script switch off causes the gate error to appear. But because `GateEditorPanel` is a controlled component (its state lives in the parent via `onChange`), clicking the toggle only calls `onChange` — the prop does not update in the same render. The test accounts for this correctly in principle, but the implementation mounts a second component (`render(...)` on line 800) rather than using `rerender` on the first. The `queryByTestId` on line 802 comes from the first render's closure but scans the full document, which now contains two mounted components. The assertion passes because the second component's DOM is found — not because the toggle caused an update on the first.

The test is not wrong in effect (it does confirm that the gate error appears when `script === undefined`), but it is testing a separate render, not the toggle interaction it claims to test. Consider: use the `rerender` returned by the original `render()` call, or split into two tests.

---

### P3 — Preset overwriting custom timeout has no negative-case test

The NOTE comment on `applyScriptPreset` (line 216) documents that presets reset the timeout to 30s. The two preset tests only use gates that already have `timeoutMs: 30000`, so they don't verify the reset behavior from a non-default value. A small additional case such as: start with `timeoutMs: 60000`, apply a preset, assert `timeoutMs === 30000` would fully pin this behavior.

---

### Summary

The second iteration is a clean improvement. All substantive feedback from the first review has been addressed. The two P2 issues above are test quality concerns only — production logic is correct. The implementation is ready to merge.

PR #1212, opened 2026-04-02
fix: prevent duplicate leader/worker question messages

lsm:
## Summary
- Add idempotency guard in `onLeaderTerminalState` and `onWorkerTerminalState` to skip duplicate `waiting_for_input` events when the group is already marked as waiting for a question from the same session role
- SessionObserver is a stateless relay that fires on every `session.updated` event (e.g. from draft saves via `updateQuestionDraft()`), which caused duplicate "asked a question" group events to accumulate linearly with each occurrence

## Root cause
`ProcessingStateManager.updateQuestionDraft()` emits a `session.updated` DaemonHub event with `status: 'waiting_for_input'` on every draft save. Since `SessionObserver` fires its callback on every matching event without deduplication, each draft save re-triggered `onLeaderTerminalState`/`onWorkerTerminalState`, which appended another status event to `task_group_events` without checking if one already existed.

## Test plan
- Added unit tests verifying that calling `onWorkerTerminalState`/`onLeaderTerminalState` with `waiting_for_input` multiple times produces only one group event
- All 1914 existing room unit tests pass

### DIFF
--- a/packages/daemon/src/lib/room/runtime/room-runtime.ts
+++ b/packages/daemon/src/lib/room/runtime/room-runtime.ts
@@ -640,7 +640,16 @@ export class RoomRuntime {
 
 		// Check if worker is waiting for user input (asked a question)
 		// Pause routing to leader — task resumes when question is answered
+		// Guard: skip if already waiting — SessionObserver is a stateless relay that fires
+		// on every session.updated event (e.g. draft saves), so we must not append duplicate
+		// group events each time.
 		if (terminalState.kind === 'waiting_for_input') {
+			if (group.waitingForQuestion && group.waitingSession === 'worker') {
+				log.debug(
+					`Worker ${group.workerSessionId} already waiting for input — skipping duplicate event`
+				);
+				return;
+			}
 			log.info(`Worker ${group.workerSessionId} is waiting for user input - pausing task`);
 			this.groupRepo.setWaitingForQuestion(groupId, true, 'worker');
 			this.appendGroupEvent(groupId, 'status', {
@@ -1042,7 +1051,16 @@ export class RoomRuntime {
 
 		// Check if leader is waiting for user input (asked a question)
 		// Pause — task resumes when question is answered
+		// Guard: skip if already waiting — SessionObserver is a stateless relay that fires
+		// on every session.updated event (e.g. draft saves), so we must not append duplicate
+		// group events each time.
 		if (terminalState.kind === 'waiting_for_input') {
+			if (group.waitingForQuestion && group.waitingSession === 'leader') {
+				log.debug(
+					`Leader ${group.leaderSessionId} already waiting for input — skipping duplicate event`
+				);
+				return;
+			}
 			log.info(`Leader ${group.leaderSessionId} is waiting for user input - pausing task`);
 			this.groupRepo.setWaitingForQuestion(groupId, true, 'leader');
 			this.appendGroupEvent(groupId, 'status', {
--- a/packages/daemon/tests/online/cross-provider/cross-provider-model-switch.test.ts
+++ b/packages/daemon/tests/online/cross-provider/cross-provider-model-switch.test.ts
@@ -740,7 +740,7 @@ describe('Cross-Provider Model Switching (MiniMax <-> GLM)', () => {
 			const initialModel = initialSystemInit.model as string | undefined;
 			expect(initialModel).toBeDefined();
 
-			await waitForIdle(daemon, sessionId, 60000);
+			await waitForIdle(daemon, sessionId, 90000);
 
 			// --- Phase 2: Switch model (cross-provider GLM → MiniMax) ---
 			const switchResult = (await daemon.messageHub.request('session.model.switch', {
@@ -757,7 +757,7 @@ describe('Cross-Provider Model Switching (MiniMax <-> GLM)', () => {
 			const postSwitchSystemInit = await postSwitchSystemInitPromise;
 
 			const postSwitchModel = postSwitchSystemInit.model as string | undefined;
-			await waitForIdle(daemon, sessionId, 60000);
+			await waitForIdle(daemon, sessionId, 90000);
 
 			expect(postSwitchModel).toBeDefined();
 			expect(postSwitchModel).not.toBe(initialModel);
--- a/packages/daemon/tests/unit/room/room-runtime-flow.test.ts
+++ b/packages/daemon/tests/unit/room/room-runtime-flow.test.ts
@@ -226,6 +226,44 @@ describe('RoomRuntime flow', () => {
 			expect(afterResume.waitingForQuestion).toBe(false);
 			expect(afterResume.waitingSession).toBeNull();
 		});
+
+		it('should not append duplicate group events on repeated waiting_for_input (idempotency)', async () => {
+			await createGoalAndTask(ctx);
+			ctx.runtime.start();
+			await ctx.runtime.tick();
+
+			const groups = ctx.groupRepo.getActiveGroups('room-1');
+			const group = groups[0];
+
+			// First waiting_for_input — should append a status event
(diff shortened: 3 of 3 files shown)


lsm (commented):

**Recommendation: APPROVE**

Clean, minimal fix for the listener leak bug. The implementation is correct:

1. **Guard logic is sound.** Both `onWorkerTerminalState` and `onLeaderTerminalState` check `group.waitingForQuestion && group.waitingSession === '<role>'` before appending a duplicate group event. The condition is tight — it only guards the specific role, so a worker waiting won't block a leader event or vice versa.

2. **Early return is safe.** The code after the guard (setWaitingForQuestion, appendGroupEvent, emitTaskUpdateById) is all idempotent state-setting that was already applied on the first call, so returning early loses nothing.

3. **Tests are well-structured.** Both worker and leader paths are tested with triple invocations, asserting exactly one status event and correct waiting state. The test comments explain the real-world trigger (SessionObserver firing on draft saves).

4. **Comments explain the "why."** The inline comments in the production code explain why the guard is needed (SessionObserver is a stateless relay), which will help future maintainers.

No issues found.

lsm (commented):

**Recommendation: APPROVE**

---

### Summary

The fix is minimal, correct, and well-targeted. The idempotency guards in both `onWorkerTerminalState` and `onLeaderTerminalState` use the existing `waitingForQuestion`/`waitingSession` fields — state that was already being maintained — so there is no new state machine complexity introduced. Two unit tests confirm the core invariant.

---

### Correctness Analysis

**State machine lifecycle is preserved.**
The guard fires only when `waitingForQuestion=true AND waitingSession === '<role>'`. The flag is cleared on any non-`waiting_for_input` terminal event (lines 663-665 and 1074-1075), so a genuinely new question in a later turn correctly bypasses the guard after the human answers. No regression risk here.

**Early return skips the right things.**
On the second+ `waiting_for_input` in `onWorkerTerminalState`, the guard returns before re-calling `setWaitingForQuestion`, `appendGroupEvent`, and `emitTaskUpdateById`. The `activeSession` clear-and-emit at lines 636-639 is above the guard and fires unconditionally for the relevant condition, so task state is still updated if needed. Same structure holds in `onLeaderTerminalState` (lines 1043-1050 above the guard).

**Root cause is accurately described in the PR.**
`SessionObserver` is indeed stateless and fires on every `session.updated` event including draft saves via `updateQuestionDraft()`. The guard is the correct layer to absorb the noise.

---

### Test Coverage

**What is covered:**
- Worker: three repeated `waiting_for_input` calls produce exactly one status event, and the group remains marked `waitingForQuestion=true, waitingSession='worker'`.
- Leader: same test mirrored for the leader path.
- Pre-existing test at line 200 covers the resume-after-question flow (flag clears on `idle`).

**Missing test — ask-answer-ask-again sequence (P3).**
There is no test verifying that after the human answers (clearing the flag via an `idle` terminal event), a subsequent `waiting_for_input` is allowed through and produces a second group event. The logic is clearly correct by reading the code, but the test gap means a future refactor could silently break this without a failing test. Low priority given the simplicity of the guard.

---

### Minor Observations

**`payloadJson` string-contains check in tests (cosmetic, P3).**
`events.filter(e => e.payloadJson?.includes('Worker asked a question'))` is correct since `payloadJson` is stored as a raw JSON string (confirmed in `session-group-repository.ts` line 975). No issue.

**`spawnAndRouteToLeader` correctly sets `leaderHasWork=true` (confirmed).**
The `onWorkerTerminalState(kind: 'idle')` call in the helper triggers `routeWorkerToLeader`, which sets `leaderHasWork=true` before injecting. The leader idempotency test correctly passes the existing `leaderHasWork` guard.

**Cross-role overwrite is not introduced by this PR.**
A theoretical edge case where `waitingForQuestion=true, waitingSession='leader'` and `onWorkerTerminalState(waiting_for_input)` fires would bypass the guard and overwrite the leader's state. This pre-existed before this PR and is not made worse by the change.

---

### Verdict

The implementation is correct, targeted, and well-explained. CI passes across all available checks. The only gap is the absence of a "second question" round-trip test, which is a P3 nit given the logic is trivially verifiable by code reading.

lsm (commented):

### Verdict: APPROVE

Clean, well-scoped fix. The root cause analysis is sound: `SessionObserver` fires on every `session.updated` event including draft saves, which re-triggers the terminal state handlers and appends duplicate group events.

**Correctness:** The idempotency guard reads `group` fresh from the repository at the top of each handler (`this.groupRepo.getGroup(groupId)`), then checks `waitingForQuestion` and `waitingSession` before the duplicate `setWaitingForQuestion` + `appendGroupEvent` calls. Since SQLite ops are synchronous and single-threaded in Bun, there is no TOCTOU race here. The guard correctly returns early, preserving the existing waiting state without side effects.

**Symmetry:** Both `onWorkerTerminalState` (line 647) and `onLeaderTerminalState` (line 1058) receive the same treatment, each checking for the appropriate session role (`'worker'` vs `'leader'`). The logic after the guard (clearing the waiting flag on non-`waiting_for_input` terminal states at lines 663 and 1074) remains unaffected.

**Tests:** Two new unit tests cover the exact scenario -- calling the handler 3 times and asserting only 1 group event is created. They verify both the event count and the persisted group state. Good coverage for the fix.

**Timeout bump:** The `waitForIdle` increase from 60s to 90s in the cross-provider MiniMax test is a reasonable response to observed 68.8s runs against a 120s test-level timeout. No concerns there.

No issues found.

lsm (commented):

**Recommendation: APPROVE**

### Summary

The fix correctly addresses the root cause. `SessionObserver` is intentionally a stateless relay (its own doc comment already says "Consumers must be idempotent"), and the guard is placed at precisely the right point in both `onWorkerTerminalState` and `onLeaderTerminalState`: after the harmless `clearActiveSession` no-op but before any event is appended or state changed. The two unit tests cover the duplicate-suppression behavior and also confirm the group state remains `waitingForQuestion=true` after the guard fires. The timeout bump from 60 s to 90 s in the cross-provider test is a straightforward latency-budget fix with an overall test timeout of 120 s, so there is enough headroom.

### Correctness

The guard condition `group.waitingForQuestion && group.waitingSession === 'worker'` (and the symmetric leader case) is exact. It will only skip the event if the *same role* is already parked in the waiting state, so a transition from worker-waiting to leader-waiting (or any other permutation) is never accidentally suppressed. The early return does not skip the `clearActiveSession` call because that block executes unconditionally above the guard. On the second and subsequent duplicate events `task.activeSession` is already `null`, so the `updateTaskStatus` call is a guarded no-op.

### Test coverage

Both new test cases exercise three consecutive `waiting_for_input` calls (first + two duplicates) and assert that exactly one status event is stored and that `waitingForQuestion`/`waitingSession` are still set correctly after the guards fire. The happy-path round-trip test (question → idle clears the flag) was already present and is unaffected.

**P3 - test assertion uses `payloadJson?.includes(...)` string match:** The tests assert the status event by matching the raw serialised `payloadJson` string (`e.payloadJson?.includes('Worker asked a question')`). If the event text in room-runtime.ts ever changes, the filter would produce 0 matches and `expect([]).toHaveLength(1)` would still catch the regression correctly — so there is no silent-pass risk. It is slightly fragile to couple the test to the exact serialised string, but it is not a correctness problem.

**P3 - `waitForIdle` budget vs outer test timeout:** Both `waitForIdle` calls are now 90 s, and the test itself has a 120 s outer timeout. That leaves only 30 s total margin across both phases before Bun kills the test. If Minimax latency increases further, the test will hit the outer timeout producing a less informative error message. Consider bumping the outer timeout to 180 s to maintain the same per-phase margin, though this is a judgement call on acceptable CI wall time.

Both observations are nits. The implementation is correct, the root cause analysis matches the code, and the fix is minimal and well-scoped.

